Issue #1168, opened 2014-11-26
Exploratory Testing: DI & configuration


rowanmiller:

mikary:
Configuration

connection strings:
* [x] From a single Configuration file
 * [x] No connection string definition
 * [x] ConnectionString defined in context
 * [x] ~~ConnectionStringKey defined in context~~
 * [ ] ConnectionString "Name=xxx" syntax
 * [x] conflicting definition (connection string and key defined) #1391
* [x] Multiple configuration files / overrides #1390
 * [x] Name / Full Name #1395
 * [x] conflicting definition (connection string and key defined in different files) #1390

MaxBatchSize (Relational):
* [x] Load from configuration
* [x] Range validation

CommandTimeout (Relational):
* [x] Load from configuration #1422
* [x] Range validation

DbContextOptionsParser Parser (RawOptions dictionary):
* [x] Key Lookup
* [x] Recursive path (nested properties appear in dictionary)

Migrations:
* Parameterless constructor only (Definition in Startup.cs)
 * [x] Using Configuration
 * [x] Using AddDbContext Action
 * [x] Using OnConfiguring
 * [x] Using Configuration and AddDbContext Action (Action override)
 * [x] Using Configuration, AddDbContext Action, and OnConfiguring override (OnConfiguring 

DI

DbContext constructors and patterns:
* Default
 * [x] Using Configuration
 * [x] Using AddDbContext Action
 * [x] Using OnConfiguring
 * [x] Using Configuration and AddDbContext Action (Action overrides)
 * [x] Using Configuration, AddDbContext Action, and OnConfiguring override (OnConfiguring overrides)
* DbContextOptions
 * [x] Using Configuration (InvalidOperationException)#1401
 * [x] Using AddDbContext Action (InvalidOperationException)
* IServiceProvider
 * [x] Using Configuration
 * [x] Using AddDbContext Action
 * [x] Using Configuration and AddDbContext Action (Action overrides)
* IServiceProvider and DbContextOptions
 * [x] Using Configuration (InvalidOperationException)
 * [x] Using AddDbContext Action (InvalidOperationException)

Verify that we are not creating additional service providers or using the wrong service provider
* Console Application
 * [x] Custom service provider and implicit service provider
  * ServiceProvider.GetService uses calling provider
  * new Context() uses new instance first call and cached for subsequent calls

* Web Application
 * [x] Service provider configured in Startup.CS, injected in Controller
[...]

PR #1422, opened 2015-01-17
Loading CommandTimeout from Configuration, enhanced OptionsExtensions tests

mikary:
While working on #1168, I discovered that CommandTimeout settings were not being loaded from Configuration. This PR adds the missing functionality, and unit tests. I also reworked the unit tests for the ```SqlServerOptionsExtensions``` for consistency and more complete coverage of the configuration behaviors.
@lukewaters @ajcvickers

### DIFF
--- a/src/EntityFramework.Relational/RelationalOptionsExtension.cs
+++ b/src/EntityFramework.Relational/RelationalOptionsExtension.cs
@@ -14,10 +14,13 @@ namespace Microsoft.Data.Entity.Relational
     public abstract class RelationalOptionsExtension : DbContextOptionsExtension
     {
         private const string ConnectionStringKey = "ConnectionString";
+        private const string CommandTimeoutKey = "CommandTimeout";
+        private const string MaxBatchSizeKey = "MaxBatchSize";
 
         private string _connectionString;
         private DbConnection _connection;
         private int? _commandTimeout;
+        private int? _maxBatchSize;
 
         public virtual string ConnectionString
         {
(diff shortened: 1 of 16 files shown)


src/EntityFramework.SqlServer/SqlServerOptionsExtension.cs
```diff
@@ -36,14 +40,14 @@ protected override void Configure(IReadOnlyDictionary<string, string> rawOptions
             base.Configure(rawOptions);
             if (!_maxBatchSize.HasValue)
             {
-                var MaxBatchSizeConfigurationKey = ProviderPrefix + ":" + MaxBatchSizeKey;
+                var maxBatchSizeConfigurationKey = ProviderPrefix + Constants.KeyDelimiter + MaxBatchSizeKey;
                 string maxBatchSizeString;
```
ajcvickers: Why do we look for ProviderPrefix here, but not for the command timeout case?

src/EntityFramework.SqlServer/SqlServerOptionsExtension.cs
```diff
@@ -36,14 +40,14 @@ protected override void Configure(IReadOnlyDictionary<string, string> rawOptions
...
+                if (rawOptions.TryGetValue(maxBatchSizeConfigurationKey, out maxBatchSizeString))
                 {
                     int maxBatchSizeInt;
                     if (!Int32.TryParse(maxBatchSizeString, out maxBatchSizeInt))
                     {
-                        throw new InvalidOperationException(Strings.IntegerConfigurationValueFormatError(MaxBatchSizeConfigurationKey, maxBatchSizeString));
+                        throw new InvalidOperationException(Strings.IntegerConfigurationValueFormatError(maxBatchSizeConfigurationKey, maxBatchSizeString));
                     }
                     _maxBatchSize = maxBatchSizeInt;
                 }
```
ajcvickers: Can we reuse the parsing code from above--the two pieces of code seem almost identical.

ajcvickers:
I am lost as to where we have ended up with MaxBatchSize and CommandTimeout. Can you summarize all the places each can be set, if and how they are provider-specific, and what we consider valid values and where we check for these?

mikary:
Had a talk with @rowanmiller and @divega about the Relational vs Provider-Specific config settings. We are thinking that the ```CommandTimeout``` and ```MaxBatchSize``` should both be specified at the top (```EntityFramework:DbContext```) level in config but loaded in by provider-specific (```SqlServerOptionsExtension```) code.

For this change, we will put both public properties on the ```RelationalOptionsExtension``` but have the code for loading the settings on ```SqlServerOptionsExtension```. The ```ProviderPrefix``` will go away for now, but may be added again at a later date for provider-specific overrides on the relational settings.

mikary:
Correction: The relational settings will be parsed and stored by the abstract class ```RelationalOptionsExtensions``` , and it will be up to the individual providers to use or ignore these settings on the provider-specific derived types.

src/EntityFramework.Relational/RelationalOptionsExtension.cs
```diff
@@ -47,24 +50,54 @@ public virtual DbConnection Connection
         public virtual int? CommandTimeout
         {
             get { return _commandTimeout; }
+            [param: CanBeNull]
             set
             {
                 if (value.HasValue && value < 0)
```
ajcvickers: Can command timeout be zero? If so, the exception message is wrong. If not, the check here is wrong.

src/EntityFramework.Relational/RelationalOptionsExtension.cs
```diff
@@ -47,24 +50,54 @@ public virtual DbConnection Connection
...
                 {
-                    throw new ArgumentException(Strings.InvalidCommandTimeout);
+                    throw new InvalidOperationException(Strings.InvalidCommandTimeout);
```
ajcvickers: Why are these not ArgumentExceptions?

ajcvickers:
:shipit:

AndriySvyryd:
If max batch size is specified at the top-level wouldn't it make it impossible to specify different values for different providers? Since this is a perf tweaking setting it would be rare that the same value is optimal for different providers.

mikary:
If we are using multiple providers for the same DbContext, it would not be possible to specify different values in Configuration unless we add provider specific overrides. Is it a common scenario to use the same context with multiple providers?

AndriySvyryd:
I think it is common to have use the config file since it can be changed without recompiling and redeploying the app. It's fine as long as we maintain the ability to add a provider specific override.

rowanmiller:
Yeah the plan we briefly discussed was having top level settings but then later adding support for adding provider specific sub-sections.

rowanmiller:
BTW you currently can't swap providers from the config file, you have to change code to swap providers. Of course, you could have conditional code that swaps based on config settings... but in that case you can always include the different setting in the conditional code.

src/EntityFramework.Relational/RelationalOptionsExtension.cs
mikary: To avoid duplicating the range checking logic, these exceptions will be thrown for an invalid setting from Configuration as well as setting the wrong value on the instance.

PR #1435, opened 2015-01-19
[Tiny] Switch to MSBuild of VS 2015

smitpatel:
Builds correctly with VS 2015 (CTP edition) & Sql Server 2012 (SP2)

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -10,7 +10,7 @@ k-standard-goals
 
 #write-package-references target='initialize' if='!IsMono && !IsTeamCity'
   var programFilesX86='${Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86)}'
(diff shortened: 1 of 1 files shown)


smitpatel:
Should I also create issue to switch to Sql server 2014?

divega:
> Should I also create issue to switch to Sql server 2014?

Yes, if we don't have one already.

bricelam:
:shipit:

julielerman:
Is it correct that the build.cmd no longer works with VS2013? The output where it fails says its targeting MSBuild\14.0 and I'm getting a file not found error. If that's the case, you should update the wiki (https://github.com/aspnet/EntityFramework/wiki/Getting-and-Building-the-Code).

ErikEJ:
@julielerman why not do a PR? 😁

julielerman:
Please see https://github.com/aspnet/EntityFramework/issues/1570

bricelam:
It is intentional; we're using C# 6 features now. Yes, wikis always need updating. ;) That's why I avoid them.

PR #1438, opened 2015-01-20
tree printer rework - reusing build in ExpressionBuilder and storing the TreeNode in internal variable rather than returning it as a result of a Visit

maumar:
@anpete @divega

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionPrinter.cs
+++ b/src/EntityFramework.Core/Query/ExpressionPrinter.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Data.Entity.Query
 
             var visitor = new ExpressionPrintingVisitor();
             var nodePrinter = new TreeNodePrinter();
-            var node = visitor.VisitExpression(expression);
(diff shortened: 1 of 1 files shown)


anpete:
Hey, did you look at the pretty printer in relinq?

maumar:
Yes I looked at it, and its really nice, just need to do a bit of formatting to make it readable for my taste. However @divega  suggested it would be good to get this thing in regardless (after applying your CR suggestion)

anpete:
@divega Do we need two?

divega:
I talked to @maumar and the differences (or deficiencies) in formating have an impact on his productivity debugging query bugs. So yes, I think temporarily it is reasonable to have two expression printers. Long term we can look at overriding Relinq's and potentially contributing back improvements if they want them.

divega:
BTW @maumar will file a bug to track what we do long term.

anpete:
Sounds good. It would be ideal if we could override relinq's as part of this change.

maumar:
meged ad98141231c8871ed090b039dd657031854aa5c8

PR #1442, opened 2015-01-20
Query: Enables ThenInclude for full multi-level include support.

anpete:


anpete:
cc @maumar

src/EntityFramework.Core/Extensions/QueryableExtensions.cs
```diff
@@ -1298,9 +1299,10 @@ public static class QueryableExtensions
 
         public static readonly MethodInfo IncludeMethodInfo
             = typeof(QueryableExtensions)
-                .GetTypeInfo().GetDeclaredMethod("Include");
+                .GetTypeInfo().GetDeclaredMethods("Include")
+                .Single(mi => mi.GetParameters().Any(pi => pi.Name == "navigationPropertyPath"));
```
bricelam: Should these be matched by parameter type instead?

anpete: This seemed easier/simpler.

bricelam:
:shipit:

PR #1446, opened 2015-01-21
React to runtime renaming

ChengTian:
- Change default packages path
- Change KRE to DOTNET

@davidfowl @divega

### DIFF
--- a/tools/EntityFramework.targets
+++ b/tools/EntityFramework.targets
@@ -22,10 +22,10 @@
     <Task>
       <Code Type="Fragment" Language="cs">
         <![CDATA[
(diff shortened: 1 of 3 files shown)


tools/EntityFramework.targets
```diff
@@ -22,10 +22,10 @@
     <Task>
       <Code Type="Fragment" Language="cs">
         <![CDATA[
-        var packagesFolder = Environment.GetEnvironmentVariable("KRE_PACKAGES")
+        var packagesFolder = Environment.GetEnvironmentVariable("DOTNET_PACKAGES")
```
davidfowl: What a hack :stuck_out_tongue_closed_eyes:

PR #1464, opened 2015-01-24
Add a convention to replace shadow foreign keys with non-shadow ones.

AndriySvyryd:
Properties are be matched using the following patterns:
1. Navigation to principal name + corresponding principal key property name
2. Principal type name + corresponding principal key property name
3. Corresponding principal key property name
4. PK on the dependent entity for unique FKs

Add a way to specify non-primary key to InternalEntityBuilder.


src/EntityFramework.Core/Metadata/EntityTypeExtensions.cs
```diff
@@ -1,6 +1,7 @@
-﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
anpete: Why is this an extension method?

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
```diff
@@ -76,11 +75,40 @@ private InternalKeyBuilder Key(IReadOnlyList<Property> properties, Configuration
...
+            if (properties == null)
+            {
+                return null;
+            }
+
+            return _keyBuilders.GetOrAdd(
+                () => Metadata.TryGetKey(properties),
+                () => Metadata.AddKey(properties),
+                key => new InternalKeyBuilder(key, ModelBuilder),
+                null,
```
anpete: named arg?

src/EntityFramework.Core/Metadata/ModelConventions/ConventionsDispatcher.cs
```diff
@@ -12,16 +12,19 @@ public class ConventionsDispatcher
     {
         public ConventionsDispatcher()
         {
-            EntityTypeConventions = new List<IEntityTypeConvention>();
+            OnEntityTypeAddedConventions = new List<IEntityTypeConvention>();
+            OnForeignKeyAddedConventions = new List<IRelationshipConvention>();
         }
 
-        public virtual IList<IEntityTypeConvention> EntityTypeConventions { get; }
+        public virtual IList<IEntityTypeConvention> OnEntityTypeAddedConventions { get; }
```
anpete: Consider a name that doesn't read like a hook method.

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
```diff
@@ -706,49 +713,211 @@ public virtual InternalRelationshipBuilder Relationship([NotNull] ForeignKey for
...
+                // Create foreign key properties in shadow state if no properties are specified
+                var baseName = (string.IsNullOrEmpty(navigationToPrincipal) ? principalType.SimpleName : navigationToPrincipal);
+
+                var fkProperties = new List<Property>();
+                foreach (var keyProperty in principalKey.Properties)
+                {
+                    var index = -1;
+                    while (true)
+                    {
+                        var name = baseName + keyProperty.Name + (++index > 0 ? index.ToString() : "");
```
anpete: string.Empty?

anpete:
:shipit:

src/EntityFramework.Core/Metadata/EntityTypeExtensions.cs
AndriySvyryd: Will move to EntityType

src/EntityFramework.Core/Metadata/ModelConventions/ConventionsDispatcher.cs
AndriySvyryd: I'm open to suggestions. It represents a list of entity type conventions that run when a new entity type is added.
Note that in the future there might be also entity type conventions that run when something different happens.

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
AndriySvyryd: We've discussed this with Arthur and "" is as readable in most cases and uses less characters.

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
AndriySvyryd: I'll add it, but you won't like it. It's onNewKeyAdded:

anpete: Is it a delegate? :-)

src/EntityFramework.Core/Metadata/ModelConventions/ConventionsDispatcher.cs
anpete: EntityTypeAddedConventions?

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
AndriySvyryd: it is!

src/EntityFramework.Core/Metadata/ModelConventions/ConventionsDispatcher.cs
AndriySvyryd: ok

Issue #1426, opened 2015-01-18
Multiple Where method calls on one IQueryable, only the last one generates SQL statement.


RickyLin:
I assume it's a quite common situation that we need to call `Where` method multiple times on one IQueryable object. Code similar to this:

```c#
IQueryable<Customer> query = myDbContext.Customers;
if ( contactName != null )
    query = query.Where(c => c.ContactName == contactName);

if ( contactTitle != null )
    query = query.Where(c => c.ContactTitle == contactTitle);

Customer[] customers = query.ToArray();
```

I know I can convert the preceding code to this:

```c#
IQueryable<Customer> query = myDbContext.Customers.Where( c =>
    (contactName == null || c.ContactName == contactName)
    && (contactTitle == null || c.ContactTitle == contactTitle)
);
```

and EF 7 will generate SQL statement similar to this ( when contactName is not null but contactTitle is null):

```sql
WHERE (1 = 0 OR ContactName = @p0) AND ( 1 = 1 OR ContactTitle = NULL )
```

I don't thinks the generated SQL statement is ideal though. It would be better if we can support multiple Where method calls on one IQueryable and the preceding SQL statement would be just
`WHERE ContactName = @p0`

It seems simple to get EF 7 support this feature, I have tried it out in my fork and it works. The problem is that how I should write test code for this feature? There are dozens of `Where_*` test methods in test projects, should I duplicate all of them and modify the duplicated test methods to call 2 or more Where methods?

rowanmiller:
@RickyLin - Based on the SQL you included for the collapsed version of the query it looks like contactName was null (the `1 = 0` in the SQL indicates this). Can you confirm that this is not the reason that the Where method from the original logic is not being applied (since in your logic the `Where(c => c.ContactName == contactName)` won't get called if contactName is null).

RickyLin:
@rowanmiller - for the collapsed version of the query, the generated sql statement is correct.
Here is an working example, running on regular Console Application, .NET Framework 4.5.1, EF 7 beta 2.

I think the sql statement generated by (3) is correct. 
The problem is (2), I want to create a pull request for this issue so for (2) EF can generate the correct sql statement `WHERE WHERE [c].[ContactName] = @p0 AND [c].[ContactTitle] = @p1`. However, as my first post said, I have no idea how to create test code for my modification in test projects. The modification I did to address this issue is quite simple. #1439 is exactly the same problem as mine.

```c#
using System;
using System.Linq;
[...]

PR #1471, opened 2015-01-26
supporting multiple Where method calls on one IQueryable

RickyLin:
This is the pull request for #1426. Somehow I can't get the EntityFramework.SqlServer.FunctionalTests run successfully. all kinds of weird errors occurred during the test running which was not because of my test code. I have Sql Server 2014 Express with LocalDb installed. I can see lots of databases are created in my "Home" folder, other tests can run successfully except the SqlServer.FunctionalTests.

Can you guys run all the tests successfully?

The first error I got in SqlServer.FunctionalTests is:
```
Microsoft.Data.Entity.SqlServer.FunctionalTests.ComputedColumnTest.Can_use_computed_columns [FAIL]
      System.Data.SqlClient.SqlException : Database 'ComputedColumns' already exists. Choose a different database name.
```
The full output of my build process is [here (gist), text output of 3M large](https://gist.github.com/RickyLin/d6a921c7dfb841dc25d6)

### DIFF
--- a/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
@@ -227,13 +227,11 @@ namespace Microsoft.Data.Entity.Relational.Query
             {
                 var filteringVisitor = new FilteringExpressionTreeVisitor(this);
                 var predicate = filteringVisitor.VisitExpression(whereClause.Predicate);
-                if (selectExpression.Predicate == null)
+                if (predicate != null)
                 {
-                    selectExpression.Predicate = predicate;
-                }
-                else if (predicate != null)
-                {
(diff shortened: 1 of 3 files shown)


divega:
@RickyLin we changed the tests to use SQL Server 2014 LocalDB recently. If you had ran the tests before with VS 2014 it is likely you already have the same database files attached to the previous version of LocalDB. So you will have to go to the old instance and drop all the test databases before the tests can pass again.

anpete:
Here is the script to do that: https://github.com/aspnet/EntityFramework/blob/dev/tools/DropAllDatabases.sql

anpete:
Looks good to me.

RickyLin:
@divega - Thank you for your help. I can run all the tests (including the one I added) successfully now.

Eilon:
CLA confirmed.

anpete:
Merged, thanks!

Issue #1461, opened 2015-01-23
Investigate IModel caching behavior


anpete:
A couple of things to consider:

1) Should models be cached per provider?
2) Models are currently only cached by the DefaultModelSource, should we instead cache the result of IModelSource.GetModel?

PR #1474, opened 2015-01-26
Query: Part 1 of automatic query caching.

anpete:
Implements the basic caching architecture sans funcletization - It works because relinq partially evaluates captured variables during query parsing. We end up with a single cache entry per partially evaluated expression tree. Funcletization is TBD..

The cache itself is a simple ThreadSafeDictionaryCache, but is designed to be pluggable so that, for example, https://github.com/aspnet/Caching/tree/dev/src/Microsoft.Framework.Cache.Memory could be used.

Verified with manual perf testing.

### DIFF
--- a/src/EntityFramework.Core/Extensions/EntityServiceCollectionExtensions.cs
+++ b/src/EntityFramework.Core/Extensions/EntityServiceCollectionExtensions.cs
@@ -14,6 +14,7 @@ using Microsoft.Data.Entity.Utilities;
 using Microsoft.Framework.ConfigurationModel;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.OptionsModel;
+using Remotion.Linq.Parsing.Structure;
 
 // ReSharper disable once CheckNamespace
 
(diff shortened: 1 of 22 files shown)


src/EntityFramework.Core/Extensions/EntityServiceCollectionExtensions.cs
```diff
@@ -60,6 +61,8 @@ public static class EntityServiceCollectionExtensions
                 .AddSingleton<StoreGeneratedValuesFactory>()
                 .AddSingleton<ValueGeneratorSelector>()
                 .AddSingleton<StateEntryMetadataServices>()
+                .AddSingleton<IQueryParser, EntityQueryParser>()
+                .AddSingleton<ICompiledQueryCache, CompiledQueryCache>()
```
AndriySvyryd: Is the cache provider specific? If so a provider-specific singleton should be registered instead.

AndriySvyryd:
:sheep: :it:

src/EntityFramework.Core/Extensions/EntityServiceCollectionExtensions.cs
anpete: It should be. Spoke to @ajcvickers and he was investigating how to do this (we also need it for cached models, which we are also not currently doing) - Or do you already know how to do this? :-)

ajcvickers: @anpete @AndriySvyryd Andriy and I talked about the same plan that I mentioned to you. Pretty sure it will work.

anpete: Tracking with #1461. Will create a new bug for updating query caching.

Issue #1314, opened 2014-12-18
EntityType collection members should have a total order


AndriySvyryd:
To make the model less dependent on the order in which it's configured ForeignKeys, Indexes, Navigations and Keys properties should be totally ordered like Properties.

rowanmiller:
@mikary please discuss with @AndriySvyryd

PR #1477, opened 2015-01-26
Added absolute order to EntityType Keys, Foreign Keys, Indexes, and Navigations

mikary:
Adding total order to address #1314 and fixing tests broken by this change.
@AndriySvyryd @bricelam @lajones

### DIFF
--- a/src/EntityFramework.Core/Metadata/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/EntityType.cs
@@ -16,11 +16,24 @@ namespace Microsoft.Data.Entity.Metadata
     public class EntityType : MetadataBase, IEntityType
     {
         private readonly object _typeOrName;
-        private readonly LazyRef<List<ForeignKey>> _foreignKeys = new LazyRef<List<ForeignKey>>(() => new List<ForeignKey>());
-        private readonly LazyRef<List<Navigation>> _navigations = new LazyRef<List<Navigation>>(() => new List<Navigation>());
-        private readonly LazyRef<List<Index>> _indexes = new LazyRef<List<Index>>(() => new List<Index>());
-        private readonly List<Property> _properties = new List<Property>();
-        private readonly LazyRef<List<Key>> _keys = new LazyRef<List<Key>>(() => new List<Key>());
+        private readonly LazyRef<SortedDictionary<IReadOnlyList<Property>, ForeignKey>> _foreignKeys
+            = new LazyRef<SortedDictionary<IReadOnlyList<Property>, ForeignKey>>(()
+                => new SortedDictionary<IReadOnlyList<Property>, ForeignKey>(PropertyListComparer.Instance));
+
+        private readonly LazyRef<SortedDictionary<string, Navigation>> _navigations
+            = new LazyRef<SortedDictionary<string, Navigation>>(() =>
+            new SortedDictionary<string, Navigation>(StringComparer.Ordinal));
+
+        private readonly LazyRef<SortedDictionary<IReadOnlyList<Property>, Index>> _indexes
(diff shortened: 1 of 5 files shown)


src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -197,7 +197,7 @@ public virtual Key AddKey([NotNull] IReadOnlyList<Property> properties)
         {
             Check.NotEmpty(properties, "properties");
 
-            var key = TryGetKey(properties);
+            var key = TryGetPrimaryKey(properties);
```
AndriySvyryd: Was this an intentional change?

mikary: yes, TryGetKey searches first for a primary key and then through the ```_keys``` This preserves the logic but still lets us use the BinarySearch

AndriySvyryd: I see. Why not change the TryGet methods to use binary search instead?

mikary: The BinarySearch requires an instance, we use it in some of the Properties methods, but we have a performance note about the impact of creating a surrogate instance for the sake of searching.

We have some options to consider with respect to surrogates, binary search, and lazy if we want to start down that path on this PR.

AndriySvyryd: Yes, we should measure the impact of this at some point, but for now we should be consistent. Either use binary search in all TryGet methods or in none.

test/EntityFramework.Core.Tests/Metadata/EntityTypeTest.cs
```diff
@@ -179,7 +179,7 @@ public void Can_add_and_get_a_key()
 
             Assert.NotNull(key2);
             Assert.Same(key2, entityType.GetKey(idProperty));
-            Assert.Equal(new[] { key1, key2 }, entityType.Keys.ToArray());
+            Assert.Equal(new[] { key2, key1 }, entityType.Keys.ToArray());
```
lajones: Nit: this test looks like it belongs in Keys_are_ordered_by_property_count_then_property_names() below. Here would just add some non-overlapping keys and ensure you can retrieve them.

test/EntityFramework.Core.Tests/Metadata/EntityTypeTest.cs
```diff
@@ -269,6 +269,20 @@ public void Removing_a_key_throws_if_it_referenced_from_a_foreign_key_in_the_mod
...
+        {
+            var customerType = new EntityType(typeof(Customer), new Model());
+            var idProperty = customerType.GetOrAddProperty(Customer.IdProperty);
+            var nameProperty = customerType.GetOrAddProperty(Customer.NameProperty);
+
+            var k2 = customerType.GetOrAddKey(nameProperty);
+            var k3 = customerType.GetOrAddKey(new[] { idProperty, nameProperty });
+            var k1 = customerType.GetOrAddKey(idProperty);
+
+            Assert.True(new[] { k1, k2, k3 }.SequenceEqual(customerType.Keys));
```
lajones: Nit: maybe add another key k4 on (idProperty, abcProperty) just to prove that even in composite keys it orders them by name.

lajones:
:shipit:

AndriySvyryd:
:shipit:

mikary:
Updated for PR feedback, there are enough changes that it could use another look before I merge.

lajones:
:shipit:

AndriySvyryd:
nice :shipit:

PR #1482, opened 2015-01-27
Honor configuration in write-package-references target

bricelam:
To write package references for the Release configuration, use the following commands.
```cmd
SET Configuration=Release
build initialize
```

/cc @anpete


anpete:
:shipit:

Issue #1269, opened 2014-12-11
Tests: Using Task.Result causing deadlocks in SqlServer.FunctionalTests


bricelam:
I've increased the number of threads for now, but per xunit/xunit#88, we should come up with a solution that doesn't block in the constructors. (i.e. Invoke the task in the constructor, await the result when you need it.)

PR #1483, opened 2015-01-27
Stop using async database creation in test fixtures (Issue #1269)

ajcvickers:
This can cause deadlocks because the code runs from a constructor, which cannot be async.


AndriySvyryd:
:shipit:

Issue #1457, opened 2015-01-23
Query : "ExecuteReader requires an  open and available Connection" thrown when trying to execute a query after a failed query executed before


maumar:
Repro:

```
namespace Repro1457
{
    public class Level1
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public ICollection<Level2> OneToMany_Required { get; set; }
        public ICollection<Level2> OneToMany_Optional { get; set; }
    }

    public class Level2
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int Level1_Required_Id { get; set; }
        public int? Level1_Optional_Id { get; set; }

        public ICollection<Level3> OneToMany_Required { get; set; }
        public ICollection<Level3> OneToMany_Optional { get; set; }

        public Level1 OneToMany_Required_Inverse { get; set; }
        public Level1 OneToMany_Optional_Inverse { get; set; }
    }

    public class Level3
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int Level2_Required_Id { get; set; }
        public int? Level2_Optional_Id { get; set; }

        public Level2 OneToMany_Required_Inverse { get; set; }
        public Level2 OneToMany_Optional_Inverse { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Level1> LevelOne { get; set; }
        public DbSet<Level2> LevelTwo { get; set; }
        public DbSet<Level3> LevelThree { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Entity<Level1>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);

            modelBuilder.Entity<Level1>().HasMany(e => e.OneToMany_Required).WithOne(e => e.OneToMany_Required_Inverse).Required(true);
            modelBuilder.Entity<Level1>().HasMany(e => e.OneToMany_Optional).WithOne(e => e.OneToMany_Optional_Inverse).Required(false);

            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);

[...]

PR #1486, opened 2015-01-28
fix to issue #1457

maumar:
Problem was that in some cases we would try to close connection more times than we tried to open it. This would result in a corrupted state that would prevent opening the connection by subsequent queries.
This could happen in case of multi-level include. During that scenario we create multiple queries (data readers) and combine their results together in the end. Once data is retrieved there is a cleanup code that disposes of the participating enumerators.
However, if one of the queries throw or at some point during Include chain there is no more data (i.e. collection is empty) - it is possible that not all readers will be used, and therefore connection will be opened fewer times than assumed.
Cleanup code was trying to close connection once per expected reader, regardless of whether it was being used or not - which lead to this bug.

Fix is to only attempt to close connection if the associated reader has been used. Additionally I made QueryingEnumerable.Dispose() method idempotent.

### DIFF
--- a/src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
+++ b/src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
@@ -95,11 +95,15 @@ namespace Microsoft.Data.Entity.Relational.Query
             {
                 if (!_disposed)
                 {
-                    _disposed = true;
+                    if (_reader != null)
+                    {
+                        _enumerable._relationalQueryContext.Connection?.Close();
+                        _reader.Dispose();
+                    }
 
-                    _reader?.Dispose();
                     _command?.Dispose();
-                    _enumerable._relationalQueryContext.Connection?.Close();
+
+                    _disposed = true;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


maumar:
@anpete

maumar:
I didn't include tests as part of this checkin. I have a suite of tests ready but it's currently blocked on two more Include bugs (which seem related/similar):
https://github.com/aspnet/EntityFramework/issues/1460
https://github.com/aspnet/EntityFramework/issues/1452

I was planning to keep #1457 open till I am able to get those tests in. Or should I wait with the commit until I have them all in place?

src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
```diff
@@ -95,11 +95,15 @@ public void Dispose()
...
                 {
-                    _disposed = true;
+                    if (_reader != null)
+                    {
+                        _enumerable._relationalQueryContext.Connection?.Close();
+                        _reader.Dispose();
+                    }
 
-                    _reader?.Dispose();
                     _command?.Dispose();
```
anpete: _command can be moved into the conditional block. Actually, perhaps it doesn't even need to be field - can you try making it a local of MoveNext? Can't remember if it is OK to dispose the command while the reader is open.

anpete: One other thing, it is good practice to dispose resources in reverse acquisition order so dispose the reader before the connection close.

anpete:
:shipit: Fine to check this in before the other stuff.

divega:
@maumar please add a better description of what this is in the title before you push.

maumar:
sent another pull request for this

PR #1491, opened 2015-01-28
Fix #1463 - Include :: Multi-level include doesn't work with collections...

### LINKED ISSUE #1463: Include :: Multi-level include doesn't work with collections of type List<T> etc (incorrect overload of Include chosen by compiler)
Repro:

```

    public class Customer
    {
        public int Id { get; set; }
        public List<Order> Orders { get; set; }
    }

    public class Order
    {
        public int Id { get; set; }
        public List<OrderLine> OrderLines { get; set; }
    }

    public class OrderLine
    {
        public int Id { get; set; }
    }

    {
        public DbSet<Customer> Customers { get; set; }

        {
            modelBuilder.Entity<Customer>().HasMany(c => c.Orders);
            modelBuilder.Entity<Order>().HasMany(o => o.OrderLines);
        }

        {
            options.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "Repro1463", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }
    }

    class Program
    {
        {
            {
                ctx.Database.EnsureCreated();
                
                // won't compile
                ctx.Customers.Include(c => c.Orders).ThenInclude(o => o.OrderLines);
            }
        }
    }

```

anpete:
... of type List<T> etc (incorrect overload of Include chosen by compiler)

Implemented solution proposed by @divega in bug discussion.

### DIFF
--- a/src/EntityFramework.Core/Extensions/QueryableExtensions.cs
+++ b/src/EntityFramework.Core/Extensions/QueryableExtensions.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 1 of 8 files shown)


test/EntityFramework.Core.FunctionalTests/IncludeAsyncTestBase.cs
```diff
@@ -575,21 +577,6 @@ var order
         }
 
         [Fact]
-        public virtual async Task Include_multi_level_reference_then_include_collection_predicate()
```
maumar: It would be good to convert those tests into what would they look in the new API

src/EntityFramework.Core/Extensions/QueryableExtensions.cs
```diff
@@ -1310,75 +1311,59 @@ public static class QueryableExtensions
...
         }
 
         public static readonly MethodInfo ThenIncludeMethodInfo
             = typeof(QueryableExtensions)
                 .GetTypeInfo().GetDeclaredMethods("ThenInclude")
-                .Single(mi => mi.GetParameters().Any(pi => pi.Name == "navigationPropertyPath"));
+                .Single();
 
         public static IIncludableQueryable<TEntity, TProperty> ThenInclude<TEntity, TPreviousProperty, TProperty>(
-            [NotNull] this IIncludableQueryable<TEntity, TPreviousProperty> source,
```
divega: If I am interpreting the code correctly with the removal of this overload you can now only use `ThenInclude()` after an `Include()` that landed on a collection (i.e. the new overload). I am tweaking my mental model to something like this: "dot through" the navigation property path that you want inside the `Include()` method but collections will disrupt the path so you can use `ThenInclude()` as a way to escape them. 

E.g. you can now do this:

``` C#
orders.Include(c => c.OrderLines).ThenInclude(ol => o.Product.Category)
```

But you cannot longer do this:
``` C#
orders.Include(c => c.OrderLines).ThenInclude(ol => o.Product).ThenInclude(p=> p.Category)
```
Personally I like the approach better (I especially like that the superfluous versions of `ThenInclude()` won't be offered by Intellisense anymore) but I would like others to be able to chime in /cc @ajcvickers @rowanmiller

anpete: Correct. As I understood it, that was a constraint of your proposal, but I am happy to be wrong about that :-)

test/EntityFramework.Core.FunctionalTests/IncludeAsyncTestBase.cs
anpete: I had versions of those already :-)

src/EntityFramework.Core/Extensions/QueryableExtensions.cs
divega: The new overload worked alongside the existing ones when when I tried (i.e. they didn't seem to confuse the compiler), although the one that landed on a collection was definitively redundant.  Let's chat about the pros and cons.

anpete: I don't mind either way.

ajcvickers: I prefer to use either Include/ThenInclude consistently for any kind of navigation property, rather than using dotting sometimes and then dropping out to ThenInclude when I can't dot. But if others think that ThenInclude should only be used for dropping out, then I won't push back.

maumar: I prefer using Include/ThenInclude for references as well - seems more consistent and (from testing perspective) more composable.

anpete: Going to start without the other overload but will consider adding if we get feedback on it.

ajcvickers: @anpete I would really like to here what @rowanmiller thinks.

anpete: That's fine. I will check this in but am happy to send another PR if required.

PR #1492, opened 2015-01-28
Enforce name=xyz syntax for named connection strings

### LINKED ISSUE #1480: Enforce "name=xyz" syntax for named connection strings in configuration
Per #1391 we decided to stop using `"ConnectionStringKey" : "xyz"` as a way to specify that the actual connection string is defined in a configuration key `"xyz"`. We started supporting the form `"ConnectionString" : "name=xyz"` instead. But as a consequence of adopting parts of the EF6 implementation we also started supporting the form `"ConnectionString" : "xyz"` (i.e. given any "xyz" string that does not contain an equals sign we assume it is a configuration key that needs to be dereferenced to obtain the actual connection string). 

After discussing with @rowanmiller we decided that we should remove support for the latter since:

1. We want to keep the number of rules customers need to understand to the minimum 
2. We don't really need yet another way of configuring a connection string
3. In the future we may potentially support other kinds of connection strings which may not contain equal signs and which however are not configuration keys, e.g. hypothetically some non-relational store could use a URI as its connection string. When we do so it would be nice if we didn't need to special case such store (or relational, or any particular kind of store) and that we could instead have a single version of this code reading all sorts of connection strings from configuration across all kinds of databases.

smitpatel:
This enforces that named connection strings must start with name (removing trailing white spaces & case-insensitive) and must contain exactly 1 '=' sign.
fixes #1480

### DIFF
--- a/src/EntityFramework.Core/Infrastructure/DbContextOptionsParser.cs
+++ b/src/EntityFramework.Core/Infrastructure/DbContextOptionsParser.cs
@@ -67,23 +67,19 @@ namespace Microsoft.Data.Entity.Infrastructure
 
                 if (key.Equals(ConnectionStringKey, StringComparison.OrdinalIgnoreCase))
                 {
-                    var redirectionKey = string.Empty;
                     // Check if the value is redirection to other key
                     var firstequals = value.IndexOf('=');
-                    if (firstequals < 0)
-                    {
-                        redirectionKey = value;
-                    }
-                    else if ((value.IndexOf('=', firstequals + 1) < 0)
+                    if ((firstequals > 0)
+                        && (value.IndexOf('=', firstequals + 1) < 0)
                         && (value.Substring(0, firstequals).Trim().Equals(
-                          "name", StringComparison.OrdinalIgnoreCase)))
+                            "name", StringComparison.OrdinalIgnoreCase)))
                     {
-                        redirectionKey = value.Substring(firstequals + 1).Trim();
-                    }
-
-                    if (!string.IsNullOrEmpty(redirectionKey) && !configuration.TryGet(redirectionKey, out value))
-                    {
-                        throw new InvalidOperationException(Strings.ConnectionStringNotFound(redirectionKey));
+                        var redirectionKey = value.Substring(firstequals + 1).Trim();
+                        if (string.IsNullOrEmpty(redirectionKey)
+                            || !configuration.TryGet(redirectionKey, out value))
+                        {
+                            throw new InvalidOperationException(Strings.ConnectionStringNotFound(redirectionKey));
+                        }
                     }
                 }
 
--- a/test/EntityFramework.Core.Tests/DbContextOptionsParserTest.cs
+++ b/test/EntityFramework.Core.Tests/DbContextOptionsParserTest.cs
@@ -20,14 +20,14 @@ namespace Microsoft.Data.Entity.Tests
                 {
                     new MemoryConfigurationSource
                         {
-                            { "EntityFramework:" + typeof(MyContext).Name + ":ConnectionString", "Data Source=MyConnectionString" }
+                            { "EntityFramework:" + typeof(MyContext).Name + ":ConnectionString", "MyConnectionString" }
                         }
                 };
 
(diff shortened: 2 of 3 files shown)


test/EntityFramework.Core.Tests/DbContextOptionsParserTest.cs
```diff
@@ -20,14 +20,14 @@ public void Connection_string_is_found_using_context_name()
                 {
                     new MemoryConfigurationSource
                         {
-                            { "EntityFramework:" + typeof(MyContext).Name + ":ConnectionString", "Data Source=MyConnectionString" }
+                            { "EntityFramework:" + typeof(MyContext).Name + ":ConnectionString", "MyConnectionString" }
```
ajcvickers: I don't really understand why these tests were changed. Can you explain?

smitpatel: These changes are not necessary. Previously it was "MyConnectionString" only.
When changes according to EF6 were brought in, it started assuming value without '=' sign as named connection string (which we are removing here). So it was changed to "Data Source=MyConnectionString".

ajcvickers: Okay, so we still have tests for the case where there is no indirection? That is, one "=" but not preceded by "name", and multiple "="s, one of which may or may not be "name"?

smitpatel: I have added tests for all possible combinations.

src/EntityFramework.Core/Infrastructure/DbContextOptionsParser.cs
```diff
@@ -67,23 +67,19 @@ public class DbContextOptionsParser
...
-
-                    if (!string.IsNullOrEmpty(redirectionKey) && !configuration.TryGet(redirectionKey, out value))
+                            "name", StringComparison.OrdinalIgnoreCase)))
                     {
-                        throw new InvalidOperationException(Strings.ConnectionStringNotFound(redirectionKey));
+                        var redirectionKey = value.Substring(firstequals + 1).Trim();
+                        if (string.IsNullOrEmpty(redirectionKey)
+                            || !configuration.TryGet(redirectionKey, out value))
+                        {
+                            throw new InvalidOperationException(Strings.ConnectionStringNotFound(redirectionKey));
```
smitpatel: @ajcvickers - should we throw different exception is the connection string is like "name=". In such case we look for empty key in configuration.

test/EntityFramework.Core.Tests/DbContextOptionsParserTest.cs
```diff
@@ -136,43 +136,78 @@ public void Indirect_connection_string_is_read_removing_whitespaces()
...
             var rawOptions = new DbContextOptionsParser().ReadRawOptions(config, typeof(MyContext), new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase));
 
             Assert.Equal(1, rawOptions.Count);
-            Assert.Equal("Data Source=MyConnectionString", rawOptions["ConnectionString"]);
+            Assert.Equal("Name=Data:DefaultConnection:ConnectionString;Key=Value", rawOptions["ConnectionString"]);
         }
 
         [Fact]
-        public void Indirect_connection_string_must_have_only_one_equal_sign()
+        public void Indirect_connection_string_must_start_with_name_1()
```
smitpatel: @ajcvickers  This test and next 3 are verifying behavior when connection string doesn't start with "name=".
Suggestion for test names?

src/EntityFramework.Core/Infrastructure/DbContextOptionsParser.cs
ajcvickers: I don't think it is necessary to have a different exception.

test/EntityFramework.Core.Tests/DbContextOptionsParserTest.cs
ajcvickers: It seems like these are not "indirect" connection strings. These are three cases where having an equal sign in the string does not cause redirection to happen.  The first is a single key/value pair. The second is a string starting with "=" and the third is just a single "=". Does that help with naming?

ajcvickers:
:shipit:

PR #1494, opened 2015-01-28
fix to issue #1457 - "ExecuteReader requires an open and available Connection" thrown when trying to execute a query after a failed query executed before

maumar:
Problem was that in some cases we would try to close connection more times than we tried to open it. This would result in a corrupted state that would prevent opening the connection by subsequent queries.
This could happen in case of multi-level include. During that scenario we create multiple queries (data readers) and combine their results together in the end. Once data is retrieved there is a cleanup code that disposes of the participating enumerators.
However, if one of the queries throw or at some point during Include chain there is no more data (i.e. collection is empty) - it is possible that not all readers will be used, and therefore connection will be opened fewer times than assumed.
Cleanup code was trying to close connection once per expected reader, regardless of whether it was being used or not - which lead to this bug.

Fix is to only attempt to close connection if the associated reader has been used. Additionally I made QueryingEnumerable.Dispose() method idempotent.

### DIFF
--- a/src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
+++ b/src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
@@ -45,7 +45,6 @@ namespace Microsoft.Data.Entity.Relational.Query
         {
             private readonly AsyncQueryingEnumerable<T> _enumerable;
 
-            private DbCommand _command;
             private DbDataReader _reader;
 
             private T _current;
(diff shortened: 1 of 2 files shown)


maumar:
@anpete - same thing as before, incorporated CR feedback, build is passing

anpete:
:shipit:

maumar:
merged - c568126519d9434a2bc895962c0f3f89b866638f

PR #1496, opened 2015-01-29
Functional tests for graph manipulation and fixes to fixup bugs

ajcvickers:
These tests cover adding, removing, and changing relationships between entities. Dimensions are:
* Multiplicity (1:1, 1:*)
* Required/optional relationships
* PK <-> PK and PK <-> FK
* New (Added) entities or existing (Unchanged) entities
* Changing relationship at principal end, dependent end, or using FK
* PK principal keys or alternate principal keys

Note that these tests do no cover deleting entities, composite keys, or unidirectional navigations. Also, we don't yet handle conceptually null FKs.


anpete:
:shipit:

Issue #1089, opened 2014-11-18
Testing: Decisions from Design Meeting


anpete:
Implement decisions on fixtures etc. from design meeting

PR #1497, opened 2015-01-29
Make IModelSource provider-specific to allow providers to use different convention sets

### LINKED ISSUE #1461: Investigate IModel caching behavior

AndriySvyryd:
Add cross store tests that build models for different providers using the same service provider
Rename ConventionsDispatcher to ConventionDispatcher
Extract ConventionSet class from ConventionDispatcher

Resolves #1461
Part of #1089


src/EntityFramework.Core/Metadata/ModelConventions/ConventionSet.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Data.Entity.Metadata.ModelConventions
+{
+    public class ConventionSet
+    {
+        public ConventionSet()
+        {
+            EntityTypeAddedConventions = new List<IEntityTypeConvention>();
+            ForeignKeyAddedConventions = new List<IRelationshipConvention>();
```
ajcvickers: Initialize inline?

ajcvickers:
:shipit:

src/EntityFramework.Core/Metadata/ModelConventions/ConventionSet.cs
AndriySvyryd: Good idea, haven't tried those yet

PR #1501, opened 2015-01-29
Remove uses of CreateScratchAsync from test fixtures

ajcvickers:
I missed these cases in the previous change.


AndriySvyryd:
:shipit:

PR #1502, opened 2015-01-29
Only reference .NET Core packages on aspnetcore50

bricelam:
This prevents .NET Core assemblies from being added to net451 projects causing reference assembly conflicts.

@Eilon, @divega & @rowanmiller If possible, I think we should take this for beta3. Without it, using EF in net451 projects is crippled. The workaround without this is to remove the .NET Core references and binding redirects that get automatically added by NuGet,


divega:
:shipit:

Issue #1485, opened 2015-01-27
SqlServerGearsOfWarQueryTest includes initialization SQL in assert


ajcvickers:
/cc @maumar 

After changing the database initialization to be non-async (See #1483), the first SqlServerGearsOfWarQueryTest that runs now includes the database creation SQL in the SQL log, causing it to fail. The log should only be used for the test query, or the log should be cleared after initialization has happened.

Tests are currently disabled.

maumar:
fixed in 3ae3a838c019ab28e17336557ad17334accfaf40

PR #1504, opened 2015-01-30
fix to #1485 - enable GearsOfWar sql tests disabled by recent test change

maumar:
Issue was that GearsOfWar uses SqlLogger to validate the expected query, and with recent change, the logger could contain database creation SQL on top of the actual query that was tested. Fix is to clear log after the database has been created


maumar:
@ajcvickers

AndriySvyryd:
:shipit:

maumar:
merged - 3ae3a838c019ab28e17336557ad17334accfaf40

PR #1506, opened 2015-01-30
Remove async overloads of Add and related code

ajcvickers:
Based on API review decision. Async versions of Add were included because when value generation is being used it may occasionally be necessary to make a database call to get a new sequence. However, there are two problems with this:
* It is not possible to have async code paths for all places where Add may be called. For example, if a new entity is detected from a call to INotifyPropertyChanged.
* All higher-level code paths that may call Add need to have async versions as well. This spreads quicky--for example DetectChanged, AttachGraph.

There are two options for an application that wants to ensure that every possible database call is async:
* Use a value generation strategy that never queries the database on Add
* Make the database call to get sequence values explicitly and set key values directly before Adding new entities.


AndriySvyryd:
all that red :ship:

Issue #823, opened 2014-10-08
Commands: Use startup project


rowanmiller:
...for config files, data directory, relative paths, etc.

There are TODOs in the code for this, just logging an issue to track the work too

lukewaters:
Fixed in #1416

bricelam:
This isn't finished. We need to enable this functionality on the ASP.NET commands too.

bricelam:
We also need to consider whether the `-StartUpProject` parameter is needed on the NuGet commands.

lukewaters:
Fixed in #1416 #1503 and #1507

PR #1507, opened 2015-01-30
getting currentDirectory of startupProject and operating in that directo...

lukewaters:
...ry part of #823

### DIFF
--- a/src/EntityFramework.Commands/Program.cs
+++ b/src/EntityFramework.Commands/Program.cs
@@ -22,12 +22,14 @@ namespace Microsoft.Data.Entity.Commands
     {
         private readonly string _projectDir;
         private readonly string _rootNamespace;
+        private readonly ILibraryManager _libraryManager;
         private readonly MigrationTool _migrationTool;
         private CommandLineApplication _app;
 
-        public Program([NotNull] IApplicationEnvironment appEnv)
+        public Program([NotNull] IApplicationEnvironment appEnv, [NotNull] ILibraryManager libraryManager)
         {
             Check.NotNull(appEnv, "appEnv");
+            Check.NotNull(libraryManager, "libraryManager");
 
             _projectDir = appEnv.ApplicationBasePath;
             _rootNamespace = appEnv.ApplicationName;
@@ -36,6 +38,7 @@ namespace Microsoft.Data.Entity.Commands
             var assemblyName = new AssemblyName(appEnv.ApplicationName);
             var assembly = Assembly.Load(assemblyName);
             _migrationTool = new MigrationTool(loggerProvider, assembly);
+            _libraryManager = libraryManager;
         }
 
         public virtual int Main([NotNull] string[] args)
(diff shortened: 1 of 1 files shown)


src/EntityFramework.Commands/Program.cs
```diff
@@ -36,13 +38,20 @@ public Program([NotNull] IApplicationEnvironment appEnv)
...
+            {
+                _startupProjectDir = Path.GetDirectoryName(libraryPath);
+            }
         }
 
         public virtual int Main([NotNull] string[] args)
         {
             Check.NotNull(args, "args");
 
-            // TODO: Enable subcommands in help
```
lukewaters: fixed in 128aabd2c

src/EntityFramework.Commands/Program.cs
```diff
@@ -36,13 +38,20 @@ public Program([NotNull] IApplicationEnvironment appEnv)
             var assemblyName = new AssemblyName(appEnv.ApplicationName);
             var assembly = Assembly.Load(assemblyName);
             _migrationTool = new MigrationTool(loggerProvider, assembly);
+
+            var library = libraryManager.GetLibraryInformation(assembly.GetName().Name);
```
bricelam: `assembly` is the current project. We need to add a `-startUpProject` parameter to each command that executes user code and set the current directory based on that.

src/EntityFramework.Commands/Program.cs
bricelam: No, this comment was about enabling `ef help migration add` which currently throws.

lukewaters: ```ef migration add help``` works, but you have a point, I’ll put the comment back in

src/EntityFramework.Commands/Program.cs
```diff
@@ -263,6 +309,19 @@ public virtual int ShowHelp([CanBeNull] string command)
 
             return 0;
         }
+
+        public virtual string GetProjectPath(string projectName)
```
bricelam: `projectName` may be `null` if the user didn't specify one. `GetLibraryInformation` could throw.

src/EntityFramework.Commands/Program.cs
```diff
@@ -36,13 +38,14 @@ public Program([NotNull] IApplicationEnvironment appEnv)
...
             _migrationTool = new MigrationTool(loggerProvider, assembly);
+            _libraryManager = libraryManager;
         }
 
         public virtual int Main([NotNull] string[] args)
         {
             Check.NotNull(args, "args");
 
-            // TODO: Enable subcommands in help
+            var returnDirectory = Directory.GetCurrentDirectory();
```
bricelam: Not used

src/EntityFramework.Commands/Program.cs
bricelam: Please re-add

bricelam:
Can probably repeat yourself less if you introduce a pattern like this...
```C#
return Execute(
    startupProject,
    () =>
    {
        DoStuff();

        return 0;
    });
```
...and move the directory changing to the `Execute` method.

src/EntityFramework.Commands/Program.cs
```diff
@@ -43,6 +46,7 @@ public virtual int Main([NotNull] string[] args)
             Check.NotNull(args, "args");
 
             // TODO: Enable subcommands in help
+            var returnDirectory = Directory.GetCurrentDirectory();
```
bricelam: Not used

src/EntityFramework.Commands/Program.cs
```diff
@@ -191,15 +209,34 @@ public virtual int AddMigration([CanBeNull] string name, [CanBeNull] string cont
...
+            var startupProjectDir = GetProjectPath(startupProject);
+            if (startupProjectDir != null)
+            {
+                Directory.SetCurrentDirectory(startupProjectDir);
+            }
+
             _migrationTool.AddMigration(name, context, _rootNamespace, _projectDir).ToArray();
 
+            Directory.SetCurrentDirectory(returnDirectory);
```
bricelam: Should probably do this in a `finally` so it gets reset even when the `_migrationTool` throws.

lukewaters:
Ping!

bricelam:
:shipit:

PR #1508, opened 2015-01-30
Add dependencies between unit test projects

AndriySvyryd:
Move shared files


ajcvickers:
Could you summarize the advantages of this change?

AndriySvyryd:
Avoids the pain of adding shared files to projects
Avoids linked files that confuse VS

ajcvickers:
:shipit:

Issue #1493, opened 2015-01-28
RelationalDbContextOptions namespace?


dazinator:
I'm writing a relational provider, and came across the `RelationalDbContextOptions` class - which is in the EntityFramework.Relational assembly, but it's namespace is `Microsoft.Data.Entity.SqlServer.Extensions`

I am just wondering why it's in a `SqlServer` namespace - shouldn't it be provider agnostic?

divega:
Thanks for letting us know. It is possibly that we missed cleaning things up after refactoring. @mikary can you take a look?

PR #1518, opened 2015-02-02
Moving relational options class out of SQL specific namespace

mikary:
Correction for issue #1493 
Talked with @rowanmiller about the change, there are a few other "Extensions" classes in the relational assembly but no extensions namespace yet. We will likely revisit this in API review, but for now this change is consistent with the existing code.

### DIFF
--- a/src/EntityFramework.Relational/RelationalDbContextOptions.cs
+++ b/src/EntityFramework.Relational/RelationalDbContextOptions.cs
@@ -4,7 +4,7 @@
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 
-namespace Microsoft.Data.Entity.SqlServer.Extensions
+namespace Microsoft.Data.Entity.Relational
 {
     public class RelationalDbContextOptions
     {


src/EntityFramework.Relational/RelationalDbContextOptions.cs
```diff
@@ -4,7 +4,7 @@
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 
-namespace Microsoft.Data.Entity.SqlServer.Extensions
+namespace Microsoft.Data.Entity
```
ajcvickers: Why is this not in M.D.E.Relational?

mikary: That would make just as much sense. There are a few other "Extensions" related classes in the assembly in the M.D.E namespace (```RelationalBuilderExtensions``` ```RelationalDatabaseExtensions```, and ```RelationalMetadataExtensions```) and I was just matching what was there.

ajcvickers: We typically put extension methods into a namespace which makes them show up in Intellisense when people have common using directives in their class. This is often M.D.E. But in this case the class does not have extension methods.

ajcvickers:
:shipit:

PR #1521, opened 2015-02-02
API docs for DbContext

rowanmiller:
Not commenting collection overloads of Add/Update/Attach/Remove just yet as the APIs are about to change

### DIFF
--- a/src/EntityFramework.Core/DbContext.cs
+++ b/src/EntityFramework.Core/DbContext.cs
@@ -20,6 +20,26 @@ using Microsoft.Framework.OptionsModel;
 
 namespace Microsoft.Data.Entity
 {
+    /// <summary>
+    /// A DbContext instance represents a session with the data store and can be used to query and save
+    /// instances of your entities. DbContext is a combination of the Unit Of Work and Repository patterns.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Typically you create a class that derives from DbContext and contains <see cref="DbSet{TEntity}" /> 
+    /// properties for each entity in the model. If the <see cref="DbSet{TEntity}" /> properties have a public setter, 
+    /// they are automatically initialized when the instance of the derived context is created.
+    /// </para>
+    /// <para>
+    /// Override the <see cref="OnConfiguring(DbContextOptions)"/> method to configure the data store (and other options) to be 
+    /// used for the context.
+    /// </para>
+    /// <para>
+    /// The model is discovered by running a set of conventions over the entity classes found in the <see cref="DbSet{TEntity}" />
+    /// properties on the derived context. To further configure the model that is discovered by convention, you can 
+    /// override the <see cref="OnModelCreating(ModelBuilder)"/> method.
+    /// </para>
+    /// </remarks>
     public class DbContext : IDisposable, IAccessor<IServiceProvider>
     {
         private static readonly ThreadSafeDictionaryCache<Type, Type> _optionsTypes = new ThreadSafeDictionaryCache<Type, Type>();
@@ -30,6 +50,10 @@ namespace Microsoft.Data.Entity
 
         private bool _initializing;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DbContext" /> class. The <see cref="OnConfiguring(DbContextOptions)"/> 
+        /// method will be called to configure the data store (and other options) to be used for this context.
+        /// </summary>
         protected DbContext()
         {
             var serviceProvider = DbContextActivator.ServiceProvider;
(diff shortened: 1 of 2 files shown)


bricelam:
@rowanmiller I'll need to see your CLA first :wink:

bricelam:
I didn't review the actual content (I'll let someone with intimate DbContext knowledge do that), but the XML looks good.

rowanmiller:
@ajcvickers would like your input on the actual docs when you have a chance

src/EntityFramework.Core/DbContext.cs
```diff
@@ -20,6 +20,26 @@
 
 namespace Microsoft.Data.Entity
 {
+    /// <summary>
+    /// A DbContext instance represents a session with the data store and can be used to query and save
+    /// instances your entities. DbContext is a combination of the Unit Of Work and Repository patterns.
```
ajcvickers: "of your entities"?

src/EntityFramework.Core/DbContext.cs
```diff
@@ -171,14 +236,38 @@ private void InitializeSets(IServiceProvider serviceProvider, DbContextOptions o
...
 
+        /// <summary>
+        /// Override this method to configure the data store (and other options) to be used for this context.
+        /// This method is called for each instance of the context that is created.
+        /// </summary>
+        /// <param name="options">
+        /// The options for this context. Data stores (and other extensions) typically define extension methods 
+        /// on this object that allow you to configure the context.
+        /// </param>
         protected internal virtual void OnConfiguring(DbContextOptions options)
```
ajcvickers: I think it would be worth documenting that the options passed into this method is cloned from the DbContextOptions that was passed to the constructor or obtained from the service provider. This allows different options to be used for each session even when only one DbContextOptions is registered in the service provider.

src/EntityFramework.Core/DbContext.cs
```diff
@@ -171,14 +236,38 @@ private void InitializeSets(IServiceProvider serviceProvider, DbContextOptions o
...
         }
 
+        /// <summary>
+        /// Saves all changes made in this context to the underlying data store.
+        /// </summary>
+        /// <returns>
+        /// The number of state entries written to the underlying data store.
+        /// </returns>
         [DebuggerStepThrough]
         public virtual int SaveChanges()
```
ajcvickers: Document that this method automatically calls DetectChanges unless AutoDetectChanges is turned off.

src/EntityFramework.Core/DbContext.cs
```diff
@@ -243,6 +347,14 @@ public virtual void Dispose()
...
 
+        /// <summary>
+        /// Gets an <see cref="EntityEntry{TEntity}" /> for the given entity providing access to
+        /// information the context is tracking for the given the entity and the ability 
+        /// to perform actions on the entity.
+        /// </summary>
+        /// <typeparam name="TEntity"> The type of the entity. </typeparam>
+        /// <param name="entity"> The entity to get the entry for. </param>
+        /// <returns> The entry for the given entity. </returns>
         public virtual EntityEntry<TEntity> Entry<TEntity>([NotNull] TEntity entity) where TEntity : class
```
ajcvickers: Ditto document calls DetectChanges.

src/EntityFramework.Core/DbContext.cs
```diff
@@ -296,6 +455,17 @@ private EntityEntry EntryWithoutDetectChanges([NotNull] object entity)
...
+        /// be removed from the data store when <see cref="SaveChanges"/> is called.
+        /// </summary>
+        /// <typeparam name="TEntity"> The type of the entity. </typeparam>
+        /// <param name="entity"> The entity to remove. </param>
+        /// <returns>
+        /// The <see cref="EntityEntry{TEntity}" /> for the entity. This entry provides access to
+        /// information the context is tracking for the the entity and the ability to perform 
+        /// actions on the entity.
+        /// </returns>
         public virtual EntityEntry<TEntity> Remove<TEntity>([NotNull] TEntity entity) where TEntity : class
```
ajcvickers: Not sure if it is worth documenting here, but if the entity is currently Added, then it will be Detached following a call to Remove.

src/EntityFramework.Core/DbContext.cs
```diff
@@ -477,15 +694,29 @@ private List<EntityEntry> GetOrCreateEntries(IEnumerable<object> entities)
...
+        /// </summary>
         public virtual IModel Model
             => _contextServices.Value.ScopedServiceProvider.GetRequiredServiceChecked<DbContextService<IModel>>().Service;
 
+        /// <summary>
+        /// Creates a set to perform operations for a given entity type in the model.
+        /// </summary>
+        /// <typeparam name="TEntity"> The type of entity for which a set should be returned. </typeparam>
+        /// <returns> A set for the given entity type. </returns>
         public virtual DbSet<TEntity> Set<TEntity>()
```
ajcvickers: Might be worth explicitly stating that this is the entry point for LINQ queries.

ajcvickers:
:shipit:

src/EntityFramework.Core/DbContext.cs
rowanmiller: Adjusted summary to look like this:

```
        /// <summary>
        /// <para>
        /// Override this method to configure the data store (and other options) to be used for this context.
        /// This method is called for each instance of the context that is created.
        /// </para>
        /// <para>
        /// If you passed an instance of <see cref="DbContextOptions"/> to the constructor of the context (or
        /// provided an <see cref="IServiceProvider"/> with <see cref="DbContextOptions"/> registered) then
        /// it is cloned before being passed to this method. This allows the options to be altered without
        /// affecting other context instances that are constructed with the same <see cref="DbContextOptions"/>
        /// instance.
        /// </para>
        /// </summary>
```

rowanmiller: Actually, moving that to remarks instead... you don't really need to know that to use the API

src/EntityFramework.Core/DbContext.cs
rowanmiller:
```
        /// <remarks>
        /// This method will automatically call <see cref="ChangeTracker.DetectChanges"/> to discover any changes
        /// to entity instances before saving to the underlying data store. This can be disabled via
        /// <see cref="ChangeTracker.AutoDetectChangesEnabled"/>.
        /// </remarks>
```

src/EntityFramework.Core/DbContext.cs
rowanmiller:
```
        /// <remarks>
        /// If the entity is already tracked in the <see cref="EntityState.Added"/> state then the context will
        /// stop tracking the entity (rather than marking it as <see cref="EntityState.Deleted"/>) since the 
        /// entity was previously added to the context and does not exist in the data store.
        /// </remarks>
```

src/EntityFramework.Core/DbContext.cs
rowanmiller:
```
        /// <summary>
        /// Creates a set to perform operations for a given entity type in the model. LINQ queries against
        /// <see cref="DbSet{TEntity}"/> will be translated into queries against the data store.
        /// </summary>
```

PR #1522, opened 2015-02-02
Make DbSet abstract and add IAccessor<IServiceProvider> to it and DbContext

ajcvickers:
Based on API review notes here: https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes:-January-16,-2015#api-review-1-dbcontext-dbset-database

Note that we decided to make DbSet an abstract base class and to explicitly implement its interfaces. The explicitly implemented methods cannot be abstract, so they throw NotImplementedException. After discussion with @divega and @rowanmiller we decided to make all the methods virtual and throw.


AndriySvyryd:
:shipit:

ajcvickers:
Merged

Issue #1510, opened 2015-01-30
Change DI contracts to stateless classes with parameterless constructor


AndriySvyryd:
The service contracts should be represented by field-less classes without non-default constructors. Default implementations should be separated to base classes
Pros:
•	We can make some API changes without breaking implementation
•	We can add helper methods inline instead of using extensions
•	They can be mocked easily
•	Implementations can have different fields and constructor parameters
Cons:
•	Cannot implement more than one service in a single class

Proposed changes summary:
IValueGeneratorFactory -> 
SimpleValueGeneratorFactory<TValueGenerator> -> ValueGeneratorFactory<TValueGenerator>
IValueGenerator -> ValueGenerator
ValueGeneratorSelector -> ValueGeneratorSelector, ValueGeneratorSelectorBase
ValueGeneratorCache -> ValueGeneratorCache, ValueGeneratorCacheBase
IValueReader -> ValueReader
DbSetFinder -> DbSetFinder, CachingDbSetFinder
DbSetSource -> DbSetSource, CachingDbSetSource
DbSetInitializer -> DbSetInitializer, DbSetInitializerBase
EntityKeyFactorySource -> EntityKeyFactorySource, EntityKeyFactorySourceBase
ClrPropertyGetterSource -> IClrPropertyGetterSource
ClrPropertySetterSource -> IClrPropertySetterSource
ClrCollectionAccessorSource -> IClrCollectionAccessorSource, ClrAccessorSource
ClrAccessorSource<IClrPropertyGetter> -> ClrAccessorSource
EntityMaterializerSource -> EntityMaterializerSource, EntityMaterializerSourceBase
MemberMapper -> MemberMapper, MemberMapperBase
ICompiledQueryCache -> CompiledQueryCache
CompiledQueryCache -> CompiledQueryCacheBase
ForeignKeyValuePropagator -> ForeignKeyValuePropagatorBase
DataStoreSelector -> DataStoreSelector, DataStoreSelectorBase
StateEntryFactory -> StateEntryFactory, StateEntryFactoryBase
NavigationFixer -> NavigationFixer, NavigationFixerBase
ChangeDetector -> ChangeDetector, ChangeDetectorBase
StateEntryNotifier -> StateEntryNotifier, StateEntryNotifierBase
StateEntrySubscriber -> StateEntrySubscriberBase
IDbContextOptions -> IDbContextOptions, DbContextOptions
DbContextOptions -> DbContextOptionsBase
[...]

PR #1523, opened 2015-02-02
Changing Metadata services to abstract classes without non-default constructors

### LINKED ISSUE #1181: ModelBuilder is not dynamically selected by provider in the correct way
ModelBuilder is added as a common service in AddEntityFramework but an attempt is made to specialize in AddAzureTableStorage.

If the ModelBuilder needs to be specialized, then it needs to be dynamically selected as part of the provider services such that multiple providers can register different model differs without one hiding another.

Note that the cross-store functional tests are not currently catching this issue.

AndriySvyryd:
Don't register CompositeEntityKeyFactory and ValueGeneratorSelector in DI
Make ModelBuilderFactory a provider-only service
Refactor EntityServiceCollectionExtensionsTest

Resolves #1181
Part of #1510


ajcvickers:
Why removing CompositeEntityKeyFactory  and ValueGeneratorSelector from DI?

AndriySvyryd:
They aren't needed in DI

src/EntityFramework.Core/Storage/DataStoreServices.cs
```diff
@@ -51,6 +51,11 @@ public abstract class DataStoreServices
             get { return p => new DbContextService<DataStoreConnection>(() => GetStoreServices(p).Connection); }
         }
 
+        public static Func<IServiceProvider, DbContextService<ModelBuilderFactory>> ModelBuilderFactoryFactory
```
AndriySvyryd: This might need a better name

ajcvickers:
So, your're saying that nobody could ever possibly want to change the implementation of CompositeEntityKeyFactory. Even though it interacts with value readers which can have quirks depending on the provider implementation?

AndriySvyryd:
No, I'm saying that when you want to change CompositeEntityKeyFactory you need to implement your own EntityKeyFactorySource, this is consistent with how you would change the SimpleEntityKeyFactory.

ajcvickers:
I can see that that is consistent. But the SimpleEntityKeyFactory case is due to limitations in what we can resolve in DI due to needing an instance of a generic type for which we don't know the generic type parameter in advance. It seems to me that CompositeEnittyKeyFactory is a service and can be resolved from DI and hence it should be resolved from DI.

ajcvickers:
:shipit:

Issue #1528, opened 2015-02-03
Include :: Exception thrown for multi-level include dotting over one-one navigation into one to many deeper inside the include chain


maumar:
```
namespace Repro1528
{
    public class Level1
    {
        public int Id1 { get; set; }
        public string Name { get; set; }

        public ICollection<Level2> OneToMany_Required { get; set; }
        public ICollection<Level2> OneToMany_Optional { get; set; }
    }

    public class Level2
    {
        public int Id2 { get; set; }
        public string Name { get; set; }

        public int Level1_Required_Id { get; set; }
        public int? Level1_Optional_Id { get; set; }

        public ICollection<Level3> OneToMany_Required { get; set; }
        public ICollection<Level3> OneToMany_Optional { get; set; }

        public Level1 OneToMany_Required_Inverse { get; set; }
        public Level1 OneToMany_Optional_Inverse { get; set; }
    }

    public class Level3
    {
        public int Id3 { get; set; }
        public string Name { get; set; }

        public int Level2_Required_Id { get; set; }
        public int? Level2_Optional_Id { get; set; }

        public Level2 OneToMany_Required_Inverse { get; set; }
        public Level2 OneToMany_Optional_Inverse { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Level1> LevelOne { get; set; }
        public DbSet<Level2> LevelTwo { get; set; }
        public DbSet<Level3> LevelThree { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Entity<Level1>().Key(e => e.Id1);
            modelBuilder.Entity<Level2>().Key(e => e.Id2);
            modelBuilder.Entity<Level3>().Key(e => e.Id3);

            modelBuilder.Entity<Level1>().Property(e => e.Id1).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id2).GenerateValueOnAdd(false);

            modelBuilder.Entity<Level1>().HasMany(e => e.OneToMany_Required).WithOne(e => e.OneToMany_Required_Inverse).Required(true);
            modelBuilder.Entity<Level1>().HasMany(e => e.OneToMany_Optional).WithOne(e => e.OneToMany_Optional_Inverse).Required(false);

            modelBuilder.Entity<Level2>().Property(e => e.Id2).GenerateValueOnAdd(false);
[...]

Issue #1460, opened 2015-01-23
Include :: "Sequence contains more than one matching element" thrown in some cases of multi level include


maumar:
Repro:

```
    public class Level1
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public ICollection<Level2> OneToMany_Optional { get; set; }
    }

    public class Level2
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int Level1_Required_Id { get; set; }
        public int? Level1_Optional_Id { get; set; }

        public ICollection<Level3> OneToMany_Optional { get; set; }
        public Level1 OneToMany_Optional_Inverse { get; set; }
    }

    public class Level3
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int Level2_Required_Id { get; set; }
        public int? Level2_Optional_Id { get; set; }

        public ICollection<Level4> OneToMany_Optional { get; set; }

        public Level2 OneToMany_Optional_Inverse { get; set; }
    }

    public class Level4
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int Level3_Required_Id { get; set; }
        public int? Level3_Optional_Id { get; set; }

        public Level3 OneToMany_Optional_Inverse { get; set; }
    }


    public class MyContext : DbContext
    {
        public DbSet<Level1> LevelOne { get; set; }
        public DbSet<Level2> LevelTwo { get; set; }
        public DbSet<Level3> LevelThree { get; set; }
        public DbSet<Level4> LevelFour { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Entity<Level1>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);

            modelBuilder.Entity<Level1>().HasMany(e => e.OneToMany_Optional).WithOne(e => e.OneToMany_Optional_Inverse).Required(false);

            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);

            modelBuilder.Entity<Level2>().HasMany(e => e.OneToMany_Optional).WithOne(e => e.OneToMany_Optional_Inverse).Required(false);

[...]

Issue #1452, opened 2015-01-21
Multiple multi-level collection/reference Include can fail


ajcvickers:
Repro code:
```
using (var context = CreateContext())
{
    var root = context.Roots
        .Include(e => e.RequiredChildren).ThenInclude(e => e.Children)
        .Include(e => e.OptionalChildren).ThenInclude(e => e.Children)
        .Include(e => e.RequiredSingle).ThenInclude(e => e.Single)
        .Include(e => e.RequiredNonPkSingle).ThenInclude(e => e.Single)
        .Include(e => e.OptionalSingle).ThenInclude(e => e.Single)
        .Single();
}
```

Generated SQL:
```
SELECT TOP(@p0) [e].[AlternateId], [e].[Id], [r].[Id], [r0].[Id], [r1].[Id], [r1].[RootId], [r2].[BackId], [r2].[Id], [o].[Id], [o].[RootId], [o0].[BackId], [o0].[Id]
FROM [Root] AS [e]
LEFT JOIN [RequiredSingle1] AS [r] ON [r].[Id] = [e].[Id]
LEFT JOIN [RequiredSingle2] AS [r0] ON [r0].[Id] = [r].[Id]
LEFT JOIN [RequiredNonPkSingle1] AS [r1] ON [r1].[RootId] = [e].[Id]
LEFT JOIN [RequiredNonPkSingle2] AS [r2] ON [r2].[BackId] = [r1].[Id]
LEFT JOIN [OptionalSingle1] AS [o] ON [o].[RootId] = [e].[Id]
LEFT JOIN [OptionalSingle2] AS [o0] ON [o0].[BackId] = [o].[Id]
ORDER BY [e].[Id]

SELECT [o].[Id], [o].[ParentId]
FROM [Optional1] AS [o]
INNER JOIN (
    SELECT DISTINCT TOP(@p0) [e].[Id]
    FROM [Root] AS [e]
) AS [e] ON [o].[ParentId] = [e].[Id]
ORDER BY [e].[Id], [o].[Id] AS [Id0]
	SqlServerGraphUpdatesTest.cs(35,0)
```

Entities and context, etc.:
```
protected class Root
{
    public int Id { get; set; }
    public Guid AlternateId { get; set; }

    public ICollection<Required1> RequiredChildren { get; set; } = new List<Required1>();
    public ICollection<Optional1> OptionalChildren { get; set; } = new List<Optional1>();
[...]

Issue #1362, opened 2015-01-07
Include :: multi level include may produce invalid SQL


maumar:
Consider the following model:

```
    public class Customer
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public Order Order { get; set; }
    }

    public class Order
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public List<OrderDetail> OrderDetails { get; set; }
    }

    public class OrderDetail
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public Order Order { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Customer> Customers { get; set; }
        public DbSet<Order> Orders { get; set; }
        public DbSet<OrderDetail> OrderDetails { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Customer>().OneToOne(c => c.Order);
            modelBuilder.Entity<Order>().OneToMany(o => o.OrderDetails);
            modelBuilder.Entity<Customer>().Property(o => o.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Order>().Property(o => o.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<OrderDetail>().Property(o => o.Id).GenerateValueOnAdd(false);
        }

        protected override void OnConfiguring(DbContextOptions options)
        {
            options.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "MultiLevelInclude2", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();


                var c1 = new Customer
                {
                    Id = 1,
                    Name = "Customer1"
                };

                var o1 = new Order
                {
                    Id = 1,
                    Name = "Order1"
                };

                var od10 = new OrderDetail
                {
                    Id = 10,
                    Name = "OrderDetail 1 0"
                };

                ctx.Customers.Add(c1);
                ctx.Orders.Add(o1);
                c1.Order = o1;

                ctx.SaveChanges();
            }

            using (var ctx = new MyContext())
            {
                var result = ctx.Customers.Include(c => c.Order.OrderDetails).ToList();
                Console.WriteLine(result);
            }
        }
    }
```

[...]

PR #1532, opened 2015-02-03
fix to include bugs - 1452 and 1460

maumar:
#1452 - Multiple multi-level collection/reference Include can fail
#1460 - "Sequence contains more than one matching element" thrown in some cases of multi level include

#1452 - problem appears in include queries in which Id columns of Included entities have same name. When this happens, we uniquefy them (see #1362). However, for more complex multi-level include, we use those keys in sort clause in one stage of the include but then we need to uniquefy it in the stage that comes after. This would result in SQL that tried to apply alias in the sort clause which is invalid. Fix is to not generate the alias for the order by during sql generation. Also another issue around this area was that we were passing original column name rather than aliased one to the order clause in the SelectExpression.

#1460 - problem appears in multi-level include queries that include same entity type multiple times in the chain. We uniquefy the names, however when constructing join conditions we only have column original name and table it comes from. This would lead to errors, since we assumed there would always be only one column in the projection list, that would match the criteria for the join condition. Fix is to use the last column from the list of candidates - as we always stick the column from the most recently joined table at the end of the projection list.

Some tests are currently blocked by another issue - #1528

### DIFF
--- a/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
+++ b/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
@@ -96,7 +96,16 @@ namespace Microsoft.Data.Entity.Relational.Query.Sql
 
                 VisitJoin(selectExpression.OrderBy, t =>
                     {
-                        VisitExpression(t.Expression);
+                        var columnExpression = t.Expression as ColumnExpression;
+                        if (columnExpression != null)
+                        {
+                            // strip column expression from alias - it should not be generated for ORDER BY clause
+                            VisitExpression(new ColumnExpression(columnExpression.Name, columnExpression.Property, columnExpression.Table));
+                        }
+                        else
+                        {
+                            VisitExpression(t.Expression);
+                        }
 
                         if (t.OrderingDirection == OrderingDirection.Desc)
                         {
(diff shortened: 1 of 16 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -96,7 +96,16 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
 
                 VisitJoin(selectExpression.OrderBy, t =>
                     {
-                        VisitExpression(t.Expression);
+                        var columnExpression = t.Expression as ColumnExpression;
+                        if (columnExpression != null)
+                        {
+                            // strip column expression from alias - it should not be generated for ORDER BY clause
+                            VisitExpression(new ColumnExpression(columnExpression.Name, columnExpression.Property, columnExpression.Table));
```
maumar: this is not very clear (i.e. its not obvious that we are stripping alias here) - I added the comment, alternatively I could introduce the state into the visitor that would add/ignore alias depending on the context in the ColumnExpression visit method itself

anpete: Just inline the relevant code from VisitColumnExpression here..

anpete:
:shipit:

PR #1541, opened 2015-02-04
Implement PropertyEntry.OriginalValue

### LINKED ISSUE #1396: Implement PropertyEntry.OriginalValue
Issue #1200 is about implementing the methods on EntityEntry that get all current values, original values, etc. But on PropertyEntry there is already a CurrentValue property which is much less hassle to use than dropping down to the low-level StateEntry. It should be easy to implement a similar OriginalValue property which would then make things like concurrency resolution easier.

smitpatel:
fixes #1396


ajcvickers:
:shipit:

Issue #1517, opened 2015-02-02
Use AddRange, RemoveRange, etc. for multiple entity overloads


ajcvickers:
In the API review meeting we decided on the following overloads:

```
public class DbContext
{
    public virtual EntityEntry Add(object entity)
    public virtual void Add(params object[] entities)
    public virtual EntityEntry<TEntity> Add<TEntity>(TEntity entity) where TEntity : class
    public virtual void Add<TEntity>(IEnumerable<TEntity> entities)
    public virtual EntityEntry<TEntity> Add<TEntity>(TEntity entity) where TEntity : class
}
```
But as we know (because we keep finding out but not remembering. :smile:) overload resolution doesn't work great when choosing between object and IEnumerable. For example:

```
using (var context = new EarlyLearningCenter())
{
    // These call the correct overloads:
    context.Add(new Product()); // Calls correct overload
    context.Add((object)new Product()); //Calls correct overload
    context.Add(new Product(), new Product()); //Calls correct overload

    // This calls the single object method with TEntity == List<Product>
    context.Add(new List<Product> { new Product(), new Product() });

    // This calls the single object method with TEntity == Product[]
    context.Add(new[] { new Product(), new Product() });

}
```

I suggest that we learn from the BCL and from the old stack and stop trying to be too clever with overloads: we should use AddRange for the methods that take multiple entities.

PR #1544, opened 2015-02-04
Use AddRange, RemoveRange, etc. for multiple entity overloads (Issue #1517)

ajcvickers:
Also add IEnumerable overload and change return type to void.


bricelam:
:shipit:

Issue #1393, opened 2015-01-13
Include :: data corruption for multi level include with optional and required relationships


maumar:
```
    public class CogTag
    {
        public Guid Id { get; set; }
        public string Note { get; set; }

        public string GearNickName { get; set; }
        public int? GearSquadId { get; set; }
        public virtual Gear Gear { get; set; }
    }

    public class Gear
    {
        public Gear()
        {
            Reports = new List<Gear>();
        }

        // composite key
        public string Id { get; set; }
        public int SquadId { get; set; }

        public string FullName { get; set; }

        public virtual CogTag Tag { get; set; }
        public virtual Squad Squad { get; set; }

        public string LeaderNickname { get; set; }
        public int LeaderSquadId { get; set; }

        // 1 - many self reference
        public virtual ICollection<Gear> Reports { get; set; }
    }

    public class Squad
    {
        public Squad()
        {
            Members = new List<Gear>();
        }

        // non-auto generated key
        public int Id { get; set; }
        public string Name { get; set; }

        public virtual ICollection<Gear> Members { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Gear> Gears { get; set; }
        public DbSet<Squad> Squads { get; set; }
        public DbSet<CogTag> Tags { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Gear>(b =>
            {
                b.Key(g => new { g.Id, g.SquadId });
                b.OneToMany(g => g.Reports).ForeignKey(g => new { g.LeaderNickname, g.LeaderSquadId });
                b.OneToOne(g => g.Tag, t => t.Gear).ForeignKey<CogTag>(t => new { t.GearNickName, t.GearSquadId });
            });

            modelBuilder.Entity<CogTag>(b =>
            {
                b.Key(t => t.Id);
                b.Property(t => t.Id).GenerateValueOnAdd();
            });

            modelBuilder.Entity<Squad>(b =>
            {
                b.Key(s => s.Id);
                b.OneToMany(s => s.Members, g => g.Squad).ForeignKey(g => g.SquadId);
                b.Property(t => t.Id).GenerateValueOnAdd();
            });
        }

        protected override void OnConfiguring(DbContextOptions options)
        {
[...]

PR #1547, opened 2015-02-04
Fix to #1393 - data corruption for multi level include with optional and required relationships

maumar:
Problem was that for include queries we would produce INNER JOIN for required relationships and OUTER JOIN for optional relationships. However for multi-level navigations this could cause data corruption. If one navigation was optional, all consequent joins should be OUTER JOINs as well, otherwise incorrect number of rows could be returned. 

Fix is to only produce INNER JOINs if there was no OUTER JOIN produced before.


maumar:
@anpete

anpete:
:shipit:

maumar:
merged ddacf6e0a5f9f2322bab19e9b2b75bb332f9a379

Issue #1417, opened 2015-01-15
Metadata :: stack overflow when processing metadata of a model including 1:1 references and self reference


maumar:
consider the following:

```
    public class Level1
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public Level2 OneToOne_Required_PK { get; set; }
        public Level2 OneToOne_Optional_PK { get; set; }

        public Level1 OneToOne_Optional_Self { get; set; }
    }

    public class Level2
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public Level1 OneToOne_Required_PK_Inverse { get; set; }
        public Level1 OneToOne_Optional_PK_Inverse { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Level1> LevelOne { get; set; }
        public DbSet<Level2> LevelTwo { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Entity<Level1>().OneToOne(e => e.OneToOne_Required_PK, e => e.OneToOne_Required_PK_Inverse);
            modelBuilder.Entity<Level1>().OneToOne(e => e.OneToOne_Optional_PK, e => e.OneToOne_Optional_PK_Inverse);

            modelBuilder.Entity<Level1>().OneToOne(e => e.OneToOne_Optional_Self);
        }

        protected override void OnConfiguring(DbContextOptions options)
        {
            options.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "IncludeComplexTests", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }

    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var l1_01 = new Level1 { Name = "L1 01", };

                ctx.Add(l1_01);

                ctx.SaveChanges();
            }
        }
    }
```

This will throw stack overflow. Infinite call appears in:
Microsoft.Data.Entity.Metadata.ForeignKeyExtensions.GetRootPrincipals

mikary:
This issue does not appear to reproduce anymore.

[...]

Issue #1205, opened 2014-12-03
Model validation


AndriySvyryd:
The first time the model is used we should check it for validity to provide actionable exceptions.
This would include ambiguities found by conventions (#1171 is related).

rowanmiller:
@AndriySvyryd  can you drive design on this one

mikary:
Validation should also check for unbreakable key cycles like the ones described in #1417

mikary:
After further discussion on #1417, it looks like the validation should ensure that there is exactly one value generator per sequence of keys (key chain), and cycles should be fine as long as a generator exists.

AndriySvyryd:
Ambiguities found by conventions will be logged by conventions, see #1171

AndriySvyryd:
Fixed in 7741bfdc7a1372ac92301752333d597c14d8915e

PR #1548, opened 2015-02-05
Fix: Added check for cycles in GetRootPrincipals to prevent infinite recursion

mikary:
Fix for #1417 throwing ```InvalidOperationException``` in place of ```OutOfMemoryException```
@AndriySvyryd


src/EntityFramework.Core/Metadata/ForeignKey.cs
```diff
@@ -133,5 +133,10 @@ bool IForeignKey.IsRequired
         {
             get { return IsRequired ?? DefaultIsRequired; }
         }
+
+        public override string ToString()
+        {
+            return Properties.Select(p => EntityType.Name + "." + p.Name).Join(" ");
```
AndriySvyryd: This would repeat the entity type name, perhaps something like "EntityType {p1, p2}" ?

src/EntityFramework.Core/Metadata/ForeignKeyExtensions.cs
```diff
@@ -96,8 +98,21 @@ public static INavigation GetNavigationToDependent([NotNull] this IForeignKey fo
...
+
+        private static IEnumerable<IProperty> GetRootPrincipals(IForeignKey foreignKey, int propertyIndex, List<IForeignKey> visitedKeys)
+        {
             Check.NotNull(foreignKey, "foreignKey");
 
+            if (visitedKeys.Contains(foreignKey))
+            {
+                visitedKeys.Add(foreignKey);
+
+                throw new InvalidOperationException(Strings.CircularDependency(visitedKeys.Join(" -> ")));
```
AndriySvyryd: This will print all visited keys, not just the ones forming the cycle. You need to find where the repeated foreignKey is and start from there.

ajcvickers:
Is this the correct place to do this check? If this is an invalid model, then would it not be better to either block the creation of the invalid model or add this to the things we do during model validation. Doing it each time we need to find the principal means we fail for an invalid model in a more unpredictable way, rather than failing fast at the beginning. Also, we do the check many times, instead of just once.

mikary:
@ajcvickers @AndriySvyryd 
I've tried moving the cycle detection logic from the ```ForeignKeyExtensions``` to ```EntityType``` so that we can detect this issue as early as possible, but there are still a few things to consider.

* Up to this point, I have been assuming that the appropriate response to a model with a cycle in the metadata like this is a negative case and that we should throw (but with a recoverable exception instead of a ```StackOverflowException```). @AndriySvyryd has pointed out that there may be cycles like this where we have a nullable key and the cycle may be broken, but I don't believe any of the logic to support this kind of cycle breaking exists at this time.
* Assuming this kind of cycle is an invalid model state, adding detection at this point prevents the model from enter an invalid state, even if the cycle would have been broken later in the model building process. @ajcvickers expressed a concern over how this impacts conventions, but when I spoke with @AndriySvyryd he suggested that conventions shouldn't put the model into an invalid state, and if an exception is thrown, conventions can catch and react to the issue (i.e. use a different key to accomplish the same end goal).
* There is an open issue for model validation #1205 and I have made note of the related issue for validation, but we are already performing some basic validation (proper entity type on properties, no duplicate foreign keys) in the ```EntityType```.

I haven't finalized the changes, and there are a number of tests that begin to fail as a result of these changes due to another issue with the One-To-One mapping defaulting to PKs instead of using shadow state FKs.

mikary:
Until we have the changes to model building to ensure that each key chain (graph) contains a single generation property, these changes break a significant amount of functionality (i.e. 275 of the SQL Server functional tests fail)

divega:
> Until we have the changes to model building to ensure that each key chain (graph) contains a single generation property

@mikary is there an issue tracking that work? Could you cross reference it here?

AndriySvyryd:
This is tracked by https://github.com/aspnet/EntityFramework/issues/1205
I already have the changes ready, but they depend on https://github.com/aspnet/EntityFramework/pull/1741

Thanks,
Andriy

mikary:
This PR is ready for review again now that the other PR has been completed.
@ajcvickers @AndriySvyryd @divega

src/EntityFramework.Core/ChangeTracking/Internal/KeyPropagator.cs
```diff
@@ -99,22 +99,16 @@ private bool TryPropagateValue(InternalEntityEntry entry, IProperty property)
 
         private ValueGenerator TryGetValueGenerator(IProperty property)
         {
-            foreach (var foreignKey in property.EntityType.ForeignKeys)
+            if (property.IsKey())
```
AndriySvyryd: This is redundant since it's already checked before the call to TryGetValueGenerator.

AndriySvyryd:
:shipit:

PR #1550, opened 2015-02-05
Fix to #1528 - Exception thrown for multi-level include dotting over one-one navigation into one to many deeper inside the include chain

maumar:
Problem was that when we tried to extract list of PropertyInfos that include chain consists of, we were doing it incorrectly - always assuming it would be a simple property (or list of those inside anonymous type).
Fix is to use a proper method that expects complex properties instead.

### DIFF
--- a/src/EntityFramework.Core/Utilities/ExpressionExtensions.cs
+++ b/src/EntityFramework.Core/Utilities/ExpressionExtensions.cs
@@ -89,6 +89,34 @@ namespace System.Linq.Expressions
             return propertyInfos != null && propertyInfos.Length == 1 ? propertyInfos[0] : null;
         }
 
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

src/EntityFramework.Core/Utilities/ExpressionExtensions.cs
```diff
@@ -89,6 +89,34 @@ var propertyPath
             return propertyInfos != null && propertyInfos.Length == 1 ? propertyInfos[0] : null;
         }
 
+        public static PropertyInfo[] GetComplexPropertyAccess(this LambdaExpression propertyAccessExpression)
```
maumar: will add [NotNull]

anpete:
:shipit:

maumar:
merged c12501a2d884d240f83de7a6b236c9a37af28db2

PR #1553, opened 2015-02-06
Clean up relationships after Ignore

### LINKED ISSUE #1215: Calling Ignore should produce the same model as if the ignored object didn't exist in the first place
Since in the current paradigm the conventions run before Ignore we need to make the best effort to produce the expected model.

Implement Remove methods on internal metadata builders:
- [x] Remove entity type
  - [x] Removes all FKs on it and FKs referencing it
- [x] Remove property
  - [x] Simple property
    - [x] Calls Remove on all FKs, Indexes and keys that reference it
  - [x] Navigation property
    - [ ] Calls Remove on associated FK discovered by convention
- [x] Remove FK
  - [x] Removes any otherwise unused shadow properties that it spanned
  - [ ] Removes the relationship if it has same or lower configuration source, otherwise creates a new shadow FK
  - [ ] Removes unreachable entities if invoked from Ignore
- [x] Remove Key (Setting a PK calls this for the previous one)
  - [x] Removes any otherwise unused shadow properties that it spanned
  - [x] Removes referencing FKs
- [x] Remove Index
  - [x] Removes any otherwise unused shadow properties that it spanned

AndriySvyryd:
Remove unreachable entity type discovered by convention after ignoring a navigation
Remove relationship discovered by convention after ignoring the associated navigations
Replace a foreign key discovered by convention when a contained property is ignored

Resolves #1215

### DIFF
--- a/src/EntityFramework.Core/EntityFramework.Core.csproj
+++ b/src/EntityFramework.Core/EntityFramework.Core.csproj
@@ -297,7 +297,8 @@
     <Compile Include="Update\DbUpdateConcurrencyException.cs" />
     <Compile Include="Update\DbUpdateException.cs" />
     <Compile Include="Utilities\AsyncLock.cs" />
-    <Compile Include="Utilities\ModelUndirectedGraphAdapter.cs" />
+    <Compile Include="Utilities\ModelNavigationsGraphAdapter.cs" />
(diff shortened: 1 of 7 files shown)


anpete:
:shipit:

Issue #740, opened 2014-10-01
Handle differences in nullability between foreign and principal key properties


ajcvickers:
Fixup often involves taking an FK value and using it to set a principal key value, or vice versa. However, sometimes nullability is different between the two properties; the appropriate value conversions should be performed in such cases.

ajcvickers:
Note for triage: I don't know if this is still an issue since incremental changes may have fixed the issues. There is one place in the code that references this issue--search for #740. Other than that, test coverage should be added where needed and issues fixed as found. Anyone could work on this.

rowanmiller:
Ask @ajcvickers  about this one

mikary:
Reviewed and removed the comment relating to #740 and added test / fix in PR #1557
New test covers non-nullable FK to nullable primary case.

PR #1557, opened 2015-02-06
Fix to #740 Test coverage for different nullability in foreign and principal keys

mikary:
:white_check_mark: Add test for non-nullable FK with nullable PK
:bug: Fix InvalidOperationException in Query exposed by new test
@ajcvickers @AndriySvyryd


AndriySvyryd:
:shipit:

Issue #1385, opened 2015-01-11
Attach should not reject changes


ajcvickers:
Issue #742 made changing state from Modified to Unchanged be a reject changes operation, as it was in EF6. However, because Attach uses this directly it means that attaching an object that is Modified now also rejects changes. There are scenarios where this would be correct, but Attach mostly means "the object as it is now is how it is in the database" and hence if the object happens to already be modified with changes then these changes should not be rejected. This was the behavior in EF6.

Here is a full list of state change behaviors in EF6. "Special" behaviors bold.I believe that this bug covers the only one that is still different in EF7. (Other than throwing on Remove for detached objects, which I don't think we want to replicate in EF7.)

Action/Initial state | Unchanged | Modified | Added | Deleted | Detached
-----------------------|----------------|------------|----------|------------|-------------
Set Unchanged | Unchanged | **Unchanged (reject)** | Unchanged | Unchanged | Unchanged
Attach | Unchanged | Unchanged (accept) | Unchanged | Unchanged | Unchanged
Set Added | Added | Added | Added | Added | Added
Add | Added | Added | Added | Added | Added
Set Deleted | Deleted | Deleted | Deleted | Deleted | Deleted
Remove | Deleted | Deleted | **Detached** | Deleted | **Throws**
Set Modified | Modified | Modified | Modified | Modified | Modified
Set Detached | Detached | Detached | Detached | Detached | Detached

lukewaters:
Fixed in #1562

PR #1562, opened 2015-02-06
Issue #1385 Attach should not reject changes

lukewaters:
This adds a code path to attach entities and update saved values to the current values

### DIFF
--- a/src/EntityFramework.Core/DbContext.cs
+++ b/src/EntityFramework.Core/DbContext.cs
@@ -455,7 +455,11 @@ namespace Microsoft.Data.Entity
         {
             Check.NotNull(entity, "entity");
 
-            return SetEntityState(entity, EntityState.Unchanged);
+            var entry = EntryWithoutDetectChanges(entity);
+            var stateEntry = entry.StateEntry;
+            stateEntry.SetEntityState(EntityState.Unchanged, acceptChanges: true);
+
+            return new EntityEntry<TEntity>(entry.Context, stateEntry);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


lukewaters:
@ajcvickers

src/EntityFramework.Core/ChangeTracking/EntityEntry.cs
```diff
@@ -32,6 +32,13 @@ public virtual EntityState State
             }
         }
 
+        public virtual void AttachState(EntityState value)
```
ajcvickers: What is this method for?

src/EntityFramework.Core/DbContext.cs
```diff
@@ -511,6 +511,16 @@ private EntityEntry EntryWithoutDetectChanges([NotNull] object entity)
             return entry;
         }
 
+        private EntityEntry<TEntity> SetEntityStateAttached<TEntity>(TEntity entity, EntityState entityState) where TEntity : class
+        {
+            var entry = EntryWithoutDetectChanges(entity);
+
+            entry.AttachState(entityState);
+
+            return entry;
```
ajcvickers: I don't think it should be done this way. We don't want to add new surface to EntityEntry for this. Instead, get the StateEntry itself, set the state to Unchanged passing in true for the acceptChanges flag. Then create an EntityEntry around that StateEntry and return it.

ajcvickers: Also, make sure that all the Attach and AttachRange methods on DbContext and DbSet have this behavior, and add tests to ensure this.

src/EntityFramework.Core/ChangeTracking/EntityEntry.cs
lukewaters: Removed

ajcvickers:
:shipit:

whyi:
:shipit:

PR #1564, opened 2015-02-06
Adding doc comments to remaining public APIs in root namespace

rowanmiller:
Skipping some of types that will be moved out in the near future

### DIFF
--- a/src/EntityFramework.Core/EntityState.cs
+++ b/src/EntityFramework.Core/EntityState.cs
@@ -3,12 +3,37 @@
 
 namespace Microsoft.Data.Entity
 {
+    /// <summary>
+    /// The state in which an entity is being tracked by a context.
+    /// </summary>
     public enum EntityState
     {
+        /// <summary>
+        /// The entity is not being tracked by the context.
+        /// </summary>
         Unknown = 0,
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


src/EntityFramework.Core/EntityState.cs
```diff
@@ -3,12 +3,37 @@
...
+    /// </summary>
     public enum EntityState
     {
+        /// <summary>
+        /// The entity is not being tracked by the context.
+        /// </summary>
         Unknown = 0,
+
+        /// <summary>
+        /// The entity is being tracked by the context and exists in the database. Its property
```
lajones: Nit: Being picky but "exists in the database" only as far as EntityFramework knows. It could have been deleted/changed by something outside of EF's knowledge. But not sure how you state this succinctly. Perhaps "existed in the database last time the framework checked"?

Similarly below.

lajones:
:shipit:

src/EntityFramework.Core/EntityState.cs
ajcvickers: I think the simpler description is better. It is true that the database might be different for a number of reasons, but the point of having this state is to assert that the state of the entity being tracked is the same as the state in the database. I think the simple one line description captures this well.

ajcvickers:
:shipt:

PR #1572, opened 2015-02-09
Query: Caching - Implements full compiled query cache.

### LINKED ISSUE #1490: Query Caching: Cache should be per provider/model
Related to #1461

anpete:
- Rewrite of the LINQ provider front end to better intercept query execution.
- Implementation of query cache including partial evaluating parameterizer.
- Removal of explicit query exception interception; now compiled into query.
- Additional parameterization of query executors (capture less).
- Change to DataStore: DataStores now return query executor delegates, not results.
- Adaptive SQL gen: SQL gen'd per query execution allowing parameter values to influence SQL (null compensation).

TBD: Integration of MemoryCache in place of current naive impl.

Also fixes #1490


divega:
:shipit:

PR #1577, opened 2015-02-10
Tidy up files in root folder of EntityFramework.Core

rowanmiller:
Leaving just the M.D.E public API surface in root folder


src/EntityFramework.Core/Utilities/INotifyPropertyChanging.cs
```diff
@@ -7,16 +7,16 @@
 #if !NET45 && !ASPNET50 && !ASPNETCORE50
 // TODO: Remove this file when VS can build against ".NET Portable 5.0"
 
-namespace Microsoft.Data.Entity
+namespace Microsoft.Data.Entity.Utilities
```
rowanmiller: Put this one in `Utilities` rather than `Internal` because it is only used to build in VS and isn't included in the binaries we actually package up.

ajcvickers:
:shipit:

Issue #1515, opened 2015-02-02
Add support for key sentinel values


ajcvickers:
As per design meeting discussion: https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes:-January-29,-2015#key-value-sentinels

PR #1578, opened 2015-02-10
Add support for key sentinel values (Issue #1515)

ajcvickers:
Changes core metadata to add support for an invalid sentinel value on any property. Entities cannot be tracked with key values that match the sentinel just as they cannot be tracked with null key values. The default sentinel is default(T).

Will file new issues for:
- Fluent API
- Convention for setting sentinels for enum types
- Ensuring consistency of sentinels between principal/dependent keys


anpete:
:shipit:

Issue #1488, opened 2015-01-28
Include :: 'Index was outside the bounds of the array' thrown for some cases of multi include queries


maumar:
Repro:

```
    public class Level1
    {
        public int Id1 { get; set; }
        public string Name { get; set; }

        public Level2 OneToOne_Required_PK { get; set; }
        public Level2 OneToOne_Optional_PK { get; set; }

        public Level2 OneToOne_Required_FK { get; set; }
        public Level2 OneToOne_Optional_FK { get; set; }

        public ICollection<Level2> OneToMany_Optional { get; set; }
    }

    public class Level2
    {
        public int Id2 { get; set; }
        public string Name { get; set; }

        public int Level1_Required_Id { get; set; }
        public int? Level1_Optional_Id { get; set; }

        public Level3 OneToOne_Required_PK { get; set; }
        public Level3 OneToOne_Optional_PK { get; set; }

        public Level3 OneToOne_Required_FK { get; set; }
        public Level3 OneToOne_Optional_FK { get; set; }

        public ICollection<Level3> OneToMany_Optional { get; set; }

        public Level1 OneToOne_Required_PK_Inverse { get; set; }
        public Level1 OneToOne_Optional_PK_Inverse { get; set; }
        public Level1 OneToOne_Required_FK_Inverse { get; set; }
        public Level1 OneToOne_Optional_FK_Inverse { get; set; }

        public Level1 OneToMany_Required_Inverse { get; set; }
        public Level1 OneToMany_Optional_Inverse { get; set; }
    }

    public class Level3
    {
        public int Id3 { get; set; }
        public string Name { get; set; }

        public int Level2_Required_Id { get; set; }
        public int? Level2_Optional_Id { get; set; }

        public Level2 OneToOne_Required_FK_Inverse { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Level1> LevelOne { get; set; }
        public DbSet<Level2> LevelTwo { get; set; }
        public DbSet<Level3> LevelThree { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Entity<Level1>().Key(e => e.Id1);
            modelBuilder.Entity<Level2>().Key(e => e.Id2);
            modelBuilder.Entity<Level3>().Key(e => e.Id3);

            modelBuilder.Entity<Level1>().Property(e => e.Id1).GenerateValueOnAdd(false);
[...]

Issue #1478, opened 2015-01-27
Include :: 'index out of range' exception thrown in some cases of multi-include queries (need to run both queries in the repro)


maumar:
Repro:

```

    public class Level1
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public Level2 OneToOne_Required_PK { get; set; } 
        public Level2 OneToOne_Optional_PK { get; set; } 

        public Level2 OneToOne_Optional_FK { get; set; } 

        public ICollection<Level2> OneToMany_Optional { get; set; } 
        public ICollection<Level1> OneToMany_Optional_Self { get; set; } 
        public Level1 OneToMany_Optional_Self_Inverse { get; set; } 
    }

    public class Level2
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int Level1_Required_Id { get; set; }
        public int? Level1_Optional_Id { get; set; }

        public Level1 OneToOne_Required_PK_Inverse { get; set; }
        public Level1 OneToOne_Optional_PK_Inverse { get; set; }
        public Level1 OneToOne_Required_FK_Inverse { get; set; }
        public Level1 OneToOne_Optional_FK_Inverse { get; set; }

        public Level1 OneToMany_Optional_Inverse { get; set; } 
    }

    public class MyContext : DbContext
    {
        public DbSet<Level1> LevelOne { get; set; }
        public DbSet<Level2> LevelTwo { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Entity<Level1>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);

            modelBuilder.Entity<Level1>().HasOne(e => e.OneToOne_Required_PK).WithOne(e => e.OneToOne_Required_PK_Inverse).ReferencedKey<Level1>(e => e.Id).Required(true); 
            modelBuilder.Entity<Level1>().HasOne(e => e.OneToOne_Optional_PK).WithOne(e => e.OneToOne_Optional_PK_Inverse).ReferencedKey<Level1>(e => e.Id).Required(false);
            modelBuilder.Entity<Level1>().HasOne(e => e.OneToOne_Optional_FK).WithOne(e => e.OneToOne_Optional_FK_Inverse).ForeignKey<Level2>(e => e.Level1_Optional_Id).Required(false);
[...]

Issue #1433, opened 2015-01-19
Include :: IndexOutOfRange when running include queries with one-to-many and one-to-many self reference


maumar:
Repro:

```
    public class Level1
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public List<Level2> OneToMany_Optional { get; set; }
        public List<Level1> OneToMany_Optional_Self { get; set; }
        public Level1 OneToMany_Optional_Self_Inverse { get; set; }
    }

    public class Level2
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int? Level1_Optional_Id { get; set; }
        public Level1 OneToMany_Optional_Inverse { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Level1> LevelOne { get; set; }
        public DbSet<Level2> LevelTwo { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Entity<Level1>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level1>().OneToMany(e => e.OneToMany_Optional, e => e.OneToMany_Optional_Inverse).Required(false);
            modelBuilder.Entity<Level1>().OneToMany(e => e.OneToMany_Optional_Self, e => e.OneToMany_Optional_Self_Inverse).Required(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);
        }

        protected override void OnConfiguring(DbContextOptions options)
        {
            options.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "IncludeBug_IndexOutsideBounds", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var l1_01 = new Level1 { Id = 1, Name = "L1 01", };
                var l1_02 = new Level1 { Id = 2, Name = "L1 02", };
                var l1_03 = new Level1 { Id = 3, Name = "L1 03", };

[...]

PR #1588, opened 2015-02-11
Fix to multiple include bugs - #1433, #1478, #1487 #1488

maumar:
Problem was that if a query was executed multiple times, or after another query, some of the entities were stored in state manager and not in the reader buffer. However when accessing PK/FK we were always using the buffered readers. This could cause multitude of errors, e.g buffer being empty, trying to access element of the reader that doesn't exist, is of a wrong type, or worst case - cause data corruption.

Fix is to mark first element of the buffer for a given entity with a null value, if that entity is already in state manager - later, when we try to retrieve the keys we check for that null, and if we find it we use state manager instead.


maumar:
@anpete

anpete:
:shipit:

rowanmiller:
@maumar - is this merged now? If so, can you close this.

maumar:
merged in 617ad0aaa964949e4a620ef5a73fa4bbe2f12374

Issue #474, opened 2014-07-31
Query: Improve translation of String's StartsWith, EndsWith and Contains


divega:
**PROVIDERS BEWARE:**

Linq translation for methods Contains, EndsWith and StartsWith that we have in the Relational package uses LIKE operator, which may return incorrect results if the value parameter (what we are searching for) contains wildcard characters, e.g. '%' or '_'.

This issue addresses SqlServer and Sqlite providers, but all other providers will still use the old translation. Each provider that can be affected by this should implement their own MethodCallTranslators for Contains, EndsWith and StartsWith. 

Currently in EF7 a LINQ query like this:

```C#
var things = Things.Where(t => t.Name.StartsWith("a"));
```

Gets translated to SQL like this (note that I am simplifying the query and expanding parameter values for clarity):

```SQL
SELECT * FROM Things WHERE Name LIKE 'a%' ;
```

However, in order to return correct results, a LINQ query like this:

```C#
var underscoreAThings = Things.Where(t => t.Name.StartsWith("_a"));
```

Should be translated to SQL like this:

```SQL
SELECT * FROM Things WHERE Name LIKE '~_a%' ESCAPE '~';
```

The escaping accounts for SQL wildcard characters in the input string which should not be treated as wildcards (we can add a separate `Like()` method for passing patterns, but that belongs in a separate work item). 

When the input string is store correlated (e.g. is another column in the database instead of parameter or a literal in the query) using LIKE in the translation correctly becomes more difficult, e.g. it would be hard to perform the required escaping in SQL. 

In general for cases in which LIKE doesn't work well we can fall back to alternative translations that don't rely on LIKE, e.g. for `String.StartsWith()`:

```C#
var underscoreAThings = Things.Where(t => t.Name.StartsWith(t.Prefix));
```

```SQL
SELECT * FROM Things WHERE CHARINDEX(Prefix, Name) = 1 OR Prefix='';
```

Note that `CHARINDEX()` won't match an empty string but `String.StartsWith("")` always return `true`, that's why we add the Prefix ='' condition.

The main disadvantage of this translation is that it is not sargable. That can be addressed with a hybrid translation, e.g.:

```SQL
SELECT * FROM Things WHERE Name LIKE Prefix+'%' AND (CHARINDEX(Prefix, Name) = 1 OR Prefix = '');
```
 
[...]

PR #1589, opened 2015-02-11
Query: Caching: Added test coverage for more parameterization scenarios....

anpete:
... Fixed corresponding bugs.

### DIFF
--- a/src/EntityFramework.Core/Query/CompiledQueryCache.cs
+++ b/src/EntityFramework.Core/Query/CompiledQueryCache.cs
@@ -142,7 +142,7 @@ namespace Microsoft.Data.Entity.Query
             var cacheKey
                 = dataStore.Model.GetHashCode().ToString()
                   + isAsync
-                  + parameterizedQuery;
+                  + query;
 
             var compiledQuery
                 = _memoryCache.GetOrSet(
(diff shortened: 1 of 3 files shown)


test/EntityFramework.SqlServer.FunctionalTests/SqlServerQueryTest.cs
```diff
@@ -1238,7 +1441,7 @@ public override void String_StartsWith_MethodCall()
             Assert.Equal(
                 @"SELECT [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[CustomerID], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
 FROM [Customers] AS [c]
-WHERE [c].[ContactName] LIKE 'M' + '%'",
+WHERE [c].[ContactName] LIKE @__LocalMethod1_0 + '%'",
```
divega: Can't remember if we talked about this already but these should use escaping, e.g. in case the value is something like "_m".

divega:
:shipit:

test/EntityFramework.SqlServer.FunctionalTests/SqlServerQueryTest.cs
divega: I had already filed this as #474.

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -225,63 +237,64 @@ public override Expression VisitExpression(Expression expression)
...
-                            : fieldInfo.Name;
+                        var memberExpression = (MemberExpression)expression;
+                        var @object = Evaluate(memberExpression.Expression, out parameterName);
 
-                        parameterValue = fieldInfo.GetValue(@object);
+                        var fieldInfo = memberExpression.Member as FieldInfo;
 
-                        return true;
-                    }
+                        if (fieldInfo != null)
```
divega: @anpete thanks for making this immune to https://entityframework.codeplex.com/workitem/1108 :smile:

anpete: I'm not sure that does what you think :-) But I know how to make it do what you suggest...

divega: Do'h, you are right, I was thinking `@object` should not be null. I was thinking you can just fallback to the lambda compilation to fail with the right exception. Want me to file a bug? :smile:

Issue #1539, opened 2015-02-03
Have a way to obtain a generic PropertyEntry<TEntity, TProperty> from an EntityEntry<TEntity> and a property name string


divega:
Currently we return the generic `PropertyEntry` only from the lambda overload of `Property()`.

smitpatel:
fixed via aff88940320650f9af8f3852124d8da94d0e93d4

Issue #1538, opened 2015-02-03
Generic PropertyEntry should have typed CurrentValues and OriginalValues


divega:
In previous versions of EF we came up with a clever conversion-based pattern to flip between generic and non-generic versions of entry types. We now think the conversion mechanism was overkill and we don't want to do the same in EF7, but we can use simple member hiding to obtain `TProperty` typed `CurrentValue` and `OriginalValue` properties. Note that without this there isn't really much of a point on having a generic version of `PropertyEntry`.

smitpatel:
fixed via aff88940320650f9af8f3852124d8da94d0e93d4

PR #1597, opened 2015-02-11
Enhancements to Generic PropertyEntry

smitpatel:
Adds typed CurrentValue & OriginalValue to Generic PropertyEntry #1538 
Adds generic Property function in Generic EntityEntry which returns Generic PropertyEntry #1539

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/EntityEntry`.cs
+++ b/src/EntityFramework.Core/ChangeTracking/EntityEntry`.cs
@@ -6,6 +6,7 @@ using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.ChangeTracking.Internal;
 using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.ChangeTracking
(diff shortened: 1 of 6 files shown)


src/EntityFramework.Core/ChangeTracking/PropertyEntry`.cs
```diff
@@ -11,5 +11,19 @@ public PropertyEntry([NotNull] StateEntry stateEntry, [NotNull] string name)
...
+            get { return (TProperty)base.CurrentValue; }
+            [param: CanBeNull]
+            set { base.CurrentValue = value; }
+        }
+
+        public virtual new TProperty OriginalValue
+        {
+            get { return (TProperty)base.OriginalValue; }
+            [param: CanBeNull]
+            set { base.OriginalValue = value; }
```
anpete: Should this be settable?

divega: Yes. It is useful for setting original values for concurrency or FK properties. You think it shouldn't?

src/EntityFramework.Core/ChangeTracking/EntityEntry`.cs
```diff
@@ -26,5 +26,13 @@ public EntityEntry([NotNull] DbContext context, [NotNull] StateEntry stateEntry)
 
             return new PropertyEntry<TEntity, TProperty>(StateEntry, propertyInfo.Name);
         }
+
+        public virtual PropertyEntry<TEntity, TProperty> Property<TProperty>(
```
divega: Do we have a negative test cases for passing wrong name and/or type? Sorry if I missed it.

smitpatel: I added test for wrong property name
Currently there is no mechanism to throw if wrong type is passed.

src/EntityFramework.Core/ChangeTracking/EntityEntry`.cs
```diff
@@ -28,5 +28,13 @@ public EntityEntry([NotNull] DbContext context, [NotNull] InternalEntityEntry in
 
             return new PropertyEntry<TEntity, TProperty>(InternalEntry, propertyInfo.Name);
         }
+
+        public virtual PropertyEntry<TEntity, TProperty> Property<TProperty>(
+            [NotNull] string propertyName)
+        {
+            Check.NotNull(propertyName, nameof(propertyName));
+
+            return new PropertyEntry<TEntity, TProperty>(InternalEntry, propertyName);
```
divega: It appears that this won't throw if TProperty is the wrong type because the strong types wrong on the generic PropertyEntry is pure sugar around the non-generic, e.g. the type is not even passed or verified in the constructor. You would very likely get an invalid cast exception later when trying to set or get the values. I believe we should validate the type and throw a useful exception but I would like to hear what @ajcvickers has to say.

ajcvickers: Yes, we should do early validation.

smitpatel: Added validation and test.

divega:
:shipit:

PR #1598, opened 2015-02-11
API review renaming and namespace changes (no functional changes)

ajcvickers:
For core classes and ChangeTracker classes.
Also some additional code cleanup.


tools/Resources.cs
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+
```
rowanmiller: Is this file supposed to be here?

rowanmiller:
:sheep: :pig:

tools/Resources.cs
ajcvickers: Reverted.

PR #1599, opened 2015-02-12
Don't derive IForeignKey from IKey

### LINKED ISSUE #1034: IForeignKey should not derive from IKey
I believe we decided this previously in [a design meeting](https://github.com/aspnet/EntityFramework/wiki/Entity-Framework-Design-Meeting-Notes---September-11,-2014#metadata-for-keys-foreign-keys-and-indexes), but that may have just been for `IIndex`.

Foreign keys are not keys. They reference either a primary key or an alternate/unique key, but they do not themselves identity the entity type on which they're defined. Our current implementation uses inheritance for the `Properties` and `EntityType` properties, but these should just be copied into `IForeignKey` since the semantics of inheritance here are not correct.

AndriySvyryd:
Resolves #1034

### DIFF
--- a/src/EntityFramework.Core/Metadata/ForeignKey.cs
+++ b/src/EntityFramework.Core/Metadata/ForeignKey.cs
@@ -10,7 +10,7 @@ using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Metadata
 {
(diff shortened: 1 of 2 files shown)


lukewaters:
:shipit:

bricelam:
:shipit: :shipit: :shipit: :stuck_out_tongue_winking_eye:

PR #1600, opened 2015-02-12
Remove SchemaQualifiedName

### LINKED ISSUE #1072: Remove SchemaQualifiedName
In some parts of the code we use this and in other parts we use schema and table name separately. In some places it makes sense to treat them separately, such as in the code that accesses and manipulates annotations. However, when it comes to code that is reasoning about table names in relational code we should probably consistently use this class or if it doesn't make sense then get rid of it.

smitpatel:
Resolves #1072


smitpatel:
Parsing method in SchemaQualifiedName was being used in unit testing only. Is that required? If yes where it should be put?

src/EntityFramework.Relational/SqlGenerator.cs
```diff
@@ -71,23 +73,26 @@ public abstract class SqlGenerator
...
 
         public virtual void AppendInsertCommand(
             [NotNull] StringBuilder commandStringBuilder,
-            SchemaQualifiedName schemaQualifiedName,
+            [NotNull] string tableName,
+            [CanBeNull] string schemaName,
             [NotNull] IReadOnlyList<ColumnModification> writeOperations)
         {
             Check.NotNull(commandStringBuilder, "commandStringBuilder");
+            Check.NotNull(tableName, "tableName");
```
bricelam: We typically use `Check.NotEmpty()` for strings.

src/EntityFramework.Relational/Update/ModificationCommand.cs
```diff
@@ -147,5 +151,15 @@ public virtual void PropagateResults([NotNull] IValueReader reader)
...
+
+        public virtual string EscapedTableName => Escape(TableName);
+        public virtual string EscapedSchemaName => Escape(SchemaName);
+
+        private static string Escape(string name)
+        {
+            return name.IndexOfAny(new[] { ']', '[', '.' }) != -1
+                ? "[" + name.Replace("]", "]]") + "]"
+                : name;
+        }
```
bricelam: Are these used anywhere? If not, remove.

bricelam:
No need to keep the parsing functionality.

bricelam:
After addressing feedback, :shipit:

src/EntityFramework.Relational/Update/ModificationCommand.cs
smitpatel: There is a test which uses them. Though test works without escaping also. Do we need escaping code somewhere while generating sql?

bricelam: We have escaping elsewhere. This type of escaping is SQL Server-specific anyway and doesn't belong in Relational.

Issue #1453, opened 2015-01-21
Topological sort :: exception thrown in case of circular dependency is useless


maumar:
Currently exception looks like this:

Unhandled Exception: System.InvalidOperationException: A circular model foreign
key dependency was detected: Microsoft.Data.Entity.Relational.Update.ModificationCommand -> Microsoft.Data.Entity.Relational.Update.ModificationCommand -> Microsoft.Data.Entity.Relational.Update.ModificationCommand.

we do ToString() on the verteces - instead we should extract useful infromation from them (i.e. entity name and its key)

Issue #1454, opened 2015-01-21
Topological sort :: Exception during SaveChanges thrown for models that specify both ForeignKey and ReferenceKey on 1:1 relationship


maumar:
```
using Microsoft.Data.Entity;
using Microsoft.Data.Entity.Metadata;
using System.Data.SqlClient;
using System.Linq;

namespace Repro1454
{
    public class Level1
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public Level2 OneToOne_Required_FK { get; set; }
        public Level2 OneToOne_Optional_FK { get; set; }
    }

    public class Level2
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int Level1_Required_Id { get; set; }
        public int? Level1_Optional_Id { get; set; }

        public Level3 OneToOne_Required_PK { get; set; }
        public Level3 OneToOne_Optional_PK { get; set; }

        public Level3 OneToOne_Required_FK { get; set; }
        public Level3 OneToOne_Optional_FK { get; set; }

        public Level1 OneToOne_Required_FK_Inverse { get; set; }
        public Level1 OneToOne_Optional_FK_Inverse { get; set; }
    }

    public class Level3
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public int Level2_Required_Id { get; set; }
        public int? Level2_Optional_Id { get; set; }

        public Level2 OneToOne_Required_PK_Inverse { get; set; }
        public Level2 OneToOne_Optional_PK_Inverse { get; set; }
        public Level2 OneToOne_Required_FK_Inverse { get; set; }
        public Level2 OneToOne_Optional_FK_Inverse { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Level1> LevelOne { get; set; }
        public DbSet<Level2> LevelTwo { get; set; }
        public DbSet<Level3> LevelThree { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            base.OnModelCreating(modelBuilder);

            modelBuilder.Entity<Level1>().Property(e => e.Id).GenerateValueOnAdd(false);
            modelBuilder.Entity<Level2>().Property(e => e.Id).GenerateValueOnAdd(false);

            modelBuilder.Entity<Level1>().HasOne(e => e.OneToOne_Required_FK).WithOne(e => e.OneToOne_Required_FK_Inverse).ForeignKey<Level2>(e => e.Level1_Required_Id).Required(true);
[...]

PR #1604, opened 2015-02-13
Add Multigraph with topological sort and cycle breaking

mikary:
This is part of the fix for #1453. It does not improve the error messages, but with cycle breaking a verbose cycle message may not be necessary
@bricelam @maumar

### DIFF
--- a/src/EntityFramework.Core/Utilities/Multigraph.cs
+++ b/src/EntityFramework.Core/Utilities/Multigraph.cs
@@ -1,0 +1,254 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
+
+namespace Microsoft.Data.Entity.Utilities
+{
+    public class Multigraph<TVertex, TEdge> : Graph<TVertex>
+    {
+        private readonly HashSet<TVertex> _vertices = new HashSet<TVertex>();
+        private readonly HashSet<TEdge> _edges = new HashSet<TEdge>();
+        private readonly Dictionary<TVertex, Dictionary<TVertex, List<TEdge>>> _successorMap = new Dictionary<TVertex, Dictionary<TVertex, List<TEdge>>>();
+
+        public virtual IEnumerable<TEdge> Edges
+        {
+            get { return _edges; }
+        }
+
+        public virtual IEnumerable<TEdge> GetEdges([NotNull] TVertex from, [NotNull] TVertex to)
+        {
+            Dictionary<TVertex, List<TEdge>> successorSet;
+            if (_successorMap.TryGetValue(from, out successorSet))
+            {
+                List<TEdge> edgeList;
+                if (successorSet.TryGetValue(to, out edgeList))
+                {
+                    return edgeList;
+                }
+            }
+            return Enumerable.Empty<TEdge>();
+        }
+
+        public virtual void AddVertex([NotNull] TVertex vertex)
+        {
+            Check.NotNull(vertex, nameof(vertex));
+
+            _vertices.Add(vertex);
+        }
+
+        public virtual void AddVertices([NotNull] IEnumerable<TVertex> verticies)
+        {
+            Check.NotNull(verticies, nameof(verticies));
+
+            _vertices.UnionWith(verticies);
+        }
+
+        public virtual void AddEdge([NotNull] TVertex from, [NotNull] TVertex to, [NotNull] TEdge edge)
+        {
+            AddEdges(from, to, new[] { edge });
+        }
+
+        public virtual void AddEdges([NotNull] TVertex from, [NotNull] TVertex to, [NotNull] IEnumerable<TEdge> edges)
+        {
+            Check.NotNull(from, nameof(from));
+            Check.NotNull(to, nameof(to));
+            Check.NotNull(edges, nameof(edges));
+
+            if (!_vertices.Contains(from))
+            {
+                throw new InvalidOperationException(Strings.GraphDoesNotContainVertex(from));
+            }
+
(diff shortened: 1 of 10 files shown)


bricelam:
Looks good. Want to take a stab at updating `ModelDiffer` to break cycles? [Here](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Migrations/Infrastructure/ModelDiffer.cs#L127-L146) is the current implementation.

src/EntityFramework.Core/Utilities/Multigraph.cs
```diff
@@ -0,0 +1,192 @@
...
+
+                    var canidateVertices = predecessorCounts.Keys.ToList();
+                    var canidateIndex = 0;
+
+                    // Iterrate over the unsorted verticies
+                    while (canidateIndex < canidateVertices.Count && !broken && canBreakEdge != null)
+                    {
+                        var canidateVertex = canidateVertices[canidateIndex];
+
+                        // Find verticies in the unsorted portion of the graph that have edges to the canidate
```
divega: typo "canidate"

divega:
@mikary This PR just adds some building blocks and doesn't attempt to fix any issues, correct?I would like to see this applied to solving the problem before it gets checked in.

mikary:
@bricelam @divega I added basic cycle breaking logic to the ModelDiffer, though it is perhaps overly defensive and I haven't added cycle error messages yet. Is this about what we want the pattern to look like?

src/EntityFramework.Relational/Migrations/Infrastructure/ModelDiffer.cs
```diff
@@ -140,14 +139,27 @@ protected virtual IReadOnlyList<MigrationOperation> Sort([NotNull] IEnumerable<M
...
+                        {
+                            alterOperations.Add(foreignKeyOperation);
+                        }
+                        else
+                        {
+                            return false;
+                        }
+                    }
+                    return true;
+                }).Select(k => createTableOperations[k]))
```
bricelam: This expression got complex. You should probably store the sorted create table operations in a temporary variable before the return statement.

src/EntityFramework.Relational/Migrations/Infrastructure/ModelDiffer.cs
```diff
@@ -140,14 +139,27 @@ protected virtual IReadOnlyList<MigrationOperation> Sort([NotNull] IEnumerable<M
...
             }
 
             return dropOperations
                 .Concat(createSequenceOperations)
-                .Concat(createTableGraph.TopologicalSort().SelectMany(l => l).Select(k => createTableOperations[k]))
+                .Concat(createTableGraph.TopologicalSort((principalTable, key, foreignKeyOperations) =>
+                {
+                    foreach (var foreignKeyOperation in foreignKeyOperations)
+                    {
+                        if (createTableOperations[key].ForeignKeys.Remove(foreignKeyOperation))
```
bricelam: Is this ever going to return `false`?

bricelam:
You should add (at lease) a simple test showing that the model differ can break foreign key cycles.

src/EntityFramework.Relational/Migrations/Infrastructure/ModelDiffer.cs
mikary: It shouldn't, the keys are all ones we've already iterated over previously.

src/EntityFramework.Core/Utilities/Multigraph.cs
```diff
@@ -0,0 +1,197 @@
...
+                            var incommingNeighbour = incommingNeighbours[neighbourIndex];
+                            // Check to see if the edge can be broken
+                            if (canBreakEdge(incommingNeighbour, candidateVertex, _successorMap[incommingNeighbour][candidateVertex]))
+                            {
+                                predecessorCounts[candidateVertex]--;
+                                if (predecessorCounts[candidateVertex] == 0)
+                                {
+                                    sortedQueue.Add(candidateVertex);
+                                    predecessorCounts.Remove(candidateVertex);
+                                    broken = true;
```
AndriySvyryd: You can replace the inner while with a foreach and use break statement

mikary:
Updated based on PR feedback and added unit test for ```ModelDiffer``` Also fixed key not found issue #1454 in ```BidirectionalAdjacencyListGraph``` since I was adding comments to that file anyway.

AndriySvyryd:
:shipit:

src/EntityFramework.Relational/Migrations/Infrastructure/ModelDiffer.cs
```diff
@@ -140,14 +139,24 @@ protected virtual IReadOnlyList<MigrationOperation> Sort([NotNull] IEnumerable<M
...
+                (principalTable, key, foreignKeyOperations) =>
+                {
+                    foreach (var foreignKeyOperation in foreignKeyOperations)
+                    {
+                        createTableOperations[key].ForeignKeys.Remove(foreignKeyOperation);
+                        alterOperations.Add(foreignKeyOperation);
+                    }
+                    return true;
+                })
+                .Select(k => createTableOperations[k]);
```
bricelam: I'd call `.ToList()` on this. It seems bad to modify the `alterOperations` collection as it lazily evals in the return statement.

mikary: The ```TopologicalSort``` call isn't lazy so ```ToList()``` would only really impact the ```Select(k => createTableOperations[k])``` which is fetching from a dictionary. That being said the lazyness is an implementation detail and not necessarily part of the contract so I could see adding it just to be explicit.

Issue #649, opened 2014-09-11
Query: support conditional expressions


divega:
E.g. db.Products.Select(p => new {p.Name, IsExpensive = p.Price >  100});

divega:
Assigning to @maumar since he is already working on a few issue that we think are related.

smitpatel:
We have support for conditional expressions in where clause already. #4141 adds support in projection.

anpete:
@smitpatel This one is projection, too.

PR #1605, opened 2015-02-13
Fixes #1439 - Ignore Convert expressions in relational filtering visitor...

### LINKED ISSUE #1439: Where() on properties of the short type does not generate a sql WHERE clause.
I have a table `Table1` with the following (relevant) columns: Id, SmallintId, IntId. SmallintId is of type smallint not null in SQL Server, IntId is of type int not null, and the `Table1` POCO model has a SmallintId property of the `short` type and an IntId of the `int` type.

I have the following observations:
The following queries does not produce any sql `WHERE` clause:

    context.Table1.Where(x => x.SmallintId == 17)
    context.Table1.Where(x => x.IntId == 42).Where(x => x.SmallintId == 17)

The following queries does produce a sql `WHERE` clause but *only* for the IntId column:

    context.Table1.Where(x => x.IntId == 42)
    context.Table1.Where(x => x.SmallintId == 17).Where(x => x.IntId == 42)

This issue might be related with #1426.

anpete:
....

### DIFF
--- a/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
@@ -256,8 +256,7 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
                     return Expression.Not(operand);
                 }
             }
-            else if (expression.NodeType == ExpressionType.Convert
-                     && expression.Type.IsNullableType())
+            else if (expression.NodeType == ExpressionType.Convert)
             {
                 var operand = VisitExpression(expression.Operand);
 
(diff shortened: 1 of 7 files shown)


src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
```diff
@@ -256,8 +256,7 @@ protected override Expression VisitUnaryExpression([NotNull] UnaryExpression exp
                     return Expression.Not(operand);
                 }
             }
-            else if (expression.NodeType == ExpressionType.Convert
-                     && expression.Type.IsNullableType())
+            else if (expression.NodeType == ExpressionType.Convert)
```
divega: Is this visitor applied on all parts of the query?

divega:
:shipit:

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
anpete: Most places it matters. I will need to do similar for #649.

PR #1608, opened 2015-02-13
Reference Caching assemblies in Commands functional tests (Fixes #1591)

### LINKED ISSUE #1591: Migrations ExecutorTest fail after adding InMemory cache
Tests throw TargetInvocationException when running in VS due to not being able to load the cache assembly.

bricelam:


ajcvickers:
:shipit:

PR #1609, opened 2015-02-13
Enable EF6 Perf Tests

rowanmiller:
Also upgrading EF6 tests to EF6.1.2
Have tested that build.cmd now passes on a CI agent as it previously failed when these tests were enabled


lajones:
:shipit:

lukewaters:
:shipit:

PR #1614, opened 2015-02-13
Use IAccessor pattern for EntityFrameworkServicesBuilder

ajcvickers:
Also move IAccessor to infrastructure namespace


AndriySvyryd:
:shipit:

PR #1616, opened 2015-02-13
[TINY] Disable failing .NET Core perf tests

rowanmiller:
These tests are hitting the same dynamic code generation issues that have resulted in us disabling a number of other test projects on CoreCLR. They don't show up on standard CI because perf tests are run on a separate CI server.


bricelam:
:shipit:

Issue #1251, opened 2014-12-08
Query :: join + orderby + count produces invalid sql


maumar:
Consider the following query:

```
(from c in cs
                 join o in os on c.CustomerID equals o.CustomerID
                 orderby c.CustomerID
                 select c).Count()
```

This produces incorrect sql:

```
SELECT COUNT(*)
FROM [Customers] AS [c]
INNER JOIN [Orders] AS [o] ON [c].[CustomerID] = [o].[CustomerID]
ORDER BY [c].[CustomerID]
```

which throws:

Column "Customers.CustomerID" is invalid in the ORDER BY clause because it is not contained in either an aggregate function or the GROUP BY clause.

We should probably prune the orderby call as it doesn't do anything

maumar:
fixed in 5ca3664a766edd2d47c04efbe03d7d1d78c12a83 and 1f8938316864038a69d9ec1b28a239f5dd28799e

PR #1618, opened 2015-02-13

maumar:
SELECT COUNT(*)
FROM [Customer] AS [c]
ORDER BY [c].[Id]

Fix is to remove any orderings when querying for count, just like we already do for some other result operators (e.g. Any)


maumar:
@anpete

PR #1621, opened 2015-02-14
Add a model validation step after model creation.

AndriySvyryd:
Add a model validation rule to detect shadow keys.

### DIFF
--- a/src/EntityFramework.Core/Metadata/IModel.cs
+++ b/src/EntityFramework.Core/Metadata/IModel.cs
@@ -27,6 +27,9 @@ namespace Microsoft.Data.Entity.Metadata
         IEnumerable<IForeignKey> GetReferencingForeignKeys([NotNull] IEntityType entityType);
 
         [NotNull]
+        IEnumerable<IForeignKey> GetReferencingForeignKeys([NotNull] IKey key);
+
+        [NotNull]
         IEnumerable<IForeignKey> GetReferencingForeignKeys([NotNull] IProperty property);
     }
 }
--- a/src/EntityFramework.Core/Properties/Strings.resx
+++ b/src/EntityFramework.Core/Properties/Strings.resx
@@ -1,7 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-
 <root>
-    <!-- 
+  <!-- 
     Microsoft ResX Schema 
     
     Version 2.0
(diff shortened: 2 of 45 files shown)


src/EntityFramework.Core/Infrastructure/ThrowingModelValidator.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Data.Entity.Infrastructure
+{
+    public class ThrowingModelValidator : ModelValidatorBase
```
ajcvickers: Not sure we should include the validator that throws on warnings in product code at this time. YAGNI and if we do want this it may need to have stuff like the ability to ignore warnings. For now, I think anyone who wants this could easily write it themselves.

src/EntityFramework.Core/Metadata/IModel.cs
```diff
@@ -27,6 +27,9 @@ public interface IModel : IMetadata
         IEnumerable<IForeignKey> GetReferencingForeignKeys([NotNull] IEntityType entityType);
 
         [NotNull]
+        IEnumerable<IForeignKey> GetReferencingForeignKeys([NotNull] IKey key);
```
ajcvickers: I'm not sure if these should be in IModel or extension methods. Fine with having them in IModel for now.

src/EntityFramework.Core/Properties/Strings.resx
```diff
@@ -1,7 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-
 <root>
-    <!-- 
+  <!--
```
ajcvickers: Are the changes here just formatting?

ajcvickers: Ah, we can now have curlies in the the messages?

ajcvickers:
:shipit:

src/EntityFramework.Core/Infrastructure/ThrowingModelValidator.cs
AndriySvyryd: I'll move it to test code

src/EntityFramework.Core/Metadata/IModel.cs
AndriySvyryd: This allows implementations to be better optimized

src/EntityFramework.Core/Properties/Strings.resx
AndriySvyryd: That's just what VS did to the file while editing it.

src/EntityFramework.Core/Metadata/IModel.cs
ajcvickers: Yes, that's why I'm not sure whether it should be or not. :-)

PR #1622, opened 2015-02-14
API review changes to ChangeTracker

ajcvickers:
- Use IAccessor for StateManager
- Remove all but most basic TrackGraph methods
  - Includes removal of related infrastcuture
  - Made the KeyValueEntityTracker public so it can be easily used with TrackGraph


divega:
:shipit:

PR #1624, opened 2015-02-15
API review updates to ChangeTracker interfaces

ajcvickers:
Removed Sidecar property notifications as they were no longer needed. Any property change is treated the same way regardless of whether it goes through a transparent sidecar or not.

Split IRelationshipListener into three interfaces.


bricelam:
:shipit:

PR #1626, opened 2015-02-15
ValueGeneration naming/namespace changes

ajcvickers:
Also some code cleanup. No functional changes.


src/EntityFramework.SqlServer/SequentialGuidValueGenerator.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 using System.Threading;
-using Microsoft.Data.Entity.Identity;
+using Microsoft.Data.Entity.ValueGeneration;
```
divega: Filed #1627 before we forget. Also consider using `nameof` in the `Next()` method.

divega:
:shipit:

PR #1630, opened 2015-02-16
Structural changes to value generation

ajcvickers:
- Introduction of abstract base classes
- Move SequentialGuidValueGenerator into core
- Move SequenceName into SQL Server generator
- Some additional cleanup


divega:
:shipit:

PR #1633, opened 2015-02-16
Move all value generator type information into the generator

ajcvickers:
This removes the need to pass IProperty to the Next method. Instead the value generator factory will always create a generator specifically for the type being targetted. As requested, the value generators are also now generic, with the Next method types appropriately. The generators do not attempt to avoid boxing since all uses just call the version of Next that returns an object and hence will always box anywhere. However, we could with the generic method start creating new generators in the future that do not box if it becomes useful to do so.


anpete:
:shipit:

PR #1636, opened 2015-02-16
Metadata: Adds initial support for entity type inheritance.

anpete:
Also some work on setting up fixtures for inheritance FTs.

NB: Metadata support is not complete here. This should be enough to get started on feature work though.

### DIFF
--- a/src/EntityFramework.Core/Metadata/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/EntityType.cs
@@ -7,6 +7,7 @@ using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Utilities;
@@ -15,7 +16,7 @@ namespace Microsoft.Data.Entity.Metadata
 {
     public class EntityType : MetadataBase, IEntityType
     {
-        private readonly object _typeOrName;
+        private static readonly char[] _simpleNameChars = { '.', '+' };
 
         private readonly LazyRef<SortedDictionary<IReadOnlyList<Property>, ForeignKey>> _foreignKeys
             = new LazyRef<SortedDictionary<IReadOnlyList<Property>, ForeignKey>>(()
@@ -36,12 +37,14 @@ namespace Microsoft.Data.Entity.Metadata
             = new LazyRef<SortedDictionary<IReadOnlyList<Property>, Key>>(() =>
                 new SortedDictionary<IReadOnlyList<Property>, Key>(PropertyListComparer.Instance));
 
+        private readonly object _typeOrName;
+
+        private EntityType _baseType;
         private Key _primaryKey;
 
         private int _shadowPropertyCount;
         private int _originalValueCount;
         private bool _useEagerSnapshots;
-        private static readonly char[] _simpleNameChars = { '.', '+' };
 
         /// <summary>
         ///     This constructor is intended only for use when creating test doubles that will override members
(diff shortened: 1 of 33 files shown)


src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -76,38 +82,88 @@ public EntityType([NotNull] Type type, [NotNull] Model model)
...
+                {
+                    if (RootType == value.RootType)
+                    {
+                        throw new InvalidOperationException(Strings.CircularInheritance(this, value));
+                    }
+
+                    foreach (var property in value.Properties.Where(property => _properties.ContainsKey(property.Name)))
+                    {
+                        throw new InvalidOperationException(Strings.DuplicateProperty(property.Name, Name));
+                    }
```
anpete: I need to add a check that we don't have any keys here...Any other validations?

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -76,38 +82,88 @@ public EntityType([NotNull] Type type, [NotNull] Model model)
...
+                }
+
+                _baseType = value;
+
+                UpdateIndexes();
+                UpdateShadowIndexes();
+                UpdateShadowIndexes();
```
anpete: Will fix.

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -76,38 +82,94 @@ public EntityType([NotNull] Type type, [NotNull] Model model)
...
-                var type = _typeOrName as Type;
-                return type != null ? type.FullName : (string)_typeOrName;
+                if (_baseType == value)
+                {
+                    return;
+                }
+
+                if (value != null)
+                {
+                    if (RootType == value.RootType)
```
AndriySvyryd: If A derives from B and C derives from B than this would make it impossible to change A to derive from C however there's no cycle,

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -329,6 +412,14 @@ public virtual IReadOnlyList<Key> Keys
             }
         }
 
+        private void ThrowifDerivedEntity([CallerMemberName] string caller = null)
```
AndriySvyryd: Didn't know that throwif was a word.

test/EntityFramework.Core.FunctionalTests/InheritanceTestBase.cs
```diff
@@ -0,0 +1,34 @@
...
+        where TFixture : InheritanceFixtureBase, new()
+    {
+//        [Fact]
+//        public virtual void Can_query_all_animals()
+//        {
+//            using (var context = CreateContext())
+//            {
+//                Assert.Equal(2, context.Animals.ToList().Count);
+//            }
+//        }
```
AndriySvyryd: If you want to leave this test just comment out [Fact], otherwise delete

AndriySvyryd:
Shouldn't Indexes, FKs, etc also be inherited or is that will just come later?

anpete:
@AndriySvyryd That will come later.

src/EntityFramework.Core/Metadata/EntityType.cs
anpete: Typo, will fix.

src/EntityFramework.Core/Metadata/EntityType.cs
anpete: You could set it to null first ;-) But I will improve this anyway.

test/EntityFramework.Core.FunctionalTests/InheritanceTestBase.cs
anpete: Fixed.

AndriySvyryd:
:shipit:

PR #1643, opened 2015-02-17

maumar:
SELECT COUNT(*)
FROM [Customer] AS [c]
ORDER BY [c].[Id]

Fix is to remove any orderings when querying for count.

There was another small issue that has been fixed - when calculating whether client eval for filter is required, we would use a common variable for the entire query. This could cause issues if multiple filters were present in the query, some being client eval, some being server eval. Fix is to use a variable local to the particular filter operation, and |= the results with the global variable.

### DIFF
--- a/src/EntityFramework.Core/Query/QueryOptimizer.cs
+++ b/src/EntityFramework.Core/Query/QueryOptimizer.cs
@@ -9,6 +9,7 @@ using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Clauses.ExpressionTreeVisitors;
+using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Transformations;
 
 namespace Microsoft.Data.Entity.Query
@@ -102,6 +103,20 @@ namespace Microsoft.Data.Entity.Query
             {
                 queryAnnotation.QuerySource = fromClause;
             }
+        }
+
+        public override void VisitResultOperator(ResultOperatorBase resultOperator, QueryModel queryModel, int index)
+        {
+            if (resultOperator is AnyResultOperator || resultOperator is CountResultOperator || resultOperator is LongCountResultOperator)
+            {
+                var orderByClauses = queryModel.BodyClauses.OfType<OrderByClause>().ToList();
+                foreach (var orderByClause in orderByClauses)
+                {
+                    queryModel.BodyClauses.Remove(orderByClause);
+                }
+            }
+
+            base.VisitResultOperator(resultOperator, queryModel, index);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


maumar:
@anpete

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -242,6 +242,135 @@ public override void Count_with_predicate()
...
+FROM [Orders] AS [o]",
+                Sql);
+        }
+
+        public override void OrderBy_Count_with_predicate_client_eval_mixed()
+        {
+            base.OrderBy_Count_with_predicate_client_eval_mixed();
+
+            Assert.Equal(
+                @"TBD",
```
maumar: fixed

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -242,6 +242,135 @@ public override void Count_with_predicate()
...
+                Sql);
+        }
+
+        public override void OrderBy_Where_Count_with_predicate_client_eval()
+        {
+            base.OrderBy_Where_Count_with_predicate_client_eval();
+
+            Assert.Equal(
+                @"TBD",
```
maumar: fixed

src/EntityFramework.Core/Query/QueryOptimizer.cs
```diff
@@ -103,5 +104,19 @@ in _queryAnnotations
                 queryAnnotation.QuerySource = fromClause;
             }
         }
+
+        public override void VisitResultOperator(ResultOperatorBase resultOperator, QueryModel queryModel, int index)
+        {
+            if (resultOperator is AnyResultOperator || resultOperator is CountResultOperator || resultOperator is LongCountResultOperator)
```
anpete: I think we are missing some result operators: I think the test can just be resultOperator is ValueFromSequenceResultOperatorBase && !(resultOperator is ChoiceResultOperatorBase).

src/EntityFramework.Core/Query/QueryOptimizer.cs
```diff
@@ -103,5 +104,19 @@ in _queryAnnotations
...
+            {
+                var orderByClauses = queryModel.BodyClauses.OfType<OrderByClause>().ToList();
```
anpete: Think you can do this more efficiently with a reverse for loop.

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -1872,6 +1872,106 @@ public virtual void Count_with_predicate()
         }
```
anpete: Can you add this test:

```
// Base
[Fact]
public virtual void Take_Count_OrderBy()
{
    AssertQuery<Order>(
           os => os.OrderBy(o => o.OrderID).Take(5).Count());
}

// SQLServer

        public override void Take_Count_OrderBy()
        {
            base.Take_Count_OrderBy();
            
            Assert.Equal(
                @"SELECT COUNT(*)
FROM (
    SELECT TOP(5) [o].[CustomerID], [o].[OrderDate], [o].[OrderID]
    FROM [Orders] AS [o]
    ORDER BY [o].[OrderID]
) AS [t0]",
                Sql);
        }
```

Issue #830, opened 2014-10-08
Reverse engineer model from database


rowanmiller:
Looks like we didn't have a work item tracking this. We need tooling to reverse engineer a model from a database. For the initial cut this should include the functionality from EF6.x (ability to select tables etc.).

This does not include the ability to incrementally update the model based on changes to the database, https://github.com/aspnet/EntityFramework/issues/831 tracks that.

lajones:
See #2465. 

From @ErikEJ: Is this a good time to start investigating using Reverse Engineering with SQLCE, or is it too early?

From @divega: @ErikEJ I personally believe it is a good time for you to get started with it. If you hit issues or have any feedback we can make adjustments.

From ErikEJ: @divega Sounds great, is there some docs/design specs? Or a drawing? :smile:

 I have a few questions:
 1: Does this info still apply: https://stackoverflow.com/questions/29300777/is-there-an-entity-framework-7-database-first-poco-generator? 
 2: If so, how does the command know which provider to use?
 3: Is the rev-eng package simply installed as needed? (EntityFramework.SqlServer.Design)
 4: Will I "simply" implement the equivalent of EntityFramework.SqlServer.Design ?

lajones:
@ErikEJ @divega 

1. The commands listed in https://stackoverflow.com/questions/29300777/is-there-an-entity-framework-7-database-first-poco-generator are still accurate.
2. That's the problem - at the moment it's hard-coded to use EntityFramework.SqlServer.Design. @rowanmiller are you ok if I add back in the --provider flag to pass in the name of the RevEng provider assembly? Also if the argument is not provided do you want to a) error or b) default to the existing value?
3. At the moment you must reference EntityFramework.SqlServer.Design in your project.json and hence it is installed when package restore is run on the project. For your new provider your new package should be referenced in project.json instead.
4. Yes. Hopefully simply :-). You must have a class that implements IDatabaseMetadataModelProvider. I'll write this up but in the meantime details below.

A) On IDatabaseMetadataModelProvider, GenerateMetadataModel(string connectionString) generates an IModel containing EntityTypes. DbContextTemplate() and EntityTypeTemplate() return the relevant templates as strings (you get to look them up however you want - however note #2418 where we'll be providing a way for customers to override the default templates).

B) First the DbContextTemplate template is executed being passed an instance of DbContextGeneratorModel. On DbContextGeneratorModel  you'll find a property MetadataModel - which is the model you generated in step A) above.

[...]

PR #1646, opened 2015-02-18
Changing CSharpCodeGeneratorHelper to be a service which is passed in.

lajones:
See issue #830. Updating CSharpCodeGeneratorHelper to no longer be a singleton and instead be a service which is passed in.

### DIFF
--- a/src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
+++ b/src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
@@ -9,6 +9,7 @@ using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Relational.Design.CodeGeneration;
 using Microsoft.Data.Entity.Utilities;
 using Microsoft.Framework.Logging;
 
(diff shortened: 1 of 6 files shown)


src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
```diff
@@ -22,15 +23,24 @@ public class ReverseEngineeringGenerator
...
             if (Logger == null)
             {
                 throw new ArgumentException(typeof(ReverseEngineeringGenerator).Name + " cannot find a service of type " + typeof(ILogger).Name);
             }
+
+            CSharpCodeGeneratorHelper = (CSharpCodeGeneratorHelper)serviceProvider.GetService(typeof(CSharpCodeGeneratorHelper));
+            if (CSharpCodeGeneratorHelper == null)
+            {
+                throw new ArgumentException(typeof(ReverseEngineeringGenerator).Name + " cannot find a service of type " + typeof(CSharpCodeGeneratorHelper).Name);
+            }
```
bricelam: There's an extension method for doing this--`GetRequiredService<T>()`.

bricelam:
:shipit:

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
lajones: Will do. Thanks.

lajones:
Updated and checked in - commit d5f5629.

PR #1647, opened 2015-02-18
API docs for ModelBuilder.cs

rowanmiller:


rowanmiller:
BTW the diff won't show up online because it is too large (there is a lot of API in this file). Need to pull changes locally or use https://github.com/aspnet/EntityFramework/pull/1647.patch (it's all additive comments so the lack of syntax highlighting isn't horrible).

ajcvickers:
I wonder if this constructor is a pit of failure:
```
        /// <summary>
        ///     Initializes a new instance of the <see cref="ModelBuilder" /> class with an empty model.
        ///     The builder will not use any conventions and the entire model must be explicitly configured. 
        ///     To specify conventions to be used, use a constructor that accepts a <see cref="ConventionSet"/>.
        /// </summary>
        public ModelBuilder()
            : this(new Model())
        {
        }
```
especially since in the old stack it would give you a fully functional, convention-based model builder.

ajcvickers:
This is really an API review comment (as was the last one) but I'm concerned about the use of "internal" here:
```
protected internal ModelBuilder([NotNull] InternalModelBuilder internalBuilder)
```

ajcvickers:
"Returns an object that can be used to configure a given entity in the model." Entity _type_?

ajcvickers:
"// TODO Add info about value generators when we have API sorted out for adding/customizing them" I think this will more often be used without also setting a value generator explicitly because the provider will determine the correct generator to use based on the type of the property and what it supports. I think it would be worth documenting this.

ajcvickers:
UseStoreDefault:

Configures whether a default value generated by the store is used for this property when the entity is saved and no value has been set.

Practically speaking, this means that if you create a new entity, don't set a value for this property, and save (insert) that entity, then the column default will be used and propagated back to the entity. If you do set a value, then the value set will be saved.

ajcvickers:
Not sure if I understand this:

```
If multiple reference key properties are specified, the order of reference key properties should match the order that the primary key or unique index properties were configured on the principal entity.
```

It would be unusual to use this call to point to a primary key because we do that by default. It would also be unusual to explicitly specify a unique index because one will be created by convention once you use this call to specify an alternate principal key. Also, you could use the same properties as the principal key or any unique index but in a different order, in which case you would be creating a new alternate key/unique index different from the existing one. This would be unusual, but is perfectly valid.

ajcvickers:
"one-to-many or zeroOrOne-to-many relationship." I don't think we should use these terms--I had a discussion with Diego about this a while back. They imply a level of abstraction that doesn't really exist, especially when you get to 1:1 type relationships. EDM and the old state manager gave some illusion that they existed, but even that was largely smoke and mirrors. In reality, all that Required does it determine whether the FK is nullable or not. If we want to describe that in non-relational terms we could talk about whether the reference navigation can be null or not.

ajcvickers:
:shipit:

rowanmiller:
Regarding the 'pit of failure constructor' I think it is... I took a few notes on the API surface as I was going and that was one of them. I think we need to make it more obvious how to construct with the default set of conventions.

PR #1648, opened 2015-02-18
Make value generators share fewer features and allow scoped value generators

ajcvickers:
Conceptually a value generator is a singleton service that should persist across scopes. This is the way they were previously implemented. However, this makes it impossible for a value generators to use scoped services, such as the connection, unless those services are passed to the generator instance. This change allows scoped value generators to be created such that they can use scoped services. This allows a generator to be created that can use, for example, the connection. However, care must then be taken to ensure that the state of the generator persists across scopes (for which a cache is provided.)

In addition, two of the reusable features of value generators--the ability to cache by something other than a property, and the ability to use a pool of generators--have been removed. These are now built into the HiLo/Sequence generator, which was the only generator currently using these features. Other generators cannot be pooled and are always cached by property unless they also implement these services inside their own code.


src/EntityFramework.Core/ValueGeneration/HiLoValueGenerator.cs
```diff
@@ -1,97 +1,24 @@
...
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.ValueGeneration
 {
-    /// <summary>
-    ///     Acts as a <see cref="ValueGenerator" />  by requesting a block of values from the
-    ///     underlying data store and returning them one by one. Will ask the underlying
-    ///     data store for another block when the current block is exhausted.
-    /// </summary>
     public abstract class HiLoValueGenerator<TValue> : ValueGenerator<TValue>
```
AndriySvyryd: Why are the comments removed?

ajcvickers: At one point in the iteration the doc was wrong, so I removed it. (Since it's not documentation that has been "officially" written yet--it's just a developer comment.) I think things gravitated back a bit so it is probably a somewhat true statement again, so I can add it back if you want.

src/EntityFramework.Core/ValueGeneration/ValueGeneratorSelectorContract.cs
```diff
@@ -1,10 +1,13 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
+
 namespace Microsoft.Data.Entity.ValueGeneration
 {
-    public abstract class SimpleTemporaryValueGenerator<TValue> : SimpleValueGenerator<TValue>
+    public abstract class ValueGeneratorSelectorContract
```
AndriySvyryd: We should discuss the name in API review

ajcvickers: I'm sure we will. :-) Or maybe make it an interface.

AndriySvyryd:
:shipit:

PR #1651, opened 2015-02-19
Adding ApiConsistencyTest for *.Design projects

lajones:
Depends on PR#1646

Adding ApiConsistencyTest for both EntityFramework.Relational.Design and EntityFramework.SqlServer.Design and updated the code to match (virtual's, [NotNull]'s).

Plus removed a reference to the no-longer-existent EntityFramework.ReverseEngineering project in EntityFramework-2014.sln.


src/EntityFramework.Relational.Design/CodeGeneration/CSharpCodeGeneratorHelper.cs
```diff
@@ -5,26 +5,27 @@
...
 using System.Reflection;
+using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Relational.Design.CodeGeneration
 {
     public class CSharpCodeGeneratorHelper
     {
-        public virtual void SingleLineComment(string comment, IndentedStringBuilder sb)
+        public virtual void SingleLineComment([NotNull] string comment, [NotNull] IndentedStringBuilder sb)
```
bricelam: Should also add `Check.NotNull()` calls.

bricelam:
:shipit:

src/EntityFramework.Relational.Design/CodeGeneration/CSharpCodeGeneratorHelper.cs
lajones: Will do. Thanks.

lajones:
Updated and checked in - commit f4e5717.

PR #1656, opened 2015-02-19
API docs for Microsoft.Data.Entity.ChangeTracking.*

rowanmiller:


src/EntityFramework.Core/ChangeTracking/ChangeTracker.cs
```diff
@@ -12,6 +12,11 @@
 namespace Microsoft.Data.Entity.ChangeTracking
 {
     // This is the app-developer facing public API to the change tracker
+    /// <summary>
+    ///     Provides access to change tracking information and operations for entity instances the context is tracking.
```
ajcvickers: Probably not important, but entries can be obtained for entities that are not being tracked such that these APIs can be used to consistently access any entity and potentially start tracking it. This might seem pedantic, and it may not be necessary to put in the API docs, but it is something Diego pushed hard for because it can make code to change state (such as used in the TrackGraph methods) simpler to write,

ajcvickers: Although I guess since the Entry method is not on ChangeTracker, maybe this doesn't apply here. :-)

src/EntityFramework.Core/ChangeTracking/ChangeTracker.cs
```diff
@@ -45,8 +59,26 @@ protected ChangeTracker()
             _context = context;
         }
 
+        /// <summary>
+        ///     <para>
+        ///         Gets or sets a value indicating whether the <see cref="ChangeTracker.DetectChanges()" /> method is called automatically 
+        ///         by methods of <see cref="DbContext" /> and related classes.
+        ///     </para>
+        ///     <para>
+        ///         The default value is true. This ensures the context is aware of any changes to tracked entity instances before performing
```
ajcvickers: Practical consideration for GitHub reviews--at least in Chrome on my machine it is hard to read these long lines.

src/EntityFramework.Core/ChangeTracking/ChangeTracker.cs
```diff
@@ -71,12 +109,47 @@ private void TryDetectChanges()
...
+        ///         If an entity is discovered that is already tracked by the context, that entity is not processed (and it's
+        ///         navigation properties are not traversed).
+        ///     </para>
+        /// </summary>
+        /// <param name="rootEntity"> The entity to begin traversal from. </param>
+        /// <param name="callback"> 
+        ///     An action to configure the change tracking information for each entity. For the entity to begin being tracked,
+        ///     the <see cref="EntityEntry.State"/> must be set.
+        /// </param>
         public virtual void TrackGraph([NotNull] object rootEntity, [NotNull] Action<EntityEntry> callback)
```
ajcvickers: Consider a reference to the KeyValueEntityTracker, or if we change this, then whatever mechanism allows people to easily use the key value to determine Added or Unchanged/Modified state.

src/EntityFramework.Core/ChangeTracking/EntityEntry.cs
```diff
@@ -34,12 +64,27 @@ public virtual EntityState State
             }
         }
 
+        /// <summary>
+        ///     Gets the internal entry that is tracking information about this entity.
+        /// </summary>
         public virtual InternalEntityEntry InternalEntry { get; }
```
ajcvickers: Should this be the IAccessor pattern?

ajcvickers:
Good docs. :-) :shipit:

src/EntityFramework.Core/ChangeTracking/ChangeTracker.cs
rowanmiller: I'll add it to the docs on ChangeTracker.Entry... oops, I mean DbContext.Entry... would have expected Entry to be with all the other change tracking stuff :smile:

src/EntityFramework.Core/ChangeTracking/ChangeTracker.cs
rowanmiller: Yeah I have a note taken to discuss KeyValueEntityTracker. Personally I'm not sure how much value it adds given the logic is really simple to write inline... but we should discuss. I added a note to revisit the docs once we discuss.

src/EntityFramework.Core/ChangeTracking/EntityEntry.cs
rowanmiller: Yeah, that's what we decided in API review. Figured the docs can be kept inline as we change the surface.

PR #1658, opened 2015-02-19

maumar:
SELECT COUNT(*)
FROM [Customer] AS [c]
ORDER BY [c].[Id]

Fix is to remove any orderings when querying for count.

There was another small issue that has been fixed - when calculating whether client eval for filter is required, we would use a common variable for the entire query. This could cause issues if multiple filters were present in the query, some being client eval, some being server eval. Fix is to use a variable local to the particular filter operation, and |= the results with the global variable. CR: Andrew

### DIFF
--- a/src/EntityFramework.Core/Query/QueryOptimizer.cs
+++ b/src/EntityFramework.Core/Query/QueryOptimizer.cs
@@ -9,6 +9,7 @@ using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Clauses.ExpressionTreeVisitors;
+using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Transformations;
 
 namespace Microsoft.Data.Entity.Query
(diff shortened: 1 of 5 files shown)


maumar:
@anpete

src/EntityFramework.Core/Query/QueryOptimizer.cs
```diff
@@ -103,5 +104,22 @@ in _queryAnnotations
                 queryAnnotation.QuerySource = fromClause;
             }
         }
+
+        public override void VisitResultOperator(ResultOperatorBase resultOperator, QueryModel queryModel, int index)
+        {
+            var containsPagingOperator = queryModel.ResultOperators.Any(r => r is TakeResultOperator) || queryModel.ResultOperators.Any(r => r is SkipResultOperator);
```
anpete: Break long lines. Also you don't need to execute this line until _after_ you match the resultOperator in the line below.

anpete:
:shipit:

maumar:
merged in 5ca3664a766edd2d47c04efbe03d7d1d78c12a83 and 1f8938316864038a69d9ec1b28a239f5dd28799e

Issue #1271, opened 2014-12-11
Make all key properties have the GenerateValueOnAdd flag set by convention


ajcvickers:
We are making a change so that the default behavior for properties that are part if keys is to have value generation enabled. Several points:
* Even with this flag set values will only be generated if the value of the property has not been set (is the default) when the entity is added. This means that use of non-generated keys where values have been set should not be affected.
* There is chance of collision of real key space and temporary key space, but we think this is unlikely to be a problem in the wild. Where collisions might occur they can be avoided by always setting key values explicitly so value generation never kicks in or by configuring different value generators that do not collide.
* Always generating keys provides a more flexible experience for dealing with disconnected graphs since temporary values will be used which can then be changed before saving
* An exception will be thrown if key generation is needed but no key generator is configured This can be avoided by setting the values explicitly before add or adding a new value generator.
* An exception will be thrown if a temporary key value is still set when the state of the object is changed to Unchanged, Modified, or Deleted. This will happen if SaveChanges is given temp values and doesn't generate real values.

ajcvickers:
A couple of notes: 
- This needs to be done for any property that is a principal key, not only for primary keys.
- We could consider changeing Property.DefaultGenerateValueOnAdd to:
```
protected virtual bool DefaultGenerateValueOnAdd => this.IsKey();
```
Or even to something more complicated that determines if this is the root principal in the chain. This is similar to what we do for IsReadOnly. However, it could be making the core metadata too magic/convention-based and hence we may not want to do it this way, and we may want to consider not doing IsReadOnly this way as well.

ajcvickers:
Another note. Currently is a property has GenerateValueOnAdd set and is not marked as using a SQL Server sequence, then it will be assumed to be SQL Server Identity. This doesn't work because properties such as alternate principal keys need to have GenerateValueOnAdd set but should not be treated as SQL Server Identity by either Migrations or the update pipeline.

It seems like there should be a SQL Server-specific convention which determines which column, if any, should be an Identity column and annotate appropriately. Migrations should do nothing other than honor the annotation. The value generator selector should use a temporary generator for any property that is not marked as using a sequence. The update pipeline should do nothing for columns that are not annotated as Identity. If temp values still exist in columns other than the Identity column then the existing exception for this case is correct.

smitpatel:
fixed via 5201798fe95d94e4c66f79f58ad4ad62a3ca878a

PR #1659, opened 2015-02-19
Set GenerateValueOnAdd flag for all Key properties by convention

### LINKED ISSUE #1126: SqlServer: Autogenerated integer values for composite PK
In present implementation SqlServer auto generates values for integer columns even if its a part of composite primary key which is not desirable by design.

smitpatel:
Implements #1271 
Added convention to set GenerateValueOnAdd flag as true whenever new key is added.
Flag is not set for Dependent end of Foreign Key
Resolves #1126 
Since above enhancement sets GenerateValueOnAdd flag as true for multiple properties in composite PK,
added sql server specific convention to add annotation to the property which should be used as Identity column.

### DIFF
--- a/src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
+++ b/src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
@@ -1,0 +1,62 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Metadata.ModelConventions;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.SqlServer.Metadata.ModelConventions
+{
+    public class SqlServerKeyConvention : IKeyConvention
+    {
+        private const string IdentityKeySuffix = "Id";
+        public virtual InternalKeyBuilder Apply(InternalKeyBuilder keyBuilder)
+        {
+            Check.NotNull(keyBuilder, "keyBuilder");
+
+            var key = keyBuilder.Metadata;
+            var properties = key.Properties;
+
+            if (key == key.EntityType.TryGetPrimaryKey())
+            {
+                var identityProperty = DiscoverIdentityProperty(properties);
+                if (identityProperty != null)
+                {
+                    var entityBuilder = keyBuilder.ModelBuilder.Entity(identityProperty.EntityType.Name, ConfigurationSource.Convention);
+                    ConfigureIdentityProperty(entityBuilder.Property(identityProperty.PropertyType, identityProperty.Name, ConfigurationSource.Convention));
+                }
+            }
+            return keyBuilder;
+        }
+
+        protected virtual IProperty DiscoverIdentityProperty(IReadOnlyList<Property> properties)
+        {
+            var identityProperty = properties
+                .Where(p => p.PropertyType.IsInteger())
+                .FirstOrDefault(p => string.Equals(p.Name, IdentityKeySuffix, StringComparison.OrdinalIgnoreCase));
+
+            if (identityProperty == null)
+            {
+                var entityType = properties.First().EntityType;
+
+                identityProperty = properties
+                    .Where(p => p.PropertyType.IsInteger())
+                    .FirstOrDefault(p => string.Equals(p.Name, entityType.SimpleName + IdentityKeySuffix, StringComparison.OrdinalIgnoreCase));
+            }
+
+            return identityProperty ?? properties.FirstOrDefault(p => p.PropertyType.IsInteger());
+        }
+
+        protected virtual void ConfigureIdentityProperty([NotNull] InternalPropertyBuilder propertyBuilder)
+        {
+            Check.NotNull(propertyBuilder, "propertyBuilder");
+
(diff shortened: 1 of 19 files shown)


ajcvickers:
@smitpatel Please have @AndriySvyryd review this for the way it uses conventions.

test/EntityFramework.Core.Tests/ChangeTracking/Internal/ChangeDetectorTest.cs
```diff
@@ -1802,20 +1802,25 @@ private static IModel BuildModel()
...
-            builder.Entity<Product>()
-                .HasOne(e => e.Tag).WithOne(e => e.Product)
-                .ReferencedKey<Product>(e => e.TagId)
-                .ForeignKey<ProductTag>(e => e.ProductId);
+            builder.Entity<Product>(b =>
+                {
+                    b.HasOne(e => e.Tag).WithOne(e => e.Product)
+                        .ReferencedKey<Product>(e => e.TagId)
+                        .ForeignKey<ProductTag>(e => e.ProductId);
+                    b.Property(e => e.TagId).GenerateValueOnAdd(false);
```
ajcvickers: Can you explain why these tests needed to be changed?

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
```diff
@@ -0,0 +1,62 @@
...
+                if (identityProperty != null)
+                {
+                    var entityBuilder = keyBuilder.ModelBuilder.Entity(identityProperty.EntityType.Name, ConfigurationSource.Convention);
+                    ConfigureIdentityProperty(entityBuilder.Property(identityProperty.PropertyType, identityProperty.Name, ConfigurationSource.Convention));
+                }
+            }
+            return keyBuilder;
+        }
+
+        protected virtual IProperty DiscoverIdentityProperty(IReadOnlyList<Property> properties)
```
ajcvickers: I'm not sure we should be setting Identity by convention for one property in a composite key. Would like to hear what @divega thinks.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            return identityProperty ?? properties.FirstOrDefault(p => p.PropertyType.IsInteger());
+        }
+
+        protected virtual void ConfigureIdentityProperty([NotNull] InternalPropertyBuilder propertyBuilder)
+        {
+            Check.NotNull(propertyBuilder, "propertyBuilder");
+
+            propertyBuilder.Annotation(SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.ValueGeneration, SqlServerValueGenerationStrategy.Identity.ToString(), ConfigurationSource.Convention);
```
ajcvickers: Is there some reason that the sugar API can't be used here? If so we should consider if the sugar API needs to be changed.

src/EntityFramework.SqlServer/Migrations/SqlServerModelDiffer.cs
```diff
@@ -123,11 +123,10 @@ protected override IEnumerable<MigrationOperation> Add(IProperty target)
...
         private SqlServerValueGenerationStrategy? GetValueGenerationStrategy(IProperty property) =>
-            property.SqlServer().ValueGenerationStrategy
-            ?? property.EntityType.Model.SqlServer().ValueGenerationStrategy
-            ?? (property.GenerateValueOnAdd && property.PropertyType.IsInteger() && property.IsPrimaryKey()
-                ? SqlServerValueGenerationStrategy.Identity
-                : default(SqlServerValueGenerationStrategy?));
+            property.GenerateValueOnAdd
+            ? (property.SqlServer().ValueGenerationStrategy
+                ?? property.EntityType.Model.SqlServer().ValueGenerationStrategy)
+            : default(SqlServerValueGenerationStrategy?);
```
ajcvickers: Won't this mean that if I set the value generation strategy on the model, then any property marked as GenerateValueOnAdd will becomes an Identity column? If so, I don't think this works for the same reason that the previous convention didn't work--multiple properties will become Identity columns.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        private const string IdentityKeySuffix = "Id";
+        public virtual InternalKeyBuilder Apply(InternalKeyBuilder keyBuilder)
+        {
+            Check.NotNull(keyBuilder, "keyBuilder");
+
+            var key = keyBuilder.Metadata;
+            var properties = key.Properties;
+
+            if (key == key.EntityType.TryGetPrimaryKey())
```
ajcvickers: Maybe we need an IPrimaryKeyConvention?

test/EntityFramework.Core.Tests/Metadata/ModelConventions/KeyDiscoveryConventionTest.cs
```diff
@@ -0,0 +1,131 @@
...
+using System.Linq;
+using Microsoft.Data.Entity.Internal;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Metadata.ModelConventions;
+using Moq;
+using Moq.Protected;
+using Xunit;
+
+namespace Microsoft.Data.Entity.Tests.Metadata.ModelConventions
```
ajcvickers: I don't see tests for checking that GenerateValueOnAdd is set for principal keys (primary or alternate) except where the properties are also foreign keys. Including more than one link in the chain to the root principal. Is this already covered somewhere by existing tests?

Also, we should have tests that the cases involving cycles that @mikary has been looking at are also handled appropriately.

smitpatel:
@ajcvickers - I had created this PR to get initial feedback from @AndriySvyryd on conventions related changes. Sorry I forgot to mention that. This is still WIP.
Since I have got feedback, I am closing this for now. Once done I will create new PR.

Issue #1642, opened 2015-02-17
Query :: Bad sql generated for query with bool parameter in filter


maumar:
Repro:

```
    public class Customer
    {
        public int Id { get; set; }
    }

    public class Order
    {
        public int Id { get; set; }
        public int CustomerId { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Customer> Customers { get; set; }
        public DbSet<Order> Orders { get; set; }

        protected override void OnConfiguring(DbContextOptions options)
        {
            options.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "Repro1642", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();
            }

            using (var ctx = new MyContext())
            {
                ctx.Customers.OrderBy(o => ClientEvalSelectorStateless()).Where(o => ClientEvalPredicateStateless()).Count(c => ClientEvalPredicate(c));
            }
        }

        public static bool ClientEvalPredicateStateless()
        {
            return true;
        }

        protected static bool ClientEvalPredicate(Customer customer)
        {
            return customer.Id > 1;
        }

        private static int ClientEvalSelectorStateless()
        {
            return 42;
        }
    }
```

Exception:

Unhandled Exception: System.Data.SqlClient.SqlException: An expression of non-bo
olean type specified in a context where a condition is expected, near '@__Client
EvalPredicateStateless_1'.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolea
n breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception
, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObj
ect stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand
 cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler,
TdsParserStateObject stateObj, Boolean& dataReady)
[...]

PR #1660, opened 2015-02-20
fix to #1642 -Query :: Bad sql generated for query with bool parameter in filter

maumar:
Error is for queries that are using bool parameter (closure) in a filter query. We would produce query like: SELECT c.Id FROM Customers as c WHERE @prm, which is invalid.

Fix is to apply similar compensation as we do for constant bools in filter, to produce query like: SELECT c.Id FROM Customers as c WHERE @prm = 1

### DIFF
--- a/test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
+++ b/test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
@@ -258,6 +258,17 @@ namespace Microsoft.Data.Entity.FunctionalTests
         }
 
         [Fact]
+        public virtual void Where_simple_closure_constant()
+        {
+            // ReSharper disable once ConvertToConstant.Local
+            var predicate = true;
+
+            AssertQuery<Customer>(
+                cs => cs.Where(c => predicate),
+                entryCount: 91);
+        }
+
+        [Fact]
         public virtual void Where_simple_closure_via_query_cache()
         {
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -1946,14 +1957,14 @@ public virtual void OrderBy_Count_with_predicate_client_eval()
         }
 
         // error #1642
```
maumar: removed

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -1946,14 +1957,14 @@ public virtual void OrderBy_Count_with_predicate_client_eval()
...
-        //[Fact]
+        [Fact]
         public virtual void OrderBy_Count_with_predicate_client_eval_mixed()
         {
             AssertQuery<Order>(os => os.OrderBy(o => o.OrderID).Count(o => ClientEvalPredicateStateless()));
         }
 
         // error #1642
```
maumar: removed

anpete:
:shipit:

maumar:
merged in ef46f6c295d50ac24c07a84e877ab76a42bfd8dc

PR #1661, opened 2015-02-20
Improve TopologicalSort error message in CommandBatchPreparer

mikary:
Fix for issue #1453, attached ForeignKey data to the ```CommandBatchPreparer``` topological sort to indicate relationships causing circular dependencies.
Moved Batching version of topological sort algorithm from ```BidirectionalAdjacencyListGraph``` to ```Multigraph``` and ported tests.


AndriySvyryd:
Add a test where batching fails due to a cycle to verify the actual exception message.
Remove BidirectionalAdjacencyListGraph since it doesn't appear to be used anywhere.
:shipit:

src/EntityFramework.Core/Utilities/Multigraph.cs
```diff
@@ -231,6 +231,120 @@ public virtual IReadOnlyList<TVertex> TopologicalSort()
...
+
+        public virtual IReadOnlyList<List<TVertex>> BatchingTopologicalSort(
+            [CanBeNull] Func<IEnumerable<Tuple<TVertex, TVertex, IEnumerable<TEdge>>>, string> formatCycle)
+        {
+            var currentRootsQueue = new List<TVertex>();
+            var predecessorCounts = new Dictionary<TVertex, int>();
+
+            foreach (var vertex in _vertices)
+            {
+                var count = GetIncomingNeighbours(vertex).Count();
```
AndriySvyryd: It would be ideal if you only used GetOutgoingNeighbours when populating predecessorCounts since it's a O(1) operation, but this is not necessary until we get perf numbers.

mikary: I would also like to factor some reusable parts of the two topological sort algorithms together to make this more DRY, but it's not to urgent yet.

PR #1662, opened 2015-02-20
Remove non-cancellation token overloads of QueryableExtensions

rowanmiller:
Make CancellationToken an optional param on remaining overloads


anpete:
:shipit:

PR #1680, opened 2015-02-20
Metadata: Inheritance: Adds more fundamental inheritance support.

anpete:
- EntityTypes can now be abstract.
- Fixed bug in entity type index maintenance.
- ForeignKey now maintains referenced entity type independently of referenced key so we can handle FKs to derived types.
- Added EntityLoadInfo to query pipeline as a way to track which materializer is associated with a given value reader.
- Modified InMemoryDatabase to maintain a table for each concrete entity type.
- Lots of C# 6 goodness.

### DIFF
--- a/src/EntityFramework.Core/Metadata/EntityMaterializerSource.cs
+++ b/src/EntityFramework.Core/Metadata/EntityMaterializerSource.cs
@@ -51,12 +51,17 @@ namespace Microsoft.Data.Entity.Metadata
             var materializer = entityType as IEntityMaterializer;
             if (materializer != null)
             {
-                return materializer.CreatEntity;
+                return materializer.CreateEntity;
             }
 
             if (!entityType.HasClrType)
             {
                 throw new InvalidOperationException(Strings.NoClrType(entityType.Name));
+            }
+
+            if (entityType.IsAbstract)
+            {
+                throw new InvalidOperationException(Strings.CannotMaterializeAbstractType(entityType));
             }
 
             return _cache.GetOrAdd(entityType.Type, k => BuildDelegate(entityType));
--- a/src/EntityFramework.Core/Metadata/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/EntityType.cs
@@ -39,8 +39,10 @@ namespace Microsoft.Data.Entity.Metadata
 
         private readonly object _typeOrName;
 
-        private EntityType _baseType;
         private Key _primaryKey;
+        private EntityType _baseType;
+
+        public event EventHandler<Property> PropertyMetadataChanged;
 
         private int _shadowPropertyCount;
         private int _originalValueCount;
@@ -61,14 +63,11 @@ namespace Microsoft.Data.Entity.Metadata
         /// <param name="type">The .NET entity type that this metadata object represents.</param>
         /// <param name="model">The model associated with this entity type.</param>
         public EntityType([NotNull] Type type, [NotNull] Model model)
+            : this(
+                (object)Check.NotNull(type, nameof(type)),
+                Check.NotNull(model, nameof(model)))
         {
-            Check.NotNull(type, nameof(type));
-            Check.NotNull(model, nameof(model));
-            Check.ValidEntityType(type, "type");
-
-            _typeOrName = type;
-
-            Model = model;
+            Check.ValidEntityType(type, nameof(type));
 
             _useEagerSnapshots = !this.HasPropertyChangingNotifications();
         }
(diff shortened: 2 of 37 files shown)


src/EntityFramework.Core/Metadata/EntityMaterializerSource.cs
```diff
@@ -59,6 +59,11 @@ public EntityMaterializerSource([NotNull] MemberMapper memberMapper)
                 throw new InvalidOperationException(Strings.NoClrType(entityType.Name));
             }
 
+            if (entityType.IsAbstract)
+            {
+                throw new InvalidOperationException(Strings.CantMaterializeAbstractType(entityType));
```
AndriySvyryd: nit:" Cannot" instead of "Cant"

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -80,13 +80,19 @@ public EntityType([NotNull] Type type, [NotNull] Model model)
...
+        }
 
-            _typeOrName = name;
+        private EntityType(object typeOrName, Model model)
+        {
+            _typeOrName = typeOrName;
 
             Model = model;
+
+            _hasDerivedTypes = new LazyRef<bool>(() => GetDerivedTypes().Any());
```
AndriySvyryd: don't store this until perf tests show it's on a hotpath

src/EntityFramework.Core/Properties/Strings.resx
```diff
@@ -453,4 +453,10 @@
   <data name="GraphDoesNotContainVertex" xml:space="preserve">
     <value>The edge cannot be added because the graph does not contain vertex '{vertex}'.</value>
   </data>
+  <data name="CantMaterializeAbstractType" xml:space="preserve">
+    <value>Unable to build a materializer for entity type '{entityType}' because it is abstract. Either make it non-abstract or consider mapping at least one derived type.</value>
```
AndriySvyryd: You can avoid talking about a materializer and just say that instances of an abstract entity type cannot be created

src/EntityFramework.Core/Query/EntityLoadInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
+
+namespace Microsoft.Data.Entity.Query
+{
+    public struct EntityLoadInfo
```
AndriySvyryd: What's the advantage of this?
Just grouping two references doesn't justify a new type IMO.

src/EntityFramework.Core/Query/QueryBuffer.cs
```diff
@@ -267,28 +266,29 @@ var entityKeyFactory
...
-                            currentNavigationIndex + 1,
-                            querySourceRequiresTracking)
-                            .WithCurrentCulture();
-
-                        return targetEntity;
-                    })
+                            return targetEntity;
+                        })
                     .Where(e => e != null)
                     .ToList(cancellationToken)
```
AndriySvyryd: Should be called ToListAsync

src/EntityFramework.Core/Query/QueryBuffer.cs
```diff
@@ -267,28 +266,29 @@ var entityKeyFactory
...
-                        var targetEntity
-                            = GetTargetEntity(
-                                targetEntityType,
-                                entityKeyFactory,
-                                keyProperties,
-                                valueReader,
-                                bufferedEntities,
-                                querySourceRequiresTracking);
+                await relatedEntitiesLoaders[currentNavigationIndex](primaryKey, relatedKeyFactory)
+                    .Select(async (eli, ct) =>
```
AndriySvyryd: This is so hard to read. Use a foreach.

src/EntityFramework.Core/Query/EntityLoadInfo.cs
anpete: Interesting, are you saying that Tuple<,> or KeyValuePair should not exist?

In this case, I have methods that return these and I prefer this over Tuple.

src/EntityFramework.Core/Metadata/EntityMaterializerSource.cs
anpete: Will change.

src/EntityFramework.Core/Metadata/EntityType.cs
anpete: Will change.

src/EntityFramework.Core/Properties/Strings.resx
anpete: Will change - these are pending our favorite review anyway...

src/EntityFramework.Core/Query/QueryBuffer.cs
anpete: The Ix-Async operators don't have the suffix.

src/EntityFramework.Core/Query/EntityLoadInfo.cs
AndriySvyryd: I see, it's ok then.

src/EntityFramework.Core/Query/QueryBuffer.cs
anpete: Subjective.

AndriySvyryd: The final call is nested 4 times and two of the calls are async, it is objectively complex for a non-fluent call.

AndriySvyryd:
:shipit:

PR #1681, opened 2015-02-20
[TINY] Make MethodInfo's exposed from QueryableExtensions internal

rowanmiller:
Cleaning up the public API on this class (these ones are not private because they are used in the expression nodes)


anpete:
:shipit:

PR #1682, opened 2015-02-21
Use project.json for .csproj package references via kpm list

bricelam:
This gets rid of our silly `<PackageReference>` elements and makes `project.json` the source of truth for both csproj and kproj-based solutions. It also enables us to use some of the newer contracts like `System.IO.FileSystem` from the csproj-based solution.

Unfortunately, there is a bug in .NET Framework 4.6 that prevents the tests from running in the csproj-based solution. We probably want to wait until the fix is released publicly before merging.


tools/EntityFramework.targets
```diff
@@ -2,83 +2,49 @@
...
-        Directory.CreateDirectory(Path.GetDirectoryName(TargetsFile));
         using (var writer = new StreamWriter(TargetsFile, append: false))
         {
             writer.WriteLine("<Project xmlns='http://schemas.microsoft.com/developer/msbuild/2003'>");
             writer.WriteLine("  <ItemGroup>");
 
-            var installedPackages = Directory.EnumerateDirectories(packagesFolder).Select(d => Path.GetFileName(d))
-                .ToArray();
-            foreach (var package in PackageReferences)
+            foreach (var assemblyPath in File.ReadLines(PackageReferences).Where(File.Exists))
```
bricelam: @troydai and I decided that using `File.Exists()` is a good contract not to break for the output of `kpm list --assemblies`.

troydai: :+1:

bricelam:
This depends on aspnet/Universe#167.

src/EntityFramework.Commands/EntityFramework.Commands.csproj
```diff
@@ -10,15 +11,18 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.Data.Entity.Commands</RootNamespace>
     <AssemblyName>EntityFramework.Commands</AssemblyName>
-    <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
+    <DefaultLanguage>en-US</DefaultLanguage>
     <FileAlignment>512</FileAlignment>
+    <ProjectTypeGuids>{786C830F-07A1-408B-BD7F-6EE04809D6DB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
+    <TargetFrameworkProfile>Profile7</TargetFrameworkProfile>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
bricelam: Note to self, we should leave this as net451 so we get test coverage on that framework.

MrTomWhite:
:+1: This is an awesome fix.

bricelam:
Closing for now. Will re-open when unblocked

bricelam:
New approach: Same thing, but net45 projects instead of .NET Core 5

PR #1686, opened 2015-02-23
Update xUnit.net to 2.0-rc3

bricelam:
This depends on aspnet/aspnet.xunit#2


lajones:
Looks good :shipit: Presume we should do the same for the 2 .csproj files I'm about to check in for the *.Design.Tests as well (https://github.com/aspnet/EntityFramework/pull/1678)?

bricelam:
@lajones Eventually, yes. It may be a while before this gets merge though. :smile: xUnit.net 2.0-rc3 was just released today and we'll need to coordinate updating all of the other repos.

divega:
:shipit:

PR #1687, opened 2015-02-23
API docs for Microsoft.Data.Entity.Extensions.*

rowanmiller:
Docs in QueryableExtensions are copied from EF6.x except for the following (which need a detailed review):
- Include
- AsAsyncEnumerable
- AsNoTracking
- Load

### DIFF
--- a/src/EntityFramework.Core/Extensions/EntityServiceCollectionExtensions.cs
+++ b/src/EntityFramework.Core/Extensions/EntityServiceCollectionExtensions.cs
@@ -6,12 +6,12 @@ using JetBrains.Annotations;
 using Microsoft.Data.Entity;
 using Microsoft.Data.Entity.ChangeTracking;
 using Microsoft.Data.Entity.ChangeTracking.Internal;
-using Microsoft.Data.Entity.ValueGeneration;
 using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Query;
 using Microsoft.Data.Entity.Storage;
 using Microsoft.Data.Entity.Utilities;
+using Microsoft.Data.Entity.ValueGeneration;
 using Microsoft.Framework.Cache.Memory;
 using Microsoft.Framework.ConfigurationModel;
 using Microsoft.Framework.Logging;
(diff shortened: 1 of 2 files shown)


src/EntityFramework.Core/Extensions/EntityServiceCollectionExtensions.cs
```diff
@@ -21,10 +21,47 @@
...
+        ///         Adds the services required by the core of Entity Framework to an <see cref="IServiceCollection" />.
+        ///         You only need to use this functionality when you want Entity Framework to resolve the services it uses
+        ///         from an external <see cref="IServiceProvider" />. If you are not using an external
+        ///         <see cref="IServiceProvider" /> Entity Framework will take care of creating the services it requires.
+        ///     </para>
+        ///     <para>
+        ///         The data store you are using will also define extension methods that can be called on the returned
+        ///         <see cref="EntityFrameworkServicesBuilder" /> to register the services for the data store. For example,
+        ///         when using EntityFramework.SqlServer you would call 
+        ///         <c>collection.UseEntityFramework(config).UseSqlServer()</c>.
```
ajcvickers: Add, not use. :-)

src/EntityFramework.Core/Extensions/EntityServiceCollectionExtensions.cs
```diff
@@ -93,6 +130,35 @@ public static class EntityServiceCollectionExtensions
             return new EntityFrameworkServicesBuilder(serviceCollection, configuration);
         }
 
+        /// <summary>
+        ///     Registers the given context as a service in the <see cref="IServiceCollection" />. This will ensure services that
+        ///     the context uses are resolved from the <see cref="IServiceProvider" /> and any Entity Framework configuration 
+        ///     found in the configuration passed to <see cref="AddEntityFramework" /> will be honored.
```
ajcvickers: I don't know how to make this better, but if I read this I still wouldn't really know what AddDbContext does or when I should use it.

ajcvickers:
I wonder if it would be better to take a slightly different approach to documenting AddEntityFramework and AddDbContext. Rather than focusing on what these methods do, maybe the "summary" documentation should point to an example of how they are used, maybe even being specific to ASP.NET, which will be _by far_ the most common place these methods would be used. Then move the "what they actually do" documentation all into the remarks section for people who are attempting to drill down into the actual underlying functionality.

ajcvickers:
With regard to Include, given that ThenInclude can only be used after collections I think it is important to document the dotting syntax that is needed for including multiple levels of references.

ajcvickers:
With regard to, AsNoTracking, is it still true that we do identity resolution? If so, I think we need to discuss this with regard to the unbounded streaming query functionality.

rowanmiller:
* Opened an issue on the AsNoTracking topic.
* As disussed, I will remove docs from Include for the moment and will open an issue to consider our ideas for reducing the complexity.
* Will take a stab at your idea for DI APIs (I like it)

ajcvickers:
:shipit:

PR #1689, opened 2015-02-24
First set of Functional tests for EF.Relational.Design

lajones:
@divega @bricelam Would appreciate some feedback this approach to functional tests for RevEng.

Have tried to avoid Mocks (per Diego's comments) but found I needed to use them in one place (may be able to remove later by making the outputting of files into a service).

Tests are not by any means exhaustive. Will add more once the general approach is considered OK.

Note: changed the generator to receive the actual provider instance and leave the looking up of this provider in the assembly to the Commands code.


lajones:
Part of https://github.com/aspnet/EntityFramework/issues/1669

bricelam:
These still feel like unit tests to me; `TestDatabaseMetadataModelProvider` feels like a mock. I think for functional tests, it would good to hit a database and/or compile code. The rest of the coverage can probably be done in unit tests. The two big functional scenarios I see are:
* Create various database objects, verify SqlServerMetadataModelProvider outputs a correct model
* Create various models, generate code, verify code, compile code, verify model

divega:
@bricelam @lajones have you guys talked about how to move forward with this?

lajones:
@divega @bricelam Will do - wanted to do it F2F.

PR #1702, opened 2015-02-25
Standardize naming of IServiceCollection EF extension method class

rowanmiller:
Previously called EntityServiceCollectionExtensions but we don't use an EF prefix on other extension method classes (QueryableExtensions, TaskExtentions).
Duplicate type names won't matter for extension methods... unless you had a conflict on extension method signature... which is then going to cause an issue when you try to call it anyways


rowanmiller:
BTW the other option, if we are nervous about duplicate type names, is to add a prefix to all our extension method classes.

ajcvickers:
:shipit:

Eilon:
BTW remember that if someone is doing a web search based on the type name that they might have a much harder time finding proper search results. I'm a fan of having the extension method containing classes have long descriptive names because in general you never see them, but in the few times it matters, the long name can be helpful. Also consider cases where people are using Reflector, dotPeek, ILSpy, etc. trying to find a type - they'll get tons of matches.

rowanmiller:
Yeah search is a really good point and I think that warrants going with the prefix approach. Do folks think 'EntityFramework' is an ok prefix? @divega @ajcvickers @anpete @bricelam

divega:
Sounds good to me.

ajcvickers:
:ship: for changing all our extension method classes to use the prefix.

Eilon:
BTW the engineering guidelines discuss this a bit: https://github.com/aspnet/Home/wiki/Engineering-guidelines#extension-method-patterns

> The class name of an extension method container (also known as a "sponsor type") should generally follow the pattern of  `<Feature>Extensions` ,  `<Target><Feature>Extensions` , or  `<Feature><Target>Extensions` . For example:

Issue #1670, opened 2015-02-20
RevEng: Add DefaultValue fluent API into generated code


lajones:
Part of #830.

Also investigate what to do about DefaultExpressions.

lajones:
Done with the above 2 check-ins.

Issue #1672, opened 2015-02-20
RevEng: Make ModelUtilities a Service


lajones:
Part of #830.

Currently ModelUtilities is set up as a singleton. Change to be a service (makes it more testable).

lajones:
Consider doing the same for SqlServerLiteralUtilities (see #1710).

lajones:
Looked into SqlServerLiteralUtilities, but decided I'd rather not pass that in as a service as that would make Commands dependent on SqlServer.Design as well as Relational.Design. The SqlServerLiteralUtilities functionality is basically just a helper with SQL Server specific syntax which I'd like to keep within SqlServer.Design.

lajones:
Done with commit d2f7162.

PR #1710, opened 2015-02-26
Adding DefaultValue code generation

lajones:
See Issue #1670.

This deals with fixed default values only. There are also expressions which are returned in the metadata - e.g. newid() and getdate() - which I will address separately.

### DIFF
--- a/src/EntityFramework.Relational.Design/CodeGeneration/CSharpUtilities.cs
+++ b/src/EntityFramework.Relational.Design/CodeGeneration/CSharpUtilities.cs
@@ -157,6 +157,11 @@ namespace Microsoft.Data.Entity.Relational.Design.CodeGeneration
             return value.ToString(CultureInfo.InvariantCulture) + "f";
         }
 
+        public virtual string GenerateLiteral(double value)
+        {
+            return value.ToString(CultureInfo.InvariantCulture) + "D";
+        }
+
         public virtual string GenerateLiteral(TimeSpan value)
         {
             return "new TimeSpan(" + value.Ticks + ")";
@@ -193,13 +198,21 @@ namespace Microsoft.Data.Entity.Relational.Design.CodeGeneration
             return "@\"" + EscapeVerbatimString(value) + "\"";
         }
 
-        public virtual string GenerateLiteral([NotNull] object value)
+        public virtual string GenerateLiteralForUnknownType([NotNull] object value)
         {
             Check.NotNull(value, nameof(value));
 
             if (value.GetType().GetTypeInfo().IsEnum)
             {
                 return Enum.Format(value.GetType(), value, "D");
+            }
+
+            var generateLiteralMethodInfo =
+                typeof(CSharpUtilities)
+                    .GetRuntimeMethod("GenerateLiteral", new Type[] { value.GetType() });
+            if (generateLiteralMethodInfo != null)
+            {
+                return (string)generateLiteralMethodInfo.Invoke(this, new object[] { value });
             }
 
             return string.Format(CultureInfo.InvariantCulture, "{0}", value);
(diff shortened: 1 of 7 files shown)


src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGenerator.cs
```diff
@@ -465,6 +471,22 @@ public virtual string GenerateColumnTypeFacetConfiguration([NotNull] IProperty p
             return null;
         }
 
+        public virtual string GenerateDefaultValueFacetConfiguration([NotNull] IProperty property)
+        {
+            Check.NotNull(property, nameof(property));
+
+            // output columnType if decimal or datetime2 to define precision and scale
```
lajones: I will remove the comment - too much copy/paste.

src/EntityFramework.Relational.Design/CodeGeneration/CSharpUtilities.cs
```diff
@@ -202,6 +207,14 @@ public virtual string GenerateLiteral([NotNull] object value)
                 return Enum.Format(value.GetType(), value, "D");
             }
 
+            var generateLiteralMethodInfo =
+                typeof(CSharpUtilities)
+                    .GetRuntimeMethod("GenerateLiteral", new Type[] { value.GetType() });
+            if (generateLiteralMethodInfo != null)
+            {
+                return (string)generateLiteralMethodInfo.Invoke(this, new object[] { value });
+            }
```
bricelam: We use `GenerateLiteral((dynamic)value)` to do this in other places.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -57,6 +58,7 @@ public SqlServerMetadataModelProvider([NotNull] IServiceProvider serviceProvider
             Check.NotNull(serviceProvider, nameof(serviceProvider));
 
             _logger = serviceProvider.GetRequiredService<ILogger>();
+            _sqlServerLiteralUtilities = new SqlServerLiteralUtilities(_logger);
```
bricelam: Should this be a service?

src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                return ConvertSqlServerBitLiteral(sqlServerDefaultValue);
+            }
+
+            //TODO: decide what to do about byte[] default values and the values
+            // newid() and getdate() for Guid and DateTime respectively
+
+            var parseMethodInfo =
+                propertyType
+                    .GetRuntimeMethod("Parse", new Type[] { typeof(string) });
```
bricelam: It might be better to use the `Convert` class than to look for instance methods named *Parse*.

bricelam:
:shipit:

src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
divega: Apparently `Convert` isn't available everywhere. Do we need this code to work everywhere?

src/EntityFramework.Relational.Design/CodeGeneration/CSharpUtilities.cs
lajones: Will do. Thanks.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
lajones: I thought about it, but it seemed very specific to this particular provider and not something other provider writers would find useful to override. But I have a couple of other issues where I'm considering changing to a service - I've added considering this to #1672).

src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
lajones: I had to add a dependency on System.Runtime.Extensions. After that Convert was available everywhere. So have changed to use Convert.ChangeType().

lajones:
Checked in - commit 1647b7f.

PR #1712, opened 2015-02-26
[SMALL] Added in ability to interpret the DefaultValue as a DefaultExpression.

lajones:
Part of fix for issue #1670. Depends on #1710.

### DIFF
--- a/src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
+++ b/src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
+using System.Text.RegularExpressions;
 using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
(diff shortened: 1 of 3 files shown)


src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
```diff
@@ -137,4 +157,10 @@ public virtual string ConvertSqlServerStringLiteral([NotNull] string sqlServerSt
             return null;
         }
     }
+
+    public class DefaultExpressionOrValue
+    {
+        public virtual string DefaultExpression { get; [param: NotNull] set; }
+        public virtual object DefaultValue { get;[param: NotNull] set; }
+    }
```
bricelam: Is there a different way to factor this that doesn't require a union type?

bricelam:
:shipit:

src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
lajones: I'm open to suggestions. I did think about it because it seems ugly, but there is generic code which we want to apply to all SQL Server values (remove the surrounding parens and potentially reporting the failure to understand the value at all), then we want to determine whether that value is an expression or a value - so that the EF code knows which Fluent API to apply. Unfortunately these concepts are both stored in the same value in the database. So this code need to report one of 3 things: i) I don't understand this value at all, ii) I understand it as an expression and here's the expression, iii) I understand it as a value and here's the value. So this seemed the most pragmatic way of doing it. Another alternative was out params but that seemed even uglier.

lajones:
Checked in - commit 967bf2a

PR #1716, opened 2015-02-26
Allow direct constructor injection of dynamic services

ajcvickers:
This change removes the DbContextService class and instead registers DI functions that return dynamic services directly. This means that the services can be used directly in constructor arguments, and consequentially are resolved eagerly rather than lazily.

The most significant change from this is that DataStore could not longer depend on StateManager. This was resolved by doing some DI-enabling in query, since query was using DataStore to obtain most of its services. There is probably more that can be done to reduce DataStore coupling by DI-enabling query.

In addition to this change, the DataStoreServices implementations of providers are now implemented using ServiceLocator. This is a good change because DataStoreServices is now used only as the source for functions registered in DI that allow the constructor injection to work, and using service locator avoids coupling of all the data store services together in one place where they will all be loaded together.

Other changes were in tests where we were relying on lazy evaluation never happening to avoid having to have valid services. The eager evaluation means we always need a valid service.


anpete:
:shipit:

PR #1719, opened 2015-02-27
Make the first navigation specified when configuring self referencing relationship be the navigation to principal.

AndriySvyryd:
Throw when configuring self referencing relationship with both navigations using the same property.
Reformat ModelBuilder.

### DIFF
--- a/src/EntityFramework.Core/Properties/Strings.resx
+++ b/src/EntityFramework.Core/Properties/Strings.resx
@@ -459,4 +459,7 @@
   <data name="ForeignKeyReferencedEntityKeyMismatch" xml:space="preserve">
     <value>The provided referenced entity key '{referencedKey}' is not a key on the entity type '{referencedEntityType}'.</value>
   </data>
+  <data name="NavigationToSelfDuplicate" xml:space="preserve">
+    <value>The navigation property '{navigation}' cannot be used for both ends of the relationship. Either specify it on the principal side or the dependent side.</value>
+  </data>
 </root>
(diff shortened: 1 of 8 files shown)


src/EntityFramework.Core/Properties/Strings.resx
```diff
@@ -459,4 +459,7 @@
   <data name="ForeignKeyReferencedEntityKeyMismatch" xml:space="preserve">
     <value>The provided referenced entity key '{referencedKey}' is not a key on the entity type '{referencedEntityType}'.</value>
   </data>
+  <data name="NavigationToSelfDuplicate" xml:space="preserve">
+    <value>The navigation property '{navigation}' cannot be used for both ends of the relationship. Either specify it on the principal side or the dependent side.</value>
```
ajcvickers: I wonder if this is helpful enough given that this is something that people often try to do? But then I'm not sure what to say that doesn't get into principal/dependent terminology that will be confusing. Maybe @rowanmiller has an idea. Also, is there any way we can support this mapping--that is the "spouse" mapping where any entity only has a principal or only has a dependent, and never both? Not saying we do it now, but if there is a reasonable way to do it we could file an issue for it.

ajcvickers:
:shipit:

src/EntityFramework.Core/Properties/Strings.resx
AndriySvyryd: You could do it on the clr entity type: Map the relationship using two private nav properties and access them through a unmapped Spouse property that will enforce this.

divega: It would be nicer to allow the nav-prop to be mapped twice for that case. But I don't think we need to do it now.

AndriySvyryd: It will require special code in fixup and query since we will need to get the fk value to determine whether this navigation is to principal or to dependent. I don't think its worth doing.

ajcvickers: I'm fine with us not doing it.

PR #1727, opened 2015-02-28
Query: Adds a SQL predicate optimizer that optimizes two cases:

anpete:
1) x.Foo = 1 OR x.Foo = 2 becomes x.Foo IN (1, 2)
2) x.Foo != 1 AND x.Foo != 2 becomes x.Foo NOT IN (1, 2)


ajcvickers:
:shipit:

divega:
Looks good, although the use of the word "optimization" is possibly a stretch. This in general may make the SQL nicer though.

Issue #1668, opened 2015-02-20
RevEng: Consult conventions to see whether Fluent API is necessary 


lajones:
Part of #830.

Investigate whether we can update the conventions code so we can directly ask it what it would produce without a given fluent API - so that if it is the same we can skip outputting the fluent API code in OnModelCreating().

lajones:
Examples are: when do we need to output the Key() API? Revisit the current assumptions about when to output the Table(), Column() and ColumnType() APIs.

rowanmiller:
Even if we don't use the conventions, we need to not generate fluent api for stuff that will be taken care of by convention

lajones:
Column() is issued only if the property name we're generating is different from the underlying column name. Planning on leaving that as is.

lajones:
Last part of this was deciding what to do about the relationship code which we currently output.

We decided that we'd like to continue outputting that code as it keeps things simpler (no having to know the ins and outs of when the convention will or will not define it for you) plus it helps educate customers on how those APIs work.

So closing this as nothing more to do.

PR #1732, opened 2015-03-02
Do not issue Key() fluent API if that's what the convention would pick anyway

lajones:
Part of addressing #1668

### DIFF
--- a/src/EntityFramework.Core/Metadata/ModelConventions/KeyConvention.cs
+++ b/src/EntityFramework.Core/Metadata/ModelConventions/KeyConvention.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Data.Entity.Metadata.ModelConventions
             return entityBuilder;
         }
 
-        protected virtual IReadOnlyList<Property> DiscoverKeyProperties([NotNull] EntityType entityType)
+        public virtual IReadOnlyList<Property> DiscoverKeyProperties([NotNull] EntityType entityType)
         {
             Check.NotNull(entityType, nameof(entityType));
 
--- a/src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGenerator.cs
+++ b/src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGenerator.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Relational.Design.CodeGeneration;
 using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Metadata.ModelConventions;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
@@ -25,7 +26,7 @@ namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
                 {
                     new Tuple<string, string>("ModelBuilder", "modelBuilder")
                 };
-
+        private static readonly KeyConvention _keyConvention = new KeyConvention();
 
         private List<string> _usedNamespaces = new List<string>()
                 {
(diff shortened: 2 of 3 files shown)


src/EntityFramework.Core/Metadata/ModelConventions/KeyConvention.cs
```diff
@@ -33,7 +33,7 @@ public virtual InternalEntityBuilder Apply(InternalEntityBuilder entityBuilder)
             return entityBuilder;
         }
 
-        protected virtual IReadOnlyList<Property> DiscoverKeyProperties([NotNull] EntityType entityType)
+        public virtual IReadOnlyList<Property> DiscoverKeyProperties([NotNull] EntityType entityType)
```
lajones: @AndriySvyryd At the moment I catch the exceptions that this method can throw. But should we consider adding a throwOnError bool instead? Then you could pass true and I could pass false (and have the code return null if the flag is false and it would have thrown).

AndriySvyryd: This exception will likely be converted to a log message.

lajones: @AndriySvyryd - when the conventions architecture calls this code, yes. But for this code I just care whether you can find a key or not - if it fails to find a key I don't need to know or report why. I just want to know that it has failed and hence that there is no key that would be chosen by convention.

AndriySvyryd: If it fails it will just return an empty list

AndriySvyryd:
:shipit:

src/EntityFramework.Core/Metadata/ModelConventions/KeyConvention.cs
lajones: @AndriySvyryd What about the multiple matching properties InvalidOperationException?

AndriySvyryd: It will just return an empty list in that case

lajones: @AndriySvyryd  ??? - lines 53-57 of KeyConvention currently throw in that situation. Am I missing a rebase or something?

AndriySvyryd: I am talking about the future version of the when I add logging to conventions.

lajones: @AndriySvyryd OK. Let's change this when we you add logging then. At the moment if I came across that situation RevEng would throw, when really all it needs to do is treat it as "no key found".

ajcvickers: I don't think we should be using exceptions for flow control, especially when we own the code that is throwing the exception.

ajcvickers: @divega Do you have an opinion?

divega: What about removing the exception as part of this change instead of waiting for logging support to happen? Would that break many tests? It seems Andriy is already tracking that change but a TODO might help point out where the warning needs to be logged. Thoughts?

AndriySvyryd: Fine by me

lajones: You mean remove throwing the exception and just return null? That works for me.

AndriySvyryd: Not null, empty list.

lajones: Yep - sorry - that's what I meant. OK. Will do. Thanks.

lajones:
Checked in - commit 0a187f8

PR #1733, opened 2015-03-02
Introduce interfaces for dynamic data store services

ajcvickers:
This change introduces interfaces for all data store services (except Database) that are dynamically selected based on the data store that is being used by the current context instance. This makes our DI use cleaner because we are always resolving these services by a contract type, rather than sometimes by the default implementation we have.

In addition, data store-specific interfaces have also been created for these servics. This means that these are also resolved from DI by contract and any replacement no longer needs to extend from our implementation.

Currently we still have the "empty interfaces" because we are regsitering everything in DI by data-store specific contract even if there is currently no difference between the base-specific contract and the data-store specific contract. This makes replacement of the service with a data-store specific one very easy. However, a further change may make this unnecessary and I will investigate that.

Database has not been updated to use a contract interface because it is a facade API that is resolved from DI. I'm going to investigate this further and will likely change the code to stop resolving facade APIs from DI since I believe the extension mechanisms we have don't require this and not doing it will allow the different goals of the two types of API to be better met.


anpete:
:shipit:

Issue #624, opened 2014-09-04
Raw store access APIs


divega:
In previous versions of EF we had handy "raw SQL" methods such as SqlQuery and ExecuteSqlCommand that allowed to drop down to the underlying relational provider to execute SQL queries directly (i.e. bypassing the EF mapping) but still take advantage of some of the services provided by EF, e.g. connection management, transaction handling, object materialization and change tracking. 

In EF7 since we support relational as well as non relational providers we should figure out how to extend this concept to specific providers through a consistent pattern, e.g. for Azure Table Storage it probably makes sense to provide access to the underlying objects from the .NET storage client library, while for relational stores being able to access the underlying connection and having something like SqlQuery and ExecuteSqlCommand defined as extension methods probably makes sense.

RickyLin:
hmm, any progress on this?

rowanmiller:
@RickyLin - We haven't started on this yet. It's a fairly isolated feature, so it will be something we work on a little later in the cycle.

mikary:
We have added a ```FromSql``` method for relational ```DbSet<T>```s that allows for materialization, change tracking, and additional querying off of a custom SQL statement. See #1845 and #1871 for more details and examples

NicoJuicy:
In the meantime, you could probably use a extension method like this 

public static int ExecuteSqlCommand(this RelationalDatabase database, string sql)
    {
        using (var connection = database.Connection)
        {
            using (var command = connection.DbConnection.CreateCommand())
            {
                command.CommandText = sql;
                return command.ExecuteNonQuery();
            }
        }

    }

bricelam:
@NicoJuicy Looks [eerily familiar](http://stackoverflow.com/a/30487747/475031)... :wink:

mikary:
@rowanmiller most of the functionality in this issue is covered by a combination of ```FromSql```, #1862, and #1863 

The only other facet is using a raw command on a non-relational store. Should we still keep this issue open as a reminder on the backlog, or is this something we can close?

rowanmiller:
@mikary yes I agree that those cover the remaining cases from this item

rowanmiller:
non-relational providers can do whatever makes sense for that particular store, so I don't think we need a general item.

kant2002:
Would be good to provide async version of #1863 (ExecuteSqlCommandAsync) as it was earlier. It's strange to have all API's Async but for this one only sync version.

rowanmiller:
[...]

PR #1734, opened 2015-03-02
Add RelationalDbSet for raw query

mikary:
Add custom relational queries off of DbSet to provide functionality for #624
@ajcvickers  @anpete @divega

### DIFF
--- a/src/EntityFramework.Core/Query/EntityQueryable`.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryable`.cs
@@ -2,19 +2,28 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Utilities;
 using Remotion.Linq;
 
 namespace Microsoft.Data.Entity.Query
 {
     public class EntityQueryable<TResult>
-        : QueryableBase<TResult>, IAsyncEnumerable<TResult>
+        : QueryableBase<TResult>, IAsyncEnumerable<TResult>, IAnnotatable
     {
+        private EntityQueryProvider _provider;
+
+        private readonly LazyRef<Annotatable> _annotatable
+            = new LazyRef<Annotatable>(
+                () => new Annotatable());
+
         public EntityQueryable([NotNull] EntityQueryProvider provider)
             : base(Check.NotNull(provider, nameof(provider)))
         {
+            _provider = provider;
         }
 
         public EntityQueryable([NotNull] EntityQueryProvider provider, [NotNull] Expression expression)
@@ -22,11 +31,59 @@ namespace Microsoft.Data.Entity.Query
                 Check.NotNull(provider, nameof(provider)),
                 Check.NotNull(expression, nameof(expression)))
         {
+            _provider = provider;
+        }
+
+        public virtual EntityQueryable<TResult> Clone()
+        {
+            return new EntityQueryable<TResult>(_provider);
         }
 
         IAsyncEnumerator<TResult> IAsyncEnumerable<TResult>.GetEnumerator()
         {
             return ((IAsyncQueryProvider)Provider).ExecuteAsync<TResult>(Expression).GetEnumerator();
+        }
+
+        public virtual Annotation AddAnnotation([NotNull] string annotationName, [NotNull] string value)
+        {
+            Check.NotEmpty(annotationName, nameof(annotationName));
+            Check.NotNull(value, nameof(value));
+
+            return _annotatable.Value.AddAnnotation(annotationName, value);
+        }
+
+        public virtual string this[[NotNull]string annotationName]
+        {
+            get
+            {
+                Check.NotNull(annotationName, annotationName);
+                return _annotatable.Value[annotationName];
+            }
+        }
+
+        public virtual IEnumerable<IAnnotation> Annotations
+        {
+            get
+            {
+                return _annotatable.Value.Annotations;
+            }
+        }
+
+        public virtual Annotation GetAnnotation([NotNull]string annotationName)
+        {
+            Check.NotNull(annotationName, nameof(annotationName));
+
+            return _annotatable.Value.GetAnnotation(annotationName);
+        }
+
+        public override string ToString()
+        {
+            return _annotatable.Value.Annotations.Count() == 0
+                ? base.ToString()
+                : string.Format("{0} ({1})",
+                    base.ToString(),
+                    string.Join(", ", _annotatable.Value.Annotations.Select(annotation =>
+                        string.Format("{0} = {1}", annotation.Name, annotation.Value))));
         }
     }
 }
--- a/src/EntityFramework.Core/Query/EntityQueryable`.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryable`.cs
@@ -14,16 +14,14 @@ namespace Microsoft.Data.Entity.Query
     public class EntityQueryable<TResult>
         : QueryableBase<TResult>, IAsyncEnumerable<TResult>, IAnnotatable
     {
-        private EntityQueryProvider _provider;
-
-        private readonly LazyRef<Annotatable> _annotatable
-            = new LazyRef<Annotatable>(
-                () => new Annotatable());
+        private readonly EntityQueryProvider _provider;
+        private readonly Annotatable _annotatable;
 
         public EntityQueryable([NotNull] EntityQueryProvider provider)
             : base(Check.NotNull(provider, nameof(provider)))
         {
             _provider = provider;
+            _annotatable = new Annotatable();
         }
 
         public EntityQueryable([NotNull] EntityQueryProvider provider, [NotNull] Expression expression)
@@ -32,11 +30,19 @@ namespace Microsoft.Data.Entity.Query
                 Check.NotNull(expression, nameof(expression)))
         {
             _provider = provider;
+            _annotatable = new Annotatable();
         }
 
         public virtual EntityQueryable<TResult> Clone()
         {
-            return new EntityQueryable<TResult>(_provider);
+            var clone = new EntityQueryable<TResult>(_provider);
+
+            foreach (var annotation in _annotatable.Annotations)
+            {
+                clone.AddAnnotation(annotation.Name, annotation.Value);
+            }
+
+            return clone;
         }
 
         IAsyncEnumerator<TResult> IAsyncEnumerable<TResult>.GetEnumerator()
@@ -44,45 +50,21 @@ namespace Microsoft.Data.Entity.Query
             return ((IAsyncQueryProvider)Provider).ExecuteAsync<TResult>(Expression).GetEnumerator();
         }
 
-        public virtual Annotation AddAnnotation([NotNull] string annotationName, [NotNull] string value)
-        {
-            Check.NotEmpty(annotationName, nameof(annotationName));
-            Check.NotNull(value, nameof(value));
+        public virtual Annotation AddAnnotation([NotNull] string annotationName, [NotNull] string value) => _annotatable.AddAnnotation(annotationName, value);
 
-            return _annotatable.Value.AddAnnotation(annotationName, value);
-        }
+        public virtual string this[[NotNull]string annotationName] => _annotatable[annotationName];
 
-        public virtual string this[[NotNull]string annotationName]
-        {
-            get
-            {
-                Check.NotNull(annotationName, annotationName);
-                return _annotatable.Value[annotationName];
-            }
-        }
+        public virtual IEnumerable<IAnnotation> Annotations => _annotatable.Annotations;
 
-        public virtual IEnumerable<IAnnotation> Annotations
-        {
-            get
-            {
-                return _annotatable.Value.Annotations;
-            }
-        }
-
-        public virtual Annotation GetAnnotation([NotNull]string annotationName)
-        {
-            Check.NotNull(annotationName, nameof(annotationName));
-
-            return _annotatable.Value.GetAnnotation(annotationName);
-        }
+        public virtual Annotation GetAnnotation([NotNull]string annotationName) => _annotatable.GetAnnotation(annotationName);
 
         public override string ToString()
         {
-            return _annotatable.Value.Annotations.Count() == 0
+            return _annotatable.Annotations.Count() == 0
                 ? base.ToString()
                 : string.Format("{0} ({1})",
                     base.ToString(),
-                    string.Join(", ", _annotatable.Value.Annotations.Select(annotation =>
+                    string.Join(", ", _annotatable.Annotations.Select(annotation =>
                         string.Format("{0} = {1}", annotation.Name, annotation.Value))));
         }
     }
--- a/src/EntityFramework.Core/Query/ExpressionTreeVisitors/EntityQueryableExpressionTreeVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionTreeVisitors/EntityQueryableExpressionTreeVisitor.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Query.ExpressionTreeVisitors
@@ -22,12 +23,14 @@ namespace Microsoft.Data.Entity.Query.ExpressionTreeVisitors
             if (constantExpression.Type.GetTypeInfo().IsGenericType
                 && constantExpression.Type.GetGenericTypeDefinition() == typeof(EntityQueryable<>))
             {
-                return VisitEntityQueryable(((IQueryable)constantExpression.Value).ElementType);
+                return VisitEntityQueryable(
+                    ((IQueryable)constantExpression.Value).ElementType,
+                    (IAnnotatable)constantExpression.Value);
             }
 
             return constantExpression;
         }
 
-        protected abstract Expression VisitEntityQueryable([NotNull] Type elementType);
+        protected abstract Expression VisitEntityQueryable([NotNull] Type elementType, IAnnotatable annotatable);
     }
 }
--- a/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
@@ -6,10 +6,8 @@ using System.Collections.Generic;
 using System.Data.Common;
 using System.Linq;
 using System.Linq.Expressions;
-using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Infrastructure;
-using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Query;
 using Microsoft.Data.Entity.Query.ExpressionTreeVisitors;
 using Microsoft.Data.Entity.Relational.Query.Expressions;
@@ -71,54 +69,11 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
             return base.VisitMethodCallExpression(methodCallExpression);
         }
 
-        protected override Expression VisitConstantExpression(ConstantExpression constantExpression)
-        {
-            Check.NotNull(constantExpression, nameof(constantExpression));
-
-            if (constantExpression.Type.GetTypeInfo().IsGenericType
-                && constantExpression.Type.GetGenericTypeDefinition() == typeof(EntityQueryable<>))
-            {
-                var sql = ((IAnnotatable)constantExpression.Value)["Sql"];
-
-                if (sql != null)
-                {
-                    return VisitRawSqlQueryable(((IQueryable)constantExpression.Value).ElementType, sql);
-                }
-            }
-
-            return base.VisitConstantExpression(constantExpression);
-        }
-
-        protected override Expression VisitEntityQueryable(Type elementType)
+        protected override Expression VisitEntityQueryable(Type elementType, IAnnotatable annotatable)
         {
             Check.NotNull(elementType, nameof(elementType));
+            Check.NotNull(annotatable, nameof(annotatable));
 
-            return CreateQuery(elementType,
-                (entityType, tableName, alias) =>
-                    new TableExpression(
-                        tableName,
-                        QueryModelVisitor.QueryCompilationContext.GetSchema(entityType),
-                        alias,
-                        _querySource));
-        }
-
-        protected virtual Expression VisitRawSqlQueryable(Type elementType, string sql)
-        {
-            Check.NotNull(elementType, nameof(elementType));
-            Check.NotNull(sql, nameof(sql));
-
-            return CreateQuery(elementType,
-                (entityType, tableName, alias) =>
-                    new RawSqlDerivedTableExpression(
-                        sql,
-                        alias,
-                        _querySource));
-        }
-
-        private Expression CreateQuery(
-            Type elementType,
-            Func<IEntityType, string, string, TableExpressionBase> createTableExpression)
-        {
             var queryMethodInfo = RelationalQueryModelVisitor.CreateValueReaderMethodInfo;
             var entityType = QueryModelVisitor.QueryCompilationContext.Model.GetEntityType(elementType);
 
@@ -129,7 +84,23 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
                 ? tableName.First().ToString().ToLower()
                 : _querySource.ItemName;
 
-            selectExpression.AddTable(createTableExpression(entityType, tableName, alias));
+            if (annotatable["Sql"] != null)
+            {
+                selectExpression.AddTable(
+                    new RawSqlDerivedTableExpression(
+                        annotatable["Sql"],
+                        alias,
+                        _querySource));
+            }
+            else
+            {
+                selectExpression.AddTable(
+                    new TableExpression(
+                        tableName,
+                        QueryModelVisitor.QueryCompilationContext.GetSchema(entityType),
+                        alias,
+                        _querySource));
+            }
 
             QueryModelVisitor.AddQuery(_querySource, selectExpression);
 
--- a/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
@@ -6,7 +6,10 @@ using System.Collections.Generic;
 using System.Data.Common;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Query;
 using Microsoft.Data.Entity.Query.ExpressionTreeVisitors;
 using Microsoft.Data.Entity.Relational.Query.Expressions;
@@ -68,25 +71,65 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
             return base.VisitMethodCallExpression(methodCallExpression);
         }
 
+        protected override Expression VisitConstantExpression(ConstantExpression constantExpression)
+        {
+            Check.NotNull(constantExpression, nameof(constantExpression));
+
+            if (constantExpression.Type.GetTypeInfo().IsGenericType
+                && constantExpression.Type.GetGenericTypeDefinition() == typeof(EntityQueryable<>))
+            {
+                var sql = ((IAnnotatable)constantExpression.Value)["Sql"];
+
+                if (sql != null)
+                {
+                    return VisitRawSqlQueryable(((IQueryable)constantExpression.Value).ElementType, sql);
+                }
+            }
+
+            return base.VisitConstantExpression(constantExpression);
+        }
+
         protected override Expression VisitEntityQueryable(Type elementType)
         {
             Check.NotNull(elementType, nameof(elementType));
 
+            return CreateQuery(elementType,
+                (entityType, tableName, alias) =>
+                    new TableExpression(
+                        tableName,
+                        QueryModelVisitor.QueryCompilationContext.GetSchema(entityType),
+                        alias,
+                        _querySource));
+        }
+
+        protected virtual Expression VisitRawSqlQueryable(Type elementType, string sql)
+        {
+            Check.NotNull(elementType, nameof(elementType));
+            Check.NotNull(sql, nameof(sql));
+
+            return CreateQuery(elementType,
+                (entityType, tableName, alias) =>
+                    new RawSqlDerivedTableExpression(
+                        sql,
+                        alias,
+                        _querySource));
+        }
+
+        private Expression CreateQuery(
+            Type elementType,
+            Func<IEntityType, string, string, TableExpressionBase> createTableExpression)
+        {
             var queryMethodInfo = RelationalQueryModelVisitor.CreateValueReaderMethodInfo;
             var entityType = QueryModelVisitor.QueryCompilationContext.Model.GetEntityType(elementType);
 
             var selectExpression = new SelectExpression();
             var tableName = QueryModelVisitor.QueryCompilationContext.GetTableName(entityType);
 
-            selectExpression
-                .AddTable(
-                    new TableExpression(
-                        tableName,
-                        QueryModelVisitor.QueryCompilationContext.GetSchema(entityType),
-                        _querySource.ItemName.StartsWith("<generated>_")
-                            ? tableName.First().ToString().ToLower()
-                            : _querySource.ItemName,
-                        _querySource));
+            var alias = _querySource.ItemName.StartsWith("<generated>_")
+                ? tableName.First().ToString().ToLower()
+                : _querySource.ItemName;
+
+            selectExpression.AddTable(createTableExpression(entityType, tableName, alias));
 
             QueryModelVisitor.AddQuery(_querySource, selectExpression);
 
--- a/src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
+++ b/src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
@@ -10,26 +10,22 @@ using Remotion.Linq.Parsing;
 
 namespace Microsoft.Data.Entity.Relational.Query.Expressions
 {
-    public class TableExpression : TableExpressionBase
+    public class RawSqlDerivedTableExpression : TableExpressionBase
     {
-        public TableExpression(
-            [NotNull] string table,
-            [CanBeNull] string schema,
+        public RawSqlDerivedTableExpression(
+            [NotNull] string sql,
             [NotNull] string alias,
             [NotNull] IQuerySource querySource)
             : base(
-                Check.NotNull(querySource, nameof(querySource)),
-                Check.NotEmpty(alias, nameof(alias)))
+                  Check.NotNull(querySource, nameof(querySource)),
+                  Check.NotEmpty(alias, nameof(alias)))
         {
-            Check.NotEmpty(table, nameof(table));
+            Check.NotEmpty(sql, nameof(sql));
 
-            Table = table;
-            Schema = schema;
+            Sql = sql;
         }
 
-        public virtual string Table { get; }
-
-        public virtual string Schema { get; }
+        public virtual string Sql { get; }
 
         public override Expression Accept([NotNull] ExpressionTreeVisitor visitor)
         {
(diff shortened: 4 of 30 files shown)


src/EntityFramework.Core/Utilities/ServiceProviderExtensions.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Microsoft.Framework.DependencyInjection
 {
-    internal static class ServiceProviderExtensions
+    public static class ServiceProviderExtensions
```
anpete: If you need these, share them, rather than making this public.

anpete:
Can you please describe what's in this change?

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -148,5 +158,82 @@ var keyFactory
                     Expression.Call(queryMethodInfo, queryMethodArguments),
                     _readerParameter));
         }
+
+        protected Expression VisitRelationalCustomQueryable(string sql, Type elementType)
```
anpete: Don't duplicate the code below. Refactor both into something that pivots on the source table expr.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -163,6 +163,23 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
             }
         }
 
+        public virtual Expression VisitCustomTableExpression(CustomTableExpression customTableExpression)
+        {
+            Check.NotNull(customTableExpression, nameof(customTableExpression));
+
+            if (customTableExpression.Schema != null)
```
anpete: Don't think this is needed, is it?

src/EntityFramework.Relational/Query/Expressions/CustomTableExpression.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            Check.NotEmpty(table, nameof(table));
+
+            Table = table;
+            Schema = schema;
+        }
+
+        public virtual string Table { get; }
+
+        public virtual string Schema { get; }
```
anpete: Don't think this is needed, is it?

src/EntityFramework.Relational/Query/Expressions/CustomTableExpression.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Relational.Query.Sql;
+using Microsoft.Data.Entity.Utilities;
+using Remotion.Linq.Clauses;
+using Remotion.Linq.Parsing;
+
+namespace Microsoft.Data.Entity.Relational.Query.Expressions
+{
+    public class CustomTableExpression : TableExpressionBase
```
anpete: Can we rename this to "RawSqlDerivedTableExpression"?

src/EntityFramework.Relational/Query/Expressions/CustomTableExpression.cs
mikary: This is how the ```DefaultSqlQueryGenerator``` is able to distinguish a table from a raw query so it knows to make a sub query.

src/EntityFramework.Relational/Query/Expressions/CustomTableExpression.cs
mikary: "Custom" is a pretty bad name, it would probably be better as "RawSql" in most places

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
mikary: Would we want an optional expression parameter so the existing code path can use the element type as it was but we still have a way to pass in the query string?

src/EntityFramework.Relational/Query/RelationalCustomQueryProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Query;
+using Microsoft.Data.Entity.Storage;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Relational.Query
+{
+    class RelationalCustomQueryProvider : IAsyncQueryProvider
```
anpete: Let's try and do this using a custom LINQ operator instead. Take a look at how AsNoTracking and Include are inplemented. The difference here is that the op will be defined in Relational and so we will need to figure out how it gets registered - currently this happens in CompiledQueryCache.CreateNodeTypeProvider

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Relational;
+using Microsoft.Data.Entity.Utilities;
+
+// ReSharper disable once CheckNamespace
+
+namespace Microsoft.Data.Entity
+{
+    public static class RelationalDbSetExtensions
```
anpete: This can go away and the new operator can just hang from IQueryable<T>

test/EntityFramework.Relational.FunctionalTests/QueryRelationalTest.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+    public class QueryRelationalTest<TFixture> : QueryTestBase<TFixture>
+        where TFixture : NorthwindQueryFixtureBase, new()
+    {
+        public QueryRelationalTest(TFixture fixture)
+            : base(fixture)
+        {
+        }
+
+        protected const string customQueryableSimple = "SELECT * FROM Customers";
```
anpete: Inline these. Optimize for readability in tests.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
anpete: No, just extract a new method and call it from both places. It may be DRY'er to have the new method take a Func<TableExpressionBase> as a factory for the table expr. but I would see what it looks like to just build the table expr. externally first.

src/EntityFramework.Relational/Query/Expressions/CustomTableExpression.cs
anpete: Meant Schema.

src/EntityFramework.Relational/Query/RelationalCustomQueryProvider.cs
anpete: Ideas for the operator name:

1) "FromSql"

cc @divega

divega: I had the same in mind. +1

mikary:
Updated based on PR feedback

src/EntityFramework.Core/Query/EntityQueryable`.cs
```diff
@@ -28,5 +34,46 @@ IAsyncEnumerator<TResult> IAsyncEnumerable<TResult>.GetEnumerator()
...
+        public virtual Annotation GetAnnotation([NotNull]string annotationName)
+        {
+            Check.NotNull(annotationName, nameof(annotationName));
+
+            return _annotatable.Value.GetAnnotation(annotationName);
+        }
+
+        public override string ToString()
+        {
+            return base.ToString() + string.Join(", ", _annotatable.Value.Annotations.Select(annotation => annotation.Value));
```
maumar: fix the formatting - e.g. space between base.ToString() and annotations, maybe wrap annotations in some kind of parenthesis.

src/EntityFramework.Core/Query/EntityQueryable`.cs
```diff
@@ -28,5 +34,46 @@ IAsyncEnumerator<TResult> IAsyncEnumerable<TResult>.GetEnumerator()
...
+        }
+
+        private class Annotatable : MetadataBase
```
maumar: Consider more explicit name - not super important since it's private class

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -68,18 +69,25 @@ protected override Expression VisitMethodCallExpression([NotNull] MethodCallExpr
             return base.VisitMethodCallExpression(methodCallExpression);
         }
 
-        protected override Expression VisitEntityQueryable(Type elementType)
+        protected override Expression VisitConstantExpression(ConstantExpression constantExpression)
         {
-            Check.NotNull(elementType, nameof(elementType));
+            var entityQueryable = constantExpression.Value as IEntityQueryable;
```
maumar: missing check.notnull

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -68,18 +69,25 @@ protected override Expression VisitMethodCallExpression([NotNull] MethodCallExpr
...
 
-            var queryMethodInfo = RelationalQueryModelVisitor.CreateValueReaderMethodInfo;
-            var entityType = QueryModelVisitor.QueryCompilationContext.Model.GetEntityType(elementType);
+            if (entityQueryable != null)
```
maumar: collapse into one condition

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -87,6 +95,33 @@ protected override Expression VisitEntityQueryable(Type elementType)
...
+                            ? tableName.First().ToString().ToLower()
+                            : _querySource.ItemName,
+                        _querySource));
+        }
+
+        private Expression VisitSelectExpression(
+            Type elementType,
+            Func<IEntityType, string, TableExpressionBase> createTableExpression)
+        {
+            Check.NotNull(elementType, nameof(elementType));
```
maumar: i would move the check to the public surface and remove it from here

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -87,6 +95,33 @@ protected override Expression VisitEntityQueryable(Type elementType)
...
                             : _querySource.ItemName,
                         _querySource));
+        }
+
+        protected virtual Expression VisitRawSqlQueryable(Type elementType, string rawSql)
+        {
+            return VisitSelectExpression(elementType,
+                (entityType, tableName) =>
+                    new RawSqlDerivedTableExpression(
+                        rawSql,
```
maumar: consider not passing rawsql as closure to avoid extra string allocation (especially that rawsql can be arbitrarily large)

src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
```diff
@@ -0,0 +1,46 @@
...
+            var specificVisitor = visitor as ISqlExpressionVisitor;
+
+            return specificVisitor != null
+                ? specificVisitor.VisitRawSqlDerivedTableExpression(this)
+                : base.Accept(visitor);
+        }
+
+        public override string ToString()
+        {
+            return RawSql + " " + Alias;
```
maumar: add parentheses around RawSql for more readability

src/EntityFramework.Relational/RelationalDbSet`.cs
```diff
@@ -0,0 +1,36 @@
...
+        public RelationalDbSet([NotNull]DbSet<TEntity> dbSet)
+        {
+            Check.NotNull(dbSet, nameof(dbSet));
+
+            _queryProvider
+                = new LazyRef<EntityQueryProvider>(
+                    () => ((IAccessor<IServiceProvider>)dbSet).Service.GetRequiredService<EntityQueryProvider>());
+        }
+
+        public virtual IQueryable<TEntity> FromSql([NotNull]string query)
```
anpete: Use extension method rather than custom DbSet.

test/EntityFramework.Relational.FunctionalTests/QueryRelationalTest.cs
```diff
@@ -0,0 +1,88 @@
...
+
+using System;
+using System.Linq;
+using Microsoft.Data.Entity.FunctionalTests;
+using Microsoft.Data.Entity.FunctionalTests.TestModels.Northwind;
+using Xunit;
+
+namespace Microsoft.Data.Entity.Relational.FunctionalTests
+{
+    public class QueryRelationalTest<TFixture> : QueryTestBase<TFixture>
```
anpete: Don't inherit here as the base tests are going to be run twice. Refactor the asserts into a test helper if required.
Make this test base class abstract.

test/EntityFramework.Relational.FunctionalTests/AsyncQueryRelationalTest.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Data.Entity.FunctionalTests;
+using Microsoft.Data.Entity.FunctionalTests.TestModels.Northwind;
+using Xunit;
+
+namespace Microsoft.Data.Entity.Relational.FunctionalTests
+{
+    public class AsyncQueryRelationalTest<TFixture> : AsyncQueryTestBase<TFixture>
```
anpete: Don't inherit here as the base tests are going to be run twice. Refactor the asserts into a test helper if required.
Make this test base class abstract.

src/EntityFramework.Core/Query/IEntityQueryable.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using Microsoft.Data.Entity.Metadata;
+
+namespace Microsoft.Data.Entity.Query
+{
+    public interface IEntityQueryable : IMetadata, IQueryable
```
anpete: What is the purpose of this interface?

src/EntityFramework.Core/Query/ExpressionTreeVisitors/EntityQueryableExpressionTreeVisitor.cs
```diff
@@ -19,10 +19,11 @@ protected EntityQueryableExpressionTreeVisitor([NotNull] EntityQueryModelVisitor
 
         protected override Expression VisitConstantExpression(ConstantExpression constantExpression)
         {
-            if (constantExpression.Type.GetTypeInfo().IsGenericType
-                && constantExpression.Type.GetGenericTypeDefinition() == typeof(EntityQueryable<>))
+            var entityQueryable = constantExpression.Value as IEntityQueryable;
```
anpete: This need to change?

src/EntityFramework.Core/Query/EntityQueryable`.cs
anpete: I think we should just change MetadataBase to be this and move it out of metadata.

src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.Data.Entity.Utilities;
+using Remotion.Linq.Clauses;
+using Remotion.Linq.Parsing;
+
+namespace Microsoft.Data.Entity.Relational.Query.Expressions
+{
+    public class RawSqlDerivedTableExpression : TableExpressionBase
+    {
+        public RawSqlDerivedTableExpression(
+            [NotNull] string rawSql,
```
anpete: "sql"

src/EntityFramework.Relational/Query/Expressions/TableExpression.cs
```diff
@@ -22,7 +22,6 @@ public class TableExpression : TableExpressionBase
                 Check.NotEmpty(alias, nameof(alias)))
         {
             Check.NotEmpty(table, nameof(table));
-            Check.NotNull(querySource, nameof(querySource));
```
anpete: Good catch.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -657,5 +668,12 @@ protected virtual string DelimitIdentifier([NotNull] string identifier)
 
             return "\"" + identifier + "\"";
         }
+
+        protected virtual string DelimitSubQuery([NotNull] string query)
```
anpete: If this is not overridden, I would just inline it.

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -1887,6 +1887,54 @@ public override void Where_chain()
...
+                Sql);
+        }
+
+        public override void From_sql_queryable_where_simple_closure_via_query_cache()
+        {
+            base.From_sql_queryable_where_simple_closure_via_query_cache();
+
+            Assert.Equal(
+                @"SELECT [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[CustomerID], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM (SELECT * FROM Customers WHERE Customers.ContactName LIKE '%o%') AS [c]
```
anpete: Start the subquery on a new line and indented - see Take_Distinct_Count for example.

src/EntityFramework.Core/Query/IEntityQueryable.cs
mikary: It is intended to simplify the casting logic in the Expression Tree Visitors.

src/EntityFramework.Core/Query/ExpressionTreeVisitors/EntityQueryableExpressionTreeVisitor.cs
mikary: Need is a strong word, but this logic is common between both visitors and the reflection work seems a bit heavyweight for identifying a type and pulling out non-generic property values

src/EntityFramework.Core/Query/EntityQueryable`.cs
mikary: I like the idea of having a non-abstract version of this class for re-usability (Trying to keep a DRY solution where we have to avoid multiple-inheritance)
Any thoughts on this thread? @bricelam

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
mikary: This probably isn't the best factoring, we could also calculate these parameters within the methods and just pass in a ```TableExpressionBase``` but there are properties of this construction that would be calculated in three different places in that factoring.

src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
mikary: That's not a pattern I've seen much before, how does it aid readability?

src/EntityFramework.Relational/RelationalDbSet`.cs
mikary: Any suggestions for making sure this doesn't leak outside of the relational context?

anpete: It only "lights up" if you are referencing relational.

src/EntityFramework.Core/Query/ExpressionTreeVisitors/EntityQueryableExpressionTreeVisitor.cs
anpete: I would prefer to leave this code as it - unless it is needed for this change.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
anpete: Move the alias construction into the main method and pass the alias to the func.

src/EntityFramework.Core/Query/EntityQueryable`.cs
bricelam: I'm fine with either inheritance or composition for sharing the implementation. If everything can inherit then yes, we should rename MetadataBase to Annotatable and move it to Infrastructure.

mikary: Added an issue (#1778) to track this, since it's not directly related to this PR.

mikary:
More updates based on PR input

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
divega: Does this comment still apply? Sounds to me like we actually only want this to be callable on a root, i.e. on a DbSet<T> by I might be missing something.

test/EntityFramework.SqlServer.FunctionalTests/FromSqlQuerySqlServerTest.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            Assert.Equal(
+                @"SELECT [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[CustomerID], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM (
+    SELECT * FROM Customers WHERE Customers.City = 'London'
+) AS [c]
+
+SELECT [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[CustomerID], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM (
+    SELECT * FROM Customers WHERE Customers.City = 'Seattle'
```
divega: If the raw SQL query contains multiple lines, will they all be indented the same?

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.Data.Entity.Utilities;
+using Microsoft.Framework.DependencyInjection;
+
+// ReSharper disable once CheckNamespace
+
+namespace Microsoft.Data.Entity
+{
+    public static class RelationalDbSetExtensions
+    {
+        public static IQueryable<TEntity> FromSql<TEntity>([NotNull]this DbSet<TEntity> dbSet, [NotNull]string query) where TEntity : class
```
divega: What is the parameter story?

divega:
@mikary @anpete @maumar Is there any more reviewing happening on this or are we good with it? I think we should figure out handling parameters (unless I am missing something) but that could go in a different commit.

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
mikary: I don't think this is applicable anymore. The operator approach resulted in unwanted client side processing and leaves questions about how to process custom SQL at an arbitrary node in the query.

src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
```diff
@@ -0,0 +1,46 @@
...
+            var specificVisitor = visitor as ISqlExpressionVisitor;
+
+            return specificVisitor != null
+                ? specificVisitor.VisitRawSqlDerivedTableExpression(this)
+                : base.Accept(visitor);
+        }
+
+        public override string ToString()
+        {
+            return Sql + " " + Alias;
```
maumar: "(" + Sql + ") AS " + Alias?

test/EntityFramework.SqlServer.FunctionalTests/FromSqlQuerySqlServerTest.cs
mikary: There isn't any logic in place to do this right now, did we want to try?

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
mikary: I was just looking to get the basic DbSet pipeline in place with this PR, we still have parameters, best-effort materialization(off of the Database) and possibly dynamic types.

src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
mikary: That doesn't seem very DRY to me, since this would be guessing at the interpretation of the SQL generator that consumes this object.

The current implementation is consistent with the other TableExpressions, but we could revisit this if it's important / useful enough.

test/EntityFramework.SqlServer.FunctionalTests/FromSqlQuerySqlServerTest.cs
anpete: Should be easy if you use a StringReader to read each line and append to the output.

src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
anpete: This is just for debugging and so do whatever makes it clear.

test/EntityFramework.SqlServer.FunctionalTests/FromSqlQuerySqlServerTest.cs
mikary: Would we need to worry about literal line-breaks in quoted strings etc.?

anpete: I don't think you can do that - at least in TSQL.

divega: Maybe it is not to bad, but the alternative is to not worry about it at all and not try to indent raw SQL.

src/EntityFramework.Relational/Query/Expressions/RawSqlDerivedTableExpression.cs
mikary: Added #1795 to track this

mikary:
Updated based on feedback

src/EntityFramework.Core/Query/EntityQueryable`.cs
```diff
@@ -28,5 +34,47 @@ IAsyncEnumerator<TResult> IAsyncEnumerable<TResult>.GetEnumerator()
         {
             return ((IAsyncQueryProvider)Provider).ExecuteAsync<TResult>(Expression).GetEnumerator();
         }
+
+        public virtual Annotation AddAnnotation([NotNull] string annotationName, [NotNull] string value)
+        {
+            Check.NotNull(annotationName, nameof(annotationName));
```
anpete: NotEmpty

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -68,25 +71,65 @@ protected override Expression VisitMethodCallExpression([NotNull] MethodCallExpr
...
+            }
+
+            return base.VisitConstantExpression(constantExpression);
+        }
+
         protected override Expression VisitEntityQueryable(Type elementType)
         {
             Check.NotNull(elementType, nameof(elementType));
 
+            return VisitSelectExpression(elementType,
```
anpete: Calling this "Visit" is confusing. How about "CreateQuery"?

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -177,6 +178,31 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
             }
         }
 
+        public virtual Expression VisitRawSqlDerivedTableExpression([NotNull] RawSqlDerivedTableExpression rawSqlDerivedTableExpression)
+        {
+            Check.NotNull(rawSqlDerivedTableExpression, nameof(rawSqlDerivedTableExpression));
+
+            _sql.Append("(")
```
anpete: Just AppendLine?

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -177,6 +178,31 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
...
+                .AppendLine();
+
+            using (_sql.Indent())
+            {
+                var reader = new StringReader(rawSqlDerivedTableExpression.Sql);
```
anpete: Disposable

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -177,6 +178,31 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
...
+
+                var line = reader.ReadLine();
+                while (line != null)
```
anpete: string line;
while ((line = reader.ReadLine()) != null) {

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
anpete: Yep, let's do that separately.

anpete: Reformat this line, move constraint to new line.

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            Check.NotNull(dbSet, nameof(dbSet));
+            Check.NotNull(query, nameof(query));
+
+            if (dbSet.Context.Database as RelationalDatabase == null)
```
anpete: Remove this check (and the Context property). If we are in this method then we can assume relational is in use.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
mikary: I think this would harm readability

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                throw new InvalidOperationException(Strings.RelationalNotInUse);
+            }
+
+            var queryProvider = ((IAccessor<IServiceProvider>)dbSet).Service.GetRequiredService<EntityQueryProvider>();
```
anpete: Rather than do this, explicitly implement IAnnotatable on DbSet<> and then provide an impl in InternalDbSet. The impl can just delegate to _entityQueryable.

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
mikary: Without this, someone can include relational and make a call on a non-relational context (i.e. InMemoryStore) and it will silently ignore the custom SQL.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
anpete: This is a well known pattern for avoiding the duplicate calls to ReadLine.

test/EntityFramework.Core.Tests/TestHelpers.cs
```diff
@@ -174,5 +177,39 @@ public ModelBuilder CreateConventionBuilder(Model model = null)
 
             return entry;
         }
+
+        public static int AssertResults<T>(
```
anpete: DRY this up everywhere. Rename to AssertQuery.

test/EntityFramework.Relational.FunctionalTests/FromSqlQueryTestBase.cs
```diff
@@ -0,0 +1,107 @@
...
+                entryCount: 6);
+
+            AssertQuery<Customer>(
+                cs => cs.FromSql("SELECT * FROM Customers WHERE Customers.City = 'Seattle'"),
+                cs => cs.Where(c => c.City == "Seattle"),
+                entryCount: 1);
+        }
+
+        [Fact]
+        public virtual void From_sql_queryable_where_simple_closure_via_query_cache()
```
anpete: Test the same query but with different FromSql.

mikary: I thought that case was covered by ```From_sql_queryable_cached_by_query```
is this test just redundant?

test/EntityFramework.Core.Tests/TestHelpers.cs
mikary: Do we want all of the ```AssertQuery``` (sync and async) moved in here?

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
mikary: I'm not sure I understand what you mean. I'm getting the provider so that I can create a query and creating a new EntityQueryable to make sure we aren't persisting any annotations beyond this call.

src/EntityFramework.Core/Query/EntityQueryable`.cs
mikary: That's a stronger constraint than we have on ```Annotatable``` itself, do we want to update that as well?

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
anpete: Seems unlikely. Either way, the way this is checked seems hacky. I don't think we can guarantee that Database will always be a RelationalDatabase. The type of the concrete Database class only seems coincidentally related to whether or not the underlying query provider understands the annotation.

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
anpete: We always have a unique EntityQueryable instance per DbSet.

test/EntityFramework.Core.Tests/TestHelpers.cs
anpete: Only where it makes it DRY'er

test/EntityFramework.Relational.FunctionalTests/FromSqlQueryTestBase.cs
anpete: Oh, I see that now. Just wondering how it works because it seems like there should be a single cache entry that is bound to the first SQL statement.

src/EntityFramework.Core/Query/EntityQueryable`.cs
anpete: Yes, seems like an oversight.

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
divega: What if the return type of this method was not just IQueryable, but something that still denotes it is a query root, e.g. DbSet? I thought about it because I was wondering how to enable the chaining we talked about today, e.g. of different provider specific raw query methods, but I think beyond that it would make sense since the method just adds information to the DbSet. Thoughts?

mikary: We could have some kind of ```DbSet<T>``` wrapper or builder returned instead of IQueryable.

I'm currently working with a change where I added ```IAccessor<EntityQueryable>``` to ```DbSet<T>```, but the extension method clones the query before annotating so it still won't chain like that yet

mikary:
Rebased, squashed, updated, and ready for another pass.
Of note:
* Simplified extension method
* DataStore now has an ```InSupported(IAnnotation)``` method that is checked when the query is compiled
* Simplified Query Tree Visitor (added ```IAnnotatable``` parameter to constant expression and simplified implementation logic)

mikary:
After talk with @anpete, looking at using QueryAnnotations to store the FromSql string and parameters.

divega:
@mikary @anpete Are we going to check in what we have and then do that or just keep iterating on this PR?

anpete:
This is mostly a rewrite so a new PR.

mikary:
Closing PR, will make a new one for the next revision.

PR #1735, opened 2015-03-02
Stop resolving Database facade API directly from DI

ajcvickers:
We want to resolve DI services by contract, and specifically by interfaces for dynamic data store services. However, we want facade classes to have explicitly implemented interfaces for people writing extension methods on them. These two things were in conflict. This change adds a simple factory abstraction to allow a contract to be resolved from DI that will generate the facade API. We might also consider not extending RelationalDatabase, etc from Database since this is not really required anymore, but that is not done here.


anpete:
:shipit:

PR #1736, opened 2015-03-02
Remove use of EntityMaterializerSource.GetMaterializer from InMemory

anpete:
Replaced with a simple materializer folded into the query execution tree.


ajcvickers:
:shipit:

Issue #1726, opened 2015-02-28
Query :: using boolProperty.Equals(true) in filter produces invalid sql


maumar:
```
ctx.Data.Where(x => x.Excluded.Equals(true)).ToList();
```

gets translated to:

```
SELECT [x].[Date], [x].[Excluded], [x].[Id], [x].[IdType]
FROM [Data] AS [x]
WHERE [x].[Excluded] = 1 = 1 = 1
```

which is invalid sql. This is caused by the fix to #717

maumar:
Fixed in 2c4b376aef3d1faf5fa1963fe5253a2c18e31f4b

Issue #717, opened 2014-09-28
Boolean expression bug


rubenalves:
If i do something like

var products = db.Products.Where(c=>c.IsActivo).ToList();  i get this error
An expression of non-boolean type specified in a context where a condition is expected

but if i do 
var products = db.Products.Where(c=>c.IsActivo == true).ToList();

it works as expected, in EF 6 it works fine.

RAlves

PR #1740, opened 2015-03-03
Fix to #1726 - using boolProperty.Equals(true) in filter produces invalid sql

maumar:
Fix is to recognize this pattern in filtering visitor and compensate accordingly.


maumar:
@anpete

anpete:
Can you do this in EqualsTranslator?

maumar:
@anpete by the time we reach EqualsTranslator the we already do the rewrite for #717, so the expression is a bit more complicated. I could do it there, but it feels like bit of extra work.

anpete:
@maumar Lets chat about this. I don't really want to hard code custom method translation handling here.

maumar:
sent a new PR about this, closing

PR #1741, opened 2015-03-03
Set GenerateValueOnAdd flag for all Key properties by convention

### LINKED ISSUE #1116: Code First: Relational: Invalid database schema created when one-to-one association
The dependent table gets created with an Identity key, which doesn't work because it's an FK.

More generally, it is typically invalid to have a store generated FK. How should we handle? Check in model? Check in update pipeline? No explicit check?

### LINKED ISSUE #1126: SqlServer: Autogenerated integer values for composite PK

smitpatel:
Implements #1271 
Added KeyConvention to set GenerateValueOnAdd flag(s) as true whenever new key is added.
If new key is existing foreign key then flag(s) are not set.
If foreign key is set on existing key then flag(s) are changed to false.
If foreign key is removed then the convention is run again on the key.

Resolves #1126 
In SqlServer provider,
for PK with only 1 property which is of common integer type, we assign SqlServerValueGenerationStrategy.Default annotation. Which is a pointer that this property can use Identity. If property level or model level value generation strategy is assigned then we use it otherwise Identity will be used as value generation strategy.
For all other cases, composite PK or PK with non-int property, we do nothing.

fixes #1116 
Due to above conventions, now the dependent end of FK doesn't have auto-generated value so there is no identity key in dependent table. ModelSnapshot still has Default as value generation strategy annotated which is valid model. And that whole area will change soon. Verified through sample project.

### DIFF
--- a/src/EntityFramework.Core/Metadata/ModelConventions/KeyConvention.cs
+++ b/src/EntityFramework.Core/Metadata/ModelConventions/KeyConvention.cs
@@ -1,82 +1,44 @@
-// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using JetBrains.Annotations;
-using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Metadata.Internal;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Metadata.ModelConventions
 {
-    public class KeyConvention : IEntityTypeConvention
+    public class KeyConvention : IKeyConvention, IForeignKeyRemovedConvention
     {
-        private const string KeySuffix = "Id";
+        public virtual InternalKeyBuilder Apply(InternalKeyBuilder keyBuilder)
+        {
+            Check.NotNull(keyBuilder, nameof(keyBuilder));
 
-        public virtual InternalEntityBuilder Apply(InternalEntityBuilder entityBuilder)
+            ConfigureKeyProperties(keyBuilder.ModelBuilder.Entity(keyBuilder.Metadata.EntityType.Name, ConfigurationSource.Convention),
+                keyBuilder.Metadata.Properties);
+
+            return keyBuilder;
+        }
+
+        protected virtual void ConfigureKeyProperties([NotNull] InternalEntityBuilder entityBuilder, [NotNull] IReadOnlyList<Property> properties)
         {
             Check.NotNull(entityBuilder, nameof(entityBuilder));
-            var entityType = entityBuilder.Metadata;
-
-            var keyProperties = DiscoverKeyProperties(entityType);
-            if (keyProperties.Count != 0
-                && entityBuilder.PrimaryKey(keyProperties.Select(p => p.Name).ToList(), ConfigurationSource.Convention) != null)
+            Check.NotNull(properties, nameof(properties));
+            foreach (var property in properties.Where(property => !entityBuilder.Metadata.ForeignKeys.SelectMany(fk => fk.Properties).Contains(property)))
             {
-                foreach (var property in keyProperties)
-                {
-                    ConfigureKeyProperty(entityBuilder.Property(property.PropertyType, property.Name, ConfigurationSource.Convention));
-                }
+                entityBuilder.Property(property.PropertyType, property.Name, ConfigurationSource.Convention).GenerateValueOnAdd(true, ConfigurationSource.Convention);
             }
-
-            return entityBuilder;
-        }
-
-        public virtual IReadOnlyList<Property> DiscoverKeyProperties([NotNull] EntityType entityType)
-        {
-            Check.NotNull(entityType, nameof(entityType));
-
-            // TODO: Honor [Key]
-            // Issue #213
-            var keyProperties = entityType.Properties
-                .Where(p => string.Equals(p.Name, KeySuffix, StringComparison.OrdinalIgnoreCase))
-                .ToList();
-
-            if (keyProperties.Count == 0)
-            {
-                keyProperties = entityType.Properties.Where(
-                    p => string.Equals(p.Name, entityType.SimpleName + KeySuffix, StringComparison.OrdinalIgnoreCase))
-                    .ToList();
-            }
-
-            if (keyProperties.Count > 1)
-            {
-                //TODO - add in logging using resource Strings.MultiplePropertiesMatchedAsKeys()
-                return Enumerable.Empty<Property>().ToList();
-            }
-
-            return keyProperties;
-        }
-
-        protected virtual void ConfigureKeyProperty([NotNull] InternalPropertyBuilder propertyBuilder)
-        {
-            Check.NotNull(propertyBuilder, nameof(propertyBuilder));
-
-            propertyBuilder.GenerateValueOnAdd(true, ConfigurationSource.Convention);
-
             // TODO: Nullable, Sequence
             // Issue #213
         }
 
-        private static bool IsCommonInteger(Type type)
+        public virtual void Apply(InternalEntityBuilder entityBuilder, ForeignKey foreignKey)
         {
-            type = type.UnwrapNullableType();
+            Check.NotNull(entityBuilder, nameof(entityBuilder));
+            Check.NotNull(foreignKey, nameof(foreignKey));
 
-            return type == typeof(int)
-                   || type == typeof(long)
-                   || type == typeof(short)
-                   || type == typeof(byte);
+            ConfigureKeyProperties(entityBuilder, foreignKey.Properties);
         }
     }
 }
--- a/src/EntityFramework.Core/Metadata/ModelConventions/KeyDiscoveryConvention.cs
+++ b/src/EntityFramework.Core/Metadata/ModelConventions/KeyDiscoveryConvention.cs
@@ -1,0 +1,57 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Metadata.ModelConventions
+{
+    public class KeyDiscoveryConvention : IEntityTypeConvention
+    {
+        private const string KeySuffix = "Id";
+
+        public virtual InternalEntityBuilder Apply(InternalEntityBuilder entityBuilder)
+        {
+            Check.NotNull(entityBuilder, nameof(entityBuilder));
+            var entityType = entityBuilder.Metadata;
+
+            var keyProperties = DiscoverKeyProperties(entityType);
+            if (keyProperties.Count != 0)
+            {
+                entityBuilder.PrimaryKey(keyProperties.Select(p => p.Name).ToList(), ConfigurationSource.Convention);
+            }
+
+            return entityBuilder;
+        }
+
+        public virtual IReadOnlyList<Property> DiscoverKeyProperties([NotNull] EntityType entityType)
+        {
+            Check.NotNull(entityType, nameof(entityType));
+
+            // TODO: Honor [Key]
+            // Issue #213
+            var keyProperties = entityType.Properties
+                .Where(p => string.Equals(p.Name, KeySuffix, StringComparison.OrdinalIgnoreCase))
+                .ToList();
+
+            if (keyProperties.Count == 0)
+            {
+                keyProperties = entityType.Properties.Where(
+                    p => string.Equals(p.Name, entityType.SimpleName + KeySuffix, StringComparison.OrdinalIgnoreCase))
+                    .ToList();
+            }
+
+            if (keyProperties.Count > 1)
+            {
+                //TODO - add in logging using resource Strings.MultiplePropertiesMatchedAsKeys()
+                return new Property[0];
+            }
+
+            return keyProperties;
+        }
+    }
+}
--- a/src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerPropertyExtensions.cs
+++ b/src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerPropertyExtensions.cs
@@ -57,7 +57,10 @@ namespace Microsoft.Data.Entity.SqlServer.Metadata
             {
                 // TODO: Issue #777: Non-string annotations
                 var value = Property[SqlServerValueGenerationAnnotation];
-                return value == null ? null : (SqlServerValueGenerationStrategy?)Enum.Parse(typeof(SqlServerValueGenerationStrategy), value);
+                var strategy = value == null ? null : (SqlServerValueGenerationStrategy?)Enum.Parse(typeof(SqlServerValueGenerationStrategy), value);
+                return strategy == SqlServerValueGenerationStrategy.Default
+                    ? (Property.EntityType.Model.SqlServer().ValueGenerationStrategy ?? SqlServerValueGenerationStrategy.Identity)
+                    : strategy;
             }
         }
 
@@ -75,9 +78,7 @@ namespace Microsoft.Data.Entity.SqlServer.Metadata
         {
             var modelExtensions = Property.EntityType.Model.SqlServer();
 
-            if (ValueGenerationStrategy != SqlServerValueGenerationStrategy.Sequence
-                && (ValueGenerationStrategy != null
-                    || modelExtensions.ValueGenerationStrategy != SqlServerValueGenerationStrategy.Sequence))
+            if (ValueGenerationStrategy != SqlServerValueGenerationStrategy.Sequence)
             {
                 return null;
             }
--- a/src/EntityFramework.SqlServer/Metadata/SqlServerPropertyBuilder.cs
+++ b/src/EntityFramework.SqlServer/Metadata/SqlServerPropertyBuilder.cs
@@ -92,5 +92,23 @@ namespace Microsoft.Data.Entity.SqlServer.Metadata
 
             return this;
         }
+
+        public virtual SqlServerPropertyBuilder UseDefaultValueGeneration()
+        {
+            _property.SqlServer().ValueGenerationStrategy = SqlServerValueGenerationStrategy.Default;
+            _property.SqlServer().SequenceName = null;
+            _property.SqlServer().SequenceSchema = null;
+
+            return this;
+        }
+
+        public virtual SqlServerPropertyBuilder UseNoValueGeneration()
+        {
+            _property.SqlServer().ValueGenerationStrategy = null;
+            _property.SqlServer().SequenceName = null;
+            _property.SqlServer().SequenceSchema = null;
+
+            return this;
+        }
     }
 }
--- a/src/EntityFramework.SqlServer/Metadata/SqlServerValueGenerationStrategy.cs
+++ b/src/EntityFramework.SqlServer/Metadata/SqlServerValueGenerationStrategy.cs
@@ -5,6 +5,7 @@ namespace Microsoft.Data.Entity.SqlServer.Metadata
 {
     public enum SqlServerValueGenerationStrategy
     {
+        Default,
         Sequence,
         Identity
     }
--- a/src/EntityFramework.SqlServer/Migrations/SqlServerModelDiffer.cs
+++ b/src/EntityFramework.SqlServer/Migrations/SqlServerModelDiffer.cs
@@ -121,13 +121,7 @@ namespace Microsoft.Data.Entity.SqlServer.Migrations
         }
 
         // TODO: Move to metadata API?
-        // See Issue #1271: Principal keys need to generate values on add, but the database should only have one Identity column.
-        private SqlServerValueGenerationStrategy? GetValueGenerationStrategy(IProperty property) =>
-            property.SqlServer().ValueGenerationStrategy
-            ?? property.EntityType.Model.SqlServer().ValueGenerationStrategy
-            ?? (property.GenerateValueOnAdd && property.PropertyType.IsInteger() && property.IsPrimaryKey()
-                ? SqlServerValueGenerationStrategy.Identity
-                : default(SqlServerValueGenerationStrategy?));
+        private SqlServerValueGenerationStrategy? GetValueGenerationStrategy(IProperty property) => property.SqlServer().ValueGenerationStrategy;
 
         #endregion
 
--- a/src/EntityFramework.SqlServer/SqlServerModelBuilderFactory.cs
+++ b/src/EntityFramework.SqlServer/SqlServerModelBuilderFactory.cs
@@ -2,10 +2,25 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Metadata.ModelConventions;
+using Microsoft.Data.Entity.SqlServer.Metadata.ModelConventions;
 
 namespace Microsoft.Data.Entity.SqlServer
 {
     public class SqlServerModelBuilderFactory : ModelBuilderFactory, ISqlServerModelBuilderFactory
     {
+        protected override ConventionSet CreateConventionSet()
+        {
+            var conventions = base.CreateConventionSet();
+
+            var sqlServerValueGenerationStrategyConvention = new SqlServerValueGenerationStrategyConvention();
+            conventions.KeyAddedConventions.Add(sqlServerValueGenerationStrategyConvention);
+
+            conventions.ForeignKeyAddedConventions.Add(sqlServerValueGenerationStrategyConvention);
+
+            conventions.ForeignKeyRemovedConventions.Add(sqlServerValueGenerationStrategyConvention);
+
+            return conventions;
+        }
     }
 }
--- a/src/EntityFramework.SqlServer/SqlServerValueGeneratorSelector.cs
+++ b/src/EntityFramework.SqlServer/SqlServerValueGeneratorSelector.cs
@@ -38,8 +38,7 @@ namespace Microsoft.Data.Entity.SqlServer
         {
             Check.NotNull(property, nameof(property));
 
-            var strategy = property.SqlServer().ValueGenerationStrategy
-                           ?? property.EntityType.Model.SqlServer().ValueGenerationStrategy;
+            var strategy = property.SqlServer().ValueGenerationStrategy;
 
             if (property.PropertyType.IsInteger()
                 && strategy == SqlServerValueGenerationStrategy.Sequence)
--- a/test/EntityFramework.Core.FunctionalTests/TestModels/MonsterContext`.cs
+++ b/test/EntityFramework.Core.FunctionalTests/TestModels/MonsterContext`.cs
@@ -218,42 +218,17 @@ namespace Microsoft.Data.Entity.FunctionalTests.TestModels
 
             modelBuilder.Entity<TBarcodeDetail>().Key(e => e.Code);
 
-            // TODO: Key should not get by-convention value generation if it is dependent of identifying relationship
-            modelBuilder.Entity<TResolution>()
-                .Property(e => e.ResolutionId)
-                .GenerateValueOnAdd(false);
-
             modelBuilder.Entity<TSuspiciousActivity>();
             modelBuilder.Entity<TLastLogin>().Key(e => e.Username);
             modelBuilder.Entity<TMessage>().Key(e => new { e.MessageId, e.FromUsername });
 
-            modelBuilder.Entity<TOrderNote>(b =>
-                {
-                    b.Key(e => e.NoteId);
-                    // TODO: Key should get by-convention value generation even if key is not discovered by convention
-                    b.Property(e => e.NoteId).GenerateValueOnAdd();
-                });
+            modelBuilder.Entity<TOrderNote>().Key(e => e.NoteId);
 
             modelBuilder.Entity<TProductDetail>().Key(e => e.ProductId);
 
-            modelBuilder.Entity<TProductWebFeature>(b =>
-                {
-                    b.Key(e => e.FeatureId);
-                    // TODO: Key should get by-convention value generation even if key is not discovered by convention
-                    b.Property(e => e.FeatureId).GenerateValueOnAdd();
-                });
+            modelBuilder.Entity<TProductWebFeature>().Key(e => e.FeatureId);
 
             modelBuilder.Entity<TSupplierLogo>().Key(e => e.SupplierId);
-
-            // TODO: Key should not get by-convention value generation if it is dependent of identifying relationship
-            modelBuilder.Entity<TCustomerInfo>()
-                .Property(e => e.CustomerInfoId)
-                .GenerateValueOnAdd(false);
-
-            // TODO: Key should not get by-convention value generation if it is dependent of identifying relationship
-            modelBuilder.Entity<TComputerDetail>()
-                .Property(e => e.ComputerDetailId)
-                .GenerateValueOnAdd(false);
 
             modelBuilder.Entity<TLicense>().Key(e => e.Name);
 
@@ -287,6 +262,7 @@ namespace Microsoft.Data.Entity.FunctionalTests.TestModels
             modelBuilder.Entity<TOrderLine>(b =>
                 {
                     b.Key(e => new { e.OrderId, e.ProductId });
+
                     b.HasOne(e => (TProduct)e.Product).WithMany().ForeignKey(e => e.ProductId);
                 });
 
@@ -310,30 +286,23 @@ namespace Microsoft.Data.Entity.FunctionalTests.TestModels
 
                     b.HasOne(e => (TResolution)e.Resolution).WithOne(e => (TComplaint)e.Complaint)
                         .ReferencedKey<TComplaint>(e => e.AlternateId);
-
-                    // Should not be needed: see Issue #1271
-                    b.Property(e => e.AlternateId).GenerateValueOnAdd();
                 });
 
             modelBuilder.Entity<TProductPhoto>(b =>
                 {
-                    b.Key(e => new { e.ProductId, e.PhotoId });
-                    // TODO: Key should get by-convention value generation even if key is not discovered by convention
-                    b.Property(e => e.PhotoId).GenerateValueOnAdd();
+                    b.Key(e => new { e.PhotoId, e.ProductId });
 
                     b.HasMany(e => (IEnumerable<TProductWebFeature>)e.Features).WithOne(e => (TProductPhoto)e.Photo)
-                        .ForeignKey(e => new { e.ProductId, e.PhotoId })
-                        .ReferencedKey(e => new { e.ProductId, e.PhotoId });
+                        .ForeignKey(e => new { e.PhotoId, e.ProductId })
+                        .ReferencedKey(e => new { e.PhotoId, e.ProductId });
                 });
 
             modelBuilder.Entity<TProductReview>(b =>
                 {
-                    b.Key(e => new { e.ProductId, e.ReviewId });
-                    // TODO: Key should get by-convention value generation even if key is not discovered by convention
-                    b.Property(e => e.ReviewId).GenerateValueOnAdd();
+                    b.Key(e => new { e.ReviewId, e.ProductId });
 
                     b.HasMany(e => (IEnumerable<TProductWebFeature>)e.Features).WithOne(e => (TProductReview)e.Review)
-                        .ForeignKey(e => new { e.ProductId, e.ReviewId });
+                        .ForeignKey(e => new { e.ReviewId, e.ProductId });
                 });
 
             modelBuilder.Entity<TLogin>(b =>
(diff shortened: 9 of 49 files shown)


test/EntityFramework.SqlServer.FunctionalTests/GraphUpdatesSqlServerTest.cs
```diff
@@ -20,7 +20,25 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+                modelBuilder.Entity<Required1>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<Required2>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<RequiredAk1>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<RequiredAk2>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<RequiredNonPkSingle1>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<RequiredNonPkSingle2>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<RequiredNonPkSingleAk1>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<RequiredNonPkSingleAk2>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<RequiredSingleAk1>().Property(e => e.Id).ForSqlServer().UseSequence();
+                modelBuilder.Entity<RequiredSingleAk2>().Property(e => e.Id).ForSqlServer().UseSequence();
```
ajcvickers: It seems unfortunate that we have to do this now. Can we think about this some more and see if we can come up with a better experience?

test/EntityFramework.Core.Tests/ChangeTracking/Internal/ChangeDetectorTest.cs
```diff
@@ -1801,20 +1801,25 @@ private static IModel BuildModel()
...
-            builder.Entity<Product>()
-                .HasOne(e => e.Tag).WithOne(e => e.Product)
-                .ReferencedKey<Product>(e => e.TagId)
-                .ForeignKey<ProductTag>(e => e.ProductId);
+            builder.Entity<Product>(b =>
+                {
+                    b.HasOne(e => e.Tag).WithOne(e => e.Product)
+                        .ReferencedKey<Product>(e => e.TagId)
+                        .ForeignKey<ProductTag>(e => e.ProductId);
+                    b.Property(e => e.TagId).GenerateValueOnAdd(false);
```
ajcvickers: What is it about TagId that means it should not have generate values on add set? And why is it that the conventions don't pick up on whatever this is and not set it?

ajcvickers: Likewise for other properties here where we are changing what the convention does.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Metadata.ModelConventions;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.SqlServer.Metadata.ModelConventions
+{
+    public class SqlServerKeyConvention : IKeyConvention
```
ajcvickers: I'm not sure that composite keys should get one property set as SQL Server Identity by default. Also, the fact that we have some essentially arbitrary rules that kick in to determine which property to use seems like not a great idea. It might be better to not set any property as Identity by convention and to let the app do this explicitly to one property if this is what the app needs.

Also, we want to switch over to make Sequences the default at some point, and we may be ready to do that now because Migrations support was holding us back, but that should now work. (We also may need a story for SQL Server versions that don't support sequences.) If using sequences becomes the default, then it's not clear when this would be useful. Also, multiple columns could use the sequence if needed, but I'm not sure we should do that by convention either.

We should discuss this with @divega.

ajcvickers:
Ideally, no application should need to switch GenerateValueOnAdd off, so if we are having to do this in our test code, then we should understand why. The main reason for this is that even if GenerateValueOnAdd is on, then a value won't be generated if the user explicitly specifies a value before adding.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
AndriySvyryd: I'm ok with not doing anything to composite keys for now.

test/EntityFramework.Core.Tests/ChangeTracking/Internal/ChangeDetectorTest.cs
smitpatel: In the tests we are not assigning the values for TagId, Because of GenerateValueOnAdd being true the value generation assigns a value. This notifies PrincipalKeyChange. But the tests expect it to be null.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
divega: +1 It seems better to me not to try to be smart about value generation for composite key members.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        private const string IdentityKeySuffix = "Id";
+        public virtual InternalKeyBuilder Apply(InternalKeyBuilder keyBuilder)
+        {
+            Check.NotNull(keyBuilder, nameof(keyBuilder));
+
+            var key = keyBuilder.Metadata;
+            var properties = key.Properties;
+
+            if (key == key.EntityType.TryGetPrimaryKey())
```
AndriySvyryd: You can use key.IsPrimaryKey

test/EntityFramework.SqlServer.FunctionalTests/GraphUpdatesSqlServerTest.cs
AndriySvyryd: I believe the last time we discussed this we decided that this case would be handled by 'lightweight conventions':

                foreach (var entityType in modelBuilder.Metadata.EntityTypes)
                {
                    var key = entityType.GetPrimaryKey();
                    if (key.Properties.Count == 1)
                    {
                        var keyProperty = key.Properties.Single();
                        modelBuilder.Entity(entityType.Name)
                            .Property(keyProperty.PropertyType, keyProperty.Name)
                            .ForSqlServer()
                            .UseSequence();
                    }
                }

AndriySvyryd:
:shipit: pending signoff from Arthur

test/EntityFramework.SqlServer.FunctionalTests/GraphUpdatesSqlServerTest.cs
ajcvickers: We have a convention that sets Identity on a primary key column if neither Identity or Sequence is explicitly set. (As noted in previous comments, this should be simplified so that it only sets Identity for single integer keys, not composites.) Can we not change this convention to set Identity or Sequence based on what is annotated at the model level?

test/EntityFramework.Core.Tests/ChangeTracking/Internal/ChangeDetectorTest.cs
ajcvickers: That makes sense given that this is a unit test for this specific behavior. Thanks.

test/EntityFramework.SqlServer.FunctionalTests/CompositeKeyEndToEndTest.cs
```diff
@@ -201,22 +201,15 @@ protected override void OnConfiguring(DbContextOptions builder)
 
             protected override void OnModelCreating(ModelBuilder modelBuilder)
             {
-                modelBuilder.Entity<Pegasus>().Key(e => new { e.Id1, e.Id2 });
+                modelBuilder.Entity<Pegasus>(b =>
+                    {
+                        b.Key(e => new { e.Id1, e.Id2 });
+                        b.Property(e => e.Id1).GenerateValueOnAdd(false);
+                        b.Property(e => e.Id2).GenerateValueOnAdd(false);
```
ajcvickers: Why do we need to switch GenerateValueOnAdd off here?

test/EntityFramework.Core.FunctionalTests/TestModels/MonsterContext`.cs
```diff
@@ -217,43 +217,18 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
             modelBuilder.Entity<TBarcodeDetail>().Key(e => e.Code);
 
-            // TODO: Key should not get by-convention value generation if it is dependent of identifying relationship
-            modelBuilder.Entity<TResolution>()
-                .Property(e => e.ResolutionId)
-                .GenerateValueOnAdd(false);
```
ajcvickers: Why do we need to switch GenerateValueOnAdd off in all the cases here?

test/EntityFramework.Core.FunctionalTests/NullKeysTestBase.cs
```diff
@@ -192,15 +192,24 @@ public virtual void OnModelCreating(ModelBuilder modelBuilder)
...
 
-                modelBuilder.Entity<WithNullableIntKey>()
-                    .HasMany(e => e.Dependents)
-                    .WithOne(e => e.Principal)
-                    .ForeignKey(e => e.Fk);
+                modelBuilder.Entity<WithNullableIntKey>(b =>
+                    {
+                        b.HasMany(e => e.Dependents)
+                            .WithOne(e => e.Principal)
+                            .ForeignKey(e => e.Fk);
```
ajcvickers: Why do we need to switch GenerateValueOnAdd off in these cases?

test/EntityFramework.Core.FunctionalTests/TestModels/MonsterContext`.cs
```diff
@@ -286,6 +261,8 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<TOrderLine>(b =>
                 {
                     b.Key(e => new { e.OrderId, e.ProductId });
+                    b.Property(e => e.OrderId).GenerateValueOnAdd(false);
```
ajcvickers: How about here?

test/EntityFramework.SqlServer.FunctionalTests/GraphUpdatesSqlServerTest.cs
AndriySvyryd: The convention runs before the model is annotated.

Sent from my Windows Phone
________________________________

ajcvickers: Does that mean we need a convention that runs when the model is annotated so that changes to annotations can be responded to?

AndriySvyryd: We can do that, but it's much more granular than any conventions we have now or expect to have in near future.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
smitpatel: After discussing with @AndriySvyryd  & @ajcvickers , 
we decided that we won't do anything for composite key. 
For PK with only 1 property and if that property is common integer type then we will add Default as value generation strategy.
Whenever SqlServerValueGenerationStrategy.Default is present, we use Identity only if there is no model level value generation strategy assigned.
In the absence of any input from what will be the value generation strategy, we use Identity.

test/EntityFramework.Core.FunctionalTests/TestModels/MonsterContext`.cs
smitpatel: This was old code where we had to turn off the value generation for dependent end of relationship.
Removed now.

smitpatel:
The previous KeyConvetion has been renamed to KeyDiscoveryConvention since it discovers the key properties. This new convention works on keys so it has been put in KeyConvention.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        protected virtual void ConfigureDefaultValueGenerationOnProperty([NotNull] InternalPropertyBuilder propertyBuilder)
+        {
+            Check.NotNull(propertyBuilder, nameof(propertyBuilder));
+
+            propertyBuilder.Annotation(
+                SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.ValueGeneration,
+                SqlServerValueGenerationStrategy.Default.ToString(),
+                ConfigurationSource.Convention);
+        }
```
ajcvickers: It's unfortunate that we don't use the sugar API here and instead restore to building annotation names and values manually. One of the advantages of the sugar API is that it creates a single place where the annotation name building happens. Can we investigate why the sugar API can't be used here? Maybe @AndriySvyryd knows?

src/EntityFramework.SqlServer/Migrations/SqlServerModelDiffer.cs
```diff
@@ -121,13 +121,21 @@ protected override IEnumerable<MigrationOperation> Add(IProperty target)
...
+                switch (property.SqlServer().ValueGenerationStrategy)
+                {
+                    case null:
+                        return property.EntityType.Model.SqlServer().ValueGenerationStrategy ?? default(SqlServerValueGenerationStrategy?);
+                    case SqlServerValueGenerationStrategy.Default:
+                        return property.EntityType.Model.SqlServer().ValueGenerationStrategy ?? SqlServerValueGenerationStrategy.Identity;
+                }
+                return property.SqlServer().ValueGenerationStrategy;
+            }
+            return default(SqlServerValueGenerationStrategy?);
```
ajcvickers: This doesn't seem right to me. It seems like if a property has GenerateValueOnAdd, but doesn't have a ValueGenerationStrategy set, then it will use the generation strategy set on the model. I think it should only use the strategy set on the model if the property is annotated with Default.

Another way of thinking about this is that the SQL Server model building conventions are now entirely responsible for determining value generation strategy for each property. Migrations doesn't even need to look at GenerateValueOnAdd anymore. It just looks if the property is annotated. If it is, then it either sets that strategy directly, or if it is annotated as default it will instead use the strategy set on the model. If no strategy is set on the model it will use the overall default--currently Identity, soon to be Sequence. However, it might would be good to remove this last set and annotate the model in every case. Then if we change the default in the future existing models will continue to work without any additional work.

src/EntityFramework.SqlServer/SqlServerValueGeneratorSelector.cs
```diff
@@ -38,8 +38,9 @@ public override ValueGenerator Select(IProperty property)
         {
             Check.NotNull(property, nameof(property));
 
-            var strategy = property.SqlServer().ValueGenerationStrategy
-                           ?? property.EntityType.Model.SqlServer().ValueGenerationStrategy;
+            var strategy = property.SqlServer().ValueGenerationStrategy == SqlServerValueGenerationStrategy.Default
+                ? (property.EntityType.Model.SqlServer().ValueGenerationStrategy ?? SqlServerValueGenerationStrategy.Identity)
+                : property.SqlServer().ValueGenerationStrategy ?? property.EntityType.Model.SqlServer().ValueGenerationStrategy;
```
ajcvickers: If possible, factor the code here and the code used in Migrations into a common place, since in both places it is about finding the effective strategy for a given property.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
AndriySvyryd: There is no sugar API for internal model builders. We can add it if you think it will add value.

src/EntityFramework.SqlServer/Migrations/SqlServerModelDiffer.cs
smitpatel: For composite PK, we don't set any annotation by conventions. What is expected behavior in such cases if user doesn't specify strategy. Presently all tests which have composite PK, uses Identity so it has been explicitly specified.

When a new key is added in SqlServer, GenerateValueOnAdd and annotation both are set. But if the key later becomes dependent end of FK then we set GenerateValue false but annotation still remains there. Therefore we are making check on GenerateValue.

ajcvickers: 1. The expected behavior is that if the property should use Identity, then that must be specified explicitly using the fluent API.

2. Annotation should be removed if it was added by convention.

src/EntityFramework.SqlServer/SqlServerValueGeneratorSelector.cs
smitpatel: Possible place to put method for refactoring?

ajcvickers: Same place ValueGenerationStrategy property is.

test/EntityFramework.SqlServer.Tests/SqlServerValueGeneratorSelectorTest.cs
```diff
@@ -45,37 +45,37 @@ public void Returns_built_in_generators_for_types_setup_for_value_generation()
...
 
-        [Fact]
-        public void Returns_sequence_value_generators_when_configured_for_model()
-        {
-            var model = BuildModel();
-            model.SqlServer().ValueGenerationStrategy = SqlServerValueGenerationStrategy.Sequence;
-            var entityType = model.GetEntityType(typeof(AnEntity));
+        //[Fact]
+        //public void Returns_sequence_value_generators_when_configured_for_model()
+        //{
```
smitpatel: @ajcvickers - This test fails because we use model annotation only if Default is applied to property.
Remove/modify?

src/EntityFramework.SqlServer/SqlServerValueGeneratorSelector.cs
smitpatel: Done.

src/EntityFramework.SqlServer/Migrations/SqlServerModelDiffer.cs
smitpatel: Updated.
@AndriySvyryd - There are changed in conventions related to FK added and removed. Can you review again?

src/EntityFramework.SqlServer/Metadata/SqlServerValueGenerationStrategy.cs
```diff
@@ -6,6 +6,7 @@ namespace Microsoft.Data.Entity.SqlServer.Metadata
     public enum SqlServerValueGenerationStrategy
     {
         Sequence,
-        Identity
+        Identity,
+        Default
```
AndriySvyryd: Default should be first

src/EntityFramework.SqlServer/Migrations/SqlServerModelDiffer.cs
```diff
@@ -121,13 +121,14 @@ protected override IEnumerable<MigrationOperation> Add(IProperty target)
...
-            property.SqlServer().ValueGenerationStrategy
-            ?? property.EntityType.Model.SqlServer().ValueGenerationStrategy
-            ?? (property.GenerateValueOnAdd && property.PropertyType.IsInteger() && property.IsPrimaryKey()
-                ? SqlServerValueGenerationStrategy.Identity
-                : default(SqlServerValueGenerationStrategy?));
+        private SqlServerValueGenerationStrategy? GetValueGenerationStrategy(IProperty property)
+        {
+            if (property.GenerateValueOnAdd)
+            {
+                return SqlServerModelExtensions.GetSqlServerValueGenerationStrategy(property);
```
AndriySvyryd: Since the convention will remove the annotation you don't need to check GenerateValueOnAdd here

test/EntityFramework.SqlServer.Tests/SqlServerValueGeneratorSelectorTest.cs
AndriySvyryd: Modify by applying Default to all properties. Also add another test that asserts that model annotation is not returned if property annotation is not set.

AndriySvyryd:
Add a SqlServer-specific ModelBuilder test that asserts that the convention removes the strategy annotations on preprties that are included in a foreign key.

src/EntityFramework.Core/Metadata/ModelConventions/KeyConvention.cs
```diff
@@ -1,82 +1,55 @@
...
-            var entityType = entityBuilder.Metadata;
+            Check.NotNull(keyBuilder, nameof(keyBuilder));
+
+            var entityBuilder = keyBuilder.ModelBuilder.Entity(keyBuilder.Metadata.EntityType.Name, ConfigurationSource.Convention);
+            var properties = keyBuilder.Metadata.Properties;
 
-            var keyProperties = DiscoverKeyProperties(entityType);
-            if (keyProperties.Count != 0
-                && entityBuilder.PrimaryKey(keyProperties.Select(p => p.Name).ToList(), ConfigurationSource.Convention) != null)
+            if (entityBuilder.Metadata.TryGetForeignKey(properties) == null)
```
AndriySvyryd: This condition will return true for a composite key where a foreign key is defined on only a subset of properties or a key whose properties are a subset of foreign key properties. Change the check to account for this and add tests.

src/EntityFramework.Core/Metadata/ModelConventions/KeyConvention.cs
```diff
@@ -1,82 +1,55 @@
...
+            Check.NotNull(entityBuilder, nameof(entityBuilder));
+            Check.NotNull(foreignKey, nameof(foreignKey));
+
+            var properties = foreignKey.Properties;
 
-            return type == typeof(int)
-                   || type == typeof(long)
-                   || type == typeof(short)
-                   || type == typeof(byte);
+            if (entityBuilder.Metadata.TryGetForeignKey(properties) == null)
```
AndriySvyryd: Same as above. Extract check to a method to avoid duplicate logic.

src/EntityFramework.Core/Metadata/ModelConventions/KeyDiscoveryConvention.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                keyProperties = entityType.Properties.Where(
+                    p => string.Equals(p.Name, entityType.SimpleName + KeySuffix, StringComparison.OrdinalIgnoreCase))
+                    .ToList();
+            }
+
+            if (keyProperties.Count > 1)
+            {
+                //TODO - add in logging using resource Strings.MultiplePropertiesMatchedAsKeys()
+                return Enumerable.Empty<Property>().ToList();
```
AndriySvyryd: This isn't related to your changes, but change this line to return new Property[0]

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerForeignKeyAddedConvention.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Metadata.ModelConventions;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.SqlServer.Metadata.ModelConventions
+{
+    public class SqlServerForeignKeyAddedConvention : IRelationshipConvention
```
AndriySvyryd: Merge with SqlServerKeyConvention and rename to SqlServerValueGenerationStrategyConvention

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerForeignKeyAddedConvention.cs
```diff
@@ -0,0 +1,37 @@
...
+                if (property.GenerateValueOnAdd == false)
+                {
+                    var entityBuilder = relationshipBuilder.ModelBuilder.Entity(property.EntityType.Name, ConfigurationSource.Convention);
+                    RemoveValueGenerationStrategyAnnotaion(entityBuilder.Property(property.PropertyType, property.Name, ConfigurationSource.Convention));
+                }
+            }
+            return relationshipBuilder;
+        }
+
+        protected virtual void RemoveValueGenerationStrategyAnnotaion([NotNull] InternalPropertyBuilder propertyBuilder)
```
AndriySvyryd: typo in 'Annotaion'

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public class SqlServerKeyConvention : IKeyConvention, IForeignKeyRemovedConvention
+    {
+        public virtual InternalKeyBuilder Apply(InternalKeyBuilder keyBuilder)
+        {
+            Check.NotNull(keyBuilder, nameof(keyBuilder));
+
+            var key = keyBuilder.Metadata;
+
+            if (key.IsPrimaryKey() && key.Properties.Count == 1 && key.Properties.First().PropertyType.IsInteger())
```
AndriySvyryd: Also need to check that ValueGeneratedOnAdd is true

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerKeyConvention.cs
```diff
@@ -0,0 +1,58 @@
...
+            Check.NotNull(entityBuilder, nameof(entityBuilder));
+            Check.NotNull(foreignKey, nameof(foreignKey));
+
+            var properties = foreignKey.Properties;
+
+            if (entityBuilder.Metadata.TryGetForeignKey(properties) == null
+                && entityBuilder.Metadata.TryGetPrimaryKey(properties) != null
+                && properties.Count == 1
+                && properties.First().PropertyType.IsInteger())
+            {
```
AndriySvyryd: No need to check for existing foreign keys, just that ValueGeneratedOnAdd is true. Extract logic to avoid duplication.

src/EntityFramework.SqlServer/Metadata/SqlServerModelExtensions.cs
```diff
@@ -72,5 +72,14 @@ public virtual Sequence GetOrAddSequence([CanBeNull] string name = null, [CanBeN
             return ((Model)Model).SqlServer().TryGetSequence(name, schema)
                    ?? AddOrReplaceSequence(new Sequence(name, schema));
         }
+
+        internal static SqlServerValueGenerationStrategy? GetSqlServerValueGenerationStrategy([NotNull] IProperty property)
```
ajcvickers: This should be public.

ajcvickers: And should be defined on ReadOnlySqlServerModelExtensions because it is only about inspecting metadata, not changing it.

src/EntityFramework.SqlServer/Metadata/SqlServerModelExtensions.cs
```diff
@@ -72,5 +72,14 @@ public virtual Sequence GetOrAddSequence([CanBeNull] string name = null, [CanBeN
...
+        {
+            Check.NotNull(property, nameof(property));
+
+            return property.SqlServer().ValueGenerationStrategy == SqlServerValueGenerationStrategy.Default
+                ? (property.EntityType.Model.SqlServer().ValueGenerationStrategy ?? SqlServerValueGenerationStrategy.Identity)
```
ajcvickers: Can you make the change to always put the default value generation strategy in the model rather than hard code it here? Or file another issue to do so if you prefer.

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerPropertyExtensions.cs
```diff
@@ -76,7 +76,7 @@ public virtual Sequence TryGetSequence()
             var modelExtensions = Property.EntityType.Model.SqlServer();
 
             if (ValueGenerationStrategy != SqlServerValueGenerationStrategy.Sequence
-                && (ValueGenerationStrategy != null
+                && ((ValueGenerationStrategy != null && ValueGenerationStrategy != SqlServerValueGenerationStrategy.Default)
```
ajcvickers: Can you use GetSqlServerValueGenerationStrategy here?

AndriySvyryd: This would fix the logic as currently it will return the model strategy if property strategy is null

test/EntityFramework.SqlServer.Tests/SqlServerValueGeneratorSelectorTest.cs
ajcvickers: Consider using the SQL Server conventions with the model builder. This will then test that given the SQL Server conventions and the model annotated with Sequence, the correct generators are used for the correct properties.

src/EntityFramework.SqlServer/SqlServerModelBuilderFactory.cs
```diff
@@ -2,10 +2,24 @@
...
     {
+        protected override ConventionSet CreateConventionSet()
+        {
+            var conventions = base.CreateConventionSet();
+
+            conventions.KeyAddedConventions.Add(new SqlServerKeyConvention());
+
+            conventions.ForeignKeyAddedConventions.Add(new SqlServerForeignKeyAddedConvention());
+
+            conventions.ForeignKeyRemovedConventions.Add(new SqlServerKeyConvention());
```
AndriySvyryd: Memory: add the same instance

test/EntityFramework.SqlServer.Tests/SqlServerValueGeneratorSelectorTest.cs
AndriySvyryd: @ajcvickers I think that would be a different, functional test that I mentioned in a previous comment. These unit tests shouldn't depend on correctness of the conventions.

ajcvickers: @AndriySvyryd I'm fine with that so long as the functional test is added.

test/EntityFramework.SqlServer.FunctionalTests/BatchingTest.cs
```diff
@@ -46,7 +47,8 @@ public BloggingContext(IServiceProvider serviceProvider, DbContextOptions option
 
             protected override void OnModelCreating(ModelBuilder modelBuilder)
             {
-                modelBuilder.Entity<Blog>().Property(b => b.Id).GenerateValueOnAdd(generateValue: false);
+                modelBuilder.Entity<Blog>().Property(b => b.Id).GenerateValueOnAdd(generateValue: false)
+                    .Metadata.RemoveAnnotation(new Annotation(SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.ValueGeneration, "_"));
```
AndriySvyryd: We should add sugar API for removing this annotation

test/EntityFramework.SqlServer.Tests/SqlServerValueGeneratorCacheTest.cs
```diff
@@ -105,6 +107,7 @@ public void Block_size_is_obtained_from_named_model_default_sequence()
                 .Entity<Robot>()
                 .Property(e => e.Id)
                 .GenerateValueOnAdd()
+                .Annotation(SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.ValueGeneration, SqlServerValueGenerationStrategy.Default.ToString())
```
AndriySvyryd: You should be able to use the sugar API in these cases.

AndriySvyryd:
:shipit:

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerModelExtensions.cs
```diff
@@ -68,5 +68,14 @@ public override Sequence TryGetSequence(string name, string schema = null)
             return FindSequence(SqlServerSequenceAnnotation + schema + "." + name)
                    ?? base.TryGetSequence(name, schema);
         }
+
+        public static SqlServerValueGenerationStrategy? GetSqlServerValueGenerationStrategy([NotNull] IProperty property)
+        {
+            Check.NotNull(property, nameof(property));
+
+            return property.SqlServer().ValueGenerationStrategy == SqlServerValueGenerationStrategy.Default
+                ? (property.EntityType.Model.SqlServer().ValueGenerationStrategy ?? SqlServerValueGenerationStrategy.Identity)
```
ajcvickers: Did you file an issue to not return Identity by default but instead always put it in the model?

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerPropertyExtensions.cs
```diff
@@ -76,7 +76,7 @@ public virtual Sequence TryGetSequence()
             var modelExtensions = Property.EntityType.Model.SqlServer();
 
             if (ValueGenerationStrategy != SqlServerValueGenerationStrategy.Sequence
-                && (ValueGenerationStrategy != null
+                && (ValueGenerationStrategy != SqlServerValueGenerationStrategy.Default
                     || modelExtensions.ValueGenerationStrategy != SqlServerValueGenerationStrategy.Sequence))
```
ajcvickers: Can you use GetSqlServerValueGenerationStrategy here?

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerModelExtensions.cs
```diff
@@ -68,5 +68,14 @@ public override Sequence TryGetSequence(string name, string schema = null)
             return FindSequence(SqlServerSequenceAnnotation + schema + "." + name)
                    ?? base.TryGetSequence(name, schema);
         }
+
+        public static SqlServerValueGenerationStrategy? GetSqlServerValueGenerationStrategy([NotNull] IProperty property)
```
ajcvickers: Why is this a static method here instead of an instance method on the property extensions?

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerModelExtensions.cs
smitpatel: filed #1811

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerModelExtensions.cs
smitpatel: moved to property extension ValueGenerationStrategy property

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerPropertyExtensions.cs
smitpatel: Simplified.

smitpatel:
Updated with feedback.

src/EntityFramework.SqlServer/Metadata/SqlServerPropertyBuilder.cs
```diff
@@ -92,5 +92,23 @@ public virtual SqlServerPropertyBuilder UseIdentity()
 
             return this;
         }
+
+        public virtual SqlServerPropertyBuilder UseDefault()
```
ajcvickers: Can't call this UseDefault because there is no indication as to what "default" this is. Change to UseDefaultValueGeneration for now, but we should review these names in API review and also review whether to change to just one method with a parameter.

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerPropertyExtensions.cs
```diff
@@ -57,7 +57,10 @@ public override object DefaultValue
             {
                 // TODO: Issue #777: Non-string annotations
                 var value = Property[SqlServerValueGenerationAnnotation];
-                return value == null ? null : (SqlServerValueGenerationStrategy?)Enum.Parse(typeof(SqlServerValueGenerationStrategy), value);
+                var strategy = value == null ? null : (SqlServerValueGenerationStrategy?)Enum.Parse(typeof(SqlServerValueGenerationStrategy), value);
```
ajcvickers: Consider using null propagation operator.

src/EntityFramework.SqlServer/Metadata/ReadOnlySqlServerModelExtensions.cs
```diff
@@ -68,5 +68,5 @@ public override Sequence TryGetSequence(string name, string schema = null)
             return FindSequence(SqlServerSequenceAnnotation + schema + "." + name)
                    ?? base.TryGetSequence(name, schema);
         }
-    }
+        }
```
ajcvickers: Fix whitespace

ajcvickers:
:shipit:

PR #1743, opened 2015-03-03
Stop resolving most value generator factories from DI

ajcvickers:
The selector is resolved from DI and is the appropriate place to make a change to which generator is used. Resolving most of the factories from DI does not add value.


anpete:
:shipit:

PR #1744, opened 2015-03-03
[TINY] Updating when we issue Table() & ColumnType() fluent APIs.

lajones:
See issue #1668.

Minor update so that we don't attempt to use Table() API if reported Schema is null.


bricelam:
:shipit:

lajones:
Checked in - commit 9333def.

PR #1745, opened 2015-03-03
Fix to #1726 - using boolProperty.Equals(true) in filter produces invalid sql

maumar:
Fix is to recognize this pattern in EqualsTranslator and compensate accordingly for boolean arguments.


maumar:
@anpete

Issue #1700, opened 2015-02-25
SqlServer: unexpected sql query for OrderBy with condition


tstar:
I have model:
```
public DbSet<Test> Tests { get; set; }

public class Test
{
	public byte Id { get; set; }
	public byte? Weight { get; set; }
}
```
Id - primary key.

I want to sort by column Weight with the null values at the end. For this, I planned to use the following code:
```
var result = Tests.OrderBy(i => i.Weight ?? byte.MaxValue).ToArray();
```
But the result was sorted without my conditions. EF7 sends to the server the following query:
```
SELECT [i].[Id], [i].[Weight]
FROM [dbo].[Test] AS [i]
ORDER BY [i].[Weight]
```
I tried different ways to replace this code:
```
var result = Tests.OrderBy(i => i.Weight == null ? 255 : i.Weight ).ToArray();
var result = Tests.Select(i => new { i.Id, Weight = i.Weight ?? 255 }).OrderBy(i => i.Weight).ToArray();
```
But did not achieve the desired result - sorting with the condition on the server side.

lukewaters:
Fixed in #1749

PR #1749, opened 2015-03-04
Added test and model for null Coalesce and Ternary operators in OrderBy

lukewaters:


src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -388,37 +393,49 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
 
                 string op;
 
-                switch (binaryExpression.NodeType)
+                var rightConstant = binaryExpression.Right as ConstantExpression;
+                if (rightConstant != null
+                    && rightConstant.Value == null)
+                {
+                    op = " IS ";
```
lajones: It seems like in here you should check whether binaryExpression.NodeType is Equal or NotEqual and output op = " IS " or op = " IS NOT " appropriately? (Any other NodeType being an error).

lukewaters: This actually looks like it is a bigger problem that I will address, it looks like we already have logic to deal with null comparisons, I just don't have my expressions in the right form.  Will fix

lukewaters:
@anpete can you take a look?

test/EntityFramework.SqlServer.FunctionalTests/QueryBugsTest.cs
```diff
@@ -352,6 +352,159 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             }
         }
 
+        [Fact]
+        public void Query_should_gen_coalesce_in_orderby_1700()
```
maumar: add tests for projection as well as its pretty common scenario

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
```diff
@@ -203,12 +203,15 @@ var innerJoinSelectExpression
...
 
-                    foreach (var columnExpression
+                    foreach (var expression
                         in innerJoinSelectExpression.OrderBy
-                            .Select(o => o.Expression)
-                            .Cast<ColumnExpression>())
+                            .Select(o => o.Expression))
                     {
-                        innerJoinSelectExpression.AddToProjection(columnExpression);
+                        var columnExpression = expression as ColumnExpression;
```
anpete: I don't think this works in general. Add a test that Includes a collection where the outer query has a OrderBy with a non-ColumnExpression.

test/EntityFramework.SqlServer.FunctionalTests/QueryBugsTest.cs
```diff
@@ -352,6 +352,159 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             }
         }
 
+        [Fact]
```
anpete: Add these tests to QueryTestBase and use Northwind.

test/EntityFramework.SqlServer.FunctionalTests/QueryBugsTest.cs
```diff
@@ -352,6 +352,159 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+        public void Query_should_gen_coalesce_in_orderby_1700()
+        {
+            CreateDatabase1700();
+
+            var loggingFactory = new TestSqlLoggerFactory();
```
anpete: Never do all of this in the test - this is why we have fixtures.

src/EntityFramework.Relational/Utilities/ExpressionExtensions.cs
```diff
@@ -16,7 +16,8 @@ public static bool IsLogicalOperation([NotNull] this Expression expression)
             Check.NotNull(expression, nameof(expression));
 
             return expression.NodeType == ExpressionType.AndAlso
-                   || expression.NodeType == ExpressionType.OrElse;
+                   || expression.NodeType == ExpressionType.OrElse
+                   || expression.NodeType == ExpressionType.Coalesce;
```
anpete: This method is really only for the conditional logical operators, so remove Coalesce.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalOrderingExpressionTreeVisitor.cs
```diff
@@ -27,44 +32,143 @@ private new RelationalQueryModelVisitor QueryModelVisitor
             get { return (RelationalQueryModelVisitor)base.QueryModelVisitor; }
         }
 
+        protected override Expression VisitConditionalExpression(ConditionalExpression conditionalExpression)
+        {
+            var selectExpression = ((RelationalQueryModelVisitor)base.QueryModelVisitor)
+                .GetSelectExpressions().First();
+
+            if (IsValidOrderbyExpression(conditionalExpression.NodeType))
```
anpete: Use a switch here.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalOrderingExpressionTreeVisitor.cs
```diff
@@ -20,6 +24,7 @@ public class RelationalOrderingExpressionTreeVisitor : DefaultQueryExpressionTre
             Check.NotNull(ordering, nameof(ordering));
 
             _ordering = ordering;
+            _adding = false;
```
anpete: Redundant.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalOrderingExpressionTreeVisitor.cs
```diff
@@ -27,44 +32,143 @@ private new RelationalQueryModelVisitor QueryModelVisitor
             get { return (RelationalQueryModelVisitor)base.QueryModelVisitor; }
         }
 
+        protected override Expression VisitConditionalExpression(ConditionalExpression conditionalExpression)
+        {
+            var selectExpression = ((RelationalQueryModelVisitor)base.QueryModelVisitor)
```
anpete: Use QueryModelVisitor above.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalOrderingExpressionTreeVisitor.cs
```diff
@@ -27,44 +32,143 @@ private new RelationalQueryModelVisitor QueryModelVisitor
...
+                .GetSelectExpressions().First();
```
anpete: This doesn't work in general. The target SelectExpression is looked up as part of member binding.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalOrderingExpressionTreeVisitor.cs
```diff
@@ -27,44 +32,143 @@ private new RelationalQueryModelVisitor QueryModelVisitor
...
+                {
+                    selectExpression.AddToOrderBy(updatedExpression, _ordering.OrderingDirection);
+                }
+
+                return updatedExpression;
+            }
+            return conditionalExpression;
+        }
+
+        protected override Expression VisitParameterExpression(ParameterExpression parameterExpression)
```
anpete: Redundant?

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -323,6 +325,11 @@ var queriesWithOrdering
             }
         }
 
+        public virtual IEnumerable<SelectExpression> GetSelectExpressions()
```
anpete: Remove this.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -429,6 +434,9 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
                     case ExpressionType.LessThanOrEqual:
                         op = " <= ";
                         break;
+                    case ExpressionType.Coalesce:
```
anpete: Don't try and do this here, just handle Coalesce separately.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -464,6 +472,33 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
...
+
+        protected override Expression VisitQuerySourceReferenceExpression(QuerySourceReferenceExpression expression)
+        {
+            _sql.Append(DelimitIdentifier(expression.ReferencedQuerySource.ItemName));
+            return expression;
+        }
+
+        protected override Expression VisitConditionalExpression(ConditionalExpression expression)
+        {
+            _sql.Append("CASE WHEN ");
```
anpete: We support CaseExpression in the tree directly - so rather than do this here, just embed the appropriate CaseExpressionNode in the OrderBy.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalOrderingExpressionTreeVisitor.cs
lukewaters: This one isn't, if I don't override the method and allow it to hit the base, we turn the parameterExpression into a MemberAccessExpression which we don't correctly sql gen

anpete: Weird. Do you know why?

test/EntityFramework.SqlServer.FunctionalTests/QueryBugsTest.cs
anpete: Lets create new issues for those.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -464,6 +472,33 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
             return binaryExpression;
         }
 
+        protected override Expression VisitMemberExpression(MemberExpression memberExpression)
```
anpete: How does a MemberExpression get into the SQL tree? It should be a ColumnExpression as a result of binding.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -464,6 +472,33 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
...
+        {
+            VisitExpression(memberExpression.Expression);
+            _sql.Append(".")
+                .Append(DelimitIdentifier(memberExpression.Member.Name));
+
+            return memberExpression;
+        }
+
+        protected override Expression VisitQuerySourceReferenceExpression(QuerySourceReferenceExpression expression)
```
anpete: As above re binding.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalOrderingExpressionTreeVisitor.cs
```diff
@@ -12,6 +15,7 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
     public class RelationalOrderingExpressionTreeVisitor : DefaultQueryExpressionTreeVisitor
     {
         private readonly Ordering _ordering;
+        private bool _adding;
 
         public RelationalOrderingExpressionTreeVisitor(
```
anpete: I don't think the solution below is quite right. For complex translation you should follow the pattern in the filtering visitor. i.e. Build up the SQL expression tree by returning it from the visitor (rather than trying to update the SelectExpression on the fly). I think an easier approach to this might be to add this translation to filtering first. Then we can go back and triangulate into ordering.

anpete: Eventually we should land in a place where there is little difference between translating expressions between filtering, ordering or projection - we should be using the same code.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalOrderingExpressionTreeVisitor.cs
lukewaters: ok

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
lukewaters: Ok, I'll pull all of the coalesce logic out to before the `else` block

src/EntityFramework.Relational/Utilities/ExpressionExtensions.cs
lukewaters: ok

lukewaters:
I still have to add tests for filtering, projection, and include

lukewaters:
@maumar and @anpete this is ready for a look

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
```diff
@@ -264,6 +248,62 @@ yield return
...
+        {
+            var columnExpression = expression as ColumnExpression;
+
+            if (columnExpression != null)
+            {
+                return new ColumnExpression(columnExpression.Alias ?? columnExpression.Name, columnExpression.Property, tableExpression);
+            }
+            switch (expression.NodeType)
+            {
+                case ExpressionType.Coalesce:
```
maumar: i would make it into if -else, rather than switch. Alternatively keep it as is, but add { } for the relevant code blocks

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -282,6 +282,38 @@ var projectionIndex
...
+                var columnExpression = expression as ColumnExpression;
+                if (columnExpression != null)
+                {
+                    AddToProjection(columnExpression);
+                }
+                else
+                {
+                    switch (expression.NodeType)
+                    {
+                        case ExpressionType.Coalesce:
```
maumar: same as above: i would make it into if -else, rather than switch. Alternatively keep it as is, but add { } for the relevant code blocks

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -282,6 +282,38 @@ var projectionIndex
             return projectionIndex;
         }
 
+        public virtual void AddToProjection([NotNull] IEnumerable<Expression> expressions)
```
maumar: I would move all that logic up the stream - process the expressions into column expressions, and use existing AddToProjection(ColumnExpresssion) instead

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -282,6 +282,38 @@ var projectionIndex
...
+                            break;
+                        case ExpressionType.Conditional:
+                            var conditionalExpression = expression as ConditionalExpression;
+                            AddToProjection(new[]
+                                {
+                                    conditionalExpression.Test,
+                                    conditionalExpression.IfFalse,
+                                    conditionalExpression.IfTrue
+                                });
+                            break;
```
maumar: add safeguard for case when we hit unexpected node here (not Column, Coalesce or Conditional) - we should throw in that case

lukewaters: Our previous behavior was to ignore any expression we didn't know how to handle rather than throwing.  I tried to preserve this behavior.  Do we want to change our default behavior to throwing for expressions we don't know how to parse?

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
```diff
@@ -264,6 +248,62 @@ yield return
...
+
+        private static Ordering UpdateColumnExpression(Ordering ordering, TableExpressionBase tableExpression)
+        {
+            var newExpression = UpdateColumnExpression(ordering.Expression, tableExpression);
+            var newOrdering = new Ordering(newExpression, ordering.OrderingDirection);
+
+            return newOrdering;
+        }
+
+        private static Expression UpdateColumnExpression(Expression expression, TableExpressionBase tableExpression)
```
maumar: this whole method looks awfully similar to SelectExpression.AddProjection you added below - see if you can DRY it up a bit - e.g. create a method that produces collection of ColumnExpressions (as in extracting columns from Coalesce/Conditional etc) - and then just use old logic - should be much simpler.

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
maumar: keep it as is then, my bad

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
```diff
@@ -218,17 +212,7 @@ in innerJoinSelectExpression.OrderBy
...
-                        var columnExpression = (ColumnExpression)ordering.Expression;
-
-                        targetSelectExpression
-                            .AddToOrderBy(
-                                columnExpression.Alias ?? columnExpression.Name,
-                                columnExpression.Property,
-                                innerJoinExpression,
-                                ordering.OrderingDirection);
-                    }
+                    UpdateOrderby(selectExpression, targetSelectExpression, innerJoinExpression);
```
maumar: UpdateOrderBy does not reflect what that name does - please rename or break it into two methods

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
lukewaters: ok

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
lukewaters: ok

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -1950,6 +1950,47 @@ public override void Where_chain()
...
+        }
+
+        public override void Ternary_orderby()
+        {
+            base.Ternary_orderby();
+
+            Assert.Equal(
+                @"SELECT [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[CustomerID], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM [Customers] AS [c]
+ORDER BY CASE WHEN (
```
anpete: Ternary should probably format into one line.

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -2598,6 +2598,67 @@ public virtual void Where_chain()
...
+
+        [Fact]
+        public virtual void Coalesce_nested_query_include()
+        {
+            using (var context = CreateContext())
+            {
+                var customers
+                    = (from c1 in context.Set<Customer>()
+
+                       from c2 in context.Set<Customer>()
```
anpete: Blank line.

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -2598,6 +2598,67 @@ public virtual void Where_chain()
                 .Where(o => o.OrderDate > new DateTime(1998, 1, 1)), entryCount: 8);
         }
 
+
+        [Fact]
+        public virtual void Coalesce_orderby()
```
anpete: Change "Coalesce" to "Null coalesce" everywhere.

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -2598,6 +2598,67 @@ public virtual void Where_chain()
...
+
+        [Fact]
+        public virtual void Coalesce_select()
+        {
+            AssertQuery<Customer>(customer => customer
+                .Select(c => new { c.CustomerID, c.CompanyName, Region = c.Region ?? "ZZ" }).OrderBy(o => o.Region));
+        }
+
+        [Fact]
+        public virtual void Ternary_orderby()
```
anpete: Change "Ternary" to "Conditional operator". Also change test name format to match others in this file. E.g. "OrderBy_conditional_operator".

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -1950,6 +1950,47 @@ public override void Where_chain()
...
+        }
+
+        public override void Coalesce_select()
+        {
+            base.Coalesce_select();
+
+            Assert.Equal(
+                @"SELECT [c].[CustomerID], [c].[CompanyName], [c].[Region]
+FROM [Customers] AS [c]
+ORDER BY COALESCE([c].[Region], 'ZZ')",
```
anpete: This SQL seems incorrect. In the query the coalesce operator is in the projection not the order by.

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
anpete: Switch is fine for NodeType.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
lukewaters: ok

lukewaters: updating with this signature `private static void UpdateOrderbyColumnBinding(IEnumerable<Ordering> orderBy, SelectExpression targetSelectExpression, JoinExpressionBase innerJoinExpression)`

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
lukewaters: Will address as part of #1798

lukewaters:
@anpete and @maumar I've got another revision ready for review

lukewaters:
@anpete I've got support for projection in the PR now too

lukewaters:
Note: this should fix #1700 and #1798

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
```diff
@@ -264,6 +254,11 @@ yield return
             }
         }
 
+        private static void UpdateOrderByColumnBinding(IEnumerable<Ordering> orderBy, SelectExpression targetSelectExpression, JoinExpressionBase innerJoinExpression)
```
anpete: Inline this...

src/EntityFramework.Relational/Query/Expressions/AliasExpression.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Relational.Query.Sql;
+using Microsoft.Data.Entity.Utilities;
+using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Parsing;
+
+namespace Microsoft.Data.Entity.Relational.Query.Expressions
+{
+    public class AliasExpression : ExtensionExpression
```
anpete: We should change ColumnExpression to use this too.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/SqlTranslatingExpressionTreeVisitor.cs
```diff
@@ -30,6 +30,13 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
 
             switch (binaryExpression.NodeType)
             {
+                case ExpressionType.Coalesce:
+                {
+                    var left = VisitExpression(binaryExpression.Left);
+                    var right = VisitExpression(binaryExpression.Right);
+
+                    return new AliasExpression(binaryExpression.Update(left, binaryExpression.Conversion, right));
```
anpete: Update? Why did I not know about this!? :-)

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/SqlTranslatingExpressionTreeVisitor.cs
```diff
@@ -78,6 +85,15 @@ var structuralComparisonExpression
             return null;
         }
 
+        protected override Expression VisitConditionalExpression(ConditionalExpression expression)
+        {
+            var test = VisitExpression(expression.Test);
+            var ifTrue = VisitExpression(expression.IfTrue);
+            var ifFalse = VisitExpression(expression.IfFalse);
+
+            return new CaseExpression(expression.Update(test, ifTrue, ifFalse));
```
anpete: I think CaseExpression is going away... @maumar ?

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -225,6 +225,15 @@ var subquery
             return subquery;
         }
 
+        public virtual void AssignAliasExpressions()
+        {
+            int aliasCount = 0;
+            foreach (var aliasExpression in _projection.OfType<AliasExpression>())
```
anpete: Aliases should be assigned at the point the expression is added to the projection. See ColumnExpression.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -589,6 +627,27 @@ public virtual Expression VisitColumnExpression(ColumnExpression columnExpressio
...
+                {
+                    _sql.Append(" AS ");
+                }
+            }
+            if (aliasExpression.Alias != null)
+            {
+                _sql.Append(DelimitIdentifier(aliasExpression.Alias));
+            }
+
+            aliasExpression.Projected = true;
```
anpete: We shouldn't mutate the expression as part of SQL gen.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
lukewaters: ok

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/SqlTranslatingExpressionTreeVisitor.cs
maumar: I want to finish null semantics first. I would say keep it as is for now, I will fix things globally when I get to it.

lukewaters: I Chatted with Maurycy and this isn't going to change for a couple weeks.  leaving as is for now

lukewaters:
The current state of this change enables Expressions in Project and Orderby but does not use Alias expressions for column expressions.  I am working on that change, but I wanted to get this change out for review in a working state.

lukewaters:
More changes have taken place, We now wrap all column expressions in alias expressions

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/EqualityPredicateExpandingVisitor.cs
```diff
@@ -18,11 +18,11 @@ public class EqualityPredicateExpandingVisitor : ExpressionTreeVisitor
                 && expression.Left.Type == typeof(bool)
                 && expression.Right.Type == typeof(bool))
             {
-                var complexLeft = !(expression.Left is ColumnExpression
+                var complexLeft = !((expression.Left as AliasExpression)?.ColumnExpression != null
```
maumar: perhaps make extension method e.g. IsColumnExpression or/and AsColumnExpression - the way it is now, it looks fairly gnarly.

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -255,11 +270,75 @@ public virtual int AddToProjection([NotNull] Expression expression)
...
+            {
+                return AddToProjection(aliasExpression);
+            }
+
             _projection.Add(expression);
 
             return _projection.Count - 1;
         }
 
+        public virtual int AddToProjection([NotNull] AliasExpression aliasExpression)
```
lukewaters: I will replace this method with one that only takes {string alias, Expression expression} so we can call without having to create an AliasExpression first

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -244,8 +259,8 @@ var subquery
             {
                 projectionIndex = _projection.Count;
 
-                _projection.Add(
-                    new ColumnExpression(column, property, FindTableForQuerySource(querySource)));
+                AddToProjection(new AliasExpression(
```
lukewaters: instead of creating an AliasExpression here, I will just call the new method I will add AddAliasToProjection(string alias, Expression expression);

src/EntityFramework.Relational/Query/Expressions/AliasExpression.cs
lukewaters: :+1:

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
```diff
@@ -199,17 +200,17 @@ var materializer
 
                     var innerJoinSelectExpression
                         = selectExpression.Clone(
-                            ((ColumnExpression)selectExpression.OrderBy.Last().Expression).TableAlias);
+                            ((AliasExpression)selectExpression.OrderBy.Last(o => o.Expression.AliasHasColumnExpression())
```
mikary: This could use some cleanup

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
```diff
@@ -343,11 +334,11 @@ var equalExpression
 
             var matchingColumnExpression
                 = projections
-                    .OfType<ColumnExpression>()
-                    .Last(p => p.Property == property);
+                    .OfType<AliasExpression>()
+                    .Last(p => p.ColumnExpression()?.Property == property);
```
mikary: Take another look at this, the semantics have changed

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/SqlTranslatingExpressionTreeVisitor.cs
mikary: Follow up on this

src/EntityFramework.Relational/Query/Expressions/AliasExpression.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            Alias = alias;
+            Expression = expression;
+        }
+
+        public virtual string Alias{ get; [param: NotNull] set; }
+
+        public virtual Expression Expression { get; [param: NotNull] set; }
+
+        public virtual bool Projected { get; set; } = false;
```
mikary: Can we remove?

src/EntityFramework.Relational/Query/Expressions/InExpression.cs
```diff
@@ -46,7 +46,7 @@ protected override Expression VisitChildren(ExpressionTreeVisitor visitor)
 
         public override string ToString()
         {
-            return Column + " IN (" + string.Join(", ", Values) + ")";
+            return Alias.ColumnExpression() + " IN (" + string.Join(", ", Values) + ")";
```
mikary: investigate more, what if it isn't a column expression?

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -147,15 +147,28 @@ public virtual bool IsDistinct
...
-                            new Ordering(
-                                new ColumnExpression(
-                                    columnExpression.Name,
-                                    columnExpression.Property,
-                                    subquery),
-                                ordering.OrderingDirection));
+                        var columnExpression = ordering.Expression as ColumnExpression;
+                        var aliasExpression = ordering.Expression as AliasExpression;
+
+                        if (columnExpression != null)
```
mikary: This code path shouldn't be hit

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -191,14 +204,17 @@ var subquery
 
             var columnAliasCounter = 0;
 
-            foreach (var columnExpression in _projection.OfType<ColumnExpression>())
+            foreach (var expression in _projection)
```
mikary: This could probably be simplified a bit

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -109,7 +109,8 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
                     VisitExpression(selectExpression.Predicate);
 
                     if (selectExpression.Predicate is ColumnExpression
-                        || selectExpression.Predicate is ParameterExpression)
+                        || selectExpression.Predicate is ParameterExpression
+                        || (selectExpression.Predicate as AliasExpression)?.Expression is ColumnExpression)
```
mikary: ```.AliasHasColumnExpression()```

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -475,103 +503,115 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
...
-            if (binaryExpression.IsLogicalOperation()
-                && (binaryExpression.Left is ColumnExpression
-                    || binaryExpression.Left is ParameterExpression))
-            {
-                _sql.Append(" = 1");
-            }
+                if (binaryExpression.IsLogicalOperation()
+                    && ( binaryExpression.Left is ColumnExpression
+                        || binaryExpression.Left is ParameterExpression
+                        || (binaryExpression.Left as AliasExpression)?.Expression is ColumnExpression))
```
mikary: Use extension method

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -475,103 +503,115 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
...
-            if (binaryExpression.IsLogicalOperation()
-                && (binaryExpression.Right is ColumnExpression
-                    || binaryExpression.Right is ParameterExpression))
-            {
-                _sql.Append(" = 1");
-            }
+                if (binaryExpression.IsLogicalOperation()
+                    && (binaryExpression.Right is ColumnExpression
+                        || binaryExpression.Right is ParameterExpression
+                        || (binaryExpression.Right as AliasExpression)?.Expression is ColumnExpression))
```
mikary: here too

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -684,7 +737,8 @@ protected override Expression VisitUnaryExpression(UnaryExpression unaryExpressi
 
                 var isColumnOrParameterOperand =
                     unaryExpression.Operand is ColumnExpression
-                    || unaryExpression.Operand is ParameterExpression;
+                    || unaryExpression.Operand is ParameterExpression
+                    || (unaryExpression.Operand as AliasExpression)?.Expression is ColumnExpression;
```
mikary: And here (probably just search for as AliasExpression)

mikary:
Looks pretty good otherwise

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
lukewaters: yeah, this is because I wrapped the column expression in an alias expression, so the situation we're trying to match has changed

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
lukewaters: how does this look 
```
var innerJoinSelectExpression
    = selectExpression.Clone(
    selectExpression.OrderBy.Select(o => o.Expression).Last(o => o.AliasHasColumnExpression())
    .GetColumnExpression().TableAlias);
```

src/EntityFramework.Relational/Query/Expressions/InExpression.cs
lukewaters: Changing to `Alias.Expression`

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
lukewaters: Verified, removing

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
lukewaters: updated to 
```
            var columnAliasCounter = 0;

            foreach (var aliasExpression in _projection.OfType<AliasExpression>())
            {
                var columnExpression = aliasExpression.GetColumnExpression();
                if (columnExpression != null && subquery._projection.OfType<AliasExpression>().Any(ae => 
                        ae.ColumnExpression()?.Name == columnExpression.Name))
                {
                    aliasExpression.Alias = "c" + columnAliasCounter++;
                }

                subquery._projection.Add(aliasExpression);
            }
```

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
lukewaters: :+1:

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
lukewaters: :+1:

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
lukewaters: :+1:

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
lukewaters: :+1:

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/SqlTranslatingExpressionTreeVisitor.cs
lukewaters: ^^ this change hasn't happened yet

src/EntityFramework.Relational/Query/Expressions/AliasExpression.cs
lukewaters: It looks like we need to keep track of this state independently of the other properties on this expression.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
maumar: just to be safe - add a null check for propery

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/SqlTranslatingExpressionTreeVisitor.cs
```diff
@@ -230,11 +248,13 @@ protected override Expression VisitMemberExpression([NotNull] MemberExpression m
...
                     memberExpression,
                     (property, querySource, selectExpression)
-                        => new ColumnExpression(
-                            _queryModelVisitor.QueryCompilationContext
-                                .GetColumnName(property),
-                            property,
-                            selectExpression.FindTableForQuerySource(querySource)));
+                        =>
+                        {
+                            return new AliasExpression(new ColumnExpression(
```
maumar: nit pick - make the formating of the call consistent - i would break the line after "return new AliasExpression("

src/EntityFramework.Relational/Query/Expressions/AliasExpression.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            return specificVisitor != null
+                ? specificVisitor.VisitAliasExpression(this)
+                : base.Accept(visitor);
+        }
+
+
+        protected override Expression VisitChildren(ExpressionTreeVisitor visitor)
+        {
+            return this;
```
maumar: should visit underlying expression here

src/EntityFramework.Relational/Query/Expressions/AliasExpressionExtensions.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Relational.Query.Expressions
+{
+    public static class AliasExpressionExtensions
```
maumar: put this into expressionextensions instead, no need for a separate class for those couple of methods

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -244,8 +248,7 @@ var subquery
             {
                 projectionIndex = _projection.Count;
 
-                _projection.Add(
-                    new ColumnExpression(column, property, FindTableForQuerySource(querySource)));
+                AddAliasToProjection(alias: null, expression: new ColumnExpression(column, property, FindTableForQuerySource(querySource)));
```
maumar: format - break lines between arguments to the AddAliasToProjection

src/EntityFramework.Relational/Query/Expressions/AliasExpressionExtensions.cs
lukewaters: :+1:

src/EntityFramework.Relational/Query/Expressions/AliasExpression.cs
lukewaters: Ok, will visit inner expression

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
lukewaters: :+1:

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IncludeExpressionTreeVisitor.cs
lukewaters: Adding `Check.NotNull(property, nameof(property())`

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -109,7 +109,8 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
                     VisitExpression(selectExpression.Predicate);
 
                     if (selectExpression.Predicate is ColumnExpression
-                        || selectExpression.Predicate is ParameterExpression)
+                        || selectExpression.Predicate is ParameterExpression
+                        || selectExpression.Predicate.AliasHasColumnExpression())
```
maumar: change here also

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -124,12 +125,24 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
                 VisitJoin(selectExpression.OrderBy, t =>
                     {
                         var columnExpression = t.Expression as ColumnExpression;
+                        var aliasExpression = t.Expression as AliasExpression;
                         if (columnExpression != null)
```
lukewaters: will remove columnexpression logic as it is never hit

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
lukewaters: yes, confirmed we never have a ColumnExpression here

maumar:
:shipit:

PR #1750, opened 2015-03-04
Allow configuring a relationship to an entity type without a primary key

### LINKED ISSUE #1603: Configuration: defining a relationship to an entity type without primary key shouldn't throw
It should be possible to build a model like this:
```
modelBuilder.Entity<CustomerDetails>()
    .HasOne(e => e.Customer)
    .WithOne(e => e.Details)
    .ReferencedKey<Customer>(e => e.Name);
modelBuilder.Entity<Customer>()
    .Key(e => e.Name);
```

AndriySvyryd:
Remove shadow keys added by convention when the primary key is set
Update relationships when a new primary key is set

Fixes #1603

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
@@ -52,27 +52,72 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                 return null;
             }
 
-            var keyToReplace = Metadata.TryGetPrimaryKey();
-            var existingKey = Metadata.TryGetKey(properties);
-
-            if (keyToReplace != null
-                && keyToReplace != existingKey
-                && !RemoveKey(keyToReplace, configurationSource).HasValue)
+            var oldPrimaryKey = Metadata.TryGetPrimaryKey();
+            var newPrimaryKey = Metadata.TryGetKey(properties);
+            if (oldPrimaryKey != null
+                && oldPrimaryKey != newPrimaryKey)
             {
-                return null;
+                if (!configurationSource.Overrides(_keyBuilders.GetConfigurationSource(oldPrimaryKey)))
+                {
+                    return null;
+                }
+
+                if (newPrimaryKey == null)
+                {
+                    newPrimaryKey = Key(properties, ConfigurationSource.Convention).Metadata;
+                }
+
+                UpdateReferencingForeignKeys(oldPrimaryKey, newPrimaryKey, configurationSource);
             }
 
-            if (existingKey != null)
+            if (newPrimaryKey != null)
             {
                 Metadata.SetPrimaryKey(properties);
             }
 
-            return _keyBuilders.GetOrAdd(
-                () => existingKey,
+            var keyBuilder = _keyBuilders.GetOrAdd(
+                () => newPrimaryKey,
                 () => Metadata.SetPrimaryKey(properties),
                 key => new InternalKeyBuilder(key, ModelBuilder),
                 onNewKeyAdded: ModelBuilder.ConventionDispatcher.OnKeyAdded,
                 configurationSource: configurationSource);
+
+            ReplaceConventionShadowKeys(keyBuilder.Metadata);
+
+            return keyBuilder;
+        }
+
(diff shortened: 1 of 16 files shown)


src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
```diff
@@ -52,27 +52,72 @@ private InternalKeyBuilder PrimaryKey(IReadOnlyList<Property> properties, Config
...
+            ReplaceConventionShadowKeys(keyBuilder.Metadata);
+
+            return keyBuilder;
+        }
+
+        private void UpdateReferencingForeignKeys(Key keyToReplace, Key newKey, ConfigurationSource configurationSource)
+        {
+            var newProperties = newKey.Properties;
+
+            var allForeignKeysReplaced = true;
```
anpete: This looks like LINQ Aggregate.

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
```diff
@@ -997,6 +1016,17 @@ private InternalIndexBuilder Index(IReadOnlyList<Property> properties, Configura
             {
                 var baseName = (string.IsNullOrEmpty(navigationToPrincipal) ? principalType.SimpleName : navigationToPrincipal);
 
+                if (principalKey == null)
+                {
+                    var principalKeyProperty = CreateUniqueProperty(
+                        "Id",
```
anpete: Should this be a constant?

src/EntityFramework.Core/Metadata/Property.cs
```diff
@@ -210,6 +210,46 @@ private void SetFlag(bool? value, PropertyFlags flag)
 
         internal static string Format(IEnumerable<IProperty> properties) => "{" + string.Join(", ", properties.Select(p => "'" + p.Name + "'")) + "}";
 
+        public static bool AreCompatible([NotNull] IReadOnlyList<Property> principalProperties,
+            [NotNull] IReadOnlyList<Property> dependentProperties)
+        {
+            return IsPropertyCountSame(principalProperties, dependentProperties)
```
anpete: Seems over factored.

anpete:
:shipit:

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
AndriySvyryd: It doesn't have to be any specific value, just a placeholder before the real PK is configured

src/EntityFramework.Core/Metadata/Property.cs
AndriySvyryd: Subjective

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
AndriySvyryd: Looks better as is. The main purpose of the call is in the side effects, not the return value.

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
anpete: "Placeholder"?

src/EntityFramework.Core/Metadata/Property.cs
anpete: I predicted that answer. Curious, do you do this for all x.Count == y.Count expressions? What is the criteria?

anpete: Also "Equal" is more correct than "Same" here.

src/EntityFramework.Core/Metadata/Internal/InternalEntityBuilder.cs
AndriySvyryd: TempId?

src/EntityFramework.Core/Metadata/Property.cs
AndriySvyryd: It looks more consistent with ArePropertyTypesCompatible here.
I'll change the name to ArePropertyCountsEqual

divega:
Blocked on #1741.

Issue #1076, opened 2014-11-17
Better navigation property names when reverse engineering a model


rowanmiller:
We should just get this right from the start and not generate ```Order.Customer```, ```Order.Customer1``` like we did in past versions.

Some comments from a blog post:

> When you start making reverse-engineer tools, please make reference names to classes meaningful. 
Example: When you have table Customer that has field KeyAccountManagerId that reference to table Employee, then reference should not be in code Customer.Employee but Customer.KeyAccountManager.
For LinqToSQL and EF6 i have written my on tools to make this possible, because Reference names like Emloyee1, and Employee2 will lead some serious business logic bugs.

lajones:
Part of #830.

lajones:
@rowanmiller #1751 implements this. However although this makes some situations better (e.g. ReportsTo on Employees in Northwind now comes out as a property called ReportsTo instead of Employees1), it also makes some situations worse. For instance in AdventureWorks there are 7 tables whose PK is BusinessEntityID (only 1 of which is actually called BusinessEntity). The foreign keys are also all called BusinessEntityID. Hence all dependent end Nav Props to any of those entity types are all called BusinessEntity (though they have the correct type) where before they would have been called the same as the type they pointed to. Looking to the future inheritance scenarios might be mixed up by this also.

rowanmiller:
I think that's probably fine, in general the FK names are probably the most likely to be correct (we use FK name right and not PK name).

lajones:
@rowanmiller Correct. FK name, not PK name.

lajones:
Done - with commit d2f7162.

PR #1751, opened 2015-03-04
Make ModelUtilities a service and better NavProp names

lajones:
Addresses #1076 and #1672.

Make ModelUtilities a service and change the algorithm for producing candidate NavProp names.

### DIFF
--- a/src/EntityFramework.Relational.Design/Utilities/ModelUtilities.cs
+++ b/src/EntityFramework.Relational.Design/Utilities/ModelUtilities.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
@@ -8,20 +9,10 @@ using JetBrains.Annotations;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Utilities;
 
-namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
+namespace Microsoft.Data.Entity.Relational.Design.Utilities
 {
     public class ModelUtilities
     {
-        private static readonly ModelUtilities _instance = new ModelUtilities();
-
-        public static ModelUtilities Instance
-        {
-            get
-            {
-                return _instance;
-            }
-        }
-
         public virtual string GenerateLambdaToKey(
             [NotNull] IEnumerable<IProperty> properties,
             [NotNull] string lambdaIdentifier)
@@ -49,7 +40,9 @@ namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
         {
             Check.NotNull(entityType, nameof(entityType));
 
-            var primaryKeyProperties = entityType.GetPrimaryKey().Properties.ToList();
+            var primaryKeyProperties =
+                entityType.TryGetPrimaryKey()?.Properties.ToList()
+                ?? new List<IProperty>();
             foreach (var property in primaryKeyProperties)
             {
                 yield return property;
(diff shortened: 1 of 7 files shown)


src/EntityFramework.Relational.Design/Utilities/ModelUtilities.cs
```diff
@@ -66,5 +59,34 @@ public virtual IEnumerable<IProperty> OrderedProperties([NotNull] IEntityType en
...
         }
+
+        public virtual string GetDependentEndCandidateNavigationPropertyName([NotNull] ForeignKey foreignKey)
+        {
+            Check.NotNull(foreignKey, nameof(foreignKey));
+
+            if (foreignKey.Properties.Count == 1)
+            {
+                var foreignKeyPropertyName = foreignKey.Properties.Single().Name;
+                if (foreignKeyPropertyName.EndsWith("id", StringComparison.OrdinalIgnoreCase))
```
lajones: Have been thinking about this and will update to strip off the "Id" only if it is specifically "Id" or "ID". "id" might be the end of some other word and "iD" just seems weird.

bricelam: Yes, this feels brittle. Instead of `EndsWith("Id")` can we do something like `EndsWith(principalPKColumnName)`? This might also allow us to handle composite keys better too. E.g. EmployerId1, EmployerId2 could become Employer instead of just Company.

It also doesn't seem to work well with databases that name things using underscores like EMPLOYER_ID.

I wonder if it's worth even trying to do this in v1. In EF6, we just uniqified the pluralized principal table name.

src/EntityFramework.Relational.Design/Utilities/ModelUtilities.cs
```diff
@@ -66,5 +59,34 @@ public virtual IEnumerable<IProperty> OrderedProperties([NotNull] IEntityType en
...
+            }
+
+            return foreignKey.ReferencedEntityType.SimpleName;
+        }
+
+        public virtual string GetPrincipalEndCandidateNavigationPropertyName([NotNull] ForeignKey foreignKey)
+        {
+            Check.NotNull(foreignKey, nameof(foreignKey));
+
+            return foreignKey.EntityType.SimpleName;
```
bricelam: Are we going to pluralize these?

bricelam:
Our usage of DI here (including the Migrations commands) doesn't gel well with the rest of the stack. We should look at ways of reusing more of the common components and patterns.

src/EntityFramework.Relational.Design/Utilities/ModelUtilities.cs
lajones: There's a whole lot more we could on pluralization/singularization. But discussed that with @rowanmiller a while back and decided not to get into that - at least for this release.

As for priority - #1076 is marked as pri0. Again I'd defer to @rowanmiller on that. And unfortunately just uniquifying the principal table name is what causes the Employee, Employee1 situation which is the main complaint of that issue.

Not sure I follow on the composite key scenario. If the foreign composite key is {EmployerId1, EmployerId2} and the principalKey is (CompanyId1, CompanyId2} how does EndsWith(matchingPrincipalPKColumnName) help me come up with "Employer" as the NavProp name? The problem with composite keys is you have to pick one name from all the column names - and it's hard to guess what would be most helpful - I could try to find a common prefix that all the column names share and then strip off "I[dD]" and uniquify that? An alternative is to use the name of the constraint in the database - but that's usually something unhelpful like FK_Customer_Employee_1 i.e. all it tells you is the names of the tables at either end. And it doesn't have to follow that pattern - the DBA is free to name it whatever they want.

src/EntityFramework.Relational.Design/Utilities/ModelUtilities.cs
lajones: Not for now at least - see comment above.

src/EntityFramework.Relational.Design/Utilities/ModelUtilities.cs
lajones: Also am happy to extend it to look for EndsWIth("_id", StringComparison.OrdinalIgnoreCase) as well.

lajones:
Re DI comment - agreed. I'll raise an issue to look at it as a whole.

bricelam:
:shipit:

src/EntityFramework.Relational.Design/Utilities/ModelUtilities.cs
bricelam: I talked with Rowan, and I'm fine with however much of this we want to implement. We should, however, review what we do by convention in a (possibly ad-hoc) design meeting.

lajones: @bricelam Sure. No problem - happy to review.

lajones:
Checked in - commit d2f7162.

PR #1752, opened 2015-03-04
Fix to #1726 - using boolProperty.Equals(true) in filter produces invalid sql (take 3)

maumar:

Fix is to revert the previous change and instead compensate in the sql generation phase - adding " = 1" terms for 'naked' column expressions as well as column expressions inside logical expression, inside WHERE predicate.

things like: products.Where(p => p.Discontinued == (p.ProductID > 100)) are still broken but the main scenarios works fine.

### DIFF
--- a/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
+++ b/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Data.Entity.Relational.Query.Sql
         private Expression _binaryExpression;
         private List<string> _parameters;
         private IDictionary<string, object> _parameterValues;
+        private Stack<bool> _insideFilter = new Stack<bool>();
 
         public virtual string GenerateSql(
             SelectExpression selectExpression, IDictionary<string, object> parameterValues)
@@ -50,6 +51,8 @@ namespace Microsoft.Data.Entity.Relational.Query.Sql
             Check.NotNull(selectExpression, nameof(selectExpression));
 
             IDisposable subQueryIndent = null;
+
+            _insideFilter.Push(false);
 
             if (selectExpression.Alias != null)
             {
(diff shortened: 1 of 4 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -98,7 +101,16 @@ public virtual Expression VisitSelectExpression(SelectExpression selectExpressio
                 _sql.AppendLine()
                     .Append("WHERE ");
 
+                _insideFilter.Push(true);
+
                 VisitExpression(selectExpression.Predicate);
+
+                if (selectExpression.Predicate is ColumnExpression)
```
anpete: This work for !Discontinued?

anpete:
EF6 translation of the still broken case: 
```
SELECT 
      [Extent1].[ProductID] AS [ProductID], 
      [Extent1].[ProductName] AS [ProductName], 
      [Extent1].[SupplierID] AS [SupplierID], 
      [Extent1].[CategoryID] AS [CategoryID], 
      [Extent1].[QuantityPerUnit] AS [QuantityPerUnit], 
      [Extent1].[UnitPrice] AS [UnitPrice], 
      [Extent1].[UnitsInStock] AS [UnitsInStock], 
      [Extent1].[UnitsOnOrder] AS [UnitsOnOrder], 
      [Extent1].[ReorderLevel] AS [ReorderLevel], 
      [Extent1].[Discontinued] AS [Discontinued]
      FROM [dbo].[Products] AS [Extent1]
      WHERE [Extent1].[Discontinued] = (CASE WHEN ([Extent1].[ProductID] > 100) THEN cast(1 as bit) WHEN ( NOT ([Extent1].[ProductID] > 100)) THEN cast(0 as bit) END)
```

maumar:
sent new pull request

PR #1756, opened 2015-03-05
Logging API changes

BrennanConroy:
@bricelam 
Need to wait for logging to be pushed

### DIFF
--- a/test/EntityFramework.Core.Tests/TestUtilities/ListLogger.cs
+++ b/test/EntityFramework.Core.Tests/TestUtilities/ListLogger.cs
@@ -12,12 +12,12 @@ namespace Microsoft.Data.Entity.Tests.TestUtilities
     {
         public ListLogger(List<Tuple<LogLevel, string>> log)
         {
-            Log = log;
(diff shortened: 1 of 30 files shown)


test/EntityFramework.Core.Tests/TestUtilities/ListLogger.cs
```diff
@@ -12,12 +12,12 @@ public class ListLogger : ILogger
     {
         public ListLogger(List<Tuple<LogLevel, string>> log)
         {
-            Log = log;
+            Logs = log;
         }
 
-        public List<Tuple<LogLevel, string>> Log { get; }
+        public List<Tuple<LogLevel, string>> Logs { get; }
```
lukewaters: I don't like this, maybe LogMessages ?

divega: +1, and I would change the constructor parameter. Otherwise just leave it as it was.

divega:
:shipit:

PR #1757, opened 2015-03-05
Preserve relationships if foreign key or principal key properties are ignored

AndriySvyryd:


ajcvickers:
:shipit:

divega:
Blocked on #1741.

PR #1759, opened 2015-03-05
Fix for #1726 - using boolProperty.Equals(true) in filter produces invalid sql

maumar:
Fix is to revert the previous change and instead compensate in the sql generation phase - adding " = 1" terms for 'naked' column expressions as well as column expressions inside logical expression, inside WHERE predicate.

things like: products.Where(p => p.Discontinued == (p.ProductID > 100)) are still broken but the main scenarios works fine.


maumar:
@anpete

anpete:
:shipit:

maumar:
Merged - 2c4b376aef3d1faf5fa1963fe5253a2c18e31f4b

Issue #1747, opened 2015-03-04
Multi-select with Contains() does not generate SQL


rmja:
The following does not generate SQL with the `IN` operator.

    var ids = new int[] {1,2,3};
    _db.Table.Where(x => ids.Contains(x.Id)).ToList();

maumar:
duplicate of https://github.com/aspnet/EntityFramework/issues/1243

maumar:
Fixed in 32cb631986fbd10a1ec5e4aaa7dcaded0eab7270

rmja:
@maumar, one small detail. What is the intended behaviour when the list of ids is empty, say:

    var ids = new int[];
    _db.Table.Where(x => ids.Contains(x.Id)).ToList();

Right now it generates the following invalid SQL:

    SELECT [x].[Id], [x].[SomeOtherColumn]
    FROM [Table] AS [x]
    WHERE [x].[Id] IN ()
    go

I know that I should probably check for the list to be empty before sending it to the database, but it would be nice, if this was handled probably inside EF.

maumar:
EF should cover this case, will fix. Thanks for reporting!

maumar:
Fixed in 5fe084e06ed743d4d8423815829b802bcb9c86f1

PR #1781, opened 2015-03-06
Fix to #1747 - Multi-select with Contains() does not generate SQL

maumar:
Fix is to translate Where(c => [list_of_constants].Constains(c.SomeProperty)) to
c.SomeProperty IN ( [list_of_constants)

We also translate !Contains... into NOT IN. Relinq converts Contains call into subquery, so we are detecting this pattern in the FilteringExpressionTreeVisitor and translate it accordingly. Argument to contains in this case will always be in the form of parameter. Parameter values are wired in during SQL generation.

### DIFF
--- a/src/EntityFramework.Relational/Query/EqualityPredicateOptimizer.cs
+++ b/src/EntityFramework.Relational/Query/EqualityPredicateOptimizer.cs
@@ -91,6 +91,7 @@ namespace Microsoft.Data.Entity.Relational.Query
                     = VisitExpression(binaryExpression.Right) as TInExpression;
 
                 if (rightInExpression != null
+                    && rightInExpression.Values != null
                     && leftColumnExpression.Equals(rightInExpression.Column))
                 {
                     return inExpressionFactory(
@@ -106,7 +107,8 @@ namespace Microsoft.Data.Entity.Relational.Query
                 var leftInExpression
                     = VisitExpression(binaryExpression.Left) as TInExpression;
 
-                if (leftInExpression != null
+                if (leftInExpression != null 
+                    && leftInExpression.Values != null
                     && rightColumnExpression.Equals(leftInExpression.Column))
                 {
                     return inExpressionFactory(
--- a/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
@@ -4,9 +4,11 @@
 using System;
 using System.Linq;
 using System.Linq.Expressions;
-using JetBrains.Annotations;
 using Microsoft.Data.Entity.Relational.Query.Expressions;
 using Microsoft.Data.Entity.Utilities;
+using JetBrains.Annotations;
+using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Parsing;
 
 namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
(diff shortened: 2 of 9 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
```diff
@@ -242,6 +244,15 @@ protected override Expression VisitUnaryExpression([NotNull] UnaryExpression exp
 
                 if (operand != null)
                 {
+                    var inExpression = operand as InExpression;
+                    if (inExpression != null)
```
anpete: Use conditional operator.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
```diff
@@ -343,6 +354,33 @@ protected override Expression VisitParameterExpression([NotNull] ParameterExpres
...
         }
 
+        protected override Expression VisitSubQueryExpression(SubQueryExpression expression)
+        {
+            Check.NotNull(expression, nameof(expression));
+
+            if (expression.QueryModel.IsIdentityQuery()
+                && expression.QueryModel.ResultOperators.Count == 1)
+            {
+                var contains = expression.QueryModel.ResultOperators.Single() as ContainsResultOperator;
```
anpete: Use First when you have already checked the length.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/FilteringExpressionTreeVisitor.cs
```diff
@@ -343,6 +354,33 @@ protected override Expression VisitParameterExpression([NotNull] ParameterExpres
...
+                if (contains != null)
+                {
+                    var parameter = expression.QueryModel.MainFromClause.FromExpression as ParameterExpression;
+                    var memberItem = contains.Item as MemberExpression;
+                    if (parameter != null && memberItem != null)
+                    {
+                        var columnExpression = (ColumnExpression)VisitMemberExpression(memberItem);
+                        _requiresClientEval = false;
```
anpete: Don't set this to false - it might already be true.

src/EntityFramework.Relational/Query/EqualityPredicateOptimizer.cs
```diff
@@ -91,6 +91,7 @@ var rightInExpression
                     = VisitExpression(binaryExpression.Right) as TInExpression;
 
                 if (rightInExpression != null
+                    && rightInExpression.Values != null
```
anpete: I think we want to avoid making InExpressions have this dual behavior.  I would rather just have Values contain the parameter expression here.

src/EntityFramework.Relational/Query/Expressions/InExpression.cs
```diff
@@ -15,9 +15,28 @@ public class InExpression : InExpressionBase
...
-            : base(
+            : this(
                 Check.NotNull(column, nameof(column)),
-                Check.NotNull(values, nameof(values)))
+                Check.NotNull(values, nameof(values)),
+                null)
+        {
+        }
+
+        public InExpression(
```
anpete: See comment above.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -259,24 +260,51 @@ public virtual Expression VisitInExpression(InExpression inExpression)
...
+                    if (!first)
+                    {
+                        _sql.Append(", ");
+                    }
+                    else
+                    {
+                        first = false;
+                    }
+
+                    _sql.Append(GenerateLiteral((dynamic)value));
```
anpete: I think there is an injection vulnerability here.

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -1932,6 +1932,84 @@ public override void Where_chain()
...
+FROM [Customers] AS [c]
+WHERE ([c].[CustomerID] IN ('ALFKI', 'ABCDE') AND [c].[CustomerID] IN ('ABCDE', 'ALFKI'))",
+                Sql);
+        }
+
+        public override void Contains_with_local_collection_complex_predicate_or()
+        {
+            base.Contains_with_local_collection_complex_predicate_or();
+
+            Assert.Equal(
```
anpete: This should optimize.

PR #1784, opened 2015-03-09
Refactored relational expression translation to more easily support tran...

anpete:
...slation enhancements.


anpete:
@lukewaters This should help with the coalesce enhancement...

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -223,104 +217,73 @@ var innerJoinExpression
...
-            var projectionCounts
-                = _queriesBySource
-                    .Select(kv => new
-                        {
-                            SelectExpression = kv.Value,
-                            kv.Value.Projection.Count
-                        })
-                    .ToList();
+            var selectExpression = TryGetQuery(queryModel.MainFromClause);
+            var requiresClientEval = selectExpression == null;
```
lukewaters: Is there any way that we could get a selectExpression back that still requires partial client evaluation?

lukewaters:
:shipit:

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
anpete: This is just tracking whether this particular WhereClause requires client eval.

PR #1787, opened 2015-03-09
Query: Relational: More refactoring of expression translation.

anpete:


src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -71,7 +71,7 @@ protected set
 
         protected abstract ExpressionTreeVisitor CreateQueryingExpressionTreeVisitor([NotNull] IQuerySource querySource);
 
-        protected virtual ExpressionTreeVisitor CreateProjectionExpressionTreeVisitor()
+        protected virtual ExpressionTreeVisitor CreateProjectionExpressionTreeVisitor([NotNull] IQuerySource querySource)
```
divega: Sorry my lack of context on this, but is the argument querySource supposed to be ignored?

divega: I see it get's used in derived classes.

divega:
:shipit:

PR #1788, opened 2015-03-10
Rename IMetadata and MetadataBase to be general purpose Annotatable

mikary:
Changes for #1778 
@rowanmiller @bricelam


src/EntityFramework.Core/Infrastructure/AnnotationNotFoundException.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Infrastructure
+{
+    public class AnnotationNotFoundException : Exception
```
anpete: We need a custom exception type for this?

src/EntityFramework.Core/Infrastructure/AnnotationNotFoundException.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public AnnotationNotFoundException()
+        {
+        }
+
+        public AnnotationNotFoundException([NotNull] string message)
+            : base(message)
+        {
+            Check.NotEmpty(message, nameof(message));
```
anpete: Check inline when arg is passed to base.

src/EntityFramework.Core/Infrastructure/AnnotationNotFoundException.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        public AnnotationNotFoundException([NotNull] string message, [CanBeNull] Exception innerException)
+            : base(message, innerException)
+        {
+            Check.NotEmpty(message, nameof(message));
```
anpete: Check inline when arg is passed to base.

anpete:
:shipit:

src/EntityFramework.Core/Infrastructure/AnnotationNotFoundException.cs
mikary: Not necessarily, the ```ModelItemNotFoundException``` seemed inappropriate for a general purpose component, but an ```InvalidOperationException``` is probably just as good.

PR #1791, opened 2015-03-10
Cleanup DbContextOptions and factor our Builder

ajcvickers:
This change cleans up the DbContextOptions classes, the way they are built and handled, and the way extensions interact. There is now a DbContextOptionsBuilder class that is used in AddEntityFramework and OnConfiguring. This is used to create DbContextOptions objects which can be passed to the DbContext constructor and/or resolved from DI. From the app perspective everything works pretty much the same except that the DbContextOptions resolved from DI is no longer cloned before being passed to OnConfiguring. However, if OnConfiguring then does anything to change the options it will effectively result in the options being cloned at this time. The DbContextOptions themselves are effectively immutable.

All additional options are now added through extensions, and the way extensions add added/updated has been simplified a bit to have easier APIs and to decouple more cleanly from the raw options that are handled by DbContextOptions itself.

Auto configuring of a single data store now happens right after OnConfiguring is called, rather than later when a data store is selected. This is cleaner and means data store selection doesn't need to mutate options.

It is now possible to depend on non-generic DbContextOptions in your DbContext constructor, but a generic DbContextOptions for your context type must be used. Non-generic DbContextOptions is now abstract.

DbContextOptions parsing is now performed by a DI service.


test/EntityFramework.InMemory.Tests/InMemoryDbContextOptionsExtensionsTest.cs
```diff
@@ -1,45 +1,45 @@
...
+//// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+//// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+//using System.Linq;
+//using Microsoft.Data.Entity.Infrastructure;
+//using Xunit;
+
+//namespace Microsoft.Data.Entity.InMemory.Tests
+//{
+//    public class InMemoryDbContextOptionsExtensionsTest
```
mikary: Are these tests broken / going away?

ajcvickers: I missed re-enabling them. Will do so.

src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGenerator.cs
```diff
@@ -137,7 +137,7 @@ public virtual void GenerateOnConfiguringCode([NotNull] IndentedStringBuilder sb
 
             Generator.CSharpCodeGeneratorHelper.BeginMethod(AccessModifier.Protected,
                 VirtualModifier.Override, "void", "OnConfiguring", sb, _onConfiguringMethodParameters);
```
bricelam: Should the values in `_onConfiguringMethodParameters` also be updated?

ajcvickers: Will do. Interesting that no tests failed...

test/EntityFramework.InMemory.Tests/InMemoryDbContextOptionsExtensionsTest.cs
```diff
@@ -1,45 +1,45 @@
...
+//        {
+//            var optionsBuilder = new DbContextOptionsBuilder();
+//            optionsBuilder.UseInMemoryStore(persist: true);
+
+//            var extension = ((IDbContextOptions)options).Extensions.OfType<InMemoryOptionsExtension>().Single();
+
+//            Assert.True(extension.Persist);
+//        }
+//    }
+//}
```
bricelam: Should this file be updated (or removed)?

ajcvickers: Missed re-enabling these--will do so.

test/EntityFramework.SqlServer.FunctionalTests/ConnectionSpecificationTest.cs
```diff
@@ -494,18 +494,17 @@ public OptionsContext(DbContextOptions<OptionsContext> options, SqlConnection co
...
+                Assert.Same(_options, optionsBuilder.Options);
+
+                optionsBuilder.UseSqlServer(_connection);
 
-                // Options was cloned
-                Assert.NotSame(options, _options);
+                Assert.NotSame(_options, optionsBuilder.Options);
 
-                Assert.Equal(1, ((IDbContextOptions)options).Extensions.OfType<FakeDbContextOptionsExtension>().Count());
+                //Assert.Equal(1, optionsBuilder.Options.Extensions.OfType<FakeDbContextOptionsExtension>().Count());
```
bricelam: Intentionally left commented out?

ajcvickers: Should be removed. Will do so.

bricelam:
:shipit:

PR #1797, opened 2015-03-10
API review cleanup

ajcvickers:
Mostly making classes internal.
Also:
 - Changing ModelValidator base class to IModelValidator interface
 - Make AddDbContext not an extension method
 - Move ModelSourceHelpers helper method to ModelSource


anpete:
:shipit:

ajcvickers:
Merged

PR #1805, opened 2015-03-12
Query: Relational: Enhances projection translation. Adds partial support...

anpete:
... for complex projection expressions. More improvements to SQL gen.


maumar:
:shipit:

Issue #1374, opened 2015-01-09
Flag to allow application data in logging


rowanmiller:
From a security point of view we decided that, by default, both the message action and state passed to ILogger.Write should not contain any application data. That is, data that comes from the data store or may be supplied by an end user of the application. Depending on the application, this data may contain sensitive information (usernames, credit card numbers, etc.). Examples include, results from queries, values stored in entity instances, and constant values used in LINQ expressions.

We'll have a simple flag that can be enabled to allow this information to be included. This is especially useful for logging queries, where it is helpful to have the parameter values logged along with the query.

This work item includes:
- [x] Creating the flag (off by default)
- [ ] Reviewing existing logging to make sure it honors the flag
- [ ] Writing a warning to the log when the flag is enabled (once per context, ServiceProvider, or whatever is easiest)
- [ ] Wherever data is filtered out of a message, include a message in place of the data letting folks know they can enable the flag to get the data (and a warning that it may result in logging sensitive info) 

We should try and add this in a way that makes it easy to use it for future logging that we add.

rowanmiller:
@lukewaters please work with @anpete to get some direction on this one

divega:
@rowanmiller @anpete @ajcvickers need to find an owner for this.

rowanmiller:
Didn't we handle this with a log level that applies across all technologies?

anpete:
This should probably be replaced by a bug to switch from logging to event notification.

lukewaters:
@rowanmiller I did add a Debug log level and a minimum level to logging a couple months ago, the minimum level prevented more detailed level messages from being distributed to registered loggers for security purposes

divega:
@lukewaters Can we assume that the states of check boxes in the check list is up to date? Thanks!

@anpete Yeah probably. I see that as part of what the new owner would need to figure out. Are you volunteering? :smile:

anpete:
Yes I am. :-)

lukewaters:
@divega The flag is in place, and I Added the LogParamenters flag in [RelationalLoggingExtensions](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Extensions/RelationalLoggerExtensions.cs) I think EF also logs a message whenever DbContext is initialized

divega:
Thanks again!

divega:
@anpete all yours!

PR #1809, opened 2015-03-12
Added the ability to log query parameters

lukewaters:
Part of issue #1374

### DIFF
--- a/src/EntityFramework.Core/Properties/Strings.resx
+++ b/src/EntityFramework.Core/Properties/Strings.resx
@@ -468,6 +468,9 @@
   <data name="NavigationToSelfDuplicate" xml:space="preserve">
     <value>The navigation property '{navigation}' cannot be used for both ends of the relationship. Either specify it on the principal side or the dependent side.</value>
   </data>
+  <data name="DebugLogWarning" xml:space="preserve">
+    <value>{debug} level logging is enabled. At this level, Entity Framework will log sensitive application data such as SQL parameter values. To hide this information configure {minimumLevel} to {recommendedLevel}</value>
+  </data>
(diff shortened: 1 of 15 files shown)


src/EntityFramework.Core/Properties/Strings.resx
```diff
@@ -462,6 +462,9 @@
   <data name="NavigationToSelfDuplicate" xml:space="preserve">
     <value>The navigation property '{navigation}' cannot be used for both ends of the relationship. Either specify it on the principal side or the dependent side.</value>
   </data>
+  <data name="DebugLogWarning" xml:space="preserve">
+    <value>Debug level logging is enabled.  At this level, Entity Framework will log sensitive application data such as Sql parameter values.  To hide this information configure ILoggerFactory.MinimumLevel to LogLevel.Verbose</value>
```
Eilon: A few little things:

1. Only one space after a period
2. Recommend using parameter substitutions in the string and passing in either `nameof(Foo)` or whatever from within the code.
3. The term "SQL" should be in all caps.

lukewaters: addressing 2. this is expected to be a static message, I don't expect anything to change, but I can parameterize the names of the log levels if you think that is necessary

Eilon: It makes it easier for the localizers if every bit of the string is actually localizable text. That way we don't have to "lock out" strings to prevent them from being overlocalized.

lukewaters:
@Eilon I've parameterized the values that I don't think we should localize, does this look good?

Eilon:
:shipit: (sorry for delay, was on vacation)

PR #1812, opened 2015-03-12
Use generic ReadValue in update pipeline and key factories

ajcvickers:
This change allows cached delegates to be created that will call the generic IValueReader.ReadValue method using the type of the property. This means that when a RelationalTypedValueReader is used the generic GetFieldValue method will be called on the underlying data reader, thereby giving the provider the contextual information needed to create objects of the correct type in cases where this cannot be determined by the database metadata alone.

This mechanism is being used in the update pipeline and for key generation. It is not yet being used in query or for reading shadow state properties.

### DIFF
--- a/src/EntityFramework.Relational/Update/ModificationCommand.cs
+++ b/src/EntityFramework.Relational/Update/ModificationCommand.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.ChangeTracking.Internal;
+using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Relational.Metadata;
 using Microsoft.Data.Entity.Utilities;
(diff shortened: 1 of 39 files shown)


src/EntityFramework.Relational/Update/ModificationCommand.cs
```diff
@@ -141,14 +147,12 @@ public virtual void PropagateResults([NotNull] IValueReader reader)
...
-            // Issue #771
             // Note that this call sets the value into a sidecar and will only commit to the actual entity
             // if SaveChanges is successful.
-            var columnOperations = ColumnModifications.Where(o => o.IsRead).ToArray();
-            for (var i = 0; i < columnOperations.Length; i++)
+            var columnOperations = ColumnModifications.Where(o => o.IsRead).ToList();
+            for (var i = 0; i < columnOperations.Count; i++)
             {
-                columnOperations[i].Value = reader.IsNull(i) ? null : reader.ReadValue<object>(i);
+                columnOperations[i].Value = columnOperations[i].BoxedValueReader.ReadValue(reader, i);
```
AndriySvyryd: Why do we need to enumerate columnOperations twice?

AndriySvyryd:
:shipit:

src/EntityFramework.Relational/Update/ModificationCommand.cs
ajcvickers: Changed.

PR #1820, opened 2015-03-13
Value generation model validator

AndriySvyryd:
Add a model validation rule to ensure there is one and only one root principal property with value generation on add enabled in a foreign key chain.


ajcvickers:
:shipit:

PR #1826, opened 2015-03-14
RevEng: Moving and simplifying look up of assembly from name

lajones:
@bricelam - just tidying up so that we pass the assembly name and look up the actual assembly in DatabaseTool instead. As agreed earlier today.

### DIFF
--- a/src/EntityFramework.Commands/DatabaseTool.cs
+++ b/src/EntityFramework.Commands/DatabaseTool.cs
@@ -16,7 +16,10 @@ namespace Microsoft.Data.Entity.Commands
 {
     public class DatabaseTool
     {
+        public static readonly string _defaultReverseEngineeringProviderAssembly = "EntityFramework.SqlServer.Design";
+
         private readonly ServiceProvider _serviceProvider;
+        private readonly LazyRef<ILogger> _logger;
 
         public DatabaseTool(
             [CanBeNull] IServiceProvider serviceProvider,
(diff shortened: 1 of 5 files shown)


src/EntityFramework.Commands/Properties/Strings.resx
```diff
@@ -60,122 +59,119 @@
...
     <value>The current CSharpMigrationOperationGenerator cannot scaffold operations of type '{operationType}'. Configure your services to use one that can.</value>
   </data>
-    <data name="UnknownLiteral" xml:space="preserve">
+  <data name="UnknownLiteral" xml:space="preserve">
     <value>The current CSharpHelper cannot scaffold literals of type '{literalType}'. Configure your services to use one that can.</value>
   </data>
+  <data name="CannotFindAssembly" xml:space="preserve">
+    <value>Unable to find asssembly with name {assemblyName}.</value>
+  </data>
 </root>
```
lajones: Not sure why all the whitespace changes. The 3 lines above are the only real change.

src/EntityFramework.Commands/DatabaseTool.cs
```diff
@@ -27,26 +30,33 @@ public class DatabaseTool
...
-            Check.NotNull(providerAssembly, nameof(providerAssembly));
+            Check.NotNull(providerAssemblyName, nameof(providerAssemblyName));
             Check.NotEmpty(connectionString, nameof(connectionString));
             Check.NotEmpty(rootNamespace, nameof(rootNamespace));
             Check.NotEmpty(projectDir, nameof(projectDir));
 
+            var assembly = GetReverseEngineerProviderAssembly(providerAssemblyName);
+            if (assembly == null)
+            {
+                _logger.Value.LogWarning(Strings.CannotFindAssembly(providerAssemblyName));
```
bricelam: You can always throw exceptions to "log" errors.

src/EntityFramework.Commands/Properties/Strings.resx
bricelam: Someone probably ran R# Code Clean-up which formats it differently than VS does by default.

bricelam:
:shipit:

bricelam:
Can we remove `Microsoft.SqlServer.Design` from this project's dependencies now?

src/EntityFramework.Commands/DatabaseTool.cs
lajones: Did it this was to be consistent with some logging that MigrationTool does. I don't really mind which way though. Would you prefer exceptions? Any particular kind of exception?

lajones:
Re  'Microsoft.SqlServer.Design'  - it's still hard-coded to use that provider and only that provider. This was per agreement with @rowanmiller a while back. In future will want to add code to the psm to allow using the DTE project to pass in the project's references and using those to look up the provider assembly but that's on the back-burner for right now.

src/EntityFramework.Commands/DatabaseTool.cs
lajones: Looks like ContextTool uses InvalidOperationException. So I'll do the same.

lajones:
Checked in with commit c2e146a.

PR #1827, opened 2015-03-14
Add IsConfigured property to DbContextOptionsBuilder

ajcvickers:
Add IsConfigured property which returns true if any UseXxx method has been called--in other words, if any extension has been added


divega:
:shipit:

ajcvickers:
Merged

Issue #1830, opened 2015-03-14
Query :: simplify negated expressions in query predicate


maumar:
We can simplify some queries containing negation:

!(!a) => a
!a == b => a != b
!a != b => a == b
!(a && b) => !a || !b
!(a || b) => !a && !b
!(a>b) => a <=b

maumar:
Fixed in 723b8b5cec00cd261dc6b8d1b68a05fe77c613de

Issue #1760, opened 2015-03-05
Query :: complex filter containing boolean expressions may produce invalid sql


maumar:
query:

```
 products.Where(p => p.Discontinued == (p.ProductID > 100))
```

EF6 translates this to:

```
SELECT 
      [Extent1].[ProductID] AS [ProductID], 
      [Extent1].[ProductName] AS [ProductName], 
      [Extent1].[SupplierID] AS [SupplierID], 
      [Extent1].[CategoryID] AS [CategoryID], 
      [Extent1].[QuantityPerUnit] AS [QuantityPerUnit], 
      [Extent1].[UnitPrice] AS [UnitPrice], 
      [Extent1].[UnitsInStock] AS [UnitsInStock], 
      [Extent1].[UnitsOnOrder] AS [UnitsOnOrder], 
      [Extent1].[ReorderLevel] AS [ReorderLevel], 
      [Extent1].[Discontinued] AS [Discontinued]
      FROM [dbo].[Products] AS [Extent1]
      WHERE [Extent1].[Discontinued] = (CASE WHEN ([Extent1].[ProductID] > 100) THEN cast(1 as bit) WHEN ( NOT ([Extent1].[ProductID] > 100)) THEN cast(0 as bit) END)
```

maumar:
fixed in 723b8b5cec00cd261dc6b8d1b68a05fe77c613de

PR #1831, opened 2015-03-14
Fix to Issues #1760 " - complex filter containing boolean expressions may produce invalid sql" and #1830 -  simplify negated expressions in query predicate

maumar:
#1760 - Query :: complex filter containing boolean expressions may produce invalid sql
#1830 - Query :: simplify negated expressions in query predicate

#1760:
Problem is when comparing expressions which are not simple (constant, parameter, column), e.g
products.Where(p => p.Discontinued == (p.ProductID > 100)). Naive translation produces invalid sql.

Fix is to expand those comparisons like so:
(a == b) => (a && b) || (!a && !b)
(a != b) => (!a && b) || (a && !b)

This is done before sql generation for perf reasons (result tree gets cached, sql does not).

#1830:
Since #1760 introduces additional (negated) expressions, we want to optimize them as much as possible.
This checkin handles the following optimizations:
!(!a) => a
!a == b => a != b
!a != b => a == b
!(a && b) => !a || !b
!(a || b) => !a && !b
!(a > b) => a <= b
!(a >= b) => a < b
!(a < b) => a >= b
!(a <= b) => a > b

Also fix some minor bugs where necessary parentheses would not be produced during sql generation

### DIFF
--- a/test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
+++ b/test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
@@ -911,6 +911,62 @@ namespace Microsoft.Data.Entity.FunctionalTests
         }
 
         [Fact]
+        public virtual void Where_bool_member_compared_to_binary_expression()
+        {
+            AssertQuery<Product>(ps => ps.Where(p => p.Discontinued == (p.ProductID > 50)), entryCount: 44);
+        }
+
+        [Fact]
+        public virtual void Where_not_bool_member_compared_to_not_bool_member()
+        {
+            AssertQuery<Product>(ps => ps.Where(p => !p.Discontinued == !p.Discontinued), entryCount: 77);
+        }
+
+        [Fact]
+        public virtual void Where_negated_boolean_expression_compared_to_another_negated_boolean_expression()
+        {
+            AssertQuery<Product>(ps => ps.Where(p => !(p.ProductID > 50) == !(p.ProductID > 20)), entryCount: 47);
+        }
+
+        [Fact]
+        public virtual void Where_not_bool_member_compared_to_binary_expression()
+        {
+            AssertQuery<Product>(ps => ps.Where(p => !p.Discontinued == (p.ProductID > 50)), entryCount: 33);
+        }
+
+        [Fact]
+        public virtual void Where_bool_parameter_compared_to_binary_expression()
+        {
+            bool prm = true;
+            AssertQuery<Product>(ps => ps.Where(p => (p.ProductID > 50) != prm), entryCount: 50);
+        }
+
+        [Fact]
+        public virtual void Where_bool_member_and_parameter_compared_to_binary_expression_nested()
+        {
+            bool prm = true;
+            AssertQuery<Product>(ps => ps.Where(p => p.Discontinued == ((p.ProductID > 50) != prm)), entryCount: 33);
+        }
+
+        [Fact]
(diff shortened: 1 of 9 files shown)


maumar:
@anpete

maumar:
@divega

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -1644,6 +1644,105 @@ public override void Where_bool_member_in_complex_predicate()
...
+WHERE ([p].[Discontinued] = 0 AND ([p].[ProductID] > 50)) OR ([p].[Discontinued] = 1 AND ([p].[ProductID] <= 50))",
+                Sql);
+        }
+
+        public override void Where_not_bool_member_compared_to_not_bool_member()
+        {
+            base.Where_not_bool_member_compared_to_not_bool_member();
+
+            Assert.Equal(
+                @"SELECT [p].[Discontinued], [p].[ProductID], [p].[ProductName], [p].[UnitsInStock]
```
maumar: this could be optimized to p.Discontinued = p.Discontinued if I run negation optimizer before equality expansion. Currently running it just after the expansion, since it optimizer more real-life cases. Running it twice seems like too much cost for the benefit, but could be done if you guys think it makes sense.

divega: Couldn't we optimize out the whole WHERE clause for this case? :smile: I don't have a strong opinion on executing the rules more than once. In principle we could design it so we can apply simplifications in a loop until either the tree doesn't change or we go over some threshold, but not sure about the cost/benefit ratio here.

maumar: We could, I was thinking to add tautology/redundant constants optimization as well. In this case however, since the properties are wrapped in a 'Not' clause, we would still have to apply 'Not' elimination before tautology elimination to optimize this out (or make tautology removal more sophisticated)

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -911,6 +911,62 @@ public virtual void Where_bool_member_in_complex_predicate()
...
+
+        [Fact]
+        public virtual void Where_bool_member_and_parameter_compared_to_binary_expression_nested()
+        {
+            bool prm = true;
+            AssertQuery<Product>(ps => ps.Where(p => p.Discontinued == ((p.ProductID > 50) != prm)), entryCount: 33);
+        }
+
+        [Fact]
+        public virtual void Where_de_morgan_or_optimizated()
```
anpete: optimizated? Cool!

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -672,12 +672,12 @@ public override void All_top_level()
 
             Assert.Equal(
                 @"SELECT CASE WHEN (
-    NOT EXISTS (
+    NOT(EXISTS (
```
anpete: Should have a space after NOT

anpete:
:shipit: :trophy:

maumar:
merged 723b8b5cec00cd261dc6b8d1b68a05fe77c613de

Issue #1819, opened 2015-03-13
Migrations: NullRef when no migration name supplied when adding a migrations (k cmd)


rowanmiller:
`k ef migration add --context UnicornStoreContext`

> System.NullReferenceException: Object reference not set to an instance of an object.
>    at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.ShowHelp(String commandName)
>    at Microsoft.Data.Entity.Commands.Program.AddMigration(String name, String context, String startupProject)
>    at Microsoft.Data.Entity.Commands.Program.<>c__DisplayClass8_1.<Main>b__6()
>    at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
>    at Microsoft.Data.Entity.Commands.Program.Main(String[] args)

Not sure if this is something we should be doing differently in our code or if it's a bug in the cmd line library we use.

bricelam:
This is a dupe of #980 (Closed as fixed by @lukewaters)

lukewaters:
I'll look into this

rowanmiller:
I have beta4-12528 referenced (3 days old) and am seeing the issue

lukewaters:
fixed in #1834

PR #1834, opened 2015-03-16
fixing up help

lukewaters:
fix for #1819


lukewaters:
@bricelam can you take a look?

bricelam:
:shipit: This will work for the specific case. We should also look at updating CommandLineUtils with a more general solution that can be reused by all its consumers.

PR #1835, opened 2015-03-16
Use generic ReadValue when reading shadow properties

ajcvickers:
Query still to do.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/EntityEntryMetadataServices.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/EntityEntryMetadataServices.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Metadata;
 
 namespace Microsoft.Data.Entity.ChangeTracking.Internal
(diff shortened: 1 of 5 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/EntityEntryMetadataServices.cs
```diff
@@ -51,5 +55,8 @@ public virtual EntityKey CreateKey(IEntityType entityType, IReadOnlyList<IProper
             => _entityKeyFactorySource
                 .GetKeyFactory(properties)
                 .Create(entityType, properties, propertyAccessor);
+
+        public virtual object ReadValueFromReader(IValueReader valueReader, IProperty property)
```
AndriySvyryd: Can this be called just ReadValue?

AndriySvyryd:
:shipit:

src/EntityFramework.Core/ChangeTracking/Internal/EntityEntryMetadataServices.cs
ajcvickers: It can, but when I had it like this I ended up with runtime bugs because the overload that was taking object was called instead of their being a compile error. This was because my code was wrong, but I decided to change the name to make it not compile in such cases.

AndriySvyryd: Since it will result in an exception anyway if the wrong one is used I would prefer the shorter name, but I don't feel strongly about it.

ajcvickers: I'd rather have compile-time checking than a runtime exception.

AndriySvyryd: ok

PR #1840, opened 2015-03-17
Add contract interfaces for more services

ajcvickers:
This makes all core services use contract interfaces, except ChangeTracker which is an API class and I will remove from DI resolution.


anpete:
:shipit:

PR #1848, opened 2015-03-18
Added key batch generation functional test

lukewaters:
This shows the expected behavior described #1540


test/EntityFramework.SqlServer.FunctionalTests/CommandConfigurationTests.cs
```diff
@@ -285,6 +286,52 @@ public void Constructed_update_statement_uses_commandTimeout_overriding_configur
             }
         }
 
+        [Theory]
+        [InlineData(51, 6)]
+        [InlineData(50, 5)]
+        [InlineData(20, 5)]
+        [InlineData(2, 2)]
+        public void Keys_generated_in_batches(int count, int expected)
```
lajones: Nit: better to name the parameters what it is a "count" of and similarly for "expected". Otherwise looks good.

Do you need other tests where you reset the Batch Size and show that the above changes?

PR #1853, opened 2015-03-18
adding type to count, removing longCount

lukewaters:
This addresses a comment from @anpete


src/EntityFramework.Relational/Query/Expressions/CountExpression.cs
```diff
@@ -15,8 +16,17 @@ public class CountExpression : ExtensionExpression
         public CountExpression()
             : base(typeof(int))
```
anpete: Sorry if I was unclear - You want to pass the type to the base ctor. All Expressions already have a Type property.

src/EntityFramework.SqlServer/Query/SqlServerQueryGenerator.cs
```diff
@@ -11,5 +14,21 @@ protected override string DelimitIdentifier(string identifier)
...
+        {
+            Check.NotNull(countExpression, nameof(countExpression));
+
+            if (countExpression.CountReturnType == typeof(long))
+            {
+                Sql.Append("COUNT_BIG(*)");
+            }
+            else
+            {
+                Sql.Append("COUNT(*)");
```
anpete: Call base.

anpete:
:shipit: after changes.

PR #1857, opened 2015-03-19
Metadata: Align entity name properties with CLR

### LINKED ISSUE #1613: Metadata: Align entity name properties with CLR
We have EntityType.Name and EntityType.SimpleName. Seems like it would be better just to match CLR here: FullName and Name.

smitpatel:
resolves #1613

### DIFF
--- a/src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
+++ b/src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
@@ -237,7 +237,7 @@ namespace Microsoft.Data.Entity.Commands.Migrations
 
             var tableName = entityType.Relational().Table;
             var schema = entityType.Relational().Schema;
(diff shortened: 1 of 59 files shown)


ajcvickers:
@rowanmiller Can you check the API review notes on this one? I thought we came to a different decision.

rowanmiller:
We decided to leave Name as-is and rename SimpleName to DisplayName and make it an extension method (though personally I don't think the extension method part is critical)

ajcvickers:
Thanks Rowan. I think it should be an extension method because then any implementation of the interface can use the same implementation.

smitpatel:
@rowanmiller - Can you update #1613  or close it and file a new issue for this task?

rowanmiller:
@smitpatel - Closed. @anpete is going to be making that changes from the API review.

PR #1859, opened 2015-03-19
React to hosting

HaoK:
@davidfowl this looks a bit weird, but I think it works...

cc @divega

### DIFF
--- a/src/EntityFramework.Commands/ContextTool.cs
+++ b/src/EntityFramework.Commands/ContextTool.cs
@@ -10,9 +10,13 @@ using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 
 #if DNX451 || DNXCORE50
+using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.FeatureModel;
 using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.Framework.ConfigurationModel;
 using Microsoft.Framework.DependencyInjection;
-using Microsoft.Framework.OptionsModel;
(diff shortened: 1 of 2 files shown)


src/EntityFramework.Commands/ContextTool.cs
```diff
@@ -52,6 +52,35 @@ private static DbContext TryCreateContextFromStartup(Type type)
...
+            {
+                private readonly Func<IFeatureCollection, Task> _application;
+
+                public StartInstance(Func<IFeatureCollection, Task> application)
+                {
+                    _application = application;
+                }
+
+                public void Dispose()
+                {
```
divega: Nothing to dispose I guess. Is this what looks weird? :smile:

HaoK: Also needing to write all this infrastructure just to get a Hosting built ApplicationServices out.  Lots of repos are needing this exact usage (creating a hosting built ISP)

HaoK:
Merged 167e85a65d96cd72715e8491c6e45309e9a46b97

src/EntityFramework.Commands/ContextTool.cs
davidfowl: We might need to look at this a bit. Seems like we might need to be able to boot the hostingEngine without a server (that's what this is). If you replace the IServerFactory you get the same effect.

divega: Good to hear this is a common need. Is it always about desig-time?

divega: I am going to guess testing is another reason.

HaoK: Yeah mostly its testing I think

PR #1860, opened 2015-03-19
[SMALL] Additional fix to Issue 1747 - Multi-select with Contains() does not generate SQL

maumar:
Problem is we were not handling the case where user would pass empty collection as argument to contains - we would produce invalid sql.
Fix is to check for number of arguments to InExpression before producing a relevant sql.
This has to happen during sql generation, since we need to know the values of the parameters.

### DIFF
--- a/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
+++ b/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
@@ -273,44 +273,57 @@ namespace Microsoft.Data.Entity.Relational.Query.Sql
 
         public virtual Expression VisitInExpression(InExpression inExpression)
         {
-            VisitExpression(inExpression.Column);
+            var inValues = ProcessInExpressionValues(inExpression.Values);
+            if (inValues.Count > 0)
+            {
+                VisitExpression(inExpression.Column);
 
-            _sql.Append(" IN (");
+                _sql.Append(" IN (");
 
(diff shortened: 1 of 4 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -273,44 +273,57 @@ public virtual Expression VisitMaxExpression(MaxExpression maxExpression)
...
+            }
 
             return inExpression;
         }
 
         protected virtual Expression VisitNotInExpression(InExpression inExpression)
         {
-            VisitExpression(inExpression.Column);
+            var inValues = ProcessInExpressionValues(inExpression.Values);
+            if (inValues.Count() > 0)
```
anpete: Count

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -273,44 +273,57 @@ public virtual Expression VisitMaxExpression(MaxExpression maxExpression)
 
         public virtual Expression VisitInExpression(InExpression inExpression)
         {
-            VisitExpression(inExpression.Column);
+            var inValues = ProcessInExpressionValues(inExpression.Values);
+            if (inValues.Count() > 0)
```
anpete: Count

anpete:
:shipit:

PR #1865, opened 2015-03-19
Enhance the ForeignKeyPropertyDiscoveryConvention to choose the best...

AndriySvyryd:
... candidate for the principal end of a non-self-referencing one-to-one relationship.

The dependent entity type will be choosen according to the following heuristics:
1. Entity type with non-shadow properties that can be used in the foreign key and are not in the primary key
2. Entity type without navigation
3. Entity type which primary key is not principal in any relationship
4. Second entity type by alphabetical order


ajcvickers:
:shipit:

Issue #1800, opened 2015-03-11
Remove usage of IConfiguration from EF


HaoK:
I'm making a pass through and removing IConfiguration from all of the other AddXyz() framework methods.  They were only being used to bind against a respective XyzOptions class.  This is still possible, but via `services.ConfigureXyz(config)` instead.  Service replacement via IConfiguration/ServiceDescriber has been cut...

EF looks to be doing something special, so EF can to decide for itself if it makes sense to continue taking IConfiguration or not as part of AddEntityFramework()...

cc @divega

divega:
@rowanmiller and I talked and agreed on a proposal which removes the current usage of IConfiguration in the EF7 stack. The default template will explicitly read the connection string configuration key and pass its value to the `DbContextOptionsBuilder`, e.g.:

``` C#
services.AddEntittyFramework()
    .AddSqlServer()
    .AddDbContext<MyDbContext>(options => 
        options.UserSqlServer(configuration.Get("MyConnectionString"));
```

This does not prevent applications from using other patterns to flow information from configuration into DbContext if the choose to do so. 

__Background:__
EF is different to the other frameworks in that there isn't framework-wide configuration and what it is interesting to configure is how EF operates on a specific model through a specific provider, which is usually embodied in a derived DbContext type. 

We did not get to implement service replacement (see #829), so until now the reason `AddEntityFramework()` was taking an IConfiguration argument was to flow it down to subsequent `AddDbContext<TContext>()` methods where the configuration settings for TContext were actually captured. 

The primary way to configure a DbContext is through the `DbContextOptionsBuilder<TContext>` class that can be injected at construction or configured in the `OnConfiguring()` method of the context. 

While it is interesting to be able to configure a context based on external configuration settings it isn't a high priority at the moment to do so in a structured way. We also want to avoid growing a large footprint in configuration as we did in EF6.

PR #1867, opened 2015-03-20
Remove use of IConfiguration

ajcvickers:
This means all configuration is now code-based only.

See Issue #1800


AndriySvyryd:
:shipit:

PR #1870, opened 2015-03-20
Make AsyncFromSqlQueryTestBase abstract

ajcvickers:
So that TestDriven.NET runner does not attempt to run the tests and fail.


AndriySvyryd:
:shipit:

PR #1871, opened 2015-03-20
Add params object array overload to FromSql

mikary:
I'm thinking we can normalize the different options (params array, anonymous type, and ```Dictionary<string, object>```) into a named parameter string in relational (where we have the extensions) and convert to the provider-specific format in the SqlQueryGenerator.

This is a first pass with just the ```FromSql(string, params object[])``` extension
@anpete @divega

### DIFF
--- a/src/EntityFramework.Relational/Query/Annotations/FromSqlAnnotation.cs
+++ b/src/EntityFramework.Relational/Query/Annotations/FromSqlAnnotation.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 
@@ -9,17 +10,28 @@ namespace Microsoft.Data.Entity.Relational.Query.Annotations
     public class FromSqlAnnotation
     {
         public FromSqlAnnotation([NotNull] string sql)
+            : this(sql, new object[0])
+        {
+        }
+
+        public FromSqlAnnotation([NotNull] string sql, [NotNull] object[] parameters)
         {
             Check.NotEmpty(sql, nameof(sql));
+            Check.NotNull(parameters, nameof(parameters));
 
             Sql = sql;
+            Parameters = parameters;
         }
 
         public virtual string Sql { get; }
 
+        public virtual object[] Parameters { get; }
+
         public override string ToString()
         {
-            return Sql;
+            return string.Format("\"{0}\" ({1})",
+                Sql,
+                string.Join(", ", Parameters.Select(p => p.ToString())));
         }
     }
 }
--- a/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
+++ b/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
@@ -186,7 +186,20 @@ namespace Microsoft.Data.Entity.Relational.Query.Sql
 
             using (_sql.Indent())
             {
-                _sql.AppendLines(rawSqlDerivedTableExpression.Sql);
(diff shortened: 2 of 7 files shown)


src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -21,5 +22,32 @@ public static IQueryable<TEntity> FromSql<TEntity>([NotNull]this DbSet<TEntity>
...
+            Check.NotNull(dbSet, nameof(dbSet));
+            Check.NotEmpty(sql, nameof(sql));
+            Check.NotNull(parameters, nameof(parameters));
+
+            var parameterDictionary = new Dictionary<string, object>();
+            var substitutions = new List<string>();
+
+            for (var index = 0; index < parameters.Count(); index++)
+            {
+                var parameterName = string.Format("P{0}", index);
```
anpete: nit: small "p"

src/EntityFramework.Relational/Query/Annotations/FromSqlAnnotation.cs
```diff
@@ -13,13 +15,29 @@ public FromSqlAnnotation([NotNull] string sql)
             Check.NotEmpty(sql, nameof(sql));
 
             Sql = sql;
+            Parameters = new Dictionary<string, object>();
```
anpete: call this()

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -21,5 +22,32 @@ public static IQueryable<TEntity> FromSql<TEntity>([NotNull]this DbSet<TEntity>
...
+
+                parameterDictionary.Add(
+                    parameterName,
+                    parameters[index]);
+
+                substitutions.Add(string.Format("{{{0}}}", parameterName));
```
anpete: Keep the ordinals in the string and call string.Format in SQL gen.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -186,7 +186,17 @@ public virtual Expression VisitRawSqlDerivedTableExpression([NotNull] RawSqlDeri
...
             {
-                _sql.AppendLines(rawSqlDerivedTableExpression.Sql);
+                var sql = rawSqlDerivedTableExpression.Sql;
+
+                foreach (var parameter in rawSqlDerivedTableExpression.Parameters)
+                {
+                    _parameters.Add(parameter.Key);
+                    _parameterValues.Add(parameter);
+
+                    sql = sql.Replace("{" + parameter.Key + "}", ParameterPrefix + parameter.Key);
```
anpete: This isn't safe in general. See comment below.

anpete:
Approach looks good.

anpete:
:shipit:

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -186,7 +186,20 @@ public virtual Expression VisitRawSqlDerivedTableExpression([NotNull] RawSqlDeri
 
             using (_sql.Indent())
             {
-                _sql.AppendLines(rawSqlDerivedTableExpression.Sql);
+                var substitutions = new List<string>();
```
anpete: nit: You could make this an array as it is fixed size. Then you can avoid the alloc below.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -186,7 +186,20 @@ public virtual Expression VisitRawSqlDerivedTableExpression([NotNull] RawSqlDeri
...
+
+                for (var index = 0; index < rawSqlDerivedTableExpression.Parameters.Count(); index++)
+                {
+                    var parameterName = string.Format("p{0}", index);
```
anpete: nit: '+' is fine here for concat - and is faster.

PR #1872, opened 2015-03-20
Sequence as default value generator for SqlServer

### LINKED ISSUE #1811: SqlServer: Default Value Generator
Presently for PK with single property, we put Default as value generation strategy.
This Default annotation translates to value generation defined on model or to Identity.
Instead of setting Identity hard coded, we should have some way to define what should be the default value generation strategy.

smitpatel:
Resolves #1811 
Add Model Conventions support
Set Sequence as default value generator for SqlServer by convention

### DIFF
--- a/src/EntityFramework.Core/EntityFramework.Core.csproj
+++ b/src/EntityFramework.Core/EntityFramework.Core.csproj
@@ -148,6 +148,7 @@
     <Compile Include="DbContext.cs" />
     <Compile Include="DbContextOptionsBuilder.cs" />
     <Compile Include="DbContextOptionsBuilder`.cs" />
+    <Compile Include="Metadata\ModelConventions\IModelConvention.cs" />
     <Compile Include="Query\AnnotateQueryExtensions.cs" />
     <Compile Include="Infrastructure\CoreOptionsExtension.cs" />
     <Compile Include="Infrastructure\DbContextOptions.cs" />
--- a/src/EntityFramework.Core/Metadata/Internal/InternalModelBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalModelBuilder.cs
@@ -185,5 +185,10 @@ namespace Microsoft.Data.Entity.Metadata.Internal
 
             return roots;
         }
+
+        public virtual InternalModelBuilder Initialize()
+        {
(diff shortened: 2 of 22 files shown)


smitpatel:
Unit tests are not added yet.

AndriySvyryd:
Looks good so far. Remember to test other repos (run "build install" for EF then run "build" in Universe)

smitpatel:
Added tests.
All repos in Universe passing except Helios. (requires older MSBuild which I don't have in my machine)

AndriySvyryd:
:shipit: 
This will likely require dropping databases, so send out an email explaining what changed and how to drop them.

rmja:
@AndriySvyryd, @smitpatel, this PR broke my setup. Every time I try and do an insert a

    SELECT NEXT VALUE FOR DefaultSequence

is sent to the database, failing with `DefaultSequence` not being an object. Is there something I have missed?

divega:
It is likely that you need to drop and recreate the database.

rmja:
@divega, I am not following. I am working on a copy of our production database. Not something generated by ef through migrations.

divega:
@rmja In that case you might need to opt out of using sequences by explicitly specifying that you want to use SQ Server IDENTITY, e.g.:

``` C#
modelBuilder.Entity<Entity>().Property(e => e.Property).ForSqlServer().UseIdentity();
```

rmja:
Thanks @divega, it works! I thought, when trying to solve this myself, that Identity in `UseIdentity()` had something to do with http://www.asp.net/identity. My bad.

divega:
Not your bad :smile: Identity is an overloaded term, so it can unfortunately be a bit confusing. Glad to help.

ajcvickers:
@rmja Note that if you always want SQL Server to use Identity, and not sequences, for properties such as integer primary keys, then you can just call UseIdentity for your model, rather than doing it for each individual property:
```
modelBuilder.ForSqlServer().UseIdentity();
```
The code for this has been updated recently, so please let us know if you try this and it doesn't work.

rmja:
@ajcvickers Yes I found that, and it seems to work fine with the current bits.

luisrudge:
Why was the sequence chosen to be the default strategy?

PR #1873, opened 2015-03-20
Match FK properties ending in "Id"

### LINKED ISSUE #1842: FK property not detected by convention when named [navigation_name]Id
Pretty sure we didn't every detect this one on EF6 (haven't checked yet) but we certainly got a number of requests to detect it. EF7 is probably a good time to add it in.

In the following code doesn't detect `Product.PrimaryCategoryId` as the FK because it doesn't follow [navigation_name][primary_key_name] convention. To have it detected it would need to be `Product.ParentCategoryCategoryId`.

Currently we introduce a shadow FK property (called `ParentCategoryCategoryId`).

```
    public class Product
    {
        public int ProductId { get; set; }
        public string DisplayName { get; set; }
        public decimal MSRP { get; set; }

        public int PrimaryCategoryId { get; set; }
        public Category PrimaryCategory { get; set; }
    }

    public class Category
    {
        public int CategoryId { get; set; }
        public string DisplayName { get; set; }

        public List<Product> Products { get; set; }
    }
```

AndriySvyryd:
Make ForeignKeyPropertyDiscoveryConvention match properties that are named [navigationName]Id, [principalName]Id and Id if the property isn't the primary key on the dependent, the principal has a simple primary key and no better matches for FK properties are found.

Fixes #1842

### DIFF
--- a/src/EntityFramework.Core/Metadata/ModelConventions/ForeignKeyPropertyDiscoveryConvention.cs
+++ b/src/EntityFramework.Core/Metadata/ModelConventions/ForeignKeyPropertyDiscoveryConvention.cs
@@ -162,24 +162,28 @@ namespace Microsoft.Data.Entity.Metadata.ModelConventions
             var foreignKeyProperties = new List<Property>();
             foreach (var referencedProperty in propertiesToReference)
             {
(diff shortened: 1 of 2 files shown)


ajcvickers:
:shipit:

PR #1875, opened 2015-03-20
Remove parameterless and internal ModelBuilder constructors

ajcvickers:
Per API review decisions, a parameterless ModelBuilder constructor that creates a ModelBuilder without conventions would be a pit of failure. Therefore, this constructor is removed and now a convention set must always be passed, which may be empty. Also verified it is pretty easy to create a ModelBuilderFactory for a given provider and get a convention builder from that.


AndriySvyryd:
:shipit:

Issue #1709, opened 2015-02-25
Simplify pattern for Include around collections/references


rowanmiller:
Currently you can only use ThenInclude after landing on a collection nav but you can't use it after landing on a reference in the previous Include call. This is super hard to describe/document... and that's a good sign it will be confusing.

Historically, trying to use overload resolution to detect collection/reference has never worked. So the proposal is to do the same Reference/Collection named pairs of methods we use elsewhere in the stack:

```
query.IncludeCollection(blog => blog.Posts).ThenIncludeReference(post => post.Author)...
```

We would still allow the dotted syntax but it wouldn't be the primary API we push folks too.

When discussing this we should also consider using the Reference/Collection names in the relationship API (rather than One/Many).

**Once decided on this we need to add API docs. I removed them for the moment because the current pattern is hard to describe and if we are going to change it I don't want to waste time working out how to word it :smile:**

divega:
> Historically, trying to use overload resolution to detect collection/reference has never worked
We actually had it working in this case. We just removed the extra overload because it seemed redundant to me.
>  Reference/Collection named pairs of methods we use elsewhere in the stack
It is not nice in this case it is not really needed.

divega:
We discussed this and we will try IncludeCollection/IncludeReference (or is it IncludeMany/IncludeOne for consistency?) because that accounts for the corner cases in which the CLR type does not matche the conceptual cardinality of the navigation property. We should consider doing #1481 at once since we are touching these methods and it might get harder to do it later.

divega:
I would like to bring this back to discussion. The more I think about it the more I believe this is going to complicate much more than simplify things.

divega:
I confirmed that this won't be necessary and that all we need to do is bring back the following "reference" overload of `ThenInclude()` which was removed in #1491:

``` C#
public static IIncludableQueryable<TEntity, TProperty> ThenInclude<TEntity, TPreviousProperty, TProperty>(
    this IIncludableQueryable<TEntity, TPreviousProperty> source,
    Expression<Func<TPreviousProperty, TProperty>> navigationPropertyPath)
    where TEntity : class
{
``` 

This is in addition to the existing "collection" overload:

``` C#
public static IIncludableQueryable<TEntity, TProperty> ThenInclude<TEntity, TPreviousProperty, TProperty>(
    this IIncludableQueryable<TEntity, ICollection<TPreviousProperty>> source,
[...]

PR #1880, opened 2015-03-22
Fix to #1709 - Simplify pattern for Include around collections/references

maumar:
Added ThenInclude overload for a reference. Also added tests that were removed when the overload was removed previously.

### DIFF
--- a/src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
+++ b/src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
@@ -8,7 +8,6 @@ using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
-using Microsoft.Data.Entity;
 using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Query;
 using Microsoft.Data.Entity.Utilities;
@@ -21,7 +20,7 @@ namespace System.Linq
     {
         #region Any/All
 
-        private static readonly MethodInfo _any = GetMethod("Any");
+        private static readonly MethodInfo _any = GetMethod(nameof(Queryable.Any));
 
         /// <summary>
         ///     Asynchronously determines whether a sequence contains any elements.
@@ -52,7 +51,7 @@ namespace System.Linq
             return ExecuteAsync<TSource, bool>(_any, source, cancellationToken);
         }
 
-        private static readonly MethodInfo _anyPredicate = GetMethod("Any", 1);
+        private static readonly MethodInfo _anyPredicate = GetMethod(nameof(Queryable.Any), 1);
 
         /// <summary>
         ///     Asynchronously determines whether any element of a sequence satisfies a condition.
@@ -87,7 +86,7 @@ namespace System.Linq
             return ExecuteAsync<TSource, bool>(_anyPredicate, source, predicate, cancellationToken);
         }
 
-        private static readonly MethodInfo _allPredicate = GetMethod("All", 1);
+        private static readonly MethodInfo _allPredicate = GetMethod(nameof(Queryable.All), 1);
 
         /// <summary>
         ///     Asynchronously determines whether all the elements of a sequence satisfy a condition.
@@ -126,7 +125,7 @@ namespace System.Linq
 
         #region Count/LongCount
 
-        private static readonly MethodInfo _count = GetMethod("Count");
+        private static readonly MethodInfo _count = GetMethod(nameof(Queryable.Count));
 
         /// <summary>
         ///     Asynchronously returns the number of elements in a sequence.
@@ -157,7 +156,7 @@ namespace System.Linq
             return ExecuteAsync<TSource, int>(_count, source, cancellationToken);
         }
 
-        private static readonly MethodInfo _countPredicate = GetMethod("Count", 1);
+        private static readonly MethodInfo _countPredicate = GetMethod(nameof(Queryable.Count), 1);
 
         /// <summary>
         ///     Asynchronously returns the number of elements in a sequence that satisfy a condition.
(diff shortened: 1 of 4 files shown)


maumar:
@anpete @divega

src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2185,22 +2185,41 @@ public static class EntityFrameworkQueryableExtensions
...
+            = typeof(EntityFrameworkQueryableExtensions)
+                .GetTypeInfo().GetDeclaredMethods("ThenInclude")
+                .Single(mi => mi.GetParameters().Any(pi => pi.Name == "navigationPropertyPath"));
+
 
         // TODO API docs once we resolve #1709
         public static IIncludableQueryable<TEntity, TProperty> ThenInclude<TEntity, TPreviousProperty, TProperty>(
             [NotNull] this IIncludableQueryable<TEntity, ICollection<TPreviousProperty>> source,
-            [NotNull] Expression<Func<TPreviousProperty, TProperty>> navigationPropertyPath)
+            [NotNull] Expression<Func<TPreviousProperty, TProperty>> collectionNavigationPropertyPath)
```
divega: The difference between the two overloads is in the source query argument, not the navprop path argument. Changing the name us very misleading.

src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2185,22 +2185,41 @@ public static class EntityFrameworkQueryableExtensions
                         new[] { source.Expression, Expression.Quote(navigationPropertyPath) })));
         }
 
-        internal static readonly MethodInfo ThenIncludeMethodInfo
+        internal static readonly MethodInfo ThenIncludeCollectionMethodInfo
```
divega: Something like `ThenIncludeAfterCollectionMethodInfo` would probably make what this is clearer.

src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2185,22 +2185,41 @@ public static class EntityFrameworkQueryableExtensions
...
             = typeof(EntityFrameworkQueryableExtensions)
                 .GetTypeInfo().GetDeclaredMethods("ThenInclude")
```
divega: OT: Can we start using nameof() for these?

maumar:
@divega - updated the PR with your comments

src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2185,23 +2184,41 @@ public static class EntityFrameworkQueryableExtensions
...
+
+        internal static readonly MethodInfo ThenIncludeAfterReferenceMethodInfo
+            = typeof(EntityFrameworkQueryableExtensions)
+                .GetTypeInfo().GetDeclaredMethods(nameof(EntityFrameworkQueryableExtensions.ThenInclude))
+                .Single(mi => mi.GetParameters().Any(pi => pi.Name == "sourceReference"));
 
         // TODO API docs once we resolve #1709
         public static IIncludableQueryable<TEntity, TProperty> ThenInclude<TEntity, TPreviousProperty, TProperty>(
-            [NotNull] this IIncludableQueryable<TEntity, ICollection<TPreviousProperty>> source,
+            [NotNull] this IIncludableQueryable<TEntity, ICollection<TPreviousProperty>> sourceCollection,
```
divega: Renaming the source parameter should be fine, but it is hard to come up with a good name. E.g. `sourceQueryIncludingCollectionNavigation` would be more accurate :stuck_out_tongue_winking_eye:. Another option would be to leave the parameter names the same and select the MethodInfos using type information, e.g. something like this: 

``` C#
internal static readonly MethodInfo ThenIncludeAfterCollectionMethodInfo =
    typeof(EntityFrameworkQueryableExtensions)
        .GetTypeInfo().GetDeclaredMethods(nameof(EntityFrameworkQueryableExtensions.ThenInclude))
        .Single(mi => mi.GetParameters()[0].ParameterType.GenericTypeArguments[1].IsGenericType);

internal static readonly MethodInfo ThenIncludeAfterReferenceMethodInfo = 
    typeof(EntityFrameworkQueryableExtensions)
        .GetTypeInfo().GetDeclaredMethods(nameof(EntityFrameworkQueryableExtensions.ThenInclude))
        .Single(mi => mi.GetParameters()[0].ParameterType.GenericTypeArguments[1].IsGenericParameter);
```

src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
```diff
@@ -21,7 +20,7 @@ public static class EntityFrameworkQueryableExtensions
     {
         #region Any/All
 
-        private static readonly MethodInfo _any = GetMethod("Any");
+        private static readonly MethodInfo _any = GetMethod(nameof(Queryable.Any));
```
divega: Thanks for trying this. It ends up looking a bit more verbose than the string literal. Let's find out what @anpete prefers.

divega: BTW, I like very much the way they in [Queryable.cs](https://github.com/dotnet/corefx/blob/master/src/System.Linq.Queryable/src/System/Linq/Queryable.cs) they take advantage of the compiler on every call, even if it potentially is less performant. Let's leave that to chat later.

src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
maumar: I will try to differentiate based on type. Proposed name seems too long to me.

divega: Agreed :smile: Thanks!

divega:
Other than the feedback provided so far, looks :shipit: to me.

src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs
anpete: Like nameof. Find doing the Queryable thing too - anything that reduces boilerplate/complexity here is great.

anpete:
:shipit:

PR #1885, opened 2015-03-23
API review: general renames and removals on builders

ajcvickers:
- Remove APIs that accept strings for type names
- Move from Many/One to Collection/Reference
- Move from Referenced to Principal
- Other miscellaneous changes


anpete:
:shipit:

PR #1886, opened 2015-03-23
Model building API changes

ajcvickers:
- Expose internal builders using IAccessor interface
- Rename internal and provider-specific classes to align with core classes


AndriySvyryd:
:shipit:

PR #1887, opened 2015-03-23
Sanitize SQL generated by SqlServerSequenceValueGenerator

mikary:
Using the ```ISqlServerSqlGenerator``` to sanitize the sequence name.
@ajcvickers @bricelam

### DIFF
--- a/src/EntityFramework.SqlServer/SqlServerSequenceValueGenerator.cs
+++ b/src/EntityFramework.SqlServer/SqlServerSequenceValueGenerator.cs
@@ -13,29 +13,33 @@ namespace Microsoft.Data.Entity.SqlServer
     public class SqlServerSequenceValueGenerator<TValue> : HiLoValueGenerator<TValue>
     {
         private readonly SqlStatementExecutor _executor;
+        private readonly ISqlServerSqlGenerator _sqlGenerator;
         private readonly ISqlServerConnection _connection;
         private readonly string _sequenceName;
 
         public SqlServerSequenceValueGenerator(
             [NotNull] SqlStatementExecutor executor,
+            [NotNull] ISqlServerSqlGenerator sqlGenerator,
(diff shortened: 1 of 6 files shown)


bricelam:
:shipit:

src/EntityFramework.SqlServer/SqlServerSequenceValueGenerator.cs
```diff
@@ -13,28 +13,30 @@ namespace Microsoft.Data.Entity.SqlServer
...
             _executor = executor;
+            _sqlGenerator = generator;
             _connection = connection;
         }
 
         protected override long GetNewLowValue()
         {
-            // TODO: Parameterize query and/or delimit identifier without using SqlServerMigrationOperationSqlGenerator
-            var sql = string.Format(CultureInfo.InvariantCulture, "SELECT NEXT VALUE FOR {0}", _sequenceName);
+            var sql = "SELECT NEXT VALUE FOR " + _sqlGenerator.DelimitIdentifier(_sequenceName);
```
anpete: Can this be added to the SQL generator so that it can be tested?

anpete:
Doesn't look like there is a test for this change?

src/EntityFramework.SqlServer/SqlServerSequenceValueGenerator.cs
mikary: Are you thinking something like ```AppendSelectValue(StringBuilder)```?

anpete: "Generate" is probably enough here, but yes, just so you can write a test for the change.

mikary:
Updated based on PR feedback, I added the method to the SqlGenerator, but the name I came up with, ``` GenerateNextSequenceValueOperation``` seems a bit too verbose to me.

PR #1891, opened 2015-03-24
Throw when System.Transactions.Transaction.Current is set

### LINKED ISSUE #1584: Throw when System.Transactions.Transaction.Current is set
When discussing https://github.com/aspnet/EntityFramework/issues/1530 we decided not to support TransactionScope. We decided we should detect if someone is trying to though and throw so it is clear it's not supported. We should have a flag to disable throwing in case folks are using TransactionScope for other reasons and EF code just happens to be hit in the block.

AndriySvyryd:
Fixes #1584

### DIFF
--- a/src/EntityFramework.Relational/RelationalConnection.cs
+++ b/src/EntityFramework.Relational/RelationalConnection.cs
@@ -21,6 +21,9 @@ namespace Microsoft.Data.Entity.Relational
         private int _openedCount;
         private int? _commandTimeout;
         private readonly LazyRef<ILogger> _logger;
+#if NET451
+        private bool _throwOnAmbientTransaction;
+#endif
 
         protected RelationalConnection([NotNull] IDbContextOptions options, [NotNull] ILoggerFactory loggerFactory)
         {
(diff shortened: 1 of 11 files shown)


AndriySvyryd:
@ajcvickers, @divega, @rowanmiller

src/EntityFramework.SqlServer/Extensions/SqlServerDbContextOptions.cs
```diff
@@ -39,6 +39,18 @@ public virtual SqlServerDbContextOptionsBuilder CommandTimeout(int? commandTimeo
             return this;
         }
 
+        public virtual SqlServerDbContextOptionsBuilder SuppressAmbientTransactionWarning()
+        {
+            var extension = new SqlServerOptionsExtension(OptionsBuilder.Options.FindExtension<SqlServerOptionsExtension>());
+            Debug.Assert(extension != null);
```
ajcvickers: Consider refactoring into a helper method and moving the assertion, which I think should be asserting that FindExtension is not returning null.

src/EntityFramework.Relational/RelationalConnection.cs
```diff
@@ -153,19 +160,47 @@ public virtual RelationalTransaction UseTransaction([CanBeNull] DbTransaction tr
...
         }
 
         public virtual async Task OpenAsync(CancellationToken cancellationToken = default(CancellationToken))
         {
+#if NET451
+            if(_throwOnAmbientTransaction
+               && System.Transactions.Transaction.Current != null)
+            {
+                throw new InvalidOperationException(Strings.AmbientTransaction);
+            }
```
ajcvickers: Consider DRY into a helper method.

ajcvickers:
:shipit:

PR #1892, opened 2015-03-24
Remove model builder interfaces and basic model builder extensions.

ajcvickers:
This means that extension methods are now written directly against the ModelBuilder classes, instead of being written in a way that can target any model builder.

Basic model builder extensions are now gone, since the basic model builder is going away anyway. Migrations no longer tries to use these extensiobns--it just writes out annotations.

Also added some new "hiding" methods to a couple of the builders to avoid breaking out of the generics.


AndriySvyryd:
:shipit:

PR #1894, opened 2015-03-24
API review change: disallow calling generic ForeignKey/PrincipalKey when...

ajcvickers:
... you didn't use an expression to get the builder

Also added a couple of new extension methods to avoid breaking out of typed sub-builders.


ajcvickers:
Will re-submit with some changes/fixes

PR #1895, opened 2015-03-24
API review change: Add generic builds for reference-reference and proper...

ajcvickers:
...ties

This changes adds generic builders for ReferenceReference relationships and properties. In the former case, this doesn't really add any functionality, but it can be done since we decided to disallow calling the generic methods unless you came from a lambda.

In the latter case, this allows extension methods to be written agains t certain property types.


AndriySvyryd:
:shipit:

ajcvickers:
Merged

PR #1896, opened 2015-03-24
EntityType properties ordered by primary key members first

mikary:
First commit removes ```LazyRef``` from the ```EntityType``` collection members as we discussed.
Second commit uses new ```IComparer<T>``` to order properties and adds tests

This solution makes the logic for ```SetPrimaryKey``` a bit strange because changes to the PK value impact the comparer behavior and the ```SortedDictionary``` assumes the contents are sorted when performing other operations (i.e. Remove).

The ```CompiledModel``` tests are broken by this change due to the property order in KoolModel.cs no longer being correct. Is this test model generated, or will the corrections need to be made by hand?

@ajcvickers @divega

### DIFF
--- a/src/EntityFramework.Core/Metadata/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/EntityType.cs
@@ -19,24 +19,20 @@ namespace Microsoft.Data.Entity.Metadata
     {
         private static readonly char[] _simpleNameChars = { '.', '+' };
 
-        private readonly LazyRef<SortedDictionary<IReadOnlyList<Property>, ForeignKey>> _foreignKeys
-            = new LazyRef<SortedDictionary<IReadOnlyList<Property>, ForeignKey>>(()
-                => new SortedDictionary<IReadOnlyList<Property>, ForeignKey>(PropertyListComparer.Instance));
+        private readonly SortedDictionary<IReadOnlyList<Property>, ForeignKey> _foreignKeys
+            = new SortedDictionary<IReadOnlyList<Property>, ForeignKey>(PropertyListComparer.Instance);
 
-        private readonly LazyRef<SortedDictionary<string, Navigation>> _navigations
-            = new LazyRef<SortedDictionary<string, Navigation>>(() =>
-                new SortedDictionary<string, Navigation>(StringComparer.Ordinal));
+        private readonly SortedDictionary<string, Navigation> _navigations
+            = new SortedDictionary<string, Navigation>(StringComparer.Ordinal);
 
-        private readonly LazyRef<SortedDictionary<IReadOnlyList<Property>, Index>> _indexes
-            = new LazyRef<SortedDictionary<IReadOnlyList<Property>, Index>>(() =>
-                new SortedDictionary<IReadOnlyList<Property>, Index>(PropertyListComparer.Instance));
+        private readonly SortedDictionary<IReadOnlyList<Property>, Index> _indexes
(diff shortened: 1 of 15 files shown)


AndriySvyryd:
The `CompiledModel` tests need to be fixed manually.

anpete:
I think we should consider moving the compiled model prototype to a separate futures branch.

ajcvickers:
I think removing the CompiledModel code from the main code base it an appropriate thing to do.

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -1049,5 +1066,42 @@ public int Compare(IReadOnlyList<Property> x, IReadOnlyList<Property> y)
...
+                    }
+                    else
+                    if (xIndex != -1)
+                    {
+                        return -1;
+                    }
+                    else if (yIndex != -1)
+                    {
+                        return 1;
+                    }
```
ajcvickers: Consider conditional operator.

test/EntityFramework.Core.Tests/DbContextTest.cs
```diff
@@ -2336,6 +2336,13 @@ public void Add_Attach_Remove_Update_do_not_call_DetectChanges()
             }
         }
 
+        private void AssertEqual(IReadOnlyList<string> expectedNames, IEnumerable<string> actualNames)
+        {
+            Assert.Equal(
+                new SortedSet<string>(expectedNames, StringComparer.Ordinal),
+                new SortedSet<string>(actualNames, StringComparer.Ordinal));
+        }
+
```
ajcvickers: Why not just assert the expected order rather than re-sorting?

test/EntityFramework.Core.Tests/ModelBuilderTest.cs
```diff
@@ -6643,16 +6643,25 @@ private class SelfRef
             public int SelfRefId { get; set; }
         }
 
+        private void AssertEqual(IReadOnlyList<string> expectedNames, IEnumerable<string> actualNames)
```
ajcvickers: Same comment again.

test/EntityFramework.Core.Tests/DbContextTest.cs
mikary: Is the actual order relevant to the behavior that the test is intended to exercise?
Enforcing order on tests outside of the metadata seems redundant to me.

ajcvickers: The order should be fixed in metadata and I would rather have less code in the tests coupled with additional coverage that the order is what it is supposed to be.

AndriySvyryd: Since here it's only 3 places I think the coupling is fine, but in the ModelBuilderTest a change in the ordering would require changes to hundreds of lines.

mikary: to be fair, I'm already fixing hundreds of tests in ```SqlServer.FunctionalTests``` to reorder columns in queries. As long as we don't make any further changes to the property order it's not that bad, I only made this change because it felt more future-proof.

ajcvickers: I'd rather keep the tests with less code, but I don't mind too much one way or the other.

ajcvickers:
:shipit:

PR #1898, opened 2015-03-25
Fix to 1742 - sometimes we produce invalid parameter name.

maumar:
Problem is that for some (quite complicated) cases we would produce invalid parameter name. We try to generate the parameter name based ProperyInfo/FieldInfo path of the MemberExpression. Most of the time this is correct, but sometimes compiler can generate names like:
((QueryBugsTest+<>c__DisplayClass21_0).CS$<>8__locals1).details).LastName, which would produce name with invalid characters (e.g. CS$<>8__locals1_details_LastName)

Fix is to remove the compiler generated prefix.

### DIFF
--- a/test/EntityFramework.SqlServer.FunctionalTests/QueryBugsTest.cs
+++ b/test/EntityFramework.SqlServer.FunctionalTests/QueryBugsTest.cs
@@ -352,6 +352,47 @@ Queen of the Andals and the Rhoynar and the First Men, Khaleesi of the Great Gra
             }
         }
 
+        [Fact]
+        public void Compiler_generated_local_closure_produces_valid_parameter_name_1742()
+        {
+            Execute1742(new CustomerDetails_1742 { FirstName = "Foo", LastName = "Bar" });
+        }
+
+        public void Execute1742(CustomerDetails_1742 details)
+        {
+            var loggingFactory = new TestSqlLoggerFactory();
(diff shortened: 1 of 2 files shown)


maumar:
@divega

test/EntityFramework.SqlServer.FunctionalTests/QueryBugsTest.cs
```diff
@@ -352,6 +352,44 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+            using (var ctx = new MyContext925(serviceProvider))
+            {
+                var firstName = details.FirstName;
+
+                var query = ctx.Customers.Where(c => c.FirstName == firstName && c.LastName == details.LastName).ToList();
+
+                var expectedSql =
+@"SELECT [c].[FirstName], [c].[LastName]
+FROM [Customer] AS [c]
+WHERE ([c].[FirstName] = @__firstName_0) AND ([c].[LastName] = @__8__locals1_details_LastName_1)";
```
divega: I talked to @maumar and while this seems correct we believe we can do better on aspect of making this human readable (which was the goal of using names from the compiler in the first place). We will file a bug for that, but we can check this improvement that makes the name a valid parameter name beta 4.

divega:
:shipit:

PR #1902, opened 2015-03-25
Remove Compiled Model prototype

mikary:
Removing the compiled model code so we don't need to maintain it (for #1896)
@divega @ajcvickers @AndriySvyryd


anpete:
:shipit:

PR #1920, opened 2015-03-27
React to hosting

HaoK:
https://github.com/aspnet/Hosting/pull/210

cc @divega

### DIFF
--- a/src/EntityFramework.Commands/ContextTool.cs
+++ b/src/EntityFramework.Commands/ContextTool.cs
@@ -34,8 +34,7 @@ namespace Microsoft.Data.Entity.Commands
 #if DNX451 || DNXCORE50
             try
             {
-                var engine = WebApplication.CreateHostingEngine(CallContextServiceLocator.Locator.ServiceProvider, config: null, configureServices: null);
-                return engine.ApplicationServices.GetService(type) as DbContext;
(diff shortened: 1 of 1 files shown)


divega:
:shipit:

bricelam:
Cool; much easier. :shipit:

HaoK:
Updated to use new overload, now even cleaner, one line with no arguments :)

src/EntityFramework.Commands/ContextTool.cs
```diff
@@ -37,12 +34,7 @@ private static DbContext TryCreateContextFromStartup(Type type)
 #if DNX451 || DNXCORE50
             try
             {
-                var context = new HostingContext
-                {
-                    ServerFactory = new ServerFactory(),
-                };
-                var instance = new HostingEngine().Start(context);
-                return context.ApplicationServices.GetService(type) as DbContext;
+                return WebApplication.CreateHostingEngine().ApplicationServices.GetService(type) as DbContext;
```
davidfowl: Holy one liner :smile:

divega: Not trying to be a wet blanket, but can you please split it? :smile:

divega: Not trying to be a wet blanket, but can you please split it? :smile:

divega:
:shipit:

PR #1925, opened 2015-03-27
disabling tests that randomly fail

maumar:
Will investigate and fix, but disabling for now so that build passes


maumar:
@mikary

mikary:
:shipit:

PR #1930, opened 2015-03-27
Add a way to accept all changes after multiple successful SaveChanges.

### LINKED ISSUE #744: Integrate updating entity state and the committing of generated values into entities with user-controlled transactions
Values generated in the update pipeline are saved in sidecars so that if the save fails these values can be discarded. Currently the values are always committed from sidecars into the actual entity instances when SaveChanges ends. But if the user is handling transactions manually then this commit should not happen until the transaction is committed.

Likewise, entity state is updated at the end of SaveChanges, but again this should only be done when the transaction is actually committed--this is the AcceptChanges functionality in EF6.

AndriySvyryd:
Fixes #744

### DIFF
--- a/src/EntityFramework.Core/DbContext.cs
+++ b/src/EntityFramework.Core/DbContext.cs
@@ -290,13 +290,34 @@ namespace Microsoft.Data.Entity
         [DebuggerStepThrough]
         public virtual int SaveChanges()
         {
+            return SaveChanges(acceptAllChangesOnSuccess: true);
+        }
+
+        /// <summary>
+        ///     Saves all changes made in this context to the underlying data store.
+        /// </summary>
+        /// <param name="acceptAllChangesOnSuccess">
+        ///     Indicates whether <see cref="ChangeTracker.AcceptAllChanges" /> is called after the changes have been
+        ///     sent succesfully to the data store.
+        /// </param>
+        /// <remarks>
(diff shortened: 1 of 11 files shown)


src/EntityFramework.Core/DbContext.cs
```diff
@@ -290,13 +290,34 @@ protected internal virtual void OnModelCreating(ModelBuilder modelBuilder)
...
+        /// <remarks>
+        ///     This method will automatically call <see cref="ChangeTracker.DetectChanges" /> to discover any changes
+        ///     to entity instances before saving to the underlying data store. This can be disabled via
+        ///     <see cref="ChangeTracker.AutoDetectChangesEnabled" />.
+        /// </remarks>
+        /// <returns>
+        ///     The number of state entries written to the underlying data store.
+        /// </returns>
+        [DebuggerStepThrough]
+        public virtual int SaveChanges(bool acceptAllChangesOnSuccess)
```
ajcvickers: Default value rather than new overload? Not sure which is best...

ajcvickers: Never mind. CancellationToken on the async version makes the default value approach less nice.

ajcvickers:
:shipit:

src/EntityFramework.Core/DbContext.cs
AndriySvyryd: Default value means that we will never change it.

Issue #1927, opened 2015-03-27
Query :: data corruption for queries with filter and && operator where one side is client eval and another is server eval


maumar:
For this case the client eval gets ignored, resulting in incorrect data returned to the customer.

Repro:

```
namespace Repro1927
{
    using System;
    using System.Data.SqlClient;
    using System.Linq;
    using Microsoft.Data.Entity;

    public class MyContext : DbContext
    {
        public DbSet<MyEntity> MyEntities { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "Repro1927", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }
    }

    public class MyEntity
    {
        public int Id { get; set; }
        public int Foo { get; set; }
        public string Bar { get; set; }
    }

    class Program
    {
        public static bool MyCompare(int argument)
        {
            return argument == 1;
        }

        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var e1 = new MyEntity { Foo = 1, Bar = "Bar" };
                var e2 = new MyEntity { Foo = 2, Bar = "Bar" };

                ctx.MyEntities.AddRange(e1, e2);
                ctx.SaveChanges();
            }

            using (var ctx = new MyContext())
            {
                var query = ctx.MyEntities.Where(e => MyCompare(e.Foo) && e.Bar == "Bar").ToList();

                // expected: 1, actual: 2
                Console.WriteLine(query.Count);
            }
        }
    }
}
```

maumar:
This has been fixed in the release branch for beta4 - 5d0b2d43f018d1fd7d5c5b217145e883b34a201c

maumar:
merged into dev in 5d0b2d43f018d1fd7d5c5b217145e883b34a201c and 4236e393d8a2c199b866331ce8379d371df79946

maumar:
Keeping open to track the improved fix work (mentioned in the pull request - https://github.com/aspnet/EntityFramework/pull/1941

maumar:
added more tests in ce3117f180738ea4f4aba5acb6505c19ba1a31c3, closing

PR #1941, opened 2015-03-27
Fix to #1927 - Query :: data corruption for queries with filter and && operator where one side is client eval and another is server eval

maumar:
Problem is that for predicates with && operator, for which one side was client eval and the other was server eval we would only process the server eval part. This could lead to data corruption (i.e. query returning more rows than intended)

Fix is to handle top level && operation differently - splitting predicate into two Where operations, one on client and one on server. This is done only for && operation and only for top level. If client eval is needed at any point and the top level expression is not AndAlso, the entire predicate will be evaluated on the client.


maumar:
@anpete @divega

anpete:
Discussed and are looking at making this visitor consistent with how our other visitors work. That is, it will  no longer return the SQL expression but a potentially rewritten input expression, used for client eval. The SQL expression will be available as a property on the visitor.

anpete:
:shipit:

PR #1942, opened 2015-03-27
Add DefaultSequence to the model

### LINKED ISSUE #1568: Add default sequence to the model when used
When I configure a model to use the default sequence, the default sequence should be added to the model. Here is an example of using the default sequence.

```C#
modelBuilder.Entity<Person>().Property(p => p.Id).ForSqlServer().UseSequence();
```

smitpatel:
Resolves #1568 
Also changed few missed nameof

### DIFF
--- a/src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerValueGenerationStrategyConvention.cs
+++ b/src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerValueGenerationStrategyConvention.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Data.Entity.Metadata.Internal;
 using Microsoft.Data.Entity.Metadata.ModelConventions;
+using Microsoft.Data.Entity.Relational.Metadata;
 
 namespace Microsoft.Data.Entity.SqlServer.Metadata.ModelConventions
 {
@@ -13,6 +14,23 @@ namespace Microsoft.Data.Entity.SqlServer.Metadata.ModelConventions
             modelBuilder.Annotation(
                 SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.ValueGeneration,
                 SqlServerValueGenerationStrategy.Sequence.ToString(),
+                ConfigurationSource.Convention);
+
+            var sequence = new Sequence(Sequence.DefaultName) { Model = modelBuilder.Metadata };
+            modelBuilder.Annotation(
+                SqlServerAnnotationNames.Prefix + RelationalAnnotationNames.Sequence + sequence.Schema + "." + sequence.Name,
+                sequence.Serialize(),
+                ConfigurationSource.Convention
+                );
+
+            modelBuilder.Annotation(
+                SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.DefaultSequenceName,
+                sequence.Name,
+                ConfigurationSource.Convention);
+
+            modelBuilder.Annotation(
+                SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.DefaultSequenceSchema,
+                sequence.Schema,
                 ConfigurationSource.Convention);
 
             return modelBuilder;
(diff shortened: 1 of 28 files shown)


src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerValueGenerationStrategyConvention.cs
```diff
@@ -70,10 +70,14 @@ public virtual InternalRelationshipBuilder Apply(InternalRelationshipBuilder rel
...
-                SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.ValueGeneration,
-                SqlServerValueGenerationStrategy.Sequence.ToString(),
-                ConfigurationSource.Convention);
+            var extensions = modelBuilder.Metadata.SqlServer();
+
+            var sequence = extensions.GetOrAddSequence();
+
+            extensions.ValueGenerationStrategy = SqlServerValueGenerationStrategy.Sequence;
+            extensions.DefaultSequenceName = sequence.Name;
+            extensions.DefaultSequenceSchema = sequence.Schema;
```
ajcvickers: Doesn't this mean that these are no longer marked as being set by convention?

ajcvickers: Also, the sequence should not be added to the model if it is not needed by the model. This is important for supporting SQL Server 2008 where creating the sequence will fail and so applications should be able to switch 100% to Identity and have no remnants of sequence functionality remaining.

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerValueGenerationStrategyConvention.cs
```diff
@@ -74,6 +74,12 @@ public virtual InternalModelBuilder Apply(InternalModelBuilder modelBuilder)
                 SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.ValueGeneration,
                 SqlServerValueGenerationStrategy.Sequence.ToString(),
                 ConfigurationSource.Convention);
+
+            var extensions = modelBuilder.Metadata.SqlServer();
+            var sequence = extensions.GetOrAddSequence();
```
bricelam: The sequence annotation should be marked as "by convention" and removed if the user calls `.UseIdentity()`. We should also file a bug to remove it if it never ends up being used.

smitpatel:
filed #1994

src/EntityFramework.SqlServer/Metadata/ModelConventions/SqlServerValueGenerationStrategyConvention.cs
```diff
@@ -15,6 +15,11 @@ public virtual InternalModelBuilder Apply(InternalModelBuilder modelBuilder)
                 SqlServerValueGenerationStrategy.Sequence.ToString(),
                 ConfigurationSource.Convention);
 
+            var extensions = modelBuilder.Metadata.SqlServer();
+            var sequence = extensions.GetOrAddSequence();
+            extensions.DefaultSequenceName = sequence.Name;
+            extensions.DefaultSequenceSchema = sequence.Schema;
```
bricelam: Should these somehow be passed `ConfigurationSource.Convention`?

src/EntityFramework.SqlServer/Metadata/SqlServerModelBuilder.cs
```diff
@@ -20,7 +20,13 @@ public SqlServerModelBuilder([NotNull] Model model)
 
         public virtual SqlServerModelBuilder UseIdentity()
         {
-            _model.SqlServer().ValueGenerationStrategy = SqlServerValueGenerationStrategy.Identity;
+            var extensions = _model.SqlServer();
+
+            extensions.RemoveSequence();
```
bricelam: We should only do this if it was added by convention.

smitpatel: at Model level it is not possible to know configuration source of the annotation. So we cannot verify if this was added by convention or not.

bricelam:
:shipit:

PR #1943, opened 2015-03-28
Change UseStoreDefault and StoreComputed to StoreGeneratedPattern

ajcvickers:
Also, mark all non-composite, not dependent, integer primary key properties as Identity by default.

This is the basic functionality of these properties. Stuff still to do includes:
- Reuse data annotation enum
- Integration with read-only semantics
- Updates to SQL Server value generation semantics to make use of new core information
- Throwing of appropriate exceptions on negative cases by making use of core information


src/EntityFramework.Core/Metadata/Internal/InternalIndexBuilder.cs
```diff
@@ -14,7 +14,7 @@ public InternalIndexBuilder([NotNull] Index index, [NotNull] InternalModelBuilde
         {
         }
 
-        public virtual bool IsUnique(bool isUnique, ConfigurationSource configurationSource)
+        public virtual bool IsUnique(bool? isUnique, ConfigurationSource configurationSource)
```
bricelam: Should the method body also be updated to remove the calls to `.HasValue` and `.Value`?

bricelam:
:shipit:

src/EntityFramework.Core/Metadata/Internal/InternalIndexBuilder.cs
ajcvickers: Will do.

PR #1944, opened 2015-03-28
enabling disabled query tests

maumar:
Using logic from SqlServerTestStore.GetOrCreateShared to prevent race conditions during database creation.


maumar:
@ajcvickers

ajcvickers:
:shipit:

maumar:
merged

PR #1945, opened 2015-03-28
Use core Identity concept and convention for SQL Server keys

ajcvickers:
Core model building now has a convention to set StoreGeneratedPattern.Identity for non-composite primary keys that are not themselves a dependent property of some other key. This information can now be used by SQL Server instead of it having its own convention to do exactly the same thing.


bricelam:
:shipit:

ajcvickers:
Merged

Issue #1890, opened 2015-03-24
Inheritance EntityType AddProperty does not check base type


mikary:
The ```AddProperty``` method in ```EntityType``` ensures that names are not duplicated for properties in the current type, but does not ensure the name is unique per hierarchy. This could be a problem if an ```EntityType``` has a CLR or shadow property with a name that collides with a shadow property on the base type.

AndriySvyryd:
This isn't a problem if we decide that we support property hiding. The property on the most derived type would be used in that case.

rowanmiller:
We should enforce this for the moment and then remove the restriction when we properly support hiding.

PR #1955, opened 2015-03-30
EntityType AddProperty checking inherited property names

mikary:
Additional test coverage and corrections/improvements to the exceptions for duplicate properties for issue #1890
@rowanmiller @anpete


anpete:
:shipit:

PR #1959, opened 2015-03-31
Use ChangeType instead of cast when reading

roji:
PostgreSQL has no byte type (tinyint), the accepted usage is to store bytes as smallint (2-byte). However, the resulting (boxed) short can't be downcast to byte. Modified to use Convert.ChangeType() instead.

I hope I'm not missing some other mechanism for custom conversion/casting of values returned from the database.

### DIFF
--- a/src/EntityFramework.Relational/RelationalObjectArrayValueReader.cs
+++ b/src/EntityFramework.Relational/RelationalObjectArrayValueReader.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Data.Entity.Relational
         {
             Debug.Assert(index >= 0 && index < Count);
 
-            return (T)_values[index];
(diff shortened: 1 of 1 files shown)


msotclas:
Hi __@roji__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

roji:
Never mind, I discovered RelationalValueReaderFactory and created a custom version of RelationalObjectArrayValueReader which does the necessary checks and conversions. Must faster than Convert.ChangeType()

PR #1962, opened 2015-03-31
Fix ThenInclude method location for Mono

rowanmiller:
IsConstructedGenericType is returning different results on Mono so
adjusting logic to something that works.
New logic for the collection overload just uses a "not" of the logic to find the reference overload.
With this change the InMemory store works on Mono 3.12.1 (at least for basic scenarios).


divega:
:shipit:

rowanmiller:
Scrapping this one, will open another PR with other changes needed for Mono too.

PR #1974, opened 2015-04-01
Workarounds to support Mono

rowanmiller:
Two workarounds needed...

**ThenInclude method location**
IsConstructedGenericType is returning different results on Mono so adjusting logic to something that works.
New logic for the collection overload just uses a "not" of the logic to find the reference overload.

**Remove use of DbDataReader.GetFieldValue<T>**
There is a bug in Mono where `GetFieldValue<T>` is overridden in `SqlDataReader` to throw even though there is now a base implementation that would work.
I've opened an issue against Mono to fix this - https://bugzilla.xamarin.com/show_bug.cgi?id=28662
**This is just a temporary workaround that will need to be done in a better way since we want the generic to help with type detection on SQLite and possibly other providers.**

### DIFF
--- a/src/EntityFramework.Relational/RelationalTypedValueReader.cs
+++ b/src/EntityFramework.Relational/RelationalTypedValueReader.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Data.Entity.Relational
 
         public virtual bool IsNull(int index) => _dataReader.IsDBNull(index);
 
-        public virtual T ReadValue<T>(int index) => _dataReader.GetFieldValue<T>(index);
+        public virtual T ReadValue<T>(int index) => (T)_dataReader.GetValue(index);
 
         public virtual int Count => _dataReader.FieldCount;
     }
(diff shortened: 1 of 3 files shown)


src/EntityFramework.Relational/RelationalTypedValueReader.cs
```diff
@@ -22,7 +22,7 @@ public RelationalTypedValueReader([NotNull] DbDataReader dataReader)
 
         public virtual bool IsNull(int index) => _dataReader.IsDBNull(index);
 
-        public virtual T ReadValue<T>(int index) => _dataReader.GetFieldValue<T>(index);
+        public virtual T ReadValue<T>(int index) => (T)_dataReader.GetValue(index);
```
divega: If you haven't already, can you please file a bug for us to consider reverting this one in the future (e.g. as soon as Mono is fixed)?

divega:
:shipit:

rowanmiller:
@roji can you confirm that the change to ReadValue<T> won't break npgsql work. We know that some datastores (such as SQLite) will want us to call GetFieldValue<T> so that they can use the T to work out the type that we are expecting. I don't think this is the case for you? This is just a temporary change to unblock Mono for the next release.

roji:
@rowanmiller nope, it shouldn't break anything. This actually seems better than casting: in Npgsql GetFieldValue<T> is implemented without any boxing going on (unlike GetValue()). So if you guys can keep it this way it means less useless heap allocation.

PR #1988, opened 2015-04-03
Extend read-only semantics per store generation discussion

ajcvickers:
Note: the naming of the concepts and whether or not they should be in an enum can be discussed in API review.

This change adds the concept of being read-only before saving a new entity as well as the concept of being read-only when updating a new entity. It also moves the check to happen as part of SaveChanges so that properties can go through a temporary state of being modified. Key values cannot go through this temporary state since the state manager relies on their values not changing.

Also added functional tests for all combinations of read-only flags and store generated patterns.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
@@ -144,12 +144,10 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
             // set all properties to modified if the entity state is explicitly set to Modified.
             if (newState == EntityState.Modified)
             {
-                _stateData.FlagAllProperties(EntityType.GetProperties().Count(), isFlagged: true);
-
-                foreach (var keyProperty in EntityType.GetProperties().Where(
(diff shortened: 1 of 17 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -225,9 +223,9 @@ public virtual void SetPropertyModified([NotNull] IProperty property, bool isMod
                 return;
             }
 
-            if (isModified && property.IsReadOnly)
+            if (isModified && property.IsKey())
```
bricelam: Hmm... I think I'm with @anpete on this--the sooner it throws, the better. It'll be hard to track down the line of code that modified it if we wait until `SaveChanges` to throw.

bricelam:
:shipit:

src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
ajcvickers: I'l bring it to a design meeting or small group discussion.

ajcvickers:
Merged

PR #1993, opened 2015-04-06
[TINY] Adding tests for All operator with predicate being partially client eval and partially server eval

maumar:
currently we push everything to the client


maumar:
@anpete

anpete:
:shipit:

PR #1995, opened 2015-04-06
API review changes: make Navigations first class on ForeignKey

ajcvickers:
And make PointsToPrincipal an extension method.


bricelam:
:shipit:

PR #1996, opened 2015-04-06
Make ChangeTracker not resolved directly from DI

ajcvickers:
ChangeTracker is a top-level API class and makes use of API patterns rather than just being a service contract. It is therefore best not resolved directly from DI.


bricelam:
:shipit:

PR #2001, opened 2015-04-07
Add contract interfaces for remaining DI Services

ajcvickers:
Mostly In-memory/SQL Server/Relational services.

Also remove protected test double constructors per API review decision.


bricelam:
:shipit:

PR #2002, opened 2015-04-07
Only add subquery in FromSql when composing on query result.

mikary:
Updates to the query annotations and SQL generation that only turns FromSql query into a subquery when necessary.
@anpete @divega


src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -85,7 +86,8 @@ protected override Expression VisitEntityQueryable([NotNull] Type elementType)
                 ? tableName.First().ToString().ToLower()
                 : _querySource.ItemName;
 
-            var fromSqlAnnotation = GetAnnotations<FromSqlAnnotation>(_querySource).SingleOrDefault();
+            var annotation = GetAnnotations<FromSqlAnnotation>(_querySource).SingleOrDefault();
+            var fromSqlAnnotation = (annotation?.ResultOperator as AnnotateQueryResultOperator)?.Expression.Value as FromSqlAnnotation;
```
anpete: I'm guessing this is addressed in the next PR? :-)

src/EntityFramework.Relational/Query/CommandBuilder.cs
```diff
@@ -1,28 +1,24 @@
...
     public class CommandBuilder
     {
-        private readonly SelectExpression _selectExpression;
-        private readonly RelationalQueryCompilationContext _relationalQueryCompilationContext;
+        private readonly Func<ISqlQueryGenerator> _sqlGeneratorFunc;
 
-        public CommandBuilder(
-            [NotNull] SelectExpression selectExpression,
-            [NotNull] RelationalQueryCompilationContext relationalQueryCompilationContext)
+        public CommandBuilder([NotNull] Func<ISqlQueryGenerator> sqlGeneratorFunc)
```
anpete: Why is this a func?

src/EntityFramework.Relational/Query/Sql/RawSqlQueryGenerator.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.Data.Entity.Relational.Query.Sql
+{
+    public class RawSqlQueryGenerator : ISqlQueryGenerator
+    {
+        private readonly Dictionary<string, object> _parameters;
+        private readonly string _rawSql;
+        private readonly object[] _rawParameters;
+
+        public RawSqlQueryGenerator([NotNull] string rawSql, [NotNull] object[] rawParameters)
```
anpete: Just call these args sql and parameters.

src/EntityFramework.Relational/Query/CommandBuilder.cs
mikary: I was preserving the behavior (new generator instance each time a command is built) but it doesn't need to be. If we drop the func we can also look at caching in the generator like one of the comments below mentions.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
mikary: yep, this is the part that inspired me to make the next PR.

src/EntityFramework.Relational/Query/Sql/ISqlQueryGenerator.cs
```diff
@@ -9,10 +9,8 @@ namespace Microsoft.Data.Entity.Relational.Query.Sql
 {
     public interface ISqlQueryGenerator
     {
-        string GenerateSql(
-            [NotNull] SelectExpression selectExpression,
-            [NotNull] IDictionary<string, object> parameterValues);
+        string GenerateSql([NotNull] IDictionary<string, object> parameterValues);
 
-        IEnumerable<string> Parameters { get; }
+        IDictionary<string, object> Parameters { get; }
```
anpete: We have a CommandParameter class we could use here.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -19,25 +19,32 @@ namespace Microsoft.Data.Entity.Relational.Query.Sql
     public class DefaultSqlQueryGenerator : ThrowingExpressionTreeVisitor, ISqlExpressionVisitor, ISqlQueryGenerator
     {
         private IndentedStringBuilder _sql;
-        private List<string> _parameters;
+        private Dictionary<string, object> _parameters;
```
anpete: This seems strange. Can we unify here?

src/EntityFramework.Relational/Query/CommandBuilder.cs
anpete: Yeah, fine to drop the func.

src/EntityFramework.Relational/Query/Sql/ISqlQueryGenerator.cs
anpete: Funnily enough it is not actually used anywhere right now.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
mikary: There's a bit of an input parameters vs output parameters thing going on here.

The generator adds to the dictionary in some cases, and in others it uses the value as a constant in the generated SQL and ignores the value in the dictionary. Rather than mutate the provided dictionary or maintain the list of the used parameters, I felt it would be cleaner to make a dictionary that only contains the key-value-pairs that need to be made into ```Paraeter```s

mikary:
Updated and ready for another look

anpete:
:shipit:

Issue #1858, opened 2015-03-19
Filter missing if wrong type passed in


ztone:
When I accidentally passed in an int as a filter to a column that is a smallint/short then the SQL generated became skewed without any kind of warning.

Also tried <code>Disdatefilter  >=</code> and just using <code>Equals()</code> like in this example.

Here is the model/context
```c#
public class Context : DbContext
{
    public DbSet<Data> Data { get; set; }

    protected override void OnConfiguring(DbContextOptions options)
    {
        options.UseSqlServer(@"..");
    }

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        base.OnModelCreating(modelBuilder);
        modelBuilder.Entity<Data>().Key(sd => new { sd.IdType, sd.Id });
    }
}

public class Data
{
        public short Provider { get; set; }
        public short DisdateFilter { get; set; }
        public long IdType{ get; set; }
        public byte Id { get; set; }
        public bool Excluded{ get; set; }
}
```
The linq query used
```c#
int start = 1403;
using (var ctx = new Context ())
{
   var query = ctx.Data.Where(x => x.DisdateFilter.Equals(start) && x.Provider.Equals(612))
      .Select(x => new { x.IdType });
}
```

Here is the skewed SQL that it produced.  Two things I notice here.  One is of course that the DisdateFilter is missing from the where clause.  But also that the SQL statement is selecting the filtered columns.
```sql
exec sp_executesql N'SELECT [x].[DisdateFilter], [x].[Provider], [x].[IdType]
FROM [Data] AS [x]
WHERE [x].[Provider] = @p0',N'@p0 smallint',@p0=612
```
When using the correct type the SQL is correct.
```sql
exec sp_executesql N'SELECT [x].[IdType]
FROM [Data] AS [x]
WHERE ([x].[DisdateFilter] = @p0 AND [x].[Provider] = @p1)',N'@p0 smallint,@p1 smallint',@p0=1403,@p1=612
```

maumar:
Investigation results: The problem in the way parameter is evaluated. For Equals method with not-matching types Equals(object, object) is being used. Parameter is wrapped in a convert and when we evaluate it, it comes out as object. Later on when we encounter parameter of type object we don't know how to translate it to SQL (in general), so this part of the predicate is done on the client.

maumar:
Filed https://github.com/aspnet/EntityFramework/issues/1927 - related to this, it should be fixed in the next beta.

[...]

PR #2003, opened 2015-04-07
Fix to #1858 - Filter missing if wrong type passed in

maumar:
Issue is for queries that use Equals(object) method for comparison. C# will return false whenever objects of different types are compared using this method, even if the underlying values are identical after conversion to one of the types.
If arguments are nullable (e.g. int?), nullability is ignored for the purpose of the comparison. E.g. 'int?' and 'int' may result in positive comparison, int? and int? as well, int? and long? will always return false.

Fix is to mimic this behavior, as well as log warning that use of Equals(object) on two objects of different types was probably not intentional.
Another fix is in the way we evaluate parameters. Before we if parameter was wrapped in the Convert node, we would evaluate it's type to the type it was converted to. In case of Equals(object), and probably other cases as well, this causes creation of parameters typed as object, which always get evaluated on the client (since we don't know how to translate parameter/constant of type 'object'.

Fix is to evaluate parameters to it's original type, and wrap the parameter into Convert node afterward. This way we preserver the intended typing but given parameters can be evaluated on the server.

### DIFF
--- a/src/EntityFramework.Core/Query/CompiledQueryCache.cs
+++ b/src/EntityFramework.Core/Query/CompiledQueryCache.cs
@@ -210,8 +210,9 @@ namespace Microsoft.Data.Entity.Query
                 {
                     var unaryExpression = (UnaryExpression)expression;
 
-                    if (unaryExpression.Type.IsNullableType()
+                    if ((unaryExpression.Type.IsNullableType() 
                         && !unaryExpression.Operand.Type.IsNullableType())
+                        || unaryExpression.Type == typeof(object))
                     {
                         e = unaryExpression.Operand;
                     }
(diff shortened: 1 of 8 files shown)


maumar:
@anpete @divega

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -213,8 +213,8 @@ public override Expression VisitExpression(Expression expression)
                 {
                     var unaryExpression = (UnaryExpression)expression;
 
-                    if (unaryExpression.Type.IsNullableType()
-                        && !unaryExpression.Operand.Type.IsNullableType())
+                    if ((unaryExpression.Type.IsNullableType() && !unaryExpression.Operand.Type.IsNullableType())
```
maumar: will break line here

maumar:
ping

divega:
Looks :shipit: to me. Best to wait for @anpete though.

anpete:
:shipit:

PR #2005, opened 2015-04-08
Use named resources for PCL compat

victorhurdugaci:
The names of the embedded resources emitted by the compiler changed.

For EF, instead of emitting
`EntityFramework.Commands.Strings.resource`, it emits `EntityFramework.Commands.Properties.Strings`. Using named resources, we can tell it to emit the expected name.

In response to:  https://github.com/aspnet/dnx/pull/1615

cc @davidfowl

### DIFF
--- a/src/EntityFramework.Commands/project.json
+++ b/src/EntityFramework.Commands/project.json
@@ -13,6 +13,9 @@
     },
     "compile": "..\\Shared\\*.cs",
     "exclude": "tools\\Handlers.cs",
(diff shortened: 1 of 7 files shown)


victorhurdugaci:
PS: Travis and AppVeyor will fail. It is expected because it cannot build without the DNX change

victorhurdugaci:
Ping @bricelam @anpete @divega

bricelam:
:shipit:

Issue #1904, opened 2015-03-25
Build one model for all data stores


ajcvickers:
We decided to change back to building a single model for all data stores. This has several implications:
- Data store specific conventions should, as a best practice, not change the structure of the model, but should only add annotations. If the structure of the model is changed, then this will result in a different structure for all data stores. For example, if a new column is added by one data store, it will show up in all other data stores as well.
- The way annotations work doesn't need to change and #1542 can be closed. 
- The way model builders and conventions work needs to be changed. This is the major work item being tracked by this bug. By default, only the conventions for the currently selected data store will run. However, it should be possible to change this in OnModelCreating to also run conventions for any other available data store.

bricelam:
The SqlServerModelDiffer will also need to be updated to handle provider-specific overrides.

bricelam:
In fact, the whole `ModelDiffer` architecture may need to change to allow multiple providers to play part in the diffing process. The implementation can be deferred until after RTM, but we should think about it up front to prevent/minimize breaking changes to the Migrations provider model.

bricelam:
We also need to review places in `EntityFramework.Relational` where we call `entityType.Relational()` (or similar) to account for provider-specific overrides.

rowanmiller:
We'll enable this when we work on multi-provider migrations etc.

PR #2006, opened 2015-04-08
Bring back SQLite (Resolves #1953)

### LINKED ISSUE #1953: Bring back SQLite

bricelam:
It's pretty crippled at the moment. Things like `.Count()` don't work because the provider returns a boxed `long` that we try to cast to an `int`. There also seems to be something wrong with the `LIKE` operator (possibly an encoding issue). Migrations isn't implemented (but `.EnsureCreated()` should more-or-less work). ...just to name a few.

### DIFF
--- a/src/EntityFramework.Relational/RelationalDataStoreCreator.cs
+++ b/src/EntityFramework.Relational/RelationalDataStoreCreator.cs
@@ -12,23 +12,54 @@ namespace Microsoft.Data.Entity.Relational
     {
         public abstract bool Exists();
 
-        public abstract Task<bool> ExistsAsync(CancellationToken cancellationToken = default(CancellationToken));
+        public virtual Task<bool> ExistsAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            return Task.FromResult(Exists());
+        }
 
         public abstract void Create();
 
-        public abstract Task CreateAsync(CancellationToken cancellationToken = default(CancellationToken));
+        public virtual Task CreateAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            Create();
+
+            return Task.FromResult(0);
+        }
 
         public abstract void Delete();
 
-        public abstract Task DeleteAsync(CancellationToken cancellationToken = default(CancellationToken));
+        public virtual Task DeleteAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            Delete();
+
+            return Task.FromResult(0);
+        }
 
         public abstract void CreateTables(IModel model);
 
-        public abstract Task CreateTablesAsync(IModel model, CancellationToken cancellationToken = default(CancellationToken));
+        public virtual Task CreateTablesAsync(IModel model, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            CreateTables(model);
+
+            return Task.FromResult(0);
+        }
 
         public abstract bool HasTables();
 
-        public abstract Task<bool> HasTablesAsync(CancellationToken cancellationToken = default(CancellationToken));
+        public virtual Task<bool> HasTablesAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+
+            return Task.FromResult(HasTables());
+        }
 
         public virtual bool EnsureDeleted(IModel model)
         {
--- a/src/EntityFramework.Relational/RelationalDatabase.cs
+++ b/src/EntityFramework.Relational/RelationalDatabase.cs
@@ -13,11 +13,11 @@ using Microsoft.Framework.Logging;
 
 namespace Microsoft.Data.Entity.Relational
 {
-    public abstract class RelationalDatabase : Database, IAccessor<IMigrator>
+    public class RelationalDatabase : Database, IAccessor<IMigrator>
     {
         private readonly IMigrator _migrator;
 
-        protected RelationalDatabase(
+        public RelationalDatabase(
             [NotNull] DbContext context,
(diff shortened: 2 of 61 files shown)


src/EntityFramework.Relational/RelationalDataStoreCreator.cs
```diff
@@ -12,23 +12,40 @@ public abstract class RelationalDataStoreCreator : IRelationalDataStoreCreator
...
 
         public abstract void Create();
 
-        public abstract Task CreateAsync(CancellationToken cancellationToken = default(CancellationToken));
+        public virtual Task CreateAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            Create();
+
+            return Task.FromResult(0);
+        }
```
ajcvickers: Are we okay with a blocking default implementation of these async methods? /cc @divega

src/EntityFramework.Relational/RelationalDatabase.cs
```diff
@@ -13,11 +13,11 @@
 
 namespace Microsoft.Data.Entity.Relational
 {
-    public abstract class RelationalDatabase : Database, IAccessor<IMigrator>
+    public class RelationalDatabase : Database, IAccessor<IMigrator>
```
ajcvickers: Do you think we should remove the SQL Server dervied class as well? I think we should be consistent.

src/EntityFramework.Relational/Update/CommandBatchPreparer.cs
```diff
@@ -77,9 +77,19 @@ protected virtual IEnumerable<ModificationCommand> CreateModificationCommands([N
                     .AddEntry(e));
         }
 
-        public abstract IRelationalPropertyExtensions GetPropertyExtensions(IProperty property);
+        public virtual IRelationalPropertyExtensions GetPropertyExtensions(IProperty property)
+        {
+            Check.NotNull(property, nameof(property));
+
+            return property.Relational();
```
ajcvickers: This seems like it will always result in using the relational annotations and not any provider-specific overrides, so it seems like this is never the correct implementation for any provider and hence should not be a default implemention in this base class.

src/EntityFramework.Sqlite/Extensions/SqliteDbContextOptionsBuilderExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Relational;
+using Microsoft.Data.Entity.Sqlite;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity
+{
+    public static class SqliteDbContextOptionsBuilderExtensions
+    {
+        public static RelationalDbContextOptionsBuilder UseSqlite([NotNull] this DbContextOptionsBuilder options, [NotNull] string connectionString)
```
ajcvickers: Assuming this is returning RelationalDbContextOptionsBuilder  because there aren't any SQLite specific options yet?

src/EntityFramework.Sqlite/SqliteDataStoreCreator.cs
```diff
@@ -0,0 +1,101 @@
...
+            try
+            {
+                using (var conneciton = _connection.CreateReadOnlyConnection())
+                {
+                    conneciton.Open();
+                }
+
+                exists = true;
+            }
+            catch (SqliteException ex) when (ex.SqliteErrorCode == SQLITE_CANTOPEN)
```
ajcvickers: C# 6 Ninja!

ajcvickers:
:shipit:

src/EntityFramework.Relational/RelationalDataStoreCreator.cs
divega: Seems consistent with the default implementation of the async methods in relational, i.e. ADO.NET. I would add cancellationToken.ThrowIfCancellationRequested() on each of them though.

bricelam: Will throw when cancelled.

src/EntityFramework.Relational/RelationalDatabase.cs
bricelam: My goal here was to provide a minimal implementation. We'd then democratize the work to flesh it out. The provider-specific RelationalDatabase classeses did seem a bit YAGNI...

src/EntityFramework.Relational/Update/CommandBatchPreparer.cs
bricelam: Query had a default implementation like this. I'm fine forcing the providers to override, but we should be consistent.

src/EntityFramework.Sqlite/Extensions/SqliteDbContextOptionsBuilderExtensions.cs
bricelam: Correct

src/EntityFramework.Relational/RelationalDataStoreCreator.cs
ajcvickers: Sounds good.

src/EntityFramework.Relational/RelationalDatabase.cs
ajcvickers: I'm fine with removing them.

src/EntityFramework.Relational/Update/CommandBatchPreparer.cs
ajcvickers: I think query should change--a default implementation that is never the correct implementation seems like a pit of failure. Happy to for you to fix it here and then file a bug on query.

src/EntityFramework.Relational/RelationalDatabase.cs
bricelam: Filed #2043

src/EntityFramework.Relational/Update/CommandBatchPreparer.cs
bricelam: Reverted and added a comment to review query (and anywhere else) in #1904

PR #2011, opened 2015-04-11
Revert hack to value readers

ajcvickers:
This change makes value readers be resolved from DI everywhere (instead of just for query) such that different data stores can use different value reader implementations. SQL Server is set to use the object array value reader, which never calls the generic methods that are not supported on Mono, and which is currently the only value reader supported by query. This allows the hack to the typed value reader to be reverted, and sets things up so that other providers can use the typed reader as intended. Note that query still needs to be updated to work with the typed value reader before it can be used across the stack.


bricelam:
W00t! :shipit:

ajcvickers:
Merged

PR #2019, opened 2015-04-15
Consolidate query annotations and decouple from result operator types

mikary:
Consolidated the ```ResultOperator```s for ```FromSql```, ```AsNoTracking```, and ```Include``` to simplify annotation extraction and to decouple Annotations from the ```ResultOperator``` type. Turned ```QueryAnnotation``` into an abstract base type to simplify filtering and casting annotations.
@anpete


anpete:
Nice refactoring! :shipit:

PR #2020, opened 2015-04-15
React to WebHostingBuilder

HaoK:
https://github.com/aspnet/Hosting/pull/237

cc @divega @davidfowl Any way to avoid CallContextServiceLocator here?

### DIFF
--- a/src/EntityFramework.Commands/ContextTool.cs
+++ b/src/EntityFramework.Commands/ContextTool.cs
@@ -12,6 +12,7 @@ using Microsoft.Data.Entity.Utilities;
 #if DNX451 || DNXCORE50
 using System.Threading.Tasks;
 using Microsoft.AspNet.Hosting;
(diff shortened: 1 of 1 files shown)


bricelam:
I can flow in the `IServiceProvider` from the `Program` class. File a bug and I'll fix.

bricelam:
:shipit: Can service building be decoupled from hosting? It feels like the `Startup.ConfigureServices` pattern of building services is orthogonal from the rest of the hosting pipeline.

HaoK:
Great, filed https://github.com/aspnet/EntityFramework/issues/2024

HaoK:
re splitting service building, I'm not sure we can, since Startup potentially can take things like IHostingEnvironment in its ctor which only come from hosting.

HaoK:
99ebd0cf3f8755bb9a25e63aa464388f374eb71e

Issue #1838, opened 2015-03-17
Query :: implement CLR null semantics for query predicates


maumar:
In c# null == null returns true, in database, it returns null. We can make some transformations to the queries to compensate for this discrepancy.

for cases where distinction between null and False is irrelevant:
a == b => (a == b || (a is null && b is null))

for cases where distinction between null and False is relevant:
a != b => (a != b) || !(a is null && b is null) || (a is null && !(b is null)) || (!(a is null) && b is null)

maumar:
Fixed in 4d053cd37f8b6c9083a462bcf0bf417df7c97b9f

PR #2022, opened 2015-04-15
Fix to #1838 - Query :: implement CLR null semantics for query predicates

maumar:
Problem is that database semantics for null equality are different from c#, resulting in unexpected results for some queries (involving null value comparisons)

Fix is to compensate this by introducing additional terms in the predicate, effectively converting 3-valued logic into 2-valued logic.

a == b -> [(a == b) && (a != null && b != null)] || (a == null && b == null))
a != b -> [(a != b) || (a == null || b == null)] && (a != null || b != null)]

also added transformations for cases when one of the operands is negated

!a == b -> [(a != b) && (a != null && b != null)] || (a == null && b == null)
!a != b -> [(a == b) || (a == null || b == null)] && (a != null || b != null)

Some terms can be removed if one of the terms is not nullable, e.g:

nullableA == nonNullableB -> (nullableA == nonNullableB) && (nullableA != null)

We do additional optimization for simple predicate of a == b where a and b are nullable:
a == b -> (a == b) || (a == null && b == null)

This is still 3-valued result, true being returned when 'a' and 'b' are in c# terms, and either false or NULL when they are not equal.
This optimization can only be applied if the predicate does not contain any NotEqual or Not operations.

### DIFF
--- a/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/CompositeRelationalExpressionTreeVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/CompositeRelationalExpressionTreeVisitor.cs
@@ -16,28 +16,61 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
             var inExpressionOptimized = 
                 new EqualityPredicateInExpressionOptimizer().VisitExpression(currentExpression);
 
-            if (inExpressionOptimized != null)
-            {
-                currentExpression = inExpressionOptimized;
-            }
+            currentExpression = inExpressionOptimized;
 
-            var equalityExpanded = 
+            var negationOptimized1 =
+                new PredicateNegationExpressionOptimizer()
+                .VisitExpression(currentExpression);
+
+            currentExpression = negationOptimized1;
+
+            var equalityExpanded =
                 new EqualityPredicateExpandingVisitor().VisitExpression(currentExpression);
 
-            if (equalityExpanded != null)
+            currentExpression = equalityExpanded;
+
+            var negationOptimized2 =
+                new PredicateNegationExpressionOptimizer()
+                .VisitExpression(currentExpression);
+
+            currentExpression = negationOptimized2;
+
+            var optimizedNullExpansionVisitor = new NullSemanticsOptimizedExpandingVisitor();
+            var nullSemanticsExpandedOptimized = optimizedNullExpansionVisitor.VisitExpression(currentExpression);
+            if (optimizedNullExpansionVisitor.OptimizedExpansionPossible)
             {
-                currentExpression = equalityExpanded;
+                currentExpression = nullSemanticsExpandedOptimized;
+            }
+            else
+            {
+                currentExpression = new NullSemanticsExpandingVisitor()
+                    .VisitExpression(currentExpression);
             }
 
-            var negationOptimized =
-                new PredicateNegationExpressionOptimizer().VisitExpression(currentExpression);
+            var negationOptimized3 =
+                new PredicateNegationExpressionOptimizer()
+                .VisitExpression(currentExpression);
 
-            if (negationOptimized != null)
+            currentExpression = negationOptimized3;
+
+            var reducedExpression = new ReducingExpressionVisitor()
+                .VisitExpression(currentExpression);
+
+            return reducedExpression;
+        }
+
+        private class ReducingExpressionVisitor : ExpressionTreeVisitor
+        {
+            public override Expression VisitExpression(Expression node)
             {
-                currentExpression = negationOptimized;
-            }
+                if (node != null && node.CanReduce)
+                {
+                    var reduced = node.Reduce();
+                    return base.VisitExpression(reduced);
+                }
 
-            return currentExpression;
+                return base.VisitExpression(node);
+            }
         }
     }
 }
--- a/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/EqualityPredicateExpandingVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/EqualityPredicateExpandingVisitor.cs
@@ -14,6 +14,9 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
         protected override Expression VisitBinaryExpression(
             [NotNull]BinaryExpression expression)
         {
+            var left = VisitExpression(expression.Left);
+            var right = VisitExpression(expression.Right);
+
             if ((expression.NodeType == ExpressionType.Equal
                 || expression.NodeType == ExpressionType.NotEqual)
                 && expression.Left.Type == typeof(bool)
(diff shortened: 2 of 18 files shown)


maumar:
@anpete @divega

src/EntityFramework.Relational/Query/Expressions/NotNullableExpression.cs
```diff
@@ -1,21 +1,21 @@
...
+using JetBrains.Annotations;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Parsing;
+using Microsoft.Data.Entity.Relational.Query.Sql;
+using System;
 
 namespace Microsoft.Data.Entity.Relational.Query.Expressions
 {
-    public class IsNotNullExpression : ExtensionExpression
+    public class NotNullableExpression : ExtensionExpression
```
maumar: this is not really a rename

test/EntityFramework.SqlServer.FunctionalTests/NullSemanticsQuerySqlServerTest.cs
```diff
@@ -19,7 +19,540 @@ public override void Compare_bool_with_bool_equal()
...
+    [e].[BoolA] = [e].[BoolB]) THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT) END <> CASE WHEN (
+    [e].[IntA] = [e].[IntB]) THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT) END
+
+SELECT [e].[Id]
+FROM [NullSemanticsEntity1] AS [e]
+WHERE CASE WHEN (
+    [e].[NullableBoolA] = [e].[BoolB]) THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT) END <> CASE WHEN (
+    [e].[IntA] = [e].[NullableIntB]) THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT) END
+
+SELECT [e].[Id]
```
maumar: for comparison, that 4th query in EF6 would be translated to:
SELECT
    [Extent1].[Id] AS [Id]
    FROM [dbo].[NullSemanticsEntity1] AS [Extent1]
    WHERE  NOT (((CASE WHEN (([Extent1].[NullableBoolA] = [Extent1].[NullableBoo
lB]) OR (([Extent1].[NullableBoolA] IS NULL) AND ([Extent1].[NullableBoolB] IS N
ULL))) THEN cast(1 as bit) WHEN ( NOT (([Extent1].[NullableBoolA] = [Extent1].[N
ullableBoolB]) AND ((CASE WHEN ([Extent1].[NullableBoolA] IS NULL) THEN cast(1 a
s bit) ELSE cast(0 as bit) END) = (CASE WHEN ([Extent1].[NullableBoolB] IS NULL)
 THEN cast(1 as bit) ELSE cast(0 as bit) END)))) THEN cast(0 as bit) END) = (CAS
E WHEN (([Extent1].[NullableIntA] = [Extent1].[NullableIntB]) OR (([Extent1].[Nu
llableIntA] IS NULL) AND ([Extent1].[NullableIntB] IS NULL))) THEN cast(1 as bit
) WHEN ( NOT (([Extent1].[NullableIntA] = [Extent1].[NullableIntB]) AND ((CASE W
HEN ([Extent1].[NullableIntA] IS NULL) THEN cast(1 as bit) ELSE cast(0 as bit) E
ND) = (CASE WHEN ([Extent1].[NullableIntB] IS NULL) THEN cast(1 as bit) ELSE cas
t(0 as bit) END)))) THEN cast(0 as bit) END)) AND ((CASE WHEN (CASE WHEN (([Exte
nt1].[NullableBoolA] = [Extent1].[NullableBoolB]) OR (([Extent1].[NullableBoolA]
 IS NULL) AND ([Extent1].[NullableBoolB] IS NULL))) THEN cast(1 as bit) WHEN ( N
OT (([Extent1].[NullableBoolA] = [Extent1].[NullableBoolB]) AND ((CASE WHEN ([Ex
tent1].[NullableBoolA] IS NULL) THEN cast(1 as bit) ELSE cast(0 as bit) END) = (
CASE WHEN ([Extent1].[NullableBoolB] IS NULL) THEN cast(1 as bit) ELSE cast(0 as
 bit) END)))) THEN cast(0 as bit) END IS NULL) THEN cast(1 as bit) ELSE cast(0 a
s bit) END) = (CASE WHEN (CASE WHEN (([Extent1].[NullableIntA] = [Extent1].[Null
ableIntB]) OR (([Extent1].[NullableIntA] IS NULL) AND ([Extent1].[NullableIntB]
IS NULL))) THEN cast(1 as bit) WHEN ( NOT (([Extent1].[NullableIntA] = [Extent1]
.[NullableIntB]) AND ((CASE WHEN ([Extent1].[NullableIntA] IS NULL) THEN cast(1
as bit) ELSE cast(0 as bit) END) = (CASE WHEN ([Extent1].[NullableIntB] IS NULL)
 THEN cast(1 as bit) ELSE cast(0 as bit) END)))) THEN cast(0 as bit) END IS NULL
) THEN cast(1 as bit) ELSE cast(0 as bit) END)))

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -1864,7 +1845,11 @@ public override void Where_bool_member_and_parameter_compared_to_binary_expressi
 
 SELECT [p].[ProductID], [p].[Discontinued], [p].[ProductName], [p].[UnitsInStock]
 FROM [Products] AS [p]
-WHERE ([p].[Discontinued] = 1 AND ((([p].[ProductID] > 50) AND @__prm_0 = 0) OR (([p].[ProductID] <= 50) AND @__prm_0 = 1))) OR ([p].[Discontinued] = 0 AND (([p].[ProductID] <= 50) OR @__prm_0 = 1) AND (([p].[ProductID] > 50) OR @__prm_0 = 0))",
+WHERE CASE WHEN (
```
maumar: I will add heuristic to use the (a && b) || (!a && !b) instead of case, when null semantics expansion is not needed in the children - it looks bit cleaner and is more sargable. WIll do this in separate checkin, per conversation with @divega

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/EqualityPredicateExpandingVisitor.cs
```diff
@@ -27,29 +30,27 @@ public class EqualityPredicateExpandingVisitor : ExpressionTreeVisitor
...
-                                Expression.Not(expression.Left),
-                                Expression.Not(right)))
-                        : Expression.OrElse(
-                            Expression.AndAlso(
-                                left,
-                                Expression.Not(right)),
-                            Expression.AndAlso(
-                                Expression.Not(left),
-                                right));
+                {
```
maumar: I will add heuristic to use the (a && b) || (!a && !b) instead of case, when null semantics expansion is not needed in the children - it looks bit cleaner and is more sargable. WIll do this in separate checkin, per conversation with @divega

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/CompositeRelationalExpressionTreeVisitor.cs
```diff
@@ -16,28 +16,61 @@ public class CompositePredicateExpressionTreeVisitor : ExpressionTreeVisitor
...
 
-            var negationOptimized =
-                new PredicateNegationExpressionOptimizer().VisitExpression(currentExpression);
+            var negationOptimized3 =
+                new PredicateNegationExpressionOptimizer()
+                .VisitExpression(currentExpression);
+
+            currentExpression = negationOptimized3;
+
+            var reducedExpression = new ReducingExpressionVisitor()
```
anpete: Can we inherit ExpressionVisitor and call base here instead?

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/EqualityPredicateExpandingVisitor.cs
anpete: More sargable, LOL!

anpete:
Really great job! :shipit:

test/EntityFramework.SqlServer.FunctionalTests/NullSemanticsQuerySqlServerTest.cs
divega: Awesome improvement!

Issue #2026, opened 2015-04-16
Query :: consider different expansion for complex equality predicates (e.g. (a == b) == c) for some cases


maumar:
Currently complex comparisons like:

(a == b) == (c) are rewritten into case statements:

(CASE WHEN a == b then 1 ELSE 0) == (CASE WHEN c == 1 then 1 ELSE 0)

Currently we expand expressions to CASE when we encounter complex expresion on either side of the equal/notequal. However we only need to expand the complex operand and leave the other one as it was.

maumar:
@anpete

maumar:
Fixed in 32aeae2c2a4a62308b0a20f0073f4ed7f11d63e2

PR #2028, opened 2015-04-17
Fix to #2026 - consider different expansion for complex equality predicates (e.g. (a == b) == c) for some cases.

maumar:
This is a performance/sqlgen improvement.
Currently expressions like (a == b) == (c) are rewritten into case statements:
(CASE WHEN a == b then 1 ELSE 0) == (CASE WHEN c == 1 then 1 ELSE 0)

This is not necessary as we only need to expand the part that is complex (a==b) and we can leave the simple part (c) intact.


anpete:
:shipit:

Issue #1861, opened 2015-03-19
Query :: contains -> IN transformation should handle nulls correctly


maumar:
This is related to null semantics work (https://github.com/aspnet/EntityFramework/issues/1838)

We should handle cases where user passes a collection that contains null. We should produce something like:

SELECT c.Id 
FROM Customers as c
WHERE c.Name IN ('Foo', 'Bar')) OR c.Name is NULL

maumar:
Fixed in 80d325e90dccd6bfbff667f1d22e83cd68b5da0f

PR #2029, opened 2015-04-18
[SMALL] Fix to #1861 - contains -> IN transformation should handle nulls correctly

maumar:
Queries translated to IN expression were not handling nulls correctly, e.g.:

from c in customers
where c.Name == "Foo" || c.Name == "Bar" || c.Name == null
select c.Id

would translate to

SELECT c.Id
FROM Customers as c
WHERE c.Name IN ('Foo', 'Bar', NULL)

This would not return customers with Name == null, because it would apply database null semantics for this comparison (null == null -> null)

Fix is to add additional term that will handle the case.

Above query will get translated to:

SELECT c.Id
FROM Customers as c
WHERE c.Name IN ('Foo', 'Bar') || c.Name IS NULL


maumar:
@anpete

anpete:
:shipit:

PR #2035, opened 2015-04-20
Allow generic GetFieldValue to be used in query

ajcvickers:
This change allows a value reader to be code genned that eagerly reads all values passing required type information to the ADO.NET provider. This requires that the types needed be captured/calculated in query. The code gennned delegates are cached and reused across the stack, not just for query.

Some additional cleanup/optimization and testing still to come. Primarily tested by running all the tests using this value reader on SQL Server, even though it is not currently used by SQL Server by default.

One test (Take_with_single_select_many) is failing with the new reader, will investigate more and/or file a bug.


src/EntityFramework.Relational/TypedValueReaderFactory.cs
```diff
@@ -0,0 +1,66 @@
...
+                    typeof(object),
+                    valueTypes.Select((t, i) => CreateGetValueExpression(
+                        dataReaderParam,
+                        Expression.Add(offsetParam, Expression.Constant(i)),
+                        t))),
+                dataReaderParam,
+                offsetParam).Compile();
+        }
+
+        private Expression CreateGetValueExpression(Expression dataReaderExpression, Expression indexExpression, Type type)
```
anpete: Duplication of code in EntityMaterializerSource?

src/EntityFramework.Relational/Query/RelationalQueryContext.cs
```diff
@@ -45,8 +47,15 @@ public virtual void RegisterDataReader([NotNull] DbDataReader dataReader)
...
 
-        public virtual IValueReader CreateValueReader(int readerIndex) 
-            => ValueReaderFactory.CreateValueReader(_activeDataReaders[_activeReaderOffset + readerIndex]);
+        public virtual IValueReader CreateValueReader([NotNull] IEntityType entityType, int readerIndex, int readerOffset)
+        {
+            Check.NotNull(entityType, nameof(entityType));
+
+            return ValueReaderFactory.CreateValueReader(
+                _activeDataReaders[_activeReaderOffset + readerIndex],
+                entityType.GetValueTypes(),
```
anpete: This looks like a perf issue. I don't think we can call GetValueTypes per value reader, we need to move it to query compile time.

src/EntityFramework.Relational/TypedValueReaderFactory.cs
ajcvickers: Its somewhat different, although there are similarities. This will likely change when we handle nulls better anyway.

src/EntityFramework.Relational/Query/RelationalQueryContext.cs
ajcvickers: I'll file a bug for the optimization.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -155,6 +156,9 @@ var keyFactory
                         });
             }
 
+            queryMethodArguments.Add(Expression.Constant((Func<IEnumerable<Type>>)(
+                () => (QueryModelVisitor.TryGetQuery(_querySource) ?? selectExpression).GetProjectionTypes())));
```
anpete: We should try and move this to type determination to query compile time.

ajcvickers: Will file a bug on it.

src/EntityFramework.Relational/Query/RelationalQueryContext.cs
anpete: I think a way to do this could be to create a ValueReaderFactoryFactory, move the types list and reader offset to that, and then bind the ValueReaderFactory into the query expression tree directly. That way, the AppDomain cache can potentially go away too.

bricelam:
I was playing with this, and queries like `db.Posts.Count()` were still failing on SQLite. I'll dig more into it...

bricelam:
I found the code that's breaking `.Count()`: [QueryMethodProvider.cs line 36](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Query/QueryMethodProvider.cs#L36). It reads directly form the `DbDataReader`.

anpete:
Nice catch. This just needs to call GetFieldValue<T>(), which we already do on the async path.

ajcvickers:
Sending out a new PR that moves the factory creation into query compilation time.

PR #2036, opened 2015-04-21
Target net45 and net45+win

### LINKED ISSUE #1965: Mono: Use of Scale/Precision in RelationalDecimalTypeMapping
Currently our code base compiles on Mono except for the use of DbParameter.Precision/Scale in RelationalDecimalTypeMapping.cs. Obviously compilation isn't a key scenario but it gives insight into what will work at runtime too.

Given this appears to be the only issue (at least in terms of API surface usage) we may want to adjust code.

```
            if (parameter.Direction == ParameterDirection.Output)
            {
                parameter.Scale = _scale;
                parameter.Precision = _precision;
            }
```

bricelam:
This may (or may not) enable Mono/Xamarin.

This resolves #1965

@AndriySvyryd Will the removal of `TransactionScopeAsyncFlowOption` be a problem? If so, we'll have to cross-compile `EntityFramework.Relational` to both `net45` and `net451`.

This depends on the following.
- aspnet/Configuration#180
- aspnet/DependencyInjection#223
- aspnet/Logging#174
- aspnet/Options#57

### DIFF
--- a/src/EntityFramework.Commands/EntityFramework.Commands.csproj
+++ b/src/EntityFramework.Commands/EntityFramework.Commands.csproj
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.Data.Entity.Commands</RootNamespace>
     <AssemblyName>EntityFramework.Commands</AssemblyName>
-    <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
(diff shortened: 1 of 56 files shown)


AndriySvyryd:
If `TransactionScopeAsyncFlowOption` is not used then the connection might sometimes be enrolled in the ambient transaction. On platforms where it's not available I'd always throw when an ambient transaction is detected in an async call instead of allowing the user to skip this check.

bricelam:
I talked to @divega and @rowanmiller, and we decided not to suppress.

Issue #2034, opened 2015-04-20
Query ::  incorrect (missing) parentheses in sql gen for some complex expressions


maumar:
We are trying to optimize out some unnecessary parentheses during sql generation. However current algorithm is too aggressive and some brackets that are needed are not being generated. 

Query:

```
            string prm1 = null;
            string prm2 = null;
            string prm3 = "Blah";

es => es.Where(e =>
                e.NullableStringB != null
                && e.NullableStringA != "Foo"
                && e.NullableStringA != prm1
                && e.NullableStringA != prm2
                && e.NullableStringA != prm3)
```

SQL:

```
__prm3_2: Blah

SELECT [e].[Id]
FROM [NullSemanticsEntity1] AS [e]
WHERE [e].[NullableStringB] IS NOT NULL AND ([e].[NullableStringA] <> 'Foo') 
OR [e].[NullableStringA] IS NULL 
AND [e].[NullableStringA] IS NOT NULL 
AND [e].[NullableStringA] IS NOT NULL 
AND ([e].[NullableStringA] <> @__prm3_2) 
OR [e].[NullableStringA] IS NULL

```

maumar:
Fixed in e172882e1e039b61fa980e4466baa049f8ac94b1

Issue #2033, opened 2015-04-20
Query :: invalid results for queries involving null-valued parameters


maumar:
We are incorrectly transforming queries like:

var prm = null;
var query = customers.Where(c => c.CustomerName != prm).Select(c => c.Id)

We would generate the following:

SELECT c.Id
FROM Customers AS c
WHERE c.CustomerName IS NOT NULL OR c.CustomerName IS NULL

which is incorrect (second term shouldn't be there)

Reason is that during null semantics expansion we don't know that prm is actually null, so we translate this to:

SELECT c.Id
FROM Customers AS c
WHERE c.CustomerName <> _prm OR c.CustomerName IS NULL

and later on we optimize out the comparison to null-valued parameter into IS NOT NULL. However it's too late. We must optimize out null-parameter comparisons before we perform null semantics expansion.

maumar:
Fixed in e172882e1e039b61fa980e4466baa049f8ac94b1

PR #2037, opened 2015-04-21
Fix to query bugs #2033 and #2034:

maumar:
#2033 - invalid results for queries involving null-valued parameters
#2034 - incorrect (missing) parentheses in sql gen for some complex expressions

2033:
We were incorrectly transforming queries that were making comparisons with null-valued parameters.
This was happening because we were performing null semantics expansion first and then getting rid of null-valued parameters comparisons.

Query:
customers.Where(c => c.CustomerName != prm).Select(c => c.Id)

Would get it's comparison expanded to:

SELECT c.Id
FROM Customers AS c
WHERE c.CustomerName <> _prm OR c.CustomerName IS NULL

Which would later be optimized into:
SELECT c.Id
FROM Customers AS c
WHERE c.CustomerName IS NOT NULL OR c.CustomerName IS NULL

which was wrong.

Fix is to perform null-valued parameter comparison pruning first, and only after that perform null semantics expansion.
This means that if we have parameters in the query, null semantics expansion can only be performed once we know the values of parameters (i.e. during sql generation).

If no parameters are used in the query, null semantics is performed early, just like before, so it's results can be cached.

2034:
Problem was that for some cases with multiple AndAlso and OrElse expressions, we would not produce enough parentheses.
Basically we were comparing the node type of the outer binary, with the node type of the inner binary. If they were the same, we would not produce a bracket.
However, if both inner and outer nodes are same (say AndAlso) but the inner node has OrElse inside, it may be required to wrap the inner AndAlso in a bracket (due to the inner OrElse)
This would not happen, and we would produce invalid results.

Fix is to produce parentheses around all binary expressions by default. Some optimizations are still performed - we don't create brackets for binary expressions between two simple terms (column, parameter, constant)
with the exception of logical binary expression for which we always produce parentheses

### DIFF
--- a/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
+++ b/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
@@ -114,7 +114,26 @@ namespace Microsoft.Data.Entity.Relational.Query.Sql
                 }
                 else
                 {
-                    VisitExpression(selectExpression.Predicate);
+                    var predicate = new NullComparisonTransformingVisitor(_parameterValues)
+                        .VisitExpression(selectExpression.Predicate);
+
+                    // we have to optimize out comparisons to null-valued parameters before we can expand null semantics 
+                    if (_parameterValues.Count > 0)
+                    {
+                        var optimizedNullExpansionVisitor = new NullSemanticsOptimizedExpandingVisitor();
+                        var nullSemanticsExpandedOptimized = optimizedNullExpansionVisitor.VisitExpression(predicate);
+                        if (optimizedNullExpansionVisitor.OptimizedExpansionPossible)
+                        {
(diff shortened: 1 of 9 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -563,16 +575,13 @@ protected override Expression VisitBinaryExpression([NotNull] BinaryExpression b
             }
             else
             {
-                var needClosingParen = false;
+                var needParentheses = !IsSimpleExpression(binaryExpression.Left)
```
maumar: would it be safe to skip brackets for all non-logical expressions?

anpete:
:shipit:

Issue #2016, opened 2015-04-14
Query: StateManager.StartTracking should track graph


anpete:
Related to #2015, the StartTracking API used by query should attach graphs.

Entities streaming from a query can be graphs because of Include. When tracking such entities, we want to track all of the entities in the graph. Previously, query would call StartTracking on any included entities, but this functionality was enabled by the QueryBuffer, which is going away.

anpete:
Superceded in https://github.com/aspnet/EntityFramework/pull/2092

PR #2038, opened 2015-04-21
Query: Foundational work on stateless query pipeline. The main change he...

anpete:
...re is to go from buffering value readers in the query buffer to having value readers move through a query as part of the result data flow.

QueryBuffer removal is pending #2016.

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionTreeVisitors/MemberAccessBindingExpressionTreeVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionTreeVisitors/MemberAccessBindingExpressionTreeVisitor.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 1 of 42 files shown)


src/EntityFramework.Core/Query/QueryBuffer.cs
```diff
@@ -135,7 +116,7 @@ public virtual void StartTracking(object entity)
             List<BufferedEntity> bufferedEntities;
             if (_byEntityInstance.TryGetValue(entity, out bufferedEntities))
             {
-                foreach (var bufferedEntity in bufferedEntities.Where(e => e != null))
+                foreach (var bufferedEntity in bufferedEntities.Skip(1).Where(e => e != null))
```
maumar: might be good to add comment here

anpete: This whole class is going away soon.

src/EntityFramework.Core/Query/ExpressionTreeVisitors/MemberAccessBindingExpressionTreeVisitor.cs
```diff
@@ -37,16 +108,44 @@ protected override Expression VisitMemberExpression(MemberExpression memberExpre
...
                 var member = memberExpression.Member;
                 var typeInfo = newExpression.Type.GetTypeInfo();
 
-                if (typeInfo.IsGenericType
-                    && typeInfo.GetGenericTypeDefinition() == typeof(IAsyncGrouping<,>))
+                if (typeInfo.IsGenericType)
                 {
-                    member = typeInfo.GetDeclaredProperty("Key");
+                    var genericTypeDefinition = typeInfo.GetGenericTypeDefinition();
+                    var asyncGrouping = genericTypeDefinition == typeof(IAsyncGrouping<,>);
```
maumar: this could be inlined

maumar:
:shipit:

PR #2041, opened 2015-04-21
Move ValueReaderFactory to query compilation context

ajcvickers:
First step to allowing the typed factory to be created at query compilation time.


anpete:
:shipit:

PR #2042, opened 2015-04-21
Introduce factory for value reader factories

ajcvickers:
This will allow type information to be used to create different dynamic factories as needed.


anpete:
:shipit:

ajcvickers:
Merged

PR #2050, opened 2015-04-22
Pretty Print Generic EntityType Names

### LINKED ISSUE #1986: Generic EntityTypes have different names across runtimes
Given the following model...
```C#
class Entity<TKey>
{
    public TKey Id { get; set; }
}

class Context : DbContext
{
    public DbSet<Entity<int>> Entities { get; set; }
}
```
...the name of the entity type will be different depending on the platform.

Runtime | Entity Type Name
------- | ----------------
dnx451 | Entity`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=**b77a5c561934e089**]]
dnxcore50 | Entity`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=**7cec85d7bea7798e**]]

Since these are *our* metadata names, we should prettify them to something like `Entity<string>`.

smitpatel:
Resolves #1986

### DIFF
--- a/src/EntityFramework.Core/Utilities/TypeNameHelper.cs
+++ b/src/EntityFramework.Core/Utilities/TypeNameHelper.cs
@@ -1,0 +1,128 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Utilities
+{
+    public static class TypeNameHelper
+    {
+        private static readonly Dictionary<Type, string> _builtInTypeNames = new Dictionary<Type, string>
+            {
+            { typeof(bool), "bool" },
+            { typeof(byte), "byte" },
+            { typeof(char), "char" },
+            { typeof(decimal), "decimal" },
+            { typeof(double), "double" },
+            { typeof(float), "float" },
+            { typeof(int), "int" },
+            { typeof(long), "long" },
+            { typeof(object), "object" },
+            { typeof(sbyte), "sbyte" },
+            { typeof(short), "short" },
+            { typeof(string), "string" },
+            { typeof(uint), "uint" },
+            { typeof(ulong), "ulong" },
+            { typeof(ushort), "ushort" }
+            };
+
+        public static string GetTypeDisplayName([NotNull] Type type, bool fullname = true)
+        {
+            var sb = new StringBuilder();
+            ProcessTypeName(type, sb, fullname);
+            return sb.ToString();
+        }
+
+        private static void AppendGenericArguments(Type[] args, int startIndex, int nArgsToAppend, StringBuilder sb, bool fullName)
+        {
+            var nArgs = args.Length;
+            if (nArgs >= startIndex + nArgsToAppend)
+            {
+                sb.Append("<");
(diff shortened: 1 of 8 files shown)


lajones:
Looks good. :shipit:

test/EntityFramework.Core.Tests/Utilities/TypeNameHelperTest.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace Microsoft.Data.Entity.Tests.Utilities
+{
+    public class TypeNameHelperTest
+    {
+        [Fact]
+        public void Can_pretty_print_CLR_full_name()
+        {
+            // Predefined Types
+            Assert.Equal("System.Int32",
```
bricelam: Can we replace built-in types? They're:
- bool
- byte
- char
- decimal
- double
- float
- int
- long
- object
- sbyte
- short
- string
- uint
- ulong
- ushort

smitpatel: Updated.

Eilon:
Looks like we've now significantly diverged the two "pretty print type" implementations. Can we share the code between the two projects by using the Common repo and shared source?

src/EntityFramework.Core/Utilities/TypeNameHelper.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Utilities
+{
+    public static class TypeNameHelper
+    {
+        private static readonly Dictionary<Type, string> builtInTypeNames = new Dictionary<Type, string>
```
bricelam: Section 3.4, paragraph 2 of the official [ASP.NET 5 Project Coding Guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#coding-guidelines) states that, "All fields MUST begin with an underscore."

Ok, so it actually doesn't... :grinning: but could you make this field start with an underscore to be consistent with the rest of our code base?

bricelam:
:shipit: Those are some mighty pretty type names.

test/EntityFramework.Core.Tests/Utilities/TypeNameHelperTest.cs
```diff
@@ -0,0 +1,127 @@
...
+            Assert.Equal("long", TypeNameHelper.GetTypeDisplayName(typeof(long)));
+            Assert.Equal("object", TypeNameHelper.GetTypeDisplayName(typeof(object)));
+            Assert.Equal("sbyte", TypeNameHelper.GetTypeDisplayName(typeof(sbyte)));
+            Assert.Equal("short", TypeNameHelper.GetTypeDisplayName(typeof(short)));
+            Assert.Equal("string", TypeNameHelper.GetTypeDisplayName(typeof(string)));
+            Assert.Equal("uint", TypeNameHelper.GetTypeDisplayName(typeof(uint)));
+            Assert.Equal("ulong", TypeNameHelper.GetTypeDisplayName(typeof(ulong)));
+            Assert.Equal("ushort", TypeNameHelper.GetTypeDisplayName(typeof(ushort)));
+        }
+        class A { }
```
bricelam: Oh, can you specify a visibility modifier on these nested types too? (`private` is fine)

smitpatel:
created aspnet/common#15

smitpatel:
Files #2054 & aspnet/Logging#175

PR #2053, opened 2015-04-23
Fix FromSql composable stored procedure tests and forced client evaluation

mikary:
@anpete @divega

### DIFF
--- a/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
@@ -106,14 +106,17 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
             var composed = true;
             if (fromSqlAnnotation != null)
             {
-                if (!fromSqlAnnotation.Sql.TrimStart().StartsWith("SELECT", StringComparison.OrdinalIgnoreCase))
+                var sqlStart = fromSqlAnnotation.Sql.Cast<char>().SkipWhile(char.IsWhiteSpace).Take(7).ToArray();
+
+                if (sqlStart.Length != 7
+                    || !char.IsWhiteSpace(sqlStart.Last())
+                    || !new string(sqlStart).StartsWith("SELECT", StringComparison.OrdinalIgnoreCase))
                 {
                     if (QueryModelVisitor.QueryCompilationContext.QueryAnnotations.OfType<IncludeQueryAnnotation>().Any())
                     {
                         throw new InvalidOperationException(Strings.StoredProcedureIncludeNotSupported);
                     }
-                    // Note: All client eval flags should be set here
-                    QueryModelVisitor.RequiresClientFilter = true;
+                    QueryModelVisitor.RequiresClientEval = true;
                     composed = false;
                 }
 
@@ -139,7 +142,7 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
                         _readerParameter
                     };
 
-            if (QueryModelVisitor.QuerySourceRequiresMaterialization(_querySource))
+            if (QueryModelVisitor.QuerySourceRequiresMaterialization(_querySource) || QueryModelVisitor.RequiresClientEval)
             {
                 var materializer
                     = new MaterializerFactory(
--- a/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
@@ -30,6 +30,9 @@ namespace Microsoft.Data.Entity.Relational.Query
         private readonly Dictionary<IQuerySource, SelectExpression> _queriesBySource
             = new Dictionary<IQuerySource, SelectExpression>();
 
+        private bool _requiresClientFilter;
+        private bool _requiresClientResultOperator;
+
         private RelationalProjectionExpressionTreeVisitor _projectionTreeVisitor;
 
         public RelationalQueryModelVisitor(
@@ -40,9 +43,14 @@ namespace Microsoft.Data.Entity.Relational.Query
             _parentQueryModelVisitor = parentQueryModelVisitor;
         }
 
-        public virtual bool RequiresClientFilter { get; set; }
-        public virtual bool RequiresClientProjection => _projectionTreeVisitor.RequiresClientEval;
-        public virtual bool RequiresClientResultOperator { get; set; }
+        public virtual bool RequiresClientEval { get; set; }
+        public virtual bool RequiresClientFilter => _requiresClientFilter || RequiresClientEval;
+        public virtual bool RequiresClientProjection => _projectionTreeVisitor.RequiresClientEval || RequiresClientEval;
+        public virtual bool RequiresClientResultOperator
+        {
+            get { return _requiresClientResultOperator || RequiresClientEval; }
+            set { _requiresClientResultOperator = value; }
+        }
 
         public new virtual RelationalQueryCompilationContext QueryCompilationContext
             => (RelationalQueryCompilationContext)base.QueryCompilationContext;
@@ -227,9 +235,9 @@ namespace Microsoft.Data.Entity.Relational.Query
         public override void VisitWhereClause(WhereClause whereClause, QueryModel queryModel, int index)
         {
             var selectExpression = TryGetQuery(queryModel.MainFromClause);
-            var requiresClientEval = selectExpression == null;
+            var requiresClientFilter = selectExpression == null;
 
-            if (!requiresClientEval)
+            if (!requiresClientFilter)
             {
                 var translatingVisitor = new SqlTranslatingExpressionTreeVisitor(this, whereClause.Predicate);
                 var sqlPredicateExpression = translatingVisitor.VisitExpression(whereClause.Predicate);
@@ -243,17 +251,17 @@ namespace Microsoft.Data.Entity.Relational.Query
                 }
                 else
                 {
-                    requiresClientEval = true;
+                    requiresClientFilter = true;
                 }
 
                 if (translatingVisitor.ClientEvalPredicate != null)
                 {
-                    requiresClientEval = true;
+                    requiresClientFilter = true;
                     whereClause = new WhereClause(translatingVisitor.ClientEvalPredicate);
                 }
             }
 
-            RequiresClientFilter |= requiresClientEval;
+            _requiresClientFilter |= requiresClientFilter;
 
             if (RequiresClientFilter)
             {
@@ -264,8 +272,9 @@ namespace Microsoft.Data.Entity.Relational.Query
         public override void VisitOrderByClause(OrderByClause orderByClause, QueryModel queryModel, int index)
         {
             var selectExpression = TryGetQuery(queryModel.MainFromClause);
+            var requiresClientOrderBy = selectExpression == null;
 
-            if (selectExpression != null)
+            if (!requiresClientOrderBy)
             {
                 var sqlTranslatingExpressionTreeVisitor
                     = new SqlTranslatingExpressionTreeVisitor(this);
@@ -292,12 +301,17 @@ namespace Microsoft.Data.Entity.Relational.Query
                 if (orderings.Count == orderByClause.Orderings.Count)
                 {
                     selectExpression.PrependToOrderBy(orderings);
-
-                    return;
+                }
+                else
+                {
+                    requiresClientOrderBy = true;
                 }
             }
 
-            base.VisitOrderByClause(orderByClause, queryModel, index);
+            if (RequiresClientEval | requiresClientOrderBy)
+            {
+                base.VisitOrderByClause(orderByClause, queryModel, index);
+            }
         }
 
         public override Expression BindMemberToValueBuffer(MemberExpression memberExpression, Expression expression)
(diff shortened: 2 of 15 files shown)


src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -20,6 +21,8 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
 {
     public class RelationalEntityQueryableExpressionTreeVisitor : EntityQueryableExpressionTreeVisitor
     {
+        private static readonly Regex selectRegex = new Regex(@"^\s*select\s+", RegexOptions.IgnoreCase);
```
anpete: underscore

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -258,8 +275,9 @@ public override void VisitWhereClause(WhereClause whereClause, QueryModel queryM
         public override void VisitOrderByClause(OrderByClause orderByClause, QueryModel queryModel, int index)
         {
             var selectExpression = TryGetQuery(queryModel.MainFromClause);
+            var clientOrderBy = selectExpression == null;
 
-            if (selectExpression != null)
+            if (!clientOrderBy)
```
anpete: Match above? i.e. "requiresClientOrderBy"

test/EntityFramework.Core.FunctionalTests/TestHelpers.cs
```diff
@@ -200,7 +200,10 @@ public ModelBuilder CreateConventionBuilder(Model model = null)
             {
                 if (assertOrder)
                 {
-                    Assert.Equal(expected, actual);
+                    for (var index = 0; index < expected.Count; index++)
```
anpete: Reason for this?

test/EntityFramework.SqlServer.FunctionalTests/NorthwindSprocQuerySqlServerFixture.cs
```diff
@@ -0,0 +1,51 @@
...
+            optionsBuilder.UseSqlServer(_testStore.Connection.ConnectionString);
+            _options = optionsBuilder.Options;
+
+            _serviceProvider.GetRequiredService<ILoggerFactory>()
+                .MinimumLevel = LogLevel.Debug;
+        }
+
+        public override NorthwindContext CreateContext()
+        {
+            return new SqlServerNorthwindContext(_serviceProvider, _options);
```
anpete: Turn off ChangeTracker.AutoDetectChangesEnabled

test/EntityFramework.Relational.FunctionalTests/FromSqlQueryTestBase.cs
```diff
@@ -147,30 +148,74 @@ public virtual void From_sql_queryable_simple_composed_include()
...
 
-        //[Fact]
+        [Fact]
         public virtual void From_sql_queryable_stored_procedure()
         {
-            AssertQuery<Product>(
-                cs => cs.FromSql("EXEC ProductsOnOrder"),
-                cs => cs.Where(p => p.UnitsOnOrder > 0),
-                entryCount: 17);
+            AssertQuery(
```
anpete: As is, I don't think these are cross provider. In this case it might be better to just make the whole SQL string a virtual property in the base.

test/EntityFramework.Relational.FunctionalTests/TestModels/NorthwindSproc/NorthwindSprocData.cs
```diff
@@ -0,0 +1,129 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Data.Entity.Relational.FunctionalTests.TestModels.NorthwindSproc
+{
+    public static class NorthwindSprocData
+    {
+        public static MostExpensiveProduct[] TenMostExpensiveProducts()
```
anpete: No need for a deep comparison for these tests. Just manually assert the number of rows in each test + any test required test specific asserts.

test/EntityFramework.Core.FunctionalTests/TestHelpers.cs
mikary: I was having test failures with lists containing the same ordered sequence of elements. It appears the comparison was on the collection being the same instance rather than having the same contents. This behavior also seems to be more consistent with the unordered version below.

test/EntityFramework.Relational.FunctionalTests/TestModels/NorthwindSproc/NorthwindSprocData.cs
mikary: The deep comparisons were important for some of the tests, in particular the forced client eval orderby in the stored procedure composition

test/EntityFramework.Core.FunctionalTests/TestHelpers.cs
anpete: The lists can't ever be the same instance. This was using the xunit assert that does a deep collection compare. We have been using the "asserter" action arg to do custom asserts so I wonder if we can do away with some of these if we change to your code above. Can you check?

mikary:
Updated based on feedback, putting off the change to ```TestHelpers``` for now.

mikary:
Rebased and updated copyright on new files
@divega @anpete

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -106,14 +109,13 @@ protected override Expression VisitEntityQueryable(Type elementType)
             var composed = true;
             if (fromSqlAnnotation != null)
             {
-                if (!fromSqlAnnotation.Sql.TrimStart().StartsWith("SELECT", StringComparison.OrdinalIgnoreCase))
+                if(!_selectRegex.IsMatch(fromSqlAnnotation.Sql))
```
anpete: NIT: Whitespace - suggest using R#, even just for code cleanup.

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -40,9 +44,23 @@ public class RelationalQueryModelVisitor : EntityQueryModelVisitor
...
-        public virtual bool RequiresClientResultOperator { get; set; }
+        public virtual bool RequiresClientEval { get; set; }
+
+        public virtual bool RequiresClientFilter => _requiresClientFilter || RequiresClientEval;
+
+        public virtual bool RequiresClientProjection => _projectionTreeVisitor.RequiresClientEval || RequiresClientEval;
+
+        public virtual bool RequiresClientResultOperator
+        {
+            get
```
anpete: NIT: Doesn't match our code style here.

test/EntityFramework.SqlServer.FunctionalTests/FromSqlSprocQuerySqlServerTest.cs
```diff
@@ -0,0 +1,83 @@
...
+                @"[dbo].[Ten Most Expensive Products]",
+                Sql);
+        }
+
+        public FromSqlSprocQuerySqlServerTest(NorthwindSprocQuerySqlServerFixture fixture)
+            : base(fixture)
+        {
+        }
+
+        protected override string TenMostExpensiveProductsSproc => "[dbo].[Ten Most Expensive Products]";
```
anpete: Dead code?

test/EntityFramework.Relational.FunctionalTests/FromSqlSprocQueryTestBase.cs
```diff
@@ -0,0 +1,208 @@
...
+            Fixture = fixture;
+        }
+
+        protected TFixture Fixture { get; }
+
+        protected abstract string TenMostExpensiveProductsSproc { get; }
+
+        protected abstract string CustomerOrderHistorySproc { get; }
+
+        protected abstract object[] CustomerOrderHistoryParameters { get; }
```
anpete: Why is this abstracted?

test/EntityFramework.Relational.FunctionalTests/AsyncFromSqlQueryTestBase.cs
```diff
@@ -77,22 +182,23 @@ protected AsyncFromSqlQueryTestBase(TFixture fixture)
 
         protected TFixture Fixture { get; }
 
-        private async Task<int> AssertQuery<TItem>(
+        private async Task AssertQuery<TItem>(
```
anpete: Remove this. Just probe the results in each test.

test/EntityFramework.Relational.FunctionalTests/AsyncFromSqlSprocQueryTestBase.cs
```diff
@@ -0,0 +1,209 @@
...
+            using (var context = CreateContext())
+            {
+                var actual = await context
+                    .Set<MostExpensiveProduct>()
+                    .FromSql(TenMostExpensiveProductsSproc)
+                    .ToArrayAsync();
+
+                Assert.Equal(10, actual.Length);
+                Assert.True(
+                    actual.Contains(
```
anpete: Just assert against the scalar props directly. Remove Equals and GetHashCode from the entities.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -20,6 +21,8 @@ namespace Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors
 {
     public class RelationalEntityQueryableExpressionTreeVisitor : EntityQueryableExpressionTreeVisitor
     {
+        private static readonly Regex _selectRegex = new Regex(@"^\s*select\s+", RegexOptions.IgnoreCase);
```
divega: @Eilon was asking me today if a regex wasn't too big of a hammer for this, so giving him a chance to gives us his feedback :smile:

mikary:
More updates from PR feedback. Currently there are 10 tests (sync and async versions of 5 tests) for the non-composed scenario that are failing due to materialization / column order issues.
Do we want to continue with this PR / disable those tests or start adding materialization changes here?
@anpete @divega

divega:
I would prefer those changes to be added here and the tests to not stop passing.

Eilon:
I would definitely not use regex for this, considering it's such a trivial check. It's not even using a compiled regex, so it'll be super slow always (and using a ton of memory). And if you *do* make it be a compiled regex, you'll be taking away regex compilation slots from the app (the slots are few and precious).

ErikEJ:
Fwiw, I use this:

if (test.ToUpperInvariant().StartsWith("SELECT", StringComparison.Ordinal) && test.Length > 6 && char.IsWhiteSpace(test[6]))

Eilon:
I wouldn't use `test.ToUpperInvariant().` because the string could be huge and that would create a copy of the whole thing, whereas we care only about the first few characters in it. Just a super naïve parser could work. Check for whitespace chars, then check for ignore-case "SELECT", then check for one more whitespace char.

ErikEJ:
Thanks, Eilon. Looking forward to the final implementation, I can probably benefit from it in my Little Tool!

mikary:
Added a replacement for the regex, for those interested

anpete:
:shipit:

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/RelationalEntityQueryableExpressionTreeVisitor.cs
```diff
@@ -106,14 +106,17 @@ protected override Expression VisitEntityQueryable(Type elementType)
             var composed = true;
             if (fromSqlAnnotation != null)
             {
-                if (!fromSqlAnnotation.Sql.TrimStart().StartsWith("SELECT", StringComparison.OrdinalIgnoreCase))
+                var sqlStart = fromSqlAnnotation.Sql.Cast<char>().SkipWhile(char.IsWhiteSpace).Take(7).ToArray();
```
Eilon: You just won at LINQ :smile:

ErikEJ: Nice :checkered_flag:

divega: It does look fancy, but unfortunately on my machine this code runs between 10% and 20% slower than the interpreted regex (and FWIW, for this simple regex, interpreted is consistently faster than compiled). Don't take me wrong, it is possible to write a version that is 10 times faster (I did it, just for the sake of it), but you have to stick to very simple parsing logic, i.e. it looks awful :smile: 

At the end of the day, I don't think it was worth switching from the regex.

divega: Correction: the simple parsing version is 30 times faster :smile:

mikary: It would almost be worth moving this (and a fair number of other bits of logic) into its own method / helper, but there's not an obvious place to hang that logic.

Eilon: Yeah I suppose this case demonstrates the price of LINQ as well.

PR #2055, opened 2015-04-23
Support generic GetFieldValue in query

ajcvickers:
This change allows a value reader to be code genned that eagerly reads all values passing required type information to the ADO.NET provider. This requires that the types needed be captured/calculated in query, which is now done as part of query compilation. The code gennned delegates are cached and reused across the stack, not just for query.

Some additional cleanup/optimization and testing still to come. Primarily tested by running all the tests using this value reader on SQL Server, even though it is not currently used by SQL Server by default.


bricelam:
:shipit:

PR #2056, opened 2015-04-24
Use typed read method for Count, All, etc.

ajcvickers:
This allows the ADO.NET provider to know what type of .NET object to create. Ideally we would just use GetFieldValue, but not doing that now because it doesn't work on Mono.

Considered using the IValueReader abstraction, but this would always result in allocation of an array which is not needed, so decided against it.


bricelam:
:shipit:

ajcvickers:
Merged

PR #2065, opened 2015-04-24
Only read values from data reader that will be used

ajcvickers:
Multiple value readers can be used against the same DbDataReader using different offsets into that data reader. This change ensures that only the section of the data reader actually used is read and the buffer is no bigger than it needs to be.


bricelam:
:shipit:

ajcvickers:
Merged

PR #2067, opened 2015-04-24
Post-process DBNulls when using non-typed value reader factory

ajcvickers:
This is a precursor to replacement of IValueReader with ValueBuffer struct. Attempted to do this only when the property value can be null, but this is hard to determine so talked with Andrew and decided to process all values for now.


bricelam:
:shipit:

ajcvickers:
Merged

Issue #2027, opened 2015-04-17
Query :: improve the way we build IsNull expressions for null semantics


maumar:
In null semantics we depend on checking whether a particular expression is null or not, e.g
a == b gets translated into (a == b) || (a == null && b == null)

Currently we assume 'virtal' nature of nulls, that is: f(a, b) == null <=> a == null || b == null
This is correct for most cases, however not always.
For example:
COALESCE: a ?? b == null <=> a == null && b == null
CONDITIONAL: condition ? ifTrue : ifFalse == null <=> (condition && (ifTue == null)) || (!condition && ifFale == null)
LOGICAL AND: a && b == null <=> (a == null && b) || (a && b == null)
LOGICAL OR: a || b == null <=> (a == null && !b) || (!a && b == null)

Currently we can produce invalid results for some null semantics cases with those kinds of operands.

maumar:
Fixed in 50def6473bd450d90129248e45dc0ea5ef86152c

PR #2068, opened 2015-04-25
Fix to #2027 - improve the way we build IsNull expressions for null semantics

maumar:
Problem was that in null semantics, when we test whether expression is null or not, we assumed incorrectly that the expression is null if either of it's parameters is null.
This works for most cases (e.g. functions), but not for all.
Exapmles are COALESCE and CASE-WHEN-THEN-ELSE.

Fix is to refactor the visitor that is used for that purpose, before it was just extracting nullable expressions, now it builds the entire IsNullExpression tree.

Also fixed some minor issues with case expression - we were assuming it would always produce expression of type bool, but it's no longer the case.


src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IsNullExpressionBuildingVisitor.cs
```diff
@@ -0,0 +1,167 @@
...
+            var inExpression = expression as InExpression;
+            if (inExpression != null)
+            {
+                return expression;
+            }
+
+            var caseExpression = expression as CaseExpression;
+            if (caseExpression != null)
+            {
+                if (caseExpression.When is ConditionalExpression)
```
maumar: should I just inline the VisitConditionalExpression here instead?

maumar:
@anpete

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/IsNullExpressionBuildingVisitor.cs
anpete: This seems fine.

test/EntityFramework.SqlServer.FunctionalTests/NullSemanticsQuerySqlServerTest.cs
```diff
@@ -681,6 +727,94 @@ public override void Where_select_not_equal_nullable_with_null_value_parameter()
...
+        }
+
+        public override void Where_not_equal_with_conditional()
+        {
+            base.Where_not_equal_with_conditional();
+
+            Assert.Equal(
+                @"SELECT [e].[Id]
+FROM [NullSemanticsEntity1] AS [e]
+WHERE ((([e].[NullableStringC] <> CASE WHEN ((([e].[NullableStringA] = [e].[NullableStringB] AND ([e].[NullableStringA] IS NOT NULL AND [e].[NullableStringB] IS NOT NULL)) OR ([e].[NullableStringA] IS NULL AND [e].[NullableStringB] IS NULL))) THEN [e].[NullableStringA] ELSE [e].[NullableStringB] END) OR ([e].[NullableStringC] IS NULL OR CASE WHEN ((([e].[NullableStringA] = [e].[NullableStringB] AND ([e].[NullableStringA] IS NOT NULL AND [e].[NullableStringB] IS NOT NULL)) OR ([e].[NullableStringA] IS NULL AND [e].[NullableStringB] IS NULL))) THEN [e].[NullableStringA] ELSE [e].[NullableStringB] END IS NULL)) AND ([e].[NullableStringC] IS NOT NULL OR CASE WHEN ((([e].[NullableStringA] = [e].[NullableStringB] AND ([e].[NullableStringA] IS NOT NULL AND [e].[NullableStringB] IS NOT NULL)) OR ([e].[NullableStringA] IS NULL AND [e].[NullableStringB] IS NULL))) THEN [e].[NullableStringA] ELSE [e].[NullableStringB] END IS NOT NULL))",
```
anpete: It would be great if we could format this better.

anpete:
:shipit:

PR #2072, opened 2015-04-26
Additional true/false literal fix

roji:
The case expression visitor generated CAST(1 AS BIT) for true, but in PostgreSQL the BIT type isn't a valid boolean.

Added TypedTrueLiteral and TypedFalseLiteral to allow overriding.


maumar:
Merged 545315f36a7c397f9a1bb672ceff37b19cff26c3, thanks!

Issue #921, opened 2014-10-17
Add params argument to DbContextActivator.CreateInstance<TContext>()


divega:
The underlying `ITypeActivator.CreateInstance<T>()` allows passing an extra `params object[] parameters` argument that can be used to fill constructor arguments that aren't registered as services. Passing through the argument would be super easy and although it wouldn't be used for DI activation of the DbContext it could simplify other scenarios, e.g. testing of user-derived DbContexts with custom constructors without having to put everything on the DI container.

divega:
This was actually added in 9137045e06d7f56c45c175e9b0a677cba9733a7c for Beta6.

PR #2073, opened 2015-04-27
fixed to #921 - Add params argument to DbContextActivator.CreateInstance...

RickyLin:
this PR addresses #921. I have to add it as an overload method because of [AddScoped method](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Core/Infrastructure/EntityFrameworkServicesBuilder.cs#L74).
In addition, a test seems to require all public api arguments must have `[NotNull]`attribute?

### DIFF
--- a/src/EntityFramework.Core/Internal/DbContextActivator.cs
+++ b/src/EntityFramework.Core/Internal/DbContextActivator.cs
@@ -32,5 +32,19 @@ namespace Microsoft.Data.Entity.Internal
                 _serviceProvider = null;
             }
         }
+
+        public static TContext CreateInstance<TContext>([NotNull] IServiceProvider serviceProvider, [NotNull] params object[] parameters)
+        {
+            try
+            {
+                _serviceProvider = serviceProvider;
+
+                return (TContext)ActivatorUtilities.CreateInstance(serviceProvider, typeof(TContext), parameters);
+            }
+            finally
+            {
+                _serviceProvider = null;
+            }
+        }
     }
 }
--- a/test/EntityFramework.Core.Tests/DbContextTest.cs
+++ b/test/EntityFramework.Core.Tests/DbContextTest.cs
@@ -2034,6 +2034,29 @@ namespace Microsoft.Data.Entity.Tests
             }
         }
 
+        [Fact]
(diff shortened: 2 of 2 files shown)


ajcvickers:
@divega Can you take a look at this since it is your enhancement request that it resolves?

divega:
@RickyLin could you add a test?

RickyLin:
@divega I have added a test.

rowanmiller:
Hey All,

~Rowan

src/EntityFramework.Core/Internal/DbContextActivator.cs
```diff
@@ -21,11 +21,16 @@ public static IServiceProvider ServiceProvider
 
         public static TContext CreateInstance<TContext>([NotNull] IServiceProvider serviceProvider)
         {
+            return CreateInstance<TContext>(serviceProvider, new object[0]);
```
divega: Nit: I would just keep the original call to `ActivatorUtilities.CreateInstance()` and avoid creating a new object[] for the default case.

test/EntityFramework.Core.Tests/DbContextTest.cs
```diff
@@ -2031,6 +2031,29 @@ public void Context_with_options_and_options_action_can_be_used_as_service()
             }
         }
 
+        [Fact]
+        public void Context_with_parameters_can_be_created_By_dbcontextactivator_createinstance()
```
divega: Test looks good to me. It seems we didn't have anything covering `DbContextActivator`. Thanks! 

@ajcvickers any concerns with the test?

divega:
As I mentioned in a comment I think we should leave the existing overload of `CreateInstance<TContext>()` the way it was, but otherwise :shipit:

RickyLin:
I have reverted the default `CreateInstance<TContext>` method.

test/EntityFramework.Core.Tests/DbContextTest.cs
ajcvickers: Looks good.

ajcvickers:
@RickyLin Can you please squash your changes into one commit?

RickyLin:
Why I didn't get the notification.
Anyway, @ajcvickers, I have squashed the changes into one commit.

PR #2075, opened 2015-04-27
Replace IValueReader with ValueBuffer struct

ajcvickers:
This change reflects the fact that currently query requires a buffer and that buffer is always an object array. Also, that buffer is guaranteed to have objects of the correct type (because that is the contract with the data store) and will also not have nulls. This means that IValueReader is acting as a buffer for boxed values. This change makes that explicit and also allows a struct to be used.

Note that we are considering making this buffer use actual materialized entities instead of an object array, but that work isn't scheduled yet.


anpete:
:shipit:

Issue #1954, opened 2015-03-30
Setting EntityType.BaseType does not throw if Child property duplicates new parent property


mikary:
The following test will fail to throw the expected exception:

```cs
        [Fact]
        public void Setting_base_type_throws_when_child_and_parent_contain_duplicate_property()
        {
            var model = new Model();

            var a = new EntityType(typeof(A), model);
            a.AddProperty("E", typeof(string));
            a.AddProperty("G", typeof(string));

            var b = new EntityType(typeof(B), model);

            var c = new EntityType(typeof(C), model);
            c.AddProperty("E", typeof(string));
            c.AddProperty("G", typeof(string));
            c.BaseType = b;

            Assert.Throws<InvalidOperationException>(() => b.BaseType = a);
        }

```

Property duplication checks currently propagate through an ```OnPropertyMetadataChanged``` event. Fixing this issue will probably require a different factoring.

PR #2077, opened 2015-04-27
Fix for issue #1954 - not throwing on duplicate property

mikary:
Fix for #1954
Switching from event based to reference based propagation of property metadata changes to allow for full traversals when adding a base type to an existing tree.
@ajcvickers @anpete

### DIFF
--- a/src/EntityFramework.Core/Metadata/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/EntityType.cs
@@ -37,8 +37,6 @@ namespace Microsoft.Data.Entity.Metadata
         private Key _primaryKey;
         private EntityType _baseType;
 
-        public event EventHandler<Property> PropertyMetadataChanged;
-
         private int _shadowPropertyCount;
 
         private bool _useEagerSnapshots;
@@ -95,6 +93,8 @@ namespace Microsoft.Data.Entity.Metadata
                     return;
                 }
 
+                _baseType = null;
+
                 if (value != null)
                 {
                     if (value.InheritsFrom(this))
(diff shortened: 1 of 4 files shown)


src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -77,7 +71,13 @@ private EntityType(object typeOrName, Model model)
 
             Model = model;
 
+            _foreignKeys = new SortedDictionary<IReadOnlyList<Property>, ForeignKey>(PropertyListComparer.Instance);
```
anpete: Did anything change here? Why have these been moved?

anpete:
Can you describe the change a little more? In particular, the "reference based propagation" approach.

mikary:
Updated based on feedback

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -37,9 +37,11 @@ public class EntityType : Annotatable, IEntityType
         private Key _primaryKey;
         private EntityType _baseType;
 
-        public event EventHandler<Property> PropertyMetadataChanged;
+        private IEnumerable<EntityType> DerivedTypes => Model.EntityTypes.Where(t => t.BaseType == this);
```
anpete: Make this a method and rename to something like GetDirectDescendents

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -95,8 +97,15 @@ public virtual EntityType BaseType
                     return;
                 }
 
+                _baseType = null;
+
                 if (value != null)
                 {
+                    if (Model != value.Model)
```
anpete: Is there any point doing this if we are going to force prevent it at the API level. If we are checking, then we need to check for other possible instances of the problem. E.g. relationships.

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -812,6 +809,35 @@ public virtual Property FindProperty([NotNull] string propertyName)
                 : BaseType?.FindProperty(propertyName);
         }
 
+        private IEnumerable<Property> FindProperties(string[] propertyNames)
+            => propertyNames.Select(n => FindProperty(n)).Where(p => p != null);
+
+        private IEnumerable<Property> FindDerivedProperties(string propertyName)
+            => FindDerivedProperties(new[] { propertyName });
+
+        private IEnumerable<Property> FindDerivedProperties(string[] propertyNames)
```
anpete: Could you use EntityType.GetDerivedTypes().SelectMany(et => et.Properties) here?

src/EntityFramework.Core/Metadata/EntityType.cs
mikary: I think we have some inconsistencies in how aggressively we are guarding against building invalid models.

For example AddKey verifies that the the first element in the list of provided properties are from the appropriate EntityType, but doesn't examine the remaining properties match or even exist in the model. While we do have a model validator that will run later, it seems like we could do a lot more in terms of sanity checking for the users who want to build their own model

anpete: Can you create a bug?

anpete:
Not sure if there is anything outstanding here...

:shipit: if not.

Issue #2074, opened 2015-04-27
Null Semantics :: allow users to opt-out from using null comparison expansion (mostly for perf reasons)


maumar:
By default we expand comparisons to match CLR semantics, but this introduces additional terms to the query (IS NULL, mostly). We should allow customers to opt out of this and just use regular database semantics.

maumar:
Fixed in 406d17f7c465007ff33b02688844bf4cc8fa77cf

PR #2081, opened 2015-04-28
Fix to #2074 - allow users to opt-out from using null comparison expansion

maumar:
Added UseDatabaseNullSemanticsQueryAnnotation that can be used to annotate the query with information that it should not expand null comparisons.


maumar:
@anpete @mikary

src/EntityFramework.Relational/EntityFramework.Relational.csproj
```diff
@@ -148,6 +148,7 @@
     <Compile Include="NonTypedValueReaderFactory.cs" />
     <Compile Include="NonTypedValueReaderFactoryFactory.cs" />
     <Compile Include="Query\Annotations\FromSqlQueryAnnotation.cs" />
+    <Compile Include="Query\Annotations\UseDatabaseNullSemanticsQueryAnnotation.cs" />
```
maumar: Any ideas for a better (shorter) name?

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -31,5 +32,12 @@ public static IQueryable<TEntity> FromSql<TEntity>([NotNull]this DbSet<TEntity>
 
             return dbSet.AnnotateQuery(new FromSqlQueryAnnotation(sql, parameters));
         }
+
+        public static IQueryable<TEntity> UseDatabaseNullSemantics<TEntity>([NotNull] this DbSet<TEntity> dbSet) where TEntity : class
```
mikary: this should probably be an extension off of ```IQueryable<T>``` instead of ```DbSet<T>```

src/EntityFramework.Relational/EntityFramework.Relational.csproj
divega: Not shorter, but maybe more precise: `UseRelationalNullComparisons()`

maumar:
updated the pull request

src/EntityFramework.Relational/Internal/AnnotatedDbSet`.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using Microsoft.Data.Entity.Query;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Relational.Internal
+{
+    public class AnnotatedDbSet<TEntity>
```
anpete: What is this type for?

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/CompositeRelationalExpressionTreeVisitor.cs
```diff
@@ -1,14 +1,29 @@
...
 {
     public class CompositePredicateExpressionTreeVisitor : ExpressionTreeVisitor
     {
+        private ICollection<QueryAnnotation> _queryAnnotations;
+
+        public CompositePredicateExpressionTreeVisitor([NotNull] ICollection<QueryAnnotation> queryAnnotations)
+        {
+            Check.NotNull(queryAnnotations, nameof(queryAnnotations));
+
+            _queryAnnotations = queryAnnotations;
```
anpete: Just pass in the bool "useRelationalNullComparisons "

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -947,5 +949,19 @@ var columnExpression
                 return base.VisitBinaryExpression(expression);
             }
         }
+
+        private class ReducingExpressionVisitor : ExpressionTreeVisitor
```
anpete: Didn't you add this somewhere else already?

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -31,5 +33,14 @@ public static IQueryable<TEntity> FromSql<TEntity>([NotNull]this DbSet<TEntity>
 
             return dbSet.AnnotateQuery(new FromSqlQueryAnnotation(sql, parameters));
         }
+
+        public static DbSet<TEntity> UseRelationalNullComparisons<TEntity>([NotNull] this DbSet<TEntity> dbSet) where TEntity : class
```
divega: After thinking a bit more about it, I would like us to bring the name back one step closer to how the context-level flag looked like in the EF6 API, so I propose we use `UseRelationalNullSemantics()`. 

The reason is that I believe we should use the presence of the annotation to control the SQL generation for other null semantics-related behaviors, such as `String.Concat()` with nulls ([which we implemented in EF6](http://entityframework.codeplex.com/workitem/2457)), and `Enumerable.Sum()` and `Queryable.Sum()` for empty sets, in which we could automatically compensate so that [no casting is needed](http://data.uservoice.com/forums/72025-entity-framework-feature-suggestions/suggestions/2410716-result-of-sum-of-an-empty-set-should-be-0-rather).

divega: @mikary @anpete Anyone remembers why `FromSql()` returns an `IQueryable<TEntity>`? Looks like with this design it should also return a `DbSet<TEntity>` same as the new operator and that both should work in any order applied to the same query.

src/EntityFramework.Relational/Internal/AnnotatedDbSet`.cs
mikary: Would we rather use ```InternalDbSet<TEntity>``` for this? We could add a constructor overload to avoid the new type.

ajcvickers: Why does there need to be another type at all? Can it not be that we create a new real DbSet instance which contains the annotation and return that?

maumar: removed, going to re-use InternalDbSet

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
maumar: this used to be in composite visitor, but now needs to be moved to sql gen, since we need NotNullExpression (which is reducible) so that we know whether we should apply null semantics in the sql gen or not.

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
anpete: Agreed. @mikary?

mikary: Not really, I'm still not convinced that there's a good reason to switch back to a DbSet after FromSql. If we have a compelling reason to compose on FromSql as a DbSet, it's not a breaking change to switch from IQueryable, but until we have a compelling reason, it only seems to enable cases like
```cs
context.Set<Customer>().FromSql(myQueryString).UseRelationalNullSemantics().AddRange(myCustomers);
```

anpete: Yes, it removes the strange API ordering restriction and doesn't seem to have any downsides? I guess it could simplify the impl too.

mikary: I don't like what it does to the API surface for intelisense, it's like having ```Where``` return a DbSet. I don't see FromSql behaving as a query configuration option so much as a part of the query itself that just happens to need to be the first clause so we know how to alias/compose it.

divega: > I don't like what it does to the API surface for intelisense

The only possible downside I see in the Intellisense experience is that you would now be able to chain multiple calls to `FromSql()` on the same query. It is fair point. Most likely code that did this would be written by mistake. However this not worse than being able to call `.ToList().ToList()` as long as we follow consistent rules on how this is interpreted, e.g. we could pick between "last one wins" (my preference) and more than one call results in a run time error. 

> it's like having Where return a DbSet,,,

Why is that? It is true that `FromSql() and `UseRelationalNullSemantics()` are different. `FromSql()` applies to query roots (and you can have more than one root in a query) while `UseRelationalNullSemantics()` is meant to apply to the whole query regardless of how many roots it has. Associating the call to `UseRelationalNullSemantics()` with one of the roots is just a pragmatic compromise (although it is yet another reason for me to prefer the introduction of explicit "relational comparison" operators and a context-level flag over this design). Once you have more than one "query operator" than can only be assigned to the root, it is really very desirable that they can be called in any order.

maumar:
@anpete  - updated the pull request

test/EntityFramework.Relational.FunctionalTests/NullSemanticsQueryTestBase.cs
```diff
@@ -441,18 +524,18 @@ protected void AssertQuery<TItem>(Func<IQueryable<TItem>, IQueryable<TItem>> que
             var actualIds = new List<int>();
             var expectedIds = new List<int>();
 
-            expectedIds.AddRange(l2oQuery(_oracleData.Set<TItem>().ToList().AsQueryable()).Select(e => e.Id).OrderBy(k => k));
+            //expectedIds.AddRange(l2oQuery(_oracleData.Set<TItem>().ToList().AsQueryable()).Select(e => e.Id).OrderBy(k => k));
```
divega: Are the commented lines in this file intentional?

maumar: No, will undo

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
mikary: > The only possible downside I see in the Intellisense experience is that you would now be able to chain multiple calls to ```FromSql()``` on the same query. It is fair point. Most likely code that did this would be written by mistake. 

We were also talking about ```AsNoTracking``` being a 'configuration' option of this type. If we return a ```DbSet``` we also make something like this compile:
```cs
context.Set<Foo>().AsNoTracking().Attach(myFoo);
...
context.SaveChanges();
```
Which looks like it's attaching an untracked entity, but since it's just a query annotation on the DbSet in a code path that isn't actually being visited, the annotation does nothing.

My objection here would go away if we had the extension off of a different type without the extra DbSet functionality, for example:
```cs
public static RootQueryable<TEntity> FromSql<TEntity>([NotNull]this RootQueryable<TEntity> rootQueryable, [NotNull]string sql)
```

> it is really very desirable that they can be called in any order.

I'm not convinced that it is desirable. Part of the reason I left the return type as ```IQueryable``` was due to YAGNI. ```FromSql``` was the only extension that we needed rooted to the beginning of the query. Now that we are adding other options, we were commenting that this pattern isn't very good for the other configuration options anyway.

The original scenario we were discussing with ```FromSql``` was some kind of cross-store situation where each provider has some kind of composable raw command operation, i.e. ```FromSql``` and ```FromRedis``` so we could have something like:

```cs
context.Set<Foo>().FromSql(sqlCommand).FromRedis(redisCommand).Where(f => f.bar == 5);
```

While there is clearly some utility to being able to do this, it also seems to say "I have a context and I don't know what provider it has, but try this." Given that the concept of composable opaque raw commands is fairly new, and that we are talking about cross-store functionality, I think it would be safer to leave it as is rather than introduce a potential pit of failure.

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
```diff
@@ -30,5 +32,14 @@ public static IQueryable<TEntity> FromSql<TEntity>([NotNull] this DbSet<TEntity>
                     new FromSqlQueryAnnotation(
                         Check.NotEmpty(sql, nameof(sql)),
                         Check.NotNull(parameters, nameof(parameters))));
+
+        public static DbSet<TEntity> UseRelationalNullSemantics<TEntity>([NotNull] this DbSet<TEntity> dbSet) where TEntity : class
```
divega: (trying to bring the discussion to the present diff to make finding it less annoying)

> we also make something like this compile: `context.Set<Foo>().AsNoTracking().Attach(myFoo);`...

I see. You are right on that.

> My objection here would go away if we had the extension off of a different type without the extra DbSet functionality

I agree this would be a way to improve it. Let's file a bug on this issue and discuss more options on how to improve it. We can hold off making changes to `FromSql()` until we decide.

Re you other comment I do think it is important to keep this composable and hopefully we will be able to come up with something good. I cannot think how support expressing alternative raw root queries for different providers could become a pit of failure.

divega:
There is a bunch of thing to follow up on but I am fine with pushing this if @anpete finished reviewing.

src/EntityFramework.Relational/RelationalDbSetExtensions.cs
anpete: Right now, AsNoTracking _is_ a real ResultOperator and so it makes sense to exclude it from this discussion.

How about this:

```c#
context.Products
    .ConfigureQuery(o => 
        { 
            o.FromSql("blah") });
            o.UseRelationalNullSemantics();
        })
     .Where(); // ...
```

mikary: Opened #2106 to track this and copied over what we have discussed so far

src/EntityFramework.Core/Extensions/Internal/SharedStringExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Data.Entity.Utilities;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Extensions.Internal
+{
+    public static class SharedStringExtensions
```
anpete: Put this under Query.

src/EntityFramework.Core/Internal/InternalDbSet.cs
```diff
@@ -36,6 +36,14 @@ public InternalDbSet([NotNull] DbContext context)
                         ((IAccessor<IServiceProvider>)_context).Service.GetRequiredService<IEntityQueryProvider>()));
         }
 
+
+        public InternalDbSet([NotNull] IQueryable<TEntity> source)
+        {
```
anpete: Won't this cause a null ref if the user calls any method that needs _context?

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/NullSemanticsExpressionVisitorBase.cs
```diff
@@ -16,5 +18,15 @@ protected Expression BuildIsNullExpression(Expression expression)
 
             return isNullExpressionBuilder.ResultExpression;
         }
+
+        protected override Expression VisitExtensionExpression(ExtensionExpression expression)
+        {
+            if (expression is NotNullableExpression)
```
anpete: Conditional op?

src/EntityFramework.Core/Extensions/Internal/SharedStringExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public static bool IsGeneratedName([NotNull] this string argument)
```
anpete: Can you re-target this extensions to IQuerySource?

anpete:
:shipit: pending comments

PR #2083, opened 2015-04-28
Remove ForRelational

ajcvickers:
This change promotes the general-purpose relational extensions to be first class on the fluent API meaning that there is no longer a ForRelational call. This reflects the idea that relational is special and that the experience should be optimized as such. All other provider extensions, including extensions specific to a given relational provider, should continue use the existing pattern--ForSqlServer, etc.


lajones:
Looks good. Could you also search for "ForRelational()" in [DbContextCodeGeneratorHelper.cs](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGeneratorHelper.cs) and replace all calls to new FacetConfiguration("ForRelational()", xyz) with just new FacetConfiguration(xyz) as well?

Otherwise :shipit:

Issue #2082, opened 2015-04-28
Design and command tests failing when run in Visual Studio


ajcvickers:
Normal development environment, using the .csproj files, TestDriven.NET as the runner.

```
------ Test started: Assembly: EntityFramework.Commands.FunctionalTests.dll ------

Test 'Microsoft.Data.Entity.Commands.ExecutorTest+SimpleProjectTest.GetMigrations_works_cross_domain' failed: System.AggregateException : One or more errors occurred.
---- System.IO.FileNotFoundException : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.
---- The following constructor parameters did not have matching fixture data: SimpleProject project

    ----- Inner Stack Trace #1 (System.IO.FileNotFoundException) -----
    at Microsoft.Data.Entity.Relational.Design.Templating.Compilation.MetadataReferencesProvider..ctor(IServiceProvider serviceProvider)
    at Microsoft.Data.Entity.Commands.DatabaseTool..ctor(IServiceProvider serviceProvider, ILoggerProvider loggerProvider)
    at Microsoft.Data.Entity.Commands.Executor..ctor(Object logHandler, IDictionary args)
    ----- Inner Stack Trace #2 (Xunit.Sdk.TestClassException) -----

Test 'Microsoft.Data.Entity.Commands.ExecutorTest+SimpleProjectTest.GetContextType_works_cross_domain' failed: System.AggregateException : One or more errors occurred.
---- System.IO.FileNotFoundException : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=1.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified.
---- The following constructor parameters did not have matching fixture data: SimpleProject project

    ----- Inner Stack Trace #1 (System.IO.FileNotFoundException) -----
    at Microsoft.Data.Entity.Relational.Design.Templating.Compilation.MetadataReferencesProvider..ctor(IServiceProvider serviceProvider)
    at Microsoft.Data.Entity.Commands.DatabaseTool..ctor(IServiceProvider serviceProvider, ILoggerProvider loggerProvider)
    at Microsoft.Data.Entity.Commands.Executor..ctor(Object logHandler, IDictionary args)
[...]

PR #2084, opened 2015-04-28
Fix for issue 2082 - problems when running tests in VS

lajones:
@ajcvickers Fix for #2082.


ajcvickers:
:sheep::it:

test/EntityFramework.Commands.FunctionalTests/ExecutorTest.cs
```diff
@@ -58,11 +58,13 @@ public SimpleProject()
                                     BuildReference.ByName("System.Interactive.Async", copyLocal: true),
                                     BuildReference.ByName("System.Runtime, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a"),
                                     BuildReference.ByName("System.Data, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"),
+                                    BuildReference.ByName("System.Reflection.Metadata, Version=1.0.19.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a", copyLocal: true),
```
bricelam: Can we not put versions, public key tokens, etc. here?

bricelam: (Doing so will decouple it from package upgrades.)

lajones: Yep. It works without them. I've already committed to dev based on previous sign-off but I'll go through, change all the System.Reflection.Metadata references and re-commit.

bricelam: Cool. The framework ones (e.g. System.Runtime) for whatever reason needed to be full assembly names, but any that we just use the short name for, we should.

lajones:
Fixes checked in with commits 3325df4 & 5bbd6e0.

PR #2092, opened 2015-04-30
Query: Re-simplifying query pipeline executor construction by reverting ...

anpete:
...IValueReader result "bundling". Instead, changed QueryBuffer to use weak reference pattern so that stale, intermediate result entities can be GC'ed. Streamlined querybuffer to no longer track related entities brought in through Include. These are now tracked via a graph walk from the root. Some small performance improvements to ValueBuffer creation.


ajcvickers:
:sheep::it:

src/EntityFramework.Core/project.json
```diff
@@ -23,6 +23,7 @@
                 "System.Collections": { "version": "", "type": "build" },
                 "System.Diagnostics.Debug": { "version": "", "type": "build" },
                 "System.Diagnostics.Tools": { "version": "", "type": "build" },
+                "System.Dynamic.Runtime": { "version": "", "type": "build" },
```
divega: @bricelam good idea to look at this since you have a PR out that touches the same file and we were not sure this was the minimal way to add the dependency.

bricelam: This is correct.

Issue #1951, opened 2015-03-30
Migrations: Separate statements with empty lines


bricelam:
For example:
```C#
migration.DropTable(
    name: "Table1");

migration.DropSequence(
    name: "DefaultSequence");

migration.CreateSequence(
    name: "MySequence",
    type: "bigint",
    startWith: 1L,
    incrementBy: 10);

migration.DropTable(
    name: "Table2");

migration.DropTable(
    name: "Table3");
```

PR #2097, opened 2015-04-30
ModelSnapshot Bugfixes

### LINKED ISSUE #924: Migrations: The C# model code generator does not generate code for candidate keys.

### LINKED ISSUE #1273: Migrations :: Nullability is not added to the model snapshot, resulting in migrations repeatedly doing same work
Consider the following model:

```
    public class Customer
    {
        public int Id { get; set; }

        public Details Details { get; set; }
    }

    public class Details
    {
        public string Id { get; set; }
    }

    {
        public DbSet<Customer> Customers{ get; set; }

        {
            modelBuilder.Entity<Details>().Property(typeof(int), "ShadowKey");
            modelBuilder.Entity<Customer>(m =>
            {
                m.OneToOne(s => s.Details).Required().ForeignKey(typeof(Details), "ShadowKey");
            });
        }
        {
            options.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "IncludeBash", IntegratedSecurity = true }.ToString());
        }
    }
```

Add-Migration will produce the first migration. However every consecutive migration (without making any changes to the model) will also result in a non-empty migration like so:

```
public override void Up(MigrationBuilder migrationBuilder)
        {
            migrationBuilder.DropPrimaryKey("Details", "PK_Details");
            
            migrationBuilder.AlterColumn("Details", "Id", c => c.String(nullable: false));
            
            migrationBuilder.AddPrimaryKey("Details", "PK_Details", new[] { "Id" }, isClustered: true);
        }
        
        public override void Down(MigrationBuilder migrationBuilder)
        {
            migrationBuilder.DropPrimaryKey("Details", "PK_Details");
            
            migrationBuilder.AlterColumn("Details", "Id", c => c.String());
            
            migrationBuilder.AddPrimaryKey("Details", "PK_Details", new[] { "Id" }, isClustered: true);
        }
```

### LINKED ISSUE #1484: Migrations: Replace usage of BasicModelBuilder with direct calls to core metadata
Currently model snapshots get scaffolded as calls to BasicModelBuilder, but the latter is going away as described in #908.  

Although most scenarios we originally envisioned for BasicModelBuilder can be handled by a "regular" ModelBuilder without the conventions, after discussing with @bricelam and @ajcvickers we agreed that the ModelBuilder API is not the best choice for model snapshots.

Instead we would like to use direct calls to the core metadata API.  

We were considering a JSON representation of the model, described in #1355, but we believe that now should be moved to the backlog.

### LINKED ISSUE #1631: Generated migration derived class uses enum with no using directive
version="7.0.0-beta4-12265" 
targetFramework="net451

Add-Migration creates migration derived class using enum from Domain Models but has no using directive, hence code cannot compile.

Error is:
The type or namespace name 'AnimalGroup' could not be found (are you missing a using directive or an assembly reference?)	c:\projects\2013\SightingsWebApp\SightingsWebApp\Migrations\20150216133708_InitialCreate.Designer.cs	32	36	SightingsWebApp

Domain Model:

<pre>
    public class ZooAnimalSighting
    {
        public int ZooAnimalSightingId { get; set; }
        public string Sighter { get; set; }
        public DateTime DateOf { get; set; }
        public bool IsSightingReliable { get; set; }

        public virtual Animal Animal { get; set; }
    }

    public class Zoo
    {
        public int ZooId { get; set; }
        public string City { get; set; }

        public virtual ICollection&lt;Animal&gt; Animals { get; set; }
    }

    public class Animal
    {
        public int AnimalId { get; set; }
        public float Weight { get; set; }
        public AnimalGroup? Group { get; set; }
        public virtual Zoo Zoo { get; set; }
    }

    public enum AnimalGroup
    {
        Invertebrate,
        Fish,
        Amphibian,
        Reptile,
        Mammal,
        Bird
    }
</pre>

### LINKED ISSUE #1976: Migrations :: index added indefinitely (with every migration)
OnModelCreating:

```
modelBuilder.Entity<City>().Index("Shadow", "Name").Unique(true).ForRelational().Name("My index ']; DROP DATABASE FOO; GO; --");
```

Migration (every time):

```
            migration.CreateIndex(
                name: "My index ']; DROP DATABASE FOO; GO; --",
                table: "City",
                columns: new[] { "Shadow", "Name" },
                unique: true);
```

smitpatel:
Update ModelBuilder to allow string based methods
Use ModelBuilder without conventions in Model Snapshot (fixes #1484)
Model Snapshot now scaffolds following data
* Model
  - Annotations
  - EntityTypes
* EntityType
  - Annotations
  - Properties
  - Primarty Key
  - Indexes
  - Foreign Keys
  - Alternate Keys (fixes #924)
* Property (fixes #1631)
  - Annotations
  - IsNullable (fixes #1273)
  - StoreGeneratedPattern
  - MaxLength
  - GenerateValueOnAdd
  - ConcurrencyToken
* Index (fixes #1976)
  - Annotations
  - IsUnique
* Foreign Key
  - Annotations
  - IsRequired
  - IsUnique


smitpatel:
This is ready for review.

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -136,7 +138,7 @@ protected enum GenerateEntityTypeOptions
             stringBuilder
                 .AppendLine()
                 .Append("b.Property<")
-                .Append(_code.Reference(property.ClrType.UnwrapEnumType()))
+                .Append(_code.Reference(property.ClrType.UnwrapNullableType().UnwrapEnumType()))
```
bricelam: Is it OK that we lose the type's nullability here? Should it just unwrap the enum type? (e.g. `MyEnum?` would become `int?`)

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -29,7 +29,7 @@ public virtual void Generate([NotNull] IModel model, [NotNull] IndentedStringBui
             Check.NotNull(model, nameof(model));
             Check.NotNull(stringBuilder, nameof(stringBuilder));
 
-            stringBuilder.Append("var builder = new BasicModelBuilder()");
+            stringBuilder.Append("var builder = new ModelBuilder(new ConventionSet())");
```
bricelam: Can you file a bug to pass this in? (Not that I'll forget or anything... :suspect:)

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -244,13 +294,44 @@ protected virtual void GenerateEntityTypeAnnotations([NotNull] IEntityType entit
...
-                .Append(", ")
-                .Append(string.Join(", ", foreignKey.Properties.Select(p => _code.Literal(p.Name))))
-                .Append(")");
+                .Append(")")
+                .AppendLine();
+
+            using (stringBuilder.Indent())
+            {
+                stringBuilder
+                    .AppendLine(".InverseReference()")
```
bricelam: What if it was a 1:*? If it doesn't matter, `.InverseCollection()` would feel like a more natural default to me.

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
smitpatel: Calling `UnwrapEnumType()` on `MyEnum?` returns `MyEnum?` only. But it make sense to have `int?` for `MyEnum?`.(i.e. `UnwrapEnumType()` returns nullable of underlying type for nullable enum)
Where should I add this logic? ModelSnapshot or modify `UnwrapEnumType()` ?

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -244,13 +294,44 @@ protected virtual void GenerateEntityTypeAnnotations([NotNull] IEntityType entit
...
+                    .Append(".ForeignKey(")
+                    .Append(_code.Literal(foreignKey.EntityType.Name))
```
bricelam: Is this needed? It seems like we already have both entity types at this point:
```C#
mb.Entity("Post") // Dependent
    .Reference("Blog") // Principal
    .InverseCollection()
    .ForeignKey("BlogId");
```

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -244,13 +294,44 @@ protected virtual void GenerateEntityTypeAnnotations([NotNull] IEntityType entit
...
+                    .Append(", ")
+                    .Append(string.Join(", ", foreignKey.Properties.Select(p => _code.Literal(p.Name))))
+                    .Append(")");
+
+                if (foreignKey.PrincipalKey != foreignKey.PrincipalEntityType.GetPrimaryKey())
+                {
+                    stringBuilder
+                        .AppendLine()
+                        .Append(".PrincipalKey(")
+                        .Append(_code.Literal(foreignKey.PrincipalEntityType.Name))
```
bricelam: (Same as above)

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
smitpatel: filed #2128

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
ajcvickers: @smitpatel UnwrapEnumType should be fixed. Make sure to check other places it is used as well for correctness and any possible code simplifications that can be done once it works correctly.

src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
```diff
@@ -221,6 +235,17 @@ public virtual IndexBuilder Index([NotNull] params string[] propertyNames)
...
+            [NotNull] string relatedTypeName,
+            [CanBeNull] string navigationName = null)
+        {
+            Check.NotEmpty(relatedTypeName, nameof(relatedTypeName));
+
+            var relatedEntityType = Builder.ModelBuilder.Entity(relatedTypeName, ConfigurationSource.Explicit).Metadata;
+
+            return new CollectionNavigationBuilder(CollectionBuilder(relatedEntityType, navigationName));
+        }
+
```
bricelam: I wonder if we should just add the methods we'll scaffold: `Reference` & `InverseCollection`

bricelam: Or maybe we just add back `ForeignKey`. @anpete wouldn't object :wink:

bricelam: More context:

Instead of having to do this...
```C#
mb.Entity<Post>().Reference<Blog>().InverseCollection().ForeignKey(p => p.BlogId);
```
...we use to have something like this.
```C#
mb.Entity<Post>().ForeignKey<Blog>(p => p.BlogId);
```

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+        {
+            public string Id { get; set; }
+        }
+
+        #region Model
+
+        [Fact]
+        public void Model_annotations_are_stored_in_snapshot()
+        {
+            var builder = new ModelBuilderFactory().CreateConventionBuilder();
```
bricelam: Could move this into the `Test` template method. Call would look something like:
```
Test(
    builder => builder.Annotation(...),
    "...",
    m =>
    {
        Assert.Whatever(...);
    });

```

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+
+            var code =
+                 @"var builder = new ModelBuilder(new ConventionSet());
+
+builder.Entity(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithOneProperty"", b =>
+    {
+        b.Property<int>(""Id"")
+            .GenerateValueOnAdd()
+            .StoreGeneratedPattern(StoreGeneratedPattern.Identity)
+            .Annotation(""OriginalValueIndex"", 0);
```
bricelam: Could you file a bug to blacklist unused core annotations like this?

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+        b.Key(""Id"");
```
bricelam: We probably want do to something similar to #1951 here too.

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+    });
+
+return builder.Model;
+";
+            Test(builder.Model, code,
+                o =>
+                    {
+                        Assert.Equal(2, o.EntityTypes.Count);
+                        Assert.Equal("Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithOneProperty", o.EntityTypes[0].Name);
+                        Assert.Equal("Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithTwoProperties", o.EntityTypes[1].Name);
```
bricelam: FYI, xUnit.net has an idiom for this

```C#
Assert.Colletction(
    o.EntityTypes,
    t => Assert.Equal("EntityWithOneProperty", t.Name),
    t => Assert.Equal("EntityWithTwoProperties", t.Name));
```

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+                o =>
+                {
+                    Assert.Equal(2, o.EntityTypes.First().GetPrimaryKey().Properties.Count);
+                    Assert.Equal("Id", o.EntityTypes.First().GetPrimaryKey().Properties[0].Name);
+                    Assert.Equal("AlternateId", o.EntityTypes.First().GetPrimaryKey().Properties[1].Name);
+                });
+        }
+
+        [Fact]
+        public void Indexes_are_stored_in_snapshot()
```
bricelam: May want a case for composite indices too.

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+            .Annotation(""OriginalValueIndex"", 1);
+        b.Key(""Id"");
+    });
+
+builder.Entity(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithTwoProperties"", b =>
+    {
+        b.Reference(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithOneProperty"")
+            .InverseReference()
+            .ForeignKey(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithTwoProperties"", ""AlternateId"")
+                .Unique();
```
bricelam: May want to check with @lajones on how reverse engineering scaffolds relationships. I don't think the additional indent should be here on `.Unique()`.

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+            Test(builder.Model, code,
+                o =>
+                {
+                    Assert.Equal(1, o.FindEntityType(typeof(EntityWithTwoProperties)).GetForeignKeys().Count());
+                    Assert.Equal("AlternateId", o.FindEntityType(typeof(EntityWithTwoProperties)).GetForeignKeys().First().Properties[0].Name);
+                });
+        }
+
+        [Fact]
+        public void Alternate_keys_are_stored_in_snapshot()
```
bricelam: Sweet ...another person I've influenced to **not** call them *candidate* keys. :smirk:

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+        b.Key(""Id"");
+    });
+
+builder.Entity(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithOneProperty"", b =>
+    {
+        b.Reference(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithTwoProperties"")
+            .InverseReference()
+            .ForeignKey(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithOneProperty"", ""Id"")
+            .PrincipalKey(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithTwoProperties"", ""AlternateId"")
+                .Unique();
```
bricelam: It might read better to keep `Unique` immediately after `ForeignKey` if possible.

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -0,0 +1,784 @@
...
+            .Annotation(""OriginalValueIndex"", 0);
+        b.Key(""Id"");
+    });
+
+return builder.Model;
+";
+            Test(builder.Model, code,
+                o =>
+                {
+                    Assert.Equal(3, o.EntityTypes[0].GetProperty("Id").Annotations.Count());
```
bricelam: Probably shouldn't assert that there are three if we only really care about one of them.

bricelam:
Should add some unit tests for the new ModelBuilder methods.

bricelam:
This looks amazing; well done!

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
smitpatel: filed #2130

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
smitpatel: `InverseReference()` & `InverseCollection()` are used as per uniqueness of relationship.

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
smitpatel: for `1:1` case the dependent type needs to be explicitly mentioned. For `1:*` case we are not generating the dependent type name again.

src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
smitpatel: @bricelam - Should we add extra methods for consistency or just the one which is used? (i.e. `Reference()`,`InverseReference()` & `InverseCollection()`)

test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
smitpatel: Updated

src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
bricelam: For now, I'd say just the ones we use. The others can wait until we enable unbound navigation properties.

src/Shared/SharedTypeExtensions.cs
```diff
@@ -76,6 +76,15 @@ public static bool IsPrimitive(this Type type)
                || type.IsNonIntegerPrimitive();
 
         public static Type UnwrapEnumType(this Type type)
-            => type.GetTypeInfo().IsEnum ? Enum.GetUnderlyingType(type) : type;
+        {
+            var isNullable = false;
+            if (type.IsNullableType())
```
bricelam: Could just do
```C#
var isNullable = type.IsNullableType();
if (isNullable)
```

smitpatel:
Updated with feedback.

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -121,8 +123,15 @@ protected enum GenerateEntityTypeOptions
...
 
+            var firstProperty = true;
             foreach (var property in properties)
             {
+                if (!firstProperty)
+                {
+                    stringBuilder.AppendLine();
+                }
+
+                firstProperty = false;
```
bricelam: Should set inside the if statement to avoid assigning every iteration

bricelam:
:shipit:

PR #2102, opened 2015-05-01
ValueGeneratorSelectors unwrapping nullable types

### LINKED ISSUE #478: Value generators doesn't work with nullable int keys
Consider the following secnario:

public class Foo
    {
        public int? Id { get; set; }
    }

    {
        public DbSet<Foo> Foos { get; set; }

        //protected override void OnModelCreating(ModelBuilder builder)
        //{
        //    builder.Model.GetEntityType(typeof(Foo)).GetProperty("Id").ValueGenerationOnAdd = ValueGenerationOnAdd.Client;
        //}

        {
            options.UseSqlServer(@"Data Source=.\SQLEXPRESS;Initial Catalog=NullableIntBug;Integrated Security=True;MultipleActiveResultSets=True");
        }
    }

    class Program
    {
        {
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var foo = new Foo
                {
                    Name = "MyFoo"
                };

                ctx.Foos.Add(foo);
                ctx.SaveChanges();
            }
        }
    }

This throws the following:

The instance of entity type 'Foo' cannot be tracked because it has a null primary key. Either set the key explicitly or consider using an IValueGenerator to generate unique key values.

Even after adding explicit value generation for the key (uncomment the lines in the sample above), EF still throws:

No client value generator has been configured for property 'Foo.Id'. To use value generation for properties of type 'Nullable`1' the data store must configure an appropriate value generator.

We should probably support this scenario.

0xdeafcafe:
Unwrapping nullable types, before passing them into their selector. Updated tests to check for nullable Guid's.

Not 100% sure on the workings of the selectors, so this might be the wrong approach. It should also fix #478.

### DIFF
--- a/src/EntityFramework.Core/ValueGeneration/ValueGeneratorSelector.cs
+++ b/src/EntityFramework.Core/ValueGeneration/ValueGeneratorSelector.cs
@@ -29,8 +29,7 @@ namespace Microsoft.Data.Entity.ValueGeneration
         {
             Check.NotNull(property, nameof(property));
 
-            var propertyType = property.ClrType;
-
+            var propertyType = property.ClrType.UnwrapNullableType();
             if (propertyType == typeof(Guid))
             {
                 return _guidFactory.Create(property);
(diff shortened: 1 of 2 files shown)


msotclas:
Hi __@0xdeafcafe__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.msopentech.com.
    </span>

TTYL, MSOTBOT;

msotclas:
__@0xdeafcafe__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, MSOTBOT;

PR #2103, opened 2015-05-01
Clean up project.json files and if directives

bricelam:


lajones:
Looks good. If it were possible to add comments (where they are not already there) for why certain #if's are for DNX451, others are for !DNXCORE50 etc that might help in later reviews of the same code. But don't let that hold you up. :shipit:

Issue #2087, opened 2015-04-28
No value generator is available for properties of type 'Decimal'


scottengle:
I am working with an existing legacy database table that contains a primary key defined as "numeric(18, 0)". In Entity Framework 7.0.0-beta3, this is mapping to a CLR type of Decimal. The Microsoft.Data.Entity.SqlServer.SqlServerValueGeneratorSelector (and its base class Microsoft.Data.Entity.Identity.ValueGeneratorSelector) do not have a comparison for decimal types defined.

Per the error message (see below), I have two options: 1) set a value for the property before adding the entity or 2) configure a value generator for properties of type 'Decimal'. The database is not configured to allow for inserts that include the primary key, so I believe I'm left with the second option. 

How do I configure a value generator for properties of type 'Decimal'?

When I try to modify items in an existing DBSet, I get this error message:
```
System.NotSupportedException was unhandled by user code
  HResult=-2146233067
  Message=The 'Id' on entity type 'Bar' does not have a value set and no value generator is available for properties of type 'Decimal'. Either set a value for the property before adding the entity or configure a value generator for properties of type 'Decimal'.
  Source=EntityFramework.Core
  StackTrace:
       at Microsoft.Data.Entity.Identity.ValueGeneratorSelector.Select(IProperty property)
       at Microsoft.Data.Entity.SqlServer.SqlServerValueGeneratorSelector.Select(IProperty property)
       at Microsoft.Data.Entity.Identity.ValueGeneratorCache.GetGenerator(IProperty property)
       at Microsoft.Data.Entity.ChangeTracking.ValueGenerationManager.MayGetTemporaryValue(IProperty property)
       at Microsoft.Data.Entity.ChangeTracking.StateEntry.MayGetStoreValue(IProperty property)
       at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
       at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
       at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
       at Microsoft.Data.Entity.ChangeTracking.StateEntry.MayGetStoreValue()
       at Microsoft.Data.Entity.ChangeTracking.StateEntry.PrepareToSave()
       at Microsoft.Data.Entity.ChangeTracking.StateManager.<>c.<SaveChanges>b__28_1(StateEntry e)
       at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
[...]

PR #2107, opened 2015-05-01
Add temp value generators for all common property types

ajcvickers:
This allows some form of temporary value generation to happen for any kind of strange key types. Fixes Issue #2087.


lajones:
:ship: :it:

PR #2109, opened 2015-05-01
Fix #1730 - Corrected impl of non-generic CreateQuery and implemented no...

### LINKED ISSUE #1730: Provider.CreateQuery is not castable to IQueryable<T>
Not entirely sure this is a bug, but it used to work before.
Code is hypothetical, but easy to reproduce.

    IQueryable<DomainPackage> domainPackages = this.domainPackageRepository.GetAll();
    // The following should work and do exactly nothing
    domainPackages = (IQueryable<DomainPackage>)domainPackages.Provider.CreateQuery(domainPackages.Expression);

Throws

    Unable to cast object of type 'Microsoft.Data.Entity.Query.EntityQueryable`1[System.Object]' to type 'System.Linq.IQueryable`1[Hostilling.Models.DomainPackage]'.

This bug breaks System.Linq.Dynamic for example https://github.com/kahanu/System.Linq.Dynamic/blob/master/Src/System.Linq.Dynamic/DynamicLinq.cs#L50

Offending line:
https://github.com/aspnet/EntityFramework/commit/bf6360e53f623ffb602bd89a6d678bd85ec8094a#diff-df9548d4cd9a1b427338a0dcb62e4f54L42

    return CreateQuery<object>(expression);

Which explicitly creates IQueryably<object> instead of the apropriate type.

anpete:
...n-generic Execute.

### DIFF
--- a/src/EntityFramework.Core/Query/EntityQueryProvider.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryProvider.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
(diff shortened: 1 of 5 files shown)


ajcvickers:
:shipit:

Issue #1803, opened 2015-03-12
Query :: replace CaseExpression with BCL SwitchExpression


maumar:

maumar:
Fixed in 6b830287fb423586cc250314973ffec4aa948281

PR #2111, opened 2015-05-02
Fix to #1803 - Remove CaseExpression

maumar:
This change removes CaseExpression that has been introduced in EF, and used ConditionalExpression instead. Using this instead of Case/Switch doesn't allow for the full expressiveness of T-SQL CASE WHEN THEN, but it allows creating c# conditional (?:), as well as simple CASE-WHEN-THEN-ELSE that we use currently for our queries.


maumar:
@anpete

anpete:
:shipit:

PR #2127, opened 2015-05-06
Use DataStoreServices instead of just DI for provider-specific service selection

ajcvickers:
The DbContext creates a DI scope for each session and delegates are registered in DI to resolve provider-specific services for that scope based on which provider has been chosen in OnConfiguring. This change extends the use of this mechanism in two ways:
* It is used to select a concrete type from DI, rather than still always pulling an contract type from DI. In other words, the mapping from interface to conrete type is now done here instead of by DI.
* It is used to resolve services that would not otherwise need to be scoped in a provider-specific way.

The consequence of this is that the empty interfaces that were previously used by DI to select one service or another are now no longer needed. Providers now only have their own interfaces when actually adding API surface to the contract.

This also means that common implementations can sometimes also be used. However, it where there is a Singleton that acts as a provider-specific cache, this must always still be a different type for each provider.

Note that even though concrete types are registered in DI, all types resolved from DI (e.g. in constructors) are still contract types. The only place where the concrete types are pulled from DI is in the implementations of IDataStoreServices where the provider-specific mapping is happening.


anpete:
:shipit:

ErikEJ:
This is great! :clap:

PR #2129, opened 2015-05-06
Remove MaxParallelThreads configuration from SqlServer.FunctionalTests

bricelam:
/cc @mikary


anpete:
:shipit:

bricelam:
I'm still seeing deadlocks without this. Two threads get blocked in `AsyncLock.Lock()`. Something isn't releasing the semaphore--possibly a task continuation in `LockAsync()` that is waiting on a free thread to dispatch it. Any ideas, @AndriySvyryd?

bricelam:
Closing for now. Will re-open after we investigate

PR #2137, opened 2015-05-07
Remove travis and appveyor CI support

bricelam:
We'll add these back when we get the agents setup correctly to build and run tests.


lajones:
:shipit:

PR #2139, opened 2015-05-07
Proper TableName for Generic EntityTypes

### LINKED ISSUE #2014: List of Generic Entity
Its possible to code the following
```
public class GenericContext : DbContext
{
    public DbSet<Card> Cards { get; set; }
}
public class Card
{
    public int Id { get; set; }
    public List<Progress<State>> State { get; set; }
    public List<Progress<Risk>> Risk { get; set; }
}

public class Progress<T>
{
    public T State { get; set; }
    public DateTime Stamp { get; set; }
}
public enum State
{
    Active,
    Closed,
    New,
    Removed,
    Resolved
}
public enum Risk
{
    Low,
    Medium,
    High
}
```

smitpatel:
Fixes #2014 
The issue here was the `ClrType.Name` was being used for table name. For `Progress<State>` & `Progress<Risk>`, it gives same result as ``Progress`1``.
We are able to generate distinct EntityTypes but during ModelDiffer stage the CreateTableOperations will point to same table due to above reason, which gives exception while adding migration.

### DIFF
--- a/test/EntityFramework.Core.Tests/Metadata/EntityTypeExtensionsTest.cs
+++ b/test/EntityFramework.Core.Tests/Metadata/EntityTypeExtensionsTest.cs
@@ -39,5 +39,21 @@ namespace Microsoft.Data.Entity.Tests.Metadata
 
             modelMock.Verify(m => m.GetReferencingForeignKeys(entityType), Times.Once());
         }
+
+        [Fact]
+        public void Can_get_proper_table_name_for_generic_entityType()
+        {
+            var entityType = new EntityType(typeof(A<int>), new Model());
+
+            Assert.Equal(
+                "A<int>",
+                ((IEntityType)entityType).DisplayName());
+
+        }
+
+        public class A<T>
+        {
+
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


bricelam:
Could you add a unit test to ensure we don't regress again in the future.

smitpatel:
Test added.

test/EntityFramework.Core.Tests/Metadata/EntityTypeExtensionsTest.cs
```diff
@@ -39,5 +40,22 @@ public void Can_get_referencing_foreign_keys()
 
             modelMock.Verify(m => m.GetReferencingForeignKeys(entityType), Times.Once());
         }
+
+        [Fact]
+        public void Can_get_proper_table_name_for_generic_entityType()
+        {
+            var modelMock = new Mock<Model>();
```
bricelam: Just `new Model()` (no need for a mock)

test/EntityFramework.Core.Tests/Metadata/EntityTypeExtensionsTest.cs
```diff
@@ -39,5 +40,22 @@ public void Can_get_referencing_foreign_keys()
...
+            var entityType = new EntityType(typeof(A<int>), modelMock.Object);
+
+            Assert.Equal(
+                TypeNameHelper.GetTypeDisplayName(typeof(A<int>), false),
```
bricelam: It's better to use a string literal here. I refer to tests like this as, "Testing the product by re-implementing the product." In other words, you're asserting that the product does what the product does instead of asserting that the product returns the correct value.

bricelam:
:shipit:

Issue #2143, opened 2015-05-08
CSharpHelper depends on the local system's culture


MichaelKetting:
When running the tests in CSharpHelperTest on a system using a culture other than US English, the tests fail because the date and number formats are different from the expectation.

Test output (abbreviated) when running with German Culture settings:

Microsoft.Data.Entity.Commands.Utilities.CSharpHelperTest.Literal_works_when_DateTime [FAIL]
Assert.Equal() Failure
Expected: DateTime.Parse("3/12/2015 12:00:00 AM")
Actual:   DateTime.Parse("12.03.2015 00:00:00")

Microsoft.Data.Entity.Commands.Utilities.CSharpHelperTest.Literal_works_when_DateTimeOffset [FAIL]
Assert.Equal() Failure
Expected: ···teTimeOffset.Parse("3/12/2015 12:00:00 AM -07:00")
Actual:   ···teTimeOffset.Parse("12.03.2015 00:00:00 -07:00")

Microsoft.Data.Entity.Commands.Utilities.CSharpHelperTest.Literal_works_when_decimal [FAIL]
Assert.Equal() Failure
Expected: 4.2m
Actual:   4,2m

Microsoft.Data.Entity.Commands.Utilities.CSharpHelperTest.Literal_works(value: 4,2, expected: "4.2") [FAIL]
Assert.Equal() Failure
Expected: 4.2
Actual:   4,2

Microsoft.Data.Entity.Commands.Utilities.CSharpHelperTest.Literal_works(value: 4,2, expected: "4.2f") [FAIL]
Assert.Equal() Failure
Expected: 4.2f
Actual:   4,2f

ErikEJ:
Good PR!

MichaelKetting:
Thanks :)

bricelam:
(Update from PR #2144)

This is actually a product bug--not a test bug. We should be generating valid C# regardless of the culture.

bricelam:
Fixed by #2157

PR #2144, opened 2015-05-08
#2143 Fix culture sensitive tests

MichaelKetting:
CSharpHelperTest uses US-English culture formatting when asserting the text. This fails on machines using other default culture settings. By setting the culture in the setup (and resetting it in tear down), the test can run successfully on non-US systems as well.

### DIFF
--- a/test/EntityFramework.Commands.Tests/Utilities/CSharpHelperTest.cs
+++ b/test/EntityFramework.Commands.Tests/Utilities/CSharpHelperTest.cs
@@ -2,12 +2,27 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Globalization;
+using System.Threading;
 using Xunit;
 
 namespace Microsoft.Data.Entity.Commands.Utilities
 {
-    public class CSharpHelperTest
+    public class CSharpHelperTest : IDisposable
     {
+        private readonly CultureInfo _backupCultureInfo;
+
+        public CSharpHelperTest()
+        {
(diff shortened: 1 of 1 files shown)


MichaelKetting:
The AppVeyor error does not seem related to my commit:
>Specify which project or solution file to use because the folder contains more than one project or solution file.

It also happens in other pull requests. Is there something I can/should do on my end?

bricelam:
@MichaelKetting Ignore Appveyer; it was never setup correctly. @Eilon was going to disable it for our repo.

bricelam:
Hmm, I think this is actually a product bug--not a test bug. We should be generating valid C# regardless of the culture. Instead, let's update [CSharpHelper](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Commands/Utilities/CSharpHelper.cs) to format using the invariant culture.

MichaelKetting:
Oh, yeah, now that you mention it. i guess CSharpHelper and output like "DateTimeOffset.Parse (..." should have been dead giveaways :)

MichaelKetting:
@bricelam Since I already got started on the original discovery... I think I should have some time this weekend to fix this issue and update the pullrequest. Would propose using invariant culture formatting for the number, time, and date time values.

bricelam:
Sounds great! Thank you for spending the time to contribute this. If you don't get around to it, no worries; just close the PR and someone on our team will eventually get assigned to the bug you filed.

Eilon:
I removed the EF project from AppVeyor.

MichaelKetting:
Will be submitting new PR shortly with bugfix for production code instead of test.

Issue #2145, opened 2015-05-08
TemporaryDateTimeOffsetValueGenerator does not work for time zones with a positive UTC offset


MichaelKetting:
When using the TemporaryDateTimeOffsetValueGenerator in a time zone with a positive offset, e.g. Central Europe, an ArgumentException is thrown.
>The UTC time represented when the offset is applied must be between year 0 and 10,000. Parameter name: offset

The reason is perfectly explained in this Stackoverflow post: http://stackoverflow.com/questions/13799214/the-utc-time-represented-when-the-offset-is-applied-must-be-between-year-0-and-1

MichaelKetting:
@ajcvickers I'm assuming the actual time zone used is not important since your commit comment indicated the type is just for generating temp data. I'm proposing changing changing this to always use UTC with offset=0 for the temp value. Pullrequest will be coming shortly.

PR #2146, opened 2015-05-08
#2145 Fix ArgumentException in TemporaryDateTimeOffsetValueGenerator …

MichaelKetting:
…for time zones with a positive UTC offset

DateTimeOffset will fail with an ArgumentException when attempting to initialize with a DateTime that results in a negative offset. This can happen when the actual DateTime value used has a positive UTC offset, e.g. the local time in Europe.

I propose switching the DateTimeOffset to always use a UTC offset = 0 and increment from there.


ajcvickers:
@MichaelKetting Good catch. I'll merge this once the CLA is signed.

ajcvickers:

MichaelKetting:
You're welcome! :smile:

PR #2157, opened 2015-05-10
#2143 Fix CSharpHelper to generate valid C# code …

MichaelKetting:
…regardless of current culture

* DateTime, TimeSpan, and DateTimeOffset are generated as their individual constructor call instead of their Parse-methods.
* Double values are generated with support for round tripping to ensure the max/min-values can also be interpreted.
* Decimal and integer types are ensured to always print in numeric notation, not exponential notation.

I went with ctor-calls for the date-types instead of parsing to make the format more discoverable/avoid ambiguities. If needed, readability could be improved by adding optimized code for offsets that only contain hours, dates that do not contain a time component, etc.

### DIFF
--- a/src/EntityFramework.Commands/Utilities/CSharpHelper.cs
+++ b/src/EntityFramework.Commands/Utilities/CSharpHelper.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 using JetBrains.Annotations;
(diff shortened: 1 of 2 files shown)


src/EntityFramework.Commands/Utilities/CSharpHelper.cs
```diff
@@ -177,17 +178,29 @@ public virtual string Identifier([NotNull] string name, [CanBeNull] ICollection<
...
 
         public virtual string Literal(char value) => "\'" + (value == '\'' ? "\\'" : value.ToString()) + "\'";
-        public virtual string Literal(DateTime value) => "DateTime.Parse(\"" + value + "\")";
-        public virtual string Literal(DateTimeOffset value) => "DateTimeOffset.Parse(\"" + value + "\")";
-        public virtual string Literal(decimal value) => value + "m";
-        public virtual string Literal(double value) => value.ToString();
-        public virtual string Literal(float value) => value + "f";
+
+        public virtual string Literal(DateTime value) =>
+            String.Format(
```
bricelam: We tend to use the C# built-in type names when available (e.g. `string` instead of `String`)

src/EntityFramework.Commands/Utilities/CSharpHelper.cs
```diff
@@ -177,17 +178,29 @@ public virtual string Identifier([NotNull] string name, [CanBeNull] ICollection<
...
+                "new DateTime({0}, {1}, {2}, {3}, {4}, {5}, {6}, DateTimeKind.{7})",
```
bricelam: :smile: Yeah, using `Parse` probably wasn't the best idea.

bricelam:
This looks great! Thank you for helping out. :shipit: I'll merge it after doing a sanity test run on my machine.

src/EntityFramework.Commands/Utilities/CSharpHelper.cs
MichaelKetting: Oops, sorry about that. We do that in re-linq, too, but I actually (thought that I) saw the type name being used so I figured... Turns out, that it was "Strings" (plural) :blush:

PR #2173, opened 2015-05-12
Enable Fluent_api_methods_should_not_return_void

### LINKED ISSUE #2148: Renable Fluent_api_methods_should_not_return_void
It's being removed because it only ran against BasicModelBuilder, which is being removed. We should consider running it against all builder types and investigating those that return void--currently Ignore returns void.

smitpatel:
Resolves #2148


anpete:
:shipit:

PR #2186, opened 2015-05-13
Blacklist core annotations from ModelSnapshot

### LINKED ISSUE #2130: ModelSnapshot: Blacklist unused core annotations
Annotations like `OriginalValueIndex` are appearing in ModelSnapshot which are not required for migration. Blacklist such annotations from ModelSnapshot

smitpatel:
Resolves #2130

### DIFF
--- a/src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
+++ b/src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
@@ -16,6 +16,12 @@ namespace Microsoft.Data.Entity.Commands.Migrations
     {
         private readonly CSharpHelper _code;
 
+        private static readonly List<string> _ignoredAnnotations = new List<string>
+            {
+                CoreAnnotationNames.OriginalValueIndexAnnotation,
+                CoreAnnotationNames.ShadowIndexAnnotation
+            };
+
         public CSharpModelGenerator([NotNull] CSharpHelper code)
         {
             Check.NotNull(code, nameof(code));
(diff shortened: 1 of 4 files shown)


src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -16,6 +16,12 @@ public class CSharpModelGenerator
     {
         private readonly CSharpHelper _code;
 
+        private readonly List<string> _coreAnnotations = new List<string>
+            {
+                "OriginalValueIndex",
```
bricelam: We should probably create a CoreAnnotationNames class (similar to RelationalAnnotationNames or SqlServerAnnotationNames) so we're not copying strings around.

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -16,6 +16,12 @@ public class CSharpModelGenerator
     {
         private readonly CSharpHelper _code;
 
+        private readonly List<string> _coreAnnotations = new List<string>
```
bricelam: A name like `_ignoredAnnotations` might be more appropriate.

bricelam: This can also be static and possibly of type `string[]`.

bricelam:
:shipit:

src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
divega: Could we have metadata in the annotation itself on what kind of annotation is instead of maintaining external lists?

smitpatel: filed #2226

Issue #2160, opened 2015-05-11
QuerySqlServerTest has culture-dependent assertion


MichaelKetting:
Testcase QuerySqlServerTest.Where_chain is culture sensitive. When run with German system culture, the test fails with:
Expected: __p_0: 1/1/1998 12:00:00 AM\r\n\r\nSELECT [o].[Order···
Actual:   __p_0: 01.01.1998 00:00:00\r\n\r\nSELECT [o].[OrderI···

There's two options for this:0
1)  Either, the diagnostic output written by RelationalLoggerExtensions.LogParameters() should always be formatted in the local system's culture. In this case, the culture should be set and reset in the test fixture. 
2) Or, the logger should always use the same formatting, then a culture, e.g. Invariant, should be chosen and possibly even a specific formatting for Date-types (RFC1123, sortable, etc).

@ajcvickers: Since you did the Logger, do you have a preference for what you'd like to see in a PR?

rowanmiller:
@maumar can you investigate what would make the most sense to do here

rowanmiller:
We should try setting the current culture in the test

dougbu:
@maumar stealing this bug because I fixed it last weekend...

divega:
Hey @dougbu, can you explain the fix?

dougbu:
@divega using the `[ReplaceCulture]` attribute from Microsoft.AspNet.Testing on the offending tests solves the problem.

divega:
Thanks! Talked to @dougbu and it seems this should work in most cases where it can reasonably work, e.g. it won't do magic for async tests on 4.5.1 because the culture is in TLS but it should be good enough.

MichaelKetting:
@dougbu Should this fix already be part of the current dev HEAD?

dougbu:
@MichaelKetting no, not in `dev` yet. See PR #2321.

MichaelKetting:
@dougbu  Good to know. Thanks!

dougbu:
b6118c9

ErikEJ:
Another dependency (on AspNet.Testing) - is that good?

PR #2194, opened 2015-05-14
#2160 Fix dependency on local system culture in QuerySqlServerTest by…

MichaelKetting:
… setting a fixed culture.

As discussed over in #2160 :smile:

### DIFF
--- a/test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -1,21 +1,22 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Globalization;
 using System.Linq;
+using System.Threading;
 using Microsoft.Data.Entity.FunctionalTests;
 using Microsoft.Data.Entity.FunctionalTests.TestModels.Northwind;
 using Microsoft.Data.Entity.Relational.FunctionalTests;
 using Microsoft.Data.Entity.Storage;
 using Xunit;
 
-#if DNXCORE50
-using System.Threading;
-#endif
-
 namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
 {
-    public class QuerySqlServerTest : QueryTestBase<NorthwindQuerySqlServerFixture>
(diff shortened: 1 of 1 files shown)


test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -2635,6 +2636,13 @@ public override void Selected_column_can_coalesce()
         public QuerySqlServerTest(NorthwindQuerySqlServerFixture fixture)
             : base(fixture)
         {
+            _backupCultureInfo = Thread.CurrentThread.CurrentCulture;
```
divega: We should use CultureInfo.CurrentCulture instead so that it can flow through async calls and work on Core.

MichaelKetting: Actually, the tests project is set to .NET 4.5, so CultureInfo only has a getter available. Should I still use CultureInfo for getting and Thread for setting?

divega: Ah, good point. @bricelam, what do you think? Could se update the test project?

bricelam: Yes, we should be OK to use .NET 4.6 for our test projects.

MichaelKetting: Good to know. I've created an issue for this (#2203) and will be sending a PR shortly. Will be updating the PR for the current issue once the test projects are have been switched to .NET 4.6.

MichaelKetting:
Since @davidfowl confirmed that .NET 4.6 doesn't fully work at this time and @divega agrees to just use Thread.CultureInfo while there are no async tests depending on it, there's nothing left to do for me with the PR and it's ready for a second go at getting merged. (I've also rebased on top of dev)

rowanmiller:
Hey All,

~Rowan

bricelam:
@MichaelKetting dnx46 is in; do you want to update this to use it before merging?

MichaelKetting:
@bricelam Thanks for the update. Want to first submit the test-project switch to .net 4.6 (#2203). Just have to figure out first if some weird exception is local to my system, a dnx problem or something else weird.

MichaelKetting:
Fixed via alternate in b6118c9

PR #2197, opened 2015-05-14
Fix #2179 - table presence detection not working on SQL Azure DB V12

### LINKED ISSUE #2179: HasTables()/HasTableAsync does not work with Azure SQL Database V12
The following TSQL command (in CreateHasTablesCommand()) always return 1 with "empty" Azure SQL Database V12 because a database_firewall_rules view is added by default in empty databases (script_deployments and script_deployment_status views are added as well after a migration from old Azure SQL Database version to V12):

IF EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES) SELECT 1 ELSE SELECT 0

As this command is used in the HasTables()/HasTablesAsync() method in the SqlServerDataStoreCreator class to check if the table should be created, tables are never created.

ErikEJ:
Fixes issue #2179 by adding a check for the table type in the check for the precense of any user created tables, as the INFORMATION_SCHEMA.TABLES view contains rows on a blank database on Azure SQL DB V12.
Modifeid query verfied against a V12 database (and against localdb by running the tests)

### DIFF
--- a/src/EntityFramework.SqlServer/SqlServerDataStoreCreator.cs
+++ b/src/EntityFramework.SqlServer/SqlServerDataStoreCreator.cs
@@ -88,7 +88,7 @@ namespace Microsoft.Data.Entity.SqlServer
             => _sqlGenerator.Generate(_modelDiffer.GetDifferences(null, model), model);
 
         private string CreateHasTablesCommand()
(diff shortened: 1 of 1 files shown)


ajcvickers:

PR #2198, opened 2015-05-14
Query: Relational: Removed MARS requirement. Query pipeline now trans…

anpete:
…parently buffers outer results when MARS is disabled and an inner reader is about to be opened.

### DIFF
--- a/src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
+++ b/src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Data.Common;
 using System.Diagnostics;
@@ -43,6 +44,7 @@ namespace Microsoft.Data.Entity.Relational.Query
             private readonly AsyncQueryingEnumerable _queryingEnumerable;
 
             private DbDataReader _dataReader;
+            private Queue<ValueBuffer> _buffer;
 
             private bool _disposed;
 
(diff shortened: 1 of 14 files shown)


test/EntityFramework.SqlServer.FunctionalTests/SqlServerTestStore.cs
```diff
@@ -480,10 +483,8 @@ public static string CreateConnectionString(string name)
             return new SqlConnectionStringBuilder
                 {
                     DataSource = @"(localdb)\MSSQLLocalDB",
-                    // TODO: Currently nested queries are run while processing the results of outer queries
-                    // This either requires MARS or creation of a new connection for each query. Currently using
-                    // MARS since cloning connections is known to be problematic.
-                    MultipleActiveResultSets = true,
+                    MultipleActiveResultSets = new Random().Next(0, 2) == 1,
```
divega: I have been told in the past that something like this was a bad idea. Just checking... :smile:

src/EntityFramework.Relational/Query/QueryingEnumerable.cs
```diff
@@ -57,50 +60,83 @@ public Enumerator(QueryingEnumerable queryingEnumerable)
...
+                    _buffer = new Queue<ValueBuffer>();
+
+                    while (_dataReader.Read())
+                    {
+                        _buffer.Enqueue(
+                            _queryingEnumerable._commandBuilder.ValueBufferFactory
+                                .CreateValueBuffer(_dataReader));
+                    }
+
+                    _dataReader.Dispose();
```
divega: Worth wrapping in a finally clause to make sure it does get disposed even if there is a transient failure?

src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
```diff
@@ -57,40 +59,76 @@ public AsyncEnumerator(AsyncQueryingEnumerable queryingEnumerable)
...
+                    _buffer = new Queue<ValueBuffer>();
+
+                    while (await _dataReader.ReadAsync(cancellationToken))
+                    {
+                        _buffer.Enqueue(
+                            _queryingEnumerable._commandBuilder.ValueBufferFactory
+                                .CreateValueBuffer(_dataReader));
+                    }
+
+                    _dataReader.Dispose();
```
divega: Same as with the sync version, it may be worth wrapping the loop into a try block and the dispose into the finally block.

divega:
:shipit:

src/EntityFramework.Relational/Query/QueryingEnumerable.cs
anpete: Will add using.

src/EntityFramework.Relational/Query/AsyncQueryingEnumerable.cs
anpete: Will add using.

test/EntityFramework.SqlServer.FunctionalTests/SqlServerTestStore.cs
anpete: It might be :-)

PR #2200, opened 2015-05-14
Query: Relational: Fixes concurrency issue in SQL gen.

anpete:


src/EntityFramework.Relational/Query/CommandBuilder.cs
```diff
@@ -50,9 +51,11 @@ public class CommandBuilder
                 command.CommandTimeout = (int)connection.CommandTimeout;
             }
 
-            command.CommandText = _sqlGenerator.GenerateSql(parameterValues);
+            var sqlGeneratorFactory = _sqlGeneratorFactory();
```
divega: Should this variable be sqlGenerator?

anpete: Yep, thanks.

Issue #2188, opened 2015-05-14
Improve chances of tests passing on all supported SQL Server editions


ErikEJ:
See https://github.com/aspnet/EntityFramework/issues/2179 - PR in the works

ErikEJ:
Merged

PR #2201, opened 2015-05-15
Fix for issue #2188 - allow tests to pass on Azure SQL Database V12

ErikEJ:
Executing just

    SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES

will return rows on Azure SQL Database V12 even with no user created tables present, see #2179 

This fix limits the rows returned by only returning user created tables (as done already elsewhere in the product code)


ajcvickers:
This is now merged. (Actually yesterday!) Thanks for the contribution!

Issue #2203, opened 2015-05-15
Switch test projects to use .NET 4.6


MichaelKetting:
As suggested by @divega and @bricelam in #2160, the test projects would benefit from switching to .NET 4.6 by enabling the use of .NET 4.6 APIs, e.g. setable CultureInfo.CurrentCulture/CurrentUICulture properties.

ErikEJ:
Concerned if that will break my 4.5 dependent provider?

MichaelKetting:
It should not since this only affects the test projects within the EntityFramework solution. The production projects remain on .NET 4.5.

MichaelKetting:
.NET 4.6 is not fully supported yet: framework dnx46 can't be used at this time.

PR is closed until dnx46 is available. Note to self: also switch from Thread.CurrentCulture to CultureInfo.CurrentCulture when doing the switch.

bricelam:
`dnx46` support is coming in aspnet/dnx#1929

natemcmaster:
Can we reverse this? We can't run our build our test projects against mono (Linux/OSX) if we use `net46`.

MichaelKetting:
The issue originally popped up when it came to using the correct culture in the tests. The actual fix was later done via the ReplaceCulture attribute in XUnit, IRC. No idea if the code actually/still has a requirement for .NET 4.6 to compile.

PR #2204, opened 2015-05-15
#2203 Switch test projects to .NET 4.6…

MichaelKetting:
… to enable use of features such as settable CultureInfo.CurrentCulture/CurrentUICulture properties.

This is in response to @divega and @bricelam suggesting the switch in #2160.

### DIFF
--- a/test/EntityFramework.Commands.FunctionalTests/EntityFramework.Commands.FunctionalTests.csproj
+++ b/test/EntityFramework.Commands.FunctionalTests/EntityFramework.Commands.FunctionalTests.csproj
@@ -11,9 +11,10 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.Data.Entity.Commands</RootNamespace>
     <AssemblyName>EntityFramework.Commands.FunctionalTests</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <NuGetPackageImportStamp>199587f0</NuGetPackageImportStamp>
+    <TargetFrameworkProfile />
   </PropertyGroup>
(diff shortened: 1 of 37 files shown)


bricelam:
You'll also need to update the `project.json` files from `dnx451` to `dnx46`.

MichaelKetting:
Ah, thanks. Hmm... got the following error now on the console via build.cmd:

System.InvalidOperationException: The current runtime target framework is not compatible with 'EntityFramework.Commands.FunctionalTests'.

Current runtime Target Framework: 'DNX,Version=v4.5.1 (dnx451)'
  Type: CLR
  Architecture: x86
  Version: 1.0.0-beta5-11735

Where else do I have to set the FX version? Right now, I've just switched the framework-identifier in the project.json file for the test projects. The production projects still read 'dnx45' and list the assemblies.

bricelam:
Interesting. As far as I know, dnx451 packages are compatible with dnx46 projects. I'll investigate more.

Eilon:
@bricelam I'm pretty sure they *should* be compatible... that's the whole idea :smile: Can ask a package person for more info on this... I think maybe @anurse implemented some of the "what's compatible with what" logic.

MichaelKetting:
@Eilon I just pushed the project.json changes into the PR as well so you guys have the full story available :smile: 
Also noticed that yeah, back with dnx45 and dnx451 there was no problem with having different framework versions. I can't rule out that I'm just missing some setting with dnvm etc but I am using the beta 5 clr, so...

davidfowl:
dnx46 doesn't work yet

MichaelKetting:
@davidfowl Ah, thanks, that's good to know. 
@bricelam, @divega  So, should I shelve this PR until later and we keep using .NET 4.5 and Thread.CurrentCulture over on #2160 for the time being?

divega:
Ok. If there are currently no aync tests which depend on culture Thread.CurrentCulture should work for now. Thanks!

MichaelKetting:
Closing PR until dnx46 is available.

Issue #2121, opened 2015-05-05
Investigate failing SQLite tests


bricelam:
I've implemented them in [bricelam:sqlitest2](https://github.com/aspnet/EntityFramework/compare/dev...bricelam:sqlitest2), but they're not passing.

- [x] AsyncFromSqlQueryTest
- [x] AsyncQueryTest
- [x] BuiltInDataTypesTest
- [x] FromSqlQueryTest
- [x] GearsOfWarQueryTest
- [ ] GraphUpdatesTest
- [x] IncludeOneToOneTest
- [x] InheritanceTest
- [x] MappingQueryTest
- [x] MonsterFixupTest
- [x] NullSemanticsQueryTest
- [x] OptimisticConcurrencyTest
- [x] QueryTest
- [x] StoreGeneratedTest
- [x] TransactionTest

rowanmiller:
@maumar can you look at the query tests that are failing

bricelam:
@natemcmaster You could start looking at some of the non-query tests. OptimisticConcurrencyTest & TransactionTest seem like good starting points.

ErikEJ:
@natemcmaster Welcome back!

natemcmaster:
BuiltInDataTypesTest continues to have problems on DNX core, even with #2230. Failures are due to the table not being created correctly.

natemcmaster:
Blocked on a bug in update pipeline. #2238

PR #2206, opened 2015-05-15
Change CommandParameter name to include the symbol prefix (@)

natemcmaster:
Also updates the tests that rely on logging to use new parameter name format

This paves the way for fixing #2121


bricelam:
:shipit:

PR #2221, opened 2015-05-18
Refactor relational metadata extensions

bricelam:
Changes
- Add IRelationalMetadataExtensionsAccessor
    - Use it in the update pipeline
- Don't derive SqlServerMetadataExtensions from RelationalMetadataExtensions

I considered calling the I**Shared**RelationalExtensions interfaces *Virtual* or *Overridable*. They basically designate which Relational metadata a provider should override. It separates them from the ones that they should not override like DescriminatorColumn and DefaultValue.


ajcvickers:
@bricelam Can you explain a bit about why you are making these changes so I can review from that perspective?

bricelam:
@ajcvickers Added a bit more info. Let me know if I need to add more.

Issue #960, opened 2014-10-24
Make sure that our multithreaded code works correctly across different platforms


divega:
EF contains some custom types that we use for thread safe code such as AsyncLock,  ThreadSafeDictionaryCache, and ThreadSafeLazyRef in which we might be making assumptions about the memory model. We should make sure the locking patterns are safe to use on other platforms.

divega:
Moving this out of @AndriySvyryd's plate since this can be done by someone else.

ajcvickers:
Closing since we are now using standard types for all these things.

PR #2223, opened 2015-05-18
React to Caching api review changes

kichalla:
@Tratcher @anpete 

There is no `GetOrSet` api anymore.

### DIFF
--- a/src/EntityFramework.Core/Query/CompiledQueryCache.cs
+++ b/src/EntityFramework.Core/Query/CompiledQueryCache.cs
@@ -30,6 +30,8 @@ namespace Microsoft.Data.Entity.Query
     {
         public const string CompiledQueryParameterPrefix = "__";
 
+        private static readonly object _compiledQueryLockObject = new object();
+
         private class CompiledQuery
         {
             public Type ResultItemType;
@@ -147,16 +149,15 @@ namespace Microsoft.Data.Entity.Query
                   + new ExpressionStringBuilder()
                       .Build(query);
 
-            var compiledQuery
-                = _memoryCache.GetOrSet(
-                    cacheKey,
-                    Tuple.Create(parameterizedQuery, dataStore),
-                    c =>
-                        {
-                            var tuple = (Tuple<Expression, IDataStore>)c.State;
-
-                            return compiler(tuple.Item1, tuple.Item2);
-                        });
+            CompiledQuery compiledQuery;
+            lock (_compiledQueryLockObject)
+            {
+                if (!_memoryCache.TryGetValue(cacheKey, out compiledQuery))
+                {
+                    compiledQuery = compiler(parameterizedQuery, dataStore);
+                    _memoryCache.Set(cacheKey, compiledQuery);
+                }
+            }
 
             return compiledQuery;
         }


Tratcher:
Nice. :shipit:

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -147,16 +147,12 @@ var cacheKey
...
-                    c =>
-                        {
-                            var tuple = (Tuple<Expression, IDataStore>)c.State;
-
-                            return compiler(tuple.Item1, tuple.Item2);
-                        });
+            CompiledQuery compiledQuery;
+            if(!_memoryCache.TryGetValue(cacheKey, out compiledQuery))
+            {
+                compiledQuery = compiler(parameterizedQuery, dataStore);
```
anpete: How does this guarantee that _compiler_ is only invoked once? Or did GetOrSet not have those semantics previously?

cc @Tratcher

Tratcher: GetOrSet never guaranteed that.

anpete: OK, figured that given this change. 

_compiler_ is really expensive here - Any recommended pattern for doing single init here? 

It superficially seems like it would be a nice feature of the IMemoryCache API.

Tratcher: Cache a Lazy<T>? It will get put into the cache fast and then everyone will block resolving it. It's still not guaranteed unique, but it's much closer.

anpete: It is unique with LazyThreadSafetyMode.ExecutionAndPublication, I believe.

Double-check around TryGetValue?

kichalla:
Updated. /cc: @anpete @Tratcher

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -147,16 +149,15 @@ var cacheKey
...
-                    cacheKey,
-                    Tuple.Create(parameterizedQuery, dataStore),
-                    c =>
-                        {
-                            var tuple = (Tuple<Expression, IDataStore>)c.State;
-
-                            return compiler(tuple.Item1, tuple.Item2);
-                        });
+            CompiledQuery compiledQuery;
+            lock (_compiledQueryLockObject)
```
divega: I think we shouldn't lock unless we didn't get an entry from the cache. On the other hand not sure the double-check pattern is safe for cross platform (see #960). We should probably go back to the original change (i.e. let a the compiler method be invoked more than once) and file a separate issue for optimizing this.

divega: BTW, it would be cool if we could `AsyncLock` this for the async path.

anpete: We should keep the lock. We had a mistaken assumption that GetOrSet guaranteed a single invocation of the creation delegate. The lock fixes that and is unlikely to cause perf issues.

Double-checked should be OK here because MemoryCache is itself thread safe.

BTW, we don't need AsyncLock here because query compilation is a purely sync operation.

anpete:
:shipit: @kichalla

src/EntityFramework.Core/Query/CompiledQueryCache.cs
divega: I am fine with taking this offline. Will explain what I mean.

divega: And by "taking offline" I meant we can merge this change as is now and discuss further improvements later.

PR #2224, opened 2015-05-18
Reintroduce IRelationalMetadataExtensionProvider

bricelam:
Adds an abstract way of accessing provider-specific overrides of relational metadata extensions. The update pipeline now uses it.


ajcvickers:
:shipit:

PR #2225, opened 2015-05-18
Short param names for common override methods

rowanmiller:
Use shorter parameter names for OnModelCreating and OnConfiguring since
they are used repetitively within the user code in those methods.

### DIFF
--- a/src/EntityFramework.Core/DbContext.cs
+++ b/src/EntityFramework.Core/DbContext.cs
@@ -253,11 +253,11 @@ namespace Microsoft.Data.Entity
         ///     affecting other context instances that are constructed with the same <see cref="DbContextOptions" />
         ///     instance.
         /// </remarks>
-        /// <param name="optionsBuilder">
+        /// <param name="options">
         ///     A builder used to create or modify options for this context. Data stores (and other extensions)
         ///     typically define extension methods on this object that allow you to configure the context.
         /// </param>
-        protected internal virtual void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        protected internal virtual void OnConfiguring(DbContextOptionsBuilder options)
         {
         }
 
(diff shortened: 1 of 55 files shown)


anpete:
Not a big fan of this change. It's generally better practice to use full names and the customer can always shorten them if they need to. "model" is especially problematic because we have the concept of a Model, and you could end up writing: model.Model in OnModelCreating.

rowanmiller:
We discussed in last API review, but we can discuss again if needed. I find it really annoying having such long names when writing code in these methods.

divega:
I agree with @anpete, and did not remember what we discussed in API review :smile:. FWIW, 'builder' is a slightly better short name for all these cases if short is the high order bit.

rowanmiller:
I think we discussed it before you were there. `builder` was actually the original proposal, but folks thought (and I agree) that model/options read better in the code. I'd be fine with `builder` though. Short is the high order bit, IntelliSense helps with the longer names but I find that having to choose an item from the list is longer than just typing out the name. It sounds ridiculous I know, but when I was writing app code and typing it over and over again (and getting the wrong item from IntelliSense) I just found it frustrating.

anpete:
I would have to disagree that "short is the high order bit" here. [Clarity is generally more important](https://signalvnoise.com/posts/3250-clarity-over-brevity-in-variable-and-method-names) (code is read many more times than it is written).

Besides that, we had "modelBuilder" in EF6, which existing users are familiar with and have likely internalized to some degree. "optionsBuilder" lines up nicely with that providing a bridge from the customer's mental model.

danielmarbach:
AppVeyor is not configured correctly as described [here](https://github.com/aspnet/EntityFramework/pull/2237#issuecomment-104031767). This PR is also failing.

rowanmiller:
Discussed in API review and decided not to do this

PR #2230, opened 2015-05-19
BuildInDataTypeTests: move EnsureCreated() call into test setup

natemcmaster:
Should help fix the race conditions when running tests on DNX Core.


ajcvickers:
:shipit:

PR #2231, opened 2015-05-19
Fixed concurrency issue in CommandConfigurationTest

ajcvickers:
The test was calling EnsureCreated multiple times in a non-thread safe manner. Switched to use a fixture instead. Also changed the database name since the same name was being used for a different database in different tests.

### DIFF
--- a/test/EntityFramework.SqlServer.FunctionalTests/CommandConfigurationTest.cs
+++ b/test/EntityFramework.SqlServer.FunctionalTests/CommandConfigurationTest.cs
@@ -21,18 +21,51 @@ using Xunit;
 
 namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
 {
-    public class CommandConfigurationTests : IDisposable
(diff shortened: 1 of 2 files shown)


test/EntityFramework.SqlServer.FunctionalTests/CommandConfigurationTest.cs
```diff
@@ -21,18 +20,51 @@
...
 {
-    public class CommandConfigurationTests : IDisposable
+    public class CommandConfigurationTest : IClassFixture<CommandConfigurationTest.CommandConfigurationTestFixture>
     {
-        private readonly IServiceProvider _serviceProvider = new ServiceCollection()
-            .AddEntityFramework()
-            .AddSqlServer()
-            .ServiceCollection()
-            .BuildServiceProvider();
+        private const string DatabaseName = "NotKettleChips";
```
bricelam: :smirk:

bricelam:
:shipit:

PR #2237, opened 2015-05-20
Removed await in favor of returning a task

danielmarbach:
I removed a few unnecessary awaits in favor of returning a task.

This is my first PR to Microsoft so please be patient ;)

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/StateManager.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/StateManager.cs
@@ -329,10 +329,10 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
         protected virtual int SaveChanges(
             [NotNull] IReadOnlyList<InternalEntityEntry> entriesToSave) => _dataStore.SaveChanges(entriesToSave);
 
-        protected virtual async Task<int> SaveChangesAsync(
+        protected virtual Task<int> SaveChangesAsync(
             [NotNull] IReadOnlyList<InternalEntityEntry> entriesToSave,
             CancellationToken cancellationToken = default(CancellationToken))
-            => await _dataStore.SaveChangesAsync(entriesToSave, cancellationToken);
+            => _dataStore.SaveChangesAsync(entriesToSave, cancellationToken);
 
         public virtual void AcceptAllChanges()
         {


danielmarbach:
It seems that AppVeyor is not correctly configured, see

> Specify which project or solution file to use because the folder contains more than one project or solution file.

rowanmiller:
@danielmarbach you can ignore AppVeyor, it's not working yet. We've just disabled it for the moment so that it won't show up on new PRs (we agree it is confusing).

danielmarbach:
@rowanmiller ok.

src/EntityFramework.Core/ChangeTracking/Internal/StateManager.cs
```diff
@@ -302,19 +302,17 @@ public virtual int SaveChanges(bool acceptAllChangesOnSuccess)
 
             if (!entriesToSave.Any())
             {
-                return 0;
+                return Task.FromResult(0);
             }
 
             try
             {
-                var result = await SaveChangesAsync(entriesToSave, cancellationToken);
```
danielmarbach: Arg. Although it safes allocations it now breaks the  https://github.com/aspnet/EntityFramework/blob/dev/test/EntityFramework.Core.Tests/DbContextTest.cs#L2174 test in the async mode because the `Dispose` happens before the `SaveChangesAsync` is done. `IDisposable` and async is a real pain!

danielmarbach:
Now it's ready.

src/EntityFramework.Core/DbContext.cs
```diff
@@ -397,7 +397,7 @@ public virtual Task<int> SaveChangesAsync(CancellationToken cancellationToken =
 
             try
             {
-                return await stateManager.SaveChangesAsync(acceptAllChangesOnSuccess, cancellationToken);
+                return stateManager.SaveChangesAsync(acceptAllChangesOnSuccess, cancellationToken);
```
divega: I am concerned about short-circuiting await here because I imagine it could cause some exceptions to stop being logged because they occur after the first time the thread is yield.

divega: @AndriySvyryd what do you think?

PinpointTownes: Hum, removing `await` here will prevent exceptions thrown asynchronously from being caught by the `catch` block: is this really what you want?

src/EntityFramework.Core/Query/AsyncLinqOperatorProvider.cs
```diff
@@ -147,7 +147,7 @@ public EnumeratorExceptionInterceptor(ExceptionInterceptor<T> exceptionIntercept
                             _inner = _exceptionInterceptor._innerFactory().GetEnumerator();
                         }
 
-                        return await _inner.MoveNext(cancellationToken);
+                        return _inner.MoveNext(cancellationToken);
```
PinpointTownes: Same remark.

src/EntityFramework.Core/DbContext.cs
PinpointTownes: Meh, @divega was too fast for me :smile:

divega: I like your explanation better @PinpointTownes :smile:

danielmarbach: Good point. Wouldn't it save allocations if we add a continuewith onfaulted and log there?

PinpointTownes: Not sure it's really worth it. Have you tried creating a benchmark?

danielmarbach: No I haven't. Tried it out and it makes the code ugly. So I'll revert and force push just one minor change.

danielmarbach:
Based on previous discussion here the updated version

AndriySvyryd:
:shipit:

divega:
Thanks for your contribution @danielmarbach 

FWIW, I think it is possible that we could get rid of more awaits in the EF7 code. In general I would expect the impact to be bigger in places in which we await in tight loops or in multiple levels of nesting, e.g. in the async query enumeration, but it might require some refactoring.

danielmarbach:
@divega Thanks for the feedback. Is still something wrong with that PR, because you didn't merge yet?

divega:
No, nothing wrong. We are just temporarily queuing up some of the PRs until we do the branching for the next preview release. We should be good to merge this in a few days.

rowanmiller:
Hey All,

~Rowan

divega:
This is ready to merge. @ajcvickers can you take care?

ajcvickers:

PR #2240, opened 2015-05-20
Use CancellationToken when possible

danielmarbach:


rowanmiller:
Hey All,

~Rowan

divega:
This looks good to me. Assigning to @lajones so he can merge it and consider adding testing for cancellation in the reverse engineering testing he has been working on.

lajones:
Note: We can pass the cancellationToken in but if you go one step further in the stack you'll see that ExecuteTemplateAsync() calls the override of the abstract method ExecuteAsync() (no cancellationToken) which the templating engine generates from the template. So though we can pass the cancellationToken down through the stack it will not have any effect. To have it take effect we'd have to talk to the Razor team about having the engine generate an ExecuteAsync(CancellationToken) method (and then updating our RazorReverseEngineeringBase ExecuteTemplateAsync(CancellationToken) to call that instead). I'll look into it.

danielmarbach:
Which would be a good change I'd say :)

lajones:
Including @NTaylorMullen who can comment on this from the Razor POV.

NTaylorMullen:
If you want to flow the cancellation token control down to the Razor page there's better ways to do that via `Model`, `ViewContext`, injected properties etc. No need to have the `ExecuteAsync` method take it in.

lajones:
An injected property might work well. We can set that before we call ExecuteTemplateAsync() and check it in each of the WriteXXX() methods.

divega:
I talked to @lajones and he mentioned that adding the cancellation capability on the API has relatively low value unless we actually make the commands cancellable, e.g. in a similar way to [how ASP.NET Hosting does it](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Program.cs#L42), but without waiting at the end :

``` C#
var appShutdownService = host.ApplicationServices.GetRequiredService<IApplicationShutdown>();
Console.CancelKeyPress += delegate { appShutdownService.RequestShutdown(); };
``` 

This hooks up (Ctrl+C) and (Ctrl+Break) to a cancellation token in the IApplicationShutdown implementation, which is part of DNX. you need to then flow the same cancellation token (`appShutdownService.ShutdownRequested`) to any async methods that implement the commands.

Note that we could use any arbitrary `CancellationToken` but `IApplicationShutdown` seems good for consistency. 

@danielmarbach would like to give this and the improvement mentioned by @lajones above (about setting the `CancellationToken` to a property in RazorReverseEngineeringBase) a try?

danielmarbach:
I can give it a try. How urgent is it for you? I'm in Oslo this week and booked out

divega:
For sure it can wait until next week. Thanks!

divega:
This is now superseded by https://github.com/aspnet/EntityFramework/pull/2462 which makes the reverse engineering code async throughout. Thanks go to @danielmarbach. Although we may have not taken your PR, it helped us realize we could take things in this direction and we appreciate that.

Issue #1669, opened 2015-02-20
RevEng: add tests


lajones:
Part of #830. Need to come up with a testing strategy and implement more tests for Reverse Engineering.

lajones:
Added end-to-end test on release branch (dnx451 only) with commit 1d973a1.

bricelam:
(Copying my comment from an earlier PR:)

I think for functional tests, it would good to hit a database and/or compile code. The two big functional scenarios I see are:
* Create various database objects, verify SqlServerMetadataModelProvider outputs a correct model
* Create various models, generate code, verify code, compile code, verify model

lajones:
Merged the end-to-end test for dnx451 into dev branch with commit 776124e.

lajones:
E2E test is updated to work on dnxcore50 as well. See #2452.

lajones:
Adding compilation test with https://github.com/aspnet/EntityFramework/pull/2575

rowanmiller:
@lajones what is still outstanding here?

lajones:
The standard RevEng capabilities are covered through the end-to-end tests, but would like to add more unit tests.

rowanmiller:
Closing, feel free to incrementally add coverage as you work on things.

PR #2241, opened 2015-05-20
Adding E2E Tests for SqlServer.Design

lajones:
See #1669.
Also added FileService to abstract outputting files so can override for tests. And ensured resources have same name when built using VS as when built using command line.


test/EntityFramework.SqlServer.FunctionalTests/SqlServerTestStore.cs
```diff
@@ -83,7 +83,7 @@ private SqlServerTestStore CreateShared(Action initializeDatabase)
             return this;
         }
 
-        public static async Task CreateDatabaseIfNotExistsAsync(string name, string scriptPath = null)
+        public static async Task CreateDatabaseAsync(string name, string scriptPath = null, bool recreateIfAlreadyExists = false)
```
lajones: The previous functionality was only replacing the database if it did not already exist. But in iterating over this during testing I found I needed to replace it multiple times. This also sets us up for the future so that if the DB has to change again (which it probably will as we find more test cases) that can be done easily.

That said, if recreateIfAlreadyExists is false (the default) then the functionality is identical to what was there before.

test/EntityFramework.SqlServer.FunctionalTests/SqlServerTestStore.cs
```diff
@@ -178,7 +210,11 @@ public static void CreateDatabaseIfNotExists(string name, string scriptPath = nu
 
                                 if (appBase != null)
                                 {
-                                    scriptPath = Path.Combine(appBase, Path.GetFileName(scriptPath));
+                                    while (scriptPath.StartsWith(@"..\"))
```
lajones: The previous functionality assumed the script was in the top-level dir relative to appBase. This now allows it to be in a sub-directory.

test/EntityFramework.SqlServer.Design.FunctionalTests/ReverseEngineering/E2ETests.cs
```diff
@@ -0,0 +1,133 @@
...
+                @"For columnId [dbo][AllDataTypes][sql_variantColumn]. Could not find type mapping for SQL Server type sql_variant. Skipping column.",
+                @"For columnId [dbo][AllDataTypes][xmlColumn]. Could not find type mapping for SQL Server type xml. Skipping column.",
+                @"For columnId [dbo][AllDataTypes][geographyColumn]. Could not find type mapping for SQL Server type geography. Skipping column.",
+                @"For columnId [dbo][AllDataTypes][geometryColumn]. Could not find type mapping for SQL Server type geometry. Skipping column.",
+                @"For columnId [dbo][TableWithUnmappablePrimaryKeyColumn][TableWithUnmappablePrimaryKeyColumnID]. Could not find type mapping for SQL Server type hierarchyid. Skipping column.",
+                @"For columnId: [dbo][PropertyConfiguration][PropertyConfigurationID]. The SQL Server data type is tinyint. This will be mapped to CLR type byte which does not allow ValueGenerationStrategy Identity. Generating a matching Property but ignoring the Identity setting.",
```
lajones: I will fix the extra ':' character after columnId

test/EntityFramework.SqlServer.Design.FunctionalTests/ReverseEngineering/E2ETests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            var generator = new ReverseEngineeringGenerator(serviceProvider);
+            var filePaths = generator.GenerateAsync(configuration).Result;
+
+            Assert.Equal(_E2ETestExpectedWarnings, logger.WarningMessages);
+            Assert.Equal(0, logger.InformationMessages.Count);
+            Assert.Equal(0, logger.VerboseMessages.Count);
+
+            var expectedFullFilePaths = _E2ETestExpectedFileNames.Select(name => TestOutputDir + @"\" + name);
+            Assert.Equal(expectedFullFilePaths, filePaths);
```
lajones: I'm in two minds about this. The collection form of the Assert is easy to use. But if it fails you're not likely to get back useful output about why it is failing - it'll more than likely be buried somewhere in the '...' which is output by this to avoid outputting long lines. Perhaps it would be better to loop over each filePath individually asserting. That way if something goes wrong the output will say at exactly which offset it went wrong. Similarly for the file contents below.

lajones:
Closing in favor of #2257 (which has a couple of the issues above addressed and is rebased on top of the latest code).

Issue #2069, opened 2015-04-25
Order by Guid.NewGuid() returns same order every time


dylanparry:
I am attempting to order my table in a random order so that I can select a random row from it. The problem appears to be that the ```OrderBy(…)``` part of the Linq statement is being ignored and no SQL is generated for it.

Here’s what I’m using:

```csharp
var items = await this.context.Items.OrderBy(a => Guid.NewGuid()).FirstAsync();
```

And here’s the SQL that’s being generated for it:

```sql
exec sp_executesql N'SELECT TOP(@p0) [a].[ID], … [other columns omitted]
FROM [Items] AS [a]',N'@p0 int',@p0=1
```

As you can see there’s no ```ORDER BY``` being appended to the SQL, which results in the same "random" row being returned every single time.

I had originally intended on using paging with ```.Skip(…).FirstAsync()``` to select a random row, but as I’m stuck using SQL Server 2008 for the foreseeable future, that doesn’t work either (See https://github.com/aspnet/EntityFramework/issues/1326)

Can anyone confirm that the ```ORDER BY``` isn’t being appended for other cases, and if so is there a workaround?

Thanks.

divega:
@maumar this issue sounds somewhat familiar. Any chance we have a fix already or was it something different?

maumar:
Currently we will produce the following query:

```
SELECT TOP(1) [a].[Id], [a].[Name]
FROM [Items] AS [a]
ORDER BY @__NewGuid_0
```

which will fail with the following exception:
Additional information: The SELECT item identified by the ORDER BY number 1 contains a variable as part of the expression identifying a column position. Variables are only allowed when ordering by an expression referencing a column name.

This is because we funcletize Guid.NewGuid() into a constant. In order for this to work we would have to recognize the method and send it directly to the server, i.e. 

```
SELECT TOP(1) [a].[Id], [a].[Name]
FROM [Items] AS [a]
ORDER BY NEWID()
```

maumar:
Per conversation with Diego and Andrew: plan is to disable partial evalulation for methods and delegate calls and attempt to translate some methods to SQL (e.g. Guid.NewGuid() -> NEWID() )

Users would be able to opt-in to the partial eval by passing the result of the method as a closure, like so:

```
var myGuid = Guid.NewGuid();
await this.context.Items.OrderBy(a => myGuid).FirstAsync();
```

maumar:
[...]

PR #2249, opened 2015-05-21
Fix to #2069 - Guid.NewGuid() returns same order every time

maumar:
Basically the issue is that we used to always attempt to pre-evaluate every method call and property that was not bound to the query (i.e. taking a column as an argument, or where column in an instance of the method/property).
This was generally desirable because users could inline some helper methods into their queries if they needed, rather than have to evaluate it manually outside the query and pass it as closure.
However, this is incorrect for functions that are not deterministic (e.g. Guid.NewGuid(), Datetime.Now). Those functions should always be evaluated on the server.
Users should be able to extend this list of non-deterministic functions (or rather functions that should never be pre-evaluated on the client) to include their own methods (TODO)

Fix is to provide this functionality. On top of that translation for many other BCL functions have been added. Before function that was bound to the query would be evaluated on the client after the information from database was obtained.
This could lead to severe perf issues, if such function was part of the predicate. With this change, function will be translated to corresponding SQL construct, so that entire predicate can be evaluated on the server, resulting less data returned to the client.

Function evaluation is performed by Relinq, and there is no good way to hook into the process, so what we do is temporarily wrap appropriate MethodCallExpressions and MemberExpresions with another expression that prevents evaluation of it's contents.
This way wrapped functions will be preserved and can be passed further into the query pipeline. During sql translation we recognize the methods that we know translations for and translate to appropriate SqlFunctionExpression. Functions that we don't know how to translate will behave as all functions used to before this change - i.e. will be evaluated on the client once all the data is received from the database.


maumar:
@anpete (query), 
@ajcvickers & @mikary (DI), 
@divega  (overall)

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -364,9 +367,190 @@ private static QueryParser CreateQueryParser()
...
+            public static readonly MethodInfo DbContextSetMethodInfo
+                = typeof(DbContext).GetTypeInfo().GetDeclaredMethod("Set");
+
+            private MethodInfo[] _nonDeterministicMethodInfos;
+
+            public FunctionEvaluationDisablingVisitor()
+            {
+                _nonDeterministicMethodInfos = new MethodInfo[]
+                {
+                    typeof(Guid).GetTypeInfo().GetDeclaredMethod("NewGuid"),
```
maumar: those will be provided from outside (configurable in Database?)

src/EntityFramework.Core/Query/ExpressionTreeVisitors/ReducingExpressionTreeVisitor.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        public override Expression VisitExpression([NotNull] Expression node)
+            => node != null
+               && node.CanReduce
+                ? VisitExpression(node.Reduce())
+                : base.VisitExpression(node);
+
+        protected override Expression VisitSubQueryExpression(SubQueryExpression expression)
+        {
+            expression.QueryModel.TransformExpressions(VisitExpression);
```
maumar: @MichaelKetting  shouldn't this be the default behavior for VisitSubquery?

src/EntityFramework.Relational/IRelationalFunctionTranslationProvider.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Data.Entity.Relational.Query.Methods;
+using System.Collections.Generic;
+
+namespace Microsoft.Data.Entity.Relational
+{
+    public interface IRelationalFunctionTranslationProvider
```
maumar: any suggestions for a better name?

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -837,6 +850,27 @@ protected override Exception CreateUnhandledItemException<T>(T unhandledItem, st
         protected virtual string GenerateLiteral([NotNull] object value)
             => string.Format(CultureInfo.InvariantCulture, "{0}", value);
 
+        protected virtual string GenerateLiteral(int value)
```
maumar: missing anything here?

src/EntityFramework.Relational/RelationalDataStore.cs
```diff
@@ -27,6 +27,7 @@ public abstract class RelationalDataStore : DataStore, IRelationalDataStore
         private readonly IBatchExecutor _batchExecutor;
         private readonly IRelationalConnection _connection;
         private readonly IDbContextOptions _options;
+        private readonly IRelationalFunctionTranslationProvider _methodCallTranslatorProvider;
```
maumar: will rename to _functionTranslationProvider

test/EntityFramework.Core.FunctionalTests/BuiltInDataTypesTestBase.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Data.Entity.FunctionalTests
     public abstract class BuiltInDataTypesTestBase<TFixture> : IClassFixture<TFixture>
         where TFixture : BuiltInDataTypesFixtureBase, new()
     {
-        [Fact]
+        //[Fact]
```
maumar: this is just so that Sqlite tests pass (failing randomly, unrelated to the change) - this will not be checked in

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -2969,6 +2978,94 @@ protected static string LocalMethod2()
             return "m";
         }
 
+        //[Fact]
```
maumar: Will uncomment and add more tests, need to fix in memory data for OrderDetails collection first

src/EntityFramework.SqlServer/Query/Methods/StringSubstringTranslator.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using Microsoft.Data.Entity.Relational.Query.Methods;
+using Microsoft.Data.Entity.Relational.Query.Expressions;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.SqlServer.Query.Methods
+{
+    public class StringSubstringTranslator : IMethodCallTranslator
```
anpete: You could DRY some of these up by introducing some base classes for common translators. E.g. StringToLower/Upper etc.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -837,6 +850,27 @@ protected override Exception CreateUnhandledItemException<T>(T unhandledItem, st
...
+            => value.ToString();
+
+        protected virtual string GenerateLiteral(short value)
```
anpete: Most of these were handled by the object overload.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
```diff
@@ -837,6 +850,27 @@ protected override Exception CreateUnhandledItemException<T>(T unhandledItem, st
...
+            => value.ToString();
+
+        protected virtual string GenerateLiteral(byte value)
+            => value.ToString();
+
+        protected virtual string GenerateLiteral(decimal value)
+            => value.ToString();
+
+        protected virtual string GenerateLiteral(double value)
+            => value + "E0";
```
anpete: Make the format string an abstract prop like the date ones.

src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs
maumar: Problem was that I had to add double overload, and once I did that, the int (dynamic) would resolve to the double overload, rather than object.

src/EntityFramework.Core/Query/CompiledQueryCache.cs
anpete: I was hoping we wouldn't need this. Can we just disable all of them?

src/EntityFramework.Core/Query/ExpressionTreeVisitors/ReducingExpressionTreeVisitor.cs
MichaelKetting: @maumar Yes, you are abosolutely correct about that one. The historic reason for this is that the QueryModel itself is mutable. So, you should always visit a clone of the QueryModel, not the original QueryModel and return an new SubQueryExpression. Originally, we did not want to pay the performance penality for cloning the QueryModel, but that's not really that big of an issue. So I think this should get changed for the 2.0 release and I've create an issue for it : https://www.re-motion.org/jira/browse/RMLNQ-73

Thanks, Michael

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -364,9 +367,190 @@ private static QueryParser CreateQueryParser()
...
+        {
+            public Expression Process(Expression expressionTree)
+            {
+                var newExpressionTree = new FunctionEvaluationEnablingVisitor().VisitExpression(expressionTree);
+
+                return newExpressionTree;
+            }
+        }
+
+        private class FunctionEvaluationDisablingVisitor : ExpressionTreeVisitorBase
```
MichaelKetting: @maumar You're right about the extension point lacking in re-linq in regards to the partial evaluation. I dug around a bit in our issue archive and pulled this one here out: https://www.re-motion.org/jira/browse/RMLNQ-7. It's basically the same thing you are doing, only integrated :smile: I've pushed it into the 2.0 release plan and am hopeful that I'll get around to it in the nearish future. If so, I'll get back to you with a PR for your PR.

src/EntityFramework.SqlServer/Extensions/SqlServerEntityServicesBuilderExtensions.cs
```diff
@@ -40,7 +42,8 @@ public static EntityFrameworkServicesBuilder AddSqlServer([NotNull] this EntityF
                     .AddScoped<SqlServerModelDiffer>()
                     .AddScoped<SqlServerMigrationSqlGenerator>()
                     .AddScoped<SqlServerDataStoreCreator>()
-                    .AddScoped<SqlServerHistoryRepository>());
+                    .AddScoped<SqlServerHistoryRepository>()
+                    .AddScoped<IRelationalFunctionTranslationProvider, SqlServerFunctionTranslationProvider>());
```
ajcvickers: This should not be registered as the interface here. Since it is a provider specific service, it needs to go through the provider specific mapping mechanism, which means this interface should be registered as function that is resolved at runtime. The registration here should be for the provider-specific type directly, which is then used by the mapping.

src/EntityFramework.Sqlite/Extensions/SqliteEntityFrameworkServicesBuilderExtensions.cs
```diff
@@ -34,7 +34,8 @@ public static EntityFrameworkServicesBuilder AddSqlite([NotNull] this EntityFram
                     .AddScoped<SqliteDataStoreConnection>()
                     .AddScoped<SqliteMigrationSqlGenerator>()
                     .AddScoped<SqliteDataStoreCreator>()
-                    .AddScoped<SqliteHistoryRepository>());
+                    .AddScoped<SqliteHistoryRepository>()
+                    .AddScoped<IRelationalFunctionTranslationProvider, SqliteFunctionTranslationProvider>());
```
ajcvickers: Same here.

src/EntityFramework.SqlServer/SqlServerFunctionTranslationProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+        private List<IMethodCallTranslator> _sqlServerMethodCallTranslators;
+
+        private List<IPropertyTranslator> _sqlServerPropertyTranslators;
+
+        public SqlServerFunctionTranslationProvider([NotNull] ILoggerFactory loggerFactory)
+            : base(loggerFactory)
+        {
+            _sqlServerMethodCallTranslators = new List<IMethodCallTranslator>
+            {
+                new NewGuidTranslator(),
```
anpete: Did you check whether any of these had any possible impl. in relational?

src/EntityFramework.Relational/RelationalEntityServicesBuilderExtensions.cs
```diff
@@ -41,6 +42,7 @@ public static EntityFrameworkServicesBuilder AddRelational([NotNull] this Entity
                 .AddScoped<RelationalDatabaseFactory>()
                 .AddScoped<RelationalValueGeneratorSelector>()
                 .AddScoped<CommandBatchPreparer>()
+                .AddScoped<RelationalFunctionTranslationProvider>()
```
ajcvickers: This needs to be `.AddScoped(p => GetStoreServices(p).RelationalFunctionTranslationProvider;` to register the mapping function.

src/EntityFramework.Relational/IRelationalFunctionTranslationProvider.cs
anpete: Can we go back to the more command-oriented approach where we have the Translate method and composite translators? Can we just keep IMethodCallTranslator and add IMemberTranslator?

src/EntityFramework.Core/Query/CompiledQueryCache.cs
maumar: @MichaelKetting Awesome, thanks for looking into this!

src/EntityFramework.Core/Query/CompiledQueryCache.cs
maumar: spoke with Andrew in person and it seems that we are ok with having the list (disabling evaluation for all methods prevents inlining some custom methods users would create)

maumar:
@anpete @ajcvickers  new version is up

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -365,9 +368,199 @@ private static QueryParser CreateQueryParser()
...
+                    : expression;
+
+                return _nonDeterministicMethodInfos.Contains(expression.Method)
+                    ? (Expression)new MethodCallEvaluationPreventingExpression(newMethodCall)
+                    : newMethodCall;
+            }
+
+            private bool IsQueryable(Expression expression)
+            {
+                if (expression == null)
```
anpete: Ternary?

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -365,9 +368,199 @@ private static QueryParser CreateQueryParser()
...
                             new PartialEvaluatingExpressionTreeProcessor(),
+                            new FunctionEvaluationEnablingProcessor(),
                             new TransformingExpressionTreeProcessor(ExpressionTransformerRegistry.CreateDefault())
                         })));
 
+        private class FunctionEvaluationEnablingProcessor : IExpressionTreeProcessor
+        {
+            public Expression Process(Expression expressionTree)
+            {
+                var newExpressionTree = new FunctionEvaluationEnablingVisitor().VisitExpression(expressionTree);
```
anpete: Inline variable.

src/EntityFramework.Core/Query/CompiledQueryCache.cs
anpete: These should be static.

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -365,9 +368,199 @@ private static QueryParser CreateQueryParser()
...
+                : base(argument.Type)
+            {
+                Check.NotNull(argument, nameof(argument));
+
+                MethodCall = argument;
+            }
+
+            public override bool CanReduce
+            {
+                get
```
anpete: Formatting.

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -1009,7 +1009,7 @@ var properties
 
             var properties = new List<IPropertyBase>();
 
-            while (memberExpression != null)
+            while (memberExpression != null && memberExpression.Expression != null)
```
anpete: Null conditional?

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -365,9 +368,199 @@ private static QueryParser CreateQueryParser()
...
+
+                return this;
+            }
+        }
+
+        private class PropertyEvaluationPreventingExpression : ExtensionExpression
+        {
+            public virtual MemberExpression MemberExpression { get; private set; }
+
+            public PropertyEvaluationPreventingExpression([NotNull] MemberExpression argument)
```
anpete: Move ctor up.

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -365,9 +368,199 @@ private static QueryParser CreateQueryParser()
...
+
+                return new SubQueryExpression(clonedModel);
+            }
+        }
+
+        private class MethodCallEvaluationPreventingExpression : ExtensionExpression
+        {
+            public virtual MethodCallExpression MethodCall { get; private set; }
+
+            public MethodCallEvaluationPreventingExpression([NotNull] MethodCallExpression argument)
```
anpete: Move ctor up.

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/SqlTranslatingExpressionTreeVisitor.cs
```diff
@@ -215,19 +220,19 @@ var arguments
...
-                            methodCallExpression.Method,
-                            arguments);
-
-                    return _queryModelVisitor.QueryCompilationContext.MethodCallTranslator
-                        .Translate(boundExpression);
+                    var boundExpression = operand != null
+                        ? Expression.Call(operand, methodCallExpression.Method, arguments)
+                        : Expression.Call(methodCallExpression.Method, arguments);
+
+                    var translatedExpression = _queryModelVisitor.QueryCompilationContext.CompositeMethodCallTranslator.Translate(boundExpression);
```
anpete: Formatting.

src/EntityFramework.Relational/Query/Expressions/SqlFunctionExpression.cs
```diff
@@ -0,0 +1,53 @@
...
+            [NotNull] Type returnType)
+            : base(returnType)
+        {
+            FunctionName = functionName;
+            Arguments = new ReadOnlyCollection<Expression>(arguments.ToList());
+        }
+
+        public virtual string FunctionName { get; [param: NotNull] set; }
+
+        public virtual ReadOnlyCollection<Expression> Arguments { get; private set; }
```
anpete: IReadOnlyCollection

src/EntityFramework.Relational/Query/Expressions/SqlFunctionExpression.cs
```diff
@@ -0,0 +1,53 @@
...
+    public class SqlFunctionExpression : ExtensionExpression
+    {
+        public SqlFunctionExpression(
+            [NotNull] string functionName,
+            [NotNull] IEnumerable<Expression> arguments,
+            [NotNull] Type returnType)
+            : base(returnType)
+        {
+            FunctionName = functionName;
+            Arguments = new ReadOnlyCollection<Expression>(arguments.ToList());
```
anpete: I don't think we use this pattern in other places.

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -833,10 +846,32 @@ protected override Exception CreateUnhandledItemException<T>(T unhandledItem, st
 
         private const string DateTimeFormat = "yyyy-MM-ddTHH:mm:ss.fffK";
         private const string DateTimeOffsetFormat = "yyyy-MM-ddTHH:mm:ss.fffzzz";
+        private const string FloatingPointFormat = "{0}E0"; 
 
         protected virtual string GenerateLiteral([NotNull] object value)
             => string.Format(CultureInfo.InvariantCulture, "{0}", value);
 
+        protected virtual string GenerateLiteral(int value)
```
anpete: Any of these need CultureInfo?

src/EntityFramework.SqlServer/Query/Methods/MathCeilingTranslator.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Reflection;
+using Microsoft.Data.Entity.Relational.Query.Methods;
+using Microsoft.Data.Entity.Relational.Query.Expressions;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.SqlServer.Query.Methods
+{
+    public class MathCeilingTranslator : IMethodCallTranslator
+    {
+        public virtual Expression Translate([NotNull] MethodCallExpression methodCallExpression)
```
anpete: Introduce a base class and make Translate a template method.

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -2285,6 +2300,43 @@ public override void String_Contains_MethodCall()
                 Sql);
         }
 
+//        public override void Where_math_abs()
```
anpete: These supposed to be commented out?

test/EntityFramework.CrossStore.FunctionalTests/EndToEndTest.cs
```diff
@@ -114,6 +115,11 @@ public SqliteEndToEndTest(SqliteCrossStoreFixture fixture)
             : base(fixture)
         {
         }
+
+        public override void Can_save_changes_and_query()
```
anpete: Can remove.

src/EntityFramework.Relational/Query/RelationalCompositeMemberTranslator.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Microsoft.Data.Entity.Relational.Query.Methods;
+using Microsoft.Framework.Logging;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Relational.Query
+{
+    public abstract class RelationalCompositeMemberTranslator : IMemberTranslator
```
ajcvickers: If it's possible and reasonable for providers to not have any custom translators (such as appears to be the case now for SQLite) then don't make this abstract and force the provider to implement an override that does nothing. And then in, for example, AddSqlite just register the base implementation.

Other than that, the DI code now looks good.

src/EntityFramework.Relational/Query/RelationalCompositeMemberTranslator.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        private List<IMemberTranslator> _relationalTranslators;
+
+        public RelationalCompositeMemberTranslator([NotNull] ILoggerFactory loggerFactory)
+        {
+            _relationalTranslators = new List<IMemberTranslator>();
```
ajcvickers: Inline initialization and make field readonly. Also for other classes in this change.

src/EntityFramework.Relational/Query/RelationalCompositeMemberTranslator.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.Framework.Logging;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Relational.Query
+{
+    public abstract class RelationalCompositeMemberTranslator : IMemberTranslator
+    {
+        private List<IMemberTranslator> _relationalTranslators;
+
+        public RelationalCompositeMemberTranslator([NotNull] ILoggerFactory loggerFactory)
```
ajcvickers: Don't take a loggerFactory if its never going to be used or passed to anything else. Also for other classes in this change.

maumar:
@anpete @ajcvickers  new version is up, applied CR feedback, added a few more tests, and few method translators for Sqlite. More tests to come with a different checkin.

src/EntityFramework.Core/Query/CompiledQueryCache.cs
```diff
@@ -187,9 +190,183 @@ private static QueryParser CreateQueryParser()
...
 
+        private class FunctionEvaluationEnablingProcessor : IExpressionTreeProcessor
+        {
+            public Expression Process(Expression expressionTree)
+            {
+                return new FunctionEvaluationEnablingVisitor().VisitExpression(expressionTree);
+            }
+        }
+
+        //private class FunctionEvaluationDisablingVisitor : ExpressionTreeVisitorBase
```
anpete: Dead?

src/EntityFramework.Relational/Query/ExpressionTreeVisitors/SqlTranslatingExpressionTreeVisitor.cs
```diff
@@ -119,9 +119,14 @@ protected override Expression VisitConditionalExpression(ConditionalExpression e
             var ifTrue = VisitExpression(expression.IfTrue);
             var ifFalse = VisitExpression(expression.IfFalse);
 
-            Equals(ifTrue.Type, ifFalse.Type);
+            if (test != null && ifTrue != null && ifFalse != null)
+            {
+                Equals(ifTrue.Type, ifFalse.Type);
```
anpete: What is this line for?

anpete:
:shipit:

ErikEJ:
@maumar @anpete  Please do not delete branches, I am a provider writer, and need to see what has changed in an easy way

Issue #875, opened 2014-10-15
Create provider-specific extensions for SQLite


ajcvickers:
Not only for things unique to SQLite, but also so that things like column types can be set differently on SQLite. Places in the code that currently just use .Relational() when SQLite is in use should be updated.

bricelam:
Reassigning to myself since I'm working in this area for Migrations.

PR #2250, opened 2015-05-21
Sqlite: Implement metadata

bricelam:
Resolves: #875

Note, this uses the new syntax we discussed in API Review today. Fore example:
```C#
modelBuilder.Entity<Person>().SqliteTable("people");
```

Also note, `entityType.Sqlite().Schema` and `model.Sqlite().Sequences` return `null` and empty since those concepts don't exist on SQLite.


src/EntityFramework.Sqlite/Metadata/SqliteMetadataExtensionProvider.cs
```diff
@@ -8,12 +8,11 @@ namespace Microsoft.Data.Entity.Sqlite.Metadata
...
     public class SqliteMetadataExtensionProvider : IRelationalMetadataExtensionProvider
     {
-        // TODO: Update with #875
-        public virtual IRelationalEntityTypeExtensions Extensions(IEntityType entityType) => entityType.Relational();
-        public virtual IRelationalForeignKeyExtensions Extensions(IForeignKey foreignKey) => foreignKey.Relational();
-        public virtual IRelationalIndexExtensions Extensions(IIndex index) => index.Relational();
-        public virtual IRelationalKeyExtensions Extensions(IKey key) => key.Relational();
-        public virtual IRelationalModelExtensions Extensions(IModel model) => model.Relational();
-        public virtual IRelationalPropertyExtensions Extensions(IProperty property) => property.Relational();
+        public virtual IRelationalEntityTypeExtensions Extensions(IEntityType entityType) => entityType.Sqlite();
```
lajones: Should these return the ISqlliteXXXExtensions instead?

lajones:
:shipit:

src/EntityFramework.Sqlite/Metadata/SqliteMetadataExtensionProvider.cs
bricelam: No; they're defined on the interface which is a Relational concept.

ErikEJ:
@bricelam I am trying to understand the purpose of implementing all this in each provider, can you "enlighten" me ? :-)

bricelam:
It's so that you can override things on a per-provider basis. E.g. one table name on SQLite, one on SQL Server, and one everywhere else. @rowanmiller commented on Friday that we may want to consider just removing it.

ErikEJ:
@bricelam  Thanks for the heads up, I will just leave it for now

Issue #2153, opened 2015-05-09
Help with my configuration: Could not load file or assembly 'System, Version=4.0.0.0 ...


GuardRex:
Feels like I haven't configured something correctly. This is a CoreCLR project on Win7 x64 with runtime `dnx-coreclr-win-x64.1.0.0-beta5-xxxxx`. Latest xxxxx is 11760 (5/14).

**project.json dependencies**
```c#
EntityFramework.SqlServer: 7.0.0-beta5-*
EntityFramework.Commands: 7.0.0-beta5-*
```
**project.json commands**
```c#
"ef": "EntityFramework.Commands"
```
**Startup.cs ConfigureServices**
```c#
services.AddEntityFramework()
    .AddSqlServer()
    .AddDbContext<DbDbContext>(options => options.UseSqlServer(config.Get("Database")));
```
**Model**
```c#
namespace MyApp.Models {
    public class DbDbContext : DbContext {

        public DbSet<Test> Tests { get; set; }

        protected override void OnModelCreating(ModelBuilder builder) {
            builder.Entity<Test>().Property(typeof(int), "id").ForSqlServer().UseIdentity();
            base.OnModelCreating(builder);
        }

    }

    public class Test {
        [Key]
        public int id { get; set; }
        public string field1 { get; set; }
        public int field2 { get; set; }
    }
}
```
Executing from an elevated command line `dnx . ef` results in ...
```
System.IO.FileNotFoundException: Could not load file or assembly 'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' or one of its dependencies. The system cannot find the file specified.
File name: 'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' ---> System.IO.FileNotFoundException: Could not load the specified file.
File name: 'System'
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyName(AssemblyName assemblyName)
   at System.Runtime.Loader.AssemblyLoadContext.Resolve(IntPtr gchManagedAssemblyLoadContext, AssemblyName assemblyName)
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks)
[...]

PR #2251, opened 2015-05-21
[Small] Change MetadataReferencesProvider for dnxcore50

lajones:
@bricelam @GuardRex
Fix for #2153.
Set up MetadataReferencesProvider to initialize using different references on dnxcore50. Also set up to initialize them lazily - previously MetadataReferencesProvider would try to find these references as soon as you created an instance of it (which was typically done by migrations any time you tried any migration command). With this fix if you never actually use MetadataReferencesProvider then you never initialize.


GuardRex:
@lajones Thanks ... I'm looking forward to this.

bricelam:
:shipit:

lajones:
Checked in with commit beb430b.

PR #2254, opened 2015-05-22
fix small typo

va5ili5:
the "EF7 is not recommend for:" should be "EF7 is not recommended for:"


natemcmaster:
Thanks for catching our tpyos.

natemcmaster:
Merged 739a7233bbd49a7130e55a48188d3dc2a59f5a0c

PR #2255, opened 2015-05-22
Version Core and Relational functional tests

### LINKED ISSUE #2252: Versioning of EF Test MyGet packages
Great that those test packages are now available for easy comsumprion in for example providers. But there seems to be a issue with the versioning, as both packages have version 1.0.0.  It would be useful if they had the same version number as the "daily" build.

     EntityFramework.Core.FunctionalTests  v1.0.0

bricelam:
Fixes #2252

### DIFF
--- a/test/EntityFramework.Core.FunctionalTests/project.json
+++ b/test/EntityFramework.Core.FunctionalTests/project.json
@@ -1,4 +1,6 @@
 {
+  "version": "7.0.0-*",
+  "description": "Reusable EntityFramework.Core functional tests for provider writers.",
   "dependencies": {
     "EntityFramework.InMemory": "7.0.0-*"
   },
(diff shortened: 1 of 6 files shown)


lajones:
:shipit: Looks good. By why only these test projects?

bricelam:
@lajones They're the only ones we "ship".

lajones:
Sure - they're the only ones needed right now - but wouldn't it be more consistent to have them all the same?

bricelam:
Tests don't need versions. I'm really just adding a version to the nupkgs. The other ones don't ever produce nupkgs--they're ran from source.

bricelam:
Eventually we'll even remove these ones once we have a long-term solution for testing providers.

lajones:
OK. Not a big issue.

PR #2257, opened 2015-05-22
Adding E2E Tests for SqlServer.Design

lajones:
See #1669.
Adding FileService to abstract outputting files so can override for tests.
Ensuring .cshtml resources have same name when built using VS as when built using command line.
Fixing some error messages.


test/EntityFramework.SqlServer.FunctionalTests/SqlServerTestStore.cs
```diff
@@ -83,7 +83,7 @@ private SqlServerTestStore CreateShared(Action initializeDatabase)
             return this;
         }
 
-        public static async Task CreateDatabaseIfNotExistsAsync(string name, string scriptPath = null)
+        public static async Task CreateDatabaseAsync(string name, string scriptPath = null, bool recreateIfAlreadyExists = false)
```
lajones: The previous functionality was only replacing the database if it did not already exist. But in iterating over this during testing I found I needed to replace it multiple times. This also sets us up for the future so that if the DB has to change again (which it probably will as we find more test cases) that can be done easily.

That said, if recreateIfAlreadyExists is false (the default) then the functionality is identical to what was there before.

test/EntityFramework.SqlServer.Design.FunctionalTests/ReverseEngineering/E2ETests.cs
```diff
@@ -0,0 +1,150 @@
...
+            var expectedFileContents = InitializeExpectedFileContents();
+
+            var generator = new ReverseEngineeringGenerator(serviceProvider);
+            var filePaths = generator.GenerateAsync(configuration).Result;
+
+            Assert.Equal(_E2ETestExpectedWarnings.Count, logger.WarningMessages.Count);
+            // loop over warnings instead of using the collection form of Assert.Equal()
+            // to give better error messages if it does fail. Similarly for file paths below.
+            var i = 0;
+            foreach (var expectedWarning in _E2ETestExpectedWarnings)
```
lajones: I'm in two minds about this. The collection form of the Assert is easy to use. But I found that if it fails you're not likely to get back useful output about why it is failing - more than likely the actual error will be buried somewhere in the '...' which is output by this to avoid outputting long lines for long collections. So instead have chosen to loop over each warning asserting individually. That way if something goes wrong the output will say at exactly which offset it went wrong. Similarly for the file paths below.

src/EntityFramework.Relational.Design/Templating/Compilation/MetadataReferencesProvider.cs
```diff
@@ -43,6 +43,8 @@ public virtual List<MetadataReference> GetApplicationReferences()
 
         private void InitializeReferences()
         {
+            _isInitialized = true;
```
lajones: Moved to prevent a stack overflow with AddReferenceFromName() calling InitializeReferences() repeatedly.

lajones:
@bricelam @divega - ping - could I get a review please?

test/EntityFramework.SqlServer.FunctionalTests/SqlServerTestStore.cs
bricelam: Could you avoid recreating by using cancelled transactions in the tests?

bricelam:
:shipit:

test/EntityFramework.SqlServer.FunctionalTests/SqlServerTestStore.cs
lajones: After discussion am leaving this approach as is to allow for future changes. It's unlikely that this test's schema will stay the same in the near term.

lajones:
Checked in with commit a787a61

lajones:
Fixed in release branch and only on dnx451. Had problems with dnxcore50 - see #2352.

PR #2260, opened 2015-05-22
:art: Move default constraint naming to relational metadata extensions

bricelam:
Note, the constraint names are now based on the on the domain model, not the database artifacts.


ajcvickers:
:shipit:

PR #2265, opened 2015-05-24
React to aspnet/Configuration#194

dougbu:
- may not completely unblock build but moves it much further along


dougbu:
FYI only. Will push immediately.

dougbu:
131e2df

Issue #2090, opened 2015-04-29
Calling AddSqlServer more than once shouldn't result in an error.


matthewDDennis:
I am getting the following error when I first use the ContentTaxonomyDbContext.

> An exception of type 'System.InvalidOperationException' occurred in EntityFramework.Core.dll but was not handled in user code
>
> Additional information: The data stores 'SqlServerDataStore' 'SqlServerDataStore' are configured. A context can only be configured to use a single data store.

I have the following method 
```c#
        public static IServiceCollection AddSqlTaxonomyLookups(this IServiceCollection services, string connectionString)
        {
            var entityFrameworkServicesBuilder = new EntityFrameworkServicesBuilder(services);
            entityFrameworkServicesBuilder
                                          .AddSqlServer() // no error if this is commented out
                                          .AddDbContext<ContentTaxonomyDbContext>(options =>
                options.UseSqlServer(connectionString));

            AddLookupServices(services);

            return services;
        }
```
that I call at the end of the Startup.ConfigureServices

```c#
        // This method gets called by the runtime. Use this method to add services to the container.
        public void ConfigureServices(IServiceCollection services)
        {
            // Add Application settings to the services container.
            services.Configure<AppSettings>(Configuration.GetSubKey("AppSettings"));

            // Add EF services to the services container.
            services.AddEntityFramework()
                .AddSqlServer()
                .AddDbContext<CodeProjectIdentityDbContext>(options =>
                    options.UseSqlServer(Configuration["Data:Identity:ConnectionString"]));
             ...
            // Add our services
            services.AddSqlTaxonomyLookups(Configuration["Data:Content:ConnectionString"]);
        }
```

matthewDDennis:
Additionally, I think that calling any *services.AddXXX* extension method multiple times should not be an issue.

In a modular design, a Module might want to call *services.AddEntityFramework(...)* to ensure that it is registered, without knowing if any other code has called it previously.

mikary:
We looked into resolving this exception in #2271, but after some discussion decided it was more appropriate to throw on repeated calls to AddXXX for providers. PR #2330 updates the AddXXX call to throw earlier with a more descriptive error message.

matthewDDennis:
you could just use TryAdd.

-----Original Message-----

mikary:
[...]

PR #2271, opened 2015-05-26
Select distinct data store services to allow multiple calls to AddProvider

mikary:
Fix for #2090, throwing when multiple registrations of the same ```IDataStoreSource``` implementation exist.
Two possible solutions to the issue are:
* Ensuring the class is only added once - This is non-trivial because there are multiple ```IDataStoreSource```s expected in the cross-store scenario and we either need to track if the provider-specific implementation has been added or query the container to see if a registration exists
* Ignoring duplicates when selecting the ```IDataStoreSource``` in the ```DataStoreSelector```

This PR uses the second approach, filtering in the ```DataStoreSelector```
@rowanmiller

### DIFF
--- a/src/EntityFramework.Core/Storage/DataStoreSelector.cs
+++ b/src/EntityFramework.Core/Storage/DataStoreSelector.cs
@@ -27,7 +27,10 @@ namespace Microsoft.Data.Entity.Storage
 
             _serviceProvider = serviceProvider;
             _contextOptions = contextOptions;
-            _sources = sources == null ? new IDataStoreSource[0] : sources.ToArray();
+            _sources = sources == null
+                ? new IDataStoreSource[0]
+                : sources.Distinct((source, storeSource) => source.GetType() == storeSource.GetType())
+                    .ToArray();
         }
 
         public virtual IDataStoreServices SelectDataStore(ServiceProviderSource providerSource)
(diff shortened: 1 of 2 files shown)


ajcvickers:
This is fine as in as far as it goes, but in general calling AddXxx methods multiple times will cause problems whenever those methods register a service where multiple services are supposed to be registered for the given type. For example, calling AddEntityFramework more than once will still cause issues because the listeners will be registered more than once.

So, if we're saying that calling the AddXxx methods more than once is okay, then we should have a general solution to make that work, otherwise we should say that calling AddXxx must only be done once.

divega:
@ajcvickers Is it the case that this fix takes care of the `AddXyzProvider()` methods but we would need a different approach for `AddEntityFramework()`?

ajcvickers:
@divega I believe that is the case for our AddXxxProvider methods as they currently stand, but it may not remain the case depending on what gets added, and it may not be the case for other providers that we don't own.

mikary:
We have at least a few options for establishing a pattern for AddXxxProvider. The solution in this PR does feel a bit like a band-aid on a bigger problem. The first solution (either inspecting the container as we build or having some static member to track if a call has been made) seems like a more complete solution

mikary:
closing to open PR with detect and throw on add call.

PR #2273, opened 2015-05-26
Rename DbContextOptions to EntityOptions and update all related naming.

ajcvickers:


anpete:
:shipit:

lajones:
Looks like you missed SqlServerDbContextTemplate.cshtml template in SqlServer.Design which also has a reference to DbContextOptionsBuilder? If you like I can check this in with the E2E test code (which is out for review - see #2257)?

ajcvickers:
@lajones That file has been updated--not sure why you are not seeing it.

lajones:
@ajcvickers Yep - I see it now - though I swear I searched for it 3 times before I sent the comment. Ah, well.

PR #2277, opened 2015-05-26
More SQLite testing

natemcmaster:
Depends on https://github.com/aspnet/Microsoft.Data.Sqlite/pull/84 for the transaction tests to work correctly.

TODO: DbConnection is not always disposed correctly--> test *.db files are left behind.

### DIFF
--- a/test/EntityFramework.Sqlite.FunctionalTests/SqliteTestStore.cs
+++ b/test/EntityFramework.Sqlite.FunctionalTests/SqliteTestStore.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Data.Common;
+using System.Diagnostics;
 using System.IO;
 using System.Threading;
 using Microsoft.Data.Entity.Relational.FunctionalTests;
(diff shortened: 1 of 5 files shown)


anpete:
:ship: :it:

test/EntityFramework.Sqlite.FunctionalTests/SqliteTestStore.cs
```diff
@@ -88,29 +102,30 @@ private DbCommand CreateCommand(string commandText, object[] parameters)
...
 
-        public static string CreateConnectionString(string name) =>
+        public static string CreateConnectionString(string name, bool sharedCache = false) =>
             new SqliteConnectionStringBuilder
-                {
-                    DataSource = name + ".db"
-                }
+            {
+                DataSource = name + ".db",
+                SharedCacheMode = sharedCache
```
bricelam: Any reason not to just always use a shared cache?

test/EntityFramework.Sqlite.FunctionalTests/SqliteTestStore.cs
```diff
@@ -17,8 +18,17 @@ public class SqliteTestStore : RelationalTestStore
...
-        public static SqliteTestStore CreateScratch() => 
-            new SqliteTestStore("scratch-" + Interlocked.Increment(ref _scratchCount)).CreateTransient();
+        public static SqliteTestStore CreateScratch(bool sharedCache = false)
+        {
+            string name;
+            do
+            {
+                name = "scratch-" + Interlocked.Increment(ref _scratchCount);
+            }
+            while (File.Exists(name + ".db"));
```
bricelam: With this, you can re-add the built-in types tests.

bricelam:
:shipit:

test/EntityFramework.Sqlite.FunctionalTests/SqliteTestStore.cs
natemcmaster: I can only think of one reason: it is not the default SQLite behavior.

PR #2279, opened 2015-05-27
Query: Don't throw when no orderby present when skip operation.

### LINKED ISSUE #1607: Skip requires OrderBy for DefaultSqlQueryGenerator
https://github.com/aspnet/EntityFramework/blob/bf6360e53f623ffb602bd89a6d678bd85ec8094a/src/EntityFramework.Relational/Query/Sql/DefaultSqlQueryGenerator.cs#L291

Skip and Take should not require OrderBy by default, as only a single product (MS SQL Server) has this limitation. It's simply a wrong place to put that check in.

anpete:
- Append ORDER BY 1 on SQL Server.

Fix #1607

@divega Updated with ORDER BY 1

### DIFF
--- a/src/EntityFramework.Core/Query/CompiledQueryCache.cs
+++ b/src/EntityFramework.Core/Query/CompiledQueryCache.cs
@@ -234,7 +234,7 @@ namespace Microsoft.Data.Entity.Query
 
                         var parameterValue = Evaluate(e, out parameterName);
 
-                        var compilerPrefixIndex = parameterName.LastIndexOf(">");
+                        var compilerPrefixIndex = parameterName.LastIndexOf(">", StringComparison.Ordinal);
                         if (compilerPrefixIndex != -1)
                         {
                             parameterName = parameterName.Substring(compilerPrefixIndex + 1);
--- a/src/EntityFramework.Relational/Properties/Strings.Designer.cs
+++ b/src/EntityFramework.Relational/Properties/Strings.Designer.cs
@@ -109,14 +109,6 @@ namespace Microsoft.Data.Entity.Relational
         }
 
         /// <summary>
-        /// A query containing the Skip operator must include at least one OrderBy operation.
-        /// </summary>
-        public static string SkipNeedsOrderBy
-        {
-            get { return GetString("SkipNeedsOrderBy"); }
-        }
-
-        /// <summary>
         /// The SQL Server sequence '{sequenceName}' was already specified with a different definition.
         /// </summary>
         public static string SequenceDefinitionMismatch([CanBeNull] object sequenceName)
@@ -341,14 +333,6 @@ namespace Microsoft.Data.Entity.Relational
         }
 
         /// <summary>
-        /// The required column '{column}' was not present in the results of a 'FromSql' operation.
-        /// </summary>
-        public static string FromSqlMissingColumn([CanBeNull] object column)
-        {
-            return string.Format(CultureInfo.CurrentCulture, GetString("FromSqlMissingColumn", "column"), column);
-        }
-
-        /// <summary>
         /// Generating down script for migration '{migration}'.
         /// </summary>
         public static string GeneratingDown([CanBeNull] object migration)
@@ -386,6 +370,14 @@ namespace Microsoft.Data.Entity.Relational
         public static string UsingConnection([CanBeNull] object database, [CanBeNull] object dataSource)
         {
             return string.Format(CultureInfo.CurrentCulture, GetString("UsingConnection", "database", "dataSource"), database, dataSource);
+        }
+
+        /// <summary>
+        /// The required column '{column}' was not present in the results of a 'FromSql' operation.
+        /// </summary>
+        public static string FromSqlMissingColumn([CanBeNull] object column)
+        {
+            return string.Format(CultureInfo.CurrentCulture, GetString("FromSqlMissingColumn", "column"), column);
         }
 
         private static string GetString(string name, params string[] formatterNames)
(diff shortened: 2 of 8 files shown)


divega:
Can we ORDER BY on PKs if they are available instead of NEWID?

anpete:
We could, but I'm not sure introducing a strong ordering is the right thing with respect to the semantics of the query.

divega:
I strongly believe it will match expectations and be more useful more often than not: DbSets will return ordered lists on the keys if no filters or explicit ordering is applied because of clustered indexes. Filtering using LINQ to Objects would always preserve ordering but a relational implementation won't because it can swithch to another index, however as soon as you apply a paging operator over a query that doesn't contain explicit ordering you will expect it to behave as a sequence, i.e. to preserve the "natural" order of the keys. Most important you expect the order to be deterministic and not random. Otherwise you cannot actually use it for paging.

anpete:
I'm not sure I agree with all of your assumptions :-)

> DbSets will return ordered lists on the keys if no filters or explicit ordering is applied because of clustered indexes.

Doesn't this depend on the underlying store? I don't think one can assume a deterministic ordering of DbSets. Also, bear in mind that there are non-relational stores. For example, results streaming from an in-memory store have a non-deterministic order.

> Filtering using LINQ to Objects would always preserve ordering but a relational implementation won't because it can swithch to another index, however as soon as you apply a paging operator over a query that doesn't contain explicit ordering you will expect it to behave as a sequence, i.e. to preserve the "natural" order of the keys.

What about this L2O query: myDictionary.Skip(5)

> Most important you expect the order to be deterministic and not random. Otherwise you cannot actually use it for paging.

That is not my expectation, at least not for DbSets, for the reasons listed above.

divega:
> I'm not sure I agree with all of your assumptions :-)

Sorry if the way I was trying to explain was too terse an chaotic. I wrote the previous comment on my phone and this maybe better to talk about in person. FWIW this happens to be a long standing feature gap between LINQ to SQL and EF which has been discussed many times before.

>> DbSets will return ordered lists on the keys if no filters or explicit ordering is applied because of clustered indexes.

> Doesn't this depend on the underlying store? 

It does, but the key point is that the idea that DbSets are "stored" (i.e. can be retrieved) in the "natural" order of the PKs is already part of the mental model for many customers. I was just mentioning one of the reasons for that.

> For example, results streaming from an in-memory store have a non-deterministic order

In most cases rows will be retrieved in some order (unless the stores intentionally randomize the output as you propose). The canonical case for LINQ is an in-memory IEnumerable which has an order. LINQ query operators on IEnumerable are order preserving unless they need to not be. 

So I am saying is that we recognize the idea that the data is stored ordered by PK.

I am not saying that we should now append an implicit OrderBy() operation on every query that doesn't have one, but that if an order dependent query operator is applied on a query that isn't ordered in many cases we do have a default order to fall back to: PK ordering.

>> Most important you expect the order to be deterministic and not random. Otherwise you cannot actually use it for paging.

> That is not my expectation, at least not for DbSets, for the reasons listed above.

I am also asserting that falling back to PK ordering is much more useful 
than random ordering. If I write a view with paging using something like `context.Products.Skip((page-1) * pageSize).Take(pageSize)` I really do hope that as I skip to different pages I will see a distinct page of products, not just a random sample of size pageSize. If I go back to a previous page, I expect to see the same products I saw before.

> What about this L2O query: myDictionary.Skip(5)

Can you explain? Are you saying that DbSets should mimic the behavior of an IDictionary instead of IEnumerable?

anpete:
> > What about this L2O query: myDictionary.Skip(5)
> Can you explain? Are you saying that DbSets should mimic the behavior of an IDictionary instead of IEnumerable?

No, what I was getting at is that you seem to be saying that L2O queries are deterministic and I was providing a counter example. 

Agree we may need to chat in person, but here are a few more data points:

> the key point is that the idea that DbSets are "stored" (i.e. can be retrieved) in the "natural" order of the PKs is already part of the mental model for many customers.

I have no evidence to either confirm or deny this. What I can say is (and I confirmed with @ajcvickers) is that DbSet's have no defined ordering semantics and it is erroneous to assume they do. DbSet ordering is non-deterministic and store dependent.

> In most cases rows will be retrieved in some order (unless the stores intentionally randomize the output as you propose). The canonical case for LINQ is an in-memory IEnumerable which has an order. LINQ query operators on IEnumerable are order preserving unless they need to not be.

When I said "in-memory" I should have been clearer - I mean using our in-memory provider. Result ordering is non-deterministic with that provider because it uses dictionaries internally.

> So I am saying is that we recognize the idea that the data is stored ordered by PK.

As above, I don't think we can assume that.

> I am also asserting that falling back to PK ordering is much more useful than random ordering

Perhaps, but it feels like we are making an app level decision here.

divega:
Ok, let's talk about this more in depth in person tomorrow then, but for now:

I believe you and I maybe referring to two different "levels" or "flavors" of determinism:

1. If I enumerate the results twice in the same application without making changes on the store I will get things in the same order both times
2. I know what the order is based on and I can repro what the order is easily.

Strictly speaking only (1) is required to have sane paging and I believe storing things on a dictionary in the InMemory store does exhibit (1), so no need to do anything for InMemory. On the other hand SQL Server just does not let us run certain queries without an explicit ORDER BY so unless we are ok with these failing we need to pick what key to sort on.

Now, appending ORDER BY NEWID() breaks both (1) and (2). ORDER BY on the keys (when available) achieves (1). Now (2) is a bonus because of all the possible orders this - I claim - is the most expected one.  

> DbSet's have no defined ordering semantics and it is erroneous to assume they do

Sure, but I am not talking in terms of mathematical correctness. I am asking for the behavior to be useful and predictable, and I believe we achieve that by introducing the key ordering when necessary.

divega:
@anpete you still have time to replace 1 with your favorite number (assuming it isn't 1) or 0 to save trees :smile:

:shipit:

Thanks @bricelam

bricelam:
Would `NULL` save even more trees? :palm_tree: :deciduous_tree: :evergreen_tree:

heraclesaraujo:
Let me point about PK ordering, what if my store uses random PK, NEWID in SQL Server our I let MongoDB generate a ObjectID at runtime? won't PK ordering by default break that "natural order" of items? maybe assuming no ordering is not an EF responsability

PR #2288, opened 2015-05-29
Commands: Use startup project to discover services

### LINKED ISSUE #2024: Flow IServiceProvider to ContextTool
src/EntityFramework.Commands/ContextTool.cs needs the ISP to avoid relying on CallContextServiceLocator for Hosting

### LINKED ISSUE #2219: Use start-up project for  ConfigureServices
I am trying to set up a project where the business logic (including data model) is separated from UI/web. The context has only a ctor that takes only `DbContextOptions` because I want to use `InMemoryStore` for tests and SqlServer when actually running the app. I am trying to add a migration from the "engine" project and provide the web project as the start up project since this is where I configure EF to use SqlServer (in the `ConfigureServices` method). This is on Beta4 bits.

`dnx . ef migration add initial -s StarTrader`

```
Information: [LoaderContainer]: Load name=EntityFramework.Relational
Information: [NuGetAssemblyLoader]: Loaded name=EntityFramework.Relational in 3ms
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Data.Entity.Commands.Program.GetProjectPath(String projectName)
   at Microsoft.Data.Entity.Commands.Program.ExecuteInDirectory(String startupProject, Func`1 invoke)
   at Microsoft.Data.Entity.Commands.Program.AddMigration(String name, String context, String startupProject)
   at Microsoft.Data.Entity.Commands.Program.<>c__DisplayClass7_1.<Main>b__6()
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.Data.Entity.Commands.Program.Main(String[] args)
```

There is a recent [change](#f3749bd69672d1bcb0e2d977f5509471b0faf4d9) that touched this code so it might be already fixed in beta5

bricelam:
Resolves #2219 & resolves #2024

Depends on aspnet/Hosting#284

Verification: Manual bashing

### DIFF
--- a/src/EntityFramework.Commands/ContextTool.cs
+++ b/src/EntityFramework.Commands/ContextTool.cs
@@ -10,40 +10,50 @@ using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 
 #if DNX451 || DNXCORE50
-using System.Threading.Tasks;
 using Microsoft.AspNet.Hosting;
-using Microsoft.Framework.Runtime.Infrastructure;
 #endif
 
 namespace Microsoft.Data.Entity.Commands
 {
     public class ContextTool
     {
-        public static DbContext CreateContext([NotNull] Type type)
+        private readonly IServiceProvider _services;
+
+        public ContextTool([CanBeNull] IServiceProvider services)
+        {
+            _services = services;
+        }
+
+        public virtual DbContext CreateContext([NotNull] Type type, [CanBeNull] string startupAssemblyName)
         {
             Check.NotNull(type, nameof(type));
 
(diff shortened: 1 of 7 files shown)


ajcvickers:
:shipit:

Just a thought. When AddDbContext is called it will register the context in D.I. as both its specific type and as just DbContext. So if the AddDbContext mechanism is being used and only a single context type is being used, then the context instance (and its type) could be obtained from the service provider even without first knowing or discovering it. This might be advantageous for cases where the context is in a different assembly. But then again, it might be best just to take a more general approach to that scenario such that we always use the exact type name of the context class.

PR #2299, opened 2015-05-29
Fix #2052 - Register IMemoryCache as singleton.

### LINKED ISSUE #2052: EntityFramework adds IMemoryCache as transient which breaks it totally
EF is registering `IMemoryCache` as transient in DI, meaning it's kinda pointless. Other frameworks, e.g. MVC, do a `TryAdd` so they don't replace it. Our default templates add EF services before any others, so this is breaking anything else that relies on `IMemoryCache`, e.g. the `CacheTagHelper`.
  https://github.com/aspnet/EntityFramework/blob/128da35a60e7e54f901ecfaaff6f29678a7ccd43/src/EntityFramework.Core/Extensions/EntityFrameworkServiceCollectionExtensions.cs#L110

anpete:


lajones:
:shipit:

PR #2300, opened 2015-05-29
Update relational type mapping

ajcvickers:
This change updates the relational type mapping code such that:
- Specifying store types (e.g. nvarchar(128)) are recognized and parameters are created appropriately
- Specifying MaxLength is recognized and used in Migrations and in parameters
- DbType is only specified where needed--for example for DateTime2 on SQL Server

The update pipeline and Migrations use the type mapping. Query does not use it yet--it just creates ad-hoc parameters)

SQL Server has been fully tested. SQLite is not yet tested.

Using provider-specific type enums is not supported--we'll do that if we find a case that needs it.

Type conversions are not yet supported--the ADO.NET provider has to be able to work with your property type and your store type without going through an intermediate CLR type.


anpete:
:shipit:

src/EntityFramework.Relational/RelationalTypeMapper.cs
```diff
@@ -2,101 +2,209 @@
...
+            var typeName = property.Relational().ColumnType;
+            if (typeName != null)
             {
-                return mapping.Item2;
+                return GetOrAddNameMapping(typeName.ToLowerInvariant());
             }
 
-            if (propertyType.GetTypeInfo().IsEnum)
+            RelationalTypeMapping mapping;
+            return SimpleMappings.TryGetValue(property.ClrType.UnwrapEnumType().UnwrapNullableType(), out mapping)
```
bricelam: Should this be `.UnwrapNullableType().UnwrapEnumType()` to work with nullable enums?

src/EntityFramework.Relational/RelationalTypeMapper.cs
```diff
@@ -2,101 +2,209 @@
...
+        protected virtual RelationalTypeMapping MapFromName(
+            [NotNull] string typeName,
+            [NotNull] string typeNamePrefix,
+            int? firstQualifier,
+            int? secondQualifier)
+        {
+            Check.NotEmpty(typeName, nameof(typeName));
+            Check.NotEmpty(typeNamePrefix, nameof(typeNamePrefix));
+
+            throw new NotSupportedException("Unrecognized/supported store type.");
```
bricelam: Localize?

src/EntityFramework.Relational/RelationalTypeMapper.cs
ajcvickers: @smitpatel fixed UnwrapEnumType so it works either way now.

src/EntityFramework.Relational/RelationalTypeMapper.cs
ajcvickers: Oops. Good catch!

src/EntityFramework.Sqlite/SqliteTypeMapper.cs
```diff
@@ -2,60 +2,63 @@
...
-            string specifiedType,
-            string storageName,
-            Type propertyType,
-            bool isKey,
-            bool isConcurrencyToken)
+        private readonly RelationalTypeMapping _integer = new RelationalTypeMapping("INTEGER");
+        private readonly RelationalTypeMapping _real = new RelationalTypeMapping("REAL");
+        private readonly RelationalTypeMapping _blob = new RelationalTypeMapping("BLOB");
+        private readonly RelationalTypeMapping _text = new RelationalTypeMapping("TEXT");
+        private readonly RelationalTypeMapping _decimal = new RelationalScaledTypeMapping("TEXT", 18, 2);
```
bricelam: Hmm... Why bother setting precision and scale?

ajcvickers: I'll change it to never set it if SQLite will always ignore it.

src/EntityFramework.SqlServer/SqlServerTypeMapper.cs
```diff
@@ -2,84 +2,156 @@
...
-                if (isKey)
-                {
-                    return _keyByteArrayMapping;
-                }
+            return TryMapSized(typeNamePrefix, new[] { "nvarchar", "national char varying", "national character varying" }, firstQualifier)
+                   ?? TryMapSized(typeNamePrefix, new[] { "varbinary", "binary varying" }, firstQualifier, DbType.Binary)
+                   ?? TryMapSized(typeNamePrefix, new[] { "varchar", "char varying", "character varying" }, firstQualifier, DbType.AnsiString)
+                   ?? TryMapScaled(typeNamePrefix, new[] { "decimal", "numeric", "dec" }, firstQualifier, secondQualifier)
+                   ?? TryMapScaled(typeNamePrefix, new[] { "float", "double precision" }, firstQualifier, secondQualifier)
+                   ?? TryMapScaled(typeNamePrefix, new[] { "datetime2" }, firstQualifier, secondQualifier, DbType.DateTime2)
```
bricelam: `datetime2` only supports precision, not scale.

test/EntityFramework.Relational.Tests/Model/RelationalTypeMappingTest.cs
```diff
@@ -16,67 +11,198 @@ namespace Microsoft.Data.Entity.Relational.Tests.Model
...
+            var parameter = new RelationalScaledTypeMapping("decimal", DbType.Decimal, 10, 6)
+                .CreateParameter(new TestCommand(), "Name", 9.99m, isNullable: false);
+
+            Assert.Equal(ParameterDirection.Input, parameter.Direction);
+            Assert.Equal("Name", parameter.ParameterName);
+            Assert.Equal(9.99m, parameter.Value);
+            Assert.Equal(DbType.Decimal, parameter.DbType);
+            Assert.False(parameter.IsNullable);
+            //Assert.Equal(10, parameter.Scale);
+            //Assert.Equal(6, parameter.Precision);
```
bricelam: Should these be commented out? (below too)

src/EntityFramework.SqlServer/SqlServerTypeMapper.cs
ajcvickers: This code works correctly for only precision or for precision and scale. In the negative case that you get it wrong, creating the database with Migrations will fail. I decided not to add an additional check here because it seems unlikely checking the negative case here will ever be of real use.

test/EntityFramework.Relational.Tests/Model/RelationalTypeMappingTest.cs
bricelam: You could add an extension method to cast to IDbParameter on net45.

src/EntityFramework.Relational/RelationalTypeMapper.cs
bricelam: Thanks, @smitpatel

src/EntityFramework.Sqlite/SqliteTypeMapper.cs
bricelam: Yep, just like max lengths on strings, precision and scale are also ignored. :smile: There are [some TODOs](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite/SqliteParameter.cs#L13-L15) to light up more on SqliteParameter, but most things are ignored by SQLite in the name of dynamic typing.

src/EntityFramework.SqlServer/SqlServerTypeMapper.cs
bricelam: Sounds good.

bricelam:
:shipit:

PR #2301, opened 2015-05-29
Prevent Stack Overflow in Reverse Engnerring plus adding E2E Tests

lajones:
@divega @muratg @danroth27 
This was reviewed just after the release branch got created - so it didn't quite make it in time. See https://github.com/aspnet/EntityFramework/pull/2257 review based on dev.
The vital part of this is the change to src/EntityFramework.Relational.Design/Templating/Compilation/MetadataReferencesProvider.cs  - without it we get a stack overflow when doing Reverse Engineering.

The rest is adding an E2E test so that we catch this next time.


divega:
@lajones is this still open for a reason? Didn't we merge it already?

lajones:
Yes it's merged - I just ended up opening too many pull requests about this. Closing - see #2257.

Issue #2168, opened 2015-05-12
AddScoped vs AddSingleton


ErikEJ:
.AddSingleton<ISqlStatementExecutor, SqlStatementExecutor>()
      .AddSingleton<ISqlServerCeTypeMapper, SqlServerCeTypeMapper>()
      .AddSingleton<ISqlServerCommandBatchPreparer, SqlServerCommandBatchPreparer>()

The corresponding ones in the SQLite provider are added as Scoped - why?

ajcvickers:
@ErikEJ Services should be added as scoped if:
* They depend on any services/data that is coupled to the current session--e.g. the context itself, the state manager, etc.
* They depend on any services that are scoped
  * This includes depending on any provider-specific services referenced through a non-provider specific contract interface, since these are all registered as scoped

Services need to be added as singletons if they contain caches that should persist beyond the lifetime of a single context instance.

It looks like SqlStatementExecutor isn't currently a service that can have a provider specific implementation, but it looks like we should probably change this so that providers can override if needed--does CE need its own implemenation? It is registered as Singleton because the only thing it depends on is logging, which is not scoped.

It looks like SqliteTypeMapper should really be registered as singleton because it maintains re-used mappings--a cache essentially. I am currently working in this area and the interface will change anyway. I'll make sure to update the registration.

CommandBatchPeparers are registered as scoped for SQLite and SQL Server, and this seems correct because there are dependencies on other scoped services.

ErikEJ:
@ajcvickers Thanks Arthur. 
Yes, that confirms my (limited) understanding.

I had a typo in my question, which should have said:
The corresponding ones in the SQLite provider are added as **Scoped** - why?

I will add all three as singletons in the CE provider, and I do not need a SqlStatementExecutor override.

ErikEJ:
Merged

PR #2308, opened 2015-06-01
Make service scoping consistent between SQLite and SQL Server providers

ajcvickers:
This change makes the SqliteTypeMapper Singleton so that the type mappings are cached, which is consistent with what SQL Server does.

Also, ISqlServerSqlGenerator is made into a server that can be overidden by providers if needed.

See Issue #2168


bricelam:
:shipit:

PR #2311, opened 2015-06-02
Make two Transaction tests overrideable

### LINKED ISSUE #2304: Error in TransactionsTest
In test Query_uses_explicit_transaction - getting error due to unsupported isolation level: Result Message:	System.ArgumentException : The IsolationLevel value ReadUncommitted is invalid.

https://msdn.microsoft.com/en-us/library/system.data.sqlserverce.sqlcetransaction.isolationlevel(v=vs.100).aspx

What would be the best way to work around this?

ErikEJ:
Fixes #2304 - some ADO.NET providers do not support ReadUncommited isolation level


natemcmaster:
LGTM

PR #2313, opened 2015-06-02
Add functional tests for SQLite type mappings

ajcvickers:
SQLite type mapping is extremely simple so this is mostly enabling the provider-agnostic tests for SQLite--which all work, even the ones that don't work on SQL Server!--and then adding SQLite specific tests that make sure that nothing barfs.

Note that size, precision, and scale are all ignored by SQLite, so tests for these just show that things still round-trip without issue.

### DIFF
--- a/test/EntityFramework.Sqlite.FunctionalTests/BuiltInDataTypesSqliteFixture.cs
+++ b/test/EntityFramework.Sqlite.FunctionalTests/BuiltInDataTypesSqliteFixture.cs
@@ -1,0 +1,140 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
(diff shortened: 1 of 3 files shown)


ErikEJ:
@ajvickers I would like to default string to nvarchar(4000), but would also like to be able to map ntext (similar to nvarchar(max) - any suggestion on how to do that with the latest type mapping api incarnation?

ajcvickers:
@ErikEJ In your _simpleMappings make the entry for "string" point to a `RelationalSizedTypeMapping("nvarchar(4000)", 4000)` and then in your _simpleNameMappings add an entry for ntext that goes to `RelationalTypeMapping("ntext")`. If the app doesn't specify any column type, then it will get an nvarchar(4000), but if the column type is set to ntext, then that will be used.

ErikEJ:
Great,  just missing the second half. Thanks!

bricelam:
Awesome. :shipit:

PR #2322, opened 2015-06-03
dotnet all the things

bricelam:
Depends on aspnet/Caching#69, aspnet/Logging#189 & aspnet/Microsoft.Data.Sqlite#87


lajones:
:shipit:

PR #2323, opened 2015-06-03
[SMALL] Adding tests for the #2069 - Order by Guid.NewGuid() returns same order every time

maumar:
Also porting OrderDetail data from Northwind so that the results match our baselines.

### DIFF
--- a/test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
+++ b/test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
@@ -2987,95 +2987,92 @@ namespace Microsoft.Data.Entity.FunctionalTests
             return "m";
         }
 
-        //[Fact]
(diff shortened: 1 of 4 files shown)


maumar:
@anpete

anpete:
@maumar Reminder, I think you need to do something so that you show up as "cla-not-requred".

test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs
```diff
@@ -3535,6 +3540,14 @@ protected QueryTestBase(TFixture fixture)
         {
             using (var context = CreateContext())
             {
+
+                var foo = NorthwindData.Set<TItem>().Count();
```
anpete: ??

anpete:
:shipit:

Issue #2316, opened 2015-06-03
RelationalLogger logs differently on different cultures


ErikEJ:
In particular parameter values, which makes some Asserts brittle - PR in the works, that avoids dependency on Mirosoft.AspNet.Testing

ErikEJ:
Merged

PR #2328, opened 2015-06-04
Parameter values should be logged using InvariantCulture

ErikEJ:
Fixes issue #2316


test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -2389,7 +2389,7 @@ public override void Where_chain()
             base.Where_chain();
 
             Assert.Equal(
-                @"@__p_0: 1/1/1998 12:00:00 AM
+                @"@__p_0: 01/01/1998 00:00:00
```
MichaelKetting: Hmm...I think you're missing the Sqlite test there.

ErikEJ: @MichaelKetting  I think you are right! :-)

bricelam:
:smiling_imp: Such a rebel...

PR #2330, opened 2015-06-04
Add exception for repeated call to add provider methods

mikary:
Fix for #2090, alternate solution to #2271 
@rowanmiller @ajcvickers 

@bricelam - Could you verify that the SQLite changes are reasonable?


ajcvickers:
:shipit: Consider pulling the boiler-plate code into a common place exposed as an "internal" utility method in core, Since we are going to do it in all our providers, other providers may want to reuse our method. But I'm also okay leaving as if if you think its better.

test/EntityFramework.Sqlite.Tests/SqliteTestHelpers.cs
```diff
@@ -0,0 +1,29 @@
...
+        public new static SqliteTestHelpers Instance { get; } = new SqliteTestHelpers();
+
+        public override EntityFrameworkServicesBuilder AddProviderServices(EntityFrameworkServicesBuilder builder)
+        {
+            return builder.AddSqlite();
+        }
+
+        protected override void UseProviderOptions(EntityOptionsBuilder optionsBuilder)
+        {
+            optionsBuilder.UseSqlite(new SqlConnection("Database=DummyDatabase"));
```
bricelam: `Database=:memory:` would keep it more isolated

bricelam: Not that it should ever be used...

bricelam:
Sqlite stuff looks fine

mikary:
Updated based on feedback

mikary:
Closing to create new PR as the intent of this PR has changed.

PR #2346, opened 2015-06-08
Fix for E2E RevEng test

lajones:
Part of issue #2257.
Recent check-in a787a61 seems to have caused a problem interpreting the DbContext template. This fix adds back in the @ symbols before if and foreach in part of the template.


bricelam:
:shipit:

lajones:
Thanks v much.

Cheers,

laj

PR #2352, opened 2015-06-09
Not working on dnxcore50 - references problem during compilation of templates

lajones:
Removing support for dnxcore50 until we can fix.

@Eilon @divega @rowanmiller @bricelam

### DIFF
--- a/src/EntityFramework.SqlServer.Design/project.json
+++ b/src/EntityFramework.SqlServer.Design/project.json
@@ -15,6 +15,6 @@
   "resource": "ReverseEngineering/Templates/*.cshtml",
   "frameworks": {
     "net45": { },
-    "dnxcore50": { }
+    "dnx451": { }
   }
 }
--- a/src/EntityFramework.SqlServer.Design/project.json
+++ b/src/EntityFramework.SqlServer.Design/project.json
@@ -14,7 +14,6 @@
   },
   "resource": "ReverseEngineering/Templates/*.cshtml",
   "frameworks": {
-    "net45": { },
-    "dnx451": { }
+    "net45": { }
   }
 }
--- a/test/EntityFramework.SqlServer.Design.FunctionalTests/project.json
+++ b/test/EntityFramework.SqlServer.Design.FunctionalTests/project.json
@@ -10,7 +10,6 @@
     "test": "xunit.runner.aspnet"
   },
   "frameworks": {
-    "dnx451": { },
-    "dnxcore50": { }
+    "dnx451": { }
   }
 }


src/EntityFramework.SqlServer.Design/project.json
```diff
@@ -15,6 +15,6 @@
   "resource": "ReverseEngineering/Templates/*.cshtml",
   "frameworks": {
     "net45": { },
-    "dnxcore50": { }
+    "dnx451": { }
```
Eilon: Hmm so why is `dnx451` being *added* here?

lajones: @bricelam It got removed during the clean-up that was done around all of the changes to dotnet etc. It appears to work even if you just specify net45. Personally I'd rather be explicit about the fact that we support dnx451 as well but if Brice says it's OK not to have that framework in there then that's OK by me too.

Eilon: If it's not needed I don't think I'd put it there... @bricelam thoughts?

bricelam: It's not needed. I think we should avoid cross-compiling to platforms when we don't need too. It just make the nupkg files bigger than they need to be.

test/EntityFramework.SqlServer.Design.FunctionalTests/project.json
```diff
@@ -10,7 +10,6 @@
     "test": "xunit.runner.aspnet"
   },
   "frameworks": {
-    "dnx451": { },
-    "dnxcore50": { }
+    "dnx451": { }
```
bricelam: There doesn't seem to be any reason these tests *shouldn't* be able to run on `dnxcore50`. Is there a test fix we can make instead?

src/EntityFramework.SqlServer.Design/project.json
lajones: @Eilon @bricelam  OK. I've added another commit - to just use net45. Tests pass either way.

src/EntityFramework.SqlServer.Design/project.json
```diff
@@ -14,7 +14,6 @@
   },
   "resource": "ReverseEngineering/Templates/*.cshtml",
   "frameworks": {
-    "net45": { },
-    "dnxcore50": { }
+    "net45": { }
```
bricelam: Why doesn't this work on dnxcore50? We need to be able to reverse engineer on CoreCLR...

lajones: It's a problem with the references that get passed to the compile step of the templates. The references look correct but the compiler complains that it doesn't have a "BinaryOperation" member that  it's expecting. I'm working on it. But want to get this checked in so we can have a working release branch without muting this test while I figure out what's going wrong.

bricelam:
:shipit: Sorry, I didn't have enough context. Unblocking for now is good.

lajones:
Fixed with commit 1d973a1. Test running successfully on dnx451.

PR #2354, opened 2015-06-09
Add validation for distinct table names

### LINKED ISSUE #2170: A better error message when the a duplicate table name is used would be an improvement
Right now if you accidentally set two entities to point to the same table, adding a migration indicates:

```
modelBuilder.Entity<Something>( c => c.Table( "Something" ) );
modelBuilder.Entity<Sommat>( c => c.Table( "Something" ) );
```

> PM> Add-Migration Test
> An item with the same key has already been added.

This is not very descriptive of what is wrong.  A better error might at least indicate the key to help identify what the duplication is.  Better still would be to indicate the issue has to do with table names.

Stack trace from running with -Verbose:

> System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Microsoft.Data.Entity.Relational.Migrations.Infrastructure.ModelDiffer.Sort(IEnumerable`1 operations)
   at Microsoft.Data.Entity.Relational.Migrations.Infrastructure.ModelDiffer.GetDifferences(IModel source, IModel target)
   at Microsoft.Data.Entity.Commands.Migrations.MigrationScaffolder.ScaffoldMigration(String migrationName, String rootNamespace)
   at Microsoft.Data.Entity.Commands.MigrationTool.AddMigration(String migrationName, String contextTypeName, String rootNamespace, String projectDir)
   at Microsoft.Data.Entity.Commands.Executor.<AddMigrationImpl>d__8.MoveNext()
   at Microsoft.Data.Entity.Commands.Executor.OperationBase.<>c__DisplayClass6_0`1.<Execute>b__0()
   at Microsoft.Data.Entity.Commands.Executor.OperationBase.Execute(Action action)

EF 7.0.0-beta5-13171

smitpatel:
fixes #2170

### DIFF
--- a/src/EntityFramework.Relational/RelationalModelValidator.cs
+++ b/src/EntityFramework.Relational/RelationalModelValidator.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.Data.Entity.Relational
+{
+    public class RelationalModelValidator : LoggingModelValidator
+    {
+        public RelationalModelValidator([NotNull] ILoggerFactory loggerFactory)
+            : base(loggerFactory)
+        {
+        }
+
+        public override void Validate(IModel model)
+        {
+            base.Validate(model);
+            EnsureDistinctTableNames(model);
+        }
+
+        protected void EnsureDistinctTableNames(IModel model)
+        {
+            var tables = new HashSet<string>();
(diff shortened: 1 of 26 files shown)


ajcvickers:
Good job! :shipit:

src/EntityFramework.Relational/RelationalModelValidator.cs
```diff
@@ -0,0 +1,40 @@
...
+            EnsureDistinctTableNames(model);
+        }
+
+        protected void EnsureDistinctTableNames(IModel model)
+        {
+            var tables = new HashSet<string>();
+            foreach (var entityType in model.EntityTypes.Where(et => et.BaseType == null))
+            {
+                var tableName = entityType.Relational().Schema + "." + entityType.Relational().Table;
+                if (tables.Contains(tableName))
```
ilmax: This check is redundant, this could be simplified with if (!tables.Add(trableName))

PR #2355, opened 2015-06-09
SQLite: Add monster fixup test and fix getters in query complication

natemcmaster:
More tests for SQLite and a minor fix to query compilation.

To make MonsterFixup tests work on SQLite, the model must use MessageId, PhotoId, and ReviewId as the key for the model.

Almost completes #2121

### DIFF
--- a/test/EntityFramework.Sqlite.FunctionalTests/MonsterFixupSqliteTest.cs
+++ b/test/EntityFramework.Sqlite.FunctionalTests/MonsterFixupSqliteTest.cs
@@ -1,0 +1,83 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Data.Entity.ChangeTracking.Internal;
+using Microsoft.Data.Entity.FunctionalTests;
+using Microsoft.Data.Entity.FunctionalTests.TestModels;
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Utilities;
+using Microsoft.Data.Sqlite;
+using Microsoft.Framework.DependencyInjection;
+
+namespace Microsoft.Data.Entity.Sqlite.FunctionalTests
+{
+    public class MonsterFixupSqliteTest : MonsterFixupTestBase
+    {
+        private static readonly HashSet<string> _createdDatabases = new HashSet<string>();
+
+        private static readonly ConcurrentDictionary<string, AsyncLock> _creationLocks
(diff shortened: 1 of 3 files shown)


test/EntityFramework.Sqlite.FunctionalTests/MonsterFixupSqliteTest.cs
```diff
@@ -0,0 +1,83 @@
...
+                }
+            }
+        }
+
+        public override void OnModelCreating<TMessage, TProductPhoto, TProductReview>(ModelBuilder builder)
+        {
+            base.OnModelCreating<TMessage, TProductPhoto, TProductReview>(builder);
+            builder.Entity<TMessage>().Key("MessageId");
+            builder.Entity<TProductPhoto>().Key("PhotoId");
+            builder.Entity<TProductReview>().Key("ReviewId");
```
ajcvickers: Is this because SQLite doesn't support composite keys? Also, why using the string overload instead of the lambda overload?

natemcmaster: It's because SQLite only supports server value generation for one column per table.

And we use the string overload because of the generic constraints on TMessage in MonsterFixupTestBase.cs

ajcvickers: Hmmm. I didn't realize the model required server value generation on more than column--it didn't when it was originally created because it used SQL Server Identity, which can only be on one column as well. It must have morphed into this at some point, either intentionally or not. It's not a huge deal, but we might just want to make it consistent for all providers.

It used to be the case that the lambda overload could be done with a cast to the interface type. It was a bit ugly, but at least still allowed automatic refactoring, etc. For example, `builder.Entity<TMessage>().Key(e => ((IMessage)e).MessageId);`

ajcvickers:
:shipit: (Either as-is or consistent across providers)

test/EntityFramework.Sqlite.FunctionalTests/MonsterFixupSqliteTest.cs
natemcmaster: The tests do not require generation for multiple columns. But because SQLite only supports generation for one column AND only if that column is the primary key.

Also, added added a few lines so we use generic constraints better ==> no need to use string overload. Will squash before merging.

natemcmaster: Update: got ahead of myself. There was a good reason to use the string overload. This test setup does not agree well with Change Tracking.

```
The instance of entity type 'Microsoft.Data.Entity.FunctionalTests.TestModels.ChangedOnlyMonsterContext+ProductReview' cannot be tracked because another instance of this type with the same key is already being tracked. For new entities consider using an IIdentityGenerator to generate unique key values.
```

Issue #2246, opened 2015-05-21
Assigning boolean in Select statement throws invalid cast.


MikeWasson:
I have a DbContext like this:

    public class BookContext : DbContext
    {
        public DbSet<Author> Authors { get; set; }
     }

This code throws an invalid cast

    var x = BookContext.Authors
        .Select(x => new
        {
            Text = x.FirstMidName,
            Value = x.AuthorID.ToString(),
            Selected = true  // Throws invalid cast
        }).ToList();

Works fine without 'Selected = true'. 

Stack trace:

```
System.InvalidCastException was unhandled by user code
  HResult=-2147467262
  Message=Specified cast is not valid.
  Source=EntityFramework.Relational
  StackTrace:
       at Microsoft.Data.Entity.Relational.RelationalObjectArrayValueReader.ReadValue[T](Int32 index)
       at lambda_method(Closure , QuerySourceScope )
       at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
       at Microsoft.Data.Entity.Query.LinqOperatorProvider.ExceptionInterceptor`1.EnumeratorExceptionInterceptor.MoveNext()
       at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
       at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
       at ContosoBooks.Controllers.BookController.GetAuthorsListItems(Int32 selected) in C:\Users\mwasson\Documents\Visual Studio 2015\Projects\sample\src\ContosoBooks\Controllers\BookController.cs:line 143
       at ContosoBooks.Controllers.BookController.Create() in C:\Users\mwasson\Documents\Visual Studio 2015\Projects\sample\src\ContosoBooks\Controllers\BookController.cs:line 45
  InnerException: 
```

maumar:
Problem here is that we are sending the following query to the database:

```
SELECT [x].[FirstMidName], [x].[AuthorID], 1
FROM [Author] AS [x]
```

This then gets stored in the value buffer as {string, int, int}, rather than {string, int, bool} - hence the conversion fails, when we try to cast from int to bool.

Possible solution is to evaluate the constant on the client, rather than sending it to server. We need to make sure that we still translate constants that are bound to the source query somehow (part of a more complex expression involving a column, or an argument to a function)

divega:
Wouldn't 'CAST (1 AS BIT)" work?

maumar:
[...]

PR #2366, opened 2015-06-11
[SMALL] Fix to #2246 - Assigning boolean in Select statement throws invalid cast

maumar:
Problem is that a row projection containing a constant bool gets translated into integer (0 or 1) since 1 and 0 are the literal bit values for true and false. When we try to materialize, we try to cast integer 0 or 1 into bool, which causes an error.
Fix is to not send standalone constants to the store, but instead deal with them on the client. This is consistent with our approach to functions (funcletize as much as we can)

### DIFF
--- a/test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -773,6 +773,56 @@ FROM [Customers] AS [c]",
                 Sql);
         }
 
+        public override void Select_anonymous_one()
+        {
+            base.Select_anonymous_one();
+
+            Assert.Equal(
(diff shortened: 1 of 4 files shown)


test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -788,7 +838,7 @@ public override void Select_constant_null_string()
             base.Select_constant_null_string();
 
             Assert.Equal(
-                @"SELECT NULL
+                @"SELECT 1
```
maumar: is this fine, or should we only do client side for the mulitple property projections?

maumar:
@anpete

anpete:
:shipit:

PR #2367, opened 2015-06-11
Updates to relational type mapping

ajcvickers:
This change uses default type mappings in query and adds code to set Size appropriately on string and binary parameters for SQL Server.


divega:
:shipit: 

Looks good to me. May make sense to have @anpete check on the query pipeline changes.

ajcvickers:
Merged. Happy to make any changes based on feedback from @anpete if necessary.

PR #2370, opened 2015-06-11
Disable shadow state entity

### LINKED ISSUE #748: Disable use of shadow state entities except for model snapshot
Shadow state entities (that is entities without any backing CLR types) are not fully supported in many places; most notably query. We should fail early if code tries to use them. However, the Migrations model snapshot does make use of them, so this should be disabled.

smitpatel:
resolves #748


divega:
:shipit:

PR #2371, opened 2015-06-12
Improving SQLite migration support

natemcmaster:
Throws NotSupportedException for all migration operations that cannot be expressed in a single SQL operation or that cannot be executed. 

TODO: Implement table rebuilds to imitate the behavior of some unsupported operations.


src/EntityFramework.Sqlite/Migrations/SqliteHistoryRepository.cs
```diff
@@ -3,51 +3,134 @@
...
+
+            _connection = connection;
+            _contextKey = context.GetType().FullName;
+            _sql = sql;
+        }
+
         public virtual string BeginIfExists(string migrationId)
         {
-            throw new NotImplementedException();
+            throw new NotSupportedException(Strings.MigrationScriptGenerationNotSupported);
```
bricelam: This should say that generating *idempotent* scripts isn't supported. You can still generated scripts.

bricelam: @ErikEJ hit a similar limitation in the SQL Server Compact provider. I'll look for a better way we can factor this whole class.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
```diff
@@ -57,14 +59,104 @@ public override void Generate(RenameTableOperation operation, IModel model, SqlB
...
         {
-            throw new NotImplementedException();
+            throw new NotSupportedException(Strings.InvalidMigrationOperation);
+        }
+
+        #endregion
+
+        #region Invalid schema operations
+
+        // EF.SQLite does not support ATTACH/DETACH DATABASE and therefore not support creating schema
```
bricelam: `ATTACH DATABASE` technically defines a catalog name, not a schema. But the lines are pretty blurry here--especially on Oracle.

src/EntityFramework.Sqlite/Properties/Strings.resx
```diff
@@ -0,0 +1,135 @@
...
+    <value>Cannot add column to table '{tableName}' with non-constant default '{defaultValue}'.</value>
+  </data>
+  <data name="InvalidMigrationOperation" xml:space="preserve">
+    <value>SQLite cannot support this migration operation.</value>
+  </data>
+  <data name="MigrationScriptGenerationNotSupported" xml:space="preserve">
+    <value>Generating scripts for migration is not supported by SQLite.</value>
+  </data>
+  <data name="MultipleDatabasesNotSupported" xml:space="preserve">
+    <value>Operations involving multiple databases in SQLite is not currently supported.</value>
```
bricelam: As above, I think this should say SQLite doesn't support schemas. Multiple attached databases is arguably different.

src/EntityFramework.Sqlite/Migrations/SqliteHistoryRepository.cs
natemcmaster: This probably belongs in a separate discussion thread, but one way we could work around this limitation is to (1) produce a separate file for each migration containing only that migration's SQL statements relevant and (2) generate a Python or LUA (or something else) script that conditionally executes each migration's statements.

src/EntityFramework.Sqlite/Properties/Strings.resx
```diff
@@ -0,0 +1,135 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="AddColumnWithDefaultExprNotSupported" xml:space="preserve">
+    <value>Cannot add column to table '{tableName}' with non-constant default '{defaultValue}'.</value>
```
bricelam: I wonder if we should mention something like "This operation requires a table rebuild." in the appropriate messages.

src/EntityFramework.Sqlite/Properties/Strings.resx
```diff
@@ -0,0 +1,135 @@
...
+  </data>
+  <data name="InvalidMigrationOperation" xml:space="preserve">
+    <value>SQLite cannot support this migration operation.</value>
+  </data>
+  <data name="MigrationScriptGenerationNotSupported" xml:space="preserve">
+    <value>Generating scripts for migration is not supported by SQLite.</value>
```
bricelam: (as above) should clarify that this is just *idempotent* scripts.

src/EntityFramework.Sqlite/Properties/Strings.resx
natemcmaster: Agreed. I'll update the PR to say this instead.

src/EntityFramework.Sqlite/SqliteSqlGenerator.cs
```diff
@@ -10,6 +10,8 @@ namespace Microsoft.Data.Entity.Sqlite
 {
     public class SqliteSqlGenerator : SqlGenerator
     {
+        public override string DelimitIdentifier(string tableName, string schemaName) => base.DelimitIdentifier(tableName);
```
bricelam: Hmm, I think it's an error to ever call this with a schema name. Ignoring this might be hiding other errors. Asking for the SQLite-specific schema name in metadata will always return null. If callers are getting the schema some other way, it's wrong.

bricelam: Oh I see, users can bypass metadata using migration operations. It might be interesting to log a warning here.

test/EntityFramework.Sqlite.Tests/Metadata/SqliteHistoryRepositoryTest.cs
```diff
@@ -0,0 +1,159 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Data.Entity.Relational;
+using Microsoft.Data.Entity.Relational.Migrations.History;
+using Microsoft.Data.Entity.Relational.Migrations.Operations;
+using Microsoft.Data.Entity.Sqlite.Migrations;
+using Microsoft.Data.Sqlite;
+using Moq;
```
bricelam: We generally discourage the use of Moq.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
```diff
@@ -22,6 +22,18 @@ public class SqliteMigrationSqlGenerator : MigrationSqlGenerator
             _sql = sqlGenerator;
         }
 
+        public override void Generate(AddColumnOperation operation, IModel model, SqlBatchBuilder builder)
+        {
+            Check.NotNull(operation, nameof(operation));
+
+            if (operation.DefaultExpression != null
+                && operation.DefaultExpression.Contains("CURRENT_"))
```
bricelam: Why only if it contains "CURRENT_"?

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
```diff
@@ -22,6 +22,18 @@ public class SqliteMigrationSqlGenerator : MigrationSqlGenerator
...
+            {
+                throw new NotSupportedException(Strings.AddColumnWithDefaultExprNotSupported(operation.Table, operation.DefaultExpression));
```
bricelam: In general, we should only throw if it would result in invalid SQL. Otherwise, we should let the database throw.

src/EntityFramework.Sqlite/SqliteSqlGenerator.cs
natemcmaster: Yup, this needed changing because of migrations sql generation. I can add a log warning...sidenote: have we establish logging conventions for EF?

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
natemcmaster: Good point. This should be BeginsWith

natemcmaster: More detail: CURRENT_TIMESTAMP, CURRENT_DATE and CURRENT_TIME are not supported when adding a column. These can only be set as the default when creating a table.

src/EntityFramework.Sqlite/SqliteSqlGenerator.cs
bricelam: No, we still need to sort out logging conventions. For now, a TODO would suffice

test/EntityFramework.Sqlite.Tests/Metadata/SqliteHistoryRepositoryTest.cs
natemcmaster: What do we use instead?

bricelam:
:shipit:

PR #2376, opened 2015-06-12
Make IDataStoreCreators get the model through DI

ajcvickers:
We were making choices about which methods to pass the context model to based on heuristics about when it might be needed. However, it was always possible to get the model in any of the methods. This change cleans up the code and allows the model to be used in any method for which it is useful. This potentially makes it more likely that the semantics of deleting a database or checking for tables will be inconsistent, but as discussed in a recent API review the best thing some providers can do is to use the current model for things like deletion.


divega:
:shipit:

Issue #2175, opened 2015-05-12
Swap back to identity by default (rather than sequence)


rowanmiller:
Adding this so that we can discuss in triage and make a decision (I didn't see any other issues tracking the need to make a decision here).

PR #2382, opened 2015-06-15
Swap back to Identity as the default for SQL Server

ajcvickers:
Also remove all SQL Server-specific ways to change whether a property is considered StoreGneratedPattern.Identity or not. This is a core level concept that should be handled by the core.

Issue #2175.


divega:
:shipit:

ErikEJ:
@ajcvickers This is great! Saved me reams of code on the SQLCE provider

Issue #2176, opened 2015-05-12
Remove KeyValueEntityTracker


rowanmiller:
When discussing https://github.com/aspnet/EntityFramework/issues/1916 we decided to remove this helper class. It's a very simple line of code and it makes it more obvious to the developer how simplistic the logic of this approach is... and easier for them to expand the logic to suit their domain model.

PR #2384, opened 2015-06-15
Remove KeyValueEntityTracker

ajcvickers:
Force apps to write their own logic so that logic is explicitly part of the app and cannot be misconstrued to do something else.

Issue #2176


divega:
:shipit:

PR #2386, opened 2015-06-15
Rename EntityOptions back to DbContextOptions

ajcvickers:
Because:
- We decided to stay with DbContext and not EntityContext at the top
- We decided that we didn't want to have another "prefix" word--just use Db
- These are the options to use for a given DbContext instance anyway


natemcmaster:
:shipit:

ErikEJ:
NOOOOOO! 😃

PR #2390, opened 2015-06-16
React to Roslyn API change

troydai:
error CS0619: 'MetadataReference.CreateFromAssembly(Assembly)' is
obsolete: 'Use CreateFromFile(assembly.Location) instead'


davidfowl:
:shipit:

bricelam:
Oh good! The whole "Don't use `CreateFromAssembly` for in-memory assemblies" was rather confusing.

PR #2392, opened 2015-06-16
Remove GenerateValueOnAdd from fluent API

ajcvickers:
The problem with this is that it is really an internal detail which should never be switched off even when explicit key values are being used. Occasionally it cab be useful to switch it on for non-key, non-discriminator properties, but this would not be a common case and dropping down to core metadata is fine for this.

Also, removed it from the model snapshot since it really has no impact on store generation--StoreGeneratedPattern is for that. Removing it from the snapshot makes diffs easier to handle.

Also, did a name change that got missed when the reat of the metadata API review changes were applied.

### DIFF
--- a/test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
+++ b/test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
@@ -70,7 +70,6 @@ namespace Microsoft.Data.Entity.Commands.Migrations
 builder.Entity(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+EntityWithOneProperty"", b =>
     {
         b.Property<int>(""Id"")
-            .GenerateValueOnAdd()
             .StoreGeneratedPattern(StoreGeneratedPattern.Identity);
 
         b.Key(""Id"");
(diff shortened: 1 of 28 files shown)


test/EntityFramework.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -451,29 +435,27 @@ public void Property_maxLength_is_stored_in_snapshot()
         }
 
         [Fact]
-        public void Property_generateValueOnAdd_is_stored_in_snapshot()
+        public void Property_generateValueOnAdd_is_nont_stored_in_snapshot()
```
smitpatel: typo. Also name should be changed according to rename.

ajcvickers: @smitpatel Will fix. Are you going to finish reviewing this or should  somebody else look as well?

smitpatel: Looks good to me. Though I suggest somebody else should look as well.

ajcvickers: I'll merge now. Happy to respond to any other feedback as we move forward.

Issue #1102, opened 2014-11-19
Reorganize ModelBuilderTest


AndriySvyryd:
There are many scenarios that need to be covered for ModelBuilder making the ModelBuilderTest class hard to navigate when looking for coverage gaps, there is also a lot of duplicate tests in BasicModelBuilderTest. Depending on the outcome of #908 these classes need to be reorganized into a collection of suites corresponding to the targeted scenarios and whether they depend on the default set of conventions.

Consider creating a ModelBuilderTests subfolder to group the resulting collection.

ajcvickers:
Clearing milestone. What I didn't realize when we triaged this was that the tests for the non-generic and string-based APIs are currently disabled meaning we have very little coverage for those APIs.

divega:
@smitpatel this is related to #908. Please talk to @AndriySvyryd about how to go about it.

PR #2393, opened 2015-06-16
Refactor ModelBuilder tests and add more coverage.

AndriySvyryd:
Remove redundant CollectionReferenceBuilder

Part of #1102


src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
```diff
@@ -80,11 +80,16 @@ public virtual EntityTypeBuilder Annotation([NotNull] string annotation, [NotNul
...
         public virtual KeyBuilder Key([NotNull] params string[] propertyNames)
         {
-            Check.NotNull(propertyNames, nameof(propertyNames));
+            Check.NotEmpty(propertyNames, nameof(propertyNames));
 
             return new KeyBuilder(Builder.PrimaryKey(propertyNames, ConfigurationSource.Explicit));
         }
 
+        /// <summary>
+        ///     Creates a new unique index for this entity type if one does not already exist over the specified properties.
```
ajcvickers: Nope! This is specifcially NOT about creating a unique "index". :-) This is for cases where you want an unique constraint on a column, but don't want an index, and don't want it to be used at one end or the other of a relationship. (Or, put differently, you might want those things, in which case you will need to make some other call anyway and this call will be redundant.)

src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
```diff
@@ -80,11 +80,16 @@ public virtual EntityTypeBuilder Annotation([NotNull] string annotation, [NotNul
...
+        /// </summary>
+        /// <param name="propertyNames"> The names of the properties that make up the unique index. </param>
```
ajcvickers: Not an index.

src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
```diff
@@ -80,11 +80,16 @@ public virtual EntityTypeBuilder Annotation([NotNull] string annotation, [NotNul
...
+        /// <returns> An object that can be used to configure the unique index. </returns>
```
ajcvickers: Not an index.

src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder`.cs
```diff
@@ -74,6 +74,19 @@ public virtual KeyBuilder Key([NotNull] Expression<Func<TEntity, object>> keyExp
             return new KeyBuilder(Builder.PrimaryKey(keyExpression.GetPropertyAccessList(), ConfigurationSource.Explicit));
         }
 
+        /// <summary>
+        ///     Creates a new unique index for this entity type if one does not already exist over the specified properties.
```
ajcvickers: Not an index,

test/EntityFramework.Core.Tests/ModelBuilderGenericTest.cs
```diff
@@ -0,0 +1,337 @@
...
+        {
+            public GenericTestEntityTypeBuilder(EntityTypeBuilder<TEntity> entityTypeBuilder)
+            {
+                EntityTypeBuilder = entityTypeBuilder;
+            }
+
+            private EntityTypeBuilder<TEntity> EntityTypeBuilder { get; }
+            public override EntityType Metadata => EntityTypeBuilder.Metadata;
+
+            public override TestEntityTypeBuilder<TEntity> Annotation(string annotation, object value)
```
ajcvickers: Consider expression body syntax for lots of these methods.

ajcvickers:
:shipit:

src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
AndriySvyryd: We call it "unique index" in the .PrincipalKey docs. Should that be changed to "unique constraint" as well?

ajcvickers: Probably, yes.

PR #1958, opened 2015-03-31
Made RelationalLoggerExtensions public

roji:
While extending ReaderModificationCommandBatch and overring Execute(), it was impossible to use
RelationalLoggerExtensions.LogCommand().

PR #2396, opened 2015-06-17
Made Logger extensions public

ErikEJ:
Fixes #1958


rowanmiller:
@ErikEJ per comment on the original request from @roji 
> given that we are making this class public can you make sure we check parameters for null on all public methods and move it to the  Microsoft.Data.Entity.Relational  namespace.

ErikEJ:
@rowanmiller  Yes, I saw that, where is that not the case in this PR - I looked but could not find any place...?

ErikEJ:
@anpete @rowanmiller PR updated

rowanmiller:
@ErikEJ yep that's what we meant. @anpete can you process this one now.

ErikEJ:
@anpete PR updated with latest changes in dev

PR #2399, opened 2015-06-17
Query: Relational: First round of subquery lifting. 

anpete:
Lift subqueries in MainFromClauses.

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionVisitors/MemberAccessBindingExpressionVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionVisitors/MemberAccessBindingExpressionVisitor.cs
@@ -19,6 +19,8 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors
     {
         private readonly EntityQueryModelVisitor _queryModelVisitor;
         private readonly bool _inProjection;
+        
+        private bool _inMember;
 
         public MemberAccessBindingExpressionVisitor(
             [NotNull] QuerySourceMapping querySourceMapping,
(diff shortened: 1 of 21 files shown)


anpete:
@maumar @mikary

src/EntityFramework.Core/Query/ExpressionVisitors/MemberAccessBindingExpressionVisitor.cs
```diff
@@ -103,9 +116,13 @@ protected override Expression VisitMethodCall(MethodCallExpression methodCallExp
                     methodCallExpression.Method.GetGenericMethodDefinition(),
                     QueryExtensions.PropertyMethodInfo))
             {
+                _inMember = true;
```
maumar: will that work for nested method calls, or should we use stack instead?

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -82,14 +82,18 @@ public virtual bool HandlesQuerySource([NotNull] IQuerySource querySource)
         {
             Check.NotNull(querySource, nameof(querySource));
 
-            return _tables.Any(tableExpression => tableExpression.QuerySource == querySource);
+            return _tables.Any(tableExpression
+                => tableExpression.QuerySource == querySource
```
maumar: te for consistency

anpete: Fixed.

src/EntityFramework.Core/Query/ExpressionVisitors/MemberAccessBindingExpressionVisitor.cs
anpete: I believe it will work. Nested member expressions can't contain additional QSREs and we only do it for Property method calls.

maumar:
:shipit:

Issue #2071, opened 2015-04-26
OrderBy() not translating to SQL ORDER BY?


roji:
I'm trying to make the `FunctionalTests.QueryTestBase.Take_Distinct()` test pass on PostgreSQL, and am running into some issues - am probably misunderstanding something.

This test runs the query `os.OrderBy(o => o.OrderID).Take(5).Distinct()`. The `SelectExpression` that gets generated from this has `_orderBy` empty, and the SQL that gets generated from that accordingly has no `ORDER BY` clause:

```sql
SELECT DISTINCT "o"."OrderID", "o"."CustomerID", "o"."EmployeeID", "o"."OrderDate" FROM "Orders" AS "o" LIMIT 5
```
(note that in PostgreSQL we have `LIMIT 5`, in SqlServer it's probably `TOP 5`)

Since there is no `ORDER BY` clause, the actual elements returned by this query aren't well-defined, and therefore it seems impossible to assert on them (my test fails)... What am I missing here?

rowanmiller:
@maumar can you help out with this one

anpete:
Bug. Poaching

roji:
Test passes for me, closing.

PR #2401, opened 2015-06-17
Fix: #2071 - OrderBy() not translating to SQL ORDER BY?

anpete:
- Improve translation of Distinct such that OrderBys are preserved when the ordering columns are present in the projection.


anpete:
@maumar @mikary

maumar:
:shipit:

PR #2411, opened 2015-06-18
Break up ModelBuilderTest into smaller files

### LINKED ISSUE #1102: Reorganize ModelBuilderTest

AndriySvyryd:
Add tests for non-generic relationship API

Fixes #1102


ajcvickers:
:shipit:

PR #2417, opened 2015-06-18
Initial database facade refactoring.

ajcvickers:
The initial change removes the relational and provider-specific facade classes in favor of extension methods. All the other refactoring renaming will build on this change. Added RelationalConnection extension method until all the transaction stuff is changed to flow directly off .Database.


anpete:
:shipit:

PR #2422, opened 2015-06-19
Fix #783 - replace Property extension method on object in fluent API

### LINKED ISSUE #783: Property extension method shows up in fluent API and can be confusing
The extension method 
```
public static TProperty Property<TProperty>([NotNull] this object entity, [NotNull] string propertyName)
```
is in the Microsoft.Data.Entity namespace and defined on object, which means is shows up all over the place--as is, I believe, the intention. However, having it show up in the fluent API where there are other Property methods can be quite confusing. Likewise for the DbContext API in the future.

mikary:
A new Ef helper class as an alternative to ```object.Property<T>()``` in the fluent API
@divega @rowanmiller

### DIFF
--- a/src/EntityFramework.Core/Query/EF.cs
+++ b/src/EntityFramework.Core/Query/EF.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
+using System;
+
+namespace Microsoft.Data.Entity.Query
+{
+    public static class EF
+    {
+        public static TProperty Property<TProperty>([NotNull] object entity, [NotNull] string propertyName)
+        {
+            throw new InvalidOperationException(Strings.PropertyMethodInvoked);
+        }
+    }
+}
+
--- a/src/EntityFramework.Core/EntityFramework.Core.csproj
+++ b/src/EntityFramework.Core/EntityFramework.Core.csproj
@@ -183,6 +183,7 @@
     <Compile Include="ModelBuilder.cs" />
     <Compile Include="Query\Annotations\QueryAnnotation.cs" />
     <Compile Include="Query\Annotations\IncludeQueryAnnotation.cs" />
+    <Compile Include="Query\EF.cs" />
     <Compile Include="Query\EntityLoadInfo.cs" />
     <Compile Include="Query\EntityTrackingInfo.cs" />
     <Compile Include="Query\ExpressionEvaluationHelpers.cs" />
--- a/src/EntityFramework.Core/Properties/Strings.Designer.cs
+++ b/src/EntityFramework.Core/Properties/Strings.Designer.cs
@@ -341,11 +341,11 @@ namespace Microsoft.Data.Entity.Internal
         }
 
         /// <summary>
-        /// The Property&lt;T&gt; extension method may only be used within LINQ queries.
+        /// The EF.Property&lt;T&gt; method may only be used within LINQ queries.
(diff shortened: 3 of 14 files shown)


rowanmiller:
@divega and I discussed a while back and I *think* we concluded that we would be better off sticking with DbFunctions from EF6 (even though this isn't technically a function in the database). We should discuss as a team though.

ajcvickers:
@rowanmiller @divega I'm happy with DbFunctions.

src/EntityFramework.Core/Infrastructure/EfMethodInfo.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Data.Entity.Query;
+using System.Reflection;
+
+namespace Microsoft.Data.Entity.Infrastructure
+{
+    public static class EfMethodInfo
```
ajcvickers: Note sure this needs its own class. Why is it being moved from where it was before?

src/EntityFramework.Core/Query/Ef.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
+using System;
+
+namespace Microsoft.Data.Entity.Query
+{
+    public static class Ef
```
ajcvickers: Slight preference for "EF" rather than "Ef" if we go with this, but I'm also happy with DbFunctions or pretty much any other name.

src/EntityFramework.Core/Infrastructure/EfMethodInfo.cs
mikary: It could remain on ```QueryExtensions``` since it is a query extension related method. Still seems a bit awkward either way since it needs to be available outside of core but shouldn't show up when dotting on ```Ef```

ajcvickers:
:shipit: if we can agree on the name.

src/EntityFramework.Core/Infrastructure/EfMethodInfo.cs
anpete: It could be moved to EntityQueryModelVisitor.

src/EntityFramework.Core/Query/Ef.cs
anpete: Agree, prefer EF.

anpete:
Not a big fan of DbFunctions because to me it seems counterintuitive.

Another nice thing about a single "EF" entry point is that we could do:

```
EF.DbFunctions
```

Which means that extension methods can be added like we do elsewhere.

src/EntityFramework.Core/Infrastructure/EfMethodInfo.cs
mikary: Added #2456 to track this to get the other changes in more quickly

src/EntityFramework.Core/Query/EF.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
+using System;
+
+namespace Microsoft.Data.Entity.Query
```
divega: Wrong namespace. Should be `Microsoft.Data.Entity` for discoverability.

PR #2430, opened 2015-06-19
Use extension method to clean up IAccessor use

ajcvickers:
This change uses a couple of extension methods to allow access to underlying services without having to cast to IAccessor<TService> every time. It means that this kind of low-level access is available and discoverable whenever the .Infrastructure namespace is imported, but is otherwise still hidden in normal apps.

Top level objects (DbContext, DbSet, Database) that expose IAccessor<IServiceProvider> allow use of the second extension method overload to automatically go through the service provider indirection and obtain any service registerd in DI.


bricelam:
:shipit: :+1: Much better!

src/EntityFramework.Core/Metadata/Builders/ReferenceNavigationBuilder.cs
```diff
@@ -154,6 +154,6 @@ protected virtual InternalRelationshipBuilder InverseReferenceBuilder([CanBeNull
             return builder;
         }
 
-        private InternalRelationshipBuilder Builder => ((IAccessor<InternalRelationshipBuilder>)this).Service;
+        private InternalRelationshipBuilder Builder => this.GetService();
```
divega: I assume in this case you didn't need to specify the type because `ReferenceNavigationBuilder` only implements `IAccessor<TService>` once. It might just be me, but this seems to make the code opaque.

divega: A bit OT but this also makes me wonder again about the name of the `IAccessor<TService>` interface (in fact it has been bothering me for a while). Wouldn't it be better to call it `IServiceAccessor<TService>`?

ajcvickers: We should discuss in API review. We could consider renaming or having bespoke extension methods for different services.

divega: Sounds good. From my perspective it would be enough if we specify the generic argument even when it is redundant to improve readability. Of course there isn't a way to enforce that...

ajcvickers: We'd also have to make sure ReShaper doesn't just remove it on cleanup.

divega: Ah, good point.

PR #2431, opened 2015-06-19
Migrations: Remove provider-specific model differs

bricelam:
Instead of override the ModelDiffer, providers instead tell us which annotations they care about. The ModelDiffer then compares them and copies them to any resulting operation.


ajcvickers:
:shipit:

PR #2433, opened 2015-06-20
Implementation of DropColumn for SQLite using migration transforms.

natemcmaster:
This is a first-cut of implementing a migration operation that requires table-rebuild.

Deliberate design choices:
 - Modifying the operation is provider specific and happens just before SQL generation
 - Migration API remains unchanged
 - Relies on the model to determine which columns are used in the table-rebuild.

Future improvements:
 - Add support for the many migration operations not available to SQLite
 - Allow users to opt-in for potentially-destructive table-rebuilds, instead of running them by-default.
 - Better logging/warning users of what is going to happen.

### DIFF
--- a/src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
+++ b/src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
@@ -1,0 +1,83 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Relational.Migrations.Infrastructure;
+using Microsoft.Data.Entity.Relational.Migrations.Operations;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Sqlite.Migrations
+{
+    public class SqliteOperationTransformer
+    {
+        private readonly IModelDiffer _differ;
+
+        public SqliteOperationTransformer([NotNull] IModelDiffer differ)
+        {
+            Check.NotNull(differ, nameof(differ));
+
+            _differ = differ;
+        }
+
+        protected IList<MigrationOperation> TransformOperation(MigrationOperation operation, IModel model) => 
+            new[] { operation };
+
+        // TODO prevent multiple DropColumnOperations on the same table
+        private IList<MigrationOperation> TransformOperation(DropColumnOperation operation, IModel model)
+        {
+            Check.NotNull(model, nameof(model));
+
+            // TODO ensure this temporary table does not conflict with an existing table
+            var tempTableName = operation.Table + "_temp";
+
+            // TODO find more efficient way to get a create table operation. Expose ModelDiffer.Add?
+            var createTableOperation = _differ.GetDifferences(null, model)
+                .FirstOrDefault(o => (o as CreateTableOperation)?.Name == operation.Table);
+
+            if (!(createTableOperation is CreateTableOperation))
+            {
+                return new MigrationOperation[] { operation };
+            }
+
+            return new[]
+            {
+                new RenameTableOperation
+                {
+                    Name = operation.Table,
+                    NewName = tempTableName
+                },
+                createTableOperation,
+                new MoveDataOperation
+                {
+                    OldTable = tempTableName,
+                    NewTable = operation.Table,
+                    Columns = ((CreateTableOperation)createTableOperation)
+                        .Columns
+                        .Select(c => c.Name)
+                        .ToArray()
+                },
+                new DropTableOperation
+                {
+                    Name = tempTableName
+                }
+            };
+        }
+
+        public virtual IReadOnlyList<MigrationOperation> Transform(
+            [NotNull] IReadOnlyList<MigrationOperation> operations,
+            [CanBeNull] IModel model = null)
+        {
+            Check.NotNull(operations, nameof(operations));
+
+            var finalOperations = new List<MigrationOperation>();
+            foreach (var operation in operations)
+            {
+                finalOperations.AddRange(TransformOperation((dynamic)operation, model));
+            }
+            return finalOperations.AsReadOnly();
+        }
+    }
+}
--- a/test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
+++ b/test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
@@ -1,0 +1,160 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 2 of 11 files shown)


bricelam:
Looks good, but I think we should flesh the feature out a bit more before merging into dev.

natemcmaster:
Just added some more thorough testing. Also, these changes are local to SQLite only. 

As we discussed, we will build support for table-rebuilds first in SQLite, and afterwards find table-rebuild abstraction for other providers.

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
```diff
@@ -0,0 +1,82 @@
...
+        // TODO prevent multiple DropColumnOperations on the same table
+        private IList<MigrationOperation> TransformOperation(DropColumnOperation operation, IModel model)
+        {
+            Check.NotNull(model, nameof(model));
+
+            // TODO ensure this temporary table does not conflict with an existing table
+            var tempTableName = operation.Table + "_temp";
+
+            var createTableOperation = _differ.GetDifferences(null, model)
+                .FirstOrDefault(o => (o as CreateTableOperation)?.Name == operation.Table);
```
bricelam: We need a more efficient way to do this; creating all the operations and throwing away all but one is wasteful. Exposing `ModelDiffer.Add(IEntityType source, IEntityType target)` might be enough.

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            if (!(createTableOperation is CreateTableOperation))
+            {
```
bricelam: It will always be a `CreateTableOperation`. Should this be `if (createTableOperation  == null)`

test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
```diff
@@ -0,0 +1,168 @@
...
+    {
+        private readonly DbContextOptionsBuilder _options;
+        private readonly IServiceProvider _provider;
+        private readonly SqliteTestStore _testStore;
+
+        [Fact]
+        public void DropColumn_rebuilds_table()
+        {
+            _testStore.ExecuteNonQuery(@"CREATE TABLE Table1 (Id INT PRIMARY KEY, Col1, Col2); 
+INSERT INTO Table1 (Col1, Col2) Values('dropped value','preserved entry');");
```
bricelam: Could use EF to setup the test. Create tables using `context.Database.EnsureCreated()` and insert as normal

test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            Migrate(up => { up.DropColumn("Col1", "Table1"); }, targetModel =>
+                {
+                    targetModel.Entity<UpdatedTableType>(b =>
+                        {
+                            b.Key(p => p.Id);
+                            b.Property(p => p.Id)
+                                .StoreGeneratedPattern(StoreGeneratedPattern.Identity);
```
bricelam: Unnecessary. `StoreGeneratedPattern` doesn't influence much on SQLite yet.

test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            using (var context = CreateContext())
+            {
+                // Assert data moved
+                var all = context.Set<UpdatedTableType>().ToList();
+                Assert.Equal(1, all.Count);
+                Assert.Equal(1, all[0].Id);
+                Assert.Equal("preserved entry", all[0].Col2);
+
+                // Assert Primary Key generates new values
```
bricelam: Asserting this doesn't seem related to the functionality you're testing. SQLite always generates new values for `INTEGER PRIMARY KEY` columns

test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                var model = context.GetService<IMigrationModelFactory>().CreateModel(migration.BuildTargetModel);
+                var command = context.GetService<IMigrationSqlGenerator>().Generate(operations, model);
+
+                using (var transaction = context.Database.GetDbConnection().BeginTransaction())
+                {
+                    context.GetService<ISqlStatementExecutor>()
+                        .ExecuteNonQuery(context.Database.GetRelationalConnection(), transaction, command);
+                    transaction.Commit();
+                }
```
bricelam: Any way to leverage parts of the Migrations pipeline for this?

test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
```diff
@@ -0,0 +1,168 @@
...
+                .BuildServiceProvider();
+
+            _options = new DbContextOptionsBuilder(new DbContextOptions<TestContext>());
+            _options.UseSqlite(_testStore.Connection);
+        }
+
+        private void AssertColumns(string tableName, List<string> columnNames)
+        {
+            var command = _testStore.Connection.CreateCommand();
+            command.CommandText = $"PRAGMA table_info({tableName});";
```
bricelam: Would be neat to leverage the materializer and do something like `context.Columns.FromSql("PRAGMA table_info({0})", tableName).ToList();`

test/EntityFramework.Sqlite.Tests/Migrations/SqliteOperationTransformTest.cs
```diff
@@ -0,0 +1,58 @@
...
+                    b.Property<string>("3").Column("col3");
+                    b.Key("3");
+                });
+            var transformer = new SqliteOperationTransformer(
+                new ModelDiffer(new SqliteTypeMapper(), new SqliteMetadataExtensionProvider()));
+            var actual = transformer.Transform(new[] { input }, model.Model);
+
+            Assert.Collection(actual, op1 =>
+                {
+                    var rename = (RenameTableOperation)op1;
```
bricelam: May get better failures with
```C#
var rename = Assert.IsType<RenameTableOperation>(op1);
```

bricelam:
:shipit:

test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
natemcmaster: True, but this was much easier.

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
natemcmaster: Will add as a TODO

test/EntityFramework.Sqlite.FunctionalTests/MigrationTest.cs
natemcmaster: No, but it would be nice to have an API for this in migrations so I don't have to recreate ApplyMigration()

PR #2440, opened 2015-06-22
Flatten connection and transaction APIs onto Database

ajcvickers:
As per API discussion, instead of indirecting through RelationalConnection and RelationalTransaction there is now a flat API off context.Database.

Also removed non-relational transaction base class and introduced an interface.


src/EntityFramework.Relational/RelationalDatabaseExtensions.cs
```diff
@@ -18,10 +21,56 @@ public static void ApplyMigrations([NotNull] this Database database)
...
+
+        public static Task OpenConnectionAsync(
+            [NotNull] this Database database, 
+            CancellationToken cancellationToken = default(CancellationToken))
+            => GetRelationalConnection(database).OpenAsync(cancellationToken);
+
+        public static void CloseConnection([NotNull] this Database database)
+            => GetRelationalConnection(database).Close();
+
+        public static DbTransaction GetDbTransaction([NotNull] this Database database)
```
AndriySvyryd: Shouldn't this return a IRelationalTransaction and be called GetTransaction? We are trying to discourage using the DbTansaction directly.

src/EntityFramework.Relational/IRelationalConnection.cs
```diff
@@ -17,21 +17,21 @@ public interface IRelationalConnection : IDataStoreConnection, IDisposable
 
         DbConnection DbConnection { get; }
 
-        RelationalTransaction Transaction { get; }
+        IRelationalTransaction Transaction { get; }
 
         int? CommandTimeout { get; set; }
 
         DbTransaction DbTransaction { get; }
```
AndriySvyryd: I don't remember whether we discussed this in the API review, but this property should be removed to discourage the use of DbTransaction directly.

src/EntityFramework.Relational/RelationalDatabaseExtensions.cs
ajcvickers: We decided in API review to flatten and not return RelationalTransaction anymore, and likewise for RelationalConnection. We don't want people to do stuff on either DbConnection or DbTransaction directly, but we still wanted to provide a way to get access, and this is what we decided on. I think it could be worth discussing whether the drop-down should be more hidden than this. I can file a bug on that so we can discuss again.

src/EntityFramework.Relational/IRelationalConnection.cs
ajcvickers: Same comment as before. We want to discourage people using it, but at the same time we want people to be able to get to it in some way.

src/EntityFramework.Relational/RelationalDatabaseExtensions.cs
AndriySvyryd: We should discuss this. I think having it just on IRelationalTransaction is enough.

AndriySvyryd:
:shipit:

src/EntityFramework.Relational/RelationalDatabaseExtensions.cs
ajcvickers: I have a list of things that I am going to file issues on--I have added it to that list to ensure we discuss it.

PR #2441, opened 2015-06-22
Rename Database class to DatabaseFacade

ajcvickers:
Per API review. Also removed DatabaseFactory since it no longer adds value for a pure facade class.


lajones:
:shipit:

ajcvickers:
Merged

PR #2447, opened 2015-06-23
Allow disabling testing of DateTimeOffset, byte, TimeSpan

roji:
PostgreSQL has no DateTimeOffset of byte types, this commit allows disabling some tests.

As an alternative, I can submit a PR breaking large multi-type tests such as `Can_query_using_any_data_type` to single-type tests which would be ignorable by specific providers with `[Fact(Skip)]`

### DIFF
--- a/test/EntityFramework.Core.FunctionalTests/BuiltInDataTypesTestBase.cs
+++ b/test/EntityFramework.Core.FunctionalTests/BuiltInDataTypesTestBase.cs
@@ -96,6 +96,7 @@ namespace Microsoft.Data.Entity.FunctionalTests
             using (var context = CreateContext())
             {
                 var entity = context.Set<BuiltInDataTypes>().Single(e => e.Id == 11);
+                var entityType = context.Model.GetEntityType(typeof(BuiltInDataTypes));
 
                 var param1 = (short)-1234;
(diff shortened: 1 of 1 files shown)


ErikEJ:
Would be reallly great if you could do the same for TimeSpan, pleeease! https://github.com/aspnet/EntityFramework/issues/2425

roji:
Ah :) @ErikEJ we seem to have the same sort of problems :) Will add a commit for TimeSpan.

roji:
Push-forced a different commit which implements the ignore in the way other fields were already ignored (e.g. unsigned int) instead of a new mechanism. Also added TimeSpan.

ErikEJ:
LGTM !

ajcvickers:
@roji Looks good--thanks for doing this. Could you sign the CLA? Once we get that I will merge it.

roji:
@ajcvickers I somehow thought I already did this (at least I'm sure I signed something for a roslyn contribution)... Can you point me the way?

ajcvickers:
@roji Here's the boilerplate message that is usually posted automatically: 

In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.dotnetfoundation.org.

Not sure why it didn't get posted this time--maybe something is wrong on our end.

roji:
Sleek!

PR #2448, opened 2015-06-23
Refactor ReaderModificationCommandBatch

### LINKED ISSUE #1956: ReaderModificationCommandBatch and affected rows
I'm working on an EF7 provider for PostgreSQL (with Npgsql), it's my very first foray into the EF codebase so I hope I won't be saying too much nonsense.

Looking at `ReaderModificationCommandBatch.Execute()`, I notice that it retrieves the affected rows (in the two ConsumeResultSet* methods). It appears to do this by reading the result of a `SELECT @@ROWCOUNT` appended by `SqlServerSqlGenerator.AppendSelectAffectedCountCommand()`.

PostgreSQL has no equivalent query; affected row counts are sent within the protocol itself, and are currently exposed on the ADO.NET level via the [IDbDataReader.RecordsAffected](https://msdn.microsoft.com/en-us/library/system.data.idatareader.recordsaffected(v=vs.110).aspx) property. In my NpgsqlModificationCommandBatch I can override Execute() and check against that property, but wanted to make sure I'm going down the right path:

* Is there a special reason to use an SqlServer-specific feature (`SELECT @@ROWCOUNT`) rather than the standard ADO.NET API for this? Does this have to do with the fact that RecordsAffected is only supposed to be populated after the reader is closed, and therefore gives the aggregate rows rather than resultset-by-resultset? Note that the PostgreSQL protocol *does* provide affected rows on a resultset-by-resultset basis, so this information can in theory be exposed to the EF provider if absolutely necessary.
* It seems a bit odd for `ReaderModificationCommandBatch`, a generic class within EntityFramework.Relational to encode logic that is so specific to the SqlServer adapter, while at the same time containing other, more generic functionality.
* You should probably consider making `SqlGenerator.AppendSelectAffectedCountCommand` virtual rather than abstract.

Thanks for any advice on this...

roji:
The previous implementation of ReaderModificationCommandBatch was written around the assumption that the number of affected rows is fetched via an additional SQL query appended to the batch. This isn't for PostgreSQL, where the affected rows are only returned in the protocol and exposed via a special Npgsql API.

Closes #1956

### DIFF
--- a/src/EntityFramework.Relational/Update/AffectedCountModificationCommandBatch.cs
+++ b/src/EntityFramework.Relational/Update/AffectedCountModificationCommandBatch.cs
@@ -1,0 +1,259 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data.Common;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.ChangeTracking.Internal;
+using Microsoft.Data.Entity.Update;
+using Microsoft.Data.Entity.Utilities;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.Data.Entity.Relational.Update
+{
+    /// <summary>
+    /// A <see cref="ReaderModificationCommandBatch"/> for providers which append an SQL query to find out
+    /// how many rows were affected (see <see cref="SqlGenerator.AppendSelectAffectedCountCommand"/>).
+    /// </summary>
+    public abstract class AffectedCountModificationCommandBatch : ReaderModificationCommandBatch
+    {
+        private readonly List<bool> _resultSetEnd = new List<bool>();
+
+        protected AffectedCountModificationCommandBatch(
+            [NotNull] ISqlGenerator sqlGenerator)
+            : base(sqlGenerator)
+        {
+        }
+
+        // contains true if the command at the corresponding index is the last command in its result set
+        // the last value will not be read
+        protected IList<bool> ResultSetEnds => _resultSetEnd;
+
+        public override bool AddCommand(ModificationCommand modificationCommand)
+        {
+            _resultSetEnd.Add(true);
+            var added = base.AddCommand(modificationCommand);
+            if (!added)
+            {
+                _resultSetEnd.RemoveAt(_resultSetEnd.Count - 1);
+            }
+            return added;
+        }
+
+        protected override void Consume(DbDataReader reader, DbContext context)
+        {
+            Debug.Assert(ResultSetEnds.Count == ModificationCommands.Count);
+            var commandIndex = 0;
+
+            try
+            {
+                var actualResultSetCount = 0;
+                do
+                {
+                    commandIndex = ModificationCommands[commandIndex].RequiresResultPropagation
+                        ? ConsumeResultSetWithPropagation(commandIndex, reader, context)
+                        : ConsumeResultSetWithoutPropagation(commandIndex, reader, context);
+                    actualResultSetCount++;
+                } while (commandIndex < ResultSetEnds.Count
+                         && reader.NextResult());
+
+                Debug.Assert(commandIndex == ModificationCommands.Count,
+                    "Expected " + ModificationCommands.Count + " results, got " + commandIndex);
+#if DEBUG
+                var expectedResultSetCount = 1 + ResultSetEnds.Count(e => e);
+                expectedResultSetCount += ResultSetEnds[ResultSetEnds.Count - 1] ? -1 : 0;
+
+                Debug.Assert(actualResultSetCount == expectedResultSetCount,
+                    "Expected " + expectedResultSetCount + " result sets, got " + actualResultSetCount);
+#endif
+            }
+            catch (DbUpdateException)
+            {
+                throw;
+            }
+            catch (Exception ex)
+            {
+                throw new DbUpdateException(
+                    Strings.UpdateStoreException,
+                    ex,
+                    ModificationCommands[commandIndex].Entries);
+            }
+        }
+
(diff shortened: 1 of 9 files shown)


src/EntityFramework.Relational/Update/AffectedCountModificationCommandBatch.cs
```diff
@@ -0,0 +1,313 @@
...
+        {
+            var added = base.AddCommand(modificationCommand);
+            if (added)
+            {
+                _resultSetEnd.Add(true);
+            }
+            return added;
+        }
+
+        public override int Execute(
```
AndriySvyryd: The code outside "using reader" has no assumptions about the reader contents so it can be pulled up.

src/EntityFramework.Relational/Update/AffectedCountModificationCommandBatch.cs
```diff
@@ -0,0 +1,313 @@
...
+                        context,
+                        ex,
+                        commandIndex < ModificationCommands.Count ? ModificationCommands[commandIndex].Entries : new InternalEntityEntry[0]);
+                }
+            }
+
+            return commandIndex;
+        }
+
+        private int ConsumeResultSetWithPropagation(int commandIndex, DbDataReader reader, DbContext context)
```
AndriySvyryd: The Consume* methods should be protected virtual

src/EntityFramework.Relational/Update/AffectedCountModificationCommandBatch.cs
```diff
@@ -0,0 +1,313 @@
...
+                if (!await reader.ReadAsync(cancellationToken))
+                {
+                    var expectedRowsAffected = rowsAffected + 1;
+                    while (++commandIndex < ResultSetEnds.Count
+                           && !ResultSetEnds[commandIndex - 1])
+                    {
+                        expectedRowsAffected++;
+                    }
+
+                    throw new DbUpdateConcurrencyException(
```
AndriySvyryd: The code for throwing this exception is pretty generic, so it can be extracted as a method and pulled up

AndriySvyryd:
You should also make SqlGenerator.AppendSelectAffectedCountCommand virtual

src/EntityFramework.Relational/Update/ReaderModificationCommandBatch.cs
```diff
@@ -248,181 +221,16 @@ protected virtual void UpdateCachedCommandText(int commandPosition)
...
-
-                if (!reader.Read())
-                {
-                    var expectedRowsAffected = rowsAffected + 1;
-                    while (++commandIndex < ResultSetEnds.Count
-                           && !ResultSetEnds[commandIndex - 1])
-                    {
-                        expectedRowsAffected++;
-                    }
+        protected abstract void Execute(DbDataReader reader, DbContext context);
```
AndriySvyryd: Since the command has already been executed consider renaming this to `Consume`

AndriySvyryd:
@roji please also update  `EntityFramework.Relational.Tests\UpdateReaderModificationCommandBatchTest.cs` and run `build.cmd`

roji:
Thanks for your patience @AndriySvyryd, I'm a newbie contributing changes to EF7.

I've fixed `EntityFramework.Relational.Tests\UpdateReaderModificationCommandBatchTest.cs` to inherit `AffectedCountModificationCommandBatch` and it now compiles fine with dnx451. However, net46 fails with the following exception:
```
C:\projects\npgsql\external\EntityFramework\test\EntityFramework.Relational.Tests\Migrations\Infrastructure\MigrationIdGeneratorTest.cs(17,13): error CS0012: The type 'Regex' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Text.RegularExpressions, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
```

Running build.cmd (I haven't been doing this till now, just `dnu compile` on specific projects) fails on dnxcore50 on EntityFramework.Commands since System.IServiceProvider doesn't seem to exist.

I guess something's wrong with the way I've set up my environment, but these errors at least seem unrelated to this PR. Please let me know if you'd like any further fixes.

AndriySvyryd:
@roji Sounds like you are using an older dnx, try running `packages\KoreBuild\build\dnvm.cmd upgrade` from the enlistment root. The current version is beta6-12120
After upgrading run 'git clean -xdf' to start from a clean state.

As for the PR you also need to update `test\EntityFramework.Relational.Tests\Update\BatchExecutorTest.cs`, the mocks for `ModificationCommandBatch` need to return a non-null value from `.ModificationCommands`

roji:
Fixed the mocks, thanks again for the patience. Rebased on latest dev and push-forced.

As for the environment, I'm on 1.0.0-beta6-12120 (clr, x86) and am getting the build errors [above](https://github.com/aspnet/EntityFramework/pull/2448#issuecomment-115374259).

AndriySvyryd:
Now `Keys_generated_in_batches` in `test\EntityFramework.SqlServer.FunctionalTests\CommandConfigurationTest.cs` and `Batches_are_divided_correctly_with_two_inserted_columns` in `test\EntityFramework.SqlServer.FunctionalTests\BatchingTest.cs` are failing.

You can also run the tests using http://testdriven.net/

roji:
Fixed and rebased on latest dev.

Test fail list for EntityFramework.Relational.Tests, EntityFramework.SqlServer.FunctionalTests looks OK.

Do you guys happen to have a publicly-accessible build server where I can track see the build status of my PRs, failing tests etc?

AndriySvyryd:
We do not have a publicly visible CI server, since everyone with commit rights have internal access. But the are some investigations going on that would make it easier to make sure the PR passes the tests.

roji:
OK, thanks again for your patience!

PR #2452, opened 2015-06-23
Fix to make Reverse Engineering work on dnxcore50

lajones:
Enabled the E2E test on dnxcore50 as well.

The important part of the change is using the explicit data type instead of var in cases where a) the variable being used was dependent on Model, which is a dynamic property declared on the base class, and b) that variable was later used in a binary operation.

Also took the chance to remove @ symbols that weren't being used and to make the spacing of foreach's and if's consistent across both templates.

### DIFF
--- a/src/EntityFramework.SqlServer.Design/ReverseEngineering/Templates/SqlServerDbContextTemplate.cshtml
+++ b/src/EntityFramework.SqlServer.Design/ReverseEngineering/Templates/SqlServerDbContextTemplate.cshtml
@@ -51,16 +51,16 @@ var className = Model.ClassName ?? Model.Helper.ClassName(Model.ConnectionString
     {
         protected override void OnConfiguring(DbContextOptionsBuilder options)
         {
-            options.UseSqlServer(@Model.Helper.VerbatimStringLiteral(@Model.ConnectionString));
+            options.UseSqlServer(@Model.Helper.VerbatimStringLiteral(Model.ConnectionString));
         }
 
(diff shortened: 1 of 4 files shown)


src/EntityFramework.SqlServer.Design/ReverseEngineering/Templates/SqlServerDbContextTemplate.cshtml
```diff
@@ -98,7 +98,7 @@ var firstEntity = true;
         {
 @:                entity.Property(e => e.@(propertyConfig.Property.Name))
             var lineCount = 0;
-            @foreach (var line in propertyConfigurationLines)
+            @foreach (string line in propertyConfigurationLines)
```
bricelam: May want to add a comment saying they can't be `var`.

bricelam:
:shipit: Should this go into release?

lajones:
@bricelam Discussed release with @divega yesterday - decided it's too late.

src/EntityFramework.SqlServer.Design/ReverseEngineering/Templates/SqlServerDbContextTemplate.cshtml
lajones: Sure - will do.

bricelam:
Yep, I realized that after I asked. :grinning:

lajones:
Checked in fix with commit 6702541.

Issue #2421, opened 2015-06-18
Migrations: Defer type mapping


bricelam:
Like in EF6, we should defer type mapping until migration apply time. For example:

```C#
migrationBuilder.CreateTable(
    name: "MyTable",
    columns: table => new
    {
        Id = table.Column<int>()
    });

migrationBuilder.AlterColumn<int>(
    name: "Id"
    table: "MyTable");

migrationBuilder.AddColumn<string>(
    name: "Name"
    table: "MyTable");

migrationBuilder.CreateSequence<long>("MySequence");
```

PR #2453, opened 2015-06-23
Count right, SQLite.

### LINKED ISSUE #2432: SQLite: Use AUTOINCREMENT when possible
[AUTOINCREMENT](http://sqlite.org/autoinc.html) prevents the reuse of ROWIDs from previously deleted rows.

Without this, you could hit some weird concurrency issues. Consider the following.

```C#
using (var context1 = new Context())
using (var context2 = new Context());
{
    var alfred1 = context1.People.First(p => p.Id = 2);
    var alfred2 = context2.People.First(p => p.Id = 2);

    context1.People.Remove(alfred);
    context1.People.Add(new Person { Name = "Batman" });
    context1.SaveChanges(); // Adds Batman with Id = 2

    alfred2.LastName = "Pennyworth";
    context2.SaveChanges(); // Updates Batman!
}
```

natemcmaster:
Fix #2432. Uses AUTOINCREMENT keyword in SQLite to handle quirks in their rowid algorithm.

### DIFF
--- a/src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
+++ b/src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Relational.Migrations.Infrastructure;
+using Microsoft.Data.Entity.Sqlite.Metadata;
+
+namespace Microsoft.Data.Entity.Sqlite.Migrations
+{
+    public class SqliteMigrationAnnotationProvider : MigrationAnnotationProvider
+    {
+        public override IEnumerable<IAnnotation> For(IProperty property)
+        {
+            if (property.StoreGeneratedPattern == StoreGeneratedPattern.Identity)
+            {
+                yield return new Annotation(SqliteAnnotationNames.Prefix + SqliteAnnotationNames.Autoincrement, true);
+            }
+        }
+    }
+}
--- a/src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
+++ b/src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
@@ -5,10 +5,12 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Relational;
 using Microsoft.Data.Entity.Relational.Migrations.Operations;
 using Microsoft.Data.Entity.Relational.Migrations.Sql;
+using Microsoft.Data.Entity.Sqlite.Metadata;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Sqlite.Migrations
@@ -85,6 +87,43 @@ namespace Microsoft.Data.Entity.Sqlite.Migrations
                 .Append(columnList)
                 .Append(" FROM ")
                 .Append(_sql.DelimitIdentifier(operation.OldTable));
+        }
+
+        public override void Generate(CreateTableOperation operation, IModel model, SqlBatchBuilder builder)
+        {
+            // Lifts a primary key definition into the typename.
+            // This handles the quirks of creating integer primary keys using autoincrement, not default rowid behavior.
+            if (operation.PrimaryKey?.Columns.Length == 1)
+            {
+                var columnOp = operation.Columns?.FirstOrDefault(o => o.Name == operation.PrimaryKey.Columns[0]);
+                if (columnOp != null)
+                {
+                    columnOp.AddAnnotation(SqliteAnnotationNames.Prefix + SqliteAnnotationNames.InlinePrimaryKey, true);
+                    operation.PrimaryKey = null;
+                }
+            }
+
+            base.Generate(operation, model, builder);
+        }
+
+        public override void ColumnDefinition(string schema, string table, string name, string type, bool nullable, object defaultValue, string defaultExpression, IAnnotatable annotatable, IModel model, SqlBatchBuilder builder)
+        {
+            base.ColumnDefinition(schema, table, name, type, nullable, defaultValue, defaultExpression, annotatable, model, builder);
+
+            var columnAnnotation = annotatable as Annotatable;
+            var inlinePk = columnAnnotation?.FindAnnotation(SqliteAnnotationNames.Prefix + SqliteAnnotationNames.InlinePrimaryKey);
+
+            if (inlinePk != null
+                && (bool)inlinePk.Value)
+            {
+                builder.Append(" PRIMARY KEY");
+                var autoincrement = columnAnnotation.FindAnnotation(SqliteAnnotationNames.Prefix + SqliteAnnotationNames.Autoincrement);
+                if (autoincrement != null
+                    && (bool)autoincrement.Value)
+                {
+                    builder.Append(" AUTOINCREMENT");
+                }
+            }
         }
 
         #region Invalid migration operations
--- a/test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
+++ b/test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.Data.Entity.Relational.Migrations.Operations;
 using Microsoft.Data.Entity.Relational.Migrations.Sql;
+using Microsoft.Data.Entity.Sqlite.Metadata;
 using Xunit;
 
 namespace Microsoft.Data.Entity.Sqlite.Migrations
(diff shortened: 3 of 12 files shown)


src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
```diff
@@ -87,6 +87,25 @@ public virtual void Generate([NotNull] MoveDataOperation operation, [CanBeNull]
...
+        public override void Generate(CreateTableOperation operation, IModel model, SqlBatchBuilder builder)
+        {
+            // Lifts a primary key definition into the typename.
+            // This handles the quirks of creating integer primary keys using autoincrement, not default rowid behavior.
+            if (operation.PrimaryKey?.Columns.Length == 1)
+            {
+                var column = operation.Columns?.FirstOrDefault(c => c.Name == operation.PrimaryKey.Columns[0]);
+                if (column != null
+                    && (column.Type.Equals("INTEGER", StringComparison.InvariantCultureIgnoreCase)
+                        || column.Type.Equals("INT", StringComparison.InvariantCultureIgnoreCase)))
```
bricelam: Must be exactly "INTEGER" any other value won't work.

bricelam: We could probably just override the user-specified type with "INTEGER" if the CLR type is an integral type. I'll look into this with #2421.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
```diff
@@ -87,6 +87,25 @@ public virtual void Generate([NotNull] MoveDataOperation operation, [CanBeNull]
...
+                {
+                    column.Type = "INTEGER PRIMARY KEY AUTOINCREMENT";
```
bricelam: Hmm, I think we should use annotations for autoincrement and inline primary key.

test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
```diff
@@ -47,33 +63,6 @@ public void RestartSequenceOperation_not_supported()
...
-                    Type = "int",
-                    IsNullable = true,
-                    DefaultExpression = "10"
-                });
-
-            Assert.Equal(
-                @"ALTER TABLE ""People"" ADD ""Age"" int DEFAULT (10);" + EOL,
-                Sql);
-        }
-
```
bricelam: Why were these removed?

natemcmaster: They were testing for a scenario we removed several commits ago  (throwing on CURRENT_TIMESTAMP in the SQL generator). The base class still tests add column.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
natemcmaster: Right. "INT" will not work with autoincrement. The typemapper should be returning INTEGER, but this catches some cases in our testing where we specify "int" (and possibly where a user specifies "int").

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
natemcmaster: To clarify, you're suggesting we apply this logic during model building instead and then use annotations to generate a different CreateTable operation? Or some other place...

test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
bricelam: We decided to pass the SQL through to the database and let it throw. These test that we're generating the "correct" (or at least valid-to-parse) SQL.

natemcmaster:
Will update when #2431 is merged.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
bricelam: I think we should either...
1. Always override the type to be INTEGER
2. Only add AUTOINCREMENT when the type is exactly INTEGER

For now, let's detect if the type merely *contains* "INT" and override it with INTEGER. This is similar logic to how SQLite determines column affinity.

After #2421 is in, we can be even smarter an base it on the CLR type instead of the column type name.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
bricelam: To capture what we discussed yesterday:

1. Add an annotation like `Sqlite:Autoincrement = true` to the property during model diffing
2. Add an annotation like `Sqlite:InlinePrimaryKey = true` to the property as an implementation detail of the SqliteMigrationSqlGenerator

natemcmaster:
@bricelam added the new implementation.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+                    columnType = null;
+                }
+                if (columnType != null
+                    && columnType.Equals("integer", StringComparison.CurrentCultureIgnoreCase)
+                    && clrType != typeof(bool)
+                    && clrType != typeof(char)
+                    && clrType != typeof(byte)
+                    )
+                {
+                    yield return new Annotation(SqliteAnnotationNames.Prefix + SqliteAnnotationNames.Autoincrement, true);
```
bricelam: This may make more sense on the property

src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+            base.For(key);
+        }
+
+        public override IEnumerable<IAnnotation> For(IProperty property)
+        {
+            if (property.IsPrimaryKey()
+                && property.EntityType.GetPrimaryKey().Properties.Count == 1)
+            {
+                yield return new Annotation(SqliteAnnotationNames.Prefix + SqliteAnnotationNames.InlinePrimaryKey, true);
```
bricelam: Not needed here. This can just be an implementation detail of the `SqliteMigrationSqlGenerator`.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+                    && columnType.Equals("integer", StringComparison.CurrentCultureIgnoreCase)
+                    && clrType != typeof(bool)
+                    && clrType != typeof(char)
+                    && clrType != typeof(byte)
+                    )
+                {
+                    yield return new Annotation(SqliteAnnotationNames.Prefix + SqliteAnnotationNames.Autoincrement, true);
+                }
+            }
+            base.For(key);
```
bricelam: Don't call base on these.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
```diff
@@ -87,6 +88,38 @@ public virtual void Generate([NotNull] MoveDataOperation operation, [CanBeNull]
...
+            var inlinePrimaryKeyAnnotation = operation.Columns
+                .FirstOrDefault(c => c.FindAnnotation(inlinePrimaryKey) != null)?[inlinePrimaryKey];
+
+            if (inlinePrimaryKeyAnnotation != null
+                && (bool)inlinePrimaryKeyAnnotation)
+            {
+                var column = operation.Columns?.FirstOrDefault(c => c.Name == operation.PrimaryKey.Columns[0]);
+                if (column != null)
+                {
+                    column.Type += " PRIMARY KEY";
```
bricelam: Don't abuse the `Type` property like this. Override how the SQL gets generated in `ColumnDefinition`

test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
bricelam: Add these tests back.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
natemcmaster: Agreed. Moved it to property.

natemcmaster:
@bricelam cleaned up the implementation to respond to your feedback. Better?

src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+                }
+                catch (NotSupportedException)
+                {
+                    columnType = null;
+                }
+                if (columnType != null
+                    && columnType.Equals("integer", StringComparison.CurrentCultureIgnoreCase)
+                    && clrType != typeof(bool)
+                    && clrType != typeof(char)
+                    && clrType != typeof(byte)
```
bricelam: Does the CLR type actually matter? If it's an INTEGER column, things should "just work" for any CLR type.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+                try
+                {
+                    columnType = property.Sqlite().ColumnType ?? _typeMapper.GetDefaultMapping(property.ClrType).DefaultTypeName;
+                }
+                catch (NotSupportedException)
+                {
+                    columnType = null;
+                }
+                if (columnType != null
+                    && columnType.Equals("integer", StringComparison.CurrentCultureIgnoreCase)
```
bricelam: Also only if `property.StoreGeneratedPattern == StoreGeneratedPattern .Identity` This will let you turn it off in the case that the keys are managed by the application.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
```diff
@@ -87,6 +89,68 @@ public virtual void Generate([NotNull] MoveDataOperation operation, [CanBeNull]
...
+                    .Append(" DEFAULT (")
+                    .Append(defaultExpression)
+                    .Append(")");
+            }
+            else if (defaultValue != null)
+            {
+                builder
+                    .Append(" DEFAULT ")
+                    .Append(_sql.GenerateLiteral((dynamic)defaultValue));
+            }
```
bricelam: :frowning: Is there a way to not have to re-implement the base method? I think SQL Server calls base then tacks on `IDENTITY` to the end of the string.

test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
```diff
@@ -26,25 +27,86 @@ public void Insert_into_select()
             Assert.Equal("INSERT INTO \"RebuiltTable\" (\"col1\", \"col2\")" + EOL + "SELECT \"col1\", \"col2\" FROM \"OldTable\";" + EOL, Sql);
         }
 
-        [Fact]
-        public void CreateSchemaOperation_not_supported()
+        public override void CreateTableOperation()
         {
-            var ex = Assert.Throws<NotSupportedException>(() => Generate(new CreateSchemaOperation()));
-            Assert.Equal(Strings.SchemasNotSupported, ex.Message);
+            // replaced with test below
```
bricelam: Keep the original and add a new method instead? "Maintainable code" takes on a different meaning in test code--it's often OK to copy-and-paste in tests since it's better to be explicit about what's being tested in each case.

bricelam:
:shipit: (once feedback is addressed)

src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
bricelam: Also, an auto-generated bool key sounds like fun: 1, 2, 3... becomes true, true, true... on the client

natemcmaster: I think technically it would work, but users may not expect this behavior if the CLR type isn't a integer-ish type (including short, long, and unsigned variants.)

src/EntityFramework.Sqlite/Migrations/SqliteMigrationSqlGenerator.cs
natemcmaster: SQLites is very specific about the order of words. 'integer [not null] autoincrement primary key' != 'integer primary key autoincrement not null'.

src/EntityFramework.Sqlite/Migrations/SqliteMigrationAnnotationProvider.cs
natemcmaster: Summing up what we discussed: we will let users shoot themselves in the foot with autoincremented boolean. We only check now for StoreGeneratedPattern == Identity

PR #2454, opened 2015-06-23
Start using "database" instead of "data store" in class names, etc.

ajcvickers:
As per API review decision.


src/EntityFramework.Core/Update/DbUpdateException.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Microsoft.Data.Entity.Update
 {
-    public class DbUpdateException : DataStoreException
+    public class DbUpdateException : DatabaseException
```
lajones: Wasn't this DatabaseUpdateException?

src/EntityFramework.InMemory/InMemoryDatabase.cs
```diff
@@ -1,171 +1,90 @@
...
+            [NotNull] ILoggerFactory loggerFactory)
+            : base(
+                Check.NotNull(model, nameof(model)),
+                Check.NotNull(entityKeyFactorySource, nameof(entityKeyFactorySource)),
+                Check.NotNull(entityMaterializerSource, nameof(entityMaterializerSource)),
+                Check.NotNull(clrPropertyGetterSource, nameof(clrPropertyGetterSource)),
+                Check.NotNull(loggerFactory, nameof(loggerFactory)))
         {
-            Check.NotNull(model, nameof(model));
+            Check.NotNull(persistentStore, nameof(persistentStore));
```
lajones: Do we need Check.NotNull(options)?

src/EntityFramework.Core/Update/DbUpdateException.cs
ajcvickers: DbUpdateException will go to SaveChangesException
DataStoreException will be removed if possible

I'm doing these changes in stages because there are so many. This change is just general renaming from data store to database, with some conflicts resolved by other renames. I still have other more specific changes to come.

src/EntityFramework.InMemory/InMemoryDatabase.cs
ajcvickers: Yes.

src/EntityFramework.Core/Update/DbUpdateException.cs
lajones: Sounds good.

lajones:
:shipit:

PR #2462, opened 2015-06-24
RevEng: Hook up IApplicationShutdown Cancel event and use async methods where possible

lajones:
@AndriySvyryd @divega Updating as agreed. Using .GetAwaiter().GetResult() instead of .Result returns the underlying exception rather than an AggregateException if an exception is thrown.

### DIFF
--- a/src/EntityFramework.Commands/DatabaseTool.cs
+++ b/src/EntityFramework.Commands/DatabaseTool.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Data.Entity.Commands
             };
 
             var generator = new ReverseEngineeringGenerator(_serviceProvider);
-            return generator.GenerateAsync(configuration).Result;
+            return generator.GenerateAsync(configuration).GetAwaiter().GetResult();
         }
     }
 }
--- a/src/EntityFramework.Commands/DatabaseTool.cs
+++ b/src/EntityFramework.Commands/DatabaseTool.cs
@@ -4,6 +4,8 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Commands.Utilities;
 using Microsoft.Data.Entity.Internal;
@@ -44,11 +46,12 @@ namespace Microsoft.Data.Entity.Commands
             _serviceProvider.AddService(typeof(ITemplating), new RazorTemplating(compilationService, metadataReferencesProvider));
         }
 
-        public virtual IEnumerable<string> ReverseEngineer(
+        public virtual async Task<IEnumerable<string>> ReverseEngineerAsync(
             [NotNull] string providerAssemblyName,
             [NotNull] string connectionString,
             [NotNull] string rootNamespace,
-            [NotNull] string projectDir)
+            [NotNull] string projectDir,
+            CancellationToken cancellationToken = default(CancellationToken))
         {
             Check.NotNull(providerAssemblyName, nameof(providerAssemblyName));
             Check.NotEmpty(connectionString, nameof(connectionString));
@@ -70,7 +73,7 @@ namespace Microsoft.Data.Entity.Commands
             };
 
             var generator = new ReverseEngineeringGenerator(_serviceProvider);
-            return generator.GenerateAsync(configuration).GetAwaiter().GetResult();
+            return await generator.GenerateAsync(configuration, cancellationToken);
         }
     }
 }
--- a/src/EntityFramework.Commands/Executor.cs
+++ b/src/EntityFramework.Commands/Executor.cs
@@ -8,6 +8,8 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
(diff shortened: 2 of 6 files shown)


src/EntityFramework.Commands/DatabaseTool.cs
```diff
@@ -70,7 +70,7 @@ public class DatabaseTool
             };
 
             var generator = new ReverseEngineeringGenerator(_serviceProvider);
-            return generator.GenerateAsync(configuration).Result;
+            return generator.GenerateAsync(configuration).GetAwaiter().GetResult();
```
divega: Why not make things async throughout? AFAIK DNX supports async `Program.Main`. Is it hard to use async methods from PowerShell?

cc @bricelam

lajones:
@danielmarbach Note: this addresses the changes you included in #2240 as well.

danielmarbach:
Ok. Yeah didn't have time to do it. Btw. There is a [NotNull] missing in the Program

lajones:
Thanks @danielmarbach - well-spotted. Will fix.

src/EntityFramework.Commands/Program.cs
```diff
@@ -222,7 +232,11 @@ public virtual int Main([NotNull] string[] args)
                             "[connectionString]",
                             "The connection string of the database");
 
-                    revEng.OnExecute(() => ReverseEngineer(connectionString.Value));
+                    revEng.OnExecute(async () =>
+                        {
+                            return await ReverseEngineerAsync(
+                                connectionString.Value, _applicationShutdown.ShutdownRequested);
+                        });
```
lajones: Note: Have updated the above to simpler async () => await ... syntax.

src/EntityFramework.Commands/DatabaseTool.cs
```diff
@@ -70,7 +73,7 @@ public class DatabaseTool
             };
 
             var generator = new ReverseEngineeringGenerator(_serviceProvider);
-            return generator.GenerateAsync(configuration).Result;
+            return await generator.GenerateAsync(configuration, cancellationToken);
```
AndriySvyryd: No need to use await here, just return what GenerateAsync returns

src/EntityFramework.Commands/Executor.cs
```diff
@@ -249,13 +251,18 @@ public ReverseEngineer([NotNull] Executor executor, [NotNull] object resultHandl
...
 
-        public virtual IEnumerable<string> ReverseEngineerImpl(
-            [NotNull] string providerAssemblyName, [NotNull] string connectionString) =>
-                _databaseTool.ReverseEngineer(providerAssemblyName, connectionString, _rootNamespace, _projectDir);
+        public virtual async Task<IEnumerable<string>> ReverseEngineerImplAsync(
+            [NotNull] string providerAssemblyName,
+            [NotNull] string connectionString,
+            CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return await _databaseTool.ReverseEngineerAsync(
```
AndriySvyryd: remove await

src/EntityFramework.Commands/Program.cs
AndriySvyryd: You can remove async/await

AndriySvyryd:
:shipit:

src/EntityFramework.Commands/Executor.cs
```diff
@@ -249,13 +251,18 @@ public ReverseEngineer([NotNull] Executor executor, [NotNull] object resultHandl
 
                 var providerAssemblyName = DatabaseTool._defaultReverseEngineeringProviderAssembly;
 
-                Execute(() => executor.ReverseEngineerImpl(providerAssemblyName, connectionString));
+                Execute(() => executor.ReverseEngineerImplAsync(providerAssemblyName, connectionString).GetAwaiter().GetResult());
```
divega: Can this be made async?

divega: Or is this the PowerShell stuff?

divega: I guess it is :smile:

src/EntityFramework.Commands/DatabaseTool.cs
lajones: If you do that the compiler complains that this method has no await in it.

src/EntityFramework.Commands/Executor.cs
lajones: As above

src/EntityFramework.Commands/Executor.cs
lajones: Yep - this is the place where we agreed the sync-async crossover would take place for Powershell.

src/EntityFramework.Commands/DatabaseTool.cs
AndriySvyryd: you need to remove the async keyword as well

src/EntityFramework.Commands/Program.cs
lajones: Nope - produces compiler errors warning that the async and await are necessary.

divega: I suspect in this case it is needed.

lajones:
OK. Have talked with Andriy and sorted out the async calls.

lajones:
Fixed with commit c75abcf.

PR #2470, opened 2015-06-25
Option to not test dirty reads

roji:
In PostgreSQL, dirty reads don't ever occur, even in isolation level "read uncommitted". Added an option to skip the dirty read check in the transaction tests.

http://www.postgresql.org/docs/9.4/static/transaction-iso.html


rowanmiller:
@AndriySvyryd please process this one

AndriySvyryd:
@roji could you rebase on dev?

roji:
Rebased

Issue #2456, opened 2015-06-23
Remove ValueBufferProperty query extension


mikary:
Additional cleanup for #2422

PR #2472, opened 2015-06-25
Remove ValueBufferProperty query extension

mikary:
Move static EF class into Microsoft.Data.Entity namespace
Fix for #2456, #2422
@anpete @ajcvickers


anpete:
:shipit:

PR #2473, opened 2015-06-25
API review changes

ajcvickers:
- Remove members from public IDatabase contract that are not publicly consumed
- Remove DatabaseException as YAGNI
- Do not rename DbUpdateException or DbUpdateConcurrencyException even though we had decided to because it doesn't meet the bar in guiding principles for changing the name of existing classes.
- Remove unused IRelationalDataStore
- Remove IDatabaseConnection
- Make IDatabaseProviderSelector internal
- Remove auto-configuration of providers--A UseXxx method must always be called
- Move ProviderInvariantName to Services class to make it easier to access from anywhere and make it return provider assembly name by default
- Rename "Name" to "ConstraintName", "IndexName", and "KeyName"
- Rename "Table" to "ToTable" to align with old name
- Renamed metadata XxxExtensions classes to XxxAnnotations
- Rename "Colunn" to "HasColumnName" to align with old name
- Rename "ColunnType" to "HasColumnType" to align with old name
- Note: Have not removed ForSqlServer yet


AndriySvyryd:
:shipit:

Issue #2274, opened 2015-05-26
Chaining multiple Include with ThenInclude throws System.InvalidOperationException


nicbourgoin:
Summary:

var field = context.Fields
   .Include(x => x.Label.Globalizations)
   .ThenInclude(x => x.Language)
   .Include(x => x.Placeholder.Globalizations)
   .ThenInclude(x => x.Language)
   .ToList();

I get the following exception thrown:

System.InvalidOperationException Invalid attempt to read when no data is present.

I asked the question on stackoverflow @
http://stackoverflow.com/questions/30407829/chaining-multiple-include-with-theninclude-throws-system-invalidoperationexcepti

anpete:
Hi @nicbourgoin, can you post the full stack trace?

nicbourgoin:
@anpete 

InvalidOperationException: Invalid attempt to read when no data is present.

System.Data.SqlClient.SqlDataReader.CheckDataIsReady(Int32 columnIndex, Boolean allowPartiallyReadColumn, Boolean permitAsync, String methodName)


System.Data.SqlClient.SqlDataReader.TryReadColumn(Int32 i, Boolean setTimeout, Boolean allowPartiallyReadColumn)


System.Data.SqlClient.SqlDataReader.GetValues(Object[] values)


Microsoft.Data.Entity.Relational.RelationalObjectArrayValueReader..ctor(DbDataReader dataReader)


Microsoft.Data.Entity.Relational.RelationalObjectArrayValueReaderFactory.Create(DbDataReader dataReader)


Microsoft.Data.Entity.Relational.Query.RelationalQueryContext.CreateValueReader(Int32 readerIndex)


Microsoft.Data.Entity.Relational.Query.QueryMethodProvider.ReferenceIncludeRelatedValuesStrategy.<GetRelatedValues>d__5.MoveNext()


System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()


System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()


System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)


System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)


Microsoft.Data.Entity.Query.QueryBuffer.Include(Object entity, IReadOnlyList`1 navigationPath, IReadOnlyList`1 relatedEntitiesLoaders, Int32 currentNavigationIndex, Boolean querySourceRequiresTracking)


[...]

PR #2474, opened 2015-06-26
Fix to #2274 - Chaining multiple Include with ThenInclude throws Sysem.InvalidOperationException

maumar:
Problem is that for complex queries with multiple includes we were attempting to materialize entities based on incorrect readers. Basically the way it works is that when we process include we build a factories that produce entities from actively opened readers.

Those were being build in isolation, starting fresh for each IncludeSpecification. Problem is that multiple include chains can leave readers open, that the other include chain was not aware of. This would produce errors like NRE, index outside bounds, invalid cast etc, when attempting to access incorrect reader while materializing entities.

Fix is to correctly compute indeces (which reader should be accessed for each entity) that each include chain needs to access. Computation is being done in the reverse order because this is the order that readers will eventually be executed in. Also we only compute indeces for Reference includes - Collections don't need them as they always open a new reader.

### DIFF
--- a/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
@@ -221,13 +221,23 @@ namespace Microsoft.Data.Entity.Query
 
                             return new IncludeSpecification(annotation.QuerySource, navigationPath);
                         })
+                    .OrderBy(a => a.NavigationPath.First().PointsToPrincipal())
                     .ToList();
+
+            IncludeNavigations(queryModel, includeSpecifications);
+        }
+
+        protected virtual void IncludeNavigations(
+            [NotNull] QueryModel queryModel,
+            [NotNull] IReadOnlyCollection<IncludeSpecification> includeSpecifications)
+        {
+            Check.NotNull(queryModel, nameof(queryModel));
+            Check.NotNull(includeSpecifications, nameof(includeSpecifications));
 
             var querySourceTracingExpressionVisitor
                 = new QuerySourceTracingExpressionVisitor();
 
-            foreach (var includeSpecification in includeSpecifications
-                .OrderBy(a => a.NavigationPath.First().PointsToPrincipal()))
+            foreach (var includeSpecification in includeSpecifications)
             {
                 var resultQuerySourceReferenceExpression
                     = querySourceTracingExpressionVisitor
@@ -250,10 +260,9 @@ namespace Microsoft.Data.Entity.Query
                             Strings.LogIncludingNavigation);
 
                     IncludeNavigations(
-                        includeSpecification.QuerySource,
+                        includeSpecification,
                         _expression.Type.GetSequenceType(),
                         accessorLambda,
-                        includeSpecification.NavigationPath,
                         QuerySourceRequiresTracking(includeSpecification.QuerySource));
 
                     QueryCompilationContext
@@ -292,10 +301,9 @@ namespace Microsoft.Data.Entity.Query
         }
 
         protected virtual void IncludeNavigations(
-            [NotNull] IQuerySource querySource,
+            [NotNull] IncludeSpecification includeSpecification,
             [NotNull] Type resultType,
             [NotNull] LambdaExpression accessorLambda,
-            [NotNull] IReadOnlyList<INavigation> navigationPath,
             bool querySourceRequiresTracking)
         {
             // template method
--- a/src/EntityFramework.InMemory/Query/InMemoryQueryModelVisitor.cs
+++ b/src/EntityFramework.InMemory/Query/InMemoryQueryModelVisitor.cs
@@ -30,19 +30,18 @@ namespace Microsoft.Data.Entity.InMemory.Query
         }
 
         protected override void IncludeNavigations(
-            IQuerySource querySource,
+            IncludeSpecification includeSpecification,
             Type resultType,
             LambdaExpression accessorLambda,
-            IReadOnlyList<INavigation> navigationPath,
             bool querySourceRequiresTracking)
         {
-            Check.NotNull(querySource, nameof(querySource));
+            Check.NotNull(includeSpecification, nameof(includeSpecification));
             Check.NotNull(resultType, nameof(resultType));
             Check.NotNull(accessorLambda, nameof(accessorLambda));
-            Check.NotNull(navigationPath, nameof(navigationPath));
 
             var primaryKeyParameter = Expression.Parameter(typeof(EntityKey));
             var relatedKeyFactoryParameter = Expression.Parameter(typeof(Func<ValueBuffer, EntityKey>));
+            var navigationPath = includeSpecification.NavigationPath;
 
             Expression
                 = Expression.Call(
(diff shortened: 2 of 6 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
```diff
@@ -148,6 +149,9 @@ var materializer
 
                     targetTableExpression = joinedTableExpression;
 
+                    var readerIndex = _readerIndeces[0];
```
maumar: shall I use Queue instead?

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -118,11 +118,47 @@ var compositePredicateVisitor
...
+                    }
+                    else
+                    {
+                        if (openedNewReader)
+                        {
+                            indeces.Add(openedReaderCount);
+                        }
+                        else
+                        {
+                            indeces.Add(0);
```
divega: It is spelled indices, right? :sweat_smile: :

maumar: ESL, will fix :)

anpete: "indexes" is fine too.

src/EntityFramework.InMemory/Query/InMemoryQueryModelVisitor.cs
```diff
@@ -29,11 +29,17 @@ protected override ExpressionVisitor CreateQueryingExpressionVisitor(IQuerySourc
             return new InMemoryEntityQueryableExpressionVisitor(this, querySource);
         }
 
+        protected override Dictionary<IncludeSpecification, List<int>> BuildNavigationIndexMap(List<IncludeSpecification> includeSpecifications)
```
anpete: Return null in base and remove this.

anpete: Actually, can we make this relational only?

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -226,6 +226,7 @@ var navigationPath
             var querySourceTracingExpressionVisitor
                 = new QuerySourceTracingExpressionVisitor();
 
+            var navigationIndexMap = BuildNavigationIndexMap(includeSpecifications);
```
anpete: Re: making this relational only: Extract Method on the foreach loop. Call it "IncludeNavigations" and make it protected virtual, then override it in relational to compute the indexes before calling base.

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -118,11 +118,47 @@ var compositePredicateVisitor
             }
         }
 
+        protected override Dictionary<IncludeSpecification, List<int>> BuildNavigationIndexMap(List<IncludeSpecification> includeSpecifications)
+        {
+            var openedReaderCount = 0;
+            var navigationIndexMap = new Dictionary<IncludeSpecification, List<int>>();
+            foreach (var includeSpecification in includeSpecifications
```
anpete: Move the sort in IncludeNavigations up, so that you don't need to do it here.

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
anpete: Or a dictionary of INav -> index?

maumar: This was my first thought, however there can be multiple instances of same INav pointing to different indexes. This could be done with the ordered list, and by removing the elements accordingly in the IncludeVisitor, but I don't like this idea, since it makes IncludeVisitor instances share state and become inter-dependent.

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
maumar: went with the indexes

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -221,13 +221,23 @@ var navigationPath
...
                         })
+                    .OrderBy(a => a.NavigationPath.First().PointsToPrincipal())
                     .ToList();
 
+            IncludeNavigations(queryModel, includeSpecifications);
+        }
+
+        protected virtual void IncludeNavigations(
+            [NotNull] QueryModel queryModel,
+            [NotNull] IEnumerable<IncludeSpecification> includeSpecifications)
```
anpete: IReadOnlyCollection

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -292,10 +301,9 @@ let entityType
         }
 
         protected virtual void IncludeNavigations(
-            [NotNull] IQuerySource querySource,
+            [NotNull] IncludeSpecification includeSpecification,
```
anpete: Nice.

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
```diff
@@ -29,6 +30,7 @@ public class IncludeExpressionVisitor : ExpressionVisitorBase
             [NotNull] IQuerySource querySource,
             [NotNull] IReadOnlyList<INavigation> navigationPath,
             [NotNull] RelationalQueryCompilationContext queryCompilationContext,
+            [NotNull] List<int> readerIndexes,
```
anpete: IReadOnlyList

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -119,22 +121,65 @@ var compositePredicateVisitor
...
+                foreach (var navigation in includeSpecification.NavigationPath)
+                {
+                    if (navigation.IsCollection())
+                    {
+                        openedNewReader = true;
+                        openedReaderCount++;
+                    }
+                    else
+                    {
+                        if (openedNewReader)
```
anpete: Ternary here

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -119,22 +121,65 @@ var compositePredicateVisitor
...
+
+            base.IncludeNavigations(queryModel, includeSpecifications);
+        }
+
+        private Dictionary<IncludeSpecification, List<int>> BuildNavigationIndexMap(
+            IEnumerable<IncludeSpecification> includeSpecifications)
+        {
+            var openedReaderCount = 0;
+            var navigationIndexMap = new Dictionary<IncludeSpecification, List<int>>();
+            foreach (var includeSpecification in includeSpecifications.Reverse())
```
anpete: Nice to add a line break before the foreach loops

anpete: Consider replacing the imperative loops with a LINQ query.

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
```diff
@@ -148,6 +149,9 @@ var materializer
 
                     targetTableExpression = joinedTableExpression;
 
+                    var readerIndex = _readerIndexes[0];
```
anpete: Can you change to loop above to use a counter and then index here?

anpete:
:shipit:

Issue #2425, opened 2015-06-19
BuiltInDataTypesTest methods fail


ErikEJ:
Can_query_using_any_data_type
 and 
Can_query_using_any_nullable_data_type() 

fail, as min value for SQLCE datetime is 1753!

ErikEJ:
Fixed, or will do a pr if not...

ErikEJ:
PR now merged, thanks

PR #2475, opened 2015-06-26
No year before 1753!

ErikEJ:
Fixes issue #2425


ajcvickers:
@ErikEJ The problem with this change is that I explicitly did use a default DateTime object to make sure it works, at least when it can work. This is because it didn't work out-of-the-box on SQL Server with the old stack and this was a pain point for people. So I think we need a change here that allows the test to continue to verify that this works when it can, but with a provider opt-out if it just isn't possible to make it work for a given provider.

ErikEJ:
@ajcvickers  Ahh, makes sense. How about a property to set the value, and leave it at the default if not set?

ajcvickers:
@ErikEJ Sounds good.

ErikEJ:
@ajcvickers PR updated

ajcvickers:
Thanks Erik. Looks like you missed the override in the in-memory provider, but I added it and merged.

ErikEJ:
@ajcvickers Thanks a lot!

PR #2480, opened 2015-06-26
Remove 'help' subcommand from command-line tool

ChengTian:
Reacting to aspnet/dnx#1116, which removes an named argument.

This should fix CI build.


bricelam:
Oh lame. We should have "fixed" it, not removed it...

ChengTian:
:smiley:

PR #2481, opened 2015-06-26
Make it possible for providers to change IParameterNameGeneratorFactory service

ajcvickers:
Per request from Oracle to generate Oracle-specific parameter names. To use, add an override something like this in your XxxDatabaseProviderServices class:

```
public override IParameterNameGeneratorFactory ParameterNameGeneratorFactory => GetService<XxxParameterNameGeneratorFactory>();
```

And register XxxParameterNameGeneratorFactory in your AddXxx method.

### DIFF
--- a/src/EntityFramework.Relational/RelationalDatabaseProviderServices.cs
+++ b/src/EntityFramework.Relational/RelationalDatabaseProviderServices.cs
@@ -27,13 +27,15 @@ namespace Microsoft.Data.Entity.Relational
 
         public override IQueryContextFactory QueryContextFactory => GetService<RelationalQueryContextFactory>();
         public override IValueGeneratorSelector ValueGeneratorSelector => GetService<RelationalValueGeneratorSelector>();
+        public override IModelValidator ModelValidator => GetService<RelationalModelValidator>();
+
         public virtual IRelationalTypeMapper TypeMapper => GetService<RelationalTypeMapper>();
         public virtual IMigrationAnnotationProvider MigrationAnnotationProvider => GetService<MigrationAnnotationProvider>();
         public virtual IBatchExecutor BatchExecutor => GetService<BatchExecutor>();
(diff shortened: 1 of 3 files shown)


src/EntityFramework.Relational/RelationalDatabaseProviderServices.cs
```diff
@@ -27,13 +27,15 @@ protected RelationalDatabaseProviderServices([NotNull] IServiceProvider services
...
         public override IValueGeneratorSelector ValueGeneratorSelector => GetService<RelationalValueGeneratorSelector>();
+        public override IModelValidator ModelValidator => GetService<RelationalModelValidator>();
+
         public virtual IRelationalTypeMapper TypeMapper => GetService<RelationalTypeMapper>();
         public virtual IMigrationAnnotationProvider MigrationAnnotationProvider => GetService<MigrationAnnotationProvider>();
         public virtual IBatchExecutor BatchExecutor => GetService<BatchExecutor>();
         public virtual IRelationalValueBufferFactoryFactory ValueBufferFactoryFactory => GetService<TypedValueBufferFactoryFactory>();
         public virtual ICommandBatchPreparer CommandBatchPreparer => GetService<CommandBatchPreparer>();
         public virtual ISqlStatementExecutor SqlStatementExecutor => GetService<SqlStatementExecutor>();
-        public override IModelValidator ModelValidator => GetService<RelationalModelValidator>();
```
divega: Is this removal related?

divega:
:shipit

src/EntityFramework.Relational/RelationalDatabaseProviderServices.cs
ajcvickers: It's not removed. It's just moved up a few lines to be with all the other overrides.

divega: D'oh! missed it.

PR #2500, opened 2015-06-29
Fix bugs in DropColumn in SQLite

natemcmaster:
Fixes limitations in the original implementation of DropColumn for SQLite which didn't take into consideration multiple drop columns, or operations made redundant by the table-rebuild.

Also, fixed a few minor bugs in csharp generation for migrations.

### DIFF
--- a/src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
+++ b/src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Data.Entity.Sqlite.Migrations
     public class SqliteOperationTransformer
     {
         private readonly IModelDiffer _differ;
+        private readonly ISet<string> _tableRebuilds = new HashSet<string>();
 
         public SqliteOperationTransformer([NotNull] IModelDiffer differ)
         {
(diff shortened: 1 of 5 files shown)


src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
```diff
@@ -22,38 +23,66 @@ public SqliteOperationTransformer([NotNull] IModelDiffer differ)
...
+            => FilterRedudantOperation(operation.Table, operation);
+
+        private IList<MigrationOperation> TransformOperation(CreateIndexOperation operation, IModel model)
+            => FilterRedudantOperation(operation.Table, operation);
+
+        private IList<MigrationOperation> TransformOperation(CreateTableOperation operation, IModel model)
+            => FilterRedudantOperation(operation.Name, operation);
+
+        private IList<MigrationOperation> FilterRedudantOperation(string tableName, MigrationOperation operation)
+            => _tableRebuilds.Contains(tableName) ? null : new[] { operation };
```
bricelam: So redundant operations *before* the rebuild will execute normally?

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
```diff
@@ -64,6 +93,10 @@ private IList<MigrationOperation> TransformOperation(DropColumnOperation operati
                     Name = tempTableName
                 }
             };
+
+            rebuildOperations.AddRange(differences.Where(o => (o as CreateIndexOperation)?.Table == tableName));
```
bricelam: It's also possible that foreign keys are created out-of-line.

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
```diff
@@ -75,7 +108,12 @@ private IList<MigrationOperation> TransformOperation(DropColumnOperation operati
             var finalOperations = new List<MigrationOperation>();
             foreach (var operation in operations)
             {
-                finalOperations.AddRange(TransformOperation((dynamic)operation, model));
+                var newOps = TransformOperation((dynamic)operation, model);
```
bricelam: BTW, we'll probably have to remove a lot of this dynamic code for UWP/.NET Native

bricelam:
:shipit:

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
natemcmaster: Remove operations that are obviated by the rebuild.

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
natemcmaster: :disappointed: dang. dynamic dispatch looks so clean.

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
natemcmaster: In SQLite? I thought these can only be added in the `create table` statement.

bricelam: Yes, the model differ will pull them out if there are schema-level cycles. You're right; SQLite should probably push them back into the create table operation if possible.

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
bricelam: I was sad too... :panda_face:

PR #2503, opened 2015-06-29
Relational namespace changes

ajcvickers:
Per API review decision, this change mirrors core namespaces in the relational types and removes "relational" from the namespace path. Also, Strings are moved into the Internal namespace.


anpete:
:shipit:

PR #2510, opened 2015-06-30
Migrations: Test renames

### LINKED ISSUE #1907: Migrations: Handle transitive renames
The `ModelDiffer` needs more work when it comes to transitive renames. For example, currently renaming a column or table will will result in change operations for everything that touches that object like keys, indexes, etc.

### LINKED ISSUE #2509: Migrations generate wrong order for DropForeignKey and DropPrimaryKey
I changed the name of a column in a small database (on SQL Server 2014) I'm using to test EF7 and used "Add-Migration" to generate the code.

Running "Apply-Migration" I received the following error:

"The constraint 'PK_Address' is being referenced by table 'Meter', foreign key constraint 'FK_Meter_Address_AddressAddressId'"

Looking at the code that was generated I noticed that in the "Up" method the call to "DropPrimaryKey" came before the calls to "DropForeignKey".

    migration.DropPrimaryKey(name: "PK_Address", table: "Address");
    migration.DropForeignKey(name: "FK_Meter_Address_AddressAddressId", table: "Meter");
            
Dropping foreign keys before the primary key worked as expected.

IDE: VS 2015 RC
EF: 7.0.0-beta4
Project types: Console & Class Library
Windows 8.1

Not sure what other information may prove useful.

bricelam:
Fixes #1907, fixes #2509

### DIFF
--- a/src/EntityFramework.Relational/EntityFramework.Relational.csproj
+++ b/src/EntityFramework.Relational/EntityFramework.Relational.csproj
@@ -67,6 +67,7 @@
     <Compile Include="Metadata\Conventions\Internal\RelationalConventionSetBuilder.cs" />
     <Compile Include="Metadata\Conventions\Internal\RelationalTableAttributeConvention.cs" />
     <Compile Include="Metadata\IRelationalMetadataExtensionProvider.cs" />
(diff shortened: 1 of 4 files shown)


ajcvickers:
:shipit:

PR #2518, opened 2015-07-01
Include EF version in model

### LINKED ISSUE #1010: EntityFramework version should be included in model snapshot
The migrations pipeline needs to know what version of the product was used to generate a model snapshot so it can compensate for changes that are introduced to model generation

smitpatel:
resolves #1010


bricelam:
:shipit:

PR #2519, opened 2015-07-01
Migrations: Handle adding required properties

### LINKED ISSUE #1275: Migrations: Default value not scaffolded for non nullable fields causing error
Consider a database with tables and user data. Now if I change my model to add a property, say `DateTime` which is non nullable. Running migrations generates the necessary code but doesn't set a default value for the column. This causes to throw an ugly EF exception when applying the migrations that there is no default value set. 

This seems to have worked in EF 6.0 where a default value was set for the column

bricelam:
Fixes #1275


ajcvickers:
:shipit:

Issue #1863, opened 2015-03-19
Add support for execution of a raw command on the data store


mikary:
Provide a simple mechanism for executing a raw command (SQL statement) that does not return a value

mikary:
@rowanmiller is this a feature we would expect to have parody with ExecuteStoreSql from EF6? If so should we consider moving it into the 7.0.0 milestone?

rowanmiller:
Clearing milestone so that we can address in triage

divega:
We missed creating the async version of ExecuteSqlCommand. Reopening to track that.

PR #2520, opened 2015-07-02
Add ExecuteStoreSql extension method to RelationalDatabaseFacade

mikary:
Adding new extension for #1863


src/EntityFramework.Relational/Infrastructure/RelationalStoreSqlExecutor.cs
```diff
@@ -0,0 +1,55 @@
...
+            [NotNull] ISqlStatementExecutor statementExecutor,
+            [NotNull] IRelationalConnection connection,
+            [NotNull] IRelationalTypeMapper typeMapper)
+        {
+            _statementExecutor = statementExecutor;
+            _connection = connection;
+            _typeMapper = typeMapper;
+        }
+
+        public virtual void ExecuteStoreSql([NotNull] string sql, [NotNull] params object[] parameters)
```
divega: This is the equivalent to dbContext.Database.ExecuteSqlCommand in EF6. We may come up with a better name but I would suggest we keep it the same until we can discuss it in API review.

divega: (of course I was trying to say that about the actual method in RelationalDatabaseFacadeExtensions, but regardless, we should be consistent).

mikary:
Could I have you take a look at this one @AndriySvyryd ?

AndriySvyryd:
:shipit:

Issue #2365, opened 2015-06-11
ArgumentNullException when querying with nullable byte, short, or enum parameters


ajcvickers:
See "Can_query_with_null_parameters_using_any_nullable_data_type" to reproduce.

Stack trace:

>
Test 'Microsoft.Data.Entity.SqlServer.FunctionalTests.BuiltInDataTypesSqlServerTest.Can_query_with_null_parameters_using_any_nullable_data_type' failed:
	System.ArgumentNullException : Value cannot be null.
Parameter name: value
	C:\aspnet\EntityFramework\src\Shared\Check.cs(24,0): at Microsoft.Data.Entity.Utilities.Check.NotNull[T](T value, String parameterName)
	Query\CommandParameter.cs(18,0): at Microsoft.Data.Entity.Relational.Query.CommandParameter..ctor(String name, Object value, RelationalTypeMapping typeMapping)
	Query\Sql\DefaultQuerySqlGenerator.cs(841,0): at Microsoft.Data.Entity.Relational.Query.Sql.DefaultQuerySqlGenerator.VisitParameter(ParameterExpression parameterExpression)
	at System.Linq.Expressions.ParameterExpression.Accept(ExpressionVisitor visitor)
	at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
	Query\Sql\DefaultQuerySqlGenerator.cs(669,0): at Microsoft.Data.Entity.Relational.Query.Sql.DefaultQuerySqlGenerator.VisitBinary(BinaryExpression binaryExpression)
	at System.Linq.Expressions.BinaryExpression.Accept(ExpressionVisitor visitor)
	at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
	Query\Sql\DefaultQuerySqlGenerator.cs(669,0): at Microsoft.Data.Entity.Relational.Query.Sql.DefaultQuerySqlGenerator.VisitBinary(BinaryExpression binaryExpression)
	at System.Linq.Expressions.BinaryExpression.Accept(ExpressionVisitor visitor)
	at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
	Query\Sql\DefaultQuerySqlGenerator.cs(154,0): at Microsoft.Data.Entity.Relational.Query.Sql.DefaultQuerySqlGenerator.VisitSelect(SelectExpression selectExpression)
[...]

PR #2523, opened 2015-07-02
Fix to #2365 - ArgumentNullException when querying with nullable byte, short, or enum parameters

maumar:
Problem was that in some cases of column == variable, compiler wraps the column expression into a Convert. As later stage in the query pipeline we try to recognize the pattern where column == null_valued_parameter and convert it to IsNull(column). However, we were not detecting it properly for columns wrapped in Convert. 

Fix is to unwrap converts in the pattern detection logic. We don't really need the information about the converted type, since the result expression is either the original (we didn't find the right pattern) or IsNull(column), that returns save boolean value regardless of the cast.

### DIFF
--- a/src/EntityFramework.Relational/Extensions/ExpressionExtensions.cs
+++ b/src/EntityFramework.Relational/Extensions/ExpressionExtensions.cs
@@ -33,19 +33,13 @@ namespace System.Linq.Expressions
 
         public static bool IsSimpleExpression([NotNull] this Expression expression)
         {
-            var unaryExpression = expression as UnaryExpression;
+            var unwrappedExpression = expression.RemoveConvert();
 
-            if (unaryExpression != null
-                && unaryExpression.NodeType == ExpressionType.Convert)
-            {
-                return IsSimpleExpression(unaryExpression.Operand);
-            }
-
-            return expression is ConstantExpression
-                   || expression is ColumnExpression
-                   || expression is ParameterExpression
-                   || expression is LiteralExpression
-                   || expression.IsAliasWithColumnExpression();
+            return unwrappedExpression is ConstantExpression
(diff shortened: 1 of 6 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Extensions/ExpressionExtensions.cs
```diff
@@ -47,5 +47,15 @@ public static bool IsSimpleExpression([NotNull] this Expression expression)
                    || expression is LiteralExpression
                    || expression.IsAliasWithColumnExpression();
         }
+
+        public static Expression UnwrapConvert([NotNull] this Expression expression)
```
anpete: Validate args

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -941,8 +941,8 @@ var parameter
                         && parameterValue == null)
                     {
                         var columnExpression
-                            = expression.Left.TryGetColumnExpression()
-                              ?? expression.Right.TryGetColumnExpression();
+                            = expression.Left.UnwrapConvert().TryGetColumnExpression()
```
anpete: Any other places we do this explicitly and could change to use the extension method?

anpete:
:shipit:

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
maumar: there is one more place that I found (in the null semantics) but it actually needs to know the conversion type of the original. Will make another sweep.

PR #2525, opened 2015-07-02
[SQLite] Always use foreign keys

natemcmaster:
Always calls `PRAMGA foreign_keys = ON;` after opening a new connection.

### DIFF
--- a/src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
+++ b/src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
@@ -2,6 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Data.Common;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Storage;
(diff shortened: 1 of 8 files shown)


src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
```diff
@@ -20,5 +22,25 @@ public SqliteDatabaseConnection([NotNull] IDbContextOptions options, [NotNull] I
...
+            base.Open();
+            EnableForeignKeys();
+        }
+        public override async Task OpenAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            await base.OpenAsync(cancellationToken);
+            EnableForeignKeys();
+        }
+
+        private void EnableForeignKeys()
```
bricelam: Need a switch to turn this off too.

bricelam: Possibly similar to SQL Server's [SuppressAmbientTransactionWarning](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.SqlServer/Extensions/SqlServerDbContextOptionsBuilder.cs#L52).

natemcmaster:
@bricelam updated. Can be switched off by using `.UseSqlite().SuppressForeignKeysEnforcement()`

ErikEJ:
:thumbsup:

src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
```diff
@@ -12,13 +15,45 @@ namespace Microsoft.Data.Entity.Sqlite
 {
     public class SqliteDatabaseConnection : RelationalConnection
     {
+        private readonly bool _enforceForeignKeys = true;
```
bricelam: Is caching this an issue? Would the user want to change it over the lifetime of this instance? E.g. turn it off for seed data, but turn it on for the rest of the app's operations.

test/EntityFramework.Sqlite.FunctionalTests/ForeignKeyTest.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public ForeignKeyTest()
+        {
+            _testStore = SqliteTestStore.CreateScratch();
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void It_enforces_foreign_key(bool suppress)
```
bricelam: These tests could be simplified to just assert that `PRAGMA FOREIGN_KEYS;` returns `0` or `1`.

src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
natemcmaster: This is cached, but only affects connection EF opens. For someone seeding data, they could turn it off by executing a pragma statement.

bricelam: Maybe SuppressForeignKeysEnforcement is YAGNI. We should look at what it's like to suppress it globally and one-off and discuss with others.

bricelam: Regardless, it looks like we don't expose a way of changing DbContextOptions over the lifetime of an instance, so caching is fine for now.

bricelam:
:shipit:

PR #2528, opened 2015-07-02
Scenarios for value generation on SQL Server

ajcvickers:
Organized as documentation-style tests that clearly show what happen in different scenarios. I tried refactoring to be provider-agnostic but they lost a lot of value. Underlying functionality is covered by more specific lower level tests in provider agnostic ways.

Also included change to use DefaultValueSql to allow server-side GUID keys to be used. See #

Changes from today's meeting not yet implemented, but TODOs in the code.


anpete:
:shipit:

PR #2543, opened 2015-07-06
Remove EntityFramework meta-package

bricelam:


natemcmaster:
How are we going to redirect people to the new packages they should use?

smitpatel:
Since we modifying makefile here, want to ask what is exactly being generated at https://github.com/aspnet/EntityFramework/blob/dev/makefile.shade#L75

bricelam:
@natemcmaster All [the walkthroughs](http://ef.readthedocs.org/) instruct people to install the provider package. E.g. `EntityFramework7.SqlServer`, `EntityFramework7.Sqlite`, etc. We decided to try this and wait for feedback as part of the [NuGet Package Naming](https://github.com/aspnet/EntityFramework/wiki/NuGet-Package-Naming) discussions.

bricelam:
@smitpatel It instructs NuGet that the package can be installed into a Universal Windows App even though there is no runtime assembly to reference.

natemcmaster:
Readme.html has been removed. Are we going to put this into the provider packages?

rowanmiller:
The readme was only important when folks might be installing EntityFramework expecting to get EF6... this won't be the case now that the packages don't overlap. BTW this is only the first step in a set of package name updates to help with major versioning.

natemcmaster:
LGTM.

PR #2548, opened 2015-07-06
Replace StoreGeneratedPattern enum fluent API with separate methods

ajcvickers:
Per design meeting decision. Enum retained for core metadata but renamed to align with fluent API. GenerateValueOnAdd renamed to RequiresValueGenerator to indicate that this is a property that needs a ValueGenerator implementation of some sort.


smitpatel:
:shipit:

Issue #2057, opened 2015-04-24
Sql query with enum produces a where clause with the string value instead of the underlying (int) type


rmja:
I have the following enum definition:

    public enum Discriminator
    {
        Something = 1
    }

The sql produced when filtering using the enum like this:

    _db.Addresses.Where(x => x.Discriminator == Discriminator.Something).ToList();

is:

    SELECT [x].[Id], [x].[Discriminator]
    FROM [network].[Address] AS [x]
    WHERE [x].[Discriminator] = Something

It should instead be (according to #1381):

    SELECT [x].[Id], [x].[Discriminator]
    FROM [network].[Address] AS [x]
    WHERE [x].[Discriminator] = 1

As a small fix I have changed the `GenerateLiteral()` in `DefaultSqlQueryGenerator.cs` to:

        protected virtual string GenerateLiteral([NotNull] object value)
        {
			if (value is Enum)
			{
				return string.Format(CultureInfo.InvariantCulture, "{0:d}", value);
			}
			else
			{
				return string.Format(CultureInfo.InvariantCulture, "{0}", value);
			}
        }

divega:
@natemcmaster want to give this one a try?

natemcmaster:
@divega Sure. I'll take a wack at it.

natemcmaster:
@rmja Can you provide a repro project? I was unable to get the error.

I created [this sample project](https://gist.github.com/natemcmaster/039c66671b18894c64f7) to reproduce the error and this is the SQL generated:

```sql
@__p_0: 1

SELECT [x].[Id], [x].[Discriminator]
FROM [Address] AS [x]
WHERE [x].[Discriminator] = @__p_0
```

rmja:
Hi @natemcmaster. This bug was previously seen for simple queries like the one I describe in the original post, but since then, it has somehow been fixed. However, when I use the Enum as a discriminator for a polymorphic setup, then the error is still present. I have made a reproducing example here: https://gist.github.com/rmja/691506a58907c07f9ce2

My setup is slightly different, as it uses an ef base that I have modified to allow for discriminators of type object instead of string. Here is the patch:

```
$ git diff --cached
diff --git a/src/EntityFramework.Relational/Metadata/IRelationalEntityTypeExtensions.cs b/src/EntityFramework.Relational/Metadata/IRelationalEntityTypeExtensions.cs
index 80fb2cd..b1cf1de 100644
[...]

PR #2549, opened 2015-07-06
Fix #2444 - change data type of DiscriminatorValue to object.

### LINKED ISSUE #2444: Make DiscriminatorValue object instead of string
The DiscriminatorValue is currently a string, but should be object to allow for e.g. enums to discriminate derived entity types. See e.g. #2057 where a naiive try results in invalid SQL for enum.

mikary:
Fix is as described in #2057, Query uses the `DiscriminatorValue` in a `ConstantExpression` and the `DefaultQuerySqlGenerator` needs to translate `Enum`s accordingly.

The SQL for the provided example now generates as:
```
@__p_0: 1
SELECT [x].[Id], [x].[Discriminator]
FROM [Address] AS [x]
WHERE ([x].[Discriminator] IN (2, 1) AND [x].[Discriminator] = @__p_0)
```

@anpete

### DIFF
--- a/src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -873,6 +873,9 @@ namespace Microsoft.Data.Entity.Query.Sql
         protected virtual string GenerateLiteral([NotNull] object value)
             => string.Format(CultureInfo.InvariantCulture, "{0}", value);
 
+        protected virtual string GenerateLiteral([NotNull] Enum value)
+            => string.Format(CultureInfo.InvariantCulture, "{0:d}", value);
+
         protected virtual string GenerateLiteral(int value)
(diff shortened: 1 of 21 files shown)


src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -861,6 +861,9 @@ protected override Exception CreateUnhandledItemException<T>(T unhandledItem, st
         protected virtual string GenerateLiteral([NotNull] object value)
             => string.Format(CultureInfo.InvariantCulture, "{0}", value);
 
+        protected virtual string GenerateLiteral([NotNull] Enum value)
```
anpete: Probably want a test for this - doesn't have to be inheritance related.

mikary:
Added inheritance tests, appears to be the only way to get coverage on the SQL generation change.

anpete:
:shipit:

Issue #329, opened 2014-06-17
Sqlite Migrations: Table rebuilds


bricelam:
In SQLite, the following operations will require rebuilding the table - https://docs.microsoft.com/en-us/ef/core/providers/sqlite/limitations.

Other databases also have a smaller set of limitations that require a table rebuild (i.e. changing Identity on a column in SQL Server).

bricelam:
Re-opening. I've regressed this.

natemcmaster:
I've discovered a limitation that makes me wonder if the complication of rebuilds will exceed their value.

The limitation `PRAGMA foreign_keys=on/off;` does not work within a transaction.

Consider the following setup:

Table "Parent" - no FK's
Table "Children" - references "Parent" via a FK

1. Scenario 1 - renaming  "Parent" to "NewTableName". If the `alter table` statement executes with `PRAGMA foreign_keys=ON;`, then SQLite will handle updating the "Children" reference automatically. Otherwise, it will break the FK relationship.
2. Scenario 2 - rebuilding "Parent" ( to imitate drop column or some other operation not natively available). This must be executed with `PRAGMA foreign_keys=OFF;` otherwise the rebuild looks to SQLite like Scenario 1 and "Children" will end up reference a table which is deleted in the rebuild.

Scenario 1 and 2 could be resolved by just executing the proper `PRAGMA foreign_keys` before renaming/rebuilding if not for this limitation,
> This pragma is a no-op within a transaction; foreign key constraint enforcement may only be enabled or disabled when there is no pending BEGIN or SAVEPOINT.

This means we cannot execute an entire migration within one transaction, but need to break SQLite migrations into multiple transactions. Reverting to the beginning of a migration becomes impossible.

ErikEJ:
You are entering a World of Pain :smiley:

natemcmaster:
Moving to backlog to work on high-priority items. 
Consider again after #1141 and #2558.

bricelam:
@natemcmaster Did you try the approach recommended in the SQLite documentation: [Making Other Kinds Of Table Schema Changes](http://sqlite.org/lang_altertable.html#otheralter)? Triggers will be lost. Views may be broken. Indexes would be re-created from the model snapshot. But otherwise it looks like it'll work.

bricelam:
Making this a pri1 issue given the new information.

natemcmaster:
[...]

PR #2550, opened 2015-07-06
[SQLite] Enable column renames in migrations

natemcmaster:
Enables column renames with a table rebuild.

Also, improve implementation of transforms and cleanup testing.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -54,6 +54,7 @@ var sqliteTestDir='test/EntityFramework.Sqlite.FunctionalTests/bin/${E("Configur
     var sourceDir='${Directory.EnumerateDirectories(packagesDir, "Microsoft.Data.Sqlite/*").OrderByDescending(d => d).First() + "/redist/x86"}'
     copy outputDir='${sqliteTestDir}'
     copy outputDir='test/EntityFramework.CrossStore.FunctionalTests/bin/${E("Configuration")}/x86'
+    copy outputDir='test/EntityFramework.Sqlite.Tests/bin/${E("Configuration")}/x86'
 
 #nuspec-pack target="compile"
     var DNX_BUILD_VERSION='${E("DNX_BUILD_VERSION")}'
--- a/src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
+++ b/src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
@@ -7,6 +7,7 @@ using JetBrains.Annotations;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Migrations.Infrastructure;
 using Microsoft.Data.Entity.Migrations.Operations;
+using Microsoft.Data.Entity.Sqlite.Migrations.Operations;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Sqlite.Migrations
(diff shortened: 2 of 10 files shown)


makefile.shade
```diff
@@ -54,6 +54,7 @@ var sqliteTestDir='test/EntityFramework.Sqlite.FunctionalTests/bin/${E("Configur
     var sourceDir='${Directory.EnumerateDirectories(packagesDir, "Microsoft.Data.Sqlite/*").OrderByDescending(d => d).First() + "/redist/x86"}'
     copy outputDir='${sqliteTestDir}'
     copy outputDir='test/EntityFramework.CrossStore.FunctionalTests/bin/${E("Configuration")}/x86'
+    copy outputDir='test/EntityFramework.Sqlite.Tests/bin/${E("Configuration")}/x86'
```
bricelam: If tests need a database, they may belong in the functional test project.

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
```diff
@@ -23,20 +24,41 @@ public SqliteOperationTransformer([NotNull] IModelDiffer differ)
...
+            TableRebuildOperation rebuild;
+            if (!_tableRebuilds.TryGetValue(operation.Table, out rebuild))
+            {
+                return new MigrationOperation[] { operation };
+            }
+            var move = rebuild.Operations.OfType<MoveDataOperation>().First();
+            if (move.ColumnMapping.ContainsKey(operation.Name))
+            {
+                move.ColumnMapping.Remove(operation.Name);
+            }
```
bricelam: This feels wrong. What is it doing? Consider the following.
```C#
mb.RenameColumn("Name", table: "Person", newName: "FirstName");
mb.AddColumn("Name", table: "Person");
mb.Sql("UPDATE Person SET Name = FirstName || ' Smith'");
```

But the code here would remove the mapping from Name->FirstName during the table rebuild?

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
```diff
@@ -48,55 +70,60 @@ private IList<MigrationOperation> TransformOperation(CreateTableOperation operat
...
+                        Name = tempTableName
+                    }
                 }
             };
 
-            rebuildOperations.AddRange(differences.Where(o => (o as CreateIndexOperation)?.Table == tableName));
-
-            return rebuildOperations;
+            rebuild.Operations.AddRange(differences
+                .OfType<CreateIndexOperation>()
```
bricelam: Is there a bug on the foreign keys as mentioned in [#2500](https://github.com/aspnet/EntityFramework/pull/2500#discussion_r33796074)?

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
natemcmaster: This code is adjusting the list of columns that need to me moved. New columns do not need to be moved from the old table.

This specific example, however, is a difficult edge case. If the code snippet didn't include the Sql operation, and were only
```c#
mb.RenameColumn("Name", table: "Person", newName: "FirstName");
mb.AddColumn<string>("Name", table: "Person");
```
then the user might be expecting a new column "Name" to be empty and the old Name column to be moved to FirstName. In SQL Server, this would be the case. But in SQLite, these two operations must be in different migrations for this to work (with the current code).

src/EntityFramework.Sqlite/Migrations/SqliteOperationTransformer.cs
natemcmaster: Yes, foreign keys are going to be a little messy. Let's hold a more thorough discussion about what level of rebuilds we want to support.

natemcmaster:
See https://github.com/aspnet/EntityFramework/issues/329#issuecomment-119293848

natemcmaster:
We will be putting this on the backlog to focus on higher priority items. 

Track on #329

PR #2551, opened 2015-07-07
Allow multiple providers for reverse engineering.

lajones:
See issue #830. This should free the way up to allow for RevEng providers beyond our standard SQL Server one.

Also it updates the way we locate the design-time provider to instead use an attribute on the runtime
provider.

### DIFF
--- a/src/EntityFramework.Commands/DatabaseTool.cs
+++ b/src/EntityFramework.Commands/DatabaseTool.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
@@ -14,6 +15,7 @@ using Microsoft.Data.Entity.Relational.Design.ReverseEngineering;
 using Microsoft.Data.Entity.Relational.Design.Templating;
 using Microsoft.Data.Entity.Relational.Design.Templating.Compilation;
 using Microsoft.Data.Entity.Relational.Design.Utilities;
+using Microsoft.Data.Entity.Storage;
 using Microsoft.Data.Entity.Utilities;
 using Microsoft.Framework.Logging;
 
@@ -21,8 +23,6 @@ namespace Microsoft.Data.Entity.Commands
 {
     public class DatabaseTool
     {
-        public static readonly string _defaultReverseEngineeringProviderAssembly = "EntityFramework.SqlServer.Design";
-
         private readonly ServiceProvider _serviceProvider;
         private readonly LazyRef<ILogger> _logger;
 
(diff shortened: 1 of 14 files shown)


ErikEJ:
@lajones So I could simply include the RevEng implementation in the base provider, rather than requiring an new NuGet package to be installed?

lajones:
@ErikEJ Yep. If you just use TypeName in the ProviderDesignTimeServicesAttribute and do not set AssemblyName it will assume the design-time type it needs is in the same assembly as the runtime provider (i.e. what you're calling the base provider). I thought you might like that :-)

The downside of course is that your runtime assembly will be that much bigger with code you don't actually need at runtime - but in your case that may be only a small increase.

src/EntityFramework.Commands/DatabaseTool.cs
```diff
@@ -47,26 +47,46 @@ public class DatabaseTool
...
             Check.NotEmpty(connectionString, nameof(connectionString));
             Check.NotEmpty(rootNamespace, nameof(rootNamespace));
             Check.NotEmpty(projectDir, nameof(projectDir));
 
-            var assembly = Assembly.Load(new AssemblyName(providerAssemblyName));
-            if (assembly == null)
+            Assembly runtimeProviderAssembly = null;
+            try
+            {
+                runtimeProviderAssembly = Assembly.Load(new AssemblyName(runtimeProviderAssemblyName));
```
bricelam: We may have to do something with paths for the NuGet commands--MSBuild won't copy a referenced assembly to the bin directory unless it's used during compilation.

bricelam: Alternatively, I suppose the NuGet commands could copy it.

src/EntityFramework.Commands/DatabaseTool.cs
```diff
@@ -75,5 +95,44 @@ public class DatabaseTool
...
+                designTimeProviderAssembly = Assembly.Load(new AssemblyName(providerAssemblyName));
+            }
+            catch (Exception exception)
+            {
+                throw new InvalidOperationException(
+                    Strings.CannotFindDesignTimeProviderAssembly(providerAssemblyName), exception);
+            }
+
+            var type = designTimeProviderAssembly.GetExportedTypes()
+                .FirstOrDefault(t => t.FullName == providerTypeFullName);
```
bricelam:
```C#
var type = designTimeProviderAssembly.GetType(providerTypeFullName);
```

src/EntityFramework.Commands/DatabaseTool.cs
```diff
@@ -75,5 +95,44 @@ public class DatabaseTool
...
+            if (type == null)
+            {
+                throw new InvalidOperationException(
+                    Strings.DesignTimeAssemblyProviderDoesNotContainSpecifiedType(
+                        designTimeProviderAssembly.FullName,
+                        providerTypeFullName));
+            }
+
+            if (!typeof(IDatabaseMetadataModelProvider).IsAssignableFrom(type))
```
bricelam: We need one more level of indirection between the `IDatabaseMetadataModelProvider` and the `designTimeType`.

For example:
```C#
var designTimeServicesFactory = (IServiceProviderFactory)Activator.CreateInstance(designTimeType);
IServiceProvider designTimeServices = designTimeServicesFactory.Create();
var databaseMetadataModelProvider = designTimeServices.GetService<IDatabaseMetadataModelProvider>();
```

ErikEJ:
@lajones unfortunatly, I will also pull in some other dependencies (Microsoft.Aspnet.Razor), that I do not need in the provider, so maybe it is better to keep it seperate, and in line with what you do for SQL Server provider

src/EntityFramework.Commands/Program.cs
```diff
@@ -223,9 +223,14 @@ public virtual int Main([NotNull] string[] args)
                     var connectionString = revEng.Argument(
                             "[connectionString]",
                             "The connection string of the database");
+                    var providerAssemblyName = revEng.Argument(
+                            "[provider]",
+                            "The assembly name of the runtime provider");
```
bricelam: I think *runtime provider* can be just *provider*. Provider consumers shouldn't need to know there's a difference.

src/EntityFramework.Commands/Properties/Strings.resx
```diff
@@ -195,4 +195,16 @@
   <data name="ReusingSnapshotName" xml:space="preserve">
     <value>Reusing model snapshot name '{name}'.</value>
   </data>
+  <data name="CannotFindDesignTimeProviderAssembly" xml:space="preserve">
+    <value>Unable to find design-time provider assembly with name {designTimeProviderAssemblyName}. Ensure that the assembly is referenced by the project.</value>
```
bricelam: Do we want to be more presumptuous and tell users to install the {designTimeProviderAssemblyName} NuGet package? (cc @rowanmiller)

src/EntityFramework.Commands/DatabaseTool.cs
lajones: @bricelam Hmmm. Am not seeing this problem at the moment with the tests I've done. Let's investigate.

src/EntityFramework.Commands/Properties/Strings.resx
rowanmiller: Yes :smile:

src/EntityFramework.Commands/DatabaseTool.cs
rowanmiller: Agreed, that will also enable us to use DI to inject services into the design time components (https://github.com/aspnet/EntityFramework/issues/1769)

bricelam:
For all we worried about and delayed doing this, it turned out great! :shipit:

src/EntityFramework.Commands/DatabaseTool.cs
lajones: Sure - that sounds OK.

src/EntityFramework.Commands/Program.cs
lajones: Hmmm. I had the other feeling. Since there are likely to be 2 packages (the runtime one and the design-time one - see e.g. ErickEJ's comments above) it was good to distinguish which one you meant - and users will in general need to know the difference?

src/EntityFramework.Commands/Properties/Strings.resx
lajones: Yep - totally agree - much better. Will do.

src/EntityFramework.Commands/DatabaseTool.cs
lajones: OK. Will do.

src/EntityFramework.Commands/Program.cs
rowanmiller: I agree with Brice, we are just going to treat the package that has the main (runtime) components as "the provider". If the provider requires a second package to light up design time scenarios then we can provide a nice error message around that (as Brice suggested in another comment) but conceptually the runtime package is "the provider".

lajones: OK. Will do.

ErikEJ:
Looking forward to get started, looks like there will be a lot of non-DRY, as sqlce is so similar to its big Brother 😊

lajones:
Fixed above issue with commit 28c29a1. @ErikEJ - stand by and I'll check in moving of SqlServerNavigationProperty, SqlServerNameMapper etc into Relational.Design as well - so you don't have to replicate those.

ErikEJ:
@lajones Thanks for your effort. I have begun the Model work, and will await your recent PR for the rest-

PR #2552, opened 2015-07-07
Improve core metadata inheritance support

AndriySvyryd:
Add inheritance support to EntityType.Navigations and EntityType.GetForeignKeys()
Use the root type for key values
Expose EntityType on EntityKey
Use inherited FKs and referencing FKs when finding related entities
Make the principal entity type required on FK constructor
Rename Navigation.EntityType, Property.EntityType, Key.EntityType, Index.EntityType and ForeignKey.EntityType to DeclaringEntityType
Move IEntityType.GetProperty(string) and IEntityType.GetNavigation(string) to extension methods
Add FindForeignKey, FindIndex and FindKey to IEntityType

### DIFF
--- a/src/EntityFramework.Core/Metadata/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/EntityType.cs
@@ -47,8 +47,7 @@ namespace Microsoft.Data.Entity.Metadata
         /// <param name="type">The .NET entity type that this metadata object represents.</param>
         /// <param name="model">The model associated with this entity type.</param>
         public EntityType([NotNull] Type type, [NotNull] Model model)
-            : this(
-                (object)Check.NotNull(type, nameof(type)),
+            : this((object)Check.NotNull(type, nameof(type)),
                 Check.NotNull(model, nameof(model)))
         {
             Check.ValidEntityType(type, nameof(type));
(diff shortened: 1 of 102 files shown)


AndriySvyryd:
I will add more tests for the changed code when I'll add Code First inheritance support.

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -993,6 +961,16 @@ public int Compare(IReadOnlyList<Property> x, IReadOnlyList<Property> y)
...
             }
+
+            public bool Equals(IReadOnlyList<Property> x, IReadOnlyList<Property> y)
+            {
+                return Compare(x, y) == 0;
+            }
+
+            public int GetHashCode(IReadOnlyList<Property> obj)
+            {
+                return obj.Aggregate(0, (hash, p) => hash ^ p.GetHashCode());
```
ajcvickers: Expression body

src/EntityFramework.Core/Metadata/EntityTypeExtensions.cs
```diff
@@ -20,6 +20,11 @@ public static IEnumerable<IEntityType> GetDerivedTypes([NotNull] this IEntityTyp
             return GetDerivedTypes(entityType.Model, entityType);
         }
 
+        public static IEnumerable<EntityType> GetDerivedTypes([NotNull] this EntityType entityType)
+        {
+            return ((IEntityType)entityType).GetDerivedTypes().Cast<EntityType>();
```
ajcvickers: Check null

ajcvickers: Never mind. I see it is checked in the overload. In that case, expression body. :-) And other places here as well.

ajcvickers:
:shipit:

PR #2556, opened 2015-07-07
Fix #2192. Throw only one level of AggregateException.

### LINKED ISSUE #2192: Many levels of AggregateException (requires update of ix-async)
[As per SO question here](http://stackoverflow.com/q/30233542/592111)

 I made a simple mistake with some linq (used Skip and forgot to include my OrderBy clause).

The exception that was thrown by this included a number of nested aggregate exceptions.

The code that generates (and catches) the exception is:

    int[] newIds;
    try
    {
        newIds = await db.Products
            .Where(p => p.PortalId == portalId)
            .Skip(ids.ProductIds.Count) //Skip the rows already read
            .Take(takeTotal) //get the next block
            .Select(p => p.ProductId)
            .ToArrayAsync();
    }
    catch (AggregateException ex)
    {
        Console.WriteLine(ex.Message);
        newIds = new int[] { };
    }

However the aggregate exception that I got from this was (this is dumped to the immediate window from the catch block shown above):

>System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.InvalidOperationException: A query containing the Skip operator must include at least one OrderBy operation. at Microsoft.Data.Entity.Relational.Query.Sql.DefaultSqlQueryGenerator.GenerateLimitOffset(SelectExpression selectExpression) at Microsoft.Data.Entity.Relational.Query.Sql.DefaultSqlQueryGenerator.VisitSelectExpression(SelectExpression selectExpression) at Microsoft.Data.Entity.Relational.Query.Expressions.SelectExpression.Accept(ExpressionTreeVisitor visitor) at Microsoft.Data.Entity.Relational.Query.Sql.DefaultSqlQueryGenerator.GenerateSql(SelectExpression selectExpression, IDictionary`2 parameterValues) etc etc

The code was originally running in a WebApi controller with a bunch of levels of async-await, so I thought that the AggregateException layers might be a symptom of that, however I've also tested with a console app just running the EF code, and I still get the nested exceptions. I assume that the levels of exception are due to layers of async-await within the EF code itself. 

As the the comments on the SO question, it can be worked around by using `GetBaseException()`, however I mainly wanted to query wether it was intentional that the exception exiting from the EF code would look like this or whether you'd look to wrap it/flatten it for more pleasant public consumption.

laskoviymishka:
I added specific catch for AggregateException.  [As per SO answer here](http://stackoverflow.com/questions/30233542/many-nested-aggregateexceptions) i just applied GetBaseException() method to catched exception for logger (to log actual error) and rethrow this actual error to clean all redundant layers of aggregation.

### DIFF
--- a/src/EntityFramework.Core/Query/AsyncLinqOperatorProvider.cs
+++ b/src/EntityFramework.Core/Query/AsyncLinqOperatorProvider.cs
@@ -149,6 +149,16 @@ namespace Microsoft.Data.Entity.Query
 
                         return await _inner.MoveNext(cancellationToken);
                     }
+                    catch (AggregateException aggregateException)
+                    {
+                        _exceptionInterceptor._queryContext.Logger.LogError(
+                           new DatabaseErrorLogState(_exceptionInterceptor._queryContext.ContextType),
+                           aggregateException.GetBaseException(),
+                           (state, exception) =>
+                               Strings.LogExceptionDuringQueryIteration(Environment.NewLine, exception));
+
+                        throw aggregateException.GetBaseException();
+                    }
                     catch (Exception e)
                     {
                         _exceptionInterceptor._queryContext.Logger.LogError(
--- a/src/EntityFramework.Relational/Migrations/IMigrator.cs
+++ b/src/EntityFramework.Relational/Migrations/IMigrator.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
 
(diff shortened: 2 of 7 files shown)


rowanmiller:
Hey,

Thanks for the pull request. We would like this to be fixed by adjusting the use of async so that we don't get AggregrateException's in the first place, rather than catching and re-throwing in all the required places (you've only covered one of the many places in this change). The other issue with this approach is that you loose the stack trace etc. from the original exception.

~Rowan

PR #2562, opened 2015-07-07
[SMALL] Moving non-SqlServer-specific code into Relational.Design for re-use

lajones:
See #830. This is mostly just moving classes from SqlServer.Design to Relational.Design. Other providers can use these classes if we move them there. Also renamed NavigationConfiguration to RelationshipConfiguration which better describes it.

@ErikEJ - this will help with your provider.


bricelam:
:shipit:

lajones:
Fixed with commit 3487c25

PR #2568, opened 2015-07-08
Make default sequence increment 1 when sequence created explicitly

ajcvickers:
A sequence created by UseSqlServerSequenceHiLo will still have an increment (block size) of 10. If a sequence is created explicitly and then used for HiLo, then whatever the increment value set in the explicitly created sequence will be used.


bricelam:
:shipit:

ajcvickers:
Merged

PR #2569, opened 2015-07-08
[Migrations] Backing out rebuilds for SQLite

natemcmaster:
Lots of red.

Removing drop-column support for SQLite until we can support rebuilds properly.

Current progress is on https://github.com/natemcmaster/EntityFramework/tree/table-rebuild.

See #329 and #2550


bricelam:
Un-:shipit:

PR #2573, opened 2015-07-08
Fix #1459. Better error message for migrations commands.

### LINKED ISSUE #1459: Migrations: Improve error message when commands not installed in project that contains the context
When `entityframework.commands` is installed in one project in a solution but not the project that contains the context, the migrations powershell commands throw a error message that could be more helpful.  

It would be more useful if EF either detected this situation or caught this exception.

```
PM> add-migration fist -verbose
Using project 'ClassLibrary1'
Build started...
Build succeeded.
Using directory 'C:\projects\ConsoleApplication11\ClassLibrary1\bin\Debug\'.
Using assembly 'ClassLibrary1.dll'.
add-migration : Exception calling "CreateInstanceAndUnwrap" with "8" argument(s): "Could not load file or assembly 'EntityFramework.Commands' or one of its dependencies. The system cannot find the file 
specified."
At line:1 char:1
+ add-migration fist -verbose
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Add-Migration], MethodInvocationException
    + FullyQualifiedErrorId : FileNotFoundException,Add-Migration
 ```

### PR DESCRIPTION
**Pull request by natemcmaster**

### CODE REVIEW

**Comment by lajones** [comment]
Looks good to me provided that is the only possible cause of that error. :shipit:

**Comment by bricelam** [comment]
No, I suspect there are many reasons it could fail. Instead, check to see if the package is install into the project before getting here and throw.

**Comment by natemcmaster** [comment]
If it is other reasons, then a different type of exception will thrown (unless I've misunderstood https://msdn.microsoft.com/en-us/library/3c4f1xde(v=vs.110).aspx). This only catches FileNotFoundException.

**Comment by bricelam** [comment]
FileNotFoundException could also be thrown if a different assembly that the user's code uses is not available.

Here is the eager-checking code.

```ps1
$componentModel = Get-VSComponentModel
$packageInstaller = $componentModel.GetService([NuGet.VisualStudio.IVsPackageInstallerServices])
$package = $packageInstaller.GetInstalledPackages() | ? Id -eq EntityFramework.Commands
if (!($package)) {
    throw 'Yo, teh EntityFramework.Commands ain''t there.'
}
```

natemcmaster:
Yes, the (edited) snippet works. Updated the PR.

bricelam:
:shipit:

PR #2575, opened 2015-07-08
RevEng: Add testing that the generated files compile.

lajones:
Also update the code generation to produce ToTable() instead of old Table() API.
And updated the exceptions to no longer carry around a GeneratedCode property as we were not using it.

See #1669.


bricelam:
:shipit:

lajones:
Fixed with commit ccc0372.

PR #2576, opened 2015-07-08
Fix sentinel bug for nullable types

ajcvickers:
Nullable types were still using the CLR default as a sentinel value in simple key factories. This was because the key factory is, by design, for the unwrapped type (e.g. for nullable FK to non-nullable PK associations.) This change makes a special key factory for the case where the sentinel is null to avoid needing to ever cast the sentinel value to the key type.


anpete:
:shipit:

Issue #2554, opened 2015-07-07
Relational: Incorrect column binding with the same column name


ravengerUA:
We have two entities with the same column name (eg. Name)

Entities
````sql
[Table("award")]
public class Award
{
    [Column("award_id")]
    public int Id { get; set; }

    [Column("name")]
    public string Name { get; set; }

    [Column("country_id")]
    public int CountryId { get; set; }
    public Country Country { get; set; }
}

[Table("countries")]
public class Country
{
    [Column("country_id")]
    public int Id { get; set; }

    [Column("name")]
    public string Name { get; set; }
}
````
and we have this code
````
from a in db.Awards
where a.Id == 1
join c in db.Countries on a.CountryId equals c.Id
select new
{
    AwardId = a.Id,
    Name = a.Name,
    CountryId = c.Id,
    CountryName = c.Name
};
````
We get an error
````
The multi-part identifier "c.country_id" could not be bound.
The multi-part identifier "c.name" could not be bound.
````
But if we write "where a.Id == 1" after inner join, then the code starts work

EF 6.x doesn't have this bug.

ErikEJ:
beta4, beta5 or latest nightly?

ravengerUA:
The latest nightly (7.0.0-beta6-13679)

rowanmiller:
@ravengerUA could you share the SQL that is being sent to the database?

ravengerUA:
This SQL
````sql
SELECT TOP(1) [a].[award_id], [a].[name], [c].[country_id], [c].[name]
FROM [awards] AS [a]
WHERE [a].[award_id] = @__awardId_0
````

maumar:
Fixed in 45eed86223434eb180b81ec1074e3abb36e78778

PR #2577, opened 2015-07-08
[SMALL] Fix to #2554 - Relational: Incorrect column binding with the same column name

maumar:
Problem was that we incorrectly assumed that join body clause will be placed directly after the main clause in the QueryModel. In fact, body clauses are added to the collection in the order in which the query was built. This caused us to be unable to find the previous query source (which we want to join with) and therefore we would not be able to produce a join query on the server.

Fix is to go through list of previously processed body clauses until we find a previous query source, rather than just checking one (previous) body clause.

### DIFF
--- a/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
@@ -220,10 +220,7 @@ namespace Microsoft.Data.Entity.Query
 
             if (selectExpression != null)
             {
-                var previousQuerySource
-                    = index == 0
-                        ? queryModel.MainFromClause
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -253,6 +250,23 @@ var previousQuerySource
             }
         }
 
+        private IQuerySource FindPreviousQuerySource(QueryModel queryModel, int index)
+        {
+            for (int i = index; i >= 0; i--)
+            {
+                var candidate = i == 0
```
anpete: Move this check ahead of loop?

anpete:
:shipit:

PR #2578, opened 2015-07-08
Skip sentinel values when generating keys

ajcvickers:
This covers the case where the sentinel has been set to, for example, -1, and then the temp key generator tries to generate -1 as a temp key.


lajones:
You have an extra underscore in method name "Sentinel_value_on_nullable_property__can_be_skipped_by_value_generator". Otherwise :shipit:

PR #2582, opened 2015-07-09
Improve Reverse-Engineer command tab expansion for installed packages

### LINKED ISSUE #2570: NuGet Commands: Tab-Expand -Provider Parameter
We should enable tab-expansion on the `Reverse-Engineer -Provider` parameter. Just the list of installed NuGet packages would be a good start. Something like this:

```ps1
Register-TabExpansion 'Reverse-Engineer' @{
    'Provider' = {
        Get-Package | select Id
    }
}
```

ErikEJ:
Fixes #2570

### DIFF
--- a/src/EntityFramework.Commands/tools/EntityFramework.psm1
+++ b/src/EntityFramework.Commands/tools/EntityFramework.psm1
@@ -283,7 +283,15 @@ function Remove-Migration {
 Register-TabExpansion Reverse-Engineer @{
     Project = { GetProjects }
     StartupProject = { GetProjects }
-    Provider = { Get-Package | select -ExpandProperty Id }
+    Provider = {
+        param ($context)
+
+        $projectName = $context.Project
+        if (!($projectName)) {
+            $projectName = (Get-Project).ProjectName
+        }
+        return Get-Package -ProjectName $projectName | select -ExpandProperty Id
+    }
 }
 
 <#
--- a/src/EntityFramework.Commands/tools/EntityFramework.psm1
+++ b/src/EntityFramework.Commands/tools/EntityFramework.psm1
@@ -283,6 +283,7 @@ function Remove-Migration {
 Register-TabExpansion Reverse-Engineer @{
     Project = { GetProjects }
     StartupProject = { GetProjects }
+    Provider = { Get-Package | select -ExpandProperty Id }
 }
 
 <#


src/EntityFramework.Commands/tools/EntityFramework.psm1
```diff
@@ -283,6 +283,7 @@ function Remove-Migration {
 Register-TabExpansion Reverse-Engineer @{
     Project = { GetProjects }
     StartupProject = { GetProjects }
+    Provider = { Get-Package | select -ExpandProperty Id }
```
bricelam: We should filter it to the packages installed into the current project. Something like...
```ps1
Provider = {
    param ($context)

    $projectName = $context.Project
    if (!($projectName)) {
        $projectName = (Get-Project).ProjectName
    }

    return Get-Package -ProjectName $projectName | select -ExpandProperty Id
}
```

lajones: Don't think we have a $context for Reverse-Engineer? But we already look up $Project (see the Register-TabExpansion already in place). Can we use that instead?

ErikEJ: How about: 

    Get-Package  -Project (Get-Project).ProjectName | select -Expandproperty Id

bricelam: `$context` is just a DTO of the other specified parameters on the command invokation. Think of it as the context of tab-expansion. Nothing to do with DbContext or a -Context parameter.

lajones: Darn! That sucks. Let's rename it e.g. TabExpansionContext or something else then if we can. Assuming you don't enter a value for the project on the command line (which is likely the default way you'd do it), does $context.Project contain the default value of the project or just null?

bricelam: No, but I covered that case in the code snippet above. `$context` is the de-facto name, and we use it other places in the PowerShell module. Feel free to change if you want, but it doesn't feel overloaded to me in the context of tab-expansion logic.

lajones: How do you rename it? I don't see where it is declared. It would certainly help me understanding the Powershell code.

bricelam: :smile: `param($context)` is the declaration--That's how you define method parameters. Think of that block like `(object context) => { }` in C#.

lajones: @bricelam OK. Makes sense now. Thanks.

bricelam:
:shipit:

PR #2585, opened 2015-07-09
:wrench: Use test fixtures for perf tests

rowanmiller:


smitpatel:
:shipit:

test/EntityFramework.Microbenchmarks.EF6/ChangeTracker/DbSetOperationTests.cs
```diff
@@ -151,12 +145,27 @@ public void Update(MetricCollector collector, bool autoDetectChanges)
...
+        private Customer[] GetAllCustomersFromDatabase()
         {
-            using (var context = new OrdersContext(_connectionString))
+            using (var context = new OrdersContext(_databaseFixture.ConnectionString))
             {
                 return context.Customers.ToArray();
             }
         }
+
+        public class DatabaseFixture
```
anpete: Don't duplicate this below, make it public and share it.

test/EntityFramework.Microbenchmarks.EF6/ChangeTracker/DbSetOperationTests.cs
```diff
@@ -30,7 +24,7 @@ public DbSetOperationTests()
         [BenchmarkVariation("AutoDetectChanges Off", false)]
         public void Add(MetricCollector collector, bool autoDetectChanges)
         {
-            using (var context = new OrdersContext(_connectionString))
+            using (var context = new OrdersContext(_databaseFixture.ConnectionString))
```
anpete: Consider defining a base test class that has the fixture field and a "CreateContext" template method (this is what we do in the runtime).

test/EntityFramework.Microbenchmarks.EF6/ChangeTracker/DbSetOperationTests.cs
rowanmiller: Will do. I was originally going to do that but it seemed like overkill to share two partial lines of code (given the initialization is already in a component that just takes the counts)... but I guess it sets it up to better add stuff in the future. I think I'll also add the CreateContext method you suggested to the fixture rather than a base class for the tests.

PR #2587, opened 2015-07-09
Prevent unnecessary parameterization of enum constants

mikary:
Prevent unnecessary parameterization of enum constants
Add test coverage for query with nullable enum parameters

@anpete

### DIFF
--- a/src/EntityFramework7.Core/Query/ExpressionVisitors/ParameterExtractingExpressionVisitor.cs
+++ b/src/EntityFramework7.Core/Query/ExpressionVisitors/ParameterExtractingExpressionVisitor.cs
@@ -66,6 +66,11 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors
 
             if (expression.NodeType == ExpressionType.Convert)
             {
+                if(expression.RemoveConvert() is ConstantExpression )
+                {
+                    return expression;
+                }
+
                 var unaryExpression = (UnaryExpression)expression;
 
                 if ((unaryExpression.Type.IsNullableType()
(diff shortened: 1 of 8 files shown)


test/EntityFramework.Core.FunctionalTests/TestModels/GearsOfWarModel/GearsOfWarModelInitializer.cs
```diff
@@ -65,48 +65,57 @@ public static void Seed(GearsOfWarContext context)
 
                 var marcusLancer = new Weapon
```
maumar: now that I think about it, it would be nice to add 'Marcus' Hammer of Dawn" with null ammunition type - for completeness, and also to make sure that null semantics works well with enums

src/EntityFramework.Core/Query/ExpressionVisitors/ParameterExtractingExpressionVisitor.cs
```diff
@@ -64,6 +64,18 @@ public override Expression Visit(Expression expression)
 
             var e = expression;
 
+            while(e.NodeType == ExpressionType.Convert)
```
maumar: formatting - add space after while/if

anpete: Use ExpressionExtensions.RemoveConvert

test/EntityFramework.Core.FunctionalTests/TestModels/GearsOfWarModel/GearsOfWarModelInitializer.cs
anpete: Seems like the Hammer of Dawn would always have ammo though.

maumar: We were considering Butcher's Cleaver, but most likely it will not pass policheck...

mikary:
Updated with requested changes.

maumar:
:shipit:

smitpatel:
Question:
https://github.com/aspnet/EntityFramework/blob/dev/test/EntityFramework7.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerFixture.cs#L40
Why aren't we deleting database every time like the other fixtures? Whenever the model changes for this, tests fail till database is deleted manually.

maumar:
@smitpatel no reason at all, we should do it.

PR #2602, opened 2015-07-10
Migrations: Implement inheritance

### LINKED ISSUE #1706: Migrations: Implement inheritance support
- Update the differ to understand TPH.
- ~~Consider gen'ing a check constraint for discriminator value set.~~ (Created #2594)

bricelam:
Resolves #1706

### DIFF
--- a/src/EntityFramework.Relational/EntityFramework.Relational.csproj
+++ b/src/EntityFramework.Relational/EntityFramework.Relational.csproj
@@ -67,7 +67,10 @@
     <Compile Include="Metadata\Conventions\Internal\RelationalConventionSetBuilder.cs" />
     <Compile Include="Metadata\Conventions\Internal\RelationalTableAttributeConvention.cs" />
     <Compile Include="Metadata\IRelationalMetadataExtensionProvider.cs" />
+    <Compile Include="Metadata\RelationalEntityTypeExtensions.cs" />
+    <Compile Include="Metadata\RelationalModelExtensions.cs" />
     <Compile Include="Metadata\RelationalPropertyAnnotationsExtensions.cs" />
+    <Compile Include="Metadata\RelationalPropertyExtensions.cs" />
     <Compile Include="Migrations\Infrastructure\ModelDifferContext.cs" />
     <Compile Include="RelationalReferenceReferenceBuilderExtensions.cs" />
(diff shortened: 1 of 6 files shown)


src/EntityFramework.Core/Metadata/EntityTypeExtensions.cs
```diff
@@ -310,5 +310,12 @@ public static IKey GetKey([NotNull] this IEntityType entityType, [NotNull] IRead
 
             return key;
         }
+
+        public static IEnumerable<IKey> GetDeclaredKeys([NotNull] this IEntityType entityType)
+        {
+            Check.NotNull(entityType, nameof(entityType));
+
+            return entityType.GetKeys().Where(k => k.EntityType == entityType);
+        }
```
bricelam: @AndriySvyryd Any reason this one was "missed"? Does anything else need to be done in this area?

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -305,7 +305,6 @@ public virtual Key AddKey([NotNull] Property property)
         public virtual Key AddKey([NotNull] IReadOnlyList<Property> properties)
         {
             Check.NotEmpty(properties, nameof(properties));
-            ThrowIfDerivedEntity();
```
AndriySvyryd: We need to be careful about expanding the scope of inheritance for RTM. If supporting keys on non-root types is something we think is high priority we need to make sure it works across the stack (query, fixup, code first, etc.)

bricelam: @anpete @ajcvickers @divega @rowanmiller Thoughts? Should we scope to not allow alternate keys on derived types? Code First appeared to superficially work after removing this.

rowanmiller: I'm ok with limiting it for now and enabling it when we have time to test it properly

divega: +1

bricelam: Will undo. Filed #2611.

AndriySvyryd:
:shipit:

PR #2604, opened 2015-07-11
Make EntityType.FindKey account for inheritance

AndriySvyryd:
Make InternalRelationshipBuilder.Required behave more consistently as to which properties it affects
Remove unused members from EntityType
Add inheritance helper members for ForeignKey


ajcvickers:
:shipit:

Issue #107, opened 2014-04-22
Convention-based model builder support for Data Annotations


divega:
We should try to support at least the very basic ones, e.g. KeyAttribute for alpha, provided that the DataAnnotations build is ready for K.
Attributes to be implemented:
- [x] Key
- [x] Required
- [x] MaxLength
- [x] NotMapped
- [x] ConcurrencyCheck
- [x] TimeStamp
- [x] Table
- [x] Column
- [x] DatabaseGenerated
- [x] StringLength

Relationship
- [x] ForeignKey
- [x] InverseProperty
- [x] Required on Navigations

Not supported yet
- ComplexType

divega:
@smitpatel please talk to @AndriySvyryd about the details.

paul-daly:
I ran into another scenario where this issue causes problems.  

Because .MaxLengh() isn't being honored on string properties, nvarchar(MAX) fields are generated in the db.  Unique constraints aren't supported on nvarchar(MAX) fields, so an error is thrown during db generation not for applying the .MaxLength() annotation, but when you apply a Unique constraint on a string property where you think you've limited the length.

jasonwilliams200OK:
:+1: 
Please also provide annotation for uniqueness constraint as a model `convention`, so we don't have to "drop down" to [manual model builder](https://github.com/aspnet/EntityFramework/pull/866#issuecomment-98000296) `configuration` probably like so:

```c#
[Unique]
public int MySurrogateKey { get; set; }
// here MySurrogateKey can be null as a unique value
```

```c#
[UniqueNotNull]
public int MySurrogateKey { get; set; }
// here MySurrogateKey can not be null
```

```c#
[UniqueAllowNull]
public int MySurrogateKey { get; set; }
// here MySurrogateKey can be null multiple times
```

copy: @Tratcher, @davidfowl, @NTaylorMullen

smitpatel:
filed #2504 & #2505 
Attributes to be implemented has been added in first post. We need to decide which relationship attributes we want to support.

divega:
We should consider StringLength which is a synonym of MaxLength only applicable to strings.

rowanmiller:
+1 for StringLength.
We also need to discuss `[ForeignKey]` and '[InverseProperty]`. My gut feel is we should probably support them with the same rules we had in EF6.

divega:
+1 for [ForeginKey] and [InverseProperty]

smitpatel:
[...]

PR #2614, opened 2015-07-13
Data Annotations Part 3

smitpatel:
part of #107 
Add support for navigation added conventions
NotMapped attribute for properties/navigations
Required attribute for navigations

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/RequiredNavigationAttributeConvention.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/RequiredNavigationAttributeConvention.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.DataAnnotations;
+using System.Reflection;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
+{
+    public class RequiredNavigationAttributeConvention : NavigationAttributeConvention<RequiredAttribute>
+    {
+        public override InternalRelationshipBuilder Apply(InternalRelationshipBuilder relationshipBuilder, Navigation navigation, RequiredAttribute attribute)
+        {
+            Check.NotNull(relationshipBuilder, nameof(relationshipBuilder));
+            Check.NotNull(navigation, nameof(navigation));
+            Check.NotNull(attribute, nameof(attribute));
+
+            if (!navigation.PointsToPrincipal() || navigation.DeclaringEntityType.ClrType?.GetProperty(navigation.Name).PropertyType.TryGetSequenceType() != null)
+            {
+                return relationshipBuilder;
+            }
+            return relationshipBuilder.Required(true, ConfigurationSource.DataAnnotation) ?? relationshipBuilder;
+        }
+    }
+}
--- a/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -256,7 +256,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                    && (canOverrideSameSource || configurationSource != currentConfigurationSource);
         }
 
-        public virtual bool Navigation(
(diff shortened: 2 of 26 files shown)


src/EntityFramework.Core/Metadata/Conventions/Internal/RequiredNavigationAttributeConvention.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public class RequiredNavigationAttributeConvention : NavigationAttributeConvention<RequiredAttribute>
+    {
+        public override InternalRelationshipBuilder Apply(InternalRelationshipBuilder relationshipBuilder, Navigation navigation, RequiredAttribute attribute)
+        {
+            Check.NotNull(relationshipBuilder, nameof(relationshipBuilder));
+            Check.NotNull(navigation, nameof(navigation));
+            Check.NotNull(attribute, nameof(attribute));
+
+            return relationshipBuilder.Required(true, ConfigurationSource.DataAnnotation) ?? relationshipBuilder;
```
AndriySvyryd: I think we should only do this if the [Required] is on a reference navigation, putting it on a collection navigation doesn't have the same semantics.

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -285,15 +285,17 @@ private bool CanRemove(ForeignKey foreignKey, ConfigurationSource configurationS
                 ? this
                 : ModelBuilder.Entity(foreignKey.DeclaringEntityType.Name, ConfigurationSource.Convention);
 
+            var builder = fkOwner._relationshipBuilders.Value.TryGetValue(foreignKey, ConfigurationSource.Convention) ?? fkOwner.Relationship(foreignKey, true, ConfigurationSource.Convention);
```
AndriySvyryd: you should be able to just call `fkOwner.Relationship(foreignKey, true, ConfigurationSource.Convention)`

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -312,17 +314,23 @@ private bool CanRemove(ForeignKey foreignKey, ConfigurationSource configurationS
...
             {
-                if (_ignoredProperties.HasValue)
+                if (_ignoredProperties.HasValue && _ignoredProperties.Value.ContainsKey(navigationName))
                 {
+                    ConfigurationSource existingConfigurationSource;
+                    _ignoredProperties.Value.TryGetValue(navigationName, out existingConfigurationSource);
+                    if (!configurationSource.Overrides(existingConfigurationSource))
+                    {
+                        return builder;
+                    }
```
AndriySvyryd: This is redundant as it's already checked in `CanSetNavigation`

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -903,19 +911,28 @@ private InternalIndexBuilder Index(IReadOnlyList<Property> properties, Configura
                     Debug.Assert(requiredSet || !isRequired.Value);
                 }
 
+                foreignKey.IsRequired = isRequired.Value;
```
AndriySvyryd: This shouldn't be necessary as IsRequired is calculated from the contained properties. Instead this should be changed to an Assert and if it fails then the logic needs to be fixed.

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -386,7 +388,7 @@ public virtual bool Ignore([NotNull] string propertyName, ConfigurationSource co
             var navigation = Metadata.FindNavigation(propertyName);
             if (navigation != null)
             {
-                if (!Navigation(null, navigation.ForeignKey, navigation.PointsToPrincipal(), configurationSource, canOverrideSameSource: false))
+                if (Navigation(null, navigation.ForeignKey, navigation.PointsToPrincipal(), configurationSource, canOverrideSameSource: configurationSource != ConfigurationSource.Explicit) == null)
```
AndriySvyryd: I changed my mind on this, I don't think this is the right way to fix it. Ignore behavior should be the same for any configuration source. The problem here is that we don't record the ConfigurationSource for navigations, that's what should be fixed.

AndriySvyryd:
:shipit:, Adding a separate configuration source for navigations can be done in a different commit

smitpatel:
merged with 03fe8f97ee1ed3618ad5c0257c89d8e3aa307488

PR #2620, opened 2015-07-14
[TINY] Support additional metadata on perf runs

rowanmiller:
Allows extra info to be set (config file or environment variable) that
will then be written to the database. This enables scenarios such as
recording the commit hash against results during perf runs.

### DIFF
--- a/test/EntityFramework7.Microbenchmarks.Core/BenchmarkConfig.cs
+++ b/test/EntityFramework7.Microbenchmarks.Core/BenchmarkConfig.cs
@@ -20,7 +20,8 @@ namespace EntityFramework.Microbenchmarks.Core
                 RunIterations = bool.Parse(config.Get("benchmarks:runIterations")),
                 ResultsDatabase = config.Get("benchmarks:resultsDatabase"),
                 BenchmarkDatabaseInstance = config.Get("benchmarks:benchmarkDatabaseInstance"),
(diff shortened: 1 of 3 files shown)


test/EntityFramework7.Microbenchmarks.Core/BenchmarkConfig.cs
```diff
@@ -20,7 +20,8 @@ public class BenchmarkConfig
                 RunIterations = bool.Parse(config.Get("benchmarks:runIterations")),
                 ResultsDatabase = config.Get("benchmarks:resultsDatabase"),
                 BenchmarkDatabaseInstance = config.Get("benchmarks:benchmarkDatabaseInstance"),
-                ProductReportingVersion = config.Get("benchmarks:productReportingVersion")
+                ProductReportingVersion = config.Get("benchmarks:productReportingVersion"),
+                ExtraInfo = config.Get("benchmarks:extraInfo")
```
anpete: Consider a better name for this. Perhaps something like "CustomData"?

anpete:
:shipit:

test/EntityFramework7.Microbenchmarks.Core/BenchmarkConfig.cs
rowanmiller: Sure, I'll change it :smile:

PR #2621, opened 2015-07-14
Changes to default and computed value SQL

ajcvickers:
Specifically:
- Make computed expression relational, not SQL Server, specific
- When using fluent API to set default or computed SQL, set value generation appropriately
- Use one metadata slot for column SQL with value generation deciding how it is used
- Update Migrations to work with these changes
- Generate temp values for GUIDs unless store generation has been asked for
- Renaming based on API review decisions (although we may want to revisit, since its not very consistent)


test/EntityFramework7.Relational.Tests/Migrations/Sql/MigrationSqlGeneratorTestBase.cs
```diff
@@ -47,6 +47,20 @@ public virtual void AddColumnOperation_with_defaultValueSql()
         }
 
         [Fact]
+        public virtual void AddColumnOperation_with_computed_column_SQL()
```
bricelam: Override and assert the SQL in the SQL Server tests.

bricelam:
:shipit:

test/EntityFramework7.Relational.Tests/Migrations/Sql/MigrationSqlGeneratorTestBase.cs
ajcvickers: Done.

Issue #1769, opened 2015-03-06
Commands: Review Dependency Injection patterns


lajones:
We use different patterns of DI in Migrations Commands and RevEng. Review to see if we can re-use the patterns from the rest of the stack.

lajones:
See #1751

lajones:
Changes for better dependency injection for RevEng checked in with PR https://github.com/aspnet/EntityFramework/pull/2623

PR #2623, opened 2015-07-14
Changing ReverseEngineering to use a better form of dependency injection

lajones:
See #1769.

If this looks like a good approach I'll update migrations to take the same approach.

### DIFF
--- a/src/EntityFramework.Commands/DatabaseTool.cs
+++ b/src/EntityFramework.Commands/DatabaseTool.cs
@@ -8,22 +8,21 @@ using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Commands.Utilities;
-using Microsoft.Data.Entity.Internal;
-using Microsoft.Data.Entity.Relational.Design.CodeGeneration;
 using Microsoft.Data.Entity.Relational.Design.ReverseEngineering;
-using Microsoft.Data.Entity.Relational.Design.Templating;
-using Microsoft.Data.Entity.Relational.Design.Templating.Compilation;
-using Microsoft.Data.Entity.Relational.Design.Utilities;
 using Microsoft.Data.Entity.Storage;
 using Microsoft.Data.Entity.Utilities;
+using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Logging;
+#if DNX451 || DNXCORE50
+using Microsoft.Framework.Runtime;
+#endif
 
 namespace Microsoft.Data.Entity.Commands
 {
     public class DatabaseTool
     {
         private readonly ServiceProvider _serviceProvider;
-        private readonly LazyRef<ILogger> _logger;
+        private readonly ILoggerProvider _loggerProvider;
 
         public DatabaseTool(
             [CanBeNull] IServiceProvider serviceProvider,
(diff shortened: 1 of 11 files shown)


src/EntityFramework.Commands/DatabaseTool.cs
```diff
@@ -81,7 +70,12 @@ public class DatabaseTool
             var designTimeAssemblyName =
                 designTimeServicesTypeAttribute.AssemblyName ?? runtimeProviderAssemblyName;
 
-            var designTimeProvider = GetDesignTimeProvider(designTimeTypeName, designTimeAssemblyName);
+            var designTimeMetadataProviderFactory =
+                GetDesignTimeMetadataProviderFactory(designTimeTypeName, designTimeAssemblyName);
+            var serviceCollection = SetupInitialServices();
+            designTimeMetadataProviderFactory.AddMetadataProviderServices(serviceCollection);
+            designTimeMetadataProviderFactory.Create(serviceCollection);
```
bricelam: Remove this line?

src/EntityFramework.Relational.Design/ReverseEngineering/DesignTimeMetadataProviderFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            serviceCollection.AddTransient<CSharpCodeGeneratorHelper, CSharpCodeGeneratorHelper>();
+            serviceCollection.AddTransient<ModelUtilities, ModelUtilities>();
+            serviceCollection.AddTransient<ICompilationService, RoslynCompilationService>();
+            serviceCollection.AddTransient<MetadataReferencesProvider, MetadataReferencesProvider>();
+            serviceCollection.AddTransient<ITemplating, RazorTemplating>();
+            serviceCollection.AddTransient<ReverseEngineeringGenerator, ReverseEngineeringGenerator>();
+        }
+
+        public abstract IDatabaseMetadataModelProvider Create([NotNull] ServiceCollection serviceCollection);
```
bricelam: Why is this needed? Can't commands just get an `IDatabaseMetadataModelProvider` from the service provider?

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerDesignTimeMetadataProviderFactory.cs
```diff
@@ -1,20 +1,25 @@
...
+        public override IDatabaseMetadataModelProvider Create([NotNull] ServiceCollection serviceCollection)
         {
-            Check.NotNull(serviceProvider, nameof(serviceProvider));
+            Check.NotNull(serviceCollection, nameof(serviceCollection));
 
-            return new SqlServerMetadataModelProvider(serviceProvider);
+            var serviceProvider = serviceCollection.BuildServiceProvider();
+            var logger = serviceProvider.GetRequiredService<ILogger>();
+            var modelUtilities = serviceProvider.GetRequiredService<ModelUtilities>();
+            return new SqlServerMetadataModelProvider(logger, modelUtilities);
```
bricelam: Yeah, this is silly. The thing that calls this could do the same if `SqlServerMetadataModelProvider` was registered as an `IDatabaseMetadataModelProvider`.

bricelam:
:shipit: So much better! Assign the bug to me now. I'll port the pattern to Migrations and look into a few more ideas I've had in this area.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerDesignTimeMetadataProviderFactory.cs
lajones: Yeah - that's fine. I want to keep the majority of the service-registering in `DesignTimeMetadataProviderFactory` - so other providers can use it. But I'll update this so instead of a separate Create() method it just overrides `DesignTimeMetadataProviderFactory.AddMetadataProviderServices()`, calls the base class and then adds the service  'SqlServerMetadataModelProvider`  as an  `IDatabaseMetadataModelProvider` as you say.

lajones:
Changes checked into release with commit a952a03, dev with commit 65cd660. Assigning to you for the Migrations changes.

PR #2630, opened 2015-07-15
[TINY] Enable result saving for EF6 perf tests

rowanmiller:
The perf infrastructure uses EF7 to persist results. Enabling saving of
results for EF6 tests now that we support EF6/EF7 loaded in the same
AppDomain.
Also fixing a couple of incorrect namespaces in test files.


natemcmaster:
:shipit: - assuming "runIterations" isn't going to run on TC

rowanmiller:
Actually that runIterations value shouldn't change. I'll revert it back to false :smile:

PR #2632, opened 2015-07-15
Fix Commands.FunctionalTests in csproj-based solution

bricelam:
Roslyn is compiled against an older version of Immutable Collections. Adding a binding redirect to compensate.

@smitpatel


ajcvickers:
:shipit: Please :shipit:!

bricelam:
@lajones It looks like `SqlServer.Design.FunctionalTests` needs this too.

lajones:
@bricelam - @ajcvickers already did it for me in commit aa55e98. Thanks.

Issue #1989, opened 2015-04-03
Make rowversion optimistic concurrency resolution easier


ajcvickers:
With the recent changes to read-only semantics and computed properties the resolution of optimistic concurrency issues has becomes harder. This is because previously (and in the old stack) values of computed properties would never be sent to the database and would always be back propagated from the database, so it only mattered what the original value was.

Now if the property is modified and the property is set as read-only after save, then an exception is thrown. This means that the property needs to be explicitly marked as non-modified after applying changes to current or original values. Further, if current or original values are different, then DetectChanges will mark the property as modified again, so this must either be switched off or original and current values must be made the same to avoid this.

The idea we had when discussing this was to add some form of higher-level functionality to make this easier. We could also consider changing the the low-level read-only behavior and/or the way detect changes and/or the update pipeline works for read-only, computed properties.

divega:
We need to make simple concurrency control simple to use again. Either by adding a new feature or by changing what we do with read-only properties again.

PR #2634, opened 2015-07-15
Make optimistic concurrency usable

ajcvickers:
Introduces a new flag on property indicating that store generated values should always be used regardless of the current value in the property or whether or not it is marked as modified. By default any property that is a concurrency token and generated on add and update will also get this flag set, although it can be switched off. This makes rowversion optimistic concurrency just work out of the box.

See #1989


bricelam:
:shipit:

PR #2625, opened 2015-07-15
Add EntityTypeBuilder.BaseType

AndriySvyryd:
Adjust the ModelBuilder logic to account for base types
Discover abstract CLR types

Part of #1704

Issue #2597, opened 2015-07-10
Model Snapshot: Implement Inheritance Support


bricelam:
Depends on #1704

rowanmiller:
@smitpatel is this done?

smitpatel:
This is done. ModelSnapshot includes call to BaseType in entity declaration, which should do everything required. Closing this issue. If we find something missing in future, will file a new issue.

PR #2640, opened 2015-07-16
Include BaseType info in model snapshot

smitpatel:
Part of #2597 
Account for changes of #2625  into model snapshot.


test/EntityFramework7.Commands.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -119,6 +129,38 @@ public void EntityType_annotations_are_stored_in_snapshot()
...
+
+        b.Key(""Id"");
+    });
+
+builder.Entity(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+DerivedEntity"", b =>
+    {
+        b.Property<string>(""Name"");
+    });
+
+builder.Entity(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+DerivedEntity"").BaseType(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+BaseEntity"");
```
bricelam: We may want a line break before calling `.BaseType()`

bricelam:
:shipit: (May also want review from @AndriySvyryd)

src/EntityFramework7.Core/Metadata/EntityTypeExtensions.cs
```diff
@@ -304,6 +304,12 @@ public static IKey GetPrimaryKey([NotNull] this IEntityType entityType)
             return primaryKey;
         }
 
+        public static IKey GetDeclaredPrimaryKey([NotNull] this IEntityType entityType)
```
AndriySvyryd: Should be named FindDeclaredPrimaryKey

src/EntityFramework7.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -97,21 +116,21 @@ protected enum GenerateEntityTypeOptions
...
+                    if ((options & GenerateEntityTypeOptions.Declared) != 0)
                     {
-                        GenerateProperties(entityType.GetProperties(), stringBuilder);
+                        GenerateProperties(entityType.GetDeclaredProperties(), stringBuilder);
 
-                        GenerateKey(entityType.GetPrimaryKey(), stringBuilder);
+                        GenerateKey(entityType.GetDeclaredPrimaryKey(), stringBuilder);
 
-                        GenerateIndexes(entityType.GetIndexes(), stringBuilder);
+                        GenerateIndexes(entityType.GetDeclaredIndexes(), stringBuilder);
```
AndriySvyryd: In the near-future Indexes and Keys will be able to use inherited properties, so they should be configured after BaseType is set

smitpatel: If that is the case then how about,
Sort entities so that base types are declared before derived types.
In derived type declaration, make call to base type before any other declaration.
thoughts? @bricelam

AndriySvyryd: Or you could group the code by member type, rather than by entity:
* BaseType
* Properties
* Indexes
* Keys
* Relationships

bricelam: I don't like the idea of breaking up entities into different "parts". Foreign keys make sense since they could be cyclic, but everything else should be in a self-contained call to `.EntityType()` in my opinion. Ideally, I'd like the FKs inline too--just pulling out one in the cycle if needed.
 cc @rowanmiller

rowanmiller: Agreed with @bricelam, although I do think it is fine to always pull FKs out separately since relationships are somewhat conceptually self contained (I don't think it's worth the complexity of working out which ones to split out).

bricelam: Yeah, the simplest logic would be to topologically sort entity types by hierarchy then configure foreign keys afterward.

divega: As a user, I kind of like it more if relationships are separate. I guess this is very subjective.

Issue #2181, opened 2015-05-13
Include tries to use PK rather than alternate key


rowanmiller:
In a model that uses an alternate key Include still tries to join to the PK. This either results in incorrect SQL or an exception if the types don't match.

```csharp
using System.Collections.Generic;
using System.Linq;
using Microsoft.Data.Entity;

namespace ConsoleApp1
{
    public class Program
    {
        public void Main(string[] args)
        {
            using (var db = new MyContext())
            {
                db.Database.EnsureCreated();
                db.TaxReturns.Include(r => r.Person).ToList();
            }
        }
    }

    public class MyContext : DbContext
    {
        public DbSet<Person> People { get; set; }
        public DbSet<TaxReturn> TaxReturns { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder builder)
        {
            builder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=Test;Trusted_Connection=True;MultipleActiveResultSets=True;");
        }

        protected override void OnModelCreating(ModelBuilder builder)
        {
            builder.Entity<Person>()
                .Collection(p => p.TaxReturns)
                .InverseReference(r => r.Person)
                .ForeignKey(r => r.SSN)
                .PrincipalKey(p => p.SSN);
        }
    }

    public class Person
    {
        public int PersonId { get; set; }
        public string Name { get; set; }
        public string SSN { get; set; }

        public List<TaxReturn> TaxReturns { get; set; }
    }

    public class TaxReturn
    {
        public int TaxReturnId { get; set; }

        public string SSN { get; set; }
        public Person Person { get; set; }
    }
}
```

> System.InvalidOperationException was unhandled by user code
  HResult=-2146233079
  Message=No coercion operator is defined between types 'System.Int32' and 'System.String'.
  Source=System.Core
  StackTrace:
       at System.Linq.Expressions.Expression.GetUserDefinedCoercionOrThrow(ExpressionType coercionType, Expression expression, Type convertToType)
       at System.Linq.Expressions.Expression.Convert(Expression expression, Type type, MethodInfo method)
       at Microsoft.Data.Entity.Relational.Query.ExpressionTreeVisitors.IncludeExpressionTreeVisitor.BuildJoinEqualityExpression(INavigation navigation, IReadOnlyList`1 primaryKeyProperties, TableExpressionBase targetTableExpression, TableExpressionBase joinExpression)
[...]

PR #2641, opened 2015-07-16
[BETA6/SMALL] Fix to #2181 - Include tries to use PK rather than alternate key

maumar:
Issue was that when building include expressions, we always assumed that joins should happen based on primary keys. However users can specify alternate key, and for those cases it should be used instead. Fix is to use correct navigation.ForeignKey.PrincipalKey to determine properties that should be used to create a join.


maumar:
@anpete @ajcvickers

src/EntityFramework7.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
```diff
@@ -298,7 +290,7 @@ var newExpression
 
             for (var i = 0; i < navigation.ForeignKey.Properties.Count; i++)
             {
-                var primaryKeyProperty = primaryKeyProperties[i];
+                var primaryKeyProperty = navigation.ForeignKey.PrincipalKey.Properties[i];
```
smitpatel: Should also change variable name to principalKeyProperty considering it may not be primaryKeyProperty

test/EntityFramework7.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerFixture.cs
```diff
@@ -38,7 +38,7 @@ public override SqlServerTestStore CreateTestStore()
                     using (var context = new GearsOfWarContext(_serviceProvider, optionsBuilder.Options))
                     {
                         // TODO: Delete DB if model changed
-
+                        context.Database.EnsureDeleted();
```
smitpatel: :+1:

smitpatel:
LGTM though someone else should give sign-off

src/EntityFramework7.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
maumar: will do!

anpete:
:shipit:

Issue #2619, opened 2015-07-14
Commands: "Lazily" use EntityFramework.Relational.Design to enable UWP


bricelam:
`EntityFramework.Commands` references `EntityFramework.Relational.Design` which references `Microsoft.AspNet.Razor`.

`Microsoft.AspNet.Razor` is not available on `netcore50`. 87a892a1f4ae3a2a3f2ca9484d50d77e4bba8322 mitigated this by not installing `EntityFramework.Relational.Design` into `netcore50` projects.

In theory, this should work because it is only ever used by Reverse Engineer which doesn't need to work on `netcore50` right now. It looks like, however types from `EntityFramework.Relational.Design` are being initialize when they're not needed. I suspect this because when I do `Add-Migration Test`, I get the following error.

> Could not load file or assembly 'EntityFramework.Relational.Design, Version=7.0.0.0, 
Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.

We should make the use of `EntityFramework.Relational.Design` in `EntityFramework.Commands` more lazy to unblock Migrations commands on `netcore50` projects.

bricelam:
Note that I mean JIT-time lazy, not runtime lazy.

divega:
Isn't it the case in that we would need the parts of Razor we are using to be available everywhere? Sounds like a reason to reconsider parts of our plans.

bricelam:
Technically, it never runs on `netcore50`. At design-time, it swaps in the `net45` version of the Commands assembly and executes on Full .NET.

bricelam:
We could re-package Razor (and any of its dependencies) into the Relational.Design package for design-time.

rowanmiller:
@lajones can you work on this as a priority as it is blocking commands on UWP.

lajones:
Fixed by https://github.com/aspnet/EntityFramework/pull/2642

PR #2642, opened 2015-07-16
Load DatabaseTool (and MigrationTool) lazily.

lajones:
@bricelam Addresses #2619


bricelam:
:shipit:

lajones:
Thanks. Checked into release with commit 356230b, dev with commit #e72970a.

PR #2643, opened 2015-07-16
Write perf results to diagnostics

rowanmiller:
This allows us to switch on diagnostics and see the perf results in VS
output window or on the command line. This makes it easy to use perf
tests as a harness when working on perf improvements.
Also adding config files so that the diagnostic messages can be easily
switched on.

### DIFF
--- a/test/EntityFramework7.Microbenchmarks.Core/BenchmarkRunSummary.cs
+++ b/test/EntityFramework7.Microbenchmarks.Core/BenchmarkRunSummary.cs
@@ -68,6 +68,15 @@ namespace EntityFramework.Microbenchmarks.Core
             MemoryDeltaStandardDeviation = StandardDeviation(memoryDeltas, MemoryDeltaAverage);
         }
 
+        public override string ToString()
+        {
+            return $@"{TestClass}.{TestMethod} (Variation={Variation})
+    Warmup Iterations: {WarmupIterations}
+    Collection Iterations: {Iterations}
+    Time Elapsed (95th Percentile): {TimeElapsedPercentile95}ms
+    Memory Delta (95th Percentile): {MemoryDeltaPercentile95}";
+        }
+
         private static long Percentile(IEnumerable<long> results, double percentile)
         {
             return results.OrderBy(r => r).ElementAt((int)(results.Count() * percentile));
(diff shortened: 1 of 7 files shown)


test/EntityFramework7.Microbenchmarks.Core/BenchmarkRunSummary.cs
```diff
@@ -68,6 +68,15 @@ public void PopulateMetrics()
             MemoryDeltaStandardDeviation = StandardDeviation(memoryDeltas, MemoryDeltaAverage);
         }
 
+        public string GetResultString()
```
anpete: ToString?

test/EntityFramework7.Microbenchmarks.Core/BenchmarkTestCaseRunner.cs
```diff
@@ -39,6 +40,8 @@ public class BenchmarkTestCaseRunner : XunitTestCaseRunner
         private static Lazy<string> _framwork = new Lazy<string>(() => ".NETFramework");
```
anpete: Typo

anpete: Not sure why this needs to be lazy.

anpete:
:shipit:

test/EntityFramework7.Microbenchmarks.Core/BenchmarkRunSummary.cs
rowanmiller: Yeah I went back and forwards on that. I couldn't decide... I'll swap it :smile:

test/EntityFramework7.Microbenchmarks.Core/BenchmarkTestCaseRunner.cs
rowanmiller: Here is the full code:

```
#if DNX451 || DNXCORE50
        private static Lazy<string> _framwork = new Lazy<string>(() => 
        {
            var services = Microsoft.Framework.Runtime.Infrastructure.CallContextServiceLocator.Locator.ServiceProvider; 
            var env = (Microsoft.Framework.Runtime.IRuntimeEnvironment)services.GetService(typeof(Microsoft.Framework.Runtime.IRuntimeEnvironment)); 
            return "DNX." + env.RuntimeType;
        });
#else
        private static Lazy<string> _framwork = new Lazy<string>(() => ".NETFramework");
#endif
```

anpete: Are there code paths where .Value is not accessed?

Issue #2494, opened 2015-06-29
Decide on hiding of DbConnection and DbTransaction


ajcvickers:
Consider that:
- In general we want people to manipulate connections and transactions through our methods
- Those methods are now flattened onto the context.Database API--our connection and transaction objects do not now need to be used
  - Note that we do still return our transaction object from methods like UseTransaction
- There needs to be some way for apps to get hold of the underlying DbConnection/DbTransaction objects for playing nice with other code that uses these objects directly

So we currently have:
```
   context.Database.GetDbConnection();
   context.Database.GetDbTransaction();
```

This allows apps to get the underlying objects, but it also makes it very easy to get the underlying objects such that it will encourage not using our APIs for manipulation of those objects. Should they be more hidden and if so how? Perhaps IAccessor pattern? Perhaps through getting the EF connection/transaction objects, but then why the flattening?

rowanmiller:
We'll leave connection and remove transaction (you can get it off RelationalTransaction)

PR #2644, opened 2015-07-16
Remove GetDbTransaction method

ajcvickers:
Can get it off the returned RelationalTransaction when needed. See Issue #2494


bricelam:
:shipit:

Issue #1704, opened 2015-02-25
Code First: Implement inheritance support


anpete:
* [x] APIs for configuring base type
* [ ] Base type discovery convention
* [x] Support for referencing inherited entity type members
* [x] Support for ignoring discovered base type
* [x] Support for ignoring inherited properties
* [x] APIs for relational discriminator annotations
* [x] Discriminator introduction convention (readonly by default)
* [x] Relational model validation rule to ensure correct inheritance mapping

bricelam:
@AndriySvyryd Two questions:

1. Will Code First take care uniquifying/collapsing properties in the hierarchy? For example, what would happen to the `Name` properties:

        class Person
        {
        }

        class Student : Person
        {
            public string Name { get; set; }
        }

        class Teacher : Person
        {
            public string Name { get; set; }
        }

2. Do you also intend to also update [CSharpModelGenerator](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs) to serialize inheritance metadata in model snapshots?

AndriySvyryd:
@bricelam 
1. Depends. When using a ModelBuilder with just the core conventions the `EntityType`s for `Student` and `Teacher` each will have a `Name` property, while the one for `Person` will have none.
When also using the relational convention the TPH convention would create a discriminator shadow property and uniquify all the column names that would end up mapped to `Persons` table. In this case it will result in columns `Name` and `Name1` named deterministically.
2. No, that is not part of this issue.

bricelam:
Sounds good. Thanks.

AndriySvyryd:
Discriminator API:
```
// If it's a string we just keep using type name for values
modelBuilder.Entity<Plant>().Discriminator<string>("Type_Of_Entity");
modelBuilder.Entity<Plant>().Discriminator(p => p.TypeOfEntity);

modelBuilder.Entity<Rose>().Discriminator(p => p.Genus).Value("Rose");
modelBuilder.Entity<Daisy>().Discriminator(p => p.Genus).Value("Daisy");

modelBuilder.Entity<Plant>().Discriminator(p => p.Genus);
modelBuilder.Entity<Rose>().Discriminator().Value("Rose");
modelBuilder.Entity<Daisy>().Discriminator().Value("Daisy");

// plus support for typeof() overloads
modelBuilder.Entity<Plant>().Discriminator(p => p.Genus)
[...]

PR #2655, opened 2015-07-17
Include BaseType info in model snapshot

smitpatel:
part of #2597 
Added to release to account for changes of #1704


bricelam:
:shipit:

PR #2656, opened 2015-07-17
Fix test projects that are still failing when running tests in VS

ajcvickers:
Same solution as for the other cases of this--app.config with binding redirect.


anpete:
:shipit:

PR #2659, opened 2015-07-17
Remove column order metadata and API

ajcvickers:
We don't use it for key ordering anymore and it isn't preserved across multiple migrations, so we are removing for now. Order can be changed by editing the migration. Could re-introduce a limited form in the future depending on feedback.


bricelam:
:shipit: :red_circle:

ajcvickers:
Merged

Varorbc:
The same field will be written in a basic model, to create a new type of time, it will go to inherit this basic model, if there are 100 classes to inherit this basic model, then you need to modify multiple migrated files, and restore Column Order metadata and API as long as the modification of the basic model it.

Varorbc:
@ajcvickers The same field will be written in a basic model, to create a new type of time, it will go to inherit this basic model, if there are 100 classes to inherit this basic model, then you need to modify multiple migrated files, and restore Column Order metadata and API as long as the modification of the basic model it.

PR #2663, opened 2015-07-18
Make non-contract methods on UpdateSqlGenerator protected

ajcvickers:
These methods were public on the abstract class even though they do not exist on the IUpdateSqlGenerator contract interface, through which all public access should go. I decided to make them protected to provide full flexibility for provider-specific derivations of the abstract class.


bricelam:
:shipit:

ajcvickers:
Merged

PR #2670, opened 2015-07-20
Make AddEntityFramework and AddProvider extensions idempotent

mikary:
Make AddEntityFramework and AddProvider extensions idempotent
Update TestHelpers classes for DRY service collection registration tests

General DI and csproj/json updates
@bricelam @maumar @ajcvickers


ajcvickers:
:shipit:

Issue #2635, opened 2015-07-15
Migrations: Tidy up History Repository API


bricelam:
The history repository area of Migrations could use some tidying up:
- [ ] Share more implementation across providers
- [ ] Reuse more components
    * `SqlStatementExecutor` (See #2497)
    * SQL query generation
    * DDL generation
- [ ] Make context key a first-class concept

PR #2672, opened 2015-07-20
Make ContextKey first-class

bricelam:
Part of #2635


src/EntityFramework7.Core/Internal/ModelSource.cs
```diff
@@ -46,7 +46,8 @@ public virtual IModel GetModel(DbContext context, IConventionSetBuilder conventi
 
             var productVersion = typeof(ModelSource).GetTypeInfo().Assembly
                 .GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
-            model[CoreAnnotationNames.ProductVersionAnnotation] = productVersion;
+            model.SetProductVersion(productVersion);
+            model.SetContextKey(context.GetType().FullName);
```
smitpatel: Should use `DisplayName()` function from `TypeExtensions` for better handling of nested classes.

bricelam: Then I can't do crazy hacks like this inside my provider. :wink: 
```C#
var context = Activator.CreateInstance(Type.GetType(model.GetContextKey()));
```
But on a serious note, is it worth it? It seems to only make generic types prettier and you can't use open generic types with Migrations anyway.

divega: So it doesn't not help with nested types?

smitpatel: As long as the string `Type.FullName` is kept intact, there won't be any issue.
Considering above usage, it is not worth. :smile:

ajcvickers: I didn't think we still had a restriction that the DbContext could not be a generic type? Or a nested type in a generic type for that matter. Is it only Migrations that imposes this restriction?

ajcvickers: I found a test that does `using (var context = new BloggingContext<TBlog>()` so it is at least nominally supported.

bricelam: We need to work out how to get an instance of it, but after that, you're right, it should be possible.

bricelam: Will update to use `DisplayName` and make a note to test generic context types with Migrations.

bricelam: @divega No, just generic types.

bricelam:
Revoking. We're going to try getting rid of the ContextKey instead.

PR #2676, opened 2015-07-21
Sequence API and behavioral consistency with core metadata

ajcvickers:
Specifically:
- Sequence objects now becomes a pure facade over their annotations in the model so that they can be mutated like first-class metadata objects.
- Removed default constants
- Pushed use of default name to be a purely HiLo thing
- Added generic fluent and non-generic overload for sequence type
- Restricted use of the annotation to inside the Sequence class implementation
- Renamed TryGetSequence to FindSequence

### DIFF
--- a/test/EntityFramework7.Relational.Tests/Metadata/RelationalBuilderExtensionsTest.cs
+++ b/test/EntityFramework7.Relational.Tests/Metadata/RelationalBuilderExtensionsTest.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Metadata.Builders;
 using Microsoft.Data.Entity.Tests;
 using Xunit;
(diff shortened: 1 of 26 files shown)


test/EntityFramework7.Relational.Tests/Metadata/RelationalBuilderExtensionsTest.cs
```diff
@@ -297,20 +296,20 @@ public void Can_create_named_sequence_with_convention_builder()
...
         {
             Assert.Equal("Snook", sequence.Name);
             Assert.Null(sequence.Schema);
             Assert.Equal(1, sequence.IncrementBy);
-            Assert.Equal(Sequence.DefaultStartValue, sequence.StartValue);
+            Assert.Equal(1, sequence.StartValue);
             Assert.Null(sequence.MinValue);
             Assert.Null(sequence.MaxValue);
-            Assert.Same(Sequence.DefaultType, sequence.ClrType);
+            Assert.Same(typeof(long), sequence.ClrType);
```
bricelam:
```C#
Assert.IsType<long>(sequence.ClrType);
```

bricelam:
:shipit:

PR #2681, opened 2015-07-22
Update Microsoft.CodeAnalysis.CSharp package

### LINKED ISSUE #2661: Commands: Microsoft.CodeAnalysis.CSharp missing on UWP
Repro:

1. Create new UWP (latest Win10 tooling) project
2. Open "project.json"
2. Add `"EntityFramework7.Commands" : "(latest version number)"` to the dependencies.
4. Verify the package has been installed by running the Nuget package manager console, `Get-Package -ProjectName {name}`
5. `Add-Migration` should be available, but instead it throws `The term 'Add-migration' is not recognized as the name of a cmdlet`

Commands on UWP are installed if the package is added via Nuget GUI or `Install-Package`

(There is probably not much we can do here except offer a workaround for users. This is a bug on UWP tooling.)

#1242

bricelam:
Also adds a missing dependency to it on Commands for the netcore50 framework.

Fixes #2661

### DIFF
--- a/src/EntityFramework.Commands/project.json
+++ b/src/EntityFramework.Commands/project.json
@@ -13,6 +13,10 @@
   "namedResource": {
     "EntityFramework.Commands.Strings": "Properties/Strings.resx"
   },
+  "dependencies": {
+    "EntityFramework.Relational": "7.0.0-*",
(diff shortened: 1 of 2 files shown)


ajcvickers:
:shipit:

PR #2683, opened 2015-07-22
NuGet Commands: Disable automatic positional binding

bricelam:
With this, only parameters with an explicit position can be unqualified. For example, where before you could do...
```ps1
Add-Migration MyMigration MyProject
```
...you must now do...
```ps1
Add-Migration MyMigration -Project MyProject
```
This aligns better with the DNX commands.


natemcmaster:
:shipit:

PR #2685, opened 2015-07-22
Refactor core metadata extension classes so providers can re-use

ajcvickers:
Specifically:
- Build into the core classes the ability to specify a provider annotation prefix and have that used first before looking up with the relational prefix. This lets the same classes be used by both the relational assembly and providers
- Collapse the read-only and read-write classes together. This means that an application could now cast from the read-only interface to the mutable class, which would break in cases where the interface is not backed by a mutable model. But this is no different that the rest of our core metadata.
- Remove empty interfaces where providers were not adding additional metadata to the extensions.


AndriySvyryd:
I don't like the cast to mutable in principle, but in this case I don't think it would cause any problems in practice.
:shipit:

PR #2689, opened 2015-07-22
Update internal builders to manipulate relational metadata using API

ajcvickers:
Extends the API sugar so it can be used to set annotations without falling back to direct manipulation of the annotations themselves.

This completes the API review note to access relational metadata through names, except for in Migrations where we discussed this in a different API review which is being handled by @bricelam


AndriySvyryd:
:shipit:

ajcvickers:
Merged

PR #2696, opened 2015-07-23
[small] Fix to CSharpHelper

### LINKED ISSUE #2674: Migrations: Dot missing from namespace WebApplication1Migrations
It looks like 95e318b143ac660cb56fe30c0e81e5f94ba69387 removed the dots separating the namespace parts in the scaffolded code. Instead of `WebApplication1.Migrations`, the namespace is now `WebApplication1Migrations`.

natemcmaster:
Fix #2674.

### DIFF
--- a/src/EntityFramework.Commands/Utilities/CSharpHelper.cs
+++ b/src/EntityFramework.Commands/Utilities/CSharpHelper.cs
@@ -138,7 +138,7 @@ namespace Microsoft.Data.Entity.Commands.Utilities
                 builder.Append(name.Substring(partStart));
             }
 
-            if (!SyntaxFacts.IsIdentifierStartCharacter(builder[0]))
+            if (builder.Length == 0 || !SyntaxFacts.IsIdentifierStartCharacter(builder[0]))
             {
                 builder.Insert(0, "_");
             }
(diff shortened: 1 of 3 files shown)


lajones:
:shipit:

src/EntityFramework.Commands/Utilities/CSharpHelper.cs
```diff
@@ -166,6 +166,23 @@ public virtual string Identifier([NotNull] string name, [CanBeNull] ICollection<
...
         }
 
+        public virtual string Namespace([NotNull] params string[] name)
+        {
+            Check.NotNull(name, nameof(name));
+
+            var @namespace = new StringBuilder();
+            foreach (var piece in name.Where(p => !string.IsNullOrEmpty(p)).SelectMany(p => p.Split('.')))
+            {
+                var identifier = Identifier(piece);
```
bricelam: Hmm, not sure it's right to reuse `Identifier`. I don't think you can escape keywords in namespaces using `@int`. Also fixing-up to remove the leading `_` won't always work because you can't have a namespace named '13', for example.

natemcmaster: Identifier should add the leading _, not remove it. This is a valid namespace:
```c#
namespace @int._13
{
    public class Abc { }
}
```

bricelam: Sounds good.

Issue #1924, opened 2015-03-27
InternalRelationshipBuilder.NavigationToDependent should configure uniqueness


AndriySvyryd:
Currently setting a navigation to dependent that doesn't match the relationship uniqueness will throw, instead it should change it.

PR #2698, opened 2015-07-23
Bugfixes in relationshipbuilder

smitpatel:
Problem:
Previously all the modifications to relationshipbuilder was possible after navigations are added therefore preserving all the information provided to `Relationship()`. With the addition of OnNavigationAdded Conventions, the relationshipbuilder can be mutated before it is built fully which causes missing relationship information, upgrading ConfigurationSource for relationshipbuilder at wrong times, Removing entityTypes which are still exploring relationship (but not built them yet).

Solution:
`ReplaceForeignKey` preserves the configurationSource unless defining properties of foreignkey changed.
EntityTypes will be added with ConfigurationSource.DataAnnotation or higher and will be changed to the original configuration source once it finishes configuring it by conventions. (this stops entityTypes being removed which hasn't explored the relationships yet)
This requires behavior of Ignore to be changed allowing to remove the model item added with same source.

Build status:
Working

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
@@ -74,7 +74,6 @@ namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
 
                 foreach (var navigationPairCandidate in navigationPairCandidates)
                 {
-                    var targetEntityTypeBuilder = navigationPairCandidate.Key;
                     var navigationCandidates = navigationPairCandidate.Value.Item1;
                     var reverseNavigationCandidates = navigationPairCandidate.Value.Item2;
 
@@ -93,11 +92,20 @@ namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
 
                     foreach (var navigationCandidate in navigationCandidates)
                     {
-                        TryBuildRelationship(entityTypeBuilder, targetEntityTypeBuilder, navigationCandidate, reverseNavigationCandidates.SingleOrDefault());
+                        var targetEntityTypeBuilder = entityTypeBuilder.ModelBuilder.Entity(FindCandidateNavigationPropertyType(navigationCandidate), ConfigurationSource.Convention);
+                        TryBuildRelationship(
+                            entityTypeBuilder,
+                            targetEntityTypeBuilder,
+                            navigationCandidate,
+                            reverseNavigationCandidates.SingleOrDefault());
                     }
                 }
             }
 
+            // While running conventions on entityType, its source will be DataAnnotation or higher
+            // Which means that entity won't be removed while being configured even if it is unreachable
+            // This takes care of removing such unreachable entities (being run after we are done building relationships using this entity)
+            entityTypeBuilder.ModelBuilder.RemoveEntityTypesUnreachableByNavigations(ConfigurationSource.DataAnnotation);
             return entityTypeBuilder;
         }
 
--- a/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -75,6 +75,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                 () => Metadata.SetPrimaryKey(properties),
                 key => new InternalKeyBuilder(key, ModelBuilder),
                 ModelBuilder.ConventionDispatcher.OnKeyAdded,
+                configurationSource,
                 configurationSource);
 
             ReplaceConventionShadowKeys(keyBuilder.Metadata);
@@ -139,6 +140,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                 () => Metadata.AddKey(properties),
                 key => new InternalKeyBuilder(key, ModelBuilder),
                 ModelBuilder.ConventionDispatcher.OnKeyAdded,
+                configurationSource,
                 configurationSource);
         }
 
@@ -201,6 +203,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                     () => Metadata.AddProperty(propertyName, propertyType, shadowProperty),
                     property => new InternalPropertyBuilder(property, ModelBuilder, configurationSource),
                     ModelBuilder.ConventionDispatcher.OnPropertyAdded,
+                    configurationSource,
                     configurationSource);
             }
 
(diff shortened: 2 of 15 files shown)


src/EntityFramework.Core/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
```diff
@@ -93,11 +92,16 @@ public virtual InternalEntityTypeBuilder Apply(InternalEntityTypeBuilder entityT
 
                     foreach (var navigationCandidate in navigationCandidates)
                     {
-                        TryBuildRelationship(entityTypeBuilder, targetEntityTypeBuilder, navigationCandidate, reverseNavigationCandidates.SingleOrDefault());
+                        TryBuildRelationship(
+                            entityTypeBuilder,
+                            entityTypeBuilder.ModelBuilder.Entity(FindCandidateNavigationPropertyType(navigationCandidate), ConfigurationSource.Convention),
```
AndriySvyryd: Assign the found entity type to a variable so it's more readable

src/EntityFramework.Core/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
```diff
@@ -93,11 +92,16 @@ public virtual InternalEntityTypeBuilder Apply(InternalEntityTypeBuilder entityT
...
+                            navigationCandidate,
+                            reverseNavigationCandidates.SingleOrDefault());
                     }
                 }
             }
 
+            entityTypeBuilder.ModelBuilder.RemoveEntityTypesUnreachableByNavigations(ConfigurationSource.DataAnnotation);
```
AndriySvyryd: Add a comment why this is needed so we don't forget

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -366,7 +366,7 @@ public virtual bool Ignore([NotNull] string propertyName, ConfigurationSource co
             var property = Metadata.FindDeclaredProperty(propertyName);
             if (property != null)
             {
-                if (!RemoveProperty(property, configurationSource, canOverrideSameSource: false).HasValue)
+                if (!RemoveProperty(property, configurationSource, canOverrideSameSource: true).HasValue)
                 {
                     if (configurationSource == ConfigurationSource.Explicit)
```
AndriySvyryd: No need to throw for explicit as it will never fail

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -1092,6 +1092,11 @@ private InternalIndexBuilder Index(IReadOnlyList<Property> properties, Configura
                 return null;
             }
 
+            if (navigationToDependentName != null && isUnique.HasValue)
+            {
+                builder = builder.Unique(isUnique.Value, configurationSource);
```
AndriySvyryd: This is ok for now, but ideally should be handled in the Navigation() method. At least add a comment so it can be moved when #1924 is fixed

src/EntityFramework.Core/Metadata/Internal/InternalModelBuilder.cs
```diff
@@ -32,26 +32,56 @@ public InternalModelBuilder([NotNull] Model metadata, [NotNull] ConventionSet co
...
+            }
+
+            var entityTypeBuilder = _entityTypeBuilders.GetOrAdd(
+                    () => Metadata.FindEntityType(name),
+                    () => Metadata.AddEntityType(name),
+                    entityType => new InternalEntityTypeBuilder(entityType, ModelBuilder),
+                    ConventionDispatcher.OnEntityTypeAdded,
+                    configurationSource.Max(ConfigurationSource.DataAnnotation));
+
+            return entityTypeBuilder != null ? _entityTypeBuilders.ChangeConfigurationSource(entityTypeBuilder.Metadata, configurationSource) : null;
```
AndriySvyryd: This should be handled in the method passed as the onNewKeyAdded parameter to reduce code duplication

src/EntityFramework.Core/Metadata/Internal/InternalModelBuilder.cs
```diff
@@ -93,7 +123,7 @@ public virtual bool Ignore([NotNull] string name, ConfigurationSource configurat
             var entityType = Metadata.FindEntityType(name);
             if (entityType != null)
             {
-                if (!Remove(entityType, configurationSource, canOverrideSameSource: false))
+                if (!Remove(entityType, configurationSource, canOverrideSameSource: true))
                 {
                     if (configurationSource == ConfigurationSource.Explicit)
```
AndriySvyryd: Exception will never be thrown

src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -200,6 +200,7 @@ public virtual InternalRelationshipBuilder Invert(ConfigurationSource configurat
                 null,
                 ((IForeignKey)Metadata).IsUnique,
                 _isRequiredConfigurationSource.HasValue ? ((IForeignKey)Metadata).IsRequired : (bool?)null,
+                true,
```
AndriySvyryd: Bool parameter values should be qualified

src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -469,6 +471,7 @@ private InternalRelationshipBuilder ReferenceInvertIfNeeded(EntityType entityTyp
                 principalProperties,
                 isUnique,
                 isRequired,
+                upgradeSource,
```
AndriySvyryd: "shouldUpgradeSource" better conforms with the guidelines

src/EntityFramework.Core/Metadata/Internal/MetadataDictionary.cs
```diff
@@ -64,6 +64,21 @@ public class MetadataDictionary<TKey, TValue>
...
+            {
+                _values.Remove(key);
+                tuple = new Tuple<TValue, ConfigurationSource>(tuple.Item1, configurationSource);
+                _values.Add(key, tuple);
+
+                return tuple.Item1;
+            }
+
+            return null;
+        }
```
AndriySvyryd: Instead of adding this method you could add another parameter to GetOrAdd for the temporary configuration source while `onNewKeyAdded` is invoked

AndriySvyryd:
Reviewed

smitpatel:
Updated.

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -381,7 +379,7 @@ public virtual bool Ignore([NotNull] string propertyName, ConfigurationSource co
             var navigation = Metadata.FindDeclaredNavigation(propertyName);
             if (navigation != null)
             {
-                if (Navigation(null, navigation.ForeignKey, navigation.PointsToPrincipal(), configurationSource, canOverrideSameSource: configurationSource != ConfigurationSource.Explicit) == null)
+                if (Navigation(null, navigation.ForeignKey, navigation.PointsToPrincipal(), configurationSource, canOverrideSameSource: true) == null)
                 {
                     if (configurationSource == ConfigurationSource.Explicit)
```
AndriySvyryd: Remove this

AndriySvyryd:
Looks good so far

smitpatel:
Fixed all tests. Build working now.

AndriySvyryd:
:shipit:

PR #2699, opened 2015-07-23
ModelSnapshot: Topologically Sort entityType by hierarchy

### LINKED ISSUE #2669: Model Snapshot: Group by Entity Type
Per [our previous discussion](https://github.com/aspnet/EntityFramework/pull/2640#discussion-diff-34847063), instead of configuring the model in different phases, we should group configuration by entity type. To do this, the high-level algorithm should be:

1. Topologically sort entity types by hierarchy.
2. Configure everything but relationships/foreign keys
3. Configure relationships/foreign keys.

smitpatel:
resolves #2669

### DIFF
--- a/src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
+++ b/src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
@@ -42,16 +42,26 @@ namespace Microsoft.Data.Entity.Commands.Migrations
                 stringBuilder.AppendLine(";");
             }
 
-            GenerateEntityTypes(model.EntityTypes, stringBuilder);
+            GenerateEntityTypes(Sort(model.EntityTypes), stringBuilder);
         }
 
         [Flags]
         protected enum GenerateEntityTypeOptions
         {
             Declared = 1,
-            BaseType = 2,
-            Relationships = 4,
-            Full = Declared | BaseType | Relationships
+            Relationships = 2,
+            Full = Declared | Relationships
+        }
+
+        private IReadOnlyList<IEntityType> Sort(IReadOnlyList<IEntityType> entityTypes)
+        {
+            var entityTypeGraph = new Multigraph<IEntityType, int>();
+            entityTypeGraph.AddVertices(entityTypes);
+            foreach (var entityType in entityTypes.Where(et => et.BaseType != null))
+            {
+                entityTypeGraph.AddEdge(entityType.BaseType, entityType, 0);
+            }
+            return entityTypeGraph.TopologicalSort();
         }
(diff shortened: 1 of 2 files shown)


src/EntityFramework.Commands/Migrations/CSharpModelGenerator.cs
```diff
@@ -48,16 +48,15 @@ public virtual void Generate([NotNull] IModel model, [NotNull] IndentedStringBui
                 stringBuilder.AppendLine(";");
             }
 
-            GenerateEntityTypes(model.EntityTypes, stringBuilder);
+            GenerateEntityTypes(new ModelBaseTypeGraphAdapter(model).TopologicalSort(), stringBuilder);
```
AndriySvyryd: Why can't you use a Multigraph instead of creating a new Graph implementation?

smitpatel: Multigraph takes Edge of a particular type. In the case of base type, there are no edge objects. Moreover edges are defined by themselves by BaseType property and doesn't required to be added explicitly. Also topological sort of multigraph is implemented to handle breaking/formatting cycles which doesn't arise in basetype.
Multigraph can also be used in this place with building the graph appropriately. Whichever way it is preferred is fine by me.

AndriySvyryd: You can use any type for the Edge here, it doesn't matter
If you are adding a new implementation of the topological sort then you need to add tests for it, it's your choice

smitpatel: @bricelam - what you say? :smile:

bricelam: Use `int` and set it to `0`. :smile: We don't need to break cycles, so they doesn't actually matter.

bricelam: @mikary Did the a Graph implementation that didn't have edge objects go away?

mikary: The old implementation without cycle breaking or edge values `BidirectionalAdjacencyListGraph` went away in b4bf8b291e4d0ad2e4c642e41faf02d5922ea64b because it wasn't being used. If it makes sense for this situation, we could bring it back.

smitpatel: Switched to MultiGraph

bricelam:
:shipit:

PR #2701, opened 2015-07-24
Remove .Relational for consuming metadata at runtime

ajcvickers:
At runtime, metadata should be consumed, usually indirectly, through the current provider's API to ensure that any provider-specific overrides are honored. This means it is never correct to use .Relational to consume metadata after the model has been built, although it is often valid to use it to set metadata when constructing the model.

This change removes the read-only interface-based .Relational methods and fixes everywhere they were being misused.


AndriySvyryd:
:shipit:

PR #2702, opened 2015-07-24
Code First inheritance fixes

### LINKED ISSUE #2671: Metadata: Prohibit mixed hierarchies
Assuming that the principal use case for entity type hierarchies is querying from the root CLR type (like using Set\<TEntity\>) then adding shadow entity types in the hierarchy or making its topology different from the corresponding CLR hierarchy doesn't add much value and could make the implementation unnecessarily complex.
Purely shadow hierarchies would still be valid.

AndriySvyryd:
Prohibit mixed hierarchies and add more consistency checks to core metadata

When a relationship on the derived type conflicts with one on the new base type, but the inverse navigation points to the derived type then remove the relationship from the base type.

Fix inheritance bugs in the internal builders and conventions

Rename InternalRelationshipBuilder.NavigationToPrincipal and NavigationToDependent to DependentToPrincipal and PrincipalToDependent

Fixes #2671


ajcvickers:
"shipit"

PR #2710, opened 2015-07-24
Configure relationship uniqueness while setting navigationToDependent

### LINKED ISSUE #1924: InternalRelationshipBuilder.NavigationToDependent should configure uniqueness

smitpatel:
Resolves #1924

### DIFF
--- a/src/EntityFramework.Core/Metadata/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/EntityType.cs
@@ -611,7 +611,7 @@ namespace Microsoft.Data.Entity.Metadata
             Debug.Assert(targetClrType != null, "targetClrType != null");
             if (navigation.IsCollection())
             {
-                var elementType = clrProperty.PropertyType.TryGetElementType(typeof(IEnumerable<>));
+                var elementType = clrProperty.PropertyType.TryGetSequenceType();
 
                 if (elementType == null
                     || !elementType.GetTypeInfo().IsAssignableFrom(targetClrType.GetTypeInfo()))
--- a/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -312,8 +312,26 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                     _ignoredProperties.Value.Remove(navigationName);
                 }
 
-                fkOwner._relationshipBuilders.Value.UpdateConfigurationSource(foreignKey, configurationSource);
-                var navigation = Metadata.AddNavigation(navigationName, foreignKey, pointsToPrincipal);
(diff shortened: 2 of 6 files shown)


smitpatel:
If this looks fine then I will add tests.

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -581,7 +581,7 @@ public virtual Navigation AddNavigation([NotNull] string name, [NotNull] Foreign
             Debug.Assert(targetClrType != null, "targetClrType != null");
             if (navigation.IsCollection())
             {
-                var elementType = clrProperty.PropertyType.TryGetElementType(typeof(IEnumerable<>));
+                var elementType = clrProperty.PropertyType.TryGetSequenceType();
```
smitpatel: @AndriySvyryd - Any reason we were using `TryGetElementType(typeof(IEnumerable<>))` instead of `TryGetSequenceType`? `TryGetSequenceType` checks for `IEnumerable<>` and `IAsyncEnumerable<>` both.

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -315,8 +315,19 @@ private bool CanRemove(ForeignKey foreignKey, ConfigurationSource configurationS
...
                 }
 
-                fkOwner._relationshipBuilders.Value.UpdateConfigurationSource(foreignKey, configurationSource);
-                var navigation = Metadata.AddNavigation(navigationName, foreignKey, pointsToPrincipal);
+                if (!pointsToPrincipal)
+                {
+                    var elementType = Metadata.ClrType?.GetPropertiesInHierarchy(navigationName).FirstOrDefault()?.PropertyType.TryGetSequenceType();
+                    if (elementType == null)
+                    {
+                        builder = builder.Unique(true, configurationSource);
```
AndriySvyryd: Only set uniqueness if the CLR property was found

src/EntityFramework.Core/Metadata/EntityType.cs
AndriySvyryd: Probably because that method didn't exist yet.

AndriySvyryd:
Also change `InternalRelationshipBuilder.Unique` to set `PrincipalToDependent` to `null` if it's not compatible

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
AndriySvyryd: It also seems useful to put the `ClrType` on the `Navigation`, so we don't need to look it up again.

smitpatel:
Added tests.

AndriySvyryd:
:shipit:

PR #2722, opened 2015-07-27
Query: Allow provider customization of SQL parameter names in Default…

anpete:
…QuerySqlGenerator.


bricelam:
:shipit: I'm curious, what would a provider need to do beside changing the prefix?

anpete:
@bricelam Remove disallowed characters such as leading underscores.

PR #2733, opened 2015-07-28
Migrations: Test & tidy up the history repository

### LINKED ISSUE #2497: SqlServerHistoryRepository.Exists doesn't use SqlStatementExecutor
Everywhere else in the stack uses this abstraction when performing ExecuteNonQuery and ExecuteScalar, so we should probably be consistent and use it here as well unless there is a reason it can't be used.

### LINKED ISSUE #2635: Migrations: Tidy up History Repository API

### LINKED ISSUE #2695: Migrations revert in the wrong order when going to "0"
Apply-Migration 0 unapplies in the wrong order. It starts with the oldest migration. The first should be last and the last should be first.

bricelam:
Resolves #2497, resolves #2635 & fixes #2695


test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqliteFixture.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
ErikEJ: Wrong filename...

src/EntityFramework.Relational/Migrations/History/IHistoryRepository.cs
```diff
@@ -11,12 +10,13 @@ namespace Microsoft.Data.Entity.Migrations.History
...
-        MigrationOperation GetInsertOperation([NotNull] IHistoryRow row);
-        MigrationOperation GetDeleteOperation([NotNull] string migrationId);
-        string BeginIfNotExists([NotNull] string migrationId);
-        string BeginIfExists([NotNull] string migrationId);
-        string EndIf();
+        IReadOnlyList<HistoryRow> GetAppliedMigrations();
+        string GetCreateScript();
+        string GetCreateIfNotExistsScript();
+        string GetInsertScript([NotNull] HistoryRow row);
+        string GetDeleteScript([NotNull] string migrationId);
```
natemcmaster: Why are we switching away from using MigrationOperation? Doesn't this mean we have now duplicated SQL generation for deleting rows?

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
```diff
@@ -111,7 +131,7 @@ public SqlStatementExecutor([NotNull] ILoggerFactory loggerFactory)
 
                             command.ExecuteNonQuery();
                         }
-                        return null;
+                        return default(object);
```
natemcmaster: Should this be `default(T)`?

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqliteFixture.cs
bricelam: Oops :astonished:

src/EntityFramework.Relational/Migrations/History/IHistoryRepository.cs
bricelam: Yes, and I would like to get rid of that duplication. I tried my best, but was not able to reuse the Update pipeline in it's current state. The default implementation of this interface, `HistoryRepository` should, for the most part, hide that duplication from provider writers. In the future we should be able to swap it out to use more of the Update pipeline and not break providers.

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqliteFixture.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            services.AddEntityFramework().AddSqlServer();
+        }
+
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            optionsBuilder.UseSqlServer("" +
+                new SqlConnectionStringBuilder
+                {
+                    DataSource = @"(localdb)\MSSQLLocalDB",
```
natemcmaster: Hard coded connection strings :shock: ... we still need a better solution for configuring our testing infrastructure.

natemcmaster:
:shipit: (with TODOs on update pipeline)

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
bricelam: There is no `T` here. It could also remain null and the call could be updated to `Execute<object>(`, but I liked this better.

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqliteFixture.cs
bricelam: When in Rome... :wink:

src/EntityFramework.Relational/Migrations/History/IHistoryRepository.cs
bricelam: Filed #2737 too.

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
natemcmaster: :facepalm: you're right. I didn't look carefully at the context of this change. Carry on.

PR #2734, opened 2015-07-28
Make look up by alternate key faster to help with fixup time

ajcvickers:
Changes the identity map to keep track of entities by any principal key, thereby allowing fast lookup of entities at the principal end of a relationship. Also, slight change to tracking of entries created for entities that are not yet tracked. These are now tracked as weak references so that in the case where an application is looping over many detached entities without actually tracking most of them the memory use will be bounded.


bricelam:
:shipit:

anpete:
How much faster?

Issue #2713, opened 2015-07-24
RevEng: Refactor to split schema acquisition from more common functionality


divega:
Currently the reverse engineering provider for SQL Server contains the logic that retrieves the schema information from the database into an IModel and also the logic that translates that into an enriched IModel containing entities, relationships and annotations that is ready for the code-gen phase. 

Ideally, providers should only need to implement the initial schema acquisition and could delegate the rest to common functionality in EF. 

One proposal to address this is to lift the common functionality into base relational provider classes. This would allow the providers to still override the behaviors that they wish to. Creating a reverse engineering provider for a database in which schema acquisition is fundamentally different from SQL Server should help polish such design.

divega:
@bricelam re the last point, do we have a bug for reverse engineering for SQLite that we can talk about in triage? The one that was closed today was specifically for UWP projects, and I couldn't find one for SQLite.

bricelam:
Created #2721

lajones:
@ErikEJ Please note this might change your SQLCE provider too (make it simpler, I hope). 

I'm planning on having an example checked in soon - but basics are there's a new abstract base class, ReverseEngineeringMetadataModelProvider, which does all the conversion from the relational IModel to the enhanced IModel used for code-gen. You should simply need to override the method ConstructRelationalModel() and produce an IModel that directly represents the DB i.e. has entities (=  tables), properties (= columns) and foreign keys. After that ReverseEngineeringMetadataModelProvider takes over. However there will be requirements on the way the relational IModel is set up (what annotations we expect etc). I'll list them when I have the complete list.

ErikEJ:
Sounds great, will save me 600 lines of code, I did not make any significant changes anyway: https://github.com/ErikEJ/EntityFramework7.SqlServerCompact/blob/master/src/Design40/ReverseEngineering/SqlCeMetaDataModelProvider.cs

ErikEJ:
Why is this line actually neccessary?
https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs#L616

lajones:
[...]

Issue #2465, opened 2015-06-24
RevEng: Strive for code generation templates that are not provider specific


divega:
Currently we may generate DbContext code that uses `ForSqlServer()` in the model generation or at least we may have logic in the template that is SQL Server specific. This issue is about avoiding that if possible and using relational APIs instead.

There is value in having templates that can be customized once and then used against different providers. There is also value in not requiring providers to develop their own templates in order to fully support the reverse engineering experiences we want to enable, e.g. what is described in #2415 (although this is less important if they can just ship copies of our template in their reveng packages).

On the other hand we are not sure yet that our templates won't require provider specific logic and we can't assume that other providers won't need to extend code generation to accommodate their own concepts. 

This came up with in a conversation with @ajcvickers and I just wanted to make sure we capture it. He and @lajones can provide mode details about what we have now, what is possible and the pros and cons of this.

ErikEJ:
Is this a good time to start investigating using Reverse Engineering  with SQLCE, or is it too early?

divega:
@ErikEJ I personally believe it is a good time for you to get started with it. If you hit issues or have any feedback we can make adjustments.

ErikEJ:
@divega  Sounds great, is there some docs/design specs? Or a drawing? :smile:  
I have a few questions:
1: Does this info still apply: https://stackoverflow.com/questions/29300777/is-there-an-entity-framework-7-database-first-poco-generator? 
2: If so, how does the command know which provider to use?
3: Is the rev-eng package simply installed as needed? (EntityFramework.SqlServer.Design)
4: Will I "simply" implement the equivalent of EntityFramework.SqlServer.Design ?

divega:
No docs yet. I will defer to @lajones on answering any questions and or drawing it for you :smile:

lajones:
@ErikEJ This feels like a more general question about how RevEng currently works - moving to the general issue tracking RevEng: #830. But note: this issue will have an effect on how you implement that.

lajones:
Fixed with commit 5c30958

lajones:
@ErikEJ Hi Erik. The changes are now in. I'll write it up but broadly you should only have to implement the ConstructRelationalModel() method in your provider and override the ExtensionsProvider getter now.

See https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs for an example.

[...]

Issue #623, opened 2014-09-04
.NET Native support


divega:
Supporting running EF based applications in .NET native is important, especially for Windows Universal Apps in the short term. This was always part of the plan for EF7 and we have made specific choices to make this easier, e.g. avoiding taking dependencies on functionality that is not available when working in .NET Native, investigating how we could use Roslyn to optimize some aspects of the application using EF at compile time, etc. This is a general placeholder for all remaining work we may need to do to support .NET Native. We need to start with investigating what the gaps are and once we have made progress we can create new work items for individual tasks we identify.

bricelam:
See also [Migrating Your Windows Store App to .NET Native](https://msdn.microsoft.com/en-us/library/dn600634.aspx)

divega:
@natemcmaster here is the.NET Native issue I promised I would assign to you :smile:

natemcmaster:
Current status: a blank Windows Universal app with EF 7 referenced causes the app to crash during startup before the app can start.  The same app runs just fine when not using the .NET Native toolchain.

Attempting to debug using [the help instructions provided in the dialog](https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(vs.debug.error.app_activation_failure);k(TargetFrameworkMoniker-.NETCore,Version%3Dv4.5.1)&rd=true) do not provide an answer as to why this crashes app.

![image](https://cloud.githubusercontent.com/assets/2696087/8014792/70670b3e-0b88-11e5-9bc7-551d48866b84.png)

bricelam:
@natemcmaster Remind me to show you where to get the latest Tools for Windows 10. You may encounter less red herrings than with the public previews.

natemcmaster:
Status updated to blocked. Waiting for now on improvements to the .NET Native toolchain that will allow us to find the source of these crashes.

bricelam:
I just tried this for fun on the latest bits and got:

> Exception thrown: '$65_System.Reflection.MissingRuntimeArtifactException' in Unknown Module.
An exception of type '$65_System.Reflection.MissingRuntimeArtifactException' occurred in Unknown Module. but was not handled in user code
[...]

PR #2741, opened 2015-07-29
Moving RevEng provider code down into Relational.Design (where possible)

lajones:
Fix issue #2465 - making code generation templates non-provider specific.
Fix issue #2713 - allow providers to only provide the relational model of the DB and have the inherited base class provide the enhanced model for code-generation.
Also fix up some error messages.
Also updated to produce AddGeneratedNever() on a property which the KeyConvention would normally set up as AddGeneratedOnAdd() but where the underlying database column does not have identity set.


src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+        {
+            get
+            {
+                return ReadFromResource(
+                    typeof(ReverseEngineeringMetadataModelProvider).GetTypeInfo().Assembly,
+                    EntityTypeTemplateResourceName);
+            }
+        }
+
+        public virtual EntityTypeCodeGeneratorHelper EntityTypeCodeGeneratorHelper(EntityTypeGeneratorModel model)
```
lajones: @ajcvickers It feels odd to me that the DbContextCodeGeneratorHelper method is abstract whereas the EntityTypeCodeGeneratorHelper method is virtual. That arises because DbContextCodeGeneratorHelper has the IRelationalMetadataExtensionProvider we discussed. I can make the EntityTypeCodeGeneratorHelper method abstract and simply override it SqlServerMetadataModelProvider to return new EntityTypeCodeGeneratorHelper() but that seems unnecessary?

src/EntityFramework.Relational.Design/ReverseEngineering/EntityTypeCodeGeneratorHelper.cs
```diff
@@ -27,5 +29,101 @@ public virtual IEnumerable<IProperty> OrderedEntityProperties
...
+                    // find navigation properties for foreign keys from another EntityType which reference this EntityType
+                    foreach (var foreignKey in otherEntityType
+                        .GetForeignKeys().Where(fk => fk.PrincipalEntityType == GeneratorModel.EntityType))
+                    {
+                        var navigationPropertyName =
+                            (string)foreignKey[ReverseEngineeringMetadataModelProvider.AnnotationNamePrincipalEndNavPropName];
+                        if (((EntityType)otherEntityType)
+                            .FindAnnotation(ReverseEngineeringMetadataModelProvider.AnnotationNameEntityTypeError) == null)
+                        {
+                            if (!foreignKey.IsUnique)
```
natemcmaster: Why does this check matter?

src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGeneratorHelper.cs
```diff
@@ -29,8 +29,14 @@ protected DbContextCodeGeneratorHelper([NotNull] DbContextGeneratorModel generat
 
         public virtual DbContextGeneratorModel GeneratorModel { get; }
 
-        // default ordering is by Name, which is what we want here
-        public virtual IEnumerable<IEntityType> OrderedEntityTypes() => GeneratorModel.MetadataModel.EntityTypes;
+        public virtual IEnumerable<IEntityType> OrderedEntityTypes()
+        {
+            // default ordering is by Name, which is what we want here but
+            // do not configure EntityTypes for which we had an error when generating
+            return GeneratorModel.MetadataModel.EntityTypes
```
natemcmaster: Can use c# 6.0 fat arrows here.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+        public const string AnnotationPrefix = "ReverseEngineeringMetadataModelProvider:";
+        public const string AnnotationNameDependentEndNavPropName = AnnotationPrefix + "DependentEndNavPropName";
+        public const string AnnotationNamePrincipalEndNavPropName = AnnotationPrefix + "PrincipalEndNavPropName";
+        public const string AnnotationNameEntityTypeError = AnnotationPrefix + "EntityTypeError";
+
+        public const string NavigationNameUniquifyingPattern = "{0}Navigation";
+        public const string SelfReferencingPrincipalEndNavigationNamePattern = "Inverse{0}";
+
+        public static readonly string DbContextTemplateResourceName =
+            typeof(ReverseEngineeringMetadataModelProvider).GetTypeInfo().Assembly.GetName().Name
```
natemcmaster: Can we just use a string literal? (Thinking about #623)

src/EntityFramework.Relational.Design/ReverseEngineering/EntityTypeCodeGeneratorHelper.cs
lajones: If the foreign key is unique you have a 1:1 (or 0..1:0..1 etc) relationship. NavPropInitializers are used to create a new empty HashSet for a NavigationProperty which represents a 1:N relationship.

src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGeneratorHelper.cs
lajones: OK. Will update.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+
+        private readonly Dictionary<EntityType, EntityType> _relationalEntityTypeToCodeGenEntityTypeMap =
+            new Dictionary<EntityType, EntityType>();
+        private readonly Dictionary<Property, Property> _relationalPropertyToCodeGenPropertyMap =
+            new Dictionary<Property, Property>();
+
+        public virtual ILogger Logger { get; private set; }
+        public virtual ModelUtilities ModelUtilities { get; private set; }
+
+        public ReverseEngineeringMetadataModelProvider([NotNull] ILogger logger, [NotNull] ModelUtilities modelUtilities)
```
natemcmaster: Since this is in the reverseengineering namespace, can we drop "ReverseEngineering" from the name of classes?

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+            return ConstructCodeGenModel(relationalModel, nameMapper);
+        }
+
+        public virtual MetadataModelNameMapper GetNameMapper([NotNull] IModel relationalModel)
+        {
+            Check.NotNull(relationalModel, nameof(relationalModel));
+
+            return new MetadataModelNameMapper(
+                relationalModel,
+                entity => ((EntityType)entity).Relational().TableName,
```
natemcmaster: I believe @ajcvickers said we should be avoiding calling this extension since we may need provider-specific annotations. Can we abstract this?

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+            }
+        }
+
+        public virtual EntityTypeCodeGeneratorHelper EntityTypeCodeGeneratorHelper(EntityTypeGeneratorModel model)
+        {
+            return new EntityTypeCodeGeneratorHelper(model);
+        }
+
+        public virtual void AddReferencesForTemplates(MetadataReferencesProvider metadataReferencesProvider)
+        {
```
natemcmaster: Empty method?

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
lajones: We could. Obviously if the assembly name changes we'd need to remember to change this. And there have been assembly name changes recently (EF -> EF7 -> EF), but then again the tests would fail if we forgot. So I'm OK with the string approach if we're taking that approach generally throughout the code.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+
+            var codeGenModel = new Entity.Metadata.Model();
+            foreach (var relationalEntityType in relationalModel.EntityTypes.Cast<EntityType>())
+            {
+                var codeGenEntityType = codeGenModel
+                    .AddEntityType(nameMapper.EntityTypeToClassNameMap[relationalEntityType]);
+                _relationalEntityTypeToCodeGenEntityTypeMap[relationalEntityType] = codeGenEntityType;
+                codeGenEntityType.Relational().TableName = relationalEntityType.Relational().TableName;
+                codeGenEntityType.Relational().Schema = relationalEntityType.Relational().Schema;
+                var errorMessage = relationalEntityType[AnnotationNameEntityTypeError];
```
natemcmaster: It seems odd to me that we use Annotations as the mechanism for passing error messages...not sure if there is a better way, but it just doesn't seem like this is consistent with how annotations are used in other places in our stack.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+
+        public static readonly string DbContextTemplateResourceName =
+            typeof(ReverseEngineeringMetadataModelProvider).GetTypeInfo().Assembly.GetName().Name
+            + ".ReverseEngineering.Templates.DbContextTemplate.cshtml";
+
+        public static readonly string EntityTypeTemplateResourceName =
+            typeof(ReverseEngineeringMetadataModelProvider).GetTypeInfo().Assembly.GetName().Name
+            + ".ReverseEngineering.Templates.EntityTypeTemplate.cshtml";
+
+        private readonly Dictionary<EntityType, EntityType> _relationalEntityTypeToCodeGenEntityTypeMap =
```
natemcmaster: Nit: this is a really long variable name.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
lajones: There was a time when this class was not abstract. But now it is I could make this abstract too. However in the SQL Server case we actually don't need to pass any additional references - so my thought was why make the provider provide this method when I could do provide base functionality which would work for most providers here.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+                if (primaryKey != null)
+                {
+                    codeGenEntityType.SetPrimaryKey(
+                        primaryKey.Properties
+                            .Select(p => _relationalPropertyToCodeGenPropertyMap[p])
+                            .ToList());
+                }
+            } // end of loop over all relational EntityTypes
+
+            AddForeignKeysToCodeGenModel(relationalModel, codeGenModel);
```
natemcmaster: Maybe using the builder pattern would be better here instead of passing codeGenModel in a chain through several layers of calls.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
lajones: It's the same pattern used for Microsoft.Data.Entity.SqlServer.SqlServer*.cs, Microsoft.Data.Entity.Sqlite.Sqlite*.cs.. Plus "MetadataModelProvider" just sounds too generic to me and doesn't really indicate what it's for. So I'd prefer to keep it this way.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
lajones: I'll look into it. In RevEng it matters less because RevEng is constructing the model in the first place. But if it's easy, sure.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
```diff
@@ -0,0 +1,319 @@
...
+            [NotNull] Property relationalProperty, [NotNull] Property codeGenProperty)
+        {
+            Check.NotNull(relationalProperty, nameof(relationalProperty));
+            Check.NotNull(codeGenProperty, nameof(codeGenProperty));
+
+            codeGenProperty.IsNullable = relationalProperty.IsNullable;
+
+            if (codeGenProperty.Name != relationalProperty.Relational().ColumnName)
+            {
+                codeGenProperty.Relational().ColumnName = relationalProperty.Relational().ColumnName;
```
natemcmaster: Same as above. Not sure if we should be using .Relational() since Sqlite and Sqlserver override this.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
lajones: Hmmm. I want to attach a piece of information to a particular EntityType and pick it up again later. I thought that was a prime candidate for annotations. If I don't then I'll have to construct some map from EntityType to error message and pass that around. So I thought why not just use the existing mechanism? Remember this model is thrown away once the code is generated - so no code outside of RevEng will see or have to interpret those annotations.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
lajones: What would you suggest? Long variable names don't bother me - I'd rather be precise and have other devs understand exactly what it is meant for, rather than save them a few keystrokes typing the name, especially as we have code completion. But I'd be happy to, e.g., remove the 2 instances of "Type" from the name.

natemcmaster:
2 thoughts:

 - Testing. Is it also possible to abstract some of the testing? (Don't need to do in this PR).
 - Abstraction. Just by inspection, I speculate we will need to make some changes to our abstraction choices as we build SQLite and other relational providers. (No action item: just discussion).

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
lajones: Yeah - that's a good idea - I'll put it in as a TODO.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
lajones: I'll look into it - as above.

src/EntityFramework.Relational.Design/ReverseEngineering/EntityTypeCodeGeneratorHelper.cs
natemcmaster: :+1: Thanks for clarifying.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
natemcmaster: It's too bad `nameof` doesn't work on assemblies. We can leave this as-is. I may change later as I try to improve our .NET Native-unfriendly code.

lajones:
Yep - we should talk.
Re testing - I have a separate testing issue - let's talk about the abstraction before I carry on with that.
Re abstraction here - a SQLite provider _should_ only have to provide an override of ConstructRelationalModel(). If you find you're doing more than that then we should definitely talk.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
divega: Per https://github.com/aspnet/EntityFramework/pull/2701 `Relational()` was removed from the consumption (i.e. read-only) interfaces because if you use it you would miss the overrides. Not sure this usage has problems, because I don't know if how this method fits in the big picture.

lajones: It doesn't cause problems in this case because in this code the provider is also in charge of putting the annotations on in the first place - so the provider can decide whether they are Relational() annotations (which they currently are) or provider-specific annotations. That said I think it would be better practice to update this to do as Nate is suggesting.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringMetadataModelProvider.cs
smitpatel: `_relationalToCodeGenEntityTypeMap`?

lajones: Sounds good. Will do.

natemcmaster:
:shipit: (with the TODO's we discussed in person)

natemcmaster:
Merged with 5c309584492ef86daee9f397ca8ceab2b1f40226

Issue #2740, opened 2015-07-29
RevEng :: reverse engineering a table with column of type float is incorrect - querying the resulting DbSet throws InvalidCastException


maumar:
Problem is that we reverse engineer column of type "float" to CLR float, however when we query that table, the values of that column are of CLR type double. 

Due to the way we materialize entities (use GetValue(i) on a reader, and cast the result to appropriate type, then assign to a relevant property in the entity), we are getting invalid cast exceptions for this scenario.

Fix would be to correctly reverse engineer "float" columns into properties of CLR type "double".

We should also make sure that there are no other type that exhibit this behavior.

lajones:
See http://entityframework.codeplex.com/SourceControl/latest#src/EntityFramework.SqlServer/Resources/System/Data/SqlClient/System.Data.Resources.SqlClient.SqlProviderServices.ProviderManifest.xml. Looking at EF6 - looks like we used SQL(float) -> CLR(double) and SQL(real) -> CLR(float). 

Note also we are missing rowversion.

lajones:
Fixed by PR https://github.com/aspnet/EntityFramework/pull/2743

PR #2743, opened 2015-07-30
Fixing wrong and missing SQL Server type mappings

lajones:
Fix for #2740 - the SQL Server type float was incorrectly mapped to C# float (should be double) and SQL Server type real was incorrectly mapped to C# double (should be float).

Also we were missing a rowversion mapping - we only had its synonym, timestamp.


src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -613,9 +613,10 @@ public static string ConstructIdForCombinationOfColumns([NotNull] IEnumerable<st
             }
 
             if (tableColumn.IsStoreGenerated
+                || tableColumn.DataType == "rowversion"
```
ErikEJ: The datatype name returned fro the schema query in all current versions of SQL Server is always "timestamp" and never "rowversion"

src/EntityFramework.SqlServer.Design/Utilities/SqlServerTypeMapping.cs
```diff
@@ -54,6 +54,7 @@ public static class SqlServerTypeMapping
                 //{ "hierarchyid", typeof(TODO) },
                 //{ "sql_variant", typeof(TODO) },
                 //{ "table", typeof(TODO) },
+                { "rowversion", typeof(byte[]) },
```
ErikEJ: Same comment as above

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
divega: And, just to be sure, by "current" in this case you mean "existing"?

ErikEJ: Yes, all existing

lajones: My mistake. I was making the code consistent with what EF6 did, but it looks like (as usual) Erik is right. I'll update.

lajones:
@ErikEJ @divega - see new commit - updated as requested.

divega:
:shipit:

Is it expected that the E2E.sql is treated by GitHub (or maybe by git?) as binary?

ErikEJ:
Not if the file is saved as ansi cp 1252, but if it is saved as unicode, then yes.

lajones:
Checked in with commit fd723d3

PR #2744, opened 2015-07-30
Allow to set Property.ClrType

AndriySvyryd:
Remove ClrType and IsShadowProperty from the Property constructor
Move CLR type validation to ModelValidator
Make IModel.GetReferencingForeignKeys extension methods and rename to FindReferencingForeignKeys


ajcvickers:
:shipit:

PR #2745, opened 2015-07-30
Improve fixup perf by using index for dependent lookup

ajcvickers:
That is, an index is made for any entity on the dependent end of a relationship such that when the principal is later brought into the context it is fast to find all the dependents that must be fixed up to it.

We may want to make building this index lazy, but that requires more investigation.


AndriySvyryd:
:shipit:

Issue #2715, opened 2015-07-25
ModelSnapshot: Ensure using directives present for annotation types


Bidou44:
I just installed the beta7-13841 and ran the `Script-Migration' command. It was working with the previous release, now I receive the following error: 

PM> Script-Migration
Specified cast is not valid.

Update: same behavior with Add-Migration

tinganho:
I'm getting the error `The name 'SqlServerIdentityStrategy' does not exist in the current context`.

NoralK:
Same here on beta7-12274
```
E:\Projects\HelloMvc\Migrations\PersonDBContextModelSnapshot.cs(16,66): DNX,Version=v4.5.1 error CS0103: The name 'SqlServerIdentityStrategy' does not exist in the current context
E:\Projects\HelloMvc\Migrations\PersonDBContextModelSnapshot.cs(22,74): DNX,Version=v4.5.1 error CS0103: The name 'SqlServerIdentityStrategy' does not exist in the current context
E:\Projects\HelloMvc\Migrations\20150726005437_MyFirstMigration.cs(17,74): DNX,Version=v4.5.1 error CS0103: The name 'SqlServerIdentityStrategy' doesnot exist in the current context
E:\Projects\HelloMvc\Migrations\20150726005437_MyFirstMigration.Designer.cs(26,66): DNX,Version=v4.5.1 error CS0103: The name 'SqlServerIdentityStrategy' does not exist in the current context
E:\Projects\HelloMvc\Migrations\20150726005437_MyFirstMigration.Designer.cs(32,74): DNX,Version=v4.5.1 error CS0103: The name 'SqlServerIdentityStrategy' does not exist in the current context
```

branciat:
Also persist on dnx beta7-12275 with entityframework beta7-13844

bricelam:
To clarify, you generated migration classes using an older version of EF (e.g. Beta 6) and you're trying to run `Script-Migration` on Beta 7?

If you delete your migrations (and model snapshot) and add a new one, does Script-Migration succeed?

branciat:
I delete the migration files for each update of entity framework.
In the new migration files is missing an using on the assembly which include the SqlServerIdentityStrategy.

NoralK:
+1 same for me

NoralK:
With 7.0.0-beta7-13854 I cleared out all migrations and ran the following updated command:
```
dnx . ef migrations add MyFirstMigration --context PersonDBContext
```

[...]

Issue #2545, opened 2015-07-06
Sqlite Migration AutoIncrement wrong cast (string instead of boolean)


angelsix:
If you run ef migration add ... on an Sqlite backed model, it generates the cs file and adds:

                columns: table => new
                {
                    Id = table.Column(type: "INTEGER", nullable: false)
                        .Annotation("Sqlite:Autoincrement", true),
...

Where Annotation is the OperationBuilder<TOperation>'s Annotation function.

That functions parameter is a string, yet the underlying Operation.AddAnnotation wants an object that gets explicitly cast to a boolean in SqliteMigrationSqlGenerator at         public override void ColumnDefinition(string schema, string table, string name, string type, bool nullable, object defaultValue, string defaultExpression, IAnnotatable annotatable, IModel model, SqlBatchBuilder builder)

So this fails to cast. All the fix is needed is to add an overload or change the string parameter in OperationBuilder<TOperation>.Annotation to accept an object or boolean directly.

bricelam:
Oops, looks like this was missed while doing 43e9ae89857859801325fab849d43e179232d3d3

angelsix:
I cannot compile EF as ILogger seems to be missing from EntityFramework.Core so the sooner you can push a nightly build to https://www.myget.org/F/aspnetvnext/ that would be great and I can start using EF7 :)

PR #2753, opened 2015-07-30
Migrations: Deffer type mapping until SQL generation

### LINKED ISSUE #2421: Migrations: Defer type mapping

bricelam:
Resolves #2421

I did some manual bashing to verify this end-to-end. Migrations is pretty broken at the moment due to #2545 & #2715.

### DIFF
--- a/src/EntityFramework.Commands/Migrations/CSharpMigrationOperationGenerator.cs
+++ b/src/EntityFramework.Commands/Migrations/CSharpMigrationOperationGenerator.cs
@@ -54,7 +54,10 @@ namespace Microsoft.Data.Entity.Commands.Migrations
             Check.NotNull(operation, nameof(operation));
             Check.NotNull(builder, nameof(builder));
 
-            builder.AppendLine(".AddColumn(");
+            builder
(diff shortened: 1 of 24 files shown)


lajones:
Looks good :shipit:. It looks like you're assuming that a sequence can only be int or long. If there is no way (even by mistake) that a sequence can be set up with a type which is not int or long that's fine.

bricelam:
@lajones Correct, we [check in Metadata](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Metadata/Sequence.cs#L141-L144) that it's `byte`, `long`, `int`, or `short`.

PR #2756, opened 2015-07-30
Add support for InverseProperty

smitpatel:
part of #107 

Add support for  `InverseProperty`
Changes to NotMapped for Navigation
Now after `KeyDiscoveryConvention`, `NotMappedNavigationAttributeConvention` will run which will ignore appropriate navigations. After that `InversePropertyAttributeConvention` will run which will create relationships which are defined using `InversePropertyAttribute`. Afterwards `RelationshipDiscoveryConvention` will run which will try to build relationships based on navigations which are not ignored and not already used in relationship with `InversePropertyAttribute`.

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/InversePropertyAttributeConvention.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/InversePropertyAttributeConvention.cs
@@ -1,0 +1,78 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Data.Entity.Internal;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
+{
+    public class InversePropertyAttributeConvention : NavigationAttributeEntityTypeConvention<InversePropertyAttribute>
+    {
+        public override InternalEntityTypeBuilder Apply(InternalEntityTypeBuilder entityTypeBuilder, PropertyInfo navigationPropertyInfo, InversePropertyAttribute attribute)
+        {
+            Check.NotNull(entityTypeBuilder, nameof(entityTypeBuilder));
+            Check.NotNull(navigationPropertyInfo, nameof(navigationPropertyInfo));
+            Check.NotNull(attribute, nameof(attribute));
+
+            if (!entityTypeBuilder.CanAddNavigation(navigationPropertyInfo.Name, ConfigurationSource.DataAnnotation))
+            {
+                return entityTypeBuilder;
+            }
+
+            var targetType = navigationPropertyInfo.FindCandidateNavigationPropertyType();
+            var targetEntityTypeBuilder = entityTypeBuilder.ModelBuilder.Entity(targetType, ConfigurationSource.DataAnnotation);
+            if (targetEntityTypeBuilder == null)
+            {
+                return entityTypeBuilder;
+            }
+
+            // The navigation could have been added when the target entity type was added
+            if (!entityTypeBuilder.CanAddNavigation(navigationPropertyInfo.Name, ConfigurationSource.DataAnnotation))
+            {
+                return entityTypeBuilder;
+            }
+
+            var inverseNavigationPropertyInfo = targetType.GetRuntimeProperties().FirstOrDefault(p => string.Equals(p.Name, attribute.Property, StringComparison.OrdinalIgnoreCase));
+
+            if (inverseNavigationPropertyInfo == null
+                || inverseNavigationPropertyInfo.FindCandidateNavigationPropertyType() != entityTypeBuilder.Metadata.ClrType)
+            {
+                throw new InvalidOperationException(
+                    Strings.InvalidNavigationWithInverseProperty(navigationPropertyInfo.Name, entityTypeBuilder.Metadata.ClrType, attribute.Property, targetType));
+            }
+
+            if (inverseNavigationPropertyInfo == navigationPropertyInfo)
+            {
+                throw new InvalidOperationException(
+                    Strings.SelfReferencingNavigationWithInverseProperty(
+                        navigationPropertyInfo.Name,
+                        entityTypeBuilder.Metadata.ClrType,
+                        navigationPropertyInfo.Name,
+                        entityTypeBuilder.Metadata.ClrType));
+            }
+
+            // Check for InversePropertyAttribute on the inverseNavigation to verify that it matches.
+            var inverseAttribute = inverseNavigationPropertyInfo.GetCustomAttribute<InversePropertyAttribute>(true);
+            if (inverseAttribute != null
+                && inverseAttribute.Property != navigationPropertyInfo.Name)
+            {
+                // TODO: Log error that InversePropertyAttributes are not pointing at each other
+                var inverseNavigation = targetEntityTypeBuilder.Metadata.FindNavigation(inverseNavigationPropertyInfo.Name);
+                if (inverseNavigation != null)
+                {
+                    targetEntityTypeBuilder.RemoveRelationship(inverseNavigation.ForeignKey, ConfigurationSource.DataAnnotation);
+                }
+                return entityTypeBuilder;
+            }
+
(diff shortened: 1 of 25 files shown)


src/EntityFramework.Core/Metadata/Conventions/Internal/InversePropertyAttributeConvention.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            return entityTypeBuilder;
+        }
+
+        private static void TryBuildRelationship(
+            [NotNull] InternalEntityTypeBuilder sourceBuilder,
+            [NotNull] InternalEntityTypeBuilder targetBuilder,
+            [NotNull] PropertyInfo navigationToTarget,
+            [CanBeNull] PropertyInfo navigationToSource)
+        {
```
smitpatel: this is being duplicated in `InversePropertyAttributeConvention` and `RelationshipDiscoveryConvention`. Any good place to put it in common place?

src/EntityFramework.Core/Metadata/Conventions/Internal/NavigationAttributeConvention.cs
```diff
@@ -9,7 +10,7 @@
 
 namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
 {
-    public abstract class NavigationAttributeConvention<TAttribute> : INavigationConvention
+    public abstract class NavigationAttributeConvention<TAttribute> : INavigationConvention, IEntityTypeConvention
```
smitpatel: I have added both interface on same class. Though the attributes which are part of `INavigationConvention` will have methods of `IEntityTypeConvention` as not implemented and vice-versa. Should divide this in 2 parts or this is fine?

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -212,7 +212,7 @@ private InternalPropertyBuilder InternalProperty(Type propertyType, string prope
 
         public virtual bool CanAddNavigation([NotNull] string navigationName, ConfigurationSource configurationSource)
             => CanAdd(navigationName, isNavigation: true, configurationSource: configurationSource)
-               && Metadata.FindNavigation(navigationName) == null;
+               && (Metadata.FindNavigation(navigationName) == null || CanRemove(Metadata.FindNavigation(navigationName).ForeignKey, configurationSource, true));
```
smitpatel: I have modified `CanAddNavigation` so that navigation can be added if it can remove the foreign key of existing navigation.

src/EntityFramework.Core/Metadata/Conventions/Internal/InversePropertyAttributeConvention.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            if (!entityTypeBuilder.CanAddNavigation(navigationPropertyInfo.Name, ConfigurationSource.DataAnnotation))
+            {
+                return entityTypeBuilder;
+            }
+
+            var targetType = navigationPropertyInfo.FindCandidateNavigationPropertyType();
+            var targetEntityTypeBuilder = entityTypeBuilder.ModelBuilder.Entity(targetType, ConfigurationSource.DataAnnotation);
+            if (targetEntityTypeBuilder == null
+                || !entityTypeBuilder.CanAddNavigation(navigationPropertyInfo.Name, ConfigurationSource.DataAnnotation))
```
AndriySvyryd: Add a comment explaining why we need to make this check again

src/EntityFramework.Core/Metadata/Conventions/Internal/InversePropertyAttributeConvention.cs
AndriySvyryd: Add as another overload of InternalEntityTypeBuilder.Relationship and return the new relationship

src/EntityFramework.Core/Metadata/Conventions/Internal/NavigationAttributeConvention.cs
AndriySvyryd: Divide it, I don't see any advantage in keeping them together.

src/EntityFramework.Core/Metadata/Conventions/Internal/NotMappedNavigationAttributeConvention.cs
```diff
@@ -12,12 +12,18 @@ public class NotMappedNavigationAttributeConvention : NavigationAttributeConvent
...
+
+        public override InternalEntityTypeBuilder Apply(InternalEntityTypeBuilder entityTypeBuilder, PropertyInfo navigationPropertyInfo, NotMappedAttribute attribute)
+        {
+            Check.NotNull(entityTypeBuilder, nameof(entityTypeBuilder));
+            Check.NotNull(navigationPropertyInfo, nameof(navigationPropertyInfo));
             Check.NotNull(attribute, nameof(attribute));
 
-            var entityTypeBuilder = relationshipBuilder.ModelBuilder.Entity(navigation.DeclaringEntityType.Name, ConfigurationSource.Convention);
-            return entityTypeBuilder.Ignore(navigation.Name, ConfigurationSource.DataAnnotation) ? null : relationshipBuilder;
+            entityTypeBuilder.Ignore(navigationPropertyInfo.Name, ConfigurationSource.DataAnnotation);
```
AndriySvyryd: If Ignore is successful this should return null

AndriySvyryd:
Add another test class DataAnnotationsTestBase derived from ModelBuilderTestBase. There add tests for `NotMapped` and `InverseProperty` enabling `RelationshipDiscoveryConvention` to discover a relationship that would have been ambiguous. Add a check and a test that if the navigation referenced by  `InverseProperty` has an  `InverseProperty` pointing to a different navigation then it throws.

src/EntityFramework.Core/Metadata/Conventions/Internal/NotMappedNavigationAttributeConvention.cs
smitpatel: This is part of EntityTypeAddedConventions - returning null from here means the entityType has been removed and further convention should not run. It is not the case here. This is just ignoring property but the entityType still exists in the model.

AndriySvyryd: If this returns null then `NavigationAttributeConvention.Apply` should stop processing other attributes on this property

smitpatel:
added tests.

src/EntityFramework.Core/Metadata/Conventions/Internal/InversePropertyAttributeConvention.cs
```diff
@@ -0,0 +1,78 @@
...
+                        navigationPropertyInfo.Name,
+                        entityTypeBuilder.Metadata.ClrType));
+            }
+
+            // Check for InversePropertyAttribute on the inverseNavigation to verify that it matches.
+            var inverseAttribute = inverseNavigationPropertyInfo.GetCustomAttribute<InversePropertyAttribute>(true);
+            if (inverseAttribute != null
+                && inverseAttribute.Property != navigationPropertyInfo.Name)
+            {
+                // Log error that InversePropertyAttributes are not pointing at each other
```
AndriySvyryd: Should be a "TODO:"

test/EntityFramework.Core.Tests/Metadata/Internal/InternalEntityTypeBuilderTest.cs
```diff
@@ -1006,11 +1006,9 @@ public void Navigation_returns_same_value()
 
             Assert.True(dependentEntityBuilder.CanAddNavigation(Order.CustomerProperty.Name, ConfigurationSource.Convention));
             Assert.Same(foreignKeyBuilder, dependentEntityBuilder.Navigation(Order.CustomerProperty.Name, foreignKeyBuilder.Metadata, pointsToPrincipal: true, configurationSource: ConfigurationSource.DataAnnotation));
-            Assert.False(dependentEntityBuilder.CanAddNavigation(Order.CustomerProperty.Name, ConfigurationSource.Explicit));
 
             Assert.True(principalEntityBuilder.CanAddNavigation(Customer.OrdersProperty.Name, ConfigurationSource.Convention));
             Assert.Same(foreignKeyBuilder, principalEntityBuilder.Navigation(Customer.OrdersProperty.Name, foreignKeyBuilder.Metadata, pointsToPrincipal: false, configurationSource: ConfigurationSource.DataAnnotation));
-            Assert.False(principalEntityBuilder.CanAddNavigation(Customer.OrdersProperty.Name, ConfigurationSource.Explicit));
```
AndriySvyryd: Just change it to Assert.True?

AndriySvyryd:
:shipit:

PR #2757, opened 2015-07-30
Migrations: Fix annotation issues

### LINKED ISSUE #2545: Sqlite Migration AutoIncrement wrong cast (string instead of boolean)

### LINKED ISSUE #2715: ModelSnapshot: Ensure using directives present for annotation types

bricelam:
Fixes #2545 & fixes #2715

### DIFF
--- a/src/EntityFramework.Commands/Migrations/CSharpMigrationGenerator.cs
+++ b/src/EntityFramework.Commands/Migrations/CSharpMigrationGenerator.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Commands.Utilities;
 using Microsoft.Data.Entity.Internal;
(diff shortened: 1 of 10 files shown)


lajones:
:shipit:

PR #2758, opened 2015-07-31
Unify command building with minimal integration of feature areas.

mikary:
@bricelam please look at the migrations changes, mostly shimmed new code in via SqlBatchBuilder

@ajcvickers please look at update changes, ~~hack in `ReaderModificationCommandBatch.PopulateParameters` is a workaround for invalid parameter names (i.e. `"@@p0"`) fixing would require either adding `ITypeMapper` to the `ModificationCommandBatch` constructor dependencies (non-trivial), or taking `ITypeMapper` as a parameter to `CommandBuilder.CreateDbCommand` which seems inappropriate.~~

@anpete please look at query changes, the query integration was probably the cleanest in this PR and shows more of the general use case I would expect for this kind of command infrastructure.


src/EntityFramework.Relational/Storage/Commands/RelationalCommandBuilder.cs
```diff
@@ -0,0 +1,113 @@
...
+            Check.NotEmpty(name, nameof(name));
+
+            var parameter = _parameterList.GetOrAdd(name, value, property);
+
+            _stringBuilder.Append(parameter.Name);
+
+            return this;
+        }
+
+        public virtual RelationalCommandBuilder AppendFormat([NotNull] string sqlFragment, [NotNull] params object[] parameters)
```
bricelam: The name of this method is misleading. Perhaps something like `AppendWithParameters` would be better...

bricelam:
Changes to Migrations look good.

bricelam:
Remember to file a bug on transaction suppression.

mikary:
@AndriySvyryd since @ajcvickers is out and it's also related to your feature area, could you take a look at this?

src/EntityFramework.Relational/Update/ReaderModificationCommandBatch.cs
```diff
@@ -101,51 +100,49 @@ protected virtual void UpdateCachedCommandText(int commandPosition)
...
         {
-            var parameterName = columnModification.ParameterName;
-            var originalParameterName = columnModification.OriginalParameterName;
-
-            if (parameterName != null
-                || originalParameterName != null)
+            if (columnModification.ParameterName != null)
             {
-                var property = columnModification.Property;
+                var parameterName = columnModification.ParameterName.Substring(1);
```
AndriySvyryd: You need to do this because `RelationalCommandBuilder` uses `ParameterNameGenerator`, right?
Can't you just create a `RelationalCommandBuilder` instance in `CreateStoreCommand` that has a noop `ParameterNameGenerator`?

ErikEJ:
Maybe the providers-beware label should be added to this?

src/EntityFramework.Relational/Update/ReaderModificationCommandBatch.cs
mikary: I'm leaning toward moving the `ParameterNameGenerator` out of the `RelationalCommandBuilder` since there isn't any other provider-specific / SQL generation logic in the builder.

AndriySvyryd: With that the Update changes will look good

mikary:
PR updated based on comments / feedback

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -24,38 +25,44 @@ namespace Microsoft.Data.Entity.Query.Sql
...
 
         public virtual SelectExpression SelectExpression => _selectExpression;
 
-        public virtual string GenerateSql(IDictionary<string, object> parameterValues)
+        protected virtual IParameterNameGeneratorFactory ParameterNameGeneratorFactory
+            => _parameterNameGeneratorFactory;
+
+        protected virtual ParameterNameGenerator ParameterNameGenerator => _parameterNameGenerator;
+
+        public virtual RelationalCommand GenerateSql([NotNull] IDictionary<string, object> parameterValues)
```
maumar: This should probably be renamed

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -67,12 +74,10 @@ public virtual string GenerateSql(IDictionary<string, object> parameterValues)
                 .Create(_selectExpression.GetProjectionTypes().ToArray(), indexMap: null);
         }
 
-        public virtual IReadOnlyList<CommandParameter> Parameters => _commandParameters;
-
-        protected virtual IndentedStringBuilder Sql => _sql;
+        protected virtual RelationalCommandBuilder Sql => _sql;
```
maumar: rename

src/EntityFramework.Relational/Query/Sql/ISqlQueryGenerator.cs
```diff
@@ -5,14 +5,13 @@
 using System.Data.Common;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Storage;
+using Microsoft.Data.Entity.Storage.Commands;
 
 namespace Microsoft.Data.Entity.Query.Sql
 {
     public interface ISqlQueryGenerator
```
maumar: rename

src/EntityFramework.Relational/Query/Sql/ISqlQueryGenerator.cs
```diff
@@ -5,14 +5,13 @@
...
     {
-        string GenerateSql([NotNull] IDictionary<string, object> parameterValues);
-
-        IReadOnlyList<CommandParameter> Parameters { get; }
+        RelationalCommand GenerateSql([NotNull] IDictionary<string, object> parameterValues);
```
maumar: rename

src/EntityFramework.Relational/Storage/Commands/RelationalCommandBuilder.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public class RelationalCommandBuilder
+    {
+        private readonly IndentedStringBuilder _stringBuilder = new IndentedStringBuilder();
+        private readonly RelationalParameterList _parameterList = new RelationalParameterList();
+
+        public virtual RelationalCommandBuilder AppendLine()
+            => AppendLine(string.Empty);
+
+        public virtual RelationalCommandBuilder Append([NotNull]object o)
```
maumar: formatting, add space after [NotNul]

src/EntityFramework.Commands/Migrations/CSharpMigrationOperationGenerator.cs
```diff
@@ -1133,20 +1133,8 @@ protected virtual void Generate([NotNull] SqlOperation operation, [NotNull] Inde
...
                 builder.Append(_code.Literal(operation.Sql));
 
-                if (operation.SuppressTransaction)
-                {
-                    builder
-                        .AppendLine(",")
-                        .Append("suppressTransaction: true");
-                }
-
                 builder.Append(")");
```
bricelam: Could simplify all of this a bit further to just:
```C#
builder
    .Append(".Sql(")
    .Append(_code.Literal(operation.Sql))
    .Append(")");
```

test/EntityFramework.Sqlite.Tests/Storage/SqliteTypeMappingTest.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Data;
+using System.Data.Common;
+using Microsoft.Data.Sqlite;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public class SqliteTypeMappingTest : RelationalTypeMappingTest
```
bricelam: Thanks, these [should have been re-enabled a while ago](https://github.com/aspnet/EntityFramework/pull/2277#discussion_r31185027).

src/EntityFramework.Relational/Infrastructure/RelationalSqlExecutor.cs
```diff
@@ -1,24 +1,33 @@
...
 using JetBrains.Annotations;
-using Microsoft.Data.Entity.Query;
+using Microsoft.Data.Entity.Sql;
 using Microsoft.Data.Entity.Storage;
+using Microsoft.Data.Entity.Storage.Commands;
+using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Infrastructure
 {
     public class RelationalSqlExecutor
```
bricelam: What value does this add over `SqlStatementExecutor`?

src/EntityFramework.Relational/Storage/ISqlStatementExecutor.cs
```diff
@@ -6,43 +6,46 @@
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
-using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Storage.Commands;
 
 namespace Microsoft.Data.Entity.Storage
 {
     public interface ISqlStatementExecutor
```
bricelam: So much cleaner!

bricelam:
Still looks good from a Migrations point of view.

src/EntityFramework.Relational/Infrastructure/RelationalSqlExecutor.cs
mikary: This class provides an entry point for the `ExecuteSqlCommand` extension and it basically serves as a place to inject the `IRelationalConnection`. The `object[]` overload could move into the `SqlStatementExecutor`, but I think we need to leave the connection out since there are cases where we use different connections in migrations (you'd probably know more about that than I do).

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
```diff
@@ -191,5 +183,26 @@ public SqlStatementExecutor([NotNull] ILoggerFactory loggerFactory)
                 connection.Close();
             }
         }
+
+        protected virtual async Task<T> ExecuteAsync<T>(
+            IRelationalConnection connection,
+            Func<Task<T>> action,
```
anpete: Pass connection to Func instead of capturing in the closure.

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
```diff
@@ -147,30 +138,31 @@ public SqlStatementExecutor([NotNull] ILoggerFactory loggerFactory)
...
-            IRelationalConnection connection,
-            DbTransaction transaction,
-            string sql)
+        public virtual Task<DbDataReader> ExecuteReaderAsync(
+            [NotNull] IRelationalConnection connection,
+            [NotNull] string sql,
+            CancellationToken cancellationToken = default(CancellationToken))
         {
             Check.NotNull(connection, nameof(connection));
             Check.NotNull(sql, nameof(sql));
```
anpete: NotEmpty?

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
```diff
@@ -147,30 +138,31 @@ public SqlStatementExecutor([NotNull] ILoggerFactory loggerFactory)
...
 
-            return Execute(
+            return ExecuteAsync(
                 connection,
                 () =>
-                {
-                    var command = new SqlBatch(sql).CreateCommand(connection, transaction);
-                    Logger.LogCommand(command);
+                    {
+                        var command = new RelationalCommand(sql).CreateDbCommand(connection, _typeMapper);
```
anpete: Missing Dispose?

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
```diff
@@ -147,30 +138,31 @@ public SqlStatementExecutor([NotNull] ILoggerFactory loggerFactory)
                 connection,
                 () =>
                     {
-                        var command = new SqlBatch(sql).CreateCommand(connection, transaction);
+                        var command = new RelationalCommand(sql).CreateDbCommand(connection, _typeMapper);
```
anpete: Missing Dispose?

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
```diff
@@ -17,128 +17,119 @@ namespace Microsoft.Data.Entity.Storage
...
-                        foreach (var sqlBatch in sqlBatches)
-                        {
-                            var command = sqlBatch.CreateCommand(connection, transaction);
-                            Logger.LogCommand(command);
-
-                            command.ExecuteNonQuery();
-                        }
-                        return default(object);
-                    });
+                        var command = new RelationalCommand(sql).CreateDbCommand(connection, _typeMapper);
```
anpete: Missing Dispose?

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
anpete: Perhaps move DbCommand creation here so you can centralize and dispose in one place?

src/EntityFramework.Relational/Storage/ISqlStatementExecutor.cs
```diff
@@ -6,43 +6,46 @@
...
     {
-        Task ExecuteNonQueryAsync(
+        void ExecuteNonQuery(
```
anpete: Unify to a single version of each method. params?

src/EntityFramework.Relational/Storage/Commands/RelationalParameterList.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Microsoft.Data.Entity.Storage.Commands
+{
+    public class RelationalParameterList
+    {
+        private readonly List<RelationalParameter> _parameters = new List<RelationalParameter>();
+
+        public virtual IReadOnlyList<RelationalParameter> RelationalParameters => _parameters;
+
+        public virtual RelationalParameter GetOrAdd(
+            [NotNull] string name,
```
anpete: NotEmpty

src/EntityFramework.Relational/Query/Sql/RawSqlQueryGenerator.cs
```diff
@@ -27,47 +29,50 @@ public class RawSqlQueryGenerator : ISqlQueryGenerator
...
             Check.NotNull(sql, nameof(sql));
             Check.NotNull(parameters, nameof(parameters));
+            Check.NotNull(typeMapper, nameof(typeMapper));
 
             _selectExpression = selectExpression;
             _sql = sql;
             _inputParameters = parameters;
-            _commandParameters = new List<CommandParameter>();
             TypeMapper = typeMapper;
+            _parameterNameGeneratorFactory = new ParameterNameGeneratorFactory();
```
anpete: Is this pending DI?

mikary: Yes, should be trivial to add to the query DI change

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -24,38 +25,44 @@ namespace Microsoft.Data.Entity.Query.Sql
...
             [NotNull] SelectExpression selectExpression,
-            [CanBeNull] IRelationalTypeMapper typeMapper)
+            [NotNull] IRelationalTypeMapper typeMapper)
         {
             Check.NotNull(selectExpression, nameof(selectExpression));
+            Check.NotNull(typeMapper, nameof(typeMapper));
 
             _selectExpression = selectExpression;
             TypeMapper = typeMapper;
+            _parameterNameGeneratorFactory = new ParameterNameGeneratorFactory();
```
AndriySvyryd: Why is this not obtained from DI?

AndriySvyryd:
Update changes look good

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
mikary: Getting an item from DI in this part of query is difficult and I'm working on a large change to query that will make fixing this trivial, so this is just an intermediate solution.

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
mikary: I agree, but I think that is beyond the scope of this PR. Once we have the basic unified commands through, I'd like to do separate PRs for each of the feature areas to cleanup/refactor around the changes.

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
mikary: Are you thinking we would move the `DbCommand` and `DbParameter` creation logic out of `RelationalCommand` and `RelationalParameter` and have everything depend on the `SqlStatementExecutor`?

That sounds like it could be a good factoring, but I would suggest we save that kind of change for a later PR so that the individual feature areas can be adjusted to use the executor first as there are still some areas that are not very DI friendly.

src/EntityFramework.Relational/Infrastructure/RelationalSqlExecutor.cs
bricelam: SQL Server needs to execute `CREATE DATABASE` statements against `master`, but that's the only special case I can think of. We should look at simplifying these in another wave. There are a few components in Migrations that depend on `IRelationalConnection` just to pass it into the executor. That can probably be simplified.

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
anpete: No, was just calling out a way to DRY up the ExecuteX methods. Specifically, the call to CreateDbCommand and the missing using.

mikary: Ah, yes, that does improve the factoring quite a bit. :)

src/EntityFramework.Relational/Storage/ISqlStatementExecutor.cs
mikary: I don't think the params pattern helps here, it requires a lot of existing code to add `ToArray()` calls and makes consistency with the async version weird:
```cs
ExecuteQueryAsync(
    [NotNull] CancellationToken cancellationToken,
    [NotNull] params RelationalCommand[] relationalCommands)
```
this is also more of a provider surface than public surface, so we might want to revisit in API review but this doesn't seem like a pri0 decsision

src/EntityFramework.Relational/EntityFramework.Relational.csproj
```diff
@@ -264,15 +267,15 @@
     <Compile Include="Update\IBatchExecutor.cs" />
     <Compile Include="Update\ICommandBatchPreparer.cs" />
     <Compile Include="Update\IModificationCommandBatchFactory.cs" />
-    <Compile Include="Update\IParameterNameGeneratorFactory.cs" />
+    <Compile Include="Sql\IParameterNameGeneratorFactory.cs" />
```
bricelam: Do we need to introduce a new namespace? Should this go under `Storage` instead?

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -713,9 +713,6 @@ protected override Expression VisitChildren(ExpressionVisitor visitor)
             return this;
         }
 
-        public override string ToString() 
-            => new DefaultQuerySqlGenerator(this, null).GenerateSql(new Dictionary<string, object>());
-
```
bricelam: Is this removal intentional?

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -24,38 +25,44 @@ namespace Microsoft.Data.Entity.Query.Sql
     public class DefaultQuerySqlGenerator : ThrowingExpressionVisitor, ISqlExpressionVisitor, ISqlQueryGenerator
     {
         private readonly SelectExpression _selectExpression;
+        private readonly IParameterNameGeneratorFactory _parameterNameGeneratorFactory;
```
bricelam: No need for a field; just use a read-only auto-property.

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
```diff
@@ -17,174 +17,169 @@ namespace Microsoft.Data.Entity.Storage
...
-                cancellationToken);
+            foreach (var command in relationalCommands)
+            {
+                Execute<object>(
+                    connection,
+                    command,
+                    c => c.ExecuteNonQuery());
+            }
+
+            connection.Close();
```
bricelam: Wrap inside `finally` so reference counting doesn't get of on exceptions.

src/EntityFramework.Relational/Storage/SqlStatementExecutor.cs
```diff
@@ -17,174 +17,169 @@ namespace Microsoft.Data.Entity.Storage
...
-            {
-                connection.Close();
+                await ExecuteAsync(
+                    connection,
+                    command,
+                    async c => await c.ExecuteNonQueryAsync(cancellationToken),
+                    cancellationToken);
             }
+
+            connection.Close();
```
bricelam: `finally`

bricelam:
:shipit: After addressing [new feedback](https://github.com/aspnet/EntityFramework/pull/2758#discussion-diff-37682496), of course.

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
mikary: Yep, this is just a tweak for the `RelationalExpressionPrinter` related to some of the `IRelationalTypeMapper` changes in `DefaultQuerySqlGenerator`

Issue #2730, opened 2015-07-28
Method translators should cache reflection objects


anpete:
The following translators are not caching reflected method information:

ConvertTranslator
MathRoundTranslator
MathTruncateTranslator

maumar:
Fixed in 71fe9fd2fa1081f8b66ef219a1eef93830b9cd54

Issue #2728, opened 2015-07-28
SqlServer provider - MathRoundTranslator


bubibubi:
In EF7 SQL Server provider, MathRoundTranslator (but also other Math related translators) the source code is similar to this (I added only some comments):

```C#
public virtual Expression Translate([NotNull] MethodCallExpression methodCallExpression)
{
    // Retrieve all methods with 1 parameter or with 2 parameters where the first parameter is an int
    var methodInfos = typeof(Math).GetTypeInfo().GetDeclaredMethods("Round").Where(m =>
            m.GetParameters().Count() == 1
            || (m.GetParameters().Count() == 2 && m.GetParameters()[1].ParameterType == typeof(int)));

    // Is the methodCallExpression one of the method retrieved
    if (methodInfos.Contains(methodCallExpression.Method))
    {
        // TODO: Check with a test case. The problem here could be that the second parameter is always a 0 also if we specify 2 parameters
        var arguments = new[] { methodCallExpression.Arguments[0], Expression.Constant(0) };

        return new SqlFunctionExpression("Round", arguments, methodCallExpression.Type);
    }

    return null;
}
```

In both overloads arguments is always 
methodCallExpression.Arguments[0], Expression.Constant(0)

In case of one parameter I can understand it (ROUND in T-SQL expects 2 parameters) but in case of two parameters I expected 
methodCallExpression.Arguments[0], methodCallExpression.Arguments[1]

How can it work so implemented?

divega:
@maumar can you take a look?

maumar:
@bubibubi  is correct, this is a bug

anpete:
Related to #2730

maumar:
Fixed in 71fe9fd2fa1081f8b66ef219a1eef93830b9cd54

PR #2761, opened 2015-07-31
Fix to query issues #2728, #2730

maumar:
#2728 :: SqlServer provider - MathRoundTranslator
#2730 :: Method translators should cache reflection objects

Also added support for Modulo operator, and cleaned up code around Convert methods and added tests for all supported ConvertTo methods.


maumar:
@anpete

anpete:
:shipit:

PR #2765, opened 2015-08-01
Add environment option to migration apply command

ph1ll:
The Entity Framework migration command currently does not support applying migrations to an alternative connection string. This PR will add an "environment" option to the apply command as shown below:

```
Usage: ef migration apply [arguments] [options]

Arguments:
  [migration]  The migration to apply. Use '0' to unapply all migrations

Options:
  -c|--context <context>                The context class to use
  -s|--startupProject <startupProject>  The start-up project to use
  -e|--environment <environment>        The environment to use
  -h|--help                             Show help information
```

The idea being your model uses either "Startup*Environment*" classes or makes use of "config.*Environment*.json" files, as per http://docs.asp.net/en/latest/fundamentals/environments.html.

You can then run a command like `dnx . ef migration apply -e Staging` which will pass the environment option to `WebHostBuilder` and call your project startup with the correct `IHostingEnvironment`.

Please let me know if any chances are required for a merge, or if the team already has a different direction for this.

### DIFF
--- a/src/EntityFramework.Commands/ContextTool.cs
+++ b/src/EntityFramework.Commands/ContextTool.cs
@@ -24,12 +24,12 @@ namespace Microsoft.Data.Entity.Commands
             _services = services;
         }
 
-        public virtual DbContext CreateContext([NotNull] Type type, [CanBeNull] string startupAssemblyName)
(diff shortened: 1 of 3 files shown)


rowanmiller:
Hey,

We are planning to change how all this is implemented because the concept of environments is very ASP.NET specific. Here is the item tracking this work https://github.com/aspnet/EntityFramework/issues/2357. We will revisit environments once we have completed this redesign.

~Rowan

PR #2773, opened 2015-08-03
Add runtime directives for .NET Native

natemcmaster:
These are suboptimal directives, and this doesn't completely solve the #623, but it's a step in the right direction.

### DIFF
--- a/src/EntityFramework.Core/Properties/EntityFramework.Core.rd.xml
+++ b/src/EntityFramework.Core/Properties/EntityFramework.Core.rd.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
+  <Library>
+    <Assembly Name="EntityFramework.Core" Activate="All" />
+    <!-- TODO optimize this for types that actually need reflection -->
+  </Library>
+</Directives>
--- a/src/EntityFramework.Relational/Properties/EntityFramework.Relational.rd.xml
+++ b/src/EntityFramework.Relational/Properties/EntityFramework.Relational.rd.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
+  <Library>
+    <!-- TODO optimize this for types that actually need reflection -->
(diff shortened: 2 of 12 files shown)


src/EntityFramework.Relational/Properties/EntityFramework.Relational.rd.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
+  <Library>
+    <!-- TODO optimize this for types that actually need reflection -->
+    <Assembly Name="EntityFramework.Relational" Activate="All" />
+    <Type Name="Microsoft.Data.Entity.Migrations.Operations.MigrationOperation" Dynamic="Required All" />
+    <Type Name="Microsoft.Data.Entity.Update.UpdateSqlGenerator">
+      <Method Name="GenerateLiteral" Signature="System.Object" Dynamic="Required" />
+    </Type>
```
bricelam: We should consider removing the use of `dynamic` in this code.

bricelam:
:shipit:

src/EntityFramework.Core/Properties/EntityFramework.Core.rd.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
+  <Library>
+    <Assembly Name="EntityFramework.Core" Activate="All" />
+    <!-- TODO optimize this for types that actually need reflection -->
```
Eilon: Do we want to leave the TODO's in here or do we have bugs to track them?

Eilon: I say this in part because I believe this raw content will be shipped to customers :smile: (Granted, it'll be obscured inside a resource, but just the thought of it... :smile: )

natemcmaster: We are tracking initial .NET Native support with #623 . When it actually works, then we can then we can move on to optimizing. Till then, let's leave this as a TODO.

Eilon: Got it, thanks!

PR #2774, opened 2015-08-03
Perf test improvements

rowanmiller:
* Tracking and NoTracking variations for query tests - also moved query
creation to be outside timing so that only execution is timed
* Altered variation name on some tests so that EF6/EF7 scenarios align
(makes it easier to compare figured in reports, pivot table, etc)
* Swapped to SQL Client for result saving :-1:  - we reverted the name
changes that allowed EF7 and EF6 side-by-side so the EF6 results were
failing to save (because we were using EF7 to save them)


anpete:
:shipit:

Issue #2721, opened 2015-07-27
SQLite: Reverse Engineer


bricelam:
Implement all the things required to do
```ps1
Scaffold-DbContextTemplate EntityFramework.Sqlite
Scaffold-DbContext 'Data Source=Chinook.sqlite' EntityFramework.Sqlite
```

lajones:
@natemcmaster - I would wait til https://github.com/aspnet/EntityFramework/pull/2741 is in before you start.

PR #2780, opened 2015-08-05
SQLite reveng

natemcmaster:
Implements ReverseEngineeringMetadataModelProvider for SQLite.

#2721 

TODO - E2E testing for SQlite


natemcmaster:
cc: @lajones

src/EntityFramework.Sqlite.Design/EntityFramework.Sqlite.Design.csproj
```diff
@@ -0,0 +1,112 @@
...
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.Framework.DependencyInjection.Abstractions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null" />
+    <Reference Include="Microsoft.Framework.Logging.Abstractions">
+      <HintPath>..\..\..\..\.dnx\packages\Microsoft.Framework.Logging.Abstractions\1.0.0-beta7-11626\lib\net45\Microsoft.Framework.Logging.Abstractions.dll</HintPath>
```
lajones: Having a hard-coded path here doesn't seem right. Can we see if we can find a way around this?

src/EntityFramework.Sqlite.Design/Properties/Strings.resx
```diff
@@ -0,0 +1,126 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="ForeignKeyScaffoldError" xml:space="preserve">
+    <value>Could not scaffold the foreign key {depTable}({depProperties}) =&gt;{principalTable}({principalProperties}). One of the tables or properties in this foreign key could not be found.</value>
```
lajones: Say which table / property was the problem.

src/EntityFramework.Sqlite.Design/Properties/Strings.resx
```diff
@@ -0,0 +1,126 @@
...
+  </data>
+  <data name="UnrecognizedType" xml:space="preserve">
+    <value>The property '{propertyName}' on table '{tableName}' could not be scaffolded because the column type was either not defined or not recognized.</value>
```
lajones: Say what the bad column type was.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDbContextCodeGeneratorHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Microsoft.Data.Entity.Sqlite.Design.ReverseEngineering
+{
+    public class SqliteDbContextCodeGeneratorHelper : DbContextCodeGeneratorHelper
+    {
+        public SqliteDbContextCodeGeneratorHelper([NotNull] DbContextGeneratorModel generatorModel)
+            : base(Check.NotNull(generatorModel, nameof(generatorModel)))
+        {
+        }
+
+        protected override IRelationalMetadataExtensionProvider RelationalExtensions { get; }
```
lajones: Point to the SqliteMetadataExtensionProvider already in EntityFramework.Sqlite.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// </summary>
+        /// <param name="modelBuilder"></param>
+        /// <param name="tables"></param>
+        public static void ParseTableDefinition(ModelBuilder modelBuilder, string tableName, string sql)
+        {
+            var entityBuilder = modelBuilder.Entity(tableName);
+            var i = 0;
+            var statements = ParseStatements(sql).ToList();
+            for (i = 0; i < statements.Count; i++)
+            {
```
lajones: Nit: Since you already declare it shouldn't need i=0

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+using Microsoft.Data.Sqlite;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.Data.Entity.Sqlite.Design.ReverseEngineering
+{
+    public class SqliteMetadataModelProvider : ReverseEngineeringMetadataModelProvider
+    {
+        private readonly SqliteReverseTypeMapper _typeMapper;
+        private readonly ILogger _logger;
+
```
lajones: Shouldn't need _logger. `ReverseEngineeringMetadataModelProvider` has a public Logger property which is set by the constructor.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+                master.CommandText = "SELECT type, name, sql, tbl_name FROM sqlite_master";
+                using (var reader = master.ExecuteReader())
+                {
+                    while (reader.Read())
+                    {
+                        var type = reader.GetFieldValue<string>(0);
+                        var name = reader.GetFieldValue<string>(1);
+                        var sql = reader.GetValue(2) as string; // TODO https://github.com/aspnet/Microsoft.Data.Sqlite/issues/111
+                        var tableName = reader.GetFieldValue<string>(3);
+
```
lajones: What happens if any of the above are null? Does the sqlite_master table allow that?

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+                        if (type == "table"
+                            && name != "sqlite_sequence")
+                        {
```
lajones: Presume this is because sqlite_sequence is a system table. Are there any other sqlite_* tables you should avoid?

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
```diff
@@ -0,0 +1,145 @@
...
+                    return true;
+                });
+            if (index != null)
+            {
+                index.IsUnique = true;
+            }
+        }
+
+        private static IEnumerable<string> SafeSplit(string str, char separator)
+        {
```
lajones: Nit: perhaps add a comment saying what this is supposed to do and /or rename. From its name I didn't expect it to do everything it does. Maybe make non-static and add a few tests as well?

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
```diff
@@ -0,0 +1,145 @@
...
+                        else if (c == ')')
+                        {
+                            nestedLevel--;
+                        }
+                    }
+                }
+                else if (str[idx] == '"')
+                {
+                    while (str[++idx] != '"'
+                           && idx < str.Length)
```
lajones: Can strings contain escaped double-quotes? As in C# "\""? If so don't think the algorithm works. Also if strings can contain '(' or ')' not sure that won't interfere with the above else if?

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+            var modelBuilder = new ModelBuilder(new ConventionSet());
+
+            using (var connection = new SqliteConnection(connectionString))
+            {
+                connection.Open();
+                var tables = new Dictionary<string, string>();
+                var indexes = new List<SqliteIndex>();
+                var master = connection.CreateCommand();
+                master.CommandText = "SELECT type, name, sql, tbl_name FROM sqlite_master";
+                using (var reader = master.ExecuteReader())
```
lajones: Here you

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+                {
+                    while (reader.Read())
+                    {
+                        var type = reader.GetFieldValue<string>(0);
+                        var name = reader.GetFieldValue<string>(1);
```
lajones: Nit: Here you hard-code the 0, 1, 2... offsets. Elsewhere you use enums. I'd be consistent.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+            Cid,
+            Name,
+            Type,
+            NotNull,
+            DefaultValue,
+            Pk
+        }
+
+        private void GetTableInfo(SqliteConnection connection, ModelBuilder modelBuilder, ICollection<string> tables)
+        {
```
lajones: Nit: I'd expect a GetXXX method to return XXX's. Maybe a better name would be BuildTablesAndPrimaryKeys()?

src/EntityFramework.Sqlite.Design/EntityFramework.Sqlite.Design.csproj
natemcmaster: I'll clean that out. Resharper put it in automagically.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
natemcmaster: They are guaranteed to exist. Only the SQL field can be empty.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
natemcmaster: sqlite_master, but it will never show up in this result set.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
natemcmaster: In SQLite, escaped double quotes use two sequential double quotes : `CREATE TABLE "Name "" Has Quote" ()`. When within a string literal, matching parenthesis doesn't matter.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
natemcmaster: Here the select statement allows me to specify order of columns. The return of pragma statements do not and could potentially change with SQLite version.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
natemcmaster: Good point. I'll update to reflect better what is occurring.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                "CONSTRAINT",
+                "PRIMARY",
+                "UNIQUE",
+                "CHECK",
+                "FOREIGN"
+            };
+
+        /// <summary>
+        ///     Performs reveng on the CREATE dml to compenstate for limitations in SQLite's pragma statements
```
bricelam: We probably don't want this on MSDN...

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// </summary>
+        /// <param name="modelBuilder"></param>
+        /// <param name="tables"></param>
```
bricelam: Write descriptions or remove these.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+            Id,
+            Seq,
+            Table,
+            From,
+            To,
+            OnUpdate,
+            OnDelete,
+            Match
+        }
```
bricelam: If these can change, it may be better to get them by name, not ordinal.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+        private void GetForeignKeysList(SqliteConnection connection, ModelBuilder modelBuilder, ICollection<string> tables)
+        {
+            //TODO case insensitive match on metadata
+            foreach (var tableName in tables)
+            {
+                var fkList = connection.CreateCommand();
+                fkList.CommandText = $"PRAGMA foreign_key_list(\"{tableName.Replace("\"", "\"\"")}\");";
+                // TODO https://github.com/aspnet/Microsoft.Data.Sqlite/issues/112
+
+                var keys = new Dictionary<int, ForeignKey>();
```
bricelam: :trollface: Foreign key constraints aren't keys

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -0,0 +1,309 @@
...
+                using (var reader = fkList.ExecuteReader())
+                {
+                    while (reader.Read())
+                    {
+                        var id = reader.GetFieldValue<int>((int)ForeignKeyList.Id);
```
bricelam: You could avoid boxing by using `.GetInt32()` instead. ...since this is *so* perf-critical. :wink:

src/EntityFramework.Sqlite.Design/SqliteReverseTypeMapper.cs
```diff
@@ -0,0 +1,84 @@
...
+            foreach (var rules in _typeRules)
+            {
+                matchedAffinity = rules(typeName);
+                if (matchedAffinity != null)
+                {
+                    return matchedAffinity;
+                }
+            }
+
+            return null;
```
bricelam: I wonder if we should default to something like `byte[]` or `string` since SQLite will just coerce values anyway.

src/EntityFramework.Sqlite.Design/SqliteReverseTypeMapper.cs
```diff
@@ -0,0 +1,84 @@
...
+                { "blob", typeof(byte[]) },
+                { "real", typeof(double) },
+                { "numeric", typeof(double) },
+                { "datetime", typeof(DateTime) },
+                { "date", typeof(DateTime) },
+                { "timestamp", typeof(int) },
+                { "bigint", typeof(long) },
+                { "unsigned big int", typeof(long) },
+                { "guid", typeof(Guid) }
+            };
```
bricelam: I don't think we should maintain a list like this. Having the affinity rules and a good default type (like `string`) is probably good enough. The user can always hand-edit any types we guessed wrong.

bricelam: I also think it would be fine to only reverse engineer `long`, `double`, `string`, and `byte[]` columns. It's the purist implementation.

src/EntityFramework.Sqlite.Design/project.json
```diff
@@ -0,0 +1,24 @@
+﻿{
+  "version": "7.0.0-*",
+  "description": "Design-time Entity Framework functionality for  SQLite",
```
bricelam: double space: `for  SQLite`

bricelam:
Let's discuss the type mapping with a few more people. Otherwise, it looks good to me.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
divega: FWIW `GetFieldValue<int>()` shouldn't need to box, unless the implementation does it.

divega:
> Let's discuss the type mapping with a few more people

Can we chat in tomorrow's design meeting?

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
bricelam: There was some weird IL limitation about having to box before returning a value type in place of an unconstrained generic type. See [the implementation](https://github.com/aspnet/Microsoft.Data.Sqlite/blob/dev/src/Microsoft.Data.Sqlite/SqliteDataReader.cs#L395). I'd love to know how to get around it!

anpete: Perhaps this issue? http://stackoverflow.com/questions/2031010/cast-value-type-to-generic

anpete: One slight optimization to GetFieldValue<> would be to move more commonly used types "up" so that less comparisons are needed on average.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
lajones: Am I missing something? I'd have thought that example would prove the point? Above wouldn't you start trying to split again after Name" instead of after Has Quote"? Similarly what happens if your string looks like this: ( xxx ")" ) - wouldn't you want the paren-matching to ignore the end-paren within the double quotes?

lajones:
I would discuss the type mapping with Brice and, as discussed, you'll need to rebase as the underlying change has had changes. But overall looks good.

natemcmaster:
@lajones where did you push those underlying changes? aspnet:150724_WorkOnBeta7_01 doesn't look like it has changed.

src/EntityFramework.Sqlite.Design/SqliteReverseTypeMapper.cs
ErikEJ: If you look at the declared type, maybe at least you should include all the ones here: https://www.sqlite.org/datatype3.html - at least the ones resulting in a affinity that is not "NUMERIC"

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
natemcmaster: Yup, verified. Added some tests and fixed.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
natemcmaster: GetOrdinal doesn't work correctly with the results of pragma statements.

natemcmaster:
Updated with response to feedback.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
```diff
@@ -0,0 +1,176 @@
...
+            else
+            {
+                return identifier;
+            }
+
+            return identifier.Substring(1, identifier.Length - 2).Replace($"{quote}{quote}", quote.ToString());
+        }
+
+        // For internal use. Only designed to split a valid SQLite statements.
+        internal static IEnumerable<string> SafeSplit(string str, char separator)
```
bricelam: Internal for testing? We've tended to move away from that pattern...

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
bricelam: :smile: I wondered if they even had names

src/EntityFramework.Sqlite.Design/SqliteReverseTypeMapper.cs
```diff
@@ -0,0 +1,52 @@
...
+        private readonly Func<string, Type>[] _typeRules =
+            {
+                name => Contains(name, "INT") ? typeof(long) : null,
+                name => Contains(name, "CHAR") || Contains(name, "CLOB") || Contains(name, "TEXT") ? typeof(string) : null,
+                name => Contains(name, "BLOB") ? typeof(byte[]) : null,
+                name => Contains(name, "REAL") || Contains(name, "FLOA") || Contains(name, "DOUB") ? typeof(double) : null,
+            };
+
+        private static bool Contains(string haystack, string needle) 
+            => haystack.IndexOf(needle, StringComparison.OrdinalIgnoreCase) >= 0;
```
bricelam: Isn't there an overload of the static `string.Contains()` method that takes a `StringComparison` value?

bricelam:
:shipit: (from a SQLite perspective)

lajones:
@natemcmaster underlying changes checked in with commit 5c30958

src/EntityFramework.Sqlite.Design/SqliteReverseTypeMapper.cs
natemcmaster: Wish there was. I looked.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDmlParser.cs
natemcmaster: ...to what? protected? or public on methods, internal on class?

bricelam: Probably a public class in an `*.Internal` namespace with public methods

bricelam: But it's probably fine to leave it as is for now

natemcmaster:
Squashed and ready to merge to dev. @lajones any final sign off?

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteContextCodeGeneratorHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Data.Entity.Sqlite.Design.ReverseEngineering
+{
+    public class SqliteContextCodeGeneratorHelper : DbContextCodeGeneratorHelper
+    {
+        public SqliteContextCodeGeneratorHelper(
+            [NotNull] DbContextGeneratorModel model, 
+            [NotNull] IRelationalMetadataExtensionProvider extensionsProvider)
+            : base(model, extensionsProvider)
+        {
```
lajones: Did you not want to override ClassName? The default produces a fixed name. The SQL Server one (by contrast) interprets the connection string as a SQL Server connection string and tries to use the Initial Catalog.

natemcmaster: I'm fine leaving it as "ModelContext" by default. It's easier than trying to sanitize a file path into a context name.

lajones:
:shipit: Looks good.

PR #2787, opened 2015-08-06
:bar_chart: Reporting for benchmark results

rowanmiller:
This is a pragmatic start at some reporting for benchmark results. There
is certainly lots of room for improvement (adopting a plugin to render
pivot tables, ajax loading of data for charts, etc.) but we can evolve
as we work out what information is useful to us.

The site will be hosted hooked up to our performance CI results, but you
can also just run it locally and it will look at the same local database
used to store local benchmark runs (which should be useful when working
on local perf improvements).


bricelam:
:shipit: (I didn't actually review any code, just how it was added to the repo)

.gitignore
```diff
@@ -26,4 +26,6 @@ nuget.exe
 *.sln.ide
 *.lock.json
 *.db
-.vs/
\ No newline at end of file
+.vs/
+lib/
+node_modules/
```
bricelam: :frowning: Looks like there's no way in KoreBuild to exclude things from [the targets](https://github.com/aspnet/Universe/blob/dev/build/_k-standard-goals.shade#L39-L52) that run npm, bower & grunt.

rowanmiller: Discussed and agreed we should move this to a separate repo. I think once we RTM we may want to split all the benchmark stuff out to a separate repro anyway since we'll be testing various versions and not just the current dev branch.

rowanmiller:
Moving to another repo since it will be messy to exclude this from build/restore etc.

PR #2792, opened 2015-08-07
:art: Defaults for benchmark Iterations

rowanmiller:
Standardizing on a 100 iterations and 1 warmup iteration for benchmarks.
Moving them to be default values and only overriding where text
execution time make 100 iterations prohibitive.


anpete:
:shipit:

PR #2793, opened 2015-08-07
Fix #290 Query DI Part 1 - CompiledQueryCache and Database classes

### LINKED ISSUE #290: Query: DI enable
...when architecture stabilizes

mikary:
Move compilation code out of cache to allow for scoped injection
Decouple Database classes from query compilation, eliminates need for provider-
specific relational Database classes

@maumar @anpete


src/EntityFramework.Core/EntityFramework.Core.csproj
```diff
@@ -217,6 +217,14 @@
     <Compile Include="Query\Annotations\QueryAnnotation.cs" />
     <Compile Include="Query\Annotations\IncludeQueryAnnotation.cs" />
     <Compile Include="EF.cs" />
+    <Compile Include="Query\Compiler\CompiledQuery.cs" />
```
anpete: Please don't introduce new namespaces - we have already done API review here.

src/EntityFramework.Core/EntityFramework.Core.csproj
```diff
@@ -224,17 +232,24 @@
     <Compile Include="Query\Expressions\PropertyEvaluationPreventingExpression.cs" />
     <Compile Include="Query\ExpressionVisitors\ExpressionStringBuilder.cs" />
     <Compile Include="Query\ExpressionVisitors\ExpressionVisitorBase.cs" />
-    <Compile Include="Query\ExpressionVisitors\FunctionEvaluationDisablingVisitor.cs" />
+    <Compile Include="Query\ExpressionVisitors\NullEvaluatableExpressionFilter.cs" />
+    <Compile Include="Query\IQueryCompilationContextFactory.cs" />
+    <Compile Include="Query\Preprocessor\Internal\FunctionEvaluationDisablingVisitor.cs" />
```
anpete: New namespace.

anpete: I don't think we have to pull out all nested classes. Some of them are implementation detail.

src/EntityFramework.Core/EntityFrameworkServiceCollectionExtensions.cs
```diff
@@ -104,19 +104,34 @@ public static class EntityFrameworkServiceCollectionExtensions
...
+        private static IServiceCollection AddQuery(this IServiceCollection serviceCollection)
+        {
+            return serviceCollection
+                .AddSingleton<IMemoryCache, MemoryCache>()
+                .AddSingleton<ICompiledQueryCache, CompiledQueryCache>()
+                .AddScoped<IEntityQueryProvider, EntityQueryProvider>()
+                .AddScoped<IQueryExecutor, QueryExecutor>()
+                .AddScoped<IQueryPreprocessor, QueryPreprocessor>()
+                .AddScoped<QueryCompiler>()
+                .AddScoped<CompiledQueryCacheKeyGenerator>()
```
anpete: Shouldn't this be registered via interface? Same with QueryCompiler.

src/EntityFramework.Core/Query/Compiler/CompiledQuery.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Data.Entity.Query.Compiler
+{
+    public class CompiledQuery
```
anpete: This class is an implementation detail of the cache.

src/EntityFramework.Sqlite/Query/SqliteQueryCompilationContextFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+using Microsoft.Data.Entity.ChangeTracking.Internal;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Query.Methods;
+using Microsoft.Data.Entity.Storage;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.Data.Entity.Query
+{
+    public class SqliteQueryCompilationContextFactory : IQueryCompilationContextFactory
```
anpete: Isn't the point of this change to make this go away?

src/EntityFramework.Core/EntityFramework.Core.csproj
mikary: I would rather put them in M.D.E.Query.Internal, they are implementation details but the purpose of the precompiler is to run through a series of transformations and the details shouldn't matter (Trying for more of an open/closed principal approach)

src/EntityFramework.Core/EntityFrameworkServiceCollectionExtensions.cs
mikary: It's registered as a provider service to allow custom cache key generation and this implementation is simply the default. It's the same with QueryCompiler, though at this point only the InMemory provider needs to override.

src/EntityFramework.Sqlite/Query/SqliteQueryCompilationContextFactory.cs
mikary: It probably will go away in the long run, but this PR is intended to be an incremental change. We can hold off on merging if it would be better to have the full implementation but I didn't want too overwhelming of a PR

src/EntityFramework.Core/EntityFramework.Core.csproj
anpete: That's fine when they are not purely implementation detail.

src/EntityFramework.Core/EntityFrameworkServiceCollectionExtensions.cs
anpete: I'm referring to lines 127, 128

mikary: There are exceptions required for the `DatabaseProviderServices`, for example the `ValueGeneratorSelector` is registered this way (though in that case both InMemory and Relational have implementations registered so it might not be necessary).

I originally had the Core compiler as an abstract class with InMemory and Relational implementations, but the logic for relational just duplicated what happens in core.

anpete: OK, I guess I must be missing something about how this works. I guess another question is: Is it required to be this way in order to work?

mikary: Anything else would be going against the pattern for `DatabaseProviderServices` or result in duplicated code (registering as a class in the AddProvider methods or being completely abstract and using the same logic in InMemory and Relational)

src/EntityFramework.Core/Query/Compiler/CompiledQueryCacheKeyGenerator.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq.Expressions;
+using JetBrains.Annotations;
```
maumar: nit - move JetBrains using to the end

src/EntityFramework.Core/Query/QueryCompiler.cs
```diff
@@ -0,0 +1,192 @@
...
+
+                    return _database.CompileAsyncQuery<TResult>(queryModel);
+                });
+        }
+
+        private Func<QueryContext, TResult> MapQueryExecutor<TResult>(QueryModel queryModel, Type resultItemType)
+        {
+            if (resultItemType == typeof(TResult))
+            {
+                var func = _database.CompileQuery<TResult>(queryModel);
```
maumar: compiledQuery rather than func?

src/EntityFramework.Core/Query/QueryPreprocessor.cs
```diff
@@ -0,0 +1,236 @@
...
+            {
+                var clonedModel = expression.QueryModel.Clone();
+
+                clonedModel.TransformExpressions(Visit);
+
+                return new SubQueryExpression(clonedModel);
+            }
+        }
+
+        public class NullEvaluatableExpressionFilter : EvaluatableExpressionFilterBase
```
maumar: private?

src/EntityFramework.Core/Query/QueryPreprocessor.cs
```diff
@@ -0,0 +1,236 @@
...
+                return base.VisitMethodCall(methodCallExpression);
+            }
+        }
+
+        private class FunctionEvaluationDisablingVisitor : ExpressionVisitorBase
+        {
+            public static readonly MethodInfo DbContextSetMethodInfo
+                = typeof(DbContext).GetTypeInfo().GetDeclaredMethod("Set");
+
+            private static readonly MethodInfo[] _nonDeterministicMethodInfos =
```
maumar: fix formatting

src/EntityFramework.Core/Query/QueryPreprocessor.cs
```diff
@@ -0,0 +1,38 @@
...
+            var partialEvaluationInfo = EvaluatableTreeFindingExpressionVisitor
+                .Analyze(functionEvaluationDisabledExpression, new NullEvaluatableExpressionFilter());
+
+            var parameterExtractedExpression = new ParameterExtractingExpressionVisitor(partialEvaluationInfo, queryContext)
+                .Visit(functionEvaluationDisabledExpression);
+
+            return parameterExtractedExpression;
+        }
+
+        private class NullEvaluatableExpressionFilter : EvaluatableExpressionFilterBase
```
anpete: Don't duplicate this..

src/EntityFramework.Core/Query/QueryPreprocessor.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Query.ExpressionVisitors;
+using Microsoft.Data.Entity.Utilities;
+using Remotion.Linq.Parsing.ExpressionVisitors.TreeEvaluation;
+
+namespace Microsoft.Data.Entity.Query
+{
+    public class QueryPreprocessor : IQueryPreprocessor
```
anpete: This new abstraction seems unrelated to the DI change.

mikary:
Updated to remove unrelated abstractions

anpete:
:shipit:

PR #2797, opened 2015-08-07
Add support for ForeignKeyAttribute

smitpatel:
part of #107

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
@@ -1,0 +1,190 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
+{
+    public class ForeignKeyAttributeConvention : IForeignKeyConvention
+    {
+        public virtual InternalRelationshipBuilder Apply(InternalRelationshipBuilder relationshipBuilder)
+        {
+            Check.NotNull(relationshipBuilder, nameof(relationshipBuilder));
+
+            var foreignKey = relationshipBuilder.Metadata;
+
+            var fkPropertyOnPrincipal = FindCandidateDependentPropertyThroughEntityType(foreignKey.PrincipalEntityType, foreignKey.PrincipalToDependent?.Name);
+            var fkPropertyOnDependent = FindCandidateDependentPropertyThroughEntityType(foreignKey.DeclaringEntityType, foreignKey.DependentToPrincipal?.Name);
+
+            if (!string.IsNullOrEmpty(fkPropertyOnDependent)
+                && !string.IsNullOrEmpty(fkPropertyOnPrincipal))
+            {
+                // TODO: Log Error that unable to determine principal end based on foreign key attributes
+                var principalTypeNavigationName = foreignKey.PrincipalToDependent?.Name;
+                var dependentTypeNavigationName = foreignKey.DependentToPrincipal?.Name;
+
+                var dependentEntityTypebuilder = relationshipBuilder.ModelBuilder.Entity(foreignKey.DeclaringEntityType.Name, ConfigurationSource.Convention);
+                var removedConfigurationSource = dependentEntityTypebuilder.RemoveRelationship(foreignKey, ConfigurationSource.DataAnnotation);
+
+                if (removedConfigurationSource == null)
+                {
+                    return relationshipBuilder;
+                }
+
+                var principalEntityTypeBuilder = relationshipBuilder.ModelBuilder.Entity(foreignKey.PrincipalEntityType.Name, ConfigurationSource.Convention);
+
+                dependentEntityTypebuilder.Relationship(
+                    principalEntityTypeBuilder,
+                    dependentEntityTypebuilder,
+                    navigationToPrincipalName: dependentTypeNavigationName,
+                    navigationToDependentName: null,
+                    configurationSource: ConfigurationSource.DataAnnotation);
+
+                principalEntityTypeBuilder.Relationship(
+                    dependentEntityTypebuilder,
+                    principalEntityTypeBuilder,
+                    navigationToPrincipalName: principalTypeNavigationName,
+                    navigationToDependentName: null,
+                    configurationSource: ConfigurationSource.DataAnnotation);
+
+                return null;
+            }
+
+            var fkPropertiesOnPrincipalToDependent = FindCandidateDependentPropertiesThroughNavigation(relationshipBuilder, pointsToPrincipal: false);
+            var fkPropertiesOnDependentToPrincipal = FindCandidateDependentPropertiesThroughNavigation(relationshipBuilder, pointsToPrincipal: true);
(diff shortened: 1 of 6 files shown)


test/EntityFramework.Core.Tests/ModelBuilderTest/TestModel.cs
```diff
@@ -198,5 +198,81 @@ private class BookLabel
...
+
+            [ForeignKey("Author")]
+            public int AuthorFkProp { get; set; }
+
+            public int AuthorFkNav { get; set; }
+
+            [ForeignKey("AuthorFkNav")]
+            public Author Author { get; set; }
+
+            [ForeignKey("PostDetailsFk,")]
```
natemcmaster: Is there supposed to be a dangling comma?

smitpatel: Yup. The value is supposed to be comma separated list of foreign key properties. If the incorrect list is provided then the attribute will be ignored. There is a test verifying this scenario which uses this navigation.

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -0,0 +1,187 @@
...
+                && fkPropertiesOnPrincipalToDependent != null
+                && !fkPropertiesOnDependentToPrincipal.SequenceEqual(fkPropertiesOnPrincipalToDependent))
+            {
+                // TODO: Log error that mismatch in foreignKey Attribute on both navigations
+                return relationshipBuilder;
+            }
+
+            var fkPropertiesOnNavigation = fkPropertiesOnDependentToPrincipal ?? fkPropertiesOnPrincipalToDependent;
+
+            if (fkPropertiesOnNavigation?.Count == 0)
```
natemcmaster: Could you squish this check with the if check on line 70?

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            var navigationFkAttribute = navigation != null
+                ? GetForeignKeyAttribute(navigation.DeclaringEntityType, navigation.Name)
+                : null;
+
+            if (navigationFkAttribute != null)
+            {
+                var properties = navigationFkAttribute.Name.Split(',').Select(p => p.Trim()).ToList();
+
+                if (properties.Any(string.IsNullOrWhiteSpace))
```
natemcmaster: If we skipped the check here, will this get caught later when looking up properties?

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -0,0 +1,187 @@
...
+                var removedConfigurationSource = dependentEntityTypebuilder.RemoveRelationship(foreignKey, ConfigurationSource.DataAnnotation);
+
+                if (removedConfigurationSource == null)
+                {
+                    return relationshipBuilder;
+                }
+
+                var principalEntityTypeBuilder = relationshipBuilder.ModelBuilder.Entity(foreignKey.PrincipalEntityType.Name, ConfigurationSource.Convention);
+
+                dependentEntityTypebuilder.Relationship(principalEntityTypeBuilder, dependentEntityTypebuilder, dependentTypeNavigationName, null, ConfigurationSource.DataAnnotation);
```
natemcmaster: Nit: (passing along Brice's styling suggestions) when passing null literal, it's more readable to include the variable name. [Example](https://github.com/aspnet/EntityFramework/commit/4a11781499ea95165cf4bc11c55b1b5778064c0d#diff-4ae577e6726dfbb2a31b20bd07e1af96R53)

natemcmaster:
:+1: for lots of testing.

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
smitpatel: will do. :smile: The conditions were pretty complex before that I had to keep this separate.

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
smitpatel: If we skip check here, it means the empty property will be passed in the list. When the list is used to look into existing properties the `FindProperty()` on line 90 will throw due to incorrect parameter.

natemcmaster: Right, won't this work too for letting the user know they have a bad param passed into ForeignKeyAttribute? Seems like we could skip the check here and let it percolate.

smitpatel: At present we are logging error and ignoring the attribute. If user is using a logger which throws on error then only code will throw otherwise the code will ignore and move ahead.
This triggers interesting discussion whether we should throw in case of data annotations (or conventions). We definitely do not want to throw when things are just ambiguous. In worst case user will not get a model as expected. Which is fine if the user using annotations in wrong way.
In this case we can throw and tell user that there is typo in the name. But it is not possible to know the user intent in all cases.

natemcmaster: IMO it is okay to ignore attributes that are irrelevant. But in this specific case, it would be better for users if our code threw errors because the attribute parameter specified is invalid.

smitpatel:
@AndriySvyryd - Can you review this?

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -0,0 +1,192 @@
...
+                    dependentEntityTypebuilder,
+                    principalEntityTypeBuilder,
+                    navigationToPrincipalName: principalTypeNavigationName,
+                    navigationToDependentName: null,
+                    configurationSource: ConfigurationSource.DataAnnotation);
+
+                return null;
+            }
+
+            var fkPropertyOnEntityType = fkPropertyOnDependent ?? fkPropertyOnPrincipal;
```
AndriySvyryd: This variable doesn't save much repetition and the code would be easier to read without it

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -0,0 +1,192 @@
...
+            return newRelationshipBuilder ?? relationshipBuilder;
+        }
+
+        private ForeignKeyAttribute GetForeignKeyAttribute(EntityType entityType, string propertyName)
+        {
+            return entityType.ClrType?.GetRuntimeProperties().
+                FirstOrDefault(p => string.Equals(p.Name, propertyName, StringComparison.OrdinalIgnoreCase))?.GetCustomAttribute<ForeignKeyAttribute>(true);
+        }
+
+        private string FindCandidateDependentPropertyThroughEntityType(EntityType entityType, string navigationName)
```
AndriySvyryd: This should return a collection so we could detect when there are several properties with annotation on one side and one on the other. In this case we should remove the relationship (add a test case)

AndriySvyryd:
:shipit:

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
smitpatel: filed #2829 to drive discussion around this and make a decision.

PR #2799, opened 2015-08-07
Store alternate keys without relationship in model snapshot

### LINKED ISSUE #2788: EF7-beta6 : "Superkey" produces a migration issue
Hello,

I'm having an issue with superkeys (Primary Keys / Alternate Keys)

http://pastebin.com/zsdX6wmV

The code above is correctly detected by migration change but when he create "DbContextModelSnapshot", the snapshot is wrong and cause outdated migration

smitpatel:
Resolves #2788 

In model snapshot for any key
1. If its a primary key then it will be stored using `Key()`
2. If its a principal key for relationship then it will be stored using `PrincipalKey()`
3. All other keys are stored using `AlternateKey()`


bricelam:
:shipit:

PR #2800, opened 2015-08-08
Add Code First API to configure the discriminator column

AndriySvyryd:
Add API for setting annotations on internal builders that returns success status


test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
```diff
@@ -13,35 +11,14 @@ public override void OnModelCreating(ModelBuilder modelBuilder)
...
-            plant.Relational().DiscriminatorProperty = plantDiscriminatorProperty;
-
-            var rose = modelBuilder.Entity<Rose>().Metadata;
-            rose.Relational().DiscriminatorValue = PlantGenus.Rose;
-
-            var daisy = modelBuilder.Entity<Daisy>().Metadata;
-            daisy.Relational().DiscriminatorValue = PlantGenus.Daisy;
+            modelBuilder.Entity<Plant>().Discriminator(p => p.Genus)
+                .HasValue(typeof(Rose), PlantGenus.Rose)
+                .HasValue(typeof(Daisy), PlantGenus.Daisy);
```
AndriySvyryd: @rowanmiller What do you think about this API?

test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
```diff
@@ -13,35 +11,14 @@ public override void OnModelCreating(ModelBuilder modelBuilder)
         {
             base.OnModelCreating(modelBuilder);
 
-            // TODO: Code First this
+            // TODO: remove this when the discriminator convention is added
+            modelBuilder.Entity<Animal>().Discriminator()
+                .HasValue(typeof(Eagle), "Eagle")
+                .HasValue(typeof(Kiwi), "Kiwi");
```
AndriySvyryd: I'm not sure we need the parameterless overload. Since the user will specify the values perhaps we should force them to specify the type explicitly.

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -167,6 +165,8 @@ private bool InheritsFrom(EntityType entityType)
             return false;
         }
 
+        public virtual EntityType RootType() => (EntityType)((IEntityType)this).RootType();
```
anpete: This is unfortunate but can't think of an easy solution.

test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
anpete: Is this for the case where we introduce a shadow prop?

test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
anpete: Not sure. A few things spring to mind:

1) No generics for the derived types.
2) Feels more natural to configure the derived via Entity<TDerived>() but I guess you came up with this to keep strong typing.

E.g.

modelBuilder.Entity<Plant>().DiscriminatorProperty(p => p.Genus)
modelBuilder.Entity<Rose>().DiscriminatorValue(PlantGenus.Rose)

cc @divega, @bricelam, @ajcvickers

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -666,8 +664,8 @@ private IEnumerable<Navigation> FindDerivedNavigations(IEnumerable<string> names
                 .Cast<Navigation>();
```
smitpatel: for index and FK cast has been removed. Is this required?

test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
AndriySvyryd: @anpete 
1. There's no point for specifying derived type via a generic argument, since there's nothing to deduce it from and there's no API downstream that would benefit from it
2. Yes, I prefer the value to be strongly typed in the API. You can still add the value from the derived type:
```
modelBuilder.Entity<Rose>().Discriminator().HasValue(typeof(Rose), PlantGenus.Rose)
```
We could add sugar for this so it looks like yours

test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
AndriySvyryd: Yes

src/EntityFramework.Core/Metadata/EntityType.cs
AndriySvyryd: I can remove it

test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
anpete: @AndriySvyryd 

1) We also do it just because it just looks better and is easier to type. E.g. modelBuilder.Ignore<Foo>();
2) Yeah, your example looks kinda strange because its WET :-)

test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
anpete: So there is another one that takes the shadow prop name and is generic on the type? If so, then this no-arg one could go away.

test/EntityFramework.Relational.FunctionalTests/InheritanceRelationalFixture.cs
divega: It would be good to discuss this all up tomorrow in the design meeting. 

FWIW, I don't like the API a lot at first glance but I would like to make sure I understand what we are trying to achieve.

AndriySvyryd: We'll discuss this in the API review. It wouldn't affect the bulk of the commit though.

anpete:
:shipit:

PR #2814, opened 2015-08-11
Fix to 2274 - Chaining multiple Include with ThenInclude throws in some cases when included collections are empty mid-chain

maumar:
Problem is that when creating creating query plan for include we statically assign each index for each ReferenceEntity materializer that indicates from which reader the data should be taken, in order to materialize this entity.
However, in some cases, we don't open all the readers that we expect - this can happen if a reader doesn't return any results. We then short circuit the include chain, and don't open any readers that were suppose to follow.

Fix is to introduce a structure that holds active readers for include (on top of the old _activeQueries for all other readers), which we don't fill in like a list, but fill it based on the index that the reader is supposed to have (filling the blanks with nulls temporarily). This way we ensure that reader always land in their place, and are available at the appropriate index, that we introduced in the query plan. This happens whether all of them were opened, or some were short-circuited.

The information about which reader should be placed at which index is also added to the query plan. We do this only for the include-related readers.

### DIFF
--- a/src/EntityFramework.Relational/Query/RelationalQueryContext.cs
+++ b/src/EntityFramework.Relational/Query/RelationalQueryContext.cs
@@ -15,6 +15,7 @@ namespace Microsoft.Data.Entity.Query
     public class RelationalQueryContext : QueryContext
     {
         private readonly List<IValueBufferCursor> _activeQueries = new List<IValueBufferCursor>();
+        private readonly List<IValueBufferCursor> _activeIncludeQueries = new List<IValueBufferCursor>();
 
         private int _activeIncludeQueryOffset;
 
@@ -33,7 +34,7 @@ namespace Microsoft.Data.Entity.Query
 
         public virtual IRelationalConnection Connection { get; }
 
-        public virtual void RegisterValueBufferCursor([NotNull] IValueBufferCursor valueBufferCursor)
+        public virtual void RegisterValueBufferCursor([NotNull] IValueBufferCursor valueBufferCursor, int? queryIndex)
         {
             Check.NotNull(valueBufferCursor, nameof(valueBufferCursor));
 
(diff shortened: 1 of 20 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/RelationalQueryContext.cs
```diff
@@ -44,10 +45,15 @@ public virtual void RegisterValueBufferCursor([NotNull] IValueBufferCursor value
             }
 
             _activeQueries.Add(valueBufferCursor);
+
+            if (queryIndex != null && queryIndex.Value > 0)
```
anpete: Use null conditional

src/EntityFramework.Relational/Query/RelationalQueryContext.cs
```diff
@@ -58,6 +64,25 @@ public virtual void RegisterValueBufferCursor([NotNull] IValueBufferCursor value
...
```
anpete: null conditional

src/EntityFramework.Relational/Query/RelationalQueryContext.cs
```diff
@@ -65,10 +90,25 @@ public virtual void DeregisterValueBufferCursor([NotNull] IValueBufferCursor val
...
+            if (index >= 0)
+            {
+                _activeIncludeQueries[index] = null;
+            }
         }
 
         public virtual ValueBuffer GetIncludeValueBuffer(int queryIndex)
-            => _activeQueries[_activeIncludeQueryOffset + queryIndex].Current;
+        {
+            if (queryIndex == 0)
```
anpete: Ternary

anpete:
Nice work.

:shipit:

src/EntityFramework.Relational/Query/RelationalQueryContext.cs
maumar: doesn't work with Nullable<T>

anpete: Good point, missed that.

PR #2821, opened 2015-08-11
Query: Relational: Adds nav prop to nav prop/null comparison translat…

anpete:
…ions.

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
@@ -23,6 +23,18 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors
 
         private class NavigationJoin
         {
+            public static void RemoveNavigationJoin(
+                ICollection<NavigationJoin> navigationJoins, NavigationJoin navigationJoin)
+            {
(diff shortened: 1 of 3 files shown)


anpete:
@maumar

src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
```diff
@@ -99,6 +113,70 @@ protected override Expression VisitConstant(ConstantExpression constantExpressio
...
+                    if (constantExpression != null
+                        && constantExpression.Value == null)
+                    {
+                        newRight = rightNavigationJoin.JoinClause.OuterKeySelector;
+
+                        NavigationJoin.RemoveNavigationJoin(_navigationJoins, rightNavigationJoin);
+                    }
+                    else
+                    {
+                        newRight = rightNavigationJoin.JoinClause.InnerKeySelector;
```
maumar: are we guaranteed not to have convert on the right side of the binary op?

maumar:
:shipit:

src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
anpete: Removed the RemoveConvert from above as it was no longer needed.

PR #2823, opened 2015-08-11
AttributeConventions should run on private properties

### LINKED ISSUE #2766: PropertyAttributeConvention does not apply to non-public properties
Both `PropertyAttributeConvention <T>` and `NavigationAttributeConvention<T>` does not apply to protected and private properties. They use method `Type.GetProperty(name)` to get CLR property which always returns `null` for non-public ones.

smitpatel:
resolves #2766


natemcmaster:
:shipit:

PR #2831, opened 2015-08-12
Migrations: Make first-round API Review changes

bricelam:
Lots of renames and a few minor refactorings...


src/EntityFramework.Commands/Migrations/CSharpMigrationOperationGenerator.cs
```diff
@@ -307,7 +303,7 @@ protected virtual void Generate([NotNull] AlterColumnOperation operation, [NotNu
                 }
 
                 builder.AppendLine(",")
-                    .Append("nullable: ")
+                    .Append("isNullable: ")
```
AndriySvyryd: Couldn't these strings be replaced by nameof()?

anpete: Likely only in the scope of the owning method.

anpete: "isNullable" should be "nullable" to be idiomatic.

AndriySvyryd:
:shipit:

src/EntityFramework.Commands/Migrations/CSharpMigrationOperationGenerator.cs
bricelam: @anpete We have *is* prefixes on Fluent API parameters:
* IndexBuilder
    * Unique(isUnique)
    * SqlServerClustered(isClustered)
* PropertyBuilder
    * Required(isRequired)
    * ConcurrencyToken(isConcurrencyToken)
* ReferenceCollectionBuilder/ReferenceReferenceBuilder
    * Required(isRequired)
* KeyBuilder
    * SqlServerClustered(isClustered)

I'll leave for now, and we can discuss in the next API Review.

Issue #2820, opened 2015-08-11
Testing: rev-eng for SQLite


natemcmaster:
Add more extensive testing for SQLite reverse engineering. Current tests only examine simple case.
- [ ] Test custom templates
- [x] Test logging
- [x] More complex databases

PR #2833, opened 2015-08-12
Improvements to reveng on SQLite

natemcmaster:
Add nullable reverse typemapping.
Remove detection of "AUTOINCREMENT" -- this annotation is really only for migrations.
Detect Unique columns before adding foreign keys.
Assign uniqueness to foreign keys.
Add more functional testing.

#2820


lajones:
:shipit: Looks good.

Issue #1767, opened 2015-03-06
Support GreaterThan LessThan etc for string comparisons


cherrydev:
I believe in earlier versions the expression that would, for example, generate a SQL statement of: `param > 'a' ` would be `.Where(x => String.CompareTo(x.param, "a", StringComparison.Ordinal)) > 0`. Perhaps the StringComparison.Ordinal was unnecessary and defaulted to a sane value. In EF7, the parser just punts when trying to compare strings and performs the comparison in-memory.

maumar:
Fixed in eeb26eaea19f42320c2c00928f1be9d5aa21ae16

toralux:
I still have issue with #1767, #3464, #5369 as the LINQ expression is evaluated locally and not on server-side as expected.

warn: Microsoft.EntityFrameworkCore.Query.Internal.SqlServerQueryCompilationContextFactory[8]
      The LINQ expression '([x].StringField.CompareTo("2000") > 0)' could not be translated and will be evaluated locally.

Running "Microsoft.EntityFrameworkCore": "1.1.0"

rowanmiller:
@toralux please open a new issue and include a code example

PR #2836, opened 2015-08-13
Fix to #1767 - Support GreaterThan LessThan etc for string comparisons

maumar:
Added support for string.Compare(string, string) method, which translates to:
CASE WHEN (strA = strB) THEN 0 ELSE (CASE WHEN (strA > strB) THEN 1 ELSE -1 END) END

We also recognize pattern like: string.Compare(strA, strB) == 0
This translates to much simpler form: strA = strB (also works with !=, >, >=, <, <=)

Providers need to opt-in to the sting order comparability in order for those optimizations to take effect.


maumar:
@anpete

src/EntityFramework.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -25,6 +27,8 @@ public class SqlTranslatingExpressionVisitor : ThrowingExpressionVisitor
 
         private readonly bool _bindParentQueries;
         private readonly bool _inProjection;
+        private static MethodInfo _stringCompareMethodInfo = typeof(string).GetTypeInfo().GetDeclaredMethods("Compare")
```
maumar: we could also support string.Compare(string, string, StringComparison) and ignore the third argument - not sure if it's worth though

src/EntityFramework.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -174,6 +195,53 @@ private static Expression UnfoldStructuralComparison(ExpressionType expressionTy
...
+                var rightConstant = expression.Right as ConstantExpression;
+                if (rightConstant != null && rightConstant.Type == typeof(int) && (int)rightConstant.Value == 0 && leftMethod != null)
+                {
+                    if (leftMethod.Method == _stringCompareMethodInfo)
+                    {
+                        var newleftString = Visit(leftMethod.Arguments[0]);
+                        var newRightString = Visit(leftMethod.Arguments[1]);
+
+                        return new SqlFunctionExpression(
+                            "__StringComparisonTemplate",
```
maumar: not really happy about the name, also might be worthwhile to store the "magic constant" in one place and reuse it. Thoughts? I need it here, in the StringCompareTranslator and in DefaultQuerySqlGenerator - what would be a good place to store it?

src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
```diff
@@ -174,7 +174,12 @@ var rightNavigationJoin
                 }
             }
 
-            return Expression.MakeBinary(binaryExpression.NodeType, newLeft, newRight);
+            return Expression.MakeBinary(
```
anpete: Curious, when are the extra params needed?

src/EntityFramework.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
anpete: Can the third arg affect translations semantics?

src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
maumar: string concatenation (strA + strB)

src/EntityFramework.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
divega: A static in StringCompareTranslator?

divega: BTW, not to change this PR but it would be very nice if in the future we could figure out a way that adding functionality like this didn't require touching so many places, e.g. if it was somehow possible that the changes required to the visitor, the translation and the SQL generation for a specific operation were encapsulated in a single place.

maumar: Moved this code from SqlTranslatingVisitor into a separate (post-processing) step, after we are done with the query model. In the future we can make it more extensible, but all the post processing should happen there

maumar:
@anpete @divega updated with CR feedback

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -129,10 +129,14 @@ var compositePredicateVisitor
                         .GetCustomQueryAnnotations(RelationalQueryableExtensions.UseRelationalNullSemanticsMethodInfo)
                         .Any());
 
+            var stringCompareOptimizer = new StringCompareExpressionOptimizer();
             foreach (var selectExpression in _queriesBySource.Values.Where(se => se.Predicate != null))
             {
                 selectExpression.Predicate
                     = compositePredicateVisitor.Visit(selectExpression.Predicate);
+
+                selectExpression.Predicate
```
maumar: done only for predicate for now enabling it for projection etc will require some extra work - simply making SelectExpression more mutable might be problematic, because many visitors depend on reference equality comparison to determine whether given expression has changed - if we end up mutating SelectExpression , we might miss some of those changes in a nested subquery scenarios

src/EntityFramework.Core/Query/ExpressionVisitors/FunctionEvaluationDisablingVisitor.cs
```diff
@@ -42,7 +42,8 @@ protected override Expression VisitMethodCall(MethodCallExpression expression)
             var newObject = Visit(expression.Object);
             var newArguments = VisitAndConvert(expression.Arguments, "VisitMethodCall");
 
-            var newMethodCall = newObject != expression.Object || newArguments != expression.Arguments
+            var newMethodCall = newObject != expression.Object || 
+                !Enumerable.SequenceEqual(newArguments, expression.Arguments)
```
anpete: SequenceEqual is an extension method. Isn't reference equality enough here?

src/EntityFramework.Core/Query/Expressions/MethodCallEvaluationPreventingExpression.cs
```diff
@@ -31,7 +32,7 @@ protected override Expression VisitChildren(ExpressionVisitor visitor)
             var newArguments = visitor.VisitAndConvert(MethodCall.Arguments, "VisitChildren");
 
             if (newObject != MethodCall.Object
-                || newArguments != MethodCall.Arguments)
+                || !Enumerable.SequenceEqual(newArguments,MethodCall.Arguments))
```
anpete: As above.

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
```diff
@@ -0,0 +1,129 @@
...
+                        typeof(bool));
+                }
+            }
+
+            return base.VisitBinary(node);
+        }
+
+        private bool TryOptimizeStringCompare(ConditionalExpression outerConditional, out Expression leftString, out Expression rightString)
+        {
+            // we are looking for expression like:
```
anpete: What does the optimization look like?

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
```diff
@@ -0,0 +1,129 @@
...
+using Microsoft.Data.Entity.Query.Expressions;
+using Microsoft.Data.Entity.Query.Methods;
+
+namespace Microsoft.Data.Entity.Query.ExpressionVisitors
+{
+    public class StringCompareExpressionOptimizer : RelinqExpressionVisitor
+    {
+        protected override Expression VisitBinary(BinaryExpression node)
+        {
+            if (node.NodeType != ExpressionType.Equal && node.NodeType != ExpressionType.NotEqual
```
anpete: Switch statement?

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
anpete: This doesn't matter if nothing holds on to the stale state.

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -78,6 +79,22 @@ public virtual string GenerateSql(IDictionary<string, object> parameterValues)
         protected virtual string TypedTrueLiteral => "CAST(1 AS BIT)";
         protected virtual string TypedFalseLiteral => "CAST(0 AS BIT)";
 
+        protected virtual Dictionary<ExpressionType, string> BinaryOperatorMap => new Dictionary<ExpressionType, string>
+        {
```
anpete: This should be static. Provide a template method for extension.

src/EntityFramework.SqlServer/Query/Methods/StringReplaceTranslator.cs
```diff
@@ -12,13 +12,13 @@ namespace Microsoft.Data.Entity.SqlServer.Query.Methods
 {
     public class StringReplaceTranslator : IMethodCallTranslator
     {
+        private static readonly MethodInfo _methodInfo = typeof(string).GetTypeInfo().GetDeclaredMethods("Replace")
+            .Where(m => m.GetParameters()[0].ParameterType == typeof(string))
```
anpete: nameof(string.Replace)

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -2949,6 +2949,175 @@ public override void String_Contains_MethodCall()
...
+    THEN 0 ELSE CASE
+        WHEN
+            ([c].[CustomerID] > 'M' + [c].[CustomerID])
+        THEN 1 ELSE -1
+    END
+END >= 42)
+
+SELECT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM [Customers] AS [c]
+WHERE (CASE
```
anpete: Can we improve the formatting here?

src/EntityFramework.SqlServer/Query/Methods/StringSubstringTranslator.cs
```diff
@@ -12,13 +12,13 @@ namespace Microsoft.Data.Entity.SqlServer.Query.Methods
 {
     public class StringSubstringTranslator : IMethodCallTranslator
     {
+        private static readonly MethodInfo _methodInfo = typeof(string).GetTypeInfo().GetDeclaredMethods("Substring")
```
anpete: nameof

src/EntityFramework.Core/Query/ExpressionVisitors/FunctionEvaluationDisablingVisitor.cs
maumar: VisitAndConvert returns a new instance of the list even if the arguments are not changed

anpete: Are you sure? I've seen a lot of code that just does reference equality on the collection.

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
maumar: strA [operator] strB, basically this method recognizes the pattern and extracts strA and strB to be used to construct the resulting expression (line 45)

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
maumar: was planning to do this in a next checkin - recognize the pattern like this and transform it to:
CASE 
    WHEN 
        condition1
    THEN 
        result1
    WHEN
        condition2
    THEN
        result2
    ELSE
        result3
END

maumar:
@anpete pushed an updated version, please take a look

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -840,6 +823,36 @@ public virtual Expression VisitLiteral(LiteralExpression literalExpression)
 
         public virtual Expression VisitSqlFunction(SqlFunctionExpression sqlFunctionExpression)
         {
+            if (sqlFunctionExpression.FunctionName == StringCompareTranslator.StringCompareMethodName)
+            {
+                var arguments = sqlFunctionExpression.Arguments.ToList();
+                if (sqlFunctionExpression.Type == typeof(bool))
```
maumar: should I use two different names for this, rather than just distinguish based on return type? Those are:
bool Method(string left, string right, ExpressionType op)
int Method(string left, string right)

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Linq.Expressions;
+using Remotion.Linq.Parsing;
+using Microsoft.Data.Entity.Query.Expressions;
+using Microsoft.Data.Entity.Query.Methods;
+
+namespace Microsoft.Data.Entity.Query.ExpressionVisitors
+{
+    public class StringCompareExpressionOptimizer : RelinqExpressionVisitor
+    {
+        private static Dictionary<ExpressionType, ExpressionType> _operatorMap = new Dictionary<ExpressionType, ExpressionType>
```
anpete: readonly

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
```diff
@@ -0,0 +1,69 @@
...
+            {  ExpressionType.LessThanOrEqual, ExpressionType.GreaterThanOrEqual },
+            {  ExpressionType.GreaterThan, ExpressionType.LessThan },
+            {  ExpressionType.GreaterThanOrEqual, ExpressionType.LessThanOrEqual },
+            {  ExpressionType.Equal, ExpressionType.Equal },
+            {  ExpressionType.NotEqual, ExpressionType.NotEqual },
+        };
+
+        protected override Expression VisitBinary(BinaryExpression node)
+        {
+            if (node.NodeType != ExpressionType.Equal && node.NodeType != ExpressionType.NotEqual
```
anpete: Weren't you going to make an extension method for this?

anpete: Can you just use the map above for this check?

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            var leftSqlFunction = node.Left as SqlFunctionExpression;
+            var rightConstant = node.Right as ConstantExpression;
+            if (leftSqlFunction != null 
+                && leftSqlFunction.FunctionName == StringCompareTranslator.StringCompareMethodName
+                && leftSqlFunction.Type == typeof(int)
+                && rightConstant != null 
+                && rightConstant.Type == typeof(int) 
+                && (int)rightConstant.Value == 0)
```
anpete: Do you need the Type checks?

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
```diff
@@ -0,0 +1,69 @@
...
+                var arguments = leftSqlFunction.Arguments.ToList();
+                return new SqlFunctionExpression(
+                    StringCompareTranslator.StringCompareMethodName,
+                        new[] { arguments[0], arguments[1], Expression.Constant(node.NodeType) },
+                        typeof(bool));
+            }
+
+            var leftConstant = node.Left as ConstantExpression;
+            var rightSqlFunction = node.Right as SqlFunctionExpression;
+            if (rightSqlFunction != null
```
anpete: Do you need the Type checks?

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -30,6 +31,22 @@ public class DefaultQuerySqlGenerator : ThrowingExpressionVisitor, ISqlExpressio
         private IDictionary<string, object> _parameterValues;
         private int _rawSqlParameterIndex;
 
+        private static Dictionary<ExpressionType, string> _binaryOperatorMap => new Dictionary<ExpressionType, string>
```
anpete: readonly

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -940,6 +953,11 @@ protected virtual string GenerateParameterName([NotNull] string parameterName)
             return parameterName;
         }
 
+        protected virtual Dictionary<ExpressionType, string> GetBinaryOperatorMap()
+        {
```
anpete: I think this fit better if the signature was something like: string GenerateBinaryOperator(ExpressionType)

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
anpete: Is this code path still needed?

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
```diff
@@ -0,0 +1,69 @@
...
+                && rightSqlFunction.FunctionName == StringCompareTranslator.StringCompareMethodName
+                && rightSqlFunction.Type == typeof(int)
+                && leftConstant != null
+                && leftConstant.Type == typeof(int)
+                && (int)leftConstant.Value == 0)
+            {
+                var arguments = rightSqlFunction.Arguments.ToList();
+                return new SqlFunctionExpression(
+                    StringCompareTranslator.StringCompareMethodName,
+                        new[] { arguments[0], arguments[1], Expression.Constant(_operatorMap[node.NodeType]) },
```
anpete: I'm still missing how this works with the bool return type and node type third argument.

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
maumar: StringCompareMethod can have a return type of either int (with 2 arguments), which we want to optimize, eg:
SqlFunction_returning_int(strA, strB) > 0
this gets optimized into 
SqlFunction_returning_bool(strA, strb, '>')
How if we have a nested scenario we want to make sure that we wont incorrectly try to optimize out the latter expression. We could probably get rid of one of the type checks (assuming that if one of them is int, the second one won't be a bool), but might as well leave both, just in case

src/EntityFramework.Relational/Query/ExpressionVisitors/StringCompareExpressionOptimizer.cs
maumar: 2-argument function returning bool, compared to constant 0 will be transformed into 2-argument function returning bool, the third argument being the operator between the original function and the constant 0, eg:
func(strA, strB) >= 0  -> func(strA, strB, '>=')
During sql generation, we recognize both of those, and depending on the return type, we generate either:
CASE WHEN strA = strB THEN 0 ELSE CASE WHEN strA > strB THEN 1 ELSE -1 (for the 2 arg and int return type)

or strA [operator, the third argument] strB (for the 3 arg and bool return type)

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
maumar: yes, in case we can't optimize expression to the simple form, e.g. 
String.Compare(strA, strB) > 42, or should we just simplify this to "false"?

src/EntityFramework.Relational/Query/ExpressionTranslators/StringCompareTranslator.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using Microsoft.Data.Entity.Query.Expressions;
+using Remotion.Linq.Parsing;
+
+namespace Microsoft.Data.Entity.Query.ExpressionTranslators
+{
+    public class StringCompareTranslator : IExpressionFragmentTranslator<BinaryExpression>, IExpressionFragmentTranslator
```
maumar: two separate interfaces to work around the covariance issue in RelationalCompositeExpressionFragmentTranslator. Any better way to deal with this problem?

maumar:
@anpete @divega  new version up design has been considerably changed from previous iterations, please take a look

src/EntityFramework.Relational/Query/ExpressionTranslators/RelationalCompositeExpressionFragmentTranslator.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.Data.Entity.Query.ExpressionTranslators
+{
+    public class RelationalCompositeExpressionFragmentTranslator : ICompositeExpressionFragmentTranslator
+    {
+        private List<IExpressionFragmentTranslator> _relationalTranslators;
+
+        public RelationalCompositeExpressionFragmentTranslator()
+        {
+            _relationalTranslators = new List<IExpressionFragmentTranslator>
```
anpete: Initialize inline?

src/EntityFramework.Relational/Query/ExpressionTranslators/StringCompareTranslator.cs
```diff
@@ -0,0 +1,133 @@
...
+                var rightString = visitor.Visit(arguments[1]);
+                if (leftString == null || rightString == null)
+                {
+                    return null;
+                }
+
+                if ((int)rightConstant.Value == 0)
+                {
+                    // Compare(strA, strB) > 0 => strA > strB
+                    return new StringCompareExpression(leftString, rightString, expression.NodeType);
```
anpete: Move NodeType to first arg to match BinaryExpression?

src/EntityFramework.Relational/Query/ExpressionTranslators/StringCompareTranslator.cs
```diff
@@ -0,0 +1,133 @@
...
+                }
+
+                if ((int)rightConstant.Value == 1 && expression.NodeType == ExpressionType.Equal)
```
anpete: Can probably simplify this code by first checking the constant and then the node type.

src/EntityFramework.Relational/Query/ExpressionTranslators/StringCompareTranslator.cs
```diff
@@ -0,0 +1,133 @@
...
+                }
+
+                if ((int)rightConstant.Value == -1 && expression.NodeType == ExpressionType.GreaterThan)
+                {
+                    // Compare(strA, strB) > -1 => strA >= strB
+                    return new StringCompareExpression(leftString, rightString, ExpressionType.GreaterThanOrEqual);
+                }
+            }
+
+            var leftConstant = expression.Left as ConstantExpression;
```
anpete: Can you extract method on these two code blocks and then call the helper with the left/right flipped?

src/EntityFramework.Relational/Query/ExpressionTranslators/StringCompareTranslator.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Linq.Expressions;
+using System.Reflection;
+using Microsoft.Data.Entity.Query.Expressions;
+using Remotion.Linq.Parsing;
+
+namespace Microsoft.Data.Entity.Query.ExpressionTranslators
+{
+    public class StringCompareTranslator : IExpressionFragmentTranslator<BinaryExpression>, IExpressionFragmentTranslator
+    {
+        private static readonly Dictionary<ExpressionType, ExpressionType> _operatorMap = new Dictionary<ExpressionType, ExpressionType>
```
anpete: Is the map part of this still used?

maumar:
@anpete new version is up - applied recent CR feedback

src/EntityFramework.Relational/Query/ExpressionTranslators/RelationalCompositeExpressionFragmentTranslator.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.Data.Entity.Query.ExpressionTranslators
+{
+    public class RelationalCompositeExpressionFragmentTranslator : IExpressionFragmentTranslator
+    {
+        private List<IExpressionFragmentTranslator> _relationalTranslators;
+
+        public RelationalCompositeExpressionFragmentTranslator()
+        {
+            _relationalTranslators = new List<IExpressionFragmentTranslator> { new StringCompareTranslator() };
```
anpete: Initialize inline?

src/EntityFramework.Relational/Query/ExpressionTranslators/RelationalCompositeExpressionFragmentTranslator.cs
```diff
@@ -0,0 +1,35 @@
...
+                if (result != null)
+                {
+                    return result;
+                }
+            }
+
+            return null;
+        }
+
+        protected virtual IReadOnlyList<IExpressionFragmentTranslator> Translators => _relationalTranslators;
```
anpete: What is this used for?

src/EntityFramework.Relational/Query/ExpressionTranslators/StringCompareTranslator.cs
```diff
@@ -0,0 +1,116 @@
...
+                && methodCall.Method == _methodInfo
+                && methodCall.Type == typeof(int)
+                && constant != null
+                && constant.Type == typeof(int))
+            {
+                var arguments = methodCall.Arguments.ToList();
+                var leftString = arguments[0];
+                var rightString = arguments[1];
+
+                if ((int)constant.Value == 0)
```
anpete: Cache the casted value in a local?

anpete: Looks better here BTW.

src/EntityFramework.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -44,6 +44,22 @@ public class SqlTranslatingExpressionVisitor : ThrowingExpressionVisitor
...
 
+        public override Expression Visit(Expression expression)
+        {
+            var translatedExpression
+                = _queryModelVisitor
+                    .QueryCompilationContext
+                    .CompositeExpressionFragmentTranslator
+                    .Translate(expression);
+
+            if (translatedExpression != null && translatedExpression != expression)
```
anpete: Nice.

anpete:
:shipit:

src/EntityFramework.Relational/Query/ExpressionTranslators/RelationalCompositeExpressionFragmentTranslator.cs
maumar: other providers can add their own translators, we use similar pattern for MethodCallTranslator

anpete: It's readonly :-) Better to hide the underlying list and provide an Add method the provider can call. Make this change in MethodCallTranslator too if required.

PR #2838, opened 2015-08-13
Fix bugs in SQLite reveng foreign key

### LINKED ISSUE #2820: Testing: rev-eng for SQLite

natemcmaster:
Handles quirks in SQLite's foreign key behavior which allows FK to a unique index.

Fix bug in model generation.

Adds more testing to bring it up to par with SqlServer.Design testing. Fix #2820


natemcmaster:
cc: @lajones after this I'll be switching to work on different parts of EF. If there is anything else to add for SQLite reveng or  #2820, let me know.

ErikEJ:
@lajones  I assume you know that SQL Server also allows foreign keys to a unique index (instead of a primary key) ?

natemcmaster:
On a related note, should we reverse engineering generate keys in the metadata for each unique index on a table?

lajones:
@ErikEJ @natemcmaster - raised issue #2840 to consider issues raised by alternate keys / indexes etc.

lajones:
@natemcmaster other than the issues raised in #2840 - I would make sure you are testing all nullable and non-nullable data types that SQLite supports, and composite keys. SQLite also raises the issue of what happens if say a string is stored in a column which is nominally an integer, does that affect reveng at all?

lajones:
:shipit: Looking through - it looks like you already covered what I was suggesting above.

natemcmaster:
@lajones thanks. It's worth mentioning here that we are only reverse engineer the column types from the CREATE statements. Although we could be more rigorous by analyzing the actual data, that's a rabbit hole I don't want to go down. For now, we do the simple thing in choosing clr types.

PR #2841, opened 2015-08-14
Adopting latest CoreCLR packages

troydai:
@bricelam


bricelam:
:shipit:

troydai:
@bricelam I can't get all the tests passed locally with or without this change. Can you try run the tests against this change before I check in?

PR #2845, opened 2015-08-14
Remove binding redirect of System.Collections.Immutable

troydai:
To prevent assembly load loop in DNX.

@bricelam @davidfowl


lajones:
You can probably just remove the files altogether if we no longer need the binding redirect.

test/EntityFramework.Commands.FunctionalTests/App.config
```diff
@@ -4,7 +4,6 @@
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
         <assemblyIdentity name="System.Collections.Immutable" publicKeyToken="b03f5f7f11d50a3a"/>
-        <bindingRedirect oldVersion="1.1.36.0" newVersion="1.1.37.0"/>
```
divega: Does the assemblyBinding element have any purpose after removing the bindingRedirect?

troydai:
I'll remove the entire file.

davidfowl:
Those files never worked until @anurse 's checkin yesterday which is causing problems.

troydai:
:watch:

troydai:
I'll merge soon to unblock the CI.

divega:
:shipit: to unblock the CI. I assume these were added to workaround some issue. If the issue comes back we can revisit.

natemcmaster:
We needed those to run tests in Visual Studio. Most of the team uses TestDriven.NET.

PR #2822, opened 2015-08-11
Allow the user to specify the project sub-directory and namespace

lajones:
Fix for issue #2750.

Issue #2750, opened 2015-07-30
RevEng: Change output directory (and namespace)


bricelam:
In EF6, you could add the model to a specific folder. (E.g. `Models`.) The namespace would also reflect the target directory. (E.g. `WebApplication1.Models`) We should consider adding similar functionality in EF7.

lajones:
+1

lajones:
Fixed with PR #2847

PR #2847, opened 2015-08-14
Allow the user to specify the project sub-directory and namespace

lajones:
Fix for issue #2750. (Replaces PR #2822).
Also updated -dnx.sln file to include SQLite projects (and generated their .xproj files).

### DIFF
--- a/src/EntityFramework.Commands/Executor.cs
+++ b/src/EntityFramework.Commands/Executor.cs
@@ -261,11 +261,11 @@ namespace Microsoft.Data.Entity.Commands
 
                 var connectionString = (string)args["connectionString"];
                 var providerAssemblyName = (string)args["provider"];
-                var @namespace = (string)(args.Contains("namespace") ? args["namespace"] : null);
+                var customNamespace = (string)(args.Contains("namespace") ? args["namespace"] : null);
                 var relativeOutputDir = (string)(args.Contains("relativeOutputDir") ? args["relativeOutputDir"] : null);
 
                 Execute(() => executor
-                    .ReverseEngineerImplAsync(providerAssemblyName, connectionString, @namespace, relativeOutputDir)
+                    .ReverseEngineerImplAsync(providerAssemblyName, connectionString, customNamespace, relativeOutputDir)
                     .GetAwaiter().GetResult());
             }
         }
@@ -273,12 +273,12 @@ namespace Microsoft.Data.Entity.Commands
         public virtual Task<IReadOnlyList<string>> ReverseEngineerImplAsync(
             [NotNull] string providerAssemblyName,
             [NotNull] string connectionString,
-            [CanBeNull] string @namespace,
+            [CanBeNull] string customNamespace,
             [CanBeNull] string relativeOutputDir,
             CancellationToken cancellationToken = default(CancellationToken))
         {
             return CreateDatabaseTool().ReverseEngineerAsync(
-                providerAssemblyName, connectionString, @namespace ?? _rootNamespace,
+                providerAssemblyName, connectionString, customNamespace, _rootNamespace,
                 _projectDir, relativeOutputDir, cancellationToken);
         }
 
--- a/src/EntityFramework.Commands/Executor.cs
+++ b/src/EntityFramework.Commands/Executor.cs
@@ -261,18 +261,25 @@ namespace Microsoft.Data.Entity.Commands
 
                 var connectionString = (string)args["connectionString"];
                 var providerAssemblyName = (string)args["provider"];
+                var @namespace = (string)(args.Contains("namespace") ? args["namespace"] : null);
+                var relativeOutputDir = (string)(args.Contains("relativeOutputDir") ? args["relativeOutputDir"] : null);
 
-                Execute(() => executor.ReverseEngineerImplAsync(providerAssemblyName, connectionString).GetAwaiter().GetResult());
+                Execute(() => executor
+                    .ReverseEngineerImplAsync(providerAssemblyName, connectionString, @namespace, relativeOutputDir)
+                    .GetAwaiter().GetResult());
             }
         }
 
         public virtual Task<IReadOnlyList<string>> ReverseEngineerImplAsync(
             [NotNull] string providerAssemblyName,
             [NotNull] string connectionString,
+            [CanBeNull] string @namespace,
+            [CanBeNull] string relativeOutputDir,
             CancellationToken cancellationToken = default(CancellationToken))
         {
             return CreateDatabaseTool().ReverseEngineerAsync(
-                providerAssemblyName, connectionString, _rootNamespace, _projectDir, cancellationToken);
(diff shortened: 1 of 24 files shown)


test/EntityFramework.Relational.Design.Tests/ReverseEngineering/ReverseEngineeringGeneratorTests.cs
```diff
@@ -0,0 +1,55 @@
...
+                Assert.Throws<ArgumentException>(
+                    () => GetGenerator().GenerateAsync(configuration).GetAwaiter().GetResult()).Message);
+
+            configuration.ProjectPath = "NonEmptyProjectPath";
+            Assert.Equal(Strings.NamespaceRequired,
+                Assert.Throws<ArgumentException>(
+                    () => GetGenerator().GenerateAsync(configuration).GetAwaiter().GetResult()).Message);
+
+            configuration.RelativeOutputPath = @"\NotARelativePath";
+            Assert.Equal(Strings.NotRelativePath(@"\NotARelativePath"),
```
lajones: @natemcmaster - re your previous comment in #2822 - have added some testing that the path has to be relative.

lajones:
Ping...

src/EntityFramework.Commands/Executor.cs
```diff
@@ -261,18 +261,25 @@ public ReverseEngineer([NotNull] Executor executor, [NotNull] object resultHandl
...
         public virtual Task<IReadOnlyList<string>> ReverseEngineerImplAsync(
             [NotNull] string providerAssemblyName,
             [NotNull] string connectionString,
+            [CanBeNull] string @namespace,
+            [CanBeNull] string relativeOutputDir,
             CancellationToken cancellationToken = default(CancellationToken))
         {
             return CreateDatabaseTool().ReverseEngineerAsync(
-                providerAssemblyName, connectionString, _rootNamespace, _projectDir, cancellationToken);
+                providerAssemblyName, connectionString, @namespace ?? _rootNamespace,
```
natemcmaster: It would be more clear that `_rootNamespace` is the fallback value if it were assigned on line 264 rather than in the function call here.

src/EntityFramework.Commands/Program.cs
```diff
@@ -130,6 +130,12 @@ public virtual int Main([NotNull] string[] args)
                             var provider = scaffold.Argument(
                                 "[provider]",
                                 "The provider to use. For example, EntityFramework.SqlServer");
+                            var @namespace = scaffold.Option(
+                                "-n|--namespace <context_namespace>",
+                                "Overrides the namespace of the generated classes.");
```
natemcmaster: Drop the word "overrides". 

Add the second clause in the powershell help. This makes it a more clear that namespace and output directory are optional.

src/EntityFramework.Commands/tools/EntityFramework.psm1
```diff
@@ -320,6 +320,12 @@ Register-TabExpansion Scaffold-DbContext @{
 .PARAMETER Provider
     Specifies the provider to use. For example, EntityFramework.SqlServer.
 
+.PARAMETER Namespace
+    Overrides the namespace for the output classes. If omitted, the root namespace of the project is used.
```
natemcmaster: ~~Overrides~~
Same as before

src/EntityFramework.Commands/Executor.cs
lajones: Fair enough. I'll fix.

natemcmaster:
So questions on behavior: what should happen for when the user specifies:

 - Relative, up levels `../../output`
 - Absolute `C:\output\`
 - Non-existent folder `./folder_that_does_exists_yet`

src/EntityFramework.Commands/Program.cs
lajones: OK. Sounds good.

lajones:
Absolute - you'll get an error. Non-existent - it'll create the folder for you. 

For up-levels I haven't decided yet - and of course it can be more complicated still e.g. "looks\like\subdir\..\..\..\..\..\..\but\really\is\not". At the moment it will create the directory where you specify even if it is outside the project (provided of course you have permission to do so). I've been thinking of turning the full path into some kind of canonical path and then checking that that canonical path still starts with the project path i.e. you are absolutely, even including all possible ".." parts, still a sub-directory. And throwing an error if not. But was planning on a TODO for that.

natemcmaster:
Absolute - is this deliberate? Or just an implementation error? We should still allow absolute paths.

Non-existent -- good :+1:

Up-levels -- we can collapse them using Path.Combine and Path.GetFullPath

lajones:
The approach to absolute is deliberate. I want to only allow the user to specify a path which is within the project. The project path is already passed in automatically by either Program.cs or Executor.cs - so all I need is the relative path.

Up-levels - yes - sounds good.

bricelam:
IMHO, it seems silly like overkill to let users specify the namespace. We should always just gen as `RootNamespace.Directory`. If the user follows some non-standard namespace, they can change it after generation.

lajones:
I agree that we should update the namespace to reflect the output directory. I will add that (and that makes it even more important that we only allow relative paths).
The purpose of being able to override the namespace is for if you are generating lots of files. I guess it's still true that you can fix that with one big search-and-replace but it's also easy for us to provide the ability. I guess I won't fight hard to keep the feature if there are more people who don't like it.

bricelam:
One of our guiding principles in the EF7 commands was to have fewer command-line parameters and smarter conventions.
/cc @rowanmiller in case he has an opinion.

src/EntityFramework.Commands/Executor.cs
lajones: Actually this code is in a separate class and doesn't have access to _rootNamespace.

lajones:
@bricelam - true. But this is not mandatory - if the user omits it we do as you say. And it seems (to me) to set us up well for when we integrate with VS and have a dialog that is asking these questions. But as I say, if there are more people who don't want this feature then I can easily remove.

src/EntityFramework.Commands/Executor.cs
natemcmaster: Good point. nvm

rowanmiller:
Agree that we don't need this knob, it's a widely accepted coding guideline that namespaces match directory structure. If you really want to change it then you can just add the templates and modify it there.

lajones:
@bricelam @rowanmiller OK. Latest commit removes that functionality.

lajones:
@natemcmaster have put in the code to use GetFullPath() to check that relativeOutputDir must be a relative sub-directory of the project dir.

natemcmaster:
IMO absolute paths should still be allowed, but the code looks fine. :shipit:

natemcmaster:
Merged 5b19fbbff82987ba9e1aafe051ff8c4fd02bf8cf

PR #2853, opened 2015-08-17
Allow setting DateTime.UtcNow as default value for DateTime columns

dcga:
Currently if "getutcdate()" is set as the DefaultValue of a DateTime column, the quoted string "getutcdate()" is generated by EF.

### DIFF
--- a/src/EntityFramework.SqlServer/Query/Methods/DateTimeNowTranslator.cs
+++ b/src/EntityFramework.SqlServer/Query/Methods/DateTimeNowTranslator.cs
@@ -15,10 +15,16 @@ namespace Microsoft.Data.Entity.SqlServer.Query.Methods
         public virtual Expression Translate([NotNull] MemberExpression memberExpression)
         {
             if (memberExpression.Expression == null
-                && memberExpression.Member.DeclaringType == typeof(DateTime)
-                && memberExpression.Member.Name == "Now")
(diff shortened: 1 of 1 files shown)


divega:
@dcga is there a reason you closed the PR?

ajcvickers:
@divega This should be set with DefaultExpression, or whatever we decided to call that API.

bricelam:
Related: [EF 7 set initial default value for DateTime column](http://stackoverflow.com/q/32049742/475031) on StackOverflow

divega:
Although it is unrelated to the title the change in the PR seemed to add support for DateTime.UtcNow to our LINQ provider. I haven't tested (and there were not tests) but it seemed to be a good change at first glance :smile:

dcga:
@divega your last comment sums up the reason on why I closed the PR.
At first I thought it was related to the StackOverflow question linked above, then I realized it wasn't related, and on top of it I didn't have tests.
Thanks

divega:
@dcga please consider filing a bug for `DateTime.UtcNow` support and submitting a new PR. For testing it should be enough to replicate what we do for `DateTime.Now` (see [here] (https://github.com/aspnet/EntityFramework/blob/4a58d4d43e2eb4aac9ef7c4ee0410bc45d946beb/test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs#L877)).

PR #2856, opened 2015-08-17
Use DI in reveng

natemcmaster:
Remove code-gen helper construction to a factory.
Rename base class to RelationalMetadataModelProvider.
Changes the implementation requirements for providers.
Removes unused code from CSharp helpers.

There are more places to use DI, but enabling DI there will require more significant restructuring.


src/EntityFramework.Relational.Design/CodeGeneration/CSharpCodeGeneratorHelper.cs
```diff
@@ -11,9 +11,9 @@
 
 namespace Microsoft.Data.Entity.Relational.Design.CodeGeneration
 {
-    public class CSharpCodeGeneratorHelper
+    public static class CSharpCodeGeneratorHelper
```
bricelam: Interesting, we seem to have come full circle here. See #1646

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringConfiguration.cs
```diff
@@ -1,18 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System.Reflection;
-using JetBrains.Annotations;
-
-namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
-{
-    public class ReverseEngineeringConfiguration
-    {
```
lajones: The idea here was that if a provider required that you pass in additional info they could just override the ReverseEngineeringConfiguration class and pass that into an overridden ReverseEngineeringGenerator. You could define a new method into which to pass this info but then you'll still be inheriting the old method. So prefer the old way.

src/EntityFramework.Relational.Design/CodeGeneration/CSharpCodeGeneratorHelper.cs
lajones: Agreed. Plus I thought we were avoiding static methods if we could so as to make it more testable (you could write your own CSharpCodeGeneratorHelper, override the methods you needed and pass it in to test other parts of the code). That said we're no longer using Moq where this was a very common approach - so may not be as important any more.

src/EntityFramework.Relational.Design/ReverseEngineering/EntityTypeCodeGeneratorHelper.cs
```diff
@@ -6,43 +6,38 @@
...
+            [NotNull] ModelUtilities modelUtilities)
         {
             Check.NotNull(generatorModel, nameof(generatorModel));
             Check.NotNull(extensionsProvider, nameof(extensionsProvider));
+            Check.NotNull(modelUtilities, nameof(modelUtilities));
 
             GeneratorModel = generatorModel;
             ExtensionsProvider = extensionsProvider;
+            _modelUtilities = modelUtilities;
         }
```
lajones: Not sure why are we putting this in 2 classes (the DbContext and EntityType Helpers) and hence keeping copies of the pointer in multiple different objects when having it on the Generator meant we would only have it once?

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
```diff
@@ -203,42 +187,11 @@ public class ReverseEngineeringGenerator
...
-        {
-            Check.NotNull(provider, nameof(provider));
+        public virtual string GetDbContextTemplateFileName() 
+            => _provider.GetType().GetTypeInfo().Assembly.GetName().Name + "." + DbContextTemplateFileName;
 
-            return provider.GetType().GetTypeInfo().Assembly.GetName().Name + "." + EntityTypeTemplateFileName;
-        }
-
-        private static void CheckConfiguration(ReverseEngineeringConfiguration configuration)
-        {
```
lajones: Am I missing where you are checking these now? In my other PR for instance this is where we check that the output path passed in is relative. I'd prefer to fail fast with a helpful message.

src/EntityFramework.Relational.Design/ReverseEngineering/EntityTypeCodeGeneratorHelper.cs
natemcmaster: Looser coupling. Moving this to DI removes the responsibility of `Generator` to provide access to an instance of utilities class.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringConfiguration.cs
natemcmaster: Removing this simplifies the code. We don't yet support provider-specific parameters on the command line, so it seems unnecessary to add the complexity.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
natemcmaster: This is [already checked](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Commands/Program.cs#L137) by the command line tooling

src/EntityFramework.Relational.Design/CodeGeneration/CSharpCodeGeneratorHelper.cs
natemcmaster: I was planning to revert this. Making it static was the easiest way to reduce the number of dependencies `ReverseEngineeringGenerator` and `EntityTypeTemplate.cshtml` have on each other.

natemcmaster:
As discussed in person wtih @lajones, we will have an API/design meeting to review whether to make incremental changes such as this PR or to do a bigger overhaul of reverse engineering.

natemcmaster:
Updated this PR after discussing changes with @lajones. 

Providers beware - this changes provider APIs for reveng.

src/EntityFramework.Relational.Design/EntityFramework.Relational.Design.csproj
```diff
@@ -46,8 +46,6 @@
     <Compile Include="..\Shared\LoggingExtensions.cs">
       <Link>LoggingExtensions.cs</Link>
     </Compile>
-    <Compile Include="CodeGeneration\CSharpCodeGeneratorHelper.cs" />
-    <Compile Include="CodeGeneration\CSharpUtilities.cs" />
```
lajones: Did you mean to remove CSharpUtilities as well? In DesignTimeMetadataProviderFactory you add an instance of CSharpUtilities as a service.

src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGeneratorHelper.cs
```diff
@@ -18,15 +19,19 @@ public abstract class DbContextCodeGeneratorHelper
         private const string DefaultDbContextName = "ModelContext";
         protected static readonly KeyDiscoveryConvention _keyDiscoveryConvention = new KeyDiscoveryConvention();
         protected static readonly KeyConvention _keyConvention = new KeyConvention();
+        private readonly ModelUtilities _modelUtilities;
```
lajones: Protected - so derived classes can reference?

src/EntityFramework.Relational.Design/ReverseEngineering/EntityTypeCodeGeneratorHelper.cs
```diff
@@ -6,43 +6,34 @@
...
+using Microsoft.Data.Entity.Relational.Design.Utilities;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
 {
     public class EntityTypeCodeGeneratorHelper
     {
-        public EntityTypeCodeGeneratorHelper([NotNull] EntityTypeGeneratorModel generatorModel,
-            [NotNull] IRelationalMetadataExtensionProvider extensionsProvider)
+        private readonly ModelUtilities _modelUtilities;
```
lajones: As above.

src/EntityFramework.Relational.Design/ReverseEngineering/MetadataModelNameMapper.cs
```diff
@@ -6,7 +6,7 @@
 using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Metadata;
-using Microsoft.Data.Entity.Relational.Design.CodeGeneration;
+using Microsoft.Data.Entity.Relational.Design.Templating;
 using Microsoft.Data.Entity.Utilities;
```
lajones: What did we move from CodeGeneration to Templating that we need to do this?

src/EntityFramework.Relational.Design/EntityFramework.Relational.Design.csproj
lajones: Oh I see - you moved CSharpUtilities. I'm not sure that CSharpUtilities is just for Templating, but I have another bug to review that - so maybe it's fine for now.

test/EntityFramework.Relational.Design.Tests/ReverseEngineering/ReverseEngineeringGeneratorTests.cs
```diff
@@ -15,50 +13,40 @@ public void Throws_exceptions_for_incorrect_configuration()
...
             };
-            Assert.Equal(Strings.ProviderRequired,
-                Assert.Throws<ArgumentException>(
-                    () => GetGenerator().GenerateAsync(configuration).GetAwaiter().GetResult()).Message);
 
-            configuration.Provider = new NullDatabaseMetadataModelProvider();
             Assert.Equal(Strings.ConnectionStringRequired,
                 Assert.Throws<ArgumentException>(
-                    () => GetGenerator().GenerateAsync(configuration).GetAwaiter().GetResult()).Message);
+                    () => configuration.CheckValidity()).Message);
```
lajones: This is much better. Thanks.

src/EntityFramework.Relational.Design/ReverseEngineering/DbContextCodeGeneratorHelper.cs
natemcmaster: Yup. Fixed this.

lajones:
Looks good to me. But I'd like @ErikEJ to take a look as well - to see how it affects his provider. In particular he'll have to provide an implementation of CodeGeneratorHelperFactory.

natemcmaster:
Ok. @ErikEJ I will squash and rebase before merging later this afternoon unless you see major problems in these changes.

ErikEJ:
LGTM :shipit: - but unable to give it a try currently due to https://github.com/aspnet/EntityFramework/issues/2889

natemcmaster:
Heads up @ErikEJ  (and other provider writers), the CodeGeneratorHelperFactory is mostly a temporary fix for DI, and will likely change soon as we considering larger changes to code gen.

PR #2859, opened 2015-08-17
Migrations: Compare values (not references) when diffing Property.DefaultValue

### LINKED ISSUE #2802: SqlServer: Error when applying a second migration - Issue with default values on bit columns
There's 2 bugs going on here...

Given that I have
-- done an initial migration
-- applied the migration to the database
-- That I have made no code changes between the initial and second migration creation, hence I should have ZERO changes to apply (bug 1)
-- Have made a second migration
-- Attempted to apply it.(bug 2)

At this point, the update fails on all of the bit columns in the database (that are defined initially with a DEFAULT value and again no schema changes were made in the code first so there was no reason to generate the statement)

I receive the following error
System.Data.SqlClient.SqlException (0x80131904): Incorrect syntax near the keyword 'DEFAULT'.

Looking at the .cs file created by the second migration an example column is defined this way

     migration.AlterColumn(
                name: "Active",
                table: "ClueRelationTypes",
                type: "bit",
                nullable: false,
                defaultValue: true);

If I output the scripts, for the same column, I see this

     ALTER TABLE [ClueRelationTypes] ALTER COLUMN [Active] bit NOT NULL DEFAULT 1;

SqlServer doesn't allow specifying a default value in an ALTER TABLE statement, so hence it fails.

Let me know if you need any further information to reproduce the two issues.

Christine

bricelam:
Fixes #2802

### DIFF
--- a/src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
+++ b/src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
@@ -423,7 +423,7 @@ namespace Microsoft.Data.Entity.Migrations.Internal
             if (isNullableChanged
                 || columnTypeChanged
                 || sourceAnnotations.GeneratedValueSql != targetAnnotations.GeneratedValueSql
-                || sourceAnnotations.DefaultValue != targetAnnotations.DefaultValue
+                || !Equals(sourceAnnotations.DefaultValue, targetAnnotations.DefaultValue)
                 || HasDifferences(MigrationsAnnotations.For(source), targetMigrationsAnnotations))
             {
                 var isDestructiveChange = (isNullableChanged && isSourceColumnNullable)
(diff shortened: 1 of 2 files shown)


natemcmaster:
:ship: :it:

src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -423,7 +423,7 @@ protected virtual IEnumerable<MigrationOperation> Diff(IProperty source, IProper
             if (isNullableChanged
                 || columnTypeChanged
                 || sourceAnnotations.GeneratedValueSql != targetAnnotations.GeneratedValueSql
-                || sourceAnnotations.DefaultValue != targetAnnotations.DefaultValue
+                || !object.Equals(sourceAnnotations.DefaultValue, targetAnnotations.DefaultValue)
```
anpete: object is redundant

anpete:
:shipit:

src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
bricelam: :smile: I know, but only because *everything*\* is an object... Will change; it just feels weird.

PR #2868, opened 2015-08-18
SQL Server: Create unique indices using WHERE..IS NOT NULL

### LINKED ISSUE #1711: SqlServer: Generate indexes using WHERE NOT NULL
Per @divega's [comment](https://github.com/aspnet/EntityFramework/pull/1610#discussion_r25408282), this would make it behavior more consistent with other relational databases. We may also have to do work to enable this in the runtime.

bricelam:
This is to support 0..1:0..1 relationships.

Resolves #1711

### DIFF
--- a/src/EntityFramework.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
+++ b/src/EntityFramework.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
@@ -186,6 +186,28 @@ namespace Microsoft.Data.Entity.Migrations
             }
         }
 
+        protected override void Generate([NotNull] CreateIndexOperation operation, [CanBeNull] IModel model, [NotNull] SqlBatchBuilder builder)
+        {
+            base.Generate(operation, model, builder);
+
+            var clustered = operation[SqlServerAnnotationNames.Prefix + SqlServerAnnotationNames.Clustered] as bool?;
+            if (operation.IsUnique && clustered != true)
+            {
+                builder.Append(" WHERE ");
+                for (var i = 0; i < operation.Columns.Length; i++)
+                {
+                    if (i != 0)
+                    {
+                        builder.Append(" AND ");
+                    }
+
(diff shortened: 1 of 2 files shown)


ajcvickers:
:shipit:

BradBarnich:
Is there a preferred way to opt-out of this? 

Trying to port an application where we depend on the existing Sql Server behavior to prevent duplicates.

bricelam:
You should be able to create a unique constraint (aka alternate key) instead of a unique index to accomplish that.

BradBarnich:
That is actually how I started, but that sets the columns to be not null. See #4415

bricelam:
You can manually edit the generated migration and replace `CreateIndex` with a hand-written `Sql` call containing the `CREATE INDEX` statement you need.

bricelam:
But wait, you're asking for a nullable column where only one row can contain null for that column?

BradBarnich:
Yes, that it the behavior of unique index in SQL Server. 

for a table (A not null, B null), unique index on (A,B):
I can store (1, null), (1,1), (1,2) and trying to store (1,null) again will fail. 

This app expects that behavior, which I understand is different from other database products. 

Manually modifying the migration is a fine workaround

bricelam:
Ah, I see it makes more sense for composite indexes. :smiling:

marchy:
WORKAROUND: For those seeking a workaround, we ended up adding another 'UniqueunessKey' column which serialized a value containing all the other columns (ie: simple string concatenation with a '|' separator)."

This way all nullable/non-nullable column values would be considered as part of the uniqueness semantics and map to a single column that a unique index can be defined on, while avoiding the multiple column and nullable values weirdness that either the database or EF Core exhibits.

**Example:**
INSERT: (foo: "A", bar: null) ==> ALLOWED
followed by
INSERT: (foo: "A", bar: null) ==> UNIQUE INDEX VIOLATION (as expected)

Hope this helps!

PR #2878, opened 2015-08-19
Migrations: Restart sequence when start value changes

### LINKED ISSUE #2811: Migrations: Scaffold RestartSequence when StartValue changes
This came out of API Review. We decided that changing it after it has been created has very undefined semantics. The one scenario we thought of that this would enable is changing it after it's been created, but before it's been used.

bricelam:
Resolves #2811


lajones:
:shipit:

PR #2880, opened 2015-08-19
[TINY] Adjust to change in ConfigurationSection.Key

rowanmiller:
ConfigurationSection.Key now returns the full key rather than just the
key within the section.

### DIFF
--- a/test/EntityFramework.Microbenchmarks.Core/BenchmarkConfig.cs
+++ b/test/EntityFramework.Microbenchmarks.Core/BenchmarkConfig.cs
@@ -22,7 +22,7 @@ namespace EntityFramework.Microbenchmarks.Core
             return new BenchmarkConfig
             {
                 RunIterations = bool.Parse(config["benchmarks:runIterations"]),
-                ResultDatabases = resultDatabasesSection.GetChildren().Select(s => resultDatabasesSection[s.Key]).ToArray(),
+                ResultDatabases = resultDatabasesSection.GetChildren().Select(s => config[s.Key]).ToArray(),
                 BenchmarkDatabaseInstance = config["benchmarks:benchmarkDatabaseInstance"],
                 ProductReportingVersion = config["benchmarks:productReportingVersion"],
                 CustomData = config["benchmarks:customData"]


ajcvickers:
:shipit:

rowanmiller:
cc @HaoK @divega just to make sure this change was intentional (no pushback... just checking :smile:)

HaoK:
@kirthik This was the regression I was talking about I guess... is this intentional?

HaoK:
I'll pushback for you @rowanmiller this looks like a regression, I'd think the key should be relative to the section, not the full path from the root.  That's just weird

divega:
cc @kirthik, who made the change and touched this line previously. I think this is kind of intentional, but also something we could discuss:
- The Key property was added to the IConfigurationSection interface with the intention of providing metadata about the configuration node corresponding to the instance. From that perspective the new behavior provides de most information. With the previous behavior it would have been cumbersome to obtain this information.
- Personally I wasn't even aware that there was a Key property hanging there previously.presumably it was not in the interface but in the class?  From that perspective this is a subtle breaking change that even @kirthik missed she had to react to when she previously adjusted this code. It has now became cumbersome to obtain jus the last key segment if that is what you need.
In theory we could expose both, e.g. `Key` and `FullKey`. Not sure is a good idea as it makes the API more complicated for everyone. We should really look and the scenarios... Regardless, it seems that at the minimum we should go back at the places where ConfigurationSection.Key was being used across the stack and make sure we don't leave them broken.

HaoK:
What about just splitting it into two properties like:
ParentPath and Key, so basically you can just concat the two to get the current full path behavior, and Key is untouched and remains the last segment.

divega:
We could. We should talk about this. Can you file a bug in Configuration?

HaoK:
https://github.com/aspnet/Configuration/issues/279

test/EntityFramework.Microbenchmarks.Core/BenchmarkConfig.cs
```diff
@@ -22,7 +22,7 @@ public class BenchmarkConfig
             return new BenchmarkConfig
             {
                 RunIterations = bool.Parse(config["benchmarks:runIterations"]),
-                ResultDatabases = resultDatabasesSection.GetChildren().Select(s => resultDatabasesSection[s.Key]).ToArray(),
+                ResultDatabases = resultDatabasesSection.GetChildren().Select(s => config[s.Key]).ToArray(),
```
divega: @rowanmiller I forgot to mention you could have just done `s => s.Value` with the new API here. Most likely not worth changing for now.

rowanmiller: Cool, I was going to ask why I couldn't do that... but good to hear I now can :smile:

PR #2884, opened 2015-08-19
Avoid using navigation properties for fixup when FK is source of truth

ajcvickers:
When an entity is queried or attached, the value of the FK property defines the relationship, which means there is no need to scan navigations in order to try to do fixup, thereby improving perf.


anpete:
:shipit:

PR #2885, opened 2015-08-19
Pull Migrations use up into RelationalDatabaseCreator

### LINKED ISSUE #2835: Push Migrations use down into RelationalDatabaseCreator
Both the SQLite and SQL Server providers have almost identical code to create the database using Migrations components. We should push this down into the `RelationalDatabaseCreator` class.

bricelam:
Resolves #2835


ajcvickers:
:shipit:

PR #2901, opened 2015-08-21
Block databinding with IListSource

### LINKED ISSUE #1431: Consider IListSource on DbSet/Query
In EF6 we implemented IListSource to throw and point folks to the right way to data bind. We should consider doing this again.

smitpatel:
Resolves #1431

### DIFF
--- a/src/EntityFramework.Core/DbSet`.cs
+++ b/src/EntityFramework.Core/DbSet`.cs
@@ -4,11 +4,17 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+#if NET45 || DNX451
+using System.ComponentModel;
+#endif
 using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.ChangeTracking;
 using Microsoft.Data.Entity.Infrastructure;
+#if NET45 || DNX451
+using Microsoft.Data.Entity.Internal;
+#endif
 using Microsoft.Data.Entity.Query;
 
 namespace Microsoft.Data.Entity
@@ -31,7 +37,7 @@ namespace Microsoft.Data.Entity
     ///     </para>
     /// </summary>
     /// <typeparam name="TEntity"> The type of entity being operated on by this set. </typeparam>
-    public abstract class DbSet<TEntity>
+    public abstract partial class DbSet<TEntity>
         : IOrderedQueryable<TEntity>, IAsyncEnumerableAccessor<TEntity>, IAccessor<IServiceProvider>
         where TEntity : class
     {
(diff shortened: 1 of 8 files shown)


src/EntityFramework.Core/DbSet`.cs
```diff
@@ -4,11 +4,13 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.ComponentModel;
```
divega: Unless I am missing a detail new usings should wrapped in an #if. Otherwise they will tend to be removed by cleanup.

src/EntityFramework.Core/DbSet`.cs
```diff
@@ -276,5 +281,14 @@ IQueryProvider IQueryable.Provider
...
         }
+
+#if NET45 || DNX451
+        IList IListSource.GetList()
+        {
+            throw new NotSupportedException(Strings.DataBindingWithIListSource);
+        }
+
+        bool IListSource.ContainsListCollection => false;
+#endif
```
bricelam: It may be a bit cleaner to factor this into it's own partial.
```C#
#if NET45 || DNX451
partial class DbSet<TEntity> : IListSource
{
        IList IListSource.GetList()
        {
            throw new NotSupportedException(Strings.DataBindingWithIListSource);
        }

        bool IListSource.ContainsListCollection => false;
}
#endif
```

divega: Nice. We can put it in a different file with its own using statements.

test/EntityFramework.Core.Tests/EntityFramework.Core.Tests.csproj
```diff
@@ -18,7 +18,7 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug\</OutputPath>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;NET46;DNX451</DefineConstants>
```
bricelam: Don't add `DNX451` here.

I've found the cleanest way to do this is to use `#if !DNXCORE50` in the tests. That way `NET46` isn't needed here either.

src/EntityFramework.Core/DbSet`.cs
smitpatel: :+1:

test/EntityFramework.Core.Tests/EntityFramework.Core.Tests.csproj
smitpatel: targets for Core.Tests => net46, dnx451. Do we need if for the test?

src/EntityFramework.Core/DbSet`.cs
smitpatel: Suggestion for file name? :confused:

src/EntityFramework.Core/EntityFramework.Core.csproj
```diff
@@ -19,7 +19,7 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>bin\Debug\</OutputPath>
-    <DefineConstants>TRACE;DEBUG;CSPROJ</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;CSPROJ;NET45;DNX451</DefineConstants>
```
bricelam: No `DNX` in the `*.csproj` files, please.

bricelam:
:shipit:

PR #2902, opened 2015-08-21
Migrations: Use an attribute for migration ID

### LINKED ISSUE #2832: Migrations: Change Migration.Id to an Attribute
This came up in API Review. We might be able to skip instantiating the classes in a few places. We'll most likely continue to need the property (or an extension method) however since we often need to get the ID for an instance.

bricelam:
Also avoids instantiating `Migration` classes unnecessarily.

Resolves #2832


ajcvickers:
:shipit:

PR #2903, opened 2015-08-21
Improve perf for snapshotting original values

ajcvickers:
In the common case where original values are tracked eagerly for all properties this change takes the query value buffer and uses it directly as the original values store. This essentially eliminates the cost of creating the original values snapshot, although snapshot of relationship state still takes time.

In doing this work it became apparent that sometimes values in the ValueBuffer are not of the correct type--notably for enum values. This caused problems for DetectChanges since the values were of different types and therefore were flagged as different even when no change had been made. To fix this the ValueBuffer obtained from GetValues is now post-processed to convert values as necessary. Currently this is only for enums, but in the future it could be used for any type conversion. Since values in the ValueBufferare now always of the correct type even when GetValues is used it means that the conversion code in the materializer can be removed.

### DIFF
--- a/src/EntityFramework.Relational/Storage/UntypedValueBufferFactoryFactory.cs
+++ b/src/EntityFramework.Relational/Storage/UntypedValueBufferFactoryFactory.cs
@@ -3,15 +3,101 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using Microsoft.Data.Entity.Internal;
 
 namespace Microsoft.Data.Entity.Storage
 {
     public class UntypedValueBufferFactoryFactory : IRelationalValueBufferFactoryFactory
     {
+        private struct CacheKey
+        {
+            public CacheKey(IReadOnlyList<Type> valueTypes)
+            {
+                ValueTypes = valueTypes;
+            }
+
+            public IReadOnlyList<Type> ValueTypes { get; }
+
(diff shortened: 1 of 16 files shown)


src/EntityFramework.Relational/Storage/UntypedValueBufferFactoryFactory.cs
```diff
@@ -3,15 +3,101 @@
...
+                                            valueVariable,
+                                            type.UnwrapEnumType()),
+                                        type),
+                                    typeof(object)))));
+                }
+            }
+
+            if (conversions.Count == 0)
+            {
+                return _ => { };
```
anpete: Cache this?

anpete:
:shipit:

src/EntityFramework.Relational/Storage/UntypedValueBufferFactoryFactory.cs
ajcvickers: The delegate is cached. Not sure if you are meaning to cache something else?

anpete: Meant this: ```return _ => { };``` Not sure if it matters.

ajcvickers: Changed it to return null instead to avoid creating an executing any delegate in the common case that there are no conversions needed.

PR #2904, opened 2015-08-21
Support for database schema on model builder

### LINKED ISSUE #2875: Support for Database Schema on the Model Builder
EF 6 had a support for a database schema in the model builder:
```c#
modelBuilder.HasDefaultSchema(“sales”);
```

This is a really handy feature and somewhat missing in EF7. 

Could this be introduced in future?
Are there any plans for this?

smitpatel:
resolves #2875

### DIFF
--- a/src/EntityFramework.Relational/Metadata/IRelationalModelAnnotations.cs
+++ b/src/EntityFramework.Relational/Metadata/IRelationalModelAnnotations.cs
@@ -11,5 +11,7 @@ namespace Microsoft.Data.Entity.Metadata
         ISequence FindSequence([NotNull] string name, [CanBeNull] string schema = null);
 
         IReadOnlyList<ISequence> Sequences { get; }
+
+        string DefaultSchema { get; }
     }
 }
--- a/src/EntityFramework.Relational/Metadata/RelationalModelAnnotations.cs
+++ b/src/EntityFramework.Relational/Metadata/RelationalModelAnnotations.cs
@@ -50,5 +50,14 @@ namespace Microsoft.Data.Entity.Metadata
                 Annotations.ProviderPrefix ?? RelationalAnnotationNames.Prefix,
                 Check.NotEmpty(name, nameof(name)),
                 Check.NullButNotEmpty(schema, nameof(schema)));
+
+        public virtual string DefaultSchema
+        {
+            get
+            {
+                return (string)Annotations.GetAnnotation(RelationalAnnotationNames.DefaultSchema);
+            }
+            [param: CanBeNull] set { Annotations.SetAnnotation(RelationalAnnotationNames.DefaultSchema, Check.NullButNotEmpty(value, nameof(value))); }
+        }
     }
 }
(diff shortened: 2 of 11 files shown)


src/EntityFramework.Relational/Metadata/IRelationalModelAnnotations.cs
```diff
@@ -11,5 +11,7 @@ public interface IRelationalModelAnnotations
         ISequence FindSequence([NotNull] string name, [CanBeNull] string schema = null);
 
         IReadOnlyList<ISequence> Sequences { get; }
+
+        string Schema { get; }
```
divega: I have slight preference for calling it `DefaultSchema` here as it only applies to objects that don't have it specified explicitly. Thoughts?

rowanmiller: :+1:

src/EntityFramework.Relational/Metadata/RelationalModelAnnotations.cs
```diff
@@ -41,14 +41,26 @@ public virtual IReadOnlyList<ISequence> Sequences
...
+        public virtual string Schema
+        {
+            get
+            {
+                return (string)Annotations.GetAnnotation(RelationalAnnotationNames.Schema);
+            }
+            [param: CanBeNull] set { SetSchema(value); }
+        }
+
+        protected virtual bool SetSchema(string value)
```
ajcvickers: Why is this broken out into a separate protected method?

src/EntityFramework.Relational/Metadata/RelationalModelAnnotations.cs
```diff
@@ -41,14 +41,26 @@ public virtual IReadOnlyList<ISequence> Sequences
...
+            => (Annotations.ProviderPrefix == null ? null : Sequence.FindSequence(Model, Annotations.ProviderPrefix, name, schema ?? Schema))
+               ?? Sequence.FindSequence(Model, RelationalAnnotationNames.Prefix, name, schema ?? Schema);
 
         public virtual Sequence GetOrAddSequence([CanBeNull] string name, [CanBeNull] string schema = null)
             => new Sequence(
                 (Model)Model,
                 Annotations.ProviderPrefix ?? RelationalAnnotationNames.Prefix,
                 Check.NotEmpty(name, nameof(name)),
-                Check.NullButNotEmpty(schema, nameof(schema)));
+                Check.NullButNotEmpty(schema, nameof(schema)) ?? Schema);
```
ajcvickers: If I call GetOrAddSequence without specifying a schema, then set the model schema, then make the same call again, I will get a different sequence back the second time. Is this what we want to happen? More generally, what does setting the model schema really mean when it comes to entity types, indexes, sequences, etc in the model? Does it mean if these things don't have a schema specified, then they will use the model schema? Or does it mean that when they are created they should be created with the model schema? I think it should probably be the former, in which case it looks like this Sequence code might need to change.

divega: I agree. What we are trying to accomplish is the former.

rowanmiller: :+1: it should be anything that doesn't have a schema specified when the model is completely built

smitpatel: Sequence will store the value of schema if specified while creation. `FindSequence` will work in the same way so that we always find the same sequence again even if the value of `DefaultSchema` changed. `Schema` property inside `Sequence` object will fallback to model level default schema if the schema for sequence is not specified.

smitpatel:
Updated.

ajcvickers:
:shipit:

PR #2913, opened 2015-08-24
[SMALL] Tidy up variations on perf tests

rowanmiller:
* Adjust names of Update Pipeline variations so that the comparisons
better reflect what will happen when you follow best practices (i.e.
there was no batching on EF6 and by default it is on for EF7, so
comparing those numbers gives you the typical perf for each version).
* Standardize on having a DetectChangesOn/Off variation for all DbSet
Operation tests (even for the scenarios where it theoretically makes
little difference).


ajcvickers:
:shipit:

PR #2921, opened 2015-08-25
Migrations: Don't batch commands by default

### LINKED ISSUE #2739: Migrations: Remove batching
We didn't have it in EF6. Its value-add seems minimal. It isn't supported by SQL Server Compact. We'd have to add API to `MigrationBuilder` for ending batches. Etc. We should scope it out for now.

bricelam:
Resolves #2739


ajcvickers:
:shipit:

PR #2922, opened 2015-08-25
Update compile-tests target to run during compile

### LINKED ISSUE #2889: Provider test packages no longer available
There now appear to be four dfiierent ASP.NET vNext galleries on MyGet:
1: https://www.myget.org/gallery/aspnetlitedev
2: https://www.myget.org/gallery/aspnetvnext

3: https://www.myget.org/gallery/aspnetliterelease
4: https://www.myget.org/gallery/aspnetrelease

I would like to update my product code with the most recent EF beta 7 build and run the test suite against the updated product code. This appears to be impossible at the moment, as current provider test packages do not seem to be available anywhere.

I assume that feed 1 is the one to be used going forward towards beta8 ? In that case, that feed is missing the three ".FunctionalTests " packages

I assume that feed 3 is the latest beta 7 bits, and that is where I should move to (build 15151 or later) in preparation for beta 7 - in that case, that feed is also missing the three ".FunctionalTests " packages.

I assume that feed feeds 2 and 4 will be decommisioned soon? Feed 2 seems to be in a bad state, with a mix in versions for EF product and test packages.

bricelam:
Fixes #2889

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -8,7 +8,7 @@ var AUTHORS='Microsoft Open Technologies, Inc.'
 use-standard-lifecycle
 k-standard-goals
 
-#compile-tests target='test-compile'
+#compile-tests target='compile'
   @{
     var BASE_DIR = Directory.GetCurrentDirectory();
     var TARGET_DIR = Path.Combine(BASE_DIR, "artifacts");


natemcmaster:
:shipit:

PR #2923, opened 2015-08-25
[small] What's in a nameof?

natemcmaster:
Use nameof for refactor-safe code.
Also fixup on more DI instance


src/EntityFramework.Sqlite.Design/EntityFramework.Sqlite.Design.csproj
```diff
@@ -68,9 +68,7 @@
     </Content>
     <None Include="project.json" />
   </ItemGroup>
-  <ItemGroup>
-    <Folder Include="ReverseEngineering\Templates\" />
-  </ItemGroup>
+  <ItemGroup />
```
lajones: Nit: You can just get rid of the whole <ItemGroup/>

lajones:
:shipit:

Issue #2898, opened 2015-08-21
Test: Azure SQL Database


divega:
Our functional tests run every time against LocalDB. In the recent past we have found issues when running against Azure SQL Database only through manual testing. We should figure out a way (through automation or not) to improve our coverage of this scenario.

natemcmaster:
Getting successful runs of SqlServer.FunctionalTests against SQL Azure will require infrastructure changes to our testing. 

On my most recent attempt, 1624 of 1649 tests fail, mostly due to connection timeouts, replication issues, and throttling. Most issues appear to stem from our use of >75 separate databases and our limited options. 

Tests using scratch databases will always fail until we add something to fix #2969 that because Azure immediately begins geo-replication on all standard 'S0' databases.

natemcmaster:
Status update: running *all*  functional tests against SQL Azure is not likely to work. I can get a very small subset passing, but anything that tests that uses database creation/deletion fails (which is basically all right now).

natemcmaster:
I've investigated further. I would not recommend running our full suite of SQL Server functional tests against Azure as I believe they will frequently fail without providing useful information about actual problems with SQL Azure. Instead, I would favor automated a subset of tests that focus primarily on the aspects of Azure that are different. These include:

1. Remote connectivity. Our SQL Server testing runs on the same machine or in the same subnet. SQL Azure tests will ensure SQL Client /EF are resilient enough to handle WAN connections
1. Throughput and stress. Related to 1.
1. Eventual database creation and deletion. `EnsureCreated()` and `EnsureDeleted()` fail because database creation operations on SQL Azure are subject to heavy delay. Also would allow functional testing for #2969
1. Slight differences in DDL support

I think it is also worth noting that SQL Azure v12 improves parity with SQL Server 2014 but still has [some differences.](https://azure.microsoft.com/en-us/documentation/articles/sql-database-transact-sql-information/) These differences may affect EF and should be investigated further. For instance, we need to ensure migrations creates *clustered* indexes.

> Because Azure SQL Database does not support heap tables, a table must have a clustered index. If a table is created without a clustered constraint, a clustered index must be created before an insert operation is allowed on the table.

divega:
@natemcmaster thanks a lot for investigating further. Am I understanding correctly that besides the lack of automatic retries on transient failures EF7 generally works with Azure SQL DB?

BTW, clustered indexes are created on the PK by default, so that shouldn't be a big concern.

[...]

Issue #2897, opened 2015-08-21
Test: SQL Server 2008 coverage


divega:
Our functional tests usually run on recent versions of LocalDB. We want to support older versions of SQL Server including SQL Server 2008 so we should figure out a way (through automation or not) to have good coverage for it.

natemcmaster:
Ran an initial test. Our functional tests make extensive use of SEQUENCE which is not supported in 2008.

ErikEJ:
Also OFFSET and FETCH will cause 2008 to break

natemcmaster:
Current results against SQL Server 2008 SP 1:

299 of 1645 tests in Ef.SqlServer.FunctionalTests fail.

13 fail with
> System.Data.SqlClient.SqlException : Incorrect syntax near 'OFFSET'.

21 fail with
> System.Data.SqlClient.SqlException : Incorrect syntax near 'OFFSET'. Invalid usage of the option NEXT in the FETCH statement.

262 fail with
> System.Data.SqlClient.SqlException : Unknown object type 'SEQUENCE' used in a CREATE, DROP, or ALTER statement.

3 fail with
> System.Data.SqlClient.SqlException : Unknown object type 'SEQUENCE' used in a CREATE, DROP, or ALTER statement. The name "next" is not permitted in this context. Valid expressions are constants, constant expressions, and (in some contexts) variables. Column names are not permitted.

rowanmiller:
Given we have a design principle to optimize for the latest releases and be ok with needing to do additional configuration for older versions of SQL Server we probably need to do additional config on the test model when running on SQL2008.

ajcvickers:
Also there are some tests that just won't work with older versions--for example, anything testing sequences is not going to work. We should have a way of excluding these.

anpete:
Rather than exclude, we should do what @rowanmiller says and figure out how to run them with different config (where it makes sense).

natemcmaster:
Excluding tests is easy. But what about OFFSET? Should we create a SQL2008 query generator that has a workaround?

anpete:
@natemcmaster We are already tracking this. #2180

divega:
Talked to @ajcvickers and we arrived to the conclusion that we have three classes of tests that use SEQUENCE:

1. _Our SQL Server SEQUENCE support tests:_ these will never work with SQL Server 2008 so could be skipped or ignored 
[...]

PR #2931, opened 2015-08-26
Configurable tests

natemcmaster:
Make SqlServer testing configurable. 
~~Add makefile target for running a test matrix.~~

Required to test against more database platforms  #2897 and #2898.
Also provides a way to automate these tests.


makefile.shade
```diff
@@ -105,6 +107,80 @@ var sqliteTestDir='test/EntityFramework.Sqlite.FunctionalTests/bin/${E("Configur
...
+        foreach(var test in tests.Keys)
+        {
+            Log.Info(string.Format("Running test '{0}'", test));
+            var testSettings = GetSection(tests, test);
+
+            var cmd = GetValue(testSettings, "cmd") as string;
+
+            var kCmd = GetValue(testSettings, "dnx") as string;
+            var folder = GetValue(testSettings, "folder") as string;
+            var dnvmUse = GetValue(testSettings,"dnvmUse") as string;
```
smitpatel: space after ,

makefile.shade
```diff
@@ -105,6 +107,80 @@ var sqliteTestDir='test/EntityFramework.Sqlite.FunctionalTests/bin/${E("Configur
         }
     }
 
+var matrixTestConfig='matrixTests.json'
+
+#test-matrix
```
bricelam: We can probably capture this in CI configurations instead of the makefile.

natemcmaster:
Ping!

bricelam:
Looks fine to me. You should get input from a few others before merging.

natemcmaster:
@rowanmiller this is most similar to what you did with benchmarks config. Thoughts?

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        public static SqlServerTestConfig Default()
+        {
+            var config = new ConfigurationBuilder(".")
+                .AddJsonFile("config.json", optional: true)
+                .AddJsonFile("config.test.json", optional: true)
+                .AddEnvironmentVariables()
+                .Build();
+
```
rowanmiller: Should we be doing some caching here? In the benchmarks I had a static Instance property so that we only ever ran all this logic once.

test/EntityFramework.SqlServer.FunctionalTests/README.md
```diff
@@ -0,0 +1,28 @@
+EntityFramework.SqlServer.FunctionalTests
```
rowanmiller: I think it's fine to leave this in... but I'm not sure how much value it adds. I think it's pretty obvious how to use the config file and this will probably just get out of date.

rowanmiller:
:shipit: from my perspective. Maybe others will have pushback on the static-ness of the config... but I think it makes sense.

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
natemcmaster: I don't cache because some tests modify the ConnectionStringBuilder to add custom properties.

test/EntityFramework.SqlServer.FunctionalTests/README.md
natemcmaster: K good. Then i'm gonna remove to reduce update-burden.

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
rowanmiller: I wonder if we should cache the base one and then have With return a new instance of the builder

rowanmiller: I'm only thinking of perf though... if it is fast enough then it doesn't matter :smile:

natemcmaster: The slowest part is probably reading config from file. I can easily cache that part.

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
```diff
@@ -0,0 +1,95 @@
...
+            int connectTimeout;
+            if (!int.TryParse(config[prefix + "ConnectTimeout"], out connectTimeout))
```
divega: Nit: Can you make this a bit cleaner by store the configuration section for the prefix in a variable? E.g.:
```C#
var section = config.GetSection(prefix);
// ...
if (!int.TryParse(section["ConnectTimeout"], out connectTimeout)) 
```

natemcmaster:
Responded to feedback from @divega  and @rowanmiller

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
```diff
@@ -0,0 +1,95 @@
...
+            return this;
+        }
+
+        public SqlServerTestConfig With(Action<SqlConnectionStringBuilder> connStrBuilder)
+        {
+            connStrBuilder(ConnectionStringBuilder);
+            return this;
+        }
+
+        public SqlConnectionStringBuilder ConnectionStringBuilder
```
divega: Is anyone using this? Maybe make pvt?

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
divega: > I don't cache because some tests modify the ConnectionStringBuilder to add custom properties.

But aren't you deferring the creation of the ConnectionStringBuilder until later?

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
```diff
@@ -0,0 +1,99 @@
...
+            var testConfig = new SqlServerTestConfig
+            {
+                DataSource = config["DataSource"] ?? @"(localdb)\MSSQLLocalDB",
+                IntegratedSecurity = integratedSecurity,
+                ConnectTimeout = connectTimeout,
+                UserID = config["UserID"],
+                Password = config["Password"]
+            };
+            Debug.WriteLine($"Using '{testConfig.ConnectionString}'");
+            return testConfig;
```
divega: Still think @rowanmiller is right and you can cache this because anyway you haven't created the string builder that others will modify. Of course, if this wasn't test code, I would also recommend making it immutable :smile:

divega: Oh, it already is immutble, isn't it?

divega: Ah, I think I see what the difficulty is: the connection string builder you keep around is mutable and tests can change that. I would recommend removing the field and creating the connection string builder every time. It would require removing the version of With() that takes a key and an object (it is redundant anyway) and changing the other With() to return the connection string directly (and maybe rename it?)

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqlServerFixture.cs
```diff
@@ -16,13 +15,9 @@ protected override void ConfigureServices(IServiceCollection services)
...
         {
-            optionsBuilder.UseSqlServer("" +
-                new SqlConnectionStringBuilder
-                {
-                    DataSource = @"(localdb)\MSSQLLocalDB",
-                    InitialCatalog = nameof(MigrationsSqlServerTest),
-                    IntegratedSecurity = true
-                });
+            optionsBuilder.UseSqlServer(SqlServerTestConfig.Default()
+                .With(b=> { b.InitialCatalog = nameof(MigrationsSqlServerTest); })
```
anpete: With what? Assuming this is connection string config? NIT: Formatting and braces not required.

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            connStrBuilder(ConnectionStringBuilder);
+            return this;
+        }
+
+        public SqlConnectionStringBuilder ConnectionStringBuilder
+        {
+            get
+            {
+                if (_connectionStringBuilder != null)
```
anpete: Use blank lines between blocks and members.

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqlServerFixture.cs
anpete: Can this test be standardized away from OnConfiguring?

test/EntityFramework.SqlServer.FunctionalTests/MonsterFixupSqlServerTest.cs
```diff
@@ -45,16 +44,13 @@ protected override DbContextOptions CreateOptions(string databaseName)
...
-            return new SqlConnectionStringBuilder
-            {
-                DataSource = @"(localdb)\MSSQLLocalDB",
-                MultipleActiveResultSets = true,
-                InitialCatalog = name,
-                IntegratedSecurity = true,
-                ConnectTimeout = 30
-            }.ConnectionString;
-        }
+            => SqlServerTestConfig.Default()
```
anpete: Can this test follow the standard fixture pattern?

test/EntityFramework.SqlServer.FunctionalTests/NavigationTest.cs
```diff
@@ -90,7 +90,15 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
         protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
         {
-            optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=StateManagerBug;Trusted_Connection=True;MultipleActiveResultSets=true");
+            var config = SqlServerTestConfig.Default()
```
anpete: Can this test follow the standard fixture pattern?

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestConfig.cs
```diff
@@ -0,0 +1,99 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Data.SqlClient;
+using Microsoft.Framework.Configuration;
+using System.Diagnostics;
+using System;
+
+namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
+{
+    public class SqlServerTestConfig
```
divega: Half of this class looks like a DTO for a connection string builder, and I am not sure that is needed. E.g. we could just use connection string builders, an extension method to populate it from configuration, and a default populated instance we cache (and we promise tests will never change :smile:). But give that this is test code and that there can be different opinions I don't care too much.

natemcmaster:
@divega @anpete I will revisit my sloppy implementation of `SqlServerTestConfig`. I didn't give it much design as this is test code. But I'll improve it so test writers understand better how to use it. I'll update the PR when its ready.

natemcmaster:
Updated. Moved loading environment variables into an extension method on `SqlConnectionStringBuilder`

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlConnectionStringBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            if (overwrite || string.IsNullOrEmpty(builder.Password))
+            {
+                builder.Password = config["Password"];
+            }
+
+            return builder;
+        }
+
+        private static readonly Lazy<IConfiguration> _sqlTestConfiguration = new Lazy<IConfiguration>(() =>
```
divega: Is `Lazy<T>` necessary?

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
```diff
@@ -366,7 +366,7 @@ private SqlServerTestStore CreateTransient(bool createDatabase)
 
                     await command.ExecuteNonQueryAsync();
 
-                    var userFolder = Environment.GetEnvironmentVariable("USERPROFILE");
+                    var userFolder = Environment.GetEnvironmentVariable("USERPROFILE") ?? Environment.GetEnvironmentVariable("HOME");
```
divega: Interesting to see this code... Maybe there is a good reason we are doing it in this case, but simply deleting the files has traditionally been a broken way of dropping a SQL Server database, even on LocalDB. An usual consequence AFAIR is that next time you try to create it you'll get a failure and you will have to try opening it and closing it before SQL Server records the fact that it doesn't exist anymore.

divega: Ah, I just saw the drop above. I don't understand why we would need to delete the files, but it might be ok.

divega: One possible theory: we had some test that failed to recreate the database because the files were still there after the drop had executed, and deleting the files seemed to help. I wonder if DROP DATABASE is just a fire and forget operation. Maybe it would be more robust to just poll for the files before we return, yet it might not matter too much.

@anpete it seems that you added this code in https://github.com/natemcmaster/EntityFramework/commit/6eec03d4103a45f35122eb8907771f07a9195aae. Do you remember why?

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlConnectionStringBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            var config = _sqlTestConfiguration.Value;
+
+            int connectTimeout;
+            if (!int.TryParse(config["ConnectTimeout"], out connectTimeout))
+            {
+                connectTimeout = DefaultConnectTimeout;
+            }
+
+            builder.ConnectTimeout = connectTimeout;
```
divega: Should the overwrite flag apply to this too?

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlConnectionStringBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            bool integratedSecurity;
+            if (!bool.TryParse(config["IntegratedSecurity"], out integratedSecurity))
+            {
+                integratedSecurity = DefaultIntegratedSecurity;
+            }
+
+            builder.IntegratedSecurity = integratedSecurity;
```
divega: Overwrite too?

test/EntityFramework.SqlServer.FunctionalTests/MonsterFixupSqlServerTest.cs
```diff
@@ -46,14 +46,14 @@ protected override DbContextOptions CreateOptions(string databaseName)
...
-                ConnectTimeout = 30
-            }.ConnectionString;
+            var connStrBuilder = new SqlConnectionStringBuilder
+                {
+                    MultipleActiveResultSets = true,
+                    InitialCatalog = name
+                };
+            connStrBuilder.LoadEnvConfiguration();
+
+            return connStrBuilder.ConnectionString;
```
divega: Nit: it wouldn't bother me if you just chained it all inline when you can :smile:

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlConnectionStringBuilderExtensions.cs
divega: Moreover, do you ever not overwrite? Maybe you don't need the flag.

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlConnectionStringBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
+{
+    public static class SqlConnectionStringBuilderExtensions
+    {
+        private static string DefaultDataSource = @"(localdb)\MSSQLLocalDB";
+        private static int DefaultConnectTimeout = 30;
+        private static bool DefaultIntegratedSecurity = true;
+
+        public static SqlConnectionStringBuilder LoadEnvConfiguration(this SqlConnectionStringBuilder builder, bool overwrite = true)
```
divega: Nit: maybe just ApplyConfiguration?

divega:
LGTM with some minor feedback. Should probably get some other eyes on it.

Just one last general question: do you think that this could be helpful for testing other database providers? It seems to me that with a few small tweaks the extension method could be made to work with any DbConnectionStringBuilder.

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlConnectionStringBuilderExtensions.cs
natemcmaster: Yeah, I guess the overwrite flag isn't necessary. I'll remove it.

natemcmaster:
Certainly it could be generalized, but we don't need something like this for SQLite (yet) so I don't think it's worth doing yet.

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlConnectionStringBuilderExtensions.cs
natemcmaster: Nope. Moving to static constructor.

divega:
ok, I guess @ErikEJ or @roji can send a PR if they want it generalized :smile:

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
divega: Actually I was wrong on who added this code. @AndriySvyryd, do you remember why we needed to delete the files?

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqlServerFixture.cs
```diff
@@ -16,13 +16,13 @@ protected override void ConfigureServices(IServiceCollection services)
 
         protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
         {
-            optionsBuilder.UseSqlServer("" +
-                new SqlConnectionStringBuilder
+            var connStrBuilder = new SqlConnectionStringBuilder
```
anpete: Don't abbreviate.

divega:
@natemcmaster LGTM, but I will let @anpete do another pass and give you the shipit.

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqlServerFixture.cs
anpete: Can we unify these (and the two below) to use the standard fixture pattern?

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
AndriySvyryd: @divega Your theory is correct. Dropping the database didn't delete the files right away so the following tests would fail.

natemcmaster:
@anpete https://github.com/natemcmaster/EntityFramework/commit/6c607b03b95079514b946a346d7ea44db31a5bdb using the fixture pattern.

test/EntityFramework.Sqlite.FunctionalTests/MonsterFixupSqliteTest.cs
```diff
@@ -9,19 +9,46 @@
 using Microsoft.Data.Entity.FunctionalTests;
 using Microsoft.Data.Entity.FunctionalTests.TestModels;
 using Microsoft.Data.Entity.Infrastructure;
-using Microsoft.Data.Sqlite;
 using Microsoft.Framework.DependencyInjection;
 
 namespace Microsoft.Data.Entity.Sqlite.FunctionalTests
 {
-    public class MonsterFixupSqliteTest : MonsterFixupTestBase
+    public class MonsterFixupSqliteTest : MonsterFixupTestBase<MonsterFixupSqliteFixture>
```
anpete: Use the SqliteTestStore

test/EntityFramework.SqlServer.FunctionalTests/NavigationTest.cs
```diff
@@ -87,10 +87,35 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
-            optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=StateManagerBug;Trusted_Connection=True;MultipleActiveResultSets=true");
+            var connStrBuilder = new SqlConnectionStringBuilder
+                {
+                    InitialCatalog = "StateManagerBug",
+                    MultipleActiveResultSets = true,
+                    ["Trusted_Connection"] = true
+                };
+            connStrBuilder.ApplyConfiguration();
+
+            var optionsBuilder = new DbContextOptionsBuilder();
```
anpete: Build options in ctor and cache?

anpete: Use test store?

test/EntityFramework.SqlServer.FunctionalTests/MonsterFixupSqlServerTest.cs
```diff
@@ -14,14 +14,42 @@
...
     {
         private static readonly HashSet<string> _createdDatabases = new HashSet<string>();
 
         private static readonly ConcurrentDictionary<string, AsyncLock> _creationLocks
             = new ConcurrentDictionary<string, AsyncLock>();
 
-        protected override IServiceProvider CreateServiceProvider(bool throwingStateManager = false)
+        public override async Task CreateAndSeedDatabase(Func<MonsterContext> createContext)
+        {
+            using (var context = createContext())
```
anpete: Use test store.

test/EntityFramework.SqlServer.FunctionalTests/MonsterFixupSqlServerTest.cs
```diff
@@ -36,52 +64,79 @@ protected override IServiceProvider CreateServiceProvider(bool throwingStateMana
             return serviceCollection.BuildServiceProvider();
         }
 
-        protected override DbContextOptions CreateOptions(string databaseName)
+        private DbContextOptions CreateOptions(string databaseName)
         {
+            var connectionStringBuilder = new SqlConnectionStringBuilder
```
anpete: Build once and cache?

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqlServerFixture.cs
```diff
@@ -9,20 +10,29 @@ namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
...
+                .ServiceCollection()
+                .BuildServiceProvider();
         }
 
-        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        public override MigrationsContext CreateContext()
         {
-            optionsBuilder.UseSqlServer("" +
-                new SqlConnectionStringBuilder
+            var connectionStringBuilder = new SqlConnectionStringBuilder
```
anpete: Use test store?

test/EntityFramework.SqlServer.FunctionalTests/MigrationsSqlServerFixture.cs
```diff
@@ -9,20 +10,29 @@ namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
...
                 {
-                    DataSource = @"(localdb)\MSSQLLocalDB",
-                    InitialCatalog = nameof(MigrationsSqlServerTest),
-                    IntegratedSecurity = true
-                });
+                    InitialCatalog = nameof(MigrationsSqlServerTest)
+                };
+            connectionStringBuilder.ApplyConfiguration();
+
+            var optionsBuilder = new DbContextOptionsBuilder();
```
anpete: Build once and cache?

test/EntityFramework.Core.FunctionalTests/MonsterFixupTestBase.cs
```diff
@@ -771,29 +738,27 @@ private void Can_build_monster_model_and_seed_data_using_navigations_with_deferr
...
-            await Composite_fixup_happens_when_FKs_change_test(p => CreateChangedOnlyMonsterContext(p), ChangedOnlyDatabaseName, useDetectChanges: false);
+            await Composite_fixup_happens_when_FKs_change_test(() => Fixture.CreateChangedOnlyMonsterContext(), useDetectChanges: false);
         }
 
         private async Task Composite_fixup_happens_when_FKs_change_test(
-            Func<IServiceProvider, MonsterContext> createContext, string databaseName, bool useDetectChanges)
+            Func<MonsterContext> createContext, bool useDetectChanges)
         {
-            var serviceProvider = CreateServiceProvider();
+            await Fixture.CreateAndSeedDatabase(createContext);
```
anpete: This shouldn't need to be called in each test. That's what fixtures are for. It looks like there might be multiple fixtures here and so we may need to split out each set of tests so the fixtures can vary.

test/EntityFramework.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
divega: Thanks. I will file a test bug on this to follow up when we have time.

test/EntityFramework.Core.FunctionalTests/MonsterFixupTestBase.cs
natemcmaster: I'm going to backout changes to MonsterFixup and move to a separate PR. This is expanding beyond the scope of the original PR.

natemcmaster: Also, the CreateAndSeedDatabase function is not used in every test.

natemcmaster:
Updated. Added #3007 to track monsterfixup tests

divega:
:shipit:

PR #2945, opened 2015-08-27
First cut of initialization benchmarks

rowanmiller:
Also adding AdventureWorks model. Currently the tests that require the database are skipped if you don't have AdventureWorks installed because the artifacts to create it are large (backup is 45mb, SQL script to create schema and insert data is >500mb). We could look at dynamically downloading artifacts and creating it in the future if we think it adds significant value.


natemcmaster:
Perhaps add a script that downloads and installs?

rowanmiller:
Yeah, I'll play around with it a little more in the future, but for the moment I just want to get on to adding more coverage. The skip message gives you the URL to go download the backup from, so it's super easy to get the database.

natemcmaster:
:shipit:

PR #2949, opened 2015-08-28
Remove unnecessary dependency

bricelam:
I unblocked the build with e79118a558cf217950b4388bcd806bfaf1dccb98, but it wasn't the right change. @rowanmiller Were these added by Reverse Engineer? If so, why are they needed in entity class files, @lajones?


rowanmiller:
:shipit:

rowanmiller:
It wasn't reveng that added these because I generated in a separate project and then copied things in. I don't really know what would have added them, I didn't ever see a failure from them so it must have been something in tooling.

rowanmiller:
BTW what was failing with this change, looks like CI was passing and build.cmd passed for me

bricelam:
CI (and my local box) was failing with

    error CS0234: The type or namespace name 'Data' does not exist in the namespace 'Microsoft' (are
    you missing an assembly reference?)

PR #2954, opened 2015-08-28
Temp revert of change to SelectExpression for debug tooltips

mikary:
Pretty much exactly what you suggested @anpete


anpete:
:shipit:

Issue #2840, opened 2015-08-13
RevEng: Keys/AlternateKeys/Indexes


lajones:
What code should be generated from a unique constraint which is not the primary key - probably AlternateKey()?

Consider what code should be generated when a foreign key points to such a key in the database.

Should we add fluent API to name the constraint underlying a given foreign key / alternate key / primary key on the database - so that if we regenerated the database using migrations it would use the same names.

Do we need to add a .Index() in order to have access to the Name?

smitpatel:
In model snapshot, keys are declared as follows:
- If primary key then use `Key()`
- If referenced by FK then use `PrincipalKey()` on relationshipbuilder.
- Otherwise use `AlternateKey()`

rowanmiller:
Those rules sound pretty good for reverse engineer too.

natemcmaster:
We need to verify that we are correctly matching the unique index/unique constraint behavior of SQL Server. (See https://technet.microsoft.com/en-us/library/aa224827%28v=sql.80%29.aspx?f=255&MSPPError=-2147217396)

lajones:
Have updated to no longer output HasAlternateKey() fluent API in PR #3519.

lajones:
@rowanmiller @divega @ajcvickers @natemcmaster Re discussion from meeting on 10/21/15.

- Are we saying we want to output the `HasIndex(propertyNames).HasName(indexName).IsUnique(true|false)` into the fluent API in the generated `DbContext`?
- For which indexes do we want to do that? Those whose properties do not match either the primary key or any foreign key on the table? Does it matter if the order of the properties specified in the index is different from the order specified in the PK / FK?
- IIRC we removed the matching `Index` DataAnnotation - so we can't output that if the user wants to use attributes, right?

natemcmaster:
IMHO:

**non-unique indexes**
We should always scaffold `HasIndex()`.

**unique indexes**
Scaffold as `HasIndex().IsUnique()` unless it is already in scaffolded as a key (currently only possible via .PrincipalKey()).

**`[Index]`**
Match our current data annotation conventions. (@smitpatel how do we handle IndexAttribute?)

ajcvickers:
@natemcmaster And index on the dependent properties of a unique (1:1) FK should not need an explicit Index scaffolded either, although this depends on some work that may not have been done yet in the model building API.

Also, we do not support IndexAttribute at all in EF7.

natemcmaster:
[...]

PR #2957, opened 2015-08-28
RevEng: Allow outputting DataAnnotations instead of Fluent API

lajones:
@natemcmaster this touches the SQLite provider - so you might want to have a look.

Added a flag in the templates which controls whether we use DataAnnotations attributes on the EntityType instead of FluentApi in OnModelCreating() (where they are available - some fluent API does not have an equivalent attribute).

Also removed the *CodeGeneratorHelper and the *Model classes they operated on (and which were passed into the templates) and replaced them with directly passing in a (new) ModelConfiguration object for the DbContext template and an EntityConfiguration object for the EntityType template.

Updated the tests for both SqlServer and SQLite to test both using the attributes and just using the fluent API.


src/EntityFramework.Relational.Design/ReverseEngineering/Configuration/ModelConfiguration.cs
```diff
@@ -0,0 +1,534 @@
...
+            entityConfiguration.FluentApiConfigurations.Add(
+                new KeyFluentApiConfiguration("e", key.Properties));
+
+            foreach (var property in key.Properties)
+            {
+                var propertyConfiguration =
+                    entityConfiguration.GetOrAddPropertyConfiguration(entityConfiguration, property);
+                propertyConfiguration.AttributeConfigurations.Add(
+                    new AttributeConfiguration(nameof(KeyAttribute)));
+            }
```
smitpatel: Composite PK is not possible with `Key`

src/EntityFramework.Relational.Design/ReverseEngineering/Configuration/KeyFluentApiConfiguration.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            FluentApi = string.Format(
+                CultureInfo.InvariantCulture,
+                "{0}({1} => {2})",
+                nameof(EntityTypeBuilder.Key),
+                lambdaIdentifier,
+                new ModelUtilities().GenerateLambdaToKey(properties, lambdaIdentifier));
+        }
+
+        public virtual bool HasAttributeEquivalent { get; } = true;
```
smitpatel: Composite PK is not possible with `Key` so this may not be true always.

lajones: We used to allow this? See e.g. https://msdn.microsoft.com/en-us/data/jj591583.aspx#Composite.

rowanmiller: It caused a lot of confusion in EF6 because you had to combine it with [Column] to specify an order. In EF7 you need to use the Fluent API.

lajones: OK. We don't output anything if the KeyDiscoveryConvention would identify the property as a key anyway. So I'll just always use fluent API for this and never output the [Key] attribute.

lajones: @rowanmiller @smitpatel If the key is a single property, but doesn't match the naming conventions by which we would automatically make it a key would you prefer the [Key] attribute or the fluent API?

divega: FWIW I would prefer [Key] for that case.

ajcvickers: +1

src/EntityFramework.Relational.Design/ReverseEngineering/Configuration/ModelConfiguration.cs
```diff
@@ -0,0 +1,534 @@
...
+        }
+
+        public virtual void AddTableNameConfiguration([NotNull] EntityConfiguration entityConfiguration)
+        {
+            Check.NotNull(entityConfiguration, nameof(entityConfiguration));
+
+            var entityType = entityConfiguration.EntityType;
+            if (ExtensionsProvider.For(entityType).Schema != null
+                && ExtensionsProvider.For(entityType).Schema != "dbo")
```
natemcmaster: This is a SQL Server specific condition. Postgres uses a different default schema name.

lajones: @natemcmaster  Good catch. How about I setup a virtual getter DefaultSchemaName which by default is null and other providers override. If it's set then we use it here.

src/EntityFramework.Relational.Design/ReverseEngineering/Configuration/KeyFluentApiConfiguration.cs
lajones: OK. Will do that.

test/EntityFramework.Sqlite.Design.FunctionalTests/ReverseEngineering/SqliteAttributesInsteadOfFluentApiE2ETest.cs
```diff
@@ -0,0 +1,359 @@
...
+using Microsoft.Data.Entity.Relational.Design.ReverseEngineering;
+using Microsoft.Data.Entity.Sqlite.Design;
+using Microsoft.Data.Entity.Sqlite.Design.ReverseEngineering;
+using Microsoft.Data.Entity.Sqlite.FunctionalTests;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace EntityFramework.Sqlite.Design.FunctionalTests.ReverseEngineering
+{
+    public class SqliteAttributesInsteadOfFluentApiE2ETest : E2ETestBase
```
natemcmaster: This duplicates much of what is in SqliteE2ETest. It would be better to create a test base.

src/EntityFramework.Relational.Design/ReverseEngineering/Configuration/ModelConfiguration.cs
natemcmaster: @lajones `null`. There are no schema names in SQLite.

natemcmaster:
With #2840, we will need to expand the Metadata => Configurations logic for more scenarios. I expect this will be true of further additions to reveng. Each addition expands the size and complexity of `ModelConfiguration`.

Because of the tree-like structure of metadata, I think it would be beneficial to consider restructuring the numerous `ModelConfiguration.Add___Configuration()`'s into using the tree visitor pattern. Then the transformation could occur using a collection of visitors. Provider-writers can add new visitors instead of needing to override the entire chain of nested calls.

lajones:
@natemcmaster Re the tree visitor idea - sounds good. I'll do that as a separate PR.

test/EntityFramework.Sqlite.Design.FunctionalTests/ReverseEngineering/SqliteAttributesInsteadOfFluentApiE2ETest.cs
lajones: Agreed. I'll see if I can do it simply. If not I'll do as a separate PR.

natemcmaster:
@lajones Yes. This PR is already big enough. Add a TODO (and also one for the testing duplication, as we said in person)

lajones:
@natemcmaster in commit b9ced12 have refactored the SQLite E2ETest classes so they inherit from a common base class. Have also updated so KeyAttribute is produced as asked for above and so that we now use DefaultSchemaName instead of "dbo". Will add the TODO for the tree-visitor.

lajones:
In commit 3115251 added in a file I forgot to `git add` and removed some commented out code that I forgot to actually delete.

natemcmaster:
:shipit:

lajones:
Checked in with commit 6adcb44

test/EntityFramework.SqlServer.Design.FunctionalTests/ReverseEngineering/ExpectedResults/E2E_UseAttributesInsteadOfFluentApi/AllDataTypes.expected
```diff
@@ -0,0 +1,53 @@
...
+namespace E2ETest.Namespace
+{
+    public class AllDataTypes
+    {
+        public int AllDataTypesID { get; set; }
+        [Required]
+        public long bigintColumn { get; set; }
+        public byte[] binaryColumn { get; set; }
+        [Required]
+        public bool bitColumn { get; set; }
```
smitpatel: Is `Required` necessary here? The property is of non-nullable type.

test/EntityFramework.SqlServer.Design.FunctionalTests/ReverseEngineering/ExpectedResults/E2E_UseAttributesInsteadOfFluentApi/PropertyConfiguration.expected
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.Data.Entity;
+using Microsoft.Data.Entity.Metadata;
+
+namespace E2ETest.Namespace
+{
+    public class PropertyConfiguration
+    {
+        public byte PropertyConfigurationID { get; set; }
+        [Required]
+        public int A { get; set; }
```
smitpatel: `Required` is not required

lajones: Agreed. I'll fix in another PR.

PR #2958, opened 2015-08-28
[SMALL] Tweaks to performance tests

rowanmiller:
A few miscellaneous tweaks including:
* Adding a for loop to some of the init tests as the warm numbers are so small they report as zero
* Adding missing config for some server generated properties in the EF6 model


ajcvickers:
:shipit:

Issue #2805, opened 2015-08-10
EntityTypeConfiguration<T> (or similar)


mmillican:
Edited by @rowanmiller Oct-13-2016

`EntityTypeConfiguration<T>` is a feature in EF6.x that allows you to encapsulate the configuration for an entity type in a class. 

Here is some code you can use to enable the pattern until we add support in EF Core.

``` c#
using Microsoft.EntityFrameworkCore.Metadata.Builders;

namespace Microsoft.EntityFrameworkCore
{
    public abstract class EntityTypeConfiguration<TEntity>
        where TEntity : class
    {
        public abstract void Map(EntityTypeBuilder<TEntity> builder);
    }

    public static class ModelBuilderExtensions
    {
        public static void AddConfiguration<TEntity>(this ModelBuilder modelBuilder, EntityTypeConfiguration<TEntity> configuration)
            where TEntity : class
        {
            configuration.Map(modelBuilder.Entity<TEntity>());
        }
    }
}
```

And here is a sample application that uses these types:

``` c#
using Microsoft.EntityFrameworkCore;
using Microsoft.EntityFrameworkCore.Metadata.Builders;

namespace Demo
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var db = new BloggingContext())
            {
                db.Database.EnsureCreated();
            }
        }
    }

    public class BloggingContext : DbContext
    {
        public DbSet<Blog> Blogs { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=Demo;Trusted_Connection=True;");
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.AddConfiguration(new BlogMap());
        }
    }

    public class Blog
    {
        public int BlogId { get; set; }
        public string Name { get; set; }
        public string Url { get; set; }
    }

    public class BlogMap : EntityTypeConfiguration<Blog>
    {
        public override void Map(EntityTypeBuilder<Blog> builder)
        {
            builder.ToTable("tbl_blogs");

            builder.Property(b => b.Name)
                .IsRequired()
                .HasMaxLength(200);

            builder.Property(b => b.Url)
                .IsRequired()
                .HasMaxLength(500);

            builder.HasIndex(b => b.Url);
        }
    }
}
```

AndriySvyryd:
[...]

PR #2961, opened 2015-08-29
Adds and implements conventions for model builder configuration

Grinderofl:
Adds the ability to configure ModelBuilder from outside OnModelCreating method of DbContext.

Previous versions of EF lacked a proper way of doing it, this implementation gives the ability to add multiple ```IModelBuilderConvention``` types for configuration to individual/differing ```DbContext``` types. Dependency Injection is also supported.

Also sort-of solves #2805.

```c#
services.AddEntityFramework()
    .AddDbContext<Context>(builder => builder.UseConvention<MyBuilderConvention>());


public class MyBuilderConvention : IModelBuilderConvention
{
    public void Apply(ModelBuilder modelBuilder)
    {
        // (...);
    }
}

```


AndriySvyryd:
@Grinderofl 
Thanks for sending this PR. We've discussed it and arrived at the conclusion that the API needs to be more flexible to be useful for more scenarios, see #2992
This is something that we plan to address before RTM, but you are free to send another PR

PR #2970, opened 2015-09-01
RevEng: When is Required required?

lajones:
Only output the Required fluent API or attribute when property is a nullable type.

@smitpatel Addressing your comment in https://github.com/aspnet/EntityFramework/pull/2957


smitpatel:
:shipit:

lajones:
Checked in with commit 102b210.

PR #2973, opened 2015-09-01
Fix for RevEng EntityType Templates producing error CS0656

lajones:
@bricelam This should fix the Sqlite test failures you were seeing:

      System.InvalidOperationException : There was an error running the EntityType template. Message: Template Processing Failed: At
 line 3626. Message: (45,17): error CS0656: Missing compiler required member 'Microsoft.CSharp.RuntimeBinder.Binder.BinaryOperation'


bricelam:
:shipit: Thanks for investigating

lajones:
Checked in with commit 0d00d83.

Issue #2941, opened 2015-08-27
Make FromSqlQuery tests provider-agnostic


roji:
While trying to run more tests for PostgreSQL, I ran into the FromSqlQuery functional test suite. FromSqlQueryTestBase contains specific SQL queries such as `SELECT [Region], [PostalCode]...` which won't be supported across providers (square brackets). As another example, PostgreSQL requires table and column names to be quoted to maintain case-sensitivity (so `SELECT * FROM "Customers"` and not `SELECT * FROM Customers`).

Am not sure how I'd refactor this testsuite to be useful across providers, letting providers provide the actual SQL is a possibility but we'd end up with an abstract property per test, almost.

natemcmaster:
Yes, this should be changed.

anpete:
Those ones would probably just work if we removed the delimiters.

divega:
@mikary should this be closed now or are we waiting to hear from @roji if it worked for him?

roji:
Promise to look at this a bit later

roji:
Guys, unfortunately the changes aren't enough to make this PostgreSQL compatible because of the second example I mentioned above: in PG identifier must be quoted in order to preserve uppercase characters, otherwise PG converts everything to lowercase, so `SELECT PostalCode FROM Customers` must become `SELECT "PostalCode" FROM "Customers"`. I realize this wouldn't work with SqlServer, so it seems a more fundamental change in this testsuite is needed...

anpete:
@roji Is this because the Northwind schema objects are created with the quotes?

roji:
That's right. If the identifiers in the database aren't all-lowercase then you have to use double quotes when referencing them.

You could in theory change the Northeind scheme for this but it seems like a pretty huge change. Note that I only have issues with these suite since it defines SQL assertions in the *base* class, which doesn't seem like a good idea given provider variety out there (who knows what provider will come next with some SQL quirk)

anpete:
Agreed, and that is my concern with template-izing only part of the SQL. I.e. The delimiters. Either the SQL needs to be fully portable (may be possible at the moment if we just use " everywhere), or, we allow you to replace the entire SQL statement. I also don't want us to re-invent an ad hoc SQL generation provider model in the test.

Are the quotes the only think blocking you right now?

roji:
@anpete yeah, I think so... sorry that I don't have time to test for other issues right away (can do this tomorrow)

anpete:
[...]

Issue #2493, opened 2015-06-29
Consolidate and rationalize SQL generation


ajcvickers:
Currently different parts of the stack do SQL generation in different ways. This may be fine since, for example, query and update have different mechanisms and requirements. However, we should make sure that we are using common code where it makes sense. If nothing else, code for escaping, etc. should be able to be shared. Related to this is that if a provider needs to customize how things like escaping are done, then the provider should only need to do this in one place.

rowanmiller:
Make sure we API review IUpdateSqlGenerator et al before closing this

rowanmiller:
This should be a very scoped refactoring just to ensure providers don't have to override escaping etc. in multiple places

rowanmiller:
We are happy with what we have now

PR #2975, opened 2015-09-01
Make FromSql query tests provider-agnostic #2941

mikary:
Cleanup for the FromSql tests
@anpete @natemcmaster


natemcmaster:
Yes, `SqliteUpdateSqlGenerator.DelimitIdentifier`

mikary:
Discussed with @anpete 
* Delimiters were added due to mention of case-sensitivity issue in PostgreSQL described in #2941, if this is still an issue after these changes, we can address at that point
* Schema came from following templating patterns in `SqlGeneratorTestBase`, removed
* There are actually about 3 different provider APIs for SQL generation (to be addressed by #2493)

Updated to remove extra delimiters in original tests, reverting SQL generation pattern.

natemcmaster:
LGTM

anpete:
:shipit:

PR #2976, opened 2015-09-01
Downgrade System.Data.SqlClient package to one that targets dnxcore50

bricelam:
Note, we're also working with the team to ensure it comes back. This commit is just to unblock the build.


anpete:
:shipit:

PR #2979, opened 2015-09-01
Make Add/Attach/Update methods add dependents by default

ajcvickers:
This gives them a limited form of graph behavior, as discussed in the design meeting. Passing includeDependents: false reverts to the single object behavior.

Note that the params sugar methods for AddRange, etc. do not have the flag (because there is no good place to put it with this signature) and always do the default graph behavior. If the single object behavior is needed, then it can be obtained by using the IEnumerable overload instead of the params overload.

Also, we need to decide whether calling Add/Attach/Update with graph behavior on an already tracked entity causes the state of that entity to change, and, if so, whether the state of dependents should also change. Currently if the entity is already tracked, then its state is not changed (because this is the simplest code) until we decide to do differently.

The behavior of DetectChanges has not yet been updated.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/EntityEntryGraphNode.cs
+++ b/src/EntityFramework.Core/ChangeTracking/EntityEntryGraphNode.cs
@@ -1,0 +1,40 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.ChangeTracking.Internal;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.ChangeTracking
+{
+    public class EntityEntryGraphNode : EntityGraphNodeBase<EntityEntryGraphNode>
+    {
+        public EntityEntryGraphNode(
+            [NotNull] DbContext context,
+            [NotNull] InternalEntityEntry internalEntityEntry,
+            [CanBeNull] INavigation inboundNavigation)
+            : base(internalEntityEntry, inboundNavigation)
+        {
+            Check.NotNull(context, nameof(context));
+
+            Context = context;
+            Entry = new EntityEntry(context, internalEntityEntry);
+        }
+
+        public virtual DbContext Context { get; }
+        public new virtual EntityEntry Entry { get; }
+
+        public override EntityEntryGraphNode CreateNode(
+            EntityEntryGraphNode currentNode,
+            InternalEntityEntry internalEntityEntry,
+            INavigation reachedVia)
+            => new EntityEntryGraphNode(
+                Context,
+                Check.NotNull(internalEntityEntry, nameof(internalEntityEntry)),
+                Check.NotNull(reachedVia, nameof(reachedVia)))
+                {
+                    NodeState = Check.NotNull(currentNode, nameof(currentNode)).NodeState
+                };
+    }
+}
--- a/src/EntityFramework.Core/ChangeTracking/EntityGraphNodeBase.cs
+++ b/src/EntityFramework.Core/ChangeTracking/EntityGraphNodeBase.cs
@@ -1,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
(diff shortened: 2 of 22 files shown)


src/EntityFramework.Core/ChangeTracking/EntityEntryGraphNode.cs
```diff
@@ -0,0 +1,40 @@
...
+            Check.NotNull(context, nameof(context));
+
+            Context = context;
+            Entry = new EntityEntry(context, internalEntityEntry);
+        }
+
+        public virtual DbContext Context { get; }
+        public new virtual EntityEntry Entry { get; }
+
+        public override EntityEntryGraphNode NewNode(
```
anpete: "CreateNode"?

src/EntityFramework.Core/ChangeTracking/EntityGraphNodeBase.cs
```diff
@@ -0,0 +1,35 @@
...
+            [NotNull] InternalEntityEntry internalEntityEntry, 
+            [CanBeNull] INavigation reachedVia)
+        {
+            Check.NotNull(internalEntityEntry, nameof(internalEntityEntry));
+
+            Entry = internalEntityEntry;
+            ReachedVia = reachedVia;
+        }
+
+        public virtual INavigation ReachedVia { get; }
```
anpete: Not sure about this name. InboundNavigation?

src/EntityFramework.Core/ChangeTracking/Internal/IGraphAttacher.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.ChangeTracking.Internal
+{
+    public interface IGraphAttacher
```
anpete: Consider adding "Entity" somewhere in the name.

anpete:
:shipit:

src/EntityFramework.Core/DbContext.cs
```diff
@@ -740,24 +746,24 @@ private void SetEntityStates(IEnumerable<object> entities, EntityState entitySta
...
-        {
-            Check.NotNull(entities, nameof(entities));
-
-            SetEntityStates(entities, EntityState.Added);
-        }
+        /// <param name="includeDependents">
+        ///     If true, then the context will also start tracking all dependent entities reachable from the given entities using
+        ///     the same rules. If false, then only the given entities will be tracked.
+        /// </param>
+        public virtual void AddRange([NotNull] IEnumerable<object> entities, bool includeDependents = true)
```
divega: I would like to propose we introduce an enum type instead of a bool, e.g.: 

``` C#
 enum GraphOperationBehavior {IncludeDependents, SingleObject};
``` 
(This is the best name I have, but probably makes sense to keep iterating over it :smile:)

__Pros:__
1. Leads to more readable code when you use it
2. Allows for easily exposing new behaviors in the future without breaking changes
__Cons:__
1. A bit more complicated than a bool?

Another option that can be combined with the enum or not: Move to a context-level property instead of a parameter:
__Pros:__
1. Keeps `Add()` and `Attach()` simpler and AddRange() more consistent 
2. If your application works best with a specific behavior, you only need to make the choice once
__Cons:__
1. A bit less discoverable
2. You need to remember to switch back when you wanted a behavior only to apply only to a few calls (we have been afraid of that in the past, but in most cases it turned out well)

Note that this would look similar to what we decided to do for the [context-level default query tracking behavior property](https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes---August-6,-2015#default-query-tracking-behavior) although in this case I am proposing we get rid of the fine grained per-call option.

ajcvickers: Filed this as an issue so we can discuss.

src/EntityFramework.Core/ChangeTracking/EntityEntryGraphNode.cs
ajcvickers: Done.

src/EntityFramework.Core/ChangeTracking/EntityGraphNodeBase.cs
ajcvickers: Done

src/EntityFramework.Core/ChangeTracking/Internal/IGraphAttacher.cs
ajcvickers: Done

PR #2986, opened 2015-09-02
:racehorse: Better cold start benchmarks

rowanmiller:
Exploratory testing shows that creating a fresh ServiceProvider instance isn't a very good simulation of cold start (and there also isn't an equivalent approach that can be used for EF6). Introducing a sandbox helper that uses AppDomains to create a true cold start. This will only ever run on Full .NET since we can't do the AppDomain magic on CoreCLR.

### DIFF
--- a/test/EntityFramework.Microbenchmarks.Core/ColdStartSandbox.cs
+++ b/test/EntityFramework.Microbenchmarks.Core/ColdStartSandbox.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if !DNXCORE50
+
+using System;
+
+namespace EntityFramework.Microbenchmarks.Core
+{
+    public class ColdStartSandbox : IDisposable
+    {
+        private AppDomain _domain = AppDomain.CreateDomain(
+                "Cold Start Sandbox",
+                null,
+                new AppDomainSetup { ApplicationBase =  AppDomain.CurrentDomain.BaseDirectory });
+
+        ~ColdStartSandbox()
+        {
+            Dispose(false);
+        }
+
+        public T CreateInstance<T>(params object[] args)
+        {
+            return (T)CreateInstance(typeof(T), args);
+        }
+
+        public object CreateInstance(Type type, params object[] args)
+        {
+            HandleDisposed();
+
+            return _domain.CreateInstanceAndUnwrap(
+                type.Assembly.FullName,
+                type.FullName,
+                ignoreCase: false,
+                bindingAttr: 0,
+                binder: null,
+                args: args,
+                culture: null,
+                activationAttributes: null);
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing && _domain != null)
+            {
+                AppDomain.Unload(_domain);
+                _domain = null;
+            }
+        }
+
+        private void HandleDisposed()
+        {
+            if (_domain == null)
+            {
+                throw new ObjectDisposedException(null);
+            }
+        }
+    }
+}
+
+#endif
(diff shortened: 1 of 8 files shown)


test/EntityFramework.Microbenchmarks.Core/ColdStartSandbox.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+
+namespace EntityFramework.Microbenchmarks.Core
+{
+    public class ColdStartSandbox : IDisposable
+    {
+        private AppDomain _domain = AppDomain.CreateDomain(
+                Guid.NewGuid().ToString(),
+                null,
+                new AppDomainSetup { ApplicationBase = Environment.CurrentDirectory });
```
bricelam: `CurrentDirectory` is shared, global state. You may want to use `AppDomain.CurrentDomain.BaseDirectory` instead. BTW, that's the default so you can just use the overload of `CreateDomain` that takes only the name.

test/EntityFramework.Microbenchmarks.Core/ColdStartSandbox.cs
```diff
@@ -0,0 +1,67 @@
...
+#if !DNXCORE50
+
+using System;
+
+namespace EntityFramework.Microbenchmarks.Core
+{
+    public class ColdStartSandbox : IDisposable
+    {
+        private AppDomain _domain = AppDomain.CreateDomain(
+                Guid.NewGuid().ToString(),
```
bricelam: BTW, I've since learned that these don't have to be unique. They're really just for debugging.

test/EntityFramework.Microbenchmarks/Models/AdventureWorks/AdventureWorksFixture.cs
```diff
@@ -1,31 +1,20 @@
...
-                    .GetService()
-                    .BuildServiceProvider();
-
-                return new AdventureWorksContext(ConnectionString, serviceProvider);
-            }
-            else
-            {
-                return new AdventureWorksContext(ConnectionString);
-            }
+            return new AdventureWorksContext(ConnectionString);
```
bricelam: So this doesn't run on DNXCORE50 anymore?

test/EntityFramework.Microbenchmarks.EF6/InitializationTests.cs
```diff
@@ -94,5 +130,21 @@ public void BuildModel_AdventureWorks(MetricCollector collector)
...
             Assert.Equal(67, model.ConceptualModel.EntityTypes.Count());
         }
+
+        private void RunTest(bool cold, Action<InitializationTests> test)
+        {
+            if (cold)
+            {
+                using (var sandbox = new ColdStartSandbox())
+                {
+                    var testClass = sandbox.CreateInstance<InitializationTests>(_fixture);
```
bricelam: You could have a default constructor that news the fixture up. That way it'll never cross the domain boundary and doesn't have to be serializable.

bricelam: Although I'm not sure how xUnit.net will handle it.

bricelam:
:shipit:

test/EntityFramework.Microbenchmarks.EF6/InitializationTests.cs
rowanmiller: Yeah I thought about that. This particular fixture isn't really necessary because all the config is static and there is no database existence logic - so sharing the same instance between test methods doesn't really buy anything. It is nice to keep using the same consistent fixture patterns though.

test/EntityFramework.Microbenchmarks/Models/AdventureWorks/AdventureWorksFixture.cs
rowanmiller: The cold variations do not run on core, the warm ones do.

test/EntityFramework.Microbenchmarks.Core/ColdStartSandbox.cs
rowanmiller: Sounds good, I will do that.

test/EntityFramework.Microbenchmarks.Core/ColdStartSandbox.cs
rowanmiller: I'll just use something static then like "Cold Start Sandbox"

test/EntityFramework.Microbenchmarks.Core/ColdStartSandbox.cs
rowanmiller: Found that it failed to load assemblies if I didn't specify the `ApplicationBase`, so I left it in but swapped to use `AppDomain.CurrentDomain.BaseDirectory `

rowanmiller:
@bricelam tidied up as we discussed in person (I think it's a lot cleaner), let me know if the changes look good to you

bricelam:
Looks good. Much better.

PR #2988, opened 2015-09-02
Use new packInclude property in project.json

bricelam:
Note, this will be easier to review [without whitespace changes](https://github.com/aspnet/EntityFramework/pull/2988/files?w=1).


anpete:
:shipit:

PR #2990, opened 2015-09-02
Fix #2951 - Throw when we encounter a sentinel key value during mater…

### LINKED ISSUE #2951: EF7 bug with zero value in primary key column
I have simple table Test with one column Age (int) with primary key on this column.
```
CREATE TABLE [dbo].[Test](
	[Age] [int] NOT NULL,
	CONSTRAINT [PK_Test] PRIMARY KEY CLUSTERED ([Age] ASC)
WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY])
ON [PRIMARY]
```
My table has 2 values (0, 1).
Also I have model for this table
```
modelBuilder.Entity<Test_Entity>(x =>
{
	x.ToTable("Test", "dbo");
	x.Key(i => i.Age);
});

public DbSet<Test_Entity> Test { get; set; }

public class Test_Entity
{
	public int Age { get; set; }
}

```
I need read all data to array like:
```
var result = Test.ToArray();
```
In result I am waiting something like:
```
[0] | { Age = 0 }
[1] | { Age = 1 }
```
But in fact I have:
```
[0] | null
[1] | { Age = 1 }
```
In Beta 2 this tests runs well. But in Beta 5, 6, 7, 8 I have this error.

My previous report about this problem https://github.com/aspnet/EntityFramework/issues/2117

anpete:
…ialization.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/CompositeEntityKeyFactory.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/CompositeEntityKeyFactory.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Storage;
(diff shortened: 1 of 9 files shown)


anpete:
@ajcvickers

anpete:
Going to try and make this better :-)

Issue #2403, opened 2015-06-18
Have a way to turn off null semantics at the context level


divega:
(just extracting what is actionable from https://github.com/aspnet/EntityFramework/issues/2106)

E.g.:

DbContext.Database.AsRelational().UseDatabaseNullSemantics = true;

Note that this will have an impact on query caching, e.g. at query execution time we should generate different SQL based on the same source LINQ query depending on the state of this flag.

ajcvickers:
@divega Should this be assigned to someone? Or should we triage it again? We have a note in the API review to make sure this gets done and that UseRelationalNullSemantics is then removed.

divega:
@ajcvickers It should have been assigned to @maumar but somehow it got missed. Thanks!

maumar:
I will deal with this as soon as I'm done with my current item

maumar:
fixed in 0a0415e7c04caa1efedbb1d96b7b129ed9e64c3f

PR #3008, opened 2015-09-03
Fix to #2403 - Have a way to turn off null semantics at the context level

maumar:
Removing necessity to setup NullSemantics on a per query basis - added a global switch:
context.Database.UseDatabaseNullSemantics(bool)

This also affects caching - we should produce different cache entries based on the value of UseDatabaseNullSemantics.
This can be used as a sample for other providers who want to introduce context-wide switches that affect the query.

### DIFF
--- a/src/EntityFramework.Relational/Query/RelationalCompiledQueryCacheKeyGenerator.cs
+++ b/src/EntityFramework.Relational/Query/RelationalCompiledQueryCacheKeyGenerator.cs
@@ -5,16 +5,16 @@ using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Query.ExpressionVisitors;
-using Microsoft.Data.Entity.Utilities;
 using Microsoft.Data.Entity.Storage;
+using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Query
 {
-    public class CompiledQueryCacheKeyGenerator : ICompiledQueryCacheKeyGenerator
+    public class RelationalCompiledQueryCacheKeyGenerator : ICompiledQueryCacheKeyGenerator
     {
         private readonly IModel _model;
 
-        public CompiledQueryCacheKeyGenerator([NotNull] IModel model)
+        public RelationalCompiledQueryCacheKeyGenerator([NotNull] IModel model)
         {
             Check.NotNull(model, nameof(model));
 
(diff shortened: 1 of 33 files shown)


maumar:
@anpete @mikary

anpete:
This shouldn't affect the cache key directly. It should affect it by introducing the UseRelationalNullSemantics annotation into the query expression.

src/EntityFramework.Relational/Query/RelationalCompiledQueryCacheKeyGenerator.cs
```diff
@@ -0,0 +1,70 @@
...
+                ((RelationalDatabase)database).UseDatabaseNullSemantics);
+
+        private struct CompiledQueryCacheKey
+        {
+            private readonly string _query;
+            private readonly IModel _model;
+            private readonly bool _async;
+            private readonly bool _useDatabaseNullSemantics;
+
+            public CompiledQueryCacheKey(string query, IModel model, bool async, bool useDatabaseNullSemantics)
```
maumar: alternatively we could reuse the existing CompiledQueryCacheKey, by introducing some sort of list of annotations. However this solution is bit faster (as you can rely on hashcode comparisons. Thoughts?

anpete: See my comment in the Conversation tab.

src/EntityFramework.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -30,6 +30,9 @@ public static void Migrate([NotNull] this DatabaseFacade databaseFacade)
                         sql,
                         parameters);
 
+        public static void UseDatabaseNullSemantics([NotNull] this DatabaseFacade databaseFacade, bool useDatabaseNullSemantics)
```
divega: Now that we have flattened relational extensions into the database façade this will read:

``` C#
context.Database.UseDatabaseNullSemantics(true);
```
rather than
``` C#
context.Database.AsRelational().UseDatabaseNullSemantics();
```

Hence I think we should swing back to the `UseRelationalNullSemantics` name.

divega:
Talked to @maumar and @anpete about whether we should leave the query operator version as public. At present I have two main objections:
1. The behavior seems very inconsistent, e.g. it is supposed to apply to individual query roots but it appears that it always spreads to the whole query (@maumar is looking at my repro)
2. The name is horrid. I would love it if could find a more straightforward and laymanish way to refer to the whole thing. A couple of incremental improvements we though of today:
 - WithRelationalNullSemantics()
 - UsingRelationalNullSemantics()
 - UsingRelationalNullBehavior()
 - UsingDatabaseNulls()

src/EntityFramework.Relational/RelationalDatabaseFacadeExtensions.cs
rowanmiller: :+1:

maumar:
@anpete @divega  new version up, please take a look
@rowanmiller are you fine with the name "UseRelationalNulls" instead of "UseDatabaseNullSemantics"?

src/EntityFramework.Core/Query/CompiledQueryCacheKeyGenerator.cs
```diff
@@ -20,7 +21,7 @@ public CompiledQueryCacheKeyGenerator([NotNull] IModel model)
             _model = model;
         }
 
-        public virtual object GenerateCacheKey([NotNull] Expression query, bool async)
+        public virtual object GenerateCacheKey([NotNull] Expression query, IDatabase database, bool async)
```
anpete: Don't flow the whole database, start with just the bool value of the option.

anpete: I think we are likely going to want some kind of query options objection going forward - async would live there too, and the stuff @natemcmaster is working on. But fine to postpone that for now.

anpete:
We are going to do this for now and triangulate with @natemcmaster next week.

:shipit:

PR #3011, opened 2015-09-04
[query] Add support for OFFSET and FETCH on SQL Server 2008

### LINKED ISSUE #2180: Generating SQL2012 specific code when using 2008R2
I have a query that has Skip() and Take() calls to implement paging.  The code generated is in the form
```sql
SELECT ...FROM [Article] AS [a]
WHERE [a].[PublishingStatusId] = 2
ORDER BY [a].[Updated] DESC OFFSET 0 ROWS FETCH NEXT 25 ROWS ONLY 
```
The OFFSET ...  code is SQL2012+ only.

natemcmaster:
Modifies the query structure to include the ROW_NUMBER workaround for limited support of OFFSET and FETCH on SQL 2008,

Add legacy paging extension to options builder.

```c#
options.UseSqlServer("Database=Kilimanjaro2008").UseRowNumberForPaging()
```

Fix #2180

### DIFF
--- a/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Data.Entity.Query
         public static readonly MethodInfo PropertyMethodInfo
             = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
 
-        private readonly IModel _model;
+        protected virtual IModel Model { get; }
         private readonly IQueryOptimizer _queryOptimizer;
         private readonly INavigationRewritingExpressionVisitorFactory _navigationRewritingExpressionVisitorFactory;
         private readonly ISubQueryMemberPushDownExpressionVisitor _subQueryMemberPushDownExpressionVisitor;
@@ -94,7 +94,7 @@ namespace Microsoft.Data.Entity.Query
             Check.NotNull(expressionPrinter, nameof(expressionPrinter));
             Check.NotNull(queryCompilationContext, nameof(queryCompilationContext));
 
-            _model = model;
+            Model = model;
             _queryOptimizer = queryOptimizer;
             _navigationRewritingExpressionVisitorFactory = navigationRewritingExpressionVisitorFactory;
             _subQueryMemberPushDownExpressionVisitor = subQueryMemberPushDownExpressionVisitor;
@@ -129,8 +129,6 @@ namespace Microsoft.Data.Entity.Query
         public virtual ILinqOperatorProvider LinqOperatorProvider => QueryCompilationContext.LinqOperatorProvider;
 
         public virtual StreamedSequenceInfo StreamedSequenceInfo => _streamedSequenceInfo;
-
-        protected virtual IModel Model => _model;
 
         public virtual Func<QueryContext, IEnumerable<TResult>> CreateQueryExecutor<TResult>([NotNull] QueryModel queryModel)
         {
@@ -362,7 +360,7 @@ namespace Microsoft.Data.Entity.Query
                 foreach (
                     var navigation in
                         from propertyInfo in chainedNavigationProperties
-                        let entityType = _model.FindEntityType(propertyInfo.DeclaringType)
+                        let entityType = Model.FindEntityType(propertyInfo.DeclaringType)
                         select entityType?.FindNavigation(propertyInfo.Name))
                 {
                     if (navigation == null)
@@ -1089,7 +1087,7 @@ namespace Microsoft.Data.Entity.Query
 
             while (memberExpression?.Expression != null)
             {
-                var entityType = _model.FindEntityType(memberExpression.Expression.Type);
+                var entityType = Model.FindEntityType(memberExpression.Expression.Type);
 
                 if (entityType == null)
                 {
@@ -1173,7 +1171,7 @@ namespace Microsoft.Data.Entity.Query
                         || querySource == null
                         || querySource == querySourceReferenceExpression.ReferencedQuerySource)
                     {
-                        var entityType = _model.FindEntityType(methodCallExpression.Arguments[0].Type);
+                        var entityType = Model.FindEntityType(methodCallExpression.Arguments[0].Type);
 
                         if (entityType != null)
                         {
--- a/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Data.Entity.Query
 {
     public class RelationalQueryModelVisitor : EntityQueryModelVisitor
     {
-        private readonly Dictionary<IQuerySource, SelectExpression> _queriesBySource
+        protected virtual Dictionary<IQuerySource, SelectExpression> QueriesBySource { get; }
             = new Dictionary<IQuerySource, SelectExpression>();
 
         private readonly Dictionary<IQuerySource, RelationalQueryModelVisitor> _subQueryModelVisitorsBySource
@@ -142,7 +142,7 @@ namespace Microsoft.Data.Entity.Query
         public new virtual RelationalQueryCompilationContext QueryCompilationContext
             => (RelationalQueryCompilationContext)base.QueryCompilationContext;
 
-        public virtual ICollection<SelectExpression> Queries => _queriesBySource.Values;
+        public virtual ICollection<SelectExpression> Queries => QueriesBySource.Values;
 
         public virtual RelationalQueryModelVisitor ParentQueryModelVisitor { get; }
 
@@ -160,7 +160,7 @@ namespace Microsoft.Data.Entity.Query
             Check.NotNull(querySource, nameof(querySource));
             Check.NotNull(selectExpression, nameof(selectExpression));
 
-            _queriesBySource.Add(querySource, selectExpression);
+            QueriesBySource.Add(querySource, selectExpression);
         }
 
         public virtual SelectExpression TryGetQuery([NotNull] IQuerySource querySource)
@@ -168,9 +168,9 @@ namespace Microsoft.Data.Entity.Query
             Check.NotNull(querySource, nameof(querySource));
 
             SelectExpression selectExpression;
-            return (_queriesBySource.TryGetValue(querySource, out selectExpression)
+            return (QueriesBySource.TryGetValue(querySource, out selectExpression)
                 ? selectExpression
-                : _queriesBySource.Values.SingleOrDefault(se => se.HandlesQuerySource(querySource)));
+                : QueriesBySource.Values.SingleOrDefault(se => se.HandlesQuerySource(querySource)));
         }
 
         public override void VisitQueryModel(QueryModel queryModel)
@@ -182,7 +182,7 @@ namespace Microsoft.Data.Entity.Query
             var compositePredicateVisitor
                 = _compositePredicateExpressionVisitorFactory.Create(((RelationalDatabase)_database).UseRelationalNulls);
 
-            foreach (var selectExpression in _queriesBySource.Values.Where(se => se.Predicate != null))
+            foreach (var selectExpression in QueriesBySource.Values.Where(se => se.Predicate != null))
             {
                 selectExpression.Predicate
                     = compositePredicateVisitor.Visit(selectExpression.Predicate);
@@ -315,7 +315,7 @@ namespace Microsoft.Data.Entity.Query
                                 .AddCrossJoin(selectExpression.Tables.First(), selectExpression.Projection);
                         }
 
-                        _queriesBySource.Remove(fromClause);
+                        QueriesBySource.Remove(fromClause);
 
                         Expression
                             = _queryFlatteningExpressionVisitorFactory.Create(
@@ -425,7 +425,7 @@ namespace Microsoft.Data.Entity.Query
 
                     if (predicate != null)
                     {
-                        _queriesBySource.Remove(joinClause);
+                        QueriesBySource.Remove(joinClause);
 
                         previousSelectExpression.RemoveRangeFromProjection(previousSelectProjectionCount);
 
--- a/src/EntityFramework.SqlServer/Extensions/SqlServerDbContextOptionsBuilder.cs
+++ b/src/EntityFramework.SqlServer/Extensions/SqlServerDbContextOptionsBuilder.cs
@@ -15,5 +15,11 @@ namespace Microsoft.Data.Entity.SqlServer.Extensions
 
         protected override SqlServerOptionsExtension CloneExtension()
             => new SqlServerOptionsExtension(OptionsBuilder.Options.GetExtension<SqlServerOptionsExtension>());
+
+        /// <summary>
+        ///     Use a ROW_NUMBER() in queries instead of OFFSET/FETCH. This method is backwards-compatible to SQL Server 2005.
+        /// </summary>
+        public virtual void UseRowNumberForPaging()
+            => SetOption(e => e.RowNumberPaging = true);
     }
 }
--- a/src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
+++ b/src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
@@ -1,0 +1,70 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Query.Sql;
+using Microsoft.Data.Entity.Utilities;
+using Remotion.Linq.Clauses;
+
+namespace Microsoft.Data.Entity.Query.Expressions
+{
+    public class RowNumberExpression : Expression
+    {
+        private readonly List<Ordering> _orderings = new List<Ordering>();
+
+        public RowNumberExpression(
+            [NotNull] ColumnExpression columnExpression,
+            [NotNull] IReadOnlyList<Ordering> orderings)
+        {
+            Check.NotNull(columnExpression, nameof(columnExpression));
+            Check.NotNull(orderings, nameof(orderings));
+
+            ColumnExpression = columnExpression;
+            _orderings.AddRange(orderings);
+        }
+
+        public virtual ColumnExpression ColumnExpression { get; }
+
+        public override ExpressionType NodeType => ExpressionType.Extension;
+        public override bool CanReduce => false;
+        public override Type Type => ColumnExpression.Type;
+
+        public virtual IReadOnlyList<Ordering> Orderings => _orderings;
+
+        protected override Expression Accept(ExpressionVisitor visitor)
+        {
+            Check.NotNull(visitor, nameof(visitor));
+
+            var specificVisitor = visitor as ISqlServerExpressionVisitor;
+
+            return specificVisitor != null
+                ? specificVisitor.VisitRowNumber(this)
+                : base.Accept(visitor);
+        }
+
+        protected override Expression VisitChildren(ExpressionVisitor visitor)
+        {
+            var newColumnExpression = visitor.Visit(ColumnExpression) as ColumnExpression;
+
+            var newOrderings = new List<Ordering>();
+            var recreate = false;
+            foreach (var ordering in _orderings)
+            {
+                var newOrdering = new Ordering(visitor.Visit(ordering.Expression), ordering.OrderingDirection);
+                newOrderings.Add(newOrdering);
+                recreate |= newOrdering.Expression != ordering.Expression;
+            }
+
+            if (recreate ||
+                (newColumnExpression != null && ColumnExpression != newColumnExpression))
+            {
+                return new RowNumberExpression(newColumnExpression, newOrderings);
+            }
+
+            return this;
+        }
+    }
+}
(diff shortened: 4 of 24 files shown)


src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -36,7 +36,7 @@ public abstract class EntityQueryModelVisitor : QueryModelVisitorBase
         public static readonly MethodInfo PropertyMethodInfo
             = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
 
-        private readonly IModel _model;
+        protected readonly IModel Model;
```
anpete: We typically just use auto props for this.

test/EntityFramework.SqlServer.Tests/SqlServerDbContextOptionsExtensionsTest.cs
```diff
@@ -93,5 +93,18 @@ public void Can_add_extension_with_connection_using_generic_options()
             Assert.Same(connection, extension.Connection);
             Assert.Null(extension.ConnectionString);
         }
+
+        [Fact]
+        public void Can_add_extension_with_legacy_paging()
+        {
+            var optionsBuilder = new DbContextOptionsBuilder<DbContext>();
+
+            optionsBuilder.UseSqlServer("Database=Kilimanjaro").UseLegacyPaging();
```
anpete: One question here is whether the config should be fine grained like this, or, whether you should just opt-in to the whole dialect. E.g. Use2005Dialect()

test/EntityFramework.SqlServer.FunctionalTests/NorthwindLegacyPagingQuerySqlServerFixture.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.Data.Entity.FunctionalTests;
+using Microsoft.Data.Entity.FunctionalTests.TestModels.Northwind;
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.SqlServer.FunctionalTests.TestModels;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
+{
+    public class NorthwindLegacyPagingQuerySqlServerFixture : NorthwindQueryRelationalFixture, IDisposable
```
anpete: Just subclass the SQL Server fixture and add a template method to configure the options. #dry

test/EntityFramework.SqlServer.FunctionalTests/LegacyPagingTest.cs
```diff
@@ -0,0 +1,162 @@
...
+        public LegacyPagingTest(NorthwindLegacyPagingQuerySqlServerFixture fixture, ITestOutputHelper testOutputHelper)
+            : base(fixture)
+        {
+            //TestSqlLoggerFactory.CaptureOutput(testOutputHelper);
+        }
+
+        public void Dispose()
+        {
+            //Assert for all tests that OFFSET or FETCH is never used
+            Assert.DoesNotContain("OFFSET ", Sql);
```
anpete: Sneaky

src/EntityFramework.SqlServer/Query/SqlServerQueryModelVisitorFactory.cs
```diff
@@ -0,0 +1,143 @@
...
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Query.ExpressionVisitors;
+using Microsoft.Data.Entity.Query.Internal;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Query
+{
+    public class SqlServerQueryModelVisitorFactory : IEntityQueryModelVisitorFactory
```
anpete: This needs to go away soon!

anpete:
It's looking pretty good, but we need to address query caching. Let's chat.

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
natemcmaster: Resharper strikes again.

test/EntityFramework.SqlServer.Tests/SqlServerDbContextOptionsExtensionsTest.cs
natemcmaster: We discussed a little about this in #2180. Because SQL Azure is a grab bag of features that does not align with SQL sever, we still need this little knobs. @divega https://github.com/aspnet/EntityFramework/issues/2180#issuecomment-136528449 showed preference for the ability to opt-in to specific features. Perhaps we could add a Use2005Dialect() extension  that represents a set of features.

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
```diff
@@ -0,0 +1,80 @@
...
+            var specificVisitor = visitor as ISqlServerExpressionVisitor;
+
+            return specificVisitor != null
+                ? specificVisitor.VisitRowNumber(this)
+                : base.Accept(visitor);
+        }
+
+        protected override Expression VisitChildren(ExpressionVisitor visitor)
+        {
+            visitor.Visit(ColumnExpression);
```
maumar: visit orderings as well, also if any of them change - recreate the RowNumberExpression with new arguments

test/EntityFramework.SqlServer.Tests/SqlServerDbContextOptionsExtensionsTest.cs
divega: I don't have a strong preference. The thing we once said we didn't want to do is specify the server version and pivot on that for potentially __all__ kinds of behavior. Just the SQL dialect feels smaller and potentially manageable, but it feels ambiguous and could be a slippery slope. E.g. by dialect are we just referring to the SELECT queries we are going to generate or does it affect CUD and DDL which are also SQL? If the latter, then it is almost the same thing we said we were going to avoid :smile:

ajcvickers: I prefer that we address specific cases rather than an over-arching dialog, especially since we're not trying to make everything always work with SQL 2005 (or whatever) when you have this flag turned on, and we don't have a plan to do many of these kinds of flags. So basically, I am agreeing with what @divega said. :-)

anpete: I'm fine either way, just wanted to make sure we were being explicit here. I also think @natemcmaster's idea of the higher level sugar that builds on the fine-grained stuff could be worth pursuing.

divega: I agree that is probably a good idea and we may very well end up doing it. 

So for now I would just add the method for paging and name it very explicitly after what it does, e.g. something like `UseRowNumberForPaging()` rather than `UseLegacyPaging()`.

We should then keep testing against specific versions and SKUs of SQL Server to see what else we find we need to do. 

When/if we add these sugar methods we may use them to control other aspects besides the dialect of SQL we generate, e.g. database creation options, or even default batch sizes, retry logic some day, etc.

Another thing we can consider is whether to use `optionsBuilder.UseSqlServer(...).UseXBehavior()` or `optionsBuilder.UseSqlServerX(...)`.

natemcmaster:
@anpete  added query caching.

anpete:
@natemcmaster Add a FT that checks cache entry resolution.

test/EntityFramework.SqlServer.Tests/SqlServerDbContextOptionsExtensionsTest.cs
natemcmaster: @divega how about `UseDerivedTablePaging()` ?

anpete: I prefer UseRowNumberForPaging as it is a little clearer to me.

ErikEJ: @anpete +1 - https://www.mssqltips.com/sqlservertip/2696/comparing-performance-for-different-sql-server-paging-methods/

natemcmaster:
Updated: renamed to API to `UseRowNumberForPaging()`

test/EntityFramework.SqlServer.FunctionalTests/NorthwindQuerySqlServerFixture.cs
```diff
@@ -49,4 +55,14 @@ public override NorthwindContext CreateContext()
 
         public void Dispose() => _testStore.Dispose();
     }
+
+    public class NorthwindLegacyPagingQuerySqlServerFixture : NorthwindQuerySqlServerFixture
```
anpete: Separate file. Rename to match API name?

test/EntityFramework.SqlServer.FunctionalTests/CompiledQueryCacheKeyGeneratorTest.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                var services = ((IAccessor<IServiceProvider>)context2).Service.GetService<IDbContextServices>().DatabaseProviderServices;
+                var generator = services.CompiledQueryCacheKeyGenerator;
+                key2 = generator.GenerateCacheKey(query, services.Database, false);
+            }
+
+            Assert.NotEqual(key1, key2);
+        }
+
+        public class QueryKeyCacheContext : DbContext
```
anpete: Use the fixture below? Also, this test seems like half functional, half unit :-) Might be easier just to just run a full end to end query, verify SQL, change the flag and then re-run the query.

src/EntityFramework.SqlServer/Query/Expressions/SqlServerVariableExpression.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Query.Sql;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Query.Expressions
+{
+    public class SqlServerVariableExpression : Expression
+    {
+        public SqlServerVariableExpression([NotNull] string name, [NotNull] Type type)
```
anpete: Can we just use ParameterExpression?

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -23,7 +23,7 @@ namespace Microsoft.Data.Entity.Query
 {
     public class RelationalQueryModelVisitor : EntityQueryModelVisitor
     {
-        private readonly Dictionary<IQuerySource, SelectExpression> _queriesBySource
+        protected readonly Dictionary<IQuerySource, SelectExpression> QueriesBySource
```
anpete: Property.

src/EntityFramework.SqlServer/Extensions/SqlServerDbContextOptionsBuilder.cs
```diff
@@ -15,5 +15,11 @@ public SqlServerDbContextOptionsBuilder([NotNull] DbContextOptionsBuilder option
 
         protected override SqlServerOptionsExtension CloneExtension()
             => new SqlServerOptionsExtension(OptionsBuilder.Options.GetExtension<SqlServerOptionsExtension>());
+
+        /// <summary>
+        ///     Use a paging method in select queries that is backwards-compatible to SQL Server 2005.
```
anpete: Call out RowNumber in the docs here.

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
```diff
@@ -0,0 +1,93 @@
...
+            _orderings.AddRange(orderings);
+        }
+
+        public virtual ColumnExpression ColumnExpression { get; }
+
+        public override ExpressionType NodeType => ExpressionType.Extension;
+        public override bool CanReduce => false;
+        public override Type Type => ColumnExpression.Type;
+
+        public virtual IReadOnlyList<Ordering> Orderings => _orderings.AsReadOnly();
```
anpete: We don't normally worry about this.

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
```diff
@@ -0,0 +1,93 @@
...
+                ? specificVisitor.VisitRowNumber(this)
+                : base.Accept(visitor);
+        }
+
+        protected override Expression VisitChildren(ExpressionVisitor visitor)
+        {
+            var newColumnExpression = visitor.Visit(ColumnExpression);
+
+            var recreate = false;
+            foreach (var ordering in _orderings)
```
anpete: Can you use VisitAndConvert here?

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                var expression = visitor.Visit(ordering.Expression);
+                recreate |= expression != ordering.Expression;
+            }
+
+            if (recreate || !ColumnExpression.Equals(newColumnExpression))
```
anpete: Reference equality is typically enough here.

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                return new RowNumberExpression(ColumnExpression, _orderings);
+            }
+
+            return this;
+        }
+
+        protected virtual bool Equals([NotNull] RowNumberExpression other)
```
anpete: Do you really need Equals and GetHashCode?

src/EntityFramework.SqlServer/Query/SqlServerQueryModelVisitor.cs
```diff
@@ -0,0 +1,160 @@
...
+                VisitSelectExpression(selectExpression);
+            }
+        }
+
+        private bool RequiresLegacyPaging(SelectExpression selectExpression)
+            => selectExpression.Offset.HasValue
+               && selectExpression.Offset != 0
+               && !selectExpression.Projection.Any(p => p is RowNumberExpression);
+
+        private Expression VisitSelectExpression(SelectExpression selectExpression)
```
anpete: Don't you need to use a visitor here? Sub-queries with paging could potentially occur anywhere in the tree.

natemcmaster: I should rename this method. This method isn't really a visitor. 

> Sub-queries with paging could potentially occur anywhere in the tree.

The method visits the tree recursively to adjust any selectexpressions that need adjustment.

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
natemcmaster: Figured it doesn't hurt...

src/EntityFramework.SqlServer/Query/Expressions/SqlServerVariableExpression.cs
natemcmaster: Nope. Our parameter binding will bind an empty string
`SqlCommand.AddParameterWithValue("@@RowCount",null)`

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
natemcmaster: Not sure how that API works, despite reading documentation and looking at examples...sample?

anpete: SqlFunctionExpression

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
anpete: YAGNI

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
natemcmaster: `Ordering` isn't a subtype of expression, so I don't think it will work.

anpete: Oh yeah, Ordering is kind of annoying that way.

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
```diff
@@ -0,0 +1,93 @@
...
+            var recreate = false;
+            foreach (var ordering in _orderings)
+            {
+                var expression = visitor.Visit(ordering.Expression);
+                recreate |= expression != ordering.Expression;
+            }
+
+            if (recreate || ColumnExpression != newColumnExpression)
+            {
+                return new RowNumberExpression(ColumnExpression, _orderings);
```
anpete: Looks like we have lost the new Ordering expression(s).

src/EntityFramework.SqlServer/Query/Expressions/SqlServerVariableExpression.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System;
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Query.Sql;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Query.Expressions
+{
+    public class SqlServerVariableExpression : Expression
```
anpete: I see now that this is @@RowCount, which is a System Function™. We should be able to use SqlFunctionExpression.

natemcmaster: SqlFunctionExpression would append parenthesis => `@@RowCount()`

src/EntityFramework.SqlServer/Query/SqlServerQueryModelVisitor.cs
```diff
@@ -0,0 +1,153 @@
...
+using Microsoft.Data.Entity.Storage;
+using Microsoft.Data.Entity.Utilities;
+using Remotion.Linq;
+using Remotion.Linq.Clauses;
+
+namespace Microsoft.Data.Entity.Query
+{
+    public class SqlServerQueryModelVisitor : RelationalQueryModelVisitor
+    {
+        private const string RowNumberColumnName = "__RowNumber__";
```
anpete: Do we need the underscores?

src/EntityFramework.SqlServer/Query/SqlServerQueryModelVisitor.cs
```diff
@@ -0,0 +1,153 @@
...
+                }
+            }
+        }
+
+        private bool RequiresLegacyPaging(SelectExpression selectExpression)
+            => selectExpression.Offset.HasValue
+               && selectExpression.Offset != 0
+               && !selectExpression.Projection.Any(p => p is RowNumberExpression);
+
+        private void AdjustForLegacyPaging(SelectExpression selectExpression)
```
anpete: I still can't see where you are visiting sub-queries in the predicate/projection/orderby.

src/EntityFramework.SqlServer/Query/Expressions/SqlServerVariableExpression.cs
anpete: Handle VisitSqlFunction in SqlServerQuerySqlGenerator and check for '@@'

src/EntityFramework.SqlServer/Query/Expressions/RowNumberExpression.cs
natemcmaster: Can you clarify why we must return new expression instances from `VisitChildren`?

We are inconsistent about `VisitChildren` returning new instances or not. SelectExpression, CrossApply, and CrossJoin do not. Others do.

src/EntityFramework.SqlServer/Query/SqlServerQueryModelVisitor.cs
natemcmaster: No. This is arbitrary.

src/EntityFramework.SqlServer/Query/SqlServerQueryModelVisitor.cs
natemcmaster: I visit all SelectExpression's in `selectExpression.Tables`. Can subqueries also be nested in `.OrderBy` and `.Predicate`?

anpete: Yep

src/EntityFramework.SqlServer/Query/SqlServerQueryModelVisitor.cs
```diff
@@ -0,0 +1,165 @@
...
+            {
+                var visitor = new RowNumberPagingExpressionVisitor();
+                foreach (var selectExpression in QueriesBySource.Values)
+                {
+                    visitor.Visit(selectExpression);
+                }
+            }
+        }
+
+        private class RowNumberPagingExpressionVisitor : ExpressionVisitorBase
```
natemcmaster: @anpete correct if wrong. This visitor should ensure **all** SelectExpressions are visited, including those nested in a Prediate or OrderBy.

anpete: I think so but add a few tests to be sure. We have ISqlExpressionVisitor but it's kinda awkward right now because we don't have a base impl. and so you end up stubbing all the other methods in a case like this. Also, we are inconsistent in our visitor pattern impl. across SQL expressions. We need to do some refactoring here.

natemcmaster:
@anpete fixed the bug in SelectMany queries. All SQL Server functional tests now pass on SQL 2008 when configured to use legacy paging.

natemcmaster:
Ping for sign off! Rebased to latest changes in dev.

anpete:
:shipit:

Issue #2947, opened 2015-08-27
RevEng: Remove our Razor dependency


divega:
After reviewing the outcome of trying to use Razor to generate C# code in our templates and we have come to the conclusion that it is probably not the right thing to do for the long term. There are several reasons for that, among others:

1. Razor is optimized for HTML generation and a template to generate generating C# code can't be very terse.
2. Razor templates (in particular ours) tend to contain a lot of logic which can be powerful, but we don't believe that having it inside the template necessarily provides the best experience. 

We still want to provide extensibility through some templating language in the future but haven't been able to identify a technology that we can take a dependency on with that purpose in the short term. 

We are then going to focus on having a great default experience, e.g. try to get out default generated code as good as possible for EF 7 RTM, while leaving our choice of a template language open.

That implies sacrificing the ability to customize the output code via templates for RTM and it requires either removing the current Razor-based implementation or turning it into an internal implementation detail.

rowanmiller:
Changing to "remove" rather than "internalize" being an option given https://github.com/aspnet/EntityFramework/issues/3340

lajones:
Removed with PR #3393.

Issue #2956, opened 2015-08-28
RevEng: minimize "public" API surface


rowanmiller:
We should move everything that is not consumed by providers to a .Internal namespace.

The main reason is that we need to design our template-based code gen feature (#2947) and this will have a large impact on the shape of the APIs.

rowanmiller:
@natemcmaster can you just make sure only the API we decided on in the design meeting is left public

PR #3013, opened 2015-09-04
Internalize use of Razor

lajones:
Fix for #2947.

Adding a new CodeWriter abstraction which allows the provider to define their own method of outputting code. A subclass, RazorTemplateCodeWriter, encapsulates what we currently use and is used here by both of our existing providers.
All of the templating code is moved into an Internal namespace.
Removed the ability to output and hence customize templates as the dependency on Razor is being internalized. Instead created a flag to be passed in on the command line to control whether to configure the output code using only fluent API.
Also moved the CSharpUtilities class into the Utilities namespace and update usages to use it via DI.

Notes: 
No attempt to make classes non-public yet - that will be tracked by issue 2956.


src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
```diff
@@ -76,67 +65,22 @@ public class ReverseEngineeringGenerator
...
-            return new List<string>
-                {
-                    FileService.OutputFile(projectDir, GetDbContextTemplateFileName(), dbContextTemplate),
-                    FileService.OutputFile(projectDir, GetEntityTypeTemplateFileName(), entityTypeTemplate)
-                };
+            return await CodeWriter.WriteCodeAsync(
+                modelConfiguration, outputPath, dbContextClassName, cancellationToken);
         }
 
         public virtual IModel GetMetadataModel([NotNull] ReverseEngineeringConfiguration configuration)
```
natemcmaster: Can `GetMetadataModel` be made private?

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
```diff
@@ -76,67 +65,22 @@ public class ReverseEngineeringGenerator
...
                 Namespace = @namespace,
+                UseFluentApi = configuration.UseFluentApi,
             };
             var modelConfiguration = _modelConfigurationFactory
                 .CreateModelConfiguration(metadataModel, customConfiguration);
 
             var dbContextClassName = modelConfiguration.ClassName();
-            CheckOutputFiles(configuration.ProjectPath, configuration.RelativeOutputPath, dbContextClassName, metadataModel);
+
+            CheckOutputFiles(configuration.ProjectPath,
```
natemcmaster: Can `CheckOutputFiles` be made private?

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
```diff
@@ -32,30 +30,25 @@ public abstract class RelationalMetadataModelProvider : IDatabaseMetadataModelPr
...
-            typeof(RelationalMetadataModelProvider).GetTypeInfo().Assembly.GetName().Name
-            + ".ReverseEngineering.Templates.EntityTypeTemplate.cshtml";
-
         private readonly Dictionary<EntityType, EntityType> _relationalToCodeGenEntityTypeMap =
             new Dictionary<EntityType, EntityType>();
         private readonly Dictionary<Property, Property> _relationalToCodeGenPropertyMap =
             new Dictionary<Property, Property>();
 
         public virtual ILogger Logger { get; private set; }
+        public virtual CSharpUtilities CSharpUtilities { get; private set; }
```
natemcmaster: Protected?

src/EntityFramework.Relational.Design/ReverseEngineering/CodeWriter.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            Check.NotNull(fileService, nameof(fileService));
+
+            FileService = fileService;
+        }
+
+        public virtual IFileService FileService { get;[param: NotNull] set; }
+        public virtual string FileExtension { get;[param: NotNull] set; } = DefaultFileExtension;
+
+        public virtual List<string> ReadOnlyOutputFiles(
```
natemcmaster: Not sure what ReadOnlyOutputFiles means. Can you clarify what this method is for?

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
lajones: Am planning on going through all the classes / members etc to review private /protected/public when I respond to #2956.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
lajones: See above

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
lajones: See above

src/EntityFramework.Relational.Design/ReverseEngineering/CodeWriter.cs
lajones: Of the output files this class is about to write which a) already exist and b) are read-only (so that we can raise an error message saying they need to be fixed - probably by being checked out of Source Control).

natemcmaster: Ok. It wasn’t clear to me just from the name of the method. Since this is an API providers may implement, can we clarify the behavior and name? Something like `void AssertWriteAccess(IList<string> files)` (with exceptions thrown) or `bool CheckWriteAccess(IList<string> files)`  seems more straightforward IMO

src/EntityFramework.Commands/DatabaseTool.cs
```diff
@@ -40,6 +40,7 @@ public class DatabaseTool
             [NotNull] string projectRootNamespace,
             [NotNull] string projectDir,
             [CanBeNull] string relativeOutputDir,
+            bool useFluentApi,
```
divega: I assume this being false turns DataAnnotations off. I think we always fluent APIs, so the name isn't super clear.

src/EntityFramework.Relational.Design/ReverseEngineering/CodeWriter.cs
lajones: Happy to change the name of the method, but we need to return the list of files. I want this code to be responsible for determining which files are a problem, but the calling code to decide what to do about that (check out the files automatically, throw etc.). How about `IList<string> files CheckWriteAccess(...)` with a comment that says what it does.

src/EntityFramework.Commands/DatabaseTool.cs
lajones: @divega Correct. How about `useFluentApiExclusively`? I use that elsewhere. The problem is that a correct name is `useFluentApiInsteadOfAttributesEvenWhereThereAreAttributesMatchingThatFunctionality` which is obviously ridiculous.

divega: Could we just invert it and say it is `useDataAnnotations`? You will always have to pass this parameter, so I don't think the fact that using DataAnnotations is the default for the high level experience matters here.

divega: I saw the command line only accepts just `-UseFluentApi`, as opposed to `-UseFluentApi=true` or `-UseFluentApi=false`. Assuming that this is by design, I see why this can't be `useDataAnnotations`. 
Then `useFluentApiOnly` (for here and the command line) sounds better to me than `useFluentApiExclusively`. Does it sound ok to you from the English perspective?

src/EntityFramework.Relational.Design/ReverseEngineering/CodeWriter.cs
divega: `GetReadonlyOutputFiles(...)`?

src/EntityFramework.Commands/DatabaseTool.cs
rowanmiller: I don't think we really need to append `exclusively` or `only`, we use annotations by default so you are really just telling us that you prefer fluent. The other options is a config flag like `-configStyle fluent` `-configStyle annotations`

rowanmiller: If we just have the flag to swap to fluent though... then maybe just `-fluent` is enough

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
divega: FWI, isn't this the entry point into the provider that the code generation invokes? If yes, I don't think it can be private.

src/EntityFramework.Relational.Design/ReverseEngineering/CodeWriter.cs
natemcmaster: Above I was suggestion also changing the behavior, giving CodeWriter and not Generator the responsibility to ensure file paths are accessible. Preserving existing behavior, a better method name may be `GetInaccessibleFilePaths()`.

src/EntityFramework.Commands/DatabaseTool.cs
lajones: Yep - that's by design (matches e.g. usage of `-Idempotent` for migrations). `-UseFluentApiOnly` sounds good to me (and is shorter). So then I'll use `useFluentApiOnly` as the variable name throughout.

src/EntityFramework.Relational.Design/ReverseEngineering/CodeWriter.cs
lajones: Agree we're returning paths - will change that part of name. But I prefer `GetReadOnlyFilePaths()`. We are not returning a list of files which are "inaccessible". I would read that as "cannot be reached" - which could happen for any number of reasons, including the file not currently existing - which is the default state and should return a zero-length list. We are specifically testing that they do exist and are read-only - meaning we won't be able to write to them. We don't test anything else.

src/EntityFramework.Commands/DatabaseTool.cs
lajones: I don't think `-fluent` by itself would be understandable. We know what we mean because we use it all the time, but as a new user I'd ask myself "fluent what"? How about a compromise and have it as `-FluentApi` (and also change `-u` to `-f` in `Program.cs`).

rowanmiller: `-FluentApi` sounds fine to me.

lajones:
@natemcmaster @divega @rowanmiller Comments addressed in latest commit.
@rowanmiller note: have changed the command-line flag from `-u` to `-f`, but just thought that `-f` typically stands for `--force` - which could have a meaning in this context (overwrite the output files if any exist). We don't do that yet, but if you'd prefer to use a different one-letter abbreviation in case we implement that later then let me know.

divega:
Has to be one letter? (excuse my ignorance)

lajones:
@divega Not sure if the API insists on it. But it is the convention - you come up with a double-dash option-name `--exciting-new-option`, plus a single-letter single-dash alternative `-x`.

divega:
I don't have a strong preference but maybe -u was ok.

divega:
Also about this change, how far did you go trying to remove the Razor dependency altogether vs. just internalizing, and are you planning to dig into that more in the future? From what I can see at a glance, it seems that you did not remove the dependency or the template, but may be you included the generator class produced from the template?

lajones:
This is the first step - make it so that a provider does not _have_ to depend on Razor - they are free to write a completely different implementation of CodeWriter. Next step would be to pre-generate / effectively re-implement the classes that Razor produces from the 2 templates. But then we will still need to retain the ability to re-generate those classes when we need to tweak the output. Haven't decided yet whether it's better to retain Razor and its templates under the covers (as a build step similar to how we run Resources.tt to produce .Designer.cs files from .resx files), or just maintain the class(es) that output the code.

lajones:
@divega @rowanmiller @natemcmaster Ping... I can change the single letter command-line option to `-u`. After that can I go ahead and check-in?

divega:
LGTM but @natemcmaster should sign off.

src/EntityFramework.Commands/Program.cs
```diff
@@ -132,6 +132,9 @@ public virtual int Main([NotNull] string[] args)
                             var relativeOutputPath = scaffold.Option(
                                 "-o|--output-path <path>",
                                 "Relative path to the sub-directory of the project where the classes should be output. If omitted, the top-level project directory is used.");
+                            var useFluentApiOnly = scaffold.Option(
+                                "-f|--fluent-api",
```
natemcmaster: Yes, agreeing with others that `-f` is not a good abbreviation.

natemcmaster:
:shipit:

lajones:
Checked in update above with commit 406eb92.

natemcmaster:
Assuming we can close this PR...

PR #3020, opened 2015-09-04
Query compilation improvements:

### LINKED ISSUE #1851: Query: Client eval OrderBy combined with Skip/Take gives wrong results
In the following code the `OrderByDescending` is client eval'd but the `Take` is still pushed to the database. This means you get the first 4 records from the table sorted by savings, rather than the 4 records with the biggest savings.

If we need to do client eval for an `OrderBy`, `Where`, etc. then we probably need to do the `Take` on the client too... or we could throw because perf is going to be bad sicne you'd have to pull back every record from the table.

```
            var bestDeals = db.Products
                .OrderByDescending(p => CalculateSavings(p.MSRP, p.CurrentPrice))
                .Take(4);

        private decimal CalculateSavings(decimal msrp, decimal current)
        {
            return msrp - current;
        }
```

anpete:
- No attempt is now made to ever partially eval a method call.
- Use re-linq's new EvaluatableExpressionFilter to control partial eval.
- Removed FunctionEvaluationDisablingVisitor.
- Simplified re-linq usage in QueryCompiler.

Fix #1851


anpete:
@maumar @mikary

maumar:
:shipit:

mikary:
LGTM

src/EntityFramework.Core/Query/ExpressionVisitors/ParameterExtractingExpressionVisitor.cs
```diff
@@ -83,7 +82,7 @@ public override Expression Visit(Expression expression)
                 }
             }
 
-            if (!typeof(IQueryable).GetTypeInfo().IsAssignableFrom(e.Type.GetTypeInfo()))
+            if (!typeof(IQueryable).IsAssignableFrom(e.Type))
```
natemcmaster: Makes the compiler unhappy. Will revert...

anpete: Thanks!

PR #3021, opened 2015-09-04
Update DetectChanges to bring in all detected dependents

ajcvickers:
This matches the semantics of Add making the behavior more consistent across the stack and, based on our design meeting discussion, more likely to be what is expected.


bricelam:
:shipit:

PR #3022, opened 2015-09-04
Commands: Version handler interfaces

### LINKED ISSUE #2512: Consider upgrade scenarios
We have a few places where artifacts from a previous version will be used by a newer version. We should think carefully about these scenarios and make sure we have the right information available to compensate for forward-incompatible changes. Here are a few that come to mind.
* Migration classes
* Model snapshots
* Migration history table
* NuGet command handlers

Here is some information that we might need.
* Version of Entity Framework
* Version of the provider

bricelam:
This allows us to add additional, unforeseen members to them in the future while maintaining backwards-compatibility. This would be needed after upgrading but before restarting VS, or in a solution that contains multiple versions of EF.

Resolves #2512 (other cases already accounted for)


ajcvickers:
:shipit:

PR #3024, opened 2015-09-04
Errors and data annotations

### LINKED ISSUE #2829: DataAnnotations - Decide when to throw or ignore on various errors
At present we have 2 different behavior in system on how to deal with erroneous DataAnnotations. Sometimes we throw exception and sometimes we log error (presently marked as todo since we don't have logger fully implemented yet).

In following cases we are throwing exception:
- `Timestamp` applied to non-byte array type.
- `Key` applied to multiple properties and key has not been set using fluent API. We do not allow Composite PK with annotation due to ordering.
- `InverseProperty` throws if the inverse property does not exist or of wrong type or self-reference.

In follow cases we are logging error:
- `InverseProperty` attributes present on both navigation and do not points to each other.
- `ForeignKey` on property on both the entityType. Ambiguous to determine the principal end.
- `ForeignKey` is on both navigations or navigation and property and the values do not match.
- `ForeignKey` is on multiple properties (possibly composite FK)
- `ForeignKey` value is invalid property list.
- `Key` in derived type

We should review above and aim for consistency or best user experience while using data annotation.

smitpatel:
resolves #2829

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations.Schema;
 using System.Linq;
 using System.Reflection;
+using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Metadata.Internal;
 using Microsoft.Data.Entity.Utilities;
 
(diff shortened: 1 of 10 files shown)


src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -123,13 +104,45 @@ public virtual InternalRelationshipBuilder Apply(InternalRelationshipBuilder rel
...
+        private void SplitNavigationsInSeparateRelationships(InternalRelationshipBuilder relationshipBuilder)
+        {
+            var foreignKey = relationshipBuilder.Metadata;
+            var principalTypeNavigationName = foreignKey.PrincipalToDependent?.Name;
+            var dependentTypeNavigationName = foreignKey.DependentToPrincipal?.Name;
+
+            var dependentEntityTypebuilder = relationshipBuilder.ModelBuilder.Entity(foreignKey.DeclaringEntityType.Name, ConfigurationSource.Convention);
+            var removedConfigurationSource = dependentEntityTypebuilder.RemoveRelationship(foreignKey, ConfigurationSource.Convention);
+
+            if (removedConfigurationSource == null)
```
AndriySvyryd: This isn't the right condition, as the relationship might have been configured explicitly. You need to check for the presence of conflicting annotations

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -146,17 +159,28 @@ private string FindCandidateDependentPropertyThroughEntityType(EntityType entity
...
                 {
                     candidateProperties.Add(propertyInfo.Name);
                 }
             }
+
             if (candidateProperties.Count > 1)
             {
-                // TODO: Log error that multiple ForeignKey Attribute pointing to same Navigation found.
-                return null;
+                // Composite FK on property
```
AndriySvyryd: Redundant comment

src/EntityFramework.Core/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -146,17 +159,28 @@ private string FindCandidateDependentPropertyThroughEntityType(EntityType entity
...
+                throw new InvalidOperationException(Strings.CompositeFkOnProperty(navigationName, entityType.Name));
+            }
+
+            if (candidateProperties.Count == 1)
+            {
+                var fkAttributeOnNavigation = GetForeignKeyAttribute(entityType, navigationName);
+                if (fkAttributeOnNavigation != null && fkAttributeOnNavigation.Name != candidateProperties.First())
+                {
+                    // FK on property and navitation don't point to each other
```
AndriySvyryd: Redundant comment

smitpatel:
updated

AndriySvyryd:
:shipit:

PR #3027, opened 2015-09-04
Remove all async database creation from tests

ajcvickers:
And thereby stop using AsyncLock in a way that was causing deadlocks.


divega:
:shipit:

PR #3028, opened 2015-09-04
Commands: Print stack traces by default for unexpected exceptions

### LINKED ISSUE #1044: Commands: Review what info we print on error
Currently we just print the exception message in as an error and that is all. For some messages this kind of works but for some it's not clear:

```
PM>  Apply-Migration 
Keyword not supported: 'databaseabc'.
```

In the EF6 we wrote out Exception.ToString() and then the exception message as an error. We could either do this, or add some context to the start ```ArgumentException occurred while applying migrations: Keyword not supported: 'databaseabc'.``` and tell folks to use ```-verbose``` for more info.

In other cases we also print the stack trace when it's not needed (which detracts from the useful info):

```
System.InvalidOperationException: More than one DbContext was found. Specify which one to use.
   at Microsoft.Data.Entity.Commands.ContextTool.SelectType(IEnumerable`1 types, String name)
   at Microsoft.Data.Entity.Commands.MigrationTool.GetContextType(String name)
   at Microsoft.Data.Entity.Commands.MigrationTool.GetMigrations(String contextTypeName)
   at Microsoft.Data.Entity.Commands.Program.ListMigrations(String context)
   at Microsoft.Data.Entity.Commands.Program.<>c__DisplayClass8_3.<Main>b__10()
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.Data.Entity.Commands.Program.Main(String[] args)
```

As part of this we should also make sure we print solutions to the problems. For example, in the above error message there is no info on how to specify which context to use.

bricelam:
Resolves #1044
Depends on #3022

### DIFF
--- a/src/EntityFramework.Commands/CommandException.cs
+++ b/src/EntityFramework.Commands/CommandException.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Commands
+{
+    public class CommandException : Exception
+    {
+        public CommandException()
+        {
+        }
+
+        public CommandException([NotNull] string message)
+            : base(message)
+        {
+        }
+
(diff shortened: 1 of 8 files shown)


ajcvickers:
:shipit:

Issue #2915, opened 2015-08-24
Move classes that we don't expect app devs or provider writers to use to .Internal


rowanmiller:
During API review we decided to be more aggressive about moving classes to `.Internal` namespaces when we don't expect folks to use them (rather than only moving things that we don't want to tidy up yet).

This reduces our "public" surface and allows us to focus on reviewing, documenting, avoiding breaking changes, etc. for the APIs that actually matter.

ajcvickers:
I made an initial pass at this in the reference PR which is now merged. There may be other classes in query, update pipeline, or Migrations that could be moved. Assigning to @bricelam first to have a look at Migrations.

bricelam:
This was done for Migrations in 5df22fa72e55d125528151344cadc26b8a64cefa. Assigning to @AndriySvyryd for Update.

AndriySvyryd:
All classes currently in Update are used by providers. Assigning to @anpete for Query

PR #3031, opened 2015-09-05
Move classes that we don't expect app devs or provider writers to us to .Internal

ajcvickers:
Based on API review discussion; see #2915.

This is a first pass through. Others may want to look again at:
- Query (@anpete)
- Migrations (@bricelam)
- Update (@AndriySvyryd)


anpete:
:shipit: Query.

PR #3073, opened 2015-09-09
[TINY] Fix SaveChanges init benchmark

rowanmiller:
Test would previously insert an entity type that used IDENTITY in the
database. Even though we roll the transaction back the value still gets
reserved in the database and so we were running out of IDENTITY values
after several test runs (the column is `smallint`, so this happened
after a few days). Swapping to use an entity with a non-generated key.


ajcvickers:
:shipit:

PR #3075, opened 2015-09-09
[TINY] "Cold Hard DNX" Disable cold start on DNX451

rowanmiller:
Getting the application assembly loaded into a new AppDomain is just
difficult on DNX because they are in-memory. Filed
aspnet/dnx#2663 to see if this can be made
easier.

In the meantime, just disabling the cold start tests on DNX451.


bricelam:
:shipit:

davidfowl:
I would just launch another process

PR #3079, opened 2015-09-09
Had cascade delete flag to core metadata and fluent API

ajcvickers:
Maintain fluent API from old stack; enum flag in core metatdata to allow for additional states in the future.


bricelam:
:shipit:

Issue #3017, opened 2015-09-04
Use enum instead of bool of Add/Attach behavior


ajcvickers:
From Diego:

I would like to propose we introduce an enum type instead of a bool, e.g.: 

``` C#
 enum GraphOperationBehavior {IncludeDependents, SingleObject};
``` 
(This is the best name I have, but probably makes sense to keep iterating over it :smile:)

__Pros:__
1. Leads to more readable code when you use it
2. Allows for easily exposing new behaviors in the future without breaking changes
__Cons:__
1. A bit more complicated than a bool?

Another option that can be combined with the enum or not: Move to a context-level property instead of a parameter:
__Pros:__
1. Keeps `Add()` and `Attach()` simpler and AddRange() more consistent 
2. If your application works best with a specific behavior, you only need to make the choice once
__Cons:__
1. A bit less discoverable
2. You need to remember to switch back when you wanted a behavior only to apply only to a few calls (we have been afraid of that in the past, but in most cases it turned out well)

Note that this would look similar to what we decided to do for the [context-level default query tracking behavior property](https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes---August-6,-2015#default-query-tracking-behavior) although in this case I am proposing we get rid of the fine grained per-call option.

ilmax:
What about using a context-level property representing the current context behavior (calling Add, AddRange without specifying the behavior, will use the context level property) and add an overload on Add, AddRange, ... Methods to pass a parameter for override the default behavior.

PR #3083, opened 2015-09-09
Use enum instead of bool for Add/Attach behavior

ajcvickers:
See issue #3017


divega:
:shipit: provided that @rowanmiller is happy with the name of the enum GraphOperationBehavior. Personally I wish it could be something less verbose, maybe just GraphBehavior? :smile:

rowanmiller:
Either one is fine by me :smile:

PR #3096, opened 2015-09-10
Fix #3036 - Enforce strong ordering for GroupJoin queries

### LINKED ISSUE #3036: Group joins and ordering
Hunting down the last remaining failing tests in the Npgsql provider, I came across `GroupJoin_customers_orders_count`. This test executes the following LINQ:

```c#
from c in cs
join o in os on c.CustomerID equals o.CustomerID into orders
select new { cust = c, ords = orders.Count()
```

Which causes the following SQL to be sent:

```sql
SELECT "c"."CustomerID", "c"."Address", "c"."City", "c"."CompanyName", "c"."ContactName", "c"."ContactTitle", "c"."Country", "c"."Fax", "c"."Phone", "c"."PostalCode", "c"."Region", "o"."OrderID", "o"."CustomerID", "o"."EmployeeID", "o"."OrderDate"
        FROM "Customers" AS "c"
        LEFT JOIN "Orders" AS "o" ON "c"."CustomerID" = "o"."CustomerID"
```

While this looks OK, but an `ORDER BY` on `CustomerID` is missing in the SQL: in PostgreSQL the rows come back in random order, making multiple rows for the same customer arrive in totally different areas of the resultset. This makes the grouping operation in `QueryMethodProvider._GroupJoin` fail.

I'm not sure why exactly this test passes for SqlServer and Sqlite, could be because they internally maintain insertion order whereas PostgreSQL doesn't.

My initial reaction was to simply add an `order by c.CustomerID` to the LINQ. This makes the grouping work, but there's something deeper here: shouldn't every GroupJoin operation implicitly order by the parent scope's key (or rather, by the column being used to join)?

(As a complete sidenote, should this specific query be grouped in SQL since it uses count? Or is this not yet supported?)

anpete:


anpete:
@maumar

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -841,7 +841,7 @@ public override void Take_Skip_Distinct()
                 Sql);
         }
 
-        public void Skip_when_no_order_by()
+        public void Skip_When_No_OrderBy()
```
maumar: Skip_when_no_OrderBy ?

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -2174,7 +2174,8 @@ public override void GroupJoin_DefaultIfEmpty()
             Assert.Equal(
                 @"SELECT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region], [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
 FROM [Customers] AS [c]
-LEFT JOIN [Orders] AS [o] ON [c].[CustomerID] = [o].[CustomerID]",
+LEFT JOIN [Orders] AS [o] ON [c].[CustomerID] = [o].[CustomerID]
```
maumar: would be also nice to add a case with composite key, not sure if we have one in Northwind, but there is one for sure in GearsOfWar

maumar:
:shipit:

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -444,6 +444,22 @@ var joinExpression
 
                         joinExpression.Predicate = predicate;
 
+                        if (operatorToFlatten == LinqOperatorProvider.GroupJoin)
+                        {
+                            var outerEntityType = Model.GetEntityType(previousQuerySource.ItemType);
```
divega: @anpete and I talked about cases in which the left side can be a projection (from one or more table). He is going to add a few tests to see if we should change this logic.

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
anpete: Changed.

test/EntityFramework.SqlServer.FunctionalTests/QuerySqlServerTest.cs
anpete: Will do.

PR #3100, opened 2015-09-10
Benchmark queries without caching

rowanmiller:
Adding variations of the query tests that disable the query cache. This
gives us an idea of what overhead is involved in query compilation.

For EF7 I'm using a null implementation of IMemoryCache, which means
that we still go through all the locking etc. in CompiledQueryCache.
This gives the best approximation of running a query where it does not
already exist in the cache.

For EF6 we need to use reflection to get the ObjectQuery to disable
caching :facepunch:.


divega:
:shipit:

PR #3108, opened 2015-09-11
Basic cleanup of SQL Server provider namespaces

ajcvickers:


ErikEJ:
@ajcvickers Is this something you suggest other providers to also do?

ajcvickers:
@ErikEJ We are currently doing API review of the provider assemblies. This is the first part of that work, but there are also some other changes coming. It would be a good idea to keep your provider updated so that the APIs align, but it is obviously not strictly necessary.

ErikEJ:
Thanks, I will attempt to align namespaces

bricelam:
:shipit:

rowanmiller:
@ErikEJ our general rule with providers etc. is that we don't introduce a provider specific sub-namespace. Types keep the same namespace as the type they inherit from core (or the related types in core). We do introduce some sub-namespaces when it makes sense (i.e. we introduce `Microsoft.Data.Entity.Migrations` in relational, because there is no migrations concept in core).

rowanmiller:
cc @roji since you may want to apply these same rules in Npgsql

roji:
Thanks @rowanmiller, have applied this in Npgsql

PR #3114, opened 2015-09-12
Query: Relational: Fixes operator precedence bug in SQL gen and some …

anpete:
…formatting improvements.

NB: Query SQL verification has largely been removed from the SQLite tests. The reason for this change is to make it easier to implement formatting improvements to SQL gen. The majority of SQL that we generate for SQLite is structurally equivalent to the TSQL version and only differs in the delimiter char. Where the SQL is different (paging) I have left in a small amount of verification.


anpete:
@maumar

maumar:
:shipit:

PR #3127, opened 2015-09-14
Fix #2941 - Use quoted identifiers in (Async)FromSqlQuery tests.

### LINKED ISSUE #2941: Make FromSqlQuery tests provider-agnostic

mikary:
@anpete


anpete:
:shipit:

Issue #3041, opened 2015-09-06
SqliteMigrationsSqlGenerator doesn't supports One-to-One and One-to-Many relationships to the same entity.


Suchiman:
Using `7.0.0-beta8-15623` when you try to define a One-to-One and a One-to-Many relationship between the same entities you will receive following exception:

```
Unhandled System.NotSupportedException
  HResult=-2146233067
  Message=SQLite cannot support this migration operation.
  Source=EntityFramework.Sqlite
  StackTrace:
       at Microsoft.Data.Entity.Migrations.SqliteMigrationsSqlGenerator.Generate(AddForeignKeyOperation operation, IModel model, SqlBatchBuilder builder)
       at Microsoft.Data.Entity.Migrations.MigrationsSqlGenerator.<>c.<.cctor>b__47_1(MigrationsSqlGenerator g, MigrationOperation o, IModel m, SqlBatchBuilder b)
       at Microsoft.Data.Entity.Migrations.MigrationsSqlGenerator.Generate(MigrationOperation operation, IModel model, SqlBatchBuilder builder)
       at Microsoft.Data.Entity.Migrations.MigrationsSqlGenerator.Generate(IReadOnlyList`1 operations, IModel model)
       at Microsoft.Data.Entity.Storage.RelationalDatabaseCreator.GetCreateTablesCommands()
       at Microsoft.Data.Entity.Storage.RelationalDatabaseCreator.CreateTables()
       at Microsoft.Data.Entity.Storage.RelationalDatabaseCreator.EnsureCreated()
       at Microsoft.Data.Entity.Infrastructure.DatabaseFacade.EnsureCreated()
       at EFTest.Program.Main(String[] args) in c:\users\robin\documents\visual studio 2015\Projects\EFTest\EFTest\Program.cs:Zeile 17.
       at System.AppDomain._nExecuteAssembly(RuntimeAssembly assembly, String[] args)
       at System.AppDomain.ExecuteAssembly(String assemblyFile, Evidence assemblySecurity, String[] args)
       at Microsoft.VisualStudio.HostingProcess.HostProc.RunUsersAssembly()
       at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
       at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[...]

PR #3133, opened 2015-09-14
Lift SQLite foreign key additions into create table statement

### LINKED ISSUE #2590: SQLite Migrations: Lift foreign keys to create table operation
The model differ breaks foreign key cycles in create table operations by pulling one out into a separate add foreign key operation. Unfortunately, SQLite can't add foreign keys to a table without rebuilding it. The SQLite migration SQL generator should push them back into their create table operation since cycles don't matter when foreign keys aren't enforced.

natemcmaster:
Fix #2590 

/cc: @bricelam

### DIFF
--- a/test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
+++ b/test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
@@ -19,6 +19,35 @@ namespace Microsoft.Data.Entity.Migrations
                 new SqliteMetadataExtensionProvider());
 
         [Fact]
+        public virtual void It_lifts_foreign_key_additions()
+        {
+            Generate(new CreateTableOperation
+                {
+                    Name = "Pie",
+                    Columns =
+                        {
+                            new AddColumnOperation
+                                {
+                                    ClrType = typeof(int),
+                                    Name = "FlavorId",
+                                    ColumnType = "INT"
+                                }
+                        }
+                }, new AddForeignKeyOperation
+                    {
+                        Table = "Pie",
+                        PrincipalTable = "Flavor",
+                        Columns = new[] { "FlavorId" },
+                        PrincipalColumns = new[] { "Id" }
+                    });
+
+            Assert.Equal(@"CREATE TABLE ""Pie"" (
+    ""FlavorId"" INT NOT NULL,
+    FOREIGN KEY (""FlavorId"") REFERENCES ""Flavor"" (""Id"")
+);" + EOL, Sql);
+        }
+
+        [Fact]
         public virtual void DefaultValue_formats_literal_correctly()
         {
             Generate(new CreateTableOperation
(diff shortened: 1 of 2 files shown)


test/EntityFramework.Sqlite.FunctionalTests/SqliteDatabaseCreationTest.cs
```diff
@@ -0,0 +1,110 @@
...
+using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Sqlite.FunctionalTests;
+using Microsoft.Data.Entity.Storage;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Logging;
+using Xunit;
+
+namespace Microsoft.Data.Entity.Sqlite
+{
+    public class SqliteDatabaseCreationTest : IClassFixture<SqliteDatabaseCreationFixture>
```
bricelam: No need for a functional test, IMHO. I would just assert the SQL with a new unit test in [SqliteMigrationSqlGeneratorTest](https://github.com/aspnet/EntityFramework/blob/dev/test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs).

bricelam:
:shipit:

test/EntityFramework.Sqlite.FunctionalTests/SqliteDatabaseCreationTest.cs
```diff
@@ -0,0 +1,110 @@
...
+        public void It_lifts_foreign_key_creation_to_create_table_operation()
+        {
+            using (var context = _fixture.CreateContext())
+            {
+                var creator = context.GetService<IDatabaseCreator>() as RelationalDatabaseCreator;
+                creator.CreateTables();
+            }
+
+            Assert.StartsWith("CREATE TABLE ", TestSqlLoggerFactory.Sql);
+            Assert.Equal(2, TestSqlLoggerFactory.SqlStatements.Count);
```
bricelam: This also doesn't assert that the foreign key was actually added...

natemcmaster: Good point. This test used the example from #3041. Will move to a simpler test as you suggested.

test/EntityFramework.Sqlite.Tests/Migrations/SqliteMigrationSqlGeneratorTest.cs
```diff
@@ -19,6 +19,35 @@ protected override IMigrationsSqlGenerator SqlGenerator
...
+                                {
+                                    ClrType = typeof(int),
+                                    Name = "FlavorId",
+                                    ColumnType = "INT"
+                                }
+                        }
+                }, new AddForeignKeyOperation
+                    {
+                        Table = "Pie",
+                        PrincipalTable = "Flavor",
```
bricelam: Self-refs are actually not broken out into their own operations. :smile:

Issue #2994, opened 2015-09-03
Change the semantics of Ignore<TEntity>


AndriySvyryd:
Assuming that the primary function of this call is to affect the behavior of conventions then calling `Entity<TEntity>` before or after `Ignore<TEntity>` should still result in the entity type being added to the model. However if `Ignore` is called it could mean that no conventions will run on that entity type.

AndriySvyryd:
We've decided this isn't a very important scenario and preserving EF6 behavior trumps it.

PR #3137, opened 2015-09-15
 Inherit ignored members

AndriySvyryd:
Allow overriding ignored and ignoring existing members using explicit configuration on same type, base type or derived type
Add a convention trigger for ignoring an entity type member
Make `NotMappedAtrribute` property convention run when an `EntityType` is added
Throw when adding a property with a name that conflicts with a navigation name and viceversa


ajcvickers:
We discussed this when triage for issue #2994. @rowanmiller may have notes on this but I think there was some push back around adding back an entity after it has been ignored. So I think maybe ignore always ignoring was okay with some people (but maybe not @divega) and last call wins (like for all the other APIs) was okay with most people. So basically I don't think we have a decision on what the semantics of Ignore should be.

divega:
I don't have a strong preference on this, and I don't have a strong pushback against what @AndriySvyryd is proposed, but it seems others do. I think we should discuss in person very soon.

rowanmiller:
Yeah in triage we agreed to do either:
* Ignore always means ignored (even if you call some other config)
* Last change wins (you can ignore, but then calling Entity on it would bring it back)

The first is the behavior of EF6, the second is more consistent with how other config works. We decided it didn't really matter too much which one we chose, so I had a note from triage to follow up with @AndriySvyryd to see which was easiest (I haven't got thru all the TODOs from last triage yet).

AndriySvyryd:
Ok, this commit wasn't going to be the final behavior, but I might as well do it if we agree. Just to clarify there are 3 situations when we need to decide what to do with conflicting configurations on entity type:

1. On the same type. Last change wins should be fine here.
1. On derived type. Overriding the derived type configuration should match expectations as well.
1. On base type. In this case ignoring a property that has been explicitly declared on the base will remove it. This might be surprising at first, but does enable scenarios that wouldn't work otherwise.

Note that NotMappedAttribute will still override any other attribute as there is no way to specify attribute order.

I'll modify the PR to implement this unless there is any pushback.

AndriySvyryd:
Updated

ajcvickers:
:shipit:

PR #3139, opened 2015-09-15
Flow cascade delete from metadata into Migrations

ajcvickers:
Also, fix R# block indenting back to match what Visual Studio does so that we don't have so many conflicts between R# formatted code and VS formatted code.

Currently the functional test for this is disabled in SQLite since cascade delete for SQLite does not appear to be working at this time.

### DIFF
--- a/src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
+++ b/src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
@@ -154,7 +154,7 @@ namespace Microsoft.Data.Entity.Migrations.Internal
 
                     var principalCreateTableOperation = createTableOperations.FirstOrDefault(
                         o => o.Name == addForeignKeyOperation.PrincipalTable
-                            && o.Schema == addForeignKeyOperation.PrincipalSchema);
+                             && o.Schema == addForeignKeyOperation.PrincipalSchema);
                     if (principalCreateTableOperation != null)
                     {
                         createTableGraph.AddEdge(principalCreateTableOperation, createTableOperation, addForeignKeyOperation);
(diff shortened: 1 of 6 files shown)


src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -606,7 +604,10 @@ protected virtual IEnumerable<MigrationOperation> Add([NotNull] IForeignKey targ
                 Columns = GetColumns(target.Properties),
                 PrincipalSchema = targetPrincipalEntityTypeAnnotations.Schema,
                 PrincipalTable = targetPrincipalEntityTypeAnnotations.TableName,
-                PrincipalColumns = GetColumns(target.PrincipalKey.Properties)
+                PrincipalColumns = GetColumns(target.PrincipalKey.Properties),
+                OnDelete = target.DeleteBehavior == DeleteBehavior.Cascade
+                    ? ReferentialAction.Cascade
+                    : ReferentialAction.NoAction
```
bricelam: Should this be `Restrict`? (They're the same on all databases but SQLite which will literally take no action, not error)

ajcvickers: I think that is an open question because we have also talked about using SetNull when the FK is not required. So for now I'm just enabling cascade delete and then we can decide what to do about other cases.

bricelam:
:shipit:

src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
divega: Sounds good, but let's note forget about on cascade set null please :smile:

PR #3140, opened 2015-09-15
Fix #3058 - DRY up cache key generation

### LINKED ISSUE #3058: Query: Caching: DRY up cache key generation
We are duplicating the CompiledQueryCacheKey across Core, Relational and SQL Server.

DRY up via annotations, or perhaps just cache key composition.

anpete:


src/EntityFramework.Core/Query/CompiledQueryCacheKeyGenerator.cs
```diff
@@ -21,19 +20,22 @@ public CompiledQueryCacheKeyGenerator([NotNull] IModel model)
...
+            => GenerateCacheKeyCore(query, async);
+
+        protected CompiledQueryCacheKey GenerateCacheKeyCore([NotNull] Expression query, bool async)
             => new CompiledQueryCacheKey(
                 new ExpressionStringBuilder().Build(Check.NotNull(query, nameof(query))),
                 _model,
                 async);
 
-        private struct CompiledQueryCacheKey
+        public struct CompiledQueryCacheKey
```
maumar: move to separate file or unnest from CompiledQueryCacheKeyGenerator

anpete: Prefer to keep this as impl detail so have made it protected.

maumar:
:shipit:

Issue #2498, opened 2015-06-29
Consolidate and rationalize SQL execution


ajcvickers:
Do these need to be shared between Update and Migrations?
Should ExecuteCommand use them?

mikary:
Notes from #2758:
 - Cleanup in Query `DefaultQuerySqlGenerator`, renames based on new `RelationalCommandBuilder`
 - Look at removing `RelationalSqlExecutor` and moving `IRelationalConnection` into `SqlStatementExecutor` constructor dependencies
 - Use DI for `ParameterNameGeneratorFactory` in Query

mikary:
completed and passed through API review

PR #3145, opened 2015-09-15
Move DbCommand execution into RelationalCommand

### LINKED ISSUE #1728: Dispose commands
I'm a core contributor for Npgsql, the open-source PostgreSQL ADO.NET driver.

Following up on [this issue on Codeplex](https://entityframework.codeplex.com/workitem/2306), I do hope that EF7 will properly dispose of commands once they're no longer in use. This triggered a memory leak when using EF6 with Npgsql (which we will work around).

Am cross-posting to Codeplex as well.

### LINKED ISSUE #1863: Add support for execution of a raw command on the data store

### LINKED ISSUE #3002: Possible null ref in RelationalParameter
```
Value.GetType().IsNullableType()
```

mikary:
Fix #3002 - Don't attempt to resolve type on null parameter
Fix #1863 - Add ExecuteSqlCommandAsync
Fix #1728 - Dispose all DbCommands in RelationalCommand

~~Working on fixing Sqlite test breaks resulting from logging "PRAGMA foreign_keys=ON;" commands~~

@bricelam


mikary:
Updated with Sqlite test fixes

src/EntityFramework.Relational/Migrations/HistoryRepository.cs
```diff
@@ -92,11 +92,21 @@ public abstract class HistoryRepository : IHistoryRepository
         protected abstract string ExistsSql { get; }
 
         public virtual bool Exists()
-            => _databaseCreator.Exists() && InterpretExistsResult(_executor.ExecuteScalar(_connection, ExistsSql));
+            => _databaseCreator.Exists() && InterpretExistsResult(
+                _commandBuilderFactory
+                    .Create()
```
maumar: this is a very common pattern of creating commands, consider adding some sugar method for it

mikary:
Also includes changes to `ModificationCommandBatch` to remove `DbContext` parameter as discussed in API review
@AndriySvyryd

src/EntityFramework.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -150,8 +150,7 @@ public GroupJoinAsyncEnumerator(GroupJoinAsyncEnumerable<TResult> groupJoinAsync
                 new AsyncQueryingEnumerable(
                     ((RelationalQueryContext)queryContext),
                     commandBuilder,
-                    /*queryIndex*/ null,
-                    queryContext.Logger)
+                    /*queryIndex:*/ null)
```
maumar: you can remove /* */ now

src/EntityFramework.Relational/Storage/RelationalCommand.cs
```diff
@@ -1,41 +1,124 @@
...
+                    {
+                        Logger.LogCommand(dbCommand);
+                    }
+
+                    return action(dbCommand);
+                }
+            }
+            finally
+            {
+                connection.Close();
```
ajcvickers: So does this mean that the command is always disposed and the connection is closed before the reader has been consumed? Does this always work?

src/EntityFramework.Relational/Query/Internal/AsyncQueryingEnumerable.cs
```diff
@@ -61,21 +57,17 @@ public AsyncEnumerator(AsyncQueryingEnumerable queryingEnumerable)
...
-                        {
-                            _queryingEnumerable._logger.LogCommand(command);
-
-                            await _queryingEnumerable._relationalQueryContext
-                                .RegisterValueBufferCursorAsync(this, _queryingEnumerable._queryIndex, cancellationToken);
+                        await _queryingEnumerable._relationalQueryContext
+                            .RegisterValueBufferCursorAsync(this, _queryingEnumerable._queryIndex, cancellationToken);
 
-                            _dataReader = await command.ExecuteReaderAsync(cancellationToken);
+                        _dataReader = await _queryingEnumerable._commandBuilder
```
maumar: i would break it down into two calls, since Build is synchronous

src/EntityFramework.Relational/Update/ReaderModificationCommandBatch.cs
```diff
@@ -143,88 +136,59 @@ protected virtual void UpdateCachedCommandText(int commandPosition)
             }
         }
 
-        public override void Execute(
-            IRelationalTransaction transaction,
-            IRelationalTypeMapper typeMapper,
-            DbContext context,
-            ILogger logger)
+        public override void Execute([NotNull] IRelationalConnection connection)
```
ajcvickers: Consider the code here. Previously we were creating the command, creating the reader, consuming the reader, disposing the reader, disposing the command.  Now we are creating the command, creating the reader, disposing the command, consuming the reader, disposing the reader.

src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
```diff
@@ -16,18 +16,19 @@ namespace Microsoft.Data.Entity.Sqlite
 {
     public class SqliteDatabaseConnection : RelationalConnection
     {
-        private readonly IRelationalTypeMapper _typeMapper;
+        private readonly IRelationalCommandBuilderFactory _relationalCommandBuilderFactory;
         private readonly bool _enforceForeignKeys = true;
+        private int _openedCount;
```
ajcvickers: It seems unfortunate that the derived class is now maintaining its own reference count.

src/EntityFramework.Relational/Storage/RelationalCommand.cs
mikary: yes, it seems this results in the command being disposed and the connection being closed before the reader is consumed. This works in the existing code because the calls happen to be made while the connection is already open, but even this behavior isn't necessarily guaranteed with all providers.

We could work around this sequencing issue if `ExecuteReader` takes a consume action as a parameter.

src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
mikary: This does resolve an existing issue where the Sqlite connection would send a command to enable foreign keys whenever `Open` is called, regardless of the state of the existing connection.

We considered a few different options for solving this problem:
* Check the `DbConnection.State` when `Open` is called and only enable keys when the previous state is `Closed` - Does not behave properly with a user owned connection that is already open
* Add lifetime events / template methods to `RelationalCommand` (i.e. `OnOpen`, `OnClose`) to accommodate these kinds of actions - Only required by Sqlite for now, YAGNI
* Count open/close calls in the `SqliteDatabaseConnection` - Duplicates logic in `RelationalCommand`

This seemed like the most pragmatic solution for the time being, but there is defiantly merit to using an alternate solution.

divega: Can the current count just be exposed from the base class? Open is already virtual.

src/EntityFramework.Relational/Storage/RelationalCommand.cs
mikary: The factoring for Query would be a bit strange with a consume action, as the existing behavior disposes the command and saves the reader for later iteration.

ErikEJ:
Providers beware? :smile:

src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
ajcvickers: What is the semantics of this supposed to be? Do we want the user to be able to to control this in any way? For example, if the user gives us an open connection with foreign_keys off will we always switch it on? If yes, then should we switch it off again when done or just leave their open connection with it switched on? What if they give us a connection with it on already? Do we just try to switch it on again anyway even if it is already on? Is the connection even the right object to use for this?

src/EntityFramework.Relational/Storage/RelationalCommand.cs
ajcvickers: I know that in the old stack when we attempted to retro-fit disposing of the command in this way it started to cause some things to fail. In other words, the old stack didn't ever dispose the command, but changing the code to dispose it before the reader was fully consumed caused issues--I just don't remember exactly what those issues were. Maybe this isn't an issue on the new stack for any providers we care about.

anpete: Return a reader decorator that takes care of disposing the connection?

Even better return an IEnumerable of reader similar to what we do in query.

mikary:
Yes, @ErikEJ very much so :)

src/EntityFramework.Relational/Storage/RelationalCommand.cs
mikary: I do like the idea of a reader decorator. I didn't add one already because I've been trying to avoid needlessly wrapping the ADO objects.

In some of the prototyping I've done, I was also able to use a wrapped `DbDataReader` to transparently handle MARS. (not something we need now, but potentially useful)

anpete: We also need to be careful with the perf of this code path in query

ajcvickers: I'm a bit concerned that we will end up over-wrapping and over-abstracting and we won't end up with something that is significantly easier to use or maintain and may result in worse perf.

mikary: While it doesn't necessarily address the dispose issue, should we consider removing the connection `Open`/`Close` calls out of this class and assume that part of the contract for ExecuteXxx is that the provided connection is open?

src/EntityFramework.Relational/Query/QueryMethodProvider.cs
```diff
@@ -100,8 +100,7 @@ private static TResult GetResult<TResult>(IEnumerable<ValueBuffer> valueBuffers)
             => new QueryingEnumerable(
                     ((RelationalQueryContext)queryContext),
                     commandBuilder,
-                    /*queryIndex*/ null,
-                    queryContext.Logger)
+                    /*queryIndex:*/ null)
```
maumar: remove /* */

src/EntityFramework.Relational/Storage/RelationalCommand.cs
```diff
@@ -1,41 +1,124 @@
...
+            [NotNull] IRelationalConnection connection,
+            [NotNull] Func<DbCommand, T> action)
+        {
+            connection.Open();
+
+            try
+            {
+                using (var dbCommand = CreateCommand(connection))
+                {
+                    if (Logger.IsEnabled(LogLevel.Verbose))
```
maumar: Logger.LogVerbose(() => ...)

src/EntityFramework.Sqlite/SqliteDatabaseConnection.cs
mikary: Any thoughts on this @bricelam @natemcmaster ?

natemcmaster: We should use FK enforcement by default for new and existing connections. Users can opt-out with `.UseSqlite().SuppressForeignKeysEnforcement()`.

For existing connections, we should still ensure FK enforcement matches what the user has specified by calling `PRAGMA foreign_keys=OFF/ON` on first use.

IMO - many users will pass us a connection wanting FK on, but not going to the effort to turn it on themselves. For users who want FK off, they can still use `SuppressForeignKeysEnforcement` to make sure FK stays off.

src/EntityFramework.Relational/Migrations/HistoryRepository.cs
mikary: I think we should hold off on this until the factoring of the commands and the consuming classes stabilize.
Adding a note to #2498 to add sugar methods

mikary:
Closing to rework ExecuteReader API surface for Open/Close/Dispose ordering

Issue #2357, opened 2015-06-09
Decouple design time DbContext instantiaton from migrations commands


divega:
Currently EntityFramework.Commands contains what we believed to be the implementation of EF command-line experience for DNX. However, the way we instantiate the application's DbContext (with all its dependencies wired up the same way they would be at runtime) is very ASP.NET specific and not generally applicable to all DNX code. 

Issue aspnet/Hosting#286 was about decoupling that mechanism from ASP.NET Hosting but some of us met to analyze that option and came to the conclusion that that isn't the right approach. This issue describes a different way things can be decoupled: 

EntityFramework.Commands should depend on an implementation of a contract that can load a context (we should considering keeping the shape and name of IDbContextFactory from EF6 or we can come up with something new). There could be several implementations of this interface, e.g.:
1. A standard one that mediates between ASP.NET Hosting and EF commands. It should live in its own package, e.g. EntityFramework.Commands.AspNet.
2. Potentially other standard ones for non-ASP.NET DNX applications 
3. User defined (as with the old IDbContextFactory).

This "loader" would be specified in the registration of the EF commands in the project.json file, e.g.:

``` JSON
"commands": {
    "ef":  "EntityFramework.Commands --loader Microsoft.Data.Entity.Commands.AspNetContextLoader"
},
```

We can consider falling back to other behaviors in absence of this setting.

divega:
cc @bricelam @Tratcher

vossad01:
Traditional .NET (non-DNX) support is relevant for us.  No issues with `project.json`, just wasn't sure if it would be read in a `.csproj` project.

http://stackoverflow.com/q/29110241/1072626

XcoderTeam:
What is the current workaround, after I upgrade from beta5 to beta6, I unable to do migration. on the libraries that not having "Startup" class.

rowanmiller:
This goes away based on our design meeting this morning

rowanmiller:
This is tracked by several other issues too

PR #3151, opened 2015-09-16
DNX Commands: Use development environment

bricelam:
Fixes aspnet/dnx#2706

P.S. Fleshing out how we discover services and select an environment is part of #2357


ajcvickers:
:shipit:

kscott5:
@bricelam 

How does this impact "dnx ef database update"? How do you set the aspnet_env from command-line, hosting.json or project.json?

currently its defaulting to production regardless of what I set or throws an error.

PR #3153, opened 2015-09-16
Fix #3142 - Make FromSql compose on Projections

### LINKED ISSUE #3142: FromSql: Not composed correctly when projection
Given:

```c#
var productsFromSql
    = context.Set<Product>()
        .FromSql(@"SELECT *
                FROM Products 
                WHERE Discontinued <> 1 
                AND ((UnitsInStock + UnitsOnOrder) < ReorderLevel)")
        .Select(p => p.ProductName);
```
we send this to the store:
```sql
SELECT*
FROM Products
WHERE Discontinued <> 1
AND((UnitsInStock + UnitsOnOrder) < ReorderLevel)
```
I.e. There is not outer SELECT to affect the projection.

mikary:
~~Fixing the issue and adding the missing Sqlite test cases~~
@anpete


anpete:
:shipit:

PR #3166, opened 2015-09-17
Commands: Cleanup design-time service building (and logging)

### LINKED ISSUE #1652: Failure message when migrations are not available would benefit from improvement
When migrations are not available because of the data store (provider) configuration, the failure message is not very helpful in indicating the issue.  Ideally the failure message when migrations are not available would communicate the failure is because of the data store configuration and provide hints/suggestions as to how to go about correcting the issue.

Given a project with only the following classes:

```c#
public class BlogDbContext : DbContext
{
    {
        modelBuilder.Entity<Blog>();
    }
}

public class Blog
{
}
```
An attempt to create a migration with `Add-Migration` fails with the following message:

> Unable to resolve service for type 'Microsoft.Data.Entity.Commands.Migrations.MigrationScaffolder' while attempting to activate 'Microsoft.Data.Entity.Relational.Migrations.Infrastructure.MigrationAssembly'.

Desired behavior might be a message along the lines of (example text only):

> Migrations are unavailable because the configured data store does not support Migrations.

or
> Migrations are unavailable because no data store configuration was found.

Either better conveys the issue in form the user's perspective. Information about how/where configuration can be done would make them even better.  There might also be value in keeping the technical information as well for debugging and such?

bricelam:
Fixes #1652, part of #1769


src/EntityFramework.Commands/Design/Internal/CommandLogger.cs
```diff
@@ -28,7 +29,10 @@ public abstract class CommandLogger : ILogger
             typeof(DbContextOperations).FullName,
             typeof(MigrationsOperations).FullName,
             typeof(MigrationsScaffolder).FullName,
-            typeof(Migrator).FullName
+            typeof(Migrator).FullName,
+            "Microsoft.Data.Entity.Relational.Design.ReverseEngineering.ReverseEngineeringGenerator",
+            "Microsoft.Data.Entity.Relational.Design.ReverseEngineering.RelationalMetadataModelProvider",
+            "Microsoft.Data.Entity.SqlServer.Design.Utilities.SqlServerLiteralUtilities"
         };
```
lajones: Why not use typeof() for these 3 lines as above?

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
```diff
@@ -21,21 +21,21 @@ public class ReverseEngineeringGenerator
         private readonly IDatabaseMetadataModelProvider _provider;
 
         public ReverseEngineeringGenerator(
-            [NotNull] ILogger logger,
+            [NotNull] ILoggerFactory loggerFactory,
             [NotNull] IFileService fileService,
```
lajones: Why are you changing this to be an ILoggerFactory rather than an ILogger? AFAIK all RevEng needs is a single logger and passing in an ILoggerFactory makes it look like it's going to need several.

bricelam: ILogger should not be registered in DI. This is how logging was designed to be used.

src/EntityFramework.Commands/Design/Internal/CommandLogger.cs
bricelam: That assembly is not referenced by this project. I've been thinking about alternative ways to do this including ones that don't use Logging.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
divega: AFAIK that evolved and if you ask for an ILogger you will get one for your class.

bricelam: Ok, I'll try to update these to depend on `ILogger` directly again.

bricelam:
Updated to decouple `CommandLogger` from the types that log. This required explicitly calling `ILoggerFactory.CreateLogger()`.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs
bricelam: Nevermind, I decided it was cleaner to use a well-known string, not the class names.

lajones: Hmmm. Still seems a bit odd IMHO. 1) could we define "EntityFramework.Commands" as a variable somewhere and use that throughout? But 2) it still seems odd to me that e.g. `RelationalMetadataModelProvider` and `SqlServerLiteralUtilities` both create a new `ILogger`, (even if they're both created with the same name) when AFAIK we have no reason for them not to be the same logger. It feels like something the calling code should deliver as a service?

natemcmaster: Using `ILoggerFactory` instead of sharing `ILogger` is a better pattern. This allows us greater flexibility when logging. It also makes it clearer inside the class which log it is going to.

Agreeing with @lajones - let's try to use a const string. It makes refactoring easier. We could implement via shared .cs file.

bricelam: @lajones `serviceContainer.AddLogging()` does not add an `ILogger` to the service provider. It has a bit of magic to create `ILogger` objects on the fly from the `ILoggerFactory`, but there is no way to specify the name if we do that.

@natemcmaster I was also thinking a shared `*.cs` could help.

lajones: After discussion looks like we're stuck with this approach to logging for now. :shipit: after you create the shared string.

PR #3167, opened 2015-09-17
[SMALL] Fix to 3163 - include doesn't work well with inheritance

maumar:
Problem was that during binding we were assuming that memberExpression.Expression would always be QuerySourceReferenceExpression. However in some cases (inheritance related) it is wrapped in a convert.
Fix is to add a fallback mechanism to look for this pattern as well

### DIFF
--- a/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
@@ -1105,10 +1105,13 @@ namespace Microsoft.Data.Entity.Query
 
                 properties.Add(property);
 
-                querySourceReferenceExpression
-                    = memberExpression.Expression as QuerySourceReferenceExpression;
+                querySourceReferenceExpression =
+                    memberExpression.Expression as QuerySourceReferenceExpression
+                    ?? (memberExpression.Expression as UnaryExpression)?.Operand as QuerySourceReferenceExpression;
 
-                memberExpression = memberExpression.Expression as MemberExpression;
+                memberExpression =
+                    memberExpression.Expression as MemberExpression
+                    ?? (memberExpression.Expression as UnaryExpression)?.Operand as MemberExpression;
             }
 
             return querySourceReferenceExpression != null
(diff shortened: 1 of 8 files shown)


maumar:
@anpete

test/EntityFramework.Core.FunctionalTests/TestModels/GearsOfWarModel/Gear.cs
```diff
@@ -26,6 +26,8 @@ public Gear()
 
         public MilitaryRank Rank { get; set; }
 
+        // TODO: temporary hack
+        [NotMapped]
```
maumar: currently there is bug here - we conventions discover this on officer first, then we add explicit definition in OnModelCreating, which leads to Tags being declared on both Gear and Officer, resulting in exception

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -1105,10 +1105,13 @@ var property
 
                 properties.Add(property);
 
-                querySourceReferenceExpression
-                    = memberExpression.Expression as QuerySourceReferenceExpression;
+                querySourceReferenceExpression =
```
anpete: Can you just use the RemoveConvert extension method?

anpete:
:shipit:

test/EntityFramework.Core.FunctionalTests/TestModels/GearsOfWarModel/Gear.cs
divega: > currently there is bug here - we conventions discover this on officer first, then we add explicit definition in OnModelCreating, which leads to Tags being declared on both Gear and Officer, resulting in exception

@maumar Is the bug filed? Sorry if I missed it.

test/EntityFramework.Core.FunctionalTests/TestModels/GearsOfWarModel/Gear.cs
```diff
@@ -26,6 +26,8 @@ public Gear()
 
         public MilitaryRank Rank { get; set; }
 
+        // TODO: temporary hack, remove once #1704 is fixed
+        [NotMapped]
```
maumar: @divega I talked with Andriy about this, and he is aware of this and was apparently working on this before the perf issue took precedence. Not sure what is the bug number though

maumar: @divega bug number is 1704

divega: Thanks!

PR #3175, opened 2015-09-19
Improve ModelBuilder perf

### LINKED ISSUE #3097: Perf: ModelBuilder bottleneck in ConventionDispatcher.OnEntityTypeAdded
Test `EntityFramework.Microbenchmarks.InitializationTests.BuildModel_AdventureWorks` takes around 3x longer than the EF6 equivalent. Over half the CPU time of the timed section of the test is spent in `ConventionDispatcher.OnEntityTypeAdded`

profile

The call stacks get very deep, but if you drill down through the call stack shown in the profile and then another branch that accounts for another chuck of the time in the test then they both have `RelationshipDiscoveryConvention.Apply` in the stack and adding up the numbers gives ~75% of time spent in that method. Of course, they keep going much deeper than that, so it's more complex than just optimizing the body of that method.

AndriySvyryd:
By performing entity type culling once per model and storing derived types in EntityType

Fixes #3097


smitpatel:
:shipit:

PR #3182, opened 2015-09-21
Add validation for null PK

### LINKED ISSUE #3109: [ModelValidator] - Check for entity types without Primary Keys
Presently entities without PK can exist in valid model.

smitpatel:
resolves #3109


AndriySvyryd:
:shipit:

Issue #3164, opened 2015-09-17
SQLite RevEng: Context is always named ModelContext


bricelam:
By default, it should be named `<database-name>Context`.

lajones:
@bricelam. @natemcmaster and I talked this over. A typical `database-name` for SQLite is just a file path. That's what you get if you put the connection string into SqliteConnectionStringBuilder and ask for DataSource (which is similar to what we do for SqlServer). We could just strip off all directories and just make a CSharp identifier out of the name of the file. Would that work?

divega:
FWIW, works for me: just the file name with no path or extensions + Context.

lajones:
Fix checked in with commit 9d69d87.

PR #3189, opened 2015-09-22
Reveng: Sqlite context was always ModelContext

lajones:
Fix for issue #3164.


natemcmaster:
:shipit:

lajones:
Checked in with commit 9d69d87.

PR #3190, opened 2015-09-22
RevEng: Allow user to specify context class name

### LINKED ISSUE #3047: Feature Request: Scaffolding a database should allow to specify the context name
Currently (beta7) when scaffolding a database the derived DbContext class will pick the database name suffixed with the "Context" word.

For example:
```
dnx ef dbcontext scaffold "Server=myserver;Database=Foo;user id=me;password=pwd" EntityFramework.SqlServer
```

Produces a class named `FooContext`. The command should allow passing the context name as a parameter:
```
dnx ef dbcontext scaffold "Server=myserver;Database=Foo;user id=me;password=pwd" EntityFramework.SqlServer --context-name MyDbContext
```

lajones:
Fixes #3047.

### DIFF
--- a/src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringConfiguration.cs
+++ b/src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringConfiguration.cs
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Relational.Design.Utilities;
 
 namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
 {
@@ -27,6 +28,13 @@ namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
             if (string.IsNullOrEmpty(ProjectPath))
             {
                 throw new ArgumentException(Strings.ProjectPathRequired);
+            }
+
+            if (!string.IsNullOrWhiteSpace(ContextClassName)
+                && !CSharpUtilities.Instance.IsValidCSharpIdentifier(ContextClassName))
+            {
+                throw new ArgumentException(
+                    Strings.ContextClassNotValidCSharpIdentifier(ContextClassName));
             }
 
             if (string.IsNullOrEmpty(ProjectRootNamespace))
--- a/src/EntityFramework.Relational.Design/Utilities/CSharpUtilities.cs
+++ b/src/EntityFramework.Relational.Design/Utilities/CSharpUtilities.cs
@@ -207,6 +207,32 @@ namespace Microsoft.Data.Entity.Relational.Design.Utilities
             return string.Format(CultureInfo.InvariantCulture, "{0}", value);
         }
 
+        public virtual bool IsValidCSharpIdentifier([NotNull] string identifier)
+        {
+            Check.NotEmpty(identifier, nameof(identifier));
+
+            var firstChar = identifier[0];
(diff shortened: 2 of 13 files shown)


natemcmaster:
Should we also check if the context name is a valid C# identifier?

lajones:
@natemcmaster I figured the result would be a context class that wouldn't compile and they'd learn that way. But since we've both considered it perhaps I should just fail early instead. I'll update.

lajones:
@natemcmaster all fixed - please review latest commit.

src/EntityFramework.Relational.Design/Utilities/CSharpUtilities.cs
```diff
@@ -207,6 +207,32 @@ public virtual string GenerateLiteral([NotNull] object value)
             return string.Format(CultureInfo.InvariantCulture, "{0}", value);
         }
 
+        public virtual bool IsValidCSharpIdentifier([NotNull] string identifier)
```
natemcmaster: Why not use `Microsoft.CodeAnalysis.CSharp.SyntaxFacts.IsValidIdentifier`?

src/EntityFramework.Relational.Design/Utilities/CSharpUtilities.cs
```diff
@@ -207,6 +207,32 @@ public virtual string GenerateLiteral([NotNull] object value)
...
+            }
+
+            if (_cSharpKeywords.Contains(identifier))
+            {
+                return false;
+            }
+
+            if (_invalidCharsRegex.IsMatch(identifier))
+            {
+                System.IO.File.AppendAllText(@"C:\work\150924_ContextClassName", "Identifier, " + identifier + ", matches _invalidCharsRegex" + Environment.NewLine);
```
natemcmaster: Debug?

natemcmaster:
:shipit: (with the mentioned above)

src/EntityFramework.Relational.Design/Utilities/CSharpUtilities.cs
lajones: Oops. Will fix.

src/EntityFramework.Relational.Design/Utilities/CSharpUtilities.cs
lajones: Good point. Will use that instead.

src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringConfiguration.cs
```diff
@@ -29,6 +30,13 @@ public virtual void CheckValidity()
                 throw new ArgumentException(Strings.ProjectPathRequired);
             }
 
+            if (!string.IsNullOrWhiteSpace(ContextClassName)
+                && !CSharpUtilities.Instance.IsValidCSharpIdentifier(ContextClassName))
```
divega: I am personally not sure we need to do this, but FWIW, the namespace should be a valid identifier as well.

lajones: @divega The namespace is constructed from the project and sub-directory names and already goes through the 'making sure it's a valid identifier' process.

divega: OK, thanks!

lajones:
Checked in with commit 4f133c1.

PR #3191, opened 2015-09-22
Add flag for null properties to state entry

ajcvickers:
This allows a property to marked as conceptually null even when the property type is not nullable and therefore the value can't actually be null. This in turn allows the property to transition through a temporary null state such as is needed for an FK when reparenting.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
@@ -111,7 +111,7 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
 
             if (EntityState == EntityState.Modified)
             {
-                _stateData.FlagAllProperties(EntityType.GetProperties().Count(), isFlagged: false);
+                _stateData.FlagAllProperties(EntityType.GetProperties().Count(), PropertyFlag.TemporaryOrModified, flagged: false);
             }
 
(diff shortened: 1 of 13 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -111,7 +111,7 @@ private bool PrepareForAdd(EntityState newState)
 
             if (EntityState == EntityState.Modified)
             {
-                _stateData.FlagAllProperties(EntityType.GetProperties().Count(), isFlagged: false);
+                _stateData.FlagAllProperties(EntityType.GetProperties().Count(), PropertyFlag.TemporaryOrModified, isFlagged: false);
```
anpete: isFlagged -> flagged

src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -316,11 +327,29 @@ protected virtual void WritePropertyValue([NotNull] IPropertyBase propertyBase,
 
                 if (!Equals(currentValue, value))
                 {
-                    StateManager.Notify.PropertyChanging(this, property);
+                    var writeValue = true;
+                    var asProperty = property as IProperty;
```
anpete: Consider renaming property to propertyBase as above

anpete:
:shipit:

PR #3193, opened 2015-09-22
RevEng: Allow users to specify absolute output paths.

### LINKED ISSUE #3131: RevEng: Allow absolute output paths
This came up during API Review of the Commands assembly. It is a perfectly valid scenario for users to scaffold code to a directory specified by absolute path--both under and outside of the project root.

lajones:
Fixes #3131.


natemcmaster:
:shipit:

lajones:
Checked in with commit a172942.

PR #3196, opened 2015-09-23
Dispose DbContext and service scope

### LINKED ISSUE #1113: Make sure that our use of DI is disposing contexts correctly
Also make sure that other resources owned by the context or its scope are also disposed--e.g connections. Also see #1112.

natemcmaster:
DbContext throws on use after disposal.
Properly disposes service scope.

Fix #1113.

### DIFF
--- a/src/EntityFramework.Core/DbContext.cs
+++ b/src/EntityFramework.Core/DbContext.cs
@@ -50,6 +50,8 @@ namespace Microsoft.Data.Entity
         private ILoggerFactory _loggerFactory;
 
         private bool _initializing;
+        private IServiceScope _serviceScope;
+        private bool _disposed;
 
         /// <summary>
         ///     Initializes a new instance of the <see cref="DbContext" /> class. The
@@ -166,7 +168,17 @@ namespace Microsoft.Data.Entity
 
         private IStateManager GetStateManager() => ServiceProvider.GetRequiredService<IStateManager>();
 
-        private IServiceProvider ServiceProvider => _contextServices.Value.ServiceProvider;
+        private IServiceProvider ServiceProvider
+        {
+            get
+            {
+                if (_disposed)
+                {
+                    throw new ObjectDisposedException(GetType().Name);
+                }
+                return _contextServices.Value.ServiceProvider;
+            }
+        }
 
         private IDbContextServices InitializeServices(IServiceProvider serviceProvider, DbContextOptions options)
         {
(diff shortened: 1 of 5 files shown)


ajcvickers:
:shipit:

src/EntityFramework.Core/DbContext.cs
```diff
@@ -190,10 +202,11 @@ private IDbContextServices InitializeServices(IServiceProvider serviceProvider,
...
                 _logger.LogDebug(Strings.DebugLogWarning(nameof(LogLevel.Debug), nameof(ILoggerFactory) + "." + nameof(ILoggerFactory.MinimumLevel), nameof(LogLevel) + "." + nameof(LogLevel.Verbose)));
 
-                var scopedServiceProvider = serviceProvider
+                _serviceScope = serviceProvider
                     .GetRequiredService<IServiceScopeFactory>()
-                    .CreateScope()
-                    .ServiceProvider;
+                    .CreateScope();
+
+                var scopedServiceProvider = _serviceScope.ServiceProvider;
```
divega: Nit: I would probably just inline this into the return below.

src/EntityFramework.Core/DbContext.cs
```diff
@@ -166,7 +168,17 @@ private DbContextOptions GetOptions(IServiceProvider serviceProvider)
...
         private IStateManager GetStateManager() => ServiceProvider.GetRequiredService<IStateManager>();
 
-        private IServiceProvider ServiceProvider => _contextServices.Value.ServiceProvider;
+        private IServiceProvider ServiceProvider
+        {
+            get
+            {
+                if (_disposed)
+                {
+                    throw new ObjectDisposedException(GetType().Name);
```
divega: So doing anything "meaningful" on an already disposed DbContext will throw. I think this is a neat way of doing that!

src/EntityFramework.Core/DbContext.cs
```diff
@@ -390,10 +403,8 @@ public virtual Task<int> SaveChangesAsync(CancellationToken cancellationToken =
         /// </summary>
         public virtual void Dispose()
         {
-            if (_contextServices.HasValue)
-            {
-                _contextServices.Value.Dispose();
-            }
+            _disposed = true;
+            _serviceScope?.Dispose();
```
divega: Nit: do we really expect this to ever be null?

divega:
:+1:

src/EntityFramework.Core/DbContext.cs
ajcvickers: Yes. It should be null if a context instance has been created but not used for anything.

src/EntityFramework.Core/DbContext.cs
anpete: I would consider doing the regular, boring thing here. Is this trick always guaranteed to work as the code evolves?

src/EntityFramework.Core/DbContext.cs
anpete: Any reason to not check the _disposed flag first?

divega: Got it. I was confusing `Initialize` and `InitializeServices`.

src/EntityFramework.Core/DbContext.cs
natemcmaster: No, if we add new properties or methods that don't access ServiceProvider, then they won't properly throw an ObjectDisposedException. I can add smarter tests that will help us catch these changes.

ajcvickers: I think what we have here is fine. If you want to put the check on the Database and ChangeTracker properties as well I think it would cover everything of substance.

Issue #3169, opened 2015-09-18
Query :: include with inheritance can return fewer results than expected when including optional (null valued) navigations


maumar:
TSQL that we produce for includes with inheritance is incorrect, and can yield fewer results than it should, example of the query we produce:

```
SELECT [t].[Id], [t].[GearNickName], [t].[GearSquadId], [t].[Note], [g].[Nickname], [g].[SquadId], [g].[AssignedCityName], [g].[CityOrBirthName], [g].[Discriminator], [g].[FullName], [g].[LeaderNickname], [g].[LeaderSquadId], [g].[Rank]
FROM [CogTag] AS [t]
LEFT JOIN [Gear] AS [g] ON ([t].[GearNickName] = [g].[Nickname]) AND ([t].[GearSquadId] = [g].[SquadId])
WHERE ([g].[Discriminator] = 'Officer') OR ([g].[Discriminator] = 'Gear')
ORDER BY [g].[FullName]
```

This is wrong when there is a Tag that is not related to a Gear.

Query should look something like:

```
SELECT [t].[Id], [t].[GearNickName], [t].[GearSquadId], [t].[Note], [g].[Nickname], [g].[SquadId], [g].[AssignedCityName], [g].[CityOrBirthName], [g].[Discriminator], [g].[FullName], [g].[LeaderNickname], [g].[LeaderSquadId], [g].[Rank]
FROM [CogTag] AS [t]
LEFT JOIN ( 
	SELECT * FROM [Gear] AS [g] 
	WHERE ([g].[Discriminator] = 'Officer') OR ([g].[Discriminator] = 'Gear')) AS [g] ON ([t].[GearNickName] = [g].[Nickname]) AND ([t].[GearSquadId] = [g].[SquadId])
ORDER BY [g].[FullName]
```

i.e. the Discriminator predicate should be only applied to the Gear table scan, and not the entire JOIN-ed query

maumar:
Fixed in 1a100f95c36697949c5d25791d9b132662b49243

Issue #3163, opened 2015-09-17
Query :: include doesn't work well with inheritance


maumar:
Repro:

```
using System;
using System.Collections.Generic;
using System.Data.SqlClient;
using System.Linq;
using Microsoft.Data.Entity;

namespace Repro3163
{
    public class Customer
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public List<Order> Orders { get; set; }
    }

    public class VipCustomer : Customer
    {
    }

    public class Order
    {
        public int Id { get; set; }
        public int CustomerId { get; set; }
        public string Name { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Customer> Customers { get; set; }
        public DbSet<Order> Orders { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "Repro3163", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Customer>().Collection(e => e.Orders).InverseReference().ForeignKey(e => e.CustomerId);
            modelBuilder.Entity<VipCustomer>().BaseType<Customer>();
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var o11 = new Order { Name = "O11", };
                var o12 = new Order { Name = "O13" };
                var o21 = new Order { Name = "O21" };
                var o22 = new Order { Name = "O22", };
                var o23 = new Order { Name = "O23" };
                var o3 = new Order { Name = "O3" };

                var c1 = new Customer { Name = "C1", Orders = new List<Order> { o11, o12 } };
                var c2 = new Customer { Name = "C2", Orders = new List<Order> { o21, o22, o23 } };

                ctx.Customers.AddRange(c1, c2);
                ctx.Orders.AddRange(o11, o12, o21, o22, o23);
                ctx.SaveChanges();

                var c3 = new VipCustomer { Name = "VC3", Orders = new List<Order> { o3 }, };
                ctx.Customers.Add(c3);
                ctx.SaveChanges();
            }

            using (var ctx = new MyContext())
            {
[...]

PR #3197, opened 2015-09-23
Fix to include #3163 and #3169

maumar:
#3163 - Query :: include doesn't work well with inheritance

Problem was that during binding we were assuming that memberExpression.Expression would always be QuerySourceReferenceExpression. However in some cases (inheritance related) it is wrapped in a convert.
Fix is to strip convert from the expression before checking if it's a QuerySourceReferenceExpression

#3169 - Query :: include with inheritance can return fewer results than expected when including optional (null valued) navigations

Problem was that for include we produce LEFT JOINs, however the discriminator predicate that is applied to entities with inheritance (to restrict them to only sub-types requested) was applied to the entire query.
This could cause some rows being filtered out. Fix is to detect where the discriminator predicate is needed and modify the expression accordingly - we create a subquery and stick discriminator predicate on it,
so it only applies to entities that it should.

Also added lot of tests for the area (include + inheritance) to cover other similar cases.


maumar:
@anpete

src/EntityFramework.Core/Internal/DbContextServices.cs
```diff
@@ -85,6 +85,6 @@ public virtual IDatabaseProviderServices DatabaseProviderServices
 
         public virtual IServiceProvider ServiceProvider => _provider;
 
-        public virtual void Dispose() => (_provider as IDisposable)?.Dispose();
```
maumar: will not check this in

NuGet.config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
maumar: will not check this in

src/EntityFramework.Core/Internal/DbContextServices.cs
```diff
@@ -85,6 +85,6 @@ public virtual IDatabaseProviderServices DatabaseProviderServices
...
+        public virtual void Dispose() { }// => (_provider as IDisposable)?.Dispose();
```
maumar: will not check this in

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
```diff
@@ -159,6 +161,26 @@ var materializer
                                             joinedTableExpression))) - valueBufferOffset,
                                 querySource: null);
 
+                    if (selectExpression.Predicate != oldPredicate)
```
maumar: this is quite hacky - i could make it bit better if the CreateMaterializer returned out parameter with the predicate it created and yield decision to change the predicate on select to a call higher up. Thoughts?

test/EntityFramework.Core.FunctionalTests/InheritanceRelationshipsQueryTestBase.cs
```diff
@@ -0,0 +1,791 @@
...
+            using (var context = CreateContext())
+            {
+                var query = context.BaseEntities.Include(e => e.BaseCollectionOnBase).Where(e => e.Name != "Bar");
+                var result = query.ToList();
+
+                Assert.Equal(BaseEntitiesCount, result.Count);
+            }
+        }
+
+        // doesn't work for now
```
maumar: Haven't filed the bugs yet, will do so once I get the PR in.

test/EntityFramework.SqlServer.FunctionalTests/InheritanceRelationshipsQuerySqlServerTest.cs
```diff
@@ -0,0 +1,939 @@
...
+        {
+            base.Nested_include_with_inheritance_collection_collection_reverse();
+
+            Assert.Equal(
+                @"SELECT [e].[Id], [e].[Discriminator], [e].[Name], [e].[ParentCollectionId], [e].[ParentReferenceId], [b].[Id], [b].[BaseParentId], [b].[Discriminator], [b].[Name], [b0].[Id], [b0].[Discriminator], [b0].[Name], [b0].[DerivedReferenceOnDerivedId]
+FROM [NestedCollectionBase] AS [e]
+LEFT JOIN (
+    SELECT [b].*
+    FROM [BaseCollectionOnBase] AS [b]
+    WHERE ([b].[Discriminator] = 'DerivedCollectionOnBase') OR ([b].[Discriminator] = 'BaseCollectionOnBase')
```
maumar: IN optimization should kick in here, will file a bug on this also

anpete: Yeah, I remember we noticed that before, too.

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
anpete: I though oldPredicate was always null here?

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
```diff
@@ -159,6 +161,26 @@ var materializer
...
+                    {
+                        var subquery = new SelectExpression(_sqlQueryGeneratorFactory, joinExpression.Alias);
```
anpete: Consider Extract Method here

test/EntityFramework.Core.FunctionalTests/EntityFramework.Core.FunctionalTests.csproj
```diff
@@ -78,6 +80,27 @@
     <Compile Include="TestModels\ComplexNavigationsModel\Level3.cs" />
     <Compile Include="TestModels\ComplexNavigationsModel\Level4.cs" />
     <Compile Include="TestModels\GearsOfWarModel\AmmunitionType.cs" />
+    <Compile Include="TestModels\GearsOfWarModel\Officer.cs" />
+    <Compile Include="TestModels\InheritanceRelationships\BaseCollectionOnBase.cs" />
+    <Compile Include="TestModels\InheritanceRelationships\BaseCollectionOnDerived.cs" />
+    <Compile Include="TestModels\InheritanceRelationships\BaseInheritanceRelationshipEntity.cs" />
+    <Compile Include="TestModels\InheritanceRelationships\BaseReferenceOnBase.cs" />
+    <Compile Include="TestModels\InheritanceRelationships\BaseReferenceOnDerived.cs" />
+    <Compile Include="TestModels\InheritanceRelationships\CollecionOnBase.cs" />
```
anpete: Typo

test/EntityFramework.Core.FunctionalTests/InheritanceRelationshipsQueryFixtureBase.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace Microsoft.Data.Entity.FunctionalTests
+{
+    public abstract class InheritanceRelationshipsQueryFixtureBase<TTestStore>
+        where TTestStore : TestStore
+    {
+        public abstract TTestStore CreateTestStore();
+
+        public abstract InheritanceRelationshipsContext CreateContext(TTestStore testStore);
+
+        protected virtual void OnModelCreating(ModelBuilder modelBuilder)
```
anpete: Wow!

test/EntityFramework.Core.FunctionalTests/InheritanceRelationshipsQueryTestBase.cs
```diff
@@ -0,0 +1,791 @@
...
+using Microsoft.Data.Entity.FunctionalTests.TestModels.InheritanceRelationships;
+using Xunit;
+
+namespace Microsoft.Data.Entity.FunctionalTests
+{
+    public abstract class InheritanceRelationshipsQueryTestBase<TTestStore, TFixture> : IClassFixture<TFixture>, IDisposable
+    where TTestStore : TestStore
+    where TFixture : InheritanceRelationshipsQueryFixtureBase<TTestStore>, new()
+    {
+        private const int BaseEntitiesCount = 6;
```
anpete: Inline these..

test/EntityFramework.Core.FunctionalTests/InheritanceRelationshipsQueryTestBase.cs
```diff
@@ -0,0 +1,791 @@
...
+        private const int BaseCollectionsOnDerivedCount = 7;
+        private const int NestedReferencesCount = 9;
+        private const int NestedCollectionsCount = 13;
+
+        [Fact]
+        public virtual void ValidateBaselines()
+        {
+            using (var context = CreateContext())
+            {
+                var baseEntitiesCount = context.BaseEntities.Count();
```
anpete: What is this protecting against?  We don't typically do this..

test/EntityFramework.Core.FunctionalTests/TestModels/InheritanceRelationships/InheritanceRelationshipsModelInitializer.cs
```diff
@@ -0,0 +1,234 @@
...
+                    BaseCollectionOnBase = new List<BaseCollectionOnBase> { bcob21 },
+                    DerivedCollectionOnBase = new List<DerivedCollectionOnBase> { dcob41 },
+
+                    DerivedReferenceOnDerived = drod2,
+                    BaseCollectionOnDerived = new List<BaseCollectionOnDerived> { dcod11, dcod12 },
+                    DerivedCollectionOnDerived = new List<DerivedCollectionOnDerived> { dcod21 },
+                };
+
+                context.BaseEntities.AddRange(derivedEntity1, derivedEntity2, derivedEntity3);
+                context.SaveChanges();
```
anpete: This is a good FT in itself! :smile:

test/EntityFramework.SqlServer.FunctionalTests/InheritanceRelationshipsQuerySqlServerTest.cs
```diff
@@ -0,0 +1,939 @@
...
+LEFT JOIN (
+    SELECT [b].*
+    FROM [BaseReferenceOnBase] AS [b]
+    WHERE ([b].[Discriminator] = 'DerivedReferenceOnBase') OR ([b].[Discriminator] = 'BaseReferenceOnBase')
+) AS [b] ON [b].[BaseParentId] = [e].[Id]
+WHERE [e].[Discriminator] IN ('DerivedInheritanceRelationshipEntity', 'BaseInheritanceRelationshipEntity')",
+                Sql);
+        }
+
+        //public override void Include_reference_with_inheritance2()
```
anpete: Dead?

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
maumar: if the first entity has inheritance, it's discriminator predicate is going to be there.

test/EntityFramework.Core.FunctionalTests/InheritanceRelationshipsQueryTestBase.cs
maumar: initially took the numbers from the test results - wanted to make sure that the queries don't "eat" any results. Will remove since I got my confirmation and the dataset is now stable.

test/EntityFramework.Core.FunctionalTests/TestModels/InheritanceRelationships/InheritanceRelationshipsModelInitializer.cs
maumar: yup, actually found a product bug with this code :)

test/EntityFramework.SqlServer.FunctionalTests/InheritanceRelationshipsQuerySqlServerTest.cs
maumar: base tests don't work due to product bugs, will uncomment however so I won't forget to fix to enable it once the scenario is lit up

test/EntityFramework.Core.FunctionalTests/InheritanceRelationshipsQueryFixtureBase.cs
maumar: some of this (key declarations, specifically) will go away when bugs in mapping are fixed

maumar:
@anpete  new version up

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
anpete: I see. I think this approach is fine vs the out param.

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
```diff
@@ -258,6 +270,23 @@ yield return
             }
         }
 
+        private JoinExpressionBase AdjustJoinExpression(SelectExpression selectExpression, JoinExpressionBase joinExpression)
```
anpete: Can this be a static method?

test/EntityFramework.Core.FunctionalTests/InheritanceRelationshipsQueryFixtureBase.cs
anpete: This is great coverage.

anpete:
:shipit:

src/EntityFramework.Relational/Query/ExpressionVisitors/IncludeExpressionVisitor.cs
maumar: its using non-static _sqlQueryGeneratorFactory to create SelectExpression

PR #3211, opened 2015-09-23
Block shadow keys on properties defined by conventions

### LINKED ISSUE #3156: Re-enable Shadow Keys
These were disabled by #1621. Both @rowanmiller and I have found them very useful.

smitpatel:
Resolves #3156 
As shadow keys were already allowed in model. (only warnings are logged), this change blocks the shadow keys created by convention.

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
@@ -5,17 +5,16 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Metadata.Internal;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
 {
-    public class KeyConvention : IKeyConvention, IForeignKeyRemovedConvention, IPrimaryKeyConvention
+    public class KeyConvention : IKeyConvention, IForeignKeyRemovedConvention, IPrimaryKeyConvention, IModelConvention
     {
         public virtual InternalKeyBuilder Apply(InternalKeyBuilder keyBuilder)
         {
-            Check.NotNull(keyBuilder, nameof(keyBuilder));
-
             var entityTypeBuilder = keyBuilder.ModelBuilder.Entity(
                 keyBuilder.Metadata.DeclaringEntityType.Name, ConfigurationSource.Convention);
 
@@ -26,9 +25,6 @@ namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
 
         public virtual void Apply(InternalEntityTypeBuilder entityTypeBuilder, ForeignKey foreignKey)
         {
-            Check.NotNull(entityTypeBuilder, nameof(entityTypeBuilder));
-            Check.NotNull(foreignKey, nameof(foreignKey));
-
             var properties = foreignKey.Properties;
             SetValueGeneration(entityTypeBuilder, properties);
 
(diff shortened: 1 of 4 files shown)


src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
```diff
@@ -128,5 +129,43 @@ private void SetIdentity(InternalEntityTypeBuilder entityTypeBuilder, Property p
...
+        public virtual InternalModelBuilder Apply(InternalModelBuilder modelBuilder)
+        {
+            Check.NotNull(modelBuilder, nameof(modelBuilder));
+
+            foreach (var entityType in modelBuilder.Metadata.EntityTypes)
+            {
+                var entityTypeBuilder = modelBuilder.Entity(entityType.Name, ConfigurationSource.Convention);
+                foreach (var key in entityType.GetDeclaredKeys())
+                {
+                    if (key.Properties.All(p => ((IProperty)p).IsShadowProperty && entityTypeBuilder.CanRemoveProperty(p, ConfigurationSource.Convention, canOverrideSameSource: true)))
```
ajcvickers: Why is this All and not Any?

src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
```diff
@@ -128,5 +129,43 @@ private void SetIdentity(InternalEntityTypeBuilder entityTypeBuilder, Property p
                 ((IProperty)property).ClrType,
                 ConfigurationSource.Convention)
                 ?.ValueGenerated(ValueGenerated.OnAdd, ConfigurationSource.Convention);
+
+        public virtual InternalModelBuilder Apply(InternalModelBuilder modelBuilder)
+        {
+            Check.NotNull(modelBuilder, nameof(modelBuilder));
```
ajcvickers: Don't put Check calls in internal APIs.

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -233,6 +233,18 @@ private InternalPropertyBuilder ConfigureProperty(InternalPropertyBuilder builde
             return builder;
         }
 
+        public virtual bool CanRemoveProperty([NotNull] Property property, ConfigurationSource configurationSource, bool canOverrideSameSource = true)
```
ajcvickers: Can this be private?

ajcvickers:
:shipit:

src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
smitpatel: That seems to be inconsistent. All apply function in this file have Check calls.

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
smitpatel: It is being called in `KeyConvetion.cs` - can be made internal.

src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
ajcvickers: Then they are all wrong.

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
ajcvickers: Missed that--leave it public. Definitely don't make it internal.

src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
smitpatel: Converted to any. Both will have same result since if any shadow property is created by convention for a key defined by convention then all other properties in the key will be also be shadow and defined by convention.

Issue #2751, opened 2015-07-30
RevEng: allow the user to select which tables from the database to use


lajones:
Reverse-Engineering at the moment assumes that you want to reverse engineer all the tables from the database. Allow the user to specify which tables they want.

divega:
Had an initial discussion with @lajones today, we want to have this fundamental ability exposed to our command line and also APIs so that we or someone can later build UI for it. Some ideas we had on the whiteboard of what we want to support:
- Defining the set of tables by inclusion and exclusion
- Defining what gets excluded or included by exhaustive enumeration (i.e. list of tables) or pattern matching
- For command line we can support different kinds of inputs:
 - Inline
 - Text file
 - Configuration keys in the application's configuration

lajones:
With commits d3c4001 the architecture to allow you to pass in a set of inclusions and exclusions is in. With fcfe8ef I checked in the inline command-line option which will use that architecture. However the form of that is still under discussion. So keeping open for now.

lajones:
Updated the command-line syntax as agreed in the design meeting https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes-(October-1,-2015)#table-selection-in-reverse-engineering with PR #3356.

PR #3214, opened 2015-09-23
Allow user to select tables from database.

lajones:
Partial fix for #2751. Sets up the code to allow users to select tables and tests.

Still to add: what the command line looks like.

### DIFF
--- a/src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
+++ b/src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
@@ -38,6 +38,7 @@ namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
         public virtual ModelUtilities ModelUtilities { get; }
 
         protected abstract IRelationalAnnotationProvider ExtensionsProvider { get; }
+        protected TableSelectionSet _tableSelectionSet = TableSelectionSet.InclusiveAll;
 
         protected RelationalMetadataModelProvider([NotNull] ILoggerFactory loggerFactory,
             [NotNull] ModelUtilities modelUtilities, [NotNull] CSharpUtilities cSharpUtilities)
@@ -50,9 +51,12 @@ namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
             ModelUtilities = modelUtilities;
         }
 
-        public virtual IModel GenerateMetadataModel([NotNull] string connectionString)
+        public virtual IModel GenerateMetadataModel(
+            [NotNull] string connectionString, [CanBeNull] TableSelectionSet tableSelectionSet)
         {
             Check.NotEmpty(connectionString, nameof(connectionString));
+
+            _tableSelectionSet = tableSelectionSet ?? TableSelectionSet.InclusiveAll;
 
             var relationalModel = ConstructRelationalModel(connectionString);
 
--- a/src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
+++ b/src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
+{
+    public class TableSelection
+    {
+        public const string Any = "*";
+        public static readonly TableSelection InclusiveAll = new TableSelection();
+        public static readonly TableSelection ExclusiveAll = new TableSelection() { Exclude = true };
+
+        public virtual string Schema { get;[param: NotNull] set; } = Any;
+        public virtual string Table { get;[param: NotNull] set; } = Any;
+        public virtual bool Exclude { get; set; }
+
+        public virtual bool Matches([NotNull] string schemaName, [NotNull] string tableName)
+        {
+            Check.NotEmpty(schemaName, nameof(schemaName));
+            Check.NotEmpty(tableName, nameof(tableName));
+
+            return (Schema == Any || Schema == schemaName)
+                && (Table == Any || Table == tableName);
+        }
+    }
+}
--- a/src/EntityFramework.Relational.Design/ReverseEngineering/TableSelectionSet.cs
+++ b/src/EntityFramework.Relational.Design/ReverseEngineering/TableSelectionSet.cs
@@ -1,0 +1,106 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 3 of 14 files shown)


test/EntityFramework.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
```diff
@@ -29,8 +47,8 @@ public SqlServerE2ETests(SqlServerE2EFixture fixture, ITestOutputHelper output)
         }
 
         protected override E2ECompiler GetCompiler() => new E2ECompiler
-        {
-            NamedReferences =
+            {
+                NamedReferences =
                     {
```
lajones: I'll fix these whitespace changes. Not sure how they crept in.

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
```diff
@@ -50,10 +50,16 @@ public abstract class RelationalMetadataModelProvider : IDatabaseMetadataModelPr
             ModelUtilities = modelUtilities;
         }
 
-        public virtual IModel GenerateMetadataModel([NotNull] string connectionString)
+        public virtual IModel GenerateMetadataModel(
+            [NotNull] string connectionString, [CanBeNull] TableSelectionSet tableSelectionSet)
         {
             Check.NotEmpty(connectionString, nameof(connectionString));
 
+            if (tableSelectionSet != null)
```
natemcmaster: Nit: `_tableSelectionSet = tableSelectionSet ?? TableSelectionSet.InclusiveAll;` would be clearer.

src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        public const string Any = "*";
+        public static readonly TableSelection InclusiveAll = new TableSelection();
+        public static readonly TableSelection ExclusiveAll = new TableSelection() { Exclude = true };
+
+        public virtual string Schema { get;[param: NotNull] set; } = Any;
+        public virtual string Table { get;[param: NotNull] set; } = Any;
+        public virtual bool Exclude { get; set; }
+
+        public override bool Equals(object o)
```
natemcmaster: Do we need this?

src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
```diff
@@ -0,0 +1,52 @@
...
+            {
+                return false;
+            }
+
+            return Exclude == tableSelection.Exclude
+                && Schema == tableSelection.Schema
+                && Table == tableSelection.Table;
+        }
+
+        public override int GetHashCode()
```
natemcmaster: Same

natemcmaster:
Need more tests. There is some fairly complex logic about how inclusive and exclusive lists are build. Add unit tests to ensure this behaves as expected.

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
lajones: Will do.

src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
lajones: I did have an equals comparison at one stage - but I think it got removed. I'll check.

src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
lajones: Required if Equals is required. I'll check.

lajones:
Agree on the tests - I'll add.

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
lajones: Done.

src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
lajones: No longer needed. Removed.

src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
lajones: No longer needed. Removed.

lajones:
@natemcmaster - new commit - contains fixes for all issues mentioned before.

src/EntityFramework.Relational.Design/ReverseEngineering/TableSelectionSet.cs
```diff
@@ -0,0 +1,106 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
```
natemcmaster: Internal?

src/EntityFramework.Relational.Design/ReverseEngineering/TableSelection.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Relational.Design.ReverseEngineering
```
natemcmaster: Also Internal?

natemcmaster:
The tests look good :+1: In the spirit of #2956, should we be adding more to the public API? Or can we internalize some of this? /cc: @divega ?

lajones:
@natemcmaster @divega The purpose of having them public is to allow a future UI to construct `TableSelection`s and hence a `TableSelectionSet` and pass it in directly.

natemcmaster:
Right. I mean in the .Internal namespace, but still "public class"

lajones:
@divega? Maybe I misunderstand what the Internal namespace is meant for. I thought it was for classes which we depend on internally but which we do _not_ recommend other people use. Here I would be happy for other code to use this. However, it is true that at the moment no other such code exists and usage by a UI is a way off.

divega:
@lajones #2956 says that anything that isn't consumed by provider should be moved to the internal namespace, but I can see why we may not want to follow that in this case. I would like to defer until we do the API review, which should be soon anyway.

lajones:
@natemcmaster @divega OK - so can I go ahead and check in as it stands?

divega:
Actually, let's make these internal for beta8. We can always bring them back in rc1.

lajones:
OK. Will do. Then will check in.

lajones:
Checked in with commit d3c4001.

Issue #2599, opened 2015-07-10
Unary expression visitation throws for certain 'Not (!)' subqueries


EricABC:
In `SqlTranslatingExpressionVisitor`:
https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs#L327

VisitUnary throws for a null operand in the 'ExpressionType.Not' case.  This where clause:
`.Where(t => !t.UpstreamTenantRelationships.Any())` will fail in this case.  However,
`.Where(t => t.UpstreamTenantRelationships.Count() == 0)` will succeed.  The Count version results in VisitBinary returning null to the VisitWhereClause method in `RelationalQueryModelVisitor`.

Changing VisitUnary similar to below resulted in the query executing correctly.

                case ExpressionType.Not:
                {
                    var operand = Visit(expression.Operand);
                    if (operand == null)
                        break;
                    return Expression.Not(operand);
                }

divega:
@maumar this looks like something in your area.

rowanmiller:
Navigation property translation is a work in progress. Leaving open to verify this particular scenario once that is working.

maumar:
Fixed in d2802fdaf35458e35a69f9573e57c592d43c6367

PR #3216, opened 2015-09-24
[TINY] Fix to #2599 - Unary expression visitation throws for certain 'Not (!)' subqueries

maumar:
Fix is to add a null check for for cases where we can't translate the operand


maumar:
@anpete

anpete:
:shipit:

PR #3218, opened 2015-09-24
Fix #2491: Moved Logger out of QueryContext.

### LINKED ISSUE #2491: Move Logger out of QueryContext
Per API review. Probably reference from QueryCompilationContext instead, but also see #2490

anpete:
Also more DRYing up of query components after DI change.


anpete:
@maumar

maumar:
:shipit:

PR #3220, opened 2015-09-24
Remove persist flag from in-memory database

ajcvickers:
In API review we decided to allow named in-memory databases. I will file a bug for that and we can do it if we feel it is high enough priority. The easy change for now is to just remove the persist flag and make it such that in-memory databases always persist per model, per service provider, which is what we were doing in almost all our tests anyway.


anpete:
:shipit:

ajcvickers:
Merged

poke:
After removing `persist: false`, I was just hit by this because my tests expected an empty database. Is there any way to enforce a new in-memory database now? I would actually expect that multiple test cases would all have their own in-memory database but it seems the underlying memory is shared between different facts.

rowanmiller:
@poke you would typically handle this by creating a new DbContextOptions for each test ([example](https://github.com/rowanmiller/UnicornStore/blob/master/UnicornStore/src/UnicornStore.Tests/Controllers/ShippingControllerTests.cs#L15)). We are also going to introduce the idea of named stored later on (but that's not available at the moment).

poke:
@rowanmiller Yeah, that’s what I was expecting too, but it’s not the case. The tests below will succeed when run individually, but when you run them together, it will fail because the data from whatever test was executed first is still in the context:

    public class InMemoryTest
    {
        class ExampleModel
        {
            public int Id { get; set; }
        }

        class ExampleContext : DbContext
        {
            public ExampleContext(DbContextOptions options) : base(options)
            { }

            {
                modelBuilder.Entity<ExampleModel>(e =>
                {
                    e.HasKey(x => x.Id);
                });
            }
        }

        [Fact]
        public void Test1()
        {
            var opts = new DbContextOptionsBuilder();
            opts.UseInMemoryDatabase();
            using (var db = new ExampleContext(opts.Options))
            {
                db.Add(new ExampleModel { Id = 4 });
                db.SaveChanges();

                var result = db.Set<ExampleModel>().ToList();

                Assert.Equal(1, result.Count);
                Assert.Equal(4, result[0].Id);
            }
        }

        [Fact]
        public void Test2()
        {
            var opts = new DbContextOptionsBuilder();
            opts.UseInMemoryDatabase();
            using (var db = new ExampleContext(opts.Options))
            {
                db.Add(new ExampleModel { Id = 7 });
                db.SaveChanges();

                var result = db.Set<ExampleModel>().ToList();

                Assert.Equal(1, result.Count);
                Assert.Equal(7, result[0].Id);
            }
        }
    }

This was already the case with beta7, but there you could say `persist: false` to avoid this behavior explicitly.

ajcvickers:
@poke As the code is right now you would need to create a new IServiceProvider for each test since the InMemoryDatabase is rooted in the service provider. There are various ways of doing this; the most straightforward is this:
```C#

serviceCollection
    .AddEntityFramework()
    .AddInMemoryDatabase();

var serviceProvider = serviceCollection.BuildServiceProvider();

using (var context = new SomeDbContext(serviceProvider))
{
}
```
You can pass in DbContexOptions to the context as well, or use OnConfiguring.

As Rowan said, we have a work item to created named in-memory databases, but it isn't very high priority. We would probably accept a pull-request for this if you want to try contributing it.

divega:
@ajcvickers Would calling EnsureDeleted()/EnsureCreated() also help in a scenario like this? I think it should (as long as the tests don't run in parallel, because they are effectively sharing the same database).

ajcvickers:
@divega @poke Absolutely--I guess I should have mentioned that. :-)

poke:
@divega Yes, that’s what I ended up using. I just added a `db.EnsureDeleted()` at the end of each test that I know that might create duplicate content, and then also put them all into the same xUnit collection.

@ajcvickers Interesting, thank you for that example. Why is the service provider shared between tests though? Shouldn’t tests run completely isolated without “magic” globals?

Do I understand named in-memory databases correctly in that it’s essentially allowing multiple in-memory dbs which would be identified by a name, so you would pass in a name to select a possibly existing database of the current service collection? So instead of storing only one instance, there would be e.g. a dictionary holding multiple instances which you could reference that way.

ajcvickers:
@poke There are competing requirements here. In order for EF to be performant it needs to do some caching which persists beyond the scope of a context instance. Historically this scope would be at the app domain level, but that makes it hard to keep things isolated when they need to be. So in EF7 we use the service provider as the root for all this caching. For this to work the service provider can't be re-created for each context instance because then the performance plummets. So something needs to cache the service provider. Everything is setup so that an app can do this caching itself, but that requires some knowledge of DI and service provider creation. That's a pretty big barrier for a lot of people, so if you don't want to have to care about all that stuff we will create and cache a service provider for you. If you want your tests to be completely isolated, then you should manage the service provider yourself, which is what we do in the vast majority of our own tests.

Your understanding of the named in-memory database seems correct to me.

poke:
@ajcvickers Thank your for the explanations, that helped me understand this a lot better :)

As for the named in-memory databases, I might give it a shot. No promises though, as I’m rather busy at the moment, but I’ll try to squeeze in some time for it.

PR #3223, opened 2015-09-24
Re-enable in-memory graph update test

ajcvickers:
Fix was to re-add the code that was removed and change it to clear data from the in-memory database directly rather than going through EnsureDeleted.


natemcmaster:
:shipit:

ajcvickers:
Merged

PR #3224, opened 2015-09-24
NuGet Commands: Use startup project for app base

### LINKED ISSUE #3135: NuGet Commands: No stack trace is shown
No stacktrace is printed either. The message comes from PowerShell throwing `$null`. For some reason, it looks like an exception is being thrown without a message...

I hit this using the `EntityFramework.Sqlite` provider and not having sqlite3.dll available (i.e. via aspnet/Microsoft.Data.Sqlite#134)

bricelam:
Other changes:
* Delay loading assemblies until needed
* Error better on assembly load failures
* Print stack traces (fixes #3135)

### DIFF
--- a/src/EntityFramework.Commands/Commands/Program.cs
+++ b/src/EntityFramework.Commands/Commands/Program.cs
@@ -6,7 +6,6 @@
 
 using System;
 using System.IO;
-using System.Reflection;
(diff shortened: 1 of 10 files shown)


ajcvickers:
:shipit:

PR #3231, opened 2015-09-25
Fix #1856: Introduces DisableQueryClientEvaluation relational DbConte…

### LINKED ISSUE #1856: Query: Client eval warnings and disablement
We should log warnings when we are doing client eval of filter/orderby etc...

Additionally, we should consider a big switch to disable client eval completely (and throw).

anpete:
…xt option.

- Lifts IDbContextOptions dependency to RelationalQueryModelVisitorFactory
- Client evals generate logging events at Warn level.

### DIFF
--- a/src/EntityFramework.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
+++ b/src/EntityFramework.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
@@ -50,6 +50,9 @@ namespace Microsoft.Data.Entity.Infrastructure
         public virtual TBuilder UseRelationalNulls()
             => SetOption(e => e.UseRelationalNulls = true);
 
+        public virtual TBuilder DisableQueryClientEvaluation()
+            => SetOption(e => e.IsQueryClientEvaluationEnabled = false);
+
         protected virtual TBuilder SetOption([NotNull] Action<TExtension> setAction)
         {
             Check.NotNull(setAction, nameof(setAction));
--- a/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
@@ -8,12 +8,15 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Metadata.Internal;
 using Microsoft.Data.Entity.Query.Expressions;
 using Microsoft.Data.Entity.Query.ExpressionVisitors;
 using Microsoft.Data.Entity.Query.Internal;
+using Microsoft.Data.Entity.Relational.Internal;
 using Microsoft.Data.Entity.Utilities;
+using Microsoft.Framework.Logging;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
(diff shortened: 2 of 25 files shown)


anpete:
@maumar

src/EntityFramework.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
```diff
@@ -50,6 +50,9 @@ public virtual TBuilder SuppressAmbientTransactionWarning()
         public virtual TBuilder UseRelationalNulls()
             => SetOption(e => e.UseRelationalNulls = true);
 
+        public virtual TBuilder DisableQueryClientEvaluation()
+            => SetOption(e => e.IsQueryClientEvaluationEnabled = false);
```
divega: I didn't expect this to be exclusive to relational. Do you think it won't apply to other stores or that we will need something different?

anpete: What would it mean for in-memory? no-op? We could do that, although I'm not sure to what benefit. We don't really know what it means for other stores either. E.g. Redis would likely need client eval very often.

Would it be a big problem to lift it later, if we determine it makes sense?

src/EntityFramework.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -668,8 +688,12 @@ var sqlTranslatingExpressionVisitor
 
             RequiresClientFilter |= requiresClientFilter;
 
+            var predicate = whereClause.Predicate;
+
             if (RequiresClientFilter)
             {
+                CheckClientEval(predicate);
```
maumar: CheckClientEval(whereClause.Predicate)

anpete: Fixed.

test/EntityFramework.Relational.FunctionalTests/QueryNoClientEvalTestBase.cs
```diff
@@ -0,0 +1,236 @@
...
+
+using System;
+using System.Linq;
+using Microsoft.Data.Entity.FunctionalTests.TestModels.Northwind;
+using Microsoft.Data.Entity.Relational.Internal;
+using Xunit;
+
+// ReSharper disable AccessToDisposedClosure
+
+namespace Microsoft.Data.Entity.FunctionalTests
```
maumar: add a case for filter with combination of client and server eval, as there is some relevant logic around that scenario

anpete: I think that will just throw on the same code path, but can add.

maumar:
:shipit:

src/EntityFramework.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
divega: Ok, thinking about it more, I agree it would be ok to lift it later. I wasn't thinking about in-memory or even Redis as the interesting cases for it.

test/EntityFramework.Relational.FunctionalTests/QueryNoClientEvalTestBase.cs
divega: Naïve question, but do we need coverage for queries that will __not__ throw with this flag on? (It may very well be there and I just don't know how to find it).

Issue #3227, opened 2015-09-25
RevEng: Provide default context name when unspecified


bricelam:
The following command...
```ps1
Scaffold-DbContext 'Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=Northwind;Integrated Security=True' EntityFramework.SqlServer
```
...throws the following exception.
```
System.ArgumentException: The string argument 'dbContextClassName' cannot be empty.
   at Microsoft.Data.Entity.Utilities.Check.NotEmpty(String value, String parameterName)
   at Microsoft.Data.Entity.Relational.Design.ReverseEngineering.ReverseEngineeringGenerator.CheckOutputFiles(String projectPath, String relativeOutputPath, String dbContextClassName, IModel metadataModel)
   at Microsoft.Data.Entity.Relational.Design.ReverseEngineering.ReverseEngineeringGenerator.GenerateAsync(ReverseEngineeringConfiguration configuration, CancellationToken cancellationToken)
   at Microsoft.Data.Entity.Design.DatabaseOperations.ReverseEngineerAsync(String provider, String connectionString, String outputDir, String dbContextClassName, Boolean useFluentApiOnly, CancellationToken cancellationToken)
   at Microsoft.Data.Entity.Design.OperationExecutor.<ReverseEngineerImpl>d__19.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.Data.Entity.Design.OperationExecutor.OperationBase.<>c__DisplayClass4_0`1.<Execute>b__0()
   at Microsoft.Data.Entity.Design.OperationExecutor.OperationBase.Execute(Action action)
The string argument 'dbContextClassName' cannot be empty.
```

lajones:
@bricelam I'm not seeing it on my box. And it's weird - there are already a couple of layers of default - see https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational.Design/ReverseEngineering/ReverseEngineeringGenerator.cs#L72 and https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.SqlServer.Design/ReverseEngineering/Configuration/SqlServerModelConfiguration.cs#L30

bricelam:
@lajones Were you using the DNX or NuGet commands? PowerShell might be supplying an empty string instead of null.

lajones:
[...]

PR #3233, opened 2015-09-25
[SMALL] Allow for whitespace from Scaffold-DbContext.

lajones:
@bricelam - this should fix the problem in #3227


bricelam:
:shipit:

lajones:
Checked in with commit 1e6c8c9.

PR #3234, opened 2015-09-25
Allow to reference entity type members from any level of the hierarchy.

### LINKED ISSUE #2987: Allow to override PKs/FKs on relationships that have been already configured
Consider:
```
builder.Entity<Category>().Collection(e => e.Products).InverseReference(e => e.Category)
                    .ForeignKey(e => new [] { e.CategoryId, e.CountryId })
                    .PrincipalKey(e => new [] { e.CategoryId, e.CountryId });

builder.Entity<Category>().Collection(e => e.Products).InverseReference(e => e.Category)
                    .ForeignKey(e => e.CategoryId)
                    .PrincipalKey(e => e.CategoryId);
```

The second configuration will throw currently, because the FK doesn't match the PK specified previously. However doing this might be necessary when configuring a shared model for another provider or another app. This also conflicts with the general guideline of "Last configuration wins".

AndriySvyryd:
Add OnNavigationRemoved to allow RelationshipDiscoveryConvention use the navigations that have been freed
Relationship configuration that conflicts with previous configuration will now override it instead of throwing
Refactor internal builders to remove redundant code.

Part of #1704
Fixes #2987

### DIFF
--- a/src/EntityFramework.Core/EntityFramework.Core.csproj
+++ b/src/EntityFramework.Core/EntityFramework.Core.csproj
@@ -158,6 +158,7 @@
     <Compile Include="DbContextOptionsBuilder`.cs" />
     <Compile Include="Metadata\Builders\ReferenceReferenceBuilder`.cs" />
     <Compile Include="Metadata\Conventions\ConventionSet.cs" />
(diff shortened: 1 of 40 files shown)


lajones:
Looks good to me - but it's very complicated - I would get someone else to have a look as well.

Issue #3091, opened 2015-09-10
Include() Appended to Query is Dependent on Order of Joins


brent-russell:
Issue encountered in Beta7.

When a query contains multiple joins, and an .Include() is appended to the query, the child entities are only included when the DbSet with the Include statement is the initial (first) from clause. In this specific case, the include is also a self referential parent/child relation.

Here's the example, which is probably easier to read through.

```C#
// Topics is first DbSet in from clause - include works
var query = from t in this.DbContext.Topics
			join i in this.DbContext.ItemTopics
				on t.Id equals i.TopicId
			join ci in this.DbContext.CategoryItems
				on i.RootItemId equals ci.RootItemId
			where ci.CategoryId == categoryId
			select t;

// Topics.Children is ICollection<Topic> where Topic.ParentTopicId = Topic.Id
query  = query.Include(t => t.Children);
// children are included
```

```C#
// Topics is NOT first DbSet in from clause - include does not work
var query = from ci in this.DbContext.CategoryItems
			join i in this.DbContext.ItemTopics
				on ci.RootItemId equals i.RootItemId
			join t in this.DbContext.Topics
				on i.TopicId equals t.Id
			where ci.CategoryId == categoryId
			select t;

// Topics.Children is ICollection<Topic> where Topic.ParentTopicId = Topic.Id
query  = query.Include(t => t.Children);
// children are NOT included
```

Edit - Added models:
```C#
public class Item
{
	public int Id { get; set; }
	public int? RootItemId { get; set; }	// self referential to Item.Id
	public Item RootItem { get; set; }
}

public class Topic
{
	public int Id { get; set; }
	public int? ParentTopicId { get; set; }	// self referential to Topic.Id
	public Topic ParentTopic { get; set; }
	public ICollection<Topic> Children { get; set; } = new List<Topic>();
}

public class ItemTopic
{
	public int Id { get; set; }
	public int RootItemId { get; set; }	// reference to Item.Id
	public int TopicId { get; set; }	// reference to Topic.Id
	public Item RootItem { get; set; }
	public Topic Topic { get; set; }
}

public class CategoryItem
{
	public int Id { get; set; }
	public int CategoryId { get; set; }
	public int RootItemId { get; set; }	// reference to Item.Id
	public Category Category { get; set; }
	public Item RootItem { get; set; }
}
```

Edit - OnModelCreating:
```C#
protected override void OnModelCreating(ModelBuilder modelBuilder)
{
[...]

PR #3236, opened 2015-09-25
Fix to #3091 - Include() Appended to Query is Dependent on Order of Joins

maumar:
Problem was that include annotation always assumed that it's query source is the MainFromClause. In the scenario from the bug it is not the case.
Fix is to fix-up the query source of the annotation, by looking at the referenced query source of the NavigationPath,
e.g. in nav path [c].Orders, c is a QuerySourceReference expression to appropriate query source (the one being included, even if it's not the MainFromClause)

### DIFF
--- a/src/EntityFramework.Core/Query/Internal/QueryAnnotationExtractor.cs
+++ b/src/EntityFramework.Core/Query/Internal/QueryAnnotationExtractor.cs
@@ -31,7 +31,12 @@ namespace Microsoft.Data.Entity.Query.Internal
                     .ToList())
             {
                 resultOperator.Annotation.QueryModel = queryModel;
-                resultOperator.Annotation.QuerySource = queryModel.MainFromClause;
+
+                var includeAnnotation = resultOperator.Annotation as IncludeQueryAnnotation;
+                resultOperator.Annotation.QuerySource = includeAnnotation != null
+                    ? ExtractSourceReferenceExpression(includeAnnotation.NavigationPropertyPath).ReferencedQuerySource
+                    : queryModel.MainFromClause;
+
                 queryAnnotations.Add(resultOperator.Annotation);
                 queryModel.ResultOperators.Remove(resultOperator);
             }
@@ -46,6 +51,13 @@ namespace Microsoft.Data.Entity.Query.Internal
                     .TransformExpressions(e =>
                         ExtractQueryAnnotations(e, queryAnnotations));
             }
+        }
+
+        private static QuerySourceReferenceExpression ExtractSourceReferenceExpression(MemberExpression expression)
+        {
+            var expressionWithoutConvert = expression.Expression.RemoveConvert();
+            return expressionWithoutConvert as QuerySourceReferenceExpression
+                ?? ExtractSourceReferenceExpression((MemberExpression)expressionWithoutConvert);
         }
 
         private static Expression ExtractQueryAnnotations(
--- a/test/EntityFramework.Core.FunctionalTests/GearsOfWarQueryTestBase.cs
+++ b/test/EntityFramework.Core.FunctionalTests/GearsOfWarQueryTestBase.cs
@@ -246,6 +246,160 @@ namespace Microsoft.Data.Entity.FunctionalTests
         }
 
         [Fact]
+        public virtual void Select_Where_Navigation_Included()
+        {
+            using (var context = CreateContext())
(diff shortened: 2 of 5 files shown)


maumar:
@anpete

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -207,6 +207,15 @@ protected virtual void ExtractQueryAnnotations([NotNull] QueryModel queryModel)
 
             QueryCompilationContext.QueryAnnotations
                 = _queryAnnotationExtractor.ExtractQueryAnnotations(queryModel);
+
+            foreach (var includeAnnotation in QueryCompilationContext.QueryAnnotations.OfType<IncludeQueryAnnotation>())
```
anpete: Move this into QueryAnnotationExtractor and do it at the time we add the annotation

test/EntityFramework.Core.FunctionalTests/GearsOfWarQueryTestBase.cs
```diff
@@ -246,6 +246,143 @@ public virtual void Include_navigation_on_derived_type()
...
+                Assert.True(result.All(g => g.CityOfBirth != null));
+            }
+        }
+
+        [Fact]
+        public virtual void Include_with_join_reference2()
+        {
+            using (var context = CreateContext())
+            {
+                var query = context.Tags.Join(
```
anpete: I think you may find comprehension syntax easier for joins

test/EntityFramework.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
```diff
@@ -274,6 +274,119 @@ public override void Include_navigation_on_derived_type()
...
+ORDER BY [g].[FullName]",
+                Sql);
+        }
+
+        public override void Include_with_join_and_inheritance1()
+        {
+            base.Include_with_join_and_inheritance1();
+
+            Assert.Equal(
+                @"",
```
anpete: These are disabled because of the other bug?

maumar: yes

test/EntityFramework.Core.FunctionalTests/GearsOfWarQueryTestBase.cs
maumar: not for the composite keys, its terrible

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
maumar: will do

test/EntityFramework.Core.FunctionalTests/GearsOfWarQueryTestBase.cs
anpete: on new {a.Foo, a.Bar} equals new {b.Foo, b.Bar} should work.

maumar: combination of mismatched key names and types makes intellisense act up (when I tried it earlier)

maumar:
@anpete  new version up

src/EntityFramework.Core/Query/Internal/QueryAnnotationExtractor.cs
```diff
@@ -31,7 +31,12 @@ in queryModel.ResultOperators
                     .ToList())
             {
                 resultOperator.Annotation.QueryModel = queryModel;
-                resultOperator.Annotation.QuerySource = queryModel.MainFromClause;
+
+                var includeAnnotation = resultOperator.Annotation as IncludeQueryAnnotation;
+                resultOperator.Annotation.QuerySource = includeAnnotation != null
+                    ? ((QuerySourceReferenceExpression)includeAnnotation.NavigationPropertyPath.Expression).ReferencedQuerySource
+                    : queryModel.MainFromClause;
```
anpete: Does this work for c.Foo.Bar?

maumar:
@anpete  another version up, fixed the bug with the multi-level include and added a test

anpete:
:shipit:

Issue #2916, opened 2015-08-24
Review design of ParameterNameGenerator


rowanmiller:
It currently has two concerns:
* Generating unique names within a scope via `GenerateNext()`
* Adding prefixes etc. for externally provided names via `Generate(string name)`

We should either make it ensure uniqueness always (and possibly move `name` to be an optional `nameHint` parameter) or split the concerns out.

PR #3237, opened 2015-09-25
Move common SQL generation code into unified SQL generator

mikary:
Basic refactoring for #2493


anpete:
:shipit: after interface change

src/EntityFramework.Relational/Migrations/HistoryRepository.cs
```diff
@@ -51,14 +50,14 @@ public abstract class HistoryRepository : IHistoryRepository
...
-            Check.NotNull(sql, nameof(sql));
+            Check.NotNull(sqlGenerator, nameof(sqlGenerator));
 
             _databaseCreator = (IRelationalDatabaseCreator)databaseCreator;
             _executor = executor;
             _connection = connection;
             _modelDiffer = modelDiffer;
             _migrationsSqlGenerator = migrationsSqlGenerator;
-            Sql = sql;
+            SqlGenerator = sqlGenerator;
```
bricelam: :frowning: Didn't we decide just *Sql* was OK in API Review?

src/EntityFramework.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -53,21 +53,26 @@ public class MigrationsSqlGenerator : IMigrationsSqlGenerator
 
         public MigrationsSqlGenerator(
             [NotNull] IRelationalCommandBuilderFactory commandBuilderFactory,
+            [NotNull] ISqlGenerator sqlGenerator,
             [NotNull] IUpdateSqlGenerator sql,
```
bricelam: Does this really need both or can `IUpdateSqlGenerator` be removed?

src/EntityFramework.Relational/Migrations/HistoryRepository.cs
mikary: Oops, didn't mean to change the name, this is just an artifact of some incremental changes where both generators were being used.

src/EntityFramework.Relational/Migrations/MigrationsSqlGenerator.cs
mikary: If the referenced code on `IUpdateSqlGenerator` (or any of the other generators) is general purpose, we could move it to the `ISqlGenerator`. I expect we will want to keep adding to this as we refine SQL generation (i.e. parameter prefixes for #2916, operators, etc.)

PR #3238, opened 2015-09-25
Reveng: alternate keys for SQLite

natemcmaster:
Add support for reverse engineering alternate keys for SQLite. Enabling SQL Server will be more difficult because of its metadata fetching does not account for unique indexes.

Partially completes #2840 

/cc: @lajones


bricelam:
Re: SQL Server: We should get away from querying `INFORMATION_SCHEMA.*` and use `sys.*` instead.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -292,7 +293,7 @@ private void LoadTablesAndColumns(SqliteConnection connection, ModelBuilder mode
 
                         if (keyProps.Count > 0)
                         {
-                            builder.HasKey(keyProps.ToArray());
+                            var key = builder.HasKey(keyProps.ToArray());
```
bricelam: Debugging? Can remove.

bricelam:
LGTM, but I'll let @lajones sign off since I'm a bit out-of-touch on RevEng.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
natemcmaster: Yup. Good catch.

natemcmaster:
@bricelam if I'm not mistaken, sys.* only works for SQL 2008 and up. This is a limitation we are okay with, right?

src/EntityFramework.Relational.Design/ReverseEngineering/Configuration/EntityConfiguration.cs
```diff
@@ -74,11 +74,8 @@ public virtual List<PropertyConfiguration> GetPropertyConfigurations(bool useFlu
         }
 
         public virtual List<RelationshipConfiguration> GetRelationshipConfigurations(bool useFluentApiOnly)
-        {
-            return (useFluentApiOnly
-                ? RelationshipConfigurations
-                : Enumerable.Empty<RelationshipConfiguration>())
+            => RelationshipConfigurations
+                .Where(rc => useFluentApiOnly || !rc.ForeignKey.PrincipalKey.IsPrimaryKey())
                 .ToList();
```
lajones: I'd like to change this to work the way other Get*Configurations() work to be consistent. I.e. update RelationshipConfiguration to have a HasAttributeEquivalent getter which by default is set to true. For the specific situation you outline above, set HasAttributeEquivalent to false when you create the RelationshipConfiguration. Then here if useFluentApiOnly is true return all RelationshipConfigurations, but when useFluentApiOnly is false only return the ones with HasAttributeEquivalent == false.

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
```diff
@@ -160,12 +171,16 @@ public virtual void AddForeignKeysToCodeGenModel([NotNull] IModel relationalMode
...
                         .ToList();
-                    var targetRelationalEntityType = relationalForeignKey.PrincipalEntityType;
-                    var targetCodeGenEntityType = _relationalToCodeGenEntityTypeMap[targetRelationalEntityType];
-                    var targetPrimaryKey = targetCodeGenEntityType.GetPrimaryKey();
+                    
+                    var principalCodeGenEntityType = _relationalToCodeGenEntityTypeMap[relationalForeignKey.PrincipalEntityType];
+                    var principalCodeGenKeyProperties = relationalForeignKey
+                        .PrincipalKey
+                        .Properties
+                        .Select(p => _relationalToCodeGenPropertyMap[p]).ToList().AsReadOnly();
```
lajones: Let's talk about this. I'm concerned that _relationalToCodeGenPropertyMap[p] might not always exist. E.g. if you use the select-table feature I just checked in you could select the table with the FK, but the principal end, though it exists in the database, might not be part of the relational model.

test/EntityFramework.Sqlite.Design.FunctionalTests/ReverseEngineering/SqliteAllFluentApiE2ETest.cs
```diff
@@ -21,5 +21,9 @@ public SqliteAllFluentApiE2ETest(ITestOutputHelper output)
 
         protected override string ProviderName => "EntityFramework.Sqlite.Design";
         protected override IDesignTimeMetadataProviderFactory GetFactory() => new SqliteDesignTimeMetadataProviderFactory();
+        public override void Foreign_key_to_unique_index()
+        {
+            base.Foreign_key_to_unique_index();
+        }
     }
```
lajones: Was this debug?

natemcmaster: Yes, this was debug. Will remove.

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
natemcmaster: Yes, we don't handle that case. Will test it once the table selection changes are added.

divega:
> Re: SQL Server: We should move away from querying  INFORMATION_SCHEMA.*  and just use  sys.*  instead.

Agreed. Do we have a bug on this? (couldn't find it).

natemcmaster:
@divega I'm not sure. But if we do, you can assign it to me.

divega:
> @bricelam if I'm not mistaken, sys.* only works for SQL 2008 and up. This is a limitation we are okay with, right?

I think this is ok. If someone wants to create a provider that supports older than SQL Server 2008 it can be overridden and replaced with the previous form. On the other hand, we could do this only for the information that is not available from `INFORMATION_SCHEMA`... 

> @divega I'm not sure. But if we do, you can assign it to me.

I think it is worth considering. Feel free to file the bug. But wait until we triage it before you start working on it :smile:

natemcmaster:
@divega  https://github.com/aspnet/EntityFramework/issues/3247

src/EntityFramework.Relational.Design/ReverseEngineering/Configuration/EntityConfiguration.cs
natemcmaster: Updated to add HasAttributeEquivalent.

lajones:
:shipit:

natemcmaster:
:watch: Will wait to merge till after beta8 is branched.

divega:
Talked to @natemcmaster and @lajones and we believe this is ok to go in now.

ErikEJ:
@divega  If you start converting to using sys. Views for metadata, please use them throughout to be consistent

PR #3254, opened 2015-09-28
Logging: Prefer use of ILogger<T> in providers.

anpete:


Eilon:
Nice.

bricelam:
I always wondered if this would affect service provider initialization time. I guess we'll find out.

bricelam:
:shipit:

PR #3256, opened 2015-09-28
API review changes to Relational Update classes

mikary:
API review changes from API review 11
cc: @AndriySvyryd


AndriySvyryd:
:shipit:

Issue #2294, opened 2015-05-29
Commands: Design-time scenarios


bricelam:
We should think through all the different scenarios we want to support at design-time. This includes:
* Splitting things (services, config, contexts, models, migrations, etc.) into different projects
* Context constructor and Startup.ConfigureServices patterns (See #639)
* The concept of environments (e.g. Development, Production, etc.)

There are a lot of interacting parts here:
* The project on which the command is invoked
    * Migrations in this assembly
    * Contexts in this assembly
    * Contexts referenced by the migrations
* The startup project
    * Contexts in this assembly
    * Contexts registered in services
* The migrations assembly as configured by the context
    * Migrations in this assembly

A few other related bugs: #732, #2219, #2256, #2293

bricelam:
A few related bugs: #732, #2219, #2256

divega:
I think https://github.com/aspnet/EntityFramework/issues/2357 is also very related.

bricelam:
...and #2553 and #2867... :smile: This is the "fix everything design-time related" bug.

divega:
Good luck! :smile:

bricelam:
See also the [September 17, 2015 Design Meeting Notes](https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes-(September-17,-2015)#design-time-services).

bricelam:
Closing; the remaining work is tracked by #639 & #2679.

PR #3257, opened 2015-09-28
Allow providers and users to override design-time services

### LINKED ISSUE #2871: DNX Commands: Verbose switch
We currently always print verbose output in the DNX commands. We should add a `-v|--verbose` flag to align with other command-line tools.

bricelam:
Part of #2294

To override design-time services, add a method called `ConfigureDesignTimeServices` to your `Startup` class. The startup class is located using the same logic as Hosting. To use an environment other than **Development**, use the `-Environment` (NuGet) or `--environment` (DNX) parameter.

If the provider's design-time assembly is available, their design-time services will be registered during Migrations commands.

This also adds the `--verbose` option to the DNX commands (fixes #2871)


ajcvickers:
:shipit:

PR #3258, opened 2015-09-28
Allow configuring implicit service provider

### LINKED ISSUE #3165: Core: Add DbContext.OnConfiguringServices
When a service provider is created by the framework, we should have a hook users can use to override services. One proposal is...

``` C#
class DbContext
{
    protected virtual void OnConfiguringServices([NotNull] IServiceCollection services)
    {
    }
}
```

bricelam:
To use, override `DbContext.OnConfiguringServices`.

Resolves #3165

cc @rowanmiller @ajcvickers @divega 
I question the value of doing this. Managing it externally doesn't seem like too bad of an alternative. This also doesn't allow you use a different service container (e.g. Ninject).

### DIFF
--- a/src/EntityFramework.Core/DbContext.cs
+++ b/src/EntityFramework.Core/DbContext.cs
@@ -195,7 +195,7 @@ namespace Microsoft.Data.Entity
 
                 var providerSource = serviceProvider != null ? ServiceProviderSource.Explicit : ServiceProviderSource.Implicit;
 
-                serviceProvider = serviceProvider ?? ServiceProviderCache.Instance.GetOrAdd(optionsBuilder.Options);
+                serviceProvider = serviceProvider ?? ServiceProviderCache.Instance.GetOrAdd(optionsBuilder.Options, OnConfiguringServices);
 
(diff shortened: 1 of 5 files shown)


ajcvickers:
I don't think we should do this, at least not now. Is there some specific reason that we need this now?

bricelam:
@ajcvickers No, it just came up in a design meeting. I'll recall the PR and re-submit the issue for triage.

rowanmiller:
The biggest reason was being able to replace a service without having to take over managing the service provider. For example, I want to swap out for a new migration SQL generator, currently I need to start managing my own ServiceProvider which is a pretty big change to the architecture of the app.

ajcvickers:
Replacing services correctly is a non-trivial thing to do. Just figuring out what scoping to use can be hard. Also, I could put code in here which looks correct but will never get called because I should have put it in Startup instead, I think it is going to be hard for people to understand this unless they also understand service providers, in which case managing their own isn't very hard.

If there is extensibility that we think is going to be common and we want to make easy, then I think we should look at doing that for those cases. So, for example, is it common for people to need a way to change or modify the behavior of the Migrations SQL generator? If so, then we should look at doing that in a way that will work regardless of how your DI container is setup.

PR #3259, opened 2015-09-29
UseRowNumberForPaging supports group join and complex projection.

RickyLin:
``` c#
var query = from cs in context.Customers
            join os in context.Orders on cs.CustomerID equals os.CustomerID
            select new
            {
                Contact = cs.ContactName + " " + cs.ContactTitle,
                os.OrderID
            };
var result = query.Skip(10).Take(5).ToArray();
```

The preceding query works in Sql Server 2008 now.


anpete:
Merged, thanks!

PR #3262, opened 2015-09-29
Fix #2916 - Separate unique name generation from adding prefix to parameters

### LINKED ISSUE #2916: Review design of ParameterNameGenerator

mikary:


ajcvickers:
:shipit:

PR #3264, opened 2015-09-29
Reveng: add SQLite metadata reader

natemcmaster:
Refactors SQLite metadata fetching into its own component.

/cc: @lajones

### DIFF
--- a/src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataReader.cs
+++ b/src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataReader.cs
@@ -1,0 +1,209 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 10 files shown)


src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataReader.cs
```diff
@@ -0,0 +1,209 @@
...
+                        && tableSelectionSet.Allows(TableSelection.Any, name))
+                    {
+                        databaseInfo.Tables.Add(new TableInfo
+                        {
+                            Name = name,
+                            CreateStatement = sql
+                        });
+                    }
+                    else if (type == "index")
+                    {
```
lajones: Do you want indexes whose TableName is not allowed by `TableSelectionSet`?

natemcmaster: Yes. Good catch -- will add this in.

lajones:
:shipit:

PR #3266, opened 2015-09-29
Fix conditional tests for TD.Net

### LINKED ISSUE #3192: ConditionalTheory/ConditionalFact tests don't run in TestDriven.NET

natemcmaster:
Fix #3192.

TD.Net does not recognize `IXunitTestCase`. It will only run test cases that are/inherit the type `XunitTestCase`


ajcvickers:
:shipit:

PR #3271, opened 2015-09-30
RevEng: Allow FKs to point to unique keys as well as PKs. 

### LINKED ISSUE #3105: RevEng: Nvarchar foreign key to unique column causing InvalidOperationException
I'm running EF 7 beta8-15676.

I have an existing database I'm trying to reverse engineer using the `scaffold` command:

    dnx ef dbcontext scaffold "{connectionstring}" EntityFramework.SqlServer

Within the database is a table that foreign keys to a unique column of another table (note: Not the PK). They are both `nvarchar(25)`. The simplified schema is:

    TableA (
        ID int not null PK,
        Serial nvarchar(25) not null (unique)
    )

    TableB (
        ID int not null PK,
        Serial nvarchar(25) not null (FK to TableA(Serial))
    )

When running `scaffold`, I receive the following error:

    System.InvalidOperationException: The types of the properties specified for the foreign key {'[dbo][TableB][Serial]'} on entity type '[dbo][TableB]' do not match the types of the properties in the principal key on entity type '[dbo][TableA]'.
       at Microsoft.Data.Entity.Metadata.Property.EnsureCompatible(IReadOnlyList`1 principalProperties, IReadOnlyList`1 dependentProperties, EntityType principalEntityType, EntityType dependentEntityType)
       at Microsoft.Data.Entity.Metadata.ForeignKey..ctor(IReadOnlyList`1 dependentProperties, Key principalKey, EntityType dependentEntityType, EntityType principalEntityType)
       at Microsoft.Data.Entity.Metadata.EntityType.AddForeignKey(IReadOnlyList`1 properties, Key principalKey, EntityType principalEntityType)
       at Microsoft.Data.Entity.Metadata.EntityType.GetOrAddForeignKey(IReadOnlyList`1 properties, Key principalKey, EntityType principalEntityType)
       at Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering.SqlServerMetadataModelProvider.AddForeignKeysToModel(Model relationalModel, Dictionary`2 entityTypeToForeignKeyConstraintsMap)
       at Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering.SqlServerMetadataModelProvider.AddKeysToModel(Model relationalModel)
       at Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering.SqlServerMetadataModelProvider.ConstructRelationalModel(String connectionString)
       at Microsoft.Data.Entity.Relational.Design.ReverseEngineering.ReverseEngineeringMetadataModelProvider.GenerateMetadataModel(String connectionString)
       at Microsoft.Data.Entity.Relational.Design.ReverseEngineering.ReverseEngineeringGenerator.GetMetadataModel(IDatabaseMetadataModelProvider provider, ReverseEngineeringConfiguration configuration)
       at Microsoft.Data.Entity.Relational.Design.ReverseEngineering.ReverseEngineeringGenerator.<GenerateAsync>d__25.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at Microsoft.Data.Entity.Commands.Program.<ReverseEngineerAsync>d__16.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at Microsoft.Data.Entity.Commands.Program.<>c__DisplayClass9_4.<<Main>b__11>d.MoveNext()

lajones:
Also unique keys now generate HasAlternateKey() and PrincipalKey() fluent API for SQL Server.

Fixes #3105. Follow on for PR #3238 which addressed this for SQLite.

### DIFF
--- a/src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
+++ b/src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
@@ -32,10 +32,10 @@ namespace Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering
         private Dictionary<string, ForeignKeyColumnMapping> _foreignKeyColumnMappings;
 
         // primary key, foreign key and unique constraint information constructed from database data
-        private readonly Dictionary<string, Dictionary<string, List<string>>> _constraintToColumnsMap =
-            new Dictionary<string, Dictionary<string, List<string>>>(); // 1st string is TableId, 2nd is ConstraintId, 3rd is ColumnId
-        private readonly Dictionary<string, ConstraintType> _constraintTypes =
-            new Dictionary<string, ConstraintType>();
+        private readonly Dictionary<string,
+            Dictionary<TableConstraint, SortedList<int, string>>> _tableToConstraintToColumnsMap =
+            new Dictionary<string,
+                Dictionary<TableConstraint, SortedList<int, string>>>(); // 1st string is TableId, 2nd is ColumnId
 
         // utility data constructed as we iterate over the data
         private readonly Dictionary<string, EntityType> _tableIdToEntityType = new Dictionary<string, EntityType>();
@@ -89,7 +89,7 @@ namespace Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering
 
                     var tableConstraintColumns = LoadData(
                         conn, TableConstraintColumn.Query, TableConstraintColumn.CreateFromReader, tcc => tcc.Id);
-                    CreatePrimaryForeignKeyAndUniqueMaps(tableConstraintColumns);
+                    CreateConstraintMaps(tableConstraintColumns);
                 }
                 finally
                 {
@@ -136,45 +136,38 @@ namespace Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering
             return data;
         }
 
-        public virtual void CreatePrimaryForeignKeyAndUniqueMaps(
+        public virtual void CreateConstraintMaps(
             [NotNull] Dictionary<string, TableConstraintColumn> tableConstraintColumns)
         {
             Check.NotNull(tableConstraintColumns, nameof(tableConstraintColumns));
 
-            var constraintToColumnsMap = new Dictionary<string, Dictionary<string, int>>();
             foreach (var tableConstraintColumn in tableConstraintColumns.Values)
             {
-                UpdateConstraintTypes(tableConstraintColumn);
-
-                // add this column to the list of columns for this constraint
-                Dictionary<string, int> columnIds;
-                if (!constraintToColumnsMap.TryGetValue(tableConstraintColumn.ConstraintId, out columnIds))
+                var constraintType = GetConstraintType(tableConstraintColumn);
+                if (constraintType == null)
                 {
-                    columnIds = new Dictionary<string, int>();
-                    constraintToColumnsMap[tableConstraintColumn.ConstraintId] = columnIds;
+                    // not a PK, FK or unique constraint
+                    continue;
                 }
 
-                if (!columnIds.ContainsKey(tableConstraintColumn.ColumnId))
+                var tableId = _tableColumns[tableConstraintColumn.ColumnId].TableId;
+                Dictionary<TableConstraint, SortedList<int, string>> constraintToColumnsMap;
+                if (!_tableToConstraintToColumnsMap.TryGetValue(tableId, out constraintToColumnsMap))
                 {
-                    columnIds.Add(tableConstraintColumn.ColumnId, tableConstraintColumn.Ordinal);
-                }
-            }
-
-            foreach (var keyValuePair in constraintToColumnsMap)
-            {
-                var constraintId = keyValuePair.Key;
-                var columnIdToOrdinalMap = keyValuePair.Value;
-                var tableId = _tableColumns[columnIdToOrdinalMap.First().Key].TableId;
-                Dictionary<string, List<string>> constraintIdToColumnsMap;
-                if (!_constraintToColumnsMap.TryGetValue(tableId, out constraintIdToColumnsMap))
-                {
-                    constraintIdToColumnsMap = new Dictionary<string, List<string>>();
-                    _constraintToColumnsMap.Add(tableId, constraintIdToColumnsMap);
+                    constraintToColumnsMap = new Dictionary<TableConstraint, SortedList<int, string>>();
+                    _tableToConstraintToColumnsMap[tableId] = constraintToColumnsMap; 
                 }
 
-                var orderedColumnIds = columnIdToOrdinalMap
-                    .OrderBy(kvp => kvp.Value).Select(kvp => kvp.Key).ToList();
-                constraintIdToColumnsMap.Add(constraintId, orderedColumnIds);
+                var constraint = new TableConstraint(
+                    tableConstraintColumn.ConstraintId, constraintType.Value);
+                SortedList<int, string> sortedColumns;
+                if (!constraintToColumnsMap.TryGetValue(constraint, out sortedColumns))
+                {
+                    sortedColumns = new SortedList<int, string>();
+                    constraintToColumnsMap[constraint] = sortedColumns;
+                }
+
+                sortedColumns.Add(tableConstraintColumn.Ordinal, tableConstraintColumn.ColumnId);
             }
         }
 
@@ -254,7 +247,7 @@ namespace Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering
 
         public virtual void AddPrimaryAndAlternateKeysToModel()
         {
-            foreach (var keyValuePair in _constraintToColumnsMap)
+            foreach (var keyValuePair in _tableToConstraintToColumnsMap)
             {
                 var tableId = keyValuePair.Key;
                 EntityType entityType;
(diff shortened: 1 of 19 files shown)


lajones:
@natemcmaster this also addresses #2840

test/EntityFramework.SqlServer.Design.FunctionalTests/ReverseEngineering/E2E.sql
```diff
@@ -260,7 +266,7 @@ CREATE TABLE "OneToOneSeparateFKDependent" (
 	(
 		"OneToOneSeparateFKDependentFK1", "OneToOneSeparateFKDependentFK2"
 	) REFERENCES "dbo"."OneToOneSeparateFKPrincipal" (
-		"OneToOneSeparateFKPrincipalID1", "OneToOneSeparateFKPrincipalID2"
```
natemcmaster: Rather than changing this existing test case, can we add a second one? We still want to ensure this scenario works.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -30,11 +31,11 @@ public class SqlServerMetadataModelProvider : RelationalMetadataModelProvider
...
         private Dictionary<string, ForeignKeyColumnMapping> _foreignKeyColumnMappings;
 
-        // primary key, foreign key and unique columns information constructed from database data
-        private readonly Dictionary<string, int> _primaryKeyOrdinals = new Dictionary<string, int>();
-        private readonly Dictionary<string, Dictionary<string, int>> _foreignKeyOrdinals =
-            new Dictionary<string, Dictionary<string, int>>(); // 1st string is ColumnId, 2nd is ConstraintId
-        private readonly HashSet<string> _uniqueConstraintColumns = new HashSet<string>();
+        // primary key, foreign key and unique constraint information constructed from database data
+        private readonly Dictionary<string, Dictionary<string, List<string>>> _constraintToColumnsMap =
+            new Dictionary<string, Dictionary<string, List<string>>>(); // 1st string is TableId, 2nd is ConstraintId, 3rd is ColumnId
```
natemcmaster: This would be cleaner if you introduced an internal datatype. It makes the code much more readable too.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -302,54 +248,134 @@ public virtual void AddKeysToModel([NotNull] Entity.Metadata.Model relationalMod
...
+                    var constraintType = _constraintTypes[constraintId];
+                    var columnIds = entry.Value;
+                    var tuple = MatchProperties(columnIds);
+                    var matchingProperties = tuple.Item1;
+                    var unmappedColumnIds = tuple.Item2;
+                    if (ConstraintType.PrimaryKey == constraintType)
+                    {
+                        if (unmappedColumnIds.Count == 0)
+                        {
+                            entityType.SetPrimaryKey(matchingProperties.ToList());
```
natemcmaster: Isn't the `.ToList()` redundant?

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -30,11 +31,11 @@ public class SqlServerMetadataModelProvider : RelationalMetadataModelProvider
...
 
-        // primary key, foreign key and unique columns information constructed from database data
-        private readonly Dictionary<string, int> _primaryKeyOrdinals = new Dictionary<string, int>();
-        private readonly Dictionary<string, Dictionary<string, int>> _foreignKeyOrdinals =
-            new Dictionary<string, Dictionary<string, int>>(); // 1st string is ColumnId, 2nd is ConstraintId
-        private readonly HashSet<string> _uniqueConstraintColumns = new HashSet<string>();
+        // primary key, foreign key and unique constraint information constructed from database data
+        private readonly Dictionary<string, Dictionary<string, List<string>>> _constraintToColumnsMap =
+            new Dictionary<string, Dictionary<string, List<string>>>(); // 1st string is TableId, 2nd is ConstraintId, 3rd is ColumnId
+        private readonly Dictionary<string, ConstraintType> _constraintTypes =
```
natemcmaster: This seems like a lot of overhead just to track what type of constraint we are working on. Can this be added as a property to TableConstraintColumn?

natemcmaster: Nvm. It looks like it already is a property on `TableConstraintColumn`. Why are we duplicating data into another dictionary? I can't see why we need this extra dictionary.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
lajones: Will fix.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -252,47 +234,11 @@ public virtual void AddKeysToModel([NotNull] Entity.Metadata.Model relationalMod
...
-
-                if (primaryKeyProperties.Count() > 0)
-                {
-                    entityType.SetPrimaryKey(
-                        primaryKeyProperties
-                            .OrderBy(p => _primaryKeyOrdinals[p.Name]) // note: for relational property p.Name is its columnId
-                            .ToList());
-                }
-                else
+                if (entityType.FindPrimaryKey() == null)
```
natemcmaster: Nit: post-rc1, we could consider cleaning up the ordering of logic. We have a mix of model building and validation all mixed into one giant `ConstructRelationalModel` step. It could be easier to manage by separating into distinct steps in our reveng process.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
lajones: Hmmm. Will think about it. This is just a map of tables -> their constraints -> the ordered list of columns which make up those constraints.

Having said that I had a think about this last night. I will change the `List<>` to a `SortedList<>`. Then I can inject the data directly and read it out in sort order - which is all I really need.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -302,54 +248,134 @@ public virtual void AddKeysToModel([NotNull] Entity.Metadata.Model relationalMod
...
-                        var foreignKey = fromColumnRelationalEntityType.GetOrAddForeignKey(
-                            foreignKeyCodeGenProperties, targetPrimaryKey, targetRelationalEntityType);
-
-                        if (_uniqueConstraintColumns.Contains(
-                            ConstructIdForCombinationOfColumns(
-                                foreignKeyConstraintRelationalPropertyList
-                                    .Select(p => p.Name)))) // relational property's name is the columnId
+                        if (unmappedColumnIds.Count == 0)
+                        {
+                            entityType.AddKey(matchingProperties.ToList());
```
natemcmaster: Nit: also post-rc1. Let's consider switching to using ModelBuilder where possible. `MatchProperties` duplicates checks we already have in `EntityTypeBuilder.HasKey(string[] propertyNames)`, which will find properties or throw a `ModelItemNotFoundException`

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -302,54 +248,134 @@ public virtual void AddKeysToModel([NotNull] Entity.Metadata.Model relationalMod
...
+                    }
+
+                    var fromProperties = tuple.Item1;
+                    var toColumnIds = FindToColumns(constraintId, fromColumnIds);
+                    if (toColumnIds == null)
+                    {
+                        continue;
+                    }
+
+                    tuple = MatchProperties(toColumnIds);
```
natemcmaster: Similar to the note above. We could simplify here by relying on `ReferenceCollectionBuilder.ForeignKey(string[] foreignKeyPropertyNames)` to do the right thing by finding properties.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
lajones: For updating the model I need to just iterate over the primary and unique keys _before_ I iterate over the foreign keys. Initially I tried storing them in separate trees, but then I found that the code that injected the data in the first place had to be replicated.

Then I tried to store the 2nd string (constraintId) as a tuple {constraintId, constraintType}. But that made the key of that dictionary the tuple, when in fact the constraintId uniquely defines what type of constraint it is. So in the end this made sense.

natemcmaster: It is still pretty messy. When adding primary/alternate keys, could you iterate over all TableConstraintColumn and just filter out foreign keys?

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
lajones: I would have liked to put generation of this message inline with creating the key as I do for other warnings that get issued. But there is a situation which applies to primary keys in particular where no PK exists in the database. In this case no `TableConstraintColumn`s exist for that table and so when I iterate over `_constraintToColumnsMap` we can't spot that it's missing. But I see your point. I can pull this particular validation down into Relational.Design. However I can imagine other situations where only the provider knows that it is a problem and then the provider will need to be responsible for issuing that error message.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
lajones: Let's discuss for post-rc1. `MatchProperties` is matching columnId's from the database to properties that it itself has built up in creating the model. I wouldn't have thought that `ModelBuilder` would have the concept of the columnId as that is specific to SQL Server? Note: column name is something different.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
lajones: See below.

test/EntityFramework.SqlServer.Design.FunctionalTests/ReverseEngineering/E2E.sql
lajones: Yes. Can do.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
natemcmaster: Good points to consider. What is here is fine for now. If this is something you think we should revisit post-rc1, let's create a separate issue to track it.

src/EntityFramework.SqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
natemcmaster: ModelBuilder can use arbitrary strings to name properties. AFAIK `columnId` is the property name, so ModelBuilder should find it just the same.

natemcmaster: But as noted, yes let's move this to a separate issue to be considered after rc1.

lajones:
@natemcmaster Updated the approach with new commit. I now use a `TableConstraint` object instead of separate dictionaries. I also use the `SortedList` approach mentioned. Plus I updated the tests as suggested. Should be good to go. Other issues I'll open for post-rc1.

natemcmaster:
:shipit:

lajones:
Checked in with commit 6d6d1d3.

PR #3276, opened 2015-09-30
Replace cross apply with lateral join

roji:
To have the EF core follow the SQL standard more closely, replaced cross apply with lateral join. SqlServer overrides and uses cross apply.


natemcmaster:
This looks like a good change as `CROSS APPLY` is specific to t-SQL. Only concern here: can we use cross apply as an alias for lateral join? Are they exact aliases? The best docs I found said they are *roughly* equivalent.

/cc: @anpete

roji:
I think that lateral join is a superset of SqlServer's CROSS APPLY. It seems that at least the way CROSS APPLY is *currently* being used, it can be safely replaced with lateral join. If, in the future, lateral joins are generated by the core in contexts where CROSS APPLY does not work, a distinction will have to be made...

rowanmiller:
We discussed in triage and agreed that this would be a good change

natemcmaster:
:shipit:

PR #3278, opened 2015-09-30
Implement cascade delete in the state manager

ajcvickers:
This change ensures that tracked entities are deleted as appropriate during SaveChanges and also that navigations to deleted entities are cleared after SaveChanges happens. (We don't do this immediately as in the old stack and we don't null FKs so as to do as little shredding of the graph as possible.)

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
@@ -398,35 +398,8 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
             }
         }
 
-        public virtual bool PrepareToSave()
+        public virtual InternalEntityEntry PrepareToSave()
         {
-            if (HasConceptualNull)
-            {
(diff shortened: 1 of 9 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -450,8 +423,68 @@ public virtual bool PrepareToSave()
...
+                    EntityType.GetProperties().First(p => _stateData.IsPropertyFlagged(p.Index, PropertyFlag.Null)).Name,
+                    EntityType.DisplayName()));
+            }
+        }
+
+        public virtual void CascadeDelete()
+        {
+            foreach (var fk in EntityType.FindReferencingForeignKeys())
+            {
+                foreach (var dependent in StateManager.GetDependents(this, fk).ToList())
```
anpete: Consider SelectMany

anpete:
:shipit:

ajcvickers:
Merged

PR #3284, opened 2015-09-30
Remove IEntityQueryProvider

ajcvickers:
It adds nothing to the contract of IAsyncQueryProvider, so use this instead.


lajones:
:shipit:

Issue #3110, opened 2015-09-11
Route all statement execution through ISqlStatementExecutor


ajcvickers:
From API review: "Everything should be routed thru here. Move logic from here to RelationalCommand."

mikary:
The `SqlStatementExecutor` disposes `DbCommand`s and closes the connection before the `DbDataReader` can be consumed.
- Update pipeline uses a `Consume(DbDataReader)` pattern that could be provided as an `Action<T>` allowing the executor to control the life-cycle of the ADO objects
- Query
 - Requires a streaming / lazy reader, calling `Dispose` and closing connection outside of the executor
 - Reads need to be performed by an `IValueBufferFactory` to ensure proper typed / untyped reads occour
   - In `FromSql` case, requires access to `DbDataReader` before `IValueBufferFactory` can be constructed

PR #3285, opened 2015-09-30
Add SqlCommandBuilder to unify building RelationalCommands from raw SQL

mikary:
Incremental change for #3110, Start to decouple command building from ISqlStatementExecutor, DRY up user command parameter processing.

### DIFF
--- a/src/EntityFramework.Relational/Storage/Internal/SqlCommandBuilder.cs
+++ b/src/EntityFramework.Relational/Storage/Internal/SqlCommandBuilder.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Storage.Internal
+{
+    public class SqlCommandBuilder : ISqlCommandBuilder
+    {
+        private readonly IRelationalCommandBuilderFactory _commandBuilderFactory;
+        private readonly ISqlGenerator _sqlGenerator;
+        private readonly IParameterNameGeneratorFactory _parameterNameGeneratorFactory;
+
+        public SqlCommandBuilder(
+            [NotNull] IRelationalCommandBuilderFactory commandBuilderFactory,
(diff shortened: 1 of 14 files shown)


test/EntityFramework.Relational.Tests/Storage/SqlCommandBuilderTest.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            var command = builder.Build("SQL COMMAND TEXT");
+
+            Assert.Equal("SQL COMMAND TEXT", command.CommandText);
+            Assert.Equal(0, command.Parameters.Count);
+        }
+
+        [Fact]
+        public virtual void Builds_RelationalCommand_with_emtpy_parameter_list()
+        {
```
lajones: Nit: typo "emtpy"

src/EntityFramework.Relational/Storage/Internal/SqlCommandBuilder.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            _commandBuilderFactory = commandBuilderFactory;
+            _sqlGenerator = sqlGenerator;
+            _parameterNameGeneratorFactory = parameterNameGeneratorFactory;
+        }
+
+        public virtual RelationalCommand Build(
+            [NotNull] string sql,
+            [CanBeNull] IReadOnlyList<object> parameters = null)
+        {
```
lajones: Nit: Doesn't affect anything but here parameters can be null. But it seems that all the points where you call it you check that it's not null?

lajones:
:shipit:

src/EntityFramework.Relational/Storage/Internal/SqlCommandBuilder.cs
mikary: This is actually intentional, the `SqlCommandBuilder` will be replacing part of the functionality of the `SqlStatementExecutor` which manages both command building and execution. Some existing calls in the `DatabaseCreator`, `HistoryRepository` and a few other classes will make use of the null parameter, but it was impractical to include those changes in this PR without expanding the scope of the PR and including changes that would immediately be reverted / reworked in the next PR.

lajones: OK. Sounds good.

PR #3290, opened 2015-10-01
Trigger KeyDiscoveryConventions when new property added

### LINKED ISSUE #2944: Conventions: Key discovery not firing for added shadow properties
Given the following code, a key is not discovered for the `Person` entity type.
```C#
class Person
{
}

class Context : DbContext
{
    {
        modelBuilder.Entity<Person>(
            x =>
            {
                x.Property<int>("Id");

                // UNDONE: Uncomment to workaround
                // x.Key("Id");
            });
    }
}
```

smitpatel:
resolves #2944


ajcvickers:
:shipit:

Issue #2490, opened 2015-06-29
Be consistent about Logger creation and management


ajcvickers:
We need to decide on when to create an ILogger from an ILoggerFactory and be consistent about it. For example, how granular should the ILogger be? What should its lifetime be? How should it be shared across classes, if at all? Exampels--should one be created per query? Per context? Per database?

As part of this we should also review usage of things like RelationalLoggerExtensions

anpete:
Done for runtime.

PR #3300, opened 2015-10-01
Logging: All-up logging improvements:

### LINKED ISSUE #1374: Flag to allow application data in logging

anpete:
- New DbContextOption "LogSqlParameterValues()" replacing use of Debug log level to control sensitive data logging.
- Use a wrapping ILogger<T> in relational to transparently enforce sensitive data logging.
- More consistentizing of logging usage, DI etc.
- Better structured logging of DbCommands.
- Use assembly scoped enums for event ids.
- Use event ids everywhere.

Fix #1374, #2490


src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
```diff
@@ -110,4 +114,54 @@ private static IRelationalDatabaseProviderServices GetProviderServices(IServiceP
             return providerServices;
         }
     }
+
+    public interface ILogSqlParameterValuesAccessor
```
ajcvickers: Move to its own file?

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
```diff
@@ -110,4 +114,54 @@ private static IRelationalDatabaseProviderServices GetProviderServices(IServiceP
...
+    {
+        bool LogSqlParameterValues { get; }
+}
+
+    public class SensitiveDataLogger<T> : ILogger<T>, ILogSqlParameterValuesAccessor
```
ajcvickers: Move to its own file?

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
```diff
@@ -32,7 +34,9 @@ public static EntityFrameworkServicesBuilder AddRelational([NotNull] this Entity
         {
             Check.NotNull(builder, nameof(builder));
 
-            builder.GetService().TryAdd(new ServiceCollection()
+            builder.GetService()
+                .Replace(new ServiceDescriptor(typeof(ILogger<>), typeof(SensitiveDataLogger<>), ServiceLifetime.Scoped))
```
ajcvickers: Does this working depend on when AddEntityFramework is called? In other words, can I accidentally defeat this if I do some logging registration in Startup after AddEntityFramework has already been called?

If so, do we care?

ajcvickers:
:shipit:

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
anpete: Will do.

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
anpete: Will do.

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
anpete: Discussed offline and decided to introduce a new type of ILogger instead.

PR #3304, opened 2015-10-01
:bug: Minor bug fixes for benchmarks infra

rowanmiller:
* Adding a check to only store results if all iterations pass.
* Overriding GetUniqueID to avoid exception that was being swallowed by
xunit but resulted in warnings when diagnostic logging is switched on
(the default implementation serializes method arguments to get the
unique name - but we have the concept of Variation that serves that same
purpose).
* Since the above change introduced a new TestCase type that is used
when not running all iterations, I also took the opportunity to make
this more efficient by using a null collector that avoids garbage
collection etc. This turned out to be the cause of the extreme slow down on DNX... so also removing the logic that skipped tests on DNX.


natemcmaster:
:shipit:

PR #3305, opened 2015-10-02
Adds TelemetrySource event publication for DbCommand execution.

anpete:


Eilon:
Niiiiice! @rynowak @avanderhoorn FYI

src/EntityFramework.Relational/Extensions/TelemetrySourceExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+    internal static class TelemetrySourceExtensions
+    {
+        public static void WriteCommand(
+            [NotNull] this TelemetrySource telemetrySource, 
+            [NotNull] string telemetryName, 
+            [NotNull] DbCommand command)
+        {
+            if (telemetrySource.IsEnabled(telemetryName))
+            {
+                telemetrySource.WriteTelemetry(telemetryName, command);
```
rynowak: This should be an anonymous object like:

```
telemetrySource.WriteTelemetry(telemetryName, new { command = command });
```

The object that you pass in here gets "splatted" into method parameters on the listener adapter.

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
```diff
@@ -32,8 +33,12 @@ public static EntityFrameworkServicesBuilder AddRelational([NotNull] this Entity
         {
             Check.NotNull(builder, nameof(builder));
 
+            var telemetrySource = new TelemetryListener("Microsoft.AspNet"); // copied from Hosting
```
davidfowl: This can't be right

src/EntityFramework.Relational/Query/Internal/AsyncQueryingEnumerable.cs
```diff
@@ -69,6 +73,7 @@ await _queryingEnumerable._relationalQueryContext.Connection
                                     _queryingEnumerable._relationalQueryContext.ParameterValues))
                         {
                             _queryingEnumerable._logger.LogCommand(command);
+                            _queryingEnumerable._telemetrySource.WriteCommand("Microsoft.Data.Entity.BeforeExecuteReader", command);
```
avanderhoorn: Is it possible to get an `Microsoft.Data.Entity.AfterExecuteReader` as well? That way we can run timers between the two?

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
divega: Ok. Can someone explain what it should be? We only add a TelemetrySource if there isn't one already in the container. So if someone wants to consume EF telemetry, they will need to handle both Microsoft.AspNet and whatever we put here.

src/EntityFramework.Relational/Query/Internal/AsyncQueryingEnumerable.cs
avanderhoorn: Also do you only ever execute `ExecuteReaderAsync`? Do you ever run the sync equivalent or `ExecuteScalar`\`ExecuteScalarAsync`?

avanderhoorn: Reason I ask is that it would be good to be able to know when someone does an Async call vs not... we can flag this to the user so they know that they aren't running it as an async call.

avanderhoorn:
Looks great! Made the general comment about having an `After` event as well as a `Before` event. Knowing if the query is `Async` or not, and what type of command is being executed (i.e. `Scalar` vs not). Lastly, it would be good to know if an exception is thrown. We have access to unhanded exceptions, but it would be good to have access lower down just incase anything upstream swallows it, we can still let the user know that the command didn't complete and why.

src/EntityFramework.Relational/Query/Internal/AsyncQueryingEnumerable.cs
anpete: The sync code is in QueryingEnumerable

src/EntityFramework.Relational/Extensions/TelemetrySourceExtensions.cs
anpete: Will fix.

src/EntityFramework.Relational/Query/Internal/AsyncQueryingEnumerable.cs
avanderhoorn: Thanks! And given that we have sync code, could you add a property to that anonymous object to indicate whether the query was sync or not?

anpete:
Pushed a new commit.

src/EntityFramework.Relational/Query/Internal/AsyncQueryingEnumerable.cs
anpete: @avanderhoorn Implemented in latest commit.

src/EntityFramework.Relational/Internal/RelationalTelemetry.cs
```diff
@@ -0,0 +1,47 @@
...
+                    new
+                    {
+                        Command = command,
+                        ExecuteMethod = executeMethod,
+                        IsAsync = async,
+                        Exception = exception
+                    });
+            }
+        }
+    }
```
avanderhoorn: I wondering if its worth creating a separate `WriteCommandError` method? It seems a little weird that the `Before`\`After` events have an `Exception` property that is not relevant do them.

avanderhoorn:
Thanks for the updated properties and added events!

src/EntityFramework.Relational/Internal/RelationalTelemetry.cs
anpete: Do you still want the non-exception part of the payload in the error case?

avanderhoorn: Yes please. Because in an exception case we don't get the `After` message. Keeping that data would be inline with the way other cases like this work.

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
divega: Talked to @rynowak about this. In our usage of `TelemetrySource` we do only expect on in the DI container but at the same time the name passed in the constructor is only informational and we don't expect consumers to have to do any special handling or filtering based on the name (e.g. they will just resolve the service type to get it). So changing this string to be EF specific isn't very important but seems more correct and should be harmless.

anpete: OK, will change it.

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
```diff
@@ -32,8 +33,12 @@ public static EntityFrameworkServicesBuilder AddRelational([NotNull] this Entity
         {
             Check.NotNull(builder, nameof(builder));
 
+            var telemetrySource = new TelemetryListener("Microsoft.Data.Entity");
```
ajcvickers: What happens here when AddRelational is called multiple times?

src/EntityFramework.Relational/Internal/RelationalTelemetry.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Data.Common;
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.Data.Entity.Internal
+{
+    internal static class RelationalTelemetry
```
ajcvickers: Why internal?

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
anpete: We use TryAdd below

src/EntityFramework.Relational/Internal/RelationalTelemetry.cs
anpete: These are internal extension methods.

ajcvickers:
:sheep: :it:

avanderhoorn:
Yay!!! Thanks guys!!!

src/EntityFramework.Relational/Query/Internal/AsyncQueryingEnumerable.cs
```diff
@@ -68,12 +73,30 @@ await _queryingEnumerable._relationalQueryContext.Connection
...
                         {
-                            _queryingEnumerable._logger.LogCommand(command);
-
                             await _queryingEnumerable._relationalQueryContext
                                 .RegisterValueBufferCursorAsync(this, _queryingEnumerable._queryIndex, cancellationToken);
 
-                            _dataReader = await command.ExecuteReaderAsync(cancellationToken);
+                            _queryingEnumerable._logger.LogCommand(command);
+
+                            WriteTelemetry(RelationalTelemetry.BeforeExecuteCommand, command);
```
kceiw: Seems like it's also logging sensitive data. Is it a concern?

avanderhoorn: Guidance on Diagnostic Source is that because all of this is occurring inproc, there aren't privacy concerns from a publishing standpoint. From a subscriber perspective, the owness is on them to insure that they meet privacy guidelines). Additionally, getting access to this data without Diagnostics Source is possible, it just takes a little more effort. Lastly, we are going through the process of getting similar event notifications put directly in ADO (timing is further out at the current time) and the same policy that I've mentioned for this context applies there.

src/EntityFramework.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -147,11 +148,13 @@ public GroupJoinAsyncEnumerator(GroupJoinAsyncEnumerable<TResult> groupJoinAsync
             QueryContext queryContext,
             CommandBuilder commandBuilder,
             ISensitiveDataLogger logger,
+            TelemetrySource telemetrySource,
```
kceiw: In case you're not aware of it, TelemetrySource is refactored as DiagnosticSource.

avanderhoorn: Yep see #3422

Issue #3298, opened 2015-10-01
Reverse Engineering tests fail when run from VS with TestDriven.NET


ajcvickers:
```
------ Test started: Assembly: EntityFramework.SqlServer.Design.FunctionalTests.dll ------

Test 'Microsoft.Data.Entity.SqlServer.Design.FunctionalTests.ReverseEngineering.SqlServerE2ETests.E2ETest_AllFluentApi' failed:
	System.IO.FileNotFoundException : Could not find file 'C:\aspnet\EntityFramework\test\EntityFramework.SqlServer.Design.FunctionalTests\bin\Debug\ReverseEngineering\ExpectedResults\E2E_AllFluentApi\OneToOneFKToUniqueKeyDependent.expected'.
	at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
	at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
	at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
	at System.IO.StreamReader..ctor(String path, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize, Boolean checkHost)
	at System.IO.File.InternalReadAllText(String path, Encoding encoding, Boolean checkHost)
	at System.IO.File.ReadAllText(String path)
	ReverseEngineering\FileSystemFileService.cs(42,0): at Microsoft.Data.Entity.Relational.Design.ReverseEngineering.FileSystemFileService.RetrieveFileContents(String directoryName, String fileName)
	ReverseEngineering\FileSet.cs(35,0): at Microsoft.Data.Entity.Relational.Design.FunctionalTests.ReverseEngineering.FileSet.Contents(String filepath)
	ReverseEngineering\FileSet.cs(33,0): at Microsoft.Data.Entity.Relational.Design.FunctionalTests.ReverseEngineering.FileSet.Contents(Int32 index)
[...]

PR #3307, opened 2015-10-02
Fix problem with not copying expected files to output dir

lajones:
Fix for #3298.


lajones:
@ajcvickers @natemcmaster

natemcmaster:
:shipit:

lajones:
Checked in with commit f1ce78e.

PR #3311, opened 2015-10-02
Allow setting cascade action of SetNull

ajcvickers:
This means that a principal with optional dependents can be deleted and without loading the dependents and the FKs will be set to null automatically by the database.

This is not turned on by default because SQL Server will frequently error out indicating that it generates cycles in the schema.


ilmax:
Wow, I was waiting for this since a while! Nice work EF team!

bricelam:
:shipit:

ajcvickers:
Merged

Ponant:
It might be a good idea to include intellisense describing what the enum ReferentialAction does, too.

PR #3323, opened 2015-10-05
Make TypeMapper protected

roji:
Provider implementations of MigrationSqlGenerator may need to access the type mapper.


natemcmaster:
LGTM.

PR #3326, opened 2015-10-05
Enable cascade delete tests on SQLite

ajcvickers:
The issue was that the test store is creating and opening the connection and then opening a transaction, which meant that our code to enable FKs for the connection had no effect, which is, as I understand it, by design. The fix is to have the test store enable FKs before starting the Transaction.


natemcmaster:
:shipit:

ajcvickers:
Merged

Issue #2679, opened 2015-07-22
DNX Commands: Make ef a global command (enables --starupProject)


davidfowl:
Today users need to run `dnx . ef`, it would be better if they could just run a global command on any project without having to do project modification. To do this properly we'd need to have a way to load a runtime environment to harvest information for loading EF things.

/cc @anurse

bricelam:
Exactly, it executes in the context of your project. I tried going down this path, but things got ugly fast. xUnit.net also tried to do it, but it doesn't work at all since it never loads the "target" project.

anurse:
Yeah, this would probably be best tracked via an issue in DNX to enable global commands to operate in project context.

rowanmiller:
Closing as there is nothing actionable from the EF side at the moment. If the design of global commands changes such that the EF commands could make use of them then we will reconsider.

bricelam:
Re-opening. We need this to properly enable `--startupProject`

bricelam:
Another (easier to implement) option would be to always invoke this on the startup project and have a `--targetProject` option that points to the project with the `DbContext` (actually `Migration`s).

totht91:
I have a workaround for the "migration not working if dbcontext is in another project" 
* Create a Startup.cs class in the class library project's root.
* Add the following content:
```C#
/// <summary>
/// We need a Startup class in this class library,
/// because the ef command line migration tool cannot work,
/// if the DbContext is in a different startup project
/// </summary>
public class Startup
{
    public Startup(IHostingEnvironment env, IApplicationEnvironment appEnv)
    {
        // link the original startup project's root for the config.json
        var builder = new ConfigurationBuilder()
            .SetBasePath(appEnv.ApplicationBasePath + "/../MyApp.Web") 
            .AddJsonFile("config.json")
            .AddJsonFile($"config.{env.EnvironmentName}.json", optional: true);

        builder.AddEnvironmentVariables();
        Configuration = builder.Build();
    }

    public IConfigurationRoot Configuration { get; set; }

    public void ConfigureServices(IServiceCollection services)
    {
        services.AddEntityFramework()
            .AddSqlServer()
            .AddDbContext<ApplicationDbContext>(options => options.UseSqlServer(Configuration["Data:DefaultConnection:ConnectionString"]));
    }

    public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
    {
    }
}
```
* Add ef commands to the class library's project.json
```json
[...]

PR #3327, opened 2015-10-05
Commands: Discover contexts registered in the service provider

### LINKED ISSUE #2293: Commands: Discover contexts registered in the service provider
(Reported by @glen-84 in #2256)

I'm able to add a DbContext from another assembly using `services.AddEntityFramework().AddDbContext<My.ForumDbContext>()`, but the context is not listed when you run `dnx . ef context list` – is it possible to make the migration tools see all DbContexts added in this manner (i.e. via the EntityFrameworkServicesBuilder)?

### LINKED ISSUE #2867: Enable migrations in ASP.NET Class library (package)
Hi,
I'm having DAL (Data Access Layer) class library package in ASP.NET MVC 6 where my entities and DbContext are located and I try to enable migrations  using "dnx . ef migration add MigrationName".
The result it an InvalidOperationException saying: 

System.InvalidOperationException: A type named 'StartupProduction' or 'Startup'
could not be found in assembly 'DAL'.

I wouldn't like to move my DAL structure to the Web application, so is there any workaround about it? Thanks!

bricelam:
Fixes #2293

With this, it's now possible to have your `Startup`, `DbContext`, and `Migration` classes in three separate projects. But not on DNX yet because it's blocked by #2679.

Also stops throwing on DNX if `Startup` isn't found (fixes #2867)

Verification: Manual bashing :hammer:

### DIFF
--- a/src/EntityFramework.Commands/Design/DbContextOperations.cs
+++ b/src/EntityFramework.Commands/Design/DbContextOperations.cs
@@ -7,16 +7,13 @@ using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Design.Internal;
 using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Migrations;
 using Microsoft.Data.Entity.Utilities;
-using Microsoft.Extensions.Logging;
-
-#if DNX451 || DNXCORE50
-using Microsoft.AspNet.Hosting;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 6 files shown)


ajcvickers:
:shipit:

francischeung:
Hi Brice,
What is (or will be) the DNX command to add a migration if Startup and DbContext are in separate projects?
Thanks!

bricelam:
As stated above, we need to do #2679 first, but the command will be:
```cmd
cd src\MyMigrationsProject
dnx ef migrations add MyMigration --startupProject MyStartupProject
```
Note, the `DbContext` could be in a separate, third project so long as it's referenced by **MyStartupProject** and **MyMigrationsProject**.

PR #3331, opened 2015-10-05
Rename EF.SqlServer => EF.MicrosoftSqlServer

### LINKED ISSUE #3270: Consider renaming EntityFramework.SqlServer package and assembly to allow side-by-side loading of EF6 and EF7
This came up in a conversation with @rowanmiller and @bricelam today and we wanted to capture it before we forget: 

We believe that currently the fact that the EntityFramework.SqlServer.dll assembly is named the same between EF6 and EF7 is the only reason both cannot be used side by side. If that proves to be truth, we could consider renaming the assembly (probably in EF7 it would make the most sense, and then the package should probably follow suit). 

Some ideas:

- EntityFramework.MSSqlServer
- EntityFramework.SqlClient

natemcmaster:
Close #3270.

Will add an issue to Announcements once this is merged.

Wish we had an "everyone-beware" tag...


natemcmaster:
cc: @rowanmiller @divega Are we ready to push out this rename?

PR #3334, opened 2015-10-05
DNX Commands: Add --json option

### LINKED ISSUE #2883: DNX Commands: Add --json switch for tool-consumable output
One of the features that the Visual Studio ASP.NET team would like to support in `beta8` is the ability to [publish support for EF migrations](https://github.com/aspnet/EntityFramework/wiki/Visual-Studio-Publish-Integration).

In previous versions we had a difficult time discovering the contexts correctly for the publish dialog. For DNX projects we have `dnx ef context list`. We would like to use this to simplify the discovery process and to ensure we get the list correct. The issue with this is that the output of `dnx ef context list` is not in a format that is readable by VS. I'm proposing that we add a new switch `--json` to indicate that JSON should be returned instead of text.

@rowanmiller @vijayrkn @mayurid

bricelam:
Resolves #2883

### DIFF
--- a/src/EntityFramework.Commands/Commands/Program.cs
+++ b/src/EntityFramework.Commands/Commands/Program.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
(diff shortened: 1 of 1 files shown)


src/EntityFramework.Commands/Commands/Program.cs
```diff
@@ -401,21 +443,64 @@ public virtual int ListContexts()
...
                         {
-                            _logger.Value.LogInformation(migration.Id);
-                            any = true;
-                        }
+                            var builder = new IndentedStringBuilder();
+
+                            builder.AppendLine("[");
 
-                        if (!any)
+                            using (builder.Indent())
```
natemcmaster: Can we leverage existing JSON serializers?

bricelam: Why? :smile: The format is static and the inputs are well-known, so there's little value in adding a new dependency.

natemcmaster: Basically just to make code easier to inspect. I thought DNX already provided a json serializer...but if not, then nvm this as I'd also prefer not to take a new dependencies.

bricelam: It looks like there is an [internal JSON parser](https://github.com/aspnet/dnx/tree/dev/src/Microsoft.Dnx.Runtime.Json.Sources), but no serializer.

natemcmaster:
:shipit:

Issue #3329, opened 2015-10-05
Reverse Engineer: Simplify provider surface to simple database orientated object model


rowanmiller:
We decided to stop using `IModel` to try and represent the database schema and instead use a simple, purpose built object model. We also want to refactor things so that in most cases providers only need to implement the surface area to return this model (plus the code fragment for `OnConfiguring` and the context class name) - the rest of the pipeline should be provider agnostic. 

The rest of the surface area will go in a `.Internal` namespace - folks can still use it if needed, but we may change it in future releases.

rowanmiller:
Assigning to @natemcmaster as he has already prototyped something like this.

cc @roji @ErikEJ as an heads up (this will come for RC)

cc @lajones

natemcmaster:
Closed with https://github.com/aspnet/EntityFramework/pull/3439

PR #3335, opened 2015-10-05
[api-review] Rename M.D.E.Relational.Design namespace to M.D.E.Scaffolding.

### LINKED ISSUE #2956: RevEng: minimize "public" API surface

natemcmaster:
Responding to API review changes.

Moving implementation details to .Internal namespace. (Fix #2956.)

Remaining TODO from API review: unify with command's startup conventions.


natemcmaster:
Let's do #3329 first.

Issue #1667, opened 2015-02-20
RevEng: Use EF to load metadata for SqlServer provider


lajones:
Part of #830. 

As part of this investigate whether it might be better to directly query the sys tables.

ErikEJ:
Highly recommend to use sys views, as can be seen from the improved fk Query, and now is a good time to do this

dazinator:
Would using the 'GetSchema()' method of the 'DbConnection' object (SqlConnection) be a possibility here? This method exposes various schema collections (tables, foreign keys etc) - and would future proof EF should you decide to support wrapped / alternate ADO.net providers that support T-SQL but don't have any concept of sys.views / sys.tables.  https://msdn.microsoft.com/en-us/library/ms254969(v=vs.110).aspx

popcatalin81:
@dazinator sys.views / sys.tables, are regular views in the sys schema (Just like any other view).

dazinator:
@popcatalin81 yup, I realise that haha :)
My point is really that it's already possible to obtain certain schema information via the ADO.NET provider - without hand cranking any t-sql commands at all. This would insulate your code, should the SQL team decide to modify those views, drop them in future, or for example, if those views are different in different SQL versions, you don't have to have different queries for different versions of SQL server (The ADO.NET provider for SQL server will respect the version of SQL server and obtain metadata in the appropriate way for that version when you use the GetSchema method) 
It's just a suggestion, just wondered if you had already looked at that this perhaps and dismissed it for some reason.
For background, should alternate ADO.NET providers #1676 ever be supported by EF in the future, then it also means the ADO.NET provider can implement the GetSchema() method and does not have to support sys.views / sys.tables. It also means that an ADO.NET provider that wrapped the SQL ADO.Net provider could potentially "add in / inject" metadata and expose that to EF - which may or may not have some novel uses!

divega:
FWIW, `GetSchema()` is not a viable "future proof" solution. Besides any gap between the results of `GetSchema()` and the metadata needs of EF7 reverse engineering, there is no `DataTable` in .NET Core.

dazinator:
@divega that's interesting about datatable . I would have thought that system.data.sqlclient would throw runtime errors without it. Rather than pester you I'll give it a go and see what I can learn around that!
Thanks

dazinator:
[...]

PR #3337, opened 2015-10-06
Domain model for relational database

### LINKED ISSUE #3247: Reveng: update metadata fetching for SQL Server
Reverse engineering for SQL Server uses `INFORMATION_SCHEMA.*` but should be updated to use `sys.*` instead.

natemcmaster:
Provides abstraction that provider writers will implement.

(See [./src/EntityFramework.Relational.Design/Model](https://github.com/aspnet/EntityFramework/tree/namc/metadata/src/EntityFramework.Relational.Design/Model)) to view the domain model for this PR.

Also adds an implementation for SQLite and SQL Server.

TODO (will come in separate PRs)
- Modify SQL Server design to use metadata reader
- Move scaffolding to commands project

Fix #3247


src/EntityFramework.Relational.Design/Model/ForeignKey.cs
```diff
@@ -4,20 +4,24 @@
...
 
+        //optional
+        public virtual string Name { get; [param: CanBeNull] set; }
+        public virtual string SchemaName { get; [param: CanBeNull] set; }
+
         // TODO foreign key triggers
-        //public virtual string OnUpdate { get; [param: NotNull] set; }
+        //public virtual string OnDelete { get; [param: NotNull] set; }
 
         // TODO https://github.com/aspnet/EntityFramework/issues/333
```
smitpatel: move TODO above OnDelete

natemcmaster:
cc @ajcvickers @lajones @bricelam This is in response to our API review of Relational.Design. SQLite and SQL Server already use domain models to fetch metadata about schema. This unifies the model for both providers. In second PR, I will update reveng to work off of this domain model.

src/EntityFramework.Sqlite.Design/SqliteMetadataReader.cs
```diff
@@ -5,37 +5,45 @@
...
+        public virtual Database GetDatabaseInfo([NotNull] string connectionString, [NotNull] TableSelectionSet tableSelectionSet)
         {
-            _typeMapper = typeMapper;
-        }
-
-        public virtual DatabaseInfo ReadDatabaseInfo([NotNull] SqliteConnection connection, [CanBeNull] TableSelectionSet tableSelectionSet = null)
-        {
-            Check.NotNull(connection, nameof(connection));
+            Check.NotEmpty(connectionString, nameof(connectionString));
+            Check.NotNull(tableSelectionSet, nameof(tableSelectionSet));
```
lajones: You have [CanBeNull] but then Check.NotNull() on tableSelectionSet?

src/EntityFramework.Relational.Design/Model/Table.cs
```diff
@@ -3,12 +3,14 @@
...
+    public class Table
     {
         public virtual string Name { get; [param: NotNull] set; }
-        public virtual string SchemaName { get; [param: NotNull] set; }
-        public virtual string CreateStatement { get; [param: NotNull] set; }
+
+        // optional
+        public virtual string SchemaName { get; [param: CanBeNull] set; }
+        public virtual string CreateStatement { get; [param: CanBeNull] set; }
     }
```
lajones: CreateStatement feels odd to me. Is there any way you can (in Sqlite) parse the CreateStatement into the info you need and store only that on Table?

natemcmaster: Yes, this is a good point. This is a carry-over from the previous implementation. It can be removed.

src/EntityFramework.Sqlite.Design/SqliteMetadataReader.cs
natemcmaster: Are you looking at an outdated diff? This isn't in the changes anymore.

lajones:
You have a partial rename to the Scaffolding namespace as part of this. Think it would be better to include that in PR #3335.

src/EntityFramework.Relational.Design/Model/Database.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.Data.Entity.Scaffolding.Model
+{
+    public class Database
+    {
+        public virtual string Name { get; [param: CanBeNull] set; }
+        public virtual List<Column> Columns { get; [param: NotNull] set; } = new List<Column>();
+        public virtual List<ForeignKey> ForeignKeys { get; [param: NotNull] set; } = new List<ForeignKey>();
+        public virtual List<Index> Indexes { get; [param: NotNull] set; } = new List<Index>();
+        public virtual List<Table> Tables { get; [param: NotNull] set; } = new List<Table>();
+    }
```
lajones: Are the setters required? Don't we always just Add() to the list?

src/EntityFramework.Relational.Design/Model/Column.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Scaffolding.Model
+{
+    public class Column
+    {
+        public virtual string TableName { get; [param: NotNull] set; }
```
divega: I would like to hear what others think about having more of an object oriented domain model, e.g. Table could have a list of Columns and then Column wouldn't have schema, table and ordinal members. Basically what was implicitly proposed here: http://entityframework.codeplex.com/wikipage?title=Schema%20Information. 

As it stands this seems to be a purely relational domain model of the relational model :smile:, and I don't necessarily think that is bad thing, but I believe we should make an explicit decision.

Also, I wouldn't block this PR on this at all (we can always change it if we decide we need to change it).

ErikEJ: @divega +1 for a object oriented model!

natemcmaster: For the sake of discussion I'll add here that using flat, DTO objects was deliberate. It keeps the code simple and makes it easier for our provider writers to plug-in to reverse engineering. EF provides the code to find the relationships between tables, foreign keys, columns, and indexes.

bricelam: If we go with an OO model, we should make them consistent with the [Create/Add*Operation classes](https://github.com/aspnet/EntityFramework/tree/dev/src/EntityFramework.Relational/Migrations/Operations) in Migrations.

natemcmaster: Does anyone have reasons for preferring an OO model? The more I consiter it, the more I prefer the flat DTO.

IMO in the design-time scenarios, a completely OO model adds unnecessary complexity *for provider writers*. With OO, providers would need to also manage:
 - foreign keys that reference tables and columns not yet fetched (topological sorting)
 - principal tables that are referenced but filtered from selection
 - optimizing perf (implies...)
   - having temporary dictionaries of columns and tables (thus avoiding `O(n)` lookup in a list)
   - ensuring the dictionary keys are unique ids. e.g. schema, name, and table name

These are not impossible, but again, for what benefit do we get from OO? Why require all providers to duplicate effort to handle the scenarios above?

ajcvickers: I don't have a strong preference so long as any OO model has weak binding between the objects such that it is easy to populate from the database schema without substantial sorting and/or multiple passes being required. From the discussions I have been in over past few months it seems like the flat DTO approach is probably the most pragmatic one--so I basically agree with @natemcmaster. However, I don't have a strong preference as long as what we end up with is easy to create and easy to consume.

ErikEJ: @natemcmaster  great advocacy. If dto is the easiest to implement and requires minimal amount of provider code, I am all for it!

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerMetadataReader.cs
```diff
@@ -0,0 +1,226 @@
...
+                    if (tableSelectionSet.Allows(table.SchemaName, table.Name))
+                    {
+                        tables.Add(table);
+                    }
+                }
+            }
+            return tables;
+        }
+
+        private List<Column> GetColumns(SqlConnection connection, TableSelectionSet tableSelectionSet)
```
anpete: It is sad (and ironic) that we aren't using EF to do this.

bricelam: +1, it makes sense on SQLite since they're retrieved via `PRAGMA` calls, but we should try to do it for MicrosoftSqlServer (issue #1667)

bricelam: BTW, I had a prototype for testing [here](https://github.com/bricelam/EntityFramework/blob/migtests/test/EntityFramework.SqlServer.FunctionalTests/TestUtilities/InformationSchema.cs).

anpete: @bricelam Nice! Most other relational providers will work this way, too.

natemcmaster:
Updated with feedback from design meeting today. Please note this is an incremental step. This PR only defines the domain model that will be used in the abstract IModel builder. The next PR will provider-override points as decided earlier today.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataReader.cs
```diff
@@ -0,0 +1,254 @@
...
+                while (reader.Read())
+                {
+                    var table = new Table
+                    {
+                        SchemaName = reader.GetString(0),
+                        Name = reader.GetString(1)
+                    };
+
+                    if (_tableSelectionSet.Allows(table.SchemaName, table.Name))
+                    {
```
lajones: _tableSelectionSet doesn't work this way any more - you'll need to sync up and see `SqlServerMetadataProvider.AllowedBySelections()`

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataReader.cs
```diff
@@ -0,0 +1,254 @@
...
+    schema_name(t.schema_id) as [schema], 
+    t.name as [table], 
+    type_name(c.user_type_id) as [typename],
+    c.name as [column_name], 
+    c.column_id as [ordinal],
+    c.is_nullable as [nullable],
+    i.is_primary_key,
+	object_definition(c.default_object_id) as [default_sql],
+    CAST(c.precision as int) AS [precision],
+    CAST(c.scale as int) AS [scale],
```
lajones: As discussed, when testing this will need to make sure datetime precision and numeric precision are both tested - as they are separate things in `INFORMATION_SCHEMA`.

src/EntityFramework.Relational.Design/EntityFramework.Relational.Design.csproj
```diff
@@ -97,6 +100,9 @@
     <Compile Include="ReverseEngineering\ReverseEngineeringConfiguration.cs" />
     <Compile Include="ReverseEngineering\ReverseEngineeringGenerator.cs" />
     <Compile Include="ReverseEngineering\RelationalMetadataModelProvider.cs" />
+    <Compile Include="ReverseEngineering\Internal\TableSelection.cs" />
+    <Compile Include="ReverseEngineering\Internal\TableSelectionSet.cs" />
+    <Compile Include="ReverseEngineering\Internal\TableSelectionSetBuilder.cs" />
```
lajones: Nit: `TableSelection` and `TableSelectionSetBuilder` no longer exist. Need to sync up. Also not sure why the entries moved - but no big deal.

src/EntityFramework.Relational.Design/Model/ForeignKey.cs
```diff
@@ -0,0 +1,26 @@
...
+        public virtual IList<Column> To { get; [param: NotNull] set; } = new List<Column>();
+
+        //optional
+        public virtual string Name { get; [param: CanBeNull] set; }
+
+        // TODO https://github.com/aspnet/EntityFramework/issues/3330
+        //public virtual ReferentialAction OnDelete { get; [param: NotNull] set; }
+
+        // TODO foreign key triggers
+        //public virtual string OnUpdate { get; [param: NotNull] set; }
```
lajones: Nit: I know they're commented out but shouldn't the return types for `OnDelete` and `OnUpdate` be the same?

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -41,39 +47,116 @@ public override IModel ConstructRelationalModel([NotNull] string connectionStrin
...
-            using (var connection = new SqliteConnection(connectionString))
+        private void LoadEntityTypes(ModelBuilder modelBuilder, SchemaInfo schemaInfo)
+        {
+            foreach (var table in schemaInfo.Tables)
             {
-                connection.Open();
+                modelBuilder.Entity(table.Name, builder =>
+                    {
+                        builder.ToTable(table.Name);
+
```
lajones: Is this missing the table's schema name?

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/SqlServerMetadataReaderTest.cs
```diff
@@ -0,0 +1,202 @@
...
+                    });
+        }
+
+        [Fact]
+        public void It_reads_foreign_keys()
+        {
+            _testStore.ExecuteNonQuery("CREATE SCHEMA db2");
+            var sql = "CREATE TABLE dbo.Ranges ( Id INT IDENTITY (1,1) PRIMARY KEY);" +
+                      "CREATE TABLE db2.Mountains ( RangeId INT NOT NULL, FOREIGN KEY (RangeId) REFERENCES Ranges(Id) )";
+            var dbInfo = GetDatabaseInfo(sql);
```
lajones: Would add a test where the foreign key has 2 or more columns. We've had bugs in that area before.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
natemcmaster: There are no schemas in SQLite.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataReader.cs
natemcmaster: I'll add a TODO. (This will be in the next PR.)

src/EntityFramework.Relational.Design/EntityFramework.Relational.Design.csproj
natemcmaster: Yup, working on a rebase right now.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -41,39 +47,116 @@ public override IModel ConstructRelationalModel([NotNull] string connectionStrin
...
+                    .HasColumnName(column.Name);
 
-            return modelBuilder.Model;
+                if (!string.IsNullOrEmpty(column.DataType))
+                {
+                    property.HasColumnType(column.DataType);
+                }
+
+                if (!string.IsNullOrEmpty(column.DefaultValue))
+                {
```
lajones: Nit: default value of empty string may be allowed and be different from default value of NULL. Suggest changing this to only test for null.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
lajones: Doh. Nvm.

lajones:
:shipit: - with changes suggested.

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/SqlServerMetadataReaderTest.cs
natemcmaster: Added a test :)

natemcmaster:
Rebased and updated this PR. Any final comments?

lajones:
Looks good to me :shipit:

ErikEJ:
@natemcmaster Hi, just to understand this - is the plan to remove the Model folder (and classes) and also remove the XxxMetaDataModelProvider from the provider design components?

natemcmaster:
@ErikEJ not exactly. We made some design decisions late last week. I will write a post  tomorrow in the wiki and post a link.

ErikEJ:
@natemcmaster  Thanks, just trying to keep up!

natemcmaster:
@ErikEJ Here is summary of what we decided. https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes-October-8,-2015 More details to follow (i.e. exact class names)

ErikEJ:
@natemcmaster  could you open a discussion issue for this? One question that springs to mind; when would a provider choose one API over the other?

natemcmaster:
@ErikEJ the discussion has been scattered around different places on our repo. Probably the best place to discuss now is #3423

Issue #3281, opened 2015-09-30
[Test] TestModelSource does not run ModelBuiltConventions


smitpatel:

rowanmiller:
Feel free to add more detail and remove milestone if you don't think this is pri0 for RC

smitpatel:
This bug does not affect the product but due to this #3280 was not caught before. Fix for this should be simple and I will include it with #3280 fix.

PR #3338, opened 2015-10-06
Check for multiple KeyAttribute for base type only, Run ModelBuiltConventions for tests

### LINKED ISSUE #3280: KeyAttributeConvention throws while composite keys used with inheritence
Once the model is built `KeyAttributeConvention` checks if the for composite PK and verifies that it has not been set using data-annotation. Present this does not check for base type and in derived entity tries to set PK which throws `InvalidOperationException`.

Repro:
````
public class Gear
    {
        // composite key
        public string Nickname { get; set; }
        public int SquadId { get; set; }
    }

    public class Officer : Gear
    {
    }

    {
        public DbSet<Gear> Gears { get; set; }

        {
            optionsBuilder.UseSqlServer(
                @"Server=(localdb)\MSSQLLocalDB;Database=_TestApp5;Integrated Security=True;MultipleActiveResultSets=True;Connect Timeout=30");
        }

        {
            modelBuilder.Entity<Gear>().HasKey(g => new { g.Nickname, g.SquadId });
            modelBuilder.Entity<Officer>().BaseType<Gear>();
        }
    }
````

smitpatel:
resolves #3280 & #3281


AndriySvyryd:
:shipit:

PR #3348, opened 2015-10-06
Add interfaces to RelationalCommand and RelationalCommandBuilder

mikary:
Decouples from dependencies for testing and replacing implementation in DI
part of changes for #2498


maumar:
:shipit:

mikary:
@anpete @ajcvickers Could I have one of you two take a look at this PR? I know we've had some discussion in the past about protected parameter-less constructors vs. interfaces and I'd like to be sure this approach is acceptable.

src/EntityFramework.Relational/Storage/IRelationalCommandBuilder.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public interface IRelationalCommandBuilder
```
anpete: Can you make this smaller and use extension methods?

src/EntityFramework.Relational/Storage/IRelationalCommand.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Data.Common;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public interface IRelationalCommand
+    {
+        string CommandText { get; }
+
+        IReadOnlyList<RelationalParameter> Parameters { get; }
```
anpete: By design that parameter type isn't interface?

test/EntityFramework.Relational.Tests/TestUtilities/FakeRelationalCommandBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Storage;
+
+namespace Microsoft.Data.Entity.TestUtilities
+{
+    public class FakeRelationalCommandBuilder : IRelationalCommandBuilder
```
anpete: Can you just use the real one?

src/EntityFramework.Relational/Storage/IRelationalCommand.cs
mikary: The `RelationalParameter` type is only a poco, while the other classes have dependencies for type mapping, logging, ect. I could add an interface to the parameters if that seems like it would make more sense / for consistency.

src/EntityFramework.Relational/Storage/IRelationalCommandBuilder.cs
mikary: The `AddParameter` versions could be replaced with a single version and a few extensions (i.e. `AddParameter(RelationalParameter)`). The other methods are just to expose `IndentedStringBuilder` methods, do you think we should rework them somehow?

test/EntityFramework.Relational.Tests/TestUtilities/FakeRelationalCommandBuilder.cs
mikary: The real one brings along provider-specific dependencies and I was hoping to use this version as a basis for additional unit tests when the `SqlStatementExecutor` is removed. My thought here was to have a trivial implementation to allow fakes like the one in `ReaderModificationCommandBatchTest` without having to fully implement the interface.

src/EntityFramework.Relational/Storage/IRelationalCommandBuilder.cs
mikary: On closer inspection, the `AddParameter` methods depend on the `IRelationalTypeMapper` and couldn't be converted without exposing the service. There really aren't any obvious candidates for  conversion to extension methods.

anpete: Think logging. For methods below that have overloads, create a single version that handles all cases.

anpete: Dependencies should be a detail of the impl.

mikary:
Updated with suggested changes

src/EntityFramework.Relational/Storage/IRelationalCommandBuilder.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        IndentedStringBuilder IndentedStringBuilder { get; }
+
+        IRelationalCommandBuilder AddParameter(
+            [NotNull] string name,
+            [CanBeNull] object value,
+            [NotNull] Func<IRelationalTypeMapper, RelationalTypeMapping> mapType,
+            [CanBeNull] bool? nullable);
```
anpete: Don't need CanBeNull here.

src/EntityFramework.Relational/Storage/RelationalCommandBuilderExtensions.cs
```diff
@@ -0,0 +1,139 @@
...
+    public static class RelationalCommandBuilderExtensions
+    {
+        public static IRelationalCommandBuilder Append(
+            [NotNull] this IRelationalCommandBuilder commandBuilder,
+            [NotNull] object o)
+        {
+            Check.NotNull(commandBuilder, nameof(commandBuilder));
+            Check.NotNull(o, nameof(o));
+
+            commandBuilder.IndentedStringBuilder.Append(o);
```
anpete: Can we change the name of the property to SqlBuilder or CommandTextBuilder?

mikary:
Updated again

anpete:
:shipit:

PR #3349, opened 2015-10-06
:racehorse: Adjust benchmarks that are too quick

rowanmiller:
Some of our benchmarks complete too quickly and this tends to lead to
inconsistent results. In general it seems results are pretty consistent
if each iteration takes at least 50ms to complete and ideally 100ms.
Having at least 100 iterations also helps since we report on the 95th
percentile and this gives a bit of room for a few outliers.

Adjusting our benchmarks to try and fit with these numbers. In some
tests I've just increased the amount of data we deal with and in others
we run the same operation multiple times (depending on what works best
for the operation being tested).

Verified that this does not add any significant overhead for the single
iteration of each test that we run during build.cmd (~40sec for EF7 tests
and ~80sec for EF6 tests on my laptop - which is generally pretty slow).


ajcvickers:
:shipit:

Issue #3249, opened 2015-09-28
RevEng: varchar column with default null constraint generates warning


StevenBonePgh:
The command 

    Scaffold-DbContext -Connection "<connection string>" -Provider EntityFramework.SqlServer

Generates the following warning

    Unable to interpret the string ULL as a SQLServer string literal.
    For column [dbo][MyTable][SomeVarchar] unable to convert default value (NULL) into type System.String. Will not generate code setting a default value for the property [dbo][MyTable][SomeVarchar] on entity type [dbo][MyTable].

If [MyTable][SomeVarchar] column is defined as [NullableInt16] `[SomeVarchar]`(6) NULL,
and the following constraint exists

    ALTER TABLE [dbo].[MyTable] ADD  CONSTRAINT [DF_MyTable_SomeVarchar]  DEFAULT (NULL) FOR [SomeVarchar]

lajones:
Fixed with PR #3350

Issue #3248, opened 2015-09-28
RevEng: smallint column with default null constraint generates warning


StevenBonePgh:
The command 

    Scaffold-DbContext -Connection "<connection string>" -Provider EntityFramework.SqlServer

Generates the following warning

    For column [dbo][MyTable][NullableInt16] unable to convert default value (NULL) into type System.Nullable`1[System.Int16]. Will not generate code setting a default value for the property [dbo][MyTable][NullableInt16] on entity type [dbo][MyTable].

If [MyTable][NullableInt16] column is defined as [NullableInt16] [smallint] NULL,
and the following constraint exists

    ALTER TABLE [dbo].[MyTable] ADD  CONSTRAINT [DF_MyTable_NullableInt16]  DEFAULT (NULL) FOR [NullableInt16]

lajones:
Fixed with PR #3350

PR #3350, opened 2015-10-06
RevEng: Allow null as default value

lajones:
Fix for #3248 and #3249.

### DIFF
--- a/src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
+++ b/src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
@@ -121,6 +121,15 @@ namespace Microsoft.Data.Entity.SqlServer.Design.Utilities
                 };
             }
 
+            if (propertyType.IsNullableType()
+                && "NULL" == sqlServerDefaultValue)
+            {
+                return new DefaultExpressionOrValue()
+                {
+                    DefaultValue = null
+                };
+            }
+
             propertyType = propertyType.UnwrapNullableType();
 
             if (typeof(string) == propertyType)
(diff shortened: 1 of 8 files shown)


ajcvickers:
I tried to review this but I'm struggling to understand the scope. Do we only do the DEFAULT(NULL) if the application explicitly sets "NULL" as the default value SQL? Or do we also do it if CLR default value is set to null? And what f the CLR default is not set, but the type is nullable? And for that matter, can we distinguish between setting null and it just being null because it wasn't set?

It seems to me that the only time we should create DEFAULT(NULL) is when the application has explicitly set the default value SQL to "NULL", and even that case is a little dubious. In all other cases we should just not set the default for nullable columns because it will be null anyway.

lajones:
@ajcvickers I think you have this the wrong way round? This is not a case of the application affecting the metadata of the database but the other way around. Previously, if a user created a column in their database and put `DEFAULT (NULL)` in their column definition, when Reverse Engineering tried to understand that default value in order to generate the appropriate `HasDefaultValue()` calls it would issue a warning that we didn't understand the value `(NULL)`. This fix updates the code so we no longer output that warning.

Following our last conversation my understanding was that there was never a need to output a `HasDefaultValue(null)` in the generated code because in order for the database to allow that situation the column must be nullable and the default value for any nullable type that we generate is null anyway. So in the expected results for the added test columns note that they do _not_ have that API call. But please let me know if I misunderstood.

ajcvickers:
@lajones Okay, that makes sense. Let me look again.

src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
```diff
@@ -121,6 +121,15 @@ public virtual string ConvertSqlServerStringLiteral([NotNull] string sqlServerSt
                 };
             }
 
+            if (propertyType.IsNullableType()
+                && "NULL" == sqlServerDefaultValue)
```
ajcvickers: Is it always uppercase NULL? Also, the style we use everywhere else is to put the string literal on the right of the operator.

ajcvickers:
:shipit:

src/EntityFramework.SqlServer.Design/Utilities/SqlServerLiteralUtilities.cs
lajones: Thanks. Yes, it is capitalized even if you put some different capitalization in the column definition. I'll change the order of the comparison.

lajones:
Checked in with commit 8600030.

Issue #3072, opened 2015-09-09
Investigate disabling sentinel behavior for non-generated keys


rowanmiller:
When a key property is configured to be non-generated (I.e. `modelBuilder.Entity<Blog>().Property(b => b.BlogId).ValueGeneratedNever();`) we should see if we can remove the idea of a reserved sentinel value since we have seen this cause confusion/frustration for some folks who want to be able to insert this value.

The biggest issue with doing this is not with PK values but with FKs. Currently we use the sentinel value to represent "conceptual null" when you remove a relationship (via nav props) but the FK property can not be set to null. We're hitting some cases where this doesn't work though, so we are probably going to introduce a secondary "conceptual null" flag like we had in EF6.

If all these changes go as expected, the only place we would use sentinels is to work out whether to generate a key value or not when you add an entity (i.e. you have database generated values but you want to insert explicit values for some entities). We may just want to remove the global concept of sentinel value all together and just have this logic baked into the Add pipeline.

tstar:
I vote with both hands for disabling this feature by default. I have a database first model with lots of of tables with tinynt primary key. This is a big problem, even for reading data from these tables. Every day I struggle with new and new exceptions when trying to read data: "Unable to materialize an entity of type 'Test' because it has an invalid key value. Ensure that the key properties of the entity type have the correct sentinel value configuration.". It is very difficult to analyze each table in an attempt to allocate a value to the sentinel value and hope that it will not be added to the table in the future. Also, a big problem is that this process can not be automated. The process of migration to EF7 should be as simple as possible. Any example from documentation should work on any data. Table with boolean primary key shows what incredible problems adds sentinel value. I am sure that the sentinel value must be an optional feature that can be enabled if desired.

tstar:
It is available in version 7.0.0-rc1-15945?

Issue #1581, opened 2015-02-10
Create convention to set non-CLR default for enum key sentinel value


ajcvickers:
When enums are used as keys the default(T) sentinel value will often not be appropriate because default(T) is usually a valid value in the enum. There should be a convention that sets, for example, -1, as the sentinel for enum keys.

ajcvickers:
Sentinels going away, so this is not needed.

PR #3354, opened 2015-10-06
Remove sentinel values

ajcvickers:
Note that currently the default value is now always used to try to do key generation and hence is treated differently from other values, but implementation of #3072 will allow those cases to work if key generation is explicitly turned off and all key values and foreign key values are always specified explicitly.


bricelam:
:sheep: :it:

smitpatel:
if sentinel values are being removed then #1581 still required?

ajcvickers:
@smitpatel Nope--I'll close it.

PR #3355, opened 2015-10-06
:bug: Fix ServiceProviderCache Hash

rowanmiller:
The hashing algorithm for `ServiceProvider` was using the hash code on
the instance for anything that was added as an instance. This was broken
when we started adding `TelemetryListener` because we were adding a
different instance each time `AddRelational()` was called. 

This showed up as a massive slow down in our perf tests because we effectively
stopped caching anything when the context is responsible for managing
the `ServiceProvider`.

Two changes as part of this:
* Stop using the actual instance when calculating the hash and just use
the type.
* Swap to using `AddSingleton` instead of `AddInstance` for
`TelemetryListener`. Not strictly needed due to the previous change, but
it seems cleaner.

### DIFF
--- a/src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
+++ b/src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
@@ -33,12 +33,10 @@ namespace Microsoft.Data.Entity.Infrastructure
         {
             Check.NotNull(builder, nameof(builder));
 
-            var telemetrySource = new TelemetryListener("Microsoft.Data.Entity");
-            
             builder.GetService()
                 .TryAdd(new ServiceCollection()
-                .AddInstance<TelemetrySource>(telemetrySource)
-                .AddInstance(telemetrySource)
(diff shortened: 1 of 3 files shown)


ajcvickers:
:shipit:

src/EntityFramework.Relational/Infrastructure/RelationalEntityFrameworkServicesBuilderExtensions.cs
```diff
@@ -33,12 +33,10 @@ public static EntityFrameworkServicesBuilder AddRelational([NotNull] this Entity
         {
             Check.NotNull(builder, nameof(builder));
 
-            var telemetrySource = new TelemetryListener("Microsoft.Data.Entity");
-            
             builder.GetService()
                 .TryAdd(new ServiceCollection()
-                .AddInstance<TelemetrySource>(telemetrySource)
-                .AddInstance(telemetrySource)
+                .AddSingleton(s => new TelemetryListener("Microsoft.Data.Entity"))
```
divega: Interesting. Should Hosting also add this as a singleton?

divega: It actually shouldn't make a difference.

PR #3357, opened 2015-10-06
:fire: Resolve build warning for unused variable

rowanmiller:
InMemoryCommandLogger._logger was producing a build warning. Removing it
because it is never used anywhere.


natemcmaster:
Can we leave this in? This is present for easy debugging similar to https://github.com/aspnet/EntityFramework/blob/2d986407aef6451148a2b2cc613fafa3e3e33647/test/EntityFramework.MicrosoftSqlServer.FunctionalTests/QuerySqlServerTest.cs#L23

rowanmiller:
Should it be uncommented then? At the moment it is just dead code (hence the build warning)

natemcmaster:
Yes, it's intentional dead code. It just makes it easier to debug reveng. If uncommented, our test output on CI gets littered with the logging.

rowanmiller:
Updated to introduce a global flag, that way there is no build warning and you can just set the flag to true to debug.

natemcmaster:
:shipit:

PR #3358, opened 2015-10-06
Commands: Start new namespace if default contains migrations for another DbContext

### LINKED ISSUE #3124: Commands: Start a new namespace when generating migrations for a different context
If you have multiple contexts in your project, you can create multiple migrations with the same name. For example:

    dnx ef migrations add Example -c ContextA
    dnx ef migrations add Example -c ContextB

This will result in two `<timestamp>_Example.cs` files in the `Migrations` folder. Since the generated classes are partial, this only throws an error because the class then receives two `DbContextAttributes`.

In comparison, when trying to create two migrations with the same name for a single context, the command correctly fails with an error that the name is already used by an existing migration.

Because of how the migrations are created, adding a migration should always fail if one with the same name already exists, even if it is for another context.

In the long run, it might be more useful to either prefix the migrations by the context name automatically or even move them into a subfolder/subnamespace.

### LINKED ISSUE #3243: A way to use custom migrations directory?
Seems command `dnx ef migrations add` doesn't support custom directories to save migrations to.
So, having multiple database contexts will produce difficulties in generating migrations.
Or am I wrong and it can be done some way?

bricelam:
Resolves #3124

Also adds an `-OutputDir` parameter (resolves #3243)

### DIFF
--- a/src/EntityFramework.Commands/Commands/Program.cs
+++ b/src/EntityFramework.Commands/Commands/Program.cs
@@ -198,6 +198,9 @@ namespace Microsoft.Data.Entity.Commands
                             var name = add.Argument(
                                 "[name]",
                                 "The name of the migration");
+                            var outputDir = add.Option(
+                                "-o|--outputDir",
+                                "The directory (and sub-namespace) to use. If omitted, \"Migrations\" is used.");
                             var context = add.Option(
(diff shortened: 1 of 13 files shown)


ajcvickers:
:shipit:

PR #3367, opened 2015-10-07
Add typemapping from string to CLR types

natemcmaster:
One more step towards #3329 

This just moves existing reverse- and forward-type mapping under one roof.


natemcmaster:
cc @ajcvickers (I believe you did most of the typemapping work, right?)

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
```diff
@@ -92,6 +92,61 @@ public SqlServerTypeMapper()
                 };
         }
 
+        private readonly Dictionary<string, Type> _clrMappings
```
ajcvickers: Can this be combined with _simpleNameMappings, possibly with the introduction of a CLR type pn the RelationalTypeMapping? Otherwise the provider is still providing the mappings twice.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteDesignTimeMetadataProviderFactory.cs
```diff
@@ -22,7 +23,7 @@ public override void AddMetadataProviderServices([NotNull] IServiceCollection se
                 .AddSingleton<ICompilationService, RoslynCompilationService>()
                 .AddSingleton<RazorTemplating>()
                 .AddSingleton<IDatabaseMetadataModelProvider, SqliteMetadataModelProvider>()
-                .AddSingleton<SqliteReverseTypeMapper>()
+                .AddSingleton <SqliteTypeMapper>()
                 .AddSingleton<IRelationalAnnotationProvider, SqliteAnnotationProvider>()
```
smitpatel: <space>

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
natemcmaster: Actually, that was my first approach, but reverted to this when I realized the list of datatypes we support forwards and backwards are not the same.

ajcvickers: Doesn't that mean we are either missing some mappings or one side or the other will have redundancy, which doesn't matter? For example, having a mapping for "rowversion" even though reverse engineering will never use it doesn't seem to be a problem.

natemcmaster: Yes, I suppose it means we have some redundancy. On second thought and further analysis, we can reduce this redundancy.

Question then. If we add a property `RelationalTypeMapping.ClrType`, should this always be true : `mapper.GetMapping(Type input).ClrType == input`? This will add some bloat to our current code.

ajcvickers: The RelationalTypeMapping can be obtained from a number of inputs. If you ask for it from a CLR type and then look at the mapping, then you will see that it is for the CLR type. I think this is fine and don't really see it as bloat. To have an entirely separate dictionary of mapping information seems more like bloat to me. But maybe I am not understanding.

natemcmaster: No, it's minor, bloat is the wrong word. I'll update the PR with changes.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataReader.cs
```diff
@@ -14,9 +15,9 @@ namespace Microsoft.Data.Entity.Sqlite.Design.ReverseEngineering
 {
     public class SqliteMetadataReader
     {
-        private readonly SqliteReverseTypeMapper _typeMapper;
+        private readonly SqliteTypeMapper _typeMapper;
 
-        public SqliteMetadataReader([NotNull] SqliteReverseTypeMapper typeMapper)
+        public SqliteMetadataReader([NotNull] SqliteTypeMapper typeMapper)
```
mikary: `IRelationalTypeMapper`?

natemcmaster: Same difference. This is going to be pulled out in #3337 anyways.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerDesignTimeMetadataProviderFactory.cs
```diff
@@ -25,6 +26,7 @@ public override void AddMetadataProviderServices([NotNull] IServiceCollection se
                 .AddSingleton<IRelationalAnnotationProvider, SqlServerAnnotationProvider>()
                 .AddSingleton<SqlServerLiteralUtilities>()
                 .AddSingleton<ConfigurationFactory, SqlServerConfigurationFactory>()
+                .AddSingleton<SqlServerTypeMapper>()
```
ajcvickers: @natemcmaster @bricelam I would like to understand the philosophy around which design-time services are registered and how they are registered. My questions:
* Do we expect runtime services to be available without them being added here, or are the two configurations completely separate? If we do expect runtime services to be available, then we shouldn't be registering the type mapper again.
* Are we following the same pattern as for the runtime by always resolving contracts through contract interfaces and never through non-interface types? If so, then most of these registrations should be added as interfaces, and all constructors resolved by DI should only use interfaces.
* How are we dealing with provider-specific services? Is it always true that only one provider will have services registered at a time? This isn't true for runtime services, so if we are using runtime services at design time, then it won't be true at design time either. Assuming it is not true, how do we select the correct service for the provider that is currently being used? At runtime, the UseSqlServer, etc. methods coupled with the scope of the context and the indirection delegates registered by the provider are used. Is this also what is used at design time?

natemcmaster:
Updated. RelationalTypeMapping now has a property `ClrType`. NB: To disambiguate between the extension method that gets mappings based on value.GetType() and getting a type by name, I've renamed the extension method.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerDesignTimeMetadataProviderFactory.cs
natemcmaster: > Do we expect runtime services to be available without them being added here, or are the two configurations completely separate?

Currently, we do not expect runtime services to be added, only services needed for design.

> Are we following the same pattern as for the runtime by always resolving contracts through contract interfaces and never through non-interface types?

Not yet. Currently we are inconsistent about this. Should we update?

> How are we dealing with provider-specific services? Is it always true that only one provider will have services registered at a time?

Yes, only one provider. Design-time is invoked by the command line with the provider specified as a required parameter. Each provider adds the services they require.

natemcmaster: Somewhate related: this PR will need to rebase against https://github.com/aspnet/EntityFramework/pull/3369 before merging, which improved design-time to use the startup convention used by other commands.

ajcvickers: @natemcmaster Thanks for clarifying. I think we should only depend on interface contracts throughout the stack, so yes, I think the design time contracts should be updated, but this can be done as a separate work item. It should be done for RC because it is potentially public surface.

natemcmaster: Will add an issue to ensure all services follow this pattern. I just updated to use `IRelationalTypeMapper` instead.

src/EntityFramework.Relational/Storage/RelationalTypeMapping.cs
```diff
@@ -11,18 +11,22 @@ namespace Microsoft.Data.Entity.Storage
 {
     public class RelationalTypeMapping
     {
-        public RelationalTypeMapping([NotNull] string defaultTypeName, DbType? storeType = null)
+        public RelationalTypeMapping([NotNull] string defaultTypeName, [CanBeNull] Type clrType, DbType? storeType = null)
```
ajcvickers: In what situation can clrType be null?

natemcmaster: Sqlite.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
```diff
@@ -104,8 +105,14 @@ private void AddColumns(EntityTypeBuilder builder, Table table)
 
         private void AddColumn(EntityTypeBuilder builder, Column column)
         {
-            // TODO log bad datatypes
-            var clrType = _typeMapper.GetClrType(column.DataType, nullable: column.IsNullable);
+            // TODO log bad datatypes/catch exception
+            var clrType = _typeMapper.GetMapping(column.DataType)?.ClrType;
```
ajcvickers: Code for GetMapping seems to throw if it can't return a mapping, so why code here to handle it returning null?

src/EntityFramework.Sqlite/Storage/Internal/SqliteTypeMapper.cs
```diff
@@ -50,6 +52,47 @@ public SqliteTypeMapper()
...
+        ///     Returns a clr type for a SQLite column type. Defaults to typeof(string).
+        ///     It uses the same heuristics from
+        ///     <see href="https://www.sqlite.org/datatype3.html">"2.1 Determination of Column Affinity"</see>
+        /// </summary>
+        public override RelationalTypeMapping GetMapping(string typeName)
+        {
+            if (string.IsNullOrEmpty(typeName))
+            {
+                return _default;
+            }
```
ajcvickers: Contract on interface says that typeName cannot be null, so either the contract is wrong, or this code is not needed.

natemcmaster: This handles the empty-type oddity of SQLite's loose type affinities. We want to still return a default type mapping here.

src/EntityFramework.Sqlite.Design/ReverseEngineering/SqliteMetadataModelProvider.cs
natemcmaster: Will remove this reduntancy.

src/EntityFramework.Relational/Storage/RelationalTypeMapping.cs
ajcvickers: What does null mean? In other words, in what situation does null get set and how is that null value consumed?

src/EntityFramework.Sqlite/Storage/Internal/SqliteTypeMapper.cs
ajcvickers: So if it is acceptable to pass null to the method, then the parameter should be marked as CanBeNull in the contract and all implementors should honor that.

src/EntityFramework.Relational/Storage/RelationalTypeMapping.cs
natemcmaster: Another situation: (see [RelationalTypeMapperExtensions.cs](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Storage/RelationalTypeMapperExtensions#L11)) - a caller asks for mapping of a null object or of DBNull. It seems unclear to me when this would every be useful, but we already have it in our code. Should we remove?

src/EntityFramework.Sqlite/Storage/Internal/SqliteTypeMapper.cs
natemcmaster: I think `[NotNull]` is a better contract. SQLite is the one-off case. The situation I want to handle here is an empty string, not necessary null reference type. If you want, we can throw for null but let an empty string pass through.

src/EntityFramework.Relational/Storage/RelationalTypeMapping.cs
ajcvickers: I would move the _nullTypeMapping singleton into the RelationalTypeMapping class and have it use a private constructor and then ensure anyone using the public constructor must pass in a CLR type.

natemcmaster: Sounds like a good solution. I'll add this.

src/EntityFramework.Sqlite/Storage/Internal/SqliteTypeMapper.cs
ajcvickers: If a caller can ever pass null and it not be an error, then the contract of NotNull is wrong. I think it is okay, but unusual, to still allow empty strings in this case.

natemcmaster:
@ajcvickers I've updated with your feedback.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -204,8 +209,18 @@ public virtual void AddPropertiesToModel([NotNull] Entity.Metadata.Model relatio
...
                 // Note: foreign key properties appear just like any other property in the relational model.
-                Type clrPropertyType;
-                if (!SqlServerTypeMapping._sqlTypeToClrTypeMap.TryGetValue(tc.DataType, out clrPropertyType))
+
+                Type clrPropertyType = null;
+                try
+                {
+                    clrPropertyType = _typeMapper.GetMapping(tc.DataType).ClrType;
+                }
+                catch (NotSupportedException)
```
ajcvickers: Something is wrong if we have to use exceptions to control non-exceptional program flow. It probably means we need to have a non-throwing way to get this information from the type mapper.

src/EntityFramework.Relational/Storage/RelationalTypeMapping.cs
```diff
@@ -11,18 +11,31 @@ namespace Microsoft.Data.Entity.Storage
...
+
             StoreType = storeType;
+            ClrType = clrType;
         }
 
         public virtual string DefaultTypeName { get; }
 
         public virtual DbType? StoreType { get; }
 
+        [CanBeNull]
```
ajcvickers: We made a decision some time ago to consistently not use these attributes for return values. I can see why it makes sense to do so here, but that could be said of lots of other code too. I think we should either be consistent and not use the attribute here or revisit with the team the decision we made.

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerMaxLengthMapping.cs
```diff
@@ -11,8 +12,8 @@ public class SqlServerMaxLengthMapping : RelationalTypeMapping
     {
         private readonly int _maxSpecificSize;
 
-        public SqlServerMaxLengthMapping([NotNull] string defaultTypeName, DbType? storeType = null)
-            : base(defaultTypeName, storeType)
+        public SqlServerMaxLengthMapping([NotNull] string defaultTypeName, Type clrType, DbType? storeType = null)
```
ajcvickers: clrType needs a NotNull attribute. Did the API consistency tests not pick this up? If so, we should file a bug on that.

src/EntityFramework.Relational/Storage/RelationalSizedTypeMapping.cs
```diff
@@ -10,19 +11,17 @@ namespace Microsoft.Data.Entity.Storage
 {
     public class RelationalSizedTypeMapping : RelationalTypeMapping
     {
-        public RelationalSizedTypeMapping(
-            [NotNull] string defaultTypeName,
-            DbType? storeType,
+        public RelationalSizedTypeMapping([NotNull] string defaultTypeName, 
+            [CanBeNull] Type clrType,
```
ajcvickers: Why CanBeNull here when base contract is NotNull?

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
natemcmaster: Agreed. I've added a TryGet version in https://github.com/aspnet/EntityFramework/pull/3412 (https://github.com/aspnet/EntityFramework/pull/3412/files#diff-3e11ae5d5fae6bf61730a666b0764bbd)

src/EntityFramework.Relational/Storage/RelationalSizedTypeMapping.cs
natemcmaster: Missed this one. Will update.

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerMaxLengthMapping.cs
natemcmaster: Will add.

ajcvickers:
:shipit:

PR #3372, opened 2015-10-07
[TINY] Can use IsPrimaryKey on an alternate key in absence of a primary key

lajones:
Found a small bug where calling IsPrimaryKey on an alternate key on an EntityType that does not have a PK causes an exception. RevEng comes across this scenario for e.g. AdventureWorks. Changed to no longer throw and added a test.

### DIFF
--- a/src/EntityFramework.Core/Metadata/KeyExtensions.cs
+++ b/src/EntityFramework.Core/Metadata/KeyExtensions.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Data.Entity.Metadata
         {
             Check.NotNull(key, nameof(key));
 
-            return key.EntityType.GetPrimaryKey() == key;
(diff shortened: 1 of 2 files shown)


smitpatel:
:shipit:

src/EntityFramework.Core/Metadata/KeyExtensions.cs
```diff
@@ -12,7 +12,7 @@ public static bool IsPrimaryKey([NotNull] this IKey key)
         {
             Check.NotNull(key, nameof(key));
 
-            return key.EntityType.GetPrimaryKey() == key;
+            return key == key.EntityType.FindPrimaryKey();
         }
```
smitpatel: we have `IsPrimaryKey()` on key if it can be used.

lajones:
Checked in with commit f3c7726.

PR #3373, opened 2015-10-07
[SMALL] RevEng: Remove unused usings from EntityType template

### LINKED ISSUE #3229: RevEng: Unnecessary usings in entity class code
Entity classes include the following using directives.

```C#
using Microsoft.Data.Entity;
```

I can't think of any entity class code that would require them.

lajones:
Fixes #3229. Fix is tiny - just affects a large number of expected results files.


smitpatel:
:shipit:

lajones:
Checked in with commit 1f3f1eb.

PR #3375, opened 2015-10-08
Add validation for unique column names

### LINKED ISSUE #2813: Relational: Add a validation rule to ensure unique column names
This convention would run after the model is built and set unique column names for all the properties with default mapping that conflict with another column name.

smitpatel:
resolves #2813


AndriySvyryd:
:shipit:

smitpatel:
checked in with e6e9449d20712b24583cd3bb162abdf61c9d2948

PR #3379, opened 2015-10-08
Bring anonymous property casing inline with Diagnostics Source usage recommendations

avanderhoorn:


avanderhoorn:
// @rynowak @divega @anpete

rynowak:
The suggestion here is to make these camelCase instead of PascalCase. It's more natural for consumers to write it that way and the consumers are the ones that have the harder job of matching things up correctly.

anpete:
This is unfortunate. Why can't it just be case insensitive?

divega:
I agree with @anpete. Why is camelCase more natural by the way?

anpete:
@divega I think because the props are mapped to arguments on the consumer side.

rynowak:
It's more natural because the consumer is going to write a method definition, and they would typical write the arguments that way.

> This is unfortunate. Why can't it just be case insensitive?

These are programmatic identifiers. Making the top level items case-insensitive makes it seem like other parts of the system should be case-insensitive, which ends with you having case-insensitive properties on the proxy objects themselves. I'd rather not go down that route.

anpete:
As @avanderhoorn has discovered, this seems like a pit of failure. It needs to be idiomatic on both sides.

rynowak:
Actually, yeah that makes a little sense to think of it that way. Idiomatic for both sides. OK I'm convinced.

avanderhoorn:
Closing this off given that @rynowak is going to be making the changes at his end.

PR #3381, opened 2015-10-08
Suppress obsolete warning for TelemetrySource (temporarily)

anpete:


anpete:
cc @bricelam is case I did this incorrectly :fried_shrimp:

bricelam:
Hmm, I don't see an [option for noWarn](https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime/ProjectReader.cs#L453-L474); this may not be supported in `project.json`

bricelam:
...but we get [three suppressed](https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Compilation.CSharp.Common/CompilerOptionsExtensions.cs#L28-L30) for free.

bricelam:
Ah, we want #2042

bricelam:
:shipit: (can remove them from `project.json` if you want)

anpete:
Thanks @bricelam will use #pragma I guess.

Eilon:
Do we have a bug to track moving to the new DiagnosticsSource and removing the `#pragmas`?

PR #3385, opened 2015-10-08
:racehorse: Widen tables for benchmarks

rowanmiller:
The domain model used for many of the benchmarks only had a few columns
per table, which is probably not representative of the most common
scenarios that EF is used in. This expands them to have more columns.

I also collapsed data seeding into the test fixture, because now it is helpful
to have the helper methods for creating entities with their fields populated
available in the test code.

I did look at using AdventureWorks for all the tests (we use it for the
initialization tests), but it gets hard to zero in on a particular
scenario using the data patterns and counts that are present in the
database.


rowanmiller:
BTW one side effect of this is that we will need to clear out our benchmark history. There really isn't a way to adjust the existing results to retrospectively adjust for these changes. Unfortunate to loose the nice burn down graph of the perf improvements :-1:

anpete:
@rowanmiller Re-run the new code against old commits :-P

anpete:
Or, better, make the new graph go down again soon :-)

natemcmaster:
:shipit:

Issue #3103, opened 2015-09-11
Must be reducible node exception when performing Join


kskalski:
Doing following join:
`ScheduledPayouts.Join(Assets, p => p.Good.Id, a => a.Good.Id, (p, a) => new Tuple<Asset, ScheduledPayout>(a, p));`

throws following exception:
```
   w System.Linq.Expressions.Expression.ReduceAndCheck()
   w System.Linq.Expressions.Expression.ReduceExtensions()
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteExtensionExpression(Expression expr, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteMemberExpression(Expression expr, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.Rewrite[T](Expression`1 lambda)
   w System.Linq.Expressions.Expression`1.Accept(StackSpiller spiller)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteLambdaExpression(Expression expr, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.Add(Expression node)
   w System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.AddArguments(IArgumentProvider expressions)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteMethodCallExpression(Expression expr, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.Add(Expression node)
   w System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.AddArguments(IArgumentProvider expressions)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteMethodCallExpression(Expression expr, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   w System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.Add(Expression node)
[...]

PR #3386, opened 2015-10-09
Partial fix to #3103 - Must be reducible node exception when performing Join

maumar:
Problem is in a query with join, that has a navigation property in inner key selector.
We normally try to expand the navigation property into a join clause, however if done for inner key selector, leads to chicken-and-egg problem:

e.g.
from p in ctx.Payouts
join a in ctx.Assets on p.Id equals a.Good.Id
select p

translated to:
from p in ctx.Payouts
join g in ctx.Good on a.Id equals g.Id // invalid, because a was not declared yet
join a in ctx.Assets on g.Id equals a.Id

(alternatively)

from p in ctx.Payouts
join a in ctx.Assets on g.Id equals a.Id // invalid, because g was not declared yet
join g in ctx.Good on a.Id equals g.Id

Is some cases this can we worked around by translating navigation property into a FK:
from p in ctx.Payouts
join a in ctx.Assets on p.Id equals a.GoodId
select p

This is only possible if we actually have foreign key on this side of the navigation, and if the navigation itself is simple (not nested, not composite and we are accessing PK after the navigation).

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
```diff
@@ -251,6 +261,29 @@ var innerQuerySourceReferenceExpression
                 ?? base.VisitMember(memberExpression);
         }
 
+        private Expression CreateForeignKeyMemberAccess(MemberExpression memberExpression, INavigation navigation)
+        {
+            var principalKey = navigation.ForeignKey.PrincipalKey;
+            if (principalKey.Properties.Count == 1)
```
anpete: Curious, why does composite not work?

maumar:
opened a new one

PR #3387, opened 2015-10-09
Align Property/Navigation discovery with typemapping

### LINKED ISSUE #2588: Align property discovery convention with type mapping
When a TEntity CLR type is passed to `ModelBuilder.Entity<TEntity>()` we need a mechanism to identify scalar properties and navigation properties (plus complex type/value object properties in the future) on it.
 
Currently `PropertyDiscoveryConvention` uses `SharedTypeExtensions.IsPrimitive()` to identify "primitive" properties on a given entity's CLR type. 

Unfortunately the heuristic perpetuates the idea of a canonical set of primitive types supported by EF, while in reality type support should vary on a provider basis (e.g. while SQL Server does not support many unsigned int types, other providers do not support DateTimeOffset and some providers may have native support for types that EF does not know about, etc.) and in the long term even become extensible via user-provided custom type mappers.

One way to improve this would be to have an API that the convention can use to find out if a specific property type can be handled by the "current" type mapping (i.e. the provider's type mapping and in the future, any registered custom type mappers). 

- If the type can be handled then we know it can be treated as a scalar type by EF. 
- If the type cannot be handled as a scalar, we'll try to shred it into its individual components (e.g. the property could be an entity or collection of entities, etc. so we would need to reflect on it to further discover properties). 
- If none of the above succeeds we would need to fail the mapping of the property. 

Note that this approach assumes that at model building time we would have access to type mapping (which I am not sure is currently possible).

### LINKED ISSUE #2612: Property discovery convention should throw when property type cannot be mapped
See #2588.

smitpatel:
resolves #2588 & #2612 
Extends `PropertyDiscovery`, `InverseProperty`, `RelationshipDiscovery`, `ForeignKeyAttribute` conventions in relational which uses typemapper to determine primitive property/navigation.
Core uses the current `IsPrimitive` function for the same.
Relational and its derived conventionset builder takes typemapper as argument
Add PropertyMappingConvention which verifies that all added properties are mapped to datastore and all candidate properties from CLR type are either added to the model/ignored.


smitpatel:
Open to suggestions for the names of the conventions.

src/EntityFramework.Core/Metadata/Conventions/Internal/PropertyMappingConvention.cs
```diff
@@ -0,0 +1,87 @@
...
+
+                    clrProperties.ExceptWith(entityType.Properties.Select(p => p.Name));
+
+                    clrProperties.ExceptWith(entityType.Navigations.Select(p => p.Name));
+
+                    var entityTypeBuilder = modelBuilder.Entity(entityType.ClrType, ConfigurationSource.Convention);
+
+                    clrProperties.RemoveWhere(p => entityTypeBuilder.IsIgnored(p, ConfigurationSource.Convention));
+
+                    if (clrProperties.Count > 0)
```
whyi: Would
```
if (clrProperties.Any()) 
```
read better?

src/EntityFramework.Core/Metadata/Conventions/Internal/PropertyMappingConvention.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Linq;
+using System.Reflection;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
+{
+    public class PropertyMappingConvention : IModelConvention
```
AndriySvyryd: PropertyMappingValidationConvention?

src/EntityFramework.Core/Metadata/Conventions/Internal/PropertyMappingConvention.cs
```diff
@@ -0,0 +1,87 @@
...
+
+                    var entityTypeBuilder = modelBuilder.Entity(entityType.ClrType, ConfigurationSource.Convention);
+
+                    clrProperties.RemoveWhere(p => entityTypeBuilder.IsIgnored(p, ConfigurationSource.Convention));
+
+                    if (clrProperties.Count > 0)
+                    {
+                        foreach (var clrProperty in clrProperties)
+                        {
+                            var actualProperty = entityType.ClrType.GetRuntimeProperties().First(pi => pi.Name == clrProperty);
```
AndriySvyryd: Use GetRuntimeProperty(clrProperty)

src/EntityFramework.Core/Metadata/Conventions/Internal/PropertyMappingConvention.cs
```diff
@@ -0,0 +1,87 @@
...
+                                {
+                                    throw new InvalidOperationException(CoreStrings.NavigationNotAdded(actualProperty.Name, entityType.Name));
+                                }
+                            }
+                            else
+                            {
+                                targetType = actualProperty.PropertyType;
+                                targetType = targetType.TryGetSequenceType() ?? targetType;
+                                targetType = targetType.UnwrapNullableType();
+                                if (!targetType.GetTypeInfo().IsInterface)
```
AndriySvyryd: You could move this check to IsCandidateProperty

AndriySvyryd:
:shipit:

src/EntityFramework.Core/Metadata/Conventions/Internal/PropertyMappingConvention.cs
smitpatel: Do you mean inside `IsCandidateProperty` function? Filtering out Interface type properties since we do not add them to model.

AndriySvyryd: Yes, since we would not add them to the model then they are not candidates and thus should be filtered out early.

ajcvickers: Sorry, I haven't looked at the code closely, just seeing the comments. When you say "we would not add them to the model" does this mean we throw indicating we don't know how to map them? I think the only thing we should silently ignore (not map) based on our current thinking would be read-only non-navigation properties, since these are likely to be computed properties and we think in the majority of cases the application would want them not mapped. (Although it could be argued that we also throw for these.) For every other property that we can't map we should throw unless it is explicitly ignored.

AndriySvyryd: @ajcvickers So you think we should throw on properties of interface or abstract type, private properties, etc.?

ajcvickers: @AndriySvyryd Private properties, probably not. The idea is to throw on anything that the user might reasonably expect to be included in the model but isn't because of limitations in what we can map. So, things that are likely intended to be navigation properties, but aren't mapped as such, then yes. Maybe we should get together and figure out whether to throw or ignore for various cases so we are all on the same page. //cc @rowanmiller @divega

src/EntityFramework.Core/Metadata/Conventions/Internal/PropertyMappingConvention.cs
smitpatel: Micro-optimization :wink:

PR #3392, opened 2015-10-09
Fix minor bugs in TableSelectionSet

natemcmaster:
Also adding a default constructor for syntactic sugar.

cc @lajones


lajones:
:shipit:

PR #3395, opened 2015-10-09
[SMALL] Record architecture for benchmarks

rowanmiller:
Capture the architecture that a run was performed on. Also adding logic
to upgrade an existing benchmark database to add the required column if
not present.


lajones:
:shipit:

PR #3398, opened 2015-10-09
Query: Context level tracking flag and AsTracking modifier

anpete:


src/EntityFramework.Core/Query/CompiledQueryCacheKeyGenerator.cs
```diff
@@ -27,18 +29,22 @@ protected CompiledQueryCacheKey GenerateCacheKeyCore([NotNull] Expression query,
...
             private readonly bool _async;
 
-            public CompiledQueryCacheKey([NotNull] string query, [NotNull] IModel model, bool async)
+            public CompiledQueryCacheKey(
+                [NotNull] string query, [NotNull] IModel model, bool trackQueryResults, bool async)
             {
                 _query = query;
                 _model = model;
+                _trackQueryResults = trackQueryResults;
                 _async = async;
```
whyi: Although this is existing code, I wonder if this was discussed. That the 'async' is a keyword but used as a variable name here. Isn't there naming convention for this case? (conflict between variable name and keyword).
* modified : adding @lukewaters to the discussion.

test/EntityFramework.Core.FunctionalTests/AsNoTrackingTestBase.cs
```diff
@@ -71,7 +71,7 @@ var customers
                         .ToList();
 
                 Assert.Equal(6, customers.Count);
-                Assert.Equal(1, context.ChangeTracker.Entries().Count());
+                Assert.Equal(0, context.ChangeTracker.Entries().Count());
```
whyi: Assert.IsFalse(context.ChangeTracker.Entries().Any(), "because ....") reads better?

src/EntityFramework.Core/Query/CompiledQueryCacheKeyGenerator.cs
```diff
@@ -12,12 +12,14 @@ namespace Microsoft.Data.Entity.Query
...
         private readonly IModel _model;
+        private readonly bool _trackQueryResults;
 
-        public CompiledQueryCacheKeyGenerator([NotNull] IModel model)
+        public CompiledQueryCacheKeyGenerator([NotNull] IModel model, [NotNull] DbContext context)
         {
             Check.NotNull(model, nameof(model));
 
             _model = model;
+            _trackQueryResults = context.ChangeTracker.TrackQueryResults;
```
ajcvickers: Can this get stale if the value is mutated in the ChangeTracker after the CompiledQueryCacheKeyGenerator is created? I'm not sure what the lifetime of the CompiledQueryCacheKeyGenerator is.

src/EntityFramework.Core/Query/CompiledQueryCacheKeyGenerator.cs
ajcvickers: Looks fine as it to me.

src/EntityFramework.Core/Query/CompiledQueryCacheKeyGenerator.cs
anpete: I think you are right. I have a test but obv something is wrong.

ajcvickers: In other words, do both of these work:
```C#
using (var context = new Context())
{
  context.ChangeTracker.TrackQueryResults = false;
  var foo = context.SomeSet.ToList(); // Not tracked

  context.ChangeTracker.TrackQueryResults = true;
  var foo = context.SomeOtherSet.ToList(); // Tracked
}
```
```C#
using (var context = new Context())
{
  context.ChangeTracker.TrackQueryResults = false;
  foreach(var e in context.SomeSet) // Not tracked
  {
    context.ChangeTracker.TrackQueryResults = true;
    var foo = context.SomeOtherSet.ToList(); // Tracked
  }
}
```

anpete: Yep, will add. The test I have uses two contexts.

test/EntityFramework.MicrosoftSqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerFixture.cs
```diff
@@ -58,7 +58,12 @@ public override ComplexNavigationsContext CreateContext(SqlServerTestStore testS
             optionsBuilder.UseSqlServer(testStore.Connection);
 
             var context = new ComplexNavigationsContext(_serviceProvider, optionsBuilder.Options);
+
+            context.ChangeTracker.AutoDetectChangesEnabled = false;
```
ajcvickers: Why is AutoDetectChangesEnabled being disabled for all these tests?

anpete: Just making consistent the way we do fixtures for query only tests. I guess we could remove them as it shouldn't affect perf in the read-only case, right?

ajcvickers: I think we should remove them. It should not affect the perf or functionality if detecting changes isn't needed, and if somehow detecting changes is being used, then disabling it might mask a test or product bug.

anpete: Will do.

anpete:
@ajcvickers Changes made.

ajcvickers:
:shipit:

PR #3399, opened 2015-10-09
Move SqlStatementExecutor functionality into RelationalCommand

### LINKED ISSUE #3205: SqliteRelationalConnection enables foreign keys each time open is called
The SqliteRelationalConnection should only enable foreign keys when the underlying connection changes state to Open

mikary:
Incremental change for #3110 
Fix #3205 - `SqliteRelationalConnection`

Additional work to be done relates to fixing calls on `SqlStatementExecutor.ExecuteReader` which currently do not dispose `DbCommands`, removing `CreateDbCommand` from the public interface of `IRelationalCommand`, and general cleanup to consumers of the new API

### DIFF
--- a/src/EntityFramework.Relational/Storage/Internal/RelationalCommand.cs
+++ b/src/EntityFramework.Relational/Storage/Internal/RelationalCommand.cs
@@ -1,30 +1,229 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Data.Common;
+using System.Diagnostics.Tracing;
+using System.Threading;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Storage.Internal
 {
     public class RelationalCommand : IRelationalCommand
     {
+#pragma warning disable 0618
         public RelationalCommand(
+            [NotNull] ISensitiveDataLogger logger,
+            [NotNull] TelemetrySource telemetrySource,
             [NotNull] string commandText,
             [NotNull] IReadOnlyList<RelationalParameter> parameters)
         {
+            Check.NotNull(logger, nameof(logger));
+            Check.NotNull(telemetrySource, nameof(telemetrySource));
             Check.NotNull(commandText, nameof(commandText));
             Check.NotNull(parameters, nameof(parameters));
 
+            Logger = logger;
+            TelemetrySource = telemetrySource;
             CommandText = commandText;
             Parameters = parameters;
         }
 
+        protected virtual ISensitiveDataLogger Logger { get; }
+
+        protected virtual TelemetrySource TelemetrySource { get; }
+#pragma warning restore 0618
+
         public virtual string CommandText { get; }
 
         public virtual IReadOnlyList<RelationalParameter> Parameters { get; }
+
+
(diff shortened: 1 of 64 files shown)


src/EntityFramework.Relational/Storage/Internal/RelationalCommand.cs
```diff
@@ -1,31 +1,230 @@
...
+
         public virtual string CommandText { get; }
 
         public virtual IReadOnlyList<RelationalParameter> Parameters { get; }
 
+
+        public virtual void ExecuteNonQuery([NotNull] IRelationalConnection connection)
+            => Execute(
+                Check.NotNull(connection, nameof(connection)),
+                (c) =>
```
anpete: Parens.

src/EntityFramework.Relational/Storage/RelationalDataReader.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public class RelationalDataReader : IDisposable
+    {
+        private readonly DbCommand _command;
+        private readonly DbDataReader _reader;
+
+        private bool _diposed;
```
anpete: Typo

src/EntityFramework.Sqlite/Storage/Internal/SqliteRelationalConnection.cs
```diff
@@ -44,13 +45,32 @@ public class SqliteRelationalConnection : RelationalConnection
         public override void Open()
         {
             base.Open();
-            EnableForeignKeys();
+
+            _openedCount++;
```
anpete: Isn't counting also done in the base class?

test/EntityFramework.Relational.Tests/Storage/RelationalCommandTest.cs
```diff
@@ -0,0 +1,769 @@
...
+using Microsoft.Data.Entity.Storage.Internal;
+using Microsoft.Data.Entity.Tests.TestUtilities;
+using Microsoft.Data.Entity.TestUtilities;
+using Microsoft.Data.Entity.TestUtilities.FakeProvider;
+using Microsoft.Extensions.Logging;
+using Xunit;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public class RelationalCommandTest
```
anpete: Wow, this is a lot of setup code for enabling a few unit tests. Would have probably been better just to create simple functionals that use real providers. And if you are going to do this style, why not just use Moq and avoid a bunch of work?

src/EntityFramework.Sqlite/Storage/Internal/SqliteRelationalConnection.cs
mikary: This code does duplicate the counting in the base class. In #3145 we discussed the possibility of adding lifetime events or exposing the connection count in the `RelationalConnection`, but at the time we favored doing the counting in this class as a special case, rather than expanding the API surface

test/EntityFramework.Relational.Tests/Storage/RelationalCommandTest.cs
mikary: The FakeProvider classes do add a fair amount of code and they were originally something I wrote when more of the ADO objects were being passed around. I also spoke with @bricelam  at the time and I believe he noted that there are compatibility issues between Moq and Core and that we would like to get away from using it.

If we are getting away from Moq, I would expect these classes to be useful for other areas, such as `RelationalConnectionTest`, `ReaderModificationCommandBatchTest`, and any future relational unit tests that depend on provider implementations.

While we could cover most of these scenarios in functional tests, we would end up with more tests like the `CommandConfigurationTest` that seem fairly fragile and tend to fall half way between unit and functional test

anpete:
:shipit:

PR #3412, opened 2015-10-12
Improvements to SQLite metadata reader

### LINKED ISSUE #3404: RevEng: Respect IsStoreGenerated
Looking at the SqlServerMetaDataReader, I cannot see any TODOs for supporting the "IsStoreGenerated" property, and the sys.columns query does not include that information. (But I could be overlooking something)

natemcmaster:
Fix bug in primary key ordering.

SQLite reverse typemapping only adds ColumnType annotation if it has a custom type, instead of adding for every type.

Note:
Once approved, will merge into dev after #3367 is merged.

Fixes #3404

### DIFF
--- a/src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
+++ b/src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CannotFindTypeMappingForColumn" xml:space="preserve">
+    <value>Could not find type mapping for column {columnName} with data type {dateType}. Skipping column.</value>
+  </data>
   <data name="ConnectionStringRequired" xml:space="preserve">
     <value>ConnectionString is required to generate code.</value>
   </data>
--- a/src/EntityFramework.Sqlite.Design/ReverseEngineering/CSharpNamer.cs
+++ b/src/EntityFramework.Sqlite.Design/ReverseEngineering/CSharpNamer.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Relational.Design.Utilities;
+using Microsoft.Data.Entity.Utilities;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Data.Entity.Sqlite.Design.ReverseEngineering
+{
+    public class CSharpNamer<T>
+    {
+        private readonly Func<T, string> _nameGetter;
+        private Dictionary<T, string> _nameCache = new Dictionary<T, string>();
+
+        public CSharpNamer([NotNull] Func<T, string> nameGetter)
+        {
+            Check.NotNull(nameGetter, nameof(nameGetter));
+
+            _nameGetter = nameGetter;
+        }
+
+        public virtual string GetName([NotNull] T item)
+        {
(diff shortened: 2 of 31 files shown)


src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CannotFindTypeMappingForColumn" xml:space="preserve">
+    <value>Could not find type mapping for column {columnName} width data type {dateType}. Skipping column.</value>
```
smitpatel: typo: `width`

test/EntityFramework.Sqlite.Design.Tests/CSharpNamerTest.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.Data.Entity.Sqlite.Design
+{
+    public class CSharpNamerTest
+    {
+        [Theory]
+        [InlineData("Name with space", "Name_with_space")]
+        [InlineData("namespace", "_namespace")]
+        [InlineData("@namespace", "@namespace")]
+        [InlineData("8ball", "_8ball")]
+        public void Santizes_name(string input, string output)
```
lajones: Nit: typo -> Sanitizes

test/EntityFramework.Sqlite.Design.Tests/RelationalMetadataModelProviderTest.cs
```diff
@@ -0,0 +1,569 @@
...
+                }
+            };
+            var model = (EntityType)_provider.GetModel(info).EntityTypes.Single();
+
+            Assert.Equal(keyProps, model.GetPrimaryKey().Properties.Select(p => p.Relational().ColumnName).ToArray());
+        }
+
+        [Fact]
+        public void Indexes_and_alternate_keys()
+
```
lajones: Nit: extra blank line

lajones:
:shipit:

natemcmaster:
Fixed typos and formatting. :watch: on #3367

src/EntityFramework.Sqlite.Design/ReverseEngineering/CSharpNamer.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            Check.NotNull(item, nameof(item));
+
+            if (_nameCache.ContainsKey(item))
+            {
+                return _nameCache[item];
+            }
+
+            var name = CSharpUtilities.Instance.GenerateCSharpIdentifier(GenerateName(item), null);
+            _nameCache[item] = name;
```
whyi: Just a question. Why is this line not
```
_nameCache.Add(item, name);
```
? As it seems that there's no corresponding key exists in _nameCache

natemcmaster: This does the same thing.

whyi: In this context I think [] is add or update if I remember correctly while .Add method would throw an exception. In single-thread scenario the code would do the same thing, given that you're already calling ContainsKey above. However, what would happen if someone call the method in multi-threaded code? Would one that is written by thread 0 be overwritten by another thread 1 silently? Just by looking at the method name, GetName I think people would assume it is thread-safe (i.e read-only operation) but it isn't.. Any thought?

whyi: * Adding @lukewaters to the discussion.

natemcmaster: I don't think this matters. Even using `.Add()` does not guarantee thread-safeness in this section of code, which is why we have a our own utility class `ThreadSafeDictionaryCache`.

Issue #3368, opened 2015-10-07
Tests: Dead code in ChangeTrackingTestBase


anpete:
Lots of dead locals. Review whether they can be removed.

rowanmiller:
Fixed in https://github.com/aspnet/EntityFramework/commit/a80e7d3cd72c969e2fd1ea5bc4b0839085f085d1

PR #3413, opened 2015-10-12
Fix #3397, #3368: InMemoryStore should not leak references to interna…

### LINKED ISSUE #3397: Tests: ChangeTrackingInMemoryTest fails when run in isolation
Passes when run as part of the full suite.

anpete:
…l data object arrays.


ajcvickers:
:shipit:

PR #3414, opened 2015-10-13
Update System.Collections.Immutable version

pranavkm:


test/EntityFramework.Commands.FunctionalTests/project.json
```diff
@@ -6,6 +6,7 @@
     "EntityFramework.Commands": "7.0.0-*",
     "EntityFramework.Core.FunctionalTests": "7.0.0-*",
     "EntityFramework.MicrosoftSqlServer": "7.0.0-*",
+    "System.Collections.Immutable": "1.1.37",
```
divega: I don't understand why this is now needed or what it has to be this version (the EF runtime depends on 1.1.36).

divega: Anyway, thanks for fixing if this was blocking the build.

pranavkm: DNX picked up a new set of Roslyn packages earlier today. Perhaps that caused this issue?

divega: Makes sense. But it is weird. The reference wasn't even here before (was it picked up transitively from the runtime?) And the runtime still depends on 1.1.36, I think on purpose.

@bricelam should probably take a look when possible.

Again, thanks!

Issue #3288, opened 2015-09-30
RevEng: ModelBuilder Approach?


lajones:
For the future. Consider taking a ModelBuilder approach to building the relational and codeGen models. Are there any (e.g. error) situations where this won't work?

natemcmaster:
Also worth noting: SQLite.Design constructs a model from metadata using `ModelBuilder` and SQL server only makes call on `Model`. It may be possible to unite the implementations into one "relational" metadata => model process.

rowanmiller:
@lajones @natemcmaster is there any functional benefit in making this change? i.e. would we ever do this work at the expense of other features :smile:

rowanmiller:
BTW I don't just mean benefit from an end user perspective... maintainability etc. would also be a valid reason

natemcmaster:
We did this in our RC1 refactoring of the ScaffoldingModel factory.

Issue #3077, opened 2015-09-09
Reverse Engineer: allow customizing the model after it is generated, before being passed to code gen


rowanmiller:
We want a simple "seam" where folks can get the `Model` just before it is passed to code generation and manipulate it. This is particularly important since we are removing templating for the moment.

natemcmaster:
This is `MetadataReader.GetModel()`. The method is virtual and can be overwritten.

divega:
I think this item was to track how the application developer would do it, not the provider. @rowanmiller?

natemcmaster:
Ah, yes. Who are "folks"?

divega:
:smile:

divega:
Reopening for now.

abelepereira:
I didn't undertand if templating via .cshtml is being definitely replaced for the ```GetModel()``` approach or if the @rowanmiller 's "for the moment" means it will come back on a later release.
The cshtml templating was actually a pretty nice way to enable [the unsupported] pluralization: https://empiricalthoughts.wordpress.com/2015/09/12/entity-framework-7-database-model-pluralization-part-22/

divega:
We have decided to leave the choice of a template based code-gen technology open for now. It is true that Razor enabled plugging in your own pluralization logic... after all you can do pretty much anything with it, but it also leads to complicated templates that mix concerns like this with the actual code generation. One of the ideas we decided to explore is to provide a hook that allows any customization of the model (covered by this issue) plus very simple code gen at the end that just binds to the resulting model and focuses only on what code to generate and its formatting.

rowanmiller:
Closing in favor of https://github.com/aspnet/EntityFramework/issues/4038

PR #3423, opened 2015-10-13
Refactor reverse engineering API

### LINKED ISSUE #3404: RevEng: Respect IsStoreGenerated

natemcmaster:
This splits reverse engineering at several levels. The entry point for provider writers is `MetadataModelProvider.GetModel()`.
The framework provides some implementation helpers.

 - `MetadataModelProvider` has a protected method `AddNavigationProperties` which should be called on a mostly-ready `IModel`. This will add navprops for all foreign keys.
 - Providers can implement `RelationalMetadataModelProvider`. If given `SchemaInfo` and a type mapper, it will build and `IModel`.
 - Providers can override methods in `RelationalMetadataModelProvider` to get specific control over how metadata is build, if the default is unsuitable.

Closes #3404, #3077, #3288, #3247 and #3329.

TODO
 [ ] also update how a provider passes their `UseProvider()` method name to the code gen step.
 [ ] move everything to M.D.E.Scaffolding (.Internal)


test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/ReverseEngineering/ExpectedResults/E2E_AllFluentApi/TableWithUnmappablePrimaryKeyColumn.expected
```diff
@@ -1 +1 @@
-// Unable to generate EntityType TableWithUnmappablePrimaryKeyColumn. Unable to identify any primary key columns in the underlying SQL Server table [dbo].[TableWithUnmappablePrimaryKeyColumn].
+// Unable to generate EntityType TableWithUnmappablePrimaryKeyColumn. Unable to identify the primary key for table 'TableWithUnmappablePrimaryKeyColumn'.
```
lajones: Please include the schema name for SQL Server tables.

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
```diff
@@ -144,14 +145,14 @@ public void E2ETest_UseAttributesInsteadOfFluentApi()
...
-                            @"Could not find type mapping for column [dbo][AllDataTypes][geometryColumn] with data type geometry. Skipping column.",
-                            @"For column [dbo][PropertyConfiguration][PropertyConfigurationID]. This column is set up as an Identity column, but the SQL Server data type is tinyint. This will be mapped to CLR type byte which does not allow the SqlServerValueGenerationStrategy.IdentityColumn setting. Generating a matching Property but ignoring the Identity setting.",
-                            @"Could not find type mapping for column [dbo][TableWithUnmappablePrimaryKeyColumn][TableWithUnmappablePrimaryKeyColumnID] with data type hierarchyid. Skipping column.",
-                            @"Unable to identify any primary key columns in the underlying SQL Server table [dbo].[TableWithUnmappablePrimaryKeyColumn]."
+                            RelationalDesignStrings.CannotFindTypeMappingForColumn("AllDataTypes.geographyColumn", "geography"),
+                            RelationalDesignStrings.CannotFindTypeMappingForColumn("AllDataTypes.geometryColumn", "geometry"),
```
lajones: Include schema name as above. This applies to all error messages for SQL Server.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -23,577 +19,120 @@ namespace Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering
...
-                if (entityType.FindPrimaryKey() == null)
-                {
-                    var errorMessage = SqlServerDesignStrings.NoPrimaryKeyColumns(
-                        entityType.Relational().Schema,
-                        entityType.Relational().TableName);
-                    entityType.AddAnnotation(AnnotationNameEntityTypeError, errorMessage);
-                    Logger.LogWarning(errorMessage);
-                }
+                propertyBuilder.Metadata.SetMaxLength(null);
+                propertyBuilder.HasColumnType($"{column.DataType}({column.Scale})");
```
lajones: Add a comment here saying that for datetime types the "precision" is stored in the "scale" column.

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
```diff
@@ -1,285 +1,297 @@
...
         {
             CoreAnnotationNames.OriginalValueIndexAnnotation,
             CoreAnnotationNames.ShadowIndexAnnotation
         };
 
-        public const string AnnotationPrefix = nameof(RelationalMetadataModelProvider) + ":";
-        public const string AnnotationNameDependentEndNavPropName = AnnotationPrefix + "DependentEndNavPropName";
-        public const string AnnotationNamePrincipalEndNavPropName = AnnotationPrefix + "PrincipalEndNavPropName";
-        public const string AnnotationNameEntityTypeError = AnnotationPrefix + "EntityTypeError";
+        private CSharpNamer<Column> _columnNamer;
```
lajones: Just checking: column names can need to be uniquified too. E.g. 2 columns can clash if one is called `123Col` and the other is `_123Col`. So long as column names make sure that they don't use already-used column names you should be fine.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -23,577 +19,120 @@ namespace Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering
...
             }
 
-            AddForeignKeysToModel();
-        }
-
-
-        public virtual void AddPrimaryAndAlternateKeysToModel()
-        {
-            foreach (var keyValuePair in _tableToConstraintToColumnsMap)
+            // undo quirk in reverse type mapping to litters code with unnecessary nvarchar annotations
```
lajones: How come we can't avoid this in the type mapping itself?

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataReader.cs
```diff
@@ -251,20 +251,27 @@ private void GetForeignKeys()
...
                     var fromColumnName = reader.GetString(5);
                     var fromColumn = _tableColumns[ColumnKey(fkInfo.Table, fromColumnName)];
                     fkInfo.From.Add(fromColumn);
 
-                    var toColumnName = reader.GetString(6);
-                    var toColumn = _tableColumns[ColumnKey(fkInfo.PrincipalTable, toColumnName)];
-                    fkInfo.To.Add(toColumn);
+                    if (fkInfo.PrincipalTable != null)
+                    {
+                        var toColumnName = reader.GetString(6);
```
lajones: Feels like we should put out a warning that we can't locate the table on the principal end of the FK as well.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
natemcmaster: It's a quirk in how maxlength is handled by SQL Server's typemapper. Because this is specific to SQL Server, we need to override the base case. SQL Server's default type mapping for `System.String` is `nvarchar(max)` which does not equal `nvarchar` (which is actually `nvarchar(1)`).  In other words, ` typeMapper("nvarchar").DefaultTypeMapping != "nvarchar" `

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataReader.cs
natemcmaster: We do, just not here. See `RelationalMetadataModelProvider.AddPrimaryKey`.

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
natemcmaster: This is for the "codegen" model. I'll update so they are always unique.

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
natemcmaster: Will update.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
lajones: OK.

src/EntityFramework.Relational.Design/ReverseEngineering/RelationalMetadataModelProvider.cs
lajones: Thanks.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataReader.cs
lajones: Sort of. There we say we can't find the primary key on a table. Here we are unable to find the table at the other end of a foreign key. Though it's likely that the latter is caused by the former I'd like to say what code generation does about it here. Maybe some message like:

`For foreign key {fkDefinition}: unable to find principal table {principalTableName} in model. Skipping generation of foreign key.`

But feel free to add a TODO - don't let this hold up checking this in.

lajones:
:shipit: with agreed changes.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
divega: How does that work? Is it something we should correct on the layer below?

natemcmaster: I don't thinkg EF/ADO.NET needs to do anything to represent DateTime with limited precision. This is here so the data type flows through to migrations.

lajones: @divega - it's a quirk of the SQL Server sys tables. Datetime precision gets stored in the Scale column. Weird, huh?! (BTW we didn't see it in INFORMATION_SCHEMA because those views have the concept of a DatetimePrecision column - and we used that instead). What I mean here is that the datatype we're producing has what would normally be called "precision" coming form the "Scale" column. That is _not_ a typo and I thought it would be good to have a comment saying so.

divega: Ok, I might be missing context on this bug:

1. Although @bricelam filed a bug on switching from INFORMATION_SCHEMA to SYS, I haven't seen any solid argument yet to actually do it and I would like to understand better why we are doing it. 
2. I believe I am looking at SQL Server specific code and hence it seems that it is ok to compensate for a quirk here. Can you confirm that? Regardless, based on our recent conversations I am expecting the goal to be that that the code that translates from the database schema domain model to the IModel will be mostly common relational implementation. If that is really what we are trying to do then I believe the compensation from this quirk should happen in the layer in which we materialize the database schema domain model instances.

divega: Besides all that I agree with @lajones that a comment conveying why we have to do this would apply.

natemcmaster: @divega 

> Although I know @bricelam filed a bug on switching from INFORMATION_SCHEMA to SYS, I haven't seen any solid argument yet to actually do it and I would like to understand better why we are doing it.

I believe the reasonsing was that sys.* tables adhere to the T-SQL standard.

> I believe I am looking at SQL Server specific code and hence it seems that it is ok to compensate for a quirk here. Can you confirm that? 

Yes, this is SQL Server specific. It would be trivial to put into the metadata layer, but as this seems like a typemapping issue, I left it in the part of code that actually does type mapping.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataReader.cs
natemcmaster: My mistake. I misread which lines you were commenting on. We catch the missing principal table with `RelationalMetadataModelProvider.LogFailedForeignKey` which goes something like:

> Could not scaffold the foreign key 'dbo.DepTable (FkProp1)'. One of the referenced properties or the referenced table could not be found.

lajones: May be more helpful to say what the referenced table and properties are. But no big deal. :shipit:

natemcmaster: ...would have been trivial to do with a flat DTO... :no_mouth:

PR #3430, opened 2015-10-14
Refactor update surface to use RelationalCommand instead of DbCommand

mikary:
Factoring change to remove raw DbCommand use from Update
@AndriySvyryd @ajcvickers


AndriySvyryd:
:shipit:

PR #3431, opened 2015-10-14
Fix to #3103 - Must be reducible node exception when performing Join

maumar:
Problem:
We were incorrectly expanding navigations that appeared in the inner key selector of a join expression (right side).
Normally we expand nav props into additional join. E.g.:

from p in ctx.Payouts
join a in ctx.Assets on p.Good.Id equals a.Id
select p

converts to:

from p in ctx.Payouts
join g from ctx.Goods on p.Id equals g.Id
join a in ctx.Assets on g.Id equals a.Id
select p

However if navigation propery is in the inner key selector, like so:

from p in ctx.Payouts
join a in ctx.Assets on p.Id equals a.Good.Id
select p

we encounter chicken-and-egg problem:

from p in ctx.Payouts
join g in ctx.Good on a.Id equals g.Id // invalid, because a was not declared yet
join a in ctx.Assets on g.Id equals a.Id

or

from p in ctx.Payouts
join a in ctx.Assets on g.Id equals a.Id // invalid, because g was not declared yet
join g in ctx.Good on a.Id equals g.Id

Solution:

Some navigation properties can be translated into FKs, which removes the problem:

from c in Customers
join d in Details on c.Id equals d.Customer.Id
select c

can be translated into:

from c in Customers
join d in Details on c.Id equals c.CustomerId
select c

This can only be done for a subset of queries - join needs to be based on key property, FK must be pointing the right way, and key must not be composite.

For all other cases we translate this into a subquery:

from c in Customers
join d in Details on c.Id equals
(from c2 in Customers
 where c2.CustomerId equals d.Id
 select c2.Id).FirstOrDefault()

In case of nested navigation property we expand the first one into a subquery, and add extra joins to the subquery, just like we did originally. Only the first navigation is "problematic", the following ones can be safely translated to joins

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -12,6 +13,7 @@ using Microsoft.Data.Entity.Utilities;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Parsing;
 
 namespace Microsoft.Data.Entity.Query.ExpressionVisitors
@@ -21,6 +23,7 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors
         private readonly EntityQueryModelVisitor _queryModelVisitor;
         private readonly List<NavigationJoin> _navigationJoins = new List<NavigationJoin>();
 
+        private NavigationRewritingQueryModelVisitor _navigationRewritingQueryModelVisitor;
         private QueryModel _queryModel;
 
         private class NavigationJoin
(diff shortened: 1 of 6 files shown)


maumar:
@anpete

src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
```diff
@@ -85,13 +87,27 @@ public NavigationRewritingExpressionVisitor([NotNull] EntityQueryModelVisitor qu
...
+        }
+
         public virtual void Rewrite([NotNull] QueryModel queryModel)
         {
             Check.NotNull(queryModel, nameof(queryModel));
 
             _queryModel = queryModel;
 
-            queryModel.TransformExpressions(Visit);
+            var queryModelVisitor = new NavigationRewritingQueryModelVisitor(Visit, VisitInnerKeySelector);
```
maumar: this seems a bit sketchy, but not sure if there is better (cannonical) way of doing this - two visitors working on one expression tree, sharing some state

maumar:
@anpete  new version is up

anpete:
@maumar OK will try and get to this later today.

src/EntityFramework.Core/Query/ExpressionVisitors/NavigationRewritingExpressionVisitor.cs
```diff
@@ -400,5 +519,62 @@ private ConstantExpression CreateEntityQueryable(IEntityType targetEntityType)
         [UsedImplicitly]
         private static EntityQueryable<TResult> _CreateEntityQueryable<TResult>(IAsyncQueryProvider entityQueryProvider)
             => new EntityQueryable<TResult>(entityQueryProvider);
+
+        private class NavigationRewritingQueryModelVisitor : QueryModelVisitorBase
+        {
+            private Func<Expression, Expression> _transformation;
+
+            public bool IsInsideInnerKeySelector { get; private set; }
```
anpete: private field

anpete:
:shipit:

PR #3433, opened 2015-10-15
Stop generating temporary values for keys that are StoreGenerated.Never.

ajcvickers:
This means that explicit values, including zeros, can always be inserted if value generation is not being used.

Because every entity, even those not yet saved, must have a valid, unique key, this change means that key values must always be set if value generation is not being used. This is true even in cases where the properties are FKs because if there is no navigation to a principal entity, then adding the dependent first will fail since it can't get the key from the parent entity and can't use the parent entity's value generation to create a valid key since the parent entity doesn't have value generation.


AndriySvyryd:
:shipit:

ajcvickers:
Merged.

Issue #3438, opened 2015-10-15
Create default implementation of IDatabaseModelFactory


natemcmaster:
It is maybe possible to provide a base implementation of IMetadataReader so that provider-writers only need to give us SQL queries to make reveng work. Use `INFORMATION_SCHEMA`

Related https://github.com/aspnet/EntityFramework/issues/1667

rowanmiller:
We should look at using EF when we do this

natemcmaster:
Made some discoveries from an attempted implementation in EF. 

Need to have:
 - https://github.com/aspnet/EntityFramework/issues/3945
 - https://github.com/aspnet/EntityFramework/issues/3943

Nice to have:
 - https://github.com/aspnet/EntityFramework/issues/3944
 - https://github.com/aspnet/EntityFramework/issues/3942
 - #621

rowanmiller:
As a workaround to #3943 you should be able to put these things in shadow state and then after you execute the query run some additional logic to copy the values out of the change tracker into the annotations.

divega:
@natemcmaster is the materialization hook you mention in #3943 a need to have in the sense that you cannot resolve this issue (#3438) without it? I would like to understand if it can be worked around.

natemcmaster:
Listing the limitations above was more about minimizing workarounds and driving useful features in EF. They can all be worked around.

@divega #3943 can be a "nice to have" since we don't plan to support multiple rows => one entity (#3945).

> copy the values out of the change tracker

@rowanmiller this was my original approach. Of course this means I *have to* use the change tracker ( see https://github.com/aspnet/EntityFramework/issues/2282#issuecomment-161070921)

rowanmiller:
@natemcmaster we agreed not to use EF for loading metadata, but is the more general scenario covered by this issue (providers just needing to provide SQL) done?

natemcmaster:
This general scenario is not done. We still only have a Sql Server and SQLite implementation with no shared implementation.

PR #3439, opened 2015-10-15
[reveng] Move GetModel method into base implementation

natemcmaster:
So SQL Server and SQLite don't both implement the same IMetadataReader => IModel call

/cc @divega  @lajones


src/EntityFramework.Sqlite.Design/SqliteDesignTimeServices.cs
```diff
@@ -19,7 +19,7 @@ public class SqliteDesignTimeServices
         public virtual void ConfigureDesignTimeServices([NotNull] IServiceCollection serviceCollection)
         {
             serviceCollection
-                .AddSingleton<MetadataModelProvider, SqliteMetadataModelProvider>()
+                .AddSingleton<MetadataModelProvider, RelationalMetadataModelProvider>()
```
divega: I vaguely remember a discussion on whether provider should still need to derive and have their own implementation (or was it their own interfaces?). Things have been in flux and I can't remember if this was a concern or not. @ajcvickers would know.

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -18,7 +18,6 @@ namespace Microsoft.Data.Entity.SqlServer.Design.ReverseEngineering
 {
     public class SqlServerMetadataModelProvider : RelationalMetadataModelProvider
```
divega: It seems interesting to think about the code left in this class in terms of "how do we get rid of it". E.g. it raises a few questions:

1. Should the datetime precision quirk be normalized in the schema reader so that we don't need to do it here?
2. What is different about SQL server treatment of default values? Would it make sense to abstract that into a component that all providers can supply? 
3. Why do we need to special case timestamp here? Should reverse type mapping handle it?

I see this as discussion items (possibly bugs) not blocking this PR.

divega:
:shipit: LGTM but it would be good if @lajones also reviews it.

lajones:
LGTM too. :shipit:

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
natemcmaster: > Should the datetime precision quirk be normalized in the schema reader so that we don't need to do it here?

T-SQL maps 1:1 into `Table.Scale`. INFORMATION_SCHEMA does not. If we switch back, we should just be consistent about where this info is store.

> What is different about SQL server treatment of default values? Would it make sense to abstract that into a component that all providers can supply?

We could abstract into a feature of reveng, but for now it is very SQL Server specific. @lajones should we  file an work item to abstract this?

> Why do we need to special case timestamp here? Should reverse type mapping handle it?

I'll update to move this into the metadata reader. It's trivial.

divega: > T-SQL maps 1:1 into  Table.Scale . INFORMATION_SCHEMA does not

Regardless of whether we do #3438 I still believe it is better to push the compensation down to the provider's schema reader. There is no strong need to have a 1:1 mapping if there is a quirk. You already have a ton of `CASE` in that SQL so it seems to be just a matter of getting the right ones! 

> We could abstract into a feature of reveng, but for now it is very SQL Server specific

Ok, I will wait for the bug. 

> I'll update to move this into the metadata reader. It's trivial.

Why is this not a type mapping concern? Do we fundamentally lack the ability to have database types that are synonyms or something?

natemcmaster: > Why is this not a type mapping concern? Do we fundamentally lack the ability to have database types that are synonyms or something?

No, typemapping works fine. We have alias support. This was just a matter of getting `ValueGeneration` set correctly in the IModel.

src/EntityFramework.Sqlite.Design/SqliteDesignTimeServices.cs
natemcmaster: Should we abstract an interface for `MetadataModelProvider` then?

src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
divega: Got it. That makes sense to me now. 

Now what is confusing to me is that `SqlServerMetadataReader` seems to [handle timestamp and not rowversion](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataReader.cs#L158) :smile: 

``` C#
IsComputed = reader.GetBoolean(12) || dataTypeName == "timestamp" 
```
I don't even understand why this is `||` as opposed to `&&`.

divega: Talked to @natemcmaster in person and that part makes more sense to me now. Thanks!

Issue #3161, opened 2015-09-17
Inheritance :: error when adding Customer and VipCustomer to a db set in one "save changes"


maumar:
Repro:

```
using System.Data.SqlClient;
using Microsoft.Data.Entity;

namespace Repro3161
{
    public class Customer
    {
        public int Id { get; set; }
        public string Name { get; set; }
    }

    public class VipCustomer : Customer
    {
    }

    public class MyContext : DbContext
    {
        public DbSet<Customer> Customers { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "Repro3161", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<VipCustomer>().BaseType<Customer>();
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var c1 = new Customer { Name = "C1", };
                var c2 = new Customer { Name = "C2", };
                var c3 = new VipCustomer { Name = "VC3", };

                ctx.Customers.AddRange(c1, c2, c3);
                ctx.SaveChanges();
            }
        }
    }
}

```

This throws:
The instance of entity type 'Repro3161.VipCustomer' cannot be tracked because another instance of this type with the same key is already being tracked. For new entities consider using an IIdentityGenerator to generate unique key values.

   at Microsoft.Data.Entity.ChangeTracking.Internal.StateManager.StartTracking(InternalEntityEntry entry) in D:\k\EntityFramework\src\EntityFramework.Core\ChangeTracking\Internal\StateManager.cs:line 222
   at Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges) in D:\k\EntityFramework\src\EntityFramework.Core\ChangeTracking\Internal\InternalEntityEntry.cs:line 177
[...]

PR #3441, opened 2015-10-15
Inheritance conventions

### LINKED ISSUE #1704: Code First: Implement inheritance support

AndriySvyryd:
Add conventions that discover base and derived types that are already in the model using the associated CLR type information
Adjust conventions to correctly handle inheritance
Allow setting the base type to null
Move non-shadow property validation to Property

Fixes #1704

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/ValueGenerationManager.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/ValueGenerationManager.cs
@@ -38,7 +38,9 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
                     }
                     else
                     {
-                        var valueGenerator = _valueGeneratorSelector.Select(property, entry.EntityType);
+                        var valueGenerator = _valueGeneratorSelector.Select(property, property.IsKey()
+                            ? property.DeclaringEntityType
+                            : entry.EntityType);
 
                         Debug.Assert(valueGenerator != null);
 
--- a/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
@@ -427,7 +427,7 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
                 }
             }
 
-            var properties = MayGetStoreValue();
+            var properties = FindPropertiesThatMayGetStoreValue();
             if (properties.Any())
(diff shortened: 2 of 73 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/ValueGenerationManager.cs
```diff
@@ -38,7 +38,9 @@ public virtual void Generate(InternalEntityEntry entry)
                     }
                     else
                     {
-                        var valueGenerator = _valueGeneratorSelector.Select(property, entry.EntityType);
+                        var valueGenerator = _valueGeneratorSelector.Select(property, property.IsKey()
+                            ? property.DeclaringEntityType
+                            : entry.EntityType);
```
smitpatel: this part would fix #3161 ?

AndriySvyryd: correct

divega: Probably a silly question, but why not always DeclaringEntityType?

AndriySvyryd: That's a good question and the reason is discriminator properties.

divega: I see... but discriminators and keys are not the only cases of properties that are associated to value generators, are they? Should we explicitly store in the model if the generator is supposed to be associated with the type or the whole hierarchy?

divega: Talked to @AndriySvyryd offline. Apparently right now it is not easy to setup things for value generation outside of keys and discriminators so this assumption should work.

AndriySvyryd:
Ping. This needs to go in for RC

ajcvickers:
@divega @smitpatel Since you have already being looking at this could one of you finish the review? If you can't, then I'll start looking.

smitpatel:
:shipit:

Issue #3112, opened 2015-09-11
Refactor RelationalCommand and its builder


ajcvickers:
From API review:  Remove RelationalParameter and just have RelationalCommand take a RelationalParameterList. 

Also, if it still matters after the above is done, RelationalCommand should not use a "params" parameter.

mikary:
Removed the `RelationalParameterList` and changed the "params" constructor. I think the new factoring should satisfy the intent of this issue
- [ ] API review required

rowanmiller:
@mikary can this be closed now?

mikary:
Changes are done other than API review changes, I think this issue is fine to close at this point.

PR #3442, opened 2015-10-15
Fix #3110 - Move logic from ISqlStatementExecutor into IRelationalCommand

### LINKED ISSUE #3110: Route all statement execution through ISqlStatementExecutor

mikary:
Allow caller to manually control connection Open / Close
Remove CreateDbCommand from IRelationalCommand for #3112

Updates for Migrations in `Migrator` and `RelationalDatabaseCreator`
Updates to Query for `RelationalDataReader` and connection management changes to `RelationalConnection`
@bricelam @anpete


mikary:
If possible, I'd like to get this reviewed soon since it's API surface changing
(ping) @bricelam @anpete

mikary:
(Rebased over Telemetry/Diagnostic changes)

src/EntityFramework.Relational/Migrations/HistoryRepository.cs
```diff
@@ -154,21 +151,15 @@ public virtual IReadOnlyList<HistoryRow> GetAppliedMigrations()
...
             if (await ExistsAsync(cancellationToken))
             {
-                await _connection.OpenAsync(cancellationToken);
-                try
+                var command = _sqlCommandBuilder.Build(GetAppliedMigrationsSql);
+
+                using (var reader = await command.ExecuteReaderAsync(_connection))
                 {
-                    using (var reader = await _executor.ExecuteReaderAsync(_connection, GetAppliedMigrationsSql, cancellationToken))
+                    while (await reader.DbDataReader.ReadAsync(cancellationToken))
```
anpete: This looks strange

src/EntityFramework.Relational/Storage/RelationalDataReader.cs
```diff
@@ -10,18 +10,21 @@ namespace Microsoft.Data.Entity.Storage
 {
     public class RelationalDataReader : IDisposable
```
anpete: I think it might just be a case of renaming this to make the consuming code look less strange.

DataReaderHolder or something like that.

mikary: maybe DataReaderContext? The pattern seems consistent with DbContext in terms of being a scoped disposable with the expectation of cleanup on related items.

divega: Keep in mind a pattern such as _[MethodName]Result_ for naming this.

src/EntityFramework.Relational/Migrations/HistoryRepository.cs
mikary: With the proposed name change and a rename on the property, would this look more reasonable?

```cs
using(var readerContext = await command.ExecuteReaderAsync(_connection, cancellationToken))
{
    while(readerContext.Reader.ReadAsync(cancellationToken)
    {
        // ...
    }
}

```

src/EntityFramework.Relational/Storage/RelationalDataReader.cs
divega: BTW, `ExecuteReaderResult` that has a Reader property seems strange too, but maybe it is the method that needs a rename since it is currently called `ExecuteReader` after the object that it is supposed to return. We could break with ADO.NET (I don't think aligning has ever been a goal) and call it something else, e.g. something like `ExecuteQuery` (symmetric to `ExecuteNonQuery`) and then the return type could be something like `QueryResult` which would be an IDisposable and contain a Reader property.

anpete: What about move a few of the common methods like Read/ReadAsync up to the RelationalReader but keep the DbDataReader property for advanced consumption, such as with query?

divega: Not sure it is worth doing. It seems that we would need to add field access to it as well for any meaningful consumption to work without dropping down to the DbDataReader.

anpete: Agree but we only do the meaningful consumption inside ValueBuffer factories and so that part is essentially opaque.

mikary: That's also true for `AffectedCountModificationCommandBatch`, but `HistoryRepository` is using `GetString` off of the reader itself. Not really a huge deal, but there's no ValueBufferFactory involved.

anpete: Yeah, agree it would be fine to just leave Migrations as is.

divega: I had a quick chat with @anpete and we ended up coming back (full circle) to believe that this class it is alright the way it is. `RelationalDataReader` as a name is consistent with other relational wrappers we have created which are also "partial decorators" that layer some functionality over the underlying ADO.NET object they hold. On consumption you would still get the underlying `DbDataReader` and call `Read[Async]()` on it.

anpete: I was still in favor of lifting the ReadX methods in our discussion :smile:

But I don't feel strongly about it. Will leave it up to @mikary

anpete:
:shipit:

src/EntityFramework.Relational/Storage/RelationalDataReader.cs
mikary: I think it's pretty good where it is now and we're down to the point of grooming the API surface. After this PR we should be able to API review all of the commands related code we skipped over in previous reviews.

cc @rowanmiller

PR #3449, opened 2015-10-16
Use metadata features in reveng

natemcmaster:
Last set of changes to complete #2956

As we discussed in design meeting, we want to make it easier for providers to simply say, "here is my 'UseSqlServer' equivalent method name."

Changes:
- Leverage the already existing "DefaultSchema" annotation in relational
- Remove need to override model configuration factories.
- Add metadata extensions to relational design.
- Apply consistent annotations patterns to relational design.
- Fix a bug in ValueGeneratedNever with SqlServer.
- Add a "DatabaseName" annotation to relational

Last step: rename everything. ( see #3335 )


src/EntityFramework.MicrosoftSqlServer.Design/ReverseEngineering/SqlServerMetadataModelProvider.cs
```diff
@@ -46,6 +49,31 @@ protected override PropertyBuilder AddColumn([NotNull] EntityTypeBuilder builder
             return propertyBuilder;
         }
 
+        protected override EntityTypeBuilder AddPrimaryKey([NotNull] EntityTypeBuilder builder, [NotNull] Table table)
```
lajones: It feels like we're repeating the code from `KeyConvention` here - with the risk that the two pieces of code get out of sync. We used to have a method in `KeyConvention` which defined when it would automatically decide to add `ValueGeneratedOnAdd()` and here we just consulted that method. I'd like to go back to doing that.

natemcmaster: Agreed. This would be cleaner. Will do.

lajones: Sounds good. :shipit:

natemcmaster: FYI - I'm adding as a TODO due to upcoming changes in conventions proposed today by Andriy.

PR #3451, opened 2015-10-16
Fix #2909 - Make RelationalConnection check connection state on open

### LINKED ISSUE #2909: DbContext configured with an external DbConnection raises an exception when the DbConnection is already opened.
The issue: the DbContext tries to open the connection even if it is already opened. 
This happens when we configure a DbContext with a DbConnection in a *Closed* state and then open the DbConnection outside of EntityFramework. 
When the DbContext needs to use the DbConnection, it will call Open (or OpenAsync) on the abstract class *RelationnalConnection* without verifiying if the DbConnection is owned or not.
This results on an InvalidOperationException: **The connection was not closed. The connection's current state is open.**

Sample code:

```
string connectionString = "Put_your_connection_string";

using( DbConnection c = new SqlConnection( connectionString ) )
{
    IServiceProvider sp = TestHelper.Instance.CreateServiceProvider( config =>
    {
        config.AddEntityFramework().AddSqlServer().AddDbContext<SampleDbContext>( o => o.UseSqlServer(c));
    });
    // PASS OK.
    sp.GetRequiredService<SampleDbContext>().Database.ExecuteSqlCommand( "select 1 from dbo.myTable" );
    Assert.True( c.State == System.Data.ConnectionState.Closed );

    c.Open();
    // FAIL KO.
    Assert.Throws<InvalidOperationException>(
        () => sp.GetRequiredService<SampleDbContext>().Database.ExecuteSqlCommand( "select 1 from CK.tUser" ) );
}
```

mikary:
Makes sure EF will only open a connection if it is closed, and only close a connection that was opened by EF.

### DIFF
--- a/src/EntityFramework.Relational/Storage/RelationalConnection.cs
+++ b/src/EntityFramework.Relational/Storage/RelationalConnection.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Data.Entity.Storage
         private readonly LazyRef<DbConnection> _connection;
         private readonly bool _connectionOwned;
         private int _openedCount;
+        private bool _openedInternally;
         private int? _commandTimeout;
         private readonly ILogger _logger;
 #if NET45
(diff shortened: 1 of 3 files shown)


ajcvickers:
:shipit:

Issue #3432, opened 2015-10-15
FromSql throws ArgumentNullException when passing a null parameter


opiants:
Suppose we have a stored procedure

	CREATE PROCEDURE GetPersons @departmentid INT AS SELECT * FROM People WHERE DepartmentId = COALESCE(@departmentid, DepartmentID)

Calling that stored procedure using the code below causes an exception.

	int? departmentId = null;
	db.Set<Person>().FromSql(@"GetPersons @departmentid = {0}", departmentId)

Full stack trace:

	ArgumentNullException: Value cannot be null. Parameter name: value

    Microsoft.Data.Entity.Utilities.Check.NotNull[T](T value, String parameterName)
    Microsoft.Data.Entity.Query.CommandParameter..ctor(String name, Object value, RelationalTypeMapping typeMapping)
    Microsoft.Data.Entity.Query.Sql.DefaultQuerySqlGenerator.VisitRawSqlDerivedTable(RawSqlDerivedTableExpression rawSqlDerivedTableExpression)
    Microsoft.Data.Entity.Query.Expressions.RawSqlDerivedTableExpression.Accept(ExpressionVisitor visitor)
    System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
    Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
    Microsoft.Data.Entity.Query.Sql.DefaultQuerySqlGenerator.<VisitJoin>b__30_0(Expression e)
    Microsoft.Data.Entity.Query.Sql.DefaultQuerySqlGenerator.VisitJoin[T](IReadOnlyList`1 items, Action`1 itemAction, Action`1 joinAction)
    Microsoft.Data.Entity.Query.Sql.DefaultQuerySqlGenerator.VisitJoin(IReadOnlyList`1 expressions, Action`1 joinAction)
    Microsoft.Data.Entity.Query.Sql.DefaultQuerySqlGenerator.VisitSelect(SelectExpression selectExpression)
    Microsoft.Data.Entity.Query.Expressions.SelectExpression.Accept(ExpressionVisitor visitor)
    System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
    Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
    Microsoft.Data.Entity.Query.Sql.DefaultQuerySqlGenerator.GenerateSql(IDictionary`2 parameterValues)
[...]

PR #3458, opened 2015-10-16
Add test coverage for FromSql with null parameter value #3432

mikary:
Issue appears to be resolved in beta8, adding a test to prevent regression


ajcvickers:
:shipit:

PR #3462, opened 2015-10-17
[SMALL] RevEng: Generate POCO classes as partial

### LINKED ISSUE #3428: RevEng: Generate partial classes
According to [this StackOverflow question](http://stackoverflow.com/q/33045923/475031), we don't generate `partial` classes.

In the past, we've had the rule to always generate partial classes when scaffolding code.

lajones:
Fixes #3428. (Fix looks large but it's mostly just the expected results files.)

Also removed an unused dependency on `ModeUtilities` in `ReverseEngineeringGenerator`.


smitpatel:
:shipit:

lajones:
Checked in with commit 65bf6c0.

PR #3463, opened 2015-10-17
Add more sugar API for discriminators

AndriySvyryd:
Part of #1704

### DIFF
--- a/src/EntityFramework.Relational/Metadata/Internal/RelationalEntityTypeBuilderAnnotations.cs
+++ b/src/EntityFramework.Relational/Metadata/Internal/RelationalEntityTypeBuilderAnnotations.cs
@@ -50,11 +50,9 @@ namespace Microsoft.Data.Entity.Metadata.Internal
 
             return true;
         }
-
-        [CanBeNull]
+        
         public virtual DiscriminatorBuilder HasDiscriminator() => DiscriminatorBuilder(null, null);
-
-        [CanBeNull]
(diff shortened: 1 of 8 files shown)


src/EntityFramework.Relational/Metadata/Internal/RelationalEntityTypeBuilderAnnotations.cs
```diff
@@ -62,12 +62,19 @@ public virtual DiscriminatorBuilder HasDiscriminator([CanBeNull] Type discrimina
...
 
-            return DiscriminatorBuilder(null, discriminatorType);
+            return DiscriminatorProperty != null
+                   && DiscriminatorProperty.ClrType == discriminatorType
+                ? DiscriminatorBuilder(null, null)
+                : DiscriminatorBuilder(null, discriminatorType);
         }
 
-        [CanBeNull]
+        [NotNull]
```
ajcvickers: We decided to not use these attributes for return values. We should be consistent about that, or we should revisit that decision.

ajcvickers:
:shipit:

src/EntityFramework.Relational/Metadata/Internal/RelationalEntityTypeBuilderAnnotations.cs
AndriySvyryd: Removed

Issue #3317, opened 2015-10-03
MSSQL - incorrect sql - Where(x => subQuery.Any(y => x.xxx == y.yyy)


LonDat:
Hi, I have code:
```C#
var qSrvMail = EntityContext.Current.Query<SrvMailAddress>(); // DbSet<SrvMailAddress>
var qSubMail = EntityContext.Current.Query<KontaktMail>(); // DbSet<KontaktMail>

// PROBLEM => !qSrvMail.Any(m => m.Mail == k.Mail)
string[] mails = qSubMail.Where(k => !qSrvMail.Any(m => m.Mail == k.Mail)).Select(k => k.Mail).Distinct().ToArray();
```

It's generate this SQL:
```SQL
SELECT DISTINCT [k].[Mail]
FROM [KontaktMail] AS [k]
WHERE CASE
    WHEN (
        SELECT CASE
            WHEN EXISTS (
                SELECT 1
                FROM [SrvMailAddress] AS [m]
                WHERE [m].[Mail] = [k].[Mail])
            THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT)
        END
    ) /* = 1 <<< SOMETHING LIKE THIS MISSING!!! */
    THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT)
END = 0
```
This SQL statement is broken (`An expression of non-boolean type specified in a context where a condition is expected, near 'THEN'.`)

My environment:
MSSQL 2014 Developer Ed.
EF 7 - dev branch (today synced - 2015/10/3)

EF7 ROCK (is't not ironic :D)
Thx for check, Milan

aconstant:
As stated by @LonDat, the SQL generated is missing a ```= 1``` condition, when using the Any() method.

A workaround is to call ```Count(expression) > 0``` instead of ```Any(expression)```, but that should be less optimized, since it results in a SQL call to : ```Count(*) > 0```

trailheadtechnology:
So is there a fix in the main branch for this?

Issue #3235, opened 2015-09-25
Query :: include with inheritance and join doesn't work in some cases


maumar:
Repro:

```
    public class Customer
    {
        public int Id { get; set; }
        public string Name { get; set; }

        public List<Order> Orders { get; set; }
    }

    public class VipCustomer : Customer
    {
        public Foo FooProp { get; set; }
    }

    public class Foo
    {
        public int Id { get; set; }
        public string Name { get; set; }
    }

    public class Order
    {
        public int Id { get; set; }
        public int CustomerId { get; set; }
        public string Name { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Customer> Customers { get; set; }
        public DbSet<Order> Orders { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "Repro3235", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Customer>().HasMany(e => e.Orders).WithOne().ForeignKey(e => e.CustomerId);
            modelBuilder.Entity<VipCustomer>().BaseType<Customer>();
            modelBuilder.Entity<VipCustomer>().HasOne(e => e.FooProp).WithOne().Required(false);
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var o11 = new Order { Name = "O11", };// OrderDetails = new List<OrderDetail> { od111, od112 } };
                var o12 = new Order { Name = "O13" };
                var o21 = new Order { Name = "O21" };
                var o22 = new Order { Name = "O22", };// OrderDetails = new List<OrderDetail> { od221, od222, od223 } };
                var o23 = new Order { Name = "O23" };
                var o3 = new Order { Name = "O3" };


                var c1 = new Customer { Name = "C1", Orders = new List<Order> { o11, o12 } };
                var c2 = new Customer { Name = "C2", Orders = new List<Order> { o21, o22, o23 } };

                ctx.Customers.AddRange(c1, c2);
                ctx.Orders.AddRange(o11, o12, o21, o22, o23);
                ctx.SaveChanges();

[...]

PR #3469, opened 2015-10-17
Need a discuss and review, couple fixes and changes

multiarc:
#3235 fix - more than a fix, actually we could allow to use types as base types without strict declaration of base on model for example if we want separated tables with the same base entity and  ignore keys/properties from base type in child.

#3317 fix - select statement in SQL couldn't result in a bool, so just check the expression if it's select on binary expression and add " = 1"
Also MSSQL tests does have some missed parts on SQL generation if we have different FK/PK names and result SQL query was contain wrong table reference values because of small mistake. That was fixed in this commit.

### DIFF
--- a/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
@@ -378,7 +378,7 @@ namespace Microsoft.Data.Entity.Query
                     var navigation = Model.FindEntityType(navigationType)?.FindNavigation(propertyInfo.Name);
                     if (navigation == null)
                     {
-                        throw new Exception($"Cannot find navigation {propertyInfo}.");
+                        return null;
                     }
 
                     boundChainedNavigations.Add(navigation);
--- a/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
@@ -300,7 +300,7 @@ namespace Microsoft.Data.Entity.Query
 
                                         return BindChainedNavigations(
                                             navigations,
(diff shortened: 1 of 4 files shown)


rowanmiller:
@anpete can you take a general look at these changes.

anpete:
Assigning to @maumar as he has already looked at both of these issues.

multiarc:
Regarding EntityQueryModelVisitor: 
I was trying to add type reference from expression if nothing found in model. This was added to allow correct navigation reference resolving when base type wasn't registered as entity and no relations was registered in model for base class.

```CSharp
public abstract class OptionValue
{
        public int Id { get; set; }
        public string Value { get; set; }
}

public abstract class DynamicData<TOptionValue>
        where TOptionValue: OptionValue
{
        public int Id { get; set; }
        public ICollection<TOptionValue> OptionValues { get; set; }
}

public class OrderOptionValue: OptionValue
{
        //...
}

public class Order: DynamicData<OrderOptionValue>
{
        //...
}
```

So when you try to resolve OptionValues collection in Order it will try to resolve OptionValue base abstract Type, but technically it should get OrderOptionValue with correct reference to right table.

The main point of all this is we can't change/ignore relations easily using Base declaration. 
We could have any kind of abstraction for base entity and don't mind how this declared or pointed to model so we declare just concrete classes and never think about base.

anpete:
@multiarc Our main priority right now is to fix bugs and so that was the focus of this change. Feel free to open a new issue if you have ideas for how we can improve mapping around inheritance and nav. props.

divega:
@anpete  :+1: 

@multiarc Thanks for your contribution. If I understand the scenario you described above correctly I agree it would be useful to make it work, so if it still not working a new issue (so that we can triage it and prioritize it) and a new PR would help.

PR #3478, opened 2015-10-19
Incremental steps to netnative support

natemcmaster:
Changes necessary to make the very basic UWP app work in Release mode (after .NET Native mangles it).

Will still require a patch to Relinq (https://github.com/re-motion/Relinq/pull/6)

Makes significant progress on #623

### DIFF
--- a/src/EntityFramework.Core/Utilities/Imply.cs
+++ b/src/EntityFramework.Core/Utilities/Imply.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#pragma warning disable 0169 
+
+using Microsoft.Data.Entity.Storage;
+using System;
+using Microsoft.Data.Entity.ChangeTracking.Internal;
+using Microsoft.Data.Entity.Metadata.Internal;
+
+namespace Microsoft.Data.Entity.Utilities
+{
+    // This code exists only to trick the ILC compliation to include metadata 
+    // about combinations of entity types and our internal types.
+    // This is the jumping off point for a reasoning about what generic types
+    // may exist at runtime.
+    // https://github.com/aspnet/EntityFramework/issues/3477
+    internal class ImpliedEntityType<TEntity>
+        where TEntity : class
+    {
+        InternalImplyValuesAndTypes<TEntity> EntityValueProp;
+
+        InternalImpliesTypes<TEntity> EntityProp;
+        InternalImpliesTypes<char> CharProp;
+        InternalImpliesTypes<short> Int16Prop;
+        InternalImpliesTypes<ushort> UInt16Prop;
+        InternalImpliesTypes<int> Int32Prop;
+        InternalImpliesTypes<uint> UInt32Prop;
+        InternalImpliesTypes<long> Int64Prop;
+        InternalImpliesTypes<long> UInt64Prop;
+        InternalImpliesTypes<double> DoubleProp;
+        InternalImpliesTypes<decimal> DecimalProp;
+        InternalImpliesTypes<float> FloatProp;
+        InternalImpliesTypes<string> StringProp;
+        InternalImpliesTypes<bool> BoolProp;
+        InternalImpliesTypes<byte> ByteProp;
+        InternalImpliesTypes<Guid> GuidProp;
+        InternalImpliesTypes<TimeSpan> TimeSpanProp;
+        InternalImpliesTypes<DateTime> DateTimeProp;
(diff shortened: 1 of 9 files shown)


ajcvickers:
How does this effect perf? Does it make the benchmarks we have significantly slower?

ajcvickers:
Also, have we considered cross-compiling so we only do this for UWP apps? Since perf is a priority it seems counter-productive to on the one hand be trying to improve perf, while with the other hand removing code that was specifically added to help with perf.

natemcmaster:
I suspect perf will be worse. I can actually test the numbers to see how much.

We are not currently set up to cross compile to UWP. @bricelam would know better how we could set that up.

Alternatively, we could catch the runtime exception thrown only on .NET Native and then invoke this slower path which preserves the original path for all other cases.

ajcvickers:
I think we should discuss after seeing how much it affects our current benchmarks.

src/EntityFramework.Core/Utilities/Imply.cs
```diff
@@ -0,0 +1,79 @@
...
+    // This is the jumping off point for a reasoning about what generic types
+    // may exist at runtime.
+    // https://github.com/aspnet/EntityFramework/issues/3477
+    internal class ImpliedEntityType<TEntity>
+        where TEntity : class
+    {
+        InternalImplyValuesAndTypes<TEntity> EntityValueProp;
+
+        InternalImpliesTypes<TEntity> EntityProp;
+        InternalImpliesTypes<char> CharProp;
```
divega: Is there a way we can do this in providers since primitive type support is tied to providers?

natemcmaster:
To my surprise, this actually improves perf:

```c#
        [Benchmark(Iterations =10)]
        public void ConstructGetterSource(IMetricCollector collector)
        {
            var entityType = new EntityType(typeof(Poco), new Model());
            var prop = entityType.AddProperty("Id", typeof(int));
            for (var i = 0; i < 100000; i++)
            {
                var source = new ClrPropertyGetterSource();
                var setter = new ClrPropertySetterSource();
                collector.StartCollection();
                source.GetAccessor(prop);
                setter.GetAccessor(prop);
                collector.StopCollection();
            }
        }
```

Code | TimeElapsedAverage | TimeElapsed95
------|------|------
Original | 2266 | 2383
With changes in this PR | 1840 | 1889

ajcvickers:
@natemcmaster That looks good--what is the impact on our existing benchmarks?

natemcmaster:
@ajcvickers The changes to existing perf benchmarks was hardly significant. I had to run 100,000 iterations to get numbers I could believe were significantly different.

rowanmiller:
Just checking that you got those numbers on a release build and not a debug one (since that is the perf we care about)

natemcmaster:
Good point. Those were Debug. I'll re-run on release mode.

ajcvickers:
@natemcmaster I misunderstood the code when I looked at it last night. I see now that it doesn't change the cached delegate path, and hence I'm not surprised that the overall benchmark numbers are okay.

I can say :shipit: for those code changes. The "implies" stuff and how it is supposed to work is a mystery to me, so somebody who knows about that should look.

natemcmaster:
@ajcvickers Good to know. 

Everyone else, there is gonna be a whole lot more awful "implies" stuff, so brace yourselves. This stuff exists only to trick ILC (the .NET Native compilation step) to keep metadata about generic methods and types we use at run time.

src/EntityFramework.Core/Utilities/Imply.cs
natemcmaster: Perhaps a better method is to expose a utility class publicly and let providers pass in their primitive types.

ajcvickers:
@natemcmaster :shipit:

src/EntityFramework.Core/Utilities/Imply.cs
divega: I like that. Also, after discussing it in person today I kept thinking and here are some of my thoughts:

- The goal is to make .NET Native support as transparent as possible to application developers (as opposed to provider writers)
- I would like to see providers do whatever they need to do so that their valid primitive types are supported in .NET native in code that is near their type mapper
- If we do this for some well known set of types in the core it actually increases the chances that providers are just going to miss that they need to do it for all the types they support in their type mapper until they have some customer trying to use a particular type that is not part of the subset on .NET Native.

natemcmaster: @divega I added some of the related actionable items of these in https://github.com/aspnet/EntityFramework/issues/623#issuecomment-149737147.

natemcmaster:
Merged with https://github.com/aspnet/EntityFramework/commit/c35ce4d3632331d14ff8a3ab8ac59c54a660a535

PR #3479, opened 2015-10-19
Fix #3422 - React to Telemetry renames

### LINKED ISSUE #3422: Build warnings from TelemetrySource in tests
Presumably after 2887b8b95

mikary:
Reacting to change in aspnet/EventNotification#29
Follows change in aspnet/Hosting#423

cc @anpete @divega @JunTaoLuo


divega:
:shipt: (when it is time)

cc @rynowak @avanderhoorn

rynowak:
> (when it is time)

**The time has come** - we've notified all of our partners that this is happening. I'm not aware of anything that we need to wait for.

Issue #2495, opened 2015-06-29
Create non-internal state entry interface for consuming by SaveChanges


ajcvickers:
When implementing SaveChanges, providers currently consume InternalEntityEntry objects. We should inside consider creating a simplified interface that is not internal and exposes only those things needed by providers to implement SaveChanges.

anpete:
Are we OK with breaking this interface in the future?

ajcvickers:
@anpete Yes. It falls into the category of being an interface that will almost certainly only be implemented for testing, and even that should be rare. @divega and @rowanmiller both agreed it would be okay to break it in the future.

anpete:
:shipit:

PR #3481, opened 2015-10-20
Create non-internal state entry interface for consuming by SaveChanges

ajcvickers:
Issue #2495.

The new interface is IUpdateEntry. An interface has been extracted from EntityKey and called IKeyValue, and this has also been made public. Also, a ToEntityEntry method allows update exceptions to provide access back to the original public EntityEntry objects.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/ArraySidecar.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/ArraySidecar.cs
@@ -22,7 +22,11 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
 
         public override bool CanStoreValue(IPropertyBase property) => Index(property) != -1;
 
-        protected override object ReadValue(IPropertyBase property) => _values[IndexChecked(property)];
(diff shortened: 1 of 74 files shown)


ajcvickers:
@ErikEJ @roji Your providers may need to react to this depending on how much you override in the update pipeline.

src/EntityFramework.Core/ChangeTracking/Internal/CompositeEntityKey.cs
```diff
@@ -10,11 +10,11 @@
 namespace Microsoft.Data.Entity.ChangeTracking.Internal
 {
     [DebuggerDisplay("{DebuggerDisplay,nq}")]
-    public class CompositeEntityKey : EntityKey
+    public class CompositeKeyValue : KeyValue
     {
```
smitpatel: file name changes. also for other classes.

ajcvickers: Done

PR #3482, opened 2015-10-20
Query: Many efficiency improvements to the runtime execution model.

anpete:
- Significant reduction in average compiled query code size
- Replace of QueryResultScopes with TransparentIdentifier struct
- Transparent identifiers only introduced when needed
- Exception interceptor no longer takes a Func
- Simplifications to query compiler
- No longer re-using parameter expressions. cc @maumar


test/EntityFramework.MicrosoftSqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
```diff
@@ -117,54 +104,48 @@ var customers
...
+                var3.Region = (string) object prm2.get_Item(10)
+                var3
                     }
                     , 
                     allowNullResult: False
                 )
             )
             , 
-            querySource: from Customer c in value(EntityQueryable`1[FunctionalTests.TestModels.Northwind.Customer]), 
+    entityAccessor: Unhandled expression type: Default
```
maumar: that's a problem, I will fix this in some later checkin, the expression printer should be handling all the Expression nodes that appear in the query

maumar:
we used to compensate for the ParameterExpression.VisitLambda re-use in the ExpressionPrinter, that code can be now removed

anpete:
@maumar Yep, thought you would like the parameters :smile:

maumar:
:shipit:

Issue #3443, opened 2015-10-16
Decide which CLR properties would not be added to the model


smitpatel:
If entity's CLR type is provided - We discover properties which can be mapped and add them to model via convention. We throw exception if a property exists in CLR type but not added to model as property/navigation or not ignored or target entity type for navigation not ignored.
For which kind of properties we should not throw such exception?
Few examples:
- Read only
- abstract
- Interface type

divega:
@smitpatel could you put together a list of variations and examples, including the current behavior in EF7 and the proposed outcome so that we can quickly go over it together?

We should also consider if we are able to find the underlying field for the property (my understanding is that we do this by convention or using an attribute @ajcvickers mentioned there is an attribute you can use on the property to indicate the name of the field, but I couldn't find it).

ajcvickers:
@divega I looked in the code and its actually not an attribute but rather an annotation on the property called "BackingField".

smitpatel:
implemented via 3c364f30061f64f419e45e1b8ed90d02af1ce479

PR #3487, opened 2015-10-20
Block CLR types which are not mapped by TypeMapper or cannot be added…

smitpatel:
… as navigation

Implements #3443 
As decided in the meeting,
Candidate properties are the properties which have getter & are not static.
Interfaces are not valid navigations.
If any candidate property is not added to model, exception will be thrown.
For scalar properties, CLR type must be mapped by TypeMapper regardless of the column type.


smitpatel:
@ajcvickers - Is this signed off by you?

ajcvickers:
:shipit:

PR #3488, opened 2015-10-20
Fix Commands.FunctionalTest failures in VS - correct diagnostic assembly name

mikary:
Commands functional tests are failing in VS after 4de4695a931ad4943ba76ebb62ff26bbf7ba35d7, updated Assembly name appears to be incorrect.

cc @divega


natemcmaster:
:shipit:

PR #3489, opened 2015-10-20
Rename M.D.E.Relational.Design namespace to M.D.E.Scaffolding.

### LINKED ISSUE #2956: RevEng: minimize "public" API surface

natemcmaster:
Responding to API review changes.
Moving code gen to .Internal namespace

Close #2956.

Related TODO https://github.com/aspnet/EntityFramework/issues/3490


natemcmaster:
cc @rowanmiller  and @lajones

ErikEJ:
No file renames?

lajones:
Looks good :shipit:

natemcmaster:
@ErikEJ Only where class names changed. This is mostly a namespace update. The assembly names are also the same.

ErikEJ:
Once this is in, will it be a good time to re-enable the sqlce design provider?

natemcmaster:
@ErikEJ yes. We have been hard at working polishing the provider-api for reverse engineering. We'd love feedback on it.

natemcmaster:
:watch: api review. Class renames to be included as well.

ErikEJ:
@natemcmaster I wonder if it could be possible to call the code generation somehow from a VS xtension - that would be a natural next step for me, as I already have similar UI for  EF4 & EF6 ?

natemcmaster:
@erikej Your UI would need to use similar startup logic found in the commands project to ensure services are configured correctly. And btw, most of the actually code gen code will move into Commands soon. See #3490

rowanmiller:
@ErikEJ we are planning to build an extension and have a pre-release available on VS Gallery as close to RC1 as possible (hopefully at the same time it is released).

natemcmaster:
Updated with class renames.

### Providers Beware...

To keep this PR smaller, the following will be added in a followup PR.

- [ ] Remove IMethodNameProvider and use annotation instead. Throw when missing.
- [ ] Make TableSelectionSet immutable
- [ ] Make EntityType errors annotations on model, rather than including empty entity types in the scaffolding model
- [ ] Discover/respsect IsClustered indexes on SqlServer
- [ ] Change ColumnModel to use EF's value generation heuristics rather than the current, ambiguously-named flags
- [ ] Possibly remove ExplicitValueGenerationNever annotation

lajones:
:shipit:

Issue #3180, opened 2015-09-21
A column has been specified more than once in the order by list. Columns in the order by list must be unique.


Venomed:
I'm running the following query to test out relation configuration in context and get the above error:

```csharp
context.Snapshots.Where(x => x.CorrelationId == correlationId)
        .Include(x => x.OrganisationSnapshot)
            .ThenInclude(x => x.Organisation)
        .Include(x => x.OrganisationSnapshot)
            .ThenInclude(x => x.OrganisationSnapshotSubscriptions)
                .ThenInclude(x => x.Subscription)
                    .ThenInclude(x => x.Application)
        .Include(x => x.UserSnapshots)
            .ThenInclude(x => x.User)
        .Include(x => x.UserSnapshots)
            .ThenInclude(x => x.UserSnapshotSubscriptions)
                .ThenInclude(x => x.Subscription)
                    .ThenInclude(x => x.Application)
        .Take(10);
```

if I break it down to:

```csharp
context.Snapshots.Where(x => x.CorrelationId == correlationId)
        .Include(x => x.OrganisationSnapshot)
            .ThenInclude(x => x.Organisation)
        .Include(x => x.OrganisationSnapshot)
            .ThenInclude(x => x.OrganisationSnapshotSubscriptions)
                .ThenInclude(x => x.Subscription)
                    .ThenInclude(x => x.Application)
        .Take(10);
```

or 

```csharp
context.Snapshots.Where(x => x.CorrelationId == correlationId)
        .Include(x => x.OrganisationSnapshot)
            .ThenInclude(x => x.Organisation)
        .Include(x => x.OrganisationSnapshot)
            .ThenInclude(x => x.OrganisationSnapshotSubscriptions)
                .ThenInclude(x => x.Subscription)
                    .ThenInclude(x => x.Application)
        .Take(10);
```
then it works fine.

Profiler shows that SQL is indeed messed up:

```sql
exec sp_executesql N'SELECT TOP(10) [x].[Id], [other column names snipped for brevity]
FROM [Snapshots] AS [x]
LEFT JOIN [OrganisationSnapshots] AS [o] ON [o].[Id] = [x].[Id]
LEFT JOIN [Organisations] AS [o0] ON [o].[OrganisationId] = [o0].[Id]
LEFT JOIN [OrganisationSnapshots] AS [o1] ON [o1].[Id] = [x].[Id]
WHERE [x].[CorrelationId] = @__correlationId_0
[...]

PR #3493, opened 2015-10-20
Fix to #3180 - A column has been specified more than once in the order by list. Columns in the order by list must be unique.

maumar:
Problem happens for specific combination of nested include statements. When we include collection, we inject ORDER by clauses. We choose them based on the QuerySource of the tables in the select expression. If select expression has multiple tables, added by previous include reference, both tables have the same QuerySource. This makes the choice of order by table ambiguous and may lead to us ordering by the same column twice.

Fix is to add additional metadata to TableExpressionBase that represents the type of entity it is associated with.

The fix doesn't remove the issue entirely. It can still appear in some cases with self reference includes (where Tables have the same type), or if Includes are done on a subquery (which doesn't have IEntityType associated with it)


maumar:
@anpete

anpete:
@maumar Close this?

PR #3494, opened 2015-10-20
Fix #2891 - Promote transaction API to core

### LINKED ISSUE #2891: Promote transaction API to core (& no-op implementation for in-memory)
Spun off from #2866 

The primary scenario is testing relational data access code against in-memory provider

mikary:
Adding concept of transactions to `DatabaseFacade` in core

### DIFF
--- a/src/EntityFramework.Core/Storage/IDbContextTransactionManager.cs
+++ b/src/EntityFramework.Core/Storage/IDbContextTransactionManager.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public interface IDbContextTransactionManager
+    {
+        IDbContextTransaction BeginTransaction();
+
+        Task<IDbContextTransaction> BeginTransactionAsync(CancellationToken cancellationToken = default(CancellationToken));
+
+        void CommitTransaction();
+
+        void RollbackTransaction();
+    }
+}
--- a/src/EntityFramework.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
+++ b/src/EntityFramework.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
@@ -12,9 +12,13 @@ namespace Microsoft.Data.Entity
 {
     public static class InMemoryDbContextOptionsExtensions
     {
-        public static void UseInMemoryDatabase([NotNull] this DbContextOptionsBuilder optionsBuilder)
-            => ((IDbContextOptionsBuilderInfrastructure)Check.NotNull(optionsBuilder, nameof(optionsBuilder)))
-                .AddOrUpdateExtension(
-                    new InMemoryOptionsExtension());
+        public static InMemoryDbContextOptionsBuilder UseInMemoryDatabase([NotNull] this DbContextOptionsBuilder optionsBuilder)
+        {
+            Check.NotNull(optionsBuilder, nameof(optionsBuilder));
+
+            ((IDbContextOptionsBuilderInfrastructure)optionsBuilder).AddOrUpdateExtension(new InMemoryOptionsExtension());
+
+            return new InMemoryDbContextOptionsBuilder(optionsBuilder);
+        }
     }
 }
--- a/src/EntityFramework.InMemory/Infrastructure/InMemoryDbContextOptionsBuilder.cs
+++ b/src/EntityFramework.InMemory/Infrastructure/InMemoryDbContextOptionsBuilder.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Infrastructure.Internal;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Infrastructure
+{
+    public class InMemoryDbContextOptionsBuilder
+    {
+        public InMemoryDbContextOptionsBuilder([NotNull] DbContextOptionsBuilder optionsBuilder)
+        {
+            Check.NotNull(optionsBuilder, nameof(optionsBuilder));
+
+            OptionsBuilder = optionsBuilder;
+        }
+
+        protected virtual DbContextOptionsBuilder OptionsBuilder { get; }
+
+        public virtual InMemoryDbContextOptionsBuilder IgnoreTransactions()
+            => SetOption(e => e.IgnoreTransactions = true);
+
+        protected virtual InMemoryDbContextOptionsBuilder SetOption([NotNull] Action<InMemoryOptionsExtension> setAction)
+        {
+            Check.NotNull(setAction, nameof(setAction));
+
+            var extension = new InMemoryOptionsExtension(OptionsBuilder.Options.GetExtension<InMemoryOptionsExtension>());
+
+            setAction(extension);
+
+            ((IDbContextOptionsBuilderInfrastructure)OptionsBuilder).AddOrUpdateExtension(extension);
+
+            return this;
+        }
+    }
+}
--- a/src/EntityFramework.InMemory/Infrastructure/Internal/InMemoryOptionsExtension.cs
+++ b/src/EntityFramework.InMemory/Infrastructure/Internal/InMemoryOptionsExtension.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 
@@ -8,6 +9,23 @@ namespace Microsoft.Data.Entity.Infrastructure.Internal
 {
     public class InMemoryOptionsExtension : IDbContextOptionsExtension
     {
+        private bool _ignoreTransactions;
+
+        public InMemoryOptionsExtension()
+        {
+        }
+
+        public InMemoryOptionsExtension([NotNull] InMemoryOptionsExtension copyFrom)
+        {
+            _ignoreTransactions = copyFrom._ignoreTransactions;
+        }
+
+        public virtual bool IgnoreTransactions
+        {
+            get { return _ignoreTransactions; }
+            set { _ignoreTransactions = value; }
+        }
+
         public virtual void ApplyServices(EntityFrameworkServicesBuilder builder)
         {
             Check.NotNull(builder, nameof(builder));
--- a/src/EntityFramework.InMemory/Properties/InMemoryStrings.resx
+++ b/src/EntityFramework.InMemory/Properties/InMemoryStrings.resx
@@ -1,7 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-
 <root>
-    <!-- 
+  <!-- 
     Microsoft ResX Schema 
     
     Version 2.0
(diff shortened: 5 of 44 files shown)


src/EntityFramework.Core/Storage/IConnection.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public interface IConnection
```
ajcvickers: I think it would be worth discussing with @divega and @anpete whether or not we should introduce a non-relational connection concept just because we have a non-relational transaction concept. We had this in the past and I'm pretty sure somebody really pushed not to have it.

ajcvickers: Also, IConnection and ITransaction are very general interface names. I wonder if we need something more specific to EF.

src/EntityFramework.InMemory/Storage/InMemoryTransaction.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public class InMemoryTransaction : ITransaction
```
ajcvickers: I'm still concerned that this no-op implementation gives the impression that transactions are supported by in-memory, which really they are not. Should we maybe log a warning, or even throw with a provider flag that could be set to allow no-op transactions? Would be interested to hear what others think.

src/EntityFramework.Core/Storage/IConnection.cs
AndriySvyryd: We don't need `IConnection` for `ITransaction`. The concept of a transaction is not related to the concept of a connection. There might be a global transaction on some data stores.

src/EntityFramework.Relational/Storage/IRelationalTransaction.cs
```diff
@@ -1,16 +1,12 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Data.Common;
 using Microsoft.Data.Entity.Infrastructure;
 
 namespace Microsoft.Data.Entity.Storage
 {
-    public interface IRelationalTransaction : IDisposable, IAccessor<DbTransaction>
+    public interface IRelationalTransaction : ITransaction, IAccessor<DbTransaction>
```
ajcvickers: Does IRelationalTransaction have any value now? It seems to be an empty interface.

src/EntityFramework.Relational/Update/Internal/BatchExecutor.cs
```diff
@@ -21,7 +21,7 @@ public class BatchExecutor : IBatchExecutor
             {
                 if (connection.Transaction == null)
                 {
-                    startedTransaction = connection.BeginTransaction();
+                    startedTransaction = (IRelationalTransaction)connection.BeginTransaction();
```
ajcvickers: Its a bit unfortunate that we need to do a cast here. I'm not sure of any way around it.

src/EntityFramework.InMemory/Storage/InMemoryTransaction.cs
divega: I agree with this. I don't feel strongly on whether this should be a warning or a error + opt out.

src/EntityFramework.Core/Storage/IConnection.cs
divega: Agreed with both @ajcvickers and @AndriySvyryd but I am finding it very hard to imagine good names for the types.

src/EntityFramework.InMemory/Storage/InMemoryTransaction.cs
divega: Feeling slightly inclined to do error + opt out option. It feels like a warning could be more easily missed.

src/EntityFramework.Relational/Update/Internal/BatchExecutor.cs
divega: I am likely missing something but I couldn't find in either `Execute()` or `ExecuteAsync()` any place in which we are using the transaction as anything but `ITransaction`, which BTW is expected since `IRelationalTransaction` is now empty (and should be removed).

src/EntityFramework.Relational/Storage/RelationalConnection.cs
```diff
@@ -92,11 +92,11 @@ protected RelationalConnection([NotNull] IDbContextOptions options, [NotNull] IL
         public virtual DbTransaction DbTransaction => Transaction?.GetService();
```
divega: Can't remember how we ended up exposing both the DbTransaction and I don't know that the introduction of the new interface changes anything, but I have to say it feels weird and I wonder if there is a better way.

divega: The names of the properties also seems weird. It is like the word "Current" is missing there.

src/EntityFramework.InMemory/Storage/InMemoryTransaction.cs
ajcvickers: @divega I think that would be my preference too. Would like to hear from @rowanmiller.

rowanmiller: Agreed, exception with a flag

src/EntityFramework.Relational/Storage/IRelationalTransaction.cs
mikary: It seems unnecessary and could be removed, but as you mention, we would still need a cast to `IAccessor` to access the DbTransaction

src/EntityFramework.Core/Storage/IConnection.cs
mikary: I'm not sure what this change would look like in terms of the `DatabaseFacade` if we kept the concept of connection out of core. We currently extend the facade through extension methods and it seems like we would have to add a `DatabaseFacadeFactory` if we wanted to customize the behavior at the provider level.

ajcvickers: @mikary The DatabaseFacade method needs to get a service on which it calls BegingTransaction, but that service doesn't need to be the connection. I don't see how it requires any changes to the design of DatabaseFacade.

mikary:
Updated based on PR feedback

src/EntityFramework.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
```diff
@@ -12,9 +12,14 @@ namespace Microsoft.Data.Entity
...
     {
-        public static void UseInMemoryDatabase([NotNull] this DbContextOptionsBuilder optionsBuilder)
-            => ((IDbContextOptionsBuilderInfrastructure)Check.NotNull(optionsBuilder, nameof(optionsBuilder)))
-                .AddOrUpdateExtension(
-                    new InMemoryOptionsExtension());
+        public static InMemoryDbContextOptionsBuilder UseInMemoryDatabase([NotNull] this DbContextOptionsBuilder optionsBuilder)
+        {
+            Check.NotNull(optionsBuilder, nameof(optionsBuilder));
+
+            ((IDbContextOptionsBuilderInfrastructure)Check.NotNull(optionsBuilder, nameof(optionsBuilder)))
```
natemcmaster: You can remove the duplicated call to Check.NotNull on optionsBuilder.

src/EntityFramework.InMemory/Infrastructure/Internal/InMemoryOptionsExtension.cs
```diff
@@ -8,6 +9,23 @@ namespace Microsoft.Data.Entity.Infrastructure.Internal
...
+        }
+
+        public InMemoryOptionsExtension([NotNull] InMemoryOptionsExtension copyFrom)
+        {
+            _throwOnTransaction = copyFrom._throwOnTransaction;
+        }
+
+        public virtual bool ThrowOnTransaction
+        {
+            get { return _throwOnTransaction; }
```
natemcmaster: Auto prop?

src/EntityFramework.Relational/Storage/DbContextTransactionExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public static class DbContextTransactionExtensions
+    {
+        public static DbTransaction GetDbTransaction([NotNull] this IDbContextTransaction dbContextTransaction)
+        {
+            Check.NotNull(dbContextTransaction, nameof(dbContextTransaction));
+
+            var accessor = (IAccessor<DbTransaction>)dbContextTransaction;
```
natemcmaster: Possible InvalidCastException?

src/EntityFramework.Relational/Storage/IRelationalConnection.cs
```diff
@@ -10,27 +10,21 @@
...
 {
-    public interface IRelationalConnection : IDisposable
+    public interface IRelationalConnection : ITransactionManager, IDisposable
     {
         string ConnectionString { get; }
 
         DbConnection DbConnection { get; }
 
-        IRelationalTransaction Transaction { get; }
+        IDbContextTransaction CurrentTransaction { get; }
```
natemcmaster: Any reason we need to can't keep this property name? (IMO shorter is better)

mikary: @divega requested this change to alleviate confusion about the semantics of the transaction (i.e. new transaction, ambient transaction, etc.).

src/EntityFramework.InMemory/Infrastructure/Internal/InMemoryOptionsExtension.cs
mikary: Matching the pattern from `RelationalOptionsExtension` and using non-standard default value. I could go either way on this.

src/EntityFramework.Relational/Storage/IRelationalConnection.cs
natemcmaster: Sounds good. :+1:

src/EntityFramework.InMemory/Infrastructure/Internal/InMemoryOptionsExtension.cs
natemcmaster: `public virtual bool ThrowOnTransaction { get; set; } = true`. Seems simple to me, but this is nit that won't prevent a signoff.

src/EntityFramework.Core/Storage/ITransactionManager.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public interface ITransactionManager
```
divega: I don't love it but I think this should be called IDbContexTransactionManager :smile:

mikary:
made some more updates from PR feedback

src/EntityFramework.InMemory/Infrastructure/InMemoryDbContextOptionsBuilder.cs
```diff
@@ -0,0 +1,38 @@
...
+        public InMemoryDbContextOptionsBuilder([NotNull] DbContextOptionsBuilder optionsBuilder)
+        {
+            Check.NotNull(optionsBuilder, nameof(optionsBuilder));
+
+            OptionsBuilder = optionsBuilder;
+        }
+
+        protected virtual DbContextOptionsBuilder OptionsBuilder { get; }
+
+        public virtual InMemoryDbContextOptionsBuilder SuppressTransactionException()
```
ajcvickers: Naming of SuppressTransactionException? @rowanmiller @divega

src/EntityFramework.InMemory/Infrastructure/Internal/InMemoryOptionsExtension.cs
ajcvickers: Its a matter of not calling a virtual property from a constructor. I personally think this is a guideline we can break in some circumstances, but then again when there is a simple way of not breaking the guideline we may as well go with it.

src/EntityFramework.InMemory/Properties/InMemoryStrings.resx
```diff
@@ -60,77 +59,74 @@
...
-    <data name="LogSavedChanges" xml:space="preserve">
+  <data name="LogSavedChanges" xml:space="preserve">
     <value>Saved {count} entities to in-memory store.</value>
   </data>
-    <data name="InMemoryNotInUse" xml:space="preserve">
+  <data name="InMemoryNotInUse" xml:space="preserve">
     <value>In-Memory-specific methods can only be used when the context is using an In-Memory database.</value>
   </data>
+  <data name="TransactionsNotSupported" xml:space="preserve">
+    <value>Transactions are not supported by the in-memory store. To suppress this Exception call SuppressTransactionException() when overriding DbContext.OnConfiguring.</value>
```
ajcvickers: Message wording? @rowanmiller @divega

src/EntityFramework.InMemory/Infrastructure/InMemoryDbContextOptionsBuilder.cs
divega: I don't have any strong candidates in mind but one idea would be to call it IgnoreTransactions. Currently it will flip between the exception and no-op. If anyone every makes InMemory transactional it will flip between having actual transactions and not having them (for back compat).

src/EntityFramework.InMemory/Storage/InMemoryTransaction.cs
```diff
@@ -0,0 +1,44 @@
...
+        public InMemoryTransaction([CanBeNull] InMemoryOptionsExtension inMemoryOptionsExtension)
+        {
+            _inMemoryOptionsExtension = inMemoryOptionsExtension;
+        }
+
+        public virtual void Commit()
+        {
+            if(_inMemoryOptionsExtension.ThrowOnTransaction)
+            {
+                throw new InvalidOperationException(InMemoryStrings.TransactionsNotSupported);
```
ajcvickers: I think it is fine to throw only on BeginTransaction and assume that once they have got one of these then it is because they went through BeginTransaction with the exception disabled. Unless there is some other easy way of getting one of these that I am missing.

src/EntityFramework.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -62,32 +62,24 @@ public static void OpenConnection([NotNull] this DatabaseFacade databaseFacade)
...
-            => GetRelationalConnection(databaseFacade).BeginTransaction();
-
-        public static Task<IRelationalTransaction> BeginTransactionAsync(
-            [NotNull] this DatabaseFacade databaseFacade,
-            CancellationToken cancellationToken = default(CancellationToken))
-            => GetRelationalConnection(databaseFacade).BeginTransactionAsync(cancellationToken);
-
-        public static IRelationalTransaction BeginTransaction([NotNull] this DatabaseFacade databaseFacade, IsolationLevel isolationLevel)
+        public static IDbContextTransaction BeginTransaction([NotNull] this DatabaseFacade databaseFacade, IsolationLevel isolationLevel)
             => GetRelationalConnection(databaseFacade).BeginTransaction(isolationLevel);
```
ajcvickers: Won't this, and other methods that use GetRelationalConnection, fail when using in-memory? That would mean that test code could not use these methods when testing against in-memory. Is that what we want?

src/EntityFramework.InMemory/Storage/InMemoryTransaction.cs
mikary: Yeah, this seemed like a bit of overkill when I was writing it. Other than calling BeginTransaction, this class shouldn't be instantiated.

src/EntityFramework.InMemory/Properties/InMemoryStrings.resx
mikary: Wording is similar to `RelationalStrings.AmbientTransaction`

src/EntityFramework.Core/Storage/IDbContextTransactionManager.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public interface IDbContextTransactionManager
```
ajcvickers: Not a huge fan of this name. @divega @rowanmiller Any ideas?

src/EntityFramework.InMemory/Infrastructure/InMemoryDbContextOptionsBuilder.cs
ajcvickers: I like IgnoreTransactions.

src/EntityFramework.Core/Storage/IDbContextTransactionManager.cs
divega: I am not a fan either, but I proposed this name because I didn't like ITransactionManager better. I can be convinced. Another thought that crossed my mind was to call this interface a factory which would imply the method name should be Create (we would still keep BeginTransaction on the top level API). But not sure it is a good idea.

ajcvickers: Yep, factory and Create is good.

src/EntityFramework.Relational/RelationalDatabaseFacadeExtensions.cs
mikary: Do we want to promote the concept of IsolationLevel to core?

src/EntityFramework.Core/Storage/IDbContextTransactionManager.cs
mikary: Reverted to `IDbContextTransactionManager` as the service is now responsible for holding a reference to the active transaction. The name `ICurrentTransaction` has been suggested, but that may cause confusion with the `RelationalConnection.CurrentTransaction` property that maintains a reference to an `IDbContextTransaction` instance

mikary:
More updates from our discussions, including using relational transaction extensions without throwing on in-memory provider and having a scoped transaction service with facade transaction instance

src/EntityFramework.InMemory/Storage/InMemoryTransaction.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Data.Entity.Storage
+{
+    public class InMemoryTransaction : IDbContextTransaction
+    {
+        public InMemoryTransaction()
```
ajcvickers: Don't need constructor; compiler generates automatically.

src/EntityFramework.InMemory/Storage/InMemoryTransactionManager.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        private readonly InMemoryOptionsExtension _inMemoryOptionsExtension;
+
+        public InMemoryTransactionManager([NotNull] IDbContextOptions options)
+        {
+            Check.NotNull(options, nameof(options));
+
+            _inMemoryOptionsExtension = options.Extensions
+                .OfType<InMemoryOptionsExtension>()
+                .FirstOrDefault();
```
ajcvickers: This uses FirstOrDefault, which implies the result could be null, but the null is never handled anywhere else.

src/EntityFramework.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -62,32 +62,57 @@ public static void OpenConnection([NotNull] this DatabaseFacade databaseFacade)
...
-            => GetRelationalConnection(databaseFacade).BeginTransaction(isolationLevel);
-
-        public static Task<IRelationalTransaction> BeginTransactionAsync(
+        public static IDbContextTransaction BeginTransaction([NotNull] this DatabaseFacade databaseFacade, IsolationLevel isolationLevel)
+        {
+            var transactionManager = GetTransactionManager(databaseFacade);
+
+            var relationalTransactionManager = transactionManager as IRelationalTransactionManager;
+
+            if (relationalTransactionManager != null)
```
ajcvickers: Factor into single return statement.

src/EntityFramework.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -62,32 +62,57 @@ public static void OpenConnection([NotNull] this DatabaseFacade databaseFacade)
...
 
-        public static void RollbackTransaction([NotNull] this DatabaseFacade databaseFacade)
-            => GetRelationalConnection(databaseFacade).Transaction.Rollback();
+            if (relationalTransactionManager != null)
+            {
+                return relationalTransactionManager.UseTransaction(transaction);
+            }
+            else
+            {
+                return transactionManager.BeginTransaction();
```
ajcvickers: Not sure we should go as far as making a UseTransaction with a DbTransaction on relational behave as a BeginTransaction on other providers. @divega?

src/EntityFramework.Relational/Storage/RelationalConnection.cs
```diff
@@ -131,40 +129,64 @@ private IRelationalTransaction BeginTransactionWithNoPreconditions(IsolationLeve
...
+        public virtual void CommitTransaction()
+        {
+            if(CurrentTransaction != null)
+            {
+                CurrentTransaction.Commit();
+            }
+            else
+            {
+                throw new InvalidOperationException(RelationalStrings.NoActiveTransaction);
+            }
```
ajcvickers: Inverse the if and drop the else.

src/EntityFramework.Relational/RelationalDatabaseFacadeExtensions.cs
divega: Update: We agreed that UseTransaction() is only relational.

ajcvickers:
:shipit: After possibly factoring transaction code out of RelationalConnection.

src/EntityFramework.InMemory/Storage/InMemoryTransactionManager.cs
divega: Should it be SingleOrDefault()?

divega: Or First()?

divega: Or Single()? :smile:

src/EntityFramework.Relational/Storage/RelationalTransaction.cs
```diff
@@ -35,15 +36,13 @@ public class RelationalTransaction : IRelationalTransaction
                 throw new InvalidOperationException(RelationalStrings.TransactionAssociatedWithDifferentConnection);
             }
 
-            Connection = connection;
+            _relationalConnection = connection;
 
             _transaction = transaction;
```
divega: Nit: call this _dbTransaction or something to make it less confusing.

src/EntityFramework.InMemory/Storage/InMemoryTransactionManager.cs
mikary: Looks like other places use the FirstOrDefault with null checking for extensions. Seems reasonable to check here

mikary:
Looked into factoring transaction code out of RelationalConnection, but it ends up being messy due to possibility of multiple connections (i.e. SqlServer connection to master), extra DI parameters (most classes that depend on transactions also depend on connection) and potential circular dependencies (connection depends on CurrentTransaction for command building, transaction depends on connection).
Discussed with @ajcvickers and will leave changes as is.

PR #3500, opened 2015-10-21
doc comments for root types in core

rowanmiller:
Remaining types in the root namespace of EntityFramework.Core

The things I think it would be good to read over in detail are:
* EF.Property
* EntityFrameworkServiceCollectionExtensions.AddEntityFramework
* GraphBehavior
* DbContextOptionsBuilder.EnableSensitiveDataLogging

### DIFF
--- a/src/EntityFramework.Core/DbContext.cs
+++ b/src/EntityFramework.Core/DbContext.cs
@@ -243,11 +243,19 @@ namespace Microsoft.Data.Entity
         ///     This method is called for each instance of the context that is created.
         /// </summary>
         /// <remarks>
-        ///     If you passed an instance of <see cref="DbContextOptions" /> to the constructor of the context (or
-        ///     provided an <see cref="IServiceProvider" /> with <see cref="DbContextOptions" /> registered) then
-        ///     it is cloned before being passed to this method. This allows the options to be altered without
-        ///     affecting other context instances that are constructed with the same <see cref="DbContextOptions" />
-        ///     instance.
+        ///     <para>
+        ///         If you passed an instance of <see cref="DbContextOptions" /> to the constructor of the context (or
+        ///         provided an <see cref="IServiceProvider" /> with <see cref="DbContextOptions" /> registered) then
+        ///         it is cloned before being passed to this method. This allows the options to be altered without
+        ///         affecting other context instances that are constructed with the same <see cref="DbContextOptions" />
+        ///         instance.
+        ///     </para>
+        ///     <para>
+        ///         In situations where an instance of <see cref="DbContextOptions" /> may or may not have been passed
+        ///         to the constructor, you can use <see cref="DbContextOptionsBuilder.IsConfigured"/> to determine if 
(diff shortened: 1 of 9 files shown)


src/EntityFramework.Core/ModelBuilder.cs
```diff
@@ -58,6 +53,11 @@ public ModelBuilder([NotNull] ConventionSet conventions, [NotNull] Model model)
             _builder = new InternalModelBuilder(model, conventions).Initialize();
         }
 
+        /// <summary>
+        ///     Checks whether the current model is a valid model that can be used by a context. An exception is thrown if the
+        ///     model is not valid.
+        /// </summary>
+        /// <returns> The same builder instance so that multiple calls can be chained. </returns>
         public virtual ModelBuilder Validate()
```
smitpatel: This actually runs the model built conventions and not validate the model. ModelValidator is called from different place.

rowanmiller: Perhaps a silly question... why is it called "Validate" then? :smile:

smitpatel: I think when first time Model Built conventions were added, it used to verify that composite PK is not set by Data annotations. Now the conventions ran from here are - `KeyAttributeConvention`, `ModelCleanupConvention` & `PropertyMappingValidationConvention`.
Some of them does validation though `ModelValidator` is different from this since ModelValidator runs on IModel, while this runs on ModelBuilder.

rowanmiller: @AndriySvyryd tells me this method is going away, so I will just remove the docs for the moment

src/EntityFramework.Core/DbContext.cs
```diff
@@ -243,11 +243,19 @@ private void InitializeSets(IServiceProvider serviceProvider, DbContextOptions o
...
-        ///     provided an <see cref="IServiceProvider" /> with <see cref="DbContextOptions" /> registered) then
-        ///     it is cloned before being passed to this method. This allows the options to be altered without
-        ///     affecting other context instances that are constructed with the same <see cref="DbContextOptions" />
-        ///     instance.
+        ///     <para>
+        ///         If you passed an instance of <see cref="DbContextOptions" /> to the constructor of the context (or
+        ///         provided an <see cref="IServiceProvider" /> with <see cref="DbContextOptions" /> registered) then
+        ///         it is cloned before being passed to this method. This allows the options to be altered without
+        ///         affecting other context instances that are constructed with the same <see cref="DbContextOptions" />
+        ///         instance.
```
ajcvickers: This is not true anymore. It will be the same instance, so this paragraph can probably just go away. The way it works now is that the DbContextOptions itself is immutable. When the builder operates on it, it creates a new DbContextOptions with the changes applied to it. I think the practical upshot of this is that it will "just work" as expected, even when both passing in a DbContextOptions and doing something to it per instance in OnConfiguring, which is why this paragraph can go away.

ajcvickers:
:shipit:

PR #3501, opened 2015-10-21
[SMALL] Fix to #3180 - A column has been specified more than once in the order by list. Columns in the order by list must be unique.

maumar:
Problem happens for specific combination of nested include statements. When we include reference we add a JOIN on the table that is being included. Issue is that QuerySource of the new table is copied from it's parent table. When include collection, we introduce a new query, but also add order by on the previous query. We determine which table to order based on the query source, but since query sources are the same, the choice is ambiguous. This may lead to queries that try to sort the same column twice, or trying to sort based on column that doesn't exist.

Fix is to always apply includes in a deterministic order that would prevent this situation from happening - first include collections, and only then references.


maumar:
@anpete

anpete:
:shipit:

maumar:
merged a9f0dc05f68b926e2a2ef4d7bfd018fa3871699a

PR #3504, opened 2015-10-22
Changes to use generations TFM

pranavkm:


anpete:
:shipit:

src/EntityFramework.Core/DbSet`.cs
```diff
@@ -314,7 +314,7 @@ IQueryProvider IQueryable.Provider
         }
     }
 
-#if NET45 || DNX451
+#if NET451 || NET451
```
smitpatel: redundant OR

divega: Yes. This was a quick fix for the CI and I suspect @pranavkm did it with a script. I think we should do a pass over the TFMs and #ifs we have in our code and make sure we don't have less or more than what we need. We should revisit if we need to add back plain .NET 4.5.

ErikEJ: @divega Does that mean that you no longer support a provider targeting .NET 4.5 plain? (Like the SQLCE provider currently does) ?

divega: I think it should actually mean that EF itself requires 4.5.1. An ADO.NET provider built for 4.5 should still work since the provider model was extended in compatible ways and there shouldn't be any problem loading it in .NET 4.5.1.
But regardless, this maybe just a temporary change while we stabilize things using the new framework monikers across all of the Microsoft.Extensions.* packages. Stay tuned.

ErikEJ: I mean that the EF.SqlServerCompact40 package only requires 4.5 currently! But maybe I need to change that to 4,5,1?

src/EntityFramework.Core/project.json
```diff
@@ -84,27 +65,6 @@
         "System.Threading": "4.0.11-beta-*",
         "System.Threading.Tasks": "4.0.11-beta-*"
       }
-    },
-    "dotnet": {
```
natemcmaster: No. We need this. Using "-*" for system packages breaks UWP.

natemcmaster:
Was this merged with https://github.com/aspnet/EntityFramework/commit/7e0ba479a2cc1242ff7d6688687a740bd8a606f4?

pranavkm:
Yup.

src/EntityFramework.Core/project.json
```diff
@@ -26,26 +26,7 @@
...
-        "System.Linq": { "version": "4.0.0.0", "type": "build" },
-        "System.Linq.Expressions": { "version": "4.0.0.0", "type": "build" },
-        "System.Linq.Queryable": { "version": "4.0.0.0", "type": "build" },
-        "System.ObjectModel": { "version": "4.0.0.0", "type": "build" },
-        "System.Reflection": { "version": "4.0.0.0", "type": "build" },
-        "System.Reflection.Extensions": { "version": "4.0.0.0", "type": "build" },
-        "System.Resources.ResourceManager": { "version": "4.0.0.0", "type": "build" },
-        "System.Runtime": { "version": "4.0.0.0", "type": "build" },
-        "System.Runtime.Extensions": { "version": "4.0.0.0", "type": "build" },
-        "System.Threading": { "version": "4.0.0.0", "type": "build" }
```
bricelam: Won't this break installing on net451? The design-time facades shouldn't be listed in the nuspec

bricelam: (But they were needed on DNX)

pranavkm: Why does listing these as framework references break on net451?

davidfowl: yes it will

davidfowl: Tho I dunno if all break. System.Runtime certainly does

pranavkm: What's the fix here - add a separate `DNX451` node?

ErikEJ: Please don't break .NET 4.5.1...

bricelam: Yes, that's one of the main reasons we had both net45 and dnx451 listed

pranavkm: @davidfowl wants to fix this in DNX as a pack time gesture. @davidfowl, would this be an appropriate work item to track this issue: https://github.com/aspnet/dnx/issues/2031?

PR #3505, opened 2015-10-22
Add IMutableModel API

AndriySvyryd:
Cleanup Model API


src/EntityFramework.Core/ChangeTracking/Internal/NavigationFixer.cs
```diff
@@ -34,13 +34,13 @@ public class NavigationFixer : INavigationFixer
         }
 
         public virtual void ForeignKeyPropertyChanged(InternalEntityEntry entry, IProperty property, object oldValue, object newValue)
-            => PerformFixup(() => ForeignKeyPropertyChangedAction(entry, property, oldValue, newValue));
+            => PerformFixup(() => ForeignKeyPropertyChangedAction(entry, property));
```
divega: Do we drop newValue and oldValue from the calling method?

AndriySvyryd: That would require changing the interface. @ajcvickers would know better if we wanted to do that.

src/EntityFramework.Core/Metadata/EntityType.cs
```diff
@@ -657,10 +654,18 @@ public virtual IEnumerable<Navigation> FindDerivedNavigations([NotNull] string n
         public virtual IEnumerable<Navigation> FindNavigationsInHierarchy([NotNull] string propertyName)
             => ToEnumerable(FindNavigation(propertyName)).Concat(FindDerivedNavigations(propertyName));
 
-        public virtual Navigation RemoveNavigation([NotNull] Navigation navigation)
+        public virtual Navigation RemoveNavigation([NotNull] string name)
```
ajcvickers: Why are we making these changes from strong Navigation parameter to weak string parameter, strong ForeignKey parameter to weak property list parameter, and so on? Is this really all part of the same change?

AndriySvyryd: The effective parameter strength is the same, since the only thing we used from it is the id value. Passing in just the id makes implementing the interface easier and the API more consistent.

src/EntityFramework.Core/Metadata/IMutableProperty.cs
```diff
@@ -0,0 +1,22 @@
...
+        new IMutableEntityType DeclaringEntityType { get; }
+        new Type ClrType { get; [param: NotNull] set; }
+        new bool? IsNullable { get; set; }
+        new ValueGenerated? ValueGenerated { get; set; }
+        new bool? IsReadOnlyBeforeSave { get; set; }
+        new bool? IsReadOnlyAfterSave { get; set; }
+        new bool? RequiresValueGenerator { get; set; }
+        new bool? IsShadowProperty { get; set; }
+        new bool? IsConcurrencyToken { get; set; }
+        new bool? StoreGeneratedAlways { get; set; }
```
ajcvickers: I assume that once other changes related to configuration source are made then these will not be nullable. Also I have never seen "new: used in an interface like this. I hope there aren't any gotchas. ;-)

src/EntityFramework.Core/ChangeTracking/Internal/NavigationFixer.cs
ajcvickers: I don't think we need to change the interface at this time. It's internal anyway, so it doesn't really matter.

src/EntityFramework.Core/Metadata/IMutableProperty.cs
AndriySvyryd: Correct. `new` works just as one would expect as far as I can tell.

ajcvickers:
This seems like a massive change to metadata. It would be really good to do proper API review of it, but I know time is very tight. I'm signing off on it, but I don't know it is really "right". :sheep::it:

rowanmiller:
Lets review it tomorrow.

rowanmiller:
BTW I don't mean we should hold off on checking in... we can make changes as needed on what we have.

src/EntityFramework.Core/ChangeTracking/Internal/NavigationFixer.cs
divega: I asked originally because I was surprised the parameters didn't matter anymore and wanted to make sure this change was intentional. If you guys know why it is ok to drop the parameter values on the floor, then no problem.

ajcvickers: It's okay because they are not being used in ForeignKeyPropertyChangedAction.

PR #3508, opened 2015-10-22
[TINY] Remove initialNodeState from ChangeTracker.TrackGraph

rowanmiller:
Discussed with @ajcvickers and concluded that this is a pretty advanced
concept and can be removed from this top level API. It is still present
in the EntityGraphNode and EntityGraphIterator. You can also use it with
TrackGraph if you want to, you just need to check for a null state and
set it within the callback.


divega:
:shipit: a bit puzzled that this doesn't required any test to change :smile:

julielerman:
ahh will ignore that parameter then! :)

PR #3509, opened 2015-10-22
[api-review] Updates behavior of scaffolding. 

natemcmaster:
Leftovers from API review on scaffolding
(see https://github.com/aspnet/EntityFramework/pull/3489#issuecomment-150025846).

NB: *not* including unscaffoldable entity types in the model had far reaching affects that required removing the "ReferredToByTableWIthUnmappablePrimaryKey" test from SqlServer.

Changes:
- Remove IMethodNameProvider and use annotation instead. Throw when missing.
- Make TableSelectionSet immutable
- Make EntityType errors annotations on model, rather than including empty entity types in the scaffolding model
- Respect IsClustered indexes on SqlServer
- Change ColumnModel to use ValueGenerated enum
- Remove ExplicitValueGenerationNever annotation


lajones:
@natemcmaster Can you explain more about the removal of the `ReferredToByTableWIthUnmappablePrimaryKey` test? That situation does occur e.g. in the AdventureWorks2012 DB. And BTW can you make sure that running `dnx ef dbcontext scaffold` on AdventureWorks2012 does work before checking in - it seems to be broken right now by some earlier changes?

natemcmaster:
The model used to have "invalid" entity types (one with and EntityTypeError annotation). `EntityTypeWriter` relied on this in code-gen to produce a comment  [such as this one](https://github.com/aspnet/EntityFramework/pull/3509/files#diff-964df143256ade24dba50510dc2db519L12). We don't include these invalid entity types anymore.

natemcmaster:
Btw, this needs to be updated quite a bit to reflect changes made in de4f73e9517ba75f69064e45ff30d2755a3981d3

lajones:
@natemcmaster  Yep - I remember that the mtg decided to move the comment from the generated `EntityType` to the where the `DbSet`'s are defined in the `DbContext` instead. But the scenario of having a table in the DB with a FK which points to a table which whose primary key we can't map should remain as a scenario - I think we should keep the test but update what we expect as results.

natemcmaster:
@lajones then we will need to restructure somewhat. We cannot add a foreign key to the model without the existence of the principal entity type, so we need another annotation to carry this data through to code gen.

lajones:
@natemcmaster Yep understood :disappointed: But we will have to do something to handle this. The scenario does exist in real databases (e.g. AdventureWorks2012).

natemcmaster:
@lajones I just tested. Scaffolding completes w/o crashing on AdventureWorks. [Results here](https://gist.github.com/natemcmaster/31d3a40ce9f802933327)

divega:
@natemcmaster @lajones just to check, is the only difference that any entities that could have potentially had an FK and navigation pointing to 'Production.ProductDocument', 'Production.Document', 'Person.PersonPhone' used to be generated with a comment and now will still be generated but without any comments? If that is the case I think it is fine. We can file a bug to try to bring those comments back by placing an annotation in those entities.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -197,12 +206,18 @@ private void GetIndexes()
                     if (index == null
                         || index.Name != indexName)
                     {
-                        var table = _tables[TableKey(tableName, schemaName)];
+                        TableModel table;
+                        if(!_tables.TryGetValue(TableKey(tableName, schemaName), out table))
+                        {
+                            continue;
```
lajones: Feels like we should log a warning here: "Unable to find table XXX for index YYY".

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
```diff
@@ -35,10 +35,22 @@ public class SqlServerScaffoldingModelFactory : RelationalScaffoldingModelFactor
...
+            var model = base.Create(connectionString, tableSelectionSet);
+            model.Scaffolding().UseProviderMethodName = nameof(SqlServerDbContextOptionsExtensions.UseSqlServer);
+            return model;
+        }
+
         protected override PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder builder, [NotNull] ColumnModel column)
         {
             var propertyBuilder = base.VisitColumn(builder, column);
 
+            if(propertyBuilder == null)
```
lajones: When does this happen? Is there an error logged for it?

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
```diff
@@ -59,7 +76,7 @@ protected override EntityTypeBuilder VisitPrimaryKey([NotNull] EntityTypeBuilder
             var pkColumns = table.Columns.Where(c => c.PrimaryKeyOrdinal.HasValue).ToList();
```
lajones: Do we have a bug logged for using KeyConvention? IIRC it's not just that the PK consists of 1 column - it's also that the column has to be of a specific set of types?

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
natemcmaster: Added as a TODO on https://github.com/aspnet/EntityFramework/issues/3454

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
natemcmaster: The base implementation will log when it encounters an error and return null. This is just passing this failed condition up the call stack.

src/EntityFramework.Relational.Design/Internal/Configuration/ModelConfiguration.cs
```diff
@@ -124,9 +117,16 @@ public virtual void AddEntityConfigurations()
 
         public virtual void AddConnectionStringConfiguration()
         {
+            var methodName = Model.Scaffolding().UseProviderMethodName;
+
+            if (string.IsNullOrEmpty(methodName))
+            {
+                throw new Exception("Cannot scaffolding connection string. Missing the provider name annotation.");
```
lajones: I think we use OperationException. Also exception messages should be localized. Also there is no workaround - so I think the exception should say something like "Contact your provider and tell them that they must define the UseProviderMethodName in the XXX method of their YYYY class."

src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
```diff
@@ -132,8 +132,20 @@
...
     <value>There was an error running the EntityType template. Message: {errorMessage}</value>
   </data>
-  <data name="ForeignKeyScaffoldError" xml:space="preserve">
-    <value>Could not scaffold the foreign key '{foreignKeyName}'. One of the referenced properties or the referenced table could not be found.</value>
+  <data name="ForeignKeyScaffoldErrorPrincipalKeyNotFound" xml:space="preserve">
+    <value>Could not scaffold the foreign key '{foreignKeyName}'. A key for '{columnsList}' was not found in the principal entity type '{principalEntityType}'.</value>
+  </data>
+  <data name="ForeignKeyScaffoldErrorPrincipalTableNotFound" xml:space="preserve">
+    <value>Could not scaffold the foreign key '{foreignKeyName}'. The referenced table could not be found.</value>
+  </data>
```
lajones: Add the name of the referenced table.

src/EntityFramework.Relational.Design/Internal/Configuration/ModelConfiguration.cs
natemcmaster: This is a provider-writer-time exception. Users should never see it. I'll add clarification that provider-writers need to add an annotation.

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -256,8 +250,10 @@ protected virtual EntityTypeBuilder VisitIndexes([NotNull] EntityTypeBuilder bui
         protected virtual IndexBuilder VisitIndex([NotNull] EntityTypeBuilder builder, [NotNull] IndexModel index)
         {
             var properties = index.Columns.Select(GetPropertyName).ToArray();
-            if (properties.Any(i => i == null))
+
+            if (properties.Count(p => builder.Metadata.FindProperty(p) != null) != properties.Length)
             {
+                // TODO log when index cannot be scaffolding because of missing columns
```
lajones: Make sure there's a bug for this.

src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
natemcmaster: This is for when the table was filtered. We don't know the name.

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -359,14 +363,12 @@ private IMutableForeignKey TryVisitForeignKey(ModelBuilder modelBuilder, Foreign
...
-                RelationalDesignStrings.ForeignKeyScaffoldError(foreignKey.DisplayName));
-
-        protected virtual void VisitNavigationProperties([NotNull] IModel model)
+        protected virtual void VisitNavigationProperties([NotNull] ModelBuilder modelBuilder)
         {
             // TODO perf cleanup can we do this in 1 loop instead of 2?
+            var model = modelBuilder.Model;
             var modelUtilities = new ModelUtilities();
+
             Check.NotNull(model, nameof(model));
```
lajones: Shouldn't this line be referring to `modelBuilder` and be at the start of the method?

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
natemcmaster: https://github.com/aspnet/EntityFramework/issues/3456

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/ReverseEngineering/E2E.sql
```diff
@@ -327,24 +323,10 @@ CREATE TABLE "OneToOneFKToUniqueKeyDependent" (
 
 GO
 
-CREATE TABLE "ReferredToByTableWithUnmappablePrimaryKeyColumn" (
```
lajones: I thought we'd fixed this. Can't we put this test back in? Just expect a comment in the `DbContext` file instead of an EntityType.cs file?

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
natemcmaster: Yup. Missed this in refactoring.

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/ReverseEngineering/E2E.sql
natemcmaster: Why? It doesn’t test anything not covered in other tests.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
lajones: Sounds good.

src/EntityFramework.Relational.Design/Internal/Configuration/ModelConfiguration.cs
lajones: Sounds good. But I still think we're supposed to localize. Also shouldn't it be "Cannot scaffold ..." rather than "Cannot scaffolding ..."?

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
natemcmaster: https://github.com/aspnet/EntityFramework/issues/3534

src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
lajones: Are we throwing this too late then? Can we instead throw back where we construct the ForeignKeyModel with the null PrinicpalTable - we know it's going to cause problems later and at that point we know what the sys tables are telling us the table on the other end of the FK is. Then we can include that in the message to the user warning that perhaps it was filtered out by the table selection.

src/EntityFramework.Relational.Design/Internal/Configuration/ModelConfiguration.cs
natemcmaster: Updated.

natemcmaster:
@divega yes, the only behavior change is that we no longer generate about the could-have-been nav prop.

src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
natemcmaster: I don’t think that is useful information to output on the console. The user already knows about filtered tables/schemas as they are specified on the command line.

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/ReverseEngineering/E2E.sql
lajones: I think you're basing this on the fact that part of this tested the commented out the NavProp which has gone away with this fix - but we are planning to fix that at which point we'd need this test again. And in any case the scenario remains valid. Otherwise where do we have coverage of generating a class where there is a FK which refers to it but the target of the FK is in a class which we are not generating because of an error?

src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
lajones: I'm not suggesting outputting the filters themselves, but the table that you were trying to target does seem like useful information. A warning message such as "Could not scaffold the foreign key {FKname} because its target table {principalTableName} is not included in this model. Was it filtered out by the table selection?" makes the user go "Oh yes - I should just include that table in my filters" which seems more useful to me.

test/EntityFramework.Sqlite.Design.FunctionalTests/ReverseEngineering/Expected/AllFluentApi/NoPrincipalPk/NoPrincipalPkFluentApiContext.expected
```diff
@@ -19,5 +19,7 @@ namespace E2E.Sqlite
         }
 
         public virtual DbSet<Dependent> Dependent { get; set; }
+
+        // Unable to generate entity type for table 'Principal'.
```
lajones: We seem to have removed the reason why we couldn't generate the entity type. Is this necessary because you are generating this message in a different part of the code now and so don't have that info? If so could you add "Please see the warning messages." to the end? I like messages to guide the user towards what to do to fix the problem.

natemcmaster: Updated. Restored the test and updated error messages.

src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
```diff
@@ -132,8 +132,20 @@
   <data name="ErrorRunningEntityTypeTemplate" xml:space="preserve">
     <value>There was an error running the EntityType template. Message: {errorMessage}</value>
   </data>
-  <data name="ForeignKeyScaffoldError" xml:space="preserve">
-    <value>Could not scaffold the foreign key '{foreignKeyName}'. One of the referenced properties or the referenced table could not be found.</value>
+  <data name="ForeignKeyScaffoldErrorPrincipalKeyNotFound" xml:space="preserve">
+    <value>Could not scaffold the foreign key '{foreignKeyName}'. A key for '{columnsList}' was not found in the principal entity type '{principalEntityType}'.</value>
+  </data>
+  <data name="ForeignKeyScaffoldErrorPrincipalTableNotFound" xml:space="preserve">
+    <value>Could not scaffold the foreign key '{foreignKeyName}'. The referenced table could not be found. This most likely occurred because the referenced table was excluded from scaffolding.</value>
```
lajones: Well this wasn't quite what I wanted but I understand you want to get this in so go ahead for now. But could you add revisiting this to one of the bugs you raised or a new one please?

lajones:
:shipit:

PR #3511, opened 2015-10-22
[SMALL] RevEng: ensure it understands default values for datetime columns…

### LINKED ISSUE #3250: RevEng: datetime field with default constraint generates warning
The command 

    Scaffold-DbContext -Connection "<connection string>" -Provider EntityFramework.SqlServer

Generates the following warning

    For column [dbo][MyTable][MyDateTime] unable to convert default value ('3000-01-01T00:00:00') into type System.DateTime. Will not generate code setting a default value for the property [dbo][MyTable][MyDateTime] on entity type [dbo][MyTable].

If [MyTable][MyDateTime] column is defined as [MyDateTime] `[datetime]` NOT NULL,
and the following constraint exists

    ALTER TABLE [dbo].[MyDateTime] ADD  CONSTRAINT [DF_MyTable_MyDateTime]  DEFAULT ('3000-01-01T00:00:00') FOR [MyDateTime]

lajones:
...which are fixed values.

Fixes #3250.

### DIFF
--- a/src/EntityFramework.MicrosoftSqlServer.Design/Internal/SqlServerLiteralUtilities.cs
+++ b/src/EntityFramework.MicrosoftSqlServer.Design/Internal/SqlServerLiteralUtilities.cs
@@ -132,6 +132,14 @@ namespace Microsoft.Data.Entity.Scaffolding.Internal
 
             propertyType = propertyType.UnwrapNullableType();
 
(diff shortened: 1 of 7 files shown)


natemcmaster:
:shipit:

lajones:
Checked in with commit fc0d25c.

PR #3512, opened 2015-10-22
Change service provider cache for better performance

ajcvickers:
Previously we would build a cache key based on all the services. Now we just look at which extensions have been added and use that. This should have the same behavior based on the following assumptions:
- This is only used when we are building the service provider internally, and there are no hooks to modify services in this case. If such a hook is added, then it will need to be able to influence the cache key creation
- Each provider adds unchanging set of services--i.e. there is no logic which says sometimes add this service, sometimes don't.

### DIFF
--- a/src/EntityFramework.Core/Internal/ServiceProviderCache.cs
+++ b/src/EntityFramework.Core/Internal/ServiceProviderCache.cs
@@ -18,28 +18,26 @@ namespace Microsoft.Data.Entity.Internal
 
         public virtual IServiceProvider GetOrAdd([NotNull] IDbContextOptions options)
         {
-            var services = new ServiceCollection();
-            var builder = services.AddEntityFramework();
-
-            foreach (var extension in options.Extensions)
-            {
-                extension.ApplyServices(builder);
-            }
-
             // Decided that this hashing algorithm is robust enough. See issue #762.
             unchecked
             {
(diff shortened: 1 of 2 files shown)


src/EntityFramework.Core/Internal/ServiceProviderCache.cs
```diff
@@ -18,28 +18,26 @@ public class ServiceProviderCache
...
-                extension.ApplyServices(builder);
-            }
-
             // Decided that this hashing algorithm is robust enough. See issue #762.
             unchecked
             {
-                var key = services.Aggregate(0, (t, d) => (t * 397) ^ CalculateHash(d).GetHashCode());
-
-                return _configurations.GetOrAdd(key, k => services.BuildServiceProvider());
+                var key = options.Extensions.Aggregate(0, (t, e) => (t * 397) ^ e.GetType().GetHashCode());
```
anpete: If this path is hot then consider de-linqing.

anpete:
:shipit:

src/EntityFramework.Core/Internal/ServiceProviderCache.cs
ajcvickers: It is currently 1% of total time. I am still looking at other changes.

PR #3513, opened 2015-10-22
Doc comments for ChangeTracking.*

rowanmiller:
It would be good to get a detailed review of what I wrote for QueryTrackingBehavior

@ajcvickers can you take a quick look at what I wrote for the GraphNode classes and make sure it is technically correct.


ajcvickers:
:shipit:

PR #3514, opened 2015-10-22
Query: Performance: Only use QueryBuffer on queries that require it:

anpete:
- Tracked queries, queries that use Include and queries that access shadow state.


natemcmaster:
:shipit:

PR #3515, opened 2015-10-22
Replace ThreadSafeDictionaryCache with ConcurrentDictionary

ajcvickers:
ThreadSafeDictionaryCache was implemented on top of ImmutableDictionary, possibly before ConcurrentDictionary was available, but also for perceived performance gains especially where reading from the cache (as opposed to adding to it) dominates. However, perf testing is showing the ConcurrentDictionary is considerably (30-40%) faster even on read-heavy scenarios, so replacing to use this instead.


smitpatel:
:shipit:

ajcvickers:
Merged

PR #3516, opened 2015-10-22
Add support for simple queries in .Net Native.

natemcmaster:
Still unsupported: queries that
 - Use projections
 - Use async
 - Use join on custom types
 - Use shadow properties (`EF.Property`)

About half of the tests in [QueryTestBase](https://github.com/aspnet/EntityFramework/blob/dev/test/EntityFramework.Core.FunctionalTests/QueryTestBase.cs) pass on .NET Native.

Reminder: this will not work at all until https://github.com/re-motion/Relinq/pull/6 is merged and available on nuget.

This is incomplete, but submitting for review because this restructures how we get ILC to code gen the generic methods we need.


anpete:
:shipit:

PR #3518, opened 2015-10-22
Refactor dependency injection usage in entity framework

pakrym:
@davidfowl

### DIFF
--- a/src/EntityFramework.Commands/Commands/Program.cs
+++ b/src/EntityFramework.Commands/Commands/Program.cs
@@ -23,25 +23,16 @@ namespace Microsoft.Data.Entity.Commands
 {
     public class Program
     {
-        private readonly IApplicationShutdown _applicationShutdown;
-        private readonly bool _useConsoleColors;
-        private readonly IServiceProvider _dnxServices;
+        private static readonly bool _useConsoleColors;
 
-        public Program([NotNull] IServiceProvider dnxServices)
+        static Program()
         {
-            Check.NotNull(dnxServices, nameof(dnxServices));
-
-            var runtimeEnv = dnxServices.GetRequiredService<IRuntimeEnvironment>();
-            _applicationShutdown = dnxServices.GetRequiredService<IApplicationShutdown>();
-            _useConsoleColors = runtimeEnv.OperatingSystem == "Windows";
-            _dnxServices = dnxServices;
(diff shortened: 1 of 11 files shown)


src/EntityFramework.Commands/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -8,23 +8,20 @@
...
-                foreach (var service in runtimeServices.Services)
-                {
-                    services.AddTransient(service, _ => dnxServices.GetService(service));
-                }
-            }
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.Application));
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.Runtime));
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.AssemblyLoadContextAccessor));
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.AssemblyLoaderContainer));
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.LibraryManager));
```
davidfowl: Add compilation services

natemcmaster:
Looks good, but before signing off, let's verify it doesn't break any of the scenarios in https://github.com/aspnet/EntityFramework/issues/2294. By inspection, I'm not sure they will all work.

natemcmaster:
:shipit:

src/EntityFramework.Commands/Commands/Program.cs
```diff
@@ -23,26 +23,17 @@ namespace Microsoft.Data.Entity.Commands
...
-            _dnxServices = dnxServices;
+            _useConsoleColors = PlatformServices.Default.Runtime.OperatingSystem == "Windows";
         }
 
-        public virtual int Main([NotNull] string[] args)
+        public static int Main([NotNull] string[] args)
         {
             Check.NotNull(args, nameof(args));
 
-            Console.CancelKeyPress += (_, __) => _applicationShutdown.RequestShutdown();
```
bricelam: Wait, we still need to cancel the async call in this case

bricelam:
Was this supposed to go into RC1? I thought this work wasn't happening until RC2...

src/EntityFramework.Commands/Commands/Program.cs
bricelam: Fixed in #3567

PR #3519, opened 2015-10-22
[SMALL] RevEng: No longer output AlternateKey() Fluent API.

lajones:
Partial fix for #2840.

One output from the meeting on 10/21/15 was that we should no longer output the HasAlternateKey() fluent API. This PR fixes that.


natemcmaster:
:shipit:

lajones:
Checked in with commit dd2bb41.

PR #3522, opened 2015-10-23
Remove throwing Get methods from Metadata API

AndriySvyryd:


src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -300,12 +300,12 @@ protected virtual IMutableForeignKey VisitForeignKey([NotNull] ModelBuilder mode
 
             try
             {
-                var dependentEntityType = modelBuilder.Model.GetEntityType(GetEntityTypeName(foreignKey.Table));
-                var principalEntityType = modelBuilder.Model.GetEntityType(GetEntityTypeName(foreignKey.PrincipalTable));
+                var dependentEntityType = modelBuilder.Model.FindEntityType(GetEntityTypeName(foreignKey.Table));
+                var principalEntityType = modelBuilder.Model.FindEntityType(GetEntityTypeName(foreignKey.PrincipalTable));
```
natemcmaster: This is going to lead to NullReferenceExceptions. In this case (and others), the "Get" method doubles as an assertion.

AndriySvyryd: Yes, we decided that NRE is ok if it's caused by a bug.

natemcmaster: In this case, it is not a bug, but expected behavior. This is easy to get around by adding a check for null after calling .Find, but it needs to be there for scaffolding to work correctly. Did testing not catch this?

AndriySvyryd: In that case I'll add null checks and remove code that catches the exceptions.
Tests are still running.

natemcmaster:
:shipit:

PR #3528, opened 2015-10-23
[SMALL] RevEng: Changing DesignTimeProviderServicesAttribute

### LINKED ISSUE #3111: Use fully qualified assembly name in DesignTimeProviderServicesAttribute
Combine assembly name and type name into one string to be consistent with other places that take a type name as a string. Add a package name property to be used for logging/error messages.

lajones:
Fixes #3111.
@ajcvickers Let me know if this is what you intended.


ajcvickers:
:shipit:

lajones:
Checked in with commit 71f0bf8.
@ErikEJ Note: you'll need to update your provider to work the same way. Sorry for the bother, but the change is small.

ErikEJ:
@lajones Thanks, I will be staring over with the provider soon, now that the dust has settled and once  #3553 has been resolved

PR #3530, opened 2015-10-23
[small] :bug: Validate property without CLR type set

smitpatel:
The PropertyMappingValidation threw an exception if there is property without CLR type set is present in the model. Converted to use `IProperty.ClrType` so that we use the default value for clr type.


AndriySvyryd:
:shipit:

Issue #3452, opened 2015-10-16
Error when return complex linq to json object The EF.Property<T> method may only be used within LINQ queries.


Anderman:
When returning a linq query with a subsquery I get an error. This query was working in Beta7 but not after the upgrade to beta8.
```
 public JsonResult Index([FromBody]DataTablesRequest dTRequest)
        {
            var result = (from u in DbContext.Users
                        select new
                        {
                            u.Id,
                            Logins = u.Logins.Select(l => new { LoginProvider = l.LoginProvider })
                        });
            return new JsonResult(result);
}
```
or a linq expression like this give the same error
```
var result = (from u in DbContext.Users
     select new
     {
          Logins = u.Logins.Select(l => new { LoginProvider = l.LoginProvider }).ToArray()
      }).ToArray();
```
The Error message

```
warning : [Microsoft.Data.Entity.Query.QueryCompilationContext] The LINQ expression '(Property([u], "UserId") == Property([l], "Id"))' could not be translated and will be evaluated locally.
error   : [Microsoft.AspNet.Diagnostics.DeveloperExceptionPageMiddleware] An unhandled exception has occurred while executing the request
System.InvalidOperationException: The EF.Property<T> method may only be used within LINQ queries.
   at Microsoft.Data.Entity.EF.Property[TProperty](Object entity, String propertyName)
   at lambda_method(Closure , QueryResultScope )
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeList(JsonWriter writer, IEnumerable values, JsonArrayContract contract, JsonProperty member, JsonContainerContract collectionC
ontract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContrac
t, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionCont
ract, JsonProperty containerProperty)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeValue(JsonWriter writer, Object value, JsonContract valueContract, JsonProperty member, JsonContainerContract containerContrac
[...]

PR #3533, opened 2015-10-23
[SMALL] Fix to Issue #3452 - Error when return complex linq to json object The EF.Property<T> method may only be used within LINQ queries.

maumar:
Problem was that we incorrectly that relationship would flow in one direction always assigning foreign key property to the outer key selector and primary key to the inner selector for the comparison.
Fix is to check the direction of the navigation and correctly construct the MemberExpressions. Also making sure we compensate for differences in nullabilities between keys, when necessary.


maumar:
@anpete

PR #3536, opened 2015-10-23
Fix #2915 - Make more query stuff internal.

### LINKED ISSUE #2915: Move classes that we don't expect app devs or provider writers to use to .Internal

anpete:


divega:
:shipit:

rowanmiller:
This makes me very happy... less stuff to document :tada: :smile:

PR #3539, opened 2015-10-24
Reacting to fix for #3161.

maumar:
We had multiple SaveChanges in the InheritanceRelationshipsModelInitializer when adding entities of two different types in a hierarchy. With the fix to 3161 they are no longer needed and can be removed.


maumar:
@smitpatel

smitpatel:
:shipit:

PR #3542, opened 2015-10-24
Create shadow ForeignKey properties based on NavigationToPrincipal Name

### LINKED ISSUE #3503: Regression? Foreign key properties now based on type name rather than property name
I have an entity class with the following properties:

```c#
public virtual ContourTemplate TailContour { get; set; }
public virtual ContourTemplate NoseContour { get; set; }
public virtual ContourTemplate MidContour { get; set; }
```

In beta7, the column names generated were `TailContourId`, `NoseContourId` and `MidContourId`. That's perfect and exactly what it should be. I've had to make some changes and in the process have dumped my old migrations from beta 7 and created new ones with beta 8. The new migrations generated by beta 8 now look like this though:

```c#
ContourTemplateId = table.Column<int>(nullable: false),
ContourTemplateId1 = table.Column<int>(nullable: false),
ContourTemplateId2 = table.Column<int>(nullable: false),
```

This is less than ideal. Why the change?

**Edit**  
I just added `[Column("TailContourId")]` and so forth to each property and it had no effect on the generated migration code.

**Edit**  
Looks like adding `ForeignKey` in place of `Column` does the trick. Original question still stands though.

smitpatel:
resolves #3503 
Issue here was when we were creating the relationship we create the foreign key first and then add navigation. Therefore the `NavigationToPrincipalName` was always null while creating foreign key, hence navigation name was never used while creating shadow properties for FK.
Solution - Pass the Navigation to Principal Name as argument even when creating FK (even though navigation is not set) so that the shadow properties have appropriate names.
@AndriySvyryd - Is this correct way to fix this? Anymore tests needed?

### DIFF
--- a/src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Builders/EntityTypeBuilder.cs
@@ -318,7 +318,7 @@ namespace Microsoft.Data.Entity.Metadata.Builders
         protected virtual InternalRelationshipBuilder CollectionBuilder(
             [NotNull] EntityType relatedEntityType, [CanBeNull] string navigationName)
             => Builder.ModelBuilder.Entity(relatedEntityType.Name, ConfigurationSource.Explicit)
-                .Relationship(Builder, ConfigurationSource.Explicit)
+                .Relationship(Builder, null, ConfigurationSource.Explicit)
                 .DependentEntityType(relatedEntityType, ConfigurationSource.Explicit)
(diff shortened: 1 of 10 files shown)


AndriySvyryd:
@smitpatel This change makes the caller to supply the navigation name twice, this shouldn't be necessary. Fixing this bug shouldn't require changing the calling code.
Setting the navigation to principal on the relationship should result in a call to `InternalRelationshipBuilder.Relationship` which should recreate the foreign key with correct property names. The bug is somewhere in this chain.

PR #3547, opened 2015-10-24
Small API changes discussed with others while writing doc comments

rowanmiller:
Moving CoreOptionsExtension to Internal
Moving QueryTracking behavior to root namespace (because it is set on a top level DbContext.ChangeTracker API)
Renaming IAccessor


ajcvickers:
:shipit:

PR #3559, opened 2015-10-26
Remove duplicate test case.

JoshVarty:
This duplicate helped catch a bug in a test runner I was building, but I'm guessing you don't need it. :)


natemcmaster:
Thanks @JoshVarty.

PR #3562, opened 2015-10-26
Collapse EntityGraphNode hierarchy & remove context from EntityEntry ctor

rowanmiller:
* Simplifies things by combining the EntityEntry and InternalEntityEntry implementations of EntityGraphNode. Using a similar pattern to what we use elsewhere by having InternalEntityEntry hidden by IInfrastructure. Also making EntityEntry creation lazy so that we don't create them when we are running internal operations that just use the InternalEntityEntry API.
* Also removing the DbContext parameter on the EntityEntry ctor since this can now be obtained from the InternalEntityEntry (combining into this commit since it simplifies the changes a little).


src/EntityFramework.Core/ChangeTracking/EntityEntryGraphNode.cs
```diff
@@ -12,37 +14,53 @@ namespace Microsoft.Data.Entity.ChangeTracking
     ///     Provides access to change tracking information and operations for a node in a 
     ///     graph of entities that is being traversed.
     /// </summary>
-    public class EntityEntryGraphNode : EntityGraphNodeBase<EntityEntryGraphNode>
+    public class EntityEntryGraphNode : IInfrastructure<InternalEntityEntry>
     {
+        private readonly InternalEntityEntry _internalEntityEntry;
+        private readonly Lazy<EntityEntry> _entityEntry;
```
ajcvickers: Use LazyRef instead since it is lightweight and this doesn't need to be thread-safe. But, there really isn't any point in doing this--the Lazy/LazyRef allocates an object anyway, so if we want to return the same instance every time, then just create it eagerly. If it's okay to create a new instance when asked for, then do that, and consider making a method rather than a property.

ajcvickers:
:shipit:

src/EntityFramework.Core/ChangeTracking/EntityEntryGraphNode.cs
rowanmiller: It's nice not to create them when we use this internally, so I'll just go with the new instance each time and we can optimize later if needed.

PR #3566, opened 2015-10-26
[SMALL] Move *AnnotationNames => .Internal

rowanmiller:
The public surface on these are the extension methods we expose. The
only place we are directly using the names (outside of the assembly they
belong) are in the Tests and a couple of places in our design code. The
raw names are not something that provider writers or app developers
should really need to use.


smitpatel:
:shipit:

PR #3567, opened 2015-10-26
Commands: Re-add cancellation

bricelam:
This was (inadvertently?) removed in #3518

cc @pakrym @lajones


pakrym:
Yes,sorry, that was accidental.

lajones:
:shipit:

PR #3568, opened 2015-10-26
Move caching of property access delegates to the model

ajcvickers:
This is because the dictionary access involved in looking up the delegates is a significant perf hit, and co-location of the accessors with the property/navigation for which they are used removes this overhead.

Since he model is not coupled to the service provider, this means that the delegate construction code is now not a replaceable service. The mitigation for this in the long term is to allow extensibility of the model to contain delegates created in different ways. For example, the factories used to create the delegates could be annotated onto the model.

Note that the delegates are not stored as annotations because the lazy creation needs to happen after the model is "built" and must be thread-safe, neither of which are supported by annotations. Also, this is pert-sensitive code and the lookup needs to be fast.


src/EntityFramework.Core/Metadata/Internal/Property.cs
```diff
@@ -16,6 +17,8 @@ namespace Microsoft.Data.Entity.Metadata.Internal
     [DebuggerDisplay("{DeclaringEntityType.Name,nq}.{Name,nq} ({ClrType?.Name,nq})")]
     public class Property : Annotatable, IMutableProperty
     {
+        private volatile IClrPropertyGetter _getter;
```
anpete: Why not use LazyInitializer?

src/EntityFramework.Core/Metadata/Internal/Navigation.cs
```diff
@@ -15,6 +15,10 @@ namespace Microsoft.Data.Entity.Metadata.Internal
     [DebuggerDisplay("{DeclaringEntityType.Name,nq}.{Name,nq}")]
     public class Navigation : Annotatable, IMutableNavigation
     {
+        private volatile IClrPropertyGetter _getter;
```
anpete: Why not use LazyInitializer?

src/EntityFramework.Core/Metadata/Internal/Property.cs
ajcvickers: What is the advantage?

src/EntityFramework.Core/Metadata/IPropertyBase.cs
```diff
@@ -9,5 +10,7 @@ public interface IPropertyBase : IAnnotatable
     {
         string Name { get; }
         IEntityType DeclaringEntityType { get; }
+        IClrPropertyGetter Getter { get; }
+        IClrPropertySetter Setter { get; }
```
AndriySvyryd: I don't think these should be part of this interface. They are redundant and are really just for perf improvement.
As we've discussed previously a cleaner way of adding perf API is through explicit interfaces. An extention method would try to cast to it and get the value or fallback to calculating the value if the interface isn't implemented.

src/EntityFramework.Core/Metadata/Internal/Property.cs
anpete: 1) Using a built-in (and recommended) lazy-initialization primitive instead of rolling our own.
2) Having a consistent pattern for such initialization -  we use LazyInitializer elsewhere.
3) Making our code easier to understand.

src/EntityFramework.Core/Metadata/IPropertyBase.cs
anpete: Yep, especially because these presumably don't apply to shadow props?

ajcvickers: @AndriySvyryd @anpete Can you guys remind me what the pattern is that will satisfy the following:
- Be fast--i.e. avoid dictionary lookup, etc.
- Be thread safe
- Work even if the implementation of IProperty is not our implementation

AndriySvyryd: Our `Property` would implement this like in this PR.
For other implementations that don't implement the new interface we'll fallback to calculating the value without any caching. This would be slow, but just as functional. And they can easily copy our implementation to get the perf.
@divega Are you ok with this?

divega: As far as I can understand it and if is fast enough, yes... I think an extension method that tries to cast to an optional interface and falls back to the slow implementation is a generally good pattern.

If I am missing something, can we discuss this in person?

src/EntityFramework.Core/Metadata/Internal/Property.cs
ajcvickers: This is what the code is now:

```C#
private volatile IClrPropertyGetter _getter;

public virtual IClrPropertyGetter Getter
    => _getter ?? (_getter = new ClrPropertyGetterFactory().Create(this));
```

This is what I have with LazyInitializer:
```C#
// Warning: Never access this field directly as access needs to be thread-safe
private IClrPropertyGetter _getter;

public virtual IClrPropertyGetter Getter
    => LazyInitializer.EnsureInitialized(ref _getter, () => new ClrPropertyGetterFactory().Create(this));
```
The LazyInitializer version is about 25% slower for the case where it is already initialized. Breaking the inline lambda out into a method does not have any impact on performance.

@divega Which would you prefer?

src/EntityFramework.Core/Metadata/IPropertyBase.cs
ajcvickers: @divega What do you mean by, "if it is fast enough"? Do you mean if the slow implementation is fast enough or if the fast implementation is fast enough? The fast implementation should be almost as fast as the current implementation (additional method call and cast). The slow implementation will probably be between 2-3 orders of magnitude slower.

divega: I mean if the cost of having to cast to the optional interface and any other overhead associated with the implementation of the extension method doesn't dilute the optimization.

divega: And yes, I imagine in this case it will be fine, but the pattern might not be applicable in other situations if the optimization is not as significant.

ajcvickers:
PR updated.

AndriySvyryd:
:shipit:

src/EntityFramework.Core/Metadata/Internal/Property.cs
divega: @ajcvickers Need more information. What is the incidence of this 25% slowness in the overall path you are optimizing with this? Is it the same in 32 and 64 bits? Same across CoreCLR and .NET? 

My main concern with the volatile approach is that I personally don't know if it is safe to use across all architectures (see #960) and without knowing for sure I would stick to not rolling our own. But if you have done the groundwork to be confident that it is the case then I am fine with it. 

Besides that, based on what I have read I am a bit surprised that LazyInitializer is slower. It uses [`Volatile.Read<T>()`](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Threading/Volatile.cs#L189) which is supposed to be able to leverage at runtime an implementation that is optimized for the platform and architecture. My only guess is that at this level of granularity a few levels of method calls might be have a higher incidence on the execution time than the perf of the memory barrier itself.

ajcvickers:
Merged

PR #3569, opened 2015-10-26
doc comments for immutable metadata API

rowanmiller:
Starting with just the core immutable API since there will be a lot of copy paste into the mutable one and the extensions methods... so we'll get this right before I copy paste mistakes everywhere :smile:


src/EntityFramework.Core/Metadata/IProperty.cs
```diff
@@ -1,20 +1,78 @@
...
+        ///     Gets a value indicating whether this property can contain null.
+        /// </summary>
         bool IsNullable { get; }
+
+        /// <summary>
+        ///     Gets a value indicating whether or not this property can be modified before the entity is
+        ///     saved to the database. If true, an exception will be thrown if a value is assigned to
+        ///     this property when it is in the <see cref="EntityState.Added"/> state.
+        /// </summary>
         bool IsReadOnlyBeforeSave { get; }
```
rowanmiller: @ajcvickers can you read over this and IsReadOnlyAfterSave, IsStoreGeneratedAlways, ValueGenerated, and RequiresValueGenerator to make sure I have them accurate.

src/EntityFramework.Core/Metadata/IEntityType.cs
```diff
@@ -8,27 +8,137 @@
 
 namespace Microsoft.Data.Entity.Metadata
 {
+    /// <summary>
+    ///     Represents an entity in a <see cref="Model"/>.
```
AndriySvyryd: IModel

src/EntityFramework.Core/Metadata/DeleteBehavior.cs
```diff
@@ -3,10 +3,24 @@
 
 namespace Microsoft.Data.Entity.Metadata
 {
+    /// <summary>
+    ///     Indicates how a delete operation is applied to dependent entities in a relationship when the principal is deleted.
+    /// </summary>
     public enum DeleteBehavior
     {
+        /// <summary>
+        ///     The delete operation is not applied to dependent entities. The dependent entities remain unchanged.
```
smitpatel: This reads more like no action than restrict.

src/EntityFramework.Core/Metadata/DeleteBehavior.cs
```diff
@@ -3,10 +3,24 @@
 
 namespace Microsoft.Data.Entity.Metadata
 {
+    /// <summary>
+    ///     Indicates how a delete operation is applied to dependent entities in a relationship when the principal is deleted.
```
ajcvickers: Note that this is now not only when the principal is deleted, but also when the relationship to the principal is severed even if the principal is not deleted.

src/EntityFramework.Core/Metadata/IProperty.cs
```diff
@@ -1,20 +1,78 @@
...
+
+        /// <summary>
+        ///     Gets a value indicating whether or not this property can be modified after the entity is
+        ///     saved to the database. If true, an exception will be thrown if a new value is assigned to
+        ///     this property after the entity exists in the database.
+        /// </summary>
         bool IsReadOnlyAfterSave { get; }
```
ajcvickers: IsReadOnly ones sound good. One minor point is that the exception is not thrown until SaveChanges is called, but what you have is pretty clear and I would err on the side of clearness rather than detail.

src/EntityFramework.Core/Metadata/IProperty.cs
ajcvickers: They all look good to me.

ajcvickers:
:shipit:

rowanmiller:
Merged https://github.com/aspnet/EntityFramework/commit/be0d4028d24c5911c2a829c5fa84d06bec43702e

PR #3570, opened 2015-10-27
Undo #3516 due to compiler crash

natemcmaster:
Revert changes made in https://github.com/aspnet/EntityFramework/pull/3516 (type inference for .NET Native).

These changes cause ILC to crash due to OutOfMemoryExceptions.

```
 C:\Program Files (x86)\MSBuild\Microsoft\.NetNative\x86\ilc\IlcInternals.targets(886,5): error : Internal compiler error: Exception of type 'System.OutOfMemoryException' was thrown.
```


natemcmaster:
Simple projects will cause compilation to hit the upper limit on what ILC can infer about our internal generic methods use with user types. I suspect the crash is due to the hundreds of permutations of `.Join<T1,T2,T3,T4>()`. We need a different approach.

ajcvickers:
:sheep::it:

natemcmaster:
Closing in favor of a PR that will be made to the release branch (whenever that happens.)

Issue #2514, opened 2015-07-01
Metadata: Allow indexes and foreign keys to use inherited properties


AndriySvyryd:
The ModelBuilder should lift the FK or index when there is one declared on the same properties on a derived type

rowanmiller:
@AndriySvyryd if you don't think this is critical for 7.0.0 then come discuss as we may postpone it

AndriySvyryd:
7.0.0 and pri0 are appropriate for this

rowanmiller:
@AndriySvyryd can you add some more details on what this means?

AndriySvyryd:
Consider
```
class Base
{
  public int Id { get; set; }
}
class Derived : Base
{
  public string Name { get; set; }
}
```

You should be able to create an index like this:
```
modelBuilder.Entity<Derived>().Index(b => new [] { b.Id, b.Name });
```

Also if you had created an index on the derived entity type:
```
  modelBuilder.Entity<Derived>().Index(b => new [] { b.Id });
```

You should be able to override it on base:
```
  modelBuilder.Entity<Base>().Index(b => new [] { b.Id });
```
and viceversa.
Same applies to FKs.

rowanmiller:
@AndriySvyryd can you discuss with folks what we want to do before doing it. Either design meeting, or just in the team room.

smitpatel:
Partially implementing this for indexes alongwith #700

smitpatel:
for indexes implemented in bc4609bcb34f57325635e3f3c0a1c20606509cd5
Pending for foreignkeys

smitpatel:
Should base type lift the FK from derived type if defined on same features? or block it?

PR #3572, opened 2015-10-27
Allow indexes to use inherited properties

smitpatel:
part of #2514

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/EntityType.cs
@@ -701,7 +701,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                 throw new InvalidOperationException(CoreStrings.DuplicateIndex(Property.Format(properties), DisplayName(), duplicateIndex.DeclaringEntityType.DisplayName()));
             }
 
-            var index = new Index(properties);
+            var index = new Index(properties, this);
             _indexes.Add(properties, index);
 
             return index;
--- a/src/EntityFramework.Core/Metadata/Internal/MetadataHelper.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/MetadataHelper.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Data.Entity.Internal;
 
 namespace Microsoft.Data.Entity.Metadata.Internal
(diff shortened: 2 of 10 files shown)


src/EntityFramework.Core/Metadata/Internal/EntityType.cs
```diff
@@ -735,10 +735,10 @@ public virtual Index AddIndex([NotNull] IReadOnlyList<Property> properties)
             var duplicateIndex = FindIndexesInHierarchy(properties).FirstOrDefault();
             if (duplicateIndex != null)
             {
-                throw new InvalidOperationException(CoreStrings.DuplicateIndex(Property.Format(properties), DisplayName(), duplicateIndex.DeclaringEntityType.DisplayName()));
+                throw new InvalidOperationException(CoreStrings.DuplicateIndex(Property.Format(properties), Name, duplicateIndex.DeclaringEntityType.Name));
```
AndriySvyryd: Why are you changing this?

src/EntityFramework.Core/Metadata/Internal/MetadataHelper.cs
```diff
@@ -26,5 +27,18 @@ public static void CheckSameEntityType(IReadOnlyList<Property> properties, strin
                 }
             }
         }
+
+        public static void CheckPropertiesInEntityType(IReadOnlyList<Property> properties, EntityType entityType, string argumentName)
+        {
+            foreach (var property in properties)
+            {
+                if (property.DeclaringEntityType != entityType
+                    && entityType.FindProperty(property.Name) != property)
```
AndriySvyryd: You can make this check faster:
```
if(!property.DeclaringEntityType.IsAssignableFrom(entityType))
```

test/EntityFramework.Core.Tests/Metadata/Internal/IndexTest.cs
```diff
@@ -41,14 +42,21 @@ public void Can_create_unique_index_from_properties()
         [Fact]
         public void Constructor_check_arguments()
         {
+            // ReSharper disable AssignNullToNotNullAttribute
```
AndriySvyryd: This test can be removed, we don't test argument validation anymore

src/EntityFramework.Core/Metadata/Internal/EntityType.cs
smitpatel: I will revert this for current PR. Filed #3583

AndriySvyryd:
:shipit:

PR #3577, opened 2015-10-27
Add test for foreign key property names for created shadow properties

smitpatel:
Added test for #3503 so that we don't regress in future


lajones:
:shipit:

PR #3580, opened 2015-10-27
Query: Perf: Adds a faster path for constructing primary key values.

anpete:
Some other minor perf improvements to the query hot path.

### DIFF
--- a/src/EntityFramework.Core/Query/Internal/QueryBuffer.cs
+++ b/src/EntityFramework.Core/Query/Internal/QueryBuffer.cs
@@ -47,13 +47,9 @@ namespace Microsoft.Data.Entity.Query.Internal
         }
 
         public virtual object GetEntity(
-            IEntityType entityType,
-            IKeyValue keyValue,
-            EntityLoadInfo entityLoadInfo,
-            bool queryStateManager)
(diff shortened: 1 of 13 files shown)


src/EntityFramework.Core/Query/Internal/QueryBuffer.cs
```diff
@@ -269,34 +252,27 @@ var keyValueFactory
                 navigationPath,
                 currentNavigationIndex,
                 await relatedEntitiesLoaders[currentNavigationIndex](primaryKeyValue, relatedKeyFactory)
-                .Select(async (eli, ct) =>
-                    {
-                        var keyValue
-                            = keyValueFactory
-                                .Create(keyProperties, eli.ValueBuffer);
+                    .Select(async (eli, ct) =>
```
AndriySvyryd: if this is a hot path don't use async keyword and don't use a lambda

AndriySvyryd:
:shipit:

src/EntityFramework.Core/Query/Internal/QueryBuffer.cs
anpete: Not looking at this path...yet :-)

PR #3581, opened 2015-10-27
Convert IEntityType.GetNavigations() and related methods to extention methods

AndriySvyryd:
The same data can be accessed through IForeignKey
The implementation will still keep track of the navigations for perf reasons

### DIFF
--- a/src/EntityFramework.Core/EntityFramework.Core.csproj
+++ b/src/EntityFramework.Core/EntityFramework.Core.csproj
@@ -86,6 +86,7 @@
     <Compile Include="Internal\IDatabaseProviderSelector.cs" />
     <Compile Include="Internal\IndentedStringBuilder.cs" />
     <Compile Include="Internal\LazyRef.cs" />
+    <Compile Include="Metadata\Internal\ICanGetNavigations.cs" />
     <Compile Include="Metadata\Internal\INavigationAccessors.cs" />
     <Compile Include="Metadata\Internal\IPropertyBaseAccessors.cs" />
     <Compile Include="Metadata\Internal\KeyExtensions.cs" />
--- a/src/EntityFramework.Core/Extensions/EntityTypeExtensions.cs
+++ b/src/EntityFramework.Core/Extensions/EntityTypeExtensions.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Data.Entity
             return entityType.FindForeignKey(new[] { property }, principalKey, principalEntityType);
         }
 
-        public static IEnumerable<IForeignKey> FindReferencingForeignKeys([NotNull] this IEntityType entityType)
+        public static IEnumerable<IForeignKey> GetReferencingForeignKeys([NotNull] this IEntityType entityType)
         {
             Check.NotNull(entityType, nameof(entityType));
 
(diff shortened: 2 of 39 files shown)


src/EntityFramework.Core/Extensions/EntityTypeExtensions.cs
```diff
@@ -102,6 +102,19 @@ public static INavigation FindNavigation([NotNull] this IEntityType entityType,
...
+        public static INavigation FindNavigation([NotNull] this IEntityType entityType, [NotNull] string name)
+        {
+            Check.NotNull(entityType, nameof(entityType));
+            Check.NotNull(name, nameof(name));
+
+            return entityType.GetNavigations().FirstOrDefault(n => n.Name.Equals(name));
+        }
+
+        public static IEnumerable<INavigation> GetNavigations([NotNull] this IEntityType entityType)
+            => Check.NotNull(entityType, nameof(entityType)).BaseType?.GetDeclaredNavigations()
```
smitpatel: Shouldn't this be `BaseType?.GetNavigations()` ?

AndriySvyryd: Yes! Will add a test.

src/EntityFramework.Core/EntityFramework.Core.csproj
```diff
@@ -86,6 +86,7 @@
     <Compile Include="Internal\IDatabaseProviderSelector.cs" />
     <Compile Include="Internal\IndentedStringBuilder.cs" />
     <Compile Include="Internal\LazyRef.cs" />
+    <Compile Include="Metadata\Internal\ICanGetNavigations.cs" />
```
smitpatel: Name :smile:

smitpatel:
LGTM

src/EntityFramework.Core/Metadata/Internal/ForeignKeyExtensions.cs
```diff
@@ -157,23 +120,30 @@ public static INavigation FindNavigationToInHierarchy([NotNull] this IForeignKey
...
 
-            if (foreignKey.IsIntraHierarchical())
+            return foreignKey.IsIntraHierarchical()
+                ? foreignKey.GetNavigations()
+                : foreignKey.FindNavigations(foreignKey.PrincipalEntityType.IsAssignableFrom(entityType));
+        }
+
+        private static IEnumerable<INavigation> FindNavigations(
+            this IForeignKey foreignKey, bool toPrincipal)
+        {
```
smitpatel: Question: Why this is `IEnumerable` with yield when there can be only 1 navigation in each direction for given foreign key?

src/EntityFramework.Core/EntityFramework.Core.csproj
AndriySvyryd: I am open to suggestions

src/EntityFramework.Core/Metadata/Internal/ForeignKeyExtensions.cs
AndriySvyryd: Because here I want an empty IEnumerable if there's no navigation instead of one with a null

smitpatel:
:shipit:

Issue #3394, opened 2015-10-09
RevEng: Split Relationship Fluent API Line in OnModelCreating()


lajones:
Currently for RevEng in `OnModelCreating()` we generate the fluent API configuring a relationship all on one line e.g.:

```
            modelBuilder.Entity<OneToOneFKToUniqueKeyDependent>(entity =>
            {
                entity.HasOne(d => d.OneToOneFKToUniqueKeyDependentFK).WithOne(p => p.OneToOneFKToUniqueKeyDependent).HasPrincipalKey<OneToOneFKToUniqueKeyPrincipal>(p => new { p.OneToOneFKToUniqueKeyPrincipalUniqueKey1, p.OneToOneFKToUniqueKeyPrincipalUniqueKey2 }).HasForeignKey<OneToOneFKToUniqueKeyDependent>(d => new { d.OneToOneFKToUniqueKeyDependentFK1, d.OneToOneFKToUniqueKeyDependentFK2 });
            }
```

This can get pretty long. Consider doing the following instead:

```
            modelBuilder.Entity<OneToOneFKToUniqueKeyDependent>(entity =>
            {
                entity.HasOne(d => d.OneToOneFKToUniqueKeyDependentFK)
                    .WithOne(p => p.OneToOneFKToUniqueKeyDependent)
                    .HasPrincipalKey<OneToOneFKToUniqueKeyPrincipal>(p => new { p.OneToOneFKToUniqueKeyPrincipalUniqueKey1, p.OneToOneFKToUniqueKeyPrincipalUniqueKey2 })
                    .HasForeignKey<OneToOneFKToUniqueKeyDependent>(d => new { d.OneToOneFKToUniqueKeyDependentFK1, d.OneToOneFKToUniqueKeyDependentFK2 });
            }
```

divega:
Minor edit on the title to make it a bit stronger :smile:

lajones:
Note: HasIndex() method chain should also be considered for this - see #3579.

lajones:
Split the relationship config lines with PR #3588.

lajones:
Last part checked in with commit d0f53e6.

PR #3588, opened 2015-10-28
RevEng: Fix the way relationship config lines appear in InModelCreated()

lajones:
Partial fix for #3394.


natemcmaster:
What remains to finish #3394?

:shipit:

lajones:
@natemcmaster Thanks. What's missing is the splitting of HasIndex() lines we discussed recently. Want to do that separately.

lajones:
Checked in with 4414581.

PR #3589, opened 2015-10-28
DNX Commands: Add --targetProject option

### LINKED ISSUE #2679: DNX Commands: Make ef a global command (enables --starupProject)

bricelam:
Resolves #2679 (for now)

### DIFF
--- a/src/EntityFramework.Commands/Commands/Program.cs
+++ b/src/EntityFramework.Commands/Commands/Program.cs
@@ -69,6 +69,9 @@ namespace Microsoft.Data.Entity.Commands
                             var context = update.Option(
                                 "-c|--context <context>",
                                 "The DbContext to use. If omitted, the default DbContext is used");
+                            var targetProject = update.Option(
+                                "-p|--targetProject <project>",
+                                "The project with the Migration classes. If omitted, the current project is used.");
                             var environment = update.Option(
                                 "-e|--environment <environment>",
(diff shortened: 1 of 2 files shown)


src/EntityFramework.Commands/Commands/Program.cs
```diff
@@ -67,6 +67,9 @@ public static int Main([NotNull] string[] args)
                             var context = update.Option(
                                 "-c|--context <context>",
                                 "The DbContext to use. If omitted, the default DbContext is used");
+                            var targetProject = update.Option(
+                                "-p|--targetProject <project>",
+                                "The project with the Migration classes. If omitted, the current projet is used.");
```
lajones: Typo: projet -> project. And same in a few more places below.

lajones:
:shipit:

src/EntityFramework.Commands/Commands/Program.cs
bricelam: Doh. :smile:

davidfowl: Not unixy enough.

`--target-project`

bricelam: I'll send a PR to kabob-case all the things.

PR #3591, opened 2015-10-28
Improvements to query on netnative

natemcmaster:
Fixes a bug on UWP that causes the ILC compiler to crash with `OutOfMemoryException`

Removed changed generics from netnative type inference. 
Add more support for async queries.


src/EntityFramework.Core/Properties/EntityFramework.Core.rd.xml
```diff
@@ -18,5 +24,44 @@
       </Type>
     </Assembly>
 
+    <Type Name="System.Linq.AsyncEnumerable">
+      <MethodInstantiation Name="Any" Arguments="Microsoft.Data.Entity.Storage.ValueBuffer" Dynamic="Required" />
```
lajones: Just checking: did you mean "Required" here? Elsewhere you use "Required All"

natemcmaster: Intentional. "Required All" is not a valid option for MethodInstantiation

lajones: OK. As far as I'm concerned :shipit: - but you should probably get at least one more pair of eyes to look at this.

PR #3592, opened 2015-10-28
Core: Combine SimpleKeyValue and CompositeKeyValue into a single struct.

anpete:
This allows for an alloc free key access pattern in runtime generated code. I.e. query. (TBD)

cc @ajcvickers


ajcvickers:
:shipit:

PR #3600, opened 2015-10-29
Fix #2891 - Promote transaction API to core

### LINKED ISSUE #2891: Promote transaction API to core (& no-op implementation for in-memory)

mikary:
Opening PR to replace #3494 for including in RC1


mikary:
Not approved for RC1, merged in #3494 instead.

PR #3601, opened 2015-10-29
DNX Commands: Kabob-case all the things

bricelam:
@davidfowl said it's more idiomatic...


divega:
I personally like it better like this but I would like to hear what others think.

anpete:
Looks better to me.

smitpatel:
:+1:

lajones:
:+1:

natemcmaster:
:+1: I think that's enough team votes to :shipit:

PR #3605, opened 2015-10-29
Avoid checking fixup in state manager when query is known to have done it

ajcvickers:
Query always fixes up navigations between entities returned by a given query, regardless of whether the query is going to be tracked or not. If the state manager ends up tracking only these entities, then it does not need to do any fixup work. Not doing this work can improve query tracking performance by about 22% for this common case.

The state manager switches out of this mode when:
- A second query starts
- An entity is attached or added
- An entity changes to the Modified or Added state, since this could be an FK change


anpete:
Looks great!

What do you think about this potential simplification: Rather than keep track of whether it's the first query via flowing EntityTrackingSource into StartTracking, instead we send a single "tracking query is starting" notification to the StateManager. There is a natural place to do this inside ILinqOperatorProvider.TrackEntities.

ajcvickers:
@anpete I'll give that a try

ajcvickers:
Closing this. Will open a new PR with the suggested change.

PR #3607, opened 2015-10-30
Add ConventionalAnnotation to store ConfigurationSource in instead of InternalMetadataBuilder

AndriySvyryd:
Part of #3476


ajcvickers:
:shipit:

PR #3612, opened 2015-10-30
[rc1] Upgrade re-linq

natemcmaster:
The upgrade to re-linq includes patch for UWP support. (see https://github.com/re-motion/Relinq/pull/6)


ajcvickers:
:shipit:

PR #3614, opened 2015-10-30
:memo: doc comments for ValueGeneration.*

rowanmiller:


src/EntityFramework.Core/ValueGeneration/IValueGeneratorCache.cs
```diff
@@ -7,8 +7,29 @@
...
+    /// </summary>
     public interface IValueGeneratorCache
     {
+        /// <summary>
+        ///     Gets the existing value generator from the cache, or creates a new one if one is not present in
+        ///     the cache.
+        /// </summary>
+        /// <param name="property"> The property to get the value generator for. </param>
+        /// <param name="entityType"> 
+        ///     The entity type that the value generator will be used for. This may be different from the type
```
natemcmaster: I think the second clause here could be simplified to say `When called on inherited properties or derived entity types, this entity type may be different from the declared entity type on <paramref name="property" />`

natemcmaster:
:shipit:

PR #3628, opened 2015-11-02
Query: Perf: Remove closures and dynamic delegates from compiled quer…

anpete:
…y trees.

- Replaces lambda shapers with new IShaper<T> abstraction.


test/EntityFramework.MicrosoftSqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
```diff
@@ -35,35 +35,7 @@ var customers
...
-        querySource: from Customer <generated>_0 in value(Internal.EntityQueryable`1[FunctionalTests.TestModels.Northwind.Customer]), 
-        valueBuffer: prm1, 
-        valueBufferOffset: 0, 
-        queryContext: prm0, 
-        entityType: FunctionalTests.TestModels.Northwind.Customer, 
-        trackingQuery: True, 
-        keyValueFactory: SimpleKeyValueFactory`1, 
-        materializer: (ValueBuffer prm2) => 
-        {
-            var var3
```
maumar: bit of a bummer that this gets compiled in

maumar:
:shipit:

test/EntityFramework.MicrosoftSqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
anpete: Yeah, but it does mean less to compile. If we want we can keep a string representation of the materializer ET around for this.

Issue #3633, opened 2015-11-02
Query :: QuerySourceTracingExpressionVisitor doesn't prune EF.Property calls when looking for query sources 


maumar:
repro:

```
using System;
using System.Collections.Generic;
using System.Data.SqlClient;
using System.Linq;
using Microsoft.Data.Entity;

namespace Repro3633
{
    public class A
    {
        public int Id { get; set; }

        public int CId { get; set; }
        public virtual C C { get; set; }
    }

    public class C
    {
        public int Id { get; set; }

        public virtual ICollection<A> A { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<A> As { get; set; }
        public DbSet<C> Cs { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(new SqlConnectionStringBuilder { DataSource = ".", InitialCatalog = "Repro3633", MultipleActiveResultSets = true, IntegratedSecurity = true }.ToString());
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<A>().HasOne(p => p.C).WithMany(p => p.A).HasForeignKey(p => p.CId);
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                C c = new C();
                ctx.Cs.Add(c);

                A a = new A();
                a.C = c;
                a.CId = c.Id;
                ctx.As.Add(a);

                ctx.SaveChanges();
            }

            using (var ctx = new MyContext())
            {
                var query = from a in ctx.As.Include(a => a.C)
                            select new 
                            {
                                C = new 
                                {
                                    Id = a.C.Id
                                }
                            };

                var result = query.ToList();
            }
        }
    }
}

```

throws:

Unhandled Exception: System.ArgumentException: The given expression 'new DtoA() {C = new DtoC() {Id = Property([a], "CId")}}' does not contain the searched expression '[a]' in a nested NewExpression with member assignments or a MemberBindingExpression.
Parameter name: fullExpression
   at Remotion.Linq.Clauses.ExpressionVisitors.AccessorFindingExpressionVisitor.FindAccessorLambda(Expression searchedExpression, Expression fullExpression, ParameterExpression inputParameter)
[...]

PR #3635, opened 2015-11-03
[SMALL] Fix to #3633 - Query :: QuerySourceTracingExpressionVisitor doesn't prune EF.Property calls when looking for query sources

maumar:
QuerySourceTracingExpressionVisitor is looking for query sources that are being materialized, so that we know which includes need to be applied to the query. The logic filters out references that are not being projected (e.g. property access), however we were not pruning property access done via EF.Property method.

Fix is to add the missing overload.


maumar:
@anpete

anpete:
:shipit:

divega:
@maumar is there a reason this is not in?

anpete:
ping @maumar

maumar:
I remember this fix uncovered a different bug, not sure if its fixed - will double check

natemcmaster:
Confirmed. This introduced a bug into GroupJoin. I'm working on a different fix.

PR #3641, opened 2015-11-03
Improve perf of relationship snapshot

ajcvickers:
This is done by calculating indexes for the properties that need to be snapshotted and using an array sidecar to store the snapshot.

Perf improvement is about 15%, with also about 10% improvement in memory usage.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/RelationshipsSnapshot.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/RelationshipsSnapshot.cs
@@ -1,33 +1,28 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Metadata;
-using Microsoft.Data.Entity.Utilities;
+using Microsoft.Data.Entity.Metadata.Internal;
 
 namespace Microsoft.Data.Entity.ChangeTracking.Internal
 {
-    // TODO: Consider using ArraySidecar with pre-defined indexes
-    // Issue #741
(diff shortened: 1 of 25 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/RelationshipsSnapshot.cs
```diff
@@ -1,33 +1,28 @@
...
-        {
-            var entityType = entry.EntityType;
+        protected override int Index(IPropertyBase property) => property.GetRelationshipIndex();
 
-            return entityType.GetKeys().SelectMany(k => k.Properties)
-                .Concat(entityType.GetForeignKeys().SelectMany(fk => fk.Properties))
-                .Distinct()
-                .Concat<IPropertyBase>(entityType.GetNavigations());
+        protected override void ThrowInvalidIndexException(IPropertyBase property)
+        {
```
anpete: Should property be used here?

src/EntityFramework.Core/Metadata/Internal/PropertyIndexes.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Data.Entity.Metadata.Internal
+{
+    public class PropertyIndexes
```
anpete: Consider a struct

anpete:
:shipit:

src/EntityFramework.Core/ChangeTracking/Internal/RelationshipsSnapshot.cs
ajcvickers: This should never get called now, outside of an application error. I'm planning to remove in another checkin.

src/EntityFramework.Core/Metadata/Internal/PropertyIndexes.cs
ajcvickers: I don't think it's going to help making it a struct, but I'll look into it.

Issue #3464, opened 2015-10-17
StringCompareExpression throws 100% of the times


thatotherone:
This is related to #1767
Discovered in EntityFramework.Relational 7.0.0-beta8, but is present in current master as far as I can tell.

When `string.Compare()` is used in the `Where` clause of relational DbSet it ends up throwing:

```
error   : [Microsoft.AspNet.Diagnostics.ExceptionHandlerMiddleware] An unhandled exception has occurred: Extension node must override the property Expression.Type.
System.InvalidOperationException: Extension node must override the property Expression.Type.
   at System.Linq.Expressions.Expression.get_Type()
   at System.Linq.Expressions.Expression.AndAlso(Expression left, Expression right, MethodInfo method)
   at Microsoft.Data.Entity.Query.RelationalQueryModelVisitor.VisitWhereClause(WhereClause whereClause, QueryModel queryModel, Int32 index)
   at Remotion.Linq.QueryModelVisitorBase.VisitBodyClauses(ObservableCollection`1 bodyClauses, QueryModel queryModel)
   at Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)
   at Microsoft.Data.Entity.Query.EntityQueryModelVisitor.VisitQueryModel(QueryModel queryModel)
   at Microsoft.Data.Entity.Query.RelationalQueryModelVisitor.VisitQueryModel(QueryModel queryModel)
   at Microsoft.Data.Entity.Query.Internal.SqlServerQueryModelVisitor.VisitQueryModel(QueryModel queryModel)
   at Microsoft.Data.Entity.Query.EntityQueryModelVisitor.CreateQueryExecutor[TResult](QueryModel queryModel)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Data.Entity.Query.QueryCompiler.<>c__DisplayClass16_0`1.<CompileQuery>b__0()
   at Microsoft.Data.Entity.Query.Internal.CompiledQueryCache.GetOrAddQuery[TResult](Object cacheKey, Func`1 compiler)
   at Microsoft.Data.Entity.Query.QueryCompiler.Execute[TResult](Expression query)
```

It appears that StringCompareExpression doesn't override its return type. Fix is simple:

```
[...]

PR #3647, opened 2015-11-03
[TINY] Fix to #3464 - StringCompareExpression throws 100% of the times

maumar:
StringCompareExpression was missing Type property overload.


maumar:
@anpete

anpete:
:shipit:

PR #3649, opened 2015-11-04
Strong name everything.

cesarbs:
https://github.com/aspnet/dnx/issues/2615

Pushing here to see how the test run goes. I'm getting local test failures after strong naming. If they happen here too, I'll file an issue.

cc @davidfowl @Eilon @muratg @moozzyk @natemcmaster @bricelam


cesarbs:
Now I see that we don't have AppVeyor and Travis for this repo. I'll enter an issue for the test failure I'm seeing.

cesarbs:
https://github.com/aspnet/EntityFramework/issues/3650

Issue #700, opened 2014-09-25
Create explicit index objects for every index


ajcvickers:
Currently some indexes are inferred from keys by Migrations when using relational models. Per [design meeting notes](https://github.com/aspnet/EntityFramework/wiki/Entity-Framework-Design-Meeting-Notes---September-11,-2014#metadata-for-keys-foreign-keys-and-indexes) every index should be created explicitly in the model.

bricelam:
We need to be careful to avoid creating duplicate indexes in the database. Most databases create an implicit index for PRIMARY KEY and UNIQUE constraints. If you then explicitly create an INDEX on the same columns, a new, duplicate index will be maintained by the database.

bricelam:
Marking for re-triage; this should probably be `pri0`. Without this, Migrations won't create unique indexes for 0..1:0..1 relationships effectively making them 0..1:* in the database.

Here are the main things that would change in Metadata:

Before | After
------ | -----
Key.IsClustered | Key.Index.IsClustered
ForeignKey.IsUnique | ForeignKey.Index.IsUnique

Migrations would then promote the indices' configuration to the `CREATE..PRIMARY KEY` AND `CREATE..UNIQUE` clauses and skip creating indexes associated with keys.

For foreign keys, everything should "just work" in Migrations since there is typically no implicit index on `FOREIGN KEY` constraints.

smitpatel:
Blocked by #2514

smitpatel:
@bricelam - Adding index for FK introduces extra migration operation for index creation which causes MigrationsModelDifferTests to fail. Is that expected behavior?

bricelam:
Yes, the tests will need to be updated.

smitpatel:
blocked by #3648 
GraphUpdate tests are failing.

smitpatel:
> Decision:
We’ll create indices by convention for each foreign key. The user owns this index and is free to update or delete it. The index will not be used when reasoning about foreign keys or relationships.
We won’t create indices for keys. Keys will continue to mirror some of the metadata that can also be defined on index (e.g. SqlServerIsClustered) in both the core metadata (as annotations) and the Fluent API. If an index is created over the same columns as a key, both a primary key/unique constraint and an index will be created in the database (i.e. we won’t collapse them in Migrations).
-Brice


Implemented in #3764

PR #3657, opened 2015-11-05
Update topological sort to account for unique constraints on FK

### LINKED ISSUE #3648: Topological Sort of CommandBatch gives incorrect ordering
When we assign new dependent entity to principal end navigation for required fk, the old dependent entity gets deleted and new entity is added to database. Though commands are sorted in a way that we add new entity first and then delete the old entity. For one-to-one FK the fk property should be unique. Therefore the order in which sql sent to database throws exception. This is not happening at present because of #700 

Following is the test failure out codebase. Trying to generate standalone repro if possible.

Test
`GraphUpdatesWithSequenceSqlServerTest.Save_required_non_PK_one_to_one_changed_by_reference(changeMechanism = ChangeMechanism.Principal, useExistingEntities = false)`
Sql sent to database
```` SQL
INSERT INTO [RequiredNonPkSingle1] ([Id], [RootId])
VALUES (@p0, @p1);
SELECT @@ROWCOUNT;
DELETE FROM [RequiredNonPkSingle2]
WHERE [Id] = @p2;
SELECT @@ROWCOUNT;
INSERT INTO [Root] ([Id], [AlternateId])
VALUES (@p3, @p4);
SELECT @@ROWCOUNT;
',N'@p0 int,@p1 int,@p2 int,@p3 int,@p4 uniqueidentifier',@p0=181,@p1=1,@p2=31,@p3=180,@p4='D52E97D8-367E-4EDE-1CD5-08D2E4AE65A2'

DELETE FROM [RequiredNonPkSingle1]
WHERE [Id] = @p5;
SELECT @@ROWCOUNT;
INSERT INTO [RequiredNonPkSingle2] ([Id], [BackId])
VALUES (@p6, @p7);
SELECT @@ROWCOUNT;
',N'@p5 int,@p6 int,@p7 int',@p5=30,@p6=182,@p7=181
````

smitpatel:
resolves #3648 
We sort commands in same batch based on EntityState, which puts delete before add. But if the due to other operations, if add/delete on same entity ends up in different batches then add might be performed before delete. For foreign key which are unique this causes exception on database.
Solution is to add foreign key dependency between commands for unique foreign key if the same values are being re-used. So that delete always happen before add/update.
At present this error is not being seen because of #700 because all `0..1:0..1` relationships are stored as `0..1:*` in database.

### DIFF
--- a/src/EntityFramework.Relational/Update/Internal/CommandBatchPreparer.cs
+++ b/src/EntityFramework.Relational/Update/Internal/CommandBatchPreparer.cs
@@ -78,10 +78,13 @@ namespace Microsoft.Data.Entity.Update.Internal
         // according to these rules:
         //
         // 1. Commands adding rows or modifying the candidate key values (when supported) must precede
-        //     commands modifying or adding rows that will be referencing the former
+        //     commands adding or modifying rows that will be referencing the former
         // 2. Commands deleting rows or modifying the foreign key values must precede
         //     commands deleting rows or modifying the candidate key values (when supported) of rows
         //     that are currently being referenced by the former
+        // 3. Commands deleting rows or modifying the foreign key values must precede
+        //     commands adding or modifying the foreign key values to the same values
+        //     if foreign key is unique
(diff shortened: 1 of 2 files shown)


AndriySvyryd:
:shipit:

PR #3669, opened 2015-11-05
Move InternalEntityTypeBuilder state to EntityType

AndriySvyryd:
Part of #3476


smitpatel:
There would be a lot of other places where `EntityType.Builder` would replace creating entity type builder from model builder especially in `InternalRelationshipBuilder` otherwise looks fine.
:+1: for having access to builder from metadata.
:shipit:

AndriySvyryd:
I will optimize the calling code in one go once I've moved all the state out of the builders.

Issue #2752, opened 2015-07-30
Review exception messages to ensure we don't show any instance data


rowanmiller:

mikary:
- [ ] Review Strings.resx
 - [ ] Commands
 - [ ] Core
 - [ ] InMemory
 - [ ] Microsoft.SqlServer
 - [ ] Microsoft.SqlServer.Design
 - [ ] Relational
 - [ ] Relational.Design
 - [ ] Sqlite
 - [ ] Sqlite.Design
- [ ] Locate raw string messages

mikary:
Exceptions requiring modification
  - ExpressionParameterizationException - passes expression tree that may contain parameter instance data

mikary:
Error messages not in resx file
  - [x] ExpressionPrinter.VisitUnary
  - [x] ExpressionPrinter.UnhandledExpressionType
  - [x] ExpressionPrinter.UnhandledOperation
  - [x] ModificationCommand.AddEntry
  - [x] CommandLogger.Log

mikary:
cc @blowdart here is the issue for reviewing exception messages that I mentioned the other day.

mikary:
Reviewed and removed instance data and connection strings from error messages.

Note: Instance data is referenced in `DbUpdateException` exception object, but not included in the message. Consuming code is expected to intercept these exceptions if the data is not to be accessible.

PR #3684, opened 2015-11-06
Remove unused exception and log strings from resource files

mikary:
Discovered while reviewing messages for #2752


anpete:
:shipit:

Issue #2646, opened 2015-07-16
Reverse Engineer: UWP Projects


bricelam:
The big issue we have to sort out here is how to get Razor into the project to use at design-time. One option is to redistribute it (and any of its dependencies that also won't install on `uwp10` projects) as part of the `EntityFramework7.Relational.Design` package.

rowanmiller:
Discussed and concluded we'll only do this if folks ask for it. Targeting an existing database from UWP seems like a less common scenario (at least until we support targeting non-local storage).

ErikEJ:
An opportunity for a third party Tool, perhaps?

rowanmiller:
@ErikEJ could be, though I'm not sure how often you have an existing local SQLite database that you want to start accessing with EF.

ErikEJ:
Yes, but I have tools already in place that lets you migrate from SQL Server to sqlite, so a possible RAD workflow if you want to resuse your SQL Server based schema could be:
SQL Server => SQLite db => EF7 codegen

bricelam:
Re-opening for triage. After #3687, we're really close to having this. The rest of the work should be relatively cheap. If nothing else, this could be Backlog, up-for-grabs.

bricelam:
This "works" (as much as it can). It could be made better by #6128, but I don't think there's anything else to do here.

bricelam:
The workaround for when it doesn't work is to reverse engineer into a console app and copy the files to the UWP one.

PR #3687, opened 2015-11-06
Move scaffolding code generation into EntityFramework.Commands assembly

### LINKED ISSUE #3490: Scaffolding: move implementation of code gen into Commands project
Will require splicing some classes (CSharpUtilities and ModelUtilities) so we don't have circular dependencies.

natemcmaster:
Fix #3490.

 - Move E2E scaffolding testing into EF.Commands.FunctionalTests
 - Remove the now empty EF.Relation.Design.FunctionalTests assembly/project/package
 - Remove unused strings from RelationalDesignStrings
 - Split some internal classes to divide code-gen functionality


natemcmaster:
cc @bricelam @lajones

ErikEJ:
Pls remember to publish the new Commands.FunctionalTest package on Myget and NuGet feeds

bricelam:
We can probably simplify the packages a bit more. I'll take a closer look on Monday.

bricelam:
I don't like the idea of providers needing to use EF.Commands.FunctionalTests. It feels like we could refactor our provider-agnostic tests better to remove the dependency on EF.Commands.

natemcmaster:
Our E2E tests right now don't follow a consistent pattern. It would be nice to help providers write better E2E tests, but we need to cleanup our test infrastructure ourselves. Ideally, we could have a general E2E testing on commands that we can re-use for migrations, reveng, update, etc.

bricelam:
Product changes look good. We're one step closer to #2646 :wink:

bricelam:
Discussed in-person: We don't want to "ship" EF.Commands.FunctionalTests. We'll re-introduce EF.Relational.Design.FunctionalTests and make it depend on EF.Commands.

lajones:
As far as I'm concerned :shipit:. But maybe check with @bricelam as well.

natemcmaster:
@bricelam  I updated to reverse changes you made in https://github.com/aspnet/EntityFramework/pull/3686. I just double checked and this does not regress on https://github.com/aspnet/EntityFramework/issues/3681.

bricelam:
:shipit:

PR #3688, opened 2015-11-06
[TINY] Some fixes to expression printer around parameters

maumar:
remove the workaround re-used parameters, since we no longer do this. Also avoid throwing exception if the parameter is not there (sometimes would happen if trying to print expression fragment, e.g. lambda body)

### DIFF
--- a/src/EntityFramework.Core/Query/Internal/ExpressionPrinter.cs
+++ b/src/EntityFramework.Core/Query/Internal/ExpressionPrinter.cs
@@ -264,25 +264,12 @@ namespace Microsoft.Data.Entity.Query.Internal
         {
             _stringBuilder.Append("(");
 
-            foreach (var prm in node.Parameters)
+            foreach (var parameter in node.Parameters)
             {
-                string prmName = null;
+                _parametersInScope.Add(parameter, parameter.Name);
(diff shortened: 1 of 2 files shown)


maumar:
@anpete

src/EntityFramework.Core/Query/Internal/ExpressionPrinter.cs
```diff
@@ -266,20 +266,8 @@ protected override Expression VisitLambda<T>(Expression<T> node)
...
-                {
-                    prmName = _parametersInScope[prm];
-                }
-                else
-                {
-                    prmName = "prm" + _parametersInScope.Count;
-                    _parametersInScope.Add(prm, prmName);
-                }
-
+                var prmName = "prm" + _parametersInScope.Count;
```
anpete: We need to use the actual parameter name

maumar:
@anpete  updated

src/EntityFramework.Core/Query/Internal/ExpressionPrinter.cs
```diff
@@ -266,21 +266,8 @@ protected override Expression VisitLambda<T>(Expression<T> node)
 
             foreach (var prm in node.Parameters)
```
anpete: prm -> parameter

anpete:
:shipit:

Issue #3594, opened 2015-10-28
Fix the build on non-Windows platforms


cesarbs:
Related to https://github.com/aspnet/Universe/issues/304. EntityFramework is currently disabled because we're not able to clone and build it on Linux/OS X to then run tests on CoreCLR.

natemcmaster:
Multiple issues

1. `mono nuget.exe` dies on 4.2.0 (necessary for build.sh)
2. We need to build against Mono 4.2.0 or newer
3. `dnu build` dies on projects with `net46`

natemcmaster:
Blocked on https://github.com/dotnet/corefx/issues/4273

natemcmaster:
Blocked on Mono 4.2.1 bug https://bugzilla.xamarin.com/show_bug.cgi?id=35906

muratg:
Does this work against CoreCLR?  I'm more worried about CoreCLR compat (although Mono would be also nice to have.)

natemcmaster:
I wrote a script for our Travis CI [see travis-build.sh](https://github.com/natemcmaster/EntityFramework/blob/linux-tests/tools/travis-build.sh) that builds and runs only on CoreCLR on Linux. Most of our unit testing appears to work fine. Travis doesn't run SQL Server e2e testing though. I'm working on getting a Linux agent that will test SQL Server connectivity using CoreCLR automatically. It's been so broken up to this point that I've just been manual testing it.

natemcmaster:
All tests now run x-plat on CoreCLR (see #3781). The project will build against Mono, but testing will never fully work unless Mono updates their implementation of SqlClient.

natemcmaster:
Blocked on https://github.com/xunit/xunit/issues/564. Xunit always hangs even if parallelization is turned off.

divega:
@natemcmaster FYI I have updated your check list to include enabling tests in CI on OSX and separated the Mono-specific issues which I think aren't critical to fix.

natemcmaster:
Mostly fixed with fdecc52feb91b42ea7159de5a461526bf589c9a3 which skips failing tests on version of Mono <4.2.0. (Mostly b/c there are some Xunit issues with 4.0, but these are external to EF)

divega:
@natemcmaster Unless it is already completed we need an active bug to track our part of the work for 

- [ ] Need to  Need to enable tests on CoreCLR in CI under OSX (waiting for aspnet/Universe#304)

natemcmaster:
[...]

PR #3689, opened 2015-11-07
[rc1] Temporarily add reference System.Net.Security 

natemcmaster:
Workaround for unresolved issues in dnu restore on linux. (See https://github.com/dotnet/corefx/issues/4230).

As long as users don't use MARS, SqlClient on Linux/OSX appears to be *mostly ok.


bricelam:
:shipit:

natemcmaster:
After some discussion, we decided not to merge this. External changes should fix this for the next release. Enabling Linux testing (see #3594) should help us catch these problems sooner.

natemcmaster:
As a workaround, users should include this in their project.json dependencies:

```json
"Microsoft.NETCore.Platforms": "1.0.1-beta-*"
```

PR #3690, opened 2015-11-08
Query: Perf: Introduces RelationalCommand cache for query.

anpete:
- Each compiled query keeps a cache of generated RelationalCommands keyed by parameter nullability (or structural equality for composite parameter values).
- Adds NotParameterizedAttribute to core. This can be applied to parameters of custom LINQ operators to prevent parameterization. Used by FromSql.
- Flow parameter value dictionaries as IReadOnlyDictionary.
- Rename CommandBuilder to ShaperCommandContext to separate from the language around relational command construction.


src/EntityFramework.Core/Query/NotParameterizedAttribute.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Data.Entity.Query
+{
+    /// <summary>
+    ///     Signals that custom LINQ operator parameter should not be parameterized during query compilation.
+    /// </summary>
+    public sealed class NotParameterizedAttribute : Attribute
```
divega: Add AttributeUsage for the cases we honor? (I think it is just method).

anpete: Will do.

src/EntityFramework.MicrosoftSqlServer/Query/Sql/Internal/SqlServerQuerySqlGeneratorFactory.cs
```diff
@@ -14,40 +13,38 @@ public class SqlServerQuerySqlGeneratorFactory : ISqlQueryGeneratorFactory
...
-            Check.NotNull(sqlCommandBuilder, nameof(sqlCommandBuilder));
 
             _commandBuilderFactory = commandBuilderFactory;
             _sqlGenerator = sqlGenerator;
             _parameterNameGeneratorFactory = parameterNameGeneratorFactory;
-            _sqlCommandBuilder = sqlCommandBuilder;
         }
 
-        public virtual ISqlQueryGenerator CreateGenerator(SelectExpression selectExpression)
+        public virtual IQuerySqlGenerator CreateSqlQueryGenerator(SelectExpression selectExpression)
```
divega: So there is IQuerySqlGenerator and IQuerySqlGenerator? :confused:

divega: OT: Consider naming of this method to something like `CreateDefaultSqlQueryGenerator()` should make it easier to understand the difference with `CreateFromSqlQueryGenerator()`.

src/EntityFramework.Relational/EntityFramework.Relational.csproj
```diff
@@ -302,9 +304,9 @@
     <Compile Include="Query\Sql\ISqlExpressionVisitor.cs" />
     <Compile Include="Query\RelationalQueryCompilationContext.cs" />
     <Compile Include="Query\RelationalQueryModelVisitor.cs" />
-    <Compile Include="Query\Sql\ISqlQueryGenerator.cs" />
+    <Compile Include="Query\Sql\IQuerySqlGenerator.cs" />
```
divega: Would this change affect providers?

anpete: Possibly, they would typically derive from DefaultQuerySqlGenerator.

src/EntityFramework.MicrosoftSqlServer/Query/Sql/Internal/SqlServerQuerySqlGeneratorFactory.cs
divega: With Query and SQL in

divega: With `Query` and `Sql` in whatever the right order is :smile:

anpete: Will improve this some more.

src/EntityFramework.Relational/EntityFramework.Relational.csproj
divega: I think I see usage of the interface name in the source code of our own provider, but it could be confusing the interfaces. If you can confirm we can tag this as providers-beware.

anpete: Yep, we should add the tag.

src/EntityFramework.MicrosoftSqlServer/Query/Sql/Internal/SqlServerQuerySqlGeneratorFactory.cs
anpete: Changed to CreateDefault and CreateFromSql per our factory naming pattern.

divega: Much nicer! Thanks!

src/EntityFramework.Relational/Query/Sql/FromSqlQuerySqlGenerator.cs
```diff
@@ -14,51 +13,39 @@
 
 namespace Microsoft.Data.Entity.Query.Sql
 {
-    public class RawSqlQueryGenerator : ISqlQueryGenerator
+    public class FromSqlQuerySqlGenerator : DefaultQuerySqlGenerator
```
divega: I was struggling to understand why it was ok to use inheritance here so I asked @anpete and it turns out this class is only used in the non-composable case (in which only a FromSql() call needs to be processed). Since we are making a breaking change for relational providers around the name of the class (and related methods) my preference would be to use a name that conveys that it only applies to the non-composable case. 

That said if we are going to end up removing this type (which according to @anpete is a clear possibility) then I think this name is fine.

anpete: Renamed to FromSqlNonComposedQuerySqlGenerator. Added base class for QuerySqlGenerators and moved the FromSql generator to internal.

src/EntityFramework.Relational/Storage/Internal/RelationalCommand.cs
```diff
@@ -322,7 +327,9 @@ public class RelationalCommand : IRelationalCommand
                 executeMethod,
                 async: async);
 
-        private DbCommand CreateCommand(IRelationalConnection connection)
+        private DbCommand CreateCommand(
+            IRelationalConnection connection,
+            IReadOnlyDictionary<string, object> parameterValues)
```
divega: I have mixed feelings about having one `RelationalCommand` type that supports two separate calling patterns at the same time:

1. Pass parameters in constructor, execute once and discard
2. Pass parameter metadata in constructor, execute multiple times rebinding parameter values each time

Perhaps this is just clever and I need to wrap my head around it :smile: but it makes me wonder if having only the second option would be a forcing function for us to make other improvements where we use this class. I will file a bug.

anpete: Agree we can move everything to fully separate construction from execution.

src/EntityFramework.Relational/Storage/RelationalParameter.cs
```diff
@@ -1,30 +1,46 @@
...
 
-        public virtual RelationalTypeMapping RelationalTypeMapping { get; }
+        public virtual void AddDbParameter(DbCommand command, object value)
+        {
+            Check.NotNull(command, nameof(command));
 
-        public virtual bool? Nullable { get; }
+            command.Parameters
+                .Add(_relationalTypeMapping
+                    .CreateParameter(command, _name, value ?? _value, _nullable));
```
divega: I suspect using `RelationalCommand` in hybrid mode (i.e. passing non-null initial values in the constructor, then sometimes passing parameter values on the `Execute` methods) would break because any null value passed on `Execute` would be ignored in favor of the initial value (which could be different from null). I assume we are never really using `RelationalCommand` in "hybrid" mode, so it doesn't have any immediate consequence, but I took note of this in https://github.com/aspnet/EntityFramework/issues/3704 as another potential reason allowing both calling patterns may be broken.

src/EntityFramework.Relational/Storage/Internal/RelationalCommand.cs
divega: https://github.com/aspnet/EntityFramework/issues/3704

divega:
:shipit: can't wait to see the numbers!

PR #3695, opened 2015-11-09
Fix #3655 - ValidEntityType message parameters reversed.

### LINKED ISSUE #3655: ArgumentException for ValidEntityType appears to have parameters reversed
Exception is thrown with:
```cs
public static Type ValidEntityType(Type value, [InvokerParameterName] [NotNull] string parameterName)
{
    // ...
    throw new ArgumentException(CoreStrings.InvalidEntityType(parameterName, value));
    // ...
}
```

but target signature is:
```cs
public static string InvalidEntityType([CanBeNull] object type, [CanBeNull] object argumentName)
{
    return string.Format(CultureInfo.CurrentCulture, GetString("InvalidEntityType", "type", "argumentName"), type, argumentName);
}
```

mikary:


divega:
:shipit:

PR #3697, opened 2015-11-09
Throw when instance of unmapped derived type is used

### LINKED ISSUE #743: Make sure we throw when an instance of an unmapped derived type is use
Regardless of whether inheritance is being mapped in the model, it is always possible that the runtime type of an instance passed to the context is actually an unmapped derived type. In EF6 we explicitly disallowed this, except that we made a special exception for proxies. The fact that we made this exception shows that sometimes using an unmapped derived type can be useful, so it maybe useful to allow this in EF7. However, doing so means that code cannot rely on a certain type, which may result in additional complexity and/or subtle bugs.

smitpatel:
Resolves #743 
~~Re-introduced `ModelItemNotFoundException`~~

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/StateManager.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/StateManager.cs
@@ -92,6 +92,11 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
 
                 var entityType = _model.FindEntityType(entity.GetType());
 
+                if (entityType == null)
+                {
+                    throw new InvalidOperationException(CoreStrings.EntityTypeNotFound(entity.GetType().DisplayName(false)));
+                }
+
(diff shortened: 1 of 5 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/StateManager.cs
```diff
@@ -85,6 +85,11 @@ public virtual InternalEntityEntry GetOrCreateEntry(object entity)
 
                 var entityType = _model.FindEntityType(entity.GetType());
 
+                if (entityType == null)
+                {
+                    throw new ModelItemNotFoundException(CoreStrings.EntityTypeNotFound(entity.GetType().DisplayName(false)));
```
ajcvickers: I think we can just throw InvalidOperationException here. Pretty sure we decided in API review to get rid of the ModelItemNotFoundException and I don't think we want to bring it back.

ajcvickers:
:shipit: (without the new exception type)

PR #3699, opened 2015-11-09
Migrations: Diff column types using the TypeMapper

### LINKED ISSUE #3434: Change of MaxLength not migrated to DB
I'm using beta8. So I don't think this is a dupe of #882, but maybe a followup on it.
I followed the steps given [here](http://ef.readthedocs.org/en/latest/getting-started/full-dotnet.html).

If I do ...
```
modelBuilder.Entity<Blog>()
    .Property(b => b.Url)
    .IsRequired()
    .HasMaxLength(256);
```
... and run the initial migration, I get a Blog table with a column defined as `Url (nvarchar(256), not null`, as I would expect.
If I subsequently change this to ...
```
modelBuilder.Entity<Blog>()
    .Property(b => b.Url)
    .IsRequired()
    .HasMaxLength(512);
```
... and do `Add-Migration` / `Update-Database`, the field stays as it was in the db, even though the change in `HasMaxLength` is present in the `ModelSnapshot` class:
```
b.Property<string>("Url")
    .IsRequired()
    .Annotation("MaxLength", 512);
```
If the only thing I change for the second migration is the value for `HasMaxLength` as done above, I get the following Migration code:
```
public partial class MySecondMigration : Migration
{
    {
    }
    {
    }
}
```
Which probably is the reasony why the database isn't changed. If I manually change this to ...
```
public partial class MySecondMigration : Migration
{
    {
        migrationBuilder.AlterColumn<string>("Url", "Blog", "nvarchar(512)");
    }
    {
        migrationBuilder.AlterColumn<string>("Url", "Blog", "nvarchar(256)");
    }
}
```
... applying and reverting the second migration works as expected.
I assume it's a bug that a change of MaxLength in the model is not migrated to the database? Because if I remove the `.IsRequired()` from the model, the column is altered as it should be if I run the migration, so subsequent column changes actually do seem to work. If I modify two properties at once (ie. change `IsRequired()` and `HasMaxLength`), only the change of the first one is migrated to the db.

bricelam:
Resolves #3434

### DIFF
--- a/src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
+++ b/src/EntityFramework.Relational/Migrations/Internal/MigrationsModelDiffer.cs
@@ -11,6 +11,7 @@ using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Metadata.Internal;
 using Microsoft.Data.Entity.Migrations.Operations;
+using Microsoft.Data.Entity.Storage;
 using Microsoft.Data.Entity.Utilities;
 
 namespace Microsoft.Data.Entity.Migrations.Internal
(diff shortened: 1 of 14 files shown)


ajcvickers:
:shipit:

PR #3701, opened 2015-11-09
Query: Perf: Improve performance of typed ValueBuffer factories by:

anpete:
1) Use DbDataReader.GetX methods when available for target type.
2) Only box result when required.
3) Only perform IsDBNull check when target is nullable.

Yields ~25% speed up


ajcvickers:
:shipit:

divega:
Nice!

PR #3702, opened 2015-11-09
Create specialized structs for original values, store generated values, and relationship snapshots

ajcvickers:
This replaces the general purpose sidecar code resulting in:
- Fewer memory allocations needed when tracking
- Less memory used when tracking in the common cases
- Optimized code paths based on how each of the structs is used


src/EntityFramework.Core/ChangeTracking/Internal/StoreGeneratedValues.cs
```diff
@@ -1,25 +1,72 @@
...
+            public void SetValue(IPropertyBase propertyBase, object value)
+            {
+                var index = propertyBase.GetStoreGeneratedIndex();
+
+                Debug.Assert(index != -1);
+
+                _values[index] = value ?? NullSentinel.Value;
+            }
+
+            private sealed class NullSentinel
```
anpete: private static readonly object NullSentinel  = new object();

This?

anpete:
:shipit:

Issue #3642, opened 2015-11-03
RevEng: Sqlite: Indexes for PK are scaffolded wrongly


smitpatel:
For SQLite, we are scaffolding indexes which are defined on PK properties. If there is no unique keyword in index then we scaffold `unique=false` for index. According to #700 , key's uniqueness will be based on the underlying index object. This causes bug if the PK is used as FK for relationship. Effectively making FK non-unique even though its unique.

rowanmiller:
Created a [task to look at the all-up design around indexes](https://github.com/aspnet/EntityFramework/issues/3710) (and listed this as an item that needs attention).

smitpatel:
Included fix in #3714

Issue #3710, opened 2015-11-10
Review how we create/handle indexes in Reverse Engineer


rowanmiller:
We've had a few issues around indexes, so rather than making isolated decisions without the all-up context we are creating an issue to take a holistic look at what we do.

Can you put together a summary of how we handle indexes and what needs to change. This should include the things discussed in:
* https://github.com/aspnet/EntityFramework/issues/3642
* https://github.com/aspnet/EntityFramework/issues/3664

natemcmaster:
When fixing, reenable `SqliteDatabaseModelFactoryTest.It_assigns_uniqueness_to_pk_foreign_key` (see https://github.com/aspnet/EntityFramework/pull/3714#discussion-diff-44565007R255). cc @smitpatel

lajones:
Per discussion on 11/24/15 with @rowanmiller and @ajcvickers we need the following:

1. FKs: If the name of the index matches that which the convention would automatically create then no need to add a separate index fluent API at all. But if the name is different will need to add it and then the runtime will use that index (matching based on FK properties in the correct order).
2. PKs: may need to define the PK and add ConstraintName() if the name of the underlying index is different from what the convention would assume.
3. Alternate Keys: we already generate these. Check that we correctly issue fluent API defining the name.

Issue #3642 was addressed by commit 9b8ea39 and PR #3764.
Issue #3664 will be revisited during the work above.

lajones:
Note to self: point 3 above is already done (see https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs#L303).

lajones:
Note: for point 2 above the correct API is `HasName()`.

lajones:
Fix checked in with commit dbb5c9f.

Issue #3551, opened 2015-10-25
Allow multiple foreign keys to use the same set of properties


AndriySvyryd:
Principal key and principal entity type should be included in the identifier of an FK

PR #3714, opened 2015-11-10
Create explicit index object for Keys & ForeignKeys

### LINKED ISSUE #700: Create explicit index objects for every index

smitpatel:
Resolves #700 
`Key` & `ForeignKey` has Index object associated. `IsUnique` and `IsClustered` from key/foreignkey moved to index object.
Need review from @bricelam for Migrations & @AndriySvyryd for metadata.

### DIFF
--- a/src/EntityFramework.Core/Metadata/IForeignKey.cs
+++ b/src/EntityFramework.Core/Metadata/IForeignKey.cs
@@ -47,6 +47,11 @@ namespace Microsoft.Data.Entity.Metadata
         INavigation PrincipalToDependent { get; }
 
         /// <summary>
+        ///     Gets the index object defined on the foreign key properties.
+        /// </summary>
+        IIndex Index { get; }
+
+        /// <summary>
         ///     Gets a value indicating whether the values assigned to the foreign key properties are unique.
         /// </summary>
         bool IsUnique { get; }
--- a/src/EntityFramework.Core/Metadata/IMutableForeignKey.cs
+++ b/src/EntityFramework.Core/Metadata/IMutableForeignKey.cs
@@ -73,9 +73,14 @@ namespace Microsoft.Data.Entity.Metadata
         IMutableNavigation HasPrincipalToDependent([CanBeNull] string name);
 
         /// <summary>
-        ///     Gets or sets a value indicating whether the values assigned to the foreign key properties are unique.
+        ///     Gets the index object defined on the foreign key properties.
         /// </summary>
-        new bool? IsUnique { get; set; }
+        new IMutableIndex Index { get; }
+
+        /// <summary>
+        ///     Gets a value indicating whether the values assigned to the foreign key properties are unique.
+        /// </summary>
+        new bool? IsUnique { get; }
 
         /// <summary>
         ///     Gets or sets a value indicating if this relationship is required. If true, the dependent entity must always be
--- a/src/EntityFramework.Core/Metadata/Internal/Index.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/Index.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 
@@ -49,5 +50,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
 
         [UsedImplicitly]
         private string DebuggerDisplay => Property.Format(Properties);
+
+        public virtual bool IsInUse() => (DeclaringEntityType.FindKey(Properties) != null) || DeclaringEntityType.FindForeignKeys(Properties).Any();
     }
 }
--- a/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -86,9 +86,13 @@ namespace Microsoft.Data.Entity.Metadata.Internal
             => HasKey(GetOrCreateProperties(clrProperties, configurationSource), configurationSource);
 
         private InternalKeyBuilder HasKey(IReadOnlyList<Property> properties, ConfigurationSource configurationSource)
-            => properties == null
-                ? null
-                : GetOrAdd(
+        {
+            if (properties == null)
+            {
+                return null;
+            }
+
+            var keyBuilder = GetOrAdd(
                     GetOrCreateProperties(properties, configurationSource),
                     Metadata.FindDeclaredKey,
                     (e, c) => e.UpdateConfigurationSource(c),
@@ -96,6 +100,17 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                     k => k.Builder,
                     configurationSource,
                     onNewKeyAdded: ModelBuilder.ConventionDispatcher.OnKeyAdded);
+
+            if (keyBuilder != null)
+            {
+                ModelBuilder.Entity(keyBuilder.Metadata.DeclaringEntityType.Name, ConfigurationSource.Convention)
+                    .HasIndex(keyBuilder.Metadata.Properties, configurationSource)
+                    .IsUnique(true, configurationSource);
+            }
+
+            return keyBuilder;
+        }
+
 
         public virtual ConfigurationSource? RemoveKey(
             [NotNull] Key key, ConfigurationSource configurationSource, bool canOverrideSameSource = true)
@@ -111,6 +126,11 @@ namespace Microsoft.Data.Entity.Metadata.Internal
             {
                 var removed = foreignKey.DeclaringEntityType.Builder.RemoveForeignKey(foreignKey, configurationSource);
                 Debug.Assert(removed.HasValue);
+            }
+
+            if (key.Index != null)
+            {
+                RemoveIndex(key.Index, configurationSource);
             }
 
             var removedKey = Metadata.RemoveKey(key.Properties);
@@ -600,12 +620,6 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                 return null;
             }
 
-            foreach (var index in Metadata.GetIndexes().Where(i => i.Properties.Contains(property)).ToList())
-            {
-                var removed = RemoveIndex(index, configurationSource);
-                Debug.Assert(removed.HasValue);
-            }
-
             var detachedRelationships = property.FindContainingForeignKeys().ToList()
                 .Select(DetachRelationship).ToList();
 
@@ -614,6 +628,12 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                 detachedRelationships.AddRange(key.FindReferencingForeignKeys().ToList()
                     .Select(DetachRelationship));
                 var removed = RemoveKey(key, configurationSource);
+                Debug.Assert(removed.HasValue);
+            }
+
+            foreach (var index in Metadata.GetIndexes().Where(i => i.Properties.Contains(property)).ToList())
+            {
+                var removed = RemoveIndex(index, configurationSource);
                 Debug.Assert(removed.HasValue);
             }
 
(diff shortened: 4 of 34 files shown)


src/EntityFramework.Core/Metadata/IForeignKey.cs
```diff
@@ -47,6 +47,11 @@ public interface IForeignKey : IAnnotatable
...
 
         /// <summary>
+        ///     Gets the index object defined on the foreign key properties.
+        /// </summary>
+        IIndex Index { get; }
+
+        /// <summary>
         ///     Gets a value indicating whether the values assigned to the foreign key properties are unique.
         /// </summary>
         bool IsUnique { get; }
```
AndriySvyryd: IsUnique should be removed

src/EntityFramework.Core/Metadata/IMutableForeignKey.cs
```diff
@@ -73,9 +73,14 @@ public interface IMutableForeignKey : IForeignKey, IMutableAnnotatable
...
-        ///     Gets or sets a value indicating whether the values assigned to the foreign key properties are unique.
+        ///     Gets the index object defined on the foreign key properties.
         /// </summary>
-        new bool? IsUnique { get; set; }
+        new IMutableIndex Index { get; }
+
+        /// <summary>
+        ///     Gets a value indicating whether the values assigned to the foreign key properties are unique.
+        /// </summary>
+        new bool? IsUnique { get; }
```
AndriySvyryd: IsUnique should be removed

src/EntityFramework.Core/Metadata/Internal/Index.cs
```diff
@@ -49,5 +50,20 @@ public virtual ConfigurationSource UpdateConfigurationSource(ConfigurationSource
 
         [UsedImplicitly]
         private string DebuggerDisplay => Property.Format(Properties);
+
+        public virtual bool NotInUse()
```
AndriySvyryd: Negate and rename to IsInUse

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -600,15 +620,15 @@ public void Attach(InternalEntityTypeBuilder entityTypeBuilder)
...
             foreach (var index in Metadata.GetIndexes().Where(i => i.Properties.Contains(property)).ToList())
             {
                 var removed = RemoveIndex(index, configurationSource);
                 Debug.Assert(removed.HasValue);
             }
 
-            var detachedRelationships = property.FindContainingForeignKeys().ToList()
-                .Select(DetachRelationship).ToList();
-
             foreach (var key in Metadata.GetKeys().Where(i => i.Properties.Contains(property)).ToList())
```
AndriySvyryd: Shouldn't the keys be removed before the indexes?

src/EntityFramework.Core/Metadata/Internal/InternalIndexBuilder.cs
```diff
@@ -17,26 +17,82 @@ public InternalIndexBuilder([NotNull] Index index, [NotNull] InternalModelBuilde
         {
         }
 
-        public virtual bool IsUnique(bool? isUnique, ConfigurationSource configurationSource)
+        public virtual bool IsUnique(bool unique, ConfigurationSource configurationSource)
+            => IsUnique(unique, configurationSource, runConventions: true);
+
+        public virtual bool IsUnique(bool isUnique, ConfigurationSource configurationSource, bool runConventions)
```
AndriySvyryd: rename parameter to "unique"

src/EntityFramework.Core/Metadata/Internal/InternalIndexBuilder.cs
```diff
@@ -17,26 +17,82 @@ public InternalIndexBuilder([NotNull] Index index, [NotNull] InternalModelBuilde
...
         {
-            if (configurationSource.CanSet(_isUniqueConfigurationSource, Metadata.IsUnique.HasValue)
-                || Metadata.IsUnique.Value == isUnique)
+            if (!CanSetUnique(isUnique, configurationSource))
```
AndriySvyryd: If unique is false this should also check for keys using this index and remove them.

src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -1192,11 +1136,7 @@ private bool CanSetPrincipalKey(IReadOnlyList<Property> properties, Configuratio
                                       ? Metadata.PrincipalKey.Properties
                                       : null);
 
-            isUnique = isUnique ??
-                       (_isUniqueConfigurationSource.HasValue
-                        && _isUniqueConfigurationSource.Value.Overrides(configurationSource)
-                           ? ((IForeignKey)Metadata).IsUnique
-                           : (bool?)null);
+            isUnique = isUnique ?? Metadata.IsUnique;
```
AndriySvyryd: The ConfigurationSource check should be preserved. I'm surprised this hasn't caused a test failure.
You can add methods to get the ConfigurationSource.

src/EntityFramework.Core/Metadata/Internal/InternalIndexBuilder.cs
smitpatel: filed #3716

src/EntityFramework.Core/Metadata/IForeignKey.cs
ajcvickers: I don't  remember it being the intention that foreign keys, which define relationships in our metadata, would no longer control whether the relationship is 1:1 or 1:*. Does this mean that you can no longer have a 1:1 relationship in the model without also having an index?

I think we should discuss in the design meeting.

src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
smitpatel: Luckily no test failures :wink: 
`_isUniqueConfigurationSource` was being used for following purposes:
- While adding navigation to align uniqueness with collection/reference navigation - I think it would be better to pass that task to `CanSetUnique` on IndexBuilder since there can be other relationships defined over same properties. If the uniqueness based on navigation is clashing with some other foreign key defined over same properties then navigation should not be added.
- While setting isUnique - Same as above since it also needs to take care of other relationships over same properties.
- While replacing the foreignkey - the case mentioned above - It should be used in this case so that if the foreign key source is higher than isUnique configuration source then FK should have ability to change uniqueness though I am not sure how would that work with other FKs defined on same properties.

bricelam:
Migrations changes look good.

test/EntityFramework.Sqlite.Design.FunctionalTests/SqliteDatabaseModelFactoryTest.cs
```diff
@@ -251,7 +251,8 @@ FOREIGN KEY (BuddyId) REFERENCES Friends(Id)
             Assert.Equal(table.FindPrimaryKey(), fk.PrincipalKey);
         }
 
-        [Fact]
+        // TODO: See issue #3710
+        //[Fact]
```
natemcmaster: :cry: Can you add this to #3710 to ensure it is re-enabled?

smitpatel: is it reported on #3642 which is superseded by #3710

MarcoLoetscher:
Please note the following issues:
[EF creates index subsumed by clustered key](https://entityframework.codeplex.com/workitem/2535)
[EF creates unnecessary index](https://entityframework.codeplex.com/workitem/2706)
[Too many indexes on One-to–Zero-or-One Relationship](https://entityframework.codeplex.com/workitem/2851)
Thank you

smitpatel:
@MarcoLoetscher - regarding issues you mentioned,
- As for unnecessary indexes, there won't be any. That is the reason for this code change. Essentially now every key & foreign key will have index object associated with it. And there can be only one index on a fixed set of properties. So if a key & a foreign key uses the same set of properties, there will be only one index for both of them. That should resolve out last 2 issues where PK & FK is on same set of properties. We will be creating only one index for that.

- For the index subsumed by clustered key - By looking at the issue, I suppose the index on fk property is unnecessary because it is part of clustered key. Clustered key is SQL specific default for PK. Since EF7 targets other datastores also, the index on FK property may not be unnecessary.

test/EntityFramework.Sqlite.Design.FunctionalTests/SqliteDatabaseModelFactoryTest.cs
smitpatel: Fixed.

src/EntityFramework.Sqlite.Design/Internal/SqliteDmlParser.cs
```diff
@@ -75,7 +75,8 @@ public static void ParseColumnDefinition(TableModel table, string statement)
                 return;
             }
 
-            if (statement.IndexOf(" UNIQUE", StringComparison.OrdinalIgnoreCase) > 0)
+            if (statement.IndexOf(" UNIQUE", StringComparison.OrdinalIgnoreCase) > 0
+                || statement.IndexOf(" PRIMARY KEY", StringComparison.OrdinalIgnoreCase) > 0)
```
lajones: I believe these changes fix #3642. Could you update that issue as well? The Design changes look good to me but I would get others to check the other changes.

src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
smitpatel: After discussing in-person, everything is working correctly so looks like this check is not required. If in future if some bug arises, we can revisit.

smitpatel: Updated and added a test.

src/EntityFramework.Core/Metadata/IForeignKey.cs
smitpatel: Foreign key does control the relationship is 1:1 or 1:* but it won't store the information with in itself. It will modify the associated index uniqueness appropriately. So you cannot have unique & non-unique FKs using same set of properties ( #3551 enables that scenario). It also mean that you can no longer have 1:1 relationship in the model without having unique index on same properties which is present situation in migration and database. All our relationships in database are non-unique regardless of FK's uniqueness. Correcting it exposed #3648 because we were inserting duplicate values in unique FK temporarily.

I will bring this to design meeting for further discussion.

src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
smitpatel: How does this look now?

AndriySvyryd: looks good

AndriySvyryd:
:shipit: pending design discussion

test/EntityFramework.Sqlite.Design.FunctionalTests/SqliteDatabaseModelFactoryTest.cs
natemcmaster: :+1:

src/EntityFramework.Sqlite.Design/Internal/SqliteDmlParser.cs
```diff
@@ -90,7 +91,8 @@ public static void ParseColumnDefinition(TableModel table, string statement)
         public static void ParseConstraints(TableModel table, string statement)
         {
             var constraint = statement.Split(' ', '(')[0];
-            if (constraint.Equals("UNIQUE", StringComparison.OrdinalIgnoreCase))
+            if (constraint.Equals("UNIQUE", StringComparison.OrdinalIgnoreCase)
+                || statement.Equals(" PRIMARY KEY", StringComparison.OrdinalIgnoreCase))
```
natemcmaster: Tests? I don't think this will work because of the space at the beginning of the string literal.

smitpatel:
closing this in favour of #3764

PR #3715, opened 2015-11-10
Query: Perf: Use iterator for sync Include

anpete:


maumar:
:shipit:

src/EntityFramework.Relational/Query/QueryMethodProvider.cs
```diff
@@ -177,27 +176,24 @@ var relatedEntitiesLoaders
...
+                    .Include(
+                        entityAccessor == null ? innerResult : entityAccessor(innerResult), // TODO: Compile time?
+                        navigationPath,
+                        relatedEntitiesLoaders,
+                        querySourceRequiresTracking);
+
+                yield return innerResult;
+            }
+
+            foreach (var includeRelatedValuesStrategy in includeRelatedValuesStrategies)
```
divega: Should the rest be wrapped in a finally block?

test/EntityFramework.Microbenchmarks/Query/Profile.cs
```diff
@@ -22,12 +21,26 @@ var connectionString
                 = $@"Server={BenchmarkConfig.Instance.BenchmarkDatabaseInstance};Database=Perf_Query_Simple;Integrated Security=True;MultipleActiveResultSets=true;";
 
             _context = new OrdersContext(connectionString);
-            _query = _context.Products.AsNoTracking().Where(p => p.Retail < 15);
+
+            //_query = _context.Products.AsNoTracking().Where(p => p.Retail < 15);
```
divega: Intentional?

src/EntityFramework.Relational/Query/QueryMethodProvider.cs
anpete: No, this will be handled by the iterator state machine.

test/EntityFramework.Microbenchmarks/Query/Profile.cs
anpete: Yeah, I'm adding the class to make profiling easier.

src/EntityFramework.Relational/Query/QueryMethodProvider.cs
divega: I don't understand why it but OK if you do :smile: (anyway not sure what happened to the PR)

PR #3723, opened 2015-11-11
Move error messages into resource files

mikary:
Moving string literal error messages discovered in #2752 into appropriate resource files

### DIFF
--- a/src/EntityFramework.Core/Properties/CoreStrings.Designer.cs
+++ b/src/EntityFramework.Core/Properties/CoreStrings.Designer.cs
@@ -53,6 +53,30 @@ namespace Microsoft.Data.Entity.Internal
         }
 
         /// <summary>
+        /// Unhandled node type: {nodeType}
+        /// </summary>
+        public static string UnhandledNodeType([CanBeNull] object nodeType)
+        {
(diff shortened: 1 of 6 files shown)


lajones:
:shipit: Some of the error messages in ExpressionPrinter don't seem very useful (e.g. about what you should do next) but if they're just meant for debug or are made clear by the context in which they're issued then no problem.

src/EntityFramework.Commands/Design/Internal/CommandLogger.cs
```diff
@@ -85,7 +86,7 @@ public virtual IDisposable BeginScope([NotNull] object state)
                     WriteDebug(message.ToString());
                     break;
                 default:
-                    Debug.Fail("Unexpected event type: " + logLevel);
+                    Debug.Fail( CommandsStrings.UnexpectedLogEvent(logLevel));
```
bricelam: Debug strings shouldn't be resources. They'll just be a waste of space on Release builds.

Eilon: :+1:

mikary:
I believe the error messages are generally intended to be useful for debugging. We should also have a chance to revisit them in API review now that they are included in the resources.

PR #3729, opened 2015-11-11
Fix some static analysis warnings

bricelam:
Mostly...
* Unused code
* Parameter name alignment with base types
* Mispeligns


anpete:
:shipit:

PR #3735, opened 2015-11-12
Remove SqliteDmlParser

natemcmaster:
It was never necessary to do this. I just didn't know about `PRAGMA index_list`.


natemcmaster:
cc @smitpatel @lajones

lajones:
Looks much better than what was there before. :shipit:
BTW - is there anything which tells whether an index is for the primary key?

natemcmaster:
Yup. The "origin" column. 
> **PRAGMA schema.index_list(table-name);**

> This pragma returns one row for each index associated with the given table.

> Output columns from the index_list pragma are as follows:

> 1. A sequence number assigned to each index for internal tracking purposes.
1. The name of the index.
1. "1" if the index is UNIQUE and "0" if not.
1. "c" if the index was created by a CREATE INDEX statement, "u" if the index was created by a UNIQUE constraint, or "pk" if the index was created by a PRIMARY KEY constraint.
1. "1" if the index is a partial index and "0" if not.

https://www.sqlite.org/pragma.html#pragma_index_list

lajones:
Then we might consider doing the same as the SQL Server one does and not return PK indexes at all. But don't let that stop you from checking this in.

natemcmaster:
I agree. We will need to address this for https://github.com/aspnet/EntityFramework.Docs/issues/61 and when the dust settles on the FK/indexes changes to the model.

Issue #3583, opened 2015-10-28
Exception Messages: Name of entityType


smitpatel:
We have inconsistencies in what do we use for name of entity type when throwing exception. At present we use `EntityType.Name` and `EntityType.DisplayName()` both with no fixed guideline which one to use for particular case. We need to make it consistent or define rules when one is preferred over the other.

divega:
Clearing up milestone so that we can see this in triage and determine its priority.

rowanmiller:
Dupe of https://github.com/aspnet/EntityFramework/issues/734

PR #3737, opened 2015-11-12
:bug: object is not a valid navigation

### LINKED ISSUE #3727: System.Object being discovered as EntityType through navigation.
I tried today to update my solution to RC1 from beta8 and get the exception ```The derived type 'FieldDefinitionEntity' cannot have keys other than those declared on the root type.```. The root type is not mapped, only the derivied type. Should the mapping throw this exception anyway?
```
public class FieldDefinitionEntity : FieldDefinitionEntityBase
{
    public virtual bool MultiCulture { get; set; }
}

public abstract class FieldDefinitionEntityBase
{
    public virtual string FieldType { get; set; }
    public virtual string Id { get; set; }
}
```
Mapping
```
builder.Entity<FieldDefinitionEntity>(b =>
{
    b.ToTable("FieldDefinition", "Products");
    b.HasKey(p => p.Id);
});
```

smitpatel:
Fixes #3727 
Blocked `object` being discovered as entity type. Entity type with clr type `System.Object` is invalid entity type (due to absence of key) which is verified in model validator. Since System.Object is base type for any clr type, we set this entity type as base type for all other entities. Therefore we need to block it from being discovered in first place.
Also modified exception message in order to give information on what is the base type.

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/EntityType.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/EntityType.cs
@@ -289,7 +289,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
         {
             if (_baseType != null)
             {
-                throw new InvalidOperationException(CoreStrings.DerivedEntityTypeKey(Name));
+                throw new InvalidOperationException(CoreStrings.DerivedEntityTypeKey(DisplayName(), _baseType.DisplayName()));
             }
 
             if (_primaryKey != null)
@@ -379,7 +379,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
 
             if (_baseType != null)
             {
-                throw new InvalidOperationException(CoreStrings.DerivedEntityTypeKey(Name));
+                throw new InvalidOperationException(CoreStrings.DerivedEntityTypeKey(DisplayName(), _baseType.DisplayName()));
             }
 
             foreach (var property in properties)
--- a/src/EntityFramework.Core/Properties/CoreStrings.resx
+++ b/src/EntityFramework.Core/Properties/CoreStrings.resx
@@ -361,7 +361,7 @@
     <value>The '{factory}' cannot create a value generator for property '{property}' on entity type '{entityType}'. Only integer properties are supported.</value>
   </data>
   <data name="DerivedEntityTypeKey" xml:space="preserve">
-    <value>The derived type '{derivedType}' cannot have keys other than those declared on the root type.</value>
+    <value>The derived type '{derivedType}' cannot have keys other than those declared on the root type '{rootType}'.</value>
   </data>
   <data name="CircularInheritance" xml:space="preserve">
     <value>The entity type '{entityType}' cannot inherit from '{baseEntityType}' because '{baseEntityType}' is a descendent of '{entityType}'.</value>
--- a/src/Shared/PropertyInfoExtensions.cs
+++ b/src/Shared/PropertyInfoExtensions.cs
@@ -31,7 +31,8 @@ namespace System.Reflection
 
             if (isPrimitiveProperty(targetType)
                 || targetType.GetTypeInfo().IsInterface
-                || targetType.GetTypeInfo().IsValueType)
+                || targetType.GetTypeInfo().IsValueType
+                || targetType == typeof(object))
             {
                 return null;
             }
--- a/test/EntityFramework.Core.Tests/Metadata/ModelConventions/RelationshipDiscoveryConventionTest.cs
+++ b/test/EntityFramework.Core.Tests/Metadata/ModelConventions/RelationshipDiscoveryConventionTest.cs
@@ -861,6 +861,8 @@ namespace Microsoft.Data.Entity.Tests.Metadata.Conventions
         {
             public int Id { get; set; }
 
+            public object Object { get; set; }
+
             public static OneToManyDependent Static { get; set; }
 
             public OneToManyDependent WriteOnly
--- a/src/EntityFramework.Core/Properties/CoreStrings.Designer.cs
+++ b/src/EntityFramework.Core/Properties/CoreStrings.Designer.cs
@@ -53,30 +53,6 @@ namespace Microsoft.Data.Entity.Internal
         }
 
         /// <summary>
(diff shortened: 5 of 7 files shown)


test/EntityFramework.Core.Tests/Metadata/ModelConventions/RelationshipDiscoveryConventionTest.cs
```diff
@@ -861,6 +861,8 @@ private class EntityWithNoValidNavigations
         {
             public int Id { get; set; }
 
+            public object Object { get; set; }
```
divega: Just to make sure, would it make sense to have a test in which:
1. A CLR entity type derives from a base type that different from System.Object but isn't a valid entity type
2. There is a property of that base type somewhere in the model?

src/EntityFramework.Core/Properties/CoreStrings.resx
```diff
@@ -361,7 +361,7 @@
     <value>The '{factory}' cannot create a value generator for property '{property}' on entity type '{entityType}'. Only integer properties are supported.</value>
   </data>
   <data name="DerivedEntityTypeKey" xml:space="preserve">
-    <value>The derived type '{derivedType}' cannot have keys other than those declared on the root type.</value>
+    <value>The derived type '{derivedType}' cannot have keys other than those declared on the root type {rootType}.</value>
```
AndriySvyryd: Add quotes

src/EntityFramework.Core/Metadata/Internal/EntityType.cs
```diff
@@ -379,7 +379,7 @@ public virtual Key AddKey([NotNull] Property property, ConfigurationSource confi
 
             if (_baseType != null)
             {
-                throw new InvalidOperationException(CoreStrings.DerivedEntityTypeKey(Name));
+                throw new InvalidOperationException(CoreStrings.DerivedEntityTypeKey(Name, _baseType.Name));
```
AndriySvyryd: Use .DisplayName()

AndriySvyryd:
:shipit:

src/Shared/PropertyInfoExtensions.cs
```diff
@@ -31,7 +31,8 @@ public static Type FindCandidateNavigationPropertyType(this PropertyInfo propert
 
             if (isPrimitiveProperty(targetType)
                 || targetType.GetTypeInfo().IsInterface
-                || targetType.GetTypeInfo().IsValueType)
+                || targetType.GetTypeInfo().IsValueType
+                || targetType == typeof(object))
```
ajcvickers: I think this might be the same thing @divega is saying, but it doesn't seem to me that special casing "object" is the right thing to do here because object is not any more special than any other type that might be the base type for all or some entities in the model,

src/EntityFramework.Core/Metadata/Internal/EntityType.cs
smitpatel: Standing issue #3583. Which one you prefer here?

AndriySvyryd: .DisplayName(), since there's no ambiguity

src/Shared/PropertyInfoExtensions.cs
AndriySvyryd: It is special because it is much less likely to be an entity type than any other base type

test/EntityFramework.Core.Tests/Metadata/ModelConventions/RelationshipDiscoveryConventionTest.cs
smitpatel: We have test to verify 
1. Non-primitive types are discovered as entity type
2. When base type is present as entity type in the model then base type will be set for derived types
3. When invalid entity is present in model throw exception
Above scenario is just combination of all of them and should not be necessary.
Though I can add test if you want. :smile:

ajcvickers: I would like to see this test too. I'm not understanding how this fix will make the test Diego suggested pass.

divega: I think it is important that we understand the behavior in this general case because `System.Object` should just be just an example of it.

smitpatel: General behavior is,
If user have 2 entity types in the model added explicitly/conventionally, and if they are in CLR hierarchy then we will create TPH. We will throw exception if user tries to create key in derived type.
We will throw exception if there is any invalid entity type (entity without PK) regardless of above.

`System.Object` is kind of stands out because,
1. We know that it is not valid entity type (unless shadow PK defined but if we want to support that we have to remove this check)
2. object is base type for every clr type hence it will be set as base type for every other entity type.

This change is not the fix for the user reported exception.
User has property of `object` type which cannot be mapped by typemapper. Hence user should be notified to exclude it from model since discovering `object` as entity type throws misleading exception.
For any type other than `object`, there won't be clr inheritance unless user has defined it using derived class. At which point it is fair to throw above exception notifying user that we have put both entities in same hierarchy by convention (or there is invalid entity type in the model)

divega: @ajcvickers I talked to @smitpatel and @AndriySvyryd and right now I understand the reason for this design and I am ok with it. Hopefully I can explain it to you tomorrow before I forget :smile:

ajcvickers: @divega After looking through it all again it seems to me that the only real issue is that the original exception message didn't specify that we had pulled "object" into the model. With an appropriate change to the message (possibly more than what is in this PR, although that is a good start) it seems that is all we need to do. In other words, I don't think we also need to special case object. But I look forward to hearing why special casing object is a better thing to do than just having a better exception message for the general case.

Issue #3598, opened 2015-10-29
DbContext.Database.ExecuteSqlCommand should return rows affected


rowanmiller:
In EF6 the method returns an int for rows affected... currently the method is void in EF7.

mikary:
Fixed in #3738

PR #3738, opened 2015-11-12
API review changes relating to RelationalCommands

mikary:
API review changes (excluding `ISqlCommandBuilder`).
Could use another look on RelationalTypeMapper to be sure the get mappings signatures are correct


mikary:
Fixes #3598 - `ExecuteSqlCommand` should return rows affected

src/EntityFramework.Relational/Storage/Internal/RelationalCommandBuilder.cs
```diff
@@ -33,7 +33,7 @@ public class RelationalCommandBuilder : IRelationalCommandBuilder
             _typeMapper = typeMapper;
         }
 
-        public virtual IndentedStringBuilder CommandTextBuilder { get; } = new IndentedStringBuilder();
+        public virtual IndentedStringBuilder Instance { get; } = new IndentedStringBuilder();
```
divega: Why is this `Instance`? Is that what we decided? Looks weird but I may be missing something.

divega: Looking at the usage, I think I remember now. It still looks weird though :smile: `Text` or `Sql` would probably look better IMO, but I am fine with `Instance` if others really like it.

divega:
@mikary I remember in the API review we saw an `Execute` method that potentially needed to aggregate rows affected from multiple statements. Can't remember what we decided but I think we said we would do it if possible. Is that included in this PR?

src/EntityFramework.Relational/Storage/Internal/RelationalCommandBuilder.cs
mikary: It seems a bit strange, but I believe this is consistent with other internal services that are provided only for the sake of extensions. We could back it with a private field with a more descriptive name, but that seemed unnecessary.

mikary:
@divega We have two methods in `RelationalCommandExtensions` which run non-query commands and return `void` and `Task` rather than aggregating. We made this class internal in this change, but I didn't modify the return type because it's not clear what we need when aggregating results (i.e. an `IEnumerable<int>`, an `int` with the sum of the results excluding -1's, etc.). For now I think it's best that we YAGNI on the return value until we have a need to interpret the results.

mikary:
Discussed `ISqlCommandBuilder` changes with @anpete and settled on a rename to `IRawSqlCommandBuilder` as an incremental solution. Also eliminiated `RelationalSqlExecutor`.

divega:
Renames look good to me. @mikary can you please get final sign off from @anpete?

src/EntityFramework.Relational/Storage/Internal/RelationalCommandBuilder.cs
anpete: Instance seems wrong. To me, "Instance" usually denotes the singleton pattern. My vote would be to leave the name and also make it an explicit impl. - It should only really be used via EMs

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -179,8 +179,16 @@ protected virtual PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder builde
 
             RelationalTypeMapping mapping;
 
-            if (!_typeMapper.TryGetMapping(column.DataType, out mapping)
-                || mapping.ClrType == null)
+            try
```
anpete: We generally don't want to use exceptions in this way.

mikary: I would like to do something else here, but there isn't a method on the public API that doesn't throw. The implementation does have `FindMapping` methods for the other two patterns, but we are making those protected as a part of these changes. I'm not sure we have much of an alternative here.

anpete: We should change or add API. Ideally a single method on the interface combined with companion sugar extension methods. cf Logging etc.

mikary:
Updated the `Instance` method to be an explicit interface implementation
Moved the `Find` methods to the `IRelationalTypeMapper` and converted the `Get` methods to extensions so that we don't have to use any strange catch patterns

anpete:
:shipit:

PR #3739, opened 2015-11-12
React to changes in Testing

natemcmaster:
Remove Xunit functionality that exists in the Microsoft.AspNet.Testing project.

:watch: https://github.com/aspnet/Testing/pull/186


bricelam:
Have you ensured things still works with TestDriven.NET?

natemcmaster:
Yes, I tested TestDriven.Net. It should work the same.

test/EntityFramework.Core.FunctionalTests/project.json
```diff
@@ -6,8 +6,7 @@
   },
   "dependencies": {
     "EntityFramework.InMemory": "7.0.0-*",
-    "xunit.assert": "2.1.0-*",
-    "xunit.extensibility.execution": "2.1.0-*"
+    "Microsoft.AspNet.Testing": "1.0.0-*"
```
bricelam: We'll also need to ensure this gets published to NuGet for our provider writers...

natemcmaster:
My intention here was to align with AspNet.Testing so we can re-use their platform-detection pieces. This comes with the cost of an additional dependency. Would it be better to just copy pieces of AspNet.Testing that we need instead?

Issue #3499, opened 2015-10-21
Exception with in-list where clause across joined table


jparish9:
Windows 10, VS 2015, IIS Express, SQL Server Express 2014, EF 7.0.0-beta8, ASP.NET 6.0.0-beta8

I am seeing a "Query source has already been associated with an expression" exception when trying to run an in-list where clause across a joined table.

Models:
```c#
public class Blog
{
        public int BlogId { get; set; }

        public string Status { get; set; }
}
```

```c#
public class Post
{
        public int PostId { get; set; }

        public int BlogId { get; set; }

        public string PostText { get; set; }

        public Blog Blog { get; set; }
}
```

Query 1 (works):
```c#
_dbContext.Blogs
                .Where(b => new[] { "A", "B", "C" }.Contains(b.Status))
                .ToList();
```
Query 2 (exception):
```c#
_dbContext.Posts
                .Where(p => new[] { "A", "B", "C" }.Contains(p.Blog.Status))
                .ToList();
```

Stack trace:
```c#
Query source has already been associated with an expression
   at Remotion.Linq.Clauses.QuerySourceMapping.AddMapping(IQuerySource querySource, Expression expression)
   at Microsoft.Data.Entity.Query.EntityQueryModelVisitor.VisitJoinClause(JoinClause joinClause, QueryModel queryModel, Int32 index)
   at Microsoft.Data.Entity.Query.RelationalQueryModelVisitor.<>n__0(JoinClause joinClause, QueryModel queryModel, Int32 index)
   at Microsoft.Data.Entity.Query.RelationalQueryModelVisitor.<>c__DisplayClass58_0.<VisitJoinClause>b__0()
   at Microsoft.Data.Entity.Query.RelationalQueryModelVisitor.OptimizeJoinClause(JoinClause joinClause, QueryModel queryModel, Int32 index, Action baseVisitAction, MethodInfo operatorToFlatten, Boolean outerJoin)
   at Microsoft.Data.Entity.Query.RelationalQueryModelVisitor.VisitJoinClause(JoinClause joinClause, QueryModel queryModel, Int32 index)
   at Remotion.Linq.Clauses.JoinClause.Accept(IQueryModelVisitor visitor, QueryModel queryModel, Int32 index)
   at Remotion.Linq.QueryModelVisitorBase.VisitBodyClauses(ObservableCollection`1 bodyClauses, QueryModel queryModel)
[...]

PR #3750, opened 2015-11-13
Fix to #3499 - Exception with in-list where clause across joined table

maumar:
Problem was that during navigation expansion we would add the created join clauses into the query model that the navigation was used. However, sometimes the navigation references a query source from the outside query, and the joins should be added there instead.

Fix is to keep track of hierarchy of query models and testing whether the given query model is the right one to add the join to (i.e. it contains a declaration of the query source that the join references)

Also added some minor fixes - we were not performing navigation expansions for result operators and we were not translating EF.Property() inside Contains result operator properly.


maumar:
@anpete

anpete:
:shipit:

PR #3754, opened 2015-11-13
fixing build break due to recent DI changes

maumar:


maumar:
@bricelam

smitpatel:
and I was stuck on this :cry:

pranavkm:
Are you on an old build of DI? If you're having trouble switch to volatile. You shouldn't need to do this change.

bricelam:
Oh wow, our local mirror (at 15957) is significantly behind the ci (at 15983). @victorhurdugaci?

victorhurdugaci:
There's no more local mirror. There's only a cache of the local coherence. If you need a cache, you have to create job that does that.

We changed it so that all the test builds use the same configuration and do the same thing

victorhurdugaci:
I thought you're not using it for anything else other that the CI after syncing with @natemcmaster

victorhurdugaci:
Btw, the cache is now in `D:\feeds\Coherence`

src/EntityFramework.Commands/Design/Internal/StartupInvoker.cs
```diff
@@ -95,7 +95,7 @@ protected virtual IServiceCollection ConfigureHostServices([NotNull] IServiceCol
             => services
 #if DNX451 || DNXCORE50
                 .ImportDnxServices()
-                .AddSingleton<IHostingEnvironment>(new HostingEnvironment { EnvironmentName = _environment })
+                .AddSingleton<IHostingEnvironment>(_ => new HostingEnvironment { EnvironmentName = _environment })
```
natemcmaster: Why do we need the lambdas? Assuming you are reacting to https://github.com/aspnet/DependencyInjection/commit/ff7eb2f2c9bc584205f143f9ca23feee024f0ee6, it doesn't look  like this needs to change.

maumar: our build is/was using old version of DI, will close this and wait for the right version to propagate

PR #3763, opened 2015-11-17
Query: Async GroupBy improvements:

anpete:
- Created EF impl. of async GroupBy using await.
- Use IGrouping when client-eval'ing GroupBy because groups are always buffered.
- Fixes sporadic async GroupBy test failure.


divega:
:shipit:

PR #3764, opened 2015-11-17
Add index on foreign key properties by convention

smitpatel:
Implements #700

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/Index.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/Index.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Utilities;
 
@@ -49,5 +50,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
 
         [UsedImplicitly]
         private string DebuggerDisplay => Property.Format(Properties);
+
+        public virtual bool IsInUse() => DeclaringEntityType.FindForeignKeysInHierarchy(Properties).Any();
     }
 }
--- a/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -600,12 +600,6 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                 return null;
             }
 
-            foreach (var index in Metadata.GetIndexes().Where(i => i.Properties.Contains(property)).ToList())
-            {
-                var removed = RemoveIndex(index, configurationSource);
-                Debug.Assert(removed.HasValue);
-            }
-
             var detachedRelationships = property.FindContainingForeignKeys().ToList()
                 .Select(DetachRelationship).ToList();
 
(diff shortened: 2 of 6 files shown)


src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -617,6 +611,12 @@ public void Attach(InternalEntityTypeBuilder entityTypeBuilder)
                 Debug.Assert(removed.HasValue);
             }
 
+            foreach (var index in Metadata.GetIndexes().Where(i => i.Properties.Contains(property)).ToList())
+            {
+                var removed = RemoveIndex(index, configurationSource);
+                Debug.Assert(removed.HasValue);
```
ajcvickers: Will this assert fail if the application creates an index explicitly on a property created by convention?

src/EntityFramework.Core/Metadata/Internal/Index.cs
```diff
@@ -31,7 +32,7 @@ public class Index : ConventionalAnnotatable, IMutableIndex
 
         public virtual IReadOnlyList<Property> Properties { get; }
         public virtual EntityType DeclaringEntityType { get; }
-        public virtual InternalIndexBuilder Builder { get; [param: CanBeNull] set; }
+        public virtual InternalIndexBuilder Builder { get;[param: CanBeNull] set; }
```
bricelam: BTW, this formatting bug is fixed in Update 1.

smitpatel: Will revert this

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
smitpatel: I have just moved the code block here so that we detach relationships before removing indexes because detaching relationship will remove FK and conventional index defined for that FK.
This assert should never fail. This is `RemoveProperty` function. When we add any component in the model, we update the configuration source (CS) of the property to have highest CS of all component defined using this property. At the start of this function we check if given CS is able to remove this property. If it can remove the property then all components defined using this property should get removed. Hence assert should always pass.

ajcvickers: Sounds good. I didn't know we did this: "we update the configuration source (CS) of the property to have highest CS of all component defined using this property."

ajcvickers:
:shipit:

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
AndriySvyryd: @ajcvickers Yes, we do that for any referenced member of the model. However we don't downgrade the CS if the referencing member is removed, that's something that would be fixed with https://github.com/aspnet/EntityFramework/issues/2869

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -670,7 +670,15 @@ private RelationshipBuilderSnapshot DetachRelationship([NotNull] ForeignKey fore
...
-            RemoveShadowPropertiesIfUnused(foreignKey.Properties);
+            var index = Metadata.FindIndex(foreignKey.Properties);
+            if (index != null
+                && !index.IsInUse())
+            {
+                // Remove index if created by convention
+                index.DeclaringEntityType.Builder.RemoveIndex(index, ConfigurationSource.DataAnnotation);
+            }
+
+            RemoveShadowPropertiesIfUnused(foreignKey.Properties.Where(p => p.DeclaringEntityType.FindDeclaredProperty(p.Name) != null).ToList());
```
smitpatel: This fails. I don't think we set builder to null while removing property.

Issue #3282, opened 2015-09-30
[Inheritence] RequiresValueGenerator is not handled properly


smitpatel:
When there is conflicting relationship between baseType and derivedType, we remove the conflicting relationship but this marks the `RequiresValueGenerator` as true which gets merged with base type property. Since base type still have the relationship it throws following error:
`System.InvalidOperationException: The property 'SquadId' on entity type 'Gear' has value generation on add enabled and it is included in the foreign key {'SquadId'}. Foreign key properties should not have 
value generation on add enabled unless they are part of a cycle where they are the only properties with value generation on add enabled.`

Repro:
````
public class Squad
    {
        public int Id { get; set; }

        public virtual ICollection<Gear> Members { get; set; }
    }

    public class Gear
    {
        public int Id { get; set; }
        public int SquadId { get; set; }
        public virtual Squad Squad { get; set; }
    }

    public class Officer : Gear
    {
    }

    public class MyContext : DbContext
    {
        public DbSet<Gear> Gears { get; set; }
        public DbSet<Squad> Squads { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(
                @"Server=(localdb)\MSSQLLocalDB;Database=_TestApp5;Integrated Security=True;MultipleActiveResultSets=True;Connect Timeout=30");
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Officer>().BaseType<Gear>();
            // Defining relationship explicitly does not fix the issue either
            modelBuilder.Entity<Squad>(b =>
            {
                b.HasKey(s => s.Id);
                b.HasMany(s => s.Members).WithOne(g => g.Squad).ForeignKey(g => g.SquadId);
            });
        }
    }
````

AndriySvyryd:
We keep `RequiresValueGenerator` up to date using conventions, but it doesn't seem like something a user would want to turn off, so I suggest calculating the correct value as the default, this would be the most robust solution.
Having said that we can still fix this bug using conventions.

smitpatel:
This does not repro anymore.
@AndriySvyryd - Any insights what may have fixed this?

AndriySvyryd:
We changed the way `RequiresValueGenerator` is used so it should work even if it set to `true` in this case. However we should still add a test to ensure the conventions set it to 'false'
@ajcvickers might have more suggestions

[...]

PR #3775, opened 2015-11-17
Add test for RequireValueGenerator handling

smitpatel:
#3282 got fixed by other changes.
Removed all work-arounds and added test


bricelam:
:shipit:

PR #3776, opened 2015-11-17
Create Shadow Property for FK

### LINKED ISSUE #3576: Shadow property is not created by .ForeignKey(Type, string)
... and is not taken into account if defined late in the Model.

Repro: this code should not throw an exception, but actually complains that `DocumentId` is not found or defined as a shadow property

```c#
    class Program
    {
        {
            var builder = new DbContextOptionsBuilder();
            var conventionBuilder = new CoreConventionSetBuilder();
            var modelBuilder = new ModelBuilder(
                conventionBuilder.CreateConventionSet()
                );

            modelBuilder.Entity<Document>(x =>
            {
                // I want the DocumentId FK on the PersonByName table
                x.HasOne(typeof(PersonByName), "PersonByName")
                    .WithOne()
                    .ForeignKey(typeof(PersonByName), "DocumentId");
            });

            modelBuilder.Entity<PersonByName>(x =>
            {
                // Defines shadow property DocumentId explicitely
                x.Property(typeof(int), "DocumentId");
            });

            modelBuilder.Validate();

            builder.UseModel(modelBuilder.Model);
            var dbFileName = Path.GetTempFileName();
            builder.UseSqlite(@"Data Source=" + dbFileName + ";Cache=Shared");

            // Ensure the database is created
            var dbContext = new DbContext(builder.Options);
            dbContext.Database.EnsureCreated();

            try {
                var p1 = new PersonByName { Name = "Bill" };
                var d1 = new Document { PersonByName = p1 };
                dbContext.Set<PersonByName>().Add(p1);
                dbContext.Set<Document>().Add(d1);
                dbContext.SaveChanges();

                IQueryable<Document> documents = dbContext.Set<Document>();
                Console.WriteLine(1 == documents.Count());

                Console.WriteLine(null != documents
                    .Where(x => x.PersonByName.Name == "Bill").FirstOrDefault());

               
            }
            finally
            {
                dbContext.Dispose();
                File.Delete(dbFileName);
            }
        }
    }

    public class Document
    {
        public int Id { get; set; }
        public PersonByName PersonByName { get; set; }
    }

    public class PersonByName
    {
        public int Id { get; set; }
    }
```

smitpatel:
fixes #3576

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -870,7 +870,10 @@ namespace Microsoft.Data.Entity.Metadata.Internal
             var principalType = ModelBuilder.Entity(principalEntityTypeName, configurationSource);
             return principalType == null
                 ? null
-                : HasForeignKeyInternal(principalType, GetOrCreateProperties(propertyNames, configurationSource), configurationSource);
+                : HasForeignKeyInternal(
+                    principalType,
+                    GetOrCreateProperties(propertyNames, configurationSource, principalType.Metadata.FindPrimaryKey()?.Properties),
+                    configurationSource);
         }
 
         public virtual InternalRelationshipBuilder HasForeignKey(
@@ -1259,7 +1262,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
             }
         }
 
-        public virtual IReadOnlyList<Property> GetOrCreateProperties([CanBeNull] IEnumerable<string> propertyNames, ConfigurationSource configurationSource)
+        public virtual IReadOnlyList<Property> GetOrCreateProperties([CanBeNull] IEnumerable<string> propertyNames, ConfigurationSource configurationSource, [CanBeNull] IEnumerable<Property> referencedProperties = null)
         {
             if (propertyNames == null)
             {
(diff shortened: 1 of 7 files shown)


src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -861,9 +861,25 @@ private static IndexBuildersSnapshot DetachIndexes(IEnumerable<Index> indexesToD
...
+                var newRelationshipBuilder = relationshipBuilder.HasForeignKey(propertyNames, configurationSource);
+                if (newRelationshipBuilder == null)
+                {
+                    RemoveForeignKey(relationshipBuilder.Metadata, configurationSource);
+                    return null;
+                }
+                return newRelationshipBuilder;
+            }
+
+            return HasForeignKeyInternal(principalType, GetOrCreateProperties(propertyNames, configurationSource), configurationSource);
```
AndriySvyryd: You could unify this new code with the code in `HasForeignKey` by moving it to `GetOrCreateProperties` and adding a parameter of `IEnumerable<Type>` to it containing the principal properties CLR types

AndriySvyryd:
:shipit:

src/EntityFramework.Core/Metadata/Internal/InternalEntityTypeBuilder.cs
smitpatel: Should we create properties when the principal key is shadow?

AndriySvyryd: I think we decided we wouldn't. You can pass a null if the corresponding principal properties is shadow.

smitpatel:
@AndriySvyryd - Can use another look at the PR.

src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -902,11 +902,23 @@ private bool CanInvert(ConfigurationSource configurationSource)
...
+            var dependentEntityTypeBuilder = ModelBuilder.Entity(Metadata.DeclaringEntityType.Name, configurationSource);
+            var principalProperties = Metadata.PrincipalKey.Properties;
+
+            var properties = dependentEntityTypeBuilder
+                .GetOrCreateProperties(
+                    propertyNames,
+                    configurationSource,
+                    principalProperties.Count != propertyNames.Count
+                    ? null
+                    : principalProperties.Select(p => p.IsShadowProperty ? null : p.ClrType));
```
AndriySvyryd: It looks like you can further reduce duplication by just passing through the principal properties and moving the filtering logic to `GetOrCreateProperties`

AndriySvyryd:
Looks neat
:shipit:

PR #3777, opened 2015-11-17
Use generic snapshot types to reduce heap allocations and total memory used when tracking

ajcvickers:
Previously each original value and relationship snapshot value was stored as an object, which for reference types meant boxing and a heap allocation. Now value types are stored directly in the snapshot without boxing or extra heap allocation. This also sets us up for faster code in places like DetectChanges where we can gen type specific differs that will avoid boxing.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/SnapshotFactoryFactory.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/SnapshotFactoryFactory.cs
@@ -1,0 +1,150 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Metadata.Internal;
+
+namespace Microsoft.Data.Entity.ChangeTracking.Internal
+{
+    public abstract class SnapshotFactoryFactory
+    {
+        public virtual Func<InternalEntityEntry, ISnapshot> Create([NotNull] IEntityType entityType)
+        {
+            var count = GetPropertyCount(entityType);
+
+            if (count == 0)
+            {
+                return e => Snapshot.Empty;
+            }
+
+            var entryParameter = Expression.Parameter(typeof(InternalEntityEntry), "entry");
+
+            var types = new Type[count];
+            var propertyBases = new IPropertyBase[count];
+
+            foreach (var propertyBase in entityType.GetPropertiesAndNavigations())
+            {
+                var index = GetPropertyIndex(propertyBase);
+
+                if (index >= 0)
+                {
+                    types[index] = (propertyBase as IProperty)?.ClrType ?? typeof(object);
+                    propertyBases[index] = propertyBase;
+                }
+            }
+
+            Expression constructorExpression;
+            if (count > Snapshot.MaxGenericTypes)
+            {
+                var snapshotExpressions = new List<Expression>();
+
+                for (var i = 0; i < count; i += Snapshot.MaxGenericTypes)
+                {
+                    snapshotExpressions.Add(
+                        CreateSnapshotExpression(
+                            entityType.ClrType,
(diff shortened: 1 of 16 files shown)


smitpatel:
`snapshot.cs` file. :fearful: 
I am not reviewing, I was just having a peek at changes.

natemcmaster:
:scream: .NET Native is not going to like `typeof(Snapshot<,,,,,,,,,,,,,,,,,,,,,,,,,,,,,>).MakeGenericType(types);`

anpete:
@natemcmaster Haha, yep! Code gen?

src/EntityFramework.Core/ChangeTracking/Internal/SnapshotFactoryFactory.cs
```diff
@@ -0,0 +1,162 @@
...
+        [UsedImplicitly]
+        private static HashSet<object> SnapshotCollection(IEnumerable collection)
+        {
+            if (collection == null)
+            {
+                return null;
+            }
+
+            var snapshot = new HashSet<object>(ReferenceEqualityComparer.Instance);
+            foreach (var product in collection)
```
anpete: Use ctor that takes a collection and a comparer.

divega:
@anpete @natemcmaster if it doesn't make the compiler crash :crystal_ball:

natemcmaster:
@anpete If only it were a matter of code-gen. The real issue here is trying to infer all the possible permutations of `Snapshot<T0, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, T23, T24, T25, T26, T27, T28, T29>`. ILC's inference cough, spits, and dies after trying to infer just 4 generic types. I think this is a nail in the coffin for trying to get perfect runtime-type inference.

Can we still make .NET Native still work? Likely yes, but this means we can no longer expect users will not need to add custom `rd.xml` directives against our internal types.

divega:
We may need to consider #ifs and keeping a plain implementation for .NET Native in this case.

anpete:
@natemcmaster @divega Perhaps we need to strategize the snapshotting approach? We could fallback to the old approach on native.

divega:
Yep. Strategize but at compile time :smile:

anpete:
@ajcvickers Can you make the non-generic Snapshot ctor private?

anpete:
@divega Yep, but could just be user configured, right?

anpete:
@ajcvickers Is case 0 in CreateSnapshotType ever hit?

divega:
Yeah, I think I might be wrong. My mental model is not crips enough about what works and what doesn't. E.g. .NET Native compiler may actually not mind the Snapshot<...> types but may not be capable of generating code that will actually instantiate them, so as long as we never tried to do it at runtime for .NET Native we might be ok without #ifs. @natemcmaster of course is the expert.

ajcvickers:
@anpete  Re-non-generic snapshot constructor can be made private. It used to be hit in case 0 mentioned by @divega, but then I added in a short-circuit, so it doesn't get hit now.

anpete:
:shipit:

ajcvickers:
@divega @natemcmaster @rowanmiller I think you guys know my opinion on this (we're writing valid, good, .NET code with measurable reasons for doing so, and if .NET native doesn't work then it should be fixed) but if you guys want to fork the code in some way to workaround the fact that .NET native doesn;t work, then we should come up with a pattern for this. We should keep testability in mind--whatever we come up with should allow us to run pretty much all our tests with both approaches otherwise bugs are likely to creep in on the fork that isn't tested.

ajcvickers:
@divega @natemcmaster @rowanmiller Also, keep in mind that I plan to use the fact that we have this type of snapshot in other places in the code--e.g. DetectChanges and fixup--and also use this pattern for shadow values for the same reasons.

src/EntityFramework.Core/ChangeTracking/Internal/SnapshotFactoryFactory.cs
ajcvickers: Done.

divega:
FYI @natemcmaster confirmed that this breaks .NET Native support. 

The .NET Native compiler can be improved to fail more gracefully but I don't think an approach that requires generating all the possible snapshot types would ever work. There are just too many. We either need another approach that can generate at compile time only the types we will use or to maintain an alternate code path that doesn't use the snapshot types.

PR #3780, opened 2015-11-18
Migrations: Fix some scaffolding bugs

### LINKED ISSUE #3388: Migrations: Wrong migration generates when DbContext implementation is generic
Subject.
If I try dnx . ef migrations add with generic context implementation, model snapshot generate wrong attribute
[DbContext(typeof(SiteContext`17))]

bricelam:
Highlights:
* Stops escaping contextual C# keywords out of context
* Handle generic types (fixes #3388)

### DIFF
--- a/src/EntityFramework.Commands/Migrations/Design/CSharpHelper.cs
+++ b/src/EntityFramework.Commands/Migrations/Design/CSharpHelper.cs
@@ -40,101 +40,60 @@ namespace Microsoft.Data.Entity.Migrations.Design
             "__reftype",
             "__refvalue",
             "abstract",
-            "add",
-            "alias",
             "as",
-            "ascending",
-            "assembly",
-            "async",
-            "await",
             "base",
             "bool",
             "break",
-            "by",
             "byte",
             "case",
             "catch",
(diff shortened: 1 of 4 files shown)


ajcvickers:
:shipit:

src/EntityFramework.Commands/Migrations/Design/CSharpHelper.cs
```diff
@@ -40,101 +40,60 @@ public class CSharpHelper
             "__reftype",
             "__refvalue",
             "abstract",
-            "add",
-            "alias",
             "as",
-            "ascending",
-            "assembly",
-            "async",
-            "await",
```
divega: Careful with await. It is contextual but is in context in many places where the others aren't.

bricelam: The only things we currently generate identifiers for (using this code) are class names in their definition and inside `typeof`.

divega: Sounds great then.

PR #3783, opened 2015-11-18
Move SQL Server concepts into provider-specific scaffolding model

natemcmaster:
Reverting a design decision we made for RC1. 

Use inheritance instead of making the database model a grab-bag of features from multiple providers. Fix https://github.com/aspnet/EntityFramework/issues/3766.

This pattern made it easier to resolve a quirk scaffolding "DateTimePrecision" on SqlServer. (see https://github.com/aspnet/EntityFramework/issues/3454).

### DIFF
--- a/src/EntityFramework.MicrosoftSqlServer.Design/EntityFramework.MicrosoftSqlServer.Design.csproj
+++ b/src/EntityFramework.MicrosoftSqlServer.Design/EntityFramework.MicrosoftSqlServer.Design.csproj
@@ -43,6 +43,8 @@
       <DesignTime>True</DesignTime>
       <DependentUpon>Resources.tt</DependentUpon>
     </Compile>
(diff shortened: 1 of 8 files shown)


bricelam:
We may need another design meeting...

natemcmaster:
FWIW provider-writers can already use inheritance to get around missing fields in `DatabaseModel`.

roji:
Thanks for this, I'll be integrating some PostgreSQL-specific metadata into the scaffolding meeting soon.

natemcmaster:
Ping for review. Per design-meeting discussion, I tested an implementation of this functional change using `Annotatable` instead, but found inheritance to be more straightforward. Furthermore, `Annotatable` has down-the-road complications as we try to use EF to materialize `DatabaseModel`.

lajones:
Looks good to me :shipit:

PR #3785, opened 2015-11-18
Stop taking key from query

ajcvickers:
State manager doesn't need to take key from query since it can be obtained from the entity


anpete:
:shipit:

PR #3789, opened 2015-11-18
Use generic snapshots for shadow values

ajcvickers:
Improves the perf of creating shadow value storage and storage uses less memory.

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/ISnapshotFactorySource.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/ISnapshotFactorySource.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.Data.Entity.ChangeTracking.Internal;
+using Microsoft.Data.Entity.Storage;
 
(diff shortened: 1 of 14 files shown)


src/EntityFramework.Core/Metadata/Internal/ISnapshotFactorySource.cs
```diff
@@ -10,5 +11,7 @@ public interface ISnapshotFactorySource
     {
         Func<InternalEntityEntry, ISnapshot> OriginalValuesFactory { get; }
         Func<InternalEntityEntry, ISnapshot> RelationshipSnapshotFactory { get; }
+        Func<ValueBuffer, ISnapshot> ShadowValuesFactory { get; }
+        Func<ISnapshot> EmptyShadowValuesFactory { get; }
```
AndriySvyryd: Why not return and cache the empty ISnapshot itself?

AndriySvyryd:
:shipit:

PR #3800, opened 2015-11-19
Use LogSensitiveData flag to limit instance data in exception and log messages

mikary:
Fix for exception messages identified by #2752.
Promote `ISensitiveDataLogger` to core.


src/EntityFramework.Relational/Storage/RelationalConnection.cs
```diff
@@ -217,8 +222,13 @@ public virtual void Open()
             {
                 _logger.LogVerbose(
                     RelationalLoggingEventId.OpeningConnection,
-                    ConnectionString,
-                    RelationalStrings.RelationalLoggerOpeningConnection);
+                    new RelationalConnectionLogData(
```
lajones: Nit. This seems like an odd way to do it, but if you can't pass `_connection.Value.Database` and `_connection.Value.DataSource` directly then OK.

src/EntityFramework.Core/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -18,25 +19,30 @@ public class ParameterExtractingExpressionVisitor : ExpressionVisitorBase
         public static Expression ExtractParameters(
             [NotNull] Expression expression,
             [NotNull] QueryContext queryContext,
-            [NotNull] IEvaluatableExpressionFilter evaluatableExpressionFilter)
+            [NotNull] IEvaluatableExpressionFilter evaluatableExpressionFilter,
+            [NotNull] ISensitiveDataLogger logger)
```
lajones: Re the [NotNull], what gets passed here if `SensitiveDataLoggingEnabled` is false?

mikary: `ISensitiveDataLogger` will exist either way, the flag is just a setting from `DbContextOptions` that can be accessed through this interface. Checking the value also logs a warning if sensitive data logging is enabled.

src/EntityFramework.Relational/Storage/RelationalConnection.cs
mikary: Could work with an anonymous type. We use a strong type for LogCommand (`DbCommandLogData`) which is a public type. Not sure if there is a preference / pattern we want to establish for these.

anpete: Use a nominal type when it needs to be consumable. E.g. from tests.

src/EntityFramework.Relational/Storage/RelationalConnection.cs
```diff
@@ -270,5 +285,19 @@ public virtual void Dispose()
                 _openedCount = 0;
             }
         }
+
+        private class RelationalConnectionLogData
```
anpete: This is either public or use an anon type.

src/EntityFramework.Core/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
lajones: OK. No problem then.

lajones:
LGTM :shipit:

PR #3808, opened 2015-11-19
Query: Simplifications post @ajcvickers removal of StartTracking key …

anpete:
…requirement

- Removed entity key requirement from EntityTrackingInfo
- Removed QueryBuffer requirement for simple tracking queries.


ajcvickers:
:shipit:

Issue #3758, opened 2015-11-14
Including Many-To-Many Relation in Select


chbach:
Hi, 

I have a many to many relationship between my models `NewsEntry` and `Tag` via the "pivot" model `Taggings`. On the index view of my NewsEntry page I don't want to include all columns of `NewsEntry` in my query result. However, it somehow throws an error when selecting the `Taggings` relation. Here is the working version including all fields: 

```c#
var query = Db.NewsEntries.Include(e => e.Taggings);
```

Producing the following query: 
```sql
SELECT "t"."NewsEntryId", "t"."TagId"
FROM "Tagging" AS "t"
INNER JOIN (
    SELECT DISTINCT "t0".*
    FROM (
        SELECT "e"."Id", "e"."Abstract", "e"."Date", "e"."Image", "e"."Querystring", "e"."Text", "e"."Title", "e"."URL"
        FROM "NewsEntry" AS "e"
        ORDER BY "e"."Date" DESC, "e"."Id"
        LIMIT 12 OFFSET 0
    ) AS "t0"
) AS "e" ON "t"."NewsEntryId" = "e"."Id"
```

Then I tried to map the result to a smaller version of the object like this: 

```c#
var entries = query.Select(e => new NewsEntry { 
    Id = e.Id,
    Title = e.Title,
    URL = e.URL,
    Taggings = e.Taggings,
    Date = e.Date
});
```

Producing the following error: 

```txt
ArgumentException: Argument types do not match
System.Linq.Expressions.Expression.Bind (System.Reflection.MemberInfo member, System.Linq.Expressions.Expression expression) <0x5aeac80 + 0x000bf> in <filename unknown>, line 0
```

If I don't include the taggings in the select, everything works as expected.

I found out that e.Taggings somehow still only includes `int` values (after converting it to a list) and hasn't been populated with the Tagging objects. My guess is the query returns just a row containing the TagId and EF later automagically populates the relation, so I would have to select the TagId. But my model does not contain a field for TagId. Is there some way around it or do I have to select all columns for now?

Thanks for your help, 

Christoph

maumar:
@chbach could you provide full code listing of the entities used as well as DBContext? We have made some bug fixes in this area recently and with functional repro I could quickly check whether this issue has been addressed already

chbach:
Thanks for your quick reply. Here are the classes:

Models: 

```c#
public class NewsEntry
{
    public int Id { get; set; }
    public String Title { get; set; }
[...]

PR #3812, opened 2015-11-20
Fix to #3758 - Including Many-To-Many Relation in Select

maumar:
Problem was that we were making some incorrect assumptions when materializing collections directly - we always assumed there will be a subquery present in those cases, and also we were not materializing correct collection types.
Fix is to inject subquery to navigation access, if the subquery was not present before. Also added a client-side method that materializes a correct collection, based on the information in the INavigation.

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/ClrCollectionAccessorFactory.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/ClrCollectionAccessorFactory.cs
@@ -18,6 +18,9 @@ namespace Microsoft.Data.Entity.Metadata.Internal
         private static readonly MethodInfo _createAndSet
             = typeof(ClrCollectionAccessorFactory).GetTypeInfo().GetDeclaredMethods("CreateAndSet").Single();
 
+        private static readonly MethodInfo _create
+            = typeof(ClrCollectionAccessorFactory).GetTypeInfo().GetDeclaredMethods("CreateCollection").Single();
+
         public virtual IClrCollectionAccessor Create([NotNull] INavigation navigation)
         {
             var accessor = navigation as IClrCollectionAccessor;
@@ -65,6 +68,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
 
             Action<TEntity, TCollection> setterDelegate = null;
             Func<TEntity, Action<TEntity, TCollection>, TCollection> createAndSetDelegate = null;
+            Func<TCollection> createDelegate = null;
 
             var setter = property.SetMethod;
             if (setter != null)
@@ -78,11 +82,15 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                     createAndSetDelegate = (Func<TEntity, Action<TEntity, TCollection>, TCollection>)_createAndSet
                         .MakeGenericMethod(typeof(TEntity), typeof(TCollection), concreteType)
                         .CreateDelegate(typeof(Func<TEntity, Action<TEntity, TCollection>, TCollection>));
+
+                    createDelegate = (Func<TCollection>)_create
+                        .MakeGenericMethod(typeof(TCollection), concreteType)
+                        .CreateDelegate(typeof(Func<TCollection>));
                 }
             }
 
             return new ClrICollectionAccessor<TEntity, TCollection, TElement>(
-                property.Name, getterDelegate, setterDelegate, createAndSetDelegate);
+                property.Name, getterDelegate, setterDelegate, createAndSetDelegate, createDelegate);
         }
 
         // ReSharper disable once UnusedMember.Local
@@ -96,6 +104,14 @@ namespace Microsoft.Data.Entity.Metadata.Internal
             var collection = new TConcreteCollection();
             setterDelegate(entity, collection);
             return collection;
+        }
+
+        // ReSharper disable once UnusedMember.Local
+        private static TCollection CreateCollection<TCollection, TConcreteCollection>()
+            where TCollection : class
+            where TConcreteCollection : TCollection, new()
+        {
+            return new TConcreteCollection();
         }
     }
 }
--- a/src/EntityFramework.Core/Metadata/Internal/ClrICollectionAccessor.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/ClrICollectionAccessor.cs
@@ -16,17 +16,22 @@ namespace Microsoft.Data.Entity.Metadata.Internal
         private readonly Func<TEntity, TCollection> _getCollection;
         private readonly Action<TEntity, TCollection> _setCollection;
         private readonly Func<TEntity, Action<TEntity, TCollection>, TCollection> _createAndSetCollection;
+        private readonly Func<TCollection> _createCollection;
+
+        public virtual Type CollectionType => typeof(TCollection);
 
         public ClrICollectionAccessor(
             [NotNull] string propertyName,
             [NotNull] Func<TEntity, TCollection> getCollection,
             [CanBeNull] Action<TEntity, TCollection> setCollection,
-            [CanBeNull] Func<TEntity, Action<TEntity, TCollection>, TCollection> createAndSetCollection)
+            [CanBeNull] Func<TEntity, Action<TEntity, TCollection>, TCollection> createAndSetCollection,
+            [CanBeNull] Func<TCollection> createCollection)
         {
             _propertyName = propertyName;
             _getCollection = getCollection;
             _setCollection = setCollection;
             _createAndSetCollection = createAndSetCollection;
+            _createCollection = createCollection;
         }
 
         public virtual void Add(object instance, object value)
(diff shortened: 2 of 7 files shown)


maumar:
@anpete  @ajcvickers

src/EntityFramework.Core/Metadata/Internal/ClrICollectionAccessor.cs
```diff
@@ -45,6 +50,26 @@ public virtual void AddRange(object instance, IEnumerable<object> values)
...
+        public virtual object CreateAndSet(IEnumerable<object> values)
+        {
+            var collection = (TCollection)_concreteCollectionType?.GetDeclaredConstructor(null)?.Invoke(null);
+
+            if (collection == null)
+            {
+                throw new InvalidOperationException(CoreStrings.NavigationCannotCreateType(_propertyName, typeof(TEntity).FullName, typeof(TCollection).FullName));
+            }
+
+            foreach (TElement value in values)
```
maumar: worth drying up?

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -33,7 +33,7 @@ public abstract class EntityQueryModelVisitor : QueryModelVisitorBase
         public static readonly MethodInfo PropertyMethodInfo
             = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
 
-        protected virtual IModel Model { get; }
+        public virtual IModel Model { get; }
```
maumar: not happy about this - I need model in the subquery injector to find the EntityType and then INavigation. Alternatively, I could pass it in constructor to the NavigationRewritingExpressionVisitor. Thoughts?

src/EntityFramework.Core/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -591,10 +594,71 @@ public override void VisitOrdering(Ordering ordering, QueryModel queryModel, Ord
...
+                            var mainFromClause = new MainFromClause(targetType.Name.Substring(0, 1).ToLower(), targetType, node);
+                            var selecor = new QuerySourceReferenceExpression(mainFromClause);
+
+                            var subqueryModel = new QueryModel(mainFromClause, new SelectClause(selecor));
+                            var subqueryExpression = new SubQueryExpression(subqueryModel);
+
+                            var resultCollectionType = navigation.GetCollectionAccessor().CollectionType;
+
+                            var result = Expression.Call(
+                                _materializeCollectionNavigationMethodInfo.MakeGenericMethod(targetType);
```
maumar: will change ; into , at the end of 639

src/EntityFramework.Core/Metadata/Internal/ClrCollectionAccessorFactory.cs
```diff
@@ -82,7 +82,7 @@ public virtual IClrCollectionAccessor Create([NotNull] INavigation navigation)
             }
 
             return new ClrICollectionAccessor<TEntity, TCollection, TElement>(
```
anpete: What is the difference between TCollection and concreteType?

src/EntityFramework.Core/Metadata/Internal/IClrCollectionAccessor.cs
```diff
@@ -12,5 +13,7 @@ public interface IClrCollectionAccessor
         void AddRange([NotNull] object instance, [NotNull] IEnumerable<object> values);
         bool Contains([NotNull] object instance, [NotNull] object value);
         void Remove([NotNull] object instance, [NotNull] object value);
+        object CreateAndSet([NotNull] IEnumerable<object> values);
```
anpete: I would just call this Create - analogous to the collection ctor that takes a sequence.

src/EntityFramework.Core/Metadata/Internal/ClrCollectionAccessorFactory.cs
maumar: E.g TCollection = ICollection<Foo> and concrete type = HashSet<Foo>.

src/EntityFramework.Core/Metadata/Internal/IClrCollectionAccessor.cs
maumar: Sounds good

src/EntityFramework.Core/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -591,10 +594,71 @@ public override void VisitOrdering(Ordering ordering, QueryModel queryModel, Ord
...
+                    }
+
+                    return base.VisitMember(node);
+                }
+
+                private static readonly MethodInfo _materializeCollectionNavigationMethodInfo
+                    = typeof(SubqueryInjector).GetTypeInfo()
+                        .GetDeclaredMethod(nameof(MaterializeCollectionNavigation));
+
+                public static ICollection<TEntity> MaterializeCollectionNavigation<TEntity>(INavigation navigation, IEnumerable<object> elements)
```
anpete: private?

src/EntityFramework.Core/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -591,10 +594,71 @@ public override void VisitOrdering(Ordering ordering, QueryModel queryModel, Ord
...
+                {
+                    _model = model;
+                }
+
+                protected override Expression VisitSubQuery(SubQueryExpression expression)
+                {
+                    return expression;
+                }
+
+                protected override Expression VisitMember(MemberExpression node)
```
anpete: You should use one of the bind methods on EntityQueryModelVisitor

src/EntityFramework.Core/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -591,10 +594,71 @@ public override void VisitOrdering(Ordering ordering, QueryModel queryModel, Ord
                 => ordering.TransformExpressions(_parentVisitor.Visit);
 
             public override void VisitSelectClause(SelectClause selectClause, QueryModel queryModel)
-                => selectClause.TransformExpressions(_parentVisitor.Visit);
+            {
+                var newSelector = _subqueryInjector.Visit(selectClause.Selector);
+
+                selectClause.Selector = newSelector;
+
+                selectClause.TransformExpressions(_parentVisitor.Visit);
```
anpete: Why do we revisit here?

maumar:
@anpete @ajcvickers  new version up

src/EntityFramework.Core/Metadata/Internal/ClrICollectionAccessor.cs
```diff
@@ -45,6 +50,26 @@ public virtual void AddRange(object instance, IEnumerable<object> values)
             }
         }
 
+        public virtual object Create(IEnumerable<object> values)
+        {
+            var collection = (TCollection)_concreteCollectionType?.GetDeclaredConstructor(null)?.Invoke(null);
```
ajcvickers: It seems strange that everywhere else in this class we use pre-prepared delegates to do the requested operations rather than using Reflection. Maybe it doesn't matter here because of the way query uses the Create method, but is there some reason not to follow the same pattern as the other methods?

src/EntityFramework.Core/Metadata/Internal/ClrICollectionAccessor.cs
```diff
@@ -45,6 +50,26 @@ public virtual void AddRange(object instance, IEnumerable<object> values)
...
+
+            if (collection == null)
+            {
+                throw new InvalidOperationException(CoreStrings.NavigationCannotCreateType(_propertyName, typeof(TEntity).FullName, typeof(TCollection).FullName));
+            }
+
+            foreach (TElement value in values)
+            {
+                if (!collection.Contains(value))
```
ajcvickers: Is this needed? The collection doesn't exist before the method is called, so it can't already have duplicates. The only reason this check should be needed is if the caller is passing in duplicates. is that something query will do?

maumar: was erring on side of caution, will remove and indeed it *shouldn't* happen

ajcvickers: Should there be a version of AddRange that does not check for duplicates that can be called by query when it knows that it has materialized a new entity and hence the collection is empty, and knows that it won't be passing any duplicates? When people use lists for collections and there are a lot of related entities the duplicate checking can get expensive. /cc @anpete

maumar:
@ajcvickers @anpete  new version up - applied Arthur's feedback

ajcvickers:
Collection accessor stuff looks good.

anpete:
:shipit:

Issue #3811, opened 2015-11-19
.OrderBy on column 2 or more navigation properties away + .Include throws exception


jparish9:
Windows 10, VS 2015, IIS Express, SQL Server Express 2014, EF 7.0.0-rc1-final, ASP.NET 6.0.0-rc1-final

After upgrading to RC1, I am seeing an exception `Unable to cast object of type 'System.Linq.Expressions.FieldExpression' to type 'System.Linq.Expressions.ParameterExpression'.` when querying using `.OrderBy` on a navigation property table 2 or more tables away from the base table, plus `.Include`.  This was working in beta8.  I reduced the test case as much as possible to illustrate:

Models:

``` c#
    public class Blog
    {
        public int BlogId { get; set; }

        public string Title { get; set; }
    }
```

``` c#
    public class Post
    {
        public int PostId { get; set; }

        public int BlogId { get; set; }

        public Blog Blog { get; set; }

        public string PostText { get; set; }
    }
```

``` c#
    public class Comment
    {
        public int CommentId { get; set; }

        public int PostId { get; set; }

        public Post Post { get; set; }

        public int CommentText { get; set; }
    }
```

The query that causes an exception is:

``` c#
   _dbContext.Comments
                    .Include(c => c.Post)
                    .OrderBy(c => c.Post.Blog.BlogId)
                    .ToList();
```

Stack trace:

```
Unable to cast object of type 'System.Linq.Expressions.FieldExpression' to type 'System.Linq.Expressions.ParameterExpression'.
   at Microsoft.Data.Entity.Query.ExpressionVisitors.Internal.IncludeExpressionVisitor.VisitMethodCall(MethodCallExpression expression)
   at System.Linq.Expressions.MethodCallExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
   at Microsoft.Data.Entity.Query.ExpressionVisitors.ExpressionVisitorBase.Visit(Expression expression)
   at System.Linq.Expressions.ExpressionVisitor.VisitArguments(IArgumentProvider nodes)
   at System.Linq.Expressions.ExpressionVisitor.VisitMethodCall(MethodCallExpression node)
   at Microsoft.Data.Entity.Query.ExpressionVisitors.Internal.IncludeExpressionVisitor.VisitMethodCall(MethodCallExpression expression)
   at System.Linq.Expressions.MethodCallExpression.Accept(ExpressionVisitor visitor)
[...]

PR #3819, opened 2015-11-20
[SMALL] Fix to #3811 - .OrderBy on column 2 or more navigation properties away + .Include throws exception

maumar:
Problem was that in IncludeExpressionVisitor -> VisitMethodCall, when building entity accessor from a MemberExpression, we incorrectly assumed that member expression will always be one level.

Fix is to add a method that will go down the Expression chain to find a ParameterExpression which is the root of the given MemberExpression

### DIFF
--- a/src/EntityFramework.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors.Internal
                             : Expression
                                 .Lambda(
                                     memberExpression,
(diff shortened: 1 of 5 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -112,6 +112,17 @@ var entityAccessor
             return base.VisitMethodCall(node);
         }
 
+        private ParameterExpression FindRootParameterExpression(MemberExpression memberExpression)
```
anpete: I think we have code like this already (perhaps in Bind) can you unify if so?

maumar:
@anpete  new version up

anpete:
:shipit:

PR #3827, opened 2015-11-20
[SMALL] RelationalCommand: Less closures and awaits.

anpete:
- Decompiled code size goes from 887 loc to 544.

### DIFF
--- a/src/EntityFramework.Relational/Storage/Internal/RelationalCommand.cs
+++ b/src/EntityFramework.Relational/Storage/Internal/RelationalCommand.cs
@@ -46,31 +46,17 @@ namespace Microsoft.Data.Entity.Storage.Internal
             bool manageConnection = true)
             => Execute(
                 Check.NotNull(connection, nameof(connection)),
-                (cmd, con) =>
-                    {
-                        using (cmd)
-                        {
-                            return cmd.ExecuteNonQuery();
-                        }
-                    },
-                RelationalDiagnostics.ExecuteMethod.ExecuteNonQuery,
+                nameof(ExecuteNonQuery),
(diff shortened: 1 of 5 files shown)


ajcvickers:
:shipit:

src/EntityFramework.Relational/Storage/Internal/RelationalCommand.cs
```diff
@@ -142,34 +97,17 @@ public class RelationalCommand : IRelationalCommand
...
-
-        protected virtual T Execute<T>(
+            => (RelationalDataReader)
+                await ExecuteAsync(
+                    Check.NotNull(connection, nameof(connection)),
+                    nameof(ExecuteReader),
+                    openConnection: manageConnection,
+                    closeConnection: false,
+                    cancellationToken: cancellationToken,
+                    parameterValues: parameterValues);
```
AndriySvyryd: You can also remove this await using this extension method:
```
        public static Task<TDerived> Cast<T, TDerived>(this Task<T> task)
            where TDerived : T
        {
            var tcs = new TaskCompletionSource<TDerived>();
            task.ContinueWith(t =>
                {
                    if (t.IsFaulted)
                        tcs.TrySetException(t.Exception.InnerExceptions);
                    else if (t.IsCanceled)
                        tcs.TrySetCanceled();
                    else
                        tcs.TrySetResult((TDerived)t.Result);
                }, TaskContinuationOptions.ExecuteSynchronously);
            return tcs.Task;
        }
```

anpete: Nice, will use that.

PR #3828, opened 2015-11-20
Remove dependent map from state manager

ajcvickers:
We were previously keeping track of all dependent entries referenced from a principal such that lookups for these were fast. However, we are currently only using these in three cases. In two of these, if the principal has a navigation property to the dependent(s), then we use that instead. Otherwise we fall back to a linear scan of all entries, which seems to be okay for when we need it. This allows the dependent map to be removed which has a big impact on initial query perf and reduces the amount of storage needed to track entities.


AndriySvyryd:
:shipit:

PR #3833, opened 2015-11-21
:bug: Handle Identity in KeyConvention properly

### LINKED ISSUE #3817: Identity set for foreign key property
EF RC2-16378
Issue here is, we find both navigations and create relationship by convention using `BlogId` as fk property. When we configure relationship explicitly the navigation `Blog` is removed from relationship, conventions re-discover it and while configuring it assigns `BlogId` as foreignkey. But after this foreign key removed conventions run on previous relationship building which sets identity for `BlogId` even though it is part of another relationship. This causes exception
`Cascading foreign key 'FK_BlogMetadata_Blog_BlogId' cannot be created where the referencing column 'BlogMetadata.BlogId' is an identity column.`

repro:
````C#
public class Blog
    {

        public virtual BlogMetadata BlogMetadata { get; set; }
    }

    public class BlogMetadata
    {
        [Key, ForeignKey("Blog")]

        public string Metadata { get; set; }

        public virtual Blog Blog { get; set; }
    }

        {
            modelBuilder.Entity<Blog>().HasOne(b => b.BlogMetadata).WithOne();
        }

smitpatel:
Resolves #3817

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/KeyConvention.cs
@@ -37,13 +37,7 @@ namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
         {
             var properties = foreignKey.Properties;
             SetValueGeneration(properties.Where(property => property.IsKey()));
-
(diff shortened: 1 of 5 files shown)


ajcvickers:
:shipit:

PR #3853, opened 2015-11-23
[SMALL] Fix #3773 - Introduce enum to control query client-eval behavior.

### LINKED ISSUE #3773: Query: Swap client evaluation settings to allow/throw/warn (rather than just allow/throw)
Currently we have a flag to enable/disable client eval in queries. Enabling it writes a warning to the log each time something is client evaluated. We should swap to an enum so that folks can opt out of the warnings.

anpete:


divega:
:shipit:

PR #3869, opened 2015-11-24
Use OUTPUT INTO for generated values in SqlServer to support tables with triggers

### LINKED ISSUE #1441: Support for inserting/updating tables with triggers (not use OUTPUT when batching off)
Let there be a table `Table1` in SQL Server with an AFTER INSERT trigger. The produced SQL is:

    INSERT INTO [Table1] ([Column1])
    OUTPUT INSERTED.[Id], INSERTED.[ComputedColumn]
    VALUES (@p0);
    ',N'@p0 int',@p0=1

The problem is that SQL reports the following error for such a statement if a trigger exists on the table:
*The target table 'Table1' of the DML statement cannot have any enabled triggers if the statement contains an OUTPUT clause without INTO clause.*

I have been trying a possible fix where the following code is changed for `AppendBulkInsertOperation()` in SqlServerSqlGenerator.cs:

                var operations = modificationCommands[i].ColumnModifications;
                var writeOperations = operations.Where(o => o.IsWrite).ToArray();
                var readOperations = operations.Where(o => o.IsRead).ToArray();

                if (readOperations.Length > 0)
                {
                    commandStringBuilder
                        .Append("DECLARE @inserted TABLE (")
                        .AppendJoin(readOperations.Select(c => string.Format("{0} {1}",
                                DelimitIdentifier(c.ColumnName),
                                typeMapper.GetTypeMapping(c.Property.SqlServer().ColumnType, c.Property.SqlServer().Column, c.Property.UnderlyingType, false, false).StoreTypeName
                            )))
                        .Append(")")
                        .Append(BatchCommandSeparator)
                        .AppendLine();
                }

                AppendInsertCommandHeader(commandStringBuilder, schemaQualifiedName, writeOperations);
                if (readOperations.Length > 0)
                {
                    AppendOutputClause(commandStringBuilder, readOperations);
                    commandStringBuilder
                        .AppendLine()
                        .Append("INTO @inserted");
                }
                AppendValuesHeader(commandStringBuilder, writeOperations);
                AppendValues(commandStringBuilder, writeOperations);
                for (var j = 1; j < valueSetCount; j++)
                {
                    commandStringBuilder.Append(",").AppendLine();
                    AppendValues(commandStringBuilder, modificationCommands[j].ColumnModifications.Where(o => o.IsWrite).ToArray());
                }
                commandStringBuilder.Append(BatchCommandSeparator).AppendLine();

                if (readOperations.Length > 0)
                {
                    commandStringBuilder
                        .Append("SELECT ")
                        .AppendJoin(readOperations.Select(c => DelimitIdentifier(c.ColumnName)))
                        .Append(" FROM @inserted")
                        .Append(BatchCommandSeparator)
                        .AppendLine();
                }
                else
                {
                    AppendSelectAffectedCountCommand(commandStringBuilder, schemaQualifiedName);
                }

The following query is then produced, and it seems to work as expected. I am not sure about the `GetTypeMapping()` on the typeMapper.

    DECLARE @inserted table (Id int, ComputedColumn int)
    INSERT INTO [Table1] ([Column1])
    OUTPUT INSERTED.[Id], INSERTED.[ComputedColumn]
    INTO @inserted
    VALUES (@p0)
    SELECT * FROM @inserted;
    ',N'@p0 int',@p0=1

AndriySvyryd:
Add commandPosition argument to methods on IUpdateSqlGenerator to provide a unique id for a modification command in a batch

Fixes #1441

### DIFF
--- a/src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
+++ b/src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
@@ -1,10 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Storage;
 using Microsoft.Data.Entity.Utilities;
 
@@ -12,23 +14,26 @@ namespace Microsoft.Data.Entity.Update.Internal
 {
     public class SqlServerUpdateSqlGenerator : UpdateSqlGenerator, ISqlServerUpdateSqlGenerator
     {
-        public SqlServerUpdateSqlGenerator([NotNull] ISqlGenerator sqlGenerator)
+        private readonly IRelationalTypeMapper _typeMapper;
+
+        public SqlServerUpdateSqlGenerator([NotNull] ISqlGenerator sqlGenerator,
+            [NotNull] IRelationalTypeMapper typeMapper)
             : base(sqlGenerator)
         {
+            _typeMapper = typeMapper;
         }
 
-        public override void AppendInsertOperation(
-            StringBuilder commandStringBuilder,
-            ModificationCommand command)
+        public override void AppendInsertOperation(StringBuilder commandStringBuilder, ModificationCommand command, int commandPosition)
         {
             Check.NotNull(command, nameof(command));
 
-            AppendBulkInsertOperation(commandStringBuilder, new[] { command });
+            AppendBulkInsertOperation(commandStringBuilder, new[] { command }, commandPosition);
         }
 
         public virtual ResultsGrouping AppendBulkInsertOperation(
             StringBuilder commandStringBuilder,
-            IReadOnlyList<ModificationCommand> modificationCommands)
+            IReadOnlyList<ModificationCommand> modificationCommands,
+            int commandPosition)
         {
             Check.NotNull(commandStringBuilder, nameof(commandStringBuilder));
             Check.NotEmpty(modificationCommands, nameof(modificationCommands));
@@ -51,10 +56,15 @@ namespace Microsoft.Data.Entity.Update.Internal
                 var writeOperations = operations.Where(o => o.IsWrite).ToArray();
                 var readOperations = operations.Where(o => o.IsRead).ToArray();
 
+                if (readOperations.Length > 0)
+                {
+                    AppendDeclareGeneratedTable(commandStringBuilder, readOperations, commandPosition);
+                }
+
                 AppendInsertCommandHeader(commandStringBuilder, name, schema, writeOperations);
                 if (readOperations.Length > 0)
                 {
-                    AppendOutputClause(commandStringBuilder, readOperations);
+                    AppendOutputClause(commandStringBuilder, readOperations, commandPosition);
                 }
                 AppendValuesHeader(commandStringBuilder, writeOperations);
                 AppendValues(commandStringBuilder, writeOperations);
@@ -65,9 +75,13 @@ namespace Microsoft.Data.Entity.Update.Internal
                 }
                 commandStringBuilder.Append(SqlGenerator.BatchCommandSeparator).AppendLine();
 
-                if (readOperations.Length == 0)
+                if (readOperations.Length > 0)
                 {
-                    AppendSelectAffectedCountCommand(commandStringBuilder, name, schema);
+                    AppendSelectGeneratedCommand(commandStringBuilder, readOperations, commandPosition);
+                }
+                else
+                {
+                    AppendSelectAffectedCountCommand(commandStringBuilder, name, schema, commandPosition);
                 }
             }
 
@@ -76,9 +90,7 @@ namespace Microsoft.Data.Entity.Update.Internal
                 : ResultsGrouping.OneResultSet;
         }
 
-        public override void AppendUpdateOperation(
-            StringBuilder commandStringBuilder,
-            ModificationCommand command)
+        public override void AppendUpdateOperation(StringBuilder commandStringBuilder, ModificationCommand command, int commandPosition)
         {
             Check.NotNull(commandStringBuilder, nameof(commandStringBuilder));
             Check.NotNull(command, nameof(command));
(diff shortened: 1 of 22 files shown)


src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
```diff
@@ -91,30 +103,80 @@ public SqlServerUpdateSqlGenerator([NotNull] ISqlGenerator sqlGenerator)
...
+                AppendSelectAffectedCountCommand(commandStringBuilder, name, schema, commandPosition);
             }
         }
 
+        private void AppendDeclareGeneratedTable(StringBuilder commandStringBuilder, ColumnModification[] readOperations, int commandPosition)
+        {
+            commandStringBuilder
+                .Append($"DECLARE @generated{commandPosition} TABLE (")
+                .AppendJoin(readOperations.Select(c =>
+                    $"{SqlGenerator.DelimitIdentifier(c.ColumnName)} {GetTypeNameForCopy(c.Property)}"))
```
divega: Beware that @bricelam hates this :smile:

src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
```diff
@@ -91,30 +103,80 @@ public SqlServerUpdateSqlGenerator([NotNull] ISqlGenerator sqlGenerator)
...
+            else
+            {
+                AppendSelectAffectedCountCommand(commandStringBuilder, name, schema, commandPosition);
             }
         }
 
+        private void AppendDeclareGeneratedTable(StringBuilder commandStringBuilder, ColumnModification[] readOperations, int commandPosition)
+        {
+            commandStringBuilder
+                .Append($"DECLARE @generated{commandPosition} TABLE (")
```
divega: Cool!

src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
```diff
@@ -91,30 +103,80 @@ public SqlServerUpdateSqlGenerator([NotNull] ISqlGenerator sqlGenerator)
...
+
+        private string GetTypeNameForCopy(IProperty property)
+        {
+            var mapping = _typeMapper.GetMapping(property);
+            var typeName = mapping.DefaultTypeName;
+            if (property.IsConcurrencyToken
+                && (typeName.Equals("rowversion", StringComparison.OrdinalIgnoreCase)
+                    || typeName.Equals("timestamp", StringComparison.OrdinalIgnoreCase)))
+            {
+                return property.IsNullable ? "varbinary(8)" : "binary(8)";
```
divega: Would it be better if even this lived in the type mapper for SQL Server?

test/EntityFramework.MicrosoftSqlServer.FunctionalTests/DataAnnotationSqlServerTest.cs
```diff
@@ -56,9 +56,12 @@ public override void DatabaseGeneratedAttribute_autogenerates_values_when_set_to
 @p2: 00000000-0000-0000-0000-000000000003
 
 SET NOCOUNT OFF;
+DECLARE @generated1 TABLE ([UniqueNo] int);
 INSERT INTO [Sample] ([MaxLengthProperty], [Name], [RowVersion])
 OUTPUT INSERTED.[UniqueNo]
-VALUES (@p0, @p1, @p2);",
+INTO @generated1
```
divega: Have you had any chance to measure if this has a perf impact?

divega:
Did you think about adding actual triggers to some tables in the functional tests?

divega:
I had a few questions but this looks :shipit: provided that you give me satisfactory answer :smile:

Not removing the "needs-review"  tag in case you want to get another pair of eyes on it.

src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
bricelam: Yes, I do. Concatenation is cheaper.

AndriySvyryd: Everywhere in this file or just this instance?

src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
AndriySvyryd: Maybe, if it is useful in any other places. I slightly prefer to have it private until we want to share it.

test/EntityFramework.MicrosoftSqlServer.FunctionalTests/DataAnnotationSqlServerTest.cs
AndriySvyryd: No, since I don't think perf has big impact on whether we do this.

AndriySvyryd:
I'll try to add FTs with triggers, but it might take some time.

test/EntityFramework.MicrosoftSqlServer.FunctionalTests/DataAnnotationSqlServerTest.cs
divega: When we discussed the bug we stated that if the solution had a significant perf impact we should consider keeping a path that doesn't support triggers and an opt-in flag for the behavior that supports triggers (that runs slower).

AndriySvyryd: Ok, we can do that after we get the perf numbers from the CI machine.

src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
divega: I am fine with that. It might actually be the only case in which we will ever need this anyway.

divega:
> I'll try to add FTs with triggers, but it might take some time.

Ok, can you file a bug on that and then :shipit:?

AndriySvyryd:
Added FTs

divega:
Neat! :shipit:

PR #3876, opened 2015-11-24
[SMALL] Provide abstraction for scaffolding default values

lajones:
For conversion of strings from SQL to internal C# data types move concepts which are not specific to SQL Server from `SqlServerLiteralUtilities` to a class available to all relational providers.


lajones:
Note to self - need to include change to .csproj to include `TSqlConversionUtilities`.

natemcmaster:
:shipit: (assuming tests pass..looks like this commit is pre-travis and pre-appveyor)

lajones:
Checked in with commit 2641923.

divega:
@lajones, I am curious, do we handle default values that cannot be interpreted as literals somewhere else in scaffolding? E.g. if a column default is bound to a scalar function such as `GETDATE()`, `HOST_ID` or `NEWID`?

lajones:
@divega Yes. See https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.MicrosoftSqlServer.Design/Internal/SqlServerLiteralUtilities.cs#L53. We first attempt to see if it matches the pattern for a function. If so we produce the `DefaultExpression()` API. If not we see if we can interpret it as a value and produce `DefaultValue()` API.

divega:
Nice! Thanks.

Issue #3616, opened 2015-10-30
Projection with multiple aggregates only running last one


davezych:
If you have a projection with multiple aggregates (I've only tried with combinations of `Sum` and `Count`) only the last one seems to get run, and each property gets the results from that last query. 

dnx: `1.0.0-rc1-16048`
EF: `7.0.0-rc1-16147`

Repro:

    public partial class MyContext : DbContext
    {
        public MyContext()
        {

        }
        
        public DbSet<Forum> Forums { get; set; }
        public DbSet<Topic> Topics { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(
                @"Server=(localdb)\mssqllocaldb;Database=EF7.AspNet5;Trusted_Connection=True;");
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
        }
    }
    
    public class Forum
    {
        public Forum()
        {
            Topics = new HashSet<Topic>();
        }

        public int Id { get; set; }
        public string Title { get; set; }

        public ICollection<Topic> Topics { get; set; }
    }

    public class Topic
    {
        public int Id { get; set; }
        public string Title { get; set; }
        public int ForumId { get; set; }
        public int Counter { get; set; }

        [ForeignKey(nameof(ForumId))]
        [InverseProperty("Topics")]
        public Forum Forum { get; set; }
    }

Query:

    var forums = _context.Forums.Include(f => f.Topics).Select(f => new
            {
                f.Id,
                Counter = f.Topics.Sum(t => t.Counter),
                TopicCount = f.Topics.Count(),
            }).ToList();

Sample data:

    var forum = new Forum();
    forum.Title = "SAMPLE";

    var topic = new Topic();
    topic.Title = "SAMPLE";
    topic.Forum = forum;
    topic.Counter = 10;

    forum.Topics.Add(topic);

    var topic2 = new Topic();
    topic2.Title = "SAMPLE2";
    topic2.Forum = forum;
    topic2.Counter = 9;

    forum.Topics.Add(topic2);

    context.Forums.Add(forum);

    context.SaveChanges();

Running the above query with this sample data results in:

    Id: 1
    Counter: 2
    TopicCount: 2

When we should get a `Counter` of `19`. This is the resulting SQL query:

    SELECT (
         SELECT COUNT(*)
         FROM [Topic] AS [t]
[...]

PR #3881, opened 2015-11-24
[SMALL] Fix to #3616 - Projection with multiple aggregates only running last one

maumar:
Problem was that when we were adding a subquery to the projection we were clearing everything that was in the projection list before. This is not a problem for scalars, because they are all re-added when we bind the query, but complex expressions are not added and are ignored. Fix is to only remove column projection elements instead.

### DIFF
--- a/src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
@@ -325,7 +325,7 @@ namespace Microsoft.Data.Entity.Query.Expressions
 
             if (expression is SelectExpression)
             {
-                ClearProjection();
+                ClearColumnProjections();
             }
 
             _projection.Add(expression);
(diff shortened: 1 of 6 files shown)


maumar:
@anpete

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -472,6 +472,23 @@ public virtual void ClearProjection()
             IsProjectStar = true;
         }
 
+        public virtual void ClearColumnProjections()
+        {
+            for (int i = _projection.Count - 1; i >= 0; i--)
+            {
+                var aliasExpression = _projection[i] as AliasExpression;
+                if (aliasExpression != null && aliasExpression.Expression.RemoveConvert() is ColumnExpression)
```
anpete: Shouldn't be any Convert nodes should there?

maumar:
@anpete  new version up

anpete:
:shipit:

PR #3884, opened 2015-11-24
Run ForeignKeyAddedConventions only if the ForeignKey is still present

### LINKED ISSUE #3805: Value cannot be null. Parameter name: navigation
I upgraded from EF6 SQLServer to EF7 SQLite -7-rc1-final.
Full .NET WPF
The context is not in the startup assembly.

Stack trace: https://gist.github.com/HalidCisse/02de3081aa6faafaac70

smitpatel:
Resolves #3805


AndriySvyryd:
:shipit:

PR #3886, opened 2015-11-25
RevEng: Updating FluentApiConfigurations to allow multiple lines …

lajones:
…and removing the ability for them to be prefixed by the ForXXX() concept which we no longer use.

This fixes the last part of #3394 - allowing the HasIndex() fluent API line to be split.


natemcmaster:
:shipit:

lajones:
Checked in commit d0f53e6.

PR #3898, opened 2015-11-25
Workaround anonymous type bug in diagnostics

natemcmaster:
Use concrete type instead of anonymous type

Workaround for https://github.com/dotnet/corefx/issues/4672


lajones:
:shipit:

Issue #3902, opened 2015-11-26
Tests using scratch localdb are flaky


AndriySvyryd:
```
System.Data.SqlClient.SqlException : Unable to open the physical file "C:\Users\asplab\Microsoft.Data.SqlServer.Scratch_15.mdf". Operating system error 2: "2(The system cannot find the file specified.)".
Unable to open the physical file "C:\Users\asplab\Microsoft.Data.SqlServer.Scratch_15.mdf". Operating system error 2: "2(The system cannot find the file specified.)".
Could not restart database "Microsoft.Data.SqlServer.Scratch_15". Reverting to the previous status.
ALTER DATABASE statement failed.
File activation failure. The physical file name "C:\Users\asplab\Microsoft.Data.SqlServer.Scratch_15_log.ldf" may be incorrect.
File activation failure. The physical file name "C:\Users\asplab\Microsoft.Data.SqlServer.Scratch_15_log.ldf" may be incorrect.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlCommand.RunExecuteNonQueryTds(String methodName, Boolean async, Int32 timeout, Boolean asyncWrite)
   at System.Data.SqlClient.SqlCommand.InternalExecuteNonQuery(TaskCompletionSource`1 completion, String methodName, Boolean sendToPipe, Int32 timeout, Boolean asyncWrite)
   at System.Data.SqlClient.SqlCommand.ExecuteNonQuery()
[...]

Issue #3749, opened 2015-11-13
Scratch databases not disposed during testing


natemcmaster:
![image](https://cloud.githubusercontent.com/assets/2696087/11153815/5d0a703a-89f0-11e5-86cf-799a90824399.png)

ErikEJ:
What is the problem ? :smile:

natemcmaster:
If testing fails, the following run will also fail because it does not start with a clean db file. Example: 

![image](https://cloud.githubusercontent.com/assets/2696087/11153995/5b54f278-89f1-11e5-8beb-b6bbc4644615.png)

ErikEJ:
Maybe use inmemory? https://sqlite.org/inmemorydb.html

bricelam:
@ErikEJ We'd have to use named, shared-cache, in-memory databases and manage the lifetime in our testing framework because once all the connections are closed, the database goes away. It was just easier to use an on-disk database for now.

natemcmaster:
This is not longer limited to SQLite. Remarking for triage as SQL Server tests fail often because the previous run failed to cleanup the scratchdb.mdf file.

![image](https://cloud.githubusercontent.com/assets/2696087/11385464/fc382f36-92cd-11e5-94f5-b834461cb15c.png)

smitpatel:
Most EF builds on CI are failing due to this flakiness.

AndriySvyryd:
Even though the scratch dbs are deleted the localdb issue is still present, I've opened a different issue to track it: #3902

PR #3927, opened 2015-11-30
Wait for localdb database files to be deleted instead of deleting them

AndriySvyryd:
Part of #3902

### DIFF
--- a/test/EntityFramework.MicrosoftSqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
+++ b/test/EntityFramework.MicrosoftSqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
@@ -19,8 +19,6 @@ namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
     {
         public const int CommandTimeout = 90;
 
-        private static int _scratchCount;
-
         public static SqlServerTestStore GetOrCreateShared(string name, Action initializeDatabase)
             => new SqlServerTestStore(name).CreateShared(initializeDatabase);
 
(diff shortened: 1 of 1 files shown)


AndriySvyryd:
First stab at the issue. @natemcmaster @bricelam

natemcmaster:
Why do we need to wait? Can't we safely assume that if the "DROP DATABASE" command executes that SQL Server will cleanup files appropriately?

smitpatel:
`DROP DATABASE` command cleans up files appropriately but we can run into a case where the command is issued but we tried to use the same name before it gets deleted by file system.

AndriySvyryd:
This is needed if the last ran didn't delete the database and we are trying to create it with the same name.

natemcmaster:
It sounds like this is trying to compensate for weak collision detection in the "CreateTransient" logic.

AndriySvyryd:
and to cleanup state from previous runs

natemcmaster:
I would favor a solution that instead makes `SqlServerTestStore.CreateTransient` impervious to bad. [SqliteTestStore does this](https://github.com/aspnet/EntityFramework/blob/18439602ee448c15b0dfdafa23d455a856204a08/test/EntityFramework.Sqlite.FunctionalTests/SqliteTestStore.cs#L27) in the CreateTransient logic because cleanup is not guaranteed to execute. Caveat: this led to #3749, but this seems to be due to an issue in Dispose not being called correctly by xunit.

AndriySvyryd:
There are other reasons that could make Dispose not run and leave the DB around. Without the cleanup logic we would need to drop them manually by connecting to the CI agents. Sqllite dbs are much easier to cleanup.

AndriySvyryd:
Anyway the AppVeyor builds failed with the new timeout message, so I'll try Nate's suggestion

natemcmaster:
Yeah, this is related to a second problem. This cleanup assumes localdb, which we do not always use. AppVeyor uses SQL Server 2014.

AndriySvyryd:
I don't think it does, we wouldn't get this error:
```
System.Exception : Timeout when waiting for 'C:\Users\appveyor\Microsoft.Data.SqlServer.Scratch_1.mdf' to be deleted.
```

AndriySvyryd:
AppVeyor passes now

natemcmaster:
:shipit:

Issue #3861, opened 2015-11-23
I get errors when performing reverse engineering of my database and do not know because it is produced


micball:
In new console project run nugget console:

Install-Package EntityFramework.MicrosoftSqlServer –Pre
Install-Package EntityFramework.Commands –Pre
Install-Package EntityFramework.MicrosoftSqlServer.Design –Pre

Finaly execute:

Scaffold-DbContext -provider EntityFramework.MicrosoftSqlServer -connection "Server=Sealand;Database=Maldivas_oran;Trusted_Connection=True;"

The first test show:

System.Data.SqlTypes.SqlNullValueException: Data is Null. This method or property cannot be called on Null values.
   at System.Data.SqlClient.SqlBuffer.get_String()
   at System.Data.SqlClient.SqlDataReader.GetString(Int32 i)
   at Microsoft.Data.Entity.Scaffolding.SqlServerDatabaseModelFactory.GetIndexes()
   at Microsoft.Data.Entity.Scaffolding.SqlServerDatabaseModelFactory.Create(String connectionString, TableSelectionSet tableSelectionSet)
   at Microsoft.Data.Entity.Scaffolding.RelationalScaffoldingModelFactory.Create(String connectionString, TableSelectionSet tableSelectionSet)
   at Microsoft.Data.Entity.Scaffolding.SqlServerScaffoldingModelFactory.Create(String connectionString, TableSelectionSet tableSelectionSet)
   at Microsoft.Data.Entity.Scaffolding.Internal.ReverseEngineeringGenerator.GetMetadataModel(ReverseEngineeringConfiguration configuration)
   at Microsoft.Data.Entity.Scaffolding.Internal.ReverseEngineeringGenerator.GenerateAsync(ReverseEngineeringConfiguration configuration, CancellationToken cancellationToken)
   at Microsoft.Data.Entity.Design.DatabaseOperations.ReverseEngineerAsync(String provider, String connectionString, String outputDir, String dbContextClassName, List`1 schemas, List`1 tables, Boolean useDataAnnotations, CancellationToken cancellationToken)
   at Microsoft.Data.Entity.Design.OperationExecutor.<ReverseEngineerImpl>d__19.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
[...]

PR #3933, opened 2015-12-01
RevEng: Adding warning messages for potential errors in metadata

lajones:
Fix for #3861. We experienced some problems where we were receiving NULLs in metadata we expected to be non-null or we were unable to find tables for column metadata or columns for indexes/FKs etc.

So have added guard code and warning messages for if we experience this again.


lajones:
Closing in favor of #3934.

PR #3934, opened 2015-12-01
Reveng: Guarding against duplicate or bad metadata and adding warning messages

lajones:
Fix for #3861. We had problems with Foreign Keys or Indexes with same name on different tables. We also experienced some problems where we were receiving NULLs in metadata we expected to be non-null or we were unable to find tables for column metadata or columns for indexes/FKs etc. I've fixed the duplicate issue and now we guard against bad metadata and issue warning messages.

### DIFF
--- a/src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
+++ b/src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
@@ -4,10 +4,14 @@
 using System;
 using System.Collections.Generic;
 using System.Data.SqlClient;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Migrations;
+using Microsoft.Data.Entity.Scaffolding.Internal;
 using Microsoft.Data.Entity.Scaffolding.Metadata;
 using Microsoft.Data.Entity.Utilities;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Data.Entity.Scaffolding
 {
@@ -22,6 +26,15 @@ namespace Microsoft.Data.Entity.Scaffolding
         private static string TableKey(TableModel table) => TableKey(table.Name, table.SchemaName);
         private static string TableKey(string name, string schema = null) => "[" + (schema ?? "") + "].[" + name + "]";
         private static string ColumnKey(TableModel table, string columnName) => TableKey(table) + ".[" + columnName + "]";
+
+        public SqlServerDatabaseModelFactory([NotNull] ILoggerFactory loggerFactory)
+        {
+            Check.NotNull(loggerFactory, nameof(loggerFactory));
+
+            Logger = loggerFactory.CreateCommandsLogger();
+        }
+
+        public virtual ILogger Logger { get; }
 
         private void ResetState()
         {
@@ -59,8 +72,9 @@ namespace Microsoft.Data.Entity.Scaffolding
         private void GetTables()
         {
             var command = _connection.CreateCommand();
-            command.CommandText = "SELECT schema_name(t.schema_id) AS [schema], t.name FROM sys.tables AS t " +
-                                  $"WHERE t.name <> '{HistoryRepository.DefaultTableName}'";
+            command.CommandText =
+                "SELECT schema_name(t.schema_id) AS [schema], t.name FROM sys.tables AS t " +
+                $"WHERE t.name <> '{HistoryRepository.DefaultTableName}'";
             using (var reader = command.ExecuteReader())
             {
                 while (reader.Read())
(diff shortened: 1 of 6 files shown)


src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -119,13 +133,28 @@ ELSE null
...
+                    if (!_tables.TryGetValue(TableKey(tableName, schemaName), out table))
+                    {
+                        Logger.LogWarning(
+                            SqlServerDesignStrings.UnableToFindTableForColumn(columnName, schemaName, tableName));
+                        continue;
+                    }
+
                     var dataTypeName = reader.GetString(2);
-                    var nullable = reader.GetBoolean(5);
+                    var nullable = reader.IsDBNull(5) ? false : reader.GetBoolean(5);
```
natemcmaster: This column (`sys.columns.is_nullable`) is never DbNull

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -119,13 +133,28 @@ ELSE null
...
                     var tableName = reader.GetString(1);
+                    var columnName = reader.GetStringOrNull(3);
                     if (!_tableSelectionSet.Allows(schemaName, tableName))
                     {
                         continue;
                     }
 
+                    if (string.IsNullOrEmpty(columnName))
+                    {
+                        Logger.LogWarning(SqlServerDesignStrings.ColumnNameEmpty(schemaName, tableName));
```
natemcmaster: YAGNI. Columns must have a name.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -119,13 +133,28 @@ ELSE null
...
+                        continue;
+                    }
+
+                    TableModel table;
+                    if (!_tables.TryGetValue(TableKey(tableName, schemaName), out table))
```
natemcmaster: This is going to blowup the log output when a user filters a table. We should only log this warning if we hit this unexpectedly i.e. table was allowed by TableSelectionSet AND we didn't find it in `_tables`

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
lajones: I looked at the definition for the view. It is nullable.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -174,41 +202,50 @@ ELSE null
...
+                    var tableName = reader.GetString(1);
+                    var indexName = reader.GetStringOrNull(2);
                     if (!_tableSelectionSet.Allows(schemaName, tableName))
                     {
                         continue;
                     }
 
+                    if (string.IsNullOrEmpty(indexName))
+                    {
+                        Logger.LogWarning(SqlServerDesignStrings.IndexNameEmpty(schemaName, tableName));
```
natemcmaster: Likewise. YAGNI. Indexes must have names.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
lajones: I looked at the definition of the view. It is nullable. I agree that's weird and would mean something strange but I'd rather issue a warning than die with the KeyNotFoundException that people are seeing.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -216,14 +253,33 @@ private void GetIndexes()
                         {
                             Table = table,
                             Name = indexName,
-                            IsUnique = reader.GetBoolean(3),
-                            IsClustered = (reader.GetString(5) == "CLUSTERED") ? true : default(bool?)
+                            IsUnique = reader.IsDBNull(3) ? false : reader.GetBoolean(3),
```
natemcmaster: is_unique is always defined.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
lajones: We do this. See the lines just before this.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -216,14 +253,33 @@ private void GetIndexes()
...
                         };
                         table.Indexes.Add(index);
                     }
-                    var columnName = reader.GetString(4);
-                    var column = _tableColumns[ColumnKey(index.Table, columnName)];
-                    index.Columns.Add(column);
+
+                    var columnName = reader.GetStringOrNull(4);
+                    ColumnModel column = null;
+                    if (string.IsNullOrEmpty(columnName))
```
natemcmaster: YAGNI

natemcmaster:
Overall this feels like overdone defensive programming. We have a well-defined, expected content coming from sys.* tables and can program against that expectation.

Also LogWarning seems too high for most of these. LogVerbose instead?

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
lajones: As above. Weirdly this is not true. In fact this was the case that caused the error the user was seeing. Basically I've guarded against every situation where the appropriate sys view definition allows nulls.

lajones:
As mentioned - the user has found themselves with a DB where these issues do happen. Yes, they are weird schemas and the user has probably made a mistake in creating the schema that way. But I'd much rather issue a warning message than have scaffolding fail with a strange exception. Disagree on the level too - you are probably going to run into problems later on - you need the warning that e.g. the index is missing a column etc.

rowanmiller:
I agree with using the warning level too... as noted the data being returned doesn't really make sense so we should clearly surface that we don't know how to handle it and are just skipping it.

natemcmaster:
I didn't read #3861 before making above comments. Wow, yes that is weird that a user got the database into this state. Nameless indexes and columns? How is that even possible?

But since apparently these edge cases are possible, let's defend. :shipit:

lajones:
Thanks. And yes - I totally get how weird this is. Checked in with commit 454968a.

PR #3939, opened 2015-12-01
Add a non-empty check for navigation name in relationship API

### LINKED ISSUE #3911: Can't remove migration because of empty navigation property
Hi,
I just created a migration then tried to remove it.  When I tried to remove it, I got a weird error, and I'm not really sure what's causing it:

The navigation property '' cannot be added to the entity type 'MyType' because a navigation property with the same name already exists on entity type 'MyType'.

Note the empty property name.
I was able to create the migration somehow... I just can't remove it.

Here are the 2 stack trace methods before the exception:

   at Microsoft.Data.Entity.Metadata.Builders.ReferenceNavigationBuilder.WithOneBuilder(String reference)
   at Microsoft.Data.Entity.Metadata.Builders.ReferenceNavigationBuilder.WithOne(String reference)

AndriySvyryd:
Fixes #3911


lajones:
Looks good. :shipit:

PR #3946, opened 2015-12-02
:bug: Remove derived type from directlyDerivedTypes list when removing entity type

### LINKED ISSUE #3894: Removing derived type does not update list of directly derived type in the base type
See https://github.com/aspnet/EntityFramework/issues/3555

I don't have a code sample yet, but I have some more information.
I reproduced this using EF7 RC1 final, .Net Framework 4.6 in VS2015.

The null reference happens in this section of Microsoft.Data.Entity.Metadata.Internal.InternalModelBuilder.Remove:

```csharp
            foreach (var directlyDerivedType in entityType.GetDirectlyDerivedTypes().ToList())
            {
                var removed = Entity(directlyDerivedType.Name, ConfigurationSource.Convention)
                    .HasBaseType(entityType.BaseType, configurationSource);
                Debug.Assert(removed != null);
            }
```
"Entity(directlyDerivedType.Name, ConfigurationSource.Convention)" ends up returning null because "IsIgnored(name, configurationSource)" in the "Entity" method is true.

In the case that I encountered, these are the variable values in this section:

entityType - System.Xml.Linq.XNode
directlyDerivedType.Name - System.Xml.Linq.XDocument
entityType.BaseType - null
configurationSource - DataAnnotation

---

It appears that while the code has changed, the problem still exists in the Dev branch (as of commit 26276266), in this case, "directlyDerivedType.Builder" returns null.
```csharp
            foreach (var directlyDerivedType in entityType.GetDirectlyDerivedTypes().ToList())
            {
                var derivedEntityTypeBuilder = directlyDerivedType.Builder
                    .HasBaseType(entityType.BaseType, configurationSource);
                Debug.Assert(derivedEntityTypeBuilder != null);
            }
```

---

For a workaround, I ended up adding the following to my code.
I did not need this in Beta8.

```csharp
            modelBuilder.Ignore<Version>();
            modelBuilder.Ignore<XAttribute>();
            modelBuilder.Ignore<XElement>();
            modelBuilder.Ignore<XName>();
            modelBuilder.Ignore<XNamespace>();
            modelBuilder.Ignore<XNode>();
            modelBuilder.Ignore<XDocument>();
```
It appears that the types listed above come from entity properties in my model that are ignored using this method:
```csharp
            modelBuilder.Entity<T>(
                c =>
                {
                    c.Ignore( x => x.Prop);
                } );
```

smitpatel:
Resolves #3894

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/InternalModelBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalModelBuilder.cs
@@ -112,6 +112,10 @@ namespace Microsoft.Data.Entity.Metadata.Internal
                 return false;
             }
 
+            // Set base type as null to remove the entityType from directly derived types of the base type
+            var baseType = entityType.BaseType;
+            entityType.Builder.HasBaseType((EntityType)null, configurationSource);
+
             Metadata.Ignore(entityType.Name, configurationSource);
 
             var entityTypeBuilder = entityType.Builder;
@@ -130,7 +134,7 @@ namespace Microsoft.Data.Entity.Metadata.Internal
             foreach (var directlyDerivedType in entityType.GetDirectlyDerivedTypes().ToList())
             {
                 var derivedEntityTypeBuilder = directlyDerivedType.Builder
-                    .HasBaseType(entityType.BaseType, configurationSource);
+                    .HasBaseType(baseType, configurationSource);
                 Debug.Assert(derivedEntityTypeBuilder != null);
             }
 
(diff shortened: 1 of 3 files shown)


src/EntityFramework.Core/Metadata/Internal/InternalModelBuilder.cs
```diff
@@ -112,6 +112,9 @@ private bool Ignore(EntityType entityType, ConfigurationSource configurationSour
                 return false;
             }
 
+            var baseType = entityType.BaseType;
```
lajones: Nit: Not immediately obvious when reading the code why you're doing this. So I would add a comment. But otherwise looks good. :shipit:

AndriySvyryd:
So it was not necessary to add the extra `IsIgnored` check?
:shipit:

smitpatel:
@AndriySvyryd - The only time when entityTypeBuilder is null is due to it being in derived types list.

PR #3947, opened 2015-12-02
Use SCOPE_IDENTITY() for SqlServer inserts if only one row is being inserted

### LINKED ISSUE #3893: Can we avoid some of the overhead from OUTPUT INTO changes
There is a significant perf impact for no batching case

AndriySvyryd:
Fixes #3893


divega:
Nice that we already had the code :smile: :shipit:

Issue #3856, opened 2015-11-23
Reverse Engineer: Require a flag to overwrite existing files


rowanmiller:
Per discussion in #3849 we overwrite existing files, we should probably error and require a flag to do this.

lajones:
Added -f|--force flag to command line (= -Force flag for powershell). Commit c698d65.

PR #3949, opened 2015-12-02
RevEng: Adding Force flag to prevent existing files being overwritten by mistake

lajones:
Fix for #3856.

### DIFF
--- a/src/EntityFramework.Commands/Commands/Program.cs
+++ b/src/EntityFramework.Commands/Commands/Program.cs
@@ -157,6 +157,10 @@ namespace Microsoft.Data.Entity.Commands
                             var dbContextClassName = scaffold.Option(
                                 "-c|--context <name>",
                                 "Name of the generated DbContext class.");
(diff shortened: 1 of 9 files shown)


src/EntityFramework.Commands/Commands/Program.cs
```diff
@@ -655,13 +660,15 @@ public virtual int RemoveMigration([CanBeNull] string context)
                 [CanBeNull] List<string> schemaFilters,
                 [CanBeNull] List<string> tableFilters,
                 bool useDataAnnotations,
+                bool force,
```
natemcmaster: "force" is good as a command line parameter but in code its meaning is obscure. "overwriteFiles" would be more clear.

lajones: Sounds good. Will fix.

natemcmaster:
:shipit: w/ feedback updates

lajones:
Checked in with updates - commit c698d65.

PR #3957, opened 2015-12-02
Conditionally skip SQL Server tests on Linux w/o external server

natemcmaster:
Use Xunit to skip assemblies rather than test runner. 

Future proofing against pending build system changes.

### DIFF
--- a/test/EntityFramework.MicrosoftSqlServer.FunctionalTests/Utilities/SqlServerConfiguredConditionAttribute.cs
+++ b/test/EntityFramework.MicrosoftSqlServer.FunctionalTests/Utilities/SqlServerConfiguredConditionAttribute.cs
@@ -2,16 +2,22 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
+using System.Data.SqlClient;
+using Microsoft.Data.Entity.FunctionalTests.TestUtilities;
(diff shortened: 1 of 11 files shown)


test/EntityFramework.MicrosoftSqlServer.FunctionalTests/Utilities/SqlServerConfiguredConditionAttribute.cs
```diff
@@ -2,16 +2,22 @@
...
 
 namespace Microsoft.Data.Entity.SqlServer.FunctionalTests
 {
-    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly)]
     public class SqlServerConfiguredConditionAttribute : Attribute, ITestCondition
     {
-        public bool IsMet => SqlServerTestStore.IsConfigured;
+    	private static string DataSource = new SqlConnectionStringBuilder(SqlServerTestStore.CreateConnectionString("sample")).DataSource;
+    	private bool IsLocalDb = DataSource.StartsWith("(localdb)");
```
lajones: Would do this case-insensitive. E.g. I often use "(LocalDB)".

lajones:
Looks good other than small comment above. :shipit:

Issue #3830, opened 2015-11-20
Relative directory passed to Scaffold-DbContext for OutputDirectory parameter creates a project directory called "."


esmith-blackrat:
**Relevant Assemblies**
EF 7.0.0-rc1-final
EntityFramework.Commands.7.0.0-rc1-final
EntityFramework.MicrosoftSqlServer.Design.7.0.0-rc1-final

Visual Studio 2015, .NET 4.6 Console Application

When running `Scaffold-DbContext` from the Package Manager Console, if you pass a directory in the style of ".\Models" to argument `-OutputDirectory`, a project folder is added called "." with the child directory "Models".

As a user, I assumed the `-OutputDirectory` parameter expected (or at least would work with) this type of path as it does with other PowerShell cmdlets.

As a side note, if you delete the "." folder from within Visual Studio, it deletes as much as it can from the project folder on the file system, which is unfortunate.

lajones:
Fixed with PR #3959.

PR #3959, opened 2015-12-03
RevEng: Canonicalizing output paths

lajones:
Fix for #3830. For Scaffold-DbContext we send back a list of the paths to the generated files to powershell so that it can update the VS project to include them in the project. When the output directory looked like ".\Models" we were sending back filenames like "X:\Project\Path\\.\Models\MyEntity.cs". The extra "." was being interpreted by the VS AddFromFile() API as a real part of the name of the file. This fixes that by canonicalizing the name of the path before it get sent back so that the above now becomes "X:\Project\Path\Models\MyEntity.cs".

### DIFF
--- a/src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
+++ b/src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
@@ -142,7 +142,7 @@ namespace Microsoft.Data.Entity.Scaffolding.Internal
         ///  an <see cref="NamespaceAndOutputPaths"> object containing the canonicalized paths
         ///  and the namespace
         /// </returns>
-        public static NamespaceAndOutputPaths ConstructNamespaceAndCanonicalizedPaths(
+        internal static NamespaceAndOutputPaths ConstructNamespaceAndCanonicalizedPaths(
             [NotNull] string rootNamespace,
             [NotNull] string projectPath,
             [CanBeNull] string outputPath)
@@ -150,7 +150,10 @@ namespace Microsoft.Data.Entity.Scaffolding.Internal
             Check.NotEmpty(rootNamespace, nameof(rootNamespace));
             Check.NotEmpty(projectPath, nameof(projectPath));
 
-            // strip off any directory separator chars at end of project path
+            // Strip off any directory separator chars at end of project path
+            // to ensure that when we strip off the canonicalized project path
+            // later to form the canonicalized relative path we strip off the
+            // correct number of characters.
             for (var projectPathLastChar = projectPath.Last();
                 directorySeparatorChars.Contains(projectPathLastChar);
                 projectPathLastChar = projectPath.Last())
@@ -186,9 +189,9 @@ namespace Microsoft.Data.Entity.Scaffolding.Internal
                 canonicalizedFullOutputPath, canonicalizedRelativeOutputPath);
         }
 
-        public class NamespaceAndOutputPaths
+        internal class NamespaceAndOutputPaths
         {
-            public NamespaceAndOutputPaths(
+            internal NamespaceAndOutputPaths(
                 [NotNull] string @namespace,
                 [NotNull] string canonicalizedFullOutputPath,
                 [CanBeNull] string canonicalizedRelativeOutputPath)
@@ -201,9 +204,9 @@ namespace Microsoft.Data.Entity.Scaffolding.Internal
                 CanonicalizedRelativeOutputPath = canonicalizedRelativeOutputPath;
             }
 
-            public virtual string Namespace { get; }
-            public virtual string CanonicalizedFullOutputPath { get; }
-            public virtual string CanonicalizedRelativeOutputPath { get; }
+            internal string Namespace { get; }
+            internal string CanonicalizedFullOutputPath { get; }
+            internal string CanonicalizedRelativeOutputPath { get; }
         }
     }
 }
--- a/src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
+++ b/src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
@@ -52,8 +52,11 @@ namespace Microsoft.Data.Entity.Scaffolding.Internal
             configuration.CheckValidity();
 
             var metadataModel = GetMetadataModel(configuration);
+
+            var outputPaths = ConstructCanonicalizedPaths(configuration.ProjectPath, configuration.OutputPath);
+
             var @namespace = ConstructNamespace(configuration.ProjectRootNamespace,
-                    configuration.ProjectPath, configuration.OutputPath);
+                    outputPaths.CanonicalizedRelativeOutputPath);
 
             var customConfiguration = _configurationFactory
                 .CreateCustomConfiguration(
@@ -67,16 +70,12 @@ namespace Microsoft.Data.Entity.Scaffolding.Internal
                 ? modelConfiguration.ClassName()
                 : customConfiguration.ContextClassName;
 
-            var outputPath = string.IsNullOrEmpty(configuration.OutputPath)
-                ? configuration.ProjectPath
-                : (Path.IsPathRooted(configuration.OutputPath)
-                    ? configuration.OutputPath
-                    : Path.Combine(configuration.ProjectPath, configuration.OutputPath));
-
-            CheckOutputFiles(outputPath, dbContextClassName, metadataModel, configuration.OverwriteFiles);
+            CheckOutputFiles(outputPaths.CanonicalizedFullOutputPath,
+                dbContextClassName, metadataModel, configuration.OverwriteFiles);
 
             return CodeWriter.WriteCodeAsync(
-                modelConfiguration, outputPath, dbContextClassName, cancellationToken);
+                modelConfiguration, outputPaths.CanonicalizedFullOutputPath,
+                dbContextClassName, cancellationToken);
         }
 
         public virtual IModel GetMetadataModel([NotNull] ReverseEngineeringConfiguration configuration)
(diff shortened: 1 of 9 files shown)


src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
```diff
@@ -154,24 +176,40 @@ public virtual IModel GetMetadataModel([NotNull] ReverseEngineeringConfiguration
...
+        public class CanonicalizedOutputPaths
+        {
+            public CanonicalizedOutputPaths(
+                [NotNull] string canonicalizedFullOutputPath,
+                [CanBeNull] string canonicalizedRelativeOutputPath)
             {
-                @namespace += "." + CSharpUtilities.Instance.GenerateCSharpIdentifier(pathPart, null);
+                Check.NotEmpty(canonicalizedFullOutputPath, nameof(canonicalizedFullOutputPath));
+
+                CanonicalizedFullOutputPath = canonicalizedFullOutputPath;
```
divega: Why repeate the words of the name if the class in each property of the class?

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
```diff
@@ -154,24 +176,40 @@ public virtual IModel GetMetadataModel([NotNull] ReverseEngineeringConfiguration
...
-            {
-                // canonicalizedOutputPath is outside of project - so just use root namespace
-                return rootNamespace;
-            }
+            var canonicalizedFullProjectPath = Path.GetFullPath(projectPath);
+            var canonicalizedFullOutputPath =
+                string.IsNullOrEmpty(outputPath)
+                    ? canonicalizedFullProjectPath
+                    : Path.IsPathRooted(outputPath)
+                        ? Path.GetFullPath(outputPath)
```
divega: Are you sure that `Path.Combine()` doesn't do this check already?

divega:
I might be wrong, but a quick look at the change gives me the impression that something simpler could work and be more maintainable. I would the usual reviewers for thus area to chip in though. Cc @natemcmaster, @bricelam

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
lajones: Will change them.

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
lajones: Just checked and you're right. Will fix.

divega:
@lajones note that Travis build for this PR is failing. @natemcmaster noticed last night that it seems to be a legitimate issue with the PR. Apparently either something in the new path handling logic or on one of the new tests doesn't work well in non-Windows platforms.

lajones:
@divega - feel free to suggest. The only thing I've found that canonicalizes is Path.GetFullPath(). And this algorithm was already in use to produce the namespace - I've just adapted it so we can use it both for the output paths and the namespace. I would have preferred out params and skipping the `CanonicalizedOutputPaths` class altogether, but that's against the current coding practice. Have added new commit 7e7a205 which combines `ConstructOutputPaths()` and `ConstructNamespace()` to make it simpler.

natemcmaster:
The failing test on Linux is:
```
ReverseEngineeringGeneratorTests.Constructs_correct_canonical_paths

rootNamespace: "Root.Namespace", 
projectPath: "/project/Path", 
outputPath: "", 
expectedNamepace: "Root.Namespace", 
canonicalizedFullOutputPath: "project/Path", 
canonicalizedRelativeOutputPath: ""

      Assert.Equal() Failure
                ↓ (pos 0)
      Expected: project/Path
      Actual:   /project/Path
                ↑ (pos 0)
```

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
```diff
@@ -133,19 +131,24 @@ public virtual IModel GetMetadataModel([NotNull] ReverseEngineeringConfiguration
...
+        /// <returns>
+        ///  an <see cref="NamespaceAndOutputPaths"> object containing the canonicalized paths
+        ///  and the namespace
+        /// </returns>
+        public static NamespaceAndOutputPaths ConstructNamespaceAndCanonicalizedPaths(
+            [NotNull] string rootNamespace,
+            [NotNull] string projectPath,
+            [CanBeNull] string outputPath)
         {
-            Check.NotEmpty(rootNamespace, nameof(rootNamespace));
```
natemcmaster: Was this check intentionally removed?

lajones:
@natemcmaster re Linux test typo on my end. Should be fixed with commit c698808. Let me know if not.

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
lajones: Nope - well spotted. Will put back in.

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
```diff
@@ -154,24 +157,52 @@ public virtual IModel GetMetadataModel([NotNull] ReverseEngineeringConfiguration
...
+                    ? canonicalizedFullProjectPath
+                    : Path.GetFullPath(Path.Combine(projectPath, outputPath));
+
+            var canonicalizedRelativeOutputPath =
+                canonicalizedFullOutputPath == canonicalizedFullProjectPath
+                ? string.Empty
+                : canonicalizedFullOutputPath.StartsWith(canonicalizedFullProjectPath)
+                    ? canonicalizedFullOutputPath
+                        .Substring(canonicalizedFullProjectPath.Count() + 1)
+                    : null;
```
natemcmaster: Not completely sure if this is the same, but this looks almost the same as `Uri.MakeRelativeUri`

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
```diff
@@ -154,24 +157,52 @@ public virtual IModel GetMetadataModel([NotNull] ReverseEngineeringConfiguration
...
+
+            var @namespace = rootNamespace;
+            if (!string.IsNullOrEmpty(canonicalizedRelativeOutputPath))
             {
-                // canonicalizedOutputPath is outside of project - so just use root namespace
-                return rootNamespace;
+                foreach (var pathPart in canonicalizedRelativeOutputPath
+                    .Split(directorySeparatorChars, StringSplitOptions.RemoveEmptyEntries))
```
natemcmaster: Could we use `Uri.Segments` to make this more x-plat safe?

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
lajones: I tried `Uri` stuff a while back. Don't remember exactly what the problem was now but would rather keep full control over what we are doing.

src/EntityFramework.Commands/Scaffolding/Internal/ReverseEngineeringGenerator.cs
lajones: Hmmm. It seems to be passing on both Linux and Windows platforms already? I'd rather not complicate it unless I have to.

natemcmaster:
Per @divega about simplifying, here are a few suggestions:
 - `ConstructNamespaceAndCanonicalizedPaths` can be split into separate functions. These are two separate concerns.
 - The class `NamespaceAndOutputPaths` could be a tuple. This is never used by other components.
 - Reuse functionality in `System.Uri` (where possible.)

lajones:
@natemcmaster @divega actually `ConstructNamespaceAndCanonicalizedPaths` was 2 separate functions in the 1st commit - I put them together to simplify. And they are not unrelated - you need the canonicalized relative (both the canonicalized and the relative are important) path to construct the namespace. If you put them as 2 separate functions then either you have to pass the canonicalized relative path into the `ConstructNamespace` function - which makes it odd as a public standalone method - how would you check that what is passed in is canonicalized, what would you do if it wasn't relative? Or you can pass `ConstructNamespace` the same args as `ConstructOutputPaths` (plus `rootNamespace`) and call `ConstructOutputPaths` from `ConstructNamespace` to make sure you get the canonicalized relative path but then you would just be calling the same function twice on the same data? It seemed more efficient to just do it all at once.

I agree it could return a `Tuple` - to me that's actually less clear than what I've done here but I will do it if it makes it clearer to others.

Re `Uri` - I just did a little testing and without using `Path.GetFullPath()` `Uri`s start throwing `UriFormatException` if they come across the "." or ".." syntax. So would much rather use `Path`.

Re overall simplicity: `ConstructNamespaceAndCanonicalizedPaths` seems to me to be a pretty simple little helper function very similar to what was in `ConstructNamespace` before. Agree it is very specific to what it does and is not used elsewhere - I would have liked to make it private but I kept it public so we could test it.

natemcmaster:
> I agree it could return a Tuple - to me that's actually less clear than what I've done here but I will do it if it makes it clearer to others.

Agreed. Anonymous types or private type instead? A public type adds API surface for something that  isn't an API. `NamespaceAndOutputPaths` is really just a named data structure for marshaling data from one function to another. 

> ConstructNamespaceAndCanonicalizedPaths was 2 separate functions in the 1st commit - I put them together to simplify

I still suggest splitting. From a design perspective, constructing a namespace and canonicalizing paths are separate responsibilities. Avoiding a precondition check is not reason enough to push them together.

> I would have liked to make it (ConstructNamespaceAndCanonicalizedPaths) private but I kept it public so we could test it.

Make it `internal` so it can still be tested but is not part of our public API. We should do this regardless of how the "split or don't" decision is solve above.

lajones:
@natemcmaster Happy to do internal. And will see what I can do to make `NamespaceAndOutputPaths` not public but still testable. Still disagree on splitting the methods. To construct the namespace you need the canonicalized relative path. The only  "precondition check" I could imagine putting in would be to also pass it the `projectPath` and `outputPath` and going through the same functionality as before to check that what has been passed in is the correct canonicalized relative path. At that point I might as well not bother passing the canonicalized relative path at all but just call a new `ConstructCanonicalizedPaths` internally from within `ConstructNamespace` and we'll be repeating the same code calculations. Not sure why the problem - we did exactly that before this check-in - we just weren't exposing the canonicalized paths as part of what was being returned from the method.

lajones:
Have updated it so the class `NamespaceAndOutputPaths` is internal and so is the method that uses it. Also found that 2 tests (including the one we're talking about here) should now be in `EntityFramework.Commands.Tests` to match their source code - so pulled them into there at the same time.

natemcmaster:
Discussed in person with @lajones. We will keep the function and it's datatype internal. :shipit:

lajones:
Checked in with commit 036140f.

PR #3968, opened 2015-12-03
Query: Various fixes and minor perf improvements to query front-end:

anpete:
- Replace ET stringification hashing with ET EqualityComparer. ~100% improvement in dictionary key resolution micro-benchmark.
- Parameterizer reworked. Code simplifications and minor perf improvements. Constants at the top-level (outside lambdas) are now parameterized correctly. Previously was causing cache bloat for Skip/Take queries.
- Relational: SelectExpression Limit/Offset are now expressions allowing query parameterization.
- Fixed a bug with SQL Date/DateTime literals - Looks like the SQL Server specific literal formats were inadvertently removed with the RelationalCommandBuilder refactoring work.
- Corrected parameterization of FromSql queries. FromSql arguments now represented as an expression.


ajcvickers:
:shipit:

PR #3969, opened 2015-12-03
Enforce CLR inheritance in model using validation

### LINKED ISSUE #3860: Make inheritance mapping conventions more "assertive"
**Conclusion for the moment is that we will just block having types that inherit in CLR but not in the model.** This will give an exception in this case - perhaps not the clearest message, but at least you don't inadvertently get the wrong thing.

<hr />

Currently it is possible to break an inheritance hierarchy by providing entity key configuration on derived types first and then conflicting configuration on a base type.

We should consider throwing an exception at conflicting key configuration regardless of the order. 

@rowanmiller could you paste the repro code here?

AndriySvyryd:
Fixes #3860

### DIFF
--- a/src/EntityFramework.Core/Properties/CoreStrings.resx
+++ b/src/EntityFramework.Core/Properties/CoreStrings.resx
@@ -507,4 +507,7 @@
   <data name="MultipleNavigationsSameFk" xml:space="preserve">
     <value>There are multiple navigations in entity type '{entityType}' which are pointing to same set of properties - '{propertyList}' using ForeignKeyAttribute.</value>
   </data>
+  <data name="InconsistentInheritance" xml:space="preserve">
+    <value>The entity type '{entityType}' should derive from '{baseEntityType}' to reflect the hierarchy of the corresponding CLR types.</value>
+  </data>
 </root>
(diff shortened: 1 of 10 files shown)


src/EntityFramework.Core/Properties/CoreStrings.resx
```diff
@@ -507,4 +507,7 @@
   <data name="MultipleNavigationsSameFk" xml:space="preserve">
     <value>There are multiple navigations in entity type '{entityType}' which are pointing to same set of properties - '{propertyList}' using ForeignKeyAttribute.</value>
   </data>
+  <data name="InconsistentInheritance" xml:space="preserve">
+    <value>The entity type '{entityType}' should derive from '{baseEntityType}' to reflect the hierarchy of the corresponding CLR types.</value>
```
ajcvickers: Looks good, but the message could be confusing because you will see something like, "The entity type 'DiscontinuedProduct' should derive from 'Product' to reflect the hierarchy of the corresponding CLR types." and if you got here by mistake you probably won't know what to do. That being said, I don't know how to improve it. :-) /cc @rowanmiller

ajcvickers:
:shipit:

PR #3970, opened 2015-12-03
Allow generic, non-boxing access to state entry values

ajcvickers:
Specifically, current values (including shadow), original values, and relationship snapshot values

This is a building block for some upcoming changes to the identity cache. The code is not used much yet.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
@@ -225,7 +225,41 @@ namespace Microsoft.Data.Entity.ChangeTracking.Internal
         internal static readonly MethodInfo ReadShadowValueMethod
             = typeof(InternalEntityEntry).GetTypeInfo().GetDeclaredMethod(nameof(ReadShadowValue));
 
-        public virtual object ReadShadowValue(int shadowIndex) => null;
+        [UsedImplicitly]
+        protected virtual T ReadShadowValue<T>(int shadowIndex) => default(T);
+
+        internal static readonly MethodInfo ReadOriginalValueMethod
+            = typeof(InternalEntityEntry).GetTypeInfo().GetDeclaredMethod(nameof(ReadOriginalValue));
+
+        [UsedImplicitly]
+        private T ReadOriginalValue<T>(IProperty property, int originalValueIndex) 
+            => _originalValues.GetValue<T>(this, property, originalValueIndex);
+
+        internal static readonly MethodInfo ReadRelationshipSnapshotValueMethod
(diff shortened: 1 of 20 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -225,7 +225,41 @@ public virtual void MarkAsTemporary([NotNull] IProperty property, bool isTempora
         internal static readonly MethodInfo ReadShadowValueMethod
             = typeof(InternalEntityEntry).GetTypeInfo().GetDeclaredMethod(nameof(ReadShadowValue));
 
-        public virtual object ReadShadowValue(int shadowIndex) => null;
+        [UsedImplicitly]
+        protected virtual T ReadShadowValue<T>(int shadowIndex) => default(T);
+
+        internal static readonly MethodInfo ReadOriginalValueMethod
+            = typeof(InternalEntityEntry).GetTypeInfo().GetDeclaredMethod(nameof(ReadOrighinalValue));
```
anpete: "ReadOrighinalValue'

src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -225,7 +225,41 @@ public virtual void MarkAsTemporary([NotNull] IProperty property, bool isTempora
         internal static readonly MethodInfo ReadShadowValueMethod
             = typeof(InternalEntityEntry).GetTypeInfo().GetDeclaredMethod(nameof(ReadShadowValue));
 
-        public virtual object ReadShadowValue(int shadowIndex) => null;
+        [UsedImplicitly]
+        protected virtual T ReadShadowValue<T>(int shadowIndex) => default(T);
```
anpete: private?

src/EntityFramework.Core/ChangeTracking/Internal/Snapshot.cs
```diff
@@ -4656,6 +4856,9 @@ public struct Snapshot<T0>
 
         private T0 _value0;
 
+        public T GetValue<T>(int index)
+            => ((Func<Snapshot<T0>, T>)_valueReaders[index])(this);
```
anpete: Can this case be removed by changing the type of the array above?

anpete:
:shipit:

src/EntityFramework.Core/ChangeTracking/Internal/Snapshot.cs
ajcvickers: Not really, because the T is not a class generic, and this is because the calling code is supplying T, which as far as the compiler is concerned may be different from T0, even though we ensure at runtime that they are the same.

PR #3971, opened 2015-12-03
[TINY] More command info when multiple contexts found

### LINKED ISSUE #1981: Migrations :: add information how to deal with multiple contexts
If project has multiple contexts defined and one tries to perform migration operation the following error pops up:

```
System.InvalidOperationException: More than one DbContext was found. Specify which one to use.
```

However when I display migration tool info it doesn't say how to specify the context. We should probably add this information to the initial message

rowanmiller:
Fixes #1981


lajones:
:shipit:

PR #3972, opened 2015-12-03
[SMALL] Align Infrastructure/Internal types

### LINKED ISSUE #3374: Be consistent about namespaces
`ModelSource` is in MDE.Infrastructure
`SqlServerModelSource` is in MDE.Internal
`SqliteModelSource` is in MDE.Infrastructure.Internal

rowanmiller:
Equivalent types for SQLite and SQL Server were in different namespaces.
Applying our rule that types should follow the namespace of the type
they inherit (plus a `.Internal` if they are "internal").
Fixes #3374


lajones:
:shipit:

PR #3973, opened 2015-12-04
Reverse engineer sequences for SQL Server

### LINKED ISSUE #3510: Scaffolding: reverse engineer sequences
Scaffolding currently only generates code for identity columns. Detect sequences and add the appropriate annotations to code gen.

natemcmaster:
Fix #3510 

~~NB important change: `ISequence.IncrementBy` changed from `int` to `long` because `INCREMENT BY` can be greater than the range of an integer. The affected migrations and model building APIs.~~

### DIFF
--- a/src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
+++ b/src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
@@ -156,4 +156,10 @@
   <data name="ExistingFiles" xml:space="preserve">
     <value>The following file(s) already exist in directory {outputDirectoryName}: {existingFiles}. Use the Force flag to overwrite these files.</value>
   </data>
+  <data name="SequencesRequireName" xml:space="preserve">
+    <value>Sequence name cannot be null or empty. Entity Framework cannot model a sequence that does not have a name.</value>
+  </data>
+  <data name="BadSequenceType" xml:space="preserve">
+    <value>For sequence '{sequenceName}'. Unable to scaffold because it uses an unsupported type: '{typeName}'.</value>
+  </data>
 </root>
--- a/src/EntityFramework.Relational/Metadata/Sequence.cs
+++ b/src/EntityFramework.Relational/Metadata/Sequence.cs
@@ -18,6 +18,14 @@ namespace Microsoft.Data.Entity.Metadata
         private readonly IModel _model;
         private readonly string _annotationName;
 
+        public static readonly Type DefaultClrType = typeof(long);
+        public const int DefaultIncrementBy = 1;
+        public const int DefaultStartValue = 1;
+
+        public static readonly long? DefaultMaxValue = default(long?);
+        public static readonly long? DefaultMinValue = default(long?);
+        public static readonly bool DefaultIsCyclic = default(bool);
+
         public Sequence(
             [NotNull] IMutableModel model,
             [NotNull] string annotationPrefix,
@@ -38,9 +46,9 @@ namespace Microsoft.Data.Entity.Metadata
                 {
                     Name = name,
                     Schema = schema,
-                    ClrType = typeof(long),
-                    IncrementBy = 1,
-                    StartValue = 1
+                    ClrType = DefaultClrType,
+                    IncrementBy = DefaultIncrementBy,
+                    StartValue = DefaultStartValue
                 });
             }
         }
@@ -131,16 +139,21 @@ namespace Microsoft.Data.Entity.Metadata
             }
         }
 
+        public static IReadOnlyCollection<Type> SupportedTypes { get; } = new[]
+        {
+            typeof(byte),
+            typeof(long),
+            typeof(int),
+            typeof(short)
+        };
+
         public virtual Type ClrType
         {
             get { return GetData().ClrType; }
             [param: NotNull]
             set
             {
-                if ((value != typeof(byte))
-                    && (value != typeof(long))
-                    && (value != typeof(int))
-                    && (value != typeof(short)))
+                if (!SupportedTypes.Contains(value))
                 {
                     throw new ArgumentException(RelationalStrings.BadSequenceType);
                 }
--- a/test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/App.config
+++ b/test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/App.config
@@ -1,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<!-- binding redirect so TD.NET gets the right assembly version -->
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="System.Collections.Immutable" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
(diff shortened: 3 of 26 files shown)


natemcmaster:
cc @lajones

natemcmaster:
cc @bricelam (this causes signature changes to migrations)

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/ReverseEngineering/ExpectedResults/SequenceContext.expected
```diff
@@ -0,0 +1,31 @@
...
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+        {
+            modelBuilder.HasSequence("CountByTwo").IncrementsBy(2);
+
+            modelBuilder.HasSequence("CyclicalCountByThree")
+                .StartsAt(6)
+                .IncrementsBy(3)
+                .HasMax(27)
+                .HasMin(0)
```
lajones: Nit: I know it doesn't matter but it would read more clearly to me if we put `HasMin()` before `HasMax()`.

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/App.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="System.Collections.Immutable" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+        <bindingRedirect oldVersion="1.1.36.0" newVersion="1.1.37.0"/>
```
lajones: Back to this again, huh? Could we add a comment saying why - just so others know what's going on. Similarly for other App.configs.

src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
```diff
@@ -156,4 +156,10 @@
   <data name="ExistingFiles" xml:space="preserve">
     <value>The following file(s) already exist in directory {outputDirectoryName}: {existingFiles}. Use the Force flag to overwrite these files.</value>
   </data>
+  <data name="SequencesRequireName" xml:space="preserve">
+    <value>Sequence name cannot be null or empty. EF cannot model a sequence that does not have a name.</value>
```
lajones: Nit: I believe we're supposed to spell out Entity Framework rather than EF.

src/EntityFramework.Relational.Design/Properties/RelationalDesignStrings.resx
```diff
@@ -156,4 +156,10 @@
...
+  </data>
+  <data name="BadSequenceType" xml:space="preserve">
+    <value>For sequence '{sequenceName}'. Unable to scaffold because it uses and unsupported type: '{typeName}'.</value>
```
lajones: Typo: "and" should be "an"

lajones:
Looks good to me. I'll leave @bricelam to give the final sign-off.

src/EntityFramework.Relational/Metadata/ISequence.cs
```diff
@@ -10,7 +10,7 @@ public interface ISequence
         string Name { get; }
         string Schema { get; }
         long StartValue { get; }
-        int IncrementBy { get; }
+        long IncrementBy { get; }
```
bricelam: Does this need to change? Incrementing by anything over two billion seems a little ridiculous...

bricelam:
Migrations changes seem fine assuming we really want to change from `int` to `long` in the core API.

src/EntityFramework.Relational/Metadata/ISequence.cs
natemcmaster: I agree it's probably an unlikely scenario, but this is still theoretically possible in SQL Server. 

```sql
CREATE SEQUENCE BigStepSequence INCREMENT BY 300000000000 -- works in SQL Server
```

natemcmaster:
If we would rather keep `int`, then there is potential for data loss in reverse engineering. The risk is likely minimal since people rarely increment sequences by a step greater than the range of `Int32`

src/EntityFramework.Relational/Metadata/ISequence.cs
bricelam: `long` is fine. The SQL standard doesn't really prescribe the type beyond "numeric".

test/EntityFramework.MicrosoftSqlServer.Design.FunctionalTests/App.config
natemcmaster: TD.NET, that's why. I'll add as a comment.

natemcmaster:
Updated with typo fixes. @bricelam is that a sign off then?

src/EntityFramework.Relational/Metadata/ISequence.cs
ajcvickers: @bricelam @natemcmaster We did actually discuss this when designing ISequence and decided to keep it as int even though long is possible. I don't mind too much one way or the other, but the team previously came out on the side of int.

src/EntityFramework.Relational/Metadata/Sequence.cs
```diff
@@ -131,16 +139,21 @@ public virtual int IncrementBy
             }
         }
 
+        public static ISet<Type> SupportedTypes { get; } = new HashSet<Type>
```
bricelam: Do we expose `ISet<T>` properties anywhere else? Seems like `IReadOnlyCollection<T>` would be more idiomatic.

bricelam:
LGTM

src/EntityFramework.Relational/Metadata/ISequence.cs
natemcmaster: I'm pretty much impartial here too. Knowing this was a deliberate design choice changes my opinion. I'd rather keep the interface the same and side the team choice of using `int`. Ass increments > 2 bil are probably never done, I'll revert to use `int` instead.

src/EntityFramework.Relational/Metadata/Sequence.cs
natemcmaster: Good catch. I'll update.

src/EntityFramework.Relational/Metadata/ISequence.cs
divega: I would be fine either way, as long as we are consistent across the stack.

natemcmaster:
Updated.

src/EntityFramework.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -494,7 +494,7 @@ public class MigrationsSqlGenerator : IMigrationsSqlGenerator
         protected virtual void SequenceOptions(
             [CanBeNull] string schema,
             [NotNull] string name,
-            [NotNull] int increment,
+            [NotNull] long increment,
```
bricelam: Missed one.

natemcmaster: Good catch. Updated.

bricelam:
:shipit:

Issue #3186, opened 2015-09-22
Query: [Nav Prop Translation] Optional navigation translates to INNER JOIN instead of LEFT OUTER JOIN


kennywangjin:
I have model: 

```
public Class Item

{

    public int Id {get; set;}
   
    public int? ParentId {get; set;}
    
    public virtual Item Parent {get;set;}

}
```
when the query dbContenxt.Items.Where(item=>item.ParentId==null || item.Parent.ParentId==null)

it returns just the items matching the second filter: item.Parent.ParentId==null, the first filter doesn't work!
if I remove the second filter and just use the first one, then I could get the items matching the first filter.

Is it a bug or any misunderstanding of the framework?

maumar:
This is a legitimate bug. The problem is that we create the following sql:

```
SELECT [i].[Id], [i].[Name], [i].[ParentId]
FROM [Item] AS [i]
INNER JOIN [Item] AS [i.Parent] ON [i].[ParentId] = [i.Parent].[Id]
WHERE [i].[ParentId] IS NULL OR [i.Parent].[ParentId] IS NULL
```

where the correct sql would be:

```
SELECT [i].[Id], [i].[Name], [i].[ParentId]
FROM [Item] AS [i]
LEFT JOIN [Item] AS [i.Parent] ON [i].[ParentId] = [i.Parent].[Id]
WHERE [i].[ParentId] IS NULL OR [i.Parent].[ParentId] IS NULL
```

(LEFT JOIN instead of INNER JOIN)

cda-1:
Hey @maumar , I just want to be sure you know that the inner join problem crops up for all cases of projection on an optional navigation property, not just OP's case where it's in a Where clause. It seems you are aware but I wanted to be 100% sure and offer another example:
```c#
    return await Properties
        //.OrderBy(p => p.Id)
        //.Skip(20)
        .Select(p => new PropertyDto
        {
            Id = p.Id,
            ClientId = p.ClientId,
            ClientName = p.Client.Name,
            Name = p.Name,
            Address1 = p.Address1,
            Address2 = p.Address2,
            City = p.City,
            State = p.State,
            Zip = p.Zip,
            Phone = p.Phone,
            Fax = p.Fax,
            LocalPropertyId = p.LocalPropertyId,
            EmailAddresses = p.EmailAddresses,
            OwnerName = p.OwnerName,
            SendFileClient = p.SendFileClient,
            Units = p.Units,
[...]

PR #3974, opened 2015-12-04
Adding (disabled) tests for #3186 and associated fixes

maumar:
- Enabling ComplexNavigationTests to run on InMemory.

- Adding (disabled) tests for #3186

- Fixing bug in query that manifests when running InMemory - when creating joins, where key selectors have different types (e.g. int and int?) we were making incorrect casting at times which would cause exceptions if one of the values was null. Fix is to always convert to nullable type.

- Fixing bug in nav prop optimization (which was changing order.Customer.Id into order.CustomerId - before we used to do this for nav props which were only one level, non-composite, and which access PK on the other side of the navigatio (turning this into FK call). However we should not be doing that for optional navigations. In that case the initial Expression type is different than type after optimization (e.g. optional FK being nullable int, and PK being int). We could try to compensate for this in some cases (join, binary expression etc), but for other cases it's problematic (e.g. when projecting into anonymous type). Fix is to reduce the scope of the optimization, and only apply it for required navigations.

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -254,7 +254,8 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors.Internal
                             if (navigations.Any())
                             {
                                 if ((navigations.Count == 1)
-                                    && navigations[0].IsDependentToPrincipal())
+                                    && navigations[0].IsDependentToPrincipal() 
+                                    && navigations[0].ForeignKey.IsRequired)
                                 {
                                     var foreignKeyMemberAccess = CreateForeignKeyMemberAccess(node, navigations[0]);
                                     if (foreignKeyMemberAccess != null)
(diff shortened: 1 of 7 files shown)


maumar:
@anpete

maumar:
new version up

src/EntityFramework.Core/EntityFrameworkServiceCollectionExtensions.cs
```diff
@@ -156,7 +156,8 @@ private static IServiceCollection AddQuery(this IServiceCollection serviceCollec
                 .AddScoped(p => GetProviderServices(p).EntityQueryableExpressionVisitorFactory)
                 .AddScoped(p => GetProviderServices(p).ExpressionPrinter)
                 .AddScoped(p => GetProviderServices(p).ResultOperatorHandler)
-                .AddScoped(p => GetProviderServices(p).ProjectionExpressionVisitorFactory);
+                .AddScoped(p => GetProviderServices(p).ProjectionExpressionVisitorFactory)
+                .AddScoped(p => GetProviderServices(p).NavigationRewritingExpressionVisitorFactory);
```
anpete: We should discuss this. I would prefer to keep this in core if possible.

maumar: We could do this, but then we would have to compensate for the type changes in a few places (I know of join and binary expressions but there might be more). I will send a new pull request with that alternative, see which one you like better.

maumar:
@anpete new version up

anpete:
:shipit:

Issue #3843, opened 2015-11-22
Reverse Enginer: Support type aliases (user-defined data types)


jonnybee:
I tried to reverse engineer the AdventureWorks 2008 database and got errors on custom datatype is SQL Server during reverse enginneer. 

"Could not find type mapping for column 'Person.AddressType.Name' with data type 'Name'. Skipping column."

Moved the following to #3941 ~~Reverse engineering also failed to create default value on Decimal datatype: 
"For column Purchasing.ShipMethod.ShipBase unable to convert default value ((0.00)) into type System.Decimal. Will not generate code setting a default value for the property ShipBase on entity type ShipMethod."~~ 

This did work in EF6 and should also work in EF7!

rowanmiller:
@lajones - if a custom type is just an alias for a built-in type (i.e. it's a "user-defined data type" on SQL Server) then we should be able to scaffold it right? We just use the type mapper to get the CLR type for the property based on the underlying built-in type... and then specify the user-defined data type using the Fluent API.

```SQL
CREATE TYPE [dbo].[Name] FROM [varchar](max) NOT NULL
```

lajones:
@rowanmiller @natemcmaster  - re aliased data types - if they are well-known and only ever mean the same thing on different instances of SQL Server (e.g. `Name` always means `varchar(max)`) then it would be easy to hard-code them into the type mapper. But if we need to run a query to look up the underlying data types for aliased data-types on that particular instance then I would need to look into how that metadata works.

Re the decimal value - that's a bug - that looks like a perfectly well-defined decimal default value to me. I should investigate why the code is not picking it up.

jonnybee:
These types are typically defined in a database / domain  specific manner as User Defined Data Types, 

Example from AdventureWorks: 

- AccountNumber 	: NVARCHAR(15)
- Flag			: BIT, NOT NULL
- Name			: NVARCHAR(50)
- OrderNumber	: NVARCHAR(25)
- Phone			: NVRACHAR(25) 

In my opinion EF should be able to map these directly to the actual type - not as an alias in the type mapping.

jonnybee:
@lajones 
The issue with default value is locale/globalization.

This fails: 

    Thread.CurrentThread.CurrentCulture = CultureInfo.GetCultureInfo("nb-NO");
    Convert.ChangeType("0.00", typeof(Decimal));

You must use the overload that specifies an IFormatProvider to make sure that the proper format is used. 

[...]

PR #3975, opened 2015-12-04
RevEng: Support Type Aliases

lajones:
Fix for #3843 - allow columns to use user-defined data types as the column type.

Note: I'm open to reducing this down to only the `SqlServerTypeMapper`, `SqlServerDatabaseModelFactory` etc. But I wanted to put this wider approach out there so that other providers could re-use the same architecture i.e. they would only have to provide their own implementation of `XyzDatabaseModelFactory.GetUserDefinedDataTypes()` and after that it would "just work". The price is `GetUserDefinedDataTypes()` has to run before the first call to `Create()` and so it has to manage its own connection.

### DIFF
--- a/src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
+++ b/src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
@@ -61,6 +61,7 @@ namespace Microsoft.Data.Entity.Scaffolding
                  // TODO actually load per-user
                 _databaseModel.DefaultSchemaName = "dbo";
 
+                GetTypeAliases();
                 GetTables();
                 GetColumns();
                 GetIndexes();
@@ -68,6 +69,43 @@ namespace Microsoft.Data.Entity.Scaffolding
                 return _databaseModel;
             }
         }
+
+        private void GetTypeAliases()
+        {
+            var command = _connection.CreateCommand();
+            command.CommandText = @"SELECT
+  [type_name],
+  [underlying_system_type]
+  FROM
+  (SELECT
+    t1.[name] as [type_name],
+    ( CASE WHEN t1.[xusertype] = t1.[xtype] THEN NULL
+      ELSE
+        ( SELECT t2.[name]
+          FROM [sys].[systypes] AS t2
+          WHERE t2.[xusertype] = t2.[xtype]
+          AND t2.[xusertype] = t1.[xtype] )
+      END) as [underlying_system_type]
+    FROM [sys].[systypes] AS t1
+  ) AS t
+  WHERE [underlying_system_type] IS NOT NULL";
+
+            using (var reader = command.ExecuteReader())
+            {
+                while (reader.Read())
+                {
+                    var alias = reader.GetString(0);
+                    var underlyingSystemDataType = reader.GetString(1);
+                    var typeAlias = new TypeAliasModel
+                    {
+                        Alias = alias,
+                        SystemType = underlyingSystemDataType
+                    };
+
+                    _databaseModel.TypeAliases.Add(typeAlias);
+                }
+            }
+       }
 
         private void GetTables()
         {
--- a/src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
+++ b/src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
@@ -35,9 +35,6 @@ namespace Microsoft.Data.Entity.Storage.Internal
         private readonly RelationalTypeMapping _decimal = new RelationalTypeMapping("decimal(18, 2)", typeof(decimal));
         private readonly RelationalTypeMapping _time = new RelationalTypeMapping("time", typeof(TimeSpan));
 
-        private readonly Dictionary<string, RelationalTypeMapping> _simpleNameMappings;
-        private readonly Dictionary<Type, RelationalTypeMapping> _simpleMappings;
-
         public SqlServerTypeMapper()
         {
             _simpleNameMappings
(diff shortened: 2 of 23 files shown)


lajones:
Note to provider-writers - with this approach you will need to implement ` XyzDatabaseModelFactory.GetUserDefinedDataTypes(string)` - it's safe to just return null if you have no such types.

roji:
Here's an Npgsql perspective... PostgreSQL supports user-defined types beyond simple aliases (composite types, enums). Since these vary from database to database, it would be good if there were a way to load them as well (with a connection string). In other words, it's oddly specific to have a mechanism so specific for loading user-defined data types with a connection string while neglecting all other cases.

Also, the name "user-defined data types" has no meaning outside SqlServer (and seems to imply something much more than a simple alias). Perhaps "type aliases" is a more generic naming?

Finally, I may be missing stuff, but design-wise isn't it simpler/better to have the database model factory return type aliases as part the DatabaseModel returned by the Create method, rather than defining an additional separate GetUserDefinedDataTypes?

ilmax:
@roji if I understand the topic correctly, [Oracle](https://docs.oracle.com/cd/A91202_01/901_doc/server.901/a88856/c14ordb.htm) use the same name for the same thing, so (at least to me) the name UDT looks good.

roji:
@ilmax be careful, the EF7 is discussed user-defined *data* types here - these are simple aliases. User-defined types (without data) is a different SqlServer feature that is not supported. See https://github.com/aspnet/EntityFramework/issues/3843#issuecomment-161086035.

lajones:
@ilmax @roji is correct. These are just aliases for underlying system types - we are not supporting user-defined types just yet. And thanks @roji - I like the name `TypeAlias`.

Let me look into the idea of just adding `TypeAliases` to the `DatabaseModel` instead and I'll post a response here.

ilmax:
@roji thank you for the clarification, so you are right, a better name should help avoid confusion!

roji:
A development on something I said above...

Type aliases means that you support database-specific things in the TypeMapper - types that are present in one database but not in another. If you allow this, you may want to consider modelling the entire TypeMapper around this, it may solve issues such as newer types introduced in later versions of a database.

I know that allowing the TypeMapper to query the database for types has been discussed before. I'm only raising it again because type aliases are bringing this back in a restricted way.

src/EntityFramework.Relational.Design/IDatabaseModelFactory.cs
```diff
@@ -8,6 +9,7 @@ namespace Microsoft.Data.Entity.Scaffolding
 {
     public interface IDatabaseModelFactory
     {
+        Dictionary<string, string> GetUserDefinedDataTypes([NotNull] string connectionString);
```
natemcmaster: Since this is so SQL Server specific, we shouldn't put this on the interface.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -371,6 +371,47 @@ private void GetForeignKeys()
             }
         }
 
+        public virtual Dictionary<string, string> GetUserDefinedDataTypes(
```
natemcmaster: This should just be part of `DatabaseModel` (See https://github.com/aspnet/EntityFramework/pull/3783).

natemcmaster:
cc @ajcvickers (typemapping discussion)

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
```diff
@@ -110,12 +107,6 @@ public SqlServerTypeMapper()
 
         protected override string GetColumnType(IProperty property) => property.SqlServer().ColumnType;
 
-        protected override IReadOnlyDictionary<Type, RelationalTypeMapping> GetSimpleMappings()
```
natemcmaster: This removal is causing build errors.

lajones:
Suggest you all wait on the comments. As I mentioned above I'm going to re-do this based on the `TypeAlias` suggestion from @roji. I'm part way through and it addresses a lot of your comments @natemcmaster . Anyway I've removed the `needs-review tag` for now, I'll re-add it when I've updated the PR.

lajones:
@roji - this issue will just be for type aliases - as it is causing significant numbers of columns from SQL Server DBs to issue the "can't recognize the data type for this column so skipping it" warning message.

However all that does is say "this alias really means this underlying system data type, so map it to a C# type in the same way you would that type". If you'd like to have a general discussion about loading _all_ data types from the database itself I think that's a separate request - could you raise another issue if you'd like to discuss that please? At the moment I don't see how it would work - how would the DB know to which C# type to map its internal data types? But I could be missing something - so feel free to raise.

lajones:
@roji @natemcmaster Updated to a new commit using the `TypeAlias` approach and have re-added the `needs-review` tag.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerDatabaseModelFactory.cs
lajones: It is with the 2nd commit.

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
lajones: Yep - now fixed.

src/EntityFramework.Relational.Design/IDatabaseModelFactory.cs
lajones: Fixed with 2nd commit.

lajones:
Chatted with @natemcmaster and decided we need a more SQL Server specific approach. Closing this. Will open another PR with that approach.

roji:
Dealing with this as an SQL Server-specific feature might be the way to go.

I'm personally interested in general in whether the type mapper should have access to a specific database, i.e. accepting a connection or connection string. Aside from SQL Server type aliases which seem to require this, PostgreSQL also allows types to be defined in a specific database etc.

PR #3984, opened 2015-12-05
Generate using directives for reveng properties

roji:
When reverse engineering, providers may map database columns to CLR properties from arbitrary namespaces. Generate the required using directives for these.

Note that if the property type happens to come from another *assembly*, this will fail. This happens in Npgsql, which defines some types for PostgreSQL columns which don't correspond to built-in CLR types.


lajones:
This looks good. :shipit:

PR #3989, opened 2015-12-06
More precise index scaffolding warning

roji:
Missing property is only one reason for failing to scaffold an index. Providers may override VisitIndex and return null for other reasons, in which case it's their responsibility to log a warning.


lajones:
I'm fine with this. @natemcmaster any comments?

natemcmaster:
Also fine with this change. @roji can you first cleanup build errors? They are due to remaining references to `RelationalDesignStrings.UnableToScaffoldIndex` in *test\EntityFramework.MicrosoftSqlServer.Design.FunctionalTests\ReverseEngineering\SqlServerE2ETests.cs*

lajones:
@roji - do you think this will be checked in soon? I need to make the same change as part of the fix for #3710. Or if you prefer I can just include your change with mine?

roji:
Sorry for taking long with the cleanup, real life gets in the way... I've rebased on dev, corrected the reference and push-forced to this PR.

lajones:
No worries. That looks good. :shipit:

lajones:
@roji Shall I merge it?

roji:
@lajones sure thing, unless there are other issues it looks good to me...

PR #3995, opened 2015-12-07
Skip tests on Mono that fail due to external bugs

### LINKED ISSUE #3594: Fix the build on non-Windows platforms

natemcmaster:
Mono has numerous bugs in LINQ, SqlClient, remoting, and others. We thoroughly test on CoreCLR on X-Plat.

Passes on Mono 4.2.0.

Running korebuild does not work yet on Mono 4.0.x. Xunit hangs indefinitely on <4.2.0 (see https://github.com/xunit/xunit/issues/564) even with parallelization turned off.

### DIFF
--- a/test/EntityFramework.Core.FunctionalTests/TestUtilities/Xunit/MonoVersionConditionAttribute.cs
+++ b/test/EntityFramework.Core.FunctionalTests/TestUtilities/Xunit/MonoVersionConditionAttribute.cs
@@ -1,0 +1,79 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+
+namespace Microsoft.Data.Entity.FunctionalTests.TestUtilities.Xunit
+{
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = false)]
+    public class MonoVersionConditionAttribute : Attribute, ITestCondition
+    {
+        private Version _min;
+        private Version _max;
+        private Version _skip;
+        public string Min { get { return _min.ToString(); } set { _min = new Version(value); } }
+        public string Max { get { return _max.ToString(); } set { _max = new Version(value); } }
+        public string Skip { get { return _skip.ToString(); } set { _skip = new Version(value); } }
+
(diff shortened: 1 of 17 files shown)


test/EntityFramework.Relational.Tests/Migrations/Internal/MigrationsModelDifferTest.cs
```diff
@@ -317,7 +318,8 @@ public void Add_column_with_computed_value()
                 });
         }
 
-        [Theory]
+        [ConditionalTheory]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Odd type comparisons for empty byte arrays")]
```
lajones: Nit: Mention Mono in SkipReason?

test/EntityFramework.Core.FunctionalTests/TestUtilities/Xunit/MonoVersionConditionAttribute.cs
```diff
@@ -0,0 +1,79 @@
...
+                }
+
+                if (_min == null && _max == null)
+                {
+                    return true;
+                }
+
+                if (_min == null)
+                {
+                    return Current <= _max;
```
lajones: Current can return null -  what happens if it is null? Similarly in the comparisons below.

natemcmaster: The comparison would return null. This code should never be reached though.

test/EntityFramework.Relational.Tests/Migrations/Internal/MigrationsModelDifferTest.cs
natemcmaster: will update.

natemcmaster:
Updated.

test/EntityFramework.Core.FunctionalTests/TestUtilities/Xunit/MonoVersionConditionAttribute.cs
lajones: OK :shipit:

PR #3999, opened 2015-12-07
Upgrade source for navigation being used by foreign key attribute

### LINKED ISSUE #3832: ForeignKeyAttribute should upgrade sources for navigation based on which it was found
If explicit configuration is used to exclude a navigation in a relationship which is being pointed by `ForeignKey` attribute, we create relationship using that navigation by convention, while setting foreign key property on that which is found by convention, we match it with explicit relationship and the navigation is removed again. And this goes into infinite loop causing stack overflow.
Repro
````C#
public class Blog
    {
        public int Id { get; set; }
        
        public virtual IList<BlogMetadata> BlogMetadata { get; set; }
    }

    public class BlogMetadata
    {
        [ForeignKey("Blog")]
        public int Id { get; set; }
        
        public virtual Blog Blog { get; set; }
    }
        {
            modelBuilder.Entity<BlogMetadata>().HasOne<Blog>().WithMany(b => b.BlogMetadata);
        }
````

### PR DESCRIPTION
**Pull request by smitpatel**
Resolves #3832

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
@@ -68,13 +68,9 @@ namespace Microsoft.Data.Entity.Metadata.Internal
             if (newRelationshipBuilder != null
                 && newRelationshipBuilder.Metadata.Builder == null)
             {
-                var navigationToPrincipal = Metadata.DeclaringEntityType.FindNavigation(navigationToPrincipalName ?? "");
-                var navigationToDependent = Metadata.PrincipalEntityType.FindNavigation(navigationToDependentName ?? "");
-                if (navigationToDependent.ForeignKey == navigationToPrincipal.ForeignKey)
-                {
-                    return navigationToDependent.ForeignKey.Builder;
-                }
-                return null;
+                return FindForeignKey(
+                    navigationToPrincipalName,
+                    navigationToDependentName);
             }
 
             return newRelationshipBuilder;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs [hunk 1]
```diff
@@ -1212,11 +1218,25 @@ private bool CanSetPrincipalKey(IReadOnlyList<Property> properties, Configuratio
...
+                {
+                    return null;
+                }
+                var conflictingForeignKeyWithSameProperties = conflictingRelationships.FirstOrDefault(r => r.Metadata.Properties.SequenceEqual(dependentProperties));
+                if (conflictingForeignKeyWithSameProperties == null
+                    || !conflictingForeignKeyWithSameProperties.CanSetForeignKey(null, conflictingForeignKeyWithSameProperties.Metadata.DeclaringEntityType, configurationSource))
+                {
+                    return null;
+                }
+
```
AndriySvyryd: We should also return null if there's any other conflicting foreign key that can't be removed

smitpatel: Wouldn't above 2 ifs will take care of it?

smitpatel: No, it wouldn't. :cry:

src/EntityFramework.Core/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -1540,11 +1560,24 @@ private bool CanSetPrincipalKey(IReadOnlyList<Property> properties, Configuratio
                     ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyRemoved(removedForeignKey.Item1, removedForeignKey.Item2);
                 }
 
+                if (newRelationshipBuilder != null
```
AndriySvyryd: newRelationshipBuilder should never be null here

AndriySvyryd:
:shipit:

PR #4011, opened 2015-12-08
[Tiny] :bug: Do not run relationship discovery convention on entity type whi…

### LINKED ISSUE #3845: RelationshipDiscoveryConvention rediscovers entity while it is being ignored
I was playing around with async loading, and added the follow method to one of my Models:

	[NotMapped]
	public NotifyTaskCompletion<Geometry> StartImage
	{
		get; set;
	}

When I tried to add a migration (Add-Migration), the command never completed, and just kept chewing through CPU.

NotifyTaskCompletion is as follows:

	using System;
	using System.ComponentModel;
	using System.Threading.Tasks;

	namespace GymBuilder
	{
		public sealed class NotifyTaskCompletion<TResult> : INotifyPropertyChanged
		{
			public NotifyTaskCompletion(Task<TResult> task)
			{
				Task = task;
				if (!task.IsCompleted)
				{
					TaskCompletion = WatchTaskAsync(task);
				}
			}
	
			private async Task WatchTaskAsync(Task task)
			{
				try
				{
					await task;
				}
				catch {}
	
				var propertyChanged = PropertyChanged;
				if (propertyChanged == null) return;
	
				propertyChanged(this, new PropertyChangedEventArgs("Status"));
				propertyChanged(this, new PropertyChangedEventArgs("IsCompleted"));
				propertyChanged(this, new PropertyChangedEventArgs("IsNotCompleted"));
	
				if (task.IsCanceled)
				{
					propertyChanged(this, new PropertyChangedEventArgs("IsCanceled"));
				}
				else if (task.IsFaulted)
				{
					propertyChanged(this, new PropertyChangedEventArgs("IsFaulted"));
					propertyChanged(this, new PropertyChangedEventArgs("Exception"));
					propertyChanged(this, new 	PropertyChangedEventArgs("InnerException"));
					propertyChanged(this, new PropertyChangedEventArgs("ErrorMessage"));
				}
				else
				{
					propertyChanged(this, new 	PropertyChangedEventArgs("IsSuccessfullyCompleted"));
					propertyChanged(this, new PropertyChangedEventArgs("Result"));
				}
			}
	
			public Task<TResult> Task { get; private set; }
			public Task TaskCompletion { get; private set; }
			public TResult Result {	get	{ return (Task.Status == 	TaskStatus.RanToCompletion) ? Task.Result : default(TResult); } }
			public TaskStatus Status { get { return Task.Status; } }
			public bool IsCompleted { get { return Task.IsCompleted; } }
			public bool IsNotCompleted { get { return !Task.IsCompleted; } }
			public bool IsSuccessfullyCompleted { get { return Task.Status == 	TaskStatus.RanToCompletion; } }
			public bool IsCanceled { get { return Task.IsCanceled; } }
			public bool IsFaulted { get { return Task.IsFaulted; } }
			public AggregateException Exception { get { return Task.Exception; } }
			public Exception InnerException { get { return (Exception == null) ? null : 	Exception.InnerException; } }
			public string ErrorMessage { get { return (InnerException == null) ? null : 	InnerException.Message; } }
	
			public event PropertyChangedEventHandler PropertyChanged;
		}
	}

smitpatel:
…le it is being removed

Resolves #3845 
The issue: While removing entity type, we reset base type for all directly derived types. Setting base type triggers relationship discovery convention on oldbasetype and current entityType both. Since we were using `oldBaseType.Builder` which is set to null after base type is set for all derived types. It discovered relationships with ignored entity type too and added them to model again. The exception only appears when there is another derived type which gets foreign key with ignored type due to conventions. Since the related entity type builder is found using model builder, it returned null for ignored type.

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
@@ -293,7 +293,10 @@ namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
 
         public virtual bool Apply(InternalEntityTypeBuilder entityTypeBuilder, EntityType oldBaseType)
         {
-            var oldBaseTypeBuilder = oldBaseType?.Builder;
+            var oldBaseTypeBuilder = oldBaseType != null
+                && !entityTypeBuilder.ModelBuilder.IsIgnored(oldBaseType.Name, ConfigurationSource.Convention)
+                ? oldBaseType.Builder
+                : null;
             if (oldBaseTypeBuilder != null)
             {
                 ApplyOnRelatedEntityTypes(entityTypeBuilder.ModelBuilder, oldBaseTypeBuilder.Metadata);
@@ -313,7 +316,7 @@ namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
 
             foreach (var relatedEntityType in relatedEntityTypes)
             {
-                var relatedEntityTypeBuilder = modelBuilder.Entity(relatedEntityType.Name, ConfigurationSource.Convention);
+                var relatedEntityTypeBuilder = relatedEntityType.Builder;
                 Apply(relatedEntityTypeBuilder);
             }
         }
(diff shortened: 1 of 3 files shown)


src/EntityFramework.Core/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
```diff
@@ -293,7 +293,7 @@ private IReadOnlyList<RelationshipCandidate> FindRelationshipCandidates(Internal
 
         public virtual bool Apply(InternalEntityTypeBuilder entityTypeBuilder, EntityType oldBaseType)
         {
-            var oldBaseTypeBuilder = oldBaseType?.Builder;
+            var oldBaseTypeBuilder = oldBaseType != null ? entityTypeBuilder.ModelBuilder.Entity(oldBaseType.Name, ConfigurationSource.Convention) : null;
```
AndriySvyryd: Could you call `IsIgnored(oldBaseType.Name, oldBaseType.Name)` instead?

AndriySvyryd:
:shipit:

PR #4015, opened 2015-12-08
[Tiny] Block shadow keys created by convention

### LINKED ISSUE #3978: Self Referencing Entity
I'm not sure if this is a bug or just me doing something wrong but if I create a simple entity that can optionally have children of the same type, or a parent of the same type I'm unable to setup the model to produce the correct database.

I have a set of terms which can have child terms under them to build a taxonomy. When I create the migration it produces and additional column for what should be the FK to the parent term. I originally tried with annotations and later with fluent with the same issue.

This is the Term entity
```
    public class Term
    {
        public int Id { get; set; }
        public int CategoryId{ get; set; }
        public int? ParentId { get; set; }
        public Term Parent { get; set; }
        public ICollection<Term> Children { get; set; }
    }
```

Fluent model builder

```
        {

            builder.Entity<Term>()
                .HasKey(t => new { t.Id, t.CategoryId });

            builder.Entity<Term>()
                .HasOne(x => x.Parent)
                .WithMany(x => x.Children)
                .HasForeignKey(x => x.ParentId)
                .IsRequired(false);
        }
```

This is what a snippet of the migration looks like but the issue for me is the non nullable ParentId1
```
columns: table => new
                {
                    Id = table.Column<int>(nullable: false),
                    CategoryId = table.Column<int>(nullable: false),
                    Name = table.Column<string>(nullable: true),
                    ParentId = table.Column<int>(nullable: true),
                    ParentId1 = table.Column<int>(nullable: false)
                },
```

I've added a repo with the code to reproduce this to https://github.com/stevejgordon/EF7SelfRefTest

smitpatel:
Resolves #3978 
We had check in place, it just wasn't running. :smile:

### DIFF
--- a/src/EntityFramework.Core/Metadata/Conventions/Internal/CoreConventionSetBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Conventions/Internal/CoreConventionSetBuilder.cs
@@ -55,6 +55,7 @@ namespace Microsoft.Data.Entity.Metadata.Conventions.Internal
 
             conventionSet.ModelBuiltConventions.Add(new ModelCleanupConvention());
             conventionSet.ModelBuiltConventions.Add(keyAttributeConvention);
+            conventionSet.ModelBuiltConventions.Add(keyConvention);
             conventionSet.ModelBuiltConventions.Add(new PropertyMappingValidationConvention());
 
             conventionSet.NavigationAddedConventions.Add(new RequiredNavigationAttributeConvention());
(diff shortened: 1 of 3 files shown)


AndriySvyryd:
We also have `ModelValidator.EnsureNoShadowKeys`. Why isn't that throwing?

AndriySvyryd:
:shipit:

smitpatel:
filed #4016

PR #4023, opened 2015-12-09
Update exception msg for circular dependency

### LINKED ISSUE #3147: Better exception message for circular dependency in FK cycle
Hi, 
Working with the new dnx-clr-win-x86.1.0.0-Beta7.

I'm getting the following message : 
InvalidOperationException: A circular dependency was detected: 'Schema' {'FutureId'} -> 'Occurrence' {'ID'}, 'Occurrence' {'SchemaID'} -> 'Schema' {'ID'}, 'OccurrenceHistory' {'OccurrenceID'} -> 'Occurrence' {'ID'}.

Where classes architecture is :
- Schema holds all Occurrences under "OccurrenceHistory" enumerable.
- Schema holds Future instance of the occurrence.
- Schema holds Past instance of that occurrence.
(that is, future is the next link in the chain, and past is the previous link in the chain. the history refers to all links whatsoever)

So, i'm not so sure why i get this error, it should work (unless i'm missing something), 
it worked on Beta6 but had some other issues, where it wasn't able to load the Future item with .Include.

Thanks,
Eliran.

smitpatel:
resolves #3147 
@rowanmiller

### DIFF
--- a/src/EntityFramework.Core/Internal/Multigraph.cs
+++ b/src/EntityFramework.Core/Internal/Multigraph.cs
@@ -271,8 +271,8 @@ namespace Microsoft.Data.Entity.Internal
                 // TODO: Support cycle-breaking?
 
                 var currentCycleVertex = predecessorCounts.First(p => p.Value != 0).Key;
-                var cycle = new List<TVertex>();
-                cycle.Add(currentCycleVertex);
+                var cyclicWalk = new List<TVertex>();
+                cyclicWalk.Add(currentCycleVertex);
                 var finished = false;
                 while (!finished)
                 {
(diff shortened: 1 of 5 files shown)


smitpatel:
Working on correcting the cycle.

rowanmiller:
:shipit:

smitpatel:
Updated cycle detection to remove tail of a cycle.

src/EntityFramework.Core/Internal/Multigraph.cs
```diff
@@ -292,18 +292,34 @@ public virtual IReadOnlyList<List<TVertex>> BatchingTopologicalSort()
                 }
                 cycle.Reverse();
```
AndriySvyryd: You can rename `cycle` to `cyclicWalk` and `lastCycle` to `cycle`

AndriySvyryd:
:shipit:

src/EntityFramework.Core/Internal/Multigraph.cs
smitpatel: :+1:

mikary:
I think the best option for this would be to change the interpretation of the `formatCycle` lambda. Right now it's expected to produce a string that is just the formatted data portion of the message, we could have it return the entire error message instead to support custom error messages like this one.

PR #4026, opened 2015-12-09
Query: Fix #3668 - Subqueries with parameters skip parameterization

### LINKED ISSUE #3668: Subqueries with parameters skip parameterization
This may be related to #3317.

If I create a query that I use as a subquery of another query, e.g.:

```c#
var petsOfType = _dbContext.PetOwners.Where(x => x.Pet.Type == petType)

var owners = _dbContext.People
    .Where(x => petsOfType.Where(d => d.PersonId == x.PersonId).Count() > 0)
    .ToList()
```

The parameters to the subquery are cached along with the query itself, which leads to wrong results from the database when the parameters change.  The results are correct for the first execution only.

The executed queries reflect the lack of parameterization and hint at the entire query being cached.

From SQL Profiler, first execution, petType = "dog":

```sql
SELECT [x].[PersonId], [x].[Name], [x].[PetPetId]
FROM [Person] AS [x]
WHERE (
	SELECT COUNT(*)
	FROM [PetOwner] AS [x]
	INNER JOIN [Pet] AS [x.Pet] ON [x].[PetId] = [x.Pet].[PetId]
	WHERE ([x.Pet].[Type] = 'dog') AND ([x].[PersonId] = [x].[PersonId])
) > 0
```

From SQL Profiler, second execution, petType = "cat":

```sql
SELECT [x].[PersonId], [x].[Name], [x].[PetPetId]
FROM [Person] AS [x]
WHERE (
	SELECT COUNT(*)
	FROM [PetOwner] AS [x]
	INNER JOIN [Pet] AS [x.Pet] ON [x].[PetId] = [x.Pet].[PetId]
	WHERE ([x.Pet].[Type] = 'dog') AND ([x].[PersonId] = [x].[PersonId])
) > 0
```

anpete:
- When queryables are captures in the closure, during parameterization we need to eval the member expression and then visit the sub-expression.

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
@@ -15,35 +15,53 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors.Internal
 {
     public class ParameterExtractingExpressionVisitor : ExpressionVisitor
     {
+        private static readonly TypeInfo _queryableTypeInfo = typeof(IQueryable).GetTypeInfo();
+
         public static Expression ExtractParameters(
             [NotNull] Expression expression,
             [NotNull] QueryContext queryContext,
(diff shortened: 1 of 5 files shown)


divega:
:shipit:

PR #4033, opened 2015-12-10
RevEng: Don't produce unnecessary HasIndex() API

lajones:
If it would already be assumed due to it supporting a primary or foreign key, but make allowance for the extra API which is necessary if the default name of the index would be different.

Fix for #3710.


lajones:
Needs a few changes - will re-open when done.

PR #4037, opened 2015-12-10
React to (and use) aspnet/Microsoft.Data.Sqlite#192

bricelam:
@natemcmaster

### DIFF
--- a/src/EntityFramework.Sqlite/Storage/Internal/SqliteDatabaseCreator.cs
+++ b/src/EntityFramework.Sqlite/Storage/Internal/SqliteDatabaseCreator.cs
@@ -6,15 +6,19 @@ using JetBrains.Annotations;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Migrations;
 using Microsoft.Data.Entity.Utilities;
+using Microsoft.Data.Sqlite;
 
 namespace Microsoft.Data.Entity.Storage.Internal
 {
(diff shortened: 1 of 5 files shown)


src/EntityFramework.Sqlite/Storage/Internal/SqliteDatabaseCreator.cs
```diff
@@ -32,7 +37,22 @@ public override void Create()
             Connection.Close();
         }
 
-        public override bool Exists() => true;
+        public override bool Exists()
+        {
+            using (var readOnlyConnection = _connection.CreateReadOnlyConnection())
```
natemcmaster: Seems like there could be complications here with in-memory databases.

bricelam: Hmm... I suspect it won't really matter. `Exists` will unexpected return `false`, but calling `EnsureCreated` will behave as expected.

natemcmaster: In this scenario, wouldn’t this try to recreate already existing tables?

bricelam: No, databases are created by SQLite when you open the connection (see the `Create` method above). If the database already exists, this will no-op. Creating the tables, etc. is done in a separate step.

natemcmaster:
:shipit:

Issue #4038, opened 2015-12-10
Code templates for scaffolding entity types and DbContext from an existing database


rowanmiller:
We have one seam at the moment where folks can get a hold of the IModel we build from the database and manipulate it. We haven't really done any usability testing on this to see if it is good for changing table name conventions, etc.. 

We will also want some higher level extensibility points, such as code generation templates.

abelepereira:
Where is this IModel manipulation being allowed? I guess some sort of extension over the default ef command (EntityFramework.Commands) need to be implemented, but where? Also: does this "extension" for model customization implies creating a new command assembly or will the default EntityFramework.Commands allow some sort of DI?

Regards

rowanmiller:
The seam we have at the moment is designed for providers to use... so yes, you probably need to implement at least some parts of the provider to use it. This issue is about creating some high level seams that can be used by the application developer.

dazinator:
Just wanted to check my understanding - as of the present release there is no way for an application developer to customise the code generated via scaffolding (a provider writer can), although in earlier releases there was a razor based template approach which has since been pulled? So essentially to do stuff like pluralisation etc we are waiting for the resolution of this issue - is that correct?

rowanmiller:
@dazinator yes, that is correct. Post the 7.0.0 release we plan to provide customizable code-gen templates and code based interception points where you can customize the reverse engineered model and how it is written out to code.

idg10:
One problem with this approach is that the existing model seems to support just one Name for an entity, which is used for both the generated type, and also for the property on the context. This means that if you want an entity type called `Customer` but and you want the `DbSet<Customer>` on your context to be called `Customers`, it doesn't look like you can do it, because the code gen pulls the name of both the type and that property from the `IEntitySet.Name` property.

I've worked through an attempt to singularize some names. (I'm working with a database where all the tables have plural names. Leaving this as is produces correct-sounding dbset properties, but wrong-sounding entity types. E.g., `new Customers` is a weird way to create a single customer. Singular names produce wrong-sounding dbset props, but I find that slightly less disturbing.)

[...]

PR #4039, opened 2015-12-10
RevEng: Don't produce HasIndex() API if it would already be assumed

lajones:
Due to supporting a primary key, but make allowance for the outputting
HasName() (on a primary key) or HasConstraintName() (on a foreign key) as needed.

Fix for #3710.

### DIFF
--- a/test/EntityFramework.Sqlite.Design.FunctionalTests/ReverseEngineering/Expected/AllFluentApi/OneToMany/OneToManyFluentApiContext.expected
+++ b/test/EntityFramework.Sqlite.Design.FunctionalTests/ReverseEngineering/Expected/AllFluentApi/OneToMany/OneToManyFluentApiContext.expected
@@ -14,7 +14,8 @@ namespace E2E.Sqlite
         {
             modelBuilder.Entity<OneToManyDependent>(entity =>
             {
-                entity.HasKey(e => new { e.OneToManyDependentID1, e.OneToManyDependentID2 });
+                entity.HasKey(e => new { e.OneToManyDependentID1, e.OneToManyDependentID2 })
+                    .HasName("sqlite_autoindex_OneToManyDependent_1");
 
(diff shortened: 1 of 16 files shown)


test/EntityFramework.Sqlite.Design.FunctionalTests/ReverseEngineering/Expected/AllFluentApi/OneToMany/OneToManyFluentApiContext.expected
```diff
@@ -14,7 +14,8 @@ namespace E2E.Sqlite
         {
             modelBuilder.Entity<OneToManyDependent>(entity =>
             {
-                entity.HasKey(e => new { e.OneToManyDependentID1, e.OneToManyDependentID2 });
+                entity.HasKey(e => new { e.OneToManyDependentID1, e.OneToManyDependentID2 })
+                    .HasName("sqlite_autoindex_OneToManyDependent_1");
```
natemcmaster: Can we suppress this behavior for SQLite? SQLite ignores attempts to set the constraint name on the primary key.

```sql
CREATE TABLE People ( Id INT, CONSTRAINT "PK_People" PRIMARY KEY ( Id ) );
-- done.
SELECT * FROM sqlite_master;
-- shows the index created is named "sqlite_autoindex_People_1"
```

lajones: Discussed this with @natemcmaster offline. SQLite won't do anything with it this extra API and it should cause no problems. This seems to be something we should address in a SQLite-specific codegen layer, but we don't have that at the moment pending decisions in #4038. So checking in as is. Will discuss as part of #4038.

natemcmaster:
:shipit:

lajones:
Checked in with commit dbb5c9f.

PR #4041, opened 2015-12-10
Fix a bug with concatenation of string and non-string, (e.g. c.Name + 5)

maumar:
Problem is twofold - compiler translates this into a BinaryAdd operation, with a string Concat(object, object) method info, and casts the int into object. When we translate query and encounter a object-valued constant we don't know how to translate it. Another issue is that when generating SQL, we need to explicitly cast the non-string to varchar (so that types match). Currently we are always ignoring casts (were treating them as means to have consistently typed expression trees, but they were not needed in actual SQL generation).

Fix is to add fragment translator for the BinaryExpression.Add that recognizes concatenation of string with non-string, unboxing the non-string constant and wrapping it in a (new) ExplicitCast node. This node is then translated into proper CAST ([non-string] AS [string-type]). String-type name is determined using appropriate RelationalTypeMapper

### DIFF
--- a/src/EntityFramework.Relational/Query/ExpressionTranslators/StringConcatTranslator.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionTranslators/StringConcatTranslator.cs
@@ -1,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Query.Expressions;
+
+namespace Microsoft.Data.Entity.Query.ExpressionTranslators
+{
+    public class StringConcatTranslator : IExpressionFragmentTranslator
+    {
+        private static MethodInfo _stringConcatMethodInfo = typeof(string).GetTypeInfo().GetDeclaredMethods(nameof(string.Concat))
+            .Where(m => m.GetParameters().Count() == 2 && m.GetParameters()[0].ParameterType == typeof(object) && m.GetParameters()[1].ParameterType == typeof(object))
+            .Single();
+
+
+        public virtual Expression Translate([NotNull] Expression expression)
+        {
+            var binaryExpression = expression as BinaryExpression;
+            if (binaryExpression != null && binaryExpression.NodeType == ExpressionType.Add && binaryExpression.Method == _stringConcatMethodInfo)
+            {
+                var newLeft = binaryExpression.Left.Type != typeof(string)
+                    ? new ExplicitCastExpression(HandleNullTypedConstant(binaryExpression.Left.RemoveConvert()), typeof(string))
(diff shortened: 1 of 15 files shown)


maumar:
@divega @mikary

divega:
:shipit: but I am still curious: if this was actually a regression how was it working before?

src/EntityFramework.Relational/Query/ExpressionTranslators/StringConcatTranslator.cs
```diff
@@ -0,0 +1,59 @@
...
+            .Where(m => m.GetParameters().Count() == 2 && m.GetParameters()[0].ParameterType == typeof(object) && m.GetParameters()[1].ParameterType == typeof(object))
+            .Single();
+
+
+        public Expression Translate([NotNull] Expression expression)
+        {
+            var binaryExpression = expression as BinaryExpression;
+            if (binaryExpression != null && binaryExpression.NodeType == ExpressionType.Add)
+            {
+                if (binaryExpression.Left.Type == typeof(string) && binaryExpression.Right.Type != typeof(string))
```
divega: Actually discussed this code a bit more in depth with @maumar. We believe it would be slightly cleaner (and possibly work for more cases) if we pivot just on the operation being a string concatenation and then we explicitly cast any arguments we find not to be of type string. I.e. no need to say "if left is string but right is not string or viceversa".

src/EntityFramework.Relational/Query/Expressions/ExplicitCastExpression.cs
```diff
@@ -0,0 +1,49 @@
...
+        protected override Expression VisitChildren(ExpressionVisitor visitor)
+        {
+            var newOperand = visitor.Visit(Operand);
+
+            return (newOperand != Operand)
+                ? new ExplicitCastExpression(newOperand, _type)
+                : this;
+        }
+
+        public override string ToString() => "CAST(" + Operand.ToString() + " AS " + _type.Name + ")";
```
divega: We talked about translating this to STR() in SQL Server, but STR leaves leading blanks so we would need LTRIM and in general we don't see much benefit.

src/EntityFramework.Relational/Query/ExpressionTranslators/StringConcatTranslator.cs
```diff
@@ -0,0 +1,52 @@
...
+            .Where(m => m.GetParameters().Count() == 2 && m.GetParameters()[0].ParameterType == typeof(object) && m.GetParameters()[1].ParameterType == typeof(object))
+            .Single();
+
+
+        public virtual Expression Translate([NotNull] Expression expression)
+        {
+            var binaryExpression = expression as BinaryExpression;
+            if (binaryExpression != null && binaryExpression.NodeType == ExpressionType.Add && binaryExpression.Method == _stringConcatMethodInfo)
+            {
+                var newLeft = binaryExpression.Left.Type != typeof(string)
```
divega: Nit: extract method and apply to both binaryExpression.Left and binaryExpression.Right?

PR #4043, opened 2015-12-11
Dissalow defining FKs on properties that are part of an inherited key

### LINKED ISSUE #2837: Dissalow defining FKs on properties that are part of an inherited key
Consider this model:
```
private class Burger
{
    public int Id { get; set; }
    public Bun Bun { get; set; }
}

private class Ingredient
{
    public int Id { get; set; }
}

private class Bun : Ingredient
{
    public Burger Burger { get; set; }
}
```

There's an FK from `Bun.Id` to `Burger.Id`, however there are no FKs defined from `Ingredient`. Should `Ingredient.Id` use its own value generator? Should it use value generator from `Burger.Id`? What if there's another type derived from `Ingredient` and it has an FK from `Id` to a different property? Should we prohibit defining FKs on properties that are part of an inherited key?.

AndriySvyryd:
Fixes #2837


smitpatel:
:shipit:

Issue #4036, opened 2015-12-10
DbContext.X.Where(x => x.YRef.ID == 1) throws "The binary operator Equal is not defined for the types Nullable..."


LonDat:
Classes:
```C#
public class X {
  public int ID { get; set; }
  public Y YRef { get; set; }
}
public class Y {
  public int ID { get; set; }
}
```
Using (exception when):
```
X xxx = dbContext.Xs.Single(x => x.YRef.ID == 1);
```
Exception:
```
System.InvalidOperationException occurred
  HResult=-2146233079
  Message=The binary operator Equal is not defined for the types 'System.Nullable`1[System.Int32]' and 'System.Int32'.
  Source=System.Core
  StackTrace:
       at System.Linq.Expressions.Expression.GetEqualityComparisonOperator(ExpressionType binaryType, String opName, Expression left, Expression right, Boolean liftToNull)
       at System.Linq.Expressions.Expression.Equal(Expression left, Expression right, Boolean liftToNull, MethodInfo method)
       at System.Linq.Expressions.Expression.MakeBinary(ExpressionType binaryType, Expression left, Expression right, Boolean liftToNull, MethodInfo method, LambdaExpression conversion)
       at System.Linq.Expressions.Expression.MakeBinary(ExpressionType binaryType, Expression left, Expression right, Boolean liftToNull, MethodInfo method)
       at Microsoft.Data.Entity.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.VisitBinary(BinaryExpression node) in C:\LonDatSources\Git\Repos\EntityFramework\src\EntityFramework.Core\Query\ExpressionVisitors\Internal\NavigationRewritingExpressionVisitor.cs:line 225
  InnerException: 

```
Callstack from debugging:
```
 	System.Core.dll!System.Linq.Expressions.Expression.GetEqualityComparisonOperator(System.Linq.Expressions.ExpressionType binaryType, string opName, System.Linq.Expressions.Expression left, System.Linq.Expressions.Expression right, bool liftToNull)	Unknown
 	System.Core.dll!System.Linq.Expressions.Expression.Equal(System.Linq.Expressions.Expression left, System.Linq.Expressions.Expression right, bool liftToNull, System.Reflection.MethodInfo method)	Unknown
[...]

PR #4045, opened 2015-12-11
Fix issue #4036 - DbContext.X.Where(x => x.YRef.ID == 1) throws "The binary operator Equal is not defined for the types Nullable..."

maumar:
Issue was that the navigation property -> foreign key optimization was too aggressive. Problem with the optimization is that sometimes FK is of different type than PK, and it may cause type inconsistencies.
We incorrectly assumed that it can be performed for required navigations. However one can declare required navigation that has nullable FK, while PK is non-nullable.

Fix is to compare types of PK and FK instead.


maumar:
@divega

bricelam:
:shipit:

Issue #3291, opened 2015-10-01
Add a guard to prevent interleaving of async calls


divega:
We did this in EF6 to prevent multiple threads from using the same context concurrently, and to guide users to the right async patterns.

See #3240.

mikary:
The EF6 implementation of guards around async calls appears to focus on wrapping top level API calls in semaphore checks via memory barrier (using Interlocked exchanges to increment and decrement a `ThrowingMonitor` instance attached to the `ObjectContext`).

EF6 updates the semaphore on the following operations:
 - RefreshEntitiesAsync
 - SaveChangesAsync
 - ExecuteStoreCommandAsync
 - ExecuteStoreQueryAsync
 - GetResultAsync (Query)

It also throws based on the semaphore value on these operations:
 - RefreshAsync
 - RefreshEntities
 - SaveChanges
 - SaveChangesAsync
 - ExecuteFunction
 - ExecuteStoreCommand
 - ExecuteStoreCommandAsync
 - ExecuteStoreQuery
 - ExecuteStoreQueryAsync
 - GetEnumerator (Query)
 - GetEnumeratorAsync (Query)
 - GetResult (Query)
 - GetResultAsync (Query)
 - ExecuteSqlQuery
 - ExecuteSqlQueryAsync
 - Find
 - FindAsync

Note: The `GetResultAsync` semaphore encapsulates query compilation and execution, but not enumeration due to the method returning an `ObjectResult<T>`

mikary:
Based on discussions with the team so far, this enhancement has the following requirements / considerations:
 - The guard should never throw when the context is being used properly
 - The async entry points for customer code should all be covered
 - There are no special cases (i.e. AsNoTracking queries are not thread safe)
 - The performance impact should be kept to a minimum

We can get the same effect as the EF6 `ObjectContext` from a scoped service with a similar implementation to `ThrowingMonitor`.

There are fewer entry points to the current EF7 implementation
 - SaveChanges (Async)
 - EnsureCreated (Async)
 - EnsureDeleted (Async)
 - GetEnumerator - on InternalDbSet
 - AsyncEnumerable - on InternalDbSet
 - ExecuteSqlCommand (Async)

Notes on Query:
[...]

PR #4051, opened 2015-12-11
Add concurrency detection for query, save changes, and ADO command execution

mikary:
Changes for #3291
@divega @ajcvickers @AndriySvyryd @anpete

### DIFF
--- a/src/EntityFramework.Core/Internal/ConcurrencyDetector.cs
+++ b/src/EntityFramework.Core/Internal/ConcurrencyDetector.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.Data.Entity.Internal
+{
+    public class ConcurrencyDetector : IConcurrencyDetector
+    {
+        private bool _isInCriticalSection;
+
+        public virtual void EnterCriticalSection()
+        {
+            if(_isInCriticalSection)
+            {
+                throw new NotSupportedException(CoreStrings.ConcurrentMethodInvocation);
+            }
+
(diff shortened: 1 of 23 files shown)


mikary:
For Query, it was not necessary to add guards on the `QueryCompiler` for the single result path because the guard is already compiled into the query.

src/EntityFramework.Core/Internal/ConcurrencyDetector.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Microsoft.Data.Entity.Internal
+{
+    public class ConcurrencyDetector : IConcurrencyDetector
+    {
+        private int _isInCriticalSection;
+
+        public virtual void EnterCriticalSection()
+        {
+            if(_isInCriticalSection++ != 0)
```
AndriySvyryd: Using the previous value when setting has more potential of ending up with an unexpected value in concurrency scenarios making this potentially less effective. Just use a bool and assign a constant

src/EntityFramework.Core/Properties/CoreStrings.resx
```diff
@@ -516,4 +516,7 @@
   <data name="PrincipalEntityTypeNotInRelationship" xml:space="preserve">
     <value>You are configuring a relationship between '{dependentEntityType}' and '{principalEntityType}' but have specified a foreign key targetting '{entityType}'. The foreign key must be targetting a type that is part of the relationship.</value>
   </data>
+  <data name="ConcurrentMethodInvocation" xml:space="preserve">
+    <value>A second operation started on this context before a previous asynchronous operation completed. Use 'await' to ensure that any asynchronous operations have completed before calling another method on this context. Any instance members are not guaranteed to be thread safe.</value>
```
ajcvickers: This message is no longer appropriate because we are not just doing this for async operations anymore.

ajcvickers:
:shipit: Looks good to me once comments addressed.

PR #4052, opened 2015-12-11
[2-liner] Commands: Look for Program then App if Startup not found

### LINKED ISSUE #3802: Commands: Look for ConfigureServices in Program
I created a Console project and configured it to read the connection string during Startup and the app ran fine as expected.  However, when I tried to use migrations using EF.Commands, I got an error saying that no database providers were configured.

To work around that I ended up creating a DbContextFactory which the Commands read and things worked fine.

https://github.com/dneimke/coffeestore/blob/develop/src/EF7Demo.CoffeeStore/Model/DbContextFactory.cs

I was surprised that I had to create the Factory as I do not need to do this for an ASP Web Application.  Is this a different behavior for apps other than ASP.NET Web Application and was I right to create the Factory to work around this?

bricelam:
Resolves #3802 

Verification: Manual (automation wasn't worth generating assemblies)


smitpatel:
:shipit:

PR #4054, opened 2015-12-11
Abstract identity maps for perf and more efficient memory usages

ajcvickers:
We currently have the concept of an IKeyValue to represent values of primary, alternate, and foreign keys. This interface is implemented by a struct, but is always referenced through the interface, which means values are always boxed and hence require heap allocations. This change introduces identity maps for the state manager and query buffer that, for simple keys, use the actual type of the key rather than an IKeyValue. This is achieved by identity maps specific for the key type with YAGI interfaces to perform operations on those identity maps efficiently.

There is still work left to do here:
- IKeyValue is still used by the in-memory provider
- The update pipeline uses IKeyValue
- Composite and binary keys are still represented by IKeyValue


ajcvickers:
@maumar @divega Could one of you guys look at this since Andrew is out?

bricelam:
:shipit:

ajcvickers:
Merged

PR #4055, opened 2015-12-12
Update in-memory store to not use IKeyValue

ajcvickers:
Follows the pattern used elsewhere of dynamically creating dictionaries with the actual key type. Also removed use of ImmutableCollections.


bricelam:
:shipit:

ajcvickers:
Merged

PR #4056, opened 2015-12-12
Make setting the primary key set IsRequired on properties via convention

### LINKED ISSUE #3460: DeleteBehavior is not updated to Cascade if requiredness is set on FK after the relationship is discovered
Deletes on required associations are supposed to cascade unless the user explicitly opts out. It seems that if the requiredness of the association is set after we first learn about it the delete behavior doesn't get updated. 

Repro:

``` C#
using System;
using System.Data.SqlClient;
using System.Linq;
using Microsoft.Data.Entity;

namespace ConsoleApplication20
{
    internal class Program
    {
        private static void Main(string[] args)
        {
            using (var context = new IdentityDbContext())
            {
                var fk =
                    context.Model.FindReferencingForeignKeys(context.Model.FindEntityType(typeof(Employee))).First();
                Console.WriteLine(
                    $"FK on Employee.Detail is required: {fk.IsRequired}");
                Console.WriteLine(
                    $"Delete behavior on Employee.Detail is {fk.DeleteBehavior}");  // Required
            }
        }
    }

    public class EmployeeDetail
    {
        public virtual string EmployeeId { get; set; }
        public virtual byte[] Photo { get; set; }
    }

    public class Employee
    {
        public virtual string Id { get; set; }
        public virtual EmployeeDetail Detail { get; set; }
    }

    public class IdentityDbContext : DbContext
    {
        public DbSet<Employee> Users { get; set; }

        {
            modelBuilder.Entity<Employee>(b =>
            {
                b.HasOne(u => u.Detail)
                    .WithOne()
                    .HasForeignKey<EmployeeDetail>(pd => pd.EmployeeId);
            });

            // if key configuration of EmployeeDetail happens before the association configuration 
            // delete behavior will correctly be set to cascade 
            modelBuilder.Entity<EmployeeDetail>(b =>
            {
                b.HasKey(r => r.EmployeeId);
            });
        }

        {
            optionsBuilder.UseLocalDB(GetType().Name);
        }
    }

    public static class DbContextOptionsExtensions
    {
        public static SqlServerDbContextOptionsBuilder UseLocalDB(this DbContextOptionsBuilder optionsBuilder,
            string databaseName = null, Action<SqlConnectionStringBuilder> action = null)
        {
            var connectionStringBuilder = new SqlConnectionStringBuilder
            {
                DataSource = "(localdb)\\mssqllocaldb",
                InitialCatalog = databaseName
            };
            action?.Invoke(connectionStringBuilder);
            return optionsBuilder.UseSqlServer(connectionStringBuilder.ConnectionString);
        }
    }
}
```

AndriySvyryd:
Fixes #3460

### DIFF
--- a/src/EntityFramework.Core/EntityFramework.Core.csproj
+++ b/src/EntityFramework.Core/EntityFramework.Core.csproj
@@ -116,6 +116,7 @@
     <Compile Include="Internal\IndentedStringBuilder.cs" />
     <Compile Include="Internal\ConcurrencyDetector.cs" />
     <Compile Include="Internal\LazyRef.cs" />
+    <Compile Include="Metadata\Conventions\Internal\IKeyRemovedConvention.cs" />
     <Compile Include="Metadata\Internal\ConventionalAnnotatable.cs" />
     <Compile Include="Metadata\Internal\ConventionalAnnotation.cs" />
     <Compile Include="Metadata\Internal\ICanGetNavigations.cs" />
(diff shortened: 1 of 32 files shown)


ajcvickers:
Is it now possible to build a model with a non-required primary key? If so, we should add a rule that throws for this in model validation.

AndriySvyryd:
Added a check to `AddKey` to ensure no property is nullable to complement a check in `IsNullable`

ajcvickers:
:shipit:

Issue #4074, opened 2015-12-14
TPH: Validate shared columns and constraints


bricelam:
It is possible to share columns, foreign keys, and indexes across a hierarchy. For example:
```C#
class Animal {
    public int Id { get; set; }
}

class Cat : Animal {
    public string Breed { get; set; }
}

class Dog: Animal {
    public string Breed { get; set; }
}
```
We should validate (or combine) the configuration on each of the properties. Because conflicts:
```C#
modelBuilder.Entity<Cat>().Property(c => c.Breed).HasMaxLength(30);
modelBuilder.Entity<Dog>().Property(d => d.Breed).HasMaxLength(15);
```

divega:
It the current behavior last one wins?

bricelam:
The only place it currently matters is Migrations. I just did `Distinct((x, y) => x.Name == y.Name)`, so one (undefined) of them will win.

bricelam:
:trollface: It's almost like we should maintain a separate model for SSpace with some sort of mapping between the two...

divega:
:see_no_evil:

rowanmiller:
Triage: For the moment we will throw if the two columns are not configured in the same way. We can relax this in the future.

PR #4075, opened 2015-12-14
Migrations: Allow shared columns and constraints within a hierarchy

### LINKED ISSUE #3061: Migrations: Allow many properties mapped to one column
This is to unblock the scenario where two derived types have properties with the same name that could be mapped to the same property in TPH.
For the first iteration we could just use the column generated from either property.
As a minimal safeguard we could throw if the generated columns are different in any way, without trying to merge the configurations.

bricelam:
The behavior on metadata conflicts is undefined. (See #4074)

Fixes #3061


ajcvickers:
:shipit:

src/EntityFramework.Relational/Metadata/RelationalEntityTypeExtensions.cs
```diff
@@ -1,22 +0,0 @@
...
-using System.Collections.Generic;
-using System.Linq;
-using JetBrains.Annotations;
-using Microsoft.Data.Entity.Metadata.Internal;
-
-namespace Microsoft.Data.Entity.Metadata
-{
-    public static class RelationalEntityTypeExtensions
-    {
-        public static IEnumerable<IForeignKey> GetForeignKeysInHierarchy([NotNull] this IEntityType entityType)
```
divega: I am personally ok with removing these extension methods from the public surface, but want to make sure we are all in agreement. 

cc @rowanmiller

bricelam: Additional context: They were added for (and only used internally by) the model differ to support TPH, but this PR required a more specialized usage so I removed these.

PR #4077, opened 2015-12-14
Stop using IKeyValue in update pipeline

ajcvickers:
Switched to use same key factories as other parts of the stack. This currently reuses internal code used in the core, but it could be changed to duplicate some of that code and only use public APIs.


bricelam:
:shipit:

ajcvickers:
Merged

PR #4084, opened 2015-12-14
Commands: Include types without default constructors in searches

### LINKED ISSUE #3752: Better exception message when DbContext does not have default ctor
Tested on Full .NET, in a unit test project, with EntityFramework.Commands v7.0.0-rc2-16292.

The DbContext class lives in a shared project (i.e. the glorified linked files project type) that is referenced by the unit test project, acting as the startup project.

Calling Add-Migration and providing only a name gives error:
No DbContext was found. Ensure that you're using the correct assembly and that the type is neither abstract nor generic.

Explicitly specifying the full name (includes namespace) of the DbContext with the -Context parameter and I get an error in the form:
No DbContext named 'MyNameSpace.MyDbContext' was found.

I thought at first the shared project might be causing problems, so I repeated the test in a new console project, exact same problem.  It's a public, non-abstract and non-generic class that was functioning perfectly fine in beta6 or 7 (I forget which now).

### LINKED ISSUE #3803: System.TypeLoadException when generating migration in UWP app
I'm generating a migration in my existing UWP app and I get a TypeLoad exception.

I have the following nuget packages installed for SQLite
```
"EntityFramework.Commands": "7.0.0-rc1-final",
"EntityFramework.SQLite": "7.0.0-rc1-final",
```

The full text of the error is
```
System.TypeLoadException: Could not find Windows Runtime type 'Microsoft.Graphics.Canvas.UI.Xaml.CanvasAnimatedControl'. ---> System.NotSupportedException: Types from custom Windows Runtime components are not supported in desktop applications.
   --- End of inner exception stack trace ---
   at System.Signature.GetSignature(Void* pCorSig, Int32 cCorSig, RuntimeFieldHandleInternal fieldHandle, IRuntimeMethodInfo methodHandle, RuntimeType declaringType)
   at System.Signature..ctor(IRuntimeMethodInfo methodHandle, RuntimeType declaringType)
   at System.Reflection.RuntimeConstructorInfo.GetParametersNoCopy()
   at System.Reflection.RuntimeConstructorInfo.GetParameters()
   at System.SharedTypeExtensions.<>c.<GetConstructibleTypes>b__19_1(ConstructorInfo c)
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at System.SharedTypeExtensions.<>c.<GetConstructibleTypes>b__19_0(TypeInfo t)
   at System.Linq.Enumerable.<>c__DisplayClass6_0`1.<CombinePredicates>b__0(TSource x)
   at System.Linq.Enumerable.WhereArrayIterator`1.MoveNext()
   at Microsoft.Data.Entity.Design.DbContextOperations.FindContextTypes()
   at Microsoft.Data.Entity.Design.DbContextOperations.FindContextType(String name)
   at Microsoft.Data.Entity.Design.DbContextOperations.CreateContext(String contextType)
   at Microsoft.Data.Entity.Design.MigrationsOperations.AddMigration(String name, String outputDir, String contextType)
   at Microsoft.Data.Entity.Design.OperationExecutor.<AddMigrationImpl>d__7.MoveNext()
   at Microsoft.Data.Entity.Design.OperationExecutor.OperationBase.<>c__DisplayClass4_0`1.<Execute>b__0()
   at Microsoft.Data.Entity.Design.OperationExecutor.OperationBase.Execute(Action action)
Could not find Windows Runtime type 'Microsoft.Graphics.Canvas.UI.Xaml.CanvasAnimatedControl'.
```

If it helps, the CanvasAnimatedControl comes from the Win2D nuget package and is not local to the project.

bricelam:
Instead of not finding anything, this will inform the user that a default constructor is missing (fixes #3752). This also avoids looking at constructor parameters whose types may not be available on UWP projects (fixes #3803).

Verification: In addition to the automated test, I manual verified the specific scenarios.

### DIFF
--- a/test/EntityFramework.Core.Tests/Utilities/TypeExtensionsTest.cs
+++ b/test/EntityFramework.Core.Tests/Utilities/TypeExtensionsTest.cs
@@ -9,6 +9,10 @@ using System.Linq;
 using System.Reflection;
 using Xunit;
 
+#if DNXCORE50
+using Moq;
+#endif
+
 namespace Microsoft.Data.Entity.Tests.Utilities
 {
     public class TypeExtensionsTest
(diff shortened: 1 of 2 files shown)


test/EntityFramework.Core.Tests/Utilities/TypeExtensionsTest.cs
```diff
@@ -442,5 +442,55 @@ private enum SomeEnum
...
+        }
+
+        private class SomeTypeWithoutDefaultCtor
+        {
+            public SomeTypeWithoutDefaultCtor(int value)
+            {
+            }
+        }
+
+        // TODO: Understand why this doesn't compile on dnxcore
```
divega: FWIW, the reason it does not compile for me is that the ref version of System.Reflection has an internal parameter-less constructor for Assembly. I don't know why is that, but it might be a reason to file a bug on that or to take a different testing approach.

divega:
A bit concerned with the test #if but otherwise :shipit:

bricelam:
It looks like the `System.Reflection` contract has an `internal` (not `protected`) default `Assembly` constructor. We can only mock on Full .NET.

bricelam:
Confirmed. Using Moq works on CoreCLR (but not Full.NET because of ISerializable). Will check-in with coverage for both.

divega:
I filed https://github.com/dotnet/corefx/issues/4997.

PR #4086, opened 2015-12-14
Make RelationalDatabaseCreator.CreateTables atomic

### LINKED ISSUE #3908: After fail of EnsureCreatedAsync method database will be in inconsistent state. Why we are not using transaction?
Hello,
I'm not sure that it's a bug, but at least some unexpected behavior.

In our code we are using something like that.
      using (var db = serviceProvider.GetService<MyShuttleContext>())
            {
                    var databaseCreated = await db.Database.EnsureCreatedAsync();
            }
as a result of this call we catch an exception with text "{"Introducing FOREIGN KEY constraint 'FK_Ride_Driver_DriverId' on table 'Ride' may cause cycles or multiple cascade paths. Specify ON DELETE NO ACTION or ON UPDATE NO ACTION, or modify other FOREIGN KEY constraints.\r\nCould not create constraint or index. See previous errors."}"

I would expect that database as a result will be empty, but in fact it was partially creates.

bricelam:
Fixes #3908

### DIFF
--- a/src/EntityFramework.Relational/Storage/RelationalDatabaseCreator.cs
+++ b/src/EntityFramework.Relational/Storage/RelationalDatabaseCreator.cs
@@ -69,10 +69,28 @@ namespace Microsoft.Data.Entity.Storage
         }
 
         public virtual void CreateTables()
-            => GetCreateTablesCommands().ExecuteNonQuery(Connection);
+        {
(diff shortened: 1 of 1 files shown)


lajones:
Looks good :shipit:

divega:
:+1:

Issue #3417, opened 2015-10-13
RevEng: Scaffolding fails silently on VB projects


ErikEJ:
It would be helpful to show an error message instead https://github.com/ErikEJ/EntityFramework7.SqlServerCompact/issues/242#issuecomment-147589772

rowanmiller:
We should just be generating C# (or a warning would be fine too).

lajones:
@rowanmiller @ErikEJ - I tried running Scaffold-DbContext against a VB project with build 7.0.0-rc2-16542 and that does _not_ fail. Instead it produces C# code in the VB project successfully. So if we want that functionality we just leave the code as is. However I have created a PR, see #4087, which does throw an error from Scaffold-DbContext if the user attempts to run it from a VB project, which we can use if preferred.

rowanmiller:
Showing an error is better... so lets get your PR in

lajones:
See comments in PR #4087. We decided to leave the code as is - for now it will generate C# code in a VB project.

midnait:
https://github.com/aspnet/EntityFramework/issues/4210

Migrations and Scaffold "translated" to VB.

PR #4087, opened 2015-12-15
[TINY] RevEng: Throw error if user attempts to run Scaffold-DbContext in a VB project.

lajones:
Fix for #3417.

### DIFF
--- a/src/EntityFramework.Commands/tools/EntityFramework.psm1
+++ b/src/EntityFramework.Commands/tools/EntityFramework.psm1
@@ -409,6 +409,9 @@ function Scaffold-DbContext {
     $values = ProcessCommonParameters $StartupProject $Project
     $dteStartupProject = $values.StartupProject
     $dteProject = $values.Project
+    if ([string]::Compare($dteProject.Kind.ToString(), "{F184B08F-C81C-45F6-A57F-5ABD9991F28F}", $True) -eq 0) {
+        throw "Project $Project is a Visual Basic project which is not yet supported by Scaffold-DbContext."
+    }
 
     $artifacts = InvokeOperation $dteStartupProject $Environment $dteProject ReverseEngineer @{
         connectionString = $Connection


src/EntityFramework.Commands/tools/EntityFramework.psm1
```diff
@@ -409,6 +409,9 @@ function Scaffold-DbContext {
     $values = ProcessCommonParameters $StartupProject $Project
     $dteStartupProject = $values.StartupProject
     $dteProject = $values.Project
+    if ([string]::Compare($dteProject.Kind.ToString(), "{F184B08F-C81C-45F6-A57F-5ABD9991F28F}", $True) -eq 0) {
+        throw "Project $Project is a Visual Basic project which is not yet supported by Scaffold-DbContext."
+    }
```
rowanmiller: Does this cover all VB.NET projects, or just class library?

bricelam:
Why doesn't it just spit out C# code? The project models are almost identical...

src/EntityFramework.Commands/tools/EntityFramework.psm1
ErikEJ: I would say all relevant project types: http://www.mztools.com/articles/2008/mz2008017.aspx

bricelam: In EF6, we detected using `project.CodeModel.Language == CodeModelLanguageConstants.vsCMLanguageVB`. I believe this caught all projects including Web Sites.

bricelam:
I don't like the idea of blocking users from doing this. In Migrations, they can override the `MigrationsCodeGenerator` design-time service with a VB-specific one and everything should work. It would be nice to have the same seam in RevEng.

bricelam:
I believe we said in the past that we wouldn't necessarily support VB, but we also wouldn't explicitly block it either.

src/EntityFramework.Commands/tools/EntityFramework.psm1
lajones: I'll test a few more project types, but the above code was taken from `VsUtils` in EF6.

lajones:
@bricelam As mentioned in the issue #3417 I'm happy to just leave the code as is - in which case it does successfully output C# code into the VB project. But @rowanmiller is in favor of the error message...

rowanmiller:
Oh, right, yeah I missed that we would be blocking someone from implementing support themselves... yeah we should not block that.

rowanmiller:
Ok, yeah... just C# is fine

lajones:
OK. I'll close this and update the issue to say we're leaving it as is.

PR #4099, opened 2015-12-16
Split out query compilation benchmarks

rowanmiller:
Removing compilation from general query tests and adding a separate
class with more targeted tests (we can add more as needed in the
future).

Also restructuring fixtures a little, so that scenario specific settings
(like disabling batching and disabling query cache) are handled in the
test class specific fixtures.


ajcvickers:
:shipit:

PR #4106, opened 2015-12-16
Fix #4018 - Substring start index off by one in generated SQL

### LINKED ISSUE #4018: SubString startIndex is one character off in the generated SQL
Using 7.0.0-rc1-final

I have an `NVARCHAR(50)` column named `Code` and I'm trying to return the first character of that column.

The C# projection uses `p => p.Code.Substring(0, 1)` (In C#, `SubString` is 0 based so it made sense to me to pass 0 as the startIndex)

The generate SQL was `SUBSTRING([m].[Code], 0, 1)` which returns an empty string instead of returning the first character. 

When playing with the generated SQL query and changing it to `SUBSTRING([m].[Code], 1, 1)`, it behaves as expected, returning the first character.

Note: The current workaround is to use `p => p.Code.Substring(1, 1)` with a comment explaining why.

mikary:


lajones:
Maybe add some tests for the single argument overload of Substring() and maybe a comment saying that SQL Server is 1-based whereas C# is 0-based just to explain why you're doing what you're doing. But it all looks good :shipit:

src/EntityFramework.MicrosoftSqlServer/Query/ExpressionTranslators/Internal/SqlServerStringSubstringTranslator.cs
```diff
@@ -19,7 +19,15 @@ public virtual Expression Translate(MethodCallExpression methodCallExpression)
...
-                    new[] { methodCallExpression.Object }.Concat(methodCallExpression.Arguments))
+                    new[] {
+                        methodCallExpression.Object,
+                        methodCallExpression.Arguments[0].NodeType == ExpressionType.Constant
+                            ? (Expression)Expression.Constant(
+                                (int)((ConstantExpression) methodCallExpression.Arguments[0]).Value + 1)
+                            : Expression.Add(
+                                methodCallExpression.Arguments[0],
+                                Expression.Constant(1)),
+                        methodCallExpression.Arguments[1] })
```
divega: Do we have the same issue in SQLite?

divega: I can answer the question myself. We don't have the same problem because we don't have a SQLite version of this method translator... In fact it seems there are several method translators we have only implemented for SQL Server but we could implement for SQLite as well.

mikary:
@lajones At the moment, `Substring` is only translated for the two argument overload so the call is going to be forced into client eval. I considered adding tests, but it doesn't seem like an explicitly supported code path.

src/EntityFramework.MicrosoftSqlServer/Query/ExpressionTranslators/Internal/SqlServerStringSubstringTranslator.cs
mikary: We don't have a `IMethodCallTranslator` implementation for `Substring` on SQLite so all of the calls are going through client eval. SQLite does have a `substr(X,Y,Z)` function which uses 1-based counting for string indexes if we want to add a translator for it.

divega: Filed #4108 for triage. We shouldn't block this PR on the outcome.

divega:
@mikary @lajones adding support for the single argument overload would be good, but fixing this existing one to return correct results was higher priority. Feel free to file a bug, but I am sure there are many more methods that are missing and could be supported.

src/EntityFramework.MicrosoftSqlServer/Query/ExpressionTranslators/Internal/SqlServerStringSubstringTranslator.cs
```diff
@@ -19,7 +19,16 @@ public virtual Expression Translate(MethodCallExpression methodCallExpression)
                 ? new SqlFunctionExpression(
                     "SUBSTRING",
                     methodCallExpression.Type,
-                    new[] { methodCallExpression.Object }.Concat(methodCallExpression.Arguments))
+                    new[] {
+                        methodCallExpression.Object,
+                        // Accomidate for SQL Server assumption of 1-based string indexes
```
ErikEJ: Typo => Accomodate

PR #4107, opened 2015-12-16
Stop using IKeyValue for composite keys

ajcvickers:
Instead using object[] directly in identity maps, with appropriate equality comparer. No major gains from this, but it will allow us to remove IKeyValue which ultimately will be deleted.

Also, using custom compares for simple keys for perf and to reduce boxing.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/CompositePrincipalKeyValueFactory.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/CompositePrincipalKeyValueFactory.cs
@@ -1,37 +1,152 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using JetBrains.Annotations;
-using Microsoft.Data.Entity.Infrastructure;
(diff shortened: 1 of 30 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/CompositePrincipalKeyValueFactory.cs
```diff
@@ -1,37 +1,152 @@
...
+        private object[] CreateFromEntry(
+            InternalEntityEntry entry,
+            Func<InternalEntityEntry, IProperty, object> getValue)
+        {
+            var values = new object[_properties.Count];
+            var index = 0;
+
+            foreach (var property in _properties)
+            {
+                values[index++] = getValue(entry, property);
```
lajones: Should this be `if ((values[index++] = getValue(entry, property)) == null) { return null; }` as in `CompositeDependentValueFactory` above or are you ok if some of the objects passed back are null?

ajcvickers: The principal key will not be null because we don't allow null principal keys.

lajones: OK. If a user can't do that (even by mistake) then :shipit:

ajcvickers:
Merged

PR #4111, opened 2015-12-17
Benchmarks for raw SQL queries

### LINKED ISSUE #3351: Perf coverage for FromSql()
We should add microbenchmarks for FromSql(). A few aspects that come to mind that might be important to understand:

- Comparison with SqlQuery() in EF6
- Simple `FromSql()` queries vs. `FromSql()` plus LINQ composition
- Composable (i.e. `SELECT`) vs. non-composable queries
- Caching (see #2315 as an issue that can possibly affect it)

rowanmiller:
Resolves #3351


ajcvickers:
:shipit:

PR #4122, opened 2015-12-17
Push work of getting dependents into the state manager identity map

ajcvickers:
This is faster and also means we no longer user IKeyValue here.


lajones:
Looks good. :shipit:

ajcvickers:
Merged

PR #4128, opened 2015-12-18
Fix #3962 - Navigation with OfType results in null parameter

### LINKED ISSUE #3962: Navigaton property not work whe used OfType
This code raise error
```C#
public class NavPropertyError
    {
        class TestEntity
        {
            public int Id { get; set; }
            public string Field1 { get; set; }
        }

        class TestEntityDetail
        {
            public int Id { get; set; }
            public TestEntity TestEntity { get; set; }
        }

        class Context: DbContext
        {
            public DbSet<TestEntity> Entities { get; set; }
            public DbSet<TestEntityDetail> EntityDetails { get; set; }
        }

        [Fact]
        public void MyTestMethod()
        {
            serviceCollection
                .AddEntityFramework()
                .AddInMemoryDatabase()
                .AddDbContext<Context>(o => o.UseInMemoryDatabase());

            var serviceProvider = serviceCollection.BuildServiceProvider();
            var context = serviceProvider.GetRequiredService<Context>();
            var target = context.EntityDetails;

            target
                .OfType<TestEntityDetail>()
                .Select(ed => ed.TestEntity.Field1)
                .ToArray();
        }
}
```
If remove call OfType, then no error.

mikary:
Resolving #3962 by updating the `IAsyncQueryProvider` on all parent visitors when it is found. The `IQueryable` can be at an arbitrary depth depending on the shape of the original `Expression`.

A potential alternative would be to use a ThreadStatic field, though this is only ever done in `DbContextActivator` in the current code base.

### DIFF
--- a/src/EntityFramework.Core/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/EntityFramework.Core/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -146,6 +146,13 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors.Internal
             {
                 _entityQueryProvider
                     = (node.Value as IQueryable)?.Provider as IAsyncQueryProvider;
+
+                var parent = _parentvisitor;
+                while (parent != null)
+                {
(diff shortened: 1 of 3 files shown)


ajcvickers:
:shipit:

PR #4132, opened 2015-12-18
Split Add/Attach/Update into pairs of methods

ajcvickers:
We decided in the design meeting to use AddSingle and AddWithChildren. But for the Range method the Single seemed weird, so talked with Rowan and we are doing Add/AddWithChildren and AddRange/AddRangeWithChildren.


ajcvickers:
Note: Diego and Rowan probably need to review this so people can hold off reviewing for now.

julielerman:
since nothing in wiki design meeting notes yet, is this instead of graphbehavior parameter? Does Add just simply disappear? Does WithChildren still mean dependents only? Not principals? Thanks. (Thanks @ErikEJ  for calling my attention to this one :) )

rowanmiller:
@julielerman this change is designed to make it more obvious what is going on (and that Add doesn't behave like it did in EF6). I'll put up the design meeting notes shortly. 'WithChildren' means the same thing as 'IncludeDependents'.

julielerman:
Thanks. This bit of the APIs is of the most interest to me. I find myself talking about them more than other features *because* of the behavior change when the methods look like they always have. And I spend even more time making sure people grok the WithDependents because it's not just everything that's attached. I had to have one of you hit me over the head before I realized that even though the enum name can't be much more explicit, I hadn't stopped to process what it meant. Now I have a bunch of relevant tests in my samples (https://github.com/julielerman/EF7WhoAreYou) which I have updated for RC1. No matter how much we say "this is not EF6, don't expect things to act the same", people will anyway. So I think this is a smart move.

ajcvickers:
New pull request submitted.

PR #4135, opened 2015-12-18
Remove many-to-many relationship from database model

### LINKED ISSUE #3996: RevEng: Consider getting rid of the many-to-many associations in relational design metadata
Looking at https://github.com/aspnet/EntityFramework/issues/3945 and discussing with @natemcmaster  I realized we have a few collections representing many-to-many associations in the relational metadata model, e.g. IndexModel has Columns and ForeginKeyModel has both Columns and PrincipalColumns of type `IList<ColumnModel>`. In turn, each ColumnModel can participate of multiple indexes or FKs.

Note that the ordinal position of each column in the index or FK already has a representation in the property since the type is `IList<T>` which is a an ordered collection type (although this may make it hard to leverage EF to materialize this model since EF does not support ordered collections)

For IndexModel.Columns we know we are missing payload (e.g. direction) and it is easy to imagine providers adding their own payload. Unless we have a "join entity" that represents the participation of a column in an index, there wouldn't be a clear place to add new properties or annotations. 

For ForeginKeyModel the matching of a component of the FK and the corresponding component of the PK is relying on the order between Columns and PrincipalColumns being consistent. In my opinion a more natural way to represent that would be to have an list of foreign key components each of which references the dependent column and the primary column it needs to match.

natemcmaster:
Removes implicit column ordering in index and foreign models e.g. ~~`IList`~~ => `ICollection`

Fix #3996

cc @divega @lajones

### DIFF
--- a/src/EntityFramework.Relational.Design/Metadata/ForeignKeyModel.cs
+++ b/src/EntityFramework.Relational.Design/Metadata/ForeignKeyModel.cs
@@ -17,8 +17,7 @@ namespace Microsoft.Data.Entity.Scaffolding.Metadata
         [CanBeNull]
         public virtual TableModel PrincipalTable { get; [param: CanBeNull] set; }
 
-        public virtual IList<ColumnModel> Columns { get; } = new List<ColumnModel>();
-        public virtual IList<ColumnModel> PrincipalColumns { get; } = new List<ColumnModel>();
+        public virtual ICollection<ForeignKeyColumnModel> Columns { get; } = new List<ForeignKeyColumnModel>();
 
         [NotNull]
         public virtual string Name { get; [param: CanBeNull] set; }
@@ -29,6 +28,6 @@ namespace Microsoft.Data.Entity.Scaffolding.Metadata
         //public virtual ReferentialAction OnUpdate { get; [param: NotNull] set; }
 
         public virtual string DisplayName
-            => Table?.DisplayName + "(" + string.Join(",", Columns.Select(f => f.Name)) + ")";
+            => Table?.DisplayName + "(" + string.Join(",", Columns.OrderBy(f => f.Ordinal).Select(f => f.Column.Name)) + ")";
     }
 }
--- a/src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
+++ b/src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Data.Entity.Scaffolding
             return modelBuilder;
         }
 
-        protected virtual ModelBuilder VisitSequences([NotNull] ModelBuilder modelBuilder, [NotNull] IList<SequenceModel> sequences)
+        protected virtual ModelBuilder VisitSequences([NotNull] ModelBuilder modelBuilder, [NotNull] ICollection<SequenceModel> sequences)
         {
             Check.NotNull(modelBuilder, nameof(modelBuilder));
             Check.NotNull(sequences, nameof(sequences));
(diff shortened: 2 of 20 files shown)


test/EntityFramework.Sqlite.Design.FunctionalTests/SqliteScaffoldingModelFactoryTest.cs
```diff
@@ -0,0 +1,290 @@
...
+);";
+            var table = GetModel(sql).FindEntityType("Friends");
+
+            var fk = table.FindForeignKeys(new[] { table.FindProperty("Id") }).SingleOrDefault();
+
+            Assert.True(fk.IsUnique);
+        }
+
+        [Fact]
+        public void It_does_not_assigns_uniqueness_to_foreign_key()
```
lajones: Nit: typo "assigns" -> "assign"

test/EntityFramework.Sqlite.Design.FunctionalTests/SqliteScaffoldingModelFactoryTest.cs
```diff
@@ -0,0 +1,290 @@
...
+            Assert.NotEmpty(parent.GetReferencingForeignKeys());
+            Assert.NotEmpty(children.GetForeignKeys());
+
+            var principalKey = children.FindForeignKeys(children.FindProperty("ParentId")).SingleOrDefault().PrincipalKey;
+            Assert.Equal("Parent", principalKey.DeclaringEntityType.Name);
+            Assert.Equal("Id", principalKey.Properties[0].Name);
+        }
+
+        [Fact]
+        public void It_loads_multiple_key_references()
```
lajones: Nit: This looks more like `It_loads_composite_key_references`? When it said multiple I was expecting multiple foreign keys in one table - which might also be interesting...

test/EntityFramework.Relational.Design.Tests/RelationalDatabaseModelFactoryTest.cs
```diff
@@ -241,10 +241,37 @@ public void Indexes_and_alternate_keys()
...
+            {
+                Name = "IDX_C2_C1",
+                IndexColumns = {
+                    new IndexColumnModel { Column = table.Columns.ElementAt(1) },
+                    new IndexColumnModel { Column = table.Columns.ElementAt(0) }
+                },
+                IsUnique = false
+            });
+            table.Indexes.Add(new IndexModel
+            { /*Name ="UNQ_C3_C1",*/
```
lajones: Very small nit: maybe put comment on next line.

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -344,7 +344,10 @@ protected virtual IndexBuilder VisitIndex([NotNull] EntityTypeBuilder builder, [
             Check.NotNull(builder, nameof(builder));
             Check.NotNull(index, nameof(index));
 
-            var propertyNames = index.Columns.Select(GetPropertyName).ToArray();
+            var propertyNames = index.IndexColumns
+                .OrderBy(ic=>ic.IndexOrdinal)
```
lajones: Nit: spaces around `=>`

src/EntityFramework.Relational.Design/Metadata/ForeignKeyModel.cs
```diff
@@ -29,6 +28,6 @@ public class ForeignKeyModel : Annotatable
         //public virtual ReferentialAction OnUpdate { get; [param: NotNull] set; }
 
         public virtual string DisplayName
-            => Table?.DisplayName + "(" + string.Join(",", Columns.Select(f => f.Name)) + ")";
+            => Table?.DisplayName + "(" + string.Join(",", Columns.Select(f => f.Column.Name)) + ")";
```
lajones: Nit: depending on where this is used, may be better to sort the columns by ordinal before combining them in `string.Join()`

lajones:
Overall - looks good. Few small comments but nothing major :shipit:

natemcmaster:
Updated w/ changes from comments.

PR #4136, opened 2015-12-18
Improve perf of Add/Attach/Update by caching key metadata

ajcvickers:
Properties keep track of which keys they belong to.

### DIFF
--- a/src/EntityFramework.Core/Metadata/Internal/IPropertyKeyMetadata.cs
+++ b/src/EntityFramework.Core/Metadata/Internal/IPropertyKeyMetadata.cs
@@ -1,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Data.Entity.Metadata.Internal
+{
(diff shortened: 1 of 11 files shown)


ajcvickers:
Note: Diego and Rowan probably need to review this so people can hold off reviewing for now.

src/EntityFramework.Core/Metadata/Internal/IPropertyKeyMetadata.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Metadata.Internal
+{
+    public interface IPropertyKeyMetadata
+    {
+        IKey PrimaryKey { get; [param: CanBeNull] set; }
+        IReadOnlyList<IKey> Keys { get; [param: CanBeNull] set; }
+        IReadOnlyList<IForeignKey> ForeignKeys { get; [param: CanBeNull] set; }
```
AndriySvyryd: Why are the setters part of the interface?

ajcvickers: Left over from previous slightly different implementation. I will remove them.

AndriySvyryd:
:shipit:

PR #4137, opened 2015-12-18
Fix #2227 - Better exception message for no database provider configured

### LINKED ISSUE #2227: Improve exception messages for "no data store configured"
We have seen people hit this exception several times in the betas: #2191 #2156 #2094.

The current exception message (_No data stores are configured. Configure a data store by overriding OnConfiguring in your DbContext class or in the AddDbContext method when setting up services._) does not account for one of the common causes: customers will often have their derived DbContext correctly setup in their service configuration but will then bypass DI in the creation of the context by creating the context with the new keyword.

I believe we should try to say something more helpful in the exception message, e.g. something like this:

_No data stores are configured. You can configure a data store by calling the AddDbContext method when setting up services and then instantiating the derived DbContext class using dependency injection, or by overriding OnConfiguring in your DbContext class._

Note that the fact it is at all possible to new a DbContext without passing any parameters to the constructor is an unfortunate effect of the decision we made to enable default constructors to work with ASP.NET and to have a DbContext with only a default constructor in the default application templates. We could also revert that decision although it has some unwanted consequences.

mikary:
Updated based on wording suggested in issue #2227.
@divega @rowanmiller

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.Designer.cs
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.Designer.cs
@@ -197,7 +197,7 @@ namespace Microsoft.EntityFrameworkCore.Internal
         }
 
         /// <summary>
-        /// No database providers are configured. Configure a database provider by overriding OnConfiguring in your DbContext class or in the AddDbContext method when setting up services.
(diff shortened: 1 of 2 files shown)


rowanmiller:
I'm going to propose the following. @divega what do you think?

> A database has not been configured for this context. If you are not using dependency injection, override the OnConfiguring method in your context class to configure the database. If you are using dependency injection, use the AddDbContext method to register your context as a service and configure the database. In ASP.NET 5 this is done in the ConfigureServices method of your Startup class. Also ensure that you are resolving the context from services and are not directly creating instances of it.

divega:
Replace ASP.NET 5 with ASP.NET Core. Otherwise looks good to me :shipit:

PR #4138, opened 2015-12-18
Stop using Enum.IsDefined

ajcvickers:
It was taking 25% of the time in the Update perf test.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/EntityEntry.cs
+++ b/src/EntityFramework.Core/ChangeTracking/EntityEntry.cs
@@ -1,10 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Diagnostics;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.ChangeTracking.Internal;
 using Microsoft.Data.Entity.Infrastructure;
+using Microsoft.Data.Entity.Internal;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Utilities;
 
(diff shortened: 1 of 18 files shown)


src/EntityFramework.Core/ChangeTracking/EntityEntry.cs
```diff
@@ -58,7 +60,11 @@ public virtual EntityState State
             get { return _internalEntityEntry.EntityState; }
             set
             {
-                Check.IsDefined(value, nameof(value));
+                if (value < 0
+                    || (int)value > 4)
```
AndriySvyryd: Why magic numbers?

src/EntityFramework.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
```diff
@@ -52,7 +53,12 @@ public virtual TBuilder UseRelationalNulls()
 
         public virtual TBuilder QueryClientEvaluationBehavior(QueryClientEvaluationBehavior queryClientEvaluationBehavior)
         {
-            Check.IsDefined(queryClientEvaluationBehavior, nameof(queryClientEvaluationBehavior));
+            if (queryClientEvaluationBehavior < 0
+                || (int)queryClientEvaluationBehavior > 2)
```
AndriySvyryd: Magic numbers

src/EntityFramework.Core/ChangeTracking/EntityEntry.cs
ajcvickers: Would you prefer named constants? Or using values from the enum itself? Or explicitly listing each value? None of these options will prevent this code from breaking if a new enum value is added, so I went for the simplest thing. Other ideas welcome, as long as the code is fast.

AndriySvyryd: Put the result of `Enum.GetValues(typeof(EntityState)).Cast<int>().Max()` in a static readonly field

ajcvickers: Will do.

AndriySvyryd:
:shipit:

ajcvickers:
Merged

PR #4139, opened 2015-12-18
Cache services in DbContext

ajcvickers:
Avoid requesting them from DI multiple times. (Even if our use of DI gets faster this is on a hot path where not using DI at all will matter.)


AndriySvyryd:
:shipit:

ajcvickers:
Merged

PR #4141, opened 2015-12-18
Boolean in select

### LINKED ISSUE #649: Query: support conditional expressions

### LINKED ISSUE #3618: Invalid SQL when projection includes a comparison to a Nullable type
int? selectedId = null;

    var results = await DbContext.Departments
        .Select(x => new {Selected = x.Id == selectedId})
        .ToListAsync();

generates this SQL:

    exec sp_executesql N'SELECT [x].[Id] = @__selectedId_0
    FROM [Department] AS [x]',N'@__selectedId_0 int',@__selectedId_0=NULL

This looks to be similar to https://github.com/aspnet/EntityFramework/issues/2450 (apologies if it is a duplicate).

### LINKED ISSUE #3703: Incorrect syntax generated for select new { bool }
Beta8 / SQL Server 2014
Database model:

```
public class Blog
{
        public bool IsHipster { get; set; } // <=
        public List<Post> Posts { get; set; }
}
```

Repository model:

```
public class RepositoryBlog
{
        public bool IsHippy { get; set; } // <=
        public List<Post> Posts { get; set; }
}
```

Query 1:

```
List<RepositoryBlog> x = (from b in db.Blogs
	select new RepositoryBlog
	{
		BlogId = b.BlogId,
		IsHippy = !b.IsHipster, // Note negation
		Url = b.Url
	}).ToList();
```

Generated SQL:

```
SELECT [b].[BlogId], [b].[IsHipster] = 0, [b].[Url]
FROM [Blog] AS [b]
```

SQL Server error:

```
Incorrect syntax near '='.
```

Query 2:

```
List<RepositoryBlog> x = (from b in db.Blogs
  select new RepositoryBlog
  {
	  BlogId = b.BlogId,
	  IsHippy = b.IsHipster ? false : true, // Attempted workaround
	  Url = b.Url
  }).ToList();
```

Generated SQL:

```
SELECT [b].[BlogId], CASE
    WHEN [b].[IsHipster]
    THEN CAST(0 AS BIT) ELSE CAST(1 AS BIT)
END, [b].[Url]
FROM [Blog] AS [b]
```

SQL Server error:

```
An expression of non-boolean type specified in a context where a condition is expected, near 'THEN'.
```

Twice the fun!

smitpatel:
resolves #3703 
Issue: SQL does not allow not operation on Boolean type directly. They must be compared with true or false. Fix is to convert unary operations on Boolean type to comparison operation.
resolves  #3618 
Issue: SQL does not allow comparison operation into projection. They must be wrapped in `Case When Then` block.
Fix: Whenever there is comparison operation into projection, convert it to conditional expression.
If comparison operation is inside conditional expression already then do no convert.
Issue: Comparison follows null semantics.
Implementation: While processing `SelectExpression` process the projection accordingly.
resolves #649 for projections

### DIFF
--- a/src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -143,7 +143,7 @@ namespace Microsoft.Data.Entity.Query.Sql
                     _relationalCommandBuilder.Append(", ");
                 }
 
-                VisitJoin(selectExpression.Projection);
+                VisitProjection(selectExpression.Projection);
 
                 projectionAdded = true;
             }
@@ -225,6 +225,13 @@ namespace Microsoft.Data.Entity.Query.Sql
             }
 
             return selectExpression;
+        }
+
+        protected virtual void VisitProjection([NotNull] IReadOnlyList<Expression> projections)
+        {
+            var nullComparisonTransformer = new NullComparisonTransformingVisitor(_parametersValues);
+
+            VisitJoin(projections.Select(e => nullComparisonTransformer.Visit(e)).ToList());
         }
 
         protected virtual void GenerateOrderBy([NotNull] IReadOnlyList<Ordering> orderings)
--- a/src/EntityFramework.MicrosoftSqlServer/Query/Sql/Internal/SqlServerQuerySqlGenerator.cs
+++ b/src/EntityFramework.MicrosoftSqlServer/Query/Sql/Internal/SqlServerQuerySqlGenerator.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
@@ -8,6 +9,7 @@ using Microsoft.Data.Entity.Query.Expressions;
 using Microsoft.Data.Entity.Query.Expressions.Internal;
 using Microsoft.Data.Entity.Storage;
 using Microsoft.Data.Entity.Utilities;
+using Remotion.Linq.Parsing;
 
 namespace Microsoft.Data.Entity.Query.Sql.Internal
 {
@@ -20,10 +22,10 @@ namespace Microsoft.Data.Entity.Query.Sql.Internal
             [NotNull] IRelationalTypeMapper relationalTypeMapper,
             [NotNull] SelectExpression selectExpression)
             : base(
-                  relationalCommandBuilderFactory, 
-                  sqlGenerationHelper, 
-                  parameterNameGeneratorFactory, 
-                  relationalTypeMapper, 
+                  relationalCommandBuilderFactory,
+                  sqlGenerationHelper,
+                  parameterNameGeneratorFactory,
+                  relationalTypeMapper,
                   selectExpression)
         {
         }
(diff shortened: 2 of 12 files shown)


src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -143,7 +143,14 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
                     _relationalCommandBuilder.Append(", ");
                 }
 
-                VisitJoin(selectExpression.Projection);
+                var comparisonTransformer = new ProjectionComparisonTransformingVisitor();
+                var nullComparisonTransformer = new NullComparisonTransformingVisitor(_parametersValues);
```
anpete: DRY this up with predicate case below.

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -143,7 +143,14 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
                     _relationalCommandBuilder.Append(", ");
                 }
 
-                VisitJoin(selectExpression.Projection);
+                var comparisonTransformer = new ProjectionComparisonTransformingVisitor();
```
anpete: One question: Should this handling be in our base SQL gen vs SQL Server? Is it appropriate for the general case?

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -143,7 +143,14 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
...
+                var nullComparisonTransformer = new NullComparisonTransformingVisitor(_parametersValues);
+                var transformedProjections = selectExpression.Projection
+                    .Select(comparisonTransformer.Visit)
+                    .Select(nullComparisonTransformer.Visit)
+                    .ToList();
+
```
anpete: I'm assuming we do this rewrite as part of SQL gen because it can be dependent on parameter values?

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
smitpatel: This works on MySQL, looks like SQL server specific issue.

smitpatel:
Updated
Split converting comparison to case block to SqlServer Specific. Null semantics are required for SQLite too so it is in relational.

anpete:
@smitpatel Does this potentially resolve #649?

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1125,7 +1132,7 @@ protected virtual bool TryGenerateBinaryOperator(ExpressionType op, [NotNull] ou
         protected override Exception CreateUnhandledItemException<T>(T unhandledItem, string visitMethod)
             => new NotImplementedException(visitMethod);
 
-        private class NullComparisonTransformingVisitor : RelinqExpressionVisitor
+        protected class NullComparisonTransformingVisitor : RelinqExpressionVisitor
```
anpete: This need to be protected?

src/EntityFramework.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -143,7 +145,10 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
                     _relationalCommandBuilder.Append(", ");
                 }
 
-                VisitJoin(selectExpression.Projection);
+                var transformedProjection = TransformProjection(selectExpression)
```
anpete: Looks good. I think we could make the pattern more consistent by tweaking it slightly: Introduce a method VisitProjection(IReadOnlyList<Expression>). In the base class it would run the null comparison transformer and call VisitJoin. In the SQL Server override it would run the projection comparison transformer and then call base. Do you think that would work?

smitpatel: Changed it. Looked better pattern as it allows provider to change the way projection is processed.

smitpatel:
#649 is resolved by this at least for projections.

anpete:
:shipit:

PR #4142, opened 2015-12-19
Property flag perf improvements

ajcvickers:
Hot path for Update looks at a lot of flags and doesn't do much else, so making flag lookup faster helps a lot.


AndriySvyryd:
I don't like it, but if it gives significant perf improvement, then :shipit:

ajcvickers:
@AndriySvyryd Do you think we should just dump the flag code and list out all the bools separately? Given the way the metadata model has evolved it doesn't seem like this is worth saving memory like this, but I retained the compact storage mainly because it was already there.

AndriySvyryd:
Yes, I would opt for listing them out. But we can do it later, when we add a readonly implementation.

ajcvickers:
Merged

PR #4151, opened 2015-12-21
Map indexed properties to fit within SQL Server 900-byte limitation

### LINKED ISSUE #1071: nvarchar(max) can't be used with indexed columns
(See [How to create an index for a string property in Entity Framework 7](http://stackoverflow.com/q/26950625/475031))

Similar to how we use `nvarchar(450)` on key columns, we should do something similar for indexed columns. I assume this also applies to `varbinary(max)`.

natemcmaster:
Fix #1071 

Ref: https://msdn.microsoft.com/en-us/library/ms188783.aspx

cc @ajcvickers

### DIFF
--- a/src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
+++ b/src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Linq;
 using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Utilities;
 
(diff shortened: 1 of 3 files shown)


src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
```diff
@@ -144,5 +146,9 @@ protected override RelationalTypeMapping FindCustomMapping(IProperty property)
                         _varbinarymax, _varbinarymax, _varbinary900, _rowversion)
                     : null;
         }
+
+        // indexes in SQL Server have a max size of 900 bytes
+        protected override bool RequiresKeyMapping([NotNull] IProperty property)
```
ajcvickers: Don't put NotNull on overrides. (R# will show this as grayed out "dead code".)

ajcvickers:
:shipit:

PR #4156, opened 2015-12-22
Create cache for dependent entries when requested first time

ajcvickers:
This greatly improves perf for querying dependents first, followed by principals.

### DIFF
--- a/src/EntityFramework.Core/ChangeTracking/Internal/DependentsMap.cs
+++ b/src/EntityFramework.Core/ChangeTracking/Internal/DependentsMap.cs
@@ -1,0 +1,99 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.Data.Entity.Metadata;
+
+namespace Microsoft.Data.Entity.ChangeTracking.Internal
+{
+    public class DependentsMap<TKey> : IDependentsMap
+    {
+        private readonly IEntityType _entityType;
+        private readonly IPrincipalKeyValueFactory<TKey> _principalKeyValueFactory;
+        private readonly IDependentKeyValueFactory<TKey> _dependentKeyValueFactory;
+        private readonly Dictionary<TKey, List<InternalEntityEntry>> _map;
+
+        public DependentsMap(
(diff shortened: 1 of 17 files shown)


src/EntityFramework.Core/ChangeTracking/Internal/DependentsMap.cs
```diff
@@ -0,0 +1,99 @@
...
+            _entityType = foreignKey.DeclaringEntityType;
+            _principalKeyValueFactory = principalKeyValueFactory;
+            _dependentKeyValueFactory = dependentKeyValueFactory;
+            _map = new Dictionary<TKey, List<InternalEntityEntry>>(principalKeyValueFactory.EqualityComparer);
+        }
+
+        public virtual void Add(InternalEntityEntry entry)
+        {
+            TKey key;
+            if (_entityType.IsAssignableFrom(entry.EntityType)
```
AndriySvyryd: Should we throw or Assert on `else` ?

src/EntityFramework.Core/Metadata/Internal/ForeignKey.cs
```diff
@@ -465,5 +467,8 @@ private static bool ArePropertyTypesCompatible(IReadOnlyList<IProperty> principa
 
         // Note: This is set and used only by IdentityMapFactoryFactory, which ensures thread-safety
         public virtual object DependentKeyValueFactory { get; set; }
+
+        // Note: This is set and used only by IdentityMapFactoryFactory, which ensures thread-safety
+        public virtual Func<IDependentsMap> DependentsMapFactory { get; set; }
```
AndriySvyryd: Implement the interface explicitly?

src/EntityFramework.Core/ChangeTracking/Internal/DependentsMap.cs
ajcvickers: Good point. Let me look to see if the check can be made a debug assert.

src/EntityFramework.Core/Metadata/Internal/ForeignKey.cs
ajcvickers: Why? It's on internal code already.

AndriySvyryd: To make it even less likely to be used. This is optional, as it doesn't matter much.

AndriySvyryd:
:shipit:

ajcvickers:
Merged

PR #4157, opened 2015-12-22
Switch off auto DetectChanges for EF6 fixup tests

ajcvickers:
Because it doesn't happen in EF7 and dominates the time in the EF6 runs.


rowanmiller:
Rather than doing this, I think we should have a 'Default' and 'AutoDetectChangesOff' variation. The comparison we have is good in that by default you just get better performance now... but I agree the comparison without detect changes is more interesting for tuning our code base.

ajcvickers:
@rowanmiller I'm fine with that. Do you want to do it? :-)

rowanmiller:
Sure, I'll do it now

ajcvickers:
Superseded by Rowan's pull request.

Issue #4095, opened 2015-12-15
Duplicate variable names when batching inserts with only default values


mikary:
When attempting to add multiple instances with default values to the context, a `System.Data.SqlClient.SqlException` is thrown on `SaveChanges` with the message "The variable name '@generated5' has already been declared. Variable names must be unique within a query batch or stored procedure."

**Reproduction**
```cs
    public class Program
    {
        public void Main(string[] args)
        {
            using (var Sql = new TestDataModel())
            {
                Sql.Database.EnsureDeleted();
                Sql.Database.EnsureCreated();

                var pictures = new List<Picture>();

                for (var index = 0; index < 15; index++)
                {
                    var picture = new Picture();
                    Sql.Pictures.Add(picture);
                    pictures.Add(picture);
                }

                for (var index = 0; index < 5; index++)
                {
                    var package = new PicturePackage
                    {
                        Pictures = new List<Picture>
                        {
                            pictures[3 * index],
                            pictures[3 * index + 1],
                            pictures[3 * index + 2]
                        }
                    };

                    Sql.PicturePackages.Add(package);
                }

                Sql.SaveChanges();
            }
        }
    }

    // Sample database with object relations.
    public class TestDataModel : DbContext
    {
        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(
                new SqlConnectionStringBuilder
                {
                    DataSource = @"(localdb)\MSSQLLocalDB",
                    MultipleActiveResultSets = true,
                    InitialCatalog = "BatchingDefaults",
                    IntegratedSecurity = true,
                    ConnectTimeout = 30
                }.ConnectionString);
        }

        public virtual DbSet<Picture> Pictures { get; set; }
        public virtual DbSet<PicturePackage> PicturePackages { get; set; }
    }

    public class Picture
    {
        public int PackageId { get; set; }

        [Required]
        [ForeignKey(nameof(PackageId))]
        public PicturePackage Package { get; set; }

        [Key]
        public int Id { get; set; }

    }

    public class PicturePackage
    {
        [InverseProperty(nameof(Picture.Package))]
        public virtual ICollection<Picture> Pictures { get; set; }

        [Key]
        public int Id { get; set; }
    }
```

**Generated SQL**
```sql
SET NOCOUNT ON;
DECLARE @generated5 TABLE ([Id] int);
INSERT INTO [PicturePackage]
[...]

PR #4171, opened 2015-12-24
Add a strong order to the generated values returned from batches of inserts for Sql Server

### LINKED ISSUE #4080: Inserted entities has different ID (DB vs instance) after SaveChanges
Hi guys...
I have problem with inserting more entities (same class).
My environment: MSSQL 2014 (this trouble may appears on MSSQL 2012+).

Situation:
Create 500 instances of one type entity a call SaveChanges.
You generate one insert with many values and identity insert into temporary table...but:
"There is no guarantee that the order in which the changes are applied to the table and the order in which the rows are inserted into the output table or table variable will correspond." (source: https://msdn.microsoft.com/en-us/library/ms177564.aspx or http://stackoverflow.com/questions/11222043/table-valued-function-order-by-is-ignored-in-output).

I have table with identity (no big deal...default identity(1, 1)) and DbReader output is:
```
--ID int (PK, identity(1, 1)):
1. 508
2. 466
3. 462
4. 493
5. 504
6. 520
7. 496
8. 497
9. 494
10. 277
11. 281
12. 263
13. 476
...
```
**This is fatal when you need continue work with instances :(**

Thx for fix,
Milan

AndriySvyryd:
Fixes #4080, #4095

### DIFF
--- a/src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
+++ b/src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using JetBrains.Annotations;
@@ -15,6 +16,8 @@ namespace Microsoft.Data.Entity.Update.Internal
     public class SqlServerUpdateSqlGenerator : UpdateSqlGenerator, ISqlServerUpdateSqlGenerator
     {
         private readonly IRelationalTypeMapper _typeMapper;
+        private readonly string _tableId = "_t";
+        private readonly string _joinTableId = "_j";
 
         public SqlServerUpdateSqlGenerator([NotNull] ISqlGenerationHelper sqlGenerationHelper,
             [NotNull] IRelationalTypeMapper typeMapper)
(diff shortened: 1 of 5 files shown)


src/EntityFramework.MicrosoftSqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
```diff
@@ -154,28 +293,112 @@ private string GetTypeNameForCopy(IProperty property)
...
+            var rowNumberColumnName2 = SqlGenerationHelper.DelimitIdentifier("_RowNumber2");
+            if (keyOperations != null)
+            {
+                var keyColumns = string.Join(", ", keyOperations.Select(c => SqlGenerationHelper.DelimitIdentifier(c.ColumnName, _tableId)));
+                commandStringBuilder
+                    .Append("FROM (SELECT ")
+                    .AppendJoin(readOperations.Select(c => SqlGenerationHelper.DelimitIdentifier(c.ColumnName)))
+                    .Append(", " + SqlGenerationHelper.DelimitIdentifier(positionColumnName) + ",")
+                    .AppendLine()
+                    .AppendFormat("ROW_NUMBER() OVER (PARTITION BY {0} ORDER BY {1}) AS {2},",
```
divega: Is there any coverage for the cases in which this SQL gets generated? (Couldn't find it in this PR, but I was using a mobile browser)

AndriySvyryd: Yes, it's covered by the functional test, but it don't assert the actual SQL since it's rather esoteric.

divega: Got it. Thanks!

ajcvickers:
:sheep::it:

Issue #4168, opened 2015-12-23
Scaffold a subset of tables - silently ignores / skips tables that cannot be found


dazinator:
I have an existing database that I'd like to scaffold a `DbContext` from, but only for a subset of the tables in this database.

I tried: 

`dnx ef dbcontext scaffold "connstring" EntityFramework.MicrosoftSqlServer --outputDir Models --table dbo.table1 --table dbo.table2`

UPDATE:  ~~This doesn't work - it seemed to only scaffold the second table.~~ 

This does work, but the table names are case sensitive, and if you get the casing wrong, then it seems to silently skip the scaffolding of that table.

I think it should output a warning to say that a table was skipped for some reason.

ErikEJ:
Have you tried "-t" ie no double hyphens?

dazinator:
Oh yikes - tried using `-t dbo.table1 -t dbo.table2` and this time it worked.. which led me to re-look at the issue. `--table dbo.table1 --table dbo.table2` also worked! So how come it didn't work when I first tried? Well.. the table names are case sensitive, and if you get the casing wrong it appears to be silently ignored by the command line making me think that it was only ever doing one table..

I think if it doesn't like a table it should probably output a warning of some kind :)

Thanks

ErikEJ:
I think it should not be case sensitive at all!

rowanmiller:
> I think it should not be case sensitive at all

Agreed, as I don't think any of the major relational databases let table names only differ by case.

> I think if it doesn't like a table it should probably output a warning of some kind to say that it skipped scaffolding for a table because it couldn't find it

Also agreed, since you explicitly asked for that table to be included

rowanmiller:
@lajones we should warn if the table is not found. Case sensitivity should be up to the provider (and ours should not be case sensitive).

lajones:
Note: case-sensitivity was fixed by @ErikEJ with bfc8da1 (Thanks Erik).

lajones:
Note: each provider interprets the `--schema` and `--table` selections according to its own rules (e.g. the SQL Server provider allows you to pass `[schema].[table]` as well as `schema.table`) - and so will also need to be responsible for the functionality that interprets whether those selections have been satisfied.

lajones:
Fix checked in with PR#4252. Commit 50c51e7.

sreejithvijayan:
If anyone got the below error, 

[...]

PR #4172, opened 2015-12-24
RevEng: Fix for issue #4168 (part 1)

ErikEJ:
- Make TableSelectionSetExtension implementation case insensitive
- Fix R# warnings
- Update test to work with case-insensitive table name


natemcmaster:
Looks good.

Issue #3670, opened 2015-11-05
Logging: Binary parameter values logged as System.Byte[]


ErikEJ:
Any binary parameter values are logged as System.Byte[], see for example TimestampAttribute_throws_if_value_in_database_changed()

I would be valuable if the actual values were logged as a string with:

     "0x" + 
     for (int i = 0; i < buffer.Length; i++)
     {
         _sbScript.Append(buffer[i].ToString("X2", System.Globalization.CultureInfo.InvariantCulture));
      }

But maybe impose a limit on the size??

PR #4184, opened 2015-12-28
Parameter logging now logs the value of byte arrays

ErikEJ:
Fixes issue #3670

### DIFF
--- a/src/EntityFramework.Relational/Extensions/DbCommandLogData.cs
+++ b/src/EntityFramework.Relational/Extensions/DbCommandLogData.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Data.Entity.Storage
             [NotNull] string commandText,
             CommandType commandType,
             int commandTimeout,
-            [NotNull] IReadOnlyDictionary<string, string> parameters,
+            [NotNull] IReadOnlyDictionary<string, object> parameters,
             long? elapsedMilliseconds)
         {
             CommandText = commandText;
@@ -28,7 +28,7 @@ namespace Microsoft.Data.Entity.Storage
         public virtual string CommandText { get; }
         public virtual CommandType CommandType { get; }
         public virtual int CommandTimeout { get; }
-        public virtual IReadOnlyDictionary<string, string> Parameters { get; }
+        public virtual IReadOnlyDictionary<string, object> Parameters { get; }
         public virtual long? ElapsedMilliseconds { get; }
     }
 }
--- a/src/EntityFramework.Relational/Extensions/DbCommandLogData.cs
+++ b/src/EntityFramework.Relational/Extensions/DbCommandLogData.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Data.Entity.Storage
             [NotNull] string commandText,
             CommandType commandType,
             int commandTimeout,
-            [NotNull] IReadOnlyDictionary<string, object> parameters,
+            [NotNull] IReadOnlyDictionary<string, string> parameters,
             long? elapsedMilliseconds)
         {
             CommandText = commandText;
@@ -28,7 +28,7 @@ namespace Microsoft.Data.Entity.Storage
         public virtual string CommandText { get; }
         public virtual CommandType CommandType { get; }
         public virtual int CommandTimeout { get; }
-        public virtual IReadOnlyDictionary<string, object> Parameters { get; }
+        public virtual IReadOnlyDictionary<string, string> Parameters { get; }
         public virtual long? ElapsedMilliseconds { get; }
     }
 }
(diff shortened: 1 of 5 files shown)


test/EntityFramework.MicrosoftSqlServer.FunctionalTests/DataAnnotationSqlServerTest.cs
```diff
@@ -178,7 +178,7 @@ public override void TimestampAttribute_throws_if_value_in_database_changed()
 
 @p0: 1
 @p1: ModifiedData
-@p2: System.Byte[]
+@p2: 0x00000000000007D1
```
ajcvickers: This is going to get very long for things like images. Maybe if the length is greater than 16, then do just the first and last eight bytes--something like "0x0102030405060708...0102030405060708" @divega @rowanmiller Thoughts?

rowanmiller: I'm torn... having the full value is nice... but then again perhaps the full value is of less benefit when it's binary data. The same could be true for string data too, `Post.Content` could be a very sizable string.

ErikEJ: Given that parameter logging is something you opt-in for, I vote for logging the full values

anpete: Agree with @ajcvickers. I think it would be fine to just truncate the data. To @rowanmiller's point, we should consider doing the same thing for strings.

divega: I can see that logging parameters of arbitrary large sizes could cause problems. I can also see that this probably isn't worth adding another knob to :smile: So I guess I would personally truncate at a larger size than 16, so I would still get the full value for more cases. I also probably wouldn't bother logging the tail of the value when I truncate. 

BTW, isn't how parameter values get formatted for the log and truncated potentially a type mapping concern?

ErikEJ: So in conslusion: Log the first 16 bytes.
If more than 16 bytes long, log the first 16 bytes and append ... (3 dots)
0x01020304050607080102030405060708...
Is that how to do it for now?

divega: By "at a size larger size than 16" I was trying to suggest to pick a larger number, e.g. 32, as the limit. I don't feel strongly about any of this though.

ErikEJ: Ah, got it. Let's go for 32 for now, then?

anpete: Still too small? Perhaps 50?

ErikEJ: how about 64 ?

divega: I guess it could help to agree on what the problem is with too long parameters. I believe there are two main ones:
1. Too large log files would turn logging unusable
2. Too many cycles spent formatting would turn logging unusable
3. Too long parameter values could turn the console log unreadable, so picking a lenght that multiplied by two (since byte[] is rendered as hex) plus a typical lenght of a parameter name, colon, space and the ellipsis at the end doesn't make it overflow a line in a typical console. 

To me the 3rd issue makes 32 a reasonable number, but I am not sure this is a strong argument.

ErikEJ: @divega Had not thought about 3 - in that case 32 makes a lot of sense

src/EntityFramework.Relational/Extensions/DbCommandLogData.cs
```diff
@@ -15,7 +15,7 @@ public class DbCommandLogData
             [NotNull] string commandText,
             CommandType commandType,
             int commandTimeout,
-            [NotNull] IReadOnlyDictionary<string, object> parameters,
+            [NotNull] IReadOnlyDictionary<string, string> parameters,
```
anpete: We probably want to keep this strongly typed for structured logging.

anpete:
@ErikEJ Sorry if I was unclear. We want the parameter values in DbCommandLogData to be the original typed values, not strings. The custom parameter formatting should only happen when we are rendering a string via the formatter delegate.

ErikEJ:
@anpete Ah. so you mean do the Parameter Formatting here?

           state.Parameters 
                             .Select(kv => $"{kv.Key}='**{kv.Value}**'") 
                             .Join(),

anpete:
@ErikEJ Yep, we want the original objects to flow through as a structure logging payload.

anpete:
@ErikEJ Are you planning to update this?

ErikEJ:
@anpete Yes, I just got stuck in the dotnet chaos - planning to submit a new PR

ErikEJ:
Closing and will attempt a new PR

PR #4197, opened 2015-12-30
[uwp] Add command to generate directives for runtime types

### LINKED ISSUE #3801: Automate runtime directive generation on UWP
Extend the UWP build process to add project-specific runtime directives. This ideally runs post-CSC pre-ILC. This addresses impossible problem of *statically* generating directives for all possible `Snapshot<T1...T30>` types (minimum 2.35x10^32 are possible).

See https://github.com/aspnet/EntityFramework/pull/3777#issuecomment-157545631.

natemcmaster:
Adds powershell command `Scaffold-Directives`. This will automatically create a file in the UWP project called "EntityFramework.g.rd.xml" this file contains runtime directives for the types user projects will need at runtime.

Removes the chained generic-type inference which was causing the compiler produce heavily bloated binaries or to crash with out-of-memory exceptions.

Fix #3801 

TODO:
- [ ] Improve coverage of generate runtime directives for queries using TransparentIdentifier and anonymous types
- [x] Add analyzer to improve maintainability of runtime-generation ( see https://github.com/aspnet/EntityFramework/pull/4199)
- [ ] Catch metadata missing exception and throw user-friendly error to educate on existence of the new command


natemcmaster:
cc @rowanmiller @bricelam

src/EntityFramework.Commands/Design/DbContextOperations.cs
```diff
@@ -45,8 +45,23 @@ public class DbContextOperations
...
+        {
+            Check.NotNull(contextType, nameof(contextType));
+
+            Func<DbContext> factory;
+            if(!FindContextTypes().TryGetValue(contextType, out factory))
+            {
+                throw new OperationException(CommandsStrings.NoContext);
+            }
+            return CreateContext(factory);
+        }
```
bricelam: To better match the existing code, this should just `return CreateContext(FindContextType(contextType).Value)` and the logic here should be moved to a new `FindContextType(Type)` method.

src/EntityFramework.Commands/Design/Internal/DesignTimeServicesBuilder.cs
```diff
@@ -83,18 +86,25 @@ private void ConfigureProviderServices(string provider, IServiceCollection servi
...
+                    .AddTransient(_ => contextServices.GetService<ILoggerFactory>())
+                    .AddTransient(_ => contextServices.GetService<IMigrationsAssembly>())
+                    .AddTransient(_ => contextServices.GetService<IMigrationsIdGenerator>())
+                    .AddTransient(_ => contextServices.GetService<IMigrationsModelDiffer>())
+                    .AddTransient(_ => contextServices.GetService<IMigrator>())
+                    .AddTransient(_ => contextServices.GetService<IModel>())
+            // for directive generation
+                    .AddTransient(_ => contextServices.GetService<IStateManager>())
+                    .AddTransient(_ => contextServices.GetService<IInternalEntityEntryFactory>())
+                    .AddTransient<RuntimeTypeDiscoverer>()
```
bricelam: This one isn't coming from the `DbContext`'s services, so it probably doesn't belong in this method. Perhaps `ConfigureServices`?

src/EntityFramework.Commands/Design/Internal/DesignTimeServicesBuilder.cs
```diff
@@ -83,18 +86,25 @@ private void ConfigureProviderServices(string provider, IServiceCollection servi
...
-                .AddTransient(_ => contextServices.GetService<IDbContextOptions>())
-                .AddTransient(_ => contextServices.GetService<IHistoryRepository>())
-                .AddTransient(_ => contextServices.GetService<ILoggerFactory>())
-                .AddTransient(_ => contextServices.GetService<IMigrationsAssembly>())
-                .AddTransient(_ => contextServices.GetService<IMigrationsIdGenerator>())
-                .AddTransient(_ => contextServices.GetService<IMigrationsModelDiffer>())
-                .AddTransient(_ => contextServices.GetService<IMigrator>())
-                .AddTransient(_ => contextServices.GetService<IModel>());
+        {
+            // for migrations
```
bricelam: I don't think there's value in differentiating; I'm sure they all use things like `DbContex` and `ILoggerFactory`...

src/EntityFramework.Commands/Design/Internal/DesignTimeServicesBuilder.cs
```diff
@@ -83,18 +86,25 @@ private void ConfigureProviderServices(string provider, IServiceCollection servi
...
-                .AddTransient(_ => contextServices.GetService<IDatabaseProviderServices>())
-                .AddTransient(_ => contextServices.GetService<IDbContextOptions>())
-                .AddTransient(_ => contextServices.GetService<IHistoryRepository>())
-                .AddTransient(_ => contextServices.GetService<ILoggerFactory>())
-                .AddTransient(_ => contextServices.GetService<IMigrationsAssembly>())
-                .AddTransient(_ => contextServices.GetService<IMigrationsIdGenerator>())
-                .AddTransient(_ => contextServices.GetService<IMigrationsModelDiffer>())
-                .AddTransient(_ => contextServices.GetService<IMigrator>())
-                .AddTransient(_ => contextServices.GetService<IModel>());
+        {
```
bricelam: My expression-bodied syntax! :scream_cat:

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+            typeof(double),
+            typeof(bool),
+            typeof(string),
+            typeof(char),
+            typeof(byte),
+            typeof(TimeSpan),
+            typeof(DateTime),
+            typeof(DateTimeOffset),
+            typeof(Guid)
+        };
```
bricelam: Hmm... We don't really have a well-defined list of primitives anymore. E.g. on SqlServer, [SqlInt32](https://msdn.microsoft.com/en-us/library/system.data.sqltypes.sqlint32.aspx) might be considered a primitive type at some point...

bricelam: Could we get a list from the provider?

bricelam: Also, should this include nullable versions?

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+            _model = model;
+            _stateManager = stateManager;
+            _entityEntryFactory = entityEntryFactory;
+        }
+
+        public virtual ICollection<MemberInfo> Discover([CanBeNull] params Assembly[] assemblies)
+        {
+            _entityTypes = (from e in _model.GetEntityTypes()
+                            where e.HasClrType()
+                            select e.ClrType).ToList();
```
bricelam: I recently discovered a new pattern that I prefer for mixing comprehension syntax and methods:
```C#
Enumerable.ToList(
    from i in items
    select i);
```
I'm also a firm believer in artistic license, so pick whichever one floats your boat. :boat:

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+                                       from collection in new[]
+                                       {
+                                           typeof(IEnumerable<>),
+                                           typeof(IAsyncEnumerable<>),
+                                           typeof(IOrderedEnumerable<>),
+                                           typeof(ICollection<>),
+                                           typeof(IList<>),
+                                           typeof(HashSet<>),
+                                           typeof(List<>)
+                                       }
```
bricelam: Like the primitives, this list also doesn't feel fully defined.

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+                var genericMethods = from assembly in assemblies
+                                     from t in assembly.DefinedTypes
+                                     from m in t.DeclaredMethods
+                                     where m.GetCustomAttributes<GenericMethodFactoryAttribute>().Any()
+                                     from boundMethod in MakeGenericMethods(m)
+                                     select boundMethod;
+
+                members.AddRange(genericMethods);
+
+            }
```
bricelam: The StyleCop [rejects](http://www.stylecop.com/docs/SA1013.html) your closing curly bracket spacing. :trollface:

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+
+            members.AddRange(_entityTypes.Select(t => t.GetTypeInfo()));
+            members.AddRange(_entityTypesCollections.Select(t => t.GetTypeInfo()));
+
+            foreach (var entityType in _model.GetEntityTypes()
+                .Where(e => e.HasClrType())
+                .OfType<EntityType>())
```
bricelam: Oh weird. I'm not sure I like using a placeholder type for this. Could these be expanded eagerly in `MakeGenericMethods(GenericMethodFactoryAttribute)`?

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+
+            return members;
+        }
+
+        private IEnumerable<MethodInfo> MakeGenericMethods(MethodInfo declaringMethod)
+        {
+            var usages = declaringMethod.GetCustomAttributes<GenericMethodFactoryAttribute>();
+            foreach (var usage in usages)
+            {
+                var methodDefinition = string.IsNullOrEmpty(usage.MethodName)
```
bricelam: Could this fallback behavior be moved to the MethodName getter?

bricelam: (same for `TargetType`)

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+                    ? declaringMethod
+                    : (usage.TargetType ?? declaringMethod.DeclaringType)
+                        .GetTypeInfo()
+                        .GetDeclaredMethods(usage.MethodName)
+                        .FirstOrDefault(m => m.GetGenericArguments().Length == usage.TypeArguments.Length);
+
+                Debug.Assert(methodDefinition != null,
+                    "Could not find target method of " + nameof(GenericMethodFactoryAttribute) +
+                    $" on {declaringMethod.DeclaringType?.FullName}.{declaringMethod.Name}. Check the type arguments align");
```
bricelam: Feels like this should be an exception, not an assert.

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+                foreach (var typeArgs in BindTypeArguments(new ArraySegment<Type>(usage.TypeArguments)).Distinct())
+                {
+                    MethodInfo boundMethod;
+                    try
+                    {
+                        boundMethod = methodDefinition.MakeGenericMethod(typeArgs);
+                    }
+                    catch (ArgumentException)
+                    {
+                        continue;
```
bricelam: Log a warning?

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+                    }
+                    yield return boundMethod;
+                }
+            }
+        }
+
+        private IEnumerable<Type[]> BindTypeArguments(ArraySegment<Type> typeArguments)
+        {
+            if (typeArguments.Count == 0)
```
bricelam: Hmm, seems impossible...

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+                typeParameters = _navPropTypes;
+            }
+            else if (nextUnbound == typeof(TypeArgumentCategory.Primitives))
+            {
+                typeParameters = _primitiveTypes;
+            }
+            else if (nextUnbound == typeof(TypeArgumentCategory.PrimitiveCollections))
+            {
+                typeParameters = _primitiveTypesCollections;
+            }
```
bricelam: Yeah, I really don't like these marker types. Maybe we need two attributes--one that can take types and one that can take categories.

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+        private IEnumerable<TypeInfo> GetKeyTypes(EntityType entityType)
+        {
+            foreach (var key in entityType.GetKeys())
+            {
+                yield return key.GetWeakReferenceIdentityMapFactory().Invoke().GetType().GetTypeInfo();
+            }
+        }
```
bricelam: Could avoid generating an enumerator for this at compile time and just `return entityType.GetKeys().Select(key => key.GetWeakReferenceIdentityMapFactory().Invoke().GetType().GetTypeInfo());

bricelam: `yield`, like `async` should be avoided when possible. They're backed by a lot of compiler magic.

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+            yield return entityType.OriginalValuesFactory(entry).GetType().GetTypeInfo();
+            yield return entityType.RelationshipSnapshotFactory(entry).GetType().GetTypeInfo();
+            yield return entityType.GetEmptyShadowValuesFactory()().GetType().GetTypeInfo();
+        }
+
+        private IEnumerable<TypeInfo> GetPropertyTypes(EntityType entityType)
+        {
+            foreach (var prop in entityType.GetProperties())
+            {
+                prop.GetPropertyAccessors();
```
bricelam: Do you need this method's side effects? Could you add a comment--it isn't obvious.

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
```diff
@@ -0,0 +1,276 @@
...
+
+                if (prop.IsShadowProperty)
+                {
+                    continue;
+                }
+
+                yield return prop.GetGetter().GetType().GetTypeInfo();
+                yield return prop.GetSetter().GetType().GetTypeInfo();
```
bricelam: A lot of these methods are returning `.GetType().GetTypeInfo()`. It might be cleaner to let them return `IEnumerable<Type>` and project `.GetTypeInfo()` in the thing that consumes them.

src/EntityFramework.Commands/Design/ProjectOperations.cs
```diff
@@ -0,0 +1,98 @@
...
+using Microsoft.Data.Entity.Metadata.Internal;
+using Microsoft.Data.Entity.Scaffolding;
+using Microsoft.Data.Entity.Storage;
+using Microsoft.Data.Entity.Utilities;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Data.Entity.Design
+{
+    public class ProjectOperations
```
bricelam: It feels like this should be merged into `DbContextOperations`.

src/EntityFramework.Commands/Design/ProjectOperations.cs
```diff
@@ -0,0 +1,98 @@
...
+                startupAssemblyName,
+                environment);
+
+            var startup = new StartupInvoker(startupAssemblyName, environment);
+            _servicesBuilder = new DesignTimeServicesBuilder(startup);
+        }
+
+        public virtual DirectiveFiles GenerateRuntimeDirectives()
+        {
+            var generator = new DirectiveGenerator();
```
bricelam: Should this come from the design-time service provider? Users could theoretically provide their own implementation...

src/EntityFramework.Commands/Design/ProjectOperations.cs
```diff
@@ -0,0 +1,98 @@
...
+            var members = new List<MemberInfo>();
+
+            foreach (var contextType in _contextOperations.GetContextTypes())
+            {
+                var contextTypeName = contextType.GetTypeInfo().FullName;
+
+                _logger.Value.LogInformation(CommandsStrings.LogUseContext(contextTypeName));
+
+                using (var context = _contextOperations.CreateContext(contextType))
```
bricelam: This repeats a lot of work that was just performed by `_contextOperations.GetContextTypes()`. Maybe we should have a `CreateContexts` method.

src/EntityFramework.Commands/Design/ProjectOperations.cs
```diff
@@ -0,0 +1,98 @@
...
+                        context.GetInfrastructure().GetRequiredService<IDbContextServices>().DatabaseProviderServices.GetType().GetTypeInfo().Assembly));
+
+                    _logger.Value.LogDebug(CommandsStrings.EndRuntimeTypeDiscovery(members.Count - start, contextTypeName));
+                }
+            }
+            var xml = generator.GenerateXml(members);
+
+            var filename = Path.Combine(_projectDir, "Properties", "EntityFramework.g.rd.xml");
+
+            if (!Directory.Exists(Path.GetDirectoryName(filename)))
```
bricelam: I think you can just call Create. It will no-op if it already exists.

src/EntityFramework.Commands/Scaffolding/Internal/DirectiveFiles.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+
+namespace Microsoft.Data.Entity.Design
+{
+    public class DirectiveFiles
+    {
+        public virtual string GeneratedFile { get; [param: NotNull] set; }
+    }
```
bricelam: Do we ever anticipate generating other artifacts? If not, this type can probably go away.

src/EntityFramework.Commands/Scaffolding/DirectiveGenerator.cs
```diff
@@ -0,0 +1,144 @@
...
+                sb.Append("{")
+                    .Append(SerializeTypeArgs(typeArgs))
+                    .Append("}");
+            }
+
+            return sb.ToString();
+        }
+
+        private static string SerializeTypeArgs(Type[] typeArgs) 
+            => (typeArgs.Length == 0)
```
bricelam: Impossible?

src/EntityFramework.Commands/Scaffolding/DirectiveGenerator.cs
```diff
@@ -0,0 +1,144 @@
...
+            var document = new XDocument(new XDeclaration("1.0", "utf-8", "yes"));
+            document.Add(AutogenComment);
+
+            var root = new XElement(ns + "Directives", new XAttribute("xmlns", ns));
+            document.Add(root);
+
+            var library = new XElement(ns + "Library");
+            root.Add(library);
+
+            foreach (var member in members)
```
bricelam: This file might get diffed. We may want to add some explicit sorting to minimize differences.

src/EntityFramework.Commands/tools/EntityFramework.psm1
```diff
@@ -426,6 +426,63 @@ function Scaffold-DbContext {
...
+}
+
+
+<#
+.SYNOPSIS
+    Scaffolds a runtime directive file for .NET Native.
+
+.DESCRIPTION
+    Scaffolds a runtime directive file for .NET Native. 
+	These directives instruct the compiler to include metadata for types needed at runtime.
```
bricelam: No tabs :space_invader:

bricelam: (other places and in `EntityFramework.psd1` too)

bricelam: P.S. I use the [Productivity Power Tools](https://visualstudiogallery.msdn.microsoft.com/34ebc6a2-2777-421d-8914-e29c1dfa7f5d) to catch this.

src/EntityFramework.Commands/tools/EntityFramework.psm1
```diff
@@ -426,6 +426,63 @@ function Scaffold-DbContext {
...
+        [string] $Project,
+        [string] $StartupProject,
+        [string] $Environment)
+
+	$values = ProcessCommonParameters $StartupProject $Project
+    $dteStartupProject = $values.StartupProject
+	$config = $dteStartupProject.ConfigurationManager.ActiveConfiguration.ConfigurationName
+
+	if ( $config -ne "Debug") {
+		Write-Warning "Cannot run directive scaffolding in '$config' mode. Set the solution to 'Debug' and try again."
```
bricelam: Technically, we can't even run it in Debug mode if they've checked **Compile with the .NET Native tool chain**. It looks like we can detect this with the following.
```ps1
GetProperty $dteStartupProject.ConfigurationManager.ActiveConfiguration.Properties ProjectN.UseDotNetNativeToolchain
```

bricelam: Also, this is terminating so it should `throw` instead of warn.

src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -236,27 +236,27 @@ public virtual void MarkAsTemporary([NotNull] IProperty property, bool isTempora
             = typeof(InternalEntityEntry).GetTypeInfo().GetDeclaredMethod(nameof(ReadShadowValue));
 
         [UsedImplicitly]
-        protected virtual T ReadShadowValue<T>(int shadowIndex) => default(T);
+        protected internal virtual T ReadShadowValue<T>(int shadowIndex) => default(T);
```
bricelam: We're generally against `internal` members...

src/EntityFramework.Core/Internal/GenericMethodFactoryAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        ///     Name of the method created
+        /// </summary>
+        public virtual string MethodName { get; [param: NotNull] set; }
+
+        /// <summary>
+        ///     Type which contains the method to be created
+        /// </summary>
+        public virtual Type TargetType { get; [param: NotNull] set; }
```
bricelam: I would also have a constructor overload for this. (`.ctor(Type, string, Type[])`)

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
bricelam: Scratch that. I thought this was `TypeArgumentCategory.EntityTypes`

src/EntityFramework.Core/Internal/TypeArgumentCategory.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.Data.Entity.Metadata;
+using Microsoft.Data.Entity.Query;
+
+namespace Microsoft.Data.Entity.Internal
+{
+    /// <summary>
+    ///     Represents categories of type arguments that may be used at runtime.
+    ///     Runtime-directive scaffolding uses these categories to compute possible combinations of type arguments
+    /// </summary>
+    public static class TypeArgumentCategory
```
bricelam: Again, I think this should be a flags enum that is either a separate property on the attribute or a separate attribute.

bricelam: Hmm... I'm starting to see why it was done this way. You could have `SomeMethod<TEntity, int>` and so they need to be mixed in the array.

bricelam: I wonder if a syntax like this would work
```C#
[GenericMethodFactory(new[] { TypeArgument.EntityTypes, TypeArgument.Of<int>() })]
```

src/EntityFramework.Commands/Design/ProjectOperations.cs
```diff
@@ -0,0 +1,98 @@
...
+                    members.AddRange(discoverer.Discover(typeof(EntityType).GetTypeInfo().Assembly, 
+                        typeof(RelationalDatabase).GetTypeInfo().Assembly,
+                        context.GetInfrastructure().GetRequiredService<IDbContextServices>().DatabaseProviderServices.GetType().GetTypeInfo().Assembly));
+
+                    _logger.Value.LogDebug(CommandsStrings.EndRuntimeTypeDiscovery(members.Count - start, contextTypeName));
+                }
+            }
+            var xml = generator.GenerateXml(members);
+
+            var filename = Path.Combine(_projectDir, "Properties", "EntityFramework.g.rd.xml");
```
bricelam: Will this change with MSBuild integration? E.g. Generate it behind the scenes under `obj/` and include it in the compilation. If so, maybe this command should have an `-OuputFile` parameter.

src/EntityFramework.Core/Internal/GenericMethodFactoryAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        public GenericMethodFactoryAttribute([NotNull] string methodName, [NotNull] params Type[] typeArguments)
+            : this(typeArguments)
+        {
+            Check.NotEmpty(methodName, nameof(methodName));
+
+            MethodName = methodName;
+        }
+
+        public GenericMethodFactoryAttribute([NotNull] params Type[] typeArguments)
```
bricelam: It looks like we don't use this overload, can remove for now.

bricelam:
LGTM. We may want to go over the high-level design in a meeting to get feedback/buy-off from the team.

src/EntityFramework.Commands/Design/ProjectOperations.cs
natemcmaster: Yes, I think an output path would be generally useful. I'll add to the contract.

However, I don’t think there is a need to surface this in powershell. In fact, I specifically avoided. I'm considering adding this to our uwp MSBuild integration (in place of the invisible, auto-gen approach).

```xml
<Error Condition="!Exists('Properties\EntityFramework.g.rd.xml')" Text="hey! you need to run rdxml gen" />
```
This may have its own problems (if user renames the file), but it could be generally useful as a default.

src/EntityFramework.Core/Internal/TypeArgumentCategory.cs
natemcmaster: Yup, you figured out why I needed mark types. Although it reads well, `TypeArgument.Of<int>()` isn't a constant expression so it can't be in an attribute parameter list.

src/EntityFramework.Core/ChangeTracking/Internal/InternalEntityEntry.cs
natemcmaster: Is preserving `nameof()` references a compelling enough reason to allow this?

src/EntityFramework.Commands/Scaffolding/DirectiveGenerator.cs
natemcmaster: Yes. This was left behind as I moved code around.

src/EntityFramework.Commands/Scaffolding/Internal/DirectiveFiles.cs
natemcmaster: It is possible. I have some potential use cases in mind, like dropping multiple files or multiple types of files. I'd like to leave as we may do further changes to output a we build the msbuild client.

src/EntityFramework.Core/Internal/TypeArgumentCategory.cs
bricelam: Doh, silly attributes.

src/EntityFramework.Commands/Design/ProjectOperations.cs
natemcmaster: Seems like YAGNI.

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
natemcmaster: As with primitives, this discoverer does not (yet) make an exhaustive list. This is a TODO - increase coverage.

src/EntityFramework.Commands/Design/Internal/RuntimeTypeDiscoverer.cs
natemcmaster: For now, Sqlite and inmemory are our only types that provider that works on UWP. This could be improved to use the type mapper to make sure all supported types are included. And yes, this should be expanded to include nullables.

src/EntityFramework.Commands/Scaffolding/DirectiveGenerator.cs
natemcmaster: Good idea. I’ll sort alphabetically.

natemcmaster:
@bricelam I've updated this PR with updates to the little tweaks. 

I also removed `ProjectOperations` and put it in rdxml gen in `DbContextOperations`. This prevented redundant runs of FindContextTypes

src/EntityFramework.Commands/Design/Internal/DesignTimeServicesBuilder.cs
```diff
@@ -68,6 +69,7 @@ protected virtual void ConfigureServices([NotNull] IServiceCollection services)
                 .AddSingleton<CSharpMigrationOperationGenerator>()
                 .AddSingleton<CSharpSnapshotGenerator>()
                 .AddSingleton<MigrationsCodeGenerator, CSharpMigrationsGenerator>()
+                .AddTransient<RuntimeTypeDiscoverer>()
```
bricelam: Can this be a singleton service?

src/EntityFramework.Core/Internal/GenericMethodFactoryAttribute.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.Data.Entity.Utilities;
+
+namespace Microsoft.Data.Entity.Internal
+{
+    /// <summary>
+    ///     Method uses reflection to create a generic method. e.g. MethodInfo.MakeGenericMethod().
+    ///     Runtime-directive scaffolding uses this attribute to compute the possible runtime arguments that may be passed to MakeGenericMethod.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = false)]
+    public class GenericMethodFactoryAttribute : Attribute
```
bricelam: `CallsMakeGenericMethodAttribue`?

bricelam:
:shipit:

src/EntityFramework.Core/Internal/GenericMethodFactoryAttribute.cs
natemcmaster: :+1:  I like this name. Will update PR.

src/EntityFramework.Commands/Design/Internal/DesignTimeServicesBuilder.cs
natemcmaster: Yes. Will update.

JensKuehner:
Hi,
I tried the new feature. One hint is to restart visual studio after you have upgrated the nuget packages.
Now I got the folowing errors in the nuget console:
PM> scaffold-directives
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Data.Entity.Metadata.Internal.ClrPropertySetterFactory.Create(PropertyInfo property)
   at Microsoft.Data.Entity.Metadata.Internal.ClrAccessorFactory`1.Create(IPropertyBase property)
   at Microsoft.Data.Entity.Metadata.Internal.Property.<get_Setter>b__115_0()
   at System.Threading.LazyInitializer.EnsureInitializedCore[T](T& target, Func`1 valueFactory)
   at System.Threading.LazyInitializer.EnsureInitialized[T](T& target, Func`1 valueFactory)
   at Microsoft.Data.Entity.Metadata.Internal.Property.get_Setter()
   at Microsoft.Data.Entity.Metadata.Internal.PropertyBaseExtensions.GetSetter(IPropertyBase propertyBase)
   at Microsoft.Data.Entity.Design.Internal.RuntimeTypeDiscoverer.<GetRuntimeTypes>d__17.MoveNext()
   at System.Linq.Enumerable.<SelectManyIterator>d__22`3.MoveNext()
   at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
   at System.Collections.Generic.List`1.AddRange(IEnumerable`1 collection)
   at Microsoft.Data.Entity.Design.Internal.RuntimeTypeDiscoverer.Discover(Assembly[] assemblies)
   at Microsoft.Data.Entity.Design.DbContextOperations.GenerateRuntimeDirectives()
   at Microsoft.Data.Entity.Design.OperationExecutor.<ScaffoldRuntimeDirectivesImpl>d__21.MoveNext()
   at Microsoft.Data.Entity.Design.OperationExecutor.OperationBase.<>c__DisplayClass4_0`1.<Execute>b__0()
   at Microsoft.Data.Entity.Design.OperationExecutor.OperationBase.Execute(Action action)
Object reference not set to an instance of an object.
PM> 

thanks
Jens Kühner

natemcmaster:
@JensKuehner can you open an new issue with this info and share your model?

JensKuehner:
ok. Do it tomorrow.

JensKuehner:
I found it. I cannot use eg. .Ignore(nameof(MyProp)). I need to use .Ignore(p => p.MyProp)

natemcmaster:
@JensKuehner can you clarify? Is `Scaffold-Directives` not working because you forgot to use `.Ignore()`? Or does using `.Ignore()` cause the error?

JensKuehner:
I have some readonly properties. I need to ignore them. And I need the expression for propertyies p => p.MyProp instead of nameof.

JensKuehner:
If I would specify the property name as string via nameof it wont work.
  modelBuilder
            .Entity<Employee>(e =>
            {
                e.Property(p => p.FirstName).IsRequired();
                e.Property(p => p.Pin).IsRequired();
                e.Ignore(p => p.DisplayName);
            })

rowanmiller:
@JensKuehner if you are still seeing the issue with `nameof` can you open a new issue and include a complete code listing (or project) that we can use to reproduce the issue.

PR #4199, opened 2015-12-30
Create analyzer for MakeGenericMethod

natemcmaster:
A Roslyn analyzer that finds all usages of `MethodInfo.MakeGenericMethod` that (1) don't have a `GenericMethodFactory` attribute (see #4197) OR (2) have the inspection disabled via comment.

:warning: The error level is set to "Information" for now. I'll upgrade to "Warning" once we solve the problem

The analyzer only runs in VS2015. Doesn't run or build on CI and is never published. This is designed only for EF project authors, not general use.

cc @bricelam


natemcmaster:
Not worth adding until https://github.com/dotnet/cli/issues/83

PR #4229, opened 2016-01-06
Fix #3732 - Include and Take generate incorrect SQL without Skip

### LINKED ISSUE #3732: Include and Take generate incorrect SQL without Skip
**Version:** Beta 8
**Model:** EFGetStarted.Blog
**DB:** SQLite

**Code:**
```csharp  
db.Blogs.Include(x => x.Posts).OrderBy(x => x.Url).Take(10).ToList(); 
```
*Note: The OrderBy call is not necessary to repro, but more clearly shows the problem (that not all posts are included)*

**SQL:**
```sql
SELECT "x"."BlogId", "x"."Url"
FROM "Blog" AS "x"
ORDER BY "x"."Url", "x"."BlogId"
LIMIT 10

SELECT "p"."PostId", "p"."BlogId", "p"."Content", "p"."Title"
FROM "Post" AS "p"
INNER JOIN (
    SELECT DISTINCT "x"."Url", "x"."BlogId"
    FROM "Blog" AS "x"
    LIMIT 10
) AS "x" ON "p"."BlogId" = "x"."BlogId"
ORDER BY "x"."Url", "x"."BlogId"
```
*(There should be an ORDER BY in the inner query)*

**Workaround:**
```csharp  
db.Blogs.Include(x => x.Posts).OrderBy(x => x.Url).Skip(0).Take(10).ToList(); 
```

anpete:
When lifting OrderBys on Include subqueries we were always clearing the OrderBy on the subquery. However, if the subquery has a limit or offset specified, then we need to leave the OrderBy intact.

### DIFF
--- a/src/EntityFramework.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors.Internal
                     = ((ConstantExpression)node.Arguments[2]).Value
                         as Shaper;
 
-                if ((entityShaper != null)
+                if (entityShaper != null
                     && entityShaper.IsShaperForQuerySource(_querySource))
                 {
                     var resultType = node.Method.GetGenericArguments()[0];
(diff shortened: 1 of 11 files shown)


anpete:
@mikary @maumar

maumar:
:shipit:

PR #4230, opened 2016-01-06
Skip tests that fail on Mono < 4.2

### LINKED ISSUE #3594: Fix the build on non-Windows platforms

natemcmaster:
Close #3594.

Skip tests that won't pass on Mono 4.0.x

cc @cesarbs

### DIFF
--- a/test/EntityFramework.Core.FunctionalTests/AsNoTrackingTestBase.cs
+++ b/test/EntityFramework.Core.FunctionalTests/AsNoTrackingTestBase.cs
@@ -10,10 +10,11 @@ using Xunit;
 
 namespace Microsoft.Data.Entity.FunctionalTests
 {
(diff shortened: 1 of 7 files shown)


divega:
:shipit:

So all new tests in this classes should be skipped in Mono < 4.2? Is there a specific pattern that will make the tests fail that would be worth describing in a comment alongside the class level attribute?

natemcmaster:
> So all new tests in this classes should be skipped in Mono < 4.2?

All new tests marked as `[ConditionalFact]` will be skipped. The test condition on the class applies to all contained methods.

>  Is there a specific pattern that will make the tests fail that would be worth describing in a comment alongside the class level attribute?

Pretty much any test that uses query will fail.

divega:
> Pretty much any test that uses query will fail.

Got it. And that is why you even made tests that had `[Fact]` commented out `[ConditionalFact]` :smile:

PR #4239, opened 2016-01-06
Allow self referencing relationships without navigations

### LINKED ISSUE #4069: Migration fails for class with multiple self references
I have a class with two self references. Like this one:

	public class AClass
	{
		public int Id { get; set; }

		public AClass AClassFirst { get; set; }

		public AClass AClassPrevious { get; set; }
	}

when I call "dnx ef database update" I get:

	The navigation property '' cannot be added to the entity type 
	'AocSample.Models.AClass' because a navigation property with 
	the same name already exists on entity type 'AocSample.Models.AClass'.

This does not occur if there is only one self reference or if I reference another class two times. This happens with RC1 and with the source I fetched yesterday. Call Stack with line numbers attached:
[ef-error-multiple-self-references.txt](https://github.com/aspnet/EntityFramework/files/61166/ef-error-multiple-self-references.txt)

PS: First, I'm new to EF, .NET 5, etc. So I hope I do not oversee something. Second I do not know if the following is also considered an issue I should post somewhere: I had big troubles with dnvm/dnx/dnu and building EF from the source because of charset issues. Although I do not even use blanks and German umlauts in file and folder names I had chosen a Windows user name with accents in it (about 15 years ago and kept it). The problem is that cmd is not unicode aware. So the path dnvm adds (including the user directory name with the special characters) cannot be found later. I switched to PowerShell so I could executed the database migrations. But when I wanted to build EF from the source this did not work because the build script calls cmd (with dnu) multiple times. I found a hack to change the cmd charset to one which is unicode aware. I tried to set the path globally. Did not help. Finally I found that the runtimes can be installed machine wide (global). Then I could build EF with the latest runtime. But Visual Studio installed the runtime (older versions) again in the user directory. I did no try any further but finally changed my user name.

smitpatel:
Fixes #4069

### DIFF
--- a/src/EntityFramework.Core/Metadata/Builders/ReferenceNavigationBuilder.cs
+++ b/src/EntityFramework.Core/Metadata/Builders/ReferenceNavigationBuilder.cs
@@ -114,6 +114,7 @@ namespace Microsoft.Data.Entity.Metadata.Builders
         protected virtual InternalRelationshipBuilder WithOneBuilder([CanBeNull] string reference)
         {
             if (Builder.Metadata.IsSelfReferencing()
+                && (reference != null)
                 && (ReferenceName == reference))
             {
                 throw new InvalidOperationException(CoreStrings.DuplicateNavigation(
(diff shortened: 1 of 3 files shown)


anpete:
:shipit:

PR #4240, opened 2016-01-06
Allow Fk to use inherited properties

### LINKED ISSUE #2514: Metadata: Allow indexes and foreign keys to use inherited properties

smitpatel:
Resolves #2514


AndriySvyryd:
:shipit:

PR #4250, opened 2016-01-07
Fix #1909 - Do not append Select when result operator performs projection

### LINKED ISSUE #1909: Getting InvalidCastException when grouping over a projection
I'm not sure if this has been already addressed or not. I'm testing it with Beta 3 version.

The following code:
```
var db = new SampleContext();
var result = db.Blogs
	.Select(x => new { x.BlogId })
	.GroupBy(x => x.BlogId);
result.ToList();
```
is producing InvalidCastException:
```
System.InvalidCastException: Unable to cast object of type 'Microsoft.Data.Entity.Query.QuerySourceScope`1[<>f__AnonymousType0`1[System.Int32]]' to type 'Microsoft.Data.Entity.Query.QuerySourceScope`
   at Microsoft.Data.Entity.Query.QuerySourceScope.GetResult[TResult](IQuerySource querySource)
```

Here is the context and model declaration:

```
    public class SampleContext : DbContext
	{

		protected override void OnConfiguring(DbContextOptions builder)
		{
			builder.UseSqlServer(@"Server=(localdb)\v11.0;Database=Samples;Trusted_Connection=True;");
		}
	}

	public class Blog
	{
	}
```

mikary:
Fixes exception thrown due to type mismatch between operator an expression sequence type
@anpete @maumar

### DIFF
--- a/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
@@ -21,6 +21,7 @@ using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Clauses.ExpressionVisitors;
+using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Clauses.StreamedData;
 
 namespace Microsoft.Data.Entity.Query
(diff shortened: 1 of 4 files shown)


src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -33,6 +34,12 @@ public abstract class EntityQueryModelVisitor : QueryModelVisitorBase
         public static readonly MethodInfo PropertyMethodInfo
             = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
 
+        private static readonly HashSet<Type> _projectingResultOperators = new HashSet<Type>
```
anpete: Do this inline and save the allocation here

anpete:
:shipit:

PR #4252, opened 2016-01-07
[RevEng] Warn if user selects a schema or table and that selection is not satisfied

### LINKED ISSUE #4168: Scaffold a subset of tables - silently ignores / skips tables that cannot be found

lajones:
Fixes #4168.


src/EntityFramework.Relational.Design/TableSelectionSet.cs
```diff
@@ -22,11 +24,24 @@ public TableSelectionSet([CanBeNull] IReadOnlyList<string> tables)
...
+        public class Selection
+        {
+            public Selection([NotNull] string selection)
+            {
+                Check.NotEmpty(selection, nameof(selection));
+
+                Name = selection;
+            }
+
+            public virtual string Name { get; [param: NotNull] set; }
```
lajones: Note to self: I think `Text` might be a better name for this. It represents the text the user inputted on the command-line after the `-s` or `-t` flag.

test/EntityFramework.MicrosoftSqlServer.Design.Tests/SqlServerTableSelectionSetExtensionsTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            Assert.True(schemaASelectionWithSquareBrackets.IsSatisfied);
+            Assert.False(schemaBSelection.IsSatisfied);
+            Assert.False(schemaBSelectionWithSquareBrackets.IsSatisfied);
+            Assert.True(table0Selection.IsSatisfied);
+            Assert.True(table0SelectionWithSquareBrackets.IsSatisfied);
+            Assert.False(table1Selection.IsSatisfied);
+            Assert.False(table1SelectionWithSquareBrackets.IsSatisfied);
+
+            // When Allows() runs against a non-selected schema/table no further schema or table selection is marked true
+            Assert.False(SqlServerTableSelectionSetExtensions.Allows(tableSelectionSet, "schemaNonExistent", "tableNotThere"));
```
lajones: Note to self: `"schemaNonExistent"` should be `"schemaOther"`. It does exist in the database. Similarly for `"tableNotThere"`.

natemcmaster:
It appears we iterate over `IList` multiple times. I'm concerned this have poor performance even with a few hundred selections. We can make this more efficient by using a different collection type, such as `HashSet` or `Dictionary`. We don't need `IList` because order does not matter and selections are unique.

lajones:
@natemcmaster If `S` is number of selections and `N` is number of tables in the database then, yes, perf characteristics are `O(S * N)`. But, in order to raise the error message, we have to do this as it might be the `N`th table that satisfies the `S`th selection. This is slower than before (where we just stopped at the first selection that matched) - but on average that was still `S/2` comparisons - so still `O(S * N)`.

For SQLite it's possible that we could do something with a simple `HashSet` as we could strip out any duplicate selections (presumably typos) that the user has entered and then it would be certain that 1 table would only satisfy 1 selection. But for SQL Server this is more complicated as 1 table can satisfy multiple table selections if the user has e.g. somehow entered e.g. both `-t table1` and `-t schema1.table1` - entries which we cannot combine without knowing what data we are going to see. So unfortunately I think we do have to check every selection against every table.

But I share your concern. How about I raise a separate issue to look into seeing if there are any perf improvements we can make instead?

src/EntityFramework.Sqlite.Design/SqliteTableSelectionSetExtensions.cs
```diff
@@ -8,14 +8,38 @@ namespace Microsoft.Data.Entity.Scaffolding
...
+        /// to mark that they have been satisfied.
+        /// </summary>
+        /// <param name="tableSelectionSet"> the <see cref="TableSelectionSet" /> to test </param>
+        /// <param name="tableName"> name of the database table to check </param>
+        /// <returns> whether or not the table is allowed </returns>
         public static bool Allows(this TableSelectionSet tableSet, string tableName)
         {
-            if ((tableSet == null)
-                || (tableSet.Tables.Count == 0))
+            // mark all schema selections as satisfied as sqlite ignores them
```
lajones: @natemcmaster @rowanmiller I'd like some advice on this. At the moment I've coded it so that any `-s` entries the user makes for SQLite are basically silently ignored - they will not cause a warning saying that the selection is not satisfied, nor will they cause any warning that using a `-s` flag under SQLite is basically meaningless. I can continue to take this approach or I can:

A) take out this section - which means if the user enters any `-s` flags all of them will raise warnings saying they are unsatisfied.
B) check for `-s` flags much earlier in the process - the question then is do I a) error - i.e. issue an error message and stop the process; or b) warn that `-s` flag is meaningless but continue, and when I continue do I also raise the warning messages that the `-s` flags are unsatisfied?

rowanmiller: My vote would be warn if the flag is present and then just ignore it from that point on

natemcmaster: I’ll second that. Warn but carry on ignoring `-s`.

lajones: Will do.

natemcmaster:
> How about I raise a separate issue to look into seeing if there are any perf improvements we can make instead?

@lajones Do you have perf numbers? Although academically perf could be improved, this a "user-time" tool. I doubt users will care if scaffolding takes 5 seconds instead of 3.

My primary concern is that perf for even 100-200 tables can be "slow" (which is again, a matter of user-preception.) But I haven't measured how this PR impacts perf to in order to compare. My estimate could be wrong.

lajones:
@natemcmaster No perf numbers yet. I'll give it a try on AdventureWorks.

rowanmiller:
Yeah we should at least sanity test perf before pushing this if we think it might be an issue (my guess is that this extra in-memory work won't add much overhead).

lajones:
@natemcmaster @rowanmiller Did a quick test on a DB with 1000 tables, selecting 600. Numbers consistently came out between 11 and 12 secs. For comparison AdventureWorks (71 tables, selecting all 71 individually) took around 10.5 secs and a small DB with only 2 tables took around 10 secs. So I don't think this is the performance bottleneck.

rowanmiller:
Those look fine. Do we know what the numbers were before the change?

lajones:
@rowanmiller Just tried it - and I got a range of timings for the 3 DBs mentioned above which were similar to the range I got without the change.

rowanmiller:
:+1:

natemcmaster:
Ok. Glad it doesn't kill perf. Then don't worry about algorithmic improvements for now, but let's add a TODO in code.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerTableSelectionSetExtensions.cs
```diff
@@ -9,26 +10,67 @@ namespace Microsoft.Data.Entity.Scaffolding
...
 
-            if (tableSelectionSet.Schemas.Contains(schemaName))
+            var result = false;
+
+            foreach (var pattern in schemaPatterns)
             {
-                return true;
+                var patternToMatch = pattern.Replace("{schema}", schemaName);
+                var matchingSchemaSelections = tableSelectionSet.Schemas.Where(
+                    s => s.Name.Equals(patternToMatch, StringComparison.OrdinalIgnoreCase));
```
natemcmaster: nit: code formatting. We should run R# over our design projects (separate PR).

src/EntityFramework.Relational.Design/TableSelectionSet.cs
```diff
@@ -22,11 +24,24 @@ public TableSelectionSet([CanBeNull] IReadOnlyList<string> tables)
...
+            public virtual bool IsSatisfied { get; [param: NotNull] set; }
```
natemcmaster: `Matched`

src/EntityFramework.Relational.Design/TableSelectionSet.cs
natemcmaster: How about `Pattern`?

lajones: I thought about that but I'd rather not as it makes me think we're inputting things like "Abc*" which matches anything starting with Abc - which we don't support.

src/EntityFramework.Relational.Design/TableSelectionSet.cs
lajones: Has to be IsMatched I think, but otherwise I'm fine with that.

lajones:
@natemcmaster Sounds good. Have updated as suggested and force-pushed. Ready for sign-off.

src/EntityFramework.Sqlite.Design/Properties/SqliteDesignStrings.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="UsingSchemaSelectionsWarning" xml:space="preserve">
+    <value>You have specified some schema selections. This Sqlite provider does not support these and they will be ignored. Note: it does support table selections.</value>
```
natemcmaster: Nit: `SQLite`
Also, I would shorten to one sentence if possible e.g. `Scaffolding from a SQLite database will ignore the schema filters argument.`

natemcmaster:
My only, final design suggestion is that this changes `TableSelectionSet` from immutable to mutable. We should reflect this in our usage of readonly interfaces.

src/EntityFramework.Sqlite.Design/Properties/SqliteDesignStrings.resx
lajones: OK. Will fix.

lajones:
Note: only the `Selection` classes themselves are mutable - to allow them to be marked as matched - and come to think of it the `Text` property should not be - I'll fix that. The `TableSelectionSet` itself is not mutable.

natemcmaster:
Making `Text` immutable is a good solution. This is what I really had in mind, that you could (in theory) change the selection set by updating the string. With that change, :shipit:

lajones:
Checked in with commit 50c51e7.

PR #4263, opened 2016-01-08
adding more tests (currently disabled) for left outer join cases 

maumar:
also added some more complex groupjoin + selectmany + defaultifempty queries


maumar:
@anpete

anpete:
:shipit:

Issue #2992, opened 2015-09-02
Need a way for adding Entitytypes to DbContext using ModelBuilder for scaffolding


PradeepKadubandi:
I am logging this after our discussion about scaffolding.

When scaffolding fetches the DbContext from the application using reflection, we need a way to dynamically add entity sets to DbContext (to support a scenario where user selected to scaffold a model which is not a DbSet property in the selected DbContext) - this would avoid us to have to compile the application again one more time to figure out the metadata ...

AndriySvyryd:
To make this flexible and useful for other scenarios we came up with several requirements:
1. Be able to add actions that run before or after `OnModelCreating` without changing the service provider
2. Be able to avoid calling `OnModelCreating`
3. Be able to create a model cache key for a given context instance

One way of satisfying these is with the following singleton service:
```
public interface IModelCustomizer
{
    event Action<ModelBuilder, ModelCacheKey> BeforeOnModelCreating;
    event Action<ModelBuilder, ModelCacheKey> AfterOnModelCreating;

    ModelCacheKey GetCacheKey(DbContext context);

    void Customize(ModelBuilder modelBuilder, ModelCacheKey cacheKey);
}
public class ModelCacheKey
{
    public virtual DbContext Context { get; protected set; }

    public override int GetHashCode()

    public override bool Equals(object obj)
}
```
`ModelSource` would use it instead of invoking `OnModelCreating` directly.

rowanmiller:
Closing as we discussed with @sayedihashimi and it's not clear if this is still needed. If the conclusion is that it is needed, we will re-open and implement.

Grinderofl:
I disagree with the singleton service described here. This is a scenario whereby each separate DbContext instance should have its own individual configuration, that isn't exposed to all other dbcontext instances, but rather directly injected into the dbcontext. Perhaps an addition to DbContextServices with ```IEnumerable<ModelBuilderOverride> Overrides {get; set;}```? This would also extend to `ModelBuilder.AddOverride(ModelBuilderOverride)` or `ModelBuilder.AddOverridesFromAssembly(Assembly)`

PR #4265, opened 2016-01-09
Adds and implements ModelCustomizer

Grinderofl:
Provides a possible implementation for a way to alter modelbuilder from outside dbcontext.
Resolves: #2992 and #2805

### DIFF
--- a/src/EntityFramework.Core/EntityFrameworkServiceCollectionExtensions.cs
+++ b/src/EntityFramework.Core/EntityFrameworkServiceCollectionExtensions.cs
@@ -90,6 +90,8 @@ namespace Microsoft.Extensions.DependencyInjection
                 .AddSingleton<IMemberMapper, MemberMapper>()
                 .AddSingleton<IFieldMatcher, FieldMatcher>()
                 .AddSingleton<ICoreConventionSetBuilder, CoreConventionSetBuilder>()
+                .AddSingleton<IModelCustomizer, ModelCustomizer>()
+                .AddSingleton<IModelCacheKeyFactory, ModelCacheKeyFactory>()
                 .AddSingleton<LoggingModelValidator>()
                 .AddScoped<IKeyPropagator, KeyPropagator>()
                 .AddScoped<INavigationFixer, NavigationFixer>()
--- a/src/EntityFramework.Core/Infrastructure/IModelCacheKeyFactory.cs
+++ b/src/EntityFramework.Core/Infrastructure/IModelCacheKeyFactory.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+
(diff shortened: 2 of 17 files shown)


rowanmiller:
@AndriySvyryd can you take a quick look at this one and see if the design looks good. We may still hold off to post-7.0.0 to take this one.

AndriySvyryd:
@Grinderofl, sorry I didn't explain the purpose of the model cache key. It is to be used for the model cache in ModelSource. So `ConcurrentDictionary<Type, IModel> _models` would become `ConcurrentDictionary<ModelCacheKey, IModel> _models`, the default implementation of `ModelCacheKey.Equals` would just compare the stored `DbContext` types. Note that `ModelCacheKey` shouldn't store the `DbContext` instance to avoid leaking memory. This means that `IModelCustomizer.Customize` should take a `DbContext` instead of a `ModelCacheKey`

This would allow customizing the model differently depending on the `DbContext` instance data. E.g. one could put a property for a custom schema to be used for the particular `DbContext` instance, then implement custom `IModelCustomizer` that configures the model appropriately and a custom `ModelCacheKey` that would also use the schema for equals comparison.

We don't have a consensus oh what is the best way to add custom actions, so to make the PR simpler and more likely to go in sooner we can remove the requierement to add actions without modifying the service provider. This would make the default implementation `ModelCustomizer` very simple: it would just call `OnModelCreating`, so `ModelCustomizerOptions` and `ModelCustomizerOptionsSetup` can be removed. As long as `IModelCustomizer` is in DI, the users can choose and implement whatever makes more sense for their scenario. You could move `ModelCustomizerOptions` and `ModelCustomizerOptionsSetup` to the test to ensure that it can still be used like that.

Grinderofl:
That requirement of not modifying service provider did cause a lot of headaches. I've updated the PR and removed the options. I think it makes sense to just be able to swap out the service, as most other features can be altered this way.

src/EntityFramework.Core/EntityFrameworkServiceCollectionExtensions.cs
```diff
@@ -18,6 +18,7 @@
 using Microsoft.Data.Entity.ValueGeneration;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
```
AndriySvyryd: This shouldn't be needed

AndriySvyryd:
Could you also make the `ModelCacheKey` changes?

AndriySvyryd:
@Grinderofl We've discussed this as a team and decided that the interface should be split into these two:
```
    public interface IModelCacheKeyFactory
    {
        object Create([NotNull] DbContext context);
    }
```
```
    public interface IModelCustomizer
    {
        void OnModelCreating([NotNull] ModelBuilder modelBuilder, [NotNull] DbContext dbContext);
    }
```
`ModelCustomizer` can still implement both of them, but will need to be registered twice in DI

Could you make these changes, rebase on latest dev and squash the commits?

AndriySvyryd:
Thanks for your contribution!

Issue #2272, opened 2015-05-26
Column ordering


anpete:
**Outcome from much discussion on this topic...**

In EF6 we tried having column order match the order of properties in the class. The issue is that reflection can return a different order on different architectures. Although the order of properties in the model does not theoretically matter... it actually created a lot of very difficult to debug issues on EF6 where the order had a subtle effect on conventions, model diffing, etc. We managed to fix some of them (after many releases) but there are still some issues we have never really fixed. For that reason, metadata is stored in a deterministic (alphabetical) order in metadata in EF7.

In EF6, we also had a column ordering API, but this only impacts the initial create of the table. If you later change the order, migrations does not handle this (it would require a table rebuild). Also, if you add new columns, they are always appended to the table (since anything else also requires a table rebuild).

We could, however, store an ordering hint as an annotation on the model (based on the reflection order). This would _only_ affect the order that migrations writes the columns in a create table call. That way it doesn't matter if it changes when you run your app on different platforms, since it's just an annotation that is ignored for the most part. If we wanted to, we could also flow an order that was specified in `[Column]` annotations into the hint field.

bricelam:
Only for keys?

anpete:
No, for fine grained control of table layout.

ChristineBoersen:
Just to add to this, adding the HasColumnOrder attribute isn't currently respected either and instead the columns are added in alphabetical order. 

Is there a way to work around this incomplete feature in beta 6  ?

ErikEJ:
@ChristineBoersen Why do you need to manage the colum order?

ChristineBoersen:
There are times you need to write code at the database level and column order can become more important (inserting into a table without full column specs in a trigger for example). On a wide table, it can save time and since columns in sql databases don’t change order once written easily, is usually fairly safe.

 

In addition, I prefer to look at the model and data in the same column order 99% of the time so I am not hunting for information visually (example comparing query output debugging a query versus framework output).  Maybe it’s a little OCD on my part as well as the consistency just “feels right”.

 

That or worst case obsolete the currently non-implemented .HasColumnOrder attribute so people know it is incomplete at least. :)

 

Christine

jmalatia:
Just created a db with 7.0.0-beta7-15540 and it created the table columns in alphabetical order...  I would expect it to create the table in the same order as the model is written as in EF6.  Db Column order is important for review as well as other things.

[...]

PR #4275, opened 2016-01-11
RevEng: Generate Properties in same order as columns in DB

### LINKED ISSUE #4062: Reveng: entity class properties in the same order of table fields
In reverse engineer please consider  to generate (eventually optional) the entity class with (field) properties listed not in alphabetic order but  in the same order of the fields in related table.

lajones:
Fixes #4062. 

Note: this uses a Scaffolding-only ColumnOrdinal annotation for now. If we decide to add column ordering in a more general way to the model as a whole (see discussion in #2272) then we should change this code to use that instead.

### DIFF
--- a/src/EntityFramework.Commands/Scaffolding/Internal/EntityTypeWriter.cs
+++ b/src/EntityFramework.Commands/Scaffolding/Internal/EntityTypeWriter.cs
@@ -5,6 +5,7 @@ using System.Collections.Generic;
 using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.Data.Entity.Internal;
+using Microsoft.Data.Entity.Metadata;
 using Microsoft.Data.Entity.Scaffolding.Internal.Configuration;
 using Microsoft.Data.Entity.Utilities;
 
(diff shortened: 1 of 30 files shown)


src/EntityFramework.Commands/Scaffolding/Internal/EntityTypeWriter.cs
```diff
@@ -104,7 +105,7 @@ public virtual void AddConstructor()
 
         public virtual void AddProperties()
         {
-            foreach (var property in ScaffoldingUtilities.OrderedProperties(_entity.EntityType))
```
natemcmaster: Can we remove `ScaffoldingUtilities.OrderedProperties` then?

natemcmaster:
Can you add to the `RelationalDatabaseModelFactoryTest` unit test as well?

src/EntityFramework.Commands/Scaffolding/Internal/EntityTypeWriter.cs
lajones: Afraid not. It's also used to decide in what order the fluent API is generated in OnModelCreating(). There I thought it made more sense to keep alphabetical - at least I know I would find that more useful than having to search in a long list for the one I want.

lajones:
@natemcmaster Will do.

lajones:
@natemcmaster Have force pushed with latest commit - including extra test - and renamed that file as agreed.

natemcmaster:
:shipit:  (merge conflicts? probably due to r# cleanup)

PR #4279, opened 2016-01-11
[uwp] Add build error if runtime directives are not present

natemcmaster:
We stripped out automatic inference in #4197. UWP projects *will fail* without generating runtime directives. We can help users catch this early with msbuild

(If this idea get's approved, I will replace the URL in the error message to a fwlink)

cc @bricelam


bricelam:
LGTM. I'll let @rowanmiller sign-off on the experience and messages.

rowanmiller:
:shipit: from my perspective

rowanmiller:
Link to use in the message http://go.microsoft.com/fwlink/?LinkId=722449

PR #4282, opened 2016-01-11
Fix #4260 - bug in interaction between rownumberpaging and include

### LINKED ISSUE #4260: SQL2008 query bug with Include and skip
Changes made in 804458aa576723d8732f65605effde5b7fb7145a cause IncludeSqlServerTest.Include_with_skip to start to fail on SQL Server 2008:

```
System.InvalidOperationException : Sequence contains no matching element
   at System.Linq.Enumerable.Last[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.Data.Entity.Query.ExpressionVisitors.Internal.IncludeExpressionVisitor.
```

Full stack trace on our CI server

cc @anpete

natemcmaster:
cc @anpete

### DIFF
--- a/src/EntityFramework.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
+++ b/src/EntityFramework.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
@@ -207,7 +207,11 @@ namespace Microsoft.Data.Entity.Query.ExpressionVisitors.Internal
                 else
                 {
                     var principalTable
-                        = selectExpression.Tables.Last(t => t.QuerySource == querySource);
+                        = (selectExpression.Tables.Count == 1)
+                            && selectExpression.Tables.OfType<SelectExpression>().Any(s => s.Tables.Any(t => t.QuerySource == querySource))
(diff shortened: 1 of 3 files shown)


src/EntityFramework.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -207,7 +207,11 @@ yield return
                 else
                 {
                     var principalTable
-                        = selectExpression.Tables.Last(t => t.QuerySource == querySource);
+                        = (selectExpression.Tables.Count == 1)
```
anpete: Can we use SelectExpression.GetTableForQuerySource instead?

natemcmaster: Nope.

natemcmaster: Tried, but this method gets `FirstOrDefault` and we need `Last()`.

anpete:
:shipit:

PR #4292, opened 2016-01-13
Query: Relational: Initial support for DefaultIfEmpty operator.

anpete:
@maumar


src/EntityFramework.Relational/Query/Internal/RelationalResultOperatorHandler.cs
```diff
@@ -255,65 +255,66 @@ private static Expression HandleCount(HandlerContext handlerContext)
...
 
-                return (Expression)_transformClientExpressionMethodInfo
-                    .MakeGenericMethod(minExpression.Type)
-                    .Invoke(null, new object[] { handlerContext });
-            }
+            selectExpression.ClearTables();
 
-            return handlerContext.EvalOnClient();
-        }
+            var emptySelectExpression = new SelectExpression(selectExpression.QuerySqlGeneratorFactory, "empty");
```
mikary: Why not use `handlerContext.SelectExpressionFactory`?

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -57,6 +57,8 @@ public SelectExpression([NotNull] IQuerySqlGeneratorFactory querySqlGeneratorFac
             ? _projection[0].Type
             : base.Type;
 
+        public IQuerySqlGeneratorFactory QuerySqlGeneratorFactory => _querySqlGeneratorFactory;
```
mikary: No need to expose the service if we use the factory in `RelationalResultOperatorHandler`

src/EntityFramework.Relational/Query/Internal/RelationalResultOperatorHandler.cs
anpete: That's better, thanks.

src/EntityFramework.Relational/Query/Expressions/SelectExpression.cs
anpete: Yay.

mikary:
:shipit:

PR #4295, opened 2016-01-13
Add explicit dependency to system.net.security.

natemcmaster:
Workaround https://github.com/dotnet/corefx/issues/5359 on Linux


natemcmaster:
cc @bricelam. This is a one-liner

natemcmaster:
nvm. corefx fix is on its way.

PR #4296, opened 2016-01-13
[RevEng]: For SQL Server also use type aliases to determine data type

lajones:
Fix for #3843. Now we have the annotation approach in place use that approach and make using type aliases specific to the SQL Server provider. (But I'll mark as "Providers Beware" in case other providers (@ErikEJ?) would like to do similar things).

### DIFF
--- a/src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
+++ b/src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
@@ -49,6 +49,29 @@ namespace Microsoft.Data.Entity.Scaffolding
             return propertyBuilder;
         }
 
+        protected override Type GetTypeMapping([NotNull] ColumnModel column)
+        {
+            RelationalTypeMapping mapping = null;
+            if (column.DataType != null)
+            {
+                string underlyingDataType = null;
+                var typeAliases = column.Table.Database.SqlServer().TypeAliases;
+                if (typeAliases != null)
+                {
+                    typeAliases.TryGetValue(column.DataType, out underlyingDataType);
+                }
+
+                mapping = TypeMapper.FindMapping(underlyingDataType ?? column.DataType);
+            }
+
+            if (mapping?.ClrType == null)
+            {
+                return null;
+            }
+
+            return column.IsNullable ? mapping.ClrType.MakeNullable() : mapping.ClrType;
+        }
+
         protected override KeyBuilder VisitPrimaryKey([NotNull] EntityTypeBuilder builder, [NotNull] TableModel table)
         {
             var keyBuilder = base.VisitPrimaryKey(builder, table);
--- a/src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
+++ b/src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
@@ -26,11 +26,11 @@ namespace Microsoft.Data.Entity.Scaffolding
         internal const string SelfReferencingPrincipalEndNavigationNamePattern = "Inverse{0}";
 
         protected virtual ILogger Logger { get; }
+        protected virtual IRelationalTypeMapper TypeMapper { get; }
 
         private Dictionary<TableModel, CSharpUniqueNamer<ColumnModel>> _columnNamers;
         private readonly TableModel _nullTable = new TableModel();
         private CSharpUniqueNamer<TableModel> _tableNamer;
-        private readonly IRelationalTypeMapper _typeMapper;
         private readonly IDatabaseModelFactory _databaseModelFactory;
 
         public RelationalScaffoldingModelFactory(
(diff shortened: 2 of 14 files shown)


ErikEJ:
Thanks, @lajones - but user defined types do not exist in sql compact

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -264,18 +264,12 @@ protected virtual PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder builde
...
 
-            var mapping = column.DataType != null
-                ? _typeMapper.FindMapping(column.DataType)
-                : null;
-
-            if (mapping?.ClrType == null)
+            var clrType = FindClrType(column);
+            if (clrType == null)
             {
-                Logger.LogWarning(RelationalDesignStrings.CannotFindTypeMappingForColumn(column.DisplayName, column.DataType));
```
natemcmaster: Can we keep log warning here? Then we don't have to duplicate this in the override.

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -30,7 +30,7 @@ public class RelationalScaffoldingModelFactory : IScaffoldingModelFactory
         private Dictionary<TableModel, CSharpUniqueNamer<ColumnModel>> _columnNamers;
         private readonly TableModel _nullTable = new TableModel();
         private CSharpUniqueNamer<TableModel> _tableNamer;
-        private readonly IRelationalTypeMapper _typeMapper;
+        protected readonly IRelationalTypeMapper _typeMapper;
```
natemcmaster: `TypeMapper`

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -315,6 +309,23 @@ protected virtual PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder builde
             return property;
         }
 
+        protected virtual Type FindClrType([NotNull] ColumnModel column)
```
natemcmaster: `GetTypeMapping` ? Consistent w/ pattern we use on the other apis in this class

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
```diff
@@ -49,6 +49,31 @@ protected override PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder build
             return propertyBuilder;
         }
 
+        protected override Type FindClrType([NotNull] ColumnModel column)
+        {
+            var mapping = column.DataType != null
+                ? _typeMapper.FindMapping(column.DataType)
```
natemcmaster: This calls `_typeMapper.FindMapping` twice. Can we just get the type alias first and then ask for mapping once?

natemcmaster:
Other than comments above, LGTM

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
lajones: Is this a R# thing?

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
lajones: OK. Will do.

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
natemcmaster: R# catches this. Everywhere else in our project we use protected properties and private fields. So this should probably be `protected IRelationalTypeMapper TypeMapper { get; }`

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
lajones: Yes - will do.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
lajones: Sure.

src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs
lajones: Yep - have updated it to match `Logger`.

lajones:
@natemcmaster All comments addressed in 2nd commit. I'll go ahead and check in.

natemcmaster:
:shipit:

lajones:
Checked in with commit 701f32e.

PR #4303, opened 2016-01-14
Remove GraphBehavior enum and track all reachable entities

ajcvickers:
Issue #4424

This makes the behavior for Add/Attach/Update be to track all reachable entities, as in EF6. Single object operations can be done with the EntityEntry API. In the future we made add a context-wide flag to modify this behavior.

Also, make Remove still be single object, but do attachment of reachable entities such that cascade deletes will be processed correctly on SaveChanges.


ajcvickers:
Ping. This has been out for review for almost two weeks now.

ajcvickers:
@divega Updated with changes to graph behavior from #4424

divega:
:shipit: 

Nice to see things getting simpler.

PR #4322, opened 2016-01-15
Throw when DbContext.Set<T>() is called where T is not in the model

### LINKED ISSUE #4301: Prevent DbContext.Set<T>() from creating sets for types not in the model
Currently we create sets for any type... but they fail ungracefully if used to do anything (NullReferenceException etc.). We should just throw in the Set<T> method if T is not in the model.

natemcmaster:
Fix #4301 

cc @rowanmiller

### DIFF
--- a/src/EntityFramework.Core/Properties/CoreStrings.resx
+++ b/src/EntityFramework.Core/Properties/CoreStrings.resx
@@ -528,4 +528,7 @@
   <data name="ConcurrentMethodInvocation" xml:space="preserve">
     <value>A second operation started on this context before a previous operation completed. Any instance members are not guaranteed to be thread safe.</value>
   </data>
+  <data name="InvalidSetType" xml:space="preserve">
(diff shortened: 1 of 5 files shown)


src/EntityFramework.Core/Properties/CoreStrings.resx
```diff
@@ -528,4 +528,7 @@
   <data name="ConcurrentMethodInvocation" xml:space="preserve">
     <value>A second operation started on this context before a previous operation completed. Any instance members are not guaranteed to be thread safe.</value>
   </data>
+  <data name="InvalidSetType" xml:space="preserve">
+    <value>Cannot create DbSet for type '{typeName}' because this type was not configured in the model.</value>
```
rowanmiller: Small tweak to exception message...

> Cannot create a DbSet for '{typeName}' because this type is not included in the model for the context.

natemcmaster: I updated the message.

anpete:
:shipit:

Issue #4320, opened 2016-01-15
EF does not install and build on UWP


natemcmaster:
Since dropping the "netcore50" TFM, we have not been able to install and build EF on UWP. Tracking progress on external blocking issues here.

- [x] EntityFramework.Sqlite - Fixed with #4343
- [ ] EntityFramework.Commands

natemcmaster:
Blocked on https://github.com/dotnet/corefx/issues/5456

PR #4338, opened 2016-01-20
Fix .NET Native queries with shadow properties

### LINKED ISSUE #3604: Queries with shadow properties fail on .NET Native
Any query that uses `EF.Property(e, "SomeProperty")` fails on UWP, .NET Native with this error:

`System.InvalidOperationException: The EF.Property<T> method may only be used within LINQ queries.`

natemcmaster:
Depends on #4320 

Fix #3604

cc @anpete

### DIFF
--- a/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
+++ b/src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
@@ -32,8 +32,14 @@ namespace Microsoft.Data.Entity.Query
         public static readonly ParameterExpression QueryContextParameter
             = Expression.Parameter(typeof(QueryContext), "queryContext");
 
-        public static readonly MethodInfo PropertyMethodInfo
-            = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
+        private static readonly string EFTypeName = typeof(EF).FullName;
+
+        public static bool IsPropertyMethod([CanBeNull] MethodInfo methodInfo) =>
+            methodInfo != null
+            && methodInfo.IsGenericMethod
+            // string comparison because MethodInfo.Equals is not .NET Native-safe
+            && methodInfo.Name == nameof(EF.Property)
+            && methodInfo.DeclaringType.FullName == EFTypeName;
 
         private readonly IQueryOptimizer _queryOptimizer;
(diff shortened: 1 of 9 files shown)


src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -32,9 +32,22 @@ public abstract class EntityQueryModelVisitor : QueryModelVisitorBase
...
-        public static readonly MethodInfo PropertyMethodInfo
+        private static readonly MethodInfo PropertyMethodInfo
             = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
 
+        private const string EFTypeName = "Microsoft.Data.Entity.EF";
+
+        public static bool IsPropertyMethod([CanBeNull] MethodInfo methodInfo) =>
+            methodInfo != null
+            && methodInfo.IsGenericMethod
+#if NETCORE50
```
anpete: I would be OK with just the NETCORE50 approach if we wanted to simplify further.

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
```diff
@@ -32,9 +32,22 @@ public abstract class EntityQueryModelVisitor : QueryModelVisitorBase
         public static readonly ParameterExpression QueryContextParameter
             = Expression.Parameter(typeof(QueryContext), "queryContext");
 
-        public static readonly MethodInfo PropertyMethodInfo
+        private static readonly MethodInfo PropertyMethodInfo
             = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
 
+        private const string EFTypeName = "Microsoft.Data.Entity.EF";
```
anpete: Consider typeof(EF).FullName

anpete:
:shipit:

src/EntityFramework.Core/Query/EntityQueryModelVisitor.cs
natemcmaster: So, not worried about the perf difference between string comparison and reference comparisons?

anpete: No, it should be negligible in the context of query compilation.

PR #4343, opened 2016-01-20
Re-add netcore50

natemcmaster:
Add netcore50 TFM to libraries required to run SQLite provider on UWP.

TODO: support commands project on UWP.

Partial fix for #4320 

cc @bricelam


bricelam:
:shipit:

Issue #4312, opened 2016-01-14
RevEng: generated props .HasMaxLength(x).HasColumnType("varchar") create varchar(1) columns in DB


umu24013:
The dnx command with the **-a** option
*dnx ef dbcontext scaffold &lt;connection string&gt; EntityFramework.MicrosoftSqlServer -a -c &lt;context name&gt; -o &lt;output dir&gt;*

generates incomplete entity properties for the varchar data type in the OnModelCreating method:

for **nvarchar(*x*)** it generates **.HasMaxLength(*x*)**
but for **varchar(*x*)** it generates **.HasMaxLength(*x*).HasColumnType("varchar")**
where **x** is data length

Then, when I call Database.EnsureCreated() from the generated dbcontext, it creates varchar(**1**) columns in the DB rather then varchar(*x*).

As a result, the SQL Server throws en exception "String or binary data would be truncated" when trying to write to those castrated columns.
![varchar 1](https://cloud.githubusercontent.com/assets/16705608/12332556/2a9f3a18-bb00-11e5-9a92-192547af2765.png)

lajones:
Thanks for reporting this @umu24013. I'll look into it.

rowanmiller:
Once we have `IsUnicode` API (https://github.com/aspnet/EntityFramework/issues/3420) the code should be ```.HasMaxLength(x).IsUnicode(false)```. In the meantime it should just be ```HasColumnType("varchar(x)"). The HasMaxLength call isn't needed because the length is specified in the datatype... but if it's less work just to leave it in, then that is fine (we'll want it when we swap to `IsUnicode` anyway).

lajones:
Fix is checked in with #4348, commit 443c52b.

Details: using both `HasColumnType()` and `HasMaxLength()` at the same time doesn't work. Because the column type contains (either implicitly or explicitly) a max length we just ignore the `HasMaxLength()` call and use the column type defined by `HasColumnType()`. When scaffolding generated e.g. `.HasColumnType("varchar").HasMaxLength(100)` it was expecting that migrations would treat that as an underlying data type of `varchar(100)`. What actually happened, as mentioned at the top of this issue, was that the `HasMaxLength()` call was ignored and migrations assumed a data type of `varchar` which SQL Server treats as `varchar(1)`.

[...]

PR #4348, opened 2016-01-20
[RevEng] Output the HasColumnType() and HasMaxLength() APIs correctly

lajones:
Fix for #4312.

For SQL Server this means for `nvarchar` columns (or synonyms) we use `HasMaxLength()` and not `HasColumnType()`, for other string types we use `HasColumnType()` and not `HasMaxLength()`. Similarly for `varbinary` (or synonyms) use `HasMaxLength()` and not `HasColumnType()`, but for other binary column types use `HasColumnType()` and not `HasMaxLength()`.


src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
```diff
@@ -18,6 +19,11 @@ public class SqlServerScaffoldingModelFactory : RelationalScaffoldingModelFactor
     {
         private const int DefaultTimeTimePrecision = 7;
 
+        private static readonly ISet<string> _dataTypesForbiddingLength =
```
natemcmaster: Anyway we can capture this in the SqlServer type mapper? I thought it was already designed to handle lengths.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
```diff
@@ -144,16 +145,64 @@ private PropertyBuilder VisitTypeMapping(PropertyBuilder propertyBuilder, Column
...
+                    }
+                }
             }
 
             return propertyBuilder;
         }
 
+        private bool IsTypeAlias(ColumnModel column)
+        {
+            var typeAliases = column.Table.Database.SqlServer().TypeAliases;
```
natemcmaster: Possible to simplify?
`=> column.Table.Database.SqlServer().TypeAliases?.ContainsKey(column.DataType) == true;`

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
```diff
@@ -144,16 +145,64 @@ private PropertyBuilder VisitTypeMapping(PropertyBuilder propertyBuilder, Column
...
+                        // what we did above).
+                        propertyBuilder.Metadata.SetMaxLength(null);
+                    }
+                }
             }
 
             return propertyBuilder;
         }
 
+        private bool IsTypeAlias(ColumnModel column)
```
natemcmaster: `HasTypeAlias`?

natemcmaster:
Seems like we could make improvements to reuse type mapping API. Perhaps we need new API, such as `IRelationalTypeMapper.FindMapping(string typeName, int length)`. We already have some API for handling sized type mappings (`RelationalSizedTypeMapping`)

Thoughts on this @ajcvickers ?

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
lajones: Sure.

src/EntityFramework.MicrosoftSqlServer.Design/SqlServerScaffoldingModelFactory.cs
lajones: OK. Will do.

lajones:
@natemcmaster @ajcvickers Nit: It would definitely have to be `int? length`.

The problem is that different data types interpret having a null max length differently. For `image`, `text` etc it is legal (in fact expected) and means "max length doesn't apply to this data type", for `nvarchar`, `varbinary` etc it means "equivalent to `nvarchar(max)`, `varbinary(max)`" etc. For other types it would make no sense e.g. `nchar`, `char` and would be an error.

natemcmaster:
>  It would definitely have to be int? length.

Or an overload.

> The problem is that different data types interpret having a null max length differently....

To clarify, my point was that this logic could be placed in `SqlServerTypeMapper` instead of `SqlServerScaffoldModelFactory`.

lajones:
@natemcmaster Sort of. This is not just about type mapping. It would be perfectly legal for us to output e.g. `HasColumnType("nvarchar(100)")` but we have chosen instead to write out `HasMaxLength(100)` and rely on the fact that the code we produce will define the CLR type for that column as a `string` and the default mapping for a string is back to `nvarchar`.

So in the end we need 2 pieces of information out: 1) should we use `HasColumnType()` or `HasMaxLength()` and 2) if `HasColumnType()`, what is the full definition of the column type we want to use, including any length qualifications.

I'll try to put all of that in `SqlServerTypeMapper` and see what it looks like.

lajones:
@ajcvickers Have updated as suggested above - could you give the final sign-off since Nate is away?

lajones:
Removing "Needs Review" while I rebase this on the renamed codebase and get it working again. Will put it back on when ready for review again.

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
```diff
@@ -36,6 +38,11 @@ public class SqlServerTypeMapper : RelationalTypeMapper
         private readonly RelationalTypeMapping _decimal = new RelationalTypeMapping("decimal(18, 2)", typeof(decimal));
         private readonly RelationalTypeMapping _time = new RelationalTypeMapping("time", typeof(TimeSpan));
 
+        private static readonly ISet<string> _dataTypesForbiddingMaxLength =
+            new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "image", "ntext", "text" };
```
jdlundquist: What about rowversion and uniqueidentifier - would these map to byte[] and string CLR types, respectively, and need to be excluded?

Also, I don't think date, datetime, or smalldatetime allow max length (at least per SQL Server, not sure about `HasMaxLength()`), but those would be excluded implicitly by CLR type in `MaxLengthQualifiedDataType`

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
```diff
@@ -117,6 +124,66 @@ public SqlServerTypeMapper()
...
+        /// <param name="maxLength">
+        ///   the max length of the same column as defined by SQL Server metadata
+        /// </param>
+        /// <returns>
+        ///   a tuple of the max-length-qualified SQL Server type name, e.g. nvarchar(max),
+        ///   which can be used by HasColumnType() and the max length that should be
+        ///   defined on that same property through HasMaxLength()
+        /// </returns>
+        public virtual Tuple<string, int?> MaxLengthQualifiedDataType(
+            [NotNull] string unqualifiedTypeName, int? maxLength)
```
jdlundquist: What about decimal/numeric that take a precision and scale vs. a single max length? anything to consider for that or outside the scope of #4312 ?

src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs
lajones: Please don't review just yet - I'm in the process of updating this and will add the "Needs Review" tag back when it's ready. Nonetheless I take your point and will check those scenarios.

lajones:
I tried to put the code in `SqlServerTypeMapper` but what we are doing in scaffolding right there is overriding already assigned column types and max lengths. And sometimes it's important _not_ to override the existing values - so it's not really a generic "output column type and max length for this column" method which would be more generally useful throughout the stack - it's quite specific to what scaffolding needs. I agree such a method would be useful and I'll raise another issue to suggest looking into `SqlServerTypeMapper` and redesigning so that there is such a method. But for this issue, after discussion with @natemcmaster, I've moved most of the code back into `SqlServerScaffoldingModelFactory`. I did have to leave in one change to `SqlServerTypeMapper` as a `binary` data type should not be mapped to `_varbinary` otherwise it thinks its DefaultTypeName is `varbinary` and we end up outputting `HasMaxLength()` instead of `HasColumnType()`.

That said there are real synonyms for some types e.g. `varbinary` has the synonym `binary varying`. So I've updated the tests so we now test DDL using all data types including synonyms and check that we get the correct `HasColumnType()` and `HasMaxLength()` fluent API as a result of that.

natemcmaster:
LGTM. :shipit:

lajones:
Checked in with commit 443c52b.

lajones:
Note: opened issue #4425 to track trying to make the `SqlServerTypeMapper` APIs more useful as discussed with @natemcmaster.

Issue #3890, opened 2015-11-25
dbContext.<Entity>.Update() is creating new records


DiegoZoracKy:
When i do something like:
```javascript
var e = new Entity();
e.Id = 100;
e.Name = "Name 100";
dbContext.Entities.Update(e);
dbContext.SaveChanges();
```
a new record is created.

Is this the correct behaviour?

rowanmiller:
@DiegoZoracKy it should attempt to update an existing record, not insert a new one. Can you share a full code listing that we can use to reproduce the issue? (i.e. the code for `Entity` and your context)

DiegoZoracKy:
@rowanmiller i've created something simple to show the case. See:

```javascript
using Microsoft.AspNet.Mvc;
using System.Linq;
using Microsoft.Data.Entity;
using Newtonsoft.Json.Linq;
		
public class TestUpdate
{
	public int Id { get; set; }
	public string Name { get; set; }
}

public class MyContext : DbContext
{
	public DbSet<TestUpdate> TestUpdates { get; set; }		
}			

[Route("api/[controller]")]
public class TestController : Controller
{
	private MyContext dbContext;

	public TestController(MyContext context)
	{
		dbContext = context;
	}

	[HttpGet]
	public IActionResult Get()
	{
		var e = new TestUpdate();
		e.Name = "Name 100";
		dbContext.TestUpdates.Update(e);
		dbContext.SaveChanges();

		return new ObjectResult(dbContext.TestUpdates.AsNoTracking().ToList());
	}
}	
```

Everytime you hit api/test (GET) a new record will be saved.

DiegoZoracKy:
@rowanmiller  i've found what it seems to be another weird behavior on `Update` method. Let's suppose that in my previous example, `AnotherClass` exists on the project and it makes a reference to `TestUpdate`, like:

```javascript
	public class AnotherClass
	{
    	public int Id { get; set; }
    	public virtual List<TestUpdate> TestUpdate { get; set; }
	}
```

Thinking about SQLServer, this will create a column `AnotherClassId` on the `TestUpdate` Table, with a reference to records on `AnotherClass`, right?

The problem is that when i do `.Update` on `TestUpdate`, this reference is being set to `NULL`.

rowanmiller:
@DiegoZoracKy regarding the first issue you reported...

[...]

PR #4351, opened 2016-01-20
Entity given to Attach/Update does not use key to determine state

ajcvickers:
Issue #3890

When a graph of entities is given to Attach or Update, the state of those entities is set to Unchanged/Modified unless the entity is using key value generation and the key value is not set, in which case the state is set to Added. But for the actual entity passed to these methods (as opposed to additional entities discovered in the graph) we decided that the state should be the requested state regardless of whether or not the key value is set.


anpete:
:shipit:

ajcvickers:
Merged

Issue #4020, opened 2015-12-09
DbContext.X.Where().Include(x => x.Y) may load Y as added entity


LonDat:
Hi,
I have 2 classes (X and Y) with **INotifyPropertyChanged**
Somethink like:
```C#
public class X : INotifyPropertyChanged
{
    public Y YRef { 
            get { return ...; } 
            set { ... = value; NotifyPropertyChange(...); } // NotifyPropertyChange is important!
    }
    // some PK + properties

}
public class Y : INotifyPropertyChanged
{
    // some PK + properties
}
```
When I execute:
```C#
dbContext.Xs.Where<X>(x => x.ID == 1).Include(x => x.YRef).ToArray();
```
everything is OK...I have 2 entries in unmodified state...BUT:
```C#
dbContext.Xs.Single<X>(x => x.ID == 1); // it's cache instance in entry
dbContext.Xs.Where<X>(x => x.ID == 1).Include(x => x.YRef).ToArray();
```
Instance (type Y) in YRef property has Added state.

I mean, that YRef call PropertyChanged and StateManager start tracking entity...and because entry is Detached...state is set to Added.

Is it understandable? Thx.
PS: sry...my English is not perfect :)

divega:
@ajcvickers did you see this? Seems interesting.

LonDat:
Any news/progress? :)

PR #4359, opened 2016-01-21
Ensure notification entities brought in by query are tracked as Uncha…

ajcvickers:
…nged

Issue #4020

The issue is that when query sets the navigation property this is detected by the change tracker and the new entity is marked as Added. There doesn't seem to be a clean way to prevent this, although it possibly could be done by looking at the stack or by setting some kind of thread-local flag to tell the change tracker that query is making the changes. Instead, when query is done and tells EF to start tracking the entity we now explicitly set the state to Unchanged even if the entity is already being tracked.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/QueryContextFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/QueryContextFactory.cs
@@ -12,17 +12,22 @@ namespace Microsoft.EntityFrameworkCore.Query
     {
         protected QueryContextFactory(
             [NotNull] IStateManager stateManager,
-            [NotNull] IConcurrencyDetector concurrencyDetector)
+            [NotNull] IConcurrencyDetector concurrencyDetector,
+            [NotNull] IChangeDetector changeDetector)
         {
             Check.NotNull(stateManager, nameof(stateManager));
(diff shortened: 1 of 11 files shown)


anpete:
@ajcvickers Could we just disable processing of change notifications during nav prop load? E.g. We have access to StateManager.Notify from QueryBuffer.LoadNavigationProperties.

ajcvickers:
@anpete Updated

src/Microsoft.EntityFrameworkCore/Query/QueryContextFactory.cs
```diff
@@ -10,19 +10,24 @@ namespace Microsoft.EntityFrameworkCore.Query
...
+            [NotNull] IConcurrencyDetector concurrencyDetector,
+            [NotNull] IChangeDetector changeDetector)
         {
             Check.NotNull(stateManager, nameof(stateManager));
             Check.NotNull(concurrencyDetector, nameof(concurrencyDetector));
+            Check.NotNull(changeDetector, nameof(changeDetector));
 
             StateManager = stateManager;
             ConcurrencyDetector = concurrencyDetector;
+            _changeDetector = changeDetector;
```
anpete: Make this a protected virtual property so that overrides of CreateQueryBuffer can access it.

anpete:
Nice. :shipit:

PR #4378, opened 2016-01-22
[TINY] Link to SQLite limitation docs from exceptions

### LINKED ISSUE #4103: Link to docs about SQLite limitations when a migration operation is not supported
Need to write the doc and then update the exception message in the SQL generator to forward to it via a FWLink

rowanmiller:
Resolves #4103


lajones:
:shipit:

PR #4379, opened 2016-01-22
[SMALL] Improve HasKey on derived type exception message

### LINKED ISSUE #3822: Improve exception message when we can't map a property
My entity includes a property that only has a getter, e.g. something like this:

    public string FullName
    {
        get { return FirstName + " " + LastName; }
    }

Previously, this was automatically ignored by EF when building the model. But now, I receive the following error:

> No backing field could be discovered for property 'Example.User.FullName' and the property does not have a setter. Either use a backing field name that can be matched by convention, annotate the property with a backing field, or define a property setter.

I had to add an explicit `entity.Ignore(x => x.FullName)` to the model builder, in order to fix this. But before RC1, this was just working fine without having to explcitly ignore properties. When and why was this changed?

### LINKED ISSUE #3863: HasKey() throws confusing exception about redefining key if no inheritance was intended
Work item is to add info about how you an get into this case.

<hr />

From https://github.com/aspnet/EntityFramework/issues/3727#issuecomment-159079828 by @miroslavsiska: the following throws 

`InvalidOperationException: The derived type 'CashDesk.Models.AllergenQuantityMaterial' cannot have keys other than those declared on the root type.` 

Note that this is still a negative scenario, e.g. there cannot be a navigation property of the base type if the base type does not define the key. But the exception isn't very helpful if the type wasn't intended as an entity type in the first place.

Repro code:

``` C#
public class AllergenQuantity
{              
    public Guid UnitId { get; set; }
    public virtual Unit Unit { get; set; }         

    [Display(Name = "Množství obsahu")]
    public decimal Quantity { get; set; }
}


 public class Unit 
{
    public Unit()
    {
        this.Id = Guid.NewGuid();
    }
    public virtual ICollection<AllergenQuantity> AllergenQuantity { get; set; }

    [Key]
    public Guid Id { get; set; }     

    [Display(Name = "Jednotka")]

    [Display(Name = "Typ Jednotky")]
    public string UnitType { get; set; }

    [Display(Name = "Použité v kulturách")]
    public string Cultures { get; set; }
}


  public class TenantDbContext : IdentityDbContext<TenantUser, TenantRole, Guid>
{
    private string _connectionString { get; set; }

    public TenantDbContext(DbContextOptions<TenantDbContext> options) : base(options) 
    {
        this._connectionString = "CON STRING";
    }

    public TenantDbContext()       
    {
    }   
    public TenantDbContext(string ConnectionString)
    {
        this._connectionString = ConnectionString;
    }
    public static TenantDbContext Create(string ConnectionString)
    {
        return new TenantDbContext(ConnectionString);
    }

    { 
        optionsBuilder.UseSqlServer(_connectionString);  
    }

    //  BASE
    public DbSet<Unit> Unit { get; set; }

    //  Stock
    public DbSet<Allergen> Allergen { get; set; }  
    public DbSet<AllergenQuantityMaterial> AllergenQuantityMaterial { get; set; }        
        //Inherited from AllergenQuantity
    public DbSet<Material> Material { get; set; }

    {      

        // Throws
        builder.Entity<AllergenQuantityMaterial>().HasKey(l => new { l.MaterialId, l.AllergenId });  
    }
}
```

rowanmiller:
Resolve #3863

A common cause of this exception is unintentionally including a base
type in the model. Adding info about how to stop this happening to the
exception message.

**Old Message:** The derived type 'Address' cannot have keys other than
those declared on the root type 'AddressBase'.

**New Message:** A key cannot be configured on 'Address' because it is a
derived type. The key must be configured on the root type 'AddressBase'.
If you did not intend for 'AddressBase' to be included in the model,
ensure that it is not included in a DbSet property on your context,
referenced in a configuration call to ModelBuilder, or referenced from a
navigation property on a type that is included in the model.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.Designer.cs
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.Designer.cs
@@ -653,7 +653,7 @@ namespace Microsoft.EntityFrameworkCore.Internal
         }
 
         /// <summary>
-        /// The derived type '{derivedType}' cannot have keys other than those declared on the root type '{rootType}'.
+        /// A key cannot be configured on '{derivedType}' because it is a derived type. The key must be configured on the root type '{rootType}'. If you did not intend for '{rootType}' to be included in the model, ensure that it is not included in a DbSet property on your context, referenced in a configuration call to ModelBuilder, or referenced from a navigation property on a type that is included in the model.
         /// </summary>
         public static string DerivedEntityTypeKey([CanBeNull] object derivedType, [CanBeNull] object rootType)
         {
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
@@ -358,7 +358,7 @@
     <value>The '{factory}' cannot create a value generator for property '{property}' on entity type '{entityType}'. Only integer properties are supported.</value>
   </data>
   <data name="DerivedEntityTypeKey" xml:space="preserve">
-    <value>The derived type '{derivedType}' cannot have keys other than those declared on the root type '{rootType}'.</value>
+    <value>A key cannot be configured on '{derivedType}' because it is a derived type. The key must be configured on the root type '{rootType}'. If you did not intend for '{rootType}' to be included in the model, ensure that it is not included in a DbSet property on your context, referenced in a configuration call to ModelBuilder, or referenced from a navigation property on a type that is included in the model.</value>
   </data>
   <data name="CircularInheritance" xml:space="preserve">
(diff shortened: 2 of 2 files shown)


rowanmiller:
@divega can you take a look at the wording

divega:
IMO the part about "referenced in a configuration call to ModelBuilder" isn't very clear but I cannot think of a better way other than maybe mentioning the Entity method.

I think this is a big improvement so if you cannot think of anything else :shipit:

ajcvickers:
Looks good to me.

rowanmiller:
Yeah I agree it is not super clear... but I also could not come up with anything better :cry:

rowanmiller:
Merged https://github.com/aspnet/EntityFramework/commit/7fc5cd8c9bc6469de831f25ac48b99ff201d29a5

PR #4380, opened 2016-01-22
[SMALL] Improve exception when no backing field found

### LINKED ISSUE #3822: Improve exception message when we can't map a property

rowanmiller:
Providing a more helpful exception message if you include a "getter
only" property in the model and no backing field is found by convention.

Resolves #3822 

Sample Message

> The property 'ConsoleApp1.Customer.Name' does not have a setter and no backing field was found by convention. Either remove the property from the model, add a property setter, or configure a backing field (see http://go.microsoft.com/fwlink/?LinkId=723277).

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.Designer.cs
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.Designer.cs
@@ -173,7 +173,7 @@ namespace Microsoft.EntityFrameworkCore.Internal
         }
 
         /// <summary>
-        /// No backing field could be discovered for property '{entityType}.{property}' and the property does not have a setter. Either use a backing field name that can be matched by convention, annotate the property with a backing field, or define a property setter.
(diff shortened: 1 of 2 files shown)


divega:
:shipit:

rowanmiller:
Merged https://github.com/aspnet/EntityFramework/commit/9263a1df7ffede5b20c03528e1bc9ba04d803191

Issue #3925, opened 2015-11-30
Commands: .NET Core CLI Implementation


bricelam:
DNX commands will be going away with the transition to [.NET Core CLI](https://github.com/dotnet/cli). We'll need to create a .NET Core CLI tools package instead. Like the NuGet PMC commands, any version of the command should be able to work with any version of the framework.

Known Issues
------------

- [x] System.DllNotFoundException: Unable to load DLL 'sni.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E) (#4953)
- [x] Unable to load DLL 'sqlite3': The specified module could not be found. (Exception from HRESULT: 0x8007007E) (dotnet/cli#2289)
- [x] Could not load file or assembly 'System.Data.SqlClient, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified. (dotnet/cli#2289)
- [x] Support netcoreapp1.0 apps
- [x] Hosting environment not setup correctly. https://github.com/aspnet/EntityFramework/issues/5065
- [x] Newtonsoft.Json 7.0.0 missing at runtime https://github.com/NuGet/Home/issues/2462
- [x] Newtonsoft.Json 8.0.0 could not be loaded at runtime https://github.com/dotnet/cli/issues/2514. Workaround in #5082 
- [x] ~~Tools using ConfigurationBuilder without setting basepath cannot find config files.~~ won't fix. User's need to set base path to IApplicationEnvironment.ApplicationBasePath.
- [x] Need to change to nested-dispatch model because `AssemblyLoadContext` does not provide the necessary isolation https://github.com/dotnet/cli/issues/2514
- [x] Cannot use dotnet-ef on a netstandard library or standalone app with nested-dispatch https://github.com/dotnet/cli/issues/2645

Usage
-----
Your `project.json` should look something like this.

```JSON
{
    "compilationOptions": {
        "emitEntryPoint": true
    },
    "tools": {
        "Microsoft.EntityFrameworkCore.Tools": { 
            "version": "1.0.0-*",
            "imports": "portable-net452+win81"
         }
    },
    "dependencies": {
        "Microsoft.EntityFrameworkCore.Sqlite": "1.0.0-*",
        "Microsoft.EntityFrameworkCore.Sqlite.Design": { "version": "1.0.0-*", "type": "build" },

[...]

PR #4381, opened 2016-01-22
Add (non-working) .NET Core CLI commands

bricelam:
Part of #3925 

This commit puts a stake in the ground. I'm still getting assembly resolution errors locally that may or may not be related to CoreCLR/CLI issues. Most negative cases aren't handled very well. Etc. I'll continue to test and iterate on this, but I think it's at a good point to merge.

For reference, here's [ye olde DNX Program.cs](https://github.com/aspnet/EntityFramework/blob/16dc2107105930e4a8b3bea4ac0679269d22d888/src/EntityFramework.Commands/Commands/Program.cs).

### DIFF
--- a/src/dotnet-ef/MigrationsCommand.cs
+++ b/src/dotnet-ef/MigrationsCommand.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.EntityFrameworkCore.Commands
+{
+    public class MigrationsCommand
+    {
+        public static void Configure([NotNull] CommandLineApplication command)
+        {
+            command.Description = "Commands to manage your migrations";
+
+            command.HelpOption("-h|--help");
+
+            command.Command("add", MigrationsAddCommand.Configure);
+            command.Command("list", MigrationsListCommand.Configure);
+            command.Command("remove", MigrationsRemoveCommand.Configure);
+            command.Command("script", MigrationsScriptCommand.Configure);
+
+            command.OnExecute(() => command.ShowHelp());
+        }
+    }
+}
--- a/src/dotnet-ef/project.json
+++ b/src/dotnet-ef/project.json
@@ -1,0 +1,24 @@
+{
+  "version": "1.0.0-*",
+  "description": "Entity Framework Core Commands for .NET Core CLI.",
+  "compilationOptions": {
+    "warningsAsErrors": true,
+    "emitEntryPoint": true
+  },
+  "compile": "..\\Shared\\*.cs",
+  "dependencies": {
+    "Microsoft.DotNet.Cli.Utils": "1.0.0-*",
+    "Microsoft.DotNet.ProjectModel.Loader": "1.0.0-*",
+    "Microsoft.EntityFrameworkCore.Commands": "1.0.0-*",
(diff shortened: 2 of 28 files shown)


NuGet.config
```diff
@@ -2,6 +2,8 @@
 <configuration>
   <packageSources>
     <add key="AspNetVNext" value="https://www.myget.org/F/aspnetcidev/api/v3/index.json" />
+    <add key="dotnet-cli" value="https://www.myget.org/F/dotnet-cli/api/v3/index.json" />
+    <add key="nugetbuild" value="https://www.myget.org/F/nugetbuild/api/v3/index.json" />
```
bricelam: @Eilon @pranavkm @victorhurdugaci It's a chicken-and-egg problem, but are we going to add the CLI packages to our feed?

victorhurdugaci: For the xunit runner I actually manually pushed some CLI packages to our feed. If you need something from there, I think we should do the same thing as we do for CLR: gated-clone with some sort of verification.

pranavkm: Yup! They had some issues with their packaging which was why I was avoiding pulling their packages in, but I could set up the CoreCLR configuration to clone the CLI packages.

victorhurdugaci: There's another problem. Since we don't always have the latest CLR, the packages in the dotnet-cli feed might require never versions.

smitpatel:
:+1:

src/dotnet-ef/project.json
```diff
@@ -0,0 +1,23 @@
+{
+  "version": "1.0.0-*",
+  "description": "Entity Framework Core Commands for .NET Core CLI.",
+  "repository": {
```
victorhurdugaci: Not used anymore

NuGet.config
bricelam: I'm not sure mismatched versions will be an issue. The common utility libraries are rather self-contained, so using different versions in different tools should be fine.

src/dotnet-ef/MigrationsCommand.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
```
victorhurdugaci: Required?

NuGet.config
victorhurdugaci: My worry is that you'll not be able to restore because the packages on the CLI feed reference a newer version of the CLR which is not part of their feed and we only have the older version on aspnetcidev/volatile

lajones:
:shipit: As you mention more error handling would be good, but this is a great first step.

NuGet.config
bricelam: Every tool runs in its own CoreCLR process, so different versions of the same assembly are perfectly fine.

src/dotnet-ef/MigrationsCommand.cs
bricelam: Yes, to satisfy EF's [API consistency tests](https://github.com/aspnet/EntityFramework/blob/dev/test/Microsoft.EntityFrameworkCore.Tests/ApiConsistencyTestBase.cs)

NuGet.config
bricelam: So what's the course of action here? I assume other tools are also starting to need these...

src/dotnet-ef/project.json
bricelam: It looks like you added these in f2a1c5dbea6446f7acc5e40cb68750a660b0a694. We're free to remove them now?

PR #4397, opened 2016-01-25
Only discover properties with public getter & any setter

### LINKED ISSUE #4314: Only discover properties with public getter and public/private/etc. setter

smitpatel:
Resolves #4314


ajcvickers:
:shipit:

Issue #1940, opened 2015-03-27
RevEng: Decide where we store connection string


rowanmiller:
Currently we just write it to code. We are really trying to help folks write secure apps and this is an ant-pattern so we need to work out what to do. This will be complex because the ideal solution is going to be different for different project types.

lajones:
We discussed this and decided we will continue to output the connection string in OnConfiguring(). But in addition we will output a #warning indicating that the connection string may contain a password and with a comment containing a link to a web page which explains what you can do to fix that. Moving to @rowanmiller to decide the wording and the link.

rowanmiller:
Here is my proposed message:

> In order to protect potentially sensitive information in your connection string, you should move it out of source code. See http://go.microsoft.com/fwlink/?LinkId=723263 for guidance on storing connection strings.

rowanmiller:
cc @divega above proposal look ok to you?

divega:
Looks good.

rowanmiller:
@lajones can you make this change in the generated code

lajones:
Fixed by PR #4398. Commit d9b5da1.

PR #4398, opened 2016-01-26
[RevEng] Compiler warning re potential sensitive information in connection string

lajones:
@rowanmiller Fix for #1940.


src/Microsoft.EntityFrameworkCore.Commands/Scaffolding/Internal/DbContextWriter.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
 {
     public class DbContextWriter
     {
+        public const string SensitiveInformationFwLink = "http://go.microsoft.com/fwlink/?LinkId=723263";
```
rowanmiller: The URL should just be part of the resource and not formatted in (allows localizing content later if we do that).

rowanmiller:
:shipit: (once the link is combined into the resource string)

lajones:
Checked in with the change requested. Commit d9b5da1.

Issue #3722, opened 2015-11-11
Subquery.Where(Expression<Func<Entity, bool>>) throws exception


Tasteful:
When adding a subquery from expression an exception it thrown as below.
```
{
"message":"An error has occurred.",
"exceptionMessage":"Could not parse expression 'value(Microsoft.Data.Entity.Query.EntityQueryable`1[ProductEntity+FieldDataEntity]).Any(__expr_0)': The given arguments did not match the expected arguments: Object of type 'System.Linq.Expressions.TypedParameterExpression' cannot be converted to type 'System.Linq.Expressions.LambdaExpression'.",
"exceptionType":"System.NotSupportedException",
"stackTrace":"   at Remotion.Linq.Parsing.Structure.MethodCallExpressionParser.CreateExpressionNode(Type nodeType, MethodCallExpressionParseInfo parseInfo, Object[] additionalConstructorParameters)
   at Remotion.Linq.Parsing.Structure.MethodCallExpressionParser.Parse(String associatedIdentifier, IExpressionNode source, IEnumerable`1 arguments, MethodCallExpression expressionToParse)
   at Remotion.Linq.Parsing.Structure.ExpressionTreeParser.ParseMethodCallExpression(MethodCallExpression methodCallExpression, String associatedIdentifier)
   at Remotion.Linq.Parsing.Structure.QueryParser.GetParsedQuery(Expression expressionTreeRoot)
   at Remotion.Linq.Parsing.ExpressionVisitors.SubQueryFindingExpressionVisitor.Visit(Expression expression)
   at System.Linq.Expressions.ExpressionVisitor.VisitLambda[T](Expression`1 node)
   at Remotion.Linq.Parsing.ExpressionVisitors.SubQueryFindingExpressionVisitor.Visit(Expression expression)
   at Remotion.Linq.Parsing.Structure.MethodCallExpressionParser.ProcessArgumentExpression(Expression argumentExpression)
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Remotion.Linq.Parsing.Structure.MethodCallExpressionParser.Parse(String associatedIdentifier, IExpressionNode source, IEnumerable`1 arguments, MethodCallExpression expressionToParse)
[...]

PR #4431, opened 2016-01-29
Failing test and fix for: Subquery.Where(Expression<Func<Entity, bool>>) throws exception

Tasteful:
Failing test for #3722


Tasteful:
If value of the evaluated parameter is an expression, use the expression instead of an expression-parameter that not can evaluated.

Fix for #3722

Tasteful:
Regression bug when same parametername is used
if same parameter name is used in expression the SQL that is generated will use same table alias for the subquery.

As temporary fix I added an ISqlGenerator.GenerateAlias that create unique alias, see (code from RC1)
https://github.com/Tasteful/EntityFramework/commit/cade7508e60b6ec4c2c71dd9b2e5e57f3423cd37 but you probably have a better way to do that on :)

anpete:
Blocked on #4537

Issue #4304, opened 2016-01-14
Query optimization: Remove unnecessary conditionals and casts from generated SQL


mikary:
The query test:
```cs
        [ConditionalFact]
        public virtual void Where_subquery_on_collection()
        {
            AssertQuery<Product, OrderDetail>((pr, od) =>
                pr.Where(
                        p => od
                            .Where(o => o.ProductID == p.ProductID)
                            .Select(odd => odd.Quantity).Contains<short>(5)));
        }
```

Produces the SQL:
```sql
SELECT [p].[ProductID], [p].[Discontinued], [p].[ProductName], [p].[UnitsInStock]
FROM [Products] AS [p]
WHERE (
    SELECT CASE
        WHEN 5 IN (
            SELECT [o].[Quantity]
            FROM [Order Details] AS [o]
            WHERE [o].[ProductID] = [p].[ProductID]
        )
        THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT)
    END
) = 1
```

Which could be improved by removing the `CASE` / `WHEN` in the `WHERE` clause:
```sql
SELECT [p].[ProductID], [p].[Discontinued], [p].[ProductName], [p].[UnitsInStock]
FROM [Products] AS [p]
WHERE (
    5 IN (
        SELECT [o].[Quantity]
        FROM [Order Details] AS [o]
        WHERE [o].[ProductID] = [p].[ProductID]
    )
)
```

mikary:
Had a discussion with @divega about this issue. The current implementation builds the `Case` and `Cast` into the expression to deal with boolean return types for SqlServer. Not all providers need the cast to return the correct value and as implemented it appears in subqueries and clauses where it isn't necessary. We should look into a solution that only impacts the SqlServer implementation, such as adding the cast in the `SqlServerQuerySqlGenerator`.

This issue also relates to `ResultOperator`s for `Any` and `All`. Updating title to reflect this.

PR #4433, opened 2016-01-29
Query optimization - Remove SELECT CASE for boolean values in WHERE clause

mikary:
Fix for #4304
@anpete @divega


mikary:
Could you also take a look @maumar

maumar:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1069,6 +1069,15 @@ protected override Expression VisitUnary(UnaryExpression expression)
                         return VisitIsNotNull(isNullExpression);
                     }
 
+                    if(expression.Operand is ExistsExpression)
```
anpete: Formatting.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/ConditionalRemovingExpressionVisitor.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class ConditionalRemovingExpressionVisitor : ExpressionVisitorBase
+    {
+        public override Expression Visit([CanBeNull] Expression node)
+        {
+            var selectExpression = node as SelectExpression;
+
+            var conditionalExpression = selectExpression?.Projection.FirstOrDefault() as ConditionalExpression;
+
+            if (selectExpression?.Projection.Count == 1
```
anpete: Duplicate null check of selecExpression. Suggest doing the projection count test before pulling out the conditional expression.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/ConditionalRemovingExpressionVisitor.cs
```diff
@@ -0,0 +1,33 @@
...
+                && conditionalExpression != null
+                && conditionalExpression.Type == typeof(bool)
```
anpete: Could use null-conditional here and omit the null comparison above.

anpete:
:shipit:

PR #4439, opened 2016-01-30
Simplified rules for principal/dependent determination in one-to-one relationship

### LINKED ISSUE #4313: ModelBuilder: Avoid "guessing" principal/dependent in 1:1 relationships
If there isn't a clear principal/dependent in a 1:1 relationship then we should throw and instruct the developer to explicitly configure them (by identifying the FK property).

smitpatel:
Resolves #4313


smitpatel:
This takes care of cases when there are 2 reference navigations.
Will submit separate PR for single reference navigation case.
I have written tests to cover all possible scenarios. Let me know if any scenario is missing.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -528,4 +528,7 @@
   <data name="InvalidSetType" xml:space="preserve">
     <value>Cannot create a DbSet for '{typeName}' because this type is not included in the model for the context.</value>
   </data>
+  <data name="AmbiguousOneToOneRelationship" xml:space="preserve">
+    <value>Principal and dependent side cannot be determined for one to one relationship between entity types '{declaringEntityType}' and '{principalEntityType}' involving navigations - '{dependentToPrincipalNavigation}' and '{principalToDependentNavigation}'.</value>
```
smitpatel: @rowanmiller & @divega - Any improvement for exception message?

ajcvickers:
@smitpatel Can you write down the rules we are now using?

smitpatel:
One-to-one relationships
Identified using
- `HasOne().WithOne()` fluent API, or
- A pair of reference navigations
If user specifies principal/dependent side using
- `ForeignKeyAttribute` in non-ambiguous way, or
- `HasForeignKey` fluent API, or
- `HasPrincipalKey` fluent API
Then appropriate side will be marked as dependent.
In the absence of any of above, conventions will try to find a matching property(-ies) in both sides. IFF only one side has matching property(-ies) then that side will be used as dependent.
If principal/dependent ends are configured in some way then only compatible PK can be used as FK.
If principal/dependent cannot be determined then exception will be thrown during ModelBuiltConvention running.

ajcvickers:
:shipit:

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
rowanmiller: Here is my proposal:

> The child/dependent side could not be determined for the one-to-one relationship that was detected between '{declaringEntityType}'.'{dependentToPrincipalNavigation}' and '{principalEntityType}'.'{principalToDependentNavigation}'. To identify the child/dependent side of the relationship, configure the foreign key property. See <FWLink> for more details.

rowanmiller: FWLink would probably go to here http://docs.efproject.net/en/latest/modeling/relationships.html#one-to-one

smitpatel: Navigations can be null (using fluent API).
Link for more details?

rowanmiller: I think it can be <TypeName>.<Navigation> if there is a navigation property and just <TypeName> if there is none.

rowanmiller: http://go.microsoft.com/fwlink/?LinkId=724062

PR #4460, opened 2016-02-02
Fix build failures on core clr

smitpatel:
- Add missing virual
- Disable failing InMemory Tests
- Disable asserting SQL in tests where statements are generated in different order for different runtime

With this changes `build.cmd` finishes fully.


lajones:
Looks good. :shipit:

PR #4464, opened 2016-02-02
Cleanup database model factories

### LINKED ISSUE #3454: Cleanup implementation of SqlServerDatabaseModelFactory
Current implementation (done by yours truly) needs some TLC.

- [x] Don't use index numbers to fetch fields
- [x] Quirks for datetime precision are only present in the comments
- [x] Queries for metadata
- [x] Server-size eval of `TableSelectionSet`. Current method pulls all into memory.
- [x] Implement fetching the DefaultSchemaName

Could be done as a part of #3438

natemcmaster:
Also fix minor bug in filtering indexes on views in reverse engineering.

Close #3454 

cc @lajones


src/Microsoft.EntityFrameworkCore.Relational.Design/Internal/FieldDbDataReader.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Data.Common;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
+{
+    public class FieldDbDataReader : IDisposable
+    {
+        private readonly DbDataReader _reader;
+        private readonly IDictionary<string, int> _indexes = new Dictionary<string, int>();
```
lajones: StringComparer.OrdinalIgnoreCase?

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -149,30 +155,21 @@ private void GetSequences()
...
                         Database = _databaseModel,
-                        SchemaName = reader.GetStringOrNull(dboIdx),
-                        Name = reader.GetStringOrNull(nameIdx),
-                        DataType = reader.GetStringOrNull(typeIdx),
-                        IsCyclic = reader.GetBoolean(cycleIdx),
-                        IncrementBy = reader.GetInt32(incrIdx),
-                        Start = reader.GetInt64(startIdx),
-                        Min = reader.GetInt64(minIdx),
-                        Max = reader.GetInt64(maxIdx)
+                        SchemaName = reader.GetFieldValue<string>("schema_name"),
```
divega: Not sure this is cleaner than what it was already in place:
1. IMO, the shape and the name of the FieldDbDataReader are weird even for an internal type
 - It is an incomplete wrapper for DbDataReader when the mail goal is to extend what `GetFieldValue<T>()` does 
 - The name doesn't seem to mean anything
2. It may not matter in this case but accessing the fields by index is generally faster than doing it by name. Incidentally using EF for this would have made the code nicer and use indexes (because shapers use indexes) at the same time :smile:

lajones:
@natemcmaster GitHub has stopped accepting inline comments for reasons I don't understand. So for this line: `var nullable = reader.GetFieldValue<bool>("nullable");` in `SqlServerDatabaseModelFactory` I think it should be `bool?` to match what it did before.

lajones:
@natemcmaster Also can you check that for all `GetFieldValue<bool?>()`s we do want the value to be `false` if we receive `NULL` from the DB? I seem to remember there was one instance where we wanted it to be `true` instead but I can't find it now.

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
natemcmaster: @divega I certainly debated whether to include this internal class or not. I'm not convinced it is the right move either. If manually mapping field names => field index via dozens of `.GetOrdinal()` calls is okay, I'll revert to that approach instead.

src/Microsoft.EntityFrameworkCore.Relational.Design/Internal/FieldDbDataReader.cs
```diff
@@ -0,0 +1,41 @@
...
+            Check.NotNull(reader, nameof(reader));
+
+            _reader = reader;
+        }
+
+        public virtual T GetFieldValue<T>([NotNull] string name)
+        {
+            if (!_indexes.ContainsKey(name))
+            {
+                _indexes.Add(name, _reader.GetOrdinal(name));
```
divega: DbDataReaders already map field names to field indexes. I don't think we have enough evidence that adding our own map here will be better. The only optimization that is worth considering is to avoid using the map for each row, i.e. only use it once at the beginning of a query and store the field index as we were doing in some of the code you have replaced in this PR. If we convince ourselves that we don't need that optimization then I would suggest just calling the string based `GetXyz()` APIs on the reader returned by the provider as that will be less code and probably as fast as maintaining our own map.

divega: E.g. we could have extension methods that take the field name and return the value by calling the `GetOrdinal()` method every time.

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
divega: Would it be worth doing some perf testing to understand the value of getting the indexes only once per query vs. getting them every time in the reverse engineering scenario?

BTW, I have seen in some of the code in the rest of the PR we have hardcoded "magic" numbers for the field ordinals. I think that is brittle as changes to the SQL queries will cause it to break. Once we pick a pattern we should use it consistently.

src/Microsoft.EntityFrameworkCore.Relational.Design/Internal/FieldDbDataReader.cs
natemcmaster: Maybe I missed the point of your remark. This class caches the field/index map for each reader and doesn't refresh it per row. From a reveng perspective, the perf gain is probably not worth adding more code. We could settle for something simpler, such as an extension as simple as 
```csharp
public T GetFieldValue<T>(this DbDataReader reader, string name) 
    => reader.GetFieldValue<T>(reader.GetOrdinal(name));
```

natemcmaster:
@lajones I double checked the usages of `Get<bool?>()`. The semantics should be preserved with this change.

@divega I updated this PR to use an extension instead of a wrapping class. I think it looks better even if the perf may be slightly slower.

lajones:
@natemcmaster ok - looks good. I'll let @divega sign-off.

src/Microsoft.EntityFrameworkCore.Relational.Design/Internal/FieldDbDataReader.cs
divega: Yep, I knew it didn't refresh it per row, but it did the name/ordinal lookup per field per row which the previous code avoided. 

Anyway, I agree the simple extension method you have in the new iteration makes the code cleaner. We can always come back to optimize this code if we find perf issues.

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -151,28 +157,19 @@ private void GetSequences()
...
-                var maxIdx = reader.GetOrdinal("maximum_value");
-                var startIdx = reader.GetOrdinal("start_value");
-                var incrIdx = reader.GetOrdinal("increment");
-
                 while (reader.Read())
                 {
                     var sequence = new SequenceModel
                     {
                         Database = _databaseModel,
-                        SchemaName = reader.GetStringOrNull(dboIdx),
```
divega: Is `SqlDataReaderExtensions.GetStringOrNull()` still in use after this change?

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
```diff
@@ -291,28 +257,14 @@ private void GetForeignKeys()
             }
         }
 
-        private static ReferentialAction? ConvertToReferentialAction(string onDeleteAction)
-        {
-            switch (onDeleteAction.ToUpperInvariant())
+        private static readonly IReadOnlyDictionary<string, ReferentialAction> ReferentialActionMap 
+            = new Dictionary<string, ReferentialAction>(StringComparer.OrdinalIgnoreCase)
             {
-                case "RESTRICT":
```
divega: Nit: Personally I prefer the low-tech implementation that was there before. Also, not sure if it matters but I think the new implementation will throw if SQLite ever starts supporting a different option, while in the old implementation it would return null, correct?

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
natemcmaster: Good catch. No, it's not used anymore. Will remove.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
natemcmaster: >  the new implementation will throw if SQLite ever starts supporting a different option

Seems unlikely, but since the "low-tech" implementation works, I'll leave it alone.

natemcmaster:
Ping @divega I've updated the PR

divega:
:shipit:

PR #4465, opened 2016-02-02
HasDefaultValueSql should update ValueGeneratedOnAdd if not set explicitly

### LINKED ISSUE #4328: OnModelCreating : ValueGenerated.OnAddOrUpdate results in ValueGeneratedOnAdd()
I am using EF version 7.0.0-rc2-16649 and dnx  1.0.0-rc2-16357

I use the following code in my context OnModelCreating() method, in order to automatically update the "DateMod" column on all entities which have the property :
```
foreach (var entityType in builder.Model.GetEntityTypes())
{
    var dateModProp = entityType.FindProperty("DateMod");
    if (dateModProp != null)
    {
        dateModProp.ValueGenerated = ValueGenerated.OnAddOrUpdate;
        builder.Entity(entityType.ClrType).Property(typeof(DateTime), "DateMod").HasDefaultValueSql("getutcdate()");
    }
}
```
I then updated an entity, and DateMod did not change.
After looking in the model snapshot generated code, I can see :
```
 b.Property<DateTime>("DateMod")
  .ValueGeneratedOnAdd()
  .HasAnnotation("Relational:GeneratedValueSql", "getutcdate()");
```
I think the method **ValueGeneratedOnAddOrUpdate()** should be called instead.

smitpatel:
resolves #4328 
If user has set ValueGenerated for property explicitly HasDefaultValueSql will not modify it.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/RelationalPropertyBuilderExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/RelationalPropertyBuilderExtensions.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 
 // ReSharper disable once CheckNamespace
@@ -52,7 +54,12 @@ namespace Microsoft.EntityFrameworkCore
             Check.NotNull(propertyBuilder, nameof(propertyBuilder));
             Check.NullButNotEmpty(sql, nameof(sql));
 
-            propertyBuilder.ValueGeneratedOnAdd();
+            var property = (Property)propertyBuilder.Metadata;
+            if (ConfigurationSource.Convention.Overrides(property.GetValueGeneratedConfigurationSource()))
+            {
+                property.SetValueGenerated(ValueGenerated.OnAdd, ConfigurationSource.Convention);
+            }
+
             propertyBuilder.Metadata.Relational().GeneratedValueSql = sql;
 
             return propertyBuilder;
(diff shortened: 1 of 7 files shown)


src/Microsoft.EntityFrameworkCore.Relational/RelationalPropertyBuilderExtensions.cs
```diff
@@ -52,7 +53,11 @@ public static class RelationalPropertyBuilderExtensions
             Check.NotNull(propertyBuilder, nameof(propertyBuilder));
             Check.NullButNotEmpty(sql, nameof(sql));
 
-            propertyBuilder.ValueGeneratedOnAdd();
+            if (((Property)propertyBuilder.Metadata).GetValueGeneratedConfigurationSource() != ConfigurationSource.Explicit)
```
divega: Should the plain `HasDefaultValue()` (without `Sql`) contain similar logic? I can't remember if there is a reason we only do it in the SQL version.

smitpatel: `HasDefaultValue()` does not contain any logic for ValueGenerated since it does not required value generator. @ajcvickers  can comment more on that.
Another question - Should `HasComputedColumnSql()` should have similar logic? It sets `ValueGenerated.AddOrUpdate`.

ajcvickers: @divega @smitpatel I think we probably should do it on HasDefaultValue as well so that the update pipeline will back-propagate the value. If the value is explicitly set, then the update pipeline should do nothing with it anyway. But it is possible that this will cause issues (bugs?) with other parts of the stack, such as the creation of identity columns by MIgrations. if so, we may want to fix the issues, or we may want to revisit.

HasComputedColumnSql should again only set AddOrUpdate if no other value has been explicitly set.

rowanmiller: :+1: on also setting it for `HasDefaultValue` - you can get an entity instance that doesn't match database values after `SaveChanges` without it.

smitpatel: Updated.
Added `ForSqliteHasDefaultValue` API which was not present before. (not sure of reason)

smitpatel:
Side-effect: If user uses `HasDefaultValue`, it will set ValueGenerated.OnAdd explicitly. After that even if user uses `HasComputedColumnSql` that Sql will be used as default value sql only.

ajcvickers:
@smitpatel If we set ValueGenerated implicitly or the application, then it should probably be marked as set by convention so that it can later be overridden by a different convention.

smitpatel:
Updated to have the value generated to be set by ConfigurationSource.Convention

ajcvickers:
:shipit:

PR #4466, opened 2016-02-02
Improve perf of initial fixup for Add

ajcvickers:
Initial testing shows about 65% perf improvement for Add/AddRange.

The change is to handle the case where there is a dependent being tracked before a principal and the dependent has a reference to the principal, but the principal does not have a reference to the dependent. In this case when we later start tracking the principal and it potentially gets its PK value, we need to go back and propagate that value to the FK. But finding the dependent entity at that time requires an expensive scan, so instead we keep a dictionary to look at later.


divega:
:shipit: Nice!

PR #4467, opened 2016-02-02
Fix #4308 - InvalidCastException for decimal without significant digits

### LINKED ISSUE #4308: Specified cast is not valid
In my example, I have a `Time` class with a `Minutes` field defined using the `short` data type.

If I try to get the total decimal hours like this:
`var totalHours = _dbContext.Times.Sum(t => t.Minutes / 60m);`

I get the error:

`An exception of type 'System.InvalidCastException' occurred in EntityFramework.Core.dll but was not handled in user code. Additional information: Specified cast is not valid.`

This, however, works fine:
`var totalHours = _dbContext.Times.Sum(t => t.Minutes / 60);`

mikary:
Corrects issue where generated SQL for `decimal` types did not always include digits following the decimal (invalid cast from boxed `int` to `decimal` and incorrect use of integer math on server-side computations).
CC: @bricelam @natemcmaster @anpete

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalSqlGenerationHelper.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalSqlGenerationHelper.cs
@@ -14,6 +14,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
     public class RelationalSqlGenerationHelper : ISqlGenerationHelper
     {
         protected virtual string FloatingPointFormat => "{0}E0";
+        protected virtual string DecimalFormat => "0.0###########################";
         protected virtual string DateTimeFormat => @"yyyy-MM-dd HH\:mm\:ss.fffffff";
         protected virtual string DateTimeOffsetFormat => @"yyyy-MM-dd HH\:mm\:ss.fffffffzzz";
 
@@ -68,7 +69,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
             => value.ToString();
 
         protected virtual string GenerateLiteralValue(decimal value)
-            => string.Format(value.ToString(CultureInfo.InvariantCulture));
+            => string.Format(value.ToString(DecimalFormat, CultureInfo.InvariantCulture));
 
         protected virtual string GenerateLiteralValue(double value)
             => string.Format(CultureInfo.InvariantCulture, FloatingPointFormat, value);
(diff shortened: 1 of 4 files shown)


natemcmaster:
LGTM

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalSqlGenerationHelper.cs
```diff
@@ -14,6 +14,8 @@ namespace Microsoft.EntityFrameworkCore.Storage
     public class RelationalSqlGenerationHelper : ISqlGenerationHelper
     {
         protected virtual string FloatingPointFormat => "{0}E0";
+        protected virtual int DecimalPrecision => 18;
```
divega: Curious, why 18?

mikary: Pulled the value from the `DefaultDecimalPrecision` in the `SqlGenerator` for EF6

EF6 also has a fair amount of logic around the total length of the string and conditionally adds a server side `decimal()` call.

divega: I wonder if 30 would be a better number so that there is no data loss for any .NET decimal instance.

mikary: Do we want to consider a solution closer to what EF6 does? Upon closer inspection it looks like a longer decimal (i.e. 30 characters long) would result in SQL that looks more like:

```sql
cast( 1.1234567890123456789012345678 as decimal(30))
```

divega: I don't know if/when we need the cast. I would test without it and only add it if necessary. I was just pointing out that using 18 will cause rounding and data loss for some decimal instances, and it seems unnecessary to go through that data loss given that SQL Server supports even more precision.

divega: Also if we had to add the cast, I imagine it is possible we would need to provide the right scale, not just precision. Yet another reason to avoid it if possible.

mikary:
Updated based on feedback

mikary:
Updated and rebased
ping @divega

divega:
:shipit:

PR #4477, opened 2016-02-03
Avoid recursively copied content

smitpatel:
The issue: When we build our design functional tests we copy .expected files to bin & .testpublish folders. On the subsequent runs it copies files from those folders too. Hence recursively creating more and more folders inside bin & .testpublish. All this works fine till the path becomes too long at which point build-tests phase fails.

Fix: We are restricting our scope to ReverseEngineering folder where we have created files. Exclude syntax in project.json is not able to work correctly somehow. I am trying to create a standalone repro to file bug on dotnet cli.


bricelam:
:shipit:

smitpatel:
@bricelam - review :smile:

natemcmaster:
:shipit:

PR #4479, opened 2016-02-03
Add additional property in function tests due to SQL CE limitation

### LINKED ISSUE #4455: Recent test change exposes SQL CE limitation
The changes to the test model in fix 
https://github.com/aspnet/EntityFramework/commit/788aee079f6aec2b077113f18cec82322da09336 has exposed an issue with the SQL Server Compact provider. 
The changes prevent any tests in InheritanceRelationshipsQuerySqlCeTest from passing 
see: https://github.com/ErikEJ/EntityFramework.SqlServerCompact/issues/299

I am posting the issue in this repo for advice.
The new seeding causes the following statement, which is not supported by SQL Server Compact to be generated:

    INSERT INTO [ReferencedEntity]\r\nDEFAULT VALUES;

I think it was also an issue before this fix, but just not covered by any of my tests. In fact I cannot really see usage of "DEFAULT VALUES" covered by any functional tests at all.

This is the definition of [ReferencedEntity]:

     -- Script Date: 02-02-2016 14:30  - ErikEJ.SqlCeScripting version 3.5.2.58
    CREATE TABLE [ReferencedEntity] (
      [Id] int IDENTITY (1,1) NOT NULL
    );
    GO
    ALTER TABLE [ReferencedEntity] ADD CONSTRAINT [PK_ReferencedEntity] PRIMARY KEY ([Id]);
    GO
 
The "real" issue here is that SQL Server Compact is not designed to work with a table with just a single IDENTITY column

smitpatel:
Resolves #4455


ajcvickers:
:shipit:

Issue #4399, opened 2016-01-26
SQL Server bulk insert doesn't allow inserting varchar column values longer than 1 character


jdlundquist:
### Title
SQL Server bulk insert generates table without specifying varchar column size

### Functional impact
For SQL Server, a runtime exception is thrown and you cannot save changes on a `DbContext` when trying to insert more than one of the same type of entity if that entity has an underlying `varchar` column and you pass a string value longer than 1 character in length.

### Minimal repro steps
My actual usage scenario involves various tables and entities, but the following is my hypothesis of steps to repro:

1. Create a database with a table having a column that is `varchar(max)` (or perhaps any size greater than 1), e.g. `MyTable.ExampleColumn`
2. Scaffold (or hand-build) a derived class from `DbContext` and map `MyTable.ExampleColumn` with `.HasColumnType("varchar(max)")`
3. Create multiple instances of the `MyTable` entity and put a string value in `ExampleColumn` longer than 1 character
4. Add the multiple entities via `DbContext.Add`
5. Call `DbContext.SaveChanges()` or `DbContext.SaveChangesAsync()`

### Expected result
Adding multiple entities of the same type to a `DbContext` would insert successfully into the database without throwing an exception, even when they have string column values containing data longer than 1 character

### Actual result
A runtime exception is thrown: `SqlException: String or binary data would be truncated.`
None of the changes to the `DbContext` are saved to the database.

### Further technical details
I think the starting point for digging into the issue is: `Microsoft.EntityFrameworkCore.Update.Internal.SqlServerUpdateSqlGenerator `.  The `GetTypeNameForCopy` method (used by `AppendDeclareTable`) should probably include a length for applicable data types (e.g. varchar, nvarchar, etc.) when generating the columns for the temp table.

Looking at the code statically without running the EF7 code directly, here's my suggestion for a couple high level things that could be done to make this work:
[...]

PR #4482, opened 2016-02-03
Use type mapper correctly in SQL Server batch updates

ajcvickers:
See Issue #4399

The issue was that the update pipeline was always asking the type mapper for the type to use even if a type had been explicitly set for the property. Fixed this and added tests for inserting all mapped types as both single inserts and multiple inserts both with and without Identity columns.


divega:
:shipit:

Issue #4287, opened 2016-01-12
Mapping column to UInt32 throws an invalid cast exception


TrabacchinLuigi:
doing so, ef7 will correctly generate a bigger datatype for sql (bigint) but is unabe to "re-hidratate" the class instance with a value, it will throw an invalid cast exception.
Actually i was trying to have UInt32 as Id in my POCO classes.

TrabacchinLuigi:
Well ... actually mapping uint to a bigint could also be considered a waste of space, because of the same amount of bits between uint and int.
But maybe for the consistency of the information it's better to not change this behaviour...

Just saying it would fit into an int column...

rowanmiller:
**Notes for triage:** We do create a schema for the model, so migrations is handling the `uint` and creating a `bigint` column. Both query and save fail though.

SaveChanges exception

```
Microsoft.Data.Entity.DbUpdateException was unhandled
  HResult=-2146233088
  Message=An error occurred while updating the entries. See the inner exception for details.
  Source=EntityFramework.Relational
  StackTrace:
       at Microsoft.Data.Entity.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)
       at Microsoft.Data.Entity.Update.Internal.BatchExecutor.Execute(IEnumerable`1 commandBatches, IRelationalConnection connection)
       at Microsoft.Data.Entity.Storage.RelationalDatabase.SaveChanges(IReadOnlyList`1 entries)
       at Microsoft.Data.Entity.ChangeTracking.Internal.StateManager.SaveChanges(IReadOnlyList`1 entriesToSave)
       at Microsoft.Data.Entity.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
       at Microsoft.Data.Entity.DbContext.SaveChanges(Boolean acceptAllChangesOnSuccess)
       at Microsoft.Data.Entity.DbContext.SaveChanges()
       at EF7ConsoleApplication.Program.Main(String[] args) in C:\Users\rowmil\Documents\Visual Studio 2015\Projects\Console.CI\Console.CI\Program.cs:line 18
       at System.AppDomain._nExecuteAssembly(RuntimeAssembly assembly, String[] args)
       at System.AppDomain.ExecuteAssembly(String assemblyFile, Evidence assemblySecurity, String[] args)
       at Microsoft.VisualStudio.HostingProcess.HostProc.RunUsersAssembly()
[...]

PR #4485, opened 2016-02-04
Remove unsupported primitive types from SQL Server type mapper

ajcvickers:
See Issue #4287

The SQL Server type mapper would produce mappings for sbyte, ushort, uint, and ulong, even though these are currently not supported and cannot be supported without type conversions.


lajones:
:shipit:

divega:
:shipit:

PR #4486, opened 2016-02-04
Improve RelationalExpressionPrinter to also print shaper lambdas. 

maumar:
Since the lambdas are compiled into the expression we are now producing a string representation for them before compiling, and store it in the shaper.


maumar:
@anpete

maumar:
new version up

anpete:
@maumar I don't think we should flow the materializer string around. We need a solution that is pay-per-play. Easiest might just be to log the materializer expression at the time it is being created.

maumar:
will open another one

Issue #753, opened 2014-10-01
Handle creating delegates for private setters on base types


ajcvickers:
Properties can be defined in a base type with a private setter and then overridden. In such cases the setter does not show up in the .NET Reflection metadata for the property in the derived type. We therefore need to use special code to go down the hierarchy and find the private setter on the base type so it can be mapped to a delegate and called. (See EF6 codebase for examples.) (The same is true for getters so we should probably support both cases, although I've never seen the getter case in the wild.)

MatthewMWR:
+1. I got nailed by this my first time out trying EF7 (experienced as per https://github.com/aspnet/EntityFramework/issues/1026 ).

phillip-haydon:
So EF 7 is supporting private setters ?

rowanmiller:
@phillip-haydon They are actually supported already. The following code works with EF7 Beta 4. Private property setters were actually supported in previous version of EF too.

```c#
using Microsoft.Data.Entity;
using System;

namespace ConsoleApplication39
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var db = new BloggingContext())
            {
                db.Database.EnsureCreated();

                db.Blogs.Add(new Blog("romiller.com"));
                db.SaveChanges();

                foreach (var item in db.Blogs)
                {
                    Console.WriteLine(item.Url);
                }
            }
        }
    }

    public class BloggingContext : DbContext
    {
        public DbSet<Blog> Blogs { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=PrivateSetters;Trusted_Connection=True;");
        }
    }

    public class Blog
    {
        private Blog()
        { }

        public Blog(string url)
        {
            Url = url;
        }

        public int BlogId { get; private set; }
        public string Url { get; private set; }
    }
}
```

PR #4491, opened 2016-02-04
Handle private setters in ClrPropertySetterFactory #753

davidroth:
This is a fix for issue #753 (Handle creating delegates for private setters on base types)

As described in the original issue:

_"Properties can be defined in a base type with a private setter and then overridden. In such cases the setter does not show up in the .NET Reflection metadata for the property in the derived type. We therefore need to use special code to go down the hierarchy and find the private setter on the base type so it can be mapped to a delegate and called."_

The fix checks if a setter method can be found on the `PropertyInfo` instance passed to `Create()`.
If the setter is null, we first check if the property is defined on the *DeclaringType* (This is the case, if the property is not virtual and not overwritten). If this is not successfull, the code walks down the inheritance hierarchy to check if the setter is declared on a base type instead. I have also added an additional `InvalidOperationException` (former: `NullReferenceException`), so that there is more context if no setter could be found at all.

I have added various test cases for handling both virtual and non-virtual properties with private setters, both cases with 2 layers of inheritance to verify the hierarchy walking code.

Happy to hear your feedback :)

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/ClrPropertySetterFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/ClrPropertySetterFactory.cs
@@ -10,18 +10,31 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
     {
         public override IClrPropertySetter Create(PropertyInfo property)
         {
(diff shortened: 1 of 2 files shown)


ajcvickers:
This has now been merged. Thanks for the great contribution!

PR #4496, opened 2016-02-04
Re-enable installation on net451

bricelam:
...again.

Note, dotnet/cli#1012 didn't help us much here because `Remotion.Linq`, `System.Diagnostics.DiagnosticSource`, `xunit.abstractions`, `xunit.assert`, and `xunit.extensibility.core` fall outside of its rules.

@victorhurdugaci

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/project.json
+++ b/src/Microsoft.EntityFrameworkCore.Relational/project.json
@@ -24,6 +24,9 @@
     "net451": {
       "frameworkAssemblies": {
         "System.Data": "",
+        "System.Diagnostics.Tracing": { "type": "build" },
+        "System.Runtime": { "type": "build" },
+        "System.Threading": { "type": "build" },
         "System.Transactions": ""
       }
     },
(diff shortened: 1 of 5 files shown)


victorhurdugaci:
@anurse @davidfowl what do you think?

bricelam:
Re-linq has been update (see [this issue](https://www.re-motion.org/jira/browse/RMLNQ-94)). I'll remove the facades from the project.json files under src.

ErikEJ:
@bricelam Does that mean that the current build on aspnetcidev is not useable with a .NET 4.5.1 only provider? (Ie I must wait for this change to be included in the aspnetcidev feed)

bricelam:
No, it will work, you just can't install your packages into a packages.config-based (i.e. non-project.json) project.

ErikEJ:
But my provider is packages.config- based :cry:

bricelam:
Then it's probably been broken since 65380c94d59b56e963c5bdcdd0d2cceb55470a1e

ErikEJ:
I will patiently await a new build on aspnetcidev, then. Thanks!

bricelam:
For reference, you're waiting for `1.0.0-rc2-16874` or higher.

PR #4498, opened 2016-02-05
Allow conventions to create one-to-one relationship with single reference navigation

### LINKED ISSUE #4313: ModelBuilder: Avoid "guessing" principal/dependent in 1:1 relationships

smitpatel:
Resolves #4313 

If a relationship
- Has dependent to principal navigation of reference type
- Has principal to dependent navigation set as null
- Has matching property(-ies) on principal side
- Has no matching property(-ies) on dependent side
Then conventions will try to invert and make one-to-one relationship using the single navigation

Conventions will not match "Id" as a matching property since it is ambiguous.
If "Id" is PK for dependent entity then conventions will use PK property for FK as per existing rules


ajcvickers:
:shipit:

PR #4500, opened 2016-02-05
Throw in model validation for unqualified types like "nvarchar" and "varbinary"

ajcvickers:
If the app really wants a data type like this with a length of one, then it should use, for example,  "nvarchar(1)". This stops apps from putting, for example, nvarchar and then setting a max length of 16 in the API and expecting this to be transformed into the type nvarchar(16).

Also, some cleanup in the type mapper and model validator classes.


rowanmiller:
:white_check_mark: Exception message looks good

AndriySvyryd:
:shipit:

Issue #4488, opened 2016-02-04
[Question/Issue] Adding entities with and without keys in the same context. The instance of entity type 'XXXX' cannot be tracked because another instance of this type with the same key is already being tracked.


danielcirket:
So, I have a DbSet<T> for which at runtime/startup I want to seed some initial/test data if it doesn't exist.

However i'm running into some issues when adding the first entry with a set key (will be a default value of -1), then trying to add the next item with no Id set to be generated, I get an exception thrown for the key already existing in the dbset.

For the entity type I'm using this as a base class:

```C#
public abstract class Entity
{
    public virtual int? Id { get; set; }
    public virtual DateTime CreatedDate { get; set; }
    public virtual int CreatedById { get; set; }
    public virtual DateTimeLastModifiedDate { get; set; }
    public virtual int LastModifiedById { get; set; }
    public virtual bool IsDeleted { get; set; }
}
```

With the entity as:

```C#
public class Country : Entity
{
   public string Name { get; set; }
}
```

(I have tried this with different combinations of the Id property having the UseSqlServerIdentityColumn() and ValueGeneratedOnAdd() and with both removed (Adding them seemed to make no difference, but wanted to try as many options as possible in case I was missing something).

```C#
public static void AddEntityDefaults<T>(this EntityTypeBuilder<T> source) where T : Entity    
{
    source.HasKey(x => x.Id);
    source.Property(x => x.Id).UseSqlServerIdentityColumn();
    source.Property(x => x.Id).ValueGeneratedOnAdd();
    source.Property(x => x.CreatedDate).HasDefaultValue(DateTime.Now).ValueGeneratedOnAdd();
    source.Property(x => x.LastModifiedDate).HasDefaultValue(DateTime.Now).ValueGeneratedOnAddOrUpdate();
    source.Property(x => x.IsDeleted)
        .HasDefaultValue(false)
        .IsRequired();
}
```

And for the 'seeding' I'm using this (Which works perfectly for a single entity OR a number of entities where the Id is null for all items). I have tried wrapping this in a transaction, and running an SQL command to enable then disable identity insert, saving the entity with the Id set before adding the additional items and simply adding them one at a time then saving.

This works for adding a single with the Id set to -1:

```C#
private static void InitialiseCountries(this DataContext source)
{
    if (source.Countries.Any())
        return;

    using (var transaction = source.Database.BeginTransaction())
    {
[...]

PR #4503, opened 2016-02-05
Fix for issue #4488

danielcirket:
Initial temp value starts at int.MinValue as per Rowan's comment.


danielcirket:
Messed up the commit, fixing.

PR #4504, opened 2016-02-05
Fix for #4488 

danielcirket:
as per Rowan's comment


rowanmiller:
It's not quite that simple... we were actually making the change on our end too https://github.com/aspnet/EntityFramework/pull/4505. Always important to run unit tests before submitting a PR :wink:.

danielcirket:
This ended up being the wrong branch I referenced anyway for some reason, knew it wasn't quite as simple as that, sorry about should have rechecked before submitting.

PR #4505, opened 2016-02-05
Change numeric temp value generators to use values that are even less likely to be used as real keys

ajcvickers:
See Issue #4488

This requires different start values and increment/decrement behavior for different types due to size limitations.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ValueGeneration/Internal/TemporaryByteValueGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore/ValueGeneration/Internal/TemporaryByteValueGenerator.cs
@@ -1,10 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Threading;
+
 namespace Microsoft.EntityFrameworkCore.ValueGeneration.Internal
 {
-    public abstract class TemporaryNumberValueGenerator<TValue> : ValueGenerator<TValue>
+    public class TemporaryByteValueGenerator : TemporaryNumberValueGenerator<byte>
     {
-        public override bool GeneratesTemporaryValues => true;
+        private int _current;
+
+        public override byte Next() => (byte)Interlocked.Decrement(ref _current);
     }
 }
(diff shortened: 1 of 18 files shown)


davidroth:
@ajcvickers Out of curiosity: Why do you start at int.MinValue + **1000** ? 
What is the benefit of adding **1000**?

ajcvickers:
@davidroth Because it was pointed out that some apps might use int.MinValue as a sentinel. Given that the idea is not to do something that can always work (apps can change the service or not use temp values in extreme cases) but will work most of the time by not colliding, this was just an extra bit of making it not collide in most cases. It's not very important, and the 1000 value was pulled out of...the air. :-)

AndriySvyryd:
Why can't we make the temporality part of the key value? This would avoid all collisions. The temporality would be propagated to FKs during fixup. Explicit FK values would only match explicit PK values.

ajcvickers:
@AndriySvyryd I tried this a long time ago and it had several issues. For example, when application code tried to set an FK using something like post.CategoryID = category.Id, it wouldn't work. This was really confusing because looking at the entities everything matched up fine. Also, we could no longer just use an int, long, or whatever as the key value in identity maps and similar, which would prevent us using the optimized non-boxing services for this. We could create structs with an extra bit in, but it is again more complicated, more memory, and likely at least a bit  Basically, it seems like we should just keep this simple.

AndriySvyryd:
@ajcvickers In my opinion `post.CategoryID = category.Id` not working if `category.Id` hasn't been assigned is not so bad, but if the team thinks otherwise, so be it.
I would however suggest adding more information to the exception message if a collision does occur with a temporary value specifically.

divega:
> I would however suggest adding more information to the exception message if a collision does occur with a temporary value specifically.

I think this is a good idea. Personally I would prefer to get a good exception for collisions with temporary values instead of seeing temporary values start on arbitrary numbers. Temporary values is something sooner or later as a developer using EF Core I need to understand.

ajcvickers:
@divega Agreed. Somebody should file a bug for this.

divega:
#4525

AndriySvyryd:
:shipit:

PR #4506, opened 2016-02-05
Match compatible PK as FK properly

### LINKED ISSUE #4451: Setting principal key should run fk discovery convention
Setting principal key determines principal/dependent ends explicitly which can help in figure out ambiguity. This should trigger foreignkey property discovery convention so that appropriate type from dependent side can be used as fk property.

smitpatel:
resolves #4451 

Conditions
- One-to-One relationship (non-self-PK-referencing)
- No matching properties on either side
Either of following two
- Principal Key is set with source higher than convention, or
- Principal End & Uniqueness set with source higher than convention

Issue 1: For one-to-one relationship configured conventionally, we don't match compatible PK since principal end is ambiguous. If on such relationship, `HasPrincipalKey` is used then principal end is specified by user and we should try matching PK. But if the `HasPrincipalKey` calls the same principal key as the current one then conventions are not run.

Issue 2: Using `HasOne` API sets principal end explicitly. So we may use half-configured relationship to determine side and assign FK properties even though when relationship is full-configured it can be ambiguous. Therefore check for principal end CS should happen with Unique CS since `WithOne` or `WithMany` sets uniqueness.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyPropertyDiscoveryConvention.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyPropertyDiscoveryConvention.cs
@@ -9,7 +9,7 @@ using Microsoft.EntityFrameworkCore.Metadata.Internal;
 
 namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
 {
-    public class ForeignKeyPropertyDiscoveryConvention : IForeignKeyConvention, INavigationConvention, IPropertyConvention
+    public class ForeignKeyPropertyDiscoveryConvention : IForeignKeyConvention, INavigationConvention, IPropertyConvention, IPrincipalEndConvention
     {
         public virtual InternalRelationshipBuilder Apply(InternalRelationshipBuilder relationshipBuilder)
         {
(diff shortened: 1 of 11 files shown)


smitpatel:
Will be making some changes to this. Once updated I'll put needs-review tag again.

smitpatel:
Ready for review.

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyPropertyDiscoveryConvention.cs
```diff
@@ -46,7 +46,8 @@ public virtual InternalRelationshipBuilder Apply(InternalRelationshipBuilder rel
                 }
 
                 // Only match with PK if the principal end is set
-                if ((!ConfigurationSource.Convention.Overrides(foreignKey.GetPrincipalEndConfigurationSource())
+                if (((!ConfigurationSource.Convention.Overrides(foreignKey.GetPrincipalEndConfigurationSource())
+                       && !ConfigurationSource.Convention.Overrides(foreignKey.GetIsUniqueConfigurationSource()))
```
AndriySvyryd: If the principal end is ambiguous, then it's CS should be null. So the only condition here should be
```
foreignKey.GetPrincipalEndConfigurationSource() != null
```

If there are places like `HasOne` that set the principal end CS when they shouldn't then they need to be fixed.

smitpatel:
@AndriySvyryd - Updated.

AndriySvyryd:
:shipit:

PR #4507, opened 2016-02-05
[TINY] Fixes for benchmark CI

rowanmiller:
Because we now consume artifacts from the central CI build (rather than
cloning and building on the perf agent), we need to copy config files so
that they are part of the artifacts from the build.

Also removing redundant TestFramework attribute from EF6 tests.


ajcvickers:
:shipit:

PR #4508, opened 2016-02-05
Store DefaultValue annotation of enum type as underlying type value

### LINKED ISSUE #4061: Migrations: Convert enum values to underlying type
When creating a migration to add an `enum` column, the namespace of the enum is not included in the resulting migration.

EG: 

```
                name: "Status",
                table: "CommandRecord",
                nullable: false,
                defaultValue: CommandStatus.Pending);
```

This results in `DNX,Version=v4.5.1 error CS0103: The name 'CommandStatus' does not exist in the current context` when running `dnx ef database update`

smitpatel:
resolves #4061

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.Relational.Tests/Metadata/RelationalBuilderExtensionsTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.Relational.Tests/Metadata/RelationalBuilderExtensionsTest.cs
@@ -146,6 +146,23 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Tests
         }
 
         [Fact]
+        public void Can_set_column_default_value_of_enum_type()
+        {
+            var modelBuilder = CreateConventionModelBuilder();
+
+            modelBuilder
+                .Entity<Customer>()
+                .Property(e => e.EnumValue)
+                .HasDefaultValue(MyEnum.Tue);
+
+            var property = modelBuilder.Model.FindEntityType(typeof(Customer)).FindProperty("EnumValue");
+
+            Assert.Equal(typeof(ulong), property.Relational().DefaultValue.GetType());
+            Assert.Equal((ulong)2, property.Relational().DefaultValue);
+            Assert.Equal(ValueGenerated.OnAdd, property.ValueGenerated);
+        }
+
+        [Fact]
         public void Can_set_key_name()
         {
             var modelBuilder = CreateConventionModelBuilder();
@@ -1000,10 +1017,18 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Tests
             Assert.Same(typeof(int), sequence.ClrType);
         }
 
+        private enum MyEnum : ulong
+        {
+            Sun,
+            Mon,
+            Tue
+        }
+
         private class Customer
         {
             public int Id { get; set; }
             public string Name { get; set; }
+            public MyEnum EnumValue { get; set; }
 
             public IEnumerable<Order> Orders { get; set; }
         }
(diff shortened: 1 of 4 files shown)


test/Microsoft.EntityFrameworkCore.Relational.Tests/Metadata/RelationalBuilderExtensionsTest.cs
```diff
@@ -146,6 +146,23 @@ public void Setting_column_default_value_does_not_modify_explicitly_set_value_ge
...
+
+            modelBuilder
+                .Entity<Customer>()
+                .Property(e => e.EnumValue)
+                .HasDefaultValue(MyEnum.Tue);
+
+            var property = modelBuilder.Model.FindEntityType(typeof(Customer)).FindProperty("EnumValue");
+
+            Assert.Equal(typeof(ulong), property.Relational().DefaultValue.GetType());
+            Assert.Equal((ulong)2, property.Relational().DefaultValue);
```
bricelam: This behavior is odd to me, but the team seems to be OK with it for now. cc @rowanmiller

bricelam:
:shipit:

test/Microsoft.EntityFrameworkCore.Relational.Tests/Metadata/RelationalBuilderExtensionsTest.cs
divega: I agree it is odd. But it seems fundamental to the way we figured out we could fix this issue. Do you have a different approach in mind?

bricelam: The way I saw the fix originally was to "erase" the enum value in the model snapshot during serialization, and compensate for the mismatch between the current model and the model snapshot in the model differ.

rowanmiller: I like this approach better... to me the enum is just sugar for the actual value. I'm not super passionate either way... just feels like we are best insulated from renamed enum types/values that way.

ajcvickers: I think we should do whatever is easier. We don't support arbitrary types in annotations because we can't serialize and deserialize them. We also don't constrain what kinds of types can be mapped by providers. This means if a user puts a default value of some arbitrary type on a property, then we will bomb. As far as I know we are not attempting to fix this for v1, but we decided we do need to support enums because they are very common. So the question was how to do this in the simplest way possible. If we had a generalized mechanism to support arbitrary types, then we would use that. Until then, I say do whatever is easier.

divega: :+1: on taking the easiest possible approach that works for now. I was thinking I like @bricelam's idea slightly better because I suspect when/if we add a more general type mapping feature it is going to be desirable to be able to set and get the default value as an instance of the actual type (and we'll probably have to tackle serialization as a separate aspect). However I am not even sure that enums in general should be mapped as primitive types as opposed to model types (see one of the reasons at http://www.postgresql.org/docs/9.2/static/datatype-enum.html).

smitpatel: This approach is easier since Enum is out of picture as soon as we store in metadata. Which avoids any extra dealing with Enum types later.
Should I merge with this?

divega: I am ok with it.

PR #4509, opened 2016-02-06
Remove dnx451 in all test code

natemcmaster:
Fix https://github.com/aspnet/EntityFramework/issues/4495


bricelam:
This will break running tests in Visual Studio on ASP.NET 5 RC1

natemcmaster:
I think we should do this even though VS tooling won't let us run full .NET tests. We haven't been running tests such as `GetContextType_works_cross_domain` for weeks since they are only full .NET tests AND have are `[ConditionalFact]`'s

bricelam:
cc @divega #4550 will also break running tests in VS. But IMHO, keeping the "daily dev workflow" working is our highest priority right now.

natemcmaster:
Ping again. We need this on CI. The `-noappdomain` solution doesn't work and dnxcore50 tests should still run in VS.

natemcmaster:
See #4638

Issue #3115, opened 2015-09-12
Support passing DbParameter instances to raw SQL APIs


Vasimovic:
Example:-

Instead of:
var UserType = dbcontext.Set().FromSql("dbo.SomeSproc @p0, @p1", 45, "Ada")

Use named parameters:

object[] sqlParams = {
  new SqlParameter("@Id", 45),
  new SqlParameter("@Name", "Ada")
};

var UserType = dbcontext.Set().FromSql("dbo.SomeSproc", sqlParams)

If SqlParameter is not supported then something like this would suffice:

var parameterDictionary = new Dictionary
{
  {"@Id", 45},
  {"@Name", "Ada"}
};

This would make it easier to find parameters in SQL profiler, also by using a list of named parameters there would be no need to worry about parameter order

rowanmiller:
**Note for triage:** We should support passing in DbParameters as well as just the raw values (we supported this in EF6), but it doesn't work at the moment.

tuespetre:
Dictionaries are cool and all but the 'micro-orm' approach of allowing anonymous objects to be passed in is much more convenient.

prasannapattam:
I also vote for anonymous object approach to pass in the named parameters.

vmudinas:
Hi, is were a way to get FromSql working if I don't have Id  or identity column ?

mikes-gh:
Using ExecuteSQLCommand

Whats the best way of getting the value of an SqlParameter of type System.Data.ParameterDirection.Output as it stands.  Do I need to go back to SqlCommand

rowanmiller:
@mikes-gh - yes, for the moment.

tessSnap:
How do we return an output value ? For example I have a stored procedure that returns 0 if everything goes well and 1 if an error occurs .

divega:
@tessSnap Yes, dropping down to ADO.NET is going to be the solution until we remove this limitation.

mikes-gh:
For those of you using SQL command I found this useful to get a connection string from existing context.

someContext.Database.GetDbConnection().ConnectionString

then use a new connection .  

I tried to borrow context connection and cast to SQLConnection for my command but doing that breaks the context you borrowed from.

prasannapattam:
This code is working for me. I am using Connection object. 

            SqlConnection connection = (SqlConnection)dbContext.Database.GetDbConnection();
            connection.Open();
[...]

PR #4518, opened 2016-02-08
Fix #3704 - Remove values from RelationalCommand construction

### LINKED ISSUE #3704: RelationalCommand: stricter separation of construction and execution
PR https://github.com/aspnet/EntityFramework/pull/3690 makes the minimal changes required to enable caching and reuse of `RelationalCommand`: parameter values can now be passed in each individual `Execute` method call in addition to the constructor. 

The change doesn't preclude the existing pattern from being used though, i.e. you create a RelationalCommand passing both metadata and values for the parameters in the constructor, call one of the `Execute` methods with a null or empty dictionary of parameter values and then discard the `RelationalCommand`. 

Although the option is not exercised in our codebase, it is even possible to use hybrid patterns in which sometimes initial parameter values are used and other times the "fresh" parameter values passed to the various `Execute` methods are used (but from reading the code I think this won't always work as expected because null parameters passed to the `Execute` methods are ignored in favor of the corresponding initial values).

@anpete and I had a chat about this and we believe there are potential benefits in terms of code cleanliness and efficiency in forcing all usages of `RelationalCommand` to separate the construction of the command's parameters from the binding of parameter values at execution time.

mikary:
Allow providing parameter values as arguments or cached property value

Cleans up `RelationalCommand`:
 - When the `RelationalCommand` is built, no parameter values are referenced (query scenario)
 - Values can be added to the command after building (other scenarios)
  - Cached values will be used by default
  - If values are passed in, they will override the cached values (all or none)

These changes are an incremental step for #3115. Changes for that issue will include command and parameter building API consistency and extensibility cleanup.

@divega @maumar @anpete

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
@@ -244,6 +244,12 @@
   <data name="ClientEvalWarning" xml:space="preserve">
     <value>The LINQ expression '{expression}' could not be translated and will be evaluated locally.</value>
   </data>
+  <data name="NullParameterValue" xml:space="preserve">
+    <value>The argument '{argumentName}' can not be null when parameters have been configured.</value>
+  </data>
+  <data name="MissingParameterValue" xml:space="preserve">
+    <value>No value provided for requred parameter '{parameter}'.</value>
+  </data>
   <data name="RelationalLoggerExecutedCommand" xml:space="preserve">
     <value>Executed DbCommand ({elapsed}ms) [Parameters=[{parameters}], CommandType='{commandType}', CommandTimeout='{commandTimeout}']{newLine}{commandText}</value>
   </data>
--- a/src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalCommand.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalCommand.cs
@@ -14,33 +14,39 @@ namespace Microsoft.EntityFrameworkCore.Storage
 
         IReadOnlyList<IRelationalParameter> Parameters { get; }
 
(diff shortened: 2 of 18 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
```diff
@@ -244,6 +244,12 @@
   <data name="ClientEvalWarning" xml:space="preserve">
     <value>The LINQ expression '{expression}' could not be translated and will be evaluated locally.</value>
   </data>
+  <data name="NullParameterValue" xml:space="preserve">
+    <value>The argument '{argumentName}' can not be null when parameters have been configured.</value>
+  </data>
+  <data name="MissingParameterValue" xml:space="preserve">
+    <value>No value provided for requred parameter '{parameter}'.</value>
```
divega: typo

src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalCommand.cs
```diff
@@ -14,33 +14,39 @@ public interface IRelationalCommand
...
             [NotNull] IRelationalConnection connection,
+            [CanBeNull] IReadOnlyDictionary<string, object> parameterValues = null,
             bool manageConnection = true,
             CancellationToken cancellationToken = default(CancellationToken));
 
         RelationalDataReader ExecuteReader(
             [NotNull] IRelationalConnection connection,
-            bool manageConnection = true,
-            [CanBeNull] IReadOnlyDictionary<string, object> parameterValues = null);
+            [CanBeNull] IReadOnlyDictionary<string, object> parameterValues = null,
```
divega: "providers-beware"? (assuming providers can re-implement this)

mikary:
Closing this PR after talking with @divega
Too much of what changes here depends on / relates to what happens for #3115 and it would be better to review all of the changes together.

PR #4532, opened 2016-02-10
Handle lifting of relationships between entity types in the same hierarchy

AndriySvyryd:
Make InternalRelationshipBuilder.DependentEntityType() and InternalRelationshipBuilder.PrincipalEntityType() not try to invert the relationship
Add InternalRelationshipBuilder.RelatedEntityTypes() that handles inversion, as well as lifting of dependent or principal ends
  This avoids ambiguity between lifting and inverting relationships between types in the same hierarchy
Make ForeignKeyPropertyDiscoveryConvention match shadow properties that weren't added by convention

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 
 namespace Microsoft.EntityFrameworkCore.Metadata.Internal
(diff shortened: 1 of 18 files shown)


ajcvickers:
:shipit:

src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -693,107 +597,86 @@ private bool CanSetUnique(bool isUnique, ConfigurationSource configurationSource
...
-                {
-                    if (!configurationSource.Overrides(Metadata.GetPrincipalKeyConfigurationSource()))
+                    if (runConventions)
                     {
-                        return null;
+                        builder = ModelBuilder.Metadata.ConventionDispatcher.OnPrincipalEndSet(builder);
                     }
 
-                    principalProperties = new Property[0];
+                    builder.Metadata.UpdatePrincipalEndConfigurationSource(configurationSource);
```
smitpatel: builder can be null after running conventions.
Also shouldn't source be upgraded before running convention. (since some conventions do rely on CS)

AndriySvyryd: Will fix

PR #4533, opened 2016-02-10
[SMALL] Fix framework reporting in benchmarks

rowanmiller:
We used to rely of conditional compilation to report Full .NET vs DNX,
but as part of moving to .NET CLI this was removed in
a80fd5277892dd1e9eee5c451df0b494ee98a9ff. We also no longer recompile
for each test run (we use central artifacts from CI build) - so
conditional compilation doesn't work anyway. Changing to just report CLR
vs CoreCLR and we can always report additional details via build
configuration parameters if needed.


rowanmiller:
@bricelam could you take a quick look at this one

bricelam:
:shipit:

PR #4537, opened 2016-02-11
Generate Unique alias for table names

### LINKED ISSUE #3826: Table alias collision in generated SQL with subquery (MSSQL)
Windows 10, VS 2015, IIS Express, SQL Server Express 2014, EF 7.0.0-rc1-final, ASP.NET 6.0.0-rc1-final

I am seeing incorrect SQL generated with subqueries when the tables involved begin with the same letter.

 Models:
```c#
public class Blog
{

        public ICollection<BlogPost> BlogPosts { get; set; }
}
```
```c#
public class BlogPost
{
        public int BlogPostId { get; set; }

        public string PostText { get; set; }
}
```
FK setup:
```c#
    modelBuilder.Entity<Blog>().HasMany(b => b.BlogPosts).WithOne(p => p.Blog);
```

Query:
```c#
   _dbContext.Blogs.Where(b => b.BlogPosts.Count == 0).ToList();
```

Generated SQL:
```sql
SELECT [b].[BlogId], [b].[Title]
FROM [Blog] AS [b]
WHERE (
    SELECT COUNT(*)
    FROM [BlogPost] AS [b]
    WHERE [b].[BlogId] = [b].[BlogId]
) = 0
```

With test data of 2 blogs, 2 posts for blog 1 and 0 for blog 2, this query incorrectly returns 0 results (should return one result, namely blog 2).  If `BlogPost` in the subquery is aliased to something other than [b] it works fine, e.g.:

```sql
SELECT [b].[BlogId], [b].[Title]
FROM [Blog] AS [b]
WHERE (
    SELECT COUNT(*)
    FROM [BlogPost] AS [bp]
    WHERE [bp].[BlogId] = [b].[BlogId]
) = 0
```

smitpatel:
resolves #3826 
Presently it is code change only. If approaches looks right then I will update SQL generated for tests and add more test to verify this scenario.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
@@ -135,7 +135,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
                 var targetEntityType = navigation.GetTargetType();
                 var targetTableName = _relationalAnnotationProvider.For(targetEntityType).TableName;
-                var targetTableAlias = targetTableName[0].ToString().ToLower();
+                var targetTableAlias = _queryCompilationContext.CreateUniqueTableAlias(targetTableName[0].ToString().ToLower());
 
                 if (!navigation.IsCollection())
                 {
@@ -179,7 +179,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
                         selectExpression.Predicate = oldPredicate;
                         selectExpression.RemoveTable(joinExpression);
-                        selectExpression.AddTable(newJoinExpression);
+                        selectExpression.AddTable(newJoinExpression, createUniqueAlias: false);
                         joinExpression = newJoinExpression;
                     }
 
@@ -223,7 +223,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                                 OrderingDirection.Asc);
                     }
 
-                    var targetSelectExpression = _selectExpressionFactory.Create();
+                    var targetSelectExpression = _selectExpressionFactory.Create(_queryCompilationContext);
 
                     targetTableExpression
                         = new TableExpression(
@@ -232,7 +232,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                             targetTableAlias,
                             querySource);
 
-                    targetSelectExpression.AddTable(targetTableExpression);
+                    targetSelectExpression.AddTable(targetTableExpression, createUniqueAlias: false);
 
                     var materializer
                         = _materializerFactory
@@ -290,8 +290,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         private JoinExpressionBase AdjustJoinExpression(
             SelectExpression selectExpression, JoinExpressionBase joinExpression)
         {
-            var subquery = new SelectExpression(_querySqlGeneratorFactory, joinExpression.Alias);
-            subquery.AddTable(joinExpression.TableExpression);
+            var subquery = new SelectExpression(_querySqlGeneratorFactory, _queryCompilationContext) { Alias = joinExpression.Alias };
+            // Don't create new alias when adding tables to subquery
+            subquery.AddTable(joinExpression.TableExpression, createUniqueAlias: false);
             subquery.IsProjectStar = true;
             subquery.Predicate = selectExpression.Predicate;
 
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -22,6 +22,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 #endif
 
         private readonly IQuerySqlGeneratorFactory _querySqlGeneratorFactory;
+        private readonly RelationalQueryCompilationContext _queryCompilationContext;
         private readonly List<Expression> _projection = new List<Expression>();
         private readonly List<TableExpressionBase> _tables = new List<TableExpressionBase>();
         private readonly List<Ordering> _orderBy = new List<Ordering>();
@@ -35,22 +36,27 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 
         public virtual Expression Predicate { get; [param: CanBeNull] set; }
 
-        public SelectExpression([NotNull] IQuerySqlGeneratorFactory querySqlGeneratorFactory)
+        public SelectExpression([NotNull] IQuerySqlGeneratorFactory querySqlGeneratorFactory,
+            [NotNull] RelationalQueryCompilationContext queryCompilationContext)
             : base(null, null)
         {
             Check.NotNull(querySqlGeneratorFactory, nameof(querySqlGeneratorFactory));
+            Check.NotNull(queryCompilationContext, nameof(queryCompilationContext));
 
             _querySqlGeneratorFactory = querySqlGeneratorFactory;
+            _queryCompilationContext = queryCompilationContext;
         }
 
         public SelectExpression(
             [NotNull] IQuerySqlGeneratorFactory querySqlGeneratorFactory,
+            [NotNull] RelationalQueryCompilationContext queryCompilationContext,
             [NotNull] string alias)
-            : base(null, Check.NotNull(alias, nameof(alias)))
+            : this(querySqlGeneratorFactory, queryCompilationContext)
         {
-            Check.NotNull(querySqlGeneratorFactory, nameof(querySqlGeneratorFactory));
+            Check.NotNull(alias, nameof(alias));
 
-            _querySqlGeneratorFactory = querySqlGeneratorFactory;
+            // When assigning alias to select expression make it unique
+            Alias = queryCompilationContext.CreateUniqueTableAlias(alias);
         }
 
         public override Type Type => _projection.Count == 1
@@ -62,7 +68,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             Check.NotNull(alias, nameof(alias));
 
             var selectExpression
-                = new SelectExpression(_querySqlGeneratorFactory, alias)
+                = new SelectExpression(_querySqlGeneratorFactory, _queryCompilationContext, alias)
                 {
                     _limit = _limit,
                     _offset = _offset,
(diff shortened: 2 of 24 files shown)


smitpatel:
@anpete @maumar

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryCompilationContext.cs
```diff
@@ -78,5 +82,35 @@ public virtual SelectExpression FindSelectExpression([NotNull] IQuerySource quer
...
+            if (string.IsNullOrEmpty(currentAlias))
+            {
+                return currentAlias;
+            }
+
+            var uniqueAlias = currentAlias;
+            var counter = 0;
+
+            int _;
+            if (int.TryParse(currentAlias.Substring(1), out _))
```
maumar: we could change this to work the same for one letter aliases as well as multi letter aliases, i.e.
currently we have:
a -> a0, a1, a2
bb -> bb0, bb00, bb000

we should aim for 
bb -> bb0, bb1, bb2, bb3

smitpatel: updated.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -179,7 +179,7 @@ var materializer
 
                         selectExpression.Predicate = oldPredicate;
                         selectExpression.RemoveTable(joinExpression);
-                        selectExpression.AddTable(newJoinExpression);
+                        selectExpression.AddTable(newJoinExpression, false);
```
anpete: Use named parameter to clarify false here.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -290,8 +290,9 @@ yield return
         private JoinExpressionBase AdjustJoinExpression(
             SelectExpression selectExpression, JoinExpressionBase joinExpression)
         {
-            var subquery = new SelectExpression(_querySqlGeneratorFactory, joinExpression.Alias);
-            subquery.AddTable(joinExpression.TableExpression);
+            var subquery = new SelectExpression(_querySqlGeneratorFactory, _queryCompilationContext) { Alias = joinExpression.Alias };
+            // Don't create new alias when adding tables to subquery
+            subquery.AddTable(joinExpression.TableExpression, false);
```
anpete: Named parameter.

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -35,22 +36,28 @@ public class SelectExpression : TableExpressionBase
...
             _querySqlGeneratorFactory = querySqlGeneratorFactory;
+            _queryCompilationContext = queryCompilationContext;
         }
 
         public SelectExpression(
             [NotNull] IQuerySqlGeneratorFactory querySqlGeneratorFactory,
+            [NotNull] RelationalQueryCompilationContext queryCompilationContext,
             [NotNull] string alias)
-            : base(null, Check.NotNull(alias, nameof(alias)))
+            : base(null, null)
```
anpete: Is a null alias allowed here now?

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -35,22 +36,28 @@ public class SelectExpression : TableExpressionBase
 
         public virtual Expression Predicate { get; [param: CanBeNull] set; }
 
-        public SelectExpression([NotNull] IQuerySqlGeneratorFactory querySqlGeneratorFactory)
+        public SelectExpression([NotNull] IQuerySqlGeneratorFactory querySqlGeneratorFactory,
+            [NotNull] RelationalQueryCompilationContext queryCompilationContext)
             : base(null, null)
         {
             Check.NotNull(querySqlGeneratorFactory, nameof(querySqlGeneratorFactory));
```
anpete: queryCompilationContext null check?

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryCompilationContext.cs
```diff
@@ -78,5 +84,30 @@ public virtual SelectExpression FindSelectExpression([NotNull] IQuerySource quer
                  select selectExpression)
                     .First();
         }
+
+        public virtual string CreateUniqueTableAlias()
+            => CreateUniqueTableAlias(SystemAliasPrefix);
+
+        public virtual string CreateUniqueTableAlias([NotNull] string currentAlias)
+        {
+            Debug.Assert(currentAlias != null);
```
anpete: Regular Check.NotNull is fine

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryCompilationContext.cs
```diff
@@ -78,5 +84,30 @@ public virtual SelectExpression FindSelectExpression([NotNull] IQuerySource quer
...
+
+            if (string.IsNullOrEmpty(currentAlias))
+            {
+                return currentAlias;
```
anpete: Never reach here due to assert above.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryCompilationContext.cs
```diff
@@ -20,6 +22,10 @@ public class RelationalQueryCompilationContext : QueryCompilationContext
         private readonly List<RelationalQueryModelVisitor> _relationalQueryModelVisitors
             = new List<RelationalQueryModelVisitor>();
 
+        private const string SystemAliasPrefix = "t";
+        private Regex NumericEnding = new Regex("[0-9]*$", default(RegexOptions), TimeSpan.FromMilliseconds(1000.0));
```
anpete: Lets try and do this without Regex. Use Char.IsNumber while iterating backwards.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/NullSemanticsQuerySqlServerTest.cs
```diff
@@ -678,9 +678,9 @@ public override void Join_uses_database_semantics()
             base.Join_uses_database_semantics();
 
             Assert.Equal(
-                @"SELECT [e1].[Id], [e2].[Id], [e1].[NullableIntA], [e2].[NullableIntB]
-FROM [NullSemanticsEntity1] AS [e1]
-INNER JOIN [NullSemanticsEntity2] AS [e2] ON [e1].[NullableIntA] = [e2].[NullableIntB]",
+                @"SELECT [e].[Id], [e0].[Id], [e].[NullableIntA], [e0].[NullableIntB]
```
anpete: It looks like we are losing the behavior of trying to re-use the range variable name. We need to try and preserve this behavior. I.e. Only unique-ify when necessary.

smitpatel:
@anpete - Updated with different approach. I checked few tests to verify that we are preserving user-defined names.

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
smitpatel: Added check.

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
smitpatel: null alias is still not allowed. We unique-fy alias before assigning it therefore we don't use it in direct call to base.

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
```diff
@@ -165,7 +165,7 @@ var predicate
 
             if (predicate != null)
             {
-                var innerSelectExpression = handlerContext.SelectExpressionFactory.Create();
+                var innerSelectExpression = handlerContext.SelectExpressionFactory.Create(handlerContext.QueryModelVisitor.QueryCompilationContext);
```
maumar: break those long lines into 2 separate ones, we usually do something like:

var innerSelectExpression 
   = handlerContext.SelectExpressionFactory.Create(handlerContext.QueryModelVisitor...)

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
```diff
@@ -200,9 +200,9 @@ public override void Join_navigation_in_inner_selector_translated_to_subquery()
                 @"SELECT [e2].[Id], [e1].[Id]
 FROM [Level2] AS [e2]
 INNER JOIN [Level1] AS [e1] ON [e2].[Id] = (
-    SELECT TOP(1) [subQuery0].[Id]
-    FROM [Level2] AS [subQuery0]
-    WHERE [subQuery0].[Level1_Optional_Id] = [e1].[Id]
+    SELECT TOP(1) [subQuery00].[Id]
```
maumar: we can probably change the base generated name from "subquery0" to "subquery", since it's being uniquefied anyway

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryCompilationContext.cs
```diff
@@ -78,5 +84,36 @@ public virtual SelectExpression FindSelectExpression([NotNull] IQuerySource quer
...
         }
+
+        public virtual string CreateUniqueTableAlias()
+            => CreateUniqueTableAlias(SystemAliasPrefix);
+
+        public virtual string CreateUniqueTableAlias([NotNull] string currentAlias)
+        {
+            Check.NotNull(currentAlias, nameof(currentAlias));
+
+            if (string.IsNullOrEmpty(currentAlias))
```
anpete: Redundant check

smitpatel: Its for Empty Check. if currentAlias is empty then we return empty. Is there any other way to check?

smitpatel:
Updated with feedback. Added some tests.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryCompilationContext.cs
```diff
@@ -78,5 +84,36 @@ public virtual SelectExpression FindSelectExpression([NotNull] IQuerySource quer
...
         }
+
+        public virtual string CreateUniqueTableAlias()
+            => CreateUniqueTableAlias(SystemAliasPrefix);
+
+        public virtual string CreateUniqueTableAlias([NotNull] string currentAlias)
+        {
+            Check.NotNull(currentAlias, nameof(currentAlias));
+
+            if (string.Equals(currentAlias, ""))
```
anpete: Or, currentAlias.Length == 0

anpete:
@smitpatel Is it just that not all test changes have been reverted from the earlier approach? Because, if not, it looks like we are still unique-fying too much.

smitpatel:
Updated with 1 more change which reduced changes required.
There are mainly 3 reasons for changes occurring
1. We are unique-fying across query model. Therefore if multiple SQL statements are executed aliases will be unique across all. Which was not the case previously.
2. Subquery aliases - During include processing when we generate a subquery for inner join, previously we used same alias as inner table. We are making it unique now.
3. Client Eval - If the query is using client eval then we visit the query model first time using alias names but fails to generate SQL out of it. So when we process query model again for client eval, they will use different aliases.

Most changes in tests look to be consequence of one of the above. If there is any particular test which seems to be unique-fy extra then point me to it and I will investigate it.

smitpatel:
Update to above:
4. Subquery depth - Previously we used t + subquery depth as alias so that push down subquery gets different alias, now we are unique-fying it so we don't use depth anymore. Which changes it to start from `t` instead of `t0` as before.

anpete:
@smitpatel Thanks for the clear explanation. 1,2 & 4 are OK. Can you see if there is any easy fix to 3? Otherwise :shipit:

smitpatel:
For point 3 I tried looking into avoiding it. But we traverse the queryModel 2 times and both generate different `SelectExpression`. And we don't have any way to know what initiated QueryModelVisitor. The only way to do it without changing a lot of things in background would be to remove all generated aliases if client eval starts but it can have side-effects. So I did not attempt that way.

Issue #4547, opened 2016-02-12
Query :: query source materialized as ValueBuffer rather than entity for some complex cases involving subqueries and conditional operator - causes NRE/InvalidCast


maumar:
The following query will throw during execution:

```
                var query = from e3 in context.LevelThree
                            join e1 in context.LevelOne
                            on
                                (int?)e3.Id
                            equals
                                (
                                    from subQuery2 in context.LevelTwo
                                    join subQuery3 in context.LevelThree
                                    on
                                        subQuery2 != null ? (int?)subQuery2.Id : null
                                    equals
                                        subQuery3.Level2_Optional_Id
                                    into
                                        grouping
                                    from subQuery3 in grouping.DefaultIfEmpty()
                                    select subQuery3 != null ? (int?)subQuery3.Id : null
                                ).FirstOrDefault()
                            select e1.Id;
```

It is because 

```
subQuery3 != null ? (int?)subQuery3.Id : null
```

required subQuery3 to be materialized as Entity, but instead it is materialized as ValueBuffer.

This is the current query plan for the query:

```
(QueryContext queryContext) => IEnumerable<int> _Select(
    source: IEnumerable<TransparentIdentifier<ValueBuffer, ValueBuffer>> _Join(
        outer: IEnumerable<ValueBuffer> ProjectionQuery(
            queryContext: queryContext, 
            entityType: FunctionalTests.TestModels.ComplexNavigationsModel.Level3
        )
        , 
        inner: IEnumerable<ValueBuffer> ProjectionQuery(
            queryContext: queryContext, 
            entityType: FunctionalTests.TestModels.ComplexNavigationsModel.Level1
        )
        , 
        outerKeySelector: (ValueBuffer e3) => (Nullable<int>) (int) object e3.get_Item(0), 
        innerKeySelector: (ValueBuffer e1) => Nullable<int> FirstOrDefault(
            source: IEnumerable<Nullable<int>> _Select(
                source: IEnumerable<TransparentIdentifier<TransparentIdentifier<Level2, IEnumerable<ValueBuffer>>, ValueBuffer>> _SelectMany(
                    source: IEnumerable<TransparentIdentifier<Level2, IEnumerable<ValueBuffer>>> _GroupJoin(
                        outer: IEnumerable<Level2> EntityQuery(
                            queryContext: queryContext, 
                            entityType: FunctionalTests.TestModels.ComplexNavigationsModel.Level2, 
                            key: Key, 
[...]

Issue #4539, opened 2016-02-12
Query :: Join flattening fails for some complex cases involving SelectMany


maumar:
QueryFlattener doesn't work for specific shape of queries - e.g. Join-SelectMany

```
                var query = from l1 in context.LevelOne
                            join l1_Optional in context.LevelTwo on (int?)l1.Id equals EF.Property<int?>(l1_Optional, "Level1_Optional_Id") into grouping
                            from l1_Optional in grouping.DefaultIfEmpty()
                            from l2 in context.LevelTwo
                            join l2_Required_Reverse in context.LevelOne on l2.Level1_Required_Id equals l2_Required_Reverse.Id
                            select new { l1_Optional, l2_Required_Reverse };
```

this throws:

Additional information: Unable to cast object of type 'System.Linq.Expressions.Expression`1[System.Func`3[Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor+TransparentIdentifier`2[Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor+TransparentIdentifier`2[Microsoft.EntityFrameworkCore.Storage.ValueBuffer,System.Collections.Generic.IEnumerable`1[Microsoft.EntityFrameworkCore.FunctionalTests.TestModels.ComplexNavigationsModel.Level2]],Microsoft.EntityFrameworkCore.FunctionalTests.TestModels.ComplexNavigationsModel.Level2],Microsoft.EntityFrameworkCore.Storage.ValueBuffer,Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor+TransparentIdentifier`2[Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor+TransparentIdentifier`2[Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor+TransparentIdentifier`2[Microsoft.EntityFrameworkCore.Storage.ValueBuffer,System.Collections.Generic.IEnumerable`1[Microsoft.EntityFrameworkCore.FunctionalTests.TestModels.ComplexNavigationsModel.Level2]],Microsoft.EntityFrameworkCore.FunctionalTests.TestModels.ComplexNavigationsModel.Level2],Microsoft.EntityFrameworkCore.Storage.ValueBuffer]]]' to type 'System.Linq.Expressions.ConstantExpression'.

[...]

PR #4554, opened 2016-02-13
Initial support for optional navigation translation into LEFT OUTER JOIN.

maumar:
Currently we translate all navigations into INNER JOINs, which may produce invalid results for some queries.

Fix is to translate optional navigations into LEFT OUTER JOIN. In order to do that we convert optional navigation into SelectMany-GroupJoin-DefaultIfEmpty expression, that translates to LOJ.
We also need to add null checks in various places because now the intermittent results can be null.

e.g.
from c in ctx.Customers
where c.Detail.Name == "Foo"
select c

will get translated into:

from c in ctx.Customers
join d in ctx.Detail on c.Id equals d.CustomerId into grouping
from d in grouping.DefaultIfEmpty()
where (d != null ? d.Name : null) == "Foo"
select c;

Also fixed a number of bugs uncovered by this change, due to generating significantly more complex queries in some cases.

Known issues:
- DefaultIfEmpty() cannot be translated into SQL so everything happening after it is computed on the client (filters, projections, nested navigations) (#4588),
- Optional navigations don't work with Include (#4589),
- Optional navigations don't work with some queries involving SelectMany operator (#4539),
- Optional navigations don't work for soem complex queries involving subqueries and/or navigation inside inner key selector of a Join statement (#4547).

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IsNullExpressionBuildingVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IsNullExpressionBuildingVisitor.cs
@@ -9,13 +9,16 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 {
     public class IsNullExpressionBuildingVisitor : RelinqExpressionVisitor
     {
+        private bool _nullConstantAdded = false;
+
         public virtual Expression ResultExpression { get; private set; }
 
         protected override Expression VisitConstant(ConstantExpression node)
         {
-            if (node.Value == null)
+            if (node.Value == null && !_nullConstantAdded)
             {
-                AddToResult(node);
+                AddToResult(new IsNullExpression(node));
+                _nullConstantAdded = true;
             }
 
             return node;
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -48,17 +48,69 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                 IQuerySource querySource,
                 INavigation navigation,
                 JoinClause joinClause,
+                IEnumerable<IBodyClause> additionalBodyClauses,
+                bool optionalNavigationInChain,
+                bool dependentToPrincipal,
+                QuerySourceReferenceExpression querySourceReferenceExpression)
+                : this(
+                      querySource, 
+                      navigation, 
+                      joinClause, 
+                      null, 
+                      additionalBodyClauses, 
+                      optionalNavigationInChain,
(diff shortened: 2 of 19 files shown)


maumar:
@anpete  @divega

maumar:
new version up

maumar:
New version up

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IsNullExpressionBuildingVisitor.cs
```diff
@@ -9,13 +9,16 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
...
     public class IsNullExpressionBuildingVisitor : RelinqExpressionVisitor
     {
+        private bool _nullConstandAdded = false;
+
         public virtual Expression ResultExpression { get; private set; }
 
         protected override Expression VisitConstant(ConstantExpression node)
         {
-            if (node.Value == null)
+            if (node.Value == null && !_nullConstandAdded)
```
anpete: Typo

anpete:
@maumar Might be good if you link the known issues listed above to corresponding bugs.

maumar:
@anpete did that for some issues but I think not for all, will open issues for missing ones

maumar:
@anpete new version up, fixed the typo as well as some sql validation (which was broken on core clr), and added the bug numbers of the known issues in the description

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -214,22 +287,40 @@ var rightNavigationJoin
...
                     else
                     {
-                        newRight = rightNavigationJoin.JoinClause.InnerKeySelector;
+                        newRight = rightJoin?.InnerKeySelector;
                     }
                 }
             }
 
+            // TODO: should we special case other methods? e.g. Equals(object, object)?
+            if (node.NodeType != ExpressionType.ArrayIndex
```
anpete: The hard-coded string.Concat methodinfo seems concerning. Can you elaborate?

maumar: this is to unify the types of binary operation after nav rewrite. During the rewrite types get changed from non-nullable to nullable, e.g.

from c in ctx.Customers
where c.Detail.Id == 5
select c

this compensation needs to happen because c.Detail.Id will be changed to int?.

We take object into account also for cases where property is compared to null constant (which has it's type set to object). 

string.Concat is special cased for cases like:

from c in ctx.Customers
select c.Detail.Name + 1

This will use BinaryOperator with a method (string.Concat), and for this case we don't want to change the type of that second argument because it won't match the method signature.

However I think we can remove the special case for object and this method altogether, currently. This will cause null constant be evaluated on the client, but since everything is evaluated on the client anyway it doesn't make much difference. Once SelectMany-GroupJoin-DefaultIfEmpty is being properly translated to sql, we can re-evaluate this scenatio (we already have tests in place)

anpete:
:shipit:

PR #4575, opened 2016-02-16
Fix #3704 - Explicit separation of command construction and execution

### LINKED ISSUE #3115: Support passing DbParameter instances to raw SQL APIs

### LINKED ISSUE #3704: RelationalCommand: stricter separation of construction and execution

mikary:
Fix #3115 - Support passing DbParameter instances to ExecuteSqlCommand
and FromSql

I've made the changes we discussed in the meeting / email.

I'm still not happy with the name for the parameter value caching version of the builders / commands and leaning more towards going back to a single builder that doesn't cache at all.

@anpete @divega


mikary:
Updated to remove command and builder version that caches parameter values.

src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
```diff
@@ -244,6 +244,15 @@
   <data name="ClientEvalWarning" xml:space="preserve">
     <value>The LINQ expression '{expression}' could not be translated and will be evaluated locally.</value>
   </data>
+  <data name="MissingParameterValue" xml:space="preserve">
+    <value>No value provided for requred parameter '{parameter}'.</value>
```
divega: Typo on "required".

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1125,14 +1122,18 @@ protected override Expression VisitParameter(ParameterExpression expression)
...
             Check.NotNull(expression, nameof(expression));
 
-            object value;
-            if (_parametersValues.TryGetValue(expression.Name, out value))
-            {
-                var name = _sqlGenerationHelper.GenerateParameterName(expression.Name);
+            var name = _sqlGenerationHelper.GenerateParameterName(expression.Name);
 
-                _relationalCommandBuilder.AppendParameter(name, value, expression.Type, expression.Name);
+            if (_relationalCommandBuilder.ParameterBuilder.Parameters.All(p => p.InvariantName != expression.Name))
```
divega: I am not sure why this part changed but isn't this potentially a hot path? The LINQ query over all parameters could be expensive.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -51,10 +51,15 @@ public static void Migrate([NotNull] this DatabaseFacade databaseFacade)
...
 
-                return databaseFacade
+                var command = databaseFacade
                     .GetService<IRawSqlCommandBuilder>()
-                    .Build(sql, parameters)
-                    .ExecuteNonQuery(GetRelationalConnection(databaseFacade));
+                    .Build(sql, parameters);
+
+                return command
+                    .Item1
```
divega: I am not too fond of using tuples for this and would consider defining a type for this instead. Let's hear what others think. @anpete?

src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalCommand.cs
```diff
@@ -16,31 +16,35 @@ public interface IRelationalCommand
 
         int ExecuteNonQuery(
             [NotNull] IRelationalConnection connection,
+            [NotNull] IReadOnlyDictionary<string, object> parameterValues = null,
```
divega: Isn't the [NotNull] attribute contradicting that we have a default? Did you intend to remove the default value?

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
mikary: This was just a factoring change. The logic already existed in `RelationalCommand`:
```cs
        public virtual void AddParameter(IRelationalParameter relationalParameter)
        {
            Check.NotNull(relationalParameter, nameof(relationalParameter));

            if (relationalParameter.InvariantName == null
                || _parameters.All(p => p.InvariantName != relationalParameter.InvariantName))
            {
                _parameters.Add(relationalParameter);
            }
        }
```

This isn't the main hot path scenario, this is only hit when we have a cache miss on the command cache.

The previous implementation ignored duplicate definitions by default. This seems like a dangerous assumption and moving the logic here revealed an issue with the `ReaderModificationCommandBatchTest.CreateStoreCommand_creates_parameters_for_each_ModificationCommand` test.

src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalCommand.cs
mikary: The default is correct, the value should be `[CanBeNull]` since it is optional for parameterless commands.

mikary:
Updated for `FromSql` command caching fixes.

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
```diff
@@ -45,14 +45,25 @@ public override bool Equals(object obj)
                             return false;
                         }
 
-                        if (value is IEnumerable
-                            && value.GetType() != typeof(string)
-                            && value.GetType() != typeof(byte[]))
+                        if(value is IEnumerable)
```
anpete: formatting

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
```diff
@@ -45,14 +45,25 @@ public override bool Equals(object obj)
...
                         {
-                            // TODO: This doesn't always need to be deep.
-                            // We could add a LINQ operator parameter attribute to tell us.
-                            return StructuralComparisons
-                                .StructuralEqualityComparer.Equals(value, otherValue);
+                            var type = value.GetType();
+
+                            if(type == typeof(object[]))
+                            {
+                                // Only compare lengths for FromSql parameters
```
anpete: Nit: Core can't really know about FromSql here.

src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalParameterList.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace Microsoft.EntityFrameworkCore.Storage
+{
+    public interface IRelationalParameterList
```
anpete: I'm not sure this type is necessary. It also doesn't look like a traditional list type.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
anpete: Agree.

mikary: Would we rather have glorified tuple (a poco with properties for the `RelationalCommand` and the `IDictionary<string, object>`) or more of a wrapper / facade?

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
mikary: True, but it would be misleading to say this was here for any other reason.

I wonder if it would be better in the long run to use some kind of parameter metadata to determine how to compare / cache.

src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalParameterList.cs
mikary: It's a bit strange, but this class exists for the sake of API surface and `CompositeRelationalParameter`s. While this is more of an internal/provider level API, it might be worth having a design meeting/discussion about the options and patterns here.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
divega: I would suggest creating the simplest type that could work (maybe the POCO), try to find a good name for it and then looking at it again to see if it encapsulation would be compelling. There are no hard rules. You have to use some refactoring instinct :smile:

divega: Also, is this the only place that returns a command and a list of parameters?

mikary: This only really happens in two places, here and in the `ReaderModificationCommandBatch`. It might be useful in migrations in the future, but we currently don't use parameters in any of the commands there.

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
mikary: On second thought, this isn't core. We don't have the concept of Commands outside of Relational.

anpete: *nothing to see here*

mikary:
Updated from feedback and discussion

anpete:
:shipit:

Issue #4549, opened 2016-02-12
ExpressionEqualityComparer NullReferenceException


EricABC:
Some low-hanging fruit here, nameless parameters throw a NullReferenceException on `ParameterExpression.Name.GetHashCode();` inside `ExpressionEqualityComparer` line 101 when dynamically building expression trees (ie. using 'Expression.Parameter(typeof(something))' without a parameter name).

rowanmiller:
@EricABC is this manifesting in some way when you are using EF? Or just a theoretical thing that could be hit based on looking at the code?

EricABC:
Yup, it did in fact occur for me.  I have a sparse-matrix model with dynamic stored schema where I build a set of property filters on a query dynamically.  On one of my properties didn't have a name and that was the culprit.  As soon as I named it the problem was solved for me and everything worked like a charm.  Admittedly I have a few customizations in place, including an implementation for https://github.com/aspnet/EntityFramework/issues/242 that inherits off of a few internal namespaced items (just serializes stuff in shadow properties), but none of it would have effected the call path to caching the query, which is what ultimately gets the hashcode.  If you prefer I can get you an unadulterated repro.

EricABC:
Repro at https://github.com/EricABC/EFBugReproIssue4549

maumar:
fixed in da1a1a839f869564482d37b519c52e3b2b95b1c6

PR #4576, opened 2016-02-16
[TINY] Fix to #4549 - ExpressionEqualityComparer NullReferenceException

maumar:


maumar:
@anpete

divega:
Is the indentation actually wrong or is it just the diff? Why didn't this fail in any of our tests? Should we add one?

Otherwise :shipit:

anpete:
:shipit:

maumar:
Fixed the indentation. We don't have the test for, but I don't think we really need one. This only repros if you manually create ParameterExpression using the Expression builder API. I will take a look at other expressions though to see if there are more NREs like this.

Issue #4590, opened 2016-02-17
ToEntityEntry 


meikeric:
I realize this is an edge use case, but none the less, it is one I am up against.

I am writing a custom dynamic entity tracker/state manager, essentially a dynamic JSON Object -> database implementation. It leverages EntityFramework 7 since you guys have done the heavy lifting.  However, I've run into an issue which requires a concrete implementation of EntityEntry instead of an interface which ends up being tightly coupling some pieces because of one interface method: 

"src/Microsoft.EntityFrameworkCore/Update/IUpdateEntry.cs" requires "ToEntityEntry()" for "Microsoft.EntityFrameworkCore/DbUpdateException.cs".

Can DbUpdateException rely on IUpdateEntity instead of EntityEntry, thus removing the need for the concrete implementation of EntityEntry and ToEntityEntry?

EntityEntry is pretty heavy and to have to write a decent amount of code to return a new EntityEntry via ToEntityEntry just for an exception message seems extreme.  It's declared once in IUpdateEntry.cs, it's implemented once in InternalEntityEntry.cs, and it's invoked once in DbUpdateException.cs.  I can't find **anywhere** in the repo that DbUpdateException.Entries is used, therefore I don't foresee an impact.

EDIT: Looking through the code, I don't see why a few changes wouldn't resolve the tight coupling.

Line 20:
`private readonly LazyRef<IReadOnlyList<EntityEntry>> _entries;`
becomes
`private readonly LazyRef<IReadOnlyList<IUpdateEntry>> _entries;`

Line 30:
`_entries = new LazyRef<IReadOnlyList<EntityEntry>>(() => new List<EntityEntry>());`
becomes
`_entries = new LazyRef<IReadOnlyList<IUpdateEntry>>(() => new List<IUpdateEntry>());`

Line 59:
`_entries = new LazyRef<IReadOnlyList<EntityEntry>>(() => entries.Select(e => e.ToEntityEntry()).ToList());`
becomes
`_entries = new LazyRef<IReadOnlyList<IUpdateEntry>>(entries);`

Line 66:
`public virtual IReadOnlyList<EntityEntry> Entries => _entries.Value;`
become
`public virtual IReadOnlyList<IUpdateEntry> Entries => _entries.Value;`

meikeric:
I went ahead and created a pull request. https://github.com/aspnet/EntityFramework/pull/4592

rowanmiller:
[...]

PR #4592, opened 2016-02-18
Changed InternalEntityEntry to IUpdateEntry where logical.

meikeric:
Made OriginalValues and StoredGeneratedValues structs public.
Added indexer to IUpdateEntry

This is so that custom Entity tracking can be created.


rowanmiller:
Closing per discussion on the related issue (#4590). 

>  DbUpdateException.Entries  is used by application code when it needs to resolve a database exception. For example, if there is a concurrency exception, then it can use the data and operations from  EntityEntry  to show the user the different values and allow them to resolve the issues.

> IUpdateEntry  is the API designed for database providers to persist changes to the database. 

> Given these two APIs are designed for completely different audiences it does not make sense to change them.

We are willing to consider a different approach to enabling the scenario in question. We can use the issue to discuss this.

PR #4598, opened 2016-02-18
Query: Relational: Include: Compilation and SQL improvements/fixes.

anpete:
- Includes now composed over query function at top-level like for in-memory. Simplifies ET rewrite and result accessing logic. Improves Include performance for client-eval scenarios.
- Introduces simpler and more robust SQL translation for collection Include. In particular, changes from inner join subquery to left outer join so that we can return a row for each parent even when no child rows exist. This allows for correct result streaming in client-eval scenarios.
- Some refactoring of the large Include compiler method in IncludeExpressionVisitor.
- Renamed OuterJoin in SQL ETs to LeftOuterJoin.


anpete:
@maumar

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
```diff
@@ -196,59 +204,75 @@ public override void Include_multiple_include_then_include()
...
     WHERE [g].[Discriminator] IN ('Officer', 'Gear')
-) AS [c] ON [g].[CityOrBirthName] = [c].[Name]
-LEFT JOIN [CogTag] AS [c0] ON ([c0].[GearNickName] = [g].[Nickname]) AND ([c0].[GearSquadId] = [g].[SquadId])
-WHERE ([g].[Discriminator] = 'Officer') OR ([g].[Discriminator] = 'Gear')
-ORDER BY [c].[Nickname], [c].[Name]",
+) AS [t0]
+LEFT JOIN (
+    SELECT [g].*
+    FROM [Gear] AS [g]
+    WHERE ([g].[Discriminator] = 'Officer') OR ([g].[Discriminator] = 'Gear')
```
divega: A bit OT but I wonder why some queries have `[g].[Discriminator] IN ('Officer', 'Gear')` and some others `([g].[Discriminator] = 'Officer') OR ([g].[Discriminator] = 'Gear')`

anpete: cc @maumar perhaps a bug in EqualityPredicateInExpressionOptimizer?

divega: @maumar @anpete please help me file a good bug on this :smile:

anpete: #4603

maumar:
One issue I found when playing around with this pull request, we are currently broken for cases with GroupJoin if the key selector is a property wrapped in convert. We will generate invalid SQL, trying to order by the same key twice.

This happens because we produce two ordering (int?)c.Id and c.Id (which both get translated to the same sql), and our check for duplicates doesn't catch it

The simple fix could be to modify OuterJoinOrderingExtractor, to filter out the converts:

```
            protected override Expression VisitUnary(UnaryExpression node)
            {
                if (node.NodeType == ExpressionType.Convert)
                {
                    return Visit(node.Operand);
                }

                return base.VisitUnary(node);
            }
```

anpete:
I need to do a bit more work on this to address the GroupJoin issue.

cc @maumar

Issue #4577, opened 2016-02-16
CLI Commands: Support .NET Framework-only projects


bricelam:
Currently, in order to use the .NET Core CLI EF commands, your project must target a .NET Core-compatible framework. We could remove this requirement by starting a new net451 process to execute the commands in this case.

~~Currently **blocked** on dotnet/cli#2126~~

### Known Issues

- [ ] Cannot use dotnet-ef on net451 class library project https://github.com/dotnet/cli/issues/2919
```
No executable found matching command "Microsoft.EntityFrameworkCore.Tools.Cli"
```
- [ ] dotnet-ef cannot load the startup assembly. See below https://github.com/aspnet/EntityFramework/issues/4577#issuecomment-217496222

```
System.IO.FileNotFoundException: Could not load file or assembly 'StartupProj' or one of its dependencies. The system cannot find the file specified. File name: 'StartupProj'
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint,
```

divega:
> your project must target a .NET Core-compatible framework

Does that mean they run fine in .NET 4.6 or not?

bricelam:
No, the user's assembly needs to be compiled  against the .NET Core contracts.

ErikEJ:
But all the commands work in the PMC, I assume?

rowanmiller:
@ErikEJ correct. PMC will light up for xproj in RC2

bricelam:
Yes, if you want to target only .NET 4.5.1, using the *.csproj + NuGet commands will work as usual.

rowanmiller:
For folks using VS, the primary way to invoke these commands is Package Manager Console (which will be enabled in RC2). That means for the developer workflow, this would only be needed for someone developing against Full .NET and wanting to use the command line tools. This seems like it will be a small set of folks.

One thing we didn't consider in triage though is that this could be very useful for build/deployment integration.

rowanmiller:
Discussed and decided that our general purpose CLI that runs on assemblies (https://github.com/aspnet/EntityFramework/issues/646) rather than our .NET CLI specific commands is the thing that will be best suited to build/deployment/etc. Given that, closing this out but we can reconsider if we see customers wanting this in real applications.

bricelam:
Re-opening. It looks like this will happen as part of dotnet/cli#2020

rowanmiller:
[...]

PR #4602, opened 2016-02-18
CLI Commands: Automatically build the project

bricelam:
Part of #3925

Additional changes:
* Allow `--verbose` anywhere
* Log stack traces at a higher verbosity level
* Log project information
* Throw on incompatible frameworks (see #4577)


src/dotnet-ef/OperationExecutor.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            Reporter.Verbose.WriteLine(("Using startup project '" + startupProject + "'.").Bold().Black());
+            Reporter.Verbose.WriteLine(("Using project '" + project + "'.").Bold().Black());
+
+            var startupProjectContext = GetCompatibleProjectContext(startupProject);
+
+            Reporter.Verbose.WriteLine("Build started...".Bold().Black());
+
+            var buildCommand = Command.Create(
+                Path.Combine(Environment.GetEnvironmentVariable("DOTNET_HOME"), "bin", "dotnet" + Constants.ExeSuffix),
```
bricelam: @davidfowl Good enough?

bricelam: Doh! `DOTNET_HOME` is being removed in dotnet/cli#1478

src/dotnet-ef/Program.cs
```diff
@@ -61,6 +64,20 @@ public static int Main([NotNull] string[] args)
...
         }
 
+        private static void HandleVerboseOption(ref string[] args)
+        {
+            for (int i = 0; i < args.Length; i++)
+            {
+                if (args[i] == "-v" || args[i] == "--verbose")
+                {
+                    Environment.SetEnvironmentVariable(CommandContext.Variables.Verbose, bool.TrueString);
+                    args = args.Take(i).Concat(args.Skip(i + 1)).ToArray();
```
lajones: If you find a `-v` or `--verbose` at say, `i = 4`, then having re-calculated `args` wouldn't you need to start at `i = 4` again in order to look at whatever was originally at `i = 5` but now is at `i = 4` instead of just carrying on with the `i++` in the loop?

src/dotnet-ef/OperationExecutor.cs
lajones: Seems like we should raise an error if `DOTNET_HOME` is not defined?

src/dotnet-ef/OperationExecutor.cs
```diff
@@ -0,0 +1,177 @@
...
+                new[] { "build", startupProject, "-f", startupProjectContext.TargetFramework.GetShortFolderName() });
+            if (buildCommand.Execute().ExitCode != 0)
+            {
+                throw new OperationException("Build failed.");
+            }
```
lajones: Add the exit code to the message.

src/dotnet-ef/OperationExecutor.cs
```diff
@@ -0,0 +1,177 @@
...
+{
+    public class OperationExecutor
+    {
+        private readonly LazyRef<DbContextOperations> _contextOperations;
+        private readonly LazyRef<DatabaseOperations> _databaseOperations;
+        private readonly LazyRef<MigrationsOperations> _migrationsOperations;
+
+        public OperationExecutor([CanBeNull] string startupProject, [CanBeNull] string environment)
+        {
+            var project = Path.Combine(Directory.GetCurrentDirectory(), Project.FileName);
```
lajones: Am I misunderstanding? It looks like you are expecting the project to be one level down from the current directory? But I would hope we could run a command from any directory we like and just pass in the project directory to which we want to point?

src/dotnet-ef/Program.cs
bricelam: The loop breaks after this. We find the first one, process and remove it, then stop

src/dotnet-ef/OperationExecutor.cs
bricelam: Useless information. The user can manually build their project to see the errors.

bricelam: (The experience is more-or-less the same in VS using the NuGet commands)

src/dotnet-ef/OperationExecutor.cs
bricelam: You currently run the tool in the project's directory (i.e. the one containing the `project.json`). We could add a `--project` parameter if we wanted. It just seemed YAGNI to me. cc @rowanmiller

bricelam: Oh, I remember why I didn't. `dotnet.exe` can only find the `dotnet-ef` if you run it in the same directory as the project.

src/dotnet-ef/OperationExecutor.cs
lajones: I don't know - a message that just says "Build failed." doesn't seem very useful. Maybe include what the user needs to do to see the errors?

src/dotnet-ef/OperationExecutor.cs
bricelam: Sounds like we're going to assume it's on the path. (see dotnet/cli#1150) I'll update the PR.

src/dotnet-ef/OperationExecutor.cs
rowanmiller: I think it's ok as-is. This is the same experience we have in VS. If the build fails I think it's pretty intuitive that you build the project to see errors.

src/dotnet-ef/OperationExecutor.cs
lajones: OK. As long as the error message you get if you're in the wrong directory is helpful.

src/dotnet-ef/OperationExecutor.cs
lajones: OK. Seems like a small helpful thing we could do for the user to me - but no big deal.

src/dotnet-ef/Program.cs
lajones: OK. Sounds fine so long as if there are multiple `-v` or `--verbose` flags it doesn't cause some strange error later on.

src/dotnet-ef/OperationExecutor.cs
bricelam: It's more-or-less useful. `dotnet.exe` reports the following.
> No executable found matching command "dotnet-ef"

bricelam: lol, it's no worse than what DNX did. :wink: 
> Error: Unable to load application or execute command 'ef'.

src/dotnet-ef/Program.cs
bricelam: They're swallowed gracefully by the parser. :cocktail:

src/dotnet-ef/OperationExecutor.cs
lajones: Yeah - I've seen that error message and not been clear on what to do myself. If it's easy for us to add something that says "you need to be in the project directory to run this command" I'd do that. In general I think we should aim for a higher level of usefulness on the error messages than was typically provided by DNX :wink:

src/dotnet-ef/Program.cs
lajones: OK. Sounds good.

lajones:
:shipit:

PR #4609, opened 2016-02-19
Fix #4603 - Ensure we optimize predicates of collection Include queries.

### LINKED ISSUE #4603: Collection Include queries not optimized
We need to run CompositePredicateExpressionVisitorFactory over them.

anpete:


divega:
:shipit:

Issue #4458, opened 2016-02-02
Investigate Sql statements ordering in tests


smitpatel:
For tests
`Where_nav_prop_reference_optional1_via_DefaultIfEmpty` & `Where_nav_prop_reference_optional2_via_DefaultIfEmpty`
The generated SQL has 2 statements. Test on coreclr and full desktop has different ordering of the SQL statements. Disabled SQL assertion till investigated.

rowanmiller:
If ordering doesn't matter then we should make the test pass with either order

maumar:
This is due to Enumerable.Join having different implementation between coreclr and net451. Filed a bug on corefx team to track this: https://github.com/dotnet/corefx/issues/6477

JonHanna:
Hi. In case anyone isn't following the corefx issue, I'm arguing there we should keep the difference (it allows us to short-circuit on some cases where the results of the join are empty). It seems that the issue it caused you was purely with testing rather than EF itself, and as such #4611 is all you need.

If someone can think of a reason why I'm wrong on that, please visit https://github.com/dotnet/corefx/issues/6477.

maumar:
this is due to optimization done in corefx - source collections are executed in different order now to allow some short circuiting. The discrepancy is by design and we just need to be mindful in out tests.

PR #4611, opened 2016-02-20
Fix build break

smitpatel:
There are further tests being affected by #4458 breaking the build.
Converted Assert.Equal to Assert.Contains to verify SQL ignoring the ordering of statements

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -2212,11 +2212,14 @@ INNER JOIN [Orders] AS [o] ON ([c].[CustomerID] = [o].[CustomerID]) AND ([c].[Cu
         {
             base.Join_client_new_expression();
 
-            Assert.Equal(
+            // See issue#4458
+            Assert.Contains(
                 @"SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
-FROM [Orders] AS [o]
+FROM [Orders] AS [o]",
+                Sql);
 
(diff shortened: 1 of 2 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -3363,25 +3386,25 @@ public override void Projection_when_arithmetic_expression_precendence()
...
-//        public override void Projection_when_arithmetic_mixed()
-//        {
-//            //base.Projection_when_arithmetic_mixed();
-//        }
-//
-//        public override void Projection_when_arithmetic_mixed_subqueries()
-//        {
-//            //base.Projection_when_arithmetic_mixed_subqueries();
-//        }
+        //        public override void Projection_when_arithmetic_expressions()
```
divega: Is this change intentional? Do you know why is it commented out?

divega:
:shipit:

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
smitpatel: It was already commented out. TODO note above it says complex projection translation.
Autoformat on file changed formatting. I will revert it so the commit does not mislead in future.

maumar:
Interesting. I ran full build when working on LOJ translation and fixed all tests that were affected for me. Possibly the issue is intermittent?

smitpatel:
Doesn't look like intermittent. Once I have seen failure, it always failed for future runs. (also check CI history).

divega:
@smitpatel @maumar How hard would it be to get a minimal repro of the behavior difference? I would like us to pass this along as an issue for the CoreCLR team. I know Maurycy started but not sure how far along he got.

PR #4620, opened 2016-02-22
Parameter logging now logs the value of byte arrays

### LINKED ISSUE #3670: Logging: Binary parameter values logged as System.Byte[]

ErikEJ:
Fixes #3670

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalLoggerExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalLoggerExtensions.cs
@@ -38,14 +38,14 @@ namespace Microsoft.EntityFrameworkCore.Storage
                             command.CommandTimeout,
                             command.Parameters
                                 .Cast<DbParameter>()
-                                .ToDictionary(p => p.ParameterName, p => logParameterValues ? FormatParameterValue(p.Value) : "?"),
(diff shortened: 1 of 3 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalLoggerExtensions.cs
```diff
@@ -37,7 +38,7 @@ var logParameterValues
                             command.CommandTimeout,
                             command.Parameters
                                 .Cast<DbParameter>()
-                                .ToDictionary(p => p.ParameterName, p => logParameterValues ? p.Value : "?"),
+                                .ToDictionary(p => p.ParameterName, p => logParameterValues ? FormatParameterValue(p.Value) : "?"),
```
anpete: Parameter values need to stay strongly typed here. Move the formatting to the formatter delegate.

ErikEJ: Moved!

ErikEJ:
Call to Format... moved

PR #4626, opened 2016-02-23
Add high-level cancellation token check to async queries

### LINKED ISSUE #4317: ToListAsync() is not handling cancellation correctly.
With Sqlite database provider and QueryTrackingBehavior set to NoTracking, ToListAsync() does not handle cancellation correctly. On cancellation, it should either throw TaskCanceledException or complete the operation (i.e., ignore cancellation). But, it instead succeeds returning a partial list of records.

Test code attached.
[NoTrackingTests.zip](https://github.com/aspnet/EntityFramework/files/91552/NoTrackingTests.zip)

natemcmaster:
This ensures partial results are never returned, as was found to be the case in #4317.

Close #4317

cc @anpete

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
@@ -127,6 +127,8 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                     try
                     {
                         _exceptionInterceptor._queryContext.ConcurrencyDetector.EnterCriticalSection();
+                        // TODO remove this when/if bug is resolved in Ix-Async https://github.com/Reactive-Extensions/Rx.NET/issues/166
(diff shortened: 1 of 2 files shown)


natemcmaster:
FWIW I can't find a bug in Ix-Async

divega:
@natemcmaster does the strange cancellation behavior repro at all with Ix-Async?

natemcmaster:
@divega I've been trying to repro the same partial-results-returned behavior w/o success

natemcmaster:
@anpete I can find the source of the bug in ix-async or our code. The high-level cancellation token check solves the issue. Are you okay with this even though we don't know why get the the async bug?

anpete:
I want to take one more look at this before we do the workaround

anpete:
OK, filed Rx.NET#166

@natemcmaster :shipit: on the workaround.

natemcmaster:
I'll add a TODO linking to the issue before I merge

Issue #3076, opened 2015-09-09
Reverse Engineer: Can we just scaffold string properties for `xml`, `hierarchyid`, `geography`, `geometry`, etc


rowanmiller:
Some of these datatypes may just work with an implicit conversion to/from string. We need to investigate which ones work (or can work with minimal changes) and then update reverse engineer to scaffold a string property for them.

nteague22:
I do know that stored procedures as of Sql Server 2012 can accept string ADO parameter values when the parameter type is marked as HierarchyId, is there a simple way to override and create a new derivative of SqlOperation that extends these checks?

bricelam:
@nteague22 It may require a few changes, but we should be able to set the parameter type using the [SqlServerTypeMapper](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.SqlServer/Storage/Internal/SqlServerTypeMapper.cs).

nteague22:
Excellent, I really am impressed with the work you guys produce and the team as a whole!  Even in beta, EF7 makes my life easier on running ETL and other data migratory operations

wangkanai:
How is issue going? I still that the https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs is unchanged.
Hoping to see this feature add by rc1 :)

bricelam:
@wangkanai Have you tried manually adding the properties to your model?

wangkanai:
@bricelam  Do you mean like mapping a object property [System.Data.Spatial] (https://msdn.microsoft.com/en-us/library/system.data.spatial(v=vs.110).aspx) to [Spatial Data Types](https://msdn.microsoft.com/en-us/library/bb964711.aspx) in demo application? **No** :smile:, don't know where to start, still studying your codebase.
Or add [RelationalTypeMapping](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Storage/RelationalTypeMapping.cs) to your [SqlServerTypeMapper.cs](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.MicrosoftSqlServer/Storage/Internal/SqlServerTypeMapper.cs) class?
```
private readonly RelationalTypeMapping _geography = new RelationalTypeMapping("geography");
private readonly RelationalTypeMapping _geometry = new RelationalTypeMapping("geometry");
```
[...]

PR #4629, opened 2016-02-23
[SMALL, RevEng] Map SQLServer xml columns to string properties.

lajones:
Partial fix for #3076. Update RevEng SQL Server provider to map xml columns to string properties.


natemcmaster:
Have we considered the [limitations of xml data type](https://msdn.microsoft.com/en-us/library/hh403385.aspx#Anchor_4)? This simple PR works from database => model. Does EF handle xml columns types the other way?

lajones:
@natemcmaster Yes - I did some testing and EF handles insert, update and delete going the other way. You can get into problems if you try to store some (but not all) non-XML e.g. `">'<"` causes problems but just `"someText>"` does not. You get a `DbUpdateException` which explains (in the inner exception) that it had trouble parsing the string as XML. Discussed these limitations with @rowanmiller and decided they were OK.

rowanmiller:
Yeah I've done some playing around with it for demos etc. and it works for the mainline cases just fine.

natemcmaster:
Ok. Changes here look fine then. :shipit:

lajones:
Checked in with commit a8b185f.

PR #4631, opened 2016-02-23
Commands: Cleanup parameters

bricelam:
@lajones

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/Design/OperationExecutor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Commands/Design/OperationExecutor.cs
@@ -35,12 +35,13 @@ namespace Microsoft.EntityFrameworkCore.Design
             var projectDir = (string)args["projectDir"];
             var rootNamespace = (string)args["rootNamespace"];
 
-            var startupAssembly = Assembly.Load(new AssemblyName(startupTargetName));
+            var assemblyLoader = new AssemblyLoader();
+            var startupAssembly = assemblyLoader.Load(startupTargetName);
 
             Assembly assembly;
             try
             {
-                assembly = Assembly.Load(new AssemblyName(targetName));
+                assembly = assemblyLoader.Load(targetName);
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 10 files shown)


src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
```diff
@@ -396,10 +396,10 @@ function Scaffold-DbContext {
         [string] $Connection,
         [Parameter(Position = 1, Mandatory = $true)]
         [string] $Provider,
-        [string] $OutputDirectory,
-        [string] $ContextClassName,
-        [string[]] $Schemas,
-        [string[]] $Tables,
+        [string] $OutputDir,
+        [string] $Context,
+        [string[]] $Schemas = @(),
```
lajones: Is this correct? Don't you have to reference the Schemas `.PARAMETER` in some way? Similarly for Tables below.

src/Microsoft.EntityFrameworkCore.Commands/Design/OperationExecutor.cs
```diff
@@ -290,8 +290,8 @@ public ReverseEngineer([NotNull] OperationExecutor executor, [NotNull] object re
                 var provider = (string)args["provider"];
                 var outputDir = (string)args["outputDir"];
                 var dbContextClassName = (string)args["dbContextClassName"];
-                var schemaFilters = (string[])args["schemaFilters"] ?? new string[0];
-                var tableFilters = (string[])args["tableFilters"] ?? new string[0];
+                var schemaFilters = (IEnumerable<string>)args["schemaFilters"];
+                var tableFilters = (IEnumerable<string>)args["tableFilters"];
```
lajones: Does this work find with the new command line? What happens if a user passes no `-s` or `-t` options on the command line?

bricelam: Yes. `CommandOption.Values` will return an empty list if none are specified.

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
bricelam: I'm not following... The only change here is to give them a default value of `@()` which is PowerShell for an empty array. That way if they're not specified, they're empty instead of null.

lajones: Nm. I was mis-reading it.

src/Microsoft.EntityFrameworkCore.Commands/Design/OperationExecutor.cs
lajones: OK. Sounds good.

lajones:
:shipit:

PR #4550, opened 2016-02-12
Move build to dotnet

ryanbrandenburg:

PR #4638, opened 2016-02-25
Use dotnet-test

### LINKED ISSUE #4495: Full .NET xunit is not running ConditionalTheory and ConditionalFact tests
When we transitioned to dotnet cli, we started using `xunit.console.exe` to run full .NET tests. Currently, this is not finding tests marked with `[ConditionalTheory]` or `[ConditionalFact]`.

Example: Microsoft.EntityFrameworkCore.Commands.FunctionalTests should have 109 tests, but only 98 are discovered. The [tests](https://github.com/aspnet/EntityFramework/files/118126/tests.txt) should in include `OperationExecutorTest`, but they are not discovered.

Related: https://github.com/aspnet/Testing/issues/203 (EF implements its own discoverers, but problably requires the same fix Testing needs.)

natemcmaster:
Supercedes #4509 and #4550 

VS still doesn't like this. :cry: 

Fix #4495 

cc @bricelam @Eilon


test/Microsoft.EntityFrameworkCore.Commands.FunctionalTests/Design/Internal/ForwardingProxyTest.cs
```diff
@@ -1,11 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-#if DNX451
+#if (NET451 || DNX451)
```
pranavkm: Shouldn't need the parantheses.

natemcmaster:
@bricelam fixed the tests that use roslyn to compile. Everything suceeds on dotnet-test

test/Microsoft.EntityFrameworkCore.Commands.FunctionalTests/Design/Internal/ForwardingProxyTest.cs
natemcmaster: True, but I left it deliberately so it will be easy to search & replace when we strip out dnx451.

natemcmaster:
To the concern of VS tooling: I found how to launch tests from VS. But it is about as ugly as the current command-line setup we have.

Here is what it takes (on my machine...haven't tried elsewhere.)

- Remove "net451" from all project.json (source and test)
- dotnet-restore from command line. DO NOT restore packages from VS. This will fail.
- Unload all except the test project you want to work on. (Test Discovery has a bug that starts test discovery on the wrong projects in the wrong order.)
- Build the test project. (This will fail.) But test discovery for Test Explorer should start anyways. (See Output > Tests" to see if it works.)
- If tests are discovered, you can run them by selecting the list and *rightclick* > Run Selected Tests. (Don't "Run all". This tries to build which fails and cancels the test run.)

smitpatel:
@natemcmaster - I just ran `dotnet restore` on command line and build solution in VS. Build fails as usual but all tests are discovered.

natemcmaster:
@smitpatel Hmm. Weird. Which version of ASP.NET Web tooling? I have 
`ASP.NET and Web Tools 2015 (RC1 Update 1)   14.1.20203.0` (Help > About MS VS)

smitpatel:
`RC1 Update 1 - 14.1.11120.0` I am behind!

natemcmaster:
Well, good news. If you didn't update with the latest "greatest" version of WTE, then these changes shouldn't break you too bad.

natemcmaster:
@divega @bricelam despite our best efforts, VS will only sort-of work. We need to push this through anyways as the netstandard rename is coming along anyways. Can we merge?

PR #4654, opened 2016-02-26
Convert relational annotations to fluent api in model snapshot

### LINKED ISSUE #4382: ModelSnapshot is different after migration is added and then removed
While looking at code that was about to get checked in on my project I noticed that the ApplicationDbContextModelSnapshot had changed, even though I had reverted a change I thought I wanted to make.

Specifically I added a migration with some changes to an existing entity, but didn't like them so I used `dnx ef database update [prior_migration]` and `dnx ef migrations remove`  When I was checking out Team Explorer I noticed the differences were there.

Not sure it matters, because they seem functionally equivalent, but also unclear why they should be different when the model state is the same and wanted to bring it up in case something might cause issues in a more general release to broader audience.

The changes occur in *all entities* in the model (not just the one I tried to change) and looks something like this:

       modelBuilder.Entity("MyApp.Models.AnswerOption", b =>
                {
                    b.ToTable("AnswerOption");

                    b.Property<int>("Id")
                        .ValueGeneratedOnAdd();

                    b.Property<int?>("QuestionId");

                    b.Property<string>("Text");

                    b.HasKey("Id");

                    b.HasIndex("QuestionId");
                });

to

            modelBuilder.Entity("MyApp.Models.AnswerOption", b =>
                {
                    b.Property<int>("Id")
                        .ValueGeneratedOnAdd();

                    b.Property<int?>("QuestionId");

                    b.Property<string>("Text");

                    b.HasKey("Id");

                    b.HasIndex("QuestionId");

                    b.HasAnnotation("Relational:TableName", "AnswerOption");
                });

the difference being that it initially uses `b.ToTable("AnswerOption");` but after the migrate/remove migrate uses `b.HasAnnotation("Relational:TableName", "AnswerOption");`

if `HasAnnotation` is supposed to mean that my entity has that annotation in its definition, then it is worth noting that it does not have any attributes at all (on the class or any properties)

Again, hasn't been a problem for me, but seemed worth mentioning. (using RC2-16649)

smitpatel:
Resolves #4382 
This converts all the relational annotations (except sequence & `DatabaseName`) in model snapshot to fluent API.
Let me know if I missed any :smile:

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/Migrations/Design/CSharpSnapshotGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Commands/Migrations/Design/CSharpSnapshotGenerator.cs
@@ -8,6 +8,7 @@ using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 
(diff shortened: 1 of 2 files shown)


ajcvickers:
Do we plan to leave provider-specific annotations as annotations? Or will there be another PR to handle things like ForSqlServerToTable?

rowanmiller:
> I'm fine with leaving provider-specific annotations as annotations for now

:+1: 

> it would be good to convert to fluent API in the future. @smitpatel Maybe create an issue for this?

I think we should wait for some more data before spending time on this one. My guess is that very few people will be using the provider specific "overloads".

ajcvickers:
:shipit:

PR #4656, opened 2016-02-27
Query: Relational: GroupJoin/GroupBy compiler fixes and refactorings:

anpete:
- Include now works with GroupJoin/GroupBy
- Tracking now works with GroupJoin/GroupBy
- Materializer factories are now strongly typed.
- Corrected some project default namespaces.
- Simplified compilation/runtime of Include slightly.

@maumar


EntityFramework.sln
```diff
@@ -73,6 +72,11 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "dotnet-ef", "src\dotnet-ef\
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "dotnet-ef.Tests", "test\dotnet-ef.Tests\dotnet-ef.Tests.xproj", "{93D7AB4E-2E04-4217-BBE5-43EC0D54344F}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{E113FE7C-DAA8-49E8-AC8D-AF2B109F5033}"
```
anpete: Will revert

global.json
```diff
@@ -1,3 +0,0 @@
-{
-    "projects": [ "src", "test" ]
```
anpete: Will revert

maumar:
:shipit:

maumar:
you can also enable ComplexNavigationsQueryTestBase -> Include_with_optional_navigation (i will add more tests once the change is in)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/QuerySourceTracingExpressionVisitor.cs
```diff
@@ -62,7 +63,14 @@ protected override Expression VisitQuerySourceReference(QuerySourceReferenceExpr
 
                     if (groupJoinClause != null)
                     {
-                        Visit(groupJoinClause.JoinClause.InnerSequence);
+                        if (groupJoinClause.JoinClause.Equals(_targetQuerySource))
```
maumar: this is way nicer than what I was trying to do - this should unblock another group of LOJ tests

maumar:
After adding some tests, I found an issue for a multi level include along with LOJ. In some cases the orderby we generate is (int?)c.Id, l.Id. When we lift the query, those orderbys are added to the projection, but because one of them is wrapped in convert, the uniquefying logic doesn't trigger. 

Fix is to simply to remove convert when extracting orderings from a join:

in RelationalQueryModelVisitor -> class OuterJoinOrderingExtractor (line 519 ish)

change:

```
expressions.Add(binaryExpression.Left);
```

to

```
expressions.Add(binaryExpression.Left.RemoveConvert());
```

anpete:
@maumar Made changes, thanks.

Issue #2476, opened 2015-06-26
Commands: Drop Database


bricelam:
In the past, we've talked about having a command to delete the database (i.e. call `EnsureDeleted`). Something like...
```cmd
dnx ef database delete
```

benmillane:
Would it be possible to have options similar to what we've had in the EF SetInitializer method e.g. DropCreateDatabaseAlways or DropCreateDatabaseIfModelChanges. Something like

`dnx . ef migration add NewMigration 
 dnx . ef migration apply -db DropCreateDB`

Not sure on the limitations of the commandline stuff as i'm new to it, but enjoying it so far, just feels like it's lacking in power currently.

bricelam:
`dotnet ef database drop` was added by @laskoviymishka in PR #4657.

I attempted to add `Drop-Database` in PR #4907, but ran into issues marshalling script blocks across domain boundaries.

To complete this issue, we need to...

1. Add a way to get the database name and server from PowerShell (e.g. a `Get-DbContext` command)
2. Add the `Drop-Database` command. The code in [bricelam:fist](https://github.com/aspnet/EntityFramework/compare/dev...bricelam:fist) is a good starting point.

natemcmaster:
Added in #6300

PR #4657, opened 2016-02-27
Add Command: Drop Database

laskoviymishka:
Add new Commands: Drop Database (issue #2476)
- Usage sample - `dnx ef database delete`

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/Design/DbContextOperations.cs
+++ b/src/Microsoft.EntityFrameworkCore.Commands/Design/DbContextOperations.cs
@@ -47,6 +47,7 @@ namespace Microsoft.EntityFrameworkCore.Design
             _runtimeServices = startup.ConfigureServices();
         }
 
+        public virtual void DropDatabase([CanBeNull] string contextName, [NotNull] Func<string, bool> confirmCheck)
         public virtual DbContext CreateContext([CanBeNull] string contextType)
             => CreateContext(FindContextType(contextType).Value);
 
--- a/src/dotnet-ef/DatabaseDeleteCommand.cs
+++ b/src/dotnet-ef/DatabaseDeleteCommand.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.EntityFrameworkCore.Commands
+{
+    public class DatabaseDeleteCommand
+    {
+        public static void Configure([NotNull] CommandLineApplication command)
+        {
+            command.Description = "Drop the database for specific enviroment";
+
+            var startupProject = command.Option(
+                "-s|--startup-project <project>",
+                "The startup project to use. If omitted, the current project is used.");
+            var environment = command.Option(
+                "-e|--environment <environment>",
+                "The environment to use. If omitted, \"Development\" is used.");
+            var context = command.Option(
+                "-c|--context <context>",
+                "The DbContext to use. If omitted, the default DbContext is used");
+
+            var force = command.Option(
+                "-f|--force",
+                "The DbContext to use. If omitted, the default DbContext is used");
+            command.HelpOption();
+            command.VerboseOption();
+
+            command.OnExecute(
+                () => Execute(context.Value(), startupProject.Value(), environment.Value(), force.HasValue()));
+        }
+
+        private static int Execute(string context, string startupProject, string environment, bool isForced)
+        {
+            new OperationExecutor(startupProject, environment)
+                .DropDatabase(
+                    context,
+                    message =>
+                    {
+                        Reporter.Output.WriteLine(message);
+                        var readedKey = Console.ReadKey().KeyChar;
+
+                        return isForced || (readedKey == 'y') || (readedKey == 'Y');
+                    });
+
+            return 0;
+        }
+    }
+}
--- a/src/dotnet-ef/OperationExecutor.cs
+++ b/src/dotnet-ef/OperationExecutor.cs
@@ -28,6 +28,9 @@ namespace Microsoft.EntityFrameworkCore.Commands
         private readonly LazyRef<DatabaseOperations> _databaseOperations;
         private readonly LazyRef<MigrationsOperations> _migrationsOperations;
 
+        private const string DataDirEnvName = "ADONET_DATA_DIR";
+        private const string DefaultConfiguration = "Debug";
+
         public OperationExecutor([CanBeNull] string startupProject, [CanBeNull] string environment)
         {
             var project = Path.Combine(Directory.GetCurrentDirectory(), Project.FileName);
@@ -57,6 +60,13 @@ namespace Microsoft.EntityFrameworkCore.Commands
             }
 
             Reporter.Verbose.WriteLine("Build succeeded.".Bold().Black());
+
+            var runtimeOutputPath = startupProjectContext.GetOutputPaths(DefaultConfiguration)?.RuntimeOutputPath;
+            if (!string.IsNullOrEmpty(runtimeOutputPath))
+            {
+                // TODO set data directory in AppDomain when/if this supports desktop .NET
+                Environment.SetEnvironmentVariable(DataDirEnvName, runtimeOutputPath);
+            }
 
             var projectFile = ProjectReader.GetProject(project);
             var startupAssemblyName = new AssemblyName(startupProjectContext.ProjectFile.Name);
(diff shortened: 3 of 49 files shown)


divega:
@bricelam your thoughts on this?

bricelam:
Perfect. We just need to review the command name and come up with a PowerShell counterpart.

smitpatel:
Should we have confirmation message to avoid deleting the database by mistake?

laskoviymishka:
@smitpatel confirmation would be good :+1:  there is no fun to drop your production database accidentally :smile: 
I could add this feature if EF team confirm.

laskoviymishka:
Or we could allow this command only for dev enviroment. May be even remove options for changing enviroment :smile:

natemcmaster:
+1 for adding a confirmation message, but only if we add a `--force` option to skip the message.

rowanmiller:
Agreed, confirmation and `--force`

laskoviymishka:
Add confirm message. If force (`-f` or `--force`) option presented skip confirm check.
I not found any confirm message like behaivior in Microsoft.Extensions.CommandLineUtils so i implemented extension method inside `CommandLineApplicationExtensions`

src/dotnet-ef/DatabaseDeleteCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            command.Confirm(
+                "Are you sure? (y/N)",
+                () => Execute(context.Value(), startupProject.Value(), environment.Value()),
+                () =>
+                {
+                    Reporter.Output.WriteLine(" Cancel droping database. Use -f|--force to skip.");
+                    return 1;
+                },
+                () => force.HasValue());
```
bricelam: I would rather see this inline in the `Execute` method for now. If we need it more generally later, we can always extract it.

src/dotnet-ef/OperationExecutor.cs
```diff
@@ -104,6 +105,14 @@ public OperationExecutor([CanBeNull] string startupProject, [CanBeNull] string e
                     rootNamespace));
         }
 
+        public virtual void DropDatabase([CanBeNull] string contextName)
+        {
+            using (var context = _contextOperations.Value.CreateContext(contextName))
+            {
+                context.Database.EnsureDeleted();
```
bricelam: We should make sure this logs something like "Dropping database 'Northwind' on server '(localdb)\MSSQLLocalDB'..." and add it to [the list of loggers that get forwarded to the console](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Commands/Design/Internal/CommandLogger.cs#L18-L22).

src/dotnet-ef/DatabaseDeleteCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+                "Force confirm message. If omitted, confirm message not used");
+            command.HelpOption();
+            command.VerboseOption();
+
+            command.Confirm(
+                "Are you sure? (y/N)",
+                () => Execute(context.Value(), startupProject.Value(), environment.Value()),
+                () =>
+                {
+                    Reporter.Output.WriteLine(" Cancel droping database. Use -f|--force to skip.");
```
bricelam: I don't think this information is necessary. It's expected that if they cancel, nothing happens.

src/dotnet-ef/DatabaseDeleteCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+                "-c|--context <context>",
+                "The DbContext to use. If omitted, the default DbContext is used");
+            var force = command.Option(
+                "-f|--force",
+                "Force confirm message. If omitted, confirm message not used");
+            command.HelpOption();
+            command.VerboseOption();
+
+            command.Confirm(
+                "Are you sure? (y/N)",
```
bricelam: Idiomatic PowerShell is something like the following. I'm not sure if there is idiomatic CLI; I couldn't find any examples of confirmation in the DNX or CLI repositories. Regardless, I think we should at least tell them the database and server name here.

    Confirm
    Are you sure you want to perform this action?
    Performing the operation "Drop Database" on target "'Northwind' on server '(localdb)\MSSQLLocalDB'".
    [Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "Y"):

src/dotnet-ef/DatabaseDeleteCommand.cs
laskoviymishka: Sure, i think it would be nice to see such feature in DNX repo (Microsoft.Extensions.CommandLineUtils namespace) i'll create issue for them :smile:

src/dotnet-ef/DatabaseDeleteCommand.cs
laskoviymishka: Well i try to found any confirm like behavior in asp net libs and have no success. According to my experience usually command line prompt looks this way - some bla-bla, are really sure and (y/N) where capital letter is default option. But i could change this message to this one (and implement all this options)

src/dotnet-ef/OperationExecutor.cs
laskoviymishka: :+1: good mention. i'll check it.

bricelam: (If it doesn't, you can just file a bug--no need to fix in this PR)

laskoviymishka: It doesn't i'll fix it tommorow

laskoviymishka: Ok, i'll create bug after merge.

src/dotnet-ef/DatabaseDeleteCommand.cs
bricelam: Oh no, I don't want it to look like PowerShell. :smile: It was just an illustration of what information and options PowerShell engineers thought were important.

I like your convention. Let's make it a thing. Just add a bit more information. Something like:

    Are you sure you want to drop the database 'Northwind' on server '.\SQLEXPRESS'? (y/N)
    _

bricelam: Where database is `DbConnection.Database` and server is `DbConnection.DataSource`.

devochka-valera: This option looks perfect! :heart_eyes:

laskoviymishka: This one looks perfect :heart:

src/Microsoft.EntityFrameworkCore.Commands/Design/DbContextOperations.cs
```diff
@@ -45,6 +45,18 @@ public class DbContextOperations
             _runtimeServices = startup.ConfigureServices();
         }
 
+        public virtual void DropDatabase([CanBeNull] string contextName, [NotNull] Func<string, bool> confirmCheck)
+        {
+            using (var context = CreateContext(contextName))
+            {
+                if (confirmCheck(
+                    $"Are you sure you want to drop the database '{context.Database.GetDbConnection().Database}' on server '{context.Database.GetDbConnection().DataSource}'? (y/N)"))
```
bricelam: I think this message needs to be generated by the caller (it will be different for PowerShell). Make `confirmCheck` take the two pieces of information instead of the message, and format them into a message in `DatabaseDeleteCommand.Execute`.

bricelam:
Do you also want to take a stab at implementing the PowerShell command? (If not, I'll do it as part of merging you PR.)

laskoviymishka:
I could do this, where I could find power shell commands? :)

src/Microsoft.EntityFrameworkCore.Commands/Design/DbContextOperations.cs
laskoviymishka: Great, I'll fix it.

bricelam:
They're in [src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1)

laskoviymishka:
@bricelam sorry for late response, but my knowledge of :fist:  :shell: (power shell) is not good enough to add this commands, could you please assist with these :smile_cat:

bricelam:
lol, no problem; I'd be happy to add them. Is this PR ready to merge then?

laskoviymishka:
@bricelam i think so, everything beside :fist: :shell: is ready

bricelam:
Will be merged with #4907.

@laskoviymishka would you mind rebase-and-squashing on top of `dev`?

laskoviymishka:
@bricelam OK, but i do not know how to do it :smile:

bricelam:
From command line:
```sh
git checkout dev
git pull upstream dev
git checkout dropDatabaseCommand
git rebase -i dev # change all but the first PICKs to SQUASH
git push origin dropDatabaseCommand:dropDatabaseCommand --force
```
Or you can do it with [TrotoiseGit](https://tortoisegit.org/):
image

laskoviymishka:
@bricelam done, but i not sure that everything is OK :sweat_smile:

bricelam:
Merged as bd6d4f7e8a38aa2f44e942cfac806e3c5c89facf

PR #4661, opened 2016-02-28
Fix #2942: Script-Migrations throws when no migrations.

### LINKED ISSUE #2942: Script-Migrations throws when no migrations
Calling `Script-Migration` on a project with no migrations throws the following exception.

    System.InvalidOperationException: Sequence contains no elements
       at System.Linq.Enumerable.Last[TSource](IEnumerable`1 source)
       at Microsoft.Data.Entity.Migrations.Internal.Migrator.GenerateScript(String fromMigration, String toMigration, Boolean idempotent)
       at Microsoft.Data.Entity.Commands.MigrationTool.ScriptMigration(String fromMigrationName, String toMigrationName, Boolean idempotent, String contextTypeName, String startupAssemblyName)
       at Microsoft.Data.Entity.Commands.Executor.ScriptMigrationImpl(String fromMigrationName, String toMigrationName, Boolean idempotent, String contextTypeName)
       at Microsoft.Data.Entity.Commands.Executor.ScriptMigration.<>c__DisplayClass0_1.<.ctor>b__0()
       at Microsoft.Data.Entity.Commands.Executor.OperationBase.<>c__DisplayClass5_0`1.<Execute>b__0()
       at Microsoft.Data.Entity.Commands.Executor.OperationBase.Execute(Action action)

laskoviymishka:
Add check for empty migrations when create migrate script. 
If so show show error to reporter with message `There is no migration`


bricelam:
This doesn't look like the right fix. We shouldn't error if there are no migrations. The behavior should be basically the same as `Update-Database` with no migrations. You should basically just get a script that creates the `__EFMigrationsHistory` table.

laskoviymishka:
Ok, I'll change my PR

laskoviymishka:
Hi, after little research i reimplement my PR.
Current implementation of GenerateScript was refactored according to aproach that used in Emigrate and Migrate Async methods (pretty much same operation so we could share logic between of it). These methods was refactored long time ago, but generate script part remain same.
Migrate and GenerateScript methods have absolutely same logic about how to get migrations which should be executed. Only difference is how we use this migrations after we populate them. In migrate methods we execute it right after generating. In GenerateScript we build Indented StringBuilder string which contain all necessary SQL scripts. So i extract shared logic into separate method and reuse it in both flows.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
```diff
@@ -196,107 +207,76 @@ public virtual void Migrate(string targetMigration = null)
...
-                fromMigration = Migration.InitialDatabase;
-            }
-            else if (fromMigration != Migration.InitialDatabase)
-            {
-                fromMigration = _migrationsAssembly.GetMigrationId(fromMigration);
-            }
+            var builder = new IndentedStringBuilder();
 
-            if (string.IsNullOrEmpty(toMigration))
+            if (migrationsToApply.Any(t => t.GetId() == firstMigrationId) || !(migrationsToApply.Any() || migrationsToRevert.Any()))
```
bricelam: This doesn't feel right. We should only create the history table if `fromMigration` is null, empty, or `Migration.InitialDatabase`.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
```diff
@@ -196,107 +207,76 @@ public virtual void Migrate(string targetMigration = null)
             string toMigration = null,
             bool idempotent = false)
         {
-            var migrations = _migrationsAssembly.Migrations;
+            var firstMigrationId = _migrationsAssembly.Migrations.Select(t => t.Key).FirstOrDefault();
+            var skippedMigrations = _migrationsAssembly.Migrations
+                .Where(t => string.Compare(t.Key, fromMigration, StringComparison.OrdinalIgnoreCase) < 0)
+                .Select(t => t.Key);
```
bricelam: Nit: Can we reuse the concept of applied migrations here instead of introducing the concept of "skipped".

bricelam:
Looking good. I love the refactoring! (was on my TODO list)

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
laskoviymishka: In this context *imho*  i think this more suitable word (since we not apply migration that already was applied). But i could change it :)

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
laskoviymishka: Well we create script when migration include firstMigration (left OR statement) or if we not have any migrations (right OR statement). Well this conditions little bit odd, i should change it :sweat_smile:

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
bricelam: Nope, I was just nit-picking. I'm happy with whatever you think is best.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
```diff
@@ -196,107 +207,76 @@ public virtual void Migrate(string targetMigration = null)
...
-                fromMigration = Migration.InitialDatabase;
-            }
-            else if (fromMigration != Migration.InitialDatabase)
-            {
-                fromMigration = _migrationsAssembly.GetMigrationId(fromMigration);
-            }
+            var builder = new IndentedStringBuilder();
 
-            if (string.IsNullOrEmpty(toMigration))
+            if (toMigration == Migration.InitialDatabase || string.IsNullOrEmpty(fromMigration))
```
bricelam: Shouldn't this be `if (fromMigration ...`?

bricelam: It might be worth adding some tests for this.

laskoviymishka: make sense, i'll add it.
This new dotnet cli have really weak tooling right now, and debugging test is real pain :cry:

bricelam: Amen to that. If you run `build initialize` then open `EntityFramework.sln` it bypasses most of the CLI and Web Tooling workflow and lets you use plain-ol' csproj on Full .NET. You can run the unit tests using [TestDriven.NET](http://testdriven.net/).

laskoviymishka: @bricelam shame on me :scream_cat:, you are right this should be ```if (fromMigration ...```

bricelam:
Merged as 0822ffc4851ca472f35a3068a4ce108f84e71fb9

PR #4667, opened 2016-02-29
SqlServer: Treat String literal as unicode by default

### LINKED ISSUE #4622: Filtering NVARCHAR column is missing "N" when using string literal
Hi,

I've a very simple `Forum` entity:

```
public class Forum
    {
        public int Id { get; set; }

        public virtual List<Post> Posts{ get; set; }
    }
```
And nothing extra related to Forum in the `OnModelCreating` method.
EF correctly generates Name as _NVARCHAR(MAX)_ and filtering with a variable generates the correct SQL:

    var var = "דורון";
    var scoops = _context.Forums.Single(x => x.Name == var);

```
exec sp_executesql N'SELECT TOP(2) [x].[Id], [x].[Description], [x].[Name]
FROM [Forums] AS [x]
WHERE [x].[Name] = @__var_0',N'@__var_0 nvarchar(4000)',@__var_0=N'דורון'
```

But when I use a string literal:

    var scoops = _context.Forums.Single(x => x.Name == "דורון");
EF generates the following SQL:

```
SELECT TOP(2) [x].[Id], [x].[Description], [x].[Name]
FROM [Forums] AS [x]
WHERE [x].[Name] = 'דורון'
```

Note that the parameter is treated as VARCHAR instead of **N**VARCHAR

* Using RC1
* The generated SQL was taken from SQL Server profiler.

smitpatel:
resolves #4622

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
@@ -222,7 +222,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations
             }
 
             builder
-                .Append("IF SCHEMA_ID(N")
+                .Append("IF SCHEMA_ID(")
                 .Append(SqlGenerationHelper.GenerateLiteral(operation.Name))
                 .Append(") IS NULL EXEC(N'CREATE SCHEMA ")
                 .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name))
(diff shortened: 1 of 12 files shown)


src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -491,7 +491,7 @@ protected override void ForeignKeyAction(ReferentialAction referentialAction, Re
                 .AppendLine(" = [d].[name]")
                 .AppendLine("FROM [sys].[default_constraints] [d]")
                 .AppendLine("INNER JOIN [sys].[columns] [c] ON [d].[parent_column_id] = [c].[column_id]")
-                .Append("WHERE ([d].[parent_object_id] = OBJECT_ID(N'");
+                .Append("WHERE ([d].[parent_object_id] = OBJECT_ID('");
```
smitpatel: Nice catch! Will update at all places affected.

natemcmaster:
:shipit: with those changes

PR #4669, opened 2016-03-01
Validate properties mapped to same column has same type mapping.

### LINKED ISSUE #4251: error mapping sql types in derived types
This is an odd one I found while trying to repo a different issue. Notice that SpecialUser1 and 2 both derive from ApplicationUser. Also notice that they both define "Sex" but of different type. Probably this is an error on the dev's part, but in more complex hierarchies maybe that happens. 

The problem is that EF scripts successfully into the database with "sex" column of type int, while SpecialUser2.Sex is type string!  I'd expect an error to be thrown.

This is with RC2 16649. Did you loosen up error checking since RC1 - because that might explain why I can't repo the other issue I had where defining the same property in two sibling derived types threw an error, even if they used the same type. (https://github.com/aspnet/EntityFramework/issues/4224)

using Microsoft.AspNet.Identity.EntityFramework;
using Microsoft.Data.Entity;

namespace WebApplication5
{
    public class ApplicationUser : IdentityUser
    {
    }
    public class SpecialUser1 : ApplicationUser
    {
        //same property in SpecialUser1 and SpecialUser2
        public char Sex { get; set; }
    }

    public class SpecialUser2 : ApplicationUser
    {
        public string SomethingElse { get; set; }
        public string Sex { get; set; } //same property NAME in SpecialUser1 and SpecialUser2
        public string SomethingElse2 { get; set; }
    }

    public class ApplicationDbContext : IdentityDbContext<ApplicationUser>
    {
        public DbSet<ApplicationUser> ApplicationUsers { get; set; }

        //two derived classes with same property NAME as each other, but different types
        public DbSet<SpecialUser1> Users1 { get; set; }
        public DbSet<SpecialUser2> Users2 { get; set; }

        {
        }
    }
}

smitpatel:
Resolves #4251 
If the properties which are being added to single table has same name and different type mapping then throw exception.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
@@ -70,15 +70,47 @@ namespace Microsoft.EntityFrameworkCore.Internal
 
         protected virtual void EnsureDistinctColumnNames([NotNull] IModel model)
         {
+            var groupedEntityTypes = new Dictionary<string, List<IEntityType>>();
             foreach (var entityType in model.GetEntityTypes())
             {
-                var columns = new HashSet<string>();
-                foreach (var property in entityType.GetProperties())
+                var annotations = _relationalExtensions.For(entityType);
+                var tableName = annotations.Schema + "." + annotations.TableName;
+                if (!groupedEntityTypes.ContainsKey(tableName))
                 {
-                    var name = _relationalExtensions.For(property).ColumnName;
-                    if (!columns.Add(name))
+                    groupedEntityTypes[tableName] = new List<IEntityType>();
+                }
+                groupedEntityTypes[tableName].Add(entityType);
+            }
+
+            foreach (var table in groupedEntityTypes.Keys)
+            {
+                var properties = groupedEntityTypes[table].SelectMany(et => et.GetDeclaredProperties());
+                var propertyTypeMappings = new Dictionary<string, IProperty>();
+
+                foreach (var property in properties)
+                {
+                    var columnName = _relationalExtensions.For(property).ColumnName;
+                    IProperty duplicateProperty;
+                    if (propertyTypeMappings.TryGetValue(columnName, out duplicateProperty))
                     {
-                        ShowError(RelationalStrings.DuplicateColumnName(name, entityType.Name, property.Name));
+                        var currentTypeString = _relationalExtensions.For(property).ColumnType ?? _typeMapper.GetMapping(property).DefaultTypeName;
+                        var previousTypeString = _relationalExtensions.For(duplicateProperty).ColumnType ?? _typeMapper.GetMapping(duplicateProperty).DefaultTypeName;
+                        if (!(currentTypeString).Equals(previousTypeString, StringComparison.OrdinalIgnoreCase))
+                        {
+                            ShowError(RelationalStrings.DuplicateColumnName(
+                                duplicateProperty.DeclaringEntityType.DisplayName(),
+                                duplicateProperty.Name,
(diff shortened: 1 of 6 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
```diff
@@ -248,7 +248,7 @@
     <value>Executed DbCommand ({elapsed}ms) [Parameters=[{parameters}], CommandType='{commandType}', CommandTimeout='{commandTimeout}']{newLine}{commandText}</value>
   </data>
   <data name="DuplicateColumnName" xml:space="preserve">
-    <value>Cannot use column name '{columnName}' in entity '{entityType}' for property '{property}' since it is being used for another property.</value>
+    <value>Property '{property1}' in entity '{entityType1}' is mapped to column '{columnName}' in '{table}'. Property '{property2}' in entity '{entityType2}' is mapped to the same column but uses different type.</value>
   </data>
```
smitpatel: @rowanmiller to review the exception message.

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
```diff
@@ -70,15 +70,43 @@ protected virtual void EnsureDistinctTableNames([NotNull] IModel model)
...
+
+                foreach (var property in properties)
                 {
-                    var name = _relationalExtensions.For(property).ColumnName;
-                    if (!columns.Add(name))
+                    var columnName = _relationalExtensions.For(property).ColumnName;
+                    IProperty duplicateProperty;
+                    if (propertyTypeMappings.TryGetValue(columnName, out duplicateProperty))
+                    {
+                        if (_typeMapper.FindMapping(duplicateProperty) != _typeMapper.FindMapping(property))
```
AndriySvyryd: This might be too strict. Perhaps comparing the type string would be more correct:
```
(_relationalExtensions.For(duplicateProperty).ColumnType ?? _typeMapper.GetMapping(duplicateProperty).DefaultTypeName).Equals(
    _relationalExtensions.For(property).ColumnType ?? _typeMapper.GetMapping(property).DefaultTypeName,
    StringComparison.OrdinalIgnoreCase)
```
@ajcvickers Thoughts?

AndriySvyryd:
:shipit: for the rest of the implementation

src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
rowanmiller: I think it can be more concise with

'{entityType1}.{property1}' and '{entityType2}.{property2}' are both mapped to column '{columnName}' in '{table}' but are configured to use different data types.

rowanmiller: BTW do we have the types? If so, we should include that in the message too...

> '{entityType1}.{property1}' and '{entityType2}.{property2}' are both mapped to column '{columnName}' in '{table}' but are configured to use different data types ('{dataType1}' and '{dataType2}').

smitpatel: updated with second message since we have type info available

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
smitpatel: Changed to string based comparison of types.

smitpatel:
Updated.

AndriySvyryd:
:shipit:

PR #4670, opened 2016-03-01
Revive the *.csproj solution

bricelam:
With this, `build initialize` will transform the `project.lock.json` files into MSBuild items.

I tried to use NuGet's project.json MSBuild integration, but in the end it was impossible to use the same `project.json` file for both .NET Core CLI and NuGet. The contention was primarily around project-to-project references.


davidfowl:
Nooooooooo

bricelam:
This was an EF team decision. *.xproj is too painful at the moment. We'll reevaluate again when it stabilizes.

Eilon:
@davidfowl I discussed this with several EF folks, it's critical to unblock progress.

@bricelam I heard there should be new internal builds of tooling this week so hopefully it's not a long wait!

src/Microsoft.EntityFrameworkCore.Commands/project.json
```diff
@@ -12,9 +8,6 @@
     "xmlDoc": true
   },
   "compile": "..\\Shared\\*.cs",
-  "namedResource": {
```
natemcmaster: Why don't we need this anymore? Does dotnet-cli automatically include resx files now?

natemcmaster:
:shipit: x 2

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/project.json
```diff
@@ -8,12 +8,9 @@
     "Microsoft.Extensions.Configuration.Json": "1.0.0-*",
     "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.0.0-*"
   },
-  "content": ["**/*.sql", "config.json"],
-  "publishExclude": [".testPublish/**", "obj/**", "bin/**"],
+  "content": [ "**/*.sql", "config.json" ],
+  "publishExclude": [ ".testPublish/**", "obj/**", "bin/**" ],
```
bricelam: Cool. Will do.

Issue #4019, opened 2015-12-09
Specifying attribute Table="SomeDatabase.dbo.SomeTable" won't work.


zpul:
I need to access a Table from another database located in the same server, so I did specify its location in the table name but this do not work.

The problem are the angled brackets put in the name of the table. Can I avoid inserting this brackets into the table name?

I receive this exception:

   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at System.Data.SqlClient.SqlDataReader.get_MetaData()
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, SqlDataReader ds)
   at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean asyncWrite)
   at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, String method)
   at System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior, String method)
   at System.Data.SqlClient.SqlCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at System.Data.Common.DbCommand.ExecuteReader()
[...]

PR #4672, opened 2016-03-01
Imlementation for Support specifying catalog for table mapping (initial PR)

laskoviymishka:
(fix issue #4019)

- Add database property to relational type annotations
- Follow this property from explicit mapping to default query sql generator (to create proper table alias including database name)
- Enhance relational model validator to including database name (this will allow have same table and schemas for different databases)
- Enhance model configuration - add generator for proper toTable method if database is presented

There are still some open questions:
1) There is no database/catalog propery in TableAttribute. 

So we could not implement data annotation mapping to different database. There is a possible solution by introducing additional attribute for this need (like DatabaseAttribute) with single proprte Name. In this case usage could be fairly simple

```
[Database("Pink")]
[Table("Unicorns", Schema="Fluffy")]
class DifferentDatabaseItem{ }
```

Instead of:
```
 modelBuilder.Entity<MappingQueryTestBase.DifferentDatabaseItem>(e =>
            {
                e.Metadata.Relational().TableName = "Unicorns";
                e.Metadata.Relational().Schema = "Fluffy";
                e.Metadata.Relational().Database = "Pink";
            });
```
But there is a lot more implementation needed to add such attribute. It require add Database property to every migration command and follow it from code gen to query generation.

2) There is issues with navigation properties and foreign keys between databases. We could join tables but could not generate proper foreign key without hacks.


rowanmiller:
Hey,

We're going to close this one out as we would need a complete implementation of catalog support in order to take it at this stage. Some of the key things missing from this so far are migrations support and a top level fluent API (currently needs to be set via lower level metadata APIs).

There are also some open design issue here. Most notably, should this be a relational concept, or specific to SQL Server. Feel free to open up an issue to start hashing these design concerns out - or use the existing #4019.

We are very willing to take this feature, but given that the pull request needs more than just some minor changes before we would accept it, we're going to close the PR for the moment.

~Rowan

PR #4673, opened 2016-03-01
Fix #4253: Cleanup code generated in ModelSnapshot

### LINKED ISSUE #4253: Cleanup code generated in ModelSnapshot (extra line breaks, etc.)
For almost all entities model snapshot scaffolds the table name. If there is nothing else generated after `ToTable` call then blank line is added. If there are entity type annotations generated then there are 2 blank lines before HasAnnotation call.

```
builder.Entity(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+DerivedType"", b =>
    {
        b.HasBaseType(""Microsoft.Data.Entity.Commands.Migrations.ModelSnapshotTest+BaseType"");

        b.ToTable(""DerivedType"");

    });
```

laskoviymishka:
Change append line to single append. Add this line appender inside of next line builder.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/Migrations/Design/CSharpSnapshotGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Commands/Migrations/Design/CSharpSnapshotGenerator.cs
@@ -180,7 +180,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Design
                     .AppendLine()
                     .Append("b.ToTable(")
                     .Append(_code.Literal(entityType.DisplayName()))
-                    .AppendLine(");");
+                    .Append(");");
             }
         }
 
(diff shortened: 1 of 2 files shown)


smitpatel:
This will actually create extra line break if there are not other configurations before property declarations.
Like
````c#
builder.Entity("EntityA", b =>
    {

         b.Property<int>("Id");
    });
````
Apparently, the way model snapshot tests are written this won't occur at present because `ToTable` is always scaffolded but that is not the case always. See #4382 . So it drills down to where we want to add extra line break. :smile: 
Also `ToTable` calls are going away with #4654 . I'll take a look if the extra line break can be removed.

laskoviymishka:
@smitpatel should i change my PR somehow?

laskoviymishka:
@smitpatel should i merge my fix with your PR? and does it have any sense now? :sweat_smile: 
Your PR looks very scary, may be it better to reimplement in with new PR rather than merge.

smitpatel:
Apparently with #4654 going in, the issue this PR fix no longer exists.
The current way still leaves space for additional line breaks.
Configurations in entity type always adds line above but does not add below. This avoids un-necessary line break at the end. There are 2 exceptions to this.
1. First property does not add line above since there is nothing before (in the absence of base type)
2. BaseType does not add line above since it always comes first but it adds line before to compensate for the first property.

Now in this whole setup if there are some configurations not declared then it will leave extra line breaks.
If there are no properties (of course in derived type) but any other configuration then extra line breaks will occur between those 2. If Base type is the only thing then there will be extra line at the end.
Since they are not much frequent and does not create merge issues for model snapshot its very low priority.

laskoviymishka:
great!
Anyway there is still some work for me left, this is cool :+1:
So i'll close this pr, and create new one for this super minor issue :smiley cat:
Should i create new issue that you described?

smitpatel:
@laskoviymishka - I have updated existing issue #4253

PR #4679, opened 2016-03-01
Better handling for shadow FKs on attach, and general cleanup of change detection and fixup

ajcvickers:
ChangeDetector is now only responsible for detecting changes. It doesn't do any updating of state. NavigationFixter does all update of state that happens when a property or navigation changes, or when an entity is initially tracked, and is responsible for keeping the snapshots and dependent maps in sync.

We now no longer set properties (e.g. FKs, navigations) on entities that are not yet tracked. Instead we keep a side dictionary of these entities and do any necessary fixup when the entity is ultimately tracked.

This makes our behavior more consistent as to when we set modified or not. In particular, if an entity with a shadow state property is attached, then we will set its FK based on navigation values and the resulting entity will not be marked as modified. We also do the same thing now for non shadow state properties to be consistent.

The dependent map handles conceptually null FKs, although this could be optimized.


divega:
:shipit:

PR #4680, opened 2016-03-02
Fix #4527 - Support update to specified fields with Attach InMemory

### LINKED ISSUE #4527: Partial updating with InMemoryDatabases
Instead of pulling a full database object into code, updating it and then saving it in the database I like to do partial updates on rows with entity framework like this:
```csharp
var efObject = new EfObject
{
	Id = participantTaskId,
	Date = DateTime.UtcNow
    //Has other properties that are not being set
};
EntityEntry<EfObject> entry = this.dbContext.EfObjects.Attach(efObject);
entry.Property(ps => ps.Date).IsModified = true;
await this.dbContext.SaveChangesAsync();
```
And it always seems to work fine. But recently when doing automated testing with the in memory databases this will update the efObject to only having an id and a date. All other properties for that object will be default values, even if they had values before. 

I am assuming this is a bug and will be fixed. Is there any work around for this besides doing a full update?

mikary:
Throw DbUpdateConcurrencyException on invalid updates to InMemory store

@divega @ajcvickers


src/Microsoft.EntityFrameworkCore.InMemory/Storage/Internal/InMemoryTable.cs
```diff
@@ -27,10 +27,43 @@ public virtual void Create(IUpdateEntry entry)
             => _rows.Add(CreateKey(entry), CreateValueBuffer(entry));
 
         public virtual void Delete(IUpdateEntry entry)
-            => _rows.Remove(CreateKey(entry));
+        {
+            var key = CreateKey(entry);
+
+            if (_rows.ContainsKey(key))
+            {
+                _rows.Remove(CreateKey(entry));
```
ajcvickers: Calling CreateKey twice here.

src/Microsoft.EntityFrameworkCore.InMemory/Storage/Internal/InMemoryTable.cs
```diff
@@ -27,10 +27,43 @@ public virtual void Create(IUpdateEntry entry)
...
+            }
+            else
+            {
+                throw new DbUpdateConcurrencyException("", new[] { entry });
```
ajcvickers: Need an exception message here.

src/Microsoft.EntityFrameworkCore.InMemory/Storage/Internal/InMemoryTable.cs
```diff
@@ -27,10 +27,43 @@ public virtual void Create(IUpdateEntry entry)
...
+                    valueBuffer[index] = entry.IsModified(properties[index])
+                        ? entry.GetCurrentValue(properties[index])
+                        : _rows[key][index];
+                }
+
+                _rows[CreateKey(entry)] = valueBuffer;
+            }
+            else
+            {
+                throw new DbUpdateConcurrencyException("", new[] { entry });
```
ajcvickers: Exception message.

src/Microsoft.EntityFrameworkCore.InMemory/Storage/Internal/InMemoryTable.cs
```diff
@@ -27,10 +27,43 @@ public virtual void Create(IUpdateEntry entry)
...
+        }
 
         public virtual void Update(IUpdateEntry entry)
-            => _rows[CreateKey(entry)] = CreateValueBuffer(entry);
+        {
+            var key = CreateKey(entry);
+
+            if (_rows.ContainsKey(key))
+            {
+                var properties = entry.EntityType.GetProperties().ToArray();
```
ajcvickers: Use ToList here. Doesn't matter much, but slightly more efficient for short term use of the data structure.

test/Microsoft.EntityFrameworkCore.FunctionalTests/UpdatesTestBase.cs
```diff
@@ -0,0 +1,267 @@
...
+                var entry = context.Products.Attach(
+                    new Product
+                    {
+                        Id = productId,
+                        Price = 1.99M
+                    });
+
+                entry.Property(p => p.Price).IsModified = true;
+
+                Assert.Equal(false, entry.Property(p => p.DependentId).IsModified);
```
ajcvickers: Use Assert.False.

test/Microsoft.EntityFrameworkCore.FunctionalTests/UpdatesTestBase.cs
```diff
@@ -0,0 +1,267 @@
...
+                var entry = context.Products.Attach(
+                    new Product
+                    {
+                        Id = new Guid("3d1302c5-4cf8-4043-9758-de9398f6fe10"),
+                        Name = "Apple Fritter"
+                    });
+
+                entry.Property(c => c.Name).IsModified = true;
+
+                Assert.Throws<DbUpdateConcurrencyException>(
```
ajcvickers: Assert correct exception message here. (Would have caught that there is no exception message in the product code.)

test/Microsoft.EntityFrameworkCore.FunctionalTests/UpdatesTestBase.cs
```diff
@@ -0,0 +1,267 @@
...
+        {
+            using (var context = CreateContext())
+            {
+                var entry = context.Products.Remove(
+                    new Product
+                    {
+                        Id = new Guid("3d1302c5-4cf8-4043-9758-de9398f6fe10")
+                    });
+
+                Assert.Throws<DbUpdateConcurrencyException>(
```
ajcvickers: Assert message.

test/Microsoft.EntityFrameworkCore.FunctionalTests/UpdatesTestBase.cs
```diff
@@ -0,0 +1,267 @@
...
+            TestStore = fixture.CreateTestStore();
+        }
+
+        protected TFixture Fixture { get; }
+
+        protected TTestStore TestStore { get; }
+
+        public void Dispose()
+        {
+            TestStore.Dispose();
```
ajcvickers: Consider using expression body syntax where applicable.

ajcvickers:
:shipit: after comments addressed.

PR #4685, opened 2016-03-02
:bug: Process all relationship candidates

### LINKED ISSUE #4682: ModelBuilder fails depending on properties names
`Add-Migration` fails with below classes.

> The navigation 'N' on entity type 'Console.MN' has not been added to the model, or ignored, or target entityType ignored.

If I rename `N` property with `M` (or any name starting with letter < `M`) in `MN` class, migration runs fine.

```
using Microsoft.Data.Entity;
using System;

namespace Console
{
    public class MT
    {
        public Guid Id { get; set; }
        public MN RN { get; set; }
    }

    public class MN
    {
        public Guid Id { get; set; }

        public MN N { get; set; }

        public MN PN { get; set; }

        public IList<MNL> MNLs { get; set; }
    }

    public class MNL
    {
        public Guid PNId { get; set; }

        public Guid CNId { get; set; }

        public MN PN { get; set; }

        public MN CN { get; set; }
    }

    {
        public DbSet<MT> MTs { get; set; }

        {
            optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=MyDatabase;Trusted_Connection=True;");
        }
    }

    class Program
    {
        {

        }
    }  
}
```

smitpatel:
resolves #4682 
Issue: When we encountered multiple relationship candidates, we were falling out of the loop if find ambiguous candidate. Therefore rest of candidates were not being process to build relationships. The issue did not surfaced so far because such missing relationships are found by the relationship discovery convention running on the other side. Which fails for self-ref relationship.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
@@ -276,7 +276,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
                             .RemoveForeignKey(existingForeignKey, ConfigurationSource.Convention);
                     }
 
-                    break;
+                    continue;
                 }
 
                 foreach (var navigation in relationshipCandidate.NavigationProperties)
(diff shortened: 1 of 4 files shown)


AndriySvyryd:
:shipit:

Issue #4540, opened 2016-02-12
Revisit using DbSet property names for table names by default


divega:
Just adding an issue for this since it came up recently in conversations and a few of us seem to like the resulting experience more than what we have. 

Per https://github.com/aspnet/EntityFramework/issues/3060#issuecomment-183069466:

> ... because this is the way I use it in the code, I expect it to be the same in the DB.

PR #4688, opened 2016-03-02
Use DbSet property name as table names

ajcvickers:
Issue #4540

The DbSet property name from the context is used as the name of the table mapped to by that entity type.

If the table name is specified explicitly with ToTable, then this is used instead.

The entity type name is used as the table name for entity types in the model that do not have a corresponding DbSet property. If there is more than one DbSet property for the same entity type, then we fall back to using the entity type name.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/Internal/RelationalModelSource.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/Internal/RelationalModelSource.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal;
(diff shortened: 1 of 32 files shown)


divega:
@AndriySvyryd could you review this one?

src/Microsoft.EntityFrameworkCore/Metadata/IMutableEntityType.cs
```diff
@@ -16,7 +18,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata
     ///         Once the model is built, <see cref="IEntityType" /> represents a ready-only view of the same metadata.
     ///     </para>
     /// </summary>
-    public interface IMutableEntityType : IEntityType, IMutableAnnotatable
+    public interface IMutableEntityType : IEntityType, IMutableAnnotatable, IInfrastructure<InternalEntityTypeBuilder>
```
AndriySvyryd: I was thinking that `EntityType` would implement this interface, not `IMutableEntityType`, since `InternalEntityTypeBuilder` is using `EntityType`.

ajcvickers: Talked to @AndriySvyryd and will revert to casting to EntityType and add a note to the public conventions bug.

AndriySvyryd:
:shipit:

PR #4689, opened 2016-03-02
Allow multiple foreign keys to use the same set of properties

### LINKED ISSUE #3551: Allow multiple foreign keys to use the same set of properties

AndriySvyryd:
Improve shadow key validation message

Fixes #3551


smitpatel:
:shipit:

PR #4696, opened 2016-03-03
Tiny: Only resolve ILogger<> and ISensitiveDataLogger<> from scoped services

ajcvickers:
This clears the way for a logger or logger provider to be set at the context level, either through the constructor or some other sugar mechanism. This may be needed for proposed changes in the way we use D.I. but even if it isn't it will be useful for logging sugar.


src/Microsoft.EntityFrameworkCore.InMemory/Storage/Internal/InMemoryStore.cs
```diff
@@ -97,7 +90,7 @@ public virtual IReadOnlyList<InMemoryTableSnapshot> GetTables(IEntityType entity
             return data;
         }
 
-        public virtual int ExecuteTransaction(IEnumerable<IUpdateEntry> entries)
+        public virtual int ExecuteTransaction(ILogger<InMemoryDatabase> logger, IEnumerable<IUpdateEntry> entries)
```
anpete: Consider keeping entries as the first arg.

anpete:
:shipit:

PR #4713, opened 2016-03-07
CLI Commands: Specify RIDs when building project context

bricelam:
This fixes some assembly/native library resolution errors.

Part of #3925


divega:
:shipit:

PR #4717, opened 2016-03-07
More x-plat testing

natemcmaster:
Extend test-compile target to publish .NET Core test apps for SQL Server testing

cc @divega will better enable testing .NET Core => Sql Azure. cref #2898


lajones:
LGTM :shipit:

Issue #3456, opened 2015-10-16
Add more logging to scaffolding


natemcmaster:
We only log a small amount of information, and almost exclusively just for console output. We can add more thorough logging for provider writers who are building/testing their own reveng implementations.

Also, revisit the logging we produce on the console to ensure it provides useful information for end-users.

natemcmaster:
- [ ] Index on column that could not be scaffolded (see https://github.com/aspnet/EntityFramework/pull/3509/files#r42923752)

lajones:
Fix checked in with commit 2d14e47. Addresses both the issues above.

PR #4719, opened 2016-03-07
[RevEng] Updating some existing logging messages plus adding Trace messages

lajones:
For the SQL Server and Sqlite RevEng providers. Fix for #3456.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Design/RelationalScaffoldingModelFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Design/RelationalScaffoldingModelFactory.cs
@@ -32,6 +32,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
         private readonly TableModel _nullTable = new TableModel();
         private CSharpUniqueNamer<TableModel> _tableNamer;
         private readonly IDatabaseModelFactory _databaseModelFactory;
+        private readonly HashSet<ColumnModel> _unmappedColumns = new HashSet<ColumnModel>();
 
         public RelationalScaffoldingModelFactory(
             [NotNull] ILoggerFactory loggerFactory,
@@ -269,6 +270,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
             var clrType = typeMapping?.ClrType;
             if (clrType == null)
             {
+                _unmappedColumns.Add(column);
                 Logger.LogWarning(RelationalDesignStrings.CannotFindTypeMappingForColumn(column.DisplayName, column.DataType));
                 return null;
             }
@@ -340,17 +342,18 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
                 return null;
             }
 
-            var keyProps = keyColumns.Select(GetPropertyName)
-                .Where(name => builder.Metadata.FindProperty(name) != null)
-                .ToArray();
-
-            if (keyProps.Length != keyColumns.Count)
+            var unmappedColumns = keyColumns
+                .Where(c => _unmappedColumns.Contains(c))
+                .Select(c => c.Name).ToArray();
+            if (unmappedColumns.Any())
             {
-                Logger.LogWarning(RelationalDesignStrings.PrimaryKeyErrorPropertyNotFound(table.DisplayName));
+                Logger.LogWarning(RelationalDesignStrings.PrimaryKeyErrorPropertyNotFound(
+                    table.DisplayName,
+                    string.Join(CultureInfo.CurrentCulture.TextInfo.ListSeparator, unmappedColumns)));
                 return null;
             }
 
-            return builder.HasKey(keyProps);
+            return builder.HasKey(keyColumns.Select(GetPropertyName).ToArray());
         }
 
         protected virtual EntityTypeBuilder VisitIndexes([NotNull] EntityTypeBuilder builder, [NotNull] ICollection<IndexModel> indexes)
@@ -371,21 +374,22 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
             Check.NotNull(builder, nameof(builder));
             Check.NotNull(index, nameof(index));
 
-            var propertyNames = index.IndexColumns
+            var indexColumns = index.IndexColumns
                 .OrderBy(ic => ic.Ordinal)
-                .Select(ic => GetPropertyName(ic.Column))
-                .ToArray();
-
-            if (propertyNames.Count(p => builder.Metadata.FindProperty(p) != null) != propertyNames.Length)
+                .Select(ic => ic.Column).ToArray();
+            var unmappedColumns = indexColumns
+                .Where(c => _unmappedColumns.Contains(c))
+                .Select(c => c.Name).ToArray();
+            if (unmappedColumns.Any())
             {
-                Logger.LogWarning(RelationalDesignStrings.UnableToScaffoldIndexMissingProperty(index.Name));
+                Logger.LogWarning(RelationalDesignStrings.UnableToScaffoldIndexMissingProperty(
+                    index.Name,
+                    string.Join(CultureInfo.CurrentCulture.TextInfo.ListSeparator, unmappedColumns)));
                 return null;
             }
 
-            var columnNames = index.IndexColumns
-                .OrderBy(ic => ic.Ordinal)
-                .Select(ic => ic.Column.Name);
-
+            var columnNames = indexColumns.Select(c => c.Name);
+            var propertyNames = indexColumns.Select(GetPropertyName).ToArray();
             if (index.Table != null)
             {
                 var primaryKeyColumns = index.Table.Columns
@@ -402,9 +406,9 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
                             .GetDefaultKeyName(
                                 index.Table.Name,
                                 true, /* is primary key */
-                                primaryKeyColumns.Select(c => GetPropertyName(c))))
+                                primaryKeyColumns.Select(GetPropertyName)))
                     {
-                        builder.HasKey(propertyNames.ToArray()).HasName(index.Name);
+                        builder.HasKey(propertyNames).HasName(index.Name);
                     }
                     return null;
                 }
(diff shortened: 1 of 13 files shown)


lajones:
@Providers. This doesn't break you, but you might want to consider providing similar logging messages in your own providers.

lajones:
cc @natemcmaster - you'll probably want to check the changes to the sqlite provider.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
```diff
@@ -69,127 +81,172 @@ public virtual DatabaseModel Create(string connectionString, TableSelectionSet t
...
 
-        private void GetColumns()
-        {
-            foreach (var table in _databaseModel.Tables)
+            using (var command = _connection.CreateCommand())
             {
-                var command = _connection.CreateCommand();
-                command.CommandText = $"PRAGMA table_info(\"{table.Name.Replace("\"", "\"\"")}\");";
+                command.CommandText =
+                    "SELECT name, tbl_name FROM sqlite_master WHERE type = 'index'";
```
natemcmaster: Good. This starts a simplification I began writing but never got around to committing. We can simplify this even further by just relying on "PRAGMA index_list" you added below to provide a list of all indexes on a table.

natemcmaster: See https://github.com/natemcmaster/EntityFramework/blob/50436db8a8f4a280a2548eb2a6216bf55bf17d35/src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs for reference

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
```diff
@@ -69,127 +81,172 @@ public virtual DatabaseModel Create(string connectionString, TableSelectionSet t
...
 
                 using (var reader = command.ExecuteReader())
                 {
-                    var ordinal = 0;
                     while (reader.Read())
                     {
-                        var isPk = reader.GetValueOrDefault<bool>("pk");
```
natemcmaster: IMO `isPk` makes this section more understandable.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
```diff
@@ -69,127 +81,172 @@ public virtual DatabaseModel Create(string connectionString, TableSelectionSet t
...
+                        while (reader.Read())
                         {
-                            index.IsUnique = isUnique;
+                            var columnName = reader.GetValueOrDefault<string>("name");
+                            var dataType = reader.GetValueOrDefault<string>("type");
+                            var primaryKeyOrdinal = reader.GetValueOrDefault<int>("pk");
+                            var notNull = reader.GetValueOrDefault<bool>("notnull");
+                            var defaultValue = reader.GetValueOrDefault<string>("dflt_value");
+
+                            Logger.LogTrace(SqliteDesignStrings.FoundColumn(
```
natemcmaster: Nit: coding preference. Instead of creating all of these local variables, logging them, and then setting as the property on another local variable, just move this log below ColumnModel instantiation and log `SqliteDesignStrings.FoundColumn(table.Name, column.Name, column.DataType....` etc

src/Microsoft.EntityFrameworkCore.Sqlite.Design/Properties/SqliteDesignStrings.resx
```diff
@@ -117,6 +117,36 @@
...
+  <data name="FoundIndex" xml:space="preserve">
+    <value>Found index with name: {indexName} on table: {tableName}.</value>
+  </data>
+  <data name="FoundIndexColumn" xml:space="preserve">
+    <value>Found index column on index {indexName} on table {tableName}, column name: {columnName}, ordinal: {ordinal}.</value>
+  </data>
+  <data name="FoundTable" xml:space="preserve">
+    <value>Found table with name: {name}.</value>
+  </data>
+  <data name="FoundUniqueIndex" xml:space="preserve">
```
natemcmaster: Can this be combined with FoundIndex? something like `Found index: Name={name}, Table={tableName}, IsUnique={IsUnique}`

src/Microsoft.EntityFrameworkCore.Sqlite.Design/Properties/SqliteDesignStrings.resx
```diff
@@ -117,6 +117,36 @@
...
+    <value>Determined that index: {indexName} on table {tableName} is unique.</value>
+  </data>
+  <data name="IndexTableNotInSelectionSet" xml:space="preserve">
```
natemcmaster: If we simplify the fetching (as I suggest above), we won't need this error messages as indexes will only be fetched after tables have been filtered.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
lajones: I can add the local variable `isPk` back in as shorthand for `primaryKeyOrdinal != 0` and use that. But I'd rather that what goes out in the trace message is just exactly what we got back from the DB, no logic on top of that, to be consistent with all the other trace messages.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
lajones: The diff engine on this didn't come out very well. I didn't add in "PRAGMA index_list". It was already in `GetIndexes()`. But let's discuss.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/Properties/SqliteDesignStrings.resx
lajones: If we do your suggestion of moving finding the indexes into `GetIndexes()` - then yes there will only be the one message.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/Properties/SqliteDesignStrings.resx
lajones: Noted.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
lajones: Yep - that's probably OK. Will do.

natemcmaster:
I'm concerned we are over-logging. Unless I'm mistaken, I believe all of these **trace** level log messages will appear when a user executes commands with `-Verbose`. @bricelam can you verify what commands logging does with trace level messages?

src/Microsoft.EntityFrameworkCore.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -585,6 +587,30 @@ protected virtual void VisitNavigationProperties([NotNull] ModelBuilder modelBui
             }
         }
 
+        private Tuple<List<IMutableProperty>, IEnumerable<string>> MappedPropertiesAndUnmappedColumns(
```
natemcmaster: This function feels quirky. I have an idea for how we can make this cleaner. Let's discuss

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
lajones: Actually, no. `primaryKeyOrdinal` is what we read, the setting of `IsNullable` and `PrimaryKeyOrdinal` on the `ColumnModel` requires logic So would rather just output what we receive. But your approach does work for `IndexModel` - so I've updated that to work as you suggest.

bricelam:
Trace will only be written if `--verbose` is specified. Here is the [console](https://github.com/aspnet/EntityFramework/blob/dev/src/dotnet-ef/ConsoleCommandLogger.cs) and [PowerShell](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1#L540-L546) loggers.

lajones:
@natemcmaster I've force-pushed with the updated changes.

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
```diff
@@ -67,42 +78,37 @@ public virtual DatabaseModel Create(string connectionString, TableSelectionSet t
...
+            using (var command = _connection.CreateCommand())
             {
-                while (reader.Read())
-                {
-                    var type = reader.GetValueOrDefault<string>("type");
-                    var name = reader.GetValueOrDefault<string>("name");
-                    var tableName = reader.GetValueOrDefault<string>("tbl_name");
+                command.CommandText =
+                    "SELECT name FROM sqlite_master" +
+                    " WHERE type = 'table' AND name <> 'sqlite_sequence'";
```
natemcmaster: Syntax error? I believe this should be `!=` in sqlite.

Also, do we not filter the migrations history table?

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
```diff
@@ -146,50 +161,61 @@ private void GetIndexes()
...
+                            {
+                                Name = reader.GetValueOrDefault<string>("name"),
+                                Table = table,
+                                IsUnique = reader.GetValueOrDefault<bool>("unique")
+                            };
+
+                            Logger.LogTrace(SqliteDesignStrings
+                                .FoundIndex(index.Name, table.Name, index.IsUnique));
+
+                            table.Indexes.Add(index);
```
natemcmaster: :+1: much cleaner

natemcmaster:
Nice! :shipit:

src/Microsoft.EntityFrameworkCore.Sqlite.Design/SqliteDatabaseModelFactory.cs
bricelam: +1 for `<>`. It's more "ANSI"

lajones:
Checked in with commit 2d14e47.

Issue #4627, opened 2016-02-23
Exception msg for properties not included in the model


JanLenoch:
### Title
PropertyMappingValidationConvention throws upon scaffolding a new migration in a full .NET solution.

### Functional impact
Migration cannot be created (scaffolded).

### Minimal repro steps
1. Open the [ZIP archive](https://onedrive.live.com/redir?resid=A58EF47BA9A2178B!114720&authkey=!AN-ENf-QKOU4Ly8&ithint=file%2czip) and restore the solution.
2. Within the SqlServerAccess project, restore NuGet packages.
3. In the Package Manager Console, add a new migration (e.g. `Add-Migration Migration01`).

### Expected result
New migration should be scaffolded.

### Actual result
The following error message appears in the Package Manager Console window:
System.ArgumentNullException: Value cannot be null.
Parameter name: propertyInfo
   at Microsoft.Data.Entity.Utilities.Check.NotNull[T](T value, String parameterName)
   at Microsoft.Data.Entity.Metadata.Conventions.Internal.RelationalPropertyMappingValidationConvention.FindCandidateNavigationPropertyType(PropertyInfo propertyInfo)
   at Microsoft.Data.Entity.Metadata.Conventions.Internal.PropertyMappingValidationConvention.Apply(InternalModelBuilder modelBuilder)
   at Microsoft.Data.Entity.Metadata.Conventions.Internal.ConventionDispatcher.OnModelBuilt(InternalModelBuilder modelBuilder)
   at Microsoft.Data.Entity.Metadata.Internal.InternalModelBuilder.Validate()
   at Microsoft.Data.Entity.ModelBuilder.Validate()
   at Microsoft.Data.Entity.Infrastructure.ModelSource.CreateModel(DbContext context, IConventionSetBuilder conventionSetBuilder, IModelValidator validator)
   at Microsoft.Data.Entity.Infrastructure.ModelSource.<>c__DisplayClass8_0.<GetModel>b__0(Type k)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.Data.Entity.Infrastructure.ModelSource.GetModel(DbContext context, IConventionSetBuilder conventionSetBuilder, IModelValidator validator)
   at Microsoft.Data.Entity.Internal.DbContextServices.CreateModel()
   at Microsoft.Data.Entity.Internal.LazyRef`1.get_Value()
[...]

PR #4720, opened 2016-03-08
Throw a better exception when a property was not discovered, but could be a primitive property

### LINKED ISSUE #4484: Property mapping validation provides unhelpful message for primitive type that cannot be mapped
If the entity has a property of a type that cannot be mapped by the provider and is not a primitive type, then the exception message is:
`The property '{property}' on entity type '{entityType}' has CLR type which is not supported by current provider and it has not been configured to use any supported column type.`
However, if the property is a primitive type, such as uint for SQL Server, then the message is:
`The property '{property}' on entity type '{entityType}' has not been added to the model or ignored.`

It seems to me that we should not be treating primitive types differently in this case.

AndriySvyryd:
Throw a better exception when a property of interface type was not discovered

Resolves #4484, #4627

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
@@ -466,13 +466,16 @@
     <value>The entity type '{entityType}' cannot be removed because '{derivedEntityType}' is derived from it. All derived entity types must be removed or redefined before the entity type can be removed.</value>
   </data>
   <data name="NavigationNotAdded" xml:space="preserve">
-    <value>The navigation '{navigation}' on entity type '{entityType}' has not been added to the model, or ignored, or target entityType ignored.</value>
+    <value>Unable to determine the relationship represented by navigation property '{entityType}.{navigation} of type '{propertyType}'. Either manually configure the relationship, or ignore this property from the model.</value>
   </data>
   <data name="PropertyNotAdded" xml:space="preserve">
-    <value>The property '{property}' on entity type '{entityType}' has not been added to the model or ignored.</value>
+    <value>The property '{entityType}.{property}' could not be mapped, because it is of type '{propertyType}' which is not a supported primitive type or a valid entity type. Either explicitly map this property, or ignore it.</value>
   </data>
   <data name="PropertyNotMapped" xml:space="preserve">
-    <value>The property '{property}' on entity type '{entityType}' has CLR type which is not supported by current provider and it has not been configured to use any supported column type.</value>
+    <value>The property '{entityType}.{property}' is of type '{propertyType}' which is not supported by current database provider. Either change the property CLR type or manually configure the database type for it.</value>
+  </data>
+  <data name="InterfacePropertyNotAdded" xml:space="preserve">
+    <value>The property '{entityType}.{navigation}' is of an interface type ('{propertyType}'). If it is a navigation property manually configure the relationship for this property by casting it to a mapped entity type, otherwise ignore the property from the model.</value>
   </data>
   <data name="NavigationForWrongForeignKey" xml:space="preserve">
     <value>The navigation property '{navigation}' on entity type '{entityType}' cannot be associated with foreign key {targetFk} because it was created for foreign key {actualFk}.</value>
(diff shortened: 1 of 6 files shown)


AndriySvyryd:
@rowanmiller for exception message review

ajcvickers:
:shipit: I like the new messages.

rowanmiller:
Can you summarize when these get thrown? I think we can improve the "has not been added to the model" wording as we add all properties to the model by default. Perhaps it's more a case of "could not be added to the model". I can work on the wording if you just let me know the situations in which each message is thrown :smile:

AndriySvyryd:
@rowanmiller Currently we only discover and add primitive properties that can be mapped. If such property is present on the CLR type or if it was added explicitly, but without specifying the provider type `PropertyNotMapped` is thrown.
`NavigationNotAdded` is thrown for properties that could be navigations, but weren't discovered, probably because there's an ambiguity matching the inverse, convention logging would show more info.
`InterfacePropertyNotAdded` is for interface type properties, we never discover them.
`PropertyNotAdded` is for everything else that still could be a property, so mostly value types that aren't considered primitive in Core or `object`.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -466,13 +466,16 @@
     <value>The entity type '{entityType}' cannot be removed because '{derivedEntityType}' is derived from it. All derived entity types must be removed or redefined before the entity type can be removed.</value>
   </data>
   <data name="NavigationNotAdded" xml:space="preserve">
-    <value>The navigation '{navigation}' on entity type '{entityType}' has not been added to the model, or ignored, or target entityType ignored.</value>
+    <value>The navigation property '{navigation}' of type '{propertyType}' on entity type '{entityType}' has not been added to the model. Either manually configure a relationship using this property, ignore the property or ignore the target entity type.</value>
   </data>
```
rowanmiller: I think the implementation details of what it means to be "added to the model" are leaking out here. Conceptually (for the app developer) all properties (with a public getter/setter are included in the model) but EF can't work out what to do with this one.

Here is my suggestion:

> Unable to determine the relationship represented by navigation property '{entityType}.{navigation} of type '{propertyType}'. Either manually configure the relationship, or ignore this property from the model.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -466,13 +466,16 @@
...
   <data name="PropertyNotAdded" xml:space="preserve">
-    <value>The property '{property}' on entity type '{entityType}' has not been added to the model or ignored.</value>
+    <value>The property '{property}' on entity type '{entityType}' has not been added to the model, because it is of type '{propertyType}' which is not recognized as a supported primitive type or a valid entity type. Either explicitly map this property or ignore it.</value>
   </data>
```
rowanmiller: > The property '{entityType}.{property}' could not be mapped, because it is of type '{propertyType}' which is not a supported primitive type or a valid entity type. Either explicitly map this property, or ignore it.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -466,13 +466,16 @@
...
   <data name="PropertyNotMapped" xml:space="preserve">
-    <value>The property '{property}' on entity type '{entityType}' has CLR type which is not supported by current provider and it has not been configured to use any supported column type.</value>
+    <value>The property '{property}' on entity type '{entityType}' is of type '{propertyType}' which is not supported by current provider. Either change the property CLR type or manually configure the provider type for it.</value>
+  </data>
```
rowanmiller: > The property '{entityType}.{property}' is of type '{propertyType}' which is not supported by current database provider. Either change the property CLR type or manually configure the database type for it.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -466,13 +466,16 @@
...
+  <data name="InterfacePropertyNotAdded" xml:space="preserve">
+    <value>The property '{navigation}' of type '{propertyType}' on entity type '{entityType}' has not been added to the model. If it is a navigation property manually configure a relationship using this property by casting it to a mapped entity type otherwise ignore the property.</value>
   </data>
```
rowanmiller: > The property '{entityType}.{navigation}' is of an interface type ('{propertyType}'). If it is a navigation property manually configure the relationship for this property by casting it to a mapped entity type, otherwise ignore the property from the model.

rowanmiller:
Suggested some updated exception messages that I think match better with what is conceptually happening for the app developer rather than what is happening in our stack. Let me know if you have pushback.

divega:
:+1: I like the suggested messages and the conceptual approach.

smitpatel:
:+1:

Issue #4668, opened 2016-02-29
Reduce magic in the interaction of EF Core and DI


divega:
### TL;DR
This is a proposal to remove the magic that allows derived `DbContext` types with default or parameter-less constructors to be used with DI. The general idea is to embrace the constructor injection pattern when using DI so that it becomes very clear what services are being injected and also so that the non-DI usages can be clearly differentiated. 

### Details
Currently we allow derived `DbContexts` that are supposed to be resolved from DI (e.g. the default application context in an ASP.NET Core application) to have parameter-less constructors. We actually went out of our way to enable this through some complex logic in `DbContextActivator` and the `AddDbContext()` method.

While this allows removing a lot of boilerplate code from derived `DbContext` constructors, it also has some undesired consequences:

1. __It is confusing:__ In the previews we have seen many customers mixing up the various patterns for setting the context options with DI and without DI (i.e. using the `OnConfiguring()` method) and then creating the context instance without DI (i.e. using the `new` keyword) or with DI, which in the worst cases will fail, and in some of the better cases may cause performance issues.

2. The magic doesn't reflect or play well with the patterns customers actually need to learn in order to interact with the DI system correctly in the rest of their application.

3. The resulting constructors are completely opaque about which application services the `DbContext` depends on (and everything downstream needs to use service locator pattern), let alone the fact that it is sharing the application's `IServiceProvider` to resolve its own internal services.

We have started brainstorming and we are going to continue discussing some ideas on how to improve this. Here are some initial thoughts around the alternative of removing the magic. The main disadvantage of this approach is that derived DbContext will need to define a constructor to be used with DI (until now we were getting away with default constructors). It seems that the level of control and clarity about what is going on you gain might be worth the pain though.

The changes would involve all of the following:

1. Get rid of the magic: Remove `DbContextActivator` and the delegate registration that uses it in `AddDbContext()`. 

2. Expand the `DbContext` constructors to include arguments for all the shared application services EF would use within a `DbContext`, including logging, diagnostics and memory cache.

3. Keep the `DbContextOptions` argument to be able to resolve context options that are configured externally.

4. Keep the `IServiceProvider` argument for cases in which a specific service provider (could be the application’s one or a different one) is configured externally. 

[...]

PR #4725, opened 2016-03-08
Reduce magic in the interaction of EF Core and DI

ajcvickers:
See issue #4668 for full description of changes.

This change removes EntityFrameworkServicesBuilder since it is no longer needed for AddDbContext and moves all EF methods that add services to IServiceCollection. Providers will need to react to this by moving AddMyProvider methods to be extension methods of IServiceCollection.


ajcvickers:
Merged

PR #4728, opened 2016-03-08
[small] Use GetValues for SqlServer reverse engineering on Mono

natemcmaster:
Fixes failing reveng tests on Mono.

cc @lajones


lajones:
:shipit:

Issue #4730, opened 2016-03-08
Review usages of NotSupportedException


anpete:
From MSDN: "NotSupportedException indicates that no implementation exists for an invoked method or property."

However we are using the exception in a more general purpose way. E.g. From ConcurrencyDetector.EnterCriticalSection to indicate a concurrent access. We should consider using InvalidOperationException in some of these cases.

divega:
:+1: I think we should change it.

PR #4731, opened 2016-03-08
Fixes/improvements to ConcurrencyDetector:

anpete:
1) Fixed race-condition in EnterCriticalSection.
2) Switched to Disposable pattern to streamline usage.
3) Use InvalidOperationException instead of NotSupportedException. see #4730

cc @mikary


mikary:
Looks much better to me
:shipit:

Issue #4710, opened 2016-03-07
Commands: Don't terminate when Startup isn't constructible


bricelam:
Currently, if we can't construct `Startup`, the commands that use a `DbContext` will fail. Instead, we should warn and continue looking for the `DbContext` using the other mechanisms.

bricelam:
Closing this. We'll take it into consideration as part of #4546.

bricelam:
Re-opening. We should do this.

Issue #4546, opened 2016-02-12
Sort out design time interaction with Startup in ASP.NET Core


rowanmiller:
Our tooling currently runs code in Startup.cs to get the context from services. This is messy though because our design time environment is not the same as the environment the app really runs in. We need to revisit this.

As part of this we need to try and remove ASP.NET specific dependencies from our general commands package so you don't get references to `Microsoft.AspNetCore.Hosting.*` etc in a UWP/WinForms/WPF project (see https://github.com/aspnet/EntityFramework/issues/4830).

Related issues:
 * https://github.com/aspnet/EntityFramework/issues/4494
 * https://github.com/aspnet/EntityFramework/issues/4367

natemcmaster:
See also https://github.com/aspnet/EntityFramework/pull/5037. In this change, we specifically workaround limitations in UWP and startupinvoker. We will need a different abstraction when/if we stop cross-compiling to UWP. #5045

natemcmaster:
We will likely need to react to this: https://github.com/aspnet/Hosting/pull/714
cc @divega

natemcmaster:
Related issue: https://github.com/aspnet/EntityFramework/issues/5304#issuecomment-217956745

Recent Hosting templates have changed to loading configuration in `Program.Main`. EF can't (and shouldn't) call `Program.Main` at design-time, yet we still need to load configuration to get connection strings.

bricelam:
We discussed this today. We came up with a solution that would keep some of the ASP.NET-specific code outside of the main tooling package, but decided that the cost and complexity of implementing it wasn't worth the marginal value it would add.

As part of #5334, we are going to author and structure the packages in a way that reduces the amount of dependencies that get pulled into your project and utilize NuGet & CLI features to prevent the ones we do pull in from getting deployed with your application.

I'm closing this issue and splitting off the remaining work into #5616, #5615 & #5617

PR #4733, opened 2016-03-08
Commands: Inject IHostingEnvironment, IApplicationEnvironment, etc. into Startup

### LINKED ISSUE #4677: CLI Commands: `dbcontext scaffold` has two -s options
`--schema` and `--startup-project`

bricelam:
Other small changes
* Move `OperationHandlers.cs` for easier debugging
* Remove short form of `--schema` (fixes #4677)
* Remove `#if DNX451`
* Canonicalize `project.json`

@NTaylorMullen This *is* the PR you're looking for... :wave:

Part of #3925


src/dotnet-ef/DbContextScaffoldCommand.cs
```diff
@@ -37,7 +37,7 @@ public static void Configure([NotNull] CommandLineApplication command)
                 "-o|--output-dir <path>",
                 "Directory of the project where the classes should be output. If omitted, the top-level project directory is used.");
             var schemas = command.Option(
-                "-s|--schema <schema>",
+                "--schema <schema>",
```
lajones: @bricelam This is OK for now if you need to get this in. But I'd prefer to come up with a new short-form for `--schema`. It's odd that `--table` still has one but `--schema` does not.

NTaylorMullen:
What's the override route for the case when you have custom services (from `Program.Main`) being injected into your `Startup.cs`? It still looks like `Startup.cs` is constructed every time no matter what without fallback to `IDbContextFactory`.

src/dotnet-ef/DbContextScaffoldCommand.cs
bricelam: This was the fix we decided on in triage. You can create a new issue if you think it's common enough to deserve a short name. Personally, I think we should remove even more short names.

bricelam:
@NTaylorMullen We'll that out as part of #4546. Skipping `Startup` will be done as part of #4710. But I think this should unblock 80% of scenarios (including yours).

src/dotnet-ef/DbContextScaffoldCommand.cs
divega: If table is going to be more commonly used than schema then I think having a short name for table makes sense and I don't think we need symmetry between table and schema. Otherwise I think we should reassess.

rowanmiller: > Personally, I think we should remove even more short names.

Agreed, I think we've gone a little overboard with them. To me they are most useful when you type by hand a lot (`-verbose` etc.). I imagine things like table/schema/etc. will generally be stored in a script etc.

lajones: I think users will find the lack of symmetry odd. But I'm OK if we wait and see. It will be easy to add some abbreviation back in if we find it's wanted. Since we have no exclusion syntax, I do expect that the average number of `-t|--table`s on the command line will exceed the number of `--schema`s, sometimes significantly.

rowanmiller:
:shipit:

PR #4734, opened 2016-03-09
Fix #4605 - Do not cache commands with Contains on Enumerable parameter.

### LINKED ISSUE #4605: Query: Revisit Enumerable.Contains command caching
Different SQL is generated for Enumerable.Contains depending on the contents of the enumerable. The second level (command) caching of these commands does not currently have any kind of eviction policy and may grow out of control.

Possible solutions to the issue include:
* Introduce an eviction policy for the command cache
* Only add commands to the cache if the same cache key is encountered more than once
* Do not enter Enumerable.Contains commands into the cache
* Parameterize the Enumerable (TVP) in the query
* Support partial command construction / SQL generation

mikary:
May need to revisit the `CommandCacheKey` comparison after  #4575 is merged
@anpete

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
@@ -45,25 +45,11 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                             return false;
                         }
 
-                        if (value is IEnumerable)
+                        if (value is IEnumerable
+                            && value.GetType() == typeof(object[]))
                         {
-                            var type = value.GetType();
-
-                            if (type == typeof(object[]))
-                            {
-                                // Only compare lengths for FromSql parameters
-                                return ((object[])value).Length == (otherValue as object[])?.Length;
-                            }
-
-                            if (type != typeof(string)
-                                && type != typeof(byte[]))
-                            {
-                                // Comparision for contains parameters
-                                // TODO: This doesn't always need to be deep.
-                                // We could add a LINQ operator parameter attribute to tell us.
-                                return StructuralComparisons
-                                    .StructuralEqualityComparer.Equals(value, otherValue);
-                            }
+                            // FromSql parameters must have the same number of elements
+                            return ((object[])value).Length == (otherValue as object[])?.Length;
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


anpete:
Can you add tests?

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/IQuerySqlGenerator.cs
```diff
@@ -12,6 +12,8 @@ public interface IQuerySqlGenerator
     {
         IRelationalCommand GenerateSql([NotNull] IReadOnlyDictionary<string, object> parameterValues);
 
+        bool CanCacheCommand { get; }
```
anpete: IsCacheable?

mikary:
Updated based on feedback / discussion
Added a test to cover the Contains on Enumerable parameter in a subquery case, though this may not be necessary now that the IsCacheable state is computed after SQL gen.

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -28,6 +28,7 @@ public class DefaultQuerySqlGenerator : ThrowingExpressionVisitor, ISqlExpressio
         private IRelationalCommandBuilder _relationalCommandBuilder;
         private IReadOnlyDictionary<string, object> _parametersValues;
         private ParameterNameGenerator _parameterNameGenerator;
+        private bool _hasParameterLiterals;
```
anpete: Can we remove this and use IsCacheable directly?

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
```diff
@@ -82,9 +81,22 @@ public override bool Equals(object obj)
...
+            IRelationalCommand relationalCommand;
+            var key = new CommandCacheKey(parameters);
+
+            if (!_commandCache.TryGetValue(key, out relationalCommand))
+            {
+                var generator = QuerySqlGeneratorFactory();
+
+                relationalCommand = generator.GenerateSql(parameters);
+
+                if(generator.IsCacheable)
```
anpete: Formatting. Consider R# code cleanup.

anpete:
:shipit:

Issue #3889, opened 2015-11-25
Query: [Nav Prop Translation] SelectMany generates wrong SQL


srollinet:
When I execute this query with the RC1 version
```c#
var test1 = context.Sports
	.Where(s => s.Id == 1)
	.SelectMany(s => s.Teams)
	.SelectMany(s => s.Players)
	.Select(p => p.Name)
	.ToList();
```
The generated SQL looks like
```sql
SELECT [s].[Id], [s].[Name], [s].[TeamId], [s].[Id], [s0].[Id], [p].[Name]
FROM [Player] AS [s]
CROSS JOIN [Team] AS [s0]
CROSS JOIN [Player] AS [p]
WHERE [s].[Id] = @__s1_Id_0
```

1) the tables and columns are wrong (s should be Sport)
2) the join conditions are missing -> cartesian product
3) less important, but too many columns are selected. Only [p].[Name] is really usefull

and with this query
```c#
var test2 = context.Teams
	.Where(t => t.SportId == 1)
	.SelectMany(t => t.Players)
	.Select(p => p.Name)
	.ToList();
```
The generated SQL looks like
```sql
SELECT [t].[Id], [t].[Name], [t].[TeamId], [t].[Id], [p].[Name]
FROM [Player] AS [t]
CROSS JOIN [Player] AS [p]
WHERE [t].[SportId] = @__s1_Id_0
```
Same problems, but here an exception is thrown because [t].[SportId] doesn't exist

vuorinem:
+1 (just ran into this issues with very similar query)

mejje:
Is there a known workaround for this bug?

srollinet:
You can join manually rather than using the navigation properties.

```c#
var test1b = (from s in context.Sports
			  join t in context.Teams on s.Id equals t.SportId
			  join p in context.Players on t.Id equals p.TeamId
			  where s.Id == 1
			  select p.Name).ToList();
```
has the expected result

adrien-constant:
Any ETA on this?

The workaround works, but this prevents code reuse and that's a pain to rewrite the code with manual joins when you want to call SelectMany.

rowanmiller:
@adrien-constant this is on our list of issues to fix for our next pre-release (it's being fixed as part of a larger area of work on navigation property support in query)

maumar:
fixed in 7d27bf6527391cd46dbb09210f7e40f7998e549e

PR #4737, opened 2016-03-09
Fix to #3889 - Query: [Nav Prop Translation] SelectMany generates wrong SQL

maumar:
Issue was that we were applying our standard navigation expansion logic to SelectMany queries (e.g. customers.SelectMany(c => c.Orders)

This resulted in queries like:

SELECT * FROM Orders as o

which is correct for the basic case, however gives invalid SQL for slightly complex scenarios, e.g. with filter:
customers.Where(c => c.CustomerPK == 1).SelectMany(c => c.Orders)

SELECT * FROM Orders as o
WHERE o.CustomerPK = 1

Fix is to always translate the navigations inside SelectMany into inner joins instead.
This will produce queries like:

SELECT * FROM Customers as c
JOIN Orders as o ON c.Id = o.CustomerFK
WHERE c.CustomerPK = 1

This will have to be modified when Many-Many relationships are introduces, as some of the navigations will have to be translated into CROSS JOINs

Also added template visitor for transforming QueryModel as we now have two instances where this functionality is used.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -350,6 +350,17 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
                             if (navigations.Any())
                             {
+                                var outerQuerySourceReferenceExpression = new QuerySourceReferenceExpression(qs);
+
+                                if (_navigationRewritingQueryModelVisitor.AdditionalFromClauseBeingProcessed != null 
+                                    && navigations.Last().IsCollection())
+                                {
+                                    return RewriteSelectManyNavigationsIntoJoins(
+                                        outerQuerySourceReferenceExpression,
+                                        navigations,
+                                        _navigationRewritingQueryModelVisitor.AdditionalFromClauseBeingProcessed);
+                                }
+
                                 if ((navigations.Count == 1)
                                     && navigations[0].IsDependentToPrincipal())
                                 {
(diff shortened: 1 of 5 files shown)


maumar:
@anpete

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -536,51 +546,13 @@ var navigationJoin
...
-                    }
-
-                    joinClause.InnerKeySelector = innerKeySelector;
+                    QuerySourceReferenceExpression innerQuerySourceReferenceExpression;
+                    var joinClause = BuildJoinFromNavigation(
+                        querySourceReferenceExpression,
+                        navigation,
+                        targetEntityType,
+                        optionalNavigationInChain,
+                        out innerQuerySourceReferenceExpression);
```
maumar: I don't like the out parameter, but couldn't figure out a nice way to factor this method. Better to return KVP instead or is this fine?

anpete: This is fine for private impl.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/QueryModelVisitor.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using Microsoft.EntityFrameworkCore.Utilities;
+using JetBrains.Annotations;
+using Remotion.Linq;
+using Remotion.Linq.Clauses;
+using Remotion.Linq.Parsing;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
+{
+    public class QueryModelVisitor : QueryModelVisitorBase
```
anpete: How about ExpressionTransformingQueryModelVisitor? Can the pattern be applied anywhere else?

maumar: done, the only other place I could find was EntityQueryModelVisitor, but I don't think the pattern would be that useful there

maumar:
@anpete new version up

anpete:
:shipit:

PR #4740, opened 2016-03-09
[SMALL, Reveng] Add ability to deal with temporal tables for SQL Server 2016

### LINKED ISSUE #4697: RevEng: SQL Server Provider: Update tables metadata query for SQL Server 2016
SQL Server 2016 has a concept of a "System-Versioned" table. Each such table, say `XYZ`, is backed by an `XYZ_Archive` table. But the `XYZ_Archive` table does not have a primary key - so we can't generate an EntityType for it. This is correct as that table is automatically filled in by the system, but to avoid warning messages about all of these tables we should exclude them from the metadata query which returns the list of tables.

In SQL Server 2016, there is a new column on `sys.tables` called `temporal_type`. If the value for that is 1 then it's a backing table (called a "History Table" by SQL Server). But this column does not exist on earlier versions of SQL Server so we need to do something like:

```
IF CAST((select serverproperty(‘ProductMajorVersion’)) AS INT) >= 13
BEGIN
-- execute query with WHERE temporal_type <> 1
END
ELSE
BEGIN
-- execute query we have now
END
```

lajones:
Fixes #4697.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
@@ -17,6 +17,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
     public class SqlServerDatabaseModelFactory : IDatabaseModelFactory
     {
         private SqlConnection _connection;
+        private Version _serverVersion;
         private TableSelectionSet _tableSelectionSet;
         private DatabaseModel _databaseModel;
         private Dictionary<string, TableModel> _tables;
@@ -71,6 +72,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
 
                 _databaseModel.DatabaseName = _connection.Database;
 
+                _serverVersion = GetServerVersion();
                 if (SupportsSequences)
                 {
                     GetSequences();
@@ -86,18 +88,21 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
             }
         }
 
-        private bool SupportsSequences
+        private Version GetServerVersion()
         {
-            get
+            Version v;
+            if (Version.TryParse(_connection.ServerVersion, out v))
             {
-                Version v;
-                if (Version.TryParse(_connection.ServerVersion, out v))
-                {
-                    return v.Major >= 11;
-                }
-                return false;
+                return v;
             }
+
+            return null;
         }
+
+        private bool SupportsSequences => _serverVersion != null && _serverVersion.Major >= 11;
+
+        private string TemporalTableWhereClause =>
+            _serverVersion != null && _serverVersion.Major >= 13 ? " AND t.temporal_type <> 1" : string.Empty;
 
         private void GetDefaultSchema()
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -350,8 +357,9 @@ private void GetIndexes()
     INNER JOIN sys.columns c ON ic.object_id = c.object_id AND c.column_id = ic.column_id
     INNER JOIN sys.tables t ON t.object_id = i.object_id
 WHERE object_schema_name(i.object_id) <> 'sys'
-    AND object_name(i.object_id) <> '" + HistoryRepository.DefaultTableName + @"'
-ORDER BY object_schema_name(i.object_id), object_name(i.object_id), i.name, ic.key_ordinal";
+    AND object_name(i.object_id) <> '" + HistoryRepository.DefaultTableName + @"'" +
+    TemporalTableWhereClause +
+"ORDER BY object_schema_name(i.object_id), object_name(i.object_id), i.name, ic.key_ordinal";
```
lajones: Note: have added a space before `ORDER BY` - so that there is whitespace between it and the previous clause.

rowanmiller:
@natemcmaster  can you take a look at this one

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -71,6 +72,7 @@ public virtual DatabaseModel Create(string connectionString, TableSelectionSet t
 
                 _databaseModel.DatabaseName = _connection.Database;
 
+                _serverVersion = GetServerVersion();
```
natemcmaster: Could you just make this a one-liner? 
`Version.TryParse(_connection.ServerVersion, out _serverVersion)`

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -86,19 +88,22 @@ public virtual DatabaseModel Create(string connectionString, TableSelectionSet t
...
-                    return v.Major >= 11;
-                }
-                return false;
+                return v;
             }
+
+            return null;
         }
 
+        private bool SupportsSequences => _serverVersion != null && _serverVersion.Major >= 11;
```
natemcmaster: `_serverVersion?.Major >= 11` ..and elsewhere

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
lajones: Sure.

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
lajones: OK.

natemcmaster:
:shipit:

lajones:
Checked in with commit e96af05.

PR #4750, opened 2016-03-10
Update to Reduce magic in the interaction of EF Core and DI

ajcvickers:
See issue #4668. This is an update following a discussion yesterday that moves all configuration onto DbContextOptions and makes this the only object passed to the DbContext constructor. This allows the builder pattern that already exists for DbContextOptions to be used for external services and for setting the internal service provider. AddDbContext does this automatically for external services, but a DbContextOptions object can also be easily built and either configured in DI or passed directly to the context with new.

In addition, the UseMyProvider methods have been updated to use nested closure for provider-specific configuration. This makes for a very nice chaining story.

There is a new overload of AddDbContext that makes the application service provider available to teh configuration delegate. This is rarely needed, but one use is to set the application service provider to be used as the internal EF service provider if this is needed/desired.

Still to do:
- Look at negative cases when services are not setup and create better exception messages.
- Rename AddSqlServer, etc to AddEntityFrameworkSqlServer, etc.
- Extract diagnostics as one of the external services on DbContextOptions.


src/Microsoft.EntityFrameworkCore.SqlServer/Extensions/SqlServerDbContextOptionsExtensions.cs
```diff
@@ -16,11 +17,14 @@ public static class SqlServerDbContextOptionsExtensions
...
+        /// <param name="optionsBuilder"> A builder for setting options on the context. </param>
         /// <param name="connectionString"> The connection string of the database to connect to. </param>
-        /// <returns> An options builder to allow additional SQL Server specific configuration. </returns>
-        public static SqlServerDbContextOptionsBuilder UseSqlServer(
-            [NotNull] this DbContextOptionsBuilder optionsBuilder, [NotNull] string connectionString)
+        /// <param name="sqlServerOptionsAction">An optional action to allow additional SQL Server specific configuration.</param>
+        /// <returns> The options builder so that further configuration can be chained. </returns>
+        public static DbContextOptionsBuilder UseSqlServer(
+            [NotNull] this DbContextOptionsBuilder optionsBuilder, 
+            [NotNull] string connectionString,
```
anpete: Should this just happen in the action now? Or, is this the pattern for required config?

ajcvickers: Yep. This is the pattern for required config.

anpete:
Question: Do we still need the DbContextOptions/Builder that are generic on TContext? What are they used for?

anpete:
:shipit: :tada:

ajcvickers:
Merged

Grinderofl:
Referencing #4784 

This change has caused application service provider resolution to fail when adding plain DbContext (rather than subclassed one) because the default constructor is preferred. May be for DependencyInjection

rmja:
@rowanmiller One could argue that this change should be emphasized on the https://github.com/aspnet/Announcements repo, as it breaks code when one resolves the db provider in `AddDbContext<T>()` like this:

```
services.AddDbContext<MyContext>(options =>
{
        options.UseSqlServer(connectionString);
});
```

As `MyContext` now _requires_ a constructor similar to this one:

```
public MyContext(DbContextOptions options)
        : base(options)
{
}
```

PR #4757, opened 2016-03-10
Make AddMyProvider methods act like top-level service methods

ajcvickers:
Since these now hang directly off IServiceCollection they should behave in the same way as other methods that are here. Specifically:
- Renamed to indicate that they are methods for adding EntityFramework with a specific provider
- Add their dependencies automatically--that is, call AddEntityFramework

This means that AddEntityFramework is now a infrastructure method, like AddRelational, that is called by providers. It has therefore been moved to the Infrastructure namespace.


ajcvickers:
Could somebody take a quick look at this? It's a very straightforward change and once it is reviewed I can start merging the D.I. changes.

anpete:
:shipit:

anpete:
Might be nice to have nested closure here, too

ajcvickers:
Merged

PR #4760, opened 2016-03-10
Tiny: Move DbContextOptions to root namespace

ajcvickers:
Since it will now be referenced very commonly by applications.


anpete:
:shipit:

ajcvickers:
Merged

Issue #3289, opened 2015-10-01
[Inheritence] Setting base type can merge 2 different relationships


smitpatel:
When base type and derived type both have relationship with one entity and if the relationships use same foreignkey properties (configured by conventions) then while setting base type, relationship from derived type is removed even though both relationships are different (navigationToDependent is different)

Repro:
````
public class AnEntity
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public BaseReference BaseReference { get; set; }
        public DerivedReference DerivedReference { get; set; }
    }

    public class BaseReference
    {
        [NotMapped]
        public int Id { get; set; }
        public string Name { get; set; }

        public int? BaseParentId { get; set; }
        [NotMapped]
        public AnEntity BaseParent { get; set; }
    }

    public class DerivedReference : BaseReference
    {
    }

    public class MyContext : DbContext
    {
        public DbSet<AnEntity> DerivedEntities { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(
                @"Server=(localdb)\MSSQLLocalDB;Database=_TestApp6;Integrated Security=True;MultipleActiveResultSets=True;Connect Timeout=30");
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<AnEntity>().HasOne(e => e.BaseReference).WithOne();
            modelBuilder.Entity<AnEntity>().HasOne(e => e.DerivedReference).WithOne();

            modelBuilder.Entity<AnEntity>().HasKey(e => e.Id);

            modelBuilder.Entity<DerivedReference>().BaseType<BaseReference>();

            modelBuilder.Entity<BaseReference>().HasKey(e => e.Id);
        }
    }
`````
```
Console.WriteLine(context.Model.GetEntityType(typeof(BaseReference)).GetForeignKeys().Count());
Console.WriteLine(context.Model.GetEntityType(typeof(DerivedReference)).GetForeignKeys().Count());

//expected output
1
2
//actual output
1
1
```

smitpatel:
This does not repro anymore :relieved: 
@AndriySvyryd - This was reported during beta 8 when we did not have `BaseTypeDiscoveryConvention`. In latest code, base type is set even before relationships are configured. Any test scenario to cover this?

AndriySvyryd:
[...]

PR #4761, opened 2016-03-10
Add test: Setting base type does not merge relationships between base…

smitpatel:
… type and derived type if they are different

Test for #3289. It already fixed in latest code.


anpete:
:shipit:

PR #4762, opened 2016-03-10
[TINY] Fixing small issue in SqlTranslatingExpressionVisitor 

maumar:
...when we would get null argument exception when trying to translate coalesce expression whose arguments were not translatable to sql (i.e. needed to be client eval-ed)


maumar:
@anpete

anpete:
:shipit:

PR #4763, opened 2016-03-10
Explicitly set UTF8 encoding on migration files output

### LINKED ISSUE #4574: Wrong latin SQL encoding in migrations
When I add a migration with SQL command and latin characters, I get them replaced by "?" when updating database.

Let's see an example:

    public partial class TestEncodingMigration : Migration
    {
        {
            migrationBuilder.Sql(@"some text with latin chars : á é í ñ");
        }

        {
        }
    }

And this is the script I get from that migration:

    PM> dnx ef migrations script <some prev migration> -c MyDbContext
    
    some text with latin chars: ? ? ? ?
    ;
    
    GO
    
    INSERT INTO [__EFMigrationsHistory] ([MigrationId], [ProductVersion])
    VALUES (N'20160216162901_TestEncodingMigration', N'7.0.0-rc1-16348');

As you can see, all latin characters have been replaced by "?"

natemcmaster:
Fix #4574
cc @bricelam

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/Migrations/Design/MigrationsScaffolder.cs
+++ b/src/Microsoft.EntityFrameworkCore.Commands/Migrations/Design/MigrationsScaffolder.cs
@@ -289,7 +289,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Design
                 }
 
                 _logger.Value.LogInformation(CommandsStrings.RevertingSnapshot);
-                File.WriteAllText(modelSnapshotFile, modelSnapshotCode);
+                File.WriteAllText(modelSnapshotFile, modelSnapshotCode, Encoding.UTF8);
             }
 
             return files;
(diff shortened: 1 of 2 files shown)


bricelam:
From the docs:
> This method uses UTF-8 encoding without a Byte-Order Mark (BOM). ... If it is necessary to include a UTF-8 identifier at the beginning of a file, use the WriteAllText method overload with UTF8 encoding.

bricelam:
:shipit:

bricelam:
Also, were there any other calls to `File.WriteAllText()` in the product? If so, we should do the same there.

natemcmaster:
I believe it's actually a quirk in the VS text editor. In some cases, if the file doesn't have the BOM and you add UTF8 characters, it still saves as ASCII.

And yes, I found one more "writealltext" call in reveng. Added the UTF8 option there too.

PR #4766, opened 2016-03-11
Fix build break

smitpatel:
- AddCaching is split into 2 methods. Method useful for us AddMemoryCache
- AddLogging & AddMemoryCache methods are void functions now

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Extensions.DependencyInjection
             [CanBeNull] Action<IServiceProvider, DbContextOptionsBuilder> optionsAction)
             where TContext : DbContext
         {
-            serviceCollection.AddCaching();
+            serviceCollection.AddMemoryCache();
             serviceCollection.AddLogging();
 
             serviceCollection.TryAddSingleton(p => DbContextOptionsFactory<TContext>(p, optionsAction));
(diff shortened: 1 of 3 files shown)


smitpatel:
@natemcmaster

ajcvickers:
We should fix that AddMemoryCache returns void and so breaks chaining, but :shipit: for for now.

natemcmaster:
cc @javiercn  who made the change in https://github.com/aspnet/Caching/commit/a6c044c1590b16c98df1cc6d01eca1140a86ee99.

smitpatel:
`AddLogging` also returns void. `AddOptions` return service collection. We have inconsistencies.

natemcmaster:
Are you using cidev or volatiledev?

javiercn:
@ajcvickers AddMemoryCache returning void is intentional. We only want methods with builders (like EF, Identity or MVC) to return something. We've done this change across all repositories. @Eilon can confirm

ajcvickers:
@javiercn @Eilon @divega @rowanmiller We often write code like this:
```C#\
var appServiceProivder = new ServiceCollection()
    .AddLogging()
    .AddCaching()
    .BuildServiceProvider();
```
This is really clean and easy to read. When we are no longer able to chain we have to write this instead:
```C#
serviceCollection.AddLogging();
serviceCollection.AddCaching();
var appServiceProivder = serviceCollection.BuildServiceProvider();
```
This additional clutter detracts from the overall readability of the code.

Furthermore, the benefits of this kind of pattern are cascading. For example, it lets us use expression body syntax to again create very clear, very readable code:
```C#
protected virtual void ConfigureServices([NotNull] IServiceCollection services)
        => services
        .AddSingleton<CSharpHelper>()
            .AddSingleton<CSharpMigrationOperationGenerator>()
            .AddSingleton<CSharpSnapshotGenerator>()
            .AddSingleton<MigrationsCodeGenerator, CSharpMigrationsGenerator>()
            .AddScaffolding()
            .AddLogging();
```
This was the reason when these methods where designed a couple of years ago they allowed chaining. I'm not sure why we think this is no longer relevant. Also, returning the service collection as opposed to void doesn't force anybody to use chaining if they don't want to, so what is the downside?

Eilon:
I chatted with @ajcvickers this morning, we'll discuss this over email.

PR #4767, opened 2016-03-11
Add PMC cmdlet support for xproj

### LINKED ISSUE #978: Commands: PMC cmdlets for xproj
Having to drop down to the cmd line and use dnx is very jarring compared to staying in VS.

We need to do some work to enable this, but most of it is in NuGet and xproj tooling:
* https://github.com/NuGet/Home/issues/1174
* https://github.com/aspnet/Tooling-Internal/issues/443

natemcmaster:
The powershell commands invoke dotnet-ef when executing on an xproj. It will throw errors if dotnet-ef is not installed in the tools section of the targeted project.

Fixes #978 

cc @bricelam

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
+++ b/src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
@@ -43,7 +43,24 @@ function Use-DbContext {
 
     $dteProject = GetProject $Project
     $dteStartupProject = GetStartupProject $StartupProject $dteProject
-    $contextTypeName = InvokeOperation $dteStartupProject $Environment $dteProject GetContextType @{ name = $Context }
+    if (IsDotNetProject $dteProject) {
+        $contextTypes = GetContextTypes $Project $StartupProject $Environment
+        $candidates = $contextTypes | ? { $_ -ilike "*$Context" }
+        $exactMatch = $contextTypes | ? { $_ -eq $Context }
+        if ($candidates.length -gt 1 -and $exactMatch -is "String") {
+            $candidates = $exactMatch
+        }
+        
+        if ($candidates.length -lt 1) {
+            throw "No DbContext named '$Context' was found"
+        } elseif ($candidates.length -gt 1 -and !($candidates -is "String")) {
+            throw "More than one DbContext named '$Context' was found. Specify which one to use by providing its fully qualified name."
+        }
+
+        $contextTypeName=$candidates
+    } else {
+        $contextTypeName = InvokeOperation $dteStartupProject $Environment $dteProject GetContextType @{ name = $Context }
+    }
 
     $EFDefaultParameterValues.ContextTypeName = $contextTypeName
     $EFDefaultParameterValues.ProjectName = $dteProject.ProjectName
@@ -105,15 +122,24 @@ function Add-Migration {
     $dteProject = $values.Project
     $contextTypeName = $values.ContextTypeName
 
-    $artifacts = InvokeOperation $dteStartupProject $Environment $dteProject AddMigration @{
-        name = $Name
-        outputDir = $OutputDir
-        contextType = $contextTypeName
-    }
+    if (IsDotNetProject $dteProject) {
+        $options = ProcessCommonDotnetParameters $dteProject $dteStartupProject $Environment $contextTypeName
+        if($OutputDir) {
+            $options += "--output-dir", (NormalizePath $OutputDir)
+        }
+        $files = InvokeDotNetEf $dteProject -Json migrations add $Name @options
+        $DTE.ItemOperations.OpenFile($files.MigrationFile) | Out-Null
+    } else {
+        $artifacts = InvokeOperation $dteStartupProject $Environment $dteProject AddMigration @{
+            name = $Name
+            outputDir = $OutputDir
+            contextType = $contextTypeName
+        }
 
-    $artifacts | %{ $dteProject.ProjectItems.AddFromFile($_) | Out-Null }
-    $DTE.ItemOperations.OpenFile($artifacts[0]) | Out-Null
-    ShowConsole
+        $artifacts | %{ $dteProject.ProjectItems.AddFromFile($_) | Out-Null }
+        $DTE.ItemOperations.OpenFile($artifacts[0]) | Out-Null
+        ShowConsole
+    }
 
     Write-Output 'To undo this action, use Remove-Migration.'
 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
```diff
@@ -489,6 +562,31 @@ function ProcessCommonParameters($startupProjectName, $projectName, $contextType
...
+    }
+    if ($contextTypeName) {
+        $options += "--context",$contextTypeName
+    }
+    return $options
+}
+
+function IsDotNetProject($project) {
+    $xprojKind="{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}"
+    return $project.Kind -eq $xprojKind
```
bricelam: Is there a difference between the .NET Core and ASP.NET Core project kinds?

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
```diff
@@ -105,15 +122,23 @@ function Add-Migration {
...
-        name = $Name
-        outputDir = $OutputDir
-        contextType = $contextTypeName
-    }
+    if (IsDotNetProject $dteProject) {
+        $options = ProcessCommonDotnetParameters $dteProject $dteStartupProject $Environment $contextTypeName
+        if($OutputDir) {
+            $options += "--output-dir", (NormalizePath $OutputDir)
+        }
+        InvokeDotNetEf $dteProject migrations add $Name @options | Out-Null
```
bricelam: Can we open the file?

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
```diff
@@ -503,6 +601,53 @@ function ShowConsole {
...
+    } catch {
+        Write-Verbose $_.Exception.Message
+        throw "Invalid JSON file in $projectJson"
+    }
+    if ($projectDef.tools) {
+        $t=$projectDef.tools | Get-Member dotnet-ef
+    }
+    if (!$t) {
+        $projectName = $project.ProjectName
+        throw "Cannot execute this command because 'dotnet-ef' is not installed in project '$projectName'. Add 'dotnet-ef' to the 'tools' section in project.json."
```
bricelam: This limitation feels weird. We should brainstorm about alternatives

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
```diff
@@ -503,6 +601,53 @@ function ShowConsole {
...
+        if ($LASTEXITCODE -ne 0) {
+            if (!($verboseOutput) -and $output) {
+                # most often occurs when dotnet-ef didn't install
+                throw $output
+            }
+            throw $verboseOutput
+        }
+        Write-Verbose $($output -join [Environment]::NewLine)
+        # dotnet commands log verbose output to stderr
+        Write-Verbose $($verboseOutput -join [Environment]::NewLine)
```
bricelam: Hmm, Warning, Information, etc. semantics will be lost.

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
```diff
@@ -105,15 +122,23 @@ function Add-Migration {
     $dteProject = $values.Project
     $contextTypeName = $values.ContextTypeName
 
-    $artifacts = InvokeOperation $dteStartupProject $Environment $dteProject AddMigration @{
-        name = $Name
-        outputDir = $OutputDir
-        contextType = $contextTypeName
-    }
+    if (IsDotNetProject $dteProject) {
```
bricelam: Would it be cleaner to branch in InvokeOperation? I know it would have to change a bit... Maybe using an enum-like structure to abstract the differences. Example: `InvokeOperation $operation.AddMigration ...` Where `AddMigration` had string properties for `migrations add` and `AddMigration`

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
natemcmaster: Yes, this is the part I like the least about piping through to dotnet-ef. It's quirky and it leaves room for enhancement.

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
natemcmaster: > Would it be cleaner to branch in InvokeOperation?

Maybe. This was my original approach, but found it difficult to make it any cleaner than my current implementation. If only we could use powershell classes... :-/

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
natemcmaster: We would need to parse the log output of the command to find the location it used....seemed like an unsafe precedent.

But on this note (related to the logging below), can make enhancement to dotnet-ef to produce more powershell-friendly output? e.g. dbcontext list --json is actually pretty useful.

bricelam: Adding more `--json` flags on commands would be great! They're there for situations like this.

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
bricelam: We could adopt a node-style output to help with parsing:

    warn: This is a warning
    info: This is information

Or we could pass through ANSI escape sequences and parse them out to determine logging level.

(I don't actually like either of these ideas)

natemcmaster: Yeah, not sure I like either but can't think of anything better. I filed https://github.com/aspnet/EntityFramework/issues/4771 to address this

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
natemcmaster: I haven't been able to get an installation of WTE to work yet...so not sure.

natemcmaster: How about instead, `$project.FileName.EndsWith(".xproj")`?

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
natemcmaster: Updated. Added a `--json` flag to "migrations add" and updated the cmdlet to use it.

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
natemcmaster: Alternative approaches:
 - install dotnet-ef into the project
 - bundle dotnet-ef in the commands package. Execute `corehost dotnet-ef ...`
 - make dotnet-ef a console app as well

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
bricelam: :+1: Nuget has [code like that](https://github.com/NuGet/NuGet.BuildTasks/blob/dev/src/Microsoft.NuGet.Build.Tasks/Microsoft.NuGet.targets#L22).

bricelam:
:shipit: This is good enough for RC2.

We talked about exploring the possibility of bypassing dotnet-ef in Triage. Ideally, the PowerShell commands for csproj and xproj would go down more-or-less the same code path. This would depend heavily on .NET Core unification...

PR #4769, opened 2016-03-11
Throw helpful exceptions when services or provider are not configured

### LINKED ISSUE #4665: Improve exception message when using derived DbContext with no services
This was reported in https://github.com/aspnet/dnx/issues/3366. I haven't tried to repro it but apparently if you create a context with no `OnConfiguring()` method and try to use it this way:

``` C#
var command = _ctx.Database.GetDbConnection();
```

You can get an `InvalidOperationException` with the message __"No service for type 'Microsoft.Data.Entity.Storage.IRelationalConnection' has been registered"__.

Currently the most common reason this can happen is that the user has only configured the derived DbContext to be used from the DI container but then he uses the parameterless constructor to create an instance.

ajcvickers:
The common case here is a context that does not have OnConfiguring and does not pass in a DbContextOptions to its constructor, or the DbContextOptions passed does has not had any providers configured on it.

It is also possible to get no EF services at all, but this now requires making a call to UseInternalServiceProvider with a service provider that does not have EF services in it.

Also fixes #4665.

@rowanmiller for exception message wording

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Infrastructure/AccessorExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/Infrastructure/AccessorExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 
@@ -41,7 +42,17 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
         /// <param name="accessor"> The object exposing the service provider. </param>
         /// <returns> The requested service. </returns>
         public static TService GetService<TService>([NotNull] this IInfrastructure<IServiceProvider> accessor)
-            => Check.NotNull(accessor, nameof(accessor)).Instance.GetRequiredService<TService>();
+        {
+            Check.NotNull(accessor, nameof(accessor));
+
+            var service = accessor.Instance.GetService<TService>();
+            if (service == null)
+            {
+                throw new InvalidOperationException(CoreStrings.NoProviderConfigured);
+            }
+
+            return service;
+        }
 
         /// <summary>
         ///     <para>
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
@@ -187,10 +187,10 @@
     <value>The database providers {storeNames}are configured. A context can only be configured to use a single database provider.</value>
   </data>
   <data name="NoProviderConfigured" xml:space="preserve">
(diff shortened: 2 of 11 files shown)


src/Microsoft.EntityFrameworkCore/Infrastructure/AccessorExtensions.cs
```diff
@@ -41,7 +42,17 @@ public static class AccessorExtensions
...
         /// <returns> The requested service. </returns>
         public static TService GetService<TService>([NotNull] this IInfrastructure<IServiceProvider> accessor)
-            => Check.NotNull(accessor, nameof(accessor)).Instance.GetRequiredService<TService>();
+        {
+            Check.NotNull(accessor, nameof(accessor));
+
+            var service = accessor.Instance.GetService<TService>();
+            if (service == null)
+            {
+                throw new InvalidOperationException(CoreStrings.NoProviderConfigured);
```
anpete: Isn't this method too general a place to throw this specific error? It seems like we also use the method to resolve non provider services.

ajcvickers: We use it to resolve EF services, and with the recent changes to D.I. registration if the app has not used a provider, then there typically will not be any EF services at all registered. The solution is to use a provider. So while it may not technically be a provider service that we failed to resolve, the guidance to the user is still the same.

While it is possible, but now unlikely, that the app may have EF services registered but not provider services, this method would then pass for all EF services but fail for provider services, in which case the message is still appropriate.

The only other case is that this method is being used for some other non-EF service, and that service is not registered. We could try to detect this case but it seems pretty obscure so I didn't attempt to do so.

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -187,10 +187,10 @@
...
+    <value>No database provider has been configured for this DbContext. A provider can be configured by overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider. If AddDbContext is used, then also ensure that your DbContext type accepts a DbContextOptions object in its constructor.</value>
   </data>
-  <data name="NoProviderServices" xml:space="preserve">
-    <value>No database providers are available. Ensure that database provider services are added inside the call to AddEntityFramework on your ServiceCollection.</value>
+  <data name="NoEfServices" xml:space="preserve">
+    <value>Entity Framework services have not been added to the internal service provider. Either remove the call to UseInternalServiceProvider so that EF will manage its own internal services, or ensure that the service provider has Entity Framework and provider services added using AddEntityFrameworkMyProvider where MyProvider is the name of the database provider being used.</value>
   </data>
```
rowanmiller: > Entity Framework services have not been added to the internal service provider. Either remove the call to UseInternalServiceProvider so that EF will manage its own internal services, or use the method from your database provider to add the required services to the service provider (e.g. AddEntityFrameworkSqlServer).

rowanmiller:
:shipit:

PR #4777, opened 2016-03-12
Workaround for CLI issue where restore fails with

pranavkm:
An item with the same key has already been added. Key: System.Threading
Unspecified exception.


pranavkm:
cc @davidfowl

PR #4813, opened 2016-03-16
Update dotnet-ef to target netstandardapp1.5

bricelam:


src/dotnet-ef/OperationExecutor.cs
```diff
@@ -171,9 +171,9 @@ private ProjectContext GetCompatibleProjectContext(string projectPath)
             if (framework == null)
             {
                 throw new OperationException(
-                    "The project '" + projectFile.Name + "' doesn't target a framework compatible with DNX Core 5.0. You must " +
-                    "target a compatible framework like 'dotnet5.4' before using the Entity Framework Core .NET Core CLI " +
-                    "Commands.");
+                    "The project '" + projectFile.Name + "' doesn't target a framework compatible with .NET Standard " +
+                    "App 1.5. You must target a compatible framework like 'netstandard1.3' before using the Entity " +
```
Eilon: Recommend changing `like` to `such as` because this is a proper example.

src/dotnet-ef/OperationExecutor.cs
```diff
@@ -171,9 +171,9 @@ private ProjectContext GetCompatibleProjectContext(string projectPath)
             if (framework == null)
             {
                 throw new OperationException(
-                    "The project '" + projectFile.Name + "' doesn't target a framework compatible with DNX Core 5.0. You must " +
-                    "target a compatible framework like 'dotnet5.4' before using the Entity Framework Core .NET Core CLI " +
-                    "Commands.");
+                    "The project '" + projectFile.Name + "' doesn't target a framework compatible with .NET Standard " +
```
natemcmaster: Are we not globalizing strings for dotnet-ef?

bricelam:
Recalling. This doesn't solve anything. We need to update our command to follow the guidance in dotnet/cli#1970 anyway.

src/dotnet-ef/OperationExecutor.cs
bricelam: Not yet. When they do, we will. :wink:

PR #4873, opened 2016-03-21
Set data directory to output path

### LINKED ISSUE #4765: CLI commands puts SQLite files in the nuget folder containing dotnet-ef
A relative path in the SQLite connection string + dotnet-ef  = the root for the relative path is the folder containing dotnet-ef.dll. This is because Microsoft.Data.Sqlite relies on AppContext.BaseDirectory, [which is Package.Current.InstalledLocation.Path](https://github.com/dotnet/corefx/blob/master/src/System.AppContext/src/System/AppContext.cs#L28).

image

To reproduce:
```c#
    public class TestContext : DbContext
    {
        {
            optionsBuilder.UseSqlite("Filename=./test.db");
        }
    }
```
```
dotnet ef migrations add First
dotnet ef database update
```

natemcmaster:
Fix #4765 

cc @bricelam

### DIFF
--- a/src/dotnet-ef/OperationExecutor.cs
+++ b/src/dotnet-ef/OperationExecutor.cs
@@ -28,6 +28,9 @@ namespace Microsoft.EntityFrameworkCore.Commands
         private readonly LazyRef<DatabaseOperations> _databaseOperations;
         private readonly LazyRef<MigrationsOperations> _migrationsOperations;
 
+        private const string DataDirEnvName = "ADONET_DATA_DIR";
+        private const string DefaultConfiguration = "Debug";
+
         public OperationExecutor([CanBeNull] string startupProject, [CanBeNull] string environment)
         {
             var project = Path.Combine(Directory.GetCurrentDirectory(), Project.FileName);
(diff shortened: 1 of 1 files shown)


src/dotnet-ef/OperationExecutor.cs
```diff
@@ -58,6 +61,13 @@ public OperationExecutor([CanBeNull] string startupProject, [CanBeNull] string e
 
             Reporter.Verbose.WriteLine("Build succeeded.".Bold().Black());
 
+            // TODO parameterize "configuration"
```
natemcmaster: Is this already being tracked in an issue?

bricelam: Nope. I think I saw this assumption--that design-time always uses `Debug`--in a few other places across the stack.

src/dotnet-ef/OperationExecutor.cs
```diff
@@ -58,6 +61,13 @@ public OperationExecutor([CanBeNull] string startupProject, [CanBeNull] string e
...
+            var runtimeOutputPath = startupProjectContext.GetOutputPaths(DefaultConfiguration)?.RuntimeOutputPath;
+            if (!string.IsNullOrEmpty(runtimeOutputPath))
```
bricelam: When is this `null`/empty?

src/dotnet-ef/OperationExecutor.cs
```diff
@@ -58,6 +61,13 @@ public OperationExecutor([CanBeNull] string startupProject, [CanBeNull] string e
...
+            {
+                Environment.SetEnvironmentVariable(DataDirEnvName, runtimeOutputPath);
```
bricelam: Can you add a TODO to call `AppDomain.SetData("DataDirectory", ...)` on `net45`? It sounds like this tool will be going that direction...

bricelam:
:shipit:

src/dotnet-ef/OperationExecutor.cs
natemcmaster: I found odd corner cases, mostly due to restore errors. I think this is just due to the fact that we currently build against DependencyModel 1585. They've made lots of updates in the last 300 builds of cli.

natemcmaster:
Updated & merged

PR #4883, opened 2016-03-22
Stop registering DbContext as an internal service

ajcvickers:
Because when people use the same service provider for both the application and our internal services and don't create a derived DbContext class, then the two registrations can often collide. See #4750.

The fix is to use the type ICurrentDbContext internally which wraps the context for the current scope.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Internal/ICurrentDbContext.cs
+++ b/src/Microsoft.EntityFrameworkCore/Internal/ICurrentDbContext.cs
@@ -1,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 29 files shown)


src/Microsoft.EntityFrameworkCore/Internal/ICurrentDbContext.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.EntityFrameworkCore.Internal
+{
+    public interface ICurrentDbContext
```
divega: `IDbContextAdapter`? :trollface:

divega:
:shipit: I am not 100% about the name of the interface but I cannot think of anything better.

PR #4886, opened 2016-03-22
Allow ILoggers to be resolved before context has been created

ajcvickers:
We intercept ILogger creation to ensure EF uses the correct logger when the UseLoggerProvider method has been used. However, this meant that when sharing a service provider we would sometimes fail to provider a logger if the context had not yet been initialized. This change fixes that by falling back to regular service provider resolution.


AndriySvyryd:
:shipit:

PR #4888, opened 2016-03-22
[TINY] Fixing possible null reference in NavigationRewritingExpressionVisitor

maumar:
adding some missing null checks


maumar:
@anpete

anpete:
:shipit:

PR #4889, opened 2016-03-23
Update pipeline hot path optimizations

AndriySvyryd:
Improve perf of SqlServerTypeMapper.FindCustomMapping()
Improve perf of Property.FindContainingIndexes()
Restart parameter naming for each batch


ajcvickers:
:shipit:

PR #4892, opened 2016-03-23
Fix to #4547 - Query :: query source materialized as ValueBuffer rather than entity for some complex cases involving subqueries and conditional operator - causes NRE/InvalidCast

maumar:
Issue is that some queries reference result of SelectMany-GroupJoin-DefaultIfEmpty in a subquery scenario but those underlying query sources were not marked for materialization.
Fix is to add those additional sources to the list of queries to be materialized, similar to what we do for the non-subquery case.


maumar:
@anpete

anpete:
:shipit:

PR #4894, opened 2016-03-23
Commands: Avoid serializing load exceptions across boundary

### LINKED ISSUE #4796: Migrations :: exception thrown when trying to add migration using regular dotnet project (csproj)
Exception is as follows:

Add-Migration : Exception calling "CreateInstanceAndUnwrap" with "8" argument(s): "Type 'Microsoft.EntityFrameworkCore.Design.OperationException' in assembly 
'Microsoft.EntityFrameworkCore.Commands, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' is not marked as serializable."
At line:1 char:1
+ Add-Migration Mig01 -Verbose

bricelam:
Resolves #4796


lajones:
Add a comment in the code saying why this is necessary and then :shipit:

PR #4896, opened 2016-03-24
Make Attach never have side effect of setting dependent entities as Modified

ajcvickers:
This means that graph Attach calls behave as expected no matter which entity is attached first or which mechanism is used, and work correctly with shadow FKs.

It also means that re-parenting an attached dependent cannot be done by setting a new principal to point to it, and then attaching that principal. But attaching the principal and then setting the navigation property will work.

/cc @divega @rowanmiller


anpete:
:shipit:

PR #4901, opened 2016-03-24
Remove-Migration -Force

natemcmaster:
cc @bricelam 
Fix https://github.com/aspnet/EntityFramework/issues/4812.
Allows user to force remove the last migration, skipping over database check.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/Design/MigrationsOperations.cs
+++ b/src/Microsoft.EntityFrameworkCore.Commands/Design/MigrationsOperations.cs
@@ -141,7 +141,7 @@ namespace Microsoft.EntityFrameworkCore.Design
         }
 
         public virtual MigrationFiles RemoveMigration(
-            [CanBeNull] string contextType)
+            [CanBeNull] string contextType, bool force)
         {
             using (var context = _contextOperations.CreateContext(contextType))
             {
@@ -150,7 +150,7 @@ namespace Microsoft.EntityFrameworkCore.Design
 
                 var scaffolder = services.GetRequiredService<MigrationsScaffolder>();
 
-                var files = scaffolder.RemoveMigration(_projectDir, _rootNamespace);
+                var files = scaffolder.RemoveMigration(_projectDir, _rootNamespace, force);
 
                 _logger.Value.LogInformation(CommandsStrings.Done);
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.EntityFrameworkCore.Commands/Design/MigrationsOperations.cs
```diff
@@ -141,7 +141,7 @@ public class MigrationsOperations
         }
 
         public virtual MigrationFiles RemoveMigration(
-            [CanBeNull] string contextType)
+            [CanBeNull] string contextType, bool force = false)
```
bricelam: No default value. This is not a user-facing API, so it will probably do more harm (by being forgotten) than good.

src/Microsoft.EntityFrameworkCore.Commands/Migrations/Design/MigrationsScaffolder.cs
```diff
@@ -188,7 +188,7 @@ public class MigrationsScaffolder
                     : @namespace;
 
         // TODO: DRY (file names)
-        public virtual MigrationFiles RemoveMigration([NotNull] string projectDir, [NotNull] string rootNamespace)
+        public virtual MigrationFiles RemoveMigration([NotNull] string projectDir, [NotNull] string rootNamespace, bool force = false)
```
bricelam: No default value.

src/dotnet-ef/OperationExecutor.cs
```diff
@@ -135,8 +135,8 @@ public virtual void UpdateDatabase([CanBeNull] string targetMigration, [CanBeNul
             [CanBeNull] string contextType)
             => _migrationsOperations.Value.ScriptMigration(fromMigration, toMigration, idempotent, contextType);
 
-        public virtual MigrationFiles RemoveMigration([CanBeNull] string contextType)
-            => _migrationsOperations.Value.RemoveMigration(contextType);
+        public virtual MigrationFiles RemoveMigration([CanBeNull] string contextType, bool force = false)
```
bricelam: No default value.

bricelam:
:shipit:

src/Microsoft.EntityFrameworkCore.Commands/Design/MigrationsOperations.cs
natemcmaster: Good point. Will remove

bricelam: I realized this after finding a few places where `CancellationToken`s weren't being passed downstream.

PR #4905, opened 2016-03-25
Query: Async: Force serialized execution of concurrently executing co…

anpete:
…ntext-affinitized query Tasks.

cc @AndriySvyryd @divega


AndriySvyryd:
Could you add a test for Query and one for Update that could fail without this?

anpete:
@AndriySvyryd This is only for query right now. @smitpatel has tests coming that exercise the single query part of this. I will add one for the multi-query scenario.

AndriySvyryd:
:shipit:

PR #4906, opened 2016-03-25
[TINY] Reorder parameters in exceptions thrown from PropertyMappingValidationConvention

### LINKED ISSUE #4746: type and property name backwards in error message for invalid property
```
System.InvalidOperationException: 
The property 'Urls.List<string>' could not be mapped, because it is of type 'Pet' which is not a
supported primitive type or a valid entity type. Either explicitly map this property, or ignore it.
```

The type and property name are swapped in the error message.

Model Type:
```
    public class Pet
    {
        public int Id { get; set; }

        public Category Category { get; set; }

        [Required]

        public List<string> Urls { get; set; }

        public List<Tag> Tags { get; set; }

        [Required]
        public string Status { get; set; }
    }
```

AndriySvyryd:
Fixes #4746


divega:
:shipit:

PR #4907, opened 2016-03-25
NuGet Commands: Drop-Database

### LINKED ISSUE #2476: Commands: Drop Database

bricelam:
This includes changes from #4657. Only 03f95a27246c799f7388e11fc83b3d9bd0e2e87d needs review.

cc @natemcmaster @laskoviymishka (aka Rarity)

Resolves #2476

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
+++ b/src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
@@ -50,7 +50,7 @@ function Use-DbContext {
         if ($candidates.length -gt 1 -and $exactMatch -is "String") {
             $candidates = $exactMatch
         }
-        
+
         if ($candidates.length -lt 1) {
             throw "No DbContext named '$Context' was found"
         } elseif ($candidates.length -gt 1 -and !($candidates -is "String")) {
@@ -498,7 +498,7 @@ function Scaffold-DbContext {
 
         $artifacts | %{ $dteProject.ProjectItems.AddFromFile($_) | Out-Null }
         $DTE.ItemOperations.OpenFile($artifacts[0]) | Out-Null
-        
+
         ShowConsole
     }
 }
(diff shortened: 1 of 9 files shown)


src/dotnet-ef/DatabaseDropCommand.cs
```diff
@@ -39,9 +39,10 @@ private static int Execute(string context, string startupProject, string environ
             new OperationExecutor(startupProject, environment)
                 .DropDatabase(
                     context,
-                    message =>
+                    (database, dataSource) =>
                     {
-                        Reporter.Output.WriteLine(message);
+                        Reporter.Output.WriteLine(
+                            $"Are you sure you want to drop the database '{database}' on server '{dataSource}'? (y/N)");
                         var readedKey = Console.ReadKey().KeyChar;
```
natemcmaster: shouldn't we skip read altogether this if `--force` was used?

natemcmaster:
It looks like `--force` on dotnet-ef doesn't actually skip the prompt.

Otherwise, :shipit:

bricelam:
Thanks. I'll do a bit bashing on this too before merging.

src/Microsoft.EntityFrameworkCore.Commands/tools/EntityFramework.psm1
```diff
@@ -513,6 +513,70 @@ function Enable-Migrations {
...
+            InvokeDotNetEf $dteProject database drop @options | Out-Null
+        }
+    } else {
+        InvokeOperation $dteStartupProject $Environment $dteProject  DropDatabase @{
+            contextType = $contextTypeName
+            confirmCheck = [Func[string, string, bool]]{
+                param ($database, $dataSource)
+
+                return $PSCmdlet.ShouldProcess("$database on $dataSource")
+            }
```
bricelam: Bad new. This results in `Cannot serialize delegates over unmanaged function pointers, dynamic methods or methods outside the delegate creator's assembly.` :disappointed:

bricelam:
Recalling. We need to rethink how the confirmation works. One idea @natemcmaster had was to get the information in a separate command (e.g. `Get-DbContext` / `dotnet ef dbcontext show --json`), then prompt, then call drop.

PR #4911, opened 2016-03-25
Shadow properties created for Foreign key by copying over principal k…

### LINKED ISSUE #4895: Shadow FK created by replicating the property type from principal key are non-nullable
I tried applying `[ForeignKey(name)]` on a reference navigation property that didn't have a corresponding FK property in the CLR type to pick a column name that was different from the one that the convention would produce. 

This had the unexpected side effect that it made the FK non-nullable (which in turns activated cascade deletes, caused queries with `Include()` on that navigation property to translate to INNER JOINs, etc). 

None of these things happened if the name I specified in the attribute matched what the convention would pick.

I haven't dug into the implementation but I wonder if this affects more than just the nullability of the property, and if the relationship builder should produce nullable FKs by default. 

**Repro**

``` C#
using Xunit;

namespace ForeignKeyName
{
    public class Tests
    {
        [Fact]
        private static void FKAttribute_should_not_change_nullability_if_it_renames_property()
        {
            {
                var et = context.Model.FindEntityType(typeof(Post));
                var fk = et.FindProperty("AuthorId");
                Assert.True(fk.IsNullable); // fails
            }
        }

        [Fact]
        private static void FKAttribute_should_not_change_nullability_if_it_does_not_rename_property()
        {
            {
                var et = context.Model.FindEntityType(typeof(Post));
                var fk = et.FindProperty("ModifiedById");
                Assert.True(fk.IsNullable); // passes
            }
        }
    }

    {

        {
        }
    }

    public class Post
    {
        [ForeignKey("AuthorId")] // does not match by convention name
        public User PostedBy { get; set; }
        [ForeignKey("ModifiedById")] // matches by convention name
        public User ModifiedBy { get; set; }
    }

    public class User
    {
        public int Id { get; set; }
    }
}
```

smitpatel:
…ey properties types use nullable types

resolves #4895

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -1462,7 +1462,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
                     else if (type != null)
                     {
                         // TODO: Log that shadow property is created by convention
-                        propertyBuilder = Property(propertyName, type, ConfigurationSource.Convention);
+                        propertyBuilder = Property(propertyName, type.MakeNullable(), ConfigurationSource.Convention);
                     }
                     else
                     {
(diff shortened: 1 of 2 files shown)


ajcvickers:
:shipit:

PR #4913, opened 2016-03-27
Adds Timestamps, InstanceId and Duration to Diagnostic Source events

avanderhoorn:
This brings DiagnosticSource usage inline with what we have been doing elsewhere. 
 - **Timing**: usage has been converted from using a new instance of `StopWatch` to using `Stopwatch.GetTimestamp()`. This has been well [documented](https://github.com/aspnet/Hosting/pull/543) to have better performance than other alternatives. 
 - **TimeStamp**: `timestamp` has ben added to the begin/end payloads so that all consumers of logging and DiagnosticSource will be showing the same times when using EF based events
 - **Correlation**: `InstanceId` has ben added to the payload so that consumers can correlate begin and end messages together in a deterministic way without haven't to maintain state across events
     - Note: Strictly speaking other areas where we are using this, the Id is "global" to the instance of the object. In this case that dosn't work as technically the command object can be used multiple times. Semantically, that means that the Id is unique to the call not the instance of the containing object, but I *think?* the still holds given that its the instance of the command being executed.
 - **Duration**: `duration` has ben added to the payload so that consumers don't have to listen to both the begin and end message to easily be able to determine how long a request took... this has the potential to make the whole system more performant as it means that less people should have the need to tap the begin message
     - Note: I removing the `RelationalDiagnosticSourceMessage` and creating the `RelationalDiagnosticSourceBeforeMessage`/`RelationalDiagnosticSourceAfterMessage` instead. Strictly speaking I didn't need to do this, but it seemed weird to have `duration` on the begin message.
     - Note: I updated the centralized logging implementation to access the start/end timestamp instead of `elapsedMilliseconds` so that the calculations of the milliseconds would only occur as needed when logging is enabled. 

Note: I changed the position of when the events get run to be slightly more consistent with other areas where we are doing this with Try/Catch blocks. In these cases, we have been running the end event as the last thing inside of the try block... In this case doing that means that means that the main semantic difference is that we would be running the event before a potential connection close. That meant that I shifted the begin event to after the potential connection open. Given that the timing data we are looking at is around command execution time, this seemed liked a ok change.

Let me know if there is any feedback.

@anpete @divega

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalLoggerExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalLoggerExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Data.Common;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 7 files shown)


avanderhoorn:
@anpete Thanks!!!

PR #4917, opened 2016-03-28
Add support for building commands with transaction suppression in migrations

mikary:
Adds `MigrationCommandList` to support consistent transaction use/suppression in migrations. Changes `MigrationCommandListBuilder` (RelationalCommandListBuilder) to opt into command suppression when `EndCommand` is called. It is now the responsibility of the `Generate` method to terminate commands and statements.
cc @bricelam


src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandList.cs
```diff
@@ -0,0 +1,128 @@
...
+                        command.RelationalCommand.ExecuteNonQuery(connection);
+                    }
+
+                    transaction?.Commit();
+                }
+                catch (Exception)
+                {
+                    transaction?.Rollback();
+                    throw;
+                }
```
bricelam: This catch block isn't necessary. The `Dispose` in the `finally` will rollback the transaction.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandList.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Migrations
+{
+    public class MigrationCommandList
```
bricelam: This class doesn't add much value. The Execute methods could be be private helpers in the migrator or extension methods on `IEnumerable<MigrationCommand>`.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/HistoryRepository.cs
```diff
@@ -107,9 +107,9 @@ public virtual bool Exists()
         public virtual string GetCreateScript()
         {
             var operations = _modelDiffer.GetDifferences(null, _model.Value);
-            var commands = _migrationsSqlGenerator.Generate(operations, _model.Value);
+            var commandLists = _migrationsSqlGenerator.Generate(operations, _model.Value);
```
bricelam: This isn't a list of lists, is it?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
```diff
@@ -100,21 +99,20 @@ public virtual void Migrate(string targetMigration = null)
...
+
+            foreach (var commandList in commandLists)
             {
-                await ExecuteAsync(command(), cancellationToken);
+                await commandList().ExecuteNonQueryAsync(_connection, cancellationToken);
             }
         }
 
-        private IEnumerable<Func<IReadOnlyList<IRelationalCommand>>> GetMigrationCommands(
+        private IEnumerable<Func<MigrationCommandList>> GetMigrationCommandLists(
```
bricelam: Feels like this should either return `MigrationCommandList` or `IEnumerable<MigrationCommand>`, but not a list of lists.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -186,11 +198,27 @@ protected override void Generate(RenameSequenceOperation operation, IModel model
...
+            builder.Append("CREATE ");
+
+            if (operation.IsUnique)
+            {
+                builder.Append("UNIQUE ");
+            }
+
+            IndexTraits(operation, model, builder);
+
+            Index(operation, model, builder);
```
bricelam: We *really* want to avoid copy-and-paste overrides. Could you add a `bool terminateSatement` parameter to the base instead?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -117,13 +114,18 @@ public class MigrationsSqlGenerator : IMigrationsSqlGenerator
                 .Append("ALTER TABLE ")
                 .Append(SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
                 .Append(" ADD ");
+
             ColumnDefinition(operation, model, builder);
+
+            builder
+                .AppendLine(SqlGenerationHelper.StatementTerminator)
+                .EndCommand();
```
bricelam: Instead of calling `EndCommand` inline, can we put it in a `virtual` `EndStatement` method instead? This should make it easier for providers to implement batching if they want.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
bricelam: Oh I see, it's a way of forcing a new transaction. I wonder if there's a better way to model that in `MigrationCommand`.

test/Microsoft.EntityFrameworkCore.SqlServer.Tests/Migrations/SqlServerMigrationSqlGeneratorTest.cs
```diff
@@ -268,7 +268,7 @@ public virtual void CreateSchemaOperation_dbo()
             Generate(new EnsureSchemaOperation { Name = "dbo" });
 
             Assert.Equal(
-                ";" + EOL,
+                "",
```
bricelam: :+1:

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandList.cs
mikary: The dispose doesn't explicitly rollback the transaction, so it would depend on the ADO dispose. Is that good enough?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandList.cs
mikary: It seems like a reasonably encapsulated unit of functionality to me. The transaction logic is relevant to the `RelationalDatabaseCreator` and the `SqlServerDatabaseCreator`, so at very least we should have a common implementation. The `IEnumerable` version would be effectively the same pattern we had in `RelationalCommandExtensions`, but that still feels like an awkward / non-oo pattern.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
mikary: It seems like we have at least a few cases where we generate a list of `MigrationCommand`s and want to merge them either with or without splitting the transactions. This could be a feature of the `MigrationCommand` / `MigrationCommandList` if it seems like it would be valuable.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
mikary: We could, but I think it would have to be a parameter for all of the `Generate` methods. Doing it that way might be for the best, since it would preserve the extensibility of the previous implementation.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
mikary: That would work pretty well. Would we want to append the statement terminator in that method as well?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandList.cs
bricelam: The ADO.NET contract is to rollback during dispose if it isn't committed.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
bricelam: > I think it would have to be a parameter for all of the Generate

I think we should evolve the API based on need. For now, we should just be able to add two overloads:

```C#
protected virtual void Generate(
    CreateIndexOperation operation,
    IModel model,
    MigrationCommandListBuilder builder)
    => Generate(operation, model, builder, terminate: true);

protected virtual void Generate(
    CreateIndexOperation operation,
    IModel model,
    MigrationCommandListBuilder builder,
    bool terminate)
{
    // ...
}
```

mikary:
Added some updates from PR feedback

mikary:
rebase and ping @anpete

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandList.cs
anpete: Agree with @bricelam. I think we just need to change this to something like MigrationCommandExecutor and pass the commands to execute.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/IMigrationsSqlGenerator.cs
```diff
@@ -5,13 +5,12 @@
...
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Migrations.Operations;
-using Microsoft.EntityFrameworkCore.Storage;
 
 namespace Microsoft.EntityFrameworkCore.Migrations
 {
     public interface IMigrationsSqlGenerator
     {
-        IReadOnlyList<IRelationalCommand> Generate(
+        MigrationCommandList Generate(
```
anpete: I don't think we should add a new type here. IReadOnlyList<MigrationCommand> seems better.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommand.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Migrations
+{
+    public class MigrationCommand
+    {
+        public MigrationCommand(
+            [NotNull] IRelationalCommand relationalCommand,
+            bool transactionSuppressed)
```
anpete: default value?

anpete:
Is there an FT for this change?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandList.cs
anpete: Related to that, IMHO this would be a better way to do scripting than what we currently have. I.e. a MigrationsScriptingExecutor vs. the duplicated migration code path in Migration.GeneratedScript.

mikary:
Updated based on feedback and discussion

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommand.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Migrations
+{
+    public class MigrationCommand
```
anpete: Given, that this is not really a command - It has no Execute like method - a name like "MigrationOperation" seems better. Did you consider subclassing RelationalCommand here?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
```diff
@@ -244,13 +246,13 @@ public virtual void Migrate(string targetMigration = null)
                         builder.AppendLine(_historyRepository.GetBeginIfExistsScript(migration.GetId()));
                         using (builder.Indent())
                         {
-                            builder.AppendLines(command.CommandText);
+                            builder.AppendLines(command.RelationalCommand.CommandText);
```
anpete: Consider Law of Demeter here

anpete:
:shipit:

Issue #4474, opened 2016-02-03
Shadow Property - ChangeTracker Modified State


CharlesToniolo:
Hello,

I'm having a strange behavior with a shadow property and i don't know if it's a bug or not.
I was able to reproduce it in the dev branch.

Given these classes:
``` 
namespace Microsoft.EntityFrameworkCore.FunctionalTests.TestModels.Northwind
{
    public class ExtraValue
    {
        public string ID { get; set; }
        public string Description { get; set; }
    }
}
 
namespace Microsoft.EntityFrameworkCore.FunctionalTests.TestModels.Northwind
{

    public class Order
    {
        //public string ExtraValueID { get; set; } Shadow Property
        public ExtraValue ExtraValue { get; set; }

        public int OrderID { get; set; }
        public string CustomerID { get; set; }
        public int? EmployeeID { get; set; }
        ......
     }
}
``` 
I would like to have a Shadow Property called ExtraValueID to be the foreign key to the table ExtraValue.
Entity framework is automatically creating this Shadow Property but when i select the root entity with `AsNoTracking`  and `Update` the entity without changing anything like this:
``` 
var order = context.Orders.AsNoTracking()
   .Include(o => o.ExtraValue)
   .Where(o => o.OrderID.Equals(10248))
   .FirstOrDefault();

context.Update(order);
``` 

The Shadow Property `context.Entry(order).Property("ExtraValueID")` is being marked as modified with originalValue=null.
I think that this property should not be marked as modified. If i create the property ExtraValueID inside the class Order, it works.

Thanks

PR #4920, opened 2016-03-28
Introduce an Unknown state for property values

ajcvickers:
Shadow properties get this state when entities are first attached. These properties are then not considered modified even if the entire entity is marked as modified and so the values for these properties are not saved. The properties are only considered as modified or not when an actual value is set.

See #4474


src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -252,6 +261,17 @@ public virtual void MarkAsTemporary([NotNull] IProperty property, bool isTempora
             _stateData.FlagProperty(property.GetIndex(), PropertyFlag.TemporaryOrModified, isTemporary);
         }
 
+        protected virtual void MarkShadowPropertiesNotSet([NotNull] IEntityType entityType)
+        {
+            foreach (var property in entityType.GetProperties())
+            {
+                if (property.IsShadowProperty)
```
anpete: Consider LINQ or GetShadowProperties extension.

anpete:
:shipit:

Issue #4501, opened 2016-02-05
Reconsider ValueGeneratedOnAddOrUpdate + HasDefaultValueSql = Computed Column


rowanmiller:
Consider the following model where I want a `LastModified` column to be generated by the database. The best way to do this is a default value to handle inserts and a trigger for updates. I'd have to do the trigger outside of EF, but I can setup the default value in my model.

```c#
public class MyContext : DbContext
    {
        public DbSet<Blog> Blogs { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=Sample;Trusted_Connection=True;");
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Blog>()
            .Property(b => b.LastModified)
            .HasDefaultValueSql("GETDATE()")
            .ValueGeneratedOnAddOrUpdate();
        }
    }

    public class Blog
    {
        public int BlogId { get; set; }
        public DateTime LastModified { get; set; }
    }
```

The issue is that DDL assumes that ValueGeneratedOnAddOrUpdate with a default value means I want a computed column. Which I don't... since that would give the current time whenever I select the entity... but I have no way to stop if being a computed column.

```SQL
CREATE TABLE [Blog] (
    [BlogId] int NOT NULL IDENTITY,
    [Modified] AS GETDATE(),
    CONSTRAINT [PK_Blog] PRIMARY KEY ([BlogId])
);
```

ajcvickers:
This is probably because we collapse HasDefaultValueSql and HasComputedColumnSql into one piece of metadata: GeneratedValueSql. If we stop doing this, then Migrations can behave differently.

gdoron:
Exactly what I was trying to do now.
BTW, shouldn't be an EF way of defining a trigger for updates?
Something like:    

    HasValueSqlOnUpdate(string code)

With a lot of effort which probably doesn't worth it, it can even have an overload with `Expression<Func<TEntity, object>>` or something like that.
But the string overload should be relatively quite simple to implement.

What do you think @rowanmiller @ajcvickers?
I might even implement it myself and contribute the code if you think it's a good idea.

rowanmiller:
@gdoron yes I think that could be really useful - I doubt we'd get it into 1.0.0 release but I think it would provide a lot of value.

[...]

PR #4923, opened 2016-03-29
Split storage of default value SQL and computed value SQL

ajcvickers:
Because not all provider support both concepts we were storing these things as a single string and then letting other metadata determine how to use the value. However, for cases like that in issue #4501 Migrations assumes a computed column because it doesn't know that the value specified is actually a default and not computed expression. The fix is to stop combining the metadata.


AndriySvyryd:
:shipit:

natemcmaster:
cc @lajones do we need to update reveng?

lajones:
@natemcmaster - am I missing something? Looks like @ajcvickers already updated the reveng code as part of this.

natemcmaster:
@lajones I didn't investigate thoroughly, but does this change mean, in some cases, we should be scaffolding `HasComputedValueSql()` instead of `HasDefaultValueSql()`?

lajones:
@natemcmaster. I looked briefly and it looks like we should at least revisit it. At the moment we just get back an `is_computed` column and then use that to decide how to set `ValueGenerated`. I've raised issue #4944 to track this.

PR #4924, opened 2016-03-29
Reenable SQL Azure batching tests

### LINKED ISSUE #4228: Batching insert does not work on Linux with SQL Azure
Repro: https://github.com/natemcmaster/EntityFramework.PlatformTesting/blob/master/test/SqlAzure/BatchingTest.cs#L26

Inserting more than 1 entity fails on SQL Azure v12. I dug into what is happening and `DbDataReader.HasRows` is `false` when it should return true. For whatever reason, the `DbCommand.ExecuteReader()` does not return as expected.

AndriySvyryd:
Fixes #4228


AndriySvyryd:
@natemcmaster

natemcmaster:
:shipit:

PR #4927, opened 2016-03-29
Avoid performing empty updates

### LINKED ISSUE #4814: Timestamp/IsConcurrencyToken causes invalid UPDATE if no data modified (SQLServer)
I'm attaching a sample program causing UPDATE to fail if:
- Model contains a [Timestamp] property (rowversion)
- Model is saved but only the rowversion is set (and optionally other properties are set with the same values they already contained so no DB value has really to be updated).

The problem is that setting the rowversion fakes EF into thinking there is something to update, but since no property has really changed and rowversion is not saved (it is DB generated), an UPDATE with no fields to update is generated:

```
UPDATE [User] SET 
OUTPUT INSERTED.[RowVersion]
WHERE [UserId] = @p0 AND [RowVersion] = @p1;
',N'@p0 int,@p1 varbinary(8)',@p0=4,@p1=0x00000000000007D4
```

Notice the missing fields between SET and OUTPUT.

Sample code follows. Please check the last lines where the entity is updated.

```
namespace EFCoreTest
{
	class TestContext : DbContext
	{
		public DbSet<User> Users { get; set; }

		{
			optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=EFCoreTestDb2;Trusted_Connection=True;");
		}
	}

	class User
	{
		public int UserId { get; set; }

		[Timestamp]
		public byte[] RowVersion { get; set; }
	}

	class Program
	{
		{
			int id;
			byte[] version;

			using (var db = new TestContext())
			{
				db.Database.EnsureCreated();
				var user = new User { Name = "Name" };
				db.Users.Add(user);
				db.SaveChanges();
				id = user.UserId;
				version = user.RowVersion;
			}

			using (var db = new TestContext())
			{
				var user = db.Users.FirstOrDefault(u => u.UserId == id);

				// PROBLEM IS HERE
				user.Name = "Name"; // This has no effect since the value is not really changed...
				user.RowVersion = version; // This causes the problem

				db.SaveChanges();
			}
		}
	}
}
```

AndriySvyryd:
Fixes #4814

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Update/Internal/CommandBatchPreparer.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Update/Internal/CommandBatchPreparer.cs
@@ -76,7 +76,7 @@ namespace Microsoft.EntityFrameworkCore.Update.Internal
 
                         command.AddEntry(e);
                         return command;
-                    });
+                    }).Where(c => c.EntityState != EntityState.Modified || c.ColumnModifications.Any(m => m.IsWrite));
         }
 
         // To avoid violating store constraints the modification commands must be sorted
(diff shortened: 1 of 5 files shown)


ajcvickers:
:shipit:

PR #4928, opened 2016-03-29
Fix #4919 - Use transactions to ensure SQL Server database seeding does not

### LINKED ISSUE #4919: Query intermittently produces incorrect results on timeout
There appears to be an issue on the CI where several tests will fail due to incorrect results while others fail due to a `SqlException` "Timeout Expired." This appears to be a potential data corruption issue as the results are incorrect but no exception is thrown.

```
[Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.QueryNavigationsSqlServerTest.Select_Navigation] Assert.Equal() Failure
Expected: 830
Actual:   781
   at Microsoft.EntityFrameworkCore.FunctionalTests.QueryNavigationsTestBase`1.Select_Navigation() in C:\BuildAgent\work\16b2695b987e66d0\test\Microsoft.EntityFrameworkCore.FunctionalTests\QueryNavigationsTestBase.cs:line 317
   at Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.QueryNavigationsSqlServerTest.Select_Navigation() in C:\BuildAgent\work\16b2695b987e66d0\test\Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests\QueryNavigationsSqlServerTest.cs:line 52
```

Several more examples in in the CI logs.

mikary:
partially fail.
@divega @AndriySvyryd


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
```diff
@@ -485,10 +435,13 @@ public override void Dispose()
         }
 
         public static string CreateConnectionString(string name)
+            => CreateConnectionString(name, new Random().Next(0, 2) == 1);
+
+        private static string CreateConnectionString(string name, bool multipleActiveResultSets)
             => new SqlConnectionStringBuilder(TestEnvironment.DefaultConnection)
             {
                 //MultipleActiveResultSets = false,
```
AndriySvyryd: This comment can be removed

AndriySvyryd:
:shipit: Let's see if this helps

divega:
A bit OT but does anyone know what is the best way to get .sql files treated as text in GitHub so that you can get a good diff in Northwind.sql in a PR like this? I found http://stackoverflow.com/questions/28145687/why-does-git-think-my-sql-file-is-a-binary-file that suggests adding `.sql diff`` on the .gitattributes file. Another choice mentioned there is to make sure you save the file in an encoding that is supported by the tool.

mikary:
@divega We might consider doing both. It looks like adding `*.sql diff` to .gitattributes works, but the current encoding (in Northwind.sql) is difficult to read on a command line diff because it includes a lot of noise for null bytes and line endings.

divega:
@mikary sounds good. Not a super high priority but next time we make changes to .sql files I will try to ask that we do this before the PR is merged :smile:

Issue #4868, opened 2016-03-21
Boolean condition with non entity member in where() throw exception


cdie:
I don't know if this fix https://github.com/aspnet/EntityFramework/pull/4141 solves this problem, but I cannot make boolean test within a LINQ Where function with a non-entity member test.

In my example, I have a ViewModel that contains a simple flag.

In my controller method, I want to change the Where condition depending of this flag :

`var postes = _context.Set<GestPoste>().Where(m => m.Ps == guid && (viewmodel.IsTla == true ? m.Range >= 800 : m.Range < 800)).ToList();
`
This will give this SQL Query :

```
exec sp_executesql N'SELECT [m].[PERIPH_GUID], [m].[NAME], [m].[PS_GUID], [m].[RANGE]
FROM [GEST_POSTE] AS [m]
WHERE ([m].[PS_GUID] = @__guid_0) AND CASE
    WHEN @__model_IsTla_1
    THEN [m].[RANGE] >= 800 ELSE [m].[RANGE] < 800
END',N'@__guid_0 uniqueidentifier,@__model_IsTla_1 bit',@__guid_0='F6AC31E2-A2BB-403B-8D4D-264AD3A306B5',@__model_IsTla_1=1
```

Which, of course, gives an error.

My only workaround today is 

```
Func<GestPoste, bool> fun;
if (model.IsTla)
   fun = (m) => m.Range >= 800;
else
    fun = (m) => m.Range < 800;
var postes = _context.Set<GestPoste>().Where(m => m.Ps == guid && fun(m)).ToList();
```

Is this a bug ?

smitpatel:
The fix for #4141 fixes above partially. Specifically this is how the where clause would look like
```
WHERE ([m].[PS_GUID] = @__guid_0) AND CASE
    WHEN @__model_IsTla_1 = 1
    THEN [m].[RANGE] >= 800 ELSE [m].[RANGE] < 800
END
```
It would be still incorrect syntax. While fixing #4141 I hit the same issue. I couldn't find exact words in the documentation but from what I have tried, it is not possible that result of case block is a Boolean condition.
If you can translate the query in T-Sql then please provide it.

smitpatel:
Found T-Sql though it is very long expansion. It can be reduced using logical operators. Looking at the fixing it.
```
WHERE ([m].[PS_GUID] = @__guid_0) AND 
(CASE
    WHEN @__model_IsTla_1 = 1
    THEN 
		CASE
[...]

PR #4929, opened 2016-03-29
Merge conditions when ternary operator used in Where Clause to select…

smitpatel:
… filtering condition
fix for #4868 
Issue: While C# supports selecting condition based on another condition (in ternary operator) that is not supported by T-Sql since the Case Block needs to return single value.
Solution:
In the most simplest form of solution the conditions which are results (IfTrue, IfFalse) of another conditional, should be expanded to return true or false and then compare it to true or false as shown in https://github.com/aspnet/EntityFramework/issues/4868#issuecomment-202626010
Such logic can be simplified using Boolean logic rules. i.e.
`cond1 ? cond2 : cond3` translates to `(cond1 && cond2) || (!cond1 && cond3)` which generates a very concise Sql.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -3077,6 +3077,42 @@ WHERE 1 = 0",
                 Sql);
         }
 
+        public override void Where_ternary_boolean_condition()
+        {
+            base.Where_ternary_boolean_condition();
+
+            Assert.Contains(
+                    @"SELECT [p].[ProductID], [p].[Discontinued], [p].[ProductName], [p].[UnitsInStock]
+FROM [Products] AS [p]
+WHERE ((@__flag_0 = 1) AND ([p].[UnitsInStock] >= 20)) OR ((@__flag_0 <> 1) AND ([p].[UnitsInStock] < 20))",
+                    Sql);
+        }
+
+        public override void Where_ternary_boolean_condition_with_another_condition()
+        {
+            base.Where_ternary_boolean_condition_with_another_condition();
+
+            Assert.Equal(
+                    @"@__productId_0: 15
+@__flag_1: True
+
+SELECT [p].[ProductID], [p].[Discontinued], [p].[ProductName], [p].[UnitsInStock]
+FROM [Products] AS [p]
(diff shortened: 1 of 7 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -3067,6 +3067,42 @@ public override void Where_false()
...
+        }
+
+        public override void Where_ternary_boolean_condition_with_false_as_result()
+        {
+            base.Where_ternary_boolean_condition_with_false_as_result();
+
+            Assert.Contains(
+                    @"SELECT [p].[ProductID], [p].[Discontinued], [p].[ProductName], [p].[UnitsInStock]
+FROM [Products] AS [p]
+WHERE ((@__flag_0 = 1) AND ([p].[UnitsInStock] >= 20)) OR ((@__flag_0 <> 1) AND 0 = 1)",
```
smitpatel: @anpete , @maumar - Do we have something in current codebase which trims conditions like this? `0 = 1` is always false and there is actually no point of whatever is after OR.

anpete: @smitpatel Don't think so. Should be easy to add to the rather poorly named: CompositePredicateExpressionVisitor

smitpatel:
@anpete Added something another visitor for composite predicate which reduces logical operators
````
a && true => a
a && false => false
a || true => true
a || false => a
````

**Comment by anpete** [comment]
Nice! :shipit:

**Comment by smitpatel** [comment]
Working on git issue! branch lost commits somehow!

File: src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/PredicateLogicalReducingVisitor.cs [hunk 2]
```diff
@@ -0,0 +1,62 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq.Expressions;
+using Remotion.Linq.Parsing;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    public class PredicateLogicalReducingVisitor : RelinqExpressionVisitor
```
smitpatel: @anpete , @maumar - Any suggestion for better name?

anpete: PredicateReductionExpressionOptimizer? "Optimizing" is generally how we talk about this stuff and this name aligns with some existing ones.

divega:
:+1: this is a great improvement.

PR #4932, opened 2016-03-30
Make ForeignKeyAttribute configure the side it's on as the dependent side ...

### LINKED ISSUE #4885: Support using ForeignKeyAttribute on a collection navigation to influence the FK name
Almost same scenario as in #4884 but this time I hit a stack overflow while trying to workaround the wrong foreign key name by adding the `[ForeignKey("PostId")]` on the collection (that is why the repros are very similar).

``` C#
using System.Linq;
using Xunit;

namespace ForeignKeyName
{
    public class Tests
    {
        [Fact]
        private static void FK_name_matches_PK_name()
        {
            {
                var actualName = context.Model // triggers stack overflow 
                    .FindEntityType(typeof (Comment))
                    .GetForeignKeys().Single().Properties.Single().Name;
                Assert.Same("PostId", actualName); 
            }
        }
    }

    {

        {
        }
    }

    public class Post
    {
        [ForeignKey("PostId")] // Hoping to influence the name of the FK on the dependent side
        public ICollection<Comment> Comments { get; set; }
    }

    public class Comment
    {
        public int CommentId { get; set; }
    }
}
```

### LINKED ISSUE #4909: Using ForeignKeyAttribute to control the name of a FK property unexpectedly swaps principal and dependent sides and sets cardinality to 1:1
I introduced `[ForeignKey("UserId")]` in a model like the one below with the sole purpose of affecting the name of the foreign key for `Post.Author`.  

Instead EF Core went off to find that "UserId" matched the name of the principal key on `User`, and so decided to create a 1:1 identifying relationship between `Post.PostId` and `User.UserId`.

``` C#
    public class Post
    {
        [ForeignKey("UserId")]
        public User Author { get; set; }
    }

    public class User
    {
        public int UserId { get; set; }
    }

    {

        {
        }
    }
```
This is probably by design right now but I believe it is very confusing. In fact I find it strange that using `ForeignKeyAttribute` on a reference navigation property would turn a relationship around and make the reference point from principal to dependent. 

I only realized that something was wrong when the navigation property started to be populated with completely arbitrary entities.

Additional repro code:
``` C#
using System.Linq;
using Xunit;

namespace ForeignKeyName
{
    public class Tests
    {
        [Fact]
        private static void ForeignKeyAtribute_should_rename_fk_for_reference()
        {
            {
                var postForeignKeys = context.Model.FindEntityType(typeof (Post)).GetForeignKeys();
                Assert.True(
                    postForeignKeys 
                        .Any(fk=>fk.Properties
                            .Any(p=>p.Name=="UserId")));
            }
        }
        [Fact]
        private static void ForeignKeyAtribute_should_not_swap_principal_side()
        {
            {
                var userForeignKeys = context.Model.FindEntityType(typeof(User)).GetForeignKeys();
                Assert.False(
                    userForeignKeys
                        .Any(fk => fk.PrincipalKey.Properties
                            .Any(p => p.Name == "PostId")));
            }
        }
    }
```

AndriySvyryd:
… when specified on a reference navigation property.

Fixes #4909
Fixes #4885

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
@@ -78,15 +78,10 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
                 if ((fkPropertyOnDependent == null)
                     && (fkPropertyOnPrincipal == null))
                 {
-                    if (fkPropertiesOnNavigation.Any(p => foreignKey.DeclaringEntityType.FindProperty(p) == null)
-                        && fkPropertiesOnNavigation.All(p => foreignKey.PrincipalEntityType.FindProperty(p) != null))
+                    if (fkPropertiesOnPrincipalToDependent != null
+                        && foreignKey.IsUnique)
                     {
                         invertConfigurationSource = ConfigurationSource.DataAnnotation;
-                    } else if (relationshipBuilder.Metadata.GetPrincipalEndConfigurationSource() == null
-                               && fkPropertiesOnPrincipalToDependent == fkPropertiesOnNavigation
-                               && fkPropertiesOnNavigation.All(p => foreignKey.DeclaringEntityType.FindProperty(p) == null))
-                    {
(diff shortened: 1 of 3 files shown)


ajcvickers:
:shipit:

Issue #4817, opened 2016-03-16
Changes in derived related entities are not detected


AndriySvyryd:
The derived related entities are tracked as base related entities, so changes to properties that only exist on the derived type are not detected.

Add this test to `InheritanceRelationshipsQueryTestBase`:
```
[Fact]
public virtual void Changes_in_derived_related_entities_are_detected()
{
    using (var context = CreateContext())
    {
        context.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.TrackAll;

        var derivedEntity = context.BaseEntities.Include(e => e.BaseCollectionOnBase)
            .Single(e => e.Name == "Derived1(4)") as DerivedInheritanceRelationshipEntity;

        var firstRelatedEntity = derivedEntity.BaseCollectionOnBase.Cast<DerivedCollectionOnBase>().First();

        Assert.Equal(0, firstRelatedEntity.DerivedProperty);
                
        var firstRelatedEntityEntry = context.ChangeTracker.Entries<DerivedCollectionOnBase>()
            .Single(e => e.Entity == firstRelatedEntity);

        firstRelatedEntity.DerivedProperty = 1;
        context.ChangeTracker.DetectChanges();

        Assert.Equal(EntityState.Modified, firstRelatedEntityEntry.State);
        Assert.Equal(0, firstRelatedEntityEntry.Property(e => e.DerivedProperty).OriginalValue);
        Assert.Equal(1, firstRelatedEntityEntry.Property(e => e.DerivedProperty).CurrentValue);
    }
}
```

Also add `DerivedProperty` to `DerivedCollectionOnBase`

sandord:
I'm hitting this exact issue in EF Core 3.1 today.

What caught my eye is that after updating a property of the derived type with a different value, its value in `CurrentValues` equals the one in `OriginalValues`, which is obviously wrong.

ajcvickers:
@sandord This is a very old issue. If you're seeing this on EF Core 3.1 or 5.0 RC1, then please file a _new issue_ and attach a small, runnable project or post a small, runnable code listing that reproduces what you are seeing so that we can investigate.

sandord:
@ajcvickers Thanks for your reply. Luckily, I managed to figure out what was causing the issue.

I was doing this:

    var customer = await dbContext.Customers.GetRequired(customerId);

where `GetRequired()` is an extension method that had another `await` inside it, which obviously isn't very efficient anyway.

[...]

PR #4933, opened 2016-03-30
Ensure entities from query are tracked as correct type

ajcvickers:
Issue #4817

The problem was that query was passing in the base type to the state manager since the actual type varies by row. It would likely be possible to flow the actual type through from the materializer but this would require significant changes and would not be used in a lot of cases. Instead the state manager now looks up the entity type if it is not the base type, with some optimizations to that lookup made in metadata.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/ICanFindEntityType.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/ICanFindEntityType.cs
@@ -1,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Internal
+{
+    public interface ICanFindEntityType
+    {
+        IEntityType FindEntityType([NotNull] Type type);
+        IMutableEntityType AddEntityType([NotNull] string name, [CanBeNull] Type type);
+    }
+}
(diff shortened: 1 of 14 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Internal/ICanFindEntityType.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Internal
+{
+    public interface ICanFindEntityType
```
anpete: I thought we were just going to probe for our concrete metadata implementations from extensions.

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/Metadata/Internal/ICanFindEntityType.cs
ajcvickers: We were, but that information is not cached in a way that makes it fast to lookup--each type only caches its direct descendents. So this way worked out better in terms of a fast lookup.

anpete: Guess I must be missing something; thought we could just probe for Model.

ajcvickers: Not sure what you mean. We decided as a team to use this mechanism of accessing cached metadata without changing the underlying interfaces. If you mean cast to model and use it if the cache succeeds, I agree that would be easier and less work, but in the big discussion on this subject its not what we decided to do.

anpete: OK, I guess I must have missed/forgotten why we decided to add the extra interfaces.

PR #4945, opened 2016-03-31
Avoid adding redundant prefix for the shadow FK generated by convention

### LINKED ISSUE #4884: Unexpected foreign key name created by convention when PK name is entity type name + id
When the PK name is formed by combining the _type name_ + _Id_, I would expect shadow state FKs pointing to it on other entities to use the same name. Instead the FKs are of the form _type name_ + _type name_ + _id_.

I was hoping to find a test that could account for the current behavior around in [ForeignKeyPropertyDiscoveryConventionTest.cs](https://github.com/aspnet/EntityFramework/blob/dev/test/Microsoft.EntityFrameworkCore.Tests/Metadata/Conventions/Internal/ForeignKeyPropertyDiscoveryConventionTest.cs#L265), but I found those quite difficult to read :smile:

**Repro:**
``` C#
using System.Linq;
using Xunit;

namespace ForeignKeyName
{
    public class Tests
    {
        [Fact]
        static void FK_name_matches_PK_name()
        {
            {
                var actualName = context.Model.FindEntityType(typeof(Comment))
                    .GetForeignKeys().Single().Properties.Single().Name;
                Assert.Same("PostId", actualName); // actual is PostPostId
            }
        }
    }

    {

        {
        }
    }

    public class Post
    {
        public ICollection<Comment> Comments { get; set; }
    }

    public class Comment
    {
        public int CommentId { get; set; }
    }
}
```

AndriySvyryd:
Fixes #4884

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -1388,8 +1388,10 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
                 for (var i = 0; i < principalKey.Properties.Count; i++)
                 {
                     IProperty keyProperty = principalKey.Properties[i];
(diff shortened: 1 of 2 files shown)


divega:
:shipit:

Issue #4829, opened 2016-03-17
Metadata: Hard to write polymorphic code over metadata.


anpete:
We should consider creating interfaces for commonly used metadata elements such as:

``` c#
IEntityType DeclaringEntityType { get; } // Defined separately four times.
IReadOnlyList<IProperty> Properties { get; } // Defined separately three times.
```

It is often useful in tooling scenarios to be able to write polymorphic code over such elements.

rowanmiller:
We're not going to do anything on this now, we can revisit later

PR #4946, opened 2016-03-31
Move common metadata properties to interfaces for polymorphic extensibility

mikary:
Fix for #4829
cc: @divega @anpete

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/IMetadataElement.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/IMetadataElement.cs
@@ -1,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    public interface IMetadataElement
+    {
+        /// <summary>
+        ///     Gets the entity type this metadata element is declaired on.
+        /// </summary>
+        IEntityType DeclaringEntityType { get; }
+    }
+}
--- a/src/Microsoft.EntityFrameworkCore/Metadata/IMetadataNamed.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/IMetadataNamed.cs
@@ -1,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    public interface IMetadataNamed
+    {
+        /// <summary>
+        ///     Gets the name of the metadata element.
+        /// </summary>
+        string Name { get; }
+    }
+}
--- a/src/Microsoft.EntityFrameworkCore/Metadata/IMetadataProperties.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/IMetadataProperties.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    public interface IMetadataProperties
+    {
+        /// <summary>
+        /// Gets the properties that are associated with this metadata element. These properties may be declaired on a base type
+        /// in an inheritance hierarchy.
+        /// </summary>
+        IReadOnlyList<IProperty> Properties { get; }
+    }
+}
--- a/src/Microsoft.EntityFrameworkCore/Metadata/IMutableMetadataProperties.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/IMutableMetadataProperties.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    public interface IMutableMetadataProperties : IMetadataProperties
+    {
+        /// <summary>
+        /// Gets the properties that are associated with this metadata element. These properties may be declaired on a base type
+        /// in an inheritance hierarchy.
+        /// </summary>
+        new IReadOnlyList<IMutableProperty> Properties { get; }
+    }
+}
--- a/src/Microsoft.EntityFrameworkCore/Metadata/IEntityType.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/IEntityType.cs
@@ -11,17 +11,12 @@ namespace Microsoft.EntityFrameworkCore.Metadata
     /// <summary>
     ///     Represents an entity in an <see cref="IModel" />.
     /// </summary>
-    public interface IEntityType : IAnnotatable
+    public interface IEntityType : IAnnotatable, IMetadataNamed
     {
         /// <summary>
         ///     Gets the model this entity belongs to.
         /// </summary>
         IModel Model { get; }
-
-        /// <summary>
-        ///     Gets the name of the entity.
-        /// </summary>
-        string Name { get; }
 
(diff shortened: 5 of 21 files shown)


divega:
cc @ajcvickers @AndriySvyryd

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataElement.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    public interface IMetadataElement
```
AndriySvyryd: IEntityTypeMember?

ajcvickers:
@mikary  Can you quickly list each of the interfaces and the classes that implement each interface?

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataNamed.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    public interface IMetadataNamed
```
AndriySvyryd: INamed?

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataProperties.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    public interface IMetadataProperties
```
AndriySvyryd: IPropertyDefined?

src/Microsoft.EntityFrameworkCore/Metadata/IMutableMetadataProperties.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    public interface IMutableMetadataProperties : IMetadataProperties
+    {
+        /// <summary>
+        /// Gets the properties that are associated with this metadata element. These properties may be declaired on a base type
```
anpete: Typo

mikary:
Quick list of the interfaces and implementations

 - IMetadataElement (Has DeclaringEntityType)
  - IKey
  - IForeignKey
  - IIndex
  - IPropertyBase
    - IProperty (Existing relationship)
    - INavigation (Existing relationship)
  - IMutableMetadataElement (Has mutable DeclaringEntityType)
    - IMutableKey
    - IMutableForeignKey
    - IMutableIndex
    - IMutableProperty
    - IMutableNavigation

 - IMetadataNamed (Has Name)
  - IEntityType
  - IPropertyBase
    - IProperty (Existing relationship)
    - INavigation (Existing relationship)

- IMetadataProperties (Has Properties)
  - IKey
  - IForeignKey
  - IIndex
  - IMutableMetadataProperties (Has mutable Properties)
    - IMutableKey
    - IMutableForeignKey
    - IMutableIndex

ajcvickers:
Not sure we should do the named one. Does it really add value?

mikary:
I'd be fine with either having simple names for all of the interfaces or a consistent pattern (more what I was trying for) since the interface names aren't likely to be directly referenced by anyone who isn't working with metadata extensibility.

mikary:
@ajcvickers I could see it going either way.

Likewise we could theoretically use an interface for IIndex.IsUnique and IForeignKey.IsUnique, but those don't really map to the same concept so I left them out.

anpete:
@ajcvickers @mikary Not sure about the names one. I would lean towards keeping it as it could be useful in a tooling scenario.

ajcvickers:
I don't feel strongly, but for something that can be easily added later I would err on the side of not doing it just because "it could be useful".

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataProperties.cs
smitpatel: `IPropertiesDefined` ? since there is a list.

smitpatel:
100% source code change :wink:

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataElement.cs
divega: Either that or `IEntityTypeElement` (depending on what meaning you want to give to member, I feel a bit biased by CLR). It also seems that the word "Type" could be removed.

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataProperties.cs
divega: This is only for IKey, IForeignKey, IIndex which actually don't "define" but "reference" properties defined by IEntityType (in turn, IEntityType doesn't implement this interface, which is a bit strange but I guess is by-design).

AndriySvyryd: The interface name I proposed means that the object is defined by properties which is semantically different from declares. Referencing is ambiguous since IForeignKey also references properties on the principal side.

divega: > The interface name I proposed means that the object is defined by properties 

Got it now, although it isn't very clear IMO.

> Referencing is ambiguous since IForeignKey also references properties on the principal side.

But not directly in Properties, right? Aren't the properties from the principal side only in PrincipalKey?

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataElement.cs
anpete: +1 for IEntityTypeMember

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataProperties.cs
AndriySvyryd: > But not directly in Properties, right? Aren't the properties from the principal side only in PrincipalKey?

Correct

anpete: IStructuralType?

divega: That would be weird because IEntityType doesn't implement it. I guess I am biased by EF6 too :smile:

anpete: I guess EF6 had it wrong as both ETs and CTs are nominal.

divega:
I am getting the impression that this change is getting nowhere. There is nothing fundamentally wrong with adding interfaces like this but IMO the fact that figuring out names is consuming time like it is "smells" and makes me wonder if we have enough reasons to do the whole thing. Maybe it would help to see some examples from the usage that motivated the issue. @anpete could you help with that?

src/Microsoft.EntityFrameworkCore/Metadata/IMetadataProperties.cs
divega: I think they idea was that NominalType: StructuralType but the former wasn't defined in the metadata API. ETs and CTs were structural because they had a structure, not because they were not nominal :smile: RowType would be the only one that  purely structural. 

Regardless, I don't know what the right name would be.

divega:
We haven't agreed on going forward with this, but these are the latest names we came up with:

`IMetadataElement` -> `IEntityTypeMember` 
`IMetadataNamed` -> `INamed`
`IMetadataProperties` -> `IStructuralType`

smitpatel:
The issue this PR fixing is closed. Does this close now too?

divega:
We ended up deciding in triage not to pursue this for now. We can always revisit in the future when we have more scenarios that use this APIs.

PR #4947, opened 2016-03-31
Use chaining for AddXxx methods now it is supported again.

ajcvickers:
AddLogging and AddMemoryCache now support chaining again.


AndriySvyryd:
:shipit:

ErikEJ:
Providers beware?

ajcvickers:
@ErikEJ I don't think so, unless I'm missing something.

ErikEJ:
I was thinking about changes to: src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDesignTimeServices.cs  !?

ajcvickers:
@ErikEJ I don't think it has to return the IServiceCollection but to be consistent you could make the SQL CE one follow the same pattern.

ErikEJ:
Will do, thanks!

PR #4948, opened 2016-03-31
Refix fixes to build break so they won't break again

ajcvickers:
Use ifdefs to prevent R# from removing unused stuff on code cleanup


anpete:
:shipit:

Issue #4915, opened 2016-03-28
Investigate test regression in SQL Client on Linux


natemcmaster:
Possible regression in SqlClient? Started getting failures on Linux after upgrading version of SqlClient. Needs investigation.

```
Microsoft.EntityFrameworkCore.SqlServer.Tests.SqlServerConnectionTest.Creates_SQL_Server_connection_string [FAIL]
System.NotSupportedException : The keyword 'integrated security' is not supported on this platform.
Stack Trace:
     at System.Data.SqlClient.SqlConnectionString..ctor(String connectionString)
     at System.Data.SqlClient.SqlConnectionFactory.CreateConnectionOptions(String connectionString, DbConnectionOptions previous)
     at System.Data.ProviderBase.DbConnectionFactory.GetConnectionPoolGroup(DbConnectionPoolKey key, DbConnectionPoolGroupOptions poolOptions, DbConnectionOptions& userConnectionOptions)
     at System.Data.SqlClient.SqlConnection.ConnectionString_Set(DbConnectionPoolKey key)
     at System.Data.SqlClient.SqlConnection.set_ConnectionString(String value)
     at System.Data.SqlClient.SqlConnection..ctor(String connectionString)
     at Microsoft.EntityFrameworkCore.Storage.Internal.SqlServerConnection.CreateDbConnection()
     at Microsoft.EntityFrameworkCore.Internal.LazyRef`1.get_Value()
     at Microsoft.EntityFrameworkCore.Storage.RelationalConnection.get_DbConnection()
     at Microsoft.EntityFrameworkCore.SqlServer.Tests.SqlServerConnectionTest.Creates_SQL_Server_connection_string()
```

ErikEJ:
Is integrated security supported on Linux?

natemcmaster:
No, it isn't. Our test never sets `integrated security = true`, but the test fails anyways.

ErikEJ:
But the test set Trusted_Connection, which is an alias for integrated security: https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlconnection.connectionstring(v=vs.110).aspx

natemcmaster:
@ErikEJ thanks for investigating. Yup, that's the source of the issue. Looks like this test was changed by @ajcvickers in https://github.com/aspnet/EntityFramework/commit/6c1c83949e184db648d7b80a4cc36d17a6d3f91c

ErikEJ:
[...]

PR #4950, opened 2016-03-31
Remove Integrated Security/Trusted_Connection from dummy strings

ajcvickers:
Since it seems SQLClient on Linux is now verifying these

Issue #4915


divega:
Checking just in case, do we now need SQL auth credentials in the connection string?

ajcvickers:
Nope. At least, not on Windows. We don't ever open the connection.

divega:
:shipit:

Issue #4789, opened 2016-03-14
Reduce allocations from Property.Indexes and related APIs


rynowak:
These accessors https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore/Metadata/Internal/Property.cs#L423 end up allocating on every call due to the use of capturing lambdas inside the call to `LazyInitializer.EnsureInitialized`.

For a simple MVC app doing a single `INSERT` per request, accessing `Property.Indexes` generates about 2.5% of allocated bytes.
![image](https://cloud.githubusercontent.com/assets/1430011/13750218/c5f0954e-e9c1-11e5-8c09-e7b31795ef7a.png)

PR #4954, opened 2016-03-31
Use non-capturing lambdas for lazy initialization

ajcvickers:
See issue #4789

This requires a new initializer class since the BCL LazyInitializer doesn't support passing parameters to the creation delegate.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Internal/NonCapturingLazyInitializer.cs
+++ b/src/Microsoft.EntityFrameworkCore/Internal/NonCapturingLazyInitializer.cs
@@ -1,0 +1,41 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Internal
+{
+    public static class NonCapturingLazyInitializer
+    {
+        public static TValue EnsureInitialized<TParam, TValue>(
+            [CanBeNull] ref TValue target,
+            [CanBeNull] TParam param,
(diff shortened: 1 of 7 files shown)


src/Microsoft.EntityFrameworkCore/Internal/NonCapturingLazyInitializer.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public static TValue EnsureInitialized<TParam, TValue>(
+            [CanBeNull] ref TValue target,
+            [CanBeNull] TParam param,
+            [NotNull] Func<TParam, TValue> valueFactory) where TValue : class
+        {
+            if (target != null)
+            {
+                // ReSharper disable once UnusedVariable
+                var _ = _barrier; // Insert a volatile load barrier. Needed on IA64.
```
davidfowl: `Thread.MemoryBarrier()`?

ajcvickers: We decided to copy the code directly from LazyInitializer to avoid making those all-to-easy mistakes around what works on one architecture or another.

src/Microsoft.EntityFrameworkCore/Metadata/Internal/Navigation.cs
```diff
@@ -213,7 +212,7 @@ public virtual PropertyIndexes PropertyIndexes
                 }
                 else
                 {
-                    LazyInitializer.EnsureInitialized(ref _indexes, () => value);
+                    NonCapturingLazyInitializer.EnsureInitialized(ref _indexes, this, n => value);
```
anpete: Could you use the regular LazyInitializer here?

anpete:
:shipit:

PR #4961, opened 2016-04-01
Add more facet validation for shared columns

### LINKED ISSUE #4074: TPH: Validate shared columns and constraints

AndriySvyryd:
Fixes #4074

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
@@ -32,7 +32,7 @@ namespace Microsoft.EntityFrameworkCore.Internal
             base.Validate(model);
 
             EnsureDistinctTableNames(model);
-            EnsureDistinctColumnNames(model);
+            EnsureSharedColumnsCompatibility(model);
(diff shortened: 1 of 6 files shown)


lajones:
Looks good. :shipit:

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
```diff
@@ -89,12 +89,16 @@ protected virtual void EnsureDistinctColumnNames([NotNull] IModel model)
 
                 foreach (var property in properties)
```
smitpatel: should change name of the function.

PR #4962, opened 2016-04-01
Be consistent in exception types thrown

### LINKED ISSUE #611: Be consistent in the exception types thrown by core metadata
For very similar cases in core metadata we sometimes throw ArgumentException, sometimes throw InvalidOperationException, and sometimes throw ModelItemNotFoundException. We should decide what the general categories of errors are here and then be consistent in what we throw.

### LINKED ISSUE #4730: Review usages of NotSupportedException

### LINKED ISSUE #4850: EntityTypeNameEqualityComparer should be in Core
At the moment the class `EntityTypeNameEqualityComparer ` is in `InMemory`. This is a generic useful class and should be in `Core`.

AndriySvyryd:
Fixes #611
Fixes #4730
Fixes #4850


anpete:
:shipit:

PR #4963, opened 2016-04-01
Fix #4875 - EF.Property<T> throws in certain generic usages

### LINKED ISSUE #4875: Exception "The EF.Property<T> method may only be used within LINQ queries" and unnecessary client-side evaluation triggered in certain generic usages
In certain usages of the `EF.Property()` method the compiler introduces a cast to `object` for the `entity` argument of the call. This apparently prevents EF Core from correctly parsing the LINQ expression.

I have seen this lead to two different outcomes:
1. An InvalidOperationException with message "The EF.Property<T> method may only be used within LINQ queries."
2. Unnecessary client-side evaluation

I was able to produce a relatively small repro for the exception case (see below) but unfortunately the second effect is more elusive and so far only appears in a larger application (I can work with someone figuring out that part).

``` C#
using System;
using System.Data.SqlClient;
using System.Linq;
using Microsoft.EntityFrameworkCore.ChangeTracking;
using Microsoft.EntityFrameworkCore.Metadata.Builders;

namespace EFPropertyFromObject
{
    internal class Program
    {
        private static void Main(string[] args)
        {
            using (var context = new MyContex("deleteme", 4))
            {
                context.Add(new Customer {Name = "Alfred", Country = "USA"});
                context.Add(new Customer {Name = "Maria", Country = "USA"});

                context.SaveChanges();
            }
            using (var context = new MyContex("deleteme", 4))
            {
                var allCountries = context.QueryAllCountries();
                foreach (var country in allCountries) // exception is thrown here
                {
                    Console.WriteLine(country);
                }
            }
        }
    }

    public class Customer
    {
        public int Id { get; set; }
        public string Country { get; set; }
    }

    public class MyContex : DbContext
    {
        private readonly string _connectionString;
        private readonly int _tenantId;

        public MyContex(string database, int tenantId)
        {
            _connectionString =
                new SqlConnectionStringBuilder
                {
                    DataSource = @"(localdb)\mssqllocaldb",
                    InitialCatalog = database,
                    IntegratedSecurity = true
                }.ToString();
            _tenantId = tenantId;
        }

        {
            modelBuilder.Entity<Customer>(AddTenantId);
        }

        {
            optionsBuilder.UseSqlServer(_connectionString);
        }

        public override EntityEntry<TEntity> Add<TEntity>(TEntity entity)
        {
            var entry = base.Add(entity);
            entry.Property<int>("TenantId").CurrentValue = _tenantId;
            return entry;
        }

        public IQueryable<string> QueryAllCountries()
            => QueryCustomers().Select(c => c.Country).Distinct();

        public IQueryable<Customer> QueryCustomers()
            => ApplyTenantFilter(Set<Customer>());

        private IQueryable<T> ApplyTenantFilter<T>(IQueryable<T> source) //where T : class
            => source.Where(e => EF.Property<int>(e, "TenantId") == _tenantId);

        private void AddTenantId<T>(EntityTypeBuilder<T> builder) where T : class
        {
            builder.Property<int>("TenantId");
        }
    }
}
```

Note that the cast to `object` is added by the compiler to account for value types which need to be boxed in order to be passed as objects, so adding the generic constraint `where T: class` to `ApplyTenantFilter()` prevents the compiler from introducing the cast. However that doesn't seem to be a very discoverable workaround.

mikary:
Linq adds an extra Convert to object for boxing when EF.Property<T> is called in a method without generic constraints on the entity type.

Adding a step to query compilation to update the Expression before the QueryModel is generated but after caching.

cc @anpete @maumar

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
@@ -1186,9 +1186,18 @@ namespace Microsoft.EntityFrameworkCore.Query
                     || querySource == null
                     || querySource == querySourceReferenceExpression.ReferencedQuerySource)
                 {
+                    var entityExpression = methodCallExpression.Arguments[0];
+
+                    if ((entityExpression.NodeType == ExpressionType.Convert
+                        || entityExpression.NodeType == ExpressionType.ConvertChecked)
+                            && ((UnaryExpression)entityExpression).Type == typeof(object))
+                    {
+                        entityExpression = ((UnaryExpression)entityExpression).Operand;
+                    }
+
                     var entityType
                         = QueryCompilationContext.Model
-                            .FindEntityType(methodCallExpression.Arguments[0].Type);
+                            .FindEntityType(entityExpression.Type);
 
                     if (entityType != null)
                     {
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/MemberAccessBindingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/MemberAccessBindingExpressionVisitor.cs
@@ -154,7 +154,13 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
             if (EntityQueryModelVisitor.IsPropertyMethod(node.Method))
             {
                 var newArguments
-                    = VisitAndConvert(node.Arguments, "VisitMethodCall");
+                    = VisitAndConvert(
+                        new List<Expression>
+                        {
+                            node.Arguments[0].RemoveConvert(),
+                            node.Arguments[1]
+                        }.AsReadOnly(),
+                        "VisitMethodCall");
 
                 if (newArguments[0].Type == typeof(ValueBuffer))
                 {
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
@@ -106,7 +106,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
                     if (parameterInfos[i].GetCustomAttribute<NotParameterizedAttribute>() != null)
                     {
-                        var parameter = newArgument as ParameterExpression;
+                        var parameter = newArgument.RemoveConvert() as ParameterExpression;
 
                         if (parameter != null)
                         {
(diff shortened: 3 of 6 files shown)


anpete:
Can we just do RemoveConvert "at the time" I.e. In BindMethodCallExpression? That would be much simpler and more inline with how we already approach this.

mikary:
I tried doing it in BindMethodCallExpression, but that seems to be too late. It gets through the bind call but just blows up in another spot. This cast is fairly specific to the first parameter of EF.Property<T>. If I stripped it from all parameters, I was getting another exception for uses like the following (passing in a string literal as object and casting back to string in the method):

```
        [ConditionalFact]
        public virtual void Select_Property_when_shaow_unconstrained_generic_method()
        {
            AssertQuery<Employee>(es =>
                ShadowPropertySelect<Employee, string>(es, "Title"));
        }

        protected IQueryable<string> ShadowPropertySelect<T>(IQueryable<T> source, object column)
            => source.Select(e => EF.Property<string>(e, (string)column));
```

anpete:
@mikary "blows up in another spot" - with what?

mikary:
When I remove the convert in `EntityQueryModelVisitor.BindMethodCallExpression`, and exception is still thrown in `MemberAccessBindingExpressionVisitor.VisitMethodCall`. It looks like a similar issue (that visitor is also expecting the arguments to be unwrapped).

```
Rewriting child expression from type 'Microsoft.EntityFrameworkCore.FunctionalTests.TestModels.Northwind.Employee' to type 'Microsoft.EntityFrameworkCore.Storage.ValueBuffer' is not allowed, because it would change the meaning of the operation. If this is intentional, override 'VisitUnary' and change it to allow this rewrite.
System.InvalidOperationException was unhandled by user code
  HResult=-2146233079
  Message=Rewriting child expression from type 'Microsoft.EntityFrameworkCore.FunctionalTests.TestModels.Northwind.Employee' to type 'Microsoft.EntityFrameworkCore.Storage.ValueBuffer' is not allowed, because it would change the meaning of the operation. If this is intentional, override 'VisitUnary' and change it to allow this rewrite.
```

mikary:
@anpete @maumar suggested adding a `VisitUnary` to the `MemberAccessBindingExpressionVisitor` to strip out the cast (similar to the same method in `NavigationRewritingExpressionVisitor`). Doing so appears to work up until `QueryBuffer.GetPropertyValue` fails on a `Debug.Assert` to ensure the value is found)

anpete:
@mikary That assert means that it is attempting to get the value of shadow property for an entity that was not found in the query buffer. You likely need to debug the query plan to get to the bottom of that.

mikary:
@anpete @maumar and I did a bit more digging, it looks like it works if we explicitly remove convert in `MemberAccessBindingExpressionVisitor.VisitMethodCall` as in the following:

```cs
                var arguments = new List<Expression>
                {
                    node.Arguments[0].RemoveConvert(),
                    node.Arguments[1]
                }.AsReadOnly();

                var newArguments
                    = VisitAndConvert(arguments, "VisitMethodCall");
```

If we would rather deal with it where it breaks rather than as a preprocess step, it could be done this way.

anpete:
@mikary Yep, that's how we deal with this.

mikary:
Updated based on PR feedback

anpete:
:shipit:

mikary:
@anpete fixed an inheritance issue that was about to be introduced by the other changes in this PR.

anpete:
:shipit:

PR #4970, opened 2016-04-04
Small: fix missed cancellation Token when migrate async

laskoviymishka:
This is small fix.
I noticed that in history repository missed cancellation token for async command reader.


bricelam:
Already fixed by @ajcvickers in 79cb49bcac64452a09be960fbc8dbab623381ca7

PR #4974, opened 2016-04-04
Fix #4780 - Ensure that we always apply a composed projection when us…

### LINKED ISSUE #4780: EF mistreats LINQ methods chained after FromSql
I'm getting the following **warning** using **RC-1-Final**:

> The LINQ expression 'orderby ([x].ReplyToPostId ?? -1) asc, [x].IsImportantReply desc, [x].PostId asc' could not be translated and will be evaluated locally.

Code:

        public Task<List<HierarchyPost>> GetRootAsync(int postId)
        {
            return HierarchyPosts.FromSql(@"
	with postsCTE as (
	   select Id, ReplyToPostId
	   from Posts
	   where Id = @p0
	   union all
	   select Parent.Id, Parent.ReplyToPostId
	   from Posts Parent
	     join postsCTE Son on Son.ReplyToPostId = Parent.Id  -- this is the recursion
	) 
	select * from HierarchyPosts where RootId =(
	                                            select top 1 Id
	                                            from PostsCTE
	                                            where ReplyToPostId is null)", postId)
	                .OrderBy(x => x.ReplyToPostId ?? -1)
	                .ThenByDescending(x => x.IsImportantReply)
	                .ThenBy(x => x.PostId)
	                .ToListAsync();
        }
        
It's quite easy to translate the query to SQL (using `case`) and indeed it worked in earlier versions of EF.

anpete:
…ing FromSql.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
@@ -797,7 +797,8 @@ namespace Microsoft.EntityFrameworkCore.Query
 
             var sequenceType = _expression.Type.GetSequenceType();
 
-            if (selectClause.Selector.Type == sequenceType)
+            if (selectClause.Selector.Type == sequenceType
+                && selectClause.Selector is QuerySourceReferenceExpression)
             {
                 return;
             }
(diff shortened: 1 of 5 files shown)


maumar:
:shipit:

Issue #2532, opened 2015-07-03
Migrations: Store down script in history table


bricelam:
This is @rowanmiller's awesome idea for being able to revert migrations without the corresponding `*.cs` files.

Bartmax:
:+1:

PR #4985, opened 2016-04-05
Store down script

laskoviymishka:
I faced with really long manual revert migration since one of my teammates applied migration to shared DB and then suddenly changed his mind and revert all of code  :anger: . 

I think this could save couple days of pain for someone like me


divega:
@rowanmiller Thoughts on this?

laskoviymishka:
Update PR according to latest updates from dev branch

laskoviymishka:
UPD: merge latest source to PR :monkey: 
@rowanmiller any thoughts on this?

rowanmiller:
Sorry for the slow response on this one. We discussed this further and looking thru the changes this is a partially implemented feature that we don't want to take into the 1.0.0 code base. It stores the down script, but as yet there is no way to make any use of it. The feature needs some overall design work to iron out when we store the script (always, opt-in, etc.) and then how the script is used (APIs, command line, etc.).

Closing for now, but very happy to discuss a proposal for the overall design and then take a PR.

laskoviymishka:
@rowanmiller where should disscussion take place? Should i file new issue or use this one #4985?

rowanmiller:
I think you can just use #2532. Our team is pretty flat out getting RTM ready at the moment, so we may be a little slow to respond for a little while.

laskoviymishka:
@rowanmiller no problem, i'm not claiming myself as flash. So my proposal could be also little bit slow :smile_cat:

PR #4995, opened 2016-04-06
Reduce allocations during SaveChanges

### LINKED ISSUE #4535: AddRange is not performing if it is more that 50K
I have used AddRange this has a good performance when compared with EF6 AddorUpdate. At the same time when I add more records its slow down and not working: below is the stat. I have also created batch and tried inserting there was a increases in performance. but more then 50K its not working. In my case I am looking for at least to insert 50K+ one time insert. since the performance is not good I am going back to BulkInsert in EF6
 Env Records Insert(sec) Update(sec) Delete(sec)
 LocalDB 5K 7 6 1
 LocalDB 10K 22 24 9
 LocalDB 20K 48 100 48
 LocalDB 30K 60 120 60
 LocalDB 50K 180 300 120
 LocalDB 0.5K 2 2 0.4
 Azure 5K 120 240 54
 Azure 10K 300 660 120

ef7_addrange

AndriySvyryd:
Fixes #4535 

This change reduces the total number of allocations for this scenario in half


src/Microsoft.EntityFrameworkCore/Internal/NonCapturingLazyInitializer.cs
```diff
@@ -9,6 +9,20 @@ namespace Microsoft.EntityFrameworkCore.Internal
 {
     public static class NonCapturingLazyInitializer
     {
+        public static TValue EnsureInitialized<TValue>(
```
anpete: How is this different from LazyInitializer?

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/Internal/NonCapturingLazyInitializer.cs
AndriySvyryd: Will remove

rynowak:
:+1:

PR #4999, opened 2016-04-07
Ensure relationship snapshot is kept up-to-date when query loads navigations

ajcvickers:
Issue #4853

The issue here is when the second TestClass was loaded the relationship snapshot on the inverse was not updated, which in turn caused DetectChanges to misbehave. Rather than attempt to do any kind of diff at the end of query the fix is to update the relationship snapshot at the same time that query is setting or populating the actual navigation property.


AndriySvyryd:
:shipit:

Issue #4953, opened 2016-03-31
SQL Client: Unable to load DLL 'sni.dll'


mrmeek:
Downloaded latest CLI build (beta-2175), created new project with dotnet new, added aspnetcidev myget feed to nuget.config, added AspNetCore, MVCCore, and EFCore dependencies, created context and model for existing database, SQL Server query fails with:

> dbug: Microsoft.EntityFrameworkCore.Query.Internal.SqlServerQueryCompilationContextFactory[5]
      TRACKED: True
      (QueryContext queryContext) => IEnumerable<Product> _ShapedQuery(
          queryContext: queryContext,
          shaperCommandContext: SelectExpression:
              SELECT [p].[Id], [p].[Name]
              FROM [Products] AS [p]
          ,
          shaper: UnbufferedEntityShaper<Product>
      )
>
dbug: Microsoft.EntityFrameworkCore.Storage.Internal.SqlServerConnection[3]
      Opening connection to database 'ProductsDb' on server 'localhost\SQLEXPRESS'.
fail: Microsoft.EntityFrameworkCore.Query.Internal.SqlServerQueryCompilationContextFactory[1]
      An exception occurred in the database while iterating the results of a query.
      System.DllNotFoundException: Unable to load DLL 'sni.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
         at System.Data.SqlClient.SNINativeMethodWrapper.UnmanagedIsTokenRestricted(IntPtr token, Boolean& isRestricted)

Easiest way to see this in action is to download the cli-samples (https://github.com/aspnet/cli-samples), using HelloMvcApi add "Microsoft.EntityFrameworkCore.SqlServer": "1.0.0-rc2-*" to project.json dependencies, query anything with a context.

Apologies if this is a dupe of known issue in #3925, thought it was worth mentioning because it's occurring again, but, this time on any EF+SqlServer server connections- no tools/ef-commands here.

Also, I understand the churn in these repos. Let me know if there's no need to report such mundane/obvious issues right after changes in the core/cli nightlies.

mrmeek:
After looking further I see this is not an issue with EF. Targeting System.Data.SqlClient directly and opening a connection throws the same exception ala dotnet/corefx#5252 and the issue probably belongs there

rowanmiller:
Keeping this open to track the issue from the EF side, but opened https://github.com/dotnet/corefx/issues/7518 for SQL Client to fix the issue.

[...]

PR #5002, opened 2016-04-07
Load runtime and native libraries in dotnet-ef

natemcmaster:
Customizes the creation of `ProjectLoadContext` to include native and runtime libraries.

Partial fix for https://github.com/aspnet/EntityFramework/issues/3925

Work-around https://github.com/dotnet/cli/issues/2289

~~wait for fixes in CLI regarding #4953~~

Added some end-to-end testing for dotnet-ef using sqlite.


natemcmaster:
cc @anurse

natemcmaster:
:watch: pending changes to DependencyContext, we may be able to remove the hard-coded RID graph here.

anurse:
FYI: https://github.com/dotnet/cli/pull/2379

natemcmaster:
Just added a fix for loading sni.dll at design time. With this change, dotnet-ef now completely reimplements everything from  ["Microsoft.DotNet.ProjectModel.Loader"](https://github.com/dotnet/cli/tree/rel/1.0.0/src/Microsoft.DotNet.ProjectModel.Loader)... but with bug fixes.

natemcmaster:
ping @anurse for code review. I've updated to use DependencyContext now that we've taken the fix from https://github.com/dotnet/cli/pull/2379. This resolves the remaining issues on #3925 but still implements its own assemblyload context.

src/dotnet-ef/Loader/DesignTimeProjectLoadContext.cs
```diff
@@ -0,0 +1,134 @@
...
+        private bool SearchAlternateNativeLibraryName(string dllFileName, out string path)
+        {
+            if (_nativeLibraries.Count == 0)
+            {
+                path = null;
+                return false;
+            }
+
+            var alternates = new List<string>();
+            if (dllFileName.EndsWith(NativeLibrarySuffix))
```
anurse: If the DllImport ends with `.dll`, won't this fail to strip that extension off on non-Windows platforms because of the static constructor above? Might I suggest `Path.GetFileNameWithoutExtension`?

anurse:
Looks good to me. Btw, post-RC2 we'll probably just take your load context into the CLI, it looks much richer that our current one :)

src/dotnet-ef/Loader/DesignTimeProjectLoadContext.cs
natemcmaster: Good point. I'll use path instead.

natemcmaster:
> we'll probably just take your load context into the CLI

Sounds good. For now this workaround seems fine to have in dotnet-ef. When you move changes into cli, you ping me on the change?

natemcmaster:
> Looks good to me.

@anurse so is that a :shipit:?

anurse:
Yeah, I didn't know for sure if you wanted anyone from EF to sign off, but I'm :shipit:

divega:
:shipit:

PR #5003, opened 2016-04-07
Fix #4975 - EF.Property<T> property name shoud not be parameterized

### LINKED ISSUE #4975: Query caches property name on EF.Property<T> closure
When a query on a shadow property takes the property name as a parameter, the parameter name is cached on the first call and not respected in successive calls.

The following test fails on the second Assert:
```cs
        [ConditionalFact]
        public virtual void Where_shadow_closure()
        {
            var propertyName = "Title";
            var value = "Sales Representative";

            AssertQuery<Employee>(
                es => es.Where(e => EF.Property<string>(e, propertyName) == value),
                entryCount: 6);

            propertyName = "FirstName";
            value = "Steven";

            AssertQuery<Employee>(
                es => es.Where(e => EF.Property<string>(e, propertyName) == value),
                entryCount: 1);
        }
```

Generated SQL (Second query still referencing `Title` and not `FirstName`):
```sql
@__value_0: Sales Representative

SELECT [e].[EmployeeID], [e].[City], [e].[Country], [e].[FirstName], [e].[ReportsTo], [e].[Title]
FROM [Employees] AS [e]
WHERE [e].[Title] = @__value_0

@__value_0: Steven

SELECT [e].[EmployeeID], [e].[City], [e].[Country], [e].[FirstName], [e].[ReportsTo], [e].[Title]
FROM [Employees] AS [e]
WHERE [e].[Title] = @__value_0
```

mikary:
Implementing the solution proposed in #4975. Also includes minor change to `ParameterExtractingExpressionVisitor` because the attribute was not being checked due to a performance optimization.
@anpete @maumar

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.FunctionalTests/QueryTestBase.cs
+++ b/test/Microsoft.EntityFrameworkCore.FunctionalTests/QueryTestBase.cs
@@ -4689,6 +4689,24 @@ namespace Microsoft.EntityFrameworkCore.FunctionalTests
         }
 
         [ConditionalFact]
+        public virtual void Where_Property_shadow_closure()
+        {
+            var propertyName = "Title";
+            var value = "Sales Representative";
+
+            AssertQuery<Employee>(
+                es => es.Where(e => EF.Property<string>(e, propertyName) == value),
+                entryCount: 6);
+
+            propertyName = "FirstName";
+            value = "Steven";
+
+            AssertQuery<Employee>(
+                es => es.Where(e => EF.Property<string>(e, propertyName) == value),
+                entryCount: 1);
+        }
+
+        [ConditionalFact]
         public virtual void Selected_column_can_coalesce()
         {
             using (var context = CreateContext())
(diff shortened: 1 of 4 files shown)


test/Microsoft.EntityFrameworkCore.FunctionalTests/QueryTestBase.cs
```diff
@@ -4689,6 +4689,24 @@ public virtual void Where_Property_when_shadow()
         }
 
         [ConditionalFact]
+        public virtual void Where_Property_shadow_closure()
```
maumar: we could add tests where we reference static field and/or property

maumar:
:shipit:

test/Microsoft.EntityFrameworkCore.FunctionalTests/QueryTestBase.cs
mikary: Looks like we already have a few tests to cover these kinds of cases (i.e. `Where_static_field_access_closure_via_query_cache`)

PR #5013, opened 2016-04-08
Fix #4998 - NotImplementedException thrown on Where TypeIs cache key generation

### LINKED ISSUE #4998: Query ExpressionEqualityComparer throws on Is
`ExpressionEqualityComparer.GetHashCode` throws a `NotImplementedException.` for expressions with expression type `ExpressionType.TypeIs` (`TypeBinaryExpression`)

```cs
        [Fact]
        public virtual void NotImplementedException_on_is()
        {
            {
                var birds = context.Set<Animal>().Where(a => a is Bird).ToList();
            }
        }
```

mikary:
Adds hash code generation for expression type that was not covered by the existing implementation.
@anpete @divega


anpete:
:shipit:

PR #5016, opened 2016-04-08
Make Sql Azure batching test only run on not Sql Azure

AndriySvyryd:
This will let us know if the scenario works while #4228 is being fixed
@natemcmaster

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlAzure/SqlAzureBatchingTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlAzure/SqlAzureBatchingTest.cs
@@ -10,7 +10,7 @@ using Xunit.Abstractions;
 
 namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.SqlAzure
 {
-    [SqlServerCondition(SqlServerCondition.IsSqlAzure)]
+    [SqlServerCondition(SqlServerCondition.IsNotSqlAzure)]
     public class SqlAzureBatchingTest : IClassFixture<BatchingSqlAzureFixture>
     {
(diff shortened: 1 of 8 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlAzure/SqlAzureBatchingTest.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.SqlAzure
 {
-    [SqlServerCondition(SqlServerCondition.IsSqlAzure)]
+    [SqlServerCondition(SqlServerCondition.IsNotSqlAzure)]
```
natemcmaster: Are you just trying to run this test class on our normal runs of mssql functional tests? If so, it will be simple to just remove the IsSqlAzure condition.

AndriySvyryd: I know, but I think it would be useful to have this condition as well and get a clean pass on SQL Azure for Linux

natemcmaster:
This makes it clear to me that we need a broader cleanup of sql azure testing. Right now, mssql functional tests fall into IsNotSqlAzure (most tests) or IsSqlAzure. We don't have tests to work for both conditions.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlAzure/SqlAzureBatchingTest.cs
natemcmaster: Good point. Keep it.

natemcmaster:
:shipit:

Issue #4994, opened 2016-04-06
Migrations of multiple DbContexts with different connection strings


newsash:
``` c#
services.AddEntityFramework()
    .AddSqlServer()
    .AddDbContext<FirstDbContext>(options =>
        options.UseSqlServer(FIRST_CONNECTION_STRING))
    .AddDbContext<SecondDbContext>(options =>
        options.UseSqlServer(SECOND_CONNECTION_STRING));
```

`dnx ef database update -c FirstDbContext -v` writes changes to a database from the second connection string

``` c#
services.AddEntityFramework()
    .AddSqlServer()
    .AddDbContext<FirstDbContext>(options =>
        options.UseSqlServer(FIRST_CONNECTION_STRING));
    //.AddDbContext<SecondDbContext>(options =>
    //    options.UseSqlServer(SECOND_CONNECTION_STRING));
```

`dnx ef database update -c FirstDbContext -v` writes changes to a database from the first connection string as expected

These observations from "dnx-1.0.0-rc1-update1"

aschlesener:
We are experiencing the same issue (although not related to migrations), where adding a second dbcontext with a different connection string overwrites the first connection string - so both the first and the second dbcontexts use the second connection string. 

Is there any workaround for this issue until the bug is fixed?

ajcvickers:
@aschlesener @newsash  Couple of questions:
* Are you using current bits (RC2) or latest released bits (RC1)?
* Can you post what the constructors for your classes derived from DbContext looks like? In particular, I am interested if they use generic or non-generic DbContextOptions objects?

ajcvickers:
Following on from this, the fix is likely to be to use generic `DbContext<TContext>` parameters in the DbContext classes.

mikes-gh:
My startup is
```
services
                .AddDbContext<ApplicationDbContext>(options =>
                    options.UseSqlServer(Configuration["Data:default:ConnectionString"]))
                .AddDbContext<TradingCompanyDbContext>(options =>
                    options.UseSqlServer(Configuration["Data:default:ConnectionString"]));
```
I had to remove the second call to AddDbContext and use AddScoped and OnConfiguring instead  to get it to work.

A short sample of how we are meant to do it now would be a great help

mikes-gh:
ping @ajcvickers .  Sorry to ask again ( know you guys busy with rc2) A few lines of code on how multiple AddDbContext should be done would be much appreciated by me and would be helpful for others.

[...]

PR #5017, opened 2016-04-08
Throw when using multiple context types with non-generic DbContextOptions

ajcvickers:
Probably root cause of issue #4994.

If there are multiple context types registered in D.I. and each depends on non-generic DbContextOptions, then only one of them will get the correct instance. This fix checks that a context is getting its correct options and throws if not with the guidance to use generic DbContextOptions parameters in the context constructors.


divega:
:shipit:

Issue #4944, opened 2016-03-31
[RevEng] Revisit whether we need to generate HasComputedValueSql()


lajones:
See #4501 and #4923.

Given that, after #4923, the runtime dinstinguishes between the metadata for computed and default value columns, we should revisit the code in the `SqlServerDatabaseModelFactory` and ensure that we distinguish between default and computed columns as laid out in #4501 and ultimately generate the correct fluent API - perhaps in some cases generating `HasComputedValueSql()` instead of ` HasDefaultValueSql() `.

lajones:
Fix checked in with 2acd2f1.

PR #5018, opened 2016-04-08
Reverse Engineer HasComputedColumnSql() as needed.

lajones:
Fix for #4944. For computed columns used `HasComputedColumnSql()` instead of `HasDefaultValueSql()`.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Design/Metadata/ColumnModel.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Design/Metadata/ColumnModel.cs
@@ -21,6 +21,8 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Metadata
         [CanBeNull]
         public virtual string DefaultValue { get; [param: CanBeNull] set; }
 
+        [CanBeNull]
+        public virtual string ComputedValue { get; [param: CanBeNull] set; }
         public virtual int? MaxLength { get; [param: CanBeNull] set; }
         public virtual int? Precision { get; [param: CanBeNull] set; }
         public virtual int? Scale { get; [param: CanBeNull] set; }
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.Design.FunctionalTests/ReverseEngineering/Expected/AllFluentApi/SqlServerReverseEngineerTestE2EContext.expected
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.Design.FunctionalTests/ReverseEngineering/Expected/AllFluentApi/SqlServerReverseEngineerTestE2EContext.expected
@@ -271,6 +271,11 @@ namespace E2ETest.Namespace
                 entity.HasIndex(e => new { e.A, e.B })
                     .HasName("Test_PropertyConfiguration_Index");
 
+                entity.Property(e => e.ComputedDateTimeColumn)
+                    .HasColumnType("datetime")
+                    .HasComputedColumnSql("getdate()")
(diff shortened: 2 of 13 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.Design.FunctionalTests/ReverseEngineering/Expected/AllFluentApi/SqlServerReverseEngineerTestE2EContext.expected
```diff
@@ -271,14 +271,21 @@ namespace E2ETest.Namespace
                 entity.HasIndex(e => new { e.A, e.B })
                     .HasName("Test_PropertyConfiguration_Index");
 
+                entity.Property(e => e.ComputedDateTimeColumn)
+                    .HasColumnType("datetime")
+                    .HasComputedColumnSql("getdate()")
+                    .ValueGeneratedOnAddOrUpdate();
```
ErikEJ: GETDATE() is a default value, it is not computed on updates.... (unless you add some other RDBMS feature to do this)

lajones: Had a chat with @divega and we think this is a little different.

The definition of that column is `ComputedDateTimeColumn AS GETDATE()` i.e. it is a computed column, not a default value. We have a separate column to test a datetime2 column with a default value which is a computed expression (see the `WithDateDefaultExpression datetime2 NOT NULL DEFAULT(getdate())` column also on the `PropertyConfiguration` table). The 2 types of columns are treated differently in SQL Server metadata - the first is treated as a computed column (has an entry in `sys.computed_columns`); the second is not a computed column but does have a default value.

So, although technically you're right, the server does not compute a new value when a row is updated, in practice we keep the `ChangeTracker` up to date with updates by performing a select immediately after we've updated the data. Now on `SELECT` the value of this column will be updated, so the value stored in the `ChangeTracker` would likewise be updated. The closest we can come to this behavior is to assign it `ValueGeneratedOnAddOrUpdate()`.

ErikEJ: Got it, thanks!

src/Microsoft.EntityFrameworkCore.Relational.Design/Metadata/ColumnModel.cs
```diff
@@ -21,6 +21,8 @@ public class ColumnModel : Annotatable
         [CanBeNull]
         public virtual string DefaultValue { get; [param: CanBeNull] set; }
 
+        [CanBeNull]
```
ajcvickers: We decided in a design meeting some time back to not use these attributes for return values.

ajcvickers:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational.Design/Metadata/ColumnModel.cs
lajones: OK. Will remove (and similarly from the existing one above). Thx.

lajones:
Found a few more `[CanBeNull]`s on return values. Removed them too and checked in with 2acd2f1.

Issue #113, opened 2014-04-24
LINQ queries that compare related entities should compare based on their keys


Eilon:
I chatted with @divega and he suggested I log a bug. He thinks this may have worked in LINQ to SQL, but perhaps not in EF.

I have two related entities, for example, a photo gallery that contains many photos. In one part of my app I want to write this kind of query:
`_db.Photos.Where(p => p.Gallery == gallery)`
But apparently the gallery "instance" comparison doesn't work. I have to compare based on the PK value:
`_db.Photos.Where(p => p.GalleryId == galleryId)`

I'm not sure I'd say this is "hugely" important, but part of me had an expectation that this would work.

divega:
Just to add more details based on the discussion on #3141: In EF6 we did not support comparing an in-memory entity instance against an expression representing an entity in the database but we did have some support for a few patterns for comparing entities in the database, e.g. joins in grouping on an entity would in some cases work and the translation would expand the comparison between entities into a comparison of all the individual columns that were part of the entity. 

In the future in EF7.x we should consider supporting those scenarios as well, but when we do it we should consider translating comparisons between entities into comparisons between their keys.

PR #5019, opened 2016-04-08
Query: Implements entity equality comparison translation. Entity equa…

anpete:
…lity comparisons are transformed into entity key comparisons.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -23,6 +23,17 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 {
     public class SqlTranslatingExpressionVisitor : ThrowingExpressionVisitor
     {
+        private static readonly Dictionary<ExpressionType, ExpressionType> _inverseOperatorMap 
+            = new Dictionary<ExpressionType, ExpressionType>
+        {
+            { ExpressionType.LessThan, ExpressionType.GreaterThanOrEqual },
+            { ExpressionType.LessThanOrEqual, ExpressionType.GreaterThan },
+            { ExpressionType.GreaterThan, ExpressionType.LessThanOrEqual },
+            { ExpressionType.GreaterThanOrEqual, ExpressionType.LessThan },
+            { ExpressionType.Equal, ExpressionType.NotEqual },
+            { ExpressionType.NotEqual, ExpressionType.Equal }
+        };
+
         private readonly IRelationalAnnotationProvider _relationalAnnotationProvider;
         private readonly IExpressionFragmentTranslator _compositeExpressionFragmentTranslator;
         private readonly IMethodCallTranslator _methodCallTranslator;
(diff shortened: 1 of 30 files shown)


divega:
This is great! BTW you can resolve #113 with it.

Curious (because I couldn't find tests, sorry if I am missing something), do we now support:
- Comparing two entities from the database, e.g. joining on a nav prop?
- Grouping on an entity? (I know we won't translate it but I thought in memory it could work)

anpete:
@divega Comparing nav props was already largely handled by nav rewriting. There may be some DRY'ing up we need to do here (cc @maumar)

Client grouping already works as L2O - unless there is something more specific you had in mind?

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -333,8 +334,14 @@ private Expression ProcessComparisonExpression(BinaryExpression binaryExpression
...
                 = TransformNullComparison(leftExpression, rightExpression, binaryExpression.NodeType);
 
-            return nullExpression
-                   ?? Expression.MakeBinary(binaryExpression.NodeType, leftExpression, rightExpression);
+            if (nullExpression != null)
+            {
+                return nullExpression;
+            }
+
+            return leftExpression.Type == rightExpression.Type
```
maumar: should we compensate for nullability difference etc?

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/EntityEqualityRewritingExpressionVisitor.cs
```diff
@@ -0,0 +1,116 @@
...
+            Check.NotNull(queryModel, nameof(queryModel));
+
+            queryModel.TransformExpressions(Visit);
+        }
+
+        protected override Expression VisitBinary(BinaryExpression binaryExpression)
+        {
+            Check.NotNull(binaryExpression, nameof(binaryExpression));
+
+            var newExpression = base.VisitBinary(binaryExpression);
```
maumar: why is it done this way rather than

if (newExpression.NodeType == ExpressionType.Equal
                || newExpression.NodeType == ExpressionType.NotEqual)
{
  var newLeft = Visit(binaryExpression.Left);
  var newRight = Visit(binaryExpression.Right);
...
}

This seems to be the pattern we usually follow in those cases. Is there a situation where base visitor changes the NodeType of the expression?

test/Microsoft.EntityFrameworkCore.InMemory.FunctionalTests/QueryNavigationsInMemoryTest.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.EntityFrameworkCore.InMemory.FunctionalTests
 {
     public class QueryNavigationsInMemoryTest : QueryNavigationsTestBase<NorthwindQueryInMemoryFixture>
     {
-        public QueryNavigationsInMemoryTest(NorthwindQueryInMemoryFixture fixture)
+       public QueryNavigationsInMemoryTest(NorthwindQueryInMemoryFixture fixture)
```
maumar: nit: missing space before ctor definition

divega:
> Client grouping already works as L2O - unless there is something more specific you had in mind?

If I understand correctly and we perform reference equality and we never return duplicates within the same query this should be enough. Thanks.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
```diff
@@ -10,6 +10,17 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests
 {
     public class GearsOfWarQuerySqlServerTest : GearsOfWarQueryTestBase<SqlServerTestStore, GearsOfWarQuerySqlServerFixture>
     {
+        public override void Entity_equality_empty()
```
maumar: also worth adding test with optional navigation to verify that we properly compensate for the nullability change

anpete: Added. Disabled for now as I found #5043

anpete:
@maumar New version up.

maumar:
:shipit:

PR #5020, opened 2016-04-09
Minor tweaks to scripting migrations

### LINKED ISSUE #5012: "Done" message was shown when running dotnet ef migrations script
I'm working on ef migrations in web tool side. I use dotnet ef migrations script to generate T-SQL file of migrations. When this command is done, it will show a "Done" on screen, whose the code is from src/dotnet-ef/MigrationsScriptCommand.cs  (line 74 Reporter.Error.WriteLine("Done");) . This "Done" should not be handled by "Reporter.Error".

When I run the whole thing by msbuild tasks, the "Done" will be captured by msbuild.exe and it was recognized an error message. This makes people think this command has a problem and actually it is not. 

Is there an option not showing "Done" on screen or some other approach to skip "Done" ?

natemcmaster:
Fix #5012.

cc @divega

### DIFF
--- a/src/dotnet-ef/MigrationsScriptCommand.cs
+++ b/src/dotnet-ef/MigrationsScriptCommand.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.IO;
+using System.Text;
 using JetBrains.Annotations;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.Extensions.CommandLineUtils;
@@ -69,9 +70,10 @@ namespace Microsoft.EntityFrameworkCore.Commands
             else
             {
                 Reporter.Verbose.WriteLine("Writing SQL script to '" + output + "'".Bold().Black());
-                File.WriteAllText(output, sql);
+                File.WriteAllText(output, sql, Encoding.UTF8);
 
-                Reporter.Error.WriteLine("Done");
+                // TODO https://github.com/aspnet/EntityFramework/issues/4771
+                // Reporter.Error.WriteLine("Done");
             }
 
             return 0;


divega:
:shipit:

natemcmaster:
Cc @bricelam

src/dotnet-ef/MigrationsScriptCommand.cs
```diff
@@ -69,9 +70,9 @@ public static void Configure([NotNull] CommandLineApplication command)
             else
             {
                 Reporter.Verbose.WriteLine("Writing SQL script to '" + output + "'".Bold().Black());
-                File.WriteAllText(output, sql);
+                File.WriteAllText(output, sql, Encoding.UTF8);
 
-                Reporter.Error.WriteLine("Done");
+                Reporter.Output.WriteLine("Done");
```
bricelam: This goes against our general principle of only logging data to stdout.

natemcmaster: Verbose then?

lajones: `Information` would seem to be the right level to me.

natemcmaster: There are only three levels on Reporter: Error, Verbose, and Output.

lajones: :-( OK, `Verbose` sounds good then - but we should log something against `Reporter`. That doesn't seem to provide enough levels to me.

divega: @natemcmaster and I talked about this, in particular about @bricelam's [feedback](https://github.com/aspnet/EntityFramework/pull/5020/files#r59140707). It seems that what we have is consistent with the conventions of .NET CLI commands so we need to find out msdeploy can use this differently or if those conventions need revisiting.

bricelam: BTW, `Verbose` writes to stderr.

natemcmaster: But only if -v is specified, right? 

Diego and I talked in person. We are going to investigate a little more before/if we merge this. We want to avoid corner-casing this one command.

bricelam: Oh good point, yes.

natemcmaster: Logging "Done" has little value. How about we just remove altogether and revisit after RC2?

PR #5021, opened 2016-04-09
Cross-compile to netcore50

### LINKED ISSUE #4802: Latest RC2 packages do not install in latest UWP builds
The errors are currently in our dependencies, so there is likely nothing to do on our side other than test and report issues to the appropriate teams. I've reported the current issue with the `System.AppContext` package to .NET Core folks.

> System.AppContext 4.1.0-rc2-23826 provides a compile-time reference assembly for System.AppContext on UAP,Version=v10.0, but there is no run-time assembly compatible with win10-arm.
At line:1 char:1
install-package Microsoft.EntityFrameworkCore.Sqlite -pre
~~~~~~~~~~~~~
CategoryInfo : NotSpecified: (:) [Install-Package], Exception
FullyQualifiedErrorId : NuGetCmdletUnhandledException,NuGet.PackageManagement.PowerShellCmdlets.InstallPackageCommand

natemcmaster:
Fix powershell commands.

Fix #4802 

Related pull requests for dependencies: 

https://github.com/aspnet/Common/pull/107
https://github.com/aspnet/Caching/pull/178
https://github.com/aspnet/Options/pull/137
https://github.com/aspnet/PlatformAbstractions/pull/32
https://github.com/aspnet/DependencyInjection/pull/399
https://github.com/aspnet/Logging/pull/403
https://github.com/aspnet/Microsoft.Data.Sqlite/pull/242

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Commands/Design/Internal/StartupInvoker.cs
+++ b/src/Microsoft.EntityFrameworkCore.Commands/Design/Internal/StartupInvoker.cs
@@ -5,10 +5,12 @@ using System;
 using System.Linq;
 using System.Reflection;
 using JetBrains.Annotations;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.EntityFrameworkCore.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.PlatformAbstractions;
+#if !NETCORE50
+using Microsoft.AspNetCore.Hosting;
+#endif
 
 namespace Microsoft.EntityFrameworkCore.Design.Internal
 {
@@ -100,12 +102,14 @@ namespace Microsoft.EntityFrameworkCore.Design.Internal
 
         protected virtual IServiceCollection ConfigureHostServices([NotNull] IServiceCollection services)
         {
+#if !NETCORE50
             services.AddSingleton<IHostingEnvironment>(
                 new HostingEnvironment
                 {
                     ContentRootPath = _startupProjectDir,
                     EnvironmentName = _environment
                 });
+#endif
 
             services.AddLogging();
             services.AddOptions();
--- a/src/Microsoft.EntityFrameworkCore.Relational/project.json
+++ b/src/Microsoft.EntityFrameworkCore.Relational/project.json
@@ -10,13 +10,15 @@
     "xmlDoc": true
   },
   "dependencies": {
-    "Microsoft.EntityFrameworkCore": "1.0.0-*",
-    "System.Diagnostics.DiagnosticSource": "4.0.0-*"
+    "Microsoft.EntityFrameworkCore": "1.0.0-*"
   },
   "compile": "../Shared/*.cs",
   "resourceFiles": "Properties/Microsoft.EntityFrameworkCore.Relational.rd.xml",
   "frameworks": {
     "net451": {
+      "dependencies": {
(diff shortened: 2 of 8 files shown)


src/Microsoft.EntityFrameworkCore.Commands/Design/Internal/StartupInvoker.cs
```diff
@@ -100,12 +102,14 @@ private object Invoke(Type type, string[] methodNames, IServiceCollection servic
 
         protected virtual IServiceCollection ConfigureHostServices([NotNull] IServiceCollection services)
         {
+#if !NETCORE50
```
divega: For possibly improved logistics, consider merging the changes from https://github.com/aspnet/EntityFramework/pull/5037 as well, since UWP does not work until the commands work.

divega:
I am not able to validate if the changes to project.json files seem correct, but the rest LGTM.

natemcmaster:
cc @pranavkm @davidfowl can someone review/signoff on the project.json changes?

src/Microsoft.EntityFrameworkCore.Commands/tools/OperationHandlers.cs
```diff
@@ -32,17 +38,17 @@ public partial class OperationResultHandler : IOperationResultHandler
         private string _errorMessage;
         private string _errorStackTrace;
 
-        public virtual int Version => 0;
+        public virtual int Version { get { return 0; } }
```
pranavkm: Why change these?

natemcmaster: This file also needs to be compiled in c# 5.0 by powershell. See #5036

pranavkm: Shouldn't that go in a separate PR? If the hope here is that eventually we won't have to cross compile for netcore, we could revert this change in entirety.

src/Microsoft.EntityFrameworkCore.Relational/project.json
```diff
@@ -38,8 +40,29 @@
...
+    },
+    "netcore50": {
+      "dependencies": {
+        "Microsoft.NETCore.Platforms": {
+          "type": "build",
+          "version": "1.0.1-*"
+        },
+        "Microsoft.CSharp": "4.0.0",
+        "System.Data.Common": "4.0.0",
+        "System.Text.RegularExpressions": "4.0.10",
```
pranavkm: Why are some of these `type: build` and not others?

pranavkm:
Looks :shipit: to me. Might want to get @davidfowl \ @anurse \ @pakrym to look at this in case they spot issues working with the CLI.

src/Microsoft.EntityFrameworkCore.Relational/project.json
natemcmaster: In short, to make dotnet-restore pass but they don't need to be in the nuspec because these packages will only ever be installed alongside Microsoft.NETCore.UniversalWindowsPlatform.

"build" dependencies are there only to upgrade the version of the package dotnet-restore pulls. Without this (for example without "System.Dynamic.Runtime" build dependency), nuget fails. "Could not find runtime-assembly for win7-x64 for System.Dynamic.Runtime 4.0.0.

The same is true of including ms.netcore.platforms. It doesn't need to be in the nuspec.

PR #5023, opened 2016-04-09
Fix/supress ReSharper warnings for Microsoft.EntityFrameworkCore.SqlServer

AndriySvyryd:


anpete:
:shipit:

src/Microsoft.EntityFrameworkCore.Commands/Design/DbContextOperations.cs
```diff
@@ -85,7 +85,7 @@ public virtual Type GetContextType([CanBeNull] string name)
             var contexts = new Dictionary<Type, Func<DbContext>>();
 
             // Look for IDbContextFactory implementations
-            var contextFactories = _startupAssembly.GetConstructibleTypes()
+            var contextFactories = _startupAssembly.GetConstructableTypes()
```
lajones: Weird that R# insists on this. AFAICS "constructible" is more normal, but if it stops R# warnings then so be it.

lajones:
:shipit:

src/Microsoft.EntityFrameworkCore.Commands/Design/DbContextOperations.cs
AndriySvyryd: The dictionary it's using is not very flexible, e.g. It also marks behaviour as misspelled

PR #5039, opened 2016-04-11
Fix/supress ReSharper warnings for Microsoft.EntityFrameworkCore.Commands

AndriySvyryd:


src/Microsoft.EntityFrameworkCore.Commands/Scaffolding/Internal/Configuration/IndexConfiguration.cs
```diff
@@ -25,15 +25,17 @@ public class IndexConfiguration : IFluentApiConfiguration
         public virtual string LambdaIdentifier { get; }
         public virtual Index Index { get; }
 
-        public virtual bool HasAttributeEquivalent { get; }
+        public virtual bool HasAttributeEquivalent { get; set; }
```
divega: Is the setter actually needed. Can't tell from looking at the code.

AndriySvyryd: No. I'll change it to
```
public virtual bool HasAttributeEquivalent => false;
```

divega:
LGTM, probably a good idea for @lajones to take a look at the changes in scaffolding in case there are specific reasons we had things how we had them (e.g. the protected vs. virtual members in `ModelConfiguration`).

src/Microsoft.EntityFrameworkCore.Commands/Scaffolding/Internal/Configuration/ModelConfiguration.cs
```diff
@@ -18,48 +19,48 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal.Configuration
...
             [NotNull] CSharpUtilities cSharpUtilities,
             [NotNull] ScaffoldingUtilities scaffoldingUtilities)
         {
             Check.NotNull(configurationFactory, nameof(configurationFactory));
             Check.NotNull(model, nameof(model));
             Check.NotNull(customConfiguration, nameof(customConfiguration));
-            Check.NotNull(extensionsProvider, nameof(extensionsProvider));
+            Check.NotNull(annotationProvider, nameof(annotationProvider));
             Check.NotNull(cSharpUtilities, nameof(scaffoldingUtilities));
             Check.NotNull(scaffoldingUtilities, nameof(cSharpUtilities));
```
lajones: Can we change the above 2 lines so that the `nameof()` matches the object we're testing? Looks like they got mixed up at some stage.

src/Microsoft.EntityFrameworkCore.Commands/Scaffolding/Internal/Configuration/ModelConfiguration.cs
```diff
@@ -18,48 +19,48 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal.Configuration
...
-        protected const string DbContextSuffix = "Context";
-        protected const string DefaultDbContextName = "Model" + DbContextSuffix;
-        protected static readonly KeyDiscoveryConvention _keyDiscoveryConvention = new KeyDiscoveryConvention();
-        protected static readonly KeyConvention _keyConvention = new KeyConvention();
-
-        protected readonly ConfigurationFactory _configurationFactory;
-        protected List<OptionsBuilderConfiguration> _onConfiguringConfigurations;
-        protected SortedDictionary<EntityType, EntityConfiguration> _entityConfigurationMap;
+        private const string DbContextSuffix = "Context";
+        private const string DefaultDbContextName = "Model" + DbContextSuffix;
```
lajones: I think making the above 2 const's `public` (or keeping them `protected`) would be better. So that an inheritor could easily override them. Other changes from `protected` to `private` are fine.

lajones:
After the minor comments above :shipit:

PR #5044, opened 2016-04-12
Fix/supress ReSharper warnings in providers

AndriySvyryd:
@lajones

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.InMemory/Extensions/Internal/InMemoryLoggerExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.InMemory/Extensions/Internal/InMemoryLoggerExtensions.cs
@@ -5,7 +5,8 @@ using System;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.Extensions.Logging;
 
-namespace Microsoft.EntityFrameworkCore.Internal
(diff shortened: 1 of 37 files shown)


src/Microsoft.EntityFrameworkCore.InMemory/Extensions/Internal/InMemoryLoggerExtensions.cs
```diff
@@ -5,7 +5,8 @@
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.Extensions.Logging;
 
-namespace Microsoft.EntityFrameworkCore.Internal
+// ReSharper disable once CheckNamespace
+namespace Microsoft.EntityFrameworkCore
```
lajones: Are you sure on this? It's in the `Internal` directory?

AndriySvyryd: In this case the class itself is internal

lajones: Hmmm. OK.

lajones:
:shipit:

PR #5046, opened 2016-04-12
Ensure properties matching CLR properties are always marked as non-shadow

### LINKED ISSUE #4714: Ensure properties added via EntityBuilder that match CLR properties are marked as non-shadow
Currently it is possible to add a property using the string-based API with the same name as a property on the CLR type that hasn't been discovered by a convention, resulting in a shadow property by default.

AndriySvyryd:
Fixes #4714


AndriySvyryd:
Will rewrite

PR #5048, opened 2016-04-13
Fixed a bug with group join returning incorrect outer entities

joshcomley:
Suppose the following model:

    public class Order
    {
        public string Id { get; set; }
        public int? Id { get; set; }
        public Customer Customer { get; set; }
    }

    public class Customer
    {
        public int Id { get; set; }
        public List<Order> Orders { get; set; }
    }

**Note: it is crucial we have a nullable Customer entity in this relationship to trigger the group join**

Then suppose we have the following data:

- Order ID 1 - Customer ID 1
- Order ID 2 - Customer ID 1
- Order ID 3 - Customer ID 1
- Order ID 4 - Customer ID [null]

Currently your results set would look like this:

- Order ID 1 - Customer ID 1
- Order ID 1 - Customer ID 1
- Order ID 1 - Customer ID 1
- Order ID 4 - Customer ID [null]

This is (I think) because we are not checking in the group join if the outer element has changed.

I am unable to run tests on my machine, so my apologies if this is a bad pull request.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/BufferedEntityShaper`.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/BufferedEntityShaper`.cs
@@ -25,7 +25,16 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
         public override Type Type => typeof(TEntity);
 
-        public virtual TEntity Shape(QueryContext queryContext, ValueBuffer valueBuffer)
+	    public virtual object GetKey(QueryContext queryContext, ValueBuffer valueBuffer)
+	    {
+		    return queryContext.QueryBuffer.GetEntityKey(
+			    Key,
+			    new EntityLoadInfo(valueBuffer, Materializer),
+			    queryStateManager: IsTrackingQuery,
+			    throwOnNullKey: !AllowNullResult);
+	    }
+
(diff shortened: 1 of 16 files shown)


joshcomley:
Ah yep I broke the build. I can't get the main build script working on my machine, so I am unsure how to run the tests locally

natemcmaster:
@joshcomley you should be able to run `build initialize` and then open VS and run tests from there.

joshcomley:
Ah I just saw this, yes I did that on my clone and it didn't work. I then made a fork for the tiny change, and assumed that would fail on `build initialize` too, but I just tried it and it succeeds where the clone didn't, so now I can try to get the tests to pass before pushing back to this branch.

joshcomley:
The build and tests are now fixed for this change. I have added the ability to resolve the key object of the outer part of the left join to the Shaper class and relevant dependencies.

anpete:
Hi @joshcomley. Thanks for the PR! Any chance you can add a corresponding query test? You should be able to add it to ComplexNavigationsQueryTestBase - I think the model used in those tests has an optional one-to-many.

joshcomley:
@anpete Sure thing, will sort this later today or tomorrow morning! I was intending to do it but hadn't had time to research the tests to find where is the right place, so thank you for the pointer :)

maumar:
@joshcomley would you be able to clean up the pull request  (fix formatting, remove commented code, add test, rebase, squash) - we decided to take the fix to RC2

maumar:
The bug got fixed in e169d17243760ac3eece58bf8187905665ab67f6, closing this PR

Issue #266, opened 2014-05-28
Blocked: SequenceValueGenerator may fail with LocalDb due to LocalDb concurrency issue with sequences


ajcvickers:
Stack trace from CI failure:

```
The statement has been terminated.
   at System.Data.SqlClient.SqlCommand.<ExecuteDbDataReaderAsync>b__24(Task`1 result)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at Microsoft.Data.Entity.Relational.Update.BatchExecutor.<ExecuteBatchAsync>d__1.MoveNext() in C:\BuildAgent\work\9114aaddeefdc8e8\src\Microsoft.Data.Entity.Relational\Update\BatchExecutor.cs:line 68
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
   at Microsoft.Data.Entity.Relational.Update.BatchExecutor.<ExecuteAsync>d__1.MoveNext() in C:\BuildAgent\work\9114aaddeefdc8e8\src\Microsoft.Data.Entity.Relational\Update\BatchExecutor.cs:line 50
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
[...]

PR #5062, opened 2016-04-14
Re-enabling test that was disabled due to SQL Server bug

ajcvickers:
Bug is now fixed and pushed to releases/service packs. See #266.


AndriySvyryd:
:shipit:

PR #5064, opened 2016-04-14
Move abstract functional tests to src projects

### LINKED ISSUE #5055: Functional Tests: Move Specification tests to source
EF has following test projects which are being published to nuget.org
-	Microsoft.EntityFrameworkCore.Relational.FunctionalTests
-	Microsoft.EntityFrameworkCore.FunctionalTests
-	Microsoft.EntityFrameworkCore.Relational.Design.FunctionalTests

These tests depend on dotnet-test-xunit. These tests should be split into Facts which rely on xunit (which will be published to nuget.org) and runnable component (which does not get published).

smitpatel:
fixes #5055 
Move abstract classes for function tests to src folder
Package name changes for provider-writers
MS.EFCore.FunctionalTests -> MS.EFCore.Specification.Tests
MS.EFCore.Relational.FunctionalTests -> MS.EFCore.Relational.Specification.Tests
MS.EFCore.Relational.Design.FunctionalTests -> MS.EFCore.Relational.Design.Specification.Tests

Above packages are class libraries. They contain only abstract test bases & fixtures which providers can use to test by plugging in their implementation. They will be published on nuget.org.

Other changes:
Provider specific tests are derived from *.Specification.Tests directly.
EFCore.FunctionalTests contain only runnable tests. Other 2 project did not have any concrete test classes and hence removed.
Class level skip conditions are moved to derived classes instead of testbase so that if tests are running for other providers then they can verify them.

build.cmd runs successfully.
csproj based solution builds & runs all tests successfully.
xproj based solution builds fine, fails to discover any tests in my system.

Haven't incorporated namespace changes according to project name changes yet. (my resharper crashed doing so). Are namespaces supposed to be changed? @divega , @natemcmaster


smitpatel:
@natemcmaster - travis build failing due to "integrated security" though there is nothing in this PR which is changing connection strings. Any insights?

ErikEJ:
@smitpatel Hoping that other than the package name change this will not impact how I use the test libraries...

smitpatel:
@ErikEJ - It shouldn't impact the way tests are used.
The changes I had to make to SqlServer.FunctionalTests are below
1. Package name change
2. Adding dotnet-test-xunit to project.json which was earlier coming from Core.FunctionalTests transitively.
3. MonoVersionConditions on tests which fails on lower version of Mono. (depends on provider if their implementation can work on Mono below 4.2.0 or not)
4. Minor class rename to UpdatesRelationalTest to UpdatesRelationalTestBase
If you still face issue while moving to new package names, I would be happy to assist you.

src/Microsoft.EntityFrameworkCore.Specification.Tests/Microsoft.EntityFrameworkCore.Specification.Tests.csproj
```diff
@@ -0,0 +1,209 @@
...
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <Import Project="..\..\tools\EntityFramework.props" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{1A73D95E-E8B5-4F96-908C-7B040E4F7AFE}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Microsoft.EntityFrameworkCore.Specification.Tests</RootNamespace>
```
natemcmaster: The rootnamespace can be "Microsoft.EntityFrameworkCore"

src/Microsoft.EntityFrameworkCore.Specification.Tests/Microsoft.EntityFrameworkCore.Specification.Tests.xproj
```diff
@@ -6,13 +6,13 @@
   </PropertyGroup>
   <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''" />
   <PropertyGroup Label="Globals">
-    <ProjectGuid>7ba2e7db-a85f-4d3b-a610-ff941adeffd6</ProjectGuid>
-    <RootNamespace>Microsoft.EntityFrameworkCore</RootNamespace>
+    <ProjectGuid>aab3c3cd-0bdd-494b-83ad-f3295aaa13cf</ProjectGuid>
+    <RootNamespace>Microsoft.EntityFrameworkCore.Specification.Tests</RootNamespace>
```
natemcmaster: Same here.

src/Microsoft.EntityFrameworkCore.Specification.Tests/project.json
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "frameworks": {
+    "netstandard1.3": {
+      "imports": [
+        "dnxcore50",
+        "portable-net452+win81"
+      ],
+      "dependencies": {
+        "Microsoft.CSharp": "4.0.1-*",
+        "Microsoft.NETCore.Platforms": "1.0.1-*"
```
natemcmaster: I think we may not need this here....but only way to know is to try. Does this build w/o this dependency?

src/Microsoft.EntityFrameworkCore.Specification.Tests/Microsoft.EntityFrameworkCore.Specification.Tests.csproj
smitpatel: Can you write what should be root namespace for all 3 projects?
I will update them all. There was mismatch between csproj & xproj so I made them consistent with csproj.

test/Microsoft.EntityFrameworkCore.InMemory.FunctionalTests/AsyncQueryInMemoryTest.cs
```diff
@@ -2,9 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.EntityFrameworkCore.FunctionalTests;
+using Microsoft.EntityFrameworkCore.FunctionalTests.TestUtilities.Xunit;
 
 namespace Microsoft.EntityFrameworkCore.InMemory.FunctionalTests
 {
+    [MonoVersionCondition(Min = "4.2.0", SkipReason = "Async queries will not work on Mono < 4.2.0 due to differences in the IQueryable interface")]
```
natemcmaster: We'll have to try this, but I think inmemory will work just fine on Mono 4.0

natemcmaster:
Don't forget to update nugetpackageverifier.json

natemcmaster:
cc @muratg heads up: we are adding new packages.

natemcmaster:
> travis build failing due to "integrated security" though there is nothing in this PR which is changing connection strings. 

@smitpatel we need to take a closer look at the xunit test utilities. It appears to me the test conditions are not actually being evaluated correctly so the tests are running on Linux when they shouldn't.

src/Microsoft.EntityFrameworkCore.Specification.Tests/Microsoft.EntityFrameworkCore.Specification.Tests.csproj
natemcmaster: I'm not 100% sure. Let's discuss in design meeting today.

ErikEJ:
@smitpatel Got it - you are just making these test libraries an official part of the "product", and as such, they belong in src folder, not in test

natemcmaster:
> you are just making these test libraries an official part of the "product"

Correct. We don't want to ship packages that have a dependency on the coreclr test runner, and we also want to be consistent with how other aspnet packages ship their test specs. For example: https://github.com/aspnet/DependencyInjection/tree/dev/src/Microsoft.Extensions.DependencyInjection.Specification.Tests

muratg:
@natemcmaster Please make sure to update packages/packages.csv in Coherence repo as well as the nuget package verifier JSON file.

cc @pranavkm heads up

pranavkm:
If the packages are being moved to src, we should move these to the AdxShipping category in NuGetPackageVerifier.json

smitpatel:
Updated nugetpackageverifier.json
Submitted PR to coherence & coherence-signed

src/Microsoft.EntityFrameworkCore.Specification.Tests/Microsoft.EntityFrameworkCore.Specification.Tests.csproj
smitpatel: changing namespace to Microsoft.EntityFrameworkCore.Specification.Tests & related
see #5073

natemcmaster:
LGTM. Assuming we get the Linux tests passing, I don't see anything less that needs to change. (Not a :shipit: untill we get Linux passing and ask-mode permission.)

smitpatel:
travis passed 😌

Issue #3625, opened 2015-11-01
Change tracking doesn't handle empty property name properly


flagbug:
When implementing `INotifyPropertyChanged` on a model and sending an empty string as `PropertyName` in the `PropertyChangedEventArgs` event args, EF throws an `ArgumentException` with the message `The string argument 'propertyName' cannot be empty`

The remarks section of the [MSDN Documentation page] (https://msdn.microsoft.com/en-us/library/system.componentmodel.inotifypropertychanged.propertychanged(v=vs.110).aspx)  clearly states: 

> The PropertyChanged event can indicate all properties on the object have changed by using either null or String.Empty as the property name in the PropertyChangedEventArgs.

so EF should handle this properly.

Here is a stacktrace:

``` 	
EntityFramework.Core.dll!Microsoft.Data.Entity.Utilities.Check.NotEmpty(string value, string parameterName)	Unknown
 	EntityFramework.Core.dll!Microsoft.Data.Entity.Metadata.EntityType.FindProperty(string propertyName)	Unknown
 	EntityFramework.Core.dll!Microsoft.Data.Entity.Metadata.EntityType.Microsoft.Data.Entity.Metadata.IEntityType.FindProperty(string propertyName)	Unknown
 	EntityFramework.Core.dll!Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntrySubscriber.TryGetPropertyBase(Microsoft.Data.Entity.Metadata.IEntityType entityType, string propertyName)	Unknown
 	EntityFramework.Core.dll!Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntrySubscriber.SnapshotAndSubscribe.AnonymousMethod__2(object s, System.ComponentModel.PropertyChangedEventArgs e)	Unknown
>	[REDACTED].OnPropertyChanged(string propertyName) Line 175	C#
```

Going into the EF source, I found that there is an "empty or null" check at this line: https://github.com/aspnet/EntityFramework/blob/46f2c8d3fdb74d91c24192ef0f2c416a4ceba02d/src/EntityFramework.Core/Metadata/Internal/EntityType.cs#L827

I would send a PR, but I don't have enough knowledge about the internals of EF to change this behavior

PR #5072, opened 2016-04-14
Handle null/empty in property name in notification events

ajcvickers:
See #3625. Passing in null or the empty string indicates that all properties are changing/changed. We now honor this, except for properties marked as read-only. We ignore these since we don't track original values for full notification entities and hence have no way of knowing if they are really changed or not, and we cannot mark them as modified. So we assume that since they are read only they have not been changed.


lajones:
:shipit:

Issue #4988, opened 2016-04-05
AddDbContext<TContext>() is only useful if DbContext will be used in web requests in an ASP.NET application


divega:
The main issue IMO is that  `AddDbContext<TContext>()` always adds the context as a scoped service. This is very useful for the 80% scenario in ASP.NET in which a single DbContext instance is resolved and used per request, but it can be problematic or require additional knowledge (e.g. about DI scopes) if anyone is trying to use it outside this scenario, e.g.:

- Resolve an instance of `TContext` in other parts of the application, e.g. startup
- In a non-web application that uses DI
- In test code

E.g. the following code will fail:

``` C#
var serviceProvider = new ServiceCollection()
    .AddEntityFrameworkSqlServer()
    .AddDbContext<CommandContext>(
        options => options.UseSqlServer(_connectionString))
    .BuildServiceProvider();

using (var context = serviceProvider.GetService<CommandContext>())
{
    // do some work
}

using (var context = serviceProvider.GetService<CommandContext>())
{
    // this will return the same instance which has already been disposed
}
```

There a couple of things I believe we could consider doing:

1. Rename `AddDbContext<TContext>()` to `AddScopedDbContext<TContext>()`
 - In addition we could create methods for different lifetimes, e.g. `AddTransientDbContext<TContext>()`, `AddSingletonDbContext<TContext>()` and keep using the scoped version in the default ASP.NET templates.
2. Move this method to a different package that is ASP.NET specific (possibly alongside other things that are ASP.NET specific, such as things that deal with `Startup`).

rowanmiller:
Option 2 :+1: :smile: 

BTW the example of wrapping services from a ServiceProvider in a using statement is generally a bad idea anyway. The ServiceProvider should be responsible for disposing anything it creates at the appropriate time.

divega:
> the example of wrapping services from a ServiceProvider in a using statement is generally a bad idea anyway

Agreed. The example is a bit artificial because disposing is not needed or necessarily a good idea (especially since the instance is scoped). The main point is the user didn't know that nor wanted the instances to be the same.

divega:
@ajcvickers will come up with a proposal.

dance:
Could you please take a look at https://github.com/aspnet/EntityFramework/compare/dev...dance:dbcontext-lifetimes if it is right direction to go? There are no tests yet - if you approve, i'll add them.
Some thoughts:
[...]

PR #5078, opened 2016-04-14
Allow AddDbContext to add context as transient or singleton

ajcvickers:
See #4988

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
@@ -58,18 +58,27 @@ namespace Microsoft.Extensions.DependencyInjection
         ///         work if you have one derived context type registered in your <see cref="IServiceProvider" />.
         ///     </para>
         /// </param>
+        /// <param name="contextLifetime"> The lifetime with which to register the DbContext service in the container. </param>
         /// <returns>
         ///     A builder that allows further Entity Framework specific setup of the <see cref="IServiceCollection" />.
         /// </returns>
         public static IServiceCollection AddDbContext<TContext>(
             [NotNull] this IServiceCollection serviceCollection,
-            [CanBeNull] Action<DbContextOptionsBuilder> optionsAction = null)
+            [CanBeNull] Action<DbContextOptionsBuilder> optionsAction = null,
+            ServiceLifetime contextLifetime = ServiceLifetime.Scoped)
             where TContext : DbContext
-            => AddDbContext<TContext>(serviceCollection, (p, b) => optionsAction?.Invoke(b));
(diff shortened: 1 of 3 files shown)


src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
```diff
@@ -63,6 +64,13 @@ public static class EntityFrameworkServiceCollectionExtensions
         /// </returns>
         public static IServiceCollection AddDbContext<TContext>(
             [NotNull] this IServiceCollection serviceCollection,
+            ServiceLifetime contextLifetime,
+            [CanBeNull] Action<DbContextOptionsBuilder> optionsAction = null)
+            where TContext : DbContext
+            => AddDbContext<TContext>(serviceCollection, contextLifetime, (p, b) => optionsAction?.Invoke(b));
+
```
rowanmiller: Do we know how this plays with IntelliSense... I was picturing this being the last parameter so that it was out of the way... I'm worried it might show up before the overload without it in IntelliSense now. Could we make it last and options with a default value?

ajcvickers: Yes, we can make it last. It just seemed to me to read better as the first parameter before the lambda, especially when the lambda is relatively long and multi-lined.

ajcvickers:
@rowanmiller @divega Updated.

rowanmiller:
:shipit: from an API perspective

src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
```diff
@@ -58,18 +58,27 @@ public static class EntityFrameworkServiceCollectionExtensions
         ///         work if you have one derived context type registered in your <see cref="IServiceProvider" />.
         ///     </para>
         /// </param>
+        /// <param name="contextLifetime"> The lifetime with which to register the DbContext service in the container. </param>
```
divega: The description of the new parameter seems good, but I think there are a couple of paragraphs that are no longer true (unless I am missing something):

- "You do not need to expose a `<see cref="DbContextOptions" />` constructor parameter for the options to be passed to the context. If you choose to expose a constructor parameter, we recommend typing it as the generic context. If you choose to..."
-   "`<returns>` A builder that allows further Entity Framework specific setup of the `<see cref="IServiceCollection" />`."

divega:
:shipit: (after reviewing the method's XML comments)

ajcvickers:
@rowanmiller Can you look at the XML docs and suggest how you would like it?

rowanmiller:
Just merge it, I opened https://github.com/aspnet/EntityFramework/issues/5102 (assigned to me) to more generally review the docs around options, DI, etc. in light of all these changes we have made.

divega:
Sounds good.

PR #5079, opened 2016-04-14
Make the Metadata to .NET types mapping readonly

### LINKED ISSUE #4714: Ensure properties added via EntityBuilder that match CLR properties are marked as non-shadow

### LINKED ISSUE #4934: Metadata: make ClrType readonly
This referes to `IMutableEntityType.ClrType` and `IMutableProperty.ClrType`. This will allow to remove `ICanFindEntityType.AddEntityType` and make the code simpler.

AndriySvyryd:
Make IMutableEntityType.ClrType, IMutableProperty.ClrType and IMutableProperty.IsShadowProperty readonly

Properties added using EntityTypeBuilder will always be non-shadow if the name matches a CLR property.
Properties added using EntityType can be shadow even if the name matches a CLR property.

Fixes #4934
Fixes #4714


ajcvickers:
:shipit:

Issue #5076, opened 2016-04-14
Exception message when updating to RC2


ajcvickers:
This is the message you get if you update an RC1 app with a parameterless DbContext constructor to use RC2. Should we change it?

> No database provider has been configured for this DbContext. A provider can be configured by overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider. If AddDbContext is used, then also ensure that your DbContext type accepts a DbContextOptions object in its constructor.

I think we should make it clear that you should use a generic DbContextOptions type, but other than that it looks okay to me.

@rowanmiller @natemcmaster @divega

rowanmiller:
Yep I think just changing `DbContextOptions` to `DbContextOptions<TContext>` would be enough for that. 

Perhaps we should also point out that you should pass the options to the base constructor :smile: 

That would leave us with...

> No database provider has been configured for this DbContext. A provider can be configured by overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider. If AddDbContext is used, then also ensure that your DbContext type accepts a DbContextOptions<TContext> object in its constructor and passes it to the base constructor for DbContext.

ajcvickers:
@rowanmiller @divega Triage for RC2?

rowanmiller:
Sounds good to me given this is something folks will hit when upgrading an existing app to RC2

ajcvickers:
@divega Waiting for your input.

divega:
I agree this should be in RC2. The message also looks good.

divega:
Marking as RC2.

OlsonDev:
Just came across this and didn't find the mention of `<TContext>` in the exception message very helpful. I think you should also mention `ApplicationDbContext` is the default implementation. I had to create a project from a VS template to see which concrete core implementation to use.

PR #5080, opened 2016-04-14
Update exception message seen for parameterless context constructor

ajcvickers:
Emphasizes that generic DbContextOptions should be used.

See #5076


divega:
:shipit:

Issue #3253, opened 2015-09-28
Allow named in-memory stores


ajcvickers:
We decided in API review to remove the "persist" flag from the in-memory provider and replace with a string that could be used to create a named in-memory store. However, we were pretty much never using the persist: false anyway, so I just removed it and created a bug to triage adding named databases so we could triage appropriately.

rowanmiller:
At the moment, do you get one per ServiceProvider, or one per context type per ServiceProvider? Either way, I don't think we need the named database capability for RC.

ajcvickers:
You get one table per entity type per service provider.

rowanmiller:
Reactivating (presumably for the backlog) as I am writing the docs for in-memory testing and this is not very easy at the moment. I think named instances could be a good solution to avoid folks having to muck around with ServiceCollections.

NeilCross:
Sorry for bringing up an old thread, but currently the [Testing with InMemory](https://docs.efproject.net/en/latest/miscellaneous/testing.html) in the EF docs: 
mentions requiring a workaround to this issue.  

> Caution
Currently you need to use ServiceCollection and IServiceProvider to control the scope of the InMemory database, which adds complexity to your tests. In the next release after RC2, there will be improvements to make this easier, see issue #3253 for more details.

Now 1.0.0 is released and I assume this warning in the docs not required anymore, and a ServiceCollection is no longer required to instantiate an in-memory context
Can you confirm this for me?

I'm assuming the simplest way to instantiate an in-memory context now is:
> var builder = new DbContextOptionsBuilder<BloggingContext>();
builder.UseInMemoryDatabase();
var context = new BloggingContext(builder.options)

rowanmiller:
@NeilCross opened https://github.com/aspnet/EntityFramework.Docs/issues/287 to fix the docs. The docs are currently migrating to a new system, so I'll fix once the migration is done.

NeilCross:
Good to know, thanks Rowan.

logamanig:
Following code works perfectly for me,
// Requires separate DI Service provider to isolate db instance as InMemory database is also shared within 
            // the class scope since the App Service Factory is configured to per scope instance for dbcontext
            var internalServiceProviderForDbContext = new ServiceCollection()
                .AddEntityFrameworkInMemoryDatabase()
                .BuildServiceProvider();
            var options = new DbContextOptionsBuilder<PegPortalDbContext>()
[...]

PR #5081, opened 2016-04-15
Allow named in-memory stores.

ajcvickers:
Issue #3253

Makes it easier to write isolated tests while still sharing a service provider between tests.

/cc @rowanmiller For the API

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
@@ -14,6 +14,37 @@ namespace Microsoft.EntityFrameworkCore
     {
         public static DbContextOptionsBuilder UseInMemoryDatabase(
             [NotNull] this DbContextOptionsBuilder optionsBuilder,
+            [CanBeNull] string databaseName,
+            [CanBeNull] Action<InMemoryDbContextOptionsBuilder> inMemoryOptionsAction = null)
+        {
+            Check.NotNull(optionsBuilder, nameof(optionsBuilder));
+
+            var extension = optionsBuilder.Options.FindExtension<InMemoryOptionsExtension>();
+            extension = extension != null
+                ? new InMemoryOptionsExtension(extension)
+                : new InMemoryOptionsExtension();
+
+            if (databaseName != null)
+            {
+                extension.StoreName = databaseName;
+            }
+            ((IDbContextOptionsBuilderInfrastructure)optionsBuilder).AddOrUpdateExtension(extension);
+
+            inMemoryOptionsAction?.Invoke(new InMemoryDbContextOptionsBuilder(optionsBuilder));
+
+            return optionsBuilder;
+        }
+
+        public static DbContextOptionsBuilder<TContext> UseInMemoryDatabase<TContext>(
+            [NotNull] this DbContextOptionsBuilder<TContext> optionsBuilder,
+            [CanBeNull] string databaseName,
+            [CanBeNull] Action<InMemoryDbContextOptionsBuilder> inMemoryOptionsAction = null)
+            where TContext : DbContext
+            => (DbContextOptionsBuilder<TContext>)UseInMemoryDatabase(
+                (DbContextOptionsBuilder)optionsBuilder, databaseName, inMemoryOptionsAction);
+
+        public static DbContextOptionsBuilder UseInMemoryDatabase(
+            [NotNull] this DbContextOptionsBuilder optionsBuilder,
             [CanBeNull] Action<InMemoryDbContextOptionsBuilder> inMemoryOptionsAction = null)
         {
             Check.NotNull(optionsBuilder, nameof(optionsBuilder));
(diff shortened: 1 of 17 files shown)


src/Microsoft.EntityFrameworkCore.InMemory/Infrastructure/InMemoryDbContextOptionsBuilder.cs
```diff
@@ -22,6 +22,9 @@ public InMemoryDbContextOptionsBuilder([NotNull] DbContextOptionsBuilder options
         public virtual InMemoryDbContextOptionsBuilder IgnoreTransactions()
             => SetOption(e => e.IgnoreTransactions = true);
 
+        public virtual InMemoryDbContextOptionsBuilder UseNamedDatabase([NotNull] string databaseName)
+            => SetOption(e => e.StoreName = Check.NotEmpty(databaseName, nameof(databaseName)));
+
```
rowanmiller: I think I was expecting this to be an optional param on the `UseInMemoryDatabase()` method, rather than something that is chained. I'm not really fussed either way... but just curious if you considered that as an option?

ajcvickers: I was going to do that at first too, but I recalled that we had had a discussion that parameters on UseInMemoryDatabase or similar were more for required configuration, like the connection or connection string, as opposed to optional modifiers.

I'm open to doing it either way. @divega?

divega: I may need more data to form a helpful opinion, but assuming I understand how things work it seems to me that named in-memory stores is probably the pit of success and I would probably never use it without a name again. So having it as a parameter in the method feels better to me. It also feels more consistent with the other providers in which you pass the connection string here. 

(I even wonder about having to pass a null name explicitly to get the current functionality would be better than having a parameter less version of the method, but that might be one of my extreme points of view :smile:)

ajcvickers: @divega Personally I think that is a little strong. We use the in-memory store for the majority of our tests and we don't use a named instance and we have not run into issues. When we need tests to be isolated from each other we isolate using the service provider, which ensures there is no shared state between the tests. You could argue that using named stores is a pit of failure because it encourages tests to be written that share state. I think that would be too strong too.

Using a named in-memory store is good for those people who are okay (either ignorantly or pragmatically) with sharing state between tests, but don't want to share the actual store.

rowanmiller: I think the other use case for InMemory is when you are stubbing out the actual app, but don't want to hit an actual database. Admittedly in this case it would be very easy just to give it a name, so I don't think it's a big deal, but my preference would be to leave it optional. 

I really don't have a strong opinion either way though.

divega: > Using a named in-memory store is good for those people who are okay (either ignorantly or pragmatically) with sharing state between tests, but don't want to share the actual store. 

What are the possible unintended consequences of sharing the service container for a developer writing functional tests for their application code? Shouldn't we be able to assert that named in-memory stores provide enough isolation for this?

It makes more sense for me to have more complete isolation in most of our own tests (I hope we are not missing bugs we would only detect if we reused the service container though :smile:)

After this discussion I still feel that most EF Core customers using in-memory databases should look into using named ones. Discoverability and similarity with the other database providers makes me prefer the name to be an argument in an overload of `UseInMemoryDatabase()` method. You conviced me that it is ok to have a paremterless overload. We can discuss this in person.

ajcvickers:
Updated with API change to optional parameter on UseInMemoryDatabase. Still needs review.

src/Microsoft.EntityFrameworkCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
```diff
@@ -14,6 +14,37 @@ public static class InMemoryDbContextOptionsExtensions
...
+            ((IDbContextOptionsBuilderInfrastructure)optionsBuilder).AddOrUpdateExtension(extension);
+
+            inMemoryOptionsAction?.Invoke(new InMemoryDbContextOptionsBuilder(optionsBuilder));
+
+            return optionsBuilder;
+        }
+
+        public static DbContextOptionsBuilder<TContext> UseInMemoryDatabase<TContext>(
+            [NotNull] this DbContextOptionsBuilder<TContext> optionsBuilder,
+            [CanBeNull] string databaseName,
```
anpete: Should this be required on this overload?

divega:
:shipit:

divega:
I just noticed @anpete was also reviewing, so I will let him finish :smile:

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
ajcvickers: If you're calling the method dynamically it can be useful to be able to just pass null, like we do in our tests. But I don't  have a strong opinion.

anpete: I figured that might be the reason. Just wanted to check it was by design.

ajcvickers:
Merged

PR #5082, opened 2016-04-15
Temporary workaround for Newtonsoft.Json issues with AssemblyLoadContext

natemcmaster:
This is to unblock other teams.

Temporary workaround for https://github.com/dotnet/cli/issues/2514. 

We may need to change dotnet-ef's execution strategy to permanently fix version issues with using `AssemblyLoadContext` from a tool. Until that lands, we can fix most issues by just pulling higher version of Newtonsoft.Json.

### DIFF
--- a/TestAssets/TestProjects/PortableAppWithTools/TestContext.cs
+++ b/TestAssets/TestProjects/PortableAppWithTools/TestContext.cs
@@ -1,4 +1,5 @@
 using Microsoft.EntityFrameworkCore;
+using Newtonsoft.Json;
 
 namespace ConsoleApplication
 {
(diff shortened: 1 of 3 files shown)


TestAssets/TestProjects/PortableAppWithTools/TestContext.cs
```diff
@@ -13,6 +14,8 @@ public class TestContext : DbContext
     {
         protected override void OnConfiguring(DbContextOptionsBuilder options)
         {
+            var o = new object();
+            JsonConvert.SerializeObject(o);
```
lajones: I would create an issue to track fixing this, then add a TODO comment here referencing that issue, then :shipit:

NTaylorMullen:
:shipit:

TestAssets/TestProjects/PortableAppWithTools/TestContext.cs
natemcmaster: Already tracking as a part of https://github.com/aspnet/EntityFramework/issues/3925

Issue #4285, opened 2016-01-12
InMemoryDatabase raises System.InvalidOperationException when you save TPH Model


katz:
```C#
Unhandled Exception: System.InvalidOperationException: The property 'Discriminator' on entity type 'SubProduct' has a temporary value while attempting to change the entity's state to 'Unchanged'. Either set a permanent value explicitly or ensure that the database is configured to generate values for this property.
   at Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges)
   at Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState entityState, Boolean acceptChanges)
   at Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntry.AcceptChanges()
   at Microsoft.Data.Entity.ChangeTracking.Internal.StateManager.AcceptAllChanges(IReadOnlyList`1 changedEntries)
   at Microsoft.Data.Entity.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.Data.Entity.DbContext.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.Data.Entity.DbContext.SaveChanges()
```

The below is a sample console application that raises the exception.
The sample app uses EF 7.0.0-rc1-final with .NET Framework 4.6.

```C#
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using Microsoft.Data.Entity;
using Microsoft.Data.Entity.Infrastructure;
using Microsoft.Data.Entity.Metadata;
using System.ComponentModel.DataAnnotations;

namespace EF7ConsoleApplication
{
    class Program
    {
        static void Main(string[] args)
        {
            var optionsBuilder = new DbContextOptionsBuilder<ApplicationDbContext>();
            // This is the magic line
            optionsBuilder.UseInMemoryDatabase();
            var db = new ApplicationDbContext(optionsBuilder.Options);

            var subProductList = new List<SubProduct>()
            {
                new SubProduct()
                {
                    someName = "test1",
                },
                new SubProduct()
                {
                    someName = "test2",
                },
            };

            db.AddRange(subProductList);
            // The SaveChanges() raises the exception.
            db.SaveChanges();
        }
    }

    public class BaseProduct
    {
[...]

PR #5090, opened 2016-04-15
Prevent discriminator columns from failing with in-memory database

ajcvickers:
Issue #4285

The default, provider-agnostic, value generator was setup to always generate temporary values. This was a carry over from when we required this so that entities could be added in any order, but that behavior was changed and so always generating temporary values is no longer needed.

The default value generator now doesn't generate temporary values, which may not be handled by the provider--such as is the case for in-memory. Where possible default value generators are still supplied that make use of GUIDs. One of these now gets used in this scenario.

This change doesn't really fix the problem that structural changes made to the model by a provider may not be handled correctly by other providers, but it fixes the issue at hand.


anpete:
:shipit:

PR #5098, opened 2016-04-18
Fix #3425 - Throw before saving temp values that aren't generated by the store

### LINKED ISSUE #3425: Throw before saving if temp values are present and won't be generated by the database
Split off from https://github.com/aspnet/EntityFramework/issues/3390
Currently we throw after SaveChanges if temp values are present, we should throw in advance if we know the database isn't going to generate something.

mikary:
cc @ajcvickers @AndriySvyryd


ajcvickers:
:shipit:

PR #5105, opened 2016-04-18
Improve configurability of benchmark database 

rowanmiller:
We used to just allow configuring the database instance name, and
assumed integrated security etc. Updating to allow the full connection
string to be specified, then we just append the database. This is
required because we are moving the benchmark CI to use a separate
database server.


divega:
:shipit:

PR #5115, opened 2016-04-19
Upgrade version of dotnet-test-xunit to RC2

natemcmaster:
Also removes unnecessary dependency on Ms.NETCore.Platforms.

cc @pakrym @smitpatel

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.Commands.FunctionalTests/project.json
+++ b/test/Microsoft.EntityFrameworkCore.Commands.FunctionalTests/project.json
@@ -21,7 +21,7 @@
           "version": "1.0.0-*",
           "type": "platform"
         },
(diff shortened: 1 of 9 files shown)


pakrym:
`Ms.NETCore.Platforms` is required for 'runable' projects that target `net451`

natemcmaster:
oh, ok. I'll reverse that part of this change

smitpatel:
LGTM

natemcmaster:
This enables the VS test runner. When the new VS tooling becomes public, we can consider again switching back to xproj.

pranavkm:
Fixed in https://github.com/aspnet/EntityFramework/commit/809139d6748c837a2ad8e1b0f0a363406725c6d7

Issue #4375, opened 2016-01-22
Problem with OriginalValue and INotifyPropertyChanging


mehrt:
I tried this on 7.0.0-rc2-16769. What happens is, when an entity implements INotifyPropertyChanging, it seems the OriginalValue always reflects the latest update to it's property, but not the real original value. I tried this with a plain entity, an entity implementing IPropertyChanged, and a third one also implementing IPropertyChanging, which is why the model is a bit lengthy. The first two seem fine, the third one behaves differently. Unless I made a trivial error I haven't spotted in the last two hours, it's probably a bug.
Model:

```c#
using Microsoft.Data.Entity;
using System.ComponentModel;
using System.Runtime.CompilerServices;

namespace ReproOriginalValues
{
  public abstract class Changed : INotifyPropertyChanged
  {
    public event PropertyChangedEventHandler PropertyChanged;
    protected void OnPropertyChanged([CallerMemberName]string propertyName = "")
    {
      var handler = PropertyChanged;
      if (handler != null) handler(this, new PropertyChangedEventArgs(propertyName));
    }
  }

  public abstract class Changing : Changed, INotifyPropertyChanging
  {
    public event PropertyChangingEventHandler PropertyChanging;
    protected void OnPropertyChanging([CallerMemberName]string propertyName = "")
    {
      var handler = PropertyChanging;
      if (handler != null) handler(this, new PropertyChangingEventArgs(propertyName));
    }
  }

  public class Blog1
  {
    public int Blog1Id { get; set; }
    public string Url { get; set; }
  }

  public class Blog2 : Changed
  {
    private int _blog2Id;
    public int Blog2Id
    {
      get { return _blog2Id; }
      set
      {
        if (_blog2Id == value) return;
        _blog2Id = value;
        OnPropertyChanged();
      }
    }
    private string _url;
    public string Url
    {
      get { return _url; }
      set
      {
        if (_url == value) return;
        _url = value;
        OnPropertyChanged();
      }
    }
  }

  public class Blog3 : Changing
  {
    private int _blog3Id;
    public int Blog3Id
    {
      get { return _blog3Id; }
      set
      {
        if (_blog3Id == value) return;
        OnPropertyChanging();
        _blog3Id = value;
        OnPropertyChanged();
      }
    }
    private string _url;
    public string Url
    {
      get { return _url; }
      set
      {
        if (_url == value) return;
        OnPropertyChanging();
        _url = value;
        OnPropertyChanged();
      }
    }
[...]

PR #5118, opened 2016-04-19
Throw when attempting to access original value when it is not being tracked.

ajcvickers:
Issue #4375.

For entities that implement INotifyPropertyChanging, the original value is not needed by changing tracking and hence we do not store it or allocate space to store it. Previously if you asked for the original value in such cases we would give you back the current value. We now throw saying that the value is not available and directing you to the flag that will make it available. It's a pretty blunt flag right now, but we can make it more fine-grained if enough people want this behavior.

In fixing this, it turns out that the update pipeline was using the original value for primary key where clauses. This started to throw because the original value was not available. This has been changed to use the current value instead. This was complicated because the logic for choosing original or current was implemented twice and both needed to match, so this has been changed to use a common implementation.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Update/ColumnModification.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Update/ColumnModification.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Threading;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Utilities;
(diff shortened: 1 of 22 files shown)


AndriySvyryd:
:shipit:

davidroth:
@ajcvickers Wouldn't it make sense to lazily track property values if _INotifyPropertyChanging_ is implemented? For example, for a given property _Foo_, the first time INPChanging is raised for _Foo_, the value gets tracked as it`s original value. If the property never changes, then there is no overhead of tracking the value. If it is changed, it is tracked and can be used.

Or am I missing something here?

ajcvickers:
@davidroth This is basically what we do for properties for which EF itself needs original values--foreign keys and concurrency tokens. Doing it for any other property where the app needs an original value but EF doesn't is something we want to enable as an option although it is a lower priority for now. (But I may do it anyway as part of another upcoming change.) The default behavior means EF isn't using memory to store original values that it will never need.

PR #5122, opened 2016-04-19
Query: Convert order by condition to case blocks

### LINKED ISSUE #5095: Error when Ordering by condition
I have the following query to order items based on future and past dates:

```
    context.Projects.OrderBy(x => x.Finish > DateTime.UtcNow).ThenBy(x => x.Finish);
```

I get the following error:

```
    System.AggregateException: One or more errors occurred. 
---> System.AggregateException: One or more errors occurred. 
---> System.AggregateException: One or more errors occurred. 
---> System.AggregateException: One or more errors occurred. 
---> System.AggregateException: One or more errors occurred. 
---> System.Data.SqlClient.SqlException: Incorrect syntax near '>'.
Invalid usage of the option NEXT in the FETCH statement.
```

I suppose this is a bug?

smitpatel:
Issue is Order By requires value type for ordering. Having condition in order by expression is incorrect sql.
Fix is convert comparison or logical expressions in case blocks when they appear in order by expression

Fixes #5095

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -769,6 +769,15 @@ namespace Microsoft.EntityFrameworkCore.Query
                         break;
                     }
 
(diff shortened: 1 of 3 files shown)


anpete:
:shipit:

Issue #5074, opened 2016-04-14
Be consistent and correct with metadata extensions


ajcvickers:
It had been decided in the past that we would:
* Keep metadata interfaces minimal with only “core”, non-computed concepts exposed
* Store non-core concepts as annotations
* Add extension methods to access non-core concepts and computed information
* Use caching when necessary to avoid expensive computation
* Expose caching through additional interfaces that could be probed for

This has resulted in the following issues:
* Our product code and tests only ever use our implementations, which means where an extension method does some probing it will always go down one path and not the other. We never test the other path, which may be completely broken. For example:

```C#
public static bool UseEagerSnapshots([NotNull] this IEntityType entityType)
{
    Check.NotNull(entityType, nameof(entityType));

    return (entityType as EntityType)?.UseEagerSnapshots ?? false;
}
```

If the implementation of IEntityType is not ours, then this code will always return false. This is almost always wrong and will cause change tracking to not work.

* We often don’t probe for interfaces but instead for our concrete types. See the example above again. This means that even if this code was correct and was somehow able to compute a value for UseEagerSnapshots, if that computation was slow (which it would be) then there would be no way for some other implementation of IEntityType to use a cached value and obtain acceptable perf.

I think we should choose a way to be consistent and fix the code appropriately. Some options:
* Go all-in YAGNI and assert that the only implementation of our metadata interfaces for the foreseeable future will be our own implementation. In extension methods that probe have the path we don’t use throw saying you can’t do that, but you can submit a pull request exposing the behavior as an interface such that you can then do it.
* Go all in with the extensiblity mechanism and make sure that we don’t probe for concrete types and that the paths taken with other implementations actually work by writing and maintaining tests for them all. This is quite a lot of work.
* Go half in and drop the additional interfaces, but at least add tests for all implementations or our extension methods. This would allow somebody to implement our metadata, although probably not in a performant way.

Personally, I think we should go with all-in YAGNI. The interfaces we have provide a good read-only API surface, but I think as of now we should just assume that nobody else is implementing our metadata model. If and when somebody decides to do so we can take a path to make that work.

AndriySvyryd:
Another point to consider is that soon we could have a read-only model implementation for perf. Then the mutable model will not need to implement most of the optional interfaces and could be used for tests without much additional cost.
In the long-term we could consider using the compiled model generator to customize what combination of interfaces we use for tests.

ajcvickers:
[...]

PR #5136, opened 2016-04-20
Be explicit that custom metadata interface implementations are not yet supported

ajcvickers:
See issue #5074


src/Microsoft.EntityFrameworkCore/Extensions/KeyExtensions.cs
```diff
@@ -22,18 +22,6 @@ public static class KeyExtensions
...
-            if (referencingForeignKeyMetadata != null)
-            {
-                return referencingForeignKeyMetadata.ReferencingForeignKeys;
-            }
-
-            return key.DeclaringEntityType.Model.GetEntityTypes()
-                .SelectMany(e => e.GetDeclaredForeignKeys())
-                .Where(fk => fk.PrincipalKey == key);
-        }
+            => Check.NotNull(key, nameof(key)).AsKey().ReferencingForeignKeys ?? Enumerable.Empty<IForeignKey>();
```
AndriySvyryd: Move the coalesce to .ReferencingForeignKeys

AndriySvyryd: All the FindReferencing* properties should be renamed to GetReferencing*
We decided that methods returning a collection that don't take parameters should be prefixed with Get. This and the ones for Property were missed.

src/Microsoft.EntityFrameworkCore/Metadata/Internal/MetadataExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    public static class MetadataExtensions
+    {
+        public static TConcrete AsConcreteType<TInterface, TConcrete>([NotNull] this TInterface @interface, [NotNull] string methodName)
+            where TConcrete : class
+        {
+            var concrete = @interface as TConcrete;
+            if (concrete == null)
+            {
+                throw new NotSupportedException(
+                    CoreStrings.CustomMetadata(methodName, typeof(TInterface).Name, @interface.GetType().Name));
```
AndriySvyryd: The exception message is specific for metadata interfaces. Either add a check so it is only thrown for metadata interfaces or rename the method to reflect this.

src/Microsoft.EntityFrameworkCore/Extensions/KeyExtensions.cs
ajcvickers: Will do.

src/Microsoft.EntityFrameworkCore/Metadata/Internal/MetadataExtensions.cs
ajcvickers: Will rename,

AndriySvyryd:
:shipit:

src/Microsoft.EntityFrameworkCore/Extensions/KeyExtensions.cs
AndriySvyryd: Same goes for FindContaining*

Issue #5043, opened 2016-04-12
Query: Navigations: Using EF.Property with reference nav omits null compensation


anpete:
In ComplexNavigationsQueryTests:

```c#
[ConditionalFact]
public virtual void Key_equality_when_sentinel_ef_property()
{
    using (var context = CreateContext())
    {
        var query = context.LevelOne
            .Where(l => EF.Property<int>(l.OneToOne_Optional_FK, "Id") == 0);

        var result = query.ToList();

        Assert.Equal(0, result.Count);  // Fails with 5 results
    }
}
```

maumar:
Fixed in ff8bc9baef0ec2d0bb146f3d12ea1bfbfb834035

PR #5137, opened 2016-04-21
Fix to #5043 - Query: Navigations: Using EF.Property with reference nav omits null compensation

maumar:
Problem was that navigation expansion logic was only applied for MemberExpressions, but never for MethodCallExpression. Fix is to abstract that logic to work for both MemberExpressions and MethodCallExpression that happen to be EF.Property.
Also did some minor bugfixes and optimizations found along the way:
- compensating for nullability differences in SqlTranslatingExpressionVisitor, when processing comparison,
- preventing EntityEqualityRewritingVisitor to short-circuit on non-equality operations in case of nested binary expressions.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
@@ -294,7 +294,7 @@ namespace Microsoft.EntityFrameworkCore.Query
                     .Select(includeResultOperator =>
                         {
                             var navigationPath
-                                = BindNavigationPathMemberExpression(
+                                = BindNavigationPathPropertyExpression(
                                     includeResultOperator.NavigationPropertyPath,
                                     (ps, _) =>
                                         {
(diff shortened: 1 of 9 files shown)


maumar:
@anpete

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -1060,39 +1060,44 @@ private static Expression ShiftMemberAccess(Expression targetExpression, Express
...
                     properties,
                     querySourceReferenceExpression.ReferencedQuerySource);
             }
 
             return default(TResult);
         }
 
-        private IEnumerable<IPropertyBase> IterateCompositeMemberExpression(
-            MemberExpression memberExpression, out QuerySourceReferenceExpression querySourceReferenceExpression)
+        private IEnumerable<IPropertyBase> IterateCompositePropertyExpression(
```
anpete: Not sure about this. Currently we have separate "bind" methods for members vs method call I.e. BindMethodCallExpression

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -177,12 +177,14 @@ public override void Where_query_composition_entity_equality()
         {
             base.Where_query_composition_entity_equality();
 
-            Assert.StartsWith(
+            Assert.Equal(
```
anpete: I'm not sure this is correct. Can you test when the subquery returns 2 items/zero items?

maumar:
@anpete  new version up

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -1074,39 +1074,69 @@ private static Expression ShiftMemberAccess(Expression targetExpression, Express
...
                 && (querySource == null
                     || querySource == querySourceReferenceExpression.ReferencedQuerySource))
             {
-                return memberBinder(
+                return propertyBinder(
                     properties,
                     querySourceReferenceExpression.ReferencedQuerySource);
             }
 
+            if (properties.Count() > 0)
```
anpete: Multiple enumeration of properties.

anpete:
:shipit:

PR #5142, opened 2016-04-21
Refactor design-time tools

natemcmaster:
Split design-time tools into three packages so that users need only install "Microsoft.EntityFrameworkCore.Tools" and nuget will restore the tools they need for the platform they target.

The packages:

 - **.Tools**
     - A pseudo-metapackage. This contains shims to call into shared
       components.
     - **lib/net451/Microsoft.EntityFrameworkCore.Tools.dll** - Contains
       PowerShell cmdlets and APIs that bridge it with common components.
     - **lib/netcoreapp1.0/dotnet-ef.dll** - a light shim for initializing
       commands on DotNet project. This executes from "tools".
     - **lib/netcore50/Microsoft.EntityFrameworkCore.Tools.dll** - same as
       net451. For compatibility with UWP only. This is never actually
       executed.
     - **build/netcore50/*.target** - swaps netcore50 to net451 at build time
       on UWP projects
     - **tools/*** - the powershell cmdlets
 - **.Tools.Cli**
     - A program to launch commands from dotnet-cli. These execute as a
       "dependency".
     - **lib/netstandard1.5/*.dll** - Contains the entry point for executing
       commands on a DotNet project model that supports .NET Core
     - **lib/net451/*.exe** - Contains the entry point for executing commands
       on a DotNet project modle that supports desktop .NET only
 - **.Tools.Core**
     - contains shared components. This contains the implementation
       details of executing commands.

Resolves https://github.com/aspnet/EntityFramework/issues/4577
Resolves https://github.com/aspnet/EntityFramework/issues/5065
Partial fix for https://github.com/aspnet/EntityFramework/issues/3925

### Functional issues

- [x] Can't use dotnet-ef in a class library or standalone apps. This dies with "Could not resolve coreclr path". https://github.com/dotnet/cli/issues/2645
- [x] There are issues loading assemblies when startup project and current project are different. Haven't identified the source of the issue.
- [x] Cannot install into a csproj. Blocked on https://github.com/NuGet/Home/issues/2635.
- [x] The framework/configuration on the dispatcher (dotnet-ef) should flow through to the "inside-man".

### Cleanup

- [x] Put all strings into resx files
- [x] Remove workaround that runs `dotnet restore --infer-runtimes` on the tools projects. Without this, dotnet-pack fails, complaining that no runtimes were specified.
- [ ] Merge https://github.com/aspnet/Coherence/pull/79
 
cc @divega @NTaylorMullen @bricelam


src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -0,0 +1,218 @@
...
+using Microsoft.DotNet.ProjectModel;
+using Microsoft.Extensions.PlatformAbstractions;
+using NuGet.Frameworks;
+using Microsoft.EntityFrameworkCore.Design;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Migrations.Design;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore.Scaffolding.Internal;
+using Microsoft.EntityFrameworkCore.Design.Internal;
```
AndriySvyryd: Sort usings

AndriySvyryd:
🐑 🇮🇹

src/Microsoft.EntityFrameworkCore.Tools.Cli/ConsoleCommandLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        protected override void WriteDebug(string message)
+            => Reporter.Verbose.WriteLine(message.Bold().Black());
+
+        protected override void WriteError(string message)
+            => Reporter.Error.WriteLine(message.Bold().Red());
+
+        protected override void WriteInformation(string message)
+            => Reporter.Error.WriteLine(message);
```
NTaylorMullen: Error for writeinformation?

src/Microsoft.EntityFrameworkCore.Tools.Cli/ExecuteCommand.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+
+using JetBrains.Annotations;
+using Microsoft.Extensions.CommandLineUtils;
+using Microsoft.EntityFrameworkCore.Design;
+
+namespace Microsoft.EntityFrameworkCore.Tools.Cli
+{
+    public class ExecuteCommand
```
NTaylorMullen: static

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -0,0 +1,218 @@
...
+            string startupBuildBaseDir = null;
+            string startupOutputDir = null;
+
+            var startupProjectContext = GetCompatibleStartupProjectContext(startupProject);
+
+            var startupFramework = startupProjectContext.TargetFramework;
+
+            if (buildStartup)
+            {
+                Reporter.Verbose.WriteLine("Build started...".Bold().Black());
```
NTaylorMullen: Note: We're probably gonna have them remove these string extensions from the cli.utils package. For now is fine, but an FYI

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -0,0 +1,218 @@
...
+                        startupBuildBaseDir,
+                        startupOutputDir)
+                    .ForwardStdOut()
+                    .ForwardStdErr()
+                    .Execute()
+                    .ExitCode;
+
+                if (buildExitCode != 0)
+                {
+                    throw new OperationException("Build failed.");
```
NTaylorMullen: If this is going to be consumed by VS/scaffolding may want to provide some extra information here so when those tools consume ef they provide verbose enough errors when things go wrong.

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -0,0 +1,218 @@
...
+    {
+        private const string DataDirEnvName = "ADONET_DATA_DIR";
+
+        private readonly LazyRef<MigrationsOperations> _migrationsOperations;
+        private readonly LazyRef<DatabaseOperations> _databaseOperations;
+        private readonly LazyRef<DbContextOperations> _contextOperations;
+
+        public OperationExecutor([CanBeNull] string startupProject, [CanBeNull] string environment)
+        {
+            var project = Path.Combine(Directory.GetCurrentDirectory(), Project.FileName);
```
NTaylorMullen: move this into the if statement below

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -0,0 +1,218 @@
...
+            var buildStartup = true;
+            if (startupProject == null)
+            {
+                startupProject = project;
+                buildStartup = false;
+            }
+            // TODO flow through from dispatch
```
NTaylorMullen: Wont this need to be done in the dispatch layer? Reason: Users project needs to be built in order to dispatch in the first place.

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -0,0 +1,218 @@
...
+            var runtimeOutputPath = startupProjectContext.GetOutputPaths(startupConfiguration)?.RuntimeOutputPath;
+            if (!string.IsNullOrEmpty(runtimeOutputPath))
+            {
+                Environment.SetEnvironmentVariable(DataDirEnvName, runtimeOutputPath);
+#if NET451
+                AppDomain.CurrentDomain.SetData("DataDirectory", runtimeOutputPath);
+#endif
+            }
+
+            var projectFile = ProjectReader.GetProject(project);
```
NTaylorMullen: Why does this and the startupProject bits potentially fork?

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -0,0 +1,218 @@
...
+            var startupAssemblyName = startupProjectContext.ProjectFile.GetCompilerOptions(startupFramework, startupConfiguration).OutputName;
+            var assemblyName = projectFile.GetCompilerOptions(projectFramework, projectConfiguration).OutputName;
+            var projectDir = projectFile.ProjectDirectory;
+            var startupProjectDir = startupProjectContext.ProjectFile.ProjectDirectory;
+            // TODO should this match assembly name?
+            var rootNamespace = projectFile.Name;
+
+            var assemblyLoader = new AssemblyLoader(Assembly.Load);
```
NTaylorMullen: Why not just utilize Assembly.Load?

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -0,0 +1,218 @@
...
+                cancellationToken);
+
+        private ProjectContext GetCompatibleStartupProjectContext(string projectPath)
+        {
+            var projectFile = ProjectReader.GetProject(projectPath);
+            var frameworks = projectFile.GetTargetFrameworks().Select(f => f.FrameworkName);
+
+#if NET451
+            var currentFramework = FrameworkConstants.CommonFrameworks.Net451;
+            var framework = frameworks.FirstOrDefault(f => f.IsDesktop())
```
NTaylorMullen: Based on [this](https://github.com/aspnet/EntityFramework/pull/5142/files#r60815303) having frameworks here may not be interesting.

src/Microsoft.EntityFrameworkCore.Tools.Cli/ExecuteCommand.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public static EfCommandLineApplication Configure()
```
NTaylorMullen: `Create` would probably make more sense.

src/Microsoft.EntityFrameworkCore.Tools.Cli/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+            catch (Exception ex)
+            {
+                if (ex is TargetInvocationException)
+                {
+                    ex = ex.InnerException;
+                }
+
+                if (!(ex is OperationException))
+                {
+                    Reporter.Error.WriteLine(ex.ToString());
```
NTaylorMullen: This'll result in some duplicated output with the reporter writeline below

src/Microsoft.EntityFrameworkCore.Tools/Program.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    Reporter.Error.WriteLine(ex.ToString());
+                }
+
+                Reporter.Error.WriteLine(ex.Message.Bold().Red());
+                return 1;
+            }
+        }
+
+        private static void HandleVerboseOption(ref string[] args)
```
NTaylorMullen: Why not rely on dotnets verbose options/Reporter.Verbose?

src/Microsoft.EntityFrameworkCore.Tools/DispatchCommand.cs
```diff
@@ -0,0 +1,148 @@
...
+                        .ForwardStdOut()
+                        .Execute()
+                        .ExitCode;
+                    if (buildExitCode != 0)
+                    {
+                        throw new OperationException("Build failed.");
+                    }
+                }
+
+                var projectContext = new ProjectContextBuilder()
```
NTaylorMullen: Don't need a projectContext. See https://github.com/aspnet/RazorTooling/pull/65/files#diff-33c97b345caf0ab940bd32a46634dde0R73

NTaylorMullen:
:watch:

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
natemcmaster: Needs to be done in both places. Example:
```
AspNetWebApp/
    Startup.cs
Library/
    TestContext.cs
```
If I run `C:\Library> dotnet ef migrations add Initial --startup-project ../AspNetWebApp`, I need to go build "AspNetWebApp"

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
natemcmaster: Project is not necessarily the same as startupProject. I'm going to clean this code up to make it a little more clear.

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
natemcmaster: :cry: but I like these extension methods...

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
NTaylorMullen: Why build them both separately though? Can you not make the assumption that the startup project references the project that will have the migrations done?

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
natemcmaster: Build errors should be show on stderr, but I'll add the name of the project we were trying to build as well.

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
natemcmaster: Actually, this commit didn't include a change I'm about to push. In some cases, I need AssemblyLoadContext. In others Assembly.Load is fine. AssemblyLoader is just a wrapper to make this abstraction easier.

src/Microsoft.EntityFrameworkCore.Tools.Cli/Program.cs
natemcmaster: This is by design. We only log the stack trace if we get an unexpected error type. OperationException is an exception we throw intentionally to short-circuit the command execution

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
natemcmaster: In this scenario, AspNetWebApp references Library. The dispatcher only builds Library. We could make dispatcher smarter and have it build AspNetWebApp too, but this mixes dispatcher with inside-man too much IMO

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
natemcmaster: Yeah, I need to cleanup the framework logic...will add in a follow-up commit.

src/Microsoft.EntityFrameworkCore.Tools/Program.cs
natemcmaster: We'd like both of these to work: `dotnet ef -v` and `dotnet -v ef`. Does dotnet utils have an API for this now?

NTaylorMullen: Not that I know of

natemcmaster:
I've added a new commit to this PR. It includes some response to feedback, cleanup, and fixed up some bugs.

https://github.com/aspnet/EntityFramework/pull/5142/commits/27f6fa984e2936e6b0ba896322a63cf65fe75e46

src/Microsoft.EntityFrameworkCore.Tools.Cli/ExecuteCommand.cs
```diff
@@ -0,0 +1,102 @@
...
+                DispatchVersionOptionTemplate, GetVersion(),
+                noBuild ? NoBuildOptionTemplate : string.Empty,
+                verbose ? VerboseOptionTemplate : string.Empty
+            };
+
+        public static CommandLineApplication Create()
+        {
+            var app = new CommandLineApplication()
+            {
+                Name = "Microsoft.EntityFrameworkCore.Tools.Cli",
```
NTaylorMullen: If this is tied to this assembly you should use a typeof to ensure it stays consistent without needing to manually update.

test/Microsoft.EntityFrameworkCore.Tools.FunctionalTests/TestUtilities/AppDomainOperationExecutor.cs
```diff
@@ -118,9 +118,9 @@ private object InvokeOperationImpl(string operation, Hashtable arguments, bool i
 
             if (resultHandler.ErrorType != null)
             {
-                throw new OperationErrorException(
-                    resultHandler.ErrorType,
-                    resultHandler.ErrorStackTrace,
+                throw new OperationException(
+                    //resultHandler.ErrorType,
+                    //resultHandler.ErrorStackTrace,
```
lajones: TODO or remove completely?

lajones:
:shipit:

src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
```diff
@@ -31,92 +33,115 @@ public class OperationExecutor
...
-                Reporter.Verbose.WriteLine("Build started...".Bold().Black());
+                Reporter.Verbose.WriteLine(ToolsCliStrings.LogBuildFailed.Bold().Black());
 
-                var buildExitCode = BuildCommandFactory.Create(startupProject,
+                var buildExitCode = BuildCommandFactory.Create(startupProjectPath,
                         startupConfiguration,
                         startupFramework,
-                        startupBuildBaseDir,
-                        startupOutputDir)
+                        buildBasePath: null,
```
NTaylorMullen: Might be good to allow a user to specify some of these so they don't get build output spewed in their current directory.

NTaylorMullen:
Took a look at the CLI side of things and seemed alright. Lots here though and I'm not 100% fluent in what the EF tooling is responsible for.

:shipit: Feel free to bug me if you want anything specific looked over in detail.

PR #5146, opened 2016-04-21
Metadata: Allow creating shadow navigation on shadow entity types

### LINKED ISSUE #2140: Metadata: Allow unbound navigation properties
For example:

```C#
mb.Entity<Post>().Reference<Blog>("Blog").InverseCollection("Posts");
```

We could then store this in model snapshot and use it to help detect foreign key column renames.

It could also make configuring foreign keys without any navigation properties a little less awkward.

smitpatel:
Fixes #2140

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/Navigation.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/Navigation.cs
@@ -36,6 +36,15 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
             ForeignKey = foreignKey;
         }
 
+        public Navigation([NotNull] string navigationName, [NotNull] ForeignKey foreignKey)
+        {
+            Check.NotEmpty(navigationName, nameof(navigationName));
+            Check.NotNull(foreignKey, nameof(foreignKey));
+
+            Name = navigationName;
+            ForeignKey = foreignKey;
+        }
+
         public virtual string Name { get; }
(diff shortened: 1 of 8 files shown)


AndriySvyryd:
:shipit:

test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/ModelBuilderNonGenericStringTest.cs
```diff
@@ -17,16 +18,139 @@ public class ModelBuilderNonGenericStringTest : ModelBuilderNonGenericTest
     {
         public class NonGenericOneToManyType : OneToManyTestBase
```
AndriySvyryd: Could you add "String" to the names of these nested classes?

src/Microsoft.EntityFrameworkCore/Metadata/Internal/Navigation.cs
```diff
@@ -96,12 +105,8 @@ public virtual EntityType DeclaringEntityType
             var sourceClrType = sourceType.ClrType;
             if (sourceClrType == null)
             {
-                if (shouldThrow)
-                {
-                    throw new InvalidOperationException(
-                        CoreStrings.NavigationOnShadowEntity(navigationPropertyName, sourceType.DisplayName()));
```
AndriySvyryd: Also remove the resource

PR #5153, opened 2016-04-22
Various bechmark improvements

rowanmiller:
Increasing the work done in a number of the tests in order to get better numbers from our benchmark CI. See the individual commits for more specific details.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/InitializationTests.cs
+++ b/test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/InitializationTests.cs
@@ -16,10 +16,10 @@ namespace Microsoft.EntityFrameworkCore.Microbenchmarks.EF6
     public class InitializationTests : IClassFixture<AdventureWorksFixture>
     {
         [Benchmark]
-#if NETCOREAPP1_0
+        [BenchmarkVariation("Warm (10000 instances)", false, 10000)]
+#if NET451
         [BenchmarkVariation("Cold (1 instance)", true, 1)]
 #endif
-        [BenchmarkVariation("Warm (1000 instances)", false, 1000)]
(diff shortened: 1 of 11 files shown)


test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/InitializationTests.cs
```diff
@@ -16,32 +16,32 @@ namespace Microsoft.EntityFrameworkCore.Microbenchmarks.EF6
     public class InitializationTests : IClassFixture<AdventureWorksFixture>
     {
         [Benchmark]
-#if NETCOREAPP1_0
+        [BenchmarkVariation("Warm (10000 instances)", false, 10000)]
+#if NET451
         [BenchmarkVariation("Cold (1 instance)", true, 1)]
```
divega: I am curious about the reason for all the changes you made that are similar to this.

divega:
:shipit:

test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/InitializationTests.cs
rowanmiller: This stuff got refactored so much with all the TFM changes that it just ended up being wrong. Somewhere along the way the EF Core cold start tests were removed altogether and the EF6 ones ended up in a state where they never actually run... they were compiled out on NET451 (where they could run), and just skipped on NetCore (because they can't run).

rowanmiller: Actually, correction, the Core ones would have failed... but currently the benchmarks are just running against Full .NET because the Core runs were super flakey while everything was moving to CLI.

PR #5157, opened 2016-04-22
[SMALL] Async variations of query tests

rowanmiller:


test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/Query/SimpleQueryTests.cs
```diff
@@ -136,8 +181,9 @@ public void SkipTake(IMetricCollector collector, bool tracking, int queriesPerIt
         }
 
         [Benchmark]
-        [BenchmarkVariation("Default (10 queries)", 10)]
-        public void GroupBy(IMetricCollector collector, int queriesPerIteration)
+        [BenchmarkVariation("Default (10 queries)", false, 10)]
```
divega: From the perspective of the APIs offered by EF, I am not sure we should call the sync ones the default. Perhaps there is a practical reason to do it though, e.g. to avoid having to rename existing data sets?

Anyway, the comment likely applies to the `SaveChanges()` PR as well, but I didn't noticed it when I was reviewing that one.

divega:
:shipit:

test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/Query/SimpleQueryTests.cs
rowanmiller: Good point, "Default" is the name the variation has if nothing is specified, so I just used this name to match the existing data. I'll change all these to have a " - Sync" and "- Async" variation.

Issue #3316, opened 2015-10-03
Grouping with Joins using anonymous objects


gustyn:
I am using dnx Beta7 as well as "EntityFramework.SqlServer": "7.0.0-beta7",

Given the following query which is a query I have run in various forms or other in EF6 and earlier:
```c#
    var query = (from blog in Model.Blogs
        join comment in Model.Comments on blog.Id equals comment.BlogId into commentJoin
	from comentLeft in commentJoin.DefaultIfEmpty()
	group new {blog.Id, blog.Body, blog.Summary, blog.Synopsis, blog.Title} by new { blog.Id, 
       blog.Body, blog.Summary, blog.Synopsis, blog.Title}
					//into g
					//			   select new  
					//			   {
					//				   Id = g.Key.Id,
					//				   Summary = g.Key.Summary,
					//				   Synopsis = g.Key.Synopsis,
					//				   Title = g.Key.Title,
					//				   Body = g.Key.Body,
					//				   //Comments = g.Select(x => new { Author = 
                                        //                          x.Comment.AuthorId.ToString(), Text = x.Comment.Text })
					//			   }
					).FirstOrDefault(x=>x.Key.Id == id);
```

I keep getting the following exception:
```
[...]

PR #5168, opened 2016-04-25
Add test for #3316 - Left outer join group by anonymous type

mikary:
Test coverage for a scenario that used to throw in beta/rc1
@maumar @divega


maumar:
:shipit:

Issue #3208, opened 2015-09-23
Query does not apply null semantics when using .Equals with nullable parameters


mikary:
[Fact]
        public virtual void Where_equals_on_null_nullable_int_types()
        {
            int? nullableIntPrm = null;

            AssertQuery<Employee>(
                es => es.Where(e => nullableIntPrm.Equals(e.ReportsTo)),
                entryCount: 1);

            AssertQuery<Employee>(
                es => es.Where(e => e.ReportsTo.Equals(nullableIntPrm)),
                entryCount: 1);
        }

csunali:
Could that possible relate to https://github.com/AutoMapper/AutoMapper/issues/902 as well?

There seems to be an exception whenever there is a "Select" statement as `HasChildren = item.Children.Count > 0` and the Children are bound to the parent entity using a `long?` field:

```
An exception of type 'System.InvalidOperationException' occurred in mscorlib.dll but was not handled in user code

Additional information: The binary operator Equal is not defined for the types 'System.Nullable`1[System.Int64]' and 'System.Int64'.
```

The entities are as follows:

```
    public class Item
    {
        public long Id { get; set; }
        public long? ParentId { get; set; }
        public string Name { get; set; }
        public bool HasChildren { get; set; }
        public int SortIndex { get; set; }
    }

    public class ItemModel
    {
        public long Id { get; set; }
        public long? ParentId { get; set; }

        public string Name { get; set; }
        public int SortIndex { get; set; }

        public virtual ItemModel Parent { get; set; }
        public virtual SiteModel Site { get; set; }
        public virtual ItemTypeModel ItemType { get; set; }
        public virtual ICollection<ItemModel> Children { get; } = new List<ItemModel>();
    }
```

And a simple code that breaks is:

```
            IQueryable<ItemModel> query = new ApplicationDbContext().Items;
            query = query.Where(i => i.ParentId == parentId);
            List<Item> items = (from i in query.OrderBy(i => i.SortIndex)
                                select new Item
                                {
                                    Id = i.Id,
                                    ParentId = i.ParentId,
                                    Name = i.Name,
                                    SortIndex = i.SortIndex,
                                    HasChildren = i.Children.Count > 0
                                }).ToList();
```

Please consider the fact that if I remove the `HasChildren = i.Children.Count > 0` line, everything works.

csunali:
[...]

PR #5173, opened 2016-04-25
Enable test for #3208 - Query null semantics on .Equals with nullable params

mikary:
Test case is now passing and generating expected SQL
@maumar


maumar:
:shipit:

PR #5174, opened 2016-04-25
Remove use of IApplicationEnvironment

pranavkm:
Hosting is no longer going to add these services to the container and we'll be removing the interfaces entirely (only keeping the static helpers)


pranavkm:
cc @divega - who'd be the right person to review this?
cc @davidfowl

divega:
This affects commands so assigning to @natemcmaster.

natemcmaster:
> we'll be removing the interfaces entirely

@pranavkm Can you point to a PR or issue in Hosting so we know when this change has been taken in to Hosting?

pranavkm:
Hosting work item: https://github.com/aspnet/Hosting/pull/723

natemcmaster:
@pranavkm  this is going into RC2?

pranavkm:
Yup.

src/Microsoft.EntityFrameworkCore.Commands/Design/Internal/ApplicationEnvironment.cs
```diff
@@ -1,28 +0,0 @@
...
-
-namespace Microsoft.EntityFrameworkCore.Design.Internal
-{
-    public class ApplicationEnvironment : IApplicationEnvironment
-    {
-        public ApplicationEnvironment([NotNull] IApplicationEnvironment applicationEnvironment)
-        {
-            Check.NotNull(applicationEnvironment, nameof(applicationEnvironment));
-
-            ApplicationBasePath = applicationEnvironment.ApplicationBasePath;
```
natemcmaster: What is going to replace this?

pranavkm: The only reference to this was in `StartupInvoker` which was removed. There's already code that sets up `HostingEnvironment` with the right `ContentRoot`

natemcmaster:
This is going to be problematic, unless the static helpers are settable. The path of `DefaultApplicationEnvironment.ApplicationBasePath` at design time is the nuget cache. Anyone that ues Configurate.SetBasePath to ApplicationBasePath is not going to be able to resolve config files at design-time.

natemcmaster:
:shipit: 

Note to self: when this is in, we need to manually test and verify this doesn't break `--startup-project` on commands.

davidfowl:
Looks good :shipit:

PR #5180, opened 2016-04-26
Query: Async: Reduce allocations during async query.

anpete:
Replaced most commonly used IX-Async operators with async methods (await), which are more efficient.

In one async query microbenchmark we see allocations go from 1,893,939,770 bytes/34,771,015 objects to 312,770,445 bytes/5,421,289 objects.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
@@ -30,11 +30,12 @@ namespace Microsoft.EntityFrameworkCore.Query
             QueryContext queryContext,
             ShaperCommandContext shaperCommandContext,
             IShaper<T> shaper)
-            => new AsyncQueryingEnumerable(
-                (RelationalQueryContext)queryContext,
-                shaperCommandContext,
-                queryIndex: null)
-                .Select(vb => shaper.Shape(queryContext, vb)); // TODO: Pass shaper to underlying enumerable
+            => AsyncLinqOperatorProvider.
+                _Select(new AsyncQueryingEnumerable(
+                    (RelationalQueryContext)queryContext,
+                    shaperCommandContext,
+                    queryIndex: null),
+                    vb => shaper.Shape(queryContext, vb)); // TODO: Pass shaper to underlying enumerable
 
         public virtual MethodInfo DefaultIfEmptyShapedQueryMethod => _defaultIfEmptyShapedQueryMethodInfo;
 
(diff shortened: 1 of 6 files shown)


anpete:
@divega @maumar

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -169,7 +172,7 @@ public DefaultIfEmptyAsyncEnumerator(IAsyncEnumerator<ValueBuffer> enumerator)
             Func<TSource, TElement> elementSelector)
             => new GroupByAsyncEnumerable<TSource, TKey, TElement>(source, keySelector, elementSelector);
 
-        private class GroupByAsyncEnumerable<TSource, TKey, TElement> : IAsyncEnumerable<IGrouping<TKey, TElement>>
+        private sealed class GroupByAsyncEnumerable<TSource, TKey, TElement> : IAsyncEnumerable<IGrouping<TKey, TElement>>
```
divega: Just to keep my understanding up to date: didn't we originally create these classes (`GroupByAsyncEnumerable`, `GroupByAsyncEnumerator`, etc.) also as a workaround for some limitations with the Ix-Async ones?

anpete: No, these are the relational streaming versions.

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
```diff
@@ -373,11 +375,47 @@ public IQueryProvider Provider
...
+                IAsyncEnumerable<TSource> source, Func<TSource, TResult> selector)
+            {
+                _source = source;
+                _selector = selector;
+            }
+
+            public IAsyncEnumerator<TResult> GetEnumerator()
+                => new SelectAsyncEnumerator(_source.GetEnumerator(), _selector);
+
+            private sealed class SelectAsyncEnumerator : IAsyncEnumerator<TResult>
```
divega: Would the trick of making the enumerator a struct like `List<T>` work for us here? I understand it probably wouldn't have a dramatic impact.

divega:
:shipit: 

New operator implementation look super clean! And numbers look great! Let's try to get this in for RC2. A few questions:
1. Are there more Ix-Async operators we would replace if we had a bit more time?
2. How much value are we getting out of the Ix-Async dependency after this besides the common `IAsyncEnumerable<T>` interface?
3. Is the main reason for the _PascalCase in `AsyncLinqOperatorProvider` method to avoid confusion with the Ix-Async counterparts?

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
anpete: Not as it stands because we use interfaces throughout.

divega: Got it. I forget this often.

anpete:
@divega 

1) All of them potentially. Next up would be Any, All, Last.
2) The majority of ops are still IX-Async. 
3) Yes, and the fact that we have AsyncLinqOperatorProvider.Select, which is a property that returns a MethodInfo - Usually Select is dynamically bound this way. Going forward, we could split our actual impls out of this class.

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
divega: Got it. I think it would be nice if that was clear in the name, but NABD :smile:

PR #5183, opened 2016-04-26
Switch AddDbContext to use .Add instead of .TryAdd when adding DbContextOptions

natemcmaster:
Fix https://github.com/aspnet/EntityFramework/issues/5156

~~This brings https://github.com/aspnet/EntityFramework/pull/5017 into release branch~~ 

cc @divega @ajcvickers

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools.Core/Design/DbContextOperations.cs
+++ b/src/Microsoft.EntityFrameworkCore.Tools.Core/Design/DbContextOperations.cs
@@ -104,7 +104,7 @@ namespace Microsoft.EntityFrameworkCore.Design
 
             // Look for DbContext classes registered in the service provider
             var registeredContexts = _runtimeServices.GetServices<DbContextOptions>()
-                .Select(o => o.GetType().GenericTypeArguments[0]);
+                .Select(o => o.ContextType);
             foreach (var context in registeredContexts.Where(c => !contexts.ContainsKey(c)))
             {
                 contexts.Add(
--- a/src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 .AddLogging();
 
             serviceCollection.TryAddSingleton(p => DbContextOptionsFactory<TContext>(p, optionsAction));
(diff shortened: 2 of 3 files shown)


natemcmaster:
Fixes https://github.com/aspnet/EntityFramework/issues/4994 too

ajcvickers:
@natemcmaster Can you explain why it requests bringing #5017 in? As I understand it, all this change will do in the normal case is change which of the two generic DbContextOptions is returned when the non-generic one is resolved. It doesn't make #5017 any worse or better. 

We could decide we really need this fix in RC2 anyway, but I don't think that it is related to this change.

divega:
FWIW we talked in triage today and decided to bring #5017 in. I agree the two changes are somewhat orthogonal, but #5017 makes the whole experience less confusing and the code in this PR a bit cleaner.

ajcvickers:
@natemcmaster @rowanmiller @divega I think a better process here would be to make this change as a single PR that we take through ask mode, and then reference the existing PRs for the other changes and bring those through separately. This change conflates several things into one PR, some of which have already been done in other PRs in dev. This doesn't seem to be the right way to go about this.

ajcvickers:
@divega Not sure how it makes the code in this PR cleaner. It should be a one-line change--TryAdd to Add. A couple of tests might break because of the last wins/first wins implication, but it shouldn't require any other product changes.

divega:
@ajcvickers I was thinking because of `ContextType` which you added. Not sure it makes it cleaner in any other way.

ajcvickers:
@divega How is ContextType involved in the change from TryAdd to Add?

divega:
It helps removing a couple of reflection calls that Migrations was using to detect the DbContext types. That is all. They are still orthogonal.

BTW, I agree with you that from the process perspective we can take the two PRs separately.

ajcvickers:
@divega Exactly. There are several things being conflated here, only one of which is described in the description of the PR.

It feels wrong to me when in ask mode to manually merge bits and pieces from several existing commits into one PR, then add some new stuff, and give it a title that doesn't reflect what is really happening at all. Seems like sloppy process. But I will step back now and let it happen this way if that's how you guys have discussed it should happen.

divega:
> But I will step back now and let it happen this way if that's how you guys have discussed it should happen.

We didn't discuss this explicitly. I think this is very good feedback, so I have asked @natemcmaster to send new PRs.

natemcmaster:
Updated. This PR to isolate the change. The other PR is https://github.com/aspnet/EntityFramework/pull/5186

ajcvickers:
:shipit:

PR #5193, opened 2016-04-27
Query: Funcletize method calls

### LINKED ISSUE #5035: DateTime.Parse is evaluated on client
In a SQL Server 2014 table with a smalldatetime column, any sort of comparison on said column returns this error:
`SqlException: Conversion failed when converting character string to smalldatetime data type.`

The setup is basic:
```
class Entity {
    [Column(TypeName="smalldatetime")]
    public DateTime Date { get; set; }
}
...
static void main(string[] args) {
    dbContext.Entity.Where(x => x.Date > DateTime.Parse("1/1/2015")).ToList(); // crashes
}
```

I noticed via Profiler that smalldatetime parameters are done the same way as datetime2, which has a precision of up to 7 decimal points of a second. The code for this is _DateTimeFormatConst_ in **RelationalSqlGenerationHelper**. However, in T-SQL a smalldatetime can only be parsed with a precision of one second (rounded to nearest minute as the type can support) So '2015-01-01T00:00:00.0000000' wouldn't work but '2015-01-01T00:00:00' would.

As I'm sure others still use this type for whatever reason that may be needed, can there be an additional format akin to something like `yyyy-MM-dd HH:mm:ss` and an overload to _GenerateLiteralValue_ that supports smalldatetime?

smitpatel:
fixes #5035 
Funcletize method calls in query.
`Guid.NextGuid()` & `Random().Next()` are not funcletized

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
@@ -758,7 +758,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
             using (var context = CreateContext())
             {
-                Assert.Throws<NotImplementedException>(() =>
+                Assert.Throws<InvalidOperationException>(() =>
                     context.Set<Customer>()
                         .Where(c => c.City == city.Throw().InstanceFieldValue)
                         .ToList());
@@ -3496,13 +3496,13 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         [ConditionalFact]
         public virtual void OrderBy_Count_with_predicate_client_eval_mixed()
         {
-            AssertQuery<Order>(os => os.OrderBy(o => o.OrderID).Count(o => ClientEvalPredicateStateless()));
+            AssertQuery<Order>(os => os.OrderBy(o => o.OrderID).Count(o => ClientEvalPredicate(o)));
         }
 
         [ConditionalFact]
         public virtual void OrderBy_Where_Count_with_predicate_client_eval()
         {
-            AssertQuery<Order>(os => os.OrderBy(o => ClientEvalSelectorStateless()).Where(o => ClientEvalPredicateStateless()).Count(o => ClientEvalPredicate(o)));
+            AssertQuery<Order>(os => os.OrderBy(o => ClientEvalSelectorStateless()).Where(o => ClientEvalPredicate(o)).Count(o => ClientEvalPredicate(o)));
         }
 
         [ConditionalFact]
@@ -3511,7 +3511,8 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             AssertQuery<Order>(os => os.OrderBy(o => o.OrderID).Where(o => ClientEvalPredicate(o)).Count(o => o.CustomerID != "ALFKI"));
         }
 
-        [ConditionalFact]
+        //TODO: The function translated to SQL and due to being integer represent the column number
+        //[ConditionalFact]
         public virtual void OrderBy_client_Take()
         {
             AssertQuery<Employee>(es => es.OrderBy(o => ClientEvalSelectorStateless()).Take(10), entryCount: 9);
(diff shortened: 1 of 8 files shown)


smitpatel:
@anpete @divega

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -3496,13 +3496,13 @@ public virtual void OrderBy_Count_with_predicate_client_eval()
         [ConditionalFact]
         public virtual void OrderBy_Count_with_predicate_client_eval_mixed()
         {
-            AssertQuery<Order>(os => os.OrderBy(o => o.OrderID).Count(o => ClientEvalPredicateStateless()));
+            AssertQuery<Order>(os => os.OrderBy(o => o.OrderID).Count(o => ClientEvalPredicate(o)));
```
anpete: Why are these changing?

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -3511,7 +3511,8 @@ public virtual void OrderBy_Where_Count_with_predicate_client_eval_mixed()
             AssertQuery<Order>(os => os.OrderBy(o => o.OrderID).Where(o => ClientEvalPredicate(o)).Count(o => o.CustomerID != "ALFKI"));
         }
 
-        [ConditionalFact]
+        //TODO: The function translated to SQL and due to being integer represent the column number
+        //[ConditionalFact]
```
anpete: I have a fix for this in my impending PR :smile:

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/AsyncQuerySqlServerTest.cs
```diff
@@ -46,8 +46,9 @@ public class AsyncQuerySqlServerTest : AsyncQueryTestBase<NorthwindQuerySqlServe
         public override async Task String_Contains_MethodCall()
         {
             await AssertQuery<Customer>(
-                cs => cs.Where(c => c.ContactName.Contains(LocalMethod1())),
-                entryCount: 19);
+                cs => cs.Where(c => c.ContactName.Contains(LocalMethod1())), //case-insensitive
+                cs => cs.Where(c => c.ContactName.Contains(LocalMethod1().ToLower()) || c.ContactName.Contains(LocalMethod1().ToUpper())), //case-sensitive
```
anpete: Ha! You fixed this the exact same way I did.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/RowNumberPagingTest.cs
```diff
@@ -203,6 +203,12 @@ public override void String_Contains_Literal()
             // base.String_Contains_Literal()
         }
 
+        public override void String_Contains_MethodCall()
```
anpete: Why is this disabled?

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
smitpatel: `ClientEvalPredicateStateless` is static constant function so it was getting translated to SQL. I changed this so that count remains client eval and the test scenario remains the same.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/RowNumberPagingTest.cs
smitpatel: Copied from above method `String_Contains_Literal()`. Both of them are being run in `QuerySqlServerTest` and have no impact due to rownumber paging.

anpete: Isn't that true of most of the base test methods? I think our default is just to run them all anyway.

smitpatel: Enabled both.

anpete:
:shipit:

Issue #5075, opened 2016-04-14
Make notification change tracking opt-in


ajcvickers:
Currently we dynamically decide what kind of change tracking to use based on the entity types and even instances used:
* Does not implement INotifyPropertyChanged or INotifyPropertyChanging: use snapshots with change detection
* Implements INotifyPropertyChanged but not INotifyPropertyChanging: use snapshots with change notification
* Implements both INotifyPropertyChanged and INotifyPropertyChanging: use only change notification (with lazy snapshots where original values are needed)

In addition, if an entity instance has a non-notifying collection navigation property instance, then it will still need snapshot change tracking.

Also note:
* These choices will only work correctly if the entity type correctly implements the interfaces for _all_ mapped properties and navigations.
* Entity types may implement these interfaces for reasons other than EF change tracking. In fact, it may be much more likely that entity types implement these interfaces for other reasons--e.g. data binding. In such cases it may be fine to only implement the interfaces for certain properties.

I propose that we change to an opt-in mechanism to switch on notification based change tracking. The options would be:
* Snapshot with change detection (default)
* Snapshot with notifications (I assert that my entity types correctly implement INotifyPropertyChanged for all mapped properties and navigations and the collection navigation properties are all notifying collections.)
* Full notifications (I assert that my entity types correctly implement both INotifyPropertyChanging and INotifyPropertyChanged for all mapped properties and navigations and the collection navigation properties are all notifying collections.)

The APi would allow this setting to be configured for the entire model and for each entity type.

rowanmiller:
:+1:

davidroth:
@ajcvickers  :+1: for the new API.

I guess the third option (correctly implementing _INotifyPropertyChanged_+ _INotifyPropertyChanging_) will show the best performance / lowest memory usage?

@rowanmiller  Are there any benchmarks for these different configurations?

Issue #445, opened 2014-07-21
Handle INotifyCollectionChanged to enable full notification change tracking


ajcvickers:
Currently we handle INotifyPropertyChanged and INotifyPropertyChanging for detecting changes to properties without the need for DetectChanges. However, collections still require DetectChanges. The need for DetectChanges could be removed if all collections also implement INotifyCollectionChanged. Related to this is the need to make sure we know when DetectChanges and snapshots are needed and when they are not.

PR #5196, opened 2016-04-27
Make notification change tracking opt-in and implement collection notifications

ajcvickers:
Issue #5075
Issue #445
Also, updates options for PR #5118

Use of INotifyProperyChanged and INotifyPropertyChanging is now opt-in. If you opt-in, then you must also start using INotifyCollectionChanged on your collection nav props, which is now implemented. We check these things in various places, including model validation and when we start tracking entities to try to make sure that you won't think that change tracking is happening when it isn't.

Doing a Clear on ObservableCollection does not work because the event does not tell us which items were removed. This has always been a problem for EF. An implementation of ObservableCollection where Clear does work is provided, and is used by default when we initialize collections on notification entities.

/cc @rowanmiller for API changes and exception messages

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/TestModels/ChangedOnlyMonsterContext.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/TestModels/ChangedOnlyMonsterContext.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Runtime.CompilerServices;
 
@@ -86,7 +87,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestModels
 
             public void InitializeCollections()
             {
-                BadScans = BadScans ?? new HashSet<IIncorrectScan>();
+                BadScans = BadScans ?? new ObservableCollection<IIncorrectScan>();
             }
 
             public byte[] Code
@@ -608,8 +609,8 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestModels
 
             public void InitializeCollections()
             {
-                OrderLines = OrderLines ?? new HashSet<IOrderLine>();
-                Notes = Notes ?? new HashSet<IOrderNote>();
+                OrderLines = OrderLines ?? new ObservableCollection<IOrderLine>();
+                Notes = Notes ?? new ObservableCollection<IOrderNote>();
             }
 
             public int AnOrderId
@@ -856,11 +857,11 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestModels
 
             public void InitializeCollections()
             {
-                Suppliers = Suppliers ?? new HashSet<ISupplier>();
-                //Replaces = new HashSet<DiscontinuedProduct>();
-                Reviews = Reviews ?? new HashSet<IProductReview>();
-                Photos = Photos ?? new HashSet<IProductPhoto>();
-                Barcodes = Barcodes ?? new HashSet<IBarcode>();
+                Suppliers = Suppliers ?? new ObservableCollection<ISupplier>();
+                //Replaces = new ObservableCollection<DiscontinuedProduct>();
+                Reviews = Reviews ?? new ObservableCollection<IProductReview>();
+                Photos = Photos ?? new ObservableCollection<IProductPhoto>();
+                Barcodes = Barcodes ?? new ObservableCollection<IBarcode>();
                 Dimensions = Dimensions ?? new Dimensions();
             }
 
@@ -950,7 +951,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestModels
 
             public void InitializeCollections()
             {
-                Features = Features ?? new HashSet<IProductWebFeature>();
+                Features = Features ?? new ObservableCollection<IProductWebFeature>();
             }
 
             public int ProductId
@@ -988,7 +989,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestModels
 
             public void InitializeCollections()
             {
-                Features = Features ?? new HashSet<IProductWebFeature>();
+                Features = Features ?? new ObservableCollection<IProductWebFeature>();
             }
 
             public int ProductId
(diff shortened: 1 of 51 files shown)


src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -165,11 +165,20 @@
   <data name="IQueryableProviderNotAsync" xml:space="preserve">
     <value>The provider for the source IQueryable doesn't implement IAsyncQueryProvider. Only providers that implement IEntityQueryProvider can be used for Entity Framework asynchronous operations.</value>
   </data>
-  <data name="EagerOriginalValuesRequired" xml:space="preserve">
-    <value>Lazy original value tracking cannot be turned on for entity type '{entityType}'. Entities that do not implement both INotifyPropertyChanging and INotifyPropertyChanged require original values to be stored eagerly in order to correct detect changes made to entities.</value>
+  <data name="ChangeTrackingInterfaceMissing" xml:space="preserve">
+    <value>The entity type '{entityType}' is configured to use the '{changeTrackingStrategy}' change tracking strategy but does not implement the required interface '{notificationInterface}'.</value>
```
rowanmiller: `required interface '{notificationInterface}'` should probably read `required '{notificationInterface}' interface`

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -165,11 +165,20 @@
...
+  </data>
+  <data name="NonNotifyingCollection" xml:space="preserve">
+    <value>The collection set for navigation property '{navigation}' on entity type '{entityType}' does not implement 'INotifyCollectionChanged'. Any entity type configured to use the '{changeTrackingStrategy}' change tracking strategy must use collections that implement 'INotifyCollectionChanged'. Consider using 'ObservableCollection&lt;T&gt;' for this.</value>
```
rowanmiller: Maybe `The collection type being used for navigation property...`?

src/Microsoft.EntityFrameworkCore/ModelBuilder.cs
```diff
@@ -208,6 +208,13 @@ public virtual ModelBuilder Ignore([NotNull] Type type)
             return this;
         }
 
+        public virtual ModelBuilder HasChangeTrackingStrategy(ChangeTrackingStrategy changeTrackingStrategy)
```
rowanmiller: Does this imply that strategy is used for all entity types? Or only those that implement the interface?

ajcvickers: All. We never infer based on interface implementation anymore.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -165,11 +165,20 @@
...
+  </data>
+  <data name="ResetNotSupported" xml:space="preserve">
+    <value>'ObservableCollection&lt;T&gt;.Clear()' is not supported because it uses the 'INotifyCollectionChanged' 'Reset' operation, which does not supply the items removed. Either use multiple calls to 'Remove' or use a notifying collection that supports 'Clear', such as 'Microsoft.EntityFrameworkCore.ChangeTracking.ObservableCollectionWithClear&lt;T&gt;'.</value>
```
rowanmiller: Can you use a `<see cref="..." />` tag for the collection type

rowanmiller:
Looks good from my point of view (other than comments I made). Not a huge fan of the `ObservableCollectionWithClear` name... but I don't have a better alternative to propose 😖.

src/Microsoft.EntityFrameworkCore/ModelBuilder.cs
rowanmiller: 👍 that's what I was hoping 😄

ErikEJ:
ObservableCollectionClearable ?

ajcvickers:
@rowanmiller One thing I was thinking of was doing a new implementation of a collection that implements INotifyCollectionChanged that is based on a hash set. It would be good for those people who have run into perf issues with ObservableCollection being based on a list, and would probably be a better default for us. Then we could call it ObservableHashSet.

rowanmiller:
@ajcvickers that sounds great

rowanmiller:
BTW I'm fine with pushing what we have for the moment and doing that separately

ajcvickers:
Sounds good to me. Still need a dev code review. Unless we are deciding that @rowanmiller's review is enough? ;-) @divega?

rowanmiller:
I only looked at public surface and resx 😄

divega:
Looking...

src/Microsoft.EntityFrameworkCore.Specification.Tests/TestModels/ChangedOnlyMonsterContext.cs
```diff
@@ -1230,8 +1231,8 @@ public class Supplier : NotificationEntity, ISupplier
 
             public void InitializeCollections()
             {
-                Products = Products ?? new HashSet<IProduct>();
-                //BackOrderLines = new HashSet<BackOrderLine>();
+                Products = Products ?? new ObservableCollection<IProduct>();
+                //BackOrderLines = new ObservableCollection<BackOrderLine>();
```
divega: Are these lines commented out on purpose?

ajcvickers: When the tests were ported from EF6 there was no inheritance support, so the stuff that used inheritance was commented out. They are still commented out now because nobody has updated the tests following addition of inheritance support.

divega:
:shipit: 
👍 on having a HashSet that is observable. Not sure when we should do it, can you file a bug?

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
ajcvickers: @rowanmiller Not sure how this would work in an exception message. We don't do it anywhere else I can find.

rowanmiller: Haha... brain fart on my behalf... never mind 😄

Issue #4854, opened 2016-03-18
Add: NullRef when adding graph with shadow keys


bricelam:
```C#
class TestAssembly
{
    [Key]
    public string Name { get; set; }
    public ICollection<TestClass> Classes { get; } = new List<TestClass>();
}

class TestClass
{
    public TestAssembly Assembly { get; set; }
    public string Name { get; set; }
    public ICollection<Test> Tests { get; } = new List<Test>();
}

class Test
{
    public TestClass Class { get; set; }
    public string Name { get; set; }
}

class Context : DbContext
{
    public DbSet<TestAssembly> Assemblies { get; set; }
    public DbSet<TestClass> Classes { get; set; }
    public DbSet<Test> Tests { get; set; }

    protected override void OnConfiguring(DbContextOptionsBuilder options)
        => options.UseSqlite("Data Source=ConsoleApplication1.db");

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<TestClass>(
            x =>
            {
                x.Property<string>("AssemblyName");
                x.HasKey("AssemblyName", nameof(TestClass.Name));
                x.HasOne(c => c.Assembly).WithMany(a => a.Classes)
                    .HasForeignKey("AssemblyName");
            });
        modelBuilder.Entity<Test>(
            x =>
            {
                x.Property<string>("AssemblyName");
                x.HasKey("AssemblyName", "ClassName", nameof(Test.Name));
                x.HasOne(t => t.Class).WithMany(c => c.Tests)
                    .HasForeignKey("AssemblyName", "ClassName");
            });
    }
}

class Program
{
    static void Main()
    {
        using (var db = new Context())
        {
            db.Database.EnsureDeleted();
            db.Database.EnsureCreated();

            var assembly = new TestAssembly { Name = "Assembly1" };
            var testClass = new TestClass { Assembly = assembly, Name = "Class1" };
            db.Tests.Add(new Test { Class = testClass, Name = "Test1" });

            // Throws NullReferenceException
            db.SaveChanges();
        }
    }
}
```
    System.NullReferenceException: Object reference not set to an instance of an object.
        at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.CompositePrincipalKeyValueFactory.CompositeComparer.Equals(Object[] x, Object[] y)
        at System.Collections.Generic.Dictionary`2.Remove(TKey key)
[...]

PR #5197, opened 2016-04-27
Fix for NullRef when adding graph with shadow keys

ajcvickers:
Issue #4854

Switch to using Equals(x, y) instead of x.Equals(y).


anpete:
:shipit:

Issue #3804, opened 2015-11-19
EF 7 LINQ query with .Select() including collection throws ArgumentNullException when .Take() is used


jemiller0:
I found that the query below throws an ArgumentNullException when used with EF 7 RC 1. It works in EF 6. If you comment out the .Take(100), the query works. Also, it doesn't matter if there are any rows in the database. Initially, I was using .FirstOrDefault() which has the same problem, but, I found that the problem even occurs if you use .Any(). 

```c#
using Microsoft.Data.Entity;
using System.Collections.Generic;
using System.Linq;

namespace ConsoleApplication1
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var bc = new BlogContext())
            {
                bc.Database.EnsureCreated();
                for (int i = 0; i < 5; i++)
                {
                    var b = new Blog { };
                    bc.Blogs.Add(b);
                    var p = new Post { Blog = b };
                    bc.Posts.Add(p);
                }
                bc.SaveChanges();
                bc.Blogs
                    .Take(100)
                    //.Select(b2 => b2.Posts.Any())
                    .Select(b2 => b2.Posts.FirstOrDefault().Content)
                    .ToArray();
            }
        }
    }

    public class BlogContext : DbContext
    {
        public DbSet<Blog> Blogs { get; set; }
        public DbSet<Post> Posts { get; set; }
        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder) => optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=Blogs;Trusted_Connection=True;");
    }

    public class Blog
    {
        public int? Id { get; set; }
        public ICollection<Post> Posts { get; set; }
    }

    public class Post
    {
        public int? Id { get; set; }
        public string Content { get; set; }
        public int? BlogId { get; set; }
        public Blog Blog { get; set; }
    }
}
```

rowanmiller:
Appears to be the same code as https://github.com/aspnet/EntityFramework/issues/3807 but one reports bad SQL and the other an exception

jemiller0:
Yes, the code is almost the same. Basically, if you have a Take(), it fails with an exception. If you comment that out, it works, however, the generated SQL looks wrong. It works, but, it would be very inefficient if the table had a large number or rows since it isn’t filtering and is querying for the entire table (the subqueries).

Jon

rowanmiller:
[...]

Issue #3913, opened 2015-11-27
.Sum()+.Count() inside .Select can throw exception because .Sum() can be null


DavidKarlas:
Databases/EF is not my home... I will throw bunch of code and theory and you guys will probably know better if this is bug or not...

I want to display events to volunteer and also number of registered volunteers per event...
Number of volunteers is Count of all Volunteers(Foreign key ManyToMany) + Each volunteer can add few extra volunteers who don't have account on website... ExtraVolunteers...
```
        //
        // GET: /Volunteer/Index
        [HttpGet]
        public IActionResult Index()
        {
            var model = new IndexViewModel();
            model.Events.AddRange(dbContext.Events
                .Where(e => e.StartTime > DateTime.Now)
                .Include(e => e.Location)
                .Include(e => e.Volunteers)
                .Select(e => new IndexViewModel.EventIndexViewModel()
                {
                    Id = e.Id,
                    Name = e.Name,
                    LocationName = e.Location.Name,
                    StartTime = e.StartTime,
                    EndTime = e.EndTime,
                    TotalVolunteersNeeded = e.TotalVolunteersNeeded,
                    CurrentVolunteers = e.Volunteers.Sum(v => v.ExtraVolunteers)  + e.Volunteers.Count()
                }));
            return View(model);
        }
```
This produces this SQL:
```
SELECT [e].[Id], [e].[Name], [e.Location].[Name], [e].[StartTime], [e].[EndTime], [e].[TotalVolunteersNeeded], (
SELECT SUM([v].[ExtraVolunteers])
FROM [DbUserAtEvent] AS [v]
WHERE [e].[Id] = [v].[EventId]
) + (
SELECT COUNT(*)
FROM [DbUserAtEvent] AS [d]
WHERE [e].[Id] = [d].[EventId]
)
FROM [DbEvent] AS [e]
INNER JOIN [Location] AS [e.Location] ON [e].[LocationId] = [e.Location].[Id]
WHERE [e].[StartTime] > GETDATE()
```

And throws this exception in aspnet console:
```
fail: Microsoft.Data.Entity.Query.Internal.SqlServerQueryCompilationContextFactory[1]
      An exception occurred in the database while iterating the results of a query.
      System.NullReferenceException: Object reference not set to an instance of an object.
         at lambda_method(Closure , TransparentIdentifier`2 )
         at System.Linq.Enumerable.<>c__DisplayClass2`3.<CombineSelectors>b__3(TSource x)
[...]

Issue #3676, opened 2015-11-06
Usage of the "let" keyword breaks grouping


AlekseyMartynov:
Version: 7.0.0-rc1-*

Query:

    from p in nwind.Products
    let cat = p.CategoryID
    group p by cat;
    // group p by p.CategoryID; // - also fails when 'let' is present

Message:

    Expression of type '...' cannot be used for parameter of type '...' of method '...'

Full sample: https://github.com/AlekseyMartynov/ef7-bugs/tree/group-with-let

mikary:
Verified that this query throws an `ArgumentException` in `EntityFramework.MicrosoftSqlServer 7.0.0-rc1-final`, but the exception is not thrown in current dev.

The returned results appear to be incorrect due to a related issue (#5179) with navigations.

mikary:
Retested now that related issue has been resolved. The initial query completes successfully; however, a `NullReferenceException` is thrown in `LinqOperatorProvider.GetEnumerator` when enumerating the elements within the `IGrouping<TKey, TElement>`

```cs
[ConditionalFact]
public virtual void Let_group_by_nav_prop()
{
    AssertQuery<OrderDetail, IGrouping<string, OrderDetail>>(
        ods => from od in ods
                let customer = od.Order.CustomerID
                group od by customer
                into odg
                select odg,
        asserter: (l2oItems, efItems) =>
        {
            foreach (var pair in
                from l2oItem in l2oItems
                join efItem in efItems on l2oItem.Key equals efItem.Key
                select new { l2oItem, efItem })
            {
                Assert.Equal(
                    pair.l2oItem.Select(i => i.OrderID).OrderBy(i => i),
                    pair.efItem.Select(i => i.OrderID).OrderBy(i => i));
            }
        });
}
```

Issue #3674, opened 2015-11-06
Grouping with Where on a navigation prop fails


AlekseyMartynov:
Version: "7.0.0-rc1-*

Query:

    from o in nwind.Orders
    where o.Shipper.CompanyName == "Speedy Express"
    group o by o.OrderDate;

Message:

    Expression of type '...' cannot be used for parameter of type '...' of method '...'

Full sample: https://github.com/AlekseyMartynov/ef7-bugs/tree/group-navigation-where

mikary:
Verified that this query throws an `ArgumentException` in `EntityFramework.MicrosoftSqlServer 7.0.0-rc1-final`, but the exception is not thrown in current dev.

The returned results appear to be incorrect due to a related issue (#5179) with navigations.

rowanmiller:
@mikary can you verify now that #5179 is fixed

mikary:
Verified and added test coverage to prevent future regressions.

PR #5198, opened 2016-04-27
Add tests for #3804 - Select Navigation following Skip or Take

mikary:
@maumar @anpete


maumar:
you might want to hold of on this checkin - I just submitted a PR that adds better result verification for the navigation tests. I think we should take advantage of it for all the new tests that are being added

divega:
@maumar @mikary should we now revisit this PR?

mikary:
@divega I've been looking into adding tests for a number of related issues (#3804 #3674 #3676 and #3913) that have been resolved (or partially resolved in the case of #3676) and adding tests for the scenarios described in the issues to avoid regressions. All of these tests would be in `QueryNavigationsTestBase` and I'm looking at consolidating all of these tests into one PR.

mikary:
@maumar updated with tests for the listed issues

maumar:
:shipit:

Issue #5179, opened 2016-04-26
Query: Select Where Navigation returns wrong results on relational providers


mikary:
The existing test in `QueryNavigationsTestBase` returns the correct number of results but incorrect values on relational providers.

```cs
        [ConditionalFact]
        public virtual void Select_Where_Navigation()
        {
            using (var context = CreateContext())
            {
                var orders
                    = (from o in context.Set<Order>()
                       where o.Customer.City == "Seattle"
                       select o).ToList();

                Assert.Equal(14, orders.Count);
            }
        }
```

The list of orders contains 14 references to the same order (10904 on SqlServer, 10269 on Sqlite) instead of the expected results for the query. Results appear to be correct on the InMemory provider.

mikary:
Generated SQL for SqlServer:
```sql
SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate], [o.Customer].[CustomerID], [o.Customer].[Address], [o.Customer].[City], [o.Customer].[CompanyName], [o.Customer].[ContactName], [o.Customer].[ContactTitle], [o.Customer].[Country], [o.Customer].[Fax], [o.Customer].[Phone], [o.Customer].[PostalCode], [o.Customer].[Region]
FROM [Orders] AS [o]
LEFT JOIN [Customers] AS [o.Customer] ON [o].[CustomerID] = [o.Customer].[CustomerID]
ORDER BY [o].[CustomerID]
```

rowanmiller:
@maumar this should be your top priority

PR #5200, opened 2016-04-28
Fix to #5179 - Query: Select Where Navigation returns wrong results on relational providers

maumar:
Problem was that when creating groups in GroupJoin we would only look at the inner key selector to differentiate between groups. However in some cases (specifically 1-Many navigations coming from the many side - e.g. order.Customer) outer element constantly changes, while inner element stays the same. If later in the pipeline outer elements are being used, e.g. in projection, we get incorrect results.

Fix is to also look at changes between outer elements and create a new group every time either outer element or a inner key changes.

Also as part of this change we improve result verification for the navigation tests (previously we would only verify that the result count is as expected).

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/ValueBufferShaper.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/ValueBufferShaper.cs
@@ -17,6 +17,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
         public override Type Type => typeof(ValueBuffer);
 
+        public virtual object GetKey(QueryContext queryContext, ValueBuffer valueBuffer)
+            => null;
+
         public virtual ValueBuffer Shape(QueryContext queryContext, ValueBuffer valueBuffer)
             => valueBuffer;
     }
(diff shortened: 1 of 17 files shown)


maumar:
@anpete for the query part, @ajcvickers for the change tracker part, @divega for the rc2 risk assessment

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/ValueBufferShaper.cs
```diff
@@ -17,6 +17,9 @@ public ValueBufferShaper([NotNull] IQuerySource querySource)
 
         public override Type Type => typeof(ValueBuffer);
 
+        public virtual object GetKey(QueryContext queryContext, ValueBuffer valueBuffer)
+            => null;
```
maumar: not sure what to do for ValueBuffer - currenly we dont check for differences between ValueBuffer results, so there still might be come instances where we get wrong results

anpete:
@maumar I have a PR out for this. Can you turn this into a PR for the test improvements?

mikary:
@maumar Shouldn't this be a PR against release? That might impact the way we deal with / merge onto @anpete s PR

maumar:
the work was done in other commits, closing

Issue #448, opened 2014-07-21
Identity map must be updated when sidecar values are discarded after failed SaveChanges


ajcvickers:
Currently values are reverted but the state manager identity map is left in a bad state.

PR #5208, opened 2016-04-28
Update identity map after SaveChanges fails

ajcvickers:
Issue #448

Store-generated values are discarded, but the identity maps from key value to entry were not being updated. The fix is to send a changed notification for any property that had been assigned a store generated value.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/StoreGeneratedSqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/StoreGeneratedSqlServerTest.cs
@@ -2,9 +2,15 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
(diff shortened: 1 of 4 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/StoreGeneratedSqlServerTest.cs
```diff
@@ -16,6 +22,48 @@ public StoreGeneratedSqlServerTest(StoreGeneratedSqlServerFixture fixture)
         {
         }
 
+        [Fact]
+        public virtual void Exception_in_SaveChanges_causes_store_values_to_be_reverted()
+        {
+            using (var context = CreateContext())
+            {
+                var entities = new List<Darwin>();
+                for (var i = 0; i < 1000; i++)
```
natemcmaster: Do we need 1000 to repro the original bug? Or would fewer suffice? (Thinking about slow a full build is are already.)

ajcvickers: I tested with different numbers and it made no difference to the test execution time. We don't need 1000, but there is no good way I can think of to ensure that the bad one isn't first. Using a lot makes it less likely that the bad one is first.

natemcmaster:
:shipit:

Issue #5135, opened 2016-04-20
Consider moving SqlDataReaderExtension and SqlServerTableSelectionSetExtensions to internal namespace


divega:
Relative to other types in `Microsoft.EntityFrameworkCore.SqlServer.Design` they seem to be more implementation detail than public surface that we want to expose to anyone.

lajones:
Discussed this in a team meeting today and decided to put all the classes in the `M.EFCore.SqlServer.Design` assembly into `.Internal` namespaces to indicate that, although other provider-writers can take dependencies on some of those classes, we still feel free to make breaking changes to them. Same for the `M.EFCore.Sqlite.Design` assembly.

lajones:
Fixed as discussed above with PR #5209.

PR #5209, opened 2016-04-28
Moving SqlServer.Design and Sqlite.Design classes into Internal namespaces

lajones:
Fix for #5135. Just changing some namespaces.


natemcmaster:
:shipit:

lajones:
Checked in with commit 778be41.

Issue #4347, opened 2016-01-20
Null ref when attempting to back propagate non-existent results


anpete:
Entity key is configured for IDENTITY but database is misconfigured to not have IDENTITY on the column.

```
Microsoft.Data.Entity.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Data.Entity.ChangeTracking.Internal.Snapshot`2.set_Item(Int32 index, Object value)
   at Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntry.SetOriginalValue(IPropertyBase propertyBase, Object value)
   at Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntry.SetPropertyModified(IProperty property, Boolean isModified)
   at Microsoft.Data.Entity.ChangeTracking.Internal.ChangeDetector.PropertyChanged(InternalEntityEntry entry, IPropertyBase propertyBase)
   at Microsoft.Data.Entity.ChangeTracking.Internal.InternalEntityEntryNotifier.Dispatch(Action`1 action)
   at Microsoft.Data.Entity.Update.ModificationCommand.PropagateResults(ValueBuffer valueBuffer)
   at Microsoft.Data.Entity.Update.AffectedCountModificationCommandBatch.ConsumeResultSetWithPropagation(Int32 commandIndex, DbDataReader reader)
   at Microsoft.Data.Entity.Update.AffectedCountModificationCommandBatch.Consume(DbDataReader reader)
   --- End of inner exception stack trace ---
   at Microsoft.Data.Entity.Update.AffectedCountModificationCommandBatch.Consume(DbDataReader reader)
   at Microsoft.Data.Entity.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)
   at Microsoft.Data.Entity.Update.Internal.BatchExecutor.Execute(IEnumerable`1 commandBatches, IRelationalConnection connection)
   at Microsoft.Data.Entity.ChangeTracking.Internal.StateManager.SaveChanges(IReadOnlyList`1 entriesToSave)
   at Microsoft.Data.Entity.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.Data.Entity.DbContext.SaveChanges(Boolean acceptAllChangesOnSuccess)
[...]

PR #5210, opened 2016-04-28
Throw better exception when database back propagates null for non-nullable property

ajcvickers:
Issue #4347


lajones:
:shipit:

ajcvickers:
Merged

wgutierrezr:
Hi
I have this problem whit a table with a Bit field with default '0' (false). When I try to create a new record using ASP.Net Core 1.1 MVC, and EntityFrameworkCore.Sqlserver 1.1.2, passing the value for the boolean field with 'False' then it triggers this error.

Do you have any solution for this problem?

ajcvickers:
@wgutierrezr The fix for this issue was merged some time ago. Can you please file a new issue with full details of how to reproduce what you are seeing?

PR #5211, opened 2016-04-28
[TINY] Increase work in fixup query tests

rowanmiller:
We recently updated the other fixup tests to work on 5,000 customers &
10,000 orders. Upping the tests that fixup with a query to work on the
same entity counts.


lajones:
Looks good :shipit:

divega:
:shipit:

Issue #5126, opened 2016-04-20
Consider ignoring common Tooling tables eg __RefactorLog when doing reveng


mikes-gh:
2 examples I have in my reveng generated code.
__RefactorLog (used my SSDT)
dtproperties (not sure about this one I think tool is now dead so maybe moot).

Consider an ignore list for reveng 
Not sure how __EFMigrationHistory is ignored.
I cant see any scenario where the above would be useful in the dbcontext and model classes.

natemcmaster:
As you may know, dbcontext scaffolding can [take parameters](http://docs.efproject.net/en/latest/cli/powershell.html#scaffold-dbcontext) to filter on specific tables and schema. For now I consider this the best way to prevent these tables from being scaffolded. To solve this generally, we'd need a clear answer to "what are common table names that should always be excluded?"

ErikEJ:
Providing exclusion filter rather than inclusion filter would be a good start

rowanmiller:
We should handle this with the ability to exclude tables

rowanmiller:
BTW we should be ignoring system tables... not sure if __RefactorLog is a system table or not

ErikEJ:
You should add: is_ms_shipped to the tables query - and refactorlog is a user table: 

https://technet.microsoft.com/da-dk/library/ms190324(v=sql.110).aspx 

http://stackoverflow.com/questions/10029979/how-to-check-if-sql-server-tables-are-system-tables

mikes-gh:
@ErikEJ 
Good find
I confirm both the examples I quoted are in system table folder 

![system tables](https://cloud.githubusercontent.com/assets/16208742/14691442/2a30ac34-0749-11e6-98c8-fc410a740be8.png)

so the stack overflow article is the answer.  Although it doesn't look as straightforward as is_ms_shipped = 0 

> it's a combination - either is_ms_shipped from sys.tables being 1, or having a particular extended property set.

The article shows the query SSMS uses to classify.

mikes-gh:
**The following**
```
select Name
from sys.tables AS tbl
where 
	CAST(
	case 
		when tbl.is_ms_shipped = 1 then 1
		when (
			select major_id 
			from sys.extended_properties 
			where 
				major_id = tbl.object_id and 
				minor_id = 0 and 
				class = 1 and 
				name = N'microsoft_database_tools_support') 
			is not null then 1
		else 0
	end AS Bit) = 0   
```
[...]

PR #5213, opened 2016-04-28
[SMALL] Ignore system tables for reverse engineering

lajones:
Fix for #5126.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
@@ -197,32 +197,52 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
         private void GetTables()
         {
             var command = _connection.CreateCommand();
+            // for origin of the CASE statement below see https://github.com/aspnet/EntityFramework/issues/5126
             command.CommandText =
-                "SELECT schema_name(t.schema_id) AS [schema], t.name FROM sys.tables AS t " +
-                $"WHERE t.name <> '{HistoryRepository.DefaultTableName}'" +
-                TemporalTableWhereClause;
+                @"SELECT
+    schema_name(t.schema_id) AS [schema],
+    t.name,
+    CAST (
+      CASE
+        WHEN t.is_ms_shipped = 1 THEN 1
+        WHEN (
+          SELECT major_id
+          FROM  sys.extended_properties
+          WHERE major_id = t.object_id
+          AND   minor_id = 0
+          AND   class = 1
+          AND   name = N'microsoft_database_tools_support'
+        ) IS NOT NULL THEN 1
+        ELSE 0
+      END
+    AS bit) AS [is_system_object]
+    FROM sys.tables AS t " +
+    $"WHERE t.name <> '{HistoryRepository.DefaultTableName}'" + TemporalTableWhereClause;
             using (var reader = command.ExecuteReader())
             {
                 while (reader.Read())
                 {
-                    var table = new TableModel
+                    if (!reader.GetValueOrDefault<bool>("is_system_object"))
                     {
-                        Database = _databaseModel,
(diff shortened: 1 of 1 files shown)


src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -197,32 +197,52 @@ private void GetSequences()
...
+      END
+    AS bit) AS [is_system_object]
+    FROM sys.tables AS t " +
+    $"WHERE t.name <> '{HistoryRepository.DefaultTableName}'" + TemporalTableWhereClause;
             using (var reader = command.ExecuteReader())
             {
                 while (reader.Read())
                 {
-                    var table = new TableModel
+                    if (!reader.GetValueOrDefault<bool>("is_system_object"))
```
natemcmaster: Could this just be part of the SQL query? `WHERE is_system_object <> 1`

natemcmaster:
:shipit:

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
divega: Furthermore, if the condition doesn't need to be in the projection it seems we can remove the CASE and write a simpler condition in the WHERE clause.

lajones: You can't directly refer to `is_system_object` in the WHERE clause - however I can do @divega's plan. Updated - please refresh and review again. Thanks.

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -197,10 +197,22 @@ private void GetSequences()
...
+    FROM sys.tables AS t
+    WHERE t.is_ms_shipped <> 1
+    AND (SELECT major_id
+         FROM  sys.extended_properties
+         WHERE major_id = t.object_id
+         AND   minor_id = 0
+         AND   class = 1
+         AND   name = N'microsoft_database_tools_support'
+        ) IS NULL " +
+    $"AND t.name <> '{HistoryRepository.DefaultTableName}'" + TemporalTableWhereClause;
```
divega: Much nicer IMO. It would be perfect with NOT EXISTS :smile:, e.g. 

``` SQL
AND NOT EXISTS (SELECT *  
  FROM  sys.extended_properties  
  WHERE major_id = t.object_id  
  AND   minor_id = 0  
  AND   class = 1  
  AND   name = N'microsoft_database_tools_support') 
```

divega:
:shipit:

src/Microsoft.EntityFrameworkCore.SqlServer.Design/SqlServerDatabaseModelFactory.cs
```diff
@@ -197,10 +197,22 @@ private void GetSequences()
...
+            // below see https://github.com/aspnet/EntityFramework/issues/5126
             command.CommandText =
-                "SELECT schema_name(t.schema_id) AS [schema], t.name FROM sys.tables AS t " +
-                $"WHERE t.name <> '{HistoryRepository.DefaultTableName}'" +
-                TemporalTableWhereClause;
+                @"SELECT
+    schema_name(t.schema_id) AS [schema],
+    t.name
+    FROM sys.tables AS t
+    WHERE t.is_ms_shipped <> 1
```
ErikEJ: Why not: WHERE t.is_ms_shipped = 0 ?

lajones: Yep - is_ms_shipped is a `bit` column - so will do.

lajones:
Updated as suggested by @divega and @ErikEJ and checked in with commit e86848c.

PR #5214, opened 2016-04-29
adding query perf test for optional navigation

maumar:


maumar:
@rowanmiller

test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/Query/NavigationsQueryTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            _fixture = fixture;
+        }
+
+        [Benchmark]
+        [BenchmarkVariation("Sync (10 queries)", false, 10)]
+        [BenchmarkVariation("Async (10 queries)", true, 10)]
+        public async Task PredicateAcrossOptionalNavigation(IMetricCollector collector, bool async, int queriesPerIteration)
+        {
+            using (var context = AdventureWorksFixture.CreateContext())
```
rowanmiller: This should really be an instance method on the fixture. The static one just exists because the cold start tests do complicated stuff with AppDomains etc. Leave it as-is for now, I'll add an instance one later.

test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/Query/NavigationsQueryTests.cs
```diff
@@ -0,0 +1,52 @@
...
+                            await query.ToListAsync();
+                        }
+                        else
+                        {
+                            query.ToList();
+                        }
+                    }
+                }
+
+                Assert.Equal(466, query.Count());
```
rowanmiller: What kind of results do you see for this? I'm wondering with 466 results if it might execute too quickly to give us any meaningful numbers. You typically want to be seeing 50ms on your local machine at a minimum (100-300ms is ideal).

maumar: The filter actually runs on the client, so the number of results doesn't affect the time too much. However EF6 might be executing too fast. Will take a look

rowanmiller: Ok, it's probably fine then - though I assume we are aiming to translate this at some point. Looks like the store table has 701 rows and most the other tests pull back 1000 rows... so I think you are good.

test/Microsoft.EntityFrameworkCore.Microbenchmarks.EF6/Query/NavigationsQueryTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        [Benchmark]
+        [BenchmarkVariation("Sync (10 queries)", false, 10)]
+        [BenchmarkVariation("Async (10 queries)", true, 10)]
+        public async Task PredicateAcrossOptionalNavigation(IMetricCollector collector, bool async, int queriesPerIteration)
+        {
+            using (var context = AdventureWorksFixture.CreateContext())
+            {
+                var query = from s in context.Store
+                            where s.SalesPerson.Bonus > 3000
+                            select s;
```
rowanmiller: I not too worried about this, but would have a slight preference for method chaining syntax to be consistent with the other tests.

```c#
var query = context.Store
    .Where(s => s.SalesPerson.Bonus > 3000);
```

rowanmiller:
:shipit:

PR #5222, opened 2016-04-29
Query: Fix #5179: Select Where Navigation returns wrong results on re…

### LINKED ISSUE #5179: Query: Select Where Navigation returns wrong results on relational providers

anpete:
…lational providers

Fixed streaming relational GroupJoin operator to correctly re-group on outer element change.

Also fixed some places where we were failing to flow cancellation tokens.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
@@ -4331,6 +4331,16 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
                 from o in orders
                 select c);
         }
+        
+        [ConditionalFact]
+        public virtual void GroupJoin_simple3()
+        {
+            AssertQuery<Customer, Order>((cs, os) =>
+                from c in cs
+                join o in os on c.CustomerID equals o.CustomerID into orders
+                from o in orders
+                select new { o.OrderID });
+        }
 
         [ConditionalFact]
         public virtual void GroupJoin_simple_ordering()
@@ -5355,10 +5365,8 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             bool assertOrder = false,
             int entryCount = 0,
             Action<IList<object>, IList<object>> asserter = null)
-            where TItem : class
-        {
-            AssertQuery(query, query, assertOrder, entryCount, asserter);
-        }
+            where TItem : class 
+                => AssertQuery(query, query, assertOrder, entryCount, asserter);
 
         protected void AssertQuery<TItem>(
             Func<IQueryable<TItem>, object> query,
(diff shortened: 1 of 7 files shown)


anpete:
cc @divega @maumar

src/Microsoft.EntityFrameworkCore/Storage/ValueBuffer.cs
```diff
@@ -89,5 +89,28 @@ public ValueBuffer WithOffset(int offset)
         }
 
         public bool IsEmpty => _values == null;
+
+        public override bool Equals(object obj)
```
maumar: don't you need similar thing for CompositeShaper?

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -4331,6 +4331,16 @@ public virtual void GroupJoin_simple2()
                 from o in orders
                 select c);
         }
+        
+        [ConditionalFact]
+        public virtual void GroupJoin_simple3()
```
maumar: Might want to incorporate tests added in https://github.com/aspnet/EntityFramework/pull/5200, specifically files: QueryNavigationsTestBase, NorthwindData, QueryNavigationsSqlServerTest - there were a few scenarios returning wrong results that were caught by the overhaul

anpete: See my comment on the PR

maumar: Sounds good, will add the tests separately once your fix is in

src/Microsoft.EntityFrameworkCore/Storage/ValueBuffer.cs
anpete: This is ValueBuffer, not Shaper (if that's what you meant) :smile:

maumar: nevermind, default comparer for TransparentIdentifier should handle it just fine

maumar:
:shipit:

src/Microsoft.EntityFrameworkCore/Storage/ValueBuffer.cs
maumar: I meant the result of CompositeShaper.Shape - TransparentIdentifier, but it's default comparison should work since its a struct

PR #5225, opened 2016-04-29
Query: Convert IS operator in predicate to comparison with discriminator

### LINKED ISSUE #5007: Query: Inheritance filter with "is" performs client evaluation
The following query performs client filtering:
```cs
        [Fact]
        public virtual void Can_use_is_kiwi()
        {
            {
                var kiwis = context.Set<Animal>().Where(a => a is Kiwi).ToList();

                Assert.Equal(1, kiwis.Count);
            }
        }
```

With generated SQL:
```sql
SELECT [a].[Species], [a].[CountryId], [a].[Discriminator], [a].[Name], [a].[EagleId], [a].[IsFlightless], [a].[Group], [a].[FoundOn]
FROM [Animal] AS [a]
WHERE [a].[Discriminator] IN (N'Kiwi', N'Eagle')
```

Note, the same query with the where clause replaced with `OfType<Kiwi>()` filters on the server:
```sql
SELECT [a].[Species], [a].[CountryId], [a].[Discriminator], [a].[Name], [a].[EagleId], [a].[IsFlightless], [a].[Group], [a].[FoundOn]
FROM [Animal] AS [a]
WHERE [a].[Discriminator] = N'Kiwi'
```

smitpatel:
Resolves #5007

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -22,6 +22,7 @@ using Microsoft.Extensions.Logging;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Parsing;
 
 namespace Microsoft.EntityFrameworkCore.Query
 {
@@ -252,6 +253,9 @@ namespace Microsoft.EntityFrameworkCore.Query
         public override void VisitQueryModel(QueryModel queryModel)
         {
             Check.NotNull(queryModel, nameof(queryModel));
+
+            var typeIsExpressionTranslatingVisitor = new TypeIsExpressionTranslatingVisitor(QueryCompilationContext.Model, _relationalAnnotationProvider);
+            queryModel.TransformExpressions(typeIsExpressionTranslatingVisitor.Visit);
 
             base.VisitQueryModel(queryModel);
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -830,6 +835,68 @@ protected virtual void CheckClientEval([NotNull] object expression)
...
+                }
+
+                return Expression.Constant(true, typeof(bool));
+            }
+        }
+
+        private static readonly MethodInfo _efPropertyMethod
+            = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
+
+        private static Expression CreatePropertyExpression(Expression target, IProperty property)
```
smitpatel: these methods are being used in 2 places (other one being `EntityEqualityRewritingExpressionVisitor`), any good place to refactor?

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -701,6 +703,9 @@ public override void VisitWhereClause(WhereClause whereClause, QueryModel queryM
 
             if (!requiresClientFilter)
             {
+                var discriminatorPredicate = new PedicateTypeIsExpressionTranslatingVisitor(this).Visit(whereClause.Predicate);
```
maumar: typo

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
maumar: Maybe EntityQueryModelVisitor? we already have a IsPropertyMethod on there

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/InheritanceSqlServerTest.cs
```diff
@@ -30,7 +30,18 @@ public override void Can_use_is_kiwi()
             Assert.Equal(
                 @"SELECT [a].[Species], [a].[CountryId], [a].[Discriminator], [a].[Name], [a].[EagleId], [a].[IsFlightless], [a].[Group], [a].[FoundOn]
 FROM [Animal] AS [a]
-WHERE [a].[Discriminator] IN (N'Kiwi', N'Eagle')",
+WHERE [a].[Discriminator] IN (N'Kiwi', N'Eagle') AND ([a].[Discriminator] = N'Kiwi')",
```
maumar: that's something we could probably optimize (in the future)

maumar:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -830,6 +835,68 @@ protected virtual void CheckClientEval([NotNull] object expression)
             }
         }
 
+        private class PedicateTypeIsExpressionTranslatingVisitor : ExpressionVisitorBase
```
anpete: *Predicate*

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
anpete: Even introducing a new helper class would be preferable here.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
anpete: I don't think this should just be for predicates. Add some tests with *is* in the projection etc.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/InheritanceSqlServerTest.cs
smitpatel: filed #5230

smitpatel:
Updated with feedback.
The translatingvisitor now works on whole query model so projection/predicate both are translated. (added test for both).
Also removed generation of `Where 1 = 1` in query. If predicate is true then we don't generate where clause.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -830,6 +835,58 @@ protected virtual void CheckClientEval([NotNull] object expression)
             }
         }
 
+        private class TypeIsExpressionTranslatingVisitor : ExpressionVisitorBase
+        {
+            private readonly RelationalQueryModelVisitor _queryModelVisitor;
+
+            public TypeIsExpressionTranslatingVisitor(RelationalQueryModelVisitor relationalQueryModelVisitor)
```
anpete: Minor but consider reducing coupling here by only passing the required dependencies: Model and RelationalAnnotationProvider. Then you can revert the RelationalAnnotationProvider property change, too.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/EntityEqualityRewritingExpressionVisitor.cs
```diff
@@ -81,27 +81,17 @@ var newRightExpression
...
                     Expression.NewArrayInit(
                         typeof(object),
                         properties
-                            .Select(p => Expression.Convert(CreatePropertyExpression(target, p), typeof(object)))
+                            .Select(p => Expression.Convert(EntityQueryModelVisitor.CreatePropertyExpression(target, p), typeof(object)))
                             .Cast<Expression>()
                             .ToArray()));
         }
 
-        private static readonly MethodInfo _efPropertyMethod
```
anpete: Nice.

smitpatel:
updated.

anpete:
:shipit:

PR #5226, opened 2016-04-29
Improve navigation building perf

AndriySvyryd:
Refactor the internal model builders to avoid looking up the PropertyInfo for navigation properties more than once.

Add PropertyIdentity to unify the implementations of methods accepting either string or PropertyInfo, reduce code duplication and get rid of sentry values


AndriySvyryd:
This is basically a refactoring, but due to a large amount of code moved it's recommended to use CodeFlow to review InternalRelationshipBuilder.cs

ajcvickers:
:sheep::it:

PR #5228, opened 2016-04-29
Implementation of ObservableHashSet

ajcvickers:
This is an implementation of INotifyCollectionChanged that is backed by a HashSet instead of a List. This will then perform better in Contains when there are a large number of elements, which is something that ObservableCollection doesn't do well with. It also implements Clear in a way that can be used by EF notifications because the items being cleared are passed in the event.

Copies the API surface of HashSet<T>.

Also implements INotifyPropertyChanged/Changing for the Count property.

Does not have an inherent ordering and so does not have support for indexers, etc., or the notifications that go with these.


src/Microsoft.EntityFrameworkCore/ChangeTracking/ObservableHashSet.cs
```diff
@@ -0,0 +1,279 @@
...
+            => OnCollectionChanged(new NotifyCollectionChangedEventArgs(action, item));
+
+        private void OnCollectionChanged(IList newItems, IList oldItems)
+            => OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Replace, newItems, oldItems));
+
+        protected virtual void OnCollectionChanged([NotNull] NotifyCollectionChangedEventArgs e)
+            => CollectionChanged?.Invoke(this, e);
+    }
+
+    internal class ObservableHashSetSingletons
```
anpete: Curious why these are in a separate class

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/ChangeTracking/ObservableHashSet.cs
ajcvickers: Because only one instance of each is needed and if there were in the main class there would be one instance per generic type. It's not a big deal either way.

PR #5237, opened 2016-05-01
Switches the order of Configure*Services methods.

### LINKED ISSUE #5236: Entity Framework doesn't follow ASP.NET rules for hosting environment.
``` c#
public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddEntityFramework()
            .AddSqlServer()
            .AddDbContext<MyContext>(options => options.UseSqlServer(Configuration["SqlServerConnectionString"]));
    }

    public void ConfigureDevelopmentServices(IServiceCollection services)
    {
        services.AddEntityFramework()
            .AddSqlite()
            .AddDbContext<MyContext>(options => options.UseSqlite(new SqliteConnectionStringBuilder { Datasource = ":memory:" }.ToString()));
    }
}
```

When I run this with this DNX command: `"web": "Microsoft.AspNet.Server.Kestrel"` the code correctly steps through the `ConfigureDevelopmentServices` method and _NOT_ the `ConfigureServices` method.

However, when I run `dnx ef migrations add Foo` (`"ef": "EntityFramework.Commands"`), it gives me an exception complaining about not being able to find the `SqlServerConnectionString` indicating that the production code is running, not the development code.  `SqlServerConnectionString` is not available in development, only Sqlite is available in development so this error is correct but the code path leading up to it is the problem.

I tried `dnx ef migrations add --environment Development Foo` but that resulted in the same problem.  I also added a log line to `Startup` constructor to output the `hostingEnvironment` and it shows `hostingEnvironment.EnvironmentName` is `Development` (whether I add the `--environment` switch or not).

Zoltu:
Fixes #5236

This array is used in a loop that stops on the first method it finds.  It should try the more specific methods before the less specific methods.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools.Core/Design/Internal/StartupInvoker.cs
+++ b/src/Microsoft.EntityFrameworkCore.Tools.Core/Design/Internal/StartupInvoker.cs
@@ -51,7 +51,7 @@ namespace Microsoft.EntityFrameworkCore.Design.Internal
 
             return Invoke(
                 _startupType,
-                new[] { "ConfigureServices", "Configure" + _environment + "Services" },
+                new[] { "Configure" + _environment + "Services", "ConfigureServices" },
                 services) as IServiceProvider
                    ?? services.BuildServiceProvider();
         }


Zoltu:
I don't understand the Travis CI failure.  Let me know if it is unexpected (ideally with a hint as to where to look).

natemcmaster:
@Zoltu the travis CI failure was unrelated to your change. I restarted the travis build.

PR #5253, opened 2016-05-02
Fix build warnings

pranavkm:


test/Microsoft.EntityFrameworkCore.Microbenchmarks.Core/BenchmarkTestCaseBase.cs
```diff
@@ -8,10 +8,6 @@
 using Xunit.Abstractions;
 using Xunit.Sdk;
 
-#if NETCOREAPP1_0
```
pranavkm: cc @natemcmaster \ @bricelam 

I'm commonizing `dotnet-test-xunit` in order to allow us to use the dotnet test runner on desktop and netcoreapp. This line was failing to compile when I moved the dependency. However this project doesn't have any projects of it's own so I moved the runner into `Microsoft.EntityFrameworkCore.Microbenchmarks` and `Microsoft.EntityFrameworkCore.Microbenchmarks.EF6`

PR #5255, opened 2016-05-02
Query: Add translators for string functions

### LINKED ISSUE #5199: Query: Translate string.IsNullOrEmpty
I'm having a situation where I need to get data from a single table with 100k+ rows with a quite simple where clause. This query was originally created with EF6 and returned within 100ms.
The code is:
````c#
return await dbContext.Tees
	.Where(w => w.CourseId.Equals(courseId) && !string.IsNullOrEmpty(w.Name) && w.IsActive.Equals(true) && w.IsDeleted.Equals(false))
	.OrderBy(o => o.Color)
	.Select(s => new TeeDto
	{
		CourseId = s.CourseId,
		TeeId = s.TeeId,
		FrontPar = s.ParFront,
		BackPar = s.ParBack,
		CoursePar = s.CoursePar,
		DistanceBack = s.DistanceBack,
		DistanceFront = s.DistanceFront,
		DistanceTotal = s.DistanceTotal,
		Name = s.Name,
		Color = s.Color,
		Rating = s.Rating,
		Slope = s.Slope
	}).ToListAsync();
````
The query above took more than 55 secs to come back. So I used Profiler to discover that the generated SQL did not include any of the where clause except for the last one (IsDeleted).
````sql
SELECT [w].[CourseId], [w].[Name], [w].[IsActive], [w].[TeeId], [w].[ParFront], [w].[ParBack], [w].[CoursePar], [w].[DistanceBack], [w].[DistanceFront], [w].[DistanceTotal], [w].[Color], [w].[Rating], [w].[Slope]
FROM [Tee] AS [w]
WHERE [w].[IsDeleted] = 0
ORDER BY [w].[Color]
````
I changed the clauses order and pushed the `string.IsNullOrEmpty` at the end, so the query looks like
````c#
return await dbContext.Tees
	.Where(w => w.CourseId.Equals(courseId) && w.IsActive.Equals(true) && w.IsDeleted.Equals(false) && !string.IsNullOrEmpty(w.Name))
	.OrderBy(o => o.Color)
	.Select(s => new TeeDto
	{
		CourseId = s.CourseId,
		TeeId = s.TeeId,
		FrontPar = s.ParFront,
		BackPar = s.ParBack,
		CoursePar = s.CoursePar,
		DistanceBack = s.DistanceBack,
		DistanceFront = s.DistanceFront,
		DistanceTotal = s.DistanceTotal,
		Name = s.Name,
		Color = s.Color,
		Rating = s.Rating,
		Slope = s.Slope
	}).ToListAsync();
````
With the modified where clauses, the query came back under 100ms but the generated SQL was not exactly what I was hoping for; the `string.IsNullOrEmpty` was not translated at all:
````sql
exec sp_executesql N'SELECT [w].[Name], [w].[CourseId], [w].[TeeId], [w].[ParFront], [w].[ParBack], [w].[CoursePar], [w].[DistanceBack], [w].[DistanceFront], [w].[DistanceTotal], [w].[Color], [w].[Rating], [w].[Slope]
FROM [Tee] AS [w]
WHERE (([w].[CourseId] = @__courseId_0) AND ([w].[IsActive] = 1)) AND ([w].[IsDeleted] = 0)
ORDER BY [w].[Color]',N'@__courseId_0 bigint',@__courseId_0=3959
````
I had to manually recreate the `string.IsNullOrEmpty` method in order to get the results I needed, and thus the correct SQL.
````c#
return await dbContext.Tees
	.Where(w => w.CourseId.Equals(courseId) && (w.Name != null || w.Name != "") && w.IsActive.Equals(true) && w.IsDeleted.Equals(false))
	.OrderBy(o => o.Color)
	.Select(s => new TeeDto
	{
		CourseId = s.CourseId,
		TeeId = s.TeeId,
		FrontPar = s.ParFront,
		BackPar = s.ParBack,
		CoursePar = s.CoursePar,
		DistanceBack = s.DistanceBack,
		DistanceFront = s.DistanceFront,
		DistanceTotal = s.DistanceTotal,
		Name = s.Name,
		Color = s.Color,
		Rating = s.Rating,
		Slope = s.Slope
	}).ToListAsync();
````
````sql
exec sp_executesql N'SELECT [w].[CourseId], [w].[TeeId], [w].[ParFront], [w].[ParBack], [w].[CoursePar], [w].[DistanceBack], [w].[DistanceFront], [w].[DistanceTotal], [w].[Name], [w].[Color], [w].[Rating], [w].[Slope]
FROM [Tee] AS [w]
WHERE ((([w].[CourseId] = @__courseId_0) AND ([w].[Name] IS NOT NULL OR (([w].[Name] <> '''') OR [w].[Name] IS NULL))) AND ([w].[IsActive] = 1)) AND ([w].[IsDeleted] = 0)
ORDER BY [w].[Color]',N'@__courseId_0 bigint',@__courseId_0=3959
````

Am I doing this correctly or for now, the support of translating `string.IsNullOrEmpty` into SQL is not supported yet? I'm using *EF 7.0.0-rc1-final*

### PR DESCRIPTION
**Pull request by smitpatel**
Resolves #5199 
@anpete @divega - Are there anymore translators which we should have?
Currently we have following translators for string
Contains, EndsWith, Equals, StartsWith, Compare, Concat (in relational)
Length, Replace, Substring, ToLower, ToUpper (in sql server)
Adding IsNullOrEmpty in this PR.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -511,7 +511,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                     = methodCallExpression.Arguments
                         .Where(e => !(e is QuerySourceReferenceExpression)
                                     && !(e is SubQueryExpression))
-                        .Select(Visit)
+                        .Select(e => (e as ConstantExpression)?.Value is Array ? e : Visit(e))
                         .Where(e => e != null)
                         .ToArray();
 
--- a/src/Microsoft.EntityFrameworkCore.Sqlite/Query/ExpressionTranslators/Internal/SqliteStringTrimTranslator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Sqlite/Query/ExpressionTranslators/Internal/SqliteStringTrimTranslator.cs
@@ -1,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using Microsoft.EntityFrameworkCore.Query.Expressions;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
+{
+    public class SqliteStringTrimTranslator : IMethodCallTranslator
+    {
+        private static readonly IEnumerable<MethodInfo> _trims = typeof(string).GetTypeInfo()
+            .GetDeclaredMethods(nameof(string.Trim));
+
+        public virtual Expression Translate(MethodCallExpression methodCallExpression)
+        {
+            if (_trims.Contains(methodCallExpression.Method))
+            {
+                var sqlArguments = new List<Expression> { methodCallExpression.Object };
+                var charactersToTrim = methodCallExpression.Arguments.Count == 1
+                    ? (methodCallExpression.Arguments[0] as ConstantExpression)?.Value as char[]
+                    : null;
+                if (charactersToTrim?.Length > 0)
+                {
+                    sqlArguments.Add(Expression.Constant(new string(charactersToTrim), typeof(string)));
+                }
+                return new SqlFunctionExpression("trim", methodCallExpression.Type, sqlArguments);
+            }
+
+            return null;
+        }
+    }
+}
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -4829,6 +4829,104 @@ FROM [Customers] AS [c]",
                 Sql);
         }
 
+        public override void IsNullOrEmpty_in_predicate()
+        {
+            base.IsNullOrEmpty_in_predicate();
+
+            Assert.Equal(
+                @"SELECT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM [Customers] AS [c]
+WHERE [c].[Region] IS NULL OR ([c].[Region] = N'')",
+                Sql);
+        }
+
+        public override void IsNullOrEmpty_in_projection()
+        {
+            base.IsNullOrEmpty_in_projection();
+
+            Assert.Equal(
+                @"SELECT [c].[CustomerID], CASE
+    WHEN [c].[Region] IS NULL OR ([c].[Region] = N'')
+    THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT)
+END
+FROM [Customers] AS [c]",
+                Sql);
+        }
+
+        public override void IsNullOrWhiteSpace_in_predicate()
+        {
+            base.IsNullOrWhiteSpace_in_predicate();
+
+            Assert.Equal(
+                @"SELECT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM [Customers] AS [c]
+WHERE [c].[Region] IS NULL OR (LTRIM(RTRIM([c].[Region])) = N'')",
+                Sql);
+        }
+
+        public override void TrimStart_in_predicate()
+        {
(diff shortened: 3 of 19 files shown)

### CODE REVIEW

File: test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs [hunk 1]
```diff
@@ -4808,6 +4808,30 @@ public override void Substring_with_client_eval()
...
         }
 
+        public override void IsNullOrEmpty_in_predicate()
+        {
+            base.IsNullOrEmpty_in_predicate();
+
+            Assert.Equal(
+                @"SELECT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM [Customers] AS [c]
+WHERE [c].[Region] IS NULL OR ([c].[Region] = N'')",
```
anpete: @divega Any smarter translation? Built-in function etc

anpete:
@smitpatel IsNullOrWhitespace?

smitpatel:
What would `IsNullOrWhitespace` translate to?

anpete:
@smitpatel LTRIM(RTRIM( or PATINDEX - see http://stackoverflow.com/questions/350400/testing-for-whitespace-in-sql-server

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
divega: This looks good actually.

anpete: Agreed

smitpatel:
updated with feedback
functions added are
string.IsNullOrEmpty
string.IsNullOrWhiteSpace
"string".TrimStart
"string".TrimEnd
"string".Trim

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -511,7 +511,7 @@ var arguments
                     = methodCallExpression.Arguments
                         .Where(e => !(e is QuerySourceReferenceExpression)
                                     && !(e is SubQueryExpression))
-                        .Select(Visit)
+                        .Select(e => (e as ConstantExpression)?.Value is Array ? e : Visit(e))
```
anpete: I think we want to visit when the array has elements, no?

src/Microsoft.EntityFrameworkCore.Sqlite/Query/ExpressionTranslators/Internal/SqliteStringTrimTranslator.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        private static readonly IEnumerable<MethodInfo> _trims = typeof(string).GetTypeInfo()
+            .GetDeclaredMethods(nameof(string.Trim));
+
+        public virtual Expression Translate(MethodCallExpression methodCallExpression)
+        {
+            if (_trims.Contains(methodCallExpression.Method))
+            {
+                var sqlArguments = new List<Expression> { methodCallExpression.Object };
+                var charactersToTrim = methodCallExpression.Arguments.Count == 1 ? (methodCallExpression.Arguments[0] as ConstantExpression)?.Value as char[] : null;
```
anpete: Break long lines like this

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
smitpatel: The array is value of `ConstantExpression`. We never visit it regardless of it has values or not. `VisitConstant` function just verifies if the type if supported or not without looking into actual values.

anpete: Should we move the check to VisitConstant?

smitpatel: Methods which takes params args generates ConstantExpression with array values. That is the only case we need to pass the array as is. In all other cases we block array valued ConstantExpression (except byte[]). Moving this to VisitConstant could have unwanted side-effects.

anpete:
:shipit:

ErikEJ:
Providers beware?

smitpatel:
This "Providers-beware" is similarly to the other one. Here we are adding provider-specific method translators. Other providers can write similar but if they don't then it will be evaluated on client. So no breaking change.

Sun5Kong:
@anpete Why "!string.IsNullOrEmpty(r.X)" will translate to "[r].[X] IS NOT NULL AND (([r].[X] <> N'') OR [r].[X] IS NULL)" not"[r].[X] IS NOT NULL AND ([r].[X] <> N'')"?

PR #5256, opened 2016-05-02
Add transitive reference to PlatformAbstractions

ryanbrandenburg:


divega:
Hey @ryanbrandenburg, can you explain why this is needed and also why it is not in release?

ryanbrandenburg:
It's needed because PlatformAbstractions was removed as a transitive dependency from a couple of packages that these packages depend on, and it's causing the CoreCLR gate to fail.

As for being in some branches but not in others, we noticed a problem when I pushed to release, someone had force pushed and @pranavkm lost some commits. He's working on it now and I'll take things back into dev when he's got that all resolved.

ryanbrandenburg:
@pranavkm already merged this in for me.

ryanbrandenburg:
@pranavkm already merged this in for me.

PR #5268, opened 2016-05-04
[TINY] Correct references to an Internal namespace

lajones:
In #5209 I changed the namespaces of a lot of files to add `.Internal`. But I missed the couple of references (by reflection) in the `DesignTimeProviderServices` attributes.


smitpatel:
:shipit:

ErikEJ:
Providers beware?

divega:
@ErikEJ yes, in the sense that provider writers may choose to follow similar namespace layouts, but no in the sense that this shouldn't be a breaking change.

ErikEJ:
Thanks, Diego!

PR #5270, opened 2016-05-05
SQL Server datepart translation support (fixes #4582)

### LINKED ISSUE #4582: Query: Translate DateTime.Month/Day/Year etc.
I want to filter database records:
item.PurchDate.Year==2016&&item.TransactionType==PCTransactionType.Purchase (PCTransactionType is type of Enum)
when I check the log, it shows that these two conditions could not be executed in database server and must be evaluated locally.
Does EF7 support these kind of filtering?

davidroth:
This PR adds sql server translation support for the following DateTime properties in LINQ (See also: https://github.com/aspnet/EntityFramework/issues/4582):
- DateTime.Date
- DateTime.Year
- DateTime.Month
- DateTime.Day
- DateTime.DayOfYear
- DateTime.Day
- DateTime.Hour
- DateTime.Minute
- DateTime.Second
- DateTime.Millisecond

This is SQL server only atm, but I would also be willing to add support for sqlite. 
However, I thought I would gather some feedback on this part first.

Looking forward to your feedback :)

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
@@ -1099,6 +1099,91 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [ConditionalFact]
+        public virtual void Where_datetime_date_component()
+        {
+            var myDatetime = new DateTime(1998, 5, 4);
+            AssertQuery<Order>(
+                oc => oc.Where(o =>
+                o.OrderDate.Value.Date == myDatetime),
+                entryCount: 3);
+        }
+
+
+        [ConditionalFact]
+        public virtual void Where_datetime_year_component()
+        {
+            AssertQuery<Order>(
+               oc => oc.Where(o =>
+               o.OrderDate.Value.Year == 1998),
+               entryCount: 270);
+        }
+
+        [ConditionalFact]
+        public virtual void Where_datetime_month_component()
+        {
+            AssertQuery<Order>(
+                oc => oc.Where(o =>
+                o.OrderDate.Value.Month == 4),
+                entryCount: 105);
+        }
+
+        [ConditionalFact]
+        public virtual void Where_datetime_dayOfYear_component()
+        {
+            AssertQuery<Order>(
+                oc => oc.Where(o =>
+                o.OrderDate.Value.DayOfYear == 68),
+                entryCount: 3);
+        }
+
+
+        [ConditionalFact]
+        public virtual void Where_datetime_day_component()
+        {
+            AssertQuery<Order>(
+              oc => oc.Where(o =>
+              o.OrderDate.Value.Day == 4),
+              entryCount: 27);
+        }
+
+        [ConditionalFact]
+        public virtual void Where_datetime_hour_component()
+        {
+            AssertQuery<Order>(
+              oc => oc.Where(o =>
+              o.OrderDate.Value.Hour == 14),
+              entryCount: 0);
+        }
+
+
+        [ConditionalFact]
+        public virtual void Where_datetime_minute_component()
+        {
+            AssertQuery<Order>(
+              oc => oc.Where(o =>
+              o.OrderDate.Value.Minute == 23),
+              entryCount: 0);
+        }
+
+        [ConditionalFact]
+        public virtual void Where_datetime_second_component()
+        {
+            AssertQuery<Order>(
+              oc => oc.Where(o =>
+              o.OrderDate.Value.Second == 44),
+              entryCount: 0);
+        }
+
+        [ConditionalFact]
+        public virtual void Where_datetime_millisecond_component()
+        {
+            AssertQuery<Order>(
+              oc => oc.Where(o =>
+              o.OrderDate.Value.Millisecond == 88),
+              entryCount: 0);
+        }
+
+        [ConditionalFact]
         public virtual void Where_simple_reversed()
         {
             AssertQuery<Customer>(
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerDateTimeDatePartComponentTranslator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerDateTimeDatePartComponentTranslator.cs
@@ -1,0 +1,42 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq.Expressions;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
+{
+    public class SqlServerDateTimeDatePartComponentTranslator : IMemberTranslator
+    {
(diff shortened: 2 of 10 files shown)


src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -1099,6 +1099,91 @@ public virtual void Where_datetime_utcnow()
...
+              entryCount: 27);
+        }
+
+        [ConditionalFact]
+        public virtual void Where_datetime_hour_component()
+        {
+            AssertQuery<Order>(
+              oc => oc.Where(o =>
+              o.OrderDate.Value.Hour == 14),
+              entryCount: 0);
```
ErikEJ: Why entryCount: 0 ?

anpete: No results.

anpete:
@davidroth Thanks! This looks really good.

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
davidroth: Yep, its because none of the existing orders in the [Northwind](https://github.com/aspnet/EntityFramework/blob/dev/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/Northwind.sql)  test database have a timepart other than '00:00:00'.
Since the "real" testcases verify the generated SQL anyway, I thought it is not worth to add additional data just to get an entryCount > 0.

ErikEJ: You could modify Northwind.sql to have abnon-zero datetime order?

davidroth: Sure. I will push an update as soon as I find some time.

anpete: I think it is fine as is. We don't typically modify Northwind.sql - we would usually create a separate fixture for specific data requirements.

davidroth: Fine, less work for me :smiley:

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerDateTimeDatePartComponentTranslator.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
+{
+    public class SqlServerDateTimeDatePartComponentTranslator : IMemberTranslator
+    {
+        private readonly Dictionary<string, DatePart> _supportedParts = new Dictionary<string, DatePart>
```
ilmax: Just asking, does we need all those Dictionary to keep mappings? they are not static, so each time SqlServerDateTimeDatePartComponentTranslator  is newed up, this dictionary should be created as well.

What about converting dictionaries to a method?

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerDateTimeDatePartComponentTranslator.cs
```diff
@@ -0,0 +1,39 @@
...
+         {
+            { nameof(DateTime.Year), DatePart.Year },
+            { nameof(DateTime.Month), DatePart.Month },
+            { nameof(DateTime.DayOfYear), DatePart.DayOfYear },
+            { nameof(DateTime.Day), DatePart.Day },
+            { nameof(DateTime.Hour), DatePart.Hour },
```
ilmax: Mind you adding support for DateTime.Date as well?

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerDateTimeDatePartComponentTranslator.cs
divega: Interesting point. I am not sure if this is worth creating a dictionary (and a second one for the SQL generation) for this each time. A virtual method with a switch statement or even a static dictionary could be ok. @anpete What do yo think?

smitpatel: What about `DatePart` enum? Do we actually need it? Based on member name here we put `DatePart` value in `DatePartExpression` and the `SqlServerQueryGenerator` uses another dictionary to convert `DatePart` to string. I think we can use the string in the `DatePartExpression` which `QueryGenerator` would use it directly (without going through enum)

Also the string mapped are same as the name of member. We can use a list of supported parts here or switch statement for all supported parts, which would use `nameof` on `Member.Name` to put appropriate string on `DatePartExpression`

divega: Sounds good to me. I would probably still encapsulate the mapping currently implemented with `TryGetValues()`  into a method to facilitate extending  the list, but the implementation could be right passing the name through for all  currently supported parts.

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerDateTimeDatePartComponentTranslator.cs
davidroth: [DatePart ](https://msdn.microsoft.com/de-de/library/ms174420(v=sql.120).aspx)can only extract integral values from a date instance, but cannot convert a datetime into date. 
This can be done via `CAST` or `CONVERT` and I have implemented this in this PR via via `SqlServerDateTimeDateComponentTranslator.cs`

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerDateTimeDatePartComponentTranslator.cs
davidroth: Sounds good to me as well. I will remove the `DatePart` enum plus both dictionaries and implement it via `switch` instead.

davidroth: Changes pushed - no more dictionaries, no more enums :smile:

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerDateTimeDatePartComponentTranslator.cs
ilmax: Sorry @davidroth, totally missed that!

divega:
Thanks @davidroth for a great contribution to EF Core!

Issue #1378, opened 2015-01-09
Ensure PropertyChanging and PropertyChanged events are unsubscribed


ajcvickers:
This may have to be coupled to disposing the context, but it would be good to find a way to do it such that un-disposed contexts do not leak.

PR #5283, opened 2016-05-06
Unsubscribe from notification events

ajcvickers:
Issue #1378

This change ensures that entities are unsubscribed from notification events when they stop being tracked or the context is disposed.


anpete:
:shipit:

PR #5286, opened 2016-05-06
CLI commands on desktop: resolve startup assembly from external project

natemcmaster:
Partial fix for https://github.com/aspnet/EntityFramework/issues/4577

This is the minimal change we can make that will enable most users to use dotnet-ef for desktop .NET. Although the implementation will other issues for some users, I believe this is a good tradeoff to the high cost of fixing #4577 perfectly.

Why it's not perfect (simplifed): this tries to pull a "child" assembly into a "parent" app domain. The child may have dependencies that conflicts with the "parent" app domain. Should this problem arise (which may not for most users), the solution is to lift dependencies in the project.json for the "child" assembly.

The "proper" way to do this (the way we do it in Powershell) is to only ever load the "child" in an appdomain with its own dependencies, thus avoiding potential for conflict. But implementing this would mean re-working huge parts of CLI commands.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Tools.Cli/OperationExecutor.cs
@@ -91,7 +91,8 @@ namespace Microsoft.EntityFrameworkCore.Tools.Cli
 #endif
             }
 
-            var startupAssemblyName = startupProjectContext.ProjectFile.GetCompilerOptions(startupFramework, startupConfiguration).OutputName;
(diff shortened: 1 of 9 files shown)


natemcmaster:
cc @rowanmiller @divega

natemcmaster:
Ping for review.

divega:
LGTM, but let's chat and find out if others are ok with the trade offs.

natemcmaster:
Closed. Let's solve this right rather than workaround limitations. https://github.com/aspnet/EntityFramework/issues/5311

Issue #3376, opened 2015-10-08
Support self-referencing overlapping FK/PK (since generated value does not need to be propagated)


waterfallbay:
I have a ModifierGroupHeader which has a foreign key 
.ForeignKey(x => new { x.LinkedGroupHeaderId, x.AccountId, });
pointing to itself with self reference.

An exception of type 'System.InvalidOperationException' occurred in EntityFramework.Core.dll but was not handled in user code

Additional information: A circular dependency was detected: 'ModifierGroupHeader' {'LinkedGroupHeaderId', 'AccountId'} -> 'ModifierGroupHeader' {'GroupHeaderId', 'AccountId'}.

My EF7 is written as follows:
```C#
modelBuilder.Entity<ModifierGroupHeader>()
                            .Reference(x => x.ModifierGroupHeader2)
                            .InverseCollection(x => x.ModifierGroupHeader1)
                            .ForeignKey(x => new { x.LinkedGroupHeaderId, x.AccountId, });
```

rowanmiller:
__EF Team Triage:__ This issue is lacking enough information for us to be able to effectively triage it. For a guide on submitting good bug reports, read_ [Painless Bug Tracking](http://www.joelonsoftware.com/articles/fog0000000029.html). _The most important elements to help us act on a bug report are example code/steps that reproduce the issue, the outcome/results you are expecting, and the results or error message that you are receiving. Following are a few questions to help us understand this bug report, for future issues please try to include this info in the initial report._

Could you share a complete code listing (or a project) that we can use to reproduce and debug the issue you are seeing?

smitpatel:
As the exception message says - the property `AccountId` is part of both principal key and foreign key. Not sure if that is a valid relationship.

smitpatel:
discussed with @AndriySvyryd  about this. It is cycle so we are throwing correct exception but since there is only one node in the cycle we could allow it too. Though not required at present.

Example model
````
public class Node
    {
        public int ListId { get; set; }
        public int PreviousNodeId { get; set; }
        public int NextNodeId { get; set; }

        public Node PreviousNode { get; set; }
        public Node NextNode { get; set; }
    }

modelBuilder.Entity<Node>(b =>
{
    b.HasKey(e => new {e.ListId, e.PreviousNodeId});
    b.HasOne(e => e.NextNode)
        .WithOne(e => e.PreviousNode)
        .ForeignKey<Node>(e => new {e.ListId, e.NextNodeId});
});
````

[...]

PR #5287, opened 2016-05-06
Regression tests for overlapping FK/PK

ajcvickers:
Issue #3376

Added tests for both the customer model (1:*) and the model Smit added (1:1). Both work with the current bits.


anpete:
:shipit:

Issue #4425, opened 2016-01-28
Redesign SqlServerTypeMapper to provide better round-trip mappings?


lajones:
See #4348. When trying to update scaffolding to make better use of the `SqlServerTypeMapper` code it was difficult to re-use because the mappings conflate the ideas of synonyms (which result in identical SQL Server data types in the database) with the idea of CLR-types which, by default, map to the same SQL Server data type e.g. see the mapping of `image` to `_varbinary`.

It also conflates in the idea of a max length qualification. And when it's available (see #3420) will also conflate in the idea of whether a column is a Unicode column as well. For scaffolding purposes (going from SQL Server metadata to generated code) it would be good for this code to provide an API which can be passed a set of attributes (say an unqualified SQL Server data type, e.g. "varchar", and the max length, e.g. 30) and return that the full column type is "varchar(30)" but that in order to implement that we need the fluent API `.HasMaxLength(30).IsUnicode(false)` i.e. we don't, and must not, specify `HasColumnType()`. Other data types would need different combinations of those APIs.

Migrations (going from code to generated SQL Server metadata) need to know e.g. that the default SQL Server data type for a string column is `nvarchar(max)`. But scaffolding needs to know that too in order that it _not_ to generate fluent API if that is what would be assumed by default anyway.

We aim for scaffolding to generate code from which migrations would generate the same SQL Server metadata i.e. to make it "round-trippable". `SQLServerTypeMapper` seems like a good place to centralize this but looks like it needs a redesign to make that work. We would also need to figure out how much code that is useful to scaffolding could also be used for more general purposes.

divega:
@ajcvickers I had this assigned to me I believe with the purpose of giving it some thought. Re-reading it today I am not sure how much of it is still necessary. While I tend to agree that it could be worth to centralize the knowledge of how CLR types + facets map to database types for a specific database provider, I believe having logic that is runtime specific and design time specific in separate places is a reasonable compromise. 

Could you please take a look at this and provide your thoughts on whether we should have this in the backlog, do something about it before RTM or close it?

divega:
[...]

Issue #4134, opened 2015-12-18
Use column/property facets for parameter types in Update Pipeline


michaelpaulus:
### Title
HasMaxLength / HasColumnType / ForSqlServerHasColumnType don't produce parameters with that size

### Functional impact
Data Types on generated parameters don't match actual sql data types and don't work as expected.  This can lead to indexes not being used properly due to size mismatches between parameters and columns.

### Minimal repro steps

1. Create an EF 7 Project with Sql Server
2. Add a model with a key and a string property
3. Configure the model's string property to use HasMaxLength(200) or HasColumnType("nvarchar(200)") or ForSqlServerHasColumnType("nvarchar(200)")
4. Query against the string column in the model (e.g. where stringcolumn == strinvalue)
5. Look at sql server profiler and see that the parameter is created with nvarchar(4000) not nvarchar(200)
6. Update the string column as savechanges 
7. Look at the sql server profile and see that the parameter is created with nvarchar(4000) not nvarchar(200)

### Expected result
Parameter is created with the size specified by HasMaxLength or HasColumnType

### Actual result
Parameter is always being created with the _maxSpecificSize in Microsoft.Data.Entity.Storage.Internal.SqlServerMaxLengthMapping

### Further technical details
There is a comment in the method Microsoft.Data.Entity.Storage.Internal.SqlServerMaxLengthMapping.ConfigureParameter 

// For strings and byte arrays, set the max length to 8000 bytes if the data will
// fit so as to avoid query cache fragmentation by setting lots of different Size
// values otherwise always set to -1 (unbounded) to avoid SQL client size inference.

I don't agree with this comment if the user is going through the trouble of telling you what size to use.

ErikEJ:
Do you have evidence of indexes not being used?

michaelpaulus:
I'm sorry, I was going off what I learned about db design, not an actual use case.  I cannot repo the issue and after research (http://stackoverflow.com/questions/14342954/ado-net-safe-to-specify-1-for-sqlparameter-size-for-all-varchar-parameters) the functional impact seems to be wrong.  If the developer goes through the trouble of telling you the parameter size I think you should use it, instead of 4000 / 8000.  This will still avoid query cache fragmentation but produce the query that the developer expects.

Thanks.

ErikEJ:
But HasMaxLength does not define a parameter size, but a storage max size.

divega:
[...]

PR #5310, opened 2016-05-09
Generalize type mapping support for Unicode

ajcvickers:
In looking at #4425 and #4134 in became quickly apparent that the support for inferring Unicode was too specific to Unicode. What is really needed is that if the type mapping to be used can be inferred in some way, then this type mapping should be used for all relevant facets, not just Unicode. Only in cases where no inference is possible should the default mapping for the CLR type be used.


anpete:
:shipit:

Issue #4795, opened 2016-03-14
Query: Client-eval when query parameterized on weakly typed array access.


anpete:
Northwind model
Query:
```c#
data.Set<Customer>().SingleAsync(c => c.CustomerId == (string)context.Arguments["customerId"]));
```
Results:
```sql
SELECT [c].[CustomerId], [c].[CompanyName], [c].[ContactName]
FROM [Customers] AS [c]
```

PR #5326, opened 2016-05-10
Fix: #4795 - Makes method parameterization more aggressive by conside…

anpete:
…ring instance methods.

@maumar


maumar:
:shipit:

PR #5330, opened 2016-05-11
Fix #5327 - use backward-compatible powershell API in EF commands

### LINKED ISSUE #5327: Modify InvokeDotnetEf in EntityFramework.psm1 to use .Path instead of .Source to discover dotnet location
According to https://msdn.microsoft.com/en-us/library/system.management.automation.applicationinfo_members(v=vs.85).aspx, the `.Source` property on `ApplicationInfo` returned by `Get-Command` was added in Powershell 5.0. `.Path` returns the same value and works in powershell 3.0.

Line of code to change: https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFramework.psm1#L643

natemcmaster:
Also helps DNX project users who upgrade.

cc @rowanmiller @divega @pranavkm

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFramework.psm1
+++ b/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFramework.psm1
@@ -640,7 +640,11 @@ function ShowConsole {
 }
 
 function InvokeDotNetEf($project, [switch] $json, [switch] $skipBuild) {
-    $dotnet = (Get-Command dotnet).Source
+    try {
+        $dotnet = (Get-Command dotnet).Path
+    } catch {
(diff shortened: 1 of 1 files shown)


divega:
LGTM. Not super important, but should we just say *.xproj projects to be more specific?

natemcmaster:
As per https://github.com/aspnet/EntityFramework/issues/5327#issuecomment-218527404, I will merge this change to dev, not release.

@divega because of this, not sure the *.xproj change would apply.

natemcmaster:
Ping for review.

lajones:
:shipit:

natemcmaster:
Merged to dev in https://github.com/aspnet/EntityFramework/commit/31abba7391bc43656a91571aa069edd88f65ca2d

PR #5340, opened 2016-05-11
Preserve removed navigation PropertyInfo

AndriySvyryd:


lajones:
The change looks good but could you add something to say why you are doing this - what problem does it solve? And can you add a test?

src/Microsoft.EntityFrameworkCore/Metadata/Internal/ForeignKey.cs
```diff
@@ -213,14 +213,16 @@ public virtual void UpdatePrincipalToDependentConfigurationSource(ConfigurationS
                         DeclaringEntityType.Model.ConventionDispatcher.OnNavigationRemoved(
                             DeclaringEntityType.Builder,
                             PrincipalEntityType.Builder,
-                            oldNavigation.Name);
+                            oldNavigation.Name,
+                        	oldNavigation.PropertyInfo);
```
smitpatel: tabbing

smitpatel:
@lajones - From what I understand, this change stops `RelationalshipDiscoveryConvention` to run while a shadow navigation is removed because convention would not find anything new on CLR type.

AndriySvyryd:
@lajones @smitpatel is correct. This is a perf fix. By itself it doesn't improve much, but it's a prerequisite for #5358

PR #5341, opened 2016-05-11
Fix #5332 - Log warning when redundant includes detected.

### LINKED ISSUE #5332: Query: Warn when Includes are ignored
For example,  in the following query the Include/ThenInclude operators are saying "when you return  `CarPart`s in the query, include the relates  `Car`s and their related  `Driver`s". Because the query is then changed to return  `Car`s rather than  `CarPart`s, this is ignored. We should write a warning that the includes are being ignored.

```c#
db.CarParts
    .Include(cp => cp.Car).ThenInclude(c => c.Drivers)
    .Select(cp => cp.Car)  
```

anpete:
@maumar


maumar:
:shipit:

Issue #5238, opened 2016-05-01
missing "using System;" in context class


tdjastrzebski:
version: 1.0.0-rc1-final

SQL Server DB has a table with uniqueidentifier (Guid) column which has a default value "00000000-0000-0000-0000-000000000000".

Proper Fluent API gets generated:
`entity.Property(e => e.PrintFormatListId).HasDefaultValue(new Guid("00000000-0000-0000-0000-000000000000"));`
but lack of "using System;" in the file header causes compiler error CS0246: The type or namespace name 'Guid' could not be found,

lajones:
Fxied with commit 4e27d5a.

PR #5348, opened 2016-05-12
[SMALL] Generated DbContext classes need "using System;"

lajones:
Fix for #5238. One file is the fix - the rest are just updating the expected results.


lajones:
Put `providers-beware` on this because if you have tests which check the contents of generated files for your providers you will have an extra `using System;` line.

smitpatel:
Is there a way to find out only necessary imports? I think Migration files use that approach.

ErikEJ:
Thanks, @lajones

natemcmaster:
:shipit:

@smitpatel we could try to detect what we need, but this simple fix is simple IMO. Migrations has all that detection b/c migration scaffolding may pull in user types, but dbcontext scaffolding doesn't rely on existing user code.

lajones:
Agree with @natemcmaster - trying to keep it simple. Checked in with commit 4e27d5a.

Issue #5267, opened 2016-05-04
dotnet-ef scaffold references old EF package name


pranavkm:
Usage: dotnet ef dbcontext scaffold [arguments] [options]
Arguments:
  [connection]  The connection string of the database
  [provider]    The provider to use. For example, **EntityFramework.MicrosoftSqlServer**

lajones:
Fixed with commit 93c1635.

PR #5349, opened 2016-05-12
[TINY] Correcting bad example provider name.

lajones:
Fix for #5267. Example had the old provider package name. Updated to the new one.


divega:
:shipit:

lajones:
Checked in with commit #93c1635.

PR #5355, opened 2016-05-13
Updates to relational type mapping

ajcvickers:
These changes address issue #4425 while at the same time re-introducing the parsing of sizes in store type names such that these can then be used in parameters for #4134.

The mapping of strings and binary types based on facets has been split out such that the information provided by the type mapper can be interpreted more easily by the scaffolding code. This also simplifies the type mapper APIs themselves. A new class ScaffoldingTypeMapper has been introduced that can be used to determine what needs to be scaffolded for a given store type. This has not yet been integrated into scaffolding, but has unit tests to check it gives the correct answers. Scaffolding must provide this service with the full store type name (e.g. nvarchar(256)" together with information about whether the property will be a key or index or a rowversion. The service then gives back data indicating whether the type is inferred (need not be explicitly set) and, if so, whether max length and/or unicode factes needs to be set.


ajcvickers:
Ping.

src/Microsoft.EntityFrameworkCore.Relational/Storage/ByteArrayRelationalTypeMapper.cs
```diff
@@ -0,0 +1,59 @@
...
+                return RowVersionMapping;
+            }
+
+            var defaultMapping = keyOrIndex && KeyMapping != null ? KeyMapping : DefaultMapping;
+
+            if (size.HasValue
+                && size != defaultMapping.Size)
+            {
+                return size <= MaxBoundedLength
+                    ? _boundedMappings.GetOrAdd(size.Value, l => CreateBoundedMapping(l))
```
AndriySvyryd: Use method group instead of lambda

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalTypeMapper.cs
```diff
@@ -54,92 +42,93 @@ public virtual RelationalTypeMapping FindMapping(Type clrType)
...
+        public virtual RelationalTypeMapping FindMapping(string storeType)
         {
-            Check.NotNull(typeName, nameof(typeName));
+            Check.NotNull(storeType, nameof(storeType));
 
-            RelationalTypeMapping mapping;
-            return GetSimpleNameMappings().TryGetValue(typeName, out mapping)
-                ? mapping
-                : null;
+            return _explicitMappings.GetOrAdd(storeType, t => CreateMappingFromStoreType(storeType));
```
AndriySvyryd: Use method group

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalTypeMapper.cs
```diff
@@ -54,92 +42,93 @@ public virtual RelationalTypeMapping FindMapping(Type clrType)
...
 
-        protected virtual RelationalTypeMapping FindCustomMapping([NotNull] IProperty property) => null;
+            var closeParen = storeType.IndexOf(")", openParen + 1, StringComparison.Ordinal);
+            int size;
+            if (closeParen > openParen
+                && int.TryParse(storeType.Substring(openParen + 1, closeParen - openParen - 1), out size)
+                && mapping.Size != size)
+            {
+                return mapping.CreateCopy(storeType, size);
+            }
```
AndriySvyryd: Wrap this block in `if (openParen > 0)`

test/Microsoft.EntityFrameworkCore.SqlServer.Design.Tests/ScaffoldingTypeMapperSqlServerTest.cs
```diff
@@ -0,0 +1,298 @@
...
+
+using System;
+using Microsoft.EntityFrameworkCore.Scaffolding;
+using Microsoft.EntityFrameworkCore.Storage.Internal;
+using Xunit;
+
+namespace Microsoft.EntityFrameworkCore.SqlServer.Design
+{
+    public class ScaffoldingTypeMapperSqlServerTest
+    {
```
AndriySvyryd: Consider adding a test for with an uppercase string

AndriySvyryd:
:shipit:

Issue #3101, opened 2015-09-11
.Include() with Self-Join / Coalesce


brent-russell:
Beta 7 and Beta 6.

This might encompass two separate issues. The Include() method is problematic with the following query:

```C#
var query = from eVersion in _dbContext.Entities
			join eRoot in _dbContext.Entities
				on eVersion.RootEntityId.GetValueOrDefault() equals eRoot.Id
			into RootEntities
			from eRootJoined in RootEntities.DefaultIfEmpty()  // left outer join
			select eRootJoined ?? eVersion;
```

If `query = query.Include(e => e.Children);` is used, the coalesce expression `eRootJoined ?? eVersion` cannot be bound.

If `Include` is added to the second or both of the `_dbContext.Entities` lines, then a null reference exception occurs.

If `Include` is added to the first `_dbContext.Entities` line, the query runs, but does not include the children.

```C#
public class Entity
{
	public Entity()
	{
		this.Children = new Collection<Child>();
	}

	public int Id { get; set; }

	public int? RootEntityId { get; set; }

	public Entity RootEntity { get; set; }

	public ICollection<Child> Children { get; set; }
}
```

Where RootEntityId is a self-referential relation on the Entity table.

maumar:
@brent-russell  could you also share a model you are using?

brent-russell:
@maumar edited original comment to include model info.

maumar:
poaching this

maumar:
blocked by #4547

maumar:
blocked by #4925

maumar:
clearing milestone for retriage

rowanmiller:
@maumar where are we at with this issue?

rowanmiller:
@maumar ping 😄

maumar:
@rowanmiller fix is almost ready

maumar:
Fixed in 7bdf5818b10389588d74ecc03ba766fe24850298

PR #5356, opened 2016-05-13
Fix to #3101 - .Include() with Self-Join / Coalesce

maumar:
Problem was that for include we were using relinq's EntityAccessorFindingVisitor, to extract a accessor lambda to a given query source in the selector (for InMemory scenario). Problem was that EntityAccessorFindingVisitor was not meant to work for some cases that we were using it (e.g. Coalesce, Conditional).

Fix is to refactor InMemory include mechanisms so that we no longer need the accessor lambda. Now Include method is applied either around appropriate elements of the outer-most selector, or in case no Select method has been generated - around EntityQuery/OfType directly.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
@@ -62,49 +62,157 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
         protected override void IncludeNavigations(
             IncludeSpecification includeSpecification,
             Type resultType,
-            LambdaExpression accessorLambda,
+            Expression accessorExpression,
             bool querySourceRequiresTracking)
         {
             Check.NotNull(includeSpecification, nameof(includeSpecification));
             Check.NotNull(resultType, nameof(resultType));
-            Check.NotNull(accessorLambda, nameof(accessorLambda));
 
-            MethodInfo includeMethod;
+            var includeExpressionVisitor = new InMemoryIncludeExpressionVisitor(
+                _materializerFactory,
+                QueryCompilationContext,
+                LinqOperatorProvider,
+                includeSpecification,
+                accessorExpression,
+                querySourceRequiresTracking);
 
-            var resultItemTypeInfo = resultType.GetTypeInfo();
+            Expression = includeExpressionVisitor.Visit(Expression);
+        }
 
-            if (resultItemTypeInfo.IsGenericType
-                && (resultItemTypeInfo.GetGenericTypeDefinition() == typeof(IGrouping<,>)
-                    || resultItemTypeInfo.GetGenericTypeDefinition() == typeof(IAsyncGrouping<,>)))
+        private sealed class InMemoryIncludeExpressionVisitor : ExpressionVisitorBase
+        {
+            private readonly IncludeSpecification _includeSpecification;
+            private readonly IMaterializerFactory _materializerFactory;
+            private readonly QueryCompilationContext _queryCompilationContext;
+            private readonly ILinqOperatorProvider _linqOperatorProvider;
+            private readonly Expression _accessorExpression;
+            private readonly bool _querySourceRequiresTracking;
+
+            public InMemoryIncludeExpressionVisitor(
+                [NotNull] IMaterializerFactory materializerFactory,
+                [NotNull] QueryCompilationContext queryCompilationContext,
+                [NotNull] ILinqOperatorProvider linqOperatorProvider,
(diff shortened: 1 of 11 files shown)


maumar:
@anpete

src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
```diff
@@ -57,54 +58,175 @@ public class InMemoryQueryModelVisitor : EntityQueryModelVisitor
             Check.NotNull(materializerFactory, nameof(materializerFactory));
 
             _materializerFactory = materializerFactory;
+            _includeExpressionVisitor = new InMemoryIncludeExpressionVisitor(
```
anpete: DI this?

anpete:
@maumar Nice test coverage!

src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
anpete: Oh, I see this is a nested class. Does it need to be a field, or should we just create a new instance each time (we do this in other places)

src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
```diff
@@ -57,54 +58,175 @@ public class InMemoryQueryModelVisitor : EntityQueryModelVisitor
...
             }
-            else
+
+            public Expression InjectIncludeExpression(
+                [NotNull] Expression expression,
+                [NotNull] IncludeSpecification includeSpecification,
+                [NotNull] Expression accessorExpression,
+                bool querySourceRequiresTracking)
+            {
+                Check.NotNull(expression, nameof(expression));
```
anpete: Don't need checks for private surface.

src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
```diff
@@ -57,54 +58,175 @@ public class InMemoryQueryModelVisitor : EntityQueryModelVisitor
...
+                Expression, 
+                includeSpecification, 
+                accessorExpression, 
+                querySourceRequiresTracking);
+        }
 
-            var resultItemTypeInfo = resultType.GetTypeInfo();
+        private sealed class InMemoryIncludeExpressionVisitor : ExpressionVisitorBase
+        {
+            private IncludeSpecification _includeSpecification;
```
anpete: Do a R# pass to uncover fields that should be readonly

src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
```diff
@@ -139,21 +261,57 @@ public void Dispose()
...
         [UsedImplicitly]
-        private static IEnumerable<TResult> Include<TResult>(
+        private static TEntity Include<TEntity>(
+            QueryContext queryContext,
+            TEntity source,
+            IncludeSpecification includeSpecification,
+            IReadOnlyList<IRelatedEntitiesLoader> relatedEntitiesLoaders,
+            bool querySourceRequiresTracking)
+        {
+            if (includeSpecification.IsEnumerableTarget)
```
anpete: Worth moving this condition to compile time?

src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
maumar: yeah i will do the R# pass before checking this in - got the vm almost set up:)

src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
maumar: it doesn't need to be a field anymore - in one of the early iterations I was caching navigations that have already been included, but with the newer design it is no longer needed - will change

src/Microsoft.EntityFrameworkCore.InMemory/Query/Internal/InMemoryQueryModelVisitor.cs
maumar: will give it a try

maumar:
@anpete new version up

anpete:
:shipit:

PR #5358, opened 2016-05-13
Improve RelationshipDiscoveryConvention perf by keeping track of undiscovered properties

AndriySvyryd:
This improves BuildModel_AdventureWorks time ~5x

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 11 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
```diff
@@ -356,25 +366,114 @@ private void ApplyOnRelatedEntityTypes(EntityType entityType)
...
+            {
+                return true;
+            }
 
-            return propertyInfo.FindCandidateNavigationPropertyType(clrType => clrType.IsPrimitive());
+            navigationCandidates = navigationCandidates.Remove(candidate);
+            SetNavigationCandidates(entityTypeBuilder, navigationCandidates);
+            Apply(entityTypeBuilder);
+            return entityTypeBuilder.IsIgnored(ignoredMemberName, ConfigurationSource.Convention);
         }
```
smitpatel: Why is `IsIgnored` called here?

smitpatel:
:shipit:

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
AndriySvyryd: The return value of the conventions determine whether the next convention should run for this event. Since this runs when a member is ignored then we shouldn't run any other conventions (if there were any) if the member is no longer ignored (in case adding a relationship triggered another convention that unignored the member)

Issue #5347, opened 2016-05-12
Query: String concatenation on a nav prop fails


AlexBilonog:
I use EntityFrameworkCore version: 1.0.0-rc3-20831, .Net Framework: 4.6, AspNet.Mvc: 5.2.3
Code:

`var fullNames = _dbContext.Set<UserRole>().Select(r => r.User.FirstName + " " + r.User.LastName).ToList();`

Throws:
Exception thrown: 'System.InvalidOperationException' in mscorlib.dll
Additional information: The binary operator Add is not defined for the types 'System.String' and 'System.String'.
   at System.Linq.Expressions.Expression.GetUserDefinedBinaryOperatorOrThrow(ExpressionType binaryType, String name, Expression left, Expression right, Boolean liftToNull)
   at System.Linq.Expressions.Expression.Add(Expression left, Expression right, MethodInfo method)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.VisitBinary(BinaryExpression node)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.VisitBinary(BinaryExpression node)
   at Remotion.Linq.Clauses.SelectClause.TransformExpressions(Func2 transformation)
   at Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.Rewrite(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.OptimizeQueryModel(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.CreateQueryExecutor[TResult](QueryModel queryModel)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.<>c__DisplayClass19_01.<CompileQuery>b__0()
   at Microsoft.EntityFrameworkCore.Query.Internal.CompiledQueryCache.GetOrAddQuery[TResult](Object cacheKey, Func1 compiler)
[...]

PR #5359, opened 2016-05-13
[TINY] fix to #5347 - Query: Support translation of string concatenation

maumar:


maumar:
@anpete

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -332,7 +332,7 @@ var rightNavigationJoin
                 }
             }
 
-            return Expression.MakeBinary(node.NodeType, newLeft, newRight);
+            return Expression.MakeBinary(node.NodeType, newLeft, newRight, node.IsLiftedToNull, node.Method);
```
anpete: Can you do a quick check for any other instances of this?

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -5348,6 +5348,18 @@ public override void Environment_newline_is_funcletized()
                 Sql);
         }
 
+        public override void String_concat_with_navigation()
+        {
+            base.String_concat_with_navigation();
+
+            Assert.Equal(
+                @"SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate], [o.Customer].[CustomerID], [o.Customer].[Address], [o.Customer].[City], [o.Customer].[CompanyName], [o.Customer].[ContactName], [o.Customer].[ContactTitle], [o.Customer].[Country], [o.Customer].[Fax], [o.Customer].[Phone], [o.Customer].[PostalCode], [o.Customer].[Region], [o].[CustomerID] + N' '
```
divega: I had a chat with @maumar because this translation seemed strange, even if we say we will only do client side evaluation.

divega: He will look a bit more at it.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
anpete: Is node.Update better here?

maumar: There are few instances where we do equality, that the method is not preserved - we only need that for Add, right? or should we do this across the board, just to be safe?

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
anpete: Good!

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
anpete: I think we need to do it across the board to avoid more of these potential errors. node.Update seems safest (when we can use it)

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
maumar: the reason for strange query is that there is optimization in nav prop, which caused o.Customer.CustomerID to be simplified to o.CustomerID, which doesn't require access to LOJ-ed Customer table and can be translated directly into sql. Modified the test so that now we concatenate two properties requiring nav prop, and query is as expected.

maumar:
@anpete @divega  new version up

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
anpete: But why aren't we getting the projection optimization here? (no materialization required).

divega: If I understood what @maumar said to me, when we do client evaluation we project everything. Assuming that is true I am ok with saying that the spurious `[o].[CustomerID] + N' '` here is just a corner case, but I would like us to consider a separate improvement to only include things we end up binding to in the projection when we do client evaluation.

anpete: OK, I'm assuming the client-eval here is because of DefaultIfEmpty, @maumar?

Seems like a bug in the projection translation that we gen the [o].[CustomerID] + N' '

divega: > OK, I'm assuming the client-eval here is because of DefaultIfEmpty, @maumar?

@maumar said yes when I asked the same question :smile:

> Seems like a bug in the projection translation that we gen the [o].[CustomerID] + N' '

Yeah, maybe. But FWIW, it seems that part is for `o => o.CustomerID + " " ` so no it is not affected by the nav prop translation.

anpete: @divega Yeah, something seems wrong if we are both materializing that Order, but still pushing down part of the concatenation expression. Perhaps it won't matter once we start pushing down the DefaultIfEmpty.

maumar: In RelationalProjectionExpressionVisitor first we visit the entire selector and determine that we need client eval, then we defer to base implementation, which in turn visits the left (o.CustomerID + " ") and right (o.Customer.City). The left side is translatable to SQL and so it is automatically added to the SelectExpression. Then we visit the right side (which can't be translated), but the unnecessary SQL fragment has already been added. 

@anpete should we add logic that if the client eval in needed for the projection, we never add anything to the underlying SQL?

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
maumar: @anpete I switched back to Expression.MakeBinary in some cases because update doesn't work. Basically if the old expression was reference comparison but the new expression is value comparison, the update will throw. (we do that in navigation rewriting and entity equality rewriting). Will keep expression.Update anywhere else

anpete:
:shipit:

Issue #3987, opened 2015-12-06
Identifier generation strategy in reverse engineering


roji:
I can see logic in CSharpUtilities that's responsible for generating a C# identifier out of a database column. In PostgreSQL it seems that the standard case is "snake case" (employee_name). It seems right for the reverse engineering process to convert this to camel case (EmployeeName).

So I was wondering if this is something that Npgsql should do in its own version of CSharpUtilities, or whether the universal CSharpUtilities should detect snake case and perform this task (I'm guessing there are instances where SqlServer or other databases contain snake case).

natemcmaster:
This is a good question. The identifier generation needs to be revisited. 

**Currently** [`RelationalScaffoldingModelFactory.GetEntityName`](https://github.com/aspnet/EntityFramework/blob/083fc56455bfd7fb919f30c284cab272b1272c2e/src/EntityFramework.Relational.Design/RelationalScaffoldingModelFactory.cs#L83) handles generating the names. This can be overridden. This does 2 things: (1) ensures unique names for each entity type and (2) ensures these identifiers are CSharp-safe.

**Future** we will move the csharpification of identifiers into the code-gen step (i.e. RelationalScaffoldingModelFactory will no longer need to return CSharp-safe identifiers.) We haven't designed code-gen API yet.

rowanmiller:
**Triage:** We should just handle snake case in our CSharpUtilities. @roji do you want to send a PR for this?

roji:
@rowanmiller sure I'll submit a PR soon.

roji:
Note this probably means that single uncapitalized words are uppercased. If so there's no more need to detect C# reserved words and prefix them with underscore, since `volatile` becomes Volatile anyway

natemcmaster:
Yup, this was one of the intentions of [my comments above](https://github.com/aspnet/EntityFramework/issues/3987#issuecomment-162962760). `IScaffoldingModelFactory.Create` shouldn't be concerned with the targeted code-gen language.

rowanmiller:
@lajones regardless of extension points, we should update EF to handle snake case

lajones:
Fix checked in with a065b93.

lajones:
See comments in #5360. Need to add <space> as a word separator.

lajones:
Blocked pending discussion between @divega, @rowanmiller and me to decide final design.

lajones:
[...]

PR #5360, opened 2016-05-13
RevEng: Split identifiers into words and Pascal-case them.

lajones:
Fix for #3987.

The code change is fairly small but this is a fairly big change to the names of identifiers for classes and properties generated by scaffolding (and so a large number of expected results files change).

(Also note: I found we had 2 copies of the `CSharpUtilities` class - so I removed one of them.)

**Negatives of this approach**: there will need to be more `[ColumnName]` and `[TableName]` attributes (or equivalent fluent API).
**Positives**: tends to make the resulting names look more like C#. We get better (more "C#-looking")identifiers for DB table/column names which contain underscores or hyphens and also better results for camel-cased or all-caps or (some) mixed case.

This will obviously be a matter of taste but we think the positives outweigh the negatives here.


lajones:
Adding `providers-beware` because if you have tests that look at the contents of generated files - this will probably change them (depending on what table/column names you have).

smitpatel:
:shipit:

lajones:
Checked in with commit a065b93.

test/Microsoft.EntityFrameworkCore.Sqlite.Design.FunctionalTests/SqliteScaffoldingModelFactoryTest.cs
```diff
@@ -157,13 +157,13 @@ FOREIGN KEY (parentid_A, parentid_B) REFERENCES parent (id_a, id_b)
 
             var propList = new List<Property>
             {
-                (Property)children.FindProperty("ParentId_A"),
-                (Property)children.FindProperty("ParentId_B")
+                (Property)children.FindProperty("ParentIdA"),
```
divega: Sorry I am only chipping in after this has been merged but this string reveals that we weren't all exactly on the same page with the rules :disappointed:. 

In my mind one of the rules was that if a string contained mixed case we would assume that the user already had it the way they wanted it to be and would avoid any further transformation, including removing "_"s.

@rowanmiller @roji thoughts on this?

rowanmiller: What we have is consistent with what I thought we'd come up with 😄 😕 . My understanding was that the pascal casing was only skipped on a per-segment basis after we had split on the underscore... so in this case we leave it as "ParentId" rather than making it "Parentid".

lajones: @divega  My understanding was the same as @rowanmiller's. But see also 2dc6007 where I checked in an update because I had missed a test that became broken because of this. For that test there is a table called "E F" which becomes an Entity called "E_f" because we split words only on '_' and '-' and so the algorithm regards all 3 chars as one "word" which it Pascalizes to "E f" which then goes through the part of algorithm which makes sure it's valid C# and we end up with "E_f".

I'm not sure how far we want to go down this route as it's a fairly unusual case but maybe we should add in more characters which act as "word separators"?

rowanmiller: I would treat " " the same as "_" and "-".

lajones: What about '+' or '&' or any kind of whitespace? Or punctuation?

rowanmiller: That's probably overkill I think, the common ones I have seen are "-", "_", and " " (though I think underscores are the most common).

lajones: OK. Have re-opened #3987 and will submit another PR for that.

divega: @lajones @rowanmiller I don't feel super strongly about this but I would like to make sure you can see what I mean: if I have a column in the database called `ParentId_A` the chances that I expect or want reverse engineering to fiddle with it to produce a property called "ParentIdA" are almost nil IMO. E.g. if I already knew to use mixed case in "ParentId" then the underscore is likely there on purpose. It is just heuristics but I don't think there is a way to turn it off, so trying to apply the principle of least surprise.

PR #5367, opened 2016-05-15
Optimistic concurrency tests

roji:
Some fixes to the optimistic concurrency test suite to make it more usable with non-SqlServer providers. Specifically, move the usage of byte[] Versions to the SqlServer-specific fixture (not supported by PostgreSQL).

The second commit turns Version into a shadow property, which again cleans up the model somewhat for non-SqlServer (where it wouldn't be used), and seems to make sense in general for concurrency tokens.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/F1FixtureBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/F1FixtureBase.cs
@@ -16,26 +16,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         {
             // TODO: Uncomment when complex types are supported
             //builder.ComplexType<Location>();
-            modelBuilder.Entity<Chassis>(b =>
-                {
-                    b.HasKey(c => c.TeamId);
-                    b.Property(e => e.Version)
-                        .ValueGeneratedOnAddOrUpdate()
-                        .IsConcurrencyToken();
-                });
-
-            modelBuilder.Entity<Driver>(b =>
-                {
-                    b.Property(e => e.Version)
-                        .ValueGeneratedOnAddOrUpdate()
-                        .IsConcurrencyToken();
-                });
-
-            modelBuilder.Entity<Engine>(b =>
-                {
-                    b.Property(e => e.EngineSupplierId).IsConcurrencyToken();
-                    b.Property(e => e.Name).IsConcurrencyToken();
-                });
+            modelBuilder.Entity<Chassis>().HasKey(c => c.TeamId);
 
             // TODO: Complex type
             // .Property(c => c.StorageLocation);
@@ -45,40 +26,10 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
             modelBuilder.Entity<Gearbox>();
 
-            // TODO: Complex type
-            //builder
-            //    .ComplexType<Location>()
-            //    .Properties(ps =>
-            //        {
-            //            // TODO: Use lambda expression
-            //            ps.Property<double>("Latitude", concurrencyToken: true);
-            //            // TODO: Use lambda expression
-            //            ps.Property<double>("Longitude", concurrencyToken: true);
-            //        });
-
-            modelBuilder.Entity<Sponsor>(b =>
-                {
-                    b.Property(e => e.Version)
-                        .ValueGeneratedOnAddOrUpdate()
-                        .IsConcurrencyToken();
-                });
-
-            // TODO: Complex type
-            //builder
-            //    .ComplexType<SponsorDetails>()
-            //    .Properties(ps =>
-            //        {
-            //            ps.Property(s => s.Days);
-            //            ps.Property(s => s.Space);
-            //        });
             modelBuilder.Ignore<SponsorDetails>();
 
             modelBuilder.Entity<Team>(b =>
                 {
-                    b.Property(t => t.Version)
-                        .ValueGeneratedOnAddOrUpdate()
-                        .IsConcurrencyToken();
-
                     b.HasOne(e => e.Gearbox).WithOne().HasForeignKey<Team>(e => e.GearboxId);
                     b.HasOne(e => e.Chassis).WithOne(e => e.Team).HasForeignKey<Chassis>(e => e.TeamId);
                 });
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/OptimisticConcurrencyTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/OptimisticConcurrencyTestBase.cs
@@ -105,8 +105,9 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             result.Add(entityType.FindProperty("Poles"), driver.Poles);
             result.Add(entityType.FindProperty("Races"), driver.Races);
             result.Add(entityType.FindProperty("TeamId"), driver.TeamId);
-            result.Add(entityType.FindProperty("Version"), driver.Version);
             result.Add(entityType.FindProperty("Wins"), driver.Wins);
+
+
             return result;
         }
 
@@ -131,9 +132,9 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             using (var context = CreateF1Context())
             {
                 var driver = context.Drivers.Single(d => d.CarNumber == 1);
-                Assert.NotEqual(1, driver.Version[0]);
+                Assert.NotEqual(1, context.Entry(driver).Property<byte[]>("Version").CurrentValue[0]);
                 driver.Podiums = StorePodiums;
-                firstVersion = driver.Version;
+                firstVersion = context.Entry(driver).Property<byte[]>("Version").CurrentValue;
                 await context.SaveChangesAsync();
             }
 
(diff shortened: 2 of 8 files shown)


src/Microsoft.EntityFrameworkCore.Specification.Tests/F1FixtureBase.cs
```diff
@@ -17,25 +17,9 @@ protected virtual void OnModelCreating(ModelBuilder modelBuilder)
...
-                });
-
-            modelBuilder.Entity<Engine>(b =>
-                {
-                    b.Property(e => e.EngineSupplierId).IsConcurrencyToken();
-                    b.Property(e => e.Name).IsConcurrencyToken();
-                });
+            {
+                b.HasKey(c => c.TeamId);
+            });
```
smitpatel: `modelBuilder.Entity<Chasis>().HasKey(c => c.TeamId);`

roji: @smitpatel have corrected and push-forced

AndriySvyryd:
@roji You should also update `OptimisticConcurrencyTestBase` to use the shadow property

roji:
@AndriySvyryd oops, sorry for missing that. Updated `OptimisticConcurrencyTestBase` and rebased on latest dev.

AndriySvyryd:
Travis is still failing. You also need to add the shadow Version to the SQLite tests.

roji:
Really sorry for not checking this more.

I've fixed and push-forced, but I'm getting 3 test failures which I don't understand, on Calling_Reload_on_a_Modified_entity_makes_the_entity_unchanged, Calling_Reload_on_a_Deleted_entity_makes_the_entity_unchanged, Calling_Reload_on_a_Unchanged_entity_makes_the_entity_unchanged.

Sample failure:

```
Microsoft.EntityFrameworkCore.Sqlite.FunctionalTests.OptimisticConcurrencySqliteTest.Calling_Reload_on_a_Deleted_entity_makes_the_entity_unchanged [FAIL]
  System.InvalidCastException : Unable to cast object of type 'Microsoft.EntityFrameworkCore.ChangeTracking.PropertyEntry' to type 'System.Byte[]'.
  Stack Trace:
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore\ChangeTracking\Internal\Snapshot.cs(4837,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.Snapshot`2.set_Item(Int32 index, Object value)
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore\ChangeTracking\Internal\InternalMixedEntityEntry.cs(64,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalMixedEntityEntry.WritePropertyValue(IPropertyBase propertyBase, Object value)
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore\ChangeTracking\Internal\InternalEntityEntry.cs(457,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetProperty(IPropertyBase propertyBase, Object value, Boolean setModified)
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore\ChangeTracking\Internal\InternalEntityEntry.cs(416,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.set_Item(IPropertyBase propertyBase, Object value)
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\OptimisticConcurrencyTestBase.cs(24,0): at Microsoft.EntityFrameworkCore.Specification.Tests.TestExtensions.SetValues(InternalEntityEntry internalEntry, Dictionary`2 values)
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\OptimisticConcurrencyTestBase.cs(61,0): at Microsoft.EntityFrameworkCore.Specification.Tests.TestExtensions.Reload(InternalEntityEntry internalEntry, DbContext context)
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\OptimisticConcurrencyTestBase.cs(68,0): at Microsoft.EntityFrameworkCore.Specification.Tests.TestExtensions.Reload(EntityEntry entityEntry, DbContext context)
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\OptimisticConcurrencyTestBase.cs(599,0): at Microsoft.EntityFrameworkCore.Specification.Tests.OptimisticConcurrencyTestBase`2.TestReloadPositive(EntityState state)
	C:\projects\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\OptimisticConcurrencyTestBase.cs(589,0): at Microsoft.EntityFrameworkCore.Specification.Tests.OptimisticConcurrencyTestBase`2.Calling_Reload_on_a_Deleted_entity_makes_the_entity_unchanged()
```

src/Microsoft.EntityFrameworkCore.Specification.Tests/OptimisticConcurrencyTestBase.cs
```diff
@@ -74,7 +74,7 @@ public static void Reload(this EntityEntry entityEntry, DbContext context)
                 var id = ((Driver)internalEntry.Entity).Id;
                 return context.Set<Driver>()
                     .Where(d => d.Id == id)
-                    .Select(d => d.GetValues(internalEntry.EntityType))
+                    .Select(d => d.GetValues(context, internalEntry.EntityType))
```
AndriySvyryd: To query the shadow property use `EF.Property` in the 'Select' body:
```C#
.Select(d => d.GetValues(internalEntry.EntityType, EF.Property<byte[]>(d, "Version")))
```

roji:
@AndriySvyryd, while looking at this I suddenly realized the whole point was to remove the byte[] Version from the base class as it's a provider-specific thing, so putting code in OptimisticConcurrencyTestBase.c which refers to it doesn't make much sense...

One option would be to refactor the whole Reload/GetDatabaseValues mechanism into something that the provider-specific test suites could extend, but on a hunch I simply removed the setting of Version in GetDatabaseValues() and all tests passed. I'm guessing the tests don't depend on Version actually being reloaded.

I've amended to make it so, but let me know if you think it's a problem. Another solution is to bring back Version as a real non-shadow property, but leave the configuration of the concurrency token up to the specific providers. A bit less clean but it'll work.

AndriySvyryd:
@roji The tests pass for SqlServer even without that code because many tests are disabled and `Reload` is not implemented fully. For now you can just execute different logic in `GetDatabaseValues` depending on whether `entityType.FindProperty("Version")` returns `null`
Also move `Modifying_concurrency_token_only_is_noop()` to SqlServer

src/Microsoft.EntityFrameworkCore.Specification.Tests/OptimisticConcurrencyTestBase.cs
```diff
@@ -105,8 +105,9 @@ public static void Reload(this EntityEntry entityEntry, DbContext context)
             result.Add(entityType.FindProperty("Poles"), driver.Poles);
             result.Add(entityType.FindProperty("Races"), driver.Races);
             result.Add(entityType.FindProperty("TeamId"), driver.TeamId);
-            result.Add(entityType.FindProperty("Version"), driver.Version);
             result.Add(entityType.FindProperty("Wins"), driver.Wins);
+
+
             return result;
```
AndriySvyryd: extra line

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/F1SqlServerFixture.cs
```diff
@@ -59,5 +59,64 @@ public override F1Context CreateContext(SqlServerTestStore testStore)
...
+            });
+
+            // TODO: Complex type
+            //builder
+            //    .ComplexType<SponsorDetails>()
+            //    .Properties(ps =>
+            //        {
+            //            ps.Property(s => s.Days);
+            //            ps.Property(s => s.Space);
+            //        });
```
AndriySvyryd: Move these comments to the base

roji:
@AndriySvyryd sorry for being away for so long.

I don't have enough time to dive into this and fully understand why, but any attempt to turn the concurrency token (Version) into a shadow property in the tests fails for me - please see [this commit](https://github.com/roji/EntityFramework/commit/5e19c5210140b19ee8cd7aef3f792207dde17e7f) for a very minimal attempt that already fails the tests. I suspect it may have something to do with that fact that GetValues now has to pass through the context to get the current value of Version, rather than access it directly on the Clr instance as before. Not sure at all though.

Executing different logics in `GetDatabaseValues` as you proposed would require me to embed PostgreSQL-specific stuff in your test base class: in PostgreSQL I'm going to try to make use of an autoincrementing int property as a concurrency token. I can definitely hack things that way but - add an IntVersion alongside the current byte[] Version - but that's pretty dirty. This would obviate shadow properties though and would definitely work - let me know if you want me to do that.

AndriySvyryd:
@roji The reason the tests are failing is that in your change you are getting the value for `Version` from the `EntityEntry` tracked in the context, but the purpose of the method is to get the database values. You could do this using the `EF.Property` method as I mentioned previously, but then this code will need to be different for providers that have a different model. So I reimplemented `GetDatabaseValues` to dynamically gets all properties, so it's provider agnostic, see https://github.com/aspnet/EntityFramework/commit/b9ece5b791a4e87c56d5a3d5268a24a9b611f897

roji:
Great, thanks for your time @AndriySvyryd!

Will you be committing this?

AndriySvyryd:
I will if it's enough for your tests to work

roji:
Thanks again for this @AndriySvyryd, the effort you guys are spending on other platforms/database is really impressive. I'll work on optimistic concurrency in the next few days and will post back if there are any issues.

PR #5372, opened 2016-05-16
Make RelationalModelValidator's fields protected

roji:
To allow provider-specific model validators to access them.


AndriySvyryd:
@roji Also make them virtual

roji:
@AndriySvyryd done

Issue #5289, opened 2016-05-07
RevEng: DataAnnotations are not used for column types


smitpatel:
Using northwind.db
command:
Scaffold-DbContext "Data Source=C:\Users\smpatel\Documents\Visual Studio 2015\Projects\ClassLibrary2\northwind.db" Microsoft.EntityFrameworkCore.Sqlite -OutputDir .\Models -Context MyContext -DataAnnotations -Tables Employees,Orders -Project ClassLibrary1 -Environment Production -Force -Verbose
Generates following in modelbuilder code
```
modelBuilder.Entity<Orders>(entity =>
{
    entity.HasKey(e => e.OrderID)
        .HasName("sqlite_autoindex_Orders_1");

    entity.Property(e => e.OrderID).HasColumnType("int");

    entity.Property(e => e.CustomerID).HasColumnType("char");

    entity.Property(e => e.EmployeeID).HasColumnType("int");

    entity.Property(e => e.Freight)
        .HasColumnType("decimal")
        .HasDefaultValueSql("0");

    entity.Property(e => e.OrderDate).HasColumnType("datetime");

    entity.Property(e => e.RequiredDate).HasColumnType("datetime");

    entity.Property(e => e.ShipAddress).HasColumnType("nvarchar");

    entity.Property(e => e.ShipCity).HasColumnType("nvarchar");

    entity.Property(e => e.ShipCountry).HasColumnType("nvarchar");

    entity.Property(e => e.ShipName).HasColumnType("nvarchar");

    entity.Property(e => e.ShipPostalCode).HasColumnType("nvarchar");

    entity.Property(e => e.ShipRegion).HasColumnType("nvarchar");

    entity.Property(e => e.ShipVia).HasColumnType("int");

    entity.Property(e => e.ShippedDate).HasColumnType("datetime");
});
```
Some of the types here like `int`, `decimal` can be specified as DataAnnotations. But the whole model has no data-annotations for any kind of column types.

Same happens with SqlServer provider:
```
modelBuilder.Entity<Orders>(entity =>
{
    entity.HasIndex(e => e.CustomerID)
        .HasName("CustomersOrders");

    entity.HasIndex(e => e.EmployeeID)
        .HasName("EmployeesOrders");

    entity.HasIndex(e => e.OrderDate)
        .HasName("OrderDate");

    entity.HasIndex(e => e.ShipPostalCode)
        .HasName("ShipPostalCode");

    entity.HasIndex(e => e.ShipVia)
[...]

PR #5380, opened 2016-05-16
[SMALL] Using Column(TypeName=xyz) annotation instead of fluent API where appropriate

lajones:
Fix for scaffolding issue #5289. Use `[Column(TypeName = "xyz")]` instead of `HasColumnType("xyz")` when the user asks for DataAnnotations.


lajones:
Adding `providers-beware` because if you have tests which look at the contents of the generated files then, depending on the data types you test, they could change.

smitpatel:
:shipit:

lajones:
Checked in with commit 2d30f14.

PR #5381, opened 2016-05-17
Ensure discriminator values are removed when the discriminator type is changed

### LINKED ISSUE #5258: Migrations: Incorrect discriminator value generated if a derived class is referenced from an entity
Consider the following example:
```
    public abstract class Vehicle
    {
        public int Id { get; set; }
    }

    public class Car : Vehicle
    {
        public int CarparkId { get; set; }
    }

    public class Carpark
    {
        public int Id { get; set; }
        public ICollection<Car> Cars { get; set; }
    }

    {
        public DbSet<Vehicle> Vehicles { get; set; }
        public DbSet<Carpark> Carparks { get; set; }

        {
            optionsBuilder.UseSqlServer("Server=localhost;Database=MyDatabase;Trusted_Connection=True;MultipleActiveResultSets=true");
        }

        {
            modelBuilder.Entity<Vehicle>().HasDiscriminator<int>("Discriminator")
                .HasValue<Car>(1);
        }
    }
```

The `Carpark` references the derived type `Car`. This causes the following model snapshot to be generated:

```
modelBuilder.Entity("CustomAnnotationsRepro.Vehicle", b =>
                {
                    b.Property<int>("Id")
                        .ValueGeneratedOnAdd();

                    b.Property<int>("Discriminator");

                    b.HasKey("Id");

                    b.ToTable("Vehicles");

                    b.HasDiscriminator<int>("Discriminator").HasValue("Vehicle");
                });
```
The `.HasValue("Vehicle")` should not be included and the code cannot compilse as `"Vehicle"` cannot be converted to int.

Latest rc2 bits

AndriySvyryd:
Fixes #5258

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Metadata/Internal/RelationalEntityTypeBuilderAnnotations.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Metadata/Internal/RelationalEntityTypeBuilderAnnotations.cs
@@ -131,15 +131,13 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
             [CanBeNull] Func<InternalEntityTypeBuilder, InternalPropertyBuilder> createProperty,
             [CanBeNull] Type propertyType)
         {
+            var configurationSource = Annotations.ConfigurationSource;
             var discriminatorProperty = DiscriminatorProperty;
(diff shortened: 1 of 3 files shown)


smitpatel:
:shipit:

PR #5388, opened 2016-05-17
Add more logging during database drop command

### LINKED ISSUE #5316: dotnet ef database drop does not provide any feed back when y/Y is keyed
It seems like the command is hung when the key is pressed. It would be nice to get some form of response that says it's busy deleting the db.

natemcmaster:
Fix #5316.

cc @bricelam @pranavkm @rowanmiller


src/Microsoft.EntityFrameworkCore.Tools.Cli/project.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-preview1-*",
+  "version": "1.0.0-preview2-*",
```
pranavkm: Did I miss the other packages here?

pranavkm:
:shipit:

src/Microsoft.EntityFrameworkCore.Tools.Cli/project.json
natemcmaster: This was the only missed package I saw. Everything else is rc3/preview2

PR #5396, opened 2016-05-17
Core: Fix #5333 - Allows warnings to be treated as runtime errors.

### LINKED ISSUE #5333: More generic solution for severity level of warnings (throw/warn/ignore)
Currently we have a special API for what happens with client eval during queries. We want to make this a more generic concept so that we can handle other warning types (such as ignored includes https://github.com/aspnet/EntityFramework/issues/5332). 

The min-bar for RTM is to have a context wide API. It would be nice to have the ability to override these on a per-query basis, so that you can say "I'm ok with client eval for this query" etc. But that can be split off to a separate issue if we don't get to it for RTM.

anpete:
- Adds DbContextOptionsBuilder.SetWarningsAsErrors(params CoreLoggingEventId[]) and RelationalDbContextOptionsBuilder.SetWarningsAsErrors(params RelationalLoggingEventId[]) top-level configuration APIs.
- Re-uses our logging event ids as the way to identify specific warning kinds.
- Calling the API with no-args turns on errors for all event ids defined in that assembly (Core or Relational).
- Warning exception generation logic centralized in InterceptingLogger via LogWarning EF extensions.


anpete:
cc @rowanmiller @divega @ajcvickers @AndriySvyryd for API feedback

AndriySvyryd:
Wouldn't a blacklist approach be more useful, as one usually disables the warnings that are too noisy? Perhaps an overload: `SetWarningsAsErrorsExcept`
On a related note, we probably need a more granular Id system to disable specific warnings, but this can be done post-rtm.

anpete:
@AndriySvyryd Possibly. There are so few right now that I'm not sure it matters too much.

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
```diff
@@ -57,6 +57,19 @@ public virtual TBuilder QueryClientEvaluationBehavior(QueryClientEvaluationBehav
...
         }
 
+        public virtual TBuilder SetWarningsAsErrors(
+            [NotNull] params RelationalLoggingEventId[] relationalLoggingEventIds)
+        {
+            Check.NotNull(relationalLoggingEventIds, nameof(relationalLoggingEventIds));
+
+            if (relationalLoggingEventIds.Length == 0)
+            {
+                relationalLoggingEventIds = (RelationalLoggingEventId[])Enum.GetValues(typeof(RelationalLoggingEventId));
```
divega: I am not sure how often this code would be called (maybe once per request?) but we should probably cache the outcome of the `Enum.GetValues()` or maybe we should use a separate value in the enum representing "all warnings".

divega: Anyway, @anpete and I talked and he mention there are a few other options we should discuss:
- Use the new event id struct that Logging defines
- Have a bespoke type (enum perhaps) for use in this API instead of trying to reuse event ids

cc @ajcvickers @rowanmiller @bricelam

src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerDbContextOptionsBuilder.cs
```diff
@@ -20,6 +20,7 @@ protected override SqlServerOptionsExtension CloneExtension()
         /// <summary>
         ///     Use a ROW_NUMBER() in queries instead of OFFSET/FETCH. This method is backwards-compatible to SQL Server 2005.
         /// </summary>
-        public virtual void UseRowNumberForPaging() => SetOption(e => e.RowNumberPaging = true);
```
divega: Yikes, we were returning void?

src/Microsoft.EntityFrameworkCore/DbContextOptionsBuilder.cs
```diff
@@ -128,6 +128,19 @@ public virtual DbContextOptionsBuilder UseInternalServiceProvider([CanBeNull] IS
         public virtual DbContextOptionsBuilder EnableSensitiveDataLogging()
             => SetOption(e => e.IsSensitiveDataLoggingEnabled = true);
 
+        public virtual DbContextOptionsBuilder SetWarningsAsErrors(
```
divega: I believe I understand why there are core and relational versions of this API but it seems unfortunate you cannot do it only once. Maybe another thing we can chat about.

divega:
I believe the approach is good. We need to discuss a few more details.

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
rowanmiller: I think I was expecting an API more like this:

```
options.SetWarningLevel(WarningLevels.Ignore) // <- Default level for all warnings
       .SetWarningLevel(WarningLevels.Throw, Warnings.ClientEvaluation)
       .SetWarningLevel(WarningLevels.Warn, Warnings.IgnoredIncludes, Warnings.SomeOtherWarning)
```

divega: I like what @rowanmiller  suggests better, e.g. regardless of the naming, a method to which you can pass values for ignore, throw or warn. From what I remember when we discussed this in triage we said (or at least I thought to myself) this would become a full superset (and replacement) for the current `QueryClientEvaluationBehavior()` method.

I am still not sure about the granularity...

@rowanmiller FYI, when we discussed yesterday we looked at several alternatives for how to represent and configure individual warnings. The main challenge is that they are defined by several different components, e.g. we have some defined by core, some by relational, and potentially we may have provider specific warnings. It didn't look like a single enum would work, and the other alternatives didn't look very straightforward, so we left the meeting with the idea of scoping granularity out for RTM (e.g. the setting would apply to all warnings), so we can have more time to bake an API for it. What do you think?

anpete: @rowanmiller Another thing to consider is that we have enums for warnings already, in the form of our logging event ids.

rowanmiller: > so we left the meeting with the idea of scoping granularity out for RTM

Yep that's fine. I think we should have the ability to set any of the three levels though, and not just the ability to make warnings throw.

> Another thing to consider is that we have enums for warnings already, in the form of our logging event ids. 

For sure, I should have put a note in that the API names etc. were all just a sketch of a general pattern, not actually what I thought we'd have. EventId's do seem like the logical thing to control this.

src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerDbContextOptionsBuilder.cs
anpete: Yep :smile:

anpete:
@divega Updated to remove event id arguments.

divega:
@anpete thanks. As @rowanmiller said at https://github.com/aspnet/EntityFramework/pull/5396#discussion_r64087716:

> I think we should have the ability to set any of the three levels though, and not just the ability to make warnings throw.

And what about removing `QueryClientEvaluationBehavior()` for now? It seems that will be completely superseded by this API once we have the granularity, but it doesn't seem to bad to ship RTM without the granularity for that specific case.

anpete:
Sending a new PR

PR #5398, opened 2016-05-17
Remove cross compiling for netcore50 

natemcmaster:
Add custom packing build step so tools can still be installed to UWP apps.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -16,13 +16,6 @@ var AUTHORS='Microsoft Open Technologies, Inc.'
 use-standard-lifecycle
 k-standard-goals
 
-#initialize-tools target='initialize'
-  @{
-    // workaround bug in dotnet-build for netcore50 on .Tools and .Tools.Core
-    string restoreOptions = " --infer-runtimes " + E("KOREBUILD_DOTNET_RESTORE_OPTIONS");
-  }
-  dotnet-restore restore_options='${ restoreOptions } src/Microsoft.EntityFrameworkCore.Tools/'
-
 #csproj-initialize target='initialize' if='(!IsMono && !IsTeamCity && E("APPVEYOR") == null) || E("IsEFPerfBuild") != null'
   var programFilesX86='${Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86)}'
   var buildProgram='${Path.Combine(programFilesX86, "MSBuild", "14.0", "Bin", "MSBuild.exe")}'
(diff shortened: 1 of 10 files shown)


natemcmaster:
cc @pranavkm for the makefile stuff
cc @divega @Eilon once this is in, we can remove netcore50 from EF's dependencies

https://github.com/aspnet/Common/compare/namc/no-netcore50
https://github.com/aspnet/DependencyInjection/compare/namc/no-netcore50
https://github.com/aspnet/Caching/compare/namc/no-netcore50
https://github.com/aspnet/PlatformAbstractions/compare/namc/no-netcore50
https://github.com/aspnet/Microsoft.Data.Sqlite/compare/namc/no-netcore50
https://github.com/aspnet/Logging/compare/namc/no-netcore50
https://github.com/aspnet/Options/compare/namc/no-netcore50

makefile.shade
```diff
@@ -31,4 +34,50 @@ k-standard-goals
...
+      Log.Info("Repacking Nupkg: " + projectNupkg);
+
+       var extractToDirectory = projectNupkg + "-temp";
+      ZipFile.ExtractToDirectory(projectNupkg, extractToDirectory);
+
+      var nuspecFile = Files.Include(Path.Combine(extractToDirectory, "*.nuspec")).First();
+      var nuspec = new List<string>(File.ReadAllLines(nuspecFile));
+      var dependencies = Array.FindIndex<string>(nuspec.ToArray(), l => l.Contains("<dependencies>")) + 1;
+
+      var uap = nuspec.SkipWhile(l => !l.Contains("<group targetFramework=\".NETFramework4.5.1"))
```
pranavkm: Do you want to make this a tool instead? Might be more reliable to use nuget APIs to do this. Here's what we do in Coherence-Signed: https://github.com/aspnet/Coherence-Signed/blob/dev/src/PushCoherence/Program.cs#L90-L122

divega:
@natemcmaster can we chat a bit about the implications of this tomorrow?

makefile.shade
natemcmaster: Yes, this plus what we add in https://github.com/aspnet/EntityFramework/pull/5399 make me think we'd be better of making this a tool. Too bad we can't use nuspec directly or avoid compilation and yes still produce a nupkg that targets netcore50 (https://github.com/dotnet/cli/issues/897).

natemcmaster:
@divega of course, i'll be here all day :)

natemcmaster:
:watch: waiting to merge...found a potential show-stopper bug in .NET Native compiler + CoreFX. If this bug won't be fixed for RTM, then we'll need to continue cross compiling to old versions of CoreFX for UWP.

natemcmaster:
cc @bricelam

natemcmaster:
The corefx bug will be fixed for RTM. We can move ahead with this PR now.

bricelam:
:shipit:

PR #5399, opened 2016-05-17
dotnet-ef for net451-x86

### LINKED ISSUE #5285: Tools.CLI is 64 bit architecture
Not an error just a warning but this came up while using rev eng on win7 32 bit machine

C:\Program Files\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1820,5): warning MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "Microsoft.EntityFrameworkCore.Tools.Cli, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, processorArchitecture=AMD64", "AMD64". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.

natemcmaster:
Add custom build step for x86 net451 dotnet-ef command support

Fix #5285 

Depends on some variables added in #5398 

cc @NTaylorMullen @prafullbhosale

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -81,3 +81,68 @@ k-standard-goals
           // Don't care if we couldn't delete the temp directory.
       }
   }
+
+#repack-cli-tools target='compile'
+  @{    
+      if (string.IsNullOrEmpty(CONFIGURATION_LOCAL))
+      {
+          CONFIGURATION_LOCAL = "Debug";
+      }
+      
+      // Forcing to lower to match the entry in the sources project.json.
+      var configurationX86 = CONFIGURATION_LOCAL.ToLower() + "_x86";
+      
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
:shipit:

natemcmaster:
Note: I intentionally don't want to include the exe under lib/net451/ so what we don't get MSBuild errors about CPU architecture

natemcmaster:
Merged with https://github.com/aspnet/EntityFramework/commit/3c575741a9bffe57c08a70e7fa2fc1655be51833

PR #5400, opened 2016-05-17
Fix #5266 - Remove AsAsyncEnumerable from public surface.

### LINKED ISSUE #5266: Consider removing IX-async dependency
We currently are getting value from IX-async in two ways:

- Shared `IAsyncEnumerable<T>` interfaces
- In-memory implementation of async LINQ operators

A few new data points may affect the initial decision on taking this dependency:

- The .NET languages team is working on standard version the interfaces and language support for them (see https://github.com/ljw1004/roslyn/blob/features/async-return/docs/specs/feature%20-%20async%20iterators.md). We definitely will want to swap to that when it is ready 
- We have found performance issues in the LINQ operator implementation of IX-async which we solved by implementing our own version of some of the operators

The proposal is to define our own interface in an internal namespace, so that we can swap to the standard interfaces and implementations when they are available but in the meanwhile enable developers that are ok with opting in to possible breaking changes to code against our interfaces.

If there is enough demand for it, we can always create a bridge to IX-async.

The cost of removing the dependency would be mostly the cost of finishing the implementation of all LINQ operators.

anpete:


lajones:
:shipit:

Issue #5353, opened 2016-05-12
Change namespace M.EFCore.Tools.Core.Scaffolding.Internal.Configuration


lajones:
Our convention is to _end_ the namespace with `.Internal`. So we should move the files that are currently in `Microsoft.EntityFrameworkCore.Tools.Core.Scaffolding.Internal.Configuration` into `Microsoft.EntityFrameworkCore.Tools.Core.Scaffolding.Configuration.Internal`.

lajones:
Fix checked in with c6e0f04.

PR #5402, opened 2016-05-18
Move namespace to ...Configuration.Internal

lajones:
Fix for #5353. Move namespace from `...Internal.Configuration` to `...Configuration.Internal`.


smitpatel:
:shipit:

lajones:
Checked in commit c6e0f04.

PR #5403, opened 2016-05-18
Fix #5365 - Use type mapper to determine SQL translate-ability.

### LINKED ISSUE #5365: PostgreSQL-specific types ignored in query's where clause
I'm trying to run some BuiltInDataTypes functional tests on PostgreSQL-specific types. These types have been properly mapped in my NpgsqlTypeMapper, and indeed most of the tests pass quite well (it's a real pleasure to see this work).

There's one important issue though, with Can_query_using_any_mapped_data_type which places the types in a where clause. Looking at the generated SQL, I can see that with regular types the where clause is good (`WHERE ("e"."Int" = 999) AND ("e"."Bool" = $1)`). However, PostgreSQL types get dropped completely: ` WHERE "e"."Int" = 999`. Examples of CLR types which cause this are PhysicalAddress, TimeSpan and some others.

Have I messed things up in NpgsqlTypeMapper, or does the EFCore query generator somehow not accept non-standard types yet?

(~~this may be somehow related to #3617~~ probably not actually)

anpete:


smitpatel:
:shipit:

PR #5409, opened 2016-05-18
Look at all derived types when creating a column during migration.

### LINKED ISSUE #5379: Migrations: MigrationsSqlGenerator fails to find correct property for column
`FindEntityType` function
https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs#L786
`FindProperty` function
https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs#L793

`FindEntityType` can match with derived entity type also due to same mapped table in database.
Result of `FindProperty` depends on result of above. If base type is returned then properties in derived type are not found.
This causes Migrations to have no info about `IProperty` therefore in the absence of `ColumnType` clr type is used to find mapping which causes loss of information about max length.

AndriySvyryd:
Fixes #5379

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
@@ -9,6 +9,7 @@ using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Migrations.Operations;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Utilities;
@@ -783,11 +784,11 @@ namespace Microsoft.EntityFrameworkCore.Migrations
             }
         }
 
-        protected virtual IEntityType FindEntityType(
+        protected virtual IEnumerable<IEntityType> FindEntityTypes(
             [CanBeNull] IModel model,
             [CanBeNull] string schema,
             [NotNull] string tableName)
-            => model?.GetEntityTypes().FirstOrDefault(
+            => model?.GetEntityTypes().Where(
                 t => (_annotations.For(t).TableName == tableName) && (_annotations.For(t).Schema == schema));
 
         protected virtual IProperty FindProperty(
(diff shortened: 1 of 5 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -783,20 +783,20 @@ public class MigrationsSqlGenerator : IMigrationsSqlGenerator
...
                 t => (_annotations.For(t).TableName == tableName) && (_annotations.For(t).Schema == schema));
 
         protected virtual IProperty FindProperty(
             [CanBeNull] IModel model,
             [CanBeNull] string schema,
             [NotNull] string tableName,
             [NotNull] string columnName)
-            => FindEntityType(model, schema, tableName)
-                ?.GetProperties().FirstOrDefault(p => _annotations.For(p).ColumnName == columnName);
+            => FindEntityTypes(model, schema, tableName).SelectMany(e => e.GetProperties())
```
smitpatel: Shouldn't this be `GetDeclaredProperties()`? Otherwise if the entity type has derived type then `SingleOrDefault` will fail due to multiple properties matching the condition.

AndriySvyryd: Yes

test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTestBase.cs
```diff
@@ -81,6 +81,26 @@ public virtual void AddColumnOperation_with_maxLength()
...
+            => Generate(
+                modelBuilder =>
+                    {
+                        modelBuilder.Entity("Person");
+                        modelBuilder.Entity("SpecialPerson", b =>
+                            {
+                                b.HasBaseType("Person");
+                                b.Property<string>("Name").HasMaxLength(30);
+                            });
+                    },
```
smitpatel: Can you add 1 more derived type deriving from `SpecialPerson` for better coverage. (i.e. entity type with max length property has both base type and derived type).

smitpatel:
:shipit:

test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTestBase.cs
AndriySvyryd: Sure

Issue #5303, opened 2016-05-09
How to generate custom IDs?


shaulbehr:
EF7 does all the usual stuff about auto-generating auto-incrementing IDs. Now, I have some tables that require custom key generation on insert.
1. COMB Guids. i.e. Guids that get inserted (close to) sequentially so they don't wreak havoc with the clustered index.
2. One table has a custom GUID generated from the hash of some other values on the table, to ensure that the ID is consistent over all our environments.

I saw your video [here](https://channel9.msdn.com/Blogs/Seth-Juarez/Key-Generation-Strategies-in-Entity-Framework-7), but it doesn't deal with the cases I mentioned.

How do you generate custom IDs?

rowanmiller:
Are values generated in the database, or in memory prior to insert? Also, it this an existing database, or are you using migrations to create the database from the model?

shaulbehr:
We *could* generate the COMB GUIDs in the database, but a good part of the reason why we chose to have GUID keys is so that we can generate them in memory and have the values available before we insert.

This is an existing database; we're migrating to EF7 from NHibernate.

rowanmiller:
@ajcvickers can you point in the right direction for using a value generator to do this.

shaulbehr:
@ajcvickers I will be most grateful for any assistance you can offer!

shaulbehr:
Bump.  @rowanmiller , @ajcvickers could you please point me in the right direction for this?

ajcvickers:
This is not very easy to do with the current bits. We need to add some sugar that will do most of the heavy lifting for you, but until we do that here are the steps to get custom value generators to work.

1) Implement a custom value generator. Something like:
```C#
public class CustomGuidValueGenerator : ValueGenerator<Guid>
{
    public override Guid Next() => // Logic for generating values

    public override bool GeneratesTemporaryValues => false;
}
```
Note that there is already an implementation to generate sequential GUID values on SQL Server (SequentialGuidValueGenerator) and this is actually the default value generator we use for GUIDs when using SQL Server.)

2) Extend the ValueGeneratorSelector for your provider to let EF know which properties need to use the new value generator. Using InMemory as an example:
```C#
public class CustomInMemoryValueGeneratorSelector : InMemoryValueGeneratorSelector
{
    public CustomInMemoryValueGeneratorSelector(IValueGeneratorCache cache)
        : base(cache)
    {
    }

    public override ValueGenerator Create(IProperty property, IEntityType entityType)
    {
[...]

PR #5423, opened 2016-05-19
Updating ValueGenerator to pass in entry

ajcvickers:
See #5303. This allows the value generator to use state from the current entity as part of the value generation process. Doing this before RTM to avoid a breaking change doing it later.


divega:
:shipit:

Issue #5154, opened 2016-04-22
RevEng : Use unbounded navigations in scaffolded model


smitpatel:
Since updated navigation are supported in design time model now, RevEng should use them to have navigations in scaffolded model.

divega:
A bit off-topic, but any thoughts on whether we should actually commit to the term "shadow navigation" in our ubiquitous language? 

In my mind things are shadow only if at runtime their original and current values are tracked in shadow state, e.g. shadow entities don't currently exist but they may exist in the future. However these navigation properties don't have their own state so to me they are just metadata-only navigation properties or unbound navigation properties.

I am ok with embracing the term even if it doesn't feel completely accurate :smile:

smitpatel:
Updated title & issue. I did not think in that detail about the term "shadow" but not using it for navigations seems more reasonable.

divega:
@smitpatel I wasn't really pushing back. I would just like to hear what people in the team think we should call them.

ajcvickers:
@divega No reason why they couldn't have state in the future. Would be relatively easy to implement, although I'm not sure that the value would be high.

I haven't looked at the code yet, but presumably they can already or will at some point be useable in query to do things like Include.

lajones:
Note: see #5146 for the fix that allowed navigation properties to point to entities which do not yet have a backing CLR type (which is what we have when constructing the `IModel` during RevEng).

lajones:
Fix checked in with 767b3dd.

PR #5432, opened 2016-05-19
RevEng: Use unbound NavProps on the IModel instead of annotations

lajones:
Fix for #5154. This is clean up - no change to output from Scaffolding. Now that you can add "shadow" NavProps to an `IModel` we make use of them so we can stop using the artificial `ScaffoldingForeignKeyAnnotations`.

Change to `ForeignKey` is just cleaning up a whitespace problem I saw when I was doing this.


lajones:
Note for provider-writers: if you previously were using `ScaffoldingForeignKeyAnnotations` to store navigation property names, you'll have to update to use `foreignKey.HasDependentToPrincipal()` and `foreignKey.HasPrincipalToDependent()` instead.

src/Microsoft.EntityFrameworkCore.Tools.Core/Scaffolding/Configuration/Internal/ModelConfiguration.cs
```diff
@@ -674,10 +673,8 @@ public virtual void AddRelationshipConfiguration([NotNull] EntityConfiguration e
 
             foreach (var foreignKey in entityConfiguration.EntityType.GetForeignKeys())
             {
-                var dependentEndNavigationPropertyName =
-                    foreignKey.Scaffolding().DependentEndNavigation;
-                var principalEndNavigationPropertyName =
-                    foreignKey.Scaffolding().PrincipalEndNavigation;
+                var dependentEndNavigationPropertyName = foreignKey.DependentToPrincipal.Name;
+                var principalEndNavigationPropertyName = foreignKey.PrincipalToDependent.Name;
```
smitpatel: null check for navigations needed?

smitpatel:
LGTM.

src/Microsoft.EntityFrameworkCore.Tools.Core/Scaffolding/Configuration/Internal/ModelConfiguration.cs
lajones: Will do. After that can I convert the LGTM to a ship-it?

smitpatel:
If someone else wants to review otherwise :shipit:

lajones:
Checked in with commit 767b3dd.

PR #5433, opened 2016-05-19
Use WHERE EXISTS instead of INNER JOIN for collection navigation Include

### LINKED ISSUE #4903: Include() generates unnecessary DISTINCT keywords for collection queries
A query of the form `context.Posts.Include(p => p.Comments)` generates SQL like this:

``` SQL
SELECT [p].[PostId], [p].[Body]
FROM [Posts] AS [p]
ORDER BY [p].[PostId]

SELECT [c].[CommentId], [c].[Body], [c].[PostPostId]
FROM [Comment] AS [c]
INNER JOIN (
    SELECT DISTINCT [p].[PostId]
    FROM [Posts] AS [p]
) AS [p1] ON [c].[PostPostId] = [p1].[PostId]
ORDER BY [p1].[PostId]
```

This works correctly but there doesn't seem to be any need for the `DISTINCT` keyword in the subquery in any scenario as the principal key seems to always be included in the projection.

smitpatel:
when there is no ORDER BY Clause or ORDER BY is on principal key properties
resolves #4903

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
@@ -274,6 +274,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                             ? selectExpression.Tables[0]
                             : selectExpression.Tables.Last(t => t.QuerySource == querySource);
 
+                    var canGenerateExists = !IsOrderingOnNonPrincipalKeyProperties(selectExpression.OrderBy, navigation.ForeignKey.PrincipalKey.Properties);
+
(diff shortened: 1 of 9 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -306,27 +308,73 @@ var materializer
...
-                        = BuildJoinEqualityExpression(
-                            navigation,
-                            targetTableExpression,
-                            innerJoinExpression,
-                            querySource);
+                        subqueryExpression.Predicate
+                            = compositePredicateExpressionVisitor
+                                .Visit(subqueryExpression.Predicate);
+
+                        var map = navigation.ForeignKey.PrincipalKey.Properties
```
anpete: Better variable name?

anpete:
:shipit:

Issue #5191, opened 2016-04-27
Query :: invalid results returned for queries with complex projection involving FirstOrDefault and navigation property


maumar:
```cs
        [ConditionalFact]
        public virtual void Collection_select_nav_prop_first_or_default_then_nav_prop()
        {
            using (var context = CreateContext())
            {
                var customers
                    = (from c in context.Set<Customer>()
                       select new { c.Orders.FirstOrDefault().Customer }).ToList();

                Assert.Equal(91, customers.Count);
            }
        }
```

The test above returns nulls instead of actual customer values.

Query plan is as follows:

```
TRACKED: False
(QueryContext queryContext) => IEnumerable<<>f__AnonymousType122<Customer>> _Select(
    source: IEnumerable<ValueBuffer> _ShapedQuery(
        queryContext: queryContext, 
        shaperCommandContext: SelectExpression: 
            SELECT [c].[CustomerID]
            FROM [Customers] AS [c]
        , 
        shaper: ValueBufferShaper
    )
    , 
    selector: (ValueBuffer c) => new <>f__AnonymousType122<Customer>(Customer FirstOrDefault(
            source: IEnumerable<Customer> _Select(
                source: IEnumerable<Order> _Where(
                    source: IEnumerable<Order> _ShapedQuery(
                        queryContext: queryContext, 
                        shaperCommandContext: SelectExpression: 
                            SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
                            FROM [Orders] AS [o]
                        , 
                        shaper: BufferedEntityShaper<Order>
                    )
                    , 
                    predicate: (Order <generated>_1) => (string) object c.get_Item(0) == string GetValue(
                        queryContext: queryContext, 
                        entity: <generated>_1, 
                        property: CustomerID
                    )
                )
                , 
                selector: (Order <generated>_1) => <generated>_1.Customer
            )
        )
        )
)
```

maumar:
fixed in 0da202b8e7d23d94f071bd1d161716e23f6e8d8d

PR #5434, opened 2016-05-19
Fix to #5191 - Query :: invalid results returned for queries with complex projection involving FirstOrDefault and navigation property

maumar:
Problem was that we were not applying navigation expansion to some queries (when the navigation was based of a subquery, rather than query source reference), e.g.

ctx.Customers.Select(c => ctx.Orders.FirstOrDefault().Customer)

This would cause incorrect results - Customer navigation would always be null.

Fix is to push those navigations before the FirstOrDefault call, like so:

ctx.Customers.Select(c => ctx.Orders.Select(o => o.Customer).FirstOrDefault())


maumar:
@anpete

anpete:
@maumar We have a visitor for this already: SubQueryMemberPushDownExpressionVisitor - I think the problem is that it is running after nav rewrite.

maumar:
@anpete  you are right, moving SubQueryMemberPushDownExpressionVisitor  before navigation expansion solves the problem, new version up

anpete:
:shipit:

Issue #5108, opened 2016-04-19
Collection navigation include with Skip/Take/FirstOrDefault


Trungkho:
In one of my test projects, I tried this query:

```
kpiReport = context.KpiReports.OrderByDescending(r => r.RunDate)
    .Include(r => r.DepartmentResults).ThenInclude(r => r.Department)
    .Include(r => r.KpiResults).ThenInclude(r => r.Employee)
    .FirstOrDefault();
```

And the following queries are generated:
```
SELECT TOP(1) [r].[Id], [r].[RunDate]
FROM [KpiReport] AS [r]
ORDER BY [r].[RunDate] DESC, [r].[Id]
```

```
SELECT [k].[Id], [k].[Discriminator], [k].[EmployeeCode], [k].[FilePath], [k].[ParentReportId], [k].[Type], [e].[Code], [e].[Active], [e].[DepartmentCode], [e].[Level], [e].[Name]
FROM [KpiResult] AS [k]
INNER JOIN (
    SELECT DISTINCT TOP(1) [r].[RunDate], [r].[Id]
    FROM [KpiReport] AS [r]
) AS [r] ON [k].[ParentReportId] = [r].[Id]
LEFT JOIN [Employee] AS [e] ON [k].[EmployeeCode] = [e].[Code]
WHERE ([k].[Discriminator] = 'Success') OR ([k].[Discriminator] = 'Failure')
ORDER BY [r].[RunDate] DESC, [r].[Id]
```

```
SELECT [d].[Id], [d].[DepartmentCode], [d].[Headcount], [d].[InvalidCount], [d].[ParentReportId], [d].[ValidCount], [d0].[Code]
FROM [DepartmentResult] AS [d]
INNER JOIN (
    SELECT DISTINCT TOP(1) [r].[RunDate], [r].[Id]
    FROM [KpiReport] AS [r]
) AS [r] ON [d].[ParentReportId] = [r].[Id]
LEFT JOIN [Department] AS [d0] ON [d].[DepartmentCode] = [d0].[Code]
ORDER BY [r].[RunDate] DESC, [r].[Id]
```

The **DISTINCT TOP(1)** part will result in empty **DepartmentResults** and **KpiResults**

rowanmiller:
@smitpatel can you verify that this is fixed by your `EXISTS` change

smitpatel:
[...]

PR #5435, opened 2016-05-19
Query: Add test for collection include with Skip/Take/FirstOrDefault

smitpatel:
Test for #5108


src/Microsoft.EntityFrameworkCore.Specification.Tests/IncludeTestBase.cs
```diff
@@ -490,6 +490,62 @@ var customers
...
+
+        [Fact]
+        public virtual void Include_collection_order_by_non_key_with_first_or_default()
+        {
+            using (var context = CreateContext())
+            {
+                var customer
+                    = context.Set<Customer>()
+                        .Include(c => c.Orders)
+                        .OrderBy(c => c.City)
```
maumar: make this one OrderByDescending instead

maumar:
:shipit:  after the OrderByDescening change

PR #5447, opened 2016-05-20
SQL Server: Add missing condition to sys.default_constraints joins

### LINKED ISSUE #4821: Alter column selecting and dropping wrong default constraint
Build: `7.0.0-rc2-16649`

I'm using EF7 with an existing database. On one of my tables, I have a column named `Active` that is of type `int`. I set up a migration to alter the column to a `bit`. 

When running the migrations, EF looks for any default constraints on the column and removes them. The SQL generated for removing the default is incorrect. 

Here is my up migration:

```csharp
migrationBuilder.AlterColumn<bool>(
    name: "Active",
    table: "MyTable",
    nullable: false,
    defaultValue: true);
```

and the sql that is generated to remove the default constraint:

```sql
DECLARE @var0 sysname;
SELECT @var0 = [d].[name]
FROM [sys].[default_constraints] [d]
INNER JOIN [sys].[columns] [c] ON [d].[parent_column_id] = [c].[column_id]
WHERE ([d].[parent_object_id] = OBJECT_ID(N'MyTable') AND [c].[name] = N'Active');
IF @var0 IS NOT NULL EXEC(N'ALTER TABLE [MyTable] DROP CONSTRAINT [' + @var0 + ']');
ALTER TABLE [MyTable] ALTER COLUMN [Active] bit NOT NULL;
ALTER TABLE [MyTable] ADD DEFAULT 1 FOR [Active];
```

It's a bit hard to see, but the tricky bit is on the join to `sys.columns`:

```
INNER JOIN [sys].[columns] [c] ON [d].[parent_column_id] = [c].[column_id]
```

This is joining *solely* on the `column_id`. In this case, it selected all defaults on `MyTable` and "cross joined" (for lack of better term) it with any column named `Active` across any table in the db. It ended up selecting another default on `MyTable` and removed it.

It should be joining on `object_id` as well:

```sql
INNER JOIN [sys].[columns] [c] ON [d].[parent_column_id] = [c].[column_id] and [c].[object_id] = object_id(N'MyTable')
```

bricelam:
Fixes #4821

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
@@ -261,7 +261,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations
                 .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name))
                 .Append(" SET READ_COMMITTED_SNAPSHOT ON')");
 
-            EndStatement(builder, suppressTransaction:true);
+            EndStatement(builder, suppressTransaction: true);
         }
 
         protected virtual void Generate(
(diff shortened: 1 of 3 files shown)


smitpatel:
:shipit:

Issue #4833, opened 2016-03-17
Can we avoid needing Commands installed in startup project for Full .NET apps?


rowanmiller:
Currently we require commands to be installed in the startup project, this is nasty from an architecture perspective though. I'm guessing we only introduced this requirement for xproj?

```
PM> Add-Migration NextMigration
Cannot execute this command because Microsoft.EntityFrameworkCore.Commands is not installed in the startup project 'src\RocketProfiler.UI'.
```

bricelam:
It think this may have always been the case. If I remember right, VS wouldn't copy `M.EFC.Commands.dll` to the output directory if it was only transitively referenced.

natemcmaster:
Remarking for triage: this can wait until after RTM.

natemcmaster:
Scope question: are we trying to do this just for csproj/packages.config projects?

rowanmiller:
Any progress we can make would be good, if csproj is easier then that would still be a win. BTW I think it would be really good to relax this restriction... but I also don't think it's a ship stopper for tooling RTM. So I think the decision can take into account how messy the implementation is 😄

natemcmaster:
Here is my thinking. After #5334, here is the tooling in layers of dependency: 

- `.Tools.VisualStudio`/ `.Tools.DotNet`  = `init.ps1` and `dotnet-ef`
- `.Design` = `OperationExecutor`, the version-resilient design-time "entry-point" into the app
- `.Relational.Design` = other design-timey stuff.
- `.Relational` and below = the runtime stuff

The tools layer is the funny bit. `.Tools.VisualStudio` "floats" across all projects in a solution because PMC cmdlets are solution-global, and the `.Tools.DotNet` is disconnected from the app's "dependencies". But both still need the next layer down is why our tools require installing (what used to be .Commands) .Design.

To ease the restrictions, we could:
 - Squash design layers down. We could squash the code in .Design down into .Relational.Design. I'm not sure the separation is as important post #5334.  (I like this idea most)
 - Make the tools layer carry the next layer down with it. Tools copy the design dll's to the right place before running and cleans up after
 - Squash design into runtime. We could be even more aggressive and push it all into .Relational. Downsides: violates our pay-for-play ideals, increases binary size. Upsides: tools will basically work anywhere EF runtime is installed. (I like this idea least)

bricelam:
[...]

PR #5474, opened 2016-05-23
UWP Commands:  remove msbuild copy of net451

### LINKED ISSUE #3682: NuGet Commands: Remove EF.Commands.dll from bin on UWP
Currently, the commands package copies the net451 version of EntityFramework.Commands.dll to the output directory on UWP in order to work.

The problem is that .NET Native will try and compile *all* the assemblies in the output directory. We should look into a solution that doesn't leave this assembly in the output directory when compiling with .NET Native.

natemcmaster:
Only add the net451 assembly to UWP output folder during command execution

cc @bricelam 

Depends on #5398 

Fix #3682


src/Microsoft.EntityFrameworkCore.Tools/build/netcore50/Microsoft.EntityFrameworkCore.Tools.props
```diff
@@ -1,7 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8" ?>
-<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup>
-    <!-- TODO: Consider alternate solutions for working with project.json -->
-    <CopyNuGetImplementations>true</CopyNuGetImplementations>
```
bricelam: Don't we still need this to work with class library projects?

bricelam:
Could we extend this slightly to resolve #4833 too?

bricelam:
:shipit:

src/Microsoft.EntityFrameworkCore.Tools/build/netcore50/Microsoft.EntityFrameworkCore.Tools.props
natemcmaster: What does setting this property actually do?

bricelam: At one point it was the only way to actually get dependencies into the `bin` directory.

natemcmaster: Oh, so this https://github.com/NuGet/Home/issues/1084. This was fixed in VS Update 1. Since we require Update 2 for .NET Native compiler upgrades anyways, it is safe to remove this.

natemcmaster:
Merged to dev in https://github.com/aspnet/EntityFramework/commit/82b994dfea58ab975f9ee637507300d08b355ab5

PR #5476, opened 2016-05-23
PS commands: add catch around GetProperty for non-existent properties

### LINKED ISSUE #5383: Migration fails for separate UWP Class library with new EntityFrameworkCore.SQLite
I have two projects in my UWP solution : one is Blank App (Universal Windows) another is a Class Library (Universal Windows). I installed **Microsoft.EntityFrameworkCore.SQLite –Pre** and **Microsoft.EntityFrameworkCore.Tools –Pre** in the Class library project. When I tried to add the initial migration file with : **Add-Migration MyFirstMigration**, it throws a red coloured exception : 

> Add-Migration : Value does not fall within the expected range.
> At line:1 char:1
> + Add-Migration MyFirstMigration
> + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>     + CategoryInfo          : OperationStopped: (:) [Add-Migration], ArgumentException
>     + FullyQualifiedErrorId : System.ArgumentException,Add-Migration

I was following [this](http://ef.readthedocs.io/en/latest/platforms/uwp/getting-started.html) article.

[EFSQLiteUWP.UI.zip](https://github.com/aspnet/EntityFramework/files/268354/EFSQLiteUWP.UI.zip)

natemcmaster:
Fix #5383.
Possibly also fixes https://github.com/aspnet/EntityFramework/issues/5293

> The Item method throws a ArgumentException exception if the specified value cannot be found in the collection.

https://msdn.microsoft.com/en-us/library/envdte.properties.item(v=vs.120).aspx

e.g. UWP class libraries don't define "ProjectN.UseDotNetNativeToolchain" at all but UWP app projects do.

cc @bricelam

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFramework.psm1
+++ b/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFramework.psm1
@@ -887,12 +887,11 @@ function InvokeOperation($startupProject, $environment, $project, $operation, $a
 }
 
 function GetProperty($properties, $propertyName) {
-    $property = $properties.Item($propertyName)
-    if (!$property) {
+    try {
+        return $properties.Item($propertyName).Value
+    } catch {
         return $null
     }
-
-    return $property.Value
 }
 
 function GetProjectItem($project, $path) {


bricelam:
:shipit:

PR #5482, opened 2016-05-24
Add some guards to relational property metadata builders

### LINKED ISSUE #5384: Int32/Int64 properties with default value automatically annotated with SqlServerValueGenerationStrategy.IdentityColumn in migration
After migrating to EF7 RC2 we've noticed that our `int` and `long` properties that are set with `HasDefaultValue` in the `ModelBuilder` are causing the migration script to annotate columns representing those properties with an identity attribute. This, in turn, causes an exception since defaults cannot be created on identity columns.

Here's a simplified example of the DbContext:

        public class Entity
        {
            public Guid Id { get; set; }
            public int Column1 { get; set; }
        }

        public class TestContext : DbContext
        {
            public DbSet<Entity> Entitys { get; set; }

            public TestContext() { }

            {
                var serverName = @"localhost";
                var dbName = "Test";

                optionsBuilder
                    .UseSqlServer($"data source={serverName};initial catalog={dbName};Integrated Security=true;");
            }

            {
                modelBuilder.Entity<Entity>(entity =>
                {
                    entity
                        .Property(e => e.Column1)
                        .HasDefaultValue(1);
                });
            }
        }

        {
            using (var context = new TestContext())
            {
            }
        }

When creating a migration with Add-Migration, I get the following, which causes an exception when `context.Database.EnsureCreated` is called (*Defaults cannot be created on columns with an IDENTITY attribute. Table 'Entitys', column 'Column1'*): 

        {
            migrationBuilder.CreateTable(
                name: "Entitys",
                columns: table => new
                {
                    Id = table.Column<Guid>(nullable: false),
                    Column1 = table.Column<int>(nullable: false, defaultValue: 1)
                        .Annotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn)
                },
                constraints: table =>
                {
                    table.PrimaryKey("PK_Entitys", x => x.Id);
                });
        }

Is there something I'm missing, or is there a workaround for this?

AndriySvyryd:
Detect conflicting DefaultValue, DefaultValueSql, ComputedColumnSql and SqlServerValueGenerationStrategy
Rename ComputedValueSql extension method to ComputedColumnSql to be consistent with fluent API

Fixes #5384

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalModelValidator.cs
@@ -123,9 +123,9 @@ namespace Microsoft.EntityFrameworkCore.Internal
                                 table));
                         }
 
-                        var currentComputedValueSql = propertyAnnotations.ComputedValueSql ?? "";
-                        var previousComputedValueSql = previousAnnotations.ComputedValueSql ?? "";
-                        if (!currentComputedValueSql.Equals(previousComputedValueSql, StringComparison.OrdinalIgnoreCase))
+                        var currentComputedColumnSql = propertyAnnotations.ComputedColumnSql ?? "";
+                        var previousComputedColumnSql = previousAnnotations.ComputedColumnSql ?? "";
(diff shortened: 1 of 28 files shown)


ajcvickers:
:shipit:

Issue #4952, opened 2016-03-31
Test: Verify unicode behaviour of parameters


smitpatel:
with f416dd9a71a5a6a69715b4ba40a37e6f9da751ef
we have added test `Non_unicode_parameter_is_used_for_non_unicode_column()` though our current testing framework (or may be logging too) does not allow us to see the type of parameter used.

divega:
As far as I remember in EF6 we decided to log all parameter facets that were not set to their default values. Should we do something similar in EF Core?

rowanmiller:
@smitpatel can you talk to @divega and @anpete about some ideas on how to do this

ajcvickers:
@smitpatel I'm going to look at this since I need it for the parameter stuff I am working on.

Issue #5472, opened 2016-05-23
All command logging code is internal


ajcvickers:
All the code for formatting log output for commands and parameters is internal--as in, real internal, not just in an internal namespace. This makes it hard to re-use this code in providers and tests. See RelationalLoggerExtensions.LogCommandExecuted

PR #5494, opened 2016-05-24
Include parameter facets in command log data

ajcvickers:
Related to #4952

Also, start logging facets when this is likely to provide value--e.g. log size when not default, nullable when different from nullability of type, etc.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/TestSqlLoggerFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/TestSqlLoggerFactory.cs
@@ -3,13 +3,16 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data;
 using System.Globalization;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.Extensions.Logging;
 using Xunit.Abstractions;
+using System.Reflection;
 #if !NETSTANDARD1_3
 using System.Runtime.Remoting.Messaging;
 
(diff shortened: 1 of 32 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalLoggerExtensions.cs
```diff
@@ -63,27 +75,165 @@ var logParameterValues
...
+                    builder.Append(buffer[i].ToString("X2", CultureInfo.InvariantCulture));
                 }
-                stringValueBuilder.Append(buffer[i].ToString("X2", CultureInfo.InvariantCulture));
             }
 
-            return stringValueBuilder.ToString();
+            builder.Append('\'');
+        }
+
+        private static bool IsNormalDbType(DbType dbType, Type clrType)
```
anpete: Not sure what "Normal" means here. Is it "default"?

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/TestSqlLoggerFactory.cs
```diff
@@ -146,17 +153,186 @@ private static SqlLoggerData CreateLoggerData()
                 }
             }
 
+            private static string FormatParameter(DbParameterLogData parameterData)
```
anpete: Use product code?

ajcvickers: It's all internal for some reason. See #5472

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalLoggerExtensions.cs
ajcvickers: Kind of, but it's more like a heuristic. Different providers may have different defaults, or may default to nothing at all. So it's like if its a DbType that seems like it would be expected, then don't include it in the output.

anpete:
:shipit:

natemcmaster:
Can't build anymore on Linux or OSX :-/.  https://bugzilla.xamarin.com/show_bug.cgi?id=29167

PR #5501, opened 2016-05-24
Rename PowerShell cmdlets module to EntityFrameworkCore

### LINKED ISSUE #5145: Migrations PowerShell commands can't work for both EF6 and EF Core projects that are part of the same solution
The PowerShell versions of the migrations commands are global to the solution and have the same names in EF6 and EF Core but their implementations are version specific. Only one version of them can be active per solution. This can be very inconvenient while incrementally migrating existing projects to EF Core if at any point projects targeting different versions of EF are part of the same solution.

Possible workarounds customers could use are:
- Uninstall and reinstall the EF packages as needed
- Split the solution (more permanent)

In a quick brainstorming we came up with a few ideas (we haven't validated feasibility):
- Release commands that are compatible with both EF6 and EF Core
- Provide an easy way to unload or swap the active PS module

natemcmaster:
Fix #5145

Related: documenation update https://github.com/aspnet/EntityFramework.Docs/pull/170

~~Btw, we still have "Apply-Migrations" and "Enable-Migrations" in this module. Can we remove these now? Or do we still want them there to throw "obsolete" errors?~~ nevermind. Just realized Brice opened an issue a few hours before me to track this exact question.

cc @rowanmiller @bricelam @divega

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psd1
+++ b/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psd1
@@ -1,12 +1,12 @@
 @{
     # Script module or binary module file associated with this manifest
-    ModuleToProcess = 'EntityFramework.psm1'
+    ModuleToProcess = 'EntityFrameworkCore.psm1'
 
     # Version number of this module.
-    ModuleVersion = '7.0'
+    ModuleVersion = '1.0.0'
 
     # ID used to uniquely identify this module
-    GUID = '06e6973e-66a3-4cfc-9a42-3f7d159e945a'
+    GUID = 'c126fb40-c0f1-43ae-8dd0-06bb50512eb2'
 
     # Author of this module
     Author = 'Entity Framework Team'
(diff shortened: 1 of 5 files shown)


src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psd1
```diff
@@ -58,14 +58,15 @@
...
-        'Add-Migration',
+        'Enable-Migrations',
         'Apply-Migration',
-        'Update-Database',
-        'Script-Migration',
+        
+        'Add-Migration',
         'Remove-Migration',
-        'Enable-Migrations',
         'Scaffold-DbContext'
```
lajones: Does the Scaffold-DbContext line now need a comma?

natemcmaster: Oops. Yes. Will fix.

bricelam:
:shipit:

src/Microsoft.EntityFrameworkCore.Tools/tools/about_EntityFramework.help.txt
```diff
@@ -7,14 +7,14 @@
...
+    about_EntityFrameworkCore
 
 SHORT DESCRIPTION
-    Provides information about Entity Framework commands.
+    Provides information about Entity Framework Core commands.
 
 LONG DESCRIPTION
-    This topic describes the Entity Framework commands. Entity Framework is Microsoft's recommended data access technology for new
-    applications.
+    This topic describes the Entity Framework Core commands. Entity Framework is Microsoft's recommended data access technology
```
divega: I know this is merged already, sorry I still need to comment... I was thinking how we can fix it, but can we get rid of the second sentence altogether? Maybe a link to our documentation page would be more useful 

cc @rowanmiller

rowanmiller: Agreed...

> This topic describes the Entity Framework Core commands. See https://docs.efproject.net for information on Entity Framework Core.

PR #5502, opened 2016-05-24
Add explicit cast to double literals in snapshots

### LINKED ISSUE #5455: Migration error with default double value
When you have a double property that has a default double value, then `ef database update` command throws 
`Cannot set default value '0' of type 'System.Int32' on property 'XXX' of type 'System.Double' in entity type 'YYY'.`

Fluent configuration:
`mb.Property(p => p.Uptime).HasDefaultValue<double>(0.0).IsRequired();`

Created migration file:
```
migrationBuilder.AlterColumn<double>(
    name: "Uptime",
    table: "ServersStatusRate",
    nullable: false,
    defaultValue: 0);
```

Snapshot file:
```
b.Property<double>("Uptime")
     .ValueGeneratedOnAdd()
     .HasDefaultValue(0);
```

AndriySvyryd:
Fixes #5455

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools.Core/Migrations/Design/CSharpHelper.cs
+++ b/src/Microsoft.EntityFrameworkCore.Tools.Core/Migrations/Design/CSharpHelper.cs
@@ -309,7 +309,8 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Design
             "new DateTimeOffset(" + Literal(value.DateTime) + ", " + Literal(value.Offset) + ")";
 
         public virtual string Literal(decimal value) => value.ToString(CultureInfo.InvariantCulture) + "m";
-        public virtual string Literal(double value) => value.ToString("R", CultureInfo.InvariantCulture);
+        public virtual string Literal(double value) => EnsureDecimalPlaces(value.ToString("R", CultureInfo.InvariantCulture));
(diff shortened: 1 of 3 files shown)


src/Microsoft.EntityFrameworkCore.Tools.Core/Migrations/Design/CSharpHelper.cs
```diff
@@ -309,7 +309,7 @@ public virtual string Namespace([NotNull] params string[] name)
             "new DateTimeOffset(" + Literal(value.DateTime) + ", " + Literal(value.Offset) + ")";
 
         public virtual string Literal(decimal value) => value.ToString(CultureInfo.InvariantCulture) + "m";
-        public virtual string Literal(double value) => value.ToString("R", CultureInfo.InvariantCulture);
+        public virtual string Literal(double value) => "(double)" + value.ToString("R", CultureInfo.InvariantCulture);
```
bricelam: Can we just add `.0` to the end if it doesn't have any decimal places?

AndriySvyryd: yes

AndriySvyryd:
Updated

bricelam:
:shipit:

Issue #5428, opened 2016-05-19
RevEng: Do not generate ValueGenerated.Never for key columns with default values


lajones:
See #5366. For the SQL Server Scaffolding provider we currently generate `ValueGenerated.Never` for single integer or Guid key columns unless they are marked as IDENTITY columns. See https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.SqlServer.Design/Internal/SqlServerScaffoldingModelFactory.cs#L92. We need to also _not_ generate `ValueGenerated.Never` for the same key columns if the `ColumnModel` has non-null `ValueGenerated` or `DefaultValue`.

lajones:
Fix checked in with 552d6ec.

lajones:
Adding the `providers-beware` tag just in case any provider out there is re-using this code.

PR #5504, opened 2016-05-24
RevEng: Do not produce ValueGenerated.Never for PK key columns with default values

lajones:
Fix for #5428. For SQL Server We used to generate a `ValueGeneratedNever()` call for single primary key columns which generated their value through a default value which referenced a sequence. Now we no longer do that.

Also renamed some tests which said they were testing identity columns with default values (which is not allowed) - they are actually testing PK columns with a sequence (which is allowed).


ajcvickers:
:shipit:

lajones:
Checked in with 552d6ec.

PR #5507, opened 2016-05-25
Fix to 4355 - Contains on Sub query with entity type generates invalid SQL

maumar:
There are several issues around the area:
- in some cases we were forcing client evaluation too aggressively when trying to translate SubQueryExpression. Fix is to relax those criteria, so that more queries that we can translate to SQL will actually be translated.
- in other cases we were not translating expressions that could have been translated because we were not looking into queries processed by parent EntityQueryExpressionVisitor during binding,
- in general, if we were unable to translate the item expression of the contains, we were not applying client-side result operator, which would produce incorrect results when client evaluation was really needed,
- we were handling Contains in two places separately: RelationalResultOperatorHandler and SqlTranslatingExpressionVisitor -> VisitSubquery, now that the RelationalResultOperatorHandler logic has been improved, code in the the other place can be removed.


maumar:
@anpete  @mikary

mikary:
LGTM

anpete:
:shipit:

PR #5518, opened 2016-05-25
Add runtime directive for System.Collections.ArrayList

### LINKED ISSUE #5517: UWP: property 'Count' not found on type 'System.Collections.ArrayList'
This issue has been described in other bugs, but this one consolidates and provides a workaround. This will also track progress with the .NET Native team and/or EF including a fix.

## The Issue
At runtime, UWP apps using EF Core RC1 and RC2 may encounter this error but only in "Release" mode.
```
Property 'Count' was not found on type 'System.Collections.ArrayList'
```

## Reason
The .NET Native compiler, which runs on UWP apps in "release" mode, aggressively removes metadata required for reflection. This can be controlled by adding runtime directives. Although the .NET Native compiler improved greatly in 1.3.2, EF hits a common limitation of the compiler. Earlier versions of EF docs instructed users to include this directive:
```xml
 <Type Name="System.Collections.ArrayList" Dynamic="Required All"/>
```

Despite having this runtime directive, users may still encounter this error. This happens in some apps when additional dependencies also bring in System.Xml.XmlSerializer. When this occurs, the .NET Native compiler removes metadata for ArrayList.

> warning : ILTransform_0027: Ambiguous type name detected. There are types with name 'System.Collections.ArrayList' in both assembly 'System.Xml.XmlSerializer' and 'System.Collections.NonGeneric'. This directive is ignored. Consider placing this directive within an Assembly scope.

## Workaround
Users can include this in "Default.rd.xml" to instruct the compiler to include the necessary metadata.

```xml
    <Assembly Name="System.Collections.NonGeneric">
      <Type Name="System.Collections.ArrayList" Dynamic="Required All"/>
    </Assembly>
```

natemcmaster:
I was too aggressive in removing rdxml that it turns out we still need.

Fix #5517

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/Microsoft.EntityFrameworkCore.rd.xml
+++ b/src/Microsoft.EntityFrameworkCore/Properties/Microsoft.EntityFrameworkCore.rd.xml
@@ -2,7 +2,10 @@
 
 <Directives xmlns="http://schemas.microsoft.com/netfx/2013/01/metadata">
   <Library>
-    <!-- TODO optimize this for types that actually need reflection -->
     <Assembly Name="Microsoft.EntityFrameworkCore" Dynamic="Required All" />
+
+    <Assembly Name="System.Collections.NonGeneric">
+      <Type Name="System.Collections.ArrayList" Dynamic="Required All"/>
+    </Assembly>
   </Library>
 </Directives>


natemcmaster:
cc @divega

divega:
Had a chat with @natemcmaster the reason reflection metadata is needed for System.Collections.ArrayList is that Re-LINQ uses it. We can follow up with them on changing the way they are accessing it, but for now adding the directive seems to be the right thing to do.

divega:
:shipit:

Issue #5397, opened 2016-05-17
Updating entity with optional one-to-one relationship


Thealexbarney:
I just upgraded from RC1 to RC2, and I'm seeing some different behavior from RC1 with this code:
https://gist.github.com/Thealexbarney/2cfec11d8cbd211ff89e8dad51798248

When you update an entity with an optional one-to-one relationship using the other entity in that relationship, RC1 will update the foreign key, but RC2 will not, making it so you have to manually update the foreign keys.

Is this intended behavior? Did I get something wrong with my relationships? Or is this a regression?

PR #5521, opened 2016-05-25
Fix detecting reference to untracked entity

ajcvickers:
Issue #5397

The problem was that before attaching the new entity we were not recording that it had a navigation pointing to it, so then in initial fixup this was not being used to populate the FK.


anpete:
:shipit:

PR #5523, opened 2016-05-25
Fix #5439 - Select null coalesce bool throws InvalidCastException

### LINKED ISSUE #5439: RC2, using ?? operator in select fails with casting error
RC2, using ?? operator in select fails with casting error. 

`.Select(x => x.NullableBooleanProperty ?? false)` 

would fail with casting error. But the same like this is ok

`.Select(x => x.NullableBooleanProperty == null ? false : x.NullableBooleanProperty.value)`

mikary:
An exception was being thrown when projecting with null coalesce on a bool value on SQL Server because the generated SQL was not casting the numeric value back into a byte for materialization.
@maumar @anpete

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -3445,7 +3445,7 @@ WHERE [c].[CustomerID] = N'ALFKI'",
             Assert.Contains(
                     @"SELECT [p].[ProductID], [p].[Discontinued], [p].[ProductName], [p].[UnitsInStock]
 FROM [Products] AS [p]
-WHERE (@__flag_0 = 1 AND ([p].[UnitsInStock] >= 20)) OR ((@__flag_0 <> 1) AND ([p].[UnitsInStock] < 20))",
+WHERE (@__flag_0 = 1 AND ([p].[UnitsInStock] >= 20)) OR ((@__flag_0 <> CAST(1 AS BIT)) AND ([p].[UnitsInStock] < 20))",
                     Sql);
         }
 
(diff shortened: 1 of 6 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -3445,7 +3445,7 @@ public override void Where_ternary_boolean_condition()
             Assert.Contains(
                     @"SELECT [p].[ProductID], [p].[Discontinued], [p].[ProductName], [p].[UnitsInStock]
 FROM [Products] AS [p]
-WHERE (@__flag_0 = 1 AND ([p].[UnitsInStock] >= 20)) OR ((@__flag_0 <> 1) AND ([p].[UnitsInStock] < 20))",
+WHERE (@__flag_0 = 1 AND ([p].[UnitsInStock] >= 20)) OR ((@__flag_0 <> CAST(1 AS BIT)) AND ([p].[UnitsInStock] < 20))",
```
maumar: i think we should either do CAST(1 AS BIT) everywhere or just for the coalesce case (global vs local fix), rather than make it more nuanced, like it seems to be according to this example - i.e.  @__flag_0 = 1 but @__flag_0 <> CAST(1 AS BIT)

mikary:
This is also an opportunity to consider if we want the default behavior for all providers to be `CAST(1 AS BIT)` or just `1` for a bool literal. The `DefaultQuerySqlGenerator` still has a few literals (`TrueLiteral`, `FalseLiteral`, `"1 = 1"`, and `"1 = 0"`) for various expressions where it's not using the `ISqlGenerationHelper`

smitpatel:
`1 = 1` should have been gone. It was used to say true in where clause but in current code we don't generate where clause for that.
`1 = 0` is to show always false condition in where clause. Which is not related to any type.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
divega: Had a chat with @mikary and @maumar about this and there is going to be another iteration of this PR. A couple of issues are:
- We are not being very consistent about it (what @maumar called out above)
- It is not clear whether we should keep assuming the existence of BIT type and the need to CAST to it across all relational databases and have providers that don't support it opt out or whether we should do the opposite. Regardless, the next iteration of the PR should be "providers-beware".

cc @anpete

divega:
It also seems we could cast to BIT only in the top projection if that is the only thing that throws?

Issue #4608, opened 2016-02-19
Use column/property facets for parameter types in Query Pipeline


carlcamilleri:
Unless I'm missing something, parameterised queries (against SQL Server in my case) are treating all strings as NVARCHAR

I have the below table definition:
```C#
modelBuilder.Entity<Products>(entity =>
            {
                entity.ToTable("Products", "ofr");

                entity.HasIndex(e => e.ProductId).HasName("IX_Products_ProductId").IsUnique();

                entity.HasIndex(e => new { e.Id, e.ProductId, e.PublicationId }).HasName("IX_ofr_Products_PublicationId");

                entity.Property(e => e.ProductId)
                    .IsRequired()
                    .HasMaxLength(50)                    
                    .HasColumnType("VARCHAR");
            });
```

Then a query:
```C#
 var q = from p in products

                        join sp in dbContext.SiteProducts
                        on new { Id = p.Id, ProductId = p.ProductId }
                        equals new { Id = sp.ProdId, ProductId = req.ProductId}
             
                        where   sp.SiteId == req.Engine.Context.SiteId                                            
                        
                        select p;
var res = q.ToList();
```

The query generated is:
```SQL
exec sp_executesql N'SELECT TOP(1) [p].[Id], [p].[ProductId], [p].[PublicationId]
FROM [ofr].[Products] AS [p]
INNER JOIN [ofr].[SiteProducts] AS [sp] ON ([p].[Id] = [sp].[ProdId]) AND ([p].[ProductId] = @__req_ProductId_0)
WHERE [sp].[SiteId] = @__req_Engine_Context_SiteId_1',N'@__req_ProductId_0 nvarchar(4000),@__req_Engine_Context_SiteId_1 int',@__req_ProductId_0=N'350003995',@__req_Engine_Context_SiteId_1=1
```

The "ProductId" parameter is therefore treated as an NVARCHAR, obviously missing the index.

I have tested with EF6 against the same database, and the query generated correctly uses a VARCHAR parameter:
```SQL
exec sp_executesql N'SELECT 
    [Extent1].[Id] AS [Id], 
    [Extent1].[ProductId] AS [ProductId], 
    [Extent1].[PublicationId] AS [PublicationId]
    FROM  [ofr].[Products] AS [Extent1]
[...]

PR #5525, opened 2016-05-25
Use length facet for parameter size

ajcvickers:
Issues:
#4608 Use column/property facets for parameter types in Query Pipeline
#4134 Use column/property facets for parameter types in Update Pipeline

If a property length is specified, then this is used to infer the length to use for parameters relating to that property, unless the length of the data is too long, in which case unbounded length is used. Because query has code that can infer the length to use for a parameter this still means that fragmentation will not happen without always using the 4000/8000 value.


anpete:
:shipit:

PR #5527, opened 2016-05-26
RevEng: Use a naming service to produce candidate names for identifiers

### LINKED ISSUE #3987: Identifier generation strategy in reverse engineering

lajones:
Fixes #3987 - see that issue for details of the new rules for generating identifiers. Added as a `CandidateNamingService` which encapsulates the generation of candidate names from DB names. (It includes what `ModelUtilities` used to do plus adds one method).

Most of the changes are changes to expected results because of the new rules.


ajcvickers:
:shipit:

lajones:
Checked in with f2f90ca.

Issue #5519, opened 2016-05-25
Include().ThenInclude and order by throws exception


Tasteful:
Tested with and without #5443 
```
[Fact]
        public virtual void Then_include_collection_order_by_collection_column()
        {
            using (var context = CreateContext())
            {
                var customer
                    = context.Set<Customer>()
                        .Include(c => c.Orders)
                        .ThenInclude(o => o.OrderDetails)
                        .Where(c => c.CustomerID.StartsWith("W"))
                        .OrderByDescending(c => c.Orders.OrderByDescending(oo => oo.OrderDate).FirstOrDefault().OrderDate)
                        .FirstOrDefault();

                Assert.NotNull(customer);
                Assert.NotNull(customer.Orders);
                Assert.NotEmpty(customer.Orders);
                Assert.NotNull(customer.Orders.First().OrderDetails);
                Assert.NotEmpty(customer.Orders.First().OrderDetails);
            }
        }
```

throws exception (this exception after #5443 is applied)
```
Test Name:	IncludeSqlServerTest.Then_include_collection_order_by_collection_column
Test FullName:	Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.IncludeSqlServerTest.Then_include_collection_order_by_collection_column
Test Source:	C:\Git\EntityFramework\test\Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests\SqlAzure\Model\Product.cs : line 27
Test Outcome:	Failed
Test Duration:	0:00:01,783

Result StackTrace:	
at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at System.Data.SqlClient.SqlDataReader.get_MetaData()
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
[...]

PR #5443, opened 2016-05-20
Query: Include collection and order by subquery throws 'Variable names must be unique within a query batch or stored procedure.'

Tasteful:
Replaces PR #5405 (that I messad up when syncing code with dev)

@anpete @smitpatel

Issue #5421, opened 2016-05-19
Query: Lift subquery in INNER JOIN to generate flattened query


smitpatel:
For collection navigation include, we generate query of following structure at present

```
SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
FROM [Orders] AS [o]
INNER JOIN (
    SELECT DISTINCT [c].[CustomerID]
    FROM [Customers] AS [c]
    WHERE [C].[CustomerID] = N'ALFKI'
) AS [c0] ON [o].[CustomerID] = [c0].[CustomerID]
ORDER BY [c0].[CustomerID]
```

It can be written as

```
SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
FROM [Customers] AS [c]
INNER JOIN [Orders] AS [o]
ON [o].[CustomerID] = [c].[CustomerID]
WHERE [C].[CustomerID] = N'ALFKI'
ORDER BY [c].[CustomerID]
```

Most of the work this change is available in @anpete's fork.

divega:
Discussed this with @anpete and @smitpatel and I think we can do it as part of the "EXISTS" improvements @smitpatel is already doing, but would like to make the call in triage.

ErikEJ:
@divega  Will those "EXISTS" improvements also affect https://github.com/aspnet/EntityFramework/issues/2626 ?

smitpatel:
Exists improvements tracked by #4903

smitpatel:
@ErikEJ - No.

smitpatel:
While further exploring translations, I found that if the inner subquery has top or offset or needs DISTINCT then subquery may be required. We cannot convert it to flattened query.
With #4903 we have already removed inner join itself if there is no order by on non-principal-key column. So at present this translation seems to apply only the cases when there is no distinct/top/offset and there is order by on non-principal-key column. Considering that we may not need to do this change in RTM. Removing milestone. 

This translation can be applied to any Inner Join query so we may want to do it as a subquery lifting optimizer in future.

divega:
Moving to 1.0.1 for reconsideration based on @smitpatel's comment. cc @rowanmiller

ajcvickers:
@smitpatel Can we just close this one?

smitpatel:
@anpete - Thoughts?

ajcvickers:
Code has moved forward and this no longer has value.

Issue #5038, opened 2016-04-11
Query: Include & Order By subquery generates invalid query.


smitpatel:
Test
````c#
[Fact]
public virtual void Include_collection_order_by_collection_column()
{
	using (var context = CreateContext())
	{
		var customer
			= context.Set<Customer>()
				.Include(c => c.Orders)
				.Where(c => c.CustomerID.StartsWith("w"))
				.OrderByDescending(c => c.Orders.OrderByDescending(oo => oo.OrderDate).FirstOrDefault())
				.FirstOrDefault();

		Assert.NotNull(customer);
		Assert.NotNull(customer.Orders);
		Assert.NotEmpty(customer.Orders);
	}
}
```
Generated SQL
```SQL
SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
FROM [Orders] AS [o]
INNER JOIN (
    SELECT DISTINCT TOP(1) (
        SELECT TOP(1) 1
        FROM [Orders] AS [oo]
        WHERE [c].[CustomerID] = [oo].[CustomerID]
        ORDER BY [oo].[OrderDate] DESC
    ) AS [c0_0], [c].[CustomerID]
    FROM [Customers] AS [c]
    WHERE [c].[CustomerID] LIKE N'w' + N'%'
    ORDER BY (
        SELECT TOP(1) 1
        FROM [Orders] AS [oo]
        WHERE [c].[CustomerID] = [oo].[CustomerID]
        ORDER BY [oo].[OrderDate] DESC
    ) DESC, [c].[CustomerID]
) AS [c0] ON [o].[CustomerID] = [c0].[CustomerID]
ORDER BY [c0_0] DESC, [c0].[CustomerID]
````
Error - `ORDER BY items must appear in the select list if SELECT DISTINCT is specified.`
We introduce the `Distinct` keyword for include. Since the `Order By` item is subquery, even if that appears in select list too, Sql Server cannot identify them as same subquery hence throws exception.
We either need to use alias expression for such subquery or revisit usage of `Distinct`.

smitpatel:
fixed via c97c523c179df16600528d67e5048899f0131c22

PR #5531, opened 2016-05-26
order by subquery and using theninclude throws exception

Tasteful:
This contains test case and fix for #5519. This also contains the fixes from #5443.

Fixes bug #5519, #5421, #5038 
Potential fix for #4695

@smitpatel Didn't get any answer on I should include in previous PR or create a new one so created a new one. If I should include them in the other one I will merge and squash them, otherwise I will squash the commits here after #5543 is merged to get a clean PR for the theninclude bug.

ping @anpete

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -788,22 +788,27 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 
             if (columnExpression != null)
             {
-                return new ColumnExpression(columnExpression.Name, columnExpression.Property, tableExpression);
+                return columnExpression.Property == null
+                    ? new ColumnExpression(columnExpression.Name, columnExpression.Type, tableExpression)
+                    : new ColumnExpression(columnExpression.Name, columnExpression.Property, tableExpression);
             }
 
             var aliasExpression = expression as AliasExpression;
 
             if (aliasExpression != null)
             {
-                var newAliasExpression
-                    = new AliasExpression(
-                        aliasExpression.Alias,
-                        UpdateColumnExpression(aliasExpression.Expression, tableExpression))
-                    {
-                        SourceMember = aliasExpression.SourceMember
-                    };
+                var selectExpression = aliasExpression.Expression as SelectExpression;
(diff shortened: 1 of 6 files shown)


smitpatel:
@Tasteful - I have merged the other PR. Can you rebase this on latest dev & squash commits?

Tasteful:
@smitpatel - PR is rebased and squashed. Also added the extra asserts that you was adding for the other PR.

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -788,22 +788,27 @@ public override string ToString()
 
             if (columnExpression != null)
             {
-                return new ColumnExpression(columnExpression.Name, columnExpression.Property, tableExpression);
+                return columnExpression.Property == null
```
smitpatel: Having a `ColumnExpression` with null property doesn't look very good. @anpete - thoughts?

divega: @smitpatel and I talked and reviewed other places where it was already possible to use `ColumnExpression` without a property, e.g. by using the constructor that takes a type. It seems that this is ok, and that the change in `ColumnExpression.Equals()` potentially fixes a NRE we just didn't know about.

Tasteful: @smitpatel @divega Fix in `ColumnExpression.Equals()` was because of an NRE. Another solution to not allow null property in the `ColumnExpression` is to add a new expression, lets say `AliasColumnExpression` that will do the same as `ColumnExpression` that not have an property assigned. If that should be introduced my feeling is that it's another PR because it will change other places where the `ColumnExpression` is used also.

smitpatel: @Tasteful - We found some other places also where `ColumnExpression` is used without property. The absence of NRE was just co-incidence.
I am trying to merge this. After rebasing on latest dev there is minor issue in generated Sql. If you can change it now else I will update it while merging.

smitpatel:
merged with 0fe95fa88c69080a71fb87c3d4c391c2cb2fbbe7
@Tasteful  - Thank you contribution 😄

PR #5538, opened 2016-05-26
Fixes #5333 - More generic solution for severity level of warnings

### LINKED ISSUE #5333: More generic solution for severity level of warnings (throw/warn/ignore)

anpete:
- Adds ```DbContextOptionsBuilder.ConfigureWarnings(Action<WarningsConfigurationBuilder>)``` fluent API. The API has two main functions:
  1. Allow a default warning behavior to be set: ```builder.DefaultBehavior(WarningBehavior)``` default is 'Log'
  2. Configure explicit behavior for specific event ids. Throw, Log or Ignore. E.g. ```builder.Throw(CoreLoggingEventId.IncludeIgnoredWarning)```
- Re-uses our logging event ids as the way to identify specific warning kinds.
- Adds relational fluent extension methods on builder for relational event ids.
- Warning exception generation logic centralized in InterceptingLogger via LogWarning EF extensions.
- Replaces previous ad-hoc warnings: query client eval and ambient transaction detected.


anpete:
@rowanmiller @divega New version of this

rowanmiller:
API looks good to me. When we write errors, do we give the name (not value) of the corresponding EventId? I.e. is it easy for a dev to work out they need to use `CoreLoggingEventId.IncludeIgnored`?

divega:
Maybe we could rename `DefaultBehavior()` to `Default()`, e.g.:

``` C#
.ConfigureWarnings(warnings => warnings
   .Default(WarningBehavior.Throw)
   .Ignore(CoreLoggingEventId.IncludeIgnoredWarning));
```

Instead of 

``` C#
.ConfigureWarnings(warnings => warnings
   .DefaultBehavior(WarningBehavior.Throw)
   .Ignore(CoreLoggingEventId.IncludeIgnoredWarning));
```

But maybe not :smile: What do you think?

Maybe the two things we already had become slightly less discoverable, but I think this is nice! :clap::clap::clap:

rowanmiller:
@divega `Default(...)` looks good to me 👍

anpete:
@rowanmiller @divega Will change to Default - looks good to me. And, @rowanmiller, this is what errors look like:

```
Warning as error exception for warning 'IncludeIgnoredWarning': The Include operation for navigation: 'e.Nav' was ignored because the target navigation is not reachable in the final query results.
```

rowanmiller:
@anpete is there any way to easily get the fully qualified EventId enum name/value in there? Also, do we write the EventId name when we log the warning?

anpete:
@rowanmiller I think that should be do-able. The first PR did it - it was easier because the APIs were different. WRT the event id and logging, no. Are you thinking about how people go from seeing the warning in logging to making a corresponding change in ConfigureWarnings?

rowanmiller:
> Are you thinking about how people go from seeing the warning in logging to making a corresponding change in ConfigureWarnings?

Yes. In particular if they want to suppress the warning.

anpete:
@divega @rowanmiller Another thing we should consider: Generalize our event ids be renaming to remove "Logging". I.e. CoreLoggingEventId -> CoreEventId

rowanmiller:
@anpete 👍 to the rename. If we don't get the fully qualified name in to this change (which is fine) lets just make sure we have an issue tracking it.

anpete:
@rowanmiller Fully qualified done. Proposal for logged warnings:

"The LINQ expression '[c].IsLondon' could not be translated and will be evaluated locally. This warning can be configured using DbContextOptionsBuilder.ConfigureWarnings with event id 'RelationalLoggingEventId.QueryClientEvaluationWarning'."

rowanmiller:
Message looks good to me

rowanmiller:
Super happy we are getting this into RTM 😄

divega:
@ajcvickers to do final CR.

ajcvickers:
:shipit:

divega:
A few more things came to mind:
- It seems we used to throw by default on ambient transactions? Will this change it to a log by default? If yes,  I think it is OK, if no, it is inconsistent
- Should we treat in-memory's IgnoreTransaction() in a similar way? 
- Are we tying our hands when we say this API is about warnings? What if in the future we find an event that is bad enough we want it to throw by default, but we want to be able to configure it to log or be ignored? Would that also be considered a warning?

PR #5540, opened 2016-05-26
Throw on multiple relationships using one navigation instead of letting the last one win

### LINKED ISSUE #5344: Multiple one-to-one relations create duplicate foreign key field.
Have 2 entities
```

public class Customer {
  public int Id {get; set;}
  public string Name {get; set;}
  public int? ShipToAddressId {get; set;}
  public int? BillToAddressId {get;set;}
  public virtual CustomerAddress ShipToAddress {get;set;}
  public virtual CustomerAddress BillToAddress {get;set}
  public virtual ICollection<CustomerAddress> CustomerAddresses {get;set;} 
}

public class CustomerAddress {
  public int Id {get;set;}
  public int CustomerId {get;set;}
  public string Name {get;set;}
  public string Address {get;set;}
  public virtual Customer Customer {get;set'}
}
```

with the `following` Fluent API in DbContext OnModelCreating method

            var customeraddress = builder.Entity<CustomerAddress>();
            customeraddress.HasKey(t => t.Id);
            customeraddress.Property(t => t.Id).ValueGeneratedOnAdd();
            customeraddress.Property(t => t.Name).IsRequired();
            customeraddress.Property(t => t.Address).IsRequired();

            var customer = builder.Entity<Customer>();
            customer.HasKey(t => t.Id);
            customer.Property(t => t.Id).ValueGeneratedOnAdd();
            customer.Property(t => t.Name).IsRequired();
            customer.HasOne(t => t.BillToAddress)
                    .WithOne(c => c.Customer)
                    .HasForeignKey<Customer>(f => f.BillToAddressId);
            customer.HasOne(t => t.ShipToAddress)
                    .WithOne(c => c.Customer)
                    .HasForeignKey<Customer>(f => f.ShipToAddressId);   

this generates an extra field called "BillToAddressId1" in Customers Table instead of using already existing "BillToAddressId" field.
So Customers table gets created with both fields.

thanks

AndriySvyryd:
Throw when configuring a bidirectional relationship that conflicts with a previously configured biderectional relationship on one of the navigations.

Fixes #5344

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Builders/CollectionNavigationBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Builders/CollectionNavigationBuilder.cs
@@ -1,9 +1,11 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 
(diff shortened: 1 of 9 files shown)


AndriySvyryd:
@rowanmiller Please review the exception message:

> Cannot create a relationship between 'Customer.CompletedOrders' and 'Order.Customer', because there already is a relationship between 'Customer.OutstandingOrders' and 'Order.Customer'. Navigation properties can only participate in a single relationship.

rowanmiller:
:shipit: for the exception message

smitpatel:
:shipit:

PR #5541, opened 2016-05-26
[TINY] Revert "Fixing build break"

rowanmiller:
This reverts commit 242d53cfc49dd40932ff3b5cdc0639893d729ad6.

Adding in the missing "Microsoft." that caused the build to break. Not really sure why this only started failing recently since the code has been like this for a while.


AndriySvyryd:
:shipit:

PR #5542, opened 2016-05-26
Remove dependency on Microsoft.DotNet.Cli.Utils

### LINKED ISSUE #5012: "Done" message was shown when running dotnet ef migrations script

### LINKED ISSUE #5483: Release Build -> Yields M.E.PlatformAbstractions complications
Posting this here since it is the only library right now that I am using in conjunction with Template10 for Sqlite support.   It appears that the Microsoft.EntityFrameworkCore.Tools are part of the problem with M.E.PlatformAbstractions, since they are the only tool that apparently references that library

1>C:\Program Files (x86)\MSBuild\Microsoft\.NetNative\x86\ilc\IlcInternals.targets(936,5): warning : ILTransform_0000:       MCG : warning MCG0006: Unresolved P/Invoke method 'sysctl!libc' in assembly 'Microsoft.Extensions.PlatformAbstractions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' because it is not available in UWP applications. Please either use an another API , or use [DllImport(ExactSpelling=true) to indicate that you understand the implications of using non-UWP application APIs.

1>C:\Program Files (x86)\MSBuild\Microsoft\.NetNative\x86\ilc\IlcInternals.targets(936,5): warning : ILTransform_0000:       MCG : warning MCG0006: Unresolved P/Invoke method 'uname!libc' in assembly 'Microsoft.Extensions.PlatformAbstractions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' because it is not available in UWP applications. Please either use an another API , or use [DllImport(ExactSpelling=true) to indicate that you understand the implications of using non-UWP application APIs.

1>C:\Program Files (x86)\MSBuild\Microsoft\.NetNative\x86\ilc\IlcInternals.targets(936,5): warning : ILTransform_0000:       MCG : warning MCG0006: Unresolved P/Invoke method 'RtlGetVersion!ntdll' in assembly 'Microsoft.Extensions.PlatformAbstractions, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' because it is not available in UWP applications. Please either use an another API , or use [DllImport(ExactSpelling=true) to indicate that you understand the implications of using non-UWP application APIs.

---project.lock.json --
"Microsoft.EntityFrameworkCore.Tools/1.0.0-preview1-final": {
        "type": "package",
        "dependencies": {
          "Microsoft.EntityFrameworkCore.Tools.Core": "1.0.0-rc2-final",
          "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-final",
          "System.Collections.NonGeneric": "4.0.0",
          "System.IO": "4.0.10",
          "System.IO.FileSystem": "4.0.0"
        },
        "compile": {
          "lib/netcore50/Microsoft.EntityFrameworkCore.Tools.dll": {}
        },
        "runtime": {
          "lib/netcore50/Microsoft.EntityFrameworkCore.Tools.dll": {}
        }
      },

natemcmaster:
Replace use of their console reporter with our own
Copy over a few of the command line helpers
Remove use of stderr for normal execution output from dotnet-ef

Fix #5012 
Incidentally fixes #5483 by removing dependency on Microsoft.Extensions.PlatformAbstractions

cc @bricelam

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools.Cli/ConsoleCommandLogger.cs
+++ b/src/Microsoft.EntityFrameworkCore.Tools.Cli/ConsoleCommandLogger.cs
@@ -1,32 +1,76 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using JetBrains.Annotations;
-using Microsoft.DotNet.Cli.Utils;
 using Microsoft.EntityFrameworkCore.Design.Internal;
+using Newtonsoft.Json;
 
 namespace Microsoft.EntityFrameworkCore.Tools.Cli
 {
     public class ConsoleCommandLogger : CommandLogger
     {
+        private static object _lock = new object();
+        public static bool IsVerbose { get; set; }
+
+        public static void Output(string message)
+        {
+            lock (_lock)
+            {
+                Console.WriteLine(message);
+            }
+        }
+
+        public static void Error(string message)
+        {
+            lock (_lock)
+            {
+                Console.Error.WriteLine(message);
+            }
+        }
+
+        public static void Verbose(string message)
+        {
+            if (IsVerbose)
+            {
+                Output(message);
+            }
+        }
+
+        public static void Json(object result, bool delimited)
+        {
+            lock (_lock)
+            {
+                if (delimited)
+                {
+                    Console.WriteLine("//BEGIN");
+                }
+                Console.WriteLine(JsonConvert.SerializeObject(result, Formatting.Indented));
+                if (delimited)
+                {
+                    Console.WriteLine("//END");
+                }
+            }
+        }
+
         public ConsoleCommandLogger([NotNull] string name)
             : base(name)
         {
         }
 
-        protected override void WriteDebug(string message)
-            => Reporter.Verbose.WriteLine(message.Bold().Black());
+        protected override void WriteTrace(string message)
+            => Verbose(message.Bold().Black());
 
-        protected override void WriteError(string message)
-            => Reporter.Error.WriteLine(message.Bold().Red());
+        protected override void WriteDebug(string message)
+            => Verbose(message.Bold().Black());
 
         protected override void WriteInformation(string message)
-            => Reporter.Error.WriteLine(message);
-
-        protected override void WriteTrace(string message)
-            => Reporter.Verbose.WriteLine(message.Bold().Black());
+            => Output(message);
 
         protected override void WriteWarning(string message)
-            => Reporter.Error.WriteLine(message.Bold().Yellow());
+            => Output(message.Bold().Yellow());
+
+        protected override void WriteError(string message)
+            => Error(message.Bold().Red());
     }
 }
--- a/src/Microsoft.EntityFrameworkCore.Tools.Cli/ExecuteCommand.cs
+++ b/src/Microsoft.EntityFrameworkCore.Tools.Cli/ExecuteCommand.cs
@@ -1,18 +1,24 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using JetBrains.Annotations;
-using Microsoft.DotNet.Cli.Utils;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.EntityFrameworkCore.Utilities;
-using System.Linq;
 
 namespace Microsoft.EntityFrameworkCore.Tools.Cli
 {
     public class ExecuteCommand
     {
+        private static readonly Assembly ThisAssembly = typeof(ExecuteCommand).GetTypeInfo().Assembly;
+        private static readonly string AssemblyVersion = ThisAssembly
+            .GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion
+            ?? ThisAssembly.GetName().Version.ToString();
+        
+        private const string DispatcherVersionArgumentName = "--dispatcher-version";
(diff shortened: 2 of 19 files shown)


natemcmaster:
cc @NTaylorMullen this also meant removing Microsoft.Extensions.DotnetToolDispatcher.Sources. The implementation is still there, but the value of shared implementation was getting in the way of our longer-term goals for dotnet-ef. We may go back to using it as we continue to redraw the boundaries of "invoker" vs "inside-man"

NTaylorMullen:
Interesting, are you guys going to bake in the tool version verification into the EF tool?

src/Microsoft.EntityFrameworkCore.Tools.Cli/ExecuteCommand.cs
```diff
@@ -1,18 +1,24 @@
...
 namespace Microsoft.EntityFrameworkCore.Tools.Cli
 {
     public class ExecuteCommand
     {
+        private static readonly Assembly ThisAssembly = typeof(ExecuteCommand).GetTypeInfo().Assembly;
+        private static readonly string AssemblyVersion = ThisAssembly
+            .GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion
+            ?? ThisAssembly.GetName().Version.ToString();
+        
+        private const string DispatcherVersionArgumentName = "--dispatcher-version";
```
natemcmaster: @NTaylorMullen we still do tool verification for now. We needed to remove dependencies on Microsoft.DotNet.Cli.Utils and Ms.Ext.PlatformAbstractions

src/Microsoft.EntityFrameworkCore.Tools.Cli/Extensions/CommandLineApplicationExtensions.cs
```diff
@@ -33,6 +33,9 @@ public static CommandOption VerboseOption(this CommandLineApplication command)
         public static CommandOption JsonOption(this CommandLineApplication command)
             => command.Option("--json", "Use json output");
 
+        public static CommandOption JsonDelimitedOption(this CommandLineApplication command)
+            => command.Option("--json-delimited", "Use json output wrapped by '//BEGIN' and '//END'");
+
```
bricelam: Why? Can it wait until we need it?

src/Microsoft.EntityFrameworkCore.Tools.Cli/ConsoleCommandLogger.cs
```diff
@@ -1,32 +1,76 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using JetBrains.Annotations;
-using Microsoft.DotNet.Cli.Utils;
 using Microsoft.EntityFrameworkCore.Design.Internal;
+using Newtonsoft.Json;
```
bricelam: Why depend on JSON.NET? We don't parse JSON or serialize unknown types. It adds no value.

bricelam:
:shipit:

src/Microsoft.EntityFrameworkCore.Tools.Cli/ConsoleCommandLogger.cs
natemcmaster: We already depend on it, but it came transitively from Ms.DotNet.Cli.Utils, but in the spirit of removing dependencies, we could also remove this too.

src/Microsoft.EntityFrameworkCore.Tools.Cli/Extensions/CommandLineApplicationExtensions.cs
natemcmaster: We need it for pmc commands to parse the json output.

bricelam: Ah, since it's no longer the only thing on stdout. Got it. :blush:

bricelam: Does it need to be an option or should we always do it?

src/Microsoft.EntityFrameworkCore.Tools.Cli/ConsoleCommandLogger.cs
bricelam: I think it's worth it--one less piece to the puzzle.

src/Microsoft.EntityFrameworkCore.Tools.Cli/Extensions/CommandLineApplicationExtensions.cs
natemcmaster: I considered it, but figured it better to make these opt-in as comments not in the JSON spec.

bricelam: Sounds good. We can reevaluate this interaction with the new architecture.

src/Microsoft.EntityFrameworkCore.Tools.Cli/ConsoleCommandLogger.cs
natemcmaster: Ok. I'll follow up in a separate PR.

PR #5543, opened 2016-05-27
Fix to #3807 - EF Core SQL statements with no filter are generated for LINQ query using FirstOrDefault()

### LINKED ISSUE #3807: EF 7 SQL statements with no filter are generated for LINQ query using FirstOrDefault()
The following code queries for blogs and returns the first post for the blog. I checked the SQL that is executed using SQL Server Profiler. First a query for the blogs table is executed. That's fine. Then, it appears to execute a separate query for each blog against the posts table. It appears that this is wrong though. The posts queries don't have a where clause. It should be limiting those to just the blog at hand. The end result comes out correct. I'm assuming filtering is happening in memory. However, if the posts table contains lots of rows, it would be unacceptably slow. If it's doing N+1 queries, it should including a WHERE clause and filtering by id in the main table.

```c#
using Microsoft.Data.Entity;
using System.Linq;

namespace ConsoleApplication1
{
    class Program
    {
        {
            using (var bc = new BlogContext())
            {
                bc.Database.EnsureCreated();
                for (int i = 0; i < 10; i++)
                {
                    var b = new Blog { };
                    bc.Blogs.Add(b);
                    var p = new Post { Blog = b };
                    bc.Posts.Add(p);
                }
                bc.SaveChanges();
                bc.Blogs
                    //.Take(100)
                    //.Select(b2 => b2.Posts.Any())
                    .Select(b2 => b2.Posts.FirstOrDefault().Content)
                    .ToArray();
            }
        }
    }

    public class BlogContext : DbContext
    {
        //protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder) => optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=Blogs;Trusted_Connection=True;");
        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder) => optionsBuilder.UseSqlServer(@"Server=localhost;Database=Blogs;Trusted_Connection=True;");
    }

    public class Blog
    {
        public int? Id { get; set; }
        public ICollection<Post> Posts { get; set; }
    }

    public class Post
    {
        public int? Id { get; set; }
        public string Content { get; set; }
        public int? BlogId { get; set; }
    }
}
```
```SQL
SELECT [b2].[Id]
FROM [Blog] AS [b2]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
SELECT [p].[BlogId], [p].[Content]
FROM [Post] AS [p]
go
```

maumar:
Fixes #3807 

Problem was that we were not translating queries that would project first element of a collection of entities, like so:

ctx.Customers.Select(c => c.Orders.FirstOrDefault().Id)

We can translate those queries, as long as single value is projected in the subquery.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -728,6 +728,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
             var subQueryModel = expression.QueryModel;
 
+            var subQueryOutputDataInfo = subQueryModel.GetOutputDataInfo();
             if (subQueryModel.IsIdentityQuery()
                 && subQueryModel.ResultOperators.Count == 1
                 && subQueryModel.ResultOperators.First() is ContainsResultOperator)
@@ -762,10 +763,15 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                     }
                 }
             }
-            else if (!(subQueryModel.GetOutputDataInfo() is StreamedSequenceInfo))
+            else if (!(subQueryOutputDataInfo is StreamedSequenceInfo))
             {
+                var streamedSingleValueInfo = subQueryOutputDataInfo as StreamedSingleValueInfo;
+                var streamedSingleValueSupportedType = streamedSingleValueInfo != null
+                    && _relationalTypeMapper.FindMapping(streamedSingleValueInfo.DataType.UnwrapNullableType().UnwrapEnumType()) != null;
+
                 if (_inProjection
-                    && !(subQueryModel.GetOutputDataInfo() is StreamedScalarValueInfo))
+                    && !(subQueryOutputDataInfo is StreamedScalarValueInfo) 
+                    && !streamedSingleValueSupportedType)
                 {
                     return null;
                 }
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
@@ -113,6 +113,35 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
                     .Select(c => c.Orders.FirstOrDefault()));
         }
 
+
+        [ConditionalFact]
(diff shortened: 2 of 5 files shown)


maumar:
@anpete

smitpatel:
EF 7 ? :trollface:

maumar:
@anpete @smitpatel ping

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -762,8 +762,13 @@ protected override Expression VisitSubQuery(SubQueryExpression expression)
             }
             else if (!(subQueryModel.GetOutputDataInfo() is StreamedSequenceInfo))
             {
+                var streamedSingleValueInfo = subQueryModel.GetOutputDataInfo() as StreamedSingleValueInfo;
+                var streamedSingleValueSuppertedType = streamedSingleValueInfo != null
```
anpete: Typo

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -762,8 +762,13 @@ protected override Expression VisitSubQuery(SubQueryExpression expression)
             }
             else if (!(subQueryModel.GetOutputDataInfo() is StreamedSequenceInfo))
             {
+                var streamedSingleValueInfo = subQueryModel.GetOutputDataInfo() as StreamedSingleValueInfo;
```
anpete: Unify the calls to subQueryModel.GetOutputDataInfo() via a local.

anpete:
:shipit:

PR #5555, opened 2016-05-27
Migrations: Add first-class property facets

### LINKED ISSUE #5223: Migrations: Scaffold MaxLength
When I apply

```
[StringLength(60, MinimumLength = 3)] //nvarchar(MAX) to nvarchar(60)
```

to my existing model, EF doesn't pick up this change.

To repro: 

1. Create a model with string and no DataAnnotations. 
2. add migrations/update
3. apply ``[StringLength(60)0``

EF doesn't detect the change.

#5224 <-- NOT NULL is detected but not enforced

bricelam:
**Note to providers:** This changes some method signatures in `MigrationsSqlGenerator`.

Resolves #5223


ajcvickers:
:shipit:

roji:
Just noticed this change, and am wondering about the inclusion of RowVersion, which seems like a very SqlServer-specific concept, into the provider-independent core...?

ajcvickers:
@roji We are very inconsistent about it at the moment. It used to be a core concept in the old stack, and hence there is a data annotation for it. We realized it wasn't core when doing the fluent API and so didn't add it there, but it was still in the core concepts of the type mapper. Need to discuss with @bricelam @divega @rowanmiller and others about where we should end up with this.

PR #5556, opened 2016-05-27
React to update CoreCLR packages

JunTaoLuo:
This builds locally but just want to make sure using the latest `System.Collections.Immutable` with a floating version is okay.

cc @pranavkm @natemcmaster.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/project.json
+++ b/src/Microsoft.EntityFrameworkCore/project.json
@@ -21,7 +21,7 @@
     "Microsoft.Extensions.DependencyInjection": "1.0.0-*",
     "Microsoft.Extensions.Logging": "1.0.0-*",
     "Remotion.Linq": "2.0.2",
-    "System.Collections.Immutable": "1.1.37"
+    "System.Collections.Immutable": "1.2.0-*"
   },
   "frameworks": {
     "net451": {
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore/project.json
```diff
@@ -21,7 +21,7 @@
     "Microsoft.Extensions.DependencyInjection": "1.0.0-*",
     "Microsoft.Extensions.Logging": "1.0.0-*",
     "Remotion.Linq": "2.0.2",
-    "System.Collections.Immutable": "1.1.37"
+    "System.Collections.Immutable": "1.2.0-*"
```
pranavkm: cc @natemcmaster

natemcmaster:
:sheep: :it: assuming tests also pass

src/Microsoft.EntityFrameworkCore/project.json
divega: I think we decided not to depend on `System.Collections.Immutable`, but maybe they came back? Or maybe they didn't go away because they are immutable 😆

pranavkm: We could ignore this issue in Coherence if you'd like to stick to 1.1.37.

smitpatel: They returned when we decided to store the discovered navigation property infos on entity type builder as annotation inside RelationshipDiscoveryConvention. That is the only place which uses it.

divega: > We could ignore this issue in Coherence if you'd like to stick to 1.1.37. 

No, I wasn't suggesting that.

PR #5557, opened 2016-05-27
Fix OSX build on Travis.

cesarbs:
@natemcmaster


natemcmaster:
Not necessary. We don't run this build on OSX, only Linux.

As a separate issue, we could re-evaluate building on travis. We stopped because travis OSX agents were too slow and always timed out.

PR #5558, opened 2016-05-27
Fix #5545 - Prevent null compensation on Coalesce to avoid invalid cast

### LINKED ISSUE #5545: Query :: prevent Coalesce from null compensation in NavigationRewritingExpressionVisitor -> VisitBinary
If we are expanding optional navigation, we sometimes change the expression type from non-nullable to nullable (e.g. c.Order.Id, Id is int, but Order can be null, so we cast the whole expression into int?).

In the VisitBinary we compensate for the discrepancy that can sometimes happen, e.g. during comparisons, when one side was expanded and converted to a nullable type and the other side was not. We do that by converting non-nullable types to nullable if there is discrepancy. 

However we should not do that for coalesce as it is sometimes be used to force non-nullable result of an expression.

mikary:
@maumar @anpete


maumar:
:shipit:

PR #5561, opened 2016-05-27
SQL Server Migrations: Rebuild indexes on column narrowing

### LINKED ISSUE #4642: Migrations: Index wasn't dropped when altering its column
Hello.
I think there is a small problem with migrations and unique indexes. 
I created simple schema like this:
`User --> Role <----> Permission`

In my initial migration I created these entities and I added unique index on UserName column:

``` C#
modelBuilder.Entity<User>()
                .HasIndex(u => u.UserName)
                .IsUnique();
```

Then I added [Required] adnotation to UserName column and made second migration.
When I ran `dnx ef database update` I get this error.

> The index 'IX_User_UserName' is dependent on column 'UserName'.
> The index 'IX_User_UserName' is dependent on column 'UserName'.
> ALTER TABLE ALTER COLUMN UserName failed because one or more objects access this column.

The solution was simple. I had to add manually drop/create index to migration.

It is intended effect?
Does indexes shouldn't be dropped and created automatically as foreign keys ?

bricelam:
Fixes #4642


src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -55,6 +56,30 @@ protected override void Generate(MigrationOperation operation, IModel model, Mig
             Check.NotNull(operation, nameof(operation));
             Check.NotNull(builder, nameof(builder));
 
+            var property = FindProperty(model, operation.Schema, operation.Table, operation.Name);
+            var mightNarrowColumn = property != null
+                && (property.IsKeyOrForeignKey()
+                    || operation.ColumnType != null
+                    || operation.IsUnicode == false
+                    || operation.MaxLength.HasValue
+                    || !operation.IsNullable);
```
bricelam: Note this is only a heuristic. Migrations SQL generators have no way to diff against the previous model...

ajcvickers:
:shipit:

bricelam:
I'm not happy with this implementation. It means that if you do some things that might require rebuilding the indexes, they'll be rebuilt on all subsequent alters until the original thing you did is undone.

The alternative implementation is to do something in the model differ like rebuild them for every provider or annotate the alter operation with the relevant information.

We could also look at flowing in the previous model, but this would require significant work.

PR #5563, opened 2016-05-27
Fix bad MigrationSqlGenerator tests

bricelam:


smitpatel:
Because `rowversion` is sql server specific?
:shipit:

bricelam:
> Because rowversion is sql server specific?

No, because the mock type mapper can't handle `byte[]` :smile:

PR #5564, opened 2016-05-27
Remove dependency on Newtonsoft.Json from dotnet-ef

natemcmaster:
cc @bricelam


bricelam:
:shipit: Although still I prefer the way they were originally implemented using string concatenation; it kept it very explicit about what we serialized (and didn't).

PR #5565, opened 2016-05-27
Issue warning when EF6 commands are also installed in solution

### LINKED ISSUE #5145: Migrations PowerShell commands can't work for both EF6 and EF Core projects that are part of the same solution

natemcmaster:
cc @rowanmiller @bricelam 

When executing any command that also exist in EF6 powershell commands AND the EF6 commands are installed, users will see a warning.

image

Fix #5145

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.PowerShell2.psm1
+++ b/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.PowerShell2.psm1
@@ -44,6 +44,6 @@ function Enable-Migrations {
 
 function Hint-Upgrade ($name) {
     if (Get-Module | Where { $_.Name -eq 'EntityFramework' }) {
-        Write-Warning "Executing the Entity Framework Core version of '$name'. Run 'EntityFramework\$name' to execute for EF 6 and earlier."
+        Write-Warning "Both Entity Framework Core and Entity Framework 6.x commands are installed. The Entity Framework Core version is executing. You can fully qualify the command to select which one to execute, 'EntityFramework\$name' for EF6.x and 'EntityFrameworkCore\$name' for EF Core."
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.PowerShell2.psm1
```diff
@@ -1,31 +1,49 @@
...
+    Hint-Upgrade $MyInvocation.MyCommand
+    Write-Warning 'Enable-Migrations is obsolete. Use Add-Migration to start using Migrations.'
+}
+
+#
+# Private functions
+#
+
+function Hint-Upgrade ($name) {
+    if (Get-Module | ? Name -eq EntityFramework) {
```
bricelam: The `-eq` prameter on `where` is a ps 3 thing. Use a script block

natemcmaster:
:up ::date: fixed the syntax for PS v2

bricelam:
:shipit:

rowanmiller:
No space in "EF 6" 😄

rowanmiller:
Suggesting a more verbose message...

> Both Entity Framework Core and Entity Framework 6.x commands are installed. The Entity Framework Core version is executing. You can fully qualify the command to select which one to execute, 'EntityFramework\Add-Migration' for EF6.x and 'EntityFrameworkCore\Add-Migration' for EF Core.

natemcmaster:
:+1: will update with feedback

PR #5587, opened 2016-05-31
Fix build break introduced PR #5555

smitpatel:
Issues:
1. Missing Table property values in `AddColumnOperation`
2. No `StringTypeMapper` in test code.
3. Use property to get type when property is not null and maxlength is absent or same as max length annotation.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTestBase.cs
+++ b/test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTestBase.cs
@@ -150,6 +150,7 @@ namespace Microsoft.EntityFrameworkCore.Relational.Tests.Migrations
                     Table = "Person",
                     Name = "Name",
                     ClrType = typeof(string),
+                    MaxLength = 30,
                     IsNullable = true
                 });
 
@@ -362,18 +363,21 @@ namespace Microsoft.EntityFrameworkCore.Relational.Tests.Migrations
                         new AddColumnOperation
                         {
                             Name = "Id",
+                            Table = "People",
                             ClrType = typeof(int),
                             IsNullable = false
                         },
                         new AddColumnOperation
                         {
                             Name = "EmployerId",
+                            Table = "People",
                             ClrType = typeof(int),
                             IsNullable = true
                         },
                         new AddColumnOperation
                         {
                             Name = "SSN",
+                            Table = "People",
                             ClrType = typeof(string),
                             ColumnType = "char(11)",
                             IsNullable = true
--- a/test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTest.cs
@@ -79,15 +79,6 @@ namespace Microsoft.EntityFrameworkCore.Relational.Tests.Migrations
                 Sql);
         }
 
-        public override void AddColumnOperation_with_maxLength_overridden()
-        {
-            base.AddColumnOperation_with_maxLength_overridden();
-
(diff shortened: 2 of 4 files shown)


smitpatel:
@natemcmaster

test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTestBase.cs
```diff
@@ -106,7 +106,7 @@ public virtual void AddColumnOperation_with_maxLength()
                     IsNullable = true
                 });
 
-        [Fact]
+        //[Fact] TODO: See issue #5586
```
natemcmaster: If this change is going to introduce other regressions, we should investigate more thoroughly the find out what is wrong or revert #5555

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -650,8 +650,9 @@ public class MigrationsSqlGenerator : IMigrationsSqlGenerator
             if (property != null)
             {
                 // TODO: Allow unicode to be overridden with #3420
-                if (maxLength == property.GetMaxLength()
-                    && rowVersion == (property.IsConcurrencyToken && (property.ValueGenerated == ValueGenerated.OnAddOrUpdate)))
+                if ((maxLength == null
+                     || maxLength == property.GetMaxLength())
```
natemcmaster: `GetMaxLength()` returns `int?`, so shouldn't this extra check be unnecessary? Perhaps the bug is actually in ` property.GetMaxLength()`

test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTestBase.cs
smitpatel: It is not regression. It is limitation of our test base. For this test to work the typemapper needs to have `StringTypeMapper` property populated which is null in `RelationalTypeMapper`. All provider specific type mappers populate it in constructor so this test would work fine in providers but not in relational since for relational tests we use `ConcreteTypeMapper` which is just fake implementation. I will disable this test specifically for relational and let it run for SqlServer & SQLite. Does that look good?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
bricelam: +1

test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTestBase.cs
bricelam: Can we move the test to the SqlServer ones then? I had to do this with a few others because the relation mocks aren't as complete.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
smitpatel: So the check needed is either maxlenth is null then use property whatever is the value of max length there or if maxlengh is not null but same as property then use property. Any other way to describe this logic?

test/Microsoft.EntityFrameworkCore.Relational.Tests/Migrations/MigrationSqlGeneratorTestBase.cs
smitpatel: Moved to provider specific and removed from Relational.

smitpatel:
Updated.

natemcmaster:
:shipit:

Issue #5499, opened 2016-05-24
Local variable null check incorrectly translates to SQL


jacob-robertson:
I have a simple filter being passed into a `.Where()` call that is improperly using the equal operator (=) instead of `IS` to check for null values.

    var g = Guid.Parse('XXXX');
    var guitars = context.Guitars
                         .Include(g => g.Model)
                         .Where(g => guitarId == null || g.Id == guitarId);

Where `guitarId` is equal to some non-null value, translates to

    WHERE (@__guitarId_0 = NULL) OR ([g].[Id] = @__guitarId_1)

When `guitarId` is `null`, I get

    WHERE (@__guitarId_0 = NULL) OR [g].[Id] IS NULL

The first is incorrect but returns data.  The second always returns an empty data-set.  I would expect something like the following

    WHERE (@__guitarId_0 IS NULL) OR [g].[Id] IS NULL

The obvious workaround is to conditionally add the filter expression, but this does seem to be an issue with SQL translation in general.  I'm also curious as to why, in the first example the two references to `guitarId` are given two separate SQL parameters.  Shouldn't there just be one?  I've attached a working example.
                    
[EFTester.zip](https://github.com/aspnet/EntityFramework/files/280198/EFTester.zip)

c1rus:
This works before rc2 :/

gdoron:
Which build can we use if we can't wait for RTM on this?
I'm quite surprised that not even one test failed because of this...

maumar:
@gdoron the fix is not checked in yet

maumar:
fixed in 3e797a8ae3d09aeb6e580596dfc6f2c46c3b898a

maumar:
@gdoron just checked in the fix so should be available in tomorrow's nightly build, unless we get some infra issues with the build system

prasannapattam:
Where can I get the nightly build for this fix?

gdoron:
Add the MyGet feed

smitpatel:
More info on feeds : https://github.com/aspnet/Home/wiki/NuGet-feeds

PR #5588, opened 2016-05-31
[SMALL] Fix to #5499 - Local variable null check incorrectly translates to SQL

maumar:
Problem was that during null semantics rewrite we were only looking at constant and column expressions, never taking parameter expressions into account. Those can be optimized out in some cases as we know the value of the parameter during sql generation. Added another pass of PredicateReductionExpressionOptimizer after null semantics is applied to get rid of redundant true/false constants that may now be produced.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -169,8 +169,8 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
 
             if (selectExpression.Predicate != null)
             {
-
-                var constantExpression = selectExpression.Predicate as ConstantExpression;
+                var nullSemanticsPredicate = ApplyNullSemantics(selectExpression.Predicate);
+                var constantExpression = nullSemanticsPredicate as ConstantExpression;
 
                 if (constantExpression == null
                     || !(bool)constantExpression.Value)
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1238,19 +1239,45 @@ var parameter
...
+                        if (constantExpression != null)
                         {
-                            return
-                                expression.NodeType == ExpressionType.Equal
-                                    ? (Expression)new IsNullExpression(columnExpression)
-                                    : Expression.Not(new IsNullExpression(columnExpression));
+                            if (parameterValue == null && constantExpression.Value == null)
+                            {
+                                return
+                                    expression.NodeType == ExpressionType.Equal
```
maumar: i was considering return Expression.Constant(expression.NodeType == ExpressionType.Equal), but readability was somewhat worse. Thoughts?

maumar:
new version up

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
smitpatel: Either is fine or you can use a local var to store the value for readability too.

smitpatel:
:shipit:

PR #5591, opened 2016-05-31
Description and tags for shipping packages

### LINKED ISSUE #5463: Clean up package metadata and pre-Core references
# NuGet package metadata
The `project.json` file of every shipping NuGet package has two properties that must be non-empty:
1. description
2. tags

## Description guidance

Description should be at least something like this:
  "description": "ASP.NET Core MVC view rendering features. Contains common types used in most MVC applications as well as view rendering features such as view engines, views, view components, and HTML helpers.",

And if there are types in the package that are frequently referenced *in source code*, add this:
  "description": "%description%\r\nCommonly used types:\r\nNS1.Type1\r\nNS1.Type2\r\nNS2.Type1\r\nNS3.Type1",
Note: Sort the types by the full namespace+type name

## Tags guidance

The tags should generally include a "product" tag indicating the main product this is a part of (e.g. `aspnetcore`), and also "subject" or "feature" tags to associate this package with other related packages (e.g. `aspnetcoremvc`, `logging`, `json`).

Examples of tags:

```json
  "tags": [
    "aspnetcore"
  ],
```

```json
  "tags": [
    "aspnetcore",
    "aspnetcoremvc",
    "XYZ"
  ],
```

Note: Certain packages are not part of ASP.NET Core *or* Entity Framework Core, so be careful not to add those tags to the metadata of these packages.

# Other stuff to clean up related to the "Core" product rename

Need to do a search in all files in the repo for strings such as `asp.net`, `aspnet`, `entity framework`, and `entityframework`. This obviously has a lot of false positives because these strings appear in namespace names all over the place. However, it's not too hard to weed through those (despite thousands of hits) because when scrolling in the output of the commands it's easy to see breaks in the pattern that need to be investigated.

Some useful commands to run:

```
findstr /snip /c:"entityframework" *
findstr /snip /c:"entity framework" *
findstr /snip /c:"asp.net" *
findstr /snip /c:"aspnet" *
```

# Update readme.md in the root of the repo

This is a fairly simple task of ensuring the content in the readme.md file also doesn't have references to pre-Core naming.

rowanmiller:
Resolves #5463

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.InMemory/project.json
+++ b/src/Microsoft.EntityFrameworkCore.InMemory/project.json
@@ -1,6 +1,16 @@
 {
   "version": "1.0.0-rc3-*",
-  "description": "In memory data store for Entity Framework (to be used for testing purposes).",
+  "description": "In memory database provider for Entity Framework Core (to be used for testing purposes).",
+  "packOptions": {
+    "tags": [
+      "Entity Framework",
+      "Entity Framework Core",
+      "EF",
+      "Data",
+      "O/RM",
+      "In Memory"
+    ]
+  },
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
@@ -16,7 +26,7 @@
     "Microsoft.EntityFrameworkCore": "1.0.0-*"
   },
   "frameworks": {
-    "net451": {},
+    "net451": { },
     "netstandard1.3": {
       "imports": [
         "portable-net452+win81"
--- a/src/Microsoft.EntityFrameworkCore.Tools/project.json
+++ b/src/Microsoft.EntityFrameworkCore.Tools/project.json
@@ -1,6 +1,15 @@
 {
   "version": "1.0.0-preview2-*",
-  "description": "Command line tools for Entity Framework.",
+  "description": "Command line tools for Entity Framework Core.",
+  "packOptions": {
+    "tags": [
+      "Entity Framework",
+      "Entity Framework Core",
+      "EF",
+      "Data",
+      "O/RM"
+    ]
+  },
   "buildOptions": {
     "emitEntryPoint": true,
     "outputName": "dotnet-ef",
--- a/src/Microsoft.EntityFrameworkCore/project.json
+++ b/src/Microsoft.EntityFrameworkCore/project.json
@@ -1,6 +1,15 @@
 {
   "version": "1.0.0-rc3-*",
-  "description": "Entity Framework is Microsoft's recommended data access technology for new applications.",
+  "description": "Entity Framework Core is a lightweight and extensible version of the popular Entity Framework data access technology.",
+  "packOptions": {
+    "tags": [
+      "Entity Framework",
+      "Entity Framework Core",
+      "EF",
+      "Data",
+      "O/RM"
+    ]
+  },
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
--- a/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/project.json
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/project.json
@@ -1,6 +1,15 @@
 {
   "version": "1.0.0-rc3-*",
-  "description": "Reusable Microsoft.EntityFrameworkCore.Relational.Design functional tests for provider writers.",
+  "description": "Shared design-time test suite for Entity Framework Core relational database providers.",
+  "packOptions": {
+    "tags": [
+      "Entity Framework",
+      "Entity Framework Core",
+      "EF",
+      "Data",
+      "O/RM"
+    ]
+  },
   "buildOptions": {
     "keyFile": "../../tools/Key.snk",
     "nowarn": [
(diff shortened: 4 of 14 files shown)


ajcvickers:
:shipit:

src/Microsoft.EntityFrameworkCore.InMemory/project.json
```diff
@@ -1,6 +1,16 @@
 {
   "version": "1.0.0-rc3-*",
-  "description": "In memory data store for Entity Framework (to be used for testing purposes).",
+  "description": "In memory database provider for Entity Framework Core (to be used for testing purposes).",
+  "packOptions": {
+    "tags": [
+      "Entity Framework",
+      "Entity Framework Core",
```
Eilon: Sorry I'm late to the party here. So far we've been using tags without spaces, e.g. `entityframeworkcore`, `aspnetcore`, etc. And these are both new tags compared to EF 6.x and earlier. Do you feel strongly either way about this?

For example, here's what I put in for the EF Diagnostics page: https://github.com/aspnet/Diagnostics/blob/dev/src/Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore/project.json#L10-L12

src/Microsoft.EntityFrameworkCore.Tools/project.json
```diff
@@ -1,6 +1,15 @@
 {
   "version": "1.0.0-preview2-*",
-  "description": "Command line tools for Entity Framework.",
+  "description": "Command line tools for Entity Framework Core.",
```
Eilon: In all the tooling packages so far we also put in some hint about the command name, e.g. `This package includes the dotnet-ef command line tool.` This is part of the "cost" of having package names not match assembly names 😦

src/Microsoft.EntityFrameworkCore.InMemory/project.json
```diff
@@ -1,6 +1,16 @@
 {
   "version": "1.0.0-rc3-*",
-  "description": "In memory data store for Entity Framework (to be used for testing purposes).",
+  "description": "In memory database provider for Entity Framework Core (to be used for testing purposes).",
```
Eilon: Would `In-memory` be clearer?

src/Microsoft.EntityFrameworkCore.InMemory/project.json
rowanmiller: I'm not sure "entityframeworkcore" without spaces really helps, it looks kind of like a TFM based on our product name 😄. If I was searching for EF packages, I would use "Entity Framewor" or "EF" and not "entityframework".

src/Microsoft.EntityFrameworkCore.InMemory/project.json
rowanmiller: Or we could do InMemory since that's what all the APIs related to it use.

src/Microsoft.EntityFrameworkCore.Tools/project.json
rowanmiller: Will do, it also "contains" the Package Manager Console commands, so I will mention those too. BTW do we still put the hyphen in `dotnet-ef`... it seems like `dotnet ef` is more helpful since that's what you actually type.

Eilon: I'm fine either way 😄

src/Microsoft.EntityFrameworkCore.InMemory/project.json
Eilon: I'd use InMemory if it's the full type name, e.g. InMemoryWhateverProvider, but "in-memory" if it's just plain English (which seems to be the case here).

src/Microsoft.EntityFrameworkCore.InMemory/project.json
Eilon: I personally prefer space-less tags because that's the norm (or even requirement) in many systems, so it's nice to have the same tags across different systems (e.g. StackOverflow). I think having extra tags such as "entity framework" ought to be just fine as well. No harm having several tags, assuming they actually carry some meaning.

src/Microsoft.EntityFrameworkCore/project.json
```diff
@@ -1,6 +1,15 @@
 {
   "version": "1.0.0-rc3-*",
-  "description": "Entity Framework is Microsoft's recommended data access technology for new applications.",
+  "description": "Entity Framework Core is a lightweight and extensible version of the popular Entity Framework data access technology.",
```
bricelam: @divega / @Eilon Can and should we update the repository description to this too?

rowanmiller: Good point, I can do that... and I have 😄

src/Microsoft.EntityFrameworkCore.InMemory/project.json
rowanmiller: Just to throw one more convention into the mix 😄... on StackOverflow we use `entity-framework`, so how about we have:
* `Entity Framework`
* `Entity Framework Core`
* `entity-framework`
* `entity-framework-core`
* `EF`

I'm not convinced the spaceless one makes much sense... but happy to add that into the mix too if you want it...

rowanmiller: Or perhaps we don't even need the versions without "Core" in them?

Eilon: @rowanmiller sounds good to me (and I'm fine dropping the Core-less ones).

rowanmiller:
Opened new PR (https://github.com/aspnet/EntityFramework/pull/5611) with further changes discussed in this PR.

PR #5592, opened 2016-05-31
One-to-many relationships are not ambiguous

### LINKED ISSUE #5406: Duplicate foreign key generated after RC1 -> RC2 migration
Two simple models:

    public abstract class TicketBatch
    {
        [Key, DatabaseGenerated( DatabaseGeneratedOption.Identity )]
        public int Id { get; set; }
        public List<Ticket> Tickets { get; set; }
    }

    public class Ticket
    {
        [Key, DatabaseGenerated( DatabaseGeneratedOption.None )]
        public Guid Guid { get; set; }

        public int TicketBatchId { get; set; }

        public TicketBatch TicketBatch { get; set; }
    }

TicketBatch has two descendants that define only value (non-relationship) properties.

After manual migration of the project to 1.0.0-rc2-final, EF Core now wants to apply following in new migration:

                name: "TicketBatchId1",
                table: "Tickets",
                nullable: true);

    migrationBuilder.CreateIndex(
                name: "IX_Tickets_TicketBatchId1",
                table: "Tickets",
                column: "TicketBatchId1");

When I explicitly state that the foreign key is TicketBatchId in ApplicationDbContext.OnModelCreating:

    builder.Entity<Ticket>()
        .HasOne( t => t.TicketBatch )
        .WithMany( b => b.Tickets )
        .HasForeignKey( t => t.TicketBatchId );

then new column is not added. But that should not be necessary as conventions are followed.

Only similar issue I found is #5344 but I did not dig deeper.

Thanks!

smitpatel:
Fix #5406 
Set principal end correctly for explicit
For relationships which has collection navigation, don't check for matching property on principal side.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Builders/EntityTypeBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Builders/EntityTypeBuilder.cs
@@ -366,6 +366,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Builders
             => relatedEntityType.Builder
                 .Relationship(Builder, ConfigurationSource.Explicit)
                 .IsUnique(false, ConfigurationSource.Explicit)
+                .RelatedEntityTypes(Builder.Metadata, relatedEntityType, ConfigurationSource.Explicit)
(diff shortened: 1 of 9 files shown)


AndriySvyryd:
:shipit:

Issue #5573, opened 2016-05-29
Metadata: While removing key we may add FK pointing to the key again


smitpatel:
From https://github.com/aspnet/EntityFramework/issues/5449#issuecomment-222269974
> The issue is when you set  IsRequired(false)  on PK, we remove key defined on that property. Removing key removes referencing FKs but while removing FKs we run conventions which could add another FK referencing same PK again. Which blocks us from removing the key and exception is thrown.
 Fix would be to run the conventions on removed fks are after key is removed. So FKs added by convention would not target the key being removed.

```
Exception message: System.InvalidOperationException: Cannot remove key {'Id'} from entity type 'ModelApp.TicketBatch' because it is referenced by a foreign key in entity type 'ModelApp.Ticket'. 
All foreign keys must be removed or redefined before the referenced key can be removed.
Stack trace:
at Microsoft.EntityFrameworkCore.Metadata.Internal.EntityType.CheckKeyNotInUse(Key key)
at Microsoft.EntityFrameworkCore.Metadata.Internal.EntityType.RemoveKey(Key key, Boolean runConventions)
at Microsoft.EntityFrameworkCore.Metadata.Internal.InternalEntityTypeBuilder.RemoveKey(Key key, ConfigurationSource configurationSource, Boolean runConventions)
at Microsoft.EntityFrameworkCore.Metadata.Internal.InternalPropertyBuilder.IsRequired(Boolean isRequired, ConfigurationSource configurationSource)
at Microsoft.EntityFrameworkCore.Metadata.Builders.PropertyBuilder`1.IsRequired(Boolean required)
at ModelApp.MyContext.OnModelCreating(ModelBuilder modelBuilder)
at Microsoft.EntityFrameworkCore.Infrastructure.ModelSource.CreateModel(DbContext context, IConventionSetBuilder conventionSetBuilder, IModelValidator validator)
at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
at Microsoft.EntityFrameworkCore.Internal.DbContextServices.CreateModel()
at Microsoft.EntityFrameworkCore.Internal.LazyRef`1.get_Value()
at Microsoft.Extensions.DependencyInjection.ServiceProvider.ScopedCallSite.Invoke(ServiceProvider provider)
at Microsoft.Extensions.DependencyInjection.ServiceProviderExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
[...]

PR #5597, opened 2016-05-31
Metadata: RemoveKey should run all conventions after key is removed

smitpatel:
Fixes issue #5573 
Run conventions on removed FKs after Key is removed fully


test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/OneToManyTestBase.cs
```diff
@@ -1975,6 +1975,18 @@ public virtual void Shadow_property_created_for_foreign_key_is_nullable()
                 Assert.Equal(typeof(int?), modelBuilder.Model.FindEntityType(typeof(Order)).FindProperty("MyShadowFk").ClrType);
             }
 
+            [Fact]
+            public virtual void RemoveKey_does_not_add_back_foreign_key_pointing_to_the_same_key()
+            {
+                var modelBuilder = CreateModelBuilder();
+                modelBuilder.Entity<Alpha>(b =>
+                    {
+                        // Following line should not throw exception
```
AndriySvyryd: This comment is obvious 😄 

You could also assert that a key exists on Id before this and it doesn't after.

AndriySvyryd:
:shipit:

PR #5598, opened 2016-05-31
Base default constraint names on column name instead of property name.

### LINKED ISSUE #5475: Relational Metadata: Default constraint names reference properties not columns
For an entity configured with the following...
```C#
modelBuilder.Entity<Employee>().Property(e => ManagerId).HasColumnName("Manager");
```
...default constraint names currently look something like the following.

    AK_Employee_ManagerId
    IX_Employee_ManagerId
    FK_Employee_Employee_ManagerId

In EF6, they looked like this.

    AK_Employee_Manager
    IX_Employee_Manager
    FK_Employee_Employee_Manager

AndriySvyryd:
Fixes #5475


smitpatel:
:shipit:

natemcmaster:
Need to also update the reverse engineering tests.

PR #5600, opened 2016-05-31
[SMALL] Don't assume failed GetService means no provider

### LINKED ISSUE #5500: Improve Error Message: Resolving unregistered service from internal service provider
Currently we assume that resolving an unregistered service means you haven't configured a provider:

> No database provider has been configured for this DbContext. A provider can be configured by overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider. If AddDbContext is used, then also ensure that your DbContext type accepts a DbContextOptions object in its constructor and passes it to the base constructor for DbContext.

But it is particularly confusing if you try to resolve a context from the internal service provider (not a super weird thing to assume might work):

```c#
public static void MyExtensionMethod<TEntity>(this DbSet<TEntity> set)
{
    var context = set.GetService<DbContext>();
}
```

We should update the message so say something like "Could not resolve '<service>', this may be because...".

rowanmiller:
If you call an extension method such as
`DbContext.Database.GetDbConnection()` and we fail to resolve the
connection from DI, then we assumed this was because you did not
register a provider. This is often the case, but it's a misleading error
message if you are just trying to resolve some service that is not
present. Adding some pre-amble to this exception to explain this.

Note we still throw the existing message if we are trying to locate the
configured provider and there is none
(see `DatabaseProviderSelector.SelectServices()`) since we are sure that
is the problem in that case.

Fixes #5500

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Infrastructure/AccessorExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/Infrastructure/AccessorExtensions.cs
@@ -48,7 +48,7 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
             var service = accessor.Instance.GetService<TService>();
             if (service == null)
             {
-                throw new InvalidOperationException(CoreStrings.NoProviderConfigured);
+                throw new InvalidOperationException(CoreStrings.NoProviderConfiguredFailedToResolveService(typeof(TService).FullName));
             }
 
             return service;
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
@@ -579,4 +579,7 @@
   <data name="WarningLogTemplate" xml:space="preserve">
     <value>{message} This warning can be configured using DbContextOptionsBuilder.ConfigureWarnings with event id '{eventId}'.</value>
   </data>
+  <data name="NoProviderConfiguredFailedToResolveService" xml:space="preserve">
+    <value>Unable to resolve service for type '{service}'. This is often because no database provider has been configured for this DbContext. A provider can be configured by overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider. If AddDbContext is used, then also ensure that your DbContext type accepts a DbContextOptions&lt;TContext&gt; object in its constructor and passes it to the base constructor for DbContext.</value>
+  </data>
 </root>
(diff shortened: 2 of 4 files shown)


src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -579,4 +579,7 @@
   <data name="WarningLogTemplate" xml:space="preserve">
     <value>{message} This warning can be configured using DbContextOptionsBuilder.ConfigureWarnings with event id '{eventId}'.</value>
   </data>
+  <data name="NoProviderConfiguredFailedToResolveService" xml:space="preserve">
+    <value>No service of type '{service}' was found. This is often because no database provider has been configured for this DbContext. A provider can be configured by overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider. If AddDbContext is used, then also ensure that your DbContext type accepts a DbContextOptions&lt;TContext&gt; object in its constructor and passes it to the base constructor for DbContext.</value>
```
anpete: Consider the wording for this that DI uses: "Unable to resolve service for type '{0}'"

src/Microsoft.EntityFrameworkCore/Infrastructure/AccessorExtensions.cs
```diff
@@ -48,7 +48,7 @@ public static TService GetService<TService>([NotNull] this IInfrastructure<IServ
             var service = accessor.Instance.GetService<TService>();
             if (service == null)
             {
-                throw new InvalidOperationException(CoreStrings.NoProviderConfigured);
+                throw new InvalidOperationException(CoreStrings.NoProviderConfiguredFailedToResolveService(typeof(TService).Name));
```
anpete: Should this be fully qualified type name

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
rowanmiller: Yep that sounds good, will do.

src/Microsoft.EntityFrameworkCore/Infrastructure/AccessorExtensions.cs
rowanmiller: Will do. I know we've gone back and forwards on this in general... but given it's likely not a user type, I think it makes sense to fully qualify here anyway.

rowanmiller:

PR #5601, opened 2016-05-31
[SMALL] Migrations: mention assembly when no contexts found

### LINKED ISSUE #5276: Migrations: Improve error message when no context in selected project in Package Manager Console
This is usually going to be because you have the wrong project selected. We should print out what project/assembly we looked in and that you can change it with the 'Default project:' dropdown.

```
PM> Add-Migration First
No DbContext was found. Ensure that you're using the correct assembly and that the type is neither abstract nor generic.
```

rowanmiller:
This is usually going to be because you have the wrong project selected
in Package Manager Console (or because you are in the wrong directory
for CLI). Writing out the info helps folks work out they have the wrong
assembly.

Fixes #5276


rowanmiller:
@natemcmaster could you take a quick look at this one

natemcmaster:
:shipit: assuming no tests need updating. I didn't find any tests that execute this code path :-(

rowanmiller:
Correct, that string is only used in the one place in our code base. I'll run a full `build.cmd` before pushing though.

rowanmiller:

Issue #4147, opened 2015-12-21
Misleading Error Message from Microsoft.Data.Entity.Storage.RelationalConnection.CheckForAmbientTransactions()


michaelpaulus:
When using a transaction scope around EF7, I get an exception:

> System.InvalidOperationException: An ambient transaction has been detected. Connections opened by Entity Framework will not be enlisted in ambient transactions. To suppress this warning call SuppressAmbientTransactionWarning() when overriding DbContext.OnConfiguring.

I see the discussion around this on issue:
https://github.com/aspnet/EntityFramework/issues/1584

and the check-in code 

https://github.com/aspnet/EntityFramework/commit/9195c1a013fedbbe6b0822ddcd752201b2bd0f2d

That added:
```
using (new System.Transactions.TransactionScope(System.Transactions.TransactionScopeOption.Suppress)) 
```
around the opening of a connection.

The following check-in that modified the RelationalConnection class https://github.com/aspnet/EntityFramework/commit/fef4009e557c4f5444cf2a2a7026140967d6dd4a removed this check.

In my testing I am able to use a transaction scope when I call SuppressAmbientTransactionWarning() and the connection created does enlist in the ambient transaction.  This is the behavior I want, but is opposite of what the error message said.

Is there something I'm missing?  Is it safe to SuppressAmbientTransactionWarning() and use the transaction scope, safe meaning you are not going to change the behavior to prevent me from using the transaction scope?

Thanks.

rowanmiller:
@divega can you take a look at this one

divega:
I think @michaelpaulus is right that the exception message is misleading: It is not that database connection won't be auto-enlisted in ambient transactions on `Open()` when using EF7 (this is actually not up to us and it will continue to happen on desktop .NET where System.Transactions exists and providers, e.g. SqlClient are designed to auto-enlist). It is rather than EF7 is not aware of System.Tranactions and hence lacks the necessary logic to deal with those ambient transactions, e.g. it won't know not to start a local database transaction during `SaveChanges()` if there is an active ambient Transaction.

rowanmiller:
@divega to see what happens with us creating the nested transaction. If it works as expected then maybe we can just remove this exception.

michaelpaulus:
Instead of checking System.Transactions.Transaction.Current and throwing an exception, couldn't you save that the transaction exists in a local field and use that so you don't create your own transaction?

divega:
[...]

PR #5602, opened 2016-05-31
Fix #5530 - Tweaks to warnings configuration.

### LINKED ISSUE #5530: Performance of Include() on RC2
Having the below model:

```
public partial class Products
 {
        public Products()
        {
        }

        public int Id { get; set; }
        ...
        public int? DocId { get; set; }
        public virtual Documents Doc { get; set; }
}

public partial class Documents
{
        public Documents()
        {
            Products = new HashSet<Products>();
        }

        public int Id { get; set; }
        ...
        public virtual ICollection<Products> Products { get; set; }
}
```

The SQL generated by the following queries is similar and both have similar timings in SSMS (< 0ms) and execution plans:

```
var query = from p in dbContext.Products
     where p.ProductId == req.ProductId
               && p.SiteId == req.Engine.Context.SiteId                             
     select p;

query = query.Include(b => b.Doc);
var x = query.ToList();


var query = from p in dbContext.Products
     join d in dbContext.Documents on p.DocId equals  d.Id
     where p.ProductId == req.ProductId
                && p.SiteId == req.Engine.Context.SiteId
     select new { p, d };

var x = query.ToList();
```

However, the when running the code the query with the Include() is very slow compared to the one with the join. 

With the Include() we are only able to run 54,500 queries in 1 minute
With the join we get 212,500 queries in 1 minute

anpete:
- Throw by default for RelationalEventId.AmbientTransactionWarning and InMemoryEventId.TransactionIgnoredWarning


src/Microsoft.EntityFrameworkCore.InMemory/Properties/InMemoryStrings.resx
```diff
@@ -121,7 +121,7 @@
     <value>Saved {count} entities to in-memory store.</value>
   </data>
   <data name="TransactionsNotSupported" xml:space="preserve">
-    <value>Transactions are not supported by the in-memory store. To suppress this Exception call IgnoreTransactions() when overriding DbContext.OnConfiguring.</value>
+    <value>Transactions are not supported by the in-memory store. To suppress this Exception use the ConfigureWarnings API when overriding DbContext.OnConfiguring.</value>
```
divega: Is it ok to say `when overriding DbContext.OnConfiguring` although many ASP.NET customers will be doing `AddDbContext()` or should we say something more generic like `when configuring the DbContextOptions`?

cc @rowanmiller

src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
```diff
@@ -193,7 +193,7 @@
     <value>Unable to set property '{property}' as a discriminator for entity type '{entityType}' because it is not a property of '{entityType}'.</value>
   </data>
   <data name="AmbientTransaction" xml:space="preserve">
-    <value>An ambient transaction has been detected. Connections opened by Entity Framework will not be enlisted in ambient transactions. To suppress this warning call SuppressAmbientTransactionWarning() when overriding DbContext.OnConfiguring.</value>
+    <value>An ambient transaction has been detected. Entity Framework Core does not support ambient transactions. To suppress this warning call ConfigureWarnings when overriding DbContext.OnConfiguring.</value>
```
divega: :+1: for first sentence. It fixes #4147. 

For second sentence
- I believe it should say `To suppress this exception use the ConfigureWarnings API` like in the previous one, assuming we switched it back to throw by default. 
- Same concern as before that many customers will call this in `AddDbContext()` rather than overriding `OnConfiguring()`.

Still not sure about second sentence.

divega: We might also want to extend this message with a link to a transaction management topic, e.g. that explains `SaveChanges()` uses local transactions automatically and we have the `BeginTransaction()` API, etc. 

cc @rowanmiller

src/Microsoft.EntityFrameworkCore.InMemory/Properties/InMemoryStrings.resx
rowanmiller: In other parts of the stack we have used...

> overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider

so perhaps say...

> Transactions are not supported by the in-memory store. To suppress this Exception use the ConfigureWarnings API. ConfigureWarnings can be used when overriding the DbContext.OnConfiguring method or using AddDbContext on the application service provider.

src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
rowanmiller: Agreed, you can use this link  http://go.microsoft.com/fwlink/?LinkId=800142 (no page at the destination just yet)

AndriySvyryd:
:shipit: for the implementation

PR #5603, opened 2016-06-01
[SMALL] Minor exception message improvements

### LINKED ISSUE #4525: Provide more information in exception when collisions happen with temporary key values
(Capturing @AndriySvyryd's suggestion at https://github.com/aspnet/EntityFramework/pull/4505#issuecomment-181979761)

Temporary key values assigned by EF automatically share the same domain as real keys. By default we use different parts of the domain than what we expect applications to need e.g. for integer keys we use negative numbers, counting backwards. But any time a collision happens between a temporary key value and a key provided by the user (regardless of whether the user intended the value provided to be real or temporary) it would be useful to offer additional information so that the user can understand how temporary keys are being generated and decide how to best change his/her code to avoid the collision.

This is related to #4488 and possibly an alternate fix from the one implemented in https://github.com/aspnet/EntityFramework/pull/4505.

### LINKED ISSUE #5087: Usability: Mention ServiceLifetime in DbContext disposed exception
Users are likely to encounter this exception when they are using a DbContext that has been registered in DI as a Scoped service when they expect it to be Transient. This would be a good opportunity to provide a message describing how to register the DbContext with a different ServiceLifetime.

rowanmiller:
Fixes #5087
Mention that ObjectDisposedException may be due to disposing a context
that came from DI

Fixes #4525
More info about temporary values when a key collision happens

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
@@ -133,7 +133,7 @@
     <value>The expression '{expression}' is not a valid property expression. The expression should represent a property access: 't =&gt; t.MyProperty'.</value>
   </data>
   <data name="IdentityConflict" xml:space="preserve">
-    <value>The instance of entity type '{entityType}' cannot be tracked because another instance of this type with the same key is already being tracked. For new entities consider using an IIdentityGenerator to generate unique key values.</value>
+    <value>The instance of entity type '{entityType}' cannot be tracked because another instance of this type with the same key is already being tracked. When adding new entities, for most key types a unique temporary key value will be created if no key is set (i.e. if the key property is assigned the default value for its type). If you are explicitly setting key values for new entities, ensure they do not collide with existing entities or temporary values generated for other new entities. When attaching existing entities, ensure that only one entity instance with a given key value is attached to the context.</value>
   </data>
   <data name="WrongStateManager" xml:space="preserve">
     <value>Cannot start tracking InternalEntityEntry for entity type '{entityType}' because it was created by a different StateManager instance.</value>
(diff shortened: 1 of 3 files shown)


rowanmiller:
@divega or @ajcvickers could you take a look at this one

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -133,7 +133,7 @@
     <value>The expression '{expression}' is not a valid property expression. The expression should represent a property access: 't =&gt; t.MyProperty'.</value>
   </data>
   <data name="IdentityConflict" xml:space="preserve">
-    <value>The instance of entity type '{entityType}' cannot be tracked because another instance of this type with the same key is already being tracked. For new entities consider using an IIdentityGenerator to generate unique key values.</value>
+    <value>The instance of entity type '{entityType}' cannot be tracked because another instance of this type with the same key is already being tracked. When adding new entities, for most key types a unique temporary key value will be created if no key is set (i.e. if the key property is assigned the CLR default value for its type). If you are explicitly setting key values for new entities, ensure they do not collide with existing entities or temporary values generated for other new entities. When attaching existing entities, ensure that only one entity instance with a given key value is attached to the context.</value>
```
divega: Not sure I would mention CLR. It makes it sound more complicated and I think the sentence is clear if you omit it. Otherwise looks good to me.

divega:
:shipit: unless you want to wait for @ajcvickers

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
rowanmiller: 👍

rowanmiller:

PR #5604, opened 2016-06-01
CLI commands: always wrap JSON output with begin/end markers

natemcmaster:
After further review, it appears anyone scripting dotnet-ef with json output always needs to parse out other stdout output anyways. When we trigger a dotnet-build or when we add a migration, there is now other output on stdout. We might as well always log "//BEGIN" and "//END" for json output.

cc @bricelam


bricelam:
:shipit:

Issue #5581, opened 2016-05-31
Provide Migration "Hooks"


glatzert:
### The issue

When I create a ef-migration during development, that migration often will not be final and might need recreation, due to database creation errors (like incorrect cascade chains or whatever else).
When I now try to provide additional methods for `Up()` and `Down()` i will have to add them to the migration, everytime I recreate it.
That could be solved, if the `partial class MyMigration` would provide partial stub methods, which I might or might not implement like `PreUp()`, `PreDown()`, `PostUp()`, `PostDown()`.
This would allow to add StoredProcedures, Triggers or Schemas, which are not directly generated by EF in that methods and not lose that code on recreating the migration.
#### Example

 MyMigration_20160531132500.cs

``` c#
 public partial class InitialDatabaseSetup : Migration
    {
        protected override void Up(MigrationBuilder migrationBuilder)
        {
            PreUp(migrationBuilder);

            // .. original Migration Code .. //

            PostUp(migrationBuilder);
        }

        protected override void Down(MigrationBuilder migrationBuilder)
        {
            PreDown(migrationBuilder);

            // .. original Migration Code .. //

            PostDown(migrationBuilder);
        }   

        protected partial void PreUp(MigrationBuilder migrationBuilder);
        protected partial void PreDown(MigrationBuilder migrationBuilder);
        protected partial void PostUp(MigrationBuilder migrationBuilder);
        protected partial void PostDown(MigrationBuilder migrationBuilder);
    }
```

MyMigration.Programability.cs

``` c#
 public partial class InitialDatabaseSetup : Migration
    {
        protected partial void PreDown(MigrationBuilder migrationBuilder){
            // .. my own code to delete SPs .. //
        }
        protected partial void PostUp(MigrationBuilder migrationBuilder) {
            // .. my own code to create SPs .. //
        }
    }
```

bricelam:
You might be able to add them yourself by overriding `UpOperations` and `DownOperations` properties.

bricelam:
But I do like the idea of adding some hooks that are easier to use in an external partial. Perhaps like `BeforeUp`, `AfterUp`, etc. methods on `Migration` that you can override.

glatzert:
Uh.. that `UpOperations` and `DownOperations` looks awfully complicated compared to the possibility of some Before and After hooks. I suppose the hooks would be an inexpensive change, which would allow usecases like the descripted. That'd really be a "nice to have", I think 👍 . I'd also volunteer myself for an PR :)

[...]

PR #5607, opened 2016-06-01
Allow users to add custom up/down code in external partial class

glatzert:
- added 4 Methods BeforeUp, BeforeDown, AfterUp, AfterDown
 - changed upCommands and downCommands creation

Addresses #5581


rowanmiller:
We discussed as a team and we're no 100% sure that this is the right approach to take. Typically, if you have things you want to do before/after, but not have them included alongside the scaffolded code, then you would scaffold an additional blank migration and put the operations in there.

The main concern is that we don't want to introduce APIs etc. when we're not sure that it's something that many folks will want. This is the first request we've seen for the feature, so at this stage we'd rather wait for more feedback that folks want it.

mhamri:
Please merge this https://github.com/aspnet/EntityFrameworkCore/issues/18776

PR #5608, opened 2016-06-01
Change optimistic concurrency tests to use a shadow property

AndriySvyryd:
This allows other providers to use different concurrency tokens in the model


ajcvickers:
:shipit:

Issue #3928, opened 2015-11-30
SQL Server Migrations: Batch migrations commands


bricelam:
(This came up in a hallway conversation a while ago; filing to track.)

We implemented #2739 to help providers that don't support batching, but we could (if we wanted to) update our providers to continue batching commands.

Here are some of my original arguments for removing it that are also relevant to this issue:

> Its value-add seems minimal. We'd have to add API to MigrationBuilder for ending batches.

balivo:
I don't know if it helps or if it is interesting to include in the "core" of the EF7...
But I created a "shortcut" method for executing SQL statements from a file of "Resource", follows:

<code>
public static void ExecuteSqlFromEmbeddedResource(MigrationBuilder pMigrationBuilder, string pResourceName)
        {
            var _assembly = Assembly.GetExecutingAssembly();

            var _resourceName = _assembly.GetManifestResourceNames().FirstOrDefault(lbda => lbda.EndsWith(pResourceName));

            using (var _stream = _assembly.GetManifestResourceStream(_resourceName))
            {
                if (_stream != null)
                {
                    using (var _reader = new StreamReader(_stream))
                    {
                        pMigrationBuilder.Sql(_reader.ReadToEnd());
                    }
                }
            }
        }
</code>

"pResourceName" is the name of (file) in resource, generally I use the name of Migration.

bricelam:
Marking for re-triage. Doing this would be a good way to validate the APIs added in #4917.

bricelam:
I prototyped this, and #5610 includes the breaking changes necessary to enable us to implement this later.

bricelam:
I agreed that the actual implementation of this should wait until post-RTM when we have more time to verify it.

bricelam:
On a side note, implementing this will create **much cleaner** idempotent scripts. We currently put `IF` statements around each command. Implementing this would put them all inside the same `IF` statement.

bricelam:
Prototyped in [bricelam:batch](https://github.com/aspnet/EntityFrameworkCore/compare/master...bricelam:batch)

PR #5610, opened 2016-06-01
Migrations: Make MigrationsSqlGenerator.EndStatement batching-friendly

bricelam:
This decouples statement termination from `SqlCommand` creation, and enables us to implement #3928 later.


ajcvickers:
:shipit:

lajones:
This looks good to me, but does it really not change the results of any tests / shouldn't there be some tests for which this would change the results?

bricelam:
@lajones Nope, this is a pure refactoring with no functional impact.

bricelam:
Well... that satement wasn't 100% true. Providers may need to add a few `.AppendLine(SqlGenerationHelper.StatementTerminator)` calls like I did in SQL Server.

lajones:
OK - then :shipit:

PR #5618, opened 2016-06-01
Fix #5376 - escape parameters passed to dotnet-ef from PMC

### LINKED ISSUE #5376: Reverse-Engineer: Scaffold-DbContext not working on ASP.NET Core project
Found when updating this tutorial https://docs.efproject.net/en/latest/platforms/aspnetcore/existing-db.html

I'm guessing this is an issue in the way we hand off parameters to the CLI command that we shell out to.

```
PM> Scaffold-DbContext "Server=(localdb)\mssqllocaldb;Database=Blogging;Trusted_Connection=True;" Microsoft.EntityFrameworkCore.SqlServer 
The term 'localdb' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

natemcmaster:
cc @lajones @bricelam 

This also fixes out we treat verbose output and stderr.

Escapes all parameters passed to dotnet-ef with single quotes.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psm1
+++ b/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psm1
@@ -657,8 +657,6 @@ function InvokeDotNetEf($project, [switch] $json, [switch] $skipBuild) {
         throw "Cannot execute this command because 'Microsoft.EntityFrameworkCore.Tools' is not installed in project '$projectName'. Add 'Microsoft.EntityFrameworkCore.Tools' to the 'tools' section in project.json. See http://go.microsoft.com/fwlink/?LinkId=798221 for more details."
     }
 
-    $output = $null
-
     $config = $project.ConfigurationManager.ActiveConfiguration.ConfigurationName
     $arguments = "--configuration", $config
     Write-Debug "Using configuration $config"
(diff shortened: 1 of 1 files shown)


src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psm1
```diff
@@ -675,39 +673,45 @@ function InvokeDotNetEf($project, [switch] $json, [switch] $skipBuild) {
...
         Write-Debug "Finish executing command with code $exit"
         if ($exit -ne 0) {
-            if (!($verboseOutput) -and $output) {
+            if (!($stderr) -and $stdout) {
                 # most often occurs when Microsoft.EntityFrameworkCore.Tools didn't install
-                throw $output
+                throw $stdout
             }
-            throw $verboseOutput
+            throw $stderr
```
lajones: What happens if there is no stderr? I.e. does this need `if ($stderr) { ... }` around it?

natemcmaster: See line above. I could remove the `-and $stdout` predicate because that is always true...

lajones: Let me put it another way - what happens if **both** stderr and stdout are empty? Does the call at line 704 error?

natemcmaster: I suppose it would, but I do not know of any case when a call to dotnet returns with a non-zero exit code an has no output on either stderr or stdout

lajones: OK. I'd rather issue an error message in that scenario, but that's not the main point of this fix.

lajones:
:shipit:

Issue #5059, opened 2016-04-13
Repeated allocations in InterceptingLogger<T>


rynowak:
Data is from [Basic Views](https://github.com/aspnet/Mvc/issues/4175) 

Sample includes 3001 requests.

`TypeExtensions.DisplayName` is being called each time one of these are constructed, which in this scenario is 6 times per request
![image](https://cloud.githubusercontent.com/assets/1430011/14511423/473e1626-018c-11e6-859a-a5f40f85c92c.png)

Allocations are spread across `string`, `char[]`, and `StringBuilder`. I included the major ones here, but summing up everything I was able to find, this is worth about 8mb-9mb total or 2.8kb/request.

![image](https://cloud.githubusercontent.com/assets/1430011/14511372/eb78f4a0-018b-11e6-9a16-c7520ba4f202.png)
![image](https://cloud.githubusercontent.com/assets/1430011/14511395/11d2516e-018c-11e6-857f-54776789d445.png)
![image](https://cloud.githubusercontent.com/assets/1430011/14511325/976973c6-018b-11e6-8ccb-75b0acd81b3d.png)

divega:
rynowak just to make sure, is the main assertion you are making that the `TypeNameHelper.GetDisplayName()` method used in the default logging implementation is more efficient than `TypeExtensions.DisplayName()` or that we are doing too much logging on too many classes? :smile:

If the former, I think we can probably copy the implementation or use inheritance to get it.

rynowak:
`Logger<T>` is registered as a singleton and so the name is computed once per-logger-type per-app. `InterceptingLogger<T>` is scoped, so all of these names are computed once per request.

divega:
Got it. Thanks!

PR #5619, opened 2016-06-01
Reduce allocations in construction of InterceptingLogger

ajcvickers:
Cache logger name for the type to avoid calling DisplayName each time a logger instance is created.

Issue #5059


lajones:
:shipit:

PR #5622, opened 2016-06-02
Issue 'IS NULL' instead of '= null' in condition expressions during update.

### LINKED ISSUE #3023: Update Pipeline: Use C# null semantics for WHERE clauses (concurrency tokens)
I'm having some OptimisticConcurrency tests fail because the following SQL gets sent to the database:

```sql
UPDATE "Drivers" SET "Podiums" = $1 WHERE "Id" = $2 AND "Version" = $3 RETURNING "Version"
```
with $3 being equal to null.

This may work in other databases, but PostgreSQL always returns false for `= NULL` even when using a parameter. Now I can see that in a *select query* EF generates different SQL (IS NULL vs. equality operator) based on whether the parameter is null or not (I wonder how this interacts with caching, do you key on the nullness of parameters?). However, this logic seems to be totally absent from [UpdateSqlGenerator](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Update/UpdateSqlGenerator.cs#L317), where the equality operator is always used.

Any tips on how to proceed? I could override `AppendWhereCondition` and use PostgreSQL's `IS NOT DISTINCT FROM` as a workaround (it's like the equality operator except it returns true for nulls). I could also test the ColumnModification's `Value` and generate `IS NULL` or equality based on that (although is the update SQL cached somewhere?).

However, this seems like something that needs to be handled outside the provider in Relational... Ideally, shouldn't the code generating where clauses (and perhaps other SQL components) be shared between query generation and modification generation?

AndriySvyryd:
Fixes #3023

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Update/UpdateSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Update/UpdateSqlGenerator.cs
@@ -350,10 +350,22 @@ namespace Microsoft.EntityFrameworkCore.Update
             Check.NotNull(columnModification, nameof(columnModification));
 
             SqlGenerationHelper.DelimitIdentifier(commandStringBuilder, columnModification.ColumnName);
-            commandStringBuilder.Append(" = ");
-            SqlGenerationHelper.GenerateParameterName(commandStringBuilder, useOriginalValue
-                ? columnModification.OriginalParameterName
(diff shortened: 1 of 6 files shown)


AndriySvyryd:
@roji Does this work for you?

ajcvickers:
:shipit:

Issue #5623, opened 2016-06-02
EF core RC2 One to One relationship


1amirjalai:
### Steps to reproduce
Following the instruction [here](http://ef.readthedocs.io/en/latest/modeling/relationships.html#one-to-one) at the ef core doc.
and then using command line

> dotnet ef migrations add
> dotnet ef database update

the created relationship is one to many instead of one to one.

also i have tried to create models like this :
```
   public class Blog
    {
        public int BlogId { get; set; }
        public string Url { get; set; }
        public int BlogImageId { get; set; }
        public virtual BlogImage BlogImage { get; set; }
    }
```

```
    public class BlogImage
    {
        public int BlogImageId { get; set; }
        public byte[] Image { get; set; }
        public string Caption { get; set; }
        public int BlogId { get; set; }
        public virtual Blog Blog { get; set; }
    } 

```
   and the related modelbuilder
```

  modelBuilder.Entity<Blog>()
                .HasOne(p => p.BlogImage)
                .WithOne(i => i.Blog)
                .HasForeignKey<BlogImage>(b => b.BlogId);
```
with no success
### The issue

One to Many relationship is created instead of one to one

 
### Further technical details

EF Core version: "1.0.0-rc2-final"
Operating system: 
Visual Studio version: 2015 update 2 enterprise

Other details about my project setup:

smitpatel:
Can you also explain what you mean by `One to Many relationship is created instead of one to one` ?

ajcvickers:
Confirmed that the index created for the FK is not unique.
![image](https://cloud.githubusercontent.com/assets/1430078/15751956/e5761704-28a0-11e6-9c16-60a8197def91.png)

divega:
Same here.

``` C#
using System;
using System.Linq;
using Microsoft.EntityFrameworkCore;
using Xunit;

namespace Repro5623
{
    public class Tests
    {
        [Fact]
        public void One_to_one_relationships_are_one_to_one()
        {
            using (var context = new MyContext())
            {
                context.Database.EnsureDeleted();
                context.Database.EnsureCreated();
            }

            using (var context = new MyContext())
            {
                var blog = new Blog { Url = "http://blogs.msdn.com/diego" };
                var blogImage1 = new BlogImage { Caption = "Image1", Blog = blog };
                context.BlogImages.Add(blogImage1);
                context.SaveChanges();
            }

            using (var context = new MyContext())
            {
                var blog = context.Blogs.First();
[...]

Issue #5415, opened 2016-05-19
Migrations: Opt out from IS NOT NULL in index creation


rmja:
[When creating an index in migrations](https://github.com/aspnet/EntityFramework/blob/ce07379ec2d5e5e6d3e6077267bcd7c1dde58f92/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs#L198) it would be nice if it was possible to specify on a per column basis whether the column should be generated as NOT NULL. Right now all columns are NOT NULL if the index is not clustered.

More generally, it would be nice if provider specific details were implemented in seperate overrideable methods, so that one could implement this with a custom migrations SQL generator. For example:

```
protected override void Generate(CreateIndexOperation operation, IModel model, MigrationCommandListBuilder builder)
        {
            Check.NotNull(operation, nameof(operation));
            Check.NotNull(builder, nameof(builder));

            base.Generate(operation, model, builder, terminate: false);
            AppendIndexWhere(operation, model, builder);

            EndStatement(builder);
        }

protected virtual void AppendIndexWhere(CreateIndexOperation operation, IModel model, MigrationCommandListBuilder builder)
{
            var clustered = operation[SqlServerFullAnnotationNames.Instance.Clustered] as bool?;
            if (operation.IsUnique
                && (clustered != true))
            {
                builder.Append(" WHERE ");
                for (var i = 0; i < operation.Columns.Length; i++)
                {
                    if (i != 0)
                    {
                        builder.Append(" AND ");
                    }

                    builder
                        .Append(SqlGenerationHelper.DelimitIdentifier(operation.Columns[i]))
                        .Append(" IS NOT NULL");
                }
            }
}
```

In this way I can call `base.Generate()` from a derived custom SQL generator and override the provider behaviour where needed by overriding each step. This is because I cannot call `base.base.Generate()` from my custom SQL generator, so it is currently an all or nothing situation.

smitpatel:
You can override `AppendIndexWhere` in derived class and opt out of NOT NULL condition. Is it not working for you?

rmja:
[...]

PR #5631, opened 2016-06-02
Enforce uniqueness of foreign keys in the database

ajcvickers:
Issue #5623

The index created to go alongside the relationship was not getting its uniqueness from the relationship, and hence the database created by Migrations did not have a unique index for the FK column. The fix is to propagate uniqueness from the relationship metadata to the index.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -968,16 +968,16 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public virtual InternalIndexBuilder HasIndex([NotNull] IReadOnlyList<string> propertyNames, ConfigurationSource configurationSource)
-            => HasIndex(GetOrCreateProperties(propertyNames, configurationSource), configurationSource);
+            => HasIndex(GetOrCreateProperties(propertyNames, configurationSource), null, configurationSource);
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public virtual InternalIndexBuilder HasIndex([NotNull] IReadOnlyList<PropertyInfo> clrProperties, ConfigurationSource configurationSource)
-            => HasIndex(GetOrCreateProperties(clrProperties, configurationSource), configurationSource);
+            => HasIndex(GetOrCreateProperties(clrProperties, configurationSource), null, configurationSource);
 
-        private InternalIndexBuilder HasIndex(IReadOnlyList<Property> properties, ConfigurationSource configurationSource)
+        private InternalIndexBuilder HasIndex(IReadOnlyList<Property> properties, bool? unique, ConfigurationSource configurationSource)
         {
             if (properties == null)
             {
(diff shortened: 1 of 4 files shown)


divega:
:shipit:

src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -1222,13 +1229,13 @@ private static IndexBuildersSnapshot DetachIndexes(IEnumerable<Index> indexesToD
...
             bool runConventions)
         {
-            var key = Metadata.AddForeignKey(dependentProperties, principalKey, principalType, configurationSource: null, runConventions: false);
-            key.UpdateConfigurationSource(configurationSource);
+            var foreignKey = Metadata.AddForeignKey(dependentProperties, principalKey, principalType, configurationSource: null, runConventions: false);
+            foreignKey.UpdateConfigurationSource(configurationSource);
             principalType.UpdateConfigurationSource(configurationSource);
 
-            HasIndex(dependentProperties, ConfigurationSource.Convention);
+            HasIndex(dependentProperties, foreignKey.IsUnique, ConfigurationSource.Convention);
```
smitpatel: `IsUnique` function on `InternalIndexBuilder` can be used here by chaining.

ErikEJ:
@divega Wonder if this could cause issues with a sql engine that does not support filtered indexes? 
A number of tests now fail for me due to inserts failing  because of a unique index on nullable fk columns?? @ajcvickers ??

ajcvickers:
@ErikEJ Nothing about the way EF behaves has changed, so it sounds like if SQL Compact can't support enforcing uniqueness in these cases then its implementation of Migrations should not make the columns unique.

divega:
@ErikEJ, I agree with @ajcvickers. You get to choose between:
1. Creating the unique index without filters therefore allowing only one dependent row with a null FK
2. Not creating the index at all, which means the database won't enforce the relationship to be 1:0..1 at all times

I would go with (2) because (as @ajcvickers pointed out when we had a chat about this) as long as the FK is nullable, creating more than one dependent with a null FK is valid, and the unique index is only there to protect the application developer from making a mistake.

The good thing is that you can still create unique indexes if the FK is non-nullable.

ajcvickers:
@divega Wouldn't it be better to still create the index, but just not make it unique?

divega:
Yes. Good catch.

ErikEJ:
@ajcvickers @divega Thank so much for the prompt replies, I really appreciate the level of support I get form your guys. I understand the workaround, but I am a Little unsure how to get the information about the fact that a index hits a nullable FK column - maybe @bricelam has some input - I hope!

I understand that you are not facing this issue on SQL Server due to the fact that you create a filtered index (WHERE [Col] IS NOT NULL) - otherwise you would see the same issue there, caused by mulitple NULL values in the index. I hope you are aware of the implications of this, in particular when mixing "EF" owned objects with legacy objects in a large database - see the section: "Required SET Options for Filtered Indexes" here: https://msdn.microsoft.com/en-us/library/ms188783.aspx 

For SQLite:
"For the purposes of unique indices, all NULL values are considered to different from all other NULL values and are thus unique. This is one of the two possible interpretations of the SQL-92 standard (the language in the standard is ambiguous) and is the interpretation followed by PostgreSQL, MySQL, Firebird, and Oracle. Informix and Microsoft SQL Server follow the other interpretation of the standard." -  https://www.sqlite.org/lang_createindex.html

ErikEJ:
@bricelam - I did soemthing like this:

     var isNullableForeignKey = operation
                .Columns.Select(column => FindProperty(model, null, operation.Table, column))
                .Any(property => (property != null) && property.IsColumnNullable() && property.IsForeignKey());

            if (operation.IsUnique && !isNullableForeignKey)
            {
                builder.Append("UNIQUE ");
            }

bricelam:
Perfect. It might also be worth doing for all nullable columns (not just foreign keys). Most databases treat `NULL` values as distinct from each other. Ultimately, however, it's up to the application, and that's why we have issue #5415. :wink:

Issue #5634, opened 2016-06-02
Should MethodInfoExtensions be public in System.Reflection


rowanmiller:
Not sure that we should really have this (`Microsoft.EntityFrameworkCore\Extensions\MethodInfoExtensions.cs`)
```
namespace System.Reflection
{
    public static class MethodInfoExtensions
    {
             public static bool MethodIsClosedFormOf([NotNull] this MethodInfo methodInfo, ...)
    }
```

rowanmiller:
BTW it's used in Relational and InMemory, so my proposal is to move it to M.EFC.Internal namespace.

PR #5639, opened 2016-06-03
Make MethodInfoExtensions internal

ajcvickers:
Issue #5634

(Won't check this in until we have triaged that we want to do it.)


divega:
:shipit: (pending triage)

PR #5648, opened 2016-06-03
Doc comments for Core query

anpete:
Moved some lower level query components to internal.

cc @rowanmiller


src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
```diff
@@ -56,15 +70,60 @@ public class QueryCompilationContext
             TrackQueryResults = trackQueryResults;
         }
 
+        /// <summary>
+        ///     Gets the model.
+        /// </summary>
+        /// <value>
+        ///     The model.
+        /// </value>
```
rowanmiller: I haven't been including these <value> tags as they are pretty redundant - the auto-templating in VS actually doesn't emit them anymore. I don't think there is any issue including them, just don't feel like you have to include them in future docs.

rowanmiller:
:shipit:

PR #5649, opened 2016-06-03
Supressed warnings in test code.

### LINKED ISSUE #5647: Build warning in GearsOfWarQueryTestBase
```
GearsOfWarQueryTestBase.cs(605,34,605,57): warning CS0458: The result of the expression is always 'null' of type 'AmmunitionType?'
```

smitpatel:
Resolves #5647 
@anpete @pranavkm @sebastienros 

@Eilon for ASK mode


test/Microsoft.EntityFrameworkCore.Tools.Core.Tests/Utilities/JsonUtilityTest.cs
```diff
@@ -12,7 +12,9 @@ namespace Microsoft.EntityFrameworkCore.Tests.Utilities
 {
     public class JsonUtilityTest
     {
+#pragma warning disable 649
         private readonly ITestOutputHelper _output;
```
pranavkm: Couldn't you just remove this?

anpete:
@smitpatel Do we need to do it for csprojs, too?

:shipit:

test/Microsoft.EntityFrameworkCore.Tools.Core.Tests/Utilities/JsonUtilityTest.cs
smitpatel: `_output` is assigned value in ctor but it is commented out since it is there for debugging only.

anpete: Comment out field too?

smitpatel: The field is being used in file multiple times. Commenting out fields means commenting out all occurrences of usages also.

smitpatel:
Updated for csprojs

smitpatel:
Now important question
@pranavkm , @anpete @divega - Is this ask mode worthy?

divega:
To make sure I understand:

- This is a test only change, right?
- Does it make the build pass more cleanly?

If the answer to both is true I think we should bring it up.

pranavkm:
It fixes a few warnings (which is a good things). Adding `warningsAsErrors` is nice to have to prevent the recurrence of these issues.

PR #5656, opened 2016-06-04
Query: Convert boolean value to search condition in sql predicates

### LINKED ISSUE #5652: Invalid SQL statement produced from null-coalescing a nullable boolean
##### Database:
```
CREATE TABLE dbo.Items (
    Id int NOT NULL PRIMARY KEY,
    Enabled bit NULL,
    Value nvarchar(max) NOT NULL
);
```

##### Entity:
```
public class Items
{
    [Key]
    public int Id { get; set; }
    public string Value { get; set; }
    public bool? Enabled { get; set; }
}
```

##### Query:
```
var items = _context.Items.Where(i => i.Enabled ?? false).ToList();
```

##### Generated SQL
```
SELECT [i].[Id], [i].[Enabled], [i].[Value]
FROM [Items] AS [i]
WHERE COALESCE([i].[Enabled], 0)
```

##### Exception message (from SQL)
```
An expression of non-boolean type specified in a context where a condition is expected
```

### The issue

Using the null-coalescing operator with a nullable boolean (`bool?`) in a `Where` clause produces invalid SQL. I would expect this LINQ statement to produce something like this:
`WHERE (COALESCE([i].[Enabled], 0) = 1`

Is there a format that EF Core expects for a query like this?

### Further technical details
EF Core version: EntityFramework 7.0.0-rc1-final
Operating system: Windows 8
Visual Studio version: VS 2015

smitpatel:
resolves #5652 
Issue: In Sql bool values and conditions are different concept and implicit conversion is not done. Therefore we need to append `= 1` to where predicate for certain expression. So far we had a list of expression which needed it. `Coalesce` was not part of the list hence incorrect SQL generated.
Solution: Instead of adding another special case for `Coalesce` here we generalized kind of expression which are valid to be used as search condition and generated `= 1` for all the rest. The CFG was taken from T-SQL which covers all cases. (supposedly)
@anpete @maumar @divega

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -169,31 +169,13 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
 
             if (selectExpression.Predicate != null)
             {
-                var nullSemanticsPredicate = ApplyNullSemantics(selectExpression.Predicate);
-                var constantExpression = nullSemanticsPredicate as ConstantExpression;
-
-                if (constantExpression == null
-                    || !(bool)constantExpression.Value)
+                var optimizedPredicate = ApplyOptimizations(selectExpression.Predicate, searchCondition: true);
+                if (optimizedPredicate != null)
                 {
                     _relationalCommandBuilder.AppendLine()
                         .Append("WHERE ");
 
-                    if (constantExpression != null)
-                    {
-                        _relationalCommandBuilder.Append("1 = 0");
-                    }
-                    else
-                    {
-                        Visit(nullSemanticsPredicate);
-
-                        if (selectExpression.Predicate is ParameterExpression
-                            || selectExpression.Predicate.IsAliasWithColumnExpression()
-                            || selectExpression.Predicate is SelectExpression)
-                        {
-                            _relationalCommandBuilder.Append(" = ");
-                            _relationalCommandBuilder.Append(TrueLiteral);
-                        }
-                    }
+                    Visit(optimizedPredicate);
                 }
             }
 
@@ -223,7 +205,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
             return selectExpression;
         }
 
-        private Expression ApplyNullSemantics(Expression expression)
+        private Expression ApplyOptimizations(Expression expression, bool searchCondition)
         {
             var newExpression
                 = new NullComparisonTransformingVisitor(_parametersValues)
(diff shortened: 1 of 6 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1210,11 +1208,32 @@ var parameterName
...
-            return column?.Property != null 
-                ? _relationalTypeMapper.FindMapping(column.Property) 
+            return column?.Property != null
+                ? _relationalTypeMapper.FindMapping(column.Property)
                 : null;
         }
 
+        protected virtual bool IsSearchCondition(Expression expression)
+        {
+            while (expression.NodeType == ExpressionType.Convert)
```
maumar: you can just do expression.RemoveConvert()

src/Microsoft.EntityFrameworkCore.Specification.Tests/GearsOfWarQueryTestBase.cs
```diff
@@ -1454,6 +1454,17 @@ public virtual void Include_on_GroupJoin_SelectMany_DefaultIfEmpty_with_complex_
             }
         }
 
+        [ConditionalFact]
+        public virtual void Coalesce_operator_in_predicate()
```
maumar: also add tests for other cases, just to make sure

smitpatel: Which are the other cases? 😄

maumar: like, in, string compare should be more than enough

smitpatel: We already have tests which use those operators as root predicate.

maumar:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
smitpatel: I did not know it was recursive 😄

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1210,11 +1208,29 @@ var parameterName
...
                 : null;
         }
 
+        protected virtual bool IsSearchCondition(Expression expression)
+        {
+            expression = expression.RemoveConvert();
+
+            if (expression.IsComparisonOperation()
+                || expression.IsLogicalOperation()
+                || expression is LikeExpression
```
anpete: Short-circuit when NodeType != Extension?

smitpatel: AliasExpression is also Extension.

anpete: I meant you can skip checking any node that is not an extension (and not a binary).

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -186,9 +186,7 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
                     {
                         Visit(nullSemanticsPredicate);
 
-                        if (selectExpression.Predicate is ParameterExpression
-                            || selectExpression.Predicate.IsAliasWithColumnExpression()
-                            || selectExpression.Predicate is SelectExpression)
+                        if (!IsSearchCondition(nullSemanticsPredicate))
```
anpete: Doesn't the compensation need to apply to any node in the predicate tree. I.e. Does this work for coalesce as one side of a binary op?

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
smitpatel: Ah, ok!

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
smitpatel: That thought crossed my mind and then I got confused that we are only appending `= 1` at the root level of the predicate. We did have 3 cases where we compensated. Either we don't hve test coverage for those coming as a node or we are doing something to compensate that in some other way. I will look into it more.

smitpatel: Verified this. We don't have test coverage. Need to change the solution to generalize it for all cases.

smitpatel:
Holding off this PR to make further changes to generalize whole scenario.

smitpatel:
@anpete @maumar - Updated source code (for most part). Will be adding more tests for thorough verification of scenario. Does approach look fine?

smitpatel:
Updated PR. Ready for review.

anpete:
:shipit:

PR #5658, opened 2016-06-04
Add R# StringFormatMethod to FromSql

roji:
Add R# StringFormatMethod to FromSql

Will help users avoid parameter referencing and formatting issues in SQL queries passed to FromSql.

A good example is a user having curly braces somewhere in the SQL query, that will make `string.Format` bomb internally.


divega:
@ajcvickers @anpete we can take this. Can one of you review it and merge it if it is good?

anpete:
@divega Will do.

PR #5671, opened 2016-06-06
Commands: Flow app info into IDbContextFactory

### LINKED ISSUE #5615: Commands: Flow info into IDbContextFactory
There are two pieces of information that should somehow be flowed into `IDbContextFactory` so that users can properly implement it in a way that works well with the commands:
* Environment
* Startup project directory

bricelam:
Resolves #5615

Verification: Manually verified

cc @rowanmiller to review documentation


anpete:
:shipit:

PR #5676, opened 2016-06-07
Update conditional skip to exclude sequence test only on Linux/OSX

natemcmaster:
cc @BrennanConroy

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SequenceEndToEndTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SequenceEndToEndTest.cs
@@ -114,7 +114,8 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests
             }
         }
 
-        [ConditionalFact(Skip = "Test is flaky on Linux. See https://github.com/dotnet/corefx/issues/8701")]
+        [ConditionalFact]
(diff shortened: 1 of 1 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SequenceEndToEndTest.cs
```diff
@@ -114,7 +114,8 @@ private static void AddEntities(IServiceProvider serviceProvider)
             }
         }
 
-        [ConditionalFact(Skip = "Test is flaky on Linux. See https://github.com/dotnet/corefx/issues/8701")]
+        [ConditionalFact]
+        [PlatformSkipCondition(TestPlatform.Mac | TestPlatform.Linux, SkipReason = "Test is flaky on OSX/Linux. See https://github.com/dotnet/corefx/issues/8701")]
```
BrennanConroy: Any reason you're using a custom attribute in EF instead of the OSSkipConditionAttribute from Testing?

BrennanConroy:
:shipit:

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SequenceEndToEndTest.cs
natemcmaster: EF doesn't depend on Microsoft.AspNet.Testing. We have similar implementations, but needed slightly different things.

PR #5678, opened 2016-06-07
Use RuntimeEnvironment.Sources

BrennanConroy:
@natemcmaster

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/Xunit/FrameworkSkipConditionAttribute.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/Xunit/FrameworkSkipConditionAttribute.cs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.Extensions.PlatformAbstractions;
+using System.Runtime.InteropServices;
 
 namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestUtilities.Xunit
 {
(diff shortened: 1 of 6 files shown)


src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/Xunit/FrameworkSkipConditionAttribute.cs
```diff
@@ -27,27 +26,24 @@ private static bool CanRunOnThisFramework(RuntimeFrameworks excludedFrameworks)
                 return true;
             }
 
+#if NET451
```
natemcmaster: We shouldn't cross-compile this check out. It's still technically possible to run netstandard libraries on Mono.

src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/Xunit/FrameworkSkipConditionAttribute.cs
```diff
@@ -27,27 +26,24 @@ private static bool CanRunOnThisFramework(RuntimeFrameworks excludedFrameworks)
...
+            if (excludedFrameworks.HasFlag(RuntimeFrameworks.CLR))
             {
                 return false;
             }
-
-            if (excludedFrameworks.HasFlag(RuntimeFrameworks.CoreCLR)
-                &&
-                PlatformServices.Default.Runtime.RuntimeType.Equals("CoreCLR", StringComparison.OrdinalIgnoreCase))
+#else
+            if (excludedFrameworks.HasFlag(RuntimeFrameworks.CoreCLR))
```
natemcmaster: Likewise. We can't rely on how this was compiled to be sure if this is running on coreclr or not. This library is compiled for .NET Standard 1.3, which could technically run just fine on .NET 4.6.

natemcmaster:
:shipit:

Issue #5320, opened 2016-05-10
Tools: Support .NET Core Class Library projects


natemcmaster:
Add support for running "dotnet ef" on a .NET Standard and desktop .NET class library project.

This is blocked by https://github.com/dotnet/cli/issues/2645 ~~and/or https://github.com/dotnet/cli/issues/4005 (depending on our implementation approach)~~.
### Short explanation

If you cannot do `dotnet run`, you also cannot use `dotnet ef` on your project. 
### More explanation

.NET Core, unlike .NET Framework, does not have a globally installed framework. All .NET Core apps need to reference an app framework (e.g. Microsoft.NETCore.App) in order to run. .NET Core class library projects are built without a framework, so dotnet-ef cannot run on these projects.
### Workarounds

See https://docs.efproject.net/en/latest/miscellaneous/cli/dotnet.html#targeting-class-library-projects-is-not-supported for workarounds.

Kurira:
Hi,

Just to be sure, we can't use EF within a class library project (aka tests project) in RC2?

natemcmaster:
@Kurira you can use EF within a class library. You can't use **EF .NET Core CLI commands** on a class library. See https://docs.efproject.net/en/latest/cli/dotnet.html more info

Kurira:
So I would use tools from my web project and apply them to test context which resides in test project?

kjbetz:
I need to test this more, but here is my solution that seems to be working: https://github.com/aspnet/EntityFramework/issues/5460#issuecomment-221167712

ToddThomson:
Is there a published workaround for this issue? I've added:

    "frameworks": {
        "net451": {
            "buildOptions": {
                "emitEntryPoint": true
            }
        },
        "netcoreapp1.0": {
            "imports": [ "portable-net451+win8" ],
            "buildOptions": {
                "emitEntryPoint": true
            },
            "dependencies": {
                "Microsoft.NETCore.App": {
                    "type": "platform",
                    "version": "1.0.0-*"
                }
            }
        }
    },

and a main() in program.cs, but I still get the "..Entity Framework tools does not support targeting class library projects.." message.

natemcmaster:
@ToddThomson you are probably hitting https://github.com/dotnet/cli/issues/3164. Try setting the top-level emitEntryPoint build option to true.

ToddThomson:
@natemcmaster Yes. That was it.

rowanmiller:
[...]

Issue #5188, opened 2016-04-26
CLI Commands: usability improvements


natemcmaster:
Some ideas for improving the usability of dotnet ef.

- [ ] `dotnet ef database update` doesn't list the migrations being applied/reverted
- [ ] Don't require building to get output of `--help`. We can show help dialog without dispatching to inside-man.
- [x] Support `--json` on all subcommands
- [ ] Allow top-level arguments in any order. E.g. currently `dotnet ef dbcontext list --no-build` throws, but `dotnet ef --no-build dbcontext list` works. ~~Part of this is just the way that [Microsoft.Extensions.CommandLineUtils](https://github.com/aspnet/Common/tree/dev/src/Microsoft.Extensions.CommandLineUtils) works.~~ https://github.com/aspnet/Common/pull/131. We still can't completely do this with the way we invoke ef.exe from dotnet-ef.dll.
- [x] ~~Depending on the outcome of https://github.com/dotnet/cli/issues/2514, we may be able to remove "inside-man" dispatch for RTM.~~ NVM. We've pivoted away from ProjectDependencyCommandFactory to our own thing.
- [x] ~~Depending on the resolution to https://github.com/dotnet/cli/issues/2645, we may need to change Ms.EfCore.Tools to also target class libraries. (Support this by cross-compiling the library to netstandard1.5)~~ tracked here: https://github.com/aspnet/EntityFramework/issues/5320
- [x] ~~Help dialog should show options top-level args (`--framework`) even when listing args for subcommands.~~ Fixed in CommandLineUtils.

natemcmaster:
We should do this but it can be post RTM.

VanKichline:
Note that new preview 2 web application templates use 'dotnet bundle' command, so asking for 'dotnet help ef' runs the bundler/minifier and may modify the contents of the project.

natemcmaster:
@VanKichline you can workaround with the `--no-build` parameter. e.g.:
```sh
dotnet ef --no-build --help
dotnet ef --no-build migrations list 
# etc...
```

bricelam:
Closing. Filed remaining work as #6592 and #6593.

PR #5680, opened 2016-06-07
Parameterize directories used by dotnet-ef

### LINKED ISSUE #5311: CLI Commands: support loading startup from a separate project
Add support for supplying a project as a parameter in `--startup-project` that is different than the current project.

This is explicitly disabled in code. See https://github.com/aspnet/EntityFramework/issues/4577#issuecomment-217496222 for background and https://github.com/aspnet/EntityFramework/pull/5312.

natemcmaster:
Fix #5311 by adding back `--startup-project`.

NB: using `--startup-project` requires that "Microsoft.EntityFrameworkCore.Design" be installed startup project, not the current project. This partially enables class libraries. cref #5320

TODO: parameter order on command line still matters, so this option must be added between "ef" and the next sub command. E.g. `dotnet ef --startup-project ../WebApp1 database update`. This is already tracked in #5188 

This is built under the assumption that on .NET Core CLI, the current working directory is your target project, _and is not necessarily the startup project_. The reverse assumption (that dotnet-ef should always be executed on startup) is not consistent with other .NET Core CLI commands that alter files.

Worth noting: we are still using relative project directories as the value of `--startup-project`. This is to stay consistent with how .NET Core CLI handles preferences to projects. (Example see `dotnet run --help`)

cc @bricelam @divega @rowanmiller


natemcmaster:
(PS, before I request ask-mode approval, let's consider first that this is the behavior we want.)

bricelam:
Wonderful! It's exactly how I always wanted it to work.

src/Microsoft.EntityFrameworkCore.Design/CommandLine/ExecuteCommand.cs
```diff
@@ -80,12 +75,16 @@ public static CommandLineApplication Create(ref string[] args)
...
+                StartupAssembly = app.Option(StartupAssemblyOptionTemplate + " <ASSEMBLY>",
+                     "The assembly file containing the startup class."),
                 DataDirectory = app.Option(DataDirectoryOptionTemplate + " <DIR>",
-                    "The folder to use as the data directory. If not specified, the runtime output folder is used."),
+                    "The folder used as the data directory."),
                 ProjectDirectory = app.Option(ProjectDirectoryOptionTemplate + " <DIR>",
-                    "The folder to use as the project directory. If not specified, the current working is used."),
+                    "The folder used as the project directory. If not specified, the current working directory is used."),
+                StartupTargetDirectory = app.Option(StartupTargetDirectoryOptionTemplate + " <DIR>",
+                    "The folder used as the base directory for ASP.NET hosting environment. If not specified, the current working directory is used."),
```
bricelam: This isn't ASP.NET-specific. Maybe something like "...base directory for the application."

bricelam:
:shipit:

src/Microsoft.EntityFrameworkCore.Design/CommandLine/ExecuteCommand.cs
natemcmaster: Can we drop the data directory and merge them into one parameter? I can't think of a scenario in which data directory is different from the app base path.

bricelam: Classic ASP.NET uses `<base>/App_Data`

bricelam: But yes, it's probably OK to do that for now on CLI.

natemcmaster: I think I'm going to leave it, as long as there is the possible scenario that data directory may not be the app base path.

PR #5681, opened 2016-06-07
Add fallback check to determine if projects are .NET Core

natemcmaster:
Helps future-proof PMC commands against rename of *.xproj -> *.csproj

cc @bricelam

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psm1
+++ b/src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psm1
@@ -610,7 +610,7 @@ function ProcessCommonDotnetParameters($dteProject, $dteStartupProject, $Environ
 }
 
 function IsDotNetProject($project) {
-    $project.FileName -like "*.xproj"
+    $project.FileName -like "*.xproj" -or $project.Kind -eq "{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}"
 }
 
 function IsUwpProject($project) {


src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psm1
```diff
@@ -610,7 +610,7 @@ function ProcessCommonDotnetParameters($dteProject, $dteStartupProject, $Environ
 }
 
 function IsDotNetProject($project) {
-    $project.FileName -like "*.xproj"
+    $project.FileName -like "*.xproj" -or $project.Kind -eq "{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}"
```
Eilon: Is the GUID always uppercase? Or is this some kind of "soft" GUID equality comparison?

natemcmaster: `-eq` is case insensitive :)

Eilon: I refuse to learn PowerShell, so thanks 😄

Eilon:
:shipit:

src/Microsoft.EntityFrameworkCore.Tools/tools/EntityFrameworkCore.psm1
bricelam: Pet peeve: Use single quotes unless you need string interpolation!

PR #5683, opened 2016-06-07
Commands: Warn if an error occurs while calling Startup

### LINKED ISSUE #4710: Commands: Don't terminate when Startup isn't constructible

bricelam:
Fixes #4710


natemcmaster:
LGTM. :sheep: :it:

Issue #5674, opened 2016-06-07
RC2: Regression with Parent/Child relationship


rwg0:
### Steps to reproduce

Use the following code to reproduce:

```
    class Program
    {
        static void Main(string[] args)
        {
            if (File.Exists("test.db"))
                File.Delete("test.db");

            using (var context = new WidgetContext())
            {
                context.Database.Migrate();
            }

            using (var context = new WidgetContext())
            {
                for (int i = 0; i < 10; i++)
                {
                    var w = new Widget();
                    context.Widgets.Add(w);
                }

                context.SaveChanges();
            }

            using (var context = new WidgetContext())
            {
                var widgets = context.Widgets.ToList();
                widgets[0].ParentWidget = widgets[1];
                context.SaveChanges();
            }

            using (var context = new WidgetContext())
            {
                var widgets = context.Widgets.ToList();
                try
                {
                    Console.WriteLine(widgets[1].ChildWidgets.Count);
                }
                catch (Exception e)
                {
                    Console.WriteLine("oops, didn't connect childwidgets collection.");
                    Console.WriteLine(e);
                }
                widgets[0].ParentWidget = widgets[2];
                try
                {
                    context.SaveChanges();
                    Console.WriteLine("Saved change of parent.");
                }
                catch (Exception e)
                {
                    Console.WriteLine("oops, couldn't alter the parent Id");
                    Console.WriteLine(e);
                }
            }
        }
    }

    class WidgetContext : DbContext
    {
        public DbSet<Widget> Widgets { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlite("Filename=test.db");
        }
    }

    public class Widget
    {
        public int Id { get; set; }

        public int? ParentWidgetId { get; set; }

        [InverseProperty(nameof(ChildWidgets))]
        [ForeignKey(nameof(ParentWidgetId))]
        public Widget ParentWidget { get; set; }


        [InverseProperty(nameof(ParentWidget))]
        public List<Widget> ChildWidgets { get; set; }
    }
```

Add nuget packages for appropriate version of EntityFramework.Sqlite/EntityFrameworkCore.Sqlite and EntityFramework.Commands/EntityFrameworkCore.Tools.

Create an initial migration.

### The issue
On RC2, the ChildWidgets collection on each Widget when loading the entire table, so the line 
`Console.WriteLine(widgets[1].ChildWidgets.Count);`
[...]

PR #5687, opened 2016-06-07
Do state manager fixup for for more single queries

ajcvickers:
Issue #5674

The state manager was assuming the query would take care of fixup for all single queries. However, this is only true when the relationships are explicitly included in the query. For some cases such as self references and multiple relationships between entity types, fixup is needed that will not be performed by query. Therefore we will now drop out of single query mode for any query with self refs and any time multiple entities are returned. This will have a perf impact, but super simple queries will still run in single query mode. Post RTM we will try to do something better.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/StateManager.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/StateManager.cs
@@ -32,6 +32,8 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
         private IIdentityMap _identityMap1;
         private Dictionary<IKey, IIdentityMap> _identityMaps;
         private bool _needsUnsubscribe;
+        private bool? _singleQueryMode;
+        private IEntityType _singleQueryModeEntityType;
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -67,7 +69,38 @@ public class StateManager : IStateManager
...
+            }
+
+            // If we already checked for self-refs then don't check again.
+            if (_singleQueryModeEntityType == entityType)
+            {
+                return true;
+            }
+
+            // Drop out if SQM for self-refs since query will not fix them up.
+            if (_singleQueryModeEntityType != null
```
anpete: Does this work when combined with inheritance? Any existing metadata helpers you can use here? (ForeignKeyExtensions)

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/StateManager.cs
ajcvickers: Will check.

PR #5689, opened 2016-06-07
Remove imports

### LINKED ISSUE #2192: Many levels of AggregateException (requires update of ix-async)

### LINKED ISSUE #5176: Need external dependencies to support .NET Standard in order to remove the requirement for imports in project.json
Two of EF Core's dependencies, Remotion.Linq and Ix-Async currently use pre-existing portable targets to support multiple platforms. Until these packages add support for .NET Standard, libraries or applications using EF Core will need to specify additional import attributes in project.json. 

```
    Package Ix-Async 1.2.5 is not compatible with netcoreapp1.0 (.NETCoreApp,Version=v1.0). 
    Package Ix-Async 1.2.5 supports:
      - net40 (.NETFramework,Version=v4.0)
      - net45 (.NETFramework,Version=v4.5)
      - portable-net45+win8+wp8 (.NETPortable,Version=v0.0,Profile=Profile78)
    Package Remotion.Linq 2.0.2 is not compatible with netcoreapp1.0 (.NETCoreApp,Version=v1.0). 
    Package Remotion.Linq 2.0.2 supports:
      - net35 (.NETFramework,Version=v3.5)
      - net40 (.NETFramework,Version=v4.0)
      - net45 (.NETFramework,Version=v4.5)
      - portable-net45+win8+wp8+wpa81 (.NETPortable,Version=v0.0,Profile=Profile259)
```

These two issues track the necessary changes in our dependencies:

Reactive-Extensions/Rx.NET#174
https://www.re-motion.org/jira/browse/RMLNQ-98

Cc: @natemcmaster @rowanmiller

natemcmaster:
Fix #5176.
Fix #2192.

We still need imports for test code (for xunit and moq.netcore).

~~:watch: Waiting on updates to package feeds, as well as the updates to Ix-Async and re-linq.~~ Ready now.

cc @divega @Eilon

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/AsyncQueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/AsyncQueryTestBase.cs
@@ -3187,8 +3187,8 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
                 Assert.Equal(
                     CoreStrings.ConcurrentMethodInvocation,
-                    (await Assert.ThrowsAsync<AggregateException>(
-                        async () => await context.Customers.FirstAsync())).InnerException.Message);
+                    (await Assert.ThrowsAsync<InvalidOperationException>(
+                        async () => await context.Customers.FirstAsync())).Message);
             }
         }
 
--- a/src/Microsoft.EntityFrameworkCore.Design/project.json
+++ b/src/Microsoft.EntityFrameworkCore.Design/project.json
@@ -48,7 +48,6 @@
   "frameworks": {
     "net451": {},
     "netcoreapp1.0": {
-      "imports": "portable-net452+win81",
       "dependencies": {
         "Microsoft.NETCore.App": {
           "version": "1.0.0-*",
--- a/src/Microsoft.EntityFrameworkCore.InMemory/project.json
+++ b/src/Microsoft.EntityFrameworkCore.InMemory/project.json
@@ -27,10 +27,6 @@
   },
   "frameworks": {
     "net451": { },
-    "netstandard1.3": {
-      "imports": [
-        "portable-net452+win81"
-      ]
-    }
+    "netstandard1.3": { }
   }
 }
--- a/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/project.json
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/project.json
@@ -30,7 +30,7 @@
     "netstandard1.6": {
       "imports": [
         "dnxcore50",
-        "portable-net452+win81"
+        "portable-net45+win8"
       ],
       "dependencies": {
         "System.Threading.Tasks.Parallel": "4.0.1-*"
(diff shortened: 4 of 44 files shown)


Eilon:
cc @javiercn for your FYI information.

natemcmaster:
Dependencies are up now. Ping for code review.

src/Microsoft.EntityFrameworkCore.Specification.Tests/AsyncQueryTestBase.cs
```diff
@@ -3187,8 +3187,8 @@ protected static string LocalMethod2()
 
                 Assert.Equal(
                     CoreStrings.ConcurrentMethodInvocation,
-                    (await Assert.ThrowsAsync<AggregateException>(
-                        async () => await context.Customers.FirstAsync())).InnerException.Message);
+                    (await Assert.ThrowsAsync<InvalidOperationException>(
```
divega: Is this intentionally part of the PR?

divega: Ah, yeah, I remember now this is because of the improvement in the new version of IX-async.

natemcmaster: Yes. It fixes #2192

divega:
:shipit:

PR #5691, opened 2016-06-08
Avoid recursion when discovering entity types

### LINKED ISSUE #4843: ModelBuilder: Shadow property introduction doesn't work without calling Property<T>()
```C#
class TestClass
{
}

class Test
{
    public int Id { get; set; }
    public TestClass Class { get; set; }
}
```
```C#
modelBuilder.Entity<TestClass>(
    x =>
    {
        // NOTE: HasKey fails without this
        x.Property<string>("AssemblyName");

        x.HasKey("AssemblyName", "Name");
    });
modelBuilder.Entity<Test>(
    x =>
    {
        // NOTE: HasForeignKey fails without this
        x.Property<string>("AssemblyName");

        x.HasOne(t => t.Class).WithMany(c => c.Tests)
            .HasForeignKey("AssemblyName", "ClassName");
    });
```

### LINKED ISSUE #5350: modelBuilder.Entity<AnEntity>(e => { }) can cause StackOverflowException in some cases
I use EntityFrameworkCore version: 1.0.0-rc3-20831, .Net Framework: 4.6, AspNet.Mvc: 5.2.3

I cannot provide certain steps to reproduce. But can point at some conditions. I have about 100 entities. And executing this code causes StackOverflowException (empty stack trace):

`modelBuilder.Entity<AnyEntity>(entity => { /* any code or empty */ })`

This bug is always reproducible on IIS Express, but on Local IIS is reproducible only in rare cases.

I found out that this bug appears only when I uncomment some navigation property in entity, but I cannot track which one nav prop causes this bug (because have large amount of entities).

EXP: If something wrong with my model (maybe some circular referencing?), could you provide more clear exception message. Without this message I can spend much time looking for 'what is wrong with my model'

EXP: Why this is reproducible only with some conditions (or not always). Seems as a bug in model builder algorithm.

By the way: in my previous version of EFCore (January 2016 (don't remember some rc2)) this bug did not exist.

Thanks in advance.

AndriySvyryd:
Also infer FK property types from shadow principal keys, since now relationships to entity types could be created before the key discovery convention runs.

Fixes #5350
Fixes #4843


ajcvickers:
:shipit:

PR #5692, opened 2016-06-08
Fix to #5630 - Select Distinct with where/Any subquery always includes key from joined table, and returns too many rows

### LINKED ISSUE #5630: Select Distinct with where/Any subquery always includes key from joined table, and returns too many rows
### Steps to reproduce

There are 3 very simple tables in this scenario.  You could have 3 tables with a GUID and a string, and that would work fine (GUID is PKEY).  These represent a Drive/Folder/File type situation.  I am attempting to find Drives that have files.

I call the following:
``` C#
(from root in rootSet join folder in folderSet on root.StorageRootId equals folder.RootId
where fileSet.Any(file => file.FolderId == folder.FolderId)
select root.StorageRootId).Distinct();
```

This produces the following query:

``` SQL
SELECT DISTINCT [folder].[FolderId], [root].[StorageRootId]
FROM [StorageRoot] AS [root]
INNER JOIN [StorageFolder] AS [folder] ON [root].[StorageRootId] = [folder].[RootId]
WHERE EXISTS (
    SELECT 1
    FROM [StorageFile] AS [file]
    WHERE [file].[FolderId] = [folder].[FolderId])
```

And too many rows are returned because the FolderID is returned.  I wish to only have Roots.

If the where clause is removed, then the select ONLY returns the StorageRootId

This query:
``` C#
(from root in rootSet join folder in folderSet on root.StorageRootId equals folder.RootId
select root.StorageRootId).Distinct();
```

Produces this SQL:

``` SQL
SELECT DISTINCT [root].[StorageRootId]
FROM [StorageRoot] AS [root]
INNER JOIN [StorageFolder] AS [folder] ON [root].[StorageRootId] = [folder].[RootId]
```

You can see, the FolderId is not returned this time, and it shouldn't be since I explicitly select just 1 column

### The issue
The first query should not be returning the folder.FolderId column, and is causing an incorrect result set.

This only occurs when the where / Any clause is present.

```
Exception message: None
Stack trace: N/A
```

### Further technical details

Operating system: n/a

maumar:
Fixes #5630

Problem was that for correlated subquery when we visit the element that references the outer query, we add that element to the root projection. We need that potentially if the query ends up being client-evaluated. However, if client-eval is not needed, those additional projection elements can be thrown away. Currently they were not, resulting in unnecessary columns being pulled from the database, or in some cases (e.g. Distinct) could lead to incorrect results being returned.

Fix is to do a snapshot of projections before we start visiting the subquery and if it ends up being fully translated to SQL, we prune those additional projection elements.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -788,6 +788,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                         = (RelationalQueryModelVisitor)_queryModelVisitor.QueryCompilationContext
                             .CreateQueryModelVisitor(_queryModelVisitor);
 
+                    var queriesProjectionCountMapping = _queryModelVisitor.Queries
+                        .ToDictionary(k => k, s => s.Projection.Count);
+
                     queryModelVisitor.VisitSubQueryModel(subQueryModel);
 
                     if (queryModelVisitor.Queries.Count == 1
@@ -798,6 +801,11 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                         var selectExpression = queryModelVisitor.Queries.First();
 
                         selectExpression.Alias = string.Empty; // anonymous
+
+                        foreach (var mappingElement in queriesProjectionCountMapping)
+                        {
+                            mappingElement.Key.RemoveRangeFromProjection(mappingElement.Value);
+                        }
 
                         return selectExpression;
                     }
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -788,6 +788,9 @@ var queryModelVisitor
                         = (RelationalQueryModelVisitor)_queryModelVisitor.QueryCompilationContext
                             .CreateQueryModelVisitor(_queryModelVisitor);
 
+                    var queriesProjectionCountMapping = _queryModelVisitor.Queries
+                        .ToDictionary(k => k, s => s.Projection.Count);
```
anpete: OK, so we do need it for all queries?

anpete:
:shipit:

PR #5700, opened 2016-06-08
Query: Fix #5481 - RC2 Deadlock problem with CountAsync() that does n…

### LINKED ISSUE #5481: RC2 Deadlock problem with CountAsync() that does not happen with Count()
I have found an inconsistency when querying a `query.CountAsync();`. It does not throw any error nor displays any error on console, it simply gets into a deadlock state (the application is still up but the next line of code does not get to execute)

The problem does not happen when using the same query to count the items in a synchronous way: `query.Count();`

Scenario:

I have an entity that has 2 optional references to another entity
```
public class Shipment 
{
  public int Id { get; set; }
  public string CarrierName { get; set; }
  public int? CargoReportId { get; set; }
  public CargoReport CargoReport { get; set; }
  public int? AltCargoReportId { get; set; }
  public CargoReport AltCargoReport { get; set; }
}

public class Report
{
  public int Id { get; set; }
  public string Reference { get; set; }
}
```

Now I try to count the items in database with a simple query and **the synchronous call works fine but the asynchronous call gets the code onto a deadlock state**:
```
private IQueryable<Shipment> GetTestQuery(ShipmentDirection direction) {
    var shipmentsQuery = _context.Shipments
            .Include(s => s.CargoReport)
            .Include(s => s.AltCargoReport)
            .Where(s => s.CarrierName == "carrier"
                    || s.CargoReport.Reference.Contains("whatever") 
                    || s.AltCargoReport.Reference.Contains("whatever"));
    return shipmentsQuery;
}

public async Task<int> MyTestRepositoryCount()
{
  var count1 = await _context.Shipments.CountAsync(); //this works well
  var shipmentsQuery = GetTestQuery(direction);
  var count2 = shipmentsQuery.Count(); //this works well
  var count3= await shipmentsQuery.CountAsync(); //this DOES NOT work. No exception, just deadlock
  return count3; //this never executes
}
```

I know the problem has to do with _CargoReport_ and _AltCargoReport_ being a _null_ reference because If I had something like this:
```
private IQueryable<Shipment> GetTestQuery(ShipmentDirection direction) {
   var shipmentsQuery = _context.Shipments
            .Include(s => s.CargoReport)
            .Include(s => s.AltCargoReport)
            .Where(s => //s.CarrierName == "carrier" || //Noticed this condition commented out
                       s.CargoReport.Reference.Contains("whatever") 
                    || s.AltCargoReport.Reference.Contains("whatever"));
    return shipmentsQuery;
}

```
I would get the following exception when executing the `count2` line:

> Microsoft.EntityFrameworkCore.Storage.Internal.SqlServerConnection:Debug: Closing connection to database 'SpeediCargo' on server '(localdb)\mssqllocaldb'.
> Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler:Error: An exception occurred in the database while iterating the results of a query.
> System.NullReferenceException: Object reference not set to an instance of an object.
>    at lambda_method(Closure , TransparentIdentifier`2 )
>    at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
>    at System.Linq.Enumerable.Count[TSource](IEnumerable`1 source)
>    at lambda_method(Closure , QueryContext )
>    at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.<>c__DisplayClass19_1`1.<CompileQuery>b__1(QueryContext qc)
> 
> System.NullReferenceException: Object reference not set to an instance of an object.
>    at lambda_method(Closure , TransparentIdentifier`2 )
>    at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
>    at System.Linq.Enumerable.Count[TSource](IEnumerable`1 source)
>    at lambda_method(Closure , QueryContext )
>    at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.<>c__DisplayClass19_1`1.<CompileQuery>b__1(QueryContext qc)
> Exception thrown: 'System.NullReferenceException' in Microsoft.EntityFrameworkCore.dll
> Exception thrown: 'System.NullReferenceException' in mscorlib.ni.dll

So a few things here:

1. How should I have the conditions on CargoReport and AltCargoReport to handle null? **Shouldn't Entity Framework take care of that for me?**

1. Why the `.Count()` works well but the `.CountAsync()` doesn't? **shouldn't be both consistent?**

1. In case it's fine that the behaviour is not consistent. Shouldn't the `.CountAsync()` call at least **throw a null reference exception** if that's the case for the deadlock?

Thanks guys.

anpete:
…ot happen with Count()

- Replace IX-Async GroupJoin with async/await impl.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/AsyncQuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/AsyncQuerySqlServerTest.cs
@@ -11,7 +11,6 @@ using Xunit;
 using Xunit.Abstractions;
 
 #pragma warning disable 1998
-
 namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests
 {
     [MonoVersionCondition(Min = "4.2.0", SkipReason = "Async queries will not work on Mono < 4.2.0 due to differences in the IQueryable interface")]
--- a/src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
@@ -402,12 +402,128 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
         [UsedImplicitly]
         // ReSharper disable once InconsistentNaming
         private static IAsyncEnumerable<TResult> _GroupJoin<TOuter, TInner, TKey, TResult>(
-            IAsyncEnumerable<TOuter> outer,
-            IAsyncEnumerable<TInner> inner,
-            Func<TOuter, TKey> outerKeySelector,
-            Func<TInner, TKey> innerKeySelector,
-            Func<TOuter, IAsyncEnumerable<TInner>, TResult> resultSelector)
-            => outer.GroupJoin(inner, outerKeySelector, innerKeySelector, resultSelector);
+                IAsyncEnumerable<TOuter> outer,
+                IAsyncEnumerable<TInner> inner,
+                Func<TOuter, TKey> outerKeySelector,
+                Func<TInner, TKey> innerKeySelector,
+                Func<TOuter, IAsyncEnumerable<TInner>, TResult> resultSelector)
+            => new GroupJoinAsyncEnumerable<TOuter, TInner, TKey, TResult>(
+                outer, inner, outerKeySelector, innerKeySelector, resultSelector);
+
+        private sealed class GroupJoinAsyncEnumerable<TOuter, TInner, TKey, TResult> : IAsyncEnumerable<TResult>
+        {
+            private readonly IAsyncEnumerable<TOuter> _outer;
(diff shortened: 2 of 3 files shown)


anpete:
@maumar @smitpatel

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/AsyncQuerySqlServerTest.cs
```diff
@@ -4,14 +4,15 @@
 using System;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using Microsoft.EntityFrameworkCore.Specification.Tests;
```
smitpatel: Are these using needed?

anpete: Will clean up.

smitpatel:
:shipit:

PR #5701, opened 2016-06-08
Cache services in some places to avoid multiple calls to GetService.

ajcvickers:


anpete:
:shipit:

divega:
@ajcvickers @anpete is this for "1.0.1"?

ajcvickers:
@divega Depends on the bar for 1.0.1 per our discussion at lunch. If we decide to only take high priority fixes in patch releases, then I don't think this makes the cut. If we decide to take any non-feature change, then yes it would go in 1.0.1.

divega:
Thanks. The main purpose of my question was to make sure there were not things we wanted to take for 1.0.0. We can figure out where they go after (we have only defined a milestone for 1.0.1 but they might not belong there). It is good information that these would require breaking changes to internal surface.

ajcvickers:
@divega This one doesn't break internal APIs. The other one does.

PR #5703, opened 2016-06-09
Making tracking more pay-per-play in query

ajcvickers:
Lazily resolve services related to tracking from query such that they will not be loaded when only no-tracking queries are used.


anpete:
@ajcvickers Can we add a regression test? (reflection?)

divega:
@ajcvickers @anpete is this something we are planning to take in "1.0.1"?

ajcvickers:
@divega Probably have to wait for the first point release, per discussion on internal APIs at lunch.

ajcvickers:
@anpete Updated with test,

anpete:
:shipit:

PR #5709, opened 2016-06-09
Fix random Can_copy_to_array test failure

pakrym:
Test was failing in cases `CreateTestData` returned less then 10 elements causing array to be bigger then source collection because of `Math.Max`

@pranavkm @natemcmaster

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.Tests/ChangeTracking/ObservableHashSetTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.Tests/ChangeTracking/ObservableHashSetTest.cs
@@ -14,7 +14,9 @@ namespace Microsoft.EntityFrameworkCore.Tests.ChangeTracking
 {
     public class ObservableHashSetTest
     {
-        private static readonly Random _random = new Random();
+        // seed chosen by fair dice roll
+        // guaranteed to be random
+        private static readonly Random _random = new Random(4);
 
         [Fact]
         public void Can_construct()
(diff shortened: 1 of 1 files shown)


natemcmaster:
cc @ajcvickers

test/Microsoft.EntityFrameworkCore.Tests/ChangeTracking/ObservableHashSetTest.cs
```diff
@@ -159,11 +161,18 @@ public void Can_copy_to_array()
 
             Assert.Equal(orderedDistinct, arrray.Skip(100).Take(orderedDistinct.Count).OrderBy(i => i));
 
-            var toTake = Math.Max(10, orderedDistinct.Count);
+            var toTake = Math.Min(10, orderedDistinct.Count);
             arrray = new int[100 + toTake];
```
pranavkm: Is `array` a reserved keyword? If not, this seems like a typo.

test/Microsoft.EntityFrameworkCore.Tests/ChangeTracking/ObservableHashSetTest.cs
```diff
@@ -159,11 +161,18 @@ public void Can_copy_to_array()
...
             hashSet.CopyTo(arrray, 100, toTake);
 
-            Assert.Equal(orderedDistinct.Take(toTake), arrray.Skip(100).Take(toTake).OrderBy(i => i));
+            // there is no guaranteed order so just checking that all array elements are
+            // distinct and source collection contains them
+            var copied = arrray.Skip(100).Take(toTake);
+            Assert.Equal(copied.Count(), copied.Distinct().Count());
+            foreach (var element in copied)
```
pranavkm: Couldn't you sort both the arrays and do an `Assert.Equal`?

pakrym: No, I'm taking part of elements from HashSet, but it doesn't not guarantee which ones.

pranavkm: Ahh ok. Maybe use `Assert.Contains(element, orderedDistinct);` in that case. `Assert.True` wouldn't tell what was missing and what the source array was.

ajcvickers:
@pakrym Can you provide more information on the problem that is being fixed? Stack trace?

pakrym:
```
Assert.Equal() Failure
Expected: <TakeIterator>d__1<Int32> [-2109796193, -1896138343]
Actual:   OrderedEnumerable<Int32, Int32> [-2109796193, -1896138343, 0, 0, 0, ...]
   at Microsoft.EntityFrameworkCore.Tests.ChangeTracking.ObservableHashSetTest.Can_copy_to_array() in D:\b\work\3d42ba28d4e3c9\EntityFramework\test\Microsoft.EntityFrameworkCore.Tests\ChangeTracking\ObservableHashSetTest.cs:line 167
```

ajcvickers:
@pakrym Can you please disable the test and file a bug. I don't think that this is the right way to fix this, but I am OOF today and don't have time to work on it now.

bricelam:
**Triage:** Closing. We'll fix this with issue #5711

PR #5722, opened 2016-06-10
Use provider-specific truth literal in conditional

roji:
DefaultQuerySqlGenerator contains the following code for rendering conditionals:

```c#
if (expression.Test.IsSimpleExpression())
{
    _relationalCommandBuilder.Append(" = 1");
}
```

This doesn't work for PostgreSQL. We already have TrueLiteral and TypedTrueLiteral which are used elsewhere, this PR replaces the `1` with TrueLiteral.

However, it's not clear to me why a condition is necessary at all. In other words, instead of rendering `WHEN x = 1 THEN y ELSE z END`, why not simply render `WHEN x THEN y ELSE z END`? At least for PostgreSQL all the GearsOfWar tests (where a lot of conditionals happen) pass in this way.

As a workaround, I've overridden `VisitConditional` in Npgsql's QuerySqlGenerator.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -821,7 +821,8 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
 
                 if (expression.Test.IsSimpleExpression())
                 {
-                    _relationalCommandBuilder.Append(" = 1");
+                    _relationalCommandBuilder.Append(" = ");
+                    _relationalCommandBuilder.Append(TrueLiteral);
                 }
 
                 _relationalCommandBuilder.AppendLine();


divega:
> However, it's not clear to me why a condition is necessary at all. In other words, instead of rendering WHEN x = 1 THEN y ELSE z END, why not simply render WHEN x THEN y ELSE z END? 

Cool! That means in PostgreSQL Boolean values and search conditions are interchangeable. In other databases (in particular SQL Server family) we have to compensate when going both directions. It would be great if you could just override this for PostgreSQL.

roji:
@divega I don't really mind having `WHEN x = 1 THEN ...` if this is important for other providers. However, the current DefaultQuerySqlGenerator hardcodes the 1 rather than TrueLiteral, which might also be a problem for other providers...

smitpatel:
This whole statement block is no longer needed 😄 
With recent changes to convert bool to search conditions, we already convert simple expression into search condition so there is no longer any need to compensate anywhere in the code like this.

smitpatel:
@divega - Should I do PR to remove this compensations happening in other places? Also need to include converting parameters to search condition. Both complemented each other and did not raise error.
Further given that PostgreSQL doesn't require conversion and we have code for converting in Relational level, what should be the strategy for provider writers to opt out of it?

smitpatel:
The whole check is removed from `DefaultQuerySqlGenerator` in #5729

roji:
Great, thanks!

PR #5727, opened 2016-06-10
Use the startup project directory as the content root path

### LINKED ISSUE #5726: StartupInvoker fails when startup class contains AddUserSecrets
Repro: 
Create a new Web template that uses usersecrets
Run `dotnet ef dbcontext list`

Error:

> An error occurred while calling method 'ConfigureServices' on startup class 'WebApplication18.Startup'. Consider using IDbContextFactory to override the initialization of the DbContext at design-time. Error: Unable to locate a project.json at 'C:\Users\phenning\Documents\Visual Studio 2015\Projects\WebApplication18\src\WebApplication18\bin\Debug\netcoreapp1.0\'.

The issue is we are setting ContentRootPath to ApplicationBasePath, but those are not necessarily the same thing,

natemcmaster:
Fix #5726

cc @Eilon @phenning @bricelam

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Design/CommandLine/ExecuteCommand.cs
+++ b/src/Microsoft.EntityFrameworkCore.Design/CommandLine/ExecuteCommand.cs
@@ -16,7 +16,7 @@ namespace Microsoft.EntityFrameworkCore.Tools.Cli
         private const string StartupAssemblyOptionTemplate = "--startup-assembly";
         private const string DataDirectoryOptionTemplate = "--data-dir";
         private const string ProjectDirectoryOptionTemplate = "--project-dir";
-        private const string AppBaseDirectoryOptionTemplate = "--app-base-dir";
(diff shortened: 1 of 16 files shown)


src/Microsoft.EntityFrameworkCore.Design/CommandLine/ExecuteCommand.cs
```diff
@@ -111,7 +111,7 @@ private static void WriteLogo()
 
             ConsoleCommandLogger.Output("");
             ConsoleCommandLogger.Output(@"                     _/\__       ".Insert(21, Bold + White));
-            ConsoleCommandLogger.Output(@"               ---==/    \\      ");
+            ConsoleCommandLogger.Output(@"               ---==/    \\      ".Insert(20, Bold + White));
```
natemcmaster: :art: :smile:

natemcmaster:
cc @divega we want to get this one in for 1.0.0

divega:
:shipit:

Issue #5715, opened 2016-06-10
Sql Server datetime Data Type Mapped to datetime2 rather than datetime


WilliamBerryiii:
Given the following fluent mapping:
```entity.Property(e => e.Date)
                    .HasColumnName("date")
                    .HasColumnType("datetime"); ```

and a query such as:
```
var date = DateTime.UtcNow;                
var d = _context.EntityType.AsNoTracking().FirstAsync(entity => entity.Date == date);
```

The framework elects to map the query parameter `date` to a `datetime2` data type despite the specific column type mapping.  This naturally results in the exception:
```
Exception message: Microsoft.EntityFrameworkCore.Query.Internal.SqlServerQueryCompilationContextFactory:Error: An exception occurred in the database while iterating the results of a query.

Stack trace: 
System.ArgumentException: The version of SQL Server in use does not support datatype 'datetime2'.
   at System.Data.SqlClient.SqlCommand.<>c.<ExecuteDbDataReaderAsync>b__110_0(Task`1 result)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
```

Would it be possible to either correct this mapping to push explicitly declared `datetime` mapping type to the Sql Server `datetime` type here:

[{ "datetime", _datetime2 },](https://github.com/aspnet/EntityFramework/blob/14e16902c6558cd23e43826d388219ed3b9e4f97/src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerTypeMapper.cs#L111)

or conversely provide some explicit mapping option to override the default `datetime2` mapping?

EF Core version: 1.0.0-rc2-final
Operating system: Windows 
Visual Studio version:VS2105 Update 2

ErikEJ:
SQL Server 2005 is not supported by EF Core

WilliamBerryiii:
@ErikEJ Understood.  That said the data type `datetime` still exists and is fully supported in SQL Server 2008.  EF appears to have taken the stance that even though it knows that I want a `datetime` it's going to give me a `datetime2` anyway.  Perhaps this is simply an implementation detail in SQL Server 2008? 
[...]

PR #5728, opened 2016-06-10
Correctly map SQL Server datetime columns

ajcvickers:
Issue #5715

SQL Server was mapping "datetime" to DbType.DataTime2 parameters. This breaks SQL Server 2005 and also the always encrypted feature of later versions where the automatic conversions from datetime2 don't work.


divega:
:shipit:

PR #5729, opened 2016-06-10
Query: Convert Parameters of bool value to search condition in predicate

smitpatel:
resolves & supersedes #5722 
@anpete @divega

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -104,8 +104,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
         protected virtual IRelationalCommandBuilder Sql => _relationalCommandBuilder;
 
         protected virtual string ConcatOperator => "+";
-        protected virtual string TrueLiteral => "1";
-        protected virtual string FalseLiteral => "0";
         protected virtual string TypedTrueLiteral => "CAST(1 AS BIT)";
         protected virtual string TypedFalseLiteral => "CAST(0 AS BIT)";
 
@@ -819,11 +817,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
 
                 Visit(expression.Test);
 
-                if (expression.Test.IsSimpleExpression())
-                {
-                    _relationalCommandBuilder.Append(" = 1");
-                }
-
                 _relationalCommandBuilder.AppendLine();
                 _relationalCommandBuilder.Append("THEN ");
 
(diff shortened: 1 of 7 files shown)


anpete:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -104,8 +104,6 @@ public virtual IRelationalCommand GenerateSql(IReadOnlyDictionary<string, object
         protected virtual IRelationalCommandBuilder Sql => _relationalCommandBuilder;
 
         protected virtual string ConcatOperator => "+";
-        protected virtual string TrueLiteral => "1";
```
divega: So @roji has to override this to return `TRUE` for PostgreSQL. What does he need to do instead?

smitpatel: In current code,
There are 2 cases where we add `= 1` (or `"= " + TrueLiteral`)
1. If the test in conditional is SimpleExpression
2. In logical binary expression (AND/OR) either side is SimpleExpression.
The `SearchConditionTranslatingVisitor` makes sure that those cases are already converted to search conditions before they are converted to SQL.  (`ParameterExpression` was missing in that which is added in this PR, trying to find if anything else is missing.)
Therefore there won't be case arising when we need to compensate.
previously there was `expression` of value type, after the visitor it is converted to `BinaryExpression` with right node being `True` valued `ConstantExpression`.
How the value of `True` gets translated is governed by `RelationalSqlGenerationHelper` which providers can override to produce something other than `1`. In case of PostgreSQL, that returns `True`.
Hence there is no longer need of having `TrueLiteral` & `FalseLiteral` here. There will be only single place providers needs to specify what is the value of `TrueLiteral`

divega: Excellent! Thanks!

divega:
@smitpatel Travis CI broken after this?

smitpatel:
@divega - EF build at present, while restoring packages, is producing a lot of package downgrade warnings. Due to all those downgrade warnings the log size of 4MB reached on Travis CI, at which point it concludes that a test is failing on same exception producing same output stacktrace and terminates the job. The package restore didn't finish on Travis CI at all.
I am not able to reach efci atm, but last I checked both windows & Linux platform passed after this change.
On aspnetci, coherence-signed passed with this change and also Ubuntu-Universe passed. On OS X, building EF step also passed. 😄

divega:
Thanks @smitpatel. We can check what's up with the package downgrades later.

Issue #5725, opened 2016-06-10
D.I. constructors should be documented as internal


ajcvickers:
We have constructors on public types which are there for use by dependency injection and therefore take our injected internal types. We should consider documenting these as internal because application code should never need to call them and we will need to change them as service dependencies change.

PR #5730, opened 2016-06-10
Document constructors that take internal types as internal

ajcvickers:
Issue #5725

These typically would have been internal constructors in the old world. They are now public, often for D.I. or for fluent API builders. When we break internal stuff these as conceptually internal members may also get broken, so document as such.


anpete:
:shipit:

Issue #5737, opened 2016-06-12
Expose Current ConventionSet


rjperes:
As of now, the *current* ConventionSet is not publicly exposed. It is only possible, AFAIK, to get a pointer to it by using reflection to access the _conventionSet field of the ConventionDispatcher.
In order to make conventions simpler to use, can't you expose the current ConventionSet as a property somewhere?

AndriySvyryd:
@rjperes The `ConvetionSet` is not exposed because `ModelBuilder` assumes that it is readonly. The correct way to change the conventions that will be used is by deriving from the provider-specific implementation of `IConventionSetBuilder` and adding it to DI, overriding the default one. Since each provider can have a different set of conventions you would need to do it for each provider that will be used in the application.

As an example if you want to replace the `RelationshipDiscoveryConvention` that runs when an entity type is added for SqlServer, you would implement something like:
```C#

public class CustomSqlServerConventionSetBuilder : SqlServerConventionSetBuilder
{
    public CustomSqlServerConventionSetBuilder(
        [NotNull] IRelationalTypeMapper typeMapper,
        [CanBeNull] ICurrentDbContext currentContext,
        [CanBeNull] IDbSetFinder setFinder)
        : base(typeMapper, currentContext, setFinder)
    {
    }

    public override ConventionSet AddConventions(ConventionSet conventionSet)
    {
        base.AddConventions(conventionSet);

        conventionSet.ModelInitializedConventions.Add(new SqlServerValueGenerationStrategyConvention());
            
        var relationshipDiscoveryConvention =
            conventionSet.EntityTypeAddedConventions.OfType<RelationshipDiscoveryConvention>().FirstOrDefault();

        var index = conventionSet.EntityTypeAddedConventions.IndexOf(relationshipDiscoveryConvention);
        conventionSet.EntityTypeAddedConventions.RemoveAt(index);
        conventionSet.EntityTypeAddedConventions.Insert(index, new BetterRelationshipDiscoveryConvention());

        return conventionSet;
    }
}
```

Then add it to DI:
```C#
services
.AddEntityFrameworkSqlServer()
.AddScoped<SqlServerConventionSetBuilder, CustomSqlServerConventionSetBuilder>()
```

This experience will be improved by https://github.com/aspnet/EntityFramework/issues/214

rjperes:
[...]

PR #5741, opened 2016-06-13
#5737 Expose Current ConventionSet

rjperes:
Simple change, just expose the private ConventionSet so that we can add to the current conventions.
Please see https://github.com/aspnet/EntityFramework/issues/5737 and https://weblogs.asp.net/ricardoperes/implementing-missing-features-in-entity-framework-core-part-4-conventions for a description of the problem.


rjperes:
Weird... The only change was the addition of a property... if the tests are failing, then they are already failing in the source branch, from which I forked!

smitpatel:
@rjperes - Tests are failing because the property which is added needs to be virtual.

AndriySvyryd:
@rjperes Thanks for your contribution. Unfortunately it goes against the design we are pursuing. Please see https://github.com/aspnet/EntityFramework/issues/5737#issuecomment-225724598

Issue #5734, opened 2016-06-11
After second SaveChanges() EF loses added new dependents


AlexBilonog:
After second SaveChanges() EF loses added new dependents:
```
using (var db = new TheContext())
{
   var user = db.Set<User>().AsQueryable().First();
   var userDepartment = new UserDepartment { DepartmentID = 1 };
   user.UserDepartments.Add(userDepartment);
   db.SaveChanges(); // After this line: user.UserDepartments: Count = 1 (OK)
   db.SaveChanges(); // After this line: user.UserDepartments: Count = 0 (Wrong)
   db.SaveChanges(); // By executing this line exception: The association between entity types 'User' and 'UserDepartment' has been severed but the foreign key for this relationship cannot be set to null
}
```

EF Core version: 1.0.0-rc3-21338
Operating system: Windows 10
Visual Studio version: VS 2015 last updates
StackTrace:
```
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.HandleConceptualNulls()
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.GetEntriesToSave()
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChanges()
   at EFGetStarted.ConsoleApp.Program.Main() in C:\Work\Console4\EFGetStarted.ConsoleApp\Program.cs:line 21
   at System.AppDomain._nExecuteAssembly(RuntimeAssembly assembly, String[] args)
   at System.AppDomain.ExecuteAssembly(String assemblyFile, Evidence assemblySecurity, String[] args)
   at Microsoft.VisualStudio.HostingProcess.HostProc.RunUsersAssembly()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[...]

PR #5745, opened 2016-06-13
Update relationship snapshot when recording dangling entities

ajcvickers:
Issue #5734

When we encounter an untracked entity in a navigation we save it for later fixup, but we were not adding these entities to the snapshot for the navigation. This later caused change detection to be confused, which resulted in the entities being removed from the navigation.


divega:
:shipit:

PR #5749, opened 2016-06-14
More doc comments for commonly used APIs

rowanmiller:
This gets doc comments onto the bulk of APIs that would be used in most applications.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.InMemory/Extensions/InMemoryServiceCollectionExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.InMemory/Extensions/InMemoryServiceCollectionExtensions.cs
@@ -15,8 +15,40 @@ using Microsoft.Extensions.DependencyInjection.Extensions;
 // ReSharper disable once CheckNamespace
 namespace Microsoft.Extensions.DependencyInjection
 {
+    /// <summary>
+    ///     In-memory specific extension methods for <see cref="IServiceCollection"/>.
+    /// </summary>
     public static class InMemoryServiceCollectionExtensions
     {
+        /// <summary>
+        ///     <para>
+        ///         Adds the services required by the in-memory database provider for Entity Framework
+        ///         to an <see cref="IServiceCollection" />. You use this method when using dependency injection
+        ///         in your application, such as with ASP.NET. For more information on setting up dependency
+        ///         injection, see http://go.microsoft.com/fwlink/?LinkId=526890.
+        ///     </para>
+        ///     <para>
+        ///         You only need to use this functionality when you want Entity Framework to resolve the services it uses
+        ///         from an external dependency injection container. If you are not using an external
+        ///         dependency injection container, Entity Framework will take care of creating the services it requires.
+        ///     </para>
+        /// </summary>
+        /// <example>
+        ///     <code>
+        ///         public void ConfigureServices(IServiceCollection services)
+        ///         {
+        ///             services
+        ///                 .AddEntityFrameworkInMemoryDatabase()
+        ///                 .AddDbContext&lt;MyContext&gt;((serviceProvider, options) =>
+        ///                     options.UseInMemoryDatabase()
+        ///                            .UseInternalServiceProvider(serviceProvider));
+        ///         }
+        ///     </code>
+        /// </example>
+        /// <param name="services"> The <see cref="IServiceCollection" /> to add services to. </param>
+        /// <returns>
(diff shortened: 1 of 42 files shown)


src/Microsoft.EntityFrameworkCore.InMemory/Extensions/InMemoryServiceCollectionExtensions.cs
```diff
@@ -15,8 +15,40 @@
...
+        /// <summary>
+        ///     <para>
+        ///         Adds the services required by the in-memory database provider for Entity Framework
+        ///         to an <see cref="IServiceCollection" />. You use this method when using dependency injection
+        ///         in your application, such as with ASP.NET. For more information on setting up dependency
+        ///         injection, see http://go.microsoft.com/fwlink/?LinkId=526890.
+        ///     </para>
+        ///     <para>
+        ///         You only need to use this functionality when you want Entity Framework to resolve the services it uses
+        ///         from an external <see cref="IServiceCollection" />. If you are not using an external
```
divega: Not sure saying we resolve from an external `IServiceCollection` is clear. I would use `IServiceProvider` here and in the next instance in the same paragraph as well.

divega: or maybe just "dependency injection container".

rowanmiller: > or maybe just "dependency injection container".

👍

src/Microsoft.EntityFrameworkCore.InMemory/Infrastructure/InMemoryEventId.cs
```diff
@@ -1,11 +1,23 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Extensions.Logging;
+
 namespace Microsoft.EntityFrameworkCore.Infrastructure
 {
+    /// <summary>
+    ///     Values that are used as the eventId when logging messages from this assembly via <see cref="ILogger" />.
```
divega: Technically the fact that are messages from this "assembly" my matter and be correct, but it seems a bit obscure for the user. Can we use something else? E.g. "provider"?

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/RelationalEventId.cs
```diff
@@ -1,19 +1,64 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Extensions.Logging;
+
 namespace Microsoft.EntityFrameworkCore.Infrastructure
 {
+    /// <summary>
+    ///     Values that are used as the eventId when logging messages from this assembly via <see cref="ILogger" />.
```
divega: Idem re "assembly" maybe from "Entity Framework Core" this time?

src/Microsoft.EntityFrameworkCore.Sqlite/Infrastructure/SqliteServiceCollectionExtensions.cs
```diff
@@ -20,8 +20,42 @@
...
+        /// <summary>
+        ///     <para>
+        ///         Adds the services required by the SQLite database provider for Entity Framework
+        ///         to an <see cref="IServiceCollection" />. You use this method when using dependency injection
+        ///         in your application, such as with ASP.NET. For more information on setting up dependency
+        ///         injection, see http://go.microsoft.com/fwlink/?LinkId=526890.
+        ///     </para>
+        ///     <para>
+        ///         You only need to use this functionality when you want Entity Framework to resolve the services it uses
+        ///         from an external <see cref="IServiceCollection" />. If you are not using an external
```
divega: Same comment as before re service collection vs. provider/container. Applies to both instances in this paragraph.

divega:
I made a few suggestions but it looks very good. 
:shipit:

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/RelationalEventId.cs
rowanmiller: I went with `...when logging messages from a relational database provider`

PR #5756, opened 2016-06-14
Use KOREBUILD_NUGET_EXE if available

pranavkm:
Temporary workaround to use a build of nightly NuGet.exe to repack packages with `serviceable` attribute.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -47,7 +47,7 @@ k-standard-goals
       
       Log.Info("Repacking Nupkg: " + projectNupkg);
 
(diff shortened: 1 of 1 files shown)


pranavkm:
cc @natemcmaster

pranavkm:
They have a nuget.exe 3.5.0-beta on nuget.org. Might be able to use that in our builds instead.

pranavkm:
Nevermind. It's beta and we need beta2 which is when the feature was added.

natemcmaster:
:shipit:

PR #5758, opened 2016-06-14
Rename Microsoft.EntityFrameworkCore{.Tools.Core} => {.Design.Core}

### LINKED ISSUE #5746: Microsoft.EntityFrameworkCore.Tools.Core version is "1.0.0-rc3-*"
@rowanmiller just found that this package is not versioned as "1.0.0-preview2-*" although it is part of the tooling. 

@natemcmaster @bricelam is this intentional for a reason we are missing?

BTW, we actually already shipped the package with that name as "1.0.0-rc2-final" so just changing the version to "preview2" would be going back in time.

natemcmaster:
Fix #5746 

cc @Eilon @rowanmiller @bricelam @divega


divega:
LGTM but would be good to get another pair of eyes on it.

bricelam:
:shipit:

src/Microsoft.EntityFrameworkCore.Design.Core/project.json
```diff
@@ -1,6 +1,6 @@
 {
-  "version": "1.0.0-rc3-*",
-  "description": "Shared components for Entity Framework Core tools.",
+  "version": "1.0.0-preview2-*",
+  "description": "Internal implementations for Entity Framework Core design-time tools.",
```
natemcmaster: @rowanmiller is this wording okay?

rowanmiller: This would be more consistent with the wording used in the other packages...

> Shared design-time components for Entity Framework Core tools.

Issue #5771, opened 2016-06-14
Allow navigation properties to be typed as IEnumerable so long as backed by ICollection


ajcvickers:
Currently runtime delegate creation enforces that the type of the navigation property on the entity is an ICollection, even though our APIs are designed to work with IEnumerable properties.

Issue #1481, opened 2015-01-27
Query: Include and ThenInclude should support collection navigation properties that are just IEnumerable<T>


divega:
We learned yesterday while discussing #1463 with @anpete and @ajcvickers that the EF7 runtime does not constrain collection navigation properties to be of types that can be assigned to `ICollection<T>` but that `IEnumerable<T>` is generally sufficient. 

The eager loading API should relax to support the same navigation property types.

divega:
`HasMany()` already supports `IEnumerable<TRelatedEntity>` so this seems to be the only issue blocking having navigation properties that are statically of that type. 

@ajcvickers @rowanmiller @anpete note that this is currently pri1 for RTM but I wonder if it is just a matter of adding some tests and changing the signature of [this overload of `ThenInclude()`](https://github.com/aspnet/EntityFramework/blob/17fa7e74e4f0ecbc329fc0fa40671d7f821d31e1/src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs#L2201) from:

``` C#
public static IIncludableQueryable<TEntity, TProperty> 
    ThenInclude<TEntity, TPreviousProperty, TProperty>( 
        [NotNull] this IIncludableQueryable<TEntity, ICollection<TPreviousProperty>> source, 
        [NotNull] Expression<Func<TPreviousProperty, TProperty>> navigationPropertyPath) 
```
To
``` C#
public static IIncludableQueryable<TEntity, TProperty> 
    ThenInclude<TEntity, TPreviousProperty, TProperty>( 
        [NotNull] this IIncludableQueryable<TEntity, IEnumerable<TPreviousProperty>> source, 
        [NotNull] Expression<Func<TPreviousProperty, TProperty>> navigationPropertyPath) 
```

cbaxter:
@divega Based on the current state of `dev` branch, making the proposed change seems to work as you expected with no _obvious_ side-effects. 

I modified the [overload of ThenInclude](https://github.com/aspnet/EntityFramework/blob/17fa7e74e4f0ecbc329fc0fa40671d7f821d31e1/src/EntityFramework.Core/Extensions/EntityFrameworkQueryableExtensions.cs#L2201) in a local build and verified the existing tests pass via the official build script (although Azure tests skipped in my environment). I also ran through several `ThenInclude` examples in my test project and everything seems to be working like a charm. 

[...]

PR #5772, opened 2016-06-14
Allow navigation properties to be typed as IEnumerable so long as backed by ICollection

ajcvickers:
Issues #5771 and #1481


anpete:
:shipit:

Issue #797, opened 2014-10-06
Add Find method on DbSet


kirthik:
In the old world, we have Find method on Dbset that is missing in the new world. It would be nice to have this in new EF world too.

bricelam:
Note for triage: I'm marking this for re-triage since I'm seeing more and more people asking about this and we haven't done any work on the mitigation we once talked about in the early days of EF7.

jasoncavett:
Is FindAsync also in the consideration here as well or is that a separate issue?

divega:
@jasoncavett You can consider `FindAsync()` part of this issue.

rowanmiller:
I've seen some good discussion and feedback on this on the ASP.NET Insiders mailing list. We should re-discuss and maybe pull it into RTM.

tidusjar:
I would also like to add that I would like to see `FindAsync()` and `Find()` in EF7

sirpadington:
I am also going to add that it would be nice to have Find() in EF7

bricelam:
P.S. I have [an answer on StackOverflow](http://stackoverflow.com/a/29082410/475031) showing how you could write it yourself.

sirpadington:
That is awesome.  I am having a hard time trying to find where IAccessor is located. Any ideas?

bricelam:
@sirpadington In namespace `Microsoft.Data.Entity.Infrastructure`.

joeleesh:
@bricelam and @sirpadington do you find reference for IAccessor? I can not find on Microsoft.Data.Entity.Infrastructure?

bricelam:
It was renamed to `IInfrastructure<>`.

komengem:
What is the progress on this "Add Find method on DbSet #797"?

VSG24:
God we need this so much.

komengem:
Man, i was hoping we would get this in Update 2 :-1:

ErikEJ:
@komengem did you have a look at: http://stackoverflow.com/questions/29030472/dbset-doesnt-have-a-find-method-in-ef7/29082410#29082410 ?

sjb-sjb:
[...]

PR #5800, opened 2016-06-16
Implement DbSet.Find

ajcvickers:
Issue #797

Semantics are essentially the same as EF6:
- If entity with given key is being tracked, then return it without a database query
- Otherwise, query the database and track and return entity if found
- Otherwise, return null

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/CompositePrincipalKeyValueFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/CompositePrincipalKeyValueFactory.cs
@@ -40,6 +40,13 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        public virtual object CreateFromKeyValues(object[] keyValues)
+            => keyValues.Any(v => v == null) ? null : keyValues;
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
         public virtual object CreateFromBuffer(ValueBuffer valueBuffer)
         {
             var values = new object[_properties.Count];
--- a/src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
+++ b/src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
@@ -6,16 +6,21 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.ChangeTracking;
+using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Query.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 
 namespace Microsoft.EntityFrameworkCore.Internal
 {
     /// <summary>
-    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
     ///     directly from your code. This API may change or be removed in future releases.
     /// </summary>
     public class InternalDbSet<TEntity>
@@ -23,10 +28,11 @@ namespace Microsoft.EntityFrameworkCore.Internal
         where TEntity : class
     {
         private readonly DbContext _context;
+        private IStateManager _stateManager;
         private readonly LazyRef<EntityQueryable<TEntity>> _entityQueryable;
 
         /// <summary>
-        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public InternalDbSet([NotNull] DbContext context)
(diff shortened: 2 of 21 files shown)


test/Microsoft.EntityFrameworkCore.Sqlite.FunctionalTests/FindSqliteTest.cs
```diff
@@ -0,0 +1,312 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
anpete: Don't verify the SQL here unless there is a meaningful difference from SQL Server. We removed this type of verification because it added a significant maintenance burden to query.

src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/CompositePrincipalKeyValueFactory.cs
```diff
@@ -40,6 +40,13 @@ public CompositePrincipalKeyValueFactory([NotNull] IKey key)
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        public virtual object CreateFromKeyValues(object[] keyValues)
```
anpete: Did you consider avoiding these breaks by just using a ValueBuffer (possibly with these as extension methods instead)?

src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
```diff
@@ -43,101 +49,205 @@ public InternalDbSet([NotNull] DbContext context)
...
+                        CoreStrings.FindValueTypeMismatch(i, typeof(TEntity).Name, valueType.Name, propertyType.Name));
+                }
+            }
+
+            return StateManager.TryGetEntry(key, keyValues)?.Entity as TEntity;
+        }
+
+        private IStateManager StateManager => _stateManager ?? (_stateManager = _context.GetService<IStateManager>());
+
+        private static readonly MethodInfo _efPropertyMethod
```
anpete: This is already cached on EntityQueryModelVisitor. Can we unify?

src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
```diff
@@ -43,101 +49,205 @@ public InternalDbSet([NotNull] DbContext context)
...
+        {
+            var entityparameter = Expression.Parameter(typeof(TEntity), "e");
+
+            BinaryExpression predicate = null;
+            for (var i = 0; i < keyValues.Length; i++)
+            {
+                var property = keyProperties[i];
+                var equals =
+                    Expression.Equal(
+                        property.IsShadowProperty
```
anpete: We typically just use EF.Property for nominal properties, too.

src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
```diff
@@ -43,101 +49,205 @@ public InternalDbSet([NotNull] DbContext context)
...
+                                entityparameter, property.Name),
+                        Expression.Constant(
+                            keyValues[i], property.ClrType));
+
+                predicate = predicate == null ? equals : Expression.AndAlso(predicate, equals);
+            }
 
-            _context = ((InternalDbSet<TEntity>)dbSet)._context;
-            _entityQueryable = new LazyRef<EntityQueryable<TEntity>>(() => (EntityQueryable<TEntity>)source);
+            return Expression.Lambda<Func<TEntity, bool>>(predicate, entityparameter);
```
anpete: Cache this in an annotation on IKey?

test/Microsoft.EntityFrameworkCore.Sqlite.FunctionalTests/FindSqliteTest.cs
ajcvickers: It's also a way of ensuring that the database has/has not been hit. Okay if I change it to empty string verses not empty string?

src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/CompositePrincipalKeyValueFactory.cs
ajcvickers: Positions of key values in the value buffer are not necessarily the same as those in the key values array, so it would require translation to a new array, which might also have to have empty spaces for padding. This is not going to go into a patch release, so the internal API breaks are fine.

src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
ajcvickers: We can, but it would have to be made public and a new dependency introduced. Decided it was cleaner to not have the dependency.

src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
ajcvickers: Why?

src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
ajcvickers: Considered this, but would like to test perf difference before doing it.

src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
```diff
@@ -43,101 +49,205 @@ public InternalDbSet([NotNull] DbContext context)
...
+        }
+
+        private IStateManager StateManager => _stateManager ?? (_stateManager = _context.GetService<IStateManager>());
+
+        private static readonly MethodInfo _efPropertyMethod
+            = typeof(EF).GetTypeInfo().GetDeclaredMethod(nameof(EF.Property));
+
+        private static Expression<Func<TEntity, bool>> BuildPredicate(IReadOnlyList<IProperty> keyProperties, object[] keyValues)
+        {
+            var entityparameter = Expression.Parameter(typeof(TEntity), "e");
```
anpete: Casing.

src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
anpete: Just simpler when building the expression.

test/Microsoft.EntityFrameworkCore.Sqlite.FunctionalTests/FindSqliteTest.cs
anpete: Aren't you already getting that verification by using Assert.Same?

anpete:
:shipit:

PR #5804, opened 2016-06-17
Clean out test code that handled testing on Mono

natemcmaster:
We don't run tests on Mono anymore so we don't need this code.


ajcvickers:
:shipit:

Issue #5057, opened 2016-04-13
Add QueryTracking behavior to DbContextOptionsBuilder


anpete:
So it can be configured from Startup.cs

PR #5824, opened 2016-06-21
Add QueryTracking behavior to DbContextOptionsBuilder

ajcvickers:
Allows one time configuration in places like Startup.cs.

Issue #5057


src/Microsoft.EntityFrameworkCore/ChangeTracking/ChangeTracker.cs
```diff
@@ -22,16 +23,23 @@ public class ChangeTracker : IInfrastructure<IStateManager>
...
         {
             Check.NotNull(context, nameof(context));
 
             Context = context;
+            _queryTrackingBehavior = context
+                .GetService<IDbContextOptions>()
+                .Extensions
+                .OfType<CoreOptionsExtension>()
+                .FirstOrDefault()
+                .QueryTrackingBehavior;
```
anpete: Potential null ref?

ajcvickers: Will fix.

anpete:
:shipit:

PR #5829, opened 2016-06-21
Fix tests that break due to line-ending differences

natemcmaster:
Fixing the \r\n and \n woes for Linux build/test.
Fix minor bugs in build script.


AndriySvyryd:
:shipit:

Issue #5688, opened 2016-06-07
Avoid doing fixup work twice


ajcvickers:
Because of issue #5674 we started doing state manager fixup in more places than we were. This means that a lot of cases where query is doing all the fixup and the state manager would have nothing to do. To fix this, query should somehow communicate to the state manager which fixup it is going to do, and the state manager can then determine if it needs to do additional fixup or not.

PR #5849, opened 2016-06-24
Query tells state manager which relationships have been fixed up

ajcvickers:
Issue #5688

Query code will take care of fixing up entities within the query when Include is used. This means that the state manager does not need to do fixup for these relationships in cases where there are no entities tracked other than those that have come from query. This change allows query to communicate this to the state manager so that it can avoid unneeded work. The change also preserves the very simple case of a single query for a single entity type with no self-refs, in which case the state manager does not need to do work at all.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
@@ -180,18 +180,28 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
                 StateManager.StopTracking(this);
             }
 
-            StateManager.Notify.StateChanged(this, oldState, StateManager.IsSingleQueryMode(EntityType), fromQuery: false);
+            StateManager.Notify.StateChanged(this, oldState, fromQuery: false);
         }
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual void MarkUnchangedFromQuery()
+        public virtual void MarkUnchangedFromQuery([CanBeNull] ISet<IForeignKey> handledForeignKeys)
         {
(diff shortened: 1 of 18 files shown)


src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -180,18 +180,28 @@ private void SetEntityState(EntityState oldState, EntityState newState, bool acc
...
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual void MarkUnchangedFromQuery()
+        public virtual void MarkUnchangedFromQuery([CanBeNull] ISet<IForeignKey> handledForeignKeys)
         {
             StateManager.Notify.StateChanging(this, EntityState.Unchanged);
             _stateData.EntityState = EntityState.Unchanged;
-            StateManager.Notify.StateChanged(this, EntityState.Detached, StateManager.IsSingleQueryMode(EntityType), fromQuery: true);
+            StateManager.Notify.StateChanged(this, EntityState.Detached, fromQuery: true);
```
IDisposable: I find it confusing that the `StateChanging` is saying `EntityState.Unchanged` but the `StateChanged` call is saying **`EntityState.Detached`**.  This looks like a typo, but I haven't read this code base in a long time.

I'm used to event notification patterns that have the same value set in the _ing_ and _ed_ calls. Is this intentionally **Detached** not **Unchanged**?

ajcvickers: If you look at the API you will see that StateChanging contains the state that the entity will change to, while StateChanged contains the state that the entity was previously in. The current state is always available from looking at the entry.

IDisposable: Gotcha, thanks

anpete:
:shipit:

Issue #4931, opened 2016-03-30
Port EF6 Data Annotations tests


AndriySvyryd:
We want to know whether any semantics of data annotations has changed.

ajcvickers:
@AndriySvyryd Poaching this.

ajcvickers:
Assigning back to @AndriySvyryd since he has some additional tests.

PR #5875, opened 2016-06-27
Port data annotation tests from EF6

ajcvickers:
Issue #4931

Turns out we didn't have that many tests, but I ported all the relevant ones from DataAnnotationScenariosTest.


anpete:
:shipit:

Issue #5765, opened 2016-06-14
Potential Globalization Bugs


bricelam:
I'm running some static analysis on our codebase, and it looks like we may have some potential globalization bugs:

    src\Microsoft.EntityFrameworkCore.Design.Core\Design\MigrationsOperations.cs(95): warning CA1307: Microsoft.Globalization : Because the behavior of 'string.StartsWith(string)' could vary based on the current user's locale settings, replace this call in 'MigrationsOperations.SubnamespaceFromOutputPath(string)' with a call to 'string.StartsWith(string, StringComparison)'. If the result of 'string.StartsWith(string, StringComparison)' will be displayed to the user, such as when sorting a list of items for display in a list box, specify 'StringComparison.CurrentCulture' or 'StringComparison.CurrentCultureIgnoreCase' as the 'StringComparison' parameter. If comparing case-insensitive identifiers, such as file paths, environment variables, or registry keys and values, specify 'StringComparison.OrdinalIgnoreCase'. Otherwise, if comparing case-sensitive identifiers, specify 'StringComparison.Ordinal'.
    src\Microsoft.EntityFrameworkCore.Design.Core\Migrations\Design\CSharpHelper.cs(306): warning CA1305: Microsoft.Globalization : Because the behavior of 'string.Format(string, params object[])' could vary based on the current user's locale settings, replace this call in 'CSharpHelper.Literal(DateTime)' with a call to 'string.Format(IFormatProvider, string, params object[])'. If the result of 'string.Format(IFormatProvider, string, params object[])' will be displayed to the user, specify 'CultureInfo.CurrentCulture' as the 'IFormatProvider' parameter. Otherwise, if the result will be stored and accessed by software, such as when it is persisted to disk or to a database, specify 'CultureInfo.InvariantCulture'.
    src\Microsoft.EntityFrameworkCore.Design.Core\Migrations\Design\CSharpHelper.cs(316): warning CA1305: Microsoft.Globalization : Because the behavior of 'int.ToString()' could vary based on the current user's locale settings, replace this call in 'CSharpHelper.Literal(int)' with a call to 'int.ToString(IFormatProvider)'. If the result of 'int.ToString(IFormatProvider)' will be displayed to the user, specify 'CultureInfo.CurrentCulture' as the 'IFormatProvider' parameter. Otherwise, if the result will be stored and accessed by software, such as when it is persisted to disk or to a database, specify 'CultureInfo.InvariantCulture'.
[...]

PR #5879, opened 2016-06-27
Fix globalization issues

ajcvickers:
Issue #5765

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Design.Core/Properties/DesignCoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore.Design.Core/Properties/DesignCoreStrings.resx
@@ -255,4 +255,7 @@ Change your target project to the migrations project by using the Package Manage
   <data name="InvokeStartupMethodFailed" xml:space="preserve">
     <value>An error occurred while calling method '{method}' on startup class '{startupClass}'. Consider using IDbContextFactory to override the initialization of the DbContext at design-time. Error: {error}</value>
   </data>
+  <data name="CouldNotSerialize " xml:space="preserve">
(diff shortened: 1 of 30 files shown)


lajones:
Looks good. Rather than just the comment `// Interpolation OK...`, is there any better way of indicating that the code as written should be ignored as far as the globalization checks are concerned? An annotation perhaps (depends on what tool we run to detect the problem in the first place I guess)? If not then :shipit:

ajcvickers:
@lajones I'll leave it to @bricelam to comment on what he is using and whether there is a way to suppress warnings.

bricelam:
`#pragma disable` or `[SuppressMessage]` will work, but I don't think it's important until we actually automate the checks.

src/Microsoft.EntityFrameworkCore.Design.Core/Properties/DesignCoreStrings.resx
```diff
@@ -255,4 +255,7 @@ Change your target project to the migrations project by using the Package Manage
   <data name="InvokeStartupMethodFailed" xml:space="preserve">
     <value>An error occurred while calling method '{method}' on startup class '{startupClass}'. Consider using IDbContextFactory to override the initialization of the DbContext at design-time. Error: {error}</value>
   </data>
+  <data name="CouldNotSerialize " xml:space="preserve">
```
Eilon: @bricelam there's a trailing space in the key here. Can we fix that?

Eilon: (I realize it probably has no effect, but just in case.)

Issue #2455, opened 2015-06-23
RelationalTypeMapper: FK facet propagation


AndriySvyryd:
If the user specifies a store type for a principal key column that is different from the default creating a foreign key might result in a store error due to type mismatch.
This isn't too bad for explicit FKs, as the user can easily configure the store type, but for shadow state FKs and FKs found by convention it would be harder to figure out and would require more code to fix.
RelationalTypeMapper should get the store type from the principal key if it's not set on the foreign key.

ajcvickers:
@AndriySvyryd Poaching this.

PR #5889, opened 2016-06-28
Propagate type mapping from principal property to dependent property

ajcvickers:
So that if the type mapping for a primary key is changed, then the type mapping for referencing foreign keys will also change.

Issue #2455


AndriySvyryd:
:shipit:

Issue #5410, opened 2016-05-18
Update scaffolding to use ScaffoldingTypeMapper


ajcvickers:
This follows the changes made for #4425 which provides more information from the type mapper to scaffolding on what to scaffold for a given property. However, the implementation assumes that IsUnicode has been implemented and so this cannot be done until that issue is fixed--see #3420.

rowanmiller:
When this is done, verify that decimal columns have scale/precision included in column type

bricelam:
This work was also mentioned as part of #1671

AndriySvyryd:
The type for decimal properties should always be specified, even if same as default, due to #6944

ajcvickers:
@lajones Can you elaborate on why this is flagged as a breaking change?

lajones:
@ajcvickers This is part of the overall type mapping work. Though I haven't mapped out the exact API on how that will work yet, I don't think we'll manage to get through all of that without changing any of the type-mapping API. I don't know that it will specifically affect the `ScaffoldingTypeMapper` yet - we'll see.

ajcvickers:
@lajones So you are saying that using the ScaffoldingTypeMapper will affect the reverse engineering API such that it will be breaking for providers? Or that #7434 will do that, and this is part of the work needed for #7434?

lajones:
@ajcvickers The latter. I'm saying #7434 may affect the APIs on `ScaffoldingTypeMapper`.

ajcvickers:
@lajones Then I think it's okay to have only #7434 marked as breaking.

ajcvickers:
Moving this out of preview2 since it doesn't need to be done.

smitpatel:
Poaching this as it needs for #7004

smitpatel:
Fixed in #8626 & #8660 
There are some inaccuracies in how it is used filed #8697

smitpatel:
Commits 263c102e74898ccf117d8eccabff81b34235f725 & 58c8f8100b98beb3ed003c10c7f1ef59141ff050

Issue #3420, opened 2015-10-13
Provide an IsUnicode API


rowanmiller:
As pointed out in https://github.com/aspnet/EntityFramework/issues/3362 we have removed this API which forces you to specify a provider specific type in cases where you would not have needed to in EF6. We should add this into EF7.

joshmouch:
I think IsFixedLength() is probably the other one that is agnostic yet can have an easy fallback if not supported by the provider. I use that because I override the entities in Microsoft Identity 2.0, and it uses that quite a bit.

bricelam:
**Note to implementer:** Don't forget to diff the metadata in `MigrationsModelDiffer.Diff(IProperty, IProperty)`.

lajones:
*Note to implementer:* when we do this we'll need to update scaffolding to output `IsUnicode(false)` for `varchar` columns instead of outputting `HasColumnType("varchar(x)")` as we do now. See discussion in #4312.

divega:
Cleaning up milestone so that we can see this in triage.

PR #5890, opened 2016-06-28
Implement IsUnicode API

ajcvickers:
Issue #3420

I think everything in the runtime and migrations areas is implemented. The ScaffoldingTypeMapper already supports unicode, but scaffolding will not make use of this until updated as described in issue #5410.

/cc @rowanmiller Can you look at the public API and doc comments?

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Extensions/MutablePropertyExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/Extensions/MutablePropertyExtensions.cs
@@ -37,6 +37,18 @@ namespace Microsoft.EntityFrameworkCore
         }
 
         /// <summary>
+        ///     Sets a value indicating whether or not this property can persist unicode characters.
+        /// </summary>
+        /// <param name="property"> The property to set the value for. </param>
+        /// <param name="unicode"> True if the property accepts unicode characters, false if it does not, null to clear the setting. </param>
+        public static void IsUnicode([NotNull] this IMutableProperty property, bool? unicode)
+        {
+            Check.NotNull(property, nameof(property));
+
+            property[CoreAnnotationNames.UnicodeAnnotation] = unicode;
+        }
+
+        /// <summary>
         ///     Gets all foreign keys that use this property (including composite foreign keys in which this property
         ///     is included).
         /// </summary>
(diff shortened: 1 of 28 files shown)


src/Microsoft.EntityFrameworkCore/Extensions/MutablePropertyExtensions.cs
```diff
@@ -37,6 +37,18 @@ public static void SetMaxLength([NotNull] this IMutableProperty property, int? m
         }
 
         /// <summary>
+        ///     Sets a value indicating whether or not this property can persist unicode characters.
+        /// </summary>
+        /// <param name="property"> The property to set the value for. </param>
+        /// <param name="unicode"> True if the property accepts unicode characters, false if it does not, null to clear the setting. </param>
+        public static void IsUnicode([NotNull] this IMutableProperty property, bool? unicode)
```
AndriySvyryd: If it's first-class on the builder API, why not make it first-class on the Metadata API as well?

ajcvickers: Because we decided that for facets like this and max length we would make them annotations accessed by extension methods.

ajcvickers: I think we decided this mainly to keep the metadata interfaces minimal.

AndriySvyryd:
:shipit:

ajcvickers:
@rowanmiller Merged this, but would still appreciate you taking a look.

rowanmiller:
Yep looks good to me

test/Microsoft.EntityFrameworkCore.Tools.Core.FunctionalTests/Migrations/ModelSnapshotTest.cs
```diff
@@ -807,6 +807,28 @@ public void Property_maxLength_is_stored_in_snapshot()
...
+            Test(
+                builder => { builder.Entity<EntityWithStringProperty>().Property<string>("Name").IsUnicode(false); },
+                @"
+builder.Entity(""Microsoft.EntityFrameworkCore.Tools.Core.FunctionalTests.Migrations.ModelSnapshotTest+EntityWithStringProperty"", b =>
+    {
+        b.Property<int>(""Id"")
+            .ValueGeneratedOnAdd();
+
+        b.Property<string>(""Name"")
+            .HasAnnotation(""Unicode"", false);
```
bricelam: Should this be "lifted" to a Fluent API call? We do it for other relational configuration.

ajcvickers: @bricelam Good point. I'll do that.

PR #5891, opened 2016-06-29
InversePopertyAnnotation tests and fixes

### LINKED ISSUE #4931: Port EF6 Data Annotations tests

AndriySvyryd:
Moved ModelBuilder data annotations tests to Microsoft.EntityFrameworkCore.Specification.Tests
If two navigations have InversePopertyAnnotation pointing to the same property don't create a relationship
Store ambiguous navigations found by RelationshipDiscoveryConvention to avoid them being used by derived entity types
Add conventions that run when entity type is ignored to check whether any ambiguities were resolved

Fixes #4931


ajcvickers:
:shipit:

Issue #734, opened 2014-10-01
Consistenly format CLR types in exceptions messages and logging


rowanmiller:
I've been using ```'{0}'``` in the strings and just passing the ```Type``` to the formatter, this results in the fully qualified type name enclosed in single quotes... we should be consistent throughout the stack though.

divega:
Clearing up milestone so that we can see this in triage and decide the priority.

PR #5905, opened 2016-06-29
Use short, readable type names consistently in exception messages

ajcvickers:
Issue #734

Also used our name formatting methods, that did not exist when bug was filed, so that generic types are nicely formatted.


AndriySvyryd:
:shipit:

Issue #5894, opened 2016-06-29
Migrations: Different properties with same name in derived classes throws exception


jakimber:
as per [https://blogs.msdn.microsoft.com/dotnet/2016/06/27/entity-framework-core-1-0-0-available/#comment-91605](https://blogs.msdn.microsoft.com/dotnet/2016/06/27/entity-framework-core-1-0-0-available/#comment-91605) thanks for any time you can spend on this, I appreciate you are busy!

This may be a result of progress, but i'm a little confused and have spent a fair amount of time trawling the docs so would appreciate some guidance.  (Happy to post to StackOverflow if that is more appropriate too!)

### The issue
I have a model that I have been building and saving with RC2 for a while now, migrations created and updated just fine.  After my eager upgrade to 1.0.0 my migrations file still gets created just fine but `dotnet ef database update` fails with ` Sequence contains more than one matching element`.

It may be that work on 1.0.0 has made things more strict than before and I was merely getting lucky with the RC, but ideally I'd like a better fix to the workarounds that I have got.

### TL;DR;
I can no longer map two derived class properties that are named the same into a single TPH-style table.

### Steps to reproduce
After trying with my original project I have reproduced the error in a brand new vanilla project with `"Microsoft.EntityFrameworkCore.Tools": "1.0.0-preview2-final"` added in `"references"` and `"tools"`.

Here's a model that matches my design and causes the issue for me:
```
namespace Model
{
    // Base class that all three derived classes inherit
    public class MyBaseClass
    {
        public int Id { get; set; }
        public string BaseClassProperty { get; set; }
    }

    // interface that (in my design) only two of the three derived classes implement
    public interface IMyInterface
    {
        string MyInterfaceProperty { get; set; }
    }

    // First derived class, doesn't implement the interface (obviously has more to it in my real application)
    public class MyDerivedClass1 : MyBaseClass { }

    // and the classes that derive from base, and also implement the interface
    public class MyDerivedClass2 : MyBaseClass, IMyInterface
    {
        public string MyInterfaceProperty { get; set; }
    }

    public class MyDerivedClass3 : MyBaseClass, IMyInterface
    {
        public string MyInterfaceProperty { get; set; }
    }
}
```

So, before you say that interface properties are not supported and close my issue, I have subsequently tried the following:
### Attempt 1
Avoid the interface, add duplicate properties to the derived classes and map them like this:
```
protected override void OnModelCreating(ModelBuilder modelBuilder)
{
[...]

PR #5908, opened 2016-06-30
Fix for properties with shared columns in Migrations

ajcvickers:
Issue #5894

This was an issue with code added for hasLength. It assumed each column was mapped from a single property. Fix is to use any property--model validator already checks that these properties do not have different mappings.

Also added more functional tests for shared columns.


AndriySvyryd:
:shipit:

Issue #5334, opened 2016-05-11
Implement new architecture of dotnet-ef and PowerShell commands for ASP.NET Core/.NET Core


rowanmiller:
Some things to consider:
 * Do we continue to just shell to dotnet-ef from PowerShell commands? If so, how do we get better output (#4771).
 * Can we use AppDomains for xproj (#4773)

rowanmiller:
Since GitHub doesn't support multiple assignees 😄... assigning to @bricelam to catch up with @natemcmaster and other when he is back in the office to discuss where have landed and where we go from here.

natemcmaster:
Remarking for triage: this can wait until after RTM.

natemcmaster:
Additional things to consider:

- [ ] Can we change the appdomain interfaces now that everything runs through ef.exe? https://github.com/aspnet/EntityFramework/pull/6288#discussion_r74467607

PR #5919, opened 2016-06-30
Create design-time abstractions

natemcmaster:
NB: merging into a feature branch instead of dev.

Part of #5334

In order to create "ef.exe", we will need to unify types on both sides of the app domain. Currently, we do this in powershell by including "OperationHandlers.cs" which is compiled into the powershell host. Although we could continue to use transparent proxies to remote calls between types that are almost the same, if we can settle on abstractions that work, we can use ef.exe on .NET Standard libraries.

In addition, I'm adding strong types for results.

TODO: xmldocs

cc @bricelam


lajones:
Looks good to me.

bricelam:
I don't like this idea. In the past, this area has evolved too much between releases. Hence the current reflection/proxy and `IDictionary` approach.

PR #5920, opened 2016-06-30
Query: Apply search condition conversions outside of ConvertExpression

### LINKED ISSUE #5877: Query: Error using a compare with a nullable boolean
Hi, 
  I just updated to asp.net core rtm an application and I'm having a runtime error
  I have this expression:

```
clienti = db.Clienti
   .Where(cli => cli.Attivo && cli.Operativo != null && cli.Operativo.Value)
   .OrderBy(cli => cli.Descrizione)
   .ToList();

```
where Attivo is a bool and Operativo is a ?bool.
If I run this code I obtain this error message

```
Microsoft.EntityFrameworkCore.Query.Internal.SqlServerQueryCompilationContextFactory:Error: An exception occurred in the database while iterating the results of a query.
System.InvalidOperationException: The binary operator Equal is not defined for the types 'System.Nullable`1[System.Boolean]' and 'System.Boolean'.

```
Has anything changed ? This code worked fine before the upgrade.

smitpatel:
Resolves #5877 
The issue: We had `UnaryExpression` to convert values from `bool?` to `bool` in predicates for nullable columns. While visiting the `UnaryExpression.Operand` we tried to convert it to search condition by comparing it to true but `Operand` is of `bool?` type hence throwing exception.
The fix: For `UnaryExpression` with `Convert` node type, the operand is not search condition, we compare the whole expression to make a search condition if needed.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/NullSemanticsQueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/NullSemanticsQueryTestBase.cs
@@ -444,6 +444,39 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
+        public virtual void Where_nullable_bool()
+        {
+            using (var context = CreateContext())
+            {
+                context.Entities1
+                    .Where(e => e.NullableBoolA.Value)
+                    .Select(e => e.Id).ToList();
+            }
+        }
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


smitpatel:
@anpete @maumar

anpete:
:shipit:

smitpatel:
merged to 1.0.1 via add3939bd592f1c4a7f249b81d024a2ff260091d
merged to dev via 8663185a86f7a2218d1323c505c2acf6cbb1e17e

Issue #5552, opened 2016-05-27
API to get current transaction


rowanmiller:
When we moved transactions to core we lost a top level API to get the current transaction, we should add one.

rjperes:
@rowanmiller, as of now, is it possible to use a transaction in two contexts?

rjperes:
There is a way:

```
var dbtxmgr = ctx.GetInfrastructure().GetService<IDbContextTransactionManager>();
var relcon = dbtxmgr as IRelationalConnection;
var dbtx = relcon.CurrentTransaction;
var tx = dbtx.GetDbTransaction();

```
Right?

PR #5921, opened 2016-06-30
Add CurrentTransaction API to context.Database

ajcvickers:
Issue #5552

/cc @rowanmiller for public surface API docs.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Storage/IDbContextTransactionManager.cs
+++ b/src/Microsoft.EntityFrameworkCore/Storage/IDbContextTransactionManager.cs
@@ -41,5 +41,10 @@ namespace Microsoft.EntityFrameworkCore.Storage
         ///     Discards all changes made to the database in the current transaction.
         /// </summary>
         void RollbackTransaction();
+
+        /// <summary>
+        ///     Gets the current transaction.
+        /// </summary>
+        IDbContextTransaction CurrentTransaction { get; }
     }
 }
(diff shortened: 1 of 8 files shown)


lajones:
:shipit:

src/Microsoft.EntityFrameworkCore/Storage/IDbContextTransactionManager.cs
```diff
@@ -41,5 +41,10 @@ public interface IDbContextTransactionManager
         ///     Discards all changes made to the database in the current transaction.
         /// </summary>
         void RollbackTransaction();
+
+        /// <summary>
+        ///     Gets the current transaction.
+        /// </summary>
+        IDbContextTransaction CurrentTransaction { get; }
```
ajcvickers: @roji @ErikEJ @alexkeh @PrashanthIBM @gmartinezsan This property is defined on IRelationalConnection in the RTM release. Moving it to here is technically a breaking change because it is possible that a provider could have implemented this interface directly rather than through IRelationalConnection. However, we think that this is unlikely. Can you guys confirm whether or not this change would break your providers?
/cc @rowanmiller @divega

ErikEJ: I do not think so: https://github.com/ErikEJ/EntityFramework.SqlServerCompact/blob/master/src/Provider40/Storage/Internal/ISqlCeDatabaseConnection.cs

ajcvickers:
@rowanmiller Will merge now, but I can update doc comments later if needed.

ajcvickers:
Merged.

PR #5923, opened 2016-06-30
Moving sequence test into its own method.

### LINKED ISSUE #5593: Separate Sequence Test from other RevEng tests
### Steps to reproduce
In f873682 I put `[SqlServerCondition(SqlServerCondition.SupportsSequences)]` on the 2 tests `SqlServerE2ETests.E2ETest_UseAttributesInsteadOfFluentApi` and `SqlServerE2ETests.E2ETest_AllFluentApi`. To repro simply remove those attributes and run those tests on a box where the version of SQL Server is 2008. 

### The issue
This was a temporary measure to prevent CI failures. Long term the correct solution is to pull out the part of the test that depends on a `CREATE SEQUENCE` statement (see `E2E.sql`) into a separate test, remove that attribute from the 2 old tests and have only the new test depend on `[SqlServerCondition(SqlServerCondition.SupportsSequences)]`.

lajones:
Fixes #5593. Splitting a SQL Server sequence test out into a separate test method which allows the remaining tests to run even on databases such as SQL Server 2008 which do not support sequences.


natemcmaster:
:shipit:

lajones:
Checked in with commit fd456ba.

PR #5925, opened 2016-06-30
Query: Navigations: Three bug fixes.

anpete:
```
Fix: #3676 - Usage of the "let" keyword breaks grouping.
 - Fixed compiler bug in tracked, grouped queries.
Fix: #5427 - A column has been specified more than once in the order by list thrown for a complex query with orderby and navigation.
 - Fixed bug in SelectExpression.PrependToOrderBy
Fix: #4539 - Query :: Join flattening fails for some complex cases involving SelectMany
 - SelectMany after GroupJoin should not cause client-eval.
```
Also addressed most of the R# warnings in nav. rewriter.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
@@ -9,8 +9,13 @@ using Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.Northwind;
 using Microsoft.EntityFrameworkCore.Specification.Tests.TestUtilities.Xunit;
 using Xunit;
 
+// ReSharper disable InconsistentNaming
+// ReSharper disable PossibleMultipleEnumeration
+// ReSharper disable ReplaceWithSingleCallToFirstOrDefault
+// ReSharper disable ReplaceWithSingleCallToAny
+// ReSharper disable ReplaceWithSingleCallToFirst
+// ReSharper disable StringStartsWithIsCultureSpecific
 // ReSharper disable UseCollectionCountProperty
-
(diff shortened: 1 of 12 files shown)


anpete:
@maumar @smitpatel

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -4623,6 +4623,30 @@ public virtual void GroupJoin_DefaultIfEmpty3()
                 select o);
         }
 
+
+        [ConditionalFact]
+        public virtual void GroupJoin_DefaultIfEmpty_Where()
```
maumar: GroupJoin_SelectMany_Where() - no DefaultIfEmpty in the query, same thing below

anpete: Will fix.

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
```diff
@@ -43,8 +48,7 @@ var orders
             }
         }
 
-        // issue 4539
-        ////[ConditionalFact]
+        [ConditionalFact]
```
maumar: there are more tests blocked by this issue in GearsOfWar tests

maumar:
:shipit:

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
anpete: Got 'em, thanks.

PR #5942, opened 2016-07-01
Remove JsonUtility

### LINKED ISSUE #5859: Remove JsonUtility
Currently this is part of our public "Internal" surface, but is only ever intended to be used in very specific EF use cases and does not have the appropriate perf/globalization/security/etc. to be exposed. Given it's not intended to be a general use component we should make it private (or remove altogether and just string concat where we need to). #5821 is an example of what we don't want to take care of moving forwards.

natemcmaster:
Fix #5859 

cc @rowanmiller @bricelam


rowanmiller:
:shipit:

PR #5943, opened 2016-07-01
Query: Test: Add test for ?? in where predicate translated to comparison

smitpatel:
Test for #5439


smitpatel:
We already have test with `bool` column. This one is with `bool?` if we need to add. I am fine either way.

anpete:
:shipit:

PR #5948, opened 2016-07-01
Query: Reuse projections for properties in derived types mapped to sa…

### LINKED ISSUE #5909: Query: TPH query with shared column includes column multiple times.
See InheritanceSqlServerTest.Can._query_all_types_when_shared_column

Note: query returns correct results.

Model:
```C#
    public class Drink
    {
        public int Id { get; set; }
    }

    public class Coke : Drink
    {
        public int SugarGrams { get; set; }
    }

    public class Lilt : Drink
    {
        public int SugarGrams { get; set; }
    }
```
Query:
```C#
var drinks = context.Set<Drink>().ToList();
```
SQL:
```SQL
SELECT [d].[Id], [d].[Discriminator], [d].[SugarGrams], [d].[SugarGrams] 
FROM [Drink] AS [d]
WHERE [d].[Discriminator] IN (N'Lilt', N'Coke', N'Drink')
```

smitpatel:
Resolves #5909

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -430,9 +430,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 
             if (projectionIndex == -1)
             {
-                projectionIndex = _projection.Count;
-
-                AddAliasToProjection(
+                projectionIndex = AddAliasToProjection(
                     alias: null,
                     expression: new ColumnExpression(column, property, GetTableForQuerySource(querySource)));
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -581,14 +581,15 @@ var projectionIndex
                             var ce = e.TryGetColumnExpression();
 
                             return ce?.Property == columnExpression.Property
-                                   && ce?.Type == columnExpression.Type
-                                   && ce.TableAlias == columnExpression.TableAlias;
+                                   && ce?.Name == columnExpression.Name
+                                   && ce?.TableAlias == columnExpression.TableAlias;
```
anpete: Second null-conditional not required.

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -516,13 +514,15 @@ var projectionIndex
                             var ae = e as AliasExpression;
                             var ce = e.TryGetColumnExpression();
 
-                            return ce != null && ce.Property == columnExpression?.Property
-                                   && ce.TableAlias == columnExpression?.TableAlias
+                            return (ce != null
+                                && (ce.Property == columnExpression?.Property || ce.Name == columnExpression?.Name)
```
anpete: Multiple null checks on columnExpression - Is there a more efficient way to write this?

PR #5951, opened 2016-07-01
Query: Add OrderBy for definite results

### LINKED ISSUE #5220: Test issue with Substring_with_constant and Substring_with_closure
These tests assert that the first row to come out of the query is "ari" and "ria", but there's no order by to enforce which row gets selected first. The test happens to pass on SqlServer but fails on PostgreSQL.

smitpatel:
Resolves #5220 
@roji - Let us know if there are more tests which are failing due to missing order by.


anpete:
:shipit:

PR #5955, opened 2016-07-02
Make relationship discovery more flexible

roji:
Am not sure if this PR is the right way, here's what I'm trying to accomplish.

PostgreSQL supports arrays as a database type. In other words, a single column can be a an array of ints, etc. This is a much more lightweight one-to-many solution than another table with a relationship, and there are built-in operators and functions which deal with the array.

Npgsql fully supports arrays on the ADO.NET level, and I'd like to add support to the EFCore level as well. The main issue is that EFCore identifies CLR array types as relationship/navigation properties. My approach is to add an attribute `[MapAsArray]`, which would let EFCore know to leave the property alone and let it be mapped as a "scalar" (as far as EFCore is concerned). However, RelationshipDiscoveryConvention, doesn't seem to be very extensible (most of the logic is private) - this PR makes the main DiscoverRelationships method overridable.

Does the approach seem valid? Do you foresee other issues?

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
@@ -37,7 +37,11 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
         /// </summary>
         public const string AmbiguousNavigationsAnnotationName = "RelationshipDiscoveryConvention:AmbiguousNavigations";
 
-        private InternalEntityTypeBuilder DiscoverRelationships(InternalEntityTypeBuilder entityTypeBuilder)
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected virtual InternalEntityTypeBuilder DiscoverRelationships(InternalEntityTypeBuilder entityTypeBuilder)
         {
             if (!entityTypeBuilder.Metadata.HasClrType()
                 || entityTypeBuilder.ModelBuilder.IsIgnored(entityTypeBuilder.Metadata.ClrType, ConfigurationSource.Convention))


ajcvickers:
@roji Does explicitly telling EF that the property is a scalar property cause it to be mapped correctly? For example, `modelBuilder.Entity<Foo>().Property(e => e.MyArray);` Theoretically, this should prevent the convention from treating it as a navigation.

roji:
@ajcvickers it seems like I had an issue with my type mapper implementation which caused the error. I've managed to get arrays to work although I admit I'm not sure exactly why. Basically my NpgsqlTypeMapper overrides `FindMapping(Type clrType)`, and if no simple mapping is found, checks whether clrType is an array or generic list. If so it returns a proper mapping. Somehow this doesn't conflict with relationships, so I'm going to close this for now.

It would be good to understand though - does the relationship discovery automatically ignore properties for which a database mapping has been found?

smitpatel:
@roji - Yes, it does.
https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Metadata/Conventions/Internal/RelationalRelationshipDiscoveryConvention.cs has function which finds the target type for navigation. It returns null value for primitive properties and type mapper is used to determine whether a property is primitive property. Therefore as long as a CLR type can be mapped to database type by type mapper, it won't be discovered as navigation property.

roji:
Great, must have missed that. This probably means that array support isn't going to require anything special from the core.

I'm off to add scaffolding support as well. Thanks for the help!

ajcvickers:
@roji Awesome that you got it working. It would be good to know if it works with other parts of the stack--like change detection, queries, and updates. It's not something we have ever bee nable to try before, so I expect that there may be some issues.

roji:
@ajcvickers will try to do thorough testing and will let you know if I run into anything. It's really cool to see EFCore create tables with int[] etc.

alexzaytsev-newsroomly:
@ajcvickers Thanks to @roji 's contribution, EF core works fine with arrays and npgsql. However, the InMemoryProvider seems to throw this exception:

System.InvalidOperationException: The property 'MyEntity.MyProp' could not be mapped, because it is of type 'Int32[]' which is not a supported primitive type or a valid entity type. Either explicitly map this property, or ignore it using the '[NotMapped]' attribute or by using 'EntityTypeBuilder.Ignore' in 'OnModelCreating'.

Is there anything that can be done to enable that support for InMemoryDatabase?

ajcvickers:
@alexzaytsev-newsroomly Filed #11926

PR #5964, opened 2016-07-03
Make IsTypeMapped virtual

roji:
Rather than an extension method. This allows it to be overridden by providers.

The use case is that in Npgsql, the same CLR type may correspond to several store types, without any of the store types being an obvious default. In this case, `FindMapping(Type clrType)` can't return an actual mapping (since there's no default), but the property should still be a candidate and the user should be able to specify the exact store type with the fluent API.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalTypeMapper.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalTypeMapper.cs
@@ -43,6 +43,13 @@ namespace Microsoft.EntityFrameworkCore.Storage
         RelationalTypeMapping FindMapping([NotNull] string storeType);
 
         /// <summary>
+        ///     Gets a value indicating whether the given .NET type is mapped.
+        /// </summary>
+        /// <param name="clrType"> The .NET type. </param>
+        /// <returns> True if the type can be mapped; otherwise false. </returns>
+        bool IsTypeMapped([NotNull] Type clrType);
+
+        /// <summary>
         ///     Ensures that the given type name is a valid type for the relational database.
         ///     An exception is thrown if it is not a valid type.
         /// </summary>
--- a/src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalTypeMapper.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalTypeMapper.cs
@@ -116,6 +116,18 @@ namespace Microsoft.EntityFrameworkCore.Storage
         }
 
         /// <summary>
+        ///     Gets a value indicating whether the given .NET type is mapped.
+        /// </summary>
+        /// <param name="clrType"> The .NET type. </param>
+        /// <returns> True if the type can be mapped; otherwise false. </returns>
+        public virtual bool IsTypeMapped(Type clrType)
+        {
+            Check.NotNull(clrType, nameof(clrType));
+
+            return FindMapping(clrType) != null;
+        }
+
+        /// <summary>
         ///     Creates the mapping for the given database type.
         /// </summary>
         /// <param name="storeType">The type to create the mapping for.</param>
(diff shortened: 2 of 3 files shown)


ajcvickers:
@roji If a property of the given CLR type appears in an entity, then what determines which mapping is returned?

roji:
That's exactly the problem - if a property with the given CLR type simply appears with no extra instructions, the model should fail validation because there's no default mapping for that CLR type (preferably with an explicit error message). However, the user should still be able to manually set the column type with the fluent API, and this should make everything work.

I haven't yet analyzed everything that's needed to actually make this work, but at the very least this doesn't currently seem possible because the model validation simply calls `IsTypeMapped`, which itself calls `TypeMapper.FindMapping(Type clrType)`. While `FindMapping` can't return a mapping (there is no default), `IsTypeMapped` should return true.

ajcvickers:
@roji We have been working with the assumption that there is a reasonable (if possibly arbitrary) default for any CLR type that can be mapped. It would be interesting to know a bit more about the CLR types and possible mappings involved.

With regard to the PR, adding a member to IRelationalTypeMapper is a breaking change, so we would need a very compelling reason to do this now we are post RTM. So if we do need to make a change to the semantics of type mapping like this, then we will probably have to find a different way to do it. Also, it would be good to analyze the other places that FindMapping is being used. I suspect that without other changes we could hit null ref exceptions because code assumes that if the type passed the validator, then the CLR type is mappable.

roji:
@ajcvickers I understand, I assumed that since implementations of IRelationalTypeMapper are probably all subclasses of RelationalTypeMapper at this point, and that class implements the proposed method, there would be no breakage.

As a general rule, the vast majority of provider-related methods in EF are virtual so as to allow customization by providers. RelationalTypeMapperExtensions is an annoying exception to this rule - I understand these are meant to be small extension that providers in general wouldn't want to change, but here I am with a problem. Just saying this to discourage defining logic in extension methods in this way in the future.

My scenario is admittedly an edge case even for Npgsql. PostgreSQL [has several internal uint types](https://www.postgresql.org/docs/current/static/ddl-system-columns.html), which aren't really supposed to be user-facing; oid (internal row ID), xid (transaction id), cid (command id). I'm working on implementing optimistic concurrency, which depends on accessing the row's xmin column (id of last transaction which modified the row), which has type xid.

Now the way Npgsql currently works at the ADO.NET level, none of these three types are set up as a default for uint, because they're internal and users generally shouldn't be using them. However, users can still send these types by explicitly setting a parameter's NpgsqlDbType. In other words, if a user naively tries to send a uint, that will fail, but they can manually specify xmin if they choose to.

Since this is an edge case even of Npgsql, I'll close this and work around it which some special handling in NpgsqlRelationalMapper. I guess it's worth keeping in mind in case this kind of issue pops with another mapping.

ajcvickers:
@roji What you say makes sense. I can see the need to allow a type to be mapped conditionally, based on the context in which it occurs. Can you open an issue for this?

The reason IsTypeMapped is an extension method is that it does not actually form part of the contract of the type mapper. Code consuming the type mapper can assume that if a CLR type is mapped, then calling FindType for that CLR type will not return null. This means that for cases like FromSql where there is no information available other than the CLR type a type mapping will be available. So when we change this we need to make sure code that consumes the type mapper can handle the change.

With regard to all providers inheriting from the type, that is a good point. It may give us wiggle room to change the interface if we need to.

roji:
@ajvickers,

> @roji What you say makes sense. I can see the need to allow a type to be mapped conditionally, based on the context in which it occurs. Can you open an issue for this?

The uint scenario described above is not so much about context-based conditional mapping, but rather explicit-only mapping. On the ADO.NET level, that means that a user *must* set the *DbType to some explicit value (in my case, `NpgsqlDbType.Xid`); simply trying to create an NpgsqlParameter and setting its value to a uint will fail, but if you set its NpgsqlDbType to Xid that will work. The translation of this idea to the EFCore level seems to be that uint properties on entities should cause a model validation error, *unless* the column type for those properties is explicitly set.

Again, since this is such an edge case I simply worked around it, do you want me to open an issue for the above?

> The reason IsTypeMapped is an extension method is that it does not actually form part of the contract of the type mapper. Code consuming the type mapper can assume that if a CLR type is mapped, then calling FindType for that CLR type will not return null. This means that for cases like FromSql where there is no information available other than the CLR type a type mapping will be available. So when we change this we need to make sure code that consumes the type mapper can handle the change.

I understand. The point described above is that it may make sense for IsTypeMapped to return true, but for FindType to return null: that would mean that the type *can* be mapped, but that an explicit column type is required. Again, to me this simply doesn't seem important enough to implement in EFCore.

Issue #5537, opened 2016-05-26
Simple way to specify a value generator for a property


rowanmiller:
We have the built in concept of value generators, but currently you have to use some pretty low level building blocks to configure one and it's difficult to set a generator for a given property. We should add a top level API for this.

PR #5985, opened 2016-07-05
Make it easy to set a custom value generator

ajcvickers:
Issue #5537. Also see #5303.

This change added fluent API set set a custom value generator for use on a property. The value generator will be used regardless of provider and does not do anything special to ensure Migrations or the update pipeline are in sync with what is happening, but is useful to make a change to a value generator when these things do not matter or are already setup correctly.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Extensions/MutablePropertyExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/Extensions/MutablePropertyExtensions.cs
@@ -8,6 +8,7 @@ using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
+using Microsoft.EntityFrameworkCore.ValueGeneration;
 
 // ReSharper disable once CheckNamespace
 namespace Microsoft.EntityFrameworkCore
@@ -17,6 +18,30 @@ namespace Microsoft.EntityFrameworkCore
     /// </summary>
     public static class MutablePropertyExtensions
     {
+        /// <summary>
+        ///     <para>
+        ///         Sets the factory to use for generating values for this property, or null to clear any previously set factory.
+        ///     </para>
+        ///     <para>
+        ///         Setting null does not disable value generation for this property, it just clears any generator explicitly
+        ///         configured for this property. The database provider may still have a value generator for the property type.
+        ///     </para>
+        /// </summary>
+        /// <param name="property"> The property to set the value generator for. </param>
+        /// <param name="valueGeneratorFactory">
+        ///     A factory that will be used to create the value generator, or null to
+        ///     clear any previously set factory.
+        /// </param>
+        public static void SetValueGeneratorFactory(
+            [NotNull] this IMutableProperty property,
+            [NotNull] Func<IProperty, IEntityType, ValueGenerator> valueGeneratorFactory)
+        {
+            Check.NotNull(property, nameof(property));
+            Check.NotNull(valueGeneratorFactory, nameof(valueGeneratorFactory));
+
+            property[CoreAnnotationNames.ValueGeneratorFactoryAnnotation] = valueGeneratorFactory;
+        }
+
         /// <summary>
         ///     Sets the maximum length of data that is allowed in this property. For example, if the property is a <see cref="string" /> '
         ///     then this is the maximum number of characters.
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
@@ -1,10 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
+using Microsoft.EntityFrameworkCore.ValueGeneration;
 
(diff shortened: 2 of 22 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
```diff
@@ -103,6 +105,90 @@ public virtual PropertyBuilder IsUnicode(bool unicode = true)
         }
 
         /// <summary>
+        ///     <para>
+        ///         Configures the type to use to generate values for this property.
```
rowanmiller: I'd go with this...

> Configures the <see cref="ValueGenerator" /> that will generate values for this property.

ajcvickers: Configures the _type_ that will generate values for this property?

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
```diff
@@ -103,6 +105,90 @@ public virtual PropertyBuilder IsUnicode(bool unicode = true)
...
+        ///     </para>
+        ///     <para>
+        ///         Values are generated when the entity is added to the context using, for example,
+        ///         <see cref="DbContext.Add{TEntity}" />. Values are generated only when the value of the property
+        ///         is null or the CLR default.
```
rowanmiller: We've used this wording elsewhere, I think it's a little clearer, but also ok with what you have if you want to leave it...

> Values are generated only when the property is assigned the CLR default value (null for string, 0 for int, Guid.Empty for Guid, etc.).

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
```diff
@@ -103,6 +105,90 @@ public virtual PropertyBuilder IsUnicode(bool unicode = true)
...
+        ///     </para>
+        ///     <para>
+        ///         A single instance of this type will be created and used to generate values for this property in all
+        ///         instances of the entity type. The type must be instantiable and have a parameterless constructor.
+        ///     </para>
+        ///     <para>
+        ///         This method is intended for use with custom value generation. Value generation for common cases is
+        ///         usually handled automatically by the Entity Framework provider.
```
rowanmiller: We typically use database provider rather than Entity Framework provider.

> ...usually handled automatically by the database provider

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
```diff
@@ -103,6 +105,90 @@ public virtual PropertyBuilder IsUnicode(bool unicode = true)
...
+        ///         this will be used to generate values for this property in all instances of the entity type.
+        ///     </para>
+        ///     <para>
+        ///         This method is intended for use with custom value generation. Value generation for common cases is
+        ///         usually handled automatically by the Entity Framework provider.
+        ///     </para>
+        /// </summary>
+        /// <param name="factory"> A delegate that will be used to create value generator instances. </param>
+        /// <returns> The same builder instance so that multiple configuration calls can be chained. </returns>
+        public virtual PropertyBuilder HasValueGenerator([NotNull] Func<IProperty, IEntityType, ValueGenerator> factory)
```
rowanmiller: What are the use cases for this one? It seems pretty complicated for the top level API... but maybe I'm missing a use case 😄

src/Microsoft.EntityFrameworkCore/Extensions/MutablePropertyExtensions.cs
```diff
@@ -18,6 +19,24 @@ namespace Microsoft.EntityFrameworkCore
     public static class MutablePropertyExtensions
     {
         /// <summary>
+        ///     Sets the factory to use for generating values for this property, or null to clear any previously set factory.
```
rowanmiller: I'd add something to the effect of

> Setting null does not disable value generation for this property, it just clears any generator explicitly configured for this property. The database provider may still have a value generator for the property type.

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
ajcvickers: We could move it off the top level if you want. This is what all the other APIs are doing, they are just sugar for when you have a simple type. I don't think it is any more complicated that the other APIs. In some respects it is simpler because it is less magic.

rowanmiller: When I got to the resources I realized this is needed for non-parameterless constructors... which is enough of a reason to leave it here (IMNVHO).

rowanmiller:
:shipit: from my perspective. ❤️ that we are getting this in BTW 🎵 💃 🎉.

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
rowanmiller: Sounds good

anpete:
@ajcvickers @rowanmiller Is it worth going over this quickly at the design meeting?

ajcvickers:
@anpete Do you have concerns?

AndriySvyryd:
>                 annotations.Remove(annotationToRemove);

Also change this to use IgnoreAnnotations

---
Refers to: src/Microsoft.EntityFrameworkCore.Design.Core/Migrations/Design/CSharpSnapshotGenerator.cs:432 in 1183085. [](commit_id = 1183085b5de817f449c196e1849ac409626827f9, deletion_comment = False)

anpete:
@ajcvickers No, but if feels like an ideal candidate for us to kick off our 1.1 design process.

AndriySvyryd:
:shipit:

ajcvickers:
@anpete Okay. Suggest we wait until @divega is back, then I will present something.

anpete:
@ajcvickers Sounds good.

PR #5986, opened 2016-07-05
Split dotnet-ef into separate project and use nuspec to pack

### LINKED ISSUE #5684: Don't use Sake to re-pack the tooling nupkgs
dotnet-pack does not provide us enough options to produce the nupkgs we need for Ms.EF.Tools and Ms.Ef.Design. We hack around this with Sake. We should use a better tool to creating these packages.

natemcmaster:
No functional changes to product. Just rearranging files and using nuspec instead of dotnet-pack.

Close #5684

cc @pranavkm @bricelam


makefile.shade
```diff
@@ -1,13 +1,10 @@
...
-use namespace="System.Linq"
+use import="Json"
 
-default BASE_DIR_LOCAL='${Directory.GetCurrentDirectory()}'
-default BUILD_DIR_LOCAL='${Path.Combine(BASE_DIR_LOCAL, "artifacts", "build")}'
-default TOOL_PACKAGE_NAME='Microsoft.EntityFrameworkCore.Tools'
-default CLI_TOOL_PROJECT_NAME='Microsoft.EntityFrameworkCore.Design'
-default TOOL_EXE_NAME='${CLI_TOOL_PROJECT_NAME}.exe'
-default NUGET_EXE_PATH='${E("KOREBUILD_NUGET_EXE") ?? Path.Combine(BASE_DIR_LOCAL, ".build", "nuget.exe")}'
+default BASE_DIR='${Directory.GetCurrentDirectory()}'
```
pranavkm: Could these changes possibly wait for a bit? @anurse's KoreBuild v2 changes are nearly done and a bunch of this code would have to be transpiled to xml soon. Might as well start off writing this in xml

makefile.shade
```diff
@@ -36,123 +49,71 @@ k-standard-goals
...
-          // Delete temporary directory we used to repack.
-          Directory.Delete(extractToDirectory, true);
-      }
-      catch
-      {
-          // Don't care if we couldn't delete the temp directory.
-      }
+    ExecClr(NUGET_EXE_PATH, "pack src/Microsoft.EntityFrameworkCore.Design.nuspec -Verbosity detailed" +
+      " -OutputDirectory " + BUILD_DIR + 
+      " -Version 1.0.0-preview3-" + E("DOTNET_BUILD_VERSION") +
```
pranavkm: This is going to be pretty tricky to keep update to date between releases.

makefile.shade
natemcmaster: How long? I'm working on other changes that need to happen on top of this.

makefile.shade
natemcmaster: I assume you mean that it would be nicer if a script could update this instead of manually updating...as changing the makefile is not actually hard. Ironically, I didn't do this because of the msbuild/project.json changes that are coming. Is it sufficient for now though?

natemcmaster:
The work done in this PR is already obsolete. Closing as we are pivoting again.

cref https://github.com/NuGet/Home/issues/3023#issuecomment-232101111

Issue #5395, opened 2016-05-17
Remove culture dependency from OrderBy tests


Tasteful:
I'm running the following test and what I can see the result is not correct.
```
Test Name:	Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.QuerySqlServerTest.OrderBy
Test FullName:	Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.QuerySqlServerTest.OrderBy
Test Source:	C:\Git\EntityFramework\test\Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests\QuerySqlServerTest.cs : line 2876
Test Outcome:	Failed
Test Duration:	0:00:01,902

Result StackTrace:	
at Xunit.Assert.Equal[T](T expected, T actual, IEqualityComparer`1 comparer) in C:\BuildAgent\work\cb37e9acf085d108\src\xunit.assert\Asserts\EqualityAsserts.cs:line 35
   at Microsoft.EntityFrameworkCore.Specification.Tests.TestHelpers.AssertResults[T](IList`1 expected, IList`1 actual, Boolean assertOrder, Action`2 asserter) in C:\Git\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\TestHelpers.cs:line 172
   at Microsoft.EntityFrameworkCore.Specification.Tests.QueryTestBase`1.AssertQuery[TItem](Func`2 efQuery, Func`2 l2oQuery, Boolean assertOrder, Int32 entryCount, Action`2 asserter) in C:\Git\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\QueryTestBase.cs:line 5231
   at Microsoft.EntityFrameworkCore.Specification.Tests.QueryTestBase`1.AssertQuery[TItem](Func`2 query, Boolean assertOrder, Int32 entryCount, Action`2 asserter) in C:\Git\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\QueryTestBase.cs:line 5131
   at Microsoft.EntityFrameworkCore.Specification.Tests.QueryTestBase`1.OrderBy() in C:\Git\EntityFramework\src\Microsoft.EntityFrameworkCore.Specification.Tests\QueryTestBase.cs:line 2713
[...]

PR #5988, opened 2016-07-05
Make OrderBy tests run on machines with different culture settings

ajcvickers:
Issue #5395 

Fix is to make L2O queries use InvariantCulture.


AndriySvyryd:
:shipit:

Issue #5738, opened 2016-06-13
Query Perf: Selecting entire table for navigation properties in projections


ruchanb:
### Steps to reproduce
```
public class Country
{        
    public string CountryCode { get; set; }
    public string CountryName { get; set; }
    public virtual ICollection<State> States { get; set; }
}

public class State 
{
    public string StateCode { get; set; }
    public string StateName { get; set; }
    public string CountryCode { get; set; }       
    public Country Country { get; set; }
}
```

Mapping on Country
```
builder.ToTable("Country");
builder.HasKey(pr => pr.CountryCode);
builder.HasMany(m => m.States).WithOne(i => i.Country).HasForeignKey(m => m.CountryCode);

```
Mapping on State
```
builder.ToTable("State");
builder.HasKey(pr => pr.StateCode);
builder.HasOne(m => m.Country).WithMany(m => m.States).HasForeignKey(m => m.CountryCode);

```
Now the Query
```
var query = _context.Countries
                    .Where(i => i.CountryCode == "USA")
                    .Select(m => new
                                  {
                                       m.CountryName,
                                       m.CountryCode,
                                       States = m.States.Select(x => new
                                                         {
                                                            x.StateCode,
                                                            x.StateName,
                                                            x.CountryCode
                                                         })
                                  }).AsQueryable();
  return query.ToList();
```
### The issue
When I run SQL Profiler on this query I get following
```
SELECT [i].[CountryName], [i].[CountryCode]
FROM [Country] AS [i]
WHERE [i].[CountryCode] = N'USA'
```

and
```
SELECT [x].[CountryCode], [x].[StateCode], [x].[StateName]
FROM [State] AS [x]
```

I understand the two queries are not combined as in previous versions of EntityFramework. However, here the query on ``State`` is just select all without any where condition from the ``Country`` it is looking for. It does gives me ``States`` from ``'USA'`` but it is doing so in the memory, not the SQL server.

Now make the query to include about 5 countries, it gives me 5 of these queries without any condition for ``Country``
```
SELECT [x].[CountryCode], [x].[StateCode], [x].[StateName]
FROM [State] AS [x]
```

### Further technical details

EF Core version: 1.0.0-rc2-final
Operating system: Windows 10
SQL Server 2008
Visual Studio version: VS 2015 update 2
Framework Target:  net461

rowanmiller:
@anpete @maumar is this already tracked by another issue?

maumar:
[...]

PR #5990, opened 2016-07-06
Fix to #5738 - Query Perf: Selecting entire table for navigation properties in projections

maumar:

would get translated to:

This would produce outer query to fetch all the customers and for each customer another query, fetching ALL the orders each time filtering out the orders for each customer and returning only the first one on the client (because we didn't know how to handle o.CustomerId in this case)

SELECT [t].[CustomerID]
FROM [Customers] AS [c]

FROM [Orders] AS [o]

We achieve this by introducing a new query method that executes a delegate before enumerating it's source, and then executing another delegate after all elements have been enumerated or the enumerator has been disposed.
Those functions set and remove the necessary parameters just before/after the inner query is accessed.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
@@ -792,5 +792,78 @@ namespace Microsoft.EntityFrameworkCore.Query
 
             public void Dispose() => _includeCollectionIterator?.Dispose();
         }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual MethodInfo PreExecuteMethod => _preExecuteMethodInfo;
+
+        private static readonly MethodInfo _preExecuteMethodInfo
+            = typeof(AsyncQueryMethodProvider)
+                .GetTypeInfo().GetDeclaredMethod(nameof(_PreExecute));
+
+        [UsedImplicitly]
+        // ReSharper disable once InconsistentNaming
+        private static IAsyncEnumerable<TElement> _PreExecute<TElement>(
+            Action preExecuteAction, IAsyncEnumerable<TElement> source, Func<object> postExecuteAction)
+            => new PreExecutor<TElement>(preExecuteAction, source, postExecuteAction);
+
+        private sealed class PreExecutor<TElement> : IAsyncEnumerable<TElement>
+        {
+            private readonly Action _preExecuteAction;
+            private readonly IAsyncEnumerable<TElement> _innerEnumerable;
+            private readonly Func<object> _postExecuteAction;
+
+            public PreExecutor(Action preExecuteAction, IAsyncEnumerable<TElement> innerEnumerable, Func<object> postExecuteAction)
+            {
+                _preExecuteAction = preExecuteAction;
+                _innerEnumerable = innerEnumerable;
+                _postExecuteAction = postExecuteAction;
+            }
+
+            public IAsyncEnumerator<TElement> GetEnumerator() => new PreExecuteEnumerator(this);
+
+            IAsyncEnumerator<TElement> IAsyncEnumerable<TElement>.GetEnumerator() => new PreExecuteEnumerator(this);
+
+            private sealed class PreExecuteEnumerator : IAsyncEnumerator<TElement>
+            {
+                private readonly PreExecutor<TElement> _preExecutor;
+                private readonly IAsyncEnumerator<TElement> _innerEnumerator;
+                private bool _postExecuteActionComplete = false;
+
+                public PreExecuteEnumerator(PreExecutor<TElement> preExecutor)
+                {
+                    _preExecutor = preExecutor;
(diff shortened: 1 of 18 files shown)


maumar:
@anpete @smitpatel

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -792,5 +792,78 @@ public IAsyncEnumerable<EntityLoadInfo> Load(QueryContext queryContext, IInclude
 
             public void Dispose() => _includeCollectionIterator?.Dispose();
         }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual MethodInfo PreExecuteMethod => _preExecuteMethodInfo;
```
maumar: better name suggestions?

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
```diff
@@ -62,6 +62,17 @@ public override bool Equals(object obj)
             }
 
             public override int GetHashCode() => 0;
+
+            public CommandCacheKey Clone()
```
maumar: before we would use the parameter collection as cache key, and since we modify the parameter collection during execution the cache would become corrupted - so we need to clone them now, only doing it upon adding new entry to the cache so the perf impact shouldn't be too big

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -966,10 +1018,15 @@ var sqlTranslatingExpressionVisitor
 
                 foreach (var ordering in orderByClause.Orderings)
                 {
+                    var canBindPropertyToOuterParameter = CanBindPropertyToOuterParameter;
+                    CanBindPropertyToOuterParameter = false;
```
maumar: we disable this for order by, because you can't have parameter (that is integer) in the order by

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -617,6 +617,11 @@ var expression
                         .BindMethodCallExpression(methodCallExpression, CreateAliasedColumnExpressionCore);
             }
 
+            if (expression == null)
```
anpete: Ternary (can you run R# over the changes?)

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/ShaperCommandContext.cs
anpete: Use Dictionary ctor that takes a Dictionary.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1291,5 +1368,146 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
         }
 
         #endregion
+
+        public virtual bool CanBindPropertyToOuterParameter { get; private set; } = true;
```
anpete: Can this be private?

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1291,5 +1368,146 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
...
+                    return Expression.Parameter(
+                        property.ClrType,
+                        parameterName);
+                }
+            }
+
+            return null;
+        }
+
+        private readonly MethodInfo _addParameterMehodInfo
```
anpete: "Mehod"

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -792,5 +792,78 @@ public IAsyncEnumerable<EntityLoadInfo> Load(QueryContext queryContext, IInclude
...
+
+        private static readonly MethodInfo _preExecuteMethodInfo
+            = typeof(AsyncQueryMethodProvider)
+                .GetTypeInfo().GetDeclaredMethod(nameof(_PreExecute));
+
+        [UsedImplicitly]
+        // ReSharper disable once InconsistentNaming
+        private static IAsyncEnumerable<TElement> _PreExecute<TElement>(
```
anpete: Why is this generalized? Can we just make it specifically about parameter introduction?

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
anpete: Add comment in code?

maumar:
sent a new pr vs 1.0.1 branch

Issue #5941, opened 2016-07-01
InternalDbSet caches EF.Property MethodInfo in generic type


anpete:
Which means we get a separate cached instance for each closed generic.

PR #5999, opened 2016-07-06
Move EF.Property MethodInfo to common location

ajcvickers:
Issue #5941


natemcmaster:
I think EntityQueryModelVisitor also stores methodinfo for ef.property. We could probably update that one too while we're at it.

natemcmaster:
:shipit:

natemcmaster:
nvm. You got it :+1:

PR #6000, opened 2016-07-06
Use fluent API in snapshot for HasMaxLength and IsUnicode

ajcvickers:
Instead of just using the raw annotation syntax.


AndriySvyryd:
:shipit:

PR #6002, opened 2016-07-06
[1.0.1] Fix #5901 - Query: Wrong data in included navigation when usi…

### LINKED ISSUE #5901: Query: Wrong data in included navigation when using Skip() method
### Steps to reproduce
For tests I created brand new Asp.Net Core WebApplication (.NET Framework) with individual user accounts. In this app I created two simple models with one to many relationship:
classes
Then I added DbSets to context class:
context
Then I created 20 FirstModel objects with 10 SecondModel objects per each FirtsModel object and stored them in db.

If I read all FirtModel objects with included SecondModel objects everything is ok:
without
without_deb

But If i do the same but at this time I skip some items, list with SecondModel objects is empty:
skip
skip_deb

### The issue
Is described above.

### Further technical details

EF Core version: "Microsoft.AspNetCore.Identity.EntityFrameworkCore": "1.0.0"
Operating system: Win 8.1
Visual Studio version: VS 2015 with update 3

I'm writing web app from RC1 version. In Rc1 and Rc2 everything was ok.

anpete:
Tweak logic in IncludeExpressionVisitor to not apply EXISTS optimization if the parent query uses Skip without a predicate.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/IncludeTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/IncludeTestBase.cs
@@ -9,6 +9,7 @@ using Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.Northwind;
 using Microsoft.EntityFrameworkCore.Specification.Tests.TestUtilities.Xunit;
 using Microsoft.EntityFrameworkCore.Internal;
 using Xunit;
+// ReSharper disable StringStartsWithIsCultureSpecific
 
 #if NETSTANDARD1_3
 using System.Reflection;
@@ -67,8 +68,9 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
                         {
                             using (var context = CreateContext())
                             {
-                                var query = context.Set<Order>()
+                                context.Set<Order>()
                                     .Include(o => new { o.Customer, o.OrderDetails })
+                                    // ReSharper disable once ReturnValueOfPureMethodIsNotUsed
                                     .ToList();
                             }
                         }).Message);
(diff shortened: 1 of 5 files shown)


anpete:
@smitpatel

anpete:
@maumar

src/Microsoft.EntityFrameworkCore.Specification.Tests/IncludeTestBase.cs
```diff
@@ -175,6 +178,43 @@ var customers
...
+            using (var context = CreateContext())
+            {
+                var customers
+                    = context.Set<Customer>()
+                        .Skip(10)
+                        .Include(c => c.Orders)
+                        .ToList();
+
+                Assert.Equal(81, customers.Count);
+                Assert.Equal(730, customers.Where(c => c.Orders != null).SelectMany(c => c.Orders).Count());
```
maumar: this is non-deterministic and will probably fail on postgres @roji  reported similar problems a few times

src/Microsoft.EntityFrameworkCore.Specification.Tests/IncludeTestBase.cs
```diff
@@ -175,6 +178,43 @@ var customers
...
+            {
+                var customers
+                    = context.Set<Customer>()
+                        .Skip(10)
+                        .Take(5)
+                        .Include(c => c.Orders)
+                        .ToList();
+
+                Assert.Equal(5, customers.Count);
+                Assert.Equal(30, customers.Where(c => c.Orders != null).SelectMany(c => c.Orders).Count());
```
maumar: same as above

maumar:
:shipit:

Issue #5377, opened 2016-05-16
Metadata :: confusing exception when setting up Navigation with shadow FK


maumar:
repro:

```
    public class MyContext : DbContext
    {
        public DbSet<Item> Items { get; set; }
        public DbSet<Offer> Offers { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Offer>().HasOne(e => e.Item).WithOne().HasForeignKey("Offer", "ItemId").IsRequired();
        }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer("Server=.;Database=Repro5377;Trusted_Connection=True;MultipleActiveResultSets=true;");
        }
    }

    public class Offer
    {
        public Guid Id { get; set; }
        public Item Item { get; set; }
        public decimal Price { get; set; }
    }

    public class Item
    {
        public Guid Id { get; set; }
    }
```

Exception:

System.InvalidOperationException: You are configuring a relationship between 'Offer' and 'Item' but have specified a foreign key on 'Offer'. The foreign key must be defined on a type that is part of the relationship.
   at Microsoft.EntityFrameworkCore.Metadata.Builders.ReferenceReferenceBuilder.SetDependentEntityType(String dependentEntityTypeName) in D:\k\EntityFramework\src\Microsoft.EntityFrameworkCore\Metadata\Builders\ReferenceReferenceBuilder.cs:line 237
   at Microsoft.EntityFrameworkCore.Metadata.Builders.ReferenceReferenceBuilder`2.HasForeignKey(String dependentEntityTypeName, String[] foreignKeyPropertyNames) in D:\k\EntityFramework\src\Microsoft.EntityFrameworkCore\Metadata\Builders\ReferenceReferenceBuilder`.cs:line 174
   at Repro5107.MyContext.OnModelCreating(ModelBuilder modelBuilder) in D:\Projects\Repro5107\Repro5107\Program.cs:line 16
   at Microsoft.EntityFrameworkCore.Infrastructure.ModelCustomizer.Customize(ModelBuilder modelBuilder, DbContext dbContext) in D:\k\EntityFramework\src\Microsoft.EntityFrameworkCore\Infrastructure\ModelCustomizer.cs:line 8
[...]

PR #6003, opened 2016-07-06
Sugar for HasForeignKey/HasPrincipalKey

ajcvickers:
Issue #5377

Added back generic overloads so that in the common case of a shadow property in CLR entity types the generic method can be used instead of using typeof().

Also, only for these APIs, the short name of the type can be passed as a string since the resolution space is very small--it just checks the name at either end of the relationship.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Builders/ReferenceReferenceBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Builders/ReferenceReferenceBuilder.cs
@@ -188,6 +188,37 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Builders
         ///         added to the principal entity type to serve as the reference key.
         ///     </para>
         /// </summary>
+        /// <typeparam name="TDependentEntity">
+        ///     The entity type that is the dependent in this relationship (the type that has the foreign key
+        ///     properties).
+        /// </typeparam>
+        /// <param name="foreignKeyPropertyNames">
+        ///     The name(s) of the foreign key property(s).
+        /// </param>
+        /// <returns> The same builder instance so that multiple configuration calls can be chained. </returns>
+        public virtual ReferenceReferenceBuilder HasForeignKey<TDependentEntity>(
+            [NotNull] params string[] foreignKeyPropertyNames)
+            where TDependentEntity : class 
+            => HasForeignKey(typeof(TDependentEntity), foreignKeyPropertyNames);
+
+        /// <summary>
+        ///     <para>
+        ///         Configures the property(s) to use as the foreign key for this relationship.
+        ///     </para>
+        ///     <para>
+        ///         If the specified property name(s) do not exist on the entity type then a new shadow state
+        ///         property(s) will be added to serve as the foreign key. A shadow state property is one
+        ///         that does not have a corresponding property in the entity class. The current value for the
+        ///         property is stored in the <see cref="ChangeTracker" /> rather than being stored in instances
+        ///         of the entity class.
+        ///     </para>
+        ///     <para>
+        ///         If <see cref="HasPrincipalKey(System.Type,string[])" /> is not specified, then an attempt will be made to
+        ///         match the data type and order of foreign key properties against the primary key of the principal
+        ///         entity type. If they do not match, new shadow state properties that form a unique index will be
+        ///         added to the principal entity type to serve as the reference key.
+        ///     </para>
+        /// </summary>
         /// <param name="dependentEntityTypeName">
         ///     The name of the entity type that is the dependent in this relationship (the type that has the foreign
         ///     key properties).
@@ -203,7 +234,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Builders
                 SetDependentEntityType(Check.NotEmpty(dependentEntityTypeName, nameof(dependentEntityTypeName)))
                     .HasForeignKey(Check.NotEmpty(foreignKeyPropertyNames, nameof(foreignKeyPropertyNames)), ConfigurationSource.Explicit),
                 this,
-                inverted: Builder.Metadata.DeclaringEntityType.Name != dependentEntityTypeName,
+                inverted: Builder.Metadata.DeclaringEntityType.Name != ResolveEntityType(dependentEntityTypeName).Name,
                 foreignKeySet: true);
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Builders/ReferenceReferenceBuilder.cs
```diff
@@ -188,6 +188,36 @@ public virtual ReferenceReferenceBuilder HasAnnotation([NotNull] string annotati
...
         /// </summary>
+        /// <typeparam name="TDependentEntity">
+        ///     The entity type that is the dependent in this relationship (the type that has the foreign key
+        ///     properties).
+        /// </typeparam>
+        /// <param name="foreignKeyPropertyNames">
+        ///     The name(s) of the foreign key property(s).
+        /// </param>
+        /// <returns> The same builder instance so that multiple configuration calls can be chained. </returns>
+        public virtual ReferenceReferenceBuilder HasForeignKey<TDependentEntity>(
```
anpete: Should this have a class constraint to match the other APIs in CF?

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/Metadata/Builders/ReferenceReferenceBuilder.cs
ajcvickers: Will do.

test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/ModelBuilderGenericTest.cs
```diff
@@ -314,9 +314,15 @@ public GenericTestReferenceReferenceBuilder(ReferenceReferenceBuilder<TEntity, T
             public override TestReferenceReferenceBuilder<TEntity, TRelatedEntity> HasForeignKey(Type dependentEntityType, params string[] foreignKeyPropertyNames)
```
AndriySvyryd: The Type overload should be removed from base. The current implementation would become the implementation of the new overload in ModelBuilderNonGenericTest

test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/ModelBuilderNonGenericUnqualifiedStringTest.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Linq.Expressions;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace Microsoft.EntityFrameworkCore.Tests
+{
+    public class ModelBuilderNonGenericUnqualifiedStringTest : ModelBuilderNonGenericTest
+    {
+        public class NonGenericStringOneToManyType : OneToManyTestBase
```
AndriySvyryd: Why the Type suffix?

test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/OneToOneTestBase.cs
```diff
@@ -2324,39 +2324,72 @@ public virtual void Throws_on_duplicate_navigation_when_self_referencing()
                         modelBuilder.Entity<SelfRef>().HasOne(e => e.SelfRef1).WithOne(e => e.SelfRef1)).Message);
             }
 
-            [Fact]
-            public virtual void Throws_if_specified_FK_types_do_not_match()
+            [Theory]
+            [InlineData(true)]
+            [InlineData(false)]
+            public virtual void Throws_if_specified_FK_types_do_not_match(bool useGeneric)
```
AndriySvyryd: These should remain Facts, the variations are handled at class level

test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/ModelBuilderNonGenericUnqualifiedStringTest.cs
ajcvickers: Don't know how that got there...

test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/ModelBuilderGenericTest.cs
ajcvickers: Will do.

test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/ModelBuilderNonGenericUnqualifiedStringTest.cs
ajcvickers: It was in the parallel class to this. Looks like it refers to passing a Type as opposed to a string as the entity type name.

ajcvickers:
@AndriySvyryd Pushed an update.

test/Microsoft.EntityFrameworkCore.Tests/ModelBuilderTest/ModelBuilderNonGenericTest.cs
```diff
@@ -275,11 +275,11 @@ public NonGenericTestReferenceReferenceBuilder(ReferenceReferenceBuilder referen
...
 
-            public override TestReferenceReferenceBuilder<TEntity, TRelatedEntity> HasForeignKey(Type dependentEntityType, params string[] foreignKeyPropertyNames)
-                => Wrap(ReferenceReferenceBuilder.HasForeignKey(dependentEntityType, foreignKeyPropertyNames));
+            public override TestReferenceReferenceBuilder<TEntity, TRelatedEntity> HasForeignKey<TDependentEntity>(params string[] foreignKeyPropertyNames)
+                => Wrap(ReferenceReferenceBuilder.HasForeignKey<TDependentEntity>(foreignKeyPropertyNames));
 
-            public override TestReferenceReferenceBuilder<TEntity, TRelatedEntity> HasPrincipalKey(Type principalEntityType, params string[] keyPropertyNames)
-                => Wrap(ReferenceReferenceBuilder.HasPrincipalKey(principalEntityType, keyPropertyNames));
+            public override TestReferenceReferenceBuilder<TEntity, TRelatedEntity> HasPrincipalKey<TPrincipalEntity>(params string[] keyPropertyNames)
+                => Wrap(ReferenceReferenceBuilder.HasPrincipalKey<TPrincipalEntity>(keyPropertyNames));
```
AndriySvyryd: Shouldn't this implementation use the type overload?

AndriySvyryd: NVM

AndriySvyryd:
:shipit:

Issue #5717, opened 2016-06-10
Incorrect type mapping chosen for parameter


roji:
After upgrading to rc3 in the Npgsql EFCore provider, test `QueryTestBase.Where_concat_string_int_comparison3` suddenly started failing. Digging around, I traced this to what seems to be a bug in EFCore itself: the DbParameter created for the variable `i`, which is an int, is created with the type mapping for string (nvarchar in SqlServer) rather than int. This doesn't cause the SqlServer test to fail, but Npgsql always sets the NpgsqlDbType on parameter, so the `@__i_0` parameter is sent with NpgsqlDbType text. When PostgreSQL sees SQL trying to add a string and an int it raises an error.

Digging a bit further upon the [DefaultQuerySqlGenerator code that adds the parameters](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs#L1150):

```c#
_relationalCommandBuilder.AddParameter(
	parameterExpression.Name,
	parameterName,
	_typeMapping ?? _relationalTypeMapper.GetMapping(parameterExpression.Type),
	parameterExpression.Type.IsNullableType());
```

I'm not sure exactly what's going, but it seems that the type mapping inferred from the query context  (which is string) is preferred over the mapping that corresponds to the parameter itself. It seems more correct to set the parameter's type based on the actual type of the value set on it.

This is quite critical for Npgsql to work correctly. Am not sure what changes happened in rc3 to trigger this.

divega:
@roji, good catch, that is a fun one! I didn't even think that the type of the expression and the value could not match,  but it makes sense.

divega:
@roji we discussed a couple of alternative approaches for a fix in triage but ultimately decided to move the issue to 1.0.1 because we don't think it meets the bar for 1.0.1. 

For reference, if a similar issue affected users of SQL Server but had a workaround (as this one does: manually convert assign the integer values to string variables before the query) at this stage we would be moving it to 1.0.1 as well. We hope to fix it shortly after RTM.

roji:
@divega yeah, PostgreSQL is in general very strongly-typed and strict, it's a good laboratory for this kind of thing.

No problem and thanks for the quick attention, this can wait for 1.0.1.

PR #6019, opened 2016-07-07
[1.0.1] Fix: #5717 - Incorrect type mapping chosen for parameter

anpete:
Ensure we update type mapping stack when visiting ExplicitCastExpression.


anpete:
@ajcvickers

ajcvickers:
:shipit:

PR #6020, opened 2016-07-07
[1.0.1] Fix to #5738 - Query Perf: Selecting entire table for navigation properties in projections

### LINKED ISSUE #5738: Query Perf: Selecting entire table for navigation properties in projections

maumar:
Fixes #5738 

would be translated to:

This would previously produce outer query to fetch all the customers and for each customer another query, fetching ALL the orders each time filtering out the orders for each customer and returning only the first one on the client (because we didn't know how to handle o.CustomerId in this case)

SELECT [t].[CustomerID]
FROM [Customers] AS [c]

FROM [Orders] AS [o]

We achieve this by introducing a new query method that creates additional parameters before enumerating it's source, and then removing those parameters after all elements have been enumerated or the enumerator has been disposed.
This way we make sure that the additional parameters only exist in the scope that the inner query is executed.

This also forces some changes in the CommandCache - we used QueryContext.ParameterValues as cache key, but since now the parameter values are modified during query execution we need to rely on cloning them at and used the cloned values as the cache key.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
@@ -713,5 +713,100 @@ namespace Microsoft.EntityFrameworkCore.Query
 
             public void Dispose() => _includeCollectionIterator?.Dispose();
         }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual MethodInfo InjectParametersMethod => _injectParametersMethodInfo;
+
+        private static readonly MethodInfo _injectParametersMethodInfo
+            = typeof(AsyncQueryMethodProvider)
+                .GetTypeInfo().GetDeclaredMethod(nameof(_InjectParameters));
+
+        [UsedImplicitly]
+        // ReSharper disable once InconsistentNaming
+        private static IAsyncEnumerable<TElement> _InjectParameters<TElement>(
+            QueryContext queryContext,
+            IAsyncEnumerable<TElement> source,
+            string[] parameterNames,
+            object[] parameterValues)
+            => new ParameterInjector<TElement>(queryContext, source, parameterNames, parameterValues);
+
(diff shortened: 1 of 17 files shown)


maumar:
@anpete

PR #6021, opened 2016-07-07
Separate SqlServer code in optimistic concurrency tests

roji:
The optimistic concurrency test base classes contained some SqlServer-specific code, i.e. the setup of a byte[] rowversion as the concurrency column. Moved that out to the SqlServer tests.

@AndriySvyryd, this is a continuation of the discussion in #5367 (thanks for fixing those the tests in #5608.

Also just to let you guys know (@rowanmiller, @divega, @ajcvickers), I've added first-class optimistic concurrency support to the Npgsql provider, using PostgreSQL's xmin column. This is a special system column that all tables have, and which contains the id of the last transaction that touched the row ([see the docs](https://www.postgresql.org/docs/current/static/ddl-system-columns.html)). Challenges included uint support, and also making Npgsql's ModelDiffer totally ignore system columns and never produce migrations for them (as they're always there and unchangeable).

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/F1FixtureBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/F1FixtureBase.cs
@@ -20,16 +20,6 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             modelBuilder.Entity<Chassis>(b =>
                 {
                     b.HasKey(c => c.TeamId);
-                    b.Property<byte[]>("Version")
-                        .ValueGeneratedOnAddOrUpdate()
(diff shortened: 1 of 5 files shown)


AndriySvyryd:
@roji Don't forget to update `OptimisticConcurrencySqliteTest`

roji:
@AndriySvyryd you're right, just push-forced a fix. Although there's not much going on there :)

BTW it seems like at the very least optimistic concurrency should be possible in sqlite with non-value-generated-on-update concurrency tokens.

Let me know if I've missed anything else.

AndriySvyryd:
@roji We'll update the SQLite tests when we add better API for handling optimistic concurrency

PR #6022, opened 2016-07-07
[1.0.1] Fix to #5738 - Query Perf: Selecting entire table for navigation properties in projections

maumar:

would be translated to:

This would previously produce outer query to fetch all the customers and for each customer another query, fetching ALL the orders each time filtering out the orders for each customer and returning only the first one on the client (because we didn't know how to handle o.CustomerId in this case)

SELECT [t].[CustomerID]
FROM [Customers] AS [c]

FROM [Orders] AS [o]

We achieve this by introducing a new query method that creates additional parameters before enumerating it's source, and then removing those parameters after all elements have been enumerated or the enumerator has been disposed.
This way we make sure that the additional parameters only exist in the scope that the inner query is executed.

This also forces some changes in the CommandCache - we used QueryContext.ParameterValues as cache key, but since now the parameter values are modified during query execution we need to rely on cloning them at and used the cloned values as the cache key.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
@@ -713,5 +713,87 @@ namespace Microsoft.EntityFrameworkCore.Query
 
             public void Dispose() => _includeCollectionIterator?.Dispose();
         }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual MethodInfo InjectParametersMethod => _injectParametersMethodInfo;
+
+        private static readonly MethodInfo _injectParametersMethodInfo
+            = typeof(AsyncQueryMethodProvider)
+                .GetTypeInfo().GetDeclaredMethod(nameof(_InjectParameters));
+
+        [UsedImplicitly]
+        // ReSharper disable once InconsistentNaming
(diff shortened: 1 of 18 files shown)


maumar:
@anpete

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -713,5 +713,100 @@ public IAsyncEnumerable<EntityLoadInfo> Load(QueryContext queryContext, IInclude
...
+                }
+
+                public TElement Current => _innerEnumerator.Current;
+
+                public async Task<bool> MoveNext(CancellationToken cancellationToken)
+                {
+                    var result = await _innerEnumerator.MoveNext(cancellationToken);
+                    if (!result)
+                    {
+                        RemoveParameters();
```
anpete: It should be safe to just rely on Dispose to do this.

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -713,5 +713,100 @@ public IAsyncEnumerable<EntityLoadInfo> Load(QueryContext queryContext, IInclude
...
+                }
+
+                public void Dispose()
+                {
+                    if (!_parametersRemoved)
+                    {
+                        RemoveParameters();
+                    }
+
+                    _innerEnumerator?.Dispose();
```
anpete: This can't be null, right?

anpete:
Looking good. :shipit: after feedback and R#

rowanmiller:
We decided this is too high risk for 1.0.1, so just push it into dev

Issue #5254, opened 2016-05-02
Mixing sync and async APIs on SqlClient for Linux overwhelms connection pooling


natemcmaster:
The test re-enabled in https://github.com/aspnet/EntityFramework/pull/5062 fails on Linux consistently. We should investigate if this is a bug in SqlClient or in our test setup.

```
Can_use_sequence_end_to_end_from_multiple_contexts_concurrently_async
```

natemcmaster:
This may be an issue in SqlClient connection pooling on Linux.

Test setup:
50 tasks created. Each task opens a connection to SQL Server to request the next sequence value for 20 items. Tasks run concurrently. On Windows, .NET Core and .NET Framework, this test always passes just fine. On Linux, this test is flaky. It fails about 50% of the time on our CI server:
```
System.Data.SqlClient.SqlException : A network-related or instance-specific error occurred while
 establishing a connection to SQL Server. The server was not found or was not accessible. Verify 
that the instance name is correct and that SQL Server is configured to allow remote connections. 
(provider: TCP Provider, error: 40 - Could not open a connection to SQL Server)
```

Here is the log output for Linux and Windows. Log output represents actual order of execution. Each line identified with the task number.

[Windows_log.txt](https://github.com/aspnet/EntityFramework/files/273352/Windows_log.txt)
[Linux_log.txt](https://github.com/aspnet/EntityFramework/files/273351/Linux_log.txt)


As you can see in the Windows log, EF only opens about 12 connections at a time. It appears pooling is kicking it to prevent overwhelming the server. 

In the Linux log, EF tries to open hundreds of connections without waiting for currently open connections to close.

cc @saurabh500 are there known issues with SqlClient connection pooling on Linux?

saurabh500:
@natemcmaster  we havent seen such an issue for SqlClient on Linux. This seems like a networking issue. The error is seen when SqlClient cannot open a TCP connection to the Sql Server. 

Is this the only use case you are seeing this? Or can this be reproed for any other test case as well? 


cc @corivera

saurabh500:
More questions 
1. Whats the sql server version? 
2. Do you see this without pooling as well?

natemcmaster:
Made a minimal repro without EF and opened an issue: https://github.com/dotnet/corefx/issues/8701

saurabh500:
Will take a look at it. Thanks for the minimal repro.

natemcmaster:
[...]

Issue #2291, opened 2015-05-29
Support AddAsync && AddRangeAsync


davidroth:
If EF7 is configured with GenerateValueOnAdd(true), sql server sequences are used, and so every time Add(entity) is called, EF7 will send a query to the store so that it can assign a new key to the entity.

I remember I have seen AddAsync && AddRangeAsync overloads in previous Ef7 releases. Why did you remove them?

I found some notes here: https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes:-January-16,-2015 (*"Update: we decided to remove AddAsync entirely"*) but i could not find any reasoning for this change.

divega:
Good question! Sorry we left the details out of the notes. 

The removal of `AddAsync()` was a compromise and (as usual) one that we are very open to get feedback on.

Our understanding of how to best apply the Task-based async pattern to API designs is still somewhat evolving. Despite how powerful the support for async programming is in C#, synchronous APIs are still easier to use correctly in many circumstances. In particular, [it is hard to use async APIs correctly when from synchronous code](http://stackoverflow.com/questions/9343594/how-to-call-asynchronous-method-from-synchronous-method-in-c) :smile: 

In principle we create async versions of APIs to enable server applications to scale better by saving threads that would otherwise be wasted waiting for I/O operations to be completed. 

Following that principle we originally introduced `AddAsync()`in the very first prototype of EF7 to handle asynchronous implementations of value generators. But when we later on started fleshing out the rest of the API we found additional places where we had to either call into `AddAsync()` or implement the same behavior as `AddAsync()`, among others:

- Setting the state of an entity to Added
- Calling DetectChanges() on a graph containing new entities 
- Assigning a new entity to a navigation property of an entity we are tracking

Suddenly async had become “viral”: any time you use `await` the whole call stack has to be async, so we needed/wanted `ChangeStateAsync()`, `DetectChangesAsync()` (AFAIR we actually implemented those two), but how do we figure out a way to make setters on navigation properties be async?

It just looked like a slippery slope in which we would end up making our API much more complex in exchange for very little value. (After all, correctly implemented value generators should not be chatty on the network. E.g. we use a default value block size of 10, meaning that only once every ten calls to `Add()` will actually cause a network round trip. If this isn’t good enough, you can increase the block size or switch to a client-based value generator.) We didn't want to go down the path of adding all async versions of all those methods and hence we decided to remove it entirely. Did we make the best choice? We don't know with certainty. 

[...]

PR #6023, opened 2016-07-07
Bring back AddAsync

ajcvickers:
But only AddAsync and not everything else that was virally async by association. This allows a path by which an application can ensure that database calls are all async even when using value generators that make calls to the database.

Issue #2291, #5254


AndriySvyryd:
:shipit:

Issue #5522, opened 2016-05-25
Model with nullable FK fails parent child query with "Argument Types do not match"


brianro:
**When doing this, also make sure the scenario from #5883 is taken care of - dotting thru an optional nav to get a value**

If you have a model (see attached project) with class A with a nullable id field for child class E - int? EId, querying A->E fails with "argument types do not match" error.

```c#
            var item =
                (from a
                     in db.As
                 select new A
                 {
                     Id = a.Id,
                     EId = a.EId,
                     E = new E
                     {
                         Id = a.E.Id
                     }
                 }
                     ).First();
```

In attached project query_ef_test_ac fails with the following information:

Test Name:	BugDemoEF.EFTest.query_ef_test_ae
Test FullName:	BugDemoEF.EFTest.query_ef_test_ae
Test Source:	C:\Code\General\BugDemo\src\BugDemoEF\EFTests.cs : line 205
Test Outcome:	Failed
Test Duration:	0:00:01.976

Result StackTrace:	
at System.Linq.Expressions.Expression.Bind(MemberInfo member, Expression expression)
   at System.Linq.Expressions.ExpressionVisitor.Visit[T](ReadOnlyCollection`1 nodes, Func`2 elementVisitor)
   at System.Linq.Expressions.ExpressionVisitor.VisitMemberInit(MemberInitExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitMemberAssignment(MemberAssignment node)
   at System.Linq.Expressions.ExpressionVisitor.Visit[T](ReadOnlyCollection`1 nodes, Func`2 elementVisitor)
   at System.Linq.Expressions.ExpressionVisitor.VisitMemberInit(MemberInitExpression node)
   at Remotion.Linq.Clauses.SelectClause.TransformExpressions(Func`2 transformation)
   at Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.Rewrite(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.OptimizeQueryModel(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.CreateQueryExecutor[TResult](QueryModel queryModel)
[...]

PR #6031, opened 2016-07-08
Fix to #5522 - Model with nullable FK fails parent child query with "Argument Types do not match"

maumar:
Problem was that during navigation rewrite we sometimes change types of expressions, e.g.

o.Customer.Id (originally int)

would get converted to:

(o != null) ? (int?)o.CustomerId : null (change type to int?)

We try to compensate for this later, by casting back to the original type, but we missed some cases: MemberAssignment, ElementInit, NewArray.

Fix is to add the compensation for those nodes.


maumar:
@anpete

PR #6032, opened 2016-07-08
[1.0.1] Fix to #5522 - Model with nullable FK fails parent child query with "Argument Types do not match"

maumar:
Problem was that during navigation rewrite we sometimes change types of expressions, e.g.

o.Customer.Id (originally int)

would get converted to:

(o != null) ? (int?)o.CustomerId : null (change type to int?)

We try to compensate for this later, by casting back to the original type, but we missed some cases: MemberAssignment, ElementInit, NewArray.

Fix is to add the compensation for those nodes.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -398,6 +398,53 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        protected override MemberAssignment VisitMemberAssignment(MemberAssignment node)
+        {
+            var newExpression = CompensateForNullabilityDifference(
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

maumar:
@smitpatel

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -927,6 +973,20 @@ private ConstantExpression CreateEntityQueryable(IEntityType targetEntityType)
...
             => new EntityQueryable<TResult>(entityQueryProvider);
 
+        private static Expression CompensateForNullabilityDifference(Expression expression, Type originalType)
+        {
+            var newType = expression.Type;
+
+            var needsTypeCompensation = (originalType != newType)
+                && !originalType.IsNullableType()
+                && newType.IsNullableType()
+                && (originalType.UnwrapNullableType() == newType.UnwrapNullableType());
```
smitpatel: `originalType` is always same as `originalType.UnwrapNullableType()` here.

maumar: will fix

smitpatel:
:shipit:

Issue #5998, opened 2016-07-06
Add non-generic Find to context


ajcvickers:
In the old stack we had a non-generic Set method on the context, which then allowed non-generic Find to be used. We decided previously against non-generic Set for good reasons, and I don't think we should change that. However, we should consider having a non-generic Find on the context. This is useful for things like generic repositories and other generic context use.

```C#
public virtual object Find(Type entityType, params object[] keyValues)

public virtual Task<object> FindAsync(Type entityType, params object[] keyValues)

public virtual Task<object> FindAsync(Type entityType, object[] keyValues, CancellationToken cancellationToken)
```

For completeness we could also add generic versions:

```C#
public virtual TEntity Find<TEntity>(params object[] keyValues)

public virtual Task<TEntity> FindAsync<TEntity>(params object[] keyValues)

public virtual Task<TEntity> FindAsync<TEntity>(object[] keyValues, CancellationToken cancellationToken)
```

PR #6045, opened 2016-07-11
Add non-generic Find to DbContext

ajcvickers:
Issue #5998

Moved Find code into its own service and re-factored so it could also be used in a non-generic way.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Internal/IEntityFinder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Internal/IEntityFinder.cs
@@ -1,0 +1,28 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public interface IEntityFinder
+    {
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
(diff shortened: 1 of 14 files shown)


src/Microsoft.EntityFrameworkCore/Internal/IEntityFinder.cs
```diff
@@ -0,0 +1,34 @@
...
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        object Find([NotNull] params object[] keyValues);
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        Task<object> FindAsync([NotNull] params object[] keyValues);
```
anpete: Remove this and use a default value for cancellationToken.

src/Microsoft.EntityFrameworkCore/Internal/IEntityFinder`.cs
```diff
@@ -0,0 +1,35 @@
...
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        new TEntity Find([NotNull] params object[] keyValues);
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        new Task<TEntity> FindAsync([NotNull] params object[] keyValues);
```
anpete: Remove this and use a default value for cancellationToken.

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/Internal/IEntityFinder`.cs
ajcvickers: Presumably you mean by making the API non-params?

PR #6049, opened 2016-07-11
Split relationships with the same ForeingKey attribute on both sides.

### LINKED ISSUE #6017: StackOverflowException in dotnet.exe when creating migrations with one-to-one relationship configured incorrectly
When configuring my models using the following setup (simplified):

```
public class Independent {

    [Key]
    public int ID { get; set; }

    [ForeignKey("ID")]
    public virtual Dependent Dependent { get; set; }
}

public class Dependent {

    [Key]
    public int ID { get; set; }

    [ForeignKey("ID")]
    public Independent Independent { get; set; }
}
```

and running Add-Migration, rather than giving a warning regarding ambiguity in which side is dependent, dotnet throws an exception:

> Unhandled exception at 0x00007FFF798C97DE (coreclr.dll) in dotnet.exe: 0xC00000FD: Stack overflow (parameters: 0x0000000000000001, 0x000000AC03A75FF8).

AndriySvyryd:
Fixes #6017

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/DataAnnotationTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/DataAnnotationTestBase.cs
@@ -1025,6 +1025,64 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
+        public virtual void ForeignKey_to_ForeignKey_same_name()
+        {
+            var modelBuilder = CreateModelBuilder();
+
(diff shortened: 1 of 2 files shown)


smitpatel:
This would mean that user cannot specify foreign key attribute on both navigations even if the values are referring to same property(-ies). While specifying on both navigations is unnecessary but have seen some folks doing it. Could change model for them.
Otherwise :shipit:

AndriySvyryd:
@smitpatel I added another test with the property only on one side. Both scenarios were broken prior to this change, so it should be safe.

PR #6050, opened 2016-07-11
Fix running tests from VS by adding redirects for Castle.Core and Newtonsoft.Json

AndriySvyryd:


AndriySvyryd:
@natemcmaster

ajcvickers:
:shipit:

PR #6058, opened 2016-07-12
New method of cleaning databases to improve test perf

ajcvickers:
context.Database.EnsureClean method for SQLite and SQL Server.

Makes use of the reverse engineering code to query for database structure and then uses Migrations to generate SQL  to drop all the structure. Doesn't handle everything, but handles what we need for 99% of our tests.

Results from my machine (VS/TestDriven.NET)
- Before change:
 - Sqlite.FunctionalTests: 72.02 seconds
 - SqlServer.FunctionalTests: 272.74 seconds

- After change:
 - Sqlite.FunctionalTests: 71.21 seconds
 - SqlServer.FunctionalTests: 165.66 seconds

So no significant change for SQLite, but big improvement for SQL Server

Overall:
========== Total Tests: 11925 passed, 0 failed, 10 skipped, took 448.80 seconds ==========


anpete:
Hah! Is this our first Migrations FT? :stuck_out_tongue:

roji:
Wow, nice idea. Npgsql tests will definitely benefit from this as creating a database is a pretty heavy operation in PostgreSQL...

AndriySvyryd:
:shipit:

PR #6060, opened 2016-07-12
Update database model factory to allow running with existing connection

ajcvickers:
Fixes issue on C.I. where connection string doesn't have enough information to create a new connection. Added new internal interface to avoid breaking publicly shipped existing API.


AndriySvyryd:
:shipit:

Issue #6040, opened 2016-07-11
Queries projecting entities incorrectly try to inline parts of their projection even when client side projection of those elements is required


espentveit:
### Steps to reproduce
```
var list =
	context.FlexLmLicenseFiles
	.Select(
	 a => new FeatureVM
	 {
		 ExpirationDate = a.Expires,
		 FeatureName = a.Feature,
		 LicenseFile = a.Service.LicenseFilePath,
		 LicenseServer = a.Service.Server.PrincipalName,
		 Location = a.Service.Server.PrincipalName,
		 Port = a.Service.PortNumber,
		 Id = a.Id,
		 Vendor = a.Vendor,
		 Version = a.Version,
		 Capacity = a.Capacity.OrderByDescending(b => b.DateTime).Select(b => b.Value).FirstOrDefault(),
		 Used = a.Usage.OrderByDescending(b => b.DateTime).Select(b => b.Value).FirstOrDefault()
	 }
	).ToList();
```
### The issue

The system should generate a query that selects the data within an acceptable time frame.

Tested with Linq to SQL (EF6). Gives a working query:
```
SELECT [t0].[Id], [t0].[Feature] AS [FeatureName], (
    SELECT TOP (1) [t3].[Value]
    FROM [FlexLmCapacities] AS [t3]
    WHERE [t3].[LicenseFileId] = [t0].[Id]
    ORDER BY [t3].[DateTime] DESC
    ) AS [Capacity], (
    SELECT TOP (1) [t4].[Value]
    FROM [FlexLmUsages] AS [t4]
    WHERE [t4].[LicenseFileId] = [t0].[Id]
    ORDER BY [t4].[DateTime] DESC
    ) AS [Used], [t0].[Expires] AS [ExpirationDate], [t0].[Vendor], [t2].[PrincipalName] AS [LicenseServer], [t1].[LicenseFilePath] AS [LicenseFile], [t1].[PortNumber] AS [Port], [t0].[Version]
FROM [FlexLmLicenseFiles] AS [t0]
LEFT OUTER JOIN [FlexLmServices] AS [t1] ON [t1].[Id] = [t0].[ServiceId]
LEFT OUTER JOIN [FlexLmServers] AS [t2] ON [t2].[Id] = [t1].[ServerId]

```

Tested with EF Core:
Output:
```
warn: Microsoft.EntityFrameworkCore.Query.Internal.SqlServerQueryCompilationContextFactory[8]
[...]

PR #6062, opened 2016-07-13
Fix to #6040 - Queries projecting entities incorrectly try to inline parts of their projection even when client side projection of those elements is required

maumar:
Problem was that during Projection translation we would try to inline single element subqueries into the parent query. However, in case the query source was mapped to Entity as opposed to ValueBuffer we were not modifying the selector expression afterwards.
This resulted in the same subquery being processed again, later in the pipeline, producing redundant and confusing query plans (also the inlined queries were never used in the final result)

Fix is to only inline queries for which we know how to modify the selector (i.e. single element that can be extracted from a ValueBuffer)


maumar:
@anpete @smitpatel

anpete:
:shipit:

Issue #6067, opened 2016-07-13
New collection navigation is not fixed up after reparenting


AndriySvyryd:
```C#
var category = db.Set<Product>().Include(c => c.Category).ToList().First().Category;

Assert.Equal(2, category.Products.Count);

var category2 = db.Set<Category>().ToList().Single(a => a != category);

Assert.Equal(0, category2.Products.Count);

var product = category.Products.First();
category.Products.Remove(product);
product.Category = category2;
category2.Products.Add(product);
Assert.Equal(category.Id, product.CategoryId);

db.ChangeTracker.DetectChanges();

Assert.Equal(category2.Id, product.CategoryId);
Assert.Equal(category, category.Products.Single().Category);
Assert.Equal(category2, category2.Products.Single().Category); // Throws
```

PR #6068, opened 2016-07-13
Test for #6067

AndriySvyryd:
@ajcvickers

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.Tests/ChangeTracking/Internal/FixupTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.Tests/ChangeTracking/Internal/FixupTest.cs
@@ -3194,6 +3194,43 @@ namespace Microsoft.EntityFrameworkCore.Tests.ChangeTracking.Internal
             }
         }
 
+        //Issue #6067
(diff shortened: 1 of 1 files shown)


test/Microsoft.EntityFrameworkCore.Tests/ChangeTracking/Internal/FixupTest.cs
```diff
@@ -3194,6 +3194,43 @@ public void Add_principal_but_not_dependent_one_to_one_no_navs_FK_set_no_navs_se
             }
         }
 
+        //Issue #6067
+        //[Fact]
```
anpete: Need [Fact]?

AndriySvyryd: Would break the build as the test fails

anpete: Ah, mistakenly thought we were verifying the issue was fixed :relaxed:

ajcvickers:
:shipit:

Issue #5369, opened 2016-05-15
Query: Support for String.CompareTo


Villason:
Might be related to #1767.

I have some classes that returns expressions that afterwards are compiled into a final expression. This code does not translates to SQL:

```
        public override Expression<Func<Comment, bool>> GetAfterExpression(Comment after)
        {
            if (IsDescending)
            {
                return m => m.Path.CompareTo(after.Path) > 0;
            }
            else
            {
                return m => m.Path.CompareTo(after.Path) < 0;
            }
        }
```

The warning is:

```
warn: Microsoft.Data.Entity.Query.Internal.SqlServerQueryCompilationContextFactory[0]
      The LINQ expression '([f].Path.CompareTo(__after_Path_1) < 0)' could not be translated and will be evaluated locally.
```

Is this the expected behavior?

Update:

If I switch to this:

```
        public override Expression<Func<Comment, bool>> GetAfterExpression(Comment after)
        {
            if (IsDescending)
            {
                return m => string.Compare(m.Path, after.Path) > 0;
            }
            else
            {
                return m => string.Compare(m.Path, after.Path) < 0;
            }
        }
```
I get an exception with this bug: #3464

```
fail: Microsoft.AspNet.Diagnostics.DeveloperExceptionPageMiddleware[0]
      An unhandled exception has occurred while executing the request
      System.InvalidOperationException: Extension node must override the property Expression.Type.
         at System.Linq.Expressions.Expression.get_Type()
         at System.Linq.Expressions.Expression.AndAlso(Expression left, Expression right, MethodInfo method)
         at Microsoft.Data.Entity.Query.ExpressionVisitors.SqlTranslatingExpressionVisitor.VisitBinary(BinaryExpression binaryExpression)
         at System.Linq.Expressions.BinaryExpression.Accept(ExpressionVisitor visitor)
         at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
         at Microsoft.Data.Entity.Query.ExpressionVisitors.SqlTranslatingExpressionVisitor.Visit(Expression expression)
         at Microsoft.Data.Entity.Query.RelationalQueryModelVisitor.VisitWhereClause(WhereClause whereClause, QueryModel queryModel, Int32 index)
         at Remotion.Linq.Clauses.WhereClause.Accept(IQueryModelVisitor visitor, QueryModel queryModel, Int32 index)
[...]

PR #6070, opened 2016-07-13
Query: Add test for string.Compare where argument is not constant

smitpatel:
Test for #5369 
we already have support for string.Compare. For the particular query it failed because the argument was property of other object. In current code, that is converted to parameter hence correctly translating to SQL.


maumar:
:shipit:

Issue #5939, opened 2016-07-01
Queries with "Contains" broken after upgrade from RC2 to RTM


dvdobrovolskiy:
Query with lots of Where now broken and generates "Incorrect syntax near the keyword 'IN'." error.

Query is:
```
            model.Zayavkas = _context.Zayavkas.Select(s=> new Zayavka
            {
                Apartment = s.Apartment,
                ApartmentId = s.ApartmentId,
                Block = _context.Blocks.SingleOrDefault(d => d.BlockId == s.BlockId),
                BlockId = s.BlockId,
                Description = s.Description,
                DispatcherPhone = s.DispatcherPhone,
                EndDate = s.EndDate,
                Reason = _context.Reasons.SingleOrDefault(b => b.ReasonId == s.ReasonId),
                ReasonId = s.ReasonId,
                Reasontmp = s.Reasontmp,
                Responsable = _context.Users.SingleOrDefault(b => b.Id == s.ResponsableId),
                ResponsableId = s.ResponsableId,
                StartDate = s.StartDate,
                Worker = _context.Users.SingleOrDefault(b => b.Id == s.WorkerId),
                WorkerId = s.WorkerId,
                ZayavkaId = s.ZayavkaId,
                ZayavkaStatus = s.ZayavkaStatus,
                CreatedDate = s.CreatedDate,
                CreatorId = s.CreatorId,
                EditedDate = s.EditedDate,
                ModifierId = s.ModifierId,
                Creator = _context.Users.SingleOrDefault(c=>c.Id == s.CreatorId),
                Modifier = _context.Users.SingleOrDefault(m => m.Id == s.ModifierId),
                MainPhone = s.MainPhone,
                SecondaryPhone = s.SecondaryPhone,
                TertiaryPhone = s.TertiaryPhone
            })
            .Where(w => model.status.Count() > 0 ? model.status.Contains(w.ZayavkaStatus) : true)
            .Where(w => model.type.Count() > 0 ? model.type.Contains(_context.Reasons.SingleOrDefault(b => b.ReasonId == w.ReasonId).ZayavkaType) : true)
            .Where(w => model.BlockId.Count() > 0 ? model.BlockId.Contains(w.BlockId) : true)
            .Where(w => Responsable ? w.ResponsableId == user.Id: true)
            .Where(w => Worker ? w.WorkerId == user.Id : true)
            .Where(w => w.CreatedDate >= _start)
            .Where(w => w.CreatedDate < _end )
[...]

PR #6074, opened 2016-07-13
[SMALL] Fix to #5939 - Queries with "Contains" broken after upgrade from RC2 to RTM

maumar:
Problem was that we were not translating search expressions into non-search expressions for Extension expressions.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -1745,9 +1745,22 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
                     _isSearchCondition = false;
                     var newExpression = base.VisitExtension(expression);
                     _isSearchCondition = parentIsSearchCondition;
+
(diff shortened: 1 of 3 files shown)


maumar:
@smitpatel

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1745,10 +1745,26 @@ protected override Expression VisitExtension(Expression expression)
...
+
                     return expression is AliasExpression || expression is ColumnExpression || expression is SelectExpression
                         ? Expression.Equal(newExpression, Expression.Constant(true, typeof(bool)))
                         : newExpression;
                 }
+                else
+                {
+                    if (IsSearchCondition(expression))
+                    {
+                        var parentIsSearchCondition = _isSearchCondition;
```
anpete: Anyway to restructure logic to remove the duplication here?

maumar: was thinking about it also, will try

PR #6079, opened 2016-07-14
Refactor dotnet-ef to use new architecture

natemcmaster:
Refactors how dotnet-ef invokes the "inside-man". It calls "/tools/netcoreapp1.0/ef.dll" instead of using the ProjectDependencyCommandFactory invocation.

Part of #5334. NB: merges into feature branch.

Changes:

- Change dotnet-ef to invoke "ef.dll"
- Rename project to Tools.DotNet.
- Update dotnet-ef end to end tests so they can run more reliably on CI.

TODO:
 - ~~Invoke ef.exe/ef.dll from PowerShell~~ see https://github.com/aspnet/EntityFramework/pull/6112
 - Add more testing
 - xmldoc

cc @bricelam


src/Microsoft.EntityFrameworkCore.Design/Design/OperationExecutor.cs
```diff
@@ -441,11 +441,10 @@ public virtual void Execute<T>([NotNull] Func<IEnumerable<T>> action)
     }
 
 #if NET451
-    partial class OperationExecutor : MarshalByRefObject
+    public partial class OperationExecutor : MarshalByRefObject
```
bricelam: FYI, I leave these off of supplemental partials because DRY.

src/Tools.DotNet/Internal/ICommandResolver.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.DotNet.Cli.Utils;
+
+namespace Microsoft.EntityFrameworkCore.Tools.DotNet.Internal
+{
+    public interface ICommandResolver
```
bricelam: This type seems unnecessary. YAGNI

src/Tools.Shared/OperationHandlers.cs
```diff
@@ -92,53 +47,59 @@ public partial class OperationLogHandler : IOperationLogHandler
...
-    }
-
-    partial class OperationLogHandler : MarshalByRefObject
-    {
-    }
+    public partial class OperationLogHandler : MarshalByRefObject
+    { }
+    public partial class OperationResultHandler : MarshalByRefObject
+    { }
 #endif
```
bricelam: I like this pattern better than `DesignMarshalByRefObject`

bricelam:
Awesome. :shipit:

src/Tools.DotNet/Internal/ICommandResolver.cs
natemcmaster: Oh yeah, forgot about this guy. I added for (yet to be implemented) testing. But I'll pull this out as there is no need for it yet.

PR #6082, opened 2016-07-14
Make Find queries parameterized

ajcvickers:
Should make more efficient use of query cache on SQL Server.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Internal/EntityFinder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Internal/EntityFinder.cs
@@ -11,6 +11,7 @@ using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Utilities;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore/Internal/KeyValues.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Internal
+{
+    internal struct KeyValues
```
anpete: internal by design?

anpete: Could we just use ValueBuffer instead?

ajcvickers: Yes. Wanted to be nested and private, but then no good place to cache the MethodInfo because Finder is generic.

ajcvickers: Maybe. But it has stuff we don't need.

anpete: Could be worth measuring. Advantage is avoids an extra type, and we already have similar code to build the get value expressions.

anpete:
:shipit:

Issue #5672, opened 2016-06-07
Query :: Include with multiple navigations (including optional navigation) fails


ImmePak:
I`am running ASP.NET Core 1.0 application, using EF7 rc2
Part of class structure:

```
public sealed class Member
{
    public Guid Id { get; set; }
    public MemberState State { get; set; }
}
public sealed class MemberState
{
     public Guid Id { get; set; }
     public Org Org { get; set; }
}
```
So, first goal was to make field Org in MemberState nullable and I configured EF context in the following way:
```
entity.Property<Guid?>("Idf_Org");
entity.HasOne(e => e.Org).WithMany().HasForeignKey("Idf_Org").IsRequired(false);
```
It was fine before i have need to search _Members_ by _Org.Id_
First approach, that worked before i marked Org field as nullable:
```
_provider.Members.Include(m => m.State).ThenInclude(s => s.Org)
        .Where(m => m.State.Org.Id == idfOrg).ToList();
```
Now it start firing exception : _Object reference not set to an instance of an object._
Considering the fact that database not contains any record of _MemberState_ with _NULL_ in _Idf_Org_ it was quite strange. 
but this one is working O_o but give me unappropriate result (all included instances is null)
```
_provider.Members.Where(m => m.State.Org.Id == idfOrg).ToList();
```
One way i found is to do direct SQL query in that way
```
_provider.Members.FromSql("SELECT mem.* "
         + "FROM [dbo].[Member] as mem "
         + "inner join [dbo].[State] as st on st.Id = mem.Idf_State "
         + "where st.Idf_Org = {0} ", idfOrg)
         .Include(m => m.State).ThenInclude(s => s.Org).ToList();
```
**The question is follows:**
1. Is this behaviour is normal to EF ?
2. Is there a way to get queries without using direct sql in this case ? In my app _MemberState_ and included object _Org_ are involved in many application parts as reports and so on and it is unappropriate  to manage all queries  with sql.

divega:
@maumar does this look familiar? In triage we discussed that the NRE may come from `m.State.Org.Id` if `.Org` is null, but we believe you improved the expansion to compensate for that. Was that after RC2?

maumar:
@ImmePak can you provide some more information? im unable to reproduce this. Ideally provide standalone repro. Otherwise please provide code for Org class and a complete code for OnModelCreating. If you are using sample data, that would be great also.

maumar:
[...]

PR #6093, opened 2016-07-15
Fix to #5672 - Query :: Include with multiple navigations (including optional navigation) fails

maumar:
Problem was that our include logic for groupjoin was expecting outer and inner elements (on which the include was being performed) to be entities. However sometimes those elements would be TransparentIdentifiers (e.g. when the element comes from a result of SelectMany).

Fix is to also store (when necessary) the accessor from the outer/inner element to the entity that we want to include, extract the actual entity in runtime and apply include operations on that entity instead of the actual outer/inner element.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/GroupJoinInclude.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/GroupJoinInclude.cs
@@ -22,6 +22,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
         private RelationalQueryContext _queryContext;
         private IRelatedEntitiesLoader[] _relatedEntitiesLoaders;
         private GroupJoinInclude _previous;
+        private Delegate _entityAccessor;
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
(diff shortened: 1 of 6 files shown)


maumar:
@anpete

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/GroupJoinInclude.cs
```diff
@@ -75,6 +76,11 @@ public virtual void Initialize([NotNull] RelationalQueryContext queryContext)
         /// </summary>
         public virtual void Include([CanBeNull] object entity)
         {
+            if (_entityAccessor != null)
+            {
+                entity = _entityAccessor.DynamicInvoke(entity);
```
maumar: any ideas on how to avoid DynamicInvoke here? (since the argument is TransparentIdentifier I can't store the accessor as Func<object, object> and I can't add type arguments to the GroupJoinInclude itself because that would be breaking change

anpete: GroupJoinInclude is internal. Let's discuss next week.

maumar:
```cs
public virtual Type GroupJoinIncludeType => typeof(GroupJoinInclude);
```

on QueryMethodProvider is public

maumar:
opening new pull request with different approach

PR #6094, opened 2016-07-16
Pull down the relationship to the derived type for any order it can be configured.

AndriySvyryd:
This bug would prevent the user from configuring a relationship on the derived type if a convention created a relationship on the base that used any of the same members.


ajcvickers:
:shipit:

Issue #6085, opened 2016-07-15
Abstract out IIncludable/Async Extensions for Generic use


rondefreitas:
### The Problem

When building and testing real-world applications with EntityFramework, it is not always possible to take advantage of the Async and IIncludable extensions.

Many application structures abstract away the ORM code behind generic repositories and services that focus on the IQueryable interface; Along the way, lose the advantage of the EntityFrameworkQueryableExtensions, both in business logic and unit tests, without jumping through sometimes significant hoops.

Developers faced with this problem are often either unable to take advantage of the Async extensions and IIncludableQueryable in their services at all, or are forced to write overly complicated wrapping code to allow these features.
### Proposal

Let's face it, no matter how many times you beat the drum that DbSet _is_ a repository and that DbContext _is_ a Unit of Work, you will never prevent organizations from feeling the need to decouple their ORM entirely from their services project.

I propose we abstract these extensions such that they are available with all instances of IQueryable regardless of the underlying QueryProvider. I'm prepared to contribute this myself, as it's a matter mostly of porting work I've done already on another project.
- Create a new Project and Nuget package for `Microsoft.EntityFrameworkCore.Abstractions` (or `....Extensions`.)
- Default implementations would be provided for extensions such as ToListAsync, which would return a task from the result of the Synchronous version of the method unless targeting an IQueryable for which the QueryProvider implements IAsyncQueryProvider.
- The existing namespaces would remain in tact. The purpose of the project would be solely for using these extensions independent of pulling in all of Entity Framework Core.
- Default implementations of IIncludableQueryable would assume that the IQueryable source is a synchronous collection for which navigation properties are already populated. (note: EF already does this for Include/ThenInclude... they could just use a move to an abstractions project
- **_Requires Discussion**_: debug log events could be generated whenever the default, synchronous implementation is used, so the developer is aware that they are not truly executing asynchronous code.

rowanmiller:
Going to wait unitl @divega is back in the office to discuss this.

rdefreitas:
Thanks @rowanmiller.

I've gone through all of the methods and have come up with a short list of the ones I don't believe are appropriate to be moved, and why.

- *Load*() / *LoadAsync*(): This is very EF specific. The only change I'd propose here is to add the fallback of calling Load if the collection does not implement IAsyncEnumerable.
[...]

PR #6099, opened 2016-07-18
Create Abstractions library for Async and IIncludable Extensions

rdefreitas:
Create Abstractions library for Async and IIncludable Extensions
Per proposal #6085

- Create project/package Microsoft.EntityFrameworkCore.Abstractions
- Microsoft.EntityFrameworkCore now takes a dependency on ...Abstractions
- Add IEntityQueryProvider marker interface to aide in determining if a query is provided by Entity Framework.
- Async Linq extensions fallback to synchronous counterparts
- move async linq methods into AsyncQueryableExtensions with non-extensions fallback in EntityFrameworkQueryableExtensions (to reduce breaking changes)
- move AsyncQueryableExtensions to System.Linq namespace (per Guiding Principles)


rdefreitas:
phew, those functional tests with the string-based build references gave me a nice headache tracking down the issue (particularly when the build takes about 20 minutes).

Reopening this PR now that I tracked down what broke the build... ;-)

rowanmiller:
cc @natemcmaster to comment on the AppVeyor issue

natemcmaster:
I triggered a rebuild manually.

src/Microsoft.EntityFrameworkCore.Abstractions/Properties/CoreStrings.resx
```diff
@@ -565,7 +565,7 @@
     <value>Cannot create a DbSet for '{typeName}' because this type is not included in the model for the context.</value>
   </data>
   <data name="AmbiguousOneToOneRelationship" xml:space="preserve">
-    <value>The child/dependent side could not be determined for the one-to-one relationship that was detected between '{dependentToPrincipalNavigationSpecification}' and '{principalToDependentNavigationSpecification}'. To identify the child/dependent side of the relationship, configure the foreign key property. See http://go.microsoft.com/fwlink/?LinkId=724062 for more details.</value>
+    <value>The child/dependent side could not be determined for the one-to-one relationship that was detected between '{dependentToPrincipalNavigationSpecification}' and '{principalToDependentNavigationSpecification}'. To identify the child/dependent side of the relationship, configure the foreign key property. See http://go.microsoft.com/fwlink/?LinkId=724062 for more details.\</value>
```
divega: Typo?

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
```diff
@@ -325,18 +277,13 @@ public static class EntityFrameworkQueryableExtensions
         ///     <paramref name="predicate" />.
         /// </returns>
         public static Task<TSource> FirstAsync<TSource>(
-            [NotNull] this IQueryable<TSource> source,
+            [NotNull] IQueryable<TSource> source,
             [NotNull] Expression<Func<TSource, bool>> predicate,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = new CancellationToken())
```
anpete: Why are these changing?

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
```diff
@@ -83,18 +75,13 @@ public static class EntityFrameworkQueryableExtensions
         ///     predicate; otherwise, <c>false</c>.
         /// </returns>
         public static Task<bool> AnyAsync<TSource>(
-            [NotNull] this IQueryable<TSource> source,
+            [NotNull] IQueryable<TSource> source,
             [NotNull] Expression<Func<TSource, bool>> predicate,
-            CancellationToken cancellationToken = default(CancellationToken))
+            CancellationToken cancellationToken = new CancellationToken())
```
ilmax: Why there are mixed `new CancellationToken()` and `default(CancellationToken)` ?

src/Microsoft.EntityFrameworkCore.Abstractions/Extensions/Internal/QueryableExtensions.cs
```diff
@@ -21,6 +23,20 @@ public static class QueryableExtensions
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        public static bool IsAsyncEnumerable<TSource>([NotNull] this IQueryable<TSource> source)
+        {
+            Check.NotNull(source, nameof(source));
+
+            var enumerable = source as IAsyncEnumerable<TSource>;
```
ilmax: can be simplified with `return source is IAsyncEnumerable<TSource>`

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
rdefreitas: In most cases it's default... apparently I got one mixed somewhere along the way, and it may have been an auto formatting changed based on resharper settings on my machine. I'll go back and fix these up.

src/Microsoft.EntityFrameworkCore.InMemory/project.json
```diff
@@ -23,6 +23,7 @@
     }
   },
   "dependencies": {
+    "Microsoft.EntityFrameworkCore.Abstractions": "1.1.0-*",
```
rdefreitas: added to support a minor defect in a related project that was intermixing EF core with 1.0 RTM libraries. This may be redundant outside of mixed-library scenarios such as that, but it doesn't hurt anything to leave it in place.

src/Microsoft.EntityFrameworkCore.Relational/project.json
```diff
@@ -25,6 +25,7 @@
     }
   },
   "dependencies": {
+    "Microsoft.EntityFrameworkCore.Abstractions": "1.1.0-*",
```
rdefreitas: added to support a minor defect in a related project that was intermixing EF core with 1.0 RTM libraries. This may be redundant outside of mixed-library scenarios such as that, but it doesn't hurt anything to leave it in place.

anpete: @rdefreitas Starting to look at this. First couple of things:

1) Microsoft.EntityFrameworkCore.csproj no longer builds. Looks like some shared files got removed.
2) CoreStrings.resx got moved. If we need resources in Abstractions, then split them out into an AbstractionsStrings.resx and link the Resources.tt file

Thanks!

src/Microsoft.EntityFrameworkCore.Abstractions/Properties/InternalsVisibleTo.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.EntityFrameworkCore, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
divega: Why is this necessary? In principle it doesn't seem right.

src/Microsoft.EntityFrameworkCore.Abstractions/Query/Internal/IEntityQueryProvider.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using System.Linq.Expressions;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.EntityFrameworkCore.Query.Internal
```
divega: It seems weird to have internal surface in an abstractions package.

rdefreitas:
@natemcmaster it seems this will need another manual kickoff before it can be reviewed again.
> Build execution time has reached the maximum allowed time for your plan (60 minutes).

Thank you, sir.

src/Microsoft.EntityFrameworkCore.Abstractions/Query/Internal/IEntityQueryProvider.cs
rdefreitas: @divega I has split out the small necessary components from the EF Core project into the Abstractions project, and added a dependency on the abstractions in the Ef library, with a focus on minimizing breaking changes..

With the exception of classes whose namespaces were intentionally changed, I tried to keep the same namespaces intact whenever possible, particular in the case of highly related components.

Considering that all of the Internal namespaces give warnings in the code docs for outside parties not to rely on them, I suppose it's safe to change them, and in doing so, simplify the library a bit.

IEntityQueryProvider is a helper interface that was introduced in place of EntityQueryProvider and used in the extensions for IIncludableQueryable, as the EntityQueryProvider type is no longer accessible to those extensions.  Would you propose I eliminate any "Internal" namespaces and only preserve old namespaces if they would have resulted in breaking changes?

I could move everything into the Microsoft.EntityFrameCore.Abstractions namespace unless there is a good case for why it lead to breaking changes.

divega: @rdefreitas In general the bar for moving internal types to public surface his quite high, so no, I am not suggesting that we move anything. My main point is that carrying internal types in an Abstractions packages smells. It may be a symptom that what you are creating is not really an abstractions package, that the factoring needs improvement or - like with any smell - it maybe ok for some reason (although I would be suprised). Anyway, I will let you and @anpete figure that out :smile:

src/Microsoft.EntityFrameworkCore.Abstractions/Properties/InternalsVisibleTo.cs
rdefreitas: I agree that it feels wrong...

It was necessary to move CoreStrings in Abstractions so that the embedded resources could be accessed there, but oddly the shared utility classes, particularly Check, were encountering type collision issues when they existed in both libraries.

In light of the collision and in the interest of minimizing future support, I did not wish to break out CoreStrings and Check to include a different internal version in each of the 2 libraries. allowing EFCore to reference those from Abstractions.

I can go back to that issue this weekend and see if maybe there was something overlooked that led to that type collision during builds, although nothing comes to mind off-hand. If you have any suggestions I'm open to them :)

src/Microsoft.EntityFrameworkCore.Abstractions/Query/Internal/IEntityQueryProvider.cs
anpete: @rdefreitas Change the Include stuff to test for IAsyncQueryProvider.

rdefreitas: To be honest, I think these extensions, with the exception of IIncludableQueryable would feel more natural as part of System.Interactive.Async in [Reactive-Extensions/Rx.NET](https://github.com/Reactive-Extensions/Rx.NET), but then that would leave the questions of the best way to hook up special handling for EF's EntityQueryProvider, where the Includable extensions would live, and of course the overall issues of API coupling across disparate projects.

The idea here was to allow an indispensable set of Extension methods from Entity Framework to be utilized in service libraries that do not directly reference the bulk of EFCore, without the need to introduce proxy/wrapper classes which easily can get out of hand.

src/Microsoft.EntityFrameworkCore.Abstractions/Properties/InternalsVisibleTo.cs
rdefreitas: @anpete this covers the reason behind the moving of CoreStrings.resx, although I did not encounter the build issues you're speaking of regarding EntityFrameworkCore.csproj ... were you building that in isolation from the solution and is that a scenario that needs to be supported?

anpete: No, opened EntityFramework.sln and hit build.

rdefreitas:
@anpete based on feedback, here's the changes I'll be looking at this weekend... if you have any other suggestions before I dive back please let me know.

- [ ] Move CoreStrings back to EntityFrameworkCore.
- [ ] Remove InternalsVisibleTo
- [ ] Remove reference to shared NotNull checks - add subset version for the Abstractions library only, utilizing an Abstractions resource file for the error messages that should be utilized by the Abstractions lib.
- [ ] Split AsyncQueryableExtensions and IncludableExtensions (note, since the library is very focused on providing extension methods, should we break this out into partial static classes as opposed to the regions that exist now?)
- [ ] Drink more coffee :)

anpete:
@rdefreitas Looks good. To address the internal IAsyncQueryProvider issue, we are thinking that we can have IEntityQueryProvider in Abstractions and put the async query methods on it. So, IEntityQueryProvider can be used from both the async and Include extension methods. Sound reasonable?

Re the Check library, just link the whole file, like we do for everything else.

Fine to use partials to better organize the extension methods.

rdefreitas:
@anpete, that all seems fine with the exception of the Check library, since it references the CoreStrings resources from EntityFrameworkCore and this library is independent of the EFCore base lib.

Also, do I need to install a separate version of the Visual Studio tooling?  The .sln file never worked for me in VS, only the -xproj.sln

anpete:
@rdefreitas Oh, I see the problem now. Let's just lift the Check strings to the Abstractions resource file.

rdefreitas:
@anpete I ran into personal delays on my end in being able to get back to this PR, so in the interest of still having a worthwhile enhancement for 1.1.0 depending on release date, I've streamlined a separate pull request to just provide the synchronous fallback to the existing methods and will come back to this another time.

PR #6103, opened 2016-07-18
Ensure fixup doesn't get confused when already done manually

ajcvickers:
Issue #6067

The issue was that we were not checking that the inverse had not already been fixed up before changing it. Added a check to stop this. Also found related issue that if shared PK/FK of Added entity was set manually, then it shouldn't be overwritten with same value by fixup since this will stop it being processed correctly later by DetectChanges.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/GraphUpdatesTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/GraphUpdatesTestBase.cs
@@ -6,10 +6,10 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using Microsoft.EntityFrameworkCore.ChangeTracking;
-using Microsoft.EntityFrameworkCore.Specification.Tests.TestUtilities.Xunit;
 using Microsoft.EntityFrameworkCore.Infrastructure;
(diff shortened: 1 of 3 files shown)


AndriySvyryd:
:shipit: 
Should we ship this in 1.0.1?

ajcvickers:
@rowanmiller @AndriySvyryd We decided in triage not to, but I could be convinced otherwise. I am very much on the fence.

rowanmiller:
Yeah I'm on the fence too... I think the problem is we don't really know how bad the side effects could be... but we do suspect there is a possibility for invalid data being saved if the same context is used for multiple save operations (is that consistent with your thoughts @ajcvickers).

How risky do we think the fix is?

ajcvickers:
@rowanmiller That is consistent with my thoughts. I don't think the fix is medium risk. The risk is that things might break in real world models with several levels of navigations than can cause cascading actions among keys.

AndriySvyryd:
@ajcvickers Makes sense

ajcvickers:
My vote is still marginally in favor of 1.1. @rowanmiller? @AndriySvyryd?

rowanmiller:
Yep, same here

AndriySvyryd:
1.1 sounds good, since we haven't seen any real users blocked by this

PR #6104, opened 2016-07-18
Fix code scaffolding in hotfix-branch

natemcmaster:
Code scaffolding is broken in the hotfix branch because the assembly version has changed to "1.0.1.0".

I will also apply this same fix to the dev branch.

cc @Eilon @lajones

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/DesignTimeProviderServicesTest.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/DesignTimeProviderServicesTest.cs
@@ -1,0 +1,29 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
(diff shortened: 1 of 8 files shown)


lajones:
:shipit: The fix looks good. I wonder whether long term we could arrange for the version value to get substituted into the attribute as part of the build?

natemcmaster:
Agreed. I think that will be do-able when we have finer grained build controls (ahem msbuild). But I’m not sure of a good way to do this with dotnet-build.

ErikEJ:
Does this mean that the 1.0.0.x providers will have to be re-released to support the 1.0.1 EF hotfix?

rowanmiller:
@ErikEJ this shouldn't affect other providers... it only affects ours because we bumped the version number on our provider design assemblies. Of course, if you change your design assembly version then you'd also have to update the attribute in your runtime assembly.

natemcmaster:
Correct. The tooling for .NET Framework creates binding redirects to handle assembly version changes. .NET Core does the same thing with the "deps.json" file. This change only affects SQLite and SqlServer providers, not the other EF base libraries.

Issue #6106, opened 2016-07-18
Original values must be required for principal keys


ajcvickers:
We are not currently requiring original values to be saved for principal keys, but the update pipeline uses these values in certain delete scenarios. This causes some saves to fail when using notification entities.

PR #6107, opened 2016-07-19
Store original values for principal keys

ajcvickers:
Update pipeline requires them for certain delete scenarios. Issue #6106

Also add test coverage for changes/fixup using different change notification strategies.


AndriySvyryd:
:shipit:

ajcvickers:
Merged

Issue #4749, opened 2016-03-10
Query :: error during groupjoin flattening for complex query with multiple joins subquery and LOJ


maumar:
query:

```
from qRoot in ctx.Questions
join qAnswered in
(
    from instance in ctx.ExamInstances
    join answered in ctx.ExamInstanceQuestions
        on instance.StudentId equals 1
    join qVersion in ctx.Questions on answered.QuestionId equals qVersion.Id
    select qVersion
)
    on qRoot.Id equals qAnswered.RootQuestionId ?? qAnswered.Id
    into AnsweredQuestions
from qAnswered in AnsweredQuestions.DefaultIfEmpty()
select qRoot;

```

model:

```
        public class MyContext : DbContext
        {
            public DbSet<Question> Questions { get; set; }
            public DbSet<ExamInstance> ExamInstances { get; set; }
            public DbSet<ExamInstanceQuestion> ExamInstanceQuestions { get; set; }
        }

        public class Question
        {
            public int Id { get; set; }
            public int? RootQuestionId { get; set; }
            public DateTime Version { get; set; }
            public Question3582 RootQuestion { get; set; }
        }

        public class ExamInstance
        {
            public int Id { get; set; }
            public int StudentId { get; set; }
        }

        public class ExamInstanceQuestion
        {
            public int Id { get; set; }
            public int ExamInstanceId { get; set; }
            public int QuestionId { get; set; }
            public int? AnswerId { get; set; }
            public ExamInstance3582 ExamInstance { get; set; }
            public Question3582 Question { get; set; }
        }
```

Expression that is being flatten:

```
IEnumerable<TransparentIdentifier<Question3582, IEnumerable<ValueBuffer>>> _GroupJoin(
    outer: IEnumerable<Question3582> _ShapedQuery(
        queryContext: Unhandled parameter: queryContext, 
        shaperCommandContext: SelectExpression: 
            SELECT [qRoot].[Id], [qRoot].[RootQuestionId], [qRoot].[Version], [t].[RootQuestionId], [t].[Id]
            FROM [Questions] AS [qRoot]
            LEFT JOIN (
                SELECT [qVersion0].*
                FROM [ExamInstances] AS [instance0]
                INNER JOIN [ExamInstanceQuestions] AS [answered0] ON [instance0].[StudentId] = 1
                INNER JOIN [Questions] AS [qVersion0] ON [answered0].[QuestionId] = [qVersion0].[Id]
[...]

PR #6111, opened 2016-07-19
Fix to #4749 - Query :: error during groupjoin flattening for complex query with multiple joins subquery and LOJ

maumar:
Problem was that we tried to lift queries that we shouldn't have - i.e. GroupJoin whose ShapedQuery elements were wrapped in Select. Fix is to check the shape of the expression before performing the lifting - we do similar thing for SelectMany already.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -509,47 +509,45 @@ namespace Microsoft.EntityFrameworkCore.Query
         {
             var selectManyExpression = Expression as MethodCallExpression;
             if (selectManyExpression == null
-                || !selectManyExpression.Method.MethodIsClosedFormOf(LinqOperatorProvider.SelectMany))
+                || !selectManyExpression.Method.MethodIsClosedFormOf(LinqOperatorProvider.SelectMany)
+                || !IsShapedQueryExpression(selectManyExpression.Arguments[0] as MethodCallExpression, innerShapedQuery: false)
+                || !IsShapedQueryExpression((selectManyExpression.Arguments[1] as LambdaExpression)?.Body as MethodCallExpression, innerShapedQuery: true))
             {
                 return false;
             }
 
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

anpete:
@maumar Do we need a new model here? Or, can we re-use one of our existing fixtures? I would prefer the test to be cross-provider, and do SQL verification against SQL Server.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -5697,6 +5698,314 @@ public override void String_concat_with_navigation2()
...
         }
 
+
+
+
+
+
+
+
+        public class ComplexGroupJoin
```
maumar: ignore this part, its going away - residual code from different investigation

maumar:
new version up @anpete

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -737,6 +744,28 @@ var joinExpression
             }
         }
 
+        private bool CanFlattenGroupJoin()
+        {
+            var groupJoinExpression = Expression as MethodCallExpression;
```
anpete: You could collapse these conditionals into a single expression.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -514,42 +514,48 @@ private bool CanFlattenSelectMany()
...
             }
 
-            var innerShapedQuery = (selectManyExpression.Arguments[1] as LambdaExpression)?.Body as MethodCallExpression;
-            if (innerShapedQuery == null)
+            return true;
+        }
+
+        private bool IsShapedQueryExpression(MethodCallExpression shapedQueryExpression, bool innerShapedQuery)
+        {
+            if (shapedQueryExpression == null)
```
anpete: You could probably collapse these conditionals into a single expression.

anpete:
:shipit:

PR #6112, opened 2016-07-19
Refactor PMC commands to invoke ef.exe/ef.dll

natemcmaster:
NB: will merge to feature branch. Based of changes in #6079 

Part of #5334

Changes:

 - Package Manager Console commands start a new process using "ef.exe" or
   "dotnet.exe ef.dll"
 - Add --no-color and --no-appdomain to ef.exe
 - ef.exe can fallback to reflection based invocation
 - Add a "prefixed" reporter to ef.exe which helps PMC commands redirect
   stdout to PowerShell's Warning, Output, Warning, and Error streams
 - Add more --json option to 'dbcontext scaffold' and 'migrations remove'

TODO:

 - more automated testing
 - xmldoc

cc @bricelam


src/Tools.Console/Commands/DbContextScaffoldCommand.cs
```diff
@@ -99,9 +103,47 @@ public static void Configure([NotNull] CommandLineApplication command, [NotNull]
...
-            => executor.ReverseEngineer(_provider, _connection, _outputDir, _context, _schemas, _tables, _dataAnnotations, _force);
+        {
+            var filesCreated = executor.ReverseEngineer(_provider, _connection, _outputDir, _context, _schemas, _tables, _dataAnnotations, _force);
+            if (_json)
+            {
+                ReportJsonResults(filesCreated);
+            }
+        }
+
+
```
ilmax: nit: Empty line

src/Tools.Console/PrefixConsoleReporter.cs
```diff
@@ -0,0 +1,43 @@
...
+            => base.Output(PrefixLines(ErrorPrefix, message));
+
+        public override void Output(string message)
+            => base.Output(PrefixLines(OutputPrefix, message));
+
+        private static string PrefixLines(string prefix, string message)
+        {
+            var lines = message.Split(new[] { Environment.NewLine, "\n" }, StringSplitOptions.None).Select(m => prefix + m);
+
+            if (message.EndsWith(Environment.NewLine, StringComparison.OrdinalIgnoreCase))
```
ilmax: Cannot this be simplified with .Split(new[] { Environment.NewLine, "\n" }, StringSplitOptions.RemoveEmptyEntries)? Do we care about empty lines in the middle of the message?

natemcmaster: > Do we care about empty lines in the middle of the message?

Yes, this is intentional.

src/Tools.Console/OperationExecutorFactory.cs
```diff
@@ -13,19 +13,22 @@ public class OperationExecutorFactory
...
-                    projectDir: options.ProjectDirectory ?? Directory.GetCurrentDirectory(),
-                    dataDirectory: options.DataDirectory ?? Directory.GetCurrentDirectory(),
-                    contentRootPath: options.ContentRootPath,
-                    rootNamespace: options.RootNamespace,
-                    environment: options.EnvironmentName);
-#else
             try
             {
+#if NET451
+                if (!options.NoAppDomain)
```
bricelam: Why enable this? The other code path won't work unless you copy `ef.exe` to your `bin` directory, and even then the `App.config` won't get used for settings, redirects, etc.

src/Tools.Console/Commands/DbContextScaffoldCommand.cs
bricelam: &nbsp;

src/Tools.Console/Program.cs
```diff
@@ -95,14 +95,30 @@ public static int Main(string[] args)
...
             for (var i = 0; i < args.Length; i++)
             {
                 if (args[i] == "--debug")
                 {
+                    debug = true;
                     args = args.Take(i).Concat(args.Skip(i + 1)).ToArray();
+#if NET451
+                    Console.WriteLine("Waiting for debugger to attach");
+                    Console.WriteLine($"Process ID: {Process.GetCurrentProcess().Id}");
+                    while (!Debugger.IsAttached);
```
bricelam: Just `Debugger.Launch` instead?

src/Tools.VisualStudio/InvokeProcessCommand.cs
```diff
@@ -0,0 +1,250 @@
...
+    ///     <para>
+    ///         When -RedirectByPrefix is specified, each line of stdout/stderr can be redirected to 'Debug', 'Verbose', etc. based on the prefix.
+    ///         <cref see="RedirectByPrefix" />.
+    ///         When -JsonOutput is specified, the cmdlet will only output JSON objects found in stdout/stderr that have been wrapped by '//BEGIN'
+    ///         and '//END' <cref see="JsonOutput" />
+    ///     </para>
+    /// </summary>
+    [Cmdlet("Invoke", "Process")]
+    [OutputType(typeof(string))]
+    public class InvokeProcessCommand : PSCmdlet
```
bricelam: Eww... Is there any way we can get away with not owning a C#-implemented cmdlet?

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -649,126 +520,198 @@ function InvokeDotNetEf($dteProject, $dteStartupProject, [switch] $json, [switch
...
+        throw "Invalid JSON file in '$targetProjectJson'"
+    }
+
+    $frameworks = $project.frameworks | Get-Member -MemberType NoteProperty | % Name
+    if ($frameworks -is 'String') {
+        # when there is one framework listed
+        return $frameworks
+    }
+    $choices = [System.Management.Automation.Host.ChoiceDescription[]]( $frameworks | % {$i=0} { new-object System.Management.Automation.Host.ChoiceDescription "Option &$i`: $_"; $i++} )
+    $choice = $Host.Ui.PromptForChoice('Multiple target frameworks available', 'Which framework should the command use?', $choices, 0)
```
bricelam: Isn't there an active one we can just use?

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -649,126 +520,198 @@ function InvokeDotNetEf($dteProject, $dteStartupProject, [switch] $json, [switch
...
+    # only returns part of the actual output path
+    $outputPath = Join-Path $startupProjectDir (GetProperty $startupProject.ConfigurationManager.ActiveConfiguration.Properties OutputPath)
+
+    $config = $startupProject.ConfigurationManager.ActiveConfiguration.ConfigurationName
+
+    # TODO get the actual output file from VS when/if CPS API's become available
+    $assemblyName = GetProperty $startupProject.Properties AssemblyName
+
+    $arguments = @()
+    if ($tfm -like 'net45*' -or $tfm -like 'net46*') {
```
bricelam: I don't like having these numbers hard-coded here and below.

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -649,126 +520,198 @@ function InvokeDotNetEf($dteProject, $dteStartupProject, [switch] $json, [switch
...
-    
-    if ($skipBuild) {
-        $arguments += ,"--no-build"
+    $arch = GetProperty $startupProject.ConfigurationManager.ActiveConfiguration.Properties PlatformTarget
+    if ($arch -eq 'x86') {
+        $exe = Join-Path $PSScriptRoot 'net451/ef.x86.exe'
+    } elseif ($arch -eq 'AnyCPU' -or $arch -eq 'x64') {
+        $exe = Join-Path $PSScriptRoot 'net451/ef.exe'
+    } else {
+        throw "Cannot invoke command. The current configuration targets and unsupported architecture '$arch'"
```
bricelam: Should we just default to `ef.exe`?

bricelam: (and see what happens)

src/Tools.VisualStudio/InvokeProcessCommand.cs
natemcmaster: Yeah, I agree it's not ideal. But I haven't found anything in PS that gives us the functionality we want. Here was my thinking on why we need this:

- Invoke-Expression has all sorts of yucky problems with argument escaping. We've already solved most, but not all of them (such as dangling quotes). With a C# cmdlet, we can just pass in `string[]` arg instead and use a more-robust argument escaper.
- Argument escaping issues could have been mitigated by using the call operator instead `& ef.exe`, but PMC hides all output from out-of-proc calls initialized with the call operator.
- Therefore, the only way to show output would be redirect to file or variable and print its contents when the ef.exe operation completes.
- AFAIK not possible to process both stderr and stdout in parallel from code in psm1

src/Tools.Console/OperationExecutorFactory.cs
natemcmaster: UWP design-time. UWP ends up with the NETStandard .Design assembly which means the internal OperationExecutor isn't an MBRO.

src/Tools.Console/CommandLineOptions.cs
```diff
@@ -39,13 +41,27 @@ public static CommandLineOptions Parse(params string[] args)
             app.HelpOption();
             app.VersionOption(Program.GetVersion);
             var verbose = app.Option("--verbose", "Show verbose output", inherited: true);
+            var noColor = app.Option("--no-color", "Do not use color in console output");
```
bricelam: :trollface: Can we have `--im-a-robot` as sugar for `--no-color` and `--json`?

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
natemcmaster: Not AFAIK. I searched the DTE namespace but couldn't find anything.

src/Tools.Console/OperationExecutorFactory.cs
bricelam: Makes sense.

src/Tools.Console/CommandLineOptions.cs
natemcmaster: yes, absolutely yes. 🤖

src/Tools.VisualStudio/InvokeProcessCommand.cs
bricelam: Sounds justified. We should test what will happen in the future during upgrade scenarios.

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
bricelam: It might be web-tools specific (which means it'll probably change anyway)

bricelam:
:shipit:

src/Tools.VisualStudio/InvokeProcessCommand.cs
natemcmaster: Here's what I've found so far:

- binary PS modules cannot be unloaded. 
  Implications:
  - `Uninstall-Package` tools package requires VS restart b/c the .dll is still loaded in the PMC domain
  - `Remove-Module` silently fails to remove binary modules. => Upgrading 1.1 to 1.2 (without a VS restart) will still have the old version of the binary module. We could mitigate by adding full version to the name of the name of the binary module.

bricelam: Is it likely to ever actually change?

natemcmaster: Given our recent tooling history, then absolutely yes this will change. But of course I'd like to think this won't need much revision. 🔮

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
natemcmaster: Any suggestion for alternatives? 

Also, most of this will change as project.json dies. In theory we should be able to pull this from DTE/CPS instead

bricelam: Can we get just the framework identifier (without the version) e.g. .NETFramework?

natemcmaster: Not from DTE. This isn't exposed as a property on the project or in the active configuration. We could start bundling Nuget.Frameworks.dll to parse the tfm string, but I'm not sure we want to do that either.

natemcmaster: For reference, here is what DTE gives us about an xproj `Project`.
```
OutputType = Library
OutputName = 
FullPath = C:\Users\namc\dev\ef\src\Tools.DotNet
ActiveDebugFramework = 
IsClasslibraryProject = False
RootNamespace = Microsoft.EntityFrameworkCore.Tools.DotNet
AvailablePlatforms = System.Collections.ObjectModel.ReadOnlyCollection`1[System.String]
FileName = Tools.DotNet.xproj
TargetFrameworkMoniker = .NETFramework,Version=v4.5.1
IsWebProject = False
AssemblyName = Tools.DotNet
ProduceOutputsOnBuild = False
ApplicationIcon = 
TargetFramework = 262405
ActiveDebugRuntime = 
TargetDNXVersion = 
DebugCLRTypeOverride = 
UserSecretsId = 
Name = Tools.DotNet
```
(FYI Tools.DotNet only targets "netcoreapp1.0", but "TargetFrameworkMoniker" says .NETFramework for some reason, and "TargetFramework = 262405" doesn't appear to actually correlate to the TFMs in projectjson)

From active configuration
```
OutputPath = .\bin\
```

bricelam: There's enough TODOs in here to just move forward with what you have for now. :wink:

PR #6116, opened 2016-07-20
Rename DatabaseProvider`.cs to DatabaseProvider.cs

fadihania:


AndriySvyryd:
@fadihania Do you experience any issues with the current name?

anpete:
There is no non-generic version, which is why we normally would have "`" in the name. Not sure if we are fully consistent on this though.

ajcvickers:
@anpete @AndriySvyryd We are not consistent. Often, as is the case here, there used to be a non-generic version. Personally I don't see much value in doing the rename, but I don't feel strongly one way or the other.

anpete:
My vote would be to take it. I'm happy to merge it if others are OK with it.

AndriySvyryd:
I'm ok with it

ajcvickers:
I'm okay either way.

fadihania:
@AndriySvyryd No, but I don't remember that I've seen this way of naming files used anywhere else in .NET repos. So, I thought it was a mistake

Issue #6039, opened 2016-07-11
NullReferenceException if FromSql returns no data


corradolab:
```
public class SomeTotals
{
  [Key] // required by EF but unused
  public Decimal Sales { get; set; }
}

string sql = "select sum(Sales) as Sales from SomeTable where SomeField > @p0";
return context.SomeTable.FromSql(sql, aParam).FirstOrDefault();

```
Execute a Query using FromSql.<br/>
Let the query return a custom type, not directly backed by a physical table. <br/>
If the query returns no results, I get NullReferenceException. <br/>
No error if the query returns any result.

```
Exception message:
info: Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommandBuilderFactory[1]
      Executed DbCommand (27ms) [Parameters=[@p0='?', @p1='?'], CommandType='Text', CommandTimeout='30']
      SELECT    sum(Incassato) as Incassato FROM Rilevazioni WHERE  [Data]>=@p0  AND [Data]<@p1
fail: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware[0]
      An unhandled exception has occurred: Object reference not set to an instance of an object.
System.NullReferenceException: Object reference not set to an instance of an object.
   at lambda_method(Closure , ValueBuffer )
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.UnbufferedEntityShaper`1.Shape(QueryContext queryContext, ValueBuffer valueBuffer)
   at Microsoft.EntityFrameworkCore.Query.AsyncQueryMethodProvider.<>c__DisplayClass3_0`1.<_ShapedQuery>b__0(ValueBuffer vb)
   at Microsoft.EntityFrameworkCore.Query.Internal.AsyncLinqOperatorProvider.SelectAsyncEnumerable`2.SelectAsyncEnumerator.get_Current()
   at Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.<ToListAsync>d__129`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
[...]

Issue #4289, opened 2016-01-12
Query: Better diagnostics/errors/exceptions on failed queries


rowanmiller:
Because query is so perf sensitive we don't do much defensive checking of results etc. which can result in exceptions that are difficult to debug. We don't want to add these defensive checks, since that would have a significant impact on performance, but once a query fails we could run some additional logic to try and provide some guidance on why the error may have happened. In EF6 we would re-run parts of the query pipeline in a "slow but defensive" mode to try and get a better error. We don't necessarily want to take the same approach in EF7... we may just want to run some analysis on the results/model to suggest what issues could be the result of errors.

For example:
- If the database returns a null value for a column that is configured as non-nullable in the model, EF doesn't defensively check this and you get a NullReferenceException with no helpful details (see https://github.com/aspnet/EntityFramework/issues/4120). 
- If the data type of a value in the result set can not be cast to the data type configured in the model, then an InvalidCastException with no helpful details (see https://github.com/aspnet/EntityFramework/issues/3923).

divega:
Clearing milestone to have a chance to discuss this again in triage. During app week I had to work with an existing database and hit this kind of issue multiple times. The way EF Core crashed when there was a type mismatch or an unexpected null value was not helpful at all and diagnosing the cause of the error was very time consuming.

rowanmiller:
talk with @anpete  and @divega about some ideas of how we do this

PR #6121, opened 2016-07-21
Query: Better error messages when encountering invalid data.

anpete:
Surround generated ValueBuffer reads with a Try/Catch so we can throw a better exception. Explicit detection of common cases: Null ref and invalid cast.

Fix: #6039, #4289

No perf degradation observed when running tests locally. I will monitor perf runs after this goes in.


anpete:
@ajcvickers

anpete:
@divega

divega:
LGTM, but I prefer that @ajcvickers still reviews it :smile:

ajcvickers:
Nice. :shipit: from me.
@rowanmiller to review the exception messages.

rowanmiller:
:shipit: from exception message point of view

Issue #625, opened 2014-09-04
Explicit loading support


divega:
Explicitly triggering the loading of associated data through a navigation can be handy especially when loading across aggregate bondaries or when working with stores that are not natively [aggregate oriented](http://martinfowler.com/bliki/AggregateOrientedDatabase.html). In previous versions of EF we had such methods as Load() and LoadAsync() on reference and collection property entries. We also had handy Query() methods that returned the query we would use to load the data so that filters could be composed over them before executing them. We should have the equivalent functionality in EF7.

mjrousos:
+1; I've recently been porting some old EF6 code which used lazy loading and while I'm ok with not using lazy loading in the new world, some of the models are too large to entirely eagerly load. Instead, I have to explicitly load related objects as they're needed. Having first-class support for this with APIs like Load() and LoadAsync() would make porting easier.

nocommonsense:
+1 for this

arielcloud:
The docs [here](https://ef.readthedocs.io/en/latest/querying/related-data.html#explicit-loading) reference this issue.
if u close this it means the issue solved? if so - can u fix the docs? thanks!

rowanmiller:
@arielcloud it's fixed in our working code base, but not in the current stable release. We will update the docs when this is released in 1.1.

RickStrahl:
Looking at this it occurs to me that it would be nice to have a model attribute that would allow you to specify that related entities or collections should always be included. This would greatly mitigate the need for lazy or even explicit loading in many situations. 

It'd be nice if you could just say:

```csharp
public class album
{
    ...
    [Include]
    public Artist Artist {get; set;}

    [Include]
    public Track[] Tracks {get; set; }
}
````

divega:
@RickStrahl I think that is covered by #1985, but please let us know if you have additional feedback.

RickStrahl:
Thank you. Yup that looks like it would work. 

These options will make life a lot easier for generic code in base repositories.

PR #6131, opened 2016-07-21
Implement explicit loading APIs for navigation properties

ajcvickers:
Modeled closely on the EF6 APIs

Issue #625

@rowanmiller Can you look at the API and doc comments?

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/LoadTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/LoadTestBase.cs
@@ -1,0 +1,2878 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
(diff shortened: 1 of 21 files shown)


src/Microsoft.EntityFrameworkCore.Specification.Tests/LoadTestBase.cs
```diff
@@ -0,0 +1,2878 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations.Schema;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.EntityFrameworkCore.Specification.Tests
+{
+    public abstract class LoadTestBase<TFixture> : IClassFixture<TFixture>, IDisposable
+        where TFixture : LoadTestBase<TFixture>.LoadFixtureBase, new()
```
anpete: Looks like a nice test model. Would be great if the model/fixture was split out to be re-usable.

anpete:
:shipit:

PR #6133, opened 2016-07-21
[1.0.1] Fix GetReferencingForeignKeys for derived types

### LINKED ISSUE #6055: Wrong save order for one to many in combination with multi level inheritance
### Steps to reproduce

Create 4 classes, Relation (abstract), Person, Employee and Address. Employee inherits from Person, and Person implements Relation.

`public abstract class Relation {}`
`public class Person : Relation {}`
`public class Employee : Person {}`
`public class Address {}`

Relation entity and Address entity have a primary key
`[Key, DatabaseGenerated(DatabaseGeneratedOption.Identity)]`
`public int Id { get; set; }`

Relation also has a collection property for Addresses
`public virtual ICollection<Address> Addresses { get; set; }`

plus some other properties.

The DbContext class has a DbSet for each entity except Address (does not matter for what i'm seeing).
`public DbSet<Employee> Employees { get; set; }`
`public DbSet<Person> Persons { get; set; }`
`public DbSet<Relation> Relations { get; set; }`

EF creates 2 tables in the database, as expected (TPH), using a discriminator for the Relations table.

Next, create a new instance of Employee, and add a new Address to the Adresses collection. Add the Employee to the Employees DbSet, and call SaveChanges on the context.

Do the same, but with Person and Persons DbSet.

### The issue
When saving the Employee entity to the Employees DbSet, and the Employee Entity has 1 or more Addresses added it should first persist the Employee entity, and then the Address entities.

Currently it saves the Address entities first, resulting in Address entities without a relation to the Employee (Relation entity). Since there is no way to find these records they are 'lost' in the database.

This only happens when the entity is inheriting 1 class which implements another (as far as I can tell).

Note that the entry in the relation table is always correct. It has the discriminator that is expected (Person, Employee).

My expectation would be that it should fail instead of what is happening right now. (when a foreign key is added in the database, it does fail, as it should). 

I have a more extensive test project that I can attach if needed.

### Further technical details

EF Core version: 1.0.0
Database provider: Npgsql.EntityFrameworkCore.PostgreSQL 1.0.0
Operating system: Windows 10
Visual Studio version: VS 2015 Update 3

AndriySvyryd:
Use ReferenceEqualityComparer in StateManager

Fixes #6055 

I think this fix should go into 1.0.1

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/GraphUpdatesTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/GraphUpdatesTestBase.cs
@@ -31,7 +31,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         {
             using (var context = CreateContext())
             {
-                var root = context.Roots.Single();
+                var root = context.Roots.Single(IsTheRoot);
 
                 root.OptionalSingle = new OptionalSingle1();
 
@@ -44,7 +44,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         {
             using (var context = CreateContext())
             {
-                var root = context.Roots.Single();
+                var root = context.Roots.Single(IsTheRoot);
 
                 root.RequiredSingle = new RequiredSingle1();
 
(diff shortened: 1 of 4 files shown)


ajcvickers:
:shipit: and +1 for 1.0.1

PR #6135, opened 2016-07-21
Fix to #5672 - Query :: Include with multiple navigations (including optional navigation) fails

maumar:
Problem was that our include logic for groupjoin was expecting outer and inner elements (on which the include was being performed) to be entities. However sometimes those elements would be TransparentIdentifiers (e.g. when the element comes from a result of SelectMany).

Fix is to also store (when necessary) the accessor from the outer/inner element to the entity that we want to include, extract the actual entity in runtime and apply include operations on that entity instead of the actual outer/inner element.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
@@ -385,8 +385,10 @@ namespace Microsoft.EntityFrameworkCore.Query
                 private IAsyncEnumerator<ValueBuffer> _sourceEnumerator;
                 private bool _hasNext;
                 private TOuter _nextOuter;
-                private AsyncGroupJoinInclude.AsyncGroupJoinIncludeContext outerGroupJoinIncludeContext;
-                private AsyncGroupJoinInclude.AsyncGroupJoinIncludeContext innerGroupJoinIncludeContext;
+                private AsyncGroupJoinInclude.AsyncGroupJoinIncludeContext _outerGroupJoinIncludeContext;
+                private AsyncGroupJoinInclude.AsyncGroupJoinIncludeContext _innerGroupJoinIncludeContext;
+                private Func<TOuter, object> _outerEntityAccessor;
+                private Func<TInner, object> _innerEntityAccessor;
 
                 public GroupJoinAsyncEnumerator(
                     GroupJoinAsyncEnumerable<TOuter, TInner, TKey, TResult> groupJoinAsyncEnumerable)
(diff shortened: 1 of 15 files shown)


maumar:
@anpete @divega

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -418,7 +418,15 @@ var outer
 
                         if (_groupJoinAsyncEnumerable._outerGroupJoinInclude != null)
                         {
-                            await _groupJoinAsyncEnumerable._outerGroupJoinInclude.IncludeAsync(outer, cancellationToken);
+                            var outerEntityAccessor = _groupJoinAsyncEnumerable._outerGroupJoinInclude.EntityAccessor as Func<TOuter, object>;
```
anpete: Can we only do this cast once?

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -385,8 +385,10 @@ private sealed class GroupJoinAsyncEnumerator : IAsyncEnumerator<TResult>
                 private IAsyncEnumerator<ValueBuffer> _sourceEnumerator;
                 private bool _hasNext;
                 private TOuter _nextOuter;
-                private AsyncGroupJoinInclude.AsyncGroupJoinIncludeContext outerGroupJoinIncludeContext;
-                private AsyncGroupJoinInclude.AsyncGroupJoinIncludeContext innerGroupJoinIncludeContext;
+                private AsyncGroupJoinInclude.AsyncGroupJoinIncludeContext _outerGroupJoinIncludeContext;
```
anpete: Oops! :smile:

src/Microsoft.EntityFrameworkCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -418,9 +422,16 @@ var outer
 
                         _nextOuter = default(TOuter);
 
-                        if (outerGroupJoinIncludeContext != null)
+                        if (_outerGroupJoinIncludeContext != null)
                         {
-                            await outerGroupJoinIncludeContext.IncludeAsync(outer, cancellationToken);
+                            if (_outerEntityAccessor != null)
```
anpete: Use ternary

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -161,9 +171,18 @@ var groupJoinInclude
 
                 if (groupJoinInclude != null)
                 {
-                    var newArguments = methodCallExpression.Arguments.ToList();
+                    var groupJoinIncludeExpression = (Expression)Expression.Constant(groupJoinInclude);
```
anpete: Why this cast?

anpete:
:shipit: after R# cleanup.

PR #6136, opened 2016-07-21
Fix #5444 - Uwp ToListAsync() throws Exception with Message "The sema…

### LINKED ISSUE #5444: Uwp ToListAsync() throws Exception with Message "The semaphore has been disposed."
Hi,
I am using EF RC2 with an UWP App, after the update from EF RC1 the following Code lines:
`var iq= ctx.Projects.Include(p => p.Address).ThenInclude(a => a.GeoLocation);`
`return await iq.ToListAsync();`

throw an exception with the message="The semaphore has been disposed".

The same query works fine with synchronious ToList()

=> SingleOrDefaultAsync is also not working as expected.

anpete:
…phore has been disposed.

Improves exception that is generated when a context is disposed before an async query has completed.


ajcvickers:
:shipit:

Issue #6109, opened 2016-07-19
ChangeTracker shouldn't throw on conceptual nulls if one of the fk properties is nullable


AndriySvyryd:
```C#

public class Product
{
    public int Id { get; set; }
    public int PartitionId { get; set; }

    public int? CategoryId { get; set; }
    public Category Category { get; set; }
}

public class Category
{
    public int Id { get; set; }
    public int PartitionId { get; set; }

    public ICollection<Product> Products { get; set; }
}

public class StoreContext : DbContext
{
    public DbSet<Product> Products { get; set; }
    public DbSet<Category> Categories { get; set; }

    public StoreContext(DbContextOptions options)
        : base(options)
    {
    }

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<Product>()
            .HasKey(p => new { p.Id, p.PartitionId });
        modelBuilder.Entity<Category>()
            .HasKey(c => new { c.Id, c.PartitionId });

        modelBuilder.Entity<Product>().HasOne(p => p.Category)
            .WithMany(c => c.Products)
            .HasForeignKey(p => new { p.CategoryId, p.PartitionId });
    }
}

context.Products.Add(new Product { Id = 1, PartitionId = 1, Category = new Category { Id = 2, PartitionId = 1 } });
context.SaveChanges();

context.Products.Include(p => p.Category).Single().Category = null;
context.SaveChanges();
```
Throws:
`The association between entity types 'Product' and 'Category' has been severed but the foreign key for this relationship cannot be set to null. If the dependent entity should be deleted, then setup the relationship to use cascade deletes.`

PR #6142, opened 2016-07-22
Fix handling of conceptual nulls for composite keys in change tracker

ajcvickers:
Issue #6109

Fix is to clear any conceptual null if any part of the composite key is nullable, and to instead make sure that that part is null.

Also added more coverage for update scenarios with composite keys and improved perf of GraphUpdatesTest.


anpete:
:shipit:

PR #6146, opened 2016-07-23
Query: Apply search condition optimizations on join conditions

### LINKED ISSUE #4748: Query :: invalid sql generated for query with joins where key selector is a null comparison
query:

```
from instance in ctx.ExamInstances
join answered in ctx.ExamInstanceQuestions
on true equals answered.AnswerId != null
```

generated query is:

```
SELECT [answered].[ExamInstanceQuestionId], [answered].[AnswerId], [answered].[ExamInstanceId], [answered].[QuestionId]
FROM [ExamInstances] AS [instance]
INNER JOIN [ExamInstanceQuestions] AS [answered] ON 1 = [answered].[AnswerId] IS NOT NULL
```

we should be converting the null check into CASE statement, like we do for projections

smitpatel:
Resolves #4748 
Apply search condition optimizations on join conditions

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -274,19 +274,45 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
             return selectExpression;
         }
 
-        private Expression ApplyOptimizations(Expression expression, bool searchCondition)
+        private Expression ApplyOptimizations(Expression expression, bool searchCondition, bool joinCondition = false)
         {
             var newExpression
                 = new NullComparisonTransformingVisitor(_parametersValues)
                     .Visit(expression);
 
+            var binaryExpression = newExpression as BinaryExpression;
             var relationalNullsOptimizedExpandingVisitor = new RelationalNullsOptimizedExpandingVisitor();
-            var optimizedExpression = relationalNullsOptimizedExpandingVisitor.Visit(newExpression);
+            var relationalNullsExpandingVisitor = new RelationalNullsExpandingVisitor();
 
-            newExpression
-                = relationalNullsOptimizedExpandingVisitor.IsOptimalExpansion
-                    ? optimizedExpression
-                    : new RelationalNullsExpandingVisitor().Visit(newExpression);
+            if (joinCondition
+                && binaryExpression != null)
+            {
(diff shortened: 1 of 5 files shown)


smitpatel:
@anpete - Need to understand why extra order by comes for left join queries.

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -274,19 +274,45 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
...
+            var binaryExpression = newExpression as BinaryExpression;
 
-            newExpression
-                = relationalNullsOptimizedExpandingVisitor.IsOptimalExpansion
-                    ? optimizedExpression
-                    : new RelationalNullsExpandingVisitor().Visit(newExpression);
+            if (joinCondition
+                && binaryExpression != null)
+            {
+                var relationalNullsOptimizedExpandingVisitor = new RelationalNullsOptimizedExpandingVisitor();
```
maumar: those two are common for both cases - pull them out of the if block

maumar: pull this out of the if block

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -274,19 +274,45 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
...
+                var optimizedLeftExpression = relationalNullsOptimizedExpandingVisitor.Visit(binaryExpression.Left);
+
+                optimizedLeftExpression
+                    = relationalNullsOptimizedExpandingVisitor.IsOptimalExpansion
+                        ? optimizedLeftExpression
+                        : new RelationalNullsExpandingVisitor().Visit(binaryExpression.Left);
```
maumar: RelationalNullsExpandingVisitor doesn't have state so you can make one of them and reuse it

maumar:
:shipit:  otherwise

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
smitpatel: Does it have state? `IsOptimalExpansion`?

PR #6161, opened 2016-07-25
Take the principal type into account when looking for existing relationship.

### LINKED ISSUE #6115: Model Building: StackOverflowException when two properties share the same foreign key
```
public class FileRecord : RelativeEntity
    {
        public Guid DmsId { get; set; }
        public int EntityId { get; set; }
        public string MimeType { get; set; }
        public FileRecordEntity Entity { get; set; }

        [ForeignKey("EntityId")]
        public virtual Decision Decision { get; set; }
        [ForeignKey("EntityId")]
        public virtual Invitation Invitation { get; set; }
    }
```

Can i have two properties with the same foreign key?

EF Core version: 1.0.0
Operating system: Win10 64
Visual Studio version: VS2015

AndriySvyryd:
Fixes #6115


ajcvickers:
:shipit:

Issue #1201, opened 2014-12-02
Implement relationship access methods on EntityEntry


ajcvickers:
This includes:
- Reference
- Collection
- Member (including integration with existing Property methods)

These methods should provide access to:
- CurrentValue
- Back pointer to the EntityEntry
- IsLoaded
- Load
- Query

PR #6162, opened 2016-07-25
Implement IsLoaded flag

ajcvickers:
Set when an entity is loaded with Include or one Load, regardless of whether or not there are any related entities or not. Always true for reference navigation properties on tracked entities when the reference is non-null.

Issue #625
Issue #1201


ajcvickers:
@rowanmiller Can you look at the API and doc comments?

AndriySvyryd:
:shipit:

PR #6167, opened 2016-07-26
Fix #6122 - LoadAsync does not work

### LINKED ISSUE #6122: Query: LoadAsync does not work
No entities end up being tracked.
Repro:
```C#
{

        => optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=LoadAsyncTest;Trusted_Connection=True;");
}

public class Blog
{
    public int Id { get; set; }
}

public class Program
{
    public static void Main()
    {
        Task.Run(async () =>
        {
            await MainAsync();
        }).Wait();
    }

    public static async Task MainAsync()
    {
        {

            context.AddRange(new Blog(), new Blog());
            await context.SaveChangesAsync();
        }

        {
            await context.Blogs.LoadAsync();

            Console.WriteLine(context.ChangeTracker.Entries().Count());
        }
    }
}
```
Output is 0; expected output is 2. 
When run with Load instead of LoadAsync, output is 2 as expected.

Note that there are currently zero tests for LoadAsync.

anpete:
Problem: Our async Select operator was too lazy. The projector delegate was only run when the Current property was accessed, but the code in LoadAsync spins the enumerator without accessing the Current property.

Solution: Change our async Select operator to invoke the projector delegate on MoveNext. This is preferred over accessing Current during LoadAsync because it matches L2O behavior and will prevent future instances of this bug.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/AsyncQueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/AsyncQueryTestBase.cs
@@ -59,6 +59,17 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [ConditionalFact]
+        public virtual async Task LoadAsync_should_track_results()
+        {
+            using (var context = CreateContext())
+            {
+                await context.Customers.LoadAsync();
+                 
+                Assert.Equal(91, context.ChangeTracker.Entries().Count());
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


anpete:
@ajcvickers

ajcvickers:
Can you change the LoadAsync method to use this instead of ToListAsync in EntityFinder when you merge this to dev?

ajcvickers:
:shipit:

Issue #6171, opened 2016-07-26
Fixup can be inconsistent for 1:1 relationships changed in store


ajcvickers:
If an FK value for a 1:1 relationship is moved from one row (entity) in the database to another row (entity) in the database while the first is loaded, then if the second is later loaded fixup can be inconsistent because both entities will have navigation properties pointing to the single principal.

PR #6172, opened 2016-07-26
Make fixup consistent for 1:1 relationships changed in store

ajcvickers:
Issue #6171

The fix is to to make sure that any existing dependent that points to the same principal has its FK and nav props nulled out.


AndriySvyryd:
:shipit:

PR #6173, opened 2016-07-27
Fix to #474 - Query: Improve translation of String's StartsWith, EndsWith and Contains

maumar:
**PROVIDERS BEWARE:**

Linq translation for methods Contains, EndsWith and StartsWith that we have in the Relational package uses LIKE operator, which may return incorrect results if the value parameter (what we are searching for) contains wildcard characters, e.g. '%' or '_'.

This issue addresses SqlServer and Sqlite providers, but all other providers will still use the old translation. 

**Each provider that can be affected by this should implement their own MethodCallTranslators for Contains, EndsWith and StartsWith.**

We were returning wrong results for StartsWith, EndsWith and Contains with patters containing wildcard characters (%, _)

Fix is to improve the translation as follows:

Contains:
(CHARINDEX(pattern, string) > 0) OR pattern = ''

StartsWith:
(string LIKE pattern + "%" AND CHARINDEX(pattern, string) = 1) OR pattern = ''

EndsWith:
(RIGHT(string, LEN(pattern) = pattern) OR pattern = ''

We can sometimes remove the final term (pattern = ''), i.e. when the pattern is a constant. If it's value is '', we just return true, otherwise the final term is redundant and will not be generated.
We can't really do same trick for parameters without changing the way caching works (currently it only looks at the nullability of the parameter, but in this case the value itself is important)

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
@@ -198,6 +198,9 @@
   <data name="PossibleUnintendedUseOfEquals" xml:space="preserve">
     <value>Possible unintended use of method Equals(object) for arguments of different types: '{left}', '{right}'. This comparison will always return 'false'.</value>
   </data>
+  <data name="PossibleIncorrectResultsUsingLikeOperator" xml:space="preserve">
+    <value>Linq translation for method '{function}' used by this database provider can return incorrect results when the value argument contains wildcard characters (e.g. '%' or '_').</value>
+  </data>
   <data name="StoredProcedureIncludeNotSupported" xml:space="preserve">
     <value>The Include operation is not supported when calling a stored procedure.</value>
   </data>
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionTranslators/RelationalCompositeMethodCallTranslator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionTranslators/RelationalCompositeMethodCallTranslator.cs
@@ -27,11 +27,11 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators
             _methodCallTranslators
                 = new List<IMethodCallTranslator>
                 {
-                    new ContainsTranslator(),
-                    new EndsWithTranslator(),
+                    new ContainsTranslator(logger),
+                    new EndsWithTranslator(logger),
                     new EqualsTranslator(logger),
                     new IsNullOrEmptyTranslator(),
-                    new StartsWithTranslator()
+                    new StartsWithTranslator(logger)
                 };
         }
 
@@ -55,6 +55,6 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators
         /// </summary>
         /// <param name="translators"> The translators. </param>
         protected virtual void AddTranslators([NotNull] IEnumerable<IMethodCallTranslator> translators)
-            => _methodCallTranslators.AddRange(translators);
+            => _methodCallTranslators.InsertRange(0, translators);
     }
 }
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -1411,6 +1411,13 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
 
                     return expression;
                 }
+                case ExpressionType.Negate:
+                {
+                    _relationalCommandBuilder.Append("-");
+                    Visit(expression.Operand);
+
+                    return expression;
+                }
             }
 
             return base.VisitUnary(expression);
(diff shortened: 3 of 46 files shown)


maumar:
@divega @anpete

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionTranslators/RelationalCompositeMethodCallTranslator.cs
```diff
@@ -55,6 +55,6 @@ public virtual Expression Translate(MethodCallExpression methodCallExpression)
         /// </summary>
         /// <param name="translators"> The translators. </param>
         protected virtual void AddTranslators([NotNull] IEnumerable<IMethodCallTranslator> translators)
-            => _methodCallTranslators.AddRange(translators);
+            => _methodCallTranslators.InsertRange(0, translators);
```
maumar: this technically is a breaking change. Current behavior is that we run relational translators first, and after that we run provider-specific translators (so if provider wants to override our relational implementation, they can't - translation is short-circuited after the first translator is able to handle the expression). 

This change reverses the order, to allow providers to override our default implementations with their own. However if a provider already has a buggy implementation that was just not being called ever, now things could start to fail :)

divega:
`EndsWith()` should actually be = 1 instead of > 0, right?

divega:
Ah, it seems all right in the code.

divega:
Couple of additional  thoughts on `EndsWith()`:
- Perhaps reversing the strings is exesive. We could just check if the result of `CHARINDEX()` matches the length of the string minus the length of the pattern. 
- On the other hand, using `REVERSE()` might be clever if you can define a computed index on the reverse of the string, so we could actually use `LIKE` just as with `StartsWith()` (but with the reversed strings)  and there is a chance that the query would be sargable again.

maumar:
@divega that was my initial idea, but it doesn't work if you have multiple occurrences of the pattern. CHARINDEX matches the first one, so we would get some false negatives.

divega:
Ah, good point! Maybe `RIGHT(string, LEN(pattern)) = pattern`? BTW, a similar pattern could work instead of `CHARINDEX()` for `StartsWith()`. 

Anyway, the idea of `LIKE` with `REVERSE()` also seems intriguing. Let's chat?

maumar:
@divega @anpete new version up

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1632,21 +1632,27 @@ protected override Expression VisitBinary(BinaryExpression expression)
                 }
                 else
                 {
+                    Expression newLeft;
```
maumar: @smitpatel  can u take a look at this bit also?

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryBugsTest.cs
```diff
@@ -1321,6 +1323,694 @@ public class Comment5456
...
+                    .ToList().Select(r => new { r.c1.FirstName, r.c2.LastName, r.c1.NullableBool }).OrderBy(r => r.FirstName).ThenBy(r => r.LastName).ToList();
+
+                Assert.Equal(result.Count, expected.Count);
+                for (int i = 0; i < result.Count; i++)
+                {
+                    Assert.True(expected[i].FirstName == result[i].FirstName);
+                    Assert.True(expected[i].LastName == result[i].LastName);
+                }
+
+                Assert.Equal(
```
maumar: there is a discrepancy between CLR semantics and the result we have here - if the EndsWith that takes null as argumens and its compared (!=) to null, we would expect to get positive result, but we get a negative instead. Reason is that EndsWith term is converted to CASE block as it's a search condition, a boolean value is expected, and that case block never returns null. I don't think we can do anything about this case, but it shouldn't be a big deal, because C# would throw in this case anyway :)

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
smitpatel: LGTM.

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerCompositeMethodCallTranslator.cs
```diff
@@ -29,7 +29,10 @@ public class SqlServerCompositeMethodCallTranslator : RelationalCompositeMethodC
             new SqlServerStringTrimEndTranslator(),
             new SqlServerStringTrimStartTranslator(),
             new SqlServerStringTrimTranslator(),
-            new SqlServerConvertTranslator()
+            new SqlServerConvertTranslator(),
+            new SqlServerContainsCharIndexTranslator(),
```
smitpatel: nit: Alphabetical order?

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/FunkyDataQuerySqlServerFixture.cs
```diff
@@ -0,0 +1,66 @@
...
+            optionsBuilder
+                .EnableSensitiveDataLogging()
+                .UseInternalServiceProvider(_serviceProvider)
+                .UseSqlServer(testStore.Connection);
+
+            var context = new FunkyDataContext(optionsBuilder.Options);
+
+            context.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.NoTracking;
+
+            context.Database.UseTransaction(testStore.Transaction);
```
anpete: Need a transaction here?

anpete:
:shipit:

divega:
@maumar should this be "providers-beware" in case they need to provide their own translations/tweak the relational one? In that case can you describe what they will need to do?

Otherwise :shipit:

divega:
One more thing, I think it would be worth running some numbers to understand if there is any impact on perf for SQL Server and SQLite.

ErikEJ:
Please very much providers beware...

maumar:
@divega Will add the providers-beware tag. Existing providers would be affected if they don't want their translators to run (because they are broken) and run our, relational, translators instead. Fix would be to remove those buggy translators from their code.

src/Microsoft.EntityFrameworkCore.Relational/Properties/RelationalStrings.resx
```diff
@@ -198,6 +198,9 @@
   <data name="PossibleUnintendedUseOfEquals" xml:space="preserve">
     <value>Possible unintended use of method Equals(object) for arguments of different types: '{left}', '{right}'. This comparison will always return 'false'.</value>
   </data>
+  <data name="PossibleIncorrectResultsUsingLikeOperator" xml:space="preserve">
+    <value>Translation for method '{function}' used by this database provider can return incorrect results when the value argument contains wildcard characters (e.g. '%' or '_').</value>
```
maumar: @anpete @ajcvickers can you guys take a look at the message?

divega:
@maumar and I looked together at the perf and it is looking good. We also checked that the translation of `String.StrathsWith()` is sargable. 

@maumar has added a warning if the default (quirky) relational translation ends up being used.

Issue #4461, opened 2016-02-02
Change Tracker: Enable using backing fields


rowanmiller:
It is possible to configure an entity to have a getter-only property that has a backing field. The Query Pipeline will write to this field when creating entity instance. For most cases, the ChangeTracker only needs to read property values, but there are some value propagation cases where we need to set property values (generated properties, FK propagation, etc.). We should enable the ChangeTracker to use the backing field for these cases.

patrickmichalina:
:+1: 
This would be a great feature and I currently suspect I am getting bizarre behavior due to the change tracker not being able to use getter-only properties.

Skinney:
Question. Will this work if the backing field is of a different type? I'm using a pattern for validation where I wrap strings and ints in a class that checks for valid input, and allows me to change the format of the input in a simple way. Today this doesn't work so well with EF Core.

Example:

```csharp
public class Email {
  public Email(string email) {
    if (!invalidEmail) {
      throw new IllegalOperationException("invalid email");
    }

    this.Email = email;
  }

  public string ToString() {
    return Email.Trim();
  }

  public string Email { get; set; }
}
```

```csharp
public class Model {
  string email;
  public Email Email {
    get { return email == null ? null : new Email(email); }
    set { email = value.ToString(); }
  }
}
```

The backing field is here of a different type than the public property, but would this work when this issue has been dealt with?

rowanmiller:
@Skinney I think they would all need to be the same. I think what you really want is type conversion so that EF can handle a string from the database mapping to your `Email` type.

ajcvickers:
Triage: I think we should do this in 1.1

PR #6184, opened 2016-07-27
Add change tracker support for read-only properties with backing fields

### LINKED ISSUE #758: When mapping to backing fields handle cases where field is on base type

ajcvickers:
Add change tracker support for read-only properties with backing fields

Issue #4461

Previously the materializer would use the backing field, but then when a value needed to be set for fixup, etc, things would fall down. Now we generate delegates everywhere that can use the backing field.

Note that query still prefers the backing field, while fixup prefers the property.

MemberMapper was made a service of the model. Only the minimal done here--we can allow it to be set by convention or fluent API if the need to customize it is requested.

Also fixed #758 - When mapping to backing fields handle cases where field is on base type


anpete:
:shipit:

PR #6187, opened 2016-07-28
fixing failing test - adding orderby to the query so that results are deterministic

maumar:
Fixes https://github.com/aspnet/EntityFramework/issues/6186


ajcvickers:
:shipit:

PR #6189, opened 2016-07-28
Async Queryable Extension methods should fallback to Synchronous methods

rdefreitas:
Per #6099, streamlining PR at this time to only add the fallback to synchronous methods. Other features will be reopened at a later date.
- Async Linq extensions should fallback to synchronous counterparts


rdefreitas:
@anpete what is the target date for wrapping up 1.1.0?  It seems I may have time to go back and clean up the rest of it in another week or two.

anpete:
@rdefreitas A week or two would be fine :smile:

rdefreitas:
@anpete Alright then. I'll reopen the other request as well, then along with any detailed concerns.

src/Microsoft.EntityFrameworkCore/Extensions/Internal/QueryableExtensions.cs
```diff
@@ -17,6 +17,18 @@ namespace Microsoft.EntityFrameworkCore.Extensions.Internal
     /// </summary>
     public static class QueryableExtensions
     {
+		
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static bool IsAsyncEnumerable<TSource>([NotNull] this IQueryable<TSource> source)
```
anpete: I think we should rather change AsAsyncEnumerable to return null instead of throwing, instead because this approach seems problematic:
1) It doesn't do the test for IAsyncEnumerableAccessor
2) AsAsyncEnumerable will never throw now.
3) We end up testing for IAsyncEnumerable twice.
Sound OK?

rdefreitas: **As**AsyncEnumerable still throws... that method hasn't changed.  IsAsyncEnumerable is a shortcut method that is used inside some of the extension fallbacks (ToDictionaryAsync,ToListAsync). It is possible to remove the AsAsyncEnumerable method altogether and just call `source is IAsyncEnumerable<TSource>` inside of these methods.

I believe that it should still throw if calling AsAsyncEnumerable() on a collection that is not.  The As method feels like an assertion. I'm going to use this object as this object because I know it is one.  If I'm wrong, I deserve my code to blow up in my face (which null would do anyway unless you add a null check everywhere you call the method).

anpete: > AsAsyncEnumerable still throws
It will never throw because we always call IsAsyncEnumerable first. AsAsyncEnumerable is an internal impl. detail and so we are free to change it.
> would then require we add a null check everywhere we call the method
Yes. Consider the idiomatic c# pattern of using the as operator followed by a null check.

rowanmiller:
@rdefreitas just wanted to touch base to see if you were planning to keep working on this contribution?

RainingNight:
@rdefreitas  Any news?

PR #6201, opened 2016-07-28
Logging: Log machine readable structured log entries

### LINKED ISSUE #6096: Log machine readable & queryable structured log entries
Hi @rowanmiller,

Currently EF logs raw strings without leveraging ASP.NET logging extensions. 
[source code](https://github.com/aspnet/EntityFramework/blob/3ec89f76aef3be582ba3c6d5a1a5f6c88a0c3236/src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalLoggerExtensions.cs#L61-L82)
Filtering based on that string is very difficult if not impossible.

I wanted to create a pull request with a fix, I tried to compile the repository for hours but there were mismatch of assemblies I couldn't resolve for some reason.

Anyway the fix seems to be as simple as changing the code to:

    logger.LogInformation((int) RelationalEventId.ExecutedCommand,
        RelationalStrings.RelationalLoggerExecutedCommand,
        string.Format(CultureInfo.InvariantCulture, "{0:N0}", elapsedMilliseconds),
        command.Parameters
            .Cast<DbParameter>()
            .Select(
                p => new DbParameterLogData(
                            p.ParameterName,
                            logParameterValues ? p.Value : "?",
                            logParameterValues,
                            p.Direction,
                            p.DbType,
                            p.IsNullable,
                            p.Size,
                            p.Precision,
                            p.Scale))
            .Select(p => $"{p.Name}={p.FormatParameter()}")
            .Join(),
        command.CommandType,
        command.CommandTimeout,
        command.CommandText.TrimEnd()
    );
    
And change `RelationalStrings.RelationalLoggerExecutedCommand` to:

> Executed DbCommand {{elapsed}}ms [Parameters=[{{parameters}}], {{commandType}', {{commandTimeout}} {{commandText}}

(Or simply use `LoggerMessage.Define` as MVC is doing)

Which will yield a queryable log entry.

image

The first is EF log entry and the second one is using my code (not exactly as above since I couldn't compile the source code).

Now it is possible to query for all the queries which lasted more than x ms.
**Can it please be fixed before 1.1.0?**

Thanks

p.s. Please change from `elapsed` to `ElapsedMilliseconds` to be in sync with [MVC](https://github.com/aspnet/Mvc/blob/283573d6c965413516c2898657628ec8b6395a5d/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreLoggerExtensions.cs#L75)

smitpatel:
Fixes #6096

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Extensions/DbCommandLogData.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Extensions/DbCommandLogData.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
@@ -18,7 +20,8 @@ namespace Microsoft.EntityFrameworkCore.Storage
     ///         to be directly constructed in your application code.    
     ///     </para>
     /// </summary>
-    public class DbCommandLogData
+    [Obsolete("This class is obsolete. It will be removed in a future release.")]
+    public class DbCommandLogData : IEnumerable<KeyValuePair<string, object>>
     {
         /// <summary>
         ///     Initializes a new instance of the <see cref="DbParameterLogData"/> class.
@@ -76,5 +79,16 @@ namespace Microsoft.EntityFrameworkCore.Storage
         ///     Gets how many milliseconds the command took to execute (if it has completed).
         /// </summary>
         public virtual long? ElapsedMilliseconds { get; }
+
+        public virtual IEnumerator<KeyValuePair<string, object>> GetEnumerator()
+        {
+            yield return new KeyValuePair<string, object>("CommandText", CommandText);
+            yield return new KeyValuePair<string, object>("CommandType", CommandType);
+            yield return new KeyValuePair<string, object>("CommandTimeout", CommandTimeout);
+            yield return new KeyValuePair<string, object>("Parameters", Parameters);
+            yield return new KeyValuePair<string, object>("ElapsedMilliseconds", ElapsedMilliseconds);
+        }
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
     }
 }
--- a/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/TestSqlLoggerFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/TestSqlLoggerFactory.cs
@@ -14,6 +14,8 @@ using Xunit.Abstractions;
 using System.Runtime.Remoting.Messaging;
 
 #endif
+
+#pragma warning disable 618
 namespace Microsoft.EntityFrameworkCore.Specification.Tests
 {
     public class TestSqlLoggerFactory : ILoggerFactory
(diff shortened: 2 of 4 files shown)


smitpatel:
@anpete @gdoron @nblumhardt

nblumhardt:
Thanks for the loop-in @smitpatel.

This does address the basic issue of lacking structured data on the event :+1:

As a template for logging in EF going forward (I'm conscious that patterns tend to be replicated through a codebase) I have to say it looks like the hard way of doing things, given the new logging framework will handle all of this stuff itself if you use the template-based API.

E.g., at the [point of usage](https://github.com/aspnet/EntityFramework/blob/3ec89f76aef3be582ba3c6d5a1a5f6c88a0c3236/src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalLoggerExtensions.cs#L35) you could achieve a similar result with:

```csharp
var elapsed = DeriveTimespan(startTimestamp, currentTimestamp);
var paramList = command.Parameters
                       .Cast<DbParameter>()
                       .Select(p => $"{p.Name}={p.FormatParameter()}")
                       .Join();

logger.LogInformation((int)RelationalEventId.ExecutedCommand,
    "Executed DbCommand ({Elapsed} of {CommandTimeout} ms) {Parameters} {CommandType} {CommandText}",
    elapsed,
    command.CommandTimeout,
    paramList,
    command.CommandType,
    command.CommandText.TrimEnd());
```

(I know I've simplified things a fraction by stringifying the parameters here - there are better options for capturing them but not supported by all providers.)

The event passed through the logging pipeline will work the same way, but the `format` implementation will be provided automatically, as will the payload object with the attached structured properties. That cuts quite a bit of code out - the point of logging is simpler, there's one less class to drag around, and one less format method to write. I'd bet that adds up to about 100 lines of code per logging statement, total :-)

As a bonus, logging libraries that understand message templates can grab the original format string and use it to [pretty-print the events](https://nblumhardt.com/2015/06/contender-for-nets-prettiest-console/).

If client code already relies on the current format it may be too late in this instance, but going forward I'd suggest giving the template-based approach more consideration.

Hope this helps!

smitpatel:
@nblumhardt - Thanks for feedback. Template based logging was the first approach I used to solve the issue. As you described, it would make 1 less class to drag around. But considering it is going to 1.1.0 release, removing a class on public surface is breaking change. Therefore in order to keep the current mechanism mostly the same, we went with interface implementation approach.

nblumhardt:
:+1: sounds like the right thing to do.

I can see having the log data objects on the public surface area being quite a challenge over the long term - there are arguments to be made for it, but once log events are part of the API contract even internal refactorings could become breaking changes.

Would adding `[Obsolete("...")]` now be a step towards closing this up in v2?

Cheers,
Nick

gdoron:
👍
@rowanmiller would it be possible for it to be included in 1.0.1?
Without it we have no real way of monitoring our application and EF.
It would be a shame leaving such a great feature aside for 3-5 months when the sole change was making an internal class implement a tiny interface.

What do you guys think?

Thanks for implementing this and @nblumhardt  Nick thanks for your help! 👏

anpete:
@smitpatel Yeah, let's Obsolete the old pattern and file a bug to remove.

rowanmiller:
Agreed, on the Obsolete approach here.

smitpatel:
Updated to add `Obselete`

src/Microsoft.EntityFrameworkCore.Relational/Extensions/DbCommandLogData.cs
```diff
@@ -18,7 +20,8 @@ namespace Microsoft.EntityFrameworkCore.Storage
     ///         to be directly constructed in your application code.    
     ///     </para>
     /// </summary>
-    public class DbCommandLogData
+    [Obsolete("This class is obsolete. It will be removed in future release.", error: false)]
+    public class DbCommandLogData : IEnumerable<KeyValuePair<string, object>>
```
smitpatel: @rowanmiller @divega - Review on obsolete message.

anpete: "in a future release"
Do you need the error arg or is false the default?

anpete:
:shipit:

smitpatel:
filed #6946 for the removal of `DbCommandLogData`

PR #6210, opened 2016-07-30
Throw when an inherited property or navigation is ignored

### LINKED ISSUE #5864: NotMapped on overridden mapped base class property on EF6 throws, but has no effect on EF Core
Model:
```C#
private abstract class AbstractBaseEntity4
{
    public long Id { get; set; }
    public abstract string AbstractBaseClassProperty { get; set; }
}

private class BaseEntity4 : AbstractBaseEntity4
{
    public string BaseClassProperty { get; set; }
    public virtual string VirtualBaseClassProperty { get; set; }
    public override string AbstractBaseClassProperty { get; set; }
}

private class Unit4 : BaseEntity4
{
    [NotMapped]
    public override string VirtualBaseClassProperty { get; set; }
    public virtual AbstractBaseEntity4 Related { get; set; }
}
```
Config:
```C#
modelBuilder.Entity<Unit4>();
modelBuilder.Entity<BaseEntity4>();
```
In EF6, this will throw. In EF Core, the NotMapped attribute is ignored and VirtualBaseClassProperty remains part of the model.

AndriySvyryd:
Fixes #5864


ajcvickers:
:shipit:

Issue #6177, opened 2016-07-27
Query: INNER JOIN generated for navigation traversals from principal to dependents


Bartmax:
This is very counter intuitive, I'm not sure what's going on.
My goal was to get the 4 database entries, with null on Logo when Details is null.
I tried different approaches and i found it very inconsistent.

So obvious question is:  is it possible to get all rows, left join with default if empty?
Question 2: Am I doing something wrong or this is a bug?

Database has 4 entries, 2 with null details.

this query returns 2 results:
```
var query1 = await context.Games
    .Include(g=>g.Details)
    .ThenInclude(d=>d.Logo)
    .Select(g=> new { Logo = g.Details.Logo, g.Title, g.GameId })
    .ToListAsync();
```
<hr>
this query returns 4 results:
```
var query2 = await context.Games
    .Include(g=>g.Details)
    .ThenInclude(d=>d.Logo)
    .Select(g=> new { g.Title, g.GameId })
    .ToListAsync();
```
<hr>
this query throws:
`InvalidOperationException: Operation is not valid due to the current state of the object.
GetArgument`
```
var query3 = await context.Games
    .Include(g=>g.Details)
    .ThenInclude(d=>d.Logo)
    .DefaultIfEmpty()
    .Select(g=> new {  Logo = g.Details.Logo, g.Title, g.GameId })
    .ToListAsync();
```
<hr>
this query throws:
```SqlException: The column prefix 't10' does not match with a table name or alias name used in the query.
No column name was specified for column 1 of 't2'.
Invalid column name 'Title'.
Invalid column name 'GameId'.```

```
 var query4 = await context.Games
    .Include(g=>g.Details)
    .ThenInclude(d=>d.Logo)
    .DefaultIfEmpty()
    .Select(g=> new { Logo = (GameFile)null, g.Title, g.GameId })
    .ToListAsync();
```
<hr>
this query returns 4 elements, same as query2. (logo is explicit set as null):
```
var query5 = await context.Games
    .Include(g=>g.Details)
    .ThenInclude(d=>d.Logo)
    .Select(g=> new { Logo = (GameFile)null, g.Title, g.GameId })
    .ToListAsync();
```
<hr>
note: using `g.Details?.Logo` on lambda does not compile.

Bartmax:
I'm pretty sure the problem lies in that when specifying the navigation inside the `Select` the query switches from `LEFT JOIN` to `JOIN` hence my attempt to use `DefaultIfEmpty`.

[...]

PR #6224, opened 2016-08-01
Fix to #6177 - Query: INNER JOIN generated for navigation traversals from principal to dependents

maumar:
Problem was that during navigation expansion we assumed that when Foreign Key is required, then automatically INNER JOIN could be produced to associate entities within the navigation. This is only true if the navigation is from dependent to principal.
If navigation is from principal to dependent need to produce LEFT OUTER JOIN as it is valid to have a principal entity without dependent in that case.


maumar:
@divega @anpete

PR #6225, opened 2016-08-02
Fix to #6177 - Query: INNER JOIN generated for navigation traversals from principal to dependents

maumar:
Problem was that during navigation expansion we assumed that when Foreign Key is required, then automatically INNER JOIN could be produced to associate entities within the navigation. This is only true if the navigation is from dependent to principal.
If navigation is from principal to dependent need to produce LEFT OUTER JOIN as it is valid to have a principal entity without dependent in that case.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
@@ -53,10 +53,10 @@ ORDER BY [l].[Id]",
             base.Key_equality_using_property_method_required();
 
             Assert.Equal(
-                @"SELECT [l].[Id], [l].[Name], [l].[OneToMany_Optional_Self_InverseId], [l].[OneToMany_Required_Self_InverseId], [l].[OneToOne_Optional_SelfId]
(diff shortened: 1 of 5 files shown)


maumar:
@anpete @divega

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
```diff
@@ -278,25 +278,35 @@ public override void Navigation_key_access_required_comparison()
...
+LEFT JOIN [Level2] AS [e1.OneToOne_Required_FK] ON [e1].[Id] = [e1.OneToOne_Required_FK].[Level1_Required_Id]
+ORDER BY [e1].[Id]",
+                Sql);
+        }
+
+        public override void Optional_navigation_inside_method_call_translated_to_join()
+        {
+            base.Optional_navigation_inside_method_call_translated_to_join();
+
+            Assert.Equal(
```
anpete: ?

maumar: This one is disabled due to bug

anpete:
:shipit:

PR #6229, opened 2016-08-02
[TINY] adding handling of TryExpression to ExpressionPrinter

maumar:
We now wrap some expressions into Try blocks, which currently are not handled well in the ExpressionPrinter. Adding the method to display the body of the TryExpression to make debugging query plans easier

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
@@ -185,6 +185,10 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                     VisitDefault((DefaultExpression)node);
                     break;
 
+                case ExpressionType.Try:
+                    VisitTry((TryExpression)node);
+                    break;
+
                 default:
                     UnhandledExpressionType(node.NodeType);
                     break;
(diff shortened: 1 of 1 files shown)


maumar:
@anpete

maumar:
@smitpatel

src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
```diff
@@ -597,6 +601,23 @@ protected override Expression VisitDefault(DefaultExpression node)
...
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        protected override Expression VisitTry(TryExpression node)
+        {
+            _stringBuilder.Append("try { ");
+            Visit(node.Body);
+            _stringBuilder.Append(" } ");
+            foreach (var handler in node.Handlers)
+            {
+                _stringBuilder.Append("catch (" + handler.Test.Name + ") { ... } ");
```
smitpatel: `AppendLine`?
Whole code seems to have no line breaks while printing.

smitpatel:
:shipit:

src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
maumar: yeah I did that first but it was too verbose. Wrapped expression tend to be one-liners and try-catch block was taking much more space than stuff that is actually significant.

anpete:
@maumar Do we need to show these at all in the query plan?

maumar:
merged https://github.com/aspnet/EntityFramework/commit/f3cdeb429809662b84642b572c9c96418aa2bcf7

PR #6247, opened 2016-08-04
Generate non-temporary values for non-composite string and binary keys when using a relational provider

AndriySvyryd:


src/Microsoft.EntityFrameworkCore.Relational/ValueGeneration/RelationalValueGeneratorSelector.cs
```diff
@@ -45,12 +45,12 @@ public override ValueGenerator Create(IProperty property, IEntityType entityType
...
                 if (propertyType == typeof(string))
                 {
-                    return new StringValueGenerator(generateTemporaryValues: true);
+                    return new StringValueGenerator(generateTemporaryValues: false);
                 }
 
                 if (propertyType == typeof(byte[]))
                 {
-                    return new BinaryValueGenerator(generateTemporaryValues: true);
+                    return new BinaryValueGenerator(generateTemporaryValues: false);
```
ajcvickers: Fall through to core value generator since it will do this anyway.

ajcvickers:
:shipit:

Issue #6268, opened 2016-08-08
Consider whether name matching conventions should match field names


ajcvickers:
If this field is mapped:

``` C#
public class Foo
{
   private int _id;
}
```

then is will not be made the key because it does not match the property naming conventions for keys.

divega:
**Triage:** we don't believe we need to recognize "_id" as a key name, but we will monitor feedback on this.

Issue #6267, opened 2016-08-08
Decide how data annotations on fields should work


ajcvickers:
Currently data annotations on fields are used if and only if the field is mapped and has no property associated. For example, the Key attribute will be applied in this case:

``` C#
public class Foo
{
    [Key]
    private int _key;
}
```

assuming _key is a mapped field:

``` C#
builder.Entity<Foo>().Property("_key");
```

The data annotation will not be applied in this case:

``` C#
public class Foo
{
    [Key]
    private int _key;

    public int Key => _key;
}
```

It is instead expected that the data annotation will be taken from the property.

If we decide to allow data annotations to be applied to fields that have associated properties, then we also need to decide what to do when the field and the property have conflicting annotations:

``` C#
public class Foo
{
    [Column("KEY")]
    private int _key;

    [Column("ID")]
    public int Key => _key;
}
```

divega:
We like it the way @ajcvickers implemented it, e.g. an annotation on the property always override the one on the field.

Issue #6266, opened 2016-08-08
Provide a data annotation for setting backing fields


ajcvickers:
This would allow a property to be annotated with a backing field directly in the class. One advantage of this is that the nameof() syntax could be used to reference the field.

lajones:
@ajcvickers Let me know what you think of this design for the new attribute (I was thinking of putting it in EFCore.Abstractions next to `OwnedAttribute`:

```C#
namespace Microsoft.EntityFrameworkCore
{
    /// <summary>
    ///     Names the backing field associated with this property.
    /// </summary>
    [AttributeUsage(AttributeTargets.Property)]
    public sealed class BackingFieldAttribute : Attribute
    {
        /// <summary>
        ///     Initializes a new instance of the <see cref="BackingFieldAttribute" /> class.
        /// </summary>
        /// <param name="name">The name of the backing field.</param>
        public BackingFieldAttribute([NotNull] string name)
        {
            Check.NotEmpty(name, nameof(name));

            Name = name;
        }

        /// <summary>
        ///     The name of the backing field.
        /// </summary>
        public string Name { get; }
    }
}
```

ajcvickers:
@dotnet/efcore Do people think this should be its own annotation? Or should we have one attribute for property mapping and facets? Or one attribute for property mapping, one for facets?

smitpatel:
What is property mapping? And which facets are we planning to add annotations for?

ajcvickers:
Maybe we should make a list of all the things that can be configured for with the fluent API for which there are no attributes?

lajones:
Not sure this is exhaustive, but a quick look at [the docs](https://docs.microsoft.com/en-us/ef/core/modeling/) shows the following Fluent API with no attribute equivalent (note: some of these don't apply to properties and so may not be relevant here):

- configuring composite keys
- name of PK/FK constraints
- `HasPrincipalKey()`
- `HasAlternateKey()`
- cascade delete
- `HasIndex()` (and its name and included properties)
- `HasDiscriminator()`
- `HasSequence()`
- `HasField()` and `UsePropertyAccessMode()`
- value conversions
- data seeding
- `HasNoKey()`

lajones:
Notes from team discussion 2/19/20: 
In general we prefer small individual attributes over attributes which represent a combination of facets, e.g. we'd prefer `Unicode` and `Precision` as separate attributes, but we may make exceptions where concepts are closely tied together, e.g. Precision and Scale might well end up together.

[...]

Issue #4855, opened 2016-03-19
Add a way to configure property materialization to not use the backing field


AndriySvyryd:
This would allow to have logic in the property that is executed on materialization

PR #6269, opened 2016-08-08
Support mapping to fields

ajcvickers:
Issues #4357, #4855

This change adds fluent and lower-level APIs to set backing fields or properties with no fields as well as conventions for using the fields and a mechanism to provide control over when the property is used and when the field is used.

Also see #6266, #6267, and #6268.

/cc @rowanmiller for API and docs


anpete:
@ajcvickers Any guidance on specific areas of this to review?

ajcvickers:
@anpete Probably the biggest change is the way fields are found and set in metadata when using the convention or the API. Then possibly the rules for choosing a field or a property based on mode.

anpete:
:shipit:

src/Microsoft.EntityFrameworkCore/Metadata/Internal/PropertyBaseExtensions.cs
```diff
@@ -145,6 +147,179 @@ public static PropertyInfo GetPropertyInfo([NotNull] this IPropertyBase property
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static object FindMemberInfo(
+            [NotNull] this IPropertyBase propertyBase,
+            bool forConstruction,
+            bool forSet)
+        {
```
AndriySvyryd: Consider using an out parameter or return a Tuple

src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -1153,9 +1153,7 @@ private bool CanSetUnique(bool unique, ConfigurationSource? configurationSource,
                     && Metadata.DeclaringEntityType == dependentEntityType)
                 {
                     Debug.Assert(!shouldResetToPrincipal
-                                 && !shouldResetToDependent
-                                 && !shouldResetPrincipalProperties
-                                 && !shouldResetDependentProperties);
```
AndriySvyryd: Note for myself to revision this assert

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/BackingFieldConvention.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected virtual FieldInfo TryMatchFieldName(
+            [NotNull] Type entityType, [NotNull] Type propertyType, [NotNull] string propertyName)
+        {
+            var fields = entityType.GetRuntimeFields().ToDictionary(f => f.Name);
```
AndriySvyryd: We'll likely need to cache this

AndriySvyryd:
:shipit:

Issue #6261, opened 2016-08-07
DateTime converted to Nullable<DateTime> inside optional navigation property, throws exception on AddDays


HydrateEfficiently:
I believe this is a regression from RC1 or RC2 (I migrated from RC1 => RTM).

Trying to query all the orders completed within the last day:
(Note: Placement is the optional navigational property and, in the context of my solution, if it exists it indicates than the order has been placed)

```C#
var ordersCompleted = await dbContext.Orders
                .Where(o => o.IsCompleted)
                .Where(o => o.Error == OrderErrorType.None)
                .Where(o => o.Placement.Placed.AddDays(1) > now)
                .ToListAsync();
```
**Exception:**

Message: Method 'System.DateTime AddDays(Double)' declared on type 'System.DateTime' cannot be called with instance of type 'System.Nullable`1[System.DateTime]'

```
StackTrace: 
   at System.Linq.Expressions.Expression.ValidateCallInstanceType(Type instanceType, MethodInfo method)
   at System.Linq.Expressions.Expression.ValidateStaticOrInstanceMethod(Expression instance, MethodInfo method)
   at System.Linq.Expressions.Expression.Call(Expression instance, MethodInfo method, IEnumerable`1 arguments)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.VisitMethodCall(MethodCallExpression node)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.VisitBinary(BinaryExpression node)
   at Remotion.Linq.Clauses.WhereClause.TransformExpressions(Func`2 transformation)
   at Remotion.Linq.QueryModelVisitorBase.VisitBodyClauses(ObservableCollection`1 bodyClauses, QueryModel queryModel)
   at Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.Rewrite(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.OptimizeQueryModel(QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.CreateAsyncQueryExecutor[TResult](QueryModel queryModel)
[...]

PR #6274, opened 2016-08-09
Issue6261 - just test to reproduce

jenseralmeida:
Code to reproduce the issue 6261. No correction is included.

The issue was inserted by commit c4e44471 when rewriting the expression from:
`
o.Customer.Birthday.AddDays(1) > now
`
to:
`
(o.Customer != null ? o.Customer.Birthday : null).AddDays(1) > now
`
Putting in perspective, the expression rewrite was inserted to support LOJ when using some property of a optional navigation entity. More details in NavigationRewritingExpressionVisitor.RewriteNavigationsIntoJoins, lines 765 (when propertyType.MakeNullable() is callend) and 772 (when the expression is rewrited to support null check before access the property)
 
Reproduces, but not fixes, #6261


maumar:
closing this as we will have tests added along with the fix - tests will be added in ComplexNavigations model, since Northwind is somewhat static, well established model/database and we are unwilling to change it.

PR #6287, opened 2016-08-10
[TINY] SQL Server Migrations: Avoid redundant WHERE..NOT NULL on unique indices

### LINKED ISSUE #5532: Migrations: Avoid redundant IS NOT NULL on indexes
### Steps to reproduce
Create an entity with string property that got an unique-index. Field is marked as required.
```
builder.Entity<UnitOfMeasurementEntity>(b =>
            {
                b.ToTable("UnitOfMeasurementEntity", "Products");
                b.HasKey(p => p.SystemId);

                b.Property(p => p.Id)
                    .HasMaxLength(100)
                    .IsRequired();
                b.HasIndex(p => p.Id).IsUnique();
            });
```

### The issue
Index is created on non null column, index is still created as a filtered index with an where statement. Somehow the filtered index is not used during all querying and got slow response time from SQL server (SQL 2012r2).
```
IF NOT EXISTS(SELECT * FROM [__EFMigrationsHistory] WHERE [MigrationId] = N'20160526125022_FixBrokenIndex2')
BEGIN
    CREATE UNIQUE INDEX [IX_UnitOfMeasurementEntity_Id] ON [Products].[UnitOfMeasurementEntity] ([Id]) WHERE [Id] IS NOT NULL;
END
```

Migration code looks like the following
```
            migrationBuilder.CreateIndex(
                name: "IX_UnitOfMeasurementEntity_Id",
                schema: "Products",
                table: "UnitOfMeasurementEntity",
                column: "Id",
                unique: true);
```

The column for the index is non nullable and then the index does not need any filter.

### Further technical details

EF Core version: RC1
Operating system: Win10
Visual Studio version: VS2015 RC2

bricelam:
Resolves #5532

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 using System.Text;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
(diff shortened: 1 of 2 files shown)


natemcmaster:
:shipit:

PR #6289, opened 2016-08-10
[TINY] SQL Server Migrations: Ignore default value of added identity columns

### LINKED ISSUE #6035: AddColumn migrations created with bad default
Hope I'm not reporting a dup here.

Testing with SqlServer, I've added this property to my model:

```c#
builder.Entity<Blog>().Property(b => b.SomeProp).ValueGeneratedOnAdd();
```

Generating a migration creates the following code:

```c#
	name: "SomeProp",
	table: "Blogs",
	nullable: false,
	defaultValue: 0)
	.Annotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn);
```

Note the `defaultValue: 0`, which I didn't specify in the model in any way. Note also that if the column is defined when creating the table, the correct CreateTable migration is generated:

```c#
SomeProp = table.Column<int>(nullable: false)
    .Annotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn),
```

This causes an issue with Npgsql (https://github.com/npgsql/Npgsql.EntityFrameworkCore.PostgreSQL/issues/68) since the default value is examined in the migration decision process.

bricelam:
Fixes #6035

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
@@ -435,7 +435,9 @@ namespace Microsoft.EntityFrameworkCore.Migrations
                 maxLength,
                 rowVersion,
                 nullable,
-                defaultValue,
+                identity
+                    ? null
+                    : defaultValue,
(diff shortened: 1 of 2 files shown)


AndriySvyryd:
:shipit:

Issue #6292, opened 2016-08-10
Partially store-generated keys cannot be inserted on SQLite


ajcvickers:
Key has one property store-generated, one property not.

``` C#
public class Category
{
    public int Id1 { get; set; }
    public int Id2 { get; set; }
}

public class CompKeyContext : DbContext
{
    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        => optionsBuilder.UseSqlite("Data Source=CompKeyContext");

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<Category>(b =>
        {
            b.HasKey(e => new { e.Id1, e.Id2 });
            b.Property(e => e.Id1).ValueGeneratedOnAdd();
        });
    }
}

public class Program
{
    public static void Main()
    {
        using (var db = new CompKeyContext())
        {
            db.Database.EnsureDeleted();
            db.Database.EnsureCreated();

            db.Add(new Category { Id2 = 77 });

            db.SaveChanges();
        }
    }
}
```

Exception:

```
Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> Microsoft.Data.Sqlite.SqliteException: SQLite Error 19: 'NOT NULL constraint failed: Category.Id1'.
   at Microsoft.Data.Sqlite.Interop.MarshalEx.ThrowExceptionForRC(Int32 rc, Sqlite3Handle db)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at System.Data.Common.DbCommand.ExecuteReader()
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.Execute(IRelationalConnection connection, String executeMethod, IReadOnlyDictionary`2 parameterValues, Boolean openConnection, Boolean closeConnection)
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.ExecuteReader(IRelationalConnection connection, IReadOnlyDictionary`2 parameterValues, Boolean manageConnection)
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)
   --- End of inner exception stack trace ---
[...]

PR #6293, opened 2016-08-11
Handle fixup of overlapping composite keys which get propagated store generated values

ajcvickers:
Issue #6212

This happens only in cases where the FK property is getting a store generated value by propagation, and this FK property is shared between two relationships, and if the metadata is built in a way that the propagation happens in a certain order. The fix is to to also use the pre-store generated value when looking up the new principal since this may not have been updated yet in the dependent map.

Also added more coverage for store-generated composite keys in general, and added some extension methods to make it easier to dump out a human-readable form of the model.

Also, see #6292, uncovered will testing.


AndriySvyryd:
:shipit:

PR #6294, opened 2016-08-11
Add foreign key indexes using a convention

### LINKED ISSUE #5665: duplicate index created when PK is also a FK
### Steps to reproduce

Consider the following many-to-many definition, with an additional property in the MTM table so we need the intermediate entity:

    public class Foo
    {
        public int Id { get; set; }        
        public ICollection<FooBar> FooBars { get; set; }
    }

    public class Bar
    {
        public int Id { get; set; }
        public ICollection<FooBar> FooBars { get; set; }
    }

    public class FooBar
    {
        public int FooId { get; set; }
        public int BarId { get; set; }
        public string AdditionalProperty { get; set; }
    }

`OnModelCreating` contains:

            builder.Entity<Foo>(b =>
            {
                b.HasKey(x => x.Id);
                b.HasMany(x => x.FooBars).WithOne().HasForeignKey(x => x.FooId).IsRequired();
            });

            builder.Entity<Bar>(b =>
            {
                b.HasKey(x => x.Id);
                b.HasMany(x => x.FooBars).WithOne().HasForeignKey(x => x.BarId).IsRequired();
            });

            builder.Entity<FooBar>(b =>
            {
                b.HasKey(x => new { x.FooId, x.BarId });
            });

### The issue

One would expect the primary key of `FooBars` to be `FooId, BarId`, and one more index is created due to the FK for `BarId`. The foreign key `FooId REFERENCES Foos(Id)` should *not* create an additional index, because it is already covered by the PK.

However, the generated migration code is:

            migrationBuilder.CreateTable(
                name: "Bars",
                columns: table => new
                {
                    Id = table.Column<int>(nullable: false)
                        .Annotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn)
                },
                constraints: table =>
                {
                    table.PrimaryKey("PK_Bars", x => x.Id);
                });

            migrationBuilder.CreateTable(
                name: "Foos",
                columns: table => new
                {
                    Id = table.Column<int>(nullable: false)
                        .Annotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn)
                },
                constraints: table =>
                {
                    table.PrimaryKey("PK_Foos", x => x.Id);
                });

            migrationBuilder.CreateTable(
                name: "FooBars",
                columns: table => new
                {
                    FooId = table.Column<int>(nullable: false),
                    BarId = table.Column<int>(nullable: false),
                    AdditionalProperty = table.Column<string>(nullable: true)
                },
                constraints: table =>
                {
                    table.PrimaryKey("PK_FooBars", x => new { x.FooId, x.BarId });
                    table.ForeignKey(
                        name: "FK_FooBars_Bars_BarId",
                        column: x => x.BarId,
                        principalTable: "Bars",
                        principalColumn: "Id",
                        onDelete: ReferentialAction.Cascade);
                    table.ForeignKey(
                        name: "FK_FooBars_Foos_FooId",
                        column: x => x.FooId,
                        principalTable: "Foos",
                        principalColumn: "Id",
                        onDelete: ReferentialAction.Cascade);
                });

            migrationBuilder.CreateIndex(
                name: "IX_FooBars_BarId",
                table: "FooBars",
                column: "BarId");

            migrationBuilder.CreateIndex(
                name: "IX_FooBars_FooId",
                table: "FooBars",
                column: "FooId");

Note the index `IX_FooBars_FooId` which is being created.

### Further technical details

Operating system: Win 7 x64
Visual Studio version: VS 2015 Update 2

AndriySvyryd:
Avoid adding indexes when covered by keys or other indexes

Fixes #5665

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Internal/EnumerableExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/Internal/EnumerableExtensions.cs
@@ -73,20 +73,54 @@ namespace Microsoft.EntityFrameworkCore.Internal
                 return true;
             }
 
-            var firstEnumerator = first.GetEnumerator();
-            var secondEnumerator = second.GetEnumerator();
-
-            while (firstEnumerator.MoveNext())
+            using (var firstEnumerator = first.GetEnumerator())
             {
-                if (!secondEnumerator.MoveNext()
-                    || !StructuralComparisons.StructuralEqualityComparer
-                        .Equals(firstEnumerator.Current, secondEnumerator.Current))
+                using (var secondEnumerator = second.GetEnumerator())
                 {
-                    return false;
+                    while (firstEnumerator.MoveNext())
+                    {
+                        if (!secondEnumerator.MoveNext()
+                            || !StructuralComparisons.StructuralEqualityComparer
+                                .Equals(firstEnumerator.Current, secondEnumerator.Current))
+                        {
+                            return false;
+                        }
+                    }
+
+                    return !secondEnumerator.MoveNext();
+                }
+            }
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static bool StartsWith<TSource>(
+            [NotNull] this IEnumerable<TSource> first,
+            [NotNull] IEnumerable<TSource> second)
+        {
+            if (ReferenceEquals(first, second))
+            {
+                return true;
+            }
+
+            using (var firstEnumerator = first.GetEnumerator())
+            {
+                using (var secondEnumerator = second.GetEnumerator())
+                {
+                    while (secondEnumerator.MoveNext())
+                    {
+                        if (!firstEnumerator.MoveNext()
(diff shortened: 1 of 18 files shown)


src/Microsoft.EntityFrameworkCore/Internal/EnumerableExtensions.cs
```diff
@@ -88,5 +88,33 @@ public DynamicEqualityComparer(Func<T, T, bool> func)
...
+        public static bool StartsWith<TSource>(
+            [NotNull] this IEnumerable<TSource> first,
+            [NotNull] IEnumerable<TSource> second)
+        {
+            if (ReferenceEquals(first, second))
+            {
+                return true;
+            }
+
+            var firstEnumerator = first.GetEnumerator();
```
divega: If this is supposed to be somewhat general purpose (even if internal) consider "using" the enumerators so that they get disposed.

ajcvickers:
:shipit:

src/Microsoft.EntityFrameworkCore/Internal/EnumerableExtensions.cs
AndriySvyryd: ok

PR #6300, opened 2016-08-11
Add 'Drop-Database'

natemcmaster:
Close https://github.com/aspnet/EntityFramework/issues/2476

cc @bricelam

### DIFF
--- a/src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
+++ b/src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
@@ -68,7 +68,7 @@ Register-TabExpansion Add-Migration @{
     Context = { param ($tabExpansionContext) GetContextTypes $tabExpansionContext.Project $tabExpansionContext.StartupProject $tabExpansionContext.Environment }
     Project = { GetProjects }
     StartupProject = { GetProjects }
-    # Disables tab completion on output dir. 
+    # Disables tab completion on output dir.
     # PMC will show completion options relative to the solution directory, but OutputDir should be relative to project
     OutputDir = { }
 }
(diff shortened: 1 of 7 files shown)


src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -146,6 +146,90 @@ function Add-Migration {
...
+    
+.PARAMETER Force
+    Drops the database without confirmation.
+
+.LINK
+    Update-Database
+    about_EntityFrameworkCore
+#>
+function Drop-Database {
+    [CmdletBinding(PositionalBinding = $false)]
```
bricelam: `SupportsShouldProcess = $true`

This is the idiomatic way of doing confirmation and dry runs in PowerShell

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -146,6 +146,90 @@ function Add-Migration {
...
+
+    $yes = New-Object System.Management.Automation.Host.ChoiceDescription "&Yes", `
+        "Drops the database"
+
+    $no = New-Object System.Management.Automation.Host.ChoiceDescription "&No", `
+        "Cancels the command without dropping the database"
+
+    $options = [System.Management.Automation.Host.ChoiceDescription[]]($yes, $no)
+
+    $result = $host.ui.PromptForChoice($msg, "Are you sure you want to proceed?", $options, 1)
```
bricelam: Use `$PSCmdlet.ShouldProcess` instead.

natemcmaster:
Updated to be idiomatic. Here is what the different options look like.

image

rhires:
Should the default to drop a database be [Y]?

natemcmaster:
I understand the concern, but IMO it's safe to assume that users intend to drop the database if they typed the command 'Drop-Database'.

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -146,6 +146,80 @@ function Add-Migration {
...
+        throw 'Drop-Database should not be used with Universal Windows apps. Instead, call DbContext.Database.EnsureDeleted() at runtime.'
+    }
+
+    $info = InvokeOperation $values -json database drop --dry-run
+
+    ShowConsole
+
+    if (!($info)) {
+        Write-Output 'Could not find a database to remove'
+    } elseif ($Force -or
```
bricelam: Using `-Force` for this in PowerShell is an anti-pattern. Instead, the user would either set `$ConfirmPreference` or use the `-Confirm:$false` default parameter.

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -146,6 +146,80 @@ function Add-Migration {
...
+
+.PARAMETER Force
+    Drops the database without confirmation.
+
+.LINK
+    Update-Database
+    about_EntityFrameworkCore
+#>
+function Drop-Database {
+    [CmdletBinding(PositionalBinding = $false, SupportsShouldProcess = $true)]
```
bricelam: May want to also set `ConfirmImpact = 'High'` so it prompts by default in PMC.

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -146,6 +146,80 @@ function Add-Migration {
...
+    }
+
+    $info = InvokeOperation $values -json database drop --dry-run
+
+    ShowConsole
+
+    if (!($info)) {
+        Write-Output 'Could not find a database to remove'
+    } elseif ($Force -or
+        $PSCmdlet.MyInvocation.BoundParameters['WhatIf'].IsPresent -or
```
bricelam: `-WhatIf` is handled by `ShouldContinue`. It will show a differnt message and return `$false` without prompting.

src/Tools.VisualStudio/tools/EntityFrameworkCore.psm1
```diff
@@ -146,6 +146,80 @@ function Add-Migration {
...
+        $PSCmdlet.ShouldContinue(
```
bricelam: This needs to be `ShouldProcess`. `ShouldContinue` is used when processing multiple things.

natemcmaster:
Updated. Here's take 3.

image

bricelam:
:shipit:

Issue #6306, opened 2016-08-12
Make replacing internal EF services easier in the common case


ajcvickers:
Currently when an application needs to replace an internal EF service it must take over creation and management of the internal service provider. This is a significant burden as can be seen in answers for #5845 and #5303. It can also result in bad code being written when the caching requirements for the service provider are not understood--for example #5866.

It would be useful to have an API that can be called in OnConfiguring that would allow a service to be easily replaced without the application needing to take over managing the service provider.

PR #6307, opened 2016-08-12
Make it easier to replace internal EF services

ajcvickers:
Issue #6306

Adds a ReplaceService method to DbContextOptionsBuilder that is used like this:

```C#
protected internal override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
  => optionsBuilder
      .ReplaceService<IModelCustomizer, CustomModelCustomizer>()
      .UseInMemoryDatabase();
```

The replacement service automatically gets the scoping of the original registration.

An exception is thrown if this method is used in combination with UseInternalServiceProvider.

\cc @rowanmiller for API, documentation, and exception message


AndriySvyryd:
:shipit:

PR #6311, opened 2016-08-12
[tiny] Hide stack trace for common migrations operation failures

natemcmaster:
We don't need to show the stack trace for these three errors, so we should throw OperationException instead.

cc @bricelam


bricelam:
:shipit:

Issue #5346, opened 2016-05-12
Re-enable all skipped/disabled tests


rowanmiller:
Need to do a general search of the code base, but also make sure we cover these.
- https://github.com/aspnet/EntityFramework/issues/4841
- https://github.com/aspnet/EntityFramework/issues/5172

ajcvickers:
@natemcmaster So we really have zero skipped tests after this change?

lajones:
I'm not seeing any skipped tests for the specifically-referenced areas any more. And most of this has been moved/consolidated since this bug was created. On the other hand there are many, many [ConditionalFact] tests. Looking at all of them would probably need a team push.

smitpatel:
superseded by #7160

PR #6312, opened 2016-08-12
Re-enable skipped design-time tests and remove dead tests

natemcmaster:
Progress towards https://github.com/aspnet/EntityFramework/issues/5346


ajcvickers:
@natemcmaster So we really have zero skipped tests after this change?

ajcvickers:
Looks like there are skipped tests in:
* InheritanceRelationshipsQueryTestBase
* OptimisticConcurrencyTestBase
* FieldMappingTestBase
* AsyncQueryTestBase
* QueryTestBase
* ComplexNavigationsQueryTestBase
* GearsOfWarQueryTestBase

natemcmaster:
Sorry, should have been more clear. I mean we are re-enabling **some** skipped tests, not all.

ajcvickers:
@natemcmaster That's great, but then it doesn't resolve the issue assigned to Lawrence which was to scan the code for all skipped tests and either re-enable them or check that there is a bug associated with the test and that the skipped test is mentioned from that bug.

natemcmaster:
I edit the description above so #5346 is still open when the PR merges. This re-enabled tests here are only for the tests concerning design-time tools

ajcvickers:
:shipit:

PR #6327, opened 2016-08-15
Add missing assembly descriptions to tooling libraries

natemcmaster:
Unblocks ASP.NET CI

cc @ajaybhargavb @pranavkm


ajaybhargavb:
:shipit:

pranavkm:
Add it to the project.json instead. See https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Design/project.json#L3

natemcmaster:
We don't use project.json on these projects to pack.

natemcmaster:
@pranavkm to clarify...does it matter if the description is in code or in project.json? AFAIK the only reason to have it in project.json is so dotnet-pack picks this up in addition to generating AssemblyDescriptionAttribute. But since we use nuspec instead, does it matter?

pranavkm:
Fair enough. :shipit:

Issue #6330, opened 2016-08-15
Query :: null compensation logic is not working for some cases where non-nullable type is expected


maumar:
E.g.:

```
context.Tags.Where(t => t.Gear.HasSoulPatch)
```
or
```
context.Tags.Where(t => t.Gear.HasSoulPatch ? true : false)
```

Navigation expansion converts ```t.Gear.HasSoulPatch``` expression to ```bool?``` and we fail to construct Expression, because ```bool``` is expected

maumar:
Fixed in 3f532a773a453b8f77e06e169ff9e5061618662b

Issue #5613, opened 2016-06-01
Query :: optional navigations null-ref protection logic doesn't work for some cases involving multiple optional navs and method calls


maumar:
Found during investigation of https://github.com/aspnet/EntityFramework/issues/5481

model:

```
    public class Shipment
    {
        public int Id { get; set; }
        public string CarrierName { get; set; }
        public int? CargoReportId { get; set; }
        public Report CargoReport { get; set; }
        public int? AltCargoReportId { get; set; }
        public Report AltCargoReport { get; set; }
    }

    public class Report
    {
        public int Id { get; set; }
        public string Reference { get; set; }
    }

    public class MyContext : DbContext
    {
        public DbSet<Shipment> Shipments { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer("Server=.;Database=Repro5613;Trusted_Connection=True;MultipleActiveResultSets=true;");
        }
    }

```

query:

```
                var query = ctx.Shipments
                     .Where(s => s.CargoReport.Reference.Contains("whatever")
                             || s.AltCargoReport.Reference.Contains("whatever"));
```

relevant query plan fragment:

```
(t3.Outer.Outer.Inner != null ? t3.Outer.Outer.Inner.Reference : null).Contains(
            value: whatever
        )
```

this will null ref if Outer.Outer.Inner is null

maumar:
moving to 1.0.1 per triage comment in https://github.com/aspnet/EntityFramework/issues/6177

divega:
Pulling from 1.0.1 into 1.1.0 because the fix has grown complex enough that we are not longer confident we should include it in a patch release. 

cc @maumar @Eilon

maumar:
Fixed in https://github.com/aspnet/EntityFramework/commit/7472a679e1258ee4c77408bcb59817b277d6126a

PR #6331, opened 2016-08-15
Fix to #6330 - Query :: null compensation logic is not working for some cases where non-nullable type is expected

maumar:
Problem was that we were not compensating for nullability change properly for some cases, leading to compilation errors is some cases and runtime errors in others.

e.g.
context.Tags.Where(t => t.Gear.HasSoulPatch)

would get translated to:
context.Tags.Where(t => (bool)(t.Gear != null ? t.Gear.HasSoulPatch : null))

but now we will translate it to:
context.Tags.Where(t => (t.Gear != null ? t.Gear.HasSoulPatch : null) == true)

This ensures type consistency as well as prevents exception being thrown at runtime, when we would sometimes try to cast null value to bool.
We do similar rewrite for other "search conditions"

Also modified one of the test models and added tests for #5613, and fixed a few minor bugs.


maumar:
@smitpatel  @divega

smitpatel:
:shipit:

PR #6341, opened 2016-08-17
[1.0.1] Throw a better exception when shadow keys are detected

### LINKED ISSUE #6260: Mismatched property types used in a relationship
I noticed this using v1.0.0

If you specify a many to many relationship between three classes and the property types don't match:

```
public class Class1
{
    public int ID { get; set; }
}

public class Class2
{
    public int ID { get; set; }
}
public class Class2
{
    public string Class1ID { get; set; }
    public int Class2ID { get; set; }
}

// relationship defined in OnModelCreating
```

you'll get a rather obscure error message referencing shadow properties, when the problem is really that the property types don't match.

It would be nice if the code building the relationship noted that the types don't match.

AndriySvyryd:
Reorder the validation rules so this exception is thrown less frequently.

Fixes #6260

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
@@ -148,7 +148,7 @@
     <value>The collection argument '{argumentName}' must contain at least one element.</value>
   </data>
   <data name="EntityRequiresKey" xml:space="preserve">
-    <value>The entity type '{entityType}' requires a key to be defined.</value>
+    <value>The entity type '{entityType}' requires a primary key to be defined.</value>
   </data>
   <data name="KeyPropertiesWrongEntity" xml:space="preserve">
     <value>The specified key properties {key} are not declared on the entity type '{entityType}'. Ensure key properties are declared on the target entity type.</value>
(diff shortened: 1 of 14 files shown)


src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -349,10 +349,7 @@
     <value>The entity type '{type}' provided for the argument '{argumentName}' must be a reference type.</value>
   </data>
   <data name="ReferencedShadowKey" xml:space="preserve">
-    <value>The key {key} contains properties in shadow state and is referenced by a relationship from '{referencingEntityTypeWithNavigation}' to '{referencedEntityTypeWithNavigation}'. Configure a non-shadow principal key for this relationship.</value>
-  </data>
-  <data name="ReferencedShadowKeyWithoutNavigations" xml:space="preserve">
-    <value>The key {key} on entity type '{entityType}' contains properties in shadow state and is referenced by the foreign key {foreignKey} from entity type '{referencingEntityType}'. Configure a non-shadow principal key for this relationship.</value>
+    <value>The relationship from '{referencingEntityTypeOrNavigation}' to '{referencedEntityTypeOrNavigation}' defined on {foreignKeyProperties} references a key that contains properties in shadow state, because the primary key is not compatible with the foreign key properties. Configure a non-shadow principal key for this relationship or a set of compatible foreign key properties.</value>
```
ajcvickers: Do we need to mention shadow keys at all here? Maybe:

> No primary key compatible with the foreign key type was found for the relationship from '{referencingEntityTypeOrNavigation}' to '{referencedEntityTypeOrNavigation}' with foreign key properties {foreignKeyProperties} Configure a principal key or a set of compatible foreign key properties for this relationship.

Because it seems very unlikely that anyone will get this message for anything other than us creating the shadow PK for them, and the shadow stuff in the message is very distracting/confusing to people.

ajcvickers:
:shipit:

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
AndriySvyryd: By this point we know that a primary key exists, so how about this?
> The relationship from '{referencingEntityTypeOrNavigation}' to '{referencedEntityTypeOrNavigation}' with foreign key properties {foreignKeyProperties} cannot target the primary key because it is not compatible. Configure a principal key or a set of compatible foreign key properties for this relationship.

ajcvickers: I like it except that what does "is not compatible" mean? Usually the wrong type, possibly wrong number of properties.

AndriySvyryd: We could include the types as well. So formatted it would look like
> The relationship from 'OrderDetails.Order' to 'Order.OrderDetails' with foreign key properties 'OrderId : int, OrderUniqueId : Guid' cannot target the primary key 'Id : string' because it is not compatible. Configure a principal key or a set of compatible foreign key properties for this relationship.

ajcvickers: :shipit:

divega: :+1:

Issue #6337, opened 2016-08-16
Query: Projected values can be the wrong type


ajcvickers:
Consider this query:
```C#
var values = context.Drivers.AsNoTracking()
    .Where(e => e.Id == larry.Id)
    .Select(e => new object[] { e.Id, e.Name, e.Poles })
    .FirstOrDefault();
```
The types of properties Id, Name, and Poles are int, string, int respectively. So the expectation is that the projected values in the array would be of those types. However, on SQLite, the values are of type long, string, long. That is, the types are the defaults from the store and not the requested types. I expect the same issue would exist on SQL Server in places where the type is relevant.

Note that if I let LINQ to Objects to do the projection:
```C#
var values = context.Drivers.AsNoTracking()
    .Where(e => e.Id == larry.Id)
    .ToList()
    .Select(e => new object[] { e.Id, e.Name, e.Poles })
    .FirstOrDefault();
```
then the types in the array are int, string, int, as expected.

However, in the real code I can't use LINQ to Objects for this because I am using EF.Property calls to project shadow values.

This bug currently prevents GetDatabaseValues from working on SQLite.

ajcvickers:
Poaching this.

Issue #1200, opened 2014-12-02
Implement CurrentValues, OriginalValues, and GetDatabaseValues on EntityEntry


ajcvickers:

leo90skk:
when will this feacture be implemented? I need it for a update function in an update method in a service (N-Tire architecture).

rowanmiller:
Currently planned for 1.1 - see https://blogs.msdn.microsoft.com/dotnet/2016/07/29/entity-framework-core-1-1-plans/ for details.

dstep2007:
I see 1.1 was just released.  Did this make it into the release?

smitpatel:
@dstep2007 - Yes.

PR #6348, opened 2016-08-17
Implement CurrentValues, OriginalValues, and GetDatabaseValues

ajcvickers:
Issue #1200

Pretty much the same design as EF6. Exposed full property metadata instead of names since we now have nice metadata. Tests were ported from EF6. Not fully functional, especially on SqLite, until #6337 is fixed.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/EntityEntry.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/EntityEntry.cs
@@ -4,6 +4,8 @@
 using System;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
 using Microsoft.EntityFrameworkCore.Infrastructure;
@@ -27,6 +29,8 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking
     public class EntityEntry : IInfrastructure<InternalEntityEntry>
     {
         private static readonly int _maxEntityState = Enum.GetValues(typeof(EntityState)).Cast<int>().Max();
+
+        private IEntityFinderSource _entityFinderSource;
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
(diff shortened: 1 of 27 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Internal/EntityMaterializerSource.cs
```diff
@@ -158,5 +161,24 @@ var blockExpressions
...
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual Func<ValueBuffer, object> GetMaterializer(IEntityType entityType)
+            => Materializers.GetOrAdd(entityType, e =>
+                {
+                    var valueBufferParameter = Expression.Parameter(typeof(ValueBuffer), "values");
+
+                    return Expression.Lambda<Func<ValueBuffer, object>>(
+                        CreateMaterializeExpression(e, valueBufferParameter),
```
bricelam: Is there anything that was calling `CreateMaterializeExpression` than can be updated to use this new cache?

bricelam:
:shipit:

src/Microsoft.EntityFrameworkCore/ChangeTracking/EntityEntry.cs
```diff
@@ -200,5 +204,69 @@ public virtual CollectionEntry Collection([NotNull] string navigationPropertyNam
         ///     otherwise true.
         /// </summary>
         public virtual bool IsKeySet => InternalEntry.IsKeySet;
+
+        /// <summary>
+        ///     Gets the current property values for the this entity.
```
AndriySvyryd: Extra 'the'

src/Microsoft.EntityFrameworkCore/ChangeTracking/PropertyValues.cs
```diff
@@ -0,0 +1,124 @@
...
+        ///         Sets the values of this object by reading values from another <see cref="PropertyValues" />
+        ///         object.
+        ///     </para>
+        ///     <para>
+        ///         The other object must be based on the same type as this object, or a type derived
+        ///         from the type for this object.
+        ///     </para>
+        /// </summary>
+        /// <param name="propertyValues"> The object from which values should be coiped. </param>
+        public abstract void SetValues([NotNull] PropertyValues propertyValues);
```
AndriySvyryd: Consider taking a `IReadOnlyDictionary<string, object>` here and implementing it by `PropertyValues` or having it as an overload

AndriySvyryd:
:shipit:

src/Microsoft.EntityFrameworkCore/Metadata/Internal/EntityMaterializerSource.cs
ajcvickers: Query uses it, but I believe it compiles it into its own delegates that are cached at a higher level. @anpete can probably comment about whether it would make sense to combine these at all.

src/Microsoft.EntityFrameworkCore/ChangeTracking/PropertyValues.cs
ajcvickers: Seems like a good idea. I'll create an issue for it. I wonder why we didn't have it in EF6?

PR #6350, opened 2016-08-17
Remove references to issue #6292

ajcvickers:
We decided that issue #6292 is just a limitation of SQLite and we are therefore not going to do anything in EF about it. Therefore, the SQLite version of the composite key store generated tests will continue to not use store generated values and comments in the code indicating that this should change have been removed.


bricelam:
:shipit:

PR #6362, opened 2016-08-18
[TINY] Migrations: Unwrap enum types before synthesizing a default value

### LINKED ISSUE #6193: Migrations fails to adds corresponding using when using enums
Add a migration when a model has an enum fails to include the using for the appropiate namespace on the migration file and results in compilation error.

```
                name: "EnumName",
                table: "TableWithEnum",
                nullable: false,
                defaultValue: EnumName.None);
```

adding the using by hand solves the issue.

bricelam:
Fixes #6193 

Note, you may want to review [ignoring whitespace changes](https://github.com/aspnet/EntityFramework/pull/6362/files?w=1).


natemcmaster:
:shipit:

PR #6364, opened 2016-08-18
Remove the AssemblyLoader abstraction from design-time operations

natemcmaster:
This abstraction doesn't have multiple implementations anymore, and it probably won't. Assembly.Load is sufficient. I believe this was originally introduced to handle AssemblyLoadContext, but it's not really necessary as the clr handles flowing Assembly.Load into the correct AssemblyLoadContext.LoadFromAssemblyName call.

This is a breaking change on our design packages from preview2.

cc @bricelam


bricelam:
:shipit: I always imagined it would be a good abstraction for tests too, but we can add it back if and when we write tests. :wink:

Issue #2295, opened 2015-05-29
Implement missing APIs from ChangeTacker/Entry


rowanmiller:
We do not have all APIs that we historically had available:
* GetModifiedProperties
* Reload
* etc.

NullVoxPopuli:
:+1: I think I need Reload, too

shobman:
+1 for Reload

ajcvickers:
GetModifiedProperties issue resolved with commit above. Reload is being tracked by #1203

Issue #2245, opened 2015-05-21
How to detect changed properties?


Nebelkraehe:
Hello.

At the moment I´m looking for a method/property delivering the changed properties inside an entity.
I detect the changes by `EntityState` but I can´t find any access to receive only the changed properties without iterating the whole properties for changes. 

In earlier versions this information was accessible by `ObjectStateEntry.GetModifiedProperties()`. Is there any equivalent way in EF7?

Thanks in advance.

Nebelkraehe

rowanmiller:
Here is the code you need to write at the moment.

```
            var entry = context.Entry(myEntity);
            var propertyNames = entry.Metadata.GetProperties().Select(p => p.Name);
            var modifiedProperties = propertyNames.Select(p => entry.Property(p)).Where(e => e.IsModified);
```

Leaving open as this code is worse than I was expecting and we may want to bump up priority of implementing the missing APIs here.

Nebelkraehe:
Thank you very much. 
A direct access would be great.

Nebelkraehe:
How about inserting this two methods into [EntityEntry](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Core/ChangeTracking/EntityEntry.cs)-Class?
```
public virtual IEnumerable<string> GetModifiedPropertyNames()
{
	return Metadata.GetProperties().Where(e => Property(e.Name).IsModified).Select(s => s.Name);
}

public virtual IEnumerable<IProperty> GetModifiedProperties()
{
	return Metadata.GetProperties().Where(e => Property(e.Name).IsModified);
}
```

rowanmiller:
Opened up a high level item to track this and other APIs that we want to implement on ChangeTracker/Entry https://github.com/aspnet/EntityFramework/issues/2295

PR #6370, opened 2016-08-19
Add API to get PropertyEntry objects for all properties of an entity

ajcvickers:
Likewise for members, navigations, references, collections.

Issues #2295 and #2245

The issue here is that it is quite common to want to do something for all property entries for a given entity. In the question asked by the customer this was to get the modified state. This was always possible but required two steps:
- Use metadata APIs to find all the Properties
- Loop/query over this metadata getting entries

So Rowan's code looked liked this:
```C#
var entry = context.Entry(myEntity);
var propertyNames = entry.Metadata.GetProperties().Select(p => p.Name);
var modifiedEntries = propertyNames.Select(p => entry.Property(p)).Where(e => e.IsModified);
```

With these new APIs the metadata and entry steps are combined for you so you would now write this:
```C#
var modifiedEntries = context.Entry(myEntity).Properties.Where(e => e.IsModified);
```

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/EntityEntry.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/EntityEntry.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore/ChangeTracking/EntityEntry.cs
```diff
@@ -167,6 +191,21 @@ public virtual PropertyEntry Property([NotNull] string propertyName)
...
+        ///     Provides access to change tracking information and operations for all
+        ///     properties of this entity.
+        /// </summary>
+        public virtual IEnumerable<PropertyEntry> Properties
+        {
+            get
+            {
+                foreach (var property in InternalEntry.EntityType.GetProperties())
+                {
+                    yield return new PropertyEntry(InternalEntry, property);
```
AndriySvyryd: Why not call `.Select` instead of yield?

AndriySvyryd:
:shipit:

src/Microsoft.EntityFrameworkCore/ChangeTracking/EntityEntry.cs
ajcvickers: Fixed

PR #6375, opened 2016-08-19
[small] polish error handling in ef.exe

### LINKED ISSUE #6371: CLI commands: database drop throws ObjectDisposedException
### Steps to reproduce
`dotnet ef database drop` on a project that uses Startup and .AddDbContext.

### The issue
The command discovers the database correctly, but when it attempts to actually drop the database, it throws:

```
> dotnet ef database drop
This command will permanently drop the database:
    Database name : main
    Data source   : ./test.db
Are you sure you want to proceed? (y/N) y
System.ObjectDisposedException: Cannot access a disposed object. A common cause of this error is disposing a context that was resolved from dependency injection and then later trying to
 use the same context instance elsewhere in your application. This may occur if you are calling Dispose() on the context, or wrapping the context in a using statement. If you are using
dependency injection, you should let the dependency injection container take care of disposing context instances.
Object name: 'AppDbContext'.
   at Microsoft.EntityFrameworkCore.DbContext.get_InternalServiceProvider()
   at Microsoft.EntityFrameworkCore.Infrastructure.AccessorExtensions.GetService[TService](IInfrastructure`1 accessor)
   at Microsoft.EntityFrameworkCore.Design.DbContextOperations.CreateContext(Func`1 factory)
   at Microsoft.EntityFrameworkCore.Design.DbContextOperations.CreateContext(String contextType)
   at Microsoft.EntityFrameworkCore.Design.DbContextOperations.DropDatabase(String contextType)
   at Microsoft.EntityFrameworkCore.Design.OperationExecutor.DropDatabase.<>c__DisplayClass0_1.<.ctor>b__0()
   at Microsoft.EntityFrameworkCore.Design.OperationExecutor.OperationBase.Execute(Action action)
```

### Further technical details

Microsoft.EntityFrameworkCore.Tools.DotNet 1.0.0-preview3-21911

natemcmaster:
A collection of minor tweaks to polish ef.exe.

Change color output of drop database command so it displays better against default blue PowerShell background

Fix #6371 - always create new instance of DbContext for each design-time operation

cc @bricelam

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Design/Design/DbContextOperations.cs
+++ b/src/Microsoft.EntityFrameworkCore.Design/Design/DbContextOperations.cs
@@ -118,7 +118,7 @@ namespace Microsoft.EntityFrameworkCore.Design
             {
                 contexts.Add(
                     context,
-                    FindContextFactory(context) ?? (() => (DbContext)_runtimeServices.GetRequiredService(context)));
+                    FindContextFactory(context) ?? (() => (DbContext)ActivatorUtilities.CreateInstance(_runtimeServices, context)));
             }
 
             // Look for DbContext classes in assemblies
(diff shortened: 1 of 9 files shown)


bricelam:
:shipit:

Issue #6316, opened 2016-08-12
Add support for data annotations on fields


AndriySvyryd:
If a field is configured for a property and there's a data annotation on the field it should be applied to the property. Currently only the `KeyAttribute` is applied

PR #6376, opened 2016-08-19
Add support for data annotations on fields

ajcvickers:
Issue #6316

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
@@ -196,41 +196,43 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
                 return null;
             }
 
-            var candidateProperties = new List<string>();
-            foreach (var propertyInfo in entityType.ClrType.GetRuntimeProperties().OrderBy(p => p.Name))
+            string candidateProperty = null;
+            var clrType = entityType.ClrType;
+            foreach (var memberInfo in clrType.GetRuntimeProperties().Cast<MemberInfo>()
+                .Concat(clrType.GetRuntimeFields()))
             {
-                var targetType = FindCandidateNavigationPropertyType(propertyInfo);
-                if (targetType != null)
+                var propertyInfo = memberInfo as PropertyInfo;
+                if (propertyInfo != null
+                    && FindCandidateNavigationPropertyType(propertyInfo) != null)
                 {
                     continue;
                 }
 
-                var attribute = propertyInfo.GetCustomAttribute<ForeignKeyAttribute>(true);
+                var attribute = memberInfo.GetCustomAttribute<ForeignKeyAttribute>(true);
                 if ((attribute != null)
                     && (attribute.Name == navigationName))
                 {
-                    candidateProperties.Add(propertyInfo.Name);
+                    if (candidateProperty != null)
+                    {
+                        throw new InvalidOperationException(CoreStrings.CompositeFkOnProperty(navigationName, entityType.DisplayName()));
+                    }
+                    candidateProperty = memberInfo.Name;
                 }
             }
 
-            if (candidateProperties.Count > 1)
-            {
-                throw new InvalidOperationException(CoreStrings.CompositeFkOnProperty(navigationName, entityType.DisplayName()));
-            }
-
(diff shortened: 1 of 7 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -197,19 +197,22 @@ private string FindForeignKeyAttributeOnProperty(EntityType entityType, string n
...
 
             var candidateProperties = new List<string>();
-            foreach (var propertyInfo in entityType.ClrType.GetRuntimeProperties().OrderBy(p => p.Name))
+            var clrType = entityType.ClrType;
+            foreach (var memberInfo in clrType.GetRuntimeProperties().Cast<MemberInfo>()
+                .Concat(clrType.GetRuntimeFields()).OrderBy(p => p.Name))
             {
-                var targetType = FindCandidateNavigationPropertyType(propertyInfo);
+                var propertyInfo = memberInfo as PropertyInfo;
+                var targetType = propertyInfo == null ? null : FindCandidateNavigationPropertyType(propertyInfo);
```
AndriySvyryd: Split into two loops to avoid the double cast

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -197,19 +197,22 @@ private string FindForeignKeyAttributeOnProperty(EntityType entityType, string n
             }
 
             var candidateProperties = new List<string>();
-            foreach (var propertyInfo in entityType.ClrType.GetRuntimeProperties().OrderBy(p => p.Name))
+            var clrType = entityType.ClrType;
+            foreach (var memberInfo in clrType.GetRuntimeProperties().Cast<MemberInfo>()
+                .Concat(clrType.GetRuntimeFields()).OrderBy(p => p.Name))
```
AndriySvyryd: Since we throw for multiple attributes the `OrderBy` is irrelevant

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
```diff
@@ -197,19 +197,22 @@ private string FindForeignKeyAttributeOnProperty(EntityType entityType, string n
...
                     continue;
                 }
 
-                var attribute = propertyInfo.GetCustomAttribute<ForeignKeyAttribute>(true);
+                var attribute = memberInfo.GetCustomAttribute<ForeignKeyAttribute>(true);
                 if ((attribute != null)
                     && (attribute.Name == navigationName))
                 {
-                    candidateProperties.Add(propertyInfo.Name);
+                    candidateProperties.Add(memberInfo.Name);
```
AndriySvyryd: Also there is no point collecting the candidates, we can just throw when another one is found

AndriySvyryd:
:shipit:

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
ajcvickers: Tried this, but there is more code and it seems less readable, so made a minor change to not store targetType into a local but left as a single loop.

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
ajcvickers: Fixed.

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
ajcvickers: Fixed.

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyAttributeConvention.cs
AndriySvyryd: ok

PR #6377, opened 2016-08-20
Project correct value types into array

ajcvickers:
Issue #6337

The expression tree has converts in it to box the returned values. But this means we always ask for object from the reader. Therefore, use the value inside the convert when asking for the value from the data reader.


bricelam:
:shipit:

Issue #6304, opened 2016-08-11
Expose way to get simple, human-readable, representation of the model for debugging


ajcvickers:
There are some internal APIs called ToDebugString that allow a model to be dumped out in a form like this:
```
Model 1.1.0-csproj
  Game
    Properties: 
      Id (<Id>k__BackingField, int) Required PK ReadOnlyAfterSave RequiresValueGenerator ValueGenerated.OnAdd 0 0 0 -1 0
    Navigations: 
      Items (<Items>k__BackingField, ICollection<Item>) Collection ToDependent Item Inverse: Game 1 -1 1 -1 -1
      Levels (<Levels>k__BackingField, ICollection<Level>) Collection ToDependent Level Inverse: Game 2 -1 2 -1 -1
    Keys: 
      Id PK
  Item
    Properties: 
      Id (<Id>k__BackingField, int) Required PK ReadOnlyAfterSave RequiresValueGenerator ValueGenerated.OnAdd 0 0 0 -1 0
      GameId (<GameId>k__BackingField, int) Required FK Index 1 1 1 -1 1
      LevelId (<LevelId>k__BackingField, int) Required FK Index 2 2 2 -1 -1
    Navigations: 
      Game (<Game>k__BackingField, Game) ToPrincipal Game Inverse: Items 3 -1 3 -1 -1
      Level (<Level>k__BackingField, Level) ToPrincipal Level Inverse: Items 4 -1 4 -1 -1
    Keys: 
      Id PK
    Foreign keys: 
      GameId -> Game.Id ToDependent: Items ToPrincipal: Game
      GameId, LevelId -> Level.GameId, Level.Id ToDependent: Items ToPrincipal: Level
  Level
    Properties: 
      GameId (<GameId>k__BackingField, int) Required PK FK Index ReadOnlyAfterSave 0 0 0 -1 0
      Id (<Id>k__BackingField, int) Required PK ReadOnlyAfterSave 1 1 1 -1 -1
    Navigations: 
      Game (<Game>k__BackingField, Game) ToPrincipal Game Inverse: Levels 2 -1 2 -1 -1
      Items (<Items>k__BackingField, ICollection<Item>) Collection ToDependent Item Inverse: Level 3 -1 3 -1 -1
    Keys: 
      GameId, Id PK
    Foreign keys: 
      GameId -> Game.Id ToDependent: Levels ToPrincipal: Game
```

[...]

PR #6388, opened 2016-08-23
Expose way to get simple, human-readable, representation of the model for debugging

ajcvickers:
Issue #6304

Lazy-evaluated detailed representation including annotations as discussed, plus single line representation for standard ToString.


AndriySvyryd:
:shipit:

Issue #6366, opened 2016-08-19
Second level expand not working correctly


vukasinspasic:
Hello. I noticed bugs in library.

### Steps to reproduce
When I try to make query on Posts DbSet with 2 expands of 2. level:

```C#
var query = dbContext.Posts.AsQueryable();
query = query
    .Include(x => x.User)
        .ThenInclude(x => x.Roles)
    .Include(x => x.VoteDefinition)
        .ThenInclude(x => x.PossibleAnswers)
```

Database model (some properties omitted for clarity):

```C#
    public class Post
    {
        public int Id { get; set; }
        public int UserId { get; set; }
        public User User { get; set; }
        public int? VoteDefinitionId { get; set; }
        public VoteDefinition VoteDefinition { get; set; }
    }
    public class User
    {
        public int Id { get; set; }
        public IList<Post> Posts { get; set; }
        public IList<Role> Roles { get; set; }
    }
    public class Role
    {
        public int Id { get; set; }
        public int UserId { get; set; }
    }
    public class VoteDefinition
    {
        public int Id { get; set; }
        [ForeignKey("Post")]
        public int PostId { get; set; }
        public Post Post { get; set; }
        public IList<VoteDefinitionPossibleAnswer> PossibleAnswers { get; set; }
    }
    public class VoteDefinitionPossibleAnswer
    {
        public int Id { get; set; }
        public int VoteDefinitionId { get; set; }
        public VoteDefinition VoteDefinition { get; set; }
    }
```

### The issue
This generated bad SQL query resulting in User.Roles not expanded. 

### Further technical details
As you can see in generated SQL queries, in 2. SQL query, there is line 
**LEFT JOIN "VoteDefinitions" AS "v" ON "v"."PostId" = "f"."Id"**
which should be only in 3. SQL query

SQL generated by this query:
```
info: Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommandBuilderFactory[1]
Executed DbCommand (55ms) [Parameters=[@__ToLower_0='?', @__p_2='?', @__p_1='?'], CommandType='Text', CommandTimeout='30']
```
<summary>
<details>
``` SQL 
[...]

PR #6389, opened 2016-08-23
[1.0.1] Fix to #6366 - 2 level expand not working correctly

maumar:
Problem happens for complex include scenarios like so:
```c#
ctx
    .Posts.Include(x => x.User).ThenInclude(x => x.Roles)
    .Include(x => x.VoteDefinition).ThenInclude(x => x.PossibleAnswers)
    .Skip(1)
```
We generate 3 queries for this case:

- first one gets Posts and 1:1 navigations (User and Vote Definition)
```SQL
SELECT [x].[Id], [x].[Name], [x].[UserId], [x].[VoteDefinitionId], [v].[Id], [v].[PostId], [u].[Id], [u].[Name]
FROM [Posts] AS [x]
LEFT JOIN [VoteDefinitions] AS [v] ON [v].[PostId] = [x].[Id]
INNER JOIN [Users] AS [u] ON [x].[UserId] = [u].[Id]
ORDER BY [v].[Id], [u].[Id]
OFFSET @__p_0 ROWS',N'@__p_0 int
```
- second one copies the first one and joins it with Roles table to return associated Roles
```SQL
SELECT [r].[Id], [r].[Name], [r].[UserId]
FROM [Roles] AS [r]
INNER JOIN (
    SELECT DISTINCT [t0].*
    FROM (
        SELECT [v].[Id], [u].[Id] AS [Id0]
        FROM [Posts] AS [x]
        LEFT JOIN [VoteDefinitions] AS [v] ON [v].[PostId] = [x].[Id]
        INNER JOIN [Users] AS [u] ON [x].[UserId] = [u].[Id]
        ORDER BY [v].[Id], [u].[Id]
        OFFSET @__p_0 ROWS
    ) AS [t0]
) AS [u0] ON [r].[UserId] = [u0].[Id]
ORDER BY [u0].[Id], [u0].[Id0]',N'@__p_0 int
```
- third one is similar to the second one.

Problem is that in second query we are joining based on a wrong column (VoteDefinition's Id rather than User's Id). This is because we lose track of the information about which column is associated with which property.
Due to Skip() call in the original LINQ query, we generate SELECT * around the relevant column information. Later, when looking for which column to join, we peek into the SELECT expression and don't see any information there so we assume that we should join based on a property with the name we need. This works fine if there are no duplicate columns with the same name projected out (and uniquefied) from other joined tables.

Fix is, in case of a SELECT * query, to look for the relevant column information in the TablesExpressions/SelectExpressions that the SELECT * is projecting from.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
@@ -611,7 +611,14 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
             if (selectExpression != null)
             {
-                return selectExpression.Projection.ToList();
+                if (selectExpression.IsProjectStar)
+                {
+                    return selectExpression.Tables.SelectMany(ExtractProjections);
+                }
+                else
+                {
+                    return selectExpression.Projection.ToList();
+                }
             }
 
             var joinExpression = tableExpression as JoinExpressionBase;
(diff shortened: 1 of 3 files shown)


maumar:
@smitpatel

smitpatel:
:shipit:

maumar:
merged to 1.0.1 and dev, closing

Issue #4858, opened 2016-03-19
Select after complex GroupJoin lead to unpredictable result


yyjdelete:
Database: `MSSQL`
EntityFramework: `EntityFramework.Core:7.0.0-rc1-final` and `Microsoft.EntityFrameworkCore:1.0.0-rc2-20261(aspnetcidev)`

```C#
ctx.DeviceMsgUps
     .GroupJoin(ctx.UserDeviceMaps.Where(udp => udp.UserId == 1025 && udp.Status == 0 && udp.DeviceId != null), dmu => dmu.DeviceId, udp => udp.DeviceId, (key, vals) => new { key, vals })
     .Where(pair => pair.vals.Any())
     .Select(pair=>pair.key)
     .Select(key=>new { key.Format })
     .Count()
```
Get 3 result(the same with ToList).

---
```C#
ctx.DeviceMsgUps
     .GroupJoin(ctx.UserDeviceMaps.Where(udp => udp.UserId == 1025 && udp.Status == 0 && udp.DeviceId != null), dmu => dmu.DeviceId, udp => udp.DeviceId, (key, vals) => new { key, vals })
     .Where(pair => pair.vals.Any())
     .Select(pair=>pair.key)
     //.Select(key=>new { key.Format })
     .Count()
```
Get 57 result(the same with ToList).

---
```C#
ctx.DeviceMsgUps
     .GroupJoin(ctx.UserDeviceMaps.Where(udp => udp.UserId == 1025 && udp.Status == 0 && udp.DeviceId != null), dmu => dmu.DeviceId, udp => udp.DeviceId, (key, vals) => new { key, vals })
     .Where(pair => pair.vals.Any())
     //.Select(pair=>pair.key)
     //.Select(key=>new { key.Format })
     .Count()
```
Get 1 result(the same with ToList).

---
Model
```C#

    [System.CodeDom.Compiler.GeneratedCodeAttribute("EF.Reverse.POCO.Generator", "2.18.1.0")]
    public class DeviceMsgUp
    {
        public int MessageId { get; set; } // Message_ID (Primary key)
        public int DeviceId { get; set; } // Device_ID
        public byte Format { get; set; } // Format
        public byte Action { get; set; } // Action
        public string Data { get; set; } // Data (length: 1000)
        public System.DateTime? ExpireTime { get; set; } // Expire_Time
        public System.DateTime Update { get; set; } // Update
        public byte ResultStatus { get; set; } // Result_Status
[...]

PR #6395, opened 2016-08-24
[1.0.1] Fix to #4858 - Select after complex GroupJoin lead to unpredictable result

maumar:
Problem exists for complex query with groupjoin, where the inner element is a subquery, then we apply collection operator on the inner collection, but project only the outer element like so:

context.LevelOne
  .GroupJoin(
    context.LevelTwo.Where(l2 => l2.Name != "L2 01"),
    l1 => l1.Id,
    l2 => l2.Level1_Optional_Id,
    (l1, l2s) => new { l1, l2s })
  .Where(r => r.l2s.Any())
  Select(r => r.l1);

What happens is that we look at which query sources we need to materialize (based on final projection), deduce that we don't need to materialize the elements of the inner collection, so it's being represented as value buffer. Problem is that currently value buffer can't represent a null-value element, so Any() always returns true (even for empty collection), which leads to incorrect results.

Fix is to force materialization of inner collection elements as entities. We already did that, but missed the case where the inner collection was a subquery.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryNavigationsSqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryNavigationsSqlServerTest.cs
@@ -1165,9 +1165,9 @@ ORDER BY [od1].[OrderID], [od1].[ProductID]",
             base.GroupJoin_with_complex_subquery_and_LOJ_does_not_get_flattened();
 
             Assert.Contains(
-                @"SELECT [t].[CustomerID]
+                @"SELECT [t].[CustomerID], [t].[Address], [t].[City], [t].[CompanyName], [t].[ContactName], [t].[ContactTitle], [t].[Country], [t].[Fax], [t].[Phone], [t].[PostalCode], [t].[Region]
 FROM (
-    SELECT [c20].*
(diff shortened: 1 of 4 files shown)


maumar:
@smitpatel @divega

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryNavigationsSqlServerTest.cs
```diff
@@ -1165,9 +1165,9 @@ public override void GroupJoin_with_complex_subquery_and_LOJ_does_not_get_flatte
             base.GroupJoin_with_complex_subquery_and_LOJ_does_not_get_flattened();
 
             Assert.Contains(
-                @"SELECT [t].[CustomerID]
```
smitpatel: Why did this change?

smitpatel:
:shipit:

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryNavigationsSqlServerTest.cs
maumar: Now we materialize the inners as entities, so we need all the columns in the projection

maumar:
merged to 1.0.1 () and dev (1a46a54dd8e1b33ea6df1483abeb3099eb512ea6)

PR #6396, opened 2016-08-24
Improve model building perf by recreating foreign keys less frequently

AndriySvyryd:
Fix uncovered issues

This should revert the performance degradation from introducing indexes on foreign keys using conventions


ajcvickers:
:shipit:

Issue #3915, opened 2015-11-28
Add DbSet<T>.Local (or similar)


joshmouch:
You used to be able to do the following to a DbSet<T>:

````
context.EntityDbSet.Local.FirstOrDefault(searchLocalQuery)
````

But it doesn't look like you can anymore (unless I'm missing an extension assembly).

Short of this, a backup is to do the following:

````
context.ChangeTracker.Entries<YourEntity>().Any(e => e.Entity.Id == id)
````

However, that only works if you have the context.  Is there a way to get the context if you just have a DbSet<T>?

**Comment by rowanmiller** [comment]
We thought we already had a backlog item to support .Local... but it doesn't look like we do, so turning this into an enhancement to track it.

In the meantime, you can write something like this. It gives you most of what local did, the main difference is that it won't refresh the list if you query for more data. It will sync changes to the list back into the change tracker though.

```c#
public static class Extensions
{
    public static ObservableCollection<TEntity> GetLocal<TEntity>(this DbSet<TEntity> set)
        where TEntity : class
    {
        var context = set.GetService<DbContext>();
        var data = context.ChangeTracker.Entries<TEntity>().Select(e => e.Entity);
        var collection = new ObservableCollection<TEntity>(data);

        collection.CollectionChanged += (s, e) =>
        {
            if (e.NewItems != null)
            {
                context.AddRange(e.NewItems.Cast<TEntity>());
            }

            if (e.OldItems != null)
            {
                context.RemoveRange(e.OldItems.Cast<TEntity>());
            }
        };

        return collection;
    }
}
```

PR #6406, opened 2016-08-24
Implementation of DbSet.Local and data binding support

ajcvickers:
Issue #3915

Much of the implementation is the same as for EF6 and most of the tests have been ported directly. Key differences are:
- Calling .Local does not automatically call DetectChanges.
- .Local now returns a custom observable collection type rather than a type derived from ObservableCollection.

Both of these changes were driven by reported perf issues associated with the EF6 .Local. For the first point, DetectChanges can always be slow, and almost always .Local will return the same data regardless of whether DetectChanges is called.

For the second point, creating a new List containing a copy of all the tracked entities could be both slow and use a lot of memory. Also, the use of a list would mean inserts and removes from the list, which could be common, was slow. Therefore, the new type provides a view over the already tracked data without duplicating it.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/ObservableCollectionListSource.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/ObservableCollectionListSource.cs
@@ -1,0 +1,77 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if NET451
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.ChangeTracking
+{
+    /// <summary>
+    ///     <para>
+    ///         Extends <see cref="ObservableCollection{T}" /> and adds an explicit implementation of <see cref="IListSource" />.
+    ///     </para>
+    ///     <para>
+    ///         The method <see cref="IListSource.GetList" /> is implemented to return an <see cref="IBindingList" />
+    ///         implementation that stays in sync with the ObservableCollection.
+    ///     </para>
+    ///     <para>
+    ///         This class can be used to implement navigation properties on entities for use in Windows Forms data binding.
+    ///         For WPF data binding use an ObservableCollection rather than an instance of this class.
+    ///     </para>
+    /// </summary>
+    /// <typeparam name="T"> </typeparam>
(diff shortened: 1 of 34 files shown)


src/Microsoft.EntityFrameworkCore/ChangeTracking/ObservableCollectionListSource.cs
```diff
@@ -0,0 +1,77 @@
...
+    ///         The method <see cref="IListSource.GetList" /> is implemented to return an <see cref="IBindingList" />
+    ///         implementation that stays in sync with the ObservableCollection.
+    ///     </para>
+    ///     <para>
+    ///         This class can be used to implement navigation properties on entities for use in Windows Forms data binding.
+    ///         For WPF data binding use an ObservableCollection rather than an instance of this class.
+    ///     </para>
+    /// </summary>
+    /// <typeparam name="T"> </typeparam>
+    public class ObservableCollectionListSource<T> : ObservableCollection<T>, IListSource
```
bricelam: Should this derive from `ObservableHashSet`? Also should the doc comment recommend that for WPF?

bricelam: Ah, we already have that. Why does this class exist?

src/Microsoft.EntityFrameworkCore/ChangeTracking/ObservableCollectionListSource.cs
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if NET451
```
divega: Apparently `IListSource` is in netstandard1.5 and up, defined in the `System.ComponentModel.TypeConverter` package. Should we add this dependency and compile the same code for more targets?

src/Microsoft.EntityFrameworkCore/ChangeTracking/ObservableCollectionListSource.cs
ajcvickers: It exists for WinForms binding when people are using ObservableCollections as their nav props.

src/Microsoft.EntityFrameworkCore/ChangeTracking/ObservableCollectionListSource.cs
divega: Ok, last time we had that conversation I believed that `IListSource` wasn't in netstandard at all. But yes, I am not sure it being in netstandard1.5 changes things enough.

PR #6407, opened 2016-08-24
Fix to #5613 - Query :: optional navigations null-ref protection logic doesn't work for some cases involving multiple optional navs and method calls

maumar:
Problem was that we were providing null protection in case of optional navigation access, but that did not include further calling methods on top of them.
If the navigation was null, the member access would return null and we would try to call a method on it, which results in NRE.

Fix is to introduce a new reducible expression that represents conditional access to a method or a member. In memory this gets rewritten to a Block expression that recursively evaluates it's caller, assigns it to a variable and depending on the value, either returns null or performs the required access operation. This is superior to using Conditional expression that we did before, because the caller only gets evaluated once, rather than N times.

Also fixes related issue #6261 - DateTime converted to Nullable<DateTime> inside optional navigation property, throws exception on AddDays.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -9,6 +9,7 @@ using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Query.Expressions;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 using Microsoft.EntityFrameworkCore.Query.ExpressionTranslators;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
 using Microsoft.EntityFrameworkCore.Storage;
@@ -981,6 +982,12 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 return newOperand != explicitCast.Operand
                     ? new ExplicitCastExpression(newOperand, explicitCast.Type)
                     : expression;
+            }
+
+            var nullConditional = expression as NullConditionalExpression;
+            if (nullConditional != null)
+            {
+                return Visit(nullConditional.AccessOperation);
             }
 
             return base.VisitExtension(expression);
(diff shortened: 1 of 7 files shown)


maumar:
@divega @smitpatel

maumar:
@anpete

src/Microsoft.EntityFrameworkCore/Query/Expressions/NullConditionalExpression.cs
```diff
@@ -0,0 +1,117 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+// Logic in this file is based on https://github.com/bartdesmet/ExpressionFutures/
+
+using System;
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors;
+
+namespace Microsoft.EntityFrameworkCore.Query.Expressions
```
anpete: Internal?

src/Microsoft.EntityFrameworkCore/Query/Expressions/NullConditionalExpression.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// </summary>
+        /// <param name="nullableCaller"> Expression representing potentially nullable caller that needs to be tested for it's nullability. </param>
+        /// <param name="caller"> Expression representig actual caller for the access operation. </param>
+        /// <param name="accessOperation"> Expression representing access operation. </param>
+        public NullConditionalExpression(
+            [NotNull] Expression nullableCaller,
+            [NotNull] Expression caller,
+            [NotNull] Expression accessOperation)
+        {
+            NullableCaller = nullableCaller;
```
anpete: If this type remains public then need Checks here.

test/Microsoft.EntityFrameworkCore.Relational.Tests/Query/SqlTranslatingExpressionVisitorTest.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.EntityFrameworkCore.Storage;
+using Moq;
+using Xunit;
+
+namespace Microsoft.EntityFrameworkCore.Relational.Tests.Query
+{
+    public class SqlTranslatingExpressionVisitorTest
+    {
+        [Fact]
+        public void Null_conditional_expression_gets_pruned_during_sql_translation()
```
anpete: This test is likely going to cause more pain that its worth.

anpete:
:shipit:

Issue #5554, opened 2016-05-27
Fluent API: Timestamp/RowVersion


bricelam:
We have a core-level convention for `[Timestamp]`, but no corresponding `IsRowVersion` method.

In EF6, these were equivalent.

```C#
class Data
{
    [Timestamp]
    public byte[] Version { get; set; }
}
```
```C#
// Won't compile on EFCore
modelBuilder.Entity<Data>().Property(d => d.Version).IsRowVersion();
```

PR #6408, opened 2016-08-24
Adds IsRowVersion fluent API

ajcvickers:
Issue #5554

This is pure sugar for ValueGeneratedOnAddOrUpdate and IsConcurrencyToken.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/F1SqlServerFixture.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/F1SqlServerFixture.cs
@@ -60,26 +60,12 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests
         {
             base.OnModelCreating(modelBuilder);
 
-            modelBuilder.Entity<Chassis>(b =>
-                {
-                    b.Property<byte[]>("Version")
-                        .ValueGeneratedOnAddOrUpdate()
-                        .IsConcurrencyToken();
-                });
+            modelBuilder.Entity<Chassis>().Property<byte[]>("Version").IsRowVersion();
+            modelBuilder.Entity<Driver>().Property<byte[]>("Version").IsRowVersion();
 
-            modelBuilder.Entity<Driver>(b =>
-                {
-                    b.Property<byte[]>("Version")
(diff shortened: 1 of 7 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/F1SqlServerFixture.cs
```diff
@@ -60,26 +60,12 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
-
-            modelBuilder.Entity<Team>(b =>
-                {
-                    b.Property<byte[]>("Version")
-                        .ValueGeneratedOnAddOrUpdate()
-                        .IsConcurrencyToken();
-                });
+            modelBuilder.Entity<Team>().Property<byte[]>("Version")
+                .ValueGeneratedOnAddOrUpdate()
+                .IsConcurrencyToken();
```
smitpatel: Is it intentional not to use `IsRowVersion` for this one?

smitpatel:
:shipit:

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/F1SqlServerFixture.cs
ajcvickers: Yes. Both ways should do the same thing, so this adds coverage for both ways of doing it.

roji:
Are you guys sure this belongs in Microsoft.EFCore rather than SqlServer? The concept of RowVersion really doesn't seems to be very SqlServer-specific...

ErikEJ:
I am assuming that this means: The column type is RowVersion - which is quite SQL Server specific, I believe

ajcvickers:
@roji @ErikEJ The API name is very SQL Server specific, but the behavior is just to do the same as the Timestamp attribute. We discussed this one round in circles many times. In the end we went for the same name and bahevior as EF6 and before, with providers free to implement in a way that makes sense, or to not support this combination of facets.

roji:
Yes sorry, I meant "seems to be very SqlServer-specific" above (edited).

I guess this seems reasonable for backwards compatibility reasons, as long as the database-specific nature of supported concurrency tokens is clearly documented.

PR #6409, opened 2016-08-25
RowNumberPaging works when joining the same entities twice

RickyLin:
_sorry, I have to resend this PR because my original PR can't be reopened as I force-updated my original branch_

RowNumberPaging works when joining the same entities twice.

Sometimes we need to join the same table multiple times. For instance, a Product table may have entered user and update user and both of them need to join User table.

Take Northwind database as an example, I didn't find out a proper relationship that can join the same table twice, so I use Order entity and join Customer entity twice:

```csharp 
AssertQuery<Customer, Order>((cs, os) => (
                from o in os
                join ca in cs on o.CustomerID equals ca.CustomerID
                join cb in cs on o.CustomerID equals cb.CustomerID
                orderby o.OrderID
                select new
                {
                    o.OrderID,
                    CustomerIDA = ca.CustomerID,
                    CustomerIDB = cb.CustomerID,
                    ContactNameA = ca.ContactName,
                    ContactNameB = cb.ContactName,
                }).Skip(10).Take(5));
```

This test works in InMemory database, SQL Server that supports offset keyword, but not SQL Server RowNumberPaging. My PR make it work too when using RowNumberPaging.

Basically, for the test above, RowNumberPaging generates the following incorrect SQL statement:

```sql
SELECT [t].[OrderID], [t].[CustomerID], [t].[ContactName]
FROM (
    SELECT [o].[OrderID], [ca].[CustomerID], [cb].[CustomerID] AS [c0], [ca].[ContactName], [cb].[ContactName] AS [c1], ROW_NUMBER() OVER(ORDER BY [o].[OrderID]) AS [__RowNumber__]
    FROM [Orders] AS [o]
    INNER JOIN [Customers] AS [ca] ON [o].[CustomerID] = [ca].[CustomerID]
    INNER JOIN [Customers] AS [cb] ON [o].[CustomerID] = [cb].[CustomerID]
) AS [t]
WHERE ([t].[__RowNumber__] > @__p_0) AND ([t].[__RowNumber__] <= (@__p_0 + @__p_1))
```

my PR will generate correct SQL statement:

```sql
SELECT [t].[OrderID], [t].[CustomerID], [t].[c0] AS [c0], [t].[ContactName], [t].[c1] AS [c1]
FROM (
    SELECT [o].[OrderID], [ca].[CustomerID], [cb].[CustomerID] AS [c0], [ca].[ContactName], [cb].[ContactName] AS [c1], ROW_NUMBER() OVER(ORDER BY [o].[OrderID]) AS [__RowNumber__]
    FROM [Orders] AS [o]
    INNER JOIN [Customers] AS [ca] ON [o].[CustomerID] = [ca].[CustomerID]
    INNER JOIN [Customers] AS [cb] ON [o].[CustomerID] = [cb].[CustomerID]
) AS [t]
WHERE ([t].[__RowNumber__] > @__p_0) AND ([t].[__RowNumber__] <= (@__p_0 + @__p_1))
```

We found out this bug during our production development, so if this bug is confirmed, I hope this fix could be included in 1.0.1. Otherwise, it would affect our development work a lot.

Thanks.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -533,7 +533,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 
                             return (ce != null
                                     && columnExpression != null
-                                    && (ce.Property == columnExpression.Property || ce.Name == columnExpression.Name)
+                                    && ce.Name == columnExpression.Name
                                     && ce.TableAlias == columnExpression.TableAlias)
                                 || ae?.Expression == expression;
                         });
@@ -599,7 +599,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
                             var ce = e.TryGetColumnExpression();
 
                             return ce != null
-                                && ce.Property == columnExpression.Property
                                 && ce.Name == columnExpression.Name
                                 && ce.TableAlias == columnExpression.TableAlias;
                         });
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
@@ -250,6 +250,24 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [ConditionalFact]
+        public virtual void Join_Customers_Orders_Orders_Skip_Take_Same_Properties()
+        {
+            AssertQuery<Customer, Order>((cs, os) => (
+                from o in os
+                join ca in cs on o.CustomerID equals ca.CustomerID
+                join cb in cs on o.CustomerID equals cb.CustomerID
+                orderby o.OrderID
+                select new
+                {
+                    o.OrderID,
+                    CustomerIDA = ca.CustomerID,
+                    CustomerIDB = cb.CustomerID,
+                    ContactNameA = ca.ContactName,
+                    ContactNameB = cb.ContactName,
+                }).Skip(10).Take(5));
+        }
+
+        [ConditionalFact]
         public virtual void Distinct_Skip_Take()
         {
             AssertQuery<Customer>(
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -1357,6 +1357,25 @@ OFFSET @__p_0 ROWS FETCH NEXT @__p_1 ROWS ONLY",
 
         [ConditionalFact]
         [SqlServerCondition(SqlServerCondition.SupportsOffset)]
+        public override void Join_Customers_Orders_Orders_Skip_Take_Same_Properties()
+        {
+            base.Join_Customers_Orders_Orders_Skip_Take_Same_Properties();
+
+            Assert.Equal(
+                @"@__p_0: 10
+@__p_1: 5
+
+SELECT [o].[OrderID], [ca].[CustomerID], [cb].[CustomerID], [ca].[ContactName], [cb].[ContactName]
+FROM [Orders] AS [o]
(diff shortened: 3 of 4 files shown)


divega:
@maumar @smitpatel can one of you please evaluate this PR?

divega:
**Triage:** the risk bar for the patch release is very high so the inclusion of this fix in 1.0.1 is unlikely. We still need to make sure the fix doesn't regress any existing scenarios.

divega:
@smitpatel could you please create a bug with the scenario in 1.1.0, assign it to yourself and investigate the fix?

smitpatel:
filed #6437

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -534,6 +534,7 @@ var projectionIndex
                             return (ce != null
                                     && columnExpression != null
                                     && (ce.Property == columnExpression.Property || ce.Name == columnExpression.Name)
+                                    && ce.Name == columnExpression.Name
```
smitpatel: The condition based on the `Name` was introduced while fixing bug #5909 
Essentially we compare name (doing OR with property) due to shared columns in TPH scenario where properties may be different but column is same in the table. Which is correct behavior. If same column is being projected from a table, it should not repeated.
Adding condition with `&&` regress the bug #5909 
In this particular case, issue is arising because the table is a push down subquery, hence `TableAlias` is same even though projections are coming from different tables through joins. Though this does not seem to be correct way to fix it

smitpatel:
This approach does not look right since it doesn't try to make it consistent with expected behavior and it regresses the bug #5909 
Closing this PR.
#6437 is tracking issue.

RickyLin:
But it passed all tests in test projects. In the "Drink", "Coke", "Likt" tests, it verified the SQL statement, so the `&&` condition doesn't lead to repeated columns.

smitpatel:
That would simplify condition `(ce.Property == columnExpression.Property || ce.Name == columnExpression.Name)` to always true when the added condition passes. Let me check consequences of that.

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
smitpatel: Can you remove line `(ce.Property == columnExpression.Property || ce.Name == columnExpression.Name)` altogether because with these new conditions, property match does not matter.
Also change similar thing in `AddToProjection(ColumnExpression)` function.

smitpatel:
Reactivating this PR.
The issue here is
When adding expression to projection, we don't want to add same column again. To compare if 2 `ColumnExpression` are same we used backing `Property` as a measure. Due to inheritance, 2 different Properties in metadata can be mapped to same column hence optional condition with name added while fixing #5909. In this case, both properties are the same, but they are coming from different query sources. While we push down subquery for row number paging, we lose the query source information, hence we end up assuming them to be same property.
This fix removes comparison based on `Property` instead match based on column name only. It is working correctly due to 2 reasons. 1. We are initializing column expression correctly based on column name of backing property. 2. We unique-ify the name properly while pushing down subquery.
Other fix would be include querysource for column expression so we don't lose the information but that fix would be complex.
@anpete @divega - Thoughts on taking this in?

anpete:
@smitpatel Sounds good. We can revisit the QuerySource binding improvement later if needed.

smitpatel:
@RickyLin - Can you update the PR with feedback. (also rebase on latest dev).

RickyLin:
@smitpatel - Sure. Thank you for the feedback.

RickyLin:
@smitpatel - I have updated the PR. Please take a look when you get a chance. 

Thanks.

divega:
@smitpatel do we have a tests that combines the inheritance scenario of #5909, multiple joins or  relationships with the same entity type and row number paging?

divega:
I agree with @anpete we can take this if we cannot figure out anything it would regress.

smitpatel:
Merged via 9f16618bedcc62e16508992ec4c82a4fd93a630f
@RickyLin - Thank you for contribution 😄

Issue #3921, opened 2015-11-29
String based Include


annexus:
In the older version of EF you were able to do something like, which is useful for dynamic scenarios.
```
        var query = db.CampaignCreatives.AsQueryable();
        foreach (string include in includes)
        {
            query = query.Include(include);
        };
```

---

Edited July-18 by @rowanmiller

## Workaround

Here is some code that adds two string based overloads of `Include`. 
* One allows you to do EF6 style string includes - such as `context.Blogs.Include("Posts")` and `context.Blogs.Include("Posts.Author.Photo")`. 
* The second overload allows you to make use of the C# 6 `nameof` feature when including multiple levels `context.Blogs.Include(nameof(Post.Blog), nameof(Blog.Owner), nameof(Person.Photo))`

``` c#
using Microsoft.EntityFrameworkCore;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Linq.Expressions;
using System.Reflection;

namespace ConsoleApplication5
{
    public static class Extensions
    {
        private static MethodInfo _include = typeof(EntityFrameworkQueryableExtensions)
                        .GetMethod("Include");

        private static MethodInfo _thenIncludeReference = typeof(EntityFrameworkQueryableExtensions)
            .GetMethods()
            .Where(m => m.Name == "ThenInclude")
            .Single(m => m.Name == "ThenInclude" &&
                         m.GetParameters()
                          .Single(p => p.Name == "source")
                          .ParameterType
                          .GetGenericArguments()[1].Name != typeof(ICollection<>).Name);

        private static MethodInfo _thenIncludeCollection = typeof(EntityFrameworkQueryableExtensions)
            .GetMethods()
            .Where(m => m.Name == "ThenInclude")
            .Single(m => m.Name == "ThenInclude" &&
                         m.GetParameters()
                          .Single(p => p.Name == "source")
                          .ParameterType
                          .GetGenericArguments()[1].Name == typeof(ICollection<>).Name);

        public static IQueryable<T> Include<T>(this IQueryable<T> query, string include)
        {
            return query.Include(include.Split('.'));
        }

        public static IQueryable<T> Include<T>(this IQueryable<T> query, params string[] include)
        {
            var currentType = query.ElementType;
            var previousNavWasCollection = false;

            for (int i = 0; i < include.Length; i++)
            {
                var navigationName = include[i];
                var navigationProperty = currentType.GetProperty(navigationName);
                if (navigationProperty == null)
                {
[...]

PR #6417, opened 2016-08-26
String based include

ajcvickers:
Issue #3921

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
@@ -366,6 +366,9 @@
   <data name="IncludeNonBindableExpression" xml:space="preserve">
     <value>The expression '{expression}' passed to the Include operator could not be bound.</value>
   </data>
+  <data name="IncludeBadNavigation" xml:space="preserve">
+    <value>The property '{property}' is not a navigation property of entity type '{entityType}'. The 'Include(string)' method can only be used with a '.' separated list of navigation property names.</value>
+  </data>
   <data name="LogCompilingQueryModel" xml:space="preserve">
     <value>Compiling query model: '{queryModel}'</value>
   </data>
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
@@ -91,7 +91,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
             }
 
             if (declaringType == typeof(Queryable)
-                || declaringType == typeof(EntityFrameworkQueryableExtensions))
+                || (declaringType == typeof(EntityFrameworkQueryableExtensions)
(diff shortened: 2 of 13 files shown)


src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -366,6 +366,9 @@
   <data name="IncludeNonBindableExpression" xml:space="preserve">
     <value>The expression '{expression}' passed to the Include operator could not be bound.</value>
   </data>
+  <data name="IncludeBadNavigation" xml:space="preserve">
+    <value>The property '{property}' is not a navigation property of entity type '{entityType}'. The 'Include(string)' method can only be used with a '.' separated list of navigation property names.</value>
```
anpete: Consider adding an example

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -91,11 +91,13 @@ protected override Expression VisitMethodCall(MethodCallExpression methodCallExp
             }
 
             if (declaringType == typeof(Queryable)
-                || declaringType == typeof(EntityFrameworkQueryableExtensions))
+                || (declaringType == typeof(EntityFrameworkQueryableExtensions)
+                    && (!methodInfo.IsGenericMethod
+                        || methodInfo.GetGenericMethodDefinition() != EntityFrameworkQueryableExtensions.StringIncludeMethodInfo)))
```
anpete: Can you elaborate on this?

anpete:
@ajcvickers Looks good. Curious, did you consider an approach that settled on a single representation for included nav. paths? E.g. perhaps something that didn't require the new branch in EntityQueryModelVisitor.IncludeNavigations, and/or the new ResultOperator node.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
ajcvickers: The include string should not be parameterized, so In used the same mechanism as FromSql uses to not parameterize it. However, that didn't work because all methods of EntityFrameworkQueryableExtensions were being skipped, presumably for perf reasons. This allows the Include method to be handled correctly without changing the code to process all methods of EntityFrameworkQueryableExtensions.

ajcvickers:
I considered it, but I think it would have to involve moving the expression parsing code to do it each time that an Include or ThenInclude call was made, which was more code changes and also seemed contrary to the pattern used for handling these kinds of things. It could also have been done by building an expression tree from the string, but that would just have been picked apart into the component strings anyway. I also considered a new ResultOperator, which would have been clean, but would have involved more code, some of it duplicated.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
anpete: Makes sense. That is indeed a perf. shortcut.

anpete:
:shipit:

PR #6430, opened 2016-08-29
Fix breaking API changes inadvertently introduced into the 1.0.1 branch

### LINKED ISSUE #6420: Breaking change in 1.0.1 - causing the SQL CE provider to break
### Steps to reproduce
Get the latest build of 1.0.1-rtm from myget feed and use in EFCore.SqlCe provider
Build, expecting the build to pass (no breaking changes)
### The issue
The build breaks, as the FindProperty method on MigrationsSqlBuilder is no longer available, but replaced with FindProperties

https://github.com/aspnet/EntityFramework/commit/82fdf18aeed4d061cb70535a4aab9310d5520ada

### Further technical details

EF Core version: 1.0.1-rtm-201608241155
Operating system: Windows 10
Visual Studio version: VS 2015 Update 3

bricelam:
Curse that trailing whitespace character! (You may want to view the changes [ignoring whitespace changes](https://github.com/aspnet/EntityFramework/pull/6430/files?w=1))

Resolves #6420

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
@@ -12,7 +12,7 @@ using Microsoft.EntityFrameworkCore.Metadata;
 namespace Microsoft.EntityFrameworkCore.Query.Internal
 {
     /// <summary>
-    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
     ///     directly from your code. This API may change or be removed in future releases.
     /// </summary>
     public class AsyncGroupJoinInclude : IDisposable
@@ -22,9 +22,10 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
         private readonly bool _querySourceRequiresTracking;
 
         private AsyncGroupJoinInclude _previous;
+        private AsyncGroupJoinIncludeContext _currentContext;
 
         /// <summary>
-        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public AsyncGroupJoinInclude(
@@ -52,12 +53,21 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                 _previous = previous;
             }
         }
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        [Obsolete(
+            "This method is obsolete and will be removed in the 1.1.0 release. Use CreateIncludeContext instead.",
+             error: true)]
+        public virtual void Initialize([NotNull] RelationalQueryContext queryContext)
+            => CreateIncludeContext(queryContext);
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual AsyncGroupJoinIncludeContext Initialize([NotNull] RelationalQueryContext queryContext)
+        public virtual AsyncGroupJoinIncludeContext CreateIncludeContext([NotNull] RelationalQueryContext queryContext)
         {
             var asyncGroupJoinIncludeContext
                 = new AsyncGroupJoinIncludeContext(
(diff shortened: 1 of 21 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
```diff
@@ -71,13 +72,21 @@ var asyncGroupJoinIncludeContext
...
 
-            return asyncGroupJoinIncludeContext;
+            return _currentContext = asyncGroupJoinIncludeContext;
         }
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        [Obsolete]
```
bricelam: @divega Should we specify `Message` and/or `IsError` for any of these?

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
```diff
@@ -22,9 +22,10 @@ public class AsyncGroupJoinInclude : IDisposable
         private readonly bool _querySourceRequiresTracking;
 
         private AsyncGroupJoinInclude _previous;
+        private AsyncGroupJoinIncludeContext _currentContext;
```
bricelam: @smitpatel Should these get disposed with the object?

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
divega: I am leaning towards having something in the message in the lines of "This internal API is deprecated as of version 1.0.1 and is only preserved for binary compatibility with version 1.0.0. It should not be used by any code."  Not sure why we didn't use the term internal in the XML comment to describe the API though.

AndriySvyryd:
I assume this will be only checked in to 1.0.1, correct?

divega:
@AndriySvyryd that is correct.

bricelam:
@AndriySvyryd When I "merge" into dev, I'll remove all the obsolete APIs. I'll probably keep the overloads...

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
smitpatel: Yes. `_currentContext.Dispose()` should be called in `Dispose()` method.

bricelam:
@divega I took some liberties with the obsolete messages. Let me know if you want me to use yours verbatim.

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
```diff
@@ -71,16 +72,26 @@ var asyncGroupJoinIncludeContext
...
-            return asyncGroupJoinIncludeContext;
+            return _currentContext = asyncGroupJoinIncludeContext;
         }
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        [Obsolete(
+            "This method is obsolete and will be removed in the 1.1.0 release. Use IncludeAsync on the object returned by Initialize instead.",
```
divega: Unfortunatelly I confirmed the return type change in Initalize() is also a binary breaking change 😢

divega: We can chat but I think using a new method name should work.

divega:
@bricelam the obsolete messages look good. I just wish we could emphasize somehow that the methods are not part of public surface we recommend customers to use in the first place.

divega:
But really using these should be so uncommon that it probably doesn't matter.

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
bricelam: What about an overload with an `out` parameter?

bricelam: The overload seems less messy than obsoleting `Initialize` and adding `Initialize2` just to rename `Initialize2` back to `Initialize` in 1.1

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/AsyncGroupJoinInclude.cs
```diff
@@ -68,19 +78,29 @@ var asyncGroupJoinIncludeContext
 
             if (_previous != null)
             {
-                asyncGroupJoinIncludeContext.SetPrevious(_previous.Initialize(queryContext));
+                asyncGroupJoinIncludeContext.SetPrevious(_previous.CreateIncludeContext(queryContext));
             }
 
-            return asyncGroupJoinIncludeContext;
+            return _currentContext = asyncGroupJoinIncludeContext;
```
divega: Should we assign this only in `Initialize()`? My reasoning is this may have unintended side effects, e.g. keep alive the object more than needed.

divega:
:shipit:

PR #5576, opened 2016-05-30
Fix #2852 - Support Enum.HasFlag in LINQ 

fujiy:
Fix #2852 - Support Enum.HasFlag in LINQ

PR #6431, opened 2016-08-30
Support Enum.HasFlag

### LINKED ISSUE #2852: Support Enum.HasFlag in LINQ
Hi,

I'd like to see [`Enum.HasFlag`](https://msdn.microsoft.com/en-us/library/system.enum.hasflag) supported in EF's LINQ provider.
I think this feature is pretty useful. Especially with the support of enums introduced to EF in the recent versions - a great feature indeed.

smitpatel:
Resolves #2852
Built on top of PR #5576


src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionTranslators/Internal/EnumHasFlagTranslator.cs
```diff
@@ -0,0 +1,73 @@
...
+            if (ReferenceEquals(methodCallExpression.Method, _methodInfo))
+            {
+                var argument = methodCallExpression.Arguments[0];
+                argument = argument.RemoveConvert();
+
+                // ReSharper disable once PossibleNullReferenceException
+                var objectEnumType = methodCallExpression.Object.Type.UnwrapNullableType();
+                var argumentEnumType = argument.Type.UnwrapNullableType();
+
+                if (argument.NodeType == ExpressionType.Constant)
```
maumar: no need to look at the ExpressionType since you are soft-casting argument to ConstantExpression and testing it after

smitpatel:
@fujiy - Thank you for contribution. The reason it was not able to run using `IMethodCallTranslator` because when `Enum.HasFlag` is invoked with constant value, the type of constantExpression is `System.Enum` which cannot be translated by type mapper. Such case does not arise when value is passed with parameter since the parameter has correct enum type in it. Since, values with `System.Enum` type are still valid to be translated to SQL, the `SqlTranslatingExpressionVisitor` now uses the type from the value if it find ConstantExpression with `System.Enum` type. Thus, methodCallTranslator works.
Also, removed exceptions from the translator. Translators don't throw exceptions, instead it returns null so that the expression is evaluated on client and clr runtime throws appropriate exception as per the case. (or even run it if it works).

fujiy:
Thanks for letting me now :+1:

maumar:
:shipit:

Issue #6303, opened 2016-08-11
Add string-based Property API that does not require a type


ajcvickers:
When creating a shadow property it is important to know the type of that property:
```C#
builder.Entity<Foo>().Property<int>("Id");
```
But if string API is actually being used to reference a CLR property, then it would be nice to be able to do this:
```C#
builder.Entity<Foo>().Property("Id");
```
This API would throw if the "Id" property is not found, but if it is found, then its type will be used.
This becomes especially nice when mapping fields, since this requires that the string API be used:
```C#
builder.Entity<Foo>().Property("_id");
```

PR #6436, opened 2016-08-30
Add string-based Property API that does not require a type

ajcvickers:
Issue #6303


bricelam:
Can you add a test to show that it throws for undefined properties?

ajcvickers:
@bricelam Updated

anpete:
:shipit:

bricelam:
double-:shipit:

Issue #5812, opened 2016-06-20
'must be reducible node' when aggregating over a join


kbekkenutte:
Using EF Core version: 1RC2

Given this model:
```
public class Fee { 
	public Guid Id { get; set; }
	public Foo Foo { get; set; }
	public ICollection<Foo> Foos { get; set; }
}
public class Foo {
	public Guid Id { get; set; }
	public ICollection<Bar> Bars { get; set; }
}
public class Bar {
	public Guid Id { get; set; }
}
```

A simple query like this works fine:
```context.Set<Foo>().Where(x => x.Bars.Any()).ToList();```

Similarly, this query will also work (albeit it using multiple queries):
```context.Fees.Select(x => x.Foo).Where(x => x.Bars.Any()).ToList();```

However this query fails with a 'must be reducible node' exception (see stacktrace below):
```context.Fees.SelectMany(x => x.Foos).Where(x => x.Bars.Any()).ToList();```

Exception message: ```must be reducible node```
Stack trace:
```
   at System.Linq.Expressions.Expression.ReduceAndCheck()
   at System.Linq.Expressions.Expression.ReduceExtensions()
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExtensionExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.Add(Expression node)
   at System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.AddArguments(IArgumentProvider expressions)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteMethodCallExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.Add(Expression node)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteBinaryExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpressionFreeTemps(Expression expression, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.Rewrite[T](Expression`1 lambda)
[...]

PR #6438, opened 2016-08-30
[SMALL] Fix to #5812 - 'must be reducible node' when aggregating over a join

maumar:
Problem was that when we do SelectMany, we should later re-map all subsequent references to the original query source, to point to the new query source (added during SelectMany nav rewrite). However we were not doing that for query sources inside a subquery. This caused incorrect query sources to be embedded in the query, which could not be translated and hence the 'must be reducible node' exception was thrown.

Fix is to also replace query sources inside a subquery.


maumar:
@anpete @smitpatel

anpete:
:shipit:

PR #6442, opened 2016-08-30
Add support for SQL Server memory-optimized tables

### LINKED ISSUE #3122: Migrations: Operations for annotation changes on model and entity type
EF7 allows adding provider-specific traits to various entities - SqlServer has `Clustered` indices for example.

Does the migrations model differ handle alterations in such provider-specific traits, e.g. if an index becomes `Clustered`? I have no idea if this kind of change is actually supported in SqlServer, but I definitely have some examples of PostgreSQL-specific traits that can be altered (e.g. a table's tablespace).

### LINKED ISSUE #6405: Memory-optimized table support for Sql Server
Provide fluent API:
``` C#
modelBuilder.Entity<Product>().ForSqlServerIsMemoryOptimized();
```

For on-premise SQL Server a file in a memory-optimized filegroup needs to be added to the database. We will dynamically create it in the migration if it doesn't exist.

The default durability is `SCHEMA_AND_DATA` we will _not_ provide an API to change it to `SCHEMA_ONLY`. Likewise we won't have API to set `MEMORY_OPTIMIZED_ELEVATE_TO_SNAPSHOT` to false. Both can be accomplished by editing the migration.

A convention will set indexes to be nonclustered on memory-optimized tables. But the computed columns will will be up to the user to remove.

See https://msdn.microsoft.com/en-us/library/dn133165.aspx for more info on memory-optimized tables. And https://msdn.microsoft.com/en-us/library/dn246937.aspx for detailed description of limitations of memory-optimized tables.

AndriySvyryd:
Add ForSqlServerIsMemoryOptimized() EntityBuilder extension method
Add support for conventions to run on annotation set
Change conventions to not set cascade delete on foreign keys on on memory-optimized tables
Add convention to configure keys and indexes as nonclustered on memory-optimized tables
Change IsClustered extension method on builders to take a bool?

Add AlterDatabaseOperation and AlterTableOperation
Add OldAnnotations to all alter operations
Add migrations annotations providers for IModel and ISequence
Diff migrations annotations for IModel, IEntityType and ISequence
Set migrations annotations on drop operations

Add support for memory-optimized tables to RelationalDatabaseCleaner

Fixes #6405
Fixes #3122


AndriySvyryd:
@bricelam Could you take a look? I will add more tests.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -264,10 +274,12 @@ public virtual IReadOnlyList<MigrationOperation> GetDifferences(IModel source, I
                 .Concat(dropTableOperations)
                 .Concat(dropOperations)
                 .Concat(dropColumnOperations)
+                .Concat(alterDatabaseOperations)
```
bricelam: My gut feeling is that this should go just before `createSequenceOperations`.

AndriySvyryd: I have no objections

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -472,6 +547,7 @@ protected virtual IEnumerable<MigrationOperation> Add([NotNull] IEntityType targ
                 Schema = sourceAnnotations.Schema,
                 Name = sourceAnnotations.TableName
             };
+            CopyAnnotations(MigrationsAnnotations.For(source), operation);
```
bricelam: Doing this always felt wrong to me. It seems like you should never need to know any additional information to drop it...

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -581,6 +658,7 @@ protected virtual IEnumerable<MigrationOperation> Diff([NotNull] IProperty sourc
                     IsDestructiveChange = isDestructiveChange
                 };
                 CopyAnnotations(targetMigrationsAnnotations, alterColumnOperation);
+                CopyAnnotations(sourceMigrationsAnnotations, alterColumnOperation.OldAnnotations);
```
bricelam: W00t! I need this.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -633,12 +711,15 @@ protected virtual IEnumerable<MigrationOperation> Remove([NotNull] IProperty sou
...
             var sourceEntityTypeAnnotations = Annotations.For(source.DeclaringEntityType.RootType());
 
-            yield return new DropColumnOperation
+            var operation = new DropColumnOperation
             {
                 Schema = sourceEntityTypeAnnotations.Schema,
                 Table = sourceEntityTypeAnnotations.TableName,
                 Name = Annotations.For(source).ColumnName
             };
+            CopyAnnotations(MigrationsAnnotations.For(source), operation);
```
bricelam: (Same as above)

bricelam: ...and other places.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
AndriySvyryd: Well you do need it for memory-optimized to suppress the transaction

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -979,12 +1069,16 @@ protected virtual IEnumerable<MigrationOperation> Diff([NotNull] ISequence sourc
                 };
             }
 
+            var sourceMigrationsAnnotations = MigrationsAnnotations.For(source).ToList();
+            var targetMigrationsAnnotations = MigrationsAnnotations.For(target).ToList();
```
bricelam: It's interesting that you can't actually annotate a sequence in metadata, but you can synthesize annotations...

test/Microsoft.EntityFrameworkCore.SqlServer.Tests/Migrations/SqlServerMigrationSqlGeneratorTest.cs
```diff
@@ -303,6 +303,16 @@ public virtual void CreateDatabaseOperation()
                 Sql);
         }
 
+        [Fact]
+        public virtual void AlterDatabaseOperationOperation()
+        {
+            Generate(new AlterDatabaseOperation { [SqlServerFullAnnotationNames.Instance.MemoryOptimized] = true });
```
AndriySvyryd: Obscure C# feature of the day: calling indexers from object initializers

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -186,6 +219,14 @@ public class MigrationsSqlGenerator : IMigrationsSqlGenerator
         }
 
         protected virtual void Generate(
+            [NotNull] AlterDatabaseOperation operation,
+            [CanBeNull] IModel model,
+            [NotNull] MigrationCommandListBuilder builder)
+        {
+            throw new NotImplementedException();
```
bricelam: Should we no-op?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
AndriySvyryd: We could enable sequence annotations in metadata when we find a compelling scenario that needs it

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/AlterSequenceOperation.cs
```diff
@@ -13,5 +14,6 @@ public class AlterSequenceOperation : MigrationOperation
         public virtual long? MinValue { get; [param: CanBeNull] set; }
         public virtual string Name { get; [param: NotNull] set; }
         public virtual string Schema { get; [param: CanBeNull] set; }
+        public virtual Annotatable OldAnnotations { get; [param: CanBeNull] set; } = new Annotatable();
```
bricelam: I also need Old* of everything else.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
AndriySvyryd: Yes

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/AlterSequenceOperation.cs
AndriySvyryd: Will extract to base type

test/Microsoft.EntityFrameworkCore.SqlServer.Tests/Migrations/SqlServerMigrationSqlGeneratorTest.cs
bricelam: lol, one of my favorites! It inspired me to write my [C# Object Notation](http://www.bricelam.net/2015/02/01/csharp-object-notation.html) post.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/IMigrationsAnnotationProvider.cs
```diff
@@ -10,10 +10,20 @@ namespace Microsoft.EntityFrameworkCore.Migrations
...
     {
+        IEnumerable<IAnnotation> For([NotNull] IModel model);
         IEnumerable<IAnnotation> For([NotNull] IIndex index);
         IEnumerable<IAnnotation> For([NotNull] IProperty property);
         IEnumerable<IAnnotation> For([NotNull] IKey key);
         IEnumerable<IAnnotation> For([NotNull] IForeignKey foreignKey);
         IEnumerable<IAnnotation> For([NotNull] IEntityType entityType);
+        IEnumerable<IAnnotation> For([NotNull] ISequence sequence);
+
+        IEnumerable<IAnnotation> ForRemove([NotNull] IModel model);
```
bricelam: Huh? The differ never generates DropDatabase operations...

bricelam: Oh interesting. It's a special AlterDatabase when you're going to InitialDatabase

bricelam: Makes sense.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -425,6 +481,33 @@ protected virtual IEnumerable<MigrationOperation> Add([NotNull] string target)
...
+                var targetAnnotations = Annotations.For(target);
+                var alterTableOperation = new AlterTableOperation
+                {
+                    Name = targetAnnotations.TableName,
+                    Schema = targetAnnotations.Schema
+                };
+                CopyAnnotations(targetMigrationsAnnotations, alterTableOperation);
+
+                var sourceAnnotations = Annotations.For(source);
+                alterTableOperation.OldTable = new AlterTableOperation
```
bricelam: My preference is just to have `OldAnnotations`

bricelam: (Or get rid of RenameTableOperation, but we already decided not to do that in a design meeting)

bricelam: To put it differently, my preference is to not have Schema & Name on OldTable (which just leaves the annotations).

AlterColumn.OldColumn would have type, is null, etc. but not schema, table, and name.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -582,6 +668,25 @@ protected virtual IEnumerable<MigrationOperation> Diff([NotNull] IProperty sourc
...
+                    Schema = sourceEntityTypeAnnotations.Schema,
+                    Table = sourceEntityTypeAnnotations.TableName,
+                    Name = sourceAnnotations.ColumnName,
+                    ClrType = source.ClrType.UnwrapNullableType().UnwrapEnumType(),
+                    ColumnType = sourceAnnotations.ColumnType,
+                    MaxLength = source.GetMaxLength(),
+                    IsUnicode = source.IsUnicode(),
+                    IsRowVersion = source.ClrType == typeof(byte[])
+                                   && source.IsConcurrencyToken
+                                   && source.ValueGenerated == ValueGenerated.OnAddOrUpdate,
```
bricelam: Extract method.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -582,6 +668,25 @@ protected virtual IEnumerable<MigrationOperation> Diff([NotNull] IProperty sourc
                 };
                 CopyAnnotations(targetMigrationsAnnotations, alterColumnOperation);
 
+                alterColumnOperation.OldColumn = new ColumnOperation
+                {
+                    Schema = sourceEntityTypeAnnotations.Schema,
+                    Table = sourceEntityTypeAnnotations.TableName,
+                    Name = sourceAnnotations.ColumnName,
+                    ClrType = source.ClrType.UnwrapNullableType().UnwrapEnumType(),
```
bricelam: I think `UnwrapNullableType` is superfluous now. (It wasn't when this was originally written)

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
bricelam: Ditto for AlterSequenceOperation

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -120,13 +129,23 @@ public class MigrationsSqlGenerator : IMigrationsSqlGenerator
 
             builder.AppendLine(SqlGenerationHelper.StatementTerminator);
 
-            EndStatement(builder);
+            if (terminate)
+            {
+                EndStatement(builder);
+            }
```
bricelam: (I can't remember if I said this already so sorry if I have.)

We have a similar pattern that is used to not add StatementTerminator and allow appending to the end of a statement. Instead of passing terminate, I think we should pass what we really need here: suppressTransaction.

This would also play better with the batching when we have it. (See [my prototype](https://github.com/aspnet/EntityFramework/compare/dev...bricelam:batch))

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -336,21 +432,103 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
+        {
+            Check.NotNull(operation, nameof(operation));
+            Check.NotNull(builder, nameof(builder));
+
+            if (!IsMemoryOptimized(operation))
+            {
+                return;
+            }
+
+            builder.AppendLine(@"
```
bricelam: I am 100% against using verbatim strings for newlines in product code. Please put each line in an `AppendLine` call.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -336,21 +432,103 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
-                .Append(SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
-                .AppendLine(SqlGenerationHelper.StatementTerminator);
+            var memoryOptimized = IsMemoryOptimized(operation);
+            if (memoryOptimized)
+            {
+                builder
+                    .Append("ALTER TABLE  ")
+                    .Append(SqlGenerationHelper.DelimitIdentifier(operation.Table))
+                    .Append(" DROP INDEX ")
+                    .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name));
```
bricelam: This syntax is so much better! It's a shame we can't just always use it.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -336,21 +432,103 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
+		END
+	END");
+
+            EndStatement(builder, suppressTransaction: true);
+        }
+
+        protected override void Generate(AlterTableOperation operation, IModel model, MigrationCommandListBuilder builder)
+        {
+            if (IsMemoryOptimized(operation)
+                || IsMemoryOptimized(operation.OldTable))
```
bricelam: XOR?

bricelam:
:+1: Looking good

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
AndriySvyryd: Agree, but I would still call them OldTable, etc so we could put more things on them without renaming

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
AndriySvyryd: I'm following the existing pattern. I think this is more flexible than just passing suppressTransaction as it also allows appending to the end.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
AndriySvyryd: I was wondering if anyone would notice 👀

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
AndriySvyryd: good catch

src/Microsoft.EntityFrameworkCore.SqlServer/Metadata/SqlServerEntityTypeAnnotations.cs
```diff
@@ -0,0 +1,38 @@
...
+        public SqlServerEntityTypeAnnotations([NotNull] RelationalAnnotations annotations)
+            : base(annotations, SqlServerFullAnnotationNames.Instance)
+        {
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual bool IsMemoryOptimized
```
smitpatel: Given warning above, `.Internal` namespace?

AndriySvyryd: I'll remove the doc comment

AndriySvyryd:
Updated with more tests

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
bricelam: We should be consistent about whether the statement terminator is appended or not then

AndriySvyryd: Makes sense. Should it be renamed for methods that don't append the statement terminator?

AndriySvyryd:
Updated migrations scaffolding

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
bricelam: :+1: That seems to be where my thinking lead me too.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
bricelam: I think we should just make them all not append it. It seems more generally useful. Seems trivial for the caller to add it if they just wanted to control the call to `EndStatement`.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationBuilder.cs
```diff
@@ -189,7 +189,16 @@ public MigrationBuilder([CanBeNull] string activeProvider)
...
+            [CanBeNull] string computedColumnSql = null,
+            [CanBeNull] Type oldClrType = null,
+            [CanBeNull] string oldType = null,
+            [CanBeNull] bool? oldUnicode = null,
+            [CanBeNull] int? oldMaxLength = null,
+            bool oldRowVersion = false,
+            bool oldNullable = false,
+            [CanBeNull] object oldDefaultValue = null,
+            [CanBeNull] string oldDefaultValueSql = null,
+            [CanBeNull] string oldComputedColumnSql = null)
```
bricelam: I wonder if these should be grouped in the API. For example...

```c#
.AlterColumn(
    oldColumn: new ColumnModel<int>(
        isRequired: false));
```
...or...
```c#
.AlterColumn()
    .OldColumn<int>(
        isRequired: flase);
```

Might be worth discussing in a design meeting, but I also like the simplicity and consistency with Rename* of the way you have it.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationBuilder.cs
```diff
@@ -207,20 +216,45 @@ public MigrationBuilder([CanBeNull] string activeProvider)
                 IsNullable = nullable,
                 DefaultValue = defaultValue,
                 DefaultValueSql = defaultValueSql,
-                ComputedColumnSql = computedColumnSql
+                ComputedColumnSql = computedColumnSql,
+                OldColumn = new ColumnOperation
```
bricelam: Is there a way to make this `null` for existing migrations? It might be time to start using the version metadata...

bricelam: We might be able to accomplish it with an overload where `oldNullable` is required.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/IAlterMigrationOperation.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace Microsoft.EntityFrameworkCore.Migrations.Operations
+{
+    public interface IAlterMigrationOperation
+    {
+        IMutableAnnotatable OldAnnotations { get; }
+    }
```
bricelam: YAGNI? Or just good OOP? Hard to say...

AndriySvyryd: We need this for AlterOperationBuilder

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
AndriySvyryd: k

bricelam:
:shipit:

PR #6445, opened 2016-08-31
Add DbContext.Database.GetMigrations() etc. methods

### LINKED ISSUE #6110: Migrations: Top level APIs for All/Applied/Pending
These APIs on `DbContext.Database` were useful in EF6. One example of where these would be useful is manually applying each migration so that you can give status to the user (#6038).

bricelam:
Resolves #6110

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
@@ -2,8 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
@@ -24,8 +26,8 @@ namespace Microsoft.EntityFrameworkCore
         ///         if it does not already exist.
         ///     </para>
         ///      <para>
-        ///         Note that this API is mutually exclusive with DbContext.Database.EnsureCreated(). EnsureCreated does not use migrations 
-        ///         to create the database and therefore the database that is created cannot be later updated using migrations. 
+        ///         Note that this API is mutually exclusive with DbContext.Database.EnsureCreated(). EnsureCreated does not use migrations
+        ///         to create the database and therefore the database that is created cannot be later updated using migrations.
         ///     </para>
         /// </summary>
         /// <param name="databaseFacade"> The <see cref="DatabaseFacade" /> for the context. </param>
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -24,22 +26,71 @@ public static class RelationalDatabaseFacadeExtensions
...
         /// <param name="databaseFacade"> The <see cref="DatabaseFacade" /> for the context. </param>
         public static void Migrate([NotNull] this DatabaseFacade databaseFacade)
             => Check.NotNull(databaseFacade, nameof(databaseFacade)).GetService<IMigrator>().Migrate();
 
         /// <summary>
+        ///     Gets all the migrations that are defined in the configured migrations assembly.
+        /// </summary>
+        /// <param name="databaseFacade"> The <see cref="DatabaseFacade" /> for the context.</param>
+        /// <returns>The list of migrations.</returns>
+        public static IEnumerable<string> GetMigrations([NotNull] this DatabaseFacade databaseFacade)
```
anpete: Are migration ids still the best thing to return from these APIs?

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -24,22 +26,71 @@ public static class RelationalDatabaseFacadeExtensions
...
+            => Check.NotNull(databaseFacade, nameof(databaseFacade)).GetService<IMigrationsAssembly>().Migrations.Keys;
+
+        /// <summary>
+        ///     Gets all migrations that have been applied to the target database.
+        /// </summary>
+        /// <param name="databaseFacade"> The <see cref="DatabaseFacade" /> for the context.</param>
+        /// <returns> The list of migrations. </returns>
+        public static IEnumerable<string> GetAppliedMigrations([NotNull] this DatabaseFacade databaseFacade)
+            => Check.NotNull(databaseFacade, nameof(databaseFacade)).GetService<IHistoryRepository>()
+                .GetAppliedMigrations().Select(hr => hr.MigrationId).ToArray();
```
anpete: Why are these methods buffering via ToArray?

bricelam: Mostly because the underlying one does, so there's little value in streaming just the projection.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
bricelam: Yes, at a high level, that's the only thing you can really use (to pass one it into `Migrate` or one of the commands). For more advanced scenarios, you can use `IMigrationsAssembly` to get the types and instances of the migration classes, but I have no idea what you'd be trying to do with them--analysis?

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
anpete: But this will create a copy.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
anpete: Was thinking tooling etc. but it seems fine to just use IMigrationsAssembly, too,

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
bricelam: So which is the lesser of two evils--returning an enumerator with a reference to an array of objects or copying the strings to a new array? :smile:

anpete: Copying :smile: With ToArray you are both allocating an array and copying the contents. If you are going to buffer here, then at least change the return type so that the caller gets the benefit. I.e. IReadOnlyList

anpete:
:shipit:

PR #6446, opened 2016-08-31
Add support for AttachDBFilename

### LINKED ISSUE #2810: Error when specifying AttachDbFileName in connection string
_Edit: look at https://github.com/aspnet/EntityFramework/issues/2810#issuecomment-243402108 for details on the bug._

When specifying AttachDbFileName in connection string, the following error appear: Unable to Attach database file as database xxxxxxx. 
Please advise.

Thanks

axelheer:
There isn't any handling for custom filenames within sql connection strings ('AttachDBFilename'), which leads to obscure error messages. Thus, we just add proper handling of filenames.

* AttachDBFilename gets a reset while building master connection strings
* Exists checks accept error 1832 **and 5120** too as a hint for a non existing database
* SqlServerCreateDatabaseOperation gets an additional FileName property

*NOTE:* I don't get the difference between "CreationTests" and "CreatorTests", so I just added my tests to the former ones (the other ones seem to test less related stuff).

~~*NOTE:* The changes broke a unit test, which I just changed accordingly. I'm not sure, if that's ok.~~

Fixes #2810

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/Operations/SqlServerCreateDatabaseOperation.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/Operations/SqlServerCreateDatabaseOperation.cs
@@ -8,5 +8,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Operations
     public class SqlServerCreateDatabaseOperation : MigrationOperation
     {
         public virtual string Name { get; [param: NotNull] set; }
+
+        public virtual string FileName { get; [param: CanBeNull] set; }
     }
 }
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Linq;
 using System.Text;
 using JetBrains.Annotations;
@@ -492,7 +493,35 @@ namespace Microsoft.EntityFrameworkCore.Migrations
 
             builder
                 .Append("CREATE DATABASE ")
-                .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name))
+                .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name));
+
+            if (!string.IsNullOrEmpty(operation.FileName))
+            {
+                var fileName = ExpandFileName(operation.FileName);
+                var name = Path.GetFileNameWithoutExtension(fileName);
+
+                var logFileName = Path.ChangeExtension(fileName, ".ldf");
+                var logName = name + "_log";
+
+                // Match default naming behavior of SQL Server
+                logFileName = logFileName.Insert(logFileName.Length - ".ldf".Length, "_log");
+
+                builder
+                    .AppendLine()
+                    .Append("ON (NAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(name))
+                    .Append("', FILENAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(fileName))
+                    .Append("')")
+                    .AppendLine()
+                    .Append("LOG ON (NAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(logName))
+                    .Append("', FILENAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(logFileName))
+                    .Append("')");
+            }
+
+            builder
                 .AppendLine(SqlGenerationHelper.StatementTerminator)
                 .EndCommand(suppressTransaction: true)
                 .Append("IF SERVERPROPERTY('EngineEdition') <> 5 EXEC(N'ALTER DATABASE ")
@@ -502,6 +531,25 @@ namespace Microsoft.EntityFrameworkCore.Migrations
                 .Append("')")
                 .AppendLine(SqlGenerationHelper.StatementTerminator)
                 .EndCommand(suppressTransaction: true);
+        }
+
+        private static string ExpandFileName(string fileName)
+        {
+            Check.NotNull(fileName, nameof(fileName));
+
+#if NET451
+
+            if (fileName.StartsWith("|DataDirectory|", StringComparison.OrdinalIgnoreCase))
+            {
+                var dataDirectory = AppDomain.CurrentDomain.GetData("DataDirectory") as string;
+                if (string.IsNullOrEmpty(dataDirectory))
+                    dataDirectory = AppDomain.CurrentDomain.BaseDirectory;
+                fileName = Path.Combine(dataDirectory, fileName.Substring("|DataDirectory|".Length));
+            }
+
+#endif
+
+            return Path.GetFullPath(fileName);
         }
 
         protected virtual void Generate(
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerConnection.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerConnection.cs
@@ -50,10 +50,13 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public virtual ISqlServerConnection CreateMasterConnection()
-            => new SqlServerConnection(new DbContextOptionsBuilder()
-                .UseSqlServer(
-                    new SqlConnectionStringBuilder { ConnectionString = ConnectionString, InitialCatalog = "master" }.ConnectionString,
-                    b => b.CommandTimeout(CommandTimeout ?? DefaultMasterConnectionCommandTimeout)).Options, Logger);
+        {
+            var builder = new SqlConnectionStringBuilder(ConnectionString) { InitialCatalog = "master" };
+            builder.Remove("AttachDBFilename");
+
+            return new SqlServerConnection(new DbContextOptionsBuilder()
+                .UseSqlServer(builder.ConnectionString, b => b.CommandTimeout(CommandTimeout ?? DefaultMasterConnectionCommandTimeout)).Options, Logger);
+        }
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
@@ -97,7 +97,10 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
                 .Build("IF EXISTS (SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_TYPE = 'BASE TABLE') SELECT 1 ELSE SELECT 0");
 
         private IReadOnlyList<MigrationCommand> CreateCreateOperations()
-            => _migrationsSqlGenerator.Generate(new[] { new SqlServerCreateDatabaseOperation { Name = _connection.DbConnection.Database } });
+        {
+            var builder = new SqlConnectionStringBuilder(_connection.DbConnection.ConnectionString);
+            return _migrationsSqlGenerator.Generate(new[] { new SqlServerCreateDatabaseOperation { Name = builder.InitialCatalog, FileName = builder.AttachDBFilename } }); 
+        }
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
@@ -170,7 +173,9 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
         }
 
         // Login failed is thrown when database does not exist (See Issue #776)
-        private static bool IsDoesNotExist(SqlException exception) => exception.Number == 4060;
+        // Unable to attach database file is thrown when file does not exist (See Issue #2810)
+        // Unable to open the physical file is thrown when file does not exist (See Issue #2810)
+        private static bool IsDoesNotExist(SqlException exception) => exception.Number == 4060 || exception.Number == 1832 || exception.Number == 5120;
 
         // See Issue #985
         private bool RetryOnExistsFailure(SqlException exception, ref int retryCount)
@@ -189,7 +194,11 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
             // And (Number 4060):
             //   System.Data.SqlClient.SqlException: Cannot open database "X" requested by the login. The
             //   login failed.
-            if ((exception.Number == 233 || exception.Number == -2 || exception.Number == 4060)
+            // And (Number 1832)
+            //   System.Data.SqlClient.SqlException: Unable to Attach database file as database xxxxxxx.
+            // And (Number 5120)
+            //   System.Data.SqlClient.SqlException: Unable to open the physical file xxxxxxx.
+            if ((exception.Number == 233 || exception.Number == -2 || exception.Number == 4060 || exception.Number == 1832 || exception.Number == 5120)
                 && ++retryCount < 30)
             {
                 ClearPool();
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
@@ -34,24 +34,40 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.Utilities
         ///     A non-transactional, transient, isolated test database. Use this in the case
         ///     where transactions are not appropriate.
         /// </summary>
-        public static Task<SqlServerTestStore> CreateScratchAsync(bool createDatabase = true)
-            => new SqlServerTestStore(GetScratchDbName()).CreateTransientAsync(createDatabase);
+        public static Task<SqlServerTestStore> CreateScratchAsync(bool createDatabase = true, bool useFileName = false)
+            => new SqlServerTestStore(GetScratchDbName(), useFileName).CreateTransientAsync(createDatabase);
 
-        public static SqlServerTestStore CreateScratch(bool createDatabase = true)
-            => new SqlServerTestStore(GetScratchDbName()).CreateTransient(createDatabase);
+        public static SqlServerTestStore CreateScratch(bool createDatabase = true, bool useFileName = false)
+            => new SqlServerTestStore(GetScratchDbName(), useFileName).CreateTransient(createDatabase);
 
         private SqlConnection _connection;
         private SqlTransaction _transaction;
         private readonly string _name;
+        private readonly string _fileName;
         private string _connectionString;
         private bool _deleteDatabase;
 
         public override string ConnectionString => _connectionString;
 
         // Use async static factory method
-        private SqlServerTestStore(string name)
+        private SqlServerTestStore(string name, bool useFileName = false)
         {
             _name = name;
+
+            if (useFileName)
+            {
+                #if NET451
+
+                var baseDirectory = AppDomain.CurrentDomain.BaseDirectory;
+
+                #else
+
+                var baseDirectory = AppContext.BaseDirectory;
+
+                #endif
+
+                _fileName = Path.Combine(baseDirectory, name + ".mdf");
+            }
         }
 
         private static string GetScratchDbName()
@@ -71,7 +87,7 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.Utilities
         {
             CreateShared(typeof(SqlServerTestStore).Name + _name, initializeDatabase);
 
-            _connectionString = CreateConnectionString(_name);
+            _connectionString = CreateConnectionString(_name, _fileName);
             _connection = new SqlConnection(_connectionString);
 
             if (useTransaction)
@@ -185,7 +201,7 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.Utilities
                 catch (SqlException e)
                 {
                     if (++retryCount >= 30
-                        || (e.Number != 233 && e.Number != -2 && e.Number != 4060))
+                        || (e.Number != 233 && e.Number != -2 && e.Number != 4060 && e.Number != 1832 && e.Number != 5120))
                     {
                         throw;
                     }
@@ -213,7 +229,7 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.Utilities
                 catch (SqlException e)
                 {
                     if (++retryCount >= 30
-                        || (e.Number != 233 && e.Number != -2 && e.Number != 4060))
+                        || (e.Number != 233 && e.Number != -2 && e.Number != 4060 && e.Number != 1832 && e.Number != 5120))
                     {
                         throw;
                     }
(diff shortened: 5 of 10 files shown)


axelheer:
cc @bricelam

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -302,7 +303,25 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
-                .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name))
+                .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name));
+
+            if (!string.IsNullOrEmpty(operation.FileName))
+            {
+                var logFileName = Path.ChangeExtension(operation.FileName, ".ldf");
+
+                builder
+                    .Append(" ON (NAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(operation.Name))
```
bricelam: I believe the filename and database name can be different. Not sure if it matters here.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -302,7 +303,25 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
+                    .Append("', FILENAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(operation.FileName))
```
bricelam: Also need to resolve |DataDirectory| on Full .NET

src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerConnection.cs
```diff
@@ -52,7 +52,7 @@ public class SqlServerConnection : RelationalConnection, ISqlServerConnection
         public virtual ISqlServerConnection CreateMasterConnection()
             => new SqlServerConnection(new DbContextOptionsBuilder()
                 .UseSqlServer(
-                    new SqlConnectionStringBuilder { ConnectionString = ConnectionString, InitialCatalog = "master" }.ConnectionString,
+                    new SqlConnectionStringBuilder { ConnectionString = ConnectionString, AttachDBFilename = "", InitialCatalog = "master" }.ConnectionString,
```
bricelam: Our guidelines suggest using string.Empty

bricelam: Hmm, maybe this should actually be null instead

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
axelheer: I'll rewrite this just to make sure.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
axelheer: Yeah, I'll add that.

src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerConnection.cs
axelheer: `SqlConnectionStringBuilder` throws an exception if null

axelheer:
Any idea why the newly added tests fail on SQL 2008 (CI build)? I've only Local DB 2016; works fine.

src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerConnection.cs
axelheer: `builder.Remove("AttachDBFilename");` does the trick

bricelam:
The CI build server instances probably don't have access to `%USERPROFILE%`. Only LocalDB and SQL Server Express user instances (via `UserInstance=True` in the connection string) would...

axelheer:
* Skip these tests, if not using LocalDB
* Use another directory instead (which one?)

Any suggestions?

axelheer:
Okay, the older SQL Server instance seems to throw a different error, if the database does not exist.

```
System.Data.SqlClient.SqlException : Unable to open the physical file "C:\windows\TEMP\Scratch_3a57a202-c4b6-4b16-9e9b-e55c39ba06d7.mdf". Operating system error 2: "2(The system cannot find the file specified.)".
Cannot attach the file 'C:\windows\TEMP\Scratch_3a57a202-c4b6-4b16-9e9b-e55c39ba06d7.mdf' as database 'Scratch_3a57a202-c4b6-4b16-9e9b-e55c39ba06d7'.
```

Just adding a check for error number 1832 to the `IsDoesNotExist` method of `SqlServerDatabaseCreator` isn't enough then. Any idea which sql error number I should add for that? (The AppVeyor error log doesn't unveil that...) Thanks!

bricelam:
> Skip these tests, if not using LocalDB

This one.

axelheer:
Eureka! AppVeyor went green, SQL 2008 seems to like this stuff now too. :tada:

@bricelam please let me know, if there's still something I should change.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
```diff
@@ -34,24 +34,44 @@ public static SqlServerTestStore GetOrCreateShared(string name, Action initializ
...
+
+                #if NET451
+
+                var tempFolder = Environment.GetEnvironmentVariable("TEMP", EnvironmentVariableTarget.Machine);
+                _fileName = Path.Combine(tempFolder, name + ".mdf");
+
+                #else
+
+                // TODO: find proper temp directory on .NET Core. Using the current user's temp directory doesn't work,
+                // since the SQL Server Instance may not have access to it (running under different user credentials)...
```
bricelam: This should be addressed before merging. I think we should use `AppContext.BaseDirectory` and *only* run these tests on LocalDB. Extending the [SqlServerCondition](https://github.com/aspnet/EntityFramework/blob/dev/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/Utilities/SqlServerConditionAttribute.cs#L61) enum would be the way to do this.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -405,7 +406,29 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
                 .Append("CREATE DATABASE ")
-                .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name))
+                .Append(SqlGenerationHelper.DelimitIdentifier(operation.Name));
+
+            if (!string.IsNullOrEmpty(operation.FileName))
+            {
+                var fileName = ExpandFileName(operation.FileName);
+                var name = Path.GetFileNameWithoutExtension(fileName);
+
+                var logFileName = Path.ChangeExtension(fileName, ".ldf");
```
bricelam: I seem to remember some weird issues when this didn't have `_log` appended to the filename. We should be safe and just do what EF6 did.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -405,7 +406,29 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
+                var logName = name + "_log";
+
+                builder
+                    .Append(" ON (NAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(name))
```
bricelam: Likewise, EF6 included the file extension here.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -416,6 +439,25 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
+            Check.NotNull(fileName, nameof(fileName));
+
+#if NET451
+
+            if (fileName.StartsWith("|DataDirectory|", StringComparison.OrdinalIgnoreCase))
+            {
+                var dataDirectory = AppDomain.CurrentDomain.GetData("DataDirectory") as string;
+                if (string.IsNullOrEmpty(dataDirectory))
+                    dataDirectory = AppDomain.CurrentDomain.BaseDirectory;
+                fileName = Path.Combine(dataDirectory, fileName.Substring("|DataDirectory|".Length));
```
bricelam: Might as well hardcode the length here.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -405,7 +406,29 @@ protected override void Generate(EnsureSchemaOperation operation, IModel model,
...
+                builder
+                    .Append(" ON (NAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(name))
+                    .Append("', FILENAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(fileName))
+                    .Append("') LOG ON (NAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(logName))
+                    .Append("', FILENAME = '")
+                    .Append(SqlGenerationHelper.EscapeLiteral(logFileName))
+                    .Append("')");
```
bricelam: I'd prefer the `ON` and `LOG ON` clauses to be on their own line, but I'll let you decide.

src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
```diff
@@ -170,7 +173,9 @@ public override Task<bool> ExistsAsync(CancellationToken cancellationToken = def
         }
 
         // Login failed is thrown when database does not exist (See Issue #776)
-        private static bool IsDoesNotExist(SqlException exception) => exception.Number == 4060;
+        // Unable to attach database file is thrown when file does not exist (See Issue #2810)
+        // Unable to open the physical file is thrown when file does not exist (See Issue #2810)
+        private static bool IsDoesNotExist(SqlException exception) => exception.Number == 4060 || exception.Number == 1832 || exception.Number == 5120;
```
bricelam: I'd like to understand these errors a little better. We need to make sure we're only catching the errors that happen when everything else is setup right. If the error when database can't access the path is different from when the file just doesn't exist, we should only catch the not exists one. Unless, like you say SQL2008 just throws a different error.

I'll set up some test machines with various versions of SQL Server and dig into this a bit more...

src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
```diff
@@ -189,7 +194,11 @@ private bool RetryOnExistsFailure(SqlException exception, ref int retryCount)
             // And (Number 4060):
             //   System.Data.SqlClient.SqlException: Cannot open database "X" requested by the login. The
             //   login failed.
-            if ((exception.Number == 233 || exception.Number == -2 || exception.Number == 4060)
+            // And (Number 1832)
+            //   System.Data.SqlClient.SqlException: Unable to Attach database file as database xxxxxxx.
+            // And (Number 5120)
+            //   System.Data.SqlClient.SqlException: Unable to open the physical file xxxxxxx.
+            if ((exception.Number == 233 || exception.Number == -2 || exception.Number == 4060 || exception.Number == 1832 || exception.Number == 5120)
                 && ++retryCount < 30)
```
bricelam: I ran the tests against SQL Server 2008 Express and was still getting `5120` after 30 retries. :frowning:

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
axelheer: Okay, I'll use `AppDomain.BaseDirectory` for classic .NET then too (consistency).

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
axelheer: Yeah, that would match the default behavior of SQL Server too.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
axelheer: But that's not what SQL Server does by default, I just checked that to be sure.

The *Bronies* database from the integration tests get `(name = 'Bronies', filename = 'C:\Users\Axel\Bronies.mdf')` and `(name = 'Bronies_log', filename = 'C:\Users\Axel\Bronies_log.ldf')`.

I would suggest to match SQL Server default's behavior instead of EF6.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
axelheer: Yeah, just a habit to improve readability. The compiler evaluates this to a constant expression, if I remember correctly. (If not, it won't matter here anyway...)

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
axelheer: Sure, I'll add some line breaks here.

src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
axelheer: Makes sense. "Unable to attach database" as "Unable to open" made enough sense for me to add them here. Especially since we're checking for "access denied" too, which could also mean the current user just doesn't have access to an indeed existing database (which should be okay, since a then wrongly executed create database operation would fail anyway?).

src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
axelheer: Hm, I've no idea why. :thinking:

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
bricelam: Sounds good.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
bricelam: I'm also fine leaving it as is.

axelheer:
This time AppVeyor cannot connect to GitHub.com. The CI god ist not with me for this PR. 😔

axelheer:
@bricelam should have addressed all the feedback and Travis is green now too. Anything else I can do?

bricelam:
Everything looks good! I just need to find some time to bash on it a bit with various versions of SQL Server before merging.

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/Operations/SqlServerCreateDatabaseOperation.cs
```diff
@@ -8,5 +8,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Operations
     public class SqlServerCreateDatabaseOperation : MigrationOperation
     {
         public virtual string Name { get; [param: NotNull] set; }
+
+        public virtual string FileName { get; [param: NotNull] set; }
```
bricelam: `CanBeNull`

bricelam:
Would you mind rebase-and-squashing on top of dev?

axelheer:
Done.

src/Microsoft.EntityFrameworkCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
bricelam: Finished testing and everything looks good. SQL Server 2010 & 2012 returned 5120 while 2014 & 2016 returned 1832.

bricelam:
Merged. Thanks for the contribution! I made a few tweaks to the tests to help me in my testing (see 656b175e0cb02d41f4e82fe1ccaed85fc38cda80)

AzureGulf:
using VS2017 CE, C#, SQLServer 2016 LocalDb, .NET 4.6.1, EFCore 1.1.1 developing a WPF desktop app there still seems to be a problem with the connection string. Connection in App.Config is:
connectionString="Data Source=(LocalDB)\MSSQLLocalDB;AttachDbFilename=|DataDirectory|\Prices.mdf;Database=Prices.mdf;Integrated Security=True"
but DataDirectory is ignored in published (ClickOnce) version.
If running in debug (in VS) it works fine and finds the bin/Debug directory.
Only way for published version to work is if I programatically substitute the DataDirectory with the full path to the AppDomain.CurrentDomain.BaseDirectory + my relative folder (I check and create the directory if necessary). The production code also seems to ignore any settings for DataDirectory that I make to AppDomain.CurrentDomain.SetData.
In production, I am using context.Database.Migrate() to create the database (and apply migrations, still to test) so that this happens without any user intervention.
This has taken over a day to partially resolve, prior to proper testing and when combined with ongoing problems with System.Net.Http, has consumed an inordinate amount of time.
Please let me know if this AttachDbFilename issue is now fully resolved - thanks.

ajcvickers:
@AzureGulf EF doesn't know anything about DataDirectory--it just gets passed to DbConnection, which then handles it. However, it is my understanding that DataDirectory is not guaranteed to be set in some environments. So it would seem prudent to not use it in connection strings.

AzureGulf:
@ajcvickers Thanks for your response and appreciate that what you say is probably correct. However, in my ignorance of DbConnection, the current mechanism is inconsistent in that it works whilst in the VS debugging version but not in the published version. Either it should work in both environments or fail in both, with a suitable error message instead of giving a run-time error in the published version.

axelheer:
Wether `DataDirectory` gets some love or not depends on the build being used. The full .NET Framework version supports it; the .NET Standard edition not, since there is no AppDomain available.

You can find [the code here](https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore.SqlServer/Migrations/Internal/SqlServerMigrationsSqlGenerator.cs#L534-L556).

Maybe publishing "selects" the .NET Standard build? (I've never used ClickOnce before...)

efidiles:
Hey guys,

We just ran into this ourselves; looks like relative paths aren't supported with `AttachDbFileName`.  

```
System.Data.SqlClient.SqlException (0x80131904): Cannot attach the file '.\OurDbName.mdf' as database 'OurDbName'.
```

Just wanted to express that here for other people who might be bumping into this.  The "fix" is use absolute paths.

axelheer:
@efidiles You can use the mighty `|DataDirectory|` placeholder too (see [here](https://github.com/aspnet/EntityFrameworkCore/blob/a588722e10f34a12b9cea3d655fd26aafd91051b/src/EFCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs#L669-L685)).

johnnyreilly:
Thanks @axelheer - we did try actually. But we just got errors when we tried... Does ` |DataDirectory|` have any documentation?

axelheer:
@johnnyreilly none that I know of, but you can just copy the mentioned snippet to your code and test how it works... 🤔

johnnyreilly:
Thanks @axelheer - I'll try again and report back....

bricelam:
You can influence `|DataDirectory|` by calling `AppDomain.CurrentDomain.SetData("DataDirectory", "C:/My/Path")`.

bricelam:
Here is [the documentation](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/sql-server-express-user-instances#using-the-124datadirectory124-substitution-string) for it.

johnnyreilly:
Thanks so much!

zhosn:
Hey guys,
using VS2017 , C#, SQLServer 2017 LocalDb, .NET 47.2, EFCore 2.2.6 developing a WPF desktop app there still seems to be a problem when occur in **SqlServerDatabaseCreator.cs**. The connection string is dynamic,Connection string is:
connectionString="Data Source=(LocalDB)\MSSQLLocalDB;Database=Sundial;AttachDbFilename=
"C:\\Repository\\User.mdf" ;Integrated Security=True" for First execution is ok, but Connection string is:
connectionString="Data Source=(LocalDB)\MSSQLLocalDB;Database=Sundial;AttachDbFilename=
"C:\\Repository\\UserInfo.mdf" ;Integrated Security=True" for Second execution with error:System.Data.SqlClient.SqlException
**Database 'Sundial' already exists. Choose a different database name.**

Is there any way to solve the problem of adding multiple AttachDbFilename, very Thanks!

ErikEJ:
Also change the value for the Database keyword the second time

zhosn:
> Also change the value for the Database keyword the second time

Thanks @ErikEJ -But the second time I couldn't create a new database, this project had **only one database**: 'Sundial'。**DbContext.Database.EnsureCreated()** is called CreateCreateOperations() every time，but **no AlterDatabaseOperation** operation is performed

ajcvickers:
@zhosn It's not at all clear to me what you are trying to do, but if you feel there is sill a bug here, then please file a new issue and include a small, runnable project/solution or complete code listing that demonstrates the behavior you are seeing.

zhosn:
@ajcvickers Thank you. Attached is the code list.  The environment： VS2017 , C#, SQLServer 2017 LocalDb, .NET 47.2, EFCore 2.2.6
[DBtest.zip](https://github.com/aspnet/EntityFrameworkCore/files/3395896/DBtest.zip)

ajcvickers:
@zhosn Thanks for the code, but I still don't understand what it is you are attempting to do?

zhosn:
@ajcvickers What I'm doing the project need to put each table in a single Attach Db File, not like a normal database design to unify all tables in a Attach Db File, so I will need to create multiple Attach Db Files in the database, but I don't know how to create multiple data files through the entity framework core, and in the OnModelCreating() I didn't find  to create multiple additional data files, the file group and create the tables in the file group of grammar.

ajcvickers:
@zhosn As was stated above, there is a one-to-one relationship between database files and database names. In other words, the database file _is_ the database. For a given database, you can't have one table in one file, and a different table in a different file--SQL Server doesn't work like that.

zhosn:
@ajcvickers Thank you very much for your reply. At present, EF Core may not be able to solve the technical problems in my project. I started trying to use EF core for sqlite and EF core for sqlce, but both failed to solve the technical barrier of accessing multiple databases from the same Dbcontext. Now I need to find another way.

Issue #6453, opened 2016-09-02
Query :: during nav-rewrite we sometimes add null protection logic in places where it's not needed


maumar:
Every time we encounter optional navigation we add null protection logic to the member access, however this is not necessary. We only need to add null protection AFTER the first optional navigation. 
Example can be found in #6301

maumar:
fixed in

PR #6454, opened 2016-09-02
[SMALL] Fix to #6453 - Query :: during nav-rewrite we sometimes add null protection logic in places where it's not needed

maumar:
Problem was that when we rewrite optional navigation, we add null protection logic to it's property access right away. This is not needed - we only need to add null-ref protection after the first optional navigation, e.g.:

query:
customers.Select(c => c.Optional.Id)

should be translated to:
customers.Select(c => c.Optional?.Id)

and not:
customers.Select(c => c?.Optional?.Id)


maumar:
@smitpatel

smitpatel:
:shipit:

PR #6462, opened 2016-09-02
Query: Optimize trivial GroupJoin/SelectMany cases by rewriting to Jo…

anpete:
…in. E.g.
```c#
from c in Customers
join o in Orders into os
from o in os
```
is transformed into:
```c#
from c in Customers
join o in Orders
```


anpete:
@maumar

maumar:
:shipit:

PR #6464, opened 2016-09-02
Some fixes for memory optimized tables

AndriySvyryd:
Avoid changing OnDelete behavior for memory optimized tables
Avoid generating unique indexes for nullable columns for memory optimized tables


divega:
:shipit:

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -309,15 +320,16 @@ protected override void Generate(CreateIndexOperation operation, IModel model, M
...
+                if (operation.IsUnique
+                    && nullableColumns.Count == 0)
                 {
                     builder.Append("UNIQUE ");
                 }
 
                 IndexTraits(operation, model, builder);
 
                 builder
-                    .Append(" (")
```
divega: Thanks for removing that extra space :smile:

Issue #4402, opened 2016-01-26
Fix non-Windows dotnet build


natemcmaster:
The transition to dotnet CLI broke our Linux and OSX build/test support. Use this issue to track overall progress on restoring non-Windows build/test

- [ ] Linux
- [ ] OSX

ErikEJ:
And the AppVeyor build seems to be well broken also...

natemcmaster:
Blocking issues:
https://github.com/NuGet/Home/issues/1959
https://bugzilla.xamarin.com/show_bug.cgi?id=35947

victorhurdugaci:
Do you need mono 4.2? We usually use 4.0.5

natemcmaster:
Short answer: yes and no.
Longer answer: our testing is currently configured to work on Mono 4.0.x, but it skips a huge number of tests that fail due to features that were not added until 4.2.x. So tests on 4.0.x should pass, but 4.2.x is preferred.

PR #3896, opened 2015-11-25
Run all tests on CoreCLR

natemcmaster:
Run **all** tests on CoreCLR, not just functional tests. 

Fix #3781

PR #6472, opened 2016-09-04
Increate .travis.yml consistency between repos

dougbu:
- aspnet/Universe#349
- add `sudo: required` recommended for the Trusty image
- use same `env` syntax as in other repos

questions:
- why are OS X builds disabled for EF? most other repos use `osx_image: xcode7.1`
- why does EF use Mono "beta" instead of 4.0.5?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: csharp
+sudo: required
 dist: trusty
 os:
   - linux
@@ -16,8 +17,8 @@ addons:
     - zlib1g
 env:
   global:
-    DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
-    DOTNET_CLI_TELEMETRY_OPTOUT: 1
+    - DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+    - DOTNET_CLI_TELEMETRY_OPTOUT: 1
 branches:
   only:
     - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ mono:
   - 4.0.5
 os:
   - linux
-  - osx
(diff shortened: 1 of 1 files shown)


dougbu:
@divega could you or someone on your team please answer the questions above? The `.travis.yml` file used in this repo is different from the rest and I'd like to make it more consistent than I've done so far.

smitpatel:
For Mono see #4402

smitpatel:
osx was removed as part of #3896
See more discussion about it on https://github.com/dotnet/corefx/pull/4729
It was disabled because it was not working. If osx build is working on travis then we should enable it.

dougbu:
@smitpatel 
> For Mono see #4402

But does EF test using Mono at all anymore? That's not done in any other repo.

> If osx build is working on travis then we should enable it.

I'll run my script again and add another commit to enable OSX build here. We'll see.

Guess I'll experiment w/ the Mono version as well.

dougbu:
@smitpatel #3896 mentions EF CI machines. Are those OSX machines separate from the Travis builds? Do builds there use `.travis.yml` settings? Only Travis builds should be affected here but I want to confirm.

dougbu:
FYI two builds in a row failed on OSX. Failures were output timeouts in different test projects. Since these projects normally complete testing in less than 5 minutes and the output timeout is 10 minutes, I don't know what's going on or why only the EF build on OSX is affected. Made the one-line change to disable the OSX build.

Suggest someone on the EF team investigate the timeouts further. See https://travis-ci.org/aspnet/EntityFramework/jobs/157506602 for the second build w/ OSX enabled. (Can't find the first log due to restarting the run but it ended in the same way, just in a later test project.)

dougbu:
1c8c510

PR #6479, opened 2016-09-06
SQL Server Migrations: Rebuild indexes

### LINKED ISSUE #6102: SQL Server Migrations: Rebuild indices on narrowed columns
### Steps to reproduce
1. Create a many-to-many join entity initially with only navigation properties, no foreign key properties

    public class CourseInstructor
    {
        public int ID { get; set; }
        public Instructor Instructor { get; set; }
        public Course Course { get; set; }
    }

2. Later after adding migration for it and updating database, add foreign keys to it

    public class CourseInstructor
    {
        public int ID { get; set; }
        public int InstructorID { get; set; }
        public int CourseID { get; set; }
        public Instructor Instructor { get; set; }
        public Course Course { get; set; }
    }

    The application continues to run, doesn’t say you have to do a migration, doesn’t realize a migration is needed.

3. Run add-migration command, succeeds.
4. Run update-database and it fails:

    The Up method wants to change the foreign key columns from nullable to not-null, and it wants to change delete action from restrict to cascade.
    To do that, the Up method drops FK constraint, alters column, adds back FK constraint with cascade setting.  What it doesn’t do is drop the indexes that were created on those columns, so the alter column statements fail.

5. Manually add drop index and add index statements, and update-database works.

### The issue

Up method scaffolding should drop and re-add the indexes on the foreign key columns.

### Further technical details

EF Core version: "Microsoft.EntityFrameworkCore.Tools": "1.0.0-preview2-final",
Operating system: Windows 10
Visual Studio version: VS 2015

bricelam:
Resolves #6102

This also includes a fair amount of cleanup to the code and generated SQL. #YakShaving

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
@@ -329,7 +329,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
                     && targetMigrationsAnnotations.Count > 0)
                 {
                     var alterDatabaseOperation = new AlterDatabaseOperation();
-                    CopyAnnotations(targetMigrationsAnnotations, alterDatabaseOperation);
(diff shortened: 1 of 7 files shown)


ajcvickers:
:shipit:

PR #6486, opened 2016-09-07
Sql Server Migrations: Throw when altering IDENTITY

### LINKED ISSUE #5345: Sql Server Migrations: Throw when altering IDENTITY
If you go through the process of changing the  ApplicationUser  ID type to  int, then you will hit issues with EF migrations. On SQL Server, you'll get errors including:

The object 'PK_AspNetUsers' is dependent on column 'Id'.
 The object 'FK_AspNetUserRoles_AspNetUsers_UserId' is dependent on column 'Id'.
 The object 'FK_AspNetUserLogins_AspNetUsers_UserId' is dependent on column 'Id'.
 The object 'FK_AspNetUserClaims_AspNetUsers_UserId' is dependent on column 'Id'.
 ALTER TABLE ALTER COLUMN Id failed because one or more objects access this column.

bricelam:
Fixes #5345


AndriySvyryd:
:shipit:

PR #6489, opened 2016-09-08
WIP: Complex type metadata

ajcvickers:
Note: this is not a PR that will be merged

This is a work-in-progress PR containing some of the Complex Type metadata ideas that we have discussed. Ideas and feedback welcome.

@divega @rowanmiller @AndriySvyryd @anpete @bricelam


src/Microsoft.EntityFrameworkCore/Metadata/IProperty.cs
```diff
@@ -1,78 +1,21 @@
...
-        bool RequiresValueGenerator { get; }
-
-        /// <summary>
-        ///     Gets a value indicating whether this is a shadow property. A shadow property is one that does not have a
-        ///     corresponding property in the entity class. The current value for the property is stored in
-        ///     the <see cref="ChangeTracker" /> rather than being stored in instances of the entity class.
+        ///     TODO: ComplexType docs
         /// </summary>
-        bool IsShadowProperty { get; }
+        new IEntityType DeclaringType { get; }
```
AndriySvyryd: I think this is redundant with DeclaringEntityType

src/Microsoft.EntityFrameworkCore/Metadata/IMutableProperty.cs
```diff
@@ -14,62 +12,16 @@ namespace Microsoft.EntityFrameworkCore.Metadata
...
-        /// </summary>
-        new ValueGenerated ValueGenerated { get; set; }
-
-        /// <summary>
-        ///     Gets or sets a value indicating whether or not this property can be modified before the entity is
-        ///     saved to the database. If true, an exception will be thrown if a value is assigned to
-        ///     this property when it is in the <see cref="EntityState.Added" /> state.
-        /// </summary>
-        new bool IsReadOnlyBeforeSave { get; set; }
+        new IMutableEntityType DeclaringType { get; }
```
AndriySvyryd: Redundant?

src/Microsoft.EntityFrameworkCore/Metadata/IMutableStructuralType.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        ///     TODO: ComplexType docs
+        /// </summary>
+        IMutableComplexTypeReference RemoveComplexTypeReference([NotNull] string name);
+
+        /// <summary>
+        ///     TODO: ComplexType docs
+        /// </summary>
+        void PropertyMetadataChanged();
```
AndriySvyryd: This shouldn't be part of the interface

src/Microsoft.EntityFrameworkCore/Metadata/IProperty.cs
ajcvickers: It is redundant in the sense that it is always the same thing as the DeclaringEntityType, but it's always going to be there because it comes from the base interface, and since it is always going to be there and we know it will be an IEntityType, then it makes sense to me to redefine it as such.

src/Microsoft.EntityFrameworkCore/Metadata/IMutableStructuralType.cs
ajcvickers: Will move.

src/Microsoft.EntityFrameworkCore/Metadata/IComplexPropertyUsage.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    ///     Usage of a complex property in an entity type
+    ///     TODO: ComplexType docs
+    /// </summary>
+    public interface IComplexPropertyUsage : IStructuralProperty, IPropertyBase
+    {
+        /// <summary>
+        ///     TODO: ComplexType docs
+        /// </summary>
+        new IComplexType DeclaringType { get; }
```
AndriySvyryd: This seems confusing, the declaring type should be the entity type while the complex type would be the defining type. It's also redundant since you can get the defining complex type from DefiningProperty

ajcvickers: I guess it depends what we decide "DeclaringType" means in this case. It's redundant in either case because the entity type can be obtained from the ComplexTypeUsage

src/Microsoft.EntityFrameworkCore/Metadata/IProperty.cs
ajcvickers: Extension properties don't exist...

ajcvickers: But in principal I agree that if we decide we can rename DeclaringEntityType, then it cleans things up.

src/Microsoft.EntityFrameworkCore/Metadata/IComplexPropertyUsage.cs
AndriySvyryd: DeclaringType should always be the least derived type that returns this instance in .GetProperties()

---
In reply to: [78063653](https://github.com/aspnet/EntityFramework/pull/6489#discussion_r78063653) [](ancestors = 78063653)

ajcvickers: I can go with that.

src/Microsoft.EntityFrameworkCore/Metadata/IProperty.cs
AndriySvyryd: Yes, also if we didn't do any of what I proposed we would also need to hide it in INavigation and IMutableNavigation

---
In reply to: [78064211](https://github.com/aspnet/EntityFramework/pull/6489#discussion_r78064211) [](ancestors = 78064211)

src/Microsoft.EntityFrameworkCore/Metadata/IMutableStructuralType.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        /// <summary>
+        ///     TODO: ComplexType docs
+        /// </summary>
+        new IMutableModel Model { get; }
+
+        /// <summary>
+        ///     TODO: ComplexType docs
+        /// </summary>
+        IMutableStructuralProperty AddProperty([NotNull] string name, [NotNull] Type propertyType, bool shadow);
```
AndriySvyryd: Not directly related to the current changes, but I think this should be split into 3 overloads:
1. With only 2 parameters that adds a shadow property
2. With a PropertyInfo parameter that adds  a non-shadow property that later could also have a FieldInfo
3. With a FieldInfo parameter that adds a non-shadow property without a PropertyInfo

This would be more in-line with the explicitness of this layer

src/Microsoft.EntityFrameworkCore/Metadata/IProperty.cs
```diff
@@ -1,78 +1,21 @@
...
         /// <summary>
-        ///     Gets a value indicating whether this property is used as a concurrency token. When a property is configured
-        ///     as a concurrency token the value in the database will be checked when an instance of this entity type
-        ///     is updated or deleted during <see cref="DbContext.SaveChanges()" /> to ensure it has not changed since
-        ///     the instance was retrieved from the database. If it has changed, an exception will be thrown and the
-        ///     changes will not be applied to the database.
+        ///     TODO: ComplexType docs
         /// </summary>
-        bool IsConcurrencyToken { get; }
+        new string Name { get; }
```
AndriySvyryd: Is this required?

src/Microsoft.EntityFrameworkCore/Metadata/IComplexPropertyUsage.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    /// <summary>
+    ///     Usage of a complex property in an entity type
+    ///     TODO: ComplexType docs
+    /// </summary>
+    public interface IComplexPropertyUsage : IStructuralProperty, IPropertyBase
```
AndriySvyryd: Shouldn't this derive from IProperty instead?

src/Microsoft.EntityFrameworkCore/Metadata/IProperty.cs
ajcvickers: Will remove

src/Microsoft.EntityFrameworkCore/Metadata/IComplexPropertyUsage.cs
ajcvickers: Yep.

src/Microsoft.EntityFrameworkCore/Metadata/IComplexTypeUsage.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Collections.Generic;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    /// <summary>
+    ///     A referene to a usage of a complex type in an entity or complex type
+    ///     TODO: ComplexType docs
+    /// </summary>
+    public interface IComplexTypeUsage : IPropertyBase
```
AndriySvyryd: This should be renamed to IComplexTypeReferenceUsage

src/Microsoft.EntityFrameworkCore/Metadata/Internal/ComplexPropertyUsage.cs
```diff
@@ -0,0 +1,213 @@
...
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class ComplexPropertyUsage : StructuralProperty, IMutableComplexPropertyUsage
```
AndriySvyryd: If this derived from Property it would have little logic in it

src/Microsoft.EntityFrameworkCore/Metadata/IComplexTypeUsage.cs
ajcvickers: Yes, except that is very long. :-)

src/Microsoft.EntityFrameworkCore/Metadata/Internal/ComplexPropertyUsage.cs
ajcvickers: I'll look into it.

src/Microsoft.EntityFrameworkCore/Metadata/Internal/ComplexTypeUsage.cs
```diff
@@ -0,0 +1,238 @@
...
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual EntityType DeclaringEntityType { get; }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual PropertyIndexes PropertyIndexes
```
AndriySvyryd: This should need less indexes, perhaps we should use a smaller type to store them. Same for Navigation indexes

src/Microsoft.EntityFrameworkCore/Metadata/IComplexTypeUsage.cs
AndriySvyryd: That hasn't stopped us before

---
In reply to: [78074234](https://github.com/aspnet/EntityFramework/pull/6489#discussion_r78074234) [](ancestors = 78074234)

src/Microsoft.EntityFrameworkCore/Metadata/Internal/ComplexTypeUsage.cs
ajcvickers: I was thinking about that. They may need only the basic index, possibly with shadowIndex too. Navigations require a basic index and a relationship index.

src/Microsoft.EntityFrameworkCore/Metadata/Internal/ComplexPropertyDefinition.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override bool DefaultIsReadOnlyAfterSave
+            // TODO: Check is used in any key
+            => (ValueGenerated == ValueGenerated.OnAddOrUpdate)
+               && !IsStoreGeneratedAlways;
```
AndriySvyryd: The definition shouldn't have any logic for the defaults as the values it's using might not be the same as on the usage

AndriySvyryd: I think it would make more sense to make the properties on IComplexPropertyDefinition nullable so they are only used when set, this way we don't need to come up with some arbitrary default value

---
In reply to: [78076928](https://github.com/aspnet/EntityFramework/pull/6489#discussion_r78076928) [](ancestors = 78076928)

AndriySvyryd: This shows why it's semantically incorrect to derive IComplexPropertyDefinition from IStructuralProperty even though they are structurally compatible

---
In reply to: [78077484](https://github.com/aspnet/EntityFramework/pull/6489#discussion_r78077484) [](ancestors = 78077484,78076928)

ajcvickers: I'm not sure what is best for this whole thing. For example, if I make the definition of the property explicitly nullable, but some uses are in keys, then do those usages just ignore that the definition was set nullable? Or should it throw immediately? Or in validation?

ajcvickers: I think if we don't do this then we have to ask what IStructuralProperty is for, or indeed what the definitions are for. If I want to set the max length of any string property to some value, then if the way I do that is to traverse all usages of the property, then why have this? And if the answer is for query, then when used in the context of a query it surely is a structural property; it just doesn't belong to an entity type.

AndriySvyryd: I think we don't need IStructuralProperty or IStructuralType. The definition provides the defaults, so you don't have to traverse all the usages.
The usage implementation of IsNullable would be _isNullable ?? DefiningProperty.IsNullable ?? DefaultIsNullable

---
In reply to: [78080907](https://github.com/aspnet/EntityFramework/pull/6489#discussion_r78080907) [](ancestors = 78080907)

ajcvickers: But then I have to write the code twice--once for setting the default on the entity types, and again for setting the default on the complex type definitions.

AndriySvyryd: What do you mean? Are you speaking as a user or as the implementer?

---
In reply to: [78082310](https://github.com/aspnet/EntityFramework/pull/6489#discussion_r78082310) [](ancestors = 78082310)

ajcvickers: User
```C#
foreach (var structuralType in model.GetStructuralTypes())
{
    foreach (var structuralProperty in structuralType.GetProperties())
    {
        if (structuralProperty.ClrType == typeof(string))
        {
            structuralProperty.SetMaxLength(100);
        }
    }
}
```

AndriySvyryd: We've discussed this, doing it on GetEntityTypes will have the same result, unless you add a new ComplexTypeReferenceUsage after this

---
In reply to: [78083880](https://github.com/aspnet/EntityFramework/pull/6489#discussion_r78083880) [](ancestors = 78083880)

Issue #6494, opened 2016-09-08
MSBuild-ify tools


muratg:
### Microsoft.EntityFrameworkCore.Tools
- Expect to run independently, via the CLI
- Expect to run via MSBuild
- Runtime dependencies/Load Context thingy
- Project namespace (in MSBuild)
- OutputType
- TargetPath
### TODO
- Fork tools repos to `rel/1.0.0-preview3` branches for "October" release. Dev branches will continue to be used for 1.1 release.
- Use desktop msbuild to start our work until the new msbuild is ready for   consumption
### Notes
- [This](https://github.com/aspnet/Coherence-Signed/issues/375) tracks the work across the tools repos.
- [This](https://github.com/aspnet/specs/blob/master/notes/Tools.md) is the previous discussion on the topic.

bricelam:
Clearing the milestone since it sounds like the plans for the October release have changed.

bricelam:
Closing as a dupe of #6139 (cross-repo issue updated)

PR #6495, opened 2016-09-08
Remove tools

bricelam:
Unfortunately, most of the Reverse Engineer tests reused components from the tooling assemblies. (see [d8f0a53](https://github.com/aspnet/EntityFramework/pull/6495/commits/d8f0a533a6fe393976fba0f85b49f9638ed01b1a)) We'll need to rethink where these tests live and/or add more unit and functional tests to regain coverage.

Part of #6494


ErikEJ:
Yikes  😢

bricelam:
The simplest solution is probably just to move these projects into the tool repository too for now...

The motivation behind this PR is that the tools (CLI and NuGet) will be shipping on a different cadence than the runtime, so moving them to a separate reop will make branching and building simpler.

bricelam:
Recalling until we decide what to do with the RevEng tests. Filed #6500

PR #6512, opened 2016-09-12
Make operator generation more flexible

roji:
Current operator generation in queries is inflexible - the current logic selects an operator based only on the expression's NodeType. Cases where a different operator needs to be generated based on others factors (e.g. the expression's Type) are difficult to integrate.

A good example is string concatenation in PostgreSQL, which requires operator `|` rather than `+`. The current code hacks a special check for this directly into `DefaultQuerySqlGenerator`, defining a special ConcatOperator which can be overridden by providers. Another example is bitwise operators for boolean - PostgreSQL does not support `boolean & boolean`, so `ExpressionType.And` should be rendered as `&&` based on the expression's Type.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -33,7 +33,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
         private ParameterNameGenerator _parameterNameGenerator;
         private RelationalTypeMapping _typeMapping;
 
-        private static readonly Dictionary<ExpressionType, string> _binaryOperatorMap = new Dictionary<ExpressionType, string>
+        private static readonly Dictionary<ExpressionType, string> _operatorMap = new Dictionary<ExpressionType, string>
         {
             { ExpressionType.Equal, " = " },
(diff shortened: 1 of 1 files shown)


roji:
Sorry for not testing before submitting, will fix the Sqlite issues.

roji:
Amended and rebased

roji:
FYI Linux build seems to be failing for unrelated reasons

anpete:
@roji, Thanks for the PR. Unfortunately DefaultQuerySqlGenerator is public surface and so we need to do this in a non-breaking way. We do have the option of obsoleting methods, but not removing them.

roji:
Oh, I wasn't aware you were on a strict semver policy on provider-facing classes such as DefaultQuerySqlGenerator. Does that mean you expect users to subclass DefaultQuerySqlGenerator, or is it simply to provide a stable API to providers?

Regardless, I understand that this is your policy - I'll try to resubmit this change in a way that doesn't break anything.

anpete:
> Does that mean you expect users to subclass DefaultQuerySqlGenerator, or is it simply to provide a stable API to providers

Both. Anything in a .Internal namespace is fair game, btw.

Thanks!

roji:
@anpete, have pushed another version which leaves all protected APIs in place, and using them from the new `GenerateOperator(Expression)` API. It's a tiny bit awkward since it has to call the existing `{Try,}GenerateBinaryOperator`, but only if the given expression is binary etc.

PR #6518, opened 2016-09-13
Remove assert for unknown migration ops

roji:
The 1.0.1 MigrationsModelDiffer contains an assert which makes sure there are no unknown migration operations. This prevents providers (such as Npgsql) from adding their own custom migration operations and handling them in subclasses of MigrationsModelDiffer.


bricelam:
Closing since I believe this PR is no longer necessary. But thanks for submitting since it spawned a great conversation!

PR #6519, opened 2016-09-13
Remove assert for unknown migration ops

roji:
The 1.0.1 MigrationsModelDiffer contains an assert which makes sure there are no unknown migration operations. This prevents providers (such as Npgsql) from adding their own custom migration operations and handling them in subclasses of MigrationsModelDiffer.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
@@ -183,7 +183,6 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
                 }
                 else
                 {
-                    Debug.Assert(false, "Unexpected operation type: " + operation.GetType());
                     leftovers.Add(operation);
                 }
             }


bricelam:
> subclasses of MigrationsModelDiffer

Please don't subclass MigrationsModelDiffer; it is an internal class and not intended to be overridden by providers

bricelam:
AFAIK, only the services exposed by [IRelationalDatabaseProviderServices](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalDatabaseProviderServices.cs) should be replaced by providers. But @ajcvickers might know better.

ajcvickers:
@bricelam @roji Correct, a provider should only replace provider services, not core services. Otherwise any app adding two different providers will get the service from one provider or the other in an essentially arbitrary manner. If an existing core service needs provider-specific functionality, then that will require a factoring out of the service such that it goes through correct provider resolution.

roji:
Hmm, in that case I'm going to need some guidance... Here's what's going on.

[PostgreSQL has the concept of "extensions"](https://www.postgresql.org/docs/current/static/sql-createextension.html) which can be added to a database. Extensions can define new data types, new functions, and extend the database in other ways. For example, PostGIS is a prominent extension which adds GIS support to PostgreSQL. For another example, `uuid-ossp` is an extension which provides functions for generating UUIDs - without this extension PostgreSQL doesn't provide any way to generate a random UUID.

Extensions are set up on a per-database basis, by issuing `CREATE EXTENSION <name>`. They resemble other global database objects such sequences or tables (extensions can even be "created" in a schema). In the Npgsql provider I've modeled extensions much like other database entities: Npgsql's model builder has a `HasPostgresExtension()` method which allows users to specify that an extension is needed in the database created by the model. There's also an `EnsurePostgresExtensionOperation` migration which MigrationsSqlGenerator knows how to handle, generating a `CREATE EXTENSION` statement. And accordingly there needs to be a ModelDiffer which knows how to compare two models and understand that an extension has been added, etc. (note that the create extension migration must be sorted properly with respect to other migration operations which may depend on it).

All this is already working well in the 1.0.0 release - users are cleanly managing PostgreSQL extensions via their model. I definitely was not aware that MigrationsModelDiffer wasn't to be subclassed (everything is protected virtual etc.). It's important for EFCore to support migration operations beyond the standard ones, and this seems to impact MigrationsModelDiffer as well.

If there's some way to achieve custom migrations without touching MigrationsModelDiffer, I'd be happy to do it.

bricelam:
In 1.1.0, we'e added a general `AlterDatabase` operation that may work. It is generated based on differences in the annotations on `IModel` (as provided through `IMigrationsAnnotationProvider.For(IModel)`). We're using it to [ensure a FILEGROUP has been created](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs#L571) when memory-optimized tables are used.

There's no clean way to do this prior to 1.1.0, but one option would be to annotate CreateTable and/or AddColumn operations to specify which extensions are needed. Then at the beginning of the MigrationsSqlGenerator, scan all the operations for which extensions are needed and ensure they have been created. This isn't clean because it will produce CREATE IF NOT EXISTS statements for them every migration, but it should at least be functional.

roji:
I understand setting up extensions through annotations could work, although it's seems like a shame to restrict migrations to only the list supported out-of-the-box by the core. While I can see how FILEGROUP fits nicely as a database annotation, extensions really do resemble database objects like sequences or schemas. There may be other objects out there in various databases which aren't a table or sequence, and it would be good for providers to be able to support them.

Also, note that it doesn't seem practical for the provider to infer which extensions are needed by scanning migrations; the user may have embedded some function call in some column's default value. It's much better for the user to explicitly specify which extensions they want on their databases.

Would you consider allowing custom migrations in EFCore, is this a discussion you're willing to have internally?

If this really isn't a direction you'd like to take, can I at least suggest you remove the `Debug.Assert()` in 1.0.1? I can make the necessary changes to switch to database annotations for 1.1, but for 1.0 there doesn't seem to be an alternative.

divega:
@pranavkm I assume you closed this PR because it was on 1.0.1?

pranavkm:
We made the branch into a tag - https://github.com/aspnet/EntityFramework/tree/rel/1.0.1. I guess removing the branch caused this PR to be closed.

divega:
Thanks @pranavkm. Makes sense.

@roji We are releaseing 1.0.1 very soon so it is actually late to make this change for 1.0.1. Since this is just a `Debug.Assert()` I assume it is not needed for you provider to work with released packages, but maybe for you to debug using the source? If that is the case, would it be enough for you to use a forked version of the source code for now or is it needed for something else?

I think the rest of the migrations extensibility discussion in this PR deserves having its own issue to track it. Would you prefer to create it or us to copy some parts of the discussion into a new issue?

ajcvickers:
@roji We discussed this in triage and I think we are open to the idea of making Migrations more extensible. We have been holding back on some things because there was a desire to be able to scaffold Migrations for multiple providers as opposed to just the current provider. However, realistically, we probably won't get to doing that, which makes it easier to make some changes. I am going to send out a PR to make the model differ a provider service. This will allow you to override it while still allowing other providers to use the default. Beyond this, it would be useful to know what else you need to get your custom operations working across the stack. For example, what are you doing for C# code gen? Are there other services that your provider needs to replace that are not in IDatabaseProviderServices or IRelationalDatabaseProviderServices? We're basically looking for suggestions from you on what you need to extend in the way you want.

roji:
> @roji We are releaseing 1.0.1 very soon so it is actually late to make this change for 1.0.1. Since this is just a Debug.Assert() I assume it is not needed for you provider to work with released packages, but maybe for you to debug using the source? If that is the case, would it be enough for you to use a forked version of the source code for now or is it needed for something else?

Wow, I seem to have really bad timing...!

But as you say this isn't a showstopper, being just a Debug.Assert() -the main problem is me running tests on the custom migration type and triggering an assertion failure. I'll manage to work around it, probably with a forked version as you suggest.

@ajcvickers I can see you already opened #6522. I opened #6524 to generally track custom migration operations, and included an additional requirement for a provider-specific CSharpMigrationOperationGenerator. I think that's all that's needed, but I'll double-check that tomorrow.

Thanks for listening and for including support for this. I hope it makes sense as a general EFCore feature and not just as a response to PostgreSQL extensions...!

roji:
I somehow forgot that my test projects consume EFCore via nuget, which includes release builds so no `Debug.Assert()`, even less of a problem.

Issue #6522, opened 2016-09-13
Make IMigrationsModelDiffer a provider service


ajcvickers:
This will allow providers to create custom model differs without clobbering each other.

bricelam:
Given our [design meeting discussion](https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes-(September-15,-2016)) on this, I think we can close it as wontfix for now.

PR #6523, opened 2016-09-13
Make IMigrationsModelDiffer a provider service

ajcvickers:
Issue #6522

Service was already scoped, so the change is minimal.


AndriySvyryd:
`MigrationsModelDiffer.Sort` should be split into two methods so that it doesn't need to be completely reimplemented when adding an operation

ajcvickers:
@AndriySvyryd Filed #6529

bricelam:
Should `MigrationsModelDiffer` move out of `.Internal`?

ajcvickers:
@bricelam Possibly, if we are recommending that it be extended for this scenario. Currently the service is public, but our implementation is internal, which is normal for services that we expect people to replace. But if we expect the service to be replaced by extending the existing implementation, then it probably should be public.

ajcvickers:
It's probably worth saying that the point of this PR is to allow the PostgreSQL provider to do what it is currently doing in a way that is friendly in multiple provider scenarios. While we can choose to do more to support doing this, for example, without using internal code, it is valid for a provider to use internal code as long as it is understood that the provider may be broken when there is an EF update. So I think we should get this change in as a minimum step, and then choose to do more or not as appropriate.

rowanmiller:
Agreed with not needing to move the implementation out of internal at this stage

bricelam:
Given our [design meeting discussion](https://github.com/aspnet/EntityFramework/wiki/Design-Meeting-Notes-(September-15,-2016)) on this, I think we can close it now.

PR #6539, opened 2016-09-15
Remove xproj solution

bricelam:
It doesn't work anyway because it's too big for VS to handle.


ajcvickers:
:shipit:

rowanmiller:
BTW I will take care of passing along the issues we hit to the correct folks

Issue #6542, opened 2016-09-15
Change IPropertyBase.DeclaringEntityType to DeclaringType


ajcvickers:
This allows IPropertyBase to be a more general base types for properties, most notably those for complex property definitions. The main change would be that IPropertyBase now has DeclaringType rather than DeclaringEntityType, and this would be of type ITypeBase. This makes it a suitable base interface for property definitions of complex types. This is a breaking change, but one that we think will have minimal impact, especially given that the two subtypes of IPropertyBase (IProperty and INavigation) now get DeclaringEntityType.

PR #6543, opened 2016-09-15
Introduce ITypeBase and related changes

ajcvickers:
Issue #6542

This is a change that has come out of the design for complex types. The main change is that IPropertyBase now has DeclaringType rather than DeclaringEntityType. This makes it a suitable base interface for property definitions of complex types. This is a breaking change, but one that we think will have minimal impact, especially given that the two subtypes of IPropertyBase (IProperty and INavigation) now get DeclaringEntityType. Nonetheless, it seems prudent to do it in 1.1 even if complex types don't make it into 1.1. Also, there are other changes here in code that has not yet shipped, and hence it is good to make these changes before shipping.


ajcvickers:
Merged

Issue #6548, opened 2016-09-15
Remove ability to force shadowness in core metadata


ajcvickers:
Currently our core metadata allows a property to be forced to be a shadow property even if a property or field with the same name exists. It's only possible to do this on low-level APIs--our fluent APIs and conventions would never do this. Almost all of this low-level code is also marked internal.

Given where things have evolved this doesn't seem to be useful functionality and it complicates the code used for defining properties. As we add more different kinds of properties it would be nice to have all of them consistent in this regard--that is all simple without the ability to force this shadowness. Therefore, we should remove this functionality from core properties.

PR #6558, opened 2016-09-16
Remove ability to force shadowness in core metadata

ajcvickers:
Issue #6548

This boils down to two public overloads of AddProperty--one is called when the MemberInfo is known, the other when it is not known. The latter looks up the MemberInfo and then they collapse together into one method calling one constructor.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/IMutableEntityType.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/IMutableEntityType.cs
@@ -162,9 +162,8 @@ namespace Microsoft.EntityFrameworkCore.Metadata
         /// </summary>
         /// <param name="name"> The name of the property to add. </param>
         /// <param name="propertyType"> The type of value the property will hold. </param>
-        /// <param name="shadow"> Whether the property is in shadow-state. </param>
         /// <returns> The newly created property. </returns>
-        IMutableProperty AddProperty([NotNull] string name, [NotNull] Type propertyType, bool shadow);
+        IMutableProperty AddProperty([NotNull] string name, [CanBeNull] Type propertyType);
 
         /// <summary>
         ///     <para>
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/Extensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/Extensions.cs
@@ -78,7 +78,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         {
             foreach (var property in sourceEntityType.GetDeclaredProperties())
             {
-                var clonedProperty = targetEntityType.AddProperty(property.Name, property.ClrType, property.IsShadowProperty);
+                var clonedProperty = targetEntityType.AddProperty(property.Name, property.ClrType);
                 clonedProperty.IsNullable = property.IsNullable;
                 clonedProperty.IsConcurrencyToken = property.IsConcurrencyToken;
                 clonedProperty.RequiresValueGenerator = property.RequiresValueGenerator;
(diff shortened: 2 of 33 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/IMutableEntityType.cs
```diff
@@ -162,9 +162,8 @@ public interface IMutableEntityType : IEntityType, IMutableTypeBase
         /// </summary>
         /// <param name="name"> The name of the property to add. </param>
         /// <param name="propertyType"> The type of value the property will hold. </param>
-        /// <param name="shadow"> Whether the property is in shadow-state. </param>
         /// <returns> The newly created property. </returns>
-        IMutableProperty AddProperty([NotNull] string name, [NotNull] Type propertyType, bool shadow);
+        IMutableProperty AddProperty([NotNull] string name, [CanBeNull] Type propertyType);
```
AndriySvyryd: Why is propertyType optional now in all these methods?

ajcvickers: Why was it required before?

ajcvickers: Or, to be more specific, previously you could call this method with null and if a CLR property was matched, then the call would succeed. So it seems like it was never really a required parameter.

AndriySvyryd: Because it needs to be specified explicitly for shadow properties, while non-shadow use the MemberInfo overload

ajcvickers: Except it doesn't...

AndriySvyryd: It does, we throw NoPropertyType if it's null

ajcvickers: Are you suggesting that we make the string overload stop working for non-shadow properties? In 1.0 you could use it, and our code did use it. We now still use it. If you give us the MemberInfo, then we will use that, but if you don't give us the MemberInfo then we look it up for you. That was what we used to, and we still do it now. That code could be moved, but then it would be duplicated in several places, would be more of a breaking change, and I don't see that it adds value.

ajcvickers: You can create a non-shadow property with a string name and a null type.

AndriySvyryd: The benefit would be perf as we wouldn't need to get the list of CLR members in any of these overloads.

AndriySvyryd: What I am proposing will allow to force shadowness and I think it could improve the factoring in the internal builders. I'll look into it later.

ajcvickers: Btw, there are places, such as the FK data annotation convention, where we look up MemberInfos but then don't flow them through to here, and hence the MemberInfos get looked up again. This was the case before and is still the case now. But that doesn't mean this should change, it just means that in those places where we have MemberInfos we should be using them if perf is important. The couterpoint to that is that just using names makes that code simpler, so it really does depend on whether there is a perf bottleneck at all. With this API the way it is, the choice of collapsing to names or preserving known MemberInfos is preserved.

AndriySvyryd: Yep, we'll need to go through the code to ensure that MemberInfos are not discarded. Reflection is usually in the bottleneck for model building.

AndriySvyryd:
:shipit:

ajcvickers:
Merged

Issue #6570, opened 2016-09-19
Including an inherited navigation throws IndexOutOfRangeException


AndriySvyryd:
See #6571 for full repro

```C#
var character = context.Characters
    .Include(c => c.Level.Game)
    .First();
```
```
System.IndexOutOfRangeException : Index was outside the bounds of the array.
	ChangeTracking\Internal\StateData.cs(73,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.StateData.FlagProperty(Int32 propertyIndex, PropertyFlag propertyFlag, Boolean isFlagged)
	ChangeTracking\Internal\InternalEntityEntry.cs(1034,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetIsLoaded(INavigation navigation, Boolean loaded)
	Query\Internal\EntityTrackingInfo.cs(218,0): at Microsoft.EntityFrameworkCore.Query.Internal.EntityTrackingInfo.<GetIncludedEntities>d__17.MoveNext()
	at System.Linq.Enumerable.<SelectManyIterator>d__16`2.MoveNext()
	at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
	Query\Internal\QueryBuffer.cs(140,0): at Microsoft.EntityFrameworkCore.Query.Internal.QueryBuffer.StartTracking(Object entity, EntityTrackingInfo entityTrackingInfo)
	Query\QueryContext.cs(136,0): at Microsoft.EntityFrameworkCore.Query.QueryContext.StartTracking(Object entity, EntityTrackingInfo entityTrackingInfo)
	Query\Internal\LinqOperatorProvider.cs(191,0): at Microsoft.EntityFrameworkCore.Query.Internal.LinqOperatorProvider.<_TrackEntities>d__15`2.MoveNext()
	Query\Internal\LinqOperatorProvider.cs(145,0): at Microsoft.EntityFrameworkCore.Query.Internal.LinqOperatorProvider.ExceptionInterceptor`1.EnumeratorExceptionInterceptor.MoveNext()
	at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
	Query\Internal\QueryCompiler.cs(200,0): at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.<>c__DisplayClass19_1`1.<CompileQuery>b__1(QueryContext qc)
[...]

PR #6572, opened 2016-09-19
Correct navigation index calculation

ajcvickers:
Issue #6570

This issue is caused by incorrectly calculating the indexes for navigations. Navigations started counting at the end of the properties, but I'm not sure why that was since all uses appear to be independent. This lack of overlap was then "corrected for" when using the index for the IsLoaded flag, but this correction was done incorrectly when derived types were involved. The fix is to just make navigation indexes count independently from property indexes. Note, however, that relationship indexes are still combined.


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlServerEndToEndTest.cs
```diff
@@ -404,6 +404,128 @@ public void Can_enumerate_entity_set()
         }
 
         [Fact]
+        public void Can_track_an_entity_with_more_than_10_properties()
```
AndriySvyryd: You can remove this test, I will be running it anyway :)

PR #6579, opened 2016-09-21
Query: Properly modify column references while from clause is subquery

### LINKED ISSUE #6270: Using DateTime.Year in complex query fails on SQL Server
### Steps to reproduce
I have the following LINQ query, where the property `AvailableFrom` is a DateTime (not null) property mapped to a SQL `Date` type column.
```csharp
var nextYear = DateTime.UtcNow.AddYears(1).Year;
var years = await _db.Books.Select(x => x.AvailableFrom.Year)
    .OrderByDescending(x => x).Distinct()
    .Where(x => x < nextYear)
    .ToListAsync().ConfigureAwait(false);
```

### The issue
The above query works with `InMemory` database but fails on `UseSqlServer` (see below for exception)

The interesting point is that if I refactor the query, such as removing the `.Where()` clause then it works. I also have to move the `OrderByDescending()` before the selection of the `.Select()`. *I haven't tried many refactorings as it only fails when running with real database, which I don't use in Unit Tests. However I do know that moving the `.Where()` clause before the `.Select()` fails with the same exception.*

### Code that works
```csharp
var nextYear = DateTime.UtcNow.AddYears(1).Year;
var years = await _db.Books.OrderByDescending(x => x.AvailableFrom)
    .Select(x => x.AvailableFrom.Year).Distinct()                      
    .ToListAsync().ConfigureAwait(false);
var result = years.Where(x => x < nextYear)... etc.
```

**NOTE: This problem is not holding me up as I can refactored as shown above.**
 
------
```
Exception message: An unhandled exception occurred while processing the request.

SqlException: Invalid column name 'AvailableFrom'.
The multi-part identifier "x0.AvailableFrom" could not be bound.
<ExecuteDbDataReaderAsync>b__107_0
```
Stack Trace:
```
SqlException: Invalid column name 'AvailableFrom'.
 The multi-part identifier "x0.AvailableFrom" could not be bound.
<ExecuteDbDataReaderAsync>b__107_0
InnerInvoke
Execute
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
MoveNext
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
MoveNext
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
MoveNext
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
MoveNext
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
GetResult
MoveNext in BookFilterDropdowns.cs
                        Value = x.Category,
                        Text = $"{x.Category} ({x.BooksCount})"
                    });
                case BooksFilterBy.ByPublicationYear:
                    var comingSoon = _db.Books.Any(x => x.AvailableFrom > DateTime.UtcNow);
                    var nextYear = DateTime.UtcNow.AddYears(1).Year;
                    var years = await _db.Books.Select(x => x.AvailableFrom.Year)
                        .OrderByDescending(x => x).Distinct()
                        .Where(x => x < nextYear)
                        .ToListAsync().ConfigureAwait(false);
                    var result = years.Select(x => new DropdownTuple {Value = x.ToString(), Text = x.ToString()}).ToList();
                    if (comingSoon)
                        result.Insert(0, new DropdownTuple
                        {
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
GetResult
MoveNext in HomeController.cs
        /// <param name="options"></param>
        /// <returns></returns>
        [HttpGet]
        public async Task<JsonResult> GetFilterSearchContent(ListSortFilterDto options)
        {
            var service = new BookFilterDropdowns(_db);
            return Json(await service.GetFilterDropDownValuesAsync(options.FilterBy));
        }
        public IActionResult About()
        {
            ViewData["Message"] = "Your application description page.";
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
MoveNext
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
MoveNext
MoveNext
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
MoveNext
ThrowForNonSuccess
HandleNonSuccessAndDebuggerNotification
MoveNext
```

### Further technical details

EF Core version: 1.0.0
Operating system: Windows 10
Visual Studio version: VS2015, update 3

Other details about my project setup: Fairly standard ASP.NET Core app (1.0.0) with DataLayer.

smitpatel:
Fixes #6270 
Issue: There are multiple issues here
1. In projection, we have function to update column references while from is subquery but if the inner projection is aliased then it does not use alias. And in the case of specific SQL function, it tries to use it on a column which is not projected in the subquery. Fix is to use the alias if the inner projection has alias, else update column reference.
2. In where clause, when we select whole querysourcerefence, we were trying to use the subquery selector. That works only if the single column is being projected out of the subquery. But in the case if the inner subquery projection is aliased due to Sql function then it will have Sql function with references which may not be projected in subquery. Fix is to use the alias expression if single column is projected out and using alias.
3. In the select expression distinct was applied for the case where the projection is on client therefore incorrect projection was being made distinct, giving wrong result. Fix is to check for client eval of projection before applying distinct.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -973,7 +973,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                     return null;
                 }
 
-                return newLeft != stringCompare.Left 
+                return newLeft != stringCompare.Left
                     || newRight != stringCompare.Right
                         ? new StringCompareExpression(stringCompare.Operator, newLeft, newRight)
                         : expression;
@@ -1043,12 +1043,37 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 }
             }
 
-            var selector
-                = ((expression.ReferencedQuerySource as FromClauseBase)
-                    ?.FromExpression as SubQueryExpression)
-                    ?.QueryModel.SelectClause.Selector;
+            var type = expression.ReferencedQuerySource.ItemType.UnwrapNullableType().UnwrapEnumType();
 
-            return selector != null ? Visit(selector) : null;
+            if (_relationalTypeMapper.FindMapping(type) != null)
+            {
+                var selectExpression = _queryModelVisitor.TryGetQuery(expression.ReferencedQuerySource);
+
+                if (selectExpression != null)
+                {
+                    var subquery = selectExpression.Tables.FirstOrDefault() as SelectExpression;
+
+                    var innerProjectionExpression = subquery?.Projection.FirstOrDefault() as AliasExpression;
+                    if (innerProjectionExpression != null)
+                    {
+                        if (innerProjectionExpression.Alias != null)
+                        {
+                            return new ColumnExpression(
+                                innerProjectionExpression.Alias,
+                                innerProjectionExpression.Type,
+                                subquery);
+                        }
+
+                        var newExpression = selectExpression.UpdateColumnExpression(innerProjectionExpression.Expression, subquery);
+                        return new AliasExpression(newExpression)
+                        {
+                            SourceMember = innerProjectionExpression.SourceMember
+                        };
+                    }
+                }
+            }
+
+            return null;
         }
 
         /// <summary>
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -326,12 +326,21 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 
                         if (aliasExpression != null)
                         {
-                            var expression = UpdateColumnExpression(aliasExpression.Expression, subquery);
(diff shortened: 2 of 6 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -1007,12 +1007,37 @@ protected override Expression VisitQuerySourceReference(QuerySourceReferenceExpr
...
+            var type = expression.ReferencedQuerySource.ItemType.UnwrapNullableType().UnwrapEnumType();
 
-            return selector != null ? Visit(selector) : null;
+            if (_relationalTypeMapper.FindMapping(type) != null)
+            {
+                var selectExpression = _queryModelVisitor.TryGetQuery(expression.ReferencedQuerySource);
+
+                if (selectExpression != null)
+                {
+                    var subquery = (SelectExpression)selectExpression.Tables.Single();
```
anpete: Are these Single calls safe?

smitpatel: Converted to `FirstorDefault()`

smitpatel:
Updated

anpete:
:shipit:

PR #6585, opened 2016-09-21
[1.0.2] Query: Close the connection properly while loading related data

### LINKED ISSUE #6581: Query including a collection doesn't close the connection
This might be by design, but it surprised me so I am filing a bug on it. Running the following simple query results in the connection remaining open after ToList has completed:

``` C#
DbConnection dbConnection;

using (var context = new MyContext())
{
    dbConnection = context.Database.GetDbConnection();

    Console.WriteLine(dbConnection.State);

    context.Parents.Include(u => u.Children).ToList();
    //context.Children.Include(u => u.Parent).ToList();

    Console.WriteLine(dbConnection.State);
}

Console.WriteLine(dbConnection.State);
```

Output:

```
Closed
Open
Closed
```

The connection is closed after the context is disposed. Including a reference does not leave the connection open.

Full repro with model:

``` C#
{
    public DbSet<Parent> Parents { get; set; }
    public DbSet<Child> Children { get; set; }

        => optionsBuilder.UseSqlServer(
        "Data Source=(localdb)\\MSSQLLocalDB;Initial Catalog=inc-test-db; Integrated Security=True;");
}

public class Program
{
    {
        using (var context = new MyContext())
        {

            context.AddRange(
                new Parent { Children = new List<Child> { new Child(), new Child() } },
                new Parent { Children = new List<Child> { new Child(), new Child() } });

            context.SaveChanges();
        }

        DbConnection dbConnection;

        using (var context = new MyContext())
        {
            dbConnection = context.Database.GetDbConnection();

            Console.WriteLine(dbConnection.State);

            context.Parents.Include(u => u.Children).ToList();
            //context.Children.Include(u => u.Parent).ToList();

            Console.WriteLine(dbConnection.State);
        }

        Console.WriteLine(dbConnection.State);
    }
}

public class Parent
{
    public int Id { get; set; }
    public ICollection<Child> Children { get; set; }
}

public class Child
{
    public int Id { get; set; }
    public Parent Parent { get; set; }
}
```

smitpatel:
Resolves #6581 
Issue: When MARS is disabled, we buffer all the data. Which was setting datareader to null. Our dispose method was based on assumption that if datareader is non-null then only we opened connection & we should try to close it.
Fix: We should close the connection once we are done buffering all the data.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/AsyncFromSqlQueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/AsyncFromSqlQueryTestBase.cs
@@ -361,6 +361,25 @@ FROM ""Customers""")
             }
         }
 
+        [Fact]
+        public virtual async Task Include_closed_connection_opened_by_it_when_buffering()
+        {
+            using (var context = CreateContext())
+            {
+                var connection = context.Database.GetDbConnection();
+
+                Assert.Equal(ConnectionState.Closed, connection.State);
+
+                var query = await context.Customers
+                        .Include(v => v.Orders)
+                        .Where(v => v.CustomerID == "ALFKI")
+                        .ToListAsync();
+
+                Assert.NotEmpty(query);
+                Assert.Equal(ConnectionState.Closed, connection.State);
+            }
+        }
+
         protected NorthwindContext CreateContext() => Fixture.CreateContext();
 
         protected AsyncFromSqlQueryTestBase(TFixture fixture)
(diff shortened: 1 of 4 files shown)


ajcvickers:
@divega @rowanmiller Just wanted to point out that if people have written code that does a query and then something else that depends on the connection being open, then this fix would break that code.

rowanmiller:
I'm ok with that being the case, it seems pretty unlikely that anyone is doing that (I guess you'd be dropping down to a raw SQL command after executing the query).

divega:
Agreed.

Issue #1671, opened 2015-02-20
Reverse Engineer: Merge code which does similar things in Migrations


lajones:
Part of #830.

There are a couple of classes which have close equivalents in Migrations (ServiceProvider cf DesignTimeServices; CSharpUtilities cf CSharpHelper; SqlServerTypeMapping cf SqlServerTypeMapper). We should investigate merging the functionality.

rowanmiller:
@bricelam @lajones seems like we have done some of this, anything left to do here?

bricelam:
There's still a lot of duplication in the C# utility classes.

bricelam:
It looks like there is also duplicate code that transforms directory paths into namespaces.

smitpatel:
Also merge code for `GenerateAnnotations` in migration & scaffolding including `GenerateFluentApiForAnnotation`

divega:
Cleared up the milestone to get more clarity on whether this should be punted.

smitpatel:
Also, consider that DbContext writer does not generate namespaces for custom types at present.

bricelam:
Closing since this is tracked by #8853

PR #6586, opened 2016-09-21
Prettier type names for scaffolded primitive arrays

roji:


bricelam:
Can we do this in a more general way like in [this code](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Design/Migrations/Design/CSharpHelper.cs#L193-L208)?

bricelam:
See also #1671 😉

roji:
Sure, I can generate the entries for the array types from the primitives, is it worth the trouble though? It would also require using `Type.MakeArrayType()`, does that play well with .NET Native etc.?

bricelam:
@roji I think you misunderstood my comment. I'm recommending changing [CSharpUtilities.GetTypeName](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational.Design/Internal/CSharpUtilities.cs#L364) instead to add something like this:

```C#
if (type.IsArray)
{
    return GetTypeName(type.GetElementType()) + "[]";
}
```

bricelam:
(That way, we don't have to add values to `_primitiveTypeNames`)

roji:
Oops, that was pretty silly of me...

Amended this PR with your suggestion.

PR #6590, opened 2016-09-22
Query: Fix #4588 - Query :: SelectMany-GroupJoin-DefaultIfEmpty isn't…

### LINKED ISSUE #4588: Query :: SelectMany-GroupJoin-DefaultIfEmpty isn't being lifted into SQL for some complex queries, resulting in extensive client-side evaluation
This is mostly visible in scenarios with optional navigations, e.g.

```
from c in ctx.Customers
where c.Detail.Name == "Foo"
select c;
```

this gets translated into:

```
from c in ctx.Customers
join d in ctx.Details on c.Id equals d.CustomerId into grouping
from d in grouping.DefaultIfEmpty()
where d.Name == "Foo"
select c;
```

however the filter will be performed on the client resulting in unnecessary data being pulled from the database

anpete:
… being lifted into SQL for some complex queries, resulting in extensive client-side evaluation.

Adds relational GroupJoin/DefaultIfEmpty elimination.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -212,7 +212,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                     var leftExpression = Visit(expression.Left);
                     var rightExpression = Visit(expression.Right);
 
-                    return leftExpression != null && rightExpression != null
+                    return leftExpression != null 
+                            && rightExpression != null
                         ? Expression.MakeBinary(
                             expression.NodeType,
                             leftExpression,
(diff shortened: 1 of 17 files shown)


anpete:
@maumar @smitpatel

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -1007,6 +1021,28 @@ protected override Expression VisitQuerySourceReference(QuerySourceReferenceExpr
...
+                {
+                    var entityType
+                        = _queryModelVisitor.QueryCompilationContext.Model
+                            .FindEntityType(joinClause.ItemType);
+
+                    if (entityType != null)
+                    {
+                        return Visit(
+                            EntityQueryModelVisitor.CreatePropertyExpression(
+                                expression, entityType.FindPrimaryKey().Properties[0]));
```
smitpatel: Why PK & only first property

smitpatel:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
anpete: This is the case when a bare QSRE is encountered E.g. Where(c => c != null && c.CustomerID == "ALFKI")

We only need to test a single PK column because none can be null.

PR #6596, opened 2016-09-23
Add connection resiliency support for Sql Server.

AndriySvyryd:
All high-level API that need database connection will use the configured execution strategy except migrations.
Handle external connection state changes in RelationalConnection

Handle more objects in SqlServerDatabaseCleaner
Functional tests are no longer executed in a fixture-wide transaction, the database is cleaned when a fixture is created. Tests that modify a shared database do so in a local transaction
All SQL Server functional tests can be run against SQL Azure by providing the connection string in config.json, database will be created using Basic service tier or in an elastic pool if specified.

Add to .Core:
  * IExecutionStrategy
  * ExecutionStrategy
  * IExecutionStrategyFactory
  * IDatabaseProviderServices
    * IExecutionStrategyFactory ExecutionStrategyFactory
  * DatabaseFacade
    * IExecutionStrategy CreateExecutionStrategy()
    * TResult ExecuteWithStrategyInTransaction<TResult>(Func<DbContext, TResult> operation)
    * Task<TResult> ExecuteWithStrategyInTransactionAsync<TResult>(Func<DbContext, CancellationToken, Task<TResult>> operation, CancellationToken cancellationToken)

Add to .Relational:
  * RelationalDbContextOptionsBuilder
    * ExecutionStrategy(Func<ExecutionStrategyContext, IExecutionStrategy> getExecutionStrategy)

Add to .SqlServer:
  * SqlAzureExecutionStrategy
  * SqlServerDbContextOptionsBuilder
    * EnableRetryOnFailure()
    * EnableRetryOnFailure(int maxRetryCount)


AndriySvyryd:
CRR:
  @anpete Query
  @ajcvickers @divega @rowanmiller API

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
```diff
@@ -0,0 +1,339 @@
...
+        /// </summary>
+        protected virtual ILogger<IExecutionStrategy> Logger { get; }
+
+#if NET451
+        private const string ContextName = "ExecutionStrategySuspended";
+
+        /// <summary>
+        ///     Indicates whether the strategy performs any additional actions when executing an operation
+        /// </summary>
+        protected static bool Suspended
```
ajcvickers: >Suspended [](start = 30, length = 9)

Can you explain a bit how this is used and how the documentation related to the property name?

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
```diff
@@ -0,0 +1,339 @@
...
+            TState state,
+            CancellationToken cancellationToken = new CancellationToken())
+        {
+            if (Suspended)
+            {
+                return operation(state, cancellationToken);
+            }
+
+            return ExecuteAsyncImplementation(operation, state, cancellationToken);
+        }
```
ajcvickers: Consider conditional operator.

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
AndriySvyryd: Normally it should be only used by the strategy itself. It is set to true when a strategy starts an execution, so if there are any nested calls to the strategy inside the operation being executed the nested operation will be just executed, without any exception handling or retries.

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
```diff
@@ -0,0 +1,339 @@
...
+        ///     Determines whether the operation should be retried and the delay before the next attempt.
+        /// </summary>
+        /// <param name="lastException">The exception thrown during the last execution attempt.</param>
+        /// <returns>
+        ///     Returns the delay indicating how long to wait for before the next execution attempt if the operation should be retried;
+        ///     <c>null</c> otherwise
+        /// </returns>
+        protected internal virtual TimeSpan? GetNextDelay([NotNull] Exception lastException)
+        {
+            ExceptionsEncountered.Add(lastException);
```
ajcvickers: >ExceptionsEncountered.Add(lastException); [](start = 12, length = 41)

It seems a little strange that a method called GetNextDelay would add the exception to the encountered list.

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
```diff
@@ -0,0 +1,339 @@
...
+        public static T UnwrapAndHandleException<T>([NotNull] Exception exception, [NotNull] Func<Exception, T> exceptionHandler)
+        {
+            var dbUpdateException = exception as DbUpdateException;
+            if (dbUpdateException != null)
+            {
+                return UnwrapAndHandleException(dbUpdateException.InnerException, exceptionHandler);
+            }
+
+            return exceptionHandler(exception);
+        }
```
ajcvickers: Consider conditional operator.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -651,4 +651,10 @@
   <data name="InheritedPropertyCannotBeIgnored" xml:space="preserve">
     <value>The property '{property}' cannot be ignored on entity type '{entityType}', because it's declared on the base entity type '{baseEntityType}'. To exclude this property from your model, use NotMappedAttribute or Ignore method on the base type.</value>
   </data>
+  <data name="RetryLimitExceeded" xml:space="preserve">
+    <value>Maximum number of retries ({retryLimit}) exceeded while executing database operations with '{strategy}'. See inner exception for the most recent failure.</value>
+  </data>
+  <data name="ExecutionStrategyExistingTransaction" xml:space="preserve">
+    <value>The configured execution strategy '{strategy}' does not support user initiated transactions. Use the execution strategy to execute the all the operations in the transaction as a retriable unit.</value>
+  </data>
```
ajcvickers: Maybe point to the method or some other information on how to actually do this?

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
AndriySvyryd: Changed

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
AndriySvyryd: Moved out

src/Microsoft.EntityFrameworkCore/Infrastructure/DatabaseFacade.cs
```diff
@@ -135,6 +136,60 @@ public virtual void RollbackTransaction()
...
+
+        /// <summary>
+        ///     Executes the operation using the configured <see cref="IExecutionStrategy"/> inside a transaction.
+        /// </summary>
+        /// <param name="operation">
+        ///     A delegate representing an executable operation that returns the result of type <typeparamref name="TResult" />.
+        /// </param>
+        /// <typeparam name="TResult">The return type of <paramref name="operation" />.</typeparam>
+        /// <returns>The result from the operation.</returns>
+        public virtual TResult ExecuteWithStrategyInTransaction<TResult>([NotNull] Func<DbContext, TResult> operation)
```
ajcvickers: >Func<DbContext, TResult> operation [](start = 83, length = 34)

It is somewhat unfortunate that this is a Func of DbContext when the application will likely have some derived context type with DbSets on it.

src/Microsoft.EntityFrameworkCore/Infrastructure/DatabaseFacade.cs
```diff
@@ -135,6 +136,60 @@ public virtual void RollbackTransaction()
...
+        ///     Asynchronously executes the operation using the configured <see cref="IExecutionStrategy"/> inside a transaction.
+        /// </summary>
+        /// <param name="operation">
+        ///     A delegate representing an executable operation that returns the result of type <typeparamref name="TResult" />.
+        /// </param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken" /> to observe while waiting for the task to complete.</param>
+        /// <typeparam name="TResult">The return type of <paramref name="operation" />.</typeparam>
+        /// <returns>The result from the operation.</returns>
+        public virtual Task<TResult> ExecuteWithStrategyInTransactionAsync<TResult>(
+            [NotNull] Func<DbContext, CancellationToken, Task<TResult>> operation, CancellationToken cancellationToken)
```
ajcvickers: >CancellationToken cancellationToken [](start = 83, length = 35)

Why no default value for the cancellation token?

src/Microsoft.EntityFrameworkCore/Infrastructure/DatabaseFacade.cs
AndriySvyryd: Yes, that's why I added CreateExecutionStrategy(), we could remove ExecuteWithStrategyInTransaction if nobody objects ( @divega @rowanmiller )

src/Microsoft.EntityFrameworkCore/Infrastructure/DatabaseFacade.cs
AndriySvyryd: Added

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -116,11 +116,13 @@ public static IEnumerable<string> GetPendingMigrations([NotNull] this DatabaseFa
                     .GetService<IRawSqlCommandBuilder>()
                     .Build(sql, parameters);
 
-                return rawSqlCommand
-                    .RelationalCommand
-                    .ExecuteNonQuery(
-                        GetRelationalConnection(databaseFacade),
-                        parameterValues: rawSqlCommand.ParameterValues);
+                return databaseFacade.CreateExecutionStrategy().Execute(database =>
```
ajcvickers: >databaseFacade.CreateExecutionStrategy().Execute(database => [](start = 23, length = 60)

Remind me how we know that the command is idempotent or otherwise handle arbitrary commands that may or may not have side effects?

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalExecutionStrategyFactory.cs
```diff
@@ -0,0 +1,56 @@
...
+
+using System;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.EntityFrameworkCore.Storage.Internal
+{
+    public class RelationalExecutionStrategyFactory : IExecutionStrategyFactory
```
ajcvickers: >RelationalExecutionStrategyFactory [](start = 17, length = 34)

What makes this relational?

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
ajcvickers: That makes sense, but for some reason I am having trouble reading "Indicates whether the strategy performs any additional actions when executing an operation" to mean that.

src/Microsoft.EntityFrameworkCore/Infrastructure/DatabaseFacade.cs
ajcvickers: Could we make it generic and do the cast for them?

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalExecutionStrategyFactory.cs
AndriySvyryd: It reads `RelationalOptionsExtension`

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
AndriySvyryd: Any suggestions on how could it be phrased better?

src/Microsoft.EntityFrameworkCore/Infrastructure/DatabaseFacade.cs
AndriySvyryd: They would still need to specify the type in the call

src/Microsoft.EntityFrameworkCore.Specification.Tests/FindTestBase.cs
```diff
@@ -622,47 +617,47 @@ public virtual void Throws_for_bad_entity_type()
             }
         }
 
-        protected class BaseType
+        public class BaseType
```
ajcvickers: >  public class BaseType [](start = 6, length = 23)

Why are these all public? They were specifically made protected so that they don't pollute the space of public types visible to all tests, which makes Intellisense and automated tooling easier to use.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
AndriySvyryd: We don't. A safer approach would be to leave this method as is and add an overload where we start a transaction and use the execution strategy.

ajcvickers:
>             ExecuteWithStrategyInTransaction(

Is it true that in order to be correct every test now needs to use ExecuteWithStrategyInTransaction? If so, that seems like a significant reduction in the usability of the test infrastructure.

---
Refers to: src/Microsoft.EntityFrameworkCore.Specification.Tests/GraphUpdatesTestBase.cs:2817 in f66adea. [](commit_id = f66adea95c9519ce418731b7e0a71aa56bf40421, deletion_comment = False)

src/Microsoft.EntityFrameworkCore.Specification.Tests/FindTestBase.cs
AndriySvyryd: FindFixtureBase needs to be public because it's used as a generic constraint so FindContext needs to be public because it's returned from a public method so all these entity types need to be public for DbSets

AndriySvyryd: I can make them protected again by introducing casts in two places

src/Microsoft.EntityFrameworkCore.SqlServer/SqlAzureExecutionStrategy.cs
```diff
@@ -0,0 +1,158 @@
...
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.EntityFrameworkCore.Storage.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.EntityFrameworkCore
+{
+    public class SqlAzureExecutionStrategy : ExecutionStrategy
```
ajcvickers: >SqlAzureExecutionStrategy [](start = 17, length = 25)

Given that the normal way to get this is to use "EnableRetryOnFailure" which doesn't mention SQL Azure, should this really be called the SqlAzureExecutionStrategy?

AndriySvyryd:
>             ExecuteWithStrategyInTransaction(

Only the tests that make changes to a shared database that need to be reverted

---
In reply to: [249699255](https://github.com/aspnet/EntityFramework/pull/6596#issuecomment-249699255) [](ancestors = 249699255)

---
Refers to: src/Microsoft.EntityFrameworkCore.Specification.Tests/GraphUpdatesTestBase.cs:2817 in f66adea. [](commit_id = f66adea95c9519ce418731b7e0a71aa56bf40421, deletion_comment = False)

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
ajcvickers: That second option seems better to me. Would be curious what others think.

src/Microsoft.EntityFrameworkCore.SqlServer/SqlAzureExecutionStrategy.cs
AndriySvyryd: How about 'SqlServerTransientErrorHandlingExecutionStrategy'?

src/Microsoft.EntityFrameworkCore.Specification.Tests/FindTestBase.cs
ajcvickers: Would prefer them to be protected. In general I don't like the patterns we settled on that require a fixture and use of the TestDatabase and all that stuff. I came to peace with this by using a pattern that that keeps all that stuff isolated to the tests with as few public types as possible introduced just for infrastructure stuff or stuff that is only used by those particular tests.

ajcvickers:
The product code changes look good to me, but I don't like the test changes. Are the test changes strictly needed for the product changes? If not, could we separate out these two things. Maybe it is time to revisit the patterns we use for tests again?

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
ajcvickers: Indicates whether the strategy is suspended. The strategy is typically suspending why executing to avoid recursive execution from nested operations.

src/Microsoft.EntityFrameworkCore.SqlServer/SqlAzureExecutionStrategy.cs
ajcvickers: Sounds good.

AndriySvyryd:
The test changes are not strictly needed, since we don't test against SQL Azure in the main CIs, I would be OK checking them in later.

src/Microsoft.EntityFrameworkCore/Storage/ExecutionStrategy.cs
AndriySvyryd: Sounds better, though now it doesn't explicitly say how it behaves when suspended

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
AndriySvyryd: Ok

ryanelian:
I've noticed that this feature prevents you from using transaction. Is there a way to disable it temporarily?

AndriySvyryd:
@ryanelian You can execute your code inside `context.Database.CreateExecutionStrategy().Execute()` and create the transaction manually or use `context.Database.CreateExecutionStrategy().ExecuteInTransaction()`

Tieantono:
Hi @AndriySvyryd, thanks for your info  about using transaction with connection resiliency feature. I've successfully created a process using the `context.Database.CreateExecutionStrategy().Execute()` like this:
image
Now, I want to know how to use the `context.Database.CreateExecutionStrategy().ExecuteInTransaction()` method, may I have your example for how to use the `context.Database.CreateExecutionStrategy().ExecuteInTransaction()` method? Because I can't find any example for that method. Thanks in advance.

AndriySvyryd:
@Tieantono `ExecuteInTransaction` is mostly syntactic sugar, so you can do the same as you code snippet with:
```C#
context.Employee.Add(new Employee{ FullName = "Jones" });
context.Database.CreateExecutionStrategy().ExecuteInTransaction(
    c => c.SaveChanges(),
    c => c.Employee.AsNoTracking().Any(e => e.FullName == "Jones"),
    context,
    context);
```
The second argument in this method allows to verify that the operation was successful in case there is a failure while the transaction is being committed.
Note that you should add the `Employee` before `Execute` otherwise it will add another copy on each retry.

Tieantono:
@AndriySvyryd Thanks a lot for your explanation! I think those methods should have been documented on Microsoft Docs so people can easily find the references for those methods.

PR #6607, opened 2016-09-26
[1.0.2] Query: Generate Inner Join for collection include while Skip/Take

### LINKED ISSUE #6530: LINQ with Take generates incorrect SQL
### Steps to reproduce

If I run a query like 

```
 varobj = context.Lightbox.Take(50)
             .Include(l => l.LightboxImages)
            .ThenInclude(i => i.Image)
```

which is to say, take the top 50 lightboxes, including the images that have a many-to-many relationship with lightboxes.
### The issue

This frequently times out. It turns out that this is because it is generating incorrect SQL (as seen in the .NET Core debug output). The SQL it generates (with the columns pulled out for brevity) is

```
SELECT TOP(50) *
FROM [tblLightbox] AS [l]
ORDER BY [l].[LightboxId]

SELECT *
FROM [tblLightboxImages] AS [t]
LEFT JOIN [tblLBDImages] AS [t0] ON [t].[ImageId] = [t0].[ImageId]
WHERE EXISTS (
    SELECT TOP(50) 1
    FROM [tblLightbox] AS [l]
    WHERE [t].[LightboxId] = [l].[LightboxId])
ORDER BY [t].[LightboxId]
```

The `where exists` clause is incorrect. This will return every image that has a many-to-many with at least one lightbox. The SQL should look more like this: 

```
SELECT *
FROM [tblLightboxImages] AS [t]
LEFT JOIN [tblLBDImages] AS [t0] ON [t].[ImageId] = [t0].[ImageId]
WHERE [t].[LightboxId] in (
    SELECT TOP(50) [l].[LightboxId]
    FROM [tblLightbox] AS [l])
```

This will return only the images that have a lightbox ID in those top 50 lightboxes. It maps to entities properly, but with an image table with a large number of rows this query is too slow and times out.
### Further technical details

EF Core version: 1.0.0
Operating system:  WIndows 10 Anniversary, 
Visual Studio version: (e.g. VS 2013 or n/a) VS 2015

SQL Server 2016

smitpatel:
Fixes #6530 
Issue: During Include with skip/take, first query select only fixed amount of rows. Second query should join related data with only those rows. But using `WHERE EXISTS` will give matching rows for all related data. With use of Skip, if there are not sufficient matching rows then where exists can return false. With Take it will return true for any matching rows. This would give incorrect result (reported in #5901 & fixed in #6002) or fetch whole table (#6530)
Solution is to disable `Where Exists` while using skip/take.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
@@ -288,8 +288,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                             : selectExpression.Tables.Last(t => t.QuerySource == querySource);
 
                     var canGenerateExists
-                        = (selectExpression.Predicate != null
-                          || selectExpression.Offset == null)
+                        = selectExpression.Offset == null
+                          && selectExpression.Limit == null
                           && !IsOrderingOnNonPrincipalKeyProperties(
                               selectExpression.OrderBy,
                               navigation.ForeignKey.PrincipalKey.Properties);
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/IncludeTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/IncludeTestBase.cs
@@ -194,6 +194,22 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
+        public virtual void Include_collection_take_no_order_by()
+        {
+            using (var context = CreateContext())
+            {
+                var customers
+                        = context.Set<Customer>()
+                            .Take(10)
+                            .Include(c => c.Orders)
+                            .ToList();
+
+                Assert.Equal(10, customers.Count);
+                Assert.True(customers.All(c => c.Orders != null));
+            }
+        }
+
+        [Fact]
         public virtual void Include_collection_skip_take_no_order_by()
         {
             using (var context = CreateContext())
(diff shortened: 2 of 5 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -288,8 +288,8 @@ var principalTable
                             : selectExpression.Tables.Last(t => t.QuerySource == querySource);
 
                     var canGenerateExists
-                        = (selectExpression.Predicate != null
-                          || selectExpression.Offset == null)
+                        = selectExpression.Offset == null
```
maumar: add: select.Predicate != null

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -475,9 +475,9 @@ var needOrderingChanges
                     }
                 }
 
-                var index = orderingExpression is SelectExpression
-                    ? innerJoinSelectExpression.AddAliasToProjection(innerJoinSelectExpression.Alias + "_" + innerJoinSelectExpression.Projection.Count, orderingExpression)
-                    : innerJoinSelectExpression.AddToProjection(orderingExpression);
+                var index = aliasExpression != null
```
maumar: this is unnecessary

maumar:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
smitpatel: That removal was intended.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
smitpatel: updated

Issue #6429, opened 2016-08-29
Query :: NRE protection is not applied to manually created GroupJoin when trying to extract keys from collections potentially containing nulls


maumar:
Example of a query that could throw NRE:

```cs
context.LevelOne
                    .GroupJoin(
                        context.LevelOne.Where(l1 => l1.Name != "L1 01").Select(l1 => l1.OneToOne_Required_FK),
                        l1 => l1.Id,
                        l2 => l2.Level1_Optional_Id,
                        (l1, l2s) => new { l1, l2s })
                    .Where(r => r.l2s.Any())
                    .Select(r => r.l1);
```

this throws if one (or more) inner element is null - we don't apply null protection logic to the key extraction, so we try to access Level1_Optional_Id of a null-valued Level2 item.

We already do there correct thing if GroupJoin is created by us (i.e. when we rewrite optional navigations), but we should do it for all cases. Current workaround is to simply add null protection logic yourself, in a form of l2 != null ? l2.Level1_Optional_Id : null

maumar:
fixed in 9897064a34a0c815187f86a506f34af0cf4d5c9d

PR #6610, opened 2016-09-26
Fix to #6429 - NRE protection is not applied to manually created GroupJoin when trying to extract keys from collections potentially containing nulls

### LINKED ISSUE #6609: Query :: we are adding redundant navigation joins to query in some cases, resulting in correct but unnecessarily complex queries
using GearsOfWar model:

```
                var tags = new List<CogTag>();

                var gears = context.Gears
                    .Where(g => g.Tag != null && tags.Contains(g.Tag.Id))
                    .ToList();
```

produces the following query model (after navigation expansion)

```
from Gear g in Gears 
join CogTag g.Tag in CogTags
on 
new CompositeKey(new [] {Convert(Property([g], "Nickname")), Convert(Property([g], "SquadId"))}) 
equals 
new CompositeKey(new [] {Convert(Property([g.Tag], "GearNickName")), Convert(Property([g.Tag], "GearSquadId"))}) 
into IEnumerable`1 g.Tag_group 
from CogTag g.Tag in {[g.Tag_group] => DefaultIfEmpty()} 
join CogTag g.Tag in CogTags) 
on new CompositeKey(new [] {Convert(Property([g], "Nickname")), Convert(Property([g], "SquadId"))}) 
equals 
new CompositeKey(new [] {Convert(Property([g.Tag], "GearNickName")), Convert(Property([g.Tag], "GearSquadId"))}) 
into IEnumerable`1 g.Tag_group 
from CogTag g.Tag in {[g.Tag_group] => DefaultIfEmpty()} 
where (([g.Tag] != null) AndAlso {__tags_0 => Contains([g?.Tag]?.Id)}) 
select [g]
```

Basically we join CogTag twice even though it's the same reference. Problem is that second time CogTag is used inside a subquery and we are not smart enough to associate those two together. This causes additional join and therefore (currently) additional query issued to the database:

```
SELECT [g.Tag1].[Id], [g.Tag1].[GearNickName], [g.Tag1].[GearSquadId], [g.Tag1].[Note]
FROM [CogTag] AS [g.Tag1]

SELECT [g].[Nickname], [g].[SquadId], [g].[AssignedCityName], [g].[CityOrBirthName], [g].[Discriminator], [g].[FullName], [g].[HasSoulPatch], [g].[LeaderNickname], [g].[LeaderSquadId], [g].[Rank], [g.Tag].[Id], [g.Tag].[GearNickName], [g.Tag].[GearSquadId], [g.Tag].[Note]
FROM [Gear] AS [g]
LEFT JOIN [CogTag] AS [g.Tag] ON ([g].[Nickname] = [g.Tag].[GearNickName]) AND ([g].[SquadId] = [g.Tag].[GearSquadId])
WHERE [g].[Discriminator] IN (N'Officer', N'Gear') AND [g.Tag].[Id] IS NOT NULL
ORDER BY [g].[Nickname], [g].[SquadId]
```

maumar:
Problem was that for some groupjoin queries, e.g.:
```c#
var query = context.LevelOne
    .GroupJoin(
        context.LevelOne.Select(l1 => l1.OneToOne_Required_FK),
        l1 => l1.Id,
        l2 => EF.Property<int?>(l2, "Level1_Optional_Id"),
        (l1, l2s) => new { l1, l2s })
    .Select(r => r.l1)
```
In this case inner collection selector can potentially return null and therefore when translating inner key selector we should apply null protection logic. We were not doing that however, because those two operations were done in separate places, by two different instances of NavigationRewritingExpressionVisitor. Fix is to introduce additional state to the visitor, when visiting inner key selector, and then using that information when processing inner key selector.

Moreover now we are reusing the NavigationRewritingExpressionVisitor simplifying the code somewhat and allowing us to easily share state between different nav rewrite operations.

This also fixes #6609 - Query :: we are adding redundant navigation joins to query in some cases, resulting in correct but unnecessarily complex queries

Since now we are better at recognizing the same navigation joins (as they are processed by the same visitor)

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
@@ -1092,6 +1092,39 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             }
         }
 
+        [ConditionalFact]
+        public virtual void Join_navigation_translated_to_subquery_deeply_nested_required()
+        {
+            List<Level1> levelOnes;
+            List<Level4> levelFours;
+            using (var context = CreateContext())
+            {
+                levelFours = context.LevelFour.Include(e => e.OneToOne_Required_FK_Inverse.OneToOne_Required_FK_Inverse.OneToOne_Required_PK_Inverse).ToList();
+                levelOnes = context.LevelOne.ToList();
+            }
+
+            ClearLog();
+
+            using (var context = CreateContext())
+            {
+                var query = from e1 in context.LevelOne
+                            join e4 in context.LevelFour on e1.Name equals e4.OneToOne_Required_FK_Inverse.OneToOne_Required_FK_Inverse.OneToOne_Required_PK_Inverse.Name
+                            select new { Id4 = e4.Id, Name4 = e4.Name, Id1 = e1.Id, Name1 = e1.Name };
+
+                var result = query.ToList();
+
+                var expected = (from e1 in levelOnes
+                                join e4 in levelFours on e1.Name equals e4.OneToOne_Required_FK_Inverse.OneToOne_Required_FK_Inverse.OneToOne_Required_PK_Inverse.Name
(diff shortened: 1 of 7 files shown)


maumar:
@anpete @smitpatel

anpete:
:shipit:

PR #6612, opened 2016-09-27
[1.0.2] Query: Remove Sql-Server specific projection modifier for search conditions

### LINKED ISSUE #6598: Select with Ternary Operator/CASE-WHEN Server Side Evaluation
I wanted to select a few columns of a table/class into a smaller object. I used the ternary operator to get the value for one field. With 1.0.1 this produces a wrong SQL query. 

```
var result = dbContext.Foos
    .OrderBy(f => f.Name)
    .ThenBy(f => f.Id)
    .Select(f => new CodeNameClass()
    {
        Id = f.Id,
        Name = f.Name,
        Code = !string.IsNullOrEmpty(f.Code) ? f.Code : f.Id.ToString()
    })
    .AsNoTracking()
    .ToList();
```

The generated SQL is:

```
SELECT [f].[Id], [f].[Name], CASE
    WHEN CASE
        WHEN [f].[Code] IS NULL OR (([f].[Code] = N'') AND [f].[Code] IS NOT NULL)
    END
    THEN CAST(0 AS BIT) ELSE CAST(1 AS BIT)
END, [f].[Code]
FROM [Foos] AS [f]
ORDER BY [f].[Name], [f].[Id]
```

The code worked with RC2 and 1.0. As I saw from my logs the ternary operator had been evaluated on the client side till 1.0.
### Steps to reproduce

```
public class Foo
{
    public int Id { get; set; }
    public string Code { get; set; }
    // some further fields
}

public class ApplicationDbContext : DbContext
{
    public DbSet<Foo> Foos { get; set; }

    {
        optionsBuilder.UseSqlServer(@"server=(localdb)\MsSqlLocalDB;database=SampleDB");
    }
}

public class CodeNameClass
{
    public int Id { get; set; }
    public string Code { get; set; }
}

public class TestEfCoreServerEvaluation
{
    public TestEfCoreServerEvaluation()
    {
        using (var dbContext = new ApplicationDbContext())
        {
            Console.WriteLine("EnsureDeleted ...");
            dbContext.Database.EnsureDeleted();
            Console.WriteLine("Done!");
            Console.WriteLine("EnsureCreated ...");
            dbContext.Database.EnsureCreated();
            Console.WriteLine("Done!");

            var result = dbContext.Foos
                .OrderBy(f => f.Name)
                .ThenBy(f => f.Id)
                .Select(f => new CodeNameClass()
                {
                    Id = f.Id,
                    Name = f.Name,
                    Code = !string.IsNullOrEmpty(f.Code) ? f.Code : f.Id.ToString()
                })
                .AsNoTracking()
                .ToList();

            //var result = dbContext.Foos
            //  .OrderBy(f => f.Name)
            //  .ThenBy(f => f.Id)
            //  .Select(f => new CodeNameClass()
            //  {
            //      Id = f.Id,
            //      Name = f.Name,
            //      Code = f.Code == null ? f.Code : f.Id.ToString()
            //  })
            //  .AsNoTracking()
            //  .ToList();

            Console.WriteLine("");
        }
    }
}
```

The second query does not use `IsNullOrEmpty`. The generated SQL is:

```
SELECT [f].[Id], [f].[Name], [f].[Code] IS NULL, [f].[Code]
FROM [Foos] AS [f]
ORDER BY [f].[Name], [f].[Id]
```

```
Exception message:
An unhandled exception of type 'System.Data.SqlClient.SqlException' occurred in Microsoft.EntityFrameworkCore.dll

Additional information: Incorrect syntax near the keyword 'IS'.
```
### Further technical details

EF Core version: 1.0.1
Operating system: Windows 10
Visual Studio version:  VS 2015

smitpatel:
Resolves #6598 
We have `ProjectionComparisonTransformingVisitor` class in Sql-Server specific code.
https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.SqlServer/Query/Sql/Internal/SqlServerQuerySqlGenerator.cs#L153
Which somewhat does the same thing as `SearchConditionTranslatingVisitor`. In combination, both tried to make a case block hence we got into state of incorrect nested case blocks.
Sql-Server specific code should be removed and relational one should be used. The most of the changes needed in `SearchConditionTranslatingVisitor` are already done in 1.1.0 codebase.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Query/Sql/Internal/SqlServerQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Query/Sql/Internal/SqlServerQuerySqlGenerator.cs
@@ -97,13 +97,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        protected override void VisitProjection(IReadOnlyList<Expression> projections)
-        {
-            var comparisonTransformer = new ProjectionComparisonTransformingVisitor();
-            var transformedProjections = projections.Select(comparisonTransformer.Visit).ToList();
-
-            base.VisitProjection(transformedProjections);
-        }
+        protected override void VisitProjection(IReadOnlyList<Expression> projections) => base.VisitProjection(projections);
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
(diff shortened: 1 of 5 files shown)


src/Microsoft.EntityFrameworkCore.SqlServer/Query/Sql/Internal/SqlServerQuerySqlGenerator.cs
```diff
@@ -97,13 +97,7 @@ protected override void GenerateLimitOffset(SelectExpression selectExpression)
...
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        protected override void VisitProjection(IReadOnlyList<Expression> projections)
-        {
-            var comparisonTransformer = new ProjectionComparisonTransformingVisitor();
-            var transformedProjections = projections.Select(comparisonTransformer.Visit).ToList();
-
-            base.VisitProjection(transformedProjections);
-        }
+        protected override void VisitProjection(IReadOnlyList<Expression> projections) => base.VisitProjection(projections);
```
smitpatel: What to do with this method? Since it is `protected` in a public class.

anpete:
@smitpatel Please add a small description of the change, thanks!

smitpatel:
Added.

anpete:
:shipit:

PR #6619, opened 2016-09-27
Query: Fix #6458 - CompiledQueryCache use a static lock that will mak…

### LINKED ISSUE #6458: CompiledQueryCache use a static lock that will make the implementation single threaded
[CompiledQueryCache](https://github.com/aspnet/EntityFramework/blob/e60330ef54e0ab02eb4010f91b252e2ed67177bd/src/Microsoft.EntityFrameworkCore/Query/Internal/CompiledQueryCache.cs#L45) use use a static common lock that make the GetOrAddQuery method to be single threaded.

This is a problem together with #6457 that more or less every time need to compile the query.

If we check the profiling we can see that the lock have bad impact in the overall performance.
image

The lock should either be per cachekey or an try-get should exists outsite the lock to bypass and be able to fetch already compiled queries without need to wait for the lock that other thread have created.

anpete:
- Removed static lock and added logic to ensure a single compilation per cache-key.


anpete:
cc @ajcvickers

ajcvickers:
:shipit:

Issue #6597, opened 2016-09-24
"Operation is not valid due to the current state of the object" in some queries with navigation and paging


MohammadAkbari:
I have simple Forum application that each topic has many post, and each post may liked by several users. application Entity Models are as following:

    public class ForumTopic
    {
        public int ForumTopicId { get; set; }
        public string Subject { get; set; }
        public string UserId { get; set; }
        public virtual User User { get; set; }
        public virtual ICollection<ForumPost> ForumPosts { get; set; }
    }

    public class ForumPost
    {
        public int ForumPostId { get; set; }
        public string Message { get; set; }
        public int ForumTopicId { get; set; }
        public virtual ForumTopic ForumTopic { get; set; }
        public string UserId { get; set; }
        public virtual User User { get; set; }
        public virtual ICollection<ForumPostFavorite> ForumPostFavorites { get; set; }
    }

    public class ForumPostFavorite
    {
        public string UserId { get; set; }
        public User User { get; set; }
        public int ForumPostId { get; set; }
        public ForumPost ForumPost { get; set; }
    }

I want take current user favorite posts as following. but this query throws "Operation is not valid due to the current state of the object" Exception. I am sure that this query run in entity framework 6 without exception, but in Entity Framework core this exception occurs.

    public IActionResult Favorite()
    {
      var userId = _userManager.GetUserId(User);

      var list =  _dbContext.ForumPostFavorites
			.Where(e => e.UserId == userId)
			.Select(e => e.ForumPost)
			.Take(10)
			.Select(e => new ForumPostFavoriteVm
			{
				ForumPostId = e.ForumPostId,
				ForumTopicId = e.ForumTopicId,
				Message = e.Message,
				Subject = e.ForumTopic.Subject,
			}).ToList();

      return View(list);
    }

When I remove this line "Subject = e.ForumTopic.Subject" from query, the query runs without any exceptions.

maumar:
@MohammadAkbari could you also provide stack trace of the exception?

andreyshv:
I have a similar situation. If I remove `PostId` in `Blog` class I'll get

Unhandled Exception: System.InvalidCastException: Unable to cast object of type 'System.Int32' to type 'ConsoleApplication.Blog'
in queue2. But if remove `Take(5)` it works fine again.

`   public class Post
    {
        public int Id { get; set; }
        public string Text { get; set; }
    }
    public class Blog
    {
        public int Id { get; set; }
        public string Header { get; set; }

[...]

PR #6621, opened 2016-09-27
[SMALL] Fix to #6597 - "Operation is not valid due to the current state of the object" in some queries with navigation and paging

maumar:
Problem was during join flattening - we can't flatten joins whose input elements are not directly ShapedQueries (e.g. when shaped query is wrapped in select). For those cases we should delegate to client eval.
We already do that for GroupJoin, but it can also happen for regular joins.


maumar:
@anpete @smitpatel

anpete:
:shipit:

PR #6625, opened 2016-09-27
All SQL Server functional tests can be run against SQL Azure …

AndriySvyryd:
... by providing the connection string in config.json, database will be created using Basic service tier or in an elastic pool if specified.

Handle more objects in SqlServerDatabaseCleaner
Functional tests are no longer executed in a fixture-wide transaction, the database is cleaned when a fixture is created. Tests that modify a shared database do so in a local transaction
Use SqlServerTestStore.Create instead of SqlServerTestStore.CreateScratch where possible
Remove async database creation from SqlServerTestStore


AndriySvyryd:
This also makes the SqlServer FTs run 10% faster against LocalDB

src/Microsoft.EntityFrameworkCore.Specification.Tests/LoadTestBase.cs
```diff
@@ -3299,7 +3294,7 @@ protected class Parent
             public SingleCompositeKey SingleCompositeKey { get; set; }
         }
 
-        protected class Child
+        public class Child
```
ajcvickers: >public class Child [](start = 8, length = 18)

Thought we agreed that model types would not become public?

src/Microsoft.EntityFrameworkCore.Specification.Tests/NotificationEntitiesTestBase.cs
```diff
@@ -117,18 +123,19 @@ protected void SetWithNotify<T>(T value, ref T field, [CallerMemberName] string
...
                 modelBuilder.Entity<Blog>().Property(e => e.Id).ValueGeneratedNever();
                 modelBuilder.Entity<Post>().Property(e => e.Id).ValueGeneratedNever();
             }
 
-            protected void EnsureCreated()
+            protected virtual void EnsureCreated()
             {
                 using (var context = CreateContext())
                 {
-                    EnsureClean(context);
```
ajcvickers: > EnsureClean(context); [](start = 19, length = 22)

Why is EnsureClean removed here?

src/Microsoft.EntityFrameworkCore.Specification.Tests/NullKeysTestBase.cs
```diff
@@ -269,7 +269,7 @@ protected void EnsureCreated()
             {
                 using (var context = CreateContext())
                 {
-                    EnsureClean(context);
```
ajcvickers: >EnsureClean(context); [](start = 20, length = 21)

Likewise? Note that all of these use EnsureClean so that changes to database structure don't cause failures on the C.I. Most of these were EnsureDeleted followed by EnsureCreated before they became EnsureClean because of the database structure change issue. It also makes the tests more consistent--there are less likely to be bugs related to running on a clean machine verses a machine that already contains the database.

src/Microsoft.EntityFrameworkCore.Specification.Tests/LoadTestBase.cs
AndriySvyryd: missed these ones

src/Microsoft.EntityFrameworkCore.Specification.Tests/NullKeysTestBase.cs
AndriySvyryd: SqlServerTestStore now calls EnsureClean when creating a database if it already exists.

ajcvickers: Ah, I missed that. It's doing it with an empty context, so it will never create tables and this call to EnsureCreated now only ever creates tables.

PR #6626, opened 2016-09-27
Update partner package versions

pranavkm:
cc @Eilon \ @muratg \ @divega \ @bricelam

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Design/project.json
+++ b/src/Microsoft.EntityFrameworkCore.Design/project.json
@@ -25,17 +25,15 @@
     }
   },
   "dependencies": {
+    "Microsoft.AspNetCore.Hosting.Abstractions": "1.1.0-*",
     "Microsoft.EntityFrameworkCore.Relational.Design": "1.1.0-*",
-    "Microsoft.AspNetCore.Hosting.Abstractions": "1.1.0-*"
+    "NETStandard.Library": "1.6.1-*"
   },
   "frameworks": {
     "net451": {},
     "netstandard1.3": {
       "dependencies": {
-        "System.AppContext": "4.1.0-*",
-        "System.Collections.NonGeneric": "4.0.1-*",
-        "System.IO": "4.1.0-*",
-        "System.IO.FileSystem": "4.0.1-*"
+        "System.Collections.NonGeneric": "4.3.0-*"
       }
     }
   }
--- a/src/Microsoft.EntityFrameworkCore.Relational/project.json
+++ b/src/Microsoft.EntityFrameworkCore.Relational/project.json
@@ -25,7 +25,8 @@
     }
   },
   "dependencies": {
-    "Microsoft.EntityFrameworkCore": "1.1.0-*"
+    "Microsoft.EntityFrameworkCore": "1.1.0-*",
+    "NETStandard.Library": "1.6.1-*"
   },
   "frameworks": {
     "net451": {
@@ -48,10 +49,9 @@
     },
     "netstandard1.3": {
       "dependencies": {
-        "Microsoft.CSharp": "4.0.1-*",
-        "System.Data.Common": "4.1.0-*",
-        "System.Diagnostics.DiagnosticSource": "4.0.0-*",
-        "System.Text.RegularExpressions": "4.1.0-*"
+        "Microsoft.CSharp": "4.3.0-*",
+        "System.Data.Common": "4.3.0-*",
+        "System.Diagnostics.DiagnosticSource": "4.3.0-*"
       }
     }
   }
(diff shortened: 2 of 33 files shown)


pranavkm:
Haven't run the tests as yet. Verified this compiled though.

divega:
Please wait until @bricelam reviews.

src/Microsoft.EntityFrameworkCore.Design/project.json
```diff
@@ -25,17 +25,15 @@
     }
   },
   "dependencies": {
+    "Microsoft.AspNetCore.Hosting.Abstractions": "1.1.0-*",
     "Microsoft.EntityFrameworkCore.Relational.Design": "1.1.0-*",
-    "Microsoft.AspNetCore.Hosting.Abstractions": "1.1.0-*"
+    "NETStandard.Library": "1.6.1-*"
```
bricelam: Isn't this brought in transitively?

bricelam: Do we really want this in the nuspec for Full .NET?

bricelam: (That's a lot of packages you don't really need.)

src/Microsoft.EntityFrameworkCore.Relational/project.json
```diff
@@ -25,7 +25,8 @@
     }
   },
   "dependencies": {
-    "Microsoft.EntityFrameworkCore": "1.1.0-*"
+    "Microsoft.EntityFrameworkCore": "1.1.0-*",
+    "NETStandard.Library": "1.6.1-*"
   },
   "frameworks": {
     "net451": {
```
bricelam: Do these `frameworkAssemblies` still need to be referenced if we use `NETStandard.Library` on Full .NET?

src/Microsoft.EntityFrameworkCore/project.json
```diff
@@ -40,12 +41,9 @@
...
       "dependencies": {
-        "System.Collections.Concurrent": "4.0.12-*",
-        "System.ComponentModel.Annotations": "4.1.0-*",
-        "System.Linq.Queryable": "4.0.1-*",
-        "System.ObjectModel": "4.0.12-*",
-        "System.Reflection.Extensions": "4.0.1-*",
-        "System.Reflection.TypeExtensions": "4.1.0-*"
+        "System.ComponentModel.Annotations": "4.3.0-*",
+        "System.Linq.Queryable": "4.3.0-*",
+        "System.Reflection.TypeExtensions": "4.3.0-*"
```
bricelam: Eww.. How did that creep in?

bricelam:
LGTM assuming we've made conscious decisions regarding my comments.

src/Microsoft.EntityFrameworkCore.Design/project.json
pranavkm: Yes, but we decided it's easier if we just add this to every project. It might also make it easier to independently rev these if we ever create a release where downstream dependencies we own don't get updated.

pranavkm: I could make it `netstandard` specific if this is concerning. My guess was that package trimming should take care of this. cc @Eilon \ @davidfowl

bricelam:
You may need to update [some binding redirects](https://github.com/aspnet/EntityFramework/search?q=bindingRedirect) too.

src/Microsoft.EntityFrameworkCore.Relational/project.json
pranavkm: Possibly not. I'll try removing this and check if it the nuspec ends up with the same references

src/Microsoft.EntityFrameworkCore/project.json
pranavkm: 😞 there's a few in Mvc like these. Perhaps we can fix up some of these in 2.0.

src/Microsoft.EntityFrameworkCore.Design/project.json
Eilon: @pranavkm is correct. We do want this everywhere.

src/Microsoft.EntityFrameworkCore/project.json
bricelam: Filed #6628

src/Microsoft.EntityFrameworkCore.Design/project.json
bricelam: Seems fine. But it seems more DRY just to have the root EF package depend. It'll never be different between packages in the same repo.

divega: That is a very good point.  Maybe we actually don't want to do this for EF Core and its dependencies?

bricelam: IMHO, making it `netstandard`-specific would for us and our dependencies would be ideal.

Personally, I'd *always* make it `netstandard`-specific since it adds little-to-no value elsewhere.

divega: @pranavkm @Eilon should we discuss this further in person?

pranavkm: I think @DamianEdwards was planning on addressing this.

Issue #6620, opened 2016-09-27
GroupBy multiple keys throws exception in 1.0.1


StefH:
### GroupBy multiple keys throws exception

Example code

``` c#
var expected = _context.Posts
  .GroupBy(x => new { x.BlogId, x.PostDate })
  .OrderBy(x => x.Key.PostDate)
  .ToArray();
```
### Exception + StackTrace is like:

```
Test Name:  System.Linq.Dynamic.Core.Tests.EntitiesTests.Entities_GroupBy_MultiKey
Test FullName:  System.Linq.Dynamic.Core.Tests.EntitiesTests.Entities_GroupBy_MultiKey
Test Source:    C:\Projects\GitHub\System.Linq.Dynamic.Core\test\System.Linq.Dynamic.Core.Tests\EntitiesTests.GroupBy.cs : line 46
Test Outcome:   Failed
Test Duration:  0:00:00,097

Result StackTrace:  
at System.Dynamic.Utils.ExpressionUtils.ValidateOneArgument(MethodBase method, ExpressionType nodeKind, Expression arguments, ParameterInfo pi)
   at System.Linq.Expressions.Expression.Call(MethodInfo method, Expression arg0, Expression arg1, Expression arg2, Expression arg3)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.TrackEntitiesInResults[TResult](QueryModel queryModel)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.CreateQueryExecutor[TResult](QueryModel queryModel)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.<>c__DisplayClass19_0`1.<CompileQuery>b__0()
   at Microsoft.EntityFrameworkCore.Query.Internal.CompiledQueryCache.GetOrAddQuery[TResult](Object cacheKey, Func`1 compiler)
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryCompiler.Execute[TResult](Expression query)
   at Remotion.Linq.QueryableBase`1.GetEnumerator()
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source, Int32& length)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
[...]

PR #6629, opened 2016-09-28
[1.0.2] Fix to #6620 - GroupBy multiple keys throws exception in 1.0.1

maumar:
Problem was with Tracking queries involving GroupBy and OrderBy/Skip/Take/Distinct. When tracking results of group by we were trying to extract Grouping element from the result however we always assumed that the Grouping result operator would be on the main QueryModel. When Orderby/Take/Skip etc are used in conjunction with GroupBy, the GroupBy result operator is actually in the subquery. Fix is to properly extract Group Element information from the subquery when needed.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
@@ -3758,6 +3758,69 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [ConditionalFact]
+        public virtual void GroupBy_with_orderby()
+        {
+            AssertQuery<Order>(
+                os => os.GroupBy(o => o.CustomerID).OrderBy(g => g.Key),
+                asserter:
+                    (l2oResults, efResults) =>
+                    {
+                        var efGroupings = efResults.Cast<IGrouping<string, Order>>().ToList();
+
+                        foreach (IGrouping<string, Order> l2oGrouping in l2oResults)
+                        {
+                            var efGrouping = efGroupings.Single(efg => efg.Key == l2oGrouping.Key);
+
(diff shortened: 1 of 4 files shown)


maumar:
@anpete

src/Shared/SharedTypeExtensions.cs
```diff
@@ -81,6 +81,13 @@ private static bool IsInstantiable(TypeInfo type)
                && !type.IsInterface
                && (!type.IsGenericType || !type.IsGenericTypeDefinition);
 
+        public static bool IsGrouping(this Type type) => IsGrouping(type.GetTypeInfo());
```
anpete: Why two methods? Also, this helper is nice. Did you check whether it could be used elsewhere?

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -569,27 +569,34 @@ var lastTrackingModifier
...
             }
 
-            var groupResultOperator 
-                = queryModel.ResultOperators.OfType<GroupResultOperator>().LastOrDefault();
+            var outputExpression = queryModel.SelectClause.Selector;
 
-            var outputExpression
-                = groupResultOperator?.ElementSelector ?? queryModel.SelectClause.Selector;
+            var resultItemType = _expression.Type.GetSequenceType();
+            var isGrouping = resultItemType.IsGrouping();
```
anpete: New line before if

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -3758,6 +3758,69 @@ public virtual void Select_GroupBy_SelectMany()
         }
 
         [ConditionalFact]
+        public virtual void GroupBy_with_orderby()
```
anpete: Do these tests _need_ custom asserters?

anpete:
:shipit:

Issue #5544, opened 2016-05-27
Query :: force client evaluation on First() inside a query


maumar:
We are currently translating First and FirstOrDefault calls into SQL whenever we can, however we should consider forcing client eval on First, so that we can mimic L2O behavior of throwing if there are no elements in the collection. We should continue translating FirstOrDefault

rowanmiller:
> First, so that we can mimic L2O behavior of throwing if there is more than one element in the collection. 

Do you mean throwing if there is **no elements** in the collection?

maumar:
right :)

rowanmiller:
@maumar can you share some details of what happens now (i.e. wrong results, poor exception message, etc.)

maumar:
it behaves just like FirstOrDefault, i.e. it never throws.

maumar:
Clearing for re-triage. This could be a perf regression/breaking change for people who are already using it.

divega:
**EF Team Triage:** we will switch to client eval. `FirstOrDefault()` is the only one of these operators that we can translate to SQL correctly.

maumar:
Fixed in d698b57b561a7c27ff513b131937ad2d8c39de89

PR #6630, opened 2016-09-28
[SMALL] Fix to #5544 - Query :: force client evaluation on First() inside a query

maumar:
We used to try server-evaluate calls to First() in a subquery, however we are unable to properly mimic the exception that could potentially be thrown from First() method. Fix is to always force client evaluation when First() is used in a subquery.
If one wants to avoid client evaluation, FirstOrDefault() should be used.


maumar:
@smitpatel @anpete

smitpatel:
:shipit:

PR #6632, opened 2016-09-28
Implement complex type usages and properties

ajcvickers:
A ComplexTypeUsage represents the reference to a complex type on an entity type or on another complex type. It contains the ComplexProperty instances that represent the properties of the complex type and their facets in this usage.

/cc @AndriySvyryd (But anyone can review...)


ajcvickers:
Agreed. We probably should review it with others--I'll make a note to talk about it when we discuss in the design meeting.

PR #6638, opened 2016-09-29
Metadata cleanup following hierarchy/base class changes

ajcvickers:
These are changes that have been made on the ComplexTypes branch for things that I found following the change to the metadata types hierarchy and the move of some members to IPropertyBase. Making these changes to dev since they aren't strictly related to complex type support and it is better to change them sooner rather than later.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/IPropertyBase.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/IPropertyBase.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Reflection;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 
 namespace Microsoft.EntityFrameworkCore.Metadata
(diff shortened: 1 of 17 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/IPropertyBase.cs
```diff
@@ -39,5 +40,12 @@ public interface IPropertyBase : IAnnotatable
         ///     property is not known.
         /// </summary>
         FieldInfo FieldInfo { get; }
+
+        /// <summary>
+        ///     Gets a value indicating whether this is a shadow property. A shadow property is one that does not have a
+        ///     corresponding property in the entity class. The current value for the property is stored in
+        ///     the <see cref="ChangeTracker" /> rather than being stored in instances of the entity class.
+        /// </summary>
+        bool IsShadowProperty { get; }
```
AndriySvyryd: This could be an extension method now

ajcvickers: It could be, but it would be "more" of a break since it would change from property to method. We can discuss if you want, but I'm fine with leaving as a property.

AndriySvyryd: It's not a big deal, we can leave it

PR #6640, opened 2016-09-29
Don't clean test databases created with a script.

### LINKED ISSUE #6637: Northwind query test fixture massively slower after 92edab1
After 92edab108f, running a single query test against Northwind went from ~5s to ~40s on my machine.

NB: Northwind is readonly and only ever needs to be created once.

cc @AndriySvyryd

AndriySvyryd:
Fixes #6637


anpete:
:shipit: Thanks!

PR #6646, opened 2016-09-30
Code cleanup

AndriySvyryd:
Will push at noon if no one reviews


src/Microsoft.EntityFrameworkCore.Design/Design/DbContextOperations.cs
```diff
@@ -136,17 +136,17 @@ public virtual Type GetContextType([CanBeNull] string name)
             {
                 contexts.Add(
                     context,
-                    FindContextFactory(context) ?? (Func<DbContext>)(() =>
-                    {
-                        try
+                    FindContextFactory(context) ?? (() =>
                         {
```
bricelam: Does this conflict with the VS default auto-format indentation? I can't stand when they flip back and forth every commit.

src/Microsoft.EntityFrameworkCore.InMemory/Query/ExpressionVisitors/Internal/InMemoryEntityQueryableExpressionVisitorFactory.cs
```diff
@@ -39,11 +39,11 @@ public class InMemoryEntityQueryableExpressionVisitorFactory : IEntityQueryableE
...
         /// </summary>
         public virtual ExpressionVisitor Create(
-            EntityQueryModelVisitor queryModelVisitor, IQuerySource querySource)
-            => new InMemoryEntityQueryableExpressionVisitor(
-                _model,
-                _materializerFactory,
-                Check.NotNull(queryModelVisitor, nameof(queryModelVisitor)),
-                querySource);
+            EntityQueryModelVisitor queryModelVisitor, IQuerySource querySource
+        ) => new InMemoryEntityQueryableExpressionVisitor(
```
bricelam: That's an odd place to put a parenthesis

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/NullSemanticsQueryTestBase.cs
```diff
@@ -356,7 +354,7 @@ public virtual void Where_equal_nullable_with_null_value_parameter()
             string prm = null;
 
             AssertQuery<NullSemanticsEntity1>(es => es.Where(e =>
-                e.NullableStringA == prm));
+                    e.NullableStringA == prm));
```
bricelam: I don't understand the reasoning for the double indent here. It's only nice if there's something after it.

src/Microsoft.EntityFrameworkCore.InMemory/Query/ExpressionVisitors/Internal/InMemoryEntityQueryableExpressionVisitorFactory.cs
AndriySvyryd: Without the parenthesis resharper adds an extra ident to parameters and body, I think it's a bug

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/NullSemanticsQueryTestBase.cs
AndriySvyryd: Probably parenthesis and lambda rules overlapping

src/Microsoft.EntityFrameworkCore.Design/Design/DbContextOperations.cs
AndriySvyryd: VS formats it the same way. At least on my machine.

ajcvickers:
Can you summarize what issues with the current formatting rules you are attempting to fix? (I wasn't aware of anything terrible with what we had.)

src/Microsoft.EntityFrameworkCore.InMemory/Query/ExpressionVisitors/Internal/InMemoryEntityQueryableExpressionVisitorFactory.cs
AndriySvyryd: This wasn't caused by a rule change, but (probably) a bug in the current version. By default it would format it like this:
```C#
public virtual ExpressionVisitor Create(
        EntityQueryModelVisitor queryModelVisitor, IQuerySource querySource)
    => new InMemoryEntityQueryableExpressionVisitor(
        _model,
        _materializerFactory,
        Check.NotNull(queryModelVisitor, nameof(queryModelVisitor)),
        querySource);
```
Ideally it should be:
```C#
public virtual ExpressionVisitor Create(
    EntityQueryModelVisitor queryModelVisitor, IQuerySource querySource)
    => new InMemoryEntityQueryableExpressionVisitor(
        _model,
        _materializerFactory,
        Check.NotNull(queryModelVisitor, nameof(queryModelVisitor)),
        querySource);
```
Which is the closesest to statement body formatting:
```C#
public virtual ExpressionVisitor Create(
    EntityQueryModelVisitor queryModelVisitor, IQuerySource querySource)
{
    return new InMemoryEntityQueryableExpressionVisitor(
        _model,
        _materializerFactory,
        Check.NotNull(queryModelVisitor, nameof(queryModelVisitor)),
        querySource);
}
```
But I couldn't find any rule to achieve this.
So the workaround is to move the parenthesis:
```C#
public virtual ExpressionVisitor Create(
    EntityQueryModelVisitor queryModelVisitor, IQuerySource querySource
) => new InMemoryEntityQueryableExpressionVisitor(
    _model,
    _materializerFactory,
    Check.NotNull(queryModelVisitor, nameof(queryModelVisitor)),
    querySource);
```

AndriySvyryd:
The above is the only issue I've seen

src/Microsoft.EntityFrameworkCore.InMemory/Query/ExpressionVisitors/Internal/InMemoryEntityQueryableExpressionVisitorFactory.cs
ajcvickers: To me the "default" one looks better than the "workaround" one.

Issue #6580, opened 2016-09-21
Variable in a group JOIN clause added to ORDER BY as a SQL variable


wc-matteo:
### Steps to reproduce

```cs
from t1 in _dbContext.Rooms
join t2 in _dbContext.ContentRooms on new { id = t1.RoomsId, lang = _languageId } equals new {id = t2.RoomsId, lang = t2.LanguageId} into gj
from subt2 in gj.DefaultIfEmpty()
select new { t1.RoomsId, Name = (subt2 != null ? subt2.Name : null) };
```

```sql
SELECT [t1].[RoomsId], [t1].[AccomodationsId], [t1].[MinDays], [t1].[PeopleValidForEnd], [t1].[PeopleValidForStart], [t1].[PriceTypesId], [t2].[RoomsId], [t
2].[LanguageId], [t2].[Description], [t2].[Name], [t2].[Subtitle]                                                                                                 
FROM [Rooms] AS [t1]                                                                                                                                        
LEFT JOIN [ContentRooms] AS [t2] ON ([t1].[RoomsId] = [t2].[RoomsId]) AND (@___languageId_0 = [t2].[LanguageId])                                            
ORDER BY [t1].[RoomsId], @___languageId_0          
```

### The issue

A C# variable in a group join clause (in this case, `_languageId`) gets added to the `ORDER BY` statement as a SQL variable. 

### Exception

```
System.Data.SqlClient.SqlException: The SELECT item identified by the ORDER BY number 2 contains a variable as part of the expression identifying a column position. Variables are only allowed when ordering by an expression referencing a column name.
 at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)                              
         at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)          
         at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandl
er, TdsParserStateObject stateObj, Boolean& dataReady)                                                                                                            
         at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()                                                                                              
         at System.Data.SqlClient.SqlDataReader.get_MetaData()                                                                                                    
[...]

PR #6656, opened 2016-09-30
Query: Convert Order By constant & @@rowcount to (SELECT 1)

### LINKED ISSUE #6145: Query: OrderBy constant generates incorrect query for integer/boolean
At present in query,
`OrderBy(c => true)` translates to `ORDER BY 1`
`OrderBy(c => 3)` translates to  `ORDER BY 3`
In SqlServer, `ORDER BY` integer means order by that column number from projection which introduces unintended ordering and gives different results compared to what linq gives.
We are also using creating Order by when we do group join with default if empty using the left side of join condition which also faces above issue based on the join condition.
See tests `QuerySqlServerTest.OrderBy_true`, `QuerySqlServerTest.OrderBy_integer` & `GearsOfWarQuerySqlServerTest.Left_join_predicate_value_equals_condition`
Possible work-arounds:
We can remove the order by clause if it is using constant. Results would match this way with linq but for paging cases, order by clause is necessary.
We can introduce dummy order by as follows:
`Order by @@ROWCOUNT`
`Order by @parameter + 1` where @parameter is passed. Value doesn't matter.
@divega

smitpatel:
Fixes #6145 , #6580 
Converted OrderBy with Constant/Parameter to `(SELECT 1)`
Converted dummy `@@rowcount` to `(SELECT 1)`

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -382,6 +382,11 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
                             Visit(aliasExpression.Expression);
                         }
                     }
+                    else if (orderingExpression is ConstantExpression
+                        || orderingExpression is ParameterExpression)
+                    {
+                        _relationalCommandBuilder.Append("(SELECT 1)");
+                    }
                     else
                     {
                         Visit(ApplyOptimizations(orderingExpression, searchCondition: false));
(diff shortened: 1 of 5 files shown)


maumar:
:shipit:

ErikEJ:
Now you broke the SQL CE provider...

smitpatel:
@ErikEJ - From what I understand SQL CE was already facing issue #6145. With these changes it will throw exception instead of incorrect results.
Is there any solution in SQL CE for #6145?

divega:
Adding [providers-beware] after the fact.

ErikEJ:
@smitpatel No, no solution for SQLCE, currently is use GETDATE()

smitpatel:
@ErikEJ - After discussing with @divega , we found some failing tests for CE reported in past but all of those issues are closed. Can you file a new issue for all the tests which are failing for SQL CE so we can take this discussion there and resolve what we can.

ErikEJ:
@smitpatel  I assume you mean a consolidated list of failing test in the SQL CE provider caused by use of nested SELECT statements generated by EF.Relational ?

smitpatel:
@ErikEJ - Yes. Cause can be anything and not just nested `SELECT` (like this one scalar subquery)

ErikEJ:
Will do, thanks - close to 100% fail due to nested SELECTs, though

PR #6670, opened 2016-10-03
Update design-time reporting

### LINKED ISSUE #6643: Commands: Improve reporting architecture
Reporting in the commands is currently based on the early implementations of Logging and PowerShell. A lot has changed since it was first implemented, and it could use some love.

At the very least, we should review the `ILogHandler` interface before RTMing it.

I'd also like to make reporting a first-class concept in the `.Design` assembly and improve its integration with runtime logging.

bricelam:
This uses a logging state interface `IReportedLogState` to mark log messages that should be reported back to the user when running the design-time tools.

Resolves #6643

**Providers:** This introduces a new (better) way to report messages during reverse engineer, but the old way of using the well-known logger name will still work. There is a non-functional breaking change here where only Warning, Information, and Debug log messages will be reported back to the console. I had to update a few Trace messages in our providers.


src/Microsoft.EntityFrameworkCore.Design/Design/Internal/CommandLogger.cs
```diff
@@ -1,148 +0,0 @@
...
-using System.Linq;
-using System.Text;
-using JetBrains.Annotations;
-using Microsoft.EntityFrameworkCore.Migrations.Internal;
-using Microsoft.EntityFrameworkCore.Utilities;
-using Microsoft.Extensions.Logging;
-
-namespace Microsoft.EntityFrameworkCore.Design.Internal
-{
-    // TODO: Consider replacing with something besides Logging
```
bricelam: Done.

src/Microsoft.EntityFrameworkCore.Design/Design/OperationExecutor.cs
```diff
@@ -29,8 +29,8 @@ public OperationExecutor([NotNull] object logHandler, [NotNull] IDictionary args
             Check.NotNull(logHandler, nameof(logHandler));
             Check.NotNull(args, nameof(args));
 
-            var unwrappedLogHandler = ForwardingProxy.Unwrap<IOperationLogHandler>(logHandler);
-            var loggerProvider = new LoggerProvider(name => new CommandLoggerAdapter(name, unwrappedLogHandler));
+            var unwrappedLogHandler = ForwardingProxy.Unwrap<IOperationReportHandler>(logHandler);
+            var reporter = new OperationReporter(unwrappedLogHandler);
```
bricelam: Will rename *logHandler* to *reportHandler*.

bricelam:
@anpete Would you mind skimming through to make sure this is consistent with the latest logging best practices?

bricelam:
Providers, there is no need to react right now. See #6697

PR #6675, opened 2016-10-04
Add support for triggers that modify values in the inserted/updated row to SQL Server

### LINKED ISSUE #6044: Issue with Computed Fields using User Defined Functions
This is the same issue that was described in [https://github.com/aspnet/EntityFramework/issues/5793](https://github.com/aspnet/EntityFramework/issues/5793), but was never resolved because of a lack of code samples.  I've attached code that illustrates the issues. If it's too much code, I can make a smaller sample, but it should be pretty clear. There are actually two issues that I am running into, don't know if they are related or if the other one should be a new issue.

1) When an entity/table a computed column based on a scalar function, EF can't update the record. I have an Order table and an OrderDetail table. I have a scalar function that computes the total cost of an order (summing all of the order detail records) to display on the Order table. Here are the abbreviated details:

Property definition:
```C#
public class Order : EntityBase 
{ 
public int CustomerId { get; set; } 
public string CategoryName {get;set;} 
[DataType(DataType.Currency)]
[DatabaseGenerated(DatabaseGeneratedOption.Computed)] 
public decimal? OrderTotalComputed { get; set; } //shortened for brevity }
}
```
In my context:
```C#
OnModelCreating:
'modelBuilder.Entity(entity =>
{
entity.Property(e => e.OrderTotalComputed)
.HasColumnType("money")
.HasComputedColumnSql("Store.GetOrderTotal([Id])");
//.ValueGeneratedOnAddOrUpdate();
});'
```
When I first create an order record, it works fine. If I update an existing record I get the following error:
Result StackTrace:
```
 at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)
at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.Execute(IEnumerable1 commandBatches, IRelationalConnection connection) at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(IReadOnlyList1 entriesToSave)
at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
at SpyStore.DAL.Tests.Context.OrderTests.ShouldUpdateAnOrder() in C:\GitHub\Responsive\CodeSamples\Chapter05-EF\RTM\SpyStore.DAL\test\SpyStore.DAL.Tests\Context\OrderTests.cs:line 42
at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action1 wrapCloseInAction) at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose) at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady) at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData() at System.Data.SqlClient.SqlDataReader.get_MetaData() at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString) at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, SqlDataReader ds) at System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior) at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.Execute(IRelationalConnection connection, String executeMethod, IReadOnlyDictionary2 parameterValues, Boolean openConnection, Boolean closeConnection)
at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.ExecuteReader(IRelationalConnection connection, IReadOnlyDictionary`2 parameterValues, Boolean manageConnection)
at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)
```
Result Message:
An error occurred while updating the entries. See the inner exception for details.
Column 'inserted.OrderTotalComputed' cannot be referenced in the OUTPUT clause because the column definition contains a subquery or references a function that performs user or system data access. A function is assumed by default to perform data access if it is not schemabound. Consider removing the subquery or function from the column definition or removing the column from the OUTPUT clause.

In the attached code, execute the **SpyStore.DAL.Tests.Context.OrderTests.ShouldUpdateAnOrder** test to see the error in action.

2)    The second issue is that when I have such a computed column, if I add/update order detail records in a DbContext, the computed column doesn't update in my entity. However, if I create a new Context, I get the correct value when retrieving the record. The new DbContext test is illustrated with this test: **SpyStore.DAL.Tests.Context.OrderTests.ShouldGetOrderTotal** (this test uses a new DbContext instead of the one that seeded the data) 
The failing test (where the value doesn't get updated in the same DbContext) is here: **SpyStore.DAL.Tests.Context.OrderTests.ShouldGetOrderTotalAfterAddingAnOrderDetail** 

EF Core version: RTM (1.0.0.0)
Operating system: Win10
Visual Studio version: 2015

[SpyStore.DAL.zip](https://github.com/aspnet/EntityFramework/files/357643/SpyStore.DAL.zip)

### LINKED ISSUE #6474: SQL Server update does not get values updated by trigger.
I'm on 1.0.0 of Microsoft.EntityFramework.Core and Microsoft.EntityFramework.Core.SqlServer.

I have a table with two columns that are updated by an AFTER trigger (not an INSTEAD OF):

`__lastModified` (datetimeoffset)
`__version` (int)

There's also a `__rowVersion` column which is of type rowversion.

When I call `SaveChanges`, I don't get the values after the trigger fires, I get the values before the trigger is fired.

I logged the SQL, and this is what I get:

```
DECLARE @inserted0 TABLE ([__lastModified] datetimeoffset, [__rowVersion] varbinary(8), [__version] int);
UPDATE [stats].[EventSeries] SET [description] = @p0, [twitterScreenName] = @p1, [__lastModifiedByUserId] = @p2
OUTPUT INSERTED.[__lastModified], INSERTED.[__rowVersion], INSERTED.[__version]
INTO @inserted0
WHERE [eventSeriesId] = @p3 AND [__rowVersion] = @p4;
SELECT [__lastModified], [__rowVersion], [__version] FROM @inserted0;
```

I notice the OUTPUT statement in there (note, I've also taken a look at issue #1441 but that is an issue with EF not generating an OUTPUT INTO).

The issue is in the [documentation for the OUTPUT statement, specifically the section on the INSERTED keyword](https://msdn.microsoft.com/en-us/library/ms177564.aspx#Anchor_1) (emphasis mine):

> INSERTED is a column prefix that specifies the value added by the insert or update operation. Columns prefixed with INSERTED reflect the value **after the UPDATE, INSERT, or MERGE statement is completed but before triggers are executed**.

I've configured the `__lastModified` and `__version` columns with `ValueGeneratedOnAddOrUpdate` so the expectation is that upon return of an UPDATE/INSERT, those values would reflect what is in the database when the operation is complete.

As it stands now, the values that come back do *not* reflect what is in the database and the `DbContext` now stores the incorrect values.

The `DbContext` (and by extension, my application) is now inconsistent, and this is a HUGE problem.

There's a simple workaround, which is to refetch the entity, and that's what I'm doing now.  But given that I'm wasting network bytes bringing *something* back from the database on a call to `SaveAsync`, why not bring back the correct values?

Also disturbing is that the `__rowVersion` column isn't brought back correctly; I now have a rowversion which doesn't exist in the database so if I wanted to try to modify the entity, I can't, because the rowversion for consistency checking will never match.

It seems the simple workaround is to start a transaction and just select out the values, like so:

```
BEGIN TRANSACTION;
UPDATE [stats].[EventSeries] SET [description] = @p0, [twitterScreenName] = @p1, [__lastModifiedByUserId] = @p2
WHERE [eventSeriesId] = @p3 AND [__rowVersion] = @p4;
SELECT [__lastModified], [__rowVersion], [__version] FROM [stats].[EventSeries] where [eventSeriesId] = @p3;
COMMIT TRANSACTION;
```

And only select out the values that have been indicated as being refreshed in the database.

AndriySvyryd:
For inserts inner join the INSERTED table instead of reading values directly from it.
For updates use the relational implementation that doesn't use INSERTED table.

Fixes #6044
Fixes #6474

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Update/UpdateSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Update/UpdateSqlGenerator.cs
@@ -155,7 +155,8 @@ namespace Microsoft.EntityFrameworkCore.Update
             AppendFromClause(commandStringBuilder, name, schema);
             // TODO: there is no notion of operator - currently all the where conditions check equality
             AppendWhereAffectedClause(commandStringBuilder, conditionOperations);
-            commandStringBuilder.Append(SqlGenerationHelper.StatementTerminator).AppendLine();
+            commandStringBuilder.Append(SqlGenerationHelper.StatementTerminator).AppendLine()
+                .AppendLine();
 
             return ResultSetMapping.LastInResultSet;
         }
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/StoreGeneratedFixupTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/StoreGeneratedFixupTestBase.cs
@@ -23,7 +23,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
-        public void Add_dependent_then_principal_one_to_many_FK_set_both_navs_set()
+        public virtual void Add_dependent_then_principal_one_to_many_FK_set_both_navs_set()
         {
             ExecuteWithStrategyInTransaction(context =>
                 {
@@ -41,7 +41,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
-        public void Add_dependent_then_principal_one_to_many_FK_not_set_both_navs_set()
+        public virtual void Add_dependent_then_principal_one_to_many_FK_not_set_both_navs_set()
         {
             ExecuteWithStrategyInTransaction(context =>
                 {
@@ -59,7 +59,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
-        public void Add_dependent_then_principal_one_to_many_FK_set_no_navs_set()
+        public virtual void Add_dependent_then_principal_one_to_many_FK_set_no_navs_set()
         {
             ExecuteWithStrategyInTransaction(context =>
                 {
@@ -76,7 +76,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
-        public void Add_dependent_then_principal_one_to_many_FK_set_principal_nav_set()
+        public virtual void Add_dependent_then_principal_one_to_many_FK_set_principal_nav_set()
         {
             ExecuteWithStrategyInTransaction(context =>
                 {
@@ -94,7 +94,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
-        public void Add_dependent_then_principal_one_to_many_FK_set_dependent_nav_set()
+        public virtual void Add_dependent_then_principal_one_to_many_FK_set_dependent_nav_set()
         {
             ExecuteWithStrategyInTransaction(context =>
                 {
@@ -111,7 +111,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
-        public void Add_dependent_then_principal_one_to_many_FK_not_set_principal_nav_set()
+        public virtual void Add_dependent_then_principal_one_to_many_FK_not_set_principal_nav_set()
         {
             ExecuteWithStrategyInTransaction(context =>
                 {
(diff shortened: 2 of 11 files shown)


AndriySvyryd:
CRR: @anpete

anpete:
:shipit:

PR #6678, opened 2016-10-04
Add Microsoft.DotNet.InternalAbstractions to fix tests

pranavkm:


pranavkm:
cc @bricelam \ @divega. There's a change in DependencyModel 1.1-* that removed this dependency. Thish causes `dotnet-test-xunit` to fail. `dotnet-test-xunit` will fix it once the 1.1 build is available on NuGet.org. In the meanwhile, tests that reference DependencyModel (transitively) will require InternalAbstractions to continue working.

This was the fix we did in Mvc. However this fix would require users consuming `Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests` to also do something similar. Do you want me to change the test specification instead?

divega:
@pranavkm if I understand correctly what you are saying it seems that adding the dependency to `Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests` would lead to less changes and less pain (in this case to external provider writers) without any drawback. Does that sound right?

cc @bricelam

pranavkm:
> would lead to less changes and less pain (in this case to external provider writers) without any drawback. 

Essentially. We didn't make an equivalent change in Mvc because it's a runtime package. But since this is a test specification, this should be pretty safe to do.

pranavkm:
Regardless of how we do this, we should encourage consumers of your package not to rely on `Microsoft.DotNet.InternalAbstractions` in their code. It's being deprecated and replaced by [`Microsoft.DotNet.PlatformAbstractions`](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.DotNet.PlatformAbstractions)

divega:
> Regardless of how we do this, we should encourage consumers of your package not to rely on `Microsoft.DotNet.InternalAbstractions` in their code. It's being deprecated and replaced by `Microsoft.DotNet.PlatformAbstractions`

Ok. Is it feasible to swap now?

pranavkm:
🆙 📅

bricelam:
LGTM

PR #6689, opened 2016-10-05
Migrations: Detect renamed properties

### LINKED ISSUE #6582: Migrations: Detect renamed properties
We currently detect renaming a column, but detecting renamed properties is harder because they could easily be intended for adding and removing a column.

We could go down the path of using contextual information like #5155 & #6404 propose, but that seems a little excessive. We could also mimic EF6 where if 80% of the metadata is the same, we consider it a rename, but this yielded false positives.

I propose we consider a column renamed if *only* the name is different. This will catch the same cases as the two referenced issues, and hopefully yield less false positives.

bricelam:
Resolves #6582

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
@@ -16,7 +16,6 @@ using Microsoft.EntityFrameworkCore.Utilities;
 
 namespace Microsoft.EntityFrameworkCore.Migrations.Internal
 {
-    // TODO: Structural matching
     /// <summary>
     ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
     ///     directly from your code. This API may change or be removed in future releases.
@@ -386,8 +385,11 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
         /// </summary>
         protected virtual IEnumerable<MigrationOperation> Diff([NotNull] IEnumerable<string> source, [NotNull] IEnumerable<string> target)
             => DiffCollection(
-                source, target,
-                Diff, Add, Remove,
+                source,
+                target,
+                Diff,
+                Add,
+                Remove,
                 (s, t) => s == t);
 
         /// <summary>
@@ -410,7 +412,8 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        protected virtual IEnumerable<MigrationOperation> Remove([NotNull] string source) => Enumerable.Empty<MigrationOperation>();
+        protected virtual IEnumerable<MigrationOperation> Remove([NotNull] string source)
+            => Enumerable.Empty<MigrationOperation>();
 
         #endregion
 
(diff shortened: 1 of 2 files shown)


thomaseyde:
I wonder, how can this possibly work? We have situations where we delete one property and add another one. These are not related, but EF still guess on a rename. 

When the guess is incorrect, we end up with a database column with incorrect data.

bricelam:
We tried to be more conservative than EF6 where there were even more false positives.

Two things can help prevent false positives:
* Make your migrations atomic--adding and removing several unrelated properties in a single migration should be rare.
* Specify reasonable string lengths on properties. If two columns have different max lengths, they won't be matched.

divega:
@thomaseyde it would be great to know in what articles you got this impression, in case it is within our reach to improve their clarity in the future. FWIW, our documentation contains some more examples of when to modify migrations: https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/#adding-another-migration.

thomaseyde:
The very first [article ](https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/)from this [search](https://www.google.no/search?q=ef+core+migrations&oq=ef+core+migration&aqs=chrome.0.0j69i57j69i65j69i60j35i39l2.3752j0j7&sourceid=chrome&ie=UTF-8) show a simple workflow. Although the 2nd chapter "Add another migration" shows how to rewrite a migration, this information is easy to miss because it's not highlighted in any way.

The second [article](https://docs.microsoft.com/en-us/aspnet/core/data/ef-mvc/migrations) explains how to set up migrations, but nothing about rewriting them. It also points to the up/down methods and the snapshot file. The snapshot file looks like the end result of the migration, which lead us to believe we have to update the snapshot manually whenever we rewrite the suggested migration. 

If this is not the case, why does the snapshot look like the end result? Shouldn't it represent the result from the last migration which this one builds upon?

The next results points to non-Microsoft resources, but all of them miss the point: The added migration is just a best guess:

- https://www.learnentityframeworkcore.com/migrations
- http://www.entityframeworktutorial.net/efcore/entity-framework-core-migration.aspx
- https://www.benday.com/2017/02/14/walkthrough-entity-framework-core-1-1-with-migrations/

Then a [search](https://www.google.no/search?ei=SLG4WqzcKtOKmwWj4Jlo&q=ef+core+migrations+snapshot+file&oq=ef+core+migrations+snapshot+file&gs_l=psy-ab.3..33i160k1.1051913.1056412.0.1056726.16.15.1.0.0.0.117.1146.12j3.15.0....0...1c.1.64.psy-ab..0.15.1080...0j35i39k1j33i21k1.0.yjn10vbKfwk) for the snapshot file return an [explanation](https://www.learnentityframeworkcore.com/migrations/model-snapshot) that migrations should be removed with ´revert-migration´ so the snapshot also will be updated. This implies a manual edit of the migration will render the snapshot invalid. 

So it's not only that the community seems to miss the fact that a generated migration is just a suggestion, but the tooling itself implies we shouldn't touch the result.

bricelam:
The EF Core article is meant to be read in entirety to understand how the Migrations feature works. You rarely have to review the initial migration since it's just creating the tables from scratch. If your model is correct, the tables will be correct. The places EF has to make a educated guesses are when interpreting the differences between two models.

The section in the MVC article about the model snapshot could use some clarification. It doesn't need to show the actual code. I'm also not sure what point it's trying to make by describing these details of the Migrations implementation.

For a better understand of the model snapshot, see the [EF Core Migrations in Team Environments](https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/teams) article. If you're not working in a team environment, you can largely ignore the snapshot file.

The model snapshot loosely represents the database structure your model need in order to work. At the end of the day, EF doesn't care too much about how that structure is achieved. (e.g. data could be coming from either a table or a view; default values could be computed from default constraints or triggers; etc.) You're free to modify the migrations however you want so long as you satisfy the database structure. When you add another migration, EF looks at the difference between the snapshot and your current model to scaffold operations. If you haven't deviated from the expected structure, those new operations should still be perfectly valid even if you've altered the previous migration's operations.

> the tooling itself implies we shouldn't touch the result

The reason we open the migration file after scaffolding it is to let you review and modify it. When we scaffold something that could potentially delete data, we further caution you with a warning to **Please review the migration for accuracy.**

PR #6690, opened 2016-10-05
[TINY] Fix to #6691 - Provider specific ExpressionFragmentTranslators should override default ones

maumar:
This is a fix to https://github.com/aspnet/EntityFramework/issues/6691

We did similar change for the MethodCallTranslators.


maumar:
@anpete @smitpatel

anpete:
:shipit:

PR #6692, opened 2016-10-05
Commands: Report SQL when verbose

### LINKED ISSUE #4401: Migrations: verbose logging to include actual sql
When running an update with `--verbose` I got his message:

    C:\code\RevelarPortalSolution\src\RevelarMVC>dnx ef database update --verbose
    Finding DbContext classes...
    Using context 'ApplicationDbContext'.
    Using database 'NewRevelarPortal' on server '(local)'.
    Applying migration '20160126184050_practitionerupdates'.
    System.Data.SqlClient.SqlException (0x80131904): The ALTER TABLE statement conflicted with the FOREIGN KEY constraint "FK_Distributor_Countries_CountryId". The conflict occurred in database "NewRevelarPortal", table "dbo.Countries", column 'Id'.

Although it isn't much of a leap to figure out in this relatively small migration, it would be even nicer to see the SQL that caused the error, as well as the SQL that occurred before it - which might help me understand if this is just an ordering issue or if I need to do something more drastic to the migration. And in a big migration with many edits it could be even more useful to have that extra info.

bricelam:
Resolves #4401

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
@@ -82,7 +82,8 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
             var connection = _connection.DbConnection;
             _logger.ReportDebug(
                 RelationalEventId.MigrateUsingConnection,
-                () => RelationalStrings.UsingConnection(connection.Database, connection.DataSource));
+                connection,
+                c => RelationalStrings.UsingConnection(c.Database, c.DataSource));
 
             if (!_historyRepository.Exists())
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/Migrator.cs
```diff
@@ -365,5 +385,18 @@ protected virtual IReadOnlyList<MigrationCommand> GenerateUpSql([NotNull] Migrat
...
         }
+
+        private string FormatCommandsForReporting(IEnumerable<MigrationCommand> commands)
+        {
+            var builder = new IndentedStringBuilder();
+            foreach (var command in commands)
+            {
+                builder
+                   .AppendLine(command.CommandText)
+                   .AppendLine(_sqlGenerationHelper.BatchTerminator);
```
smitpatel: Add 1 more line here to separate each commands?

bricelam: I think `BatchTerminator` does that. I'll double-check.

bricelam: Confirmed. They are separated and consistent with `Script-Migration`.

Issue #6720, opened 2016-10-07
The temporarity of a value is not restored after an exception during SaveChanges


AndriySvyryd:
This makes the next retry fail with
> Cannot insert explicit value for identity column in table 'Products' when IDENTITY_INSERT is set to OFF.

Repro: see #6721

PR #6721, opened 2016-10-07
Add methods to ExecutionStrategy that allow suppressing exceptions that don't indicate a failure

### LINKED ISSUE #6665: Add more error codes to the transient list
Memory optimized optimistic concurrency: https://msdn.microsoft.com/en-us/library/mt668435.aspx#Anchor_5
121 VIP Swap: https://social.msdn.microsoft.com/Forums/azure/en-US/5e195f94-d4d2-4c2d-8a4e-7d66b4761510/vip-swap-and-the-semaphore-timeout-period-has-expired-errors?forum=ssdsgetstarted
49918, 49919, 49920: https://azure.microsoft.com/en-us/documentation/articles/sql-database-develop-error-messages/

### LINKED ISSUE #6680: Make ExecutionStrategy more flexible
Refactor `ExecutionStrategy` to make it easier to implement custom retry conditions, such as transaction commit failure handling.

AndriySvyryd:
Add more transient error numbers and remove one that is not transient
Rename SqlAzureExecutionStrategy to SqlServerTransientExecutionStrategy

Fixes #6680
Fixes #6665

@ajcvickers This contains tests for #6720

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Storage/ExecutionResult.cs
+++ b/src/Microsoft.EntityFrameworkCore/Storage/ExecutionResult.cs
@@ -1,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Storage
+{
+    /// <summary>
+    ///     Represents the execution state of an operation.
+    /// </summary>
+    /// <typeparam name="TResult">The type of the result.</typeparam>
+    public class ExecutionResult<TResult>
+    {
+        /// <summary>
+        ///     Creates a new instance of <see cref="ExecutionResult{TResult}" />.
+        /// </summary>
+        /// <param name="successful"><c>true</c> if the operation succeeded.</param>
+        /// <param name="result">The result of the operation if successful.</param>
+        public ExecutionResult(bool successful, [CanBeNull] TResult result)
+        {
+            IsSuccessful = successful;
+            Result = result;
+        }
+
+        /// <summary>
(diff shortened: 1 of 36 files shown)


AndriySvyryd:
@ErikEJ I filed https://github.com/aspnet/EntityFramework6/issues/83

src/Microsoft.EntityFrameworkCore/Storage/ExecutionResult.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    /// <typeparam name="TResult">The type of the result.</typeparam>
+    public class ExecutionResult<TResult>
+    {
+        /// <summary>
+        ///     Creates a new instance of <see cref="ExecutionResult{TResult}" />.
+        /// </summary>
+        /// <param name="isSuccessful"><c>true</c> if the operation succeeded.</param>
+        /// <param name="result">The result of the operation if successful.</param>
+        public ExecutionResult(bool isSuccessful, [CanBeNull] TResult result)
```
smitpatel: no prefix on parameter.

AndriySvyryd:
CRR: @ajcvickers, @divega 
Still to decide whether these methods should also be on `IExecutionStrategy` so they are available from `.Database.CreateExecutionStrategy()`, but they might not make much sense for non-transactional stores.

divega:
> Still to decide whether these methods should also be on IExecutionStrategy so they are available from .Database.CreateExecutionStrategy(), but they might not make much sense for non-transactional stores.

@AndriySvyryd and I talked and we came up with the idea of having a minimal set of methods a provider would need to implement for `IExecutionStrategy` so that we can do the rest on top. He will update the PR.

AndriySvyryd:
Updated

PR #6724, opened 2016-10-07
Remove parameters with default values from IRelationalCommand

### LINKED ISSUE #6700: Clean API of IRelationalCommand and related classes
* Methods on `IRelationalCommand` should not have parameters with default values, in particular `manageConnection` can be removed completely since assuming it's `true` always will not affect actual behavior due to how these methods are called.
 * `RelationalCommandBuilderFactory` should use a public type for the `ISensitiveDataLogger` category

AndriySvyryd:
Make RelationalCommandBuilderFactory use a public type for the ISensitiveDataLogger category

Fixes #6700

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalCommandExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalCommandExtensions.cs
@@ -1,104 +1,91 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 
+// ReSharper disable once CheckNamespace
 namespace Microsoft.EntityFrameworkCore.Storage
 {
-    /// <summary>
-    ///     <para>
-    ///         A command to be executed against a relational database.
-    ///     </para>
-    ///     <para>
-    ///         This type is typically used by database providers (and other extensions). It is generally
-    ///         not used in application code.
-    ///     </para>
-    /// </summary>
-    public interface IRelationalCommand
+    public static class RelationalCommandExtensions
     {
-        /// <summary>
-        ///     Gets the command text to be executed.
-        /// </summary>
-        string CommandText { get; }
-
-        /// <summary>
-        ///     Gets the parameters for the command.
-        /// </summary>
-        IReadOnlyList<IRelationalParameter> Parameters { get; }
-
         /// <summary>
         ///     Executes the command with no results.
         /// </summary>
+        /// <param name="command"> The command to be executed. </param>
         /// <param name="connection"> The connection to execute against. </param>
-        /// <param name="parameterValues"> The values for the parameters. </param>
         /// <returns> The number of rows affected. </returns>
-        int ExecuteNonQuery(
-            [NotNull] IRelationalConnection connection,
-            [CanBeNull] IReadOnlyDictionary<string, object> parameterValues);
+        public static int ExecuteNonQuery(
+            [NotNull] this IRelationalCommand command,
+            [NotNull] IRelationalConnection connection)
+            => command.ExecuteNonQuery(connection, parameterValues: null);
 
         /// <summary>
         ///     Asynchronously executes the command with no results.
         /// </summary>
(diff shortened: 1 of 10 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalCommandExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+{
+    public static class RelationalCommandExtensions
+    {
+        /// <summary>
+        ///     Executes the command with no results.
+        /// </summary>
+        /// <param name="command"> The command to be executed. </param>
+        /// <param name="connection"> The connection to execute against. </param>
+        /// <returns> The number of rows affected. </returns>
+        public static int ExecuteNonQuery(
```
anpete: Should these methods have Checks?

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalCommandExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+            [NotNull] this IRelationalCommand command,
+            [NotNull] IRelationalConnection connection)
+            => command.ExecuteNonQuery(connection, null);
```
anpete: Add parameter name when passing null.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommand.cs
```diff
@@ -30,22 +30,18 @@ public class MigrationCommand
```
anpete: Breaks in here?

src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalCommand.cs
```diff
@@ -34,83 +34,71 @@ public interface IRelationalCommand
         /// </summary>
```
anpete: Breaks in here?

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
```diff
@@ -267,21 +251,13 @@ public class RelationalCommand : IRelationalCommand
                     exception,
                     async: false);
 
-                if (openConnection && !closeConnection)
-                {
-                    connection.Close();
-                }
+                connection.Close();
```
anpete: Why is this changing? It seemed DRY'er in the finally and it no longer checks the flag on the error path

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalCommandExtensions.cs
AndriySvyryd: No, the delegated methods have them

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalCommandExtensions.cs
AndriySvyryd: Ok

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommand.cs
AndriySvyryd: It's unlikely this is used outside our code

src/Microsoft.EntityFrameworkCore.Relational/Storage/IRelationalCommand.cs
AndriySvyryd: It's unlikely this is used outside our code, if it is recompilation would be enough to fix.

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
AndriySvyryd: There were two checks, now there's only one - cleaner.

anpete: Is there a semantic change here, though? Any user code that handles the exception will now have the connection closed always.

AndriySvyryd: No, finally is called before propagating the exception up the stack

anpete:
:shipit:

PR #6725, opened 2016-10-07
Downgrade to 1.0.0 Microsoft.Extensions.DependencyModel

dougbu:
- workaround <https://github.com/dotnet/core-setup/issues/423>


natemcmaster:
:shipit: to unblock CI. I'm assuming you will revert once the external issue is fixed.

dougbu:
57b7656

dougbu:
Filed #6727 to do the needful once we have a newer runtime.

PR #6729, opened 2016-10-08
Fix to 6702 - Query: Null semantics issue after improved nav. translation

maumar:
Problem was that we were not adding null semantics compensation for some cases that were enabled after LOJ translation to SQL.
Specifically, when we decide whether to apply null semantics we look at the expression nullability, and for Column/Alias expression we also look at the IProperty to see if it's nullable (e.g. in case of a non-nullable string).
LOJ translation enabled scenarios where the IProperty is non-nullable, but the actual value can be null, because it comes from an optional navigation - we should be applying null semantics for those cases.

Fix is to add extra metadata to ColumnExpression telling us when the column can be null, even if the IProperty says otherwise. The field is being set during ConditionalAccessExpression translation.

Also fixed minor bugs found along the way:
- we were incorrectly applying one optimization to a query comparing negated nullable column to constant true/false
- ConditionalAccessExpression.ToString() was incorrect for cases where accessor was a method call, and not member/property

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/PredicateReductionExpressionOptimizer.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/PredicateReductionExpressionOptimizer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Linq.Expressions;
 using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Query.Expressions;
 using Remotion.Linq.Parsing;
 
 namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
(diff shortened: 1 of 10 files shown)


maumar:
@anpete @smitpatel

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/PredicateReductionExpressionOptimizer.cs
```diff
@@ -95,5 +98,15 @@ protected override Expression VisitBinary(BinaryExpression node)
...
             return base.VisitBinary(node);
         }
+
+        private bool NegatedNullableAliasOrColumn(Expression expression)
+        {
+            var unaryExpression = expression.RemoveConvert() as UnaryExpression;
+
+            return unaryExpression != null
+                && unaryExpression.Type == typeof(bool?)
+                && (unaryExpression.Operand.RemoveConvert() is AliasExpression
```
anpete: Consider our extension methods here.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -1020,6 +1020,14 @@ var nullConditionalExpression
             {
                 var newAccessOperation = Visit(nullConditionalExpression.AccessOperation);
 
+                var columnExpression = ((newAccessOperation as AliasExpression)?.Expression as ColumnExpression)
```
anpete: Consider our extension methods here.

anpete:
:shipit:

PR #6731, opened 2016-10-08
Remove FKs set by conventions when adding a key.

### LINKED ISSUE #6490: The instance of entity type X cannot be tracked because another instance of this type with the same key is already being tracked
Hi,

I'm getting this error when I try to Insert domain object containing 2 level deep sub objects:

> The instance of entity type 'Toy' cannot be tracked because another instance of this type with the same key is already being tracked. When adding new entities, for most key types a unique temporary key value will be created if no key is set (i.e. if the key property is assigned the default value for its type). If you are explicitly setting key values for new entities, ensure they do not collide with existing entities or temporary values generated for other new entities. When attaching existing entities, ensure that only one entity instance with a given key value is attached to the context.

I have shared my project here: [https://github.com/jarzimichal/ef7-tracking-issue/tree/master/src/EF7](https://github.com/jarzimichal/ef7-tracking-issue/tree/master/src/EF7)

Here is my code:

```
public abstract class Entity
    {
        [Key]
        [DatabaseGenerated(DatabaseGeneratedOption.Identity)]
        public int IdRow { get; set; }
    }

    public class Parent : Entity
    {

        public ICollection<Child> Childs { get; set; }
    }

    public class Child : Entity
    {
        public ICollection<Toy> Toys { get; set; }
    }

    public class Toy : Entity
    {
    }
```

```
        {
            var opt = new DbContextOptionsBuilder<RootContext>();
            opt.UseSqlServer("Data Source=(localdb)\\MSSQLLocalDB;Initial Catalog=ef7-test-db; Integrated Security=True;");
            //opt.UseInMemoryDatabase();

            using (IDbContext db = new RootContext(opt.Options))
            {
                var rep = new Repository(db);

                var parent = new Parent
                {
                    Name = "Parent",
                    Childs = new List<Child>
                    {
                        new Child {Name = "Child1", Toys = new List<Toy>
                        {
                            new Toy { Name = "Toy1"}, //When this line is commented, code works fine
                            new Toy { Name = "Toy2"},
                        }},
                        new Child {Name = "Child1"}
                    }
                };

                rep.Insert(parent);
                rep.SaveChanges();
            }
        }
```

I have investigated the entity in debugger and looks like EF assigned the same '-2147482647' value to IdRow property in both Toys objects.

AndriySvyryd:
Fixes #6490

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/DataAnnotationTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/DataAnnotationTestBase.cs
@@ -707,6 +707,47 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
+        public virtual void Key_property_is_not_used_for_FK_when_set_by_annotation()
+        {
+            var modelBuilder = CreateModelBuilder();
+
+            modelBuilder.Entity<Parent>();
+            modelBuilder.Entity<Child>();
+            var toy = modelBuilder.Entity<Toy>();
+
+            Assert.False(toy.Metadata.GetForeignKeys().Any(fk => fk.IsUnique == false && fk.Properties.Any(p => p.Name == nameof(Toy.IdRow))));
+
+            Validate(modelBuilder);
+        }
+
+        public class Parent
+        {
+            [Key]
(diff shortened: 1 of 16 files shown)


rowanmiller:
@smitpatel could you take a look at this one

smitpatel:
:shipit:

PR #6741, opened 2016-10-10
Commands: Log everything

### LINKED ISSUE #6697: Commands: Show full log
We discussed #6643 in a design meeting and decided on a different approach.

Instead of using `IReportedLogState`, we will whitelist the event IDs we want reported when `--verbose` isn't specified. If needed, we'll allow providers to add to the whitelist, but this is probably YAGNI.

Specifying `--verbose` will ignore the whitelist and report all events `Debug` and higher. (Or just all events depending how noisy it is.)

bricelam:
Resolves #6697

ℹ️ **Providers:** You no longer need to name your logger *Microsoft.EntityFrameworkCore.Tools* for messages to be shown in the design-time tools (i.e. `Scaffold-DbContext` and `dotnet ef dbcontext scaffold`)

Dependents:
* aspnet/EntityFramework.Tools#7


ErikEJ:
providers beware?

bricelam:
Nope. We used to use a well-known logger name for reverse engineer, but this will simply show all log messages regardless of name.

bricelam:
Providers are free to update their design-time logging to closer resemble the runtime (like this change does), but it is not required.

divega:
> Providers are free to update their design-time logging to closer resemble the runtime (like this change does), but it is not required.

That alone makes it qualify for [providers-beware] in my mind 😄

bricelam:
Maybe we need a `providers-be-aware` for non-breaking changes

ErikEJ:
Yes, Brice, I think that was what I was asking for

smitpatel:
This seems to be reverting #6692

bricelam:
@smitpatel Yes, it reverts the use of `IReportedLogData` but not the use of eventIds from that change.

bricelam:
@smitpatel Yes, the SQL will be logged by another component in the stack. Leaving that in would show it twice.

Issue #6295, opened 2016-08-11
Key.ReferencingForeignKeys can have different orderings


ajcvickers:
Depending on how the model is built, Key.ReferencingForeignKeys can get FKs added in different orders. These models should be functionally equivalent, but could cause issues with name-generation etc. This can in turn cause issues with model diffing, etc. Therefore we might want to make this collection have a fixed order. We should also look for other similar collections and do likewise.

PR #6753, opened 2016-10-12
Explicit ordering for Key.ReferencingForeignKeys

ajcvickers:
And also some other metadata collections.

Issue #6295

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/EntityType.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/EntityType.cs
@@ -185,13 +185,14 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
         private void UpdateBaseTypeConfigurationSource(ConfigurationSource configurationSource)
             => _baseTypeConfigurationSource = configurationSource.Max(_baseTypeConfigurationSource);
 
-        private readonly List<EntityType> _directlyDerivedTypes = new List<EntityType>();
+        private readonly SortedSet<EntityType> _directlyDerivedTypes = new SortedSet<EntityType>(EntityTypeNameComparer.Instance);
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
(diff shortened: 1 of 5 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Internal/EntityType.cs
```diff
@@ -185,13 +185,13 @@ public EntityType([NotNull] Type clrType, [NotNull] Model model, ConfigurationSo
...
 
-        private readonly List<EntityType> _directlyDerivedTypes = new List<EntityType>();
+        private readonly SortedSet<EntityType> _directlyDerivedTypes = new SortedSet<EntityType>(EntityTypeNameComparer.Instance);
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual IReadOnlyList<EntityType> GetDirectlyDerivedTypes() => _directlyDerivedTypes;
+        public virtual ISet<EntityType> GetDirectlyDerivedTypes() => _directlyDerivedTypes;
```
AndriySvyryd: ISet is not readonly, perhaps IReadOnlyCollection would fit better

src/Microsoft.EntityFrameworkCore/Metadata/Internal/Key.cs
```diff
@@ -117,7 +117,7 @@ public virtual IPrincipalKeyValueFactory<TKey> GetPrincipalKeyValueFactory<TKey>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual List<ForeignKey> ReferencingForeignKeys { get; [param: CanBeNull] set; }
+        public virtual ISet<ForeignKey> ReferencingForeignKeys { get; [param: CanBeNull] set; }
```
AndriySvyryd: IReadOnlyCollection?

PR #6757, opened 2016-10-12
Fix info about related entities in Attach/Update API docs

### LINKED ISSUE #6739: API docs for DbContext.Update etc. have incorrect info about state of related entities
### Steps to reproduce
```csharp
var context = new TestContext();
var author = new Author {
    AuthorId = 1,
    FirstName = "William",
    LastName = "Shakespeare"
};
author.Books.Add(new Book { Title = "Othello" });
context.Update(author);
context.SaveChanges();
```
### The issue
I would expect, based on the [summary for the Update method](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore/DbContext.cs#L640) that the ChangeTracker should contain two entries as a result of the above code, both in the `Modified` state. I would also expect a `DbUpdateException` to be thrown when EF attempts to update a book with no key value provided. However, the Book entry is in the `Added` state, which, while it makes sense, does not conform to the documentation for the method.

I would also expect the `IsKeySet` property to be `true` for the author, but `false` for the book. It is `true` for both.

### Further technical details

EF Core version: 1.0.1

```json
{
  "version": "1.0.0-*",
  "buildOptions": {
    "emitEntryPoint": true
  },
  "dependencies": {
    "Microsoft.Extensions.Configuration": "1.0.0",
    "Microsoft.Extensions.Configuration.FileExtensions": "1.0.0",
    "Microsoft.Extensions.Configuration.Json": "1.0.0",
    "Microsoft.EntityFrameworkCore.SqlServer": "1.0.1",
    "Microsoft.EntityFrameworkCore.Tools": "1.0.0-preview2-final",
    "Microsoft.NETCore.App": {
      "type": "platform",
      "version": "1.0.0"
    }
  },
  "frameworks": {
    "netcoreapp1.0": {
      "imports": "dnxcore50"
    }
  },
  "tools": {
    "Microsoft.EntityFrameworkCore.Tools": "1.0.0-preview2-final"
  }
}
```

rowanmiller:
Fixes #6739

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/DbContext.cs
+++ b/src/Microsoft.EntityFrameworkCore/DbContext.cs
@@ -486,9 +486,20 @@ namespace Microsoft.EntityFrameworkCore
         }
 
         /// <summary>
-        ///     Begins tracking the given entity, and any other reachable entities that are
-        ///     not already being tracked, in the <see cref="EntityState.Unchanged" /> state such that no
-        ///     operation will be performed when <see cref="SaveChanges()" /> is called.
+        ///     <para>
+        ///         Begins tracking the given entity in the <see cref="EntityState.Unchanged" /> state 
+        ///         such that no operation will be performed when <see cref="DbContext.SaveChanges()" /> 
+        ///         is called.
+        ///     </para>
+        ///     <para>
+        ///         A recursive search of the navigation properties will be performed to find reachable entities
+        ///         that are not already being tracked by the context. These entities will also begin to be tracked 
+        ///         by the context. If a reachable entity has its primary key value set
+        ///         then it will be tracked in the <see cref="EntityState.Unchanged" /> state. If the primary key
+        ///         value is not set then it will be tracked in the <see cref="EntityState.Added" /> state. 
+        ///         An entity is considered to have its primary key value set if the primary key property is set 
+        ///         to anything other than the CLR default for the property type.
+        ///     </para>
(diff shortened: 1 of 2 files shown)


rowanmiller:
@ajcvickers any chance you could take a look at this one

src/Microsoft.EntityFrameworkCore/DbContext.cs
```diff
@@ -501,15 +512,22 @@ private void SetEntityState(InternalEntityEntry entry, EntityState entityState)
...
-        ///         <see cref="EntityState.Unchanged" /> state and then use the returned <see cref="EntityEntry{TEntity}" />
-        ///         to mark the desired properties as modified.
+        ///         <see cref="Attach{TEntity}(TEntity)" /> to begin tracking the entity in the <see cref="EntityState.Unchanged" />
+        ///         state and then use the returned <see cref="EntityEntry" /> to mark the desired properties as modified.
+        ///     </para>
+        ///     <para>
+        ///         A recursive search of the navigation properties will be performed to find reachable entities
+        ///         that are not already being tracked by the context. These entities will also begin to be tracked 
+        ///         by the context. If a reachable entity has its primary key value set
+        ///         then it will be tracked in the <see cref="EntityState.Unchanged" /> state. If the primary key
```
ajcvickers: Modified

PR #6758, opened 2016-10-12
Query: Fix #6123 - Include on InMemory store with Single loads more t…

### LINKED ISSUE #6123: Include on InMemory store with Single loads more than it should
The following repro should load only one Post object. Two are loaded when using the in-memory store. When using SQLite or SQL Server the behavior is correct.
```C#
{

        => optionsBuilder.UseInMemoryDatabase();
        //=> optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=WonkyVB;Trusted_Connection=True;");
}

public class Blog
{
    public int Id { get; set; }

    public ICollection<Post> Posts { get; set; } 
}

public class Post
{
    public int Id { get; set; }

}

public class Program
{
    public static void Main()
    {
        {

            context.Add(new Blog { Posts = new List<Post> { new Post(), new Post() } });
            context.SaveChanges();
        }

        {
            var post = context.Posts.Include(e => e.Blog).Single(p => p.Id == 1);

            Console.WriteLine(post.Blog.Posts.Count);
        }
    }
}
```

anpete:
…han it should

- Moved injection of Include logic higher up the query tree so that it will run after filtering ops.


anpete:
@maumar

maumar:
:shipit:

Issue #6451, opened 2016-09-01
Query :: Include doesn't work with SelectMany


Bartmax:
### Steps to reproduce
```
return context.SiteBanners.Include(sb => sb.Layers).ThenInclude(l => l.SiteBannerFile)
    .SelectMany(b => b.Layers).Include(l=>l.SiteBannerFile).FirstOrDefaultAsync(l => l.FileId == id);
```

### The issue
this does populate the `SiteBannerLayer` entity but not the `SiteBannerFile`.
`SiteBannerFile` is null.

In the example above I used Include everywhere just in case.

Of course this works:
```
var banners = await context.SiteBanners.Include(sb => sb.Layers).ThenInclude(l => l.SiteBannerFile).ToListAsync();
return banners.SelectMany(b => b.Layers).FirstOrDefault(l => l.FileId == id);
```

but that's an inefficient query

### Further technical details

EF Core version: 1.0
Operating system: Windows 10
Visual Studio version: 2015

Other details about my project setup:
Here are the entities involved.
```
public class SiteBanner
{
    public SiteBanner()
    {
        Layers = new List<SiteBannerLayer>();
    }
    public int SiteBannerId { get; set; }

    public string Title { get; set; }

    public ICollection<SiteBannerLayer> Layers { get; set; }
}
public class SiteBannerLayer : ISortableFile
{
    SiteBannerLayer()
    {/*EF*/ }
    public SiteBannerLayer(FileData file, int order = 0)
    {
        if (file == null)
        {
            throw new ArgumentNullException(nameof(file));
        }
        SiteBannerFile = new SiteBannerFile(file);
        Order = order;
    }

    public int SiteBannerLayerId { get; set; }

    [Required]
    public SiteBannerFile SiteBannerFile { get; set; }

    public int Order { get; set; }

    public IFileData File => SiteBannerFile;

    public Guid FileId => SiteBannerFile.SiteBannerFileId;
        
}
public class SiteBannerFile : IFileData
{
    private const string folder = "SiteBanner";
    public SiteBannerFile() { } /*EF*/
    public SiteBannerFile(FileData fileData)
    {
        SiteBannerFileId = fileData.FileDataId;
        Filename = fileData.Filename;
        OriginalFilename = fileData.OriginalFilename;
        ContentType = fileData.ContentType;
    }

    [Key]
    public Guid SiteBannerFileId { get; private set; }

    [Required]
    public string ContentType { get; set; }

    [Required]
[...]

PR #6759, opened 2016-10-12
Fix to #6451 - Query :: Include doesn't work with SelectMany

maumar:
Problem was that we were not able to match query source we were trying to include, if it came from the SelectMany since it has different structure than just regular filter/projection etc.
Fix is to compensate for the SelectMany case when constructing IncludeSpecification by using the correct query source.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
@@ -450,9 +450,12 @@ namespace Microsoft.EntityFrameworkCore.Query
                                         includeResultOperator.NavigationPropertyPath));
                             }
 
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -450,9 +450,12 @@ var includeSpecifications
                                         includeResultOperator.NavigationPropertyPath));
                             }
 
+                            var selectManyQuerySource = ((includeResultOperator.QuerySource as AdditionalFromClause)
+                                ?.FromExpression as QuerySourceReferenceExpression)?.ReferencedQuerySource;
```
anpete: Move this logic into IncludeResultOperator.QuerySource. Why only AdditionalFromClause and not joins etc.?

anpete:
@maumar Is this stale now?

PR #6773, opened 2016-10-14
Refactor ModelValidator and make it a public provider service

### LINKED ISSUE #2416: Refactor ModelValidator
ModelValidator currently has two responsibilities - validating the model and presenting the errors to the user. These should be separated into two types to allow each to be customized separately.
Additionally since providers can have additional validators the core validator should be registered as a different service to allow overriding one without affecting the other and to enable combining the provider validators when using a shared model (see #1904)

AndriySvyryd:
Separate the validation rules between core and provider-specific
Use warnings as errors to enable exceptions to be thrown on validation warnings

Fixes #2416 

CRR: @anpete, @rowanmiller (docs)


AndriySvyryd:
@divega Yes, but we could avoid it by not changing the namespace if we don't want to make breaking changes in internal code either.

AndriySvyryd:
Undid the breaking changes to `RelationalModelValidator`

AndriySvyryd:
This has been postponed till 1.2

PR #6780, opened 2016-10-14
Move RelationalModelValidator to .Infrastructure

AndriySvyryd:


AndriySvyryd:
@roji This change was part of https://github.com/aspnet/EntityFramework/pull/6773, but we realized it would break the current build of Npgsql provider. Are you planning to release a new version that would work with EF 1.1-preview1 (sometime this month)? Otherwise we can postpone this particular change till 1.1 RTM so you have time to react.
CC: @divega, @rowanmiller

AndriySvyryd:
This has been postponed till 1.2

roji:
@AndriySvyryd sorry for not answering sooner, this slipped under my radar.

I sync my dev branch with EFCore's pretty regularly, which means the 1.1 Npgsql provider is never too far behind. I do intend on releasing a version that would be compatible with 1.1-preview1 pretty soon, so if the only reason to postpone this to 1.2 is Npgsql, you can definitely revert it and I'll make the necessary changes on my end.

(BTW if you need an urgent answer on something and I don't seem to be responding to the issue feel free to email me directly)

AndriySvyryd:
@roji No, we reconsidered and decided that even the part that wasn't breaking Npgsql was too risky for 1.1 at this point.

Issue #6380, opened 2016-08-20
Query: Incorrect table select for star projection in subquery


mstaessen:
### Steps to reproduce

See reproduction case: https://github.com/mstaessen/EFDistinctLinqToSql
### The issue

From all the results of a join table (M-N relationships, modeled as two 1-N relationships), I want to get all distinct results from one side of the join table. Generated SQL contains Disctinct clause but for wrong table.

Observed SQL:

```
SELECT [t].[Id], [t].[Source], [t].[ForeignId], [t].[CreationDate], [t].[LastUpdate], [t].[RemovalDate], [t].[Details], [t].[IsRemoved]
FROM (
    SELECT DISTINCT [x.QueryRun.QueryDefinition0].*
    FROM [QueryRunResult] AS [x0]
    INNER JOIN [Property] AS [x.Property0] ON [x0].[PropertyId] = [x.Property0].[Id]
    INNER JOIN [QueryRun] AS [x.QueryRun0] ON [x0].[QueryRunId] = [x.QueryRun0].[Id]
    INNER JOIN [QueryDefinition] AS [x.QueryRun.QueryDefinition0] ON [x.QueryRun0].[QueryDefinitionId] = [x.QueryRun.QueryDefinition0].[Id]
    WHERE [x.QueryRun.QueryDefinition0].[UserId] = @__userId_0
) AS [t]
ORDER BY [t].[CreationDate] DESC
```

Expected SQL: 

```
SELECT [t].[Id], [t].[Source], [t].[ForeignId], [t].[CreationDate], [t].[LastUpdate], [t].[RemovalDate], [t].[Details], [t].[IsRemoved]
FROM (
    SELECT DISTINCT [x.Property0].*
    FROM [QueryRunResult] AS [x0]
    INNER JOIN [Property] AS [x.Property0] ON [x0].[PropertyId] = [x.Property0].[Id]
    INNER JOIN [QueryRun] AS [x.QueryRun0] ON [x0].[QueryRunId] = [x.QueryRun0].[Id]
    INNER JOIN [QueryDefinition] AS [x.QueryRun.QueryDefinition0] ON [x.QueryRun0].[QueryDefinitionId] = [x.QueryRun.QueryDefinition0].[Id]
    WHERE [x.QueryRun.QueryDefinition0].[UserId] = @__userId_0
) AS [t]
ORDER BY [t].[CreationDate] DESC
```

Exception: 

Invalid column name 'Source'.
[...]

PR #6784, opened 2016-10-14
Query: Keep track of projected querysource in subquery to apply proper select *

### LINKED ISSUE #6657: Query: Subquery projection being selected incorrectly with required navigation filter
### Steps to reproduce
Full example

[EfCoreJoinIssue.zip](https://github.com/aspnet/EntityFramework/files/504063/EfCoreJoinIssue.zip)

### The issue
Running

    IQueryable<Post> posts = db.Posts;
            posts = posts.Where(x => x.Author.UserName == "myuser");
            posts = posts.Take(5);

            var query = (from post in posts
                         select new { post.Slug, C = post.Comments.Count() });

Generates a query that joins using the wrong columns(or alias).

If we move Take(5) to the end, or remove the Where. It works:

IQueryable<Post> posts = db.Posts;
            posts = posts.Where(x => x.Author.UserName == "myuser");
            var query = (from post in posts
                         select new { post.Slug, C = post.Comments.Count() }).Take(5);

Or:

IQueryable<Post> posts = db.Posts;
            posts = posts.Take(5);

            var query = (from post in posts
                         select new { post.Slug, C = post.Comments.Count() });

smitpatel:
Fixes #6657 
Issue: The problem here is the querysource does not need materialization so while visiting subquery the select expression we generate has no projections. Since lift subquery, the inner select expression converts from `SELECT 1` to `SELECT *`. At present we have a arbitrary logic to pick last table from select expression when there is project star. This breaks down if the projection is coming from any other table.
Solution: While visiting subquerymodel we know that which qsre is being projected out but we don't add anything to projection due to materialization is not required. Based on this qsre, we can store table alias in SelectExpression as a reference that this is the table from which we may need columns later. When we push down the subquery which converts it to project star, the subquery needs to use the table alias stored before so that we have required columns from subquery.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
@@ -139,9 +139,17 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
                 if (sqlExpression == null)
                 {
-                    if (!(node is QuerySourceReferenceExpression))
+                    var qsre = node as QuerySourceReferenceExpression;
+                    if (qsre == null)
                     {
                         QueryModelVisitor.RequiresClientProjection = true;
+                    }
+                    else
+                    {
+                        if (QueryModelVisitor.ParentQueryModelVisitor != null)
+                        {
+                            selectExpression.SetTableForProjectStar(qsre.ReferencedQuerySource);
+                        }
                     }
                 }
                 else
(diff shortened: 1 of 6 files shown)


maumar:
:shipit:

smitpatel:
Also fixes #6380

smitpatel:
@anpete - Can you take a look at this one?

anpete:
Nice @smitpatel. Two questions:

1) Do we ever need columns from multiple tables in the subquery? E.g. t1.*, t2.*
2) I think we want to remove the special case of star projection handling. I.e. A star projector is just an AliasExpression added to the SelectExpression.Projection. Would that be easy to do as part of this change?

smitpatel:
1. We may need columns from multiple tables in subquery but for all such cases which we have seen so far, we put specific projections. Till now we have picked up only 1 table when it is project star. This change just picks the correct table only. So case seems rare. Also if we need from multiple tables, we cannot do project star without aliasing so it would have to be tackled in a different way.

2. This change introduce a new property so removing it in future is breaking change so it would be better if this is done without but I think even with conversion to AliasExpression would be breaking change here. Lets discuss on that.

anpete:
:shipit:

divega:
@smitpatel please remember to file a bug for any additional improvement we want to do to this fix.

smitpatel:
Filed #6801

smitpatel:
Merged via 65f96c6db5d17f59de82d160edc8a91d3602b1ac and 9e5cfdc1f2f74a86428481e84f8404f4cec76088

Issue #6714, opened 2016-10-07
API review for 1.1


ajcvickers:
We should review APIs, paying special attention to any breaking changes. In particular, one thing we should look at are base classes used by provider-specific service implementations. See #6622 for details.

rowanmiller:
Done for preview1, will do an incremental review for 1.1 RTM

rowanmiller:
Done. There were only two API changes between Preview1 and RTM. The changes were in the lower level query components and I synced with relevant folks to make sure they were good.

PR #6798, opened 2016-10-17
RevEng: Revert breaking change in RelationalScaffoldingModelFactory

bricelam:
Part of #6714

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Design/RelationalScaffoldingModelFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Design/RelationalScaffoldingModelFactory.cs
@@ -37,17 +37,17 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
         private readonly HashSet<ColumnModel> _unmappedColumns = new HashSet<ColumnModel>();
 
         public RelationalScaffoldingModelFactory(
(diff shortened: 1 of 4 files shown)


src/Microsoft.EntityFrameworkCore.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -37,17 +37,17 @@ public class RelationalScaffoldingModelFactory : IScaffoldingModelFactory
...
             [NotNull] CandidateNamingService candidateNamingService)
         {
-            Check.NotNull(logger, nameof(logger));
+            Check.NotNull(loggerFactory, nameof(loggerFactory));
             Check.NotNull(typeMapper, nameof(typeMapper));
             Check.NotNull(databaseModelFactory, nameof(databaseModelFactory));
             Check.NotNull(candidateNamingService, nameof(candidateNamingService));
 
-            Logger = logger;
+            Logger = loggerFactory.CreateLogger<RelationalScaffoldingModelFactory>();
```
bricelam: Should this be `loggerFactory.CreateLogger(GetType())`?

ajcvickers: I think it is fine like this. This is the code I have.

PR #6800, opened 2016-10-17
API review changes

ajcvickers:
Part of issue #6714

* Rename GetField to GetFieldName
* Rename navigationPropertyName to propertyName
* Update and consolidate names of design logger Ids
* Fix enum break in RelationalEventId
* Add CurrentTransaction back to IRelationalConnection (binary compat)

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer.Design/Internal/SqlServerDatabaseModelFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer.Design/Internal/SqlServerDatabaseModelFactory.cs
@@ -224,7 +224,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
                     };
 
                     Logger.LogDebug(
-                        SqlServerDesignEventId.FoundSequence,
+                        RelationalDesignEventId.FoundSequence,
                         () => SqlServerDesignStrings.FoundSequence(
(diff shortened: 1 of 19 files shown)


src/Microsoft.EntityFrameworkCore.SqlServer.Design/Internal/SqlServerDatabaseModelFactory.cs
```diff
@@ -361,23 +361,23 @@ ELSE null
...
+                        RelationalDesignEventId.FoundColumn,
                         () => SqlServerDesignStrings.FoundColumn(
                             schemaName, tableName, columnName, dataTypeName, ordinal, nullable,
                             primaryKeyOrdinal, defaultValue, computedValue, precision, scale, maxLength, isIdentity, isComputed));
 
                     if (!_tableSelectionSet.Allows(schemaName, tableName))
                     {
                         Logger.LogDebug(
-                            SqlServerDesignEventId.ColumnNotInSelectionSet,
+                            RelationalDesignEventId.ColumnSkipped,
```
divega: OT: I am likely missing something, but I am a bit surprised that we don't short circuit earlier on `!_tableSelectionSet.Allows(schemaName, tableName)`, i.e. do we ever hit the condition that a column is skipped for any other reason than we are ignoring the whole table? If not, then what value does this add in the log? cc @lajones

Issue #6801, opened 2016-10-17
Query: Better handling of Star Projection


smitpatel:
While fixing issue #6657 in #6784 , we introduced a new property on `SelectExpression` to keep track of table to be used when doing star projection. We already have property to indicate that it is a star projection `IsProjectStar`. As an alternative to above system, we can add an AliasExpression or ColumnExpression which has name `*` and use it to denote, star projection.
Major benefit of this would be more consistent select expression as at present we set `IsProjectStar` when we need it but we don't reset it if something changes. With this change we would remove it from projections when it is not needed. Also remove special case properties from SelectExpression.

popcatalin81:
> As an alternative to above system, we can add an AliasExpression or ColumnExpression which has name * and use it to denote, star projection.

Please don't forget about multi start queries: `SELECT A.*, C.* FROM A .. JOIN B ... JOIN C`, in which case column name * seems insufficient?

smitpatel:
This approach does not work quite well.
Main issue is, we cannot have Star projection and normal projection in same Select since we cannot exactly say offset of normal projections in returned columns. Therefore if we add something in projection we need to remove the star projection. At many places we process part of query model and add stuff into projection but we need to remove it later if we have to do client eval. Having a logic that, adding something into projection can affect (remove) other things in projection is more likely to cause bugs. Also it removes `IsProjectStar` so to check star projection it requires doing filter on projection which is not perf friendly.

Also tried approach that instead of having 2 public properties `IsProjectStar` and `ProjectStarAlias`, if we can combine them into one. But we need to store alias without knowing if star projection is required or not. Therefore we need at least some public method even if we store it inside a field. Also our `SelectExpression` is not coupled properly with querysource (which is needed to create column expression). For queries like include, we send multiple queries to database even though there is only one querysource, which makes things undefined.

So current implementation looks really alright instead of bringing in complexity and potential for bugs in future.

cc: @anpete

smitpatel:
> Please don't forget about multi start queries:  SELECT A.*, C.* FROM A .. JOIN B ... JOIN C , in which case column name * seems insufficient?

[...]

PR #6811, opened 2016-10-18
Add doc comments for ProjectStarAlias & related method in SelectExpression

smitpatel:
Doc comments which were missed in #6784

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -79,8 +79,18 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
         /// </value>
         public virtual Expression Predicate { get; [param: CanBeNull] set; }
 
+        /// <summary>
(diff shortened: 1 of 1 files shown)


divega:
Talked to @smitpatel. We will probably just remove the members as part of fixing https://github.com/aspnet/EntityFramework/issues/6801.

smitpatel:
@divega, @rowanmiller - Ping! Since #6801 is not gonna be fixed.

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -79,8 +79,18 @@ public class SelectExpression : TableExpressionBase
...
+        /// <summary>
+        ///     Gets or sets the table alias to be used for star projection.
+        /// </summary>
+        /// <value>
+        ///     The table alias.
+        /// </value>
         public virtual string ProjectStarAlias { get; [param: CanBeNull] set; }
 
+        /// <summary>
+        ///     Sets table alias for star projection which corresponds to given query source
```
rowanmiller: I think this would read better as "Sets the table alias for the star projection..."

PR #6813, opened 2016-10-19
Adds overloads that use TimeSpan instead of Int32.

### LINKED ISSUE #6614: Database.SetCommandTimeout should take a `TimeSpan`, not an `Int32`
`Int32` is ambiguous (milliseconds, seconds, minutes?) and shouldn't be used as a type when you really want a duration.  Instead, `TimeSpan` should be used.

https://github.com/aspnet/EntityFramework/blob/master/src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs#L156

I don't mind submitting a PR for this, though I'm not sure what the proper deprecation strategy is for this project.  Create an overload and somehow annotate the old one?  Leave the old one and new one side-by-side (personally not a fan of this one)?  A doc comment on the old one indicating it is deprecated?  Delete the old one and accept the breaking change?

https://github.com/aspnet/EntityFramework/blob/master/src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs#L159 should similarly return a `TimeSpan`, not an `Int32`.

MicahZoltu:
I tried to follow the conventions elsewhere in the file, but feel free to let me know if I have not followed code style correctly.

The `ArgumentOutOfRangeException` seemed like the best way to handle the fact that `TimeSpan`s aren't really bounded, though with the new the API surface area the upper bound constraint seems arbitrary.  However, in most real world use cases it will probably not be hit and it is documented.  I decided to throw on negative number as well because that doesn't make sense.  The overload doesn't accept null because if you want to pass null explicitly you can simply call the other overload (transparent to the user) and if you have a `TimeSpan?` you are doing it wrong.  ;)

I added two tests, but I'll admit that they are pretty weak.  I didn't see any tests for the original overloads and I'm not sure how to author a test that will correctly spin up a `DatabaseFacade` where the `SetCommandTimeout` will actually succeed.  Feedback is welcome on how I can make a positive test for timeout setting.

The overload that returns a `TimeSpan` is likely a bit contentious, as is always the case when you have an "overload" that differs only by return type.  I can pull it out if desired, I added it mainly for consistency but I admit that it feels quite awkward.

Fixes #6614

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
@@ -211,6 +211,18 @@ namespace Microsoft.EntityFrameworkCore
 
         public static void SetCommandTimeout([NotNull] this DatabaseFacade databaseFacade, int? timeout)
             => GetRelationalConnection(databaseFacade).CommandTimeout = timeout;
+        public static void SetCommandTimeout([NotNull] this DatabaseFacade databaseFacade, TimeSpan timeout)
+        {
+            if (timeout < TimeSpan.Zero)
+            {
+                throw new ArgumentException($"Timeout must be greater than or equal to zero.  Provided: {timeout.TotalSeconds} seconds.");
+            }
+            if (timeout.TotalSeconds > Int32.MaxValue)
+            {
+                throw new ArgumentException($"Timeout must be less than or equal to Int32.MaxValue (2147483647) seconds.  Provided: {timeout.Seconds} seconds.");
+            }
+            SetCommandTimeout(databaseFacade, Convert.ToInt32(timeout.TotalSeconds));
+        }
 
         public static int? GetCommandTimeout([NotNull] this DatabaseFacade databaseFacade)
             => GetRelationalConnection(databaseFacade).CommandTimeout;
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.Tests/CommandConfigurationTests.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.Tests/CommandConfigurationTests.cs
@@ -24,6 +24,9 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.Tests
 
                     context.Database.SetCommandTimeout(null);
                     Assert.Null(context.Database.GetCommandTimeout());
+
+                    context.Database.SetCommandTimeout(TimeSpan.FromSeconds(66));
+                    Assert.Equal(66, context.Database.GetCommandTimeout());
                 }
             }
 
@@ -41,9 +44,16 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.Tests
 
                     Assert.Throws<ArgumentException>(
                         () => context.Database.SetCommandTimeout(-3));
+                    Assert.Throws<ArgumentException>(
+                        () => context.Database.SetCommandTimeout(TimeSpan.FromSeconds(-3)));
 
                     Assert.Throws<ArgumentException>(
                         () => context.Database.SetCommandTimeout(-99));
+                    Assert.Throws<ArgumentException>(
+                        () => context.Database.SetCommandTimeout(TimeSpan.FromSeconds(-99)));
+
+                    Assert.Throws<ArgumentException>(
+                        () => context.Database.SetCommandTimeout(TimeSpan.FromSeconds(UInt32.MaxValue)));
                 }
             }
 


src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -211,9 +211,29 @@ public static IDbContextTransaction BeginTransaction([NotNull] this DatabaseFaca
...
+        {
+            if (timeout < TimeSpan.Zero)
+            {
+                throw new ArgumentOutOfRangeException($"Timeout must be greater than or equal to zero.  Provided: {timeout.TotalMilliseconds} milliseconds.");
+            }
+            if (timeout.TotalMilliseconds > Int32.MaxValue)
+            {
+                throw new ArgumentOutOfRangeException($"Timeout must be less than or equal to Int32.MaxValue (2147483647) milliseconds.  Provided: {timeout.TotalMilliseconds} milliseconds.");
+            }
+            SetCommandTimeout(databaseFacade, Convert.ToInt32(timeout.TotalMilliseconds));
```
smitpatel: Should there be factor of 1000 somewhere since timeout in SqlConnection is in seconds rather than milliseconds.

MicahZoltu: If it is in Seconds I'll change it to `TotalSeconds`.  The whole reason I am creating this PR is because I have no idea what units `timeout` is actually in.  😃

MicahZoltu:
I recently changed my GitHub username from `Zoltu` to `MicahZoltu` (I created a new GitHub Organization named `Zoltu`).  I have previously done the whole CLA thing, but it appears that the bot didn't follow my name change.  What is the correct way to go about resolving that?

Side-note: The bot should really associate with users by ID, not by name.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
smitpatel: 👍 
Here is more info on ConnectionTimeout in SqlConnection.
https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlconnection.connectiontimeout(v=vs.110).aspx

smitpatel:
@MicahZoltu - Tests for original overload are here
https://github.com/aspnet/EntityFramework/blob/c9b5916a4e0fff763febea00da793ebd4113632d/test/Microsoft.EntityFrameworkCore.SqlServer.Tests/CommandConfigurationTests.cs

You can add tests in the same file which uses different way to configure them and verify if they are set correctly.

MicahZoltu:
I moved the tests into `CommandConfigurationTests.cs` (additional asserts in existing tests).  I also changed the exception thrown to `ArgumentException` to match the existing exception type.

I also made the return value of `GetCommandTimeoutTimeSpan` nullable to match the existing behavior (and address a bug in null handling).

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -211,9 +211,36 @@ public static IDbContextTransaction BeginTransaction([NotNull] this DatabaseFaca
...
+        /// <exception cref="ArgumentOutOfRangeException">If the provided timeout is greater than Int32.MaxValue (seconds) or less than 0.</exception>
+        public static void SetCommandTimeout([NotNull] this DatabaseFacade databaseFacade, TimeSpan timeout)
+        {
+            if (timeout < TimeSpan.Zero)
+            {
+                throw new ArgumentException($"Timeout must be greater than or equal to zero.  Provided: {timeout.TotalSeconds} seconds.");
+            }
+            if (timeout.TotalSeconds > Int32.MaxValue)
+            {
+                throw new ArgumentException($"Timeout must be less than or equal to Int32.MaxValue (2147483647) milliseconds.  Provided: {timeout.Seconds} seconds.");
```
smitpatel: not `milliseconds` in message.

smitpatel:
@rowanmiller - Can you scan this for API review? Milestone?

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -211,9 +211,36 @@ public static IDbContextTransaction BeginTransaction([NotNull] this DatabaseFaca
 
         public static void SetCommandTimeout([NotNull] this DatabaseFacade databaseFacade, int? timeout)
             => GetRelationalConnection(databaseFacade).CommandTimeout = timeout;
+        /// <summary>
+        /// Sets the timeout for commands run against the database facade.
+        /// </summary>
+        /// <param name="databaseFacade">The database facade to set the timeout on.</param>
+        /// <param name="timeout">The time to wait before the command is aborted.</param>
+        /// <exception cref="ArgumentOutOfRangeException">If the provided timeout is greater than Int32.MaxValue (seconds) or less than 0.</exception>
+        public static void SetCommandTimeout([NotNull] this DatabaseFacade databaseFacade, TimeSpan timeout)
```
MicahZoltu: Fixed.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -211,9 +211,36 @@ public static IDbContextTransaction BeginTransaction([NotNull] this DatabaseFaca
...
+            if (timeout.TotalSeconds > Int32.MaxValue)
+            {
+                throw new ArgumentException($"Timeout must be less than or equal to Int32.MaxValue (2147483647) seconds.  Provided: {timeout.Seconds} seconds.");
+            }
+            SetCommandTimeout(databaseFacade, Convert.ToInt32(timeout.TotalSeconds));
+        }
 
         public static int? GetCommandTimeout([NotNull] this DatabaseFacade databaseFacade)
             => GetRelationalConnection(databaseFacade).CommandTimeout;
+        public static TimeSpan? GetCommandTimeoutTimeSpan([NotNull] this DatabaseFacade databaseFacade)
```
smitpatel: @rowanmiller - Did we take any decision on this?

MicahZoltu: The only reason I can think of where I would want to get this value (rather than just set it) is if I were building a generalized monitoring library.  Since my main concern is getting the `TimeSpan` setter, I'm willing to ignore the monitoring story in an effort to move the rest of this PR forward.

MicahZoltu: Removed.  However, I did it from the GitHub GUI since I'm not at my home computer, if the build doesn't pass I may have to wait until this evening to test it locally.

rowanmiller:
@smitpatel this is good to go from my perspective.

@MicahZoltu sorry that I held this up, it had dropped of my radar of things to do.

smitpatel:
@MicahZoltu - Can you squash all commits and rebase on dev?

MicahZoltu:
Sure, though it'll have to wait until this evening when I'm at my computer.

MicahZoltu:
Rebased and squashed.

rowanmiller:
@smitpatel can you get this one merged now

smitpatel:
Merged in 3608b72f59ff110d75672569ea84e82288de6ab8
Thanks for contribution @MicahZoltu

PR #6819, opened 2016-10-19
Query: Fix #6737 - Entities not being released

### LINKED ISSUE #6737: Entities not being released
Apologies if this has been mentioned before:

The problem I am seeing, is that entities are still being held in memory, even after the Context has been disposed.

I have tried the following:

``` c#
private void Method() {
     Test();
}

 private void Test()
 {
            using (var Context = new ExampleEntities())
            {
                Context.ExampleTable.ToList();  
          }
}
```

When you view the memory heap after the method Test has completed, ExampleTable rows are still held in memory.

anpete:
- Fixed two places where compiled queries were inadvertently holding onto scoped QueryCompiler state:

1) The query expression used in the compiled query cache key via EntityQueryable<> constant expressions.
2) The compiled query delegate, via a capturing closure in QueryCompiler itself.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -13,6 +13,7 @@ using Xunit.Abstractions;
 #if NETCOREAPP1_1
 using System.Threading;
 #endif
+
 namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests
 {
     public class QuerySqlServerTest : QueryTestBase<NorthwindQuerySqlServerFixture>
@@ -21,6 +22,27 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests
             : base(fixture)
         {
             //TestSqlLoggerFactory.CaptureOutput(testOutputHelper);
+        }
+
+        [ConditionalFact]
+        public virtual void Cache_key_expressions_are_detached()
+        {
+            WeakReference wr;
+            MakeGarbage(CreateContext(), out wr);
+
+            GC.Collect();
+
+            Assert.False(wr.IsAlive);
+        }
+
+        private static void MakeGarbage(NorthwindContext context, out WeakReference wr)
+        {
+            using (context)
+            {
+                wr = new WeakReference(context.Customers.First());
+
+                Assert.True(wr.IsAlive);
+            }
         }
 
         public override void Project_to_object_array()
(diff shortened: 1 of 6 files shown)


anpete:
@ajcvickers

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -23,6 +24,27 @@ public QuerySqlServerTest(NorthwindQuerySqlServerFixture fixture, ITestOutputHel
             //TestSqlLoggerFactory.CaptureOutput(testOutputHelper);
         }
 
+        [ConditionalFact]
+        public virtual void Cache_key_expressions_are_detached()
+        {
+            WeakReference wr;
+            MakeGarbage(CreateContext(), out wr);
+
+            GC.Collect();
```
ajcvickers: Is this reliable enough? In some situations a single collect is not enough.

divega: Could be an issue if there were pending finalizers involved.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -23,6 +24,27 @@ public QuerySqlServerTest(NorthwindQuerySqlServerFixture fixture, ITestOutputHel
...
+
+            Assert.False(wr.IsAlive);
+        }
+
+        private static void MakeGarbage(NorthwindContext context, out WeakReference wr)
+        {
+            using (context)
+            {
+                wr = new WeakReference(context.Customers.First());
```
divega: @anpete I think I asked you yesterday but I don't remember the answer. What are the chances that different query operators from `First()` could introduce additional references to scoped services or state?

anpete: @divega We should be OK here. We are careful not to explicitly introduce such dependencies in the query pipeline itself. The two being fixed here are in a sense "outside" of the main pipeline, which is why we missed them.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
anpete: Was wondering this. I haven't seen any flakiness yet and so was going to start with the simple thing.

PR #6834, opened 2016-10-21
Preserve temporary flag for store generated values until AcceptChanges

ajcvickers:
Issue #6720

The fix is to stop marking the value as permanent as soon as the store generated value is generated, but instead weight until AcceptChanges when it is pushed into the entity. Also, when DiscardStoreGeneratedValues is used, a temporary value is ensured to stay temporary.


anpete:
:shipit:

PR #6838, opened 2016-10-21
Fix to #6451 - Query :: Include doesn't work with SelectMany

maumar:
Problem was that in several places we transformed query model but we were not updating the (include) annotations accordingly. This resulted in annotations going out-of-sync with the actual query model which then resulted in us not being able to find the right query source to run include on.

Fix is to update query annotations where necessary (nav rewrite, select many rewrite, join elimination).

There was additional problem with string based include - we were always assuming that the include operator was called directly on a DbSet, so string based include didn't work for cases like:

customers.SelectMany(c => c.Orders).Include("...");
orders.Select(o => o.Customer).Include("...");

Fix is to add additional metadata to the IncludeResultOperator that describes the path from the initial query source (i.e. DbSet) to the actual include method. Also refactored expression based include in similar fashion - split old NavigationPropertyPath into two expressions: PathFromQuerySource (the same as with string based include) and lambda describing properties that are being included.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/ResultOperators/Internal/IncludeExpressionNode.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ResultOperators/Internal/IncludeExpressionNode.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using JetBrains.Annotations;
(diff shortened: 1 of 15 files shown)


src/Microsoft.EntityFrameworkCore/Query/ResultOperators/Internal/IncludeExpressionNode.cs
```diff
@@ -47,19 +47,17 @@ public class IncludeExpressionNode : ResultOperatorExpressionNodeBase
         /// </summary>
         protected override ResultOperatorBase CreateResultOperator(ClauseGenerationContext clauseGenerationContext)
         {
-            var navigationPropertyPath
-                = Source.Resolve(
-                    _navigationPropertyPathLambda.Parameters[0],
-                    _navigationPropertyPathLambda.Body,
-                    clauseGenerationContext) as MemberExpression;
+            var prm = Expression.Parameter(typeof(object));
+            var pathFromQuerySource = Resolve(prm, prm, clauseGenerationContext);
```
maumar: this seems bit hacky, but don't know how to do it better

maumar:
@anpete @smitpatel

smitpatel:
Does include work with select many or not? Title is so misleading... :trollface:

maumar:
Bug was complaining that include didn't work . Fix to this bug is to make it work:)

maumar:
@anpete new version up, refactored lambda based include and added some extra tests

anpete:
:shipit:

maumar:
Merged in 93ecfeca38c45cfc3081756e0a65920db19b42a0

PR #6839, opened 2016-10-22
Query: Skip/Take with Any/All should preserve skip/take

### LINKED ISSUE #6710: LINQ Any() statement generates invalid SQL
### Steps to reproduce

Models:

    public class Blog
    {

        public List<Post> Posts { get; set; }
    }

    public class Post
    {
        public string Content { get; set; }

    }

### The issue

LINQ queries using `.Any()` are causing an incorrect SQL `EXISTS` query to be executed:

    db.Blogs.Include(m => m.Posts).OrderByDescending(m => m.Id).Skip(5).Take(10).Any()
    //translates to SQL:
    SELECT CASE WHEN EXISTS ( SELECT 1 FROM `Blogs` AS `m`) THEN TRUE ELSE FALSE END LIMIT 10 OFFSET 5

I'd expect a correct `EXISTS` statement to look more like the query in the `COUNT` statement:

	db.Blogs.Include(m => m.Posts).OrderByDescending(m => m.Id).Skip(5).Take(10).Count()
	//translates to SQL:
	SELECT COUNT(*)
	FROM (
	    SELECT `m`.`Id`, `m`.`Title`
	    FROM `Blogs` AS `m`
	    ORDER BY `m`.`Id` DESC
	    LIMIT 10 OFFSET 5
	) AS `t`

Therefore the correct `EXISTS` statement should be something like:

	SELECT CASE WHEN EXISTS (
	    SELECT 1
	    FROM `Blogs` AS `m`
	    ORDER BY `m`.`Id` DESC
	    LIMIT 10 OFFSET 5
	) THEN TRUE ELSE FALSE END

### Further technical details

EF Core version: 1.0.1
Operating system: Ubuntu 16.04

I am using the [Pomelo.EntityFramework.MySql](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql) driver (I am a contributor to the project).  I believe this issue is present in the `Microsoft.EntityFrameworkCore.Relational` package though and may affect all drivers.

smitpatel:
Fore `Any()` or `All()` we created new `SelectExpression` for inner subquery but we should clone it so that we preserve all limit, offset & orderby from it.

Fixes #6710 

Should this go to 1.2.0 or dev?

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -301,9 +301,10 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             [param: CanBeNull]
             set
             {
-                Check.NotNull(value, nameof(value));
-
-                PushDownIfLimit();
+                if (value != null)
+                {
+                    PushDownIfLimit();
+                }
 
                 _limit = value;
             }
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
@@ -173,7 +173,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
         private static Expression HandleAll(HandlerContext handlerContext)
         {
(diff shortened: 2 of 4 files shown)


anpete:
:shipit:

smitpatel:
@anpete - 1.1 or 1.2?

anpete:
@smitpatel I cleared the milestone on the bug to discuss. You can put it into 1.2 for now or wait :-)

PR #6843, opened 2016-10-22
Trivial refactor to make ORDER BY more extensible

roji:
DefaultQuerySqlGenerator's GenerateOrderBy is a single block that's difficult to extend. Moved the generation of the individual ordering clauses into a separate function to allow appending to them.

This is necessary to support PostgreSQL ORDER BY ... NULLS FIRST.

I hope this can make it into 1.1, it's really very trivial.


smitpatel:
@anpete, @divega @rowanmiller - 1.1 milestone?

anpete:
@smitpatel Fine by me.

PR #6845, opened 2016-10-23
Fix SQL Server value generation from model issue

roji:
When deciding which value generation to use, the SQL Server provider falls back to the model annotation. However, it did not properly check that the property's CLR type is compatible with the model-determined strategy. As a result, defining a ValueGeneratedOnAdd decimal on a model with strategy HiLo triggered a failure.


ajcvickers:
Would it instead be possible to fix HiLo to work with decimal?

roji:
Hmm... There's another check in `SqlServerPropertyAnnotations.SetValueGenerationStrategy` that isn't done in `GetSqlServerValueGenerationStrategy` - byte and byte? types can't be identity columns... So unifying the checks between these two functions seems to make sense regardless of adding decimal hilo support (which is definitely possible)

PR #6851, opened 2016-10-24
Query: Set table alias for star projection properly for all cases

### LINKED ISSUE #6785: Query: Track proper table alias for star projection
Whenever we generate select expression with star projection we pick last table which can be incorrect (like #6657). In #6784 , we are adding another property in SelectExpression which keeps track of the table alias for star projection. We should use the property to actually keep track of which table to project rather than choosing last one.
https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs#L210

smitpatel:
Resolves #6785 

Set table for project star wherever we generate subquery which needs project star.
~~Slight change in API for `SetTableForProjectStar` since we don't have query source perfectly matched with table alias.~~
Nuked the function to set project star. Direct setter on the property since we need to store string alias only.


anpete:
:shipit:

PR #6856, opened 2016-10-25
Don't attempt to null out PK parts of composite FK during fixup

ajcvickers:
Because this either throws because the PK property cannot be set to null, or marks the property as conceptually null which then causes issues later.

There were also two issues in the tests:
- Equals method of one of the entities was wrong
- Tests wouldn't run on SQLite because composite keys with store generated ints are not supported

Issue #6840

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/NavigationFixer.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/NavigationFixer.cs
@@ -739,7 +739,9 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
 
                 if (!StructuralComparisons.StructuralEqualityComparer.Equals(
                     dependentEntry[dependentProperty],
-                    principalValue))
+                    principalValue)
+                    || (dependentEntry.IsConceptualNull(dependentProperty)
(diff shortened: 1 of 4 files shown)


AndriySvyryd:
Oops

src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/NavigationFixer.cs
```diff
@@ -767,14 +770,23 @@ private void DelayedFixup(InternalEntityEntry entry, INavigation navigation, Int
...
                 }
             }
 
             for (var i = 0; i < foreignKey.Properties.Count; i++)
             {
-                dependentEntry[dependentProperties[i]] = null;
-                dependentEntry.StateManager.UpdateDependentMap(dependentEntry, foreignKey);
-                dependentEntry.SetRelationshipSnapshotValue(dependentProperties[i], null);
+                if (!dependentProperties[i].IsKey()
+                    || !hasNonKeyProperties)
```
AndriySvyryd: Swap order for a marginal perf gain

ajcvickers: Done

PR #6858, opened 2016-10-25
Query: Minor perf optimizations

anpete:
- Removed an allocation from the query execution code path. 
- Reverted hack introduced in fixing #4317 because it looks like the IX-Async issue has been resolved.


AndriySvyryd:
:shipit:

PR #6860, opened 2016-10-26
Lift properties from the derived types in addition to the current type when setting base type.

### LINKED ISSUE #6814: Exception says base and derived classes have same properties, but they don't
This issue can be reproduced by trying to create a database using this context and model:

	{
			: base( options ) { }

		public DbSet<A> As { get; set; }
	}

	public class A
	{
		public int ID { get; set; }
		public L M3 { get; set; }
		public IList<PBase> Ls { get; set; }
		public MT M2 { get; set; }
		public M M1 { get; set; }
	}

	public class C : OT { }

	public class MT
	{
		public int ID { get; set; }
		public IList<ST> STs { get; set; }
		public IList<OGT> OGs { get; set; }
	}

	public class M
	{
		public int ID { get; set; }
		public IList<S> Ss { get; set; }
	}

	public class ST : P
	{
		public C DC { get; set; }
	}

	public class OGT : P
	{
		public IList<OT> Os { get; set; }
	}

	public class S
	{
		public int ID { get; set; }
		public ST Type { get; set; }
	}

	public class OT : P { }

	public class L
	{
		public int ID { get; set; }
	}

	public class T : P
	{
		public P P { get; set; }
		public Q F { get; set; }
	}

	public class P : PBase { }

	public class Q : PBase { }

	public abstract class PBase
	{
		public int ID { get; set; }
		public string Stuff { get; set; }
	}

The exception text says, "The type 'OT' cannot have base type 'P' because both types include the properties: ID, Stuff."

AndriySvyryd:
Properly discover base types when they are added to the model after the derived types.
Improve the exception message for conflicting properties in the new base type.
Don't duplicate unused inherited properties that are being removed when setting a new base type. They will be discovered by conventions if they should be on the entity type.

Fixes #6814

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/DerivedTypeDiscoveryConvention.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/DerivedTypeDiscoveryConvention.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Linq;
+using System.Reflection;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 
 namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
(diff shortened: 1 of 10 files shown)


anpete:
:shipit:

Issue #6576, opened 2016-09-20
Query: issue warning for queries with paging operations (First, Take, Skip) without OrderBy


ericvb:
### Steps to reproduce

Download the source from the Automapper project (https://github.com/AutoMapper/AutoMapper).
Take the unittest ExpandCollections.cs from the project IntegrationTests.Net4. It is working with EF 6.
See the expressions and the sql generated.
Add the project IntegrationTests.NetCore from the attachment zip. It is the same project, but now for EF core.
[IntegrationTests.NetCore.zip](https://github.com/aspnet/EntityFramework/files/481719/IntegrationTests.NetCore.zip)

I sligthly adapted the original unit test.
[ExpandCollectionsEric.zip](https://github.com/aspnet/EntityFramework/files/481764/ExpandCollectionsEric.zip)
### The issue

For EF 6 we see one generated SQL query.

```
SELECT
    [Project1].[CourseId] AS[CourseId], 
    [Project1].[CourseName] AS[CourseName], 
    [Project1].[C3] AS[C1], 
    [Project1].[ContentId] AS[ContentId], 
    [Project1].[C1] AS[C2], 
    [Project1].[CategoryName] AS[CategoryName], 
    [Project1].[ShortCategory_CategoryId] AS[ShortCategory_CategoryId], 
    [Project1].[C2] AS[C3], 
    [Project1].[CategoryName1] AS[CategoryName1], 
    [Project1].[Category_CategoryId] AS[Category_CategoryId], 
    [Project1].[ContentName] AS[ContentName]
FROM(SELECT
                 [Limit1].[CourseId] AS[CourseId],
                 [Limit1].[CourseName] AS[CourseName],
                 [Join2].[ContentId] AS[ContentId],
                 [Join2].[ContentName] AS[ContentName],
                 [Join2].[Category_CategoryId] AS[Category_CategoryId],
                 [Join2].[ShortCategory_CategoryId] AS[ShortCategory_CategoryId],
                 [Join2].[CategoryName1] AS[CategoryName],
                 [Join2].[CategoryName2] AS[CategoryName1],
                 CASE WHEN ([Join2].[ContentId] IS NULL) THEN CAST(NULL AS int) WHEN([Join2].[CategoryId1] IS NOT NULL) THEN 1 END AS[C1],
[...]

PR #6862, opened 2016-10-26
Fix to #6576 - Query: issue warning for queries with paging operations (First, Take, Skip) without OrderBy

maumar:
Performing paging operations without OrderBy may lead to unpredictable results and could be hard to track down. Fix is to issue a warning in the logger if such situation happens.


src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -659,6 +659,14 @@ public override void VisitQueryModel([NotNull] QueryModel queryModel)
         {
             Check.NotNull(queryModel, nameof(queryModel));
 
+            if (queryModel.ResultOperators.Any(o => o is FirstResultOperator || o is SkipResultOperator || o is TakeResultOperator)
+                && !queryModel.BodyClauses.OfType<OrderByClause>().Any())
+            {
+                QueryCompilationContext.Logger.LogWarning(
+                    CoreEventId.CompilingQueryModel,
```
maumar: should events be more granular?

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -659,6 +659,14 @@ public override void VisitQueryModel([NotNull] QueryModel queryModel)
...
+                    () => CoreStrings.PagingOperationWithoutOrderBy(queryModel.MainFromClause.ItemType.Name));
```
maumar: using MainFromClause is not always perfect - in some cases the type can be compiler generated which doesn't tell much to the user. Are we ok with that?

maumar:
@anpete @divega @rowanmiller

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
anpete: I think this is fine.

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -6371,6 +6371,34 @@ public virtual void DefaultIfEmpty_in_subquery_nested()
...
+        public virtual void Paging_operation_without_orderby_issues_warning_subquery()
+        {
+            using (var context = CreateContext())
+            {
+                context.Customers.Where(c => c.CustomerID == "ALFKI" && c.Orders.FirstOrDefault().OrderID > 1000).ToList();
+            }
+        }
+
+        [ConditionalFact]
+        public virtual void Single_SingleOrDefault_without_orderby_doesnt_issue_warning()
```
maumar: single will throw if there is more than 1 result and if its one result then its deterministic already

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
maumar: I tried that initially but it looked really bloated for a bit more complicated query. @divega was proposing that we could assign an alias to each query model and refer to them via alias in all the log messages. If we opted for it, it would be in different pull request.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -657,4 +657,7 @@
   <data name="PropertyCalledOnNavigation" xml:space="preserve">
     <value>Cannot call Property for the property '{property}' on entity type '{entityType}' because it is configured as a navigation property. Property can only be used to configure scalar properties.</value>
   </data>
+  <data name="PagingOperationWithoutOrderBy" xml:space="preserve">
+    <value>Query on '{querySourceType}' uses paging operation (First/FirstOrDefault/Skip/Take) without OrderBy which may lead to unpredictable results.</value>
```
maumar: good idea, will do

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
anpete: Makes sense.

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
maumar: sounds like a reasonable compromise

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
maumar: one problem is for cases like:
```
customers.Where(c => c.Orders.First().Id == 1)
```

This doesn't have a filter, but we add one during nav rewrite, so we won't issue warning for this case, even though we should

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
anpete: Consider extracting this into a protected virtual method.

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
anpete: Cool, we should make sure there are no newlines, too. We could just use QM.ToString(), which is single line formatted.

src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
anpete: Do it before nav. rewrite?

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
maumar: im adding no-formatting option to expression/qm printers. default tostring leaves full type names which "steal" way to much space sometimes

maumar:
@anpete @divega  new version up

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -360,6 +364,39 @@ var entityEqualityRewritingExpressionVisitor
...
+            }
+
+            public override void VisitQueryModel(QueryModel queryModel)
+            {
+                if (queryModel.ResultOperators.Any(o => o is SkipResultOperator || o is TakeResultOperator)
+                    && !queryModel.BodyClauses.OfType<OrderByClause>().Any())
+                {
+                    _logger.LogWarning(
+                        CoreEventId.CompilingQueryModel,
+                        () => CoreStrings.RowLimitingOperationWithoutOrderBy(queryModel.Print(true, QueryModelStringLengthLimit)));
```
anpete: Add Parameter name in Print

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
```diff
@@ -30,10 +30,10 @@ var customers
                 Assert.NotNull(customers);
                 Assert.StartsWith(
                     @"    Compiling query model: 
-'from Customer <generated>_0 in DbSet<Customer>
+'from Customer <generated>_0 in DbSet<Customer>
```
anpete: This extra space expected?

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RecursiveQueryModelVisitorBase.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Linq.Expressions;
+using Remotion.Linq.Clauses.Expressions;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public abstract class RecursiveQueryModelVisitorBase : ExpressionTransformingQueryModelVisitor<ExpressionVisitorBase>
```
anpete: Going to try and unify to TransformExpressions here

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
anpete: Is it hard to fix this extra space?

anpete:
:shipit:

PR #6870, opened 2016-10-26
Query: Copy OrderBy in PushDownSubquery

### LINKED ISSUE #6736: Include with OrderBy with Skip(0) cause OrderBy gets ignored
Use Include with OrderBy with Skip(0) will cause OrderBy gets ignored.
### Steps to reproduce

This is the standalone project for reproduce the problem
[EFExample.zip](https://github.com/aspnet/EntityFramework/files/518084/EFExample.zip)

The main code is as follows

```
using (var context = new ExampleContext())
{
    var authors = context.Set<Author>();
    Console.WriteLine("order by age");
    var result = authors
        .Include(a => a.Articles)
        .OrderBy(a => a.Age).Skip(0).Take(100).ToList();
    foreach (var author in result)
    {
        Console.WriteLine(author.Name);
    }
    Console.WriteLine();

    Console.WriteLine("order by age descending");
    result = authors
        .Include(a => a.Articles)
        .OrderByDescending(a => a.Age).Skip(0).Take(100).ToList();
    foreach (var author in result)
    {
        Console.WriteLine(author.Name);
    }
    Console.ReadLine();
}
```

The output is

```
order by age
Steve
John

order by age descending
Steve
John
```
### The issue

EF generated following SQL

```
exec sp_executesql N'SELECT [t].[Id], [t].[Age], [t].[CreateTime], [t].[Deleted], [t].[Name], [t].[UpdateTime]
FROM (
    SELECT [a].[Id], [a].[Age], [a].[CreateTime], [a].[Deleted], [a].[Name], [a].[UpdateTime], ROW_NUMBER() OVER(ORDER BY [a].[Age]) AS [__RowNumber__]
    FROM [Author] AS [a]
) AS [t]
WHERE ([t].[__RowNumber__] > @__p_0) AND ([t].[__RowNumber__] <= (@__p_0 + @__p_1))
ORDER BY [t].[Id]',N'@__p_0 int,@__p_1 int',@__p_0=0,@__p_1=100
go
```
### Further technical details

EF Core version: 1.0.1
Operating system: Windows 8.1

smitpatel:
Resolves #6736 
Issue for the specific case of #6736 is, when we do RowNumberPaging, we remove `OrderBy` clause. For a simple query, results are returned correctly but when we have include, we add ordering on PK property, which changes the order causing incorrect results.
For more general case, it can arise whenever we push down a subquery since outer query will not have any order by and Include can introduce one.

Fix would be to copy over the order by so that data is always in fixed order and will not be change. Also clear order by from a subquery level if there is not limit/offset.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -283,6 +283,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
                 if (_offset != null)
                 {
                     PushDownSubquery();
+                    ClearOrderBy();
                 }
 
                 _isDistinct = value;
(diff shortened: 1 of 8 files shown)


anpete:
:shipit:

PR #6874, opened 2016-10-26
Query: Refactor to unified SQL function representation.

anpete:
- Removes custom SQL nodes for Min, Max, Sum, Count and DatePart.
- Adds SqlFragmentExpression for easier extensibility.


anpete:
@maumar @smitpatel

PR #6898, opened 2016-10-29
fix(Migrations): better ordering when adding computed columns

### LINKED ISSUE #6853: Migrations: Attempt better ordering when adding computed columns
### Steps to reproduce

I have added some additional fields to ApplicationUser.

```
public class ApplicationUser : IdentityUser
{
    public string FirstName { get; set; }
    public string LastName { get; set; }
    public string FullName { get; set; }      
}
```

The `FullName` is computed property. So, in the context, I have:

```
 {
     builder.Entity<ApplicationUser>()
     .Property(p => p.FullName)
     .HasComputedColumnSql("[FirstName] + ' ' + [LastName]");

 }
```
### The issue

In result, I get the following migration file:

```
{
    migrationBuilder.AddColumn<string>(
            name: "FirstName",
            table: "AspNetUsers",
            nullable: true);

   migrationBuilder.AddColumn<string>(
            name: "FullName",
            table: "AspNetUsers",
            nullable: true,
            computedColumnSql: "[FirstName] + ' ' + [LastName]");

    migrationBuilder.AddColumn<string>(
           name: "LastName",
           table: "AspNetUsers",
           nullable: true);
}
```

As you can see, the computed proeprty is added before `LastName` and I get error: 
`Invalid column name 'LastName'.`

After moving the creation of the computed property to the end, updating database finished wuthout error.
### Further technical details

EF Core version: 1.0.1
Operating system: Windows 10
Visual Studio version: Visual Studio 2015 Community

laskoviymishka:
**Please check if the PR fulfills these requirements**
- [x] The code builds and tests pass (verified by our automated build checks)
- [x] Commit messages follow this format

```
    Modify sort to handle computed SQL add columns at last

    Fixes #6853
```
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Code meets the expectations our engineering guidelines. https://github.com/aspnet/Home/wiki/Engineering-guidelines.

Please review the guidelines for CONTRIBUTING.md for more details.
Fixes #6853

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
@@ -129,7 +129,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
             var createTableOperations = new List<CreateTableOperation>();
             var alterDatabaseOperations = new List<MigrationOperation>();
             var alterTableOperations = new List<MigrationOperation>();
-            var columnOperations = new List<MigrationOperation>();
+            var columnOperations = new List<ColumnOperation>();
             var alterOperations = new List<MigrationOperation>();
             var restartSequenceOperations = new List<MigrationOperation>();
             var constraintOperations = new List<MigrationOperation>();
@@ -178,7 +178,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
                 }
                 else if (_columnOperationTypes.Contains(type))
                 {
-                    columnOperations.Add(operation);
+                    columnOperations.Add(operation as ColumnOperation);
                 }
                 else if (_alterOperationTypes.Contains(type))
                 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -279,7 +279,8 @@ public virtual IReadOnlyList<MigrationOperation> GetDifferences(IModel source, I
                 .Concat(alterDatabaseOperations)
                 .Concat(createSequenceOperations)
                 .Concat(alterTableOperations)
-                .Concat(columnOperations)
+                .Concat(columnOperations.Where(t => string.IsNullOrWhiteSpace(t.ComputedColumnSql)))
+                .Concat(columnOperations.Where(t => !string.IsNullOrWhiteSpace(t.ComputedColumnSql)))
```
bricelam: The reason for the groups (`alterTableOperations`, `alterOperations`, etc.) is so we don't have to iterate over lists multiple times. It would be better to split `columnOperations` into two lists--`columnOperations` and `computedColumnOperations`. Then the `Where` methods aren't needed here because they're put in the right collection the first time we iterate through the operations.

laskoviymishka: @bricelam sounds reasonable, i'll do it.

bricelam: Cool, thanks for going the extra mile. :+1:

PR #6908, opened 2016-10-31
WIP(Translators): support for translating ToString() to sql server

### LINKED ISSUE #6264: Support for translating ToString() to SQL
### Steps to reproduce

Convert int to String only works in local (by Linq)

[User].Id.ToString().Contains("1")
### The issue

Who i can convert int to string (by the server)?

```
Exception message:
info: Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommandBuilderFactory[1]
      Executed DbCommand (9ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
      SELECT [User].[Id], [User].[Name], [User].[Email]
      FROM [Users] AS [User]
warn: Microsoft.EntityFrameworkCore.Query.Internal.SqlServerQueryCompilationContextFactory[8]
      The LINQ expression '(([User].Id.ToString().Contains("1") OrElse [User].Name.Contains("1")) OrElse [User].Email.Contains("1"))' could not be translated and will be evaluated locally. To configure this warning use the DbContextOptionsBuilder.ConfigureWarnings API (event id 'RelationalEventId.QueryClientEvaluationWarning'). ConfigureWarnings can be used when overriding the DbContext.OnConfiguring method or using AddDbContext on the application service provider.
```

I use this method

```
public Expression<Func<T, Boolean>> WhereForSearch<T>(string[] words)
        {
            Expression general = null;
            if (words == null || words.Length == 0)
            {
                general = Expression.Equal(Expression.Constant(1), Expression.Constant(1));
                return Expression.Lambda<Func<T, Boolean>>(general, new ParameterExpression[] { ParameterExpression });
            }

            Dictionary<Type, object>[] vals = new Dictionary<Type, object>[words.Length];

            for (int x = words.Length - 1; x >= 0; x--)
                vals[x] = ParseHelper.ConvertFromString(words[x], Listable.Types);

            for (int x = 0, m = words.Length; x < m; x++)
            {
                Dictionary<Type, object> val = vals[x];
                Expression pe = null;
                foreach (CacheLinqExpresion.Item p in Searchable.Properties)
                {
                    object v1;
                    // Check same type
                    if (!val.TryGetValue(p.Type, out v1)) continue;

                    // Convert to string
                    Expression ex = p.Value;
                    if (p.Type != ParseHelper.TypeString)
                    {
                        ///  x.ToString().Contains(v1)
                        v1 = words[x];
                        ex = Expression.Call(ex, CrudCacheProvider.MethodToString);
                        ex = Expression.Call(ex, CrudCacheProvider.MethodContains, Expression.Constant(v1));
                        //ex = Expression.Equal(ex, Expression.Constant(v1));
                    }
                    else
                    {
                        ex = Expression.Call(ex, CrudCacheProvider.MethodContains, Expression.Constant(v1));
                    }
                    pe = pe == null ? ex : Expression.OrElse(pe, ex);
                }

                if (pe == null) continue;

                general = (general == null) ? pe : Expression.AndAlso(general, pe);
            }

            return Expression.Lambda<Func<T, Boolean>>(general, new ParameterExpression[] { ParameterExpression });
        }
```
### Further technical details

EF Core version: (found in project.json or packages.config)
  "Microsoft.AspNetCore.Identity.EntityFrameworkCore": "1.0.0",
    "Microsoft.EntityFrameworkCore": "1.0.0",

laskoviymishka:
- [x] The code builds and tests pass (verified by our automated build checks)
- [x] Commit messages follow this format
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Code meets the expectations our engineering guidelines. 

```
Add new translator for to string in sql server translators. Add corresponding tests
```

Fixes #6264

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerObjectToStringTranslator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerObjectToStringTranslator.cs
@@ -29,11 +29,26 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
                     returnType: methodCallExpression.Type,
                     arguments: new[]
                     {
-                        new SqlFragmentExpression("VARCHAR(MAX)"),
+                        new SqlFragmentExpression(GetConvertBase(methodCallExpression.Object?.Type.Name)),
                         methodCallExpression.Object
                     });
             }
             return null;
+        }
+
+        private string GetConvertBase(string typeName)
+        {
+            switch (typeName)
+            {
+                case nameof(Int64):
+                    return "VARCHAR(20)";
+                case nameof(Int32):
+                    return "VARCHAR(10)";
+                case nameof(Int16):
+                    return "VARCHAR(5)";
+                default:
+                    return "VARCHAR(MAX)";
+            }
         }
     }
 }
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerObjectToStringTranslator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerObjectToStringTranslator.cs
@@ -3,7 +3,7 @@
 namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
 {
     using System;
-    using System.Linq;
+    using System.Collections.Generic;
     using System.Linq.Expressions;
     using Microsoft.EntityFrameworkCore.Query.Expressions;
 
@@ -13,6 +13,13 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
     /// <seealso cref="Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.IMethodCallTranslator" />
     public class SqlServerObjectToStringTranslator : IMethodCallTranslator
     {
+        private static readonly Dictionary<string, string> _typeMapping = new Dictionary<string, string>
+        {
+            [nameof(Int64)] = "VARCHAR(20)",
+            [nameof(Int32)] = "VARCHAR(10)",
+            [nameof(Int16)] = "VARCHAR(5)"
+        };
+
         /// <summary>
         ///     Translates the given method call expression.
         /// </summary>
@@ -29,7 +36,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
                     returnType: methodCallExpression.Type,
                     arguments: new[]
                     {
-                        new SqlFragmentExpression(GetConvertBase(methodCallExpression.Object?.Type.Name)),
+                        new SqlFragmentExpression(GetConvertBase(methodCallExpression.Object?.Type.UnwrapEnumType().Name)),
                         methodCallExpression.Object
                     });
             }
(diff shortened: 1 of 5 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -6277,6 +6277,29 @@ public override void Subquery_member_pushdown_does_not_change_original_subquery_
...
+                Sql);
+        }
+
+
+        public override void Select_expression_to_string()
+        {
+            base.Select_expression_to_string();
+
+            Assert.Equal(
+                @"SELECT CONVERT(VARCHAR(MAX), [o].[EmployeeID])
```
anpete: Does size matter here. I.e. for some expressions we know the column size from metadata.

cc @divega

laskoviymishka: @anpete i believe when we convert size is not really matter.

divega: It think it may matter more if it is in the projected results, e.g. `varchar(max)`/`nvarchar(max)` can perform poorly. In general what input types are we supporting with this translation? We probably need some special casing for different types to make the translation correct and once we are special casing it should be easy to pick adequate sizes.

laskoviymishka: @divega that sounds reasonable. I need to add more logic into conversion.

laskoviymishka:
@anpete please review. Add a little more intelegence into varchar count. Sorry for long replay.

I review a little bit possibility to implement DateTime conversion as part of this feature and come to conclusion that this would be to not good solution. Since date conversion in MSSQL quite different than .net one. I believe that some kind of extension method for this conversion would be helpfull:
Like
```
_db.Entity.Select(t => t.DateColumn.ToString(SomeDateConversionOptionEnum))
```

Which will be translated into 
```
SELECT CONVERT(varchar(10), [MyDateTimecolumn], 20) AS DateColumn
```

Where SomeDateConversionOptionEnum will be mapped to one of Date and Time Styles. (see [msdn description](https://msdn.microsoft.com/en-us/library/ms187928.aspx?f=255&MSPPError=-2147217396)


Any thoughts on this one?

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
anpete: @laskoviymishka I think this test needs to be updated.

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerObjectToStringTranslator.cs
```diff
@@ -0,0 +1,54 @@
...
+                        new SqlFragmentExpression(GetConvertBase(methodCallExpression.Object?.Type.Name)),
+                        methodCallExpression.Object
+                    });
+            }
+            return null;
+        }
+
+        private string GetConvertBase(string typeName)
+        {
+            switch (typeName)
```
anpete: Can we switch to using a Dictionary<Type, string> lookup for this?

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerObjectToStringTranslator.cs
```diff
@@ -0,0 +1,54 @@
...
+        public virtual Expression Translate(MethodCallExpression methodCallExpression)
+        {
+            if (methodCallExpression.Method.Name == nameof(ToString))
+            {
+                return new SqlFunctionExpression(
+                    functionName: "CONVERT",
+                    returnType: methodCallExpression.Type,
+                    arguments: new[]
+                    {
+                        new SqlFragmentExpression(GetConvertBase(methodCallExpression.Object?.Type.Name)),
```
anpete: What about when methodCallExpression.Object?.Type is nullable or an enum? We have helpers UnwrapNullableType and UnwrapEnum that you can use here.

anpete:
@laskoviymishka Ping...

laskoviymishka:
Sorry for long response :(. I'll fix it today

laskoviymishka:
@anpete fixed. please reveiw

anpete:
Merged, thanks!

Issue #6893, opened 2016-10-28
Allow navigation backing fields to be assign-compatible


ajcvickers:
Currently the field must be the exact type of the property, which is too restrictive.

PR #6909, opened 2016-10-31
Allow navigation backing fields to be assign-compatible

ajcvickers:
Issue #6893

This lets backing fields of, for example, `List<T>` be found and used for navigation properties of `IEnumerable<T>` while still allowing fields of `int?` to be found and used for properties of `int`.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/BackingFieldConvention.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/BackingFieldConvention.cs
@@ -71,22 +71,29 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
             var typeInfo = propertyType.GetTypeInfo();
 
             FieldInfo fieldInfo;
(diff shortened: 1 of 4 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/BackingFieldConvention.cs
```diff
@@ -72,21 +72,29 @@ protected virtual void Apply([NotNull] PropertyBase propertyBase)
 
             FieldInfo fieldInfo;
             return (fields.TryGetValue("<" + propertyName + ">k__BackingField", out fieldInfo)
-                    && fieldInfo.FieldType.GetTypeInfo().IsAssignableFrom(typeInfo))
+                    && typeInfo.IsAssignableFrom(fieldInfo.FieldType.GetTypeInfo()))
```
AndriySvyryd: In what cases is this not the same type?

ajcvickers: I will remove that check.

PR #6917, opened 2016-11-01
Fix to 6879 - ArgumentOutOfRangeException when using DefaultIfEmpty() on a collection navigation in conjunction with SelectMany

maumar:
Problem was that when processing SelectMany we were not taken into consideration that the source could be a subquery. This happens when query has an explicit result operator attached to the navigation inside SelectMany, like so:

from u in Users
from c in u.Claims.DefaultIfEmpty()
select u

Fix is to take this into account - recognize when we are dealing with subquery and translate it into GroupJoin. There could be some additional body clauses and/or result operators in the subquery and they need to be incorporated into the GroupJoin as well.
Refactored the logic that was handling the GroupJoin creation to accommodate for this case.

We effectively translate the original query into:

from u in Users
join c in Claims on u.Id equals c.UserId into _group
from c in _group.DefaultIfEmpty()
select u


maumar:
@anpete

maumar:
@anpete added more tests for the join with subquery

anpete:
:shipit:

Issue #6896, opened 2016-10-28
Query: Rewrite Principal key access to FK access only if FK entity is not projected out


smitpatel:
Test Required_navigation_take_required_navigation

```
from Level2 l2 in 
    (from Level3 l3 in DbSet<Level3>
    order by l3.OneToOne_Required_FK_Inverse.Id asc
    select l3.OneToOne_Required_FK_Inverse)
    .Take(__p_0)
select l2.OneToOne_Required_FK_Inverse.Name
```

Gets translated to this

```
from Level2 l2 in 
    (from Level3 l3 in DbSet<Level3>
    join Level2 l3.OneToOne_Required_FK_Inverse in DbSet<Level2>
    on Property(l3, "Level2_Required_Id") equals Property(l3.OneToOne_Required_FK_Inverse, "Id")
    order by Property(l3, "Level2_Required_Id") asc
    select l3.OneToOne_Required_FK_Inverse)
    .Take(__p_0)
join Level1 l2.OneToOne_Required_FK_Inverse in DbSet<Level1>
on Property(l2, "Level1_Required_Id") equals Property(l2.OneToOne_Required_FK_Inverse, "Id")
select l2.OneToOne_Required_FK_Inverse.Name
```

In the subquery, we are converting nav prop access to join. In select clause we are projecting out navigation (principal entity) so there is not much value in converting orderby to work on fk property instead of principal key property. It breaks copying order by on outer query.

maumar:
Fixed in 347e33e79e8722e7e5242fd4f357b9b4e653635e

PR #6925, opened 2016-11-02
Fix to #6896 - Query: Rewrite Principal key access to FK access only if FK entity is not projected out

maumar:
Problem was that during nav rewrite we sometimes can simplify navigation like: order.Customer.Id into order.CustomerId.
This however is problematic for some complex queries like so:

var query = context.LevelThree
    .Select(l3 => l3.OneToOne_Required_FK_Inverse)
    .OrderBy(l3 => l3.Id)
    .Take(10)
    .Select(l2 => l2.OneToOne_Required_FK_Inverse.Name);

In this case we try to lift the orderby to the outer query. If the orderby gets simplified the query becomes invalid, because now we try to order by Level3.FK_ID, but what we project out from the subquery is Level2.
Fix is to not perform this optimization for orderby inside a subquery if we project the entity on which property we try to order.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -39,6 +39,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         private bool _insideInnerSequence;
         private bool _innerKeySelectorRequiresNullRefProtection;
         private bool _insideInnerKeySelector;
+        private bool _insideOrderBy;
 
         private class NavigationJoin
         {
@@ -579,10 +580,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                         additionalFromClauseBeingProcessed);
                 }
 
-                if (navigations.Count == 1
-                    && navigations[0].IsDependentToPrincipal())
+                if (navigations.Count == 1 && navigations[0].IsDependentToPrincipal())
                 {
-                    var foreignKeyMemberAccess = CreateForeignKeyMemberAccess(propertyName, declaringExpression, navigations[0]);
+                    var foreignKeyMemberAccess = TryCreateForeignKeyMemberAccess(propertyName, declaringExpression, navigations[0]);
                     if (foreignKeyMemberAccess != null)
                     {
                         return foreignKeyMemberAccess;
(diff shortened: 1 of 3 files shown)


maumar:
@smitpatel

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
```diff
@@ -1835,8 +1833,35 @@ public override void Required_navigation_take_required_navigation()
...
+
+        public override void Required_navigation_order_by_and_select_no_subquery()
+        {
+            base.Required_navigation_order_by_and_select_no_subquery();
+
+            Assert.Equal(
+                @"SELECT [l3.OneToOne_Required_FK_Inverse].[Id], [l3.OneToOne_Required_FK_Inverse].[Date], [l3.OneToOne_Required_FK_Inverse].[Level1_Optional_Id], [l3.OneToOne_Required_FK_Inverse].[Level1_Required_Id], [l3.OneToOne_Required_FK_Inverse].[Name], [l3.OneToOne_Required_FK_Inverse].[OneToMany_Optional_InverseId], [l3.OneToOne_Required_FK_Inverse].[OneToMany_Optional_Self_InverseId], [l3.OneToOne_Required_FK_Inverse].[OneToMany_Required_InverseId], [l3.OneToOne_Required_FK_Inverse].[OneToMany_Required_Self_InverseId], [l3.OneToOne_Required_FK_Inverse].[OneToOne_Optional_PK_InverseId], [l3.OneToOne_Required_FK_Inverse].[OneToOne_Optional_SelfId]
+FROM [Level3] AS [l3]
+INNER JOIN [Level2] AS [l3.OneToOne_Required_FK_Inverse] ON [l3].[Level2_Required_Id] = [l3.OneToOne_Required_FK_Inverse].[Id]
+ORDER BY [l3].[Level2_Required_Id]",
```
smitpatel: This should be actually `[l3.OneToOne_Required_FK_Inverse].[Id]` since we are projecting out `[l3.OneToOne_Required_FK_Inverse]` columns.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -624,6 +624,106 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+
+        private Expression TryCreateForeignKeyMemberAccess(string propertyName, Expression declaringExpression, INavigation navigation)
+        {
+            var canPerformOptimization = true;
+            if (_insideOrderBy)
+            {
+                var qsre = (declaringExpression as MemberExpression)?.Expression as QuerySourceReferenceExpression;
+                if (qsre == null)
+                {
+                    var methodExpression = declaringExpression as MethodCallExpression;
```
smitpatel: nit: `methodCallExpression`

PR #6930, opened 2016-11-03
Query: :arrow_up: Upgrade Ix-Async Package

### LINKED ISSUE #6328: Consider upgrading to Ix-Async 3.1
The Rx team has just completed a near-rewrite of a majority of the Ix Async operators. The new pattern follows the System.Linq approach of using iterator objects to minimize allocations and recursion (in favor of state machines). https://github.com/Reactive-Extensions/Rx.NET/pull/237

To upgrade, to 3.1, we'll need to test the changes, possible remove internal workarounds, and investigate the effect of other changes in this update (notably: Ix 3.1 uses NETStandard.Library instead of granular BCL dependencies, which breaks from EF/ASP.NET's pattern of using granular dependencies. cref https://github.com/aspnet/Coherence-Signed/issues/362)

cc @onovotny @anpete

smitpatel:
resolves #6328 by removing our workarounds from `AsyncLinqQueryProvider`

Also moved code in the file for managing operators in better way. If this looks good, I will do the same for `LinqQueryOperatorProvider` too.

**NOTE** - This is not to be merged yet. For the feedback purposes only. Upgrade will be done, once Ix-Async releases RTM.


smitpatel:
@anpete - Any other workaround pending which should be removed?

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
```diff
@@ -640,154 +490,190 @@ private sealed class SelectAsyncEnumerator : IAsyncEnumerator<TResult>
...
-            using (var asyncEnumerator = source.GetEnumerator())
-            {
-                return await asyncEnumerator.MoveNext(cancellationToken);
-            }
-        }
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static IAsyncEnumerable<T> ToAsyncEnumerable<T>([NotNull] IEnumerable<T> source)
```
smitpatel: This can be converted to private method and expose only `MethodInfo` just like other operators for consistency.

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
```diff
@@ -26,6 +26,8 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
     /// </summary>
     public class AsyncLinqOperatorProvider : ILinqOperatorProvider
     {
+        #region EnumerableAdapters
```
smitpatel: Our custom enumerable adapter classes have different protection levels like `private/protected/internal`. Some of them are also sealed. Should they be consistent if not being used elsewhere?

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
```diff
@@ -640,154 +490,190 @@ private sealed class SelectAsyncEnumerator : IAsyncEnumerator<TResult>
...
+
+        // Set operations
+        private static readonly MethodInfo _concat = GetMethod("Concat", 1);
+        private static readonly MethodInfo _except = GetMethod("Except", 1);
+        private static readonly MethodInfo _intersect = GetMethod("Intersect", 1);
+        private static readonly MethodInfo _union = GetMethod("Union", 1);
+
+        #endregion
+
+        #region CustomOperators
```
smitpatel: We have only 1 custom operator left which is `GroupBy`
Our `GroupBy` operator returns `IAsyncEnumerable<IGrouping<TKey, TElement>>` whereas Ix-Async returns `IAsyncEnumerable<IAsyncGrouping<TKey, TElement>> `.

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
smitpatel: Also class `TaskResultAsyncEnumerable` is being used here internally only still it is made public sealed in a different file. (`.Internal` namespace). Due to being public, now we have unit testing for it too. Which is unnecessary if its kept private.

anpete:
@smitpatel Do a Find Usages on IAsyncEnumerable.GetEnumerator and/or IAsyncEnumerator.MoveNext

smitpatel:
Did another scan and removed few other methods. ping @anpete

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2532,18 +2514,7 @@ public static void Load<TSource>([NotNull] this IQueryable<TSource> source)
...
-            var asyncEnumerable = source.AsAsyncEnumerable();
-
-            using (var enumerator = asyncEnumerable.GetEnumerator())
-            {
-                while (await enumerator.MoveNext(cancellationToken))
-                {
-                }
-            }
-        }
+            => await source.AsAsyncEnumerable().ForEachAsync(_ => { }, cancellationToken);
```
smitpatel: @anpete - There is no LoadAsync method but this trick works also. Looks good?

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
onovotny: FWIW, this has been this way in 3.0 as well:
https://github.com/Reactive-Extensions/Rx.NET/blob/v3.0.0/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1166

`GroupBy` returned an `IAsyncGrouping`. That said, as an implementation detail, our `Grouping` impl does implement both `IAsyncGrouping` and `IGrouping`:
https://github.com/Reactive-Extensions/Rx.NET/blob/v3.1.0-rc2/Ix.NET/Source/System.Interactive.Async/Grouping.cs#L394

If you need to only expose `IGrouping`, then we can add notes in there to ensure that never changes so a cast would be safe. Just let us know.

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2130,21 +2130,7 @@ public static class EntityFrameworkQueryableExtensions
...
-            {
-                while (await asyncEnumerator.MoveNext(cancellationToken))
-                {
-                    list.Add(asyncEnumerator.Current);
-                }
-            }
-
-            return list;
-        }
+            => await source.AsAsyncEnumerable().ToList(cancellationToken);
```
anpete: Remove await

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2169,11 +2155,7 @@ public static class EntityFrameworkQueryableExtensions
         public static async Task<TSource[]> ToArrayAsync<TSource>(
             [NotNull] this IQueryable<TSource> source,
             CancellationToken cancellationToken = default(CancellationToken))
-        {
-            Check.NotNull(source, nameof(source));
-
-            return (await source.ToListAsync(cancellationToken)).ToArray();
-        }
+            => await source.AsAsyncEnumerable().ToArray(cancellationToken);
```
anpete: Remove await

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
anpete: The previous version is likely more efficient and straight-forward enough so I think we can just leave it.

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2747,18 +2718,7 @@ public static void Load<TSource>([NotNull] this IQueryable<TSource> source)
...
-
-            using (var asyncEnumerator = source.AsAsyncEnumerable().GetEnumerator())
-            {
-                while (await asyncEnumerator.MoveNext(cancellationToken))
-                {
-                    action(asyncEnumerator.Current);
-                }
-            }
-        }
+            => await source.AsAsyncEnumerable().ForEachAsync(action, cancellationToken);
```
anpete: Remove await.

smitpatel:
Updated.
@anpete - Can you give feedback on rest of comments above?

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
anpete: Re: 1, IIRC accessibility was changed to make .NET Native work. Could be we don't need the changes anymore. They are ideally private sealed. Also take a look at whether some can go away by using IX version. E.g. I seem to remember IX had something like ToAsyncEnumerable.

anpete: Re 2: Fine to move that into AsyncLinqOperatorProvider.

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
anpete: Also see comment above.

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
anpete: @smitpatel Can you see what it looks like to switch to the IX version?

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
smitpatel: Converted all but one to `private sealed`. `EnumerableAdapter<TResult>` is private not sealed since we derive from it in `OrderedEnumerableAdapter<TResult>`.
`ToAsyncEnumerable` is converted to IX version already. These are the classes which are adapters or specific to EF only.
Moved `TaskResultAsyncEnumerable` and made it private sealed. Removed its testing.

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
smitpatel: Converted the return type to use `IAsyncGrouping` and all works fine.
We can do either casting or change return type.

src/Microsoft.EntityFrameworkCore/Query/Internal/AsyncLinqOperatorProvider.cs
smitpatel: Changed though the usage had to change to make cast to `AsyncLinqOperatorProvider` since this method is not part of `ILinqOperatorProvider`.

anpete:
:shipit:

PR #6933, opened 2016-11-04
Set the filter for nullable unique indexes using a convention

### LINKED ISSUE #6794: Model Definition: Ability to override IS NOT NULL clauses of an index
> This is for SQL Server and I am using EF Core 1.0.1. 

Assume that I have the below entity:

``` csharp
public class Person
{
    public string Lastname { get; set; }
    public DateTime? DeletedOn { get; set; }
}
```

Inside `OnModelCreating`, I also have this to create a multi-column unique index:

``` csharp
{
    modelBuilder.Entity<Person>().HasIndex(x => new { x.Name, x.Lastname, x.DeletedOn }).IsUnique();
}
```

This causes EF Core to generate something similar to below:

``` csharp
CREATE UNIQUE NONCLUSTERED INDEX [IX_People_Name_Lastname_DeletedOn] ON [dbo].[People]
(
    [Name] ASC,
    [Lastname] ASC,
    [DeletedOn] ASC
)
WHERE ([Name] IS NOT NULL AND [Lastname] IS NOT NULL AND [DeletedOn] IS NOT NULL)
```

However, I want not to allow `DeletedOn` column to be null for this index. So, I am try to make EF Core to generate the below index creation script for me:

``` csharp
CREATE UNIQUE NONCLUSTERED INDEX [IX_People_Name_Lastname_DeletedOn] ON [dbo].[People]
(
    [Name] ASC,
    [Lastname] ASC,
    [DeletedOn] ASC
)
WHERE ([Name] IS NOT NULL AND [Lastname] IS NOT NULL)
```

**QUESTION: Is this possible to do this with EF Core fluent API?**

I dug a bit into the fluent API and was able to find some things related to `HasAnnotation` API but was not able to understand enough to achieve this.

### LINKED ISSUE #6901: SQL Server: Set nullable unique index filter using convention

AndriySvyryd:
Fixes #6794

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/AnnotationComparer.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/AnnotationComparer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
 
 namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestUtilities
 {
@@ -30,7 +31,8 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestUtilities
             }
 
             return x.Name == y.Name
-                   && x.Value.Equals(y.Value);
+                   && (x.Name == CoreAnnotationNames.ValueGeneratorFactoryAnnotation
+                       || x.Value.Equals(y.Value));
         }
 
         public int GetHashCode(IAnnotation obj) => obj.Name.GetHashCode() ^ obj.Value.GetHashCode();
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
@@ -421,25 +421,6 @@ namespace Microsoft.EntityFrameworkCore.Migrations
             else
             {
                 base.Generate(operation, model, builder, terminate: false);
-
-                var clustered = operation[SqlServerFullAnnotationNames.Instance.Clustered] as bool?;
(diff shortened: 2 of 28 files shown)


AndriySvyryd:
@bricelam or @smitpatel could you take a look?

src/Microsoft.EntityFrameworkCore.Specification.Tests/TestUtilities/AnnotationComparer.cs
```diff
@@ -30,7 +31,8 @@ public bool Equals(IAnnotation x, IAnnotation y)
             }
 
             return x.Name == y.Name
-                   && x.Value.Equals(y.Value);
+                   && (x.Name == CoreAnnotationNames.ValueGeneratorFactoryAnnotation
```
smitpatel: Why is this needed for the change?

AndriySvyryd: To allow the ModelBuilder tests that now run with SqlServer conventions (that add a discriminator) to pass

smitpatel:
:shipit:

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -421,25 +421,6 @@ protected override void Generate(DropTableOperation operation, IModel model, Mig
...
-                        if (i != 0)
-                        {
-                            builder.Append(" AND ");
-                        }
-
-                        builder
-                            .Append(SqlGenerationHelper.DelimitIdentifier(nullableColumns[i]))
-                            .Append(" IS NOT NULL");
-                    }
-                }
```
bricelam: Where does it get translated to SQL now?

smitpatel: It will be stored on `index.SqlServer().Filter` by conventions wherever applicable. Migrations use it from there.

Issue #6835, opened 2016-10-21
Conflict with decimal column type in generated batch


yarmenteros:
### The issue

I have problems with decimal types when they are saved in the database. The problem is caused by a conflict between the decimal type precision (decimal(18,2)) used in temporary table into the batch and the decimal type precision (decimal(3,3)) used to define the corresponding parameter into the same batch.

In my case the decimal value is 0.589, and in the database it's saved as a rounded value:  0.590. My column type onto the table is decimal(9,3).

**My question is:** Why if was well determined the the type and precision to be used as parameter, at the same time it can't use the same type and precision to the same column into temporary table used in batch?
### The Batch:

Affected column: **Cost** 
Column and Type defined in temporal table: **[Cost] decimal(18, 2)**
Column Type and value defined as parameter: @p71 **decimal(3,3)**
The value before run the batch: @p71=**0.589**
The saved value: **0.590**

```
exec sp_executesql 
N'SET NOCOUNT ON;
DECLARE @toInsert1 
       TABLE ([CompanyGroupShipKey] uniqueidentifier, 
                     [Cost] decimal(18, 2), 
                     [CreationDate] datetime2, 
                     [CreationUser] nvarchar(max), 
                     [InventoryKey] uniqueidentifier, 
                     [LastModifiedDate] datetime2, 
                     [LastModifiedUser] nvarchar(max), 
                     [LotId] int, 
                     [Notes] nvarchar(max), 
                     [OrderDetailKey] uniqueidentifier, 
                     [ProductKey] uniqueidentifier, 
                     [ProductUnitMeasureKey] uniqueidentifier, 
                     [Quantity] decimal(18, 2), 
                     [ReceivingHeadKey] uniqueidentifier, 
                     [RowNumber] smallint, 
                     [_Position] [int]);
INSERT INTO @toInsert1
VALUES (@p55, @p56, @p57, @p58, @p59, @p60, @p61, @p62, @p63, @p64, @p65, @p66, @p67, @p68, @p69, 0),
(@p70, @p71, @p72, @p73, @p74, @p75, @p76, @p77, @p78, @p79, @p80, @p81, @p82, @p83, @p84, 1);

DECLARE @inserted1 
       TABLE ([ReceivingDetailKey] uniqueidentifier, 
              [CompanyGroupShipKey] uniqueidentifier, 
              [Cost] decimal(18, 2), 
              [CreationDate] datetime2, 
              [CreationUser] nvarchar(max), 
[...]

PR #6944, opened 2016-11-04
Log a warning if a decimal property does not have an explicit type mapping.

AndriySvyryd:
Part of #6835

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerModelValidator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerModelValidator.cs
@@ -1,0 +1,48 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.EntityFrameworkCore.Storage.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.EntityFrameworkCore.Infrastructure
+{
+    public class SqlServerModelValidator : RelationalModelValidator
+    {
+        public SqlServerModelValidator(
+            [NotNull] ILogger<RelationalModelValidator> loggerFactory,
+            [NotNull] IRelationalAnnotationProvider relationalExtensions,
+            [NotNull] IRelationalTypeMapper typeMapper)
+            : base(loggerFactory, relationalExtensions, typeMapper)
+        {
+        }
+
+        public override void Validate(IModel model)
+        {
+            base.Validate(model);
+            EnsureNoDefaultDecimalMapping(model);
+        }
+
+        protected virtual void EnsureNoDefaultDecimalMapping([NotNull] IModel model)
+        {
+            foreach (var property in model.GetEntityTypes().SelectMany(t => t.GetDeclaredProperties())
+                .Where(p => p.ClrType == typeof(decimal) && p.SqlServer().ColumnType == null))
+            {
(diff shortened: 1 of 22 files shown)


src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerModelValidator.cs
```diff
@@ -0,0 +1,48 @@
...
+        public override void Validate(IModel model)
+        {
+            base.Validate(model);
+            EnsureNoDefaultDecimalMapping(model);
+        }
+
+        protected virtual void EnsureNoDefaultDecimalMapping([NotNull] IModel model)
+        {
+            foreach (var property in model.GetEntityTypes().SelectMany(t => t.GetDeclaredProperties())
+                .Where(p => p.ClrType == typeof(decimal) && p.SqlServer().ColumnType == null))
```
ajcvickers: Is it only decimals? What about floats and doubles? What about string mappings where we create, for example, nvarchar(128) by convention? Likewise for binary?

divega: Didn't we have something already that detects when `varchar` and `nvarchar` are set without size because the default size is 1 and it is unlikely to be the user's intent? How does that one work? Would it make sense to unify the approach?

ajcvickers: This isn't about setting "decimal" column type with no precision/scale. It's about always having to set some column type.

divega: Got it. Thanks for clarifying. I had forgotten that detail.

AndriySvyryd: @ajcvickers The default float and double mappings already have the max precision. String and binary truncations always result in errors.

ajcvickers: @AndriySvyryd How about datetime? Specifically, conversions between datetime and datetime2?

ajcvickers:
If we do this, then I think we should coordinate with scaffolding. The general behavior for scaffolding is that if the type of the column matches our default type mapping, then don't explicitly set the type. But this would mean that when scaffolding the bug on scaffolding is fixed, it will not generate code to set the store type for decimal(18,2). The runtime will then generate a warning which is a) bad because it comes from code we scaffolded and b) bad because all the types are actually in sync.

Therefore, if we start doing this warning, then scaffolding should always generated an explicit type for any decimal column. Which would be unfortunate, but better than generating a warning from scaffolded code.

src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerModelValidator.cs
AndriySvyryd: @ajcvickers datetime2 is the default and there are no values in datetime that cannot be represented by datetime2

AndriySvyryd:
@ajcvickers I updated #5410

AndriySvyryd:
Updated to use max for strings and byte[]

src/Microsoft.EntityFrameworkCore.SqlServer/Properties/SqlServerStrings.resx
```diff
@@ -138,4 +138,7 @@
   <data name="TransientExceptionDetected" xml:space="preserve">
     <value>An exception has been raised that is likely due to a transient failure. If you are connecting to a SQL Azure database consider using SqlAzureExecutionStrategy.</value>
   </data>
-</root>
+  <data name="DefaultDecimalTypeColumn" xml:space="preserve">
+    <value>No type was specified for the decimal column '{property}' on entity type '{entityType}', this will cause values to be silently truncated if they do not fit in the default precision and scale. Explicitly specify the SQL server column type that can accomadate all the values using 'ForSqlServerHasColumnType()'.</value>
```
ajcvickers: >'{entityType}', this [](start = 84, length = 20)

"'{entityType}', this" should be "'{entityType}'. This"

PR #6945, opened 2016-11-04
Query: Issue propert client eval warning for Last & LastOrDefault

### LINKED ISSUE #6929: Expecting  client eval warnings on Last and LastOrDefault query extensions
### Steps to reproduce
A simple Last or LastOrDefault query.

db.MyEntities.LastOrDefault();

I'm using the SQL Server provider.
Provider is configured to throw on client eval.

### The issue
Actual query against database is not filtered:

I was seeing this in the ILoggerProvider logging and then verified in SQL Profiler:

SELECT [s].[Id], [s].[Name]
FROM [Samurais] AS [s]

 Therefore I can only assume the full set is returned and the Last evaluation is happening in memory.

 I can see that the [throw on client eval] config is working because an Average method does cause an exception to throw.

No exception thrown for Last or LastOrDefault, 

### system
EF Core version: 1.1.0-preview1-final
Operating system: windows 10
Visual Studio version: VS2015

smitpatel:
Resolves #6929 

Issue: we always set ClientEval as false even though it was required. Added proper handling.

Rules:
Last with order by can always be translated to Sql. Do not log warning.
LastOrDefault with order by gets translated but we still do client eval for the exception (just like FirstOrDefault), Do not log warning.
Warn when LastOrDefault is used inside subquery.
Warn when Last is used without order by because we need to evaluate on client to find the last record.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/WarningsTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/WarningsTestBase.cs
@@ -80,7 +80,69 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
                 Assert.NotNull(query2);
             }
         }
-      
+
+        [Fact]
+        public virtual void LastOrDefault_with_order_by_does_not_issue_client_eval_warning()
+        {
+            using (var context = CreateContext())
+            {
+                var query1 = context.Customers.Where(c => c.CustomerID == "ALFKI" && c.Orders.OrderBy(o => o.OrderID).LastOrDefault().OrderID > 1000).ToList();
+                Assert.NotNull(query1);
+
+                var query2 = context.Customers.OrderBy(c => c.CustomerID).LastOrDefault();
+                Assert.NotNull(query2);
+            }
+        }
+
+        [Fact]
+        public virtual void Last_with_order_by_does_not_issue_client_eval_warning_if_at_top_level()
+        {
+            using (var context = CreateContext())
+            {
+                var query = context.Customers.OrderBy(c => c.CustomerID).Last();
+                Assert.NotNull(query);
+            }
+        }
+
+        [Fact]
+        public virtual void Last_without_order_by_issues_client_eval_warning()
+        {
+            using (var context = CreateContext())
+            {
+                Assert.Equal(CoreStrings.WarningAsErrorTemplate(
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/WarningsTestBase.cs
```diff
@@ -80,7 +80,62 @@ public virtual void Single_SingleOrDefault_without_orderby_doesnt_issue_warning(
                 Assert.NotNull(query2);
             }
         }
-      
+
+        [Fact]
+        public virtual void Last_with_order_by_does_not_issue_client_eval_warning()
```
maumar: We should always client eval Last because in case collection is empty it should throw an exception. LastOrDefault we could translate safely

smitpatel: What is current behavior with `First`?

smitpatel: Do we issue client eval warning for top level `First()` (when it is not inside subquery)? Due to exception, we would evaluate it on client anyway.

divega: `First()` in that case may be evaluated on the client to be able to throw but as long as the filtering happens on the server I believe it would be confusing to issue the warning.

maumar: yeah I think you guys are right, my concern was that if someone builds on top of Last(), they would want to know that those additional operations are evaluated on the client, but we shall issue warning for those, when they come. Perhaps we could add this as a test, i.e. use Last in the subquery and then compose on top of the result

```

```

maumar:
:shipit:

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/WarningsTestBase.cs
divega: >  Perhaps we could add this as a test...

Sounds good to me.

smitpatel:
@maumar @divega - I updated PR. There was small error which I corrected.
Current behavior is this:
If there is no order by then it will issue warning since we need to do client eval only.
If there is order by then
- `LastOrDefault` does not issue any warning.
- `Last` issues warning if its inside subquery. Top level `Last` does not issue warning.

I have updated test accordingly.

PR #6947, opened 2016-11-04
Move the derived foreign key properties validation to a common method.

### LINKED ISSUE #6792: Strange FK exception depending on property names
I'm mapping a complex domain model and encountered the exception, "The property 'ID' cannot be part of a foreign key on 'T' because it is contained in a key defined on a base entity type."  Looks like the throwing of this exception was introduced in #2837.

At first I thought this was due to either a limitation of TPH mapping in general, or the current EF core implementation specifically.  But now I think it's a bug in EF Core.  It was difficult to reproduce in isolation since it obscurely depends on property names, too.

Trying to create a database using this context and model will reproduce the issue:

```
{
        : base( options ) { }

    public DbSet<L> Ls { get; set; }
    public DbSet<PBase> PBases { get; set; }
    public DbSet<P> Ps { get; set; }
    public DbSet<Q> Qs { get; set; }
    public DbSet<T> Ts { get; set; }
}

public class L
{
    public int ID { get; set; }
}

public abstract class PBase
{
    public int ID { get; set; }
}

public class P : PBase
{
}

public class Q : PBase
{
}

public class T : P
{
    public P P { get; set; }
    public Q F { get; set; }
}
```

A workaround for this example is to rename the context's "Ls" DbSet property to "Zs".  Unfortunately, I've already come across a case in my real domain model where this simple hack doesn't help.  That is, it removes the exception for one class but that, in turn, causes the same exception to occur for another class.  Removing one of the properties but not the other doesn't remove the exception, either.

AndriySvyryd:
Add more details to the exception message.

Fixes #6792

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/EntityType.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/EntityType.cs
@@ -643,12 +643,18 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
                 {
                     throw new InvalidOperationException(CoreStrings.ForeignKeyPropertiesWrongEntity(Property.Format(properties), this.DisplayName()));
                 }
-
-                if (actualProperty.GetContainingKeys().Any(k => k.DeclaringEntityType != this))
-                {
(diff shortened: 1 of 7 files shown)


ajcvickers:
:shipit:

Issue #6826, opened 2016-10-20
Migrations: Parameterless DbContext constructor is used in preference to overload taking DbContextOptions


AndriySvyryd:
Using Startup.cs:

``` C#
public void ConfigureServices(IServiceCollection services)
{
    Console.WriteLine("ConfigureServices");

    services.AddDbContext<DataContext>(options =>
        options.UseSqlServer(Configuration.GetConnectionString(name: "DefaultConnection"))
            .EnableSensitiveDataLogging());
}
```

And context:

``` C#
public class DataContext : DbContext
{
    public DataContext()
    {
        Console.WriteLine("Parameterless ctor");
    }

    public DataContext(DbContextOptions<DataContext> options)
        : base(options)
    {
        Console.WriteLine("Options ctor");
    }
}
```

Output is

> ConfigureServices
> Parameterless ctor
> System.InvalidOperationException: No database provider has been configured for this DbContext. A provider can be configured by overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider. If AddDbContext is used, then also ensure that your DbContext type accepts a DbContextOptions<TContext> object in its constructor and passes it to the base constructor for DbContext.

bricelam:
It looks like we call `ActivatorUtilities.CreateInstance()`. I must have different semantics. We could try to call `IServiceProvider.GetService()` first and fallback...

anpete:
@AndriySvyryd @bricelam @divega Is this a DI bug? Did anyone file it?

divega:
Yes, someone did :smile: https://github.com/aspnet/DependencyInjection/issues/197, and it got moved to the backlog.

divega:
> We could try to call IServiceProvider.GetService() first and fallback...

@bricelam FWIW, there is already `ActivatorUtilities.GetServiceOrCreateInstance()` that has those semantics.

OskarKlintrot:
Should this be working in the latest (stable) release? I've the same behavior in `Microsoft.EntityFrameworkCore.Design` v1.1.2.

bricelam:
@OskarKlintrot No, it was fixed in the nightly builds of 2.0.0 for a while, but got removed (by #8328) before 2.0.0-preview1.

Re-opening so we can consider putting the fix in 1.1.3.

OskarKlintrot:
Thanks for the answer @bricelam, got a bit confused!

streamtree:
[...]

PR #6950, opened 2016-11-05
PERF: Fix #6923 - Implement DbContext pooling for high scale scenarios.

### LINKED ISSUE #6923: PERF: Implement DbContext pooling for high scale scenarios
- [x] Add overload of AddDbContext that enables pooling
- [x] OnConfiguring is never run. Configuration is external only.
- [x] Add state reset (context behavior flags, internal StateManager state)
- [x] Add Dispose event redirection.
- [ ] Logging
- [x] Pool cleanup

anpete:
- Initial impl. of context pooling.


anpete:
@ajcvickers @AndriySvyryd

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
```diff
@@ -0,0 +1,117 @@
...
+                Interlocked.Decrement(ref _count);
+
+                Debug.Assert(_count >= 0);
+
+                ((IDbContextPoolable)context).Resurrect(_configurationSnapshot);
+
+                return context;
+            }
+
+            context = ActivatorUtilities.CreateInstance<TContext>(serviceProvider);
```
ajcvickers: This probably has the same issues as #6826

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            if (!_takenSnapshot)
+            {
+                // Race to set this
+                _takenSnapshot = true;
+                _configurationSnapshot = ((IDbContextPoolable)context).SnapshotConfiguration();
```
ajcvickers: Doesn't this need memory barrier protection?

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
anpete: Is this a DI bug? I can't think of any other way to do this - We can't call GetService here.

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
anpete: Can you elaborate? The intent of this is we might end up calling Snapshot multiple times, but we don't care which one we end up with - they should all be equivalent.

ajcvickers: How do we guarantee that the object has been fully constructed when viewed from another thread? Pretty sure this needs some form of memory barrier.

ajcvickers:
Clean up transaction state? Reset command timeout?

anpete:
@ajcvickers Thanks, was sure I had missed stuff. With transactions, is it just clearing IRelationalConnection.CurrentTransaction?

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
anpete: I see. Would changing the snapshot to be a class and removing the flag be a better option?

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
divega: https://github.com/aspnet/DependencyInjection/issues/197

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
ajcvickers: I don't see anyway to do it without a memory barrier.

ajcvickers:
Not sure about the transactions. @AndriySvyryd would know.

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
anpete: OK, it looks like we should validate that the context type does not have a default ctor.

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
anpete: Oh, so another thread could see _configurationSnapshot != null even though _configurationSnapshot was not fully constructed?

ajcvickers: Yes

AndriySvyryd:
Calling `IRelationalConnection.Dispose` should be enough to reset the transaction state

AndriySvyryd:
I think `IDbContextPoolable` should provide a way to subscribe to the resetting, so non-core services are able to participate

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
ajcvickers: I wonder if this is a symptom of a wider issue. The DbContext class is being registered in D.I. as a scoped service, but it isn't really used that way because of the pool. So what happens if it depends on other scoped services? How does it get the correct instance of those services?

Two other architectures to consider:
* Make the pool explicit such that other services (e.g. controllers) depend on the pool and grab a context from the pool when needed. The DbContext would never be registered in D.I. at all.
* Be like DbConnection where the instance of DbContext itself is created by D,I, in the normal way, so it gets correctly injected other services, but the guts of DbContext is obtained from a pool.

anpete:
@AndriySvyryd It would be great to see an example of where this is needed to understand what the solution might look like.

AndriySvyryd:
@anpete The `IRelationalConnection` above for instance.

anpete:
@AndriySvyryd Great! Will take a look. BTW, it looks like it is safe to call RelationalConnection.Dispose but then keep using it?! Is this by design?

AndriySvyryd:
@anpete I believe it's because there was no real downside to this approach and it was less code

anpete:
@AndriySvyryd Added provider services reset in latest commit.

src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
```diff
@@ -73,6 +75,128 @@ public static class EntityFrameworkServiceCollectionExtensions
...
+        ///     </para>
+        ///     <para>
+        ///         This overload has an <paramref name="optionsAction" /> that provides the applications <see cref="IServiceProvider" />.
+        ///         This is useful if you want to setup Entity Framework to resolve its internal services from the primary application service
+        ///         provider.
+        ///         By default, we recommend using the other overload, which allows Entity Framework to create and maintain its own
+        ///         <see cref="IServiceProvider" />
+        ///         for internal Entity Framework services.
+        ///     </para>
+        /// </summary>
```
AndriySvyryd: We should mention in the summary for both overloads that polling would be enabled so no state should be maintained on the context and `OnConfiguring` will not be called

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace Microsoft.EntityFrameworkCore.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class DbContextPool<TContext> : IDbContextPool, IDisposable
+        where TContext : DbContext
+    {
+        private const int DefaultPoolSize = 32;
```
AndriySvyryd: Different hardware configurations need different values, so it would be more useful if this was based on `ThreadPool.GetMinThreads`, though with current API it would be rarely used.

AndriySvyryd:
:shipit: contingent on @ajcvickers signoff

ajcvickers:
I'm still not happy with the use of scoped services like this. Feels like pit of failure to me since you have to understand the scoped-per-request-but-not-really semantics. If @divega is fine with it, then he's the boss. To me it feels like there are better architectural options.

divega:
I would prefer to sort out @ajcvickers' concerns in person. I would also like to see if we can use `ActivatorUtilities.CreateFactory()` to avoid reflection/expression compilation overhead each time we create a new instance.

src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
```diff
@@ -73,6 +75,128 @@ public static class EntityFrameworkServiceCollectionExtensions
...
+        /// </param>
+        /// <param name="poolSize">
+        ///     Enables DbContext instance pooling and sets the maximum number of instances retained by the pool.
+        ///     Instance pooling can increase throughput in high-scale scenarios such as web servers by re-using
+        ///     DbContext instances, rather than creating new instances for each request.
+        /// </param>
+        /// <returns>
+        ///     The same service collection so that multiple calls can be chained.
+        /// </returns>
+        public static IServiceCollection AddDbContext<TContext>(
```
divega: Another thing to chat about: Is it better to add these new overloads of `AddDbContext<TContext>()` that take `poolSize` or instead have a new method, e.g. `AddDbContextPool<TContext>()` that makes it a bit clearer that you are forking from the regular behavior of `AddDbContext`, for which `poolSize` is optional (or not specified in all overloads) so that we can pick a good default for you, and for which we can also have very specific and separate documentation/xml comments.

davidroth:
@ajcvickers Out of curiosity:  Can you share some benchmarks about this improvement?

anpete:
@ajcvickers @AndriySvyryd @divega Added a commit which I think addresses the decisions from today. Let me know if I missed something.

src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
```diff
@@ -63,6 +66,104 @@ public static class EntityFrameworkServiceCollectionExtensions
...
+        ///     Sets the maximum number of instances retained by the pool.
+        /// </param>
+        /// <returns>
+        ///     The same service collection so that multiple calls can be chained.
+        /// </returns>
+        public static IServiceCollection AddDbContextPool<TContext>(
+            [NotNull] this IServiceCollection serviceCollection,
+            [NotNull] Action<IServiceProvider, DbContextOptionsBuilder> optionsAction,
+            int poolSize = 128)
+            where TContext : DbContext
```
divega: Assuming we can and it makes sense to use the API that @AndriySvyryd had suggested for an optimal pool size, I think we should consider making it of type int? or using overloads instead of an optional parameters.

divega: The rest LGTM.

src/Microsoft.EntityFrameworkCore/Internal/DbContextPool.cs
```diff
@@ -0,0 +1,150 @@
...
+                        | BindingFlags.NonPublic);
+
+            if (ctors.Length == 1)
+            {
+                var parameters = ctors[0].GetParameters();
+
+                if (parameters.Length == 1
+                    && (parameters[0].ParameterType == typeof(DbContextOptions) 
+                        || parameters[0].ParameterType == typeof(DbContextOptions<TContext>)))
+                {
```
ajcvickers: @anpete @divega Non-generic options is fine. Hopefully this makes it work with non-derived DbContext. Maybe add a test for that?

anpete: @ajcvickers Added the test. Had to loosen the ctor validation logic slightly to ignore NonPublic ctors.

src/Microsoft.EntityFrameworkCore/EntityFrameworkServiceCollectionExtensions.cs
anpete: @divega ThreadPool.GetMinThreads is not what we want here, at least not for web apps. For a web app, what we really want is something close to the average number of concurrent requests (FYI IIS defaults to 5000 for maxConcurrentRequests for example). So I think we may just need to pick a number that is somewhat sensible.

PR #6972, opened 2016-11-09
Update SqlServerRetryingExecutionStrategy

### LINKED ISSUE #6970: Update SQL Server execution strategy
1205, 41301 and 41839 should be considered transient.
41301, 41302, 41305, 41325 and 41839 should be retried much more aggressively, with all retries happening within 100ms from the first failure.
See https://msdn.microsoft.com/en-us/library/mt668435.aspx#Conflict%20Detection%20and%20Retry%20Logic for more information

AndriySvyryd:
Add 1205, 41301 and 41839 to the transient errors list.
Use a shorter retry delay for memory-optimized errors.

Fixes #6970

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/SqlServerRetryingExecutionStrategy.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/SqlServerRetryingExecutionStrategy.cs
@@ -15,7 +15,7 @@ namespace Microsoft.EntityFrameworkCore
 {
     public class SqlServerRetryingExecutionStrategy : ExecutionStrategy
     {
-        private readonly ICollection<int> _errorNumbersToAdd;
+        private readonly ICollection<int> _additionalErrorNumbers;
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Microsoft.EntityFrameworkCore.SqlServer/SqlServerRetryingExecutionStrategy.cs
```diff
@@ -118,5 +118,43 @@ protected override bool ShouldRetryOn(Exception exception)
...
+
+            return baseDelay;
+        }
+
+        private bool IsMemoryOptimizedError(Exception exception)
+        {
+            var sqlException = exception as SqlException;
+            if (sqlException != null)
+            {
+                foreach (SqlError err in sqlException.Errors)
```
Eilon: `var`?

AndriySvyryd: nop, that's an old-school non-generic `IEnumerable`

PR #6989, opened 2016-11-10
Add EntityTypeConfiguration<TEntity>

### LINKED ISSUE #2805: EntityTypeConfiguration<T> (or similar)

rowanmiller:
Resolves #2805

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/IEntityTypeConfiguration`.cs
+++ b/src/Microsoft.EntityFrameworkCore/IEntityTypeConfiguration`.cs
@@ -1,0 +1,25 @@
+﻿using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using System;
+
+namespace Microsoft.EntityFrameworkCore
+{
+    /// <summary>
+    ///     Allows configuration for an entity type to be factored into a separate class,
+    ///     rather than in-line in <see cref="DbContext.OnModelCreating(ModelBuilder)"/>.
+    ///     Implement this interface, applying configuration for the entity in the
+    ///     <see cref="Configure(EntityTypeBuilder{TEntity})"/> method,
+    ///     and then apply the configuration to the model using 
+    ///     <see cref="ModelBuilder.ApplyConfiguration{TEntity}"/>
+    ///     in <see cref="DbContext.OnModelCreating(ModelBuilder)"/>.
+    /// </summary>
+    /// <typeparam name="TEntity"> The entity type to be configured. </typeparam>
+    public interface IEntityTypeConfiguration<TEntity>
+        where TEntity : class
+    {
+        /// <summary>
+        ///     Configures the entity of type <typeparamref name="TEntity"/>.
+        /// </summary>
+        /// <param name="builder"> The builder to be used to configure the entity type. </param>
+        void Configure(EntityTypeBuilder<TEntity> builder);
+    }
+}
--- a/src/Microsoft.EntityFrameworkCore/Microsoft.EntityFrameworkCore.csproj
+++ b/src/Microsoft.EntityFrameworkCore/Microsoft.EntityFrameworkCore.csproj
@@ -168,6 +168,7 @@
     <Compile Include="EntityFrameworkQueryableExtensions.cs" />
     <Compile Include="EntityFrameworkServiceCollectionExtensions.cs" />
     <Compile Include="EntityState.cs" />
+    <Compile Include="IEntityTypeConfiguration`.cs" />
     <Compile Include="Extensions\EntityTypeExtensions.cs" />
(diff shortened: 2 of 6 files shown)


src/Microsoft.EntityFrameworkCore/EntityTypeConfiguration`.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// <summary>
+    ///     Allows configuration for an entity type to be factored into a separate class,
+    ///     rather than in-line in <see cref="DbContext.OnModelCreating(ModelBuilder)"/>.
+    ///     Derive from this class, override <see cref="Configure(EntityTypeBuilder{TEntity})"/>,
+    ///     and then apply the configuration to the model using 
+    ///     <see cref="ModelBuilder.ApplyConfiguration{TEntity}"/>
+    ///     in <see cref="DbContext.OnModelCreating(ModelBuilder)"/>.
+    /// </summary>
+    /// <typeparam name="TEntity"> The entity type to be configured. </typeparam>
+    public abstract class EntityTypeConfiguration<TEntity>
```
ajcvickers: Why is this not an interface?

rowanmiller: Because the thought didn't even cross my mind... I was just doing it like we did in EF6 (eyeroll from @anpete 😁). I'll swap it to an interface.

anpete:
May want to create a separate bug for auto-discovery of these, which is mentioned in the original bug.

anpete:
Are we OK with only having a generic version?

rowanmiller:
I don't know that we want to auto-scan for these but I think an AddFromAssembly like we had in EF6 is an ok idea.

I played with having a non-generic one too - mostly so that I could have an ApplyConfiguration<TConfiguration>() method (which is super messy with just the generic version). I'm happy to add it if we think it's helpful. I decided not to because you end up with public stuff on the generic version that is distracting (public Type EntityType { get;} etc.)... but happy to add it back.

ErikEJ:
@sjh37 !!

sjh37:
A nice improvement. I'll make sure to add this to the EF core reverse generator.

ErikEJ:
@sjh37 Yes, it should allow you to split the files in a similar way to how you do it now

rowanmiller:
Decided in design meeting to just keep the generic version. Updated PR to use interface rather than abstract base class.

anhphan:
Not sure keeping Generic style is a good idea, currently I have a self implement of mapping and I have something like this for registration:
```
        {
            var collections = ServiceProvider.GetServices<IModelBuilder>();
            foreach (var callback in collections)
            {
                callback.OnModelCreating(modelBuilder);
            }
        }
```

For auto scan, I used ApplicationParts to registry Mapping class to DI.

Diaskhan:
After on cofiguring we could access to dbset of dbcontext? 
How usually manipulate with dynamically loading poco objects ?

Issue #6990, opened 2016-11-10
Stop special-casing the root entity on Attach


ajcvickers:
Attach sets the state of all untracked entities in the reachable graph to Unchanged, unless the key for the entity has not been set and is using value generation. The same is true for Update/Modified. This would work great for many common cases of disconnected entities.

However, it doesn't work in many cases only because we special case the root entity and force its state to Unchanged even if the key is not set. This is so that you don't call Attach on a single entity and end up with it being Added, which was considered an unexpected behavior.

We should stop special-casing the root entity because this would make Attach more generally useful. It probably isn't even a breaking change because attaching an entity with a store generated key that does not have a key value set will throw either immediately or when a second entity does the same thing.

JonPSmith:
Hi @ajcvickers,

I think this is a good change - it makes a lot of sense. Here is some feedback on the announcement.

I saw your [announcement](https://github.com/aspnet/EntityFramework/issues/8923#issuecomment-310728410), and I noted that:
- It didn't say in the title that `Update` has also been changed, although later in the announcement it talked about `Update`.
- Also, I don't think your announcement defines the change as clearly as note at the top of this post, i.e. it doesn't say that it will set the root entity State to `Added` for default PK, or non-database generated PK.
- Can I suggest that you look at the original issue #4424 which outlined how EF Core differs from EF6.x., and see if it needs editing (it isn't obvious to me what you would add, but you may have a better idea.  
I think the initial comment, written by Rowan Miller, is a very useful summary on *why* the changes were made. Julie Lerman refers to it in [one of her articles](https://msdn.microsoft.com/en-us/magazine/mt767693.aspx?f=255&MSPPError=-2147217396), and I also refer to it in the book I am writing. 

PS. I did run a unit test around non-database generated primary keys (see [This Unit Test](https://github.com/JonPSmith/EfCoreInAction/blob/Chapter09/Test/UnitTests/DataLayer/Ch09_NonDbGeneratedKey.cs)) on `v2.0.0-preview-final` I can see both `Attach` and `Update` have changed.

bertstomphorst:
@ajcvickers it seems that this fix doesn't work when the (nested) entities key is a DatabaseGenerated int, because the IsKeySet returns True in that scenario (and the key is set to a negative number). Just seen this on EF 2.2.3, and fixed the unwanted add by setting the State to Modified when key>0.

ajcvickers:
[...]

PR #6991, opened 2016-11-10
Stop special-casing the root entity on Attach

ajcvickers:
Issue #6990

Makes Attach/Update more useful by properly using the value-generation semantics built into the model.

Note that we haven't decided to do this yet, but I did the work to validate the approach. I will not merge until we decide to do it.


anpete:
:shipit:

PR #6994, opened 2016-11-10
Make InExpression accept broader type of argument

smitpatel:
since Sql supports In with following expressions
- ParameterExpression
- ConstantExpression
- ColumnExpression
- AliasExpression

@anpete @maumar @divega

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/EqualityPredicateInExpressionOptimizer.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/EqualityPredicateInExpressionOptimizer.cs
@@ -73,16 +73,16 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
             if (leftAliasExpression == null)
             {
-                leftAliasExpression = equalityType == ExpressionType.Equal
+                leftAliasExpression = (equalityType == ExpressionType.Equal
                     ? MatchInExpression(leftExpression, ref leftInValues)
-                    : MatchNotInExpression(leftExpression, ref leftInValues);
+                    : MatchNotInExpression(leftExpression, ref leftInValues)) as AliasExpression;
             }
 
             if (rightAliasExpression == null)
             {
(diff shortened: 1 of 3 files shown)


anpete:
Can we write tests that hit these new expressions?

smitpatel:
@anpete - While searching into existing test, I checked HandleContains in relational result operators, it supports Constant/Parameter but wraps it inside AliasExpression. How do we want to proceed? Both ways could work.

anpete:
@smitpatel I'm fine with moving HandleContains to the new approach.

smitpatel:
Updated for `HandleContains`.
For testing we already have test containing Constant/Column/Alias/Parameter inside InExpression.

maumar:
wont the scenarios with constant/parameter be funcletized? I think that's why we originally support only alias expression (not sure why we excluded column, perhaps it was never just a column in the queries we were looking into). Also for the tests, I would expect the expected SQL to change in those tests (i.e. contains now to be translated into SQL)

smitpatel:
@maumar - Scenarios with Constant/Parameter would be funcletized IFF search element and values both are not tied with server. Then it has deterministic answer and we would use it only. But if values to looking into is not fixed. i.e. subquery, then we need to look for a `constant IN subquery`, which is exactly what happens with contains case.
Not sure why we only supported AliasExpression but due to the limitation in `HandleContains` we actually wrap the expression in `AliasExpression` so that it could work which seems like a hack. (`new AliasExpression(expression)` is not actually an aliased expression but we are using AliasExpression in many wrong ways anyway). So SQL in tests is not changing because we are removing the work-around which was in place due to limitation.

While it is subjective since, it could work both ways - i.e. Restrict on AliasExpression but wrap it in AliasExpression whenever its not. Which works fine since we have only 1 place where we make `InExpression`. I hit this issue while fixing some of the test failures in SqlCE provider which does not support scalar subquery. One way to convert scalar subquery comparison is to use `IN` operator. While fixing it, I came across the limitation and was restricted to convert queries only when there is AliasExpression. Which gives false info to providers that they can only use InExpression with AliasExpression. The work-around we have in place is not very intuitive to figure out.

maumar:
:shipit:

Issue #6937, opened 2016-11-04
Query: Contains on element coming from optional navigation doesn't get server-evaluated


kierenj:
I'd like to help fix this, but if there's a workaround or suggested different way to express the same query that might work to get me around this in the short term, I'd be very grateful.  Any workaround suggestions are welcome!

### Steps to reproduce

Build a query with the following:

```
var set = (IQueryable<WidgetImage>)_context.Set<WidgetImage>();

// (there are some other filters here..)
 
// including this line, an "exclude any images where all of the textures IDs are excluded", breaks things
set = set.Where(i =>
    !i.UsedWidgets.Select(b =>
        b.WidgetsUsage.Brick.Texture.Id).All(t =>
            query.textures.ids.Contains(t)));

// (and here...)

var resultsProj = set.Select(w => new { w.Id }).Take(18);
```

Relevant entity models:

```
    public class WidgetImage
    {
        public Guid Id { get; set; }
        public List<ImageWidgetUsageLink> UsedWidgets { get; set; }
    }

    public class ImageWidgetUsageLink
    {
        public Guid ImageId { get; set; }
        public WidgetImage Image { get; set; }
        public Guid WidgetUsageId { get; set; }
        public WidgetUsage WidgetUsage { get; set; }
    }
    // config:
    e.HasKey(l => new { l.ImageId, l.WidgetUsageId });
    e.HasOne(l => l.Image).WithMany(i => i.UsedWidgets).IsRequired(true).HasForeignKey(l => l.ImageId);
    e.HasOne(l => l.WidgetUsage).WithMany(u => u.ImageLinks).IsRequired(true).HasForeignKey(l => l.WidgetUsageId);

    public class WidgetUsage
    {
        public Guid Id { get; set; }
        public Widget Widget { get; set; }
        public List<ImageWidgetUsageLink> ImageLinks { get; set; }
    }
    // config:
    e.HasOne(u => u.Widget).WithMany(b => b.Uses).IsRequired(true);

    public class Widget
    {
        public Guid Id { get; set; }
        public WidgetTexture Texture { get; set; }
        public List<WidgetUsage> Uses { get; set; }
    }
    // config:
    e.HasOne(b => b.Texture).WithMany(c => c.Widgets).IsRequired(false);

    public class WidgetTexture
    {
        public Guid Id { get; set; }
        public string Name { get; set; }
        public List<Widget> Widgets { get; set; }
    }
```

### The issue

[...]

PR #6998, opened 2016-11-11
Fix to #6937 - Query: Contains on element coming from optional navigation doesn't get server-evaluated

maumar:
Problem was that when trying to translate contains ResultOperator we assumed that the translatable expression would always be member access or a EF.Property method call.
However in some cases (e.g. when item is part of an optional navigation) this is not the case, and the underlying member/EF.Property is wrapped around Convert and/or NullConditional.
Fix is to perform translation on the item before before checking what shape it is. This simplifies the code and protects from similar situations, because SqlTranslatingExpressionVisitor already optimizes out a lot of redundant nodes in the expression tree.


maumar:
@anpete @smitpatel

maumar:
merged in 89659030884949531dc88aeedef74bc8cc9666f1

Issue #5459, opened 2016-05-22
Make use of prepared commands


roji:
ADO.NET allows users to prepare commands, which allows databases to cache query plans and parses. IIRC in modern version of SqlServer this doesn't have much effect (any?), but in other providers this is an important feature - repeated complicated queries in PostgreSQL definitely benefit from prepared commands.

I know there's already an SQL caching mechanism inside EFCore, and although I'm sure it would be non-trivial to cache commands as well it would be a great feature.

Definitely a lower-priority backlog feature request.

bricelam:
If we do this, we should definitely do aspnet/Microsoft.Data.Sqlite#235 :wink:

roji:
FYI, I'm working on some benchmarking and performance aspects of Npgsql, and prepared command use in EFCore would potentially be huge. I hope it gets prioritized at some point.

roji:
Note that for the next release of Npgsql (3.2), I plan to introduce an automatic/implicit preparation mechanism at the driver level - any statement used more than X times would be automatically prepared up to a certain quota, etc. This should increase performance with EFCore in a very significant way, without any change on your end.

However, it seems like EFCore may have more knowledge as to which statements to prepare - in any case it seems better-placed to make such decisions than the ADO.NET driver. So I think this feature is still relevant at the EFCore level.

bricelam:
Microsoft.Data.Sqlite will support prepared statements in version 2.1.0. PR: aspnet/Microsoft.Data.Sqlite#388

bricelam:
The biggest impact this feature would have on SQLite is in the update pipeline. Ironically, not batching statements, but instead re-using single-statement commands yields the best performance.

roji:
@bricelam is it not possible to both batch and use prepared statements at the same time? With Npgsql/PostgreSQL this yields the best performance.

bricelam:
SQLite doesn't actually support batching. 😀 We fake it in the ADO.NET layer.

smitpatel:
Take it till you make it :trollface:

roji:
Oh right, for a moment I forgot we're talking about a local database where batching would be if limited utility anyway :)

P-DHrestak:
So many years have passed, is this still not supported?
After many attempts to find a way for Npgsql prepared statements to work with EF Core still no success.

[...]

PR #7009, opened 2016-11-13
Query: Perf: Adds explicit compiled queries.

anpete:
- Adds EF.CompileQuery and EF.CompileAsyncQuery which allow for the creation of user managed
compiled query delegates. Such queries are more efficient to execute because they bypass the
CompiledQueryCache (avoiding query hashing and parameterization). Typical gains are around 3-5%

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/CompiledQueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/CompiledQueryTestBase.cs
@@ -173,8 +173,8 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         {
             var query = EF.CompileQuery(
                 (NorthwindContext context)
-                    => context.Customers.OrderBy(c => c.CustomerID).Select(c => c.CustomerID).FirstOrDefault() 
-                        + context.Orders.OrderBy(o => o.CustomerID).Select(o => o.CustomerID).FirstOrDefault());
+                    => context.Customers.OrderBy(c => c.CustomerID).Select(c => c.CustomerID).FirstOrDefault()
+                       + context.Orders.OrderBy(o => o.CustomerID).Select(o => o.CustomerID).FirstOrDefault());
 
             using (var context = CreateContext())
             {
@@ -215,12 +215,12 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
             using (var context = CreateContext())
             {
-                Assert.Equal(91, await query(context).Count());
+                Assert.Equal(91, (await query(context).ToListAsync()).Count);
             }
 
             using (var context = CreateContext())
             {
-                Assert.Equal(91, (await query(context).ToList()).Count);
+                Assert.Equal(91, (await query(context).ToListAsync()).Count);
             }
         }
 
@@ -244,12 +244,12 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
             using (var context = CreateContext())
             {
-                Assert.Equal(91, await query(context).Count());
+                Assert.Equal(91, (await query(context).ToListAsync()).Count);
             }
 
             using (var context = CreateContext())
             {
-                Assert.Equal(91, (await query(context).ToList()).Count);
+                Assert.Equal(91, (await query(context).ToListAsync()).Count);
             }
         }
 
@@ -262,12 +262,12 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
             using (var context = CreateContext())
             {
-                Assert.Equal("ALFKI", (await query(context, "ALFKI").First()).CustomerID);
+                Assert.Equal("ALFKI", (await query(context, "ALFKI").ToListAsync()).First().CustomerID);
             }
 
             using (var context = CreateContext())
             {
-                Assert.Equal("ANATR", (await query(context, "ANATR").First()).CustomerID);
+                Assert.Equal("ANATR", (await query(context, "ANATR").ToListAsync()).First().CustomerID);
             }
         }
 
@@ -318,12 +318,12 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
             using (var context = CreateContext())
             {
-                Assert.Equal("ALFKI", (await query(context, null, "ALFKI").First()).CustomerID);
+                Assert.Equal("ALFKI", (await query(context, null, "ALFKI").ToListAsync()).First().CustomerID);
             }
 
             using (var context = CreateContext())
             {
-                Assert.Equal("ANATR", (await query(context, null, "ANATR").First()).CustomerID);
+                Assert.Equal("ANATR", (await query(context, null, "ANATR").ToListAsync()).First().CustomerID);
             }
         }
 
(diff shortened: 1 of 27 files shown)


src/Microsoft.EntityFrameworkCore.Specification.Tests/CompiledQueryTestBase.cs
```diff
@@ -0,0 +1,359 @@
...
+
+            using (var context = CreateContext())
+            {
+                Assert.Equal("ANATR", query(context, null, 1, "ANATR").First().CustomerID);
+            }
+        }
+
+        [ConditionalFact]
+        public virtual void Query_with_array_parameter()
+        {
```
maumar: another interesting query could be:

context.Customers.OrderBy(c => c.Id).Select(c => c.Name).FirstOrDefault() + context.Orders.OrderBy(o => o.Id).Select(o.CustomerId).FirstOrDefault()

i.e. the root of the query is context, not dbset, curious if that gets translated into one query or will string concat be client-evaled

src/Microsoft.EntityFrameworkCore/EF.CompileAsyncQuery.cs
```diff
@@ -0,0 +1,304 @@
...
+        /// </summary>
+        /// <typeparam name="TContext">The target DbContext type.</typeparam>
+        /// <typeparam name="TParam1">The type of the first query parameter.</typeparam>
+        /// <typeparam name="TParam2">The type of the second query parameter.</typeparam>
+        /// <typeparam name="TParam3">The type of the third query parameter.</typeparam>
+        /// <typeparam name="TParam4">The type of the fourth query parameter.</typeparam>
+        /// <typeparam name="TParam5">The type of the fifth query parameter.</typeparam>
+        /// <typeparam name="TResult">The query result type.</typeparam>
+        /// <param name="queryExpression">The LINQ query expression.</param>
+        /// <returns>A delegate that can be invoked to execute the compiled query.</returns>
```
maumar: we think 5 parameters is enough? in EF6 we had 10+ IIRC

anpete: We can add more. I'm also going to enable object arrays as an alternative.

divega: In the old stack we had `ObjectContext` + 15 arguments + `TResult` = 17. I think 5 is a good start :smile:

src/Microsoft.EntityFrameworkCore/Query/Internal/CompiledAsyncTaskQuery.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Threading.Tasks;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Query.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class CompiledAsyncTaskQuery<TContext, TResult> : CompiledQueryBase<TContext, Task<TResult>>
```
maumar: nit: Make the name more indicative of what it does? (Singleton/Scalar etc) It was not obvious for me at the first glance.

src/Microsoft.EntityFrameworkCore/EF.CompileAsyncQuery.cs
```diff
@@ -0,0 +1,304 @@
...
+    public static partial class EF
+    {
+        /// <summary>
+        ///     Creates a compiled query delegate that when invoked will execute the specified LINQ query.
+        /// </summary>
+        /// <typeparam name="TContext">The target DbContext type.</typeparam>
+        /// <typeparam name="TResult">The query result type.</typeparam>
+        /// <param name="queryExpression">The LINQ query expression.</param>
+        /// <returns>A delegate that can be invoked to execute the compiled query.</returns>
+        public static Func<TContext, IAsyncEnumerable<TResult>> CompileAsyncQuery<TContext, TResult>(
```
divega: Open issue @anpete and I already talked about: this may mean the user has to add a direct reference to `System.Interactive.Async` which `IAsyncEnumerable<T> for any meaningful usage. We may make it easier by creating a few extension methods, e.g. `ToListAsync()` and `ForEachAsync()`. Another option could be to return `IQueryable<T>` or an EF specific query result type for which we support async enumeration.

src/Microsoft.EntityFrameworkCore/Query/Internal/CompiledAsyncTaskQuery.cs
anpete: Had "Singleton" but switched to Task for clarity 😄 Happy to change it though.

src/Microsoft.EntityFrameworkCore/EF.CompileAsyncQuery.cs
anpete: @maumar volunteered to add more!

src/Microsoft.EntityFrameworkCore.Specification.Tests/CompiledQueryTestBase.cs
anpete: Looks like we need to do some work to make these work. Filed #7016

src/Microsoft.EntityFrameworkCore.Specification.Tests/CompiledQueryTestBase.cs
```diff
@@ -0,0 +1,359 @@
...
+            }
+
+            using (var context = CreateContext())
+            {
+                Assert.Equal("ANATR", query(context, null, 1, "ANATR").First().CustomerID);
+            }
+        }
+
+        [ConditionalFact]
+        public virtual void Query_with_array_parameter()
```
anpete: Added

divega:
@anpete reminder to add a test that uses `FromSql()`

roji:
It may be worth thinking about this in relation with prepared statements (#5459).

The explicit compiled queries implemented here aren't bound to any context, so they don't exactly correspond (explicit compiled queries are about caching inside EF, while prepared statements are about caching database-side). However, what would the statement preparation API look like, and how would it interact with this API? It's safe to say that these 2 features will almost always go together (if one wants to cache EF-side resources, one would probably also want to cache database-side resources).

anpete:
@divega @maumar Added a wrapping async enumerable as we discussed.

anpete:
@roji AFAIK, we don't need this for SQL Server because it happens auto-magically. I'm guessing there is a good reason to make it explicit opt-in for Postgres?

roji:
@anpete yeah, SQL Server implicitly caches query plans etc. - but this isn't the standard state of affairs across databases. PostgreSQL requires you to explicitly prepare statements, and there'a a huge performance benefit associated with doing so (am pretty sure other databases exist which are similar).. ADO.NET as an API supports this via DbCommand.Prepare(), so it makes sense for EF to support preparation in at least some way.

Note that I'm looking into providing opt-in automatic query preparation in Npgsql with an LRU approach - this would make it redundant at the EF level, but it's still not done (and wouldn't affect other databases which can benefit from preparation). In fact, if you guys give some priority to preparation in EF I'll probably delay/drop the functionality in the ADO.NET later.

Basically for a strong cross-database story you really need to address this in some way...

tinchou:
Hey, sorry to ask here but I'm investigating this feature for a presentation and couldn't find a better place :).

I've seen in [this PR](https://github.com/aspnet/benchmarks/pull/179/files#diff-580ab548c69197b6c524292dcbc3ebe8R28) in aspnet/benchmarks that you switch from using `FirstAsync` to `CompileAsyncQuery` + `First`. Does that mean that you only use sync methods when compiling queries?

anpete:
@tinchou Correct. The Async versions cannot be used in expression trees because they have optional args 😦

tinchou:
Bummer! But internally we're using the same paths as the async overloads and this is true async code, right? So in the end it wouldn't matter which method you chose.

anpete:
Yeah, still async 😌

PR #7018, opened 2016-11-14
Build fixes for SqlServer 2008

smitpatel:
@anpete @maumar 
Some of the tests which relied on Supporting offset are made ConditionalFact
For, rownumberpaging we did not modify select expression if it was wrapped inside Exists. Causing invalid queries for SqlServer2008.


maumar:
:shipit:

Issue #6861, opened 2016-10-26
Query with projection and navigation properties adds extra sort columns in ORDER BY for foreign key columns


jemiller0:
### Steps to reproduce

Run a query that does a projection such as the following.

```
                var q = from b in lc.Bibs
                        select new
                        {
                            b.Id,
                            b.OclcNumber,
                            b.Title,
                            b.Author,
                            b.PublicationDate,
                            Language = b.Language.Name,
                            b.LanguageId,
                            Format = b.Format.Name,
                            b.FormatId,
                            BibStatus = b.BibStatus.Name,
                            b.BibStatusId,
                            b.BibStatusLastWriteTime,
                            b.BibStatusLastWriteUserName,
                            b.IsPublic,
                            b.Level,
                            b.CreationTime,
                            b.CreationUserName,
                            b.LastWriteTime,
                            b.LastWriteUserName
                        };
                q = q.OrderByDescending(b => b.LastWriteTime);
                q = q.Skip(200).Take(100);
                var l = q.ToList();

```

The following zip file contains the definitions for the POCO classes that are used by the above query.

[QueryOrderByBug.zip](https://github.com/aspnet/EntityFramework/files/551776/QueryOrderByBug.zip)
### The issue

EF Core 1.1.0-preview1-final generates the following SQL. There are a few problems with the generated SQL. The main one is that you can see that it is sorting by FormatId and LanguageId columns in addition to the LastWriteTime column which is the only one that was specified in the LINQ query. The Bibs table used below is a table containing 7.7 million rows. The query takes 12 seconds to execute. If I run the SQL using SQL Management Studio, but, comment out the extra order by columns, the query executes in 0 seconds. It looks like what it's doing is adding sort columns for the foreign key columns that are not required. The project above is also bringing in the BibStatus navigation property. That one is required and there isn't a sort for it.

A second problem is that the query is doing a projection, but, it looks like EF Core is bringing in all columns from the tables that are used. In this particular query, there is a column of type VARCHAR(MAX) that contains a lot of data. So, I want to avoid bringing that in.

[...]

PR #7026, opened 2016-11-16
Fix to #6861 - Query with projection and navigation properties adds extra sort columns in ORDER BY for foreign key columns

maumar:
When creating LEFT OUTER JOIN we add ORDER BY clause(s) based on the key of the outer entity. This is because out GroupJoin method is streaming and we need to have results in order to create the correct groups in a performant way.
However we can there are some optimizations we can do:
1. we don't need ORDER BY if the join is from dependent to principal because there will always be at most one customer associated with a given order.
e.g. `orders.Select(o => o.Customer.Name)`
which translates to:
```
from o in Orders
join c in Customers on o.CustomerId equal c.Id into grouping
from c in grouping.DefaultIfEmpty()
select c.Name
```
2. in case when we fully translate LOJ into SQL (which is now very often the case) we also don't need the extra ORDER BYs because we no longer perform streamed grouping on the client.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -711,13 +711,16 @@ namespace Microsoft.EntityFrameworkCore.Query
                         if (groupJoin)
                         {
                             var outerJoinOrderingExtractor = new OuterJoinOrderingExtractor();
-
                             outerJoinOrderingExtractor.Visit(predicate);
 
-                            foreach (var expression in outerJoinOrderingExtractor.Expressions)
+                            var previousOrderingCount = previousSelectExpression.OrderBy.Count;
+                            if (!outerJoinOrderingExtractor.DependentToPrincipalFound)
                             {
-                                previousSelectExpression
-                                    .AddToOrderBy(new Ordering(expression, OrderingDirection.Asc));
+                                foreach (var expression in outerJoinOrderingExtractor.Expressions)
+                                {
+                                    previousSelectExpression.AddToOrderBy(
+                                        new Ordering(expression, OrderingDirection.Asc));
+                                }
                             }
 
                             var additionalFromClause
@@ -784,6 +787,11 @@ namespace Microsoft.EntityFrameworkCore.Query
                                     QueriesBySource.Remove(joinClause);
 
                                     operatorToFlatten = LinqOperatorProvider.Join;
+
+                                    if (previousOrderingCount != previousSelectExpression.OrderBy.Count)
+                                    {
+                                        previousSelectExpression.RemoveRangeFromOrderBy(previousOrderingCount);
+                                    }
                                 }
                             }
                         }
@@ -861,7 +869,12 @@ namespace Microsoft.EntityFrameworkCore.Query
         {
             private readonly List<Expression> _expressions = new List<Expression>();
 
+            public bool DependentToPrincipalFound { get; private set; }
+
             public IEnumerable<Expression> Expressions => _expressions;
+
+            private IForeignKey _matchingCandidate;
+            private List<IProperty> _matchingCandidateProperties;
 
             public override Expression Visit(Expression expression)
             {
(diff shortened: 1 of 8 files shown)


maumar:
@smitpatel @anpete

maumar:
@ErikEJ @gmartinezsan @roji once this is checked in, you might need to update some of your test baselines

maumar:
@anpete @smitpatel new version up

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -869,17 +874,39 @@ public override Expression Visit(Expression expression)
...
+                return expression;
+            }
+
+            protected override Expression VisitBinary(BinaryExpression node)
+            {
+                if (node.NodeType == ExpressionType.Equal)
+                {
+                    var leftColumn = node.Left.RemoveConvert().TryGetColumnExpression();
+                    var rightColumn = node.Right.RemoveConvert().TryGetColumnExpression();
+                    if (leftColumn != null && rightColumn != null && leftColumn.Property.IsForeignKey() && rightColumn.Property.IsPrimaryKey())
```
smitpatel: It can be alternate key also. Not necessarily primary key.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -869,17 +874,39 @@ public override Expression Visit(Expression expression)
...
                     {
-                        case ExpressionType.Equal:
-                            _expressions.Add(binaryExpression.Left.RemoveConvert());
-                            return expression;
-                        case ExpressionType.AndAlso:
-                            return VisitBinary(binaryExpression);
+                        var foreignKeyPrincipalTypes = leftColumn.Property.GetContainingForeignKeys().Select(k => k.PrincipalEntityType.RootType());
+                        var primaryKeyDeclaringType = rightColumn.Property.DeclaringEntityType;
+                        if (foreignKeyPrincipalTypes.Contains(primaryKeyDeclaringType))
```
smitpatel: This logic to identify principal/dependent side does not work always. Current codebase allows creation of a model in which this logic picks up incorrect navigation hence wrong sides.
I am discussing with @AndriySvyryd for the validity of such models.

gmartinezsan:
@maumar Thanks , I will take a look!

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -861,25 +869,67 @@ private class OuterJoinOrderingExtractor : ExpressionVisitor
...
             {
                 var binaryExpression = expression as BinaryExpression;
 
-                if (binaryExpression != null)
+                if (binaryExpression != null && !DependentToPrincipalFound)
+                {
+                    return VisitBinary(binaryExpression);
+                }
+
+                return expression;
```
smitpatel: Any reason for not calling base here? I think it will not go down inside any of structure other than binary tree otherwise.
Also I assume you trying to calling `VisitBinary` only if `DependentToPrincipalFound` is false. Instead of doing it this way, you can have condition at the start of `VisitBinary` to just return node.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -861,25 +869,67 @@ private class OuterJoinOrderingExtractor : ExpressionVisitor
...
-                        case ExpressionType.Equal:
-                            _expressions.Add(binaryExpression.Left.RemoveConvert());
-                            return expression;
-                        case ExpressionType.AndAlso:
-                            return VisitBinary(binaryExpression);
+                        var keyDeclaringEntityType = rightProperty.GetContainingKeys().First().DeclaringEntityType;
+                        var matchingForeignKeys = leftProperty.GetContainingForeignKeys().Where(k => k.PrincipalKey.DeclaringEntityType == keyDeclaringEntityType);
+                        if (matchingForeignKeys.Count() == 1 && rightProperty.GetContainingKeys().Contains(matchingForeignKeys.Single().PrincipalKey))
+                        {
+                            var matchingForeignKey = matchingForeignKeys.Single();
```
smitpatel: twice call to `Single()` store in a variable.

smitpatel:
Visitor may not visit anything beyond binary expression. Is it by design?

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
maumar: There should be no reason to visit every node that is not BinaryExpression. Top level node is always == or || and in the VisitBinary, if the children are not BinaryExpressions themselves we add them to the order by list and return (without visiting). Will move DependentToPrincipalFound check to VisitBinary as suggested

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
maumar: 👍

PR #7044, opened 2016-11-18
Make DatabaseGeneratedOption.Identity not throw for non-integer properties on SQL Server

### LINKED ISSUE #7010: [DatabaseGenerated(DatabaseGeneratedOption.Identity)] incorrectly configures the column as SqlServer identity
### Steps to reproduce
I tried using the `[DatabaseGenerated(DatabaseGeneratedOption.Identity)]` attribute on a DateTime (which has a default value constraint on the column) as described in the [Generated Properties](https://docs.microsoft.com/en-us/ef/core/modeling/generated-properties) documentation and I got an error as shown below:

Personally I think the documentation is wrong, as the `[DatabaseGenerated(DatabaseGeneratedOption.Identity)]` attribute is pretty specific. I successfully used the Fluent API  .ValueGeneratedOnAdd() to tell EF Core there was a default value constraint.

### The issue
1. Create a table with a column 'DateOfBirth' with a default constraint on a `[datetime2]` column, e.g.  
```
[DateOfBirth] [datetime2](7) NOT NULL DEFAULT ('2000-01-01T00:00:00.000'),
```
2. Map that column to a property 'DateOfBirth'
3. Use the `[DatabaseGenerated(DatabaseGeneratedOption.Identity)]`  attribute to tell EF Core that the column has a default constraint.
4. Create the DbContext and it fails with the following stack trace

```
Exception message: Identity value generation cannot be used for the property 'DateOfBirth' on entity type 'PersonWithAddUpdateAttibutes' because the property type is 'DateTime'. Identity value generation can only be used with signed integer properties.
Stack trace:
   at Microsoft.EntityFrameworkCore.Metadata.SqlServerPropertyAnnotations.SetValueGenerationStrategy(Nullable`1 value)
   at Microsoft.EntityFrameworkCore.Metadata.Internal.SqlServerPropertyBuilderAnnotations.ValueGenerationStrategy(Nullable`1 value)
   at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.SqlServerValueGenerationStrategyConvention.Apply(InternalPropertyBuilder propertyBuilder, DatabaseGeneratedAttribute attribute, MemberInfo clrMember)
   at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.PropertyAttributeConvention`1.Apply(InternalPropertyBuilder propertyBuilder)
   at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.ConventionDispatcher.OnPropertyAdded(InternalPropertyBuilder propertyBuilder)
   at Microsoft.EntityFrameworkCore.Metadata.Internal.EntityType.AddProperty(String name, Type propertyType, MemberInfo memberInfo, ConfigurationSource configurationSource, Boolean runConventions)
   at Microsoft.EntityFrameworkCore.Metadata.Internal.EntityType.AddProperty(MemberInfo memberInfo, ConfigurationSource configurationSource, Boolean runConventions)
   at Microsoft.EntityFrameworkCore.Metadata.Internal.InternalEntityTypeBuilder.Property(Property existingProperty, String propertyName, Type propertyType, MemberInfo clrProperty, Nullable`1 configurationSource)
   at Microsoft.EntityFrameworkCore.Metadata.Internal.InternalEntityTypeBuilder.Property(String propertyName, Type propertyType, MemberInfo clrProperty, Nullable`1 configurationSource)
   at Microsoft.EntityFrameworkCore.Metadata.Internal.InternalEntityTypeBuilder.Property(MemberInfo clrProperty, ConfigurationSource configurationSource)
   at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.PropertyDiscoveryConvention.Apply(InternalEntityTypeBuilder entityTypeBuilder)
   at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.ConventionDispatcher.OnEntityTypeAdded(InternalEntityTypeBuilder entityTypeBuilder)
   at Microsoft.EntityFrameworkCore.Metadata.Internal.Model.AddEntityType(EntityType entityType, Boolean runConventions)
   at Microsoft.EntityFrameworkCore.Metadata.Internal.InternalModelBuilder.Entity(Type type, ConfigurationSource configurationSource, Boolean runConventions)
   at Microsoft.EntityFrameworkCore.ModelBuilder.Entity(Type type)
   at Microsoft.EntityFrameworkCore.Infrastructure.ModelSource.FindSets(ModelBuilder modelBuilder, DbContext context)
   at Microsoft.EntityFrameworkCore.Infrastructure.Internal.RelationalModelSource.FindSets(ModelBuilder modelBuilder, DbContext context)
   at Microsoft.EntityFrameworkCore.Infrastructure.ModelSource.CreateModel(DbContext context, IConventionSetBuilder conventionSetBuilder, IModelValidator validator)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Microsoft.EntityFrameworkCore.Internal.DbContextServices.CreateModel()
   at Microsoft.EntityFrameworkCore.Internal.LazyRef`1.get_Value()
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.<>c__DisplayClass16_0.<RealizeService>b__0(ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Microsoft.EntityFrameworkCore.DbContext.get_StateManager()
   at Microsoft.EntityFrameworkCore.DbContext.EntryWithoutDetectChanges[TEntity](TEntity entity)
   at Microsoft.EntityFrameworkCore.DbContext.SetEntityState[TEntity](TEntity entity, EntityState entityState)
   at test.UnitTests.TestPeopleContext.TestPeopleContextExistingDatabaseOk() in C:\Users\Jon\Documents\Visual Studio 2015\Projects\EfCoreInAction\test\UnitTests\TestPeopleContext.cs:line 111
```

### Further technical details

EF Core version: "Microsoft.EntityFrameworkCore.SqlServer": "1.1.0-preview1-final",
Operating system: Windows 10

### LINKED ISSUE #7049: Exception says another instance has same key, but it doesn't
The issue can be reproduced with this version:

	"Microsoft.EntityFrameworkCore.SqlServer": "1.2.0-preview*",

and this context:

	{
			: base( options ) { }

		public DbSet<L> Ls { get; set; }
		public DbSet<AL> ALs { get; set; }
	}

and this model:

	public class L
	{
		public int ID { get; set; }
	}

	public class T : P
	{
		public Q D { get; set; }
		public P P { get; set; }
		public Q F { get; set; }
	}

	public class P : PBase { }

	public class Q : PBase { }

	public abstract class PBase
	{
		public int ID { get; set; }
		public string Stuff { get; set; }
	}

	public class AL
	{
		public int ID { get; set; }
		public PBase L { get; set; }
	}

The "seeding" code is:

	dataContext.Ls.Add(
		new L {
			Ts = new[] { new T { D = new Q(), F = new Q() } }
		}
	);

The exception is:
``` 
System.InvalidOperationException was unhandled by user code
  HResult=-2146233079
  Message=The instance of entity type 'Q' cannot be tracked because another instance of this type with the same key is already being tracked. When adding new entities, for most key types a unique temporary key value will be created if no key is set (i.e. if the key property is assigned the default value for its type). If you are explicitly setting key values for new entities, ensure they do not collide with existing entities or temporary values generated for other new entities. When attaching existing entities, ensure that only one entity instance with a given key value is attached to the context.
  Source=Microsoft.EntityFrameworkCore
  StackTrace:
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(InternalEntityEntry entry)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.StartTracking(InternalEntityEntry entry)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState entityState, Boolean acceptChanges)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityGraphAttacher.PaintAction(EntityEntryGraphNode node)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph(EntityEntryGraphNode node, Func`2 handleNode)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph(EntityEntryGraphNode node, Func`2 handleNode)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph(EntityEntryGraphNode node, Func`2 handleNode)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityGraphAttacher.AttachGraph(InternalEntityEntry rootEntry, EntityState entityState)
       at Microsoft.EntityFrameworkCore.DbContext.SetEntityState(InternalEntityEntry entry, EntityState entityState)
       at Microsoft.EntityFrameworkCore.DbContext.SetEntityState[TEntity](TEntity entity, EntityState entityState)
       at Microsoft.EntityFrameworkCore.DbContext.Add[TEntity](TEntity entity)
       at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.Add(TEntity entity)
       at EFGetStarted.AspNetCore.NewDb.Startup.Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory, BloggingContext dataContext) in C:\Users\Adrian\Development\Examples\ASP.NET Core\EFGetStarted.AspNetCore.NewDb\src\EFGetStarted.AspNetCore.NewDb\Startup.cs:line 61
  InnerException: 
```

AndriySvyryd:
Make Property.RequiresValueGenerator to be a readonly extension method, this disables client-side value generation for non-key properties that don't have a custom value generator.
Make DatabaseGeneratedOption.Identity attribute just set ValueGenerated to OnAdd on SQL Server
Update ValueGenerated using conventions more consistently
Make the temporality of the client generated values depend on whether a value will be generated by the store
Add a test relational provider so that the purely relational tests can use the correct conventions

Part of #7010
Fixes #7049


AndriySvyryd:
Simplified the code by removing the ability to mutate `RequiresValueGenerator`

AndriySvyryd:
Made `ValueGenerated` only apply to store generated values

ajcvickers:
I'm having difficulty correlating these changes to what we discussed on Wednesday last week. Can we talk about it when you are back in the office?

AndriySvyryd:
Updated to be closer to the current behaviour

src/Microsoft.EntityFrameworkCore.Relational/Metadata/Conventions/Internal/RelationalValueGeneratorConvention.cs
```diff
@@ -0,0 +1,56 @@
...
+        protected override ValueGenerated? GetValueGenerated(Property property)
+        {
+            var valueGenerated = base.GetValueGenerated(property);
+            if (valueGenerated != null)
+            {
+                return valueGenerated;
+            }
+
+            var relationalProperty = AnnotationProvider.For(property);
+            return relationalProperty.ComputedColumnSql != null
```
ajcvickers: Previously we would do this immediately in the API. What is the advantage to doing it in a convention like this?

src/Microsoft.EntityFrameworkCore.SqlServer/Metadata/Conventions/Internal/SqlServerValueGeneratorConvention.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    public class SqlServerValueGeneratorConvention : RelationalValueGeneratorConvention
```
ajcvickers: Given what we discussed, why is it that SQL Server needs its own convention here?

ajcvickers:
Do you think it would be useful to split out fixing bug from changes to do things differently? It seems we're making a lot of changes to code for very minimal changes in behavior. Maybe it would help to create separate issues for things that are currently broken so we can see what actually needs to be fixed?

src/Microsoft.EntityFrameworkCore.Relational/Metadata/Conventions/Internal/RelationalValueGeneratorConvention.cs
AndriySvyryd: This way it will use the current provider-specific annotation value
So if the property is configured with SqlServerDefaultValue, and later with ComputedColumnSql currently it would end up being ValueGenerated.OnAddOrUpdate, but with this change it would be ValueGenerated.OnAdd.

---
In reply to: [91350229](https://github.com/aspnet/EntityFramework/pull/7044#discussion_r91350229) [](ancestors = 91350229)

src/Microsoft.EntityFrameworkCore.SqlServer/Metadata/Conventions/Internal/SqlServerValueGeneratorConvention.cs
AndriySvyryd: It's needed to mark key properties as ValueGenerated.OnAdd when SqlServerValueGenerationStrategy is configured on them

---
In reply to: [91352785](https://github.com/aspnet/EntityFramework/pull/7044#discussion_r91352785) [](ancestors = 91352785)

ajcvickers: So it wouldn't be needed if setting SqlServerValueGenerationStrategy marked the property as ValueGenerated.OnAdd?

AndriySvyryd:
It would probably take me a day and a half, but I can do it if you consider this to be unreviewable in the current state.

---
In reply to: [265527158](https://github.com/aspnet/EntityFramework/pull/7044#issuecomment-265527158) [](ancestors = 265527158)

ajcvickers:
I think we need to be more clear about what issues are being fixed and why. For example, I don't think that the things with DefaultValue verse ComputedValue for different providers is something that needs to be fixed because its not a realistic scenario--the model is fundamentally inconsistent whatever we do. Further, I think there are better ways of fixing this--for example, removing the underlying issue of trying to build one model for everything. With the review the way it is I can't figure out which changes are for which issues.

src/Microsoft.EntityFrameworkCore.Relational/Metadata/Conventions/Internal/RelationalValueGeneratorConvention.cs
AndriySvyryd: Ok, but independently of what we choose to be the correct behavior here the convention has more data available to it to consider.
Another issue is that currently ValueGenerated is only set from Fluent API, and not the Metadata API as we had decided not to put any conventional behavior there.

---
In reply to: [91356923](https://github.com/aspnet/EntityFramework/pull/7044#discussion_r91356923) [](ancestors = 91356923)

src/Microsoft.EntityFrameworkCore.SqlServer/Metadata/Conventions/Internal/SqlServerValueGeneratorConvention.cs
AndriySvyryd: Yes

---
In reply to: [91355358](https://github.com/aspnet/EntityFramework/pull/7044#discussion_r91355358) [](ancestors = 91355358)

AndriySvyryd:
Will split into smaller PRs, but feel free to keep commenting here.

ajcvickers:
@AndriySvyryd Consider creating issues before PRs so we can decide as a team which changes we want to take.

Issue #7003, opened 2016-11-11
Wrong SQL generated for query with group join on a subquery that is not present in the final projection


hugoterelle:
Hi,
I have 2 entities with the following code:

```
	public class Brand : RootEntity, IDisabledEntity
	{
		public string Name { get; set; }
		public bool Disabled { get; set; }
		public virtual ICollection<Smartphone> Smartphones { get; set; }
	}

	public class Smartphone : RootEntity, IDisabledEntity
	{
		public string Model { get; set; }
		public virtual Brand Brand { get; set; }
		public bool Disabled { get; set; }
	}
```
Where RootEntity and Entity are

```
    public class RootEntity : Entity, IAuditableEntity
    {
        public DateTime CreatedOn { get; set; }
        public DateTime ModifiedOn { get; set; }
    }
    public class Entity
    {
        public int Id { get; set; }
    } 
```
So nothing special.  Just a one-to-many relationship between Brand and Smartphone.
The resulting SQL table are:

```
CREATE TABLE "Brands" 
( "Id" INTEGER NOT NULL CONSTRAINT "PK_Brands" PRIMARY KEY AUTOINCREMENT, 
"CreatedOn" TEXT NOT NULL,
 "Disabled" INTEGER NOT NULL, 
"ModifiedOn" TEXT NOT NULL, 
"Name" TEXT )

CREATE TABLE "Smartphones" 
( "Id" INTEGER NOT NULL CONSTRAINT "PK_Smartphones" PRIMARY KEY AUTOINCREMENT, 
"BrandId" INTEGER, 
"CreatedOn" TEXT NOT NULL, 
"Disabled" INTEGER NOT NULL, 
"Model" TEXT, 
"ModifiedOn" TEXT NOT NULL, 
CONSTRAINT "FK_Smartphones_Brands_BrandId" FOREIGN KEY ("BrandId") REFERENCES "Brands" ("Id") ON DELETE RESTRICT )
```

I try to retrieve my "Smartphone" list with the following parameters:

* sort by brand name, then by smartphone model
* skip at least 'skip' rows
* take maximum 'take' rows

Here is the LINQ code below:

```
		public Task<List<SmartphoneDto>> GetAll(int skip, int take)
		{
			return DataContext.Set<Smartphone>()
				.Where(x => !x.Disabled)
				.Include(x => x.Brand)
				.OrderBy(x => x.Brand.Name).ThenBy(x => x.Model)
				.Skip(skip) 
[...]

PR #7046, opened 2016-11-18
[SMALL] Fix to #7003 - Wrong SQL generated for query with group join on a subquery that is not present in the final projection

maumar:
Problem was that for GroupJoin we need to mark both inputs for materialization, so we have all the elements necessary to produce joining calls.
However, if one side is a subquery, we just mark that outer query source for materialization, without digging deep into the subquery itself. This results in the actual table (which is inside) not being materialized, and that in turn produces SQL which doesn't have necessary elements to produce the correct joins.

Fix is to check if the groupjoin element in a subquery, and if so recursively mark it's elements for materialization.


maumar:
@anpete @smitpatel

smitpatel:
:shipit:

PR #7074, opened 2016-11-20
Fix #7073 AfterExecuteCommand tracing is missing the result info.

### LINKED ISSUE #7073: AfterExecuteCommand tracing is missing the result info
[Here](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs#L250) we are missing the `result` value that was set above. It is should be included in the tracing as it is needed in any case that someone will have to trace this info.

fitzchak:
Fix #7073 AfterExecuteCommand tracing is missing the result info. Now the result is added and can be used by applications that need to trace this info.


fitzchak:
Hello,

This PR is not needed, as the change in #7105 if fixing this issue also, it let us get the correct SelectRows for the statistics.

Issue #7048, opened 2016-11-18
EF Core 1.1 .ToString() translation causes exception


dotnetshadow:
### Steps to reproduce
Create a basic asp.net application for one of the controllers just insert the query code below with the model provided. As soon as the query is run the exception takes place

### The issue
The following query works in .net core **1.0.1** but doesn't work in **1.1**

**Model**
```
 public class Supplier
    {
        [Key]        
        public Guid Id { get; set; }

        [Required]
        public string Name { get; set; }

        [EmailAddress]        
        public string Email { get; set; }
        
    }
```

**Query**
```
var supplierList = _context.Suppliers.OrderBy(x => x.Name).Select(x => new SelectListItem
{
   Text = x.Name,
   Value = x.Id.ToString()
}).ToList();
```

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
Exception message:
NotSupportedException: TryExpression is not supported as a child expression when accessing a member on type 'System.Guid' because it is a value type. Construct the tree so the TryExpression is not nested inside of this expression.

System.NotSupportedException: TryExpression is not supported as a child expression when accessing a member on type 'System.Guid' because it is a value type. Construct the tree so the TryExpression is not nested inside of this expression.

   at System.Linq.Expressions.Compiler.StackSpiller.RequireNotRefInstance(Expression instance)

   at System.Linq.Expressions.Compiler.StackSpiller.RewriteMethodCallExpression(Expression expr, Stack stack)

   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)

   at System.Linq.Expressions.Compiler.StackSpiller.MemberAssignmentRewriter..ctor(MemberAssignment binding, StackSpiller spiller, Stack stack)

   at System.Linq.Expressions.Compiler.StackSpiller.BindingRewriter.Create(MemberBinding binding, StackSpiller spiller, Stack stack)

   at System.Linq.Expressions.Compiler.StackSpiller.RewriteMemberInitExpression(Expression expr, Stack stack)

   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)

   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpressionFreeTemps(Expression expression, Stack stack)

   at System.Linq.Expressions.Compiler.StackSpiller.Rewrite[T](Expression`1 lambda)

[...]

PR #7090, opened 2016-11-21
Fix: #7048 - EF Core 1.1 .ToString() translation causes exception

anpete:
Adding bad data error handling logic to materialization introduced the possibility
of hitting a known expression compiler limitation (as described here: https://github.com/dotnet/corefx/pull/13126).

This change works around the limitation by moving the read-value try-catch to a helper method, thus we no
longer insert a try-catch directly into the output ET.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
@@ -131,10 +131,26 @@ select c'
             materializer: (ValueBuffer valueBuffer) => 
             {
                 var3 = new Order()
-                var3.<OrderID>k__BackingField = try { (int) object valueBuffer.get_Item(0) } catch (Exception) { ... } 
-                var3.<CustomerID>k__BackingField = try { (string) object valueBuffer.get_Item(1) } catch (Exception) { ... } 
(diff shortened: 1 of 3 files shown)


test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
```diff
@@ -131,10 +131,26 @@ SELECT 1
             materializer: (ValueBuffer valueBuffer) => 
             {
                 var3 = new Order()
-                var3.<OrderID>k__BackingField = try { (int) object valueBuffer.get_Item(0) } catch (Exception) { ... }
```
maumar: we should also remove the special handling of try-catch in expression printer

maumar: ...and maybe make it so that TryReadValue method always prints in a single line, and without parameter names - so that the query plan doesn't get bloated too much

maumar:
:shipit: after changes in the ExpressionPrinter

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
anpete: I think we should minimize the changes for 1.1.1 but we can do it for 1.2 - sound good?

PR #7091, opened 2016-11-21
Query: Improve detection of composible FromSql queries

### LINKED ISSUE #7005: FromSql not composing with LINQ
Hi

Following `FromSql` query does not compose with LINQ. No exception is raised and the result is correct BUT `Sql Server Profiler` shows that `Where` clause is not composing with `FromSql` query.

``` c#
var result = dbContext.Permission
                .FromSql($@"
select
p.*
from Permission p
inner join
(
    select 
    r.RoleId 
    from 
    Role r
    where 
    Parents.exist('//I[.=sql:variable(""{{0}}"")]') = 1
    union all
    select {{0}} RoleId
) roles
on p.RoleId = roles.RoleId
", roleId)
                    .Where(c =>
                        c.Entity == entity &&
                        c.RoleId == roleId &&
                        c.Type == type &&
                        c.EntityId == entityId)
                        .ToList()
```

but the query in `Sql Server Profiler` is:

```sql
exec sp_executesql N'
select
p.*
from Permission p
inner join
(
    select 
    r.RoleId 
    from 
    Role r
    where 
    Parents.exist(''//I[.=sql:variable("@p0")]'') = 1
    union all
    select @p0 RoleId
) roles
on p.RoleId = roles.RoleId
',N'@p0 int',@p0=170
```

which shows `Where` clause is not included in query

EF Core version:1.0.1
Operating system: windows 10
Visual Studio version: vs 2015

smitpatel:
Resolves #7005

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/FromSqlQueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/FromSqlQueryTestBase.cs
@@ -199,6 +199,25 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
+        public virtual void From_sql_queryable_composed_after_removing_whitespaces()
+        {
+            using (var context = CreateContext())
+            {
+                var actual = context.Set<Customer>()
+                    .FromSql(@"
+    
+
+
+SELECT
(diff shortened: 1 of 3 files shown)


smitpatel:
@maumar - Updated.
Manually verified behavior with \t

Issue #7061, opened 2016-11-19
The binary operator Equal is not defined for the types 'System.Nullable`1[System.Boolean]' and 'System.Boolean' error for queries with complex predicates involving nullable boolean properties


mo-esmp:
I've updated EF from ver 1.0.1 to ver 1.1.0 and after runnig below query I got error.
```
var data = await _context.CustomerServices
                .Include(a => a.ServiceEquipment)
                .Include(a => a.ServiceAddresses)
                .ThenInclude(a => a.Address)
                .SingleOrDefaultAsync(a => a.Id == serviceId && !a.IsDeleted && !a.ServiceEquipment.IsDeleted);
```

> the binary operator Equal is not defined for the types 'System.Nullable`1[System.Boolean]' and 'System.Boolean'.
> 
>    at System.Linq.Expressions.Expression.GetEqualityComparisonOperator(ExpressionType binaryType, String opName, Expression left, Expression right, Boolean liftToNull)
>    at System.Linq.Expressions.Expression.Equal(Expression left, Expression right, Boolean liftToNull, MethodInfo method)
>    at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.SearchConditionTranslatingVisitor.VisitUnary(UnaryExpression expression)
>    at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.SearchConditionTranslatingVisitor.VisitBinary(BinaryExpression expression)
>    at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.SearchConditionTranslatingVisitor.VisitUnary(UnaryExpression expression)
>    at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.ApplyOptimizations(Expression expression, Boolean searchCondition, Boolean joinCondition)
>    at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.VisitSelect(SelectExpression selectExpression)
>    at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
>    at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.GenerateSql(IReadOnlyDictionary`2 parameterValues)
>    at Microsoft.EntityFrameworkCore.Query.Internal.ShaperCommandContext.GetRelationalCommand(IReadOnlyDictionary`2 parameters)
[...]

PR #7093, opened 2016-11-22
Fix to #7061 - the binary operator Equal is not defined for the types 'System.Nullable`1[System.Boolean]' and 'System.Boolean'

maumar:
Problem was that when we were fixing search conditions on a query, we did not take into account differences in nullability that could be created by those transformations.

e.g.:
myEntity.MyNullableBool (result type is bool?)

would get converted to:
myEntity.MyNullableBool == (bool?)true (result type bool)

If that in turn was a part of a bigger expression, e.g. AndAlso with another bool? expression, that did not have to be converted, then we would produce a type discrepancy which was resulting in an error.

Fix is to adjust types of the Left and Right expressions that we process before reconstructing a binary expression node.


maumar:
@smitpatel

PR #7099, opened 2016-11-22
Query: Remove Mutation of SelectExpression from ExistExpression

smitpatel:
@anpete


src/Microsoft.EntityFrameworkCore.SqlServer/Query/Internal/SqlServerQueryModelVisitor.cs
```diff
@@ -116,6 +116,20 @@ private class RowNumberPagingExpressionVisitor : ExpressionVisitorBase
         {
             public override Expression Visit(Expression node)
             {
+                var existExpression = node as ExistsExpression;
```
anpete: Introduce a private VisitExistsExpression the same way as VisitSelectExpression is being handled.

anpete:
:shipit:

PR #7106, opened 2016-11-23
#7105 Trace when query is done. This is needed in order to get Select…

### LINKED ISSUE #7105: Trace when query is done. This is needed in order to get SelectRows statistics.
In order to get the SelectRows from SqlStatistics we need to add this trace statement, since "System.Data.SqlClient.WriteCommandAfter" tracing is not enough, as this is before the reader is reading the results.
This let us in Entity Framework Profiler report the results for the query.

fitzchak:
Fix #7105

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
@@ -226,7 +226,8 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
                                 = new RelationalDataReader(
                                     connection,
                                     dbCommand,
-                                    dbCommand.ExecuteReader());
+                                    dbCommand.ExecuteReader(),
+                                    DiagnosticSource);
                         }
                         catch
                         {
@@ -344,7 +345,8 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
                             result = new RelationalDataReader(
                                 connection,
                                 dbCommand,
-                                await dbCommand.ExecuteReaderAsync(cancellationToken));
+                                await dbCommand.ExecuteReaderAsync(cancellationToken), 
+                                DiagnosticSource);
                         }
                         catch
                         {
--- a/src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalDataReader.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalDataReader.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Data.Common;
+using System.Diagnostics;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 
 namespace Microsoft.EntityFrameworkCore.Storage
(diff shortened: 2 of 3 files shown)


fitzchak:
Applied requested changes.

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalDataReader.cs
```diff
@@ -31,17 +34,20 @@ public class RelationalDataReader : IDisposable
         /// <param name="connection"> The connection. </param>
         /// <param name="command"> The command that was executed. </param>
         /// <param name="reader"> The underlying reader for the result set. </param>
+        /// <param name="diagnosticSource"> The diagnostic source. </param>
         public RelationalDataReader(
             [CanBeNull] IRelationalConnection connection,
             [NotNull] DbCommand command,
-            [NotNull] DbDataReader reader)
+            [NotNull] DbDataReader reader,
+            [NotNull] DiagnosticSource diagnosticSource)
```
anpete: Add Check.NotNull

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
```diff
@@ -226,7 +226,7 @@ object IRelationalCommand.ExecuteScalar(IRelationalConnection connection, IReadO
                                 = new RelationalDataReader(
                                     connection,
                                     dbCommand,
-                                    dbCommand.ExecuteReader());
+                                    dbCommand.ExecuteReader(), DiagnosticSource);
```
anpete: Move arg to newline.

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalDataReader.cs
```diff
@@ -65,6 +71,7 @@ public virtual void Dispose()
         {
             if (!_disposed)
             {
+                _diagnosticSource.WriteDataReaderDisposed(this);
```
anpete: Looks like we are firing the event before disposing the underlying reader, which is probably more useful anyway. I think we should tweak the event name to reflect that. I.e. "Disposing" instead of "Disposed".

anpete: Do you actually need the reader instance for your use case?

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalDataReader.cs
fitzchak: Done

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalDataReader.cs
fitzchak: Yes, we need it before the disposing. So I changed the name to "Disposing".
Also we need just the instance of `DbDataReader`, which has the command with the statistics and the connection. So I changed this also to send just the  DbDataReader.

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
fitzchak: Done

fitzchak:
I amended the commit with the requested changes.

Issue #7112, opened 2016-11-23
Redundant call to String.Format() in the RelationalLogger class


lecaillon:
### The issue
String.Format is called 2 times in the LogCommandExecuted function of the RelationalLoggerExtensions class. 
Remove 1 call.
```
public static void LogCommandExecuted(...)
{
...
return RelationalStrings.RelationalLoggerExecutedCommand(
                string.Format(CultureInfo.InvariantCulture, 
                string.Format(CultureInfo.InvariantCulture, "{0:N0}", elapsedMilliseconds)),
}
```

smitpatel:
Merged via ab4b6dc6038a674e055c51a6d28ad15ffe75e246

PR #7113, opened 2016-11-23
Fix to #7112 - Remove redundant call to String.Format() in LogCommand…

lecaillon:
### The fix
Just remove the unnecessary call to String.Format() ...

What a big performance boost ;-)


smitpatel:
merged via ab4b6dc6038a674e055c51a6d28ad15ffe75e246

PR #7151, opened 2016-11-28
Fix #7115 - DB query is executed with the parameters of the previous …

### LINKED ISSUE #7115: DB query is executed with the parameters of the previous query
### Steps to reproduce

Minimal solution: https://github.com/linnando/EFCoreTest

Model.cs contains the description of the model copied from https://docs.microsoft.com/en-us/ef/core/index

To reproduce the bug:
1. Execute tests in Deployment.cs (they will apply migrations and seed the test data).
2. Execute tests in MainTest.cs. Each of two tests queries the database twice. After each query, the count of found records is written to the debug trace. The first query looks for a blog with url "test" (which exists in the database) and the second one --- for a blog with url "wrong_url" (which does not). So, the expected output is 1 and 0 (in each test).

### The issue

TestGetByQueryableUrl outputs 1 and 1 instead of expected 1 and 0. In the debug mode, I see that in _both_ cases, the same query is sent to the database: "SELECT COUNT(*) FROM [Blogs] AS [blog] WHERE [blog].[Url] IN (N'test')". Although in the second time another parameter "wrong_url" is given, the query is still sent to the database with parameter "test" instead (probably, somehow cached from the previous query).

In contrast, TestGetByEnumerableUrl works correctly (sends correct queries and outputs 1 and 0), although it differs from the first test only in the signature of the auxiliary method (GetBlogsByEnumerableUrl).

### Further technical details

EF Core version: 1.1.0
Operating system: Windows 10 Enterprise N
Visual Studio version: VS Enterprise 2015 (14.0.25431.01 Update 3)

anpete:
- Fixes an issue in ExpressionEqualityComparer where we would incorrectly determine equality for constant EnumerableQuery nodes.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
@@ -43,6 +43,24 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [ConditionalFact]
+        public virtual void Method_with_constant_queryable_arg()
+        {
+            using (var context = CreateContext())
+            {
+                var count = QueryableArgQuery(context, new [] { "ALFKI" }.AsQueryable()).Count();
+
+                Assert.Equal(1, count);
+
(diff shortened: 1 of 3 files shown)


anpete:
@smitpatel

PR #7200, opened 2016-12-07
Make DatabaseGeneratedOption.Identity attribute just set ValueGenerated to OnAdd on SQL Server

AndriySvyryd:
This will go in 1.1.1

Part of #7010


AndriySvyryd:
@ajcvickers Could you take a look?

AndriySvyryd:
@smitpatel Could you take a look since @ajcvickers is OOF?

PR #7246, opened 2016-12-14
Moved tests to Specification tests to avoid code copying in providers

### LINKED ISSUE #6176: Publish Microsoft.EntityFrameworkCore.Relational.Tests as a nuget?
EFCore already provider the specification test suites as nuget packages for consumption by providers. This is great as it allows providers to automatically check if they implement EFCore behavior correctly, etc.

There's also Microsoft.EntityFrameworkCore.Relational.Tests, which contains unit (rather than functional/integration) tests on EFCore components. Aside from running these tests directly, base classes in this project are also used by provider-specific unit test suites (e.g. Microsoft.EntityFrameworkCore.SqlServer.Tests.Migrations) to test their own implementations of various EFCore relational components. Since this package isn't published as a nuget it's not possible for external providers to use it, only SqlServer and Sqlite which happen to be in the solution.

It seems like a refactor may be in order: split Microsoft.EntityFrameworkCore.Relational.Tests into one project which contains actual tests to run on relational components, and another project which contains base classes for consumption by providers (and published as a nuget).

This obviously isn't very urgent, I'm currently working around it by copying source files into my own project.

I asked for this before in https://github.com/aspnet/EntityFramework/issues/3016#issuecomment-139851543 but there were definitely more important things back then.

ErikEJ:
Moved tests to Specification tests to avoid code copying in providers 
Fixes #6176


ajcvickers:
Merged. Thanks Erik!

PR #7287, opened 2016-12-21
Make Property.RequiresValueGenerator calculated

### LINKED ISSUE #7204: Make RequiresValueGenerator readonly
Maintaining the correct RequiresValueGenerator value adds complexity, but is only needed to use a provider-specific ValueGenerator (e.g. SqlServerSequenceHiLo) on a property that wouldn't need a ValueGenerator according to core rules (such as a non-key property)

AndriySvyryd:
Fixes #7204


src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerModelValidator.cs
```diff
@@ -38,6 +39,17 @@ protected virtual void EnsureNoDefaultDecimalMapping([NotNull] IModel model)
             }
         }
 
+        protected virtual void EnsureNoNonKeyValueGeneration([NotNull] IModel model)
+        {
+            foreach (var property in model.GetEntityTypes().SelectMany(t => t.GetDeclaredProperties())
+                .Where(p => (p.SqlServer().ValueGenerationStrategy == SqlServerValueGenerationStrategy.SequenceHiLo
+                             || p.SqlServer().ValueGenerationStrategy == SqlServerValueGenerationStrategy.IdentityColumn)
+                            && !p.IsKey()))
```
ajcvickers: Will this throw if the strategy is set on the model or entity type?

src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/ValueGenerationManager.cs
```diff
@@ -98,8 +98,7 @@ public virtual void Generate(InternalEntityEntry entry)
             foreach (var property in entry.EntityType.GetProperties())
             {
                 var isForeignKey = property.IsForeignKey();
-
-                if ((property.RequiresValueGenerator || isForeignKey)
+                if ((property.RequiresValueGenerator() || isForeignKey)
                     && property.ClrType.IsDefaultValue(entry[property]))
```
ajcvickers: Does this work for cases where we generated an PK for a dependent, but now want it to be updated because it is also an FK and the principal has now been added?

src/Microsoft.EntityFrameworkCore/Extensions/PropertyExtensions.cs
```diff
@@ -31,6 +31,16 @@ public static class PropertyExtensions
         }
 
         /// <summary>
+        ///     Gets a value indicating whether this property requires a <see cref="ValueGenerator" /> to generate
+        ///     values when new entities are added to the context.
+        /// </summary>
+        public static bool RequiresValueGenerator([NotNull] this IProperty property)
```
ajcvickers: Does this need to be public API?

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
```diff
@@ -228,13 +228,12 @@ public virtual PropertyBuilder IsConcurrencyToken(bool concurrencyToken = true)
...
         /// <summary>
-        ///     Configures a property to never have a value generated when an instance of this
+        ///     Configures a property to never have a value generated by the database when an instance of this
         ///     entity type is saved.
         /// </summary>
         /// <returns> The same builder instance so that multiple configuration calls can be chained. </returns>
         /// <remarks>
-        ///     Note that temporary values may still be generated for use internally before a
-        ///     new entity is saved.
+        ///     Note that values may still be generated by a client-side value generator.
```
ajcvickers: But only if one is explicitly set, right? If so, I don't think this line is helpful.

test/Microsoft.EntityFrameworkCore.FunctionalTests/CustomValueGeneratorTest.cs
```diff
@@ -17,52 +17,6 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
     public class CustomValueGeneratorTest
     {
         [Fact]
-        public void Can_use_custom_value_generators()
```
ajcvickers: Why is this test removed?

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlServerValueGenerationScenariosTest.cs
```diff
@@ -560,34 +598,77 @@ public void Insert_with_client_generated_GUID_key()
...
+                using (var context = new BlogContextClientGuidNonKey(testStore.Name))
+                {
+                    context.Database.EnsureCreated();
+
+                    var blog = context.Add(new GuidBlog { Name = "One Unicorn" }).Entity;
+
+                    Assert.Equal(default(Guid), blog.NotId);
+
+                    // No value set on a required column
+                    Assert.Throws<DbUpdateException>(() => context.SaveChanges());
```
ajcvickers: Why does this throw?

src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerModelValidator.cs
AndriySvyryd: Yes and there's a test for this.

src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/ValueGenerationManager.cs
AndriySvyryd: The new value would be propagated, so it won't use this code path. I've seen the test for this, but can't locate it atm.

src/Microsoft.EntityFrameworkCore/Extensions/PropertyExtensions.cs
AndriySvyryd: I'll move it to Internal.

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
```diff
@@ -228,13 +228,12 @@ public virtual PropertyBuilder IsConcurrencyToken(bool concurrencyToken = true)
         }
 
         /// <summary>
-        ///     Configures a property to never have a value generated when an instance of this
+        ///     Configures a property to never have a value generated by the database when an instance of this
```
AndriySvyryd: Because it can still be generated on the client

test/Microsoft.EntityFrameworkCore.FunctionalTests/CustomValueGeneratorTest.cs
AndriySvyryd: Because a custom value generator cannot be used for a non-key property anymore. I'll add it back with a key over the properties.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlServerValueGenerationScenariosTest.cs
AndriySvyryd: Because there was no value set on a required column as the comment states

src/Microsoft.EntityFrameworkCore.SqlServer.Design/Internal/SqlServerScaffoldingModelFactory.cs
```diff
@@ -258,6 +257,10 @@ private void VisitDefaultValue(PropertyBuilder propertyBuilder, ColumnModel colu
                     {
                         propertyBuilder.HasDefaultValueSql(defaultExpression);
                     }
+                    else
+                    {
+                        ((Property)propertyBuilder.Metadata).SetValueGenerated(null, ConfigurationSource.Explicit);
```
AndriySvyryd: To avoid marking a property as ValueGeneratedOnAdd since the default value is `NULL` which is already the default for nullable properties.

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
ajcvickers: How does it get done by convention? My understanding of Never is that it really means never. That is, no real or temporary value should be generated--it is the responsibility of the application to set any needed value before the entity is tracked.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlServerValueGenerationScenariosTest.cs
ajcvickers: But NotId is setup as a key and has ValueGeneratedOnAdd set, so why it no value generated?

ajcvickers: Never mind. I was looking at the wrong context class.

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
AndriySvyryd: No, we decided that value generators will run even if ValueGenerated is set to Never so that in case a CLR default is generated it is still sent to the database. Discriminators are configured by a convention.

AndriySvyryd: I'm fine delegating to @rowanmiller, but I think correct documentation is better than documentation that is less confusing for the normal use case. Also note that this is in remarks section, so it won't be visible in IntelliSense

ajcvickers: Why not make it more clear in the documentation that this is a corner case that most users will not hit?

AndriySvyryd: Ok, is this better?
> Note that values may still be generated by a client-side value generator, if one is set explicitly or by a convention.

src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerModelValidator.cs
ajcvickers: Can you point me to the test? It seems to me that a non-key property set to OnAdd that does not have default or computed value or SQL will get the strategy from the model, which would then trigger this even though the strategy has not been set on the property.

src/Microsoft.EntityFrameworkCore/Metadata/Builders/PropertyBuilder.cs
ajcvickers: I can live with that.

src/Microsoft.EntityFrameworkCore.SqlServer/Infrastructure/SqlServerModelValidator.cs
AndriySvyryd: Yes, that's what I said, it will throw even if it hasn't been set on the property.

ajcvickers: Sorry, then the question is why? Setting it on the model is shorthand for "anytime I need you to pick a strategy, then this is the one to use." It should never cause anything to throw.

ajcvickers: Also, I don't much care if setting it on a non key property does nothing, but I can see the value in that case of warning.

ajcvickers: When did we decide that? Using DatabaseGeneratedOption.Identity is the same as setting ValueGeneratedOnAdd, which means if the value is default or temporary then the update pipeline should expect the database to generate a value.

AndriySvyryd: I don't remember the exact date when we'd discussed this with @divega. But basically there were a couple of reports where users would expect this annotation to actually configure the property as Identity. I'm fine with making it a warning though.

ajcvickers: We need to discuss in a design meeting or similar because it was a fundamental outcome of the design we agreed upon that the data annotations do exactly the same thing as the fluent API. We debated at length the term "Identity" and the confusion it could cause but ultimately decided that continuing to have it mean "value generated on add" (which is what it has always meant since it was introduced--see GUIDs) was the best thing to do in the circumstances.

Now the SQL Server value generation strategy is different. In this case it really does mean use an identity column, or, more specifically, configure an entire end-to-end key generation strategy (including DDL) to use an identity column. We certainly agreed that if the application explicitly sets this strategy for a non-key column, then it is okay to throw because the strategy is only designed for keys. This is something we did agree on in a conversation with Diego. However, if the key strategy for the model is set to Identity, it doesn't mean that non-key properties that are value generated are somehow invalid. So if we are going to do this validation, then it needs to be done in a way that distinguishes the strategy being set on the property as opposed to the model.

AndriySvyryd: Yes, we can throw for explicit and warn for derived from model.

ajcvickers: Why warn? What are we warning about?

AndriySvyryd: That if a non-key property is configured as ValueGeneratedOnAdd that it will not use the model default or Identity or HiLoSequence and if the database is created by migrations it might be different from what one expects.

ajcvickers: I think this would be an unnecessary and annoying warning. With regard to Migrations, it would still be valid for Migrations to create an identity column if possible. If Migrations is unable to do this or something else that works, then I would expect an error from Migrations.

And with regards to customer's reporting this, I haven't seen anybody be confused by the value generation strategy. I've barely seen anyone who knows about it at all. So if I, quite correctly, make a property value generated on add, and then get a warning about it referring to something that pretty much nobody knows about it, how is that helpful?

We're not going to fix the long-standing confusion of DatabaseGeneratedOption.Identy by adding warnings about the SQL Server value generation strategy.

ajcvickers: If Migrations is unable to generate an appropriate column, then it is fine/correct for it to scaffold something with a warning or that throws such that the developer can then edit this as appropriate, for example, by adding a trigger.

ajcvickers: And besides, at model validation time, we have no knowledge that Migrations are even being used.

AndriySvyryd: There is no mechanism to inform migrations that a trigger will by added by a custom SQL command.
That's why it should be a warning and I don't think there's currently a way of showing warnings from migrations.

ajcvickers: So it seems this warning would be for people that:
* Are using Migrations
* Have a clue that we have an end-to-end strategy
* But don't understand that the strategy is only for key values
* And are not intending to edit their migration to generate appropriate DDL
* So will be suprised if Migrations does not generate an Identity column

AndriySvyryd: I'll change this to throw only for explicit configuration and I opened #7295 for the warning

PR #7321, opened 2016-12-28
Update test specification to netstandard1.7

pranavkm:
This ensures references from the netstandard1.7 TFM are resolved by Microsoft.EntityFrameworkCore.Design.FunctionalTests.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/TestUtilities/BuildReference.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/TestUtilities/BuildReference.cs
@@ -6,7 +6,7 @@ using System.Collections.Generic;
 using System.Reflection;
 using Microsoft.CodeAnalysis;
 
-#if NETSTANDARD1_6
+#if NETSTANDARD1_7
 using Microsoft.Extensions.DependencyModel;
 using System.Linq;
 #endif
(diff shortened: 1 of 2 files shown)


pranavkm:
Currently the references for the dynamically compiled references come from `Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests`'s deps file. With the update to the `4.4.0` packages, the functional tests that are added as references require the a System.Runtime that is higher than the one the `netstandard1.6` reference brings in. This results in compilation issues:

```
 error CS1705: Assembly 'Microsoft.EntityFrameworkCore.Design.FunctionalTests' with identity 'Microsoft.EntityFrameworkCore.Design.FunctionalTests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' uses 'System.Runtime, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Runtime' with identity 'System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```

One way to solve this is to bump up the specifications to `netstandard1.7`. The other alternative is to ensure that none of the references come from the [`DefaultDependencyContext`](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational.Design.Specification.Tests/TestUtilities/BuildReference.cs#L18-L19). This seemed like less work. The only concern is that it's a shipping package, so I'm not sure how much you care about this.

divega:
@pranavkm if this is blocking the build we can take this simple fix right now and keep an issue open to consider a different fix. Not sure @bricelam will be able to review it.

bricelam:
LGTM

PR #7323, opened 2016-12-29
DbConnection.Dispose no longer calls DbConnection.Close in CoreCLR

pranavkm:


test/Microsoft.EntityFrameworkCore.Relational.Tests/RelationalConnectionTest.cs
```diff
@@ -237,11 +237,7 @@ public void Lazy_connection_is_recreated_if_used_again_after_being_disposed()
 
             connection.Open();
 
-#if NET451
-            // On CoreCLR, DbConnection.Dispose() calls DbConnection.Close()
```
pranavkm: I think the behavior changed here - https://github.com/dotnet/corefx/commit/7888677b3c53596085c2fa53860d9a0519eeecfd#diff-d7bff8835e6985ab9107a59a282e1a60L153.

divega: It seems so. If this is blocking the build we can take your fix, but I am going to create a corefx bug because I am not sure this change was intentional.

PR #7327, opened 2016-12-29
Changed ForeignKeyIndexConvention to update the IsUnique property

### LINKED ISSUE #7254: ModelBuilding: Changing PK of principal type loses the user configured index info on FK property
When user is explicitly configuring the index which is backing FK, if the FK is regenerated then user configuration on the index is lost in the process.
In following example, the index is explicitly set as non-unique. The FK is recreated due to change of PK on principal type. The index becomes unique again.

### Steps to reproduce
Include a complete code listing (or project/solution) that we can run to reproduce the issue.

Partial code listings, or multiple fragments of code, will slow down our response or cause us to push the issue back to you to provide code to repoduce the issue.

```c#
{
    // Configure model
    modelBuilder.Entity<Account>(etb =>
    {
        etb.HasOne(p => p.Customer)
            .WithOne(p => p.Account)
            .HasForeignKey<Account>(p => p.CustomerId);

        etb.HasIndex(p => p.CustomerId).IsUnique(false);
    });
    modelBuilder.Entity<Customer>(etb =>
    {
        etb.HasKey(p => p.AuditId);
    });
}

public class Account
{
    public int? Id { get; set; }
    public Customer Customer { get; set; }
    public int? CustomerId { get; set; }
}

public class Customer
{
    public int? AuditId { get; set; }
    public int? Id { get; set; }
    public Account Account { get; set; }
}
```

```
var entityType = db.Model.FindEntityType(typeof(Account));
var property = entityType.FindProperty(nameof(Account.CustomerId));
Console.WriteLine(entityType.FindIndex(new List<IProperty> { property}).IsUnique); // outputs true
```

### Further technical details
EF Core version: 1.1.0
Database Provider: SqlServer
Operating system: Win10
IDE: VS 2015U3

tuespetre:
of the ForeignKey when the associated index's uniqueness changes.

Fixes #7254

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
+++ b/src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
@@ -923,9 +923,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
                 }
             }
 
-            builder = builder.Metadata.SetIsUnique(unique, configurationSource, runConventions)?.Builder;
-            builder?.Metadata.DeclaringEntityType.FindIndex(builder.Metadata.Properties)?.SetIsUnique(unique, configurationSource);
-            return builder;
+            return builder.Metadata.SetIsUnique(unique, configurationSource, runConventions)?.Builder;
         }
 
         private bool CanSetUnique(bool unique, ConfigurationSource? configurationSource, out bool resetToDependent)
@@ -1792,6 +1790,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
             var oldNavigationToDependentName = oldRelationshipInverted
                 ? Metadata.DependentToPrincipal?.Name
                 : Metadata.PrincipalToDependent?.Name;
+            var oldIsUnique = newRelationshipBuilder.Metadata.IsUnique;
 
             var newRelationshipConfigurationSource = Metadata.GetConfigurationSource();
             if ((dependentProperties != null && dependentProperties.Any())
@@ -2000,11 +1999,16 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
                         principalProperties);
                 }
 
-                if (newRelationshipBuilder != null
-                    && (newRelationshipBuilder.Metadata != initialRelationship
-                        || existingRelationshipInverted == null))
+                if (newRelationshipBuilder != null)
                 {
-                    newRelationshipBuilder = ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyAdded(newRelationshipBuilder);
+                    if (newRelationshipBuilder.Metadata != initialRelationship || existingRelationshipInverted == null)
+                    {
+                        newRelationshipBuilder = ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyAdded(newRelationshipBuilder);
+                    }
+                    else if (oldIsUnique != isUnique)
+                    {
+                        newRelationshipBuilder = ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyUniquenessChanged(newRelationshipBuilder);
+                    }
                 }
 
                 foreach (var addedForeignKey in addedForeignKeys)
--- a/test/Microsoft.EntityFrameworkCore.Tests/Metadata/Conventions/Internal/ForeignKeyIndexConventionTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.Tests/Metadata/Conventions/Internal/ForeignKeyIndexConventionTest.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Xunit;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    public class ForeignKeyIndexConventionTest
+    {
+        [Fact]
+        public void Does_not_override_foreign_key_index_uniqueness_when_referenced_key_changes()
+        {
+            var modelBuilder = new ModelBuilder(new CoreConventionSetBuilder().CreateConventionSet());
+            var principalTypeBuilder = modelBuilder.Entity<PrincipalEntity>();
+            var dependentTypeBuilder = modelBuilder.Entity<DependentEntity>();
+
+            principalTypeBuilder.HasKey(t => t.InitialPrincipalId);
+            dependentTypeBuilder.HasOne(t => t.Principal).WithOne().HasForeignKey<DependentEntity>(t => t.PrincipalId);
+            dependentTypeBuilder.HasIndex(t => t.PrincipalId).IsUnique(false);
+            principalTypeBuilder.HasKey(t => t.ChangedPrincipalId);
+
+            var entityType = modelBuilder.Model.FindEntityType(typeof(DependentEntity));
+            var property = entityType.FindProperty(nameof(DependentEntity.PrincipalId));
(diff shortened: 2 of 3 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ForeignKeyIndexConvention.cs
```diff
@@ -202,7 +202,7 @@ bool IIndexUniquenessConvention.Apply(InternalIndexBuilder indexBuilder)
                 foreach (var foreignKey in index.DeclaringEntityType.GetDerivedForeignKeysInclusive()
                     .Where(fk => fk.IsUnique && AreIndexedBy(fk.Properties, fk.IsUnique, index.Properties, existingIndexUniqueness: true)))
                 {
-                    CreateIndex(foreignKey.Properties, foreignKey.IsUnique, foreignKey.DeclaringEntityType.Builder);
+                    foreignKey.IsUnique = false;
```
ajcvickers: I don't think this is the correct change. The issue is about the uniqueness on the index being overwritten This instead changes the IsUnique flag on the FK.

tuespetre: That makes sense and I will update accordingly; out of curiosity, is there anything else 'linked' to the FK's IsUnique flag I might have missed?

ajcvickers: @tuespetre In EF Core metadata, a ForeignKey object defines a relationship, and the IsUnique flag determines whether that relationship is 1:1 or 1:*. EF therefore uses this flag to determine how to handle the relationship in queries, updates, fixup, etc.

Index objects define which indexes should be created in the database when using Migrations. They are currently not reasoned about by non-Migrations code, although this could change.

When a ForeignKey is created an Index will also be created with appropriate configuration for the FK. However, the application may change that Index in some way. This should have no effect on the ForeignKey and will only change how the index is created by Migrations.

tuespetre: @ajcvickers thanks! That'll get me thinking about some scenarios. Maybe there also needs to be a test to ensure that the very thing done by this current change does not happen (i.e. test to ensure that changing an index does NOT affect a foreign key.)

AndriySvyryd: The correct fix is to remove line 927 from InternalRelationshipBuilder.cs, leftover from before we had the convention:
```
builder?.Metadata.DeclaringEntityType.FindIndex(builder.Metadata.Properties)?.SetIsUnique(unique, configurationSource);
```
And to run `OnForeignKeyUniquenessChanged` conventions in `ReplaceForeignKey` if the FK uniqueness actually changed.

rowanmiller:
@tuespetre are you planning to address the feedback on this PR?

tuespetre:
Yes, sorry for the delay. I hope to get to this soon.

tuespetre:
@ajcvickers I have made changes and rebased. @AndriySvyryd thanks for pointing me in the right direction. By removing that call and then ensuring that the 'added' convention is run for a foreign key during replacement, everything seems to work out.

Note that I did have to change one existing test in order for it to pass -- when a one-to-one is changed to a one-to-many, it still makes sense that the dependents' foreign key stays unique by convention. The existing test had it being no longer unique.

AndriySvyryd:
@tuespetre The test was correct, if the FK uniqueness changes conventions need to update the index uniqueness. To achieve this you need to add code to run `ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyUniquenessChanged()` in `InternalRelationshipBuilder.ReplaceForeignKey`

tuespetre:
@AndriySvyryd oh geez, yeah I see it now. I got the dependent/principal thing flipped around in my brain holes for some reason. Thanks

tuespetre:
@ajcvickers @AndriySvyryd alright, third time's a charm?

AndriySvyryd:
@tuespetre You are straying in the wrong direction now. `ForeignKeyIndexConvention` doesn't need to be changed, it just needs to be run at the appropriate times.
More specifically currently `InternalRelationshipBuilder.ReplaceForeignKey` invokes `newRelationshipBuilder.IsUnique` with `runConventions: false` to avoid running the conventions prematurely before the relationship is fully configured, but it doesn't call `ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyUniquenessChanged()` afterwards. The latter is what needs to be added.

tuespetre:
@AndriySvyryd so instead of the change I made [here](https://github.com/aspnet/EntityFramework/pull/7327/files#diff-fafd831e48b8bfe93b1dd176ce8ea02bL2003), add an `else` block that runs the `ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyUniquenessChanged()`?

AndriySvyryd:
@tuespetre yes, though it would be an `else if` as it should only run if the uniqueness changed

tuespetre:
:up: :date: :hash: :four:

AndriySvyryd:
@tuespetre Looks good if the tests pass 😉

tuespetre:
Well, I guess Travis decided to bail 💀

AndriySvyryd:
@tuespetre StackOverflowException. Run the tests under a debugger to see what's happening.

src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -2000,11 +1999,16 @@ private bool CanInvert(ConfigurationSource? configurationSource)
...
-                    && (newRelationshipBuilder.Metadata != initialRelationship
-                        || existingRelationshipInverted == null))
+                if (newRelationshipBuilder != null)
                 {
-                    newRelationshipBuilder = ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyAdded(newRelationshipBuilder);
+                    if (newRelationshipBuilder.Metadata != initialRelationship || existingRelationshipInverted == null)
+                    {
+                        newRelationshipBuilder = ModelBuilder.Metadata.ConventionDispatcher.OnForeignKeyAdded(newRelationshipBuilder);
+                    }
+                    else if (oldIsUnique != isUnique)
```
AndriySvyryd: Perhaps the problem is here, try instead:
`oldIsUnique != newRelationshipBuilder.Metadata.IsUnique`

tuespetre:
Hmmm. I've been running TestDriven from Visual Studio to avoid the extra wait of the whole build/package process of `build.cmd`, but apparently it doesn't report this kind of error! Guess I need to take note of the total number of tests that ran/passed.

src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
tuespetre: Something is happening in one of the conventions, so the `newRelationshipBuilder` is assigned something new. I see it runs through the `ReuniquifyTempProperties` bit, and each time it comes through it changes to `QRefId1`, `QRefId2`, and so forth.

AndriySvyryd:
@tuespetre Yeah, with TestDriven.net you need to run the tests under the debugger with the option to break when StackOverflowException is thrown, otherwise it just swallows the exception.

src/Microsoft.EntityFrameworkCore/Metadata/Internal/InternalRelationshipBuilder.cs
AndriySvyryd: Oh, perhaps a more accurate test would be
```
Metadata.IsUnique != newRelationshipBuilder.Metadata.IsUnique
```

AndriySvyryd: Perhaps a more accurate test would be
```
oldIsUnique != newRelationshipBuilder.Metadata.IsUnique
```

AndriySvyryd:
@tuespetre I found and fixed the bug in  RelationshipDiscoveryConvention that was causing infinite recursion. Thanks for the contribution!

tuespetre:
Oh, nice! Thank you!

Issue #5806, opened 2016-06-17
Query: InMemory: Rewrite equality for reference types (byte[])


tillig:
Given a context that has objects using binary types...

``` csharp
public class SqlDataContext : DbContext
{
  public DbSet<ApiKey> ApiKeys { get; set; }
}

public class ApiKey
{
  public byte[] Hash { get; set; }
}
```

If you want to query using the in-memory provider, binary comparisons don't work as expected:

``` csharp
// Always returns null because it's comparing object references
// rather than using value equality.
this.DataContext.ApiKeys.FirstOrDefaultAsync(k => k.Hash == hash)
```

Based on [some StackOverflow research](http://stackoverflow.com/questions/23175762/comparing-byte-array-in-entity-framework-where-clause) I gather in SQL Server this should work.

That said, **is this the best way to search on binary data using EF?** If different providers may behave differently, I'm guessing an alternate mechanism may be needed. For example, maybe I should store the binary as base64 encoded strings and search on that?

ajcvickers:
Notes from #8376: Notes from triage: this bug is really about deep comparisons on arrays, etc. when comparing them inside the in-memory database. Decided that this should always be a deep comparison, done by re-writing the expression, and we should consider using IStructuralComparable.

smitpatel:
This should be easy to do in new pipeline.

smitpatel:
@ajcvickers - You modified code in this area when fixing #14042 would it also fix this?

ajcvickers:
@smitpatel Likely yes, but I'm not 100% sure.

smitpatel:
Put verify-fixed label.

Issue #5936, opened 2016-07-01
Query: SqlFunction to compare byte arrays (necessary to filter on rowversion)


jnm2:
Right now you can look at https://stackoverflow.com/questions/7437970/how-to-query-code-first-entities-based-on-rowversion-timestamp-value and see how difficult this is- it's next to impossible to get right.

Since we can't have comparison operators on byte arrays, the next most idiomatic thing would be to have the following methods in SqlFunctions (or DbFunctions, if possible), naming convention following the `Expression` class:
- `bool LessThan(byte[], byte[])`
- `bool LessThanOrEqual(byte[], byte[])`
- `bool GreaterThan(byte[], byte[])`
- `bool GreaterThanOrEqual(byte[], byte[])`
### Questions
1. I don't know if there is any value in `Equal` and `NotEqual`. They would be useful if evaluated client-side, because they would be semantically purer than the `==` operator which compares byte arrays by reference. The semantics of byte array `==` are incorrect because SQL compares binary by value (as it should).
2. Could they be extension methods? Might be a nice experience, might not.
#### Casting

A function that would be cool for other purposes would be `long SqlFunctions.Cast(byte[])`, but that would not work for this scenario because SQL Server only has signed comparisons, plus it's less performant server-side:

``` sql
select case when cast(0x0FFFFFFFFFFFFFFF as bigint) < cast(0xFFFFFFFFFFFFFFFF as bigint)
    then 'unsigned' else 'signed - oops' end
```

Doing a multi-step comparison would get around the signed comparison issue, but that's hacky and slower. Casting to `char(8)` would subject you to collation comparisons.
#### Idiomatic binary type

Another alternative would be to use a Binary primitive struct that wraps a byte array and provides all the value-comparison operators and has an implicit conversion to and from a byte array. Additionally, a Timestamp primitive type ([mine](https://gist.github.com/jnm2/929d194c87df8ad0438f6cab0139a0a6)) for rowversion columns. **This puts comparison operators back on the table and would be ideal over SqlFunctions**, but obviously will have to wait for #242. If it speeds up #242 all the better!

rowanmiller:
We want to discuss this with @divega when he is back in the office

divega:
Clearing up for triage. I think for EF Core we should initially consider adding support for translating existing idioms that express this kind of comparisons between byte arrays in .NET, e.g.:

[...]

PR #7328, opened 2016-12-29
InMemory: Rewrite equality for byte[]

tuespetre:
Addresses #5806


anpete:
Hi @tuespetre, thanks for the PR!

I wonder whether we should do this rewrite across the board. In particular, on relational when we end up not translating the comparison to the server (client-eval), we would still get a reference comparison. One idea I had was that we could rewrite binary equality comparisons to SequenceEqual at the QueryModel level. Then we would just need to teach the relational translator how to turn it back into SQL equality for byte arrays. What do you think?

divega:
Related discussion #5936 cc @anpete.

tuespetre:
@anpete that sounds good, especially coming back to this PR after working heavily with Relational the past month.

Plus I see I missed an `Expression.Not` in there 😱

tuespetre:
@anpete so we do this rewrite across the board, and then on relational, a method call translator could be used to handle translating it back to SQL equality. Sound about right?

tuespetre:
I'm going to back off of this one for now.

Issue #7243, opened 2016-12-14
MongoDB Provider


crhairr:
I've been working on a personal ASP.NET project that uses MongoDB. I recently migrated to ASP.NET Core and wanted to migrate my MongoDB linq adapter to EntityFramework Core so I could integrate better with the ASP.NET Identity providers. I noticed that there was no official support for MongoDB so I pulled the source and took a few weeks to write my own provider. My first draft was on top of the public EF Core NuGet package, and I'm currently in the process of migrating it into the EntityFramework Core solution to take advantage of certain internal features.

My provider is really a glorified adapter between EF Core and the official MongoDB C# Driver. I've gotten it building within Korebuild, and I'm still in the process of expanding unit tests for the project beyond rudimentary query/store functionality tests. So far, it works well with a few caveats:
- Complex type support (integral to document databases) is lacking; I've worked around the issue, but in a way that makes the StateManager not correctly walk the object graph to detect hierarchical changes. Because of this, complex type properties have to be updated in their entirety. This should (hopefully) go away when the Complex Type feature is complete (assuming this feature supports deeply nested and not just single-level complex type properties).
- I haven't yet separated out a Non-Relational/Document layer (similar to Microsoft.EntityFrameworkCore.Relational) to handle common DocumentDB concerns so that it's easier to create other providers (eg: Couchbase), but this is relatively trivial.
- In order to keep my code clean and manageable while finalizing it, I haven't yet peppered it with XML doc comments about internal APIs that aren't meant to be directly consumed by end-users. This is a trivial copy-n-paste operation.

I know that a MonogDB provider is a major feature on the EF Core roadmap Still, I'm very interested in submitting what I've got, but I wanted to make sure that I wouldn't be stepping on any toes before I send a pull request. Worst-case scenario, I can maintain my own fork and publish my provider to my local NuGet feed until an official one is released.

ErikEJ:
Why not create your own repo for it like most other third party EF Core providers?

crhairr:
Due to time constraints, I can't currently commit to maintaining a third-party provider. I have just enough time to maintain my own provider for my own needs, or to contribute to an existing one (hence offering to submit what I have working so far so as to create an existing provider).
If the EF Core team hasn't released an official MongoDB provider by the time I do get more free time, I'll certainly revisit the option.

rowanmiller:
[...]

PR #7330, opened 2016-12-30
Initial implementation of a MongoDB Provider.

crhairr:
- Wraps the official MongoDB C# driver
 - Builds and tests run from command line (functionality tests require MongoDB 3.2 to be installed).
 - TODO: does not have a Design library or full support for Migrations.
 - TODO: does not have a separate DocumentDB layer similar to the common Relational layer for SQL providers.
 - TODO: add XML comments (currently not present to keep the code clean and legible while building the provider).

Addresses #7243.


rowanmiller:
This isn't something that our team wants to take on at this stage. If there are any volunteers in the community then we'd certainly be happy to help them take over the project (we work pretty closely with a number of provider writers).

EdwardBlair:
Excellent work! Very interested in this, been looking for a MongoDB provider for a while.  Sorry to see this PR was rejected.

Would you be willing to pop this into a standalone project/repo? I'd do it myself, but I wouldn't want to impose.

ErikEJ:
Happy to help @EdwardBlair - getting this working would also enable access to DocumentDB (as it has a MongeDB surface now)

EdwardBlair:
In its current state it doesn't quite work with EF Core 1.1.1.

It relies on the new class `Microsoft.EntityFrameworkCore.Infrastructure.CoreModelValidator` which depends on an updated version of `ModelValidator` which isn't present in 1.1.1.  It also relies on an updated version of `ModelSource` to pass along the validator.

Fortunately the interfaces they implement haven't changed. I managed to work around this by just copying the related files from EFCore to the new project, putting them in a new namespace and referencing those. Appears to work fine.

Unsure what would be the best course, have custom implementations (i.e. copy them from EFCore) of IModelValidator and IModelSource until EFCore next is released, then remove these classes and reference ones from EFCore. Or modify MongoDBProvider so it works without any Frankenstein-ery.

N.B. I am a complete novice at EFCore or anything on this scale.

crhairr:
I originally wrote this as a standalone against EFCore 1.1 and then migrated it to 1.2 so I could take advantage of some goodies in the develop branch. I've been putting off migrating it back to standalone because EF Core 1.2 is supposed to be released soon, which will mean that the internals won't have to be re-written much.

I'll just have to figure out what to do about the shared project items like Check. Hopefully, I can use Git's submodule mechanism to point to the shared subdirectory in EF Core. That would mean even fewer changes, and would make it very easy for the EF Core team to merge it in if they ever did decide to take over the code.

roji:
@crhairr and others, as someone maintaining another external EF Core provider (Npgsql), I'd strongly advise simply copying the shared project items into your own repo rather than using git submodules. The shared source files really don't change all that often and there's really no reason to go into the complexities of submodules...

It also wouldn't make it harder or easier to merge back into the main EF Core repo if that's ever on the table - which it probably won't be...

ErikEJ:
@crhairr Agree 100% with @roji  here ...

crhairr:
@roji You're probably right on both counts.

@EdwardBlair I'll go ahead and migrate this back to standalone sometime this week and create a new repo for anyone who wants to contribute. I'll copy any missing references for now so that it won't be that big of a pain to rebuild against 1.2 whenever it gets released.

crhairr:
I didn't get around to this until Sunday, but I got it converted this back to a standalone provider: https://github.com/crhairr/EntityFrameworkCore.MongoDb

It references the 1.2.0-preview build of EFCore, so there wasn't much to change aside from the DI. It just took me a little longer than expected to work out the kinks with VS2017 and fix a few tests that were failing with the updated MongoDb driver.

I'll be deleting my fork of this repo.

Issue #6534, opened 2016-09-15
Error when using async queries involving result operators and projecting a collection (N+1 queries scenario)


serber:
### Steps to reproduce

I have method named `GetSeriesData`:

```
private IQueryable<SeriesData> GetSeriesData(IQueryable<Series> query, long? userId = null)
{
    DateTime date = DateTime.Today.AddMonths(1);

    IQueryable<SeriesData> seriesDataQuery = query.Select(x => new SeriesData
    {
        Series = x,
        Subscribed = userId.HasValue && x.Subscriptions.Any(y => y.UserId == userId),
        CurrentSeasonNumber = x.Episodes.Where(z => z.ReleaseDate.HasValue && z.ReleaseDate < date).Max(y => y.SeasonNumber),
        Channel = x.Channel,
        Country = x.Channel.Country,
        ReleaseGroups = x.Episodes.SelectMany(z => z.Releases)
                                    .Select(y => y.ReleaseGroup)
                                    .OrderBy(y => y.Name)
                                    .Distinct()
                                    .Select(r => new ReleaseGroupData
                                    {
                                        ReleaseGroup = r,
                                        Subscribed = userId.HasValue && x.Subscriptions.Any(y => y.UserId == userId && y.ReleaseGroupId == r.Id)
                                    }).ToList()
    });

    return seriesDataQuery;
}
```

I call `GetSeriesData` method:

```
public async Task<SeriesData> Get(IFindSeriesParameters parameters, long? userId)
{
    DbSet<Series> seriesRepository = _dataContext.Set<Series>();
    IQueryable<Series> seriesQuery;

    if (parameters.Id.HasValue)
    {
        seriesQuery = seriesRepository.Where(x => x.Id == parameters.Id);
    }
    else if (!string.IsNullOrEmpty(parameters.UniqueUrl))
    {
        seriesQuery = seriesRepository.Where(x => x.UniqueUrl == parameters.UniqueUrl);
    }
    else
    {
        seriesQuery = seriesRepository;
    }

    SeriesData seriesData = await GetSeriesData(seriesQuery, userId).SingleOrDefaultAsync();
    return seriesData;
}
```
### The issue

Calling `SingleOrDefaultAsync` throws exception

```
Exception message:
[...]

PR #7354, opened 2017-01-05
[1.1.1] Fix to #6534 - Problems with LINQ async methods

maumar:
Problem was that during result operator processing, when creating expression for GetResult involving InjectParameter call we were not propagating CancellationToken. Fix is to call a helper method that takes care of that, rather than Expression.Call.


maumar:
@anpete @divega

PR #7355, opened 2017-01-05
Refactoring NavigationRewritingExpressionVisitor

tuespetre:
This isn't really a bug fix or a new feature so to speak. I set out to begin working on existing issues around navigations in queries and needed to wrap my head around the code, but I found it difficult to keep track of what was going on and how the behaviors of different things interacted. After some tip-toeing around, stepping through calls, and breaking some tests, I started to see what some of the little 'flags' meant and when they applied, so I wanted to separated their concerns and make their behavior a little more explicit.

I would like to work on some of the open issues concerning navigations by building upon this refactored code. I made sure all of the tests passed locally, please let me know what you think.


tuespetre:
Just a heads up to anyone reading this, I plan to implement a solution for https://github.com/aspnet/EntityFramework/issues/3792 on top of this work; maybe that could help to explain the way I refactored the code here.

Issue #7227, opened 2016-12-10
EF Core 1.1 IsModified is not working


JulienM28:
I have an issue updating my datas on an entity.

Before Updating to EF Core 1.1.0 it worked fine.

### Steps to reproduce

This one worked before but no more now. The property is not updated
```
 using (ApplicationDbContext Ctx = new ApplicationDbContext())
                    {
                        Ctx.Entry(_dico[id]).Property(key).IsModified = true;
                        Ctx.SaveChanges();
                    }
```

Here the property is correctly updated in de database.
```
 using (ApplicationDbContext Ctx = new ApplicationDbContext())
                    {
                        Ctx.Update(_dico[id]);
                        Ctx.SaveChanges();
                    }
```
Thanks for your help.

EF Core version: 1.1.0
Database Provider: Microsoft.EntityFrameworkCore.SqlServer
Operating system: Windows 10 64 bits
IDE: Visual Studio 2015 Update 3

ajcvickers:
@JulienM28 Could you provide more complete steps to reproduce this issue. The code you posted causes the updated value of the property to be saved in the tests I have run. Could you create a simple console app that reproduces the issue?

JulienM28:
Here is the complete method.
It's part of an Asp.Net Core Website using .net 4.6 Framework for things like System.Drawing


`public static void Update(string type, int id, string key, string value)
        {

            if (!key.Contains('.'))
            {
                if (value == "on") value = "true";
                if (value == "off") value = "false";
                Type typeObjet = Type.GetType("BookStarter.Models." + type);
                if (typeObjet == null) typeObjet = Type.GetType("BookStarter.Models.Books." + type);
                if (typeObjet == null) return;
                PropertyInfo prop = typeObjet.GetProperty(key);
                Type t = Nullable.GetUnderlyingType(prop.PropertyType) ?? prop.PropertyType;
                object nullableValue = (string.IsNullOrEmpty(value) || value == null) ? null : Convert.ChangeType(value, t);

                PropertyInfo[] containers = typeof(MemoryCache).GetProperties();
                foreach (PropertyInfo dictionary in containers)
                {
                    Type[] arguments = dictionary.PropertyType.GetGenericArguments();
                    Type valueType = arguments[1];
                    if (valueType == typeObjet)
                    {
                        var dico = typeof(MemoryCache).GetProperty(dictionary.Name).GetValue(null);
                        IDictionary _dico = dico as IDictionary;
[...]

PR #7361, opened 2017-01-05
Fix handling of setting property to modified on detached Entity (Note: 1.1.1 fix)

ajcvickers:
Issue #7227

Ensures that the entity is attached with the property set to modified and that fixup does not inappropriately do more than it should (i.e. marking FK is modified).

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
@@ -295,8 +295,8 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
                 }
             }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -321,7 +322,8 @@ public virtual bool IsModified(IProperty property)
 
             // Don't change entity state if it is Added or Deleted
```
divega: Is the comment still useful?

ajcvickers: Probably not.

ajcvickers:
This is fixed and merged. Leaving bug open until it is approved.

Issue #7116, opened 2016-11-24
ThenInclude Method not found


kimx:
### The issue
After upgrade to 1.1.0. Run  Query to load navigation property then got  the error as below.

Code
```
   var all = db.Category.Include(c => c.CategoryPics).ThenInclude(p => p.Picture);
       var result = all.ToList();
```

Error
```
MissingMethodException: Method not found: 'Microsoft.EntityFrameworkCore.Query.IIncludableQueryable`2<!!0,!!2> 
Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.ThenInclude(Microsoft.EntityFrameworkCore.Query.IIncludableQueryable`2<!!0,System.Collections.Generic.ICollection`1<!!1>>, System.Linq.Expressions.Expression`1<System.Func`2<!!1,!!2>>)'.
```

### Further technical details

EF Core version: 1.1.0
Operating system: 
Visual Studio version: VS 2015 Update 3

Other details about my project setup:

kimx:
Sorry~~
Please close this issue.This issue is my library project are forgot upgrade and still stay on 1.0.0.

brockallen:
I just ran into this issue. 

I have a ASP.NET Core 1.0 host app that references a library (netstandard14). The library references EF Core 1.0. The library has no SqlServer dependencies (just uses a DbContext). The host puts the DbContext in DI and configures SqlServer as the provider.

I then update the host app to ASP.NET Core 1.1 (and EF Core 1.1 and SqlServer 1.1), but still rely upon the same library that depends on EF Core 1.0. I run my code and now get this exception:

```
MissingMethodException: Method not found: 'Microsoft.EntityFrameworkCore.Query.IIncludableQueryable`2<!!0,!!2> Microsoft.EntityFrameworkCore.EntityFrameworkQueryableExtensions.ThenInclude(Microsoft.EntityFrameworkCore.Query.IIncludableQueryable`2<!!0,System.Collections.Generic.ICollection`1<!!1>>, System.Linq.Expressions.Expression`1<System.Func`2<!!1,!!2>>)'.
IdentityServer4.EntityFramework.Stores.ResourceStore.GetAllResources()
IdentityServer4.Stores.IResourceStoreExtensions+<GetAllEnabledResourcesAsync>d__2.MoveNext()
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #7362, opened 2017-01-05
Add back ThenInclude for ICollection to fix binary compat break (Note: 1.1.1 fix)

ajcvickers:
Issue #7116

Method is obsolete and not an extension method since it only needs to be there for binary compat.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
@@ -2243,10 +2243,23 @@ namespace Microsoft.EntityFrameworkCore
                     : source);
         }
 
+        internal static readonly MethodInfo ThenIncludeAfterEnumerableMethodInfo
+            = typeof(EntityFrameworkQueryableExtensions)
+                .GetTypeInfo().GetDeclaredMethods(nameof(EntityFrameworkQueryableExtensions.ThenInclude))
+                .Single(mi =>
+                    {
+                        var typeArgument = mi.GetParameters()[0].ParameterType.GenericTypeArguments[1];
+                        return typeArgument.IsGenericType && typeArgument.GetGenericTypeDefinition() == typeof(IEnumerable<>);
+                    });
+
         internal static readonly MethodInfo ThenIncludeAfterCollectionMethodInfo
             = typeof(EntityFrameworkQueryableExtensions)
                 .GetTypeInfo().GetDeclaredMethods(nameof(EntityFrameworkQueryableExtensions.ThenInclude))
-                .Single(mi => !mi.GetParameters()[0].ParameterType.GenericTypeArguments[1].IsGenericParameter);
+                .Single(mi =>
+                    {
+                        var typeArgument = mi.GetParameters()[0].ParameterType.GenericTypeArguments[1];
+                        return typeArgument.IsGenericType && typeArgument.GetGenericTypeDefinition() == typeof(ICollection<>);
+                    });
 
         internal static readonly MethodInfo ThenIncludeAfterReferenceMethodInfo
             = typeof(EntityFrameworkQueryableExtensions)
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
```diff
@@ -2298,11 +2311,34 @@ public static class EntityFrameworkQueryableExtensions
...
+        /// <typeparam name="TPreviousProperty"> The type of the entity that was just included. </typeparam>
+        /// <typeparam name="TProperty"> The type of the related entity to be included. </typeparam>
+        /// <param name="source"> The source query. </param>
+        /// <param name="navigationPropertyPath">
+        ///     A lambda expression representing the navigation property to be included (<c>t => t.Property1</c>).
+        /// </param>
+        /// <returns>
+        ///     A new query with the related data included.
+        /// </returns>
+        [Obsolete("Use overload that takes an IEnumerable<> navigation property instead.")]
```
divega: I suspect that the majority of developers that will end up using this overload are going to be application developers that will get it picked up by the compiler based on their navigation property types as opposed to library developers that explicitly choose it.

On the other hand removing this overload in 2.0 means the enumerable version will be picked up by the compiler and the code is source compatible, so no actual break as long as you are compiling the code again (which will very likely be required in 2.0).

So in this particular case isn't the obsolete message potentially more confusing than helpful?

src/Microsoft.EntityFrameworkCore/Query/ResultOperators/Internal/ThenIncludeExpressionNode.cs
```diff
@@ -24,6 +24,7 @@ public class ThenIncludeExpressionNode : ResultOperatorExpressionNodeBase
         /// </summary>
         public static readonly IReadOnlyCollection<MethodInfo> SupportedMethods = new[]
         {
+            EntityFrameworkQueryableExtensions.ThenIncludeAfterEnumerableMethodInfo,
```
divega: Could that this is all that is needed. I assume we are not adding tests because when we recompile the tests a bunch of them will start using this version?

ajcvickers: I ran all the tests while it was still a non-obsolete extension method, and all was well.

src/Microsoft.EntityFrameworkCore/EntityFrameworkQueryableExtensions.cs
ajcvickers: Not an extension method, so no "normal" code will pick it up.

divega: Talked to @ajcvickers and it turns out this is just a static method as opposed to an extension method (cool trick that @bricelam came up with), so my concerns were not justified.

src/Microsoft.EntityFrameworkCore/Query/ResultOperators/Internal/ThenIncludeExpressionNode.cs
divega: This can be ignored. @ajcvickers tested that using the overload still works.

natemcmaster:
This introduces compiler errors

```
[17:31:49]	[Step 3/4] C:\BuildAgent\work\a409e8798ad316f7\.dotnet\dotnet.exe compile-csc @C:\BuildAgent\work\a409e8798ad316f7\EntityFramework\src\Microsoft.EntityFrameworkCore\obj\Release\netstandard1.3\dotnet-compile.rsp returned Exit Code 1
[17:31:49]	[Step 3/4] C:\BuildAgent\work\a409e8798ad316f7\EntityFramework\src\Microsoft.EntityFrameworkCore\EntityFrameworkQueryableExtensions.cs(2234,45): error CS1061: 'Type' does not contain a definition for 'IsGenericType' and no extension method 'IsGenericType' accepting a first argument of type 'Type' could be found (are you missing a using directive or an assembly reference?)
[17:31:49]	[Step 3/4] C:\BuildAgent\work\a409e8798ad316f7\EntityFramework\src\Microsoft.EntityFrameworkCore\EntityFrameworkQueryableExtensions.cs(2243,45): error CS1061: 'Type' does not contain a definition for 'IsGenericType' and no extension method 'IsGenericType' accepting a first argument of type 'Type' could be found (are you missing a using directive or an assembly reference?)
[17:31:49]	[Step 3/4] 
```

Issue #7348, opened 2017-01-04
Some result operators (e.g. Distinct) after selecting a property generate incorrect SQL if query also contains optional navigation or explicit GroupJoin call


kierenj:
Same model as #6937, #6938, #7102, #7230.

This code:

```c#
            var identities = _context.Set<WidgetManufacturedIdentity>();
            var result = identities.Where(i => i.Manufacturer.Id == id && !i.Manufacturer.Retired.HasValue)
                .Select(i => i.RangeName)
                .Distinct()
                .ToArray()
                .Where(r => !string.IsNullOrEmpty(r))
                .Select(r => new { name = r });
```

Should generate a distinct set of `RangeName` property values.  Instead, it retrieves all values.

Log output:

```
info: Microsoft.EntityFrameworkCore.Storage.IRelationalCommandBuilderFactory[1]
      Executed DbCommand (33ms) [Parameters=[@__id_0='?'], CommandType='Text', CommandTimeout='30']
      SELECT DISTINCT [i].[WidgetId], [i].[BrandName], [i].[ManufacturerId], [i].[ProductName], [i].[ProductRef], [i].[RangeName], [i.Manufacturer].[Id], [i.Manufacturer].[Added], [i.Manufacturer].[ListOrder], [i.Manufacturer].[Modified], [i.Manufacturer].[Name], [i.Manufacturer].[Retired]
      FROM [WidgetManufacturedIdentity] AS [i]
      LEFT JOIN [WidgetManufacturer] AS [i.Manufacturer] ON [i].[ManufacturerId] = [i.Manufacturer].[Id]
      WHERE ([i.Manufacturer].[Id] = @__id_0) AND [i.Manufacturer].[Retired] IS NULL
```

As you can see, the `DISTINCT` is there, but the argument list to the `SELECT` is not limited to the select list I specified. 

### Further technical details

EF Core version: `1.2.0-preview1-22878`
Database Provider: `Microsoft.EntityFrameworkCore.SqlServer`
Operating system: Windows 10.0.14393
IDE: VS2015, commandline, Azure App Service

maumar:
This happens because https://github.com/aspnet/EntityFramework/issues/6647 (left join on WidgetManufacturer)

However for this case we now produce incorrect results. Before the change, Distinct was done on the client, based on only one column and therefore produced the correct results. We should aim to fix this in 1.1.1 @divega

divega:
@maumar sounds good. Feel free to assign it.

[...]

PR #7367, opened 2017-01-06
[1.1.1] Fix to #7348 - `Distinct()` after `Select()`ing a property doesn't generate correct SQL if query also contains optional navigation

maumar:
Problem was that optional navigation is translated to GroupJoin which in turn forces materialization on entire entity. This causes Distinct to be applied on all columns of that entity, even if the customer specified a subset of columns to perform Distinct on.
Fix is to always perform Distinct on the client if the query also contains GroupJoin.


maumar:
@anpete @divega

anpete:
@maumar Is it not just RequiresClientResultOperator = true?

maumar:
@anpete new version up

divega:
@maumar we think your commit message is referencing the wrong bug so it will close the wrong bug.

maumar:
@divega good catch, fixed
@anpete new version up

Issue #7001, opened 2016-11-11
Improve Exception message: Unable to create or track an entity of type 'abc' because it has a null primary or alternate key value.


rpokrovskij:
I have found the difference between Unique Index and Unique constraint: the exception in case of NULL value. In case of Unique Index you have a sqlserver exception  with message that contains column name "Cannot insert the value NULL into column...", when in case of Unique Constraint we get InvalidOperationException with generic message: **Unable to create or track an entity of type 'TestParentRecord' because it has a null primary or alternate key value.**

I propose to add filed names to exception message as sql server do.

EF Core 1.1.0-preview1-final

PR #7374, opened 2017-01-07
Improve exception messages for invalid identity and identity conflicts

ajcvickers:
Issue #7001

Includes:
- Null key and conflicting key cases
- Property names
- Help for null alternate key case
- Conflicting key values included when sensitive data logging is enabled
- Updated messages that better reflect current behavior

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
+++ b/src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
@@ -133,7 +133,10 @@
     <value>The expression '{expression}' is not a valid property expression. The expression should represent a property access: 't =&gt; t.MyProperty'.</value>
   </data>
   <data name="IdentityConflict" xml:space="preserve">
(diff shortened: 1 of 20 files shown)


src/Microsoft.EntityFrameworkCore/Properties/CoreStrings.resx
```diff
@@ -133,7 +133,10 @@
...
   </data>
   <data name="IdentityConflict" xml:space="preserve">
-    <value>The instance of entity type '{entityType}' cannot be tracked because another instance of this type with the same key is already being tracked. When adding new entities, for most key types a unique temporary key value will be created if no key is set (i.e. if the key property is assigned the default value for its type). If you are explicitly setting key values for new entities, ensure they do not collide with existing entities or temporary values generated for other new entities. When attaching existing entities, ensure that only one entity instance with a given key value is attached to the context.</value>
+    <value>The instance of entity type '{entityType}' cannot be tracked because another instance with the same key value for '{keyProperties}' is already being tracked. When attaching existing entities, ensure that only one entity instance with a given key value is attached. Consider using 'DbContextOptionsBuilder.EnableSensitiveDataLogging' to see the conflicting key values.</value>
```
AndriySvyryd: 'entity instance' is redundant, just 'entity' or 'instance' would be enough

Issue #7332, opened 2016-12-31
EntityEntry.OriginalValues APIs incorrectly return current values


Dixin:
REcently I ported some EF code to EF Core, and noticed these behaviors. After changing a tracked entity, get its EntityEntry object, then:
1. EntityEntry.OriginalValues.Clone()
  In EF, it return a clone of original values. In EF Core, it returns a clone of current values.
  Workaround in EF Core:
```
var originalValues = entry.OriginalValues.Clone(); // Return current values.
originalValues.SetValues(entry.OriginalValues); // Now original values.
```
2. EntityEntry.OriginalValues.ToObject()
  In EF, it returns an entity with original values. In EF Core, it returns an entity with current values.
  Workaround in EF Core: 
```
var originalValues = entry.OriginalValues.Clone();
originalValues.SetValues(entry.OriginalValues);
Product original = (Product)originalValues.ToObject();
```

Are these bugs, or am I using it in a wrong way?

ajcvickers:
Note for triage: confirmed this is a bug.

tonyeung:
@Dixin  thanks for the workarounds. Makes spending two hours trying to figure out if I was going crazy not so bad.

divega:
@ajcvickers I assigned this to 1.1.1 because I am not sure whether it is a regression. Please fix on the appropriate branch.

ajcvickers:
@divega Not a regression. OriginalValues did not exist before 1.1. :-)

ajcvickers:
@divega @rowanmiller Moving this to 1.1.1 as discussed.

ajcvickers:
Justification: If a developer doesn't catch that the wrong values are returned, then this could result in the wrong data being saved when doing things like resolving concurrency exception.
Risk: Low--only needs a small change to add overrides in an existing derived type.

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

AndriySvyryd:
✅ Verified

JasonLeedy:
this does not appear to actually be fixed, exhibiting this behavior in 1.1.2

AndriySvyryd:
@JasonLeedy Could you share a repro project that shows the issue?

JasonLeedy:
Yes, take me a little bit to extract it from surrounding code

raygabe:
[...]

PR #7412, opened 2017-01-13
Fix OriginalValues.Clone to really use original values (1.1.1 fix)

ajcvickers:
Issue #7332

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
@@ -540,7 +540,9 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
 
             // If setting the original value results in the current value being different from the
             // original value, then mark the property as modified.
-            if (!IsModified(property))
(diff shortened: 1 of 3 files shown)


src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -540,7 +540,9 @@ public virtual void SetOriginalValue([NotNull] IPropertyBase propertyBase, [CanB
 
             // If setting the original value results in the current value being different from the
             // original value, then mark the property as modified.
-            if (!IsModified(property))
+            if ((EntityState == EntityState.Unchanged
+                 || EntityState == EntityState.Modified)
+                && !IsModified(property))
```
AndriySvyryd: Shouldn't this be like this?
```
(EntityState == EntityState.Unchanged  
             || (EntityState == EntityState.Modified 
                && !IsModified(property)))
```

ajcvickers: I guess that is slightly more efficient given that if the state is Unchanged, then IsModified will always be false, but it is functionally equivalent.

Issue #7298, opened 2016-12-22
Child entities not persisting when using AddRangeAsync and SaveChangesAsync


cdarrigo:
Attempting to add a graph of entities where there is a 1:m relationship (parent: children) using the AddRangeAsync and Savechanges Async method will prevent the children from being persisted to the DB.

In the code below, the a collection of StoreProductEntity (Parent) are being persisted in the DBContext . Each StoreProductEntity has an IList<StoreProductPriceEntity> (child).  After calling AddRangeAsync() and SaveChangesAsync(), only the parent entities are persisted, the children are not.

This is different behavior than the corresponding synchronous methods, AddRange() and SaveChanges(), where in the entire graph (parent and children) are persisted to the DB.

```c#
[Table("StoreProduct")]
    public class StoreProductEntity
    {
        public int Id { get; set; }
        public string Name { get; set; }
        
        public virtual IList<StoreProductPriceEntity> Prices { get; set; }

    }
    [Table("StoreProductPrice")]
    public class StoreProductPriceEntity
    {
        public int Id { get; set; }
        public decimal SalesPrice { get; set; }
        public virtual StoreProductEntity StoreProduct { get; set; }
        public int StoreProductId { get; set; }
    }

    public class Provider
    {
        private readonly DataContext dataContext;

        public Provider(DataContext dataContext)
        {
            this.dataContext = dataContext;
        }
        public async Task<int> AddAndSaveAsync(IEnumerable<StoreProductEntity> entities)
        {
            await this.dataContext.StoreProducts.AddRangeAsync(entities);
            var numAdded = await this.dataContext.SaveChangesAsync();
            // StoreProductEntities are saved, the children StoreProductPrice entities are not saving.
            return numAdded;
        }
    }

    public class Sample
    {
        private readonly Provider provider;

        public Sample(Provider provider)
        {
            this.provider = provider;
        }
        public async Task<int> SaveEntitiesSample()
        {
            var entities = new List<StoreProductEntity>
            {
                new StoreProductEntity
                {
                    Name = "A",
                    Prices = new List<StoreProductPriceEntity>
                    {
                        new StoreProductPriceEntity {SalesPrice = 1m}
                    }
                },
                new StoreProductEntity
                {
                    Name = "B",
                    Prices = new List<StoreProductPriceEntity>
                    {
                        new StoreProductPriceEntity {SalesPrice = 2m}
                    }
                }
            };

[...]

PR #7430, opened 2017-01-17
Make AddAsync and AddRangeAsync include related entities (1.1.1 fix)

ajcvickers:
Issue #7298

This requires quite a lot of new async code because all the graph traversal and processing code now needs async paths.


ajcvickers:
As discussed in triage, for 1.1.1 we will put a lower risk change that does not break binary compat on internal code and ship the real fix with 2.0 instead.

PR #7437, opened 2017-01-18
Make AddAsync and AddRangeAsync include related entities (1.1.1 fix)

ajcvickers:
Issue #7298

This is a lower-risk fix replacing #7430. The fix is to call the non-async methods. There is an extremely low probability that any real-world code is making database calls in the async methods combined with an even lower probability that making those calls synchronously would cause issues, so this fix is low-risk and does not have binary-compat issues.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/DbContext.cs
+++ b/src/Microsoft.EntityFrameworkCore/DbContext.cs
@@ -470,20 +470,11 @@ namespace Microsoft.EntityFrameworkCore
         ///     <see cref="EntityEntry{TEntity}" /> for the entity. The entry provides access to change tracking
         ///     information and operations for the entity.
         /// </returns>
-        public virtual async Task<EntityEntry<TEntity>> AddAsync<TEntity>(
+        public virtual Task<EntityEntry<TEntity>> AddAsync<TEntity>(
             [NotNull] TEntity entity,
             CancellationToken cancellationToken = default(CancellationToken))
             where TEntity : class
-        {
-            var entry = EntryWithoutDetectChanges(entity);
-
-            await entry.GetInfrastructure().SetEntityStateAsync(
-                EntityState.Added,
-                acceptChanges: true,
-                cancellationToken: cancellationToken);
-
-            return entry;
-        }
+            => Task.FromResult(Add(entity));
 
         /// <summary>
         ///     <para>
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore/DbContext.cs
```diff
@@ -470,20 +470,11 @@ private void SetEntityState(InternalEntityEntry entry, EntityState entityState)
...
-            var entry = EntryWithoutDetectChanges(entity);
-
-            await entry.GetInfrastructure().SetEntityStateAsync(
-                EntityState.Added,
-                acceptChanges: true,
-                cancellationToken: cancellationToken);
-
-            return entry;
-        }
+            => Task.FromResult(Add(entity));
```
AndriySvyryd: A good practice would be to check cancellationToken before making the call for all of these methods.

src/Microsoft.EntityFrameworkCore/DbContext.cs
```diff
@@ -870,19 +852,13 @@ public virtual void AddRange([NotNull] IEnumerable<object> entities)
...
+            AddRange(entities);
 
-            foreach (var entity in entities)
-            {
-                await stateManager.GetOrCreateEntry(entity).SetEntityStateAsync(
-                    EntityState.Added,
-                    acceptChanges: true,
-                    cancellationToken: cancellationToken);
-            }
+            return Task.FromResult(0);
```
AndriySvyryd: FromResult(true) is marginally faster.

src/Microsoft.EntityFrameworkCore/DbContext.cs
ajcvickers: I thought we didn't typically do that unless we were doing some I/O task or other long running operation. At least, that's what I remember us deciding.

src/Microsoft.EntityFrameworkCore/DbContext.cs
ajcvickers: I knew one was preferred over the other, but couldn't remember which, so I searched the code and found we use FromResult(0) more frequently than FromResult(true) or FromResult(false). We should probably switch them all over...

src/Microsoft.EntityFrameworkCore/DbContext.cs
AndriySvyryd: Since this could result in a database call I think it qualifies as I/O. That's why we added the async overloads in the first place, right?

src/Microsoft.EntityFrameworkCore/DbContext.cs
AndriySvyryd: Filed https://github.com/aspnet/EntityFramework/issues/7445

benaadams: Use a TaskCache? https://github.com/aspnet/Common/blob/dev/shared/Microsoft.Extensions.TaskCache.Sources/TaskCache.cs

src/Microsoft.EntityFrameworkCore/DbContext.cs
ajcvickers: In that case, shouldn't the check be made at the time that the database call is made?

AndriySvyryd: But we don't have the cancellationToken by that time because we are using the sync path

PR #7440, opened 2017-01-18
Changing a way we extract MethodInfo to be used for TrimStart and TrimEnd translators

maumar:
It was relying on only one method with that name being declated (i.e. no overloads), which will break once additional overloads are added.

Fix is to make method selection deterministic.


maumar:
@anpete

ErikEJ:
providers beware?

Issue #7457, opened 2017-01-23
Allow only one database provider per service provider


ajcvickers:
In EF Core 1.0 there can be multiple database providers registered in the same service provider (D.I. container). However, only one can be used with any given DbContext instance. This works by defining a set of provider services and having service resolution go through a scoped indirection based on the current context so that the correct service is returned for the provider currently being used.

This was important in pre-release versions of EF where there was no real distinction between the external application service provider and the internal service provider used by EF. This meant that there was no way to use ASP.NET Core in the normal way if the application made use of two different database providers. With the switch to EF managing its own service provider be default, it now means that using ASP.NET Core in the normal way will result in two internal service providers being created--one for each database provider.

Therefore, now that we have made the internal service provider the default experience, we can switch to a simpler model where only one database provider is allowed in the service provider. This means that:
- There is no indirection mechanism for provider services, which means that:
  - Some services that had to be scoped because they varied per provider can now be singleton
  - A provider can replace any service, not just a "provider service"
  - There is no interface/base class for providers, so providers must rely more on documentation to know which services to implement
  - There is no mechanism to inject services into providers in patch releases, so we will need a different mechanism for that
- Some caching will be duplicated for applications that do use multiple providers

PR #7464, opened 2017-01-24
Allow only one database provider per service provider

ajcvickers:
Issue #7457

Note that the way AddEntityFramework works has changed. Previously, the provider indirection mechanism ensured that any service overridden by a provider would be honored. Since this is gone, we use the following approach:
- TryAdd all provider services
- TryAdd all relational services (any services already registered by the provider are not overwritten)
- TryAdd all core services (any services registered by first two steps are not overwritten)
- Any services explicitly added by application will remain regardless

This means that an application should never call AddEntityFramework directly, since if this is done before the provider services are added then things will break. Therefore this method is renamed and no-longer an extension method. (Note that applications should not have been calling it anyway, but doing so was previously harmless.)

Also note that all services registered are now interfaces. This should have been the case before except for provider-specific implementations that were still resolved through their interfaces, but some concrete services had crept in.

Work still to do:
- Revisit the scopes of services
- Consider combining core convention set builder and model-specific builder into one
- Find a new solution for InjectAdditionalServices, since there is now no longer a hook for this
- Look at ResultOperatorHandler, since it differs from all other services in how it is used


anpete:
@ajcvickers Any particular thing to look at in here?

ajcvickers:
@anpete Hard to say. Probably AddEntityFrameworkInMemoryDatabase and related provider methods are a good starting point. Most of the changes are small and related to removing code that is no longer needed.

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/ServiceCollectionRelationalProviderInfrastructure.cs
```diff
@@ -0,0 +1,97 @@
...
+                    .AddScoped<IRelationalTypeMapper, RelationalTypeMapper>()
+                    .AddScoped<IRelationalValueBufferFactoryFactory, TypedRelationalValueBufferFactoryFactory>()
+                    .AddScoped<IDatabaseCreator>(p => p.GetService<IRelationalDatabaseCreator>())
+                    .AddScoped<IDbContextTransactionManager>(p => p.GetService<IRelationalConnection>())
+                    .AddQuery());
+
+            ServiceCollectionProviderInfrastructure.TryAddDefaultEntityFrameworkServices(serviceCollection);
+        }
+
+        private static IServiceCollection AddQuery(this IServiceCollection serviceCollection)
```
AndriySvyryd: Should this be inlined as well?

ajcvickers: @AndriySvyryd I guess its a matter of preference. I don't think it adds much value, but then I don't think it causes much harm.

anpete: Would be nice if we did for all of the logical groups. I.e. AddMigrations, AddUpdate etc. Otherwise probably just better to inline.

ajcvickers: There does tend to be quite a lot of overlap, and the more groups we have, the more overlap there would be. Query was originally inlined like the rest. I think it was split out at some point because it was hard to find services, or something like that. So I guess my slight preference would be to go back to a single list of services.

anpete:
−2,871 lines, nice.

anpete:
:shipit:

ErikEJ:
@ajcvickers Ugh! Possible to get a summary of provider affecting changes?

ajcvickers:
@ErikEJ (cc @roji) I think these are the main things:

- IDatabaseProviderServices is gone. Providers must register services explicitly. See AddEntityFrameworkSqlServer for example.
- IDatabaseProvider no longer has IDatabaseProviderServices and has the invarient name directly on it.
- Providers don't need their own IModelSource
- Providers may not need their own IValueGeneratorSelector
- IDbContextOptionsExtension.ApplyServices must return true if EF core services were added. (Providers should return true.)
- All IModelValidators should inherit from CoreModelValidator
- Providers can safely replace any service. However, please let us know if you find you need to replace an internal service or extend from an internal base class.

PR #7473, opened 2017-01-24
Fix #7251 - Possible race condition affecting DbContext pooling.

### LINKED ISSUE #7251: Possible race condition affecting DbContext pooling
The original PR to have ASP.NET Core benchmarks use explicitly compiled queries (https://github.com/aspnet/benchmarks/pull/157/commits/cd246047eba454b30d9132385927f0217f24f3e0) contained a bug in the use of `Attach()` in `LoadMultipleUpdatesRows()`: if the random number generator returns duplicate IDs in the same run then `Attach()` will fail.

@mikeharder pointed out that the expected probability of picking a random number twice or more from a range of 1 to 1001 in 20 attempts should be around 2% but we were seeing around 75% of requests fail. 

Switching off `DbContext` pooling (originally introduced to the benchmarks in https://github.com/aspnet/benchmarks/pull/154) brings the failure rate to expected values.

The current theory is that somehow the state of `DbContext` instances is not always correctly reset.

anpete:
- Fixed race condition between pool return and context state reset.
- Added missing dispose checks to DbContext. These are helpful for uncovering pooling issues because pooled instances are treated as disposed when not in use.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/DbContext.cs
+++ b/src/Microsoft.EntityFrameworkCore/DbContext.cs
@@ -121,7 +121,7 @@ namespace Microsoft.EntityFrameworkCore
             }
         }
 
-        private void CheckDisposed()
+        internal void CheckDisposed()
         {
             if (_disposed)
             {
@@ -168,7 +168,7 @@ namespace Microsoft.EntityFrameworkCore
                 contextServices.Initialize(scopedServiceProvider, options, this);
 
                 _logger = scopedServiceProvider.GetRequiredService<ILogger<DbContext>>();
-
+                
                 return contextServices;
             }
             finally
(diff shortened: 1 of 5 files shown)


src/Microsoft.EntityFrameworkCore/Internal/InternalDbSet.cs
```diff
@@ -42,10 +42,17 @@ public InternalDbSet([NotNull] DbContext context)
...
-            => NonCapturingLazyInitializer.EnsureInitialized(
-                ref _entityQueryable, 
-                this, 
-                internalSet => internalSet.CreateEntityQueryable());
+        {
+            get
+            {
+                _context.CheckDisposed();
+
+                return NonCapturingLazyInitializer.EnsureInitialized(
```
ajcvickers: Looking at this now I don't think it needs thread-safe initialization. DbSets are not shared and not thread safe. But I can make this change as a separate pull request.

src/Microsoft.EntityFrameworkCore/DbContext.cs
```diff
@@ -121,7 +121,7 @@ private IServiceProvider InternalServiceProvider
             }
         }
 
-        private void CheckDisposed()
+        internal void CheckDisposed()
```
ajcvickers: It's unfortunate that this needs to be internal. Is there any other way it could be done? Explicit interface implementation, perhaps?

PR #7477, opened 2017-01-25
[1.1.1] Uniquify FK and index names if there's another FK or index with the same name configured explicitly.

### LINKED ISSUE #7082: .net core 1.1 indexes
Having trouble upgrading to .net core 1.1. following the guide
created a brand new console app, as per the guide, created as 1.0.1
ran nuget package manager to update to 1.1
generated context from database first
created basic entity framework get all on db context on random table
I get error;

The indexes {'LocationId', 'PromotionId'} on 'DrpromotionLocation' and {'PromotionId'} on 'DrpromotionLocation' are both mapped to 'DRPromotionLocation.IX_DRPromotionLocation_PromotionId' but with different columns ({'LocationId', 'PromotionID'} and {'PromotionID'}).

This is not a one off, it seems to be on any table with an index
its been generated from database first, so nothing wrong with the database
This has all worked so far on .net core 1.0.1 and previous version of .net
any help appreciated

AndriySvyryd:
This will still result in a validation error if the conflicting name is configured by convention as in this case the uniquification would be dependent on the convention order.

Fixes #7082

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Metadata/RelationalForeignKeyAnnotations.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Metadata/RelationalForeignKeyAnnotations.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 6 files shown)


ajcvickers:
@AndriySvyryd Can you explain what is happening in this bug? What is it that reverse engineering is doing that is then causing, at some point, duplicate indexes? Or what is it that is causing conventions to build an index with the same name as one from reverse engineering but is a different index? Or what...

AndriySvyryd:
@ajcvickers RevEng is doing what is supposed to: it generates an index and names it `"IX_ApplicationKey_ApplicationId"`. By convention we create an index for the FK on `ApplicationKey.ApplicationId` and also name it `"IX_ApplicationKey_ApplicationId"`.

PR #7483, opened 2017-01-25
#7479 Trace connection open and close in RelationalConnection

### LINKED ISSUE #7479: Trace connection open and close in RelationalConnection
RelationalCommand has tracing but RelationalConnection connection is missing.
Adding ConnectionOpened and ConnectionClosed will help us trace this events in order to integrate with EntityFrameworkProfiler.
A PR is on the way.

fitzchak:
Fixes #7479


anpete:
cc @avanderhoorn

avanderhoorn:
This looks great! I'm wondering if we can include some of the same info we do with the commands (i.e. `InstanceId`, `Timestamp` & `Duration`)? Also on the command we have an exception version of the close message, do we need that here for connection? 

Lastly is there anyway to deterministically way which commands where a part of which connection? Like I know that the commands have a `connection` property, but if I was trying to save a `connectionId` as part of the data that I collect command message, how would you recommend that we do that and who is responsible for tracking that. Specifically, I could try and derive that Id from `GetHash` on the connection object, but I'm not sure if that would be what we would want people to be doing or if there is any gotchas in using that value for that sort of correlation and we should provide a different id. Thoughts?

fitzchak:
I think that adding a ConnectionId (guid) to RelationalConnection would be great addition. Than it will be needed to be passed to the other methods like the commands and also I want to add TransactionCommitted and TransactionRolledback which I added in a new commit.

fitzchak:
@avanderhoorn Added connection identifier. This is really great addition!

fitzchak:
FYI - I amended the "#7479 Trace connection open and close in RelationalConnection" commit a few times until #07dce52.

fitzchak:
If anything else I'll add is the latest commit which handle errors. This is one unified way to handle all exceptions.

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
```diff
@@ -266,14 +268,7 @@ object IRelationalCommand.ExecuteScalar(IRelationalConnection connection, IReadO
...
 
-                DiagnosticSource.WriteCommandError(
-                    dbCommand,
-                    executeMethod,
-                    instanceId,
-                    startTimestamp,
-                    currentTimestamp,
-                    exception,
-                    async: false);
+                DiagnosticSource.WriteError(connection.DbConnection, connection.ConnectionId, exception);
```
avanderhoorn: Are we saying that we are removing the semantic of `WriteCommandError` and that from now on all exceptions are at the "connection" level? Note, there is code out there that is already looking for command level exceptions is tapping the `WriteCommandError` event. I'm not sure we want to simply remove this.

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalDiagnostics.cs
```diff
@@ -65,38 +78,100 @@ internal static class RelationalDiagnostics
...
+        {
+            if (diagnosticSource.IsEnabled(TransactionDisposed))
+            {
+                diagnosticSource.Write(TransactionDisposed, new
+                {
+                    Transaction = transaction,
+                    ConnectionId = connectionId
+                });
             }
         }
```
avanderhoorn: Love having these transaction events!!! Good Addition.

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
avanderhoorn: Note, I'm not saying it can't be done... but the reason its like this is because we wanted to try an ensure that a "closing" event is "always" registered. Obviously downstream systems have to assume that it mightn't be if the error was bad enough, but with this, we could still alert people to how long the command took to error, etc.

avanderhoorn:
Also did you have a chance to look at (note this could apply to the transaction as well as the connection):
> This looks great! I'm wondering if we can include some of the same info we do with the commands (i.e. InstanceId, Timestamp & Duration)?

src/Microsoft.EntityFrameworkCore.Relational/Storage/Internal/RelationalCommand.cs
fitzchak: OK, I'll bring this back.
But my thought was that the executeMethod, instanceId, Timestamp and Duration isn't needed here, as most users would want to investigate the error itself at this point.

fitzchak:
Hi, I included InstanceId, Timestamp & Duration for connection.
InstanceId is used to corolate between ConnectionOpening to ConnectionOpened or ConnectionError.
As as now for transaction I didn't add BeforeTransactionCommit, but just TransactionCommited so there is not need for InstanceId there.

fitzchak:
I fixed conflicts and rebased.

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalDiagnostics.cs
```diff
@@ -92,11 +110,218 @@ internal static class RelationalDiagnostics
...
+        }
+
+        public static void WriteConnectionClosing(this DiagnosticSource diagnosticSource,
+            DbConnection connection,
+            Guid connectionId,
+            Guid instanceId,
+            long startTimestamp,
+            bool async)
+        {
+            if (diagnosticSource.IsEnabled(WriteConnectionClosing))
```
anpete: Getting build error here: error CS1503: Argument 1: cannot convert from 'method group' to 'string'

anpete: Looks like a typo, can you address?

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalConnection.cs
```diff
@@ -74,6 +75,8 @@ protected RelationalConnection([NotNull] RelationalConnectionDependencies depend
             }
         }
 
+        public Guid ConnectionId { get; } = Guid.NewGuid();
```
anpete: virtual, doc-comments

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalConnection.cs
```diff
@@ -91,6 +94,11 @@ protected RelationalConnection([NotNull] RelationalConnectionDependencies depend
         protected virtual ILogger Logger => Dependencies.Logger;
 
         /// <summary>
+        ///     Gets the diagnostic source.
+        /// </summary>
+        protected virtual DiagnosticSource DiagnosticSource=> Dependencies.DiagnosticSource;
```
anpete: nit: formatting

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalDiagnostics.cs
```diff
@@ -92,11 +110,218 @@ internal static class RelationalDiagnostics
...
             if (diagnosticSource.IsEnabled(DataReaderDisposing))
             {
-                diagnosticSource.Write(DataReaderDisposing, dataReader);
+                diagnosticSource.Write(DataReaderDisposing, new
+                {
+                    DataReader = dataReader,
+                    ConnectionId = connectionId
+                });
             }
         }
```
avanderhoorn: We are tracking `DataReaderDisposing` but what about its opening? i.e. tracking now long the read takes. Also it would be nice to know how many records were affected.

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalDiagnostics.cs
fitzchak: Sure

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalConnection.cs
fitzchak: Done

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalConnection.cs
fitzchak: Fixed.

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalDiagnostics.cs
fitzchak: Done.
Also added the results of the commands, which is most valuable in case of ExecuteNonQuery.

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalDiagnostics.cs
```diff
@@ -42,8 +56,10 @@ internal static class RelationalDiagnostics
 
         public static void WriteCommandAfter(
```
avanderhoorn: Just wondering if we need to add `transactionId` here? i.e. if we are wanting to correlate the command not only to the connection it was a part of but also to a transaction, with this current set of data, I'm wondering how we do that?

avanderhoorn:
@anpete just wondering about that last question I had (https://github.com/aspnet/EntityFramework/pull/7483#discussion_r98745178)? Any reason for the early merge?

src/Microsoft.EntityFrameworkCore.Relational/Internal/RelationalDiagnostics.cs
fitzchak: This is another issue should be in another pull request.
I'll take a look.

fitzchak: https://github.com/aspnet/EntityFramework/pull/7522

anpete:
@avanderhoorn We are making a push to get our PR count down and I thought this one was good to go. Sorry if I jumped the gun 😄 Happy to address remaining concerns on the new PR.

avanderhoorn:
No worries :) just making sure that I hadn't missed anything.

Issue #7100, opened 2016-11-23
ExecuteSqlCommand/Async usability issues


anpete:
1) When passing a strongly typed array for the parameters arg: E.g.
```c#
context.Database.ExecuteSqlCommand(
    "SELECT * FROM Orders WHERE OrderID = {0} OR CustomerID = {1}", 
    new [] { 1, 2 }));
```
We crash because the parameters object[] ends up with a single int[] element.

2) The Async overloads require passing a cancellation token when specifying parameters.

Test coverage of these methods is low. We should beef it up.

rowanmiller:
We should look at adopting the pattern we use elsewhere for async methods (not combining `params` and `CancellationToken`). We should also look at an overload that takes an IEnumerable of parameters rather than `params`.

PR #7484, opened 2017-01-25
Change signatures of ExecuteSqlCommand/Async to be more natural to use

ajcvickers:
Issue #7100

Allow:
- Async without explicit cancellation token
- Strongly typed arrays/lists
- Not accidentally passing cancellation token as parameter

There is no overload taking `IEnumerable<TParam>` because this results in a single string parameter being passed as `IEnumerbale<char>`. A similar situation still exists for parameters that implement IList of something, but this seems much less likely. Another option would be to add an explicit overload that takes string and revert to IEnumerable, but this would only cover the string case and not any other parameter type that implements IEnumerable.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
@@ -107,8 +107,17 @@ namespace Microsoft.EntityFrameworkCore
             [NotNull] this DatabaseFacade databaseFacade,
             [NotNull] string sql,
             [NotNull] params object[] parameters)
+            => ExecuteSqlCommand(databaseFacade, sql, (IEnumerable<object>)parameters);
+
+        // Note that this method doesn't start a transaction hence it doesn't use ExecutionStrategy
+        public static int ExecuteSqlCommand(
+            [NotNull] this DatabaseFacade databaseFacade,
+            [NotNull] string sql,
+            [NotNull] IEnumerable<object> parameters)
         {
             Check.NotNull(databaseFacade, nameof(databaseFacade));
+            Check.NotNull(sql, nameof(sql));
+            Check.NotNull(parameters, nameof(parameters));
 
             var concurrencyDetector = databaseFacade.GetService<IConcurrencyDetector>();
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -106,48 +106,73 @@ public static IEnumerable<string> GetPendingMigrations([NotNull] this DatabaseFa
             [NotNull] this DatabaseFacade databaseFacade,
             [NotNull] string sql,
             [NotNull] params object[] parameters)
+            => ExecuteSqlCommand(databaseFacade, sql, (IList<object>)parameters);
+
+        // Note that this method doesn't start a transaction hence it doesn't use ExecutionStrategy
+        public static int ExecuteSqlCommand<TParam>(
+            [NotNull] this DatabaseFacade databaseFacade,
+            [NotNull] string sql,
+            [NotNull] IList<TParam> parameters)
```
anpete: Could we do ICollection?

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -106,48 +106,73 @@ public static IEnumerable<string> GetPendingMigrations([NotNull] this DatabaseFa
...
+            Check.NotNull(parameters, nameof(parameters));
 
             var concurrencyDetector = databaseFacade.GetService<IConcurrencyDetector>();
 
             using (concurrencyDetector.EnterCriticalSection())
             {
                 var rawSqlCommand = databaseFacade
                     .GetService<IRawSqlCommandBuilder>()
-                    .Build(sql, parameters);
+                    .Build(sql, parameters.Cast<object>().ToList());
```
anpete: Any way to avoid this alloc?

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
ajcvickers: Yes

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
ajcvickers: Possibly, with more code, if it is critical.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -106,48 +106,73 @@ public static IEnumerable<string> GetPendingMigrations([NotNull] this DatabaseFa
...
             }
         }
 
         // Note that this method doesn't start a transaction hence it doesn't use ExecutionStrategy
-        public static async Task<int> ExecuteSqlCommandAsync(
+        public static Task<int> ExecuteSqlCommandAsync(
+            [NotNull] this DatabaseFacade databaseFacade,
+            [NotNull] string sql,
+            CancellationToken cancellationToken = default(CancellationToken))
+            => ExecuteSqlCommandAsync(databaseFacade, sql, new object[0], cancellationToken);
```
anpete: Consider caching the empty array here and unifying with the one in ObservableHashSet. It looks like .NET 4.6 introduced Array.Empty<T>() to handle this.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
anpete: Certainly not critical. It might be good to have a team discussion on our approach to these kind of optimizations. How far do we go etc.

anpete:
:shipit:

ajcvickers:
@anpete @divega New version up that just uses object, as discussed.

src/Microsoft.EntityFrameworkCore.Relational/RelationalDatabaseFacadeExtensions.cs
```diff
@@ -122,33 +131,49 @@ public static IEnumerable<string> GetPendingMigrations([NotNull] this DatabaseFa
...
+            Check.NotNull(parameters, nameof(parameters));
 
             var concurrencyDetector = databaseFacade.GetService<IConcurrencyDetector>();
 
             using (concurrencyDetector.EnterCriticalSection())
             {
                 var rawSqlCommand = databaseFacade
                     .GetRelationalService<IRawSqlCommandBuilder>()
-                    .Build(sql, parameters);
+                    .Build(sql, parameters.Cast<object>().ToList());
```
anpete: I guess you meant to remove `.Cast<object>().ToList()`?

ajcvickers: Yes, missed that.

Issue #7119, opened 2016-11-24
System.AggregateException occurred.  InvalidOperationException: Collection was modified; enumeration operation may not execute.


mdmoura:
### Steps to reproduce

I have the following Entities in Entity Framework Core 1.1:

``` C#
    public class Question {
      public Int32 Id { get; set; }
      public Int32 AuthorId { get; set; }
      public String Content { get; set; }
      public virtual User Author { get; set; }
      public virtual ICollection<Answer> Answers { get; set; }
    } 

    public class Answer {
      public Int32 Id { get; set; }
      public Int32 QuestionId { get; set; }
      public Int32 AuthorId { get; set; }
      public String Content { get; set; }  
      public virtual Question Question { get; set; }
      public virtual User Author { get; set; }
    }

    public class User : IdentityUser<Int32> {
      public virtual ICollection<Answer> Answers { get; set; } = new List<Answer>();
      public virtual ICollection<Question> Questions { get; set; } = new List<Question>();
    }

    public class Context : IdentityDbContext<User, Role, Int32> {
      public DbSet<Answer> Answers { get; set; }
      public DbSet<Question> Questions { get; set; }

      public Context(DbContextOptions<Context> options) : base(options) { }

      protected override void OnModelCreating(ModelBuilder builder) {
        base.OnModelCreating(builder);

        builder.Entity<Question>(b =>
          b.ToTable("Questions");
          b.HasKey(x => x.Id);
          b.Property(x => x.Content).HasMaxLength(400).IsRequired(true);
          b.HasOne(x => x.Author).WithMany(x => x.Questions).HasForeignKey(x => x.AuthorId).IsRequired(true).OnDelete(DeleteBehavior.Restrict);
        });

        builder.Entity<Answer>(b =>
          b.ToTable("Answers");
          b.HasKey(x => x.Id);
          b.Property(x => x.Content).HasMaxLength(4000).IsRequired(true);
          b.HasOne(x => x.Author).WithMany(x => x.Answers).HasForeignKey(x => x.AuthorId).IsRequired(true).OnDelete(DeleteBehavior.Restrict);
          b.HasOne(x => x.Question).WithMany(x => x.Answers).HasForeignKey(x => x.AuthorId).IsRequired(true).OnDelete(DeleteBehavior.Cascade);        
        });
      }
    }
```

### The issue

I tried to create a list of questions (in the following code I am adding only one):

``` C#
        List<Question> questions = new List<Question>();

[...]

PR #7486, opened 2017-01-25
Defensively copy navigation properties during graph attach

ajcvickers:
Issue #7119

The fundamental problem here is that Answer.AuthorId is being used as the FK for two different relationships. The navigation properties for the relationships effectively pointed to different objects during fixup, which caused the FK to change value, resulting in an inconsistent graph. This kind of overlapping FKs will only work if great care is taken with the entities assigned and FK values given, which was not the case in the repro code.

All that being said, the framework should be resilient to modification of navigations while attaching, so defensive copy code has been added to ensure this.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/EntityEntryGraphIterator.cs
+++ b/src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/EntityEntryGraphIterator.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore.Infrastructure;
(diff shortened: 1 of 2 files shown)


src/Microsoft.EntityFrameworkCore/ChangeTracking/Internal/EntityEntryGraphIterator.cs
```diff
@@ -38,7 +39,7 @@ public virtual void TraverseGraph(EntityEntryGraphNode node, Func<EntityEntryGra
                 {
                     if (navigation.IsCollection())
                     {
-                        foreach (var relatedEntity in (IEnumerable)navigationValue)
+                        foreach (var relatedEntity in ((IEnumerable)navigationValue).OfType<object>().ToList())
```
anpete: Use Cast<object> to avoid an `is` check for each element?

ajcvickers: Will do

Issue #7465, opened 2017-01-24
Add paramter objects and service rewriting code to avoid breaking base class D.I. constructors in patch/point releases


ajcvickers:
When a service is overridden by a provider, the provider must call the base service constructor. This means it becomes a breaking change when base service gets new dependencies such that its constructor needs to change.

We previously implemented a mechanism whereby additional services could be injected as part of the provider service resolution mechanism. This worked because provider services were always registered through a delegate, and so we could add code to that delegate as needed to inject services. Following #7457 services are registered by the provider explicitly in the container, which is much preferable for perf, but means that there is no hook to inject services.

jnm2:
I'm super curious what you come up with. I've hit this in my own unrelated architectures.
I've been thinking about the fact that the BCL best practice for adding parameters to event handlers in a back-compatible way is by defining an `EventArgs` from the start.
Compared to straight up constructor injection it feels like quite some overhead to write a `DependencyArgs` for each class, yet we've done it for every event for years so maybe not as bad is it first seems. The `DependencyArgs` subclass could itself be recognized and constructor injected by the IoC container and it could provide getter-only autoproperties which could be added to at any time.

popcatalin81:
How about just passing a single IServiceCollection parameter to all base services, and to allow customization of per provider create a wrapper IServiceCollection for each provider.

ajcvickers:
@jnm2 @popcatalin81 Passing an IServiceProvider is probably what we will end up doing, but also what we hope to avoid doing because it results in use of the service locator pattern, which then obscures dependencies. I suspect we will use this pattern along with adding [Obsolete] to the existing constructor while at the same time introducing a new constructor with the new dependency added explicitly. Then at next major version rev we will remove the obsolete constructor.

ajcvickers:
Discussed in design meeting and decided to try the parameter object approach for services, using sealed classes. Considered composition, but this would be a big change to the architecture and design of certain services.

PR #7494, opened 2017-01-26
Service dependencies parameter class for RelationalConnection

ajcvickers:
This is a proof-of-concept implementation of parameter objects for service dependencies. Issue #7465.

Notes:
- Parameter class is sealed.
- Ended up going with Clone on the parameters object--cleaner than other things I tried, and general purpose. We can add overloads of Clone as needed when doing point releases without breaking existing code. (May want to consider whether parameter should be options.)
- Registration is not try-add and is of concrete class

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalConnectionDependencies.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalConnectionDependencies.cs
@@ -1,0 +1,74 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Utilities;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.EntityFrameworkCore.Storage
+{
+    /// <summary>
+    ///     <para>
+    ///         Service dependencies parameter class for <see cref="RelationalConnection" />
+    ///     </para>
+    ///     <para>
+    ///         This type is typically used by database providers (and other extensions). It is generally
+    ///         not used in application code.
+    ///     </para>
+    /// </summary>
+    public sealed class RelationalConnectionDependencies
+    {
+        /// <summary>
+        ///     <para>
+        ///         Creates the service dependencies parameter object for a <see cref="RelationalConnection" />.
+        ///     </para>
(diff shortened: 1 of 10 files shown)


ajcvickers:
cc @anpete @divega @AndriySvyryd

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalConnectionDependencies.cs
```diff
@@ -0,0 +1,71 @@
...
+        ///     A replacement for the current dependency of this type, or null to continue to use the current dependency.
+        /// </param>
+        /// <param name="logger">
+        ///     A replacement for the current dependency of this type, or null to continue to use the current dependency.
+        /// </param>
+        /// <returns></returns>
+        public RelationalConnectionDependencies Clone(
+            [CanBeNull] IDbContextOptions contextOptions = null,
+            [CanBeNull] ILogger<IRelationalConnection> logger = null)
+            => new RelationalConnectionDependencies(
```
AndriySvyryd: Consider using a more granular approach: `WithOptions`, `WithLogger`, etc.

This way we don't need to obsolete the overloads as we add more dependencies.

anpete: Would we strictly need to obsolete the overloads?

AndriySvyryd: If we don't obsolete them then we'll end up with redundant methods. We could avoid that altogether.

ajcvickers: @anpete @AndriySvyryd The "With" thing is nice--we've done it before--but it can result in lost of code--as it did before! It somewhat depends how often we think this will happen.

anpete: @AndriySvyryd That depends on whether we use optional args.

ajcvickers: Even if we use optional args, we will probably need to add new methods to avoid binary breaks. I think I'm leaning towards With methods. We shouldn't have inheritance hierarchies here, like we did with DbInterceptionContext, so the number of methods should be more manageable.

ajcvickers:
@anpete @AndriySvyryd @divega Pushed a version using "With..." methods.

ajcvickers:
Merged

Issue #7476, opened 2017-01-25
Query : client-side result operators may cause extensive client-side evaluation for queries with GroupJoins (and/or optional navigations) in a subquery


maumar:
query:

```
from o in context.Orders
where (
   from od in context.OrderDetails
   where o.Customer.Country == od.Order.Customer.Country
   select od).Distinct().Count() > 0
select o;
```

produces the following query plan:
```
TRACKED: True
(QueryContext queryContext) => IEnumerable<Order> _Select(
    source: IEnumerable<TransparentIdentifier<Order, Customer>> _Where(
        source: IEnumerable<TransparentIdentifier<Order, Customer>> _ShapedQuery(
            queryContext: queryContext, 
            shaperCommandContext: SelectExpression: 
                SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate], [o.Customer].[CustomerID], [o.Customer].[Address], [o.Customer].[City], [o.Customer].[CompanyName], [o.Customer].[ContactName], [o.Customer].[ContactTitle], [o.Customer].[Country], [o.Customer].[Fax], [o.Customer].[Phone], [o.Customer].[PostalCode], [o.Customer].[Region]
                FROM [Orders] AS [o]
                LEFT JOIN [Customers] AS [o.Customer] ON [o].[CustomerID] = [o.Customer].[CustomerID]
                ORDER BY [o].[CustomerID]
            , 
            shaper: CompositeShaper`3
        )
        , 
        predicate: (TransparentIdentifier<Order, Customer> t1) => int Count(
            source: IEnumerable<ValueBuffer> Distinct(
                source: IEnumerable<ValueBuffer> _Select(
                    source: IEnumerable<TransparentIdentifier<TransparentIdentifier<ValueBuffer, ValueBuffer>, IEnumerable<Customer>>> _Where(
                        source: IEnumerable<TransparentIdentifier<TransparentIdentifier<ValueBuffer, ValueBuffer>, IEnumerable<Customer>>> _GroupJoin(
                            queryContext: (RelationalQueryContext) queryContext, 
                            source: IEnumerable<ValueBuffer> _Query(
                                queryContext: queryContext, 
                                shaperCommandContext: SelectExpression: 
[...]

PR #7495, opened 2017-01-27
[1.1.1] Fix to #7476 - Query : client-side result operators may cause extensive client-side evaluation for queries with GroupJoins (and/or optional navigations) in a subquery

maumar:
When we translate GroupJoin-SelectMany-DefaultIfEmpty we modify query model - we remove additional from clause containing DefaultIfEmpty subquery.
However, if we later discover that the subquery (or a subquery higher in the stack) needs client evaluation - e.g. by having client result operator, or calling a client-side method - we need to re-visit this subquery with client evaluation in mind.

Problem was that now the query model has been modified - in this specific case SelectMany-DefaultIfEmpty clause has been removed, so we can no longer recognize that this pattern can be simplified and we introduce client GroupJoin.

Fix is to save the structure of query model (and all subquery models that it contains) before we visit it - and then, if client evaluation is needed, restore the query model to it's original shape.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -12,8 +12,10 @@ using Microsoft.EntityFrameworkCore.Query.Expressions;
 using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 using Microsoft.EntityFrameworkCore.Query.ExpressionTranslators;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
+using Microsoft.EntityFrameworkCore.Query.Internal;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Utilities;
+using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Clauses.ResultOperators;
@@ -914,6 +916,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                         = _queryModelVisitor.Queries
                             .ToDictionary(k => k, s => s.Projection.Count);
 
+                    var queryModelMapping = new Dictionary<QueryModel, QueryModel>();
+                    subQueryModel.PopulateQueryModelMapping(queryModelMapping);
+
                     queryModelVisitor.VisitSubQueryModel(subQueryModel);
 
                     if (queryModelVisitor.Queries.Count == 1
@@ -932,6 +937,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
                         return selectExpression;
                     }
+
+                    subQueryModel.RecreateQueryModelFromMapping(queryModelMapping);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


maumar:
@anpete

maumar:
will send a new one for 2.0

maumar:
turns out this is also a fix for https://github.com/aspnet/EntityFramework/issues/7198 (1.1.1 issue)

src/Microsoft.EntityFrameworkCore/Query/Internal/QueryModelExtensions.cs
```diff
@@ -27,6 +28,97 @@ public static string Print([NotNull] this QueryModel queryModel)
...
+            {
+                _mapping = mapping;
+            }
+
+            protected override Expression VisitSubQuery(SubQueryExpression expression)
+            {
+                var queryModel = expression.QueryModel;
+
+                var newQueryModel = new QueryModel(queryModel.MainFromClause, queryModel.SelectClause);
+                ShallowCopy(queryModel, newQueryModel);
```
smitpatel: How about using `QueryModelBuilder` to build the qm instead of doing a copy like this?

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -914,6 +916,9 @@ var queriesProjectionCountMapping
                         = _queryModelVisitor.Queries
                             .ToDictionary(k => k, s => s.Projection.Count);
 
+                    var queryModelMapping = new Dictionary<QueryModel, QueryModel>();
```
maumar: added this to the other place also, thanks

src/Microsoft.EntityFrameworkCore/Query/Internal/QueryModelExtensions.cs
maumar: we need to keep the same instances of QuerySources

maumar:
@anpete ping

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -932,6 +937,8 @@ var queriesProjectionCountMapping
 
                         return selectExpression;
```
anpete: Does it still work if we exit on this branch?

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -976,6 +981,8 @@ var newExpression
                 }
```
anpete: Does it still work if we exit on this branch?

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
maumar: it's fine - we only need to recreate the original shape if the client evaluation is needed (for the second pass thru query model), and if we exist here (keeping the changes to the query model) but the outer querymodel needs client evaluation, code in the outer will recreate the original shape, undoing the changes done in the inner querymodel.

anpete: That's what I thought 😄

PR #7496, opened 2017-01-27
Add Metadata support for entity types with delegated identity.

AndriySvyryd:
An entity type with delegated identity is normally resolved by following the dependent ends of FKs from regular entity types.


AndriySvyryd:
@ajcvickers @anpete @divega

AndriySvyryd:
Updated: removed definition entity types and added the delegated entity types to 'model.GetEntityTypes()'

AndriySvyryd:
Ready for review

AndriySvyryd:
Is anyone against these being renamed to owned entity types to be consistent with Fluent API?

AndriySvyryd:
Rebased so it's ready for review (again)

src/EFCore/Extensions/ModelExtensions.cs
```diff
@@ -19,12 +19,50 @@ public static class ModelExtensions
...
+        public static IEntityType FindDelegatedIdentityEntityType(
+            [NotNull] this IModel model,
+            [NotNull] Type type,
+            [NotNull] string definingNavigationName,
+            [NotNull] IEntityType definingEntityType)
+            => Check.NotNull(model, nameof(model)).AsModel().FindDelegatedIdentityEntityType(
+                Check.NotNull(type, nameof(type)),
+                Check.NotNull(definingNavigationName, nameof(definingNavigationName)),
+                (EntityType)Check.NotNull(definingEntityType, nameof(definingEntityType)));
+
```
ajcvickers: For future consideration: I think it would be good to revisit the style discussion about inlining Check calls so that expression body syntax can be used. In cases like this, maybe we should err the other way...

src/EFCore/Metadata/Internal/Property.cs
```diff
@@ -222,7 +222,7 @@ public virtual void SetValueGenerated(ValueGenerated? valueGenerated, Configurat
             }
         }
 
-        private static ValueGenerated DefaultValueGenerated => ValueGenerated.Never;
+        private ValueGenerated DefaultValueGenerated => ValueGenerated.Never;
```
ajcvickers: Why removing the static on these? (I don't mind either way, but curious on the reasoning.)

AndriySvyryd: Consistency with other Default* properties

src/EFCore/Properties/CoreStrings.resx
```diff
@@ -711,6 +711,24 @@
...
+  <data name="ClashingNonDelegatedIdentityEntityType" xml:space="preserve">
+    <value>The delegated identity entity type '{entityType}' cannot be added to the model because a entity type with the same name already exists.</value>
+  </data>
+  <data name="DelegatedIdentityBaseType" xml:space="preserve">
+    <value>The type '{entityType}' cannot have delegated identity entity type '{baseType}' as the base type.</value>
+  </data>
+  <data name="DelegatedIdentityDerivedType" xml:space="preserve">
+    <value>The delegated identity entity type '{entityType}' cannot have a base type.</value>
+  </data>
   <data name="DuplicatePropertyInList" xml:space="preserve">
```
ajcvickers: We may need to be more helpful as to what the application actually tried to do in these cases. Like, what did I do that causes the FKs to target the same entity type? Also, I think saying "delegated identity" in these messages will translate to "something of which I have no understanding and will therefore ignore" when entering the developer's brain. So we may need to come up with more of a language to talk about these things. cc @divega

src/Shared/PropertyInfoExtensions.cs
```diff
@@ -44,6 +45,20 @@ public static Type FindCandidateNavigationPropertyType(this PropertyInfo propert
...
+
+            return Equals(propertyInfo, otherPropertyInfo)
+                   || (propertyInfo.Name == otherPropertyInfo.Name
+                       && (propertyInfo.DeclaringType == otherPropertyInfo.DeclaringType
+                           || propertyInfo.DeclaringType.GetTypeInfo().IsSubclassOf(otherPropertyInfo.DeclaringType)
+                           || otherPropertyInfo.DeclaringType.GetTypeInfo().IsSubclassOf(propertyInfo.DeclaringType)
+                           || propertyInfo.DeclaringType.GetTypeInfo().ImplementedInterfaces.Contains(otherPropertyInfo.DeclaringType)
+                           || otherPropertyInfo.DeclaringType.GetTypeInfo().ImplementedInterfaces.Contains(propertyInfo.DeclaringType)));
+        }
+
```
ajcvickers: What does "SameAs" really mean here?

AndriySvyryd: That both would be mapped to the same IProperty

src/EFCore/Metadata/Internal/Model.cs
```diff
@@ -241,8 +403,7 @@ private EntityType RemoveEntityType([NotNull] EntityType entityType)
             [CanBeNull] Type type,
             ConfigurationSource configurationSource)
         {
-            ConfigurationSource existingIgnoredConfigurationSource;
-            if (_ignoredTypeNames.TryGetValue(name, out existingIgnoredConfigurationSource))
+            if (_ignoredTypeNames.TryGetValue(name, out ConfigurationSource existingIgnoredConfigurationSource))
```
smitpatel: can use `var`.

PR #7498, opened 2017-01-27
Proof-of-concept for re-writing ServiceCollection.

ajcvickers:
This is an idea @anpete had to re-write the service collection in the equivalent of AddEntityFramework so that in cases where we need to inject a new service we can do so. This is working, but the problem is that it only works if the service has been registered as a concrete type by the provider. If it is registered as an object or delegete, then we don't have a type that we can use to re-register the original registration before overwriting the current one with our delegate. Thoughts? @anpete @AndriySvyryd @divega

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Infrastructure/ServiceCollectionProviderInfrastructure.cs
+++ b/src/Microsoft.EntityFrameworkCore/Infrastructure/ServiceCollectionProviderInfrastructure.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
+using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
 using Microsoft.EntityFrameworkCore.Internal;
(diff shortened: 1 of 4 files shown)


anpete:
@ajcvickers Do we need to re-register in those cases?

ajcvickers:
@anpete Yep, you are correct. We can wrap their delegate or object directly, like I do in the most recent pull request.

src/Microsoft.EntityFrameworkCore/Infrastructure/ServiceCollectionProviderInfrastructure.cs
```diff
@@ -29,6 +31,9 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
     /// </summary>
     public static class ServiceCollectionProviderInfrastructure
     {
+        private static readonly HashSet<Type> _servicesToInject 
+            = new HashSet<Type> { typeof(IValueGeneratorSelector) };
```
anpete: Possible alternative: Probe each implementation type for IServiceInjectionSite?

src/Microsoft.EntityFrameworkCore/Internal/IServiceInjectionSite.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public interface IServiceInjectionSite
```
anpete: Could even be called IServiceLocator 😄

anpete:
I like it, seems pretty clean.

src/Microsoft.EntityFrameworkCore/Infrastructure/ServiceCollectionProviderInfrastructure.cs
```diff
@@ -106,6 +111,31 @@ public static void TryAddDefaultEntityFrameworkServices([NotNull] IServiceCollec
...
+
+                if (_servicesToInject.Contains(serviceType))
+                {
+                    var concreteType = descriptor.ImplementationType;
+
+                    serviceCollection.Add(
+                        new ServiceDescriptor(concreteType, concreteType, descriptor.Lifetime));
+
+                    serviceCollection[serviceCollection.IndexOf(descriptor)]
+                        = new ServiceDescriptor(serviceType, p => InjectAdditionalServices(p, concreteType), descriptor.Lifetime);
```
divega: Not sure what value it adds but this came up in a conversation with @anpete: instead of registering a delegate we could in theory register a decorator that implements the service contract and gets the concrete type as well as any additional services injected in its constructor. Then this decorator could set properties on the instance of the concrete type in the constructor. 

It is basically the same solution but with a decorator instead of the delegate.

ajcvickers:
Discussed in team room and decided to go with service-re-write option. New PR will be sent out.

PR #7500, opened 2017-01-27
Proof-of-concept: Providing EF methods to do service registration by providers

ajcvickers:
Another idea from @anpete. Providers would have to use our methods to register their services. Apps doing service replacement should also do this.

Issues:
- Providers must use EF methods, but things will work if they don't until the time we add a fix that injects a new service.
- Services must have a different service type and implementation type (but we control this)
- Implementation type cannot also be registered or we will replace that registration (but hard to see how this would be needed)


ajcvickers:
@anpete @AndriySvyryd @divega New idea for service injection.

ajcvickers:
Discussed in team room and decided to go with service-re-write option. New PR will be sent out.

Issue #7290, opened 2016-12-21
Query compilation error or data corruption for queries with GroupJoin wrapped in a subquery when the subset of columns is being projected


shaulbehr:
This is a pretty specific corner case, but I've managed to repro it several times using different tables.

You need two tables, one with a non-mandatory FK to the other, e.g.

```
public class Salesperson
{
    public Guid SalespersonId {get; set;}
    public string Name {get; set;}
}
public class Customer
{
    public Guid CustomerId {get;set;}
    public Guid? SalespersonId {get;set;}
    public Salesperson Salesperson {get;set;}
}
```
Now you write a query to get the IDs of the top 100 Customers whose salesperson is named Fred:
```
var query = db.Customers
    .Where(q=>q.Salesperson.Name == "Fred")
    .Take(100)
    .Select(q=>q.SalespersonId)
    .ToList();
```
Result: a nasty error with sharp, pointy teeth:
[...]

PR #7513, opened 2017-01-30
[1.1.1] Fix to #7290 - Query compilation error or data corruption for queries with GroupJoin wrapped in a subquery when the subset of columns is being projected

maumar:
Problem was that we were incorrectly reasoning about which query sources need to be materialized in the GroupJoin scenario. Previously, when we encountered GroupJoin, we marked its main from clause and the joined clause for materialization, but we would not do that recursively. This causes problems if the GroupJoin (e.g. optional navigation) is in a subquery - we would not mark it for materialization. This would lead to compilation errors in some cases, and in others (when there was a client-evaluation component) to data corruption.

Fix is to use a QueryModelVisitor to walk the entire tree which allows all necessary query sources related to GroupJoin to be marked for materialization. This incurs a performance hit for some scenarios (groupjoin/optional navigation inside a subquery) because now we bring more columns from the database.

Second part of the fix is that in LiftSubquery, when we update query sources, we would sometimes try to rewrite EntityShapers into ValueBufferShapers (if the lifted query itself is not marked for materialization). This however is not safe to do in case of GroupJoin. Fix is to only perform this conversion shapers inside ShapedQuery method.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
@@ -771,11 +771,31 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
                             where (from od in context.OrderDetails
                                    where o.Customer.Country == od.Order.Customer.Country
                                    select od).Count() > 0
+                            where o.OrderID == 10643 || o.OrderID == 10692
                             select o;
 
                 var result = query.ToList();
 
-                Assert.Equal(830, result.Count);
+                Assert.Equal(2, result.Count);
+            }
+        }
+
+        [ConditionalFact]
+        public virtual void Navigation_in_subquery_referencing_outer_query_with_client_side_result_operator_and_count()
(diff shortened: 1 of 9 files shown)


maumar:
@anpete @divega

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryNavigationsTestBase.cs
```diff
@@ -771,11 +771,31 @@ public virtual void Navigation_in_subquery_referencing_outer_query()
                             where (from od in context.OrderDetails
                                    where o.Customer.Country == od.Order.Customer.Country
                                    select od).Count() > 0
+                            where o.OrderID == 10643 || o.OrderID == 10692
```
maumar: this query now becomes N+1, so i made it run quicker (was taking 30sec on my machine before I scoped it down)

maumar:
@anpete ping

maumar:
fixed in c99151d2b27d8ffc2cfed7d27eb099c399fcbdcf

Issue #7289, opened 2016-12-21
Query : Compilation error for queries with optional navigation, client side method and parameter


cherici-simone:
Hello, after migrate pur web api project to .net core 1.1 I run into a problem with a query that previously work correct.
This is the query:

```
var activityVersion = DbContext
			.ActivityVersions
			.Include(av=>av.Activity)
			.Include(av=>av.DeployedVersions)
			.Where(av => av.Activity.Name.Equals(serviceName, StringComparison.InvariantCultureIgnoreCase)
						&& av.Activity.ActivityType == Model.ActivityType.Service
						&& av.DeployedVersions.Any(dv => dv.Space.FullPath.Equals(servicePath, StringComparison.InvariantCultureIgnoreCase)))
			.OrderByDescending(ord => ord.Version)
			.FirstOrDefault();
```
Variable serviceName and servicePath are strings (the first is the action parameter and the second is a method variable) and Model.ActivityType.Service is an enum

The error is:
**variable '__serviceName_0' of type 'System.String' referenced from scope '', but it is not defined**

And here the full stack:
`at System.Linq.Expressions.Compiler.VariableBinder.Reference(ParameterExpression node, VariableStorageKind storage)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitParameter(ParameterExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitArguments(IArgumentProvider nodes)
   at System.Linq.Expressions.ExpressionVisitor.VisitMethodCall(MethodCallExpression node)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitUnary(UnaryExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitBinary(BinaryExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitConditional(ConditionalExpression node)
   at System.Linq.Expressions.ExpressionVisitor.Visit(ReadOnlyCollection`1 nodes)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitBlock(BlockExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitBinary(BinaryExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitBinary(BinaryExpression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitBinary(BinaryExpression node)
   at System.Linq.Expressions.ExpressionVisitor.Visit(ReadOnlyCollection`1 nodes)
[...]

PR #7517, opened 2017-01-31
[1.1.1] Fix to #7289 - Query : Compilation error for queries with optional navigation, client side method and parameter

maumar:
Problem was that DefaultQueryExpressionVisitor was not visiting extension expressions. During optional navigations we introduced NullConditional expressions to protect against null references - fix is to visit this expression as well.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/DefaultQueryExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/DefaultQueryExpressionVisitor.cs
@@ -2,9 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Linq.Expressions;
 using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 using Microsoft.EntityFrameworkCore.Query.Internal;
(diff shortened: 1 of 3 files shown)


maumar:
@anpete

maumar:
ping

maumar:
@anpete ping

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/DefaultQueryExpressionVisitor.cs
```diff
@@ -81,6 +83,31 @@ protected override Expression VisitParameter(ParameterExpression node)
...
+                var newAccessOperation = Visit(nullConditionalExpression.AccessOperation);
+
+                return newNullableCaller != nullConditionalExpression.NullableCaller
+                    || newCaller != nullConditionalExpression.Caller
+                    || newAccessOperation != nullConditionalExpression.AccessOperation
+                    ? new NullConditionalExpression(newNullableCaller, newCaller, newAccessOperation)
+                    : node;
+            }
+
+            Debug.Assert(false, "Unexpected extension expression: " + node);
```
anpete: Doesn't this block the reducible node pattern? I think we want to call base here.

maumar: fixed

maumar:
@anpete new version up

maumar:
fixed in 2b1edbfbeba1f38769e9c02b21ab081c7eae5b55

PR #7522, opened 2017-02-01
#7479 Add transaction identifier info the the commands tracing

fitzchak:
Continue work on #7483 PR.
Implementing the suggestion on the comment:
https://github.com/aspnet/EntityFramework/pull/7483#discussion_r98745178


src/Microsoft.EntityFrameworkCore.Relational/Microsoft.EntityFrameworkCore.Relational.csproj
```diff
@@ -78,8 +78,6 @@
     <Compile Include="Infrastructure\RelationalEventId.cs" />
     <Compile Include="Infrastructure\RelationalOptionsExtension.cs" />
     <Compile Include="Internal\RelationalDiagnostics.cs" />
-    <Compile Include="Internal\RelationalDiagnosticSourceAfterMessage.cs" />
-    <Compile Include="Internal\RelationalDiagnosticSourceBeforeMessage.cs" />
```
avanderhoorn: Not that it bothers me too much, but wondering why these where removed? I seem to remember that @anpete created them originally for a specific use case.

avanderhoorn:
Couple of things... 

I noticed that before I got to add final feedback to the previous PR, we ended up with both the a `connectionId` and `instanceId` on the connection events, what is the difference here? I know we added `instanceId` on the command as the command could be reused multiple times and we needed to be able to track specific uses. Is this the same with the connect?

Additionally, for this PR, since we have added the `transactionId` to the command, I wonder if we should add it to the connection as well?

src/Microsoft.EntityFrameworkCore.Relational/Microsoft.EntityFrameworkCore.Relational.csproj
fitzchak: I removed them because there is no need for them and there are two comments to remove them after an issue resolved, which is already true.

fitzchak:
InstanceId is different from the ConnectionId. The instance id is an id for a specific command execution. It is shared between the beforeCommand and AfterCommand or ErrorCommand. Commands cannot be reused. But the connection id is used accross the all connection life time as it is per DbConnection.

We add the connectionId already, and the DbConnection is accessible using the command.DbConnection.
It may be wise to pass the RelationalConnection here directly in case new info would be added to this class in the future, but I'm not sure if this would be the case.

src/Microsoft.EntityFrameworkCore.Relational/Microsoft.EntityFrameworkCore.Relational.csproj
anpete: I think it's fine. It looks like they were created (900bf16dcc) as a workaround for something I'm guessing is no longer an issue.

avanderhoorn:
@fitzchak You're right, I missed the fact that it was on the command... for some reason I thought that was on the connection events. I'd be open to`RelationalConnection` being passed in, but at this point I think its just the `transactionId` missing connection...

fitzchak:
So I added the `transactionId`. 
Please let me know if there is anything else do you think we can improve here.

anpete:
@avanderhoorn Any further feedback here?

avanderhoorn:
Sorry been on paternity leave. Looks good the go!

anpete:
@fitzchak, I'm seeing test failures when trying to build this. Did you run the tests?

fitzchak:
@anpete I have run them but something changed from than. I resolved the conflict, but I cannot investigate the failing test now as I don't have VS 2017 RC installed, and I see this is a requirement now.

anpete:
@fitzchak Any plans to update this?

fitzchak:
Yes, I'll do so this in a day or two, after installing VS2017.

fitzchak:
Done

smitpatel:
@fitzchak - This PR build fails on travis CI, around diagnostics tests only. Can you have a look at it and fix the errors?

ajcvickers:
Triage decision: close this PR since it is not building and track the work to do with issue #7479.

avanderhoorn:
@fitzchak any chance that you can get things working again so that we can get this in?

PR #7552, opened 2017-02-03
Add fluent API for configuring owned entity types.

AndriySvyryd:
@anpete, @ajcvickers, @rowanmiller, @divega See OwnedTypesTestBase.cs for some examples

### DIFF
--- a/src/EFCore/Metadata/Builders/EntityTypeBuilder`.cs
+++ b/src/EFCore/Metadata/Builders/EntityTypeBuilder`.cs
@@ -4,9 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Linq.Expressions;
+using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 
@@ -35,18 +37,8 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Builders
         }
 
         /// <summary>
-        ///     Creates a new builder based on the provided internal builder. This overridden implementation creates
-        ///     <see cref="EntityTypeBuilder{TEntity}" /> instances so that logic inherited from the base class will
-        ///     use those instead of <see cref="EntityTypeBuilder" />.
-        /// </summary>
-        /// <param name="builder"> The internal builder to create the new builder from. </param>
-        /// <returns> The newly created builder. </returns>
-        protected override EntityTypeBuilder New(InternalEntityTypeBuilder builder)
-            => new EntityTypeBuilder<TEntity>(builder);
-
-        /// <summary>
         ///     Adds or updates an annotation on the entity type. If an annotation with the key specified in
-        ///     <paramref name="annotation" /> already exists it's value will be updated.
+        ///     <paramref name="annotation" /> already exists its value will be updated.
         /// </summary>
         /// <param name="annotation"> The key of the annotation to be added or updated. </param>
         /// <param name="value"> The value to be stored in the annotation. </param>
@@ -60,7 +52,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Builders
         /// <param name="name"> The name of the base type. </param>
         /// <returns> The same builder instance so that multiple configuration calls can be chained. </returns>
         public new virtual EntityTypeBuilder<TEntity> HasBaseType([CanBeNull] string name)
-            => (EntityTypeBuilder<TEntity>)base.HasBaseType(name);
+            => new EntityTypeBuilder<TEntity>(Builder.HasBaseType(name, ConfigurationSource.Explicit));
 
         /// <summary>
         ///     Sets the base type of this entity in an inheritance hierarchy.
@@ -68,7 +60,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Builders
         /// <param name="entityType"> The base type. </param>
         /// <returns> The same builder instance so that multiple configuration calls can be chained. </returns>
         public new virtual EntityTypeBuilder<TEntity> HasBaseType([CanBeNull] Type entityType)
-            => (EntityTypeBuilder<TEntity>)base.HasBaseType(entityType);
+            => new EntityTypeBuilder<TEntity>(Builder.HasBaseType(entityType, ConfigurationSource.Explicit));
 
         /// <summary>
         ///     Sets the base type of this entity in an inheritance hierarchy.
@@ -76,7 +68,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Builders
         /// <typeparam name="TBaseType"> The base type. </typeparam>
         /// <returns> The same builder instance so that multiple configuration calls can be chained. </returns>
         public virtual EntityTypeBuilder<TEntity> HasBaseType<TBaseType>()
-            => (EntityTypeBuilder<TEntity>)base.HasBaseType(typeof(TBaseType));
+            => HasBaseType(typeof(TBaseType));
 
         /// <summary>
         ///     Sets the properties that make up the primary key for this entity type.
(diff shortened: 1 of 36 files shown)


src/Microsoft.EntityFrameworkCore/Metadata/Builders/EntityTypeBuilder.cs
```diff
@@ -194,6 +195,60 @@ public virtual IndexBuilder HasIndex([NotNull] params string[] propertyNames)
 
         /// <summary>
         ///     <para>
+        ///         Configures a relationship where this entity type provides identity to
+        ///         the other type in the relationship.
```
rowanmiller: Maybe this?

> Configures a relationship where the target entity is owned by (or part of) this entity. The target entity does not have it's own key. It's key value is always controlled by the entity is belongs to.

src/Microsoft.EntityFrameworkCore/Metadata/Builders/EntityTypeBuilder`.cs
```diff
@@ -179,6 +176,33 @@ public virtual IndexBuilder HasIndex([NotNull] Expression<Func<TEntity, object>>
...
         /// <summary>
         ///     <para>
+        ///         Configures a relationship where this entity type provides identity to
+        ///         the other type in the relationship.
+        ///     </para>
+        /// </summary>
+        /// <typeparam name="TRelatedEntity"> The entity type that this relationship targets. </typeparam>
+        /// <param name="navigationExpression">
+        ///     A lambda expression representing the reference navigation property on this entity type that represents
+        ///     the relationship (<c>post => post.Blog</c>).
```
rowanmiller: I think the example should be representative of this type of relationship, so maybe `<c>customer => customer.Address</c>`

src/Microsoft.EntityFrameworkCore/Metadata/Builders/EntityTypeBuilder.cs
AndriySvyryd: That's not accurate. The owned type can be the principal in the relationship between it and the owner type.
The owner, or more specifically the navigation on the owner, just provides a way of finding the related type.

rowanmiller:
From an API perspective, this looks pretty good to me. We may come up with a better name for `Owns(...)`... but I don't think that is a bad name.

Left a few comments with doc suggestions

src/Microsoft.EntityFrameworkCore/Metadata/Builders/EntityTypeBuilder.cs
rowanmiller: Let's discuss in person 😄

AndriySvyryd:
Updated with feedback from the design meeting

test/EFCore.Tests/ModelBuilderTest/OwnedTypesTestBase.cs
```diff
@@ -0,0 +1,63 @@
...
+using Xunit;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.EntityFrameworkCore.Tests
+{
+    public abstract partial class ModelBuilderTest
+    {
+        public abstract class OwnedTypesTestBase : ModelBuilderTestBase
+        {
+            //[Fact]
```
ajcvickers: > //[Fact] [](start = 11, length = 9)

Why are the tests commented out?

src/EFCore/Metadata/Builders/EntityTypeBuilder.cs
```diff
@@ -35,13 +34,6 @@ public EntityTypeBuilder([NotNull] InternalEntityTypeBuilder builder)
             Builder = builder;
         }
 
-        /// <summary>
-        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
-        ///     directly from your code. This API may change or be removed in future releases.
-        /// </summary>
-        protected virtual EntityTypeBuilder New([NotNull] InternalEntityTypeBuilder builder)
-            => new EntityTypeBuilder(builder);
-
```
ajcvickers: Why did we used to have this and why do we not need it now?

test/EFCore.Tests/ModelBuilderTest/OwnedTypesTestBase.cs
AndriySvyryd: They don't pass without changes to conventions (next PR)

AndriySvyryd: (The tests are still valuable as a compile-time check of the API)

src/EFCore/Metadata/Builders/EntityTypeBuilder.cs
AndriySvyryd: It allowed to reuse the base implementation.
Now I realized that the public builders shouldn't have much implementation apart from calling into the internal builder, so reusing the same boilerplate code is not worth a virtual call and a cast

src/EFCore/Metadata/Builders/EntityTypeBuilder`.cs
```diff
@@ -135,28 +127,16 @@ public virtual PropertyBuilder<TProperty> Property<TProperty>([NotNull] Expressi
...
         /// <param name="propertyName"> The name of then property to be removed from the entity type. </param>
         public new virtual EntityTypeBuilder<TEntity> Ignore([NotNull] string propertyName)
-        {
-            Check.NotEmpty(propertyName, nameof(propertyName));
-
-            Builder.Ignore(propertyName, ConfigurationSource.Explicit);
-
-            return this;
-        }
+            => (EntityTypeBuilder<TEntity>)base.Ignore(propertyName);
```
smitpatel: null check gone.

AndriySvyryd: The base method has the check

test/EFCore.Tests/ModelBuilderTest/OwnedTypesTestBase.cs
```diff
@@ -0,0 +1,63 @@
...
+            public virtual void Can_use_nested_closure()
+            {
+                var modelBuilder = CreateModelBuilder();
+                var model = modelBuilder.Model;
+
+                modelBuilder.Entity<Customer>().OwnsOne(
+                    c => c.Details,
+                    r => r.HasAnnotation("foo", "bar")
+                        .HasOwnershipAnnotation("bar", "foo"));
+
```
ajcvickers: Get the concept (HasAnnotation puts the annotation on the entity type, while HasOwnershipAnnotation puts it on the FK) but not sure it is the best API. Could consider one to be YAGI, or maybe also explicitly name the entity type one to HasTypeAnnotation. @divega @anpete Ideas?

divega: I tend to agree with the suggestion but honestly I feel like I need a refresher of the whole thing to have more context.

tinchou:
Hey guys,

Would you mind sharing a quick description of what Owned Entity Types are? I can't quite figure it out from looking at [`OwnedTypesTestBase.cs`](https://github.com/aspnet/EntityFramework/blob/dev/test/EFCore.Tests/ModelBuilderTest/OwnedTypesTestBase.cs) or some of the other PRs.

Thanks!

AndriySvyryd:
@tinchou In short this enables scenarios like:
```C#
class Customer
{
    public int Id { get; set; }
    public Address BillingAddress { get; set; }
    public Address ShippingAddress { get; set; }
}

class Address
{
    public string Street { get; set; }
    public string City { get; set; }
}
```
There are two concepts here:
* __Delegated Identity Entity Types__ are entity types that are mapped to the same CLR type, since they cannot be identified by their name or CLR type they can only be accessed through a navigation from another entity type, hence their identity is delegated. It is possible to have multiple navigations to a DIET so its identity is not unique. The primary key is usually the same as the defining foreign key and usually consists of shadow properties. 
* __Owned Entity Types__ are DIETs where the defining entity type is marked as the owner. There can only be one navigation to an owned entity type. The owner must be the principal of the defining relationship. The navigation to an owned entity type will be automatically included in the query.

tinchou:
So how would the tables look like in that case? I'm thinking the following, but that wouldn't be any different from regular relationships.

Id | BillingAddressId | ShippingAddressId
---|-------------------|------
... | ... | ...

AddressId | Street | City
---|-------------------|------
... | ... | ...

Thanks!

AndriySvyryd:
The `Customer` is the principal and the `Address` DIETs are mapped to different tables, so it would be:

`Customers`:

| Id |
|---|
|...|

`Customers_BillingAddresses`:

CustomerId | Street | City
---|-------------------|------
... | ... | ...

`Customers_ShippingAddresses`:

CustomerId | Street | City
---|-------------------|------
... | ... | ...

However when we implement table splitting then by convention this would be mapped to a single table:

`Customers`:

Id | BillingAddress_Street | BillingAddress_City | ShippingAddress_Street | ShippingAddress_City
---|-------------------|------|-------------------|------
... | ... | ... | ... | ...

tinchou:
Awesome!

Issue #7271, opened 2016-12-19
Conditional sum using linq for Or condition


datasecurity:
When I use two conditions joined by an OR, the result is not correct for SQL Server.

How can I fix it?

This is my LINQ code and result in SQL (that reflection created for me):

``` C#
query.Where(p => ((p.Code == "100000") Or p.Code.EndsWith("200")));
```

``` C# 
query.Where(p => (p.year == "2015"))}
```

I added this `where `clause at runtime, now I add another extension method and it's not working:

``` C# 
query.sum(p => p.value)
```

Exception: 

> An exception of type 'System.Data.SqlClient.SqlException' occurred in Microsoft.EntityFrameworkCore.dll but was not handled in user code
> 
> Additional information: An expression of non-boolean type specified in a context where a condition is expected, near 'AND'.

SQL translated:

``` SQL
SELECT SUM([e].[Value])
FROM [acc].[Data161] AS [e]
WHERE (CASE
          WHEN RIGHT([e].[Code], LEN(N'201')) = N'201'
             THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT)
       END | 
       CASE
          WHEN RIGHT([e].[Code], LEN(N'199')) = N'199'
             THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT)
       END) 
  AND ([e].[SetadCode] = N'161')
``` 

The correct SQL should have `= 1` before the `AND`.

But without sum its works fine and add a `= 1` to SQL command

tuespetre:
This turned out not be because of the `Sum` but because of the interaction between `BinaryExpression`s and `ConditionalExpression`s.

divega:
Assigning issue and PR to @smitpatel.

liamdawson:
Does someone have a workaround for this issue?

smitpatel:
Use `OrElse` instead of `Or` (Or is bitwise operator, OrElse is logical operator)

liamdawson:
Would `AndAlso` also be the appropriate change for `And` expressions?

smitpatel:
Yes.

Issue #7520, opened 2017-02-01
Query: List of refactoring ideas


anpete:
Brain dump of potential refactorings in the query pipeline captured from a discussion today with @smitpatel and @maumar.

~- [ ] Immutable SelectExpression via facade/visitors~ Not worth
- [x] AliasExpressions everywhere or ColumnRefExpressions for referentially consistent SQL tree. (@smitpatel) #7950 
- [x] Include via QueryModel (@anpete) #7974 
~- [ ] Include/nav-prop synergy via early QM rewrites.~
~- [ ] Immutable QueryModel~
~- [ ] First-class * expression (perhaps via exploded column containment idea)~ We do not have * projection for now. We can introduce it in future.
- [x] SearchConditionTranslatingVisitor rewrite (#7554)
- [x] Subquery lifting at query model level (#7613)
- [x] Consistentize use of visitors/extension nodes, VisitChildren etc.
- [x] Apply CompositePredicateExpressionVisitor to other parts of the SelectExpression (#7608)
~- [ ] Evaluate ClientEval flags.~ We do not have client eval flag. 
- [x] Investigate NullConditionalExpression processing in RequiresMaterializationExpressionVisitor (#7746)
- [ ] clean up Nav + Entity Equality rewriting visitors (#7773)
- [ ] Clean up & Improve QueryOptimizer (#7831)
- [ ] Avoid unnecessary pushdown of subqueries (#8094)
~- [ ] Investigate DRYing up null expressions (NullableExpression and NullConditionalExpression)~ Removed NullableExpression
~- [ ] Clean-up logic around RemoveConvert (#8318)~ Nomore RemoveConvert
- [ ] Rename and clean-up EQMV.OptimizeQueryModel
- [x] log contents of compiled code (#9455)

@smitpatel, @maumar Feel free to add more detail here.

tuespetre:
[...]

PR #7554, opened 2017-02-04
Query: Refactor SearchConditionTranslatingVisitor

smitpatel:
Resolves issue #7271
part of #7520

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -321,23 +321,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
             newExpression = new PredicateReductionExpressionOptimizer().Visit(newExpression);
             newExpression = new PredicateNegationExpressionOptimizer().Visit(newExpression);
             newExpression = new ReducingExpressionVisitor().Visit(newExpression);
-
-            var searchConditionTranslatingVisitor = new SearchConditionTranslatingVisitor(searchCondition);
-
-            newExpression = searchConditionTranslatingVisitor.Visit(newExpression);
-
-            if (searchCondition && !SearchConditionTranslatingVisitor.IsSearchCondition(newExpression))
-            {
-                var constantExpression = newExpression as ConstantExpression;
-
-                if ((constantExpression != null)
-                    && (bool)constantExpression.Value)
-                {
-                    return null;
-                }
-
-                return Expression.Equal(newExpression, Expression.Constant(true, typeof(bool)));
-            }
+            newExpression = new BooleanExpressionTranslatingVisitor().Translate(newExpression, searchCondition: searchCondition);
 
             return newExpression;
         }
@@ -1138,7 +1122,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
                     }
 
                     _typeMapping = parentTypeMapping;
-                    
+
                     break;
                 }
             }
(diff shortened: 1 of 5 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
-                       || expression is ExistsExpression
-                       || expression is StringCompareExpression;
+                return newExpression;
             }
 
             protected override Expression VisitBinary(BinaryExpression expression)
             {
-                Expression newLeft;
-                Expression newRight;
+                // Visit the children
```
maumar: remove this comment - seems inaccurate (visit children comes later)

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
-                                : Expression.Convert(newExpression, typeof(bool)),
-                            Expression.Constant(true, typeof(bool)),
-                            Expression.Constant(false, typeof(bool)));
-                    }
-                }
+                _isSearchCondition = parentIsSearchCondition;
 
-                newLeft = AdjustExpressionType(Visit(expression.Left), expression.Left.Type);
-                newRight = AdjustExpressionType(Visit(expression.Right), expression.Right.Type);
+                // Update the node
```
maumar: remove this comment as seems redundant - fairly obvious what the code is doing here

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
+                expression = expression.Update(newLeft, expression.Conversion, newRight);
 
-                return expression.Update(newLeft, expression.Conversion, newRight);
+                // Do conversion
+                return _isSearchCondition
```
maumar: This pattern in repeated several times, consider DRYing it up.

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
             }
 
-            private static Expression AdjustExpressionType(Expression expression, Type expectedType)
-                => expression.Type != expectedType
-                    ? Expression.Convert(expression, expectedType)
-                    : expression;
-
             protected override Expression VisitConditional(ConditionalExpression expression)
             {
+                // Visit the children
```
maumar: remove

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
+                // Results are always values
                 _isSearchCondition = false;
-
                 var ifTrue = Visit(expression.IfTrue);
                 var ifFalse = Visit(expression.IfFalse);
 
                 _isSearchCondition = parentIsSearchCondition;
 
-                var newExpression = Expression.Condition(test, ifTrue, ifFalse);
+                // Update the node
```
maumar: remove

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
+                // NOT(A) => A == false
+                if (expression.NodeType == ExpressionType.Not
+                    && expression.Operand.IsSimpleExpression())
                 {
-                    _isSearchCondition = false;
+                    return Visit(MakeCompareToExpression(expression.Operand, compareTo: false));
                 }
 
-                if (expression.NodeType == ExpressionType.Not)
+                // Visit the children
```
maumar: remove comment

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
-                if (_isSearchCondition)
-                {
-                    if (expression.NodeType == ExpressionType.Not
-                        && expression.Operand.IsSimpleExpression())
-                    {
-                        return Expression.Equal(
-                            expression.Operand,
-                            Expression.Constant(false, expression.Operand.Type));
-                    }
+                // Update the node
```
maumar: remove comment

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
+                    ? ConvertToSearchCondition(expression)
+                    : ConvertToValue(expression);
             }
 
             protected override Expression VisitExtension(Expression expression)
             {
-                if (_isSearchCondition)
-                {
-                    var parentIsSearchCondition = _isSearchCondition;
+                // Visit the children
```
maumar: remove comment

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
...
+                       || expression.IsLogicalOperation()
+                       || expression is ExistsExpression
+                       || expression is InExpression
+                       || expression is IsNullExpression
+                       || expression is LikeExpression
+                       || expression is StringCompareExpression;
             }
 
-            protected override Expression VisitParameter(ParameterExpression expression)
+            private static Expression MakeCompareToExpression(Expression expression, bool compareTo)
```
maumar: nit: maybe 'Build' or 'Create' instead of 'Make'

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -321,23 +321,7 @@ var newExpression
...
-
-                if ((constantExpression != null)
-                    && (bool)constantExpression.Value)
-                {
-                    return null;
-                }
-
-                return Expression.Equal(newExpression, Expression.Constant(true, typeof(bool)));
-            }
+            newExpression = new SearchConditionTranslatingVisitor(searchCondition).Convert(newExpression);
```
anpete: Can we call Visit instead on Convert?

anpete:
Can we move from "SearchCondition" to "Predicate" for consistency?

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1623,245 +1607,219 @@ var constantExpression
 
         private class SearchConditionTranslatingVisitor : RelinqExpressionVisitor
         {
+            /*
```
anpete: Put this in /// Summary doc comment

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
anpete: I find the naming confusing. Is it that this visitor is a predicate translator that performs a different translation when the target expression is a search condition?

smitpatel: I also need suggestions for name. Visit would be starting method for every expression. But the code block should be applied only at top level or starting expression. I moved the extra handling at top level inside the visitor to make it more contained in itself.

smitpatel:
It was named "SearchCondition" because T-Sql uses that term. I am fine with using "Predicate" or anything else also.

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
smitpatel: How about `Translate` as method name?

This is visitor which converts value to search condition or vice-versa as needed. It is applied to predicates/join conditions & projections. After thinking more about name of the visitor,
I believe it shouldn't be named with Predicate.

anpete: Translate is fine. I was looking for symmetry with the 3 lines above, but I see why it needs a custom entry point method.

How about something like:
```c# 
new BooleanExpressionTranslatingVisitor().Translate(expression, searchCondition: false);
```

PR #7555, opened 2017-02-04
Fix some pluralisation issues in method comments

WiredUK:
Simple change to fix pluralisation issues on the XML comments of 2 `*Range` methods


AndriySvyryd:
@WiredUK Thanks for your contribution!

WiredUK:
@AndriySvyryd My pleasure! I assumed that these comments are used for the [API reference docs](https://docs.microsoft.com/en-gb/ef/core/api/microsoft.entityframeworkcore.dbset-1) so will appear there at some point in the future?

AndriySvyryd:
@WiredUK Yes, probably on the next release.

Issue #7560, opened 2017-02-06
Server-side evaluation fails if query modifiers precede condition


breyed:
The warning RelationalEventId.QueryClientEvaluationWarning is issued if `Include` precedes a condition when queries are chained. Code like this will trigger the warning:

    IQueryable<Customer> subQuery = ...
    var query = subQuery.AsTracking().Include(...).Single(condition)

By contrast, code like this does not:

    IQueryable<Customer> subQuery = ...
    var query = subQuery.Where(condition).AsTracking().Include(...).Single()

smitpatel:
Can you post the warning text also?

breyed:
> Warning as error exception for warning 'RelationalEventId.QueryClientEvaluationWarning': The LINQ expression '([c].AccountNumber == __accountNumber_2)' could not be translated and will be evaluated locally. To suppress this Exception use the DbContextOptionsBuilder.ConfigureWarnings API. ConfigureWarnings can be used when overriding the DbContext.OnConfiguring method or using AddDbContext on the application service provider.

tuespetre:
@breyed I wonder if this has more to do with the call to `Single(predicate)`. Does it work if you put `Where(predicate)` after the `AsTracking()` and `Include()` but before `Single()`?

breyed:
IQueryable<Customer> subQuery =
        from c in Db.Customers
        where c.BranchId == branchId
        orderby c.LastName, c.FirstName
        select c;

smitpatel:
@breyed - Which version of EF are you using?
I tried reproducing this, I received warning in both cases. In one it says cannot translate the equality (which is inside `Single()`) and in second it says cannot translate `Single()`. The cause is `Single` call which needs client evaluation to throw exception.
Can you share your full queries and model classes for reproduction of the issue?

smitpatel:
The subquery has an orderby clause. In Sql you cannot have orderby inside a subquery. Therefore it fails to translate outer query and shows client eval warning.
In the case when where is before as tracking, the Re-Linq query parser flattens out the whole model so there is nothing other than `Single()` outside the subquery. If Where comes after AsTracking it cannot translate it to server.

smitpatel:
Verify this is fixed when #7613 is closed.

smitpatel:
@maumar - Assigning this should be fixed with lifting of order by.

maumar:
Fixed in f022d66bc3a461851a6250536399cd3b8f8bf8d2

PR #7567, opened 2017-02-07
Add tests to verify Single predicate behavior with AsTracking/Include (for #7560)

tuespetre:
🌵 🌞


smitpatel:
This should not be required.
The problem in #7560 should be around how the query model is generated based on the condition being inside single & in where clause. Condition inside single call is probably creating a subquery in query model giving rise to warning because single inside subquery needs to be client evaluated.

Unless tracking is disabled at context level, `AsTracking` has no effect as all queries would be tracking query. Also Include is applied through query annotations so position of it should not have much effect in query model.

As the OP commented on #7560, the query involves a subquery instead of just DBSet.

PR #7578, opened 2017-02-09
Push down subquery before using COUNT/COUNT_BIG if the query has an offset (fixes #7523)

### LINKED ISSUE #7523: Skip with Count always returns 0
We are working to have Core support for IBM EntityFramework Provider.
For one of our internal test application, we have to skip few rows and and return remaining total number of rows.
Our query is:
context.ArchTable1.OrderBy(c=>c.Id).Skip(1).Count()

It always returns output:
skip count 0
We have used UseRowNumberForPaging() via optionsBuilder.

The similar Application when we run against SQL Server 2008, we do see same issue.
Generated SQL is:
SELECT [t].[c0]
FROM (
    SELECT COUNT(*) AS [c0], ROW_NUMBER() OVER(ORDER BY @@RowCount) AS [__RowNumber__]
    FROM [ArchTable1] AS [a]
) AS [t]
WHERE [t].[__RowNumber__] > 1

With manual modification in above query, if we run from SQL management Studio it returns the desired output:
  SELECT COUNT(*) AS [c0] from
    (select ROW_NUMBER() OVER(ORDER BY @@RowCount) AS [__RowNumber__]
    FROM [ArchTable1] AS [a]
) AS [t]
WHERE [t].[__RowNumber__] > 1

Now questions are:
1. Though user mentioned OrdeyBy clause, but why it's not reflecting in generated query ?
2. PushDownSubquery() basically pushes down the current query. How do we inject some new SQL (i.e. select statement to achieve above working query) in between in current selectExpression ?

tuespetre:
🌞

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
@@ -379,6 +379,11 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
         private static Expression HandleCount(HandlerContext handlerContext)
         {
+            if (handlerContext.SelectExpression.Offset != null)
+            {
+                handlerContext.SelectExpression.PushDownSubquery();
+            }
+
             handlerContext.SelectExpression
                 .SetProjectionExpression(
                     new SqlFunctionExpression(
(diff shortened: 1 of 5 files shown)


tuespetre:
:up::date:

I copied the two tests and added the ordering to the copies so both cases will be covered 'on the record'. I hope that's alright. I also added the overrides for all four tests in `RowNumberPagingTest`.

smitpatel:
Merged via eddbfa2a377a72f2330266a9f0881ac172736a19

PR #7581, opened 2017-02-10
Stop using '==' operator for reflection object comparisons.

anpete:
On .NET Core reflection types such as MethodInfo no longer override the == operator and so we get reference equality. This happens to work when the underlying runtime is caching these objects, but caching is not guaranteed. E.g. .NET Native is much less eager to cache here.

The fix is to use the instance Equals method (or object.Equals) to perform the comparison.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionTranslators/ParameterlessInstanceMethodCallTranslator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionTranslators/ParameterlessInstanceMethodCallTranslator.cs
@@ -16,8 +16,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators
     /// </summary>
     public abstract class ParameterlessInstanceMethodCallTranslator : IMethodCallTranslator
     {
-        private readonly Type _declaringType;
-        private readonly string _clrMethodName;
+        private readonly MethodInfo _methodInfo;
         private readonly string _sqlFunctionName;
 
         /// <summary>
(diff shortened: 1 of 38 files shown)


ErikEJ:
Providers beware?

roji:
Yep, seems like somethings providers should do in their translators as well...

ErikEJ:
@roji At least providers that support .NET Core...

divega:
@roji @ErikEJ added the label, but please hang in there while we work out a few more details with the .NET Native and .NET Core teams.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionTranslators/ParameterlessInstanceMethodCallTranslator.cs
```diff
@@ -43,10 +43,7 @@ public abstract class ParameterlessInstanceMethodCallTranslator : IMethodCallTra
         /// </returns>
         public virtual Expression Translate(MethodCallExpression methodCallExpression)
         {
-            var methodInfo = _declaringType.GetTypeInfo()
-                .GetDeclaredMethods(_clrMethodName).SingleOrDefault(m => !m.GetParameters().Any());
-
-            if (methodInfo == methodCallExpression.Method)
+            if (_methodInfo.Equals(methodCallExpression.Method))
```
MichalStrehovsky: I'm only looking at the diff, but you're populating `_methodInfo` with a call to `SingleOrDefault` above, which means `null` might be an expected value for `_methodInfo`.

`if (Object.Equals(_methodInfo, methodCallExpression.Method))` might be a better check in that case.

anpete: Good catch, thanks.

PR #7543, opened 2017-02-03
Group By and Outer Apply and Query Optimizations and, and...

tuespetre:
Okay, please don't be mad at me -- I'm feeling kind 👨‍🍳 of scrambled after my deep dive into this. It's a pretty big PR and touches a lot of things. I didn't plan on it being this way but as I was working on my initial goal (getting `GROUP BY` to work) I ended up breaking some things, and in order to fix them, other things broke, and I had to fix them, ad nauseum. 

Now I've finally got the tests all passing (save for one SQLite oddity, but I will cover that in a sec) and successfully rebased onto the last month's worth of commits.

The gist of what I've done:

- Added `GROUP BY` support
- Added `LEFT JOIN LATERAL`/`OUTER APPLY` support
- Optimized generated queries to avoid materialization whenever possible, resulting in much smaller SQL statements, less data over the wire, etc.

Now, there are several points I know still need to be covered:

- [x] Complex `GroupBy` keys
- [x] I'm suspicious of some of the test SQL output changes surrounding discriminator columns and would like to double-check
- [x] I think the whole `QueryFlattener` class and its factory could disappear, it's only used for one call now
- [x] Plenty more tests for `GROUP BY`

...I just thought that for as long as this has already taken, and for as large of a changeset as this is, I would get some kind of ball rolling here.

Issue #5230, opened 2016-04-29
Query: Optimize predicate on discriminator


smitpatel:
Current visiting FromClause adds predicate on discriminator column. If WhereClause (or any other way) another predicate on discriminator column is added then we should combine them in some way to improve readability.
`Can_use_is_kiwi` test
Generated SQL:

```
SELECT [a].[Species], [a].[CountryId], [a].[Discriminator], [a].[Name], [a].[EagleId], [a].[IsFlightless], [a].[Group], [a].[FoundOn]
FROM [Animal] AS [a]
WHERE [a].[Discriminator] IN (N'Kiwi', N'Eagle') AND ([a].[Discriminator] = N'Kiwi')
```

This is just readability improvement. Sql Server optimize the predicate by self for perf.

ilmax:
@smitpatel FYI I think I've addressed this, will have to double check if everything is ok and the submit a PR for evaluation.

ilmax:
I have a problem building from `build.cmd` and everything works just fine using visual studio.
The build command output is [here](https://gist.github.com/ilmax/a5a83ada6c9653c5930b552979652d0f)

Note that I didn't change anything in the project Microsoft.EntityFrameworkCore.Tools so I think I'm missing something basic. Can someone please point me to the right direction for executing all tests?

Thank you

rdefreitas:
@ilmax I haven't come across that one, but I did run into a number of other problems myself if I didn't remember to run build initialize the first time out before eventually running build.

ilmax:
I've run build initialize, but that didn't seems to solve the problem. 
I noted also that the fix is a little bit more complicate than I initially expected because if we would like to eliminate the redundant clause always (i.e. in every possible query), I have to make a little bit more work and I'm not sure the gain is worth the effort especially because we can potentially visit a very complicate query expression so all that work can slow down the perf without generating a better performing sql.

I will try to push a PR on this to gather a little bit of feedback soon

tuespetre:
Pretty sure I covered this in #7543

tuespetre:
Scratch that ❌ I did not cover this

ajcvickers:
@maumar to de-dupe.

maumar:
dupe of #11050

PR #7587, opened 2017-02-10
Optimize discriminator predicates into IN expressions in more cases

tuespetre:
This PR isolates one of the optimizations I made in #7543 into a smaller commit. It will shave two files off of that PR. This PR partially address #5230.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
@@ -28,29 +28,33 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public override Expression Visit(Expression expression)
+        protected override Expression VisitExtension(Expression expression)
         {
-            if (expression != null)
+            var selectExpression = expression as SelectExpression;
+
+            if (selectExpression?.Predicate != null)
             {
-                expression = new EqualityPredicateInExpressionOptimizer().Visit(expression);
+                var predicate = new EqualityPredicateInExpressionOptimizer().Visit(selectExpression.Predicate);
 
                 var predicateNegationExpressionOptimizer = new PredicateNegationExpressionOptimizer();
 
-                expression = predicateNegationExpressionOptimizer.Visit(expression);
+                predicate = predicateNegationExpressionOptimizer.Visit(predicate);
 
-                expression = new PredicateReductionExpressionOptimizer().Visit(expression);
+                predicate = new PredicateReductionExpressionOptimizer().Visit(predicate);
 
-                expression = new EqualityPredicateExpandingVisitor().Visit(expression);
+                predicate = new EqualityPredicateExpandingVisitor().Visit(predicate);
 
-                expression = predicateNegationExpressionOptimizer.Visit(expression);
+                predicate = predicateNegationExpressionOptimizer.Visit(predicate);
 
                 if (_useRelationalNulls)
                 {
-                    expression = new NotNullableExpression(expression);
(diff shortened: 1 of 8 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
```diff
@@ -28,29 +28,33 @@ public CompositePredicateExpressionVisitor(bool useRelationalNulls)
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public override Expression Visit(Expression expression)
+        protected override Expression VisitExtension(Expression expression)
         {
-            if (expression != null)
+            var selectExpression = expression as SelectExpression;
+
+            if (selectExpression?.Predicate != null)
```
smitpatel: This change in a way scoping down the optimizations to only predicate of select expression (applied recursively)

maumar:
This PR actually addresses https://github.com/aspnet/EntityFramework/issues/7548 - can you add test case for it? (also reflect this in the commit message)

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
maumar: +1 to what @smitpatel said. Please remove this for now, we need to look into the CompositePredicateExpressionVisitor in general, rename it and see where the optimizations can be applied to other parts of the SelectExpression. This can be done in a separate PR tho.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -277,9 +277,7 @@ var materializer
 
                     if (selectExpression.Predicate != oldPredicate)
                     {
-                        selectExpression.Predicate
-                            = compositePredicateExpressionVisitor
-                                .Visit(selectExpression.Predicate);
+                        compositePredicateExpressionVisitor.Visit(selectExpression);
```
maumar: please revert these changes for now

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -390,8 +390,7 @@ var typeIsExpressionTranslatingVisitor
 
             foreach (var selectExpression in QueriesBySource.Values)
             {
-                selectExpression.Predicate
-                    = compositePredicateVisitor.Visit(selectExpression.Predicate);
+                compositePredicateVisitor.Visit(selectExpression);
```
maumar: please revert these changes for now

tuespetre:
Would you mind if I submit a new PR for the other test case and its solution? The changes you requested here remove the optimizations seen in the test SQL output anyways, it's not really the same PR anymore.

tuespetre:
*not all of the optimizations I suppose, but any of the optimizations occurring in subqueries. I will make the changes here then.

maumar:
@tuespetre after giving is some more thought, perhaps your current solution is fine. Before we were applying it only to top level predicates, so we won't regress any optimizations, and since in your solution optimizations are still only applied to the predicates (just recursively), the name of the visitor is still adequate.

I will file an issue to track cleanup of this for later - i.e. to see if we can apply some of the optimizations in more general fashion, e.g. to projections.

Please add the extra test case in this PR and it should be good to do, sorry for randomizing you ;)

smitpatel:
`var maumar = new Random();` :trollface: 

@tuespetre - In general, there is no harm in creating new PR, but when tackling certain issue, we tend to use the same PR even if implementation changes. That is done to preserve the PR feedback comment. In case of totally change of implementation which would make the feedback on PR totally irrelevant, you should file new PR.

tuespetre:
:up::date:

smitpatel:
filed #7612

PR #7589, opened 2017-02-10
Improve the implementation of SelectExpression.HandlesQuerySource

tuespetre:
This is another change I have isolated out of #7543. When isolated like this, the main improvement it offers is avoiding 'false positives' for `IsCorrelated()` as seen in the single changed test. In the context of the rest of #7543, the correctness of `HandlesQuerySource()` gains even more significance due to a few more places it gets used.


smitpatel:
@anpete - does this isolated fix help your case?

smitpatel:
poaching this too @maumar

smitpatel:
@tuespetre - Can you elaborate reasoning behind such logic in `HandlesQuerySource`? Given a query source the method should return true, if any part of the select expression is translation of the query source in form of table or join expression. Neither should require you to look into TableExpression.QuerySource. The may be some value in unwrapping the querySource parameter if it is joinclause or processing table as JoinExpression. But needs to be tested out if there are test changes or there are scenarios (new tests) which actually needs such modification.

tuespetre:
@smitpatel there is already one test change, it is for SQL server but it would have more impact on SQLite where lateral join is not supported.

smitpatel:
Can you elaborate the logic since it seems partially incorrect based on what is the function is required to do.

tuespetre:
Which part in particular seems incorrect?

smitpatel:
`TableExpression.QuerySource as GroupJoinClause)?.JoinClause`
We shouldn't look inside of the query sources for tables. They have to be set correctly. If they are not then there is bug somewhere.

PR #7592, opened 2017-02-10
Add explicit interfaces for Shapers

tuespetre:
Another breakout from #7543, this time isolating about 9 files. 

- Added an `IShaper` interface and changed casts/tests from `Shaper` to `IShaper`
- Added an `IOffsettableShaper` interface with an `OffsettableShaper` base class and changed casts/tests from `EntityShaper` to `IOffsettableShaper`

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
@@ -239,9 +239,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 Expression.Constant(shaper));
         }
 
-        private Shaper CreateShaper(Type elementType, IEntityType entityType, SelectExpression selectExpression)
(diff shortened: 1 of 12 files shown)


tuespetre:
Hey, who got Appveyor to work? 😉

maumar:
What is the reason for this change? It doesn't seem to provide much value.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
```diff
@@ -239,9 +239,9 @@ var useQueryComposition
                 Expression.Constant(shaper));
         }
 
-        private Shaper CreateShaper(Type elementType, IEntityType entityType, SelectExpression selectExpression)
+        private IShaper CreateShaper(Type elementType, IEntityType entityType, SelectExpression selectExpression)
```
maumar: Change to interface is problematic here. We won't be able to add functionality to it without introducing a breaking change. Abstract base class doesn't have this limitation.

tuespetre:
You may be right, part of why I split these changes out. I thought I had more use for them with some other Shaper classes but that may not be valid anymore. I will double check and close this if that's the case

tuespetre:
Sorry for wasting your time with this one! 🙀

PR #7594, opened 2017-02-11
1.0.4 patch train updates

JunTaoLuo:
Updating versions for 1.0.4 patch train.


JunTaoLuo:
ping

src/Microsoft.EntityFrameworkCore/project.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.2",
+  "version": "1.0.3",
```
bricelam: Why not `1.0.2`?

cc @divega @rowanmiller @Eilon

PR #7596, opened 2017-02-11
Query: Optimize OfType operator to only query for required columns

### LINKED ISSUE #7451: Relational OfType op selects unnecessary columns
Given a hierarchy like:
```
   A
 /   \
B     C
```
the following query will select the columns from B:
```c#
context.As.OfType<C>();
```

smitpatel:
Resolves #7451 

For OfType operators of entity type remove the operator by changing the type of entityqueryable.
Move TypeIs to query model level optimization

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
@@ -10,14 +10,12 @@ using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
-using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Query.Expressions;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.ResultOperators;
-using Remotion.Linq.Parsing;
 
 namespace Microsoft.EntityFrameworkCore.Query.Internal
 {
@@ -35,7 +33,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
             public HandlerContext(
                 IResultOperatorHandler resultOperatorHandler,
                 IModel model,
-                IRelationalAnnotationProvider relationalAnnotationProvider,
                 ISqlTranslatingExpressionVisitorFactory sqlTranslatingExpressionVisitorFactory,
                 ISelectExpressionFactory selectExpressionFactory,
                 RelationalQueryModelVisitor queryModelVisitor,
@@ -47,7 +44,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                 _sqlTranslatingExpressionVisitorFactory = sqlTranslatingExpressionVisitorFactory;
 
                 Model = model;
-                RelationalAnnotationProvider = relationalAnnotationProvider;
                 SelectExpressionFactory = selectExpressionFactory;
                 QueryModelVisitor = queryModelVisitor;
                 ResultOperator = resultOperator;
@@ -56,7 +52,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
             }
 
             public IModel Model { get; }
-            public IRelationalAnnotationProvider RelationalAnnotationProvider { get; }
             public ISelectExpressionFactory SelectExpressionFactory { get; }
             public ResultOperatorBase ResultOperator { get; }
             public SelectExpression SelectExpression { get; }
@@ -97,7 +92,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                 { typeof(LastResultOperator), HandleLast },
                 { typeof(MaxResultOperator), HandleMax },
                 { typeof(MinResultOperator), HandleMin },
-                { typeof(OfTypeResultOperator), HandleOfType },
                 { typeof(SingleResultOperator), HandleSingle },
                 { typeof(SkipResultOperator), HandleSkip },
                 { typeof(SumResultOperator), HandleSum },
@@ -105,7 +99,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
             };
 
         private readonly IModel _model;
-        private readonly IRelationalAnnotationProvider _relationalAnnotationProvider;
         private readonly ISqlTranslatingExpressionVisitorFactory _sqlTranslatingExpressionVisitorFactory;
         private readonly ISelectExpressionFactory _selectExpressionFactory;
         private readonly IResultOperatorHandler _resultOperatorHandler;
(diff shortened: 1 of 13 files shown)


src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -397,22 +397,34 @@ public override void VisitQueryModel(QueryModel queryModel)
...
             }
+        }
 
-            private class RecursiveQueryModelExpressionVisitor : ExpressionVisitorBase
-            {
-                private readonly NondeterministicResultCheckingVisitor _parentVisitor;
+        /// <summary>
+        /// 
+        /// </summary>
+        protected class RecursiveQueryModelExpressionVisitor : ExpressionVisitorBase
```
smitpatel: @anpete - Any place to put this visitor? It doesn't need to be on public surface. (May somewhere in .Internal)

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1306,6 +1317,80 @@ protected virtual void WarnClientEval([NotNull] object expression)
                 () => RelationalStrings.ClientEvalWarning(expression));
         }
 
+        private class OfTypeResultOperatorOptimizer : QueryModelVisitorBase
```
anpete: Any reason not to do this in Core? If not, then QueryOptimizer would be the natural place to put this logic.

smitpatel: I guess this can be done in core too. I started in relational because extra columns/select expression is in relational. 😄

smitpatel:
New version up
- Change entityqueryable to the desired type to avoid selecting columns from sibling types
- Remove OfType of entity type is applied eagerly and removed from query model
- Lift predicate from join table when Inner Join to produce flattened query
- Convert join table to subquery in the presence of predicate in Left Outer Join
- Move TypeIsExpressionVisitor to relational OptimizeQueryModel method

src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
```diff
@@ -637,111 +635,8 @@ private static Expression HandleMax(HandlerContext handlerContext)
...
-                    && discriminatorExpression.QuerySource == _querySource)
-                {
-                    return new DiscriminatorPredicateExpression(_discriminatorPredicate, _querySource);
-                }
-
-                return expression;
-            }
-        }
+        // OfType of an entity type should be removed while optimizing query model
+        private static Expression HandleOfType(HandlerContext handlerContext) => handlerContext.EvalOnClient();
```
anpete: Can remove this and the corresponding entry in the handler map because client-eval is the default.

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -397,22 +399,97 @@ public override void VisitQueryModel(QueryModel queryModel)
...
+            => Expression.Constant(
+                _createEntityQueryableMethod
+                    .MakeGenericMethod(targetEntityType.ClrType)
+                    .Invoke(null, new object[]
+                    {
+                        entityQueryProvider
+                    }));
+
+            private static readonly MethodInfo _createEntityQueryableMethod
+                = typeof(NavigationRewritingExpressionVisitor)
```
anpete: Wrong type?

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -397,22 +399,97 @@ public override void VisitQueryModel(QueryModel queryModel)
...
+                base.VisitQueryModel(queryModel);
 
-                public RecursiveQueryModelExpressionVisitor(NondeterministicResultCheckingVisitor parentVisitor)
-                {
-                    _parentVisitor = parentVisitor;
-                }
+                var ofTypeOperators = queryModel.ResultOperators.OfType<OfTypeResultOperator>().ToList();
 
-                protected override Expression VisitSubQuery(SubQueryExpression expression)
+                foreach (var resultOperator in ofTypeOperators)
```
anpete: Can this be done in VisitResultOperator in QueryOptimizer? If so, we can DRY up some of the code below that deals with updating QSREs and annotations.

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -397,22 +399,97 @@ public override void VisitQueryModel(QueryModel queryModel)
...
+                            queryAnnotation.QuerySource = newMainFromClause;
+                            queryAnnotation.QueryModel = queryModel;
+                        }
+                    }
                 }
+
+                queryModel.TransformExpressions(new RecursiveQueryModelExpressionVisitor(this).Visit);
+            }
+
+            private ConstantExpression CreateEntityQueryable(IEntityType targetEntityType, IAsyncQueryProvider entityQueryProvider)
```
anpete: DRY this up by moving this and the nav rewrite version to a shared helper. Perhaps IAsyncQueryProviderExtensions.CreateQueryable?

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -397,22 +399,97 @@ public override void VisitQueryModel(QueryModel queryModel)
...
+                = typeof(NavigationRewritingExpressionVisitor)
+                    .GetTypeInfo().GetDeclaredMethod(nameof(_CreateEntityQueryable));
+
+            [UsedImplicitly]
+            // ReSharper disable once InconsistentNaming
+            private static EntityQueryable<TResult> _CreateEntityQueryable<TResult>(IAsyncQueryProvider entityQueryProvider)
+                => new EntityQueryable<TResult>(entityQueryProvider);
+        }
+
+        private class RecursiveQueryModelExpressionVisitor : ExpressionVisitorBase
```
anpete: If we add this, we should use it everywhere. Can be a separate PR.

smitpatel:
Updated.

src/Microsoft.EntityFrameworkCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -248,6 +253,50 @@ public override void VisitResultOperator(ResultOperatorBase resultOperator, Quer
...
+                        {
+                            queryModel.ResultOperators.RemoveAt(index);
+
+                            var newMainFromClause = new MainFromClause(
+                                oldQuerySource.ItemName,
+                                entityType.ClrType,
+                                entityQueryProvider.CreateEntityQueryable(entityType));
+                            queryModel.MainFromClause = newMainFromClause;
+
+                            var querySourceMapping = new QuerySourceMapping();
```
anpete: Can you unify any of this fix-up stuff with similar code in this class?

src/Microsoft.EntityFrameworkCore/Query/Internal/IQueryOptimizer.cs
```diff
@@ -20,6 +21,7 @@ public interface IQueryOptimizer
         /// </summary>
         void Optimize(
             [NotNull] IReadOnlyCollection<IQueryAnnotation> queryAnnotations,
+            [NotNull] IModel model,
```
anpete: Make this a ctor param.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1294,6 +1298,20 @@ public override void VisitResultOperator(ResultOperatorBase resultOperator, Quer
         }
 
         /// <summary>
+        ///     Applies optimizations to the query.
+        /// </summary>
+        /// <param name="queryModel"> The query. </param>
+        protected override void OptimizeQueryModel(QueryModel queryModel)
+        {
+            var typeIsExpressionTranslatingVisitor
```
anpete: Does this code need to move? This isn't really an optimization but rather a specific translation.

smitpatel: It causes indirect optimization in terms of materialization. We have a test with `a is Kiwi` in projection. If this is done in optimization we don't materialize. If we do during VisitQueryModel, we materialize the entity.

PR #7542, opened 2017-02-02
Upgrade to VS 2017

bricelam:
Resolves #7538

**Providers/contributors:**
1. Specification tests now target `net452` (xUnit.net dropped  `net451`)
2. `EntityFramework.sln` now requires [Visual Studio 2017 RC](https://www.visualstudio.com/vs/visual-studio-2017-rc/). For running tests, [TestDriven.NET](http://testdriven.net/) still works

cc @ErikEJ @roji @tuespetre

PR #7607, opened 2017-02-14
Bump test projects up to .NET 4.5.2

dougbu:
- aspnet/Testing#248
- xUnit no longer supports .NET 4.5.1


tools/EntityFramework.targets
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <DefineConstants>$(DefineConstants);CSPROJ</DefineConstants>
     <DefineConstants Condition=" '$(TargetFrameworkMoniker)' == '.NETFramework,Version=v4.5.1' ">$(DefineConstants);NET451</DefineConstants>
+    <DefineConstants Condition=" '$(TargetFrameworkMoniker)' == '.NETFramework,Version=v4.5.2' ">$(DefineConstants);NET452</DefineConstants>
```
dougbu: Not sure whether this will be necessary after project.json files are removed. But, for now best to cover both cases&hellip;

src/Microsoft.EntityFrameworkCore.Specification.Tests/project.json
```diff
@@ -33,7 +33,7 @@
         "Microsoft.CSharp": "4.4.0-*"
       }
     },
-    "net451": {
+    "net452": {
```
dougbu: @natemcmaster bit of a concern that I needed to change this project. Without the change, CI runs failed with missing symbols in test projects because the dependencies listed below weren't brought in. Is that a bug in transitive references?

With this change, I will also need to change other `src` projects that reference this one.

natemcmaster: Might be. We should try this on csproj and see if it works better there (#7542)

dougbu:
@bricelam the CI failures on Windows seem to be fairly random -- SQL timeouts. But the failures on OSX and Linux are real and potentially a blocker. Microsoft.EntityFrameworkCore.InMemory.FunctionalTests and two other projects fail to compile because these platforms have no System.Xml.dll reference assemblies. I've tried a couple of things and have not improved matters.

Options:
1. Forget this PR and address aspnet/Testing#248 as part of or just after #7542. This works because none of the test projects will attempt to build for desktop .NET except on Windows.
2. Remove the net451 target framework from the test/Microsoft.EntityFrameworkCore.InMemory.FunctionalTests/project.json file and do the same for the other two projects and all test projects that depend on them. This unfortunately would affect most of the test projects on Windows too. This would not be 🆗 unless we test using the .csproj test projects on Windows.
3. Work around the issue some other way. Since these projects build when targeting .NET 4.5.1 on these machines and .NET 4.5.2 should be treated as a superset, I'm not sure exactly what's going on.

Thoughts?

bricelam:
I opt for number one. I'll update that PR, and investigate any failures.

dougbu:
Thanks @bricelam 👍

Issue #2850, opened 2015-08-15
Query: Provide DbFunctions class to define methods that correspond to standard SQL operations


roji:
Not sure whether this has already been discussed (or makes sense).

There are some standard SQL operations that don't correspond to an already-existing .NET method; the example that comes to mind is LIKE. It seems that up to now, each provider invented its own way to allow users to specify these in LINQ to Entities - SqlClient has [SqlFunctions](https://msdn.microsoft.com/en-us/library/system.data.objects.sqlclient.sqlfunctions%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) and [SqlMethods](https://msdn.microsoft.com/en-us/library/system.data.linq.sqlclient.sqlmethods%28v=vs.110%29.aspx).

It seems to makes sense to pull the standard SQL functions and methods out of these SQL Server-specific classes and place them in the generic EF7 code, this way the same LINQ can be used across different providers...

ErikEJ:
I think it is already there: https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Relational/Query/Methods/ContainsTranslator.cs

divega:
@roji This does make sense in general although it is probably worth digging more into what "standard SQL operations" means :smile:

We have talked within the EF team about the need to have a way to represent any standard relational functions as well as provider specific functions that aren't well represented by LINQ operators or by standard method calls in .NET BCL classes. We came up with a pattern we decided to try in EF7 that we believe can be better than what we had in the past. We just haven't found an opportunity to start using it, but would look similar to this:

``` C# 
var legalMinors = context.People.Where(p => 
    EF.Functions.DiffYear(p.BirthDate, DateTime.Now) - 
    DateTime.Now.DayOfYear < p.BirthDate.DayOfYear ? 1 : 0 
    < legalAge);
```

The idea is to add a new `Functions` property to the static [EF class](https://github.com/aspnet/EntityFramework/blob/dev/src/EntityFramework.Core/EF.cs) (the same class where we put the `Property` method you can use to access shadow state properties in LINQ). Then EF and providers as well can define extension methods for the type of the `Functions` property. 

Here are a few (subject to change) rules for these:

1. In general, EF and providers should use a high bar for adding methods here. If there is a reasonable pattern that can be represented with LINQ operators or standard .NET BCL method calls we should enable the translation of that pattern instead.

[...]

PR #7611, opened 2017-02-14
Add EF.Functions and SQL Like

roji:
To allow relational and providers to start defining extension methods on it, for #2850.

Hope this is more or less what you guys were thinking of, let me know of any nitpicks in terms of class/file naming etc.

Note that no actual methods (e.g. Like) are defined, that can come later in a separate PR.

### DIFF
--- a/src/EFCore.Relational.Specification.Tests/RelationalQueryTestBase.cs
+++ b/src/EFCore.Relational.Specification.Tests/RelationalQueryTestBase.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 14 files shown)


roji:
Have just amended this PR. Aside from rebasing for VS2017 support, this PR now adds SQL LIKE support via `EF.Functions.Like()`, providing overloads with and without an escape character.

Note that the SqlTranslatingExpressionVisitor had to be tweaked for it to allow translating method calls on the `EFFunctions` object.

anpete:
@roji Would you mind rebasing again? We renamed the world 😬

roji:
@anpete rebased, also renamed EFFunctions -> DbFunctions (looks slightly better). Let me know if you want any changes

anpete:
@roji I'm seeing some test failures in ApiConsistencyTest, can you take a look? BTW, sorry for the delay 😄

roji:
Fixed and rebased.

PR #7619, opened 2017-02-15
Query: Fix bug in BindPropertyToOuterParameter

tuespetre:
This PR makes the following changes:

- Makes `QueriesBySource` a private API to `RelationalQueryModelVisitor`
- Makes `Queries` an `IEnumerable<SelectExpression>` instead of `ICollection<SelectExpression>`
- Renames `AddQuery` to `MapQuery` to better align with what the method is doing

This is, again, a set of changes I am splitting out from #7543. The idea is that as the query model is being visited, each query source gets mapped to a `SelectExpression` so that during binding (either within `SqlTranslatingExpressionVisitor` or `MemberAccessBindingExpressionVisitor` or etc.) there will be optimal opportunity to find or add the appropriate projection expressions.

As you can see this change already has benefits outside of the context of #7543 in that it has optimized some types of N+1 queries that would otherwise pull much more data on each of the +1 queries.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -1608,8 +1608,7 @@ namespace Microsoft.EntityFrameworkCore.Query
         {
             if (querySource != null && _canBindPropertyToOuterParameter)
             {
-                SelectExpression parentSelectExpression = null;
-                ParentQueryModelVisitor?.QueriesBySource.TryGetValue(querySource, out parentSelectExpression);
+                var parentSelectExpression = ParentQueryModelVisitor?.TryGetQuery(querySource);
                 if (parentSelectExpression != null)
                 {
                     var parameterName = OuterQueryParameterNamePrefix + property.Name;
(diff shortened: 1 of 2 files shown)


smitpatel:
Poaching this @maumar

smitpatel:
The changed test SQL is an improvement here, those tests should be using parameter injection but instead they are not because of bug here
https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs#L1633

Instead of trying to find SelectExpression by asking QueriesBySource, it should use `TryGetQuery` method which utilizes `HandlesQuerySource` method on SelectExpression.

Ideally, a querymodelvisitor (qmv) should store only SelectExpression(s), which are required to be run to generate results of query model being visited. There can be more than one SelectExpression stored it they are being stitched together. That is the reason when we lift the join clause into main select expression we remove the entry for join clause in QueriesBySource. If a query source is being lifted then there should be no need to store it in QueriesBySource, instead SelectExpression.HandlesQuerySource method should tell if particular qs is part of SelectExpression.

This change brought test SQL changes because of existing bug. Instead the underlying bug should be fixed. Leaving that aside, this does not add any value rather increases the metadata size.

tuespetre:
> instead SelectExpression.HandlesQuerySource method should tell if particular qs is part of SelectExpression

WOW! 😱 Big 'duh' moment for me, now I am seeing all of the places where a query source is not being mapped to an added join expression or subquery select expression, and how I was using these changes to compensate for that.

tuespetre:
:up::date:

smitpatel:
Merged via 2a8a3b5bd5fd39337038dd952acadee93eefa3f2
Thank you for contribution.

PR #7634, opened 2017-02-16
Add Pluralization Hook to Reverse Engineer

### LINKED ISSUE #3060: Reverse Engineering: Enable pluralization/singularization when scaffolding a model from a database
Database scaffolding has no pluralization service and the entity models are generated mapping the exact name of database objects. My suggestion is to enable this pluralization through a DI service that could be overridden at the `ef` CLI like this:

```
dnx ef dbcontext scaffold "Server=myserver;Database=Foo;user id=me;password=pwd" EntityFramework.SqlServer --use-pluralization "MyAssembly.PlurarizationService, MyAssembly"
```

And on Razor templates this could be exposed as:

``` csharp
...
@:namespace @Model.Namespace
@:{
@:    public class  @PluralizationService.Singularize(Model.EntityType.Name)
@:    {
...
```

I think the default service could be the one shipped with EF6 (`System.Data.Entity.Design.PluralizationServices.PluralizationService`)

rowanmiller:
Introduces a new IPluralizer service that is used to singularize entity type names and pluralize DbSet names. The default implementation is a no-op, so this is just a hook where folks can easily plug in their own pluralizer.

I left this separate from the existing CandidateNamingService, which is responsible for cleaning up the table name into a C# friendly identifier.

Here is what it looks like for a developer to hook in their own pluralizer (this example uses Inflector). Our code generation coverage is lacking, so I tested this against Northwind and verified by round tripping data.

```c#
public class MyDesignTimeServices : IDesignTimeServices
{
    public void ConfigureDesignTimeServices(IServiceCollection services)
    {
        services.AddSingleton<IPluralizer, MyPluralizer>();
    }
}

public class MyPluralizer : IPluralizer
{
    public string Pluralize(string name)
    {
        return Inflector.Inflector.Pluralize(name) ?? name;
    }

    public string Singularize(string name)
    {
        return Inflector.Inflector.Singularize(name) ?? name;
    }
}
```

Resolves #3060

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.Relational.Design.Tests/RelationalScaffoldingModelFactoryTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.Relational.Design.Tests/RelationalScaffoldingModelFactoryTest.cs
@@ -15,6 +15,7 @@ using Microsoft.EntityFrameworkCore.Scaffolding.Metadata;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.Extensions.Logging;
 using Xunit;
+using Microsoft.EntityFrameworkCore.Infrastructure;
 
 namespace Microsoft.EntityFrameworkCore.Relational.Design
 {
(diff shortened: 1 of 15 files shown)


rowanmiller:
Adding `providers-beware` tag as providers who derive from `RelationalScaffoldingModelFactory` will need to add the new constructor parameter (see changes to `SqlServerScaffoldingModelFactory` and `SqliteScaffoldingModelFactory` in this PR).

test/Microsoft.EntityFrameworkCore.Relational.Design.Tests/RelationalScaffoldingModelFactoryTest.cs
```diff
@@ -828,4 +893,21 @@ public class TestTypeMapper : RelationalTypeMapper
...
+                : name + "s";
+        }
+
+        public string Singularize(string name)
+        {
+            return name.EndsWith("s")
+                 ? name.Substring(0, name.Length - 1)
+                 : name;
+        }
+    }
```
bricelam: lol, let's ship it. :trollface:

bricelam:
(I edited your example to use `IDesignTimeServices`. We should recommend this over a class named `Startup`)

test/Microsoft.EntityFrameworkCore.Relational.Design.Tests/RelationalScaffoldingModelFactoryTest.cs
rowanmiller: It may be better than the EF6 one 😉

rowanmiller:
Nice, I didn't know about `IDesignTimeServices`... I'll update a recent blog post I did too 😄

jwooley:
@rowanmiller It looks like the sample from your blog doesn't handle child navigation property pluralization for ICollection<T> items. It does appear to work for the root levels in the context itself.

rowanmiller:
@jwooley good point, I'll put together a change to pluralize collections

rowanmiller:
Pluralization of collection navs addressed in https://github.com/aspnet/EntityFramework/commit/6841330be6bf3f7d37d252617b448806ed37e27e

PR #7638, opened 2017-02-16
Add parameter objects for dependencies of public services

ajcvickers:
Issue #7465

This change introduces parameter objects for the service dependencies of all public services. This allows new dependencies to be added without breaking derived classes that call the base class constructor. For factories, these dependency objects are passed down into the services created by the factory. This means that we also don't break by adding parameters to the constructors called by the factory. This is the main reason for using this approach rather than just the service rewriting approach.

The pattern is not used for internal services since no code should be inheriting from these classes. However, the service re-writer code is still in place, which means that we have some flexibility to add dependencies to internal services even in patch releases where we are being very strict about binary compatibility.

Some work remains still to do:
- Add With methods to all ...Dependencies objects to allow for service replacement
- Add sugar for registration
- Re-visit service scopes


ErikEJ:
OMG ! 😄

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/ServiceCollectionRelationalProviderInfrastructure.cs
```diff
@@ -84,7 +86,30 @@ public static void TryAddDefaultRelationalServices([NotNull] ServiceCollectionMa
...
+                .TryAddScoped<SqlTranslatingExpressionVisitorDependencies, SqlTranslatingExpressionVisitorDependencies>()
+                .TryAddScoped<RelationalProjectionExpressionVisitorDependencies, RelationalProjectionExpressionVisitorDependencies>()
+                .TryAddScoped<ParameterNameGeneratorDependencies, ParameterNameGeneratorDependencies>()
+                .TryAddScoped<RelationalQueryModelVisitorDependencies, RelationalQueryModelVisitorDependencies>()
+                .TryAddScoped<RelationalEntityQueryableExpressionVisitorDependencies, RelationalEntityQueryableExpressionVisitorDependencies>()
+                .TryAddScoped<RelationalConnectionDependencies, RelationalConnectionDependencies>()
+                .TryAddScoped<RelationalDatabaseDependencies, RelationalDatabaseDependencies>()
+                .TryAddScoped<SelectExpressionDependencies, SelectExpressionDependencies>()
+                .TryAddScoped<RelationalValueBufferFactoryDependencies, RelationalValueBufferFactoryDependencies>()
+                .TryAddScoped<RelationalQueryCompilationContextDependencies, RelationalQueryCompilationContextDependencies>();
```
AndriySvyryd: Group with the dependent service?

src/Microsoft.EntityFrameworkCore/Storage/Database.cs
```diff
@@ -24,17 +24,17 @@ namespace Microsoft.EntityFrameworkCore.Storage
     /// </summary>
     public abstract class Database : IDatabase
     {
-        private readonly IQueryCompilationContextFactory _queryCompilationContextFactory;
+        private readonly DatabaseDependencies _dependencies;
```
ajcvickers: It should be--will fix.

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/ServiceCollectionRelationalProviderInfrastructure.cs
ajcvickers: I'm going to revisit this and possibly add some specific API for registering these. Only our code should be doing it--providers should never need to.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/HistoryRepository.cs
```diff
@@ -17,26 +17,39 @@
...
+    /// <summary>
+    ///     <para>
+    ///         Service dependencies parameter class for <see cref="HistoryRepository" />
+    ///     </para>
+    ///     <para>
+    ///         This type is typically used by database providers (and other extensions). It is generally
+    ///         not used in application code.
+    ///     </para>
+    /// </summary>
+    public sealed class HistoryRepositoryDependencies
```
AndriySvyryd: Move to a separate file

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/RelationalModelValidatorDependencies.cs
```diff
@@ -0,0 +1,58 @@
...
+        ///     <para>
+        ///         Creates the service dependencies parameter object for a <see cref="RelationalModelValidator" />.
+        ///     </para>
+        ///     <para>
+        ///         This type is typically used by database providers (and other extensions). It is generally
+        ///         not used in application code.
+        ///     </para>
+        ///     <para>
+        ///         Do not call this constructor directly from provider or application code as it may change
+        ///         as new dependencies are added. Use the 'With...' methods instead.
```
ajcvickers: How about "Do not call this constructor directly from provider or application code as it may change as new dependencies are added. Instead, use this type in your constructor so that an instance will be created and injected automatically by the dependency injection container. To create an instance with some dependent services replaced, first resolve the object from the dependency injection container, then replace selected services using the 'With...' methods. Do not call the constructor at any point in this process."

src/Microsoft.EntityFrameworkCore.Relational/Migrations/HistoryRepository.cs
ajcvickers: Will do.

src/Microsoft.EntityFrameworkCore.Relational/Storage/RelationalDatabase.cs
```diff
@@ -22,31 +21,21 @@ namespace Microsoft.EntityFrameworkCore.Storage
     /// </summary>
     public class RelationalDatabase : Database
     {
-        private readonly ICommandBatchPreparer _batchPreparer;
-        private readonly IBatchExecutor _batchExecutor;
-        private readonly IRelationalConnection _connection;
+        private readonly RelationalDatabaseDependencies _relationalDependencies;
```
AndriySvyryd: Property

AndriySvyryd:
:shipit:

Issue #7601, opened 2017-02-12
Missing translators for Math functions in SqlServer provider


rpawlaszek:
There is quite a number of [math functions not translated](https://github.com/aspnet/EntityFramework/tree/c9b5916a4e0fff763febea00da793ebd4113632d/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal): ```Math.Exp, Math.Log10, Math.Log, Math.Sqrt, Math.Acos, Math.Asin, Math.Atan2, Math.Atan, Math.Cos, Math.Sin, Math.Tan```. The implementation is as easy as for the ones translated that derive from [```MultipleOverloadStaticMethodCallTranslator```](https://github.com/aspnet/EntityFramework/blob/c9b5916a4e0fff763febea00da793ebd4113632d/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionTranslators/MultipleOverloadStaticMethodCallTranslator.cs). Could I ask to implement them? I implemented them locally and they all work fine.

### Further technical details
EF Core version: all
Database Provider: Microsoft.EntityFrameworkCore.SqlServer
Operating system: Windows 10 Pro
IDE: Visual Studio 2015

rpawlaszek:
@rowanmiller what does the label "up-for-grabs" mean?

rowanmiller:
@rpawlaszek we use it to identify items that would be good candidates for a contribution from the community. It doesn't mean we're not going to do it, just that someone from the community may want to pick it up before we get to it.

rpawlaszek:
Great:) As I understand this includes implementing the functionality AND the tests, right? And if I would like to do it is [this](https://github.com/aspnet/Home/blob/dev/CONTRIBUTING.md#contributing-code-and-content) the procedure? Sorry if it is obvious, I just haven't done this ever before.

rowanmiller:
@rpawlaszek correct. It is also best to start with a small contribution for your first one. So if you plan to tackle this, probably start with one method until you get the hang of how the process works and the type of feedback you get on the code review etc.

rpawlaszek:
[...]

PR #7639, opened 2017-02-16
SqlServerMathExpTranslator implementation proposition (part of #7601)

rpawlaszek:
Summary of the changes
  - Added [SqlServerMathExpTranslator](https://github.com/rpawlaszek/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerMathExpTranslator.cs)
  - Included it in [_methodCallTranslators](https://github.com/rpawlaszek/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerCompositeMethodCallTranslator.cs#L23) (in SqlServerCompositeMethodCallTranslator)
  - Added [Where_math_exp](https://github.com/rpawlaszek/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs#L4825) test in QueryTestBase
  - Added [Where_math_exp](https://github.com/rpawlaszek/EntityFramework/blob/dev/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs#L4968) test to QuerySqlServerTest tests.

Fixes (partially, as there are more ```Math``` methods to be implemented) #7601 

Just want to try with this small piece of code to check whether it meets the requirements. If so - I can implement the rest following this pattern.


rpawlaszek:
I see that the structure of all the csprojs was changed in [dotnet migrate commit](https://github.com/aspnet/EntityFramework/commit/ff2fecdfaca0b89e856eca8426306d506389564b#diff-39cd3af016a43c9165055139843a04f2).

rpawlaszek:
Ok, I've found [how to proceed](https://help.github.com/articles/syncing-a-fork/).

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -4822,6 +4822,14 @@ public virtual void Where_math_ceiling2()
         }
 
         [ConditionalFact]
+        public virtual void Where_math_exp()
+        {
+            AssertQuery<OrderDetail>(
+                ods => ods.Where(od => Math.Exp(od.Discount) > 1),
```
smitpatel: Can you change this to `ods => ods.Where(od => od.OrderID == 11077).Where(Math.Exp(od.Discount) > 1)`
That will make the size of result set smaller and reduce test execution time.

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerMathExpTranslator.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class SqlServerMathExpTranslator : MultipleOverloadStaticMethodCallTranslator
```
smitpatel: This should actually derive from `SingleOverloadStaticMethodCallTranslator` just like `SqlServerMathPowerTranslator`.
`Math.Exp` got only 1 method and no other overload.

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -4822,6 +4822,14 @@ public virtual void Where_math_ceiling2()
...
+                entryCount: 838);
+        }
```
smitpatel: Also add same test in `AsyncQueryTestBase` too.

src/Microsoft.EntityFrameworkCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerMathExpTranslator.cs
rpawlaszek: Thank you @smitpatel for the comments. I will add/change the parts you pointed out.

src/Microsoft.EntityFrameworkCore.Specification.Tests/QueryTestBase.cs
anpete: Not sure this matters too much as the whole query is being translated.

roji:
Just a thought... With the many Math methods, it may be worth concentrating all the translators in one MathTranslator rather than creating a class per method? Just for maintainability

maumar:
@roji I was thinking the same thing

rpawlaszek:
Ok, I installed VS2017, updated the repo to aspnet/EntityFramework and applied the changes requested by @smitpatel except for the one to implement a test in ```AsyncQueryTestBase```. As I see there is no test for other ```Math``` related methods there and the class mainly tests actual async functions. Should it be still added there? 

Besides, looking at #7651 even the ones I pushed now seem already obsolete.

smitpatel:
@rpawlaszek - After #7651 (I m running final test before check-in), it would simplify the source code changes for this function and other functions you like to add. Sorry, for bit of extra work, the refactoring came up in this PR only.

As for tests, ideally tests in QueryTestBase should appear in AsyncQueryTestBase too since it uses the async pipeline of query. While translation may be correct but we still should be checking if it works E2E. Though, you are right on missing tests. AsyncQueryTestBase hasn't been up-to-date and lacking various tests. Just find a suitable place in that file and add test you write for new functions. (no need to assert SQL for those)

smitpatel:
@rpawlaszek - I have checked in the aggregation. Just rebase on latest dev. The file you need to edit now is `SqlServerMathTranslator`.
Thanks.

rpawlaszek:
@maumar Thank you. Now I'm implementing other ```Math``` methods (with tests).

maumar:
@rpawlaszek please rebase-squash on the current dev and I will push the PR

rpawlaszek:
@maumar I've squashed all my commits, but eventually there was a difference between my dev's and aspnet/EntityFramework's dev in ```SqlServerMathTranslator``` shown here under "Resolve conflicts". I merged them in my repo but from the [last commit](https://github.com/rpawlaszek/EntityFramework/commit/bba22505172543faf0182edf4c66f7d3bb0d0f83) I see there is no change at all. 

It was the first time I did rebase-squash, so please check if I did that correctly. I used [this](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history) as my guide.

Finally, the ```SqlServerMathTranslator``` prepared by @smitpatel  is now updated with all the functions from #7601, tests are prepared and they all pass.

maumar:
@rpawlaszek new changes look good. However the commit message got lost during the merge. Ideally you would want to rebase your branch on top of our dev, rather than merging dev into your branch. I like using tortoisegit for that purpose, which allows you to do squash rebase very easily and also got some nice tools for diffing and conflict resolution.

You can update the last commit message using:
git commit --amend

This is just FYI, as I have already done this and pushed the change (squash-rebase would reassign me as the PR creator, that's why you had to do it yourself). Thank you for the contribution!

rpawlaszek:
@maumar @smitpatel, thank you for the guidance and patience. I hope this update will be useful.

PR #7643, opened 2017-02-17
Skip flaky tests

pranavkm:
Unblocking patch build


bricelam:
Might hit more after this assembly. Is there a way to disable all `*.SqlServer.*` test assemblies from running?

PR #7646, opened 2017-02-17
Implement VisitChildren for InExpression and ExistsExpression

tuespetre:
- Introduce CompositeExpression to explicitly represent a set of various expressions

-----

More changes being split out of #7543. The immediate benefits are that further optimizations are applied to subqueries within `InExpression` and `ExistExpression`, and instead of using various odd means of representing a small collection of `Expression`s (such as `IReadOnlyCollection<Expression>` and `ConstantExpression` with an `Expression[]` value) we have an explicit expression type to represent that concept. 

In the longer term, `CompositeExpression` will gain a `Flatten` method which allows for more opportunities like complex/nested grouping keys and easier processing of complex/nested projections.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/ExistsExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/ExistsExpression.cs
@@ -72,6 +72,13 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
         ///     children, and if any of them change, should return a new copy of
         ///     itself with the modified children.
         /// </remarks>
-        protected override Expression VisitChildren(ExpressionVisitor visitor) => this;
+        protected override Expression VisitChildren(ExpressionVisitor visitor)
+        {
+            var expression = visitor.Visit(Expression);
+
+            return expression != Expression
+                ? new ExistsExpression(expression)
+                : this;
+        }
     }
 }
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/InExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/InExpression.cs
@@ -47,6 +47,16 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             SubQuery = subQuery;
         }
 
+        private InExpression(
+            Expression operand,
+            IReadOnlyList<Expression> values,
+            SelectExpression subQuery)
+        {
+            Operand = operand;
+            Values = values;
+            SubQuery = subQuery;
+        }
+
         /// <summary>
         ///     Gets the operand.
         /// </summary>
@@ -110,7 +120,35 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
         ///     children, and if any of them change, should return a new copy of
         ///     itself with the modified children.
         /// </remarks>
-        protected override Expression VisitChildren(ExpressionVisitor visitor) => this;
+        ///         
+        protected override Expression VisitChildren(ExpressionVisitor visitor)
+        {
+            var newOperand = visitor.Visit(Operand);
+            var newSubQuery = (SelectExpression)visitor.Visit(SubQuery);
+
+            var valuesChanged = false;
+            var newValues = new List<Expression>();
+            if (Values != null)
+            {
+                foreach (var value in Values)
+                {
+                    var newValue = visitor.Visit(value);
+                    if (newValue is BlockExpression && value is ListInitExpression)
+                    {
+                        newValues.Add(value);
+                    }
+                    else
+                    {
+                        newValues.Add(newValue);
+                        valuesChanged |= newValue != value;
+                    }
+                }
+            }
+
+            return valuesChanged || newOperand != Operand || newSubQuery != SubQuery
+                ? new InExpression(newOperand, newValues.AsReadOnly(), newSubQuery)
+                : this;
+        }
 
         /// <summary>
         ///     Creates a <see cref="String" /> representation of the Expression.
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
@@ -1963,7 +1963,7 @@ ORDER BY [t0].[SquadId]",
         LEFT JOIN (
             SELECT [t.Gear].[Nickname], [t.Gear].[SquadId], [t.Gear].[AssignedCityName], [t.Gear].[CityOrBirthName], [t.Gear].[Discriminator], [t.Gear].[FullName], [t.Gear].[HasSoulPatch], [t.Gear].[LeaderNickname], [t.Gear].[LeaderSquadId], [t.Gear].[Rank]
             FROM [Gear] AS [t.Gear]
-            WHERE ([t.Gear].[Discriminator] = N'Officer') OR ([t.Gear].[Discriminator] = N'Gear')
+            WHERE [t.Gear].[Discriminator] IN (N'Officer', N'Gear')
         ) AS [t0] ON ([t].[GearNickName] = [t0].[Nickname]) AND ([t].[GearSquadId] = [t0].[SquadId])
         WHERE (([t].[Note] <> N'K.I.A.') OR [t].[Note] IS NULL) AND ([t0].[HasSoulPatch] = 0))
     THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT)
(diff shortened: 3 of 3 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/ExistsExpression.cs
```diff
@@ -72,6 +72,16 @@ protected override Expression Accept(ExpressionVisitor visitor)
         ///     children, and if any of them change, should return a new copy of
         ///     itself with the modified children.
         /// </remarks>
-        protected override Expression VisitChildren(ExpressionVisitor visitor) => this;
+        protected override Expression VisitChildren(ExpressionVisitor visitor)
+        {
+            var expression = visitor.Visit(Expression);
+
+            if (expression != Expression)
```
smitpatel: Use `?:` syntax.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
```diff
@@ -1963,7 +1963,7 @@ SELECT 1
         LEFT JOIN (
             SELECT [t.Gear].[Nickname], [t.Gear].[SquadId], [t.Gear].[AssignedCityName], [t.Gear].[CityOrBirthName], [t.Gear].[Discriminator], [t.Gear].[FullName], [t.Gear].[HasSoulPatch], [t.Gear].[LeaderNickname], [t.Gear].[LeaderSquadId], [t.Gear].[Rank]
             FROM [Gear] AS [t.Gear]
-            WHERE ([t.Gear].[Discriminator] = N'Officer') OR ([t.Gear].[Discriminator] = N'Gear')
```
smitpatel: Given that optimizations in `CompositePredicateVisitorExpression` are running on all where predicates appearing anywhere in the select expression, these changes should be captured by it. At present it is not happening due to incorrect `VisitChildren` method in `ExistExpression` (notice that all these changes are happening when there is `Exists` in query). So these tests would be improved with just `ExistExpression` changes. What is the benefit of all other source changes here? Nothing is visible in tests.

tuespetre: I suppose you are correct here. I was going off of another comment from @maumar about splitting out the `VisitChildren` implementation of `InExpression` into its own PR.

tuespetre:
> What is justification for adding an Expression just for the Expression[]. While it may seem to simplify things in InExpression.VisitChildren, the same method can be written to take care of all the different kinds of properties it can have.

As I said above:

> In the longer term, CompositeExpression will gain a Flatten method which allows for more opportunities like complex/nested grouping keys and easier processing of complex/nested projections.

Its advantage becomes clearer in the context of other changes I still have pending.

smitpatel:
As happened earlier, add the custom Expression when you actually need it. 😄

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
```diff
@@ -1986,7 +1986,7 @@ public override void Optional_navigation_type_compensation_works_with_contains()
 WHERE (([t].[Note] <> N'K.I.A.') OR [t].[Note] IS NULL) AND [t0].[SquadId] IN (
     SELECT [g].[SquadId]
     FROM [Gear] AS [g]
-    WHERE ([g].[Discriminator] = N'Officer') OR ([g].[Discriminator] = N'Gear')
+    WHERE [g].[Discriminator] IN (N'Officer', N'Gear')
```
tuespetre: @smitpatel this change is actually due to the `InExpression`.

smitpatel: Not really. Just implementing `VisitChildren` in `ExistExpression` gives above change. I tested on latest dev.

tuespetre: For the specific test `Optional_navigation_type_compensation_works_with_contains()`?

maumar: Irrespective, I think there is value in implementing VisitChildren on InExpression. However, I do agree with @smitpatel that there is not enough value to justify creating CompositeExpression node in this PR.

Supposedly, the other PR in which the new node provides real value doesn't get accepted (due to some fundamental issue, or we decide it's too risky or there is better solution, etc) we would then end up with extra Expression that doesn't really do us much good.

smitpatel: Sorry, my mistake. I did not see full SQL for this test. Though this should be fixed easily by writing correct `VisitChildren` for `InExpression` without bringing in any additional expression.

tuespetre: Question: any thoughts on how to handle the `ListInitExpression` then? The reason I opted to use `CompositeExpression` here is because then instead of having three different properties hanging off of the `InExpression`, there could be a single `Expression Values { get; }` property that could be pattern matched.

maumar: @tuespetre 

```cs
        private InExpression(
            Expression operand,
            IReadOnlyList<Expression> values,
            SelectExpression subQuery)
        {
            Operand = operand;
            Values = values;
            SubQuery = subQuery;
        }
```

```cs
        protected override Expression VisitChildren(ExpressionVisitor visitor)
        {
            var newOperand = visitor.Visit(Operand);
            var newSubQuery = (SelectExpression)visitor.Visit(SubQuery);

            var valuesChanged = false;
            var newValues = new List<Expression>();
            if (Values != null)
            {
                foreach (var value in Values)
                {
                    var newValue = visitor.Visit(value);
                    if (newValue is BlockExpression && value is ListInitExpression)
                    {
                        newValues.Add(value);
                    }
                    else
                    {
                        newValues.Add(newValue);
                        valuesChanged |= newValue != value;
                    }
                }
            }

            return valuesChanged || newOperand != Operand || newSubQuery != SubQuery
                ? new InExpression(newOperand, newValues.AsReadOnly(), newSubQuery)
                : this;
        }
```

maumar: just tested that implementation and all tests seem to be passing

tuespetre: Alright, I will tweak that when I have time this weekend.

Having the ListInitExpression be nested inside of a list of expressions doesn't feel weird to you though?

tuespetre:
:up::date:

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
maumar: It does feel a bit hacky, but the code churn is not worth it. We might as well reconsider this once/if CompositeExpression is justified by a future change.

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/ExistsExpression.cs
maumar: please switch this to:
```
return expression != Expression
 ? new ExistsExpression(expression)
 : this;
```

We tend to use this pattern whenever we can.

tuespetre: Oh, shoot, I missed this one

tuespetre:
:up::date:

maumar:
@tuespetre looks good! Can you rebase on the latest dev?

tuespetre:
👍  and 🏁

maumar:
Thank you for the contribution!

tuespetre:
Thank you two for the patience having to deal with me :wink:

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/InExpression.cs
```diff
@@ -110,7 +120,35 @@ protected override Expression Accept(ExpressionVisitor visitor)
...
+                    else
+                    {
+                        newValues.Add(newValue);
+                        valuesChanged |= newValue != value;
+                    }
+                }
+            }
+
+            return valuesChanged || newOperand != Operand || newSubQuery != SubQuery
+                ? new InExpression(newOperand, newValues.AsReadOnly(), newSubQuery)
```
smitpatel: This line is incorrect. If we are having InExpression with subquery (where Values is null), and if the operand or subquery changes after visiting then we will create a new InExpression with Empty List as Values instead of null. That makes InExpression to produce `0=1` hence incorrect SQL.

tuespetre: 👍 indeed

PR #7650, opened 2017-02-17
Optimize query sources requiring materialization

tuespetre:
This is the big one, split out from #7543. There a couple/few other open PRs that, if merged, could somewhat reduce the size of this one, but other than that, I have tried multiple times to break this one down further with no success.


smitpatel:
@tuespetre - Can you mention other PRs which are included in this PR so we can review them first which makes it easier to review this one.

Also, given description of the PR, this should optimize the materialization in group join only. If there are any other optimizations which are not related to join/groupjoin then please split them out. There is no reason for them to be inter-dependent. They can certainly improve the groupjoin queries but the mail crux of the issue in groupjoin is forced materialization of both the sides and in this scope of PR, it should be removing that part and apply good heuristic to determine what to materialize and how to bind.

tuespetre:
@smitpatel 

Here are the other PRs I would be waiting on that this can be rebased onto:

- #7589 
- #7619 
- #7646 

As for your request:

> If there are any other optimizations which are not related to join/groupjoin then please split them out. There is no reason for them to be inter-dependent.

There is *plenty* of reason for the remaining changes to be interdependent, because changing the heuristics of which query sources require materialization ultimately affects a lot of stuff -- that's what remains in this PR. It starts with the RequiresMaterializationExpressionVisitor, which I very tediously refactored and debugged to properly account for all types of query sources. From there, through debugging and fixing the tests that broke as a result, the other changes resulted. 

What remains in this PR is what you asked for just above: "apply good heuristic to determine what to materialize and how to bind."

tuespetre:
@smitpatel what does 'blocked' mean in this case?

tuespetre:
@smitpatel @maumar @anpete @rowanmiller 

I'm feeling worried by the 'blocked' tag sans explanation. If I need to stop trying here I would like to know now rather than later.

maumar:
@tuespetre I think @smitpatel means that we should review the other pull requests that you mentioned before tackling this one. Basically to make this as simple as possible.

tuespetre:
OK, sorry, it just looked 👻 scary to me

maumar:
No worries you are not the first person who got spooked by some of our tags, and surely not the last :)

tuespetre:
Rebased + cleaned some 'churn' out of this commit, including removing `CompositeExpression`

tuespetre:
Good news @smitpatel and @maumar... I think I did find another thing I can split out (the `OuterPropertyExpression` and its related changes.) Looks like it will have some optimizations, assuming I can successfully split it out easily.

Issue #7647, opened 2017-02-17
Query/Cleanup : remove redundant compilations of subquery models when processing MainFromClause, AdditionalFromClause and Join/GroupJoin clauses


maumar:
When processing MainFromClause, AdditionalFromClause, JoinClause inner sequence and GroupJoin inner sequence, we first compile the expression and then check if its a subquery (and if so, compile again).

We should revert the order, first try to compile as subquery, and only if that fails compile the second time. This will simplify the query compilation as well as fix some more instances of issue described in https://github.com/aspnet/EntityFramework/issues/7476

tuespetre:
I have addressed this with changes in #7543. It's part of the whole 'impossible to split further' thing.

maumar:
This would be a great candidate to break off that big PR - it should be just few lines of code (call LiftSubquery before CompileMainFromClauseExpression/CompileAdditionalFromClauseExpression/CompileJoinClauseInnerSequenceExpression/CompileGroupJoinInnerSequenceExpression with some extra checks) and it provides decent value on its own.

maumar:
fixed in 49112ba68e5352a8c7d1c1f4ba6cb73176ef7e08

PR #7655, opened 2017-02-18
Fix to #7647 - Query/Cleanup : remove redundant compilations of subquery models when processing MainFromClause, AdditionalFromClause and Join/GroupJoin clauses

maumar:
This is small optimization for some queries containing subqueries. Before we used to compile them twice, throwing away the first one. This makes it harder to debug code, but also prevents translating GroupJoin queries inside subqueries.

Fix is to reverse the order in which we process those subqueries, this way we only process the subquery twice if the first pass fails (i.e. we can't translate it fully), rather than always.


maumar:
@smitpatel @anpete

smitpatel:
Hope you had good time updating all those SQL :trollface:

PR #7663, opened 2017-02-21
Refactor SqlTranslatingExpressionVisitor

tuespetre:
This commit adds the `OuterPropertyExpression`, which is a dedicated expression to represent expressions bound to parent query models. This expression allows us to defer the decision of whether to bind directly to the parent query (i.e. an `AliasExpression`) or to bind via an outer parameter. There are several improved generated SQL queries in this PR as a result.

To facilitate this change, other changes take place:

- Subquery models are now visited only once, meaning no more unnecessary uniquification (things like `FROM Orders AS o4` instead of `FROM Orders AS o`)
- The `_bindParentQueries` field/parameter to `RelationalQueryModelVisitor` and `SqlTranslatingExpressionVisitor` is now gone, instead being driven by the context of the query model visitor and its ancestors via `CanBindOuterProperties`, `CanBindOuterParameters`, and `RequiresOuterParameterInjection`
- The `_targetSelectExpression` field/parameter to `SqlTranslatingExpressionVisitor` is now gone, being replaced by a `_mutateProjections` field/parameter. This helps to make it clearer what is being done every time and place the visitor is used (also, there were cases where the `_targetSelectExpression` comparison bit would hinder things from binding correctly.)

This PR is split from (and integral to) #7650 and #7543, and this PR includes/depends on the changes that make up #7619.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/ISqlTranslatingExpressionVisitorFactory.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/ISqlTranslatingExpressionVisitorFactory.cs
@@ -3,7 +3,6 @@
 
 using System.Linq.Expressions;
 using JetBrains.Annotations;
-using Microsoft.EntityFrameworkCore.Query.Expressions;
 
 namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 {
@@ -16,18 +15,16 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         ///     Creates a new SqlTranslatingExpressionVisitor.
         /// </summary>
         /// <param name="queryModelVisitor"> The query model visitor. </param>
-        /// <param name="targetSelectExpression"> The target select expression. </param>
         /// <param name="topLevelPredicate"> The top level predicate. </param>
-        /// <param name="bindParentQueries"> true to bind parent queries. </param>
         /// <param name="inProjection"> true if we are translating a projection. </param>
+        /// <param name="addToProjections"> false to avoid adding columns to projections. </param>
         /// <returns>
         ///     A SqlTranslatingExpressionVisitor.
         /// </returns>
         SqlTranslatingExpressionVisitor Create(
             [NotNull] RelationalQueryModelVisitor queryModelVisitor,
-            [CanBeNull] SelectExpression targetSelectExpression = null,
             [CanBeNull] Expression topLevelPredicate = null,
-            bool bindParentQueries = false,
-            bool inProjection = false);
+            bool inProjection = false,
+            bool addToProjections = true);
     }
 }
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
@@ -63,20 +63,55 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
             if (EntityQueryModelVisitor.IsPropertyMethod(node.Method))
             {
-                var newArg0 = Visit(node.Arguments[0]);
+                var handledPropertyExpression
+                    = TryHandlePropertyExpression(
+                        node,
+                        node.Arguments[0].RemoveConvert(),
+                        (node.Arguments[1] as ConstantExpression)?.Value as string);
 
-                if (newArg0 != node.Arguments[0])
+                if (handledPropertyExpression != null)
                 {
-                    return Expression.Call(
-                        node.Method,
-                        newArg0,
-                        node.Arguments[1]);
+                    return handledPropertyExpression;
                 }
-
-                return node;
             }
 
+            var handledMethodCallExpression
+                = TryHandleMemberOrMethodCallExpression(node);
+
+            if (handledMethodCallExpression != null)
+            {
+                return handledMethodCallExpression;
+            }
+
+            QueryModelVisitor.RequiresClientProjection = true;
             return base.VisitMethodCall(node);
+        }
+
+        /// <summary>
+        ///     Visit a member expression.
+        /// </summary>
+        /// <param name="node"> The expression to visit. </param>
+        /// <returns>
+        ///     An Expression corresponding to the translated member.
+        /// </returns>
+        protected override Expression VisitMember(MemberExpression node)
+        {
+            Check.NotNull(node, nameof(node));
+
+            var handledExpression
+                = TryHandlePropertyExpression(
+                    node,
+                    node.Expression.RemoveConvert(),
+                    node.Member.Name)
+                ?? TryHandleMemberOrMethodCallExpression(node);
+
+            if (handledExpression != null)
+            {
+                return handledExpression;
+            }
+
+            QueryModelVisitor.RequiresClientProjection = true;
+            return base.VisitMember(node);
         }
 
         /// <summary>
@@ -90,11 +125,32 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         {
             Check.NotNull(expression, nameof(expression));
 
-            var newNewExpression = base.VisitNew(expression);
+            if (QueryModelVisitor.QueryCompilationContext.QuerySourceRequiresMaterialization(_querySource))
+            {
+                QueryModelVisitor.RequiresClientProjection = true;
+            }
+            else
+            {
+                foreach (var qsre in expression.Arguments.OfType<QuerySourceReferenceExpression>())
+                {
+                    var entityType
+                        = QueryModelVisitor.QueryCompilationContext.Model
+                            .FindEntityType(qsre.ReferencedQuerySource.ItemType);
+
+                    if (entityType != null)
+                    {
+                        // TODO: Figure out how to make use of a composite/projection shaper
+                        QueryModelVisitor.RequiresClientProjection = true;
+                        break;
+                    }
+                }
+            }
+
+            var newNewExpression = (NewExpression)base.VisitNew(expression);
 
             var selectExpression = QueryModelVisitor.TryGetQuery(_querySource);
 
-            if (selectExpression != null)
+            if (selectExpression?.Projection.Count > 0)
             {
                 for (var i = 0; i < expression.Arguments.Count; i++)
                 {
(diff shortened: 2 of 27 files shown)


tuespetre:
List of queries with improved generated SQL (ignoring those that just changed in alias names due to the uniquification):

- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-4685ea6da86b37b446cabc9dbe924057R1658
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-4685ea6da86b37b446cabc9dbe924057R2438
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2c68636402e93e86a3d0108bbd4ba1ddL2074
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-e8ff8dabc15c4214f1f4885c86b23ce2R190
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-e8ff8dabc15c4214f1f4885c86b23ce2R448
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-e8ff8dabc15c4214f1f4885c86b23ce2R1017
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2L265
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2L297
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2R313
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2R508
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2R585
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2R3115
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2R3180
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2R5521
- https://github.com/aspnet/EntityFramework/pull/7663/files#diff-2602188597986ba3e53c5b56162786d2R6207

src/Microsoft.EntityFrameworkCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
```diff
@@ -3202,7 +3202,7 @@ public virtual void GroupJoin_in_subquery_with_client_projection()
                             where (from l1_inner in context.LevelOne.ToList()
                                    join l2_inner in context.LevelTwo.ToList() on l1_inner.Id equals l2_inner.Level1_Optional_Id into grouping
                                    from l2_inner in grouping.DefaultIfEmpty()
-                                   select l1_inner).Distinct().Count() > 7
+                                   select ClientStringMethod(l1_inner.Name)).Count() > 7
```
maumar: whoops :)

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1401,6 +1559,75 @@ var discriminatorPredicate
...
+                var ancestor = _queryModelVisitor.ParentQueryModelVisitor;
+
+                if (ancestor != null && !ancestor.CanBindOuterProperties)
+                {
+                    return false;
+                }
+
+                do
+                {
+                    if (ancestor.TryGetQuery(outerPropertyExpression.QuerySource) != null)
```
maumar: NRE

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalExpressionExtensions.cs
```diff
@@ -34,5 +34,17 @@ public static bool IsSimpleExpression([NotNull] this Expression expression)
                    || unwrappedExpression is ParameterExpression
                    || unwrappedExpression.IsAliasWithColumnExpression();
         }
+
+        public static Expression MaybeAnonymousSubquery([CanBeNull] this Expression expression)
```
smitpatel: Can you elaborate what this would generate and why it is needed?

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/ISqlTranslatingExpressionVisitorFactory.cs
```diff
@@ -16,18 +15,16 @@ public interface ISqlTranslatingExpressionVisitorFactory
...
         ///     A SqlTranslatingExpressionVisitor.
         /// </returns>
         SqlTranslatingExpressionVisitor Create(
             [NotNull] RelationalQueryModelVisitor queryModelVisitor,
-            [CanBeNull] SelectExpression targetSelectExpression = null,
             [CanBeNull] Expression topLevelPredicate = null,
-            bool bindParentQueries = false,
-            bool inProjection = false);
+            bool inProjection = false,
+            bool mutateProjections = true);
```
smitpatel: Purpose of `mutateProjections`?

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
tuespetre: thanks, will change it from `do...while` to `while`

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalExpressionExtensions.cs
tuespetre: `SqlTranslatingExpressionVisitor.VisitSubQuery` could return a `SelectExpression`, but it was always setting the `Alias` to `string.Empty`, which can cause invalid SQL in some cases now that more queries can be inlined. The solution was to only make the resulting `SelectExpression` into an anonymous subquery when appropriate.

smitpatel: Ah, I see. We need to investigate why it is needed. Ideally, we start with top level select expression without setting any alias. We need alias only when we add it as subquery in another select expression. While flattening can cause one of more select expression to be joined, there is no code path which would cause select expression to split, hence we should not have to set the alias to `string.Empty` ever. If we are doing so then we are assigning alias incorrectly in first place which should be fixed.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
```diff
@@ -1706,13 +1692,10 @@ public override void GroupJoin_on_complex_subquery_and_set_operation_on_grouping
...
-    FROM [Level1] AS [l10]
-    LEFT JOIN [Level2] AS [l1.OneToOne_Required_FK] ON [l10].[Id] = [l1.OneToOne_Required_FK].[Level1_Required_Id]
-    WHERE ([l10].[Name] <> N'L1 01') OR [l10].[Name] IS NULL
-) AS [t]",
+                @"SELECT [l10].[Id], [l10].[Date], [l10].[Name], [l10].[OneToMany_Optional_Self_InverseId], [l10].[OneToMany_Required_Self_InverseId], [l10].[OneToOne_Optional_SelfId], [l1.OneToOne_Required_FK].[Id], [l1.OneToOne_Required_FK].[Date], [l1.OneToOne_Required_FK].[Level1_Optional_Id], [l1.OneToOne_Required_FK].[Level1_Required_Id], [l1.OneToOne_Required_FK].[Name], [l1.OneToOne_Required_FK].[OneToMany_Optional_InverseId], [l1.OneToOne_Required_FK].[OneToMany_Optional_Self_InverseId], [l1.OneToOne_Required_FK].[OneToMany_Required_InverseId], [l1.OneToOne_Required_FK].[OneToMany_Required_Self_InverseId], [l1.OneToOne_Required_FK].[OneToOne_Optional_PK_InverseId], [l1.OneToOne_Required_FK].[OneToOne_Optional_SelfId]
```
maumar: I think I have a relatively simple fix in progress that makes this better, so that we don't regress here

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalExpressionExtensions.cs
tuespetre: Well the string.Empty makes it a parenthesized, anonymous subquery. It's just that with new optimizations, it should only be applied on the right occasion instead of unconditionally like it was before.

where (select count(*) from ....) > 7

and so forth. Sorry for the formatting, on a phone right now

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
tuespetre: Is this a regression though? The resulting SQL is smaller

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/ISqlTranslatingExpressionVisitorFactory.cs
tuespetre: There are places where the visitor should not be adding to the projections, like in the projection visitor, some result operator handlers... this prevents every place from having to use a 'projection count map' or whatever

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
smitpatel: @maumar - Subquery is getting lifted. Better SQL.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/ISqlTranslatingExpressionVisitorFactory.cs
smitpatel: Can we name it something different? We are trying to make SelectExpression as less `Mutable` as possible. 😄

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalExpressionExtensions.cs
smitpatel: Good point. Are all such calls go through single path where we can put it instead of having a function to do it?

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/ISqlTranslatingExpressionVisitorFactory.cs
tuespetre: Sure!

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalExpressionExtensions.cs
tuespetre: They did go through a single path but now not everything coming out of that single path should have it applied

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
tuespetre: Oh it looks like the selected columns are different. (More being selected.)

tuespetre:
:up::date:

- Fixed the possible NRE pointed out
- Renamed 'mutateProjections' parameter to 'addToProjections'
- Undid the changes related to #7619

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -63,23 +63,65 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                    node,
+                    node.Expression.RemoveConvert(),
+                    node.Member.Name);
+
+            if (handledPropertyExpression != null)
+            {
+                return handledPropertyExpression;
+            }
+
+            var handledMemberExpression
```
smitpatel: Use null propagation.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -63,23 +63,65 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
 
             if (EntityQueryModelVisitor.IsPropertyMethod(node.Method))
             {
-                var newArg0 = Visit(node.Arguments[0]);
+                var handledPropertyExpression
+                    = TryHandlePropertyExpression(
+                        node,
+                        node.Arguments.FirstOrDefault().RemoveConvert(),
```
smitpatel: Possible NRE.
Also why we need to use FirstOrDefault instead of direct indexes? Property method always have 2 args.

src/Microsoft.EntityFrameworkCore.Relational/Extensions/RelationalExpressionExtensions.cs
smitpatel: I looked into usages of this function. They are being used at various expressions without any corresponding logic around it. We shouldn't have it this way.
Encapsulate this code into the particular code path with condition where this is required. Subquery with empty string as alias is pitfall for many invalid SQL. We should identify the pattern where this is required and apply for those cases only. Ideally this function should be removed.

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
```diff
@@ -30,6 +30,8 @@ public class QueryCompilationContext
     {
         private readonly IRequiresMaterializationExpressionVisitorFactory _requiresMaterializationExpressionVisitorFactory;
         private readonly IEntityQueryModelVisitorFactory _entityQueryModelVisitorFactory;
+        private readonly IDictionary<QueryModel, EntityQueryModelVisitor> _entityQueryModelVisitors
```
smitpatel: 👍

smitpatel: This change is good in a way that we have reference to QMV which visited this QM already. Hence avoiding revising same QM again. @anpete - Any side effects of this? (like visiting a QM adding something to SelectExpression of some other QM)

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
tuespetre: ✔️

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
tuespetre: We don't need to, that should be an indexer ✔️

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -90,11 +132,32 @@ protected override Expression VisitNew(NewExpression expression)
         {
             Check.NotNull(expression, nameof(expression));
 
-            var newNewExpression = base.VisitNew(expression);
+            if (QueryModelVisitor.QueryCompilationContext.QuerySourceRequiresMaterialization(_querySource))
```
tuespetre: @smitpatel, regarding `MaybeAnonymousSubquery()`:

I added this condition as a workaround for not being able to always inline a subquery into a projection (due to an `EntityShaper` being present, or otherwise.) It looks like maybe this takes care of the reason I had put in `MaybeAnonymousSubquery` so I can remove it after all. 🎉

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
tuespetre: actually changed the preceding condition to return early if the ancestor is null instead

tuespetre:
:up::date:

Was able to remove `MaybeAnonymousSubquery()`

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
smitpatel: @tuespetre - This change would improve perf for query compiler though there is a slight chance of side-effects of caching querymodelvisitors if there are any changed flags in parent qmv in between multiple visits of same QM. The only way to verify is to just make caching changes and see if no tests change. Can you submit a separate PR for that?

tuespetre: I have approached it that way before (it was one of the first split-out PRs I tried to do) but it does not work. :chicken: ➕ 🥚; also with these changes there is no multiple visiting of the same query model.

smitpatel: If caching QMVs is breaking tests that means it has unintentional side-effects and we cannot cache it. While binding/lifting, QMV peeks into parent QMV and even though same query model is being visited it can produce different results. Mostly, caching is not usable here.

tuespetre: > even though same query model is being visited it can produce different results

That's my point, there same query model will not be visited more than once

tuespetre: > If caching QMVs is breaking tests that means it has unintentional side-effects

The problem I found was not with the caching itself, but that introducing caching actually **prevents** the query models from being visited more than once, and once you do that, you have to use/account for new and different flags on the parent QMV, such as `CanBindOuterProperties`, `RequiresOuterParameterBinding`, etc.

tuespetre: I will try removing the from this PR and see if it breaks anything. If it doesn't break, I will push up a commit that doesn't have the caching. Does that sound good to you

smitpatel: I tested out caching part and it breaks many tests. In current pipeline, visiting a QM not only dependent on outer qmv flags but also mutates the QM itself. When we fail to translate the QM, we have to revert the QM. We have methods like `PopulateQueryModelMapping` & `RecreateQueryModelFromMapping` for that. That is enough evidence to say that current pipeline generate different results while visiting same query model again.

Before we introduce caching on QMVs, we need to make sure that pipeline can generate the required translation of QM which can be used under any circumstances later.

Remove the caching part from this PR. We can add it when we reach above stage.

tuespetre: Ohhh you gave me an idea for how to deal with all the unnecessary uniquification without the caching bits (for another PR, another time) and also I think I know how to get around the removal of the caching for this PR.

PR #7669, opened 2017-02-21
Query: GroupJoin queries should use new Include compiler.

anpete:
Continuing the Include refactoring to apply to GroupJoin queries with ref Includes.


anpete:
Doing some more work on this.

PR #7671, opened 2017-02-21
Add Microsoft.EntityFrameworkCore.Sqlite.Core

### LINKED ISSUE #7667: Allow custom SQLite builds (winsqlite3, SQLCipher, etc.)
`Microsoft.Data.Sqlite` has switched to SQLitePCL.raw. By introducing a `Microsoft.EntityFrameworkCore.Sqlite.Core` package, we'll enable consuming projects to choose one of the native SQLite implementations provided by SQLitePCL.raw.

For example, to use `winsqlite3.dll`, use this:
```XML
<ItemGroup>
  <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite.Core" Version="1.2.0-*"/>
  <PackageReference Include="SQLitePCLRaw.bundle_winsqlite3" Version="1.1.2"/>
</ItemGroup>
```

The current SQLitePCLRaw.bundle_* packages they provide are:
* **e_sqlite3**--An up-to-date version of SQLite configured consistently across platforms
* **green** *(default)*--Same as e_sqlite3 except on iOS where the system's SQLite is used
* **sqlcipher**--A version of SQLite with encryption (exposed via `PRAGMA key`)
* **winsqlite3**--The version of SQLite included in the UWP SDK

You could also "drop" down to using specific providers (e.g. to use the system-wide SQLite library everywhere), but that scenario is more advanced so I won't cover it here.

For more information about SQLitePCL.raw, see [the project's README file](https://github.com/ericsink/SQLitePCL.raw/blob/master/README.md).

bricelam:
Use this package to specify a specific implementation of SQLite. For example:

```XML
<ItemGroup>
  <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite.Core" Version="1.2.0-*"/>
  <PackageReference Include="SQLitePCLRaw.bundle_winsqlite3" Version="1.1.2"/>
</ItemGroup>
```

Resolves #7667


src/Microsoft.EntityFrameworkCore.Sqlite/Microsoft.EntityFrameworkCore.Sqlite.csproj
```diff
@@ -1,19 +1,25 @@
...
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
   <Import Project="..\..\tools\EntityFramework.props" />
   <PropertyGroup>
     <Description>SQLite database provider for Entity Framework Core.</Description>
     <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>$(PackageTags);SQLite</PackageTags>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
```
bricelam: @natemcmaster This doesn't seem to do anything... (still need `Update="NETStandard.Library"` below)

natemcmaster: You probably need to update https://github.com/aspnet/EntityFramework/blob/dev/tools/EntityFramework.props#L23-L25

src/Microsoft.EntityFrameworkCore.Sqlite.Core/Microsoft.EntityFrameworkCore.Sqlite.Core.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\tools\EntityFramework.props" />
+  <PropertyGroup>
+    <AssemblyName>Microsoft.EntityFrameworkCore.Sqlite</AssemblyName>
+    <PackageId>Microsoft.EntityFrameworkCore.Sqlite.Core</PackageId>
+    <RootNamespace>Microsoft.EntityFrameworkCore.Sqlite</RootNamespace>
+    <Description>SQLite database provider for Entity Framework Core.</Description>
```
natemcmaster: Would be good to explain how this is different from "M.EF.Sqlite" i.e. that a users must also provide a reference that satisfies the native SQLite requirement.

tools/EntityFramework.props
```diff
@@ -20,7 +20,4 @@
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(OutputType)' == 'library'">
```
natemcmaster: We should still keep this until we resolve https://github.com/aspnet/Coherence-Signed/issues/401. Just add a condition to exclude when DisableImplicitFrameworkReferences is set

bricelam: Thanks. I was on my way out so I decided to let the CI see if it would actually work..

src/Microsoft.EntityFrameworkCore.Sqlite.Core/Microsoft.EntityFrameworkCore.Sqlite.Core.csproj
bricelam: I don't think anyone would see it here, so I filed aspnet/EntityFramework.Docs#354 instead. Also the exception message you get points you in the right direction.

tools/EntityFramework.props
bricelam: Updated.

src/Microsoft.EntityFrameworkCore.Sqlite/Microsoft.EntityFrameworkCore.Sqlite.csproj
bricelam: Updated. Thanks!

src/Microsoft.EntityFrameworkCore.Sqlite.Core/Microsoft.EntityFrameworkCore.Sqlite.Core.csproj
natemcmaster: Docs and runtime errors may help, but I'd still add something here. "Core" has become an overloaded term that confuses users. I suspect the ".Core" suffix will mislead new users into thinking Ms.Ef.Sqlite is for "old" projects and "MS.Ef.Sqlite.Core" is for .NET Core. Docs and runtime errors don't show up in NuGet.org or VS's NuGet package manager window.

ErikEJ: The description is quite visible in the NuGet UI in VS...

bricelam: Will defer to @rowanmiller. P.S. The `.Core` suffix may not be used in the end; I just "picked something", but intend to review it with the team.

rowanmiller:
:shipit:

PR #7686, opened 2017-02-22
Remove bad translators

### LINKED ISSUE #7656: Remove StartsWith, EndsWith, Contains translator for 2.0
Relational currently contains 3 query translators that provide incorrect results - since they are implemented with SQL LIKE and don't escape, % and _ are treated as wildcards and not as literals (#474). These emit warnings about the possibly incorrect results.

2.0 seems like the right version to remove these, forcing providers to write their own, correct translators.

roji:
Remove default implementations of String.{Contains,StartsWith,EndsWith} as they use SQL LIKE without handling escape characters (%, _). Every provider needs to implement this on their own.

Closes #7656


maumar:
@roji can you rebase on the latest dev?

roji:
@maumar dne

maumar:
Merged in 3d765b25599cf3d65922057d7d0fc0938089653b. Thank you for the contribution!

Issue #7689, opened 2017-02-23
Query : improve SQL for queries with left join that generate subqueries due to discriminator predicate


maumar:
In 291a548ce8655bfa07a6834795946b1f60eac876 we started introducing subqueries to handle discriminator predicates in queries with LEFT JOINs. Currently we project all columns in the inner SelectExpression individually, we could switch to SELECT [inner_table].*

This makes for smaller SQL but will also help in the future once we stop materializing everything in LEFT JOIN queries

maumar:
fixed in d4746fb0e70df11b2c3a0e1c28213fafe15e6a13

PR #7690, opened 2017-02-23
Fix to #7689 - Query : improve SQL for queries with left join that generate subqueries due to discriminator predicate

maumar:
Fix is to now produce SELECT [alias].* FROM ... for subqueries generated due to discriminator column being present in LOJ query, instead of listing all the columns individually.


src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -693,6 +693,10 @@ var predicate
                             selectExpression.PushDownSubquery();
                             selectExpression.ExplodeStarProjection();
                             tableExpression = selectExpression.Tables.Single();
+                            var innerSelectExpression = (SelectExpression)tableExpression;
```
smitpatel: `PushDownSubquery` returns this already.

PR #7691, opened 2017-02-23
Query: Improve logic to find the correct table in the select expression

### LINKED ISSUE #7220: Incorrect SQL Generated - selecting column from wrong table when filtering on navigation property and using skip/take
When filtering on a navigation property and then paging results using Skip/Take, incorrect SQL is generated.  This causes a SQL exception (in both SQL Server and SQLite).

```
Exception message: 
    SQLite Error 1: 'no such column: q.Book.MysteryBook.AuthorID'.
Stack trace:
       at Microsoft.Data.Sqlite.Interop.MarshalEx.ThrowExceptionForRC(Int32 rc, Sqlite3Handle db)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.Execute(IRelationalConnection connection, String executeMethod, IReadOnlyDictionary`2 parameterValues, Boolean closeConnection)
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.ExecuteReader(IRelationalConnection connection, IReadOnlyDictionary`2 parameterValues)
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryingEnumerable.Enumerator.BufferlessMoveNext(Boolean buffer)
   at Microsoft.EntityFrameworkCore.Query.QueryMethodProvider.<_ShapedQuery>d__3`1.MoveNext()
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at Microsoft.EntityFrameworkCore.Query.Internal.LinqOperatorProvider.ExceptionInterceptor`1.EnumeratorExceptionInterceptor.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at IncorrectSQLGeneration.Program.Main(String[] args) in C:\Users\ablake\Documents\Visual Studio 2015\Projects\IncorrectSQLGeneration\src\IncorrectSQLGeneration\Program.cs:line 51
```

### Steps to reproduce
Project that reproduces this issue can be found at: https://github.com/austinleroy/EFCore_IncorrectSQLGeneration

Source also attached to this issue.
[IncorrectSQLGeneration.zip](https://github.com/aspnet/EntityFramework/files/642647/IncorrectSQLGeneration.zip)

Console application that describes this issue:
#### Program.cs
```c#
using Microsoft.Data.Sqlite;
using System;
using System.Linq;
using System.Threading.Tasks;

namespace IncorrectSQLGeneration
{
    public class Program
    {
        {
            var connectionStringBuilder = new SqliteConnectionStringBuilder
            {
                DataSource = ":memory:"
            };
            string connectionString = connectionStringBuilder.ToString();
            var connection = new SqliteConnection(connectionString);
            connection.Open();

            DbContextOptionsBuilder<DataContext> builder = new DbContextOptionsBuilder<DataContext>();
            //Issue appears when generating SQL for SQLite
            builder.UseSqlite(connection);
            //Issue appears when generating SQL for SQL Server
            //builder.UseSqlServer("Data Source=localhost;Initial Catalog=EFCoreBroke;Integrated Security=true;");

            //Issue is not a problem with an in-memory database
            //builder.UseInMemoryDatabase();

            var log = new Microsoft.Extensions.Logging.LoggerFactory();
            log.AddConsole(LogLevel.Debug);
            log.AddDebug();
            builder.UseLoggerFactory(log);

            using (var context = new DataContext(builder.Options))
            {

                IQueryable<Quote> quotes =
                    context.Quotes
                    .Include(q => q.Book).ThenInclude(b => b.MysteryBook);

                //Commenting out filter on navigation property causes SQL to be generated correctly
                quotes = quotes.Where(q => q.Book.MysteryBook.Hero == "Sherlock");
                quotes = quotes.OrderBy(q => q.ID);
                //Commenting out Skip/Take causes SQL to be generated correctly
                quotes = quotes.Skip(0).Take(10);

                var authorIDs = quotes.Select(q => q.AuthorID).ToList();

                foreach(var id in authorIDs)
                {
                    Console.WriteLine(id);
                }
            }
        }
    }
}
```

#### Context & Models
```c#
using System;
using System.Linq;
using System.Threading.Tasks;
using System.ComponentModel.DataAnnotations;

namespace IncorrectSQLGeneration
{
    public class DataContext : DbContext
    {
        public DataContext(DbContextOptions<DataContext> builder)
            :base(builder)
        { }

        public DbSet<Quote> Quotes { get; set; }
        public DbSet<Book> Books { get; set; }
        public DbSet<MysteryBook> Publishers { get; set; }
    }

    [Table("Quote")]
    public partial class Quote
    {
        [Key]
        public int ID { get; set; }

        public int AuthorID { get; set; }

        public int? BookID { get; set; }

        [ForeignKey("BookID")]
        public virtual Book Book { get; set; }
    }

    [Table("Book")]
    public partial class Book
    {
        [Key]
        public int ID { get; set; }
        public virtual ICollection<Quote> Quotes { get; set; }
        public virtual MysteryBook MysteryBook { get; set; }
    }

    [Table("MysteryBook")]
    public partial class MysteryBook
    {
        [Key]
        public int BookID { get; set; }
        public string Hero { get; set; }

        [ForeignKey("BookID")]
        public virtual Book Book { get; set; }
    }
}

```

### Further technical details
EF Core version: 1.1.0
Database Provider: 
    "Microsoft.EntityFrameworkCore.Sqlite": "1.1.0"
    "Microsoft.EntityFrameworkCore.SqlServer": "1.1.0"
Operating system: Windows 10
IDE: Visual Studio 2015

smitpatel:
- Use ProjectStarTable as default table when SelectExpression is unable to find table for querysource
- Change ProjectStarAlias to ProjectStarTable
- Don't bind with subquery when subquery has client side projection (fixes #7220)
- Set ProjectStarTable more correctly

The main issue here was:
We were trying to bind to subquery which had client side projection. We cannot bind directly to select expression in such case because client method can generate different result then what is returned by the query.

When similar client methods applied for Where/OrderBy, it should be able to bind but the issue arose was unable to find the correct table since we fell back to last table in subquery. The correct table would be table set by ProjectStarAlias which is set when we encounter QSRE in RelationalProjectionExpressionVisitor only if we are inside correct SelectExpression & querysource.
Further, in case of simple identity QMs, we may not go into ProjectionExpressionVisitor making us unable to set ProjectStarAlias correctly. If there is only 1 table then it is the table which should be used.

Since we now need TableExpressionBase more than the string Alias, converted ProjectStarAlias to ProjectStarTable.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
@@ -470,9 +470,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                     Alias = joinExpression.Alias
                 };
 
-            // Don't create new alias when adding tables to subquery
             subquery.AddTable(joinExpression.TableExpression);
-            subquery.ProjectStarAlias = joinExpression.Alias;
+            subquery.ProjectStarTable = joinExpression;
             subquery.IsProjectStar = true;
             subquery.Predicate = selectExpression.Predicate;
 
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -31,6 +31,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 
         private Expression _limit;
         private Expression _offset;
+        private TableExpressionBase _projectStarTable;
 
         private int _subqueryDepth = -1;
 
@@ -85,12 +86,16 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
         public virtual Expression Predicate { get; [param: CanBeNull] set; }
 
         /// <summary>
-        ///     Gets or sets the table alias to be used for star projection.
+        ///     Gets or sets the table to be used for star projection.
         /// </summary>
         /// <value>
-        ///     The table alias.
+        ///     The table.
         /// </value>
-        public virtual string ProjectStarAlias { get; [param: CanBeNull] set; }
+        public virtual TableExpressionBase ProjectStarTable
+        {
+            get { return _projectStarTable ?? (_tables.Count == 1 ? _tables.Single() : null); }
+            [param: CanBeNull] set { _projectStarTable = value; }
+        }
 
         /// <summary>
         ///     Type of this expression.
@@ -265,7 +270,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             return _tables.FirstOrDefault(te
                        => te.QuerySource == querySource
                           || ((te as SelectExpression)?.HandlesQuerySource(querySource) ?? false))
-                   ?? _tables.Last();
+                   ?? ProjectStarTable;
         }
 
         /// <summary>
@@ -396,13 +401,13 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             subquery._offset = _offset;
             subquery._isDistinct = _isDistinct;
             subquery._subqueryDepth = _subqueryDepth;
-            subquery.ProjectStarAlias = ProjectStarAlias;
+            subquery.ProjectStarTable = ProjectStarTable;
             subquery.IsProjectStar = IsProjectStar || !subquery._projection.Any();
 
             _limit = null;
             _offset = null;
             _isDistinct = false;
-            ProjectStarAlias = null;
+            ProjectStarTable = null;
 
             Predicate = null;
 
@@ -411,7 +416,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             ClearOrderBy();
 
             _tables.Add(subquery);
-            ProjectStarAlias = subquery.Alias;
+            ProjectStarTable = subquery;
 
             foreach (var ordering in subquery.OrderBy)
             {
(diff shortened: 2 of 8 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
```diff
@@ -468,7 +468,7 @@ var subquery
 
             // Don't create new alias when adding tables to subquery
```
maumar: update the comment here also

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -80,12 +81,16 @@ public class SelectExpression : TableExpressionBase
...
+        ///     Gets or sets the table to be used for star projection.
         /// </summary>
         /// <value>
-        ///     The table alias.
+        ///     The table.
         /// </value>
-        public virtual string ProjectStarAlias { get; [param: CanBeNull] set; }
+        public virtual TableExpressionBase ProjectStarTable
+        {
+            get { return _projectStarTable ?? (_tables.Count == 1 ? _tables.Single() : null); }
```
maumar: FirstOrDefault gives same result

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs
smitpatel: That is actually unrelated comment but it should have been removed during changes to `SE.AddTable`

PR #7694, opened 2017-02-23
Query: Translate aggregate functions to server only if not on subquery

### LINKED ISSUE #3792: Query: Sum over subquery produces incorrect SQL
When I run this select:

```
var list6 = context.Customers.Select(x =&gt; new
         {
            Name = x.Name,
            InvoiceTotal = x.Invoices.Sum(p => p.InvoiceDetails.Sum(d =&gt; d.TotalAmount)),
            PaymentTotal = x.Payments.Sum(p =&gt; p.PaymentAmount)
         }
         ).ToList();
```

I get this error:
Cannot perform an aggregate function on an expression containing an aggregate or a subquery.

My DB context:

```
public class AppDBContext: DbContext, { 
public DbSet&lt;Customer&gt; Customers { get; set; }
public DbSet&lt;Invoice&gt; Invoices { get; set; }
public DbSet&lt;InvoiceDetail&gt;  InvoiceDetails { get; set; }
}

public class Customer 
{
public virtual ICollection&lt;Invoice&gt; Invoices { get; set; }
}

public class Invoice
{
#region Customer
public int CustomerID { get; set; }
public virtual Customer Customer { get; set; }
#endregion Customer

public virtual ICollection&lt;InvoiceDetail&gt; InvoiceDetails { get; set; }
}

public class InvoiceDetail 
   {
      public int InvoiceID { get; set; }
      public decimal TotalAmount { get; set; }

      public virtual Invoice Invoice { get; set; }
   }
```

smitpatel:
Fixes #3792 
Aggregate over subquery generates invalid sql. For Max/Min/Average/Sum if the inner projection is a subquery (aka selectexpresion) then we force client eval.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
@@ -254,24 +254,28 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
             {
                 var expression = handlerContext.SelectExpression.Projection.First();
 
-                var inputType = expression.Type;
-                var outputType = expression.Type;
-
-                var nonNullableInputType = inputType.UnwrapNullableType();
-                if (nonNullableInputType == typeof(int)
-                    || nonNullableInputType == typeof(long))
+                if (!(expression.RemoveConvert() is SelectExpression))
                 {
-                    outputType = inputType.IsNullableType() ? typeof(double?) : typeof(double);
+
+                    var inputType = expression.Type;
+                    var outputType = expression.Type;
+
+                    var nonNullableInputType = inputType.UnwrapNullableType();
+                    if (nonNullableInputType == typeof(int)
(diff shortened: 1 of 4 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
```diff
@@ -254,24 +254,28 @@ private static Expression HandleAverage(HandlerContext handlerContext)
             {
                 var expression = handlerContext.SelectExpression.Projection.First();
 
-                var inputType = expression.Type;
-                var outputType = expression.Type;
-
-                var nonNullableInputType = inputType.UnwrapNullableType();
-                if (nonNullableInputType == typeof(int)
-                    || nonNullableInputType == typeof(long))
+                if (!(expression is SelectExpression))
```
anpete: Does expression need RemoveConvert?

smitpatel: Possibly. will update.

PR #7695, opened 2017-02-23
Refactor RequiresMaterializationExpressionVisitor

tuespetre:
This is the code at the heart of the efforts I made in #7543 and #7650. 

I don't know why it didn't occur to me to make a branch that had just this in it without removing the 'training wheels' in the `QueryCompilationContext` (the code that brute-force flags all group joins as requiring materialization.) It actually went really easy, I just had to make a tiny little patch to SelectExpression and update some tests' SQL output (one which has a MUCH better query now, another which doesn't pull unnecessary columns anymore, and the rest which just have smaller resulting SQL text.)

Whenever the team is ready to work on the things I accomplished in those other two PRs, all that would need to be done is:

- Uncomment lines 416 and 417 in `RequiresMaterializationExpressionVisitor.cs`
- Remove the code in `QueryCompilationContext` that forces all group joins to be materialized
- Have fun

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/MemberAccessBindingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/MemberAccessBindingExpressionVisitor.cs
@@ -192,7 +192,12 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
                     if (newAccessOperation != nullConditionalExpression.AccessOperation)
                     {
-                        return newAccessOperation;
+                        return Expression.Condition(
+                            test: ValueBufferNullComparisonCheck(newCaller),
+                            ifTrue: newAccessOperation.Type != nullConditionalExpression.Type
+                                ? Expression.Convert(newAccessOperation, nullConditionalExpression.Type)
+                                : newAccessOperation,
+                            ifFalse: Expression.Default(nullConditionalExpression.Type));
                     }
                 }
             }
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
@@ -1,15 +1,18 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Clauses.ResultOperators;
+using Remotion.Linq.Clauses.StreamedData;
 
 namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 {
@@ -21,10 +24,10 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
     {
         private readonly IModel _model;
         private readonly EntityQueryModelVisitor _queryModelVisitor;
-        private readonly Dictionary<IQuerySource, int> _querySources = new Dictionary<IQuerySource, int>();
-
-        private QueryModel _queryModel;
-        private Expression _selector;
+        private readonly Stack<QueryModel> _queryModelStack = new Stack<QueryModel>();
+        private readonly Dictionary<IQuerySource, int> _querySourceReferences = new Dictionary<IQuerySource, int>();
+        private readonly QuerySourceTracingExpressionVisitor _querySourceTracingExpressionVisitor
+            = new QuerySourceTracingExpressionVisitor();
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
@@ -44,17 +47,28 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         /// </summary>
         public virtual ISet<IQuerySource> FindQuerySourcesRequiringMaterialization([NotNull] QueryModel queryModel)
         {
-            _querySources.Clear();
-            _queryModel = queryModel;
-            _selector = queryModel.SelectClause.Selector;
+            // Top-level query source result operators need to be promoted manually here
+            // because unlike subquerys' result operators, they won't be promoted via
+            // HandleUnderlyingQuerySources
+            foreach (var querySourceResultOperator in queryModel.ResultOperators.OfType<IQuerySource>())
+            {
+                PromoteQuerySource(querySourceResultOperator);
+            }
 
-            _queryModel.TransformExpressions(Visit);
+            _queryModelStack.Push(queryModel);
 
-            var querySources
-                = new HashSet<IQuerySource>(
-                    _querySources.Where(kv => kv.Value > 0).Select(kv => kv.Key));
+            queryModel.TransformExpressions(Visit);
 
-            return querySources;
+            _queryModelStack.Pop();
+
+            AdjustForResultOperators(queryModel);
+
+            var querySources =
+                from entry in _querySourceReferences
+                where entry.Value > 0
+                select entry.Key;
+
+            return new HashSet<IQuerySource>(querySources);
         }
 
         /// <summary>
@@ -64,22 +78,25 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         protected override Expression VisitQuerySourceReference(
             QuerySourceReferenceExpression expression)
         {
-            AddQuerySource(expression.ReferencedQuerySource);
+            PromoteQuerySource(expression.ReferencedQuerySource);
 
             return base.VisitQuerySourceReference(expression);
         }
 
-        private void AddQuerySource(IQuerySource querySource)
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitExtension(Expression node)
         {
-            if (!_querySources.ContainsKey(querySource))
+            if (node is NullConditionalExpression nullConditionalExpression)
             {
-                _querySources.Add(querySource, 0);
+                Visit(nullConditionalExpression.AccessOperation);
+
+                return node;
             }
 
-            if (_model.FindEntityType(querySource.ItemType) != null)
-            {
-                _querySources[querySource]++;
-            }
+            return base.VisitExtension(node);
         }
 
         /// <summary>
@@ -92,16 +109,23 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
             if (node.Expression != null)
             {
-                _queryModelVisitor
-                    .BindMemberExpression(
-                        node,
-                        (property, querySource) =>
-                            {
-                                if (querySource != null)
-                                {
-                                    _querySources[querySource]--;
-                                }
-                            });
+                if (node.Expression.Type.IsGrouping() && node.Member.Name == "Key")
+                {
+                    if (node.Expression is QuerySourceReferenceExpression qsre)
+                    {
+                        DemoteQuerySource(qsre.ReferencedQuerySource);
+                    }
+                }
+                else
+                {
+                    _queryModelVisitor.BindMemberExpression(node, (property, querySource) =>
+                    {
+                        if (querySource != null)
+                        {
+                            DemoteQuerySource(querySource);
+                        }
+                    });
+                }
             }
 
             return newExpression;
@@ -113,18 +137,27 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         /// </summary>
         protected override Expression VisitMethodCall(MethodCallExpression node)
         {
-            var newExpression = base.VisitMethodCall(node);
+            var newExpression = (MethodCallExpression)base.VisitMethodCall(node);
 
-            _queryModelVisitor
-                .BindMethodCallExpression(
-                    node,
-                    (property, querySource) =>
-                        {
-                            if (querySource != null)
-                            {
-                                _querySources[querySource]--;
-                            }
-                        });
+            _queryModelVisitor.BindMethodCallExpression(node, (property, querySource) =>
+            {
+                if (querySource != null)
+                {
+                    DemoteQuerySource(querySource);
+                }
+            });
+
+            foreach (var subQueryExpression in newExpression.Arguments.OfType<SubQueryExpression>())
+            {
+                if (subQueryExpression.QueryModel.ResultOperators.LastOrDefault() is IQuerySource querySourceResultOperator)
+                {
+                    PromoteQuerySource(querySourceResultOperator);
+                }
+                else if (subQueryExpression.QueryModel.SelectClause.Selector is QuerySourceReferenceExpression qsre)
+                {
+                    PromoteQuerySource(qsre.ReferencedQuerySource);
+                }
+            }
 
             return newExpression;
         }
@@ -135,39 +168,45 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         /// </summary>
         protected override Expression VisitBinary(BinaryExpression node)
         {
-            var oldParentSelector = _selector;
+            return node.Update(
+                left: VisitBinaryOperand(node.Left, node.NodeType),
+                conversion: node.Conversion,
+                right: VisitBinaryOperand(node.Right, node.NodeType));
+        }
 
-            var leftSubQueryExpression = node.Left as SubQueryExpression;
-
-            if ((leftSubQueryExpression != null)
-                && (_model.FindEntityType(leftSubQueryExpression.Type) != null))
+        private Expression VisitBinaryOperand(Expression operand, ExpressionType comparison)
+        {
+            if (comparison == ExpressionType.Equal || comparison == ExpressionType.NotEqual)
             {
-                _selector = leftSubQueryExpression.QueryModel.SelectClause.Selector;
+                var isEntityTypeExpression = _model.FindEntityType(operand.Type) != null;
 
-                leftSubQueryExpression.QueryModel.TransformExpressions(Visit);
+                if (isEntityTypeExpression)
+                {
+                    // An equality comparison of query source reference expressions
+                    // that reference an entity query source does not suggest that
+                    // materialization of that entity type may be required. This is true
+                    // whether in a join predicate, where predicate, selector, etc.
+                    // because it is rewritten into a key equality comparison.
+                    if (operand is QuerySourceReferenceExpression)
+                    {
+                        return operand;
+                    }
+
+                    // Same as above, key comparison, except coming from a subquery
+                    if (operand is SubQueryExpression subQueryExpression)
+                    {
+                        _queryModelStack.Push(subQueryExpression.QueryModel);
+
+                        subQueryExpression.QueryModel.TransformExpressions(Visit);
+
+                        _queryModelStack.Pop();
+
+                        return operand;
+                    }
+                }
             }
-            else
-            {
-                Visit(node.Left);
-            }
 
-            var rightSubQueryExpression = node.Right as SubQueryExpression;
-
-            if ((rightSubQueryExpression != null)
-                && (_model.FindEntityType(rightSubQueryExpression.Type) != null))
-            {
-                _selector = rightSubQueryExpression.QueryModel.SelectClause.Selector;
-
-                rightSubQueryExpression.QueryModel.TransformExpressions(Visit);
-            }
-            else
-            {
-                Visit(node.Right);
-            }
-
-            _selector = oldParentSelector;
-
-            return node;
+            return Visit(operand);
         }
 
         /// <summary>
@@ -176,61 +215,35 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         /// </summary>
         protected override Expression VisitSubQuery(SubQueryExpression expression)
         {
-            var oldParentSelector = _selector;
-            var oldQueryModel = _queryModel;
+            _queryModelStack.Push(expression.QueryModel);
 
-            _selector = expression.QueryModel.SelectClause.Selector;
-            _queryModel = expression.QueryModel;
+            expression.QueryModel.TransformExpressions(Visit);
 
-            _queryModel.TransformExpressions(Visit);
+            _queryModelStack.Pop();
 
-            _selector = oldParentSelector;
-            _queryModel = oldQueryModel;
+            AdjustForResultOperators(expression.QueryModel);
 
-            var querySourceReferenceExpression
-                = expression.QueryModel.SelectClause.Selector
-                    as QuerySourceReferenceExpression;
+            var parentQueryModel = _queryModelStack.Peek();
 
-            if (querySourceReferenceExpression != null)
+            var referencedQuerySource
+                = (expression.QueryModel.SelectClause.Selector
+                    as QuerySourceReferenceExpression)?
+                        .ReferencedQuerySource;
+
+            if (referencedQuerySource != null)
             {
-                var querySourceTracingExpressionVisitor = new QuerySourceTracingExpressionVisitor();
+                var parentQuerySource =
+                    (parentQueryModel.SelectClause.Selector as QuerySourceReferenceExpression)
+                        ?.ReferencedQuerySource;
 
-                if (expression.QueryModel.ResultOperators.LastOrDefault() is DefaultIfEmptyResultOperator)
+                if (referencedQuerySource.ItemType == parentQuerySource?.ItemType)
                 {
-                    var underlyingQuerySource = (((querySourceReferenceExpression.ReferencedQuerySource as MainFromClause)
-                            ?.FromExpression as QuerySourceReferenceExpression)
-                        ?.ReferencedQuerySource as GroupJoinClause)?.JoinClause;
+                    var resultSetOperators = GetSetResultOperatorSourceExpressions(parentQueryModel);
 
-                    if (underlyingQuerySource != null)
+                    if (resultSetOperators.Any(r => r.Equals(expression))
+                        && _querySourceReferences[parentQuerySource] > 0)
                     {
-                        AddQuerySource(underlyingQuerySource);
-                    }
-                }
-
-                var resultQuerySource
-                    = querySourceTracingExpressionVisitor
-                        .FindResultQuerySourceReferenceExpression(
-                            _selector,
-                            querySourceReferenceExpression.ReferencedQuerySource);
-
-                if ((resultQuerySource == null)
-                    && !(expression.QueryModel.ResultOperators.LastOrDefault() is OfTypeResultOperator))
-                {
-                    _querySources[querySourceReferenceExpression.ReferencedQuerySource]--;
-                }
-
-                foreach (var sourceExpression 
-                    in _queryModel.ResultOperators.Select(SetResultOperationSourceExpression).Where(e => e != null))
-                {
-                    if (sourceExpression.Equals(expression))
-                    {
-                        var parentQuerySource = _selector as QuerySourceReferenceExpression;
-                        if ((parentQuerySource != null)
-                            && (_querySources[parentQuerySource.ReferencedQuerySource] > 0)
-                            && (parentQuerySource.Type == querySourceReferenceExpression.Type))
-                        {
-                            _querySources[querySourceReferenceExpression.ReferencedQuerySource]++;
-                        }
+                        PromoteQuerySource(referencedQuerySource);
                     }
                 }
             }
@@ -238,29 +251,216 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
             return expression;
         }
 
-        private static Expression SetResultOperationSourceExpression(ResultOperatorBase resultOperator)
+        private void DemoteQuerySource(IQuerySource querySource)
         {
-            var concatOperator = resultOperator as ConcatResultOperator;
-            if (concatOperator != null)
+            HandleUnderlyingQuerySources(querySource, DemoteQuerySource);
+
+            if (_querySourceReferences.ContainsKey(querySource))
             {
-                return concatOperator.Source2;
+                _querySourceReferences[querySource]--;
+            }
+        }
+
+        private void PromoteQuerySource(IQuerySource querySource)
+        {
+            HandleUnderlyingQuerySources(querySource, PromoteQuerySource);
+
+            if (!_querySourceReferences.ContainsKey(querySource))
+            {
+                _querySourceReferences[querySource] = 1;
+            }
+            else
+            {
+                _querySourceReferences[querySource]++;
+            }
+        }
+
+        private void HandleUnderlyingQuerySources(IQuerySource querySource, Action<IQuerySource> action)
+        {
+            if (querySource is GroupResultOperator groupResultOperator)
+            {
+                var keySelectorExpression
+                    = groupResultOperator.KeySelector is SubQueryExpression keySelectorSubQuery
+                        ? keySelectorSubQuery.QueryModel.SelectClause.Selector
+                        : groupResultOperator.KeySelector;
+
+                if (keySelectorExpression is QuerySourceReferenceExpression keySelectorQsre)
+                {
+                    action(keySelectorQsre.ReferencedQuerySource);
+                }
+
+                var elementSelectorExpression
+                    = groupResultOperator.ElementSelector is SubQueryExpression elementSelectorSubQuery
+                        ? elementSelectorSubQuery.QueryModel.SelectClause.Selector
+                        : groupResultOperator.ElementSelector;
+
+                if (elementSelectorExpression is QuerySourceReferenceExpression elementSelectorQsre)
+                {
+                    action(elementSelectorQsre.ReferencedQuerySource);
+                }
+            }
+            else if (querySource is GroupJoinClause groupJoinClause)
+            {
+                action(groupJoinClause.JoinClause);
+            }
+            else
+            {
+                var underlyingExpression
+                    = ((querySource as FromClauseBase)?.FromExpression)
+                        ?? ((querySource as JoinClause)?.InnerSequence);
+
+                if (underlyingExpression is SubQueryExpression subQueryExpression)
+                {
+                    var finalResultOperator = subQueryExpression.QueryModel.ResultOperators.LastOrDefault();
+
+                    if (finalResultOperator is IQuerySource querySourceResultOperator)
+                    {
+                        action(querySourceResultOperator);
+                    }
+                    else if (subQueryExpression.QueryModel.SelectClause.Selector is QuerySourceReferenceExpression qsre)
+                    {
+                        action(qsre.ReferencedQuerySource);
+                    }
+                }
+                else if (underlyingExpression is QuerySourceReferenceExpression qsre)
+                {
+                    action(qsre.ReferencedQuerySource);
+                }
+            }
+        }
+
+        private void AdjustForResultOperators(QueryModel queryModel)
+        {
+            var referencedQuerySource
+               = (queryModel.SelectClause.Selector
+                   as QuerySourceReferenceExpression)?
+                       .ReferencedQuerySource;
+
+            // The selector may not have been a QSRE but this query model may still have something that needs adjusted.
+            // Example:
+            // context.Orders.GroupBy(o => o.CustomerId).Select(g => new { g.Key, g.Sum(o => o.TotalAmount) })
+            // The g.Sum(...) will result in a subquery model like { from Order o in [g] select o.TotalAmount => Sum() }.
+            // In that case we need to ensure that the referenced query source [g] is demoted.
+            if (referencedQuerySource == null)
+            {
+                referencedQuerySource
+                    = (queryModel.MainFromClause.FromExpression
+                        as QuerySourceReferenceExpression)?
+                            .ReferencedQuerySource;
             }
 
-            var exceptOperator = resultOperator as ExceptResultOperator;
-            if (exceptOperator != null)
+            if (referencedQuerySource == null)
             {
-                return exceptOperator.Source2;
+                return;
             }
 
-            var intersectOperator = resultOperator as IntersectResultOperator;
-            if (intersectOperator != null)
+            var isSubQuery = _queryModelStack.Count > 0;
+            var finalResultOperator = queryModel.ResultOperators.LastOrDefault();
+
+            if (isSubQuery && finalResultOperator is GroupResultOperator groupResultOperator)
             {
-                return intersectOperator.Source2;
+                // These two lines should be uncommented to implement GROUP BY translation.
+                //DemoteQuerySource(referencedQuerySource);
+                //DemoteQuerySource(groupResultOperator);
+                return;
             }
 
-            var unionOperator = resultOperator as UnionResultOperator;
+            var unreachableFromParentSelector =
+                 isSubQuery && _querySourceTracingExpressionVisitor
+                     .FindResultQuerySourceReferenceExpression(
+                         _queryModelStack.Peek().SelectClause.Selector,
+                         referencedQuerySource) == null;
 
-            return unionOperator?.Source2;
+            if (finalResultOperator is SingleResultOperator
+                || finalResultOperator is FirstResultOperator
+                || finalResultOperator is LastResultOperator)
+            {
+                // If not a subquery or if reachable from the parent selector
+                // we would not want to fall through to one of the next blocks.
+                if (unreachableFromParentSelector)
+                {
+                    DemoteQuerySourceAndUnderlyingFromClause(referencedQuerySource);
+                }
+
+                return;
+            }
+
+            if (ConvergesToSingleValue(queryModel))
+            {
+                // This is a top-level query that was not Single/First/Last
+                // but returns a single/scalar value (Avg/Min/Max/etc.)
+                // or a subquery that belongs to some outer-level query that returns 
+                // a single or scalar value. The referenced query source should be 
+                // re-promoted later if necessary.
+                DemoteQuerySourceAndUnderlyingFromClause(referencedQuerySource);
+                return;
+            }
+
+            if (isSubQuery && (unreachableFromParentSelector || finalResultOperator is DefaultIfEmptyResultOperator))
+            {
+                DemoteQuerySourceAndUnderlyingFromClause(referencedQuerySource);
+                return;
+            }
+        }
+
+        private bool ConvergesToSingleValue(QueryModel queryModel)
+        {
+            var outputInfo = queryModel.GetOutputDataInfo();
+
+            if (outputInfo is StreamedSingleValueInfo || outputInfo is StreamedScalarValueInfo)
+            {
+                return true;
+            }
+
+            foreach (var ancestorQueryModel in _queryModelStack)
+            {
+                outputInfo = ancestorQueryModel.GetOutputDataInfo();
+
+                if (outputInfo is StreamedSingleValueInfo || outputInfo is StreamedScalarValueInfo)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private void DemoteQuerySourceAndUnderlyingFromClause(IQuerySource querySource)
+        {
+            DemoteQuerySource(querySource);
+
+            var underlyingQuerySource
+                = ((querySource as FromClauseBase)
+                    ?.FromExpression as QuerySourceReferenceExpression)
+                        ?.ReferencedQuerySource;
+
+            if (underlyingQuerySource != null)
+            {
+                DemoteQuerySource(underlyingQuerySource);
+            }
+        }
+
+        private static IEnumerable<Expression> GetSetResultOperatorSourceExpressions(QueryModel queryModel)
+        {
+            foreach (var resultOperator in queryModel.ResultOperators)
+            {
+                if (resultOperator is ConcatResultOperator concatOperator)
+                {
+                    yield return concatOperator.Source2;
+                }
+                else if (resultOperator is ExceptResultOperator exceptOperator)
+                {
+                    yield return exceptOperator.Source2;
+                }
+                else if (resultOperator is IntersectResultOperator intersectOperator)
+                {
+                    yield return intersectOperator.Source2;
+                }
+                else if (resultOperator is UnionResultOperator unionOperator)
+                {
+                    yield return unionOperator.Source2;
+                }
+            }
         }
     }
 }
--- a/test/Microsoft.EntityFrameworkCore.InMemory.FunctionalTests/DatabaseErrorLogStateTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.InMemory.FunctionalTests/DatabaseErrorLogStateTest.cs
@@ -107,18 +107,6 @@ namespace Microsoft.EntityFrameworkCore.InMemory.FunctionalTests
             Query_logs_DatabaseErrorLogState(c => c.Blogs.FirstOrDefaultAsync().Wait());
         }
 
-        [Fact]
-        public void Query_logs_DatabaseErrorLogState_during_scalar()
-        {
-            Query_logs_DatabaseErrorLogState(c => c.Blogs.Count());
-        }
-
-        [Fact]
-        public void Query_logs_DatabaseErrorLogState_during_scalar_async()
-        {
-            Query_logs_DatabaseErrorLogState(c => c.Blogs.CountAsync().Wait());
-        }
-
         public void Query_logs_DatabaseErrorLogState(Action<BloggingContext> test)
         {
             var loggerFactory = new TestLoggerFactory();
--- a/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/QuerySourceTracingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/QuerySourceTracingExpressionVisitor.cs
@@ -101,7 +101,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         /// </summary>
         protected override Expression VisitSubQuery(SubQueryExpression expression)
         {
-            Visit(expression.QueryModel.SelectClause.Selector);
+            expression.QueryModel.TransformExpressions(Visit);
 
             return expression;
         }
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
@@ -2394,9 +2394,9 @@ ORDER BY [t].[Name]",
             Assert.Contains(
                 @"@__p_0: 2
 
-SELECT [t].[Id], [t].[Date], [t].[Name], [t].[OneToMany_Optional_Self_InverseId], [t].[OneToMany_Required_Self_InverseId], [t].[OneToOne_Optional_SelfId], [t].[c0], [t].[c1], [t].[Level1_Optional_Id], [t].[Level1_Required_Id], [t].[c2], [t].[OneToMany_Optional_InverseId], [t].[c3], [t].[OneToMany_Required_InverseId], [t].[c4], [t].[OneToOne_Optional_PK_InverseId], [t].[c5]
+SELECT [t].[Id], [t].[Date], [t].[Name], [t].[OneToMany_Optional_Self_InverseId], [t].[OneToMany_Required_Self_InverseId], [t].[OneToOne_Optional_SelfId]
 FROM (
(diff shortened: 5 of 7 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -170,6 +170,11 @@ public virtual void AddTable([NotNull] TableExpressionBase tableExpression)
         {
             Check.NotNull(tableExpression, nameof(tableExpression));
 
+            if (_tables.Count == 0)
```
maumar: this shouldn't be needed after you rebase on current dev

tuespetre: I thought that may be the case, I saw @smitpatel's other change (`ProjectStarTable`)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/MemberAccessBindingExpressionVisitor.cs
```diff
@@ -188,11 +188,20 @@ protected override Expression VisitExtension(Expression node)
                 if (newCaller != nullConditionalExpression.Caller
                     && newCaller.Type == typeof(ValueBuffer))
                 {
-                    var newAccessOperation = Visit(nullConditionalExpression.AccessOperation);
+                    var convertedAccessOperation
```
maumar: do the conversion after the visit and only if needed

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -92,16 +101,38 @@ protected override Expression VisitMember(MemberExpression node)
...
-                        (property, querySource) =>
-                            {
-                                if (querySource != null)
-                                {
-                                    _querySources[querySource]--;
-                                }
-                            });
+                if (node.Expression.Type.IsGrouping() && node.Member.Name == "Key")
+                {
+                    var groupResultOperator
```
tuespetre: The contents of this if block should be simplified to:

```
if (node.Expression is QuerySourceReferenceExpression qsre)
{
    DemoteQuerySource(qsre.ReferencedQuerySource);
}
```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +226,288 @@ protected override Expression VisitBinary(BinaryExpression node)
...
-                    var underlyingQuerySource = (((querySourceReferenceExpression.ReferencedQuerySource as MainFromClause)
-                            ?.FromExpression as QuerySourceReferenceExpression)
-                        ?.ReferencedQuerySource as GroupJoinClause)?.JoinClause;
+                    var resultSetOperators = GetSetResultOperatorSourceExpressions(parentQueryModel);
 
-                    if (underlyingQuerySource != null)
+                    if (resultSetOperators.Any())
                     {
-                        AddQuerySource(underlyingQuerySource);
+                        var parentStatistics = _querySourceReferences[parentQuerySource];
```
maumar: remove

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +226,288 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+        {
+            HandleUnderlyingQuerySources(querySource, PromoteQuerySource);
+
+            if (!_querySourceReferences.ContainsKey(querySource))
+            {
+                _querySourceReferences[querySource] = 1;
+            }
+            else
+            {
+                _querySourceReferences[querySource] += 1;
```
maumar: nit: ++

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +226,288 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+            return expression;
+        }
+
+        private void DemoteQuerySource(IQuerySource querySource)
+        {
+            HandleUnderlyingQuerySources(querySource, DemoteQuerySource);
+
+            if (_querySourceReferences.ContainsKey(querySource))
+            {
+                _querySourceReferences[querySource] -= 1;
```
maumar: nit: --

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/MemberAccessBindingExpressionVisitor.cs
```diff
@@ -188,11 +188,20 @@ protected override Expression VisitExtension(Expression node)
...
+                            nullConditionalExpression.Type);
 
-                    if (newAccessOperation != nullConditionalExpression.AccessOperation)
+                    var newAccessOperation = Visit(convertedAccessOperation);
+
+                    if (newAccessOperation != convertedAccessOperation)
                     {
-                        return newAccessOperation;
+                        var test = ValueBufferNullComparisonCheck(newCaller);
+                        var ifTrue = newAccessOperation;
```
maumar:
```
var ifTrue = newAccessOperation.Type != nullConditionalExpression.Type ? Expression.Convert(newAccessOperation, nullConditionalExpression.Type) : nullConditionalExpression;
```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -10,6 +10,9 @@
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Clauses.ResultOperators;
+using Remotion.Linq.Clauses.StreamedData;
+using System;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
```
smitpatel: sort usings

test/Microsoft.EntityFrameworkCore.InMemory.FunctionalTests/DatabaseErrorLogStateTest.cs
```diff
@@ -107,18 +107,6 @@ public void Query_logs_DatabaseErrorLogState_during_single_async()
             Query_logs_DatabaseErrorLogState(c => c.Blogs.FirstOrDefaultAsync().Wait());
         }
 
-        [Fact]
```
smitpatel: Why these tests are removed?

maumar: they no longer materialize entities therefore exception is no longer thrown

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/MemberAccessBindingExpressionVisitor.cs
```diff
@@ -188,11 +188,20 @@ protected override Expression VisitExtension(Expression node)
...
+                        var ifFalse = Expression.Default(convertedAccessOperation.Type);
+
+                        return Expression.Condition(test, ifTrue, ifFalse);
```
maumar: For which scenarios simply returning visited AccessOperation (after we adjust for potential type difference) is not enough?

tuespetre: CompexNavigationsInMemoryTest, GroupJoin_on_right_side_being_a_subquery

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +226,288 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                            break;
+                        }
+                    }
+                }
+
+                var finalResultOperator = queryModel.ResultOperators.LastOrDefault();
+
+                if (isSubQuery && finalResultOperator is GroupResultOperator groupResultOperator)
+                {
+                    //DemoteQuerySource(referencedQuerySource);
```
smitpatel: ???

tuespetre: The GroupResultOperator cannot be demoted until GROUP BY is being implemented. Those are the two lines I mentioned in my initial comments for this PR

smitpatel: Can you add the same in code comments? PR comments are disconnected with source (and even commit)

tuespetre: Sure thing

maumar:
I hit a regression in a couple tests: ```ComplexNavigationsQuerySqlServerTest.Null_protection_logic_work_for_inner_key_access_of_manually_created_GroupJoin2```
and 
```ComplexNavigationsQuerySqlServerTest.GroupJoin_on_complex_subquery_and_set_operation_on_grouping_but_nothing_from_grouping_is_projected```

for the first one, with the following query model:

```
join Level2 l2 in 
    join Level2 l1.OneToOne_Required_FK in DbSet<Level2>
    on Property(l1, "Id") equals Property(l1.OneToOne_Required_FK, "Level1_Required_Id") into l1.OneToOne_Required_FK_group
    from Level2 l1.OneToOne_Required_FK in 
        (from Level2 l1.OneToOne_Required_FK_groupItem in l1.OneToOne_Required_FK_group
        select l1.OneToOne_Required_FK_groupItem)
        .DefaultIfEmpty()
    select l1.OneToOne_Required_FK
on (Nullable<int>) l1.Id equals (?Property([l2], "Level1_Optional_Id")?) into l2s
select l1
```

the new algorithm marks:
```from Level1 l1 in value(EntityQueryable`1[Level1])```
```join Level2 l1.OneToOne_Required_FK in value(EntityQueryable`1[Level2]) on Property([l1], "Id") equals Property([l1.OneToOne_Required_FK], "Level1_Required_Id")```

for materialization.

This is incorrect, only Level1 should be marked as the result of inner join (that was the behavior before the change). The problem doesn't surface currently because the second query source gets marked for materialization anyway in QueryCompilationContext, but once we lift that requirement from QCC that scenario will be blocked from being optimized

**Comment by tuespetre** [comment]
That is probably because I was going off of the group join clause as the query source rather than its join clause to determine what required materialization. I will adjust for that, thanks for the example

**Comment by tuespetre** [comment]
:up::date:

File: src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs [hunk 12]
```diff
@@ -176,91 +211,282 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                    }
+                }
+                else if (isSubQuery)
+                {
+                    var tracer = new QuerySourceTracingExpressionVisitor();
+                    var traced = tracer.FindResultQuerySourceReferenceExpression(
+                        _queryModelStack.Peek().SelectClause.Selector,
+                        referencedQuerySource);
+                    
+                    if (traced == null || finalResultOperator is DefaultIfEmptyResultOperator)
```
tuespetre: @maumar This seems to be what was needed.

tuespetre: On second thought, I thought it through and I'm pretty sure this last block can go away and the previous block can test for convergesToSingleValue || isSubQuery. If it's a subquery there shouldn't be a need to trace it because the new implementation of PromoteQuerySource will re-promote it when necessary.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +211,282 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                        {
+                            DemoteQuerySource(referencedQuerySource);
+                        }
+                    }
+                }
+                else if (finalResultOperator is FirstResultOperator || finalResultOperator is LastResultOperator)
+                {
+                    var choiceResultOperator = (ChoiceResultOperatorBase)finalResultOperator;
+
+                    if (isSubQuery && choiceResultOperator.ReturnDefaultWhenEmpty)
```
tuespetre: I don't think this check (`choiceResultOperator.ReturnDefaultWhenEmpty`) really needs to be here, so this block can be collapsed with the preceding block. I've tested that change on my branch with all of the optimizations and everything stayed the same test-wise.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
maumar: yup, ran the updated code alongside my changes and the queries are better indeed

maumar:
@tuespetre, I found some more issues. Those are not regressions (current code is also wrong) but perhaps you might want to look into it.

Tests in question are:
```Collection_select_nav_prop_single_or_default_then_nav_prop_nested```
```Collection_select_nav_prop_first_or_default_then_nav_prop_nested```  
```Collection_select_nav_prop_first_or_default_then_nav_prop_nested_with_orderby```

a simplified version of the query:

```
ctx.Customers.Select(c => ctx.Orders.SingleOrDefault(o => o.OrderID == 10643).Customer.City)
```

and a variation (using client-side method):

```
ctx.Customers.Select(c => ClientFirstOrDefault(ctx.Orders).Customer.City));

        public static T ClientFirstOrDefault<T>(IQueryable<T> source)
        {
            return source.FirstOrDefault();
        }
```

query model:

```
from Customer c in DbSet<Customer>
select 
    (from Order o in DbSet<Order>
    join Customer o.Customer in DbSet<Customer>
    on Property(o, "CustomerID") equals Property(o.Customer, "CustomerID") into o.Customer_group
    from Customer o.Customer in 
        (from Customer o.Customer_groupItem in o.Customer_group
        select o.Customer_groupItem)
        .DefaultIfEmpty()
    where o.OrderID == 10643
    select o.Customer)
    .SingleOrDefault().City
```

so o.Customer should be marked for materialization, because it's behind SingleOrDefault (or a client method). Like with the previous example, it's not an issue currently because we blanket-materialize all joins, but will be a problem in the future.

**Comment by maumar** [comment]
for the client-side variant, the query model is:

```
from Customer c in DbSet<Customer>
select Order ClientFirstOrDefault(
    source: DbSet<Order>
)
.Customer.City
```

and orders don't get marked for materialization either

**Comment by maumar** [comment]
After giving it a bit more thought compensating for this probably doesn't belong here. Will take a stab at it in QCC

File: src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs [hunk 14]
```diff
@@ -135,39 +148,61 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+
+                    break;
             }
 
-            _selector = oldParentSelector;
+            return operand;
+        }
 
-            return node;
+        private void AddResultOperators(QueryModel queryModel)
```
maumar: No need to pass querymodel here, also this could be more general-purpose, e.g.
```
PromoteQuerySources(IEnumerable<IQuerySource> querySources)
```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
maumar: i think we are missing test coverage, there should definitely be a difference between First and FirstOrDefault when used in a subquery. FoD we can sometimes translate but first will always result in client evaluation. Wrt LastOrDefault, we can only translate it if the query also has orderby (we change the asc <-> desc and slap FirstOrDefault instead). Perhaps we can make this change as part of QueryModel optimization (before this step), so every time you encounter LoD here, you know it needs client eval.

smitpatel: We do not modify the LoD in optimizequery model instead we do it in HandleLast here https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs#L561

tuespetre: Whether or not those result operators need to be client evaluated doesn't affect whether the underlying query source needing materialization or not though. customers.Select(c => c.Name).Single() should not materialize the whole customer

tuespetre:
@maumar on your examples, with the SingleOrDefault, only the one field is selected so there is no materialization. The ClientFirstOrDefault is definitely an issue though.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Really the method could go away altogether, just needs to be:

if (queryModel.ResultOperators.LastOrDefault() is IQuerySource querySource) { PromoteQuerySource(querySource); }

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
maumar: @smitpatel yeah, but we can move that earlier

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Or rather, `foreach (var querySourceResultOperator in queryModel.ResultOperators.OfType<IQuerySource>())`. I mean realistically only the `GroupResultOperator` is an `IQuerySource` (I think) but for the sake of correctness? 😛

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
maumar: @tuespetre this is because in your example the subquery that Single is wrapped around doesn't need the c materialized on its own (it's not projected by that particular subquery). If the query model was instead:

```
from c in customers
select
{customers.Single())}.Name
```

(that thing will be optimized by relinq, but lets assume its not)

then the Single will force materialization on the inner subquery, and since result of that subquery is c, its query source should be marked.

What is happening in the code (in AdjustForResultOperators) is, we peek into the parent selector, and if that selector can't be traced back to the result of inner query, we demote the result of that inner query. This is ok for most cases, but with Single/First etc we still need the result of that inner subquery materialized. 

However one could argue that this is not a concern of the RequiresMaterialization visitor, since it shouldn't care which operators can be translated.

tuespetre: @maumar in the new example you gave, the customer still does not need to be materialized. But if it were to select a `NotMapped` property instead of `Name`, or call some client method on it instead of accessing the `Name` member, it should be materialized.

tuespetre:
@maumar I've investigated the ClientFirstOrDefault example further.

- First, an exception is thrown when evaluating the query, because `EntityQueryableExpressionVisitor` tries to make a `SelectExpression` for it but the `querySource` is `null` because it is in the `SelectClause`. The `EntityQueryableExpressionVisitor` should skip over `EntityQueryable` expressions in such cases and leave them for evaluation during client evaluation of the projection. But if it were changed to `ClientFirstOrDefault(from o in context.Orders select o)`, that would be different story because all of a sudden it can be assessed as a subquery instead of just an `EntityQueryable`.
- Second, and related to that, during the first run through of determining the query sources requiring materialization (for the parent query) we would not expect to see the orders marked as requiring materialization but on the subsequent pass through `RequiresMaterializationExpressionVisitor` we should expect to see them marked so (happening when the projection is client-evaluated.)
- Third, about how I mentioned `ClientFirstOrDefault(from o in context.Orders select o)`, I've looked at that and there is indeed a problem there, because the orders are not marked when they should be.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
maumar: @tuespetre yeah that's because relinq optimized the case. I gave this (simplified) case assuming that   the QM will be processed in this form. A real life example would be: (with the complex navigations model - using required navs so that QueryCompilationContext doesn't compensate)

```
ctx.LevelTwo.Select(l2 => ctx.LevelTwo.First().OneToOne_Required_FK_Inverse.Name)
```

this gets converted into the following QM:

```
from Level2 l1 in DbSet<Level2>
select 
    (from Level2 <generated>_1 in DbSet<Level2>
    join Level1 <generated>_1.OneToOne_Required_FK_Inverse in DbSet<Level1>
    on Property(<generated>_1, "Level1_Required_Id") equals Property(<generated>_1.OneToOne_Required_FK_Inverse, "Id")
    select <generated>_1.OneToOne_Required_FK_Inverse)
    .First().Name
```

now, this will be translated into the following query plan:

```
TRACKED: False
(QueryContext queryContext) => IEnumerable<string> _Select(
    source: IEnumerable<ValueBuffer> _ShapedQuery(
        queryContext: queryContext, 
        shaperCommandContext: SelectExpression: 
            SELECT 1
            FROM [Level2] AS [l1]
        , 
        shaper: ValueBufferShaper
    )
    , 
    selector: (ValueBuffer l1) => Level1 First(
        source: IEnumerable<Level1> _Select(
            source: IEnumerable<TransparentIdentifier<ValueBuffer, Level1>> _ShapedQuery(
                queryContext: queryContext, 
                shaperCommandContext: SelectExpression: 
                    SELECT TOP(1) [l0].[Id], [l0].[Date], [l0].[Name], [l0].[OneToMany_Optional_Self_InverseId], [l0].[OneToMany_Required_Self_InverseId], [l0].[OneToOne_Optional_SelfId]
                    FROM [Level2] AS [l]
                    INNER JOIN [Level1] AS [l0] ON [l].[Level1_Required_Id] = [l0].[Id]
                , 
                shaper: CompositeShaper`3
            )
            , 
            selector: (TransparentIdentifier<ValueBuffer, Level1> t0) => t0.Inner
        )
    )
    .Name
)
```

Note that the inner query, inside First() is materialized into entity, so everything works fine.
However, if you remove the ```&& choiceResultOperator.ReturnDefaultWhenEmpty``` from the line 417 this query will now not be marked for materialization and will fail to compile.

From the perspective on QM translation to query plan, First, Single, SingleOrDefault are indistinguishable from a simple client method that takes a collection and returns single element.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
maumar: Inlining sounds good.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: In that case though, it should not fail to compile because the `MemberAccessBindingExpressionVisitor` will check if the type is `ValueBuffer` and use `BindMemberToValueBuffer`

maumar: @tuespetre I think you are right. Currently we can't properly bind in this case, but that's a bug in the  compilation rather than the ```RequiresMaterializationExpressionVisitor```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -135,39 +163,53 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
 
-            var leftSubQueryExpression = node.Left as SubQueryExpression;
+            return node;
+        }
 
-            if ((leftSubQueryExpression != null)
-                && (_model.FindEntityType(leftSubQueryExpression.Type) != null))
+        private Expression VisitBinaryOperand(Expression operand, ExpressionType comparison)
+        {
+            switch (comparison)
```
maumar: convert this to a single if statement instead, there is only one case + default

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +218,267 @@ protected override Expression VisitBinary(BinaryExpression node)
         /// </summary>
         protected override Expression VisitSubQuery(SubQueryExpression expression)
         {
-            var oldParentSelector = _selector;
-            var oldQueryModel = _queryModel;
+            _queryModelStack.Push(expression.QueryModel);
```
maumar: don't we need to add group result operators here also? (like we do for the parent query)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +218,267 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                    //DemoteQuerySource(groupResultOperator);
+                }
+                else if (finalResultOperator is SingleResultOperator
+                    || finalResultOperator is FirstResultOperator
+                    || finalResultOperator is LastResultOperator)
+                {
+                    if (isSubQuery)
+                    {
+                        var tracer = new QuerySourceTracingExpressionVisitor();
+                        var traced = tracer.FindResultQuerySourceReferenceExpression(
```
maumar: how about:

```
var unreachableFromParentSelector =
    tracer.FindResultQuerySourceReferenceExpression(
        _queryModelStack.Peek().SelectClause.Selector,
        referencedQuerySource) == null;

if (unreachableFromParentSelector)
{
    DemoteQuerySource(referencedQuerySource);
}
```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +218,267 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                {
+                    action(qsre.ReferencedQuerySource);
+                }
             }
-
-            return expression;
         }
 
-        private static Expression SetResultOperationSourceExpression(ResultOperatorBase resultOperator)
+        private void AdjustForResultOperators(QueryModel queryModel)
```
maumar: in general, this method is quite complicated - can you provide some description of the logic behind it? (similar to the part in VisitBinary - I found that very useful)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: We do that on the parent query because it absolutely has to be, but for subqueries, any group result operators should be promoted/demoted as necessary via `HandleUnderlyingQuerySources`

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: I will give it a shot -- I also see some restructuring that could be done to make it 'flow' better, which should help

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +218,267 @@ protected override Expression VisitBinary(BinaryExpression node)
...
-            var exceptOperator = resultOperator as ExceptResultOperator;
-            if (exceptOperator != null)
+            if (referencedQuerySource != null)
             {
-                return exceptOperator.Source2;
+                var isSubQuery = _queryModelStack.Count > 0;
+                var outputInfo = queryModel.GetOutputDataInfo();
+                var convergesToSingleValue = false;
+
+                if (outputInfo is StreamedSingleValueInfo || outputInfo is StreamedScalarValueInfo)
```
maumar: this can be calculated later, only when it's needed. Extract it to ```ConvergesToSingleValue(queryModel)```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +218,267 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                        {
+                            convergesToSingleValue = true;
+                            break;
+                        }
+                    }
+                }
+
+                var finalResultOperator = queryModel.ResultOperators.LastOrDefault();
+
+                if (isSubQuery && finalResultOperator is GroupResultOperator groupResultOperator)
```
maumar: consider:

```if (condition)
{
...

return;
}

if (condition2)
{
...

return;
}```

this way there is a good place to put some comments (just before the ```if```s)

**Comment by maumar** [comment]
@tuespetre looking good, please address the remaining feedback, squash-rebase on latest dev, add some info in the commit message and it should be good to go :)

**Comment by tuespetre** [comment]
:up::date:

File: src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs [hunk 21]
```diff
@@ -176,91 +205,287 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                return;
             }
 
-            var exceptOperator = resultOperator as ExceptResultOperator;
-            if (exceptOperator != null)
+            var isSubQuery = _queryModelStack.Count > 0;
+
+            var finalResultOperator = queryModel.ResultOperators.LastOrDefault();
+
+            var unreachableFromParentSelector =
```
maumar: compute this after the GroupBy check

tuespetre: Done ✔️

ksmithRenweb:
@tuespetre @maumar  Thank you for working on this. When it gets released it will solve some major performance issues in our app. (It should reduce one very frequent call from 1 second to 30 ms. Blame a table with 300 columns being materialized, when we only need 3 of them in the query.)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -135,39 +163,40 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
-            }
-            else
-            {
-                Visit(node.Left);
-            }
+                // An equality comparison of query source reference expressions
+                // that reference an entity query source does not suggest that
+                // materialization of that entity type may be required. This is true
+                // whether in a join predicate, where predicate, selector, etc.
+                if (operand is QuerySourceReferenceExpression && isEntityTypeExpression)
```
smitpatel: Can you update comment that QSRE referencing entity type is re-written into key equality comparison.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -135,39 +163,40 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
         /// </summary>
         protected override Expression VisitBinary(BinaryExpression node)
         {
-            var oldParentSelector = _selector;
+            VisitBinaryOperand(node.Left, node.NodeType);
+            VisitBinaryOperand(node.Right, node.NodeType);
 
-            var leftSubQueryExpression = node.Left as SubQueryExpression;
+            return node;
```
smitpatel: Why this method return same node instead of the visited node like other methods.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -135,39 +163,40 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                // that reference an entity query source does not suggest that
+                // materialization of that entity type may be required. This is true
+                // whether in a join predicate, where predicate, selector, etc.
+                if (operand is QuerySourceReferenceExpression && isEntityTypeExpression)
+                {
+                    return;
+                }
 
-            var rightSubQueryExpression = node.Right as SubQueryExpression;
+                if (operand is SubQueryExpression subQueryExpression && isEntityTypeExpression)
```
smitpatel: What is the different when operand is subquery & its entitytype expression? This subquery expression will be revisited in VisitSubquery when you call Visit(operand)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +205,286 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+            _queryModelStack.Push(expression.QueryModel);
+
+            expression.QueryModel.TransformExpressions(Visit);
 
-            _selector = expression.QueryModel.SelectClause.Selector;
-            _queryModel = expression.QueryModel;
+            _queryModelStack.Pop();
 
-            _queryModel.TransformExpressions(Visit);
+            AdjustForResultOperators(expression.QueryModel);
```
smitpatel: What is guidance on using AdjustForResultOperators?
At very top level, we add result operator before visiting query model and then adjust result operators after visiting qm. Here we did not add result operators. Few lines above we also visiting query model but we are not doing anything with result operators.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +205,286 @@ protected override Expression VisitBinary(BinaryExpression node)
...
-                    var underlyingQuerySource = (((querySourceReferenceExpression.ReferencedQuerySource as MainFromClause)
-                            ?.FromExpression as QuerySourceReferenceExpression)
-                        ?.ReferencedQuerySource as GroupJoinClause)?.JoinClause;
+                    var resultSetOperators = GetSetResultOperatorSourceExpressions(parentQueryModel);
 
-                    if (underlyingQuerySource != null)
+                    if (resultSetOperators.Any() && _querySourceReferences[parentQuerySource] > 0)
                     {
-                        AddQuerySource(underlyingQuerySource);
+                        foreach (var sourceExpression in resultSetOperators)
```
smitpatel: Will there be more than one sourceExpression matching current expression?

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +205,286 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                        ?.QueryModel.SelectClause.Selector as QuerySourceReferenceExpression)
+                            ?.ReferencedQuerySource;
+
+                if (keySelectorSubQueryQuerySource != null)
+                {
+                    action(keySelectorSubQueryQuerySource);
+                }
+
+                var elementSelectorQuerySource
+                    = (groupResultOperator.ElementSelector as QuerySourceReferenceExpression)
```
smitpatel: This is being repeated multiple times. Make a method which try safe cast to QSRE and returns referenced query source.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +205,286 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                        ?.ReferencedQuerySource;
+
+                if (elementSelectorQuerySource != null)
+                {
+                    action(elementSelectorQuerySource);
+                }
+
+                var elementSelectorSubQueryQuerySource
+                    = ((groupResultOperator.ElementSelector as SubQueryExpression)
+                        ?.QueryModel.SelectClause.Selector as QuerySourceReferenceExpression)
```
smitpatel: Similar to above, function to get selector after safe casting to Subqueryexpression

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +205,286 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+            {
+                // These two lines should be uncommented to implement GROUP BY translation.
+                //DemoteQuerySource(referencedQuerySource);
+                //DemoteQuerySource(groupResultOperator);
+
+                return;
+            }
+
+            var unreachableFromParentSelector =
+                 isSubQuery && new QuerySourceTracingExpressionVisitor()
```
smitpatel: >new QuerySourceTracingExpressionVisitor() [](start = 31, length = 41)

This can (or should) be stored in static in this class so that we don't initialize it all the time. The same instance can serve all the invocations of the function.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +205,286 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+                         referencedQuerySource) == null;
+
+            if (finalResultOperator is SingleResultOperator
+                || finalResultOperator is FirstResultOperator
+                || finalResultOperator is LastResultOperator)
             {
-                return concatOperator.Source2;
+                // If not a subquery or if reachable from the parent selector
+                // we would not want to fall through to one of the next blocks.
+                if (isSubQuery && unreachableFromParentSelector)
```
smitpatel: unreachableFromParentSelector asserts isSubQuery already.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Sure

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Because the version before my changes did the same. This visitor doesn't actually change anything, but I guess it wouldn't hurt to return `node.Update(newLeft, node.Conversion, newRight)`.

smitpatel: We should be consistent about it. Either we return same expression in everything or we send updated. Change every overridden method to return the same expression which came via parameter.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: The previous version did the same thing, it should be the same thing as the `QuerySourceReferenceExpression` case -- for key equality comparisons.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -64,22 +73,25 @@ var querySources
...
 
             return base.VisitQuerySourceReference(expression);
         }
 
-        private void AddQuerySource(IQuerySource querySource)
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitExtension(Expression node)
```
smitpatel: @maumar to verify where and how this is needed.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: The top-level `IQuerySource` result operators are added that way because it's the top level and they won't get promoted the same way they do for subqueries (via `HandleUnderlyingQuerySources`.)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: It is not thread-safe so `static` is off limits. Caching a single instance per `RequiresMaterializationExpressionVisitor` may work.

smitpatel:
Loos good 😄

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: @smitpatel it's needed because we are counting references and visiting each of the `Caller` and the `NullableCaller` and the `AccessOperation` would throw off the count (as I found when working on materialization changes.)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Shouldn't be, the old version used a `foreach` loop too, a more succinct version would be:

```
                    if (resultSetOperators.Any(r => r.Equals(expression)) 
                        && _querySourceReferences[parentQuerySource] > 0)
                    {
                        PromoteQuerySource(referencedQuerySource);
                    }
```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: This line was written before the VS2017 conversion, and now the language makes it so much nicer 😉 

```
if (groupResultOperator.KeySelector is QuerySourceReferenceExpression keySelectorQsre)
{
    action(keySelectorQsre.ReferencedQuerySource);
}
```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
maumar: Maybe we can do a check similar to the one in binary expression here - if Caller/NullableCaller is qsre of type entity, we don't promote, but otherwise we do

tuespetre: Don't we need `VisitMember` to ensure proper counting though? What if the accessed property is `NotMapped`

maumar: You would visit the AccessExpression, just it's done currently, but on top of that peek into the Callers. They *should* be just entity qsres for all the real cases, so for all those cases, the code will behave the same way. But we would also cover our bases for the weird cases (e.g. when the user manually creates NullConditionalExpression)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: I'm changing to return `node.Update(newLeft, node.Conversion, newRight)`

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Can we possibly revisit that later, if it becomes an issue

smitpatel: > Can we possibly revisit that later, if it becomes an issue

Are we removing this override fully then? If yes, then its fine to revisit later.

tuespetre: @smitpatel @maumar I'm not sure I understand. 

If we just remove this override altogether, the base implementation skips visiting the `NullConditionalExpression` entirely. The override here is at least an improvement and makes sure we count the references as we should; we can take it out but then we may not be catching things we should (like `NotMapped` properties.) 

And what about the other places that only work with `AccessOperation`, like `ExpressionEqualityComparer` and `SqlTranslatingExpressionVisitor`? That's why I'm asking if it can be revisited later: because places like that aren't doing anything special with `Caller` and `NullableCaller`, and if we're going to put that under a microscope here, shouldn't it be addressed in the other places too?

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Adding comments for this.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Adding comments for this.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre: Making a change that will make this more terse/compact but not quite enough to justify a method, IMO.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
smitpatel: Can you try removing this whole part and see how VisitSubquery works out?

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
smitpatel: It is being repeated at multiple places in same pattern. It can be simple method `TryGetSelectorFromSubquery`

tuespetre:
```
Test 'Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.RowNumberPagingTest.Where_query_composition_is_not_null' failed:
	System.InvalidOperationException : Reference equality is not defined for the types 'Microsoft.EntityFrameworkCore.Storage.ValueBuffer' and 'System.Object'.
```

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
tuespetre:
```
Test 'Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.RowNumberPagingTest.Where_query_composition_is_not_null' failed:
	System.InvalidOperationException : Reference equality is not defined for the types 'Microsoft.EntityFrameworkCore.Storage.ValueBuffer' and 'System.Object'.
```

(Posted under the wrong comment at first!)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +215,252 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+            {
+                var keySelectorExpression
+                    = groupResultOperator.KeySelector is SubQueryExpression keySelectorSubQuery
+                        ? keySelectorSubQuery.QueryModel.SelectClause.Selector
+                        : groupResultOperator.KeySelector;
+
+                if (keySelectorExpression is QuerySourceReferenceExpression keySelectorQsre)
                 {
-                    _querySources[querySourceReferenceExpression.ReferencedQuerySource]--;
+                    action(keySelectorQsre.ReferencedQuerySource);
```
tuespetre: @smitpatel It's only this block using this exact pattern, and only two times next to each other, just doesn't seem like enough to warrant a whole separate method...

tuespetre:
:up::date: added more comments, made a few things more terse.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
smitpatel: > the base implementation skips visiting the NullConditionalExpression entirely.

Wasn't aware of that. I assumed it would call VisitChildren in class. Since it is skipping over, lets keep this. @maumar - to file issue on usage of `NullConditionalExpression` (based on our discussion around caller) and link it to #7520

maumar: we were under the assumptions that the VisitChildren would get called, which would visit all the nodes. However thats not the case (we special case NullConditional in the base visitor), so this is fine as it is. I will try to come up with some tests, see if I can break it, but this can be done later. Let's leave this method as is for now.

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
smitpatel: It is being used many times like line 156 or 237. I suppose this thing is being used at multiple places even outside of this file and a good candidate for extension method. You can leave it as is though. I will update it in future refactoring.

---
In reply to: [103548170](https://github.com/aspnet/EntityFramework/pull/7695#discussion_r103548170) [](ancestors = 103548170)

src/Microsoft.EntityFrameworkCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -176,91 +205,286 @@ protected override Expression VisitBinary(BinaryExpression node)
...
+            if (finalResultOperator is SingleResultOperator
+                || finalResultOperator is FirstResultOperator
+                || finalResultOperator is LastResultOperator)
             {
-                return concatOperator.Source2;
+                // If not a subquery or if reachable from the parent selector
+                // we would not want to fall through to one of the next blocks.
+                if (isSubQuery && unreachableFromParentSelector)
+                {
+                    DemoteQuerySourceWithUnderlying(referencedQuerySource);
```
maumar: I don't like the name either, but can't come up with anything much better. Maybe DemoteQuerySourceAndItsFromClause? Method is not very complicated, so having a bad name is not a big deal. I'm fine with either.

tuespetre: `DemoteQuerySourceAndUnderlyingFromClause`?

maumar: @tuespetre sounds good to me

maumar:
@tuespetre please squash-rebase on current dev and I will merge it

tuespetre:
@maumar done 🏁

maumar:
@tuespetre merged in 67b4d2228c548189fa727eb7a790eb286e19099f Thank you for the contribution!

tuespetre:
Awesome, thanks @maumar! Question now, is anyone on the team already working on the group join materialization stuff?

maumar:
I'm almost done with a partial fix (one last issue left to figure out, I hope) that only requires materialization for queries when we actually introduce client GroupJoin. This fixes optional navs case where we can eliminate the client GroupJoin because the LOJ is flattened with SelectMany.

maumar:
Also a while back @anpete was working on a more general solution but hit some roadblocks. IIRC the problematic case was when there was nothing from the outer in the final result, so you couldn't figure out how to split the groups. We put that solution on hold for now :)

PR #7696, opened 2017-02-23
Improve binding to parent query models

tuespetre:
This is a much smaller, less radical version of #7663 that brings only enhancements (no regressions) to the plate.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -49,8 +49,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         private readonly IRelationalTypeMapper _relationalTypeMapper;
         private readonly SelectExpression _targetSelectExpression;
         private readonly Expression _topLevelPredicate;
-
-        private readonly bool _bindParentQueries;
+        
         private readonly bool _inProjection;
 
         /// <summary>
@@ -60,14 +59,12 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         /// <param name="queryModelVisitor"> The query model visitor. </param>
         /// <param name="targetSelectExpression"> The target select expression. </param>
         /// <param name="topLevelPredicate"> The top level predicate. </param>
-        /// <param name="bindParentQueries"> true to bind parent queries. </param>
         /// <param name="inProjection"> true if the expression to be translated is a LINQ projection. </param>
         public SqlTranslatingExpressionVisitor(
             [NotNull] SqlTranslatingExpressionVisitorDependencies dependencies,
             [NotNull] RelationalQueryModelVisitor queryModelVisitor,
             [CanBeNull] SelectExpression targetSelectExpression = null,
             [CanBeNull] Expression topLevelPredicate = null,
-            bool bindParentQueries = false,
             bool inProjection = false)
         {
             Check.NotNull(dependencies, nameof(dependencies));
@@ -81,7 +78,6 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
             _queryModelVisitor = queryModelVisitor;
             _targetSelectExpression = targetSelectExpression;
             _topLevelPredicate = topLevelPredicate;
-            _bindParentQueries = bindParentQueries;
             _inProjection = inProjection;
         }
 
@@ -612,47 +608,35 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 }
             }
 
-            var expression
-                = _queryModelVisitor
-                      .BindMethodCallExpression(methodCallExpression, CreateAliasedColumnExpression)
-                  ?? _queryModelVisitor.BindLocalMethodCallExpression(methodCallExpression);
-
-            if (expression == null
-                && _bindParentQueries)
-            {
-                expression
-                    = TryBindParentExpression(
-                        _queryModelVisitor.ParentQueryModelVisitor,
-                        qmv => qmv.BindMethodCallExpression(methodCallExpression, CreateAliasedColumnExpressionCore));
-            }
-
-            return expression
-                   ?? _queryModelVisitor.BindMethodToOuterQueryParameter(methodCallExpression);
+            return TryBindMemberOrMethodToAliasExpression(methodCallExpression, (expression, visitor, binder)
+                    => visitor.BindMethodCallExpression(expression, binder))
+                ?? _queryModelVisitor.BindLocalMethodCallExpression(methodCallExpression)
+                ?? _queryModelVisitor.BindMethodToOuterQueryParameter(methodCallExpression);
         }
 
         /// <summary>
         ///     Visit a member expression.
         /// </summary>
-        /// <param name="expression"> The expression to visit. </param>
+        /// <param name="memberExpression"> The expression to visit. </param>
         /// <returns>
         ///     An Expression.
         /// </returns>
-        protected override Expression VisitMember(MemberExpression expression)
+        protected override Expression VisitMember(MemberExpression memberExpression)
         {
-            Check.NotNull(expression, nameof(expression));
+            Check.NotNull(memberExpression, nameof(memberExpression));
 
-            if (!(expression.Expression is QuerySourceReferenceExpression)
-                && !(expression.Expression is SubQueryExpression))
+            if (!(memberExpression.Expression is QuerySourceReferenceExpression)
+                && !(memberExpression.Expression is SubQueryExpression))
             {
-                var newExpression = Visit(expression.Expression);
+                var newExpression = Visit(memberExpression.Expression);
 
                 if (newExpression != null
-                    || expression.Expression == null)
+                    || memberExpression.Expression == null)
                 {
                     var newMemberExpression
-                        = newExpression != expression.Expression
-                            ? Expression.Property(newExpression, expression.Member.Name)
-                            : expression;
+                        = newExpression != memberExpression.Expression
+                            ? Expression.Property(newExpression, memberExpression.Member.Name)
+                            : memberExpression;
 
                     var translatedExpression = _memberTranslator.Translate(newMemberExpression);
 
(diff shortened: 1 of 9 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1594,10 +1605,19 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
 
         private ParameterExpression BindPropertyToOuterParameter(IQuerySource querySource, IProperty property, bool isMemberExpression)
         {
-            if (querySource != null && _canBindPropertyToOuterParameter)
+            if (querySource != null && _canBindPropertyToOuterParameter && ParentQueryModelVisitor != null)
             {
-                var parentSelectExpression = ParentQueryModelVisitor?.TryGetQuery(querySource);
-                if (parentSelectExpression != null)
+                var ancestor = ParentQueryModelVisitor;
```
smitpatel: variable name -> outerQueryModelVisitor

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1594,10 +1605,19 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
 
         private ParameterExpression BindPropertyToOuterParameter(IQuerySource querySource, IProperty property, bool isMemberExpression)
         {
-            if (querySource != null && _canBindPropertyToOuterParameter)
+            if (querySource != null && _canBindPropertyToOuterParameter && ParentQueryModelVisitor != null)
```
smitpatel: remove null check on qmv

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1594,10 +1605,19 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
...
             {
-                var parentSelectExpression = ParentQueryModelVisitor?.TryGetQuery(querySource);
-                if (parentSelectExpression != null)
+                var ancestor = ParentQueryModelVisitor;
+                SelectExpression outerQuery = null;
```
smitpatel: var name -> outerSelectExpression

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1594,10 +1605,19 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
...
+
+                do
```
smitpatel: Remove do part of loop by assigning values when declaring variables.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +646,79 @@ var newMemberExpression
...
         {
-            if (queryModelVisitor == null || queryModelVisitor.RequiresClientProjection)
+            AliasExpression CreateAliasExpression(
+                IProperty property, IQuerySource querySource, SelectExpression selectExpression)
             {
-                return null;
+                var column = _relationalAnnotationProvider.For(property).ColumnName;
+                var table = selectExpression.GetTableForQuerySource(querySource);
+
+                return new AliasExpression(new ColumnExpression(column, property, table));
```
smitpatel: make inline.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +646,79 @@ var newMemberExpression
...
+                return aliasExpression;
+            });
+
+            if (bound != null)
+            {
+                return bound;
+            }
+
+            var ancestor = _queryModelVisitor.ParentQueryModelVisitor;
+            var canBindToAncestor = _queryModelVisitor.CanBindToParentQueryModel;
```
smitpatel: You may not be able to bind to parent query model visitor if it requires client projection.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +646,79 @@ var newMemberExpression
...
-                        qmv => qmv.BindMemberExpression(expression, CreateAliasedColumnExpressionCore));
-            }
+            return TryBindAliasExpression(node, (visitor, binder)
+                    => visitor.BindMemberExpression(node, binder))
+                ?? TryBindQuerySourcePropertyExpression(node)
+                ?? _queryModelVisitor.BindMemberToOuterQueryParameter(node);
+        }
 
-            if (aliasExpression == null)
+        private Expression TryBindQuerySourcePropertyExpression(MemberExpression expression)
```
smitpatel: Is this method being used elsewhere? If not then we don't need to create this method.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
tuespetre: Why? It used twice. (See: https://github.com/aspnet/EntityFramework/pull/7696/files/e7e61afd4d65c3c1d158f9bd7911440f10f32148#diff-c0990b2bb9d2ed7c8071a0c4323c01a4R708)

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
tuespetre: CanBindToParentQueryModel should never be true if the parent qmv requires client projection.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
smitpatel: Inline in the sense, you don't need to declare variables for `column` & `table` just put them directly in constructor since they are not being used.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
smitpatel: Can you point me to the place where we are setting that flag appropriately?

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
tuespetre: There are several other methods (even in this class) that are only used in one place. I feel that the readability was greatly improved by making this into its own method:

```
return TryBindAliasExpression(node, (visitor, binder)
        => visitor.BindMemberExpression(node, binder))
    ?? TryBindQuerySourcePropertyExpression(node)
    ?? _queryModelVisitor.BindMemberToOuterQueryParameter(node);
```

Additionally, this method can be expanded on (for instance, to add support for translating access of a grouping key.)

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
tuespetre: K, thanks

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
tuespetre: The flag is set by `RelationalQueryModelVisitor.VisitSubQueryModel`, which is called in `RelationalQueryModelVisitor.LiftSubQuery` and `SqlTranslatingExpressionVisitor.VisitSubQuery`.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -612,49 +608,36 @@ var boundExpression
...
         /// <summary>
         ///     Visit a member expression.
         /// </summary>
-        /// <param name="expression"> The expression to visit. </param>
+        /// <param name="node"> The expression to visit. </param>
         /// <returns>
         ///     An Expression.
         /// </returns>
-        protected override Expression VisitMember(MemberExpression expression)
+        protected override Expression VisitMember(MemberExpression node)
```
tuespetre: change to `memberExpression` or revert to `expression`?

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
smitpatel: Sounds good then.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -191,6 +189,16 @@ internal virtual bool RequiresClientSingleColumnResultOperator
...
 
         /// <summary>
+        ///     Gets or sets a value indicating whether this query model visitor will be
+        ///     able to bind directly to properties from its parent query without requiring
+        ///     parameter injection.
+        /// </summary>
+        /// <value>
+        ///     true if the query model visitor can bind to its parent's properties, false if not.
+        /// </value>
+        public virtual bool CanBindToParentQueryModel { get; protected set; } = false;
```
smitpatel: no need to set to false. Its default

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
smitpatel: FWIW, When a new query model visitor is created, the flag is set to false. So in the case of client projections we create the qmv in ProjectExpressionVisitor which would be called for client projections only.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +646,79 @@ var newMemberExpression
...
-            if (aliasExpression == null
-                && _bindParentQueries)
-            {
-                aliasExpression
-                    = TryBindParentExpression(
-                        _queryModelVisitor.ParentQueryModelVisitor,
-                        qmv => qmv.BindMemberExpression(expression, CreateAliasedColumnExpressionCore));
-            }
+            return TryBindAliasExpression(node, (visitor, binder)
+                    => visitor.BindMemberExpression(node, binder))
```
smitpatel: This is actually a closure rather than a Func since you are passing value of node into the lambda expression. If you look at the function `TryBindAliasExpression` then you will see the first parameter is never used in its method body. We don't want to create a closure here. Instead change this lambda to take the node as argument and pass the node to it in `TryBindAliasExpression`

smitpatel:
Looks good.

tuespetre:
:up::date:

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -612,22 +608,10 @@ var boundExpression
...
-            {
-                expression
-                    = TryBindParentExpression(
-                        _queryModelVisitor.ParentQueryModelVisitor,
-                        qmv => qmv.BindMethodCallExpression(methodCallExpression, CreateAliasedColumnExpressionCore));
-            }
-
-            return expression
-                   ?? _queryModelVisitor.BindMethodToOuterQueryParameter(methodCallExpression);
+            return TryBindAliasExpression(methodCallExpression, (e, visitor, binder)
```
smitpatel: e -> expression

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -646,7 +630,7 @@ protected override Expression VisitMember(MemberExpression expression)
             {
                 var newExpression = Visit(expression.Expression);
 
-                if (newExpression != null
+                if (newExpression != null
```
smitpatel: Format the doc.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +647,78 @@ var newMemberExpression
...
-
-            if (aliasExpression == null
-                && _bindParentQueries)
-            {
-                aliasExpression
-                    = TryBindParentExpression(
-                        _queryModelVisitor.ParentQueryModelVisitor,
-                        qmv => qmv.BindMemberExpression(expression, CreateAliasedColumnExpressionCore));
-            }
+            return TryBindAliasExpression(expression, (e, visitor, binder)
```
smitpatel: e -> expression

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +647,78 @@ var newMemberExpression
...
+                    => visitor.BindMemberExpression(e, binder))
+                ?? TryBindQuerySourcePropertyExpression(expression)
+                ?? _queryModelVisitor.BindMemberToOuterQueryParameter(expression);
+        }
 
-            if (aliasExpression == null)
+        private Expression TryBindQuerySourcePropertyExpression(MemberExpression expression)
```
smitpatel: memberExpression

tuespetre: oops

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
tuespetre: huh, wonder why Visual Studio diff didn't show that

tuespetre:
:up::date:

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +647,78 @@ var newMemberExpression
...
+
+            if (bound != null)
+            {
+                return bound;
             }
 
-            return binder(queryModelVisitor)
-                   ?? TryBindParentExpression(queryModelVisitor.ParentQueryModelVisitor, binder);
-        }
+            var ancestor = _queryModelVisitor.ParentQueryModelVisitor;
```
smitpatel: can we move away from this variable name. ancestor does not specify ancestor of what particularly.
Better name would be outerQueryModelVisitor (since its not direct parent always)

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +647,78 @@ var newMemberExpression
...
 
-            return aliasExpression
-                   ?? _queryModelVisitor.BindMemberToOuterQueryParameter(expression);
+            return null;
         }
 
-        private static AliasExpression TryBindParentExpression(
-            RelationalQueryModelVisitor queryModelVisitor,
-            Func<RelationalQueryModelVisitor, AliasExpression> binder)
+        private Expression TryBindAliasExpression<TExpression>(
```
smitpatel: Do we need generics here? Does a simple expression work?

tuespetre: Generic due to `MemberExpression` vs. `MethodCallExpression`

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
tuespetre: Done (also changed `canBindtoAncestor` to `canBindToOuterQueryModelVisitor`)

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
smitpatel: function name `TryBindMemberOrMethodToAliasExpression`?

smitpatel:
Can you add some test around improved binding? Multi-level queries which can be bound with outer select expression through alias expression and through parameter.

src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -663,81 +647,78 @@ var newMemberExpression
...
-            if (queryModelVisitor == null || queryModelVisitor.RequiresClientProjection)
+            AliasExpression CreateAliasExpression(
+                IProperty property, IQuerySource querySource, SelectExpression selectExpression)
+                => new AliasExpression(
+                    new ColumnExpression(
+                        _relationalAnnotationProvider.For(property).ColumnName,
+                        property,
+                        selectExpression.GetTableForQuerySource(querySource)));
+
+            var bound = binder(sourceExpression, _queryModelVisitor, (property, querySource, selectExpression) =>
```
smitpatel: boundExpression

tuespetre:
@smitpatel working on tests.

tuespetre:
:up::date: Added some tests and also a fix for an error that was revealed by the tests. Did in a separate commit so you can see the isolated change.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1605,6 +1605,8 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
 
         private const string OuterQueryParameterNamePrefix = @"_outer_";
 
+        private readonly Dictionary<string, Expression> _boundParameters = new Dictionary<string, Expression>();
```
smitpatel: Can we name it injectedParameters?

tuespetre: ✔️

tuespetre:
🏁 done

smitpatel:
@tuespetre - QueryInMemoryTests are failing.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1668,5 +1689,26 @@ var parameterWithSamePrefixCount
...
                 Expression.NewArrayInit(typeof(object), parameterValueExpressions));
         }
+
+        /// <summary>
+        ///     Lifts the outer parameters injected into a subquery into the query
+        ///     expression that is being built by this query model visitor, so that
+        ///     the subquery can be lifted.
+        /// </summary>
+        /// <param name="subQueryModelVisitor"> The query model visitor for the subquery being lifted. </param>
+        public void LiftInjectedParameters(RelationalQueryModelVisitor subQueryModelVisitor)
```
smitpatel: add virtual, NotNull annotation and null check on parameter

tuespetre:
Oops, hadn't thought to run those before pushing. (What happened to the Travis logs BTW?)

tuespetre:
:up::date: fixed the tests (needed to change from `Region` to `City`)

smitpatel:
Merged via 08bdfc91b7d36616e6994c873a268a2b2fbd6a69

PR #7703, opened 2017-02-24
Query: Add a hook to override single projection or predicate

smitpatel:
@maumar @anpete

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -17,6 +17,7 @@ using Remotion.Linq.Clauses;
 using Remotion.Linq.Parsing;
 
 // ReSharper disable SwitchStatementMissingSomeCases
+// ReSharper disable once CheckNamespace
 namespace Microsoft.EntityFrameworkCore.Query.Sql
 {
     /// <summary>
@@ -210,7 +211,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
                     _relationalCommandBuilder.Append(", ");
                 }
 
-                VisitProjection(selectExpression.Projection);
+                ProcessExpressionList(selectExpression.Projection, GenerateProjection);
 
                 projectionAdded = true;
             }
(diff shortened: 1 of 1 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -210,7 +210,7 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
                     _relationalCommandBuilder.Append(", ");
                 }
 
-                VisitProjection(selectExpression.Projection);
+                VisitJoin(selectExpression.Projection, VisitProjection);
```
maumar: rename VisitJoin to something else the name is confusing a bit

anpete: IIRC, VisitJoin is just an existing private helper method.

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -315,13 +308,26 @@ var newExpression
         }
 
         /// <summary>
-        ///     Visit the projection.
+        ///     Visit a single projection in SQL SELECT clause
+        /// </summary>
+        /// <param name="projection"> The projection expression. </param>
+        protected virtual void VisitProjection([NotNull] Expression projection) => Visit(ApplyOptimizations(projection, searchCondition: false));
```
anpete: Rename these two to GenerateX, which is the pattern we are using for these kinds of virtual methods - Visit methods should deal with handling a specific node type.

src/Microsoft.EntityFrameworkCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
maumar: Yeah, was thinking its a good opportunity to rename it now as it doesn't have anything to do with JoinExpression

smitpatel: Suggest a new name if you want change 😝

maumar: how about ProcessExpressionList?

smitpatel:
Updated.

PR #7709, opened 2017-02-24
Remove QueryFlattener and split OptimizeJoinClause

tuespetre:
This PR removes `QueryFlattener`, `QueryFlattenerFactory`, and `IQueryFlattenerFactory`, and splits `OptimizeJoinClause` into dedicated `TryFlattenJoin` and `TryFlattenGroupJoin` methods. A `TryFlattenSelectMany` and a `TryFlattenGroupJoinDefaultIfEmpty` method are also introduced.

I feel these changes are justified for the following reasons:

- `QueryFlattener` is just not feasible as a standalone component, it's a kind of 'magic box' that isn't reusable.
- `OptimizeJoinClause` already had a lot going on in it, including a huge code path that only existed for the group join scenario. The combination felt really forced and even though it was `protected`, it (like `QueryFlattener`) didn't really seem reusable or extendable.
- To add to the second point, future optimizations to materialization requirements will introduce yet more context-specific code for each of the join/group join scenarios. It will be much easier and much more maintainable to have them already split up.


tuespetre:
I'm going to come back to this later.

Issue #629, opened 2014-09-04
Seed Data


bricelam:
In EF6, this was accomplished using `DbMigrationsConfiguration.Seed()`, but migrations configurations don't exist in EF7.

- [x] Add `EntityTypeBuilder.SeedData`
- [x] Add support for shared table seeding
- [x] Add support for owned types seeding
- [x] Add seed data validation
- [x] Reuse existing code by delegating to `IStateManager`, `IEntityMaterializerSource` and `SharedTableEntryMap`
- [ ] ~~Documentation:~~ https://github.com/aspnet/EntityFramework.Docs/issues/509
- [ ] ~~Add support for navigation seeding:~~ https://github.com/aspnet/EntityFrameworkCore/issues/10000
- [ ] ~~Add `EntityTypeBuilder.SeedData` sugar that takes an entity and a property bag for shadow properties:~~ https://github.com/aspnet/EntityFrameworkCore/issues/9999
- [ ] ~~Perf: batch the operations:~~ https://github.com/aspnet/EntityFrameworkCore/issues/9997
- [ ] ~~Perf: skip optimistic concurrency checks:~~ https://github.com/aspnet/EntityFrameworkCore/issues/9998

hbopuri:
How is Seed method handled in EF7?

protected override void Seed(MyContext context){}

rowanmiller:
@hbopuri - There isn't a first class API for this at the moment (though this work item is tracking adding one in the future). Currently you would just write code in app startup to handle seed data.

hbopuri:
@rowanmiller wasn't  _*override seed*_  part of earlier EF versions? is EF7 a upgraded version of earlier versions or a completely newly designed from scratch? 

[HaBo Rocks](https://www.facebook.com/HaBoRocks)
[Harsha Bopuri](http://www.harshabopuri.com/)

rowanmiller:
@hbopuri - It's a major version and has breaking changes in it. These breaking changes are larger than we would typically do in a release, but you can read more about the reasoning behind what we are doing here http://blogs.msdn.com/b/adonet/archive/2014/10/27/ef7-v1-or-v7.aspx.

heathyates:
@rowanmiller Will this feature ship with the RTM? And totally agree with @giggio suggestion.

rowanmiller:
@heathyates - Not sure if we'll get a seed method equivalent into initial RTM. It is super easy to write the login in your app start code in the meantime.

[...]

PR #7712, opened 2017-02-24
[WIP] Data migration operations

tinchou:
See #629 

This adds data migration operations:

* InsertRowsOperation
* DeleteRowsOperation
* UpdateRowsOperation


tinchou:
@bricelam maybe you can take a look at these few changes I've made so far and let me know if there's anything to change!

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -948,16 +996,21 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
             => model?.GetEntityTypes().Where(
                 t => Dependencies.Annotations.For(t).TableName == tableName && Dependencies.Annotations.For(t).Schema == schema);
 
+        protected virtual IEnumerable<IProperty> FindProperties(
```
smitpatel: Method name should not be find.

tinchou: changing to `GetProperties`

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -948,16 +996,21 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
+            [CanBeNull] IModel model,
+            [CanBeNull] string schema,
+            [NotNull] string tableName)
+            => FindEntityTypes(model, schema, tableName)?.SelectMany(e => e.GetDeclaredProperties());
```
bricelam: Per the comment in FindProperty, this method shouldn't be needed.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/InsertRowsOperation.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Migrations.Operations
+{
+    public class InsertRowsOperation : MigrationOperation
```
bricelam: Maybe `InsertRowsOperation<T>`? Then `Rows` could be of `T`. But this might cause issues when only some of the values are specified on some rows...

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/InsertRowsOperation.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public virtual string Table { get; [param: NotNull] set; }
+        public virtual string Schema { get; [param: CanBeNull] set; }
+        public virtual object[] Rows { get; [param: NotNull] set; }
```
bricelam: `IEnumerable<object>` or maybe `IReadOnlyList<object>`

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlServerMigrationSqlGeneratorTest.cs
```diff
@@ -984,6 +984,22 @@ public virtual void RenameTableOperation()
...
 
+        public override void InsertRowsOperation()
+        {
+            base.InsertRowsOperation();
+
+            Assert.Equal(
+                "INSERT INTO [People]" + EOL +
+                "    ([Id], [Name])" + EOL +
+                "VALUES" + EOL +
+                "    (1, N'Daenerys Targaryen')," + EOL +
```
bricelam: Hmm, do most databases support adding multiple rows in a single `INSERT` statement? Is it in the SQL standard?

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/MigrationSqlGeneratorTestBase.cs
```diff
@@ -562,6 +562,26 @@ public virtual void SqlOperation()
...
                 });
 
+        [Fact]
+        public virtual void InsertRowsOperation()
+            => Generate(
+                modelBuilder =>
+                {
+                    modelBuilder.Entity("People").Property<int>("Id").IsRequired();
+                    modelBuilder.Entity("People").Property<string>("Name").IsRequired();
+                },
```
bricelam: Don't specify a model.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -624,6 +625,53 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
+            var columns = FindProperties(model, operation.Schema, operation.Table).Select(p => Dependencies.Annotations.For(p).ColumnName);
+            using (builder.Indent())
+            {
+                builder
+                    .Append("(")
+                    .AppendJoin(", ", columns.Select(Dependencies.SqlGenerationHelper.DelimitIdentifier))
+                    .AppendLine(")");
+            }
+
+            builder.AppendLine("VALUES");
```
bricelam: (And the column names on their own line.)

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandListBuilder.cs
```diff
@@ -46,6 +46,31 @@ public virtual MigrationCommandListBuilder Append([NotNull] object o)
...
+                        last = iter.Current;
+                    }
+
+                    _commandBuilder.Append(last);
+                }
+            }
+
+            return this;
+        }
+
```
tinchou: if `o` is large, the we're back to having a long string (and that's why we use a builder in the first place)

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -624,6 +625,53 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
+            Check.NotNull(operation, nameof(operation));
+            Check.NotEmpty(operation.Rows, nameof(operation.Rows));
+            Check.NotNull(model, nameof(model));
+            Check.NotNull(builder, nameof(builder));
+
+            builder
+                .Append("INSERT INTO ")
+                .AppendLine(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema));
+
+            var columns = FindProperties(model, operation.Schema, operation.Table).Select(p => Dependencies.Annotations.For(p).ColumnName);
```
tinchou: I discussed this with Andriy and we arrived at the same conclusion. Then things like the model can be null since we're not using it.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -624,6 +625,53 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
             EndStatement(builder, operation.SuppressTransaction);
         }
 
+        protected virtual void Generate(
+            [NotNull] InsertRowsOperation operation,
+            [NotNull] IModel model,
+            [NotNull] MigrationCommandListBuilder builder)
+        {
+            Check.NotNull(operation, nameof(operation));
+            Check.NotEmpty(operation.Rows, nameof(operation.Rows));
```
tinchou: An insert operation with no rows should just be empty. Then what's the point of allowing that? Also, I'll be using `operation.Keys[0].GetType().GetRuntimeProperties()` to get the column names, and I don't want to hit an index error.

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/MigrationSqlGeneratorTestBase.cs
```diff
@@ -562,6 +562,26 @@ public virtual void SqlOperation()
...
+                new InsertRowsOperation
+                {
+                    Table = "People",
+                    Rows = new[] {
+                        new Person { Id = 1, Name = "Daenerys Targaryen" },
+                        new Person { Id = 2, Name = "John Snow" },
+                        new Person { Id = 3, Name = "Arya Stark" },
+                        new Person { Id = 4, Name = "Harry Strickland" },
+                        new Person { Id = 5, /* No name */ }
```
bricelam: I'm not sure using anonymous types will work here. What do we do for column names with spaces in them? On `CreateTable` we handle it like this:
```C#
CreateTable(
    columns: x => new
    {
        FirstName = x.Column(name: "First Name")
    });
```
...but that might be a bit cumbersome here. maybe we should just do an array of arrays:
```C#
 new InsertRowsOperation
{
    Table = "People",
    Columns = new[] { "Id", "First Name", "Last Name" },
    Values = new[]
    {
        new object[] { 2, "John", "Snow"}
    }
}
```

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
tinchou: It seems to be:

```
SELECT FROM [Table] ([Column]....)
VALUES (...), (...), ....
```
I'd prefer to have values in different lines, but let's be consistent.
 
https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Update/UpdateSqlGenerator.cs#L270

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/InsertRowsOperation.cs
tinchou: Just tried and it won't compile because it can't find a best type for an implicit array. I had to do:

```cs
new[]
{
    new { Id = 1, FirstName = "Brice", LastName = default(string) },
    new { Id = 2, FirstName = default(string), LastName = "Lambson" }
}
```

Another solution is to have `new object[] {...}`, but then we'd need to do an union on all object property names. What would you recommend?

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/InsertRowsOperation.cs
tinchou: I was following the pattern on [other operations](https://github.com/aspnet/EntityFramework/blob/dev/src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/AddForeignKeyOperation.cs#L13). Should I change it anyway?

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlServerMigrationSqlGeneratorTest.cs
tinchou: [It's SQL92](https://en.wikipedia.org/wiki/Insert_(SQL)#Multirow_inserts), but I can change it to multiple inserts if that's better.

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/MigrationSqlGeneratorTestBase.cs
tinchou: Maybe use a Dictionary? `Values` could also be a tuple to get strong-typing across objects.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationCommandListBuilder.cs
bricelam: The builder isn't used here for perf. This isn't a critical path. It just makes the API easier to extend. I don't think the benefits outweigh the cost of maintaining this method.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
bricelam: Sorry, I mean return on empty instead of throw. Imagine the following when `env` is *prod*.
```C#
var users = new Rows();
if (env == "dev")
    users.Add("test", "");
else if (env == "test")
    users.Add("test", "password");

mb.Insert(users);
```

src/Microsoft.EntityFrameworkCore.Relational/Migrations/MigrationsSqlGenerator.cs
bricelam: I'd be OK putting them on their own line when there's more than one.
```SQL
INSERT INTO T (C)
VALUES ('V1'),
       ('V2');
```

bricelam: (or whatever varient we agree on)

bricelam: My rationale being that this SQL is more visible to users and we don't have to worry about hitting length limits.

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/InsertRowsOperation.cs
bricelam: Hmm, nevermind then. ☺

src/Microsoft.EntityFrameworkCore.Relational/Migrations/Operations/InsertRowsOperation.cs
bricelam: See my comment below. We may not be able to use anonymous objects.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/SqlServerMigrationSqlGeneratorTest.cs
bricelam: No, I like this better. Providers can override if it won't work. In general we just aim to generate standard-compliant SQL here.

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/MigrationSqlGeneratorTestBase.cs
bricelam: We'll have to think about it more. Maybe there's a non-cumbersome way to use anonymous objects and non-CLR-identifier column names.

bricelam: Maybe an optionally chained `Map` method:
```C#
.Map(
    x =>
    {
        x.Property(e => e.FirstName). HasColumnName("First Name");
    });
```

tinchou: something like https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types lol

tinchou: How about

```cs
// NOTE: we'll need System.ValueTuple from Nuget
ColumnNames = new[] { "Id", "Full Name" },
Keys = new [] {
    (2, "John Snow"),
    (3, "Arya Stark")
}
```

tinchou:
This *could* be merged, but I think we need to discuss:

1. Operation property names
1. What's the API for column names with spaces

Thanks!

src/Microsoft.EntityFrameworkCore.Relational.Specification.Tests/MigrationSqlGeneratorTestBase.cs
bricelam: I don't think we can use tuples. We need to work in places where C# 7 won't be available.

tinchou:
@bricelam I added some weird LINQ and helper methods, especially for working with multidimensional arrays because it's not that easy. It may not be that idiomatic or easy to understand, would you mind checking? e.g. https://github.com/aspnet/EntityFramework/pull/7712/files#diff-feb942230d2c5a19a87bd4f62adb60a4R764

We can go through the code together on Monday if you want.

src/EFCore.Design/Migrations/Design/CSharpMigrationOperationGenerator.cs
```diff
@@ -1378,6 +1379,62 @@ protected virtual void Generate([NotNull] SqlOperation operation, [NotNull] Inde
...
+                {
+                    var rowCount = operation.Values.GetLength(0);
+                    var valueCount = operation.Values.GetLength(1);
+                    for (var i = 0; i < rowCount; i++)
+                    {
+                        builder
+                            .Append("{ ")
+                            .Append(string.Join(
+                                ", ",
+                                Enumerable.Range(0, valueCount).Select(j => _code.UnknownLiteral(operation.Values[i, j]))))
```
bricelam: You already know this, but it seems silly to me not to just do this:

```C#
for (var j = 0; j < valueCount; j++}
{
    if (j != 0)
        build.Append(", ");

    builder.Append(_code.UnknownLiteral(operations.Values[i, j]));
}
```

src/EFCore.Relational.Specification.Tests/MigrationSqlGeneratorTestBase.cs
```diff
@@ -562,6 +562,81 @@ public virtual void SqlOperation()
                     Sql = "-- I <3 DDL"
                 });
 
+        [Fact]
+        public virtual void InsertRowsOperation()
+            => Generate(
```
bricelam: I discourage using expression-bodied members in tests since it ruins **Collapse to Definitions**. Which I use often when writing and running tests.

src/EFCore.Relational/Migrations/MigrationBuilder.cs
```diff
@@ -679,5 +679,25 @@ public virtual AlterOperationBuilder<AlterDatabaseOperation> AlterDatabase()
 
             return new OperationBuilder<SqlOperation>(operation);
         }
+
+        public virtual OperationBuilder<InsertRowsOperation> InsertRows(
```
bricelam: Just `Insert`?

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -622,6 +626,148 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
+                    builder
+                        .AppendLine(",")
+                        .Append("       ");
+                }
+
+                builder
+                    .Append("(")
+                    .Append(string.Join(
+                        ", ",
+                        row.Select(v => Dependencies.SqlGenerationHelper.GenerateLiteral(v))))
```
bricelam: (same comment as above)

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -622,6 +626,148 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
+            foreach (var row in EnumerateValues(operation.Keys))
+            {
+                if (firstRow)
+                {
+                    firstRow = false;
+                }
+                else
+                {
+                    builder
+                        .AppendLine(" OR")
```
bricelam: This might get confusing with composite keys. Maybe we should generate separate `DELETE` statements. Does the Update Pipeline batch like this?

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -622,6 +626,148 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
+            foreach (var row in EnumerateValues(operation.Values))
+            {
+                builder
+                    .Append("UPDATE ")
+                    .AppendLine(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
+                    .Append("SET ");
+
+                var assignments = GenerateEquals(
+                    columns.Where(c => !operation.KeyColumns.Contains(c)),
+                    row.Where((r, i) => !operation.KeyColumns.Contains(columns[i])));
```
bricelam: This operation's structure doesn't feel right. You have to do too much filtering here.

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -622,6 +626,148 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
+                        columns.Where(c => operation.KeyColumns.Contains(c)),
+                        row.Where((r, i) => operation.KeyColumns.Contains(columns[i]))))
+                    .AppendLine(Dependencies.SqlGenerationHelper.StatementTerminator);
+            }
+
+            EndStatement(builder);
+        }
+
+        private string GenerateAnd(IEnumerable<string> columns, IEnumerable<object> row)
+            => $"({string.Join(" AND ", GenerateEquals(columns, row))})";
```
bricelam: Surely there's code in the Update Pipeline that we can be leveraging for a lot of this...

src/EFCore.Relational/Migrations/Operations/DeleteRowsOperation.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Migrations.Operations
+{
+    public class DeleteRowsOperation : MigrationOperation
```
bricelam: Just `DeleteOperation`?

src/EFCore.Relational/Migrations/Operations/DeleteRowsOperation.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public virtual string Table { get; [param: NotNull] set; }
+        public virtual string Schema { get; [param: CanBeNull] set; }
+        public virtual string[] KeyColumns { get; [param: NotNull] set; }
+        public virtual object[,] Keys { get; [param: NotNull] set; }
```
bricelam: `KeyValues`?

src/EFCore.Relational/Migrations/Operations/InsertRowsOperation.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.EntityFrameworkCore.Migrations.Operations
+{
+    public class InsertRowsOperation : MigrationOperation
```
bricelam: Just `InsertOperation`?

src/EFCore.Relational/Migrations/Operations/UpdateRowsOperation.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.EntityFrameworkCore.Migrations.Operations
+{
+    public class UpdateRowsOperation : MigrationOperation
+    {
+        public virtual string Table { get; [param: NotNull] set; }
+        public virtual string Schema { get; [param: CanBeNull] set; }
+        public virtual string[] KeyColumns { get; [param: NotNull] set; }
+        public virtual string[] Columns { get; [param: NotNull] set; }
+        public virtual object[,] Values { get; [param: NotNull] set; }
```
bricelam: As stated above, this doesn't feel like the right structure. Having the values used for filtering mixed in with the values used for updating may be creating unnecessary work on the consumer.

src/EFCore.Relational/Migrations/Operations/UpdateRowsOperation.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Migrations.Operations
+{
+    public class UpdateRowsOperation : MigrationOperation
```
bricelam: Just `UpdateOperation`?

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -622,6 +626,148 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
+
+        private IEnumerable<string> GenerateEquals(IEnumerable<string> columns, IEnumerable<object> row)
+            => columns.Zip(
+                row,
+                (c, v) =>
+                    Dependencies.SqlGenerationHelper.DelimitIdentifier(c) +
+                    " = " +
+                    Dependencies.SqlGenerationHelper.GenerateLiteral(v));
+
+        private static IEnumerable<IEnumerable<object>> EnumerateValues(object[,] rows)
```
bricelam: `protected`? If providers override, they'll probably want to use this too.

test/EFCore.SqlServer.FunctionalTests/SqlServerMigrationSqlGeneratorTest.cs
```diff
@@ -1111,6 +1111,70 @@ public virtual void SqlOperation_ignores_non_go()
...
+
+        public override void UpdateRowsOperation_simple_key()
+        {
+            base.UpdateRowsOperation_simple_key();
+
+            Assert.Equal(
+                "UPDATE [People]" + EOL +
+                "SET [Full Name] = N'Daenerys Stormborn'" + EOL +
+                "WHERE ([Id] = 1);" + EOL +
+                "UPDATE [People]" + EOL +
```
bricelam: Don't batch by default (call `EndStatement` between rows).

test/EFCore.SqlServer.FunctionalTests/SqlServerMigrationSqlGeneratorTest.cs
```diff
@@ -1111,6 +1111,70 @@ public virtual void SqlOperation_ignores_non_go()
...
+                Sql);
+        }
+
+        public override void UpdateRowsOperation_simple_key()
+        {
+            base.UpdateRowsOperation_simple_key();
+
+            Assert.Equal(
+                "UPDATE [People]" + EOL +
+                "SET [Full Name] = N'Daenerys Stormborn'" + EOL +
```
bricelam: Add a test that show updating multiple columns.

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
bricelam: Well, the parenthesis help with confusion, but it still looks a bit ugly to me. 😉

src/EFCore.Relational.Specification.Tests/MigrationSqlGeneratorTestBase.cs
tinchou: But every single test in this file is written this way. Maybe I can fix all of them or we can open a new issue for clean-up.

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
tinchou: It seems [`AppendDeleteCommand`](https://github.com/aspnet/EntityFramework/blob/f386095005e46ea3aa4d677e4439cdac113dbfb1/src/EFCore.Relational/Update/UpdateSqlGenerator.cs#L128) is called once for each row to be deleted. But it's the same with `AppendUpdateCommand`.

You think the resulting SQL is not clear and we should split the commands, or just fix the code?

```sql
DELETE FROM [People]
WHERE ([Id] = 1 AND [Full Name] = N'Daenerys Targaryen') OR
      ([Id] = 2 AND [Full Name] = N'John Snow');
```

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
tinchou: I'll probably need some help to integrate with that. For now I'm writing my own SQL code as you see above.

tinchou:
Everything should be ready now. I'm only concerned about code repetition, but it doesn't seem to be a main concern given this code only deals with code generation and won't probably be changed much.

src/EFCore.Relational.Specification.Tests/MigrationSqlGeneratorTestBase.cs
bricelam: Fine to leave as is.

src/EFCore.Relational/Migrations/MigrationBuilder.cs
```diff
@@ -679,5 +679,69 @@ public virtual AlterOperationBuilder<AlterDatabaseOperation> AlterDatabase()
...
+            return new OperationBuilder<DeleteOperation>(operation);
+        }
+
+        public virtual OperationBuilder<UpdateOperation> Update(
+            [NotNull] string table,
+            [CanBeNull] string schema = null,
+            [CanBeNull] string[] keyColumns = null,
+            [CanBeNull] object[,] keyValues = null,
+            [CanBeNull] string[] columns = null,
+            [CanBeNull] object[,] values = null)
```
bricelam: This could use some refinement. I don't like that the key and column values are split in this API. It makes it less readable.

PR #7713, opened 2017-02-24
Lift more subqueries

tuespetre:
This PR depends on #7709.

It implements a ProjectionShaper that LiftSubQuery can use to... well, lift more subqueries. It also makes a few methods on EntityQueryModelVisitor `protected` instead of `private` so that the `RelationalQueryModelVisitor` can use `IntroduceTransparentScope` while flattening a group-join-default-if-empty query. This change was necessary because revisiting the join clause to create the new scope would wreak havoc when the join clause contained a subquery that could now make use of the ProjectionShaper. Changes to the Shaper classes/interface were made as necessary.


maumar:
I think @anpete has something similar in the works

tuespetre:
Cool, @anpete let me know, I will close this if you got it

tuespetre:
I will come back to this later.

anpete:
Hey @tuespetre, sorry for the delay. I like your approach here, can I re-open? 😄

tuespetre:
Sure!

anpete:
@tuespetre Great! Can you rebase, we just renamed the world.

src/Microsoft.EntityFrameworkCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1039,6 +761,28 @@ var newExpression
                                 subSelectExpression)
                             .Visit(subQueryModelVisitor.Expression);
 
+                    if (newExpression is MethodCallExpression methodCallExpression
```
anpete: What about doing this in RelationalQueryModelVisitor.VisitSelect? I think we get the goodness in more places if we do it there.

src/Microsoft.EntityFrameworkCore.Relational/Infrastructure/ServiceCollectionRelationalProviderInfrastructure.cs
```diff
@@ -74,7 +74,6 @@ public static void TryAddDefaultRelationalServices([NotNull] ServiceCollectionMa
                 .TryAddScoped<IConditionalRemovingExpressionVisitorFactory, ConditionalRemovingExpressionVisitorFactory>()
                 .TryAddScoped<ICompositePredicateExpressionVisitorFactory, CompositePredicateExpressionVisitorFactory>()
                 .TryAddScoped<IIncludeExpressionVisitorFactory, IncludeExpressionVisitorFactory>()
-                .TryAddScoped<IQueryFlattenerFactory, QueryFlattenerFactory>()
```
anpete: We don't need to change this, right?

anpete: My bad, reviewing this in GH is awkward; will pull the changes locally after rebase.

tuespetre:
Almost done 🌷

tuespetre:
:up::date:

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1048,6 +770,28 @@ var newExpression
                                 subSelectExpression)
                             .Visit(subQueryModelVisitor.Expression);
 
+                    if (newExpression is MethodCallExpression methodCallExpression
+                        && methodCallExpression.Method.MethodIsClosedFormOf(LinqOperatorProvider.Select))
+                    {
+                        var shapedQuery = (MethodCallExpression)methodCallExpression.Arguments[0];
+
+                        if (IsShapedQueryExpression(shapedQuery))
```
anpete: Hey, you see my earlier comment on this? Can we do this in RQMV.VisitSelect so we can get this optimization at the top level?

tuespetre:
My bad! I will look into that now.

anpete:
@tuespetre Hey, no problem, thanks!

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
tuespetre: Looks like that would make a lot of stuff angry. `QuerySourceUpdater` can't work as intended, top-level seems to complain about an invalid `queryContext` parameter reference, some include-related things fail, and so forth.

anpete: Some of those are caused by compiling the newShaper below. Once compiled, the delegate can have two issues:

1) Any qsres that have not been bound yet will never get bound. A solution is to detect this case (there are relatively few) and prevent the optimization.
2) Any references to the queryContext parameter will be invalid because the materializer is no longer connected to the larger expression scope. Solution is to make a new materializer delegate that is additionally parameterized by queryContext, which can then be passed to the delegate in the Shape method.

tuespetre: Good call on both of those. It also looks like the query source for the shaper can just be taken from the 'old' shaper (by making the `QuerySource` property public. Doing these three things so far has resolved most of it; I'm still seeing a few issues related to includes but I'm working on it.

anpete: WRT Includes, it's likely that the AccessorExpression is getting broken. It's fine to disable this optimization when there are Includes as that is being reworked to not use accessor expressions at all.

tuespetre: The accessor expression appears to be fine but the `resultType` of the `methodCallExpression` no longer matches it, i.e. by the time it gets to the Includes, it no longer has access to the transparent identifier to make any use of the accessor expression.

tuespetre: Question: is this really any kind of optimization for top-level queries anyways?

anpete: Yep, the accessor expression needs to be in terms of the new type. I.e. the result of the projection.

tuespetre: @anpete ohhh. Need to update the `QuerySourceMapping` when the shaper is created, then.

anpete: >  is this really any kind of optimization for top-level queries anyways?

Yes, in the sense that we don't need a client Select. Small optimization, though.

> Need to update the QuerySourceMapping when the shaper is created, then.

Yeah, or rewrite it somehow.

tuespetre: Hmmm. Updating the mapping solved the simple cases but not the complex multi-include cases due to the obliterating of access to the transparent identifier (no way to rewrite the mappings for those since there is 'nothing' to map to after that.) Looks like skipping when there are includes is the way to go

tuespetre:
:up::date:

tuespetre:
Whoa, for real? :scream: 

I did not expect this to be merged so quickly! This is going to be KILLER for complex GROUP BY query performance!

PR #7732, opened 2017-02-27
SQLite: Warn when model has schema or sequence

### LINKED ISSUE #6679: SQLite Migrations: Ignore schema
Asked by @VorSiechtIP (or a user with the same name) [on StackOverflow](http://stackoverflow.com/q/39854342/475031):

I have a shared model that will be used to create both SqlServer and SQLite databases.  I added the Table attribute onto each one of my model classes and supplied a schema:

``` C#
[Table("Sites", Schema = Common)]
```

When I migrate the SQLite database using `context.Database.Migrate()`, I get a NotSupportedException error.  I know that SQLite doesn't support schemas, so I was wondering if there was an easy way to ignore the schema attribute during migrate?

bricelam:
Also updates Migrations to ignore EnsureSchema and DropSchema operations (fixes #6679)


bricelam:
👍 Feel free to do something similar.

ErikEJ:
Will do!

Issue #7452, opened 2017-01-20
Query : invalid cast for queries with nullable bool and coalesce


maumar:
```
    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var p1 = new Person { Name = "Eo", IsGold = false };
                var p2 = new Person { Name = "Virginia", IsGold = true };
                var p3 = new Person { Name = "Darrow", IsGold = null };

                ctx.People.AddRange(p1, p2, p3);
                ctx.SaveChanges();
            }

            using (var ctx = new MyContext())
            {
                ctx.People.Select(p => p.IsGold ?? false).ToList();
            }
        }
    }


    public class MyContext : DbContext
    {
        public DbSet<Person> People { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer("Data Source=.;Initial Catalog=Repro7452;Integrated Security=SSPI;");
        }
    }

    public class Person
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public bool? IsGold { get; set; }
    }
```

exception:

System.InvalidCastException: Specified cast is not valid.
   at lambda_method(Closure , ValueBuffer )
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at Microsoft.EntityFrameworkCore.Query.Internal.LinqOperatorProvider.ExceptionInterceptor`1.EnumeratorExceptionInterceptor.MoveNext() in D:\git\EntityFramework\src\Microsoft.EntityFrameworkCore\Query\Internal\LinqOperatorProvider.cs:line 145

We generate the following query plan:

```
TRACKED: False
(QueryContext queryContext) => IEnumerable<bool> _Select(
    source: IEnumerable<ValueBuffer> _ShapedQuery(
        queryContext: queryContext, 
        shaperCommandContext: SelectExpression: 
            SELECT COALESCE([p].[IsGold], 0) AS [Coalesce]
            FROM [People] AS [p]
        , 
        shaper: ValueBufferShaper
    )
    , 
    selector: (ValueBuffer p) => (bool) object p.get_Item(0)
)
```

Problem is that COALESCE([p].[IsGold], 0) comes back as int and we are expecting bool - we should probably compensate in SQL

```sql
[...]

PR #7733, opened 2017-02-27
[SMALL] Fix to #7452 - Query : invalid cast for queries with nullable bool and coalesce

maumar:
Problem was that COALESCE (nullable_bool, 1) results in an int value, but we expect bool instead.
Fix is to add explicit cast around Coalesce that returns boolean in projections. In case of nesting, we only need to do this for the outermost COALESCE.

### DIFF
--- a/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/NullSemanticsQuerySqlServerTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/NullSemanticsQuerySqlServerTest.cs
@@ -5,14 +5,16 @@ using System;
 using Microsoft.EntityFrameworkCore.Specification.Tests;
 using Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.Utilities;
 using Xunit;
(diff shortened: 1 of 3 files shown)


maumar:
@anpete @smitpatel

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/NullSemanticsQuerySqlServerTest.cs
```diff
@@ -13,6 +13,7 @@ public class NullSemanticsQuerySqlServerTest : NullSemanticsQueryTestBase<SqlSer
         public NullSemanticsQuerySqlServerTest(NullSemanticsQuerySqlServerFixture fixture)
             : base(fixture)
```
smitpatel: Don't you need `testOutputHelper` in constructor?

maumar: good catch, updated

maumar:
new version up

Issue #7692, opened 2017-02-23
Query : add option to ExpressionPrinter to highlight non-reducible nodes


maumar:
Sometimes when debugging query plan for dangling qsre (and other non-reducible nodes) it's hard to spot them. We could add option to ExpressionPrinter to highlight their position in the query plan for easier debugging

maumar:
Fixed in 5ef0f31d7baf2591d90da76e44a66b7f57335775

PR #7737, opened 2017-02-28
[TINY] Fix to #7692 - Query : add option to ExpressionPrinter to highlight non-reducible nodes

maumar:
This is purely to help with debugging non-reducible nodes in the final query plan. Sometimes they are hard to notice.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
+++ b/src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
@@ -47,6 +47,11 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
             { ExpressionType.Or, " | " }
         };
 
+        private bool _highlightNonreducibleNodes;
+
+        private const string HighlightLeft = " ---> ";
+        private const string HighlightRight = " <--- ";
+
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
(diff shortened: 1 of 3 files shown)


maumar:
@smitpatel @anpete

src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
```diff
@@ -467,7 +491,8 @@ protected override Expression VisitMemberInit(MemberInitExpression node)
                 var assignment = node.Bindings[i] as MemberAssignment;
                 if (assignment != null)
                 {
-                    _stringBuilder.Append(assignment.Member.Name + " = " + Visit(assignment.Expression));
+                    _stringBuilder.Append(assignment.Member.Name + " = ");
```
maumar: small, unrelated bugfix

maumar:
new version up

src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
```diff
@@ -119,11 +124,30 @@ protected virtual void AppendLine([NotNull] string message = "")
...
+            return PrintInternal(expression, removeFormatting, characterLimit, false);
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual string PrintDebug(Expression expression, bool highlightNonreducibleNodes = true)
+        {
+            return PrintInternal(expression, false, null, highlightNonreducibleNodes);
```
smitpatel: use parameter names

smitpatel:
On unrelated note, also fix the bug in query model printer where the "," in order by is printed incorrectly when there are multiple ordering.

src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
maumar: problem here is that highlightNonreducibleNode comes last, so the call would have to look:
```
PrintInternal(expression, removeFormatting: false, characterLimit: null, highlightNonreducibleNodes: highlightNonreducibleNodes);
```

will add named param for the Print(..) method tho

smitpatel: It is not a problem 😛 
The way you have written in above comment is right.

maumar:
@smitpatel new version up

src/Microsoft.EntityFrameworkCore/Query/Internal/ExpressionPrinter.cs
maumar: yeah, it looks ugly tho. Also feels bit redundant to me, since the method that is being called is directly below. But I guess I can add this for consistency with the other method

smitpatel: For future references - we use commented parameter names if they are not at the end of parameters.

Issue #7368, opened 2017-01-06
Query: Added support for User Defined Scalar Functions.


pmiddleton:
I have added support for user defined scalar functions. 

My fork can be found at https://github.com/pmiddleton/EntityFramework in the udf branch.

I am looking for feedback on this feature and if there is interest in it I will submit a pull request.

This update has three main features.

* User defined scalar functions. #4319
* The EntityFunctions feature from EF 6 #2850
* The ability to intercept any dotnet method call and translate it into a function call in the EF store.

## Implementation details

I have used a different integration approach with Re-linq vs what EF Core is presently doing. Rather than performing method translation in the VisitMethodCall on the SqlTranslatingExpressionVisitor class, I have moved the translation to earlier in the transformation pipeline by useing an implementation of IExpressionTransformer<MethodCallExpression> which is called when Re-linq parses the original query. This transformer replaces method calls with DbFunctionExpressions. The DbFunctionExpressions are then translated into SqlFunctionExpressions by the SqlTranslatingExpressionVisitor.

By moving the method transformation to earlier in the Re-linq process, and introducing the DbFunctionExpression, we are now able to translate method calls inside from clauses. This is not possible in the current implementation. This ability will allow us to use functions as data sources and therefore implement Table Valued Functions for database which support this feature. 

There is a working proof of concept for this feature inside of the TVF branch of my repository. The item holding up that feature is missing support in EF Core for representing the return results of a TVF. The proof of concept registers all TVF results objects as Entities. This is not the correct solution, but it proves out that the approach for TVF code will work. What is needed to finalize the code is a lightweight representation of TVF results. The wip complex type branch appears to be a promising approach, but the TVF results are not tied to a single Entity, so something closely related to complex types is needed. Discussion needs to happen around the proper way to represent the results of TVF in the model.

## Function Setup

Methods can be marked as being translatable in two ways. For user defined functions you can use the DbFunctionAttribute. You can also use the fluent api method DbFuction which is located on the ModelBuilder class. For intercepting method calls on existing classes or adding a custom EntityFunction extension method you need to use the fluent api.

	
## User defined function
	
To create a UDF you just need to define a dotnet function which has the same signature as the database function. You then register the dotnet method using either the DbFunctionAttribute or the fluent api. You can then use that function in your queries. Calls to it will be translated to an equivalent call in the database. 
	
[...]

PR #7739, opened 2017-02-28
Add support for user defined scalar functions

pmiddleton:
This PR adds support for user defined scalar functions.  See #7368.  

@divega @roji - per the conversation in #2850 I have made some modifications and broken my changes into multiple PRs. This is the first part.  

This PR contains the api used to define UDFs and the back end changes needed to execute them.

From my original code I have pulled out the EFFunction extension methods and the changes which converted the existing supported scalar methods from using IMethodTranslator to the new API.  Those changes will come in later PRs.


anpete:
@pmiddleton I'm going to start looking at this (thanks BTW!), can you rebase to latest dev?

pmiddleton:
@anpete I have rebased to latest dev.  I look forward to your feedback.  

There is one implementation area specifically I am interested in your thoughts on.

As it stands with this PR there are two different paths which functions can take through the translation process.  All custom functions with be converted into DbFunctionExpression during Relinq query parsing and later to SqlFunctionExpression as the expression tree is walked.  Any built in functions will continue to be converted directly to SqlFunctionExpressions via IMethodCallTranslator during expression parsing.

The reason I went the DbFunctionExpression route is that it allows converting function calls which are a part of a from clause.  This will allow support of Table Value Functions as query sources.  (I have a POC of this working)

My original cut of this feature included converting all of the existing IMethodCallTranslator's to use the new DBFunction creation api defined in this PR.  The outstanding question I have is whether it is worth doing that conversion, and deprecating IMethodCallTranslator, in order to get the code down to a single function translation path.

pmiddleton:
@anpete - Have you had a chance to take a look at this PR yet?

anpete:
@pmiddleton It is next on my todo list 😄

anpete:
@pmiddleton Thanks for submitting this thorough PR! Some initial feedback/thoughts:

- I don't think we need to add functions to Core. In our new architecture, it seems like these should be implemented in relational as annotations. Similarly, the Code First APIs can be realized using our provider extension method pattern.
- In line with that, I think we can then have relational use these annotations to create and register handlers into our existing method call translator system.
- So what we end up with is an easy way to create custom method translators via fluent API/annotations, but one that integrates with our existing translation system.

Let me know what you think.

pmiddleton:
* I put things into Core to allow support in both relational and non-relational data sources.  Raven, Mongo, Hadoop all support some form of UDF.  Is better to try to come up with a shared solution for relational and non-relational in Core, or will that cause too many limitations for each system type that it is better to go with separate implementations?  I can see arguments both ways.

* In regards to tying into the existing translation system.  That was the first approach I looked at and the problem I ran into was adding support for Table Value Functions.  The implementation became more straight forward when TVFs were translated into DbFunctionExpressions earlier in the translation process. You don't have to deal with trying to convert method calls in the MainBody and AdditionalFromClauses. That being said if we want to remove code from Core then this warrants another look as well.

* Another question driving the translation timing is whether there are any advantages to knowing a function is translatable, vs needing client eval, earlier in the query transformation process.  

What are your thoughts on the non-relational support?  Would that be something better left to a non-relational base layer vs Core?

anpete:
1) I think we would go with separate implementations just for simplicity and because we don't really understand what we need for functions on those other providers. A related point is that in the PR DbFunction has a Schema property, but EF Core is very careful to keep this as a relational concept. I also think it will just greatly simplify the impl, without really taking away any of the benefits of the new APIs you are proposing.
2) I'm happy to help where I can if you hit any issues integrating with the current method translation extensibility. I think it would be fine to scope-out TVFs for this initial support, too. 
3) Not really, either the method can be translated during SQL translation, or not. If not, we will compile the call into the tree directly, which is nice if the CLR method can have an impl. because it enables testing against im-memory. That looked to be what you were going for in DbFunctionUnwindingExpressionVisitor.

pmiddleton:
I will start take a look at these changes.  I have a bunch of free time this weekend.

I might have to hit you up with questions on the annotations and the api extension point, but I'm sure I can find a good example in the codebase to follow.

I will start looking at the TVF tie in with From clauses to make sure we are not backing ourselves into a corner, but support for materializing non entity types is needed before the feature can be finished.

pmiddleton:
@anpete

I am working through migrating the code to live in the Relational project and converting the metadata classes to use annotations as discussed.  I am using the existing relational Sequence class as an example to follow for the metadata classes.

I don't understand why Sequence is internally serializing it's data before storing it in the model.  What advantage does that give vs just storing the sequence object directly?  Should this serialization approach be used for other items in Relational which store themselves in annotations on the model?  

@ajcvickers  - you added the serialization code so can you help me understand why this was done?

Thanks!

ajcvickers:
@pmiddleton Basically so that the model can be serialized and then read back. Specifically, the model snapshot, used by migrations, is a serialized form of the model. However, if the code that creates the snapshot understands the annotation and can write appropriate C# code, then this is better and the annotation can be any type. I think we added this for Sequence, so it may be that sequences don't need to be stored like this anymore. Likewise, any new annotations need to be either ignored by the snapshot (if they don't impact anything migrations would do) or serialized either by being stored as a simple type (e.g. string) or by having special code that understands them.

pmiddleton:
@anpete - Quick question before I end up writing a bunch of code that I might  not need to.

As part of the udf setup process I have a convention which processes registered dbFunctions by setting up name, arguments, etc.  In my original code I added a new property to ConventionSet to handle conventions for database functions.  I don't see anything in the relational project for extending ConventionSet in this way.  Am I missing something?  I'm trying to avoid extending ConventionSet if I can :smiley:

ajcvickers:
@AndriySvyryd to respond to last question.

AndriySvyryd:
@pmiddleton You would need to add `IModelAnnotationSetConvention` in Core similarly to `IEntityTypeAnnotationSetConvention` and the implementation would filter on the annotation name like `SqlServerMemoryOptimizedTablesConvention`

pmiddleton:
@anpete I'm looking for your thoughts on the integration with the current method translation for UDFs.  I'm debating a design choice I made.

Currently classes which derive from RelationalCompositeMethodCallTranslator are loaded into DI as singletons.  In order to perform method translation RelationalCompositeMethodCallTranslator now needs the current model in order to have access to the UDF metadata stored in the annotations.  

I ended up replacing IMethodCallTranslator on RelationalCompositeMethodCallTranslator with a new interface that has a translate method which takes the IModel as a parameter.  This works, but it involved creating a new interface which providers have to update their derived classes to use.

I'm debating if it would be better to switch RelationalCompositeMethodCallTranslator to have a Scoped lifetime and DI the IModel.  Thoughts?

I have everything up and running with the new approach.  I am currently in process of getting the remaining unit tests moved over and working.

anpete:
Great to hear that you are making progress!

I'm actually OK with either approach. Another option would be to add the IModel parameter to IMethodCallTranslator.Translate itself. 

cc @ajcvickers for input.

ajcvickers:
@pmiddleton @anpete I would probably err on the side of making the service scoped and having the IModel injected. It seems cleaner in this case. But I don't feel strongly.

pmiddleton:
@anpete 

I am getting close to being done.  I have made my changes on a different branch and I think it makes more sense to do a new PR.  Does that work for you or would you prefer I try to rebase this one?

anpete:
@pmiddleton New PR is fine.

PR #7741, opened 2017-02-28
[1.0.4] Design: Only check for migrations assembly mismatch when modifying files

### LINKED ISSUE #7710: Design: Allow script/update etc. to run when context is not in target project, but can be found in services
Listing DbContext's will return contexts that are registered in services, even if they reside in another project. Some commands (such as adding a migration) require the target project, since they generate code files, but others only need a context instance and could run by getting the context from services.

Perhaps we could also consider making this work if there is an IDbContextFactory in the target project.

This will help ASP.NET Core publishing.

cc @vijayrkn @mlorbetske

bricelam:
Resolves #7710


lajones:
:shipit:

PR #7745, opened 2017-02-28
Implementation of a more directed API for provider service registration

ajcvickers:
Issue #7457

Main points:
- Providers don't need to know scope to register--EF now knows the appropriate scope for each service. (Providers can override this, but should only do so very carefully.)
- Providers don't need to know whether to register service as multiple/enumerable service.
- Better error checking for bad registrations
- Special internal code for registering service dependencies types with check for bad registration
- Service re-writing for patch releases is internal


ajcvickers:
@bricelam This is for services for which the provider is defining the contract interface, rather than EF defining the contract interface. As such, we don't know the scoping, etc. and so the provider needs to give full registration details. It could be done by just exposing the ServiceCollectionMap on the builder as a property and having the provider call methods on that, but I prefer this way because it forces the provider to explicitly separate out its own services from its implementation of EF services, and therefore (hopefully) helps the provider writer to do each in the "correct" way.

bricelam:
Agreed. Even if it doesn't technically have to be this shape it does feel like a pit of success.

Issue #214, opened 2014-05-22
Expose model building conventions to applications


rowanmiller:
- [x] Create public interfaces for metadata API to be used by conventions.
- [x] Change `QueryFilter`, `DefiningQuery`, `IsEagerLoaded`, `IsShadowProperty` , `IsIndexedProperty`, `BeforeSaveBehavior` `AfterSaveBehavior` to extension methods.
- [x] Throw when calling `SetField` would change the shadowness of a property or the name of the identifying `MemberInfo`.
- [x] Rename `GetContainingPrimaryKey` to `FindContainingPrimaryKey`.
- [x] Throw when calling `SetField` on a shadow or a field-only property.
- [x] Create public interfaces that the internal model builders implement, e.g. `IConventionModelBuilder`. This will allow cleaner API, without the explicit `ConfigurationSource` parameter.
- [x] Throw instead of silently failing when a non-explicit configuration call is invalid. This would require adding quiz API ('CanSet').
- [x] Make model element references consistent (e.g. `Property` vs `IProperty` vs `string` vs `PropertyInfo`)
- [x] Add provider-specific extension methods for convention metadata interfaces.
- [x] Rename and move `RelationalAnnotations` and `RelationalAnnotationsBuilder` to Core.
- [x] Remove unused methods on provider classes like  `SqlServerPropertyBuilderAnnotations.ColumnName`.
- [x] Rename the `Apply` methods to be more descriptive and make the return types consistent, still allowing to stop the execution of the following conventions.
- [x] Pass `CoreConventionSetBuilderDependencies` to the constructor of all conventions.
- [x] Add convention type for configuring a type as owned and use it to remove ambiguous ownerships.
- [x] Track the model elements as they are modified by conventions, so a call to a convention never returns an element that is no longer in the model.
- [x] Add a way to delay the convention execution.
- [x] Move logic from `ModelCustomizer` to a convention.
- [x] Add `SqlServerValueGenerationStrategy.None` value that can be used to override the model default.
- [x] Filter out the core annotations changes from conventions.
- [x] Remove `PropertyMetadataChanged()` calls and don't cache indexes while the model is mutable.
- [x] Expose conventions as `List` on `ConventionSet`
- [x] Add methods to add/remove a convention to/from all appropriate convention lists.
- [x] Add a `Conventions` property of `ConventionSetBuilder` type to `ModelConfigurationBuilder` with methods to add/remove a convention

sthiakos:
[...]

PR #7747, opened 2017-03-01
Refactor ConventionDispatcher to allow delayed execution

AndriySvyryd:
- Refactor ConventionDispatcher
  - Allow delayed execution
  - Allow tracking the foreign key modified by conventions.
  - When set on ConventionDispatcher ConventionScope captures the conventions to be executed.
  - ConventionBatch sets a ConventionScope and executed the stored conventions when it's run or disposed.
  - When conventions captured in a ConventionScope are executed any triggered conventions are captured in a new ConventionScope to avoid keeping all the convention chain on the stack.
  - This changes the order of convention execution which exposed some issues in code and tests
- Allow specifying the dependend end without specifying the FK properties,
- Throw when the same property is specified several times when adding a key, foreign key or indexes.
- Remove Mock usage from ConventionDispatcherTest

Part of #214


AndriySvyryd:
This change adds a visitor to run conventions, so from this point on @anpete will be able to contribute to the model building code 😄

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ConventionDispatcher.cs
```diff
@@ -28,22 +29,45 @@ public ConventionDispatcher([NotNull] ConventionSet conventionSet)
...
+        /// </summary>
         public virtual InternalEntityTypeBuilder OnEntityTypeAdded([NotNull] InternalEntityTypeBuilder entityTypeBuilder)
         {
             Check.NotNull(entityTypeBuilder, nameof(entityTypeBuilder));
 
+            if (_scope == null)
+            {
+                return RunOnEntityTypeAdded(entityTypeBuilder);
+            }
+
```
ajcvickers: Consider having a ConventionScope (or some sugar API) that doesn't batch to avoid having to do this conditional every place.

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ConventionDispatcher.cs
```diff
@@ -510,5 +884,68 @@ public virtual bool OnPropertyNullableChanged([NotNull] InternalPropertyBuilder
...
+                }
+            }
+
+            public void Dispose()
+            {
+                if (!_ran)
+                {
+                    Run();
+                }
+            }
```
ajcvickers: Is it OK to Run if an exception is thrown and Dispose happens as part of unwinding the stack?

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ConventionDispatcher.cs
```diff
@@ -510,5 +884,68 @@ public virtual bool OnPropertyNullableChanged([NotNull] InternalPropertyBuilder
...
+                _dispatcher = dispatcher;
+                var currentScope = _dispatcher._scope;
+                dispatcher._scope = new ConventionScope(currentScope, children: null);
+                currentScope?.Add(dispatcher._scope);
+            }
+
+            private void Run()
+            {
+                _ran = true;
+                while (true)
```
ajcvickers: > while (true) [](start = 15, length = 13)

Could we avoid an infinite loop here by stopping (or at least warning) if this goes on for a very large number of iterations?

AndriySvyryd: We can throw. I don't think a warning will do much of a difference.
Is 65536 a reasonable limit?

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ConventionDispatcher.cs
AndriySvyryd: Will do

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ConventionDispatcher.cs
ajcvickers: I would hope so...

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ConventionDispatcher.cs
AndriySvyryd: The conventions should be robust enough to not throw even if the model is in a bad state due to an exception. All of our negative test cases work correctly. But in theory we should avoid running them after an exception, though I'd have to figure out how this can be detected.

ajcvickers: I think it's fine to leave as is unless we discover a real issue with it in practice.

src/Microsoft.EntityFrameworkCore/Internal/Reference.cs
```diff
@@ -0,0 +1,53 @@
...
+        public Reference([NotNull] T @object)
+            : this(@object, null)
+        {
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public Reference([CanBeNull] T @object, [CanBeNull] IReferenceRoot<T> root)
```
smitpatel: Is this `CanBeNull` true? Other constructor and setter has notnull

src/Microsoft.EntityFrameworkCore/Metadata/Builders/ReferenceNavigationBuilder.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
+using Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal;
```
smitpatel: sort using in all files

src/Microsoft.EntityFrameworkCore/Internal/Reference.cs
AndriySvyryd: Oops, the other one CanBeNull as well

src/Microsoft.EntityFrameworkCore/Metadata/Builders/ReferenceNavigationBuilder.cs
AndriySvyryd: yup

smitpatel:
:shipit:

smitpatel:
Now conventions are even more magic :trollface:

AndriySvyryd:
@smitpatel I think of it more as rule enforcement for conventions, so they will become regulated magic
Proper Unicorn

src/Microsoft.EntityFrameworkCore/Metadata/Conventions/Internal/ConventionDispatcher.cs
AndriySvyryd: The exception message would be 

> The convention invocations have reached the recursion limit. This is likely an issue in EF Core, please report it.

ajcvickers: Add a fwdlink for where to report it? @rowanmiller Maybe also suggest to include repro code and version of EF being used--we could include the version in this message.

PR #7758, opened 2017-03-01
Add an AddToPredicate method to SelectExpression...

tuespetre:
...to eliminate redundant logic.

Just a little 🍬 🍰 🍭 🍨 thing that centralizes a common operation that may be used more than it is now. @anpete had already done so in `IncludeExpressionVisitor`.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
@@ -1017,7 +1017,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 
             if (innerPredicate != null)
             {
-                Predicate = Predicate == null ? innerPredicate : AndAlso(Predicate, innerPredicate);
+                AddToPredicate(innerPredicate);
             }
 
             return innerJoinExpression;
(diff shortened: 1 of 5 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -1017,7 +1017,7 @@ public virtual PredicateJoinExpressionBase AddInnerJoin([NotNull] TableExpressio
 
             if (innerPredicate != null)
             {
-                Predicate = Predicate == null ? innerPredicate : AndAlso(Predicate, innerPredicate);
+                AddToPredicate(innerPredicate);
```
smitpatel: also make predicate private set only.

tuespetre: I thought of that too but there are too many places that set it for other reasons than combining with an existing.

smitpatel:
Other places where we can change this
`SqlServerQueryModelVisitor`

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
smitpatel: I reviewed usages. Not worth removing the setter at this stage.

tuespetre:
:up::date: to use `AddToPredicate` in `SqlServerQueryModelVisitor`

smitpatel:
Merged via 9903d3c8c643b6a92f7f439063b1cd4d087da9b2

PR #7760, opened 2017-03-01
SQL Server Migrations: Handle GO and (Backslash) utilities statements

### LINKED ISSUE #6747: MigrationBuilder.Sql split on GO statement
I need to run a separate sql script just after the migration Up logic is executed.
The following sample works fine, except the script is creating quite a few Stored Procedures etc which requires the use of GO statements.

However, the MigrationBuilder.Sql does not seem to support the GO statement. I haven't tried splitting into multiple script files but I would rather not do this as I would require a lot.

Is this something which will be supported soon. What would be the best workaround for the moment?

```
private class SomeMigration : Migration
{
    {
        migrationBuilder.Sql(File.ReadAllText(sqlFile));
    }

    {
    }
}
```

bricelam:
Resolves #6747

I found a few issues with this (it's more permissive than it should be with things after the statement), but it's more-or-less the same implementation as EF6, and nobody has reported them yet...

**Providers**: This PR is 99% SQL Server, but it also updates `MigrationsSqlGenerator` to stop tacking on a statement terminator (`;`) to the user's `.Sql()` calls. This will only have a functional impact if you've gone above and beyond the base implementation to add batching to Migrations.

@ErikEJ: You may want to support this in SQL CE too.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/Internal/SqlServerMigrationsSqlGenerator.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/Internal/SqlServerMigrationsSqlGenerator.cs
@@ -3,9 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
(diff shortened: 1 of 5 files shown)


ErikEJ:
Nice new label! (providers-fyi)

src/Microsoft.EntityFrameworkCore.SqlServer/Migrations/Internal/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -766,6 +769,48 @@ protected override void Generate(DropForeignKeyOperation operation, IModel model
...
         }
 
+        protected override void Generate([NotNull] SqlOperation operation, [CanBeNull] IModel model, [NotNull] MigrationCommandListBuilder builder)
+        {
+            Check.NotNull(operation, nameof(operation));
+            Check.NotNull(builder, nameof(builder));
+
+            var batches = Regex.Split(
+                Regex.Replace(operation.Sql, @"\\\r?\n", string.Empty),
+                @"^\s*(GO[ \t]+[0-9]+|GO)(?:\s+|$)",
```
ajcvickers: Not going to attempt to determine if this is correct... 😉

smitpatel: No timeout?

bricelam: Same as EF6... 😇

bricelam: Will add.

bricelam: I had no idea what `(?:` meant and had to look it up.

PR #7697, opened 2017-02-23
Implement GROUP BY

tuespetre:
# I don't know when to quit.

Imgur

--------

This PR (dependent on #7695 and #7696) implements GROUP BY, this time with ~~only about 15 or 16 changed files~~ <ins>less changed files, although still a good number due to test changes</ins> and a very focused approach to **getting it done without a lot of byproduct.**

PR #7767, opened 2017-03-02
Refactor RelationalProjectionExpressionVisitor

tuespetre:
More accurate marking of `RequiresClientProjection` (less false positives), which results in less duplicate visitation of child query models in many cases

Note that the few tests producing larger SQL queries are due to unnecessary subquery pushdown, because now that `RequiresClientProjection` is marked more accurately, `LiftSubQuery` is able to lift them. See the related issue: https://github.com/aspnet/EntityFramework/issues/7766

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
@@ -50,66 +50,98 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         /// <summary>
         ///     Visit a method call expression.
         /// </summary>
-        /// <param name="node"> The expression to visit. </param>
+        /// <param name="methodCallExpression"> The expression to visit. </param>
         /// <returns>
         ///     An Expression corresponding to the translated method call.
         /// </returns>
-        protected override Expression VisitMethodCall(MethodCallExpression node)
+        protected override Expression VisitMethodCall(MethodCallExpression methodCallExpression)
         {
-            Check.NotNull(node, nameof(node));
+            Check.NotNull(methodCallExpression, nameof(methodCallExpression));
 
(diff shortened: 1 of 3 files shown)


tuespetre:
@smitpatel :up::date:, rebased

tuespetre:
@smitpatel @maumar @anpete any chance of looking at this one soon? Waiting on it for #7697 and frequently needing to rebase, etc.

maumar:
@tuespetre At present @smitpatel is working on a significant refactoring of SelectExpression - it's probably best to hold off on the GROUP BY work for now (constant rebasing etc.). In fact, I would hold of on anything that touches SelectExpression - lot of stuff is going to change in that area, referential integrity of SelectExpression trees will be vastly improved and you won't be able to make adhoc changes to ColumnExpression, aliases etc. If you have some ideas for query model optimizations, those would be the best to focus on for now (like the several QM optimizations that you recently submitted - I will look at them as soon as I can)

smitpatel:
From #7950 description
> Investigate visiting NewExpression in RelationalProjectionExpressionVisitor. At present we visit it twice. The only info we get out of first visit is about client projection. We throw away translated expression. (this would fix #7844 )

Just tweaking around VisitNew to avoid visiting twice would fix #7844 and other changes in tests. It seems unnecessary to split out Member/MethodCall visiting part.

smitpatel:
Closing this PR in favour of #8072 
The test `Where_subquery_anon` fails to translate mainly due to `RequireClientProjection` flag being true due to anonymous projection. This blocks lifting of the SelectExpression. The flag is only set in the first visit of NewExpression where we don't utilize the translated expression. We need to call VisitNew method to save the member bindings so that we can lift the select expression and bind with it further. We get better aliases if we visit only once by avoid same work. Also for NewExpression we do not need client evaluation flag. The projection shaper knows how to project them out on top level. And on inner level if their properties being used somewhere then we fail to bind it causing client eval anyway. Setting RequireClientProjection is not needed.

PR #7768, opened 2017-03-02
Generate binding redirects for tests that need it

### LINKED ISSUE #7620: Investigate failing tests that use Roslyn
The tests that compile code are failing on machines with both VS 2015 and VS 2017. They were disabled in 33bf2453f39f8328ef9f0c64ca4f547b0f2e4ee4, but we should investigate the issue and re-enable them. The issue may also go away when we stop using the 4.4.0 contracts.

bricelam:
Resolves #7620

### DIFF
--- a/src/EFCore.Relational.Design.Specification.Tests/Migrations/ModelSnapshotTest.cs
+++ b/src/EFCore.Relational.Design.Specification.Tests/Migrations/ModelSnapshotTest.cs
@@ -111,7 +111,7 @@ namespace Microsoft.EntityFrameworkCore.FunctionalTests.Migrations
 
         #region Model
 
-        [Fact(Skip = "Roslyn assembly missmatch")]
+        [Fact]
         public virtual void Model_annotations_are_stored_in_snapshot()
         {
             Test(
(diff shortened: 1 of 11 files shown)


tinchou:
I just tried to run `SqlServerE2ETests` with TestDriven.NET and I got `PathTooLongException`. If I try to run `CSharpMigrationOperationGeneratorTest` everything works fine. If CI passes then I approve.

bricelam:
@tinchou I hit that too. I'll see if there's a workaround...

bricelam:
...if only we went with `EFCore` instead of `EntityFrameworkCore`... :trollface:

test/Microsoft.EntityFrameworkCore.Design.Tests/Microsoft.EntityFrameworkCore.Design.Tests.csproj
```diff
@@ -4,6 +4,11 @@
     <TargetFrameworks>net452;netcoreapp1.1</TargetFrameworks>
     <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp1.1</TargetFrameworks>
     <PreserveCompilationContext>true</PreserveCompilationContext>
+    <!-- TODO: Remove when Microsoft/vstest#428 is fixed. -->
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
+    <!-- TODO: Remove when dotnet/sdk#909 is fixed. -->
+    <OutputType>Exe</OutputType>
```
bricelam: Looks like this isn't needed.

bricelam:
Blocked on #7779

PR #7771, opened 2017-03-03
Query: Relational: only push down subqueries when needed.

### LINKED ISSUE #7766: Sometimes SelectExpressions are pushed down unnecessarily
There are many cases where, for one reason or another, `LiftSubQuery` will push a `SelectExpression` down into a subquery even though it is not required.

Some examples:

- It's being pushed down in the `MainFromClause` even though there are no `JoinClause`s, `GroupJoinClause`s, or `AdditionalFromClause`s
- It's being pushed down in a `JoinClause`, a `GroupJoinClause`, or an `AdditionalFromClause` even though the query will not be able to be flattened

tuespetre:
Fixes #7766

### DIFF
--- a/test/EFCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -497,19 +497,13 @@ WHERE (
             Assert.StartsWith(
                 @"@__p_0: 2
 
-SELECT [t].[OrderID], [t].[CustomerID], [t].[EmployeeID], [t].[OrderDate]
-FROM (
-    SELECT TOP(@__p_0) [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
-    FROM [Orders] AS [o]
-) AS [t]
-ORDER BY [t].[OrderID]
+SELECT TOP(@__p_0) [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
+FROM [Orders] AS [o]
+ORDER BY [o].[OrderID]
 
-SELECT [t1].[OrderID]
-FROM (
-    SELECT TOP(2) [o1].*
-    FROM [Order Details] AS [o1]
-) AS [t1]
-ORDER BY [t1].[ProductID], [t1].[OrderID]
+SELECT TOP(2) [o1].[OrderID]
+FROM [Order Details] AS [o1]
+ORDER BY [o1].[ProductID], [o1].[OrderID]
 
 @_outer_CustomerID2: VINET (Size = 450)
 
@@ -1745,11 +1739,8 @@ FROM [Customers] AS [c3]",
             Assert.Equal(
                 @"@__p_0: 91
 
-SELECT [t].[City]
-FROM (
-    SELECT TOP(@__p_0) [c].*
-    FROM [Customers] AS [c]
-) AS [t]",
+SELECT TOP(@__p_0) [c].[City]
+FROM [Customers] AS [c]",
                 Sql);
         }
 
@@ -1811,13 +1802,9 @@ ORDER BY [c].[CustomerID]",
             Assert.Equal(
                 @"@__p_0: 2
 
-SELECT [t].[City]
-FROM (
-    SELECT TOP(@__p_0) [c].*
-    FROM [Customers] AS [c]
-    ORDER BY [c].[CustomerID]
-) AS [t]
-ORDER BY [t].[CustomerID]",
+SELECT TOP(@__p_0) [c].[City]
+FROM [Customers] AS [c]
+ORDER BY [c].[CustomerID]",
                 Sql);
         }
 
@@ -2141,11 +2128,8 @@ FROM [Products] AS [p]",
             Assert.Equal(
                 @"@__p_0: 9
 
-SELECT [t].[EmployeeID]
-FROM (
-    SELECT TOP(@__p_0) [e].*
-    FROM [Employees] AS [e]
-) AS [t]",
+SELECT TOP(@__p_0) [e].[EmployeeID]
+FROM [Employees] AS [e]",
                 Sql);
         }
 
@@ -3877,10 +3861,13 @@ END",
 SELECT CASE
     WHEN NOT EXISTS (
         SELECT 1
-        FROM [Customers] AS [c]
-        WHERE LEN([c].[CustomerID]) <> 5
-        ORDER BY [c].[ContactName]
-        OFFSET @__p_0 ROWS FETCH NEXT @__p_1 ROWS ONLY)
+        FROM (
+            SELECT [c].*
+            FROM [Customers] AS [c]
+            ORDER BY [c].[ContactName]
+            OFFSET @__p_0 ROWS FETCH NEXT @__p_1 ROWS ONLY
+        ) AS [t]
+        WHERE [t].[CustomerID] LIKE N'A' + N'%' AND (CHARINDEX(N'A', [t].[CustomerID]) = 1))
     THEN CAST(1 AS BIT) ELSE CAST(0 AS BIT)
 END",
                 Sql);
@@ -4006,12 +3993,9 @@ ORDER BY [t].[Country]",
             base.Distinct_OrderBy2();
 
             Assert.Equal(
-                @"SELECT [t].[CustomerID], [t].[Address], [t].[City], [t].[CompanyName], [t].[ContactName], [t].[ContactTitle], [t].[Country], [t].[Fax], [t].[Phone], [t].[PostalCode], [t].[Region]
-FROM (
-    SELECT DISTINCT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
-    FROM [Customers] AS [c]
-) AS [t]
-ORDER BY [t].[CustomerID]",
+                @"SELECT DISTINCT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
+FROM [Customers] AS [c]
+ORDER BY [c].[CustomerID]",
                 Sql);
         }
 
@@ -5590,12 +5574,9 @@ WHERE [c].[CustomerID] LIKE N'A' + N'%' AND (CHARINDEX(N'A', [c].[CustomerID]) =
 
 @_outer_CustomerID: ALFKI (Size = 450)
 
-SELECT [t].[OrderDate]
-FROM (
-    SELECT TOP(3) [o].*
-    FROM [Orders] AS [o]
-    WHERE ([o].[OrderID] < 10500) AND (@_outer_CustomerID = [o].[CustomerID])
-) AS [t]",
+SELECT TOP(3) [o].[OrderDate]
+FROM [Orders] AS [o]
+WHERE ([o].[OrderID] < 10500) AND (@_outer_CustomerID = [o].[CustomerID])",
                 Sql);
         }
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -1089,6 +1165,13 @@ public virtual void AddToPredicate([NotNull] Expression predicate)
         {
             Check.NotNull(predicate, nameof(predicate));
 
+            if (_limit != null || _offset != null)
```
smitpatel: `AddToPredicate` is supposed to be a simple function. This logic should not be here at all.

tuespetre: I placed this logic here to be consistent with the several other 'simple' methods (and even setters) that use similar logic to ensure that a `SelectExpression` is pushed down into a subquery when appropriate: `set_Limit`, `set_Offset`, `SetProjectionExpression`, etc., but also because it is applicable in several places.

test/Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests/RowNumberPagingTest.cs
```diff
@@ -376,13 +376,16 @@ public override void Skip_Take_All()
 
 SELECT CASE
     WHEN NOT EXISTS (
-        SELECT [t].[c0]
+        SELECT 1
```
maumar: can we avoid this here? (previous query was better)

tuespetre: Previous query was incorrect, even though the tests passed. The produced SQL was corresponding to:

`customers.Where(c => c.CustomerID.Length != 5).Skip(...).Take(...)` 

When it should have corresponded to:

`customers.Skip(...).Take(...).Where(c => c.CustomerID.Length != 5)`

tuespetre: @maumar, the previous query generated for this test was incorrect but the test was not producing the proper conditions to show how it was incorrect.

After changing the test from `cs.OrderBy(c => c.ContactName).Skip(5).Take(10).All(p => p.CustomerID.Length == 5)` to `cs.OrderBy(c => c.ContactName).Skip(5).Take(10).All(p => !p.CustomerID.StartsWith("A"))`

Old query, returns 1:

```
SELECT CASE
    WHEN NOT EXISTS (
        SELECT [t].[c0]
        FROM (
            SELECT 1 AS [c0], ROW_NUMBER() OVER(ORDER BY [c].[ContactName]) AS [__RowNumber__]
            FROM [Customers] AS [c]
            WHERE [c].[CustomerID] LIKE N'A' + N'%' AND (CHARINDEX(N'A', [c].[CustomerID]) = 1)
        ) AS [t]
        WHERE ([t].[__RowNumber__] > 5) AND ([t].[__RowNumber__] <= (15)))
END
```

New query, returns 0:

```
SELECT CASE
    WHEN NOT EXISTS (
        SELECT 1
        FROM (
            SELECT [t0].*
            FROM (
                SELECT [c].*, ROW_NUMBER() OVER(ORDER BY [c].[ContactName]) AS [__RowNumber__]
                FROM [Customers] AS [c]
            ) AS [t0]
            WHERE ([t0].[__RowNumber__] > 5) AND ([t0].[__RowNumber__] <= (15))
        ) AS [t]        
        WHERE [t].[CustomerID] LIKE N'A' + N'%' AND (CHARINDEX(N'A', [t].[CustomerID]) = 1))
END
```

tuespetre:
:up::date:, rebased onto latest

smitpatel:
@tuespetre 
Previously we lifted subqueries only if a certain set of conditions were met (mainly depicted by `IsIdentityQuery`). While those conditions were much more restrictive in nature, this approach handles with problem from other direction i.e. lift select expression always and push down when needed. That brings in extra checks in `AddToOrderBy` & `AddToPredicate`, ideally which would have simple methods because Limit/Offset are result operators and applied after orderby/where translation. Though since this change lift select expression much freely, we need those checks which were already a subquery in query model.

I understand the original thing was not perfect and much more restrictive but is this approach better than that in some way? Can we improve lifting scenarios by applying appropriate conditions in `LiftSubquery` and avoid lifting rather than pushing down later?
The reason I ask this is because, when you do pushdown while adding orderby/where, you need to update the sql translation (LiftPredicate visitor) because when you were translating the expression you used incorrect SelectExpression (the one before push down). If we push down before actually running sqltranslatingexpressionvisitor then we don't need to deal with this overhead.

Further I am working refactoring SelectExpression to make it more consistent and referentially integral, which also removes `UpdateColumnExpression` and bunch of other methods. After my changes, the best way to bind a column with SelectExpression would become through member binder which is part of SqlTranslatingExpressionVisitor. And ad-hoc creation of ColumnExpression/AliasExpression are stopped which would introduce complexity in updating already translated SqlExpressions.

tuespetre:
> Previously we lifted subqueries only if a certain set of conditions were met (mainly depicted by IsIdentityQuery). ... Though since this change lift select expression much freely, we need those checks which were already a subquery in query model.

To be clear, this PR doesn't change which subqueries are lifted, just which `SelectExpression`s get pushed down into subqueries during lifting.

>I understand the original thing was not perfect and much more restrictive but is this approach better than that in some way?

Tighter SQL statements, especially when `GROUP BY` is thrown into the mix. This is not only for aesthetics but for performance -- much less bytes over the wire as can be seen from a number of the changed test results.

> Further I am working refactoring SelectExpression to make it more consistent and referentially integral, which also removes UpdateColumnExpression and bunch of other methods. After my changes, the best way to bind a column with SelectExpression would become through member binder which is part of SqlTranslatingExpressionVisitor. And ad-hoc creation of ColumnExpression/AliasExpression are stopped which would introduce complexity in updating already translated SqlExpressions.

I am really interested to see what you are doing with that, especially in relation to `GROUP BY`, because many of those cases turn out to be very complex. Example:

```
from c in context.Customers
join og in (from o in context.Orders group o by o.CustomerID)
    on c.CustomerID equals og.Key
select new { c.ContactName, og.Max(o => o.Total) }
```

By the time you get to translate the `Max` part of the projection to SQL, the `GROUP BY` query is pushed down and you have this:

```
SELECT [c].[ContactName]
FROM [Customers] AS [c]
JOIN (
    SELECT [o].[CustomerID]
    FROM [Orders] AS [o]
    GROUP BY [o].[CustomerID]
) AS [og]
```

...and you have to both successfully translate `MAX([o].[Total]) AS [o.Total.Max]` and `[og].[o.Total.Max]` out of that.

smitpatel:
> To be clear, this PR doesn't change which subqueries are lifted, just which SelectExpressions get pushed down into subqueries during lifting.

> Tighter SQL statements, especially when GROUP BY is thrown into the mix. This is not only for aesthetics but for performance -- much less bytes over the wire as can be seen from a number of the changed test results.

The pushdowns which were happening during LiftSubquery are happening at later stage now. If you don't pushdown a subquery then you are basically lifting it up. This is certainly improvement over generated SQL. My point is can we include some logic inside `LiftSubQuery` to do pushdown at appropriate places so that we don't have to pushdown later.

Taking simple example from tests added:
```c#
AssertQuery<OrderDetail>(ods => 
                ods.OrderBy(d => d.Quantity).ThenBy(d => d.OrderID).Take(2).OrderBy(d => d.ProductID).ThenBy(d => d.Quantity), 
                entryCount: 2); 
```
The generated query model is as follows:
```
  Optimized query model: 
  'from OrderDetail d in 
      (from OrderDetail d in DbSet<OrderDetail>
      order by d.Quantity asc, d.OrderID asc
      select d)
      .Take(__p_0)
  order by d.ProductID asc, d.Quantity asc
  select d'
```
Now in this case if we don't lift the subquery and have the SelectExpression inside it appear in a subquery select then while adding outer order by we don't need to do push down. (btw above query works correctly in current dev)

Having pushdown inside orderby/where means that we made a flat select expression for a subquery model instead of making a subquery select expression as it was required.
So there becomes 2 approaches to reach above improvement,
1. Either we don't pushdown till we need like in this PR.
2. We identify cases where we need to pushdown eagerly and of course much accurately. (this is the logic which is inaccurate atm).

So which approach would be better for us? After changes to SelectExpression 1 may be more complex than 2. But if 2 cannot solve everything for us then 1 is also fine.

While discussing in team, based on the Sql which are improving, there is another way to do this in QueryOptimizer.
Following is the query model for one such improved query
```
  'from Level2 l2 in 
      (from Level2 l2 in DbSet<Level2>
      where l2.OneToOne_Required_FK_Inverse.Name == "L1 03"
      select l2)
      .Take(__p_0)
  select l2.Name'
```
In this case since we have only select in outer query model, we can combine it with subquery and flatten the query. Optimizing at query model level makes things much easier while translating. Would you like to tackle this issue in query optimizer? 

As for group by query you posted, lets discuss on group by PR.

tuespetre:
@smitpatel good call on the pattern of the improved queries. I may not tackle this in the optimizer right now but since you have pointed that out I would say I am okay with backing off from this. If it's agreeable to you I will amend the commit to include the new test cases as well as the fix to the incorrect query that I discovered.

smitpatel:
>  fix to the incorrect query that I discovered.
Which incorrect queries?

test/EFCore.SqlServer.FunctionalTests/RowNumberPagingTest.cs
```diff
@@ -376,13 +376,16 @@ public override void Skip_Take_All()
 
 SELECT CASE
     WHEN NOT EXISTS (
-        SELECT [t].[c0]
+        SELECT 1
```
tuespetre: @smitpatel this one that @maumar commented on previously

test/EFCore.SqlServer.FunctionalTests/RowNumberPagingTest.cs
```diff
@@ -464,6 +467,49 @@ SELECT COUNT_BIG(*)
...
+FROM (
+    SELECT [d0].[OrderID], [d0].[ProductID], [d0].[Discount], [d0].[Quantity], [d0].[UnitPrice], ROW_NUMBER() OVER(ORDER BY [d0].[Quantity], [d0].[OrderID]) AS [__RowNumber__]
+    FROM [Order Details] AS [d0]
+) AS [t0]
+WHERE [t0].[__RowNumber__] > @__p_0
+ORDER BY [t0].[Quantity], [t0].[OrderID]",
+                Sql);
+        }
+
+        public override void Skip_Skip_Take()
```
tuespetre: @smitpatel IIRC this test would also fail without a fix

test/EFCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -3877,10 +3861,13 @@ public override void Skip_Take_All()
 SELECT CASE
     WHEN NOT EXISTS (
         SELECT 1
-        FROM [Customers] AS [c]
-        WHERE LEN([c].[CustomerID]) <> 5
-        ORDER BY [c].[ContactName]
-        OFFSET @__p_0 ROWS FETCH NEXT @__p_1 ROWS ONLY)
+        FROM (
```
tuespetre: @smitpatel this one

smitpatel: What is wrong with the original query here? The test itself was changed causing different SQL.

test/EFCore.SqlServer.FunctionalTests/RowNumberPagingTest.cs
smitpatel: Test changed. Original SQL looks fine.

test/EFCore.SqlServer.FunctionalTests/RowNumberPagingTest.cs
smitpatel: This could potentially fail. Though looking at generated SQL. It doesn't look very good given it has * at top level select. We should have proper column projected out. The reason we never generate SelectExpression with * at top level is you cannot accurately bind the valuebuffer to properties with star projection.

smitpatel:
We can certainly use the tests here. Also the improvement around `LiftSubquery` function is decent. Lets take this in once changes to Select Expression are merged in.

test/EFCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
tuespetre: I changed the All predicate test to demonstrate the issue. The predicate that was used before was hiding the issue. The subquery with Limit/Offset has to be pushed down before adding another predicate.

smitpatel: Good catch. It indeed is incorrect query. The modifying `AddToPredicate` does not seem to be right way to fix it.
This situation arises because of multiple result operators. And it should be fixed in same manner as done in #7578 . i.e. Result operators should be taking care of other result operators which may come before. Do you want to submit a fix for it in `HandleAll`?

tuespetre: Yes, do you have any objections to also fixing the `.Skip().Skip()` issue and including the new test cases in the same PR?

smitpatel: No objection as long as they are localized in handling result operators.

tuespetre: OK, great. Thanks!

tuespetre:
@smitpatel we'll see how the other PRs go; besides those, what would be left in this PR is not really substantial so I am going to close it.

Thanks! 🐢

Issue #7722, opened 2017-02-26
Query : error during compilation for queries with navigation properties and First/Single/client method operators inside a subquery


maumar:
Example using ComplexNavigations model:

```
ctx.LevelTwo.Select(l1 => ctx.LevelTwo.First().OneToOne_Required_FK_Inverse.Name);
```

```
ctx.LevelTwo.Where(l1 => ctx.LevelTwo.First().OneToOne_Required_FK_Inverse.Name == "L1 02");
```

maumar:
This happens  because our current binding logic doesn't know how to bind to the the inner subquery properly

maumar:
issue is caused by #8526

Issue #6647, opened 2016-09-30
Left Join (GroupJoin) always materializes elements resulting in unnecessary data pulling


qmicron:
### Steps to reproduce

```
public class Budget
{
    public Guid Id { get; set; }
    public Guid? DescriptionTextId { get; set; }
    public decimal PostCost { get; set; }
    public string PostNr { get; set; }
    public decimal PostPrice { get; set; }
    public decimal Quantity { get; set; }
    public decimal UnitCost { get; set; }
    public string UnitName { get; set; }
    public decimal UnitPrice { get; set; }
    public decimal VatAmount { get; set; }
    public decimal VatBasis { get; set; }
    public decimal VatFreeAmount { get; set; }
    public decimal VatFreePercent { get; set; }
    public decimal VatIncluded { get; set; }
    public decimal VatRate { get; set; }

    public Text DescriptionText { get; set; }
}

public class Text
{
    public Guid Id { get; set; }
    public string Plain { get; set; }
    public string Rtf { get; set; }

    public ICollection<Budget> BudgetDescriptionText { get; set; } = new List<Budget>();
}

public class DataContext : DbContext
{
    public DataContext(DbContextOptions options)
        : base(options)
    {
    }

    public DbSet<Budget> Budget => Set<Budget>();
    public DbSet<Text> Text => Set<Text>();

    protected override void OnModelCreating(ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<Budget>(entity =>
        {
            entity.HasKey(e => e.Id);
            entity.Property(e => e.Id).ValueGeneratedNever();
            entity.Property(e => e.DescriptionTextId);
            entity.Property(e => e.PostCost);
            entity.Property(e => e.PostNr).IsRequired().HasMaxLength(1000);
            entity.Property(e => e.PostPrice);
            entity.Property(e => e.Quantity);
            entity.Property(e => e.UnitCost);
            entity.Property(e => e.UnitName).HasMaxLength(100);
            entity.Property(e => e.UnitPrice);
            entity.Property(e => e.VatAmount);
            entity.Property(e => e.VatBasis);
            entity.Property(e => e.VatFreeAmount);
            entity.Property(e => e.VatFreePercent);
            entity.Property(e => e.VatIncluded);
            entity.Property(e => e.VatRate);
            entity.HasOne(e => e.DescriptionText)
                .WithMany(e => e.BudgetDescriptionText).HasForeignKey(e => e.DescriptionTextId);
        });

[...]

PR #7772, opened 2017-03-03
Fix to #6647 - Left Join (GroupJoin) always materializes elements resulting in unnecessary data pulling

maumar:
Problem was that for GroupJoin we would always force materialization on participating query sources.
We were doing this because we are not always able to correctly divide outer elements of the GroupJoin into correct groups.
However, if the GroupJoin clause is wrapped around SelectMany clause then groups don't matter because they are getting flattened by SelectMany anyway.

Fix is to recognize those scenarios and only force materialization when the correct grouping actually matters.
We can avoid materialization if the GroupJoin is followed by SelectMany clause (that references the grouping) and that the grouping itself is not present anywhere else in the query.
This addresses optional navigations, which is the 80% case. Manually created GroupJoins that are not modeling LeftOuterJoins still require additional materialization, but this can be addressed later as the priority is not nearly as high.

This change also addresses #7722 - Query : error during compilation for queries with navigation properties and First/Single/client method operators inside a subquery.

Problem here was that for some queries we don't know how to properly bind to a value buffer (when the result of binding is subquery, and not qsre).
Fix/mitigation is to recognize those scenarios and force materialization on the subqueries. This can be properly addressed in later commit (i.e. by improving the binding logic)

Also fixed several minor issues that were uncovered since we no longer do extensive materialization.

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -11,12 +12,14 @@ using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Query.Expressions;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
 using Microsoft.EntityFrameworkCore.Query.Internal;
 using Microsoft.EntityFrameworkCore.Query.ResultOperators.Internal;
 using Microsoft.EntityFrameworkCore.Query.Sql;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Utilities;
+using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 
@@ -71,7 +74,12 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
             var queryModelVisitor = (RelationalQueryModelVisitor)CreateQueryModelVisitor();
 
+            var queryModelMapping = new Dictionary<QueryModel, QueryModel>();
+            expression.QueryModel.PopulateQueryModelMapping(queryModelMapping);
+
             queryModelVisitor.VisitQueryModel(expression.QueryModel);
+
+            expression.QueryModel.RecreateQueryModelFromMapping(queryModelMapping);
 
             if (_querySource != null)
             {
--- a/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
@@ -251,7 +251,10 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
         {
             Check.NotNull(querySource, nameof(querySource));
 
-            return _tables.Any(te => te.QuerySource == querySource || te.HandlesQuerySource(querySource));
+            var processedQuerySource = PreProcessQuerySource(querySource);
+
+            return _tables.Any(te => te.QuerySource == processedQuerySource || te.HandlesQuerySource(processedQuerySource))
+                || base.HandlesQuerySource(querySource);
         }
 
         /// <summary>
--- a/src/EFCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
+++ b/src/EFCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
@@ -155,19 +155,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                 return handlerContext.EvalOnClient();
             }
 
-            if (relationalQueryModelVisitor.RequiresClientSingleColumnResultOperator
-                && !(resultOperator is SkipResultOperator
-                    || resultOperator is TakeResultOperator
-                    || resultOperator is FirstResultOperator
-                    || resultOperator is SingleResultOperator
-                    || resultOperator is CountResultOperator
-                    || resultOperator is AllResultOperator
-                    || resultOperator is AnyResultOperator
-                    || resultOperator is GroupResultOperator))
-            {
-                return handlerContext.EvalOnClient();
-            }
-
             return resultHandler(handlerContext);
         }
 
@@ -241,9 +228,8 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
                 if (!(expression.RemoveConvert() is SelectExpression))
                 {
-
-                    var inputType = expression.Type;
-                    var outputType = expression.Type;
+                    var inputType = handlerContext.QueryModel.SelectClause.Selector.Type;
+                    var outputType = inputType;
 
                     var nonNullableInputType = inputType.UnwrapNullableType();
                     if (nonNullableInputType == typeof(int)
@@ -253,7 +239,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                     }
 
                     expression = new ExplicitCastExpression(expression, outputType);
-                    var averageExpression = new SqlFunctionExpression("AVG", expression.Type, new [] { expression });
+                    var averageExpression = new SqlFunctionExpression("AVG", outputType, new [] { expression });
 
                     handlerContext.SelectExpression.SetProjectionExpression(averageExpression);
 
@@ -589,7 +575,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
                 if (!(expression.RemoveConvert() is SelectExpression))
                 {
-                    var minExpression = new SqlFunctionExpression("MIN", expression.Type, new [] { expression });
+                    var minExpression = new SqlFunctionExpression("MIN", handlerContext.QueryModel.SelectClause.Selector.Type, new [] { expression });
 
                     handlerContext.SelectExpression.SetProjectionExpression(minExpression);
 
@@ -611,7 +597,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
                 if (!(expression.RemoveConvert() is SelectExpression))
                 {
-                    var maxExpression = new SqlFunctionExpression("MAX", expression.Type, new [] { expression });
+                    var maxExpression = new SqlFunctionExpression("MAX", handlerContext.QueryModel.SelectClause.Selector.Type, new [] { expression });
 
                     handlerContext.SelectExpression.SetProjectionExpression(maxExpression);
 
@@ -666,7 +652,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
                 if (!(expression.RemoveConvert() is SelectExpression))
                 {
-                    var sumExpression = new SqlFunctionExpression("SUM", expression.Type, new [] { expression });
+                    var sumExpression = new SqlFunctionExpression("SUM", handlerContext.QueryModel.SelectClause.Selector.Type, new [] { expression });
 
                     handlerContext.SelectExpression.SetProjectionExpression(sumExpression);
 
--- a/src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -1204,26 +1204,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                 target,
                 Expression.Constant(property.Name));
 
-            if (!addNullCheck)
-            {
-                return propertyExpression;
-            }
-
-            var constantNull = property.ClrType.IsNullableType()
-                ? Expression.Constant(null, property.ClrType)
-                : Expression.Constant(null, property.ClrType.MakeNullable());
-
-            if (!property.ClrType.IsNullableType())
-            {
-                propertyExpression = Expression.Convert(propertyExpression, propertyExpression.Type.MakeNullable());
-            }
-
-            return Expression.Condition(
-                Expression.NotEqual(
-                    target,
-                    Expression.Constant(null, target.Type)),
-                propertyExpression,
-                constantNull);
+            return addNullCheck
+                ? new NullConditionalExpression(target, target, propertyExpression)
+                : propertyExpression;
         }
 
         private static bool IsCompositeKey([NotNull] Type type)
--- a/src/EFCore/Query/Internal/QuerySourceExtensions.cs
+++ b/src/EFCore/Query/Internal/QuerySourceExtensions.cs
@@ -5,6 +5,9 @@ using System;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Utilities;
 using Remotion.Linq.Clauses;
+using Remotion.Linq;
+using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Clauses.ResultOperators;
 
 namespace Microsoft.EntityFrameworkCore.Query.Internal
 {
@@ -25,5 +28,19 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
             return querySource.ItemName.StartsWith("<generated>_", StringComparison.Ordinal);
         }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static GroupJoinClause TryGetFlattenedGroupJoinClause([NotNull] this AdditionalFromClause additionalFromClause)
+            => (additionalFromClause.FromExpression is SubQueryExpression subQueryExpression
+                && subQueryExpression.QueryModel.IsIdentityQuery()
+                && subQueryExpression.QueryModel.ResultOperators.Count == 1
+                && subQueryExpression.QueryModel.ResultOperators[0] is DefaultIfEmptyResultOperator
+                && subQueryExpression.QueryModel.MainFromClause.FromExpression is QuerySourceReferenceExpression subqueryQsre
+                && subqueryQsre.ReferencedQuerySource is GroupJoinClause groupJoinClause)
+                ? groupJoinClause
+                : null;
     }
 }
--- a/src/EFCore/Query/QueryCompilationContext.cs
+++ b/src/EFCore/Query/QueryCompilationContext.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors;
@@ -19,6 +20,7 @@ using Microsoft.Extensions.Logging;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Clauses.ResultOperators;
 
 namespace Microsoft.EntityFrameworkCore.Query
 {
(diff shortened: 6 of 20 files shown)


maumar:
@anpete @smitpatel

maumar:
@tuespetre fyi, feel free to chime in as you have significant knowledge in this area

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
```diff
@@ -340,62 +343,253 @@ var requiresMaterializationExpressionVisitor
...
+
+                        foreach (var querySource in additionalSourcesRequiringMaterialization)
+                        {
+                            _state.QuerySourcesRequiringMaterialization.Add(querySource);
+                        }
+                    }
+                }
+            }
+
+            private Expression ExtractSourceExpression(Expression expression)
```
maumar: This code has been copied and modified from the EntityQueryModelVisitor.IterateCompositePropertyExpression Might be worthwhile to add a "debug"/"mock" binding. Wasn't sure if its worth to add public surface to that class - this approach is sort of a hack, ideally we would fix the binding itself to work for those scenarios. That's why I (initially) opted for copying the code into a private method that can be removed without issue once a proper fix is implemented. Thoughts?

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -318,6 +318,8 @@ var entityEqualityRewritingExpressionVisitor
             _navigationRewritingExpressionVisitorFactory
                 .Create(this).Rewrite(queryModel, parentQueryModel: null);
 
+            entityEqualityRewritingExpressionVisitor.Rewrite(queryModel);
```
smitpatel: Can we avoid rewrite twice?

maumar: I can try, but IIRC we were missing some cases if the equality rewrite happens only after nav rewrite

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
```diff
@@ -340,62 +343,253 @@ var requiresMaterializationExpressionVisitor
...
+
+                    memberExpression = expression as MemberExpression;
+                    methodCallExpression = expression as MethodCallExpression;
+                }
+
+                return expression;
+            }
+        }
+
+        private class RequiresMaterializationCompensatingVisitorState
```
maumar: I can split this into individual properties and pass them directly. In earlier iterations there was a boolean state that was shared between query model visitor and expression visitor so passing a reference was needed. Now it can be rewritten, but not sure if it's worth - this "nicely" encapsulates all the data/helpers that those visitors need. Thoughts?

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
smitpatel: why does nav rewrite depends on equality rewrite?

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/TableExpressionBase.cs
```diff
@@ -88,7 +91,33 @@ public virtual bool HandlesQuerySource([NotNull] IQuerySource querySource)
         {
             Check.NotNull(querySource, nameof(querySource));
 
-            return _querySource == querySource;
+            return _querySource == PreProcessQuerySource(querySource);
+        }
+
+        protected virtual IQuerySource PreProcessQuerySource(IQuerySource querySource)
+        {
+            var newQuerySource = ExtractGroupJoinFromSelectManySubquery(querySource);
```
tuespetre: Could the same result be achieved by assigning the appropriate `QuerySource` to `joinExpression` in `OptimizeJoinClause` (assign it the `joinClause`, and then if flattening a `DefaultIfEmpty`, assign it the `additionalFromClause` instead)?

tuespetre: @maumar these changes work but keep the 'hairiness' out of `TableExpressionBase`:

https://github.com/aspnet/EntityFramework/compare/fix7290...tuespetre:fix7290

It should be noted that the extra-hair-tastic bit (the `else if` block) is only required because we don't lift all the subqueries we could yet (i.e. lifting `_Select` using a `ProjectionShaper`) and thus the group joins aren't flattened as they could be.

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
tuespetre: No need to pass the state in -- the QueryModelVisitor and the ExpressionVisitor each can have a `public HashSet<IQuerySource> QuerySources { get; }` that they add to. In `FindQuerySourcesRequiringMaterialization`, the resulting sets from all three visitors (the original + two compensating) can be combined

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
tuespetre: Not much need for this, once inside the `BindMemberExpression`/`BindMethodCallExpression` lambda, there can only have been a single property, so `RemoveConvert()` on `node.Expression` or `node.Arguments[0]` suffices

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
```diff
@@ -340,62 +343,254 @@ var requiresMaterializationExpressionVisitor
...
+                return base.VisitMethodCall(node);
+            }
+
+            private void CompensateForUnboundablePropertyAccess(Expression expression)
+            {
+                var sourceExpression = ExtractSourceExpression(expression);
+                if (sourceExpression is SubQueryExpression subQueryExpression)
+                {
+                    var queryModel = subQueryExpression.QueryModel;
+                    if (queryModel.ResultOperators.ToList().OfType<ChoiceResultOperatorBase>().Any())
```
tuespetre: Shouldn't need to check the result operators -- if it's a subquery bound to a property, you can just promote the selector as a QSRE, using `HandleUnderlyingQuerySources` pointing at a `MarkForMaterialization` method that adds them to the collection

tuespetre:
@maumar I had to confirm all of my suspicions before making comments, here are the results in case it's of any help to you: https://github.com/aspnet/EntityFramework/compare/fix7290...tuespetre:fix7290

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
maumar: good call!

src/Microsoft.EntityFrameworkCore/Query/EntityQueryModelVisitor.cs
maumar: currently some scenarios involving navigation entity comparison. we could probably refactor that code by adding some intermediate step between nav rewrite and entity comparison, will file a bug on this

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
maumar: is selector guaranteed to be QSRE? Might be safer to just delegate to the requiresmaterialization visitor, but I agree that there is no need to check for result operators - (I think) there should always be one if we end up in this situation

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
maumar: Makes sense, simplifies this code a lot, now that both visitors don't share state

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/TableExpressionBase.cs
maumar: @smitpatel @anpete any thoughts on this - you guys know this code better than me ;) One thing worth pointing out is that we can't rely on ProjectionShaper optimization to simplify the code, since there can always be client method involved

maumar:
new version up

src/Microsoft.EntityFrameworkCore.Relational/Query/Expressions/TableExpressionBase.cs
tuespetre: Noting before I forget: The block itself was only needed to make two tests pass: GroupJoin_on_left_side_being_a_subquery and its twin.

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
tuespetre: If it's an entity property access, I would think so (or it's another subquery that returns an entity, which `HandleUnderlyingQuerySources` should handle) BUT now that you bring it up, what about a subquery that selects an entity via a constructor, like `(from c in customers select new Customer() { SingleProp = c.SingleProp }).FirstOrDefault(),SingleProp`? I'm not sure I've seen a test that covers that already.

src/Microsoft.EntityFrameworkCore/Query/QueryCompilationContext.cs
```diff
@@ -337,65 +340,133 @@ var requiresMaterializationExpressionVisitor
...
+                    CompensateForUnboundPropertyAccess(node.Arguments[0]);
+                });
+                
+                return base.VisitMethodCall(node);
+            }
+
+            private void CompensateForUnboundPropertyAccess(Expression expression)
+            {
+                var sourceExpression = expression.RemoveConvert();
+                if (sourceExpression is SubQueryExpression subQueryExpression)
```
tuespetre: ⛏ : `if (expression.RemoveConvert() is SubQueryExpression subQueryExpression)`

tuespetre: Hold up, new developments coming in

tuespetre:
@maumar, with a couple small changes, `RequiresClientSingleColumnResultOperator` can be removed. I also tried to see if the `UnboundPropertyAccessCompensatingVisitor` could be removed entirely. Here is the result of both: https://github.com/aspnet/EntityFramework/commit/84164647694f1699dfbc389773ce529899842766

maumar:
@tuespetre the change is dangerously growing in complexity ;) I thinking to keep the hack for now as its very simple and localized and will leave #7722 open. Once this change is in, you could submit a new PR with the proper fix and remove the hack.

tuespetre:
I thought so too, especially the `UnboundPropertyAccessCompensatingVisitor`. Good call to leave that for afterwards tied to #7722. 

The `RequiresClientSingleColumnResultOperator` change is not complex in itself, though. (2 changed files, other than test SQL output)

maumar:
@tuespetre agreed, especially that it's related to this change - we introduced this flag because it was messing with out groupjoins. Will incorporate this into the PR

maumar:
@tuespetre after doing some more digging, simply removing ```RequiresClientSingleColumnResultOperator ``` regresses queries like this:

```
                var query = context.Gears.GroupJoin(
                    context.Tags,
                    g => new { k1 = g.Nickname, k2 = (int?)g.SquadId },
                    t => new { k1 = t.GearNickName, k2 = t.GearSquadId },
                    (g, t) => g.HasSoulPatch).Distinct();
```

For those we still materialize extra stuff, so distinct needs to happen on the client still. 

```RequiresClientSingleColumnResultOperator ``` can be dealt with when fixing #7497. I would imagine we will do something similar to the groupjoin compensating visitor - if the groupjoin is followed by appropriate SelectMany and is only present in the query model once, we allow result operators to be translated, otherwise force client eval.

**Comment by tuespetre** [comment]
Oh, ok. I suppose `Distinct` on a `bool` would mask some errors like that!

**Comment by maumar** [comment]
new version up (adding more tests)

**Comment by maumar** [comment]
new version up, rebased on the latest changes and fixed some more bugs found along the way

File: src/EFCore/Query/ExpressionVisitors/Internal/RecursiveQueryModelVisitor.cs [hunk 7]
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class RecursiveQueryModelVisitor : ExpressionTransformingQueryModelVisitor<RecursiveQueryModelExpressionVisitor>
```
tuespetre: why hello there beautiful 💖

smitpatel: @maumar - Is it visitor to visit a visitor? :trollface:

maumar: This reminds me, we can reuse this in NondeterministicResultCheckingVisitor

tuespetre:
@maumar I have an idea that might eliminate the need for not only the `PreProcessQuerySource` stuff, but also the `SnapshotQuerySourceMapping`/`PopulateQueryModelMapping` etc. stuff.

src/EFCore/Query/ExpressionVisitors/Internal/RecursiveQueryModelVisitor.cs
```diff
@@ -0,0 +1,44 @@
...
+using Remotion.Linq;
+using Remotion.Linq.Clauses.Expressions;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class RecursiveQueryModelVisitor : QueryModelVisitorBase
```
tuespetre: Yo @maumar would you mind splitting this into its own PR so it can be merged sooner? I've got multiple places in multiple branches where I'd like to use this.

maumar: Sure

src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
```diff
@@ -129,6 +137,21 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
             return base.VisitMethodCall(node);
         }
 
+        protected override Expression VisitExtension(Expression node)
```
anpete: Base VisitExtension calls VisitChildren on the extension node. I know we have been inconsistent with this but I'd like to fix that 😄

src/EFCore/Query/ExpressionVisitors/Internal/RecursiveQueryModelVisitor.cs
smitpatel: @maumar - We talked about it yesterday only. 😄 I had implemented a visitor which would apply QueryModelVisitor recursively. Lets discuss about it.

@tuespetre - PR #7822 can utilize such recursive visitor & detect all subquery expressions but there is one test which is regressing on that due to how parsed querymodel is generated and order which optimizations are applied. If you want to tackle that issue in that way. I do not want you to block you on that so the current changes in that PR is also fine.

src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
maumar: we skip over ```NullConditionalExpression``` in ```ExpressionVisitorBase```. However we do process NullConditional in ```DefaultQueryExpressionVisitor``` so it can be safely removed here

anpete: > we skip over NullConditionalExpression in ExpressionVisitorBase

Yeah, this needs to change, too. We are breaking the pattern by doing that.

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -1204,26 +1204,9 @@ private static Expression CreatePropertyExpression(Expression target, IProperty
...
-            }
-
-            return Expression.Condition(
-                Expression.NotEqual(
-                    target,
-                    Expression.Constant(null, target.Type)),
-                propertyExpression,
-                constantNull);
+            return addNullCheck
+                ? new NullConditionalExpression(target, target, propertyExpression)
```
smitpatel: For this reason, I wonder what is difference between `caller` and `nullableCaller` in NullConditionalExpression.

src/EFCore/Query/ExpressionVisitors/Internal/RecursiveQueryModelVisitor.cs
smitpatel: #7843  is merged so you can simply this. Can you also post reference where this is being used? (I suppose there is derived class for this else this 2 visitor does nothing other than calling each other. So much Love)

src/EFCore/Query/QueryCompilationContext.cs
```diff
@@ -337,64 +339,194 @@ var requiresMaterializationExpressionVisitor
...
+                    return method.Arguments[0] as QuerySourceReferenceExpression;
+                }
+
+                return null;
+            }
+
+            private void MaterializeOptionalNavigationSource(QuerySourceReferenceExpression sourceQsre)
+            {
+                if (sourceQsre != null
+                    && sourceQsre.ReferencedQuerySource is AdditionalFromClause additionalFromClause
```
smitpatel: DRY this conditions with the GroupJoin visitor. You can put it in QuerySourceExtensions if it is used outside of this file. Or private function in this file. This thing basically takes AdditionalFromClause and find the pattern.

test/EFCore.Sqlite.FunctionalTests/ComplexNavigationsQuerySqliteTest.cs
```diff
@@ -11,5 +11,15 @@ public ComplexNavigationsQuerySqliteTest(ComplexNavigationsQuerySqliteFixture fi
             : base(fixture)
         {
         }
+
+        public override void Optional_navigation_inside_nested_method_call_translated_to_join_keeps_original_nullability_also_for_arguments()
+        {
+            base.Optional_navigation_inside_nested_method_call_translated_to_join_keeps_original_nullability_also_for_arguments();
+        }
```
smitpatel: remove these.

src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
```diff
@@ -71,8 +74,13 @@ protected override Expression VisitSubQuery(SubQueryExpression expression)
 
             var queryModelVisitor = (RelationalQueryModelVisitor)CreateQueryModelVisitor();
 
+            var queryModelMapping = new Dictionary<QueryModel, QueryModel>();
+            expression.QueryModel.PopulateQueryModelMapping(queryModelMapping);
```
smitpatel: Why did we not require this before and what changes brought this requirement now?

src/EFCore/Query/QueryCompilationContext.cs
smitpatel: `PreProcessQuerySource` uses this. Make extension method on AdditionalFromClause. Can also used in GroupJoin visitor if you override VisitAdditionalFromClause method.
Name this pattern something good in extension method like `IsMaumerJoin` :trollface:

src/EFCore/Query/QueryCompilationContext.cs
```diff
@@ -337,64 +339,194 @@ var requiresMaterializationExpressionVisitor
...
-            var groupJoinMaterializationExpressionVisitor = new RequiresMaterializationForGroupJoinExpressionVisitor();
-            var groupJoinMaterializationQueryModelVistor = new RequiresMaterializationForGroupJoinQueryModelVisitor(
-                groupJoinMaterializationExpressionVisitor,
-                _querySourcesRequiringMaterialization,
+            var groupJoinCompensatingVisitor = new GroupJoinMaterializationCompensatingVisitor(
                 requiresMaterializationExpressionVisitor);
 
-            groupJoinMaterializationExpressionVisitor.QueryModelVisitor = groupJoinMaterializationQueryModelVistor;
-            groupJoinMaterializationQueryModelVistor.VisitQueryModel(queryModel);
+            var unboundPropertyAccessCompensatingVisitor = new UnboundPropertyAccessCompensatingVisitor(
```
smitpatel: This 🔥

src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
maumar: It was a bug, we just didn't catch it because for the relevant tests we have, LOJ and IJ were returning the same results.

src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
maumar: I will do this in a separate PR

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
maumar: its relevant for the nested NCEs - nullable caller can be another NCE, whereas caller and access operation will always be normal expressions (methods/member access)

src/EFCore/Query/QueryCompilationContext.cs
maumar: Wrt ```GroupJoinMaterializationCompensatingVisitor``` I need to process GroupJoin clauses, not AdditionalFrom clauses because there can be a groupjoin without select many, and we would not catch those cases (i.e. wouldn't mark them for materialization)

maumar:
new version up

src/EFCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
```diff
@@ -666,7 +653,7 @@ private static Expression HandleSum(HandlerContext handlerContext)
 
                 if (!(expression.RemoveConvert() is SelectExpression))
                 {
-                    var sumExpression = new SqlFunctionExpression("SUM", expression.Type, new [] { expression });
+                    var sumExpression = new SqlFunctionExpression("SUM", handlerContext.QueryModel.SelectClause.Selector.Type, new [] { expression });
```
smitpatel: test failure?

maumar: yes, problematic scenario is ```entities.Select(e => (int?)e.Optional.Id)).Sum()``` which we now fully translate to sql. When converting single result element to a collection, the expected element TResult is int? but looking at selectExpression, we think its int (because we nuke all converts when translating). Fix is to look at query model instead which has appropriate typing.

tuespetre: Don't forget Min/Max! :cactus:

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -251,7 +251,9 @@ public override bool HandlesQuerySource([NotNull] IQuerySource querySource)
         {
             Check.NotNull(querySource, nameof(querySource));
 
-            return _tables.Any(te => te.QuerySource == querySource || te.HandlesQuerySource(querySource));
+            var processedQuerySource = PreProcessQuerySource(querySource);
+
+            return _tables.Any(te => te.QuerySource == processedQuerySource || te.HandlesQuerySource(processedQuerySource));
```
tuespetre: Can be simplified (only needs to check if HandlesQuerySource)

tuespetre: Also wouldn't hurt to fallback via OrElse to base.HandlesQuerySource

src/EFCore/Query/Internal/QuerySourceExtensions.cs
```diff
@@ -25,5 +28,19 @@ public static bool HasGeneratedItemName([NotNull] this IQuerySource querySource)
...
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static GroupJoinClause TryGetFlattenedGroupJoinClause([NotNull] this AdditionalFromClause additionalFromClause)
+            => (additionalFromClause.FromExpression is SubQueryExpression subQueryExpression
+                && subQueryExpression.QueryModel.IsIdentityQuery()
+                && subQueryExpression.QueryModel.ResultOperators.Count == 1
+                && subQueryExpression.QueryModel.ResultOperators[0] is DefaultIfEmptyResultOperator
+                && subQueryExpression.QueryModel.MainFromClause.FromExpression is QuerySourceReferenceExpression subqueryQsre
+                && subqueryQsre.ReferencedQuerySource is GroupJoinClause groupJoinClause)
```
tuespetre: Should probably explicitly check if the GroupJoinClause is the immediately preceding body clause in the query model

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
maumar: I tried that initially, but some queries deteriorate - CROSS JOIN gets converted to CROSS APPLY, where no correlation is present. Will leave as is.

maumar:
@anpete new version up

src/EFCore/Query/Internal/QuerySourceExtensions.cs
maumar: this actually requires query model, which we don't have in some contexts. Currently we are doing this check in the caller code, will keep as is.

src/EFCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
```diff
@@ -241,9 +228,8 @@ private static Expression HandleAverage(HandlerContext handlerContext)
 
                 if (!(expression.RemoveConvert() is SelectExpression))
                 {
-
-                    var inputType = expression.Type;
-                    var outputType = expression.Type;
+                    var inputType = handlerContext.QueryModel.SelectClause.Selector.Type;
+                    var outputType = handlerContext.QueryModel.SelectClause.Selector.Type;
```
anpete: outputType = inputType?

src/EFCore/Query/QueryCompilationContext.cs
```diff
@@ -337,64 +339,125 @@ var requiresMaterializationExpressionVisitor
...
 
-            _querySourcesRequiringMaterialization = requiresMaterializationExpressionVisitor
+            var querySourcesRequiringMaterialization = requiresMaterializationExpressionVisitor
                     .FindQuerySourcesRequiringMaterialization(queryModel);
 
-            var groupJoinMaterializationExpressionVisitor = new RequiresMaterializationForGroupJoinExpressionVisitor();
-            var groupJoinMaterializationQueryModelVistor = new RequiresMaterializationForGroupJoinQueryModelVisitor(
-                groupJoinMaterializationExpressionVisitor,
-                _querySourcesRequiringMaterialization,
+            var groupJoinCompensatingVisitor = new GroupJoinMaterializationCompensatingVisitor(
```
anpete: Anyway to unify all of this inside RequiresMaterializationExpressionVisitor?

maumar: I implemented them intentionally outside - RequiresMaterializationExpressionVisitor should just find query sources that are reachable from the final projection and/or ones that need to be materialized due to client evals. Those visitors here are just hacks/workarounds for existing issues and once those issues are resolved, they can be removed - without tinkering with the RMEV, which is itself quite complicated.

smitpatel: Tracking issues to remove hacks?

maumar: #7787 is already tracking OptionalCollectionNavigationCompensatingVisitor, will file a new issue to track GroupJoinMaterializationCompensatingVisitor once this is checked in (and 7290 gets closed) - to track the remaining work on groupjoin materialization

PR #7789, opened 2017-03-04
Remove special handling for concat operator

roji:
A special string concat operator was added at some point to support PostgreSQL (which uses || to concat strings). This is an overridable property that is by default +.

After the introduction of a more generic GenerateOperator(), this no longer makes sense - a provider that needs a special concat operator (or any other) can simply override GenerateOperator() and do that there without any special support from Relational.

Removing to simplify/cleanup DefaultQuerySqlGenerator.

### DIFF
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -153,11 +153,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
         protected virtual IRelationalCommandBuilder Sql => _relationalCommandBuilder;
 
         /// <summary>
-        ///     The default string concatenation operator SQL.
-        /// </summary>
-        protected virtual string ConcatOperator => "+";
-
-        /// <summary>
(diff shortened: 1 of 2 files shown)


divega:
@smitpatel @roji isn't the failure in Travis-ci legit?

/home/travis/build/aspnet/EntityFramework/src/EFCore.Relational.Design/bin/Debug/net451/Microsoft.EntityFrameworkCore.Relational.Design.dll
Query/Sql/Internal/SqliteQuerySqlGenerator.cs(18,35): error CS0115: 'SqliteQuerySqlGenerator.ConcatOperator': no suitable method found to override [/home/travis/build/aspnet/EntityFramework/src/EFCore.Sqlite.Core/EFCore.Sqlite.Core.csproj]

smitpatel:
@roji - SQLite also overrides concat operator, can you make changes in SqliteQuerySqlGenerator? (should be similar to changes in Npgsql provider.

roji:
Apologies, did this a bit too quickly - rebased and fixed.

smitpatel:
Merged via 52ff4b79b2bb51d20b583bffa492d298b3df88e3
Thanks for contribution @roji

Issue #7784, opened 2017-03-03
Null-Coalesce Operator Generating Incorrect SQL Sometimes


stehlikio:
It looks like the SQL generated from a null-coalesce operator changes depending on a previous statement in the Linq expression. So far I've only noticed this change in behavior when a Where statement contains a !StartsWith somewhere in the Linq expression (doesn't matter if it's before or after).

```c#
context.Employees
    .Where(employee => !employee.FirstName.StartsWith("Test"))
    .Where(employee => (employee.DirectorId ?? employee.Id) == 1)
    .ToList();
```

The problem is that after doing the null-coalesce, it tacks on ```[employee].[DirectorId] IS NOT NULL```, which is unexpected.

#### Actual SQL
```sql
SELECT [employee].[Id], [employee].[DirectorId], [employee].[FirstName]
FROM [Employee] AS [employee]
WHERE (NOT ([employee].[FirstName] LIKE N'Test' + N'%') OR (CHARINDEX(N'Test', [employee].[FirstName]) <> 1)) AND ((COALESCE([employee].[DirectorId], [employee].[Id]) = 1) AND [employee].[DirectorId] IS NOT NULL)
```

#### Expected SQL

```sql
SELECT [employee].[Id], [employee].[DirectorId], [employee].[FirstName]
FROM [Employee] AS [employee]
WHERE (NOT ([employee].[FirstName] LIKE N'Test' + N'%') OR (CHARINDEX(N'Test', [employee].[FirstName]) <> 1)) AND (COALESCE([employee].[DirectorId], [employee].[Id]) = 1)
```

### Steps to reproduce

I've provided the DDL and the following code that demonstrates a few different examples of StartsWith, !StartsWith, and different coalesce statements.

```sql
create table Employee (
    Id int identity(1, 1),
    DirectorId int,
    FirstName varchar(255)
);
```
```c#
internal class Program
{
    internal class Context : DbContext
    {
        public DbSet<Employee> Employees { get; set; }

        public Context() { }
        public Context(DbContextOptions options) : base(options) { }
    }

    [Table("Employee")]
    internal class Employee
    {
        [Column("Id")]
        public int Id { get; set; }

        [Column("DirectorId")]
        public int? DirectorId { get; set; }

        [Column("FirstName")]
        public string FirstName { get; set; }
    }

    public static void Main(string[] args)
    {
        var options = new DbContextOptionsBuilder()
            .UseSqlServer(/* ... */)
[...]

PR #7804, opened 2017-03-07
fix issue 7784.  Remove extraneous IS NOT NULL checks on equals compa…

pmiddleton:
This fixes issue #7784.

RelationalNullsExpandingVisitor was adding an extra IS NOT NULL check when an equals comparison was performed on a nullable column.  That check is only needed if a not equals comparison is being performed, or if NOT logic is involved on any of the parameters.


maumar:
Merged in 42e8bbed604b0447a57933e75caa0e58ed35d912, thank you for the contribution!

PR #7807, opened 2017-03-08
Update Resources.tt and its usage

### LINKED ISSUE #7778: Fix Resources.tt
During the move to VS 2017, we lost our resource generator. With a few tweaks, we should be able to get it working again...

bricelam:
I went for a very T4-idiomatic approach here and removed any features that we weren't using.

Fixes #7778

P.S. I'll send a separate PR to modernize the generated code.


src/EFCore/Properties/CoreStrings.Designer.cs
```diff
@@ -353,7 +353,7 @@ public static string MultipleProvidersConfigured([CanBeNull] object storeNames)
         }
 
         /// <summary>
-        /// AddDbContext was called with configuration, but the context type '{contextType}' only declares a parameterless constructor. This means that the configuration passed to AddDbContext will never be used. If configuration is passed to AddDbContext, then ‘{contextType}’ should declare a constructor that accepts a DbContextOptions&lt;{contextType}&gt; and must pass it to the base constructor for DbContext.
+        /// AddDbContext was called with configuration, but the context type '{contextType}' only declares a parameterless constructor. This means that the configuration passed to AddDbContext will never be used. If configuration is passed to AddDbContext, then �{contextType}� should declare a constructor that accepts a DbContextOptions&lt;{contextType}&gt; and must pass it to the base constructor for DbContext.
```
bricelam: I'll fix those smart quotes in the resx...

tools/Resources.tt
```diff
@@ -124,86 +76,90 @@ namespace {0}
...
         {
-            return UsingNamedArgs ? name : 'p' + name;
+            result.Resources = Enumerable.ToList(
+                from DictionaryEntry r in reader
+                let value = (string)r.Value
+                select new Resource
+                {
+                    Name = (string)r.Key,
+                    Value = value,
+                    Parameters = Regex.Matches(value, @"\{(\w+)\}").Cast<Match>().Select(m => m.Groups[1].Value)
```
AndriySvyryd: Consider allowing digits in the parameter names

bricelam: YAGNI. We'll update to fit our needs as they evolve.

PR #7809, opened 2017-03-08
Reverse Engineer: Pluralize collection navigations

rowanmiller:
Tested manually as there isn't really test coverage for any of this

As pointed out by @jwooley in https://github.com/aspnet/EntityFramework/pull/7634, I missed pluralizing collection navigation in the initial PR.

### DIFF
--- a/src/EFCore.Relational.Design/RelationalScaffoldingModelFactory.cs
+++ b/src/EFCore.Relational.Design/RelationalScaffoldingModelFactory.cs
@@ -636,6 +636,12 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
                     dependentEndNavigationPropertyName)
                 : CandidateNamingService.GetPrincipalEndCandidateNavigationPropertyName(
                     foreignKey, dependentEndNavigationPropertyName);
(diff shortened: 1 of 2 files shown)


src/EFCore.Relational.Design/RelationalScaffoldingModelFactory.cs
```diff
@@ -636,6 +636,12 @@ protected virtual void AddNavigationProperties([NotNull] IMutableForeignKey fore
                     dependentEndNavigationPropertyName)
                 : CandidateNamingService.GetPrincipalEndCandidateNavigationPropertyName(
                     foreignKey, dependentEndNavigationPropertyName);
+
+            if (!foreignKey.IsUnique && !foreignKey.IsSelfReferencing())
```
rowanmiller: I looked at doing something smarter with pluralization of self referencing navs... but it seems like whatever we generate is going to need updating anyway, and what we generate is pretty clear. So I decided to leave them as-is and skip any pluralization.

rowanmiller:
@bricelam added a test

PR #7822, opened 2017-03-09
Query: Lift more subqueries in the QueryOptimizer

tuespetre:
Query models following a simple pattern of selecting from a subquery
only to create a projection can often be flattened by lifting the
subquery. These changes do that, which results in more efficient
queries for such cases, and in the case of Relational, more compact
SQL statements.

### DIFF
--- a/src/EFCore/Query/Internal/QueryOptimizer.cs
+++ b/src/EFCore/Query/Internal/QueryOptimizer.cs
@@ -13,7 +13,6 @@ using Microsoft.EntityFrameworkCore.Utilities;
 using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
-using Remotion.Linq.Clauses.ExpressionVisitors;
 using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Transformations;
 
@@ -102,9 +101,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
         private void TryFlattenJoin(JoinClause joinClause, QueryModel queryModel)
         {
-            var subQueryExpression = joinClause.InnerSequence as SubQueryExpression;
-
-            if (subQueryExpression != null)
+            if (joinClause.InnerSequence is SubQueryExpression subQueryExpression)
             {
                 VisitQueryModel(subQueryExpression.QueryModel);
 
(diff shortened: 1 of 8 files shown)


src/EFCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -76,6 +76,8 @@ public QueryOptimizer([NotNull] IModel model)
             _queryAnnotations = queryAnnotations;
 
             VisitQueryModel(queryModel);
+
+            new MainFromClauseFlatteningQueryModelVisitor(queryAnnotations).VisitQueryModel(queryModel);
```
smitpatel: How about overriding `VisitMainFromClause` here?

tuespetre: Doesn't work, the `SubQueryFlattener` base class will not hit all of the `SubQueryExpressions`

smitpatel: You can always override `VisitMainFromClause` apply your logic and if it doesn't work then call base. Question becomes if we want to do this merging in MainFromClause only or on AdditionalFromClause too (later may be better but probably needs some more tests to be added).
We are already overriding `FlattenSubquery` there you should be getting all subquery expressions.

tuespetre: @smitpatel that is what I tried first. It doesn't get all of the subquery expressions, and the `SubQueryFromClauseFlattener` base class won't even visit the `MainFromClause` if there are result operators on the query model.

smitpatel: @tuespetre - You are right. It does not get all subquery expressions. The dual dilemma of ExpressionVisitor & QueryModelVisitor (which forces @maumar to make recursive visitors)

smitpatel: Since we are using Expression/QueryModelVisitor pattern, can you move visiter creation & entry calls to `EntityQueryModelVisitor.OptimizeQueryModel` method. `QueryOptimizer` is a visitor on query model. So it is a stand-alone thing. If we need any other visitor on QMV then we add visitor in optimize query model just like `SubQueryMemberPushDown` or `EntityEqualityRewriting`.

Rest of changes look good to accommodate it. Its really unfortunate that QueryOptimizer is suboptimal at present to deal with such cases.

tuespetre:
:up::date:

src/EFCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -135,29 +132,21 @@ public override void VisitGroupJoinClause(GroupJoinClause groupJoinClause, Query
...
-            var additionalFromClause
-                = queryModel.BodyClauses.ElementAtOrDefault(index + 1)
-                    as AdditionalFromClause;
-
-            var querySourceReferenceExpression
-                = additionalFromClause?.FromExpression as QuerySourceReferenceExpression;
-
-            if (querySourceReferenceExpression != null
-                && querySourceReferenceExpression.ReferencedQuerySource == groupJoinClause)
+            if (queryModel.BodyClauses.ElementAtOrDefault(index + 1) is AdditionalFromClause additionalFromClause
```
smitpatel: Just a side note: can we avoid having such cleanups in the PRs which fixes certain bugs/issues? It makes harder to review when finding what code path actually changed. We generally run clean up on whole code together so that we don't have to review it in detail.

tuespetre: Want me to cut this part?

tuespetre:
@smitpatel which test is regressing that you had mentioned on #6647?

smitpatel:
This PR is superseded by #7843 
@tuespetre - QueryOptimizer is good place to capture optimizations on QueryModel which are based on structure of QM. We wanted to include this change in QueryOptimizer but it was regressing a test as pointed out in #7844 . After investigating it, considering the cause of the issue may not be us, we decided that we are fine with that test change.

tuespetre:
OK. There are still cases that the `SubQueryFromClauseFlattener` will never hit, though (`Contains_with_subquery_optional_navigation_and_constant_item`).

smitpatel:
Thanks for pointing out. I will investigate into it.

PR #7823, opened 2017-03-09
Query: Relational: Fix issues with problematic Take/Skip queries

tuespetre:
Certain queries like All needed to push down the select expression
into a subquery if it had a limit or offset. Other queries like
Skip().Skip() needed to push down the select expression into a subquery
when adding an Offset to a select expression that already had an offset.

This PR includes a `SelectExpression.LiftTableReferences` function as well that is used to avoid double translation of predicates in certain scenarios where the predicate has been translated already but the select expression needs to be pushed down before the predicate is used.

### DIFF
--- a/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
@@ -305,9 +305,9 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             [param: CanBeNull]
             set
             {
-                if (value != null)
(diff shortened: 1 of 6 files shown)


smitpatel:
With #7950 , we have removed function `SelectExpression.UpdateColumnReferences` which is extensively used by `SelectExpression.LiftTableReferences` visitor in this PR. The core part of this change is no longer in SelectExpression and it would probably require reworking the fix here.

@tuespetre - I am closing this PR. If you want to tackle the problem with different fix inline with current dev, feel free to submit a new PR. Thanks.

Issue #2461, opened 2015-06-24
Migrations: Scaffold comments


bricelam:
Migrations should be able to scaffold comments for operations like.

``` C#
// This operation may result in data loss.
migrationBuilder.DropTable("Customers");
```

Providers should also be able to participate, for example, SQLite may want to write something like.

``` C#
// On SQLite, this operation results in a table rebuild.
migrationBuilder.DropColumn(name: "Birthday", table: "Customers");
```

bricelam:
Here are some thoughts I had on implementing this:

The model differ could pass the operations to the provider giving them a chance to add annotations like `CodeGen:Comment = "This will rebuild the table."`. The scaffolder would then use these special CodeGen annotations when serializing the operations.

natemcmaster:
Although a comment is helpful, I think casual users may not read the comments. When data loss is possible, could we do more assertive , such as requiring a command-line flag "--AllowDrops" or prompting y/n on migration execution?

bricelam:
We already log a yellow warning when operations that could result data loss are scaffolded. But there currently isn't a way for providers to influence whether an operation could result in data loss.

divega:
Here is another scenario we may want to generate comments for: *Add a migrations warning for ValueGeneratedOnAdd for a non-key property for SqlServer #7295*

ajcvickers:
We have already improved logging/feedback in this area, and we hope to improve it further once simple logging is implemented in EF Core. Closing this issue as there is nothing actionable here that isn't tracked elsewhere.

PR #7824, opened 2017-03-09
WIP EntityTypeBuilder.SeedData

tinchou:
See #629.


src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -1183,6 +1207,75 @@ protected virtual IEnumerable<MigrationOperation> Remove([NotNull] ISequence sou
...
+        #region SeedData
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected virtual IEnumerable<MigrationOperation> DiffSeedData(
+            [NotNull] IEntityType source,
+            [NotNull] IEntityType target,
+            [NotNull] DiffContext diffContext)
```
bricelam: Note, when the `Down` operations are being generated, `source` and `target` will be backwards from what you expect. `source` will be for the current model.

tinchou:
I'm seeing errors in `ModelSnapshotSqlServerTest` for a couple of tests I added: https://ci.appveyor.com/project/aspnetci/entityframework/build/1.2.0-preview1-4300/tests

```
System.InvalidOperationException : Build failed. Diagnostics: warning CS1701: Assuming assembly reference 'System.Runtime, Version=4.0.20.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' used by 'Microsoft.EntityFrameworkCore' matches identity 'System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' of 'System.Runtime', you may need to supply runtime policy
warning CS1701: Assuming assembly reference 'System.Runtime, Version=4.0.20.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' used by 'Microsoft.EntityFrameworkCore' matches identity 'System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' of 'System.Runtime', you may need to supply runtime policy
```

If I run the following test in my computer everything works fine:

```cs
public class ModelSnapshotSqlServerTest : ModelSnapshotTest
{
    public override void SeedData_annotations_are_stored_in_snapshot()
    {
        base.SeedData_annotations_are_stored_in_snapshot();
    }
}
```

Any ideas?

tinchou:
Other updates: to be able to reuse the update pipeline I had to make the `Insert`, `Update` and `Delete` operations single-row. This is required because `IUpdateSqlGenerator` takes a single `ModificationCommand` object that has an `IEnumerable<ColumnModification>` for that row: https://github.com/aspnet/EntityFramework/pull/7824/files#diff-d7c7182b48a3b33bb7a237506b4c4037R31

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
tinchou: Thanks for the pointer, it's mostly taken care of now :)

tinchou:
There's clearly something wrong in the `ModelSnapshotTest`s that I added, but I have no idea how they're different than the others and why they would reference a different `System.Runtime`.

It's worth noting this only happens in .NET Core, not Full Framework (as run by TestDriven.NET).

@bricelam any ideas?

test/EFCore.Relational.Tests/Migrations/MigrationSqlGeneratorTest.cs
```diff
@@ -310,89 +310,61 @@ public void Generate_doesnt_batch_by_default()
...
         {
             base.DeleteRowsOperation_composite_key();
 
+            // TODO remove rowcount
             Assert.Equal(
                 "DELETE FROM \"People\"" + EOL +
-                "WHERE (\"First Name\" = 'Hodor' AND \"Last Name\" IS NULL) OR" + EOL +
-                "      (\"First Name\" = 'Daenerys' AND \"Last Name\" = 'Targaryen');" + EOL,
+                "WHERE \"Id1\" = 2 AND \"Id2\" = 'John Snow';" + EOL +
+                "SELECT provider_specific_rowcount();" + EOL + EOL,
```
tinchou: Going through the `UpdateSqlGenerator` adds the row count SELECT. We should discuss whether this could be useful for error checking or if we could add an option to avoid generating it.

test/EFCore.Relational.Tests/TestUtilities/FakeProvider/FakeRelationalOptionsExtension.cs
```diff
@@ -48,13 +50,81 @@ public static IServiceCollection AddEntityFrameworkRelationalDatabase(IServiceCo
...
                 .TryAdd<IRelationalDatabaseCreator, FakeRelationalDatabaseCreator>();
 
             builder.TryAddCoreServices();
 
             return serviceCollection;
         }
+
+        //
+        //
+        // TODO avoid copy-pasting from CommandBatchPreparerTest
```
tinchou: Please advise on how to reuse `TestModificationCommandBatchFactory` from `CommandBatchPreparerTest`

test/EFCore.SqlServer.FunctionalTests/PropertyEntrySqlServerTest.cs
```diff
@@ -27,7 +27,7 @@ public override void Property_entry_original_value_is_set()
             Assert.Contains(
                 @"SET NOCOUNT ON;
 UPDATE [Engines] SET [Name] = @p0
-WHERE [Id] = @p1 AND [EngineSupplierId] = @p2 AND [Name] = @p3;
+WHERE [Id] = @p1 AND [EngineSupplierId] = @p3 AND [Name] = @p4;
```
tinchou: I have no idea how my code impacted this.

tinchou: Just checked with @smitpatel: UpdateSqlGenerator.cs:391 has a side effect on the ColumnModification class of generating a parameter name when calling `ColumnModification.ParameterName`. It should be fine since we're reusing code for another feature, but we could consider changing the condition to `columnModificationBase.GetType() != typeof(ColumnModification) && columnModificationBase.ParameterName == null` for 100% backwards behavior compatibility.

src/EFCore.Relational.Design.Specification.Tests/Migrations/ModelSnapshotTest.cs
```diff
@@ -1819,6 +1820,101 @@ public virtual void Do_not_generate_entity_type_builder_again_if_no_foreign_key_
 
         #endregion
 
+        #region SeedData
+
+        [FrameworkSkipCondition(RuntimeFrameworks.CoreCLR, SkipReason = "System.Runtime binding error")]
```
tinchou: I haven't been able to figure out why this is happening yet. I'll work on this during the week, but worst-case scenario I'd like to merge and open a new issue to track it.

src/EFCore.Relational/Update/ColumnModification.cs
```diff
@@ -24,55 +24,48 @@ public class ColumnModification
...
-        public virtual bool UseOriginalValueParameter => IsCondition && IsConcurrencyToken;
-
         public virtual bool UseCurrentValueParameter => IsWrite || (IsCondition && !IsConcurrencyToken);
 
-        public virtual string ParameterName
+        public override string ParameterName
             => _parameterName ?? (_parameterName = _generateParameterName());
 
-        public virtual string OriginalParameterName
+        public override string OriginalParameterName
```
AndriySvyryd: You can remove _parameterName  and _originalParameterName by adding a protected setter on the base property

src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -140,6 +140,25 @@ public virtual InternalEntityEntry GetOrCreateEntry(object entity)
         }
 
         /// <summary>
+        ///     TODO
+        /// </summary>
+        public virtual InternalEntityEntry GetOrCreateShadowEntryWithValues(IEntityType entityType, IDictionary<string, object> values)
```
AndriySvyryd: `GetOrCreateEntry(IDictionary<string, object> values, IEntityType entityType)`

src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -609,6 +628,15 @@ public virtual int SaveChanges(bool acceptAllChangesOnSuccess)
             }
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <returns></returns>
+        public virtual List<InternalEntityEntry> GetMigrationOperationsToRun()
```
AndriySvyryd: Making `GetEntriesToSave()` would be better, this is still in Internal

src/EFCore/Metadata/Internal/EntityType.cs
```diff
@@ -1709,6 +1709,32 @@ public override void OnTypeMemberIgnored(string name)
...
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual void AddSeedData(
+            [NotNull] object[] data)
+        {
+            var dicts = data.Select(d => d.GetType().GetRuntimeProperties().ToDictionary(x => x.Name, x => x.GetValue(d)));
```
AndriySvyryd: You can file this as a future improvement, but if the supplied object of the type mapped to this entity type then we should only use the mapped properties.

src/EFCore/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -1913,6 +1913,18 @@ public virtual IReadOnlyList<Property> GetActualProperties([CanBeNull] IReadOnly
         public virtual bool UsePropertyAccessMode(PropertyAccessMode propertyAccessMode, ConfigurationSource configurationSource)
             => HasAnnotation(CoreAnnotationNames.PropertyAccessModeAnnotation, propertyAccessMode, configurationSource);
 
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual InternalEntityTypeBuilder HasSeedData(
```
AndriySvyryd: This method is useless in the current form, you can just invoke `Metadata.AddSeedData` directly

test/EFCore.SqlServer.Tests/Migrations/SqlServerHistoryRepositoryTest.cs
```diff
@@ -165,10 +167,13 @@ private static IHistoryRepository CreateHistoryRepository(string schema = null)
                     new MigrationsModelDiffer(
                         new SqlServerTypeMapper(new RelationalTypeMapperDependencies()),
                         annotationsProvider,
-                        new SqlServerMigrationsAnnotationProvider(new MigrationsAnnotationProviderDependencies())),
+                        new SqlServerMigrationsAnnotationProvider(new MigrationsAnnotationProviderDependencies()),
+                        Mock.Of<IStateManager>(),
+                        Mock.Of<ICommandBatchPreparer>()),
```
AndriySvyryd: Don't use mocks as we are trying to remove all of them

src/EFCore/Metadata/Internal/EntityType.cs
tinchou: This is fixed. I moved this logic to `GetSeedData` where the `EntityType` is not mutable anymore, and so we can use `GetProperties`

src/EFCore.Relational/Infrastructure/EntityFrameworkRelationalServicesBuilder.cs
```diff
@@ -73,6 +72,7 @@ public class EntityFrameworkRelationalServicesBuilder : EntityFrameworkServicesB
                 { typeof(ICommandBatchPreparer), new ServiceCharacteristics(ServiceLifetime.Scoped) },
                 { typeof(IModificationCommandBatchFactory), new ServiceCharacteristics(ServiceLifetime.Scoped) },
                 { typeof(IIncludeExpressionVisitorFactory), new ServiceCharacteristics(ServiceLifetime.Scoped) },
+                { typeof(IMigrationsModelDiffer), new ServiceCharacteristics(ServiceLifetime.Scoped) },
```
tinchou: It's scoped because it now depends on the scoped `ICommandBatchPreparer` and `IStateManager`

tinchou:
LGTM? :)

ajcvickers:
@bricelam Can you look at this and make an assessment as to whether we can merge this into the product? Any more testing needed? Corner cases?

tinchou:
FYI:

When this is merged we should track the following:

* Enable `ModelSnapshotTest.SeedData_annotations_are_stored_in_snapshot` and `SeedData_for_multiple_entities_are_stored_in_model_snapshot`
* Write documentation
* Either use or remove the row count in the operations generated for the migration by the update pipeline
* Batch multiple migration operations in one call
* Support navigations (needs shadow navigations)
* Use SeedData in the InMemory provider
* Discuss the possibility of adding a property bag for shadow properties

src/EFCore/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -663,10 +663,12 @@ public virtual void AddRangeToCollectionSnapshot([NotNull] IPropertyBase propert
...
             {
-                object value;
-                return _storeGeneratedValues.TryGetValue(propertyBase, out value)
-                    ? value
-                    : ReadPropertyValue(propertyBase);
+                var isShadowNavigation = propertyBase is INavigation && propertyBase.IsShadowProperty;
+                return isShadowNavigation
+                    ? null
+                    : _storeGeneratedValues.TryGetValue(propertyBase, out var value)
+                        ? value
```
ajcvickers: Is this just a temporary hack to make things work for seeding? Seems like if seeding needs shadow nav props, then we should make it work correctly, rather than with this kind of special-casing.

src/EFCore/ChangeTracking/Internal/InternalEntityEntryFactory.cs
```diff
@@ -32,6 +32,7 @@ private static InternalEntityEntry NewInternalEntityEntry(IStateManager stateMan
         {
             if (!entityType.HasClrType())
             {
+                // TODO use the entity parameter to populate the InternalShadowEntityEntry
                 return new InternalShadowEntityEntry(stateManager, entityType);
```
ajcvickers: Do we need an issue to track this, or is this something that should be done before shipping?

src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -155,6 +155,36 @@ public virtual InternalEntityEntry GetOrCreateEntry(object entity)
...
+            if (entry == null)
+            {
+                object entity;
+                _trackingQueryMode = TrackingQueryMode.Multiple;
+
+                if (entityType.HasClrType())
+                {
+                    entity = Activator.CreateInstance(entityType.ClrType);
+                    entry = _factory.Create(this, entityType, entity);
+                }
```
ajcvickers: When is this path used? Generally we avoid using Activator.CreateInstance directly. We have factories setup to create entity instances already, so we should try to wire these through.

src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -155,6 +155,36 @@ public virtual InternalEntityEntry GetOrCreateEntry(object entity)
...
+                else
+                {
+                    entry = new InternalShadowEntityEntry(this, entityType);
+                    entity = entry;
+                }
```
ajcvickers: Are we sure this is safe? Could you end up with two state entries referencing the same entity?

src/EFCore/Metadata/IEntityType.cs
```diff
@@ -112,5 +112,16 @@ public interface IEntityType : ITypeBase
...
+        /// <summary>
+        ///     <para>
+        ///         Gets the seed data defined on this entity.
+        ///     </para>
+        ///     <para>
+        ///         This API should only be called after model configuration is finished.
+        ///     </para>
+        /// </summary>
+        /// <returns> The seed data defined on this entity. </returns>
+        IEnumerable<IDictionary<string, object>> GetSeedData();
```
ajcvickers: >  IEnumerable<IDictionary<string, object>> GetSeedData(); [](start = 6, length = 57)

This is returning mutable data from a non-mutable API surface. Is this okay?

src/EFCore/Metadata/IEntityType.cs
```diff
@@ -112,5 +112,16 @@ public interface IEntityType : ITypeBase
         /// </summary>
         /// <returns> The properties defined on this entity. </returns>
         IEnumerable<IProperty> GetProperties();
+
+        /// <summary>
+        ///     <para>
+        ///         Gets the seed data defined on this entity.
+        ///     </para>
+        ///     <para>
+        ///         This API should only be called after model configuration is finished.
```
ajcvickers: >This API should only be called after model configuration is finished. [](start = 20, length = 69)

Why is this? This differs from all our other metadata APIs.

src/EFCore/Metadata/IMutableEntityType.cs
```diff
@@ -198,5 +198,17 @@ public interface IMutableEntityType : IEntityType, IMutableTypeBase
...
+        ///         Adds objects as seeds on this entity.
+        ///     </para>
+        ///     <para>
+        ///         This API can take either typed or anonymous objects. This allows to pass values such as
+        ///         a foreign key that is defined in your model by convention.
+        ///     </para>
+        /// </summary>
+        /// <param name="data"> The data to seed to this entity. </param>
+        void AddSeedData([NotNull] params object[] data);
     }
```
ajcvickers: Does calling this replace all the existing seed data, or add to it? If it adds, how are things merged?

src/EFCore/Metadata/Internal/EntityType.cs
```diff
@@ -1740,6 +1743,21 @@ public override void OnTypeMemberIgnored(string name)
...
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual IEnumerable<IDictionary<string, object>> GetSeedData()
+            => _seedData.Select(seed =>
+                GetProperties() // we'll ignore navigations and invalid properties on the seeds
+                    .Select(p => seed.GetType().GetRuntimeProperty(p.Name))
+                    .Where(p => p != null)
+                    .ToDictionary(p => p.Name, p => p.GetValue(seed)));
+
```
ajcvickers: Does this work with shadow properties? (Not sure I understand what is happening here...)

rowanmiller:
@tinchou any chance you would want to rebase this on dev? I'd love to demo it as a future feature at Build Conference, but it doesn't build at the moment.

tinchou:
I will do it as soon as I have my machine running again (I'm having
technical problems :( ). An option would be to use tinchou/entityframework
in the meantime.

And let me know if I can be of any help!

src/EFCore/ChangeTracking/Internal/InternalEntityEntry.cs
tinchou: Shadow navs would be very well received because:

1. this workaround wouldn't be necessary
1. they would provide a richer experience

This is needed because the snapshot model is 100% shadow, and currently our code throws exceptions when using shadow navs.

src/EFCore/ChangeTracking/Internal/StateManager.cs
tinchou: This is linked to your other comment here: https://github.com/aspnet/EntityFramework/pull/7824/files/9c2a5b8def1f5e37875b43a7b9a217a6314f327a#r114194446

This comes from `MigrationsModelDiffer.DiffSeedData`. I believe we can make it more specific and just keep the `CreateEntry` part. We're using `StateManager.TryGetEntry(key, keyValues)` to retrieve them later on, and not this method because the `values` dictionary is not tracked.

The branch in particular that falls in `entityType.HasClrType()` is the one where the model is the current one (and not a snapshot), where we can have real entities (and not only shadow ones).

I had to instantiate an empty entity and then call `CurrentValues.SetValues(values)` because:

1. `InternalEntityEntryFactory.NewInternalEntityEntry` doesn't support dictionaries
1. `InternalEntityEntryFactory.NewInternalEntityEntry` doesn't allow null values
1. I couldn't find a factory to create empty entities

src/EFCore/ChangeTracking/Internal/StateManager.cs
tinchou: This method receives just a `values` Dictionary (akin to a bag of properties, which is not an `entity`). I think we should rename it to `CreateEntry` as I mentioned in another comment so people don't expect to get the same entry twice.

src/EFCore/Metadata/IEntityType.cs
tinchou: If you check the method implementation, it uses the property information to create the dictionary. It **needs** the configuration to be finished to be able to use that.

It could be changed to "This API only returns meaningful data after model configuration", which is also probably the case with `GetProperties` and other similar methods.

src/EFCore/Metadata/IEntityType.cs
tinchou: True. And even if you mutate it, the implementation always returns a new object. What's the signature for this kind of immutable data?

src/EFCore/Metadata/IMutableEntityType.cs
tinchou: Can be called multiple times to add. Does the documentation need clarification?

Implementation is:

```csharp
void IMutableEntityType.AddSeedData(params object[] data) => _seedData.UnionWith(data);
```

Where `_seedData` is a `HashSet<object>`.

src/EFCore/Metadata/Internal/EntityType.cs
tinchou: Shadow props, not shadow navs. This code:

1. gets all the property names for this entity (excluding navigations)
1. gets the value from the seed objects through reflection
1. creates a dictionary of { propertyName: seedValue }

Note:

1. We ignore navs
1. We ignore properties missing in the seed object
1. We ignore properties in the seed object that are not in the model
1. This could be cached, but since it's only used in the differ we are not doing that (premature optimization?)

src/EFCore/ChangeTracking/Internal/InternalEntityEntryFactory.cs
tinchou: I don't know, it's hard for me to say. This code is clearly ignoring the entity, and it could be used to pre-populate the shadow entity. Probably worth opening an issue if you plan to extend shadow entity support.

tinchou:
@AndriySvyryd in commit https://github.com/aspnet/EntityFramework/commit/f5af64f38dc55cb857a5598765c2eb0678ae62a9 you introduced some changes that added complexity to the `MigrationsModelDiffer`. It seems that now you can have a mapping from n `IEntityType`s to n `IEntityType`s.

This seems to add a lot of complexity to the `MigrationsModelDiffer.DiffSeedData` that I had implemented for the 1 to 1 `IEntityType` mapping that we supported at the time I was there.

For example, let's imagine this schema migration:

1. I had a `User` `IEntityType` with PK `User_Id` and properties `Address_City`, `Address_Country`
1. Now I have `User` `IEntityType` with PK `User_Id` and  `Address` `IEntityType` with PK `Address_Id`, FK `User_Id` and properties `City`, `Country`

I have no idea how we're doing the schema migration in this case, but let's assume this is already solved. For the data migration I would have to:

1. Find the PK of the `User` `IEntityType`
1. Find the `User`s with that PK and update them
1. Look up which `IEntityType`s have `User_Id` as FK (in this case `Address`)
1. Find the `Address`es with that FK and update them

Am I right? How would you go about this?

@ajcvickers it seems this will take some time I might not have right now, unless we find a clever way of doing it. I was ready to address PR feedback, but this is not a minor change :(.

Maybe the team can take a look/give me a hand? Thanks!

tinchou:
One option would be to update this PR to check if `TableMapping.EntityTypes.Length == 1` for both source and target. Then the problem would be reduced to what was working before, and we could open a new issue to track support for table splitting + data migration.

The downside is that if we don't have enough time for addressing that issue before shipping, the feature would be somewhat incomplete.

On the other hand, before Andriy's commit we didn't even have support for table splitting in migrations. This feature would still work for most scenarios (that don't involve table splitting), and still has a lot of value even without table splitting.

divega:
> One option would be to update this PR to check if TableMapping.EntityTypes.Length == 1 for both source and target. Then the problem would be reduced to what was working before, and we could open a new issue to track support for table splitting + data migration.

That is exactly my preference: it would be great if you could address the rest of the feedback and leave this as a well known limitation (i.e. throw `NotSupportedException`) that can be addressed in a separate PR. 

There is indeed a decision we need to make on when we can consider the feature complete enough to ship it. This should help.

ajcvickers:
@tinchou @divega Just so you guys are aware, I did an updated build for @rowanmiller to use, so I think he is unblocked now. I think it's fine in general to have a limitation at this point, but let's make sure that we a) have an issue tracking it and b) at least do the best we can to ensure that anything we do to remove the limitation later won't be a breaking change--usually done by making the code throw for now.

tinchou:
@ajcvickers can you share your fixed build? Drop me a zip in an e-mail if that is easier for you!

Regarding the exception you propose... I believe it would prevent you from generating the migration at all. I'm fine with that, but maybe it's a little overprotective. If the data migration ignored split tables for now, and at some point later in time it started recognizing them, would that be a breaking change or just a new feature? I don't foresee any API change if we change the `DiffSeedData` method to receive `TableMapping` types now.

ajcvickers:
@tinchou I can send you the code when I am back in the office, but I doubt it will help much. It's not production quality--it's just hacked to build so Rowan can use it to demo seeding assuming he doesn't also do table splitting.

Agree that we can't disable Migrations entirely. I suspect we can probably make things work in most cases, and only throw when things get wonky. But I'll need to spend some more time looking at it.

divega:
Do we have the ability to generate a warn comment in the migration? This would be a good use for that.

bricelam:
Comments are tracked by #2461

src/EFCore.Relational/Migrations/MigrationBuilder.cs
```diff
@@ -684,17 +684,15 @@ public virtual AlterOperationBuilder<AlterDatabaseOperation> AlterDatabase()
             [NotNull] string table,
             [CanBeNull] string schema = null,
             [CanBeNull] string[] columns = null,
-            [CanBeNull] object[,] values = null)
```
bricelam: Can convert a single migrations operation into multiple update operations. Allowing multiple rows in the migrations operations would also allow us to group like statements into a single method call in the migration.

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -634,45 +634,26 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
...
 
-            if (operation.Values.Length == 0)
-            {
-                return;
-            }
-
+            // TODO figure out if we could turn off this check only when it's needed
             builder
-                .Append("INSERT INTO ")
+                .Append("SET IDENTITY_INSERT ")
```
bricelam: This is SQL Server specific and shouldn't be in the base `MigrationsSqlGenerator`

tinchou:
Ping

AndriySvyryd:
@tinchou Thanks for your contribution! (better late than never 😄)

tinchou:
Wohoooo, amazing!

PR #7825, opened 2017-03-09
Make appveyor builds opt-in

natemcmaster:
EF is hogging the AppVeyor queue. It's builds take so long, and devs are not generally cancelling builds when they push new commits to a PR.

This will make running on AppVeyor opt-in. Add the word 'appveyor' to the commit message to run on AppVeyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,6 @@
+only_commits:
+  # Start a build if commit message contains 'appveyor'
+  message: /appveyor/
 init:
   - git config --global core.autocrlf true
 build_script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,5 +26,4 @@ matrix:
   fast_finish: true
 artifacts:
   - path: 'artifacts\build\*.nupkg'
-# Required for dotnet-test to work
 os: Visual Studio 2017 RC


ErikEJ:
How should external contributors respond to this ? (ie add appveyor tag, or not?)

natemcmaster:
Example commit message:
```
Make EF 10x faster with this one great trick. You'll never guess what happened next

[appveyor test]
```

Eilon:
Eh, I agree with the sentiment of this change, but I don't think it's right. It's asking people to do extra work that is not discoverable. I'd rather find a way to make this be smoother. We can look into getting more AppVeyor queues or something.

natemcmaster:
Straw poll. @anpete @ajcvickers @bricelam @smitpatel how often is AppVeyor providing useful feedback? When I was on EF, it seemed like AppVeyor failures were usually unrelated to PR changes, and that it took so long to get a full build done that we didn't bother waiting for AppVeyor to pass before merging.

ajcvickers:
@natemcmaster For me it's become totally tuned out. I honestly could not tell you whether it has passed or failed on any recent PRs because I became so used to it being broken that I stopped looking.

divega:
Another part of the question is:
- Can it be made to work really well with EF Core? (not sure if additional AppVeyor queues would do that, is it worth trying?)
-  What value would it yield if it worked? (I am going to assume there is value, because otherwise such services wouldn't exist and no one would use them)

Assuming it starts working ok and there is value, people that stopped looking long time ago will need to learn to look again at it, but that shouldn't be a problem :smile:

If the answer is that it cannot be made to work well with our codebase then rather than making it an opt in, we can disable it or remove it.

bricelam:
AppVeyor may not be the right tool for EF. If I remember right, it killed the jobs because our tests take too long. Maybe with reduced verification it could be useful.

smitpatel:
In past AppVeyor used to have unrelated issues but at some point was running fine. Though in last few months it has been too broken that even I have been ignoring it.

I never used AppVeyor as a verification to my PR before merging as I run build.cmd before merging anyway. And EF also have own ci which does same thing as AppVeyor for dev branch. Travis would provide good feedback in that scenario being tested on different OS. To me, it is a good to have PRs building especially for external contributors, who sometimes cannot get build.cmd to work (or don't run it before submitting PR). In such PRs, it is really helpful to get verification done which avoids pulling the branch in a building locally. (We can also look at the output of failed PR and point in correct direction for test fixes may be). Another usefulness would be not running build.cmd always on local machine. If there is something verifying then I can get away with just running tests in VS & running command line build only at merge time. I am uncertain how many people in team use it this way.

I believe, having at least one tool which would run all tests on PR can be helpful. If travis runs SQL server tests then we can just get rid of AppVeyor fully. Or based on our build time & load AppVeyor may not be best option for us.

For this particular PR, if instead of doing opt-in based on commit message, if we can do trigger based on PR labels or comment on PR that would be a better thing as we can trigger build on external PRs. Building dev on AppVeyor is totally unnecessary. & internal contribution PR are optional.

bricelam:
Oh I like the idea of getting AppVeyor to run *just* the tests that don't (can't) run on Travis

ajcvickers:
I think @smitpatel is right that it has more value for external contributions where people don't always follow the process of testing before submitting. However, it's only useful in any situation if it's basically reliable, which it isn't at the moment. I would be in favor of merging this for now so that it unblocks the queue for repros where it is adding value, with a follow up item to find a way or ways to make it more reliable/useful.

divega:
@ajcvickers Agreed. BTW, this was already merged and I created follow up issue https://github.com/aspnet/EntityFramework/issues/7837.

PR #7826, opened 2017-03-09
Query: Improve the logic of null expansion in join conditions with an…

smitpatel:
…onymous types

@maumar @anpete 

Due to the way LINQ processes `equals` with normal properties comparison & anonymous type comparison.

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -484,6 +484,14 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 && rightExpressions != null
                 && leftExpressions.Length == rightExpressions.Length)
             {
+                if (leftExpressions.Length == 1
+                    && expressionType == ExpressionType.Equal)
+                {
+                    var translatedExpression = TransformNullComparison(leftExpressions[0], rightExpressions[0], binaryExpression.NodeType)
+                                               ?? Expression.MakeBinary(expressionType, leftExpressions[0], rightExpressions[0]);
+                    return Expression.AndAlso(translatedExpression, Expression.Constant(true, translatedExpression.Type));
+                }
+
                 return leftExpressions
                     .Zip(rightExpressions, (l, r) =>
                         TransformNullComparison(l, r, binaryExpression.NodeType)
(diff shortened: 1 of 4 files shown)


src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -484,6 +484,14 @@ private static Expression UnfoldStructuralComparison(ExpressionType expressionTy
                 && rightExpressions != null
                 && leftExpressions.Length == rightExpressions.Length)
             {
+                if (leftExpressions.Length == 1
+                    && expressionType == ExpressionType.Equal)
+                {
+                    var translatedExpression = TransformNullComparison(leftExpressions[0], rightExpressions[0], binaryExpression.NodeType)
+                                               ?? Expression.MakeBinary(expressionType, leftExpressions[0], rightExpressions[0]);
```
maumar: use ExpressionType.Equal rather than expressionType here, just so its more clear

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -484,6 +484,14 @@ private static Expression UnfoldStructuralComparison(ExpressionType expressionTy
...
+                    return Expression.AndAlso(translatedExpression, Expression.Constant(true, translatedExpression.Type));
```
maumar: Expression.Constant(true), since translatedExpression is guaranteed to be bool

smitpatel: I actually forgot to submit updated. Will fix in future PR.

smitpatel: Is `TranslatedExpression` is guaranteed to return bool type?

smitpatel: updated in #7843

maumar: It should be but it's not that clear. I guess it is safer to leave it as is after all

smitpatel: I concluded this way:
TransformNullComparison returns either IsNullExpression or Not(IsNullExpression).
IsNullExpression as type hardcoded to bool. Not expression takes type from the inner expression so it have to bool only in this case.

PR #7839, opened 2017-03-10
Query: Relational: Introduce a new LeftOuterJoinClause class

tuespetre:
These changes introduce a new LeftOuterJoinClause class so that
the GroupJoin/SelectMany/DefaultIfEmpty pattern can be detected
and made explicit at optimization time, rather than at runtime,
which required modifications to the query model during compilation
(and complex code to compensate for such modifications.)

### DIFF
--- a/src/EFCore/Query/EntityQueryModelVisitor.cs
+++ b/src/EFCore/Query/EntityQueryModelVisitor.cs
@@ -200,6 +200,11 @@ namespace Microsoft.EntityFrameworkCore.Query
 
                 OptimizeQueryModel(queryModel, includeResultOperators);
 
+                QueryCompilationContext.Logger
+                    .LogDebug(
+                        CoreEventId.OptimizedQueryModel,
+                        () => CoreStrings.LogOptimizedQueryModel(Environment.NewLine, queryModel.Print()));
+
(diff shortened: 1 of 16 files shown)


tuespetre:
@maumar this is the idea I was talking about in #7772. It makes the whole `HandlesQuerySource` thing a lot more straightforward and removes the `PopulateQueryModelMapping/RecreateQueryModelFromMapping` thing entirely instead of needing to pepper it in more places.

maumar:
@tuespetre lets work on this refactoring after #7772 is in.

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -199,6 +199,11 @@ var includeResultOperators
 
                 OptimizeQueryModel(queryModel, includeResultOperators);
 
+                QueryCompilationContext.Logger
```
tuespetre: This needs to be in both methods (async/non-async)

tuespetre:
:up::date: with re:b:ase

smitpatel:
We are closing this pull request due to lack of response. If you decide to work on the issue again then feel free to reopen the PR and push more commits to it. We would be happy to review it.
Thank you helping us out in improving EF Core.

PR #7842, opened 2017-03-10
[experiment] Run SQL Server tests on Travis

natemcmaster:


natemcmaster:
More details:
https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-setup-docker
https://docs.travis-ci.com/user/docker/

natemcmaster:
Finally got the connection string right. It looks like the SqlServer.FunctionalTests project is able to connect and run against the mssql server container correctly. The only concern I have is that Travis likes to kill builds that don't produce console output for 10 minutes. It also kills builds that produce too much log output. We may have to add some extra console output to that test, but not too much.

bricelam:
Can we give them money to avoid this?

>The job exceeded the maximum time limit for jobs, and has been terminated.

bricelam:
Can we [give them money](https://travis-ci.com/plans) to avoid this?

>The job exceeded the maximum time limit for jobs, and has been terminated.

natemcmaster:
cc @Eilon - this runs too long for Travis. Do we need an upgraded plan to get support for long-running builds?

bricelam:
Another alternative they suggest is splitting it up into multiple jobs (using a build matrix). E.g. One for the SQL Server functionals and one for everything else.

natemcmaster:
I'll give that a try.

ajcvickers:
Maybe we should bite the bullet and just create the "fast" subset of tests...

Eilon:
We already do pay for Travis... quite a lot, in fact. But either way, 50min sounds like too much anyway, no?

natemcmaster:
@bricelam @ajcvickers I'm going to close this. It seems this experiment has failed due to long-running SQL server tests. I'll leave the branch and the code if someone from EF would like to pick it up from here.

MathieuAuclair:
@natemcmaster  Why don't you use ```travis_wait``` ? You can set a command timeout this way.

PR #7843, opened 2017-03-10
Query: Lift Subqueries when outer QM's select can be pushed down into…

### LINKED ISSUE #7831: Query: Optimize QueryOptimizer
`QueryModelOptimizer` derives from `SubQueryFromClauseFlattener` class which in turn derives from `QueryModelVisitorBase`.
- QMVBase class does not visit, QueryModels recursively, hence subQMs inside SelectClause etc are not visited by optimizer. We need ExpressionVisitor to apply QueryModelOptimizer recursively.
- Expand logic of lifting of subquery.
- Verify if we need to derive from `SubQueryFromClauseFlattener`, its code structure has a specific format which we can get rid of.

smitpatel:
… subqueryQM

Apply QueryOptimizer recursively on all QueryModels
Introduce SubQueryExpressionVisitor<IQueryModelVisitor> to apply QMV's recursively. @maumar 

Partially fix #7831

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -487,14 +487,14 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 if (leftExpressions.Length == 1
                     && expressionType == ExpressionType.Equal)
                 {
-                    var translatedExpression = TransformNullComparison(leftExpressions[0], rightExpressions[0], binaryExpression.NodeType)
-                                               ?? Expression.MakeBinary(expressionType, leftExpressions[0], rightExpressions[0]);
-                    return Expression.AndAlso(translatedExpression, Expression.Constant(true, translatedExpression.Type));
+                    var translatedExpression = TransformNullComparison(leftExpressions[0], rightExpressions[0], expressionType)
+                                               ?? Expression.Equal(leftExpressions[0], rightExpressions[0]);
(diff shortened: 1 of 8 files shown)


src/EFCore/Query/ExpressionVisitors/Internal/SubQueryExpressionVisitor.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitSubQuery(SubQueryExpression expression)
+        {
+            TransformingVisitor.VisitQueryModel(expression.QueryModel);
+            return expression;
```
maumar: nit: add newline before return

smitpatel: thanks for reminding, it should be changed to base visitor actually.

smitpatel: And base returns expression only. :trollface:

smitpatel: @maumar - Decide what you want? `return Expression` or call base? 😛

maumar: call base is "more correct" ;)

smitpatel:
@maumar - Updated.
I will submit changes to ExpressionVisitorBase in separate PR.

PR #7845, opened 2017-03-10
Query: Optimize certain types of GroupJoin/SelectMany queries

tuespetre:
In some circumstances, a GroupJoin/SelectMany pattern (with or
without DefaultIfEmpty) can be optimized by shifting body clauses
from the AdditionalFromClause into the GroupJoinClause. These changes
implement that optimization.

### DIFF
--- a/src/EFCore/Query/EntityQueryModelVisitor.cs
+++ b/src/EFCore/Query/EntityQueryModelVisitor.cs
@@ -299,6 +299,13 @@ namespace Microsoft.EntityFrameworkCore.Query
             Check.NotNull(queryModel, nameof(queryModel));
             Check.NotNull(includeResultOperators, nameof(includeResultOperators));
 
+            // First pass of optimizations
+
+            var additionalFromClauseOptimizer
+                = new AdditionalFromClauseOptimizingQueryModelVisitor();
+
+            additionalFromClauseOptimizer.VisitQueryModel(queryModel);
+
             _queryOptimizer.Optimize(QueryCompilationContext.QueryAnnotations, queryModel);
 
             var entityEqualityRewritingExpressionVisitor
@@ -307,6 +314,8 @@ namespace Microsoft.EntityFrameworkCore.Query
             entityEqualityRewritingExpressionVisitor.Rewrite(queryModel);
 
             queryModel.TransformExpressions(_subQueryMemberPushDownExpressionVisitor.Visit);
+
+            // Rewrite navigations
 
             new NondeterministicResultCheckingVisitor(QueryCompilationContext.Logger)
                 .VisitQueryModel(queryModel);
@@ -319,9 +328,15 @@ namespace Microsoft.EntityFrameworkCore.Query
             _navigationRewritingExpressionVisitorFactory
                 .Create(this).Rewrite(queryModel, parentQueryModel: null);
 
+            // Second pass of optimizations
+
+            additionalFromClauseOptimizer.VisitQueryModel(queryModel);
+
             entityEqualityRewritingExpressionVisitor.Rewrite(queryModel);
 
             queryModel.TransformExpressions(_subQueryMemberPushDownExpressionVisitor.Visit);
+
+            // Log results
 
             QueryCompilationContext.Logger
                 .LogDebug(
--- a/src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
@@ -1,0 +1,93 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Query.Internal;
+using Remotion.Linq;
+using Remotion.Linq.Clauses;
+using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Clauses.ExpressionVisitors;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class AdditionalFromClauseOptimizingQueryModelVisitor : QueryModelVisitorBase
+    {
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public override void VisitQueryModel([NotNull] QueryModel queryModel)
+        {
+            queryModel.TransformExpressions(new TransformingQueryModelExpressionVisitor<AdditionalFromClauseOptimizingQueryModelVisitor>(this).Visit);
+
+            base.VisitQueryModel(queryModel);
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public override void VisitAdditionalFromClause(
+            [NotNull] AdditionalFromClause fromClause,
+            [NotNull] QueryModel queryModel,
+            int index)
+        {
+            if (fromClause.FromExpression is SubQueryExpression subQueryExpression
+                && subQueryExpression.QueryModel.MainFromClause.FromExpression is QuerySourceReferenceExpression qsre
+                && subQueryExpression.QueryModel.SelectClause.Selector is QuerySourceReferenceExpression
+                && qsre.ReferencedQuerySource is GroupJoinClause groupJoinClause
+                && !(groupJoinClause.JoinClause.InnerSequence is SubQueryExpression)
+                && queryModel.CountQuerySourceReferences(groupJoinClause) == 1
+                && subQueryExpression.QueryModel.BodyClauses.Any()
+                && subQueryExpression.QueryModel.BodyClauses.All(c => c is WhereClause))
+            {
+                var newMainFromClause = new MainFromClause(
+                    groupJoinClause.JoinClause.ItemName,
+                    groupJoinClause.JoinClause.ItemType,
+                    groupJoinClause.JoinClause.InnerSequence);
+
+                var newSelectClause = new SelectClause(
+                    new QuerySourceReferenceExpression(newMainFromClause));
+
+                var newSubQueryModel = new QueryModel(newMainFromClause, newSelectClause);
+
+                ShiftBodyClauses(subQueryExpression.QueryModel, newSubQueryModel);
+
+                var newSubQueryExpression = new SubQueryExpression(newSubQueryModel);
+
+                groupJoinClause.JoinClause.InnerSequence = newSubQueryExpression;
+
+                if (!subQueryExpression.QueryModel.ResultOperators.Any())
+                {
+                    fromClause.FromExpression = qsre;
+                }
+            }
+        }
+
+        private void ShiftBodyClauses(QueryModel oldQueryModel, QueryModel newQueryModel)
+        {
+            var querySourceMapping = new QuerySourceMapping();
+
+            querySourceMapping.AddMapping(
+                oldQueryModel.MainFromClause,
+                new QuerySourceReferenceExpression(newQueryModel.MainFromClause));
+
+            foreach (var bodyClause in oldQueryModel.BodyClauses.ToArray())
+            {
+                bodyClause.TransformExpressions(expression =>
+                    ReferenceReplacingExpressionVisitor.ReplaceClauseReferences(
+                        expression,
+                        querySourceMapping,
+                        throwOnUnmappedReferences: false));
+
+                oldQueryModel.BodyClauses.Remove(bodyClause);
+                newQueryModel.BodyClauses.Add(bodyClause);
+            }
+        }
+    }
+}
--- a/test/EFCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
@@ -7188,6 +7188,60 @@ WHERE ([o].[OrderID] > 11000) AND [o].[OrderID] IN (
                 Sql);
         }
 
+        public override void GroupJoin_SelectMany_subquery_with_filter()
+        {
+            base.GroupJoin_SelectMany_subquery_with_filter();
+
+            Assert.Equal(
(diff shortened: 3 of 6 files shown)


src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
```diff
@@ -0,0 +1,99 @@
...
+                && subQueryExpression.QueryModel.MainFromClause.FromExpression is QuerySourceReferenceExpression qsre
+                && subQueryExpression.QueryModel.SelectClause.Selector is QuerySourceReferenceExpression
+                && qsre.ReferencedQuerySource is GroupJoinClause groupJoinClause
+                && queryModel.CountQuerySourceReferences(groupJoinClause) == 1
+                && subQueryExpression.QueryModel.BodyClauses.Any()
+                && subQueryExpression.QueryModel.BodyClauses.All(c => c is WhereClause))
+            {
+                if (groupJoinClause.JoinClause.InnerSequence is SubQueryExpression innerSequenceSubQuery)
+                {
+                    if (!innerSequenceSubQuery.QueryModel.ResultOperators.Any(r => r is TakeResultOperator || r is SkipResultOperator))
```
maumar: can you add more tests? this case doesn't seem to be covered by any of the existing ones

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
```diff
@@ -0,0 +1,99 @@
...
+            [NotNull] QueryModel queryModel, 
+            int index)
+        {
+            if (fromClause.FromExpression is SubQueryExpression subQueryExpression
+                && subQueryExpression.QueryModel.MainFromClause.FromExpression is QuerySourceReferenceExpression qsre
+                && subQueryExpression.QueryModel.SelectClause.Selector is QuerySourceReferenceExpression
+                && qsre.ReferencedQuerySource is GroupJoinClause groupJoinClause
+                && queryModel.CountQuerySourceReferences(groupJoinClause) == 1
+                && subQueryExpression.QueryModel.BodyClauses.Any()
+                && subQueryExpression.QueryModel.BodyClauses.All(c => c is WhereClause))
```
maumar: anything preventing us from translating more body clauses/result operators? e.g. with Skip/Take people often add OrderBys

tuespetre: Do you have an example? An additional from clause using OrderBy().Take() is a lateral join, whereas OrderBy().Take() in the group join is not. WhereClause is the only reasonable thing to bother shifting into the group join clause (and only when Skip/Take/Concat/etc are not present on the group join's inner sequence)

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
tuespetre: For sure, need to ensure other bases are covered there too (group join's selector must be a QSRE, cannot have any Concat/Except/Intersect/Union/GroupBy, Distinct should work)

tuespetre: Hmm.... might punt on this part for now and skip the case altogether. There are problems with query compilation. Will file a separate issue

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
```diff
@@ -0,0 +1,90 @@
...
+using Remotion.Linq.Clauses.ExpressionVisitors;
+using Remotion.Linq.Clauses.ResultOperators;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class AdditionalFromClauseOptimizingQueryModelVisitor : QueryModelVisitorBase
```
smitpatel: Is separate Visitor necessary? This does not seem to be doing anything specific which relates to any kind of state between qmv & ev. This should be part of `QueryOptimizer`.

tuespetre: Sounds good to me, checked out `SubQueryFromClauseFlattener` source and it shouldn't hinder anything: https://github.com/re-motion/Relinq/blob/develop/Core/Transformations/SubQueryFromClauseFlattener.cs#L86

tuespetre: @smitpatel scratch that. It's no good because then we miss the optimization on queries after rewriting navigations into them.

smitpatel: Do nav rewrite before query optimizer ?

tuespetre: That causes lots of problems. I think this is a chicken/egg scenario like the entity equality rewriting visitor. We could run it both before and after.

tuespetre: @smitpatel I'm going to put together another PR that runs query optimizer before and after nav rewrites so we can review the difference. I think doing so will make it much clearer that we can reliably place this optimization and any other ad-hoc optimizations into the QueryOptimizer in the future.

smitpatel: Entity-equality rewrite is not chicken/egg scenario. It is the bug (or bad assumptions) in nav rewrite. Both of them affects different kind of Expressions. EntityEquality should only deal with binary expressions where qsre are being equated whereas nav rewrite should only deal with member access expressions where members are navigation. The deterministic order is to run nav rewrite first and expand all navigations & then run entity equality rewrite to convert qsre to key property access (including new qsre introduced by nav rewrite). There is one tricky step that nav rewrite optimizes navigation key access by writing it in forum of FK property instead of principal key to avoid join (or may not still having join unsure). Though even if nav rewrite does that it should not affect entity equality because there is no qsre comparison anymore.

The overall structure of `OptimizeQueryModel` is not very well defined. We should be running all custom stuff like nav rewrite, entity equality rewrite, the new include compiler @anpete introducing, possibly `SubQueryMemberPushDownEV` (which may go in `QueryOptimizer` too) and then run the QueryOptimizer on all of that to generate the optimal QM we can get.

@maumar @anpete

tuespetre: That may be so with entity equality but `QueryOptimizer` is just going to be better off running before and after the navigation rewrites. There are subqueries that can only be flattened before the navigations are rewritten and there are subqueries that can only be flattened afterwards. Not just subquery flattening but optimizations in general (subquery member pushdown, optimizations like this one, etc.)

maumar: we have a bug for this - https://github.com/aspnet/EntityFramework/issues/7773 we can add this scenario to the pile

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
```diff
@@ -0,0 +1,90 @@
...
+            int index)
+        {
+            if (fromClause.FromExpression is SubQueryExpression subQueryExpression
+                && subQueryExpression.QueryModel.MainFromClause.FromExpression is QuerySourceReferenceExpression qsre
+                && subQueryExpression.QueryModel.SelectClause.Selector is QuerySourceReferenceExpression
+                && qsre.ReferencedQuerySource is GroupJoinClause groupJoinClause
+                && !(groupJoinClause.JoinClause.InnerSequence is SubQueryExpression)
+                && queryModel.CountQuerySourceReferences(groupJoinClause) == 1
+                && subQueryExpression.QueryModel.BodyClauses.Any()
+                && subQueryExpression.QueryModel.BodyClauses.All(c => c is WhereClause))
```
maumar: we can also do the same for Order clauses - it's maybe not very useful to write orderby in this context but it *should* be safe

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public override void VisitAdditionalFromClause(
+            [NotNull] AdditionalFromClause fromClause,
+            [NotNull] QueryModel queryModel,
+            int index)
+        {
+            if (fromClause.FromExpression is SubQueryExpression subQueryExpression
```
maumar: this optimization could be done as part of nav rewrite (RewriteNavigationIntoGroupJoin method - we are already in the right place so the bulk of the condition below can be removed, we just need to analyze the body clauses)

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
maumar: I found some issues myself when digging, this area is definitely not a bug-free zone ;)

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
maumar: yeah you are right

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
tuespetre: Should the navigation rewriter be responsible for general purpose optimizations like this, though?

maumar: Ahh that's right. User could manually write GroupJoin with a filter and we would miss this optimization. Let's keep it where it is.

maumar: please add test for that (that hits this scenario using manually created groupjoin, that doesn't go thru nav rewrite)

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
tuespetre: Sounds good

src/EFCore/Query/ExpressionVisitors/Internal/AdditionalFromClauseOptimizingQueryModelVisitor.cs
tuespetre: Added a test and see an issue related to the other query compilation bug I found with additional from clauses. Looks like this optimization needs to run before any subquery flattening; once again, something that should be run both before and after navigation rewriting.

tuespetre:
:up::date: added several tests for manual group join scenarios.

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -299,15 +299,24 @@ protected virtual void ExtractQueryAnnotations([NotNull] QueryModel queryModel)
             Check.NotNull(queryModel, nameof(queryModel));
             Check.NotNull(includeResultOperators, nameof(includeResultOperators));
 
-            _queryOptimizer.Optimize(QueryCompilationContext.QueryAnnotations, queryModel);
+            var additionalFromClauseOptimizer
+                = new AdditionalFromClauseOptimizingQueryModelVisitor();
 
             var entityEqualityRewritingExpressionVisitor
```
maumar: nit: move this bit lower (just before it's called)

test/EFCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -7188,6 +7188,66 @@ public override void Contains_with_subquery_involving_join_binds_to_correct_tabl
...
+) AS [t] ON [c].[CustomerID] = [t].[CustomerID]",
+                Sql);
+        }
+
+        public override void GroupJoin_SelectMany_subquery_with_filter_orderby_and_DefaultIfEmpty()
+        {
+            base.GroupJoin_SelectMany_subquery_with_filter_orderby_and_DefaultIfEmpty();
+
+            Assert.Equal(
+                @"SELECT [o0].[OrderID], [o0].[CustomerID], [o0].[EmployeeID], [o0].[OrderDate]
```
maumar: I would expect this to be translated to a single query, like so:

```
SELECT [t].[OrderID], [t].[CustomerID], [t].[EmployeeID], [t].[OrderDate], [c].[ContactName]
 FROM [Customers] AS [c]
 LEFT JOIN (
     SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
     FROM [Orders] AS [o]
     WHERE [o].[OrderID] > 5
     ORDER BY [o].[OrderDate]
 ) AS [t] ON [c].[CustomerID] = [t].[CustomerID]
```

maumar: i guess we can address order by in a different commit if thats significant work - it's not like we are enabling some crucial scenario with it, just nice to have

tuespetre: ORDER BY cannot be in a subquery without TOP.

maumar: ok, lets just do the 'where' - sorry for the confusion ;)

tuespetre:
:up::date:

maumar:
merged in 7441fc23e12f92b7c83a6acb8c2aa5db059746a1, thank you for the contribution!

PR #7851, opened 2017-03-11
Query: Improve ExpressionVisitorBase by making it recursive.

smitpatel:
DRY up VisitSubQuery method
Use TransformingQueryModelExpressionVisitor instead of RecursiveQueryModelExpressionVisitor

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/ExpressionVisitorBase.cs
+++ b/src/EFCore/Query/ExpressionVisitors/ExpressionVisitorBase.cs
@@ -3,8 +3,8 @@
 
 using System.Diagnostics;
 using System.Linq.Expressions;
-using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
+using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Parsing;
 
 namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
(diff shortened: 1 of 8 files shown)


smitpatel:
`MemberAccessBindingExpressionVisitor` still have VisitSubQuery and doesn't override from ExpressionVisitorBase due to the handling of VisitExtension for NullConditionalExpression.

tuespetre:
Can the `TransformingQueryModelExpressionVisitor` not be generic? As far as I can tell the generic only adds extra length to the lines that use it 😉

smitpatel:
@tuespetre - For current cases yes it does. I checked out how `ExpressionTransformingQueryModelVisitor` is used. When you derive from those classes, you actually need generic type to work with specific Visitor. For this EV we can avoid generic arguments now but in future if we need to derive from it, we need to add generics which becomes a breaking change.

@maumar , @anpete - Thoughts?

tuespetre:
@smitpatel, oh, ok, that makes perfect sense. Allows for more specialization when needed. 👍 Thanks!

smitpatel:
This one is also pending @maumar @anpete

src/EFCore/Query/ExpressionVisitors/ExpressionVisitorBase.cs
```diff
@@ -16,33 +16,34 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
...
         /// </returns>
-        public override Expression Visit([CanBeNull] Expression node)
-            => node == null
-               || node.NodeType == ExpressionType.Block
-                ? node
-                : base.Visit(node);
+        /// <param name="extensionExpression">The expression to visit.</param>
+        protected override Expression VisitExtension(Expression extensionExpression)
+        {
+            if (extensionExpression is NullConditionalExpression)
```
anpete: I think this needs to be removed because it means we never visit children on the NullConditionalExpression.

smitpatel: Removing this causes failures. I thought @maumar  has work item to handle NCE consistently at all places.

anpete: Fine if it is being handled

smitpatel: I will remove it from base and move to derived types.

smitpatel:
@anpete @maumar - Updated to remove the skipping of NullConditional from base visitor.

Issue #6924, opened 2016-11-02
Make more services singletons


AndriySvyryd:
There are currently some services that could be registered as singletons, reducing the number of allocations we have to make for each `DbContext` instance:
1. Services that no longer need to be scoped, such as `IInternalEntityEntryFactory` 
1. Services that have no mutable state, but use scoped services like `ICurrentDbContext` or `IDbContextOptions` in one of the methods and can be refactored to pass the required services in the method call itself, such as `IChangeTrackerFactory` (we should measure the performance/memory benefit of this in a case by case basis and understand if the breaking change and possibly more coupled or complex code is worth it)
1. Services that have a cache and could be refactored to use `DbContextServices.MemoryCache`, such as  `IValueGeneratorSelector` (already tracked by #5322)
1. Services that depend on one or more of the above, such as `IKeyPropagator`

Doing this would allow us to cache these services and not use DI to get them. On the other hand the DI resolution can be significantly sped up by not using delegates with `GetService` calls in them for registration, see #4133.

ajcvickers:
Some additional thoughts on this:

* Consider changing IDatabaseProviderServices so that it can be implemented as a singleton. This would mean that each provider would have a singleton instance, with the one being returned in the scope being the one fore the selected provider. It could then cache singleton provider services to avoid a GetService call. The service provider could no longer be injected and instead should be passed to each method that needs to resolve a scoped service.
* Consider not allowing a different logger per context instance. The dependency on ILogger, which is scoped because a different logger can be used for each context instance, is a big factor in forcing services to be scoped.
* Consider not allowing the model to change for each context instance, or passing the model to services that need it rather than depending on it through injection.

Issue #218, opened 2014-05-22
Logging: Design all up logging approach


rowanmiller:

rowanmiller:
We're not 100% happy with where we have landed - so we want to revisit this in 2.0. Next step is for @ajcvickers to prepare something for a design meeting.

ajcvickers:
## Logging Proposal for 2.0

### Scoping and sugar

Principle: logging is something that is setup and configured for the application, not per context instance.

To achieve this we introduce some new configuration APIs that are associated with the service provider. This is effectively application-level, or EF-global level configuration, although bound to the service provider so as not to be truly global.

We run the code to set configuration for this only once when we first use a new service provider.

This existing configuration would move:
- WarningsConfiguration
- MaxPoolSize
- IsSensitiveDataLoggingEnabled
- LoggerFactory
- MemoryCache
- ReplaceService

This means that ILogger no longer needs to be a scoped service, which solves the problem of logging forcing other services to be scoped.

Consequence/limitation: we cannot log for only one context instance and not another. This is something that we support with EF6, but does not work well with the logger abstractions we have. It would likely be ugly to implement and need something like:
- A LoggerProvider that keeps track of the context instances for which logging is enabled, removing instances when contexts are disposed.
- Passing the context instance to each Log call in a well-known way to allow for filtering

It's probably not worth doing all of this, especially given that most code just logs everything to one log anyway. For things that need correlation, we can do it in a way that is more natural to the logging abstractions, such as the work that was already done for connections.

### API

We could introduce a new "On..." method for this API, but this is probably not necessary. Instead we could hang it off of "OnConfiguring". The logging sugar API (Issue #1199) would also go here, since it is application/global configuration. Maybe something like this:

```C#
optionsBuilder
    .UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=Demo;")
    .GlobalConfiguration(b => b
        .EnableSensitiveDataLogging()
        .Log(message => Console.WriteLine(message)));
```
As stated above, this new builder would only be run once the first time a new service provider is used. This could be confusing in a similar way to OnModelCreating, but then creating a new "On..." method could also be confusing in the same way.

### Logger names and event IDs

From the Logging docs:
- "Name your loggers with a distinct prefix so they can easily be filtered or disabled."
- "A logging event may specify an integer event id. Event ids should uniquely identify a logging publication call site within a logging category (i.e. the name passed to ILoggerFactory.CreateLogger when creating the logger instance)."

[...]

PR #7853, opened 2017-03-11
Allow ILogger to be a singleton service

ajcvickers:
Issues #6924 #218

This change allows ILogger to be a singleton service and hence services that need to do logging no longer need to be scoped services. The end result turned out a bit different from discussed in #218, but follows essentially the same principles. The singleton configuration API wasn't turning out well, so I switched to extending the mechanism we already have for building new service providers as necessary. This means that, when EF is managing the internal service provider, all the options work as before, but if the app changes, for example, the warnings config, then this may require a new service provider to be built. (I added a warning if more than 20 service providers are built as a heuristic against applications doing crazy stuff where the service provider is being rebuilt all the time.)

There are a few more restrictions when the internal service provider is being controlled by the application:
* For options that are directly services (ILoggerFactory, IMemoryCache) the application must configure them directly in the internal service provider. This removes the need for any re-direction.
* For other options, they can only be configured once per service provider, as was discussed in the design meeting. At the moment this is only the logging options, but it could be used for other options if necessary going forward.

Also included here is some re-shaping of the OptionsExtensions classes to bake in their immutable nature and make it harder to accidentally mutate them. WarningsConfiguration was previously mutating, and this has been updated to not mutate. Also, WarningsConfiguration and CoreOptionsExtension are needed by providers, and hence these have now been moved to a public namespace.


src/EFCore/DbContextOptionsBuilder.cs
```diff
@@ -141,17 +194,30 @@ public virtual DbContextOptionsBuilder EnableSensitiveDataLogging()
         ///         and <see cref="EntityFrameworkQueryableExtensions.AsTracking{TEntity}(IQueryable{TEntity})" /> methods.
         ///     </para>
         ///     <para>
-        ///         The default value is <see cref="EntityFrameworkCore.QueryTrackingBehavior.TrackAll" />. This means the change tracker will
-        ///         keep track of changes for all entities that are returned from a LINQ query.
+        ///         The default value is <see cref="EntityFrameworkCore.QueryTrackingBehavior.TrackAll" />. This means 
+        ///         the change tracker willkeep track of changes for all entities that are returned from a LINQ query.
```
bricelam: *willkeep* is my new favorite word.

Issue #7189, opened 2016-12-05
Migrations: Make generated code play well with StyleCop


patriksvensson:
### Steps to reproduce
Add a new migration

### The issue
Let me start with saying "Thanks for an awesome project".

This isn't a huge problem, but for all of us that uses StyleCop, adding a new migration creates a lot of warnings (and errors for release builds where warnings-as-errors are enabled).

* Namespaces in `Migrations/MyAppDbContextModelSnapshot` are not ordered alphabetically.
* Namespaces in `Migrations/20161205224611_MyMigration.cs` are not ordered alphabetically.
* Namespaces in `Migrations/20161205224611_MyMigration.Designer.cs` are not ordered alphabetically.
* Access modifier for `Migrations/MyAppDbContextModelSnapshot` class is implicit.
* Access modifier for `Migrations/20161205224611_MyMigration.Designer.cs` class is implicit.

I would be happy to provide a PR to fix this if you think it's worth doing.
Another solution to this would be to add the standard `<auto-generated>` header to the generated files.

### Further technical details

EF Core version: (found in project.json or packages.config)
Operating system: Windows 10
Visual Studio version: VS 2015

bricelam:
The implicit modifiers in the `.Designer` partial classes is by design. It allows you to change the class's access in a single location--the migration file. We could probably mark those files as auto-generated since they'll rarely have to be modified after being generated.

rowanmiller:
@patriksvensson we'd love to take a PR on this one. Here is what we think should be done...

* Add `<auto-generated>` to the `.designer` files and to the model snapshot
* Do the alphabetical sorting of using statements

rowanmiller:
Putting on the backlog, but we'd take the PR into 1.2 if it is ready in time.

patriksvensson:
@rowanmiller I can submit a PR for this, but I'm not sure when 1.2 is due (the milestone gives me no information about this).

rowanmiller:
@patriksvensson if you can get it in sometime in the next month or so that would give us time to have it in one of the pre-releases for the next release.

patriksvensson:
@rowanmiller Sounds good to me. I have plenty of free time the next month or so, so sign me up for this one.

bricelam:
[...]

PR #7855, opened 2017-03-11
Make generated code play well with StyleCop

reaction1989:
All usings get sorted.

Added the <auto-generated> tag to Metadata and Snapshot file.

Question is wether this should also be added to the migration file.

Addresses: #7189


src/EFCore.Design/Migrations/Design/CSharpMigrationsGenerator.cs
```diff
@@ -99,6 +99,15 @@ public class CSharpMigrationsGenerator : MigrationsCodeGenerator
             return builder.ToString();
         }
 
+        private void AppendAutoGeneratedTag(IndentedStringBuilder builder)
```
ajcvickers: @bricelam Did you check these auto-generated comments are where we want them to be?

bricelam: Yes. But thinking about it more, they may not belong on the model snapshot since users are expected to manually merge conflicts...

bricelam: But we did decide (in the issue) to put them there at one point.

reaction1989: This does not prevent manual merging. Only stylecop will be quiet ☺

reaction1989:
Is there anything i should change on this one?

src/EFCore.Design/Migrations/Design/CSharpMigrationsGenerator.cs
```diff
@@ -99,6 +99,15 @@ public class CSharpMigrationsGenerator : MigrationsCodeGenerator
...
+        {
+            builder.AppendLine(@"//------------------------------------------------------------------------------");
+            builder.AppendLine(@"// <auto-generated>");
+            builder.AppendLine(@"//     This code was generated by a tool.");
+            builder.AppendLine(@"// </auto-generated>");
+            builder.AppendLine(@"//------------------------------------------------------------------------------");
```
bricelam: Hard-coded English strings are problematic for localization. Instead of this huge header, can we simplify it to just:
```c#
// <auto-generated />
```

src/EFCore.Design/Migrations/Design/CSharpMigrationsGenerator.cs
bricelam: I'm fine keeping it on the snapshot. I seem to remember calling this out when discussing the issue and we still landed on putting them there.

bricelam:
@reaction1989 sorry, I forgot I had this assigned until today 😄 Just the hard-coded English message needs to change before merging.

bricelam:
@reaction1989 sorry, I forgot I had this assigned until today 😄 Just the hard-coded English message needs to change before merging.

PR #7856, opened 2017-03-12
Query: Relational: Optimize GroupJoin/SelectMany into lateral joins

tuespetre:
🐄🌪

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/LateralJoinOptimizingQueryModelVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/LateralJoinOptimizingQueryModelVisitor.cs
@@ -1,0 +1,107 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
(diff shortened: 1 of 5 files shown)


smitpatel:
Assigning to @maumar . Changes look good. Probably the visitor can be moved inside query optimizer (either inline being in VisitAdditionalClause method or another visitors being used inside query optimizer).

Though we may need more rigorous testing to define the set of condition when this translation should be done. At present it is hard to make out structures which can benefit from this optimization.

src/EFCore/Query/ExpressionVisitors/Internal/LateralJoinOptimizingQueryModelVisitor.cs
```diff
@@ -0,0 +1,107 @@
...
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public override void VisitAdditionalFromClause(
+            [NotNull] AdditionalFromClause additionalFromClause,
+            [NotNull] QueryModel queryModel,
+            int index)
+        {
+            if (index > 0
+                && additionalFromClause.FromExpression is SubQueryExpression subQueryExpression
```
maumar: DRY it up, perhaps extension method on AdditionalFromClause? (similar set of conditions can be seen in AdditionalFromClauseOptimizingQueryModelVisitor->VisitAdditionalFromClause

smitpatel:
@maumar - Thoughts? This could be potentially done inside query optimizer rather than separate visitor. If you are not planning to rebase/fix/merge this PR then please close this.

PR #7857, opened 2017-03-12
Query: Relational: implement LEFT JOIN LATERAL / OUTER APPLY

tuespetre:
These changes simply pull up the pushed down subquery expression created by the `RelationalResultOperatorHandler` for the `DefaultIfEmptyResultOperator` and add it as a `LeftJoinLateralExpression` instead of a `CrossJoinLateralExpression`.

Another possible optimization that could be handled later is to re-order subquery result operators to 'float' `DefaultIfEmptyResultOperator`s to the end of the list, e.g. for cases like `.DefaultIfEmpty().Distinct()` or `.DefaultIfEmpty().Take(x).Skip(y)`. That would allow more queries to become `OUTER APPLY` instead of `CROSS APPLY`.

### DIFF
--- a/src/EFCore.Relational/Query/Expressions/LeftJoinLateralExpression.cs
+++ b/src/EFCore.Relational/Query/Expressions/LeftJoinLateralExpression.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Query.Sql;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Query.Expressions
+{
+    /// <summary>
+    ///     Represents a SQL LEFT JOIN LATERAL expression.
+    /// </summary>
+    public class LeftJoinLateralExpression : JoinExpressionBase
+    {
+        /// <summary>
+        ///     Creates a new instance of LeftJoinLateralExpression.
+        /// </summary>
+        /// <param name="tableExpression"> The target table expression. </param>
+        public LeftJoinLateralExpression([NotNull] TableExpressionBase tableExpression)
+            : base(Check.NotNull(tableExpression, nameof(tableExpression)))
+        {
+        }
+
+        /// <summary>
+        ///     Dispatches to the specific visit method for this node type.
+        /// </summary>
+        protected override Expression Accept(ExpressionVisitor visitor)
+        {
+            Check.NotNull(visitor, nameof(visitor));
+
+            return visitor is ISqlExpressionVisitor sqlExpressionVisitor
+                ? sqlExpressionVisitor.VisitLeftJoinLateral(this)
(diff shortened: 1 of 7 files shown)


src/EFCore.Relational/Query/Expressions/LeftJoinLateralExpression.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        /// <summary>
+        ///     Dispatches to the specific visit method for this node type.
+        /// </summary>
+        protected override Expression Accept(ExpressionVisitor visitor)
+        {
+            Check.NotNull(visitor, nameof(visitor));
+
+            var specificVisitor = visitor as ISqlExpressionVisitor;
```
tuespetre: Ugh, C#7 came out since I first wrote this

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1230,22 +1230,16 @@ private Shaper ExtractShaper(MethodCallExpression shapedQueryExpression, int off
...
-
-            if (outerSelectExpression == null)
-            {
-                return false;
-            }
-
             var innerSelectExpression = TryGetQuery(fromClause);
 
-            if (innerSelectExpression?.Tables.Count != 1)
+            if (outerSelectExpression == null || innerSelectExpression == null)
```
smitpatel: Any specific reason to remove check on tables count?

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1272,16 +1266,35 @@ var innerShapedQuery
...
-                        innerSelectExpression.Tables.First(),
+            if (correlated)
+            {
+                var finalResultOperator
+                    = ((SubQueryExpression)fromClause.FromExpression)
+                        .QueryModel.ResultOperators.LastOrDefault();
+
+                if (finalResultOperator is DefaultIfEmptyResultOperator)
+                {
+                    innerSelectExpression = (SelectExpression)innerSelectExpression.Tables[0];
```
smitpatel: value being used in all the paths, store in variable before entering the if blocks.

smitpatel: also use different variable name here. Causes bit of confusion with the table count check above, that how did we end up with 2nd table in innerSelectExpression.

smitpatel:
@anpete @maumar - I am looking to merge this PR. Please review. Especially from translation perspective.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1272,16 +1266,35 @@ var innerShapedQuery
...
+
+                    outerSelectExpression.AddLeftJoinLateral(
+                        ((LeftOuterJoinExpression)innerSelectExpression.Tables[1]).TableExpression,
+                        innerSelectExpression.Projection);
+                }
+                else
+                {
+                    outerSelectExpression.AddCrossJoinLateral(
```
maumar: add test for this - similar to SelectMany_Joined_DefaultIfEmpty but with .Distinct() or other result op after DIE

Issue #7864, opened 2017-03-13
OrderBy_correlated_subquery_lol2 seems to be a very slow test


ajcvickers:
The `OrderBy_correlated_subquery_lol2` test seems to run slowly on SQLite, In-memory, and SQL Server. We should check if there is a way to make it faster while still testing what it needs to test.

* QuerySqliteTest.OrderBy_correlated_subquery_lol2: 145 seconds
* QueryInMemoryTest.OrderBy_correlated_subquery_lol2: 86 seconds
* RowNumberPagingTest.OrderBy_correlated_subquery_lol2: 82 seconds
* QuerySqlServerTest.OrderBy_correlated_subquery_lol2: 81 seconds

Lol.

ajcvickers:
@maumar Can you make this your second priority, after the 1.1.0 to 1.1.1 regressions? We can discuss again if there are no easy wins here.

ajcvickers:
Also related to #7837.

maumar:
Fixed in c402d73796551929cdff1a50c77c326e62c309f3

smitpatel:
@divega - This does not seem to be enhancement. It was just test which used to pull a lot of data. We trimmed down the size of result to speed up. No product code changes.

PR #7867, opened 2017-03-13
[TINY] Fix to #7864 - OrderBy_correlated_subquery_lol2 seems to be a very slow test

maumar:
Problem was in the in-memory portion of the test. We would execute 830 subqueries and that took the vast majority of the time. Fix is to introduce additional predicate, cutting the number of subqueries down to 4.

### DIFF
--- a/src/EFCore.Specification.Tests/QueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/QueryTestBase.cs
@@ -3386,24 +3386,26 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [ConditionalFact]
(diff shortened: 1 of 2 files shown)


src/EFCore.Specification.Tests/QueryTestBase.cs
```diff
@@ -3386,7 +3386,7 @@ public virtual void OrderBy_ThenBy_predicate()
         }
 
         [ConditionalFact]
-        public virtual void OrderBy_correlated_subquery_lol()
+        public virtual void OrderBy_correlated_subquery1()
         {
             AssertQuery<Customer>(
                 cs => from c in cs
```
smitpatel: Possibly add some filter here too.

PR #7877, opened 2017-03-14
Fix SQL Server test timeouts

### LINKED ISSUE #7411: Investigate parallelization issues with SqlServer.FunctionalTests
These tests fail a lot on CI (and for me after I drop databases) when run in parallel. We used to mitigate this on CI using an environment variable, but this feature is being removed from KoreBuild (see aspnet/KoreBuild#157). Instead, we'll need to disable them at the assembly level by adding the following. (Or debug and fix the issues.)

```C#
[assembly: CollectionBehavior(DisableTestParallelization = true)]
```

AndriySvyryd:
* Remove the transient failure retry count limit in SqlServerDatabaseCreator.Exists so it's only limited by time (1 minute by default)
* Increase timeout used for creating SqlServer test databases to account for the increased delay when running in parallel.
* Speed up SqlServerDatabaseCreatorTest by decreasing the timeouts used
* Make SqlServer.FunctionalTests run in parallel
* Turn off SqlClient connection resiliency in tests

Fixes #7411

### DIFF
--- a/src/EFCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
+++ b/src/EFCore.SqlServer/Storage/Internal/SqlServerDatabaseCreator.cs
@@ -35,6 +35,18 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        public virtual TimeSpan RetryDelay { get; set; } = TimeSpan.FromMilliseconds(500);
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


smitpatel:
This passed on efci

ajcvickers:
@smitpatel Did efci already have databases created or was it starting from clean?

smitpatel:
all databases dropped before build.cmd starts

AndriySvyryd:
The changed that ran on efci dropped the databases.
And this still failed on aspnetci with 'no process on the other end' exception, so I'm going to try some more things

AndriySvyryd:
Fiddled with timeouts more. Now both CIs run the tests successfully.
@divega's idea to shrink the db model doesn't work consistently so we can consider just doing it manually for now

PR #7878, opened 2017-03-14
Test: Merge upstream ConditionalFact changes

bricelam:
[appveyor test]


src/EFCore.Specification.Tests/TestUtilities/Xunit/FrameworkSkipConditionAttribute.cs
```diff
@@ -27,20 +26,17 @@ private static bool CanRunOnThisFramework(RuntimeFrameworks excludedFrameworks)
...
+#if NET452
+            if (excludedFrameworks.HasFlag(RuntimeFrameworks.CLR))
             {
                 return false;
             }
-
-            if (excludedFrameworks.HasFlag(RuntimeFrameworks.CoreCLR)
-                &&
-                RuntimeInformation.FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase))
+#else
```
smitpatel: How about conditional If? combines both return false.

bricelam: I like it! But... I'll keep it the same as the upstream for now.

bricelam:
Upstream code is in [aspnet/Testing](https://github.com/aspnet/Testing) although we've deviated a bit.

smitpatel:
nit: our ConditionalAttribute can also use nameof pattern like upstream.

PR #7880, opened 2017-03-14
AppVeyor: Run only SqlServer FunctionalTests

smitpatel:
This PR enables only SqlServer Functional Tests on AppVeyor. It takes about ~23 minutes for 1 run to finish.
Since we already had made groups of testsin #7842 this PR just run the group which travis cannot run.

Also had to change to SqlServer 2016 because for some reason SqlServer 2014 is not starting up in AppVeyor.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,16 +1,14 @@
-only_commits:
-  # Start a build if commit message contains 'appveyor'
-  message: /appveyor/
+version: 1.2.0-preview1-{build}
 init:
   - git config --global core.autocrlf true
 build_script:
-  - ps: .\build.ps1
+  - ps: .\build.ps1 /p:TestGroup=SqlServer
 clone_depth: 1
 test: off
 deploy: off
 # See http://www.appveyor.com/docs/services-databases
 services:
-  - mssql2014
+  - mssql2016
 branches:
   only:
     - master
@@ -21,7 +19,7 @@ environment:
   global:
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     DOTNET_CLI_TELEMETRY_OPTOUT: 1
-    Test__SqlServer__DefaultConnection: Server=(local)\SQL2014;Database=master;User ID=sa;Password=Password12!
+    Test__SqlServer__DefaultConnection: Server=(local)\SQL2016;Database=master;User ID=sa;Password=Password12!
 matrix:
   fast_finish: true
 artifacts:
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -1,0 +1,15 @@
+<Project>
+  <ItemGroup>
+    <!-- SQL Server functional tests which needs SQL connection but no exsting database. -->
+    <SqlServerTests Include="$(RepositoryRoot)test\EFCore.SqlServer*FunctionalTests\*.csproj" />
+  </ItemGroup>
+
+  <Target Name="_FilterTestProjects" BeforeTargets="TestProjects">
+    <ItemGroup Condition="'$(TestGroup)' != ''">
+      <ProjectsToTest Remove="@(ProjectsToTest)" Condition="'$(TestGroup)' != 'All'" />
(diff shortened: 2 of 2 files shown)


smitpatel:
Facing timeout errors. putting in WIP.

appveyor.yml
```diff
@@ -1,16 +1,13 @@
-only_commits:
-  # Start a build if commit message contains 'appveyor'
-  message: /appveyor/
```
Eilon: Yay!

appveyor.yml
```diff
@@ -1,16 +1,13 @@
...
 build_script:
-  - ps: .\build.ps1
+  - ps: .\build.ps1 /p:TestGroup=SqlServer
 clone_depth: 1
 test: off
 deploy: off
 # See http://www.appveyor.com/docs/services-databases
 services:
-  - mssql2014
+  - mssql2016
```
divega: Does anyone know what exact version is this? Is it regular SQL Server or LocalDB, is it RTM or RC1? Curious because that can affect whether [instant database file initialization](https://msdn.microsoft.com/en-us/library/ms175935.aspx) can and will be enabled.

smitpatel: `SQL Server 2016 Express`

smitpatel:
This is ready for review.
Changes introduced
- Run AppVeyor on all PRs but only run SqlServer FunctionTests.
- I have preserved grouping Nate did. Though our basic requirement is SqlServer & non-SqlServer tests only.
- This does not run benchmark tests. It involves importing adventureworks in AppVeyor before calling build. I believe if all other tests are passing then benchmark would not fail.

I monitored different issues happening when AppVeyor was failing. The biggest issue was error with sql2014 service. For some reasons (associated with AppVeyor), it fails to start instance of sql server 2014. That explains the consistent failures we were having for past few months. I changed it to use Sql Server 2016 now which is working fine.

There were also timeout issues happening. I also noticed those happening on efci when I removed the environment variable for No parallel tests. Ideally the assembly condition should be sufficient enough but it doesn't seem to be the case. (I am still monitoring EFCI to check it).
Changes in #7877 fixes the timeout issues. This has been consistently passing in last 3 runs. It takes about 20 mins to finish so we don't hog up queues either.

smitpatel:
We do have `sqlcmd` available on AppVeyor. I am checking about SQLCLient ODBC drivers installation,

ajcvickers:
So, with this change, the plan is to have AppVeyor run the SQL Server tests, and have Travis for everything else?

smitpatel:
@ajcvickers - Yes. For PRs on query this can be very useful because sqlserver functional tests are more likely to fail.

build/repo.targets
```diff
@@ -0,0 +1,40 @@
...
+    <_SqlServerIsAvailable Condition="'$(OS)' != 'Windows_NT' AND '$(Test__SqlServer__DefaultConnection)' == ''">false</_SqlServerIsAvailable>
+  </PropertyGroup>
+
+  <Target Name="_FilterTestProjects" BeforeTargets="TestProjects">
+    <ItemGroup Condition="'$(TestGroup)' != ''">
+      <ProjectsToTest Remove="@(ProjectsToTest)" Condition="'$(TestGroup)' != 'All'" />
+      <ProjectsToTest Include="@(UnitTests)" Condition="'$(TestGroup)' == 'Unit'" />
+      <ProjectsToTest Include="@(FunctionalTests)" Condition="'$(TestGroup)' == 'Functional'" />
+      <ProjectsToTest Include="@(BenchmarkTests)" Condition="'$(TestGroup)' == 'Benchmarks'" />
+      <ProjectsToTest Include="@(SqlServerTests)" Condition="'$(TestGroup)' == 'SqlServer'" />
```
bricelam: Only `SqlServer` is being used. Remove `All`, `Unit`, `Functional` & `Benchmark` groups until we need them.

bricelam: (Their items too)

smitpatel: Updated.

PR #7893, opened 2017-03-15
Add scripts to clean and speedup the default LocalDB instance

AndriySvyryd:
All research and testing done by @divega

`CleanMSSQLLocalDB.cmd` drops all databases and deletes the default instance. This is useful to quickly get to a clean state, it doesn't have any perf benefits.

`ShrinkLocalDBModel.cmd` runs `DBCC ShrinkDatabase (model)` several times as this was empirically proven to yield the smallest `model` file resulting in faster database creation.
Then it changes the growth amount for the `model` database to 10%, this is the default for non-system dbs.

To use this you need to have sqlcmd and SQLClient ODBC drivers installed:
* https://www.microsoft.com/en-us/download/details.aspx?id=53591
* https://www.microsoft.com/en-us/download/details.aspx?id=53339

On my machine it sped up the SQLServer FTs on .NET Core from 4 min to 2.5 min

### DIFF
--- a/tools/ShrinkLocalDBModel.cmd
+++ b/tools/ShrinkLocalDBModel.cmd
@@ -1,0 +1,14 @@
+@echo off
+echo Shrinking the model database several times
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+echo Configuring model database to grow with default settings
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "ALTER DATABASE model MODIFY FILE (NAME=modeldev,FILEGROWTH=10%%);" 
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "ALTER DATABASE model MODIFY FILE (NAME=modellog,FILEGROWTH=10%%);"
--- a/tools/CleanMSSQLLocalDB.cmd
+++ b/tools/CleanMSSQLLocalDB.cmd
@@ -1,0 +1,8 @@
+@echo off
+sqlcmd -S "(localdb)\mssqllocaldb" -i "DropAllDatabases.sql" -o "DropAll.sql"
+sqlcmd -S "(localdb)\mssqllocaldb" -i "DropAll.sql"
+del "DropAll.sql"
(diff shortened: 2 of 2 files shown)


smitpatel:
I was wondering if we should ~blacklist~ ignore adventureworks while dropping all databases? It is never changed and have to manually restored if we run the script.

tools/ShrinkLocalDBModel.cmd
```diff
@@ -0,0 +1,14 @@
+@echo off
+echo Shrinking the model database several times
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
```
divega: Maybe it is YAGNI, but what this command does could apply to any SQL Server instance. Maybe have an optional instance name parameter? (please don't ask me how :smile:).

AndriySvyryd:
@smitpatel the script also deletes the instance, so no databases will remain. It just drops them first to ensure the physical files are deleted
@divega sure
@ErikEJ sqlcmd needs it to execute the query
@ajcvickers It's unlikely that the CIs have the dependencies installed. I had to install the ODBC client on my box

tools/ShrinkLocalDBModel.cmd
```diff
@@ -0,0 +1,14 @@
...
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
+sqlcmd -S "(localdb)\mssqllocaldb" -Q "DBCC SHRINKDATABASE (model) WITH NO_INFOMSGS"
```
AndriySvyryd: Quotes are needed for PowerShell

ErikEJ:
I wish @martincostello could have a look at this! https://github.com/martincostello/sqllocaldb

tools/ShrinkLocalDBModel.cmd
divega: ok :+1:

tools/ShrinkLocalDBModel.cmd
divega: add them in the last two lines then?

AndriySvyryd: Yes

martincostello:
Hello @ErikEJ - just catching up with this PR. Was there something specific LocalDB-related you wanted some input from me on?

ErikEJ:
@martincostello Just wondered if there was anything you could use or share re removing a localdb instance with files in the best manner?

martincostello:
@ErikEJ Nothing particularly groundbreaking. The main thing that we observed where we used LocalDB for testing extensively at a previous job, was that deleting the throw-away test instances that were created as part of the tests still left files on disk. Over time and many builds and tests, this filled up the disks on build and developer machines as the LocalDB engine leaves behind ~1MB of logs and trace files.

image

To account for that I added a method to the internals of my LocalDB Native API wrapper to delete the entire instance directory from disk: [```void DeleteInstanceFiles(string instanceName)```](https://github.com/martincostello/sqllocaldb/blob/d0ef9bae0f854155f121ca95090731aeff1e3f43/src/SqlLocalDb/SqlLocalDbApi.cs#L973-L1033).

Effectively, I just went for the nuclear option. The flexibility to do that came from not using the default instance, meaning we didn't have to worry about potential corruption, at the expense of the overhead of creating a new instance with a GUID-based name at the start of a test-run. We tried to minimise this by having an instance per-test logical test run, and creating/dropping whatever individual databases were created during the tests within the single throwaway instance.

tools/ShrinkLocalDBModel.cmd
```diff
@@ -0,0 +1,14 @@
...
+echo Configuring model to grow with default settings
```
divega: nit: Configuring model **database** to grow with default settings

AndriySvyryd:
@martincostello Thanks for the tip, we'll probably just stick to the default instance to avoid potential bloat.

PR #7895, opened 2017-03-15
[Experiment] Run shrink utility on appveyor

smitpatel:
For testing only


smitpatel:
@divega - This did not get much time gain.

divega:
@smitpatel Yes, I noticed. I have been looking around (mostly in some of the CI agents) and it appears that in some instances `model.mdf` and `model_log.ldf` are created and remain small (~2.4 mb and 0.5 mb) without having to do anything. Can't explain why different machines behave so differently :frowning_face:, but it seems in AppVeyor we don't need to do this.

smitpatel:
@divega - Should I recall this?

divega:
@smitpatel yes, I think so. We can always add it if the need arises.

PR #7912, opened 2017-03-16
Query: Streamline OptimizeQueryModel

tuespetre:
Some query optimizations can only be done before navigations are
rewritten and some can only be done after. These changes roll up
the calls to a couple of specific optimizing expression visitors
into the QueryOptimizer and call the QueryOptimizer before and
after rewriting navigations.

- I'm submitting this PR to extend the discussion started here: https://github.com/aspnet/EntityFramework/pull/7845#discussion_r106445944
- You can see in the changed test output that this already has some benefit.
- This will allow for new and additional optimizations to be invoked in `QueryOptimizer` with the assurance that the timing of the navigation rewriting will not prevent any opportunities for optimization. 
- Note that optimizations can apply both before AND after navigation rewriting.


tuespetre:
cc @smitpatel @maumar @anpete

tuespetre:
@smitpatel @maumar @anpete 

Can we please review this one before I look for any more optimizations to apply? Reason being that some optimizations are just going to have to be run twice to be effective and I feel this PR presents a much cleaner, uniform approach to managing when, where, and how those optimizations are run. 

Example:

```
from c in cs
join o in os on c.CustomerID equals o.CustomerID into lo
from o in { from x in lo where x.OrderID > 5 }
select new { c.ContactName, o.OrderID }
```

The `QueryOptimizer`'s flattening properties will cause this query model to be flattened like so:

```
from c in cs
join o in os on c.CustomerID equals o.CustomerID into lo
from x in lo
where x.OrderID > 5
select new { c.ContactName, o.OrderID }
```

So in order to optimize it the way we **really** want to (move the where clause into the `GroupJoinClause` as a subquery):

```
from c in cs
join o in { from o in os where o.OrderID > 5 } on c.CustomerID equals o.CustomerID into lo
from o in lo
select new { c.ContactName, o.OrderID }
```

This optimization needs to be run *before* any subquery flattening.

At the same time, it also needs to run *after* navigation rewriting:

```
from c in cs
from o in c.Orders.Where(x => x.OrderID > 5)
select new { c.ContactName, o.OrderID }

becomes

join o in os on c.CustomerID equals o.CustomerID into lo
from o in { from x in lo where x.OrderID > 5 }
select new { c.ContactName, o.OrderID }
```

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -414,23 +414,6 @@ public virtual void VisitSubQueryModel([NotNull] QueryModel queryModel)
 
             base.VisitAdditionalFromClause(fromClause, queryModel, index);
 
-            if (fromClause.FromExpression is QuerySourceReferenceExpression)
```
tuespetre: @smitpatel @maumar @anpete this block was here only because we have not been running the query optimizer once more after rewriting navigations.

anpete:
@tuespetre Approach seems good to me.

tuespetre:
re:b:ased

anpete:
@tuespetre Seeing some failures in SQL assertions, can you take a look?

tuespetre:
It looks like ordering differences in .NET Core, like how some tests used Contains instead of Equals before. Not sure what you want to do there

anpete:
Most of the ones I am seeing look related to recent Include changes. I.e. not order related but just incorrect aliases etc.

tuespetre:
I will look into it as soonas I can but it may be a while -- we are gaining a new family member this morning :baby:

tuespetre:
@anpete re:b:ased, and it looks like whatever was causing the SQL difference between .NET and .NET Core has been resolved

anpete:
Thanks!

PR #7929, opened 2017-03-17
Query: Remove OrderingExpressionVisitorFactory

smitpatel:
**DEAD CODE**


src/EFCore/Query/EntityQueryModelVisitorDependencies.cs
```diff
@@ -68,7 +68,6 @@ public sealed class EntityQueryModelVisitorDependencies
         ///     The <see cref="IMemberAccessBindingExpressionVisitorFactory" /> to be used when
         ///     processing the query.
         /// </param>
-        /// <param name="orderingExpressionVisitorFactory"> The <see cref="IOrderingExpressionVisitorFactory" /> to be used when processing the query. </param>
```
lajones: Isn't this a breaking change?

smitpatel: Yes it is. This is going to dev branch though.

lajones: Didn't we agree to add some label to PR's with breaking changes - so that we could highlight them all come release time? I can't remember what we agreed though.

smitpatel: I added labels.

lajones: Thx. LGTM.

PR #7976, opened 2017-03-23
Remove net451 as a cross-compile target

pranavkm:


src/EFCore.Design/Design/Internal/ForwardingProxy.cs
```diff
@@ -18,10 +18,12 @@ public static class ForwardingProxy
...
             where T : class
         {
-#if NET451
+#if NET46
             return target as T ?? new ForwardingProxy<T>(target).GetTransparentProxy();
-#else
+#elif NETSTANDARD1_3
             return (T)target;
+#else
+#error target frameworks need to be updated.
```
bricelam: Not a fan, but I can see the value...

src/EFCore.Specification.Tests/TestModels/ConcurrencyModel/Team.cs
```diff
@@ -10,10 +10,12 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.Concurren
 {
     public class Team
     {
-#if NET452
+#if NET46
         private readonly ObservableCollection<Driver> _drivers = new ObservableCollectionListSource<Driver>();
```
pranavkm: Btw, I wasn't sure if you still need the observable types. Are they replaceable by the ones in `NETStandard1.3`?

bricelam: @ajcvickers?

pranavkm: I'll merge the PR in to unblock the rest of my work. You can remove this in a follow up if it is indeed removable.

Issue #7714, opened 2017-02-24
Invalid query produced when projecting DTO containing subquery that returns a single element 


dougbu:
### Describe what is not working as expected.
Using a simplified version of the #6534 code, get an `InvalidOperationException` instead of expected compilation failure (pre-1.1.1) or result (with 1.1.1). Main simplification is this one does not involve an optional (`1 => 0,1`) relationship.

Using the out-of-the-box application database created in a .NET Core web application with Individual User Accounts, add the following to `HomeController`:
``` c#
        public async Task<List<Results>> Bug1()
        {
            var val = from c in _dbContext.Users
                      where c.EmailConfirmed
                      orderby c.Email
                      select new Results
                      {
                          Count = c.Claims.Count(x => x.Id <= 3),
                          Name = c.UserName,
                      };

            return await val.ToListAsync();
        }

        public class Results
        {
            public int Count { get; set; }

            public string Name { get; set; }
        }
```
Then, restore, build and run the application. Navigate to `/home/bug` and see the stack on the console.

### If you are seeing an exception, include the full exceptions details (message and stack trace).
```
fail: Microsoft.EntityFrameworkCore.Query.Internal.SqlServerQueryCompilationContextFactory[1]
      An exception occurred in the database while iterating the results of a query.
      System.InvalidOperationException: No value provided for required parameter '_outer_Id'.
         at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.CreateCommand(IRelationalConnection connection, IReadOnlyDictionary`2 parameterValues)
         at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.<ExecuteAsync>d__26.MoveNext()
      --- End of stack trace from previous location where exception was thrown ---
         at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
         at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
         at Microsoft.EntityFrameworkCore.Query.Internal.AsyncQueryingEnumerable.AsyncEnumerator.<BufferlessMoveNext>d__9.MoveNext()
      --- End of stack trace from previous location where exception was thrown ---
         at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
         at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

Issue #7033, opened 2016-11-17
v1.1 - The multi-part identifier "x.Id" could not be bound


ctolkien:
The following linq:

```csharp
var foo = await _context.Things
                .Select(x => new ThingViewModel
                {
                    Name = x.Name,
                    Bars = x.Bars
                        .Take(5)
                        .Select(q => new ThingViewModel.BarViewModel {
                            OtherName = q.Name,
                    })
                }).ToListAsync();
```

Generates a Select N+1 of the following:

```sql
SELECT [t].[Name]
FROM (
    SELECT TOP(5) [q0].*
    FROM [Bars] AS [q0]
    WHERE [x].[Id] = [q0].[ThingId]
) AS [t]
```

This is invalid SQL, as `[x]` doesn't exist here. Removing the `.Take(5)` allows the query to proceed without issue (it is a select n+1, but I presume this is by design for now).

ctolkien:
I should also flag, this is a regression from v1.0.

smitpatel:
We do client eval in this case producing N+1 queries.
In the second query, ParameterInjection is faulty hence we generated incorrect SQL instead of having injected parameter.

smitpatel:
Possible workarounds
1. Makes 2 queries but fetches all records from second table though only relevant columns. So if there are many `Bars` for each `Thing` it may perform poorly.
```c#
var foo1 = db.Things.Select(t => new {Id = t.Id, Name = t.Name}).ToList();
var bars =
    db.Set<Bar>()
        .Select(b => new {Id = b.ThingId, OtherName = b.Name})
        .GroupBy(t => t.Id)
        .Select(g => new {Id = g.Key, Bars = g.Take(5)})
        .ToList();

var foo = foo1.Join(bars, f1 => f1.Id, f2 => f2.Id,
    (f1, f2) => new { Name = f1.Name, Bars = f2.Bars }).ToList();
```
2. N+1 queries but only fetches required rows from `Bars` table. (Also fetches all columns)
```c#
var temp = db.Things.Select(t => new { t.Name, Bars = t.Bars.Take(5) }).ToList();
var foo = temp.Select(t => new {Name = t.Name, Bars = t.Bars.Select(b => new {OtherName = b.Name})}).ToList();
```

divega:
@smitpatel could you please take a stab at writing a justification/risk assessment for taking the fix in 1.1.1? A comment here in the bug works for me. Thanks.

smitpatel:
[...]

PR #7992, opened 2017-03-25
[1.1.2] Improved fix to #7714 - Invalid query produced when projecting DTO containing subquery that returns a single element

maumar:
Problem was that in the fix for #7033 we force client evaluation on queries that try to bind to parents, when the parent has a client projection.
However, for the case when single element is returned (e.g. cusotmers.Select(c => new DTO { Count = c.Orders.Count() })) we don't need to do client evaluation.
We actually try to merge the subquery into the parent query, but this causes a problem because earlier in the pipeline we assumed that client eval is needed (since parent QM has client projection)

Fix is to flow information about query model's StreamedDataInfo into SqlTranslatingExpressionVisitor so that we only client-eval when it's really necessary (i.e. when more than one element is returned from the subquery).
Information is stored in (proper) internal fields as we don't really want to expose this API in any way.


anpete:
@maumar Is this only for 1.1.2?

maumar:
@anpete yes, this doesn't repro in current dev bits due to number of refactorings we have done

anpete:
Worth adding new tests to dev?

maumar:
yes, I will port them once this is merged

Issue #7939, opened 2017-03-21
What's the intended use pattern of RelationalDiagnosticSource* messages?


NickCraver:
I'm trying to hook [MiniProfiler](https://github.com/MiniProfiler/dotnet) up to Entity Framework Core, but I'm not really understanding what the intended interface is. My understanding is that my best (supported) route is [DiagnosticSource](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/DiagnosticSourceUsersGuide.md), called via [`RelationalDiagnostics`](https://github.com/aspnet/EntityFramework/blob/f386095005e46ea3aa4d677e4439cdac113dbfb1/src/EFCore.Relational/Internal/RelationalDiagnostics.cs). That's the only path I see to get my hands on the `DbCommand`, `DbConnection`, etc. needed for various scenarios.

But [the `.Write()` call for DiagnosticSource](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSource.cs#L35) is:
```c#
public abstract void Write(string name, object value);
```
The way RelationalDiagnostics uses this passes in a typed object (good), [example here](https://github.com/aspnet/EntityFramework/blob/f386095005e46ea3aa4d677e4439cdac113dbfb1/src/EFCore.Relational/Internal/RelationalDiagnostics.cs#L45-L53):
```c#
diagnosticSource.Write(
    BeforeExecuteCommand,
    new RelationalDiagnosticSourceBeforeMessage
    {
        ConnectionId = connectionId,
        Command = command,
        ExecuteMethod = executeMethod,
        InstanceId = instanceId,
        Timestamp = startTimestamp,
        IsAsync = async
    });
```
...but those types passed are internal, for example here's [RelationalDiagnosticSourceBeforeMessage](https://github.com/aspnet/EntityFramework/blob/f386095005e46ea3aa4d677e4439cdac113dbfb1/src/EFCore.Relational/Storage/RelationalConnection.cs): 
```c#
namespace Microsoft.EntityFrameworkCore.Internal
{
    // TODO revert to anonymous types when https://github.com/dotnet/corefx/issues/4672 is fixed
    internal class RelationalDiagnosticSourceBeforeMessage
    {
        public Guid ConnectionId { get; set; }
        public DbCommand Command { get; set; }
        public string ExecuteMethod { get; set; }
        public bool IsAsync { get; set; }
[...]

PR #8000, opened 2017-03-27
Fix diagnostic name for RelationalDiagnostics.ConnectionError

NickCraver:
Looks like a copy/paste oops here - the diagnostic name was incorrect.

I was trying to write an adapter per #7939 and wondered why this wasn't working. Turns out I wasn't the crazy one, *this time*...

Fixes item 1 in #8001.


anpete:
Thanks!

PR #8022, opened 2017-03-29
Test: Split SqlServerDatabaseCreator tests

smitpatel:
Verifying on AppVeyor first.

In xunit each class is collection of tests. Collections are run in parallel but tests in particular collection are run sequentially.
On my dev box, SqlServer.FunctionalTests take 160s time on average. If I remove `SqlServerDatabaseCreatorTest` then it takes 70s. So CreatorTest is bottleneck here. This PR splits them out in different classes so that they can be run in parallel.
After this change, average time for SqlServer.FunctionalTests on my dev box is ~90s.


smitpatel:
AppVeyor passed.

AndriySvyryd:
You can keep them in one file to preserve the logical grouping.
Maybe they can even be just nested classes, this way they can have better names.

smitpatel:
Closing this in favour of #8028

PR #8030, opened 2017-03-30
Converted test projects to run on netcoreapp2.0

kichalla:


ajcvickers:
Let's make sure not to merge this until it works, both on the command line and in VS. All indications are that it currently does not work.

ajcvickers:
@bricelam and @smitpatel can work on this, but it would be helpful to have some more specific information so they are not starting from scratch. @kichalla Can you provide some more specific leads as to what is broken? Are there common things that don't work when doing this conversion? Are there issues with cross-compiles? Etc...

smitpatel:
Looking at logs
- System.Runtime assembly mismatch. We have referenced a fixed version of system.runtime in source code while building code in memory (though only for net46 framework, netcore calls mscorlib) that may need to updated properly.
- ❗️ `Attribute` is not defined in netstandard 2.0.
- dotnet-ef.Tests needs to be updated. Assembly load failure.

bricelam:
@kichalla Is this a known issue?

> Testhost process exited with error:
Unhandled Exception: System.TypeInitializationException: The type initializer for 'Microsoft.VisualStudio.TestPlatform.ObjectModel.EqtTrace' threw an exception. ---> System.TypeLoadException: Could not load type 'System.Threading.Monitor' from assembly 'System.Threading, Version=4.0.12.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.

kichalla:
@bricelam I haven't hit this issue in any other repo which was converted to netcoreapp2.0.

bricelam:
Getting closer... Need to tweak the tests that use Roslyn to unify netstandard 2.0 libraries.

bricelam:
@smitpatel I think this is good to go. Want to take one last review for approval?

build/dependencies.props
```diff
@@ -11,6 +11,7 @@
     <NetStandardImplicitPackageVersion>1.6.1</NetStandardImplicitPackageVersion>
     <RelinqVersion>2.1.1</RelinqVersion>
     <RoslynVersion>1.3.0</RoslynVersion>
+    <RuntimeFrameworkVersion>2.0.0-*</RuntimeFrameworkVersion>
```
smitpatel: Just a side question: Where is this used?

bricelam: SDK uses it when implicitly referencing `Microsoft.NETCore.App`

smitpatel:
Lets wait till at least one of the appveyor passes before checking in.

build/dependencies.props
bricelam: I believe this is saying: Don't use the version that ships with the SDK KoreBuild downloaded, use the latest.

kichalla: Right, the one is SDK is usually older than the package and doesn't have the fixes quick enough.

pmiddleton:
I think this PR broke the build in Visual Studio.  I just pulled latest and am now getting build errors like this

`Error	CS1703	Multiple assemblies with equivalent identity have been imported: 'C:\Users\pmiddleton\.nuget\packages\microsoft.netcore.app\2.0.0-preview1-001967-00\ref\netcoreapp2.0\System.Reflection.Emit.ILGeneration.dll' and 'C:\Users\pmiddleton\.nuget\packages\system.reflection.emit.ilgeneration\4.3.0\ref\netstandard1.0\System.Reflection.Emit.ILGeneration.dll'. Remove one of the duplicate references.	dotnet-ef.Tests	C:\code\github\Take2\EntityFramework\test\dotnet-ef.Tests\CSC	1	Active`

`Error	CS0433	The type 'IEnumerable<T>' exists in both 'System.Runtime, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' and 'System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'	dotnet-ef.Tests	C:\code\github\Take2\EntityFramework\test\dotnet-ef.Tests\CommandsTest.cs	30	Active`

smitpatel:
Things working but literally bleeding edge now.
See https://github.com/aspnet/Announcements/issues/231 for more detailed information about issues and work-around. Most likely, after all those hacks you will get it working.
Also don't forget to do git clean, clear all your nuget caches and nuke cached packages folder.

PR #8040, opened 2017-03-31
[1.1.2/TINY] making fix for issue 7714 thread safe

maumar:
Previous fix was using static field and therefore was not thread safe - multiple queries could be compiled simultaneously which could result in corrupted translations.


maumar:
merged 622e1272557e9ff8e7bf3f8e999508eab6b24a84

PR #8057, opened 2017-04-03
[1.1.2] [Tiny] Make SqlServerDatabaseModelFactoryTest use scratch database …

AndriySvyryd:
…to avoid getting in a dirty state, since `RelationalDatabaseCleaner` uses `SqlServerDatabaseModelFactory`


lajones:
@AndriySvyryd Did you check performance after this change? Admittedly a lot has gone on since then and DB initialization may well have changed, but I seem to remember that at the time there were some DB-heavy tasks that were done to initially set up the DB. If those are being repeated more than they were before it might make those tests slow.

AndriySvyryd:
@lajones I didn't measure as they definitely will be slower. But I think slow tests are better than flaky tests

lajones:
@AndriySvyryd - sort of - if they become too slow then people just don't run them. As long as you've run them and the time it takes to run them is still acceptable then that's fine.

AndriySvyryd:
@lajones On my machine running them all goes up from 4.3s to 6.6s

lajones:
@AndriySvyryd No worries then.

Issue #8054, opened 2017-04-02
Subquery with nested orderby is not lifted


anpete:
The following query:
```c#
from c1_Orders in context.Orders
join _c1 in
   (from c1 in
	(from c in context.Customers
		orderby c.CustomerID
		select c)
	.Take(2)
	from c2 in context.Customers
	select new { CustomerID = EF.Property<string>(c1, "CustomerID") })
   .Distinct()
on EF.Property<string>(c1_Orders, "CustomerID") equals _c1.CustomerID
orderby _c1.CustomerID
select c1_Orders
```
produces
```sql
@__p_0: 2

SELECT DISTINCT [t0].[CustomerID]
FROM (
    SELECT TOP(@__p_0) [c0].[CustomerID], [c0].[Address], [c0].[City], [c0].[CompanyName], [c0].[ContactName], [c0].[ContactTitle], [c0].[Country], [c0].[Fax], [c0].[Phone], [c0].[PostalCode], [c0].[Region]
    FROM [Customers] AS [c0]
    ORDER BY [c0].[CustomerID]
) AS [t0]
CROSS JOIN [Customers] AS [c20]
ORDER BY [t0].[CustomerID]

SELECT [c1_Orders].[OrderID], [c1_Orders].[CustomerID], [c1_Orders].[EmployeeID], [c1_Orders].[OrderDate]
FROM [Orders] AS [c1_Orders]
```
I think this is because of the orderby lifting happening in SE.PushDownSubquery. This query represents some Include queries that are worse with the new compiler

anpete:
@smitpatel @maumar Any ideas?

smitpatel:
What is the condition which is blocking flattening of join?

In Sql, join clause is not being materialized but inner Customer is which is weird.

anpete:
@smitpatel I think it is the presence of OrderBy without skip/take. The OrderBy is lifted to the outer table expression during Subquery push-down.

smitpatel:
If we remove the code which is lifting order by outside in PushDownSubquery, does this query work fine?

anpete:
Yes! But others don't 😬

smitpatel:
@anpete - Is this blocking issue for you? The side effects so far I have found
- Join ignores ordering for join clause. Causing ordering difference between l2o/sql queries
- Unnecessary pushdown happening when we don't lift joins.
- There is a blocking bug in ProjectionShaper.

anpete:
[...]

PR #8062, opened 2017-04-04
Query: Include: Collection SQL and generated code improvements

anpete:
- When generating collection include query models, the cloned parent subquery should explicitly project the required columns for join comparison and ordering.
- Also tweaked generated code to no longer capture QueryContext in closures.


anpete:
Ping @smitpatel @maumar

test/EFCore.SqlServer.FunctionalTests/InheritanceSqlServerTest.cs
```diff
@@ -26,15 +30,15 @@ public virtual void Common_property_shares_column()
                 var cokeType = context.Model.FindEntityType(typeof(Coke));
                 var teaType = context.Model.FindEntityType(typeof(Tea));
 
-                Assert.Equal("SugarGrams", cokeType.FindProperty("SugarGrams").Relational().ColumnName);
-                Assert.Equal("CaffeineGrams", cokeType.FindProperty("CaffeineGrams").Relational().ColumnName);
-                Assert.Equal("CokeCO2", cokeType.FindProperty("Carbination").Relational().ColumnName);
+                AssertSql("SugarGrams", cokeType.FindProperty("SugarGrams").Relational().ColumnName);
```
maumar: this should be Assert.Equal still

test/EFCore.SqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
```diff
@@ -33,17 +33,7 @@ var customers
...
-    TRACKED: True
-(QueryContext queryContext) => IEnumerable<Customer> _ShapedQuery(
-    queryContext: queryContext, 
-    shaperCommandContext: SelectExpression: 
-        SELECT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region]
-        FROM [Customers] AS [c]
-    , 
-    shaper: UnbufferedEntityShaper<Customer>
-)",
+'from Customer <generated>_0 in DbSet<Customer>",
```
maumar: we should either verify contents of optimized query model here, or remove the test altogether - in this form it doesn't provide any value over ```Queryable_simple``` test

test/EFCore.SqlServer.FunctionalTests/QueryLoggingSqlServerTest.cs
```diff
@@ -99,67 +89,8 @@ var customers
...
-            Void queryContext.QueryBuffer.StartTracking(
-                entity: entity, 
-                entityType: EntityType: Customer
-            )
-            return queryContext.QueryBuffer.IncludeCollection(0, Navigation: Customer.Orders (<Orders>k__BackingField, ICollection<Order>) Collection ToDependent Order Inverse: Customer 0 -1 1 -1 -1, Navigation: Order.Customer (<Customer>k__BackingField, Customer) ToPrincipal Customer Inverse: Orders 0 -1 2 -1 -1, EntityType: Order, value(Metadata.Internal.ClrICollectionAccessor`3[Specification.Tests.TestModels.Northwind.Customer,System.Collections.Generic.ICollection`1[Specification.Tests.TestModels.Northwind.Order],Specification.Tests.TestModels.Northwind.Order]), value(Metadata.Internal.ClrPropertySetter`2[Specification.Tests.TestModels.Northwind.Order,Specification.Tests.TestModels.Northwind.Customer]), True, entity, Convert(included[0]))
-        }
-    )
-)",
+'from Customer c in DbSet<Customer>"
+,
```
maumar: we should either verify contents of optimized query model here, or remove the test altogether - in this form it doesn't provide any value over Queryable_simple test

anpete: It is checking that we are logging everything up to an including the QueryModel, just no longer the specifics of the QM itself, which changes often.

test/EFCore.SqlServer.FunctionalTests/IncludeSqlServerTest.cs
```diff
@@ -778,46 +796,48 @@ public override void Include_duplicate_collection(bool useString)
...
+    SELECT [c4].*
+    FROM [Customers] AS [c4]
+    ORDER BY [c4].[CustomerID]
+    OFFSET 2 ROWS FETCH NEXT 2 ROWS ONLY
+) AS [t4]
+ORDER BY [t3].[CustomerID]",
+                    Sql);
 
-@__p_0: 2
+                Assert.Contains(
```
maumar: do we know why these joins are now performed on the client?

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -716,107 +804,77 @@ private static Expression CreateKeyAccessExpression(Expression target, IReadOnly
...
 
-                toQueryModel.BodyClauses.Add(outerOrderByClause);
+                    if (ordering.Expression is MemberExpression memberExpression
+                        && memberExpression.Expression
+                            is QuerySourceReferenceExpression querySourceReferenceExpression
+                        && querySourceReferenceExpression.ReferencedQuerySource == querySource)
+                    {
+                        orderingExpressionAsPropertyExpression
+                            = EntityQueryModelVisitor
+                                .CreatePropertyExpression(
```
smitpatel: Can we avoid this? Creating PropertyExpression without property.

test/EFCore.SqlServer.FunctionalTests/IncludeSqlServerTest.cs
anpete: #8054

src/EFCore/Query/Internal/AnonymousObject.cs
```diff
@@ -40,22 +49,42 @@ public struct CompositeKey
...
+                        => current + ((current * 397) ^ (argument?.GetHashCode() ?? 0)));
+            }
+        }
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public override int GetHashCode() => 0;
+        public object GetValue(int index) => _values[index];
```
smitpatel: Can we get away with having this method? Is there any other way we can correlate inner expressions of AnonymousObject from outside?

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -1208,7 +1208,7 @@ private static bool IsCompositeKey([NotNull] Type type)
         {
```
maumar: consider changing helper method name to ```IsAnonymousObject``` as welll

src/EFCore/Query/Internal/AnonymousObject.cs
```diff
@@ -40,22 +49,42 @@ public struct CompositeKey
...
-        public override bool Equals(object obj)
-            => _values.SequenceEqual(((CompositeKey)obj)._values);
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                return _values.Aggregate(
+                    0,
+                    (current, argument)
+                        => current + ((current * 397) ^ (argument?.GetHashCode() ?? 0)));
```
anpete: Yes.

src/EFCore/Query/Internal/AnonymousObject.cs
anpete: No.

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
anpete: Considered it, but I think in this case it always represents a composite key and so is more descriptive.

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -601,6 +601,37 @@ var boundExpression
...
+                        = (int)((ConstantExpression)methodCallExpression.Arguments.Single()).Value;
+
+                    var tableExpression
+                        = selectExpression
+                            .GetTableForQuerySource(
+                                querySourceReferenceExpression.ReferencedQuerySource);
+
+                    if (tableExpression is SelectExpression)
+                    {
+                        return selectExpression.Projection[projectionIndex];
```
anpete: In the first case we are translating the join predicate and so are binding directly to the join inner sequence query. In the second case we are translating the order by after the join and so are binding "through" the join expression, which has been lifted. Let me know if there is a better way here.

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -639,7 +651,7 @@ private class IncludeRemovingExpressionVisitor : RelinqExpressionVisitor
         {
             protected override Expression VisitMethodCall(MethodCallExpression node)
                 => IncludeCompiler.IsIncludeMethod(node)
-                    ? node.Arguments[0]
```
anpete: Pre-pended queryContext parameter - was previously accessed via closure.

test/EFCore.SqlServer.FunctionalTests/InheritanceSqlServerTest.cs
```diff
@@ -405,5 +412,10 @@ SELECT COUNT(*)
 ";
 
         private static string Sql => TestSqlLoggerFactory.Sql.Replace(Environment.NewLine, FileLineEnding);
+
+        private void AssertSql(string expected, string actual)
+        {
+            TestHelpers.AssertBaseline(expected, actual, _testOutputHelper);
```
smitpatel: Not specific to this PR but why do we have actual as arg here? we always pass `Sql` as actual.

anpete: Oh, nice. We should do that.

anpete:
New version up.

smitpatel:
Build failure
-- Missing NotNull annotations --
Microsoft.EntityFrameworkCore.Query.Internal.AnonymousObject.IsGetValueExpression[querySourceReferenceExpression]
Expected: False
Actual:   True

anpete:
@maumar Bug in ApiConsistencyTest :trollface: out parameter! I am fixing.

PR #8070, opened 2017-04-04
Add change tracking support for delegated identity entity types

AndriySvyryd:
Add ReferenceEntry.TargetEntry as a way to get entries for the delegated identity entities
If the navigation to owned entity is set to null the owned entity will be deleted
Changes in the inverse navigation from an owned type will be ignored as its meaning is ambiguous

### DIFF
--- a/src/EFCore/ChangeTracking/Internal/StateManager.cs
+++ b/src/EFCore/ChangeTracking/Internal/StateManager.cs
@@ -25,6 +25,9 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
         private readonly Dictionary<object, InternalEntityEntry> _entityReferenceMap
             = new Dictionary<object, InternalEntityEntry>(ReferenceEqualityComparer.Instance);
 
+        private readonly Dictionary<IEntityType, Dictionary<object, InternalEntityEntry>> _dietReferenceMap
+            = new Dictionary<IEntityType, Dictionary<object, InternalEntityEntry>>();
+
         private readonly LazyRef<IDictionary<object, IList<Tuple<INavigation, InternalEntityEntry>>>> _referencedUntrackedEntities
             = new LazyRef<IDictionary<object, IList<Tuple<INavigation, InternalEntityEntry>>>>(
                 () => new Dictionary<object, IList<Tuple<INavigation, InternalEntityEntry>>>(ReferenceEqualityComparer.Instance));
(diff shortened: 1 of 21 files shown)


src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -25,6 +25,9 @@ public class StateManager : IStateManager
         private readonly Dictionary<object, InternalEntityEntry> _entityReferenceMap
             = new Dictionary<object, InternalEntityEntry>(ReferenceEqualityComparer.Instance);
 
+        private readonly Dictionary<IEntityType, Dictionary<object, InternalEntityEntry>> _dietReferenceMap
```
ajcvickers: >_dietReferenceMap [](start = 90, length = 17)

If I understand things correctly, we need this so that the same entity instance can be shared by multiple different "owners". Is that correct? If so, is it necessary to allow this?

AndriySvyryd: Correct. It's not necessary for owned entities, but it is necessary for DIETs and we decided that all owned entities will have a DIET

ajcvickers: Why is is necessary for DIETs?

AndriySvyryd: To be able to reassign it from one reference to another. Ok, strictly speaking it's not necessary, but I feel it does improve the experience considerably.

ajcvickers: By re-assign, do you mean if I do this:
```C#
customer.Address = office.Address;
```
then, after DetectChanges, both customer and office reference the same object instance? Does this mean that the same instance can have different key values?

AndriySvyryd: I was referring to
```
customer.Address = office.Address;
office.Address = new Address();
```

ajcvickers: So is this needed only for the temporary state where customer.Address and office.Address are the same instance? Since at the end of these two lines the same instance is not shared anymore.

AndriySvyryd: Correct. As a side-effect instance sharing is possible if the PK is shadow, but this is not the goal.

ajcvickers: Signing off for now, but think we should probably re-visit.

AndriySvyryd: Perhaps this can be removed if we tracked deleted entries separately, but we would need to ensure old one is marked for deletion first.

Issue #6611, opened 2016-09-27
Hundred of queries being ran. When calling OrderBy, then Include then Where to a one to many relationship


fujiy:
### Steps to reproduce

Full example: [EfCoreMultipleQueriesIssue.zip](https://github.com/aspnet/EntityFramework/files/494360/EfCoreMultipleQueriesIssue.zip)

IQueryable<Post> queryBug = db.Posts.OrderBy(x => x.Id).Include(x => x.PostTags).Where(x => x.PostTags.Any());
//Hundreds of
//SELECT[p].[PostId]
//FROM[PostTag] AS[p]

IQueryable<Post> queryBug2 = db.Posts.OrderBy(x => x.Id).Include(x => x.PostTags).Where(x => x.PostCategories.Any());
//Hundreds of
//SELECT[p].[PostId]
//FROM[PostTag] AS[p]

IQueryable<Post> queryOk = db.Posts.Where(x => x.PostTags.Any()).OrderBy(x => x.Id).Include(x => x.PostTags);

IQueryable<Post> queryOk2 = db.Posts.Include(x => x.PostTags).Where(x => x.PostTags.Any()).OrderBy(x => x.Id);

IQueryable<Post> queryOk3 = db.Posts.Include(x => x.PostTags).OrderBy(x => x.Id).Where(x => x.PostTags.Any());
### The issue

Looks like part of the query is being evaluated locally, generating all those queries. It's not clear why.

maumar:
Initially when the query is parsed, each Include call is treated as result operator, effectively producing a subquery, unless it's the last operation, e.g.:
```
db.Posts.Include(x => x.PostTags).Where(x => x.PostTags.Any()).OrderBy(x => x.Id)
```

translates to:
```
from Post x in 
    (from Post x in DbSet<Post>
    select x)
    .Include(x.PostTags)
where 
    (from PostTag <generated>_1 in x.PostTags
    select <generated>_1)
    .Any()
order by x.Id asc
select x
```

in the next step we extract the include operators out of the query:
```
from Post x in 
    from Post x in DbSet<Post>
    select x
where 
    (from PostTag <generated>_1 in x.PostTags
    select <generated>_1)
    .Any()
order by x.Id asc
select x
```

and then we remove redundant subqueries:
```
from Post x in DbSet<Post>
where 
    (from PostTag <generated>_1 in DbSet<PostTag>
    where Property(x, "Id") == Property(<generated>_1, "PostId")
[...]

PR #8073, opened 2017-04-04
[SMALL] Fix to #6611 - Hundred of queries being ran. When calling OrderBy, then Include then Where to a one to many relationship

maumar:
Problem was that during initial parsing of query model, Include, AsTracking, AsNoTracking force subqueries to be created (since they are result operators). Those operators are then removed, but the subqueries persist, making it much harder to translate queries efficiently. We try to prune "empty" subqueries, but sometimes it is safe to lift subqueries that contain body clauses.

Fix is to recognize those patterns and perform the lift during query model optimizations. We do it for cases when the subquery is coming from the MainFromClause, it doesn't contain any result operators and is projecting it's MainFromClause only in the selector.
The optimization is the most prominent for queries with Include but will help other cases also.

### DIFF
--- a/src/EFCore/Query/Internal/QueryOptimizer.cs
+++ b/src/EFCore/Query/Internal/QueryOptimizer.cs
@@ -173,12 +173,19 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
             VisitQueryModel(subQueryModel);
 
+            var subqueryInMainClauseWithoutResultOperatorsProjectingItsMainClause
+                = fromClause is MainFromClause
+                    && !subQueryModel.ResultOperators.Any()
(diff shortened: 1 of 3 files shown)


src/EFCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -173,12 +173,19 @@ var additionalFromClause
 
             VisitQueryModel(subQueryModel);
 
+            var subqueryInMainClauseWithoutResultOperatorsProjectingItsMainClause
+                = fromClause is MainFromClause
+                    && !subQueryModel.ResultOperators.Any()
+                    && subQueryModel.SelectClause.Selector is QuerySourceReferenceExpression subquerySelectorsQsre
```
smitpatel: Simplify.

PR #8061, opened 2017-04-03
Query: Convert SetOperator.Source2 from EntityQueryable to QueryModel…

smitpatel:
… when applicable.

This causes issue when translating EntityQueryable to SelectExpression as there is no query source.

resolves #7279

PR #8074, opened 2017-04-04
Query: DRY IsEntityQueryable

smitpatel:
Follow up PR for #8061 

@anpete

### DIFF
--- a/src/EFCore/Query/Internal/QueryOptimizer.cs
+++ b/src/EFCore/Query/Internal/QueryOptimizer.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Extensions.Internal;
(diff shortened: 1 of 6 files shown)


src/EFCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -308,47 +308,34 @@ private static void ProcessSetResultOperator(ResultOperatorBase resultOperator)
...
+        private static Expression ConvertEntityQueryableToSubQuery(Expression expression)
         {
-            return expression is ConstantExpression constantExpression
-                   && constantExpression.Type.IsConstructedGenericType
-                   && constantExpression.Type.GetGenericTypeDefinition() == typeof(EntityQueryable<>);
-        }
+            if ((expression as ConstantExpression)?.IsEntityQueryable() ?? false)
+            {
+                var mainFromClause = new MainFromClause(
+                    "dummy",
```
smitpatel: @anpete - :trollface:

PR #8076, opened 2017-04-05
Query: Include: Use nav rewrite to generate skeleton collection include queries.

anpete:
Leverage nav rewrite for initial creation of collection include queries.

### DIFF
--- a/src/EFCore.Specification.Tests/FieldMappingTestBase.cs
+++ b/src/EFCore.Specification.Tests/FieldMappingTestBase.cs
@@ -233,7 +233,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         public virtual void Update_read_only_props_with_named_fields()
             => Update<BlogReadOnlyExplicit>("Posts");
 
-        [Fact]
+        [Fact(Skip = "#8068")]
         public virtual void Include_collection_write_only_props()
         {
             using (var context = CreateContext())
@@ -275,7 +275,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         public virtual void Update_write_only_props()
             => Update<BlogWriteOnly>("Posts");
 
-        [Fact]
+        [Fact(Skip = "#8068")]
         public virtual void Include_collection_write_only_props_with_named_fields()
         {
             using (var context = CreateContext())
--- a/src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
@@ -1619,10 +1619,10 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
                 Assert.Equal("Marcus", result[0].Nickname);
                 Assert.Equal(2, result[0].Weapons.Count);
-                Assert.Equal("Marcus", result[1].Nickname);
+                Assert.Equal("Baird", result[1].Nickname);
+                Assert.Equal(2, result[1].Weapons.Count);
                 Assert.Equal("Marcus", result[2].Nickname);
-                Assert.Equal("Baird", result[3].Nickname);
-                Assert.Equal(2, result[3].Weapons.Count);
+                Assert.Equal("Marcus", result[3].Nickname);
                 Assert.Equal("Marcus", result[4].Nickname);
             }
         }
@@ -1660,10 +1660,10 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
 
                 Assert.Equal("Marcus", result[0].Nickname);
                 Assert.Equal(2, result[0].Weapons.Count);
-                Assert.Equal("Marcus", result[1].Nickname);
+                Assert.Equal("Baird", result[1].Nickname);
+                Assert.Equal(2, result[1].Weapons.Count);
                 Assert.Equal("Marcus", result[2].Nickname);
-                Assert.Equal("Baird", result[3].Nickname);
-                Assert.Equal(2, result[3].Weapons.Count);
+                Assert.Equal("Marcus", result[3].Nickname);
                 Assert.Equal("Marcus", result[4].Nickname);
             }
         }
--- a/src/EFCore.SqlServer/Storage/Internal/SqlServerExecutionStrategy.cs
+++ b/src/EFCore.SqlServer/Storage/Internal/SqlServerExecutionStrategy.cs
@@ -19,7 +19,9 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
         public virtual bool RetriesOnFailure => false;
 
         public virtual TResult Execute<TState, TResult>(
-            Func<TState, TResult> operation, Func<TState, ExecutionResult<TResult>> verifySucceeded, TState state)
+            Func<TState, TResult> operation,
+            Func<TState, ExecutionResult<TResult>> verifySucceeded,
+            TState state)
         {
             try
             {
--- a/src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -1378,6 +1378,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                 var groupResultOperator = resultOperator as GroupResultOperator;
                 if (groupResultOperator != null)
                 {
+                    groupResultOperator.ElementSelector 
+                        = _subqueryInjector.Visit(groupResultOperator.ElementSelector);
+
                     var originalKeySelectorType = groupResultOperator.KeySelector.Type;
                     var originalElementSelectorType = groupResultOperator.ElementSelector.Type;
 
--- a/src/EFCore/Query/Internal/ExpressionPrinter.cs
+++ b/src/EFCore/Query/Internal/ExpressionPrinter.cs
@@ -352,7 +352,8 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
             if (blockExpression.Result != null)
             {
-                AppendLine("return " + blockExpression.Result);
+                Append("return ");
+                Visit(blockExpression.Result);
             }
 
             _stringBuilder.DecrementIndent();
@@ -735,7 +736,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
             switch (extensionExpression)
             {
                 case QuerySourceReferenceExpression qsre:
-                    StringBuilder.Append(qsre.ReferencedQuerySource.ItemName);
+                    StringBuilder.Append(qsre);
                     break;
 
                 case NullConditionalExpression nullConditional:
--- a/src/EFCore/Query/Internal/IncludeCompiler.cs
+++ b/src/EFCore/Query/Internal/IncludeCompiler.cs
@@ -16,6 +16,7 @@ using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
 using Microsoft.EntityFrameworkCore.Query.ResultOperators;
 using Microsoft.EntityFrameworkCore.Query.ResultOperators.Internal;
@@ -122,8 +123,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                 = CreateIncludeSpecifications(queryModel, includeResultOperators)
                     .GroupBy(a => a.QuerySourceReferenceExpression);
 
-            var parentOrderings = new List<Ordering>();
-
             var compiledIncludes
                 = new List<(
                     QuerySourceReferenceExpression querySourceReferenceExpression,
@@ -164,19 +163,9 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
                     if (navigation.IsCollection())
                     {
-                        var collectionIncludeQueryModel
-                            = BuildCollectionIncludeQueryModel(
-                                queryModel,
-                                navigation,
-                                includeSpecification.QuerySourceReferenceExpression,
-                                parentOrderings);
-
-                        _queryCompilationContext.AddQuerySourceRequiringMaterialization(
-                            ((QuerySourceReferenceExpression)collectionIncludeQueryModel.SelectClause.Selector).ReferencedQuerySource);
-
-                        Expression collectionLambdaExpression 
+                        Expression collectionLambdaExpression
                             = Expression.Lambda<Func<IEnumerable<object>>>(
-                                new SubQueryExpression(collectionIncludeQueryModel));
+                                Expression.Property(includeSpecification.QuerySourceReferenceExpression, navigation.PropertyInfo));
 
                         var includeCollectionMethodInfo = _queryBufferIncludeCollectionMethodInfo;
 
@@ -215,7 +204,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                                                 EntityQueryModelVisitor.CreatePropertyExpression)));
 
                         blockExpressions.Add(
-                            BuildIncludeExpressions(
+                            BuildReferenceIncludeExpressions(
                                 includeSpecification.NavigationPath,
                                 entityParameter,
                                 trackingQuery,
@@ -287,12 +276,12 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                 compiledIncludes.Add((includeGrouping.Key, includeExpression));
             }
 
-            ApplyParentOrderings(queryModel, parentOrderings);
             ApplyIncludeExpressionsToQueryModel(queryModel, compiledIncludes);
         }
 
         private static void ApplyParentOrderings(
-            QueryModel queryModel, IReadOnlyCollection<Ordering> parentOrderings)
+            QueryModel queryModel,
+            IReadOnlyCollection<Ordering> parentOrderings)
         {
             if (parentOrderings.Any())
             {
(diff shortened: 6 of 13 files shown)


src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
+
+                for (var i = 0; i < subQueryProjection.Count; i++)
+                {
+                    if (subQueryProjection[i] is UnaryExpression unaryExpression
+                        && unaryExpression.Operand is MethodCallExpression methodCallExpression
+                        && EntityQueryModelVisitor.IsPropertyMethod(methodCallExpression.Method))
+                    {
+                        subQueryProjection[i]
+                            = unaryExpression.Update(
+                                new NullConditionalExpression(
```
anpete: @maumar Any better place to apply NullConditional?

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -1378,6 +1378,9 @@ public override void VisitResultOperator(ResultOperatorBase resultOperator, Quer
                 var groupResultOperator = resultOperator as GroupResultOperator;
                 if (groupResultOperator != null)
                 {
+                    groupResultOperator.ElementSelector 
+                        = _subqueryInjector.Visit(groupResultOperator.ElementSelector);
+
```
anpete: @maumar Think this is just a bug, I had to add this here because for GroupBy, we use the projection expression from the result operator, not the SelectClause.

src/EFCore/Query/Internal/ExpressionPrinter.cs
```diff
@@ -352,7 +352,7 @@ protected override Expression VisitBlock(BlockExpression blockExpression)
 
             if (blockExpression.Result != null)
             {
-                AppendLine("return " + blockExpression.Result);
+                AppendLine("return " + Visit(blockExpression.Result));
```
maumar:
```
AppendLine("return ");
Visit(blockExpression.Result);
```

Otherwise stuff will get printed twice

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
maumar: This should be good, it needs to happen before ```base.VisitResultOperator(resultOperator, queryModel, index);``` 2 lines below, and subquery injection won't change the type, so the ```originalElementSelectorType``` won't be affected by this

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -287,12 +276,12 @@ Expression collectionLambdaExpression
                 compiledIncludes.Add((includeGrouping.Key, includeExpression));
```
maumar: nit: unnecessary parens

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
-                    new SelectClause(collectionQuerySourceReferenceExpression));
+            public List<Ordering> ParentOrderings { get; } = new List<Ordering>();
 
-            var querySourceMapping = new QuerySourceMapping();
-            var clonedParentQueryModel = parentQueryModel.Clone(querySourceMapping);
+            protected override Expression VisitMethodCall(MethodCallExpression methodCallExpression)
+            {
+                if (typeof(IQueryBuffer).GetTypeInfo()
+                        .IsAssignableFrom(methodCallExpression.Object?.Type.GetTypeInfo())
+                    && methodCallExpression.Method.Name.StartsWith("IncludeCollection", StringComparison.Ordinal))
```
maumar: why ```StartsWith``` over ```==```? also, use ```nameof (IQueryBuffer.IncludeCollection)```

anpete: Two versions (async). I technically should be checking for the methodinfos and this was meant to be temporary :trollface:

src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
```diff
@@ -1619,10 +1619,10 @@ public virtual void Include_on_GroupJoin_SelectMany_DefaultIfEmpty_with_coalesce
 
                 Assert.Equal("Marcus", result[0].Nickname);
                 Assert.Equal(2, result[0].Weapons.Count);
-                Assert.Equal("Marcus", result[1].Nickname);
+                Assert.Equal("Baird", result[1].Nickname);
+                Assert.Equal(2, result[1].Weapons.Count);
                 Assert.Equal("Marcus", result[2].Nickname);
-                Assert.Equal("Baird", result[3].Nickname);
-                Assert.Equal(2, result[3].Weapons.Count);
```
smitpatel: Does this mean we were asserting incorrectly before?

src/EFCore.SqlServer/Storage/Internal/SqlServerExecutionStrategy.cs
```diff
@@ -31,7 +31,7 @@ private SqlServerExecutionStrategy()
                 {
                     throw new InvalidOperationException(SqlServerStrings.TransientExceptionDetected, ex);
                 }
-
+
```
smitpatel: space

src/EFCore/Query/Internal/IncludeCompiler.cs
smitpatel: ^Online resharper check.

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
+                {
+                    var lambaArgument = methodCallExpression.Arguments[8];
+                    var convertExpression = lambaArgument as UnaryExpression;
```
smitpatel: ~Does remove convert work here?~
It is not even check if its a convert expression. What is it supposed to do?

src/EFCore/Query/Internal/IncludeCompiler.cs
smitpatel: Why do we need this conversion?

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
+                Expression targetParentExpression)
             {
-                var index = _querySources.IndexOf(querySource);
+                var querySourcePriorityAnalyzer
+                    = new QuerySourcePriorityAnalyzer(queryModel.SelectClause.Selector);
+
+                Expression predicate = null;
 
-                if (index != -1)
+                if (querySourcePriorityAnalyzer.AreLowerPriorityQuerySources(parentQuerySource))
```
smitpatel: Function name is pluralized but taking one arg only. Which is the other querysource?

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
-                queryModel.ResultOperators.Add(new TakeResultOperator(Expression.Constant(1)));
+                foreach (var groupResultOperator
+                    in queryModel.ResultOperators.OfType<GroupResultOperator>()
+                        .ToArray())
+                {
+                    queryModel.ResultOperators.Remove(groupResultOperator);
 
-                lastResultOperator = choiceResultOperator is LastResultOperator;
-            }
+                    var orderByClause1 = queryModel.BodyClauses.OfType<OrderByClause>().LastOrDefault();
```
smitpatel: better variable name

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
+                                targetExpression,
+                                AnonymousObject.GetValueMethodInfo,
+                                Expression.Constant(projectionIndex));
+
+                        outerOrderByClause.Orderings
+                            .Add(new Ordering(newExpression, ordering.OrderingDirection));
+
+                        toQueryModel.BodyClauses.Add(outerOrderByClause);
+
+                        if (canRemove)
```
smitpatel: Shouldn't this be 1 level outer?

test/EFCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
```diff
@@ -1993,14 +1993,15 @@ public override void Multiple_SelectMany_with_Include()
...
+FROM [Level4] AS [l1.OneToMany_Optional.OneToMany_Optional.OneToMany_Optional]
 INNER JOIN (
     SELECT DISTINCT [l1.OneToMany_Optional.OneToMany_Optional0].[Id]
     FROM [Level1] AS [l10]
     INNER JOIN [Level2] AS [l1.OneToMany_Optional0] ON [l10].[Id] = [l1.OneToMany_Optional0].[OneToMany_Optional_InverseId]
     INNER JOIN [Level3] AS [l1.OneToMany_Optional.OneToMany_Optional0] ON [l1.OneToMany_Optional0].[Id] = [l1.OneToMany_Optional.OneToMany_Optional0].[OneToMany_Optional_InverseId]
-) AS [t] ON [l3.OneToMany_Optional].[OneToMany_Optional_InverseId] = [t].[Id]
+    LEFT JOIN [Level4] AS [l1.OneToMany_Optional.OneToMany_Optional.OneToOne_Required_FK0] ON [l1.OneToMany_Optional.OneToMany_Optional0].[Id] = [l1.OneToMany_Optional.OneToMany_Optional.OneToOne_Required_FK0].[Level3_Required_Id]
```
smitpatel: Was this incorrect query earlier?

test/EFCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
```diff
@@ -1522,26 +1522,12 @@ public override void Include_on_GroupJoin_SelectMany_DefaultIfEmpty_with_coalesc
...
-    FROM [Gear] AS [g0]
-    LEFT JOIN (
-        SELECT [g20].*
-        FROM [Gear] AS [g20]
-        WHERE [g20].[Discriminator] IN (N'Officer', N'Gear')
-    ) AS [t0] ON [g0].[LeaderNickname] = [t0].[Nickname]
-    WHERE [g0].[Discriminator] IN (N'Officer', N'Gear') AND [t0].[Nickname] IS NOT NULL
-) AS [t1] ON [g2.Weapons].[OwnerFullName] = [t1].[FullName]
-ORDER BY [t1].[c0], [t1].[FullName]
+ORDER BY [t].[FullName], [g].[FullName]
```
smitpatel: Beware of ordering issues on .NET Core

test/EFCore.SqlServer.FunctionalTests/IncludeSqlServerTest.cs
```diff
@@ -43,6 +46,20 @@ SELECT 1
                 Sql);
         }
 
+        public override void Include_reference(bool useString)
+        {
+            base.Include_reference(useString);
+
+            if (!useString)
```
smitpatel: Why conditional assertion?

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
+                var joinQuerySourceReferenceExpression
+                    = CreateJoinToParentQuery(
+                        clonedParentQueryModel,
+                        clonedParentQuerySourceReferenceExpression,
+                        collectionQuerySourceReferenceExpression,
+                        navigation.ForeignKey,
+                        collectionQueryModel,
+                        subQueryProjection);
+
+                ApplyParentOrderings(
```
maumar: this applies orderings to clonedParentQueryModel and collectionQueryModel. Example InMemory test: Include_collection_on_join_clause_with_order_by_and_filter

src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
maumar: order is now different, it might be good idea to make the test more deterministic by adding explicit ordering

src/EFCore/Query/Internal/IncludeCompiler.cs
maumar: @anpete this seems like a good spot. In the future we could be smarter about applying NullConditional (only when it's needed) - If we had information about which query source comes from optional navigation chain/groupjoin. 
@smitpatel some inmemory scenarios break

smitpatel: That was what my question was for. Blanket changes to NCE. Its fine to improve this in future.

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
+                LiftOrderBy(
+                    clonedParentQuerySource,
+                    joinQuerySourceReferenceExpression,
+                    clonedParentQueryModel,
+                    collectionQueryModel,
+                    subQueryProjection);
+
+                for (var i = 0; i < subQueryProjection.Count; i++)
+                {
+                    if (subQueryProjection[i] is UnaryExpression unaryExpression
```
maumar: consider hard cast instead. AFAIU those projections are guaranteed to be wrapped in Convert(..., typeof(object))

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -386,488 +375,597 @@ private class IncludeReplacingExpressionVisitor : RelinqExpressionVisitor
...
             }
 
-            public IEnumerable<IQuerySource> GetHigherPriorityQuerySources(IQuerySource querySource)
+            private static void AdjustPredicate(
+                QueryModel queryModel,
+                IQuerySource parentQuerySource,
+                Expression targetParentExpression)
             {
-                var index = _querySources.IndexOf(querySource);
+                var querySourcePriorityAnalyzer
```
maumar: consider performing the analysis via Visit() method, rather than in ctor.

src/EFCore/Query/Internal/IncludeCompiler.cs
anpete: @smitpatel Specifically, since this value can be null and the QSRE is translated to a ValueBuffer, we fail with null ref because we end up binding to TryGetValue over ValueBuffer. I see special code in MemberBinding to deal with some of these cases but it was hard to integrate there. Ideally we would do it during member binding because we already do for BinaryExpression case.

src/EFCore/Query/Internal/IncludeCompiler.cs
anpete: @smitpatel Extra parens for clarity!

anpete: But really it's a tuple.

src/EFCore/Query/Internal/IncludeCompiler.cs
anpete: There can be multiple query sources of lower priority than the passed one.

test/EFCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
anpete: No, just different.

src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
anpete: This shouldn't change again - it's a function of the algorithm.

src/EFCore/Query/Internal/IncludeCompiler.cs
anpete: I want to remove Convert but put it back if it was there.

src/EFCore/Query/Internal/IncludeCompiler.cs
anpete: Yep, bad merge!

test/EFCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
anpete: Will check.

test/EFCore.SqlServer.FunctionalTests/IncludeSqlServerTest.cs
anpete: Re-linq gens different QM.

src/EFCore/Query/Internal/IncludeCompiler.cs
anpete: Going to leave this as is because this is simpler.

src/EFCore/Query/Internal/IncludeCompiler.cs
anpete: Still not seeing this. clonedParentQueryModel is referenced in the JoinClause of collectionQueryModel and so it is expected that collectionQueryModel looks like it has changed.

src/EFCore/Query/Internal/IncludeCompiler.cs
anpete: > Why do we need this conversion
Because AnonymousObject ctor takes object[] so we need to box any value types.

src/EFCore.Specification.Tests/FieldMappingTestBase.cs
```diff
@@ -242,7 +242,6 @@ public virtual void Include_collection_write_only_props()
             }
         }
 
-        [Fact]
```
anpete: Removed in error.

anpete:
@smitpatel @maumar New version up.

Issue #8004, opened 2017-03-27
Implement query Like function in core


anpete:
With the introduction of relational specific functions like LIKE, we are making it harder to test against In-memory. Where a client-side impl. is plausible, we should consider moving the function to core and providing the client-side impl.

ajcvickers:
Also, make sure that we follow this generally for new functions.

Issue #8050, opened 2017-03-31
Test: Multi-level hierarchy of tests requires Fact attributes to be specified again


smitpatel:
Those tests can be made part of middle layer from `QueryTestBase` to provider specific QueryTest.

It creates 3 level hierarchy which is blocked on https://github.com/xunit/xunit/issues/1068

smitpatel:
RelationalQueryTestBase was removed in #8084 .
This is no longer needed.

smitpatel:
We are hitting this issue in `ComplexNavigationOwnedQueryTestBase` now. Looking at overriding our test discoverer to make it work & remove it once xunit fixes the issue.

smitpatel:
Removing from backlog. This cannot be fixed without upgrading our specs projects to xunit beta4 or higher but specs are stuck on 2.2. We need to follow up with aspnet team on this.

PR #8084, opened 2017-04-05
Query: Core: Implement query Like function in core

anpete:
Fix: #8004

Implemented as case-insensitive by default, so as to be most useful.

### DIFF
--- a/test/EFCore.Tests/DbFunctionsTest.cs
+++ b/test/EFCore.Tests/DbFunctionsTest.cs
@@ -126,9 +126,13 @@ namespace Microsoft.EntityFrameworkCore.Tests
         {
             Assert.True(_functions.Like("abc\r\ndef", "abc%"));
             Assert.True(_functions.Like("abc\r\ndef", "abc__def"));
+            Assert.True(_functions.Like("abc\ndef", "abc_def"));
+            Assert.True(_functions.Like("abc\rdef", "abc_def"));
 
             Assert.False(_functions.Like("abc\r\ndef", "ab%c"));
             Assert.False(_functions.Like("abc\r\ndef", "abc_def"));
+            Assert.False(_functions.Like("abc\ndef", "abcdef"));
+            Assert.False(_functions.Like("abc\rdef", "abcdef"));
         }
     }
 }
--- a/src/EFCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
@@ -3174,12 +3174,12 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
                 (e, a) => Assert.Equal(e.Id, a.Id));
         }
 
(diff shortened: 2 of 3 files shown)


src/EFCore/DbFunctionsExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                return false;
+            }
+
+            if (matchExpression.Equals(pattern, StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+
+            if (matchExpression.Length == 0
```
maumar: if both are empty strings LIKE (on SqlServer) returns true

test/EFCore.Tests/DbFunctionsTest.cs
```diff
@@ -0,0 +1,135 @@
...
+            Assert.False(_functions.Like("ab", "abc"));
+        }
+
+        [Fact]
+        public void Like_when_wildcards()
+        {
+            Assert.True(_functions.Like("abc", "%"));
+            Assert.True(_functions.Like("abc", "%_"));
+            Assert.True(_functions.Like("abc", "___"));
+            Assert.True(_functions.Like("ABC", "a_c"));
```
maumar: those two are the same

smitpatel:
supersedes #8050

test/EFCore.Tests/DbFunctionsTest.cs
```diff
@@ -0,0 +1,135 @@
...
+            Assert.True(_functions.Like("abc ", "abc"));
+            Assert.True(_functions.Like("abc  ", "ABC "));
+
+            Assert.False(_functions.Like("ABC", "ab "));
+        }
+
+        [Fact]
+        public void Like_when_multiline()
+        {
+            Assert.True(_functions.Like("abc\r\ndef", "abc%"));
```
smitpatel: can we use Environment.NewLine instead of \r\n to avoid issues in cross-plat testing.

src/EFCore/DbFunctionsExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+        public static bool Like(this DbFunctions _, string matchExpression, string pattern, char escapeCharacter)
+            => LikeCore(matchExpression, pattern, escapeCharacter);
+
+        // Regex special chars defined here:
+        // https://msdn.microsoft.com/en-us/library/4edbef7e(v=vs.110).aspx
+
+        private static readonly char[] _regexSpecialChars
+            = { '.', '$', '^', '{', '[', '(', '|', ')', '*', '+', '?', '\\' };
+
+        private static readonly string _defaultEscapeRegexCharsPattern
```
maumar: should we cache this?

smitpatel:
Build failing. Update SQLite functional tests.

anpete:
@roji Perhaps PG should translate to ILIKE here?

test/EFCore.Tests/DbFunctionsTest.cs
anpete: No cross plat issue here - I added some assertions with the other flavors.

Issue #7956, opened 2017-03-22
Scaffold-DbContext generate IsRequired() on nullable unique column.


sonphnt:
Hi Everyone,

I have a column with unique index like this.
```
Name (nvarchar(100), null)
CREATE UNIQUE NONCLUSTERED INDEX [IX_Account_Name_unique] ON [dbo].[Account]
(
	[Name] ASC
)
```
The class generated by Scaffolding look like this

```
entity.HasIndex(e => e.Name)
                    .HasName("IX_Account_Name_unique")
                    .IsUnique();

entity.Property(e => e.Name)
                    .IsRequired()
                    .HasMaxLength(100);
```

Why it added IsRequired() into Name property?
Is it a bug? And is it possible to do a unique nullable column in Entity Framework?

Thanks

sonphnt:
If I set like this
`Name (nvarchar(100), not null)`

There is no IsRequired() added. It looks weird. It should be the other way around.

ErikEJ:
What is the value of the "nullable" column for that table/column when you run this:

https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore.SqlServer.Design/Internal/SqlServerDatabaseModelFactory.cs#L322

sonphnt:
Hi @ErikEJ  What do you mean. How can I run what you mean in your link?

ErikEJ:
In SQL Server management studio against your database

sonphnt:
Name column nullable value = 1 in SSMS

ErikEJ:
So it is currently 

    Name (nvarchar(100), null)

sonphnt:
Yes, exactly

sonphnt:
Did you find anything @ErikEJ ?
Thanks

ajcvickers:
We discussed this in triage and decided that we should reverse engineer what the schema says even if it is not something we would generate by default. We should also look at whether or not this should generate warnings in model validation.

sonphnt:
Hi @ajcvickers .

Yes the flow should be reversed with SQL schema. Column allow null then it generated IsRequired(), Column "not null" then there was no IsRequired() specification.

For now after scaffolding I have to change DbContext file manually to make EF querying correctly.

Thanks

lajones:
Note: this doesn't happen if you don't create the index; then `Name` is correctly _not_ assigned the `IsRequired()` fluent API.

[...]

PR #8089, opened 2017-04-06
A unique index only means an alternate key if it is the principal end of an FK

lajones:
Fixes issue #7956. 

Previously a unique index implied that the properties on which that index is based would be set as required even if the underlying columns were nullable. Now we only do this if the properties are the principal end of a foreign key. This prevents columns which just happen to use a unique index but have nothing to do with a  FK being set to required when this is not necessary.

### DIFF
--- a/src/EFCore.Relational.Design/Properties/RelationalDesignStrings.resx
+++ b/src/EFCore.Relational.Design/Properties/RelationalDesignStrings.resx
@@ -168,4 +168,7 @@
   <data name="MissingTable" xml:space="preserve">
     <value>Unable to find a table in the database matching the selected table {table}.</value>
   </data>
+  <data name="ForeignKeyPrincipalEndContainsNullableColumns" xml:space="preserve">
+    <value>The principal end of the foreign key '{foreignKeyName}' contains the following nullable columns '{columnNames}'. Entity Framework requires the properties representing those columns to be non-nullable.</value>
+  </data>
 </root>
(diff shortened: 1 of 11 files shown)


src/EFCore.Relational.Design/Properties/RelationalDesignStrings.resx
```diff
@@ -168,4 +168,7 @@
   <data name="MissingTable" xml:space="preserve">
     <value>Unable to find a table in the database matching the selected table {table}.</value>
   </data>
+  <data name="ForeignKeyPrincipalEndContainsNullableColumns" xml:space="preserve">
+    <value>The principal end of the foreign key '{foreignKeyName}' contains the following nullable columns '{columnNames}'. Entity Framework requires the properties representing those columns to be non-nullable.</value>
+  </data>
```
lajones: Forgot to mention the index here. I'll make it something like "The principal end of the foreign key '{foreignKeyName}' is supported by a unique index '{indexName}' which contains the following nullable columns '{columnNames}'...." and check in.

lajones:
Checked in with commit 439f510b7.

lajones:
Added `providers-beware`. Don't think this will have a big effect on providers, but the change affects, in some edge cases, the model that `RelationalScaffoldingModelFactory` produces. Providers inherit form this - so this is a heads-up for those changes.

smitpatel:
We should avoid having PRs without proper merge. If we merge PR with fast forward then git marks it as merged PR and also stores reference to it while looking at commit. This makes it much easier to come back to PR from a commit at later stage if needed.

lajones:
@smitpatel we should talk about this offline. I just merged from the command-line (I did not specify the ff flag) and everything looks normal to me on `gitk`.

PR #8093, opened 2017-04-06
Add Service Item to test projects.

lajones:
VS automatically adds these `<Service>` items to the test .csproj's. They do no harm and this prevents us from always having to revert them when we check in.


lajones:
Checked in with commit ae0cab9c8.

PR #8096, opened 2017-04-06
Add query support for delegated identity entity types

AndriySvyryd:
* Add QueryCompilationContext.QuerySourceEntityTypeMapping to keep track of the mapping from IQuerySource to IEntityType
* Change EntityQueryModelVisitor.IterateCompositePropertyExpression to propagate the IEntityType through navigations and rename it to MemberAccessBindingExpressionVisitor.GetPropertyPath

### DIFF
--- a/src/EFCore.InMemory/Query/ExpressionVisitors/Internal/InMemoryEntityQueryableExpressionVisitor.cs
+++ b/src/EFCore.InMemory/Query/ExpressionVisitors/Internal/InMemoryEntityQueryableExpressionVisitor.cs
@@ -51,7 +51,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         {
             Check.NotNull(elementType, nameof(elementType));
 
-            var entityType = _model.FindEntityType(elementType);
+            var entityType = QueryModelVisitor.QueryCompilationContext.FindEntityType(_querySource)
+                             ?? _model.FindEntityType(elementType);
 
             if (QueryModelVisitor.QueryCompilationContext
                 .QuerySourceRequiresMaterialization(_querySource))
--- a/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -1582,8 +1582,7 @@ namespace Microsoft.EntityFrameworkCore.Query
 
             foreach (var mapping in previousMapping)
             {
-                QueryCompilationContext.QuerySourceMapping
-                    .ReplaceMapping(mapping.Key, mapping.Value);
+                QueryCompilationContext.AddOrUpdateMapping(mapping.Key, mapping.Value);
             }
 
             var innerItemParameter
@@ -1591,7 +1590,7 @@ namespace Microsoft.EntityFrameworkCore.Query
                     innerShaper.Type,
                     joinClause.ItemName);
 
-            AddOrUpdateMapping(joinClause, innerItemParameter);
+            QueryCompilationContext.AddOrUpdateMapping(joinClause, innerItemParameter);
 
             var transparentIdentifierType
                 = CreateTransparentIdentifierType(
--- a/src/EFCore/Query/EntityQueryModelVisitor.cs
+++ b/src/EFCore/Query/EntityQueryModelVisitor.cs
@@ -1,9 +1,10 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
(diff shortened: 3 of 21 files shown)


src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1597,16 +1597,15 @@ var referencedQuerySource
...
+                QueryCompilationContext.AddOrUpdateMapping(mapping.Key, mapping.Value);
             }
 
             var innerItemParameter
                 = Expression.Parameter(
                     innerShaper.Type,
                     joinClause.ItemName);
 
-            AddOrUpdateMapping(joinClause, innerItemParameter);
+            QueryCompilationContext.AddOrUpdateMapping(joinClause, innerItemParameter);
```
smitpatel: Why this was changed? `AddOrUpdatingMapping` does the same thing. Either we keep the function and use it in all cases (including 1 above) or we remove the function.

AndriySvyryd: I removed the function as it seemed redundant

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -752,7 +756,8 @@ public override void VisitQueryModel([NotNull] QueryModel queryModel)
                     _expression.Type.GetSequenceType(),
                     fromClause.ItemName);
 
-            AddOrUpdateMapping(fromClause, CurrentParameter);
+            Expression expression = CurrentParameter;
```
smitpatel: pass directly to function?

AndriySvyryd: Refactoring gone wrong

src/EFCore/Query/ExpressionVisitors/Internal/QuerySourceTypeExpressionVisitor.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitConstant(ConstantExpression constantExpression)
+        {
+            if (constantExpression.Type.GetTypeInfo().IsGenericType
```
smitpatel: you can use `constantExpression.IsEntityQueryable` extension method.

anpete:
@AndriySvyryd Can we go over the changes in person tomorrow with the query team? :trollface:

anpete:
@AndriySvyryd @smitpatel and I think we have fixed the issue (#8105) where non-qsres were showing up in binding, hopefully you can simplify things a lot after this fix goes in.

anpete:
@AndriySvyryd If that works, we speculate that the simplest way to bind entity types will be to maintain a QuerySource -> IEntityType map.

AndriySvyryd:
@anpete The fix worked. I am making the changes and adding tests as discussed.

AndriySvyryd:
Updated. I'll send out another PR with functional tests to keep this PR size down.

anpete:
@AndriySvyryd It seems like we should be able to avoid the change to EntityQueryable because we can just build the QuerySource -> EntityType map as we bind through navs during nav rewrite. Thoughts?

AndriySvyryd:
@anpete In my next PR I found that `EntityEntry.GetDatabaseValues()` for an owned entity requires the query root to be the owned entity type. So we need to put the EntityType in the query expression to be able to generate a different cache key.

anpete:
@AndriySvyryd Can we tweak the query for GetDatabaseValues to use the nav.? E.g. customers.Where(c => c.CustomerID == "ALFKI").Select(c => c.Address)

AndriySvyryd:
@anpete Updated

src/EFCore.InMemory/Query/ExpressionVisitors/Internal/InMemoryEntityQueryableExpressionVisitor.cs
```diff
@@ -51,7 +51,10 @@ protected override Expression VisitEntityQueryable(Type elementType)
         {
             Check.NotNull(elementType, nameof(elementType));
 
-            var entityType = _model.FindEntityType(elementType);
+            if (!QueryModelVisitor.QueryCompilationContext.QuerySourceEntityTypeMapping.TryGetValue(_querySource, out var entityType))
```
anpete: DRY to method on QueryCompilationContext

src/EFCore/Query/QueryCompilationContext.cs
```diff
@@ -107,6 +107,14 @@ public class QueryCompilationContext
         public virtual QuerySourceMapping QuerySourceMapping { get; } = new QuerySourceMapping();
 
         /// <summary>
+        ///     Gets the mapping to query source
+        /// </summary>
+        /// <value>
+        ///     The mapping to query source
+        /// </value>
+        public virtual Dictionary<IQuerySource, IEntityType> QuerySourceEntityTypeMapping { get; } = new Dictionary<IQuerySource, IEntityType>();
```
anpete: Can we keep the dictionary private and add accessor methods.

src/EFCore/Query/ResultOperators/Internal/IncludeResultOperator.cs
```diff
@@ -94,7 +103,51 @@ public virtual void AppendToNavigationPath([NotNull] IReadOnlyList<PropertyInfo>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public override string ToString() 
+        public virtual INavigation[] GetNavigationPath([NotNull] QueryCompilationContext queryCompilationContext)
```
anpete: Should have told you to ignore the old Include code paths. It is all going away :trollface:

anpete:
Looks good, minor comments.

anpete:
cc @maumar for navs review.

PR #8097, opened 2017-04-06
Query: Include: Refactor to use tree data structure inside IncludeCompiler.

anpete:
Move from a list of include paths to a tree representation in preparation for making the compiler fully recursive.


anpete:
ping @smitpatel @maumar

PR #8103, opened 2017-04-07
Fixing extensibility blockers in QueryCompiler/ParameterExtractingExpressionVisitor

anpete:
Making ctor public, de-nesting class and making methods virtual.

### DIFF
--- a/src/EFCore/Query/Internal/EvaluatableExpressionFilter.cs
+++ b/src/EFCore/Query/Internal/EvaluatableExpressionFilter.cs
@@ -1,0 +1,61 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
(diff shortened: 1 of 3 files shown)


src/EFCore/Query/Internal/EvaluatableExpressionFilter.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        public override bool IsEvaluatableMethodCall(MethodCallExpression methodCallExpression)
+        {
+            if (_guidNewGuid.Equals(methodCallExpression.Method)
+                || _randomNext.Contains(methodCallExpression.Method))
+            {
+                return false;
+            }
+
+            return base.IsEvaluatableMethodCall(methodCallExpression);
```
smitpatel: Base is always returning true. Either we can simplify it here. Or if we want to safe-guard against changes in base class in future, we should consider adding call to base in MemberExpression also.

Issue #7429, opened 2017-01-17
StartsWith() filter slow with use of SQL CHARINDEX()


jemiller0:
Describe what is not working as expected.

I have a LINQ query that is using a .StartsWith() search on a string column. I found that it is being translated into the following SQL WHERE clause.

`WHERE [b].[updated_by] LIKE N'jemiller%' + N'%' AND (CHARINDEX(N'jemiller%', [b].[updated_by]) = 1)
`

The full SQL statement is.

```
SELECT [b].[bib_id], [b].[content], [b].[date_created], [b].[created_by], [b].[fast_add], [b].[former_id], [b].[unique_id_prefix], [b].[date_updated], [b].[updated_by], [b].[staff_only], [b].[status], [b].[status_updated_date], [b].[status_updated_by], [b.BibExt].[id], [b.BibExt].[author], [b.BibExt].[creation_time], [b.BibExt].[creation_user_name], [b.BibExt].[last_write_time], [b.BibExt].[last_write_user_name], [b.BibExt].[level], [b.BibExt].[oclc_number], [b.BibExt].[publication_date], [b.BibExt].[subscription_status], [b.BibExt].[title], [b.BibExt].[type]
FROM [ole_ds_bib_t] AS [b]
LEFT JOIN [uc_bib_ext] AS [b.BibExt] ON [b].[bib_id] = [b.BibExt].[id]
WHERE [b].[updated_by] LIKE N'jemiller' + N'%' AND (CHARINDEX(N'jemiller', [b].[updated_by]) = 1)
ORDER BY [b].[date_updated] DESC
OFFSET 0 ROWS FETCH NEXT 100 ROWS ONLY

```
The problem that I'm having is that the query is taking 11 seconds to complete. I'm using SQL Server 2016. The primary table that is being queried has 7.7 million rows and the column that is being filtered on has an index. If I comment out the "AND (CHARINDEX(N'jemiller', [b].[updated_by]) = 1)" part of the WHERE clause so that it's only doing a LIKE, then, the query only takes 1 second. Although, it looked like it took longer the first time through. I think the extra call to CHARINDEX() may be throwing the query optimizer in SQL Server off. If I view the execution plan in SQL Management Studio that it has 9 different steps when CHARINDEX() is included an only 7 without it. I can send the execution plan XML if that would help.

Issue #474 talks about why the CHARINDEX() check is included. Would it be possible to only include that check if the input string includes a LIKE special character such as %, _, [, or ]?

[...]

PR #8108, opened 2017-04-07
Fix to #7429 - StartsWith() filter slow with use of SQL CHARINDEX()

maumar:
Fix is to use more permanent translation:

property.StartsWith(pattern) => property LIKE pattern + '%' AND LEFT(property, LEN(pattern)) = pattern

Also fixed minor issue with null semantics where we would apply C# null semantics for comparisons to null-valued parameters

### DIFF
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -1585,6 +1585,11 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
 
                 return base.VisitBinary(expression);
             }
+
+            protected override Expression VisitExtension(Expression node)
+                => node is NullCompensatedExpression
+                    ? node
+                    : base.VisitExtension(node);
         }
 
         private class BooleanExpressionTranslatingVisitor : RelinqExpressionVisitor
--- a/src/EFCore.Specification.Tests/FunkyDataQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/FunkyDataQueryTestBase.cs
@@ -279,32 +279,32 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         {
             using (var ctx = CreateContext())
             {
-                var result1 = ctx.FunkyCustomers.Where(c => c.FirstName.StartsWith("%B")).Select(c => c.FirstName).ToList();
-                var expected1 = ctx.FunkyCustomers.Select(c => c.FirstName).ToList().Where(c => c != null && c.StartsWith("%B"));
+                var result1 = ctx.FunkyCustomers.Where(c => c.FirstName.EndsWith("%B")).Select(c => c.FirstName).ToList();
+                var expected1 = ctx.FunkyCustomers.Select(c => c.FirstName).ToList().Where(c => c != null && c.EndsWith("%B"));
                 Assert.True(expected1.Count() == result1.Count);
 
-                var result2 = ctx.FunkyCustomers.Where(c => c.FirstName.StartsWith("_r")).Select(c => c.FirstName).ToList();
-                var expected2 = ctx.FunkyCustomers.Select(c => c.FirstName).ToList().Where(c => c != null && c.StartsWith("_r"));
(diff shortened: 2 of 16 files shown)


src/EFCore.Specification.Tests/FunkyDataQueryTestBase.cs
```diff
@@ -279,32 +279,32 @@ public virtual void String_ends_with_on_argument_with_wildcard_constant()
         {
             using (var ctx = CreateContext())
             {
-                var result1 = ctx.FunkyCustomers.Where(c => c.FirstName.StartsWith("%B")).Select(c => c.FirstName).ToList();
-                var expected1 = ctx.FunkyCustomers.Select(c => c.FirstName).ToList().Where(c => c != null && c.StartsWith("%B"));
+                var result1 = ctx.FunkyCustomers.Where(c => c.FirstName.EndsWith("%B")).Select(c => c.FirstName).ToList();
```
maumar: those were suppose to be testing EndsWith, but were testing StartsWith (copy paste error)

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1573,7 +1574,7 @@ protected override Expression VisitBinary(BinaryExpression expression)
                             }
                         }
 
-                        if (parameterValue == null)
+                        if (parameterValue == null && _simplifyNullComparison)
```
smitpatel: Just check here is the `nonParameterExpression` is not `NonNullableExpression`

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1585,6 +1586,23 @@ protected override Expression VisitBinary(BinaryExpression expression)
...
             }
+
+            protected override Expression VisitExtension(Expression node)
+            {
+                if (node is NotNullableExpression)
+                {
+                    var simplifyNullComparison = _simplifyNullComparison;
+                    _simplifyNullComparison = false;
+
+                    var newExpression = base.VisitExtension(node);
```
smitpatel: This is causing ambiguity in the meaning of `NotNullableExpression`.

@anpete @maumar - We need to discuss this in person.
What does `NotNullableExpression` mean?
From name it means that the expression should never be null.
But based on its usage around null semantics, 
Does it mean that we should not look inside `NNE` while applying any kind of null semantics?
What should we interpret for `NNE` == null for? (perhaps false?)

Once we figure out what is true meaning of NNE, we can determine when do we need to wrap something inside it and also if we need to give it different name.

src/EFCore.Sqlite.Core/Query/ExpressionTranslators/Internal/SqliteEndsWithOptimizedTranslator.cs
```diff
@@ -23,26 +23,26 @@ public virtual Expression Translate(MethodCallExpression methodCallExpression)
...
-                        "substr",
-                        // ReSharper disable once PossibleNullReferenceException
-                        methodCallExpression.Object.Type,
-                        new[]
-                        {
-                            methodCallExpression.Object,
-                            Expression.Negate(new SqlFunctionExpression("length", typeof(int), new[] { patternExpression }))
-                        }),
-                    patternExpression);
+                var endsWithExpression = new NotNullableExpression(
```
anpete: Can you factor out a base class for both translators - they both look very similar

src/EFCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerEndsWithOptimizedTranslator.cs
```diff
@@ -23,26 +23,26 @@ public virtual Expression Translate(MethodCallExpression methodCallExpression)
...
-                        "RIGHT",
-                        // ReSharper disable once PossibleNullReferenceException
-                        methodCallExpression.Object.Type,
-                        new[]
-                        {
-                            methodCallExpression.Object,
-                            new SqlFunctionExpression("LEN", typeof(int), new[] { patternExpression })
-                        }),
-                    patternExpression);
+                var endsWithExpression = new NotNullableExpression(
```
anpete: Can you make a base class for both translators?

maumar:

PR #8118, opened 2017-04-10
Disable API check on tests and meta-packages

bricelam:


smitpatel:
EFCore.Specs.Tests?

bricelam:
Thanks. Was running again to make sure I got them all.

Interestingly, it never even tries to run on the following projects...
* dotnet-ef
* ef

smitpatel:
Should it be running on those projects? Is this due to lack of baseline or just no output from tools (perhaps bug in the tool)?

bricelam:
No output. The others gave a warning about no baseline. Possibly a "feature" for tools, but that shouldn't apply to `ef`...

PR #8120, opened 2017-04-10
Test: Update to AssertSql

smitpatel:
Takes params of string
Actual Sql is added as arg by the function
If only 1 string passed then exact match with actual Sql else contains match

@maumar @anpete


ErikEJ:
@smitpatel Any easy way to get rid of the two leading spaces in the VS output window?

smitpatel:
@ErikEJ - @anpete's trick, hold down `Alt` while selecting the output, it will do a box selection,

ErikEJ:
@smitpatel Thanks, that is what I am doing currently! 😄

smitpatel:
@ErikEJ - To avoid the issue of selecting the text, I updated to write new baselines to a file `QueryBaseline.cs` in repo root. I have tried to make it good enough so that it can work for provider writers too. Let me know if there are any issues for you.

ErikEJ:
@smitpatel Great! Thanks, I will give it a try soon

ErikEJ:
@smitpatel Thanks a lot for QueryBaseline.cs !

PR #8122, opened 2017-04-11
Consolidate cached services in DbContext and provide a non-internal way to access them.

AndriySvyryd:


src/EFCore/DbContext.cs
```diff
@@ -95,16 +103,19 @@ public DbContext([NotNull] DbContextOptions options)
 
             _options = options;
 
-            ServiceProviderCache.Instance.GetOrAdd(options, providerRequired: false)
-                .GetRequiredService<IDbSetInitializer>()
-                .InitializeSets(this);
+            _setInitializer = ServiceProviderCache.Instance.GetOrAdd(options, providerRequired: false)
+                .GetRequiredService<IDbSetInitializer>();
+            _setInitializer.InitializeSets(this);
         }
```
ajcvickers: It's actually intentional that the SetInitializer used here is not saved for use later. The reason is that this may not be the service provider that will be used as the internal service provider going forward, because at this time OnConfiguring has not yet been called. Mostly that isn't a problem because set initialization is done by our internal services, but in the case where some of those services are replaced, this could initialize set using non-replaced services. In this rare case if this is a problem for the app, then the app can just not use this mechanism to create DbSet instances, and this code becomes a no-op. However, if this set initializer is then saved and used later for the Set method, then it makes the problem bigger because now an app is using the non-replaced services even when it doesn't need to.

src/EFCore/DbContext.cs
```diff
@@ -44,24 +44,32 @@ namespace Microsoft.EntityFrameworkCore
...
-        private IModel _model;
-        private IInterceptingLogger<LoggerCategory.Update> _updateLogger;
         private IAsyncQueryProvider _queryProvider;
-        private IDbContextPool _dbContextPool;
+        private IModel _model;
 
         private bool _initializing;
         private IServiceScope _serviceScope;
+        private ILogger _logger;
+        private IInterceptingLogger<LoggerCategory.Update> _updateLogger;
```
ajcvickers: Bad rebase resolution?

src/EFCore/DbContext.cs
```diff
@@ -181,6 +192,61 @@ private IDbContextServices InitializeServices()
...
         /// <summary>
         ///     <para>
+        ///         Gets the cached <see cref="IDbSetInitializer" />
+        ///     </para>
+        ///     <para>
+        ///         This property is intended for use by extension methods. It is not intended to be used in
+        ///         application code.
+        ///     </para>
+        /// </summary>
+        IDbSetInitializer IInfrastructure<IDbSetInitializer>.Instance => DbSetInitializer;
```
ajcvickers: Can we document these explicitly with the "it's internal!" message? At least for the ones that are internal service contracts.

src/EFCore/DbContext.cs
AndriySvyryd: ok, I'll add a comment

src/EFCore/DbContext.cs
AndriySvyryd: yes

src/EFCore/DbContext.cs
AndriySvyryd: Sure

src/EFCore/DbContext.cs
```diff
@@ -95,6 +104,14 @@ public DbContext([NotNull] DbContextOptions options)
 
             _options = options;
 
+            // This service is not stored in _setInitializer as this may not be the service provider that will be used
```
ajcvickers: Nice comment. :wink:

src/EFCore/DbContext.cs
```diff
@@ -44,24 +44,33 @@ namespace Microsoft.EntityFrameworkCore
     ///         is discovered by convention, you can override the <see cref="OnModelCreating(ModelBuilder)" /> method.
     ///     </para>
     /// </remarks>
-    public class DbContext : IDisposable, IInfrastructure<IServiceProvider>, IDbContextPoolable
+    public class DbContext :
+        IDisposable,
+        IInfrastructure<IServiceProvider>,
```
ajcvickers: Would it be better to factor out these "cached services" into a different class, and then be an IInfrastructure of that? It would mean that users will not see all these implementations, it could be made obviously and explicitly internal, and we could add new ones without changing DbContext. Thoughts? @divega @anpete @bricelam

src/EFCore/DbContext.cs
anpete: Comments as a Service? :trollface:

src/EFCore/DbContext.cs
divega: What is the exact goal of this PR? I feel like without knowing that it is too easy to come up with silly ideas :smile_cat:

Re having this in a separate class, assuming it matters, would that preserve lazy initialization of the individual services? E.g. would that class be an IInfraestructure for each the service types? Would that mean that you need to jump two hoops to get each service, possibly defeating the purpose of the PR?

Anyway, I agree this is a concern.

AndriySvyryd:
Updated with `DbContextDependencies` to hold all the cached services. Perf is the same, about 10 times faster than resolving the service each time and 3 times faster than caching in a dictionary.

src/EFCore/Infrastructure/DbContextDependencies.cs
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Query.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Infrastructure
+{
```
ajcvickers: Why is this in Infrastructure instead of Internal?

AndriySvyryd: Misclick

Issue #6383, opened 2016-08-22
The wrong type is being selected when an extension method is used in a subquery


rchowe:
### The issue

I have a somewhat complicated expression in a subquery, which I'd like to simplify since I use it a lot, so I defined an extension method on `IEnumerable<MyModelType>` to make it clearer what's happening in the queries. When I try to run the code with the full expression in the subquery, it works, but when I use the extension method, it doesn't.
### Steps to reproduce

I have the following, somewhat contrived example (similar to the MusicStore example):

``` csharp
public class Song
{
    int Id { get; set; }
    public int Order { get; set; }
    public int Rating { get; set; }

    public Album Album { get; set; }
}

public class Album
{
    int Id { get; set; }
    public List<Song> Songs { get; set; }
}
```

In my code, I have this query, which works as expected (gets a list of albums with a song where the highest rated song on the album is in the top three songs).

``` csharp
var goodAlbums = await _db.Albums
    .Include(album => album.Songs)
    .Where(album => album.Songs.OrderByDescending(s => s.Rating).First().Order <= 3)
    .ToListAsync();
Console.WriteLine($"Good Album Count: {goodAlbums.Count()}");
```

This works as expected and generates the following SQL:

``` sql
SELECT "a"."Id"
FROM "Albums" AS "a"
WHERE (
    SELECT "s"."Order"
    FROM "Songs" AS "s"
    WHERE "a"."Id" = "s"."AlbumId"
    ORDER BY "s"."Rating" DESC
    LIMIT 1
) <= 3
ORDER BY "a"."Id"
```

However, I want to simplify getting the highest rated song in the subquery using the following extension method:

``` csharp
public static Song HighestRated(this IEnumerable<Song> songs)
{
    return songs.OrderByDescending(s => s.Rating).First();
}
```

So then when I do this query,

``` csharp
var goodAlbums = await _db.Albums
    .Include(a => a.Songs)
    .Where(a => a.Songs.HighestRated().Order <= 3)
    .ToListAsync();
```

I get this SQL

``` sql
SELECT "s0"."Id", "s0"."AlbumId", "s0"."Order", "s0"."Rating"
FROM "Songs" AS "s0"
WHERE EXISTS (
    SELECT 1
    FROM "Albums" AS "a"
    WHERE ((
        SELECT "s"."Order"
        FROM "Songs" AS "s"
        WHERE "a"."Id" = "s"."AlbumId"
        ORDER BY "s"."Rating" DESC
[...]

PR #8133, opened 2017-04-12
Fix to #6383 - The wrong type is being selected when an extension method is used in a subquery

maumar:
Problem was that when translating collection navigation outside projection we assumed it's always going to be wrapped in a subquery.
This is the case for the most part because in order to be useful inside a query, collection needs to have it's cardinality reduced to single element using a result operator (Count, Any, FirstOrDefault etc), which forces subquery on the collection.
However it is also possible to use client method, which takes collection and returns single element. For this case, subquery won't be generated and we were unable to translate such query.

Fix is to inject subqueries where needed, using same mechanism we use in case of projection. Client method forces materialization on the collection, so it's safe to use.

### DIFF
--- a/src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
@@ -2664,6 +2664,106 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             }
         }
 
+        [ConditionalFact]
+        public virtual void Client_method_on_collection_navigation_in_predicate()
+        {
+            using (var ctx = CreateContext())
+            {
+                var query = from g in ctx.Gears
+                            where g.HasSoulPatch && FavoriteWeapon(g.Weapons).Name == "Marcus' Lancer"
+                            select g.Nickname;
+
+                var result = query.ToList();
+
+                Assert.Equal(1, result.Count);
+                Assert.Equal("Marcus", result[0]);
+            }
+        }
+
+        [ConditionalFact]
+        public virtual void Client_method_on_collection_navigation_in_predicate_accessed_by_ef_property()
+        {
+            using (var ctx = CreateContext())
+            {
+                var query = from g in ctx.Gears
+                            where !g.HasSoulPatch && FavoriteWeapon(EF.Property<List<Weapon>>(g, "Weapons")).Name == "Cole's Gnasher"
+                            select g.Nickname;
+
+                var result = query.ToList();
+
+                Assert.Equal(1, result.Count);
+                Assert.Equal("Cole Train", result[0]);
+            }
+        }
+
+        [ConditionalFact]
+        public virtual void Client_method_on_collection_navigation_in_order_by()
+        {
+            using (var ctx = CreateContext())
+            {
+                var query = from g in ctx.Gears
+                            where !g.HasSoulPatch
+                            orderby FavoriteWeapon(g.Weapons).Name descending
+                            select g.Nickname;
+
+                var result = query.ToList();
+
+                Assert.Equal(3, result.Count);
+                Assert.Equal("Paduk", result[0]);
+                Assert.Equal("Dom", result[1]);
+                Assert.Equal("Cole Train", result[2]);
+            }
+        }
+
+        [ConditionalFact]
+        public virtual void Client_method_on_collection_navigation_in_additional_from_clause()
+        {
+            using (var ctx = CreateContext())
+            {
+                var query = from g in ctx.Gears.OfType<Officer>()
+                            from v in Veterans(g.Reports)
+                            orderby g.Nickname, v.Nickname
+                            select new { g = g.Nickname, v = v.Nickname };
+                            
+                var result = query.ToList();
+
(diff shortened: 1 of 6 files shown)


anpete:
Can you take a look to see if you can leverage re-linqs TransformExpressions instead of manually visiting each expression of each QM element? TransformExpressions is the idiomatic way to visit QM element expressions.

maumar:
@anpete this gets a bit tricky for cases like:

```
from c in ctx.Customer
from o in c.Orders
select whatever
```

where you get collection navigation not wrapped in a subquery. TransformExpressions would have problem distinguishing between those. We could probably consolidate some translations we do in other places, but that makes the fix more involved. I would rather do this as part of refactoring of nav rewrite visitor.

test/EFCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
```diff
@@ -2598,6 +2598,153 @@ public override void Subquery_with_result_operator_is_not_lifted()
...
+            AssertSql(
+                @"",
+                Sql);
+        }
+
+        public override void Client_method_on_collection_navigation_in_outer_join_key()
+        {
+            base.Client_method_on_collection_navigation_in_outer_join_key();
+
+            AssertSql(
```
maumar: need to use Contains here, due to client join ordering discrepancy between .net and core

src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
```diff
@@ -2648,6 +2648,100 @@ orderby g.Rank
...
+
+                Assert.Equal(2, result.Count);
+                Assert.Equal("Baird", result[0].o);
+                Assert.Equal("Baird", result[0].g);
+                Assert.Equal("Marcus", result[1].o);
+                Assert.Equal("Marcus", result[1].g);
+            }
+        }
+
+        private static Weapon FavoriteWeapon(IEnumerable<Weapon> weapons)
```
smitpatel: would static but not extension make a difference?

test/EFCore.SqlServer.FunctionalTests/GearsOfWarQuerySqlServerTest.cs
smitpatel: The brand new `AssertSql` would solve that issue :trollface:

src/EFCore.Specification.Tests/GearsOfWarQueryTestBase.cs
maumar: It shouldn't, extension method looks like any other static in Linq tree.

maumar:

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationSubqueryInjector.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// </summary>
+        public static readonly MethodInfo MaterializeCollectionNavigationMethodInfo
+            = typeof(CollectionNavigationSubqueryInjector).GetTypeInfo()
+                .GetDeclaredMethod(nameof(MaterializeCollectionNavigation));
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitMember(MemberExpression node)
```
smitpatel: `memberExpression` instead of node

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationSubqueryInjector.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            Check.NotNull(node, nameof(node));
+
+            var newMemberExpression = default(Expression);
+            if (_shouldInject)
+            {
+                newMemberExpression = _queryModelVisitor.BindNavigationPathPropertyExpression(
+                node,
```
smitpatel: indent

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationSubqueryInjector.cs
```diff
@@ -0,0 +1,133 @@
...
+                    });
+            }
+
+            return newMethodCallExpression ?? base.VisitMethodCall(node);
+        }
+
+        private static Expression InjectSubquery(Expression expression, INavigation collectionNavigation)
+        {
+            var targetType = collectionNavigation.GetTargetType().ClrType;
+            var mainFromClause = new MainFromClause(targetType.Name.Substring(0, 1).ToLowerInvariant(), targetType, expression);
```
smitpatel: indexer with 0

maumar: that returns char

maumar:
new version up

maumar:
ping @anpete @smitpatel

maumar:
rebased @anpete @smitpatel

Issue #7217, opened 2016-12-08
Improve experiece for filtering log data


rowanmiller:
At the moment you need to filter by category name, which has some issues:
 * It's hard to work out the full set of categories you want
 * You are potentially referencing internal types
 * Type names can change (i.e. you want `Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommandBuilderFactory` in 1.0 and `Microsoft.EntityFrameworkCore.Storage.IRelationalCommandBuilderFactory` in 1.1)

Even once you do this, it is hard to filter to the events that you want. Here is an attempt to filter to commands being executed, but EventIds are not globally unique... so there is no guarantee that I just get RelationalEventId.ExecutedCommand events. Even in this example, where I am just logging for one component, there is no guarantee that the command builder factory won't log something related to the EventId in core that collides with the id of RelationalEventId.ExecutedCommand.

The types we pass to state also make is pretty cumbersome to get the data I want (command text).

```c#
    public class SqlLoggerProvider : ILoggerProvider
    {
        public ILogger CreateLogger(string categoryName)
        {
            if (categoryName == typeof(Microsoft.EntityFrameworkCore.Storage.IRelationalCommandBuilderFactory).FullName)
            {
                return new SqlLogger(categoryName);
            }

            return new NullLogger();
        }

        public void Dispose()
        { }

        private class SqlLogger : ILogger
        {
            private string _test;

            public SqlLogger(string test)
            {
                _test = test;
            }

            public bool IsEnabled(LogLevel logLevel)
            {
                return true;
            }

            public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
            {
                if (eventId.Id == (int)RelationalEventId.ExecutedCommand)
                {
                    var data = state as IEnumerable<KeyValuePair<string, object>>;
                    if (data != null)
                    {
                        var commandText = data.Single(p => p.Key == "CommandText").Value;
                        Console.WriteLine(commandText);
                    }
                }
            }

            public IDisposable BeginScope<TState>(TState state)
            {
                return null;
            }
        }

        private class NullLogger : ILogger
        {
            public bool IsEnabled(LogLevel logLevel)
            {
                return false;
            }

            public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
            { }

            public IDisposable BeginScope<TState>(TState state)
            {
                return null;
[...]

PR #8136, opened 2017-04-12
Standardize and consolidate logging

ajcvickers:
Issues #7217 #218

This change:
* Consolidates interception and sensitive logging into one service and uses that service everywhere in the EF code.
* Adds a LoggerCategory class which contains nested classes that define EF logging categories.
* Used together, this means any EF service can depend on `IInterceptingLogger<TLoggingCategory>` and have D.I. create the correct service using its open generics feature. The TLoggingCategory is constrained so that only types specifically defined as LoggingCategories can be used. For example, `IInterceptingLogger<LoggingCategory.Database.Sql>`.
* Application code can get logging categories using, for example, `LoggingCategory.Database.Sql.Name`. This means that any application can immediately see in code what logging categories there are to do appropriate filtering.
* All services depend on the specific generic `IInterceptingLogger<TLoggingCategory>` that they will use--never ILogger. This makes it very clear which category is being logged to and makes it safe to share loggers where safe, and impossible where the categories should be different.


src/EFCore/DbContext.cs
```diff
@@ -76,7 +75,7 @@ protected DbContext()
             : this(new DbContextOptions<DbContext>())
         {
         }
-
+
```
lajones: Remove extra whitespace here and a couple of places below.

ajcvickers:
Merged

PR #8143, opened 2017-04-12
Introduce ShapedQueryExpression

tuespetre:
This PR introduces the `ShapedQueryExpression`, an extension expression that represents calls to the `ShapedQueryMethod` and `DefaultIfEmptyShapedQueryMethod`. By having this expression carry its associated `SelectExpression`, `Shaper`, and `ShaperCommandContext`, it can reduce down to the appropriate method call expression when necessary. This simplifies code related to query flattening and I think taking a similar approach to other concepts (like an `InjectParametersExpression`, for instance) would help bring more clarity to the processing of the query model and help keep more of the crucial data and behaviors inside of (and accessible from) the expression tree itself.

Related: #7520

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
@@ -84,7 +84,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 QueryModelVisitor.RegisterSubQueryVisitor(_querySource, queryModelVisitor);
             }
 
-            return queryModelVisitor.Expression;
+            return new NonComposableQueryExpression(queryModelVisitor.Expression);
         }
 
         /// <summary>
@@ -168,6 +168,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
             Func<IQuerySqlGenerator> querySqlGeneratorFunc = selectExpression.CreateDefaultQuerySqlGenerator;
 
+            var useQueryComposition = true;
+
             if (fromSqlAnnotation == null)
             {
                 selectExpression.AddTable(
@@ -188,7 +190,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
                 var trimmedSql = fromSqlAnnotation.Sql.TrimStart('\r', '\n', '\t', ' ');
 
-                var useQueryComposition
+                useQueryComposition
                     = trimmedSql.StartsWith("SELECT ", StringComparison.OrdinalIgnoreCase)
                       || trimmedSql.StartsWith("SELECT" + Environment.NewLine, StringComparison.OrdinalIgnoreCase)
                       || trimmedSql.StartsWith("SELECT\t", StringComparison.OrdinalIgnoreCase);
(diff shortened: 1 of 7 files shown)


src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
```diff
@@ -84,7 +84,9 @@ protected override Expression VisitSubQuery(SubQueryExpression expression)
                 QueryModelVisitor.RegisterSubQueryVisitor(_querySource, queryModelVisitor);
             }
 
-            return queryModelVisitor.Expression;
+            return queryModelVisitor.Expression is ShapedQueryExpression shapedQueryExpression
```
anpete: Do we need to reduce here? It would be nice if reduction only happens during final compile.

src/EFCore.Relational/Query/Expressions/ShapedQueryExpression.cs
```diff
@@ -0,0 +1,148 @@
...
+        public ShapedQueryExpression(
+            [NotNull] RelationalQueryCompilationContext queryCompilationContext,
+            [NotNull] SelectExpression selectExpression,
+            [NotNull] Shaper shaper,
+            [NotNull] ShaperCommandContext shaperCommandContext)
+            : this(Check.NotNull(queryCompilationContext, nameof(queryCompilationContext)),
+                  Check.NotNull(selectExpression, nameof(selectExpression)),
+                  Check.NotNull(shaper, nameof(shaper)),
+                  Check.NotNull(shaperCommandContext, nameof(shaperCommandContext)),
+                  false)
```
smitpatel: used named parameter for bool types

src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
tuespetre: To avoid the need to reduce here we could use a lightweight, reducible marker expression (`SubqueryBoundaryExpression`?) so the query model visitor doesn't try to flatten when it shouldn't. I was thinking about that for `InjectParameters` originally, in order to combine and lift multiple calls to `InjectParameters` up to a spot higher in the expression tree while respecting subquery boundaries.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1279,29 +1171,19 @@ var compositeShaper
...
-            var innerSelectExpression = TryGetQuery(joinClause);
-
-            if (outerSelectExpression == null || innerSelectExpression == null)
+                || outerShapedQuery == null
+                || innerShapedQuery == null)
             {
                 return false;
             }
+            
+            var outerSelectExpression = outerShapedQuery.SelectExpression;
```
smitpatel: Doesn't shaper knows about the SelectExpression it is going to shape already?

src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
anpete: It's fine to reduce here if we have to, was just trying to understand why. So, it's because we otherwise incorrectly lift the subquery?

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
tuespetre: I'm not sure what you mean.

src/EFCore.Relational/Query/ExpressionVisitors/RelationalEntityQueryableExpressionVisitor.cs
tuespetre: That is correct

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
smitpatel: Shaper has a way to get SelectExpression but it is pretty indirect. It is fine to use SelectExpression here though we can avoid adding a property in ShapedQueryExpression if we retrieve SelectExpression as we used to do it in current codebase.

tuespetre: I don't see any way to get a `SelectExpression` from a `Shaper` -- the `ShaperCommandContext` does hold a delegate that closes over the `SelectExpression` but I don't see a way to get at the `SelectExpression` itself.

I think the property is good to have, because the data relevant to the expression is right there in the expression, and you don't have to ask some query model visitor to find it for you. It seems to me that it has been too easy to cram data points into the `RelationalQueryModelVisitor` (`RequiresClient____` properties, `QueriesBySource`, `_subQueryModelVisitorsBySource`, `_injectedParameters`, etc.), which has resulted in a number of methods that don't really have anything to do with visiting a query model but instead give other expression visitors a means of interacting with that data (`TryGetQuery`, `BindMemberExpression`, `LiftInjectedParameters`, etc.)

With some more careful consideration we could figure out how to represent that data right within the expression tree that we are building. In this case, it would be a step away from `RelationalQueryModelVisitor.TryGetQuery` and `RelationalQueryCompilationContext.FindSelectExpression`.

tuespetre:
:up::date: 

@anpete, I did go ahead and strawman up a `NonComposableQueryExpression` for this PR. The `NonComposableQueryExpression` is essentially an expression representation of `RequiresClientEval` that can both aid in pattern matching for flattening purposes and function as a delimiter when traversing the expression tree.

I also tweaked the `Flatten` methods to show how `RequiresClientEval`, `RequiresClientJoin`, and `RequiresClientSelectMany` don't need to be checked for flattening because we can instead rely on matching against `ShapedQueryExpression`s.

smitpatel:
@tuespetre - I hope you are doing well. Its been a while we received response on PRs from you. If you are working on them and intend to update them soon then let us know. If you are busy at this time, then we would like to close pending PRs. You can always re-open them in future when you get time and start working on them.

smitpatel:
@anpete - Do you want to try rebasing it & merging it?

Issue #8077, opened 2017-04-05
Scaffold-DbContext with create default


reader-man:
Describe what is not working as expected.
When running Scaffold-DbContext
  The Generation of Fluent API for the default value in Context.OnModelCreating(ModelBuilder modelBuilder), returns SQL Server comments which affects the compilation of the Project in VS.

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
Compilation Error in intellisense.
```

### Steps to reproduce
make a new database in sql server, and run the file "sqlSampleTableToProveGenerationProblem.sql" and don't delete the first empty lines:

https://goo.gl/w1Fb5v

then generate the POCO Entities + the Context, and you will see these problematic lines:
```
                entity.Property(e => e.AnythingCol)
                    .HasColumnName("anythingCol")
                    .HasColumnType("TypeBoolean_Dom")
                    .HasDefaultValueSql("


/*==============================================================*/
/* Default: D_0                                                 */
/*==============================================================*/
create default D_0
    as 0
");
            });
```

Partial code listings, or multiple fragments of code, will slow down our response or cause us to push the issue back to you to provide code to reproduce the issue.

```c#
<code listing>
```

### Further technical details
EF Core version: 1.1
Database Provider: Microsoft.EntityFrameworkCore.SqlServer
Operating system: Win 10
IDE: Visual Studio 2017

bricelam:
Note to fixer, Migrations [handles this](https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore.Design/Migrations/Design/CSharpHelper.cs#L292-L295) by generating a verbatim string literals for multi-line strings.

jnegron357:
I'm getting the same issue except my output is just a bunch of empty lines:
`.HasDefaultValueSql("create default [DEFgetdate] as getdate()






");
`
When I look at the definition for the custom default DEFgetdate I see that it also has several lines of empty space, could it be the EF core is picking it up from there? If so then maybe it just needs a Trim() function call at the time of generation of code.

We are using Sql Server 2008 R2 (SP2) - 10.50.4042.0 (X64) 

EF Core version: 1.1.1
Database Provider: Microsoft.EntityFrameworkCore.SqlServer
Operating system: Win 7
IDE: Visual Studio Professional 2017
Version 15.1 (26403.0) Release

lajones:
[...]

PR #8145, opened 2017-04-12
Update Scaffolding to escape a string containing newlines as a verbatim string

lajones:
Fix for #8077.


lajones:
Fix checked in with cdd23ffa5.

Issue #4311, opened 2016-01-14
Query: InMemory: Incorrect QueryModel coming from ReLinq for groupjoin scenarios


anpete:
Started failing after 94a2d12b

anpete:
This looks to be related to a re-linq issue. Following up with those folks.

anpete:
Related re-linq issue: https://www.re-motion.org/jira/browse/RMLNQ-105

azabluda:
It also fails if I replace `Take` with `OfType`
```c#
public virtual void GroupJoin_DefaultIfEmpty3()
{
    AssertQuery<Customer, Order>((cs, os) =>
        from c in cs.OfType<Customer>() // .Take(1)
        join o in os on c.CustomerID equals o.CustomerID into orders
        from o in orders.DefaultIfEmpty()
        select o);
}
```
Do you think it makes sense to report my case separately or is it basically covered by #4311 / https://www.re-motion.org/jira/browse/RMLNQ-105 ?

ajcvickers:
Based on comment on #7770 it looks like there is a new version of re-linq that we can use.

ajcvickers:
@anpete Can you try the new version?

smitpatel:
Another skipped test - `GroupJoin_customers_orders_count_preserves_ordering`

smitpatel:
@MichaelKetting - I tested Remotion.Linq version 2.2.0-alpha-002 with current codebase of EF. Everything looks fine on our end. All existing test including the above failing tests passes successfully with pre-release version.

MichaelKetting:
@smitpatel Thanks for testing this! I've got one or two more things I was hoping to get done before releasing v2.2 (https://www.re-motion.org/jira/browse/RMLNQ-25 and https://www.re-motion.org/jira/browse/RMLNQ-111). Am not sure yet regarding my available time but I'd like to wait one or two weekends before making that descision and see how it goes.

anpete:
@MichaelKetting Thanks, btw we are hitting RMLNQ-111 and so it would great to get that in.

MichaelKetting:
@anpete Thanks for letting me know re RMLNQ-111. Then that's defintiely going into the upcoming release.

smitpatel:
@MichaelKetting - Hey any update/estimate on when version Re-Linq 2.2 will come out?

MichaelKetting:
[...]

PR #8161, opened 2017-04-13
[TINY] adding test for issue #7654

maumar:


test/EFCore.InMemory.FunctionalTests/ComplexNavigationsQueryInMemoryTest.cs
```diff
@@ -11,5 +11,10 @@ public ComplexNavigationsQueryInMemoryTest(ComplexNavigationsQueryInMemoryFixtur
             : base(fixture)
         {
         }
+
```
smitpatel: Instead of commenting out this way, put `ConditionalFact(Skip = "See issue#7654")` on the method so it shows up in test logs that test is being skipped.

maumar:
it is indeed #4311

PR #8167, opened 2017-04-14
Disable flaky tests on CI

smitpatel:
@pranavkm

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.Specification.Tests/DatabindingTestBase.cs
+++ b/src/Microsoft.EntityFrameworkCore.Specification.Tests/DatabindingTestBase.cs
@@ -405,7 +405,7 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


src/Microsoft.EntityFrameworkCore.Specification.Tests/DatabindingTestBase.cs
```diff
@@ -405,7 +405,7 @@ public virtual void Load_executes_query_on_DbQuery()
         }
 
         [Fact]
-        public void LocalView_is_initialized_with_entities_from_the_context()
+        public virtual void LocalView_is_initialized_with_entities_from_the_context()
```
smitpatel: @ajcvickers - This is change in Specifications project which we ship for provider writers. We need to add virtual to override the test and disable it in SqlServer. Is this breaking change?

ajcvickers: No, breaking the tests is fine. Provider writers can react to it in their tests. Doesn't impact product code.

PR #8168, opened 2017-04-14
Query: Remove unnecessary ExplodeStarProjection

### LINKED ISSUE #8131: Undo Distinct/Offset/Limit ExplodeStarProjection change caused by Include Compiler binding
Tracking bug for @smitpatel

smitpatel:
Handle StarProjection expression better in PushDownSubquery
SelectExpression.BindSubqueryProjectionIndexToSelectExpression method which binds AnonymousObject index with correct projected expression

resolves #8131


smitpatel:
@anpete - updated to simplify `ExplodeStarProjection` (since we already have list of lifted projections) & renamed methods

Issue #8176, opened 2017-04-14
Query: projecting stringColumn.Length may cause invalid cast exception, if the stringColumn is of type (n)varchar(max) on SqlServer


maumar:
This happens because LEN returns long for varchar(max) and nvarchar(max) columns, but we are expecting int.

maumar:
Fixed in e98a33dfdfd776a7477ba0079804e49ba004b502

PR #8179, opened 2017-04-14
[TINY] Fix to #8176 - Query: projecting stringColumn.Length may cause invalid cast exception, if the stringColumn is of type (n)varchar(max) on SqlServer

maumar:
Problem was that for (n)varchar(max) columns, LEN function returns long, and we are expecting int. Fix is to always explicitly cast the result of LEN to int.


smitpatel:
Fixes #8176

PR #8180, opened 2017-04-15
improvement to ExpressionPrinter for null conditional expression

maumar:
Before we relied on NullConditionalExpression.ToString(), however this way we would not highlight dangling qsres that were inside the NCE. Also DRYing up printing of arguments.


maumar:
ping @smitpatel

Issue #8060, opened 2017-04-03
Query: Make our reflection info selection logic deterministic


divega:
We often need to query reflection to find a specific `MethodInfo`, e.g. to generate the right expressions or to identify expressions that we can translate. 

Currently we have places that uses patterns that are not resilient to the introduction of new overloads. E.g. issue #8021 is about the introduction of new overloads of `String.Replace()` breaking an assumption in our code that only one overload would matched certain conditions.

In other words, we should use reflection in a way that doesn't break when non-breaking changes (such as introducing new overloads) happen :smile:

The general approach is to fully specify method parameter types when we query. We can also look at using `GetMethod(string, Type[])` when we can (as they do it in https://github.com/dotnet/corefx/blob/master/src/System.Linq.Expressions/src/System/Linq/Expressions/Common/CachedReflectionInfo.cs for `Queryable`) since it results in simpler looking code when you are fully specifying parameter types.

maumar:
fixed in 11dd5f60028a70329946dfe07409bbb4094c95a3

PR #8195, opened 2017-04-17
Fix to #8060 - Query: Make our reflection info selection logic deterministic

maumar:
Problem was that the way we extract MethodInfos for translators was always deterministic and could lead to errors if the new overloads of those methods were to be added.

Fix is to use pattern in which we fully specify the name and types of all arguments to the method - this is guaranteed to be unique.

### DIFF
--- a/src/EFCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerMathTranslator.cs
+++ b/src/EFCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerMathTranslator.cs
@@ -17,62 +17,84 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
     /// </summary>
     public class SqlServerMathTranslator : IMethodCallTranslator
     {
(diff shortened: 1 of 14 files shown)


src/EFCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerMathTranslator.cs
```diff
@@ -48,15 +73,15 @@ public virtual Expression Translate(MethodCallExpression methodCallExpression)
             var method = methodCallExpression.Method;
             if (method.DeclaringType == typeof(Math))
```
smitpatel: we don't need this condition anymore.

smitpatel:
Does this cover all the method infos?

maumar:
@smitpatel all except ConvertTo (40+ overloads). For those we keep the list of supported method names and the list of supported types, and make sure that the methods translated have only one argument - so it's also deterministic

maumar:
@smitpatel new version up

ErikEJ:
Providers beware?

PR #8198, opened 2017-04-17
Query: Allow translations through DTO objects

### LINKED ISSUE #7234: Query: Translate Distinct to SQL when applied after custom projection
I looked up for this problem and didn't find anything. Sorry if I'm making a dup or I messed up something.

This
```c#
context.Blogs
    .Select(t => new BlogDto
    {
        Url = t.Url
    })
    .Distinct()
    .ToList();
```
resulted in the following query:
```sql
SELECT [t].[Url]
FROM [Blogs] AS [t]
```
and I expected:
```sql
SELECT DISTINCT [t].[Url]
FROM [Blogs] AS [t]   
```

So, now I'm using the following workaround:

```c#
context.Blogs
    .Select(t => new
    {
        Url = t.Url
    })
    .Distinct()
    .Select(t => new BlogDto
    {
        Url = t.Url
    })
    .ToList();
```

which results in:

```sql
SELECT [t1].[Url]
FROM (
    SELECT DISTINCT [t0].[Url]
    FROM [Blogs] AS [t0]
) AS [t1]
```

### Steps to reproduce

```c#
    public class BlogDto
    {
    }

    public class BlogContext : DbContext
    {

        {
            optionsBuilder.UseSqlServer(@"Data Source=.;Initial Catalog=Blogging;Integrated Security=True;MultipleActiveResultSets=True");
            //optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=CollectionsTest;Trusted_Connection=True;ConnectRetryCount=0");
        }
    }

    public class Program
    {
        private static void Main()
        {
            using (var context = new BlogContext())
            {
                var results = context.Blogs
                    .Select(t => new BlogDto
                    {
                        Url = t.Url
                    })
                    .Distinct()
                    .ToList();
            }
        }
    }
```

Just to be clear, the following works as expected:
```c#
context.Blogs
    .Select(t => new
    {
        Url = t.Url
    })
    .Distinct()
    .ToList();
``` 
produces:

```sql
SELECT DISTINCT [t].[Url]
FROM [Blogs] AS [t]
```

I have one simple question regarding terminology. When you say 'projection' ('projecting' to some object different from the model), does that include anonymous objects? If so, the title of this issue should be "Projecting to a specific object(class) with Distinct is not working as expected"

Thank you!

### Further technical details
EF Core version: 1.1.0
Operating system: Win 8.1 64bit
IDE: Visual Studio 2015 Update 3

smitpatel:
DTO which are projected out, (from any level) are materialized correctly with typed projection shaper
Store member bindings for MemberInitExpression to allow binding after DTO projection

Resolves #7234

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
@@ -8,7 +8,6 @@ using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Query.Expressions;
-using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
 using Microsoft.EntityFrameworkCore.Query.Internal;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Utilities;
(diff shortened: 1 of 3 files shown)


src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -53,6 +52,45 @@ private new RelationalQueryModelVisitor QueryModelVisitor
...
+        /// <returns>
+        ///     An Expression corresponding to the translated member init.
+        /// </returns>
+        protected override Expression VisitMemberInit(MemberInitExpression memberInitExpression)
+        {
+            var newMemberInitExpression = base.VisitMemberInit(memberInitExpression);
+
+            var selectExpression = QueryModelVisitor.TryGetQuery(_querySource);
+
+            if (selectExpression != null)
```
anpete: DRY this with VisitNew

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -53,6 +52,45 @@ private new RelationalQueryModelVisitor QueryModelVisitor
...
+            {
+                foreach (var sourceBinding in memberInitExpression.Bindings)
+                {
+                    if (sourceBinding is MemberAssignment memberAssignment)
+                    {
+                        var sourceExpression = memberAssignment.Expression;
+
+                        if (_sourceExpressionProjectionMapping.ContainsKey(sourceExpression))
```
anpete: Why not TryGetValue?

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
smitpatel: based on the complex logic of finding member id in NewExpression (since newExpression.Members can be null), it provides no value to DRY it.

smitpatel:
Updated.

Issue #6658, opened 2016-10-01
Query: Entity Framework core 1 Group By after Union throw “must be reducible node” Exception


MohammadAkbari:
I need unoin two query and after union Group By that, but throw "must be reducible node" Exception. in following sample code are:

```
public class ForumPost
{
    public int ForumPostId { get; set; }
    public string Message { get; set; }
    public bool Deleted { get; set; }
    public int ForumTopicId { get; set; }
    public virtual ForumTopic ForumTopic { get; set; }
}

var query1 = _dbContext.ForumPosts
    .Where(e => e.Deleted == true);

var query2 = _dbContext.ForumPosts
    .Where(e => e.Deleted == false);

var query =  query1.Union(query2);

var list = query.GroupBy(e => e.ForumTopicId)
    .Select(group => new
    {
        group.Key,
        Total = group.Count()
    }).Take(10)
    .ToList();
```

Does anyone faced with this problem?

adrien-constant:
Is there an ETA on this? I have the same exception, and I have to call ToArray() before filtering, which obviously is bad for performances.

rowanmiller:
It's on our list to investigate for the next release (1.2.0). Once we understand the work required, we'll have a better idea of where this fits.

maumar:
This still reproduces on current bits. We generate the following query plan:

```
(QueryContext queryContext) => IEnumerable<<>f__AnonymousType0<int, int>> _InterceptExceptions(
    source: IEnumerable<<>f__AnonymousType0<int, int>> _Select(
        source: IEnumerable<IGrouping<int, ForumPost>> Take(
            source: IEnumerable<IGrouping<int, ForumPost>> _GroupBy(
                source: IEnumerable<ForumPost> Union(
                    first: IEnumerable<ForumPost> _ShapedQuery(
                        queryContext: queryContext, 
                        shaperCommandContext: SelectExpression: 
                            SELECT [e1].[ForumPostId], [e1].[Deleted], [e1].[ForumTopicId], [e1].[Message]
                            FROM [ForumPosts] AS [e1]
                            WHERE [e1].[Deleted] = 1, 
                        shaper: BufferedEntityShaper<ForumPost>), 
                    second: IEnumerable<ForumPost> _ShapedQuery(
                        queryContext: queryContext, 
                        shaperCommandContext: SelectExpression: 
[...]

PR #8213, opened 2017-04-18
Query: Update materialization marking for set operations

smitpatel:
Add test for set operation followed group by not translating correctly Issue #6658

**This PR does not fix the issue**

We were checking for type matching when marking set operations for materialization. That should not be required as compiler would make sure that types are compatible.

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
@@ -1,8 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/EFCore/Query/ExpressionVisitors/Internal/RequiresMaterializationExpressionVisitor.cs
```diff
@@ -1,8 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
smitpatel: @ajcvickers @Eilon - Just notifying that this file did not license not earlier (in current codebase) if there are any consequences of that.

ajcvickers: Should be fine. Thanks for adding it.

Eilon: If there's ever a missing license in a file, no worries, just add it in.

Issue #7918, opened 2017-03-16
Scaffold-DbContext Computed Column


pablotdv:
When I run the "Scaffold-DbContext" command on an existing database. I'm having problems with the calculated columns.

The script of my Table

```
CREATE TABLE [dbo].[Caixas](
	[CaixaId] [uniqueidentifier] NOT NULL,
	[Percentual] [decimal](18, 2) NOT NULL,
	[Quantidade] [decimal](18, 2) NOT NULL,
	[Valor] [decimal](18, 2) NOT NULL,
	[ValorDesconto]  AS (([Valor]*[Quantidade])/([Percentual]/(100))),
 CONSTRAINT [PK_Caixas] PRIMARY KEY CLUSTERED 
(
	[CaixaId] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
) ON [PRIMARY]

```

After running the command 

`Scaffold-DbContext "Server=.\SQLExpress;Database=TestComputedColumn;Trusted_Connection=True;" Microsoft.EntityFrameworkCore.SqlServer -OutputDir Models2`

My OnModelCreating was as follows:
```
protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Caixas>(entity =>
            {
                entity.HasKey(e => e.CaixaId)
                    .HasName("PK_Caixas");

                entity.Property(e => e.CaixaId).ValueGeneratedNever();

                entity.Property(e => e.Percentual).HasColumnType("decimal");

                entity.Property(e => e.Quantidade).HasColumnType("decimal");

                entity.Property(e => e.Valor).HasColumnType("decimal");

                entity.Property(e => e.ValorDesconto)
                    .HasColumnType("decimal")
                    .HasComputedColumnSql("[Valor]*[Quantidade])/([Percentual]/(100)")
                    .ValueGeneratedOnAddOrUpdate();
            });
        }
```
Note the attribute HasComputedColumnSql, where I have some missing parentheses.

The correct is  HasComputedColumnSql("**(**[Valor]*[Quantidade])/([Percentual]/(100)")

### Further technical details
EF Core version: netcoreapp1.1
Database Provider: (e.g. Microsoft.EntityFrameworkCore.SqlServer)
Operating system: 
[...]

PR #8217, opened 2017-04-19
Fixing problem interpreting default and computed values

lajones:
Fix for #7918.

Stop attempting to interpret the default/computed values stored by SQL Server. Just report them exactly as they are stored (including any parentheses).

Note: new entries needed in provider design-time services (@ajcvickers please see the changes I had to put into `SqlServerDesignTimeServices` and `SqliteDesignTimeServices`. I think these were missed as part of #8316).

### DIFF
--- a/src/EFCore.SqlServer.Design/Infrastructure/SqlServerDesignEventId.cs
+++ b/src/EFCore.SqlServer.Design/Infrastructure/SqlServerDesignEventId.cs
@@ -61,11 +61,13 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
 
         /// <summary>
         ///     Cannot interpret default value.
+        ///     Note: no longer used.
         /// </summary>
         CannotInterpretDefaultValueWarning,
 
         /// <summary>
         ///     Cannot interpret computed value.
+        ///     Note: no longer used.
         /// </summary>
         CannotInterpretComputedValueWarning
     }
--- a/src/EFCore.SqlServer.Design/Internal/SqlServerDesignTimeServices.cs
+++ b/src/EFCore.SqlServer.Design/Internal/SqlServerDesignTimeServices.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.EntityFrameworkCore.Infrastructure;
-using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Storage;
@@ -23,8 +22,6 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
         /// </summary>
         public virtual void ConfigureDesignTimeServices(IServiceCollection serviceCollection)
             => serviceCollection
-                .AddSingleton<ILoggingOptions, LoggingOptions>()
-                .AddSingleton<IInterceptingLogger<LoggerCategory.Scaffolding>, InterceptingLogger<LoggerCategory.Scaffolding>>()
                 .AddSingleton<IScaffoldingModelFactory, SqlServerScaffoldingModelFactory>()
                 .AddSingleton<IRelationalAnnotationProvider, SqlServerAnnotationProvider>()
                 .AddSingleton<IRelationalTypeMapper, SqlServerTypeMapper>()
--- a/src/EFCore.SqlServer.Design/Internal/SqlServerDesignTimeServices.cs
+++ b/src/EFCore.SqlServer.Design/Internal/SqlServerDesignTimeServices.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Internal;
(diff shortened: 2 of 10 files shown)


src/EFCore.SqlServer.Design/Infrastructure/SqlServerDesignEventId.cs
```diff
@@ -61,11 +61,13 @@ public enum SqlServerDesignEventId
 
         /// <summary>
         ///     Cannot interpret default value.
+        ///     Note: no longer used.
```
lajones: Did not want to delete this and the below as they are part of the public API.

src/EFCore.Sqlite.Design/Internal/SqliteDesignTimeServices.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.EntityFrameworkCore.Internal;
```
lajones: I will put this in correct alphabetical order.

src/EFCore.SqlServer.Design/Infrastructure/SqlServerDesignEventId.cs
ajcvickers: You can delete them or not--all this is changing for 2.0 anyway. Fewer merge conflicts for me if you don't delete them now, and I will do it as part of my change.

src/EFCore.SqlServer.Design/Internal/SqlServerDesignTimeServices.cs
```diff
@@ -22,6 +23,8 @@ public class SqlServerDesignTimeServices : IDesignTimeServices
         /// </summary>
         public virtual void ConfigureDesignTimeServices(IServiceCollection serviceCollection)
             => serviceCollection
+                .AddSingleton<ILoggingOptions, LoggingOptions>()
+                .AddSingleton<IInterceptingLogger<LoggerCategory.Scaffolding>, InterceptingLogger<LoggerCategory.Scaffolding>>()
```
ajcvickers: Does this not work if registered as the open generic?

src/EFCore.SqlServer.Design/Internal/SqlServerDesignTimeServices.cs
```diff
@@ -22,6 +23,8 @@ public class SqlServerDesignTimeServices : IDesignTimeServices
         /// </summary>
         public virtual void ConfigureDesignTimeServices(IServiceCollection serviceCollection)
```
ajcvickers: @bricelam From what I can tell, in DesignTimeServicesBuilder, ConfigureProviderServices, which indirectly calls this, is called after ConfigureContextServices. This means that any service that has been obtained from the context will then be overwritten if GetProviderDesignTimeServices also tries to add it. So, for example, in this case ILoggingOptions from the context will be overwritten.

Is this intentional? Would it be better for these services to be TryAdded?

lajones: @ajcvickers @bricelam Probably still need the answer to the above, but note: Scaffolding does not call `ConfigureContextServices()`. That only happens if you call `DesignTimeServicesBuilder.Build(DbContext)`. Scaffolding does not have an existing `DbContext` so it calls `DesignTimeServicesBuilder.Build(string)`.

bricelam: The order is as follows.
* ConfigureServices
* ConfigureContextServices (if using a context)
* ConfigureProviderServices
* ConfigureUserServices

Given that, I think `ILoggingOptions` and `IInterceptingLogger<>` need to be added to `DesignTimeServicesBuilder.ConfigureServices` (and overridden by `ConfigureContextServices` when possible).

The design is intentional. If a provider needs to replace a default service at design-time, they can. Likewise if a user needs to override a provider service, they can.

lajones: @bricelam Would it be better to add them to [AddScaffolding()](https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore.Design/Scaffolding/Internal/ScaffoldingServiceCollectionExtensions.cs#L23)?

Also how does the call to [AddLogging()](https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore.Design/Design/Internal/DesignTimeServicesBuilder.cs#L80) in `DesignTimeServicesBuilder.ConfigureServices()` interact with this?

ajcvickers: @bricelam Thanks, that makes sense. @lajones Yes, they should be added to AddScaffolding. With regard to AddLogging, it adds external logging dependencies. These are our own logging abstractions.

src/EFCore.SqlServer.Design/Internal/SqlServerDesignTimeServices.cs
lajones: I was able to get it to work with the generic instead. So I did that.

src/EFCore.SqlServer.Design/Internal/SqlServerDesignTimeServices.cs
lajones: Have checked in new commit implementing the above. Could one of you sign off please?

lajones:
Checked in with commit 1d86186cd.

Issue #7207, opened 2016-12-07
Query: compilation error for queries with navigations that also compare same navigation to null


alaatm:
```
Exception message:
System.ArgumentException: must be reducible node

Stack trace:
   at System.Linq.Expressions.Expression.ReduceAndCheck()
   at System.Linq.Expressions.Expression.ReduceExtensions()
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExtensionExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteVariableAssignment(BinaryExpression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteAssignBinaryExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteBlockExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExtensionExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteUnaryExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteLogicalBinaryExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteLogicalBinaryExpression(Expression expr, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteExpression(Expression node, Stack stack)
   at System.Linq.Expressions.Compiler.StackSpiller.ChildRewriter.Add(Expression node)
   at System.Linq.Expressions.Compiler.StackSpiller.RewriteBinaryExpression(Expression expr, Stack stack)
[...]

PR #8219, opened 2017-04-19
Fix to #7207 - Query: compilation error for queries with navigations that also compare same navigation to null

maumar:
Problem was in the optimization that compared navigation to null, specifically:
o.Customer == null => c.CustomerId == null

If we detected that this optimization could be performed, we assumed that the navigation [o.Customer] is no longer needed, and it was being removed from the navigation joins list.
However, if that navigation was also used for another nav access, then it should not have been removed.

Fix is to add counter to the NavigationJoins list, and only remove given navigation join from the list, if it's not referenced anywhere else in the query.

Also removed some dead code and fixed an issue where we would not apply:
c == o.Customer => c.Id == o.CustomerId

if CustomerId was optional (i.e. clr type was different than the PK on the other side)

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
@@ -32,7 +33,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
     public class NavigationRewritingExpressionVisitor : RelinqExpressionVisitor
     {
         private readonly EntityQueryModelVisitor _queryModelVisitor;
-        private readonly List<NavigationJoin> _navigationJoins = new List<NavigationJoin>();
+        private readonly NavigationJoins _navigationJoins = new NavigationJoins();
         private readonly NavigationRewritingQueryModelVisitor _navigationRewritingQueryModelVisitor;
         private QueryModel _queryModel;
         private QueryModel _parentQueryModel;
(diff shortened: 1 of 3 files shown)


src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -43,10 +44,50 @@ public class NavigationRewritingExpressionVisitor : RelinqExpressionVisitor
         private bool _insideOrderBy;
         private bool _insideMaterializeCollectionNavigation;
 
+        private class NavigationJoins : IEnumerable<NavigationJoin>
+        {
+            private readonly Dictionary<NavigationJoin, int> _navigationJoins = new Dictionary<NavigationJoin, int>();
+
+            public void Add(NavigationJoin navigationJoin)
+            {
+                _navigationJoins.TryGetValue(navigationJoin, out int count);
```
smitpatel: use var.

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -127,14 +168,14 @@ public IEnumerable<NavigationJoin> Iterate()
...
                 yield return this;
 
-                foreach (var navigationJoin in NavigationJoins.SelectMany(nj => nj.Iterate()))
+                foreach (var navigationJoin in Children.SelectMany(nj => nj.Iterate()))
                 {
                     yield return navigationJoin;
                 }
             }
 
             private void Remove(NavigationJoin navigationJoin)
```
smitpatel: remove this function.

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -43,10 +44,50 @@ public class NavigationRewritingExpressionVisitor : RelinqExpressionVisitor
...
+
+                    return true;
+                }
+
+                return false;
+            }
+
+            public IEnumerator<NavigationJoin> GetEnumerator()
+            {
+                return _navigationJoins.Keys.GetEnumerator();
```
anpete: Expression body all the things (R# is your friend :trollface: )

PR #8242, opened 2017-04-20
Query: Use member name for alias for anonymous/DTO projection

### LINKED ISSUE #8204: Query: Anonymous type projections should use member names as column aliases wherever possible
For example anonymous type with new property names:
```
var schools = db.Students.Select(p => new { 
      p.Id, 
      School = p.Name, 
      Score = p.Scores.Sum(s =>  s.Score) 
});
```
generated t-sql:
```
SELECT [p].[Id], [p].[Name], (
                SELECT SUM([s].[Score])
                FROM [Scores] AS [s]
                WHERE [p].[Id] = [s].[StudentId]
            )
            FROM [Students] AS [p]
```
improvement t-sql:
```
SELECT [p].[Id], [p].[Name] AS [School] , (
                SELECT SUM([s].[Score])
                FROM [Scores] AS [s]
                WHERE [p].[Id] = [s].[StudentId]
            ) AS [Score]
            FROM [Students] AS [p]
```

smitpatel:
Resolves #8204 

This introduces aliases when they are not needed therefore, other parts of SelectExpression should use alias only and inner expression of alias should only be visited during projection.

Changes to baseline are submitted in separate in commit for read-ability.  (and coming soon)

### DIFF
--- a/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
@@ -482,23 +482,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
                     .LiftExpressionFromSubquery(table)
                 : new ColumnExpression(_relationalAnnotationProvider.For(property).ColumnName, property, table);
 
-            return IsProjectStar
-                ? GetOrAddToStarProjection(projectedExpressionToSearch)
-                : (_projection.Find(e => _expressionEqualityComparer.Equals(e, projectedExpressionToSearch)) ?? projectedExpressionToSearch);
-        }
-
-        private Expression GetOrAddToStarProjection([NotNull] Expression expression)
-        {
-            var projectedExpression = _starProjection.Find(e => _expressionEqualityComparer.Equals(e, expression));
-
-            if (projectedExpression != null)
-            {
-                return projectedExpression;
-            }
-
-            _starProjection.Add(expression);
-
-            return expression;
+            return projectedExpressionToSearch;
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


smitpatel:
Splitting this out in multiple PRs.

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -482,23 +482,7 @@ public virtual void RemoveTable([NotNull] TableExpressionBase tableExpression)
...
-
-            if (projectedExpression != null)
-            {
-                return projectedExpression;
-            }
-
-            _starProjection.Add(expression);
-
-            return expression;
+            return projectedExpressionToSearch;
```
smitpatel: @anpete - So now binder does not return same instance. (also removed adding it to star projection because it was not used and current handling of star projection is quite inaccurate).

Side-effect of this & alias-ing, when you add something to projection or try to get projection index, you need to account for aliased projection. Which seems a decent logic.

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -293,7 +293,7 @@ private Expression ApplyOptimizations(Expression expression, bool searchConditio
```
smitpatel: This file has only code clean up no changes.

smitpatel:
Easy to review https://github.com/aspnet/EntityFramework/pull/8242/commits/0a8c1633d69ba84dc66b73c6b4aecc1b22c60aac

smitpatel:
Updated.

PR #8244, opened 2017-04-20
DRY code to unique-fy the projection expression

smitpatel:
Unique-fy the code to create uniquely aliased projection for subquery. (which was being used in AddToProjection & PushDownSubquery). This removes CreateUniqueProjectionAlias and introduces method CreateUniqueProjection which makes the projection alias unique (and replaces it in projection & order by)
Side effect of above is, for pushdownsubquery, we don't use ColumnAliasPrefix for new name. Instead we will unique-fy current column name.


smitpatel:
This is pre-work for #8204

Issue #7936, opened 2017-03-20
What's DatabaseName used for?


xyting:
Today, I test some ef core features, I assume the `test_db1` and `test_db2` in the same machine:
step1: I use database named `test_db1` in connection string
step2: I change the database name in  `OnModelCreating(ModelBuilder modelBuilder)` method like this:

```csharp
builder.Model.Relational().DatabaseName = "test_db2";
```
step3: I check where the data inserted, in test_db1? or in test_db2?

The result is data inserted into `test_db1`, so, my question is waht's the `DatabaseName` property of `RelationalModelAnnotations` used for?

ajcvickers:
@xyting The database to use is controlled only by the connection string. The DatabaseName is stored in the model as part of the reverse engineering process, but is otherwise not used.

Note for triage: Consider removing sugar API for this annotation if it is only needed by reverse engineering.

xyting:
@ajcvickers Thanks, I got it as I had investigated the source code last night. Have any easy way to dynamic change the database name initial determined by connection string?

xyting:
@ajcvickers Thanks.

bricelam:
Re-opening so we can address @ajcvickers's comment in triage.

lajones:
Added `breaking change` tag as we would be removing public API.

ajcvickers:
Should do this for 2.0, since it is a breaking change.

lajones:
Fixed with PR #8247.

xyting:
Good job, thanks!

xyting:
@lajones Make it a Scaffolding() one instead is ok if it is only needed by reverse engineering, but, the reason why I test this feature is I want to dynamically change the database name at runtime. That's to say I want support multiple databases/tables sharding (in the same maching, I assume) in the same model. For MySQL, the Schema is the same as database, so for MySQL, I can dynamically change the the Schema to To achieve the same purpose. see [this test](https://github.com/Arch/UnitOfWork/blob/master/src/Microsoft.EntityFrameworkCore.UnitOfWork/UnitOfWork.cs#L53-L82)

bricelam:
@xyting Couldn't you just use a new connection string pointing to a different database?
```C#
var newConnectionString = new SqlConnectionStringBuilder(oldConnectionString) { InitialCatalog = newDatabase }.ToString();
```
You may also be looking for #4019

xyting:
[...]

PR #8247, opened 2017-04-20
Relational().DatabaseName was confusing, make it a Scaffolding() one instead

lajones:
Fix for #7936.

Users were finding the Relational().DatabaseName annotation confusing since it had no effect on the database you were connecting to at runtime. So make it into a Scaffolding() annotation instead to make it clear what it's being used for and make it only available at design-time.

### DIFF
--- a/src/EFCore.Relational.Design/Metadata/Internal/ScaffoldingFullAnnotationNames.cs
+++ b/src/EFCore.Relational.Design/Metadata/Internal/ScaffoldingFullAnnotationNames.cs
@@ -19,12 +19,13 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Metadata.Internal
         protected ScaffoldingFullAnnotationNames(string prefix)
             : base(prefix)
         {
-            UseProviderMethodName = ScaffoldingAnnotationNames.UseProviderMethodName;
-            ColumnOrdinal = ScaffoldingAnnotationNames.ColumnOrdinal;
-            DependentEndNavigation = ScaffoldingAnnotationNames.DependentEndNavigation;
-            PrincipalEndNavigation = ScaffoldingAnnotationNames.PrincipalEndNavigation;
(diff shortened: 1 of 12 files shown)


src/EFCore.Relational.Design/Metadata/Internal/ScaffoldingFullAnnotationNames.cs
```diff
@@ -19,12 +19,13 @@ public class ScaffoldingFullAnnotationNames : RelationalFullAnnotationNames
...
-            PrincipalEndNavigation = ScaffoldingAnnotationNames.PrincipalEndNavigation;
-            EntityTypeErrors = ScaffoldingAnnotationNames.EntityTypeErrors;
-            DbSetName = ScaffoldingAnnotationNames.DbSetName;
+            UseProviderMethodName = prefix + ScaffoldingAnnotationNames.UseProviderMethodName;
+            ColumnOrdinal = prefix + ScaffoldingAnnotationNames.ColumnOrdinal;
+            DependentEndNavigation = prefix + ScaffoldingAnnotationNames.DependentEndNavigation;
+            PrincipalEndNavigation = prefix + ScaffoldingAnnotationNames.PrincipalEndNavigation;
+            EntityTypeErrors = prefix + ScaffoldingAnnotationNames.EntityTypeErrors;
+            DbSetName = prefix + ScaffoldingAnnotationNames.DbSetName;
+            DatabaseName = prefix + ScaffoldingAnnotationNames.DatabaseName;
```
lajones: I think the fact that all of these were missing the prefix was just an oversight. It worked because nothing clashed with "no prefix". But we shouldn't leave it that way. This could affect anyone (probably only providers) who were looking up annotations using hard-coded names, but hopefully people were not doing that but instead using the programmatic way of getting them anyway. Plus this was already a breaking change - so it seemed like the right time to correct this.

lajones:
Checked in with commit 7b860decd.

xyting:
Good

PR #8249, opened 2017-04-21
Add owned entity types support to more places in query

AndriySvyryd:
Record the EntityType when visiting existing JoinClause
Update QuerySourceMapping when cloning
Add provider-specific model building extensions
Add more query tests for owned entity types


src/EFCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
```diff
@@ -4249,9 +4439,11 @@ where l2.OneToOne_Required_FK.OneToMany_Optional.Select(i => i.OneToOne_Optional
         {
             using (var context = CreateContext())
             {
+                var expected = l2oQuery(ExpectedSet<TItem1>()).ToArray();
+                var actual = efQuery(Set<TItem1>(context)).ToArray();
```
maumar: switch the order so that actual executes first (in case expected has a null ref or something)

test/EFCore.SqlServer.FunctionalTests/ComplexNavigationsOwnedQuerySqlServerTest.cs
```diff
@@ -0,0 +1,45 @@
...
+            _testOutputHelper = testOutputHelper;
+            //TestSqlLoggerFactory.CaptureOutput(testOutputHelper);
+        }
+
+        protected override void ClearLog() => TestSqlLoggerFactory.Reset();
+
+        private bool SupportsOffset => TestEnvironment.GetFlag(nameof(SqlServerCondition.SupportsOffset)) ?? true;
+
+        [ConditionalFact]
+        public override void Where_nav_prop_reference_optional1()
```
maumar: remove

src/EFCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
AndriySvyryd: Wouldn't you want to see the exception in expected first, so that you don't waste time debugging actual?

maumar: if expected runs first and throws, then the product code never runs - in my experience, exceptions coming from expected (in this test suite at least) are mostly due to null refs during nav access, something that is perfectly fine for the product query.

AndriySvyryd: ok

src/EFCore.Specification.Tests/ComplexNavigationsOwnedQueryTestBase.cs
```diff
@@ -0,0 +1,270 @@
...
+            base.Required_navigation_on_a_subquery_with_First_in_projection();
+        }
+
+        // Naked instances not supported
+        public override void Key_equality_two_conditions_on_same_navigation()
+        {
+        }
+
+        // One-to-many not supported yet
+        public override void Multiple_SelectMany_with_string_based_Include()
```
maumar: add skip annotation here also? or would that be too much spam in the build.cmd?

AndriySvyryd: I think Skip is only for tests that are expected to pass

src/EFCore/Metadata/Internal/EntityMaterializerSource.cs
```diff
@@ -48,16 +48,13 @@ public class EntityMaterializerSource : IEntityMaterializerSource
...
 
         private static TValue TryReadValue<TValue>(
-            ValueBuffer valueBuffer, 
+            ValueBuffer valueBuffer,
             int index,
             IPropertyBase property = null)
         {
-            object untypedValue = null;
-
+            var untypedValue = valueBuffer[index];
```
AndriySvyryd: Yes and in that case the exception message would be wrong:
 An exception occurred while reading a database value for property 'Order.ReplacedOrderId'. The expected type was 'int?' but the actual value was null.

anpete: No existing tests fails, right? If not, then I think this is OK. We are effectively saying that seeing ValueBuffer.Empty is an error here.

AndriySvyryd: Well we throw in either case, just the exception is different

src/EFCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -81,18 +69,18 @@ public QueryOptimizer([NotNull] IModel model)
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public virtual void Optimize(
-            IReadOnlyCollection<IQueryAnnotation> queryAnnotations,
+            QueryCompilationContext queryCompilationContext,
```
anpete: ctor dependency?

AndriySvyryd: QueryCompilationContext is not in DI

src/EFCore/Metadata/Internal/EntityMaterializerSource.cs
anpete: Yes, and we are discussing whether that is correct 😄 This change asserts that a null coming from the database (incorrectly) never results in ValueBuffer.Empty appearing in the executing query.

src/EFCore/Query/Internal/QueryOptimizer.cs
anpete: Ah, I had assumed that QueryOptimizer was not in DI. Makes me wonder whether having QueryOptimizer in DI is correct as it has state - We have to be careful here 💣

src/EFCore/Metadata/Internal/EntityMaterializerSource.cs
anpete: Actually, I see that your change improves this situation! 🙌

src/EFCore.Relational/Metadata/RelationalEntityTypeAnnotations.cs
```diff
@@ -51,11 +51,16 @@ public virtual string TableName
...
                            RelationalFullAnnotationNames.Instance.TableName,
                            ProviderFullAnnotationNames?.TableName)
-                       ?? EntityType.ShortName();
+                       ?? GetDefaultTableName();
             }
             [param: CanBeNull] set { SetTableName(value); }
         }
 
+        private string GetDefaultTableName()
+            => EntityType.HasDelegatedIdentity()
```
anpete: Should we cache this?

AndriySvyryd: Sure,  more caching _never_ hurts

src/EFCore/Query/ExpressionVisitors/Internal/EntityEqualityRewritingExpressionVisitor.cs
```diff
@@ -20,17 +19,15 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
...
-        private readonly IModel _model;
+        private readonly QueryCompilationContext _queryCompilationContext;
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public EntityEqualityRewritingExpressionVisitor([NotNull] IModel model)
+        public EntityEqualityRewritingExpressionVisitor([NotNull] QueryCompilationContext queryCompilationContext)
         {
```
smitpatel: null check?

AndriySvyryd: We don't need checks on .Internal API

src/EFCore/Query/QueryCompilationContext.cs
```diff
@@ -123,6 +123,26 @@ public virtual void AddOrUpdateMapping([NotNull] IQuerySource querySource, [NotN
             => _querySourceEntityTypeMapping[querySource] = entityType;
 
         /// <summary>
+        ///     Updates the query source mappings to the new query sources
+        /// </summary>
+        /// <param name="querySourceMapping"> The new query source mapping </param>
+        public virtual void UpdateMapping(QuerySourceMapping querySourceMapping)
```
smitpatel: null annotation.

src/EFCore/Query/Internal/QueryOptimizer.cs
anpete: Actually, I see that your change improves this situation! 🙌

AndriySvyryd: Filed https://github.com/aspnet/EntityFramework/issues/8256

src/EFCore.Relational/Metadata/RelationalEntityTypeAnnotations.cs
AndriySvyryd: Since this is a relational concept, proper caching would require non-trivial infrastructure changes, I think it's better to do it as part of https://github.com/aspnet/EntityFramework/issues/8258

PR #8269, opened 2017-04-24
Test: Refactoring: Remove derived Northwind contexts from relational providers.

anpete:
- Move the Northwind database creation logic back to the test store so we can have multiple derived Northwind context types per provider.


ErikEJ:
providers beware?

PR #8274, opened 2017-04-25
Improving verification for ComplexNavigationTests so that it no longer relies on running product code to create expected results

maumar:
- added verification for include queries that does deep comparison, while taking into account specified expected includes,
- added verification for single result queries.

### DIFF
--- a/src/EFCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.ComplexNavigationsModel;
 using Microsoft.EntityFrameworkCore.Specification.Tests.TestUtilities.Xunit;
+using Microsoft.EntityFrameworkCore.Specification.TestUtilities;
 using Xunit;
 // ReSharper disable InconsistentNaming
 // ReSharper disable MergeConditionalExpression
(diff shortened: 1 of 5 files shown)


src/EFCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
```diff
@@ -27,15 +28,15 @@ protected ComplexNavigationsQueryTestBase(TFixture fixture)
...
             TestStore = Fixture.CreateTestStore();
+
+            ResultAsserter = new ComplexNavigationsQueryResultAsserter();
         }
 
         protected TFixture Fixture { get; }
 
         protected TTestStore TestStore { get; }
 
-        protected virtual void ClearLog()
```
maumar: @anpete FYI ;)

src/EFCore.Specification.Tests/ComplexNavigationsQueryTestBase.cs
```diff
@@ -321,40 +255,39 @@ public virtual void Multi_level_include_with_short_circuiting()
                     .Include(x => x.Placeholder.Globalizations)
                     .ThenInclude(x => x.Language);
 
-                var result = query.ToList();
+                var result = query.ToList().OrderBy(e => e.Name).ToList();
```
smitpatel: Why this one does not use AssertIncludeQuery?

maumar: those entities are not part of in memory data, and are only used for this one test (copied verbatim from a bug report long time ago). I should probably move this to QueryBugsTests instead

Issue #5553, opened 2016-05-27
Reverse Engineer: Scaffolding fails on Azure Sql Server v11 and SQL 2008 because sequences are not supported


mhenderson442:
Steps to reproduce
`Scaffold-DbContext -Provider "Microsoft.EntityFrameworkCore.SqlServer" -Context "DbMercurySchool" -OutputDir "DataModel"  -Connection "Server=[Sql Azure Database Server];Data Source=[Sql Azure Database];Initial Catalog=MercurySchool;Persist Security Info=False;User ID=[User ID ];Password=[My Password];Pooling=False;MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=False;Connection Timeout=30;"`

The issue: SQL error thrown

`Error Number:208,State:1,Class:16 One or more errors occurred. (Invalid object name 'sys.sequences'.)`

See attached file.

[scaffold_error.txt](https://github.com/aspnet/EntityFramework/files/286958/scaffold_error.txt)

EF Core 
- version: `1.0.0-rc2-final`
- Operating system: Win 10
- Visual Studio version: VS 2015 Community

Other details about my project setup:

So far database has one table. The schema for the table is `academy` not `dbo`.

smitpatel:
Ah Sequence issue. We are trying to get metadata about sequences but not all version of Sql Azure supports sequences so database error.
@mhenderson442 - Can you also tell us details about the server like version of Sql Azure?

mhenderson442:
The server hosting my database describes the following features:

- Server Version: V2

- Extended Transact SQL 

- API Version 12

ErikEJ:
Do a SELECT @@version

ErikEJ:
@lajones  
     DECLARE @sqlVersion numeric(4,2)
     SELECT @sqlVersion = LEFT(CAST(SERVERPROPERTY('productversion') as VARCHAR), 4)
     IF (SERVERPROPERTY('EngineEdition') == 5 AND @sqlVersion < 12) OR (@sqlVersion < 11) 
     -- No sys.sequences

mhenderson442:
`SELECT @@VERSION` outputs the following:

Microsoft SQL Azure (RTM) - 11.0.9231.349 
	May 20 2016 14:01:37 
	Copyright (c) Microsoft Corporation

`SERVERPROPERTY('EngineEdition')` = 5

`SERVERPROPERTY('productversion')` = 11.0

**Off in the weeds:**
[...]

PR #8276, opened 2017-04-25
Ignore hidden columns (used in SQL Server system versioned tables)

### LINKED ISSUE #8192: Hidden columns are scaffolded
When scaffolding  a database to EF Core, Hidden Columns are not excluded.  As a result, columns used for SQL Temporal Tables get scaffolded back into the application. As a result, if the DB has not had those turned on, it breaks the EF code against another database which does not have them.

### Steps to reproduce
Create a Database Project and scaffold it to a SQL Server.
Now run Scaffold-DbContext to build your EF context off that server.
Then in SQL Server follow these steps https://docs.microsoft.com/en-us/sql/relational-databases/tables/creating-a-system-versioned-temporal-table#alter-non-temporal-table-to-be-system-versioned-temporal-table to add Temporal System Versioning.  
Now scaffold from the database again.  Notice that the versioning columns are added, despite the fact they are marked as hidden in SQL.

ErikEJ:
fixes #8192

### DIFF
--- a/test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
+++ b/test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
@@ -351,6 +351,60 @@ CREATE INDEX Unicorn_Filtered_Index
             }
         }
 
+        [ConditionalFact]
+        [SqlServerCondition(SqlServerCondition.SupportsHiddenColumns)]
+        public void Hidden_Column()
+        {
+            using (var scratch = SqlServerTestStore.Create("WithHiddenColumns"))
+            {
+                scratch.ExecuteNonQuery(@"
+CREATE TABLE dbo.SystemVersioned
+(
+     Id int NOT NULL PRIMARY KEY CLUSTERED,
+     Name varchar(50) NOT NULL,
+     SysStartTime datetime2 GENERATED ALWAYS AS ROW START HIDDEN NOT NULL,
+     SysEndTime datetime2 GENERATED ALWAYS AS ROW END HIDDEN NOT NULL,
+     PERIOD FOR SYSTEM_TIME(SysStartTime, SysEndTime)
+)
+WITH (SYSTEM_VERSIONING = ON (HISTORY_TABLE = dbo.History));
+");
+
+                var configuration = new ReverseEngineeringConfiguration
+                {
+                    ConnectionString = scratch.ConnectionString,
+                    ProjectPath = TestProjectDir + Path.DirectorySeparatorChar,
+                    ProjectRootNamespace = TestNamespace,
+                    ContextClassName = "SystemVersionedContext",
+                    UseFluentApiOnly = true
+                };
+                var expectedFileSet = new FileSet(new FileSystemFileService(),
+                    Path.Combine("ReverseEngineering", "Expected"),
+                    contents => contents.Replace("{{connectionString}}", scratch.ConnectionString))
+                {
+                    Files = new List<string>
+                    {
+                        "SystemVersionedContext.expected",
+                        "SystemVersioned.expected",
+                    }
+                };
+
+                var filePaths = Generator.GenerateAsync(configuration).GetAwaiter().GetResult();
+
+                scratch.ExecuteNonQuery(@"
+ALTER TABLE dbo.SystemVersioned SET (SYSTEM_VERSIONING = OFF);
+DROP TABLE dbo.History;
+");
+
+                var actualFileSet = new FileSet(InMemoryFiles, Path.GetFullPath(TestProjectDir))
+                {
+                    Files = new[] { filePaths.ContextFile }.Concat(filePaths.EntityTypeFiles).Select(Path.GetFileName).ToList()
+                };
+
+                AssertEqualFileContents(expectedFileSet, actualFileSet);
+                AssertCompile(actualFileSet);
+            }
+        }
+
         protected override ICollection<BuildReference> References { get; } = new List<BuildReference>
         {
 #if NET46
--- a/test/EFCore.SqlServer.FunctionalTests/config.json
+++ b/test/EFCore.SqlServer.FunctionalTests/config.json
@@ -5,7 +5,8 @@
         "ElasticPoolName": "",
         "SupportsSequences": true,
         "SupportsOffset": true,
-        "SupportsMemoryOptimized": false
+        "SupportsMemoryOptimized": false,
+        "SupportsHiddenColumns": false
       }
     }
 }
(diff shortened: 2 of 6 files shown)


test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
```diff
@@ -351,6 +351,60 @@ PRIMARY KEY (Id)
             }
         }
 
+        [ConditionalFact]
+        [SqlServerCondition(SqlServerCondition.SupportsSequences)]
+        public void Hidden_Column()
+        {
+            using (var scratch = SqlServerTestStore.Create("SqlServerE2E"))
```
lajones: Nit: I would rename the DB to be something more easily tracked to this test e.g. HiddenColumnTest.

test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
```diff
@@ -351,6 +351,60 @@ PRIMARY KEY (Id)
...
+            {
+                scratch.ExecuteNonQuery(@"
+CREATE TABLE dbo.SystemVersioned
+(
+      Id int NOT NULL PRIMARY KEY CLUSTERED
+    , Name varchar(50) NOT NULL
+    , SysStartTime datetime2 GENERATED ALWAYS AS ROW START HIDDEN NOT NULL
+    , SysEndTime datetime2 GENERATED ALWAYS AS ROW END HIDDEN NOT NULL
+    , PERIOD FOR SYSTEM_TIME(SysStartTime, SysEndTime)
+)
```
lajones: Nit: Personally I like the "comma at the beginning of the line" convention, but as a team we decided to put them at the end of the previous line instead.

test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
```diff
@@ -351,6 +351,60 @@ PRIMARY KEY (Id)
             }
         }
 
+        [ConditionalFact]
+        [SqlServerCondition(SqlServerCondition.SupportsSequences)]
```
lajones: Does this condition just happen to coincide? Or are supporting sequences and supporting hidden columns different things? I'm wondering whether we should create a new condition `SqlServerCondition.SupportsHiddenColumns`?

smitpatel: We should create condition for SqlServer 2016

lajones: @smitpatel I don't think this is just a SQL Server 2016 issue. See bug #5553 about supporting sequences in SQL Azure as well. This could be similar, so I would prefer to keep the flag named for the function, not the version of SQL Server.

ErikEJ: I will add a new condition - #5553 is getting moot, as Azure DB v11 is being deprecated

test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
ErikEJ: Will do!

test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
ErikEJ: I will move the commas :smile:

test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/SqlServerE2ETests.cs
smitpatel: Preferable thing is to have flag based on the function. I mistakenly thought that we may end up with multiple conditions. Instead of having a large set of conditions to specify its good to have only one. But we won't hit such issue because this conditions are +ve. Means specify the one which needs to be present.

ErikEJ:
@lajones I noticed the build failed, as the NuGet v2 api had some issues earlier - maybe you can remove the lookup for v2 feed during build, and only use v3 feed?

lajones:
Sorry, @ErikEJ, because we have a new flag we also need to update the [config.json](https://github.com/aspnet/EntityFramework/blob/dev/test/EFCore.SqlServer.FunctionalTests/config.json) file to have `"SupportsHiddenColumns": true`.

ErikEJ:
@lajones Done

test/EFCore.SqlServer.FunctionalTests/config.json
```diff
@@ -5,7 +5,8 @@
         "ElasticPoolName": "",
         "SupportsSequences": true,
         "SupportsOffset": true,
-        "SupportsMemoryOptimized": false
+        "SupportsMemoryOptimized": false,
+        "SupportsHiddenColumns": true
```
smitpatel: Shouldn't this be false as MemoryOptimized? This is our default connection string for testing and we still use SqlServer2014 which does not support hidden columns.

ErikEJ: I am confused now - so you use 2016 on the CI (appveyor) and 2014 on what ?? (your dev pcs) - in that case it should be false, yes (it has no impact on the functional tests, though)

smitpatel: @ErikEJ - Did not realize this confusion 😄 Minimum requirement to build EF repo is SqlServer 2014 (or it could run with older version too may be).
We use SqlServer 2016 in appveyor because appveyor was failing to start sqlserver 2014 service and build failed always providing no feedback.
The normal dev work is on sql server 2014 as of now.

lajones: @ErikEJ @smitpatel Smit is right. I was thinking this was like `SupportsSequences`, but it's more like `SupportsMemoryOptimized` and so should be false by default.

lajones: In the meantime - I have run this test on my dev machine (which has SQL Server 2016) and it works fine.

ErikEJ:
@lajones Thanks!  I have updated the config.json file to match your test and dev environment

lajones:
@ErikEJ Looks good. Could you squash all the commits together please and I can check it in for you? Thx.

ErikEJ:
@lajones Done

lajones:
Checked in with commit 3647f579e. Thanks very much, @ErikEJ.

PR #8279, opened 2017-04-25
Use Bundled NETStandard.Library \ NETCoreApp versions instead of expl…

pranavkm:
…icitly specifying one

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -20,6 +20,6 @@
     <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(InternalAspNetCoreSdkVersion)" PrivateAssets="All" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(OutputType)' == 'library' And '$(DisableImplicitFrameworkReferences)' != 'true'">
-    <PackageReference Include="NETStandard.Library" Version="$(NetStandardImplicitPackageVersion)" />
+    <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


bricelam:
Why?

pranavkm:
From an email I sent out about this to the engineering team:

1.	Scaffolding doesn’t have a RuntimeFrameworkVersion. Consequently they are now (unwittingly) referencing the Bundled Microsoft.NETCore.App version (https://github.com/dotnet/sdk/pull/1012). This M.N.A is slightly old and references an older build of NETStandard.Library
2.	The version of NETStandard.Library on our feed is newer. Changing NETStandardImplicitVersion to 2.0.0-* causes all our projects which includes some of Scaffolding’s dependencies to reference this version.

Referencing different versions of NS.Library (more specifically microsoft.packaging.tools a dependency of NS.Library) causes builds to fail like so:

``
`\.nuget\packages\microsoft.packaging.tools\1.0.0-preview1-25221-01\build\Microsoft.Packaging.Tools.targets(60,5): error MSB4062: The "HandlePackageFileConflicts" task could not be loaded from the assembly .nuget\packages\microsoft.packaging.tools\1.0.0-preview1-5221-01\build\Microsoft.Packaging.Tools.Tasks.dll. 
\.nuget\packages\microsoft.packaging.tools\1.0.0-preview1-25217-01\build\Microsoft.Packaging.Tools.targets(60,5): error MSB4062: The "HandlePackageFileConflicts" task could not be loaded from the assembly .nuget\packages\microsoft.packaging.tools\1.0.0-preview1-25217-01\build\Microsoft.Packaging.Tools.Tasks.dll.
```

We considered the options and it seemed like doing this was something we had hoped to do (it's closer to the acquisition story for our customers), and avoids this issue https://github.com/aspnet/Coherence-Signed/issues/458 since the versions are tied to the CLI \ SDK rather than feed values.

Issue #8280, opened 2017-04-25
Tools: Ensure both netcoreapp1.x and 2.0 projects work


bricelam:
Will EFCore.Tools 2.0 work on netstandard1.x and netcoreapp1.x projects?

If the startup project is netcoreapp2.0 or net461: Yes.
If not: I don't know.

We need to decided what is supported and engineer a solution to support that. This needs to be done for 2.0.0-preview1.

bricelam:
Blocked since netcoreapp2.0 tools don't even work yet.

bricelam:
We also need to test netcoreapp2.0 projects when we can.

bricelam:
Discussed yesterday and it should work with all frameworks if you're using the 2.0 SDK.

bricelam:
Should be unblocked by the next CLI build.

bricelam:
Tested. Everything looks good. Just one issue with `netstandard2.0` class library projects: #8344

PR #8281, opened 2017-04-25
[2.0.0-preview1] Update dotnet-ef to netcoreapp2.0

bricelam:
I have no idea how much of this will even work since it's blocked on NuGet/Home#4396. I've filed #8280 to follow up with testing after that gets fixed.


natemcmaster:
FYI when you merge, some downstream repos might fail to build if they have a DotNetCliToolReference to dotnet-ef. I looked through Universe build and commented out most instances until we get https://github.com/dotnet/sdk/pull/1128. See https://github.com/aspnet/Universe/issues/506

bricelam:
Found and fixed one downstream repo in Universe that wasn't commented out.

Issue #6946, opened 2016-11-04
Logging: Use template based format string to log structured DbCommand log data


smitpatel:
At present EF logging of Db Command uses `DbCommandLogData` object with custom formatter which would generate a plain string message for logger to log. This blocks structured logging since logger would get only `DbCommandLogData` object instead of different parts of log messages (for querying). During #6096, we overcame this gap by implementing `IEnumerable<KeyValuePair<string,object>>` in #6201 which means that `DbCommandLogData` can be queried.

Current logging framework allows generation of structured log using template based string. The template string contains placeholders to fill the values of arguments and provides ability to query the log data without requiring any additional code. (See https://github.com/aspnet/EntityFramework/pull/6201#issuecomment-236076125
and example in MVC at https://github.com/aspnet/Mvc/blob/a78f77afde003c4a3fcf5dd7b6dc13dd9c85f825/src/Microsoft.AspNetCore.Mvc.Razor/Internal/MvcRazorLoggerExtensions.cs#L23)

This saves us trouble of creating custom formatter and extra type to store the state of DbCommandLog.
`DbCommandLogData` which is marked as obsolete at present would be removed when logging is implemented this way.

julielerman:
Guys -- this is the "kludge" I was talking about on Wednesday that I was confused about when you said it was implemented in 1.1. 
This is how I'm currently using & explaining it with 1.1:
![image](https://cloud.githubusercontent.com/assets/5007120/22151014/238c4cb6-deea-11e6-9223-53acaeed5c3b.png)

Am I missing something?

ErikEJ:
@julielerman I am doing this instead (based on sample code from Rowan) https://github.com/ErikEJ/EntityFramework.SqlServerCompact/blob/master/src/Provider40/Extensions/Logging/DbSimpleLogger.cs

rowanmiller:
Our logging is overly complex to work with at the moment, and we have an item on the list to address this in 2.0 (#218). Honestly, either way is fine with 1.x. `DbCommandLogData` is marked as obsolete... but we won't be removing it until 2.0 - at which point there will be a much cleaner way to do this.

smitpatel:
Removing milestone for this. Since we are re-addressing logging, should we merge this issue with #218

Issue #6802, opened 2016-10-17
Logger event IDs can be ambiguous


ajcvickers:
If each named logger is only associated with a single space of event IDs (i.e. a single enum), then event IDs can be used to identify messages. However, we sometimes share loggers between pieces of code which then log events using different event ID spaces.

For example, SqlServerScaffoldingModelFactory uses the logger from RelationalScaffoldingModelFactory. This means that event IDs from two spaces (SqlServerDesignEventId and RelationalDesignEventId) get written to the same logger, making the IDs ambiguous. We share loggers in other places, so I suspect there are other cases of this lurking in the codebase.

bricelam:
@ajcvickers Should this one be assigned to you?

ajcvickers:
Yep, at least until after we decide what to do. :-)

PR #8288, opened 2017-04-26
Update EventIds to be unique and be consistent with ILogger/DiagnosticSource

ajcvickers:
Issues #218 #6802

(Didn't update breaking change exceptions yet.)
(Doesn't cover #6946 - Template-based format strings.)

This change introduces unique event IDs and a pattern such that every time an event is logged it is sent to both ILogger and DiagnosticSource. Specifically:
* EventIds are globally unique across core/relational/any given provider. (As discussed in #218)
* The same EventIds are used for ILogger and DiagnosticSource (ILogger uses the int/name tuple, DiagnosticSource uses just the name)
* EventIds align with logger categories
* Automated tests for the EventId/LoggerExtensions pattern--every EventId has a LoggerExtension method that works as expected.
* Warnings configuration is updated to use the new event ids
* Updated IsEnabled to handle warnings as errors and ignored warnings more transparently.


src/EFCore.Design/Design/Internal/OperationLogger.cs
```diff
@@ -55,8 +54,8 @@ public virtual IDisposable BeginScope<TState>(TState state)
             Func<TState, Exception, string> formatter)
         {
             // Only show SQL when verbose
-            if (_categoryName == typeof(RelationalCommandBuilderFactory).FullName
-                && eventId.Id == (int)RelationalEventId.ExecutedCommand)
+            if (_categoryName == LoggerCategory.Database.Sql.Name
+                && eventId.Id == RelationalEventId.CommandExecuted.Id)
```
anpete: We should make EventId implement '=='

src/EFCore.Design/Infrastructure/DesignEventId.cs
```diff
@@ -1,24 +1,137 @@
...
-        ReusingNamespace,
-        ReusingDirectory,
-        RevertingSnapshot,
-        WritingMigration,
-        ReusingSnapshotName,
-        DestructiveOperation,
-        ForeignMigrations
+        // Warning: These values must not change between releases.
+        // Only add new values to the end of sections, never in the middle.
+        // Try to use <Nown><Verb> naming and be consistent with existing names.
```
anpete: Noun?

src/EFCore.Design/Design/Internal/OperationLogger.cs
anpete: Might be nice to make LoggerCategory<T> implicitly convertible to string

src/EFCore.Design/Infrastructure/DesignEventId.cs
```diff
@@ -1,24 +1,137 @@
...
+        private enum Id
```
anpete: Not sure it's worth having these private enums. Could just inline the magic number when creating the EventId.

src/EFCore.Design/Migrations/Design/MigrationsScaffolder.cs
```diff
@@ -144,19 +141,15 @@ public class MigrationsScaffolder
...
                     modelSnapshotName = lastModelSnapshotName;
                 }
             }
 
             if (upOperations.Any(o => o.IsDestructiveChange))
             {
-                _logger.LogWarning(
-                    DesignEventId.DestructiveOperation,
-                    () => DesignStrings.DestructiveOperation);
+                _logger.DestructiveOperation(upOperations.Where(o => o.IsDestructiveChange));
```
anpete: Review any payloads that allocate/do work even when logging/diagnostics could be disabled.

src/EFCore.Design/Infrastructure/DesignEventId.cs
```diff
@@ -1,24 +1,137 @@
...
 namespace Microsoft.EntityFrameworkCore.Infrastructure
 {
-    public enum DesignEventId
+    /// <summary>
+    ///     <para>
+    ///         Event IDs for design events that correspond to messages logged to an <see cref="ILogger" />
+    ///         and events sent to a <see cref="DiagnosticSource" />.
+    ///     </para>
+    ///     <para>
+    ///         These IDs are also used to with <see cref="WarningsConfigurationBuilder" /> configure the
```
lajones: "to with <cref> configure" -> "with <cref> to configure". I think this is in all the other similar classes too. (As is the noun comment by Andrew below).

src/EFCore.Design/Scaffolding/Internal/ScaffoldingServiceCollectionExtensions.cs
```diff
@@ -32,6 +33,8 @@ public static IServiceCollection AddScaffolding([NotNull] this IServiceCollectio
                 .AddSingleton<EntityTypeWriter>()
                 .AddSingleton<CodeWriter, StringBuilderCodeWriter>()
                 .AddSingleton<ILoggingOptions, LoggingOptions>()
+                .AddSingleton<DiagnosticSource>(new DiagnosticListener(LoggerCategory.Root))
+                .AddSingleton(typeof(IDiagnosticsLogger<>), typeof(DiagnosticsLogger<>))
                 .AddSingleton(typeof(IInterceptingLogger<>), typeof(InterceptingLogger<>));
```
lajones: Do we still need the `IInterceptingLogger` since Scaffolding now uses a `IDiagnosticsLogger`?

src/EFCore.Relational.Design/Properties/RelationalDesignStrings.resx
```diff
@@ -138,24 +156,54 @@
...
+    <value>Found a column on foreign key {tableName}.{fkName} with an empty or null name. Not including column in foreign key</value>
+  </data>
+  <data name="ColumnNameEmptyOnTable" xml:space="preserve">
+    <value>Found a column on table {tableName} with an empty or null name. Skipping column.</value>
+  </data>
+  <data name="UnableToFindTableForIndex" xml:space="preserve">
+    <value>For index {indexName}. Unable to find parent table {tablename}. Skipping index.</value>
+  </data>
+  <data name="IndexNameEmpty" xml:space="preserve">
+    <value>Found an index on table {tablename} with an empty or null name. Skipping index.</value>
```
lajones: Nit: "tablename" -> "tableName". Similarly for next resource.

src/EFCore.Design/Design/Internal/OperationLogger.cs
ajcvickers: I'm actually not sure what the semantics is supposed to be for EventIds that have an Id but no name when compared to EventIds that have the same Id and a name. The EventId code makes it look like the name is optional, but then the comparison current requires the name to match to compare equal, so it seems inconsistent. I have a not to file an issue on Logging for this.

I'll do the implicit conversion on LoggerCategory.

src/EFCore.Design/Infrastructure/DesignEventId.cs
ajcvickers: I did it that way at first, but using the enum helps prevent accidental reuse of the same ID as a copy/paste error.

src/EFCore.Design/Migrations/Design/MigrationsScaffolder.cs
ajcvickers: I have general tried to do this. What would you suggest here? I went for an IEnumerable that is not enumerated unless logging is enabled, which seems pretty low cost. It could be a delegate instead, but not sure that would be lower cost.

src/EFCore.Design/Scaffolding/Internal/ScaffoldingServiceCollectionExtensions.cs
ajcvickers: Yes. IDiagnosticLogger depends on IInterceptingLogger

src/EFCore.Design/Migrations/Design/MigrationsScaffolder.cs
anpete: Yeah, we could just decide we are OK allocating in these cases - IIRC we were still allocating delegates in the old impl. Otherwise we need an option to move the enabled check to the call site.

src/EFCore.Design/Infrastructure/DesignEventId.cs
anpete: Can we just add a test that does the validation instead?

src/EFCore.Design/Migrations/Design/MigrationsScaffolder.cs
ajcvickers: That's correct, so in general we are allocating a lot less now.

src/EFCore.Design/Infrastructure/DesignEventId.cs
ajcvickers: Yep...but I like it like it is. 😸 Let's briefly discuss in the design meeting tomorrow to see what others prefer.

src/EFCore/Extensions/Internal/CoreLoggerExtensions.cs
```diff
@@ -2,36 +2,441 @@
...
+                        IServiceProviders = serviceProviders
+                    });
+            }
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static void LogDebug<TLoggerCategory>(
```
anpete: Should these go away?

ajcvickers: They will at least change as part of #6946. Not sure yet if they will go away completely or not.

src/EFCore.Design/Design/Internal/OperationLogger.cs
ajcvickers: Filed https://github.com/aspnet/Logging/issues/607 on EventId

anpete:
Backlog: We could probably gen all of this boilerplate from a metadata file like we do for resx.

src/EFCore.Design/Infrastructure/DesignEventId.cs
```diff
@@ -1,24 +1,137 @@
...
+            SnapshotRemoving,
+            SnapshotFileNotFound,
+            SnapshotWriting,
+            NamespaceReusing,
+            DirectoryReusing,
+            SnapshotReverting,
+            MigrationWriting,
+            SnapshotNameReusing,
+            DestructiveOperation,
+            ForeignMigrations
```
NickCraver: IMO worth adding a comment here (as the above won't show up in a diff review later), e.g.:
```
// New Additions must happen here to prevent ID movement
```
...or make those IDs explicit in the enum to *really* prevent such a thing.

src/EFCore.Relational.Design/Infrastructure/RelationalDesignEventId.cs
```diff
@@ -1,32 +1,244 @@
...
+            IndexColumnNotNamedWarning,
+
+            // Scaffolding events
+            TableFound = CoreEventId.RelationalDesignBaseId + 1000,
+            TableSkipped,
+            ColumnSkipped,
+            IndexFound,
+            IndexColumnFound,
+            IndexColumnSkipped,
+            SequenceFound
```
NickCraver: Same comment as above - IMO this would be far more resilient to both unintentional breaks and readability with:
```c#
TableFound = CoreEventId.RelationalDesignBaseId + 1000,
TableSkipped = CoreEventId.RelationalDesignBaseId + 1001,
ColumnSkipped = CoreEventId.RelationalDesignBaseId + 1002,
IndexFound = CoreEventId.RelationalDesignBaseId + 1003,
IndexColumnFound = CoreEventId.RelationalDesignBaseId + 1004,
IndexColumnSkipped = CoreEventId.RelationalDesignBaseId + 1005,
SequenceFound = CoreEventId.RelationalDesignBaseId + 1006
```

src/EFCore.Relational.Design/Infrastructure/RelationalDesignEventId.cs
```diff
@@ -1,32 +1,244 @@
...
+        public static readonly EventId ForeignKeyNotNamedWarning = MakeScaffoldingId(Id.ForeignKeyNotNamedWarning);
+
+        /// <summary>
+        ///     A foreign key column was not found.
+        ///     This event is in the <see cref="LoggerCategory.Scaffolding" /> category.
+        /// </summary>
+        public static readonly EventId ForeignKeyColumnMissingWarning = MakeScaffoldingId(Id.ForeignKeyColumnMissingWarning);
+
+        /// <summary>
+        ///     A foreign key column was not found.
```
NickCraver: Same description as above, intentional?

src/EFCore.Relational.Design/Infrastructure/RelationalDesignEventId.cs
```diff
@@ -1,32 +1,244 @@
...
+        public static readonly EventId ColumnNotNamedWarning = MakeScaffoldingId(Id.ColumnNotNamedWarning);
+
+        /// <summary>
+        ///     An index is not named. 
+        ///     This event is in the <see cref="LoggerCategory.Scaffolding" /> category.
+        /// </summary>
+        public static readonly EventId IndexNotNamedWarning = MakeScaffoldingId(Id.IndexNotNamedWarning);
+
+        /// <summary>
+        ///     The table references by an index was not found.
```
NickCraver: referenced*

src/EFCore.Relational.Design/Metadata/SequenceModel.cs
```diff
@@ -18,5 +18,8 @@ public class SequenceModel : Annotatable
         public virtual long? Min { get; [param: CanBeNull] set; }
         public virtual long? Max { get; [param: CanBeNull] set; }
         public virtual bool? IsCyclic { get; [param: CanBeNull] set; }
+
+        public virtual string DisplayName
+            => (!string.IsNullOrEmpty(SchemaName) ? SchemaName + "." : "") + Name;
```
NickCraver: Since this is often used twice in a row - memoize it?

Also a shorter call path: 
```c#
!string.IsNullOrEmpty(SchemaName) ? (SchemaName + "." + Name) : Name;
```

NickCraver:
...so I'm guessing that's a no?

src/EFCore.Relational.Design/Infrastructure/RelationalDesignEventId.cs
ajcvickers: @NickCraver We discussed this today in design meeting and I will look at a few options for avoiding breaks.

src/EFCore.Relational.Design/Infrastructure/RelationalDesignEventId.cs
ajcvickers: Will fix.

src/EFCore.Relational.Design/Infrastructure/RelationalDesignEventId.cs
ajcvickers: Will fix.

ajcvickers:
@NickCraver Sorry for not seeing your comments. GitHub did not send me the usual notifications--not sure why. We have been discussing payload types today and I am going to go comment on #7939 and then ask a follow up question here.

ajcvickers:
@NickCraver Based on my last comment on #7939 (using nominal types for payload) would you still be using the DiagnosticName attribute?

src/EFCore.Relational/Internal/RelationalLoggerExtensions.cs
```diff
@@ -0,0 +1,935 @@
...
+            }
+        }
+
+        private static readonly double _timestampToMilliseconds 
+            = (double)TimeSpan.TicksPerSecond / (Stopwatch.Frequency * TimeSpan.TicksPerMillisecond);
+
+        private static long DeriveTimespan(long startTimestamp, long currentTimestamp)
+            => (long)((currentTimestamp - startTimestamp) * _timestampToMilliseconds);
+    }
+}
```
avanderhoorn: Great stuff!

avanderhoorn:
For my part looks great and includes a ton more than we had before! Re typed vs dynamic objects, I was a part of the original discussions in aspnet around that and fought for types messages but lost that one. So for my part, I'm all for typed objects.

NickCraver:
Sorry, crazy day here - **yes** to typed objects, absolutely. I need to think through a full consumption, but yes: I think we wouldn't need the `const` constraints given all the changes together.

All together this set of changes looks awesome for consumers. Great work!

PR #8291, opened 2017-04-26
Upgrade to System.Data.SqlClient package 4.4.0-*

smitpatel:
Add switch to use native code for localdb

PR #8295, opened 2017-04-26
[2.0.0-preview1] Upgrade to System.Data.SqlClient package 4.4.0-*

smitpatel:
Add switch to use native code for localdb
Port #8291 to 2.0.0-preview1

@ajcvickers @Eilon @muratg for approval.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -14,5 +14,11 @@
     <SystemInteractiveAsyncVersion>3.1.1</SystemInteractiveAsyncVersion>
     <TestSdkVersion>15.0.0</TestSdkVersion>
     <XunitVersion>2.2.0</XunitVersion>
+    <!--
+      * Use 4.4.0-* instead of $(CoreFxVersion) to workaround "SqlClient fails with netcoreapp2.0 on Win7/Server2008"
+      * https://github.com/dotnet/corefx/issues/18406
+      * Remove when $(CoreFxVersion) is upgraded to 4.4.0-*
+    -->
+    <SqlClientVersion>4.4.0-*</SqlClientVersion>
   </PropertyGroup>
 </Project>
--- a/src/EFCore.SqlServer/EFCore.SqlServer.csproj
+++ b/src/EFCore.SqlServer/EFCore.SqlServer.csproj
@@ -16,7 +16,7 @@
     <ProjectReference Include="..\EFCore.Relational\EFCore.Relational.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="System.Data.SqlClient" Version="$(CoreFxVersion)" />
+    <PackageReference Include="System.Data.SqlClient" Version="$(SqlClientVersion)" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.3'">
     <PackageReference Include="System.Threading.Thread" Version="$(CoreFxVersion)" />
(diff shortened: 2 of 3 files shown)


muratg:
cc @mikeharder

mikeharder:
Agreed we cannot ship preview1 until tests pass without `System.Data.SqlClient.UseLegacyNetworkingOnWindows`.

Eilon:
@smitpatel / @ajcvickers - If this gets merged, please log a bug in the 2.0.0-preview1 milestone to track undoing the switch.

mikeharder:
This change was actually merged to dev before preview1.  There are currently some test failures in Win2012-Universe in the dev branch related to EF, so we *should not* merge to preview1 at this time.

I believe the root cause is other tests which use EF, outside the EF repo, would need the `System.Data.SqlClient.UseLegacyNetworkingOnWindows` workaround.  And there may be too many impacted tests for the workaround to be feasible.

A few options:

1. Revert this change in Dev, and don't merge until `SqlClient` itself is updated so we no longer need the `System.Data.SqlClient.UseLegacyNetworkingOnWindows` workaround.
2. Temporarily set `System.Data.SqlClient.UseLegacyNetworkingOnWindows` in EF product code so all users of EF get the workaround.

Eilon:
In theory the bug in SqlClient was fixed late last night, so today we might get a build with it fixed.

smitpatel:
Changed the base.
For the options mentioned by @mikeharder,
I initially tried option 2, changing code in product to set the switch before opening connection but that did not work well. Test were still hanging with it. Probably tests were using connection opening themselves. Changing product code to set the switch may not fix all other repos depending on its test infra & regardless its not guaranteed that user will never open connection themselves without using EF. (User can pass already open connection to EF), hence after discussing with @ajcvickers , we went with only test code changes to make EF pass. This fix does not guarantee that products using EF will not hang until SqlClient is fixed.

As @Eilon mentioned above, since the bug is already fixed, we should remove the switch from this PR and try to get the package with fix ingested in our system as soon as we can.

smitpatel:
Bit unrelated to this change (which only affects netcoreapp2.0), the fix which went in dev branch, I was running it on windows 7 machine. It is stuck while testing SqlServer functional tests on net46 (error similar to appveyor). Its been like that for 15h now. Probably msbuild or xunit hang. fyi: @mikeharder

mikeharder:
Regarding the test hang, we would need to debug.  @smitpatel: If you can keep a live repro around I can investigate later today.

Let's test the latest package at https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.Data.SqlClient.  If it has the fix, I will work with @kichalla to get this version mirrored to our feed.

mikeharder:
I verified that https://github.com/dotnet/corefx/issues/18997 is fixed in `4.4.0-preview2-25227-02` at https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.Data.SqlClient.  I will get this build mirrored to https://dotnet.myget.org/feed/aspnetcore-ci-dev/package/nuget/System.Data.SqlClient, then we should be able to remove the workarounds in EF tests, and all tests should pass.

mikeharder:
I accidentally tested the preview2 build.  I will test with latest preview1 build instead.

mikeharder:
This is also fixed in `System.Data.SqlClient 4.4.0-preview1-25227-04`.  This is the version we will mirror to our feed.

mikeharder:
Packages have been mirrored to https://dotnet.myget.org/gallery/aspnetcore-ci-release, so this change can be checked into preview1 **without** the test workarounds.

mikeharder:
Packages have also been mirrored to https://dotnet.myget.org/gallery/aspnetcore-ci-dev, which should unblock tests in the dev branch.  The EF test workarounds can be reverted in dev as well.

smitpatel:
@mikeharder - Multiple questions:

I am not able to find packages on our feeds neither dev not release.
Will release feed have 4.4.0-preview1 & dev 4.4.0-preview2 respectively or do we have to specific suffix explicitly?

smitpatel:
Updated PR after discussing with Mike.

kichalla:
@smitpatel they should be with `preview1` (we have not yet started mirroring `preview2` and would like to keep it like that at least for this week)

mikeharder:
@kichalla The packages are correctly mirrored, we resolved the confusion.

PR #8296, opened 2017-04-26
[2.0.0-preview1] Core: Initial work on model-level entity type filters

anpete:
Core: Initial work on model-level entity type filters. E.g.
```c#
modelBuilder.Entity<Customer>().HasFilter(c => !c.IsDeleted);
modelBuilder.Entity<Order>().HasFilter(o => o.CustomerId == this._tenantId);
```
Such filters are automatically applied by the query compiler to any query targeting the corresponding EntityType.

Some key aspects of impl. so far:

- Client eval'd expressions are allowed.
- Nav references are not allowed. This is mostly to constrain the impl. cost.
- Evaluatable sub-trees in filter expressions are eval'd early (not parameterized) except for member accesses on DbContext objects. This permits the pattern in the second example above.
- Filters are only allowed on the root of hierarchies, again to constrain the impl. cost.

Open questions:

- Should Find ignore filters when hitting the store?
- We likely need to add an opt-out query operator.
- **Update:** What is required for owned types?

### DIFF
--- a/src/EFCore/Infrastructure/CoreModelValidator.cs
+++ b/src/EFCore/Infrastructure/CoreModelValidator.cs
@@ -4,12 +4,15 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
+using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Query;
 using Microsoft.EntityFrameworkCore.Utilities;
 
 namespace Microsoft.EntityFrameworkCore.Infrastructure
@@ -41,6 +44,80 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
             EnsureChangeTrackingStrategy(model);
             ValidateDelegatedIdentityNavigations(model);
             EnsureFieldMapping(model);
+            ValidateQueryFilters(model);
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected virtual void ValidateQueryFilters([NotNull] IModel model)
+        {
+            Check.NotNull(model, nameof(model));
+            
+            var filterValidatingExppressionVisitor = new FilterValidatingExppressionVisitor();
+
+            foreach (var entityType in model.GetEntityTypes())
+            {
+                if (entityType.Filter != null)
+                {
+                    if (entityType.BaseType != null)
+                    {
+                        ShowError(CoreStrings.BadFilterDerivedType(entityType.Filter, entityType.DisplayName()));
+                    }
+
+                    if (!filterValidatingExppressionVisitor.IsValid(entityType))
+                    {
+                        ShowError(CoreStrings.BadFilterExpression(entityType.Filter, entityType.DisplayName(), entityType.ClrType));
+                    }
+                }
+            }
+        }
+
+        private sealed class FilterValidatingExppressionVisitor : ExpressionVisitor
+        {
+            private IEntityType _entityType;
+
+            private bool _valid = true;
+
+            public bool IsValid(IEntityType entityType)
+            {
+                _entityType = entityType;
+
+                Visit(entityType.Filter.Body);
+
+                return _valid;
+            }
+
+            protected override Expression VisitMember(MemberExpression memberExpression)
+            {
+                if (memberExpression.Expression == _entityType.Filter.Parameters[0]
+                    && memberExpression.Member is PropertyInfo propertyInfo
+                    && _entityType.FindNavigation(propertyInfo) != null)
+                {
+                    _valid = false;
+
+                    return memberExpression;
+                }
+
+                return base.VisitMember(memberExpression);
+            }
+
+            protected override Expression VisitMethodCall(MethodCallExpression methodCallExpression)
+            {
+                if (methodCallExpression.IsEFProperty()
+                    && methodCallExpression.Arguments[0] == _entityType.Filter.Parameters[0]
+                    && (!(methodCallExpression.Arguments[1] is ConstantExpression constantExpression)
+                        || !(constantExpression.Value is string propertyName)
+                        || _entityType.FindNavigation(propertyName) != null))
+                {
+                    _valid = false;
+
+                    return methodCallExpression;
+                }
+
+                return base.VisitMethodCall(methodCallExpression);
+            }
         }
 
         /// <summary>
@@ -49,6 +126,8 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
         /// </summary>
         protected virtual void EnsureNoShadowEntities([NotNull] IModel model)
         {
+            Check.NotNull(model, nameof(model));
+
             var firstShadowEntity = model.GetEntityTypes().FirstOrDefault(entityType => !entityType.HasClrType());
             if (firstShadowEntity != null)
             {
@@ -62,6 +141,8 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
         /// </summary>
         protected virtual void EnsureNoShadowKeys([NotNull] IModel model)
         {
+            Check.NotNull(model, nameof(model));
+
             foreach (var entityType in model.GetEntityTypes().Where(t => t.ClrType != null))
             {
                 foreach (var key in entityType.GetDeclaredKeys())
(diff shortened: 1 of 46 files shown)


src/EFCore/Metadata/Builders/EntityTypeBuilder.cs
```diff
@@ -178,6 +179,19 @@ public virtual EntityTypeBuilder Ignore([NotNull] string propertyName)
         }
 
         /// <summary>
+        ///     Specifies a LINQ predicate expression that will automatically be applied to any queries targeting
+        ///     this entity type.
+        /// </summary>
+        /// <param name="filter">The LINQ predicate expression.</param>
+        /// <returns></returns>
+        public virtual EntityTypeBuilder HasFilter([CanBeNull] LambdaExpression filter)
```
AndriySvyryd: empty returns

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -241,7 +246,7 @@ public virtual ParameterExpression CurrentParameter
                 SingleResultToSequence(queryModel, _expression.Type.GetTypeInfo().GenericTypeArguments[0]);
 
                 TrackEntitiesInResults<TResult>(queryModel);
-
+
```
AndriySvyryd: extra whitespace

AndriySvyryd:
Are you planning to enable filters for owned types or is it scoped out for 2.0?

anpete:
@AndriySvyryd I need to chat with the...owner of that :trollface:

src/EFCore/Query/ExpressionVisitors/Internal/FilterApplyingExpressionVisitor.cs
```diff
@@ -0,0 +1,123 @@
...
+                            _parameters,
+                            _queryCompilationContext.Logger,
+                            parameterize: false,
+                            generateContextAccessors: true);
+
+                    var parameterizedFilter
+                        = (LambdaExpression)visitor.ExtractParameters(entityType.Filter);
+
+                    var mainFromClause
+                        = new MainFromClause(
```
maumar: we use this pattern in several places - consider adding helper

src/EFCore/Query/QueryContext.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.EntityFrameworkCore.Query
     /// <summary>
```
maumar: cleanup usings

src/EFCore/Infrastructure/CoreModelValidator.cs
```diff
@@ -41,6 +41,25 @@ public override void Validate(IModel model)
...
             ValidateDelegatedIdentityNavigations(model);
             EnsureFieldMapping(model);
+            EnsureNoFiltersOnDerivedTypes(model);
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected virtual void EnsureNoFiltersOnDerivedTypes([NotNull] IModel model)
```
smitpatel: validate

src/EFCore/Infrastructure/CoreModelValidator.cs
```diff
@@ -41,6 +41,25 @@ public override void Validate(IModel model)
...
+        {
+            Check.NotNull(model, nameof(model));
+
+            var invalidEntityType
```
smitpatel: derived type with query filter

src/EFCore/Metadata/Internal/EntityType.cs
```diff
@@ -119,6 +122,87 @@ public EntityType([NotNull] Type clrType, [NotNull] Model model, ConfigurationSo
...
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual LambdaExpression Filter
+        {
+            get => _filter;
+            [param: CanBeNull]
+            set
+            {
+                if (value != null
+                    && !new FilterValidatingExppressionVisitor(this).IsValid(value))
```
smitpatel: Instead of initializing every time we set filter, can we make this static and pass entity type as argument?

src/EFCore/Metadata/Internal/EntityType.cs
```diff
@@ -119,6 +122,87 @@ public EntityType([NotNull] Type clrType, [NotNull] Model model, ConfigurationSo
...
+                Visit(filter.Body);
+
+                return _valid;
+            }
+
+            protected override Expression VisitMember(MemberExpression memberExpression)
+            {
+                if (memberExpression.Expression == _targetParameter
+                    && memberExpression.Member is PropertyInfo propertyInfo
+                    && _entityType.FindNavigation(propertyInfo) != null)
```
smitpatel: Since this is checking if it is not set on navigation. Does that mean we allow filter on unmapped properties ?
Also navigation can be absent (in case of multiple navigations in each direction when fluent api configuration is required), so you can set filter in such cases on navigation and then configure relationship, what happen in such cases?
@

smitpatel: @AndriySvyryd

src/EFCore/Query/ExpressionVisitors/Internal/FilterApplyingExpressionVisitor.cs
smitpatel: Its fine either way, Just object initialization.

---
In reply to: [113814275](https://github.com/aspnet/EntityFramework/pull/8296#discussion_r113814275) [](ancestors = 113814275)

src/EFCore/Query/ExpressionVisitors/Internal/FilterApplyingExpressionVisitor.cs
```diff
@@ -0,0 +1,123 @@
...
+
+                return value;
+            }
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitSubQuery(SubQueryExpression subQueryExpression)
```
smitpatel: how about deriving from ExpressionVisitorBase?

src/EFCore/Metadata/Internal/EntityType.cs
AndriySvyryd: We should move this validation to ModelValidation. The filter will not be used during model building anyway.

smitpatel:
Lets discuss about the invalid filters in person.

src/EFCore/Metadata/Internal/EntityType.cs
anpete: Happy to move to validation step.

src/EFCore/Query/ExpressionVisitors/Internal/FilterApplyingExpressionVisitor.cs
anpete: No reason filter ETs are pure CLR ETs.

anpete:
Updated.

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -503,6 +527,57 @@ var queryExecutorExpression
...
+                return null;
+            }
+
+            contextVariableExpression
+                = Expression.Variable(_queryCompilationContext.ContextType, "context");
+
+            var blockExpressions
+                = new List<Expression>
+                {
+                    Expression.Assign(
```
maumar: CreateAssignment helper method should do the work

src/EFCore/Infrastructure/CoreModelValidator.cs
```diff
@@ -41,6 +44,78 @@ public override void Validate(IModel model)
...
+            foreach (var entityType in model.GetEntityTypes())
+            {
+                if (entityType.Filter != null)
+                {
+                    if (entityType.BaseType != null)
+                    {
+                        ShowError(CoreStrings.BadFilterDerivedType(entityType.Filter, entityType.DisplayName()));
+                    }
+
+                    if (!new FilterValidatingExppressionVisitor().IsValid(entityType))
```
smitpatel: save the object in variable outside foreach? Not sure how much it matters though.

src/EFCore/Query/EntityQueryModelVisitor.cs
smitpatel: Lets not use it here. This convert is independent of RemoveConvert.

---
In reply to: [113982711](https://github.com/aspnet/EntityFramework/pull/8296#discussion_r113982711) [](ancestors = 113982711)

smitpatel:
Few questions:
Metadata:
Presently, the filters are via fluent API only. And we are not storing ConfigurationSource. Should we store it just to future guard if we add some attribute or convention for it later. (soft-delete does not sound totally bad convention). It can be YAGNI also in which case this is fine.

Query:
With include compiler and above changes, we have a good bunch of code going through `OptimizeQueryModel` method. Perhaps the method name should be different. IFF user opt-out of query optimization, (first question would be how? 😝 ) what parts should be removed or unavailable.

anpete:
@smitpatel 

1) YAGNI is fine for this.
2) Agreed. I will add an item to query refactoring bug

Issue #7499, opened 2017-01-27
Where() on derived property fails with undefined property on base class exception


rmja:
Consider the following model:
```
public class Message
{
        public int Id { get; private set; }

        public MessageSender Sender { get; private set; }

        public ICollection<MessageRecipient> Recipients { get; private set; }
}
```
`MessageSender` and `MessageRecipient` are abstract classes that have different implementation depending on their kind.

To get the "inbox" I can issue the following query:
```
db.Messages.Where(message =>
                message.Recipients.OfType<UserMessageRecipient>().Any(x => x.UserId == userId) ||
                message.Recipients.OfType<EmailMessageRecipient>().Any(x => x.Email == email))
```

However, to get the "outbox" where the query is run on the Sender property, I cannot 
```
db.Messages.Where(message =>
                (message.Sender is UserMessageSender && ((UserMessageSender)message.Sender).UserId == userId) ||
                (message.Sender is EmailMessageSender && ((EmailMessageSender )message.Sender).Email == email))
```

The following exception is thrown:
```
Exception thrown: 'System.ArgumentException' in mscorlib.dll

Additional information: Property 'System.String UserId' is not defined for type 'Messaging.Models.MessageSender'

   at System.Linq.Expressions.Expression.Property(Expression expression, PropertyInfo property)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.<>c__DisplayClass20_0.<VisitMember>b__2(Expression e)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.RewriteNavigationsIntoJoins(QuerySourceReferenceExpression outerQuerySourceReferenceExpression, IEnumerable`1 navigations, Type propertyType, Func`2 propertyCreator, Func`2 conditionalAccessPropertyCreator)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.NavigationRewritingExpressionVisitor.<>c__DisplayClass20_0.<VisitMember>b__0(IEnumerable`1 ps, IQuerySource qs)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.BindPropertyExpressionCore[TResult](Expression propertyExpression, IQuerySource querySource, Func`3 propertyBinder)
[...]

PR #8300, opened 2017-04-27
[2.0.0-preview1] Fix to #7499 - Where() on derived property fails with undefined property on base class exception

maumar:
Problem was that we were not taking type differences and expression converts in various places during compilation. We sometimes remove converts when processing expressions, but then not accounting for them when we re-assemble the visited expression.
This caused many scenarios involving casts (e.g. accessing properties/navigations on derived types) to fail.

Fix is to start compensating for those where necessary. Created helper methods for EF.Property method call, member access and assignment and using them instead of regular Expression APIs.

Also fixed unrelated issue in the SqlTranslatingExpressionVisitor - when trying to translate MethodCall expression, we would force client evaluation, if arguments to the function was a subquery - this is not ideal, since we can translate some of the subqueries (ones that return single result).

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Query.Expressions;
@@ -375,7 +376,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 }
 
                 if (expression is MethodCallExpression methodCallExpression
-                    && EntityQueryModelVisitor.IsPropertyMethod(methodCallExpression.Method))
+                    && methodCallExpression.Method.IsEFPropertyMethod())
                 {
                     if (methodCallExpression.Arguments[0] is QuerySourceReferenceExpression querySourceCaller)
                     {
@@ -421,7 +422,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
             protected override Expression VisitMethodCall(MethodCallExpression node)
             {
-                if (EntityQueryModelVisitor.IsPropertyMethod(node.Method))
+                if (node.Method.IsEFPropertyMethod())
                 {
                     if (node.Arguments[1] is ConstantExpression propertyNameExpression && (string)propertyNameExpression.Value == _propertyName)
                     {
@@ -579,9 +580,9 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
             {
                 var arguments
                     = methodCallExpression.Arguments
-                        .Where(e => !(e is QuerySourceReferenceExpression)
-                                    && !(e is SubQueryExpression))
-                        .Select(e => (e as ConstantExpression)?.Value is Array || e.Type == typeof(DbFunctions)
+                        .Where(e => !(e.RemoveConvert() is QuerySourceReferenceExpression)
+                                && !IsNonTranslatableSubquery(e.RemoveConvert()))
+                        .Select(e => (e.RemoveConvert() as ConstantExpression)?.Value is Array || e.RemoveConvert().Type == typeof(DbFunctions)
                             ? e
                             : Visit(e))
                         .Where(e => e != null)
@@ -625,6 +626,12 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 ?? _queryModelVisitor.BindMethodToOuterQueryParameter(methodCallExpression);
         }
 
+        private bool IsNonTranslatableSubquery(Expression expression)
+            => expression is SubQueryExpression subQueryExpression
+            && !(subQueryExpression.QueryModel.GetOutputDataInfo() is StreamedScalarValueInfo 
+                || subQueryExpression.QueryModel.GetOutputDataInfo() is StreamedSingleValueInfo streamedSingleValueInfo
+                    && IsStreamedSingleValueSupportedType(streamedSingleValueInfo));
+
         /// <summary>
         ///     Visit a member expression.
         /// </summary>
@@ -636,8 +643,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         {
             Check.NotNull(memberExpression, nameof(memberExpression));
 
-            if (!(memberExpression.Expression is QuerySourceReferenceExpression)
-                && !(memberExpression.Expression is SubQueryExpression))
+            if (!(memberExpression.Expression.RemoveConvert() is QuerySourceReferenceExpression)
+                && !(memberExpression.Expression.RemoveConvert() is SubQueryExpression))
             {
                 var newExpression = Visit(memberExpression.Expression);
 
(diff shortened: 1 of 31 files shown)


maumar:
Ideally we want this in preview, since the codepaths affected are quite common, and it's quite possible that I broke/missed something. Adding @divega @ajcvickers for assessment.

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -519,8 +522,15 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
                                     node.Arguments[0],
                                     (string)((ConstantExpression)node.Arguments[1]).Value,
                                     node.Type,
-                                    e => Expression.Call(node.Method, e, node.Arguments[1]),
-                                    e => new NullConditionalExpression(e, e, Expression.Call(node.Method, e, node.Arguments[1])));
+                                    e => node.Arguments[0].Type != e.Type
```
maumar: I could maybe add helper for MethodCall expressions also, this would simplify this code and potentially safeguard scenarios around converts+methodcalls. Thoughts?

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -68,16 +68,103 @@ public abstract class EntityQueryModelVisitor : QueryModelVisitorBase
...
+        /// <param name="makeNullable"> The property indicating whether property should be made nullable automatically. </param>
         /// <returns> The newly created expression. </returns>
         public static Expression CreatePropertyExpression(
             [NotNull] Expression target,
-            [NotNull] IPropertyBase property)
-            => Expression.Call(
-                EF.PropertyMethod.MakeGenericMethod(property.ClrType.MakeNullable()),
+            [NotNull] IPropertyBase property,
+            bool makeNullable = true)
+            => CreatePropertyExpression(target, property.DeclaringType.ClrType, property.ClrType, property.Name, makeNullable);
```
maumar: is EQMV the right place for those?

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -68,16 +68,103 @@ public abstract class EntityQueryModelVisitor : QueryModelVisitorBase
...
+                Expression.Constant(propertyName));
+        }
+
+        /// <summary>
+        ///     Creates an expression that represents assignment operation.
+        /// </summary>
+        /// <remarks>
+        ///     This method compensates for potential type differences between target expression and propery declaring type.
+        /// </remarks>
+        /// <param name="left"> Expression that will be assigned to. </param>
```
maumar: ESL, any suggestions on making those descriptions less confusing?

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -625,6 +625,12 @@ var projectionIndex
                 ?? _queryModelVisitor.BindMethodToOuterQueryParameter(methodCallExpression);
         }
 
+        private bool IsNonTranslatableSubquery(Expression expression)
+            => expression is SubQueryExpression subQueryExpression
+            && !(subQueryExpression.QueryModel.GetOutputDataInfo() is StreamedScalarValueInfo 
+                || subQueryExpression.QueryModel.GetOutputDataInfo() is StreamedSingleValueInfo streamedSingleValueInfo
```
anpete: Can just do outputDataInfo != StreamedSequenceInfo

anpete:
Any ideas on better fixing Convert issue via VisitUnary I.e. the more standard way.

src/EFCore/Query/EntityQueryModelVisitor.cs
anpete: I was also thinking of moving this along with IsPropertyMethod, which is used elsewhere and looks especially weird. How about EFPropertyExtensions and make them extension methods?

src/EFCore/Query/EntityQueryModelVisitor.cs
anpete: Make them internal :trollface:

ajcvickers:
@maumar I think we should get this in preview1, once @anpete has signed off on it and if he and @divega agrees.

/cc @Eilon This might cause regressions in query, which is why getting as much feedback as possible before RTM is desirable, but I think it is low risk with regards to breaking general infrastructure/build/stability etc.

Eilon:
Ok with me

maumar:

src/EFCore/Query/QueryCompilationContext.cs
```diff
@@ -503,7 +504,7 @@ private static QuerySourceReferenceExpression GetPropertyAccessQsre(Expression e
                 }
 
                 if (expression.RemoveConvert() is MethodCallExpression method
-                    && EntityQueryModelVisitor.IsPropertyMethod(method.Method))
+                    && method.Method.IsPropertyMethod())
```
smitpatel: IsEFPropertyMethod to better clarity

smitpatel:
>             var expression = node.Expression.RemoveConvert();

If we remove this line and implement visit unary (only for convert type)

There are not many occurrences of remove convert in this file, but would like visit unary method.

---
Refers to: src/EFCore/Query/ExpressionVisitors/Internal/MemberAccessBindingExpressionVisitor.cs:245 in c37fb39. [](commit_id = c37fb398e485522e289cafe7e2904f35f848bd63, deletion_comment = False)

src/EFCore/Extensions/Internal/EFPropertyExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+        private static readonly string _efTypeName = typeof(EF).FullName;
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static bool IsPropertyMethod([CanBeNull] this MethodInfo methodInfo) =>
+            Equals(methodInfo, EF.PropertyMethod)
+            // fallback to string comparison because MethodInfo.Equals is not
+            // always true in .NET Native even if methods are the same
```
anpete: I think Equals should just work now and I suspect we can simplify this.

src/EFCore/Extensions/Internal/EFPropertyExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    public static class EFPropertyExtensions
+    {
+        private static readonly string _efTypeName = typeof(EF).FullName;
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static bool IsPropertyMethod([CanBeNull] this MethodInfo methodInfo) =>
```
anpete: Maybe add a version that accepts Expression or MethodCallExpression, too.

src/EFCore/Query/QueryCompilationContext.cs
anpete: Same with the other methods.

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
smitpatel: Why do these types differ?

src/EFCore/Extensions/Internal/EFPropertyExtensions.cs
maumar: still broken :(

divega:
I agree we should get this in preview1 if possible.

src/EFCore/Extensions/Internal/EFPropertyExtensions.cs
anpete: You tested it on .NET Native?

maumar: I goofed, will update this in next checkin

Issue #8164, opened 2017-04-13
Design: Discover DbContext types using Program.BuildWebHost()


bricelam:
If ASP.NET Core adds a well known, static accessor to the `IWebHost`, our tools can use this to instantiate the `DbContext` at runtime.

We may also want to reconsider some aspects of the previous `Startup`-based mechanism which leaked a bit into our codebase. Specificly: environment and content root directory. This includes the `--environment` option on `dotnet ef` and `ef.exe`, the `environment` and `contentRootPath` parameters on `OperationExecutor`, and the `ContentRootPath` (see also #8159) and `EnvironmentName` properties on `DbContextFactoryOptions. Issue #7480 is also probably superseded by this.

We may also want to flow arbitrary command-line arguments into the static entry point and `IDbContextFactory`. But I'm not sure how we'll handle the transition from PowerShell to command-line...

This work depends on #7588

bricelam:
As part of revisiting `DbContextFactoryOptions`, we can remove `ContentRootPath` (this was tracked by #8159)

bricelam:
#7353 may also be related to this work.

bricelam:
We currently allow a `ConfigureDesignTimeServices` method to be defined on the startup class. We should consider removing that convention and only allow implementations of `IDesignTimeServices`.

bricelam:
Questions we need to answer:
* Can we remove `--environment` and `DbContextFactoryOptions.EnvironmentName`? This would make Environment a Hosting-only concern.
* Can we remove `DbContextFactoryOptions.ContentRootPath`? This would make ContentRootPath a Hosting-only concern.
* Can we remove the logic that attempts to call `<startup>.Configure<environment>Services` directly?
* Do we try and call `<startup-class>.ConfigureDesignTimeServices` when using `Program.BuildWebHost`?
* How do we flow arguments into `BuildWebHost` from dotnet-ef/ef? From PMC?

bricelam:
Decisions:
* Remove `--environment`
* Remove `DbContextFactoryOptions`
* Require `IDesignTimeServices` to be implemented before calling `ConfigureDesignTimeServices`
* For now, arguments will be empty at design-time. Add `IDbContextFactory.Create(string[] args)` to enable flowing in the future.

Still need to discuss this:
* Can we remove the logic that attempts to call `<startup>.Configure<environment>Services` directly? (about 600 lines of code)

bricelam:
[...]

PR #8307, opened 2017-04-27
[2.0.0-preview1] Tools: Remove --environment

### LINKED ISSUE #7353: CLI Tools: Honor ASPNETCORE_ENVIRONMENT
_From @opensrcken on November 2, 2016 7:57_

Representative of the script I'm using when I encounter said bug.

```
SET ASPNETCORE_ENVIRONMENT=Test
dotnet ef database update
```

`env.EnvironmentName` does not not seem to be aware of `ASPNETCORE_ENVIRONMENT` when running the second command. Instead, it is always set to `Development`.

_Copied from original issue: dotnet/core#327_

bricelam:
Use ASPNETCORE_ENVIRONMENT instead.

Part of #8164, fixes #7353


riscie:
Short question. Wouldn't it have been possible to allow both? using ASPNETCORE_ENVIRONMENT by default but override it via --environment. 

Now when I want to run a migration on our staging environment, but do so from my dev-machine, I need to alter the ASPNETCORE_ENVIRONMENT. Before I could just do '... -e staging'. Or am I missing something?

bricelam:
In 1.x, `--environment` was a general concept that could be used outside of ASP.NET Core.

In 2.0, we removed the concept of an environment from EF Core. We just call into Hosting which happens to read ASPNETCORE_ENVIRONMENT.

So yes, we could have, but it wouldn't make any sense outside of ASP.NET Core and it would couple the EF Core Tools to Hosting.

riscie:
Thank you for elaborating @bricelam

PR #8310, opened 2017-04-27
[2.0.0-preview1] Design: Maintain Scaffolding compatibility

bricelam:


src/EFCore.Design/Design/Internal/DatabaseOperations.cs
```diff
@@ -21,6 +21,18 @@ public class DatabaseOperations
         private readonly string _rootNamespace;
         private readonly DesignTimeServicesBuilder _servicesBuilder;
 
+        // This obsolete constructor maintains compatibility with Scaffolding
+        public DatabaseOperations(
```
bricelam: :facepalm: this was supposed to be in `DbContextOperations`

bricelam: Fixed in a664dd13d

PR #8314, opened 2017-04-27
[2.0.0-preview1] Remove DbContextFactoryOptions

bricelam:
Part of #8164


src/EFCore/Infrastructure/IDbContextFactory.cs
```diff
@@ -19,8 +19,8 @@ public interface IDbContextFactory<out TContext>
         /// <summary>
         ///     Creates a new instance of a derived context.
         /// </summary>
-        /// <param name="options"> Information about the environment an application is running in. </param>
+        /// <param name="args"> Arguments provided by the design-time service. </param>
         /// <returns> An instance of <typeparamref name="TContext" />. </returns>
-        TContext Create([NotNull] DbContextFactoryOptions options);
+        TContext Create([NotNull] string[] args);
```
bricelam: Should this be `IEnumerable<string>`?

ajcvickers: Yeah, I think so.

divega: Sounds good.

bricelam: Should I call ToArray in the templates or push for IEnumerable all the way down?

bricelam: There might also be value in keeping an array to mirror Main and drive home the point that this is an entry point into the application

bricelam: It looks like both [WebHost.CreateDefaultBuilder](https://github.com/aspnet/MetaPackages/blob/rel/2.0.0-preview1/src/Microsoft.AspNetCore/WebHost.cs#L150) and [IConfigurationBuilder.AddCommandLine](https://github.com/aspnet/Configuration/blob/rel/2.0.0-preview1/src/Microsoft.Extensions.Configuration.CommandLine/CommandLineConfigurationExtensions.cs#L21) (which it calls) are `string[]`. Given that, I think we should keep it `string[]`.

ajcvickers: I'm fine with keeping an array for that reason.

PR #8317, opened 2017-04-28
Add table splitting

AndriySvyryd:


src/EFCore.Relational/Internal/RelationalAnnotationProviderExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace Microsoft.EntityFrameworkCore.Internal
+{
+    public static class RelationalAnnotationProviderExtensions
```
ajcvickers: Add API doc boilerplate code for internals.

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -20,7 +21,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
     ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
     ///     directly from your code. This API may change or be removed in future releases.
     /// </summary>
-    public class MigrationsModelDiffer : IMigrationsModelDiffer
+    public partial class MigrationsModelDiffer : IMigrationsModelDiffer
```
ajcvickers: Why partial?

src/EFCore.Relational/Properties/RelationalStrings.resx
```diff
@@ -240,10 +240,16 @@
...
+    <value>Cannot use table '{table}' for entity type '{entityType}' since it is being used for entity type '{otherEntityType}' and {specificError}.</value>
     <comment>Schema can be null</comment>
   </data>
+  <data name="SharedTableKeyNameMismatchFragment" xml:space="preserve">
+    <value>the name '{keyName}' of the primary key {primaryKey} on entity type '{entityType}' does not match the name '{otherName}' of the primary key {otherPrimaryKey} on entity type '{otherEntityType}'</value>
+  </data>
+  <data name="SharedTableNoRelationshipFragment" xml:space="preserve">
+    <value>there is no relationship between the primary key {primaryKey} on entity type '{entityType}' and the primary key {otherPrimaryKey} on entity type '{otherEntityType}'</value>
+  </data>
   <data name="KeyHasDefaultValue" xml:space="preserve">
```
ajcvickers: Generally not a good idea when these things are localized because it restricts that way expressions can be arranged in different languages. We should just create different full sentences.

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
AndriySvyryd: Because R#. Will revert

bricelam:
I skimmed through the Migrations changes, and they look good to me.

PR #8323, opened 2017-04-28
React to Logging API changes

BrennanConroy:


BrennanConroy:
Meant to open againt rel/2.0.0-preview1

src/EFCore.Design/Design/Internal/DatabaseOperations.cs
```diff
@@ -58,8 +58,8 @@ public class DatabaseOperations
 
             var services = _servicesBuilder.Build(provider);
 
-            var loggerFactory = services.GetService<ILoggerFactory>();
-            loggerFactory.AddProvider(new LoggerProvider(categoryName => new OperationLogger(categoryName, _reporter)));
+            var loggerFactory = services.GetService<ILoggerFactory>() as LoggerFactory;
+            loggerFactory?.AddProvider(new LoggerProvider(categoryName => new OperationLogger(categoryName, _reporter)));
```
Eilon: Hmm this is suspicious... what if it was a customer ILoggerFactory? What should happen?

BrennanConroy: We could allow it to call the ILoggerFactory.AddProvider method and just pragma the warning, then remove it in 3.0.0 when the API gets deleted.

smitpatel:
PR in logging repo for changes?

BrennanConroy:
https://github.com/aspnet/Logging/pull/605

src/EFCore.Design/Design/Internal/DatabaseOperations.cs
ajcvickers: What is the underlying change that results in this change being needed?

BrennanConroy: https://github.com/aspnet/Logging/pull/605/files#diff-528cfd1fcf044b63339b2b84f3b77f84R28

bricelam: > when the API gets deleted

This functionality seems pretty fundamental to a logging abstraction. What's the story going forward? Is the `ILoggerFactory` interface going away in favor of making `LoggerFactory` a public type in `Microsoft.Extensions.Logging.Abstractions`?

ajcvickers:
@BrennanConroy @Eilon Where can we find the design rationale around these changes and documentation or other information about how the logging abstractions are now intended to be used?

Eilon:
I'll set up a mtg for Monday.

BrennanConroy:
This issue is also a good reference for reasons behind the change https://github.com/aspnet/Logging/issues/420

BrennanConroy:
🆙 📅

PR #8326, opened 2017-04-28
[2.0.0-preview1] Design: Stop invoking Startup.ConfigureServices

bricelam:
Part of #8164

Note, this removes to following closure of dependencies form EFCore.Design.
* System.Globalization.Extensions
* System.Net.WebSockets
* System.Reflection.TypeExtensions
* System.Security.Claims
* System.Security.Principal
* System.Text.Encodings.Web
* Microsoft.AspNetCore.Hosting.Abstractions
* Microsoft.AspNetCore.Hosting.Server.Abstractions
* Microsoft.AspNetCore.Http.Abstractions
* Microsoft.AspNetCore.Http.Features
* Microsoft.Extensions.FileProviders.Abstractions
* Microsoft.Extensions.Hosting.Abstractions

**Providers:** If you're not already implementing `IDesignTimeServices` in the type pointed to by `[DesignTimeProviderServices]`, you'll need to after this.


smitpatel:
Correct base branch?

bricelam:
@smitpatel Yes, but only for review... 😉 It's based on #8314

bricelam:
> What exception message does the user get if they still have the old-style ConfigureServices app?

Most will get the typical error:

> No parameterless constructor was found on 'ApplicationDbContext'. Either add a parameterless constructor to 'ApplicationDbContext' or add an implementation of 'IDbContextFactory&lt;ApplicationDbContext&gt;' in the same assembly as 'ApplicationDbContext'.

There is a corner case that can happen on the first `Add-Migration` when your context has been added to the application services, but isn't in either the startup or target project. These advanced users will receive a no context/can't find context error, but the solution is the same--add an `IDbContextFactory`.

ajcvickers:
@bricelam Think we need a better message that indicates that this has happened not because they did something "wrong" but because we have made a change in how things work in 2.0. Probably with a fwlink to docs on how to convert. But happy for all this to happen later.

PR #8330, opened 2017-04-28
Allow shadow keys created explicitly

### LINKED ISSUE #7238: Tools: Confusing warning message "The key {'PropertyName'} on entity type 'TypeName' contains properties in shadow state"
I have migrated from RC2 and now I am having this warning when I generate migrations:

`The key {'Id'} on entity type 'ExtraDeliveryCostByVolume' contains properties in shadow state - {'Id'}. To configure this warning use the DbContextOptionsBuilder.ConfigureWarnings API (event id 'CoreEventId.ModelValidationWarning'). ConfigureWarnings can be used when overriding the DbContext.OnConfiguring method or using AddDbContext on the application service provider.`

I understand that `Id` property is shadow, it is true, but, why is it a warning? is there any problem on having a shadow property as primary key?

The model is configured in the follow way:

```c#
  {
    . . .
    modelBuilder.Entity<ExtraDeliveryCostByVolume>(entity =>
    {
      entity.Property<int>("Id").UseSqlServerIdentityColumn();
      entity.Property(x => x.ExtraDeliveryCost).HasColumnType("decimal(12,8)");
      entity.HasKey("Id");
      . . .
```

### Further technical details

EF Core version: 1.1
IDE: Visual Studio 2015

smitpatel:
Allow shadow PKs created by convention (which will happen only for DIET)
Throw for shadow alternate key created by convention which is principal end of relationship

Resolves #7238

### DIFF
--- a/src/EFCore/Infrastructure/CoreModelValidator.cs
+++ b/src/EFCore/Infrastructure/CoreModelValidator.cs
@@ -66,30 +66,28 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
             {
                 foreach (var key in entityType.GetDeclaredKeys())
                 {
-                    if (key.Properties.Any(p => p.IsShadowProperty))
+                    if (key.Properties.Any(p => p.IsShadowProperty)
+                        && key is Key concreteKey
+                        && ConfigurationSource.Convention.Overrides(concreteKey.GetConfigurationSource())
+                        && !key.IsPrimaryKey())
                     {
                         var referencingFk = key.GetReferencingForeignKeys().FirstOrDefault();
-                        var conventionalKey = key as Key;
-                        if (!key.IsPrimaryKey()
-                            && referencingFk != null
-                            && conventionalKey != null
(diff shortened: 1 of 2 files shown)


src/EFCore/Infrastructure/CoreModelValidator.cs
```diff
@@ -68,28 +68,28 @@ protected virtual void EnsureNoShadowKeys([NotNull] IModel model)
...
-                                    ? ""
-                                    : "." + referencingFk.PrincipalToDependent.Name),
-                                Property.Format(referencingFk.Properties, includeTypes: true),
-                                Property.Format(entityType.FindPrimaryKey().Properties, includeTypes: true)));
-                            continue;
-                        }
+                            var referencingFk = key.GetReferencingForeignKeys().FirstOrDefault();
 
-                        Dependencies.Logger.ModelValidationShadowKeyWarning(entityType, key);
+                            if (!key.IsPrimaryKey()
```
AndriySvyryd: Move the `!key.IsPrimaryKey()` clause to the first `if` for marginal perf gain

Issue #3736, opened 2015-11-12
Query :: some queries with Queryable.Cast<T> don't work


maumar:
Repro:

```
var query = from Post p in ctx.Posts
            select p.PostId;

var result = query.ToList();
```

the compiler introduces a Cast<Post>() (because type of "p" has been explicitly specified) relinq turns this into a subquery like so:

```
from p in  { from p' in ctx.Post select p' => Cast<Post>() } select p.PostId
```

then we lift the subquery:

```
from p in ctx.Posts select p.PostId => Cast<Post>()
```

this fails with the following:

Unhandled Exception: System.InvalidOperationException: No coercion operator is defined between types 'System.Int32' and 'Repro3499.Post'.
   at System.Linq.Expressions.Expression.GetUserDefinedCoercionOrThrow(ExpressionType coercionType, Expression expression, Type convertToType)
   at System.Linq.Expressions.Expression.Convert(Expression expression, Type type, MethodInfo method)
   at Remotion.Linq.Clauses.ResultOperators.CastResultOperator.GetOutputDataInfo(IStreamedDataInfo inputInfo)
   at Remotion.Linq.QueryModel.<GetOutputDataInfo>b__0(IStreamedDataInfo current, ResultOperatorBase resultOperator)
   at System.Linq.Enumerable.Aggregate[TSource,TAccumulate](IEnumerable`1 source, TAccumulate seed, Func`3 func)
   at Remotion.Linq.QueryModel.GetOutputDataInfo()
   at Microsoft.Data.Entity.Query.EntityQueryModelVisitor.SingleResultToSequence(QueryModel queryModel) in D:\k\EntityFramework\src\EntityFramework.Core\Query\EntityQueryModelVisitor.cs:line 249
   at Microsoft.Data.Entity.Query.EntityQueryModelVisitor.CreateQueryExecutor[TResult](QueryModel queryModel) in D:\k\EntityFramework\src\EntityFramework.Core\Query\EntityQueryModelVisitor.cs:line 163
   at Microsoft.Data.Entity.Storage.Database.CompileQuery[TResult](QueryModel queryModel) in D:\k\EntityFramework\src\EntityFramework.Core\Storage\Database.cs:line 68
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #8333, opened 2017-04-29
[SMALL] Fix to #3736 - Query :: some queries with Queryable.Cast<T> don't work

maumar:
Problem was that during optimization, when we flatten subqueries we didn't account for Cast result operator properly. This would fail when trying to access a property on a derived type, when derived type was coming from a subquery with Cast result operator.
Fix is to wrap qsres to that (former) subquery with Convert(s) the subquery was casted into.

Also fixed another issue around this area:

We tried to lift Distinct in some cases, which could lead to incorrect results. E.g

from outer in (ctx.Customers.Distinct())
select outer.Name

would get translated to:

(from outer in ctx.Customers
select outer.Name).Distinct()

### DIFF
--- a/src/EFCore/Query/Internal/QueryOptimizer.cs
+++ b/src/EFCore/Query/Internal/QueryOptimizer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Microsoft.EntityFrameworkCore.Internal;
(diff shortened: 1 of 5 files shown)


src/EFCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -210,14 +224,24 @@ var fromClauseData
 
                 fromClause.CopyFromSource(fromClauseData);
 
+                var newExpression = subQueryExpression.QueryModel.SelectClause.Selector;
+                var castResultOperatorTypes = subQueryModel.ResultOperators.OfType<CastResultOperator>().Select(cre => cre.CastItemType).ToList();
+                foreach (var type in castResultOperatorTypes)
```
maumar: or should I just apply last Cast operator? see #8334 for some more context

src/EFCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -175,6 +175,30 @@ var additionalFromClause
...
+                        || ro is ReverseResultOperator
+                        || ro is SkipResultOperator
+                        || ro is TakeResultOperator
+                        || ro is UnionResultOperator);
+
+            // we can lift distinct however if the outer query has result operator that doesn't care about having correct element count
+            var emptyQueryMpdelWithResultOperatorThatIgnoresDistinctAndDistinctInSubquery
+                = !queryModel.BodyClauses.Any()
+                && subQueryModel.ResultOperators.Any(ro => ro is DistinctResultOperator)
+                && queryModel.ResultOperators.Any(
```
maumar: any other operators?

maumar:
ping @anpete @smitpatel

src/EFCore/Query/Internal/QueryOptimizer.cs
```diff
@@ -175,6 +176,32 @@ var additionalFromClause
...
+                        || ro is ExceptResultOperator
+                        || ro is IntersectResultOperator
+                        || ro is OfTypeResultOperator
+                        || ro is ReverseResultOperator
+                        || ro is SkipResultOperator
+                        || ro is TakeResultOperator
+                        || ro is UnionResultOperator);
+
+            // we can lift distinct however if the outer query has result operator that doesn't care about having correct element count
+            var emptyQueryMpdelWithResultOperatorThatIgnoresElementCountAndDistinctInSubquery
```
smitpatel: typo

smitpatel:
Rebase on latest dev

maumar:
new version up @smitpatel

maumar:
ping x2 @smitpatel @anpete

PR #8339, opened 2017-04-29
[2.0.0-preview1] Core: Initial work on model-level entity type filters. E.g.

anpete:
Cherry-picking filters from dev into preview1.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
     <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />
--- a/src/EFCore.Relational.Specification.Tests/FromSqlSprocQueryTestBase.cs
+++ b/src/EFCore.Relational.Specification.Tests/FromSqlSprocQueryTestBase.cs
@@ -221,17 +221,13 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             }
         }
 
-        protected NorthwindContext CreateContext() => Fixture.CreateContext();
+        protected NorthwindContext CreateContext() => Fixture.CreateContext(QueryTrackingBehavior.NoTracking);
 
-        protected FromSqlSprocQueryTestBase(TFixture fixture)
-        {
-            Fixture = fixture;
-        }
+        protected FromSqlSprocQueryTestBase(TFixture fixture) => Fixture = fixture;
 
         protected TFixture Fixture { get; }
 
         protected abstract string TenMostExpensiveProductsSproc { get; }
-
         protected abstract string CustomerOrderHistorySproc { get; }
     }
 }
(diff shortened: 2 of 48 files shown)


anpete:
@Eilon @ajcvickers for approval. This PR adds the new model-level filters feature to preview1.

Eilon:
This is of course quite a big change. I assume we want it in Preview1 in order to get broader feedback? Does it risk regressing anything, or is it largely a new independent feature?

anpete:
> I assume we want it in Preview1 in order to get broader feedback

Yep. Specifically requested by @divega 😄 And, yes, it is a new independent feature.

Eilon:
That sounds fine to me.

ajcvickers:
@anpete Are there any tests for doing Migrations when there is a query expression in the model. Does it get ignored in the model snapshot? Does it ever cause differ issues?

anpete:
@ajcvickers No, I'm not sure how a completely new metadata element can break Migrations. @bricelam Anything to worry about here?

ajcvickers:
@anpete Approving for preview.

bricelam:
> I'm not sure how a completely new metadata element can break Migrations. @bricelam Anything to worry about here?

If it doesn't affect the database and isn't a non-serializable annotation, it should be fine. If there's ModelBuilder API, it *should* be serialized into the model snapshot, but doesn't technically need to be.

ajcvickers:
@bricelam So our approach is to serialize into the model snapshot even if it doesn't impact that s-space model?

bricelam:
Yeah, we wanted the snapshot code to be general purpose.

smitpatel:
Just asking for my own understanding,

@bricelam - So basically model snapshot is kind of serialized form of c-space model (not always accurate though). Based on snapshot and current model, we determine the diff. But diff is used to generate database operations so whatever is in model which is not affecting database is ignored in diff. (e.g. I did not see any processing for base type). Based on diff we generate the migration file which represents only changes required on s-space. So even if your c-space model changes you still can get empty migration file.

For Filters, not sure how much it matters. They are not annotations so they won't be put into snapshot. In a way, your current model always have extra filters which are missing in snapshot but differ doesn't generate anything for filters so things work regardless.

bricelam:
Correct.

I think it would be ok to say that filters are a query-only aspect of the model and thus don't need to be put in the snapshot.

I don't foresee any non-schema-related tooling using the snapshot.

As a component, it is supposed to be a fully-specified, unbound version of your model. In practice, we only care about the resulting database schema and any ambient data we can use to diff.

anpete:
> I don't foresee any non-schema-related tooling using the snapshot.

One think might be for "compiled state" models.

Another interesting thing about filters is that they would only be easily representable in the current model. I.e. They often have a strong binding to CLR types and so we would likely need to rewrite them in terms of late bound constructs.

ajcvickers:
My main concern originally was that there were annotations that wouldn't be ignored but also would not be serialized correctly. At this point I think we need to YAGNI things that we can't see having any impact on the database. Compiled state is interesting, but until we actually do it I don't know that we know what it will really require--my gut feeling is that it will need a different form of model serialization anyway.

Issue #7516, opened 2017-01-31
Query: Compiler crash for invalid Include path


jnm2:
([As asked](https://github.com/aspnet/EntityFramework/issues/7424#issuecomment-276194553) by @anpete.)

Please support eager loading when the root entity ends up in a projection.

There is not a good way to do what I need to do: load a list of entities with multiple child collections on each entity populated, plus an arbitrary calculation associated to each entity.

This would be perfect:
```c#
var results = dbContext.Entities
    .Select(entity => new
    {
        Entity = entity,
        ComplicatedServerSideCalculation = Entity.NavProperty.OtherInfo.Sum(x => x * 2)
    })
    .Include(info => info.Entity.ChildCollection1)
    .Include(info => info.Entity.ChildCollection2)
    .Include(info => info.Entity.ChildCollection3)
    .ToListAsync();
```

Intuitively, I'd also expect that this would work:
```c#
var results = dbContext.Entities
    .Include(entity => entity.ChildCollection1)
    .Include(entity => entity.ChildCollection2)
    .Include(entity => entity.ChildCollection3)
    .Select(entity => new
    {
        Entity = entity,
        ComplicatedServerSideCalculation = Entity.NavProperty.OtherInfo.Sum(x => x * 2)
    })
    .ToListAsync();
```

rowanmiller:
Have you checked if the second code listing works or not? We think that will already work.

Also, what happens with the first example (exception, ignored, etc.)?

jnm2:
@rowanmiller You're right about the second code sample! I'm surprised and happy! Given the response to the [exact same question for EF6](https://entityframework.codeplex.com/workitem/2939) I thought this was a path you decided against. This will be a major win for us when we finally get to switch to EF Core.

As for the first code sample with EF Core, I get:

```
System.InvalidCastException occurred
  HResult=0x80004002
  Message=Unable to cast object of type 'System.Linq.Expressions.NewExpression' to type 'System.Linq.Expressions.MemberExpression'.
  Source=Microsoft.EntityFrameworkCore
  StackTrace:
   at Microsoft.EntityFrameworkCore.Query.ResultOperators.Internal.IncludeResultOperator.GetQuerySource(MemberExpression expression)
   at Microsoft.EntityFrameworkCore.Query.ResultOperators.Internal.IncludeResultOperator.GetQuerySource(MemberExpression expression)
[...]

PR #8341, opened 2017-04-30
[TINY] Modification to fix for #7516 - Query: Compiler crash for invalid Include path

maumar:
Instead of ignoring includes that are not specified directly on entity type we throw NotSupportedException

### DIFF
--- a/src/EFCore/Properties/CoreStrings.resx
+++ b/src/EFCore/Properties/CoreStrings.resx
@@ -631,7 +631,7 @@
     <value>A parameterless constructor was not found on entity type '{entityType}'. In order to create an instance of '{entityType}' EF requires that a parameterless constructor be declared.</value>
   </data>
   <data name="LogIgnoredInclude" xml:space="preserve">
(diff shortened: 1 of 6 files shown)


src/EFCore/Properties/CoreStrings.resx
```diff
@@ -741,4 +741,7 @@
   <data name="BadFilterDerivedType" xml:space="preserve">
     <value>The filter expression '{filter}' cannot be specified for entity type '{entityType}'. A filter may only be applied to the root entity type in a hierarchy.</value>
   </data>
+  <data name="IncludeNotSpecifiedDirectlyOnEntityType" xml:space="preserve">
+    <value>Include operation '{operation}' can only be specified directly on an entity type.</value>
```
anpete: This may be confusing because Customer is an ET.

How about something like: "The Include operation '{include}' was invalid. All components of an Include load path must entity types.

divega: Could we say it is because `Item1` is not a navigation property defined in an entity type?

PR #8345, opened 2017-05-01
Reverting SQLitePCLRawVersion version back to 1.1.3 (issue: https:/…

kichalla:
…/github.com/ericsink/SQLitePCL.raw/issues/157)


kichalla:
We are going to use 1.1.5 instead. So closing this now.

PR #8353, opened 2017-05-02
Tests: Clean up cross-targeting

### LINKED ISSUE #8250: Specification Tests: Target netcoreapp2.0
Currently, our test specs target net46 and netstandard1.3, but we only ever test on net46 and netcoreapp2.0.

* EFCore.Specification.Tests
* EFCore.Relational.Specification.Tests
* EFCore.Relational.Design.Specification.Tests

We should remove the netstandard1.3 target and add netcoreapp2.0 instead. This would enable us in our tests to take advantage of some fixes and features that have been added.

bricelam:
Resolves #8250

### DIFF
--- a/src/EFCore.Relational.Design.Specification.Tests/EFCore.Relational.Design.Specification.Tests.csproj
+++ b/src/EFCore.Relational.Design.Specification.Tests/EFCore.Relational.Design.Specification.Tests.csproj
@@ -2,8 +2,8 @@
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
     <Description>Shared design-time test suite for Entity Framework Core relational database providers.</Description>
-    <TargetFrameworks>net46;netstandard1.6</TargetFrameworks>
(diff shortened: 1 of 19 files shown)


test/EFCore.Benchmarks.EF6/EFCore.Benchmarks.EF6.csproj
```diff
@@ -4,9 +4,6 @@
     <TargetFramework>net46</TargetFramework>
     <AssemblyName>Microsoft.EntityFrameworkCore.Benchmarks.EF6</AssemblyName>
     <RootNamespace>Microsoft.EntityFrameworkCore.Benchmarks.EF6</RootNamespace>
-    <!-- TODO: Remove when Microsoft/vstest#428 is fixed. -->
-    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
```
bricelam: We might still need these...

PR #8359, opened 2017-05-02
[WIP] Target .NET Standard 2.0

bricelam:
Blocked on mirroring the 4.4.0 corefx packages
Blocked on API Check (throws *Could not load assembly 'netstandard'.*) cc @javiercn

Open question:
* Should we test on `netcoreapp1.1` to cover the `netstandard1.3` cross-targeted assemblies?


Eilon:
BTW we should make sure this works properly with MusicStore before committing to it. I believe there have been all sorts of issues trying to get everything to run w/ NS2.0. We should make sure those are all resolved before trying to commit this.

bricelam:
This was just a tracer bullet exercise. Most of the net46 code just works. I need to exclude the |DataDirectory| tests, but​ other than that, things are passing.

bricelam:
Tests pass

bricelam:
Closing. We need to finalize our target frameworks.

Issue #8364, opened 2017-05-03
Query: comparing collection navigations to null or to each other doesn't work


maumar:
We support comparisons of qsres/reference navigations to nulls and to each other via ```EntityEqualityRewritingExpressionVisitor```, however collection navigations are not handled and cause exceptions later in the query compilation phase.

We can recognize a few patterns to make the query models much easier for us to translate:

1.) ```c.Customers == null``` --> ```c == null``` --> ```c.Id == null```
2.) ```c1.Customers == c2.Customers``` --> ```c1 == c2``` --> ```c1.Id == c2.Id```

For both cases we should issue a warning that this may not be intended behavior (e.g. c.Orders== null, could be intended to check if the customer has any orders)

maumar:
related to #5825

maumar:
fixed in c4129dc2bef433f5ce55a21ec7a417ba8f153752

Issue #5825, opened 2016-06-21
Query: Comparing collection navigation to NULL throws Null Exception


Feasoron:
### Steps to reproduce

Have a many to many relationship with a lookup table defined (the odd ID on the lookup is due legacy data we can't change. I don't think it is the issue, but I included it just to be safe)

```
    public class TableA
    {
        [Key]
        public int TableAId { get; set; }
        public virtual List<TableATableBLookup> TableATableBLookups{ get; set; }
    }

   public class TableATableBLookup
    {
        public int TableATableBLookupId { get; set; }

        public int TableAId { get; set; }
        public TableA{ get; set; }

        public String TableBId{ get; set; }
        public TableB{ get; set; }

        [Column("TableB_FK")]
        public int TableBKey { get; set; }
    }

   public class TableBId
    {
        [Key]
        [Column(TypeName = "nvarchar(450)")]
        public String TableBId{ get; set; }

        public virtual List<TableATableBLookup> TableATableBLookups{ get; set; }

        [Column("TableB_PK")]
        public int TableBKey { get; set; }   
    }

```

Then I simply try to do a lookup of rows on table a that have at least one row in table b

```
var rows= _context.TableA
                .Include(l => l.TableATableBLookups)
                .Where(b => b.TableATableBLookups!= null);
```

When this query is executed, I get an exception
### The issue

Exception message:
`Value cannot be null. Parameter name: querySource`
Stack trace:

```
   at Microsoft.EntityFrameworkCore.Utilities.Check.NotNull[T](T value, String parameterName)
   at Microsoft.EntityFrameworkCore.Query.RelationalQueryModelVisitor.AddQuery(IQuerySource querySource, SelectExpression selectExpression)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.RelationalEntityQueryableExpressionVisitor.VisitEntityQueryable(Type elementType)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.EntityQueryableExpressionVisitor.VisitConstant(ConstantExpression constantExpression)
   at System.Linq.Expressions.ConstantExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.ExpressionVisitorBase.Visit(Expression node)
[...]

Issue #8255, opened 2017-04-21
The generated identifiers in multi level queries are too long for SQLServer


AndriySvyryd:
Paste into ComplexNavigationsQueryTestBase, run in ComplexNavigationsQuerySqlServerTest:

```C#
[ConditionalFact]
public virtual void Level4_Include()
{
    using (var context = CreateContext())
    {
        var query = context.LevelOne
            .Select(l1 => l1.OneToOne_Required_PK).Where(t => t != null)
            .Select(l2 => l2.OneToOne_Required_PK).Where(t => t != null)
            .Select(l3 => l3.OneToOne_Required_PK).Where(t => t != null)
            .Select(l4 => l4.OneToOne_Required_FK_Inverse.OneToOne_Required_FK_Inverse)
            .Include(l2 => l2.OneToOne_Optional_FK);

        Assert.Equal(10, query.ToList().Count);
    }
}
```

Exception: `The identifier that starts with 'l1.OneToOne_Required_PK.OneToOne_Required_PK.OneToOne_Required_PK.OneToOne_Required_FK_Inverse.OneToOne_Required_FK_Inverse.OneT' is too long. Maximum length is 128.`

roji:
I've just run into this issue with PostgreSQL, which limits identifies to 63 characters (I see you set MaxTableAliasLength to 128). This made ComplexNavigationsQueryNpgsqlTest.Level4_Include fail.

Is setting MaxTableAliasLength in my QueryCompilationContext to 63 enough to get around this issue?

smitpatel:
@roji - That should do the trick. That property is overridable so that providers can provide max length and the function to unique-fy table names would strip down according to that.

roji:
@smitpatel ok thanks!

grahamehorner:
FYI , the same query with the same Oracle database works when using ef core 2.0? will this be fixed in previous ef core versions?

smitpatel:
@grahamehorner - The same query will work with version 2.0 compatible provider version without any changes in user code or database. Though this will not be fixed in previous versions.

Issue #8254, opened 2017-04-21
Null semantics are not carried from subquery


AndriySvyryd:
Paste into ComplexNavigationsQueryTestBase, run in ComplexNavigationsQuerySqlServerTest:
```C#
[ConditionalFact]
public virtual void Null_semantics_bug_8254()
{
    AssertQueryScalar<Level1, int>(
        l1s =>
            from l1 in l1s
            join l2Left in
                from l1p in l1s where l1p.OneToOne_Required_PK != null select l1p.OneToOne_Required_PK
                on l1.Id equals l2Left.Level1_Optional_Id into groupJoinLeft
            from l2Left in groupJoinLeft.DefaultIfEmpty()
            join l2Right in
                from l1p in l1s where l1p.OneToOne_Required_PK != null select l1p.OneToOne_Required_PK
                on l1.Id equals l2Right.Level1_Optional_Id into groupJoinRight
            from l2Right in groupJoinRight.DefaultIfEmpty()
            where (l2Left == null ? null : l2Left.Name) == "L2 05" || (l2Right == null ? null : l2Right.Name) != "L2 42"
            select l1.Id);
}
```

Generated SQL:
```SQL
SELECT [l1].[Id]
FROM [Level1] AS [l1]
LEFT JOIN (
  SELECT [l1p.OneToOne_Required_PK].*
  FROM [Level1] AS [l1p]
  LEFT JOIN [Level2] AS [l1p.OneToOne_Required_PK] ON [l1p].[Id] = [l1p.OneToOne_Required_PK].[Id]
  WHERE [l1p.OneToOne_Required_PK].[Id] IS NOT NULL
) AS [t] ON [l1].[Id] = [t].[Level1_Optional_Id]
LEFT JOIN (
  SELECT [l1p.OneToOne_Required_PK0].*
  FROM [Level1] AS [l1p0]
  LEFT JOIN [Level2] AS [l1p.OneToOne_Required_PK0] ON [l1p0].[Id] = [l1p.OneToOne_Required_PK0].[Id]
  WHERE [l1p.OneToOne_Required_PK0].[Id] IS NOT NULL
) AS [t0] ON [l1].[Id] = [t0].[Level1_Optional_Id]
WHERE ([t].[Name] = N'L2 05') OR ([t0].[Name] <> N'L2 42')
```
Expected SQL:
```SQL
SELECT [l1].[Id]
FROM [Level1] AS [l1]
LEFT JOIN (
[...]

PR #8373, opened 2017-05-03
Fix to #8364 - Query: comparing collection navigations to null or to each other doesn't work

maumar:
Also fixes (smaller in scope) #5825 - Query: Comparing collection navigation to NULL throws Null Exception

Problem was that we were not taking into account queries where collection navigation was being compared to null and/or each other. This would cause various exceptions during query compilation.

Some of the common cases can be simplified, making them much easier for the query compiler (i.e. navigation rewrite) to process.

When it comes to null comparison, we assume that collection navigation is only null if its parent entity is null, so we can actually just compare the parent entity to null instead
e.g.: customer.Orders == null ---> customer == null ---> customer.Id == null

However, it's possible that user wanted to check whether collection has any elements - we will issue a warning that the comparison to null could have unintended consequences and suggest to use .Any() instead.

When collection navigations are compared to each other we can also approximate by comparing the parent entities instead (if the collections are the same INavigation). If navigations are different, we assume they are never equal.
e.g.:
csutomer1.Orders == customer2.Orders ---> customer1.Id == customer2.Id
customer.Orders == company.ShippedOrders ---> false

This is not a perfect translation, since there can be two collections with exactly the same elements coming from different navigations (e.g. two empty collections: c.CompletedOrders, c.PendingOrders)
Therefore we also issue the warning in this case, saying that the result may not be what user expected.

### DIFF
--- a/src/EFCore.Relational.Specification.Tests/WarningsTestBase.cs
+++ b/src/EFCore.Relational.Specification.Tests/WarningsTestBase.cs
@@ -143,6 +143,26 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
             }
         }
 
+        [Fact]
+        public virtual void Comparing_collection_navigation_to_null_issues_possible_unintended_consequences_warning()
+        {
+            using (var context = CreateContext())
+            {
+                var query = context.Customers.Where(c => c.Orders != null).ToList();
+                Assert.Equal(91, query.Count);
+            }
+        }
+
+        [Fact]
+        public virtual void Comparing_two_collections_together_issues_possible_unintended_reference_comparison_warning()
+        {
+            using (var context = CreateContext())
+            {
+                var query = context.Customers.Where(c => c.Orders == c.Orders).ToList();
+                Assert.Equal(91, query.Count);
+            }
+        }
+
         protected NorthwindContext CreateContext() => Fixture.CreateContext();
 
         protected WarningsTestBase(TFixture fixture)
--- a/src/EFCore.Relational/Query/ExpressionTranslators/Internal/EqualsTranslator.cs
+++ b/src/EFCore.Relational/Query/ExpressionTranslators/Internal/EqualsTranslator.cs
@@ -37,14 +37,14 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
         {
             Check.NotNull(methodCallExpression, nameof(methodCallExpression));
 
-            if ((methodCallExpression.Method.Name == nameof(object.Equals))
-                && (methodCallExpression.Arguments.Count == 1)
-                && (methodCallExpression.Object != null))
+            if (methodCallExpression.Method.Name == nameof(object.Equals)
+                && methodCallExpression.Arguments.Count == 1
+                && methodCallExpression.Object != null)
             {
                 var argument = methodCallExpression.Arguments[0];
 
-                if ((methodCallExpression.Method.GetParameters()[0].ParameterType == typeof(object))
-                    && (methodCallExpression.Object.Type != argument.Type))
+                if (methodCallExpression.Method.GetParameters()[0].ParameterType == typeof(object)
+                    && methodCallExpression.Object.Type != argument.Type)
                 {
                     argument = argument.RemoveConvert();
 
--- a/src/EFCore.Specification.Tests/ComplexNavigationsOwnedQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/ComplexNavigationsOwnedQueryTestBase.cs
@@ -281,6 +281,11 @@ namespace Microsoft.EntityFrameworkCore.Specification.Tests
         {
         }
 
+        // #8172 - One-to-many not supported yet
+        public override void Comparing_collection_navigation_on_optional_reference_to_null()
+        {
+        }
+
         protected override IQueryable<Level2> GetExpectedLevelTwo()
             => GetExpectedLevelOne().Select(t => t.OneToOne_Required_PK).Where(t => t != null);
 
--- a/src/EFCore/Infrastructure/CoreEventId.cs
+++ b/src/EFCore/Infrastructure/CoreEventId.cs
@@ -65,6 +65,8 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
             NavigationIncluded,
             IncludeIgnoredWarning,
             QueryExecutionPlanned,
+            PossibleUnintendedCollectionNavigationNullComparisonWarning,
+            PossibleUnintendedReferenceComparisonWarning,
 
             // Model validation events
             ModelValidationShadowKeyWarning = CoreBaseId + 300,
@@ -130,10 +132,24 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
         public static readonly EventId IncludeIgnoredWarning = MakeQueryId(Id.IncludeIgnoredWarning);
 
         /// <summary>
-        ///     A navigation was ignored while compiling a query.
+        ///     A query is planned for execution.
         ///     This event is in the <see cref="LoggerCategory.Query" /> category.
         /// </summary>
         public static readonly EventId QueryExecutionPlanned = MakeQueryId(Id.QueryExecutionPlanned);
+
+        /// <summary>
+        ///     Possible uninteded comparison of collection navigation to null.
+        ///     This event is in the <see cref="LoggerCategory.Query" /> category.
+        /// </summary>
+        public static readonly EventId PossibleUnintendedCollectionNavigationNullComparisonWarning
+            = MakeQueryId(Id.PossibleUnintendedCollectionNavigationNullComparisonWarning);
+
+        /// <summary>
+        ///     Possible uninteded reference comparison.
+        ///     This event is in the <see cref="LoggerCategory.Query" /> category.
+        /// </summary>
+        public static readonly EventId PossibleUnintendedReferenceComparisonWarning
+            = MakeQueryId(Id.PossibleUnintendedReferenceComparisonWarning);
 
         private static readonly string _validationPrefix = LoggerCategory.Model.Validation.Name + ".";
         private static EventId MakeValidationId(Id id) => new EventId((int)id, _validationPrefix + id);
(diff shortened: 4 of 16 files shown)


maumar:
cc: @ajcvickers for logging infra

anpete:
@maumar Rules sound good. We should also consider an Equals -> SequenceEqual rewrite for enumerables - We have talked about this before WRT to byte[] comparisons on in-memory for example.

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationEqualityRewritingExpressionVisitor.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitBinary(BinaryExpression binaryExpression)
+        {
+            Check.NotNull(binaryExpression, nameof(binaryExpression));
+
+            var newBinaryExpression = (BinaryExpression)base.VisitBinary(binaryExpression);
```
maumar: there is some overlap between this and EntityEqualityRewritingExpressionVisitor (around comparing to null mostly). Those two could be combined to DRY it bit more, however I found it bit awkward - EEREV explicitly doesn't rewrite navigations (those are suppose to be handled in nav rewrite). Thoughts?

src/EFCore.Relational.Specification.Tests/WarningsTestBase.cs
```diff
@@ -143,6 +143,26 @@ public virtual void LastOrDefault_without_order_by_issues_client_eval_warning()
             }
         }
 
+        [Fact]
+        public virtual void Comparing_collection_navigation_to_null_issues_possible_unintended_consequences_warning()
+        {
+            using (var context = CreateContext())
+            {
+                var query = context.Customers.Where(c => c.Orders != null).ToList();
+                Assert.NotNull(query);
```
anpete: Assert result count

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationEqualityRewritingExpressionVisitor.cs
```diff
@@ -0,0 +1,160 @@
...
+             Expression rightExpressionn,
+             ExpressionType expressionType,
+             QuerySourceReferenceExpression leftNavigationQsre,
+             QuerySourceReferenceExpression rightNavigationQsre,
+             IList<IPropertyBase> leftProperties,
+             IList<IPropertyBase> rightProperties)
+        {
+            // if both collections are the same navigations, compare their parent entities (by key)
+            // otherwise we assume they are different references and return false
+            if (IsCollectionNavigation(leftNavigationQsre, leftProperties)
```
maumar: or should we just not bother with collection comparison, since it cant be perfect and force client eval (at least when we fix MaterializeCollectionNavigation)

src/EFCore.Specification.Tests/ComplexNavigationsOwnedQueryTestBase.cs
```diff
@@ -281,6 +281,10 @@ public override void Navigation_filter_navigation_grouping_ordering_by_group_key
         {
         }
 
+        public override void Comparing_collection_navigation_on_optional_reference_to_null()
```
anpete: ?

src/EFCore.Relational/Query/ExpressionTranslators/Internal/EqualsTranslator.cs
```diff
@@ -37,14 +37,14 @@ public virtual Expression Translate(MethodCallExpression methodCallExpression)
         {
             Check.NotNull(methodCallExpression, nameof(methodCallExpression));
 
-            if ((methodCallExpression.Method.Name == nameof(object.Equals))
-                && (methodCallExpression.Arguments.Count == 1)
-                && (methodCallExpression.Object != null))
+            if (methodCallExpression.Method.Name == nameof(object.Equals)
```
anpete: Can we do reflection based comparison here?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationEqualityRewritingExpressionVisitor.cs
anpete: cc @smitpatel We should change EEREV to handle all of it.

src/EFCore.Specification.Tests/ComplexNavigationsOwnedQueryTestBase.cs
maumar: collections are not supported for DIETs

anpete: Add a comment

src/EFCore/Infrastructure/CoreEventId.cs
```diff
@@ -130,11 +132,25 @@ private enum Id
...
+        /// <summary>
+        ///     Possible uninteded comparison of collection navigation to null.
+        ///     This event is in the <see cref="LoggerCategory.Query" /> category.
+        /// </summary>
+        public static readonly EventId PossibleUnintendedCollectionNavigationNullComparisonWarning
+            = MakeQueryId(Id.PossibleUnintendedCollectionNavigationNullComparisonWarning);
+
+        /// <summary>
+        ///     Possible uninteded comparison of collection navigation to null.
+        ///     This event is in the <see cref="LoggerCategory.Query" /> category.
```
ajcvickers: Thanks for fixing one copy/paste error in the comment. But then you introduced this one. ;-)

ajcvickers:
Logging stuff looks good.

src/EFCore/Infrastructure/CoreEventId.cs
smitpatel: :trollface:

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationEqualityRewritingExpressionVisitor.cs
smitpatel: EEREV does not rewrite navigation == null because nav rewrite wanted to do it 😛

smitpatel: Merge this code in EEREV, there is nothing specific here, it is just more branching.
In current dev, EntityEquality does not rewrite, navigation == null (explicitly) & it does not know how to process collection navigation comparison so it does nothing there.
This code want to relax that to bit more like following
- Do not rewrite reference navigation == null
- Rewrite collection navigation == null or collection navigation comparison.

It would be easy to add this processing in EEREV.

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationEqualityRewritingExpressionVisitor.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+
+            return null;
+        }
+
+        private Expression CreateCollectionCallerExpression(
+            QuerySourceReferenceExpression qsre,
+            IList<IPropertyBase> properties)
+        {
+            var result = (Expression)qsre;
```
smitpatel: nit: Use `Expression result = qsre.

src/EFCore/Query/ExpressionVisitors/Internal/EqualityRewritingExpressionVisitorBase.cs
```diff
@@ -0,0 +1,87 @@
...
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Query.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public abstract class EqualityRewritingExpressionVisitorBase : ExpressionVisitorBase
```
smitpatel: Merge into one and remove this class

test/EFCore.SqlServer.FunctionalTests/ComplexNavigationsQuerySqlServerTest.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Linq;
 using Microsoft.EntityFrameworkCore.Specification.Tests;
+using Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.ComplexNavigationsModel;
```
smitpatel: unused using

test/EFCore.Sqlite.FunctionalTests/ComplexNavigationsOwnedQuerySqliteTest.cs
```diff
@@ -19,35 +19,5 @@ public override void Nested_group_join_with_take()
         {
             base.Nested_group_join_with_take();
         }
-
-        [ConditionalFact(Skip = "issue #8254")]
```
smitpatel: #8254 fixed here?

Thanks for removing skip on #8255

maumar: mistake

maumar:

src/EFCore/Query/ExpressionVisitors/Internal/EntityEqualityRewritingExpressionVisitor.cs
```diff
@@ -56,7 +58,49 @@ protected override Expression VisitBinary(BinaryExpression binaryExpression)
                 var nonNullExpression = isLeftNullConstant ? newBinaryExpression.Right : newBinaryExpression.Left;
 
                 var qsre = nonNullExpression as QuerySourceReferenceExpression;
-                // If a navigation being compared to null then don't rewrite
+
+                QuerySourceReferenceExpression leftNavigationQsre;
```
smitpatel: Can this go inline? `out var` is possible.

PR #8374, opened 2017-05-03
Add a convention that configures table splitting for owned types

AndriySvyryd:
The column names are prefixed by the navigation name

### DIFF
--- a/src/EFCore.Relational/Metadata/Conventions/Internal/SharedTableConvention.cs
+++ b/src/EFCore.Relational/Metadata/Conventions/Internal/SharedTableConvention.cs
@@ -1,0 +1,102 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class SharedTableConvention :
+        IEntityTypeConvention,
+        IEntityTypeAnnotationSetConvention,
+        IForeignKeyOwnershipConvention,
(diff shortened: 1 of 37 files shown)


src/EFCore.Relational/Metadata/Conventions/Internal/SharedTableConvention.cs
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
```
ajcvickers: Boilerplate "internal" doc comments?

src/EFCore.Relational/Metadata/Conventions/Internal/SharedTableConvention.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            var foreignKey = relationshipBuilder.Metadata;
+            if (foreignKey.IsOwnership
+                && foreignKey.IsUnique)
+            {
+                SetOwnedTable(foreignKey);
+            }
+            else
+            {
+                // TODO: Restore previous value
```
ajcvickers: Is there an issue filed for this?

test/EFCore.Sqlite.FunctionalTests/ComplexNavigationsOwnedQuerySqliteTest.cs
```diff
@@ -14,6 +14,11 @@ public ComplexNavigationsOwnedQuerySqliteTest(ComplexNavigationsOwnedQuerySqlite
         {
         }
 
+        public override void Level4_Include()
```
smitpatel: redundant

src/EFCore.Relational/Metadata/Conventions/Internal/SharedTableConvention.cs
AndriySvyryd: Part of https://github.com/aspnet/EntityFramework/issues/214:

> Whenever a configuration source of an aspect is changed because it is referenced from another element this reference should be stored in a collection, so if all references are removed the aspect can be removed as well or its configuration source downgraded to what it was previously.

smitpatel:
:shipit:

PR #8382, opened 2017-05-04
Fix test issue on SqlServer2016 with SqlServer Authentication

smitpatel:
`context.Database.EnsureDeleted()` fails in the test because we pass opened connection from SqlServerTestStore. & connection uses SQL Auth so when we try to connect to master for delete, it does not contain password and log in failure is caused.

### DIFF
--- a/test/EFCore.SqlServer.FunctionalTests/MemoryOptimizedTablesTest.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/MemoryOptimizedTablesTest.cs
@@ -16,7 +16,7 @@ namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests
         [ConditionalFact]
         public void Can_create_memoryOptimized_table()
         {
-            using (var testStore = SqlServerTestStore.Create("MemoryOptimizedTablesTest"))
+            using (var testStore = SqlServerTestStore.Create("MemoryOptimizedTablesTest", deleteDatabase: true))
             {
                 var options = new DbContextOptionsBuilder()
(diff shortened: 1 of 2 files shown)


test/EFCore.SqlServer.FunctionalTests/MemoryOptimizedTablesTest.cs
```diff
@@ -35,8 +35,6 @@ public void Can_create_memoryOptimized_table()
                 using (var context = new MemoryOptimizedContext(options))
                 {
                     Assert.Equal(fastUns.Select(f => f.Name), context.FastUns.OrderBy(f => f.Name).Select(f => f.Name).ToList());
-
-                    context.Database.EnsureDeleted();
```
divega: Probably a silly question: was this just not needed? Is it because it happens later for the whole class?

AndriySvyryd:
Make the test store delete it then. Add an optional parameter: `SqlServerTestStore.Create(string name, bool deleteDatabase = false)`

test/EFCore.SqlServer.FunctionalTests/MemoryOptimizedTablesTest.cs
AndriySvyryd: If it's not deleted the tests will pass but they might not be creating the database, therefore not exercising the code

smitpatel:
Added optional parameter. Manually verified that it deletes the database after test is finished.

smitpatel:
@AndriySvyryd - ping.

PR #8386, opened 2017-05-04
:fire: Remove TestFileLogger

### LINKED ISSUE #4552: Stop generating ~/.klog/data-test.log
Something in our tests is generating a file at `%USERPROFILE%\.klog\data-test.log`. We should remove that code.

smitpatel:
Resolves #4552 

There are still many test loggers here and there. N-variation issue :trollface:


test/EFCore.InMemory.FunctionalTests/NorthwindQueryInMemoryFixture.cs
```diff
@@ -38,7 +36,7 @@ public override DbContextOptions BuildOptions(IServiceCollection serviceCollecti
                     (serviceCollection ?? new ServiceCollection())
                     .AddEntityFrameworkInMemoryDatabase()
                     .AddSingleton(TestModelSource.GetFactory(OnModelCreating))
-                    .AddSingleton<ILoggerFactory>(_testLoggerFactory)
+                    .AddSingleton<ILoggerFactory>(TestLoggerFactory.Instance)
```
anpete: Be wary of concurrency issues here. E.g. From the code it looks like setting TestOutputHelper is not thread-safe (not sure if it matters) In general, try to avoid shared state in tests where possible.

smitpatel: How about each fixture initializing own instance of TestLoggerFactory? Would that make it thread safe (or fixture isolated)

anpete: Yep, that is better.

PR #8391, opened 2017-05-05
Fix race in test database creation.

anpete:
Fix race in test database creation.


src/EFCore.Specification.Tests/InheritanceFixtureBase.cs
```diff
@@ -17,17 +19,23 @@ public InheritanceContext CreateContext(bool enableFilters = false)
 
             if (!IsSeeded)
```
smitpatel: Checking for IsSeeded twice?

anpete: Yes, so we don't enter the lock every time.

Issue #7626, opened 2017-02-15
Add some Analyzers


bricelam:
At the very least, I would like to enable the ***useful*** StyleCop rules to catch invalid XML comments. (e.g. no `<returns>` on `void` methods)

This has been working well on the Microsoft.Data.Sqlite repo.

bricelam:
Closing. I investigated FxCop and additional StyleCop rules, but they appeared to provide very little value to our codebase.

PR #8392, opened 2017-05-05
Enable StyleCop on doc comments

bricelam:
The following rules are enabled.

- `<summary>`
  - required
  - not empty
  - no "or sets" when read-only
- `<parameter>`
  - not empty
  - has name
- `<typeparam>`
  - not empty
  - has name
- `<returns>`
  - not empty
  - not when void

Part of #7626


bricelam:
I debated whether or not to enforce `<parameter>` tags be ordered the same as the method. In the end, I decided it added very little value and started down the path to the dark side of StyleCop.

src/EFCore.ruleset
```diff
@@ -0,0 +1,158 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Microsoft.Data.Sqlite" ToolsVersion="15.0">
+  <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
+    <Rule Id="SA1000" Action="None" />
```
bricelam: Sorry this file is so ugly. It's the only way to author it to get what we want.

bricelam:
My philosophy for static analysis is that it should only be used to enforce correctness and completeness, never style.

bricelam:
I'll discuss in a design meeting whether we want to take it any further than this PR.

Issue #8385, opened 2017-05-04
Query: perform null protection pattern (a != null ? a.Name : null) to NullConditionalExpression (a?.Name) translation on QueryModel level


maumar:
Currently we only perform this translation when running SqlTranslatingExpressionVisitor. However, we could apply some of them on the query model level, which would result in simpler queries, especially around such patterns on entities with composite keys.

maumar:
This also applies to relational: Complex navigation test GroupJoin_on_complex_subquery_and_set_operation_on_grouping_but_nothing_from_grouping_is_projected":

```cs
            AssertQuery<Level1>(
                  l1s =>
                      l1s.GroupJoin(
                          l1s.Where(l1 => l1.Name != "L1 01").Select(l1 => l1.OneToOne_Required_FK),
                          l1 => l1.Id,
                          l2 => l2 != null ? l2.Level1_Optional_Id : null,
                          (l1, l2s) => new { l1, l2s })
                      .Where(r => r.l2s.Any())
                      .Select(r => r.l1),
                  e => e.Id,
                  (e, a) => Assert.Equal(e.Id, a.Id));
```

produces the following sql:

```sql
SELECT [l1].[Id], [l1].[Date], [l1].[Name], [l1].[OneToMany_Optional_Self_InverseId], [l1].[OneToMany_Required_Self_InverseId], [l1].[OneToOne_Optional_SelfId], [t].[Id], [t].[Date], [t].[Level1_Optional_Id], [t].[Level1_Required_Id], [t].[Name], [t].[OneToMany_Optional_InverseId], [t].[OneToMany_Optional_Self_InverseId], [t].[OneToMany_Required_InverseId], [t].[OneToMany_Required_Self_InverseId], [t].[OneToOne_Optional_PK_InverseId], [t].[OneToOne_Optional_SelfId]
FROM [Level1] AS [l1]
LEFT JOIN (
[...]

PR #8418, opened 2017-05-09
Fix to #8385 - Query: inefficient translation of null protection logic for InMemory and some cases of relational

maumar:
Moving some of the null conditional check optimization to QueryOptimizer.
QM expressions containing patterns like:
entity != null ? entity.Property : null

get converted to:
entity?.Property

Optimization on Query Model level is limited to member access, EF.Property and TypeAs
Optimization during SQL translation additionally handles all method calls and binary expressions (except equality)

This allows us to optimize more cases (e.g. composite keys) and overall simplifies query models.
Also improved null protection translation:
- we were incorrectly simplifying equality binary operations - those are not safe to translate because of null semantics

Also fixed small bug in binding where we would not protect from unexpectedly accessing null value from ValueBuffer in NullConditionals with Convert expressions.

### DIFF
--- a/src/EFCore/Extensions/Internal/ExpressionExtensions.cs
+++ b/src/EFCore/Extensions/Internal/ExpressionExtensions.cs
@@ -305,5 +305,56 @@ namespace Microsoft.EntityFrameworkCore.Internal
 
             return Expression.MakeMemberAccess(expression, member);
         }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static bool IsNullPropagationCandidate(
+            [NotNull] this ConditionalExpression conditionalExpression,
+            out Expression testExpression,
+            out Expression resultExpression)
+        {
+            Check.NotNull(conditionalExpression, nameof(conditionalExpression));
+
+            testExpression = null;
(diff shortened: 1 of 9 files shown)


src/EFCore/Extensions/Internal/ExpressionExtensions.cs
```diff
@@ -305,5 +305,60 @@ public static IQuerySource TryGetReferencedQuerySource([NotNull] this Expression
...
+
+            if (isLeftNullConstant == isRightNullConstant)
+            {
+                return false;
+            }
+
+            if (binaryTest.NodeType == ExpressionType.Equal)
+            {
+                var ifTrueConstant = conditionalExpression.IfTrue as ConstantExpression;
+                if (ifTrueConstant == null
```
smitpatel: `!IsNullConstantExpression`

maumar:
We could ditch the QM optimization and have it only in the SqlTranslatingVisitor. We would only lose the optimization for composite keys and have InMemory query plans bit complicated.

smitpatel:
We need QM optimization. Better QM is always good.

maumar:
Actually there could be more cases that are improved by QM translation - will take a look

maumar:
new version up, no longer using visitor for QM-level rewrite per @smitpatel suggestion, also implemented nested navigations support for QM rewrite

Issue #8315, opened 2017-04-27
Argument types don't match for queries projecting conditional expression with anonymous type result


sethen:
I am using a query in my repository which you can view at this gist: https://gist.github.com/sethen/c1f3cb54e50eb36b696a1f4d79d1670c#file-gistfile1-txt-L23 which is similar to the query I found in issue #5522 -- unfortunately I am still getting the same error.  Here is my stack trace:

```
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HL4DOE8S2AI2": An unhandled exception was thrown by the application.
System.ArgumentException: Argument types do not match
   at System.Linq.Expressions.Expression.Condition(Expression test, Expression ifTrue, Expression ifFalse, Type type)
   at System.Linq.Expressions.ConditionalExpression.Update(Expression test, Expression ifTrue, Expression ifFalse)
   at System.Linq.Expressions.ConditionalExpression.Accept(ExpressionVisitor visitor)
   at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
   at System.Linq.Enumerable.SelectIListIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source, Int32& length)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.SqlTranslatingExpressionVisitor.VisitNew(NewExpression expression)
   at System.Linq.Expressions.NewExpression.Accept(ExpressionVisitor visitor)
   at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.RelationalProjectionExpressionVisitor.Visit(Expression node)
   at Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.VisitSelectClause(SelectClause selectClause, QueryModel queryModel)
   at Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)
[...]

PR #8420, opened 2017-05-09
Fix to #8315 - Argument types don't match for queries projecting conditional expression with anonymous type result

maumar:
Initial problem was that when translating anonymous type we change the expression type from the anonymous type to Expression[].
In case of conditional expression, if the second result is null constant, it's type stays the same, and when we try to update ConditionalExpression, type mismatch is thrown.
However, even if the types are compensated for, we can't translate NewExpression to SQL, apart from it being used in comparison (e.g. in composite key join scenarios)

Fix is to recognize the pattern and force client eval.

Also fixes small bug around alias generation.

### DIFF
--- a/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
@@ -613,12 +613,13 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             }
 
             var currentAlias = GetColumnName(expression);
-            var uniqueAlias = newAlias ?? currentAlias ?? ColumnAliasPrefix;
+            var uniqueAliasBase = newAlias ?? currentAlias ?? ColumnAliasPrefix;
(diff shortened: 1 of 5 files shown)


src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -613,12 +613,13 @@ private Expression CreateUniqueProjection(Expression expression, string newAlias
             }
 
             var currentAlias = GetColumnName(expression);
-            var uniqueAlias = newAlias ?? currentAlias ?? ColumnAliasPrefix;
+            var uniqueAliasBase = newAlias ?? currentAlias ?? ColumnAliasPrefix;
```
smitpatel: inline current alias here.

maumar: currentAlias is being reused later

PR #8428, opened 2017-05-09
Suppress symbols nupkgs for metapackages

natemcmaster:
Suppress producing \*.symbols.nupkg files for packages that don't include build output.


natemcmaster:
Passes on public CI. I'm going to merge to unblock internal CI. This is holding up the gatekeeper build.

PR #8429, opened 2017-05-09
Query: Add parenthesis when expanding InExpression with null values

### LINKED ISSUE #8304: Incorrect SQL Generation: Missing parenthesis with Contains() and nullable collections
In a Where of the form `Where(e => SomeCheck && nullableList.Contains(e.SomeNullableProperty)` the generated SQL code is missing parenthesis therefore generating incorrect results.

Assume an entity like:
```c#
public sealed class Entity
{
    public int EntityId { get; set; }
    public DateTime Date { get; set; }
   public int? SomeValue { get; set; }
}
```
and the following LINQ:
```c#
var myIds = new int?[]
{
       null, 
       100
};
var result = await ctx.Entities.Where(e => e.Date >= new DateTime(2017, 1, 1) && 
                                               myIds.Contains(e.SomeValue)).ToListAsync();
```
This generates:
```sql
SELECT [e].[EntityId], [e].[Date], [e].[SomeValue]
FROM [Entities] AS [e]
WHERE ([e].[Date] >= '2017-01-01T00:00:00.000') AND [e].[SomeValue] IN (100) OR [e].[SomeValue] IS NULL
```
which is missing parenthesis around [e].[SomeValue] IN (100) OR [e].[SomeValue]. Due to this, every entry where SomeValue is null is returned, independent what the Date is.

### Further technical details
EF Core version: 1.1.1
Database Provider: Microsoft.EntityFrameworkCore.SqlServer 1.1.1
Operating system: Windows 10
IDE: Visual Studio 2017 Enterprise

### Steps to reproduce
Full console solution to reproduce the issue. The interesting part is at line 39 where we see that an entity is returned with DateTime 2016-01-01. The expected result would be an empty result.

```c#
using System;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.EntityFrameworkCore.Storage;

namespace DotNetCoreTest
{
    public static class DbContextExtensions
    {
        public static void LogToConsole(this DbContext context)
        {
            var serviceProvider = context.GetInfrastructure();
            var loggerFactory = (ILoggerFactory)serviceProvider.GetService(typeof(ILoggerFactory));
            loggerFactory.AddProvider(new EFLoggerProvider());
        }
    }

    internal static class Program
    {
        {
            MainInternal().GetAwaiter().GetResult();
        }

        private static async Task MainInternal()
        {
            await InitDb();
            {
                var myIds = new int?[]
                {
                    null, 
					100
                };
                var result = await ctx.Entities.Where(e => e.Date >= new DateTime(2017, 1, 1) && myIds.Contains(e.SomeValue)).ToListAsync();
                // This should be 0 but is 1. The result includes the entity with date 2016 despite the limiting where above
                Console.WriteLine(result.Count); 
                result.ForEach(Console.WriteLine); 
            }
        }

        private static async Task InitDb()
        {
            {
                ctx.LogToConsole();
                ctx.Database.EnsureCreated();
                if (await ctx.Entities.AnyAsync())
                {
                    return;
                }
                await ctx.Entities.AddRangeAsync(new[]
                {
                    new Entity()
                    {
                        Date = new DateTime(2017, 1, 1),
                        SomeValue = 0xdead
                    },
                    new Entity()
                    {
                        Date = new DateTime(2016, 1, 1),
                        SomeValue = null
                    }
                });
                await ctx.SaveChangesAsync();
            }
        }
    }

    public sealed class MyContext : DbContext
    {
        public DbSet<Entity> Entities { get; set; }

        {
            optionsBuilder.UseSqlServer(@"Server=(localdb)\mssqllocaldb;Database=TestDb;Trusted_Connection=True;MultipleActiveResultSets=true;");
        }
    }

    public sealed class Entity
    {
        public int EntityId { get; set; }

        public DateTime Date { get; set; }

        public int? SomeValue { get; set; }

        public override string ToString()
        {
            return $"{Date} - {SomeValue?.ToString() ?? "NULL"}";
        }
    }

    public class EFLoggerProvider : ILoggerProvider
    {
        public ILogger CreateLogger(string categoryName)
        {
            if (categoryName == typeof(IRelationalCommandBuilderFactory).FullName)
            {
                return new EFLogger();
            }
            return new NullLogger();
        }

        public void Dispose()
        {
            // N/A
        }

        private class NullLogger : ILogger
        {
            public bool IsEnabled(LogLevel logLevel)
            {
                return false;
            }

            public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
            { }

            public IDisposable BeginScope<TState>(TState state)
            {
                return null;
            }
        }


        private class EFLogger : ILogger
        {
            public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
            {
                Console.WriteLine(formatter(state, exception));
            }

            public bool IsEnabled(LogLevel logLevel)
            {
                return true;
            }

            public IDisposable BeginScope<TState>(TState state)
            {
                return null;
            }
        }
    }
}
```

smitpatel:
Resolves #8304 

The issue is marked with 1.1.3
Should I create new branch?

### DIFF
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -308,7 +308,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
             newExpression = _predicateReductionExpressionOptimizer.Visit(newExpression);
             newExpression = _predicateNegationExpressionOptimizer.Visit(newExpression);
             newExpression = _reducingExpressionVisitor.Visit(newExpression);
-            newExpression = _booleanExpressionTranslatingVisitor.Translate(newExpression, searchCondition: searchCondition);
+            newExpression = _booleanExpressionTranslatingVisitor.Translate(newExpression, searchCondition);
 
             return newExpression;
         }
@@ -342,12 +342,12 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
                 var leftBooleanConstant = GetBooleanConstantValue(binaryExpression.Left);
                 var rightBooleanConstant = GetBooleanConstantValue(binaryExpression.Right);
 
-                if ((binaryExpression.NodeType == ExpressionType.Equal
-                     && leftBooleanConstant == true
-                     && rightBooleanConstant == true)
-                    || (binaryExpression.NodeType == ExpressionType.NotEqual
-                        && leftBooleanConstant == false
-                        && rightBooleanConstant == false))
+                if (binaryExpression.NodeType == ExpressionType.Equal
+                    && leftBooleanConstant == true
+                    && rightBooleanConstant == true
+                    || binaryExpression.NodeType == ExpressionType.NotEqual
+                    && leftBooleanConstant == false
+                    && rightBooleanConstant == false)
                 {
                     return;
                 }
@@ -686,7 +686,13 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
                             new InExpression(inExpression.Operand, inValuesNotNull),
                             new IsNullExpression(inExpression.Operand));
 
-                    return Visit(relationalNullsInExpression);
+                    _relationalCommandBuilder.Append("(");
+
+                    Visit(relationalNullsInExpression);
+
+                    _relationalCommandBuilder.Append(")");
+
+                    return inExpression;
                 }
 
                 if (inValuesNotNull.Count > 0)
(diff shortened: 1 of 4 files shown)


src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -686,7 +686,13 @@ var relationalNullsInExpression
                             new InExpression(inExpression.Operand, inValuesNotNull),
                             new IsNullExpression(inExpression.Operand));
 
-                    return Visit(relationalNullsInExpression);
+                    _relationalCommandBuilder.Append("(");
```
maumar: need to handle NOT IN as well

smitpatel: `Not in` uses AND operator and does not suffer from short circuiting issue.

PR #8437, opened 2017-05-11
Diagnostics - Reworks timing infrastructure

anpete:
Start capturing event start time (DateTimeOffset.UtcNow) and switch duration to TimeSpan.

- Aligns with AppInsights APIs and is generally just easier to use and more useful.

### DIFF
--- a/src/EFCore.Relational/Storage/RelationalTransaction.cs
+++ b/src/EFCore.Relational/Storage/RelationalTransaction.cs
@@ -68,33 +68,30 @@ namespace Microsoft.EntityFrameworkCore.Storage
         /// </summary>
         public virtual void Commit()
         {
-            var startTimestamp = Stopwatch.GetTimestamp();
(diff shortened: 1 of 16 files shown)


src/EFCore.Relational/Storage/RelationalTransaction.cs
```diff
@@ -106,33 +103,30 @@ public virtual void Commit()
         /// </summary>
         public virtual void Rollback()
         {
-            var startTimestamp = Stopwatch.GetTimestamp();
+            var startTime = DateTimeOffset.UtcNow;
+            var stopwatch = Stopwatch.StartNew();
```
ajcvickers: Is it OK to not stop the stopwatch? Does it consume resources or anything?

anpete: It's fine, no resources (it just stores ticks internally).

PR #8438, opened 2017-05-11
Fix #7097 - Move AsyncEnumerable to root namespace.

### LINKED ISSUE #7097: Some wrinkles in async compiled queries
1) AsyncEnumerable return type should probably be in root namespace (currently. .Query)
2) No easy way to call a singleton async operator on an AsyncEnumerable.

anpete:
Doesn't address no. 2 on #7097 - "No easy way to call a singleton async operator on an AsyncEnumerable." I think we can address it if/when we get specific feedback.


smitpatel:
Breaking change? :trollface:

Issue #7787, opened 2017-03-04
Query: NullConditional is not added to some complex queries with optional navigation and collection with subquery


maumar:
query:

```
l1s => l1s.Where(l1 => l1.OneToOne_Optional_FK.OneToMany_Optional.Distinct().Select(l3 => l3.Id).Contains(1))
```

query model before nav-rewrite:
```
from Level1 l1 in DbSet<Level1>
where 
    (from Level3 l3 in 
        (from Level3 <generated>_2 in l1.OneToOne_Optional_FK.OneToMany_Optional
        select <generated>_2)
        .Distinct()
    select l3.Id)
    .Contains(1)
select l1
```

query model after nav-rewrite:
```
from Level1 l1 in DbSet<Level1>
join Level2 l1.OneToOne_Optional_FK in DbSet<Level2>
on (Nullable<int>) Property(l1, "Id") equals Property(l1.OneToOne_Optional_FK, "Level1_Optional_Id") into l1.OneToOne_Optional_FK_group
from Level2 l1.OneToOne_Optional_FK in 
    (from Level2 l1.OneToOne_Optional_FK_groupItem in l1.OneToOne_Optional_FK_group
    select l1.OneToOne_Optional_FK_groupItem)
    .DefaultIfEmpty()
where 
    (from Level3 l3 in 
        (from Level3 <generated>_2 in DbSet<Level3>
        where (Nullable<int>) Property(l1.OneToOne_Optional_FK, "Id") == Property(<generated>_2, "OneToMany_Optional_InverseId")
        select <generated>_2)
        .Distinct()
    select l3.Id)
    .Contains(1)
select l1
```

Problem here is that we don't introduce null check for left key access of 

```
where (Nullable<int>) Property(l1.OneToOne_Optional_FK, "Id") == Property(<generated>_2, "OneToMany_Optional_InverseId")
```

```l1.OneToOne_Optional_FK``` is optional navigation, so the key access could cause null ref.

This is only problem for InMemory - when we introduce the null check we need to make sure that this null check will be optimized for relational, so that we don't get worse queries.

**Comment by maumar** [comment]
Fixed in 0bc7be5aa529349a0f96279e65ca46350b340043

**Comment by smitpatel** [comment]
This translation does not seem to be correct.
From the description on pull request.
```
Problem was happening for InMemory queries with optional navigation and collection navigation, e.g.:

from e in entities
where e.Optional.Collection.Where(predicate)
select e

those would get rewritten into:

[...]

PR #8439, opened 2017-05-11
Fix to #7787 - Query: NullConditional is not added to some complex queries with optional navigation and collection with subquery

maumar:
Problem was happening for InMemory queries with optional navigation and collection navigation, e.g.:

from e in entities
where e.Optional.Collection.Where(predicate)
select e

those would get rewritten into:

from e in entities
join o in optionals on e.Id equals o.EntityId into grouping
from o in grouping.DefaultIfEmpty()
where o.Collection.Where(predicate)
select e

and then:

from e in entities
join o in optionals on e.Id equals o.EntityId into grouping
from o in grouping.DefaultIfEmpty()
join c in collections
where o.Id == c.OptionalId && predicate
select e

Problem was that o could be null (because of the optional navigation), which would result in NRE for InMemory.

We could add null compensation on the o.Id, but then null semantics would kick in for relational producing the following predicate:

WHERE o.Id = c.OptionalId OR o.Id IS NULL AND c.OptionalId IS NULL

and we would get bad results for some cases (when o is null and collection element FK is null)

Instead, fix is to introduce new reducible expression when creating predicate for the collection navigation rewrite, that for InMemory unwraps to:

o != null && o.Id == c.OptionalId

and for relational to simple comparison predicate: o.ID = c.OptionalId

Additionally, we can now remove workaround for this issue during RequiresMaterialization calculation, which results in more efficient queries on relational (we no longer need to unnecessarily materialize some entities)

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -971,38 +971,43 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         /// </returns>
         protected override Expression VisitExtension(Expression expression)
         {
-            if (expression is StringCompareExpression stringCompare)
+            switch (expression)
             {
-                var newLeft = Visit(stringCompare.Left);
-                var newRight = Visit(stringCompare.Right);
-
-                if (newLeft == null
-                    || newRight == null)
+                case StringCompareExpression stringCompare:
                 {
-                    return null;
+                    var newLeft = Visit(stringCompare.Left);
+                    var newRight = Visit(stringCompare.Right);
+                    if (newLeft == null
+                        || newRight == null)
+                    {
+                        return null;
+                    }
+
+                    return newLeft != stringCompare.Left
+                            || newRight != stringCompare.Right
+                        ? new StringCompareExpression(stringCompare.Operator, newLeft, newRight)
+                        : expression;
                 }
-
-                return newLeft != stringCompare.Left
-                       || newRight != stringCompare.Right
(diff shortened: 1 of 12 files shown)


src/EFCore/Query/Expressions/Internal/CollectionNavigationSubqueryKeyComparisonExpression.cs
```diff
@@ -0,0 +1,105 @@
...
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Query.Expressions.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class CollectionNavigationSubqueryKeyComparisonExpression : Expression
```
maumar: this thing *desperately* needs a better name

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -1012,6 +1012,15 @@ protected override Expression VisitExtension(Expression expression)
                 }
             }
 
+            if (expression is CollectionNavigationSubqueryKeyComparisonExpression collectionNavigationSubqueryKeyComparisonExpression)
```
anpete: Can you convert these conditionals to switch statement?

src/EFCore/Query/Expressions/Internal/CollectionNavigationSubqueryKeyComparisonExpression.cs
anpete: NullKeyEqualExpression?

PR #8442, opened 2017-05-11
Query: FromSql: Adds support for interpolated strings!

anpete:
Interpolated variables are parameterized and queries are cached in the usual way.

### DIFF
--- a/src/EFCore.Relational/RelationalQueryableExtensions.cs
+++ b/src/EFCore.Relational/RelationalQueryableExtensions.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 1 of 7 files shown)


anpete:
cc @divega

src/EFCore.Relational/RelationalQueryableExtensions.cs
```diff
@@ -66,5 +71,63 @@ public static class RelationalQueryableExtensions
...
+        /// </summary>
+        /// <typeparam name="TEntity"> The type of the elements of <paramref name="source" />. </typeparam>
+        /// <param name="source">
+        ///     An <see cref="IQueryable{T}" /> to use as the base of the interpolated string SQL query (typically a <see cref="DbSet{TEntity}" />).
+        /// </param>
+        /// <param name="sql"> The interpolated string SQL query. </param>
+        /// <returns> An <see cref="IQueryable{T}" /> representing the interpolated string SQL query. </returns>
+        public static IQueryable<TEntity> FromSql<TEntity>(
+            [NotNull] this IQueryable<TEntity> source,
+            [NotNull] [NotParameterized] FormattableString sql)
```
ajcvickers: Try to see if it can work with just this overload.

src/EFCore.Relational/RelationalQueryableExtensions.cs
```diff
@@ -66,5 +71,63 @@ public static class RelationalQueryableExtensions
...
+
+        /// <summary>
+        ///     A SQL format string. This type enables overload resolution between
+        ///     the regular and interpolated FromSql overloads.
+        /// </summary>
+        public struct SqlFormat
+        {
+            public static implicit operator SqlFormat(string s)
+            {
+                return new SqlFormat(s);
```
ajcvickers: Expression body

src/EFCore.Relational/RelationalQueryableExtensions.cs
anpete: Only works if passed strings are always interpolated. I.e. start with '$'

PR #8454, opened 2017-05-12
Diagnostics: Initial commit of new EF.AppInsights package.

anpete:
- Sets up the product/test projects and enables simple AI dependency forwarding of EF generated ADO events including EF correlation ids.

TBD:

- DbContext correlation? NB: ASP.NET AI integration already correlates everything at the Request level.
- Forward other events? NB. ASP.NET AI integration already forwards all logging events as AI Events and all errors as AI Errors, so it is unclear what we should do here.

### DIFF
--- a/src/EFCore.ApplicationInsights/DiagnosticEventForwarder.cs
+++ b/src/EFCore.ApplicationInsights/DiagnosticEventForwarder.cs
@@ -1,0 +1,277 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using JetBrains.Annotations;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.ApplicationInsights
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class DiagnosticEventForwarder : IObserver<DiagnosticListener>, IDisposable
+    {
+        private const string DependencyTypeName = "SQL";
+
+        private readonly object _sync = new object();
+
+        private readonly TelemetryClient _telemetryClient;
+
+        private IDisposable _efSubscription;
+        private IDisposable _listenerSubscription;
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public DiagnosticEventForwarder([NotNull] TelemetryClient telemetryClient)
+        {
+            Check.NotNull(telemetryClient, nameof(telemetryClient));
+
+            _telemetryClient = telemetryClient;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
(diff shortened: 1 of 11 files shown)


src/EFCore.ApplicationInsights/EFCore.ApplicationInsights.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\build\common.props" />
+  <PropertyGroup>
+    <Description>Entity Framework Core Application Insights integration.</Description>
+    <TargetFrameworks>net46;netstandard1.3</TargetFrameworks>
```
bricelam: Just `<TargetFramework>netstandard1.3</TargetFramework>`--no `#if` = no cross-target

src/EFCore.ApplicationInsights/EFCore.ApplicationInsights.csproj
```diff
@@ -0,0 +1,23 @@
...
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="..\Shared\*.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\EFCore.Relational\EFCore.Relational.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="System.Data.Common" Version="$(CoreFxVersion)" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.4.0-*" />
```
bricelam: Add an `AppInsightsVersion` property to `build/dependencies.props` and use it here.

src/EFCore.ApplicationInsights/EFCore.ApplicationInsights.csproj
bricelam: (note the missing *s*)

src/EFCore.ApplicationInsights/DiagnosticEventForwarder.cs
```diff
@@ -0,0 +1,264 @@
...
+                    return;
+                }
+
+                var eventName = value.Key;
+                var eventData = value.Value;
+
+                Debug.Assert(eventName.StartsWith(LoggerCategory.Root, StringComparison.Ordinal));
+                Debug.Assert(eventData != null);
+
+                switch (eventData)
```
ajcvickers: >switch (eventData) [](start = 16, length = 18)

It's interesting that we're just using the payload type here. Is it the idea that if a new event is added that uses the same payload type, then it should be automatically handled? Or would we have to filter by event type as well instead?

src/EFCore.Relational/Diagnostics/ConnectionData.cs
```diff
@@ -38,7 +38,7 @@ public class ConnectionData
         {
             Connection = connection;
             ConnectionId = connectionId;
-            Async = async;
+            IsAsync = async;
```
ajcvickers: >IsAsync [](start = 12, length = 7)

Been wondering why we treat bools as special and give the property names an "Is" prefix?

anpete: https://msdn.microsoft.com/en-us/library/ms229012.aspx

src/EFCore.ApplicationInsights/DiagnosticEventForwarder.cs
anpete: Mostly a result of using nominal types + c# pattern matching - We don't really need the event name (although we do end up sending it to AI)

src/EFCore.Relational/Diagnostics/ConnectionData.cs
ajcvickers: "Optionally, you can also prefix Boolean properties with "Is," "Can," or "Has," but only where it adds value."

Does it add value?

anpete: You could debate that in general, but I think it certainly does for us now because of consistency - I found > 200 occurrences of this pattern in our code base.

Issue #8456, opened 2017-05-13
Make query parameter format for logging consistent


ajcvickers:
In our tests, parameters were logged like this:
`[param: 17]`

In production logs, they were logged like this:
`[param='17']`

This is because we had different code generating the strings in each case. With the fix for #6946, there is only now one piece of code. It generates logs like we have been doing for our tests, because doing otherwise would result in massive baseline updates. However, if we decide to go with a different format, then we can do the work. 👷

PR #8457, opened 2017-05-13
Use template-based logging and cached delegates

ajcvickers:
Issue #6946

The T4 template for resource strings has been updated so that if the resource name starts with "Log", then we generate an EventDefinition using information from the comment section. This brings together the event ID, log-level, and a typed, cached delegate that can be called to log the event to an ILogger. It also contains a mechanism to generate the message for use when testing and when throwing warnings-as-errors. All the XxxLoggerExtensions class now use this mechanism.

As part of this, IInterceptingLogger was consolidated into IDiagnosticsLogger. There are no longer any Log methods here because logging is done through the definitions using the cached delegates.

A fallback mechanism--RawEventDefinition--can be used for messages with more than six parameters. There are only a handful of these, and they are for design-time scenarios and so don't need to be super fast.

Part of #218

Also, see #8456


smitpatel:
Breaking changes are not specified yet.

julielerman:
NomNomNom branch. funny guys!

Issue #7975, opened 2017-03-23
Query: we should propagate optional navigations thru nav rewrite, if the initial optional navigation was created using SelectMany-GroupJoin-DefaultIfEmpty


maumar:
Repro:

```cs
    class Program
    {
        static void Main(string[] args)
        {
            using (var ctx = new MyContext())
            {
                ctx.Database.EnsureDeleted();
                ctx.Database.EnsureCreated();

                var users3 = (from user in ctx.Users
                              join ou in ctx.OrganisationUsers on user.Id equals ou.UserId into link
                              from organisation in link.DefaultIfEmpty()
                              select new
                              {
                                  Id = user.Id,
                                  Organisation = organisation == null ? "N/A" : organisation.Organisation.Name
                              }).ToList();
            }
        }
    }

    public class MyContext : DbContext
    {
        public DbSet<User> Users { get; set; }
        public DbSet<Organisation> Organisations { get; set; }
        public DbSet<OrganisationUser> OrganisationUsers { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<OrganisationUser>().HasKey(ou => new { ou.OrganisationId, ou.UserId });
            modelBuilder.Entity<OrganisationUser>().HasOne(ou => ou.Organisation).WithMany(o => o.OrganisationUsers).HasForeignKey(ou => ou.OrganisationId);
            modelBuilder.Entity<OrganisationUser>().HasOne(ou => ou.User).WithMany(u => u.OrganisationUsers).HasForeignKey(ou => ou.UserId);
        }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
           optionsBuilder.UseSqlServer(@"Server=.;Database=Repro7975;Trusted_Connection=True;MultipleActiveResultSets=true");
        }
    }

    public class User
    {
        public int Id { get; set; }
        public List<OrganisationUser> OrganisationUsers { get; set; }
    }

    public class Organisation
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public List<OrganisationUser> OrganisationUsers { get; set; }
    }

    public class OrganisationUser
    {
        public int OrganisationId { get; set; }
        public Organisation Organisation { get; set; }

        public int UserId { get; set; }
        public User User { get; set; }
    }
```

This produces the following sql:

```sql
SELECT [user].[Id], CASE
    WHEN [ou].[OrganisationId] IS NULL
[...]

Issue #7761, opened 2017-03-02
Query : Missing null propagation logic for some complex queries with manually created GroupJoin-SelectMany-DefaultIfEmpty


maumar:
Query:

```
from l1_outer in ctx.LevelOne
join subquery in
	(
		from l2_inner in ctx.LevelTwo
		join l1_inner in ctx.LevelOne on l2_inner.Level1_Required_Id equals l1_inner.Id
		select l2_inner
	)
on l1_outer.Id equals subquery.Level1_Optional_Id into grouping
from subquery in grouping.DefaultIfEmpty()
select (int?)subquery.Id;
```

problem is that in final projection subquery can be null, however do don't add null protection logic to this particular call

maumar:
another case:

```cs
from l1 in ctx.LevelOne
from l2 in l1.OneToMany_Optional.Where(l => l.Id > 5).OrderByDescending(l => l.Name).DefaultIfEmpty()
select l1.Name + " " + l2.Name
```

maumar:
Fixed in bb7eebde422e6bf9601f4c97c72492dd52f0fe41

PR #8458, opened 2017-05-13
Fix to #7975 - Query: we should propagate optional navigations thru nav rewrite, if the initial optional navigation was created using SelectMany-GroupJoin-DefaultIfEmpty

maumar:
also fix to #7761 - Query : Missing null propagation logic for some complex queries with manually created GroupJoin-SelectMany-DefaultIfEmpty

Problem for queries with required navigations that were originating from qsre that was a GroupJoin-SelectMany-DefaultIfEmpty. In those cases, we should be producing LEFT JOINs for the navigations, even though they are required (because initial qsre is already "optional"). We would produce INNER JOINs instead, which would result in data corruption - returning fewer results than expected.

Fix is to use more sophisticated logic to determine whether the source qsre is "optional" and therefore we need to expand all its navigations to LOJs.

We also apply this logic to MemberAccess and EF.Property, effectively fully protecting against accidental NREs for InMemory and client eval scenarios (we were doing that before only for results of nav rewrite, now we also do it for manually created LOJs).

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -329,7 +329,10 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
             var testExpression = isLeftNullConstant ? binaryTest.Right : binaryTest.Left;
             var resultExpression = binaryTest.NodeType == ExpressionType.Equal ? node.IfFalse : node.IfTrue;
 
-            var nullCheckRemovalTestingVisitor = new NullCheckRemovalTestingVisitor(_queryModelVisitor.QueryCompilationContext);
+            var nullCheckRemovalTestingVisitor
+                = new NullCheckRemovalTestingVisitor(
+                    _queryModelVisitor.QueryCompilationContext, 
+                    _queryModelVisitor);
 
             return nullCheckRemovalTestingVisitor.CanRemoveNullCheck(testExpression, resultExpression)
                 ? resultExpression
@@ -339,13 +342,17 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         private class NullCheckRemovalTestingVisitor : ExpressionVisitorBase
         {
             private readonly RelationalQueryCompilationContext _queryCompilationContext;
+            private readonly RelationalQueryModelVisitor _queryModelVisitor;
             private IQuerySource _querySource;
             private string _propertyName;
             private bool? _canRemoveNullCheck;
 
-            public NullCheckRemovalTestingVisitor(RelationalQueryCompilationContext queryCompilationContext)
+            public NullCheckRemovalTestingVisitor(
+                RelationalQueryCompilationContext queryCompilationContext, 
+                RelationalQueryModelVisitor queryModelVisitor)
             {
                 _queryCompilationContext = queryCompilationContext;
+                _queryModelVisitor = queryModelVisitor;
             }
 
             public bool CanRemoveNullCheck(Expression testExpression, Expression resultExpression)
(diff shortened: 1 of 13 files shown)


src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -448,12 +446,72 @@ protected override Expression VisitMember(MemberExpression node)
...
+                    && lojSubquery.QueryModel.IsIdentityQuery()
+                    && lojSubquery.QueryModel.ResultOperators.Count == 1
+                    && lojSubquery.QueryModel.ResultOperators[0] is DefaultIfEmptyResultOperator)
+                {
+                    _nullCompensationNecessityMap[qsre] = true;
+
+                    return true;
+                }
+
+                var subQuery = (qsre.ReferencedQuerySource as FromClauseBase)?.FromExpression as SubQueryExpression
```
maumar: this might be an overkill, 90% of the scenarios are covered by the simple condition above, the pattern here is needed for queries that have optional navs in joins, such as:
```
from l3 in l3s
join l2_nav in l1s.Select(ll => ll.OneToOne_Optional_FK) on l3.Level2_Required_Id equals l2_nav.Id into grouping
from l2_nav in grouping.DefaultIfEmpty()
select l2_nav.Name
```
or in general, complex manually created joins - anytime we have a subquery that contains DIE we assume it needs null protection. Should we ditch the complex pattern, to make the code much simpler (but leave those scenarios failing)? @anpete @smitpatel

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -661,9 +661,15 @@ var updatedExpression
         }
 
         private static string GetColumnName(Expression expression)
-            => (expression as AliasExpression)?.Alias
-               ?? (expression as ColumnExpression)?.Name
-               ?? (expression as ColumnReferenceExpression)?.Name;
+        {
+            var processedExpression = expression.RemoveConvert();
```
smitpatel: We should not need remove convert here at all. It is `SelectExpression` the convert will never be printed out. If we need Convert around the projection, it should be converted to explicit cast otherwise trimmed out. Also this code is specifically used for projections only. NullableExpression are also not needed. A top level projection should never be nullable expression because null semantics are not applicable on it.

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -517,16 +517,9 @@ public virtual int AddToProjection([NotNull] Expression expression, bool resetPr
...
-            {
-                var unaryExpression = (UnaryExpression)expression;
-
-                if (unaryExpression.Type.UnwrapNullableType()
-                    == unaryExpression.Operand.Type)
-                {
-                    expression = unaryExpression.Operand;
-                }
-            }
+            expression = expression.RemoveConvert();
```
maumar: should be safe to remove all converts here, right? @smitpatel

smitpatel: It should be. Though thinking about whole thing, we may get issue if we are referencing projection outside and if it contains different type.

maumar:

smitpatel:
Side question: Is this rebased on latest dev?

maumar:
ping @smitpatel @anpete

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -385,22 +390,20 @@ private void AnalyzeTestExpression(Expression expression)
                     return;
                 }
 
-                if (expression is MethodCallExpression methodCallExpression
+                if (processedExpression is MethodCallExpression methodCallExpression
                     && methodCallExpression.Method.IsEFPropertyMethod())
```
anpete: Can we call BindMethodCall here instead?

Issue #8393, opened 2017-05-05
Query: Except/Intersect methods in Ix-Async does not work well with buffering


smitpatel:
Line in the source code:
`await Task.WhenAll((Task) moveNextTask, this.fillSetTask).ConfigureAwait(false);`

Disabled tests
`Except_nested`
`Intersect_nested`

Stack trace
```
Stacktraces: 
(new) Test collection for Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.AsyncQuerySqlServerTest (1): Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.AsyncQuerySqlServerTest.Intersect_nested 
System.InvalidOperationException : There is already an open DataReader associated with this Command which must be closed first.
   at System.Data.SqlClient.SqlCommand.<>c.<ExecuteDbDataReaderAsync>b__108_0(Task`1 result)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.<ExecuteAsync>d__26.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.EntityFrameworkCore.Query.Internal.AsyncQueryingEnumerable.AsyncEnumerator.<BufferlessMoveNext>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #8478, opened 2017-05-16
Query: Testing out a fix for #8393

anpete:
- Fixes a race in async ShapedQuery execution.
- Optimizes ShapedQuery by pushing down shaper to inner enumerable.
- Removes IValueBufferCursor.Current - no longer used.

### DIFF
--- a/src/EFCore.Relational/Query/QueryMethodProvider.cs
+++ b/src/EFCore.Relational/Query/QueryMethodProvider.cs
@@ -36,16 +36,10 @@ namespace Microsoft.EntityFrameworkCore.Query
             QueryContext queryContext,
             ShaperCommandContext shaperCommandContext,
             IShaper<T> shaper)
-        {
-            // ReSharper disable once LoopCanBeConvertedToQuery
-            foreach (var valueBuffer
-                in new QueryingEnumerable(
-                    (RelationalQueryContext)queryContext,
-                    shaperCommandContext))
-            {
-                yield return shaper.Shape(queryContext, valueBuffer);
-            }
-        }
+            => new QueryingEnumerable<T>(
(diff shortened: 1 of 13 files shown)


src/EFCore.Relational/Query/QueryMethodProvider.cs
```diff
@@ -37,14 +37,10 @@ public class QueryMethodProvider : IQueryMethodProvider
...
         {
-            // ReSharper disable once LoopCanBeConvertedToQuery
-            foreach (var valueBuffer
-                in new QueryingEnumerable(
-                    (RelationalQueryContext)queryContext,
-                    shaperCommandContext))
-            {
-                yield return shaper.Shape(queryContext, valueBuffer);
-            }
+            return new QueryingEnumerable<T>(
```
maumar: expression body

src/EFCore/Query/Internal/IParameterValues.cs
```diff
@@ -26,5 +29,11 @@ public interface IParameterValues
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         object RemoveParameter([NotNull] string name);
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        void SetParameter([NotNull] string name, [CanBeNull] object value);
```
maumar: it's weird that this interface is internal, but the class that implements it (QueryContext) is public

anpete: It is only consumed by internal components and so I think this is OK.

Issue #8026, opened 2017-03-29
Object reference not set to an instance of an object. at Microsoft.EntityFrameworkCore.Query.Internal.AsyncQueryingEnumerable.AsyncEnumerator.<BufferAllAsync>d__12.MoveNext()


spottedmahn:
I'm awaiting a handful of EF Core Queries and I'm seeing this exception

```
Exception message:Object reference not set to an instance of an object
Stack trace:
   at Microsoft.EntityFrameworkCore.Query.Internal.AsyncQueryingEnumerable.AsyncEnumerator.<BufferAllAsync>d__12.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.EntityFrameworkCore.Query.RelationalQueryContext.<RegisterValueBufferCursorAsync>d__14.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.EntityFrameworkCore.Query.Internal.AsyncQueryingEnumerable.AsyncEnumerator.<BufferlessMoveNext>d__9.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.EntityFrameworkCore.Storage.Internal.SqlServerExecutionStrategy.<ExecuteAsync>d__6`2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.EntityFrameworkCore.Query.Internal.AsyncQueryingEnumerable.AsyncEnumerator.<MoveNext>d__8.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

Issue #7160, opened 2016-11-30
Skipped tests report


smitpatel:
~AsyncQueryTestBase.Mixed_sync_async_query~
```
Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.AsyncQuerySqlServerTest.Mixed_sync_async_query [FAIL]
      System.InvalidOperationException : There is already an open DataReader associated with this Command which must be closed first.
      Stack Trace:
           at System.Data.SqlClient.SqlInternalConnectionTds.ValidateConnectionForExecute(SqlCommand command)
           at System.Data.SqlClient.SqlCommand.ValidateCommand(Boolean async, String method)
           at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean asyncWrite, String method)
           at System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior)
        D:\EntityFramework\src\Microsoft.EntityFrameworkCore.Relational\Storage\Internal\RelationalCommand.cs(279,0): at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.Execute(IRelationalConnection connection, String executeMethod, IReadOnlyDictionary`2 parameterValues, Boolean closeConnection)
        D:\EntityFramework\src\Microsoft.EntityFrameworkCore.Relational\Storage\Internal\RelationalCommand.cs(142,0): at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.ExecuteReader(IRelationalConnection connection, IReadOnlyDictionary`2 parameterValues)
        D:\EntityFramework\src\Microsoft.EntityFrameworkCore.Relational\Query\Internal\QueryingEnumerable.cs(129,0): at Microsoft.EntityFrameworkCore.Query.Internal.QueryingEnumerable.Enumerator.BufferlessMoveNext(Boolean buffer)
        D:\EntityFramework\src\Microsoft.EntityFrameworkCore.SqlServer\Storage\Internal\SqlServerExecutionStrategy.cs(35,0): at Microsoft.EntityFrameworkCore.Storage.Internal.SqlServerExecutionStrategy.Execute[TState,TResult](Func`2 operation, Func`2 verifySucceeded, TState state)
[...]

Issue #8305, opened 2017-04-27
Revisit concurrency detector


ajcvickers:
Need to make sure:
* If async paths are intended to be serialized, then make sure this works for all async paths--i.e. done at appropriate level
  * Consider overhead for doing this when there really isn't any async code in the async path--i.e. when Task.FromResult is used.
* New code paths that have been added after the concurrency detector was introduced are making the right calls into the detector.
  * How do we make sure this stays up-to-date going forward?

See also #8026

anpete:
@ajcvickers We currently do the throwing check in the following:

- ExecuteSqlCommand
- ExecuteSqlCommandAsync
- SaveChanges
- SaveChangesAsync
- MoveNext (sync query)

And we do the async waiting check (interleaving) only in:

- MoveNextAsync (async query)

Obviously there are many more top-level APIs that we could add the check too. Do we want to do that?

anpete:
ping @ajcvickers

ajcvickers:
Design meeting notes:
* We will not attempt to make the context completely thread-safe since there is too much complexity in the implementation and the extra overhead for the normal single-threaded use may be too much.
* @ajcvickers to propose other places to add the check. It doesn't need to be added everywhere, but we want to add in places where it is likely to get hit so that apps will more likely throw a good exception rather than just some random exception from multi-threaded access. (Note to self--context initialization?)
* @ajcvickers to do some tests to make sure common patterns will hit the check

AndriySvyryd:
Also the concurrency detector doesn't fire reliably on `.First()`, see the `Throws_on_concurrent_query_first()` test.

AndriySvyryd:
See also https://github.com/aspnet/EntityFramework/pull/8865

PR #8489, opened 2017-05-16
Fixes for #8026 and possibly flaky test...

anpete:
AsyncQueryTestBase.Mixed_sync_async_query (part of #7160)

- Moved query async Semaphore to RelationalConnection so multiple top-level queries can be serialized.
- Moved active query buffer management to RelationalConnection so buffering works across multiple top-level queries.
- Renamed IValueBufferCursor to IBufferable

### DIFF
--- a/test/EFCore.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
@@ -16,6 +16,7 @@ using Microsoft.EntityFrameworkCore.Specification.Tests;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Storage.Internal;
 using Microsoft.Extensions.DependencyInjection;
+// ReSharper disable SuggestBaseTypeForParameter
 
 namespace Microsoft.EntityFrameworkCore.SqlServer.FunctionalTests.Utilities
 {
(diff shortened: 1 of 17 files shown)


test/EFCore.SqlServer.FunctionalTests/Utilities/SqlServerTestStore.cs
```diff
@@ -464,7 +467,7 @@ private static string CreateConnectionString(string name, string fileName, bool
         {
             var builder = new SqlConnectionStringBuilder(TestEnvironment.DefaultConnection)
             {
-                MultipleActiveResultSets = multipleActiveResultSets,
+                MultipleActiveResultSets = false, //multipleActiveResultSets,
```
smitpatel: leftover

ajcvickers:
@anpete Was there a specific part of this that you wanted me to review?

anpete:
@ajcvickers Not really, just giving a heads up on the overall approach. Also would be great if we can decide whether we need to add more concurrency checks per my comment on #8305

ajcvickers:
Overall approach looks good to me. I like using the connection to carry this stuff. Let's discuss #8305 with @divega when you are back in the office. I have questions that I think will be more productive to ask in person. 😸

anpete:
Updated the title as this is mostly about #8026 and can go in independently of the discussion about #8305

anpete:
Ping

anpete:
Ping @smitpatel

Issue #8216, opened 2017-04-18
Query: navigation rewrite fails for queries with navigation inside a subquery inside join inner key selector


AndriySvyryd:
Paste into `ComplexNavigationsQueryTestBase`:
```C#
[ConditionalFact]
public virtual void Query_source_materialization_bug_8216()
{
    AssertQueryScalar<Level1, Level2, int>(
        (l1s, l2s) =>
            from e3 in from e2 in l2s where e2.OneToOne_Required_PK != null select e2.OneToOne_Required_PK
            join e1 in l1s
                on
                (int?)e3.Id
                equals
                (
                    from subQuery2 in l2s
                    join subQuery3 in from e2 in l2s where e2.OneToOne_Required_PK != null select e2.OneToOne_Required_PK
                        on
                        subQuery2 != null ? (int?)subQuery2.Id : null
                        equals
                        subQuery3.Level2_Optional_Id
                        into
                        grouping
                    from subQuery3 in grouping.DefaultIfEmpty()
                    select subQuery3 != null ? (int?)subQuery3.Id : null
                    ).FirstOrDefault()
            select e1.Id);
}
```

Query model:
```
from Level2 e2 in DbSet<Level2>
  where [e2].OneToOne_Required_PK != null
  join Level1 e1 in DbSet<Level1>
  on (Nullable<int>) [e2].OneToOne_Required_PK.Id equals 
      (from Level2 subQuery2 in DbSet<Level2>
      join Level3 subQuery3 in 
          from Level2 e2 in DbSet<Level2>
          where [e2].OneToOne_Required_PK != null
          select [e2].OneToOne_Required_PK
      on [subQuery2] != null ? (Nullable<int>) [subQuery2].Id : null equals [subQuery3].Level2_Optional_Id into grouping
      from Level3 subQuery3 in 
          (from Level3 <generated>_1 in [grouping]
          select [<generated>_1]).DefaultIfEmpty()
      select [subQuery3] != null ? (Nullable<int>) [subQuery3].Id : null).FirstOrDefault()
  select [e1].Id
```

Exception:
```
System.ArgumentException : Property 'Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.ComplexNavigationsModel.Level3 OneToOne_Required_PK' is not defined for type 'Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.ComplexNavigationsModel.Level3'
[...]

PR #8493, opened 2017-05-17
[TINY] Fix to #8216 - Query: navigation rewrite fails for queries with navigation inside a subquery inside join inner key selector

maumar:
Problem was that navigation inside inner key selector of a JoinClause was always being rewritten to subquery (needed for #3103). However, we should only be doing this for "naked" navs - if the nav itself is inside a subquery it can be safely rewritten into a join.

Fix is to "reset" the state indicating whether we are inside join inner key selector every time we visit SubQuery.


smitpatel:
There are some test failures in `Microsoft.EntityFrameworkCore.Sqlite.FunctionalTests.ComplexNavigationsOwnedQuerySqliteTest.Query_source_materialization_bug_4547`

maumar:
Test was non deterministic, added orderby to the inner key selector query, so that it returns same results on all providers

Issue #1414, opened 2015-01-15
Have a model building "mode" where everything must be explicitly specified


ajcvickers:
This item relates to #908 in which we decided to remove the BasicModelBuilder and instead use the convention ModelBuilder with no conventions for explicitly building the entire model. After further discussion whether this will work how people want depends on what are considered "conventions". Take the FK example. Removing conventions means that an FK property will not be discovered by convention, but we have no way of knowing what the FK will be until it is specified. If it isn't, then the shadow state FK will remain in the model--which could be seen as introducing a shadow state FK by convention.

If the goal is to ensure the entire model is explicitly specified then we could do this by having a mode in which model validation will check if anything in the model was not specified explicitly. If anything is found then model validation will fail. This has the additional advantage the conventions don't need to be removed to use this mode, although removing conventions would still be possible and could be useful for things like removing reachability.

breyed:
Comparing the suggestion in this issue with that item #7561, it's worth clarifying that there are at least three levels of explicitness:

1. **Implicit columns**: Assumed presence of columns in the database even without matching properties in the POCOs. POCO property relationships assumed by convention.
2. **Explicit columns**: Database columns only correspond to POCO properties. POCO property relationships assumed by convention.
3. **Explicit model**: Database columns only correspond to POCO properties. POCO property relationships must be specified explicitly (e.g. `CustomerId` is only the key for `Customer` if a configuration says so).

This item is advocating for option 3 (assuming I understand it correctly). Item #7561 advocates for option 2. All three options are useful. I think 3 is often overkill. I'm not sure how often people would want to do that much configuration, and if they did and forgot to configure a relationship, debugging the omission might still be difficult. I like option 2 because debugging columns that don't exist was so tricky. Option 1 is great for really lean POCOs, although I wonder if it is too automagical for a default.

ajcvickers:
Closing as we haven't seen significant demand for this feature. We could revisit this decision based on feedback and we would also consider a well-written community PR for it.

AndriySvyryd:
https://github.com/aspnet/EntityFrameworkCore/issues/214 will allow to remove all non-essential conventions

Issue #977, opened 2014-10-29
Make set discovery a convention in ModelBuilder


AndriySvyryd:
Avoid coupling ModelBuilder or Model with the context.

AndriySvyryd:
Alternatively  we could extract ModelSource.CreateModel to a different service

AndriySvyryd:
fixed in f4d74e6f2a863ceed42e1b7197d4486ae25ed02a

PR #8507, opened 2017-05-18
Take 2 - Add support for user defined scalar functions 

pmiddleton:
@anpete - Here is the new PR for adding UDF support.

This is a major refactor from the last PR.

- Core UDF code moved to EFCore.Relational from EFCore
- UDF metadata now stored in annotations on the model
- UDF translation now ties into the existing function translation mechanism
- Code first APIs now use the provider extension method pattern


anpete:
@bricelam Looks like there are some minor changes to Migrations/Design here

src/EFCore.Design/Migrations/Design/CSharpSnapshotGenerator.cs
```diff
@@ -607,6 +609,16 @@ protected virtual void GenerateForeignKeyAnnotations([NotNull] IForeignKey forei
             }
         }
 
+        protected virtual void IgnoreAnnotationTypes(
+            [NotNull] IList<IAnnotation> annotations, [NotNull] params string[] annotationPrefixes)
+        {
+            Check.NotNull(annotations, nameof(annotations));
+            Check.NotNull(annotationPrefixes, nameof(annotationPrefixes));
+
+            foreach(var ignoreAnnotation in annotations.Where(a => !annotationPrefixes.All(pre => a.Name.StartsWith(pre) == false)).ToList())
```
anpete: Add braces even when single line block

anpete:
@smitpatel Can you look at CF changes?

src/EFCore/Infrastructure/ModelSource.cs
```diff
@@ -79,6 +79,8 @@ public virtual IModel GetModel(DbContext context, IConventionSetBuilder conventi
 
             FindSets(modelBuilder, context);
 
+            CreateProviderModel(modelBuilder, context) ;
```
anpete: @ajcvickers Would a relational IModelCustomizer be a better option here?

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+using Microsoft.EntityFrameworkCore.Query.Expressions;
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    /// <summary>
+    ///     Represents a db function in an <see cref="IModel" />.
+    /// </summary>
+    public class DbFunction : IDbFunction, IMethodCallTranslator
+    {
+        private SortedDictionary<string, DbFunctionParameter> _parameters
```
anpete: readonly all fields where possible.

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+            = new SortedDictionary<string, DbFunctionParameter>(StringComparer.OrdinalIgnoreCase);
+
+        private Type _returnType;
+        private readonly string _annotationName;
+        private readonly IModel _model;
```
anpete: Unused field.

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            if (dbFunctionMethodInfo.IsGenericMethod == true)
+                throw new ArgumentException(CoreStrings.DbFunctionGenericMethodNotSupported(dbFunctionMethodInfo));
+
+            _model = model;
+            _annotationName = BuildAnnotationName(annotationPrefix, dbFunctionMethodInfo);
+
+            MethodInfo = dbFunctionMethodInfo;
+            ReturnType = MethodInfo.ReturnType;
+            Name = name ?? dbFunctionMethodInfo.Name;
```
anpete: name has NotNull attribute.

src/EFCore.Relational/Infrastructure/RelationalModelValidator.cs
```diff
@@ -60,6 +61,42 @@ public override void Validate(IModel model)
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected virtual void ValidateDbFunctions([NotNull] IModel model)
+        {
+            foreach (var dbFunction in RelationalExtensions.For(model).DbFunctions)
+            {
+                if (String.IsNullOrEmpty(dbFunction.Name))
```
anpete: Use built-in type names.

src/EFCore.Relational/Infrastructure/RelationalModelValidator.cs
```diff
@@ -60,6 +61,42 @@ public override void Validate(IModel model)
...
+                if (paramIndexes.Distinct().Count() != dbFunction.Parameters.Count())
+                    ShowError(CoreStrings.DbFunctionDuplicateIndex(dbFuncName));
+
+                if (Enumerable.Range(0, paramIndexes.Length).Except(paramIndexes).Any())
+                    ShowError(CoreStrings.DbFunctionNonContinuousIndex(dbFuncName));
+
+                if (dbFunction.MethodInfo.IsStatic == false && dbFunction.MethodInfo.DeclaringType.GetTypeInfo().IsSubclassOf(typeof(DbContext)) == false)
+                    ShowError(CoreStrings.DbFunctionDbContextMethodMustBeStatic(dbFuncName));
+
+                var missingType = dbFunction.Parameters.Where(p => p.ParameterType == null).FirstOrDefault();
```
anpete: Can move predicate to FirstOrDefault and remove Where.

src/EFCore.Relational/Infrastructure/RelationalModelValidator.cs
```diff
@@ -60,6 +61,42 @@ public override void Validate(IModel model)
...
+
+                if (missingType != null)
+                    ShowError(CoreStrings.DbFunctionParameterMissingType(dbFuncName, missingType.Name));
+
+                var arrayParam = dbFunction.Parameters.Where(p => p.ParameterType.GetTypeInfo().IsArray == true).FirstOrDefault();
```
anpete: Can move predicate to FirstOrDefault and remove Where.
redundant '== true'

src/EFCore.Relational/Infrastructure/RelationalModelValidator.cs
```diff
@@ -60,6 +61,42 @@ public override void Validate(IModel model)
...
+        {
+            foreach (var dbFunction in RelationalExtensions.For(model).DbFunctions)
+            {
+                if (String.IsNullOrEmpty(dbFunction.Name))
+                    ShowError(CoreStrings.DbFunctionNameEmpty());
+
+                var paramIndexes = dbFunction.Parameters.Select(fp => fp.ParameterIndex).ToArray();
+                var dbFuncName = $"{dbFunction.MethodInfo.DeclaringType.Name}.{dbFunction.MethodInfo.Name}";
+
+                if (paramIndexes.Distinct().Count() != dbFunction.Parameters.Count())
```
anpete: nit: Use Parameters.Count property

src/EFCore.Relational/Infrastructure/RelationalModelValidator.cs
```diff
@@ -60,6 +61,42 @@ public override void Validate(IModel model)
...
+                    ShowError(CoreStrings.DbFunctionDuplicateIndex(dbFuncName));
+
+                if (Enumerable.Range(0, paramIndexes.Length).Except(paramIndexes).Any())
+                    ShowError(CoreStrings.DbFunctionNonContinuousIndex(dbFuncName));
+
+                if (dbFunction.MethodInfo.IsStatic == false && dbFunction.MethodInfo.DeclaringType.GetTypeInfo().IsSubclassOf(typeof(DbContext)) == false)
```
anpete: break long lines. We break these before '&&'

src/EFCore.Relational/Infrastructure/Internal/RelationalModelSource.cs
```diff
@@ -23,6 +23,18 @@ public RelationalModelSource([NotNull] ModelSourceDependencies dependencies)
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        protected override void CreateProviderModel([NotNull] ModelBuilder modelBuilder, [NotNull] DbContext context)
```
anpete: Don't put NotNull annotations on overriding/implementing methods.

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
```diff
@@ -0,0 +1,53 @@
...
+using System;
+using System.Linq;
+using System.Reflection;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    class RelationalDbFunctionConvention : IModelAnnotationSetConvention
```
anpete: public

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        public Annotation Apply(InternalModelBuilder modelBuilder, string name, Annotation annotation, Annotation oldAnnotation)
+        {
+            if (name.StartsWith(RelationalFullAnnotationNames.Instance.DbFunctionPrefix))
```
anpete: Add StringComparison to StartsWith call

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
```diff
@@ -0,0 +1,53 @@
...
+            { 
+                Check.NotNull(modelBuilder, nameof(modelBuilder));
+
+                var dbFunctionBuilder = new RelationalDbFunctionBuilder(annotation.Value as DbFunction) ;
```
anpete: annotation has CanBeNull

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
```diff
@@ -0,0 +1,53 @@
...
+
+                var parameters = dbFuncMethodInfo.GetParameters()
+                                    .Where(p => p.ParameterType != typeof(DbFunctions))
+                                    .Select((pi, i) => new
+                                    {
+                                        ParameterIndex = i,
+                                        ParameterInfo = pi,
+                                        DbFuncParamAttr = pi.GetCustomAttributes<DbFunctionParameterAttribute>().SingleOrDefault(),
+                                        IsParams = pi.CustomAttributes.Any(ca => ca.AttributeType == typeof(ParamArrayAttribute)),
+                                        ParameterType = pi.ParameterType
```
anpete: Redundant name

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
```diff
@@ -0,0 +1,53 @@
...
+                foreach (var p in parameters)
+                {
+                    var paramBuilder = dbFunctionBuilder.HasParameter(p.ParameterInfo.Name);
+
+                    paramBuilder.HasParameterIndex(p.DbFuncParamAttr?.ParameterIndex ?? p.ParameterIndex);
+                    paramBuilder.HasType(p.ParameterType);
+                }
+            }
+
+            return annotation ;
```
anpete: nit: formatting

src/EFCore.Relational/Infrastructure/Internal/RelationalModelSource.cs
pmiddleton: Should Check.NotNull still be performed in overriding methods for NotNull parameters?

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
pmiddleton: What's the formatting nit here?

src/EFCore.Relational/Infrastructure/Internal/RelationalModelSource.cs
anpete: Yeah, but Checks are only required for public components (not in .Internal namespace).

pmiddleton:
@anpete   

I will get these all cleared up tonight and review the remaining files for similar patterns to get those cleaned as well.

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
anpete: space before ';'

anpete:
@pmiddleton Not sure if you have R#, but it will help with a lot of these - based on our EFCore.sln.DotSettings file.

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            return model.GetAnnotations()
+                .Where(a => a.Name.StartsWith(annotationPrefix, StringComparison.Ordinal))
+                .Select(a => a.Value)
+                .Cast<IDbFunction>();
+        }
+
+        private static string BuildAnnotationName(string annotationPrefix, MethodInfo methodInfo)
+        { 
+            StringBuilder sb = new StringBuilder($"{annotationPrefix}{methodInfo.Name}") ;
```
anpete: var

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+        /// <summary>
+        ///     If set this callback is used to translate the .Net method call to a Linq Expression.
+        /// </summary>
+        public virtual Func<IReadOnlyCollection<Expression>, IDbFunction, Expression> TranslateCallback { get; [param: CanBeNull] set; }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual bool HasTranslateCallback => TranslateCallback != null;
```
anpete: This needed?

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual DbFunctionParameter FindParameter([NotNull] string name)
+        {
+            Check.NotNull(name, nameof(name));
+
+            DbFunctionParameter item;
```
anpete: out var all the things.

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual void RemoveParameter([NotNull] string name, bool shiftParameters = false)
+        {
+            Check.NotNull(name, nameof(name));
+
+            if (shiftParameters == true)
```
anpete: Redundant '== true'

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+            {
+                var paramToRemove = FindParameter(name);
+
+                if (paramToRemove == null)
+                    return;
+
+                var parameters = Parameters;
+
+                for (int i = 0; i < parameters.Count; i++)
```
anpete: use var everywhere

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+                {
+                    if (parameters[i].ParameterIndex > paramToRemove.ParameterIndex)
+                        Parameters[i].SetParameterIndex(Parameters[i].ParameterIndex - 1, false);
+                }
+            }
+
+            _parameters.Remove(name);
+        }
+        
+        private bool IsValidReturnType(Type returnType)
```
anpete: Method can be static.

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,233 @@
...
+        Expression IMethodCallTranslator.Translate([NotNull] MethodCallExpression methodCallExpression)
+        {
+            var methodArgs = methodCallExpression.Method.GetParameters().Zip(methodCallExpression.Arguments, (p, a) => new { p.Name, Argument = a });
+
+            var arguments = new ReadOnlyCollection<Expression>(
+                                    (from p in Parameters
+                                    join a in methodArgs on p.Name equals a.Name into pas
+                                    from pa in pas.DefaultIfEmpty()
+                                    orderby p.ParameterIndex
+                                    select pa.Argument).ToList());
```
anpete: possible null ref? (pa)

src/EFCore.Relational/Metadata/DbFunctionAttribute.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    /// <summary>
+    ///     Defines a user defined database function
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
```
anpete: Redundant AllowMultiple

src/EFCore.Relational/Metadata/DbFunctionAttribute.cs
```diff
@@ -0,0 +1,58 @@
...
+            Check.NotEmpty(name, nameof(name));
+
+            Schema = schema;
+            Name = name;
+        }
+
+        /// <summary>
+        ///     The name of the function in the underlying datastore.
+        /// </summary>
+        public string Name { get; [param: NotNull] set; }
```
anpete: Check.NotNull on value

src/EFCore.Relational/Infrastructure/Internal/RelationalModelSource.cs
```diff
@@ -23,6 +23,18 @@ public RelationalModelSource([NotNull] ModelSourceDependencies dependencies)
...
+        {
+            foreach (var dbFunctionMethodInfo in DbFunctionFinder.FindFunctions(context))
```
ajcvickers: Shouldn't this be done in a relational convention?

src/EFCore.Relational/Metadata/DbFunction.cs
pmiddleton: Are you asking about just the HasTranslateCallback property or the entire TranslateCallback piece?

I added The TranslateCallback as a way for users to provider their own methodcall to SqlFunctionExpression translation in the cases where the fluent api isn't expressive enough.   It gives the end user the most flexibility to perform a translation, but will it be too confusing to the average user to leave in there?

The HasTranslateCallback bool property itself can go away - I can move the != null checks to where they are actually used.

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
pmiddleton: Yea my first TA in college had us do that to more easily see missing semis, back in the era before fancy ides and good compiler errors. 😄   Old habits die hard.  I'll find a remove the rest of those.

pmiddleton:
@anpete 

I was not using R#.  I installed it last night and went over all the files.  I think I have everything cleaned up now.

src/EFCore.Relational/Infrastructure/Internal/RelationalModelSource.cs
pmiddleton: I need access to the context in order to search for methods tagged with the dbfunctionattribute.

@anpete suggested using an IModelCustomizer.  I am looking into that approach now.

pmiddleton: I took a look and I could create a RelationalModelCustomizer, which finds the dbFunctions before calling OnModelCreating, and use that for IModelCustomizer DI for relation providers.  

The other options I see would be to add the context as a parameter to IModelConvention.Apply, or add a new convention type which accepts the context on its Apply method.

Thoughts anyone?

pmiddleton:
@anpete - excuse my github ignorance please, but I have a question.  If I comment on an outdated code comment will you get notified of that?

Thanks

src/EFCore.Relational/Metadata/DbFunctionAttribute.cs
```diff
@@ -0,0 +1,70 @@
...
+        public DbFunctionAttribute()
+        {
+        }
+
+        /// <summary>
+        /// Defines a user defined database function
+        /// </summary>
+        /// <param name="schema">The schema where the function lives in the underlying datastore.</param>
+        /// <param name="name">The name of the function in the underlying datastore.</param>
+        public DbFunctionAttribute([CanBeNull] string schema, [NotNull] string name)
```
smitpatel: Should this ordering be changed? with default for schema being null (or ctor with just name). Just like the way `TableAttribute` is constructed.
With above signature, user have to specify null schema or prolly use parameter names instead of positional.

src/EFCore.Relational/Metadata/DbFunctionParameter.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        private static bool IsValidParameterType(Type parameterType)
+        {
+            var typeToCheck = parameterType.IsArray ? parameterType.GetElementType() : parameterType;
+
+            if (typeToCheck.GetTypeInfo().IsGenericType && typeToCheck.GetTypeInfo().GetGenericTypeDefinition() == typeof(Nullable<>))
+                typeToCheck = Nullable.GetUnderlyingType(typeToCheck);
+
+            return typeToCheck.GetTypeInfo().IsPrimitive
```
smitpatel: Same as valid return type on DbFunction.

src/EFCore.Relational/Metadata/Internal/DbFunctionFinder.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static IReadOnlyList<MethodInfo> FindFunctions([NotNull] DbContext context)
+        {
+            return context.GetType().GetRuntimeMethods()
+                            .Where(m =>
+                                    m.GetCustomAttributes(typeof(DbFunctionAttribute), false).Any())
```
smitpatel: Probably inherit flag but use parameter name for `false` arg. Also any reason disabling inherited DbFunctions?

src/EFCore.Relational/Query/Expressions/SqlFunctionExpression.cs
```diff
@@ -34,19 +34,35 @@ public class SqlFunctionExpression : Expression
...
+        ///     Initializes a new instance of the Microsoft.EntityFrameworkCore.Query.Expressions.SqlFunctionExpression class.
+        /// </summary>
+        /// <param name="functionName"> Name of the function. </param>
+        /// <param name="returnType"> The return type. </param>
+        /// <param name="arguments"> The arguments. </param>
+        public SqlFunctionExpression(
+            [NotNull] string functionName,
+            [NotNull] Type returnType,
+            [NotNull] IEnumerable<Expression> arguments)
+            : this(functionName, returnType, null, arguments)
```
smitpatel: name the parameter in comment like `/*schema:*/ null`

src/EFCore.Specification.Tests/EFCore.Specification.Tests.csproj
```diff
@@ -14,6 +14,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\EFCore.Relational\EFCore.Relational.csproj" />
```
smitpatel: Core.Specification does not reference Relational.

smitpatel: If you are adding something in Core.Specs which needs relational component then right place is Relational.Specs

src/EFCore/Infrastructure/ModelSource.cs
```diff
@@ -79,6 +79,8 @@ public virtual IModel GetModel(DbContext context, IConventionSetBuilder conventi
 
             FindSets(modelBuilder, context);
 
+            CreateProviderModel(modelBuilder, context);
```
smitpatel: Is there any blockers in copying this method in derived type. There can be many points of extension in the sense that when someone wants to change the model builder. i.e. when it is build, or after running dbset discovery, after running fluent api code, after running customizer, after validating.

Perhaps best if all code here can be called in derived method then it would give much better control while overriding.

test/EFCore.Relational.Tests/Metadata/DbFunctionMetadataTests.cs
```diff
@@ -0,0 +1,433 @@
...
+        public void Add_method_generic_not_supported_throws()
+        {
+            var mb = GetModelBuilder();
+
+            var expectedMessage = CoreStrings.DbFunctionGenericMethodNotSupported(MethodHmi);
+
+            Assert.Equal(expectedMessage, Assert.Throws<ArgumentException>(() => mb.HasDbFunction(MethodHmi)).Message);
+        }
+   
+        private ModelBuilder GetModelBuilderNoConventions()
```
smitpatel: CreateModelBuilder

test/EFCore.Relational.Tests/Metadata/DbFunctionMetadataTests.cs
```diff
@@ -0,0 +1,433 @@
...
+        {
+            return new ModelBuilder(new ConventionSet());
+        }
+
+        private ModelBuilder GetModelBuilder()
```
smitpatel: CreateConventionalModelBuilder may be

smitpatel: Collapse to same method with parameter to add the convention.

test/EFCore.SqlServer.FunctionalTests/DbFunctionTests.cs
```diff
@@ -0,0 +1,699 @@
...
+{
+    public static class DateTimeExtensions
+    {
+        public static bool IsDate(this string date)
+        {
+            throw new Exception();
+        }
+    }
+
+    public class DbFunctionTests : IClassFixture<NorthwindQuerySqlServerFixture>
```
smitpatel: SqlServer should be in the name,

src/EFCore.Relational/Metadata/DbFunctionAttribute.cs
pmiddleton: Yes - this was overlooked during refactoring.

src/EFCore.Relational/Metadata/Internal/DbFunctionFinder.cs
pmiddleton: Disabling inherited funtions here is a bug.  

However GetRuntimeMethods does not return static methods on base classes.  Also GetMethods is not a part of .Net Standard 1.3, so I will need to roll some custom code here to find base class methods.

src/EFCore.Specification.Tests/EFCore.Specification.Tests.csproj
pmiddleton: I am using the Northwind context for some of my tests and I added some test methods to it which had DbFunctionAttributes on them.  

I have refactored this into Relational.Specs.  The last thing I need to change is TestModelSource.  I need to somehow refactor out the call to FindFunctions.  

Right now I don't see a clean way to do this without duplicating a bunch of code in TestModelSource and NorthwindQuerySqlServerFixture.  I'll take another look with fresh eyes tomorrow.

src/EFCore/Infrastructure/ModelSource.cs
pmiddleton: @anpete and @ajcvickers both have concerns about this line as well.

The issue is timing.  I need to find the dbFunctions after the model has been created and before model validation runs.  Overriding this method doesn't give enough granularity to get the FindFunctions call where it needs to be.   I also need access to the context.  Therefore using a ModelBuiltConventions won't work.

I'm open to ideas here.

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            _returnType = returnType;
+        }
+
+        private static bool IsValidReturnType(Type returnType)
+        {
+            if (returnType.GetTypeInfo().IsGenericType && returnType.GetTypeInfo().GetGenericTypeDefinition() == typeof(Nullable<>))
+                returnType = Nullable.GetUnderlyingType(returnType);
+
+            return returnType.GetTypeInfo().IsPrimitive
```
pmiddleton: @smitpatel - We just need to make sure that the parameter/return type can be mapped to a provider type - you can't use any random classes as a parameter type.

When I originally wrote this I just missed that there was an IRelationalTypeMapper that I could use to verify the parameter type instead of hardcoding the list.  

I will switch over to use  IRelationalTypeMapper.IsTypeMapped, the problem being I don't see how I can get an IRelationalTypeMapper into the DbFunction class.  The call chain doesn't lend itself to a place where I can DI one.  I'm looking for a similar example which I can copy, but I am not finding anything.  Have any ideas?

src/EFCore/Infrastructure/EntityFrameworkServicesBuilder.cs
```diff
@@ -115,7 +116,8 @@ public class EntityFrameworkServicesBuilder
                 { typeof(IQueryTrackingListener), new ServiceCharacteristics(ServiceLifetime.Scoped, multipleRegistrations: true) },
                 { typeof(IPropertyListener), new ServiceCharacteristics(ServiceLifetime.Scoped, multipleRegistrations: true) },
                 { typeof(IResettableService), new ServiceCharacteristics(ServiceLifetime.Scoped, multipleRegistrations: true) },
-                { typeof(ISingletonOptions), new ServiceCharacteristics(ServiceLifetime.Singleton, multipleRegistrations: true) }
+                { typeof(ISingletonOptions), new ServiceCharacteristics(ServiceLifetime.Singleton, multipleRegistrations: true) },
+                { typeof(IEvaluatableExpressionFilter), new ServiceCharacteristics(ServiceLifetime.Scoped) }
```
anpete: Can this be singleton?

src/EFCore.Relational/Metadata/DbFunction.cs
anpete: @pmiddleton +1 to moving the null checks.

src/EFCore/Infrastructure/ModelSource.cs
anpete: Can we re-purpose FindSets here? We could rename it to reflect that it is a general provider-level model post-processing hook. E.g. PostProcessModel

cc @ajcvickers

src/EFCore/Infrastructure/EntityFrameworkServicesBuilder.cs
pmiddleton: No.  

The IEvaluatableExpressionFilter needs access to the current model in order to know which methods are marked as translatable.  The ParameterExtractingExpressionVisitor will try to evaluate  methods at query compilation not marked as translatable which also have constant parameters.

We could look at injecting the model directly into the ParameterExtractingExpressionVisitor, but that will involve making it a scoped service and make the code there more confusing.

src/EFCore/Infrastructure/ModelSource.cs
smitpatel: Based on how `ModelCustomizer` is, we can probably use it. We run our OnModelCreating method through customizer. Also it has access to modelBuilder & DbContext both.

src/EFCore.Relational/Metadata/DbFunction.cs
smitpatel: Sorry, I could not check code directly. What actually sets the return type?
It is something which user can set directly or we do that while initializing functions?
If it is the latter case then we can probably move the check during some sort of validation. The validation can take advantage of TypeMapper.

src/EFCore/Infrastructure/ModelSource.cs
anpete: Yeah, but this code is conceptually the same as FindSets. I.e. Customize the model based on some aspects of the DbContext CLR type - We could implement FindSets in a customizer.

src/EFCore.Relational/Metadata/DbFunction.cs
pmiddleton: This was something that was user settable, but should no longer be.  

The parameter and return types are pulled via reflection.  The user defines a function and registers it with the fluentApi, or tags via the dbfunctionattribute and then it is processed at model creation time. 

You're right I think this should move to model validation.  I will look at verifying that tonight and then making the changes.

src/EFCore/Infrastructure/EntityFrameworkServicesBuilder.cs
pmiddleton: @anpete - I'm confused on your thumbs up.  Are you for leaving the code as is, or looking into if it can be moved into ParameterExtractingExpressionVisitor?

anpete: Leave as is

src/EFCore/Infrastructure/ModelSource.cs
smitpatel: > We could implement FindSets in a customizer...

I like this part. Why `FindSets` gets special status :trollface:

anpete: Probably because it was there first

ajcvickers: DbSet properties need to be initialized without building the model.

src/EFCore.Relational/Metadata/DbFunction.cs
pmiddleton: @smitpatel - This is now changed.  Return and Parameter types are checked with the relational type mapper during model validation.

src/EFCore.Relational/Metadata/RelationalDbFunctionParameterBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+            Check.NotNull(dbFunctionParameter, nameof(dbFunctionParameter));
+
+            _dbFunctionParameter = dbFunctionParameter;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual RelationalDbFunctionParameterBuilder HasType([NotNull]Type type)
```
pmiddleton: @smitpatel @anpete - Are there any plans to ever allow end users to remove/replace the default conventions used to build the model?

If not I want to mark this method as internal because it is only used by the RelationalDbFunctionConvention and end users shouldn't be calling it.  However if that convention can be removed then end users will need this method.

smitpatel: #214 & #1414
There may be but no mechanism at present.

src/EFCore/Infrastructure/ModelSource.cs
smitpatel: I found #977 today.

src/EFCore.Relational/Metadata/RelationalDbFunctionParameterBuilder.cs
pmiddleton: I'm inclined to hide it for now and expose it once the ability to replace convention sets is there.  

Along the same lines RelationalDbFunctionParameterBuilder.HasParameter should be internal with a new FindParameter being exposed.  There is no reason for the end user to add parameters as the convention creates them from the MethodInfo.

That way there are a few less ways to mess things up for the time being, but I'm fine going either way.  

What are your thoughts?

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,211 @@
...
+
+            if (dbFunctionMethodInfo.IsGenericMethod)
+                throw new ArgumentException(CoreStrings.DbFunctionGenericMethodNotSupported(dbFunctionMethodInfo));
+
+            MethodInfo = dbFunctionMethodInfo;
+            _name = name;
+            _schema = schema;
+            _returnType = dbFunctionMethodInfo.ReturnType;
+
+            model[BuildAnnotationName(annotationPrefix, dbFunctionMethodInfo)] = this;
```
AndriySvyryd: Currently the `ConfigurationSource` is not being stored or used. This means that if, for example, the schema is configured using the `DbFunctionAttribute` it will override the schema configured using `HasDbFunction`.
A simple way to fix this is to use the `ConfigurationSource` of the annotation containing the `DbFunction`, so if any aspect is configured using Code First (`HasDbFunction`) then the convention won't change anything.
For a more granular implementation the `ConfigurationSource` of each aspect needs to be stored in `DbFunction` (Name, Schema, Parameters, MethodInfo, ReturnType) so that the convention could fill in anything not configured through Code First,

src/EFCore.Relational/Metadata/RelationalDbFunctionBuilder.cs
```diff
@@ -0,0 +1,94 @@
...
+        private DbFunctionParameter GetOrCreateParameter([NotNull] string name)
+        {
+            return _dbFunction.FindParameter(name) ?? _dbFunction.AddParameter(name);
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual void RemoveParameter([NotNull] string name, bool shiftParameters = false)
```
AndriySvyryd: We don't provide `Remove*` methods in CF API, as they aren't declarative. Remove it or move it to an Internal class.

src/EFCore.Relational/Metadata/RelationalDbFunctionParameterBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Metadata
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class RelationalDbFunctionParameterBuilder
```
AndriySvyryd: This class should be documented too

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,211 @@
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        private DbFunction([NotNull] MethodInfo dbFunctionMethodInfo, 
+            [NotNull] IMutableModel model,
+            [NotNull] string annotationPrefix,
+            [NotNull] string name,
+            [CanBeNull] string schema = null)
```
AndriySvyryd: Since `schema` is optional it doesn't need to be in the constructor parameters

src/EFCore.Relational/Metadata/Internal/DbFunctionFinder.cs
```diff
@@ -0,0 +1,38 @@
...
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public static class DbFunctionFinder
+    {
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static IReadOnlyList<MethodInfo> FindFunctions([NotNull] DbContext context)
```
pmiddleton: There is a discussion about this going on in a comment on ModelSource.cs.

src/EFCore.Relational/Metadata/DbFunction.cs
pmiddleton: Is the rule that the Code First api takes priority over attribute based configuration if both are present?

src/EFCore/Infrastructure/ModelSource.cs
AndriySvyryd: It would be ok to find sets and functions in `ModelCustomizer`. Though it would be better to refactor the way it's called to make it more extensible similarly to `ConventionSet`/`CoreConventionSetBuilder`, but this can be done in a separate PR

src/EFCore.Relational/Metadata/DbFunction.cs
AndriySvyryd: Yes, CF uses the `Explicit` `ConfigurationSource`, attributes use `DataAnnotation` and other conventions use `Convention`. This is order is used to resolve conflicts.

pmiddleton: @AndriySvyryd - I have made some changes to DbFunction so the Name and Schema properties  use ConfigurationSource.  Those changes are in another branch in my repo.  

Can you take a look there and let me know if that approach is in line with what you are thinking.

Here is a link to DbFunction.cs - 
https://github.com/pmiddleton/EntityFramework/blob/udfTake2ConfigurationSource/src/EFCore.Relational/Metadata/DbFunction.cs

If that is going down the right path I will finish making the changes to DbFunctionParameter, fix/add unit tests, etc and get it merged into this PR.

Thanks

AndriySvyryd: Yes, you are on the right track. Though you don't need to store `_configurationSource` as it is already stored in the annotation itself. You can access it by using `RelationalAnnotationsBuilder`, see https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore.Relational/Metadata/Internal/RelationalEntityTypeBuilderAnnotations.cs

pmiddleton: Great - I will keep down this path and finish these changes up tonight / tomorrow.

src/EFCore/Metadata/Conventions/ConventionSet.cs
```diff
@@ -38,6 +38,12 @@ public class ConventionSet
             = new List<IEntityTypeAnnotationSetConvention>();
 
         /// <summary>
+        ///     Conventions to run when an annotation is set or removed on a model.
+        /// </summary>
+        public virtual IList<IModelAnnotationSetConvention> ModelAnnotationSetConventions { get; }
+            = new List<IModelAnnotationSetConvention>();
```
AndriySvyryd: We changed our naming convention for these. Could you rename the interface and property to `IModelAnnotationChangedConvention` and `ModelAnnotationChangedConventions`?

pmiddleton: Will do.

src/EFCore.Relational/Metadata/DbFunction.cs
pmiddleton: I've got most of the changes done, I just need to fix the unit tests.  I should be able to get everything in this weekend so you can review again on Monday.

src/EFCore/Infrastructure/ModelSource.cs
pmiddleton: So is the consensus to move FindSets into ModelCustomizer and then derive off a RelationalModelCustomizer to run FindFunctions?

anpete: Yes, we'd like to try that.

Thanks for all your effort on this!

pmiddleton: Glad to help.  Thank you all for helping me find the places I wasn't following conventions, or missing an existing service I could be using.  There is a lot of code to learn in this project.

src/EFCore.Relational/Metadata/DbFunction.cs
pmiddleton: @AndriySvyryd 

I pushed a fix for this, and your other issues.  Let me know what you think.

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    public class RelationalDbFunctionConvention : IModelAnnotationChangedConvention
+    {
+        public virtual Annotation Apply(InternalModelBuilder modelBuilder, string name, Annotation annotation, Annotation oldAnnotation)
+        {
+            Check.NotNull(modelBuilder, nameof(modelBuilder));
+            Check.NotNull(name, nameof(name));
+
+            if (name.StartsWith(RelationalAnnotationNames.DbFunction, StringComparison.OrdinalIgnoreCase)
+                && annotation != null)
```
AndriySvyryd: Add `&& oldAnnotation == null` condition. This code doesn't need to run more than once per function.

src/EFCore.Relational/Metadata/Internal/RelationalInternalDbFunctionBuilder.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.EntityFrameworkCore.Query.Expressions;
+using Microsoft.EntityFrameworkCore.Internal;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class RelationalInternalDbFunctionBuilder
```
AndriySvyryd: You can drop `Relational` from the name

src/EFCore.Relational/Metadata/Internal/RelationalInternalDbFunctionParameterBuilder.cs
```diff
@@ -0,0 +1,61 @@
...
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class RelationalInternalDbFunctionParameterBuilder
```
AndriySvyryd: You can drop `Relational` from the name

src/EFCore.Relational/Metadata/Internal/RelationalInternalDbFunctionBuilder.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        private readonly DbFunction _dbFunction;
+
+        public RelationalInternalDbFunctionBuilder([NotNull] DbFunction dbFunction)
+        {
+            _dbFunction = dbFunction;
+        }
+
+        public virtual DbFunction DbFunction => _dbFunction;
```
AndriySvyryd: Rename this to `Metadata` for consistency with other builders.

src/EFCore.Relational/Metadata/RelationalDbFunctionBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+    ///     <para>
+    ///         Provides a simple API for configuring a <see cref="DbFunction" />.
+    ///     </para>
+    ///     <para>
+    ///         Instances of this class are returned from methods when using the <see cref="ModelBuilder" /> API
+    ///         and it is not designed to be directly constructed in your application code.
+    ///     </para>
+    /// </summary>
+    /// 
+    public class RelationalDbFunctionBuilder
```
AndriySvyryd: You can drop `Relational` from the name

src/EFCore.Relational/Metadata/RelationalDbFunctionParameterBuilder.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <summary>
+    ///     <para>
+    ///         Provides a simple API for configuring an <see cref="DbFunctionParameter" />.
+    ///     </para>
+    ///     <para>
+    ///         Instances of this class are returned from methods when using the <see cref="ModelBuilder" /> API
+    ///         and it is not designed to be directly constructed in your application code.
+    ///     </para>
+    /// </summary>
+    public class RelationalDbFunctionParameterBuilder
```
AndriySvyryd: You can drop `Relational` from the name

src/EFCore.Relational/Metadata/RelationalDbFunctionParameterBuilder.cs
```diff
@@ -0,0 +1,52 @@
...
+            Check.NotNull(dbFunctionParameter, nameof(dbFunctionParameter));
+
+            _builder = new RelationalInternalDbFunctionParameterBuilder(dbFunctionParameter);
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual IDbFunctionParameter Parameter => _builder.Parameter;
```
AndriySvyryd: Rename this and the underlying property to `Metadata`

src/EFCore/Metadata/Conventions/Internal/ConventionDispatcher.cs
```diff
@@ -93,6 +93,21 @@ public virtual bool OnEntityTypeIgnored([NotNull] InternalModelBuilder modelBuil
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        public virtual Annotation OnModelAnnotationSet(
```
AndriySvyryd: `Set` -> `Changed`

src/EFCore/Infrastructure/ModelSource.cs
AndriySvyryd: @pmiddleton Don't forget to do the above

pmiddleton: I have pushed an update for this now.

src/EFCore.Relational/Infrastructure/EntityFrameworkRelationalServicesBuilder.cs
```diff
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using JetBrains.Annotations;
+using Remotion.Linq.Parsing.ExpressionVisitors.TreeEvaluation;
```
smitpatel: sort usings

src/EFCore.Relational/Infrastructure/RelationalModelCustomizerDependencies.cs
```diff
@@ -0,0 +1,47 @@
...
+            Check.NotNull(functionFinder, nameof(functionFinder));
+
+            FunctionFinder = functionFinder;
+        }
+
+        /// <summary>
+        ///     Gets the <see cref="IDbFunctionFinder" /> that will locate the UDF functions
+        ///     on the context.
+        /// </summary>
+        public virtual IDbFunctionFinder FunctionFinder { get; }
```
smitpatel: DbFunctionFinder name? Seems inconsistent with SetFinder that way but FunctionFinder seems too generic. cc:@AndriySvyryd

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            return model.GetAnnotations()
+                .Where(a => a.Name.StartsWith(annotationPrefix, StringComparison.Ordinal))
+                .Select(a => a.Value)
+                .Cast<IDbFunction>();
+        }
+
+        private static string BuildAnnotationName(string annotationPrefix, MethodInfo methodInfo)
+        { 
+            var sb = new StringBuilder($"{annotationPrefix}{methodInfo.Name}");
```
smitpatel: Minor thing. May be round brackets around the parameters & comma separated so annotation name become more readable. (though the only place it would be visible is Model.ToString.

src/EFCore.Relational/Metadata/DbFunction.cs
```diff
@@ -0,0 +1,242 @@
...
+            = new SortedDictionary<string, DbFunctionParameter>(StringComparer.OrdinalIgnoreCase);
+
+        private string _schema;
+        private string _name;
+
+        private ConfigurationSource? _schemaConfigurationSource;
+        private ConfigurationSource? _nameConfigurationSource;
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
```
smitpatel: Not in internal namespace.

src/EFCore.Relational/Infrastructure/RelationalModelCustomizerDependencies.cs
AndriySvyryd: We can avoid having to come up with a good name by inlining this into `FindDbFunctions`.
The reason `SetFinder` is a service is, as Arthur said, because it is also used to initialize the set properties.

smitpatel:
LGTM for query side.

src/EFCore.Relational/Infrastructure/RelationalModelCustomizerDependencies.cs
pmiddleton: FindDbFunctions was just a static method on a static class until I got the bright idea to turn it into a service at 12:30 last night.  IDbSetFinder was a service and FindDbFunctions was jealous :)

I can revert this back which will remove this class and the naming issue.

src/EFCore.Relational/Metadata/DbFunctionBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+            _builder.HasName(name, ConfigurationSource.Explicit);
+
+            return this;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual DbFunctionParameterBuilder GetParameter([NotNull] string name)
```
AndriySvyryd: To be able to configure the DbFunction with just this API, without annotations, this should be changed to `HasParameter` and also add a way to set the parameter type using `DbFunctionParameterBuilder`

src/EFCore.Relational/Metadata/DbFunctionAttribute.cs
```diff
@@ -0,0 +1,70 @@
...
+        ///     The schema where the function lives in the underlying datastore.
+        /// </summary>
+        public string Schema { get; [param: CanBeNull] set; }
+    }
+
+    /// <summary>
+    ///     Define a parameter for a user defined database function
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class DbFunctionParameterAttribute : Attribute
```
AndriySvyryd: Move this to a separate file

src/EFCore.Relational/Metadata/DbFunctionBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public DbFunctionBuilder([NotNull] DbFunction dbFunction)
+        {
+            Check.NotNull(dbFunction, nameof(dbFunction));
+
+            _builder = new InternalDbFunctionBuilder(dbFunction);
+        }
+
+        public virtual IDbFunction DbFunction => _builder.Metadata;
```
AndriySvyryd: Rename to `Metadata`

src/EFCore.Relational/Metadata/DbFunctionBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual DbFunctionBuilder HasSchema([NotNull]string schema)
```
AndriySvyryd: [CanBeNull]

src/EFCore.Relational/Metadata/Internal/InternalDbFunctionBuilder.cs
```diff
@@ -0,0 +1,97 @@
...
+            _dbFunction = dbFunction;
+        }
+
+        public virtual DbFunction Metadata => _dbFunction;
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual InternalDbFunctionBuilder HasSchema([NotNull] string schema, ConfigurationSource configurationSource)
```
AndriySvyryd: [CanBeNull]

src/EFCore.Relational/Metadata/DbFunctionBuilder.cs
pmiddleton: The function parameters are discovered when the RelationalDbFunctionConvention runs.  The parameters and their types are picked up from the signature of the .Net method.  This happens regardless of how they function was added (attribute or fluent api).  Setting the parameter type should be done by changing the signature of the .Net method.

Originally I had this feature, but I pulled it.  The only scenario where you would need to add a parameter is when you have a database function which requires a constant value (like DateAdd in sql server) which doesn't have a corresponding parameter on the .Net method call.  In that case we have the TranslateWith option on DbFunction which can be used to provider a custom mapping from IDbFunction -> SqlFunctionExpression.  It seemed redundant to have both options.

src/EFCore.Relational/Metadata/DbFunction.cs
pmiddleton: Is this considered internal?  It's not in an internal namespace or folder.  If so I will remove all of these comments in this file.

smitpatel: Those doc comments are there for public classes (which requires doc comments for build to pass) but in `.Internal` namespace. Classes in `.Internal` may break in minor release too and we believe most app developer won't need to use them. So providing a good documentation is not a priority hence this stub.

src/EFCore.Relational/Metadata/DbFunctionBuilder.cs
AndriySvyryd: I agree that this is low value in this case, but generally we allow the users to supply the full configuration without having to rely on the convention. This will become more useful when we add an easy way of removing conventions.

We already have a lot of redundancy in model building because we support different coding styles (Annotations, Fluent API, Nested Closures, etc.)

src/EFCore.Relational/Metadata/DbFunction.cs
AndriySvyryd: We need a public way of changing the low-level Metadata. To be consistent with other API (like Sequence) move this class to .Internal and add `IMutableDbFunction` interface

src/EFCore.Relational/Infrastructure/RelationalModelCustomizerDependencies.cs
pmiddleton: Any input on whether we want want to remove the IDbFunctionFinder Service here and go back to a static method for FindDbFunctions, or fix the naming issue?

AndriySvyryd: My preference would be to remove the service and move the code to `RelationalModelCustomizer.FindDbFunctions`

pmiddleton: This is cleared up now.

src/EFCore.Relational/Metadata/DbFunctionBuilder.cs
pmiddleton: Ok - I will take a look at this tomorrow.  Everything else has been cleaned up.

smitpatel:
@pmiddleton - Also if you are rebasing then rebase on rel/2.0.0-preview2 branch and change base in this PR. If its good and done in time, we may want to put this in preview2 release. It would be easier to rebase/merge into dev in either case.

src/EFCore.Relational/Metadata/DbFunctionBuilder.cs
pmiddleton: @AndriySvyryd - Regarding configurationSource for this.  If a parameter was added via the convention and the user then calls .HasProperty should we replace the entire convention parameter with the explicit api parameter?  I am thinking not, since then we would lose any properties on the parameter set by the convention which the fluent api then does not set.

I am looking for an example of this pattern and I'm looking at Property on InternalEntityTypeBuilder.  There if the property already exists it is not replaced.  I just want to make sure that is the correct pattern to be following.

AndriySvyryd: Yes, that's the right thing to do. Just update the configurationSource to Explicit, configuration stays the same.

pmiddleton: @AndriySvyryd - I have made the changes to the api.  Take a look and let me know if they meet what you are thinking in regards to expanding the api surface.  

I still need to add unit tests for the new apis.

pmiddleton:
Ok I need a little git help.  What command do I need to run to rebase onto origin/rel/2.0.0-preview2.

I pulled a tracked branch of rel/2.0.0-preview locally, but I'm not sure what to run to perform the rebase.

AndriySvyryd:
@pmiddleton 
```
git rebase -i origin/rel/2.0.0-preview2
```
Then delete all the commits that aren't yours from the rebase specification that pops up

pmiddleton:
Ok I have rebased onto rel/2.0.0-preview2

src/EFCore.Relational/Metadata/DbFunctionBuilder.cs
pmiddleton: New unit tests have been added for the new apis.

AndriySvyryd:
@pmiddleton There are still 4 commits that aren't yours. Run
```
git rebase -i origin/rel/2.0.0-preview2
```
again and make sure to remove them.

pmiddleton:
@AndriySvyryd - I rebased and removed the other 4 commits.

pmiddleton:
@AndriySvyryd - There are conflicts I need to resolve.  At this point in the process can I do a pull --rebase to resolve or should I be doing a rebase -i, or does it not matter now?

anpete:
Nice work @pmiddleton!

And thanks to @smitpate, @AndriySvyryd and @bricelam for the review.

ajcvickers:
Great to see this get in for preview2. Thanks to everybody!

pmiddleton:
Thanks to everyone for the help getting this in!

PR #8512, opened 2017-05-18
Remove overriding of relational annotations

ajcvickers:
Part of #7166

Since we now build one model per database provider there is no need to allow relational and provider-specific overrides to exist together in the same model. This means that any relational annotation will always be prefixed with "Relational:" and never with, for example, "SqlServer:". Provider-specific annotations still use the provider prefix.

This particularly advantageous for the consumer, since to read, for example, the table name, a consumer only ever needs to read `entityType.Relational().TableName;` instead of having to handle the case where it has been overridden for the current provider.

Note: there is other API (e.g. ForSqlServerTableName) than can also be removed; that will be part of another PR.

Breaking change baselines not updated yet.


bricelam:
😍 It's beautiful

PR #8519, opened 2017-05-19
Unblock MonsterFixupTest for SQLCE (which only supports dates after 1753

### LINKED ISSUE #8496: Recent changes to MonsterModel breaks SQL CE provider test
https://github.com/aspnet/EntityFramework/commit/e6f0e1921216e46de774d8b55785223617982d3d
introduced use of IAuditInfo and other new Interface that contain DateTime properties. Since no value is supplied, these values default to DateTime.MinValue, which breaks INSERTs with SQL CE, as it only supports SqlDateTime.MinValue - I have helped fix this earlier, but unsure of the best way to fix now, since IAuditInfo is an owned entity. @AndriySvyryd

ErikEJ:
fixes #8496


ajcvickers:
Thanks @ErikEJ. I merged this.

PR #8520, opened 2017-05-19
Introduce Microsoft.EntityFrameworkCore.Design namespace into the runtime

### LINKED ISSUE #8499: Move and rename IDbContextFactory
We've talked about this recently. I'm submitting an issue for us to make a decision.

The two points of dislike are:
1. It doesn't convey anything about design time
2. It's tucked away in the Infrastructure namespace

### LINKED ISSUE #8553: API Review: Make DbContextFactoryOptions obsolete
Make obsolete as error anything that is used by IDbContextFactory that will not longer work in 2.0. This will result in compile time errors with a message indicating what to do, instead of just an error that the types are missing.

bricelam:
This namespace (in Microsoft.EntityFrameworkCore.dll) includes the following hooks that applications and providers can use to influence the design-time experience.

* `[DesignTimeProviderServices]`--assembly-level attribute providers use to point to their design-time services
* `IDesignTimeDbContextFactory`--implemented by applications to instantiate their DbContext at design time (resolves #8499)
* `IDesignTimeServices`--implemented by providers and applications to add or replace design-time services

Fixes #8553

### DIFF
--- a/src/EFCore/Design/IDesignTimeDbContextFactory.cs
+++ b/src/EFCore/Design/IDesignTimeDbContextFactory.cs
@@ -1,28 +1,26 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using JetBrains.Annotations;
 
-namespace Microsoft.EntityFrameworkCore.Infrastructure
+namespace Microsoft.EntityFrameworkCore.Design
 {
     /// <summary>
     ///     A factory for creating derived <see cref="DbContext" /> instances. Implement this interface to enable
     ///     design-time services for context types that do not have a public default constructor. At design-time,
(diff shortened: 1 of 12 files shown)


src/EFCore/Design/IDesignTimeDbContextFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <typeparam name="TContext">The type of the context.</typeparam>
+    public interface IDesignTimeDbContextFactory<out TContext>
+        where TContext : DbContext
+    {
+        /// <summary>
+        ///     Creates a new instance of a derived context.
+        /// </summary>
+        /// <param name="args"> Arguments provided by the design-time service. </param>
+        /// <returns> An instance of <typeparamref name="TContext" />. </returns>
+        TContext CreateDbContext([NotNull] string[] args);
```
bricelam: Note, changed from just `Create`

Issue #7004, opened 2016-11-11
Dissolve Relational.Design into just Relational and Design


bricelam:
We've briefly talked about this before, so I'm creating an issue to make any decisions.

We should either:

- Split database creation (`DbContext.Database.EnsureCreated()` and Migrations) out of `Microsoft.EntityFrameworkCore.Relational`.

or

- Merge Reverse Engineering (`Microsoft.EntityFrameworkCore.Relational.Design`) into Relational.

This would cascade into each provider. They would either need to have one more assembly (for migrations) or combine their Design package into their main one.

bricelam:
I briefly looked into this, and I think if we merge them we need to rework some of the architecture to favor composition over inheritance. Right now, providers inherit a base implementation for generating C#, but a better factoring might come if we just ask provider's for the fragments of C# they need to provide. This would push all of our C#-generating code into EFCore.Design, and the contracts for providers could move directly into EFCore.Relational.

ajcvickers:
@smitpatel This is something we would really like to get into preview2, but @bricelam is somewhat overbooked. Could you work with @bricelam to understand what is needed and try to get this either this week or early next week?

ajcvickers:
@smitpatel How close are you to getting this merged? Can we do it today, even if there are still some open issues that need to be fixed afterwards? This change impacts the packages we produce, which impacts the build, etc., so we really need to get this part in ASAP and then tidy up any lose ends after it is in.

smitpatel:
@roji @ErikEJ & other provider writers

Resolving this issue involved doing a lot of refactoring/changes on how reverse engineering pipeline worked. Following is summary of changes which went in.

Overall design
Each provider knows how to read the database. Provider will give us the `DatabaseModel` which contains all the information we need from database. (in the form of properties or annotations).
EF will convert the DatabaseModel to IModel. It will also copy over annotations from DatabaseModel so provider specific information is flowed through.
Based on IModel, EF will write code. In this process, any annotation which EF does not know how to process, it will ask Provider on how to render the annotation. This way provider will have chance to scaffold provider specific fluent API. (this part would be reused by snapshot generation in migrations in future).


What provider has to implement now?
- Extension method `Allows` on `TableSelectionSet` (@bricelam do we want to improve this? Looks very weird)
- An implementation of `IDatabaseModelFactory` which would return a `DatabaseModel`. Providers may have provider specific annotations. To annotation objects directly or use metadata extensions for that is upto provider writer.
[...]

PR #8522, opened 2017-05-19
Negate UseFluentApiOnly and make parameter UseDataAnnotations

smitpatel:
part of #7004


smitpatel:
Updated. Removed all occurence of word `UseFluentApiOnly`.

Issue #8505, opened 2017-05-17
Query: SubQueryMemberPushDown optimization is too aggressive, which may lead to invalid sql or data corruption


maumar:
Part of QM optimization is SubQueryMemberPushDownExpressionVisitor, which pushes down member from outside subquery into it. This way we get much simpler QMs and are able to translate more queries fully to sql. Problem is that we are doing it too aggresively, e.g. query:

```cs
from l1 in ctx.LevelOne
select (from l3 in ctx.LevelThree
		orderby l3.Id
		select l3).Distinct().Skip(1).FirstOrDefault().Name
```

with the following QM (before optimization):

```
from Level1 l1 in DbSet<Level1>
select 
    (from Level3 l3 in DbSet<Level3>
    order by [l3].Id asc
    select [l3])
    .Distinct()
    .Skip(1)
    .FirstOrDefault().Name
```

will get "optimized" to:

```
from Level1 l1 in DbSet<Level1>
select 
    (from Level3 l3 in DbSet<Level3>
    order by [l3].Id asc
    select [l3].Name)
    .Distinct()
    .Skip(1)
    .FirstOrDefault()
```

which is wrong. This produces invalid sql, but even if the sql would be valid, results would be incorrect - Distinct is now applied on the Name, rather than entity itself, which will produce unexpected results if the Name has duplicated values

We should not perform the pushdown if Distinct is present in the subquery ( **other operators also?** Intersect, Except, Union)

ajcvickers:
@maumar Do you think you can get this merged today?

maumar:
I hope so, hit a bunch of issues yesterday and today but I think I finally got it

maumar:
fixed in 3f8baeb79de4edda654b0674d46bdb3e48a18a02

Issue #8528, opened 2017-05-20
Query: improve subquery member pushdown for set result operators


maumar:
Currently we don't perform the pushdown if subquery contains a set result operator (Union/Concat/Except/Intersect), the reason is that we need to do the pushdown for both query sources, and at the time inner query source (one that is part of the ResultOperator object itself) is not a query source, but a constant of EntityQueryable - we convert it to query source later during translation.

smitpatel:
@maumar - Is this really needed? We only do member pushdown if member access is on FirstOrDefault (or cardinality reducing method). Set operations are not those. And even if we don't push it down, the translated query will still fetch only required data.

smitpatel:
Subquery member pushdown in 3.0 happens only when single element operator is applied. Set operations does not do member pushdown.

PR #8529, opened 2017-05-20
Fix to #8505 - Query: SubQueryMemberPushDown optimization is too aggressive, which may lead to invalid sql or data corruption

maumar:
Problem was that we were performing subquery member pushdown incorrectly in some cases, e.g. when subquery had Distinct result operator:

select
    (from Level3 l3 in DbSet<Level3>
    select [l3])
    .Distinct()
    .FirstOrDefault().Name

previously we were translating this into:

select
    (from Level3 l3 in DbSet<Level3>
    select [l3].Name)
    .Distinct()
    .FirstOrDefault()

Fix is to introduce a subquery, if Distinct result operator is present in a subquery:

select
    (from subquery in (from Level3 l3 in DbSet<Level3>
                      select [l3])
                     .Distinct()
     select [subquery].Name)
   .FirstOrDefault()

Also added tests for a number of issues discovered when fixing this one

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/SubQueryMemberPushDownExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/SubQueryMemberPushDownExpressionVisitor.cs
@@ -1,11 +1,15 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
+using Remotion.Linq.Clauses.ResultOperators;
+using Remotion.Linq;
 
 namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 {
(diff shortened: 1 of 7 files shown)


src/EFCore/Query/ExpressionVisitors/Internal/SubQueryMemberPushDownExpressionVisitor.cs
```diff
@@ -34,20 +38,48 @@ protected override Expression VisitMember(MemberExpression memberExpression)
...
-            if (subSelector is QuerySourceReferenceExpression
-                || subSelector is SubQueryExpression)
+            if (newExpression is SubQueryExpression subQueryExpression)
             {
-                var querySourceMapping = new QuerySourceMapping();
-                var subQueryModel = subQueryExpression.QueryModel.Clone(querySourceMapping);
-                _queryCompilationContext.UpdateMapping(querySourceMapping);
+                var subSelector = subQueryExpression.QueryModel.SelectClause.Selector;
+                if ((subSelector is QuerySourceReferenceExpression || subSelector is SubQueryExpression)
+                    && !subQueryExpression.QueryModel.ResultOperators.Any(ro =>
```
maumar: added #8528 to track improvement for set result operators

src/EFCore/Query/QueryCompilationContext.cs
```diff
@@ -468,6 +473,40 @@ private void MarkForMaterialization(IQuerySource querySource)
             }
         }
 
+        private class BlockedMemberPushdownCompensatingVisitor : ExpressionVisitorBase
+        {
+            public ISet<IQuerySource> QuerySources { get; } = new HashSet<IQuerySource>();
```
smitpatel: Can this logic go in materialization EV directly?

src/EFCore/Query/ExpressionVisitors/Internal/SubQueryMemberPushDownExpressionVisitor.cs
```diff
@@ -34,20 +38,48 @@ protected override Expression VisitMember(MemberExpression memberExpression)
...
+                    if (subQueryExpression.QueryModel.ResultOperators.Last() is FirstResultOperator firstResultOperator 
+                        && firstResultOperator.ReturnDefaultWhenEmpty)
+                    {
+                        var queryModel = subQueryExpression.QueryModel;
+                        queryModel.ResultOperators.Remove(firstResultOperator);
+                        queryModel.ResultTypeOverride = null;
+                        var newSubQueryExpression = new SubQueryExpression(queryModel);
 
-                return new SubQueryExpression(subQueryModel);
+                        var mainFromClause = new MainFromClause("subquery", queryModel.SelectClause.Selector.Type, newSubQueryExpression);
```
smitpatel: use generate new name as subquery prefix using the querymodel itself to avoid clashes.

src/EFCore/Query/ExpressionVisitors/Internal/SubQueryMemberPushDownExpressionVisitor.cs
```diff
@@ -34,20 +38,48 @@ protected override Expression VisitMember(MemberExpression memberExpression)
...
+                        subQueryModel.SelectClause.Selector = VisitMember(memberExpression.Update(subQueryModel.SelectClause.Selector));
+                        subQueryModel.ResultTypeOverride = subQueryModel.SelectClause.Selector.Type;
+
+                        return new SubQueryExpression(subQueryModel);
+                    }
 
-                subQueryModel.SelectClause.Selector = VisitMember(memberExpression.Update(subQueryModel.SelectClause.Selector));
-                subQueryModel.ResultTypeOverride = subQueryModel.SelectClause.Selector.Type;
+                    if (subQueryExpression.QueryModel.ResultOperators.Last() is FirstResultOperator firstResultOperator 
+                        && firstResultOperator.ReturnDefaultWhenEmpty)
```
smitpatel: What about `LastOrDefault` or `Single`?

src/EFCore/Query/QueryCompilationContext.cs
maumar: The way I think about it is that RMEV should only mark query sources that are truly needed for materialization (present in the final projection or behind client methods). All compensation needed due to bugs and shortcomings should go somewhere else - it's easier to remove them in the future, once the bugs are addressed.

src/EFCore/Query/ExpressionVisitors/Internal/SubQueryMemberPushDownExpressionVisitor.cs
maumar: They should be ok because they force client eval (similar to SingleOrDefault that we have test for -Where_subquery_distinct_singleordefault_boolean). Will double check tho.

src/EFCore/Query/QueryCompilationContext.cs
smitpatel: Sounds good.

maumar:

src/EFCore/Query/QueryCompilationContext.cs
```diff
@@ -469,6 +472,40 @@ private void MarkForMaterialization(IQuerySource querySource)
...
+                        || ro is IntersectResultOperator
+                        || ro is ExceptResultOperator))
+                {
+                    MarkForMaterialization(subQuery.QueryModel.MainFromClause);
+                }
+
+                return base.VisitMember(node);
+            }
+
+            protected override Expression VisitSubQuery(SubQueryExpression subQueryExpression)
```
smitpatel: redundant. It is there in EVBase too.

Issue #7914, opened 2017-03-16
Make StoreGeneratedAlways the default for computed columms


ajcvickers:
Currently applications have to be careful not to mark properties that are "store generated on update" as modified because this will cause the modified value to be sent to the database which almost always causes an exception because updating the computed column is not allowed. This is why we had to introduce the StoreGeneratedAlways flag and set it for concurrency tokens--without it concurrency resolution was just too hard to do.

We should consider making this the default by convention for any property marked ValueGenerated.OnAddOrUpdate. This would match how most people want to use computed columns, while still allowing the ability to use the old behavior if somebody really wants it. Also, it would mean we no longer need to treat concurrency tokens differently. It would be technically a breaking change, but I doubt there is much (any?) real-world code that would break. 

Also see #7913 which splits StoreGeneratedAlways into two flags and renames it.

Issue #7913, opened 2017-03-16
Support read-only property behaviors


ajcvickers:
Note: read-only entities is covered by issue #7586. This issue is about read/write entities with read-only properties.

The matrix of behaviors is covered by:
* Throw if value is set/changed verses ignore any set/changed value
* Insert or Update
* Property is store generated or not

For throw scenarios:
* ReadOnlyBeforeSave throws if value is set before insert
* ReadOnlyAfterSave throws if value is changed after insert
* Doesn't matter if value is store generated or not

For ignore scenarios with store-generated values
* StoreGeneratedAlways means the value set/changed will be ignored for insert or update
* Missing: way to say store-generated on update only
* Missing: way to say StoreGeneratedAlways for insert, but not update
* Missing: way to say StoreGeneratedAlways for update, but not insert

For ignore scenarios without store-generated values
* Problem with this is that after SaveChanges the database may have a different value than the client. This seems like it is usually a bad thing, and hence if value should be ignored, then it really should be made store generated.
* However, if people are okay with the value being different, and all consequences of this (fixup for FKs, etc.), then if we support StoreGeneratedAlways for all properties (changing semantics to mean ignore set/changed value) and add missing options listed above, then this would "just work".

Proposal:
* Change StoreGeneratedAlways to IgnoreValueBeforeSave and IgnoreValueAfterSave. This allows it to be set differently for inserts and updates in the same way that read-only can be.
* Respect this flag even if property is not marked store generated
* Introduce ValueGenerator.OnUpdate to support insert a value but then read-only/ignore after that

No fluent API for now, but in the future consider adding it for most common combinations.

ajcvickers:
Design meeting notes:
* Decided to do this, although not all of it is high priority
* Would be good to combine ReadOnly... and IgnoreValue... APIs into enum, possibly obsoleting the old APIs for now
* If we add OnUpdate, then we should change the enum to be a flags enum; makes the breaking change cleaner but with higher impact, but makes it more likely that providers will react correctly to the break.

PR #8533, opened 2017-05-20
Collapse and unify property value behaviors

ajcvickers:
Part of issue #7913

This change obsoletes ReadOnlyBeforeSave, ReadOnlyAferSave, and StoreGeneratedAlways into two unified facets--one for before-save behavior (Added objects) and one for after-save behavior. In each case the property value can be:
- Ignore: value never sent to database. (Equivalent to StoreGeneratedAlways, but can apply to any property and can be set differently for before and after save.)
- Throw: any non-CLR default value set or property marked as modified causes exception. (Equivalent to previous read-only behaviors.)
- UseValue: the value is used if it has been set (to non-CLR default for store-generated) or marked as modified.

Also covers issue #7914 - makes Ignore the default for before and after save when using AddOrUpdate (computed) properties.


src/EFCore/Metadata/IProperty.cs
```diff
@@ -23,33 +23,62 @@ public interface IProperty : IPropertyBase
...
+        ///         If <see cref="PropertyValueBehavior.Ignore" />, then any value
+        ///         set will be ignored when it is in the <see cref="EntityState.Added" /> state.
+        ///     </para>
         /// </summary>
+        PropertyValueBehavior BeforeSaveBehavior { get; }
+
+        /// <summary>
+        ///     <para>
+        ///         Gets a value indicating whether or not this property can be modified before the entity is
+        ///         saved to the database.
```
lajones: "before" -> "after"

ajcvickers:
Merged

PR #8534, opened 2017-05-21
TypeMapping: Fix #8010 - Add very simple "type mappings" for non-relational providers

### LINKED ISSUE #8010: Add very simple "type mappings" for non-relational providers
This came out of the discussion of type mappings (#7434) and of issue #7960. The in-memory provider doesn't really need to do type mapping, but if does have a list of "known types" that it will map. It could be useful for applications to be able to add new entries to the list, which for some subset of types would then "just work" with the in-memory database. For other types it may still be necessary to do a type conversion (#242) which is the general solution for supporting types in providers where they are not natively supported.

All this implies that maybe IRelationalTypeMapper and related classes should have very simple non-relational base types to support this.

anpete:
Lifts concept of "mappability" to core.

- Removes derived relational conventions that decided mappability via IRelationalTypeMapper - can now be handled by core conventions.
- Many code quality fixes.


src/EFCore/Storage/CoreTypeMapper.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        /// <param name="type"> The .NET type. </param>
+        /// <returns> True if the type can be mapped; otherwise false. </returns>
+        public virtual bool IsTypeMapped(Type type)
+        {
+            Check.NotNull(type, nameof(type));
+
+            return type == typeof(string)
+                   || type.GetTypeInfo().IsValueType
+                   || type == typeof(byte[]);
```
bricelam: Is this right for Core? It feels like an InMemory-specific implementation

anpete: Yeah, I debated this myself and I think it is a good question. I think you can at least make an argument for it being in Core:

- Types satisfying this condition are able to be consumed by Core framework code (StateManager etc.) and should just work. I.e. This represents the current set of assumptions made in Core about primitive types. Aside: whether or not IsValueType is accurate enough here we should discuss - we could strengthen this condition by instead checking for IEquatable. 
- Less important, but it is often convenient to have a default impl. in Core, particularly for testing.

src/EFCore/Infrastructure/EntityFrameworkServicesBuilder.cs
```diff
@@ -180,6 +181,7 @@ public virtual EntityFrameworkServicesBuilder TryAddCoreServices()
             TryAdd<IEntityFinderSource, EntityFinderSource>();
             TryAdd<IEntityMaterializerSource, EntityMaterializerSource>();
             TryAdd<ICoreConventionSetBuilder, CoreConventionSetBuilder>();
+            TryAdd<ITypeMapper, CoreTypeMapper>();
```
ajcvickers: Override the registration of ITypeMapper in EntityFrameworkRelationalServicesBuilder so that regardless of whether a service depends on ITypeMapper or IRelationalTypeMapper, they will get the same instance. Something like:
```C#
TryAdd<ITypeMapper>(p => p.GetService<IRelationalTypeMapper>())
```

anpete: Thanks, that was missed!

src/EFCore/Storage/CoreTypeMapper.cs
anpete: Created a related bug (#8537)

ajcvickers: Yeah, think we should discuss. My initial reaction was the same as @bricelam, but now I am not sure. :-)

ajcvickers:
@lajones Heads up on this one as it will likely require a merge with your work.

PR #8536, opened 2017-05-21
Add ValueGenerated.OnUpdate

ajcvickers:
Second part of issue #7913

This change adds a new ValueGenerated option for requiring a value on insert, but then using the value from the store for subsequent updates. This is a binary breaking change, but not a functional breaking change after a re-compile.


ajcvickers:
Merged

PR #8543, opened 2017-05-22
SQL Server Migrations: SET IDENTITY_INSERT ON during INSERT

bricelam:


src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -630,6 +630,13 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
             [NotNull] InsertOperation operation,
             [CanBeNull] IModel model,
             [NotNull] MigrationCommandListBuilder builder)
+            => Generate(operation, model, builder, terminate: true);
+
+        protected virtual void Generate(
+            [NotNull] InsertOperation operation,
+            [CanBeNull] IModel model,
+            [NotNull] MigrationCommandListBuilder builder,
+            bool terminate)
```
ajcvickers: Is the "terminate" pattern something we have used before, or is this a new concept? Is it something we should add generally across operations?

bricelam: It's common. It enables two scenarios:
1. Using statements as part of a larger batch. (done here)
2. Appending SQL to the end of statements.

So far, we've just done it where needed since it's not a breaking change to add more.

PR #8544, opened 2017-05-22
Migrations: Finalize CUD APIs

### LINKED ISSUE #8471: Review and update CUD APIs for migrations
As part of the data seeding work now tracked in https://github.com/aspnet/EntityFramework/issues/629, we created CUD APIs that can be used inside migrations. 

The data seeding work is currently not scheduled to make it in 2.0, but the CUD APIs are already in.

Before we release 2.0.0 we should make sure the CUD APIs are in the state we want them to be, e.g.

1. Are compatible with the usage in the current PR https://github.com/aspnet/EntityFramework/pull/7824
2. Are very usable and we are confident we won't want to make breaking changes on them after we ship.

Otherwise we should consider either fixing them or pulling them from the release.

cc @ajcvickers

bricelam:
Fixes #8471

### DIFF
--- a/src/EFCore.Relational/Migrations/MigrationBuilder.cs
+++ b/src/EFCore.Relational/Migrations/MigrationBuilder.cs
@@ -71,7 +71,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations
             => AddForeignKey(
                 name,
                 table,
-                new[] { column },
+                new[] { Check.NotEmpty(column, nameof(column)) },
                 principalTable,
                 schema,
                 principalSchema,
@@ -120,7 +120,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations
             => AddPrimaryKey(
                 name,
                 table,
-                new[] { column },
+                new[] { Check.NotEmpty(column, nameof(column)) },
                 schema);
 
         public virtual OperationBuilder<AddPrimaryKeyOperation> AddPrimaryKey(
@@ -153,7 +153,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations
             => AddUniqueConstraint(
                 name,
                 table,
-                new[] { column },
+                new[] { Check.NotEmpty(column, nameof(column)) },
                 schema);
 
         public virtual OperationBuilder<AddUniqueConstraintOperation> AddUniqueConstraint(
@@ -305,7 +305,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations
             => CreateIndex(
                 name,
                 table,
-                new[] { column },
+                new[] { Check.NotEmpty(column, nameof(column)) },
                 schema,
                 unique,
                 filter);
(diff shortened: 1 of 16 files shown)


src/EFCore.Relational/Migrations/MigrationBuilder.cs
```diff
@@ -682,11 +682,20 @@ public virtual AlterOperationBuilder<AlterDatabaseOperation> AlterDatabase()
 
         public virtual OperationBuilder<InsertOperation> Insert(
```
bricelam: Will rename to `InsertData`, etc.

src/EFCore.Relational/Migrations/MigrationBuilder.cs
```diff
@@ -697,18 +706,27 @@ public virtual AlterOperationBuilder<AlterDatabaseOperation> AlterDatabase()
...
-            [CanBeNull] object[,] keyValues = null)
+            [NotNull] string[] keyColumns,
+            [NotNull] object[] keyValues,
+            [CanBeNull] string schema = null)
+            => DeleteData(table, keyColumns, ToMultidimensionalArray(Check.NotNull(keyValues, nameof(keyValues))), schema);
+
+        public virtual OperationBuilder<DeleteDataOperation> DeleteData(
+            [NotNull] string table,
+            [NotNull] string[] keyColumns,
+            [NotNull] object[,] keyValues,
```
bricelam: I wonder if we should have overloads for a single key column. We do this in other methods that take key columns.

ajcvickers: Sounds reasonable. @divega?

divega: If it is sugar for the common case, sounds great.

bricelam: Updated. Love the terseness!

bricelam:
PR submitted: dotnet/templating#851

bricelam:
@divega Here's how I see it:

Now
===
The base `MigrationsSqlGenerator` will use the `UpdateDataOperation` objects directly to generate standard DML. If necessary (unlikely), providers can override the methods in their Migrations SQL generator to tweak it.

Later
===
We update our implementation in `MigrationsSqlGenerator` to convert the `UpdateDataOperation` objects into `ModificationCommandBase` objects and use the Update pipeline to generate the DML. Providers can safely remove their Migrations overrides since their Update pipeline implementations will become the source of truth.

divega:
@bricelam I haven't looked at https://github.com/aspnet/EntityFramework/pull/7824/ in too much detail, but maybe you have :smile: 

Did you see anything specific there that makes you think we shouldn't make the change to `IUpdateSqlGenerator` in 2.0 or do you just think it is worth doing it?

bricelam:
@AndriySvyryd Would know better about how much work is required, but I definitely think it's worth doing if we have time. The changes to Migrations are trivial (just converting from one data structure to another).

PR #8545, opened 2017-05-22
Add sugar for determining the database provider in use

ajcvickers:
Part of #7166

Main scenario here is conditional behavior in OnModelCreating now that we build a different model per provider.

### DIFF
--- a/src/EFCore/Infrastructure/DatabaseFacade.cs
+++ b/src/EFCore/Infrastructure/DatabaseFacade.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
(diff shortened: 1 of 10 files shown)


src/EFCore/Infrastructure/DatabaseFacade.cs
```diff
@@ -184,6 +186,25 @@ public virtual IDbContextTransaction CurrentTransaction
...
+        ///     <para>
+        ///         This method can only be used after the <see cref="DbContext" /> has been configured because
+        ///         it is only then that the provider is known. This means that this method cannot be used
+        ///         in <see cref="DbContext.OnConfiguring" /> because this is where application code sets the
+        ///         provider to use as part of configuring the context.
+        ///     </para>
+        /// </summary>
+        public virtual string ProviderName
+            => _context.GetService<IEnumerable<IDatabaseProvider>>()
+                ?.Select(p => p.InvariantName)
```
bricelam: Can `GetService<IEnumerable>()` return `null`?

test/EFCore.SqlServer.Tests/SqlServerDatabaseFacadeTest.cs
```diff
@@ -0,0 +1,183 @@
...
+            using (var context = new SqlServerOnConfiguringContext())
+            {
+                Assert.True(context.Database.IsSqlServer());
+            }
+        }
+
+        [Fact]
+        public void IsSqlServer_in_OnModelCreating_when_using_OnConfguring()
+        {
+            using (new SqlServerOnModelContext())
```
anpete: ?

bricelam:
We should add matching sugar to MigrationBuilder. (filed #8566)

src/EFCore/Infrastructure/DatabaseFacade.cs
ajcvickers: It's never been very clear in the spec--it may depend on the D.I. container being used.

test/EFCore.SqlServer.Tests/SqlServerDatabaseFacadeTest.cs
ajcvickers: There are asserts in the context below. But I will add some more code that checks that the asserts actually get hit.

src/EFCore/Infrastructure/DatabaseFacade.cs
bricelam: Ah, good point.

roji:
Am I understanding correctly that this won't be merged for 2.0.0? Would it be possible to specify the milestone so it's easier to understand when you guys are planning to release things?

ajcvickers:
@roji It probably will be, it's just that @bricelam created the issue and completed the work before we had a chance to triage it and put it in a milestone. He's that good. 😉

ajcvickers:
@roji Or are you asking about this PR specifically? If so, then the issue is in a milestone. We put issues in milestones rather than PRs.

roji:
@ajcvickers oh ok, thanks :) Will try to get it in for Npgsql's 2.0.0 release too then!

Issue #7131, opened 2016-11-25
Projections With Full Types Not Fixing Up Navigations on Sqlite and SQL Server


julielerman:
I've searched for a related issue as well as blog posts/docs and am not finding anything. I know it's out there somewhere but have to stop looking! 

Also related to https://github.com/aspnet/EntityFramework/issues/4007 

### Steps to reproduce

```
using System.Collections.Generic;
using System.Linq;
using Microsoft.EntityFrameworkCore;

namespace ProjectionProblem
{
  public class Parent
  {
    public Parent()
    {
      Children = new List<Child>();
    }

    public int Id { get; set; }
    public string Description { get; set; }
    public List<Child> Children { get; set; }
  }

  public class Child
  {
    public int Id { get; set; }
    public string Description { get; set; }
  }

  public class MyContext : DbContext
  {
    public DbSet<Parent> Parents { get; set; }

    protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
    {
      optionsBuilder.UseSqlite("Data source=ProjectionProblem.db");
    }
  }

  public class Program
  {
    private static void Main(string[] args)
    {
      using (var context = new MyContext())
      {
        context.Database.EnsureDeleted();
        context.Database.EnsureCreated();
        var parent = new Parent {Description = "Parent1"};
        parent.Children.Add(new Child {Description = "Child1"});
        parent.Children.Add(new Child {Description = "Child2"});
        context.Parents.Add(parent);
        context.SaveChanges();
      }
      using (var context2 = new MyContext())
      {
        var newtype = context2.Parents.Select(p => new {Parent = p, p.Children}).ToList();
      }
    }
  }
}
```

### The issue
Navigations not being fixed up in results. newtype.Children has the child objects, but newtype.Parent.Children is empty.

### Further technical details

EF Core version: 1.1
Operating system: 
Visual Studio version: vs2015 but really not applicable

Other details about my project setup:

julielerman:
Just to add to this...other projections tests are also quite messy. Again, this is probably noted somewhere that I'm just not finding ...not mentioned in the roadmap or in this repo but that could be affected by how I'm searching.

Filtered projections .. e.g.
 var newtype = context3.Parents.Select(p => 
          new { Parent = p, Children=p.Children.Where(c=>c.Description=="Child2") }).ToList();

The queries are strange. Non of them have any reference to the children. 
The results are strange. What's in the Children property is a ParameterInjector<Child>

[...]

Issue #8227, opened 2017-04-19
Query: Navs: MaterializeCollectionNavigation needs async version


anpete:
CollectionNavigationSubqueryInjector always injects a sync call to MaterializeCollectionNavigation even when the query is async.

smitpatel:
Fixed in new pipeline. Now instead of injecting sync/async method, we keep it as an custom expression and always translate to server.

PR #8556, opened 2017-05-23
Fix to #7131 - Projections With Full Types Not Fixing Up Navigations on Sqlite and SQL Server

maumar:
Problem was that entities created by MaterializeCollectionNavigation were not being tracked - TrackEntities only works on "naked" entities.

Fix is to add delegate argument to MaterializeCollectionNavigation method that would track the elements of the collection if the query is supposed to be tracked. We already have a similar logic for the Include.


maumar:
recalling for now, will try to tackle #8227 also, since they are in the same area and require significant refactoring

PR #8567, opened 2017-05-23
Fix #8557 - Adds parameterized interpolated string support to Database.ExecuteSqlCommand.

### LINKED ISSUE #8557: Update ExecuteSqlCommand methods to use string interpolation
Following same pattern as FromSql

anpete:
Adds parameterized interpolated string support to Database.ExecuteSqlCommand.

### DIFF
--- a/src/EFCore.Relational.Specification.Tests/SqlExecutorTestBase.cs
+++ b/src/EFCore.Relational.Specification.Tests/SqlExecutorTestBase.cs
@@ -3,15 +3,16 @@
 
 using System;
 using System.Data.Common;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Specification.Tests;
 using Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.Northwind;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
+// ReSharper disable ConvertToConstant.Local
 
-namespace Microsoft.EntityFrameworkCore.Specification.Tests
+namespace Microsoft.EntityFrameworkCore.Relational.Specification
 {
     public abstract class SqlExecutorTestBase<TFixture> : IClassFixture<TFixture>
         where TFixture : NorthwindQueryFixtureBase, new()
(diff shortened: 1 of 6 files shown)


src/EFCore.Relational.Specification.Tests/SqlExecutorTestBase.cs
```diff
@@ -3,15 +3,16 @@
...
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Specification.Tests;
 using Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.Northwind;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
+// ReSharper disable ConvertToConstant.Local
 
-namespace Microsoft.EntityFrameworkCore.Specification.Tests
+namespace Microsoft.EntityFrameworkCore.Relational.Specification
```
anpete: Yeah, didn't match project file.

smitpatel: AFAIK all files in Relational.Specification.Tests are in EFCore.Specification.Tests namespace.

anpete: Could be but that is not what it says in the project file. We are very inconsistent here and consequently generate a lot of R# warnings about incorrect namespaces. I would like to fix this, @ajcvickers, can you remember if we had decided on a standard convention for test project namespaces? I'm happy with any approach that is consistent :-)

ajcvickers: We did decide a long time ago, but we never really implemented the decision consistently. Also, it was a decision based on what all of ASP.NET was going to do, which I don't think is relevant anymore. Happy for us to discuss again and come up with a decision and then implement it.

anpete: Many of the test projects have "Microsoft.EntityFrameworkCore" as the default namespace. I'm happy to just make this our standard. Thoughts?

cc @bricelam

ajcvickers: Sounds good to me.

Issue #7434, opened 2017-01-18
Self-contained type mappings


divega:
Adding support for a new type on a relational database provider currently requires making changes in a few places:

1. The provider's implementation of `IRelationalTypeMapper`: used to populate parameters, configure column facets, etc.
2. The provider's implementation of `ISqlGenerationHelper`: used to generate literal values for a given type in SQL
3. Potentially, `CSharpUtilities` in `Microsoft.EntityFrameworkCore.Relational.Design` (and the equivalent types in migrations) to generate C# literals for the type
4. The definition of primitive operations/functions over the type and their LINQ translators
... and maybe other things I am missing.

This issue is about refactoring our code to get closer to a situation in which a type mapping can convey the necessary knowledge to perform all or most of those responsibilities for a given type, so that the type mapping can be added to the provider's type mapper and no further actions are required to support the new type. 

We should take into account: 

1. This should be a stepping stone towards #242, which is about having user defined type mappings, and not push us further from it.
2. This should align with #5410, which is about enabling the use of `ScaffoldingTypeMappper` in scaffolding. 
3. Can we unify the various versions of the code that generate C# (literals and otherwise) between scaffolding and migrations?
4. Does the new factoring help us decide whether we should merge the reverse engineering package into relational (one of the options described in #7004)?

per3ian:
I am trying to add `Microsoft.EntityFrameworkCore.Relational.Design ` and it added to my project.json :

```
{
  "version": "1.0.0-*",

  "dependencies": {
    "xUtilities": "1.0.0-*",
    "Microsoft.EntityFrameworkCore.Relational": "1.1.1",
    "Microsoft.EntityFrameworkCore.Relational.Design": "1.1.1",
    "Microsoft.AspNetCore.Identity.EntityFrameworkCore": "1.1.1",
    "Microsoft.EntityFrameworkCore": "1.1.1",
    "Microsoft.EntityFrameworkCore.Design": "1.1.1",
    "Microsoft.EntityFrameworkCore.SqlServer": "1.1.1",
    "Microsoft.EntityFrameworkCore.SqlServer.Design": {
      "version": "1.1.1",
      "type": "build"
    },
    "Microsoft.EntityFrameworkCore.Tools": "1.1.0",
    "NETStandard.Library": "1.6.1",
  },
  "tools": {
    "Microsoft.EntityFrameworkCore.Tools": "1.1.0-preview4-final"
  },
  "frameworks": {
    "netstandard1.6": {
[...]

PR #8568, opened 2017-05-23
Make TypeMappings self-contained (for SQL literal generation)

lajones:
This fixes the scoped-down part of #7434 - see [this comment](https://github.com/aspnet/EntityFramework/issues/7434#issuecomment-303237523) - which addresses making SQL literal generation the sole province of the `RelationalTypeMapping` class (and no longer part of `ISqlGenerationHelper`).


src/EFCore.Relational/Storage/RelationalTypeMapping.cs
```diff
@@ -105,6 +115,53 @@ public virtual RelationalTypeMapping CreateCopy([NotNull] string storeType, int?
...
+        {
+            { System.Data.DbType.Byte, "tinyint" },
+            { System.Data.DbType.Decimal, "decimal" },
+            { System.Data.DbType.Double, "float" },
+            { System.Data.DbType.Int16, "smallint" },
+            { System.Data.DbType.Int32, "int" },
+            { System.Data.DbType.Int64, "bigint" },
+            { System.Data.DbType.String, "nvarchar" },
+            { System.Data.DbType.Date, "date" }
+        };
```
lajones: Note: the above was just moved from `RelationalSqlGenerationHelper` and supports the `GenerateLiteral(DbType)` below which is used by the query pipeline. It's odd for a number of reasons and I don't think it should ever have been in `RelationalSqlGenerationHelper` to start with because what is really being asked for is a mapping from `DbType` -> storeType. But I talked this over with @maumar and the correct fix is more complicated. We decided we'd be better off doing the fix for that issue separately and later.

src/EFCore.Relational/Storage/RelationalTypeMapping.cs
```diff
@@ -187,5 +244,188 @@ public virtual RelationalTypeMapping CreateCopy([NotNull] string storeType, int?
...
+        /// </returns>
+        public virtual string GenerateSqlLiteral([CanBeNull]object value)
+        {
+            if (value == null)
+            {
+                return "NULL";
+            }
+
+            var s = value as string;
+            return s != null ? GenerateSqlStringLiteralValue(s) : GenerateSqlLiteralValue((dynamic)value);
```
ajcvickers: Do we still need a special method for strings?

lajones: The problem was the (dynamic) call. Almost everything can be converted to a string and so a `GenerateSqlLiteral(string)` method could get called when you meant to call one of the other overloads. So, yes, I believe so.

ajcvickers: Interesting. Which tests failed?

lajones: I haven't tried it this time - that was just my experience the last time I used `dynamic` - so I recognized the reason the original `RelationalSqlGenerationHelper` code had special treatment for strings and didn't attempt to change it. I can try it out if you like?

ajcvickers: Yes, please do.

lajones: Tests are fine and I'm unable to repro the problem I remember seeing with dynamic casting - so I'll go ahead and change it.

lajones:
Checked in with commit 4b8b12d01.

PR #8573, opened 2017-05-24
API review: Remove UseTransientInMemoryDatabase

ajcvickers:
It's not very useful, and it causes confusion.


test/EFCore.CrossStore.FunctionalTests/ProviderSpecificServicesTest.cs
```diff
@@ -18,7 +18,7 @@ public void Throws_with_new_when_non_relational_provider_in_use()
                     new ServiceCollection()
                         .AddEntityFrameworkInMemoryDatabase()
                         .BuildServiceProvider())
-                .UseTransientInMemoryDatabase()
+                .UseInMemoryDatabase(Guid.NewGuid().ToString())
```
anpete: Should we have a no-arg version that does the guid name by default?

divega: Would a no-arg version that generated a random name make it harder to clean up after use? I.e. users would need to create a new service provider instead of just `EnsureDeleted()`, correct?

FWIW, I filed https://github.com/aspnet/EntityFramework/issues/8548/ suggesting we add back the no-arg version as obsolete.

ajcvickers: It's interesting that @anpete is asking for a no-args version that acts like the scoped database, while @divega is suggesting we bring it back as the singleton database. To me this is a good sign that the meaning of a no-args version is always going to be ambiguous, and hence we should probably only have it as obsolete to move people away from it.

poke:
I just stumbled upon this while migrating an older 1.0-rc1 application to 2.0.

What would be the recommended way to create a short-living in-memory database for a single unit test now? I have a lot tests that look like this, where the context should not be shared across multiple tests:

```c#
var options = new DbContextOptionsBuilder<DataContext>().UseInMemoryDatabase("foo").Options;
using (var db = new MyContext(options))
{
    // add some test fixtures
}

using (var db = new MyContext(options))
{
    // do something with those fixtures
}
```

Using `Guid.NewGuid().ToString()` would of course work here but doing so still feels like the in-memory database is unnecessarily persisted across multiple unit test executions (after all, if you use the same name, you do get access to the previous data). So I would really like a way to throw the database away after the test ran, to potentially free up the memory.

Is there a way to enforce that? Or would it be better to actually share the database but call `db.Database.EnsureDeleted()` repeatedly in every test?

ajcvickers:
@poke In our own tests we use a different database name for each independent test--e.g. the test name. To clean up, use EnsureDeleted. In other words, treat it like any other database.

You could also choose to keep every unit test isolated from every other unit test by managing the service provider directly and calling UseInternalServiceProvider, but the overhead of this is probably not worth it.

poke:
@ajcvickers I see, the “treat it like any other database” makes it clear. Thanks for the suggestions :)

PR #8585, opened 2017-05-25
Remove Sqlite.Design

smitpatel:
@bricelam - Need to discuss where should Relational.Design.Specs test go. They are provider specific and also based on EFCore.Design.

This change passes all tests in SQLite. 😄


src/EFCore.Sqlite.Core/Design/Internal/SqliteDatabaseModelFactory.cs
```diff
@@ -68,9 +71,22 @@ public virtual DatabaseModel Create(string connectionString, TableSelectionSet t
...
+                // schema selections so mark all of them as matched so that we don't
+                // also log warnings about not matching each individual selection
+                tableSelectionSet.Schemas.ToList().ForEach(s => s.IsMatched = true);
+            }
+
             using (var connection = new SqliteConnection(connectionString))
             {
-                return Create(connection, tableSelectionSet);
+                var databaseModel = Create(connection, tableSelectionSet);
+                databaseModel.AddAnnotation(ScaffoldingAnnotationNames.UseProviderMethodName, nameof(SqliteDbContextOptionsBuilderExtensions.UseSqlite));
```
bricelam: This probably doesn't belong here.

src/EFCore.Sqlite.Core/Design/Internal/SqliteDatabaseModelFactory.cs
```diff
@@ -7,16 +7,19 @@
...
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Migrations;
+using Microsoft.EntityFrameworkCore.Scaffolding;
+using Microsoft.EntityFrameworkCore.Scaffolding.Internal;
 using Microsoft.EntityFrameworkCore.Scaffolding.Metadata;
+using Microsoft.EntityFrameworkCore.Scaffolding.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 
-namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
+namespace Microsoft.EntityFrameworkCore.Design.Internal
```
bricelam: I think keeping these under `Scaffolding` would be more appropriate.

src/EFCore.Sqlite.Core/Design/Internal/SqliteDesignTimeServices.cs
```diff
@@ -1,12 +1,12 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.EntityFrameworkCore.Design;
+using Microsoft.EntityFrameworkCore.Scaffolding;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Storage.Internal;
 using Microsoft.Extensions.DependencyInjection;
 
-namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
+namespace Microsoft.EntityFrameworkCore.Design.Internal
```
bricelam: This one definitely belongs in `Design`.

src/EFCore.Sqlite.Core/Properties/SqliteStrings.resx
```diff
@@ -134,4 +134,16 @@
   <data name="SequencesNotSupported" xml:space="preserve">
     <value>SQLite does not support sequences. For more information, see http://go.microsoft.com/fwlink/?LinkId=723262.</value>
   </data>
+  <data name="LogUsingSchemaSelectionsWarning" xml:space="preserve">
+    <value>Scaffolding from a SQLite database will ignore any schema selection arguments.</value>
```
bricelam: Their next question will be, "Why?" Something like this might be better: "SQLite doesn't support schemas. The specified schema selection arguments will be ignored."

src/EFCore.Sqlite.Core/Design/Internal/SqliteDatabaseModelFactory.cs
smitpatel: So to be consistent
`Migrations` namespace will have migration specific code
same goes for `Scaffolding`
And `Design` will have code common to both (more precisely any interaction with PMC/command line)
right?

bricelam:
Re the Relational.Design.Specs tests: I've always hated how end-to-end these were. We can keep them as is for now (don't merge them) and file an issue to remove their dependency on Design.

smitpatel:
Updated
```
    Remove Sqlite.Design package
    Remove Sqlite.Design.Tests
    Introduce IScaffoldingHelper interface which provider overwrites to generate provider specific code in scaffolding pipeline
```

PR #8588, opened 2017-05-25
React to breaking change latest API Check detects

dougbu:
- see PR aspnet/BuildTools#259 and issue aspnet/BuildTools#146


dougbu:
@ajcvickers was this bit of 09b1503ef9 an intentional breaking change? If not, we should revert that before I check aspnet/BuildTools#259 in.

ajcvickers:
@dougbu We review all our breaking changes through a different process, so you can assume that anything detected by the tool is intentional.

dougbu:
284f1e6c8

PR #8592, opened 2017-05-25
Tools: Block .NET Standard startup projects

### LINKED ISSUE #8452: Tools: Block .NET Standard startup projects
We've limped support for .NET Standard startup projects along, but in practice, they rarely work.

We should disable them and direct users down the path of either cross-targeting or using a startup project.

This would resolve #8344 and supersede #8349.

bricelam:
Fixes #8452

### DIFF
--- a/src/EFCore.Tools/tools/EntityFrameworkCore.psm1
+++ b/src/EFCore.Tools/tools/EntityFrameworkCore.psm1
@@ -773,15 +773,8 @@ function EF($project, $startupProject, $params, [switch] $skipBuild)
                 'a different platform and try again.'
         }
     }
-    elseif ($targetFramework -in '.NETCoreApp', '.NETStandard')
+    elseif ($targetFramework -eq '.NETCoreApp')
     {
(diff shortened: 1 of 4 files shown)


src/EFCore.Tools/tools/EntityFrameworkCore.psm1
```diff
@@ -815,6 +808,14 @@ function EF($project, $startupProject, $params, [switch] $skipBuild)
 
         $params = $dotnetParams + $params
     }
+    elseif ($targetFramework -eq '.NETStandard')
+    {
+        throw "Startup project '$($startupProject.ProjectName)' targets framework '.NETStandard'. There is no " +
+            'runtime associated with this framework, and projects targeting it cannot be executed directly. To use ' +
+            'the Entity Framework Core Package Manager Console Tools with this project, add an executable project ' +
+            'targeting .NET Framework or .NET Core that references this project, and set it as the startup project; ' +
+            'or, update this project to cross-target .NET Framework or .NET Core.'
```
ajcvickers: 👍

PR #8595, opened 2017-05-25
Only target .NET Standard 2.0

bricelam:
The only remaining `#if`s are for `AppDomain`s in the tools!


natemcmaster:
I think you set a new lines-deleted record

image

PR #8601, opened 2017-05-25
Better thread safety in InheritanceFixtureBase

anpete:
Make InheritanceFixtureBase.IsSeeded flag static so that we don't need to rely on EnsureCreated to ensure seeding only runs once.


smitpatel:
@anpete - Can you merge this?

anpete:
Merged last night.

smitpatel:
Guess I recommitted same commit 😛

Issue #8034, opened 2017-03-30
FindProperty(<Name>).Relational().ColumnType is null if column type is not specifically set


JonPSmith:
Describe what is not working as expected.

I am trying to obtain the database column type by using `entityType.FindProperty(<PropertyName>).Relational()` and I get null for `ColumnType` for string properties. If I use the same command on another, ~~non-string property it works.~~ property where I have set the type then it contains that type.

### Steps to reproduce
Include a complete code listing (or project/solution) that we can run to reproduce the issue.

Unit Test
```c#
[Fact]
public void TestGetColumnRelationalTypeOk()
{
    //SETUP
    var inMemDb = new SqliteInMemory();

    //ATTEMPT
    using (var context = inMemDb.GetContextWithSetup())
    {
        var colType = context.GetColumnRelationalType(new Book(), p => p.Description);

        //VERIFY
        colType .ShouldEqual("nvarchar(max)");
    }
}
```
The `GetColumnRelationalType` extension method is
```c#
public static string GetColumnRelationalType<TEntity, TProperty>(this DbContext context,
    TEntity source, Expression<Func<TEntity, TProperty>> model) where TEntity : class
{
    var efType = context.Model.FindEntityType(typeof(TEntity).FullName);
    var propInfo = GetPropertyInfoFromLambda(model);
    var relational = efType.FindProperty(propInfo.Name).Relational();
    return relational.ColumnType;
}
```
My Book Entity is 
```c#
public class Book                        
{
    public const int WebUrlMaxlength = 512; 

    public int BookId { get; set; }

    [Required] 
    [MaxLength(256)] 
    public string Title { get; set; }
    public string Description { get; set; }
    public DateTime PublishedOn { get; set; }
    [MaxLength(64)] 
    public string Publisher { get; set; }
    public decimal Price { get; set; }

    [MaxLength(WebUrlMaxlength)] 
    public string ImageUrl { get; set; }

    //----------------------------------------------
    //relationships

    public PriceOffer Promotion { get; set; }       
    public ICollection<Review> Reviews { get; set; }
    public ICollection<BookAuthor> 
        AuthorsLink { get; set; }                   
}
```
I haven't included my DbContext, but it is a standard layout with a `DbSet<Book> Books` property.

I have tried this with in-memory Sqlite and with SQL Server - both return null.

### Further technical details
EF Core version: 1.1.1
[...]

PR #8605, opened 2017-05-26
Make ColumnType return the mapped store type even when not set explicitly

ajcvickers:
Issue #8034, possible because of #7166

The idea here is to generate type mappings at model building type and then store them in the model. This allows `property.RelationalAnnotationNames().ColumnType` to always return the store type.

The solution is not perfect:
- A convention runs when the model is built, which means that ColumnType will remain null during OnModelCreating. This could be changed by lots of conventions that run whenever something that could change the mapping is changed.
- Because this is done by convention, and because conventions are not always run (e.g. ModelSnapshot, HistoryTable, using the ModelBuilder directly with an empty convention set) it means sometimes type mappings are not available. However, in the normal application scenario, they will always be available.

I will file an issue to look at these things post 2.0.

### DIFF
--- a/src/EFCore.Relational/Metadata/Conventions/Internal/RelationalTypeMappingConvention.cs
+++ b/src/EFCore.Relational/Metadata/Conventions/Internal/RelationalTypeMappingConvention.cs
@@ -1,0 +1,43 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Storage;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    /// <summary>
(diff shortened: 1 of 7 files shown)


src/EFCore.Relational/Metadata/Conventions/Internal/RelationalTypeMappingConvention.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual InternalModelBuilder Apply(InternalModelBuilder modelBuilder)
+        {
+            foreach (var property in modelBuilder.Metadata.GetEntityTypes().SelectMany(e => e.GetDeclaredProperties()))
+            {
+                new InternalPropertyBuilder(property, modelBuilder)
```
smitpatel: Is `property.Builder` unavailable here?

ajcvickers: Fixed.

ajcvickers:
Merged

Issue #8558, opened 2017-05-23
Propose name/location for non-interpolating string type


ajcvickers:

divega:
**Proposal:** `Microsoft.EntityFrameworkCore.RawSqlString`

The type...
- Is not nested and can be shared between `FromSql()` and `ExecuteSqlCommand()`
- Conveys an appropriate meaning when it shows up freely in Intellisense under the EFCore namespace
- Helps differentiate from the overload that takes `FormattableString`  when it shows up in Intellisense. 
- Aligns with the way we currently explain/document both overloads, e.g. for `FromSql()`:
  - `FormattableString` overload -> _Creates a LINQ query based on an interpolated string SQL query._
  -  `RawSqlString` overload -> _Creates a LINQ query based on a raw SQL query._

divega:
@ajcvickers, I talked to @anpete and he won't push back on this proposal. Let me know if you think we should still discuss it in person before he makes the change.

ajcvickers:
:shipit:

ajcvickers:
Poaching since @anpete is out today.

PR #8614, opened 2017-05-26
API review: rename SqlFormat to RawSqlString

ajcvickers:
Issue #8558

### DIFF
--- a/src/EFCore.Relational/RelationalQueryableExtensions.cs
+++ b/src/EFCore.Relational/RelationalQueryableExtensions.cs
@@ -47,7 +47,7 @@ namespace Microsoft.EntityFrameworkCore
         ///     An <see cref="IQueryable{T}" /> to use as the base of the raw SQL query (typically a <see cref="DbSet{TEntity}" />).
         /// </param>
         /// <param name="sql">
-        ///     The raw SQL query. NB. A string literal may be passed here because <see cref="SqlFormat" />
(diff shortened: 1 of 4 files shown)


src/EFCore.Relational/RelationalQueryableExtensions.cs
```diff
@@ -107,20 +107,8 @@ public static class RelationalQueryableExtensions
                     null,
                     FromSqlMethodInfo.MakeGenericMethod(typeof(TEntity)),
                     source.Expression,
-                    Expression.Constant(new SqlFormat(sql.Format)),
+                    Expression.Constant(new RawSqlString(sql.Format)),
```
divega: To replace in xml docs:
- Summary:  
  > Creates a LINQ query based on an interpolated string **representing** a SQL query.
- Param sql: 
  > The interpolated string **representing** a SQL query.

@ajcvickers feel free to use "containing" or "representing" or any better idea you might have.

src/EFCore.Relational/RelationalQueryableExtensions.cs
```diff
@@ -47,15 +47,15 @@ public static class RelationalQueryableExtensions
         ///     An <see cref="IQueryable{T}" /> to use as the base of the raw SQL query (typically a <see cref="DbSet{TEntity}" />).
```
divega: Replace in xml docs:
> This overload also accepts DbParameter instances as parameter values. This allows you to use named parameters in the SQL query string -

PR #8620, opened 2017-05-26
SQL Server Migrations: Only SET IDENTITY_INSERT when IDENTITY

### LINKED ISSUE #8618: Migrations: InsertData() causes error "Table 'table_name' does not have the identity property. Cannot perform SET operation."
Our implementation for SQL Server is incorrect because `SET IDENTITY_INSERT table_name ON` is only valid when the table does actually have an identity column. I see a few options:

1. Have a flag indicating if the statement should be generated
2. Have a separate method e.g. `InsertDataOverridingIdentity` :smile_cat:
3. Query for schema metadata in the SQL (might not work if we don't have enough permissions).

bricelam:
Fixes #8618

### DIFF
--- a/src/EFCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
+++ b/src/EFCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
@@ -828,21 +828,55 @@ namespace Microsoft.EntityFrameworkCore.Migrations
             Check.NotNull(operation, nameof(operation));
             Check.NotNull(builder, nameof(builder));
 
+            builder.Append("IF EXISTS (SELECT * FROM [sys].[identity_columns] WHERE [object_id] = OBJECT_ID(N'");
+
+            if (operation.Schema != null)
+            {
+                builder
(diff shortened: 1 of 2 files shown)


src/EFCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -828,21 +828,55 @@ protected override void Generate([NotNull] SqlOperation operation, [CanBeNull] I
             Check.NotNull(operation, nameof(operation));
             Check.NotNull(builder, nameof(builder));
 
+            builder.Append("IF EXISTS (SELECT * FROM [sys].[identity_columns] WHERE [object_id] = OBJECT_ID(N'");
```
ajcvickers: Do we know we will have permission for this query?

bricelam: No idea, but we query sys tables in other places too.

ajcvickers: Thought about it more and I think it's fine for code in Migrations that is already doing schema stuff.

Issue #7185, opened 2016-12-05
Better way to get application services from within a derived DbContext instance


alaatm:
I am trying to get a service from within a derived `DbContext` using the following code:

    ((IInfrastructure<IServiceProvider>)this).GetService<IMyService>();

The `IMyService` was previously registered in `ConfigureServices` in the `Startup` class. However, the call returns `null`. I am assuming this is because EF creates a service scope. 

My question is how can I get a service that was registered in the `Startup` class from within a derived `DbContext` instance. Or, if that makes more sense, how do I register a service in EF's service scope.

ajcvickers:
@alaatm Add a constructor parameter for the service so that it is constructor-injected into the context in the normal way.

PinpointTownes:
@ajcvickers IMHO, not being able to access the application services from EF code is really annoying, as it makes a bunch of scenarios really hard to implement. For instance, you can't create a `IModelCustomizer` that relies on a service defined in `Startup.ConfigureServices`, as EF uses its own internal service provider.

@ajcvickers was there a particular reason not to make the application service provider accessible from EF's infrastructure stuff?

ajcvickers:
@alaatm @PinpointTownes If you need to use application services as dependencies for replacements for some of EF's internal services, then you can merge the service provider's together. See this blog post for more details: https://blog.oneunicorn.com/2016/10/27/dependency-injection-in-ef-core-1-1/. Note that this does not usually require that you use the context as a service locator--it would still be better to use constructor injection for your dependencies.

PinpointTownes:
@ajcvickers thanks for the link!

Unfortunately, replacing the "internal" service provider is not a good option for ASP.NET Core apps, as the `IServiceProvider` is not available from `ConfigureServices` (well you could call `BuildServiceProvider()` to create a provider instance just for EF, but it's really not ideal).

Concerning the new `ReplaceService` option, I suppose it doesn't resolve the replaced service in the context of the application service?

ajcvickers:
@PinpointTownes This is the usual way to replace the internal service provider for ASP.NET apps:
```C#
services
    .AddEntityFrameworkInMemoryDatabase()
    .AddDbContext<MyContext>((p, b) => b
        .UseInMemoryDatabase()
        .UseInternalServiceProvider(p));
```

With regards to ReplaceService, EF has no knowledge of the application service provider unless it passed to EF using UseInternalServiceProvider.

PinpointTownes:
[...]

PR #8622, opened 2017-05-27
Make context.GetService<> fallback to the app service provider

ajcvickers:
Issue #7185

This makes it easier for applications that are using AddDbContext to get access to services registered on the container that was used for AddDbContext.


HaoK:
Awesome thanks!

PR #8629, opened 2017-05-28
Revert back to parameter logging format from v1 and update test baselines

ajcvickers:
Issue #8456

We previously used a different format for parameter logging in our tests when compared to product code logging. Now we use the same format everywhere. This change updates the product format back to what it was, and updates the test baselines to reflect this.

### DIFF
--- a/test/EFCore.Relational.Tests/Storage/RelationalCommandTest.cs
+++ b/test/EFCore.Relational.Tests/Storage/RelationalCommandTest.cs
@@ -933,7 +933,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
             foreach (var item in log)
             {
                 Assert.EndsWith(
-                    @"[Parameters=[FirstParameter: ?], CommandType='0', CommandTimeout='30']
+                    @"[Parameters=[FirstParameter='?'], CommandType='0', CommandTimeout='30']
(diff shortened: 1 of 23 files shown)


test/EFCore.Relational.Tests/Storage/RelationalCommandTest.cs
```diff
@@ -989,7 +989,7 @@ public static TheoryData CommandActions
             foreach (var item in log.Skip(1))
             {
                 Assert.EndsWith(
-                    @"[Parameters=[FirstParameter: 17], CommandType='0', CommandTimeout='30']
+                    @"[Parameters=[FirstParameter='17'], CommandType='0', CommandTimeout='30']
```
divega: Nit: Did we ever consider not including CommandType in this case? I believe it will be 0 in not specified, although documentation says default is `Table`, neither is very interesting. Also, do we always log the numeric value or do we log the name of the enum member?

ajcvickers: @divega Don't know. We never set it, so it will always be zero, which is not a valid enum value and hence is shown like this. This was done a long time ago--I doubt it was really thought about, but I don't know.

PR #8631, opened 2017-05-29
Replace existing double-dispatch method by using individual TypeMapping classes

lajones:
... for each C# type.

This is the replacement for PR #8612. Removing the double-dispatch method we had and using separate TypeMapping classes for each C# type. Sets us up to better address issue #7434 in the future.


lajones:
Note: @smitpatel to make the `HasDefaultUnicode`/`HasDefaultSize` getters more robust am thinking of changing the constructors of all the TypeMappings to take `DefaultUnicode` and `DefaultSize`. Then I'll update `HasDefaultUnicode`/`HasDefaultSize` to just return whether `IsUnicode` matches `DefaultUnicode` etc. I haven't done it yet but that sounds better to me - what do you think?

src/EFCore.Relational/Storage/ByteTypeMapping.cs
```diff
@@ -0,0 +1,75 @@
...
+    ///     </para>
+    /// </summary>
+    public class ByteTypeMapping : RelationalTypeMapping<byte>
+    {
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="ByteTypeMapping" /> class.
+        /// </summary>
+        /// <param name="storeType"> The name of the database type. </param>
+        public ByteTypeMapping([NotNull] string storeType)
+            : this(storeType, System.Data.DbType.Byte, unicode: false, size: null)
```
lajones: @ajcvickers Re your comment about requiring `DbType` on all the constructors. The 6-arg constructor below appears on every TypeMapping and that can always be used to specify the `DbType` if needed. But I found that I needed to pass in the `DbType` only for any `DateTime` or `string` type mappings. For all the rest there was only 1 suitable choice of `DbType`. So for the "convenience" constructors like the above I decided to not have it. Let me know if you strongly disagree though - I can put it back the other way, but I was always passing the same thing so I thought this was better for the "convenience" constructors.

src/EFCore.Relational/Storage/RelationalTypeMapping.cs
```diff
@@ -247,164 +176,8 @@ public virtual string GenerateSqlLiteral([CanBeNull]object value)
...
-            
-            foreach (var @byte in value)
-            {
-                stringBuilder.Append(@byte.ToString("X2", CultureInfo.InvariantCulture));
-            }
-
-            stringBuilder.Append("'");
-            return stringBuilder.ToString();
+            //LAJLAJ What should we throw here?
+            return "ERROR: RelationalTypeMapping.GenerateSqlLiteral() received object of type " + value.GetType() +", with value " + value;
```
lajones: I'd like to throw here, but can I get opinions about the type of exception?

You should never reach here (we don't in any of our tests). If you do it's a sign that your provider has set up a TypeMapping badly.

We could just always return "NULL" or we could just do `ToString(CultureInfo.InvariantCulture)` on whatever is passed in, that could lead to data corruption: any non-null value should have been handled by GenerateNonNullSqlLiteral() in whatever your current TypeMapping is. So I'd prefer to throw to alert the providers (and their users).

src/EFCore.SqlServer/Storage/Internal/SqlServerTypeMapper.cs
```diff
@@ -17,89 +17,80 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
     /// </summary>
     public class SqlServerTypeMapper : RelationalTypeMapper
     {
-        private readonly SqlServerMaxLengthMapping _nvarcharmax
-            = new SqlServerMaxLengthMapping("nvarchar(max)", typeof(string), dbType: null, unicode: true, size: null);
+        private readonly SqlServerStringTypeMapping _nvarcharmax
```
lajones: I'd like to rename these to e.g. `_unboundedString`, `_keyString`, `_unboundedAnsiString` etc as in `TestRelationalTypeMapper`. Any objections?

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -568,7 +568,15 @@ public virtual Expression VisitFromSql(FromSqlExpression fromSqlExpression)
         }
 
         private RelationalTypeMapping GetTypeMapping(object value)
-            => _typeMapping ?? Dependencies.RelationalTypeMapper.GetMappingForValue(value);
+        {
+            if (_typeMapping != null
+                && _typeMapping.ClrType.IsAssignableFrom(value.GetType()))
```
ajcvickers: Is value known to be non-null?

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -568,7 +568,15 @@ public virtual Expression VisitFromSql(FromSqlExpression fromSqlExpression)
         }
 
         private RelationalTypeMapping GetTypeMapping(object value)
-            => _typeMapping ?? Dependencies.RelationalTypeMapper.GetMappingForValue(value);
+        {
+            if (_typeMapping != null
```
ajcvickers: Consider conditional operator and expression body.

src/EFCore.Relational/Storage/ByteTypeMapping.cs
ajcvickers: I don't think that there should be any place where we specify a DbType in our provider-agnostic code. I think that the DbType should be null by default for any type mapping. Then if SQL Server, or some other provider, needs to specify a DbType that should be done when it is constructed, in the code for that provider. In other words, I think we should maintain the previous approach which is that DbType is null unless some provider decides to explicitly set it.

src/EFCore.Relational/Storage/CharTypeMapping.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        ///     Generates the SQL representation of a literal value.
+        /// </summary>
+        /// <param name="value">The literal value.</param>
+        /// <returns>
+        ///     The generated string.
+        /// </returns>
+        protected override string GenerateNonNullSqlLiteral([NotNull]char value)
+            => string.Format(CultureInfo.InvariantCulture, "'{0}'", value);
```
ajcvickers: It seems that many of these subclasses do the same thing here--that is, use string.Format (or something that can be replaced with string.Format) to generate a value, but with a different format string. I wonder if we encapsulate this into a common class, then we will be able to remove a lot of these subclasses.

src/EFCore.Relational/Storage/RelationalTypeMapping.cs
ajcvickers: ToString(CultureInfo.InvariantCulture) seems like a reasonable default here. It seems fine to say that this is what this method does if there is no override. In other words, provider code is not required to override this unless it needs to do something different for the literal. Maybe this could be combined with a string format as a constructor parameter at this level such that many types will not need to override this?

src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    ///     <para>
+    ///         Represents the mapping between a specified .NET type and a database type.
+    ///     </para>
+    ///     <para>
+    ///         This type is typically used by database providers (and other extensions). It is generally
+    ///         not used in application code.
+    ///     </para>
+    /// </summary>
+    public abstract class RelationalTypeMapping<T> : RelationalTypeMapping
```
ajcvickers: Note sure the generic is needed. I can see how it is nice in some places, but on the fence as to whether it really adds enough value.

src/EFCore.Relational/Storage/ByteTypeMapping.cs
lajones: OK. I think they're useful. But I'll put them back the other way.

src/EFCore.SqlServer/Storage/Internal/SqlServerTypeMapper.cs
lajones: I'm going to go ahead and do this since I haven't seen any objections.

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
lajones: Good catch - we'd get an `NRE`. There is one case where it is not guaranteed do I'll fix to just return `_typeMapper` in that case. (It shouldn't matter which one we use as all TypeMappers know how to respond to a null value.)

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
lajones: Will do.

src/EFCore.Relational/Storage/CharTypeMapping.cs
lajones: Yes, we can do that. BTW do you know why the relational (i.e. non-provider specific) TypeMappings for the DateTime types return `TIMESTAMP '...'`? The addition of the `TIMESTAMP` literal seems odd.

ajcvickers:
Played with the code a bit this morning. I think this general approach is fine, but I think the amount of code could be reduced quite a bit by:
- Making the SQL literal string format a parameter on RelationalTypeMapping with a default of `{0}`
- Removing subclasses for anything where this is the default--includes all the integer types
- Creating subclasses for anything that needs a custom format and passing that down as an argument to the base class constructor. For example, the DateTime classes. This means that these classes don't have to override anything
  - Looks like we could have a class here for all quoted literals with a format of `'{0}'`. This could probably always do escaping as well, removing the need for a special string class here
- Then for really special things like byte arrays, a subclass can not use the format string and just override GenerateNonNullSqlLiteral.

src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
lajones: I liked it as it helped produce compile-time rather than runtime errors. I've now found all of the compile-time errors I need for below, but it might be useful in the future when we or provider-writers produce new Type Mappings for other types.

lajones:
@ajcvickers OK. The approach you put forward above is clear (and answers a lot of my other q's above). Want to make sure that we envisage that as a good approach when, later on, we include generating C# literals and/or any special operations/functions/LINQ translators for each type mapping as well - as mentioned as one of the original goals of #7434. I'd rather not drop all those classes now only to find we need them again for that work...

src/EFCore.Relational/Storage/CharTypeMapping.cs
divega: @lajones re `TIMESTAMP '...'` notation, it is a SQL standard thing. It can be used for literals in several common databases, e.g. MySQL and PostgreSQL, but doesn't work in SQL Server. The latter even assigns a completely different meaning to the `TIMESTAMP` type from the standard.

ajcvickers:
@bricelam Do you think we should keep all the subtypes now so that they are there for C# literal generation later?

src/EFCore.Relational/Storage/CharTypeMapping.cs
lajones: @divega Good to know. Thanks!

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
```diff
@@ -337,6 +337,9 @@ public MigrationsSqlGenerator([NotNull] MigrationsSqlGeneratorDependencies depen
                 builder
                     .Append(" AS ")
                     .Append(typeMapping.StoreType);
+
+                // reset typeMapping so that we can use it with GenerateSqlLiteral() below
+                typeMapping = Dependencies.TypeMapper.GetMapping(typeof(long));
```
smitpatel: While this is micro-optimization perhaps, but we should either move this outside of if or add better comment.
We are not resetting the type mapping but we are setting it as per operation.StartValue, which is long.

lajones: I originally had it outside, but then we regen the TypeMapping even if it was the correct one to begin with. I can change the comment to "Set the typeMapping for use with operation.StartValue below" if you like?

smitpatel: Hence "micro-optimization" 😉 
I am fine with just updating comment. So that at least it becomes clearer why we update it and why we use typeof(long).

lajones:
Re comment above about  the `HasDefaultUnicode`/`HasDefaultSize` getters, @smitpatel and I discussed and we think those getters are just in the wrong place altogether. Instead we think it would be better for the TypeMapping not to know anything about its "defaults" but instead have the `ScafdfoldingTypeMapper` compare the TypeMapping it receives to the default one for that C# type and see whether it needs to scaffold IsUnicode(), HasColumnType()... that way. But we'd like to discuss that approach and deal with it in a different PR - so leaving as is for now.

src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs
lajones: Yeah - but I'd have had the complaint the other way around if I hadn't done it. I thought the comment was clear but I'll rephrase as mentioned above.

src/EFCore.SqlServer/Storage/Internal/SqlServerTypeMapper.cs
```diff
@@ -159,15 +150,19 @@ public SqlServerTypeMapper([NotNull] RelationalTypeMapperDependencies dependenci
...
-                    { typeof(bool), _bit },
-                    { typeof(byte), _tinyint },
+                    { typeof(bool), _bool },
+                    { typeof(byte), _byte },
+                    { typeof(sbyte), _sbyte },
                     { typeof(double), _double },
                     { typeof(DateTimeOffset), _datetimeoffset },
-                    { typeof(char), _int },
-                    { typeof(short), _smallint },
+                    { typeof(char), new CharTypeMapping("char(1)") },
```
bricelam: Why did this change from `int`? I don't think we should change it, but if we do, this should be `nchar`.

bricelam:
> Do you think we should keep all the subtypes now so that they are there for C# literal generation later?

Yes. I think we're mostly settled on the design for code literals, and it will require the subtypes. Doing this now prevents us from breaking providers that need to override built-in type mappings.

src/EFCore.Relational/Storage/BoolTypeMapping.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+        }
+
+        /// <summary>
+        ///     Creates a copy of this mapping.
+        /// </summary>
+        /// <param name="storeType"> The name of the database type. </param>
+        /// <param name="size"> The size of data the property is configured to store, or null if no size is configured. </param>
+        /// <returns> The newly created mapping. </returns>
+        public override RelationalTypeMapping CreateCopy([NotNull] string storeType, int? size)
```
ajcvickers: Overriding methods should not have `[NotNull]` attributes. These only go where the method is defined--in the interface or base implementation of a method. R# will show them grayed out here.

src/EFCore.SqlServer/Storage/Internal/SqlServerTypeMapper.cs
ajcvickers: Agree with @bricelam This isn't a string mapping, it's a numeric mapping.

ajcvickers:
Based on @bricelam's comment, let's leave the sub-classes.

src/EFCore.SqlServer/Storage/Internal/SqlServerTypeMapper.cs
lajones: Agree on `nchar` vs `char` comment. I don't understand the rest though - why do we want to map to an integer when we have a SQL Server data type which maps directly to a single character?

src/EFCore.Relational/Storage/BoolTypeMapping.cs
lajones: OK. Will fix.

lajones:
@ajcvickers @bricelam I will leave the sub-classes.

src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
lajones: @ajcvickers I'm in the process of removing this. Let me know if my comment above changed your mind.

ajcvickers: I'm fine with leaving it or removing it. Up to you.

lajones: OK. I'll keep it then - since I found it useful. Thanks.

lajones:
@bricelam @ajcvickers Latest commit addresses all comments above except the outstanding one of to what underlying SQL Server type to map a `char` property.

src/EFCore.Relational/Storage/BoolTypeMapping.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public BoolTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: This constructor seems unnecessary because bool mappings can never be unicode or have size.

src/EFCore.Relational/Storage/ByteArrayTypeMapping.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="ByteArrayTypeMapping" /> class.
+        /// </summary>
+        /// <param name="storeType"> The name of the database type. </param>
+        /// <param name="dbType"> The <see cref="System.Data.DbType" /> to be used. </param>
+        /// <param name="unicode"> A value indicating whether the type should handle Unicode data or not. </param>
+        /// <param name="size"> The size of data the property is configured to store, or null if no size is configured. </param>
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public ByteArrayTypeMapping(
```
ajcvickers: Remove Unicode facets. Maybe collapse into one constructor with default values?

src/EFCore.Relational/Storage/ByteTypeMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="ByteTypeMapping" /> class.
+        /// </summary>
+        /// <param name="storeType"> The name of the database type. </param>
+        /// <param name="dbType"> The <see cref="System.Data.DbType" /> to be used. </param>
+        /// <param name="unicode"> A value indicating whether the type should handle Unicode data or not. </param>
+        /// <param name="size"> The size of data the property is configured to store, or null if no size is configured. </param>
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public ByteTypeMapping(
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/CharTypeMapping.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public CharTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Remove size params.

src/EFCore.Relational/Storage/DateTimeOffsetTypeMapping.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public DateTimeOffsetTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/DateTimeTypeMapping.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public DateTimeTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/DecimalTypeMapping.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public DecimalTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/DoubleTypeMapping.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public DoubleTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/FloatTypeMapping.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public FloatTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/GuidTypeMapping.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public GuidTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/IntTypeMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public IntTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/LongTypeMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public LongTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/RelationalTypeMapping.cs
```diff
@@ -185,6 +115,11 @@ public virtual RelationalTypeMapping CreateCopy([NotNull] string storeType, int?
         public virtual bool HasNonDefaultSize { get; }
 
         /// <summary>
+        ///     Gets the string format to be used to generate SQL literals of this type.
+        /// </summary>
+        public virtual string SqlLiteralFormatString { get; } = "{0}";
```
ajcvickers: Should this be protected?

src/EFCore.Relational/Storage/RelationalTypeMapping.cs
```diff
@@ -240,171 +175,21 @@ protected virtual void ConfigureParameter([NotNull] DbParameter parameter)
...
-        public virtual string GenerateSqlLiteral([CanBeNull]object value)
+        public virtual string GenerateSqlLiteral([CanBeNull] object value)
         {
-            if (value == null)
-            {
-                return "NULL";
-            }
-
-            return GenerateSqlLiteralValue((dynamic)value);
+            return value == null
```
ajcvickers: Expression body?

src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
```diff
@@ -0,0 +1,54 @@
...
+        /// </summary>
+        /// <param name="value">The literal value.</param>
+        /// <returns>
+        ///     The generated string.
+        /// </returns>
+        public override string GenerateSqlLiteral([CanBeNull]object value)
+        {
+            return value == null
+                ? base.GenerateSqlLiteral(value)
+                : GenerateNonNullSqlLiteral((T)value);
```
ajcvickers: Is this override needed?

src/EFCore.Relational/Storage/SByteTypeMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public SByteTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/ShortTypeMapping.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public ShortTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/StringTypeMapping.cs
```diff
@@ -0,0 +1,92 @@
...
+                hasNonDefaultSize: size != Size);
+
+        /// <summary>
+        ///     Generates the escaped SQL representation of a literal value.
+        /// </summary>
+        /// <param name="literal">The value to be escaped.</param>
+        /// <returns>
+        ///     The generated string.
+        /// </returns>
+        public virtual string EscapeSqlLiteral([NotNull]string literal)
```
ajcvickers: protected?

src/EFCore.Relational/Storage/StringTypeMapping.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        ///     Generates the SQL representation of a literal value.
+        /// </summary>
+        /// <param name="value">The literal value.</param>
+        /// <returns>
+        ///     The generated string.
+        /// </returns>
+        protected override string GenerateNonNullSqlLiteral(object value)
+        {
+            return $"'{EscapeSqlLiteral((string)value)}'";
```
ajcvickers: Expression body?

src/EFCore.Relational/Storage/TimeSpanTypeMapping.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public TimeSpanTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/UIntTypeMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public UIntTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/ULongTypeMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public ULongTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

src/EFCore.Relational/Storage/UShortTypeMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public UShortTypeMapping(
+            [NotNull] string storeType,
+            [CanBeNull] DbType? dbType,
+            bool unicode,
+            int? size,
+            bool hasNonDefaultUnicode = false,
+            bool hasNonDefaultSize = false)
+            : base(storeType, dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
```
ajcvickers: Not needed for same reasons.

smitpatel:
@ajcvickers - Do we really need `HasDefaultSize/Unicode` thing?

src/EFCore.SqlServer/Storage/Internal/SqlServerByteArrayTypeMapping.cs
```diff
@@ -12,33 +14,25 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
...
+        ///     Initializes a new instance of the <see cref="SqlServerByteArrayTypeMapping" /> class.
         /// </summary>
-        public SqlServerMaxLengthMapping(
+        /// <param name="storeType"> The name of the database type. </param>
+        /// <param name="dbType"> The <see cref="System.Data.DbType" /> to be used. </param>
+        /// <param name="unicode"> A value indicating whether the type should handle Unicode data or not. </param>
+        /// <param name="size"> The size of data the property is configured to store, or null if no size is configured. </param>
+        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
+        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
+        public SqlServerByteArrayTypeMapping(
```
ajcvickers: We generally don't have doc comments like this on internal classes, but doesn't matter much.

src/EFCore.Sqlite.Core/Storage/Internal/SqliteTypeMapper.cs
```diff
@@ -15,10 +15,15 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
...
-        private static readonly SqliteTypeMapping _text = new SqliteTypeMapping("TEXT", typeof(string));
+        private const string _integerTypeName = "INTEGER";
+        private const string _realTypeName = "REAL";
+        private const string _blobTypeName = "BLOB";
+        private const string _textTypeName = "TEXT";
+
+        private static readonly LongTypeMapping _integer = new LongTypeMapping(_integerTypeName, dbType: null);
+        private static readonly DoubleTypeMapping _real = new DoubleTypeMapping(_realTypeName, dbType: null);
+        private static readonly ByteArrayTypeMapping _blob = new ByteArrayTypeMapping(_blobTypeName, dbType: null);
+        private static readonly StringTypeMapping _text = new StringTypeMapping(_textTypeName, dbType: null);
```
ajcvickers: Can these now be moved inline? Or are the referenced elsewhere still?

test/EFCore.Relational.Tests/Storage/RelationalTypeMappingTest.cs
```diff
@@ -94,81 +93,154 @@ public void Can_create_string_parameter()
         [Fact]
         public virtual void GenerateSqlLiteral_returns_bool_literal_when_true()
         {
-            var literal = new RelationalTypeMapping("bool", typeof(bool)).GenerateSqlLiteral(true);
+            var literal = new BoolTypeMapping("bool", null).GenerateSqlLiteral(true);
```
ajcvickers: Is the null needed here? Does (or can) the parameter have a default?

test/EFCore.SqlServer.FunctionalTests/BuiltInDataTypesSqlServerTest.cs
```diff
@@ -573,7 +573,7 @@ public virtual void Can_insert_and_read_back_all_mapped_data_types_set_to_null()
 @p6:  (DbType = DateTime)
 @p7:  (DbType = DateTime)
 @p8:  (DbType = DateTime2)
-@p9:  (DbType = String)
```
ajcvickers: Do we know why this changed?

ajcvickers:
@smitpatel I'm fine with them going away if it's handled somewhere else.

lajones:
@smitpatel We agreed not to try to change the `HasNonDefaultUnicode`/`HasNonDefaultSize` issue in this PR, but to have another one to address it?

smitpatel:
@lajones - After discussing with @ajcvickers about it. Certainly we don't need to block this PR for changes around HasNonDefaultUnicode/Size part. But we want to get away from it overall since its scaffolding only info. Instead scaffolding (ScaffoldingTypeMapper precisely), which already has information about the current store type and what would be the model generated data type for the property, can diff both of them to figure out whether to scaffold those facets or not.
It is fine to do in separate PR. Also, if removing those 2 properties is breaking change, then for preview2 may be we can insert default static properties in type mapping, to have API settled on final format and remove those properties once scaffolding type mapper catches up.

lajones:
@smitpatel Yes. That's what we agreed yesterday. Do it separately after this is in.

src/EFCore.Sqlite.Core/Storage/Internal/SqliteTypeMapper.cs
lajones: They are referenced in a couple of places.

src/EFCore.Relational/Storage/RelationalTypeMapping.cs
lajones: Will do

src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
lajones: Will investigate - will remove if not needed.

src/EFCore.Relational/Storage/StringTypeMapping.cs
lajones: Will do.

src/EFCore.Relational/Storage/StringTypeMapping.cs
lajones: Will do.

src/EFCore.Relational/Storage/RelationalTypeMapping.cs
lajones: Will do.

test/EFCore.Relational.Tests/Storage/RelationalTypeMappingTest.cs
lajones: Updated

test/EFCore.SqlServer.FunctionalTests/BuiltInDataTypesSqlServerTest.cs
lajones: Because DbType.String is the default (for SQL Server) if you don't specify a parameter type. But for SQL Server we now specify that properties of type DateTimeOffset now correspond to parameters of type DbType.DateTimeOffset. Is this wrong? I know we're not specifying in Relational but this is SQL Server and I thought it was right to be specific.

lajones:
@ajcvickers All review comments actioned. New commit has all constructors updated as asked for above. This commit includes putting the SQL Server TypeMapping for `char` back in. The issue to take it out again when we figure out how to cope with that in the query pipeline is #8656.

lajones:
@ajcvickers Missed updates to remove unicode references in `ByteArrayTypeMapping`. Have fixed with 4th commit.

src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
ajcvickers: Can you explain a bit how this is needed?

test/EFCore.SqlServer.FunctionalTests/BuiltInDataTypesSqlServerTest.cs
ajcvickers: Okay.

lajones:
@ajcvickers Thanks. I'll set about rebasing this based on preview2 and will submit asap.

lajones:
Checked into rel/2.0.0-preview2 with commit ef9ce8ee4, and into dev with commit ebd225acf.

lajones:
Also commits 945874cd6 (in rel/2.0.0-preview2) and a6d1fa567 (in dev) to update a breaking-changes file.

PR #8637, opened 2017-05-30
Add EventId, LogLevel, and message generation for event payloads

ajcvickers:
Issue #7939

Also adds nominal types for all events other than those in the relational design assemblies. (Since these assemblies are currently being refactored.)

Each event payload now contains all the information needed to create and log the message for the event, including overridden ToString that creates the message.

### DIFF
--- a/src/EFCore/Extensions/Internal/CoreLoggerExtensions.cs
+++ b/src/EFCore/Extensions/Internal/CoreLoggerExtensions.cs
@@ -6,9 +6,6 @@ using System.Collections.Generic;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Diagnostics;
-using Microsoft.EntityFrameworkCore.Infrastructure;
-using Microsoft.EntityFrameworkCore.Metadata;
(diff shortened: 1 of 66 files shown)


src/EFCore/Extensions/Internal/CoreLoggerExtensions.cs
```diff
@@ -29,28 +26,35 @@ public static class CoreLoggerExtensions
...
-            [NotNull] Type contextType,
+            [NotNull] DbContext context,
             [NotNull] Exception exception)
         {
             var definition = CoreStrings.LogExceptionDuringSaveChanges;
 
             definition.Log(
                 diagnostics,
-                contextType, Environment.NewLine, exception,
+                context.GetType(), Environment.NewLine, exception,
```
anpete: Duped exception arg being passed

ajcvickers: Yep. I generally didn't change any of the logger messages that we generate. So where we had the exception in the message format string it is still there. I don't think there is a way to get it into the format string using the LoggerMessage facilities without including it as one of the args, meaning it gets duplicated here. But we could choose to remove it from the message if we wanted to.

PR #8641, opened 2017-05-30
Support named connection strings when IConfiguration is in the app service provider

ajcvickers:
Issue #7185

This change allows code like this:
```C#
.AddDbContext<MyContect>(
   b => b.UseSqlServer("name=ConnectionStrings:DefaultConnection"))

```

Assuming the IConfiguration is found in the same service provider as AddDbContext is used on, then this will find "ConnectionStrings:DefaultConnection" and use it as the connection string.

Should work for all relational providers.

### DIFF
--- a/src/EFCore.Relational/Storage/Internal/NamedConnectionStringResolver.cs
+++ b/src/EFCore.Relational/Storage/Internal/NamedConnectionStringResolver.cs
@@ -1,0 +1,74 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.EntityFrameworkCore.Storage.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class NamedConnectionStringResolver : INamedConnectionStringResolver
+    {
+        private const string DefaultSection = "ConnectionStrings:";
+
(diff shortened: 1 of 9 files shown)


src/EFCore.Relational/Storage/Internal/NamedConnectionStringResolver.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// </summary>
+        public virtual string ResolveConnectionString(string connectionString)
+        {
+            var connectionName = TryGetConnectionName(connectionString);
+
+            return connectionName != null
+                ? (_options.FindExtension<CoreOptionsExtension>()
+                       ?.ApplicationServiceProvider
+                       ?.GetService<IConfiguration>()
+                       ?[connectionName]
```
divega: By convention, `"ConnectionStrings"` is the well-known section to place connection strings in `IConfiguration`. There are already a few things aligning to it:
- ASP.NET Core templates
- The environment variable configuration provider that allows overriding connection strings in this section with connection strings set in the Azure portal 
- The `GetConnectionString()` extension method

I think we should align to this too. That means allowing for `"name=Foo"` instead of or in addition to `"name=ConnectionStrings:Foo"`.

bricelam:
Should it also look in App.config on .NET Framework (and other frameworks that support it)? (Or, is there bridge between IConfiguration and ConfigurationManager?)

ajcvickers:
@bricelam Yes, ideally it should do that. @divega sent a question to Immo about this. My intent was to support IConfiguration now and open a new issue to support app.config as well. I suspect it will require us to start cross-targeting again, but let's see.

src/EFCore.Relational/Storage/Internal/NamedConnectionStringResolver.cs
ajcvickers: @divega Updated to look in "ConnectionStrings:" in addition.

divega:
@ajcvickers @bricelam re System.Confuguration, I heard we can use https://www.nuget.org/packages/System.Configuration.ConfigurationManager/, but not sure about the details of what this does when the app runs on .NET Core.

PR #8646, opened 2017-05-31
Log a warning when a DIET instance is shared.

### LINKED ISSUE #8387: Consider throwing when a DIET instance is shared
At least log a warning as mutating it could lead to unexpected results if the user is not aware it's being shared.
Once we have readonly entities, they can be shared without any issues.

AndriySvyryd:
Fixes #8387

### DIFF
--- a/src/EFCore/Extensions/Internal/CoreLoggerExtensions.cs
+++ b/src/EFCore/Extensions/Internal/CoreLoggerExtensions.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.EntityFrameworkCore.Metadata;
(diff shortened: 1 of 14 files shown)


src/EFCore/Extensions/Internal/CoreLoggerExtensions.cs
```diff
@@ -59,6 +59,31 @@ private static string SaveChangesFailed(EventDefinitionBase definition, EventDat
...
+        {
+            var definition = CoreStrings.LogDuplicateDietInstance;
+
+            definition.Log(diagnostics, diet1, diet2);
+
+            if (diagnostics.DiagnosticSource.IsEnabled(definition.EventId.Name))
+            {
+                diagnostics.DiagnosticSource.Write(
+                    definition.EventId.Name,
+                    new
```
ajcvickers: This should be a nominal type, following the pattern of other methods here. Also, consider passing actual IEntityType instances instead of just strings.

Issue #7916, opened 2017-03-16
Investigate using IConfiguration in OnConfiguring


ajcvickers:
Current POR for 2.0 is that IConfiguration will be in the app service provider for ASP.NET Core applications. We should look at how easy it will be to get access to this from OnConfiguring. This is related to #7185

I don't expect we will, but we should also briefly discuss whether the DbContextOptionsBuilder code out of Startup and into OnConfiguring.

PR #8658, opened 2017-06-01
Throw if named connection cannot be found in configuration

ajcvickers:
Issue #7916

### DIFF
--- a/src/EFCore.Relational/Properties/RelationalStrings.resx
+++ b/src/EFCore.Relational/Properties/RelationalStrings.resx
@@ -135,6 +135,9 @@
   <data name="NoConnectionOrConnectionString" xml:space="preserve">
     <value>A relational store has been configured without specifying either the DbConnection or connection string to use.</value>
   </data>
+  <data name="NamedConnectionStringNotFound" xml:space="preserve">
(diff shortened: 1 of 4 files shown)


src/EFCore.Relational/Properties/RelationalStrings.resx
```diff
@@ -135,6 +135,9 @@
   <data name="NoConnectionOrConnectionString" xml:space="preserve">
     <value>A relational store has been configured without specifying either the DbConnection or connection string to use.</value>
   </data>
+  <data name="NamedConnectionStringNotFound" xml:space="preserve">
+    <value>A named connection string was used, but the name '{name}' was not found in the application's configuration. Note that named connection strings are only supported when using 'IConfiguration' and a service provider, such as in a typical ASP.NET Core application.</value>
```
divega: Will the part about "and a service provider" stop being accurate when you address #8659?

ajcvickers: I figured we can re-write the message as it evolves. I was expecting to still say this, but add an "or" clause every time we add a new way of doing it. :-) Might be better to have a fwlink and a doc page?

divega: @ajcvickers here it is: https://go.microsoft.com/fwlink/?linkid=850912.

PR #8661, opened 2017-06-01
Design: Clean up design-time services

### LINKED ISSUE #8623: MigrationsScaffolder is a public type registered in D.I. with no contract or dependency objects
Typically all our public services:
- Are defined by a contract interface
- Use dependency object pattern so that we can add new dependencies without breaking inheritors

MigrationsScaffolder doesn't do these things. It's only added to the design-time container so maybe it doesn't matter, but it seems like it may still have the same issues as runtime services, so creating this issue to discuss.

bricelam:
Fixes #8623


src/EFCore.Design/Migrations/Design/CSharpSnapshotGeneratorDependencies.cs
```diff
@@ -0,0 +1,64 @@
...
+        ///     </para>
+        ///     <para>
+        ///         This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///         directly from your code. This API may change or be removed in future releases.
+        ///     </para>
+        /// </summary>
+        /// <param name="csharpHelper"> The C# helper. </param>
+        public CSharpSnapshotGeneratorDependencies([NotNull] ICSharpHelper csharpHelper)
+        {
+            CSharpHelper = csharpHelper;
```
ajcvickers: Check.NotNull

PR #8676, opened 2017-06-01
Remove Relational.Design package

smitpatel:
This removes the final piece.


src/EFCore.Design/Diagnostics/DesignEventId.cs
```diff
@@ -38,7 +38,18 @@ private enum Id
             MigrationWriting,
             SnapshotNameReusing,
             DestructiveOperation,
-            ForeignMigrations
+            ForeignMigrations,
+
+            // Scaffolding warning events
+            MissingSchemaWarning = CoreEventId.CoreDesignBaseId + 100,
```
bricelam: File an issue to remove all logging from Design. `IOperationReporter` should be used instead.

smitpatel: Will do. 
Sorry these commit got bit messed up since I am in process of cherry-picking commit since rebase was disaster.

bricelam: No worries; I just looked at the commit which I hand't reviewed yet.

Issue #8656, opened 2017-05-31
Remove SqlServer TypeMapping for CLR type char


lajones:
An EntityType with property `char` is not supported by the SQL Server provider. However `SqlServerTypeMapper` does have a mapping in `_clrTypeMappings` for `char` (mapping it to a SQL Server `int` column).

The reason for this is to support generation of single character literals in the Query Pipeline e.g. as arguments to functions, or `LIKE` can take an `ESCAPE` argument which is a single character - see [this test](https://github.com/aspnet/EntityFramework/blob/dev/test/EFCore.SqlServer.FunctionalTests/DbFunctionsSqlServerTest.cs#L40). Query checks that it can interpret each argument (in this case a single character) - if it fails to find a type mapping for that type then it assumes it cannot generate literals for that type.

We should remove the entry in `_clrTypeMappings` for `SqlServerTypeMapper` for `char` so that we will generate the correct error if someone tries to create a property of that type. But we need to provide an alternate means to translate the literal above. We discussed:
1. Just changing how we deal with `LIKE` to enforce that the `ESCAPE` argument is a string, or
2. Changing the way Query treats arguments in general to specifically allow it to try a `string` TypeMapping if a `char` one does not exist.
3. Updating the TypeMapper to allow a set of types which it can interpret for generation of literals even if it does not allow that type as a property type.

divega:
Additional data points:

- `GetChar` is not supported in either SQL Server or Oracle's ADO.NET provider
- Currently it is necessary for the provider to know how to generate a parameter, and not just a literal for the `escapeCharacter` argument. 
- Another potential approach would be to create some sort of fallback mapping mechanism for types like `char` and `byte`: 

- if `char` is not explicitly supported by the provider, it could fall back to a 'string' of size 1
- if `byte` is not explicitly supported by the provider, it could fall back to a `byte[]` of size 1

This feels like it is slowly turning into a short-range version of the type conversions feature, but maybe that is ok.

lajones:
Note: this came out of the work for #8631.

divega:
cc @anpete

lajones:
Note: decided for preview2 to just to update the way we deal with `LIKE` to pass in the single character as a string. And remove the `char` SQL Server TypeMapping.

ajcvickers:
@lajones Can this issue be closed?

lajones:
[...]

PR #8678, opened 2017-06-01
Update SqlServerTypeMapper to no longer allow char mappings

lajones:
Which requires we also update treatment of `LIKE` to treat the escape character as a string.

This addresses issue #8656 for preview2. We may want to revisit our approach after that as we have not addressed the wider issue e.g. being able to construct a `DbParameter` for a character arg.

### DIFF
--- a/src/EFCore/DbFunctionsExtensions.cs
+++ b/src/EFCore/DbFunctionsExtensions.cs
@@ -30,7 +30,7 @@ namespace Microsoft.EntityFrameworkCore
             [CanBeNull] this DbFunctions _,
             [CanBeNull] string matchExpression, 
             [CanBeNull] string pattern)
-            => LikeCore(matchExpression, pattern, escapeCharacter: null);
+            => LikeCore(matchExpression, pattern, escapeString: null);
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


src/EFCore/DbFunctionsExtensions.cs
```diff
@@ -30,7 +30,7 @@ public static class DbFunctionsExtensions
             [CanBeNull] this DbFunctions _,
             [CanBeNull] string matchExpression, 
             [CanBeNull] string pattern)
-            => LikeCore(matchExpression, pattern, escapeCharacter: null);
+            => LikeCore(matchExpression, pattern, escapeString: null);
```
divega: I am inclined to keep the name `escapeCharacter` even if we change the type.

divega: cc @ajcvickers @anpete

anpete: Yep, keep the name.

lajones: OK. Will do.

lajones:
Checked in to rel/2.0.0-preview2 with commit 3b3af4dc0, and to dev with commit 83395e5f6.

Issue #8584, opened 2017-05-25
Query: reuse include pipeline for queries projecting uncomposed collection navigations


maumar:
```
ctx.Customers.Select(c => c.Orders)
```

currently produces N+1 queries, however, similar query:

```
ctx.Customers.Include(c => c.Orders)
```

is much better, producing only 2 queries.

We should be able to leverage the include pipeline for the first query

maumar:
related to #8229

maumar:
fixed in 0075cb7c831bb3618bee4a84c9bfa86a499ddc6c

PR #8681, opened 2017-06-02
Fix to #8584 - Query: reuse include pipeline for queries projecting uncomposed collection navigations

maumar:
Currently when projecting collection navigation we produce N+1 queries. This change leverages include pipeline which produces 2 queries instead and joins the result on the client.

We only do this for collection navigation that are not composed on - otherwise, the query that happens after collection navigation would be performed on the client, so it's not clear that would be preferable to N+1.
Also as a side effect, we always materialize root entity (or all intermediate entities in case of multi-level navigation). This is a side effect of reusing Include pipeline, which always expects (and depends on) root element being materialized.

We leverage the include pipeline by performing a simple QM rewrite:

```
from [e] in entities
select new { [e].Id, [e].Reference.Collection }
```

will be translated to:

```
from [e] in entities
select new { [e].Id, _ProjectCollectionNavigation(
  [e], 
  prm => SuppressNavigationRewrite(prm?.Reference?.Collection) ?? new ICollection<Entity>()) }
```
additionally, we generate include result operator for [e].Reference.Collection.

When include pipeline processes the updated QM, it will apply the include on first argument to _ProjectCollectionNavigation.
Implementation of _ProjectCollectionNavigation simply invokes the Func specified as it's second argument (which accesses requested collection navigation), given the root entity (which now has include applied).

### DIFF
--- a/src/EFCore.Specification.Tests/QueryNavigationsTestBase.cs
+++ b/src/EFCore.Specification.Tests/QueryNavigationsTestBase.cs
@@ -429,7 +429,8 @@ namespace Microsoft.EntityFrameworkCore
                         {
                             Assert.Equal(pair.l2oItem.Orders, pair.efItem.Orders);
                         }
-                    });
+                    },
+                entryCount: 34);
         }
 
         [ConditionalFact]
@@ -448,7 +449,8 @@ namespace Microsoft.EntityFrameworkCore
                         {
                             Assert.Equal(pair.l2oItem.Orders, pair.efItem.Orders);
                         }
-                    });
+                    },
+                entryCount: 7);
         }
 
         [ConditionalFact]
--- a/src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
+++ b/src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
@@ -84,7 +84,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual object Create(IEnumerable<object> values)
+        public virtual object Create()
         {
             if (_createCollection == null)
             {
@@ -92,13 +92,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
                     _propertyName, typeof(TEntity).ShortDisplayName(), typeof(TCollection).ShortDisplayName()));
             }
 
-            var collection = (ICollection<TElement>)_createCollection();
-            foreach (TElement value in values)
-            {
-                collection.Add(value);
-            }
-
-            return collection;
+            return _createCollection();
         }
 
         /// <summary>
--- a/src/EFCore/Query/EntityQueryModelVisitor.cs
+++ b/src/EFCore/Query/EntityQueryModelVisitor.cs
@@ -12,13 +12,16 @@ using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors;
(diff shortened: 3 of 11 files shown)


src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
```diff
@@ -84,21 +84,15 @@ public virtual void AddRange(object instance, IEnumerable<object> values)
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual object Create(IEnumerable<object> values)
+        public virtual object Create()
```
maumar: unrelated change - it's needed to make MaterializeCollectionNavigation async. I might not get around to doing this for 2.0, but this API change helps doing it post 2.0

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -230,6 +233,134 @@ protected virtual void InterceptExceptions()
...
+                    && properties[properties.Count - 1] is INavigation lastNavigation 
+                    && lastNavigation.IsCollection())
+                {
+                    CollectionNavigationIncludeResultOperators.Add(
+                        new IncludeResultOperator(properties.Select(p => p.Name), new QuerySourceReferenceExpression(querySource)));
+
+                    var navigationMethodInfo = _navigationMethodInfo.MakeGenericMethod(querySource.ItemType, expression.Type);
+                    var parameter = Expression.Parameter(querySource.ItemType, "prm");
+                    var body = new CollectionNavigationIncludeReplacingExpressionVisitor(querySource, parameter).Visit(expression);
+                    var emptyCollection = lastNavigation.GetCollectionAccessor().Create();
```
maumar: we could capture lastNavigation here instead and gen the GetCollectionAccessor().Create() code. I can't come up with easy way to avoid referencing this constant altogether. Thoughts? Do we even care?

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -230,6 +233,134 @@ protected virtual void InterceptExceptions()
...
+
+                protected override Expression VisitQuerySourceReference(QuerySourceReferenceExpression expression)
+                    => expression.ReferencedQuerySource == _querySource
+                        ? _replacementExpression
+                        : expression;
+
+                protected override Expression VisitMember(MemberExpression memberExpression)
+                {
+                    var newExpression = Visit(memberExpression.Expression);
+                    var newMemberExpression = memberExpression.Update(newExpression);
```
maumar: We need to convert all member access and EF.Property into member access null conditionals. . This implementation is not ideal, as we can end up with NullConditionalExpression in the AccessOperation. We could do 2 separate visitors, first converting EF.Property into member access, and the second one converting member access into null conditionals. Any ideas how to do it in one pass (convert all EF.Property everywhere into member access, but convert member access to null conditional recursively only for the caller. Or again, do we even care?

src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
maumar: just realized this is internal API - I can revert this. No harm making this change tho, lmk

src/EFCore/Query/Expressions/Internal/SuppressNavigationRewriteExpression.cs
```diff
@@ -0,0 +1,79 @@
...
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Query.Expressions.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class SuppressNavigationRewriteExpression : Expression
```
maumar: alternatively, we could make ProjectCollectionNavigation expression, to capture this

src/EFCore.Specification.Tests/QueryNavigationsTestBase.cs
```diff
@@ -448,7 +449,8 @@ public virtual void Select_collection_navigation_multi_part()
                         {
                             Assert.Equal(pair.l2oItem.Orders, pair.efItem.Orders);
                         }
-                    });
+                    },
+                entryCount: 7);
```
maumar: ideally, we should be tracking 6 items, but customer is also tracked as a side effect of using include pipeline. Will add issue for this

src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
anpete: I would revert for now.

maumar:
Closing this, opened another one for preview2

smitpatel:
@maumar - You could have changed base on this PR 😕

PR #8684, opened 2017-06-02
Use verbs in convention names and the property names in ConventionSet.

### LINKED ISSUE #8672: API Review: ConventionSet changes
Use verbs in convention names and the property names in ConventionSet.
Change "Set" to "Changed".

AndriySvyryd:
Fixes #8672

### DIFF
--- a/src/EFCore/Metadata/Conventions/ConventionSet.cs
+++ b/src/EFCore/Metadata/Conventions/ConventionSet.cs
@@ -14,7 +14,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions
         /// <summary>
         ///     Conventions to run when an entity type is added to the model.
         /// </summary>
-        public virtual IList<IEntityTypeConvention> EntityTypeAddedConventions { get; } = new List<IEntityTypeConvention>();
+        public virtual IList<IEntityTypeAddedConvention> EntityTypeAddedConventions { get; } = new List<IEntityTypeAddedConvention>();
 
         /// <summary>
         ///     Conventions to run when an entity type is ignored.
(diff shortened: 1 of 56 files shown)


src/EFCore/Metadata/Conventions/ConventionSet.cs
```diff
@@ -29,18 +29,18 @@ public class ConventionSet
         /// <summary>
         ///     Conventions to run when the base entity type is set or removed.
```
smitpatel: Optional: Update the doc comments.

AndriySvyryd:
@smitpatel Did you finish reviewing the changes?

src/EFCore/Metadata/Conventions/ConventionSet.cs
```diff
@@ -60,12 +60,12 @@ public class ConventionSet
         /// <summary>
         ///     Conventions to run when a primary key is configured.
```
smitpatel: changed

src/EFCore/Metadata/Conventions/ConventionSet.cs
```diff
@@ -75,32 +75,32 @@ public class ConventionSet
         /// <summary>
         ///     Conventions to run when the uniqueness of an index is changed.
         /// </summary>
-        public virtual IList<IIndexUniquenessConvention> IndexUniquenessConventions { get; } = new List<IIndexUniquenessConvention>();
+        public virtual IList<IIndexUniquenessChangedConvention> IndexUniquenessChangedConventions { get; } = new List<IIndexUniquenessChangedConvention>();
 
         /// <summary>
         ///     Conventions to run when an annotation is set or removed on an index.
```
smitpatel: do we do processing when annotation value modifies? In that case this should be changed otherwise set/removed is also fine.

src/EFCore/Metadata/Conventions/ConventionSet.cs
```diff
@@ -75,32 +75,32 @@ public class ConventionSet
...
         /// </summary>
-        public virtual IList<IIndexAnnotationSetConvention> IndexAnnotationSetConventions { get; } = new List<IIndexAnnotationSetConvention>();
+        public virtual IList<IIndexAnnotationChangedConvention> IndexAnnotationChangedConventions { get; } = new List<IIndexAnnotationChangedConvention>();
 
         /// <summary>
         ///     Conventions to run when the principal end of a relationship is configured.
```
smitpatel: changed?

AndriySvyryd: This one is special, because it also runs when the ConfigurationSource for principal end is changed, we should probably split it into two

smitpatel: I was also wondering the splitting part of ModelBuilt & ModelInitialized convention. It may be good to align them but if they are similar kind then probably we can keep differently named same interface perhaps.

AndriySvyryd: They need to be separate so that both could be implemented in the same class

PR #8692, opened 2017-06-02
Remove obsolete remarks about forcing shadowness

ExplodingCabbage:
I believe that @ajcvickers made these remarks obsolete/nonsensical when he removed the "shadow" parameter in https://github.com/aspnet/EntityFramework/commit/847a2668125beba73878cef683922be6b2e81cad, but that he missed these remarks in his adjustment of the docs.


ajcvickers:
Thanks, Mr Cabbage. 🍰

PR #8700, opened 2017-06-02
Design: Fix namespace-directory mismatches

bricelam:


src/ef/WrappedException.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.EntityFrameworkCore.Tools
 {
-    public class WrappedException : Exception
+    internal class WrappedException : Exception
```
ajcvickers: Why?

bricelam: Everything else in the assembly is. We did it for two reasons:
* It's an app, not a library
* You used to see the public types when the package was installed into a .NET Core project. (This doesn't happen any more.)

PR #8701, opened 2017-06-02
Bring back IDatabaseModelFactory.Create which takes DbConnection for …

smitpatel:
…Clean

This fixes build on external SqlServer

Manually tested on external SqlSever 2016


src/EFCore.Relational/Scaffolding/IDatabaseModelFactory.cs
```diff
@@ -9,5 +10,6 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding
     public interface IDatabaseModelFactory
     {
         DatabaseModel Create([NotNull] string connectionString, [NotNull] TableSelectionSet tableSelectionSet);
```
bricelam: Do we have a provider service that can go from connectoinString to DbConnection? If so, we should provide a base implementation that uses it so the provider just has to implement the one that takes a DbConnection.

PR #8704, opened 2017-06-02
API Review: Review/rename query EventData classes

### LINKED ISSUE #8673: API Review: Review/rename query EventData classes
Also, look at whether the appropriate object is passed down to the payload, rather than just string.

Check these types, and any others that are query-related:
- BinaryExpressionEventData
- IncludeEventData
- NavigationPathEventData
- QueryModelExpressionEventData

anpete:
- Also fixes could-be-static method warnings in Core.

Fix #8673


src/EFCore/Diagnostics/IncludeEventData.cs
```diff
@@ -18,19 +19,19 @@ public class IncludeEventData : EventData
         /// </summary>
         /// <param name="eventDefinition"> The event definition. </param>
         /// <param name="messageGenerator"> A delegate that generates a log message for this event. </param>
-        /// <param name="includeSpecification"> The Include specification. </param>
+        /// <param name="includeResultOperator"> The Include specification. </param>
```
smitpatel: doc comment

PR #8708, opened 2017-06-02
Remove HasDefaultUnicode/HasDefaultSize properties from TypeMapping

### LINKED ISSUE #8668: Consider Removing HasNonDefaultUnicode and HasNonDefaultSize
cc @smitpatel 

See [RelationalTypeMapping](https://github.com/aspnet/EntityFramework/blob/rel/2.0.0-preview2/src/EFCore.Relational/Storage/RelationalTypeMapping.cs#L55). This class contains getters and constructors that take the arguments `HasNonDefaultUnicode` and `HasNonDefaultSize`.

Consider removing those in favor of the `ScaffoldingTypeMapper` comparing a TypeMapping to the default TypeMapping for a given CLR type to decide whether to scaffold `HasColumnType()`, `IsUnicode()`, `MaxLength()` etc. API calls.

smitpatel:
Resolves #8668 

Bring backs `_maxSpecificSize`
`ScaffoldingTypeMapper` gets default mapping and do comparison with that to generate facets.

R# ran redundant parenthesis over whole project.

### DIFF
--- a/src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
+++ b/src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
@@ -24,16 +24,12 @@ namespace Microsoft.EntityFrameworkCore.Storage
         /// <param name="dbType"> The <see cref="System.Data.DbType" /> to be used. </param>
         /// <param name="unicode"> A value indicating whether the type should handle Unicode data or not. </param>
         /// <param name="size"> The size of data the property is configured to store, or null if no size is configured. </param>
-        /// <param name="hasNonDefaultUnicode"> A value indicating whether the Unicode setting has been manually configured to a non-default value. </param>
-        /// <param name="hasNonDefaultSize"> A value indicating whether the size setting has been manually configured to a non-default value. </param>
         public RelationalTypeMapping(
             [NotNull] string storeType,
             [CanBeNull] DbType? dbType,
             bool unicode,
-            int? size,
-            bool hasNonDefaultUnicode = false,
-            bool hasNonDefaultSize = false)
-            : base(storeType, typeof(T), dbType, unicode, size, hasNonDefaultUnicode, hasNonDefaultSize)
+            int? size)
+            : base(storeType, typeof(T), dbType, unicode, size)
         {
         }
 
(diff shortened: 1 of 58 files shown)


smitpatel:
Fixing few things **WIP**

src/EFCore.Relational/Storage/RelationalTypeMapping`.cs
```diff
@@ -44,10 +40,10 @@ public abstract class RelationalTypeMapping<T> : RelationalTypeMapping
         /// <returns>
         ///     The generated string.
         /// </returns>
-        public override string GenerateSqlLiteral([CanBeNull]object value)
+        public override string GenerateSqlLiteral(object value)
         {
             return value == null
-                ? base.GenerateSqlLiteral(value)
+                ? "NULL"
                 : GenerateNonNullSqlLiteral((T)value);
```
lajones: We want the word "NULL" in only 1 place (the base class) - that's why we have this setup.

smitpatel: At present it is already at multiple places. Its in RelationalModelValidator, Query SQL gen, NullMapping etc.
There does not seem to be any value is making 1 place if we are not doing across the board.

lajones: Sorry - I disagree. I think we _should_ do this across the board - so no need to make it worse here. Though actually I think the best solution would be to have a constant somewhere set equal to "NULL" and just refer to that everywhere. If you raise a bug to do that then this is OK.

ajcvickers: I still don't understand why we even have this override. Doesn't it do the same thing as the base class?

smitpatel: Converts value to type `T` rather than object. I am uncertain if it is required.

lajones: It is required. I did some investigation today and It turned out the problem with not having the cast to `T` was to do with Enums. If you just use the base class implementation of `GenerateNonNullSqlLiteral()` then Enum literals come out as their string representation instead of their integer representation. This way it automatically casts them to the correct `int`/`long`/... value.

smitpatel: Kept the override for now.

PR #8709, opened 2017-06-02
Fix to #8584 - Query: reuse include pipeline for queries projecting uncomposed collection navigations

maumar:
Currently when projecting collection navigation we produce N+1 queries. This change leverages include pipeline which produces 2 queries instead and joins the result on the client.

We only do this for collection navigation that are not composed on - otherwise, the query that happens after collection navigation would be performed on the client, so it's not clear that would be preferable to N+1.
Also as a side effect, we always materialize root entity (or all intermediate entities in case of multi-level navigation). This is a side effect of reusing Include pipeline, which always expects (and depends on) root element being materialized.

We leverage the include pipeline by performing a simple QM rewrite:

```
from [e] in entities
select new { [e].Id, [e].Reference.Collection }
```

will be translated to:

```
from [e] in entities
select new { [e].Id, _ProjectCollectionNavigation([e], prm => prm?.Reference?.Collection ?? new ICollection<Entity>()) }
```
additionally, we generate include result operator for [e].Reference.Collection.

When include pipeline processes the updated QM, it will apply the include on first argument to _ProjectCollectionNavigation.
Implementation of _ProjectCollectionNavigation simply invokes the Func specified as it's second argument (which accesses requested collection navigation), given the root entity (which now has include applied)

### DIFF
--- a/src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
+++ b/src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
@@ -84,7 +84,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual object Create(IEnumerable<object> values)
+        public virtual object Create()
         {
             if (_createCollection == null)
             {
@@ -92,7 +92,16 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
                     _propertyName, typeof(TEntity).ShortDisplayName(), typeof(TCollection).ShortDisplayName()));
             }
 
-            var collection = (ICollection<TElement>)_createCollection();
+            return _createCollection();
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual object Create(IEnumerable<object> values)
+        {
+            var collection = (ICollection<TElement>)Create();
             foreach (TElement value in values)
             {
                 collection.Add(value);
--- a/src/EFCore/Query/EntityQueryModelVisitor.cs
+++ b/src/EFCore/Query/EntityQueryModelVisitor.cs
@@ -12,13 +12,14 @@ using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Diagnostics;
-using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
 using Microsoft.EntityFrameworkCore.Query.Internal;
+using Microsoft.EntityFrameworkCore.Query.ResultOperators;
 using Microsoft.EntityFrameworkCore.Query.ResultOperators.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 using Remotion.Linq;
@@ -230,6 +231,19 @@ namespace Microsoft.EntityFrameworkCore.Query
                     QueryContextParameter);
 
         /// <summary>
+        /// Rewrites collection navigation projections so that they can be handled by the Include pipeline.
+        /// </summary>
+        /// <param name="queryModel"> The query. </param>
+        protected virtual void RewriteProjectedCollectionNavigationsToIncludes([NotNull] QueryModel queryModel)
+        {
+            Check.NotNull(queryModel, nameof(queryModel));
+
+            var collectionNavigationIncludeRewriter = new CollectionNavigationIncludeExpressionRewriter(this);
+            queryModel.SelectClause.Selector = collectionNavigationIncludeRewriter.Visit(queryModel.SelectClause.Selector);
+            _queryCompilationContext.AddAnnotations(collectionNavigationIncludeRewriter.CollectionNavigationIncludeResultOperators);
+        }
+
+        /// <summary>
         ///     Populates <see cref="Query.QueryCompilationContext.QueryAnnotations" /> based on annotations found in the query.
         /// </summary>
         /// <param name="queryModel"> The query. </param>
@@ -259,6 +273,8 @@ namespace Microsoft.EntityFrameworkCore.Query
             new NondeterministicResultCheckingVisitor(QueryCompilationContext.Logger).VisitQueryModel(queryModel);
 
             // Rewrite includes/navigations
+
+            RewriteProjectedCollectionNavigationsToIncludes(queryModel);
 
             var includeCompiler = new IncludeCompiler(QueryCompilationContext, _querySourceTracingExpressionVisitorFactory);
 
--- a/src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
@@ -1,0 +1,144 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Reflection;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
+using Microsoft.EntityFrameworkCore.Query.ResultOperators;
+using Microsoft.EntityFrameworkCore.Query.ResultOperators.Internal;
+using Remotion.Linq.Clauses;
+using Remotion.Linq.Clauses.Expressions;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class CollectionNavigationIncludeExpressionRewriter : ExpressionVisitorBase
+    {
+        private readonly EntityQueryModelVisitor _queryModelVisitor;
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public CollectionNavigationIncludeExpressionRewriter(
+            [NotNull] EntityQueryModelVisitor queryModelVisitor)
+        {
+            _queryModelVisitor = queryModelVisitor;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual List<IQueryAnnotation> CollectionNavigationIncludeResultOperators { get; }
+            = new List<IQueryAnnotation>();
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitMember(MemberExpression node)
+            => _queryModelVisitor.BindNavigationPathPropertyExpression(
+                node,
+                (ps, qs) => RewritePropertyAccess(node, ps, qs)) ?? base.VisitMember(node);
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitMethodCall(MethodCallExpression node)
+            => _queryModelVisitor.BindNavigationPathPropertyExpression(
+                node,
+                (ps, qs) => RewritePropertyAccess(node, ps, qs)) ?? base.VisitMethodCall(node);
+
+        private Expression RewritePropertyAccess(
+            Expression expression,
+            IReadOnlyList<IPropertyBase> properties,
+            IQuerySource querySource)
+        {
+            if (querySource != null
+                && properties.Count > 0
+                && properties.All(p => p is INavigation)
+                && properties[properties.Count - 1] is INavigation lastNavigation
+                && lastNavigation.IsCollection())
+            {
+                // include doesn't handle navigations on derived class, so we can't leverage include pipeline for those cases
+                var expectedCallerType = querySource.ItemType;
+                foreach (var property in properties)
+                {
+                    if (expectedCallerType != property.DeclaringType.ClrType)
+                    {
+                        return expression;
+                    }
+
+                    expectedCallerType = property.ClrType;
+                }
+
+                var qsre = new QuerySourceReferenceExpression(querySource);
+
+                CollectionNavigationIncludeResultOperators.Add(
+                    new IncludeResultOperator(properties.Cast<INavigation>().ToArray(), qsre));
+
+                var parameter = Expression.Parameter(querySource.ItemType, querySource.ItemName);
+                var accessorBody = BuildCollectionAccessorExpression(parameter, properties);
+                var emptyCollection = lastNavigation.GetCollectionAccessor().Create();
+
+                return Expression.Call(
+                    ProjectCollectionNavigationMethodInfo
+                        .MakeGenericMethod(querySource.ItemType, expression.Type),
+                    qsre,
+                    Expression.Lambda(
+                        Expression.Coalesce(
+                            accessorBody,
+                            Expression.Constant(emptyCollection, accessorBody.Type)),
+                        parameter));
+            }
+
+            return expression;
+        }
+
+        private Expression BuildCollectionAccessorExpression(
+            ParameterExpression parameter, 
+            IEnumerable<IPropertyBase> navigations)
+        {
+            Expression result = parameter;
+            Expression memberExpression = parameter;
+            foreach (var navigation in navigations)
+            {
+                memberExpression = memberExpression.MakeMemberAccess(navigation.PropertyInfo);
+                result = new NullConditionalExpression(result, memberExpression);
+            }
+
+            return result;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitSubQuery(SubQueryExpression subQueryExpression)
+            => subQueryExpression;
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static readonly MethodInfo ProjectCollectionNavigationMethodInfo
+            = typeof(CollectionNavigationIncludeExpressionRewriter).GetTypeInfo()
+                .GetDeclaredMethod(nameof(_ProjectCollectionNavigation));
+
+        // ReSharper disable once InconsistentNaming
+        private static TResult _ProjectCollectionNavigation<TEntity, TResult>(TEntity entity, Func<TEntity, TResult> accessor)
+            => accessor(entity);
+    }
+}
--- a/src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationSubqueryInjector.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationSubqueryInjector.cs
@@ -78,6 +78,16 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         {
             Check.NotNull(methodCallExpression, nameof(methodCallExpression));
 
+            if (methodCallExpression.Method.MethodIsClosedFormOf(
+                CollectionNavigationIncludeExpressionRewriter.ProjectCollectionNavigationMethodInfo))
+            {
+                var newArgument = Visit(methodCallExpression.Arguments[0]);
+
+                return newArgument != methodCallExpression.Arguments[0]
+                    ? methodCallExpression.Update(methodCallExpression.Object, new[] { newArgument, methodCallExpression.Arguments[1] })
+                    : methodCallExpression;
+            }
+
             if (!methodCallExpression.Method.IsEFPropertyMethod())
             {
                 _shouldInject = true;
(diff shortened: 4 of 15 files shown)


maumar:
@anpete make some changes and created based on preview2 - please take a look

src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
```diff
@@ -84,21 +84,15 @@ public virtual void AddRange(object instance, IEnumerable<object> values)
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual object Create(IEnumerable<object> values)
+        public virtual object Create()
```
maumar: ended up keeping this change - we need to produce this call in the ProjectCollectionNavigation, and this change makes the call bit easier (no need to specify the unnecessary argument).

smitpatel:
>     public class ClrICollectionAccessor<TEntity, TCollection, TElement> : IClrCollectionAccessor

name of the class is weird

---
Refers to: src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs:15 in daca3b8. [](commit_id = daca3b8b3088d0ac8b36e77a43889fbbc3fc1595, deletion_comment = True)

src/EFCore/Metadata/Internal/ClrICollectionAccessor.cs
smitpatel: how about overload?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class CollectionNavigationIncludeExpressionRewriter : ExpressionVisitorBase
+    {
+        private readonly EntityQueryModelVisitor _queryModelVisitor;
+
+        private static readonly MethodInfo _getCollectionAccessorMethodInfo
+            = typeof(Metadata.Internal.NavigationExtensions).GetTypeInfo()
```
smitpatel: include in using

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationSubqueryInjector.cs
```diff
@@ -122,11 +132,17 @@ private static Expression InjectSubquery(Expression expression, INavigation coll
...
-        private static ICollection<TEntity> MaterializeCollectionNavigation<TEntity>(INavigation navigation, IEnumerable<object> elements)
+        private static ICollection<TEntity> MaterializeCollectionNavigation<TEntity>(
+            INavigation navigation, 
+            IEnumerable<TEntity> elements)
         {
-            var collection = navigation.GetCollectionAccessor().Create(elements);
+            var collection = (ICollection<TEntity>)navigation.GetCollectionAccessor().Create();
+            foreach (var element in elements)
+            {
+                collection.Add(element);
```
smitpatel: AddRange maybe

anpete:
What happens when I project: c.Product and c.Product.Orders?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationSubqueryInjector.cs
anpete: ICollection doesn't have it.

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
maumar: can't - there is already using for NavigationExtensions (the one in Core)

maumar:
@anpete 

```
(QueryContext queryContext) => IEnumerable<<>f__AnonymousType127<Level2, ICollection<Level3>>> _InterceptExceptions(
    source: IEnumerable<<>f__AnonymousType127<Level2, ICollection<Level3>>> _Select(
        source: IEnumerable<TransparentIdentifier<Level1, Level2>> _ShapedQuery(
            queryContext: queryContext, 
            shaperCommandContext: SelectExpression: 
                SELECT [l1].[Id], [l1].[Date], [l1].[Name], [l1].[OneToMany_Optional_Self_InverseId], [l1].[OneToMany_Required_Self_InverseId], [l1].[OneToOne_Optional_SelfId], [l1.OneToOne_Optional_FK].[Id], [l1.OneToOne_Optional_FK].[Date], [l1.OneToOne_Optional_FK].[Level1_Optional_Id], [l1.OneToOne_Optional_FK].[Level1_Required_Id], [l1.OneToOne_Optional_FK].[Name], [l1.OneToOne_Optional_FK].[OneToMany_Optional_InverseId], [l1.OneToOne_Optional_FK].[OneToMany_Optional_Self_InverseId], [l1.OneToOne_Optional_FK].[OneToMany_Required_InverseId], [l1.OneToOne_Optional_FK].[OneToMany_Required_Self_InverseId], [l1.OneToOne_Optional_FK].[OneToOne_Optional_PK_InverseId], [l1.OneToOne_Optional_FK].[OneToOne_Optional_SelfId]
                FROM [Level1] AS [l1]
                LEFT JOIN [Level2] AS [l1.OneToOne_Optional_FK] ON [l1].[Id] = [l1.OneToOne_Optional_FK].[Level1_Optional_Id]
                ORDER BY [l1.OneToOne_Optional_FK].[Id], 
            shaper: TypedCompositeShaper<BufferedOffsetEntityShaper<Level1>, Level1, BufferedOffsetEntityShaper<Level2>, Level2, TransparentIdentifier<Level1, Level2>>), 
        selector: (TransparentIdentifier<Level1, Level2> t1) => new <>f__AnonymousType127<Level2, ICollection<Level3>>(
            t1.Inner, 
            ICollection<Level3> _ProjectCollectionNavigation(
                entity: Level1 _Include(
                    queryContext: queryContext, 
                    entity: t1.Outer, 
                    included: new object[]{ t1.Inner }, 
                    fixup: (QueryContext queryContext | Level1 entity | Object[] included) => 
                    {
                        return !(bool ReferenceEquals(included[0], null)) ? 
                        {
                            entity.OneToOne_Optional_FK = (Level2)included[0]
                            ((Level2)included[0]).OneToOne_Optional_FK_Inverse = entity
                            return Void queryContext.QueryBuffer.IncludeCollection(
                                includeId: 0, 
                                navigation: Level2.OneToMany_Optional, 
                                inverseNavigation: Level3.OneToMany_Optional_Inverse, 
                                targetEntityType: EntityType: Level3, 
                                clrCollectionAccessor: ClrICollectionAccessor<Level2, ICollection<Level3>, Level3>, 
                                inverseClrPropertySetter: ClrPropertySetter<Level3, Level2>, 
                                tracking: False, 
                                instance: (Level2)included[0], 
                                valuesFactory: () => IEnumerable<Level3> _ShapedQuery(
                                    queryContext: queryContext, 
                                    shaperCommandContext: SelectExpression: 
                                        SELECT [l1.OneToOne_Optional_FK.OneToMany_Optional].[Id], [l1.OneToOne_Optional_FK.OneToMany_Optional].[Level2_Optional_Id], [l1.OneToOne_Optional_FK.OneToMany_Optional].[Level2_Required_Id], [l1.OneToOne_Optional_FK.OneToMany_Optional].[Name], [l1.OneToOne_Optional_FK.OneToMany_Optional].[OneToMany_Optional_InverseId], [l1.OneToOne_Optional_FK.OneToMany_Optional].[OneToMany_Optional_Self_InverseId], [l1.OneToOne_Optional_FK.OneToMany_Optional].[OneToMany_Required_InverseId], [l1.OneToOne_Optional_FK.OneToMany_Optional].[OneToMany_Required_Self_InverseId], [l1.OneToOne_Optional_FK.OneToMany_Optional].[OneToOne_Optional_PK_InverseId], [l1.OneToOne_Optional_FK.OneToMany_Optional].[OneToOne_Optional_SelfId]
                                        FROM [Level3] AS [l1.OneToOne_Optional_FK.OneToMany_Optional]
                                        INNER JOIN (
                                            SELECT DISTINCT [l1.OneToOne_Optional_FK0].[Id]
                                            FROM [Level1] AS [l10]
                                            LEFT JOIN [Level2] AS [l1.OneToOne_Optional_FK0] ON [l10].[Id] = [l1.OneToOne_Optional_FK0].[Level1_Optional_Id]
                                        ) AS [t] ON [l1.OneToOne_Optional_FK.OneToMany_Optional].[OneToMany_Optional_InverseId] = [t].[Id]
                                        ORDER BY [t].[Id], 
                                    shaper: BufferedEntityShaper<Level3>))
                        }
                         : default(Void)
                    }
                ), 
                accessor: (Level1 prm) => prm?.OneToOne_Optional_FK?.OneToMany_Optional ?? (ICollection<Level3>)object IClrCollectionAccessor GetCollectionAccessor(Level2.OneToMany_Optional).Create())
        )), 
    contextType: TestModels.ComplexNavigationsModel.ComplexNavigationsContext, 
    logger: DiagnosticsLogger<Query>, 
    queryContext: queryContext)
```

maumar:
... and query model looks like this:

```
join Level2 l1.OneToOne_Optional_FK in DbSet<Level2>
on (Nullable<int>)Property([l1], "Id") equals Property([l1.OneToOne_Optional_FK], "Level1_Optional_Id")
order by Property([l1.OneToOne_Optional_FK], "Id") asc
select new <>f__AnonymousType127<Level2, ICollection<Level3>>(
    [l1.OneToOne_Optional_FK], 
    ICollection<Level3> _ProjectCollectionNavigation(
        entity: Level1 _Include(
            queryContext: queryContext, 
            entity: [l1], 
            included: new object[]{ [l1.OneToOne_Optional_FK] }, 
            fixup: (QueryContext queryContext | Level1 entity | Object[] included) => 
            {
                return !(bool ReferenceEquals(included[0], null)) ? 
                {
                    entity.OneToOne_Optional_FK = (Level2)included[0]
                    ((Level2)included[0]).OneToOne_Optional_FK_Inverse = entity
                    return Void queryContext.QueryBuffer.IncludeCollection(
                        includeId: 0, 
                        navigation: Level2.OneToMany_Optional, 
                        inverseNavigation: Level3.OneToMany_Optional_Inverse, 
                        targetEntityType: EntityType: Level3, 
                        clrCollectionAccessor: ClrICollectionAccessor<Level2, ICollection<Level3>, Level3>, 
                        inverseClrPropertySetter: ClrPropertySetter<Level3, Level2>, 
                        tracking: False, 
                        instance: (Level2)included[0], 
                        valuesFactory: () => 
                            from Level3 l1.OneToOne_Optional_FK.OneToMany_Optional in DbSet<Level3>
                            join AnonymousObject _l1.OneToOne_Optional_FK in 
                                (from Level1 l1 in DbSet<Level1>
                                join Level2 l1.OneToOne_Optional_FK in DbSet<Level2>
                                on (Nullable<int>)Property([l1], "Id") equals Property([l1.OneToOne_Optional_FK], "Level1_Optional_Id") into l1.OneToOne_Optional_FK_group
                                from Level2 l1.OneToOne_Optional_FK in 
                                    (from Level2 l1.OneToOne_Optional_FK_groupItem in [l1.OneToOne_Optional_FK_group]
                                    select [l1.OneToOne_Optional_FK_groupItem]).DefaultIfEmpty()
                                select new AnonymousObject(new object[]{ (object)EF.Property(?[l1.OneToOne_Optional_FK]?, "Id") })).Distinct()
                            on Property([l1.OneToOne_Optional_FK.OneToMany_Optional], "OneToMany_Optional_InverseId") equals (Nullable<int>)object [_l1.OneToOne_Optional_FK].GetValue(0)
                            order by object [_l1.OneToOne_Optional_FK].GetValue(0) asc
                            select [l1.OneToOne_Optional_FK.OneToMany_Optional])
                }
                 : default(Void)
            }
        ), 
        accessor: (Level1 prm) => prm?.OneToOne_Optional_FK?.OneToMany_Optional ?? (ICollection<Level3>)object IClrCollectionAccessor GetCollectionAccessor(Level2.OneToMany_Optional).Create())
)
```

maumar:
@smitpatel @anpete  new version up - added overload for ClrICollectionAccessor.Create and the test projecting reference nav plus collection

anpete:
@maumar Nice!

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -230,6 +231,21 @@ protected virtual void InterceptExceptions()
                     QueryContextParameter);
 
         /// <summary>
+        /// Rewrites collection navigation projections so that they can be handled by the Include pipeline.
+        /// </summary>
+        /// <param name="queryModel"> The query. </param>
+        protected virtual void RewriteProjectedCollectionNavigationsToIncludes([NotNull] QueryModel queryModel)
+        {
```
anpete: Check.NotNull?

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -230,6 +231,21 @@ protected virtual void InterceptExceptions()
...
+            var collectionNavigationIncludeRewriter = new CollectionNavigationIncludeExpressionRewriter(this);
+            queryModel.SelectClause.Selector = collectionNavigationIncludeRewriter.Visit(queryModel.SelectClause.Selector);
+            _queryCompilationContext.QueryAnnotations =
+                new ReadOnlyCollection<IQueryAnnotation>(
```
anpete: Add QCC.AddAnnotations method?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+                .GetDeclaredMethod(nameof(Metadata.Internal.NavigationExtensions.GetCollectionAccessor));
+
+        private static readonly MethodInfo _clrCollectionAccessorCreateMethodInfo
```
anpete: Move static fields to top.

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+                .GetDeclaredMethod(nameof(IClrCollectionAccessor.Create));
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public CollectionNavigationIncludeExpressionRewriter(
+            [NotNull] EntityQueryModelVisitor queryModelVisitor)
+        {
+            Check.NotNull(queryModelVisitor, nameof(queryModelVisitor));
```
anpete: Check can be removed for Internal class.

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            _queryModelVisitor = queryModelVisitor;
+            CollectionNavigationIncludeResultOperators = new List<IQueryAnnotation>();
```
anpete: Use property initializer

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitMethodCall(MethodCallExpression node)
+        {
+            Expression result = null;
+            if (node.Method.IsEFPropertyMethod())
```
anpete: This check needed or will be handled by BindNavigationPathPropertyExpression?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+                    if (expectedCallerType != property.DeclaringType.ClrType)
+                    {
+                        return expression;
+                    }
+
+                    expectedCallerType = property.ClrType;
+                }
+
+                CollectionNavigationIncludeResultOperators.Add(
+                    new IncludeResultOperator(properties.Select(p => p.Name), new QuerySourceReferenceExpression(querySource)));
```
anpete: This means we end up doing model binding twice for the nav path. Perhaps add a new ctor for this pre-bound case?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                var parameter = Expression.Parameter(querySource.ItemType, "prm");
```
anpete: querySource.ItemName for parameter name?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+                var accessorBody = BuildCollectionAccessorExpression(parameter, properties);
+
+                return Expression.Call(
+                    ProjectCollectionNavigationMethodInfo
+                        .MakeGenericMethod(querySource.ItemType, expression.Type),
+                    new QuerySourceReferenceExpression(querySource),
```
anpete: re-use qsre from above.

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class CollectionNavigationIncludeExpressionRewriter : ExpressionVisitorBase
+    {
+        private readonly EntityQueryModelVisitor _queryModelVisitor;
+
+        private static readonly MethodInfo _getCollectionAccessorMethodInfo
+            = typeof(Metadata.Internal.NavigationExtensions).GetTypeInfo()
+                .GetDeclaredMethod(nameof(Metadata.Internal.NavigationExtensions.GetCollectionAccessor));
```
anpete: Resolve the collection accessor at compile time

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            return result;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected override Expression VisitSubQuery(SubQueryExpression subQueryExpression)
+            => subQueryExpression;
```
anpete: Why prune?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
anpete: This can go away anyway if we eagerly resolve the collection accessor at compile time.

anpete:
We should probably track future composition cases. E.g. c.Orders.Select(o => o.Customer) ==> Include("Orders.Customer")

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
maumar: for case when there is result operator at the end, e.g customers.Select(c => c.Orders.Count())

maumar:
@anpete will add issues to track both tracking and composition, once this one is checked in

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
anpete: What about Select(new {  context.Customers.Select(c => c.Orders) })

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
maumar: removed

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
maumar: you mean, to resolve it only when compiling accessor body Expression (rather than statically when type is created)?

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
maumar: that wouldn't get optimized, so we would produce N+1. Pruning at subquery also handles all composition cases (customers.Select(c => c.Orders.Where(o => o.Id >5)). I would rather keep it strict and simple for now, and deal with it properly when we tackle the composition case.

maumar:
new version up @anpete

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
anpete: Sounds good, but may be easy to only prune when SQ MFC is a nav?

maumar: that works as far as the QM rewrite is concerned. However currently include pipeline doesn't work for qsres inside a subquery (IncludeReplacingExpressionVisitor needs to TrasnformExpressions inside  VisitSubquery). On top of that, there seems to be some more issues - the include QM fragment is partially evaluated on the client.

anpete: OK, file an issue.

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
anpete: Call GetCollectionAccessor in the compiler and bind the result into the compiled query as a constant.

src/EFCore/Query/ExpressionVisitors/Internal/CollectionNavigationIncludeExpressionRewriter.cs
maumar: filed #8722 and added test

maumar:
new version up @anpete

src/EFCore/Query/ResultOperators/Internal/IncludeResultOperator.cs
```diff
@@ -26,13 +26,25 @@ namespace Microsoft.EntityFrameworkCore.Query.ResultOperators.Internal
...
+        private INavigation[] _navigationPath;
         private IQuerySource _querySource;
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public IncludeResultOperator(
+            [NotNull] INavigation[] navigationPath, [NotNull] Expression pathFromQuerySource)
+            : this(navigationPath.Select(n => n.Name), pathFromQuerySource)
```
anpete: @maumar Missed this earlier, but this doesn't actually prevent re-binding. The goal is to prevent going from string -> INavigation, because we already have the INavigations.

maumar: @anpete this should be fine - the ctor saves INavigation[] into a field, and GetNavigationPath which performs the binding, returns the stored value rather than computing it from strings.

PR #8719, opened 2017-06-03
Tests: Reference Microsoft.Extensions.Primitives when compiling code with Roslyn

bricelam:
@smitpatel @mikeharder


bricelam:
Waiting for all tests to pass locally. Will merge once they do.

PR #8751, opened 2017-06-06
Add more ExecuteInTransaction overloads.

### LINKED ISSUE #7431: Add a version of the ExecuteInTransaction() that supports specifying the isolation level
Currently `IExecutionStrategy.ExecuteInTransaction()` always calls `Context.Database.BeginTransaction()` with no arguments. This means if you want to use an isolation level different from the default you have resort to `Execute()` and calling `BeginTransaction()` explicitly as described at https://github.com/aspnet/EntityFramework/issues/7318#issuecomment-269742840. 

It looks like it would be easy to add an overload that takes the isolation level.

AndriySvyryd:
Fixes #7431


src/EFCore/Extensions/ExecutionStrategyExtensions.cs
```diff
@@ -371,15 +651,17 @@ public static class ExecutionStrategyExtensions
         ///     Thrown if the operation has not succeeded after the configured number of retries.
         /// </exception>
         public static Task<TResult> ExecuteInTransactionAsync<TState, TResult>(
-            [NotNull] this IExecutionStrategy strategy,
+            [NotNull] IExecutionStrategy strategy,
```
divega: And this and the one above are not extension methods because...

AndriySvyryd: ... of the `beginTransaction` parameter. This overload is for code reuse and shouldn't be used directly.

divega: And it needs to be public because... :smile:

AndriySvyryd: ... it's used in `RelationalExecutionStrategyExtensions`

AndriySvyryd: And it would be useful for non-relational providers that want to add their own overloads.

src/EFCore.Relational/Extensions/RelationalExecutionStrategyExtensions.cs
```diff
@@ -0,0 +1,324 @@
...
+        /// <param name="verifySucceeded">
+        ///     A delegate that tests whether the operation succeeded even though an exception was thrown when the
+        ///     transaction was being committed.
+        /// </param>
+        /// <param name="context"> The context that will be used to start the transaction. </param>
+        /// <typeparam name="TContext"> The type of the supplied context instance. </typeparam>
+        /// <exception cref="RetryLimitExceededException">
+        ///     Thrown if the operation has not succeeded after the configured number of retries.
+        /// </exception>
+        public static void ExecuteInTransaction<TContext>(
```
divega: @AndriySvyryd @ajcvickers FYI, this method signature allows for type inference for `TContext` but the Intellisense experience is poor: while typing things in order you won't get any help with member completion on the action or the func because you haven't typed the context variable yet.

divega: And just to confirm, moving the context parameter before the action and the func fixes that.

AndriySvyryd: The same would apply to the `state` parameter. Should we move it too?

divega: I am fine with trying that. I already lost track of what we already vs. what we are adding. Are these breaking changes?

AndriySvyryd: I don't think we should do it for `state` because it's not present on all overloads and it would be better to have it separate from `context` to reduce the overload resolution confusion.

Adding the DbContext parameter in lambdas broke all usages already.

PR #8766, opened 2017-06-07
Migrations: Disable obsolete warnings in designer files

### LINKED ISSUE #8703: Pragma out obsolete warnings in generated code
While discussing #8654, one of the sticking points was that making the existing API obsolete would result in obsolete code in model snapshots. We could start preemptively pragmaing out obsolete warnings in this generated code so in the future as long as the obsolete code keeps working, we can do it without "breaking" generated code.

bricelam:
Fixes #8703

### DIFF
--- a/src/EFCore.Design/Migrations/Design/CSharpMigrationsGenerator.cs
+++ b/src/EFCore.Design/Migrations/Design/CSharpMigrationsGenerator.cs
@@ -151,13 +151,24 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Design
                 {
                     builder
                         .AppendLine("protected override void BuildTargetModel(ModelBuilder modelBuilder)")
-                        .AppendLine("{");
+                        .AppendLine("{")
+                        .DecrementIndent()
+                        .DecrementIndent()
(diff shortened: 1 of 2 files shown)


src/EFCore.Design/Migrations/Design/CSharpMigrationsGenerator.cs
```diff
@@ -151,13 +151,24 @@ private void AppendAutoGeneratedTag(IndentedStringBuilder builder)
                 {
                     builder
                         .AppendLine("protected override void BuildTargetModel(ModelBuilder modelBuilder)")
-                        .AppendLine("{");
+                        .AppendLine("{")
+                        .DecrementIndent()
+                        .DecrementIndent()
+                        .AppendLine("#pragma warning disable 612, 618")
```
smitpatel: I wish we indented warning in Scaffolded DbContext like this. Messes up auto-format of VS atm.

bricelam: 👍 Filed #8778

PR #8767, opened 2017-06-07
Add configuration package reference because Logging dropped it

pakrym:


src/EFCore.Relational/EFCore.Relational.csproj
```diff
@@ -22,6 +22,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CSharp" Version="$(CoreFxVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="$(AspNetCoreVersion)" />
```
natemcmaster: Is config.abstractions sufficient?

pakrym: Yes, but need to fix some test projects. Working on it

pakrym:
:up::date:

pakrym:
Merging to unblock the build

Issue #8186, opened 2017-04-16
Navs: nav expansion on GJ QSRE in projection doesn't introduce parameter


anpete:
```c#
from c in context.Set<Customer>()
join o in context.Set<Order>()
on c.CustomerID equals o.CustomerID into g
where c.CustomerID == "ALFKI"
select new { c, G = g.Select(o => _Include(o, o.OrderDetails)).ToList() })
```
produces:
```sql
SELECT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region], [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
FROM [Customers] AS [c]
LEFT JOIN [Orders] AS [o] ON [c].[CustomerID] = [o].[CustomerID]
WHERE [c].[CustomerID] = N'ALFKI'
ORDER BY [c].[CustomerID]
  
SELECT [o0].[OrderID], [o0].[ProductID], [o0].[Discount], [o0].[Quantity], [o0].[UnitPrice]
FROM [Order Details] AS [o0]

SELECT [o0].[OrderID], [o0].[ProductID], [o0].[Discount], [o0].[Quantity], [o0].[UnitPrice]
FROM [Order Details] AS [o0]
```
```c#
from c in context.Set<Customer>()
join o in context.Set<Order>()
on c.CustomerID equals o.CustomerID into g
where c.CustomerID == "ALFKI"
select new { c, G = g.Select(o => _Include(o, o.Customer)).ToList() }
```
produces:
```sql
SELECT [c].[CustomerID], [c].[Address], [c].[City], [c].[CompanyName], [c].[ContactName], [c].[ContactTitle], [c].[Country], [c].[Fax], [c].[Phone], [c].[PostalCode], [c].[Region], [o].[OrderID], [o].[CustomerID],
[o].[EmployeeID], [o].[OrderDate]
FROM [Customers] AS [c]
LEFT JOIN [Orders] AS [o] ON [c].[CustomerID] = [o].[CustomerID]
WHERE [c].[CustomerID] = N'ALFKI'
ORDER BY [c].[CustomerID]

[...]

PR #8769, opened 2017-06-07
[SMALL] Fix to #8186 - Navs: nav expansion on GJ QSRE in projection doesn't introduce parameter

maumar:
Problem was that when trying to find query to bind with via outer parameter we were too strict and would not find the appropriate query. We used to share the logic for regular binding and outer parameter binding, however the latter can be more loose and therefore bind to bigger range of queries (e.g. client-side groupjoin).

This also enables additional include scenarios when GJ qsre is projected - fixed a small bug in this area that was exposed by the main fix.

### DIFF
--- a/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -244,16 +244,17 @@ namespace Microsoft.EntityFrameworkCore.Query
         ///     Try and get the active SelectExpression for a given query source.
         /// </summary>
         /// <param name="querySource"> The query source. </param>
+        /// <param name="queryForOuterParameterBinding"> True if trying to bind to the query using outer paramter, otherwise false. </param>
         /// <returns>
         ///     A SelectExpression, or null.
         /// </returns>
-        public virtual SelectExpression TryGetQuery([NotNull] IQuerySource querySource)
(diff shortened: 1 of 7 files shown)


anpete:
Looks like one for @smitpatel :trollface:

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -244,16 +244,17 @@ public virtual void AddQuery([NotNull] IQuerySource querySource, [NotNull] Selec
         ///     Try and get the active SelectExpression for a given query source.
         /// </summary>
         /// <param name="querySource"> The query source. </param>
+        /// <param name="queryForOuterParameterBinding"> True if trying to bind to the query using outer paramter, otherwise false. </param>
         /// <returns>
         ///     A SelectExpression, or null.
         /// </returns>
-        public virtual SelectExpression TryGetQuery([NotNull] IQuerySource querySource)
+        public virtual SelectExpression TryGetQuery([NotNull] IQuerySource querySource, bool queryForOuterParameterBinding = false)
```
maumar: should we create overload here instead?

maumar:
cc @ajcvickers for breaking changes stuff

maumar:
closing for now as there seem to be a much broader/better fix

PR #8772, opened 2017-06-07
Fix #8197 - Diagnostics: Report number of rows returned by server on reader dispose

### LINKED ISSUE #8197: Diagnostics: Report number of rows returned by server on reader dispose
This is information can in turn be used by a profiler, e.g. to display a warning if the number of rows goes over certain threshold. 

This is related but not the same as https://github.com/aspnet/EntityFramework/issues/5089, which is about warning on the log.

anpete:
- Track DbDataReader.Read/Async invocation count via new delegating methods on RelationalDataReader.


anpete:
@AndriySvyryd We don't seem to have any tests for relational diagnostics. Filed #8779

PR #8774, opened 2017-06-07
RevEng: Remove unnecessary async

### LINKED ISSUE #8773: RevEng: Should it be async?
Today, the entry point to Reverse Engineer `DatabaseOperations.ScaffoldContextAsync()` is async, but all of the filesystem and database interaction is done synchronously. There is no asynchronous code.

We should either make the whole thing sync, or update it to be properly asynchronous.

Today, the only consumer of this API is `ef.exe` so async would not be utilized.

bricelam:
Resolves #8773

### DIFF
--- a/src/EFCore.Design/Scaffolding/Internal/CSharpScaffoldingGenerator.cs
+++ b/src/EFCore.Design/Scaffolding/Internal/CSharpScaffoldingGenerator.cs
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Threading;
-using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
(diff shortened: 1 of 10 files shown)


src/EFCore.Design/Scaffolding/Internal/CSharpScaffoldingGenerator.cs
```diff
@@ -89,7 +84,7 @@ public class CSharpScaffoldingGenerator : ScaffoldingCodeGenerator
                 resultingFiles.EntityTypeFiles.Add(entityTypeFileFullPath);
             }
 
-            return Task.FromResult(resultingFiles);
+            return resultingFiles;
```
bricelam: 🎉

src/EFCore.Design/Scaffolding/Internal/ModelScaffolder.cs
```diff
@@ -120,7 +115,7 @@ public class ModelScaffolder : IModelScaffolder
 
             CheckOutputFiles(fullOutputPath, contextName, model, overwriteFiles);
 
-            return ScaffoldingCodeGenerator.WriteCodeAsync(model, fullOutputPath, @namespace, contextName, connectionString, useDataAnnotations, cancellationToken);
+            return ScaffoldingCodeGenerator.WriteCode(model, fullOutputPath, @namespace, contextName, connectionString, useDataAnnotations);
```
smitpatel: I wonder if we should call this WriteCode 😛

bricelam: Internal. Can always change later...

Issue #8400, opened 2017-05-07
Reverse engineer bool column with default constraint to nullable bool property


ajcvickers:
A default constraint on a bool column mapped to a bool property with ValueGenerated.OnAdd has no value in EF--see #7089 for discussion.

Instead, if the property is nullable, then:
* Not setting the property will result in the default constraint being used.
* Setting to either true or false will result in the correct value being inserted.

popcatalin81:
I really don't know if this is the correct solution, because after this change we will have to undo "two" changes made by scaffolding: 1) make the bool not nullable (as it was in DB) and 2) remove the HasDefaultValueSql, in order to have wanted behavior in the model: not nullable boolean colums with default value.

ajcvickers:
@popcatalin81 That's also a reasonable interpretation if the default constraint is simply "true", but the nullable property preserves the use of the constraint in the database, whether it generates true or false, including it sometimes generating true and sometimes false based on other conditions.

ajcvickers:
@lajones Is this a breaking change because it will generate different code than before? If so, then I don't think we need to track this as a breaking change.

lajones:
@ajcvickers Yes. And also I'm wondering about round-tripping. I.e. if I regen the DB using migrations from the model that scaffolding would produce would it produce the same DB?

ajcvickers:
@lajones Probably wouldn't create the same DB in this case, but I think that is okay. @divega?

divega:
I do think it is ok if roundtripping doesn't generate the same column type in this case. The wrinkle is in the fact that the default SQL value can be an any expression that returns a Boolean. We are optimizing reverse engineering for the common case in which it is just a constant and as @ajcvickers explained having the property as a non-nullable Boolean has very little value. But going in the opposite direction, at least in the general case the SQL default is opaque to us and we don't know for sure if it can't evaluate to null.

BTW, thinking about the original problem we are trying to solve here, I am thinking now that I would be ok if this was a warning (either in reverse engineering or in the runtime) rather than making a choice for the user to change the type. But I cannot remember if there was a reason we didn't want to do that.

lajones:
@divega
> But going in the opposite direction, at least in the general case the SQL default is opaque to us and we don't know for sure if it can't evaluate to null

[...]

PR #8777, opened 2017-06-08
Non-nullable bool column with default constraint

lajones:
Fix for #8400. 

If Reverse Engineering would map a non-nullable column to a non-nullable bool property but the column has a default constraint then a) map it to a nullable bool property instead and b) issue a warning message.

This is to allow the user of the generated classes to distinguish between actively setting the property to 'false' and not setting the property and allowing the default constraint to generate the value in the database.

### DIFF
--- a/src/EFCore.Design/Diagnostics/DesignEventId.cs
+++ b/src/EFCore.Design/Diagnostics/DesignEventId.cs
@@ -32,7 +32,8 @@ namespace Microsoft.EntityFrameworkCore.Diagnostics
             PrimaryKeyColumnsNotMappedWarning,
             ForeignKeyReferencesNotMappedTableWarning,
             ForeignKeyReferencesMissingPrincipalKeyWarning,
-            ForeignKeyPrincipalEndContainsNullableColumnsWarning
+            ForeignKeyPrincipalEndContainsNullableColumnsWarning,
+            NonNullableBoooleanColumnHasDefaultConstraintWarning
         }
 
         private static readonly string _scaffoldingPrefix = DbLoggerCategory.Scaffolding.Name + ".";
@@ -91,5 +92,11 @@ namespace Microsoft.EntityFrameworkCore.Diagnostics
         ///     This event is in the <see cref="DbLoggerCategory.Scaffolding" /> category.
         /// </summary>
         public static readonly EventId ForeignKeyPrincipalEndContainsNullableColumnsWarning = MakeScaffoldingId(Id.ForeignKeyPrincipalEndContainsNullableColumnsWarning);
+
+        /// <summary>
+        ///     A column would be mapped to a bool type, is non-nullable and has a default constraint.
+        ///     This event is in the <see cref="DbLoggerCategory.Scaffolding" /> category.
+        /// </summary>
+        public static readonly EventId NonNullableBoooleanColumnHasDefaultConstraintWarning = MakeScaffoldingId(Id.NonNullableBoooleanColumnHasDefaultConstraintWarning);
     }
 }
--- a/src/EFCore.Design/Properties/DesignStrings.Designer.cs
+++ b/src/EFCore.Design/Properties/DesignStrings.Designer.cs
@@ -584,6 +584,18 @@ namespace Microsoft.EntityFrameworkCore.Internal
                     DesignEventId.ForeignKeyPrincipalEndContainsNullableColumnsWarning,
                     _resourceManager.GetString("LogForeignKeyPrincipalEndContainsNullableColumns")));
 
+        /// <summary>
+        ///     The column '{columnName}' would normally be mapped to a non-nullable bool property, but it has a default constraint. Such a column is mapped to a nullable bool property to allow a difference between setting the property to false and invoking the default constraint.
+        /// </summary>
+        public static readonly EventDefinition<string> LogNonNullableBoooleanColumnHasDefaultConstraint
+            = new EventDefinition<string>(
+                DesignEventId.NonNullableBoooleanColumnHasDefaultConstraintWarning,
+                LogLevel.Warning,
+                LoggerMessage.Define<string>(
+                    LogLevel.Warning,
+                    DesignEventId.NonNullableBoooleanColumnHasDefaultConstraintWarning,
+                    _resourceManager.GetString("LogNonNullableBoooleanColumnHasDefaultConstraint")));
+
         private static string GetString(string name, params string[] formatterNames)
         {
             var value = _resourceManager.GetString(name);
--- a/src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
+++ b/src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
@@ -311,13 +311,13 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
 
             var clrType = typeScaffoldingInfo.ClrType;
             var forceNullable = typeof(bool) == clrType && column.DefaultValue != null;
+            if (forceNullable)
+            {
+                Logger.NonNullableBoooleanColumnHasDefaultConstraintWarning(
+                    column.DisplayName);
+            }
             if (column.IsNullable || forceNullable)
             {
-                if (forceNullable)
-                {
-                    Logger.NonNullableBoooleanColumnHasDefaultConstraintWarning(
-                        column.DisplayName);
-                }
                 clrType = clrType.MakeNullable();
             }
 
--- a/src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
+++ b/src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
@@ -310,8 +310,14 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
             }
 
             var clrType = typeScaffoldingInfo.ClrType;
-            if (column.IsNullable)
+            var forceNullable = typeof(bool) == clrType && column.DefaultValue != null;
+            if (column.IsNullable || forceNullable)
             {
+                if (forceNullable)
+                {
+                    Logger.NonNullableBoooleanColumnHasDefaultConstraintWarning(
+                        column.DisplayName);
+                }
                 clrType = clrType.MakeNullable();
             }
 
@@ -361,7 +367,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
 
             if (!column.PrimaryKeyOrdinal.HasValue)
             {
-                property.IsRequired(!column.IsNullable);
+                property.IsRequired(!column.IsNullable && !forceNullable);
             }
 
             property.Metadata.Scaffolding().ColumnOrdinal = column.Ordinal;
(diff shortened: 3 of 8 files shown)


src/EFCore.Design/Diagnostics/DesignEventId.cs
```diff
@@ -91,5 +92,11 @@ private enum Id
         ///     This event is in the <see cref="DbLoggerCategory.Scaffolding" /> category.
         /// </summary>
         public static readonly EventId ForeignKeyPrincipalEndContainsNullableColumnsWarning = MakeScaffoldingId(Id.ForeignKeyPrincipalEndContainsNullableColumnsWarning);
+
+        /// <summary>
+        ///     A column would be mapped to a bool type, is non-nullable and has a default constraint.
+        ///     This event is in the <see cref="DbLoggerCategory.Scaffolding" /> category.
+        /// </summary>
+        public static readonly EventId NonNullableBoooleanColumnHasDefaultConstraintWarning = MakeScaffoldingId(Id.NonNullableBoooleanColumnHasDefaultConstraintWarning);
```
smitpatel: If we are not logging `DiagnosticSource` event then where are these ids used?
cc: @ajcvickers

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
```diff
@@ -310,8 +310,14 @@ protected virtual PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder builde
             }
 
             var clrType = typeScaffoldingInfo.ClrType;
-            if (column.IsNullable)
+            var forceNullable = typeof(bool) == clrType && column.DefaultValue != null;
+            if (column.IsNullable || forceNullable)
             {
+                if (forceNullable)
```
smitpatel: Move warning outside of this if block & right after checking conditions.

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
```diff
@@ -310,8 +310,14 @@ protected virtual PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder builde
             }
 
             var clrType = typeScaffoldingInfo.ClrType;
-            if (column.IsNullable)
+            var forceNullable = typeof(bool) == clrType && column.DefaultValue != null;
```
smitpatel: Create variable nullable taking value from Column.IsNullable & update the variable if conditions match (in same block as log message)

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
```diff
@@ -361,7 +367,7 @@ protected virtual PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder builde
 
             if (!column.PrimaryKeyOrdinal.HasValue)
             {
-                property.IsRequired(!column.IsNullable);
+                property.IsRequired(!column.IsNullable && !forceNullable);
```
smitpatel: This would be just !nullable.

test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/Expected/NonNullBoolWithDefault.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+
+namespace E2ETest.Namespace
+{
+    public partial class NonNullBoolWithDefault
+    {
+        public int Id { get; set; }
+        public bool? TestWithDefault { get; set; }
```
smitpatel: `BooleanWithDefaultValueSql` and likewise

lajones: OK. Don't think it's important as the test already explains that but OK. Why the `Sql` addition?

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
lajones: It's a relatively minor point, but I think the way it is now is much clearer. I think it's good for the developer to know that this is happening because nullability is being forced by the scaffolding, rather than it being any intrinsic property of the underlying database model. By saying `forceNullable` you are explaining that.

src/EFCore.Design/Diagnostics/DesignEventId.cs
lajones: @ajcvickers for response to this.

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
lajones: Could. Don't feel it makes make much difference.

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
lajones: See comment above - I think this is better.

test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/Expected/NonNullBoolWithDefault.cs
smitpatel: Just because it is Bool property with DefaultValueSql set. Becomes more explicit and easy to evaluate the expected output file as in fluent API, you can check if the property has HasDefaultValueSql generated. It's not important but being more explicit in test not gonna hurt and helps person to read the tests more easily.

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
smitpatel: Its minor point but no need to compute same thing multiple times, if we are having extra variable then why not use it in well way. We start with value in database, we update the variable with whatever conditions comes up and we set nullability based on that. Today its 2 conditions, it may expand more in future. Its better to right a condensed code. We don't need variables to explain what is happening. Logging message would make it very clear why we decided to introduce nullability. And if still doesn't look very clear, feel free to write a comment for explanation.

test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/Expected/NonNullBoolWithDefault.cs
lajones: I thought `bool` was obvious a) from the name of the test and b) from the produced type. And actually `HasDefaultValueSql()` is the choice of the scaffolder - it could (and did in the past) do something different if it wanted. The fact that the underlying database model has `ColumnModel.DefaultValue` set and that's what's being tested. But OK - like I say I don't think it really matters.

src/EFCore.Design/Diagnostics/DesignEventId.cs
ajcvickers: This is still the event ID for the ILogger message.

src/EFCore.Design/Properties/DesignStrings.Designer.cs
```diff
@@ -584,6 +584,18 @@ public static string ExistingFiles([CanBeNull] object outputDirectoryName, [CanB
                     DesignEventId.ForeignKeyPrincipalEndContainsNullableColumnsWarning,
                     _resourceManager.GetString("LogForeignKeyPrincipalEndContainsNullableColumns")));
 
+        /// <summary>
+        ///     The column '{columnName}' would normally be mapped to a non-nullable bool property, but it has a default constraint. Such a column is mapped to a nullable bool property to allow a difference between setting the property to false and invoking the default constraint.
+        /// </summary>
+        public static readonly EventDefinition<string> LogNonNullableBoooleanColumnHasDefaultConstraint
+            = new EventDefinition<string>(
+                DesignEventId.NonNullableBoooleanColumnHasDefaultConstraintWarning,
```
ajcvickers: Used here.

src/EFCore.Design/Properties/DesignStrings.Designer.cs
```diff
@@ -584,6 +584,18 @@ public static string ExistingFiles([CanBeNull] object outputDirectoryName, [CanB
                     DesignEventId.ForeignKeyPrincipalEndContainsNullableColumnsWarning,
                     _resourceManager.GetString("LogForeignKeyPrincipalEndContainsNullableColumns")));
 
+        /// <summary>
+        ///     The column '{columnName}' would normally be mapped to a non-nullable bool property, but it has a default constraint. Such a column is mapped to a nullable bool property to allow a difference between setting the property to false and invoking the default constraint.
```
ajcvickers: @divega fwlink this?

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
lajones: You still need to issue the warning message only if you are forcing nullability - and not just because something is `nullable` - so you will need the `forceNullable` variable (or at least an `if` statement with the same logic). I could create `forceNullable` and then immediately create `nullable` as well to do what you want. If the variable name was something like `wasNullableInDatabaseOrNullabilityWasForced` I'd be OK with it. But that's ridiculously long. Personally I find the other way much clearer so I'm going to stick with it for now.

lajones:
@smitpatel @ajcvickers I think I've addressed everything. So I'm going to go ahead and check in soon. Let me know if I shouldn't.

src/EFCore.Design/Properties/DesignStrings.Designer.cs
smitpatel: @lajones - wait for this

smitpatel:
Also can you rebase on latest dev, both CI services are red.

lajones:
@smitpatel Sure. Np.

src/EFCore.Design/Properties/DesignStrings.Designer.cs
lajones: @ajcvickers @smitpatel @divega are we doing a fwlink within the warning message? I didn't see a response from @divega

divega: @lajones I will have it in a couple of minutes. Trying to pick a temporary landing page for it until we have the actual topic.

divega: @lajones https://go.microsoft.com/fwlink/?linkid=851278, for now it points to https://docs.microsoft.com/en-us/ef/core/modeling/generated-properties#value-generated-on-add.

lajones: Thanks, @divega. I'll add this to the message and check in.

lajones:
Checked in with commit 61650f8e2.

ryanelian:
Excuse me, can I do something to disable this behavior?

I'm trying to re-scaffold a database context from an existing database (schema manually managed / updated by SQL) but suddenly EF Core puts `?` to many columns within the entity classes, causing over 122 compile errors...

PR #8788, opened 2017-06-08
Fix to #8186 - Navs: nav expansion on GJ QSRE in projection doesn't introduce parameter

maumar:
Problem was that when trying to find query to bind with via outer parameter we were too strict and would not find the appropriate query. However it is also safe to bind to grouping qsre, even though it is wrapped by client GroupJoin.

This also enables additional include scenarios when GJ qsre is projected - fixed a small bug in this area that was exposed by the main fix.


src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1969,19 +1969,20 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
 
         private ParameterExpression BindPropertyToOuterParameter(IQuerySource querySource, IProperty property, bool isMemberExpression)
         {
-            if (querySource != null && _canBindPropertyToOuterParameter)
+            if (querySource != null && _canBindPropertyToOuterParameter && ParentQueryModelVisitor != null)
```
maumar: filed #8790 to track enabling this for broader set of scenarios - too risky/too much work to do it now.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -2020,6 +2021,20 @@ var parameterWithSamePrefixCount
             return null;
         }
 
+        private bool CanBindToQuerySourceUsingOuterParam(IQuerySource querySource)
```
smitpatel: `CanBindToParentQMV`

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1969,19 +1969,20 @@ public override Expression BindMemberToValueBuffer(MemberExpression memberExpres
 
         private ParameterExpression BindPropertyToOuterParameter(IQuerySource querySource, IProperty property, bool isMemberExpression)
         {
-            if (querySource != null && _canBindPropertyToOuterParameter)
+            if (querySource != null && _canBindPropertyToOuterParameter && ParentQueryModelVisitor != null)
             {
-                var outerQueryModelVisitor = ParentQueryModelVisitor;
-                var outerSelectExpression = outerQueryModelVisitor?.TryGetQuery(querySource);
+                var isBindable = CanBindToQuerySourceUsingOuterParam(querySource);
```
smitpatel: variable/parameters not prefixed?

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
maumar: we already have property CanBindToParentQueryModel that is being used for non outer-param binding. How about CanBindToParentQueryModelUsingOuterParam?

smitpatel: but but we need `parameter` full not `param` 😢

maumar: yup, used "parameter" in the final name

PR #8789, opened 2017-06-08
RevEng: Add language to code generating APIs

### LINKED ISSUE #8752: RevEng: Pass language to provider code generators
We currently assume that the code generated by providers is C#. We should future-proof the API by passing them the language (or file extension) that we're generating for.

bricelam:
Fixes #8752

### DIFF
--- a/src/EFCore.Design/Scaffolding/Internal/CSharpDbContextGenerator.cs
+++ b/src/EFCore.Design/Scaffolding/Internal/CSharpDbContextGenerator.cs
@@ -23,6 +23,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
     public class CSharpDbContextGenerator : ICSharpDbContextGenerator
     {
         private const string EntityLambdaIdentifier = "entity";
+        private const string Language = "CSharp";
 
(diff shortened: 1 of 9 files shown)


bricelam:
> I like the idea of passing the language in, but it feels like there should be a public API providing a list of languages that a given provider supports, from which the user can choose - is there a reason why the "CSharp" is hard-coded in a couple of different places?

There are a few well-known ones like CSharp, VisualBasic, and FSharp, but there is nothing preventing new .NET-based languages from consuming EF.

> It feels like the user should get to say what language they want to generate at the cmd-line - but I don't see that part of it?

When we support multiple languages, the language will be determined by the project. We have a lot of work to do in this area. This PR was just about future-proofing the non-Internal APIs to be able to handle it in the future.

> If the user chose a language that wasn't supported by the provider, what message would they see / how do we help them understand what went wrong and decide what to do next? (This is another reason why I think the provider should be able to say what languages it supports)

If a provider doesn't support a given language, the user will be able to add a design-time service to light it up. Some features can work without explicit language support in the provider. E.g. instead of the provider writing `.ForSqlServerIsClustered()`, we can fallback to just `.HasAnnotation("SqlServer:Clustered", true)`.

PR #8793, opened 2017-06-08
RevEng: Make TableSelectionSet Internal

### LINKED ISSUE #8713: RevEng: Clean up TableSelectionSet
The code providers have to write to use `TableSelectionSet` is a little awkward. We should look into improving it.

bricelam:
Resolves #8713


bricelam:
@smitpatel Agree we should remove it ASAP. Just scoping my changes to fit into 2.0.0-preview2. It would also be good to limit the amount of data being read by filtering on the server instead of the client.

PR #8794, opened 2017-06-09
Log connection error at debug level for Exists path

### LINKED ISSUE #8651: Exists check logs error "An error occurred using the connection..."
I am seeing a lot of errors like this when using the tools.

> An error occurred using the connection to database 'main' on server 'Test.db'.

There are two issues with them:
1. We expect these errors when checking if the database exists. Can we suppress (or dampen their log level) in this case?
2. ~~Logging errors terminates the process in PMC meaning you have to call `Update-Database` a few times before it succeeds.~~ (fixed)

smitpatel:
Resolves #8651

### DIFF
--- a/src/EFCore.Relational/Storage/IRelationalConnection.cs
+++ b/src/EFCore.Relational/Storage/IRelationalConnection.cs
@@ -44,12 +44,14 @@ namespace Microsoft.EntityFrameworkCore.Storage
         /// <summary>
         ///     Opens the connection to the database.
         /// </summary>
+        /// <param name="errorsExpected"> Indicate if the connection errors are expected and should be logged as debug message. </param>
         /// <returns> True if the underlying connection was actually opened; false otherwise. </returns>
-        bool Open();
+        bool Open(bool errorsExpected = false);
 
         /// <summary>
         ///     Asynchronously opens the connection to the database.
         /// </summary>
+        /// <param name="errorsExpected"> Indicate if the connection errors are expected and should be logged as debug message. </param>
         /// <param name="cancellationToken">
         ///     A <see cref="CancellationToken" /> to observe while waiting for the task to complete.
         /// </param>
@@ -57,7 +59,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
         ///     A task that represents the asynchronous operation, with a value of true if the connection
         ///     was actually opened.
         /// </returns>
-        Task<bool> OpenAsync(CancellationToken cancellationToken = default(CancellationToken));
+        Task<bool> OpenAsync(bool errorsExpected = false, CancellationToken cancellationToken = default(CancellationToken));
 
         /// <summary>
         ///     Closes the connection to the database.
--- a/src/EFCore.Relational/Storage/RelationalConnection.cs
+++ b/src/EFCore.Relational/Storage/RelationalConnection.cs
@@ -175,7 +175,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
                 throw new InvalidOperationException(RelationalStrings.TransactionAlreadyStarted);
             }
 
-            await OpenAsync(cancellationToken);
+            await OpenAsync(cancellationToken: cancellationToken);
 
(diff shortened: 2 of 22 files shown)


smitpatel:
Manually verified that it works. Not sure where to add tests.

src/EFCore.Relational/Storage/IRelationalConnection.cs
```diff
@@ -44,20 +44,22 @@ public interface IRelationalConnection : IRelationalTransactionManager, IDisposa
         /// <summary>
         ///     Opens the connection to the database.
         /// </summary>
+        /// <param name="logErrorAsDebug"> Indicate if the connection error should be logged as debug message. </param>
         /// <returns> True if the underlying connection was actually opened; false otherwise. </returns>
-        bool Open();
+        bool Open(bool logErrorAsDebug = false);
```
bricelam: I wonder if this should be more general. Something like `errorsExpected`...

smitpatel: @divega & @ajcvickers - to determine the name of parameter.

ajcvickers: Fine with `errorsExpected` or `logErrorAsDebug`. @divega ?

@smitpatel Please don't block on this name for merging.

src/EFCore.Relational/Storage/RelationalConnection.cs
```diff
@@ -333,11 +335,12 @@ public virtual bool Open()
         /// <param name="cancellationToken">
         ///     A <see cref="CancellationToken" /> to observe while waiting for the task to complete.
         /// </param>
+        /// <param name="logErrorAsDebug"> Indicate if the connection error should be logged as debug message. </param>
         /// <returns>
         ///     A task that represents the asynchronous operation, with a value of true if the connection
         ///     was actually opened.
         /// </returns>
-        public virtual async Task<bool> OpenAsync(CancellationToken cancellationToken = default(CancellationToken))
+        public virtual async Task<bool> OpenAsync(CancellationToken cancellationToken = default(CancellationToken), bool logErrorAsDebug = false)
```
ajcvickers: Switch the order of these around so the cancellationToken is last.

smitpatel: I tried that first. But we are using this in many places with cancellation token passed. So need named parameter or both of them. What is more preferable from user perspective (if users are supposed to use this API)?

src/EFCore.Relational/Storage/IRelationalConnection.cs
smitpatel: Changed to `errorsExpected` except for the log message where `logErrorAsDebug` is more appropriate.

src/EFCore.Relational/Storage/RelationalConnection.cs
smitpatel: Switched order.

test/EFCore.Relational.Tests/RelationalEventIdTest.cs
```diff
@@ -100,8 +100,8 @@ private class FakeRelationalConnection : IRelationalConnection
             public bool Close() => throw new NotImplementedException();
             public void CommitTransaction() => throw new NotImplementedException();
             public void Dispose() => throw new NotImplementedException();
-            public bool Open() => throw new NotImplementedException();
-            public Task<bool> OpenAsync(CancellationToken cancellationToken = default(CancellationToken)) => throw new NotImplementedException();
+            public bool Open(bool logErrorAsDebug = false) => throw new NotImplementedException();
```
divega: Just in case, did we miss a rename here or is this the one that stayed like this on purpose?

smitpatel: Missed it.

divega: I see it is in a test.

Issue #7397, opened 2017-01-12
DbSet<>.Local InvalidOperationException with ListBox.ItemsSource


ahoischen:
Using the DbSet<>.Local property as the source for an System.Windows.Controls.ListBox and then adding and removing an instance causes an InvalidOperationException to be thrown. This is not an external issue as this is following your [recommended use case for this property](https://docs.microsoft.com/de-de/ef/core/api/microsoft.entityframeworkcore.dbset-1#Microsoft_EntityFrameworkCore_DbSet_1_Local).
```
System.InvalidOperationException was unhandled
  HResult=-2146233079
  Message=Collection Remove event must specify item position.
  Source=PresentationFramework
  StackTrace:
       at MS.Internal.Data.EnumerableCollectionView.ProcessCollectionChanged(NotifyCollectionChangedEventArgs args)
       at System.Windows.Data.CollectionView.OnCollectionChanged(Object sender, NotifyCollectionChangedEventArgs args)
       at Microsoft.EntityFrameworkCore.ChangeTracking.LocalView`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)
       at Microsoft.EntityFrameworkCore.ChangeTracking.LocalView`1.StateManagerChangedHandler(InternalEntityEntry entry, EntityState previousState)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.LocalViewListener.StateChanged(InternalEntityEntry entry, EntityState oldState, Boolean fromQuery)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntryNotifier.StateChanged(InternalEntityEntry entry, EntityState oldState, Boolean fromQuery)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges)
       at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState entityState, Boolean acceptChanges)
       at Microsoft.EntityFrameworkCore.ChangeTracking.EntityEntry.set_State(EntityState value)
       at Microsoft.EntityFrameworkCore.DbContext.Remove[TEntity](TEntity entity)
       at Microsoft.EntityFrameworkCore.Internal.InternalDbSet`1.Remove(TEntity entity)
[...]

PR #8796, opened 2017-06-09
Add ToObservableCollection to LocalView so that it can be used for data binding

ajcvickers:
Issue #7397

In EF Core, DbSet.Local is a pure view over the state manager. This is much more efficient for normal local queries, etc., but doesn't work for data binding because it does not preserve order or maintain indexes. This change adds DbSet.Local.ToObservableCollection() which creates a less efficient but maintains order and indexes.


divega:
I wondering if we should consider implementing `IListSource` and `BindingList<T>`. Probably none of this is urgent.

ajcvickers:
@divega Do you mean implementing on some other class in addition to what we already have?

ajcvickers:
Since this is a minor breaking change I would love to get it into preview2. @divega Did you review this? Could someone else look at it? @smitpatel @bricelam @AndriySvyryd @anpete

divega:
@ajcvickers I meant `IListSource` on `.Local` and binding list in a different class. I can create a new issue to review how what we have works with different flavors of data binding.

ajcvickers:
Sure, we should review what we have.

PR #8806, opened 2017-06-09
Don't use a temporary table for batching inserts.

### LINKED ISSUE #8322: SQL for string-updates sometimes uses nvarchar(450) if a string field is used in an index (not key)
Entity Framework Core auto-generates SQL-Statements to cause updates to happen. When a string-value is used as a key, it will by default use nvarchar(450) instead of nvarchar(max). 

The problem is that this also happens in a few (but not all) code generated for string values which are in an index (instead of a key). Specifically, the temporary table which is used in the SQL for merge operations will incorrectly assume an nvarchar(450) and then fail when trying to place the initial nvarchar(max) value in this temp table. 

The exception will then look be: `SqlException: String or binary data would be truncated.` The full exception shown is: 
```
Microsoft.EntityFrameworkCore.DbUpdateException occurred
  HResult=0x80131500
  Message=An error occurred while updating the entries. See the inner exception for details.
  Source=Microsoft.EntityFrameworkCore.Relational
  StackTrace:
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.Execute(Tuple`2 parameters)
   at Microsoft.EntityFrameworkCore.Storage.Internal.SqlServerExecutionStrategy.Execute[TState,TResult](Func`2 operation, Func`2 verifySucceeded, TState state)
   at Microsoft.EntityFrameworkCore.ExecutionStrategyExtensions.Execute[TState,TResult](IExecutionStrategy strategy, Func`2 operation, TState state)
   at Microsoft.EntityFrameworkCore.Update.Internal.BatchExecutor.Execute(IEnumerable`1 commandBatches, IRelationalConnection connection)
   at Microsoft.EntityFrameworkCore.Storage.RelationalDatabase.SaveChanges(IReadOnlyList`1 entries)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(IReadOnlyList`1 entriesToSave)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at ToSic.Eav.Persistence.Efc.Models.EavDbContext.SaveChangesAndLogEverything(Boolean acceptAllChangesOnSuccess) in C:\Projects\eav-server\ToSic.Eav.Persistence.EFC11\ModelsEnhancements\EavDbContext.cs:line 39

Inner Exception 1:
SqlException: String or binary data would be truncated.
The statement has been terminated.
```

The SQL which was sent to the SQL Server was the following: 
```
exec sp_executesql N'SET NOCOUNT ON;
DECLARE @toInsert0 TABLE ([AttributeID] int, [ChangeLogCreated] int, [ChangeLogDeleted] int, [ChangeLogModified] int, [EntityID] int, [Value] nvarchar(450), [_Position] [int]);
INSERT INTO @toInsert0
VALUES (@p0, @p1, @p2, @p3, @p4, @p5, 0),
(@p6, @p7, @p8, @p9, @p10, @p11, 1);

DECLARE @inserted0 TABLE ([ValueID] int, [_Position] [int]);
MERGE [ToSIC_EAV_Values] USING @toInsert0 AS i ON 1=0
WHEN NOT MATCHED THEN
INSERT ([AttributeID], [ChangeLogCreated], [ChangeLogDeleted], [ChangeLogModified], [EntityID], [Value])
VALUES (i.[AttributeID], i.[ChangeLogCreated], i.[ChangeLogDeleted], i.[ChangeLogModified], i.[EntityID], i.[Value])
OUTPUT INSERTED.[ValueID], i._Position
INTO @inserted0;

SELECT [t].[ValueID] FROM [ToSIC_EAV_Values] t
INNER JOIN @inserted0 i ON ([t].[ValueID] = [i].[ValueID])
ORDER BY [i].[_Position];

',N'@p0 int,@p1 int,@p2 int,@p3 int,@p4 int,@p5 nvarchar(max) ,@p6 int,@p7 int,@p8 int,@p9 int,@p10 int,@p11 nvarchar(450)',@p0=5757,@p1=15407,@p2=NULL,@p3=NULL,@p4=21419,@p5=N'<p><em>You can&nbsp;quickly change / remove this text&nbsp;once you''ve figured out how to work with the tiles.&nbsp;</em></p>
<p>This simple&nbsp;app let''s&nbsp;you choose icons to create nice tiles. The initial output is&nbsp;optimized for&nbsp;bootstrap skins.&nbsp;Your next steps are probably:</p>
<ol>
<li>Edit / add tiles (hover over the tiles and use the toolbar like the <i class="fa fa-pencil"></i>)</li>
<li>Write something else here or change the default look of <em>all</em> tiles in this group&nbsp;(just hover over this text and click <i class="fa fa-pencil"></i> in the toolbar)</li>
<li>Change the appearence of one&nbsp;icon (<i class="fa fa-pencil"></i> then change the presentation-settings <i class="fa fa-toggle-off"></i> at the bottom of the form)</li>
<li>Modify the HTML template as you wish&nbsp;(click on the <i class="fa fa-ellipsis-h"></i> till you get <i class="fa fa-code"></i>)</li>
</ol>',@p6=5756,@p7=15407,@p8=NULL,@p9=NULL,@p10=21419,@p11=N'Content Tiles App'
```

As you can see, the `@p5 ` is correctly treated as an `nvarchar(max)` but the temporary table `@toInsert0` creates a column `[Value]` which is cast as `nvarchar(450)` - triggering this problem. 

In my case I had a DB first model which simply imported the indexes into the OnModelCreating ModelBuilder on my DbContext resulting in this index being managed:
```
entity.HasIndex(e => new { e.Value, e.ChangeLogCreated, e.EntityId, e.ChangeLogDeleted, e.AttributeId, e.ValueId })
   .HasName("IX_EAV_Values2");
```

When I commented out the information that `e.Value` is part of the index, everything worked flawlessly:
```
// 2017-04-28 disabled e.Value in this index - it also seems to affect sql queries to 450 chars on that field!
entity.HasIndex(e => new { /*e.Value,*/ e.ChangeLogCreated, e.EntityId, e.ChangeLogDeleted, e.AttributeId, e.ValueId })
    .HasName("IX_EAV_Values2");
```

It's important to note that many changes to this data type were _not affected/hurt_ by the index-information. Just some very specific ones caused this problem. 

### Steps to reproduce
My code is too complex to use when reproducing, sorry. But I hope the full explanation above clarifies everything.

### Further technical details
EF Core version: 1.1.1
Operating system: Windows 10 Dev environment, Windows 2012 R2 Web Server, IIS
IDE: Visual Studio 2017

### LINKED ISSUE #8415: Update: Relational: Potential optimization in MERGE SQL batching path?
(Based on feedback from  https://www.brentozar.com/archive/2017/05/case-entity-framework-cores-odd-sql) 

Right now when using the MERGE batch update path, we perform a final ORDER BY on the server to retrieve the store gen'd values in the correct batch order. However, it could be that the values are already in the correct order, so we could perform the ordering on the client, lazily, by detecting out of order rows when they occur. NB. This would sometimes require client buffering of the reader.

AndriySvyryd:
Fixes #8415
Fixes #8322

### DIFF
--- a/src/EFCore.SqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
+++ b/src/EFCore.SqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
@@ -40,9 +40,6 @@ namespace Microsoft.EntityFrameworkCore.Update.Internal
             IReadOnlyList<ModificationCommand> modificationCommands,
             int commandPosition)
         {
-            Check.NotNull(commandStringBuilder, nameof(commandStringBuilder));
-            Check.NotEmpty(modificationCommands, nameof(modificationCommands));
-
             if (modificationCommands.Count == 1
                 && modificationCommands[0].ColumnModifications.All(o =>
                     !o.IsKey
@@ -140,7 +137,6 @@ namespace Microsoft.EntityFrameworkCore.Update.Internal
         }
 
         private const string InsertedTableBaseName = "@inserted";
-        private const string ToInsertTableBaseName = "@toInsert";
         private const string ToInsertTableAlias = "i";
         private const string PositionColumnName = "_Position";
         private const string PositionColumnDeclaration = "[" + PositionColumnName + "] [int]";
@@ -156,28 +152,6 @@ namespace Microsoft.EntityFrameworkCore.Update.Internal
         {
             AppendDeclareTable(
                 commandStringBuilder,
-                ToInsertTableBaseName,
-                commandPosition,
-                writeOperations,
-                PositionColumnDeclaration);
-
-            commandStringBuilder.Append("INSERT INTO ").Append(ToInsertTableBaseName).Append(commandPosition);
-            AppendValuesHeader(commandStringBuilder, writeOperations);
-            AppendValues(commandStringBuilder, writeOperations, "0");
-            for (var i = 1; i < modificationCommands.Count; i++)
-            {
-                commandStringBuilder.Append(",").AppendLine();
-                AppendValues(
-                    commandStringBuilder,
-                    modificationCommands[i].ColumnModifications.Where(o => o.IsWrite).ToList(),
-                    i.ToString(CultureInfo.InvariantCulture));
-            }
-            commandStringBuilder
-                .AppendLine(SqlGenerationHelper.StatementTerminator)
-                .AppendLine();
-
-            AppendDeclareTable(
-                commandStringBuilder,
                 InsertedTableBaseName,
                 commandPosition,
                 keyOperations,
@@ -190,10 +164,10 @@ namespace Microsoft.EntityFrameworkCore.Update.Internal
                 commandStringBuilder,
                 name,
                 schema,
-                ToInsertTableBaseName,
-                commandPosition,
                 ToInsertTableAlias,
-                writeOperations);
+                modificationCommands,
+                writeOperations,
+                PositionColumnName);
             AppendOutputClause(
                 commandStringBuilder,
                 keyOperations,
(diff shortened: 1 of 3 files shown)


src/EFCore.SqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
```diff
@@ -370,16 +363,22 @@ private string GetTypeNameForCopy(IProperty property)
...
+                    }
                 }
             }
 
-            return property.ClrType == typeof(byte[])
-                   && (typeName.Equals("rowversion", StringComparison.OrdinalIgnoreCase)
-                       || typeName.Equals("timestamp", StringComparison.OrdinalIgnoreCase))
-                ? (property.IsNullable ? "varbinary(8)" : "binary(8)")
-                : typeName;
+            if (property.ClrType == typeof(byte[])
```
anpete: Should this be in the type mapper?

src/EFCore.SqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
```diff
@@ -239,39 +213,61 @@ public class SqlServerUpdateSqlGenerator : UpdateSqlGenerator, ISqlServerUpdateS
...
+                    i.ToString(CultureInfo.InvariantCulture));
+            }
+
+            commandStringBuilder
+                .Append(") AS ").Append(toInsertTableAlias)
+                .Append(" (")
+                .AppendJoin(
+                    writeOperations,
+                    SqlGenerationHelper,
+                    (sb, o, helper) => { helper.DelimitIdentifier(sb, o.ColumnName); });
```
anpete: Remove braces?

anpete:
This doesn't implement the ordering elimination optimization so we should still track that somehow.

test/EFCore.SqlServer.Tests/Update/SqlServerUpdateSqlGeneratorTest.cs
```diff
@@ -97,12 +97,10 @@ public void AppendBulkInsertOperation_appends_insert_if_store_generated_columns_
             var grouping = sqlGenerator.AppendBulkInsertOperation(stringBuilder, new[] { command, command }, 0);
 
             Assert.Equal(
-                "DECLARE @toInsert0 TABLE ([Name] nvarchar(max), [Quacks] int, [ConcurrencyToken] varbinary(max), [_Position] [int]);" + Environment.NewLine +
-                "INSERT INTO @toInsert0" + Environment.NewLine +
-                "VALUES (@p0, @p1, @p2, 0)," + Environment.NewLine +
-                "(@p0, @p1, @p2, 1);" + Environment.NewLine + Environment.NewLine +
                 "DECLARE @inserted0 TABLE ([Id] int, [_Position] [int]);" + Environment.NewLine +
-                "MERGE [dbo].[Ducks] USING @toInsert0 AS i ON 1=0" + Environment.NewLine +
+                "MERGE [dbo].[Ducks] USING (" + Environment.NewLine +
```
anpete: This change only affects one SQL baseline? Do we have enough coverage of SQL gen here?

anpete: Also, consider using verbatim strings here like we do in query - The Env.NLs add a lot of clutter.

src/EFCore.SqlServer/Update/Internal/SqlServerUpdateSqlGenerator.cs
AndriySvyryd: It's very update-specific, I doubt it could be reused.

AndriySvyryd:
I'll keep https://github.com/aspnet/EntityFramework/issues/8415 open

test/EFCore.SqlServer.Tests/Update/SqlServerUpdateSqlGeneratorTest.cs
AndriySvyryd: I believe that ideally a single change in SQL gen should require a single change in the baseline.

anpete: And I believe in 🦄 and 🌈 😄

PR #8807, opened 2017-06-09
Simplified ignored include message. Fixes #7162.

### LINKED ISSUE #7162: Query: Better warning and XML doc comments about ignored Includes
Suggestion from @mguinness to make the warning clearer. We could also FWLink to the docs on this topic.

<pre>
The Include operation for navigation: 'c.State' <b>is unnecessary and</b>
was ignored because the target navigation is not reachable in the final query results. 
</pre>

We should also doc when include is/isn't needed in the xml comments for Include.

divega:
What the customer suggested seems slightly shorter and easier to understand. It also brings back "for navigation" before the actual navigation and I removed the word "target" which didn't help IMO.

### DIFF
--- a/src/EFCore/Properties/CoreStrings.Designer.cs
+++ b/src/EFCore/Properties/CoreStrings.Designer.cs
@@ -1413,7 +1413,7 @@ namespace Microsoft.EntityFrameworkCore.Internal
                 entityType);
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


divega:
@ajcvickers updated.

anpete:
Squash commits :trollface:

divega:
@anpete I really tried. I only made one mistake :smile:

Issue #8645, opened 2017-05-31
Flow unique constraints from reverse engineering to Migrations


smitpatel:
The logic introduced in PR https://github.com/aspnet/EntityFramework/pull/8089 is inaccurate.

We should not be scaffolding unique index for unique constraints.
Practically they are same (ref: https://stackoverflow.com/questions/2675474/what-is-the-difference-between-a-unique-constraint-and-a-unique-index)

But metadata wise they are different. If there is unique constraint then we should call `HasAlternateKey` instead of HasIndex.IsUnique` regardless of it being referenced by relationship or not.

Though the underlying issue is https://github.com/aspnet/EntityFramework/issues/7956#issuecomment-291997074
> I've been looking at this offline. It's OK to not call HasAlternateKey() and it turns out we were already calling HasIndex().

At present, we never ask for metadata about the constraints to database. We generate PK based on PrimaryKeyOrdinal of column & we generate AK based on the relationship referenced columns. (therefore we incorrectly identify our unique constraints as unique index atm)

It would preserve https://github.com/aspnet/EntityFramework/issues/7956 too without complicating our logic. In #7956, there is unique index, we don't scaffold IsRequired for unique index since one null value is allowed.
If there is unique Constraint then we should scaffold `IsRequired` since SqlServer would make columns non-null too.

ajcvickers:
@smitpatel We should never generate HasAlternateKey; it is never useful. In EF terms, HasAlternateKey is a way of specifying that something is the principal end of the relationship without saying which relationship that is. But you have to tell us that in the relationship API anyway, so HasAlternateKey has no value; it's a pointless API.

So for something like this:
```C#
modelBuilder.Entity<Blog>()
    .HasMany(e => e.Posts)
    .WithOne(e => e.Blog)
    .HasPrincipalKey(e => e.BlogId);
```
You could also add this:
```C#
modelBuilder.Entity<Blog>()
    .HasAlternateKey(e => e.BlogId);
```
and it would build the same model. But there's really no point since...it builds the same model.

Why this is important is that the key values at the principal end of a relationship cannot be modified. So once something is marked with HasAlternateKey it becomes more restrictive. If it really is the principal end, then fine--it's a restriction we have. But if it has been called for something that isn't the principal end, then the model now has a restriction that it doesn't need to have, and one that historically causes issues for applications.

Maybe now is the time to finally get rid of HasAlternateKey? Or at least obsolete it.

smitpatel:
[...]

PR #8808, opened 2017-06-09
RevEng: Clean up DatabaseModel

### LINKED ISSUE #8714: RevEng: Clean up DatabaseModel
We should evaluate whether it would be worth making the DatabaseModel classes look a little more like the Migraitons operations classes.

bricelam:
Changes:
* Add PrimaryKey & UniqueConstraint
* Align names with rest of stack
* Remove non-essential properties
* Move rarely-used properties to annotations

Resolves #8714


src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
```diff
@@ -521,49 +533,44 @@ protected virtual IMutableForeignKey VisitForeignKey([NotNull] ModelBuilder mode
...
                 return null;
             }
 
-            var principalPropertiesMap = foreignKeyColumns
+            var principalPropertiesMap = foreignKey.PrincipalColumns
                 .Select(
-                    fc => new Tuple<IMutableProperty, ColumnModel>(
-                        principalEntityType.FindProperty(GetPropertyName(fc.PrincipalColumn)),
-                        fc.PrincipalColumn));
+                    fc => (property: principalEntityType.FindProperty(GetPropertyName(fc)), column: fc));
```
smitpatel: nice!

src/EFCore.Relational/Scaffolding/Metadata/Column.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class Column : Annotatable
+    {
+        public virtual Table Table { get; [param: NotNull] set; }
+        public virtual string Name { get; [param: NotNull] set; }
+        public virtual bool IsNullable { get; [param: NotNull] set; }
+        public virtual string StoreType { get; [param: CanBeNull] set; }
+        public virtual string DefaultValueSql { get; [param: CanBeNull] set; }
+        public virtual string ComputedColumnSql { get; [param: CanBeNull] set; }
+        public virtual ValueGenerated? ValueGenerated { get; set; }
```
smitpatel: CanBeNull?

AndriySvyryd:
Seems like PrimaryKey and UniqueConstraint could be the same class.
Are you planning to prefix the rest of the types with 'Database'?

src/EFCore.Sqlite.Core/Scaffolding/Internal/SqliteDatabaseModelFactory.cs
```diff
@@ -65,7 +45,7 @@ private void ResetState()
         /// </summary>
         public virtual DatabaseModel Create(string connectionString, IEnumerable<string> tables, IEnumerable<string> schemas)
         {
-            Check.NotEmpty(connectionString, nameof(connectionString));
+            Check.NotNull(connectionString, nameof(connectionString));
```
smitpatel: intentional?

smitpatel:
@AndriySvyryd - I think database prefix is used for types where it had clash with runtime metadata

src/EFCore.Relational/Scaffolding/Metadata/Column.cs
bricelam: This is implied for `Nullable<T>` properties

src/EFCore.Sqlite.Core/Scaffolding/Internal/SqliteDatabaseModelFactory.cs
bricelam: Yes. I'm pretty sure empty connection strings are valid on SQL Server and we allow them elsewhere.

bricelam:
> Seems like PrimaryKey and UniqueConstraint could be the same class.

No. (Insert Steve Carell GIF here.) They may have the same shape, but they'll ultimately (with #8645) be treated differently.

> Are you planning to prefix the rest of the types with Database?

TBD. Smit's comment is correct. I only did it where it collided with our other metadata.

bricelam:
Updated SQL Server. Just a few more TODOs to address before it's ready to merge...

AndriySvyryd:
@bricelam They can be treated differently even if they have the same type, like Key in the metadata

bricelam:
I'm fine with introducing a base type if we think it's valuable, but as standalone objects, you should be able to tell what type it is because they have different characteristics:
* `PRIMARY KEY` is `CLUSTERED` by default; `UNIQUE` constraint is `NONCLUSTERED` by default.
* `PRIMARY KEY` requires `NOT NULL` columns; `UNIQUE` constraint allow `NULL` columns.

bricelam:
I view the core metadata differently. There, they *are* the same type as far as EF is concerned. We just distinguish one as the default (or primary) one that gets used.

ErikEJ:
>> PRIMARY KEY is CLUSTERED by default; UNIQUE constraint is NONCLUSTERED by default.
- that is SQL Server specific implementation details -

bricelam:
@ErikEJ indeed. Also, the second point doesn't apply to SQLite. 😉 Just illustrating that they have different attributes based on what type of constraint they are.

bricelam:
This PR is ready for a full review.

src/EFCore.Design/Scaffolding/Internal/CSharpDbContextGenerator.cs
```diff
@@ -420,7 +420,8 @@ private void GenerateIndex(IIndex index)
 
             var annotations = index.GetAnnotations().ToList();
 
-            if (!string.IsNullOrEmpty(index.Relational().Name))
+            if (!string.IsNullOrEmpty(index.Relational().Name)
+                && index.Relational().Name != ConstraintNamer.GetDefaultName(index))
```
smitpatel: Minor: Should we swallow this in scaffoldingmodelfactory?

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
```diff
@@ -162,7 +164,7 @@ protected virtual ModelBuilder VisitSequences([NotNull] ModelBuilder modelBuilde
             return modelBuilder;
         }
 
-        protected virtual SequenceBuilder VisitSequence([NotNull] ModelBuilder modelBuilder, [NotNull] SequenceModel sequence)
+        protected virtual SequenceBuilder VisitSequence([NotNull] ModelBuilder modelBuilder, [NotNull] DatabaseSequence sequence)
```
smitpatel: Can we align parameter name with their types?

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
```diff
@@ -174,40 +176,40 @@ protected virtual SequenceBuilder VisitSequence([NotNull] ModelBuilder modelBuil
...
 
             if (sequenceType != null
                 && !Sequence.SupportedTypes.Contains(sequenceType))
             {
-                Logger.SequenceTypeNotSupportedWarning(sequence.Name, sequence.DataType);
+                Logger.SequenceTypeNotSupportedWarning(sequence.Name, sequence.StoreType);
                 return null;
             }
 
             var builder = sequenceType != null
```
smitpatel: Also to sequenceBuilder

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
```diff
@@ -334,62 +337,57 @@ protected virtual PropertyBuilder VisitColumn([NotNull] EntityTypeBuilder builde
...
             }
 
-            if (!column.PrimaryKeyOrdinal.HasValue)
+            if (!((column.Table.PrimaryKey?.Columns.Contains(column)) ?? false))
             {
                 property.IsRequired(!column.IsNullable);
             }
 
-            property.Metadata.Scaffolding().ColumnOrdinal = column.Ordinal;
+            property.Metadata.Scaffolding().ColumnOrdinal = column.Table.Columns.IndexOf(column);
```
smitpatel: Where is ColumnOrdinal used later?

smitpatel:
>             var key = dependentEntityType.GetOrAddForeignKey(

foreignKey. Variable/parameter names are not very explanatory in this file from old code. You can change them in separate PR if you want.

---
Refers to: src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs:613 in 6f3c344. [](commit_id = 6f3c344a2d33a1a4c7a9edf59da4d54413a86362, deletion_comment = False)

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
```diff
@@ -701,7 +710,7 @@ protected virtual TypeScaffoldingInfo GetTypeScaffoldingInfo([NotNull] ColumnMod
         }
 
         private static void AssignOnDeleteAction(
-            [NotNull] ForeignKeyModel fkModel, [NotNull] IMutableForeignKey foreignKey)
+            [NotNull] DatabaseForeignKey fkModel, [NotNull] IMutableForeignKey foreignKey)
```
smitpatel: :cry:

src/EFCore.Sqlite.Core/Internal/SqliteLoggerExtensions.cs
```diff
@@ -125,12 +121,11 @@ private static string SequenceConfiguredWarning(EventDefinitionBase definition,
         public static void ForeignKeyColumnFound(
             [NotNull] this IDiagnosticsLogger<DbLoggerCategory.Scaffolding> diagnostics,
             [CanBeNull] string tableName,
-            int id,
+            long id,
```
smitpatel: What is this id for? Why its type changed?

src/EFCore.Sqlite.Core/Scaffolding/Internal/SqliteDatabaseModelFactory.cs
```diff
@@ -85,295 +65,484 @@ public virtual DatabaseModel Create(DbConnection connection, IEnumerable<string>
...
         }
 
-        private void GetTables()
+        private IEnumerable<Table> GetTables(DbConnection connection, IEnumerable<string> tables)
         {
-            using (var command = _connection.CreateCommand())
+            var tableSelectionSet = new TableSelectionSet(tables);
+
+            var command = connection.CreateCommand();
+            command.CommandText = new StringBuilder()
```
smitpatel: Any specific reason to use StringBuilder? Single string block (using $ & @) would be easier to read.

smitpatel:
>     public class ReverseEngineeringConfigurationTests

Classname should be changed 

---
Refers to: test/EFCore.Design.Tests/Scaffolding/Internal/ReverseEngineeringConfigurationTests.cs:17 in 6f3c344. [](commit_id = 6f3c344a2d33a1a4c7a9edf59da4d54413a86362, deletion_comment = False)

src/EFCore.Relational/Scaffolding/Metadata/Column.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace Microsoft.EntityFrameworkCore.Scaffolding.Metadata
+{
+    public class Column : Annotatable
```
AndriySvyryd: We should be consistent and prefix everything with "Database"

src/EFCore.Design/Scaffolding/Metadata/Internal/ColumnExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Scaffolding.Metadata.Internal
+{
+    public static class ColumnExtensions
+    {
+        public static string DisplayName([NotNull] this Column column)
+        {
+            var tablePrefix = column.Table?.DisplayName();
```
AndriySvyryd: In what scenarios would `Table` be null?

smitpatel: `Column` can be created without passing in Table (default ctor) which is different from our Metadata which requires ctor parameter connecting back ref.

AndriySvyryd: Ok, but why do we allow this?

smitpatel: We can disallow it. It is coming from `legacy design`

src/EFCore.Sqlite.Core/Internal/SqliteLoggerExtensions.cs
bricelam: You can't get FK constraint names without parsing DDL. This is an internal I'd for it the native type has always been long.

src/EFCore.Design/Scaffolding/Internal/CSharpDbContextGenerator.cs
bricelam: I think that would make the model wrong. It has a name, it just happens to be the by-convention value and doesn't need to be explicitly configured.

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
bricelam: We order by it when writing the properties.

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
bricelam: lol, not sure how to interpret that comment.

I think we need another round of end-to-end cleanup. But this PR gets the public API in order.

src/EFCore.Sqlite.Core/Scaffolding/Internal/SqliteDatabaseModelFactory.cs
bricelam: Main goal was to not rely on the line endings of the file. I'm good with any format so long as it uses `Environment.NewLine`

src/EFCore.Relational/Scaffolding/Metadata/Column.cs
bricelam: I'm thinking so too. I'll tee up an API Review.

src/EFCore.Design/Scaffolding/Metadata/Internal/ColumnExtensions.cs
bricelam: Yep, I just used the existing code. We need to do another pass of internal cleanup.

src/EFCore.Design/Scaffolding/Internal/CSharpDbContextGenerator.cs
smitpatel: Then it does not align with TypeMapping system we have in place. We annotate those facets only if we need during ScaffoldingModelFactory.
I think it is fine to say that generate `Model` would be same as what runtime `Model` represent. i.e. Things don't need to be specified if they are having same values as default.

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
smitpatel: the parameter name `fkModel`. Things are quite ambiguous in naming. I think we can fix those names later since its Internal

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
bricelam: Yes. I was going to wait for API Review before aligning any of the internal code.

src/EFCore.Relational/Scaffolding/Metadata/Column.cs
bricelam: ('Cause you never know if we'll go with `Store` or `Database`. 😉)

src/EFCore.Design/Scaffolding/Internal/RelationalScaffoldingModelFactory.cs
bricelam: Filed #8853

PR #8812, opened 2017-06-10
Test: Upgrade version of xunit

### LINKED ISSUE #8487: Update to xunit 2.3
Other aspnet repos have upgrade xunit to 2.3 while EF remains on 2.2. Upgrading to 2.3 will be a non-trivial task because this upgrade bundles xunit.analyzers which will fail the build.

Possible solutions:
 - Upgrade to the xunit 2.3 runtime but exclude the analyzer. Currently difficult because of: https://github.com/dotnet/sdk/issues/1212
 - Fix all of the analyzer errors
 - Upgrade to 2.3 but add a ruleset to make xunit.analyzer errors  "Info" or "Hidden" instead: see https://github.com/dotnet/roslyn/blob/master/docs/compilers/Rule%20Set%20Format.md

FYI - here are the warnings and errors triggered by xunit.analyzers:

```
Name            Count   Meaning
----            -----   -------
xUnit1000       10      Test classes must be public
xUnit1003       420     Theory methods must have test data
xUnit1004       220     Test methods should not be skipped
xUnit1005       210     Fact methods should not have test data
xUnit1008       24      Test data attribute should only be used on a Theory
xUnit1010       1148    Type mismatch between InlineData and method parameter
xUnit1013       4848    Public method on a test class is not marked as Fact or Theory
xUnit2000       76      A literal or constant value is given as the 'actual' value in a call to Assert.Equal(expected, actual)
xUnit2002       64      Do not use Assert.NotNull() on value type 'bool'.
xUnit2003       108     Do not use Assert.Equal() to check for null value. (Use Assert.NotNull or Assert.Null)
xUnit2004       188     Do not use Assert.Equal() to check for boolean conditions. (Use Assert.True or Assert.False)
```

By the way, I suspect the total number is actually half as many. The errors/warnings appear once per target framework and tests currently target net46 and netcoreapp2.0

smitpatel:
Fixes errors suggested by analyzer
Suppressed errors which are bugs in analyzer
Enabled few tests which were missing Fact

Resolves #8487 

Pending to file issues on xUnit about bugs in the analyzer. I will file the issue and add them to pragma disable or nowarn property before merging.

### DIFF
--- a/src/EFCore.Relational.Design.Specification.Tests/Migrations/ModelSnapshotTest.cs
+++ b/src/EFCore.Relational.Design.Specification.Tests/Migrations/ModelSnapshotTest.cs
@@ -795,7 +795,7 @@ builder.Entity(""Microsoft.EntityFrameworkCore.Migrations.ModelSnapshotTest+Enti
         b.ToTable(""EntityWithStringProperty"");
     });
 ",
-                o => { Assert.Equal(false, o.GetEntityTypes().First().FindProperty("Name").IsNullable); });
+                o => { Assert.False(o.GetEntityTypes().First().FindProperty("Name").IsNullable); });
         }
 
         [Fact]
@@ -928,7 +928,7 @@ builder.Entity(""Microsoft.EntityFrameworkCore.Migrations.ModelSnapshotTest+Enti
         b.ToTable(""EntityWithTwoProperties"");
     });
 ",
-                o => { Assert.Equal(true, o.GetEntityTypes().First().FindProperty("AlternateId").IsConcurrencyToken); });
+                o => { Assert.True(o.GetEntityTypes().First().FindProperty("AlternateId").IsConcurrencyToken); });
         }
 
         [Fact]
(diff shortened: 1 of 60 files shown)


src/EFCore.Specification.Tests/AsyncQueryTestBase.cs
```diff
@@ -1600,7 +1601,7 @@ var efObjects
                     });
         }
 
-        // TODO: [ConditionalFact] See #153
+        [ConditionalFact]
```
smitpatel: This was disabled by @anpete  in Aug 2014 :trollface:

anpete: We got there! 🍾 🍾 🍾

src/EFCore.Relational.Design.Specification.Tests/Migrations/ModelSnapshotTest.cs
```diff
@@ -795,7 +795,7 @@ public virtual void Property_isNullable_is_stored_in_snapshot()
         b.ToTable(""EntityWithStringProperty"");
     });
 ",
-                o => { Assert.Equal(false, o.GetEntityTypes().First().FindProperty("Name").IsNullable); });
```
anpete: Gross!

src/EFCore.Specification.Tests/ChangeTrackingTestBase.cs
```diff
@@ -18,7 +18,7 @@ public virtual void Entity_reverts_when_state_set_to_unchanged()
             {
                 var customer = context.Customers.First();
 
-                Assert.NotEqual(customer.Phone, "425-882-8080");
```
anpete: Hehe

anpete:
🎉

PR #8829, opened 2017-06-12
Query: Generate Enum constants differently

### LINKED ISSUE #8718: Generating Literal values for Enum type
This came from discussion on PR #8708 
Particularly https://github.com/aspnet/EntityFramework/pull/8708#discussion_r119971659

At present, we have override in generic relational type mapping to generate literal. All it does is to generate cast enum to underlying type. So that we generate underlying type literal for enums.
Though this code is very unintuitive. And it would be used by Query pipeline only. (Update pipeline would use parameters need to investigate what we do with parameters of type enum). We can do this processing in DefaultQuerySqlGenerator.VisitConstant method because enums are mapped to type mapping and that function finds the type mapping for the value to generate literal.

smitpatel:
resolves #8718

### DIFF
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -1420,6 +1420,13 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
             Check.NotNull(constantExpression, nameof(constantExpression));
 
             var value = constantExpression.Value;
+
+            if (constantExpression.Type.UnwrapNullableType().IsEnum)
+            {
+                var underlyingType = constantExpression.Type.UnwrapEnumType();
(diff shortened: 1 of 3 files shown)


src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1420,6 +1420,13 @@ protected override Expression VisitConstant(ConstantExpression constantExpressio
             Check.NotNull(constantExpression, nameof(constantExpression));
 
             var value = constantExpression.Value;
+
+            if (constantExpression.Type.IsEnum)
```
anpete: Need any nullable unwrap here?

smitpatel: Updated

PR #8832, opened 2017-06-13
Test: Print debug output when AssertSql fails

### LINKED ISSUE #8759: Test: Capture full log output for failed tests
This could be particularly helpful for failed/flaky tests as we get access to the query plans and all logs which can pinpoint what went wrong & when.

cc: @anpete 

TestMessageVisitor in xunit allows to intercept ITestFailed message which contains some data. We can find something along those lines to capture the output from tests. (or we can always capture output and clear it during Testpassed.

smitpatel:
Resolves #8759 

This works only for "AssertSql" functions. We cannot output anything unless our logger captures it. And I am not able to find a hook where I can capture failed test and print out the output to ITestOutputHelper.
This will print out whole query plan whenever AssertSql fails.


smitpatel:
Will rework this bit more.

PR #8850, opened 2017-06-13
[TEST] Splitting QueryTestBase into multiple files

maumar:
Splitting QueryTestBase into different files with tests for specific areas:

- Select
- Where
- Functions
- ResultOperators
- Join and GroupJoin

There is no functional change here.


src/EFCore.Specification.Tests/QueryTestBase.Functions.cs
```diff
@@ -0,0 +1,951 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
smitpatel: Move them to single folder `QueryTestBase`

smitpatel:
Also if you like there is `QuerySqliteTest`

src/EFCore.Specification.Tests/QueryTestBase.Functions.cs
anpete: We don't use folders for grouping tests by functionality right now.

anpete:
QuerySqliteTest is small so though.

src/EFCore.Specification.Tests/QueryTestBase.Functions.cs
smitpatel: We should. The list of files inside folder is growing too large.

anpete: I'm OK with that, but then let's just create a folder for all query tests: "Query"

test/EFCore.SqlServer.FunctionalTests/QuerySqlServerTest.cs
```diff
@@ -1,8517 +0,0 @@
...
-using System;
-using System.Linq;
-using Microsoft.EntityFrameworkCore.Internal;
-using Microsoft.EntityFrameworkCore.TestModels.Northwind;
-using Microsoft.EntityFrameworkCore.TestUtilities.Xunit;
-using Xunit;
-using Xunit.Abstractions;
-using Microsoft.EntityFrameworkCore.Utilities;
-
-namespace Microsoft.EntityFrameworkCore
```
maumar: any ideas how to "convince" git that this is a rename?

smitpatel: We don't convince git. Git convinced us that we are its slave :trollface: 
Its fine whatever git says.

test/EFCore.SqlServer.FunctionalTests/Query/AsyncQuerySqlServerTest.cs
```diff
@@ -40,7 +40,11 @@ public AsyncQuerySqlServerTest(NorthwindQuerySqlServerFixture fixture, ITestOutp
                     FROM [Customers] AS [c]
                     WHERE [c].[CustomerID] = N'ALFKI', 
                 shaper: BufferedEntityShaper<Customer>), 
-            selector: (Customer c | CancellationToken ct) => Task<<>f__AnonymousType3<Customer, List<Order>>> _ExecuteAsync(
+            selector: (Customer c | CancellationToken ct) => Task<<>f__AnonymousType",
```
maumar: the only non-rename change - number on the AnonymoutType changed on my box, removed the number from baseline so that future changes won't break this test

smitpatel: I think it is fine to have it there too. The magic number changes whenever anonymous type declaration is moved from one type to another (namespace change also makes different type). If you think the test should verify this log fully then update it.

Issue #8652, opened 2017-05-31
InvalidCastException when casting from one value type to another in a simple select statement


samb0t:
[InvalidCastApp.zip](https://github.com/aspnet/EntityFramework/files/1045495/InvalidCastApp.zip)
When casting a property from a nullable double to a nullable int in a `Select` function, an InvalidCastException occurs. Likewise, the same exception occurs if the types are not nullable.

Attached is a working example of the issue. Just create the database and table first and then update the connection string in `Program.cs`.

The select looks like this:
```               
.Select(o => new DTO() {  
  Limit = o.Limit__c.HasValue ? (int?)o.Limit__c : null,   
  Slots = o.Slots__c.HasValue ? (int?)o.Slots__c : null  
  })  
  .Where(o => cIds.Contains(o.CId))  
  .ToListAsync());  
```
`Limit__c` and `Slots__c` are nullable doubles
The SQL profiler does not show any casting before the exception is thrown.
The same query was working in EF 6.x (Full Framework).
```
Exception message: Specified cast is not valid.
Stack trace:
<error errorId="7f262bc2-3b41-439b-be10-ed2c6e480a0f" application="/LM/W3SVC/2/ROOT" host="" type="System.InvalidCastException" message="Specified cast is not valid." source="Anonymously Hosted DynamicMethods Assembly" detail="System.InvalidCastException: Specified cast is not valid.;  
 at lambda_method(Closure , QueryContext , ValueBuffer );  
 at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.ProjectionShaper.TypedProjectionShaper`3.Shape(QueryContext queryContext, ValueBuffer valueBuffer);  
 at Microsoft.EntityFrameworkCore.Query.AsyncQueryMethodProvider.&lt;&gt;c__DisplayClass3_0`1.&lt;_ShapedQuery&gt;b__0(ValueBuffer vb);  
 at System.Linq.AsyncEnumerable.SelectEnumerableAsyncIterator`2.&lt;MoveNextCore&gt;d__7.MoveNext();--- End of stack trace from previous location where exception was thrown ---;  
 at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task);  
 at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task);  
[...]

PR #8861, opened 2017-06-15
Fix to #8652 - InvalidCastException when casting from one value type to another in a simple select statement

maumar:
Problem was for queries with value types being projected in a select expression when also using convert. The problem was that getValue is typed as object which then was converted to an expected type. However if the value returned by SQL was not exactly the same type, exception would get thrown due to boxing/unboxing.

Fix is to detect when we apply convert on a top level projection, and in this case use explicit cast, so that type returned by SQL was exactly the same as the type that was expected after unboxing.

Also added support for translating Negate expression, which was previously evaluated on the client.

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -52,6 +52,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
         private readonly bool _inProjection;
         private readonly NullCheckRemovalTestingVisitor _nullCheckRemovalTestingVisitor;
 
+        private bool _isTopLevelProjection;
+
         /// <summary>
         ///     Creates a new instance of <see cref="SqlTranslatingExpressionVisitor" />.
         /// </summary>
@@ -79,6 +81,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
             _topLevelPredicate = topLevelPredicate;
             _inProjection = inProjection;
             _nullCheckRemovalTestingVisitor = new NullCheckRemovalTestingVisitor(_queryModelVisitor);
+            _isTopLevelProjection = inProjection;
         }
 
         /// <summary>
@@ -107,7 +110,25 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                 return Visit(translatedExpression);
             }
 
-            return base.Visit(expression);
+            if (expression != null
+                && (expression.NodeType == ExpressionType.Convert
+                    || expression.NodeType == ExpressionType.Negate
+                    || expression.NodeType == ExpressionType.New))
+            {
+                return base.Visit(expression);
+            }
+
+            var isTopLevelProjection = _isTopLevelProjection;
+            _isTopLevelProjection = false;
+
+            try
+            {
+                return base.Visit(expression);
+            }
+            finally
+            {
+                _isTopLevelProjection = isTopLevelProjection;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


test/EFCore.SqlServer.FunctionalTests/Query/QueryNavigationsSqlServerTest.cs
```diff
@@ -380,7 +380,7 @@ public override void Select_count_plus_sum()
 
             AssertSql(
                 @"SELECT (
-    SELECT SUM([od].[Quantity])
+    SELECT SUM(CAST([od].[Quantity] AS int))
```
maumar: this is unfortunate, cast is not needed because SUM takes care of the type discrepancy, but result operators are translated after selector (in a separate step), so when translating selector we don't know that it's wrapped in a result op. We could recognize this when handling result ops, and remove the explicit cast, but it seems like an overkill. Thoughts?

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -107,7 +110,22 @@ public override Expression Visit(Expression expression)
...
-            return base.Visit(expression);
+            if ((expression is UnaryExpression || expression is NewExpression))
+            {
+                return base.Visit(expression);
+            }
+
+            var isTopLevelProjection = _isTopLevelProjection;
+            _isTopLevelProjection = false;
+
+            try
```
smitpatel: Don't use try/finally. Use the usual pattern of storing old value locally running the code and restoring the value.

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -125,70 +143,70 @@ protected override Expression VisitBinary(BinaryExpression expression)
...
-                    var right = Visit(expression.Right);
-
-                    return left != null 
-                            && right != null 
-                            && left.Type != typeof(Expression[]) 
-                            && right.Type != typeof(Expression[])
-                        ? expression.Update(left, expression.Conversion, right)
-                        : null;
-                }
+                    {
```
smitpatel: Check R# settings

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
maumar: we do both

anpete: Finally is convenient because of return

test/EFCore.SqlServer.FunctionalTests/Query/QueryNavigationsSqlServerTest.cs
anpete: Can we remove for all providers?
We could do it as SQL tree optimization.

maumar:

smitpatel:
Test failures on CI.

anpete:
smitbot! :trollface:

maumar:
fixed @smitpatel

maumar:
ping @anpete @smitpatel

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -107,7 +110,22 @@ public override Expression Visit(Expression expression)
                 return Visit(translatedExpression);
             }
 
-            return base.Visit(expression);
+            if ((expression is UnaryExpression || expression is NewExpression))
```
anpete: Check NodeType instead.

smitpatel:
😨

PR #8868, opened 2017-06-15
CLI Tools: Structure JSON of `migrations remove` and `dbcontext scaffold`

### LINKED ISSUE #7263: Design: More structured data
There are a couple operations that could use a more structure to their results:
- RemoveMigration--returns `string[]` but should be `MigrationFiles`
- ScaffoldContext--returns `string[]` but should be something that differentiates between the context and the entity types

There is a potential breaking change here that we'd need to asses since version 2.x of Tools and 1.x of Design (and vice-versa) may need to talk to each other.

This change would also manifest in `dotnet ef` (and `ef`) with the `--json` option.

bricelam:
Resolves #7263

### DIFF
--- a/src/EFCore.Design/Design/OperationExecutor.cs
+++ b/src/EFCore.Design/Design/OperationExecutor.cs
@@ -230,25 +230,17 @@ namespace Microsoft.EntityFrameworkCore.Design
             }
         }
 
-        private IEnumerable<string> RemoveMigrationImpl([CanBeNull] string contextType, bool force)
+        private IDictionary RemoveMigrationImpl([CanBeNull] string contextType, bool force)
         {
             var files = _migrationsOperations.Value
                 .RemoveMigration(contextType, force);
 
-            if (files.MigrationFile != null)
+            return new Hashtable
             {
-                yield return files.MigrationFile;
-            }
-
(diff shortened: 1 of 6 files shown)


src/EFCore.Design/Design/OperationExecutor.cs
```diff
@@ -355,13 +347,11 @@ public ScaffoldContext([NotNull] OperationExecutor executor, [NotNull] object re
                 provider, connectionString, outputDir, dbContextClassName,
```
bricelam: No. We want strongly-typed .NET APIs everywhere possible.

The purpose of `OperationExecutor` is to provide a loosely-typed compatibility layer between the tools and the design-time logic. This allows us to make changes in the design-time logic without breaking compatibility with tools. Ultimately for example, version 3.0 of the tools could work with versions 2.x AND 3.0 of the runtime.

We decided not to maintain compatibility with 1.x in 2.0 because of all the design-time changes, but this shouldn't be the case going forward.

It's particularly important for the PMC commands, since there can only be one version of them loaded per solution. This would enable users to run `Update-Database` etc. on a solution with both EF Core 2.0 and 3.0 projects.

PR #8885, opened 2017-06-16
Query: Allow FromSql to accept parameter of type DbParameter in subquery

### LINKED ISSUE #8721: Query: FromSql with DbParameters throws Must declare the scalar variable/No mapping to a relational type ... SqlParameter
Hi,
The 'FromSql' lost the parameters after 'Contains', 'GroupBy' ...

```
Exception message1: Must declare the scalar variable "@Param1"
Exception message2: No mapping to a relational type can be found for the CLR type 'SqlParameter'
```

### Steps to reproduce
CREATE TABLE [Table1]
(
[ID1]   INT IDENTITY PRIMARY KEY NOT NULL
);
GO
CREATE TABLE [Table2]
(
[ID2]   INT IDENTITY PRIMARY KEY NOT NULL,
[ID1]   INT NOT NULL
);
GO
CREATE FUNCTION [MyFunc](@Param1 int)
RETURNS TABLE AS
RETURN ( SELECT * FROM Table1 WHERE ID1 >= @Param1 );
GO

```c#
var q_sql_func = db.Table1.FromSql("SELECT * FROM [MyFunc] (@Param1)", new SqlParameter("@Param1", 123));

var q_sql_func2 = q_sql_func.Select(r => r.ID1);

//OK
var test1 = q_sql_func2.ToList();

//OK
var test2 = db.Table2.ToList();

//OK
var test3 = db.Table2.Join(q_sql_func, t2 => t2.ID1, t1 => t1.ID1, (t2,t1) => t1).ToList();

//ERROR: Must declare the scalar variable "@Param1"
var test4 = db.Table2.Where(r => q_sql_func2.Contains(r.ID1)).ToList();

//ERROR: No mapping to a relational type can be found for the CLR type 'SqlParameter'
var test5 = db.Table2
   .GroupBy(r => r.ID2)
   .Select(g => g.Key)
   .Where(id => q_sql_func2.Contains(id))
   .ToList();
```

### Further technical details
EF Core version: 2.0.0-preview2-25179
Target framework: net46
Database Provider: SqlServer

Project (Console-net46):
[EFCoreFromSqlParam.zip](https://github.com/aspnet/EntityFramework/files/1049583/EFCoreFromSqlParam.zip)

I'm sorry for my English grammar.
Thanks.

smitpatel:
Resolves #8721 

TODO:
- [x] Doc comments to be updated
- [x] Better tests with variations.

@anpete - Creating PR to get initial feedback on the approach.

### DIFF
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -465,7 +465,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
             Check.NotNull(arguments, nameof(arguments));
             Check.NotNull(parameters, nameof(parameters));
 
-            string [] substitutions = null;
+            string[] substitutions = null;
 
             // ReSharper disable once SwitchStatementMissingSomeCases
             switch (arguments.NodeType)
@@ -476,7 +476,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
 
                     if (parameters.TryGetValue(parameterExpression.Name, out var parameterValue))
                     {
-                        var argumentValues = (object [])parameterValue;
+                        var argumentValues = (object[])parameterValue;
 
                         substitutions = new string[argumentValues.Length];
 
@@ -502,14 +502,31 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
                 case ExpressionType.Constant:
                 {
                     var constantExpression = (ConstantExpression)arguments;
-                    var argumentValues = (object [])constantExpression.Value;
+                    var argumentValues = (object[])constantExpression.Value;
 
                     substitutions = new string[argumentValues.Length];
 
                     for (var i = 0; i < argumentValues.Length; i++)
                     {
                         var value = argumentValues[i];
-                        substitutions[i] = GetTypeMapping(value).GenerateSqlLiteral(value);
+
+                        if (value is DbParameter dbParameter)
+                        {
+                            if (string.IsNullOrEmpty(dbParameter.ParameterName))
+                            {
+                                dbParameter.ParameterName = SqlGenerator.GenerateParameterName(_parameterNameGenerator.GenerateNext());
+                            }
+
+                            substitutions[i] = dbParameter.ParameterName;
+
+                            _relationalCommandBuilder.AddRawParameter(
+                                dbParameter.ParameterName,
+                                dbParameter);
+                        }
+                        else
+                        {
+                            substitutions[i] = GetTypeMapping(value).GenerateSqlLiteral(value);
+                        }
                     }
 
                     break;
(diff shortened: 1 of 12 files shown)


src/EFCore.Relational/Storage/Internal/RelationalCommand.cs
```diff
@@ -401,7 +401,11 @@ public class RelationalCommand : IRelationalCommand
 
                 foreach (var parameter in Parameters)
                 {
-                    if (parameterValues.TryGetValue(parameter.InvariantName, out var parameterValue))
+                    if (parameter is RawRelationalParameter)
+                    {
+                        parameter.AddDbParameter(command, value: null);
```
smitpatel: Since IRelationalParameter.AddDbParameter` has value parameter, passing null in this case.

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -502,13 +502,20 @@ public virtual Expression VisitFromSql(FromSqlExpression fromSqlExpression)
...
+                    var argumentValues = (object[])constantExpression.Value;
 
                     substitutions = new string[argumentValues.Length];
 
                     for (var i = 0; i < argumentValues.Length; i++)
                     {
                         var value = argumentValues[i];
+                        if (value is DbParameter dbParameter)
+                        {
+                            substitutions[i] = dbParameter.ParameterName;
```
anpete: So this is to support {0} syntax with raw parameter?

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -502,13 +502,20 @@ public virtual Expression VisitFromSql(FromSqlExpression fromSqlExpression)
...
+                            _relationalCommandBuilder.AddRawParameter(dbParameter.ParameterName, dbParameter);
+                            continue;
```
anpete: Just use else instead?

smitpatel:
@anpete - Ready for review

Test patterns covered
- No format item with SqlParameter
- Format item with SqlParameter without name (generates parameter name)
- Format item with SqlParameter with name (reuse already given name)
- Format item mixed with SqlParameter without format item
- Format item missing 0 index with SqlParameter providing value

cc: @divega - If you want to check out.

src/EFCore.Relational/Storage/Internal/RawRelationalParameter.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        public string InvariantName { get; }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public void AddDbParameter(DbCommand command, object value)
+        {
+            if (value == null)
```
anpete: This ever false?

src/EFCore.Relational/Storage/Internal/RelationalCommand.cs
anpete: Change to polymorphic call on parameter

smitpatel:
Updated.

PR #8911, opened 2017-06-20
Fix #8843 - Automatic eager loading of owned types.

### LINKED ISSUE #8843: Automatic eager loading of owned types
(previously tracked as part of https://github.com/aspnet/EntityFramework/issues/2953) 

Owned types currently don't get eager loaded automatically, i.e. you need to use `Include()` like with any other entity type. Early on we said we wanted the `Include()` call to be unnecessary and we devised that we would create an annotation in the model that marked navigation properties as "always include" and a convention would annotate the navigations to owned types as such automatically. The annotation would provide a way to opt out, and we could create fluent API for it.

We are closer to be done with 2.0 and we believe the full design does not fit, but there can be smaller versions of it that do.

One of the main problems with not doing this by default is the kind of queries users need to write in order for navigations which are defined in derived types to be included. There is currently no API for Include() that works with navigations defined in derived types, hence you have to split your query, e.g.:

``` C#
var ds =context.Ds.Where(predicate);
var derived = ds.OfType<D1>().Include(d1=>d1.NavA).Concat(ds.OfType<D2>().Include(d2=>d2.NavB)).Concat(…);
```

Also, `Concat()` is evaluated in memory.

anpete:
- Adds INavigation.IsEager - navs can now be marked as eager by-default (core metadata only).
- Adds a core convention that sets IsEager=true for principal-to-dependent navs when FK.IsOwnership=true.
- Updates the query compiler to auto-synthesize Include trees for owned navs on result QSREs.
- Adds a new query test suite for testing queries involving ownership.

### DIFF
--- a/src/EFCore/Metadata/Conventions/Internal/CoreConventionSetBuilder.cs
+++ b/src/EFCore/Metadata/Conventions/Internal/CoreConventionSetBuilder.cs
@@ -110,6 +110,8 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
             conventionSet.ForeignKeyUniquenessChangedConventions.Add(foreignKeyPropertyDiscoveryConvention);
             conventionSet.ForeignKeyUniquenessChangedConventions.Add(foreignKeyIndexConvention);
 
+            conventionSet.ForeignKeyOwnershipChangedConventions.Add(new NavigationEagerLoadingConvention());
+
             conventionSet.ModelBuiltConventions.Add(new ModelCleanupConvention());
             conventionSet.ModelBuiltConventions.Add(keyAttributeConvention);
             conventionSet.ModelBuiltConventions.Add(new IgnoredMembersValidationConvention());
@@ -144,7 +146,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
             conventionSet.PropertyFieldChangedConventions.Add(maxLengthAttributeConvention);
             conventionSet.PropertyFieldChangedConventions.Add(stringLengthAttributeConvention);
             conventionSet.PropertyFieldChangedConventions.Add(timestampAttributeConvention);
-
+            
             return conventionSet;
         }
     }
--- a/src/EFCore/Metadata/Internal/InternalEntityTypeBuilder.cs
+++ b/src/EFCore/Metadata/Internal/InternalEntityTypeBuilder.cs
@@ -1729,12 +1729,15 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
                 var existingNavigation = Metadata
                     .FindNavigationsInHierarchy(navigation.Name)
                     .SingleOrDefault(n => n.GetTargetType().Name == targetEntityType.Name && n.GetTargetType().HasDefiningNavigation());
+                
                 var builder = existingNavigation?.ForeignKey.Builder;
+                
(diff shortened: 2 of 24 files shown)


src/EFCore/Metadata/Conventions/Internal/CoreConventionSetBuilder.cs
```diff
@@ -145,6 +145,8 @@ public virtual ConventionSet CreateConventionSet()
             conventionSet.PropertyFieldChangedConventions.Add(stringLengthAttributeConvention);
             conventionSet.PropertyFieldChangedConventions.Add(timestampAttributeConvention);
 
+            conventionSet.ForeignKeyOwnershipChangedConventions.Add(new NavigationEagerLoadingConvention());
```
AndriySvyryd: Move this up, closer to the other FK conventions

src/EFCore/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -1729,12 +1729,17 @@ private static IndexBuilderSnapshot DetachIndex(Index indexToDetach)
...
                 var builder = existingNavigation?.ForeignKey.Builder;
+                
                 if (builder != null)
                 {
                     builder = builder.RelatedEntityTypes(Metadata, existingNavigation.GetTargetType(), configurationSource);
                     builder = builder?.IsOwnership(true, configurationSource);
                     builder = builder?.Navigations(inverse, navigation, configurationSource);
+                    
+                    
+
```
AndriySvyryd: This is overly spacious 😄

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -306,6 +309,86 @@ protected virtual void ExtractQueryAnnotations([NotNull] QueryModel queryModel)
...
+                IQuerySourceTracingExpressionVisitorFactory querySourceTracingExpressionVisitorFactory)
+            {
+                _queryCompilationContext = queryCompilationContext;
+
+                _querySourceTracingExpressionVisitor = querySourceTracingExpressionVisitorFactory.Create();
+            }
+
+            public override void VisitMainFromClause(MainFromClause fromClause, QueryModel queryModel)
+            {
+                ApplyIncludesForOwnedNavigations(new QuerySourceReferenceExpression(fromClause), queryModel);
```
maumar: nit: use IQuerySource here and only construct QSRE when needed? (before passing to IncludeResulOperator)

test/EFCore.InMemory.FunctionalTests/Query/OwnedQueryInMemoryTest.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.EntityFrameworkCore.Query
+{
+    public class OwnedQueryInMemoryTest : OwnedQueryTestBase, IClassFixture<OwnedQueryInMemoryFixture>
+    {
+        private readonly OwnedQueryInMemoryFixture _fixture;
+
+        public OwnedQueryInMemoryTest(OwnedQueryInMemoryFixture fixture, ITestOutputHelper testOutputHelper)
+        {
+            _fixture = fixture;
+            TestLoggerFactory.TestOutputHelper = testOutputHelper;
```
smitpatel: Need to comment this out?

PR #8912, opened 2017-06-20
Always use varchar(max) or nvarchar(max) for SQL Server string command parameters

### LINKED ISSUE #5975: EF Core 1.0 RTM not filtering properly on fields of type char
### Steps to reproduce

I have created a Github project for this issue
https://github.com/iberodev/EntityFramework5975
### The issue

Suppose I have a Port entity

```
public class Port
{
    public string Code { get; set; }
}

public class PortMap
{
    public PortMap(EntityTypeBuilder<Port> entityBuilder)
    {
        entityBuilder.HasKey(p => p.Code);

        entityBuilder.Property(p => p.Code)
            .HasColumnType("char(5)");

        entityBuilder.Property(p => p.Name)
            .HasMaxLength(65);
    }
}

```

and the field Code has a `char(5)` type.

and the following data:
Code    Name
AUCNB   Canberra
AUMLB   Melbourne
AUSYD   Sydney
NZAKL   Auckland
NZTRG   Tauranga

if I use Entity to retrieve rows based on a filter (eg: the word `NZ` to retrieve only ports whose code starts with `NZ`), the SQL query Entity generates does not return any results from database.

```
public IEnumerable<Port> GetPortsFilteredInDatabase(string filter = "NZ")
{
    var query = _context.Ports.Where(p => p.Code.StartsWith(filter));
    return query.ToList(); //this list will be empty
}

public IEnumerable<Port> GetPortsFilteredInMemory(string filter = "NZ")
{
    var allPorts = _context.Ports.ToList();
    var query = allPorts.Where(p => p.Code.StartsWith(filter));
    return query.ToList(); //this list is correct
}
```

> Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommandBuilderFactory:Information: Executed DbCommand (0ms) [Parameters=[@__filter_0='?' (Size = 5)], CommandType='Text', CommandTimeout='30']
> SELECT [p].[Code], [p].[Name]
> FROM [Ports] AS [p]
> WHERE [p].[Code] LIKE @__filter_0 + '%'

Notice the question mark instead the `NZ` text when generating the SQL statement. Is that normal?

This was working fine in EF Core RC2.

Also notice that if instead CHAR(5), my code column is NVARCHAR(5) everything works well.
### Further technical details

EF Core version: 1.0.0
Operating system: Windows 10

AndriySvyryd:
Fixes #5975

### DIFF
--- a/src/EFCore.Relational/Storage/Internal/RelationalParameterBuilder.cs
+++ b/src/EFCore.Relational/Storage/Internal/RelationalParameterBuilder.cs
@@ -65,7 +65,7 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
                 new TypeMappedRelationalParameter(
                     invariantName,
                     name,
-                    typeMapping,
+                    GetParameterTypeMapping(typeMapping),
                     nullable));
         }
 
@@ -83,7 +83,7 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
                 new TypeMappedRelationalParameter(
                     invariantName,
                     name,
-                    TypeMapper.GetMapping(property),
+                    GetParameterTypeMapping(TypeMapper.GetMapping(property)),
                     property.IsNullable));
         }
 
@@ -96,18 +96,25 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
             Check.NotEmpty(invariantName, nameof(invariantName));
             Check.NotNull(buildAction, nameof(buildAction));
 
-            var innerList = new RelationalParameterBuilder(TypeMapper);
+            var parameterBuilder = Create();
 
-            buildAction(innerList);
+            buildAction(parameterBuilder);
 
-            if (innerList.Parameters.Count > 0)
+            if (parameterBuilder.Parameters.Count > 0)
             {
                 _parameters.Add(
                     new CompositeRelationalParameter(
                         invariantName,
-                        innerList.Parameters));
+                        parameterBuilder.Parameters));
             }
         }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected virtual RelationalParameterBuilder Create()
+            => new RelationalParameterBuilder(TypeMapper);
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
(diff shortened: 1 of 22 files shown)


src/EFCore.Relational/Storage/Internal/RelationalParameterBuilder.cs
```diff
@@ -96,7 +96,7 @@ public virtual void AddCompositeParameter(string invariantName, Action<IRelation
             Check.NotEmpty(invariantName, nameof(invariantName));
             Check.NotNull(buildAction, nameof(buildAction));
 
-            var innerList = new RelationalParameterBuilder(TypeMapper);
+            var innerList = Create();
```
smitpatel: Since we are here, change variable name to something better?

AndriySvyryd: `providerSpecificName`?

smitpatel:
While the title and referencing issue says to use `nvarchar(max)` why all the test changes are converting to `nvarchar(4000)`. In certain cases, its fine since it would be key. Is it case for all though?

src/EFCore.Relational/Storage/Internal/RelationalParameterBuilder.cs
smitpatel: As simple as `var parameterBuilder = CreateParameterBuilder()` would be fine too. (or add relational word too)

AndriySvyryd:
@smitpatel It seems that SQL Client uses varchar(4000) or nvarchar(4000) if the data fits

smitpatel:
Can we add a test for `nvarchar(max)` too?

AndriySvyryd:
Updated

test/EFCore.SqlServer.FunctionalTests/BuiltInDataTypesSqlServerTest.cs
```diff
@@ -343,7 +344,7 @@ public virtual void Can_insert_and_read_back_all_mapped_data_types()
 @p17='anyone!' (Nullable = false) (Size = 4000)
 @p18='Gumball Rules OK!' (Nullable = false) (Size = 4000)
 @p19='103.3'
-@p20='don't' (Nullable = false) (Size = 4000)
+@p20='" + entity.NvarcharMax + @"' (Nullable = false) (Size = -1)
```
smitpatel: Ideally we are not printing out facets which are default. So size should not be printed for `nvarchar(max)`. File a new issue to check if we have a bug.

AndriySvyryd: The default is 0,  and we correctly don't print it if the value is null or empty

smitpatel: err... 0 sized string is default. Ok.

ErikEJ:
providers fyi/beware?

PR #8919, opened 2017-06-21
Validate keys are not configured as AddOrUpdate/Computed

### LINKED ISSUE #8863: Validate keys are not configured as AddOrUpdate/Computed
Keys cannot change once the entity has been saved. This means any configuration designed to update the value after the entity has been saved is incompatible with the definition of the key. We should add a rule to model validation to check this.

AndriySvyryd:
Fixes #8863

### DIFF
--- a/src/EFCore/Infrastructure/ModelValidator.cs
+++ b/src/EFCore/Infrastructure/ModelValidator.cs
@@ -46,6 +46,7 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
             ValidateNoShadowEntities(model);
             ValidateNonNullPrimaryKeys(model);
             ValidateNoShadowKeys(model);
+            ValidateNoMutableKeys(model);
             ValidateClrInheritance(model);
             ValidateChangeTrackingStrategy(model);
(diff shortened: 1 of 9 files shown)


src/EFCore/Infrastructure/ModelValidator.cs
```diff
@@ -188,6 +189,27 @@ protected virtual void ValidateNoShadowKeys([NotNull] IModel model)
...
         /// </summary>
+        protected virtual void ValidateNoMutableKeys([NotNull] IModel model)
+        {
+            Check.NotNull(model, nameof(model));
+
+            foreach (var entityType in model.GetEntityTypes())
+            {
+                foreach (var key in entityType.GetDeclaredKeys())
+                {
+                    var mutableProperty = key.Properties.FirstOrDefault(p => p.ValueGenerated == ValueGenerated.OnAddOrUpdate);
```
smitpatel: There is also `ValueGenerated.OnUpdate` in the enum now. Should we check that also if there is a mechanism to assign that value using core metadata?

PR #8931, opened 2017-06-21
Increase the range of assertion in Concurrency_test()

smitpatel:
@anpete

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,12 @@ os:
   - linux
 #  - osx
 osx_image: xcode8.1
+addons:
+  apt:
+    sources:
+      - debian-sid
+    packages:
+      - libunwind8
 branches:
   only:
     - master
--- a/test/EFCore.SqlServer.FunctionalTests/DbContextPoolingTest.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/DbContextPoolingTest.cs
@@ -349,7 +349,7 @@ namespace Microsoft.EntityFrameworkCore
 
             await Task.WhenAll(tasks);
 
-            Assert.InRange(PooledContext.InstanceCount, 30, 50);
+            Assert.InRange(PooledContext.InstanceCount, 30, 500);
         }
 
         private readonly TimeSpan _duration = TimeSpan.FromSeconds(10);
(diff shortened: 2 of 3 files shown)


test/EFCore.SqlServer.FunctionalTests/DbContextPoolingTest.cs
```diff
@@ -349,7 +349,7 @@ public void Provider_services_are_reset()
 
             await Task.WhenAll(tasks);
 
-            Assert.InRange(PooledContext.InstanceCount, 30, 50);
+            Assert.InRange(PooledContext.InstanceCount, 50, 500);
```
anpete: Keep lower bound

smitpatel: Done

.travis.yml
```diff
@@ -10,6 +10,12 @@ os:
   - linux
 #  - osx
 osx_image: xcode8.1
+addons:
+  apt:
+    sources:
```
mikeharder: Why is the `sources` element needed?  I think just `packages\libunwind8` should be sufficient.

smitpatel: Unable to find package with just that.

mikeharder: Yeah, I found this issue: https://github.com/travis-ci/travis-ci/issues/4780.  CC: @natemcmaster

natemcmaster: libunwind8 shouldn't be required for .NET Core 2.0 IIRC

natemcmaster: Try adding this instead:
```yml
group: deprecated-2017Q2
```

Travis changed the trusty images today: https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch

smitpatel: Saw that. Is there a point of going backwards?
> We are planning to keep the previous generation of images around until September 1st, 2017.

mikeharder: @natemcmaster: Docker deps still installs libunwind8 so I assume it's required: https://github.com/dotnet/dotnet-docker-nightly/blob/master/2.0/runtime-deps/stretch/Dockerfile#L16.  Also the CLI install scripts block if libunwind not installed: https://github.com/dotnet/cli/blob/release/2.0.0-preview2/scripts/obtain/dotnet-install.sh#L194.

natemcmaster: You're write. Looks like these are still required on debian.

image

PR #8945, opened 2017-06-22
API review changes

ajcvickers:
* Rename IsEager to IsEagerLoaded
* Move enums that are used in fluent API to root namespace


smitpatel:
Weird that AppVeyor fails on API check but travis passes. Our build scripts seem to be inconsistent.
cc: @ryanbrandenburg , @natemcmaster

PR #8947, opened 2017-06-22
Remove obsolete warnings about obsolete logger APIs

### LINKED ISSUE #8926: Remove pragmas for non-obsolete ILogger.AddProvider
Since it is no-longer obsolete for 2.0. See https://github.com/aspnet/Logging/issues/616

smitpatel:
Resolves #8926

### DIFF
--- a/src/EFCore.Relational.Specification.Tests/TestSqlLoggerFactory.cs
+++ b/src/EFCore.Relational.Specification.Tests/TestSqlLoggerFactory.cs
@@ -12,8 +12,6 @@ using Microsoft.Extensions.Logging;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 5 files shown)


src/EFCore.Relational.Specification.Tests/TestSqlLoggerFactory.cs
```diff
@@ -12,8 +12,6 @@
 using Xunit;
 using Xunit.Abstractions;
 
-#pragma warning disable 618
```
ajcvickers: No restore for this one?

smitpatel: Blanket disable in whole file.

PR #8950, opened 2017-06-22
RevEng: Scaffold ForSqlServerIsClustered correctly for PK/Index

### LINKED ISSUE #8687: SQL Server RevEng: No ForSqlServerIsClustered(false) on NONCLUSTERED PKs
At present we add Clustered annotation on index when its clustered.
Points to remember,
1. By default PK is clustered but we never scaffold index for it explicitly.
2. There can be only 1 clustered index. Which means that we may need to set clustering false on PK if there is another index.
3. Annotation should be set only when we need to specify fluent API else it should remain null.

This will make processing of clustered index accurate.

smitpatel:
Resolves #8687 

Rules followed 
- If PK is marked as "CLUSTERED" don't do anything. That is default behavior for SqlServer.
- If some other index/unique constraint is marked as clustered then output `ForSqlServerIsClustered()` for index. PK will be non clustered by default.
- If PK is marked as "NONCLUSTERED" and there is no other Clustered index then we explicitly output `ForSqlServerIsClustered(false)` on PK.

### DIFF
--- a/src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
+++ b/src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
@@ -566,9 +566,9 @@ ORDER BY object_schema_name(i.object_id), object_name(i.object_id), i.name, ic.k
                             Name = indexName
                         };
 
-                        if (typeDesc == "CLUSTERED")
+                        if (typeDesc == "NONCLUSTERED")
                         {
-                            primaryKey[SqlServerAnnotationNames.Clustered] = true;
+                            primaryKey[SqlServerAnnotationNames.Clustered] = false;
                         }
 
                         Debug.Assert(table.PrimaryKey == null);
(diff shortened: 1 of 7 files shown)


src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
```diff
@@ -662,6 +662,7 @@ private void GetUniqueConstraints()
                         if (typeDesc == "CLUSTERED")
                         {
                             uniqueConstraint[SqlServerAnnotationNames.Clustered] = true;
+                            table.PrimaryKey?.RemoveAnnotation(SqlServerAnnotationNames.Clustered);
```
bricelam: Is this right? If seems like you'd be removing `Clustered = false` on the primary key. This would default it to `CLUSTERED` causing more than one clustered index to be created leading to an error in Migrations.

bricelam: (Discussed offline.) Looks good.

smitpatel: This speculation was correct. It indeed causes issues while recreating using migrations. filed https://github.com/aspnet/EntityFramework/issues/9324

PR #8951, opened 2017-06-23
RevEng: Remove DbDataReaderExtension

### LINKED ISSUE #8712: RevEng: Remove DbDataReaderExtension
This class is used by the SQL Server and SQLite providers.

smitpatel:
Resolves #8712


src/EFCore.SqlServer/Scaffolding/Internal/SqlDataReaderExtension.cs
```diff
@@ -17,13 +16,12 @@ public static class SqlDataReaderExtension
...
         /// </summary>
-        public static T GetValueOrDefault<T>([NotNull] this SqlDataReader reader, [NotNull] string name)
+        public static T GetValueOrDefault<T>([NotNull] this DbDataReader reader, [NotNull] string name)
         {
             var idx = reader.GetOrdinal(name);
             return reader.IsDBNull(idx)
                 ? default(T)
-                // : reader.GetFieldValue<T>(idx);
-                : (T)reader.GetValue(idx);
+                : reader.GetFieldValue<T>(idx);
```
smitpatel: @bricelam - Why do we need this extension method? SQLite calls `GetString` etc method directly. Is there anything different in SqlServer which requires this?

bricelam: IF the data might contain `NULL` values, you need it. Otherwise, `GetString` would try to cast `DBNull` to `string` resulting in an invalid cast exception.

bricelam: I double-checked all the places in SQLite, and there was only one column that was nullable. But I just decided to inline the logic:
```C#
var value = !reader.IsDbNull(0)
    ? reader.GetString(0)
    : null;
```

Issue #8942, opened 2017-06-22
API Review Updates for HasDbFunction, etc.


ajcvickers:
- [x] HasDbFunction
  - [x] Investigate adding HasDbFunction that takes an expression instead of a MethodInfo
  - [x] Remove name/schema from HasDbFunction in favor of leaving HasName and HasSchema on builder
- [x] DbFunctionAttribute
  - [x] Unseal
  - [x] Move to root namespace
  - [x] Change NotNull to NotEmpty on Name property
  - [x] Make `schema` optional in constructor
  - [x] Rename Name to FunctionName
- [x] DbFunctionBuilder
  - [x] Move to Builders namespace
  - [x] Remove HasParameter
  - [x] Investigate if ReturnType is needed or if it is the same as the CLR type of the method
  - [x] Rename TranslateWith to HasTranslation
  - [x] Change SqlFunctionExpression to Expression
- [x] Remove DbFunctionParameterAttribute, DbFunctionParameterBuilder, IDbFunctionParameter, IMutableDbFunctionParameter
- [x] IDbFunction/IDbMutableFunction
  - [x] Rename Name to FunctionName
  - [x] Remove Parameters
  - [x] Remove AddParameter
  - [x] Change SqlFunctionExpression to Expression
  - [x] Rename TranslateCallback to Translation

pmiddleton:
I can make most of these changes tonight. 

 DbFunctionBuilder.HasParameter and DbFunctionBuilder.ReturnType were added at the request of @AndriySvyryd.  They are needed if/when the ability to remove the default conventions is added.  HasParameter was originally GetParameter which allowed a fluent way to find a dbFunctionParameter.  

For DbFunctionParameter the only property which the users need to be able to set is Index.  It is possible to be in a situation where the database function and the .Net method parameters are in different orders and this property allows parameters to be mapped correctly.  If we drop HasParameter and DbFunctionParameterBuilder then the fluent api is lost and the user will have to fish around in the DbFunction.Parameters collection using linq.

Alternatively we can just say tough luck and if the parameter orders doesn't match up then just write a custom translation and reorder the parameters yourself.  

DbFunctionParameterAttribute was a nice to have so you could reorder parameters without needing to use the fluent api - no functionality would be lost by removing it.

anpete:
Thanks @pmiddleton!

> Alternatively we can just say tough luck and if the parameter orders doesn't match up then just write a custom translation and reorder the parameters yourself.

[...]

PR #8953, opened 2017-06-23
WIP #8942 - Refactor DbFunctions public api

pmiddleton:
Update DBFunctions public api based on #8942


pmiddleton:
Everything should be done except

* Investigate adding HasDbFunction that takes an expression instead of a MethodInfo

src/EFCore.Relational/Infrastructure/RelationalModelValidator.cs
```diff
@@ -67,11 +67,11 @@ public override void Validate(IModel model)
         /// </summary>
         protected virtual void ValidateDbFunctions([NotNull] IModel model)
         {
-            foreach (var dbFunction in model.Relational().DbFunctions)
+            foreach (DbFunction dbFunction in model.Relational().DbFunctions)
```
anpete: var

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
```diff
@@ -36,15 +36,14 @@ public virtual Annotation Apply(InternalModelBuilder modelBuilder, string name,
...
-                                        DbFuncParamAttr = pi.GetCustomAttributes<DbFunctionParameterAttribute>().SingleOrDefault(),
                                         pi.ParameterType
                                     });
 
                 foreach (var p in parameters)
                 {
                     var paramBuilder = dbFunctionBuilder.HasParameter(p.ParameterInfo.Name, ConfigurationSource.Convention);
 
-                    paramBuilder.HasIndex(p.DbFuncParamAttr?.ParameterIndex ?? p.ParameterIndex, ConfigurationSource.Convention);
+                    paramBuilder.HasIndex(p.ParameterIndex, ConfigurationSource.Convention);
```
anpete: Remove metadata/builders relating to parameters.

src/EFCore.Relational/Metadata/DbFunctionAttribute.cs
```diff
@@ -5,15 +5,15 @@
...
     /// </summary>
     [AttributeUsage(AttributeTargets.Method)]
-    public sealed class DbFunctionAttribute : Attribute
+    public class DbFunctionAttribute : Attribute
     {
-        private string _name;
+        private string _functionName;
 
         /// <summary>
         /// Defines a user defined database function
```
anpete: Add a note about the convention for function name when using the default .ctor.

test/EFCore.Relational.Tests/Metadata/DbFunctionMetadataTests.cs
```diff
@@ -80,9 +80,9 @@ public void Adding_method_fluent_only_convention_defaults()
             var modelBuilder = GetModelBuilder();
 
             var dbFuncBuilder = modelBuilder.HasDbFunction(MethodAmi);
-            var dbFunc = dbFuncBuilder.Metadata;
+            var dbFunc = dbFuncBuilder.Metadata as DbFunction;
```
anpete: Why these casts?

src/EFCore.Relational/Metadata/Internal/InternalDbFunctionBuilder.cs
```diff
@@ -87,9 +72,9 @@ private DbFunctionParameter GetOrCreateParameter(string name, ConfigurationSourc
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used 
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public virtual InternalDbFunctionBuilder TranslateWith([NotNull] Func<IReadOnlyCollection<Expression>, IDbFunction, SqlFunctionExpression> translateCallback)
+        public virtual InternalDbFunctionBuilder HasTranslation([NotNull] Func<IReadOnlyCollection<Expression>, IDbFunction, Expression> translation)
```
anpete: CanBeNull or add Check.NotNull

test/EFCore.Relational.Tests/Metadata/DbFunctionMetadataTests.cs
pmiddleton: The unit tests look at the parameters collection to make sure the convention ran correctly.  One of the changes was to remove the parameters collection from IDBFuction, so I need to either cast or use reflection so I can make sure everthing was built correctly.

The other options is to add the parameters collection back onto IDBFuction.

src/EFCore.Relational/Infrastructure/RelationalModelValidator.cs
pmiddleton: see my note below on why I had to do this.

anpete:
Looking good. I turned #8942 into a checklist and ticked what I think we have so far.

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
pmiddleton: Ok - will switch over to direct constructor calls and remove the InternalParameterBuilder

pmiddleton:
I can't make these changes until tonight - working today.

test/EFCore.Relational.Tests/Metadata/DbFunctionMetadataTests.cs
pmiddleton: I just realized there is a bigger issue with this.  DBFunciton.Translation passes an IDbFunction to the callback.  The end user won't have access to the parameters.  We either need to change the callback signature to pass DbFunction, or add the parameters collection back to IDbFunction.

Which way do you guys want to go.

anpete: We want to remove all trace of parameters including DbFunctionParameter 😄

src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
smitpatel: We actually don't need to do any processing of parameters at all. DbFunction will store methodInfo/FunctionName/Schema. The Translate will do any processing of parameters if needed else, we will just forward the args of MethodCallExpression as is.

src/EFCore.Relational/Metadata/DbFunctionAttribute.cs
```diff
@@ -5,15 +5,15 @@
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Utilities;
 
-namespace Microsoft.EntityFrameworkCore.Metadata
+namespace Microsoft.EntityFrameworkCore
```
smitpatel: Does this file move to root folder? @ajcvickers

src/EFCore.Relational/Metadata/Internal/DbFunctionParameter.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.EntityFrameworkCore.Metadata.Internal
 {
-    public class DbFunctionParameter : IMutableDbFunctionParameter
+    public class DbFunctionParameter
```
smitpatel: We don't need this class

src/EFCore.Relational/Metadata/DbFunctionAttribute.cs
ajcvickers: It should, yes.

test/EFCore.Relational.Tests/Metadata/DbFunctionMetadataTests.cs
smitpatel: Does `Translate` needs access to `IDbFunction`?
Actually `Translate` can be `Func<IReadOnlyCollection<Expression>, Expression>`
When user configures it, they know which DbFunction it is already they just need to utilize the args and construct the translated expression. The only usage DbFunction had in test was to use return type which is also removed.

pmiddleton: @smitpatel - It does have the name and schema which could be useful, but to your point if they are doing things manually they should know what they want for these values.

smitpatel: Indeed. So we should simplify it further.
Also if user is providing manual translation then name & schema should be absent. We are not strictly enforcing this but they won't be used anyway.

pmiddleton:
Updates are done.  DbFunctionParameter is gone.  :yum:

src/EFCore.Relational/DbFunctionAttribute.cs
```diff
@@ -5,18 +5,18 @@
...
     [AttributeUsage(AttributeTargets.Method)]
-    public sealed class DbFunctionAttribute : Attribute
+    public class DbFunctionAttribute : Attribute
     {
-        private string _name;
+        private string _functionName;
 
         /// <summary>
-        /// Defines a user defined database function
+        /// Defines a user defined database function.  By convention uses the .Net method name as name of the database function and the default schema.
```
smitpatel: nit: `.NET`
or @divega would know lol

pmiddleton: Capitalized.  I'm open to a better sentence if someone has one.

src/EFCore.Relational/Metadata/IDbFunction.cs
```diff
@@ -41,8 +34,8 @@ public interface IDbFunction
         Type ReturnType { get; }
 
         /// <summary>
-        ///    A translate callback for converting a method call into a sql function
+        ///    A method for converting a method call into a sql function
```
smitpatel: drop function at the end.
`A method for converting a method call into sql.
(of course user would need to understand how to generate "sql"

test/EFCore.SqlServer.FunctionalTests/Query/NorthwindDbFunctionSqlServerFixture.cs
```diff
@@ -39,7 +39,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             base.OnModelCreating(modelBuilder);
 
             modelBuilder.HasDbFunction(typeof(NorthwindDbFunctionContext).GetRuntimeMethod(nameof(NorthwindDbFunctionContext.MyCustomLength), new[] { typeof(string) }))
-                .TranslateWith((args, dbFunc) => new SqlFunctionExpression("len", dbFunc.ReturnType, args));
+                .HasTranslation((args, dbFunc) => new SqlFunctionExpression("len", dbFunc.ReturnType, args));
```
smitpatel: @anpete - Do we want the lambda to have `dbFunction`?

The only thing it is being used is for return type which user is aware of while calling the method. The other properties on DbFunction is name & schema which is also user known. If we remove DbFunction from here then we can make return type private field instead of exposing as public property.

anpete: @smitpatel So we need ReturnType?

smitpatel: We need return type when we invoke "SqlFunctionExpression"
For internal purposes, we can store it in field in `DbFunction`
For user fluent API, user can write explicitly without depending on DbFunction.
I am not seeing enough value to provide user access to `DbFunction` in the lambda expression because it will be at the same place where user is adding DbFunction.

pmiddleton: I think we can remove return type and use the return type on methodinfo to check for a supported return type.

pmiddleton: Yes the only item the user could use from dbfunction is name and schema.  They can just capture those in the lambda closure they pass into translate.

anpete: @smitpatel We wanted to try and do what @pmiddleton said about ReturnType :smile: Only question was whether or not it can be different from CLR method return type.

pmiddleton: The only place where I could see the return type being different is if someone wanted to expand or truncate a datatype on purpose.  

Say the database function returns int and you try to map it to a long, or vise versa.  What will the materializer/shaper do if the returned datatype is convertible, but not an exact match?  

I can see the end user getting into a lot of trouble with this ability.  They may try making the .NET method return one type and the database function return another and expect it to automatically convert.  Like returning a string and trying have it automatically parse to an int.

I say just force the return types to match and then use the methodInfo return type.

anpete: Sounds good. We are going to handle the type conversion issue as a separate feature anyway, so this will just be another place where it may apply.

pmiddleton: Plus if we need to add it back in at a later point that is easy to do, but once it is in the wild we can't as easily pull it back.

pmiddleton: I have pushed updated code with the return type removed.

src/EFCore.Relational/Metadata/IDbFunction.cs
```diff
@@ -23,26 +21,16 @@ public interface IDbFunction
...
+        string FunctionName { get; }
 
         /// <summary>
-        ///     The list of parameters which are passed to the underlying datastores function.
-        /// </summary>
-        IReadOnlyList<DbFunctionParameter> Parameters { get; }
-
-        /// <summary>
-        ///     The .Net method which maps to the function in the underlying datastore
+        ///     The .Net method which maps to the function in the underlying datastore.
```
divega: Another instance of ".Net" here. I would consider removing it so it is just "the method". Also, for "datastore", prefer "database."

src/EFCore.Relational/Metadata/Internal/DbFunction.cs
```diff
@@ -141,84 +120,21 @@ private void UpdateNameConfigurationSource(ConfigurationSource configurationSour
...
-        public virtual IReadOnlyList<DbFunctionParameter> Parameters => _parameters.Values.ToList();
-
-        /// <summary>
         ///     The .Net method which maps to the function in the underlying datastore
         /// </summary>
         public virtual MethodInfo MethodInfo { get; }
 
         /// <summary>
-        ///     If set this callback is used to translate the .Net method call to a Linq Expression.
+        ///     A method for converting a method call into sql.
```
divega: Consider removing ".Net" or make it uppercase.

divega: It is actually a few lines above.

src/EFCore.Relational/Metadata/IDbFunction.cs
divega: There are more.

pmiddleton:
I have added an overload for HasDbFunction which accepts a `Func<T>`.  This will allow you to define a method without needing a MethodInfo.  You will need to provide a full method call with filler parameters like this.

``` csharp 
modelBuilder.HasDbFunction(() => TestMethods.MethodA(null, default(int)));

anpete:
Merged thanks!

Issue #8465, opened 2017-05-15
InvalidOperationException updating an entity with CurrentValues.SetValues throws InvalidOperationException when the SetValues parameter does not contain a required field data.


alexdrl:
I am writing a custom save generic method for our framework in EF Core. The idea is to have a method that can make partial updates

The method is working with simple entities but throws an InvalidOperationException when the entity in entitiesAndProperties does not contain a required field data. It should not be a problem, because the entity already exists in database with the required field value, contextEntity reflects that, but seems the CurrentValues.SetValues overrides this. 

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
Exception message:
'The property 'Name' on entity type 'User' is marked as null, but this cannot be saved because the property is marked as required.'

Stack trace:
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.HandleConceptualNulls()
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.GetEntriesToSave()
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChanges(Boolean acceptAllChangesOnSuccess)
```

### Steps to reproduce
Those are the custom entity method and clases.

```c#
public IEnumerable<T> UpdateEntitiesByProperties(IEnumerable<UpdateByPropertiesData<T>> entitiesAndProperties)
        {
            List<T> entities = new List<T>();

            // TODO: (Error Management) Decide what to do with null parameters
            if (entitiesAndProperties?.Any() != true)
            {
                return entities;
            }

            foreach (UpdateByPropertiesData<T> entityAndProperties in entitiesAndProperties)
            {
                T contextEntity = this.context.Find<T>(entityAndProperties.Entity.Id);

                EntityEntry<T> changeTrackerEntityEntry = this.context.Entry(contextEntity);

                changeTrackerEntityEntry.CurrentValues.SetValues(entityAndProperties.Entity);

                // TODO: Review the possibility of having a local copy of the entityEntry modified
                // properties, for not setting them to false in this foreach. This will fix the edge
                // case when we call this method two times with the same entity and different data,
                // in which case, data is lost.
                foreach (PropertyEntry property in changeTrackerEntityEntry.Properties.Where(p => !p.Metadata.IsConcurrencyToken))
                {
[...]

Issue #8265, opened 2017-04-24
Attaching entity and setting different IsModified values to different property prevent update


ShaZeh:
I am not quite sure if this is me doing something wrong, I am pretty sure I have been using this technique without any issues for a good while. But it might be possible that I might never have noticed that some fields were not updating. It might also be possible that the update only fail under very specific circonstances.

So far I have only reproduced the issue while having two property explicitly set to false and true on the second.

Also, it seems that by simply having another entity being saved within the same transaction automatically resolved problem, the IsModified was being taken into account properly.

### Steps to reproduce
- Run the code below
- Notice that the username will remain to "name" even though the last property is set to isModified true.

```C#
using System;
using System.Collections.Generic;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.EntityFrameworkCore;
using Microsoft.EntityFrameworkCore.Storage;
using System.Data;

namespace ConsoleApplication
{
    class Program
    {
        static Context context;
        static void Main(string[] args)
        {
            Task.Run(async () => { await UpdateEntity(); }).GetAwaiter().GetResult();
        }

        static async Task UpdateEntity()
        {
            context = new Context();
            context.Database.EnsureCreated();

            var user = new Users();
            user.UserName = "name";
            user.City = "city";
            context.Add(user);
            await context.SaveChangesAsync();

            int id = user.Id;
            context.Entry(user).State = EntityState.Detached;


            user = new Users { Id = id, UserName = "New" };

            context.Attach(user);
            context.Entry(user).Property(x => x.City).IsModified = false;
            context.Entry(user).Property(x => x.UserName).IsModified = true;
            //context.Update(user);

            await context.SaveChangesAsync();

            user = await context.Users.Where(x => x.Id == id).FirstOrDefaultAsync();
            Console.WriteLine(user.UserName);
        }
    }

    public class Users
    {
        public int Id { get; set; }
        public string UserName { get; set; }
        public string City { get; set; }
    }
    public class Context : DbContext
    {
        public virtual DbSet<Users> Users { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Users>(b =>
            {
                b.HasKey(r => r.Id);

                b.Property(x => x.UserName).IsRequired();
            });
        }
        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
[...]

Issue #7798, opened 2017-03-07
Setting the IsModified flag to false on a property leaves original value untouched so the change is detected again


AndersKung:
```
using Microsoft.EntityFrameworkCore;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;

//Microsoft.EntityFrameworkCore.1.1.1

namespace ConsoleApplication
{
    class Program
    {
        static void Main(string[] args)
        {
            var context = new Context();
            var user = new Users();
            user.Id = 1;

            context.Update(user);

            user.UserName = "name";
            user.City = "city";

            OnUpdating(context, user);

            // ↓↓↓Exception↓↓↓
            // The property 'UserName' on entity type 'Users' is marked as null, but this cannot be saved because the property is marked as required.
            context.SaveChanges(); 
            
        } 

        static void OnUpdating(Context context, Users user)
        {
            context.Entry(user).Property(x => x.UserName).IsModified = false;
            var isModified = context.Entry(user).Property(x => x.UserName).IsModified; // still true
        }
    }

    public class Users
    {
        public int Id { get; set; }
        public string UserName { get; set; }
        public string City { get; set; }
    }
    public class Context : DbContext
    {
        public virtual DbSet<Users> Users { get; set; }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<Users>(b =>
            {
                b.HasKey(r => r.Id);

                b.Property(x => x.UserName).IsRequired();
            });
        }
        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder.UseSqlServer("Server=.\\SQLEXPRESS;Initial Catalog=Test;Trusted_Connection=True;MultipleActiveResultSets=true");
        }
    } 
}
```
 
[http://stackoverflow.com/questions/12661881/exclude-property-on-update-in-entity-framework](http://stackoverflow.com/questions/12661881/exclude-property-on-update-in-entity-framework)

ajcvickers:
Notes for triage: Confirmed this is a bug, but not a regression. The original value is being left as-is when setting the IsModified flag. This causes DetectChanges to detect that the current and original values are different, and so the property is marked as modified again.

Workaround: When setting the original value to be the same as the current value before marking the property as not modified. For example:
```C#
var propertyEntry = context.Entry(user).Property(x => x.UserName);
[...]

PR #8957, opened 2017-06-23
Fix IsModified issues

ajcvickers:
Issues #7798, #8265, #8465

These issues were due to incorrect handling of nulls/conceptual nulls for required but nullable properties, in particular when attaching an entity with a required property that starts off null.


ajcvickers:
Merged.

Issue #8101, opened 2017-04-07
InvalidOperationException: "Collection was modified" inside NavigationFixer


joshmouch:
I have some code that is adding an entity to the DbContext if it hasn't previously been added.  This works in most cases, however, in one case I've found, there's an exception being throw in the `ChangeTracking.Internal.NavigationFixer`.

The code is:
```
public void Add<TEntity>(TEntity entity) where TEntity : class
		{
			var entry = this.DataContext.Entry(entity);
			if (entry.State == EntityState.Detached)
			{
				this.DataContext.Add((object)entity);
			}
		}

```

The exact exception is:
```
System.InvalidOperationException occurred
  HResult=0x80131509
  Message=Collection was modified; enumeration operation may not execute.
  Source=System.Core
  StackTrace:
   at System.Collections.Generic.HashSet`1.Enumerator.MoveNext()
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.NavigationFixer.InitialFixup(InternalEntityEntry entry, ISet`1 handledForeignKeys, Boolean fromQuery)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.NavigationFixer.StateChanged(InternalEntityEntry entry, EntityState oldState, Boolean fromQuery)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntryNotifier.StateChanged(InternalEntityEntry entry, EntityState oldState, Boolean fromQuery)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState entityState, Boolean acceptChanges)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityGraphAttacher.PaintAction(EntityEntryGraphNode node)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph(EntityEntryGraphNode node, Func`2 handleNode)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityGraphAttacher.AttachGraph(InternalEntityEntry rootEntry, EntityState entityState)
[...]

PR #8959, opened 2017-06-23
Prevent Collection was modified; enumeration operation may not execute.

ajcvickers:
Issue #8101

Happens when entities are inconsistently fixed up and do shenanigans in setters. (May also happen in other scenarios, but I was able to repro it this way.) Fix is to do defensive copy.

### DIFF
--- a/src/EFCore/ChangeTracking/Internal/NavigationFixer.cs
+++ b/src/EFCore/ChangeTracking/Internal/NavigationFixer.cs
@@ -607,7 +607,7 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
                         {
                             if (principalToDependent.IsCollection())
                             {
-                                var dependents = ((IEnumerable)navigationValue).Cast<object>();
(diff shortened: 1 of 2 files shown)


src/EFCore/ChangeTracking/Internal/NavigationFixer.cs
```diff
@@ -607,7 +607,7 @@ private void DeleteFixup(InternalEntityEntry entry)
                         {
                             if (principalToDependent.IsCollection())
                             {
-                                var dependents = ((IEnumerable)navigationValue).Cast<object>();
+                                var dependents = ((IEnumerable)navigationValue).Cast<object>().ToList();
```
anpete: Consider using ArrayList here to avoid allocating during Cast.

ajcvickers: ArrayList requires an ICollection, but we only have an IEnumerable

anpete: Typical!

Issue #8617, opened 2017-05-26
Query/Test: introduce client-side ordering to QueryTests without explicit orderby, rather than using contains in the result verification


maumar:
Currently QueryTest base (and some others) verify unordered results by using Contains. This is not perfect because it's slow and can cause false positives for queries where results have duplicates. Instead, we should apply logic similar to one in ComplexNavigationsQueryTest, where we apply client-side ordering and perform normal comparison afterwards

maumar:
related to #8606

maumar:
fixed in f45292f2fd1a0edc2c364dccac00cb242b6808ae

PR #8960, opened 2017-06-23
Fix to #8617 - Query/Test: introduce client-side ordering to QueryTests without explicit orderby, rather than using contains in the result verification

maumar:
This is part 1, PR includes:
- applying client side ordering to QueryTests (rather than doing O(n^2) result comparisons for queries without order by,
- adding entry count verification (we had it in some places and not in others),
- DRYing some of the commonly used asserters

As a result, time to run those tests is cut down in half (40s to 20s). For now the helper methods are duplicated with the ones used in complex navs - will unify them in next checkin.

### DIFF
--- a/src/EFCore.Specification.Tests/Query/QueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/QueryTestBase.cs
@@ -87,8 +87,9 @@ namespace Microsoft.EntityFrameworkCore.Query
             var context = new Context();
             context.Arguments.Add("customerId", "ALFKI");
 
-            AssertQuery<Customer>(cs =>
-                    cs.Single(c => c.CustomerID == (string)context.Arguments["customerId"]));
+            AssertSingleResult<Customer>(
+                cs => cs.Single(c => c.CustomerID == (string)context.Arguments["customerId"]),
+                entryCount: 1);
         }
 
         [ConditionalFact]
@@ -235,6 +236,7 @@ namespace Microsoft.EntityFrameworkCore.Query
         {
             AssertQuery<Customer>(
                 cs => cs.Select(c => new { c }),
+                e => e.c.CustomerID,
                 entryCount: 91);
         }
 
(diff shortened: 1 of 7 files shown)


src/EFCore.Specification.Tests/Query/QueryTestBase.cs
```diff
@@ -3981,237 +3976,390 @@ protected QueryTestBase(TFixture fixture)
...
-        {
-            using (var context = CreateContext())
-            {
-                TestHelpers.AssertResults(
-                    new[] { query(NorthwindData.Set<TItem>()) },
-                    new[] { query(context.Set<TItem>()) },
-                    assertOrder);
-            }
-        }
+        private void AssertSingleResult<TItem1>(
```
maumar: all the code below will go away - it will be unified with helper methods in complex navs

smitpatel:
Test failures

maumar:
@smitpatel skipped for now, tracking here: https://github.com/aspnet/EntityFramework/issues/8962

maumar:
ping @smitpatel @anpete

src/EFCore.Specification.Tests/Query/QueryTestBase.cs
```diff
@@ -3981,237 +3980,390 @@ protected QueryTestBase(TFixture fixture)
...
-            Action<IList<object>, IList<object>> asserter = null)
-            where TItem : class
+        private void AssertQuery<TItem1, TItem2>(
+            Func<IQueryable<TItem1>, IQueryable<TItem2>, IQueryable<int>> query,
+            bool assertOrder = false)
+            where TItem1 : class
+            where TItem2 : class
+            => AssertQueryScalar22(query, query, assertOrder);
+
+        private void AssertQueryScalar22<TItem1, TItem2, TResult>(
```
smitpatel: weird name.

maumar: whoops, will clean this up

PR #8963, opened 2017-06-24
Remove GetTempFileName

jbagga:
Addresses aspnet/Coherence-Signed#510

cc @smitpatel @Eilon

### DIFF
--- a/src/dotnet-ef/Project.cs
+++ b/src/dotnet-ef/Project.cs
@@ -67,7 +67,7 @@ namespace Microsoft.EntityFrameworkCore.Tools
             }
 
             IDictionary<string, string> metadata;
-            var metadataFile = Path.GetTempFileName();
+            var metadataFile = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
             try
             {
                 var propertyArg = "/property:EFProjectMetadataFile=" + metadataFile;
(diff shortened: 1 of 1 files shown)


smitpatel:
Does the new API create the file?

bricelam:
> Does the new API create the file?

No. It also risks collisions.

bricelam:
My question is: Why aren't people deleting their temporary files? I've always wrapped these in a `try..finally` block.

smitpatel:
Looks like false alarm then.

jbagga:
@Eilon and I discussed this and he strongly suggests we use the new API. Other processes may use GetTempFileName which has caused conflicts before. You are welcome to suggest other possible solutions, other APIs etc.

ajcvickers:
@Eilon @jbagga We feel the new code is making it more flaky. If we really feel this is an issue, then please work with @bricelam and @smitpatel to come up with a solution that will be at least as robust as the code that is being replaced.

Eilon:
The problem is that *other* components don't clean up their temporary files, so then GetTempFile will fail 100% of the time with no recovery. These "other components" could be any Windows app that uses this API, which is a zillion apps, and not all are well-written. This is a real issue that we've hit in other places, not just a theoretical defensive move.

In what way do we think the new code is flaky? The new code is based on crypto APIs, so if you find a way to cause a collision, there's probably a much bigger problem...

ajcvickers:
@bricelam @smitpatel Can one of you come up with a solution that we don't think is flakey that doesn't use GetTempFileName?

@Eilon Is there an external issue that you know of tracking the problem with GetTempFileName? It seems like we're saying that we should never use this, which means that probably nobody should ever be using it until it gets fixed.

jbagga:
Moved to 2.1.0

cc @Eilon

ajcvickers:
Merged.

PR #8964, opened 2017-06-24
Query: Fix bug in IncludeCompiler where we would resolve an incorrect QSRE for grouped results.

anpete:
When GroupBy is in play we need to ensure we use ElementSelector and not SelectClause.Selector as the output expression.

### DIFF
--- a/src/EFCore/Query/Internal/QueryModelExtensions.cs
+++ b/src/EFCore/Query/Internal/QueryModelExtensions.cs
@@ -1,13 +1,17 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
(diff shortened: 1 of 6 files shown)


src/EFCore/Query/Internal/QueryModelExtensions.cs
```diff
@@ -16,7 +20,42 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
...
+            {
+                outputExpression = groupResultOperator.ElementSelector;
+            }
+            else if (queryModel.SelectClause.Selector.Type.IsGrouping())
+            {
+                var subqueryExpression
+                    = (queryModel.SelectClause.Selector
+                        .TryGetReferencedQuerySource() as MainFromClause)?.FromExpression as SubQueryExpression;
+
+                var nestedGroupResultOperator
```
maumar: add test for nested groupby

maumar: also, make recursive?

anpete: Copied existing logic from TrackEntitiesInResults, will file a bug to investigate whether it is good enough.

PR #8966, opened 2017-06-26
Add database schema creator. 

MSRiganti:
Used when creating a Database or Table with the new Database schema.
Driven by annotations.

-> [x ] The code builds and tests pass - DONE


ErikEJ:
Where are the tests?

smitpatel:
What issue this solves? There is already `EnsureSchemaOperation` which creates required schema.

bricelam:
This doesn't fit the existing Migrations architecture. Only the database needs to be handled this way. Once there is a database to connect to, all database artifacts can be created using DDL.

In Migrations, the DDL statements are represented as `MigrationsOperation` classes. (e.g. `EnsureSchemaOperation`) The database provider transforms these objects into SQL statements (e.g. `CREATE SCHEMA`) using their `IMigrationsSqlGenerator` implementation.

The scenario it looks like you're trying to solve (creating the schema specified in `ModelBuilder.HasDefaultSchema()`) should already be handled when calling `DbContext.Database.Migrate()` (or `DbContext.Database.EnsureCreated()`). If you are running into issues with it, please [submit an issue](https://github.com/aspnet/EntityFramework/issues/new).

Issue #8157, opened 2017-04-13
Investigate netcoreapp2.0 failures


bricelam:
Some tests started failing after upgrading to netcoreapp2.0. We should investigate the root cause and file appropriate issues on our dependencies.

* EFCore.InMemory.FunctionalTests
  * ComplexNavigationsQueryInMemoryTest
    - [ ] Optional_navigation_projected_into_DTO
* EFCore.Sqlite.FunctionalTests
  * ComplexNavigationsQuerySqliteTest
    - [ ] Optional_navigation_projected_into_DTO
* EFCore.SqlServer.FunctionalTests
  * ComplexNavigationsQuerySqlServerTest
    - [ ] Optional_navigation_projected_into_DTO
  * DbContextPoolingTest
    - [x] State_manager_is_reset
  * QuerySqlServerTest
    - [x] Cache_key_contexts_are_detached

bricelam:
The last two tests may just be making invalid assumptions about the implementation of `GC.Collect()`

ajcvickers:
@ajcvickers will investigate bottom two
@maumar to investigate others

bricelam:
Found a few more after upgrading the specification assemblies to netcoreapp2.0:

* EFCore.Sqlite.FunctionalTests
  * QuerySqliteTest
    - [ ] TrimEnd_in_predicate
    - [ ] TrimStart_in_predicate
* EFCore.SqlServer.FunctionalTests
  * QuerySqlServerTest
    - [ ] TrimEnd_in_predicate
    - [ ] TrimStart_in_predicate

cc @smitpatel

maumar:
more failing tests:

- [ ] Select_GroupBy_SelectMany
- [ ] Select_nested_collection_count_using_DTO
- [ ] Select_DTO_distinct_translated_to_server
- [ ] Select_DTO_constructor_distinct_translated_to_server
- [ ] Select_DTO_with_member_init_distinct_translated_to_server
- [ ] Select_DTO_with_member_init_distinct_in_subquery_used_in_projection_translated_to_server

all failing with the same underlying issue:

```
System.ArgumentOutOfRangeException : Length cannot be less than zero.
Parameter name: length
Stack Trace:
	at System.String.Substring(Int32 startIndex, Int32 length)
	at Microsoft.CSharp.RuntimeBinder.Syntax.NameTable.Add(String key, Int32 length)
[...]

PR #8969, opened 2017-06-26
Fix WeakReference tests

ajcvickers:
Part of #8157

Used pattern from SO to ensure that compiler optimizations don't impact test: https://stackoverflow.com/questions/15205891/garbage-collection-should-have-removed-object-but-weakreference-isalive-still-re


ajcvickers:
Merged

PR #8974, opened 2017-06-26
Disable xUnit Skip warnings

smitpatel:


bricelam:
Any reason not to disable in `common.props`?

smitpatel:
@bricelam - Updated common.props

Issue #7012, opened 2016-11-14
Document exceptions


denisz1:
MS SDKs typically document which exceptions are thrown.

This is crucial information. We cannot properly handle our error paths without knowing what is going wrong (and catching an `Exception` tells us nothing).

Many EF Core methods lack complete documentation, because they do not mention exceptions thrown. [MSDN was good at this](http://msdn.microsoft.com/en-us/library/system.data.entity.dbcontext.savechanges(v=vs.113).aspx#Anchor_1), but the new docs site seems incomplete.

**These are the [worst culprits](http://docs.microsoft.com/en-us/ef/core/api/microsoft.entityframeworkcore.dbcontext#Microsoft_EntityFrameworkCore_DbContext_SaveChanges):**

    DbContext.SaveChanges
    DbContext.SaveChangesAsync

We cannot rely on the source, because that can change at any time. You need to **officially document** this.

It's not hard, PLEASE consider this a priority.

*(In the meantime, please at least let us know about those two methods.)*

rowanmiller:
Note to self - look for custom exception types that we have defined

PR #8975, opened 2017-06-26
Add exception documentation for SaveChanges

ajcvickers:
Issue #7012

Text is copied from the documentation for the exception types.
Also checked that there are no other custom exception types used at runtime that do not have similar documentation. All other exceptions indicate a program/logic error, not a recoverable state, and use built-in exception types.

### DIFF
--- a/src/EFCore/DbContext.cs
+++ b/src/EFCore/DbContext.cs
@@ -319,6 +319,14 @@ namespace Microsoft.EntityFrameworkCore
         /// <returns>
         ///     The number of state entries written to the database.
         /// </returns>
+        /// <exception cref="DbUpdateException">
+        ///     An error is encountered while saving to the database.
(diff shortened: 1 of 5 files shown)


src/EFCore/DbContext.cs
```diff
@@ -319,6 +319,14 @@ protected internal virtual void OnModelCreating(ModelBuilder modelBuilder)
         /// <returns>
         ///     The number of state entries written to the database.
         /// </returns>
+        /// <exception cref="DbUpdateException">
+        ///     Thrown when an error is encountered while saving to the database.
```
bricelam: The common pattern is to not put the *Thrown if* (or *when*) part.

ajcvickers: Fixed here and other places in the code.

PR #8977, opened 2017-06-26
Test: Use different database for FiltersInheritanceTests

smitpatel:
Since InheritanceTests modify the data causing flaky errors
Also add InheritanceTests for Sqlite

@anpete @AndriySvyryd

### DIFF
--- a/test/EFCore.SqlServer.FunctionalTests/Query/InheritanceSqlServerFixture.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/Query/InheritanceSqlServerFixture.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.EntityFrameworkCore.TestModels.Inheritance;
 using Microsoft.EntityFrameworkCore.Utilities;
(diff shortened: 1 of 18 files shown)


smitpatel:
@AndriySvyryd - Is there an example of it?

AndriySvyryd:
@smitpatel Yes, Find usages

smitpatel:
@AndriySvyryd - Updated.

smitpatel:
@AndriySvyryd @anpete - Can you review again? I had to change a lot in test fixture & test base to make tests work with transaction.

test/EFCore.SqlServer.FunctionalTests/Query/InheritanceSqlServerFixture.cs
```diff
@@ -9,34 +10,54 @@ namespace Microsoft.EntityFrameworkCore.Query
...
+                .UseInternalServiceProvider(serviceProvider)
+                .Options;
+        }
 
         public override SqlServerTestStore CreateTestStore()
-            => SqlServerTestStore.GetOrCreateShared(DatabaseName, () =>
-                {
-                    using (var context = CreateContext())
+        {
+            return SqlServerTestStore.GetOrCreateShared(
```
AndriySvyryd: Why not expression body? :trollface:

PR #8980, opened 2017-06-26
Design: Handle generic types in namespace search

### LINKED ISSUE #7352: Design: Look at generic parameters when generating namespace using directives
In various places where we generate code we look at the namespaces of the types used in order to generate using directives. This was done in #5912, which has the details of where we do this. However, if the type is generic, then we should also generate using directives for the namespaces of those types.

smitpatel:
Resolves #7352


smitpatel:
@bricelam - I tried adding test but the property is identified as navigation and doesn't work unless we support such property as scalar.

Issue #7985, opened 2017-03-24
Identifying shadow FK values are not propagated when tracked as unchanged or modified


AndriySvyryd:
If the shadow PK is of nullable type 
`context.ChangeTracker.TrackGraph(principal, e => e.Entry.State = EntityState.Unchanged);`
 throws
 `Unable to track an entity of type '...' because primary key property '...' is null`

`context.Attach(principal)` works, but marks the dependent as added

ajcvickers:
@AndriySvyryd Having trouble fully understanding this. If the key value of the dependent is null, then changing state to Unchanged without setting the key value should throw. Likewise, using Attach will mark the dependent as Added because it didn't have a key value set, assuming the key is generated. If the key is not generated, then it should throw like TrackGraph does. 

What am I missing?

AndriySvyryd:
@ajcvickers The principal has the key set. How can I attach the principal and the dependent as Unchanged?

ajcvickers:
@AndriySvyryd When attaching, the primary key must be set. Because when you attach, you are saying that this entity already exists in the database, and hence already has a primary key value assigned. 

If the primary key is in shadow state, and you disconnected the entity, then you need to carry the primary key with the entity in some way so that it can then be set again on the state entry before the entity is attached.

ajcvickers:
Changed the behavior for owned types where the state must match that of the owner, and the key values can be propagated said owner. Behavior for non-owned types remains the same.

AndriySvyryd:
@ajcvickers I don't see why owned types would be different, the owned dependent is not required to exist as there's no way of enforcing it.
Carrying around the key values and setting them feels too cumbersome when dealing with an untracked entity graph. In the common cases the dependents will have the same state as the principal, and if not it's simpler to fix the state once the entities are tracked.

ajcvickers:
I don't think we can say that an existing principal and an existing dependent is any more or less common than an existing principal and an new dependent.

Also, "an owned dependent is not required to exist as there's no way of enforcing it." I think we should discuss this since it seems to be contrary to the concept of an aggregate. That is, it's okay for some entity in an aggregate to not exist, but I don't think it's necessarily okay to then later add it to the aggregate and have it inserted. But if this is really okay, then I think aggregates have the same problem as non-aggregates and I don't see anyway to make Attach work in a standard way for primary keys in shadow state--we simply have no way of knowing the the dependent is new or not and making either choice will be right or wrong in different circumstances.

@divega?

[...]

PR #8986, opened 2017-06-27
Make owned entities take their state from owner when attaching

ajcvickers:
Issue #7985

Moved key propagation inside state manager for keys that are unknown--typically shadow keys on attached entities. This prevents common cases of key not set for identifying relationships. Also added special behavior used by Attach that will take the entity state from the principal. If graph attach comes from DetectChanges, then state is still Added--because it wasn't there and now it is.


src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -278,14 +278,15 @@ public virtual InternalEntityEntry TryGetEntry(object entity)
             {
                 // ReSharper disable once CheckForReferenceEqualityInstead.2
                 if (Equals(keyValue.Key.ClrType, type)
-                    && keyValue.Value.TryGetValue(entity, out entry))
+                    && keyValue.Value.TryGetValue(entity, out InternalEntityEntry foundEntry))
```
anpete: Use out var

ajcvickers:
@AndriySvyryd @anpete Updated this based on triage discussion/decisions.

PR #8987, opened 2017-06-27
Fix #7901 - Max(), Min(), etc. do not differentiate between no results and zero

### LINKED ISSUE #7901: Max(), Min(), etc. do not differentiate between no results and zero
The `Min()` and `Max()` aggregate functions, and their async counterparts, make it difficult to differentiate between the case that there are no matching rows on which to perform the aggregate, and the minimum or maximum is zero.

Interestingly, the standard behaviour for the LINQ `Min()` and `Max()` functions is to throw `InvalidOperationException` if there are no matching rows. This is a bit of a pain—returning `null` makes more sense really—but it would be helpful if this difference was clearly documented.

    class Program
    {
        {
            // Using EF Core 1.1.1 and matching SQL Server provider in full .NET Framework
            // on Windows 10 with RTM version of VS 2017.
            {
                ctx.Database.EnsureCreated();

                // If the value being aggregated is nullable, we're fine - null
                // is returned if there are no matching rows.
                int? a = ctx.Things.Max(x => x.OtherThing);

                // We can explicitly invoke an overload of Max() that *does* return
                // a nullable value.
                int? b = ctx.Things.Max(new Func<Thing, int?>(x => x.ThingyWotsit));

                // That works after a .Select() too.
                int? c = (from s in ctx.Things
                          select s).Max(new Func<Thing, int?>(x => x.ThingyWotsit));

                // All these alternatives fail with similar errors:
                // System.ArgumentException: 'Expression of type 'System.Collections.Generic.IEnumerable`1[System.Int32]' cannot be used for return type 'System.Collections.Generic.IEnumerable`1[System.Nullable`1[System.Int32]]''
                int? d = ctx.Things.Select(x => (int?)x.ThingyWotsit).Max();

                int? e = ctx.Things.Select<Thing, int?>(x => x.ThingyWotsit).Max();

                int? f = (from s in ctx.Things
                          select s).Max<Thing, int?>(x => x.ThingyWotsit);

                int? g = ctx.Things.Select(x => x.ThingyWotsit).Max<int, int?>(x => x);
            }
        }
    }

    {
        public DbSet<Thing> Things => Set<Thing>();

        {
            optionsBuilder.UseSqlServer("Server=(localdb)\\mssqllocaldb;Initial Catalog=MyDb;Integrated Security=true");
        }
    }

    public class Thing
    {
        public int ThingId { get; set; }

        public int ThingyWotsit { get; set; }

        public int? OtherThing { get; set; }
    }

anpete:
Max, Min and Average should throw "Sequence contains no elements" when there was no data returned and the result type is non-nullable. Updated relational result operator handling to do that.

- Added a warning for the subquery throwing case.

### DIFF
--- a/src/EFCore.Relational/Properties/RelationalStrings.resx
+++ b/src/EFCore.Relational/Properties/RelationalStrings.resx
@@ -240,6 +240,10 @@
     <value>Possible unintended use of method Equals(object) for arguments of different types in expression '{expression}'. This comparison will always return 'false'.</value>
     <comment>Warning RelationalEventId.QueryPossibleUnintendedUseOfEqualsWarning object</comment>
   </data>
+  <data name="LogQueryPossibleExceptionWithAggregateOperator" xml:space="preserve">
+    <value>Possible unintended use of a potentially throwing aggregate method (Min, Max, Average) in a subquery. Client evaluation will be used and operator will throw if no data exists. Changing the subquery result type to a nullable type will allow full translation.</value>
+    <comment>Warning RelationalEventId.QueryPossibleExceptionWithAggregateOperator</comment>
+  </data>
   <data name="StoredProcedureIncludeNotSupported" xml:space="preserve">
     <value>The Include operation is not supported when calling a stored procedure.</value>
   </data>
@@ -395,6 +399,9 @@
   </data>
   <data name="UnsupportedPropertyType" xml:space="preserve">
     <value>No mapping to a relational type can be found for property '{entity}.{property}' with the CLR type '{clrType}'.</value>
+  </data>
+  <data name="NoElements" xml:space="preserve">
+    <value>Sequence contains no elements.</value>
   </data>
   <data name="LogRelationalLoggerConnectionErrorAsDebug" xml:space="preserve">
     <value>An error occurred using the connection to database '{database}' on server '{server}'.</value>
--- a/src/EFCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
+++ b/src/EFCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
@@ -244,9 +244,12 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
 
                     handlerContext.SelectExpression.SetProjectionExpression(averageExpression);
 
+                    var averageExpressionType = averageExpression.Type;
+                    var throwOnNullResult = DetermineAggregateThrowingBehavior(handlerContext, averageExpressionType);
+                    
                     return (Expression)_transformClientExpressionMethodInfo
-                        .MakeGenericMethod(averageExpression.Type)
-                        .Invoke(null, new object [] { handlerContext });
+                        .MakeGenericMethod(averageExpressionType)
+                        .Invoke(null, new object [] { handlerContext, throwOnNullResult });
                 }
             }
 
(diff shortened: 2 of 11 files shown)


test/EFCore.SqlServer.FunctionalTests/Query/QuerySqlServerTest.ResultOperators.cs
```diff
@@ -231,14 +231,16 @@ public override void Average_on_float_column_in_subquery()
...
-                @"SELECT [o].[OrderID], (
-    SELECT CAST(AVG(CAST([od].[Discount] AS real)) AS real)
-    FROM [Order Details] AS [od]
-    WHERE [o].[OrderID] = [od].[OrderID]
-) AS [Sum]
+            AssertContainsSql(
+                @"SELECT [o].[OrderID]
 FROM [Orders] AS [o]
-WHERE [o].[OrderID] < 10300");
+WHERE [o].[OrderID] < 10300",
```
maumar: this seems bad - should we add a context option to get the old behavior? or perhaps introduce a warning telling people to switch to nullable type to avoid N+1

anpete: Bad because we are changing the behavior? We already did the right thing for Single etc. and so this is just fixing that inconsistency. We will gen a client-eval warning as expected.

anpete: Going to add a new warning for this specific case.

anpete:
@maumar New version up.

src/EFCore.Specification.Tests/Query/QueryTestBase.ResultOperators.cs
```diff
@@ -254,6 +254,13 @@ public virtual void Average_on_float_column_in_subquery()
         }
 
         [ConditionalFact]
+        public virtual void Average_on_float_column_in_subquery_with_cast()
+        {
+            AssertQuery<Order>(os => os.Where(o => o.OrderID < 10300)
+                .Select(o => new { o.OrderID, Sum = o.OrderDetails.Average(od => (float?)od.Discount) }));
```
maumar: after merge you will have to add ordering func to this (because it projects anonymous type, like so:

AssertQuery<Order>(os => os.Where(o => o.OrderID < 10300)
        .Select(o => new { o.OrderID, Sum = o.OrderDetails.Average(od => (float?)od.Discount) }),
    e => e.OrderID);

src/EFCore.Relational/Properties/RelationalStrings.resx
```diff
@@ -240,6 +240,10 @@
     <value>Possible unintended use of method Equals(object) for arguments of different types in expression '{expression}'. This comparison will always return 'false'.</value>
     <comment>Warning RelationalEventId.QueryPossibleUnintendedUseOfEqualsWarning object</comment>
   </data>
+  <data name="LogQueryPossibleExceptionWithAggregateOperator" xml:space="preserve">
+    <value>Possible unintended use of a potentially throwing aggregate method (Min, Max, Average) in a subquery. Client-eval will be used and operator will throw if no data exists. Changing the subquery result type to a nullable type will allow full translation.</value>
```
maumar: client evaluation

src/EFCore.Relational/Query/Internal/RelationalResultOperatorHandler.cs
```diff
@@ -716,7 +741,7 @@ private static Expression HandleSum(HandlerContext handlerContext)
 
                     return (Expression)_transformClientExpressionMethodInfo
                         .MakeGenericMethod(sumExpression.Type)
-                        .Invoke(null, new object [] { handlerContext });
+                        .Invoke(null, new object [] { handlerContext, false });
```
smitpatel: nit: comment parameter name for  bool value?

Issue #6018, opened 2016-07-07
Reverse Engineer: Cmd parameter to disable naming logic


rowanmiller:
We have the ability to create a custom service that handles column => property naming (see https://gist.github.com/natemcmaster/353f4ab3efb4514eeaec846df28f0e24). However, some folks may want to opt out of our default logic and just have property names line up with column names. We think this makes sense to have as a top level parameter to the commands.

BTW we should have a design discussion about whether this literally just uses the column name, or makes some basic attempts to land with a valid C# identifier.

ajcvickers:
A lot of people have been asking for this, so we should try to do it for 2.0.

JuanIrigoyen:
In the previous version, I used the solution proposed by natemcmaster for regenerate my model https://gist.github.com/natemcmaster/353f4ab3efb4514eeaec846df28f0e24,

In the next version I used a modified version proposed by 
https://romiller.com/2017/02/10/ef-core-1-1-pluralization-in-reverse-engineer/

But now with the last version nothing of the two works, who can I regenerate my model for disable name-changing logic.

smitpatel:
@JuanIrigoyen - Solution posted by Nate should still work.

lajones:
@ajcvickers @smitpatel @bricelam @divega Do we need to design this? I was getting going on a simple boolean flag `--use-database-names` which would switch this on. But do we want to consider a future where users can e.g. name a `CandidateNamingService` which we then search the project for? Even if we don't do the search right now it might argue for a `--use-naming-service [service_name]` flag instead (where for right now we only recognize `PascalCase` and `UseDatabaseNames` naming services)?

divega:
@lajones can we bring this to the design meeting tomorrow? 

On first thought I like the idea of adding a very simple flag that maps directly to what seems to be a common need, e.g. `--no-pascal-case`.  

Bringing your own naming service already requires coding the service (or at least installing a NuGet package containing it), so I think it might be ok for that scenario to always require some tinkering with design time services.

But even if we later decide to do something in the command line to allow specifying custom naming services it won't necessarily obsolete the flag: We would just need to decide what happens when the user
specifies both settings. 

[...]

Issue #8853, opened 2017-06-14
RevEng: Clean up internal code


bricelam:
A few ideas:
- [ ] Move checks for default constraint names in CSharpDbContextGenerator to ScaffoldingModelFactory
- [ ] Better variable and parameter names
- [x] Rename ModelScaffolder to ReverseEngineeringScaffolder OR merge it into DatabaseOperations
- [x] Decouple RelationalScaffoldingModelFactory from IDatabaseModelFactory
- [ ] Use resources (issue #8599)
- [x] Clean up IModel annotations (issue #8689)
- [x] Remove DbDataReaderExtension (issue #8712)
- [x] Remove TableSelectionSet
- [ ] Merge code which does similar things in Migrations (issue #1671)
  - CSharpUtilities cf CSharpHelper
  - code that transforms directory paths into namespaces
  - code for GenerateAnnotations including GenerateFluentApiForAnnotation
  - DbContext writer does not generate namespaces for custom types
- [ ] Uniquification is done twice in `CSharpUtilities` and in `CSharpUniqueNamer`. We should rationalize this and figure out how configurable candidate naming, ensuring valid C# identifier, storing the names so that another name used in the same context clashes, and pluralization/singularization fits in to that process (see #8989).
- [x] Use IOperationsReporter and vanilla Logging (issue #8686)
- [x] Remove IRelationalTypeMapper dependency in RelationalScaffoldingModelFactory (use IScaffoldingTypeMapper)
- [x] Remove protected dependency properties on internal services
- [x] Remove ServiceCollection.AddScaffolding() extension method (update tests to build the full design-time service provider)
- [x] Decouple IScaffoldingCodeGenerator form I/O (see [this comment](https://github.com/aspnet/EntityFrameworkCore/pull/10356#issuecomment-346412414))

smitpatel:
`TableSelectionSet` being removed in #9315

ajcvickers:
Closing since we've done what we are going to do here. PRs still welcome for good cleanup.

PR #8989, opened 2017-06-27
UseDatabaseNames added to Scaffolding

lajones:
Fix for #6018 .

The `--use-database-names` flag can now be used to control whether scaffolding a) does or does not attempt to Pascalize the names from the database, and b) produces singularized/pluralized names.

### DIFF
--- a/src/EFCore.Design/Scaffolding/Internal/CSharpNamer.cs
+++ b/src/EFCore.Design/Scaffolding/Internal/CSharpNamer.cs
@@ -16,6 +16,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
     {
         private readonly Func<T, string> _nameGetter;
         private readonly ICSharpUtilities _cSharpUtilities;
+        private readonly Func<string, string> _singularizePluralizer;
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
@@ -27,13 +28,16 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public CSharpNamer([NotNull] Func<T, string> nameGetter, [NotNull] ICSharpUtilities cSharpUtilities)
+        public CSharpNamer([NotNull] Func<T, string> nameGetter,
+            [NotNull] ICSharpUtilities cSharpUtilities,
+            [CanBeNull] Func<string, string> singularizePluralizer)
         {
             Check.NotNull(nameGetter, nameof(nameGetter));
             Check.NotNull(cSharpUtilities, nameof(cSharpUtilities));
 
             _nameGetter = nameGetter;
             _cSharpUtilities = cSharpUtilities;
+            _singularizePluralizer = singularizePluralizer;
         }
 
         /// <summary>
@@ -49,7 +53,8 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
                 return NameCache[item];
             }
 
-            var name = _cSharpUtilities.GenerateCSharpIdentifier(_nameGetter(item), null);
+            var name = _cSharpUtilities.GenerateCSharpIdentifier(
+                _nameGetter(item), existingIdentifiers: null, singularizePluralizer: _singularizePluralizer);
             NameCache.Add(item, name);
             return name;
         }
--- a/src/EFCore.Design/Scaffolding/Internal/IScaffoldingModelFactory.cs
+++ b/src/EFCore.Design/Scaffolding/Internal/IScaffoldingModelFactory.cs
@@ -10,5 +10,6 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
     public interface IScaffoldingModelFactory
     {
         IModel Create([NotNull] string connectionString, [NotNull] IEnumerable<string> tables, [NotNull] IEnumerable<string> schemas);
+        bool UseDatabaseNames { get; set; }
     }
 }
--- a/src/EFCore.Design/Scaffolding/Internal/IScaffoldingModelFactory.cs
+++ b/src/EFCore.Design/Scaffolding/Internal/IScaffoldingModelFactory.cs
@@ -9,7 +9,6 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
 {
     public interface IScaffoldingModelFactory
     {
-        IModel Create([NotNull] string connectionString, [NotNull] IEnumerable<string> tables, [NotNull] IEnumerable<string> schemas);
-        bool UseDatabaseNames { get; set; }
+        IModel Create([NotNull] string connectionString, [NotNull] IEnumerable<string> tables, [NotNull] IEnumerable<string> schemas, bool useDatabaseNames);
     }
 }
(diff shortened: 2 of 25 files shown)


src/dotnet-ef/Properties/Resources.resx
```diff
@@ -273,6 +273,9 @@
   <data name="UnsupportedFramework" xml:space="preserve">
     <value>Startup project '{startupProject}' targets framework '{targetFramework}'. The Entity Framework Core .NET Command Line Tools don't support this framework.</value>
   </data>
+  <data name="UseDatabaseNamesDescription" xml:space="preserve">
+    <value>Use table and column names directly from the database.</value>
+  </data>
```
lajones: I did try to be more explanatory, but the explanation got too long to display easily. But if anyone has a better idea of what to say here speak up.

src/EFCore.Design/Scaffolding/Internal/IScaffoldingModelFactory.cs
```diff
@@ -10,5 +10,6 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
     public interface IScaffoldingModelFactory
     {
         IModel Create([NotNull] string connectionString, [NotNull] IEnumerable<string> tables, [NotNull] IEnumerable<string> schemas);
+        bool UseDatabaseNames { get; set; }
```
smitpatel: Pass as argument.

src/EFCore.Design/Scaffolding/Internal/CSharpNamer.cs
```diff
@@ -27,13 +28,16 @@ public class CSharpNamer<T>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public CSharpNamer([NotNull] Func<T, string> nameGetter, [NotNull] ICSharpUtilities cSharpUtilities)
+        public CSharpNamer([NotNull] Func<T, string> nameGetter,
+            [NotNull] ICSharpUtilities cSharpUtilities,
+            [CanBeNull] Func<string, string> singularizePluralizer)
```
smitpatel: The design of passing func all around seems bad. All the naming happens inside of `RelationalScaffoldingModelFactory`. It should be contained within.

src/EFCore.Design/Scaffolding/Internal/IScaffoldingModelFactory.cs
lajones: I can see the argument (if you'll forgive the pun) for that.

src/EFCore.Design/Scaffolding/Internal/CSharpNamer.cs
lajones: The problem is that `UseDatabaseNames` is used in 2 places: 1) when we originally come up with candidate names (the default is to "Pascalize" the database name), 2) when we pluralize or singularize. In the past the algorithm was:

1. Come up with candidate name
2. Make sure it's a valid C# identifier
3. Uniquify
4. Pluralize or singularize

Steps 1 and 4 were done outside of `CSharpUtilities`, but 2 and 3 were done together at one time within it.

It was rightly pointed out in the bug that step 4 ought to happen before step 3. But this involves either coming back to the code that was calling `CSharpUtilities` between steps 2 and 3 above, or passing into `CSharpUtilities` what needed to be done. Since we already used the process of passing in a `Func` for the uniquification it seemed better to me to pass in a `Func` for this too. The other way will work as well - but we will then have to have separate calls within the calling code to `CSharpUtilities` to do step 2), then perform singularize/pluralize, then call back to `CSharpUtilities` to uniquify. That will be messy too.

smitpatel: I traced back why do we need to pass, func for uniquification. The only part when it differs is naming navigations. Interestingly it has note to do it using `CSharpUniqueNamer`.

Now the default func to uniquify is in `CSharpUtilities.Uniquifier`. The only case when it uses the list of identifiers is when naming navigation because the uniquification for rest of the stack happens using `CSharpUniqueNamer`.

So apart from 1 block (navigations), we do have every part in above list modularized. 
Of course best option would be to remove special processing for navigations but if that is bit far-fetched goal for this PR then, we should split it out to a single method in `CSharpUtilities` which does only step 3 in above list.
Then we can move around pluralizer & uniquifier. And then we don't need to pass in anything to `CSharpUtilities`.

Overall design would be,
1. CandidateNamingService
2. Make C# identifier
3. Pluralizer
4. Uniquification

Step 2,4 will call method in `CSharpUtilities` (Step 1,3 are services in ScaffoldingModelFactory)
In CSharpNamer, it calls the same method as step 2.
UniqueNamer will also call the method used in step 4.

src/dotnet-ef/Properties/Resources.resx
ErikEJ: "Use database column and table names as-is" ?

src/EFCore.Design/Scaffolding/Internal/CSharpNamer.cs
lajones: I started doing the plan above, but there is also a 5th step - which is storing off the generated name, so that a future name (generated in the same context) will clash causing uniquification. That was starting to get really complicated so (after talking this over with @smitpatel) going back to the original plan for now. I added a step in #8853 to cover fixing this properly.

src/dotnet-ef/Properties/Resources.resx
lajones: Thx, @ErikEJ. I started with that - but we do still make sure they are valid C# identifiers and that they are unique - so we may change them a little.

lajones:
Checked in with commit dc7d379af.

src/dotnet-ef/Properties/Resources.resx
piyey: Hi @ErikEJ , I just notice you mention "Use database *_column_* and table names as-is", i was asking in another topic about column names, because I used this parameter when scaffolding some tables and table names wasn't PascalCased but column names was. Is this a bug?
Regards!

ErikEJ: @piyey Yes, it is a "feature" issue: https://github.com/aspnet/EntityFrameworkCore/issues/9820

siliond:
Column Names are still changed, see EQID below:

`modelBuilder.Entity<DCUTRANSLATIONS>(entity =>
            {
                entity.HasKey(e => new { e.Eqid, e.Analyte });

                entity.Property(e => e.Eqid)
                    .HasColumnName("EQID")
                    .HasMaxLength(20);
...`

Command used:
`dotnet ef dbcontext scaffold "Server=SLCOHW3DSILIO1;Database=sunrise_demo_data;Trusted_Connection=True;" "Microsoft.EntityFrameworkCore.SqlServer" -o "Models" -f -t "DCUTRANSLATIONS" --use-database-names`

[CLI](https://docs.microsoft.com/en-us/ef/core/miscellaneous/cli/dotnet)
 --use-database-names                   Use table and **column** names directly from the database.

bricelam:
@siliond Can you submit a new issue for this?

smitpatel:
@bricelam - #9820

WhistlerHusky:
Is there a way not to change column names?

simeyla:
Note: For powershell the switch is `-UseDatabaseNames`

PR #8990, opened 2017-06-27
Create a relationship with just the inverse if an ambiguity was resolved

### LINKED ISSUE #8886: Metadata: Convention does not create relationship after Fluent API resolves ambiguity
```C#
public class Friendship
{
    public int Id { get; set; }

    public int ApplicationUserId { get; set; }
    public ApplicationUser ApplicationUser { get; set; }
}

public class ApplicationUser
{
    public int Id { get; set; }
}

{
    // Configure model
    modelBuilder.Entity<Friendship>();
}
```
Above model works but below one fails
```C#
public class Friendship
{
    public int Id { get; set; }

    public int ActionUserId { get; set; }
    public ApplicationUser ActionUser { get; set; }

    public int ApplicationUserId { get; set; }
    public ApplicationUser ApplicationUser { get; set; }
}

public class ApplicationUser
{
    public int Id { get; set; }
    public IList<Friendship> Friendships { get; set; }
}

{
    // Configure model
    modelBuilder.Entity<Friendship>().HasOne(e => e.ApplicationUser).WithMany(e => e.Friendships).HasForeignKey(e => e.ApplicationUserId);
}
```
Exception thrown:
```
Unable to determine the relationship represented by navigation property 'Friendship.ActionUser' of type 'ApplicationUser'. Either manually configure the relationship, or ignore this property using the '[NotMapped]' attribute or by using 'EntityTypeBuilder.Ignore' in 'OnModelCreating'.
```

The only difference between 2 models,
First one does not have ambiguity in relationship discovery. Second one has ambiguity but it is resolved by fluent API code but convention did not run correctly afterwards.

AndriySvyryd:
Fixes #8886

### DIFF
--- a/src/EFCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
+++ b/src/EFCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
@@ -105,10 +105,11 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
                     continue;
                 }
 
+                var candidateTargetEntityType = candidateTargetEntityTypeBuilder.Metadata;
                 var entityType = entityTypeBuilder.Metadata;
-                if (relationshipCandidates.TryGetValue(candidateTargetEntityTypeBuilder.Metadata, out RelationshipCandidate existingCandidate))
+                if (relationshipCandidates.TryGetValue(candidateTargetEntityType, out var existingCandidate))
                 {
-                    if (candidateTargetEntityTypeBuilder.Metadata != entityType
+                    if (candidateTargetEntityType != entityType
                         || !existingCandidate.InverseProperties.Contains(navigationPropertyInfo))
                     {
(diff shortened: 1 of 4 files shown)


src/EFCore/Metadata/Conventions/Internal/RelationshipDiscoveryConvention.cs
```diff
@@ -105,10 +105,11 @@ private IReadOnlyList<RelationshipCandidate> FindRelationshipCandidates(Internal
                     continue;
                 }
 
+                var candidateTargetEntityType = candidateTargetEntityTypeBuilder.Metadata;
                 var entityType = entityTypeBuilder.Metadata;
-                if (relationshipCandidates.TryGetValue(candidateTargetEntityTypeBuilder.Metadata, out RelationshipCandidate existingCandidate))
```
smitpatel: `out var`

smitpatel:
Test failures in model builder tests

Issue #8369, opened 2017-05-03
Unable to cast object of type 'PropertyExpression' to type 'NullConditionalExpression' for some queries with include, and multiple orderbys


myblindy:
I'm getting the exception in the title when using a very specific `ThenBy` expression. I'll include my EF model in a comment below since it's very large. I don't have a stack trace since I'm using the NuGet package directly, but it should be easily reproducible.

### Steps to reproduce
The code that produces the exception, as stripped down as possible.
```
            var collection = DrdbContext.DrDr
                .Include(w => w.DrHistory)
                    .ThenInclude(w => w.StatusNavigation)
                .Include(w => w.DrHistory)
                    .ThenInclude(w => w.AssignedNavigation)
                .Include(w => w.ProjectNavigation)
                .Include(w => w.AtaNavigation)

            var DrCollection = await collection
                    .OrderBy(w => w.ProjectNavigation.Nn).ThenBy(w => w.ProjDr)
                    .ToListAsync();
```

Removing `ThenBy` fixes it. Swapping the lambda expressions between `OrderBy` and `ThenBy` also fixes it. Though obviously the result isn't as intended.

### Further technical details
EF Core version: v2.0.0-preview2-24770
Database Provider: Microsoft.EntityFrameworkCore.SqlServer
Operating system: Windows 7 SP1
IDE: Visual Studio 2017

myblindy:
### The Context
```
    public partial class DrdbContext : DbContext
    {
        public virtual DbSet<DrAta> DrAta { get; set; }
        public virtual DbSet<DrDr> DrDr { get; set; }
        public virtual DbSet<DrEmployee> DrEmployee { get; set; }
        public virtual DbSet<DrHistory> DrHistory { get; set; }
        public virtual DbSet<DrProj> DrProj { get; set; }
        public virtual DbSet<DrStatus> DrStatus { get; set; }

        public DrdbContext(DbContextOptions<DrdbContext> options) : base(options)
        {
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<DrAta>(entity =>
            {
                entity.ToTable("dr_ata");

                entity.Property(e => e.Id).HasColumnName("id");

                entity.Property(e => e.Descr)
                    .IsRequired()
                    .HasColumnName("descr")
                    .HasColumnType("char(100)");
            });

            modelBuilder.Entity<DrDr>(entity =>
            {
                entity.ToTable("dr_dr");

[...]

PR #8991, opened 2017-06-27
[SMALL] Fix to #8369 - Unable to cast object of type 'System.Linq.Expressions.PropertyExpression' to type 'Microsoft.EntityFrameworkCore.Query.Expressions.Internal.NullConditionalExpression'

maumar:
Problem was that we had incorrect assumption in the include compiler when performing orderby lifting. When looking for duplicate orderings we always assumed that subquery projections would be an expression of a particular shape (methood call wrapped around NullConditionalExpression). However, this is not accurate because any expression could be there if it was a previous orderby expression.

Fix is to look for duplicate expression more defensively, using soft rather than explicit cast and also taking MemberExpression into account.

### DIFF
--- a/src/EFCore.Specification.Tests/Query/ComplexNavigationsOwnedQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/ComplexNavigationsOwnedQueryTestBase.cs
@@ -354,6 +354,26 @@ namespace Microsoft.EntityFrameworkCore.Query
         {
         }
 
+        public override void Include_collection_with_multiple_orderbys_member()
+        {
+        }
+
+        public override void Include_collection_with_multiple_orderbys_property()
+        {
+        }
+
+        public override void Include_collection_with_multiple_orderbys_methodcall()
+        {
+        }
+
+        public override void Include_collection_with_multiple_orderbys_complex()
+        {
+        }
+
+        public override void Include_collection_with_multiple_orderbys_complex_repeated()
+        {
+        }
+
         protected override IQueryable<Level1> GetExpectedLevelOne()
             => ComplexNavigationsData.SplitLevelOnes.AsQueryable();
 
--- a/src/EFCore/Query/Internal/IncludeCompiler.CollectionQueryModelRewritingExpressionVisitor.cs
+++ b/src/EFCore/Query/Internal/IncludeCompiler.CollectionQueryModelRewritingExpressionVisitor.cs
@@ -512,12 +512,29 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                                     .FindIndex(
                                         e =>
                                             {
-                                                var propertyExpression = (MethodCallExpression)((NullConditionalExpression)e.RemoveConvert()).AccessOperation;
-                                                var properyQsre = (QuerySourceReferenceExpression)propertyExpression.Arguments[0];
-                                                var propertyName = (string)((ConstantExpression)propertyExpression.Arguments[1]).Value;
+                                                var expressionWithoutConvert = e.RemoveConvert();
+                                                var projectionExpression = (expressionWithoutConvert as NullConditionalExpression)?.AccessOperation
+                                                    ?? expressionWithoutConvert;
 
-                                                return properyQsre.ReferencedQuerySource == memberQsre.ReferencedQuerySource
-                                                       && propertyName == memberExpression.Member.Name;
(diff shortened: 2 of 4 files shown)


src/EFCore.Specification.Tests/Query/ComplexNavigationsOwnedQueryTestBase.cs
```diff
@@ -354,6 +354,26 @@ public override void Select_optional_navigation_property_string_concat()
         {
         }
 
+        public override void Include_collection_with_multiple_orderbys_member()
```
smitpatel: Do we know why these need to be disabled?

maumar: collection navigations are not supported in owned entites

src/EFCore/Query/Internal/IncludeCompiler.CollectionQueryModelRewritingExpressionVisitor.cs
```diff
@@ -512,12 +512,30 @@ in fromQueryModel.BodyClauses.OfType<OrderByClause>().ToArray())
...
-                                                var propertyExpression = (MethodCallExpression)((NullConditionalExpression)e.RemoveConvert()).AccessOperation;
-                                                var properyQsre = (QuerySourceReferenceExpression)propertyExpression.Arguments[0];
-                                                var propertyName = (string)((ConstantExpression)propertyExpression.Arguments[1]).Value;
-
-                                                return properyQsre.ReferencedQuerySource == memberQsre.ReferencedQuerySource
-                                                       && propertyName == memberExpression.Member.Name;
+                                                var projectionExpression = (e.RemoveConvert() as NullConditionalExpression)?.AccessOperation ?? e.RemoveConvert();
+                                                if (projectionExpression is MethodCallExpression methodCall)
+                                                {
+                                                    if (methodCall.Method.IsEFPropertyMethod())
```
smitpatel: merge If

src/EFCore/Query/Internal/IncludeCompiler.CollectionQueryModelRewritingExpressionVisitor.cs
```diff
@@ -512,12 +512,30 @@ in fromQueryModel.BodyClauses.OfType<OrderByClause>().ToArray())
...
+                                                    var projectionMemberQsre = (QuerySourceReferenceExpression)projectionMemberExpression.Expression;
+
+                                                    return projectionMemberQsre.ReferencedQuerySource == memberQsre.ReferencedQuerySource
+                                                        && projectionMemberExpression.Member.Name == memberExpression.Member.Name;
+                                                }
+
+                                                return false;
                                             });
                         }
                         else
```
anpete: Shouldn't we just end up in this else branch instead (by making the first if conditions more explicit)?

maumar: yup, just made this change

maumar:

src/EFCore/Query/Internal/IncludeCompiler.CollectionQueryModelRewritingExpressionVisitor.cs
```diff
@@ -512,12 +512,26 @@ in fromQueryModel.BodyClauses.OfType<OrderByClause>().ToArray())
                                     .FindIndex(
                                         e =>
                                             {
-                                                var propertyExpression = (MethodCallExpression)((NullConditionalExpression)e.RemoveConvert()).AccessOperation;
-                                                var properyQsre = (QuerySourceReferenceExpression)propertyExpression.Arguments[0];
-                                                var propertyName = (string)((ConstantExpression)propertyExpression.Arguments[1]).Value;
-
-                                                return properyQsre.ReferencedQuerySource == memberQsre.ReferencedQuerySource
-                                                       && propertyName == memberExpression.Member.Name;
+                                                var projectionExpression = (e.RemoveConvert() as NullConditionalExpression)?.AccessOperation ?? e.RemoveConvert();
```
smitpatel: Save result of Remove Convert

maumar: done

Issue #8899, opened 2017-06-19
Databinding: Guide users to not bind against queries directly


divega:
Currently `DbSet` implements `IListSource` to throw an informative message pointing to `DbSet.Local.ToBindingList()` and `DbSet.Local.ToObservableCollection` but our query objects do not. Binding directly to queries can cause double roundtrips and other issues, so we should consider having the implementation of `IListSource` there was well as we did in EF6.

Issue #8898, opened 2017-06-19
Databinding: Guide users to not bind against DbSet.Local directly


divega:
In EF Core `DbSet.Local` is more efficient than in EF6 but doesn't have stable order, so it is not good for databinding. For databinding you are supposed to call `ToObservableCollection` (for the WPF flavor) or `ToBindingList` (for the WinForms flavor). 

This is already in the XML documentation but nothing prevents users from binding directly to `DbSet.Local`.

We should consider implementing `IListSource.GetList()` to throw an informative exception like we do for query, but we also need to verify that doing so will help for WPF.

ajcvickers:
Leaving open to answer the WPF question.

divega:
Note for triage: This issue had one question pending: I tried binding in a WPF to an observable collection that implements IListSource and GetList was called. 

Given that in 3.0 we want to have better WPF/WinForms support, I would like to discuss if we want to do this (if we decide not to do it, maybe we should close the issue?). 

On the other hand, this reminded me of https://github.com/aspnet/EntityFrameworkCore/issues/14231, which was recently reported and says DbSet.Local isn't good for iterating either... I can't remember what DbSet.Local is good for :trollface:.

ajcvickers:
I started to work on this, but it turns out that we already did this for 2.0. See https://github.com/aspnet/EntityFrameworkCore/pull/9003

The only reason this issue was still open was to check if WPF calls the method, which has now been done., so closing.

OlegFj17:
Sorry to interrupt you. Maybe you can help.. If I bind Local.ToObservableCollection() to for exapmle DataGrid it works fine. But do I understand correctly that with this context I can only make synchronous queries to the database? Because otherwise Local collection changed from nonUI thread and an exception is thrown about that ICollectionView cannot be changed from non UI thread..  Thank you!

ajcvickers:
@roji I think this is a consequence of `ConfigureAwait false`. Fixup can now happen in a different thread, and this then means that events in the observable collection (or other notifications) are triggering calls into the U.I. on the non-U.I. thread.

/cc @AndriySvyryd

roji:
@ajcvickers we did do #23971 to preserve the synchronization context when accepting changes - is there something here with Local.ToObservableCollection that goes beyond that (am not too familiar with that...)?

ajcvickers:
@roji Yes, it gets called when materializing entities/doing fixup after _queries_.

roji:
[...]

PR #9003, opened 2017-06-28
Implement IListSource on LocalView and EntityQueryable to further direct data binding

ajcvickers:
Issues #8898 #8899

### DIFF
--- a/src/EFCore/ChangeTracking/LocalView.cs
+++ b/src/EFCore/ChangeTracking/LocalView.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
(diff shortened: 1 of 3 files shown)


ajcvickers:
No, would it change the decision to do this depending on the answer?

divega:
> No, would it change the decision to do this depending on the answer?

No, I think this PR is fine regardless. 

What I am still concerned about is the decision to deviate from what we had in EF6. I.e. .Local used to returned an ordered collection that could be used directly for WPF-like databinding but now you need another call to `ToObservableCollection()` which won't be discoverable unless in binding to these objects directly causes the exception.

divega:
That is why #8898 says 

> ... but we also need to verify that doing so will help for WPF.

ajcvickers:
I'll leave #8898 open after merging this and we can bring it back to triage to discuss the priority of doing this investigation.

PR #9008, opened 2017-06-28
Remove NetStandard.Library.NetFramework and update Moq

ryanbrandenburg:

PR #9011, opened 2017-06-29
Update dependencies.props

pranavkm:
* Update Moq to 4.7.49.
* Add NETStandardImplicitPackageVersion


smitpatel:
Also #9008 
Upgrading Moq doesn't work for us.

pranavkm:
@smitpatel have you filed a bug for this in Moq? We can't stay on a version that produces restore warnings forever.

PR #9017, opened 2017-06-29
async transaction APIs

### LINKED ISSUE #9015: Support Async Transaction APIs
@divega opened this to CoreFX a while ago: https://github.com/dotnet/corefx/issues/10975 but it doesn't look like there has been any movement.

Our ADO.NET provider, MySqlConnector has [BeginTransactionAsync](https://mysql-net.github.io/MySqlConnector/api/mysql-connection/), [CommitAsync, and RollbackAsync](https://mysql-net.github.io/MySqlConnector/api/mysql-transaction/)

We like to use them in [Pomelo.EntityFrameworkCore.MySql](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql) to avoid running Sync-Over-Async I/O and cut down on thread pool lockups.

EF does not lend itself well to supporting these Async methods.  We have to override every single method that begins, commits, or rolls back a transaction in our provider.

I think that EF should add a class that allows for the provider to define whether or not it supports Async transaction APIs, and use those instead of the Sync ones if it does.

caleblloyd:
- Fixes #9015
- I am unfamiliar with your test suite; please tell me where tests would go
- Please put design related comments in the issue #9015

### DIFF
--- a/src/EFCore.Relational/Migrations/Internal/MigrationCommandExecutor.cs
+++ b/src/EFCore.Relational/Migrations/Internal/MigrationCommandExecutor.cs
@@ -55,6 +55,18 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Internal
 
                     transaction?.Commit();
                 }
+                catch
+                {
+                    try
+                    {
+                        transaction?.Rollback();
+                    }
+                    catch
+                    {
+                        // if the connection was lost, rollback command will fail.  prefer to throw original exception in that case
+                    }
+                    throw;
+                }
                 finally
                 {
                     transaction?.Dispose();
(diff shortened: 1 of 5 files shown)


ajcvickers:
@caleblloyd Unfortunately, it's really too late in the release for us to consider this for 2.0. We will think about this post-2.0 and make a decision as to whether we want to pursue the EF APIs without the ADO.NET APIs being present. I'm going to close this PR for now, but we'll get back to you when we decide where we want to go here. We appreciate the suggestion and the time put into this.

caleblloyd:
No worries I wanted to provide a sample API of what this would look like and it was easier to do in a PR than write up in the issue. Let's keep the issue open and people can reference this PR for a future implementation option.

Issue #8615, opened 2017-05-26
Investigate flaky tests


smitpatel:
Command time in `F1SqlServerFixture` had to be increased. `DataBindingSqlServerTest` failed randomly with this exception
```
System.Data.SqlClient.SqlException : Timeout expired.  The timeout period elapsed prior to completion of the operation or the server is not responding.
---- System.ComponentModel.Win32Exception : The wait operation timed out
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
   at System.Data.SqlClient.SqlDataReader.get_MetaData()
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString)
   at System.Data.SqlClient.SqlCommand.RunExecuteReaderTds(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, Boolean async, Int32 timeout, Task& task, Boolean asyncWrite, SqlDataReader ds)
   at System.Data.SqlClient.SqlCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.Execute(IRelationalConnection connection, DbCommandMethod executeMethod, IReadOnlyDictionary`2 parameterValues, Boolean closeConnection) in C:\b\w\437c4411cf4e11ab\src\EFCore.Relational\Storage\Internal\RelationalCommand.cs:line 271
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.ExecuteReader(IRelationalConnection connection, IReadOnlyDictionary`2 parameterValues) in C:\b\w\437c4411cf4e11ab\src\EFCore.Relational\Storage\Internal\RelationalCommand.cs:line 134
[...]

PR #9020, opened 2017-06-29
Disable SqlServerDatabaseCreatorTest on CI

AndriySvyryd:
Refactor SqlServerDatabaseCreatorTest

Part of #8615

### DIFF
--- a/test/EFCore.SqlServer.FunctionalTests/SqlServerDatabaseCreatorTest.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/SqlServerDatabaseCreatorTest.cs
@@ -8,445 +8,448 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore.Infrastructure;
-using Microsoft.EntityFrameworkCore.TestUtilities.Xunit;
-using Microsoft.EntityFrameworkCore.Utilities;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Storage.Internal;
+using Microsoft.EntityFrameworkCore.TestUtilities.Xunit;
+using Microsoft.EntityFrameworkCore.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
 
+// ReSharper disable InconsistentNaming
 namespace Microsoft.EntityFrameworkCore
 {
-    public class SqlServerDatabaseCreatorExistsTest
+    // These tests are disabled in TeamCity as they are flaky, see #8615
+    public class SqlServerDatabaseCreatorTest
     {
-        [Fact]
-        public async Task Exists_returns_false_when_database_doesnt_exist()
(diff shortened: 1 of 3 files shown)


test/EFCore.SqlServer.FunctionalTests/SqlServerDatabaseCreatorTest.cs
```diff
@@ -4,449 +4,464 @@
...
 using Microsoft.EntityFrameworkCore.Storage.Internal;
+using Microsoft.EntityFrameworkCore.TestUtilities.Xunit;
+using Microsoft.EntityFrameworkCore.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
 
 namespace Microsoft.EntityFrameworkCore
 {
-    public class SqlServerDatabaseCreatorExistsTest
+    [SuppressMessage("ReSharper", "InconsistentNaming")]
```
anpete: When did this become a thing? For consistency use ```// ReSharper disable``` syntax

test/EFCore.SqlServer.FunctionalTests/SqlServerDatabaseCreatorTest.cs
```diff
@@ -4,449 +4,464 @@
...
+    public class SqlServerDatabaseCreatorTest
     {
-        [Fact]
-        public async Task Exists_returns_false_when_database_doesnt_exist()
+        public class Exists
         {
-            await Exists_returns_false_when_database_doesnt_exist_test(async: false, file: false);
-        }
+            [ConditionalFact]
+            [SqlServerCondition(SqlServerCondition.IsNotTeamCity)]
```
anpete: Curious, can these just be put on the class itself? (DRY)

AndriySvyryd: Even if we could they have different flags on different test

anpete: XUnit could combine :-)

anpete: Or we could

AndriySvyryd: That would add complexity for little value

anpete: Disagree. Elegant and DRY.

smitpatel: `IsNoTeamCity` on class.
and all tests would have `ConditionalFact`

AndriySvyryd: Ok, it works if I put it on the immediate class, so it's not completely DRY, but better

PR #9025, opened 2017-06-29
Improve literal generation for various types

### LINKED ISSUE #8259: Query: ulong literal throws RuntimeBinderException
```C#
var ulongValue = 1ul;
Enumerable.ToList(
    from _ in db.Dual
    where ulongValue == 1
    select _);
```

```
Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: The call is ambiguous between the following methods or properties: 'Microsoft.EntityFrameworkCore.Storage.RelationalSqlGenerationHelper.GenerateLiteralValue(float)' and 'Microsoft.EntityFrameworkCore.Storage.RelationalSqlGenerationHelper.GenerateLiteralValue(decimal)'
   at CallSite.Target(Closure , CallSite , RelationalSqlGenerationHelper , Object )
   at CallSite.Target(Closure , CallSite , RelationalSqlGenerationHelper , Object )
   at Microsoft.EntityFrameworkCore.Storage.RelationalSqlGenerationHelper.GenerateLiteral(Object value, RelationalTypeMapping typeMapping)
   at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.VisitConstant(ConstantExpression constantExpression)
   at System.Linq.Expressions.ConstantExpression.Accept(ExpressionVisitor visitor)
   at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
   at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.VisitBinary(BinaryExpression binaryExpression)
   at System.Linq.Expressions.BinaryExpression.Accept(ExpressionVisitor visitor)
   at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
   at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.GeneratePredicate(Expression predicate)
   at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.VisitSelect(SelectExpression selectExpression)
   at Microsoft.EntityFrameworkCore.Query.Expressions.SelectExpression.Accept(ExpressionVisitor visitor)
   at Remotion.Linq.Parsing.ThrowingExpressionVisitor.Visit(Expression expression)
   at Microsoft.EntityFrameworkCore.Query.Sql.DefaultQuerySqlGenerator.GenerateSql(IReadOnlyDictionary`2 parameterValues)
   at Microsoft.EntityFrameworkCore.Query.Internal.ShaperCommandContext.GetRelationalCommand(IReadOnlyDictionary`2 parameters)
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryingEnumerable.Enumerator.BufferlessMoveNext(Boolean buffer)
   at Microsoft.EntityFrameworkCore.Storage.Internal.NoopExecutionStrategy.Execute[TState,TResult](Func`2 operation, Func`2 verifySucceeded, TState state)
   at Microsoft.EntityFrameworkCore.ExecutionStrategyExtensions.Execute[TState,TResult](IExecutionStrategy strategy, Func`2 operation, TState state)
   at Microsoft.EntityFrameworkCore.Query.Internal.QueryingEnumerable.Enumerator.MoveNext()
   at Microsoft.EntityFrameworkCore.Query.QueryMethodProvider.<_ShapedQuery>d__3`1.MoveNext()
   at Microsoft.EntityFrameworkCore.Query.Internal.LinqOperatorProvider.<_TrackEntities>d__17`2.MoveNext()
   at Microsoft.EntityFrameworkCore.Query.Internal.LinqOperatorProvider.ExceptionInterceptor`1.EnumeratorExceptionInterceptor.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
```

smitpatel:
Resolve #8259 Self-contained type mappings so no more binding errors since we have mapping for every clr type.

Resolves https://github.com/aspnet/EntityFramework/issues/8270
After much discussion with @bricelam we ended up with
float -> Use `ToString("R")` which is roundtrip-able. 
double -> R faces issue on x64 so docs suggest using "G17"
Removed "E0" suffix for relational provider.

Resolves https://github.com/aspnet/EntityFramework/issues/8905
float in SqlServer -> Always cast to real while generating literal
double in SqlServer -> if the literal does not have E then append E0 (to make it double literal)
Since Long literals are used in migrations too we cannot cast them always. It has to be done based on functions in query.

### DIFF
--- a/src/EFCore.SqlServer/Storage/Internal/SqlServerFloatTypeMapping.cs
+++ b/src/EFCore.SqlServer/Storage/Internal/SqlServerFloatTypeMapping.cs
@@ -2,31 +2,25 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Data;
-using System.Globalization;
 using JetBrains.Annotations;
 
-namespace Microsoft.EntityFrameworkCore.Storage
+namespace Microsoft.EntityFrameworkCore.Storage.Internal
 {
     /// <summary>
-    ///     <para>
-    ///         Represents the mapping between a .NET <see cref="float" /> type and a database type.
-    ///     </para>
-    ///     <para>
-    ///         This type is typically used by database providers (and other extensions). It is generally
-    ///         not used in application code.
-    ///     </para>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
     /// </summary>
-    public class FloatTypeMapping : RelationalTypeMapping
+    public class SqlServerFloatTypeMapping : FloatTypeMapping
     {
         /// <summary>
-        ///     Initializes a new instance of the <see cref="FloatTypeMapping" /> class.
+        ///     Initializes a new instance of the <see cref="SqlServerFloatTypeMapping" /> class.
         /// </summary>
         /// <param name="storeType"> The name of the database type. </param>
         /// <param name="dbType"> The <see cref="DbType" /> to be used. </param>
-        public FloatTypeMapping(
+        public SqlServerFloatTypeMapping(
             [NotNull] string storeType,
             [CanBeNull] DbType? dbType = null)
-            : base(storeType, typeof(float), dbType)
+            : base(storeType, dbType)
         {
         }
 
@@ -37,7 +31,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
         /// <param name="size"> The size of data the property is configured to store, or null if no size is configured. </param>
         /// <returns> The newly created mapping. </returns>
         public override RelationalTypeMapping Clone(string storeType, int? size)
-            => new FloatTypeMapping(storeType, DbType);
+            => new SqlServerFloatTypeMapping(storeType, DbType);
 
         /// <summary>
         ///     Generates the SQL representation of a literal value.
@@ -48,7 +42,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
         /// </returns>
         protected override string GenerateNonNullSqlLiteral(object value)
         {
-            return ((float)value).ToString("R", CultureInfo.InvariantCulture);
+            return $"CAST({base.GenerateNonNullSqlLiteral(value)} AS {StoreType})";
         }
     }
 }
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -588,7 +588,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
         {
             return _typeMapping != null
                    && (value == null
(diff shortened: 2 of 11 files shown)


src/EFCore.SqlServer/Storage/Internal/SqlServerFloatTypeMapping.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <summary>
+        ///     Generates the SQL representation of a literal value.
+        /// </summary>
+        /// <param name="value">The literal value.</param>
+        /// <returns>
+        ///     The generated string.
+        /// </returns>
+        protected override string GenerateNonNullSqlLiteral(object value)
+        {
+            return $"CAST({base.GenerateNonNullSqlLiteral(value)} AS {StoreType})";
```
bricelam: Type names can contain spaces. (e.g. `NATIONAL CHARACTER VARYING`) Do we need to quote these in square braces?

smitpatel: We could. Or its just YAGNI may be.
We are just special casing "real" on SqlServer here. Instead of using StoreType, we can just write real itself. In any case if someone decides to override this, they may probably end up providing own SqlLiteral generation.

Let me know what should we use here. I am fine either way. Bit preference for no braces for saving few characters in Sql

bricelam: Looks like `ExplicitCastExpression` doesn't use braces either. Leave 'em off.

lajones: What does the generated "HasDefaultValueSql()" call end up looking like when we do RevEng on a SQL Server REAL column which has a default value? It looks like it may look very strange. Do we always need the `CAST()` call or only in certain (query?) situations? And if so should we provide a separate method to produce that?

smitpatel: `HasDefaultValueSql` puts up the SQL as is from database. We never generate `HasDefaultValue`.

These is how we can actually generate a real literal in SqlServer. else it will be float literal always. While it may not cause issue in some cases but once you do any operation with any other real column, the float literal will use the higher type, regardless of where it is being used. Since the point of this method is to _generate real literal_, the explicit casting is right way to do it.

The only issue with bigint is `Sequence` creation DDL where even if the StartValue is any numeric type, it will be converted to the type of the `Sequence`, perhaps we should special case migration.

lajones: OK. Good. Glad it doesn't affect that.

Issue #8921, opened 2017-06-21
Detached entities appear in ChangeTracker.Entries()


Wain123:
```
var u = ctx.Posts.First();
ctx.Entry(u).State = EntityState.Detached;
Console.WriteLine(ctx.ChangeTracker.Entries().Count());
ctx.Entry(u);
Console.WriteLine(ctx.ChangeTracker.Entries().Count());
```
This code outputs "0 1". Simply looking at u's entry after detaching u causes the entry to appear in the ChangeTracker, even though u is detached. Is this working as intended?

ajcvickers:
Note for triage: confirmed repro on 1.1.2.

PR #9032, opened 2017-06-30
Stop returning detached entries from the ChangeTracker

ajcvickers:
Issue #8921


src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -377,7 +377,8 @@ private IIdentityMap FindIdentityMap(IKey key)
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public virtual IEnumerable<InternalEntityEntry> Entries => _entityReferenceMap.Values
-            .Concat(_dietReferenceMap.Values.SelectMany(e => e.Values));
+            .Concat(_dietReferenceMap.Values.SelectMany(e => e.Values))
```
smitpatel: reference to diet 🙁

PR #9034, opened 2017-07-01
Make sure the old dependent is removed when the new one is reference stealing using navigations and/or FK properties

### LINKED ISSUE #8137: One-to-one relationship with bad data behaviors differently for bi-directional verses uni-directional relationships
I am building some tests/examples for one-to-one relationships, and I found that EF Core differs in its handling of the relationship if an inverse navigational relationship exists or not:
1. One-to-one relationship **without** a inverse navigational relationship correctly saves data
2. One-to-one relationship **with** a inverse navigational relationship detaches data that has a duplicate one-to-one entity

I found this with a test to show that duplicate one-on-one entities should throw an exception: in case 1 an exception is thrown if I have a duplicate. In case 2 no exception is thrown.

**UPDATE:** I have found change in behaviour is due to all but one of the entities that has a duplicate one-to-one relationship are set to `Detached` after SaveChanges is called. They have not been saved to the database, so they don't trigger the exception on a UNIQUE constraint.

*NOTE: I agree this seems very strange and it could be my mistake, but I have looked at it and can't see anything wrong with my code. Let me know if you spot anything.*

### Steps to reproduce
My main entity
```c#
public class Attendee 
{
    public int AttendeeId { get; set; }

    public int TicketId { get; set; } 
    public Ticket Ticket { get; set; }
}
```
My sub-entity
```c#
public class Ticket
{
    public enum TicketTypes : byte { Guest = 0, VIP = 1, Staff = 3}

    public int TicketId { get; set; }
    public TicketTypes TicketType { get; set; }

    public Attendee Attendee { get; set; }
}
```
My configuration 
NOTE: The configuration shown does NOT work (case 2), but if I replace the `WithOne(p => p.Attendee)` with `.WithOne()` then it works.
```c#
public static class AttendeeConfig
{
    public static void Configure
        (this EntityTypeBuilder<Attendee> entity)
    {
        entity.HasOne(p => p.Ticket) //#A
            .WithOne(p => p.Attendee)
            //.WithOne()
            .HasForeignKey<Attendee>(p => p.TicketId);
    }
}
```
My DbContext
```c#
public class Chapter07DbContext : DbContext
{
    public DbSet<Employee> Employees { get; set; }
    public DbSet<EmployeeShortFk> EmployeeShortFks { get; set; }
    public DbSet<Person> People { get; set; }
    public DbSet<LibraryBook> LibraryBooks { get; set; }

    //One-to-One versions
    public DbSet<Attendee> Attendees { get; set; }
    public DbSet<Ticket> Tickets { get; set; }

    public Chapter07DbContext(
        DbContextOptions<Chapter07DbContext> options)
        : base(options)
    { }

    protected override void OnModelCreating
        (ModelBuilder modelBuilder)
    {
        modelBuilder.Entity<Attendee>().Configure();
    }
}
```
My unit test is
```c#
[Fact]
public void TestOption1OneToOneDuplicateTicketBad()
{
    //SETUP
    using (var context = new Chapter07DbContext(
        SqliteInMemory.CreateOptions<Chapter07DbContext>()))
    {
        {
            var logs = new List<string>();
            SqliteInMemory.SetupLogging(context, logs);

            //ATTEMPT
            var dupTicket = new Ticket {TicketType = Ticket.TicketTypes.Guest};
            var attendees = new List<Attendee>
            {
                new Attendee {Name = "Person1", Ticket = dupTicket, },
                new Attendee {Name = "Person2", Ticket = dupTicket, },
                new Attendee {Name = "Person3", Ticket = new Ticket()}
            };
//The following code has been added to diagnose the issue
            context.AddRange(attendees);
            ListStates(context, "After AddRange", attendees);
            context.SaveChanges();
            ListStates(context, "After SaveChanges", attendees);

            //var ex = Assert.Throws<DbUpdateException>(() => context.SaveChanges());

            //VERIFY
            //ex.InnerException.Message.ShouldEqual("SQLite Error 19: 'UNIQUE constraint failed: Attendees.TicketId'.");
        }
    }
}

private void ListStates(DbContext context, string message,  List<Attendee> entities)
{
    var line = message + ": " + string.Join(", ", entities.Select(x => context.Entry(x).State));
    _output.WriteLine(line);
}
```
### Notes
The output of state of the Attendee entities in case 2 (the failure state) are:
```
After AddRange: Added, Added, Added
After SaveChanges: Detached, Unchanged, Unchanged
```

### Further technical details
EF Core version: 1.1.1
Database Provider: Microsoft.EntityFrameworkCore.Sqlite
Operating system: Windows 10
IDE: Visual Studio 2017

AndriySvyryd:
If the FK was non-identifying make sure the properties are set to null instead of the dependent being removed.

Fixes #8137

### DIFF
--- a/test/EFCore.Tests/ChangeTracking/Internal/FixupTest.cs
+++ b/test/EFCore.Tests/ChangeTracking/Internal/FixupTest.cs
@@ -9,6 +9,8 @@ using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Storage;
 using Xunit;
 
+// ReSharper disable InconsistentNaming
+// ReSharper disable AccessToDisposedClosure
 namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
 {
     public class FixupTest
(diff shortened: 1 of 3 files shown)


test/EFCore.Tests/ChangeTracking/Internal/FixupTest.cs
```diff
@@ -3769,8 +4110,6 @@ private void AssertFixup(DbContext context, Action asserts)
             asserts();
             context.ChangeTracker.DetectChanges();
             asserts();
-            context.ChangeTracker.DetectChanges();
```
ajcvickers: Leave this in please. This was added because there were cases where calling DetectChanges multiple times would result in bugs showing up.

PR #9036, opened 2017-07-01
Use nameof in StoredGeneratedTestBase

roji:
Just because it's no longer to possible to see member names in double-quotes... :)


ajcvickers:
Merged. Thanks @roji

Issue #8909, opened 2017-06-20
QueryCache misses for dynamically created queries with ConstantExpresssions or nested MemberExpressions from closure


Tasteful:
I got a lot of dupplicated entries in the `MemoryCache` that origin from EF and many of them contains the same sql query and is generated from the exact same code.

Example the following that generate an SQL query that is duplicated in cache more than 16 000 times.

![image](https://user-images.githubusercontent.com/357589/27338776-62f0cf4a-55d6-11e7-94ec-8df9f9e8e062.png)

When I checking the entity in the `MemoryCache` we can se a lot of repeated entries that is 14 338 bytes each but contains the same query. 

![image](https://user-images.githubusercontent.com/357589/27339004-df30a706-55d6-11e7-8a52-7b1c833c0214.png)

The only "special" that I can remember that I have done is that the query is built with `System.Linq.Expressions.Expression` that is added to the query with `.Where(expression)`.

What is the best approach to find the reason for this cache misses?

### Further technical details
EF Core version: 1.1.1
Database Provider: Microsoft.EntityFrameworkCore.SqlServer
Operating system: Windows 10
IDE: Visual Studio 2017

Tasteful:
It seems that is the usage of `System.Linq.Expressions.Expression` that is bad in my usage case.

- Is there any reason to not the string representation as the cache key?
- I can't find a way to change the sliding timeout for the cache entry per query, is that possible?

Tasteful:
Is it the same when using `ICollection<T>.Contains`?

example 
```
var values = new List<Guid>();
values.Add(Guid.NewGuid());
values.Add(Guid.NewGuid());
values.Add(Guid.NewGuid());
dbContext.Table1.Where(x => values.Contains(x.Id)).ToList();
```
For the contains it generate an SQL with the values in the statement, not as parameters. Will the cache handle that to replace the parameters?

anpete:
@Tasteful Sorry you are hitting this. Can you post the complete LINQ query that causes the cache miss?

Tasteful:
@anpete I will try to make a breakout of relevant code parts.

smitpatel:
> EF Core version: 1.1.1

Any reason you haven't upgraded to 1.1.2?
You filed issue in past about memory leak which was fixed in 1.1.2 https://github.com/aspnet/EntityFramework/issues/8063 😄

anpete:
Oh yeah, definitely try that 😄

Tasteful:
[...]

Issue #9014, opened 2017-06-29
Wrap Multiple Eager Loaded Queries that leverage Include in Repeatable Read transactions


caleblloyd:
I am opening this issue against the behavior observed in EF Core 1.0.  I have not had time to investigate changes in EF Core 2.0 so if by chance this has been added already please let me know.

Eager Loading with `.Include` generates multiple SQL queries.  They are not inside of a transaction, however.  This means that between the queries, new records could be inserted affecting ranges in the `WHERE` clause and causing incorrect data to get loaded.

Pomelo.EntityFrameworkCore.MySql solves this in 1.0 by wrapping queries that have in `Include` clause in a RepeatableRead transaction.  I think that this logic should be moved into EFCore.Relational.  It's pretty important for consistency's sake.

Here's an example of how we do it in 1.0:

- `HasInclude` flag in [MySqlQueryContext](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/src/Pomelo.EntityFrameworkCore.MySql/Query/Internal/MySqlQueryContext.cs#L16)
- Transaction logic in [MySqlQueryingEnumerable](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/src/Pomelo.EntityFrameworkCore.MySql/Query/Internal/MySqlQueryingEnumerable.cs#L47-L54) and [MySqlAsyncQueryingEnumerable](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/src/Pomelo.EntityFrameworkCore.MySql/Query/Internal/MySqlAsyncQueryingEnumerable.cs#L42-L55)

The transaction isolation level should be `RepeatableRead`

ajcvickers:
@caleblloyd Can you post a specific example of how the wrong data is loaded? i.e. The SQL that gets generated and what inserts happen between the SQL statements that cause the data to be wrong?

caleblloyd:
Sure.  Consider the following query that fetches a blog and it's associated posts.  Assume that `id` is a randomly assigned GUID.

**LINQ Query:**

```
await _db.Blogs.Include(m => m.Posts).OrderByDescending(m => m.Id).Take(10).ToListAsync()
```

**Generated SQL:**

```
info: Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommandBuilderFactory[1]
[...]

PR #9048, opened 2017-07-02
Fix parameter extracting of already parameter expressions

Tasteful:
When extracting parameters for caching the already created parameter expression is returnet directly instead of extracting the parameters.

Fixing #8909

@anpete I'm not finding any existing test-cases for `ParameterExtractingExpressionVisitor`, do you have any test-class where I can add more test in? Or can a validation for hashcode calculation be added to example the https://github.com/aspnet/EntityFramework/blob/f45292f2fd1a0edc2c364dccac00cb242b6808ae/src/EFCore.Specification.Tests/Query/QueryTestBase.cs#L4144 method?

### DIFF
--- a/src/EFCore.Specification.Tests/Query/QueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/QueryTestBase.cs
@@ -14,6 +14,8 @@ using Microsoft.EntityFrameworkCore.TestModels.Northwind;
 using Microsoft.EntityFrameworkCore.TestUtilities.Xunit;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
 
 // ReSharper disable InconsistentNaming
 // ReSharper disable AccessToDisposedClosure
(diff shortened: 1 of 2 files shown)


Tasteful:
Added hashcode comparison in the `AssertQuery` methods to also verify that all hashcodes is calculated correctly. Not sure if this is a good way to catch this type of errors to not need to write all the different tests-cases again or duplicate the code.

ajcvickers:
@smitpatel Can you please take a look at #8909 and this PR to try to get it in this week?

smitpatel:
I looked at this PR & the associated issue. I will need to look more into kind of ETs are generated in different scenarios mentioned here https://github.com/aspnet/EntityFramework/issues/8909#issuecomment-311712575 & understand how above fix helps that case.

ajcvickers:
@smitpatel Do you think you will be able to get that done this week?

smitpatel:
I will be able to finish investigation on the issue this week. Perhaps if the fix is easy (or above) then I can get it done this week.

Tasteful:
The real problem was not the usage of the ConstantExpression but the type of return value for ConstantExpression.

If the value for ConstantExpression was of type ParameterExpression the expression was returned directly (if I remember correctly all parameters was inlined in generated sql). For all other types of values it was creating an parameter that was used in sql.

Tasteful:
The exampelcode here https://github.com/Tasteful/bugs/tree/query-cache-expression-not-working/MemoryUsage is also updated and executed first without modifications and after with the modification.

smitpatel:
Superseded by #9014 
I took the source code fix from this PR but added different testing.
More description on the issue

Issue #9004, opened 2017-06-28
Query: compilation error for queries with Concat on two entities and Count


maumar:
query:

```
context.Gears.Concat(context.Gears).Count();
```

exception:

```
Expression of type 'System.Collections.Generic.IEnumerable`1[Microsoft.EntityFrameworkCore.Storage.ValueBuffer]' cannot be used for parameter of type 'System.Collections.Generic.IEnumerable`1[System.Object]' of method 'System.Collections.Generic.IEnumerable`1[System.Object] Concat[Object](System.Collections.Generic.IEnumerable`1[System.Object], System.Collections.Generic.IEnumerable`1[System.Object])'
	at System.Linq.Expressions.Expression.ValidateOneArgument(MethodBase method, ExpressionType nodeKind, Expression arg, ParameterInfo pi)
	at System.Linq.Expressions.Expression.Call(MethodInfo method, Expression arg0, Expression arg1)
	C:\git\EntityFramework\src\EFCore\Query\ResultOperatorHandler.cs(531,0): at Microsoft.EntityFrameworkCore.Query.ResultOperatorHandler.HandleSetOperation(EntityQueryModelVisitor entityQueryModelVisitor, Expression secondSource, MethodInfo setMethodInfo)
	C:\git\EntityFramework\src\EFCore\Query\ResultOperatorHandler.cs(144,0): at Microsoft.EntityFrameworkCore.Query.ResultOperatorHandler.HandleConcat(EntityQueryModelVisitor entityQueryModelVisitor, ConcatResultOperator concatResultOperator)
	C:\git\EntityFramework\src\EFCore\Query\ResultOperatorHandler.cs(36,0): at Microsoft.EntityFrameworkCore.Query.ResultOperatorHandler.<>c.<.cctor>b__33_4(EntityQueryModelVisitor v, ResultOperatorBase r, QueryModel __)
	C:\git\EntityFramework\src\EFCore\Query\ResultOperatorHandler.cs(89,0): at Microsoft.EntityFrameworkCore.Query.ResultOperatorHandler.HandleResultOperator(EntityQueryModelVisitor entityQueryModelVisitor, ResultOperatorBase resultOperator, QueryModel queryModel)
[...]

Issue #8525, opened 2017-05-19
Query: compilation error for queries with navigations inside set result operators and Count


maumar:
query:

```
context.Gears.Select(g => g.Weapons.Concat(g.Weapons).Count())
```

throws:

```
System.ArgumentException : Expression of type 'System.Collections.Generic.IEnumerable`1[Microsoft.EntityFrameworkCore.Storage.ValueBuffer]' cannot be used for parameter of type 'System.Collections.Generic.IEnumerable`1[System.Object]' of method 'System.Collections.Generic.IEnumerable`1[System.Object] Concat[Object](System.Collections.Generic.IEnumerable`1[System.Object], System.Collections.Generic.IEnumerable`1[System.Object])'
	at System.Linq.Expressions.Expression.ValidateOneArgument(MethodBase method, ExpressionType nodeKind, Expression arg, ParameterInfo pi)
	at System.Linq.Expressions.Expression.Call(MethodInfo method, Expression arg0, Expression arg1)
	C:\git\EntityFramework\src\EFCore\Query\ResultOperatorHandler.cs(531,0): at Microsoft.EntityFrameworkCore.Query.ResultOperatorHandler.HandleSetOperation(EntityQueryModelVisitor entityQueryModelVisitor, Expression secondSource, MethodInfo setMethodInfo)
	C:\git\EntityFramework\src\EFCore\Query\ResultOperatorHandler.cs(144,0): at Microsoft.EntityFrameworkCore.Query.ResultOperatorHandler.HandleConcat(EntityQueryModelVisitor entityQueryModelVisitor, ConcatResultOperator concatResultOperator)
	C:\git\EntityFramework\src\EFCore\Query\ResultOperatorHandler.cs(36,0): at Microsoft.EntityFrameworkCore.Query.ResultOperatorHandler.<>c.<.cctor>b__33_4(EntityQueryModelVisitor v, ResultOperatorBase r, QueryModel __)
	C:\git\EntityFramework\src\EFCore\Query\ResultOperatorHandler.cs(89,0): at Microsoft.EntityFrameworkCore.Query.ResultOperatorHandler.HandleResultOperator(EntityQueryModelVisitor entityQueryModelVisitor, ResultOperatorBase resultOperator, QueryModel queryModel)
[...]

Issue #8524, opened 2017-05-19
Query: data corruption for queries with set result operators (Concat/Union etc) and collection navigations


maumar:
query:

```
context.Gears.Select(g => g.Weapons.Concat(g.Weapons)).ToList()
```

produces bad QM:

```
from Gear g in DbSet<Gear>
select 
    (from Weapon <generated>_1 in DbSet<Weapon>
    where Property([g], "FullName") ?= Property([<generated>_1], "OwnerFullName")
    where Property([g], "FullName") ?= Property([<generated>_1], "OwnerFullName")
    select [<generated>_1]).Concat({value(Microsoft.EntityFrameworkCore.Query.Internal.EntityQueryable`1[Microsoft.EntityFrameworkCore.Specification.Tests.TestModels.GearsOfWarModel.Weapon])})
```

collection expansion predicate is applied twice to the outer query source, but is not applied to the inner one at all. This produces incorrect results (since there is no filter on the inner query source at all)

maumar:
could be related to #8186

maumar:
fixed in e1d7b1659d5f67da9db24a31038e0ca14a8c27d5

PR #9057, opened 2017-07-03
Fix to several issues around set result operators (Contains etc)

maumar:
#8524 - Query: data corruption for queries with set result operators (Concat/Union etc) and collection navigations
#8525 - Query: compilation error for queries with navigations inside set result operators and Count
#9004 - Query: compilation error for queries with Concat on two entities and Count

Problem was that for set result operators we always need to materialize both sets, even if they are wrapped around scalar result operator (e.g. Count). This is needed because set result operators are currently performed on the client.

Also, we need to inject MaterializeCollectionNavigation calls into set result operators, so that the IQuerySource that is the argument to the set operator is wrapped in its own query source.

Also fixed some minor bugs in the MaterializeCollectionNavigation injector, which surfaced once the visitor started going inside SubQueryExpressions (it wasn't necessary before because we only needed to materialize top level collection).

### DIFF
--- a/src/EFCore.Specification.Tests/Query/GearsOfWarQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/GearsOfWarQueryTestBase.cs
@@ -1715,7 +1715,75 @@ namespace Microsoft.EntityFrameworkCore.Query
             }
         }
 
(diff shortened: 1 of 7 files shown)


maumar:
@smitpatel ping

PR #9072, opened 2017-07-05
More tests for ConcurrencyDetector

ajcvickers:
Issue #8305

Also, updated text for recursive call to OnConfiguring since in my investigation this was often the first error seen in invalid multi-threading scenarios.

### DIFF
--- a/src/EFCore.Specification.Tests/ConcurrencyDetectorTestBase.cs
+++ b/src/EFCore.Specification.Tests/ConcurrencyDetectorTestBase.cs
@@ -1,0 +1,123 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore.Infrastructure;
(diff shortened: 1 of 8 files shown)


src/EFCore.Specification.Tests/ConcurrencyDetectorTestBase.cs
```diff
@@ -0,0 +1,123 @@
...
+            await ConcurrencyDetectorTest(c => c.Products.ToListAsync(), async: true);
+        }
+
+        protected virtual async Task ConcurrencyDetectorTest(Func<NorthwindContext, Task> test, bool async)
+        {
+            using (var context = CreateContext())
+            {
+                context.Products.Add(new Product());
+
+                context.GetService<IConcurrencyDetector>().EnterCriticalSection();
```
AndriySvyryd: Rewrite this avoiding using `IConcurrencyDetector` directly to cover the real use case.

ajcvickers: Very hard to make it reliable...

AndriySvyryd: Filed https://github.com/aspnet/EntityFramework/issues/9074

PR #9075, opened 2017-07-05
Design: Continue on BuildWebHost errors

### LINKED ISSUE #9073: Tools: Ignore errors in design time DbContext discovery
Any errors are thrown by creating the service container or trying to resolve `DbContext` from it in an ASP.NET Core apps should be ignored so that we can proceed to discover any `IDesignTimeDbContextFactory<T>`s. 

From https://github.com/aspnet/EntityFramework/issues/9033#issuecomment-313213884, we believe we may have regressed.

cc @bricelam @ajcvickers

bricelam:
Fixes #9073

### DIFF
--- a/src/EFCore.Design/Properties/DesignStrings.resx
+++ b/src/EFCore.Design/Properties/DesignStrings.resx
@@ -333,4 +333,7 @@ Change your target project to the migrations project by using the Package Manage
   <data name="SequencesRequireName" xml:space="preserve">
     <value>Sequence name cannot be null or empty. Entity Framework cannot model a sequence that does not have a name.</value>
   </data>
+  <data name="InvokeBuildWebHostFailed" xml:space="preserve">
+    <value>An error occurred while calling method 'BuildWebHost' on class '{startupClass}'. Continuing without the application service provider. Error: {error}</value>
+  </data>
 </root>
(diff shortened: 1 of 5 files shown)


src/EFCore.Design/Properties/DesignStrings.resx
```diff
@@ -333,4 +333,7 @@ Change your target project to the migrations project by using the Package Manage
   <data name="SequencesRequireName" xml:space="preserve">
     <value>Sequence name cannot be null or empty. Entity Framework cannot model a sequence that does not have a name.</value>
   </data>
+  <data name="InvokeBuildWebHostFailed" xml:space="preserve">
+    <value>An error occurred while calling method 'BuildWebHost' on class '{startupClass}'. Error: {error}</value>
```
ajcvickers: Can we say something that indicates it was ignored and we continued?

Maybe something like 
```
An error was ignored while calling method 'BuildWebHost' on class '{startupClass}'. Error: {error}
```
@divega?

divega: @ajcvickers agreed. Maybe `An error occurred but was ignored...`, or whatever make the most sense in English.

bricelam: How does this sound?

> An error occurred while calling method 'BuildWebHost' on class '{programClass}'. **Continuing without the application service provider.** Error: {error}

divega: I don’t have a strong preference.

PR #9080, opened 2017-07-06
Configure logging internally to not filter out Debug messages

ajcvickers:
Issue #9083

Added default configuration when we add logging to our service provider. Doesn't get used when the ILoggerFactory comes in from the outside.

Also, refactored DbContextTest into three files since it was getting too big for R#. Since the diff may be hard to read, the three tests added were:
* Can_log_debug_events_with_OnConfiguring
* Cannot_log_debug_events_with_default_UseLoggerFactory
* Can_log_debug_events_with_UseLoggerFactory_when_configured


divega:
It seems AppVeyor fails on one of the new tests.

ajcvickers:
@divega Super weird. It passed on both my machines, but not on AppVeyer. I made the change that I thought we needed to change and now it is passing on my machines and on AppVeyer.

PR #9085, opened 2017-07-06
Add NETStandard.Library.NETFramework for older versions of Visual Studio

smitpatel:
@bricelam @ajcvickers

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -20,4 +20,8 @@
   <ItemGroup>
     <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(InternalAspNetCoreSdkVersion)" PrivateAssets="All" />
   </ItemGroup>
+  <!-- TODO: Remove when TestDriven supports VS15.3+ -->
+  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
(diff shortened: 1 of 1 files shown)


bricelam:
@ryanbrandenburg any harm in keeping this a while longer?

ryanbrandenburg:
The problem is `NETStandard.Library.NETFramework` will stop shipping at some point, so there will be a stale version on the feed until we clean it, at which point this will start to fail.

smitpatel:
I think it is ok. We can remove this once it start to fail.

smitpatel:
> there will be a stale version on the feed until we clean it, at which point this will start to fail.

It is available on NuGet feed so restore will never fail even if we clean our feeds. If the stale version causes errors then we can remove it.

ryanbrandenburg:
What about the time before it fails when that package is stale?

smitpatel:
> What about the time before it fails when that package is stale?

We can always remove it if it fails. Nothing stops us from doing so in future. No need to deteriorate dev experience for that.

ErikEJ:
What issue does this solve??

bricelam:
@ErikEJ It makes developing EF Core slightly easier with VS 15.2

PR #9091, opened 2017-07-06
Properly track dependent entities when queried.

### LINKED ISSUE #9086: Cascade delete throws for owned entities
From @arjenvrh https://github.com/aspnet/EntityFramework/issues/8923#issuecomment-313422001:
I tried out the Owner Types in Preview2 with the following classes:
```C#
public class ParentClass {
    [Required]
    [Key]
    [DatabaseGenerated(DatabaseGeneratedOption.None)]
    public Guid Id { get; set; }
    
    public virtual OwnedClass Ref { get; set; }
}

public class OwnedClass
{
    public string Field1 { get; set; }
    public string Field2 { get; set; }
}
```
and the following Mapping:
```
modelBuilder.Entity<ParentClass>()
    .OwnsOne<OwnedClass>(s => s.Ref, cb =>
    {
        cb.Property(c => c.Field1).HasColumnName("OwnedField1");
        cb.Property(c => c.Field2).HasColumnName("OwnedField2");
    });
```

When I try to delete a ParentClass item it throws an error of the ChangeTracker with a KeyNotFoundException.

AndriySvyryd:
Fix shadow owned types.

Fixes #9086

### DIFF
--- a/src/EFCore/ChangeTracking/Internal/StateManager.cs
+++ b/src/EFCore/ChangeTracking/Internal/StateManager.cs
@@ -25,7 +25,7 @@ namespace Microsoft.EntityFrameworkCore.ChangeTracking.Internal
         private readonly Dictionary<object, InternalEntityEntry> _entityReferenceMap
             = new Dictionary<object, InternalEntityEntry>(ReferenceEqualityComparer.Instance);
 
-        private readonly Dictionary<IEntityType, Dictionary<object, InternalEntityEntry>> _dietReferenceMap
+        private readonly Dictionary<IEntityType, Dictionary<object, InternalEntityEntry>> _dependentTypeReferenceMap
             = new Dictionary<IEntityType, Dictionary<object, InternalEntityEntry>>();
 
         private readonly LazyRef<IDictionary<object, IList<Tuple<INavigation, InternalEntityEntry>>>> _referencedUntrackedEntities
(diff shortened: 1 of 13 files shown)


src/EFCore/Metadata/Internal/InternalRelationshipBuilder.cs
```diff
@@ -960,7 +960,8 @@ private bool CanSetUnique(bool unique, ConfigurationSource? configurationSource,
                 return false;
             }
 
-            if ((Metadata.PrincipalToDependent != null)
+            if (Metadata.PrincipalToDependent != null
+                && !Metadata.PrincipalToDependent.IsShadowProperty
```
ajcvickers: Shadow nav property?

AndriySvyryd: yes, for snapshot

src/EFCore/Metadata/Internal/InternalEntityTypeBuilder.cs
```diff
@@ -1735,6 +1735,7 @@ private static IndexBuilderSnapshot DetachIndex(Index indexToDetach)
                 if (builder != null)
                 {
                     builder = builder.RelatedEntityTypes(Metadata, existingNavigation.GetTargetType(), configurationSource);
+                    builder = builder?.IsRequired(true, configurationSource);
```
ajcvickers: Does this means Owns always means required?

AndriySvyryd: Yes, it has to be, since it creates an identifying fk.

src/EFCore/Metadata/Internal/InternalRelationshipBuilder.cs
ajcvickers: Doh. 🤦‍♂️

src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -166,28 +166,35 @@ public virtual InternalEntityEntry GetOrCreateEntry(object entity, IEntityType e
...
+        }
 
-                    entries[entity] = entry;
+        private void AddToReferenceMap(InternalEntityEntry entry)
+        {
+            var mapKey = entry.Entity ?? entry;
+            var entityType = entry.EntityType;
+            if (entityType.HasDefiningNavigation())
+            {
+                foreach (var otherDiet in _model.GetEntityTypes(entityType.Name)
```
smitpatel: can you remove "diet"

Issue #8492, opened 2017-05-16
Query: incorrect sql generated for query with owned entities, left join and predicate using inner qsre


maumar:
query:

```cs
from l1 in 
    (from l1 in l1s
     join l2 in l2s on l1.Id equals l2.Level1_Optional_Id into grouping
     from l2 in grouping.DefaultIfEmpty()
     where l2.Name != "Foo"
     select l1).Distinct()
select l1.Id
```

query model after nav rewrite:

```
from Level1 l1 in 
    (from Level1 l1 in DbSet<Level1>
    join Level2 l2 in 
        from Level1 t in DbSet<Level1>
        join Level2 t.OneToOne_Required_PK in DbSet<Level2>
        on Property([t], "Id") equals Property([t.OneToOne_Required_PK], "Id") into t.OneToOne_Required_PK_group
        from Level2 t.OneToOne_Required_PK in 
            (from Level2 t.OneToOne_Required_PK_groupItem in [t.OneToOne_Required_PK_group]
            select [t.OneToOne_Required_PK_groupItem]).DefaultIfEmpty()
        where Property([t.OneToOne_Required_PK], "Id") != null
        select [t.OneToOne_Required_PK]
    on (Nullable<int>)[l1].Id equals [l2]?.Level1_Optional_Id into grouping
    from Level2 l2 in 
        (from Level2 <generated>_1 in [grouping]
        select [<generated>_1]).DefaultIfEmpty()
    where [l2]?.Name != "Foo"
    select [l1]).Distinct()
select (int)[l1]?.Id
```

generated sql:

```sql
SELECT [t1].[Id], [t1].[OneToOne_Required_PK_Date], [t1].[OneToOne_Required_PK_Level1_Optional_Id], [t1].[OneToOne_Required_PK_Level1_Required_Id], [t1].[OneToOne_Required_PK_Name], [t1].[OneToOne_Required_PK_OneToOne_Optional_PK_InverseId], [t1].[Id]
FROM (
	SELECT DISTINCT [t0].[Id], [t0].[OneToOne_Required_PK_Date], [t0].[OneToOne_Required_PK_Level1_Optional_Id], [t0].[OneToOne_Required_PK_Level1_Required_Id], [t0].[OneToOne_Required_PK_Name], [t0].[OneToOne_Required_PK_OneToOne_Optional_PK_InverseId]
	FROM [Level1] AS [l1]
	LEFT JOIN (
[...]

PR #9092, opened 2017-07-06
Fix to #8492 - Query: incorrect sql generated for query with owned entities, left join and predicate using inner qsre

maumar:
Problem was that for queries with owned types that were using GroupJoin-SelectMany-DefaultIfEmpty pattern we were not able to find the owned entity type behind the groupjoin subquery qsre.
Usually (i.e. in case of navigations) we look for the entity type in the lookup on QueryCompilationContext, but if the groupjoin is created manually then this lookup was not populated with entry for the groupjoin subquery.

Fix is to populate the lookup with entry for groupjoin subquery qsre. We do it in nav rewrite, which also populates other entries for owned types.

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
@@ -1368,6 +1368,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         {
             private readonly CollectionNavigationSubqueryInjector _subqueryInjector;
             private readonly bool _navigationExpansionSubquery;
+            private readonly QueryCompilationContext _queryCompilationContext;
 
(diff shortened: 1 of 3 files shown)


maumar:
new version up @AndriySvyryd

src/EFCore/Query/ExpressionVisitors/Internal/NavigationRewritingExpressionVisitor.cs
```diff
@@ -1379,10 +1380,19 @@ private class NavigationRewritingQueryModelVisitor : ExpressionTransformingQuery
             {
                 _subqueryInjector = new CollectionNavigationSubqueryInjector(queryModelVisitor, shouldInject: true);
                 _navigationExpansionSubquery = navigationExpansionSubquery;
+                _queryModelVisitor = queryModelVisitor;
```
AndriySvyryd: You only need `queryModelVisitor.QueryCompilationContext`

Issue #8925, opened 2017-06-21
Post-process model snapshot to massage metadata


ajcvickers:
The model snapshot may contain provider-specific annotations that are now only supported as relational annotation. For example, `SqlServer:TableName`. After building the snapshot we could process it to change these to `Relational:Xxx` based on looking for any annotation that ends in `:SomeRelationalAnnotation`. If this results in conflict (should be rare) we can pick one and warn saying that this is not supported anymore--note that it wouldn't have worked in migrations anyway.

This hook could also be generally useful in the future for other tweaks to metadata that may result in residual code in the snapshot that needs to be massaged.

ajcvickers:
Look at the version and only do it when needed.

AndriySvyryd:
@smitpatel Poaching this

PR #9095, opened 2017-07-06
Post-process the snapshot model to remove/combine legacy provider-specific annotations

ajcvickers:
Issue #8925

Currently relies on the model elements from the snapshot being mutable, but API could return a new IModel instead if this ever becomes not true.

### DIFF
--- a/src/EFCore.Design/Design/Internal/DesignTimeServicesBuilder.cs
+++ b/src/EFCore.Design/Design/Internal/DesignTimeServicesBuilder.cs
@@ -11,6 +11,7 @@ using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Migrations.Design;
+using Microsoft.EntityFrameworkCore.Migrations.Internal;
 using Microsoft.EntityFrameworkCore.Scaffolding.Internal;
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Utilities;
@@ -95,6 +96,7 @@ namespace Microsoft.EntityFrameworkCore.Design.Internal
             => services
                 .AddTransient<MigrationsScaffolderDependencies>()
                 .AddTransient<MigrationsScaffolder>()
+                .AddTransient<ISnapshotModelProcessor, SnapshotModelProcessor>()
                 .AddTransient(_ => contextServices.GetService<ICurrentDbContext>())
                 .AddTransient(_ => contextServices.GetService<IDatabaseProvider>())
                 .AddTransient(_ => contextServices.GetService<IDbContextOptions>())
--- a/src/EFCore.Design/Migrations/Design/MigrationsScaffolder.cs
+++ b/src/EFCore.Design/Migrations/Design/MigrationsScaffolder.cs
@@ -98,7 +98,7 @@ namespace Microsoft.EntityFrameworkCore.Migrations.Design
             }
 
             var modelSnapshot = Dependencies.MigrationsAssembly.ModelSnapshot;
-            var lastModel = modelSnapshot?.Model;
+            var lastModel = Dependencies.SnapshotModelProcessor.Process(modelSnapshot?.Model);
             var upOperations = Dependencies.MigrationsModelDiffer.GetDifferences(lastModel, Dependencies.Model);
             var downOperations = upOperations.Any()
                 ? Dependencies.MigrationsModelDiffer.GetDifferences(Dependencies.Model, lastModel)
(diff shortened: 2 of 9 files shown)


src/EFCore.Design/Migrations/Internal/ModelProcessor.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public virtual IModel Process(IModel model)
+        {
+            if (model != null)
+            {
+                ProcessElement(model);
+
+                foreach (var entityType in model.GetEntityTypes())
+                {
+                    ProcessElement(entityType);
```
smitpatel: We are reaching close to `ModelVisitor` just like `ExpressionVisitor` 🎉

ajcvickers: I thought about it, but there's really not that much to traverse.

src/EFCore.Design/Migrations/Internal/ModelProcessor.cs
```diff
@@ -0,0 +1,93 @@
...
+                            var relationalName = "Relational" + stripped;
+                            var duplicate = mutableMetadata.FindAnnotation(relationalName);
+
+                            if (duplicate == null)
+                            {
+                                mutableMetadata[relationalName] = annotation.Value;
+                            }
+                            else if (!Equals(duplicate.Value, annotation.Value))
+                            {
+                                _operationReporter.WriteWarning("Doh!");
```
smitpatel: Do we need better warning? :trollface:

src/EFCore.Design/Migrations/Internal/IModelProcessor.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace Microsoft.EntityFrameworkCore.Migrations.Internal
+{
+    public interface IModelProcessor
```
AndriySvyryd: Should this be ISnapshotModelProcessor?

src/EFCore.Design/Migrations/Internal/ModelProcessor.cs
ajcvickers: Oops!

src/EFCore.Design/Migrations/Internal/IModelProcessor.cs
ajcvickers: I'm fine with ISnapshotModelProcessor or ISnapshotProcessor , but I didn't worry too much since it is internal anyway.

src/EFCore.Design/Migrations/Internal/ModelProcessor.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public ModelProcessor([NotNull] IOperationReporter operationReporter)
+        {
+            _operationReporter = operationReporter;
+            _relationalNames = new HashSet<string>(
+                typeof(RelationalAnnotationNames)
+                    .GetTypeInfo()
+                    .GetRuntimeFields()
+                    .Where(p => p.Name != nameof(RelationalAnnotationNames.Prefix))
+                    .Select(p => ((string)p.GetValue(null)).Substring(RelationalAnnotationNames.Prefix.Length - 1)));
```
AndriySvyryd: I would say this is not efficient, but this is migrations...

ajcvickers: Indeed.

src/EFCore.Design/Migrations/Internal/ModelProcessor.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            foreach (var element in metadata)
+            {
+                ProcessElement(element);
+            }
+        }
+
+        private void ProcessElement(IAnnotatable metadata)
+        {
+            if (metadata is IMutableAnnotatable mutableMetadata)
```
AndriySvyryd: We should only do this for 1.x models

src/EFCore.Design/Migrations/Internal/ModelProcessor.cs
ajcvickers: Fixed.

src/EFCore.Design/Migrations/Internal/ModelProcessor.cs
ajcvickers: Fixed.

src/EFCore.Design/Migrations/Internal/IModelProcessor.cs
ajcvickers: Changed to ISnapshotModelProcessor

PR #9103, opened 2017-07-07
Fix migration snapshot for owned types

### LINKED ISSUE #9097: Fix migration snapshot for owned types

AndriySvyryd:
Fixes #9097


src/EFCore.Design/Migrations/Design/CSharpSnapshotGenerator.cs
```diff
@@ -105,34 +108,99 @@ private IReadOnlyList<IEntityType> Sort(IReadOnlyList<IEntityType> entityTypes)
...
+                {
+                    counter++;
+                }
+
+                builderName = "b" + (counter == 0 ? "" : counter.ToString());
+            }
+            else
+            {
+                builderName = "b";
+            }
```
bricelam: `ICSharpHelper.Identifier()` can take care of uniquifying if you keep track of the names used in the current scope.

AndriySvyryd: Then I would need to flow the scope. Current approach seems simpler.

PR #9104, opened 2017-07-07
Query: Improve parametrization of variables from nested closure

### LINKED ISSUE #8909: QueryCache misses for dynamically created queries with ConstantExpresssions or nested MemberExpressions from closure

smitpatel:
Resolves #8909

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
@@ -354,7 +354,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
 
             if (parameterValue is Expression parameterExpression)
             {
-                return parameterExpression;
+                return ExtractParameters(parameterExpression);
             }
 
             if (!_parameterize)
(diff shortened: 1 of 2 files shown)


src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -352,9 +352,9 @@ private Expression TryExtractParameter(Expression expression)
         {
             var parameterValue = Evaluate(expression, out var parameterName);
 
-            if (parameterValue is Expression parameterExpression)
```
ajcvickers: Quirk

PR #9108, opened 2017-07-07
Skip first time experience on Appveyor

ryanbrandenburg:


appveyor.yml
```diff
@@ -1,4 +1,4 @@
-version: 2.0.0-preview3-{build}
+﻿version: 2.0.0-preview3-{build}
```
smitpatel: Update this to rtm too.

smitpatel:
We already do that in EF.
 https://github.com/aspnet/EntityFramework/blob/dev/appveyor.yml#L21

PR #9134, opened 2017-07-10
[2.0.0] Design: Remove OperationExecutor.GetContextType

bricelam:
@vijayrkn Publish just calls `dotnet ef`, right? You don't use this?


bricelam:
(I double checked that Scaffolding doesn't use this.)

vijayrkn:
@bricelam - Yes publish calls `dotnet ef`

PR #9138, opened 2017-07-10
Upgrade Roslyn to 2.3.0-beta4-*

mikeharder:


smitpatel:
side question can a shipping package rtm version depend on pre-release version of dependency?

mikeharder:
@smitpatel: The Roslyn version will be upgraded again to `2.3.0` (final version) before .NET Core 2.0 RTM.  We're upgrading to `2.3.0-beta4-*` now, since the final `2.3.0` won't be available until tomorrow, but we want to find issues as early as possible.

smitpatel:
I assume that we will be pinning most of our dependencies (at least the ones being used in source, before we make final package for RTM release?

mikeharder:
@smitpatel: Correct.

CC: @pranavkm

Issue #1148, opened 2014-11-26
API Documentation


rowanmiller:
Oh yeah... this one will be fun

rowanmiller:
Made good progress in RTM code base. We may add some more docs before we release RTM but moving this issue out as we can release with what we have now.

ajcvickers:
Note: RelationalQueryContextFactory has internal docs when it is not internal.

PR #9149, opened 2017-07-11
Document Microsoft.EntityFrameworkCore.Design

bricelam:
Part of #1148


bricelam:
@ajcvickers I've created the `feature/docs` branch to merge these into.

smitpatel:
Side note: This will be building on all CI servers.

bricelam:
Yep. I wanted CI.

src/EFCore.Design/Design/ICSharpHelper.cs
```diff
@@ -7,37 +7,223 @@
 
 namespace Microsoft.EntityFrameworkCore.Design
 {
+    /// <summary>
+    /// Helper for generating C# code.
```
ajcvickers: Can we use indenting consistent with the other projects? (Which is what we have R# set to do cleanup as.)

src/EFCore.Design/Design/ICSharpHelper.cs
```diff
@@ -7,37 +7,223 @@
...
+
+        /// <summary>
+        ///     Generates a byte array literal.
+        /// </summary>
+        /// <param name="values"> The byte array. </param>
+        /// <returns> The literal. </returns>
         string Literal([NotNull] byte[] values);
+
+        /// <summary>
+        ///     Generates a bool literal.
```
ajcvickers: Is it work crefing these? (Probably not since they are primitives, but maybe it could be useful?)

src/EFCore.Design/Design/OperationExecutor.cs
```diff
@@ -10,20 +10,37 @@
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Design.Internal;
 using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Utilities;
 
 namespace Microsoft.EntityFrameworkCore.Design
 {
     /// <summary>
     ///     A version-resilient, AppDomain-and-reflection-friendly facade for command operations.
```
ajcvickers: Should "refelction" have an upper-case "R"?

src/EFCore.Design/Design/OperationExecutor.cs
```diff
@@ -10,20 +10,37 @@
...
     /// </summary>
+    /// <remarks>
+    ///     Use the <c>CreateInstance</c> overloads on <see cref="AppDomain"/> and <see cref="Activator"/> with the
```
ajcvickers: Based on feedback from Rowan when he was doing API docs, we no longer do "summary" and "remarks" but instead split things with multiple "para" paragraphs. Not sure if this has changed, or if we should sometimes do one and sometimes do the other, but would be good to know/decide and be consistent.

src/EFCore.Design/Design/ICSharpHelper.cs
bricelam: Will do. Got tired of fighting VS...

src/EFCore.Design/Design/ICSharpHelper.cs
bricelam: I thought about it, but wanted to reinforce that these were more about the C# language concept and not the CLR types...

src/EFCore.Design/Design/OperationExecutor.cs
bricelam: lol, the whole sentence could probably use some rewording. Will update

src/EFCore.Design/Design/ICSharpHelper.cs
smitpatel: In the end VS wins always :trollface:

src/EFCore.Design/Design/OperationExecutor.cs
bricelam: The motivation was because remarks don't show up in intellisense--only on MSDN. I'm fine either way. My thought was that almost nobody would be using these so only having it on MSDN (or docs.microsoft.com I guess) would be sufficient.

ajcvickers: I was thinking about this last night (because I'm such an exciting person that I think about such things on my evening!) and think that the idea should be that if there are multiple paragraphs of information that are very relevant and so should show up in Intellisense, then we should use paras in the summary. Only in places where there is some incidental comment that's not super important should it go in remarks. So remarks should be quite rare, because most of the time if it's worth writing it, then it's worth showing up in the Intellisense.

src/EFCore.Design/Design/ICSharpHelper.cs
ajcvickers: I can always run R# cleanup on it after it goes in... :trollface: 😉

src/EFCore.Design/Design/ICSharpHelper.cs
ajcvickers: Agreed.

PR #9163, opened 2017-07-12
ComplexNavigations and QueryTestBase refactoring 

maumar:
Move AssertQuery from ComplexNavigations and QueryTestBase to a common helper class - QueryAsserter.

This is part of the bigger query test refactoring. In future checkins more AssertQuery methods will be migrated to the common pattern.

Currently we need to maintain two NorthwindData classes, static one for legacy tests and instance one for the new tests. Once all the tests using the static NorthwindData are migrated, the class will be removed.


src/EFCore.Specification.Tests/Query/ISetExtractor.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+
+namespace Microsoft.EntityFrameworkCore.Query
+{
+    public abstract class ISetExtractor<TContext> where TContext : DbContext
```
maumar: this provides functionality that we use in owned complex navs, where we do additional processing on context.Set<T>() before it's used in a query.

src/EFCore.Specification.Tests/TestModels/Northwind/NorthwindData.cs
```diff
@@ -10,12 +10,170 @@
...
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Query;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors;
 using Microsoft.EntityFrameworkCore.Query.Internal;
 using Microsoft.EntityFrameworkCore.Extensions.Internal;
 
 namespace Microsoft.EntityFrameworkCore.TestModels.Northwind
 {
+    public class NorthwindData2 : IExpectedData
```
maumar: this is temporary, once all the tests using static NorthwindData are moved to the new pattern, the static class will be removed

src/EFCore.Specification.Tests/Query/QueryTestBase.Select.cs
```diff
@@ -17,7 +17,7 @@ public abstract partial class QueryTestBase<TFixture>
         [ConditionalFact]
         public virtual void Select_into()
         {
-            AssertQuery<Customer>(
+            Fixture.QueryAsserter.AssertQuery<Customer>(
```
AndriySvyryd: You could extract this back into `AssertQuery` for more DRYness

maumar:
new version up @AndriySvyryd @smitpatel

src/EFCore.Specification.Tests/TestUtilities/IQueryAsserter.cs
```diff
@@ -0,0 +1,149 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.EntityFrameworkCore.TestUtilities
+{
+    public interface IQueryAsserter
```
maumar: switch to abstract class, since we are likely to add more methods?

src/EFCore.Specification.Tests/Query/ComplexNavigationsQueryTestBase.cs
```diff
@@ -1403,7 +1396,7 @@ public virtual void Include_nested_with_optional_navigation()
                 new ExpectedInclude<Level3>(l1 => l1.OneToOne_Required_FK, "OneToOne_Required_FK", "OneToOne_Optional_FK.OneToMany_Required")
             };
 
-            AssertIncludeQuery<Level1>(
+            Fixture.QueryAsserter.AssertIncludeQuery<Level1>(
```
AndriySvyryd: Missed these 2

src/EFCore.Specification.Tests/Query/QueryFixtureBase.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.EntityFrameworkCore.TestUtilities;
+
+namespace Microsoft.EntityFrameworkCore.Query
+{
+    public class QueryFixtureBase
```
AndriySvyryd: >QueryFixtureBase [](start = 17, length = 16)

This should be an interface, otherwise it won't be compatible with the new test pattern I'm working on

src/EFCore.Specification.Tests/Query/QueryTestHelpers.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Microsoft.EntityFrameworkCore
+{
+    public static class QueryTestHelpers
+    {
+        public static Func<dynamic, dynamic> GroupingSorter<TKey, TElement>()
```
AndriySvyryd: Can these be moved to QueryTestBase?

src/EFCore.Specification.Tests/TestUtilities/IQueryAsserter.cs
AndriySvyryd: Sure, give it a go. Perhaps the only abstract method we actually need is CreateContext()

---
In reply to: [127563807](https://github.com/aspnet/EntityFramework/pull/9163#discussion_r127563807) [](ancestors = 127563807)

AndriySvyryd:
:shipit:

src/EFCore.Specification.Tests/Query/QueryFixtureBase.cs
maumar: IClassFixture<TFixture> requires TFixture to be a class

AndriySvyryd: That's fine:
```C#
where TFixture : class, IQueryFixtureBase, new()
```

PR #9177, opened 2017-07-13
[WIP] Benchmarks: Use BenchmarkDotNet

smitpatel:
**This is prototype for initial feedback**

This converts `SimpleQueryTests` in EF6 to use new framework.
Benchmarks are run as console app. So nomore xunit tests.
For older version we don't need to run in each commit since they shouldn't change. They will be moved out to different solution. It will be still possible to run them manually as console app.
The current version associated benchmark will be made sure to run with build.cmd

Benchmarks are run in following way.
- Invoke `GlobalSetup`
- Invoke Benchmark method multiple times and compute results.
- Invoke `GlobalCleanup`

Framework collects data for the Benchmark method so it is isolated as much as possible by moving work of setting up etc in Global* methods.

We make sanity check on results in `GlobalCleanup` it is just to make sure we ran against correct database so our results are not totally false.

Needs feedback on
- Is the coding style looks good and flexible enough for us to convert to it?
- Anymore data to store from run?
- Anymore patterns to minimize the benchmark method to actual thing being benchmarked?

This will not be merged till all benchmark changes are made to avoid breaking current benchmark runs.

### DIFF
--- a/test/EFCore.Benchmarks.EF6/Query/SimpleQueryTests.cs
+++ b/test/EFCore.Benchmarks.EF6/Query/SimpleQueryTests.cs
@@ -4,329 +4,218 @@
 using System.Data.Entity;
 using System.Linq;
 using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
 using Microsoft.EntityFrameworkCore.Benchmarks.EF6.Models.Orders;
+using Microsoft.EntityFrameworkCore.Benchmarks.Models.Orders;
 using Xunit;
+
+// ReSharper disable ReturnValueOfPureMethodIsNotUsed
 
 namespace Microsoft.EntityFrameworkCore.Benchmarks.EF6.Query
 {
     [SqlServerRequired]
-    public class SimpleQueryTests : IClassFixture<SimpleQueryTests.SimpleQueryFixture>
+    public class SimpleQueryTests
     {
-        private readonly SimpleQueryFixture _fixture;
+        private static readonly SimpleQueryFixture _fixture = new SimpleQueryFixture();
 
-        public SimpleQueryTests(SimpleQueryFixture fixture)
+        private OrdersContext _context;
+
+        [Params(true, false)]
+        public bool Async;
+
+        [Params(true, false)]
+        public bool Tracking;
+
+        [GlobalSetup]
+        public void GlobalSetup()
         {
-            _fixture = fixture;
+            _context = _fixture.CreateContext();
+
+            Assert.Equal(1000, _context.Products.Count());
+            Assert.Equal(1000, _context.Customers.Count());
+            Assert.Equal(2000, _context.Orders.Count());
         }
 
-        [Benchmark]
-        [BenchmarkVariation("Tracking On - Sync (1 query)", true, false, 1)]
-        [BenchmarkVariation("Tracking Off - Sync (10 queries)", false, false, 10)]
-        [BenchmarkVariation("Tracking On - Async (1 query)", true, true, 1)]
-        [BenchmarkVariation("Tracking Off - Async (10 queries)", false, true, 10)]
-        public async Task LoadAll(IMetricCollector collector, bool tracking, bool async, int queriesPerIteration)
+        [GlobalCleanup]
+        public void GlobalCleanup()
         {
-            using (var context = _fixture.CreateContext())
+            _context.Dispose();
+        }
+
+        [BenchmarkDotNet.Attributes.Benchmark]
+        public async Task LoadAll()
+        {
+            var query = _context.Products
+                .ApplyTracking(Tracking);
+
+            if (Async)
             {
-                var query = context.Products
-                    .ApplyTracking(tracking);
-
-                using (collector.StartCollection())
-                {
-                    for (var i = 0; i < queriesPerIteration; i++)
-                    {
-                        if (async)
-                        {
-                            await query.ToListAsync();
-                        }
-                        else
-                        {
-                            query.ToList();
-                        }
-                    }
-                }
-
-                Assert.Equal(1000, query.Count());
-                Assert.False(tracking && (queriesPerIteration != 1), "Multiple queries per iteration not valid for tracking queries");
+                await query.ToListAsync();
+            }
+            else
+            {
+                query.ToList();
             }
         }
 
-        [Benchmark]
-        [BenchmarkVariation("Tracking On - Sync (1 query)", true, false, 1)]
-        [BenchmarkVariation("Tracking Off - Sync (10 queries)", false, false, 10)]
-        [BenchmarkVariation("Tracking On - Async (1 query)", true, true, 1)]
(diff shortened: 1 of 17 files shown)


test/EFCore.Benchmarks.EF6/Query/SimpleQueryTests.cs
```diff
@@ -1,332 +1,259 @@
...
-                        {
-                            query.ToList();
-                        }
-                    }
+                    Assert.Equal(_expectedCount, count);
+                }
+                else if (_result is List<Group> groupList)
+                {
+                    Assert.Equal(_expectedCount, groupList.Count);
+                    Assert.All(groupList, g => Assert.Equal(100, g.Products.Count()));
```
divega: Isn't the Count method checking this already?

divega: I mean the Count benchmark.

smitpatel: It first checks no. of groups And then no. of products in each group.

test/EFCore.Benchmarks.EF6/Query/SimpleQueryTests.cs
```diff
@@ -1,332 +1,259 @@
...
+                }
+                else if (_result is List<Customer> customerList)
+                {
+                    Assert.Equal(_expectedCount, customerList.Count);
+                    Assert.Equal(2 * _expectedCount, customerList.SelectMany(c => c.Orders).Count());
+                }
+                else if (_result is IList list)
+                {
+                    Assert.Equal(_expectedCount, list.Count);
```
divega: Isn't this and the 'is int' case all the only sanity checks that we need?

test/EFCore.Benchmarks.EF6/Query/SimpleQueryTests.cs
```diff
@@ -1,332 +1,259 @@
...
+            _context.Dispose();
         }
 
-        [Benchmark]
-        [BenchmarkVariation("Tracking On - Sync (1 query)", true, false, 1)]
-        [BenchmarkVariation("Tracking Off - Sync (10 queries)", false, false, 10)]
-        [BenchmarkVariation("Tracking On - Async (1 query)", true, true, 1)]
-        [BenchmarkVariation("Tracking Off - Async (10 queries)", false, true, 10)]
-        public async Task Where(IMetricCollector collector, bool tracking, bool async, int queriesPerIteration)
+        [BenchmarkDotNet.Attributes.Benchmark]
```
divega: @smitpatel and I talked about why the namespace is used. It is because we still have our own attribute also called Benchmark. It will go away.

test/EFCore.Benchmarks.EF6/Query/SimpleQueryTests.cs
smitpatel: Will remove all result checks. We just need one sanity check to verify data we don't need to verify results.
I will try doing it in ctor or GlobalSetup whichever is invoked lesser times.

test/EFCore.Benchmarks.EF6/Query/SimpleQueryTests.cs
```diff
@@ -1,332 +1,259 @@
...
         }
 
-        [Benchmark]
-        [BenchmarkVariation("Tracking On - Sync (1 query)", true, false, 1)]
-        [BenchmarkVariation("Tracking Off - Sync (10 queries)", false, false, 10)]
-        [BenchmarkVariation("Tracking On - Async (1 query)", true, true, 1)]
-        [BenchmarkVariation("Tracking Off - Async (10 queries)", false, true, 10)]
-        public async Task OrderBy(IMetricCollector collector, bool tracking, bool async, int queriesPerIteration)
+        [BenchmarkDotNet.Attributes.Benchmark]
+        public async Task Where()
```
divega: In the switch to BenchmarkDotNet we are loosing the ability to decide when we actually want to "start collection": the perf of the whole method will be reported. Hence we want to make sure that the overhead of the machinery we are using to apply tracking vs. no tracking and and async vs. non-async isn't skewing the results. @smitpatel will try comparing runs of the parameterized method vs. the a non-parameterized method.

roji:
Just a general +1 to switching to BenchmarkDotNet, which I've been using for a while with Npgsql and with which I'm very very happy.

smitpatel:
@roji - Thanks for info. It would be also useful if you can give feedback on the test pattern used in this PR. 😄

smitpatel:
@AndriySvyryd - Please have a look at this when you get time. I would like to get some feedback before I start converting all tests.

AndriySvyryd:
Looking good so far

test/EFCore.Benchmarks.EF6/Query/SimpleQueryTests.cs
roji: That's true. For cases where the added conditional overhead of benchmark parameters (e.g. sync vs. async) is important, it's always possible to break down the single benchmark into multiple (i.e. WhereSync vs. WhereAsync). However, I'm guessing the overhead here is supposed to be marginal (after all database I/O is being performed), otherwise that itself may be a performance issue :)

roji:
@smitpatel nothing in particular to say, it definitely looks OK.

One point - it would be great if your benchmarks weren't tied to SqlServer, and could be executed on, say, Npgsql. I'm thinking of something similar to the functional specification tests, which allow providers to inherit and provide their own fixture/context. This would also allow us to compare the performance of the various providers and databases, and shouldn't be too difficult.

smitpatel:
@roji - Thanks for feedback. That is good idea to benchmark against multiple providers. To avoid CR complexity, I will convert all current tests to new framework first and then branch out for different providers as making it multiple provider will require just abstracting fixture out.

I am going forward with this conversion and will pull out this PR from **WIP** once ready for review.

bricelam:
LGTM

test/EFCore.Benchmarks/v2/SqlServerBenchmarkResultProcessorv2.cs
```diff
@@ -0,0 +1,125 @@
...
+
+using System;
+using System.Data;
+using System.Data.SqlClient;
+
+namespace Microsoft.EntityFrameworkCore.Benchmarks.v2
+{
+    public class SqlServerBenchmarkResultProcessorv2
+    {
+        private static readonly string _insertCommand =
```
anpete: No EF?! :trollface:

smitpatel: Cannot use EF since this is class library shared between EF6/EFCore1.1/EFCore2.0/EFCore latest dev. 😉

anpete: Web service! :trollface:

smitpatel:
superseded by #9284

Issue #9192, opened 2017-07-17
Don't set global LogLevel


BrennanConroy:
`AddCoreServices` is setting the global minimum `LogLevel` for everyone see https://github.com/aspnet/EntityFramework/blob/2a70de4776bef1c047e36700d4a1ffa4a5bbf6c8/src/EFCore/EntityFrameworkServiceCollectionExtensions.cs#L289

This definitely looks wrong especially since the user will configure logging in `IWebHostBuilder.ConfigureLogging` and `AddCoreServices` will overwrite this.

/cc @pakrym @Eilon **This needs to be fixed for 2.0.0**

ajcvickers:
@BrennanConroy What should EF be doing to ensure that:
* If Logging has already been configured in D.I. then EF doesn't stomp on this
* If Logging has not been configured then EF can add logging with a different global default?

pakrym:
Why do you need to change global logging level?

ajcvickers:
@pakrym Because otherwise when people are using EF in non ASP.NET applications and hook into the logging they are missing a lot of messages that they expect to see. Beyond that, we don't even know of a way that they can ever start seeing the messages because there doesn't seem to be any way for them to dynamically change this later.

We're at least somewhat okay with this being an ASP.NET problem when it is an ASP.NET application and we will tell people to go configure logging in the ASP.NET way. When it's not an ASP.NET application that really isn't an option.

The fundamental problem is that prior to 2.0, the ILogger mechanism was something that we could hook into and expect to get everything we sent to it come out the other end. EF logging is built on that premise, but it no longer holds true--now stuff is filtered and we seem to have no way to stop this happening. We have a work item re-write EF logging to not use ILogger since it is no longer really suitable for our use, but since the logging changes happened late in the game for 2.0 we were not able to do this yet--we'll try to do it in 2.1. At that point we can tell people to just not use ILogger anymore. Until then we're trying to preserve as much as possible the logging functionality that we had in 1.1.

/cc @divega

pakrym:
@ajcvickers when `AddCoreServices` get's called, do you know if you are in DI or non-DI environment?

ajcvickers:
Hmmm. We made another small change to the way we hook up services which could mean this isn't needed anymore, at least for non-ASP.NET applications. I'll run some tests.

@divega Are you okay with debug messages not showing if you hook into EF logging in an ASP.NET application?

PR #9197, opened 2017-07-18
Revert setting default log level

ajcvickers:
Because we now redirect ILoggerFactory in a slightly different way it means that this is not needed when not using an ASP.NET application. When using an ASP.NET application you will get whatever filtering, etc, was setup for ASP.NET. This is different from 1.1 and may be confusing, but doesn't seem unreasonable.

Issue #9192


ajcvickers:
@Eilon @pakrym @divega @BrennanConroy

PR #9204, opened 2017-07-19
Remove reference to NETStandard.Library.NETFramework

mikeharder:
I don't think there's any customer impact to *not* take this in 2.0, since the reference is now added implicitly by the SDK when necessary.  Further, the reference was added with `PrivateAssets="All"` which means it shouldn't create a package reference in the nuspec.

However, if we want all our repos to be consistent for the 2.0 release, it should be removed.

@smitpatel or @bricelam: Can you verify whether this is still needed for TestDriven.NET?  I don't think it should be, since the SDK is now implicitly doing what we were doing explicitly before.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -20,8 +20,4 @@
   <ItemGroup>
     <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(InternalAspNetCoreSdkVersion)" PrivateAssets="All" />
   </ItemGroup>
-  <!-- TODO: Remove when TestDriven supports VS15.3+ -->
-  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
-    <PackageReference Include="NETStandard.Library.NETFramework" Version="$(NETFrameworkPackageVersion)" PrivateAssets="All" />
-  </ItemGroup>
 </Project>
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -8,7 +8,6 @@
     <InternalAspNetCoreSdkVersion>2.0.1-*</InternalAspNetCoreSdkVersion>
     <JsonNetVersion>10.0.1</JsonNetVersion>
     <MoqVersion>4.7.49</MoqVersion>
-    <NETFrameworkPackageVersion>2.0.0-*</NETFrameworkPackageVersion>
     <NETStandardImplicitPackageVersion>2.0.0-*</NETStandardImplicitPackageVersion>
     <RelinqVersion>2.1.1</RelinqVersion>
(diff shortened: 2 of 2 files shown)


smitpatel:
I tested it out again right now because it does not work still.

Afaik there is no _customer impact_ in taking this or **not** taking this in. This is needed for dev experience in EF team. Sad to say we cannot move away from TestDriven.net when built-in test explorer takes more than 6 mins just to discover tests.

If there is going to be no effect in package produced then I don't see why we should do this?

cc: @ajcvickers

mikeharder:
I'm fine with keeping this if it has no customer impact and helps the EF team dev experience.  I'm suprised it still makes a difference since the SDK is supposed to implicitly add this reference when necessary, but maybe TestDriven.Net is a special case.

smitpatel:
Its not TD.net actually. Using VS 15.2 with setting env variable to latest SDK, build fails without this. Though for everything else VS picks up things from latest SDK.

mikeharder:
Latest SDK only supports VS 15.3 though.

smitpatel:
It does work with lower VS for most part (or rather appropriate parts we need in EF).

mikeharder:
I'm fine keeping this if there's no customer impact (which I believe is the case).  The only downside is it's inconsistent with most other repos.

smitpatel:
> there's no customer impact
I don't know. When I added this back, @bricelam said there was none. But relevant people should be asked about it.

ajcvickers:
We should not do this right now. Morale on team with current dev experience is better than it has been in months. I'd rather have dev team keep working on getting things done as efficiently as possible for now. Dogfooding question is something we need to continually follow up on, but for us having a VS that works with R# and TD.NET is very important for efficiency and morale. It's also important for a lot of the tasks that we are doing in MQ since they involve cleaning up code (for which R# is invaluable) and refactoring test infra (for which R# and TD.NET is invaluable.)

mikeharder:
Tracking Issue: https://github.com/aspnet/EntityFramework/issues/9207

ErikEJ:
@ajcvickers What is MQ? 😀  and have you addressed the Test Explorer issue with the VS team?

ajcvickers:
@ErikEJ MQ = Milestone Quality, which for us is just Microsoft-speak for spending some time on infrastructure, technical debt, etc.

We have provided feedback at various times, but I think they have conflicting requirements from other users with regard to how the U.I. works, etc. Interestingly, pure speed doesn't seem to be top of the list for what people want in a test runner.

Issue #9202, opened 2017-07-18
Query: Exception when using Include on collection navigation for models also containing owned types


javiercn:
Describe what is not working as expected.
An exception is thrown when trying to include the related elements on a many to many relationship.

Exception:
```
ArgumentNullException: Value cannot be null.
 Parameter name: entityType
Microsoft.EntityFrameworkCore.Utilities.Check.NotNull<T>(T value, string parameterName)
Microsoft.EntityFrameworkCore.RelationalMetadataExtensions.Relational(IEntityType entityType)
Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.RelationalEntityQueryableExpressionVisitor.VisitEntityQueryable(Type elementType)
Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.EntityQueryableExpressionVisitor.VisitConstant(ConstantExpression constantExpression)
System.Linq.Expressions.ConstantExpression.Accept(ExpressionVisitor visitor)
System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.ReplaceClauseReferences(Expression expression, IQuerySource querySource, bool inProjection)
Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.CompileGroupJoinInnerSequenceExpression(GroupJoinClause groupJoinClause, QueryModel queryModel)
Microsoft.EntityFrameworkCore.Query.RelationalQueryModelVisitor.CompileGroupJoinInnerSequenceExpression(GroupJoinClause groupJoinClause, QueryModel queryModel)
Microsoft.EntityFrameworkCore.Query.EntityQueryModelVisitor.VisitGroupJoinClause(GroupJoinClause groupJoinClause, QueryModel queryModel, int index)
Microsoft.EntityFrameworkCore.Query.RelationalQueryModelVisitor.VisitGroupJoinClause(GroupJoinClause groupJoinClause, QueryModel queryModel, int index)
Remotion.Linq.Clauses.GroupJoinClause.Accept(IQueryModelVisitor visitor, QueryModel queryModel, int index)
Remotion.Linq.QueryModelVisitorBase.VisitBodyClauses(ObservableCollection<IBodyClause> bodyClauses, QueryModel queryModel)
Remotion.Linq.QueryModelVisitorBase.VisitQueryModel(QueryModel queryModel)
[...]

PR #9217, opened 2017-07-19
[2.0.0] Fix to #9202 - Query: Exception when using Include on collection navigation for models also containing owned types

maumar:
Problem was that when processing include collection we clone the query model, which changes references of the individual query sources inside. In case of owned types we rely on the references to determine which entity type corresponds with a given query source (using a dictionary).
However, after cloning we were not populating the dictionary with the additional entries, so the owned types corresponding to those clones query sources could not be found.

Fix is to populate the QuerySourceEntityTypeMapping with entries for the cloned query sources - like we do in other places.


maumar:
Justification:
Impact: High - it affects all queries that perform include on collection navigation, when the query model also contains a query source pointing to an owned type, especially now that owned types are included automatically.

Risk: Very low - fix is a one-liner, adding missing entries to the dictionary that we used to determine which type corresponds with which query source. We do this already in other places (e.g.SubQueryMemberPushDownExpressionVisitor)

ajcvickers:
Looks good to me but I would like at least one of @smitpatel or @anpete to review as well.

Sent bang email to @Eilon and @DamianEdwards to get 2.0 approval.

ajcvickers:
@maumar This is approved for 2.0. Please make sure the fix gets into both branches.

Issue #8842, opened 2017-06-13
Benchmarks: Rework Benchmark Infra


smitpatel:
`BenchmarkAttribute` is a `FactAttribute` but it is used on tests with parameters & alongwith `DataAttribute` like `BenchmarkVariation` which are generally used on `Theory`. As always xUnit does not like our mess and shows errors. 😝 

This errors/warnings are explicitly disabled in csproj files at present. Rework this infrastructure and remove nowarns.

anpete:
Even better, lets start using [BenchmarkDotNet](http://benchmarkdotnet.org/)

ajcvickers:
Includes getting comparison of EF6/1,.1/2.0.

smitpatel:
Decided in ad-hoc discussion today that benchmarks on current dev will not be run as xUnit test as a part of build.
- They will be included in same solution so that any API breaks/usages can be found.
- It will not run as xUnit test so if there is functional break then it will be not caught during build.cmd run though given coverage in other tests there should be minimal impact. (Also there is a way to run it manually)
- If any functional breaks happen then we can react to it.

smitpatel:
Blocked on https://github.com/dotnet/BenchmarkDotNet/issues/481

smitpatel:
We have moved to use BenchmarkDotNet for our benchmark runs.

PR #9234, opened 2017-07-20
Benchmarks: Convert Benchmarks.EF6 to use BenchmarksDotNet

smitpatel:
Move benchmarks to separate directory
Create Separate solution for benchmarks of older releases
Don't build Benchmarks.sln while build.cmd script

First part of #8842


smitpatel:
Looks like build fails.

smitpatel:
Build passes.

benchmarks/EFCore.Benchmarks.EF6/ChangeTracker/DbSetOperationTests.cs
```diff
@@ -0,0 +1,121 @@
...
+                Context.Dispose();
+            }
+        }
+
+        public class AddDataVariations : Base
+        {
+            [IterationSetup]
+            public void InitializeContext()
+            {
+                Context = Fixture.CreateContext();
```
anpete: Do this in Base?

smitpatel: Marking all methods virtual. & doing stuff in base wherever common.

benchmarks/EFCore.Benchmarks.EF6/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+            benchmarkSummaryProcessor.Process(BenchmarkRunner.Run<NavigationsQueryTests>());
+            benchmarkSummaryProcessor.Process(BenchmarkRunner.Run<QueryCompilationTests>());
+            benchmarkSummaryProcessor.Process(BenchmarkRunner.Run<RawSqlQueryTests>());
+            benchmarkSummaryProcessor.Process(BenchmarkRunner.Run<SimpleQueryTests>());
+
+            // Update
+            benchmarkSummaryProcessor.Process(BenchmarkRunner.Run<SimpleUpdatePipelineTests.Insert>());
+            benchmarkSummaryProcessor.Process(BenchmarkRunner.Run<SimpleUpdatePipelineTests.Update>());
+            benchmarkSummaryProcessor.Process(BenchmarkRunner.Run<SimpleUpdatePipelineTests.Delete>());
+            benchmarkSummaryProcessor.Process(BenchmarkRunner.Run<SimpleUpdatePipelineTests.Mixed>());
```
bricelam: Would be nice to not have to update this every time we add a new test.

smitpatel: Discussed with @bricelam . This allows us to run particular test locally if we want to run. Whenever adding new benchmarks you need to add line here to test anyway so keeping this as is also have value.

smitpatel:
Reworking a bit (converting EFCore dev tests first to compare with current numbers first)

PR #9236, opened 2017-07-21
Benchmarks: Create new project to benchmark EF Core 2.0.0

smitpatel:
Part of #8842

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -18,10 +18,13 @@
     <TestSdkVersion>15.3.0-*</TestSdkVersion>
     <XunitVersion>2.3.0-*</XunitVersion>
   </PropertyGroup>
+<!-- Following package versions must match with what is available on NuGet -->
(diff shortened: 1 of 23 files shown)


build/dependencies.props
```diff
@@ -23,5 +23,7 @@
     <BenchmarkDotNetVersion>0.10.8</BenchmarkDotNetVersion>
     <EFCore1Version>1.1.2</EFCore1Version>
     <EFCore1SqlClientVersion>4.3.1</EFCore1SqlClientVersion>
+    <EFCore2Version>2.0.0-rtm-*</EFCore2Version>
```
bricelam: This won't be available. Should it be `2.0.0-preview2-final` until `2.0.0` ships?

smitpatel: Updated to preview2-final. Will change to 2.0.0 once we ship.

PR #9239, opened 2017-07-21
[2.0.0] CLI Tools: Add --no-build

bricelam:
Verification: Manually tested


bricelam:
@smitpatel Not until we have a legitimate reason.

PR #9240, opened 2017-07-21
2.0.0-rtm to 2.1.0-preview1

ryanbrandenburg:


bricelam:
Search the codebase for `2.0.0`. There are a few other places that need to be updated.

bricelam:
Also, we'll need to update the aspnet dependencies. Will that happen in a second phase?

smitpatel:
Also change in `appveyor.yml` for book keeping.

ryanbrandenburg:
The AspNet dependencies will be updated in a second pass once other repos have a change to have their packages be available.

test/EFCore.Design.Tests/Migrations/Design/SnapshotModelProcessorTest.cs
```diff
@@ -126,7 +126,7 @@ public void Does_not_warn_for_duplicate_non_conflicting_annotations()
         public void Does_not_process_non_v1_models()
         {
             var model = new Model();
-            model.SetProductVersion("2.0.0");
+            model.SetProductVersion("2.1.0");
```
bricelam: Can keep at 2.0.0

PR #9243, opened 2017-07-21
Migrations Tests: Cleanup

bricelam:
Mostly removes model builder workarounds that have since been fixed.

Also removes some tests no longer necessary since we have per-provider models.


test/EFCore.CrossStore.FunctionalTests/TestModels/CrossStoreContext.cs
```diff
@@ -20,7 +20,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                 .Entity<SimpleEntity>(eb =>
                     {
                         eb.Property(typeof(string), SimpleEntity.ShadowPartitionIdName);
-                        eb.ToTable("RelationalSimpleEntity"); // TODO: specify schema when #948 is fixed
+                        eb.ToTable("RelationalSimpleEntity");
```
smitpatel: Should a schema be added here?

bricelam: I think the intent has been superseded by the spec tests. There's little point to doing it now.

PR #9250, opened 2017-07-24
Change xunit version in specs to 2.2.0

pranavkm:


smitpatel:
@mikeharder - They reference xUnit transitively through Specs.Tests

pranavkm:
Yup, no references in those projects to update: https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore.Relational.Design.Specification.Tests/EFCore.Relational.Design.Specification.Tests.csproj

PR #9258, opened 2017-07-25
Converting GearsOfWar tests to use AssertQuery verification infrastructure

maumar:
Making gears of war verification more comprehensive, also removing a lot of code (about 1000 lines)


src/EFCore.Specification.Tests/Query/ComplexNavigationsQueryTestBase.cs
```diff
@@ -3601,24 +3601,24 @@ public void Entries_for_detached_entities_are_removed()
...
+        //{
+        //    if (caller == null)
+        //    {
+        //        return null;
+        //    }
 
-            return expression();
-        }
+        //    return expression();
+        //}
```
AndriySvyryd: Remove

PR #9259, opened 2017-07-25
Disable IncludeSource in projects with IncludeSymbols disabled

pranavkm:


smitpatel:
@pranavkm @natemcmaster - Can any of you explain this change?

cc: @divega @ajcvickers

pranavkm:
We enabled `IncludeSource` for symbol packages via the SDK. This broke packing a couple of projects which previously had symbol package generation disabled. I re-disabled it by turning off `IncludeSource`.

ajcvickers:
@smitpatel Is there anything we need to follow up on here?

smitpatel:
@ajcvickers - No

PR #9264, opened 2017-07-26
Benchmarks: Run latest dev benchmarks using BenchmarkDotnet

smitpatel:
Part of #8842 

This creates benchmarks for current dev in new framework. There are few things changed from earlier tests. Most of these test results align with current run results (or improve them).
Pending on UpdatePipeline Batching discussion.

Based on this, snapshots for older version will be created (exact replica just different version targeted)


smitpatel:
Update pipeline tests are working right way on a remote server. This is ready for review.
(this excludes EF6 benchmarks but I will add them before merging to dev)

Issue #9074, opened 2017-07-05
Query isn't detecting concurrency reliably


AndriySvyryd:
See `Throws_on_concurrent_query_first` test in https://github.com/aspnet/EntityFramework/pull/8865

PR #9283, opened 2017-07-27
Converting async query tests to use AssertQuery verification infrastructure

maumar:
- creating AsyncQueryTestBase with all the async helper methods,
- adapting ResultAsserter to support async calls,
- removing static NorthwindData, since it's no longer used

### DIFF
--- a/src/EFCore.Specification.Tests/Query/AsyncQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/AsyncQueryTestBase.cs
@@ -4,4056 +4,476 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.EntityFrameworkCore.Infrastructure;
-using Microsoft.EntityFrameworkCore.Internal;
-using Microsoft.EntityFrameworkCore.TestModels.Northwind;
-using Microsoft.EntityFrameworkCore.TestUtilities;
-using Microsoft.EntityFrameworkCore.TestUtilities.Xunit;
-using Microsoft.Extensions.DependencyInjection;
 using Xunit;
 
-// ReSharper disable InconsistentNaming
-// ReSharper disable ConvertToExpressionBodyWhenPossible
-// ReSharper disable AccessToDisposedClosure
-// ReSharper disable AccessToModifiedClosure
-// ReSharper disable StringStartsWithIsCultureSpecific
-// ReSharper disable StringEndsWithIsCultureSpecific
 namespace Microsoft.EntityFrameworkCore.Query
 {
     public abstract class AsyncQueryTestBase<TFixture> : IClassFixture<TFixture>
-        where TFixture : NorthwindQueryFixtureBase<NoopModelCustomizer>, new()
+        where TFixture : class, IQueryFixtureBase, new()
     {
         protected AsyncQueryTestBase(TFixture fixture) => Fixture = fixture;
 
         protected TFixture Fixture { get; }
-        
-        [ConditionalFact]
-        public virtual async Task ToList_on_nav_in_projection_is_async()
-        {
-            using (var context = CreateContext())
-            {
-                var results
-                    = await context.Customers
-                        .Where(c => c.CustomerID == "ALFKI")
-                        .Select(c => new
-                        {
-                            c,
-                            Orders = ((IAsyncEnumerable<Order>)(from o in c.Orders select o)).ToList().Result
-                        })
-                        .ToListAsync();
 
-                Assert.Equal(1, results.Count);
-            }
-        }
-
-        [ConditionalFact]
-        public virtual async Task Where_subquery_correlated_client_eval()
-        {
-            await AssertQuery<Customer>(
-                cs => cs.Take(5).OrderBy(c1 => c1.CustomerID).Where(c1 => cs.Any(c2 => c1.CustomerID == c2.CustomerID && c2.IsLondon)),
-                entryCount: 1);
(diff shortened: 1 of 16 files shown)


src/EFCore.Specification.Tests/Query/AsyncQueryTestBase.cs
```diff
@@ -22,13 +22,16 @@
 // ReSharper disable StringEndsWithIsCultureSpecific
 namespace Microsoft.EntityFrameworkCore.Query
 {
-    public abstract class AsyncQueryTestBase<TFixture> : IClassFixture<TFixture>
+    public abstract class AsyncSimpleQueryTestBase<TFixture> : AsyncQueryTestBase<TFixture>
```
maumar: File names will match in the final checkin, current state is just to make CR easier.

smitpatel: codeflow makes it easy :)

src/EFCore.Specification.Tests/Query/QueryTestBase.cs
```diff
@@ -80,7 +80,7 @@ public abstract class QueryTestBase<TFixture> : IClassFixture<TFixture>
             bool assertOrder = false,
             int entryCount = 0)
             where TItem1 : class
-            => AssertQuery(query, query, elementSorter, elementAsserter, assertOrder, entryCount);
+            => Fixture.QueryAsserter.AssertQuery(query, query, elementSorter, elementAsserter, assertOrder, entryCount, isAsync: false);
```
AndriySvyryd: You could extract this value to a property 'Async' , then derive 'AsyncQueryTestBase' from 'QueryTestBase' and override the 'Async' property there

AndriySvyryd:
>         public static void Cleanup(NorthwindContext context)

Are these still used anywhere?

---
Refers to: src/EFCore.Specification.Tests/TestModels/Northwind/NorthwindData.cs:121 in 32bf020. [](commit_id = 32bf02092b814f4fa33158354488d87948c20f1c, deletion_comment = False)

src/EFCore.Specification.Tests/TestUtilities/QueryAsserter.cs
```diff
@@ -165,30 +166,168 @@ public class QueryAsserter<TContext> : QueryAsserterBase
...
             Func<IQueryable<TItem1>, IQueryable<object>> query,
             Func<dynamic, object> elementSorter = null,
             Action<dynamic, dynamic> elementAsserter = null,
             bool assertOrder = false,
-            int entryCount = 0)
+            int entryCount = 0,
+            bool isAsync = false)
             where TItem1 : class
-            => AssertQuery(query, query, elementSorter, elementAsserter, assertOrder, entryCount);
+            => await AssertQuery(query, query, elementSorter, elementAsserter, assertOrder, entryCount, isAsync);
```
AndriySvyryd: You don't need async/await if the only call you await on is the last one

src/EFCore.Specification.Tests/TestUtilities/QueryAsserterBase.cs
```diff
@@ -76,25 +108,28 @@ public abstract class QueryAsserterBase
...
 
-        public abstract void AssertQueryScalar<TItem1, TResult>(
+        public abstract Task AssertQueryScalar<TItem1, TResult>(
             Func<IQueryable<TItem1>, IQueryable<TResult>> actualQuery,
             Func<IQueryable<TItem1>, IQueryable<TResult>> expectedQuery,
-            bool assertOrder = false)
+            bool assertOrder = false,
+            bool isAsync = false)
             where TItem1 : class
             where TResult : struct;
```
AndriySvyryd: The restriction on TResult seems unnecessary

src/EFCore.Specification.Tests/TestUtilities/QueryAsserter.cs
maumar: thats what I thought, but VS is complaining: "CS4014	Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call"

test/EFCore.SqlServer.FunctionalTests/Query/AsyncQuerySqlServerTest.cs
```diff
@@ -15,19 +15,15 @@
...
 namespace Microsoft.EntityFrameworkCore.Query
 {
-    public class AsyncQuerySqlServerTest : AsyncQueryTestBase<NorthwindQuerySqlServerFixture<NoopModelCustomizer>>
+    public class AsyncSimpleQuerySqlServerTest : AsyncSimpleQueryTestBase<NorthwindQuerySqlServerFixture<NoopModelCustomizer>>
     {
-        public AsyncQuerySqlServerTest(NorthwindQuerySqlServerFixture<NoopModelCustomizer> fixture, ITestOutputHelper testOutputHelper)
+        public AsyncSimpleQuerySqlServerTest(NorthwindQuerySqlServerFixture<NoopModelCustomizer> fixture, ITestOutputHelper testOutputHelper)
             : base(fixture)
         {
-            //#9074
```
AndriySvyryd: #9074 isn't fixed yet

src/EFCore.Specification.Tests/TestUtilities/QueryAsserter.cs
AndriySvyryd: you still need to return the Task

maumar:
new version up @AndriySvyryd

Issue #9214, opened 2017-07-19
DbFunctions: Schema not specified when calling DbFunction in projection


anpete:
Seeing this on preview2; may be fixed in RTM..
```
fail: Microsoft.EntityFrameworkCore.Database.Command[200102]
      Failed executing DbCommand (1ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
      SELECT ComputePostCount([b].[BlogId])
      FROM [Blogs] AS [b]
System.Data.SqlClient.SqlException (0x80131904): 'ComputePostCount' is not a recognized built-in function name.
```

pmiddleton:
Do you have a repo example?

pmiddleton:
I can't repo on the latest code.  Where are you setting the schema; globally, via attribute, or HasSchema on the function?

anpete:
Not setting it at all. It should be defaulting to 'dbo'.

pmiddleton:
Where are you expecting the default to be happening?  DbFunctions default to empty schema.  I have always seen the global schema default to empty as well.

anpete:
Maybe this is just a new requirement based on how functions need to be called in TSQL?

pmiddleton:
Adding a default of dbo for Sql Server makes sense.  I'm not sure the best way to do it.

We can add it during sql generation, but we need to key off of something.  Keying off of a null schema makes the most sense, but the built in IMethodCallTranslators are using null for schema.  They would have to change over to using a empty string.

Otherwise we would have to find a place in the dbFunction creation pipeline to derive off a class that can set the default value.

pmiddleton:
@anpete @ajcvickers 

Do either of you have any suggestions or examples on how to inject a factory into the relational annotations?  I would like to create a SqlDbFunction class which can set the default schema.  The issue I am having is that I don't see a clean way to inject a dbFunction factory into the RelationalModelAnnotations class.    The only approach I see is storing it in the IModel and that just doesn't seem like the right way to go.

Has the team had any discussions around getting factories into the annotations?

anpete:
It may be easier to have a SQL Server IModelAnnotationChangedConvention that just defaults the schema...

pmiddleton:
Thanks - I forgot those notifications were there.  I will take a look into that approach.

pmiddleton:
and yes I realize the irony in forgetting about it when I was the one who created it :)

smitpatel:
[...]

PR #9303, opened 2017-07-31
Default Sql Server db functions to DBO schema

pmiddleton:
Set the the default schema for db functions in Sql Server to dbo.

Addresses #9214

### DIFF
--- a/src/EFCore.Relational/Metadata/Internal/DbFunction.cs
+++ b/src/EFCore.Relational/Metadata/Internal/DbFunction.cs
@@ -47,6 +47,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
             return model[BuildAnnotationName(annotationPrefix, methodInfo)] as DbFunction;
         }
 
+        private readonly IMutableModel _model;
         private string _schema;
         private string _functionName;
 
@@ -81,7 +82,8 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
 
             MethodInfo = methodInfo;
 
-            model[BuildAnnotationName(annotationPrefix, methodInfo)] = this;
+            _model = model;
+            _model[BuildAnnotationName(annotationPrefix, methodInfo)] = this;
         }
 
         /// <summary>
@@ -106,9 +108,15 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        public virtual string DefaultSchema { get; [param: CanBeNull] set;}
+      
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
         public virtual string Schema
         {
-            get => _schema;
+            get => _schema ?? _model.Relational().DefaultSchema ?? DefaultSchema;
             set => SetSchema(value, ConfigurationSource.Explicit);
         }
 
--- a/src/EFCore.SqlServer/Metadata/Conventions/Internal/SqlServerDbFunctionConvention.cs
+++ b/src/EFCore.SqlServer/Metadata/Conventions/Internal/SqlServerDbFunctionConvention.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Reflection;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    public class SqlServerDbFunctionConvention : RelationalDbFunctionConvention
+    {
+        protected override void ApplyCustomizations(InternalModelBuilder modelBuilder, string name, Annotation annotation)
+        {
+            base.ApplyCustomizations(modelBuilder, name, annotation);
+
+            ((DbFunction)annotation.Value).DefaultSchema = "dbo";
+        }
+    }
+}
--- a/src/EFCore.Relational.Specification.Tests/TestModels/Northwind/NorthwindRelationalContext.cs
+++ b/src/EFCore.Relational.Specification.Tests/TestModels/Northwind/NorthwindRelationalContext.cs
@@ -33,9 +33,9 @@ namespace Microsoft.EntityFrameworkCore.TestModels.Northwind
             modelBuilder.HasDbFunction(methodInfo)
                 .HasTranslation(args => new SqlFunctionExpression("len", methodInfo.ReturnType, args));
 
-            modelBuilder.HasDbFunction(
-                typeof(NorthwindRelationalContext)
-                    .GetRuntimeMethod(nameof(IsDate), new[] { typeof(string) }));
+            modelBuilder.HasDbFunction(typeof(NorthwindRelationalContext)
+                .GetRuntimeMethod(nameof(IsDate), new[] { typeof(string) }))
+                .HasSchema("");
         }
 
         public enum ReportingPeriod
@@ -56,13 +56,13 @@ namespace Microsoft.EntityFrameworkCore.TestModels.Northwind
             throw new Exception();
         }
 
-        [DbFunction(Schema = "dbo", FunctionName = "EmployeeOrderCount")]
+        [DbFunction(FunctionName = "EmployeeOrderCount")]
         public static int EmployeeOrderCount(int employeeId)
         {
             throw new NotImplementedException();
         }
 
-        [DbFunction(Schema = "dbo", FunctionName = "EmployeeOrderCount")]
+        [DbFunction(FunctionName = "EmployeeOrderCount")]
         public static int EmployeeOrderCountWithClient(int employeeId)
         {
             switch (employeeId)
@@ -74,13 +74,13 @@ namespace Microsoft.EntityFrameworkCore.TestModels.Northwind
             }
         }
 
-        [DbFunction(Schema = "dbo")]
+        [DbFunction]
         public static bool IsTopEmployee(int employeeId)
         {
             throw new NotImplementedException();
         }
 
-        [DbFunction(Schema = "dbo")]
+        [DbFunction]
         public static int GetEmployeeWithMostOrdersAfterDate(DateTime? startDate)
         {
             throw new NotImplementedException();
--- a/src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
+++ b/src/EFCore.Relational/Metadata/Conventions/Internal/RelationalDbFunctionConvention.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Linq;
 using System.Reflection;
+using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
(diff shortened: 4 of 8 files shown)


src/EFCore.SqlServer/Metadata/Conventions/Internal/SqlServerDbFunctionConvention.cs
```diff
@@ -0,0 +1,37 @@
...
+                && annotation != null
+                && oldAnnotation == null)
+            {
+                var dbFunction = (DbFunction)annotation.Value ;
+
+                if(dbFunction.Schema == null)
+                {  
+                    var dbFunctionBuilder = new InternalDbFunctionBuilder(dbFunction);
+
+                    dbFunctionBuilder.HasSchema("dbo", ConfigurationSource.Convention);
```
smitpatel: Should this default to "dbo" always or whatever is defined on model?

pmiddleton: I'm not understanding your question.  With this change if you don't set a schema it will default to dbo.  

One thing I just thought of is if we should use the global schema if it is set or go with this default.

@anpete thoughts?

smitpatel: > One thing I just thought of is if we should use the global schema if it is set or go with this default.

That's what I meant. If user has done 
`modelBuilder.HasDefaultSchema("MySchema");`
Then should function go there?

smitpatel: Should it be `IModelBuiltConvention`?
So that once the model is fully built, plug in missing schemas in DbFunction based on default if set on model or "dbo"

pmiddleton: Would there every be a reason someone would try to pull the schema back out of a function before finishing building the entire model?  That is the only thing I can think of where waiting for `IModelBuiltConvention` would be different.

smitpatel: Suppose user defines a function without configuring schemas & then define the default schema.

pmiddleton: Ok I switched it over to `IModelBuiltConvention`  I need to pop a few unit tests to deal with the defaultSchema vs defaulting to dbo still.

pmiddleton:
Updated and unit tests added.

smitpatel:
Can you add a functional test when default schema is specified which is different from dbo and function is defined on that schema and query picks up the schema from model correctly.

pmiddleton:
Let me make sure I am following what you are asking for.

You want the default schema set to "abc" and the function set to "xyz" and then make sure the sql uses "xyz"?  By functional test do you want it to actually hit a database and run a query?

smitpatel:
Since the convention does not do anything if model has DefaultSchema set,
What if a function is defined with no schema & model has default schema, does it produce a valid SQL for it?
One of the easiest way to verify SQL validity is to actually hit the database and run the query.

smitpatel:
I believe current functional tests covers when function is defined without schema and we use the "dbo"
We should have functional test for 
function defined without schema & with model.defaultschema set and function uses that.
function defined with schema & with model.defaultschema set & function uses the schema from function only. 
You can probably add new tests in QueryBugsSqlServerTest

pmiddleton:
If there is no schema set on the function but a default schema is set then the default schema is used.  If you don't want the default schema to be used at all then you need to set the schema on the function to an empty string.  

For Sql Server that use case would only be for calling built in functions, all udf calls require a schema.

The logic order currently is 

1.  Use the schema set directly on the function.  This schema can be an empty string.
2.  Use the default schema.  
3.  Use dbo

Any function test that hit the db need to be in `UdfDbFunctionSqlServerTests`.  That uses the Northwind database which has the setup script which we can use to create the functions in the database.

smitpatel:
> If there is no schema set on the function but a default schema is set then the default schema is used.

That's the scenario I want tested because I have doubts on that.

We added `UdfDbFunctionSqlServerTests` for UDF testing but we don't modify northwind.sql generally. At that time due to time restrictions we went that way. The issue is northwind is not re-created always.
For AppVeyor/Travis it works fine because we start with new VM which does not have any database at all so always northwind will be created from latest script.

For all local dev machines & CI machines which are not dropping database, this causes mismatch since there will be outdated northwind in sqlserver. And that leads to failure unless northwind database is manually dropped.

src/EFCore.SqlServer/Metadata/Conventions/Internal/SqlServerDbFunctionConvention.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using System.Linq;
+using System.Reflection;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    public class SqlServerDbFunctionConvention : IModelBuiltConvention
```
smitpatel: @AndriySvyryd - Name sounds good to you?

pmiddleton:
I don't see a `QueryBugsSqlServerTest` in the codebase.

Do you not want me to modify the northwind.sql and find a different approach for creating a new databsae and creating the function for each run?

smitpatel:
Ah, its `QueryBugsTest` only.
Best would be to have UDF functions test class to use northwind model but use a different database which would get recreated every time so adding more testing around udf in future does not affect northwind database. But that may be more work. If you don't have enough time for it, feel free to put it in `QueryBugsTest`.

pmiddleton:
I actually have a big family wedding the rest of this week/weekend so I won't be able to get these changes in until early next week.

smitpatel:
@pmiddleton - No rush, we have a lot of time before next release 😄

pmiddleton:
@smitpatel - I added the unit tests you were looking for.  Let me know what you think.

smitpatel:
@pmiddleton - Thanks I was looking for precisely that test since I was not able to understand how the default model schema was set on function. After looking more in code today I found out the `RelationalDbFunctionConvention` which sets name & schema from attribute or defaults.

Should `RelationalDbFunctionConvention` be ModelBuiltConvention like above rather than annotation changed convention? (it is only applied when the annotation is created & not changed). Then we may override it in SqlServer, to suggest further default to be "dbo". The downside is name & schema will be plugged in once the model is built. Not sure if that is critical to have. Upside lesser convention invocations.

What do you think?
also cc: @AndriySvyryd , @anpete

pmiddleton:
It depends if we want the user to be able to read the name and schema, set via the attribute, during OnModelCreating.  As it stands now you can, if we switch to a ModelBuiltConvention you won't be able to.

That would also be a minor change to end user behavior from 2.0 to 2.1.

smitpatel:
The test in QueryBugsTest fail if function is defined using Attribute without specifying schema. The `RelationalDbFunctionConvention` sets schema to model.DefaultSchema but if the function is found using attribute then at that time OnModelCreating hasn't run so we don't know default schema hence it fails to set any schema.

Further, if user set default schema then define function then set default schema to something else then we are not updating the schema in db function. This is because `RelationalDbFunctionConvention` would run only when adding annotation for DbFunction first time. Since we follow last one win strategy, all our metadata object goes to schema defined 2nd time but the function goes to first schema causing error. This kind of setting default schema again would be used in scenario like extensible class library using EF (like Identity), when final consumer would want to set default schema to something else)

smitpatel:
Though above is in existing code and out of the scope of PR.
@pmiddleton - Can you rebase on latest dev?

smitpatel:
filed #9360

pmiddleton:
Rebased.

pmiddleton:
Based on your findings I'm leaning towards just applying all of the defaults in a ModelBuiltConvention.  

We can make `RelationalDbFunctionConvention` a ModelBuiltConvention and have it deal with attribute & default schema.  We can then have `SqlServerDbFunctionConvention` derive and override `RelationalDbFunctionConvention` to deal with dbo.

Thoughts?

smitpatel:
@AndriySvyryd needs to approve it.

AndriySvyryd:
Instead of this being a convention you could put the check in `DbFunction.Schema` getter. This would also fix #9360

AndriySvyryd:
We would still need a SQL Server convention, but it would set a new property - `DbFunction.DefaultSchema` to "dbo" instead of `DbFunction.Schema`. And this convention needs only to run once, when the DbFunction is created.

pmiddleton:
I added a `DbFunction.DefaultSchema` and set that in the convention when the function is created.  I also added some new unit tests.  I think this gets us to where we want to be for #9360 and #9214

src/EFCore.Relational/Metadata/Internal/DbFunction.cs
```diff
@@ -106,9 +108,15 @@ private static string BuildAnnotationName(string annotationPrefix, MethodBase me
...
+        public virtual string DefaultSchema { get; [param: CanBeNull] set;}
+      
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
         public virtual string Schema
         {
-            get => _schema;
+            get => _schema ?? new RelationalModelAnnotations(_model).DefaultSchema ?? DefaultSchema;
```
smitpatel: would this be same as `_model.Relational().DefaultSchema`?

pmiddleton: yes - I will switch it to add the abstraction back in.

src/EFCore.SqlServer/Metadata/Conventions/Internal/SqlServerDbFunctionConvention.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    public class SqlServerDbFunctionConvention : IModelAnnotationChangedConvention
+    {
+        public virtual Annotation Apply(InternalModelBuilder modelBuilder, string name, Annotation annotation, Annotation oldAnnotation)
+        {
+            Check.NotNull(modelBuilder, nameof(modelBuilder));
+            Check.NotNull(name, nameof(name));
+
+            if (name.StartsWith(RelationalAnnotationNames.DbFunction, StringComparison.OrdinalIgnoreCase)
```
AndriySvyryd: Use `StringComparison.Ordinal`

AndriySvyryd: Also in `RelationalDbFunctionConvention`.
Actually you can derive this from `RelationalDbFunctionConvention` and replace it in `SqlServerConventionSetBuilder`

pmiddleton:
@AndriySvyryd  - Changes made

src/EFCore.SqlServer/Metadata/Conventions/Internal/SqlServerDbFunctionConvention.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal
+{
+    public class SqlServerDbFunctionConvention : RelationalDbFunctionConvention
+    {
+        public override Annotation Apply(InternalModelBuilder modelBuilder, string name, Annotation annotation, Annotation oldAnnotation)
+        {
+            Check.NotNull(modelBuilder, nameof(modelBuilder));
+            Check.NotNull(name, nameof(name));
+
+            base.Apply(modelBuilder, name, annotation, oldAnnotation);
```
smitpatel: Calling base this way means the check is happening twice. Instead a helper function to do the task of setting name/schema which can be overridden in derived convention could be more DRY.

pmiddleton:
@smitpatel - switched to a helper method.

smitpatel:
Merged via 7529563ec9136359a78a223f6e69441641a03924

Thanks for contribution @pmiddleton

julielerman:
Hey Paul et alia, reading the two issues on this, my understanding is that dbo is the conventional default schema. But (on EF Core 2.0.1) my queries using a function are failing (sql server returns function not found error) unless I explicitly set the schema to dbo in the function ala

```
 [DbFunction(Schema ="dbo")]
        public static string EarliestBattleFoughtBySamurai(int samuraiId)
        {
            throw new Exception();
        }
```
Am I misunderstanding something?
Thanks!

smitpatel:
@julielerman - Yes, that would be expected behavior in 2.0.1. The issue is fixed in 2.1.0-preview1-final

julielerman:
yeah .just found the note in https://github.com/aspnet/EntityFrameworkCore/issues/9663 :)
thanks!

pmiddleton:
@julielerman  - Yup I missed that one in 2.0.

pmiddleton:
@julielerman - 2.1.0-preview1-final will also allow you to use instance methods as well as static methods for db functions.  See #9755

PR #9315, opened 2017-08-01
RevEng: Remove TableSelectionSet

smitpatel:


src/EFCore.SqlServer/Internal/SqlServerLoggerExtensions.cs
```diff
@@ -227,13 +227,12 @@ private static string ByteIdentityColumnWarning(EventDefinitionBase definition,
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        public static void ForeignKeyColumnMissingWarning(
+        public static void ForeignKeyTableMissingWarning(
             [NotNull] this IDiagnosticsLogger<DbLoggerCategory.Scaffolding> diagnostics,
-            [CanBeNull] string columnName,
             [CanBeNull] string foreignKeyName,
             [CanBeNull] string tableName)
             // No DiagnosticsSource events because these are purely design-time messages
```
smitpatel: Even though comment says no `DiagnosticsSource` events there are still events defined in `SqlServerEventId`. Since that class is public but not internal I cannot remove event ids. (hence left cleaning up other verbose logging)

test/EFCore.Sqlite.Design.FunctionalTests/ReverseEngineering/SqliteE2ETestBase.cs
```diff
@@ -364,7 +364,7 @@ FOREIGN KEY (UserAltId) REFERENCES User (AltId)
 
                 var results = Generator.Generate(
                     testStore.ConnectionString,
-                    Enumerable.Empty<string>(),
+                    new List<string> {"User", "Comment"}.AsEnumerable(),
```
bricelam:
```csharp
new[] { "User", "Comment" }
```

src/EFCore.SqlServer/Internal/SqlServerLoggerExtensions.cs
smitpatel: Couldn't remove any thing.

smitpatel: Removing from logger and having event id fails a test.

PR #9332, opened 2017-08-03
Test: Update EOL pattern

### LINKED ISSUE #9147: Test: CSharpMigrationsGeneratorTests are newline dependent
Some tests fail if you download zip file of repo and run build.
```
CSharpMigrationsGeneratorTest.Migrations_compile [FAIL]
CSharpMigrationsGeneratorTest.Snapshots_compile [FAIL]
```

bricelam:
Uses `ignoreLineEndingDifferences` on `Equal` and `Environment.NewLine` with `StartsWith`, `Contains` & `EndsWith`.

Fixes #9147

### DIFF
--- a/src/EFCore.Relational.Specification.Tests/TestUtilities/TestSqlLoggerFactory.cs
+++ b/src/EFCore.Relational.Specification.Tests/TestUtilities/TestSqlLoggerFactory.cs
@@ -16,17 +16,15 @@ namespace Microsoft.EntityFrameworkCore.TestUtilities
 {
     public class TestSqlLoggerFactory : ILoggerFactory
     {
-        private const string FileLineEnding = @"
+        private const string FileNewLine = @"
 ";
 
-        private static readonly string _newLine = Environment.NewLine;
+        private static readonly string EOL = Environment.NewLine;
 
         public void AssertBaseline(string[] expected, bool assertOrder = true)
         {
             var sqlStatements
-                = _logger.SqlStatements
-                    .Select(sql => sql.Replace(Environment.NewLine, FileLineEnding))
-                    .ToList();
+                = _logger.SqlStatements;
 
             try
             {
@@ -34,21 +32,23 @@ namespace Microsoft.EntityFrameworkCore.TestUtilities
                 {
                     for (var i = 0; i < expected.Length; i++)
                     {
-                        Assert.Equal(expected[i], sqlStatements[i]);
+                        Assert.Equal(expected[i], sqlStatements[i], ignoreLineEndingDifferences: true);
                     }
                 }
                 else
                 {
                     foreach (var expectedFragment in expected)
                     {
-                        Assert.Contains(expectedFragment, sqlStatements);
+                        Assert.Contains(
+                            expectedFragment.Replace("\r", string.Empty).Replace("\n", EOL),
+                            sqlStatements);
                     }
                 }
             }
             catch
             {
                 var methodCallLine = Environment.StackTrace.Split(
-                        new[] { Environment.NewLine },
+                        new[] { EOL },
                         StringSplitOptions.RemoveEmptyEntries)[4]
                     .Substring(6);
 
@@ -56,7 +56,7 @@ namespace Microsoft.EntityFrameworkCore.TestUtilities
                 var lineIndex = methodCallLine.LastIndexOf("line", StringComparison.Ordinal);
                 var lineNumber = lineIndex > 0 ? methodCallLine.Substring(lineIndex) : "";
 
-                const string indent = FileLineEnding + "                ";
+                const string indent = FileNewLine + "                ";
 
                 var currentDirectory = Directory.GetCurrentDirectory();
                 var logFile = currentDirectory.Substring(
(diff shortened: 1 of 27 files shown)


.gitattributes
```diff
@@ -1,4 +1,4 @@
 * text=auto
-*.cs diff=csharp
+*.cs diff=csharp eol=lf
```
bricelam: Will revert. Using to test on AppVeyor.

src/EFCore.Specification.Tests/TestUtilities/AssertExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        /// <param name="expectedEndString">The string expected to be at the end of the string</param>
+        /// <param name="actualString">The string to be inspected</param>
+        /// <param name="fixEOL">If set to <c>true</c>, treats verbatim new lines in <paramref name="expectedEndString"/> as <see cref="Environment.NewLine"/>.</param>
+        /// <exception cref="ContainsException">Thrown when the string does not end with the expected string</exception>
+        public static void EndsWith(string expectedEndString, string actualString, bool fixEOL)
+            => EndsWith(FixEOL(expectedEndString, fixEOL), actualString);
+
+        /// <summary>
+        /// Verifies that two strings are equivalent.
```
smitpatel: tab

src/EFCore.Specification.Tests/TestUtilities/AssertExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        /// <param name="expected">The expected string value.</param>
+        /// <param name="actual">The actual string value.</param>
+        /// <param name="fixEOL">If set to <c>true</c>, treats verbatim new lines in <paramref name="expected"/> as <see cref="Environment.NewLine"/>.</param>
+        /// <param name="ignoreCase">If set to <c>true</c>, ignores cases differences. The invariant culture is used.</param>
+        /// <param name="ignoreLineEndingDifferences">If set to <c>true</c>, treats \r\n, \r, and \n as equivalent.</param>
+        /// <param name="ignoreWhiteSpaceDifferences">If set to <c>true</c>, treats spaces and tabs (in any non-zero quantity) as equivalent.</param>
+        /// <exception cref="EqualException">Thrown when the strings are not equivalent.</exception>
+        public static void Equal(
```
AndriySvyryd: Consider renaming to something like EqualMultiline and remove `fixEOL` parameter

smitpatel:
Can we avoid these changes for query tests altogether? Query tests using assert baseline are doing correct processing. Those tests which are verifying sql but not converted yet, are having processing too. When we have proper infra in place then creating question of which method to use is bit "unwelcomed"

bricelam:
Personally, I'd like us to take a stand and say that checking out or compiling the code on one platform and running it on another is unsupported. If we did that, we could remove all new line handling in our tests.

src/EFCore.Specification.Tests/TestUtilities/AssertExtensions.cs
bricelam: That was one option I wanted to discuss. This was the idiomatic, vanilla version that I wanted to use to start the discussion in a design meeting.

src/EFCore.Specification.Tests/TestUtilities/AssertExtensions.cs
bricelam: I'm awfully tired of satisfying R# on this point.

bricelam: (But I don't mind if someone wants to run R# cleanup periodically to catch these.)

smitpatel: Is it R# thing? I thought that was rules we had to adheres to.

bricelam: Yeah, it just happens to be the way R# formats them. I don't think we ever formalized it as a rule.

bricelam:
(But that might fall down since we actually ship the Windows compiled specification tests.)

bricelam:
(Updated with the changes we discussed.)

src/EFCore.Relational.Specification.Tests/TestUtilities/TestSqlLoggerFactory.cs
```diff
@@ -16,55 +16,55 @@ namespace Microsoft.EntityFrameworkCore.TestUtilities
...
+                        Assert.Equal(expected[i], sqlStatements[i], ignoreLineEndingDifferences: true);
                     }
                 }
                 else
                 {
                     foreach (var expectedFragment in expected)
                     {
-                        Assert.Contains(expectedFragment, sqlStatements);
+                        Assert.Contains(
+                            expectedFragment.Replace("\r", string.Empty).Replace("\n", EOL),
```
bricelam: Note: Fixing up the expected value here since it was the path of least resistance.

Issue #9111, opened 2017-07-07
Add more test coverage for tooling/RevEng


ajcvickers:

smitpatel:
@bricelam - Heads up. I am going to start working on modifying reveng tests as we discussed earlier.

smitpatel:
Summary of a recent discussion about testing of tools interface
There are 3 splits of tests needed.
1. PMC/dotnet ef calling into mock ef.exe with correct parameters. 
2. ef.exe calling into mock executor.
3. Executor calling into right entry points into (mock) design assembly.

Above would verify, from command to design assembly entry point pipeline is working correctly. 
Inside design assembly there are tests based on the scenarios.

bricelam:
We also intend to add some IModel -> C# tests.

PR #9335, opened 2017-08-03
Test: Add SqlServerDatabaseModelFactoryTest which does not use Design package

### LINKED ISSUE #9324: RevEng: Scaffold Non clustered for PK
This work is opposite of #8687 

The comment https://github.com/aspnet/EntityFramework/pull/8950/files#r123785377 was correct.
Using migrations on generated code causes issues.

The issue is as follows:
SqlServer by default creates clustered PK if there is no other "constraint" which is marked as unique. Emphasize on constraint because constraints can be declared in create table but indexes cannot be.
Since Create index is not part of create table, if there is a clustered index then SqlServer wouldn't know about it while creating table and create a clustered PK by default and later adding clustered index throws.
Since we scaffold index even for constraints, we need to have non-clustered on PK always marked so that migrations can generate correct SQL later.

smitpatel:
part of #9111 

Resolves #9324 

This still leaves some rev eng tests in SqlServer which should be moved to Design.FunctionalTests


smitpatel:
@ErikEJ @roji - This is the new pattern of testing design stuff we are moving forward.

Since there is no separate design package for providers,
- Unit tests are going to Provider.Tests
- Functional tests (esp reveng since it needs database connection) are moving to Provider.FunctionalTests

The main higher level change is, since tooling (which lives in .Design package) is separated from providers, we would be dropping `Design.Specification.Tests` project.
Instead provider just verify their component, EF.Design will verify itself & things would work together. That means no E2E testing for migration/reveng.

If provider wants to write E2E tests then they are free to do so. There is not much common code in specs package at present too.

Issue #9098, opened 2017-07-07
Revisit functional test pattern


AndriySvyryd:
We now have a new way to clean the database and setup the services, so the template for functional tests can be simplified.

smitpatel:
What is the new way? 😀

AndriySvyryd:
The "new" here is relative to the last time the pattern was revised, about 2 years ago.

ajcvickers:
"Make it easy to add a new test!" - @AndriySvyryd

AndriySvyryd:
Also ensure `SqlServerTestStore.ExecuteCommand` and `SqlServerTestStore.ExecuteCommandAsync` close the connection

smitpatel:
Once this issue is fixed, can we have a design meeting slot on how to add a new fixture & testbase?

PR #9349, opened 2017-08-07
Revisit Functional Test Pattern Part 5: New Begginnings

AndriySvyryd:
Switch relational specification tests to SharedStoreFixtureBase
Make cross store test to actually share a model
Reenable SqlServerDatabaseCreatorTest on CI

Part of #9098


smitpatel:
This has merge conflicts because of my ongoing changes in reveng tests.
Can you leave reveng tests (perhaps design.Specs.tests) as is since most of them are going away.

src/EFCore.Relational.Specification.Tests/TableSplittingTestBase.cs
```diff
@@ -3,21 +3,24 @@
...
 {
-    public abstract class TableSplittingTestBase<TTestStore>
-        where TTestStore : TestStore
+    public abstract class TableSplittingTestBase
     {
         [Fact(Skip = "#8973")]
         public void Can_query_shared()
         {
-            using (var store = CreateTestStore(OnModelCreating))
+            using (CreateTestStore(OnModelCreating))
```
smitpatel: Should this test use fixture instead of creating test store in each test?

test/EFCore.Relational.Tests/RelationalOptionsExtensionTest.cs
```diff
@@ -26,12 +27,6 @@ public void Can_set_Connection()
         }
 
         [Fact]
-        public void Throws_when_setting_Connection_to_null()
```
smitpatel: Does it not throw anymore?

test/EFCore.Sqlite.FunctionalTests/Scaffolding/SqliteDatabaseModelFactoryTest.cs
```diff
@@ -10,63 +10,39 @@
...
 using Microsoft.EntityFrameworkCore.TestUtilities;
+using Microsoft.EntityFrameworkCore.Utilities;
 using Microsoft.Extensions.Logging;
 using Xunit;
 
 // ReSharper disable InconsistentNaming
-
 namespace Microsoft.EntityFrameworkCore.Scaffolding
 {
-    public class SqliteDatabaseModelFactoryTest : IDisposable
```
smitpatel: I stumbled upon this class not using fixture (perhaps creating new db is really cheap in sqlite).

Instead of modifying all tests this way, we should adopt the pattern same as SqlServer i.e. clean up script. So we use one db only but we don't have any other objects than what test needs.

test/EFCore.SqlServer.FunctionalTests/SqlServerDatabaseCreatorTest.cs
```diff
@@ -20,8 +20,6 @@
 namespace Microsoft.EntityFrameworkCore
 {
     // Tests are split into classes to enable parralel execution
-    // These tests are disabled in TeamCity as they are flaky, see #8615
-    [SqlServerCondition(SqlServerCondition.IsNotTeamCity)]
```
smitpatel: Are we certain this works? :trollface: We can test out feature branch

src/EFCore.Relational.Specification.Tests/TableSplittingTestBase.cs
AndriySvyryd: The model is different in the tests, so there's really nothing that can be shared in a fixture

test/EFCore.Relational.Tests/RelationalOptionsExtensionTest.cs
AndriySvyryd: It does, but we usually don't test precondition checks

test/EFCore.Sqlite.FunctionalTests/Scaffolding/SqliteDatabaseModelFactoryTest.cs
AndriySvyryd: Maybe you could do this as part of the reveng test cleanup 😉

test/EFCore.SqlServer.FunctionalTests/SqlServerDatabaseCreatorTest.cs
AndriySvyryd: We did, seems to work

test/EFCore.Sqlite.FunctionalTests/Scaffolding/SqliteDatabaseModelFactoryTest.cs
smitpatel: Can you leave test as is then? It would be more work if you modify this file like this. 😞

PR #9362, opened 2017-08-08
Revisit Functional Test Pattern Part 6: The Purge

### LINKED ISSUE #9098: Revisit functional test pattern

AndriySvyryd:
Enable DbContext pooling for most of the functional tests
Move all test utilities to M.EFC.TestUtilities
Refactor ITestStoreFactory
Move throwing SaveChanges test from ThrowingMonsterStateManager to OptimisticConcurrencySqlServerTest to avoid using a non-standard service

Fixes #9098


src/EFCore/DbContext.cs
```diff
@@ -16,7 +16,7 @@
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Query.Internal;
-using Microsoft.EntityFrameworkCore.Utilities;
+using Microsoft.EntityFrameworkCore.TestUtilities;
```
smitpatel: Why does source code depends on TestUtilities? It would be better to keep utilities separate. (cannot break source utilities but can break test one)

AndriySvyryd: Should have reviewed the files before sending the PR 🤦‍♂️

AndriySvyryd:
@smitpatel Updated

src/EFCore.Specification.Tests/MonsterFixupTestBase.cs
```diff
@@ -156,46 +155,6 @@ private void Can_build_monster_model_and_seed_data_using_navigations_with_deferr
         }
 
         [Fact]
-        public virtual void Store_generated_values_are_discarded_if_saving_changes_fails()
```
smitpatel: Intentional?

AndriySvyryd: Yes, the scenario was moved to `OptimisticConcurrencySqlServerTest`

test/EFCore.SqlServer.FunctionalTests/MonsterFixupSqlServerTest.cs
```diff
@@ -3,27 +3,16 @@
...
-                .AddEntityFrameworkSqlServer();
-
-            if (throwingStateManager)
-            {
-                serviceCollection.AddScoped<IStateManager, ThrowingMonsterStateManager>();
-            }
-
-            return serviceCollection.BuildServiceProvider();
-        }
+        protected override IServiceProvider CreateServiceProvider()
```
smitpatel: Why this test needs to create own service provider?

test/EFCore.Relational.Tests/Utilities/RelationalLoggerExtensionsTest.cs
```diff
@@ -7,7 +7,7 @@
 using Microsoft.EntityFrameworkCore.Storage.Internal;
 using Xunit;
 
-namespace Microsoft.EntityFrameworkCore.Utilities
+namespace Microsoft.EntityFrameworkCore.TestUtilities
```
smitpatel: Is this correct namespace? & it does not correspond to the folder name.

test/EFCore.SqlServer.FunctionalTests/MonsterFixupSqlServerTest.cs
AndriySvyryd: Fully converting this test to the new pattern will take more time. This change is the first step.

PR #9363, opened 2017-08-08
RevEng: Remove redundant warnings & checks

### LINKED ISSUE #9293: ToListAsync never returns back after query execution
We just upgraded to EF Core 2.0 Preview 2 and I have seen this issue in EF COre 1.1 

For some of our queries we want to fetch parent records, but then we also find the related child records in one db call. 

so we have our query outlined as this

await query.Select(parent=> new 
{
ParentPropertiesProjections..
ChildCollectionModel = parent.ChildCollection.Select(t=> ....)
}.ToListAsync();

I see in the logs that EF does fire the query to get all the parent tables data, but doesn't produce any more logs after that and then the query never returns back. 

If this scenario is not supported, then the EF core framework should throw an exception for such instead of silently killing the thread. 

I don't see anything in logs at all to provide you details on it. 

Any guidance is greatly appreciated. 

I am running EF Core 2.0 on Windows 10 machine with VS 2017 Enterprise 15.3 Preview 6 against Sql Server

### LINKED ISSUE #9323: RevEng: Default/Computed Value processing
At present, DefaultValueSql does not set ValueGenerated on DatabaseColumn but ComputedValueSql sets it to AddOrUpdate.
If we are assuming that calling `HasDefaultValueSql` will set ValueGenerated by itself then same should also be true for `HasComputedValueSql` hence setting ValueGenerated for it is redundant. 
If it is incorrect assumption then, DefaultValue should set to ValueGenerated.OnAdd.

Identity must set ValueGenerated.OnAdd & rowversion must set OnAddOrUpdate.

### LINKED ISSUE #9333: RevEng: Warnings for null names for objects which requires name
We have "warnings" for sequence/column/index not having names but it is impossible to create them without name. There seem to be no way to actually hit that code path. Should we have warnings for them?

### LINKED ISSUE #9334: RevEng: Warnings for skipped table
At present when we encounter index/fk from a table which is not included by user, we generate "warning" that we couldn't find the table. That means for a skipped table we would generate multiple warnings even though that was user intent. Should we remove this unnecessary warnings?

smitpatel:
Set metadata appropriately

Resolves #9323 
resolves #9333 
resolves #9334

### DIFF
--- a/src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
+++ b/src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
@@ -294,12 +294,6 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
                         DisplayName(sequence.Schema, sequence.Name), sequence.StoreType, sequence.IsCyclic,
                         sequence.IncrementBy, sequence.StartValue, sequence.MinValue, sequence.MaxValue);
 
-                    if (string.IsNullOrEmpty(sequence.Name))
-                    {
-                        Logger.SequenceNotNamedWarning();
-                        continue;
-                    }
-
                     if (_defaultSequenceMinMax.ContainsKey(sequence.StoreType))
                     {
                         var defaultMin = _defaultSequenceMinMax[sequence.StoreType][0];
@@ -426,8 +420,7 @@ namespace Microsoft.EntityFrameworkCore.Scaffolding.Internal
             THEN c.max_length
             ELSE null
         END AS int) AS [max_length],
-    c.is_identity,
-    c.is_computed
+    c.is_identity
 FROM sys.index_columns ic
     RIGHT JOIN (SELECT * FROM sys.indexes WHERE is_primary_key = 1) AS i ON i.object_id = ic.object_id AND i.index_id = ic.index_id
     RIGHT JOIN sys.columns c ON ic.object_id = c.object_id AND c.column_id = ic.column_id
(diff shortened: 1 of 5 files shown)


src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
```diff
@@ -294,12 +294,6 @@ private void GetSequences()
                         DisplayName(sequence.Schema, sequence.Name), sequence.StoreType, sequence.IsCyclic,
                         sequence.IncrementBy, sequence.StartValue, sequence.MinValue, sequence.MaxValue);
 
-                    if (string.IsNullOrEmpty(sequence.Name))
-                    {
-                        Logger.SequenceNotNamedWarning();
-                        continue;
-                    }
-
```
bricelam: :tada:

src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
```diff
@@ -458,24 +451,17 @@ ELSE null
...
 
                     if (!_tables.TryGetValue(SchemaQualifiedKey(tableName, schemaName), out var table))
                     {
                         Logger.ColumnSkipped(DisplayName(schemaName, tableName), columnName);
                         continue;
                     }
 
-                    if (string.IsNullOrEmpty(columnName))
-                    {
-                        Logger.ColumnNotNamedWarning(DisplayName(schemaName, tableName));
```
bricelam: Can we remove these resources and APIs? If not, we should mark them as obsolete.

smitpatel: I cannot remove the APIs ( 😢 )
The logger methods are in internal namespace. The resources are in internal namespace. but there is a test which complains that for each event id there should be logger method with same name. Are we supposed to mark things in internal namespace as obsolete? 
EventIds are public so cannot break them.

bricelam: Maybe we should allow marking event IDs as obsolete in the test so it doesn't fail for them.

smitpatel: I am leaving this as of for now. Will file an issue.

Issue #3910, opened 2015-11-27
Query: Support Include/ThenInclude for navigation on derived type


davidroth:
Is there a solution for using `ThenInclude` with polymorphic types?

Example model:

``` csharp
public abstract class BasePosition : Entity
{
    public Order Order { get; set; }
}

public class GroupType : Entity
{}

public class GroupPosition : BasePosition
{
      public GroupType GroupType { get; set; }
}

public class SalesPosition : BasePosition
{
    public GroupPosition Group { get; set; }
}

public class Order : Entity
{
    private List<BasePosition> positions;

    public ICollection<BasePosition> Positions 
    => positions ?? (positions = new List<BasePosition>());
}
```

Example query:

``` csharp
context.Set<Order>()
.Include(x => x.Positions)
    .ThenInclude<SalesPosition>(x => x.Group) // Pseudo-Code, does obviously not work
    .ThenInclude<GroupPosition>(x => x.GroupType) // Pseudo-Code, does obviously not work
```

AFAIK  the only way this currently works is by manually querying the derived types after loading the base shape:

``` csharp
var o = context.Set<Order>()
.Where(o => o.Id == 1234)
.Single();

context.Set<GroupPosition>()
    .Where(x => x.Order == o)
    .Include(x => x.GroupType)
    .Load();

context.Set<SalesPosition>()
    .Where(x => x.Order == o)
    .Include(x => x.Group)
    .Load();
```

divega:
@davidroth As far as I know EF currently doesn't support any pattern you could use to express that you want query results to include navigations that only exist in derived types. Cc @anpete in case I am wrong.

As for how such pattern could look like, here is something we thought about in the past applied to the new ThenInclude method:

``` C#
context.Set<Order>()
    .Include(x => x.Positions)
    .ThenInclude(x => (x as SalesPosition).Group) 
    .Include(x => x.Positions)
    .ThenInclude(x => (x as GroupPosition).GroupType) 
```

Of course, there are probably other things that could work and for instance the null-conditional operator would be better than this if it was supported in LINQ expressions.

rowanmiller:
@divega @davidroth I think this same issue applies to Include as well as ThenInclude right? If the type you are querying is a base type then there is no way to Include navigations defined on derived types in the hierarchy.

divega:
Yes, it would apply to both. Only this particular example had it in `ThenInclude()`.

gregbty:
[...]

PR #9364, opened 2017-08-09
Fix to #3910 - Query: Support Include/ThenInclude for navigation on derived type

maumar:
It was not possible to specify include on a derived type. Include was very strict when computing navigation paths from Include method and generated code wasn't accounting for type discrepancies between caller and navigation.

Fix is to allow specifying include on derived types using string overload:
entities.Include("DerivedProperty")

or using lambda and specifying parameter type explicitly
entities.Include((Derived e) => e.DerivedProperty)

Made tweaks to the logic that computes navigation paths - now it looks for navigation on derived types also.
Added additional checks to the generated code (_Include method fixup)

old:

    fixup: (QueryContext queryContext | SomeEntity entity | Object[] included) =>
    {
        return !(bool ReferenceEquals(included[0], null))
        {
            return entity.Navigation = (OtherEntity)included[0]
        }
         : default(OtherEntity)
    }

new:

    fixup: (QueryContext queryContext | SomeEntity entity | Object[] included) =>
    {
        return !(bool ReferenceEquals(included[0], null)) && (entity is DerivedEntity) ?
        {
            return ((DerivedEntity)entity).DerivedNavigation = (OtherEntity)included[0]
        }
         : default(OtherEntity)
    }

This also allows to reuse include pipeline to project collection navigations on derived types (before it was producing N+1 queries)

For string based include, since there is no way to provide the type that the navigation is declared on, we aggressively include properties - if multiple derived types declare a property with the same name, they will all be included.

### DIFF
--- a/src/EFCore/Query/ResultOperators/Internal/IncludeResultOperator.cs
+++ b/src/EFCore/Query/ResultOperators/Internal/IncludeResultOperator.cs
@@ -26,7 +26,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ResultOperators.Internal
     public class IncludeResultOperator : SequenceTypePreservingResultOperatorBase, IQueryAnnotation
     {
         private List<string> _navigationPropertyPaths;
-        private INavigation[] _navigationPath;
+        private List<INavigation[]> _navigationPaths;
         private IQuerySource _querySource;
 
         /// <summary>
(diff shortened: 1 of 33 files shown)


src/EFCore/Query/ResultOperators/Internal/IncludeResultOperator.cs
```diff
@@ -136,23 +138,80 @@ public virtual INavigation[] GetNavigationPath([NotNull] QueryCompilationContext
                             NavigationPropertyPaths.FirstOrDefault()));
                 }
 
-                _navigationPath = new INavigation[NavigationPropertyPaths.Count];
+                var completedNavigationPaths = new List<List<INavigation>>();
```
maumar: algorithm for finding all the matching navigation paths on derived types (that match the given name pattern). We need to find at least one fully matching chain, otherwise we throw the exception. Provided that this fully matching chain is found, other chains that are only partially matching will be included also.

smitpatel:
Lets go over design meeting so we know what we are reviewing exactly :trollface:

smitpatel:
Closure?

maumar:
need to rework the translation part we talked about during the meeting

PR #9373, opened 2017-08-10
Revisit Functional Test Pattern Part 7: The Changed Monster

AndriySvyryd:
Refactor MonsterFixupTestBase to have the change notification strategy configuration in the fixture
Test only one change notification strategy per provider as the behavior is not provider-specific
Extract more methods to TestStore
Speed up ExecutionStrategyTest

Part of #9098


smitpatel:
:shipit:

PR #9392, opened 2017-08-12
Clean database from previously failed MigrationsSqlServerTest

AndriySvyryd:


AndriySvyryd:
One of the tests in MigrationsSqlServerTest. It must've hung before dropping the database.

Issue #9406, opened 2017-08-14
Test: Upgrade to 2.3.0 Beta 4


bricelam:
There are a few more analyzer issues we need to fix before upgrading.

PR #9407, opened 2017-08-14
Pin xUnit.net to 2.3.0 Beta 3 for now

bricelam:
New analyzer warnings were added in Beta 4 which fail with warnings as errors. Filed #9406 for us to do a proper upgrade.


smitpatel:
superseded by #9408

PR #9408, opened 2017-08-14
Pin xunit to 2.3.0-beta3-build3705

natemcmaster:
A new version of xunit hit nuget.org. The 2.3.0-* float picked it up and caused build failures.

http://aspnetci/viewLog.html?buildId=299594&tab=buildResultsDiv&buildTypeId=UniverseCoherence

Side note: it looks like moving to xunit 2.3.0-beta4 will require at least one source change.


smitpatel:
We should move to 2.3.0-beta4. It has fix for one of the biggest issue of EF in multi-level test hierarchy.

smitpatel:
See #8050

natemcmaster:
> We should move to 2.3.0-beta4. 

Sounds good. That work is tracked here https://github.com/aspnet/Coherence-Signed/issues/652, but feel free to do it yourself. I'm just trying to unblock the broken build.

PR #9450, opened 2017-08-17
Fix #9128: Projection with nested projection hangs/does not complete in 2.0 Preview 2

### LINKED ISSUE #9128: Projection with nested projection hangs/does not complete in 2.0 Preview 2
When we updated our code base to 2.0 Preview-2 we had significant problems with a number of our queries hanging. We initially flagged this as [AutoMapper/AutoMapper/#2184](https://github.com/AutoMapper/AutoMapper/issues/2184). But during discussions it has become apparent that this is a change at the EF Core 2 level.

The following style of query with an inner projection now fails to complete / hangs:

```
var result = await context.Documents.Select(d => new DocumentDTO
    {
        Reference = d.Reference,
        Items = d.Items.Select(i => new DocumentItemDTO
        {
            Reference = i.Reference
        }).ToList()
    }).FirstAsync();
```
It used to work in 2.0 Preview 1 (and 1.x.x) but now hangs in 2.0 Preview 2.

I appreciate that it is a problematic query, in the sense that the inner projection uses `ToList()` whilst the outer projection is `FirstAsync()` and I have some ideas for workarounds 

But overall I feel that this should continue to work, it feels natural, it is not unreasonable to request nested data and a projection in one step (we are trying to reduce the data retrieved from the db server) and I can't see how you could rewrite the inner projection as `async`. 

It seems not unreasonable to ask EF Core to resolve this type of query correctly and return the data requested asynchronously regardless of how inefficient / ugly it is.

I have created the following gists to repro this.

- [Test EF Core 2.0 Preview 1 Projection with nested projection](https://gist.github.com/bridgesquared/44b86ff80650bcf4d1a803dbc37404e9)
- [Test EF Core 2.0 Preview 2 Projection with nested projection](https://gist.github.com/bridgesquared/8648ac279383610b42f12e2b0e02fbd1)

Discussions on the AutoMapper repo have suggested that this is related to #8208 and possibly #9029.

Meanwhile, we've gone back to 2.0 Preview 1 and are trying to work out how we are going to address this in 2.0 Final. So any thoughts/fixes etc. would be much appreciated. Many thanks.

anpete:
Removes blocking by making async projections fully async even when nested queries are using sync operators.

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/ProjectionExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/ProjectionExpressionVisitor.cs
@@ -6,7 +6,12 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Extensions.Internal;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 using Remotion.Linq.Clauses.Expressions;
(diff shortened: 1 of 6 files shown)


src/EFCore/Query/ExpressionVisitors/ProjectionExpressionVisitor.cs
```diff
@@ -25,8 +30,100 @@ public class ProjectionExpressionVisitor : DefaultQueryExpressionVisitor
...
         }
 
+        private static bool TryFindAsyncMethod(string methodName, out MethodInfo asyncMethodInfo)
+        {
+            var candidateMethods
+                = typeof(AsyncEnumerable).GetTypeInfo().GetDeclaredMethods(methodName)
+                    .Where(mi =>
+                        mi.ReturnType.IsConstructedGenericType
+                        && mi.ReturnType.GetGenericTypeDefinition() == typeof(Task<>)
+                        && mi.GetParameters().Length > 0
```
AndriySvyryd: Should this be mi.GetParameters().Length == 2 ?
And also check that it only has one generic type argument

anpete:
@AndriySvyryd The real magic is in TaskLiftingExpressionVisitor, which we did for the new Include compiler.

src/EFCore/Query/ExpressionVisitors/ProjectionExpressionVisitor.cs
anpete: I should just pass the full list of generic args from the old method.

src/EFCore/Query/ExpressionVisitors/ProjectionExpressionVisitor.cs
```diff
@@ -25,8 +30,99 @@ public class ProjectionExpressionVisitor : DefaultQueryExpressionVisitor
...
+        /// <returns>
+        ///     The modified expression, if it or any subexpression was modified; otherwise, returns the original expression.
+        /// </returns>
+        protected override Expression VisitMethodCall(MethodCallExpression methodCallExpression)
+        {
+            var newExpression = base.VisitMethodCall(methodCallExpression);
+
+            switch (newExpression)
+            {
+                case MethodCallExpression newMethodCallExpression
```
smitpatel: Should have named this variable based on case block they represent

Issue #9404, opened 2017-08-14
Investigate tests failing without MARS


AndriySvyryd:
See SqlServerTestStore.cs

smitpatel:
Turns out we were leaking data readers in multiple spots causing MARS failure.

PR #9454, opened 2017-08-17
Query: Fix leaking data readers when query throws exception

smitpatel:
Leaks
- If Query fails while enumerating we leak data reader
- If DataReader disposing logger fails we leak data reader
  - Dispose reader before logging message
- If sync query throws due to concurrency check we leak data reader
  - Make enumerator creation more lazy by moving it inside concurrency detector check
  - If query is not run due to concurrency then no data readers will be created

Extend fix for #9074 to sync pipeline - sync queries do not create inner enumerator until needed

Fix issue #9404
QueryBugsTest need MARS=true because of running queries in parallel (using different contexts)
Introduce random MARS settings

### DIFF
--- a/src/EFCore.Relational/Storage/RelationalDataReader.cs
+++ b/src/EFCore.Relational/Storage/RelationalDataReader.cs
@@ -71,10 +71,30 @@ namespace Microsoft.EntityFrameworkCore.Storage
         protected RelationalDataReader([NotNull] DbDataReader reader)
         {
             // For testing
-
             Check.NotNull(reader, nameof(reader));
 
             _reader = reader;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        protected RelationalDataReader(
+            [NotNull] DbCommand command,
+            [NotNull] DbDataReader reader,
+            [NotNull] IDiagnosticsLogger<DbLoggerCategory.Database.Command> logger)
+        {
+            // For testing
+            Check.NotNull(command, nameof(command));
+            Check.NotNull(reader, nameof(reader));
+            Check.NotNull(logger, nameof(logger));
+
+            _command = command;
+            _reader = reader;
+            _logger = logger;
+            _startTime = DateTimeOffset.UtcNow;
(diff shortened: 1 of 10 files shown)


test/EFCore.SqlServer.FunctionalTests/Query/QueryBugsTest.cs
```diff
@@ -2343,7 +2343,7 @@ public class Blog9277
             Action<TContext> contextInitializer)
             where TContext : DbContext, IDisposable
         {
-            var testStore = SqlServerTestStore.CreateInitialized("QueryBugsTest");
+            var testStore = SqlServerTestStore.CreateInitialized("QueryBugsTest", multipleActiveResultSets: true);
```
smitpatel: This is necessary. Some tests are running queries in parallel in different context. We need to have MARS enabled for that.

src/EFCore.Relational/Storage/RelationalDataReader.cs
```diff
@@ -121,9 +125,6 @@ public virtual void Dispose()
                     _startTime,
                     _stopwatch.Elapsed);
 
-                _reader.Dispose();
-                _command.Parameters.Clear();
-                _command.Dispose();
                 _connection?.Close();
```
anpete: If logger fails we will not close connection?

smitpatel: yes :trollface:

src/EFCore/Query/Internal/LinqOperatorProvider.cs
```diff
@@ -288,7 +292,69 @@ public IEnumerator<TElement> GetEnumerator()
...
+        {
+            private readonly Func<T> _getElement;
+
+            /// <summary>
+            ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+            ///     directly from your code. This API may change or be removed in future releases.
+            /// </summary>
+            public ResultEnumerable([NotNull] Func<T> getElement)
+            {
+                Check.NotNull(getElement, nameof(getElement));
```
anpete: No check needed

src/EFCore.Relational/Storage/RelationalDataReader.cs
anpete: 😱 😱 😱

PR #9465, opened 2017-08-18
Update DbContextPool.cs

leonidmln:
Fixed bug in constructor to let developers use multi context


anpete:
Thanks, I've created a new PR for this with tests.

Issue #9395, opened 2017-08-12
Query: Apply CompositePredicateExpressionVisitor to join predicates


smitpatel:
At present it is applied to `SelectExpression.Predicate` only

maumar:
Also add more optimizations, e.g. (`(c.Discriminator == "Foo" ? c.FK : null) == o.PK` into `c.Discriminator == "Foo" && c.FK == o.PK`)

ajcvickers:
@maumar @smitpatel Please update description with more details and investigate for 2.1.

smitpatel:
Derived Include generates join condition using ternary to appropriately cast to derived type. 
Which generates query like this
```SQL
SELECT [l].[Name], [l].[Discriminator], [l].[LocustHordeId], [l].[ThreatLevel], [l].[DefeatedByNickname], [l].[DefeatedBySquadId], [t].[Nickname], [t].[SquadId], [t].[AssignedCityName], [t].[CityOrBirthName], [t].[Discriminator], [t].[FullName], [t].[HasSoulPatch], [t].[LeaderNickname], [t].[LeaderSquadId], [t].[Rank]
FROM [LocustLeaders] AS [l]
LEFT JOIN (
    SELECT [l.DefeatedBy].*
    FROM [Gears] AS [l.DefeatedBy]
    WHERE [l.DefeatedBy].[Discriminator] IN (N'Officer', N'Gear')
) AS [t] ON (CASE
    WHEN [l].[Discriminator] = N'LocustCommander'
    THEN [l].[DefeatedByNickname] ELSE NULL
END = [t].[Nickname]) AND (CASE
    WHEN [l].[Discriminator] = N'LocustCommander'
    THEN [l].[DefeatedBySquadId] ELSE NULL
END = [t].[SquadId])
WHERE [l].[Discriminator] IN (N'LocustCommander', N'LocustLeader')
```
Here using Case block on join condition means join cannot use any index and no longer sargable.

maumar:
fixed in 9355f64dbc5846271c36bbf6ff8196ed954751fe

PR #9472, opened 2017-08-18
Fix to #9395 - Query: Apply CompositePredicateExpressionVisitor to join predicates

maumar:
Some optimizations that are applied to predicates in SelectExpression can be safely applied to join predicates as well. Fix modifies CompositePredicateExpressionVisitor to also look into JoinExpression predicates and not only SelectExpression predicates.
Also changes the way we build join predicates themselves, when inheritance is concerned - now the discriminator predicate will be extracted out of the comparison, so that the comparisons are sargable.

Before:

((a is Derived) ? a.DerivedFK : null) == b.PK - which translated to CASE block (not sargable)

After:

a is Derived && a.DerivedFK == b.PK

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
@@ -31,7 +31,6 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         protected override Expression VisitExtension(Expression expression)
         {
             var selectExpression = expression as SelectExpression;
-
             if (selectExpression?.Predicate != null)
             {
                 var predicate = new EqualityPredicateInExpressionOptimizer().Visit(selectExpression.Predicate);
@@ -52,6 +51,10 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
                 }
 
                 selectExpression.Predicate = predicate;
+            }
+            else if (expression is PredicateJoinExpressionBase joinExpression)
+            {
+                joinExpression.Predicate = new EqualityPredicateInExpressionOptimizer().Visit(joinExpression.Predicate);
             }
 
             return base.VisitExtension(expression);
--- a/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
@@ -1104,6 +1104,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
                         ? new NullCompensatedExpression(newOperand)
                         : nullCompensatedExpression;
                 }
+                case DiscriminatorPredicateExpression discriminatorPredicateExpression:
+                    return new DiscriminatorPredicateExpression(base.VisitExtension(expression), discriminatorPredicateExpression.QuerySource);
                 default:
                     return base.VisitExtension(expression);
             }
--- a/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -25,6 +25,7 @@ using Remotion.Linq;
 using Remotion.Linq.Clauses;
 using Remotion.Linq.Clauses.Expressions;
 using Remotion.Linq.Clauses.ResultOperators;
+using Remotion.Linq.Parsing;
 
 namespace Microsoft.EntityFrameworkCore.Query
 {
@@ -1335,8 +1336,12 @@ namespace Microsoft.EntityFrameworkCore.Query
                     return base.VisitTypeBinary(typeBinaryExpression);
                 }
 
-                var entityType = _model.FindEntityType(typeBinaryExpression.TypeOperand);
+                if (!(typeBinaryExpression.Expression is QuerySourceReferenceExpression qsre))
+                {
+                    return base.VisitTypeBinary(typeBinaryExpression);
+                }
 
+                var entityType = _model.FindEntityType(typeBinaryExpression.TypeOperand);
                 if (entityType == null)
                 {
                     return base.VisitTypeBinary(typeBinaryExpression);
@@ -1352,7 +1357,7 @@ namespace Microsoft.EntityFrameworkCore.Query
                         = concreteEntityTypes[0].Relational().DiscriminatorProperty;
 
                     var discriminatorPropertyExpression
-                        = typeBinaryExpression.Expression.CreateEFPropertyExpression(discriminatorProperty);
+                        = qsre.CreateEFPropertyExpression(discriminatorProperty);
 
                     var discriminatorPredicate
                         = concreteEntityTypes
@@ -1363,7 +1368,7 @@ namespace Microsoft.EntityFrameworkCore.Query
                                         Expression.Constant(concreteEntityType.Relational().DiscriminatorValue, discriminatorPropertyExpression.Type)))
                             .Aggregate((current, next) => Expression.OrElse(next, current));
 
-                    return discriminatorPredicate;
+                    return new DiscriminatorPredicateExpression(discriminatorPredicate, qsre.TryGetReferencedQuerySource());
                 }
 
                 return Expression.Constant(true, typeof(bool));
@@ -1568,6 +1573,9 @@ namespace Microsoft.EntityFrameworkCore.Query
                 return false;
             }
 
+            var discriminatorPredicateOptimizer = new DiscriminatorPredicateOptimizingVisitor();
+            predicate = discriminatorPredicateOptimizer.Visit(predicate);
+
             QueriesBySource.Remove(joinClause);
 
             outerSelectExpression.RemoveRangeFromProjection(previousProjectionCount);
@@ -1700,6 +1708,9 @@ namespace Microsoft.EntityFrameworkCore.Query
                         Expression.Equal(joinClause.OuterKeySelector, joinClause.InnerKeySelector));
             }
 
+            var discriminatorPredicateOptimizer = new DiscriminatorPredicateOptimizingVisitor();
+            predicate = discriminatorPredicateOptimizer.Visit(predicate);
+
             QueriesBySource.Remove(joinClause);
 
             outerSelectExpression.RemoveRangeFromProjection(previousProjectionCount);
@@ -1776,6 +1787,28 @@ namespace Microsoft.EntityFrameworkCore.Query
                     groupJoinMethodCallExpression.Arguments[4]);
 
             return true;
+        }
+
+        private class DiscriminatorPredicateOptimizingVisitor : RelinqExpressionVisitor
+        {
+            protected override Expression VisitBinary(BinaryExpression binaryExpression)
+            {
+                if (binaryExpression.NodeType == ExpressionType.Equal
+                    && binaryExpression.Left.RemoveConvert() is ConditionalExpression conditionalExpression
+                    && conditionalExpression.Test is DiscriminatorPredicateExpression discriminatorPredicateExpression
+                    && conditionalExpression.IfFalse.IsNullConstantExpression())
+                {
+                    return Expression.AndAlso(
+                        discriminatorPredicateExpression.Reduce(),
+                        Expression.Equal(
+                            conditionalExpression.IfTrue.Type == binaryExpression.Right.Type
+                                ? conditionalExpression.IfTrue
+                                : Expression.Convert(conditionalExpression.IfTrue, binaryExpression.Right.Type),
+                            binaryExpression.Right));
+                }
+
+                return base.VisitBinary(binaryExpression);
+            }
         }
 
         private bool IsFlattenableGroupJoinDefaultIfEmpty(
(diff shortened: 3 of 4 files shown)


src/EFCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
```diff
@@ -53,6 +54,10 @@ protected override Expression VisitExtension(Expression expression)
 
                 selectExpression.Predicate = predicate;
             }
+            else if (joinExpression?.Predicate != null)
+            {
+                joinExpression.Predicate = new EqualityPredicateInExpressionOptimizer().Visit(joinExpression.Predicate);
```
maumar: PredicateReductionExpressionOptimizer and PredicateNegationExpressionOptimizer and are not safe to apply here - we are adding redundant terms to mark join on composite keys (this would be removed by predicate reduction)

maumar:
ping @anpete @smitpatel

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1618,94 @@ var newShapedQueryMethod
...
+        private Expression BuildJoinPredicate(Expression outerKeySelector, Expression innerKeySelector)
+        {
+            var discriminatorExtractor = new DiscriminatorExtractingExpressionVisitor(QueryCompilationContext.Model);
+            var newOuterKeySelector = discriminatorExtractor.Visit(outerKeySelector);
+            var newInnerKeySelector = discriminatorExtractor.Visit(innerKeySelector);
+            var result = Expression.Equal(newOuterKeySelector, newInnerKeySelector);
+
+            if (discriminatorExtractor.Discriminators.Any())
+            {
+                var discriminatorPredicateExpression = discriminatorExtractor.Discriminators.Values.Aggregate((c, n) => Expression.AndAlso(c, n));
```
smitpatel: Use Aggregate with seed to combine with result instead of doing it in block below this.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1618,94 @@ var newShapedQueryMethod
...
+                    Expression.AndAlso(
+                        discriminatorPredicateExpression,
+                        result),
+                    Expression.Constant(true));
+            }
+
+            return result;
+        }
+
+        private class DiscriminatorExtractingExpressionVisitor : ExpressionVisitorBase
```
smitpatel: Is it valid to derive from `ExpressionVisitorBase`? Do we want to extract discriminator inside of subquery models?

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1618,94 @@ var newShapedQueryMethod
...
+
+            public DiscriminatorExtractingExpressionVisitor(IModel model)
+            {
+                _model = model;
+            }
+
+            public Dictionary<IQuerySource, Expression> Discriminators => _discriminators;
+
+
+            protected override Expression VisitConditional(ConditionalExpression conditionalExpression)
```
smitpatel: extra line break above.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1618,94 @@ var newShapedQueryMethod
...
+            {
+                if (conditionalExpression.Test is BinaryExpression testBinary
+                    && conditionalExpression.IfFalse is ConstantExpression ifFalseConstant
```
smitpatel: Use `IsNullConstantExpression` extension method.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1618,94 @@ var newShapedQueryMethod
...
+                    && ifFalseConstant.Value == null
+                    && conditionalExpression.IfTrue is MethodCallExpression ifTrueMethodCall
+                    && ifTrueMethodCall.Method.IsEFPropertyMethod()
```
smitpatel: You can call `IsEfProperty` directly on methodcall expression. You can perhaps also do soft casting and null conditional method call comparing to true. instead of 2 blocks.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1618,94 @@ var newShapedQueryMethod
...
+
+                return base.VisitConditional(conditionalExpression);
+            }
+
+            private bool IsDicriminatorPredicate(BinaryExpression expression, out QuerySourceReferenceExpression qsre)
+            {
+                qsre = null;
+                if (expression.NodeType == ExpressionType.Equal
+                    && expression.Left is MethodCallExpression testBinaryMethodCall
+                    && testBinaryMethodCall.Method.IsEFPropertyMethod()
```
smitpatel: condense same as above

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1618,94 @@ var newShapedQueryMethod
...
+                        qsre = testBinaryMethodCallQsre;
+
+                        return true;
+                    }
+                }
+                else if (expression.NodeType == ExpressionType.OrElse
+                    && expression.Left is BinaryExpression binaryLeft
+                    && expression.Right is BinaryExpression binaryRight)
+                {
+                    var isLeftDiscriminatorPredicate = IsDicriminatorPredicate(binaryLeft, out var leftDiscriminatorQsre);
```
smitpatel: inline this

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1618,94 @@ var newShapedQueryMethod
...
+
+            private bool IsDicriminatorPredicate(BinaryExpression expression, out QuerySourceReferenceExpression qsre)
+            {
+                qsre = null;
+                if (expression.NodeType == ExpressionType.Equal
+                    && expression.Left is MethodCallExpression testBinaryMethodCall
+                    && testBinaryMethodCall.Method.IsEFPropertyMethod()
+                    && testBinaryMethodCall.Arguments[0] is QuerySourceReferenceExpression testBinaryMethodCallQsre)
+                {
+                    var declaringEntityType = _model.FindEntityType(testBinaryMethodCallQsre.Type);
```
smitpatel: This does not utilize pipeline to find qsre to entity type for queries which is used for owned entities. It would be correct result at present given owned entities does not allow inheritance atm. Not sure if we should just write safe code or react later. @anpete

src/EFCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
```diff
@@ -53,6 +54,10 @@ protected override Expression VisitExtension(Expression expression)
 
                 selectExpression.Predicate = predicate;
             }
+            else if (joinExpression?.Predicate != null)
```
smitpatel: Do we need to check for null here? Don't we have bigger issues if `PredicateJoinExpressionBase.Predicate` is null at this point?

maumar:
updated @smitpatel

src/EFCore.Relational/Query/ExpressionVisitors/Internal/CompositePredicateExpressionVisitor.cs
```diff
@@ -31,6 +31,7 @@ public CompositePredicateExpressionVisitor(bool useRelationalNulls)
         protected override Expression VisitExtension(Expression expression)
         {
             var selectExpression = expression as SelectExpression;
+            var joinExpression = expression as PredicateJoinExpressionBase;
```
smitpatel: Move it to `else if` and do `is` pattern matching.

maumar: done

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1619,87 @@ var newShapedQueryMethod
...
+                }
+
+                return base.VisitConditional(conditionalExpression);
+            }
+
+            private bool IsDicriminatorPredicate(BinaryExpression expression, out QuerySourceReferenceExpression qsre)
+            {
+                qsre = null;
+                if (expression.NodeType == ExpressionType.Equal
+                    && expression.Left is MethodCallExpression testBinaryMethodCall
```
smitpatel: this would not optimize if right side is EF.Property

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1618,6 +1619,87 @@ var newShapedQueryMethod
...
+
+            protected override Expression VisitConditional(ConditionalExpression conditionalExpression)
+            {
+                if (conditionalExpression.Test is BinaryExpression testBinary
+                    && conditionalExpression.IfFalse.IsNullConstantExpression() 
+                    && conditionalExpression.IfTrue is MethodCallExpression ifTrueMethodCall
+                    && ifTrueMethodCall.IsEFProperty()
+                    && IsDicriminatorPredicate(testBinary, out var testDiscriminatorQsre))
+                {
+                    Discriminators[testDiscriminatorQsre.ReferencedQuerySource] = testBinary;
```
smitpatel: This is overwriting for query source but that could cause unknown effects if multiple discriminator accesses are there.

smitpatel:
@maumar  - what happened to this?

maumar:
@smitpatel new version up

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -1104,6 +1104,12 @@ var equalityExpression
                         ? new NullCompensatedExpression(newOperand)
                         : nullCompensatedExpression;
                 }
+                case DiscriminatorPredicateExpression discriminatorPredicateExpression:
+                    {
```
smitpatel: avoid `{...}`

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1363,7 +1364,7 @@ var discriminatorPredicate
                                         Expression.Constant(concreteEntityType.Relational().DiscriminatorValue, discriminatorPropertyExpression.Type)))
                             .Aggregate((current, next) => Expression.OrElse(next, current));
 
-                    return discriminatorPredicate;
+                    return new DiscriminatorPredicateExpression(discriminatorPredicate, typeBinaryExpression.Expression.TryGetReferencedQuerySource());
```
smitpatel: This may be unsafe

test/EFCore.SqlServer.FunctionalTests/Query/GearsOfWarQuerySqlServerTest.cs
```diff
@@ -4047,13 +4011,7 @@ public override void Include_reference_on_derived_type_using_string()
...
     FROM [Gears] AS [l.DefeatedBy]
     WHERE [l.DefeatedBy].[Discriminator] IN (N'Officer', N'Gear')
-) AS [t] ON (CASE
-    WHEN [l].[Discriminator] = N'LocustCommander'
-    THEN [l].[DefeatedByNickname] ELSE NULL
-END = [t].[Nickname]) AND (CASE
-    WHEN [l].[Discriminator] = N'LocustCommander'
-    THEN [l].[DefeatedBySquadId] ELSE NULL
-END = [t].[SquadId])
+) AS [t] ON (([l].[Discriminator] = N'LocustCommander') AND ([l].[DefeatedByNickname] = [t].[Nickname])) AND (([l].[Discriminator] = N'LocustCommander') AND ([l].[DefeatedBySquadId] = [t].[SquadId]))
```
smitpatel: File a new issue for de-dupe.

maumar:
@smitpatel new version up

src/EFCore.Relational/Query/ExpressionVisitors/SqlTranslatingExpressionVisitor.cs
```diff
@@ -1104,6 +1104,8 @@ var equalityExpression
                         ? new NullCompensatedExpression(newOperand)
                         : nullCompensatedExpression;
                 }
+                case DiscriminatorPredicateExpression discriminatorPredicateExpression:
+                    return new DiscriminatorPredicateExpression(base.VisitExtension(expression), discriminatorPredicateExpression.QuerySource);
```
smitpatel: break long lines and leave and empty line before next case block.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1335,8 +1336,12 @@ protected override Expression VisitTypeBinary(TypeBinaryExpression typeBinaryExp
...
-                var entityType = _model.FindEntityType(typeBinaryExpression.TypeOperand);
+                if (!(typeBinaryExpression.Expression is QuerySourceReferenceExpression qsre))
+                {
+                    return base.VisitTypeBinary(typeBinaryExpression);
+                }
 
+                var entityType = _model.FindEntityType(typeBinaryExpression.TypeOperand);
                 if (entityType == null)
                 {
                     return base.VisitTypeBinary(typeBinaryExpression);
```
smitpatel: Collapse all the conditions which are branching to the base method into one if.

src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1568,6 +1573,9 @@ var predicate
                 return false;
             }
 
+            var discriminatorPredicateOptimizer = new DiscriminatorPredicateOptimizingVisitor();
```
smitpatel: Put this into CompositePredicateExpressionVisitor to make it DRY.

maumar:
addressed in 7ffc67c34a20b1ac07ade3de52497e751541ad37

PR #9476, opened 2017-08-18
Remove QueryableExtensions.Select method which was being used for int…

### LINKED ISSUE #9434: Query: Hide extension method on IQueryable
https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/EFCore/Extensions/Internal/QueryableExtensions.cs#L48

This select shows up in intellisense while using R# We should not have extension methods on IQueryable on public surface unless it is for public usage.

smitpatel:
…ernal purposes only

Resolves #9434

### DIFF
--- a/src/EFCore/Internal/EntityFinder.cs
+++ b/src/EFCore/Internal/EntityFinder.cs
@@ -5,11 +5,11 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
-using Microsoft.EntityFrameworkCore.Extensions.Internal;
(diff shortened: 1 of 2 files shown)


src/EFCore/Internal/EntityFinder.cs
```diff
@@ -285,8 +285,30 @@ private static IQueryable BuildQueryRoot(DbContext context, IEntityType entityTy
                 return (IQueryable)((IDbSetCache)context).GetOrAddSet(context.GetDependencies().SetSource, entityType.ClrType);
             }
 
-            return BuildQueryRoot(context, definingEntityType)
-                .Select(entityType.DefiningNavigationName, definingEntityType.ClrType, entityType.ClrType);
+            return BuildQueryRootWithDefiningNavigation(context, definingEntityType, entityType);
+        }
+
+        private static IQueryable BuildQueryRootWithDefiningNavigation(DbContext context, IEntityType definingEntityType, IEntityType entityType)
```
AndriySvyryd: This method can be called just `BuildQueryRoot`

Issue #9437, opened 2017-08-17
Diagnostics logging is incompatible with windows EventLog


MaximBalaganskiy:
After an upgrade to EF Core 2.0 EventLog provider throws an exception
```
Exception message: System.AggregateException: An error occurred while writing to logger(s). ---> System.ArgumentException: Invalid eventID value '100403'. It must be in the range between '0' and '65535'.
Stack trace: at System.Diagnostics.EventLogInternal.WriteEntry(String message, EventLogEntryType type, Int32 eventID, Int16 category, Byte[] rawData)
   at System.Diagnostics.EventLog.WriteEntry(String message, EventLogEntryType type, Int32 eventID, Int16 category)
   at Microsoft.Extensions.Logging.EventLog.WindowsEventLog.WriteEntry(String message, EventLogEntryType type, Int32 eventID, Int16 category)
   at Microsoft.Extensions.Logging.EventLog.EventLogLogger.WriteMessage(String message, EventLogEntryType eventLogEntryType, Int32 eventId)
   at Microsoft.Extensions.Logging.EventLog.EventLogLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.Logger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   --- End of inner exception stack trace ---
   at Microsoft.Extensions.Logging.Logger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.LoggerMessage.<>c__DisplayClass8_0`4.<Define>b__0(ILogger logger, T1 arg1, T2 arg2, T3 arg3, T4 arg4, Exception exception)
   at Microsoft.EntityFrameworkCore.Diagnostics.EventDefinition`4.Log[TLoggerCategory](IDiagnosticsLogger`1 logger, TParam1 arg1, TParam2 arg2, TParam3 arg3, TParam4 arg4, Exception exception)
   at Microsoft.EntityFrameworkCore.Internal.CoreLoggerExtensions.ContextInitialized(IDiagnosticsLogger`1 diagnostics, DbContext context, DbContextOptions contextOptions)
   at Microsoft.EntityFrameworkCore.DbContext.get_InternalServiceProvider()
[...]

PR #9479, opened 2017-08-19
Patch: Make Event ID values smaller so they will work with Windows EventLog

ajcvickers:
Issue #9437

All values now fit in a 16-bit unsigned int. There is a Quirks mode to switch back to the 2.0.0 RTM values.

### DIFF
--- a/src/EFCore.Specification.Tests/TestHelpers.cs
+++ b/src/EFCore.Specification.Tests/TestHelpers.cs
@@ -79,8 +79,16 @@ namespace Microsoft.EntityFrameworkCore
         public void TestEventLogging(
             Type eventIdType,
             Type loggerExtensionsType,
-            IDictionary<Type, Func<object>> fakeFactories)
+            IDictionary<Type, Func<object>> fakeFactories,
+            bool useLegacyIds = false)
         {
+            if (useLegacyIds)
+            {
+                // Note that we can't run the test in both modes at once because EventIds are cached.
+                // If making a change to the quirk behavior, change the default of useLegacyIds to true and re-run all tests.
+                AppContext.SetSwitch("Microsoft.EntityFrameworkCore.Diagnostics.UseLegacyEventIds", true);
+            }
+
             var eventIdFields = eventIdType.GetTypeInfo()
                 .DeclaredFields
                 .Where(p => p.FieldType == typeof(EventId))
@@ -112,6 +120,15 @@ namespace Microsoft.EntityFrameworkCore
                 }
 
                 var eventId = (EventId)eventIdField.GetValue(null);
+
+                if (useLegacyIds)
+                {
+                    Assert.InRange(eventId.Id, CoreEventId.CoreBaseId * 10, int.MaxValue);
+                }
+                else
+                {
+                    Assert.InRange(eventId.Id, CoreEventId.CoreBaseId, ushort.MaxValue);
+                }
 
                 var categoryName = Activator.CreateInstance(category).ToString();
                 Assert.Equal(categoryName + "." + eventName, eventId.Name);
--- a/src/EFCore/Diagnostics/CoreEventId.cs
+++ b/src/EFCore/Diagnostics/CoreEventId.cs
@@ -1,7 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Diagnostics;
+using Microsoft.EntityFrameworkCore.Diagnostics.Internal;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.EntityFrameworkCore.Diagnostics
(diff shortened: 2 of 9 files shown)


ajcvickers:
FYI: @NickCraver @fitzchak @roji @ErikEJ @caleblloyd See full discussion on #9437

NickCraver:
Got it. 

For what it's worth, since these weren't exposed in a way where I could rely on the constants anyway, I don't think this change will actually matter (for my use). The tradeoff is I don't have a neat `switch` for the `IObserver` or anything remotely that clean, but the upside is this doesn't actually break the code since they're compared to the spun-up runtime values.

Ping me when there's a alpha/beta/preview/whatever package to sanity check? I'm happy to do so.

ajcvickers:
Also tested manually with real Windows EventLog.

src/EFCore/Diagnostics/CoreEventId.cs
```diff
@@ -19,24 +21,36 @@ namespace Microsoft.EntityFrameworkCore.Diagnostics
     public static class CoreEventId
     {
         /// <summary>
+        ///     <para>
+        ///         The name of the switch to use with <see cref="AppContext.SetSwitch"/> to use the event ID numbers
+        ///         that were shipped with EF Core 2.0.0. These IDs are two big for use in Windows EventLog, so smaller
```
Eilon: `two` --> `too`

src/EFCore/Diagnostics/CoreEventId.cs
```diff
@@ -19,24 +21,36 @@ namespace Microsoft.EntityFrameworkCore.Diagnostics
...
+        ///     <para>
+        ///         The name of the switch to use with <see cref="AppContext.SetSwitch"/> to use the event ID numbers
+        ///         that were shipped with EF Core 2.0.0. These IDs are two big for use in Windows EventLog, so smaller 
+        ///         numbers were used starting with EF Core 2.0.1.
+        ///     </para>
+        ///     <para>
+        ///         Only set this switch if a tool was compiled against 2.0.0 and is explicitly using the legacy IDs.
+        ///     </para>
+        /// </summary>
+        public const string UseLegacyEventIdsSwitch = "Microsoft.EntityFrameworkCore.Diagnostics.UseLegacyEventIds";
```
Eilon: I don't think we should add *any* public surface area in a patch, unless it's absolutely required. Can we make this `const` be `internal` and then make it public in the next release?

The risk is that a future patch could risk a breaking change because this is an API change in a patch version (for complex reasons I can share internally).

src/EFCore/Diagnostics/CoreEventId.cs
```diff
@@ -19,24 +21,36 @@ namespace Microsoft.EntityFrameworkCore.Diagnostics
...
+
+        /// <summary>
         ///     The lower-bound for event IDs used by any Entity Framework or provider code.
         /// </summary>
-        public const int CoreBaseId = 100000;
+        public const int CoreBaseId = 10000;
```
Eilon: These probably would have been better as `public static readonly`, no? (Too late now, of course.)

src/EFCore/Diagnostics/Internal/EventIdFactory.cs
```diff
@@ -0,0 +1,48 @@
...
+using JetBrains.Annotations;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.EntityFrameworkCore.Diagnostics.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class EventIdFactory
```
Eilon: Actually I think this needs to be `internal` as well...

src/EFCore.Specification.Tests/TestHelpers.cs
```diff
@@ -79,8 +79,16 @@ private class FakeCurrentDbContext : ICurrentDbContext
         public void TestEventLogging(
             Type eventIdType,
             Type loggerExtensionsType,
-            IDictionary<Type, Func<object>> fakeFactories)
+            IDictionary<Type, Func<object>> fakeFactories,
+            bool useLegacyIds = false)
```
Eilon: And even though this is just a test, it's a "public test", so should this be done in a non-breaking way via an overload?

ajcvickers: We haven't ever tracked breaking changes in tests; we just expect consumers to update in the normal way. Would be good to discuss the rationale for this.

ErikEJ: Never been an issue for me (and I consume the tests big time)

Eilon: Yeah it was just a question, so if people who "consume the tests big time" are ok with it ( 😄 ), then I'm ok with it.

With patches my default mode is "better safe than sorry," hence the question.

fitzchak:
@ajcvickers would be best if the version of EF could be exposed in the event ID. As the events values are might change between versions so based on the event ID we'll know what object to expect.

ajcvickers:
@fitzchak Events are intended to be non-breaking across versions, starting with 2.0.

ErikEJ:
@ajcvickers Is any codes changes required in the provider for this? https://github.com/ErikEJ/EntityFramework.SqlServerCompact/blob/master/src/Provider40/Diagnostics/SqlCeEventId.cs

ajcvickers:
@ErikEJ Providers will need to recompile to get the "fixed" Ids.

ErikEJ:
@ajcvickers So I should release a "2.0.1" based provider?

ajcvickers:
@ErikEJ Unfortunately, yes.

ErikEJ:
Well, that solves this issue then: https://github.com/aspnet/EntityFrameworkCore/issues/9831 😄

PR #9494, opened 2017-08-20
Complete API documentation for remaining product assemblies

ajcvickers:
Design and provider assemblies--most code is internal and most docs are boilerplate.

Issue #1148

### DIFF
--- a/src/EFCore.SqlServer/Metadata/SqlServerEntityTypeAnnotations.cs
+++ b/src/EFCore.SqlServer/Metadata/SqlServerEntityTypeAnnotations.cs
@@ -6,24 +6,47 @@ using Microsoft.EntityFrameworkCore.Metadata.Internal;
 
 namespace Microsoft.EntityFrameworkCore.Metadata
 {
+    /// <summary>
+    ///     Properties for relational-specific annotations accessed through
+    ///     <see cref="SqlServerMetadataExtensions.SqlServer(IMutableEntityType)" />.
+    /// </summary>
     public class SqlServerEntityTypeAnnotations : RelationalEntityTypeAnnotations, ISqlServerEntityTypeAnnotations
     {
+        /// <summary>
+        ///     Constructs an instance for annotations of the given <see cref="IEntityType" />.
+        /// </summary>
+        /// <param name="entityType"> The <see cref="IEntityType" /> to use. </param>
         public SqlServerEntityTypeAnnotations([NotNull] IEntityType entityType)
             : base(entityType)
         {
         }
 
+        /// <summary>
+        ///     Constructs an instance for annotations of the <see cref="IEntityType" />
+        ///     represented by the given annotation helper.
+        /// </summary>
+        /// <param name="annotations">
+        ///     The <see cref="RelationalAnnotations" /> helper representing the <see cref="IEntityType" /> to annotate.
+        /// </param>
         public SqlServerEntityTypeAnnotations([NotNull] RelationalAnnotations annotations)
             : base(annotations)
         {
         }
 
+        /// <summary>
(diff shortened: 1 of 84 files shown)


src/EFCore.SqlServer/Metadata/SqlServerEntityTypeAnnotations.cs
```diff
@@ -6,24 +6,47 @@
...
+        ///     Indicates whether or not the type is mapped to a memory-optimized table.
+        /// </summary>
         public virtual bool IsMemoryOptimized
         {
             get => Annotations.Metadata[SqlServerAnnotationNames.MemoryOptimized] as bool? ?? false;
             set => SetIsMemoryOptimized(value);
         }
 
+        /// <summary>
+        ///     Attempts to set memory optimization using the semantics of the <see cref="RelationalAnnotations" /> in use.
```
AndriySvyryd: Remove the `RelationalAnnotations` bit. It doesn't mean much without looking at the code.

src/EFCore.SqlServer/Metadata/SqlServerPropertyAnnotations.cs
```diff
@@ -62,12 +100,28 @@ public virtual ISequence FindHiLoSequence()
             return modelExtensions.FindSequence(sequenceName, sequenceSchema);
         }
 
+        /// <summary>
+        ///     <para>
+        ///         Gets or sets the <see cref="SqlServerValueGenerationStrategy" /> to use for the property.
+        ///     </para>
+        ///     <para>
+        ///         If no strategy is set for the property, then the strategy to use will be taken from the <see cref="IModel" />
```
AndriySvyryd: Remarks?

src/EFCore.SqlServer/Metadata/SqlServerPropertyAnnotations.cs
```diff
@@ -193,6 +258,15 @@ protected virtual bool CanSetValueGenerationStrategy(SqlServerValueGenerationStr
             return true;
         }
 
+        /// <summary>
+        ///     Gets the default value set for the property.
+        /// </summary>
+        /// <param name="fallback">
```
AndriySvyryd: Rename `fallback` to something like `avoidConflicts` and make all calls use it explicitly.

src/EFCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -976,11 +1241,24 @@ protected override void Generate(SqlOperation operation, IModel model, Migration
...
             [NotNull] string name,
             [NotNull] string newName,
             [NotNull] MigrationCommandListBuilder builder) => Rename(name, newName, /*type:*/ null, builder);
 
+        /// <summary>
+        ///     Generates a rename.
+        /// </summary>
+        /// <param name="name"> The old name. </param>
+        /// <param name="newName"> The new name. </param>
+        /// <param name="type"> If not <c>null</c>, then appends literal for type. </param>
```
bricelam: The type of object being renamed. (e.g. COLUMN or INDEX)

src/EFCore.SqlServer/Metadata/SqlServerPropertyAnnotations.cs
ajcvickers: We generally perfer paras over remarks because the former show up in Intellisense.

src/EFCore.SqlServer/Metadata/SqlServerPropertyAnnotations.cs
ajcvickers: Breaking change...

PR #9514, opened 2017-08-22
Fix test to exercise literal DateTime generation

roji:
Can_query_using_any_nullable_data_type_as_literal renders built in data
types as literals, except for DateTime which for some reason sent it
to the database as a parameter.


ajcvickers:
@roji DefaultDateTime was intended to be `new DateTime()` on platforms where this can be mapped (because there are issues mapping this sometimes--such as if the default mapping for SQL Server was for datetime instead of datetime2) but irrecoverable for platforms that can't map this value--for which SQL Compact is the one that comes to mind. @ErikEJ?

So really it would be better to have a test that preserves the value of default DateTime, but which generates literals, and can be overridden by providers that don't support this.

Issue #9038, opened 2017-07-01
2.0.0 including collection navigation after optional navigation throws NRE with async


caleblloyd:
We are trying to upgrade our MySQL provider to 2.0.0. Our Discriminator Test that tests TPH Inheritance is resulting in this NullReferenceException

[Here is the Discriminator Test](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/EFCore.MySql.FunctionalTests/Tests/Models/DiscriminatorTest.cs) - read it for a good laugh, I'm put some time into making sure the Hierarchies were fun :smile:

Here is the Query that is causing the Exception:

```
				var teachers = await db.People.OfType<PersonTeacher>()
					.Where(m => m.Id >= _fixture.LowestTeacherId && m.Id <= _fixture.HighestTeacherId)
					.OrderBy(m => m.Id)
					.Include(m => m.Family)
					.ThenInclude(m => m.Members)
					.Include(m => m.Students)
					.ThenInclude(m => m.Family)
					.ThenInclude(m => m.Members)
					.ToListAsync();
```

Here is the Exception:

```
Failed   Pomelo.EntityFrameworkCore.MySql.PerfTests.Tests.Models.DiscriminatorTest.TestDiscriminator
Error Message:
 System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
   at Microsoft.EntityFrameworkCore.Query.Internal.IncludeCompiler.IncludeLoadTreeNodeBase.<_AwaitMany>d__8.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.EntityFrameworkCore.Query.Internal.IncludeCompiler.<_IncludeAsync>d__18`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal.TaskLiftingExpressionVisitor.<_ExecuteAsync>d__8`1.MoveNext()
[...]

PR #9525, opened 2017-08-22
[2.0.1/TINY] Fix to #9038 - NullReferenceException when using TPH Inheritance

maumar:
Problem was that for the fixup for _IncludeAsync method would sometimes return default(Task) (i.e. null) if there was nothing to include. This breaks the async pattern, which is expecting a Task. Fix is to return Task.CompletedTask instead.

### DIFF
--- a/src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
+++ b/src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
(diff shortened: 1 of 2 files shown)


src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
```diff
@@ -302,7 +307,11 @@ var stateManagerProperty
                         Expression.Block(
                             blockType,
                             blockExpressions),
-                        Expression.Default(blockType),
+                        blockType == typeof(Task)
+                            ? Expression.Property(
+                                null,
+                                _taskCompletedTaskPropertyInfo)
```
AndriySvyryd: Just return Expression.Constant(Task.CompletedTask)

smitpatel:
❗️ This may have introduced deadlock issues. AppVeyor started hanging after this commit.

anpete:
Why doesn't the test use await in the normal way?

PR #9555, opened 2017-08-24
Ensure .NET Core 1.1.2 is installed for the benchmarks project

natemcmaster:
Pretty soon, we will stop installing .NET Core 1.1 and 1.0 by default in the dev branch of KoreBuild. EFCore1 benchmarks still uses netcoreapp1.1. This specifies the required dependency on .NET Core 1.1.2 so it can be installed by KoreBuild.

Usage: to ensure all runtimes are installed, execute `build.sh /t:InstallDotnet`.

cref https://github.com/aspnet/BuildTools/issues/347

### DIFF
--- a/build/repo.props
+++ b/build/repo.props
@@ -5,8 +5,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- Used in the EFCore1 benchmarks project -->
-    <DotNetCoreRuntime Include="1.1.2" />
+    <!-- Only required when preparing to run the EFCore1 benchmarks project -->
+    <DotNetCoreRuntime Include="1.1.2" Condition=" '$(IsBuildingBenchmarks)' == 'true' " />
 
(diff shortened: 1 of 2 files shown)


build/repo.props
```diff
@@ -5,6 +5,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- Used in the EFCore1 benchmarks project -->
+    <DotNetCoreRuntime Include="1.1.2" />
```
natemcmaster: FYI - this will also be installed by default when you run `build.cmd` or `build.cmd /t:Restore`. Since it's not needed for a regular run, we could introduce a new condition, something like this:

```xml
<DotNetCoreRuntime Include="1.1.2"  Condition="'$(IsBuildingBenchmarks)' == 'true"/>
```

Thoughts @smitpatel ?

smitpatel: Conditional sounds good. I will set it as environment variable in Teamcity configuration.

PR #9561, opened 2017-08-24
Add support for System.Transactions

### LINKED ISSUE #5595: Enable support for System.Transactions
(Forked from https://github.com/aspnet/EntityFramework/issues/4147 which was mainly about fixing the misleading exception message we throw to indicate that we don't support `System.Transactions`)

By design we don't support `System.Transactions` (a.k.a. ambient transactions / `TransactionScope` and or `CommitableTransaction`) functionality in EF Core 1.0.0. 

In full .NET framework connection objects such as `SqlConnection` will still auto-enlist on `Open()` or `OpenAsync()` but we have code in EF Core to detect the presence of an ambient transaction an throw. 

The new `ConfigureWarnings()` API allows turning the error into a warning or to ignore it completely, but since EF Core will always just start a local transaction in the connection anyway (i.e. with `BeginTransaction()`) the feature is actually not usable. 

This issue is about the possibility of getting out of the way from `System.Transactions` by conditionally skipping the creation of the local database transaction in full .NET or in any other platform that supports it e.g. when/if https://github.com/dotnet/corefx/issues/2949 is implemented.

AndriySvyryd:
The tests currently only run for SQL Server on net461

Fixes #5595

### DIFF
--- a/src/EFCore.Relational.Specification.Tests/TransactionTestBase.cs
+++ b/src/EFCore.Relational.Specification.Tests/TransactionTestBase.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Data.Common;
 using System.Linq;
 using System.Threading.Tasks;
+using System.Transactions;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Storage;
@@ -22,6 +24,11 @@ namespace Microsoft.EntityFrameworkCore
         {
             Fixture = fixture;
             Fixture.Reseed();
+
+            if (TestStore.ConnectionState == ConnectionState.Closed)
+            {
+                TestStore.OpenConnection();
+            }
         }
 
         protected TFixture Fixture { get; set; }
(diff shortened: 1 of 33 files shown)


src/EFCore.Relational.Specification.Tests/TransactionTestBase.cs
```diff
@@ -540,17 +665,13 @@ public virtual void UseTransaction_throws_if_mismatched_connection(bool autoTran
...
-        [InlineData(false)]
-        public virtual void UseTransaction_throws_if_another_transaction_started(bool autoTransaction)
+        [Fact]
+        public virtual void UseTransaction_throws_if_another_transaction_started()
         {
             using (var transaction = TestStore.BeginTransaction())
             {
                 using (var context = CreateContextWithConnectionString())
                 {
-                    context.Database.AutoTransactionsEnabled = autoTransaction;
```
ajcvickers: Why did this change?

AndriySvyryd: Redundant test. AutoTransactionsEnabled only affects SaveChanges. This test doesn't call SaveChanges.

src/EFCore.Relational/Storage/RelationalConnection.cs
```diff
@@ -104,6 +104,53 @@ protected RelationalConnection([NotNull] RelationalConnectionDependencies depend
         public virtual IDbContextTransaction CurrentTransaction { get; [param: CanBeNull] protected set; }
 
         /// <summary>
+        ///     The currently enlisted transaction.
```
ajcvickers: I think we need more here to document what flavor of .NET transaction this "Transaction" type is. @divega?

src/EFCore.Relational/Storage/RelationalConnection.cs
```diff
@@ -355,60 +375,199 @@ public virtual bool Open(bool errorsExpected = false)
...
+                    // will not create a new transaction if an ambient transaction already exists - the ambient transaction will be used and we will
+                    // not end up in this code path). If we get here we are already in a loss-loss situation - we cannot enlist to the second transaction
+                    // as this would cause an exception saying that there is already an active transaction that needs to be committed or rolled back
+                    // before we can enlist the connection to a new transaction. The other option (and this is what we do here) is to close and reopen
+                    // the connection. This will enlist the newly opened connection to the second transaction but will also close the reader being used
+                    // by the first active query. As a result when trying to continue reading results from the first query the user will get an exception
+                    // saying that calling "Read" on a closed data reader is not a valid operation.
+                    AmbientTransaction = current;
```
ajcvickers: I may be missing something here, but is this better than just throwing? In other words, what's the valid scenario where we need to do this?

AndriySvyryd: This this EF6 behavior. If the query reader enlisted in the first transaction is never read from after this the second operation will still work. We can choose to throw instead.

ajcvickers: Not sure, would like to discuss with @divega

src/EFCore.Relational/Storage/RelationalConnection.cs
divega: System.Transactions.Transaction?

ajcvickers: @divega I was thinking of some more words than this to help people decide which to use...

divega: Ok. I misunderstood. Stepping back I think these API additions can be very confusing. Not sure additional words in the member documentation are going to help. 

I am thinking we should brainstorm something we can use in the API names as well to differentiate from the normal scenarios we already support. Right now "external transaction" comes to mind.

divega: Another approach suggested by @anpete would be to make these extension methods in the System.Transactions namespace, so people have to add a using statement to light the API up.

AndriySvyryd:
@divega, @ajcvickers I've updated with extension methods and connection keep-alive. Can you review again?

I found a scenario broken by the keep-alive:
```C#
using (new TransactionScope())
{
    // query

    using (new TransactionScope(TransactionScopeOption.Suppress))
    {
        // query
    }

    // SaveChanges
}
```

Now the middle query would still be executed in the transaction. If we tried to unenlist it would throw.

divega:
>I found a scenario broken by the keep-alive:...

Is this a scenario in which you would really need more than one connection? Then it is kind of nice if it throws...

ajcvickers:
Signing off on the latest changes.

divega:
Upgrading to providers-beware since there is new API here (`SupportsAmbientTransactions`) providers would need to implement.

bruno-garcia:
No milestone set so I wonder: When could we expect this on NuGet?

AndriySvyryd:
@bruno-garcia This is just a pull request, the corresponding issue is https://github.com/aspnet/EntityFrameworkCore/issues/5595 and it's in the 2.1.0 milestone

bruno-garcia:
My bad. I got to the issue after asking here.

codematrix:
So, I guess this is still not supported? I'm currently using 2.02 and still getting the same issue.

AndriySvyryd:
@codematrix This feature is in 2.1.0-preview1

roji:
FYI, am adding support for this in Npgsql and wanted to share some info.

Some of the tests purposefully generated an error and then continue with the TransactionScope (`SaveChanges_uses_enlisted_transaction`, `SaveChanges_uses_ambient_transaction_with_connectionString`, `SaveChanges_uses_ambient_transaction` when `closeConnection` is true). PostgreSQL However, in PostgreSQL, a transaction is put into a failed state after any error, and is unrecoverable - it can only be rolled back. So these tests must be skipped for now (which isn't great because they test some important behavior).

Ideally these tests wouldn't depend on continuing after a failure. But it's no big deal - mostly wanted you to know that you cannot rely on a transaction being recoverable after an error across databases.

AndriySvyryd:
Filed https://github.com/aspnet/EntityFrameworkCore/issues/12001

PR #9564, opened 2017-08-24
Adding opt-out mechanism for fix to #9038 - NullReferenceException when using TPH Inheritance

maumar:


src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
```diff
@@ -303,7 +303,8 @@ var stateManagerProperty
                         Expression.Block(
                             blockType,
                             blockExpressions),
-                        blockType == typeof(Task)
+                        blockType == typeof(Task) 
+                            && !(AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Query.UseLegacyAsyncInclude", out var isEnabled) && isEnabled)
```
ajcvickers: I think this says "If the switch is not present and the switch is enabled" then perform the new behavior.

I think it should be "If the switch is not present OR the switch is enabled".

Did you test with and without the switch set, at least manually? I might be getting it wrong...

ajcvickers:
@smitpatel Feel free to check in the test fix.

src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
maumar: The new behavior is in the iftrue part, old behavior is for iffalse. We do the new behavior if NOT (switch is present AND set to true)), i.e. when switch is not present OR switch is not set to true. Will push the negation into individual components, to make it less confusing. I did test (manually) for no switch and with switch set to true, will double check that it behaves correctly when switch is set to false

PR #9584, opened 2017-08-25
Query: GroupBy after Skip/Take should push down subquery

### LINKED ISSUE #9581: Query: GroupBy after Skip/Take produces incorrect results
```C#
        [ConditionalFact]
        public virtual void OrderBy_Skip_GroupBy()
        {
            AssertQuery<Order>(
                os => os.OrderBy(o => o.OrderDate).Skip(50).GroupBy(o => o.CustomerID).OrderBy(g => g.Key),
                assertOrder: true);
        }
```
Produces 
```SQL
  SELECT [o].[OrderID], [o].[CustomerID], [o].[EmployeeID], [o].[OrderDate]
  FROM [Orders] AS [o]
  ORDER BY [o].[CustomerID], [o].[OrderDate]
  OFFSET @__p_0 ROWS
```
Which is incorrect SQL and gives wrong results.

smitpatel:
Also lift GroupBy subquery when possible
This allows us to access the SelectExpression corresponding to GroupBy Result operator in ParentQMV

Resolves #9581

### DIFF
--- a/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -785,6 +785,15 @@ namespace Microsoft.EntityFrameworkCore.Query
 
             subQueryModelVisitor.VisitSubQueryModel(subQueryModel);
 
+            if ((subQueryModelVisitor.Expression as MethodCallExpression)?.Method.MethodIsClosedFormOf(QueryCompilationContext.QueryMethodProvider.GroupByMethod) == true
+                && !(querySource is AdditionalFromClause))
+            {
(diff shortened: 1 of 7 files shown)


src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -785,6 +785,16 @@ var subQueryModelVisitor
 
             subQueryModelVisitor.VisitSubQueryModel(subQueryModel);
 
+            if ((subQueryModelVisitor.Expression as MethodCallExpression)?.Method.GetGenericMethodDefinition()
+                .Equals(QueryCompilationContext.QueryMethodProvider.GroupByMethod) == true
```
anpete: Use MethodIsClosedFormOf

PR #9587, opened 2017-08-25
Improving fix for #9038 - NullReferenceException when using TPH Inheritance

maumar:
Instead of modifying the Task that we return from the problematic query, we are checking if the resulting fixup Task is null and only await on it if it's not.

### DIFF
--- a/src/EFCore/Query/Internal/IncludeCompiler.cs
+++ b/src/EFCore/Query/Internal/IncludeCompiler.cs
@@ -220,7 +220,12 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
         {
             if (entity != null)
             {
-                await fixup(queryContext, entity, included, cancellationToken);
+                var fixupTask = fixup(queryContext, entity, included, cancellationToken);
+                if (AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Query.UseLegacyAsyncInclude", out var isEnabled) && isEnabled
+                    || fixupTask != null)
+                {
+                    await fixupTask;
+                }
             }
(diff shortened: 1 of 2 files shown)


src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -220,7 +220,18 @@ public static bool IsIncludeMethod([NotNull] MethodCallExpression methodCallExpr
         {
             if (entity != null)
             {
-                await fixup(queryContext, entity, included, cancellationToken);
+                if (AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Query.UseLegacyAsyncInclude", out var isEnabled) && isEnabled)
+                {
```
anpete: @ajcvickers @divega  Any value in these quirk names containing the bug number?

anpete: We could create an extension method: IsQuirkEnabled, to encapsulate this small piece of logic.

src/EFCore/Query/Internal/IncludeCompiler.cs
```diff
@@ -220,7 +220,18 @@ public static bool IsIncludeMethod([NotNull] MethodCallExpression methodCallExpr
...
+                    await fixup(queryContext, entity, included, cancellationToken);
+                }
+                else
+                {
+                    var fixupTask = fixup(queryContext, entity, included, cancellationToken);
```
smitpatel: store fixupTask outside fhe condition and add quirk check alongwith null condition for await

PR #9588, opened 2017-08-25
Fixes version dependent tests for 1.0.5

jkotalik:
For the 1.0.5 patch.

### DIFF
--- a/src/Microsoft.EntityFrameworkCore.SqlServer.Design/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.EntityFrameworkCore.SqlServer.Design/Properties/AssemblyInfo.cs
@@ -11,5 +11,5 @@ using System.Resources;
 [assembly: AssemblyProduct("Microsoft EntityFramework Core")]
 
 #if CSPROJ
-[assembly: AssemblyVersion("1.0.4.0")]
+[assembly: AssemblyVersion("1.0.5.0")]
 #endif
--- a/test/Microsoft.EntityFrameworkCore.Tests/ModelSourceTest.cs
+++ b/test/Microsoft.EntityFrameworkCore.Tests/ModelSourceTest.cs
@@ -79,7 +79,7 @@ namespace Microsoft.EntityFrameworkCore.Tests
             var model = modelSource.GetModel(new Context1(), null,
                 new LoggingModelValidator(new Logger<LoggingModelValidator>(new LoggerFactory())));
 
-            Assert.StartsWith("1.0.4", model.GetProductVersion(), StringComparison.OrdinalIgnoreCase);
+            Assert.StartsWith("1.0.5", model.GetProductVersion(), StringComparison.OrdinalIgnoreCase);
         }
 
         private class Context1 : DbContext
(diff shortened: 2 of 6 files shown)


src/Microsoft.EntityFrameworkCore.SqlServer.Design/Properties/AssemblyInfo.cs
```diff
@@ -11,5 +11,5 @@
 [assembly: AssemblyProduct("Microsoft EntityFramework Core")]
 
 #if CSPROJ
-[assembly: AssemblyVersion("1.0.4.0")]
```
ajcvickers: There are no 1.0.x changes in EF and we don't depend on any changes in other assemblies to make EF work, so we shouldn't be bumping product assembly versions. It may be that we need to bump up test assembly dependencies; but pretty sure this does not do that.
/cc @muratg

JunTaoLuo: There were dependency updates, see https://github.com/aspnet/EntityFrameworkCore/commit/95f0663d47352805d823cc94d708d3f2e38c2aac

jkotalik: These changes were made in both of the previous 1.0.x changes: https://github.com/aspnet/EntityFrameworkCore/commit/853c79b70fe85047e372a1185950b485068772f0 and https://github.com/aspnet/EntityFrameworkCore/commit/595b76bd3eacbb168037e520f7f077e0e91e33c5

ajcvickers: What is it in Hosting that changed that means that EF has to change? Will EF not work correctly without the new version? If so, how?

I'm fine with changing test dependencies to make sure that we test EF against the latest version--as discussed with @muratg--but updating the product dependencies seems wrong.

natemcmaster: Cascading updates like this is not new. We've done it for all patch releases so far. The thinking is that if a project has only a dependency on EFCore, users won't get the patched transitive dependencies unless they know to live the transitive ref to a top-level ref _and_ update that to the latest patch.

cc @Eilon

ajcvickers: Okay, if that's what we're going to do, then that's what we're going to do...

Eilon: Right, starting with 2.0.x this goes away because the metapackage is what people use and will lift all dependencies. But in 1.0.x/1.1.x with specific package refs, this is what we do to make sure people easily get the latest updates.

test/Microsoft.EntityFrameworkCore.Tests/ModelSourceTest.cs
```diff
@@ -79,7 +79,7 @@ public void Stores_model_version_information_as_annotation_on_model()
             var model = modelSource.GetModel(new Context1(), null,
                 new LoggingModelValidator(new Logger<LoggingModelValidator>(new LoggerFactory())));
 
-            Assert.StartsWith("1.0.4", model.GetProductVersion(), StringComparison.OrdinalIgnoreCase);
+            Assert.StartsWith("1.0.5", model.GetProductVersion(), StringComparison.OrdinalIgnoreCase);
```
bricelam: Should we just update it to `"1.0."`?

jkotalik: Yeah that definitely seems like the safer thing to do. Avoids the hassle of updating everything on each patch.

bricelam: 👍 IMHO, we can let the `#if CSPROJ` code rot too. We'll update it when we need to.

PR #9598, opened 2017-08-27
fix #9320 - UDF functions use own database

### LINKED ISSUE #9320: Test: UdfDbFunctionSqlServerTests class should not use northwind database
`UdfDbFunctionSqlServerTests` is used for functional testing for UDFs. Any time we change something and want to add more test that should be the class to go to. But at present that class uses northwind database. 
We don't want to modify northwind database since anytime we modify the script, we need to drop database from machines where it is persisted. (CI & local dev boxes).
We should let this class use northwind model (or may be new model) but use own database which would get recreated every time so that adding more UDFs don't require manual dropping of database on various machines.

This should also remove currently defined UDFs in northwind.

pmiddleton:
I have updated the Sql Server UDF tests to use their own database/model instead of Northwind.

The model is very similar to the subset of Northwind which was being used previously.  Employee is now Customer and less data is loaded.

All the tests are structurally the same.


test/EFCore.SqlServer.FunctionalTests/Query/UdfDbFunctionSqlServerTests.cs
```diff
@@ -2,47 +2,245 @@
...
+                InitalizeDatabase(context);
+
+                ClearLog();
+            }
         }
 
-        protected NorthwindQuerySqlServerFixture<NoopModelCustomizer> Fixture { get; }
+        private void CreateDbFunctions(UDFSqlContext context)
+        {
+            context.Database.ExecuteSqlCommand(@"create function [dbo].[CustomerOrderCount] (@customerId int)
```
AndriySvyryd: Add a new fixture derived from `SharedStoreFixtureBase<DbContext>` and create the functions in `Seed`  method, see `SqlServerTriggersTest`

AndriySvyryd:
Don't modify `Northwind.sql` at this point it's better to just leaving it as is

smitpatel:
@AndriySvyryd - Let's discuss tomorrow in team room about modifying northwind. It would again breaking change for dev boxes but removing the UDF functions from northwind allows us to keep the databse in sync across multiple providers running the tests.

pmiddleton:
I don't see why this would be a breaking change.  The functions will still be on your local instance, but it won't cause any issues since they won't be used anymore by the unit tests.  

Let me know what is decided in your meeting.

AndriySvyryd:
@pmiddleton We've discussed this and removing the functions from `Northwind.sql` is fine. It will cause minor issues when working on a patch branch, but will be best in the long-term.

test/EFCore.SqlServer.FunctionalTests/Query/UdfDbFunctionSqlServerTests.cs
pmiddleton: I don't see a way to access the generated sql with this fixture.  Ideas on what I can use to get access to the sql?

AndriySvyryd: Add this property to the fixture:
```C#
public TestSqlLoggerFactory TestSqlLoggerFactory => (TestSqlLoggerFactory)ServiceProvider.GetRequiredService<ILoggerFactory>();
```

pmiddleton: I figured out how to get access to the TestSqlLoggerFactory.  

Should we expose an AssertSql and ClearLog method on the SharedStoreFixtureBase?

AndriySvyryd: We cannot do that as `SharedStoreFixtureBase` is at the Core level, so it is also used for non-relational provider tests. We'll consider adding a `RelationalSharedStoreFixtureBase` if this keeps coming up.

test/EFCore.SqlServer.FunctionalTests/Query/UdfDbFunctionSqlServerTests.cs
```diff
@@ -2,47 +2,248 @@
...
+            public void ClearLog()
+                => ((TestSqlLoggerFactory)ServiceProvider.GetRequiredService<ILoggerFactory>()).Clear();
+
+            public void AssertSql(params string[] expected)
+                => ((TestSqlLoggerFactory)ServiceProvider.GetRequiredService<ILoggerFactory>()).AssertBaseline(expected);
+
+            public override DbContextOptionsBuilder AddOptions(DbContextOptionsBuilder builder)
+            {
+                base.AddOptions(builder);
+                return builder.ConfigureWarnings(w => w.Default(WarningBehavior.Ignore));
```
AndriySvyryd: Consider just ignoring specific warnings

test/EFCore.SqlServer.FunctionalTests/Query/UdfDbFunctionSqlServerTests.cs
```diff
@@ -2,47 +2,248 @@
...
+        public UdfDbFunctionSqlServerTests(SqlServerUDFFixture fixture)
         {
             Fixture = fixture;
+            Fixture.ClearLog();
+        }
+
+        private SqlServerUDFFixture Fixture { get; }
+
+        public class SqlServerUDFFixture : SharedStoreFixtureBase<DbContext>
+        {
```
smitpatel: Fixture goes to bottom of the file, no? @AndriySvyryd 
e.g. https://github.com/aspnet/EntityFrameworkCore/blob/dev/test/EFCore.SqlServer.FunctionalTests/Query/IncludeOneToOneSqlServerTest.cs

AndriySvyryd: Yes

AndriySvyryd:
@pmiddleton Thanks for your contribution!

Issue #7181, opened 2016-12-05
Relax derived FK overlapping PK restriction


AndriySvyryd:
This is only an issue if the property contained in both needs a value generator

PR #9611, opened 2017-08-29
Don't throw when adding a key over a property that is part of a derived foreign key, but doesn't have value generation enabled.

AndriySvyryd:


smitpatel:
Can you explain reason for this? How would the value generation for property defined, with key & fk both being present?

AndriySvyryd:
@smitpatel This is just a missing case from #7181

PR #9618, opened 2017-08-30
Update appveyor.yml

smitpatel:
Fix regular expression used for branch matching

It seems that appveyor is not building any branches matched by regex for us. Probably seems like our regex was incorrect. This change may not work but testing it out.

I will merge PR if appveyor/travis passes.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,9 +14,9 @@ branches:
     - master
     - release
     - dev
-    - /^rel\/.*/
-    - /^feature\/.*/
-    - /^(.*\/)?ci-.*$/
+    - /rel\/.*/
+    - /feature\/.*/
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -14,9 +14,9 @@ branches:
     - master
     - release
     - dev
-    - /^rel\/.*/
-    - /^feature\/.*/
-    - /^(.*\/)?ci-.*$/
+    - /rel\/.*/
+    - /feature\/.*/
+    - /(.*\/)?ci-.*$/
```
bricelam: If `^` doesn't work, `$` probably doesn't either.

smitpatel: Let me check if `^` was real issue then will update $ too and travis too.

PR #9628, opened 2017-08-30
Use Directory.Build.props/targets

natemcmaster:
Changes:

 - Removes the Import to build/common.props from all repos, relying instead on the automatic import of Directory.Build.props/targets
 - Move EFCore.Benchmarks.EFCore from test/ to benchmarks/. It's not really a test project.
 - Only uses Internal.AspNetCore.Sdk to src/ and test/. It's not necessary for the benchmarks projects.
 - Remove duplicate PackageReference's in test csproj files.

Issue #9207, opened 2017-07-19
Remove reference to NETStandard.Library.NETFramework


mikeharder:
`build\common.props` currently contains a reference to `NETStandard.Library.NETFramework`, to improve the developer experience when using VS 15.2.  This reference should be removed as soon as possible without degrading the developer experience, to align with other aspnet repos.

smitpatel:
I believe this can be done once everyone is using dev 15.3 RTM edition.

PR #9629, opened 2017-08-30
Use PackageLineup to manage PackageReference versions

natemcmaster:
Depends on #9628 

Part of https://github.com/aspnet/KoreBuild/issues/239

This removes the Version attribute from most PackageReference items in csproj files. The build script will generate this version information during the /t:Pin build target.

For more information about lineups, see the engineering wiki.

FYI @ryanbrandenburg


Directory.Build.props
```diff
@@ -21,6 +20,9 @@
 
   <!-- TODO: Remove when TestDriven supports VS15.3+ -->
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
-    <PackageReference Include="NETStandard.Library.NETFramework" Version="$(NETStandardLibraryNETFrameworkVersion)" PrivateAssets="All" />
+    <PackageReference Include="NETStandard.Library.NETFramework" Version="2.0.0-preview2-25405-01" PrivateAssets="All">
+      <!-- This is reference temporary and won't be added to lineups. -->
```
ryanbrandenburg: Is there an issue filed about fixing this?

natemcmaster: https://github.com/aspnet/EntityFrameworkCore/issues/9207

bricelam: Pretty sure we can remove this whole thing now...

natemcmaster: Ok, I'll update this PR to take care of this too.

Directory.Build.props
```diff
@@ -18,9 +17,4 @@
   <ItemGroup>
     <EmbeddedResource Include="**\*.rd.xml" />
   </ItemGroup>
-
-  <!-- TODO: Remove when TestDriven supports VS15.3+ -->
```
smitpatel: This should not be removed.

benchmarks/Directory.Build.props
```diff
@@ -0,0 +1,25 @@
...
+
+  <PropertyGroup>
+    <!--
+      Disables build warning/errors about PackageReference items that have versions, even though lineups are in use.
+      This is required because the benchmark projects test the last RTM release of EF and EF Core for comparison with
+      the current build of EF Core.
+    -->
+    <DisablePackageReferenceVersionRestrictions>true</DisablePackageReferenceVersionRestrictions>
+
+    <BenchmarkDotNetVersion>0.10.9</BenchmarkDotNetVersion>
```
smitpatel: You can inline these versions to csproj. It is unlikely they would change. And even in case of change it is not huge overhead.

src/EFCore.SqlServer/EFCore.SqlServer.csproj
```diff
@@ -19,8 +19,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="StyleCop.Analyzers" Version="$(StyleCopAnalyzersVersion)" PrivateAssets="All" />
-    <PackageReference Include="System.Data.SqlClient" Version="$(CoreFxVersion)" />
+    <PackageReference Include="StyleCop.Analyzers" Version="$(StyleCopAnalyzersVersion)" PrivateAssets="All">
+      <!-- Build-only assets are exempt from using a version from lineups. -->
```
smitpatel: It would be good if warning system detected this on own instead of adding NoWarn on every usage.

Directory.Build.props
natemcmaster: @bricelam said we could. I don't have TD.net to try it myself.

smitpatel: Lets keep it on safe side. We already have issue to clean it up. #9207

natemcmaster: Whatever you guys want. I rolled back the change.

natemcmaster:
:up: :date: rebased on dev and update for the projects since I opened this PR

src/EFCore.SqlServer/EFCore.SqlServer.csproj
natemcmaster: The warning is meant to guide our devs away from using Version on PackageReference and instead using the lineup to select versions. Ideally, NoWarn should be rare. EF Core is the only repo where I had to add a bunch of NoWarn's. 

FYI - this may change when we start working on build-from-source as we'll want to minimize external dependencies required to build aspnet core. cc @mikeharder for details

PR #9651, opened 2017-08-31
Test: At times we need to wait :)

smitpatel:
Also testing travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,11 @@ branches:
     - master
     - release
     - dev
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -19,7 +19,11 @@ branches:
     - master
     - release
     - dev
+    - /^rel\/.*/
```
maumar: unrealted?

smitpatel: testing building of feature branches on travis

Issue #9613, opened 2017-08-29
In-memory database not persisted across service providers


ajcvickers:
Each named in-memory database is rooted in the internal service provider. This allows it to persist across multiple uses of a context without being rooted in the app-domain, which could cause leaks. Most of the time this works as expected, but sometimes an options change can cause EF to build a new internal service provider. This causes the two contexts to not be sharing the in-memory database even when it is likely expected they will.

For example:
```C#
using (var context = new FooContext(
    new DbContextOptionsBuilder()
        .UseInMemoryDatabase(nameof(FooContext))
        .Options))
{
    context.Add(new Foo());
    context.SaveChanges();
}

using (var context = new FooContext(
    new DbContextOptionsBuilder()
        .UseInMemoryDatabase(nameof(FooContext))
        .EnableSensitiveDataLogging()
        .Options))
{
    // Fails because one context has EnableSensitiveDataLogging while the other doesn't
    Debug.Assert(context.Foos.Count() == 1);
}
```

ajcvickers:
Note that this can happen if one DbContext is resolved from an IServiceProvider after being registered with AddDbContext, and the other DbContext instance is created with `new'. One workaround is to always resolve the context instance from the service provider. For example:
```C#
using (var scope = serviceProvider.CreateScope())
{
    var db = scope.ServiceProvider.GetService<MyContext>();
}
```
In the case of a web app, the IServiceProvider can be obtained by calling `IWebHost.Services`.

ajcvickers:
@radenkozec Please file a new issue including a runnable project/solution or complete code listing that demonstrates the behavior you are seeing.

radenkozec:
@ajcvickers I manages to solve the issues by following this tutorial http://www.dotnetcurry.com/aspnet-core/1420/integration-testing-aspnet-core
It seams that my issue was related to wrong dependency injection configuration.

radenkozec:
@ajcvickers I have a big problems with dependency injection and InMemory from Integration tests.
I tried to use same instance of InMemoryDatabaseRoot with no success.
What is funny this all was piece of cake in .NET Core 1. From version 2 things are much harder.
 ```
 public class TestStartup : Startup
    {
        public TestStartup(IConfiguration configuration) : base(configuration)
        {
        }
        public override void SetUpDataBase(IServiceCollection services)
        {
            var inMemoryDatabaseRoot = new InMemoryDatabaseRoot();
[...]

PR #9667, opened 2017-09-01
Allow a global (app-domain) in-memory store

ajcvickers:
Issue #9613

The internal service provider created by AddDbContext is typically different than the one created when `new` is used. This is because AddDbContext automatically sets ILoggerFactory, etc. from the application service provider, while `new` does not have access to this and so uses a different ILoggerFactory. The same thing can happen with other 'singleton options' like EnableSensitiveDataLogging.

For the in-memory database, this has two consequences:
* Since the store is rooted in the service provider, a new store is created when a situation like above forces a new internal service provider
* Since the store uses `IEntityType` instances to key its table dictionary, these instances must come from the same model to match, which is typically not the case when the model is built twice; once for each service provider

This fix adds a new flag to UseInMemoryDatabase such that the database will really be created globally (i.e. once per app domain) and entity type names are uses as keys instead of IEntityType instances. This could result in memory leaks, but this can be mitigated by seeding with well-known for the test, and/or calling EnsureDeleted appropriately to clean out the data.

### DIFF
--- a/src/EFCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
+++ b/src/EFCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
@@ -20,7 +20,10 @@ namespace Microsoft.EntityFrameworkCore
 
         /// <summary>
         ///     Configures the context to connect to an in-memory database.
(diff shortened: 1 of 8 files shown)


ajcvickers:
Superceeded by #9692

Issue #9551, opened 2017-08-24
Expression of type 'System.Object' cannot be used for parameter of type 'Microsoft.EntityFrameworkCore.Metadata.IEntityType'


dannythomas13:
Since updating to v2 I now get the following exception from one of my unit tests:

```
System.ArgumentException : Expression of type 'System.Object' cannot be used for parameter of type 'Microsoft.EntityFrameworkCore.Metadata.IEntityType' of method 'Void StartTracking(System.Object, Microsoft.EntityFrameworkCore.Metadata.IEntityType)'
Parameter name: arg1
```
In order to get the test to pass I have had to change:
```
var projects = this.context.Projects
                        .Include(p => p.Windfarm)
                        .Include(p => p.Turbines)
                        .Where(p => p.Windfarm.ClientId == clientId && p.IsActive)
                        .GroupBy(g => g.WindfarmId)
.Select(s => s.OrderByDescending(o => o.DateStarted).FirstOrDefault())
                        .ToList();
```
to the following:
```
var activeProjects = this.context.Projects
                        .Include(p => p.Windfarm)
                        .Include(p => p.Turbines)
                        .Where(p => p.Windfarm.ClientId == clientId && p.IsActive)
                        .GroupBy(g => g.WindfarmId).ToList();

            var projects = activeProjects.Select(s => s.OrderByDescending(o => o.DateStarted).FirstOrDefault())
                        .ToList();
```

smitpatel:
```C#
    public class Program
    {
        public static void Main(string[] args)
        {
            using (var db = new MyContext())
            {
                // Recreate database
                db.Database.EnsureDeleted();
                db.Database.EnsureCreated();

                // Seed database
                db.AddRange(
                    new Blog
                    {
                        Posts = new List<Post>
                        {
                            new Post(),
                            new Post(),
                            new Post()
                        }
                    },
                    new Blog
                    {
                        Posts = new List<Post>
                        {
                            new Post(),
                            new Post(),
                            new Post()
                        }
                    }
                    );


                db.SaveChanges();
            }

            using (var db = new MyContext())
            {
                // Run queries
                var query = db.Set<Post>()
                    .Include(p => p.Blog)
                    .GroupBy(p => EF.Property<int?>(p, "BlogId"))
                    .Select(g => g.OrderBy(e => e.Id).FirstOrDefault()).ToList();
            }

            Console.WriteLine("Program finished.");
        }
    }


    public class MyContext : DbContext
[...]

PR #9670, opened 2017-09-01
[2.0.1] Fix to #9551 - Expression of type 'System.Object' cannot be used for parameter of type 'Microsoft.EntityFrameworkCore.Metadata.IEntityType'

maumar:
Problem was that we made some assumptions when compiling include, that do not hold true for complex queries involving group by and other operations that follow.

Once we build _Include(...) expression and try to substitute a qsre with it, we have compensation logic in case of groupby - trying to apply the substitution on the ElementSelector of the groupby. This logic is located in QueryModelExtensions.GetOutputExpression.
However, we assume that the outermost query model is the one that contains the groupby result operator. For the queries in the bug, this is not the case - GroupBy result operator is a subquery and therefore our old logic was not detecting it correctly, causing an error.

Fix is to find a correct query model (based on the qsre used in the Include result operator) and only after the correct QM (which can be in a subquery) is obtained, perform the compensation logic in QueryModelExtensions.GetOutputExpression.

Additional fix is needed when compiling collection navigation includes. We manually craft the QM that fetches child elements, and also assume that the outermost parent QM is the one that needs to be joined to. Similar approach is used here - find the "real" QM referenced in the include and build the collection-fetching QM based on that instead.


maumar:
my flight got cancelled :(

smitpatel:
Test failures

src/EFCore/Query/ExpressionVisitors/Internal/QueryModelFindingVisitor.cs
```diff
@@ -0,0 +1,95 @@
...
+            queryModel.TransformExpressions(new TransformingQueryModelExpressionVisitor<QueryModelFindingVisitor>(this).Visit);
+
+            base.VisitQueryModel(queryModel);
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public override void VisitMainFromClause(MainFromClause fromClause, QueryModel queryModel)
```
smitpatel: Do we need visitor for this?
Specifically since we are just finding querysource, we can just compare mainfromclause to _querySource &
BodyClauses.Any(b => (b as IQuerySource) == _querySource) in single method.

ajcvickers:
@anpete @smitpatel Given that @maumar is out and this is for 2.0.1, can one of you take it over?

src/EFCore/Query/ExpressionVisitors/Internal/QueryModelFindingVisitor.cs
maumar: visitor is here just to peek into subqueries

smitpatel: We need ExpressionVisitor to peek into subqueries but we don't need QMVisitor. It can be just a private method which would work on a QM instead.

maumar: didn't we create TransformingQueryModelExpressionVisitor pattern exactly for this purpose?

smitpatel: The purpose of `TransformingQueryModelExpressionVisitor` is to apply QM transformations recursively in subquery. In this particular case, the QM visitor just walk the QM instead of transforming it into something else. Further the "walk/transformation" is simple enough that we don't need to utilize QueryModelVisitor to visit whole QM. Therefore, the whole task of this QMV can be simplified into single method which would work on QM and combining it with ExpressionVisitor to peek into subqueries simplifies code.
If we are certain that whatever is happening in `QueryModelFindingVisitor` is not going to change a lot in future then lets not put in complex solution.

smitpatel: On different dimension, there are multiple operators which are deriving from IQuerySource apart from `MainFromClause` & `AdditionalFromClause` like `JoinClause` or `GroupResultOperator` or `ConcatResultOperator` do we need to consider them too?

natemcmaster:
Can you retarget this to the rel/2.0.1 branch?

natemcmaster:
Sorry for accidental close. Thanks for reopening @smitpatel

anpete:
Superseded by #9862

PR #9692, opened 2017-09-04
Allow passing a root for in-memory databases

ajcvickers:
Alternate fix for issue #9613

The internal service provider created by AddDbContext is typically different than the one created when `new` is used. This is because AddDbContext automatically sets ILoggerFactory, etc. from the application service provider, while `new` does not have access to this and so uses a different ILoggerFactory. The same thing can happen with other 'singleton options' like EnableSensitiveDataLogging.

For the in-memory database, this has two consequences:
* Since the store is rooted in the service provider, a new store is created when a situation like above forces a new internal service provider
* Since the store uses `IEntityType` instances to key its table dictionary, these instances must come from the same model to match, which is typically not the case when the model is built twice; once for each service provider

This fix adds a new object that can be passed to UseInMemoryDatabase such that the database can be rooted globally, however the application wishes. Entity type names are uses as keys instead of IEntityType instances.

### DIFF
--- a/src/EFCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
+++ b/src/EFCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
@@ -6,6 +6,7 @@ using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Diagnostics;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Infrastructure.Internal;
+using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Utilities;
 
 // ReSharper disable once CheckNamespace
@@ -20,7 +21,10 @@ namespace Microsoft.EntityFrameworkCore
 
         /// <summary>
         ///     Configures the context to connect to an in-memory database.
-        ///     The in-memory database is shared anywhere the same name is used.
+        ///     The in-memory database is shared anywhere the same name is used, but only for a given
+        ///     service provider. To use the in-memory database globally, call
+        ///     <see cref="UseInMemoryDatabase{TContext}(DbContextOptionsBuilder{TContext},string,InMemoryDatabaseRoot,Action{InMemoryDbContextOptionsBuilder})" />
+        ///     passing an <see cref="InMemoryDatabaseRoot"/> to root the in-memory databases globally.
         /// </summary>
         /// <typeparam name="TContext"> The type of context being configured. </typeparam>
         /// <param name="optionsBuilder"> The builder being used to configure the context. </param>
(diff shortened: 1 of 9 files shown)


src/EFCore.InMemory/Storage/InMemoryDatabaseRoot.cs
```diff
@@ -0,0 +1,21 @@
...
+    ///     <see
+    ///         cref="InMemoryDbContextOptionsExtensions.UseInMemoryDatabase{TContext}(DbContextOptionsBuilder{TContext},string,System.Action{Infrastructure.InMemoryDbContextOptionsBuilder})" />
+    /// </summary>
+    public sealed class InMemoryDatabaseRoot
+    {
+        /// <summary>
+        ///     Entity Framework code will set this instance as needed. It should be considered opaque to
+        ///     application code. The type of object may change at any time.
+        /// </summary>
+        public object Instance;
```
divega: Should we make this internal?

src/EFCore.InMemory/Storage/InMemoryDatabaseRoot.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.EntityFrameworkCore.Storage
+{
+    /// <summary>
+    ///     Acts as a root for all in-memory databases such that they will be available
+    ///     across context instances and service providers as long as the same instance
+    ///     of this type is passed to
+    ///     <see
+    ///         cref="InMemoryDbContextOptionsExtensions.UseInMemoryDatabase{TContext}(DbContextOptionsBuilder{TContext},string,System.Action{Infrastructure.InMemoryDbContextOptionsBuilder})" />
+    /// </summary>
+    public sealed class InMemoryDatabaseRoot
```
divega: Seems to be a reasonable name.

src/EFCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
```diff
@@ -20,7 +21,10 @@ public static class InMemoryDbContextOptionsExtensions
 
         /// <summary>
         ///     Configures the context to connect to an in-memory database.
-        ///     The in-memory database is shared anywhere the same name is used.
+        ///     The in-memory database is shared anywhere the same name is used, but only for a given
+        ///     service provider. To use the in-memory database globally, call
+        ///     <see cref="UseInMemoryDatabase{TContext}(DbContextOptionsBuilder{TContext},string,InMemoryDatabaseRoot,Action{InMemoryDbContextOptionsBuilder})" />
+        ///     passing an <see cref="InMemoryDatabaseRoot"/> to root the in-memory databases globally.
```
divega: > To use the in-memory database globally,... to root the in-memory databases globally. 

Seems redundant.

Perhaps something like this:
``` csharp
///     The in-memory database is shared anywhere the same name is used, but only for a given 
///     service provider. To use the same in-memory database across service providers, call 
///     <see cref="UseInMemoryDatabase{TContext}(DbContextOptionsBuilder{TContext},string,InMemoryDatabaseRoot,Action{InMemoryDbContextOptionsBuilder})" /> 
///     passing a shared <see cref="InMemoryDatabaseRoot"/> on which to root the database. 
```

src/EFCore.InMemory/Storage/InMemoryDatabaseRoot.cs
ajcvickers: Documented explicitly as internal.

src/EFCore.InMemory/Extensions/InMemoryDbContextOptionsExtensions.cs
ajcvickers: Updated

ajcvickers:
Merged

PR #9698, opened 2017-09-05
Include EFCore.Design in EFCore.Runtime.sln

bricelam:
It's important to include this project when refactoring runtime code.

And hey, even I don't like to see all those tools packages sometimes. 😉


bricelam:
Metadata was my main concern. We changed the behavior of `Relational().StoreType` and needed to react in Design.

bricelam:
A lot of new features also require implementation in Design. Design uses Internal APIs too so you may break it without realizing it...

bricelam:
Do we need an EFCore.Query.sln?

bricelam:
Tools on the other hand are very isolated and only need to be changed when additional new features to them.

smitpatel:
Let's discuss in triage today. EFCore.Query.sln is also fine :)

PR #9702, opened 2017-09-06
Allow multiple IModelCustomizer

### LINKED ISSUE #9330: Allow multiple IModelCustomizer
This would allow users to add an IModelCustomizer without having to override the provider one

AndriySvyryd:
Add IModelCustomizerCollection to allow ordering the registered IModelCustomizer services

Fixes #9330


src/EFCore.Relational/Infrastructure/EntityFrameworkRelationalServicesBuilder.cs
```diff
@@ -126,7 +126,7 @@ public override EntityFrameworkServicesBuilder TryAddCoreServices()
             TryAdd<IMigrationsIdGenerator, MigrationsIdGenerator>();
             TryAdd<IKeyValueIndexFactorySource, KeyValueIndexFactorySource>();
             TryAdd<IModelSource, RelationalModelSource>();
-            TryAdd<IModelCustomizer, RelationalModelCustomizer>();
+            TryAdd<ModelCustomizer, RelationalModelCustomizer>();
```
anpete: Does this need to change?

src/EFCore/Infrastructure/IModelCustomizerCollection.cs
```diff
@@ -0,0 +1,15 @@
+using System.Collections.Generic;
+
+namespace Microsoft.EntityFrameworkCore.Infrastructure
+{
+    /// <summary>
+    ///     Contains an ordered collection of <see cref="IModelCustomizer"/>
+    /// </summary>
+    public interface IModelCustomizerCollection
```
anpete: What is the purpose of this vs just having IEnumerable<IModelCustomizer> in DI?

src/EFCore/Infrastructure/ModelSourceDependencies.cs
```diff
@@ -64,10 +65,10 @@ public sealed class ModelSourceDependencies
         public ICoreConventionSetBuilder CoreConventionSetBuilder { get; }
 
         /// <summary>
-        ///     Gets the <see cref="IModelCustomizer" /> that will perform additional configuration of the model
+        ///     Gets the <see cref="IModelCustomizer" /> collection that will perform additional configuration of the model
         ///     in addition to what is discovered by convention.
         /// </summary>
-        public IModelCustomizer ModelCustomizer { get; }
```
anpete: Break

src/EFCore/Infrastructure/ModelSourceDependencies.cs
```diff
@@ -45,15 +46,15 @@ public sealed class ModelSourceDependencies
         /// </summary>
         public ModelSourceDependencies(
             [NotNull] ICoreConventionSetBuilder coreConventionSetBuilder,
-            [NotNull] IModelCustomizer modelCustomizer,
```
anpete: Break

src/EFCore/Infrastructure/EntityFrameworkServicesBuilder.cs
```diff
@@ -66,7 +66,9 @@ public class EntityFrameworkServicesBuilder
                 { typeof(IEntityMaterializerSource), new ServiceCharacteristics(ServiceLifetime.Singleton) },
                 { typeof(ICoreConventionSetBuilder), new ServiceCharacteristics(ServiceLifetime.Singleton) },
                 { typeof(ITypeMapper), new ServiceCharacteristics(ServiceLifetime.Singleton) },
-                { typeof(IModelCustomizer), new ServiceCharacteristics(ServiceLifetime.Singleton) },
+                { typeof(ModelCustomizer), new ServiceCharacteristics(ServiceLifetime.Singleton) },
```
ajcvickers: >ModelCustomizer [](start = 25, length = 15)

No concrete types here--all our services are registered by interface contract.

ajcvickers:
@AndriySvyryd If I have an application that is currently registering an IModelCustomizer that extends from ModelCustomizer or RelationalModelCustomizer, will this change result in the customizer code in those classes being run twice? If so, could this be a breaking change, or is this code always idempotent?

Conversely, if I am registering an IModelCustomerizer specifically to change the behavior of the default, will this change result in the default behavior now being used again--i.e. potentially a break. (Example: someone specifically trying to switching off DbSet name to table name convention?

AndriySvyryd:
@ajcvickers With this change the user's code expecting to switch off the default ModelCustomizer would be broken.

We can introduce a new interface, e.g. `IAdditionalModelCustomizer` that would be additive, while preserving the current `IModelCustomizer` behavior.

src/EFCore/Infrastructure/ModelSourceDependencies.cs
AndriySvyryd: These constructors are specifically designated as breakable API

src/EFCore/Infrastructure/IModelCustomizerCollection.cs
AndriySvyryd: To allow ordering

AndriySvyryd:
Removed the breaking change

ajcvickers:
This looks functionally good, but things are so clunky now, just because of the path we have taken to get here. Would it work to change the registration for IModelCustomizer to "allow multiple" and then remove the additional interface? Ours would then be registered as a normal IModelCustomizer, and we could identify it and run it first at the point that we consume it.

AndriySvyryd:
@ajcvickers How would we identify an `IModelCustomizer` as 'ours'? Is it one derived from `ModelCustomizer` ? What if there are multiple?
Besides this would still break current code that registers a `IModelCustomizer` not derived from `ModelCustomizer` and expects the `ModelCustomizer` to not run.

ajcvickers:
@AndriySvyryd I'm okay with that break--I don't think it is why people have been using their own customizer.

That aside, how about:
* If there is only one registered, then we run it
* If multiple are registered, then:
   * If none of them derive from ModelCustomizer, then we run them all
   * If exactly one of those derives from ModelCustomizer, then we run that one first, and all the rest after
   * If more than one derives from ModelCustomizer, order by "how derived" they are, and run only the most derived one
   * If multiple come out that are the same derived, then throw or warn and do something like running only one of them or both in an arbitrary order

So, considering these types:
* ModelCustomizer
* RelationalModelCustomizer : ModelCustomizer
* MyRelationalCustomizer : RelationalModelCustomizer : ModelCustomizer
* MyCustomizer : ModelCustomizer

Then
* If ModelCustomizer registered, then run ModelCustomizer
* If ModelCustomizer, RelationalModelCustomizer registered, then run RelationalModelCustomizer
* If ModelCustomizer, RelationalModelCustomizer, MyRelationalCustomizer registered, then run MyRelationalCustomizer
* If ModelCustomizer, MyCustomizer registered, then run MyCustomizer 
* If ModelCustomizer, RelationalCustomizer, MyCustomizer registered, then warn/throw

AndriySvyryd:
@ajcvickers That can work, but I don't see that it's less clunky or easier to explain than `IAdditionalModelCustomizer`:

- To replace the default customizer register your service as `IModelCustomizer`
- To add a customizer register your service as `IAdditionalModelCustomizer`

ajcvickers:
I think we should discuss in person. Others here also have opinions. 😸

AndriySvyryd:
We decided there's a better way this could be accomplished in the future

PR #9729, opened 2017-09-07
Improve IModelCustomizer doc comment

AndriySvyryd:
And some test cleanup

### DIFF
--- a/test/EFCore.Tests/DbContextServicesTest.cs
+++ b/test/EFCore.Tests/DbContextServicesTest.cs
@@ -20,6 +20,10 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Xunit;
 
+// ReSharper disable ClassNeverInstantiated.Local
+// ReSharper disable UnusedMember.Local
+// ReSharper disable UnusedAutoPropertyAccessor.Local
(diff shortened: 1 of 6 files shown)


test/EFCore.Tests/DbContextServicesTest.cs
```diff
@@ -2967,7 +2988,9 @@ public void Throws_changing_warnings_config_with_AddDbContext_when_UseInternalSe
             {
                 var context = serviceScope.ServiceProvider.GetService<ConstructorTestContextWithOC3A>();
 
-                _ = context.Model;
+                var model = context.Model;
```
smitpatel: do we need to store in the model var? how about `Assert.NotNull(context.Model)`?

test/EFCore.Tests/DbContextServicesTest.cs
```diff
@@ -2166,8 +2172,9 @@ public void Can_use_memory_cache_before_context_exists_and_after_disposed_when_l
                 .CreateScope())
             {
                 var context = serviceScope.ServiceProvider.GetService<DbContext>();
-                _ = context.Model;
 
+
```
smitpatel: extra line break. :trollface:

Issue #9742, opened 2017-09-08
Query: Navigation rewrite throws null check on navigation


smitpatel:
```c#
        [Theory]
        [InlineData(false)]
        [InlineData(true)]
        public virtual void Include_collection_with_multiple_conditional_order_by(bool useString)
        {
            using (var context = CreateContext())
            {
                var orders
                    = useString
                        ? context.Orders
                            .Include("OrderDetails")
                            .OrderBy(o => o.OrderID > 0)
                            .ThenBy(o => o.Customer != null ? o.Customer.City : String.Empty)
                            .Take(5)
                            .ToList()
                        : context.Orders
                            .Include(c => c.OrderDetails)
                            .OrderBy(o => o.OrderID > 0)
                            .ThenBy(o => o.Customer != null ? o.Customer.City : String.Empty)
                            .Take(5)
                            .ToList();

                foreach (var order in orders)
                {
                    CheckIsLoaded(
                        context,
                        order,
                        orderDetailsLoaded: true,
                        productLoaded: false,
                        customerLoaded: false,
                        ordersLoaded: false);
                }
            }
        }
```
```
Output from Microsoft.EntityFrameworkCore.Query.IncludeSqlServerTest.Include_collection_with_multiple_conditional_order_by(useString: False):
  Entity Framework Core 2.1.0-preview1 initialized 'NorthwindRelationalContext' using provider 'Microsoft.EntityFrameworkCore.SqlServer' with options: SensitiveDataLoggingEnabled CommandTimeout=600
  
  Compiling query model: 
  '(from Order o in 
      (from Order c in DbSet<Order>
      select [c]).Include("OrderDetails")
  order by [o].OrderID > 0 asc, [o].Customer != null ? [o].Customer.City : __Empty_0 asc
  select [o]).Take(__p_1)'
  
  Including navigation: '[c].OrderDetails'
Test 'Microsoft.EntityFrameworkCore.Query.IncludeSqlServerTest.Include_collection_with_multiple_conditional_order_by(useString: True)' failed:
	System.InvalidOperationException : The operands for operator 'NotEqual' do not match the parameters of method 'op_Inequality'.
	at System.Linq.Expressions.Expression.GetMethodBasedBinaryOperator(ExpressionType binaryType, Expression left, Expression right, MethodInfo method, Boolean liftToNull)
	at System.Linq.Expressions.Expression.NotEqual(Expression left, Expression right, Boolean liftToNull, MethodInfo method)
[...]

PR #9743, opened 2017-09-09
Query: When we PushDownSubquery update aliases of ordering expressions

### LINKED ISSUE #9735: A column has been specified more than once in the order by list (MSSQL)
var res =
                ctx.B.Include(b => b.Cs).
                OrderBy(b => b.A.AId > 0).ThenBy(b => b.D != null ? b.D.Name : string.Empty).
                Take(2).ToList();

Generates the invalid SQL:
```SQL
SELECT [b.Cs].[CId], [b.Cs].[BId], [b.Cs].[Name]
FROM [C] AS [b.Cs]
INNER JOIN (
    SELECT DISTINCT [t].*
    FROM (
        SELECT TOP(@__p_1) [b0].[BId], CASE
            WHEN [b0].[AId] > 0
        END AS [c], CASE
            WHEN [b0].[DId] IS NOT NULL
            THEN [b.D0].[Name] ELSE @__Empty_0
        END AS [c0], [b0].[AId], [b0].[DId], [b.D0].[Name]
        FROM [B] AS [b0]
        LEFT JOIN [D] AS [b.D0] ON [b0].[DId] = [b.D0].[DId]
        ORDER BY CASE
            WHEN [b0].[AId] > 0
        END, [c], [b0].[BId]
    ) AS [t]
) AS [t0] ON [b.Cs].[BId] = [t0].[BId]
ORDER BY [t0].[c], [t0].[c0], [t0].[BId]
```

The inner order by uses the wrong alias. Should be
        ORDER BY CASE
            WHEN [b0].[AId] > 0
        END, [c0], [b0].[BId]

```c#
using System;
using System.Linq;

namespace EFDebug
{
    class Program
    {
        {
            var loggerFactory = new LoggerFactory();
            loggerFactory.AddDebug();
            var connection = @"";
            var ctx = new MyContext(new DbContextOptionsBuilder().UseSqlServer(connection).UseLoggerFactory(loggerFactory).Options);
            var res =
                ctx.B.Include(b => b.Cs).
                OrderBy(b => b.A.AId > 0).ThenBy(b => b.D != null ? b.D.Name : string.Empty).
                Take(2).ToList();
        }
    }
    public class A
    {
        public int AId { get; set; }
    }

    public class B
    {
        public int BId { get; set; }


        public int AId { get; set; }
        public virtual A A { get; set; }

        public virtual List<C> Cs { get; set; }
        public virtual D D { get; set; }
        public int? DId { get; set; }

        //This works
        //public int DId { get; set; }
    }

    public class C
    {
        public int CId { get; set; }
        public int BId { get; set; }
        public virtual B B { get; set; }
    }

    public class D
    {
        public int DId { get; set; }
    }

    {
        public MyContext(DbContextOptions options)
            : base(options)
        {

        }        
        public DbSet<B> B { get; set; }        
    }


//Create Table A(
//    AId int identity primary key,
//    name varchar(255)
//)
//create table D(
//        DId int identity primary key,
//        name varchar(255),
//)
//create Table B(
//    BId int identity primary key,
//    name varchar(255),
//	AId int not null references A(Aid),
//	DId int references D(Did),
//)
//create table C(
//        CId int identity primary key,
//        name varchar(255),
//		BId int references B(Bid)
//)
//insert into A(name) values('An A');
//    insert into B(name, AId) values('A B',1);
}

```
The exception is not thrown when making B.DId not nullable

### Further technical details
EF Core 2 and 2.1.0-preview1-26938
Windows, core 2.0 and .net 4.7

smitpatel:
Issue: Since OrderBy comes after Projection, for smaller sql, we refer to ordering as alias when it is present in projection. When we pushdown a subquery, we generate unique alias for each projection. This can cause change of alias for existing alias expression which may be referred by ordering. So even though ordering expression is correct one but since we only print out alias it may cause change of the expression actually used.

Fix: When pushing down subquery, we need to account for already aliased expression being in orderby. Hence we would store a map so that we can track change of expression and update orderby accordingly

Test: Due to #9742 added test in QueryBugsTest

Resolves #9735

### DIFF
--- a/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
+++ b/src/EFCore.Relational/Query/Expressions/SelectExpression.cs
@@ -337,10 +337,16 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
 
             var projectionsToAdd = IsProjectStar ? _starProjection : _projection;
             var outerProjections = new List<Expression>();
+            var aliasExpressionMap = new Dictionary<AliasExpression, Expression>();
 
             foreach (var expression in projectionsToAdd)
             {
                 var expressionToAdd = subquery.CreateUniqueProjection(expression);
+
+                if (expression is AliasExpression aliasExpression)
+                {
+                    aliasExpressionMap.Add(aliasExpression, expressionToAdd);
+                }
 
                 if (IsProjectStar)
                 {
@@ -366,8 +372,15 @@ namespace Microsoft.EntityFrameworkCore.Query.Expressions
             }
 
             subquery._tables.AddRange(_tables);
-            subquery._orderBy.AddRange(_orderBy);
 
+            foreach (var ordering in _orderBy)
+            {
+                subquery.AddToOrderBy(
+                    ordering.Expression is AliasExpression aliasExpression &&
+                    aliasExpressionMap.TryGetValue(aliasExpression, out var newExpression)
+                        ? new Ordering(newExpression, ordering.OrderingDirection)
+                        : ordering);
+            }
             subquery.Predicate = Predicate;
 
             subquery._limit = _limit;
--- a/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
+++ b/src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
@@ -1012,15 +1012,6 @@ namespace Microsoft.EntityFrameworkCore.Query
                         break;
                     }
 
-                    if (sqlOrderingExpression.IsComparisonOperation()
-                        || sqlOrderingExpression.IsLogicalOperation())
-                    {
-                        sqlOrderingExpression = Expression.Condition(
-                            sqlOrderingExpression,
-                            Expression.Constant(true, typeof(bool)),
-                            Expression.Constant(false, typeof(bool)));
-                    }
-
                     orderings.Add(
                         new Ordering(
                             sqlOrderingExpression,
--- a/test/EFCore.SqlServer.FunctionalTests/Query/QueryBugsTest.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/Query/QueryBugsTest.cs
@@ -33,8 +33,8 @@ namespace Microsoft.EntityFrameworkCore.Query
         public QueryBugsTest(SqlServerFixture fixture, ITestOutputHelper testOutputHelper)
         {
             Fixture = fixture;
-            fixture.TestSqlLoggerFactory.Clear();
-            //fixture.TestSqlLoggerFactory.SetTestOutputHelper(testOutputHelper);
+            Fixture.TestSqlLoggerFactory.Clear();
+            //Fixture.TestSqlLoggerFactory.SetTestOutputHelper(testOutputHelper);
         }
 
         protected SqlServerFixture Fixture { get; }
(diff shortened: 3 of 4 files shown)


src/EFCore.Relational/Query/RelationalQueryModelVisitor.cs
```diff
@@ -1012,15 +1012,6 @@ var sqlOrderingExpression
                         break;
                     }
 
-                    if (sqlOrderingExpression.IsComparisonOperation()
```
smitpatel: Removed this code because we already do conversion to value from condition in our SQL Gen phase in more general way. Also it allows us to match ordering by condition added in projection which was not happening earlier.

test/EFCore.SqlServer.FunctionalTests/Query/QueryBugsTest.cs
```diff
@@ -2457,6 +2457,125 @@ private SqlServerTestStore CreateDatabase9038()
...
+            public int CustomerId { get; set; }
+            public virtual Customer9735 Customer { get; set; }
+        }
+
+        public class CustomerDetails9735
+        {
+            public int Id { get; set; }
+            public string Name { get; set; }
+        }
+
```
anpete: formatting

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -27,7 +27,7 @@ public class SelectExpression : TableExpressionBase
         internal string DebugView => ToString();
 #endif
 
-        private static readonly ExpressionEqualityComparer _expressionEqualityComparer = new ExpressionEqualityComparer();
+        private static readonly ExpressionEqualityComparer ExpressionEqualityComparer = new ExpressionEqualityComparer();
```
anpete: Don't rename this.

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -337,11 +337,17 @@ public virtual SelectExpression PushDownSubquery()
 
             var projectionsToAdd = IsProjectStar ? _starProjection : _projection;
             var outerProjections = new List<Expression>();
+            var aliasExpressionMap = new Dictionary<AliasExpression, Expression>();
 
             foreach (var expression in projectionsToAdd)
             {
                 var expressionToAdd = subquery.CreateUniqueProjection(expression);
 
+                if (expression is AliasExpression ae)
```
anpete: ae -> "aliasExpression"

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -366,8 +372,19 @@ public virtual SelectExpression PushDownSubquery()
...
 
             subquery._tables.AddRange(_tables);
-            subquery._orderBy.AddRange(_orderBy);
 
+            foreach (var ordering in _orderBy)
+            {
+                if (ordering.Expression is AliasExpression aliasExpression
+                    && aliasExpressionMap.TryGetValue(aliasExpression, out var newExpression))
+                {
+                    subquery._orderBy.Add(new Ordering(newExpression, ordering.OrderingDirection));
```
anpete: Use public AddToOrderBy method instead of private field?

anpete: Consider ternary expression inside call to Add instead of if statement.

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
smitpatel: R# complained about it

smitpatel:
@anpete - Updated

test/EFCore.SqlServer.FunctionalTests/Query/QueryBugsTest.cs
```diff
@@ -1177,7 +1177,7 @@ public virtual void Repro3101_nested_coalesce1()
                                     on eVersion.RootEntityId equals (int?)eRoot.Id
                                     into RootEntities
                                 from eRootJoined in RootEntities.DefaultIfEmpty()
-                                // ReSharper disable once ConstantNullCoalescingCondition
+                                    // ReSharper disable once ConstantNullCoalescingCondition
```
anpete: Disable at file level?

Issue #9608, opened 2017-08-28
Should quirk switch names contain bug number?


maumar:
For patches we add opt-out mechanism via AppContext.TryGetSwitch(switchName, out var isEnabled). 

Should the switch name contain (or be?) a bug number that the quirk is affecting?

ajcvickers:
For new quirks, we will use something like: "Microsoft.EntityFrameworkCore.Issue9608"

PR #9752, opened 2017-09-10
Update existing 2.0.1 quirks to use issue number

ajcvickers:
Issue #9608

### DIFF
--- a/src/EFCore.InMemory/Diagnostics/InMemoryEventId.cs
+++ b/src/EFCore.InMemory/Diagnostics/InMemoryEventId.cs
@@ -69,7 +69,7 @@ namespace Microsoft.EntityFrameworkCore.Diagnostics
         {
             public static EventId Create(int id, string name)
             {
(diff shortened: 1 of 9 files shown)


src/EFCore.InMemory/Diagnostics/InMemoryEventId.cs
```diff
@@ -69,7 +69,7 @@ private static class EventIdFactory
         {
             public static EventId Create(int id, string name)
             {
-                if (AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Diagnostics.UseLegacyEventIds", out var isEnabled)
+                if (AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Issue9437", out var isEnabled)
```
anpete: Should we keep the informational description, too?

ajcvickers: I think it's pretty easy to look up the issue, and since the code gets removed once it is ported to 2.1 I don't think there is much value adding more to the code.

Issue #9605, opened 2017-08-28
Removing a dependent with an identifying one-to-many throws


AndriySvyryd:
Add this to `ChangeTrackerTest`:

```C#
[Fact]
public void Can_remove_dependent_identifying_one_to_many()
{
    using (var context = new EarlyLearningCenter())
    {
        var product = new Product();

        var order = new Order();

        var orderDetails = new OrderDetails { Order = order, Product = product };

        context.Add(orderDetails);

        order.OrderDetails.Remove(orderDetails);

        context.ChangeTracker.DetectChanges();
    }
}
```
>	System.ArgumentNullException : Value cannot be null.
Parameter name: key
	at System.Collections.Generic.Dictionary`2.Remove(TKey key)
	C:\enlistments\Universe\EntityFramework\src\EFCore\ChangeTracking\Internal\IdentityMap.cs(312,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Remove(TKey key, InternalEntityEntry entry)
	C:\enlistments\Universe\EntityFramework\src\EFCore\ChangeTracking\Internal\IdentityMap.cs(297,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Remove(InternalEntityEntry entry)
	C:\enlistments\Universe\EntityFramework\src\EFCore\ChangeTracking\Internal\StateManager.cs(459,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.StopTracking(InternalEntityEntry entry)
	C:\enlistments\Universe\EntityFramework\src\EFCore\ChangeTracking\Internal\InternalEntityEntry.cs(232,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges)
	C:\enlistments\Universe\EntityFramework\src\EFCore\ChangeTracking\Internal\InternalEntityEntry.cs(79,0): at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState entityState, Boolean acceptChanges, Boolean forceStateWhenUnknownKey)
[...]

PR #9754, opened 2017-09-10
Make sure state change happens before FKs are nulled

ajcvickers:
Issue #9605

The issue here was that FKs were being set to null even if that resulted in setting the PK to null, which then resulted in an issue when attempting to change the state of the entity that now has a null key.


src/EFCore/ChangeTracking/Internal/NavigationFixer.cs
```diff
@@ -802,9 +803,12 @@ private void DelayedFixup(InternalEntityEntry entry, INavigation navigation, Int
                 }
             }
 
+            var quirked = AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Issue9605", out var isEnabled)
```
smitpatel: Just a reminder to remove quirk since this is going to 2.1

ajcvickers:
Merged to 2.1 without quirk.

Issue #749, opened 2014-10-01
Fully implement entire entities in shadow state


ajcvickers:
That is, entities not backed by any CLR type. Also see #748.

rjperes:
May I ask what is the gain of this? Does this feature exist in any ORM? I know this is going to be used for M:M mapping in EF Core, but other that that, I honestly can't see the point. Thanks! :-)

divega:
@rjperes In general we use shadow sate to reduce the coupling between the domain model the application reasons about and the extra information EF Core might need to track and reason about. This allows for more flexibility on the shape of the types while maintaining how EF Core reasons about the model at runtime relatively simple and based on a relatively small number of different concepts. 

For individual properties in shadow state one of the main examples is how EF Core always reasons about relationships between objects in terms of foreign key properties, but CLR types that participate in those relationships don't need to contain the FKs. 

For full entities in shadow state, the main two potential scenarios to come to mind are:

1. The link entity in many-to-many associations (which you already mentioned)
2. Extensibility scenarios in which collections of new data can be added to an entity (e.g. by a user) without having to define a new CLR type to store the new data.

In both cases not having a type in the domain model representing this data is developer choice.

BradBarnich:
for my use case, I have audit tables that are maintained by trigger. I wanted the tables to be part of migrations so I did this:

```csharp
foreach ( var entityType in modelBuilder.Model.GetEntityTypes().ToArray() )
{
    var auditEntity = modelBuilder.Entity( entityType.Relational().TableName + "History" );
    foreach ( var property in entityType.GetProperties() )
    {
        auditEntity.Property( property.ClrType, property.Name );
    }
    auditEntity.Property( typeof( DateTime ), "ValidTo" );
    auditEntity.HasKey( "ID", "ValidTo" );
}
```

Also had to make a custom model validator that skips `EnsureNoShadowEntities`, but then it works.

AndriySvyryd:
We should completely drop support for shadow state entities in favor of property bag entities #9914 as they are easier to work with as a customer and would require less code changes to implement.

ajcvickers:
Closing as this particular idea (entities without any CLR object) may now not be needed since something like property bags (or even object entities with shadow properties) would cover the cases where this was thought to be useful.

lakeman:
It's a little silly that I need to define something like `internal class Dummy<T>{}` which I only intend to use for audit table migrations like the above example.

[...]

PR #9772, opened 2017-09-12
Add SeedData

AndriySvyryd:
Fixes https://github.com/aspnet/EntityFrameworkCore/issues/629

Further improvements will be done in separate PRs


smitpatel:
Rebase needed.

AndriySvyryd:
I'll rebase when the PR is approved. Incremental rebases are time-consuming as I cannot rebase the original commit.

src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -714,7 +745,11 @@ public virtual int SaveChanges(bool acceptAllChangesOnSuccess)
             }
         }
 
-        private List<InternalEntityEntry> GetEntriesToSave()
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public virtual IReadOnlyList<InternalEntityEntry> GetEntriesToSave()
```
ajcvickers: >public virtual IReadOnlyList<InternalEntityEntry> GetEntriesToSave() [](start = 8, length = 68)

Should this really be exposed publicly? What ends up using the InternalEntityEntries? Is it provider code at all?

AndriySvyryd: `IStateManager` is an Internal service

src/EFCore/ChangeTracking/Internal/StateManager.cs
```diff
@@ -157,6 +157,36 @@ public virtual InternalEntityEntry GetOrCreateEntry(object entity)
...
+                    entry = new InternalShadowEntityEntry(this, entityType);
+                    entity = entry;
+                }
+
+                _entityReferenceMap[entity] = entry;
+            }
+
+            entry.ToEntityEntry().CurrentValues.SetValues(values);
+            return entry;
+        }
```
ajcvickers: This method seems like it needs some work. There is duplicated logic here with other GetOrCreate methods, and this could make better use of the materializer.

src/EFCore/ChangeTracking/Internal/KeyPropagator.cs
```diff
@@ -100,8 +100,11 @@ private static InternalEntityEntry TryPropagateValue(InternalEntityEntry entry,
                 {
                     if (property == foreignKey.Properties[propertyIndex])
                     {
+                        // Remove when issue #749 is fixed
+                        var principal = foreignKey.DependentToPrincipal?.IsShadowProperty ?? true
+                            ? null
+                            : foreignKey.DependentToPrincipal.GetGetter().GetClrValue(entry.Entity);
                         InternalEntityEntry principalEntry = null;
```
ajcvickers: What's the consequence of just skipping propagation here?

src/EFCore/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -679,9 +679,11 @@ public virtual void AddRangeToCollectionSnapshot([NotNull] IPropertyBase propert
         {
             get
             {
-                return _storeGeneratedValues.TryGetValue(propertyBase, out var value)
-                    ? value
-                    : ReadPropertyValue(propertyBase);
+                return propertyBase is INavigation && propertyBase.IsShadowProperty // Remove when issue #749 is fixed
+                    ? null
+                    : _storeGeneratedValues.TryGetValue(propertyBase, out var value)
+                        ? value
```
ajcvickers: I'm concerned in general about the hacks for shadow nav props--for example, how they may influence other code paths rather than just seeding. Do we have a plan to either fix #749 in 2.1, or make the hacks less hacky?

AndriySvyryd: I don't think that these hacks have any functional impact on non-seeding scenarios. If we have time to do #749 in 2.1 I would prefer it instead of trying to improve the hacks further.

src/EFCore.Relational/breakingchanges.netcore.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    {
+      "TypeId": "public class Microsoft.EntityFrameworkCore.Migrations.Operations.InsertDataOperation : Microsoft.EntityFrameworkCore.Migrations.Operations.MigrationOperation",
+      "Kind": "Removal"
+    },
+    {
+      "TypeId": "public class Microsoft.EntityFrameworkCore.Migrations.Operations.UpdateDataOperation : Microsoft.EntityFrameworkCore.Migrations.Operations.MigrationOperation",
+      "Kind": "Removal"
+    },
+    {
```
ajcvickers: These are here because we added a base class? So, not really breaks?

AndriySvyryd: Yeah, seems to be a bug in the API tool

src/EFCore.Relational/Update/ColumnModification.cs
```diff
@@ -127,15 +163,26 @@ public virtual string OriginalParameterName
         /// <summary>
         ///     The original value of the property mapped to this column.
         /// </summary>
-        public virtual object OriginalValue => Entry.GetOriginalValue(Property);
+        public virtual object OriginalValue => Entry == null ? _originalValue : Entry.GetOriginalValue(Property);
```
ajcvickers: >Entry == null [](start = 47, length = 13)

What is it that allows Entry to be null now?

AndriySvyryd: The new constructor.
This is needed because when generating SQL when applying a migration we might not have an `IModel` anymore

src/EFCore/ChangeTracking/Internal/StateManager.cs
ajcvickers: I was just concerned that it might be consumed by code that could be overridden by a provider--the same reason we introduced IUpdateEntry for the update pipeline. Looks like this isn't the case, so no problem. However, did we consider returning IUpdateEntrys instead?

AndriySvyryd: I'll change it to return `IReadOnlyList<IUpdateEntry>`

AndriySvyryd:
I've added a todo list to https://github.com/aspnet/EntityFrameworkCore/issues/629 to track the outsanding issues roughly in decreasing priority order. Let me know if I missed anything.

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -310,6 +349,60 @@ public virtual IReadOnlyList<MigrationOperation> GetDifferences(IModel source, I
...
+                    }
+                    else
+                    {
+                        yield return new DeleteDataOperation
+                        {
+                            Schema = c.Schema,
+                            Table = c.TableName,
+                            KeyColumns = c.ColumnModifications.Select(col => col.ColumnName).ToArray(),
+                            KeyValues = ToMultidimensionalArray(c.ColumnModifications.Select(col => col.Value).ToArray())
+                        };
```
bricelam: We should look into accumulating these. Instead of creating one per row, if the last operation was the same type for the same table, we should just append the values.

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -1157,6 +1255,103 @@ protected virtual IEnumerable<MigrationOperation> Remove([NotNull] ISequence sou
...
+            [NotNull] DiffContext diffContext)
+        {
+            Check.NotNull(source, nameof(source));
+            Check.NotNull(target, nameof(target));
+            Check.NotNull(diffContext, nameof(diffContext));
+
+            if (source.EntityTypes.Count != 1 || target.EntityTypes.Count != 1)
+            {
+                // Table splitting not supported
+                return;
```
bricelam: Should it throw or is is made impossible at a higher level?

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -87,8 +102,24 @@ public class MigrationsModelDiffer : IMigrationsModelDiffer
...
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
         public virtual bool HasDifferences(IModel source, IModel target)
-            => Diff(source, target, new DiffContext(source, target)).Any();
+        {
+            var result = Diff(source, target, new DiffContext(source, target)).Any();
+            StateManager.ResetState();
```
bricelam: Could you hit issues if you call `HasDifferences` during `SaveChanges`? This seems like a dangerous side affect...

src/EFCore.Relational/Migrations/Operations/ModificationOperation.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// </summary>
+    public abstract class ModificationOperation : MigrationOperation
+    {
+        private IEnumerable<ModificationCommand> _modificationCommands;
+
+        /// <summary>
+        ///     The commands that correspond to this operation.
+        /// </summary>
+        public virtual IEnumerable<ModificationCommand> ModificationCommands
+            => _modificationCommands ?? (_modificationCommands = GenerateModificationCommands());
```
bricelam: This doesn't feel a property to me. Lets discuss during API review.

src/EFCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
```diff
@@ -1066,10 +1066,9 @@ protected override void Generate(SqlOperation operation, IModel model, Migration
                     .AppendLine(Dependencies.SqlGenerationHelper.StatementTerminator);
             }
 
-            base.Generate(operation, model, builder, terminate: false);
+            base.Generate(operation, model, builder);
 
             builder
-                .AppendLine(Dependencies.SqlGenerationHelper.StatementTerminator)
```
bricelam: So the last statement is in a separate batch now?

test/EFCore.Relational.Tests/Migrations/Internal/MigrationsModelDifferTest.cs
```diff
@@ -587,21 +728,95 @@ public void Drop_column()
...
+                        x.Property<string>("Name").HasColumnType("nvarchar(30)");
+                    }),
+                target => target.Entity(
+                    "Firefly",
+                    x =>
+                    {
+                        x.ToTable("Firefly", "dbo");
+                        x.Property<int>("Id");
+                        x.HasKey("Id");
+                    }),
```
bricelam: This looks like a bad merge.

test/EFCore.Relational.Tests/Migrations/Internal/MigrationsModelDifferTest.cs
```diff
@@ -575,6 +650,72 @@ public void Add_column_not_null(Type type, object expectedDefault)
...
+        {
+            Execute(
+                _ => { },
+                source => source.Entity(
+                    "Firefly",
+                    x =>
+                    {
+                        x.ToTable("Firefly", "dbo");
+                        x.Property<int>("Id");
+                        x.HasKey("Id");
```
bricelam: Unnecessary `HasKey()` calls were removed from these tests.

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
AndriySvyryd: You mean concurrently? We definitely don't support that.

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
AndriySvyryd: Yes, it's on the ToDo list: https://github.com/aspnet/EntityFrameworkCore/issues/629

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
AndriySvyryd: It's on the ToDo list: #629

src/EFCore.Relational/Migrations/Operations/ModificationOperation.cs
AndriySvyryd: No need to discuss, I'll just make it a method

test/EFCore.Relational.Tests/Migrations/Internal/MigrationsModelDifferTest.cs
AndriySvyryd: Looks more like a bad PR differ

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
bricelam: I imagined something like this:
``` csharp
db.Add(customer);

if (differ.HasDiferences(modelSnapshot, db.Model)
    throw new InvalidOperationException("Someone forgot to migrate.");

db.SaveChanges();
```

AndriySvyryd: Still doesn't look like an important enough scenario, but I am open to hearing opinions from others

src/EFCore.SqlServer/Migrations/SqlServerMigrationsSqlGenerator.cs
AndriySvyryd: Appears so. I'll revert this when I add batching back

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
bricelam: cc @divega @ajcvickers

test/EFCore.Relational.Tests/Migrations/Internal/MigrationsModelDifferTest.cs
bricelam: No, the old code didn't have `HasKey()` and used chaining (not a closure) on the target model.

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
divega: Coming very late to this comment, but I need more context. It seems it would be good to discuss it when @bricelam is present. @AndriySvyryd can you please file something so that we remember?

AndriySvyryd: I already fixed this in the new PR that's currently open

PR #9780, opened 2017-09-12
Adds quirk for #9128

anpete:


smitpatel:
This would still imply that changes in TaskBlockingExpressionVisitor would not be quirked. Is that expected?

anpete:
@smitpatel Good catch, will update.

anpete:
Take 2: I realized that we wanted to not make that change to TaskBlockingExpressionVisitor at all. I tweaked the fix so that Task Lifting can also detect TaskBlockingExpressionVisitor.

cc @ajcvickers

Issue #9309, opened 2017-07-31
Query: Output parameter in FromSql/ExecuteSqlCommand is always returns null


smitpatel:
This issue is to track porting of #9277 for patch.

Impact: Using `FromSql` with out parameters on SqlServer will never return value of out parameter. All such scenarios are broken.
Work-around:None

smitpatel:
Impact: Using FromSql with out parameters on SqlServer will never return value of out parameter. All such scenarios are broken.
Risk: Low. The fix was to clear the parameters after reader is disposed. Hence the code to clear parameter collection was moved to appropriate place and it is still part of pipeline in all scenarios.

Eilon:
This patch bug is **approved** for the 2.0.x patch. Please send a PR to the `feature/2.0.1` branch and get it reviewed and merged. When we have the `rel/2.0.1` branches ready please port the commit to that branch.

Eilon:
Hi, we have a public test feed that you can use to try out the ASP.NET/EF Core 2.0.3 patch!

To try out the pre-release patch, please refer to the following guide:
* https://github.com/aspnet/Announcements/issues/274 (info)
* https://github.com/aspnet/Home/issues/2254 (discussions and questions)

We are looking for feedback on this patch. We'd like to know if you have any issues with this patch by updating your apps and libraries to the latest packages and seeing if it fixes the issues you've had, or if it introduces any new issues. If you have any issues or questions, please reply on this issue to let us know as soon as possible.

Thanks,
Eilon

Issue #9277, opened 2017-07-27
Get output parameter value of a stored procedure using EF Core FromSql(...) is always null. 


irowbin:
Describe what is not working as expected.

Store procedure output parameter is not working as expected. The output value is always null.
I have an statement like this. I've removed other parameters except this `OUTPUT` param to pointing out the issue. The records are came but the the value of this `OUTPUT` parameter is null without the exception. Executing the proc in SSMS works just fine but with this `FromSql(...)` isn't doing anything for output parameter.

```sql
-- other param are working as expected but the OUTPUT is not so i hide them.

@todoCounts VARCHAR(MAX) OUTPUT
```
```c#
var items = _context.Set<T>().FromSql("Sp_Todo @todoCounts OUTPUT", 
 new SqlParameter("@todoCounts", SqlDbType.VarChar)
 { Direction = ParameterDirection.Output}).ToList() // no luck
```
If i add this too, then the output value coming but this is not the correct idea here. I am not seeing any exceptions. 
```c#
  _context.Database.ExecuteSqlCommand("Sp_Todo @todoCounts OUTPUT", 
 new SqlParameter("@todoCounts", SqlDbType.VarChar)
 { Direction = ParameterDirection.Output});
```

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
none
```

### Steps to reproduce

- Create a store procedure with output parameter.

```sql
 @todoCounts VARCHAR(MAX) OUTPUT
```

- Execute the StoreProc like this.

```c#
// records retrieved except OUTPUT parameter. 
context.set<MyModel>().FromSql("Sp_Todo @todoCounts OUTPUT", 
 new SqlParameter("@todoCounts", SqlDbType.VarChar)
 { Direction = ParameterDirection.Output}).ToList() 

```

### Further technical details
EF Core version: `1.1.2`
Database Provider: `1.1.2`
Operating system: Windows 10
IDE:  Visual Studio 2017 `15.2(2643.16) release`

divega:
We need to investigate if this is working on 2.0. We will consider patching this depending on the impact and the fix.

smitpatel:
Since fix for this is out in PR, removing milestone so we can discuss about patch.

mguinness:
Just to clarify that this issue is in relation to SQL Server.  The MySQL database provider has the same issue, but there is a [known issue](https://bugs.mysql.com/bug.php?id=75267) which prevents output parameters from being used with Connector/NET.

[...]

PR #9781, opened 2017-09-12
[2.0.1] Clear DbCommand.Parameters only after reader is disposed

smitpatel:
Porting #9277 as per #9309

### DIFF
--- a/src/EFCore.Relational/Storage/Internal/RelationalCommand.cs
+++ b/src/EFCore.Relational/Storage/Internal/RelationalCommand.cs
@@ -225,7 +225,12 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
             }
             finally
             {
-                dbCommand.Parameters.Clear();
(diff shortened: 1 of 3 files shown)


src/EFCore.Relational/Storage/Internal/RelationalCommand.cs
```diff
@@ -225,10 +225,9 @@ public class RelationalCommand : IRelationalCommand
             }
             finally
             {
-                dbCommand.Parameters.Clear();
```
ajcvickers: Quirk?

smitpatel: Did we decide to quirk this? I couldn't find note anywhere.

ajcvickers: We're quirking everything by default, only not quirking if there is a compelling reason not to.

smitpatel:
@ajcvickers - updated.

Issue #9627, opened 2017-08-30
Re-consider how default values are reverse engineered


ajcvickers:
Feedback on 2.0 (e.g. #9507 #9617) has made me think about what the real intent is behind having columns with default values. Specifically, I think it is likely wrong to assume that the EF model should have a store-generated property just because there is some default on the column. It may well be the cause the the application will always supply a value and so, when using EF, the default is never used. The default may instead be there because:
* It was needed when adding a new column (either by Migrations or manually)
* It is there for legacy reasons--old/non-EF app made use of it in hand-crafted SQL, but EF never will
* It's just part of the way all columns are created by the DBA or similar--the column gets a default regardless of whether it is needed or not

Going beyond the most general case, if the CLR default would generate the same value as the database default, then having EF do store generation is a no-op anyway. If an app was relying on this for bools in 1.1, and we now generate nullable bools, everything will still work, but the app needs to be updated to use nullable bools.

ajcvickers:
Triage: we will update reverse engineering to not make the property store-generated if the default value from the column matches the default value of the CLR property.

JuanIrigoyen:
This change break with the model, if you need use the default value in a entity, the logical is set these in the constructor of the class or in the property how default value. The problem is when the database modify the record after insert with a trigger, calculate fields, another procedure or function of transact-Sql that EF don´t know. For example if your default value is a function in sql server how `default([dbo].[NEWDOC_Order]())` EF never to know this. In this case the default value will be lost and you always must to read the records inserted before return it to EF, but if EF know the function, numeric values, getdate() or other basic functions these can be set in the constructor or in the properties if I have a bool field and in my model the field refer to bool?, the validations an other functions are different. In numeric class If I make a sum, I would be forced to change the function to not read the null values. If I have a default value in a table with not null in my model must be the same because I must to change my validations and programs for this cause. I´m sure that this behavior is not correct.  

For me the use of default values ​​should be something like the following:

```
public partial class Customer
{
    public Customer()
    {
        Date = DateTime.Now;    // Default Value "Getdate()"
        Total = 100;	        // Default Value 100
    }

    public string CustomerId { get; set; }
    public string Name { get; set; }

    [DefaultValue("Getdate()")]
    public DateTime Date { get; set; }

    [DefaultValue(100)]
    public decimal Total { get; set; }
}
```

// Or using default values in properties

```
public partial class Customer
{
    public Customer()
    {

    }

[...]

PR #9782, opened 2017-09-12
[2.0.1] RevEng: Don't generate nullable bool if default value is false

smitpatel:
As per #9627

### DIFF
--- a/src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
+++ b/src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
@@ -441,6 +441,15 @@ ORDER BY schema_name(t.schema_id), t.name, c.column_id";
                         defaultValue = null;
                     }
 
+                    if (!AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Issue9627", out var isEnabled) || !isEnabled)
+                    {
+                        if (defaultValue == "((0))"
+                            && (underlyingStoreType ?? storeType) == "bit")
+                        {
+                            defaultValue = null;
+                        }
+                    }
+
                     if (computedValue == "(NULL)")
                     {
                         computedValue = null;
--- a/test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/Expected/NonNullBoolWithDefault.cs
+++ b/test/EFCore.SqlServer.Design.FunctionalTests/ReverseEngineering/Expected/NonNullBoolWithDefault.cs
@@ -7,6 +7,11 @@ namespace E2ETest.Namespace
     {
         public int Id { get; set; }
         public bool? BoolWithDefaultValueSql { get; set; }
+        public bool BoolWithDefaultValueSqlFalse1 { get; set; }
+        public bool BoolWithDefaultValueSqlFalse2 { get; set; }
+        public bool BoolWithDefaultValueSqlFalse3 { get; set; }
+        public bool BoolWithDefaultValueSqlFalse4 { get; set; }
+        public bool BoolWithDefaultValueSqlFalse5 { get; set; }
         public bool BoolWithoutDefaultValueSql { get; set; }
     }
 }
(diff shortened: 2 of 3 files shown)


src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
```diff
@@ -441,6 +441,12 @@ ELSE null
                         defaultValue = null;
                     }
 
+                    if (defaultValue == "((0))"
```
ajcvickers: Will it always be "((0))" or sometimes "(0)" or "( ( 0 )) ", etc?

ajcvickers: Also, quirk?

smitpatel: I tried with different amount of `()` all of those goes to `((0))`. I will add with spaces.

smitpatel:
@ajcvickers - updated.

src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
smitpatel: Since the column type is bit whichever it is specified, SqlServer will make it `((0))`

src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
```diff
@@ -441,6 +441,15 @@ ELSE null
                         defaultValue = null;
                     }
 
+                    if (!AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Issue9627", out var isEnabled) || !isEnabled)
+                    {
+                        if (defaultValue == "((0))"
+                            && (underlyingStoreType ?? storeType) == "bit")
```
bricelam: Should this be a case-insensitive comparison?

smitpatel: I will check what happens with different casing. Though most likely it would be fixed case since whatever casing user defines storetype, SqlServer combines them into one (like nationalchar/nchar going to nchar)

ErikEJ: YAGNI

bricelam: I hope SQL Server normalizes, but I also know that SQLite doesn't. 😉

smitpatel: SqlServer does normalize it.

RandyBuchholz:
I'm reading through the discussions and I'm not convinced about default true becoming nullable. I have many tables that have an "IsActive" flag. When I create a record I want it active. It is never going to be null. Nullable bools are the aberration, not the norm. A "true" bool is only true or false. I have the need for nullable bools on rare occasions, but to default to the edge case doesn't make sense. And isn't a nullable bool a more complex type? It's not a bool is it? It's a nullable of type bool.

Worse, this is breaking bi-directional symmetry. If I create a database with a mandatory bool and default True, it comes over as nullable. If I code first a database form that, the type will now be nullable, correct? 

Also, one of the reasons to put defaults on things is to prevent any null situations.

This logic is saying that True is somehow different than False. That's a lot of baggage for a developer to carry around in their head - "When I use EF I have to remember that True is not the opposite of False, Nullable True is the opposite of False." ( not really opposite, but different) 

- K, this is a little ranty, but I hope it makes my point - Why is EF changing my design?

src/EFCore.SqlServer/Scaffolding/Internal/SqlServerDatabaseModelFactory.cs
HolisticDeveloper: > Will it always be "((0))" or sometimes "(0)" or "( ( 0 )) ", etc?

I have just run into this today because I have an old DB, originally authored with SQL Server 2000, where many bit columns have a default of "(0)". Defaults created after it was upgraded to SQL Server 2008 (we skipped 2005) use "((0))".

Issue #9524, opened 2017-08-22
Connection state not tracked correctly if store connection gets closed


AndriySvyryd:
1. Create a context with a closed existing connection or connection string
2. Call `context.Database.OpenConnection()`
3. Call `context.Database.GetDbConnection().Close()` or let it get closed due to a transient error
4. Call `context.SaveChanges()` or any other method that needs an open connection

At this point the connection will be closed, but should still be open.

PR #9792, opened 2017-09-13
Recover connection state counting after external close

ajcvickers:
Issue #9524

The issue here was the reference count was not kept correctly if the connection was closed externally when it was not expected to be. The fix is to keep the count correct in this case.

### DIFF
--- a/src/EFCore.Relational/Storage/RelationalConnection.cs
+++ b/src/EFCore.Relational/Storage/RelationalConnection.cs
@@ -241,7 +241,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
             var dbTransaction = DbConnection.BeginTransaction(isolationLevel);
 
             CurrentTransaction
-                = new RelationalTransaction(
+                = Dependencies.RelationalTransactionFactory.Create(
                     this,
                     dbTransaction,
                     Dependencies.TransactionLogger,
@@ -275,7 +275,7 @@ namespace Microsoft.EntityFrameworkCore.Storage
 
                 Open();
 
-                CurrentTransaction = new RelationalTransaction(
+                CurrentTransaction = Dependencies.RelationalTransactionFactory.Create(
                     this,
                     transaction,
                     Dependencies.TransactionLogger,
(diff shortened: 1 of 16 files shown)


src/EFCore.Relational/Storage/RelationalConnection.cs
```diff
@@ -339,10 +339,8 @@ public virtual bool Open(bool errorsExpected = false)
                 wasOpened = true;
                 ClearTransactions();
             }
-            else
-            {
-                _openedCount++;
-            }
+
+            _openedCount++;
```
AndriySvyryd: Interesting that the tests pass, but this won't work for connection resiliency. Count would be incremented on every retry.

ajcvickers: @AndriySvyryd I added a test using connection resiliency and fixed a test bug, but can't get product code to to fail. Probably going to need some help writing a test that demonstrates this.

AndriySvyryd: I found the issue in test code: `TestRelationalTransaction.ClearTransaction()` calls `_testConnection.Close()` for some reason, hiding the product bug.

ajcvickers: @AndriySvyryd I simplified TestRelationalTransaction and TestSqlServerConnection so that they don't have any logic other than causing failures...still can't reproduce the issue.

AndriySvyryd: I see, that was unrelated then. The code you added to `TestRelationalCommandBuilderFactory` hides a bug.
I've also added tests for the other places where an `ExecutionStrategy` is used: see `origin\MoreResiliencyTests` https://github.com/aspnet/EntityFrameworkCore/commit/ce4bf0d06a231680e8ceac901feed55b48796d4f

ajcvickers: New bits up with query fix.

PR #9793, opened 2017-09-13
Merge feature/2.0.1

### LINKED ISSUE #9627: Re-consider how default values are reverse engineered

smitpatel:
This fixes #9627 in dev

### DIFF
--- a/test/EFCore.SqlServer.FunctionalTests/Scaffolding/SqlServerDatabaseModelFactoryTest.cs
+++ b/test/EFCore.SqlServer.FunctionalTests/Scaffolding/SqlServerDatabaseModelFactoryTest.cs
@@ -1108,6 +1108,36 @@ CREATE TABLE DefaultComputedValues (
         }
 
         [Fact]
+        public void Default_value_of_false_for_bit_is_not_stored()
+        {
+            Test(
+                @"
+CREATE TABLE DefaultBoolValues (
+    Id int,
+    BoolTrue bit DEFAULT (1),
+    BoolFalseType1 BIT DEFAULT 0,
+    BoolFalseType2 BiT DEFAULT (0),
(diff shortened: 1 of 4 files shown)


test/EFCore.SqlServer.FunctionalTests/Scaffolding/SqlServerDatabaseModelFactoryTest.cs
```diff
@@ -1108,6 +1108,36 @@ public void Default_and_computed_values_are_stored()
...
 
         [Fact]
+        public void Default_value_of_false_for_bit_is_not_stored()
+        {
+            Test(
+                @"
+CREATE TABLE DefaultBoolValues (
+    Id int,
+    BoolTrue bit DEFAULT (1),
+    BoolFalseType1 BIT DEFAULT 0,
```
smitpatel: @bricelam - Added variation of `bit` too. all of them comes out as bit only.

PR #9807, opened 2017-09-13
Query: Include: Adds an optimized collection include client join predicate code path.

anpete:
- Avoids the use of IIncludeKeyComparer and ValueBuffers when neither of the entity types involved have any shadow key properties.
- Introduces breaking changes in IQueryBuffer

TBD: Update breaking changes log.

### DIFF
--- a/src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
+++ b/src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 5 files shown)


src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
```diff
@@ -121,19 +122,21 @@ var collectionQueryModel
...
-                        new SubQueryExpression(collectionQueryModel));
+                    = Expression.Lambda(new SubQueryExpression(collectionQueryModel));
 
                 var includeCollectionMethodInfo = _queryBufferIncludeCollectionMethodInfo;
 
                 Expression cancellationTokenExpression = null;
 
                 if (asyncQuery)
                 {
+                    var asyncEnumerableType
```
smitpatel: inline?

src/EFCore/Query/Internal/QueryBuffer.cs
```diff
@@ -1,5 +1,5 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+﻿// // Copyright (c) .NET Foundation. All rights reserved.
```
smitpatel: typo

src/EFCore/Query/Internal/IncludeCompiler.IncludeLoadTreeNode.cs
anpete: Split for clarity.

PR #9815, opened 2017-09-14
Ports updated fix for #9128 to dev.

anpete:
- Update TaskLiftingExpressionVisitor to detect blocking introduced by TaskBlockingExpressionVisitor. :trollface:
- Switched Include Compiler to use TaskBlockingExpressionVisitor instead of .Result.
- Reverted previous changes to TaskBlockingExpressionVisitor.

### DIFF
--- a/src/EFCore/Query/EntityQueryModelVisitor.cs
+++ b/src/EFCore/Query/EntityQueryModelVisitor.cs
@@ -27,6 +27,7 @@ using Remotion.Linq.Clauses.ExpressionVisitors;
 using Remotion.Linq.Clauses.ResultOperators;
 using Remotion.Linq.Clauses.StreamedData;
 using Remotion.Linq.Parsing;
+using Microsoft.EntityFrameworkCore.Extensions.Internal;
 
 namespace Microsoft.EntityFrameworkCore.Query
 {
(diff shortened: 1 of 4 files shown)


src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -540,16 +541,25 @@ var entityTrackingInfos
...
-                var newExpression = Visit(node.Expression);
+                if (IncludeCompiler.IsIncludeMethod(methodCallExpression))
+                {
+                    return methodCallExpression.Arguments[1];
+                }
+
+                if (methodCallExpression.Method
+                    .MethodIsClosedFormOf(TaskBlockingExpressionVisitor.ResultMethodInfo))
+                {
+                    var newArguments = VisitAndConvert(methodCallExpression.Arguments, "VisitMethodCall");
```
smitpatel: Interesting. Any reason we cannot use just Visit here?

anpete: Visit tries to put the method call back together, which will throw because the args are no longer compatible.

anpete:
Going to combine with #9832

PR #9834, opened 2017-09-15
[2.0.1] Update version.props to include suffix and set PackageId in csproj

natemcmaster:
Backports https://github.com/aspnet/EntityFrameworkCore/pull/9381 so repo-to-repo graph analysis works.

Set VersionSuffix on the packages produced from this branch to 2.0.1-rtm-$(BuildNumber)

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,6 +1,9 @@
-<!-- This file may be overwritten by automation. Only values allowed here are VersionPrefix and VersionSuffix.  -->
 <Project>
-    <PropertyGroup>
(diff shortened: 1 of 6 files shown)


version.props
```diff
@@ -2,5 +2,9 @@
 <Project>
     <PropertyGroup>
         <VersionPrefix>2.0.1</VersionPrefix>
+        <VersionSuffix>rtm</VersionSuffix>
```
smitpatel: Nice 😄

PR #9837, opened 2017-09-16
[2.0.1] Query: Avoid short-circuiting while using context closure variables in query filters

### LINKED ISSUE #9825: Access to context bound variable can be optimized out
Found investigating #9809. This should work:

```C#
modelBuilder
    .Entity<Report>()
    .HasQueryFilter(x => !x.IsDeleted && (IsModerated == null || IsModerated == x.IsModerated));
```
but the IsModerated == null part is optimized away.

smitpatel:
Issue:
We adopted short-circuiting logic in logical expression due to possible exceptions (which compiler would never throw in memory). This works correctly for general cases because the expression tree would have current value of closure variable hence evaluating to correct thing.
When it comes to QueryFilter, the filter has closure member access on context instance which was used during `OnModelCreating` which may have stale value of context closure variables. We parametrize such closure variables and insert their values through parameters when running query. But if such closure variable is used in complex expression which can short-circuit then we would be using wrong value for short-circuit (by this point we haven't inserted correct value from current context instance) hence generating incorrect query model.

Solution:
While evaluating filters, we need to make sure that no complex expression which has context closure variable is partially evaluatable (so it doesn't short-circuit). And when generating filters, for non-queryable member access on context (closure variables but not DbSet), we need to parametrize it.

Resolves #9825

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/FilterApplyingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/FilterApplyingExpressionVisitor.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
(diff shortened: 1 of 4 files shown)


anpete:
#9839 has an alternative approach, we should decide what to do for 2.0.1 and what to do for dev.

natemcmaster:
Can you retarget this to the rel/2.0.1 branch?

smitpatel:
Closing in #9846

PR #9838, opened 2017-09-16
Decouple StateManager from DbContext as much as possible by specifying the dependencies explicitly.

AndriySvyryd:


src/EFCore/ChangeTracking/ChangeTracker.cs
```diff
@@ -160,7 +174,13 @@ public virtual bool HasChanges()
         ///     <see cref="DbContext.SaveChanges()" /> and when returning change tracking information). You typically only need to
         ///     call this method if you have disabled <see cref="AutoDetectChangesEnabled" />.
         /// </summary>
-        public virtual void DetectChanges() => ChangeDetector.DetectChanges(StateManager);
+        public virtual void DetectChanges()
+        {
+            if (_model[Internal.ChangeDetector.SkipDetectChangesAnnotation] == null)
```
ajcvickers: Typically, if you explicitly call DetectChanges, then we always do it regardless of what settings there might be. Are we 100% sure this is a no-op?

AndriySvyryd: I just lifted the check from `ChangeDetector.DetectChanges()` we could lift it further to get the behavior that you are describing.

ajcvickers: I thought about it more and I think it is fine.

PR #9839, opened 2017-09-16
Fix #9825 - Access to context bound variable can be optimized out

anpete:
Replaced short-circuiting optimization in ParameterExtractingExpressionVisitor with cheap null compensation handling. The main advantages to this approach are:  

1) No more query cache fragmentation - short circuiting would produce multiple entries in query cache. 
2) Simpler: Less logic/rewriting inside parameter extraction.

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
@@ -294,60 +294,6 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
             return newExpression;
         }
 
-        /// <summary>
-        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
-        ///     directly from your code. This API may change or be removed in future releases.
-        /// </summary>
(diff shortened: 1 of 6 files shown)


anpete:
#9837 has an alternative approach, we should decide what to do for 2.0.1 and what to do for dev.

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -477,6 +423,14 @@ public virtual object Evaluate([CanBeNull] Expression expression, out string par
                     .Compile()
                     .Invoke();
             }
+            catch (NullReferenceException)    // null compensation
```
ajcvickers: Is the assumption here that if evaluation null-refs, then the value of the parameter must be null?

ajcvickers:
@smitpatel @anpete If you're both going to be in the office, then let's discuss to decide which way to go here.

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
anpete: Yes.

PR #9840, opened 2017-09-16
Restore GetAwaiter behavior of TaskBlockingExpressionVisitor.

anpete:
And corresponding changes to task lifting and Include compiler.

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/TaskLiftingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/TaskLiftingExpressionVisitor.cs
@@ -8,6 +8,7 @@ using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Remotion.Linq.Parsing;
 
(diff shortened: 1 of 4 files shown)


src/EFCore/Query/ExpressionVisitors/Internal/TaskLiftingExpressionVisitor.cs
```diff
@@ -150,6 +151,38 @@ protected override Expression VisitMember(MemberExpression memberExpression)
...
+                _taskExpressions.Add(
+                    Expression.Lambda<Func<Task<object>>>(
+                        Expression.Call(
+                            _toObjectTask.MakeGenericMethod(
+                                methodCallExpression.Method.ReturnType),
+                            methodCallExpression.Arguments[0])));
+
+                if (CancellationTokenParameter == null)
+                {
+                    Visit(methodCallExpression.Arguments[0]);
```
smitpatel: Why do we need to visit only in some cases?

smitpatel:
No tests?

anpete:
@smitpatel Used heavily by existing tests.

src/EFCore/Query/ExpressionVisitors/Internal/TaskLiftingExpressionVisitor.cs
anpete: Existing code but I think we use it to signal when to stop.

PR #9846, opened 2017-09-19
[2.0.1] Query: Fix for Access to context bound variable can be optimized out

### LINKED ISSUE #9825: Access to context bound variable can be optimized out

smitpatel:
**Take 3**

Issue:
We adopted short-circuiting logic in logical expression due to possible exceptions (which compiler would never throw in memory). This works correctly for general cases because the expression tree would have current value of closure variable hence evaluating to correct thing.
When it comes to QueryFilter, the filter has closure member access on context instance which was used during OnModelCreating which may have stale value of context closure variables. We parametrize such closure variables and insert their values through parameters when running query. But if such closure variable is used in complex expression which can short-circuit then we would be using wrong value for short-circuit (by this point we haven't inserted correct value from current context instance) hence generating incorrect query model. The other cause for it is QueryCache has only 1 entry for all the possible values of context bound variables in filter.

Solution:
Since evaluating filter with context bound variables with different value may not always generate a re-usable query model for all values of variables, whenever we get an expression to evaluate during funcletizing query filter, we replace context, with context parameter (which would be injected later) and return this expression without evaluating anything in-memory. Any exception to be thrown from evaluation of context bound variables will be thrown when actual parameter value will be calculated while running query.

Resolves #9825

### DIFF
--- a/src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
+++ b/src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
@@ -20,6 +20,7 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
     public class ParameterExtractingExpressionVisitor : ExpressionVisitor
     {
         private static readonly TypeInfo _queryableTypeInfo = typeof(IQueryable).GetTypeInfo();
+        private static ContextParameterReplacingExpressionVisitor _contextParameterReplacingExpressionVisitor;
 
         private readonly IEvaluatableExpressionFilter _evaluatableExpressionFilter;
         private readonly IParameterValues _parameterValues;
@@ -31,6 +32,8 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
         private PartialEvaluationInfo _partialEvaluationInfo;
 
         private bool _inLambda;
+
+        private static bool QuirkEnabled => AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Issue9825", out var enabled) && enabled;
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
@@ -48,6 +51,12 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
             _logger = logger;
             _parameterize = parameterize;
             _generateContextAccessors = generateContextAccessors;
+
+            if (_generateContextAccessors
+                && !QuirkEnabled)
+            {
+                _contextParameterReplacingExpressionVisitor = new ContextParameterReplacingExpressionVisitor();
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


smitpatel:
Quirking is pending. If the implementation is acceptable I will add quirks.

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -416,6 +417,23 @@ public virtual object Evaluate([CanBeNull] Expression expression, [CanBeNull] ou
...
 
+            if (_generateContextAccessors)
+            {
+                var contextParameterReplacingExpressionVisitor = new ContextParameterReplacingExpressionVisitor();
+                var updatedExpression = contextParameterReplacingExpressionVisitor.Visit(expression);
+
+                if (contextParameterReplacingExpressionVisitor.ContextParameterExpression != null)
+                {
+                    parameterName = updatedExpression is MemberExpression memberExpression
+                        ? memberExpression.Member.Name
```
anpete: How is this name used? Can it collide when there are multiple filters?

natemcmaster:
Can you retarget this to the rel/2.0.1 branch before merging?

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
smitpatel: #9850

smitpatel:
Updated with quirks.

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -32,6 +32,8 @@ public class ParameterExtractingExpressionVisitor : ExpressionVisitor
 
         private bool _inLambda;
 
+        private static bool QuirkEnabled => AppContext.TryGetSwitch("Microsoft.EntityFrameworkCore.Issue9825", out var enabled) && enabled;
```
anpete: private fields start with '_' - You can update R# naming settings.

anpete: Field?

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -48,6 +51,11 @@ public class ParameterExtractingExpressionVisitor : ExpressionVisitor
             _logger = logger;
             _parameterize = parameterize;
             _generateContextAccessors = generateContextAccessors;
+
+            if (_generateContextAccessors)
+            {
```
anpete: && field is null

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -403,6 +429,22 @@ var compilerPrefixIndex
             return Expression.Parameter(expression.Type, parameterName);
         }
 
+        private class ContextParameterReplacingExpressionVisitor : ExpressionVisitor
```
anpete: sealed

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -403,6 +429,22 @@ var compilerPrefixIndex
...
+        {
+            public ParameterExpression ContextParameterExpression;
+
+            protected override Expression VisitConstant(ConstantExpression node)
```
anpete: node => constantExpression :trollface:

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -403,6 +429,22 @@ var compilerPrefixIndex
...
+            {
+                if (typeof(DbContext).GetTypeInfo()
+                    .IsAssignableFrom(node.Type.GetTypeInfo()))
+                {
+                    return ContextParameterExpression ?? (ContextParameterExpression = Expression.Parameter(node.Type, "context"));
+                }
+
+                return base.VisitConstant(node);
```
anpete: nit: can just return node.

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
```diff
@@ -416,6 +458,25 @@ public virtual object Evaluate([CanBeNull] Expression expression, [CanBeNull] ou
                 return null;
             }
 
+            if (!QuirkEnabled)
+            {
+                if (_generateContextAccessors)
+                {
+                    var updatedExpression = _contextParameterReplacingExpressionVisitor.Visit(expression);
```
anpete: nit: we usually call this 'newExpression'

src/EFCore/Query/ExpressionVisitors/Internal/ParameterExtractingExpressionVisitor.cs
smitpatel: the use of out variable blocks it.

divega:
> We adopted short-circuiting logic in logical expression due to possible exceptions (which compiler would never throw in memory)

@smitpatel Do we handle query filters with patterns that would require short-circuiting after the fix?

divega:
I guess `MyContext9825` covers that. Thanks.

smitpatel:
@divega - It actually may not short-circuit but that behavior is somewhat hard to define because of having only 1 entry in query cache for all values of filters. We don't have ability to generate different select expression based on values. It would short circuit in generated SQL but it needs to be translatable and values could be computable.

rainerllera:
When will this be available?

PR #9851, opened 2017-09-19
Add CommandBatchPreparerDependencies and StateManagerDependencies

AndriySvyryd:
These will allow to create `CommandBatchPreparer` and `StateManager` instances on demand.

### DIFF
--- a/src/EFCore.Relational/Update/Internal/CommandBatchPreparerDependencies.cs
+++ b/src/EFCore.Relational/Update/Internal/CommandBatchPreparerDependencies.cs
@@ -1,0 +1,185 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.EntityFrameworkCore.Storage;
+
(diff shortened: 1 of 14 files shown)


src/EFCore.Relational/Update/Internal/CommandBatchPreparerDependencies.cs
```diff
@@ -0,0 +1,185 @@
...
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public ILoggingOptions LoggingOptions { get; }
+
+        /// <summary>
+        ///     Clones this dependency parameter object with one service replaced.
+        /// </summary>
+        /// <param name="modificationCommandBatchFactory"> A replacement for the current dependency of this type. </param>
+        /// <returns> A new parameter object with the given service replaced. </returns>
+        public CommandBatchPreparerDependencies With([NotNull] IModificationCommandBatchFactory modificationCommandBatchFactory)
```
Meir017: any reason not to make this method an extension method? (and the other `With` methods)

AndriySvyryd: We don't use extension methods unless there's a good reason to. This improves cohesion.

Meir017: These methods seem to me like utility methods and therefore they should not be inside the class, they don't use any private members or anything like that

AndriySvyryd: The whole purpose of this class is to hold and manipulate dependencies. The `With` methods are part of its responsibilities and contract.

PR #9862, opened 2017-09-20
[2.0.1] - Fix #9551 - Expression of type 'System.Object' cannot be used for parameter of type 'Microsoft.EntityFrameworkCore.Metadata.IEntityType'

### LINKED ISSUE #9551: Expression of type 'System.Object' cannot be used for parameter of type 'Microsoft.EntityFrameworkCore.Metadata.IEntityType'

anpete:
- Modify QuerySourceTracingExpressionVisitor so that it can return the "deepest" GroupBy QueryModel if one exists.
- Update the Include compiler to target the correct QueryModel when GroupBy is in play.

### DIFF
--- a/src/EFCore.Specification.Tests/Query/IncludeTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/IncludeTestBase.cs
@@ -3132,6 +3132,558 @@ namespace Microsoft.EntityFrameworkCore.Query
             }
         }
 
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public virtual void Include_collection_GroupBy_Select(bool useString)
+        {
+            using (var context = CreateContext())
+            {
+                var orders
+                    = useString
+                        ? context.Orders
+                            .Where(o => o.OrderID == 10248)
+                            .Include("OrderDetails")
+                            .GroupBy(e => e.OrderID)
+                            .Select(e => e.OrderBy(o => o.OrderID).FirstOrDefault())
+                            .ToList()
+                        : context.Orders
+                            .Where(o => o.OrderID == 10248)
+                            .Include(o => o.OrderDetails)
+                            .GroupBy(e => e.OrderID)
+                            .Select(e => e.OrderBy(o => o.OrderID).FirstOrDefault())
+                            .ToList();
+
+                foreach (var order in orders)
+                {
+                    CheckIsLoaded(
+                        context,
+                        order,
+                        orderDetailsLoaded: true,
+                        productLoaded: false,
+                        customerLoaded: false,
+                        ordersLoaded: false);
+                }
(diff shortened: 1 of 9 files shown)


src/EFCore.Specification.Tests/Query/IncludeTestBase.cs
```diff
@@ -3132,6 +3132,558 @@ public virtual void Include_specified_on_non_entity_not_supported(bool useString
...
+                        customerLoaded: false,
+                        ordersLoaded: false);
+                }
+            }
+        }
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public virtual void SelectMany_Include_reference_GroupBy_Select(bool useString)
```
smitpatel: Most tests added in a patch :trollface:

anpete: Yeah, "thanks" for that hehe

anpete:
@ajcvickers @smitpatel We need to decide on whether this should be quirked by default. I'm leaning towards having the fix enabled by default:

1) We added good test coverage with this (good work @smitpatel)
2) Because the fix itself is fairly targeted, I think the chance we break any queries that are currently working in 2.0 is low.

ajcvickers:
@anpete @smitpatel @divega For now, the fix should be enabled by default since it will be available in a pre-release build before the RTM build, and for the pre-release I think we want to gather as much information as possible. If people report regressions, then we can always flip it off before RTM if we need to. /cc @Eilon

anpete:
Added quirks

PR #9871, opened 2017-09-21
Fix #9741 - AddDbContextPool in asp.net core System.InvalidOperationException

### LINKED ISSUE #9741: AddDbContextPool in asp.net core System.InvalidOperationException
I'm using db context pooling in a Asp.Net core MVC application.
When I launch multiple simultaneous requests I get an error :

```
Exception message: System.InvalidOperationException occurred in Microsoft.EntityFramework.Core.dll
A second operation started on this context before a previous operation completed. Any instance members are not guaranteed to be thread safe.

or 
Exception message: System.ObjectDisposedException: 'Cannot access a disposed object.

code;
IQueryable<TEntity> query = DbSet;
            if (filter != null)
                query = query.Where(filter); // exception occurs here

```
Maybe I'm not using DbContext pooling the right way...

### Steps to reproduce
I launch mutiple requests in browser to the same web api controller method, which retrieves some data from the database.

```c#
 // Dependency injection is configured in Startup as follows
    services.AddScoped<IMyDbContext>(provider => provider.GetService<MyDbContext>());

// if I use service.AddScoped<IMyDbContext,MyDbContext>();
//  I get the impression db context pooling is not working, 
//  but I don't get the exception message either.

// db context pooling in Startup class
   services.AddDbContextPool<MyDbContext>(options =>
            {
                string connectionString = Configuration.GetConnectionString("MyConnection");
                options.UseSqlServer(connectionString);
            },10);
```

### Further technical details
EF Core version:Microsoft.EntityFrameworkCore.SqlServer" Version="2.0.0"
Database Provider:  Microsoft.EntityFrameworkCore.SqlServer
Operating system: Win10
IDE: Visual Studio 2017

anpete:
Problem: Pooling was not resilient to multiple DbContext.Dispose calls, which meant that a context instance could incorrectly be returned to the pool twice.
Solution: Adds a DbContext async local to track which "context" (e.g. request scope) owns the context and use it to ensure a single return to the pool.

### DIFF
--- a/src/EFCore/DbContext.cs
+++ b/src/EFCore/DbContext.cs
@@ -64,6 +64,8 @@ namespace Microsoft.EntityFrameworkCore
         private bool _initializing;
         private bool _disposed;
 
+        private AsyncLocal<bool> _leased;
+        
         /// <summary>
         ///     <para>
         ///         Initializes a new instance of the <see cref="DbContext" /> class. The
@@ -279,7 +281,7 @@ namespace Microsoft.EntityFrameworkCore
         [DebuggerStepThrough]
         internal void CheckDisposed()
         {
-            if (_disposed)
+            if (_disposed || _leased?.Value == false)
             {
                 throw new ObjectDisposedException(GetType().ShortDisplayName(), CoreStrings.ContextDisposed);
             }
@@ -482,6 +484,7 @@ namespace Microsoft.EntityFrameworkCore
             Check.NotNull(contextPool, nameof(contextPool));
 
             _dbContextPool = contextPool;
+            _leased = new AsyncLocal<bool> { Value = true };
         }
 
         DbContextPoolConfigurationSnapshot IDbContextPoolable.SnapshotConfiguration()
@@ -489,10 +492,10 @@ namespace Microsoft.EntityFrameworkCore
                 _changeTracker?.AutoDetectChangesEnabled,
                 _changeTracker?.QueryTrackingBehavior,
                 _database?.AutoTransactionsEnabled);
-
+        
         void IDbContextPoolable.Resurrect(DbContextPoolConfigurationSnapshot configurationSnapshot)
         {
-            _disposed = false;
+            _leased.Value = true;
 
             if (configurationSnapshot.AutoDetectChangesEnabled != null)
             {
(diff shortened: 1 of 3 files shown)


src/EFCore/DbContext.cs
```diff
@@ -94,7 +96,7 @@ public DbContext([NotNull] DbContextOptions options)
             }
 
             _options = options;
-
+
```
AndriySvyryd: Reformat

src/EFCore/DbContext.cs
```diff
@@ -64,6 +64,8 @@ public class DbContext :
         private bool _initializing;
         private bool _disposed;
 
+        private AsyncLocal<bool> _leased;
```
AndriySvyryd: Wouldn't this work with a `bool?` instead of `AsyncLocal`?

anpete: No.

AndriySvyryd: Can you elaborate? The tests pass with `bool?`

anpete: Are you sure? Double_dispose_concurrency_test fails for me without the async local.

Request A calls Dispose - context A is returned to pool.
Request B calls Rent and acquires context A.
Request A calls Dispose and context A is returned to pool. Boom!

AndriySvyryd: I see. The test passes unless I ran it on its own.

src/EFCore/DbContext.cs
```diff
@@ -482,17 +484,18 @@ void IDbContextPoolable.SetPool(IDbContextPool contextPool)
             Check.NotNull(contextPool, nameof(contextPool));
 
             _dbContextPool = contextPool;
+            _leased = new AsyncLocal<bool> { Value = true };
```
ajcvickers: Is this thread-safe? Depending on what caches have been flushed, thread A could see the old value for _leased even though thread B has set a new value, Thread A would then see this instance as not leased and attempt to lease it.

anpete: AFAIK, I thought that was the point of AsyncLocal. Perhaps @AndriySvyryd or @divega can comment.

ajcvickers: My understanding is that if you set a value on an AsyncLocal, then it will only be visible to the current execution context--i.e. `_leased.Value = true;` But unless the compiler is doing some magic on the _leased field, then it's just a normal field holding something that happens to be an AsyncLocal.

anpete: Oh, you mean the _leased field. SetPool is only ever call once when the context is created.

ajcvickers: Got it! :-)

PR #9872, opened 2017-09-21
Query: Add support for translating GroupBy() to SQL when aggregate operator used on group

### LINKED ISSUE #2341: Relational: Support translating GroupBy() to SQL
_Updated as of 9/21/2017 based on implementation introduced_

LINQ's `GroupBy()` operators can sometimes be translated to SQL's `GROUP BY` clauses, in particular when aggregate functions are applied in the projection. 

# Scope for 2.1 release

Our current intention is for the scope of the work in 2.1 to improve how LINQ's `GroupBy` is evaluated in this particular scenario:

*Grouping on a simple expression that references a single column or multiple columns (using an anonymous type) of the query root and then projecting any **aggregate** operation from (Count, Sum, Average, Min, Max) and/or any individual property that is part of the grouping key (this will be translated to GROUP BY in SQL)*

## What will be supported in 2.1

``` csharp

// Grouping by single column projecting aggregate or key
db.Orders.GroupBy(o => o.CustomerId).Select(g => g.Count());
db.Orders.GroupBy(o => o.CustomerId).Select(g => new { CustomerId = g.Key, Count = g.Count() });

// Grouping by multiple columns (using anonymous type) projecting aggregate or key or nested key
db.Orders.GroupBy(o => new { o.CustomerId, o.OrderDate }).Select(g => g.Sum(o => o.Cost));
db.Orders.GroupBy(o => new { o.CustomerId, o.OrderDate })
  .Select(g => new { CustomerId = g.Key.CustomerId, Sum = g.Sum(o => o.Cost) });
db.Orders.GroupBy(o => new { o.CustomerId, o.OrderDate })
  .Select(g => new { Key = g.Key, Sum = g.Sum(o => o.Cost) }); // Where Key will be anonymous object

// Grouping after complex query root
(from o in db.Orders.Where(o => o.OrderID < 10400).OrderBy(o => o.OrderDate).Take(100)
    join c in db.Customers.Where(c => c.CustomerID != "DRACD" && c.CustomerID != "FOLKO").OrderBy(c => c.City).Skip(10).Take(50)
        on o.CustomerID equals c.CustomerID
    group o by c.CustomerID)
.Select(g => new { g.Key, Count = g.Average(o => o.OrderID) });

db.Orders.OrderBy(o => o.OrderID)
    .Skip(80)
    .Take(500)
    .GroupBy(o => o.CustomerID)
    .Select(g => g.Max(o => o.OrderID));

// All above examples have group of entity types after GroupBy

// Selecting Group of anonymous types containing multiple columns
db.Orders.GroupBy(o => o.CustomerId, new {o.OrderDate, o.Price}).Select(g => g.Sum(t => t.Price));

```

## Scenarios that we are not planning to improve in the 2.1 release

1. Grouping on constants 
2. Grouping on an entity (e.g. a reference navigation property) 
3. Projecting non-aggregate scalar subqueries after grouping, e.g. `FirstOrDefault()`
4. Making groups of multiple entityTypes using anonymous types.

All the scenarios above present different variations depending on what happens after the `GroupBy`, e.g. is there an aggregate function or not, is the key mentioned in the projection or not, etc. These scenarios will still result in client evaluation.

We would appreciate if customers that care about EF Core supporting any of those scenarios that are scoped out from 2.1 to create individual issues for them, up-vote them, and keep the discussion there.

smitpatel:
Resolves #2341


src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -272,5 +293,259 @@ var outputDataInfo
...
+                _queryModelVisitor.QueryCompilationContext.AddOrUpdateMapping(
+                    _groupQuerySource,
+                    currentParameter);
+            }
+
+            private bool IsAggregateGroupBySelector(Expression expression)
+            {
+                if (expression is NewExpression newExpression)
+                {
+                    return newExpression.Arguments.Select(e => IsAggregateSubQueryExpression(e) || IsKeySelector(e))
```
maumar: return newExpression.Arguments.All(e => IsAggregateSubQueryExpression(e) || IsKeySelector(e)) seems simpler

src/EFCore/Query/EntityQueryModelVisitor.cs
```diff
@@ -1012,9 +1012,7 @@ var transparentIdentifierType
             Check.NotNull(selectClause, nameof(selectClause));
             Check.NotNull(queryModel, nameof(queryModel));
 
-            var sequenceType = _expression.Type.GetSequenceType();
-
-            if (selectClause.Selector.Type == sequenceType
+            if (selectClause.Selector.Type == _expression.Type.GetSequenceType()
```
maumar: revert this, sequenceType was used in two places, so it made sense to store it in variable.

smitpatel: :trollface: 

This was changed because the _expression can change between 2 calls. So both times `sequenceType` would evaluate to different things.

src/EFCore.Relational/Query/Expressions/SelectExpression.cs
```diff
@@ -813,6 +827,15 @@ public virtual void AddToPredicate([NotNull] Expression predicate)
         }
 
         /// <summary>
+        ///     Adds list of expressions to the GROUP BY clause of this SelectExpression
+        /// </summary>
+        /// <param name="groupingExpressions"> The grouping expressions </param>
+        public virtual void AddToGroupBy([NotNull] Expression[] groupingExpressions)
+        {
+            _groupBy.AddRange(groupingExpressions);
```
anpete: Check not null

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -182,6 +188,21 @@ var propertyCallExpressions
         /// </returns>
         public override Expression Visit(Expression expression)
         {
+            if (_topLevelProjection
+                && (QueryModelVisitor.Expression as MethodCallExpression)?.Method.MethodIsClosedFormOf(QueryModelVisitor.QueryCompilationContext.QueryMethodProvider.GroupByMethod) == true)
```
anpete: Break long line

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -272,5 +293,258 @@ var outputDataInfo
...
+            {
+                if (!_translateToSql)
+                {
+                    var sqlExpression = _selectExpression.GetProjectionForMemberInfo(node.Member);
+                    if (sqlExpression != null)
+                    {
+                        return BindSqlToValueBuffer(sqlExpression, node.Type);
+                    }
+
+                    if (node.Member.Name == "Key"
```
anpete: nameof?

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -272,5 +293,258 @@ var outputDataInfo
...
+
+                return null;
+            }
+
+            private bool CanTranslate(Expression expression)
+            {
+                // Check for Query shape
+                if (IsAggregateGroupBySelector(expression))
+                {
+                    var groupByResultOperator =
```
anpete: Return this from IsAggregateGroupBySelector?

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -250,6 +250,14 @@ public virtual Expression VisitSelect(SelectExpression selectExpression)
                 GeneratePredicate(selectExpression.Predicate);
             }
 
+            if (selectExpression.GroupBy.Any())
```
anpete: We should change to .Length > 0 for these checks to avoid allocs

src/EFCore.Specification.Tests/Query/AsyncGroupByQueryTestBase.cs
```diff
@@ -0,0 +1,1270 @@
...
+        [ConditionalFact]
+        public virtual async Task GroupBy_Property_Select_Sum()
+        {
+            await AssertQueryScalar<Order>(os => os.GroupBy(o => o.CustomerID).Select(g => g.Sum(o => o.OrderID)));
+        }
+
+        [ConditionalFact]
+        public virtual async Task GroupBy_Property_Select_Sum_Min_Max_Avg()
+        {
+            await AssertQuery<Order>(
```
anpete: Can we still throw on no element group? Is there a test for it?

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
smitpatel: `IsAggregateGroupBySelector` is actually not looking into group by query source or resultoperator. It would just verify the shape of aggregate operator query.

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
smitpatel: `nameof(???)`

src/EFCore.Specification.Tests/Query/AsyncGroupByQueryTestBase.cs
smitpatel: Can you elaborate?

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
anpete: IGrouping.Key

src/EFCore.Specification.Tests/Query/AsyncGroupByQueryTestBase.cs
smitpatel: Conclusion: `GroupBy` does not produce a group with no elements in it. For every key value pair, grouping will generate group for each key. So for every key there will be group with some value. That value can be null but a non-empty group containing null will always be there.

PR #9874, opened 2017-09-22
Add table splitting and inheritance support to data seeding.

AndriySvyryd:
Use two new instances of StateManager for the source and target model to be able to reference the correct columns.

Part of #629


src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -304,6 +308,7 @@ public virtual IReadOnlyList<MigrationOperation> GetDifferences(IModel source, I
             return dropForeignKeyOperations
                 .Concat(dropTableOperations)
                 .Concat(dropOperations)
+                .Concat(sourceDataOperations)
```
smitpatel: Can this ordering cause an issue if the Data Delete is on the dropped table? or do we not generate delete data operations for tables dropped?

AndriySvyryd: The latter. We only consider added and (potentially) changed tables

AndriySvyryd: This order is needed if the user drops a dependent table and deletes the corresponding principal seeds.

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -333,6 +338,9 @@ public virtual IReadOnlyList<MigrationOperation> GetDifferences(IModel source, I
             [CanBeNull] IModel target,
             [NotNull] DiffContext diffContext)
         {
+            _sourceStateManager = source == null ? null : new StateManager(StateManagerStateManagerDependencies.With(source));
+            _targetStateManager = target == null ? null : new StateManager(StateManagerStateManagerDependencies.With(target));
```
bricelam: I think C# needs a `¿¿` operator for this.

src/EFCore/Internal/DictionaryExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    [DebuggerStepThrough]
+    public static class DictionaryExtensions
```
bricelam: Real `internal`? (and shared source)

test/EFCore.Relational.Tests/Migrations/Internal/MigrationsModelDifferTest.cs
```diff
@@ -722,104 +772,6 @@ private enum SomeEnum
         }
 
         [Fact]
-        public void Drop_column()
```
bricelam: Lost coverage?

AndriySvyryd: Already covered by `Add_column_*` tests since they assert up and down operations

PR #9884, opened 2017-09-22
Query: Fix for projecting out bool values with ternary expression throws

### LINKED ISSUE #9818: Projecting out bool value from ternary expression throws
OneToZeroOrOne relationships do not let you check for null and provide values back in the EF query. 
 Instead you need to return the entire object from the database and then inspect the properties you are concerned about.

I also published this question to https://github.com/aspnet/EntityFrameworkCore/issues/6177 before I realized it was closed.

This code fails :
``await _repo 
   .TableNoTracking 
   .Where(x => x.Username == profileName) 
   .Select(x => new { IsAdmin = x.User != null ? x.User.IsAdmin : false }) 
   .FirstOrDefaultAsync()``

This code works fine, but we are forced to return the entire object.
`await _repo .TableNoTracking 
   .Where(x => x.Username == profileName) 
   .Select(x => new { User = x.User }) 
   .FirstOrDefaultAsync()`

### Further technical details
EF Core version: 2.0.0
Operating system: Windows 8.1
IDE: 
Microsoft Visual Studio Enterprise 2017 
Version 15.3.4
VisualStudio.15.Release/15.3.4+26730.15
Microsoft .NET Framework
Version 4.7.02053

smitpatel:
Issue: When generating conditional expression we generated typed literal for true/false but that was not happening for nullable bools inside converts. Hence whole expression was getting upcasted to int type causing InvalidCastException

Resolves #9818

### DIFF
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -165,7 +165,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
         protected virtual string TypedFalseLiteral => "CAST(0 AS BIT)";
 
         /// <summary>
-        /// Whether schemas should be generated when generating identifiers.
+        ///     Whether schemas should be generated when generating identifiers.
         /// </summary>
         protected virtual bool SupportsSchemas { get; } = true;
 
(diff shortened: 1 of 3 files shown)


src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1078,8 +1079,9 @@ protected override Expression VisitConditional(ConditionalExpression conditional
 
                 _relationalCommandBuilder.Append(" ELSE ");
 
-                if (conditionalExpression.IfFalse is ConstantExpression constantIfFalse
-                    && constantIfFalse.Type == typeof(bool))
+                if (conditionalExpression.IfFalse.RemoveConvert() is ConstantExpression constantIfFalse
+                    && constantIfFalse.Value != null
```
smitpatel: null are handled in VisitConstant properly.
On higher level we could move this special handling to VisitConstant only based on condition on which block we are in or something.

Issue #9143, opened 2017-07-11
DefaultQuerySqlGenerator not flexible enough for provider override


roji:
As part of #9126, I set about adding a PostgreSQL-specific ILikeExpression and modifying NpgsqlQuerySqlGenerator to render it correctly. I hit several limitations with DefaultQuerySqlGenerator not exposing enough functionality to subclasses:

* _typeMapping is private, and so I can't set it properly (as `VisitLike()` does)
* isSearchCondition is a private static method, so there's no way to introduce a new expression type that is identified as a search condition. This means that a useless `= TRUE` is rendered out, and may have other consequences I'm not aware of as well.

I can submit PRs for these two specific problems, but I've run into similar issues previously, so you guys may want to do a proper review of DefaultQuerySqlGenerator keeping providers in mind.

divega:
@roji Thanks for reporting this. 

> I can submit PRs for these two specific problems, but I've run into similar issues previously, so you guys may want to do a proper review of DefaultQuerySqlGenerator keeping providers in mind.

Our general approach is intentionally to avoid overdesigning the provider model and wait for feedback like yours to help us triangulate where the extensibility and additional flexibility is needed. This obviously works much better if we have time to react to the feedback... So the second bullet point from my list of regrets at https://github.com/aspnet/EntityFramework/issues/9126#issuecomment-314546027 applies here too.

Fortunately in this case I believe we can take changes like the ones you suggest later, e.g. in 2.1, without them being breaking changes. So it would be great if you can send a PR. I also hope you either do something that works (even if it is suboptimal) for 2.0 or wait until 2.1 to add the new Like variant.

smitpatel:
1. We should expose type mapping allowing providers to set it. Not sure if access as property or functions.
2. There should not be side effect other than extra = True. Though the whole class `BooleanConditionTranslatingExpressionVisitor` is private. There is no way for providers to influence it at present. For custom nodes, providers can override `VisitChildren` but at present it allows no control over if Child are supposed to be condition or not. And there is no way to extend the set of condition nodes as you figured. At one point @anpete & me were thinking of a base class for all the custom expression defined in EF (& providers) such a base class would allow to add properties like that which we can directly evaluate. Give us some time to discuss 2nd point in team. The harm of that would be none apart from bit longer SQL.

roji:
Thanks (as always) for your reactions and for caring :)

[...]

PR #9885, opened 2017-09-22
Query: Apply type facets to all arguments for string.Contains

### LINKED ISSUE #9582: Query on field non unicode string should not append N
I have setting all string use non unicode varchar and select append N in generated SQL

### Steps to reproduce
```c#
DbContext
        {
            foreach (var property in modelBuilder.Model.GetEntityTypes().SelectMany(e => e.GetProperties().Where(p => p.ClrType == typeof(string))))
            {
                property.IsUnicode(false);
            }
        }
```
```c#
 public class TipoServicioConfig : IEntityTypeConfiguration<TipoServicio>
    {
        public void Configure(EntityTypeBuilder<TipoServicio> builder)
        {
            builder.HasKey(ts => ts.Id);

            builder.Property(ts => ts.Id).IsRequired();
            builder.Property(ts => ts.Nombre).IsRequired().HasMaxLength(20);
            
           //it does not work either
          //builder.Property(ts => ts.Nombre).IsRequired().HasMaxLength(20).HasColumnType("varchar(20)");
        }
    }
```

```c#
   public class TipoServicio
    {
 
        public int Id{ get; set; }
        public string Nombre { get; set; }

    }
```
```c#
var query = DbContext.Set<TipoServicio>().Where(xx => xx.Nombre.Contains("lla")).ToList();
```

```sql
      SELECT [xx].[Id], [xx].[Nombre]
      FROM [TipoServicio] AS [xx]
      WHERE CHARINDEX(N'lla', [xx].[Nombre]) > 0
```

### Further technical details
EF Core version: 2.0.0
Database Provider:  Microsoft.EntityFrameworkCore.SqlServer
Operating system: Windows 10
IDE: Visual Studio 2017 (15.3)

smitpatel:
Resolves #9582

This solves the issue in simple cases. For complex cases we need to visit Expression Tree and infer facets and propagate them through tree.

### DIFF
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -165,7 +165,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
         protected virtual string TypedFalseLiteral => "CAST(0 AS BIT)";
 
         /// <summary>
-        /// Whether schemas should be generated when generating identifiers.
+        ///     Whether schemas should be generated when generating identifiers.
         /// </summary>
         protected virtual bool SupportsSchemas { get; } = true;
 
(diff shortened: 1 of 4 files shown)


src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -1366,6 +1366,13 @@ public virtual Expression VisitSqlFunction(SqlFunctionExpression sqlFunctionExpr
             _relationalCommandBuilder.Append(functionName);
             _relationalCommandBuilder.Append("(");
 
+            if (functionName == "CHARINDEX")
```
anpete: Hard-coded in core?

smitpatel: I thought it was same across providers.
We cannot move this to SQL server due to limitation here #9143
Recalling this. Not so easy to get this done it seems. I will remove milestone and bring it back to triage when I m back in office.

PR #9887, opened 2017-09-23
Query: Fix for incorrect QueryCache hit when chaining a streaming N+1 query

### LINKED ISSUE #9301: AsyncQueryTestBase.Mixed_sync_async_query test still flaky?
While working on my EF.Core provider I ran into some weird issues with [`Mixed_sync_async_query`](https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore.Specification.Tests/Query/AsyncSimpleQueryTestBase.cs#L98) test. After some experimenting it doesn't seem very trustworthy to me. For example it fails if I add `.ToList()` and run it against InMemory database
```c#
[ConditionalFact]
public virtual async Task Mixed_sync_async_query()
{
    {
        var results
            = (await context.Customers
                .Select(c => new
                {
                    c.CustomerID,
                    Orders = context.Orders.Where(o => o.Customer.CustomerID == c.CustomerID)
                                    .ToList() /// <= HERE!
                }).ToListAsync())
                .Select(x => new
                {
                    Orders = x.Orders
                        .GroupJoin(new[] { "ALFKI" }, y => x.CustomerID, y => y, (h, id) => new
                        {
                            h.Customer
                        })
                })
                .ToList();

        Assert.Equal(546, results.SelectMany(r => r.Orders).ToList().Count);
    }
}
```
```
Assert.Equal() Failure
Expected: 546
Actual:   830
```

I went ahead a bit further and excluded the possible influence of asynchrony and also explicitly materialized the `Orders` collection in a separate step
```c#
[ConditionalFact]
public virtual void Mixed_sync_query()
{
    {
        var list1
            = context.Customers
                .Select(c => new
                {
                    c.CustomerID,
                    Orders = context.Orders.Where(o => o.Customer.CustomerID == c.CustomerID)
                }).ToList();

        var list2 = list1.Select(x => new { x.CustomerID, Orders = x.Orders.ToList() }).ToList();

        var results = list2
            .Select(x => new
            {
                Orders = x.Orders
                    .GroupJoin(new[] { "ALFKI" }, y => x.CustomerID, y => y, (h, id) => new
                    {
                        h.Customer
                    })
            })
            .ToList();

        Assert.Equal(546, results.SelectMany(r => r.Orders).ToList().Count);
    }
}
```
and it still returns 830 rather than 546.

I have to confess I didn't check each and every line of data to prove that the test should expect 830 instead of 546, but also I can't imagine its original author doing such analysis. So for now my question is: **How can one explain different results returned with and without `.ToList()` on the indicated position?**

### Further technical details
EF Core version: 2.0.0-preview1-final
Database Provider: Microsoft.EntityFrameworkCore.InMemory
Operating system: Windows 10
IDE: Visual Studio 2017

smitpatel:
Issue:
When the query is projecting out a collection object in non-navigation case, we cause N+1 queries.
To support streaming behaviour, we inject QueryableAdapter (or async version) which would be evaluated on when enumerated.

When we are chaining on the navigation after such query, the QueryableAdapter is captured in ConstantExpression.
In our ExpressionEqualityComparer, during GetHashCode, we add such value if ConstantExpression is of type IQueryable. (so that we can fall back to Comparison)
During comparing such ConstantExpressions, we assumed if they are of type IQueryable then they are same. Which is true for IQueryable from DbSet.
But for QueryableAdapter which are also IQueryable, it could still evaluate to different results.
Another thing is, in `EntityQueryableExpressionVisitor` ConstantExpression of EntityQueryable are evaluated. But rest of them are kept intact.
Since we returned equality=true for all IQueryable, every subsequent query even though had different QueryableAdapter which gave different results we made cache hit and re-ran compiled query which has older value of QueryableAdapter (due to EntityQueryableExpressionVisitor) and would evaluate to incorrect values.

Resolves #9301

### DIFF
--- a/src/EFCore/Query/Internal/ExpressionEqualityComparer.cs
+++ b/src/EFCore/Query/Internal/ExpressionEqualityComparer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 
 // ReSharper disable SwitchStatementMissingSomeCases
@@ -434,8 +435,8 @@ namespace Microsoft.EntityFrameworkCore.Query.Internal
                     return false; // EnumerableQueries are opaque
                 }
 
-                if (a.Value is IQueryable
-                    && b.Value is IQueryable
+                if (a.IsEntityQueryable()
+                    && b.IsEntityQueryable()
                     && a.Value.GetType() == b.Value.GetType())
                 {
                     return true;
--- a/src/EFCore.Specification.Tests/Query/AsyncSimpleQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/AsyncSimpleQueryTestBase.cs
@@ -173,7 +173,7 @@ namespace Microsoft.EntityFrameworkCore.Query
                             })
                         .ToList();
 
-                Assert.Equal(546, results.SelectMany(r => r.Orders).ToList().Count);
+                Assert.Equal(830, results.SelectMany(r => r.Orders).ToList().Count);
             }
         }
 
--- a/src/EFCore.Specification.Tests/Query/SimpleQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/SimpleQueryTestBase.cs
@@ -4009,5 +4009,34 @@ namespace Microsoft.EntityFrameworkCore.Query
                     .Take(10),
                 elementSorter: e => e.Id);
         }
+
+        [ConditionalFact]
+        public virtual void Streaming_chained_sync_query()
+        {
+            using (var context = CreateContext())
+            {
+                var results
+                    = (context.Customers
+                        .Select(
+                            c => new
+                            {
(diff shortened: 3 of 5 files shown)


src/EFCore/Query/Internal/ExpressionEqualityComparer.cs
```diff
@@ -438,7 +438,13 @@ private static bool CompareConstant(ConstantExpression a, ConstantExpression b)
                     && b.Value is IQueryable
                     && a.Value.GetType() == b.Value.GetType())
                 {
-                    return true;
+                    if (a.IsEntityQueryable()
```
smitpatel: This is one possible way to resolve this issue.
I did not modify `GetHashCode` which skips IQueryable too (perhaps to verify `EnumerableQuery` here). We could perhaps add check there too.
I got this check from `EntityQueryableExpressionVisitor` which essentially states that we don't do anything with any other expression then EntityQueryable so such expression should be marked different unless they are same.

smitpatel: @anpete

anpete:
> When we are chaining on the navigation after such query, the QueryableAdapter is captured in ConstantExpression.

Is this because we eval the subquery during parameter extraction?

PR #9903, opened 2017-09-26
Query: Use correct overload of DelimitIdentifier() in DefaultQuerySqlGenerator

### LINKED ISSUE #9560: Query: Ensure ISqlGenerationHelper is used when generating identifiers
Investigate: It looks like we may be incorrectly outputing a configured schema when generating table names in Sqlite queries.

bricelam:
Fixes #9560

I reviewed all places we generate SQL for schema and table name to ensure they were consistent.

You may want to review [ignoring whitespace changes](https://github.com/aspnet/EntityFrameworkCore/pull/9903/files?w=1) since the R# settings have diverged from Visual Studio's default formatting.

### DIFF
--- a/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
+++ b/src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
@@ -165,11 +165,6 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
         protected virtual string TypedFalseLiteral => "CAST(0 AS BIT)";
 
         /// <summary>
-        ///     Whether schemas should be generated when generating identifiers.
-        /// </summary>
-        protected virtual bool SupportsSchemas { get; } = true;
-
-        /// <summary>
         ///     The default alias separator.
         /// </summary>
         protected virtual string AliasSeparator { get; } = " AS ";
@@ -651,14 +646,7 @@ namespace Microsoft.EntityFrameworkCore.Query.Sql
         {
             Check.NotNull(tableExpression, nameof(tableExpression));
 
-            if (SupportsSchemas
-                && tableExpression.Schema != null)
-            {
-                _relationalCommandBuilder.Append(SqlGenerator.DelimitIdentifier(tableExpression.Schema))
-                    .Append(".");
-            }
-
-            _relationalCommandBuilder.Append(SqlGenerator.DelimitIdentifier(tableExpression.Table))
+            _relationalCommandBuilder.Append(SqlGenerator.DelimitIdentifier(tableExpression.Table, tableExpression.Schema))
                 .Append(AliasSeparator)
                 .Append(SqlGenerator.DelimitIdentifier(tableExpression.Alias));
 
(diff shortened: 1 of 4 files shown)


src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
```diff
@@ -645,14 +645,7 @@ public virtual Expression VisitTable(TableExpression tableExpression)
         {
             Check.NotNull(tableExpression, nameof(tableExpression));
 
-            if (SupportsSchemas
```
bricelam: @anpete Can we get rid of `SupportsSchemas`? It was introduced for Oracle, but SQLite was already handling this by overriding the `SqlGenerator.DelimitIdentifier()` overload. That overload is used by Query, Update, and Migrations.

anpete: I'm fine if it works. Can you check that we don't break the Oracle sample.

anpete:
Can we decide on the whitespace formatting one or the other? We either need to update our R# settings, or the changes here should be reverted. cc @ajcvickers

ajcvickers:
Filed #9911 for R# settings

src/EFCore.Relational/Query/Sql/DefaultQuerySqlGenerator.cs
bricelam: This is turning into a yak shaving exercise. Oracle solution was broken by #9629. Need to install Oracle Database... lol, I'll keep at it.

bricelam: I'm getting thousands of test failures with the following:

> ORA-00972: identifier is too long

I'm using version 12.1. Is the sample only compatible with 12.2?

smitpatel: According to stack overflow max length was 30 upto 12.1 and was increased in 12.2.
We need to have a vm setup somewhere for this kind of testing so that all of us don't have to install oracle. May be efci host machine.

anpete: Yep, the required version is listed in the [readme](https://github.com/aspnet/EntityFrameworkCore/tree/dev/samples/OracleProvider): 12c Release 2.

bricelam: Lame. Latest Docker image is 12.1. I'll spin up a VM.

bricelam: (But when there's a 12.2 Docker image, I'll send out instructions to the team since I've got it all working.)

smitpatel:
@bricelam - You need to rebase this on latest dev. All the issues with running Oracle has been fixed.

bricelam:
Total tests: 5474. Passed: 5400. Failed: 0. Skipped: 74.

PR #9907, opened 2017-09-27
Move parts of type mapping to Core to facilitate type conversions

ajcvickers:
Issue #242

Also, start depending on runtime always having a type mapping associated with the property. This involves building the history model with conventions and updating tests that build the model manually.


src/EFCore/Metadata/Internal/CoreAnnotationNames.cs
```diff
@@ -38,5 +38,11 @@ public static class CoreAnnotationNames
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public const string PropertyAccessModeAnnotation = "PropertyAccessMode";
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public const string TypeMapping = "Relational:TypeMapping"; // "Relational:" prefix to prevent breaking changes from 2.0
```
AndriySvyryd: File an issue to fix this in 3.0

src/EFCore.Relational/Internal/RelationalPropertyExtensions.cs
```diff
@@ -27,5 +29,13 @@ public static string FormatColumns([NotNull] this IEnumerable<IProperty> propert
...
         public static string GetConfiguredColumnType([NotNull] this IProperty property)
             => (string)property[RelationalAnnotationNames.ColumnType];
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public static RelationalTypeMapping FindRelationalMapping(
+            [NotNull] this IProperty property) 
+            => (RelationalTypeMapping)property[CoreAnnotationNames.TypeMapping];
```
AndriySvyryd: Should this be `as RelationalTypeMapping` ?

src/EFCore/Metadata/Internal/CoreAnnotationNames.cs
ajcvickers: @9913

src/EFCore.Relational/Internal/RelationalPropertyExtensions.cs
ajcvickers: Maybe. If it's a relational model, then any type mapping it has should be a RelationalTypeMapping. Not sure yet what the best way to enforce this is.

AndriySvyryd: Do we really need to enforce it? As long as we can get a RelationalTypeMapping it should be fine and the relational validation already covers this

PR #9918, opened 2017-09-28
Add change tracking support for shadow navigations to non-shadow entity types

AndriySvyryd:
More tests will be added

Part of #749


src/EFCore/Metadata/Internal/ShadowCollectionAccessor.cs
```diff
@@ -0,0 +1,122 @@
...
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Metadata.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class ShadowCollectionAccessor<TCollection, TElement> : IShadowCollectionAccessor
+        where TCollection : class, IEnumerable<TElement>
```
ajcvickers: Does this need to be generic? For the collection snapshot, we just use object.

ajcvickers:
I think it would be good to discuss when we are in the office. I'd like to understand the need to move things to the state entry.

AndriySvyryd:
The methods are moved to InternalStateEntry to abstract away the shadowness in the same way it's done for the properties, without adding any dependencies to the accessor implementation.

src/EFCore/Metadata/Internal/ShadowCollectionAccessor.cs
AndriySvyryd: This class is mostly superfluous, I'll just inline it.

src/EFCore/ChangeTracking/Internal/InternalEntityEntry.cs
```diff
@@ -526,6 +527,48 @@ protected virtual void WritePropertyValue([NotNull] IPropertyBase propertyBase,
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        public virtual IEnumerable GetOrCreateCollection([NotNull] INavigation navigation)
+        {
+            return navigation.GetCollectionAccessor().GetOrCreate(Entity);
+        }
```
ajcvickers: Expression body?

AndriySvyryd: Forgot to add `Debug.Assert(!navigation.IsShadowProperty);`

src/EFCore/Metadata/Internal/ClrCollectionAccessor.cs
```diff
@@ -12,7 +13,7 @@ namespace Microsoft.EntityFrameworkCore.Metadata.Internal
     ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
     ///     directly from your code. This API may change or be removed in future releases.
     /// </summary>
-    public class ClrICollectionAccessor<TEntity, TCollection, TElement> : IClrCollectionAccessor
+    public class ClrCollectionAccessor<TEntity, TCollection, TElement> : IClrCollectionAccessor
```
ajcvickers: >ClrCollectionAccessor [](start = 17, length = 21)

The "I" is there for a reason. It means this is the implementation for ICollection, rather than just the gneral concept of some kind of collection of things on a thing.

AndriySvyryd: Still, it goes against our naming guidelines

src/EFCore/Metadata/Internal/IClrCollectionAccessor.cs
```diff
@@ -41,19 +42,19 @@ public interface IClrCollectionAccessor
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
-        object Create();
+        IEnumerable Create();
```
ajcvickers: >IEnumerable [](start = 8, length = 11)

Do we need to constrain this to only collections that implement IEnumerable?

AndriySvyryd: We already have this constraint, see https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/EFCore/Metadata/Internal/ClrCollectionAccessorFactory.cs#L42

Issue #9577, opened 2017-08-25
Query Testing: Failure of assertion framework should give better information


smitpatel:
The following output makes it ambiguous if there was error with linq query or just test. 

```
Test 'Microsoft.EntityFrameworkCore.Query.SimpleQuerySqlServerTest.OrderBy_Skip_GroupBy' failed: System.AggregateException : One or more errors occurred.
---- System.ArgumentException : At least one object must implement IComparable.
	at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
	at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
	at System.Threading.Tasks.Task.Wait()
	D:\code\EntityFrameworkCore\src\EFCore.Specification.Tests\Query\QueryTestBase.cs(83,0): at Microsoft.EntityFrameworkCore.Query.QueryTestBase`1.AssertQuery[TItem1](Func`2 query, Func`2 elementSorter, Action`2 elementAsserter, Boolean assertOrder, Int32 entryCount)
	D:\code\EntityFrameworkCore\src\EFCore.Specification.Tests\Query\SimpleQueryTestBase.GroupBy.cs(513,0): at Microsoft.EntityFrameworkCore.Query.SimpleQueryTestBase`1.OrderBy_Skip_GroupBy()
	D:\code\EntityFrameworkCore\test\EFCore.SqlServer.FunctionalTests\Query\SimpleQuerySqlServerTest.GroupBy.cs(438,0): at Microsoft.EntityFrameworkCore.Query.SimpleQuerySqlServerTest.OrderBy_Skip_GroupBy()
	----- Inner Stack Trace -----
	at System.Collections.Comparer.Compare(Object a, Object b)
	at System.Collections.Generic.ObjectComparer`1.Compare(T x, T y)
	at System.Linq.EnumerableSorter`2.CompareKeys(Int32 index1, Int32 index2)
	at System.Linq.EnumerableSorter`1.QuickSort(Int32[] map, Int32 left, Int32 right)
	at System.Linq.EnumerableSorter`1.Sort(TElement[] elements, Int32 count)
	at System.Linq.OrderedEnumerable`1.<GetEnumerator>d__1.MoveNext()
	at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
	at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
[...]

PR #9928, opened 2017-09-29
[TINY] Fix to #9577 - Query Testing: Failure of assertion framework should give better information

maumar:
Added some logic to TestHelpers.AssertResults to check for IComparable interface on the results that are about to be sorted by a default sorter (i.e. e => e)


smitpatel:
We need to remove client side ordering. It is too difficult to write a new test which doesn't project out entities

maumar:
new version up, switching to legacy, unosorted verification if the results objects are not order comparable and the custom sorter was not provided

ajcvickers:
@maumar What is the status of this PR?

maumar:
@smitpatel ping ;)

Issue #6480, opened 2016-09-06
Support mapping ulong, uint, ushort, sbyte, and char on SQL Server


markolbert:
I have the following enum:

```
    [Flags]
    public enum Factor : long
    {
        Parent = ( 1 >> 0 ),
        PastSupporter = ( 1 >> 1 ),
    }

```

When I utilize the enum in an entity class, dotnet ef migrations add has a problem:

> dotnet : System.InvalidOperationException: The property 'Voter.Factors' is of type 'Factor' which is not supported by current database provider. Either change the 
> property CLR type or manually configure the database type for it.
> At line:1 char:1
> - dotnet ef migrations add SimplifyContactsFactors
> - ```
>   ```
>   - CategoryInfo          : NotSpecified: (System.InvalidO...se type for it.:String) [], RemoteException
>   - FullyQualifiedErrorId : NativeCommandError
>   
>   at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.PropertyMappingValidationConvention.Apply(InternalModelBuilder modelBuilder)
>   at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.ConventionDispatcher.OnModelBuilt(InternalModelBuilder modelBuilder)
>   at Microsoft.EntityFrameworkCore.Infrastructure.ModelSource.CreateModel(DbContext context, IConventionSetBuilder conventionSetBuilder, IModelValidator validator)
>   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
>   
>   at Microsoft.EntityFrameworkCore.Internal.DbContextServices.CreateModel()
>   at Microsoft.EntityFrameworkCore.Internal.LazyRef`1.get_Value()
>   
>   at Microsoft.EntityFrameworkCore.Internal.DbContextServices.get_Model()
>   
>   at Microsoft.Extensions.DependencyInjection.ServiceProvider.ScopedCallSite.Invoke(ServiceProvider provider)
>   
>   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService[T](IServiceProvider provider)
>   
>   at Microsoft.Extensions.DependencyInjection.ServiceProvider.TransientCallSite.Invoke(ServiceProvider provider)
>   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ConstructorCallSite.Invoke(ServiceProvider provider)
[...]

PR #9937, opened 2017-10-02
Use initial type conversions to map unsigned types on SQL Server

ajcvickers:
Part of #242

This is not a complete implementation of type conversions. It is enough to get conversions for ulong, uint, ushort, and sbyte mapped on SQL Server. The default mappings for these types are to their signed counterparts. We can discuss if we want to change that to something else--there are pros and cons of every conversion.

### DIFF
--- a/src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
+++ b/src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
@@ -1,12 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Extensions.Internal;
 using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Query.Expressions;
 using Microsoft.EntityFrameworkCore.Query.Internal;
@@ -247,7 +249,11 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
 
                             var readValueExpression
                                 = _entityMaterializerSource
-                                    .CreateReadValueCallExpression(targetExpression, index);
+                                    .CreateReadValueExpression(
+                                        targetExpression,
+                                        expression.Type.MakeNullable(),
+                                        index,
+                                        FindProperty(sqlExpression));
 
                             var outputDataInfo
                                 = (expression as SubQueryExpression)?.QueryModel
@@ -271,6 +277,21 @@ namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors
             }
 
             return base.Visit(expression);
+        }
+
+        private static IProperty FindProperty(Expression expression)
+        {
+            switch (expression)
+            {
+                case ColumnExpression columnExpression:
+                    return columnExpression.Property;
+                case ColumnReferenceExpression columnReferenceExpression:
+                    return FindProperty(columnReferenceExpression.Expression);
+                case UnaryExpression unaryExpression:
+                    return FindProperty(unaryExpression.Operand);
+            }
+
+            return null;
         }
     }
 }
--- a/src/EFCore.Relational/Storage/RelationalTypeMapper.cs
+++ b/src/EFCore.Relational/Storage/RelationalTypeMapper.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Metadata.Internal;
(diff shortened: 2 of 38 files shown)


ajcvickers:
@divega Should we create (re-open) a specific issue for unsigned types on SQL Server so it is easier to track in release notes, etc. that this now works?

divega:
@ajcvickers We could improve the title/contents of #6480, or create a new one. Either way works for me.

ErikEJ:
@ajcvickers providers beware?

ajcvickers:
Thanks @ErikEJ /cc @roji @caleblloyd

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -272,5 +278,20 @@ var outputDataInfo
 
             return base.Visit(expression);
         }
+
+        private static IProperty FindProperty(Expression expression)
```
anpete: cc @smitpatel

ajcvickers:
Ping? Could I get a review on the type conversions code?

src/EFCore.Relational/Storage/RelationalTypeMapper.cs
```diff
@@ -22,8 +23,8 @@ namespace Microsoft.EntityFrameworkCore.Storage
     /// </summary>
     public abstract class RelationalTypeMapper : IRelationalTypeMapper
     {
-        private readonly ConcurrentDictionary<string, RelationalTypeMapping> _explicitMappings
-            = new ConcurrentDictionary<string, RelationalTypeMapping>();
+        private readonly ConcurrentDictionary<(string StoreType, Type ClrType), RelationalTypeMapping> _explicitMappings
```
anpete: Curious why this is ConcurrentDictionary?

ajcvickers: Because it's a singleton service than currently can be added to at runtime.

src/EFCore/Storage/TypeConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Storage
+{
+    /// <summary>
+    ///     Defines conversions from an object of one type in a model to an object of the same or
+    ///     different type in the store.
+    /// </summary>
+    public class TypeConverter
```
anpete: Name conflicts with type in System.ComponentModel. Suggest renaming for clarity.

ajcvickers: CoreTypeConverter?

test/EFCore.SqlServer.FunctionalTests/Query/SimpleQuerySqlServerTest.Where.cs
```diff
@@ -495,7 +495,7 @@ public override void Where_equals_using_int_overload_on_mismatched_types()
             base.Where_equals_using_int_overload_on_mismatched_types();
 
             AssertSql(
-                @"@__shortPrm_0='1' (DbType = Int32)
+                @"@__shortPrm_0='1'
```
anpete: Intended that we no longer log the type?

ajcvickers: Yep. Previously the type of the object was different from the type expected by the parameter. Now it is the same, so we don't output it.

src/EFCore/Storage/TypeConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+            Check.NotNull(convertFromStore, nameof(convertFromStore));
+
+            ConvertToStore = convertToStore;
+            ConvertFromStore = convertFromStore;
+        }
+
+        /// <summary>
+        ///     Gets the function to convert objects when writing data to the store.
+        /// </summary>
+        public virtual Func<object, object> ConvertToStore { get; }
```
anpete: I wonder whether we should just use plain old inheritance/abstract methods here for improved dispatch/alloc perf. We could still allow lambdas as sugar at the user level, but it seems fine for providers to just provide specific conversion concrete classes - They do for everything else in Type Mapping.

anpete: Another consideration is whether the conversion methods can just be defined directly on the TypeMapping? I.e. is there an advantage to this additional factoring?

ajcvickers: @anpete I think we should leave as is for now and see how the rest of the functionality ends up building on this, then we can decide to replace the converters defined in type mappings with a different pattern if it seems like the right thing.

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
smitpatel: Taking a look at it.

src/EFCore/Storage/TypeConverter.cs
anpete: More ideas: ValueConverter, DataConverter, ValueMapper

ajcvickers: Those are good, because it's not really about converting types, it's about converting values, sometimes also to a different type. I'll do ValueConverter for now, and we can discuss again in API review.

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
```diff
@@ -247,7 +249,11 @@ var targetExpression
 
                             var readValueExpression
                                 = _entityMaterializerSource
-                                    .CreateReadValueCallExpression(targetExpression, index);
+                                    .CreateReadValueExpression(
```
smitpatel: There could be associated change in the code added by group by which also uses similar,

src/EFCore.Relational/Query/ExpressionVisitors/RelationalProjectionExpressionVisitor.cs
smitpatel: There are few options here:
1. Keep this as is.
2. Make ColumnReferenceExpression to store IProperty so that we don't need to do this.

---
In reply to: [143238944](https://github.com/aspnet/EntityFramework/pull/9937#discussion_r143238944) [](ancestors = 143238944)

ajcvickers: I'll leave as-is for now, but feel free to make this cleaner as you see fit.

ajcvickers:
Merged

PR #9939, opened 2017-10-02
[2.0.1] Use MSBuild to set NuGet feeds instead of NuGet.config

natemcmaster:
Changes
 - Use MSBuild to set NuGet feeds instead of NuGet.config. This allows us to configure feeds conditionally using MSBuild properties which is not possible with NuGet.config
 - Upgrade Microsoft.Extensions.DependencyModel to the latest 2.0.x build

FYI - CI is expected to fail until we mirror Microsoft.Extensions.DependencyModel 2.0.2-servicing-25728-02 to our feed

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,4 +1,5 @@
 <Project>
+  <Import Project="build\sources.props" />
   <Import Project="build\dependencies.props" />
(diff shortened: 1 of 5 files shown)


natemcmaster:
cref https://github.com/aspnet/Universe/pull/571

PR #9942, opened 2017-10-02
Add generic type conversion

ajcvickers:
Part of #242

This allows conversions to be defined using generics, which means they can be used without additional boxing. However, in our code we often have to deal with things like nullable FKs, null values, non-exact matches (e.g. using a short when the property is an int) and it would not be a great experience to have to define multiple conversion functions to handle all of this. Therefore, we add additional code around the simple generic function to handle these cases. In the materializer, for the common case where the type of the converter matches the type of the property, the non-boxing generic converter is used. In other cases, the compensated converter is used.


ajcvickers:
Merged

PR #9954, opened 2017-10-03
Simply default(T) using C# 7.1

smitpatel:
C# 7.1 is coming :trollface:

### DIFF
--- a/src/EFCore.Relational/Query/AsyncQueryMethodProvider.cs
+++ b/src/EFCore.Relational/Query/AsyncQueryMethodProvider.cs
@@ -158,7 +158,7 @@ namespace Microsoft.EntityFrameworkCore.Query
                 }
             }
 
-            return default(TResult);
+            return default;
         }
 
         /// <summary>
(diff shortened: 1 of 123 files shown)


ErikEJ:
Does this require a specific VS version?

smitpatel:
@ErikEJ - nothing extra from what we need already.
> The C# compiler supports C# 7.1 starting with Visual Studio 2017 version 15.3, or the .NET Core SDK 2.0. However, the 7.1 features are turned off by default. To enable the 7.1 features, you need to change the language version setting for your project.

ref: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-1

ErikEJ:
Ok. Then I suggest you update the "Getting and building the code" wiki page with that info.

smitpatel:
@ErikEJ - The opt-in is per project basis. User need to specify LangVersion in csproj to get the features. Presently, all the projects in EF Core repo has C# 7.1 enabled already. So there is nothing specific is required to do for building the code. Did you mean to update something else in the wiki?

ErikEJ:
Ah, missed that 15.3 is required already... Nothing to do then! 😄

src/EFCore.Relational/Query/AsyncQueryMethodProvider.cs
```diff
@@ -338,18 +338,20 @@ private sealed class GroupJoinAsyncEnumerator : IAsyncEnumerator<TResult>
...
-                        _nextOuter = default(TOuter);
+                        _nextOuter = default;
                     }
 
                     if (_hasNext)
                     {
                         var outer
+#pragma warning disable IDE0034 // Simplify 'default' expression - Equals(object, object) causes default(object)
                             = Equals(_nextOuter, default(TOuter))
+#pragma warning restore IDE0034 // Simplify 'default' expression
```
bricelam: Did you report this anywhere? It looks like an analyzer bug.

smitpatel: https://github.com/dotnet/roslyn/issues/22505 & https://github.com/dotnet/roslyn/issues/22184 should cover it

PR #9967, opened 2017-10-04
[2.0.1] Query: Always evalaute Enum.ToString on client

### LINKED ISSUE #9894: Enum.ToString() results in SQL translation and returns underlying integer value
After upgrading from 1.1 to 2.0, I noticed a regression in my apps. Enum.ToString() calls in a query are now evaluated in the SQL database, where enums are stored as the underlying integer type, therefore giving numerical strings instead of the enum name.

The example query:

    var dtos= _context.Models
                      .Select(model => new ModelDto
                      {
                          EnumValue = model.EnumValue.ToString()
                      });

Produces the following SQL around the `EnumValue` field:

    CONVERT(VARCHAR(11), [model].[EnumValue]) AS [EnumValue]

Using AutoMapper with defaults produces the same results.

The [SqlServerObjectToStringTranslator ](https://github.com/aspnet/EntityFrameworkCore/blob/dev/src/EFCore.SqlServer/Query/ExpressionTranslators/Internal/SqlServerObjectToStringTranslator.cs#L22) uses `VARCHAR(11)` due to it being treated as an int. Later down, the call to `UnwrapEnumType()` seems to return int instead of an enum.

The enum property in the model is annotated with `[Required]` and is part of a composite index. The behavior is still present after I added a new migration with the latest tools.

### Further technical details
EF Core version: 2.0.0
Database Provider: Microsoft.EntityFrameworkCore.SqlServer (Same behavior in LocalDB and Azure)
Operating system: Windows 10 Pro x64
IDE: Visual Studio 2017 15.3.4

smitpatel:
Since server would store underlying type of enum hence server eval causes incorrect result

Resolves #9894

### DIFF
--- a/src/EFCore.Specification.Tests/Query/GearsOfWarQueryTestBase.cs
+++ b/src/EFCore.Specification.Tests/Query/GearsOfWarQueryTestBase.cs
@@ -3966,6 +3966,23 @@ namespace Microsoft.EntityFrameworkCore.Query
             }
         }
 
+        [ConditionalFact]
+        public virtual void Enum_ToString_is_client_eval()
+        {
+            using (var context = CreateContext())
+            {
+                var query = context.Gears
+                    .OrderBy(g => g.SquadId)
+                    .ThenBy(g => g.Nickname)
+                    .Select(g => g.Rank.ToString())
+                    .Take(1)
+                    .ToList();
+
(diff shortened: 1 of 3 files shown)


ErikEJ:
sqlce beware?

PR #9989, opened 2017-10-05
Validate seed data

AndriySvyryd:
Fix issues with discriminators and table splitting

Part of #629


src/EFCore/Metadata/Builders/EntityTypeBuilder`.cs
```diff
@@ -369,7 +369,16 @@ public new virtual EntityTypeBuilder<TEntity> UsePropertyAccessMode(PropertyAcce
         ///     An array of seed data.
         /// </param>
         public virtual EntityTypeBuilder<TEntity> SeedData([NotNull] params TEntity[] data)
-            => (EntityTypeBuilder<TEntity>)base.SeedData(data);
+        {
+            foreach (var entity in data)
+            {
+                if (entity.GetType() != typeof(TEntity))
```
ajcvickers: >entity.GetType() != typeof(TEntity) [](start = 20, length = 35)

This has to be an exact match?

AndriySvyryd: yes, because any derived property will be ignored

src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -357,70 +356,7 @@ public virtual IReadOnlyList<MigrationOperation> GetDifferences(IModel source, I
...
-
-        /// <summary>
-        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
-        ///     directly from your code. This API may change or be removed in future releases.
-        /// </summary>
-        protected virtual IEnumerable<MigrationOperation> DiffSeedData()
-        {
-            foreach (var stateManager in new[] { _sourceStateManager, _targetStateManager })
-            {
-                if (stateManager == null)
```
ajcvickers: > if (stateManager == null) [](start = 15, length = 26)

When is one of the state manager's null?

AndriySvyryd: when there's no corresponding model
i.e. Starting from an empty DB or going to one

src/EFCore.Relational/Update/Internal/SharedTableEntryMapFactory.cs
```diff
@@ -0,0 +1,8 @@
+namespace Microsoft.EntityFrameworkCore.Update.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public delegate SharedTableEntryMap<TValue> SharedTableEntryMapFactory<TValue>(SharedTableEntryValueFactory<TValue> valueFactory);
+}
```
ajcvickers: Named delegates are...unusual...

AndriySvyryd: Funcs were too combersome

PR #10005, opened 2017-10-06
[2.0.x] Update how PackageReference versions are set

natemcmaster:
Changes
 - Remove floating versions
 - Disable myget feeds during a Universe build
 - Use package-specific MSBuild variables. Pattern = `packageId.Pascalize() + "PackageVersion"`, with a few exceptions.

This is an implementation of the repo-API design for automatically flowing dependencies between repos, without PackageLineup. For now this will just be on rel/2.0.x servicing builds. But we're hoping this will work for dev branch builds too in the future. (cc @mikeharder)

FYI @JunTaoLuo  - this is an example of an auto-generated change that I plan to make to all Universe repos on rel/patch 2.0.1 branches


natemcmaster:
cref https://github.com/aspnet/Universe/pull/585

build/dependencies.props
```diff
@@ -1,26 +1,32 @@
 ﻿<Project>
   <PropertyGroup>
-    <AspNetCoreVersion>2.0.0</AspNetCoreVersion>
-    <CoreFxVersion>4.4.0</CoreFxVersion>
-    <DiagnosticSourceVersion>4.4.1</DiagnosticSourceVersion>
-    <EF6Version>6.1.3</EF6Version>
-    <InternalAspNetCoreSdkVersion>2.0.1-rtm-15400</InternalAspNetCoreSdkVersion>
-    <JsonNetVersion>10.0.1</JsonNetVersion>
+    <EntityFrameworkPackageVersion>6.1.3</EntityFrameworkPackageVersion>
+    <InternalAspNetCoreSdkPackageVersion>2.0.2-beta-15516</InternalAspNetCoreSdkPackageVersion>
```
smitpatel: Does this mean that we would be building against last RTM release on that particular release cycle locally? How would this be updated if they need to be? Just curious

natemcmaster: Yes, this means you're building against the last successful build, at least at the moment. The plan is to make tools to automate updating dependencies.props. We'll build these tools before we put this on dev.

For now, if you need to work on rel/2.0.1, feel free to manually update any 2.0.1-rtm-* packages to newer builds as needed.

Also FYI - the versions in this file are overridden during a run of aspnet/Universe.

PR #10010, opened 2017-10-07
Update default SQL Server type mappings for ulong, uint, ushort, sbyte

ajcvickers:
Issues #6480, #242

Now using these as the default conversions:
* sbyte -> smallint
* ushort -> int
* uint -> bigint
* ulong -> decimal(20, 0)

These use wider columns on the server put preserve semantics. The previous mappings can be used simply by specifying the store type of the column--this is how the Northwind tests are still able to use unsigned types without changing the database.


ajcvickers:
Ping @divega

ajcvickers:
Superseded by #10079

PR #10022, opened 2017-10-09
Query: Funcletize accessing array element

### LINKED ISSUE #6859: Query: Funcletize accessing an array element
### Steps to reproduce

In SSMS:
Select \* From Person Where Name = 'marcio'
return Márcio, its ok!

In EF7
string name = "marcio";
Person person = context.Persons.FirstOrDefault(p => p.Name == name);
if(person == null)    <<< yep, person is null.... 

What I'm doing wrong?

Windows 10 Pro Insider Preview build 14951
SQL Server 2016 Latin1_General_CI_AI
WPF .Net 4.6.1
VS 2015
EF Core 1.0.1
EF SQLServer 1.0.1

smitpatel:
Resolves #6859

Array access is BinaryExpression with node type of ArrayIndex. Since we always used base method for BinaryExpression except logical expression, we would end up evaluating array separately from access hence always caused client evaluation.
Fix is to process array access as whole instead of visiting left & right subtree.

### DIFF
--- a/src/EFCore.Specification.Tests/Query/SimpleQueryTestBase.Where.cs
+++ b/src/EFCore.Specification.Tests/Query/SimpleQueryTestBase.Where.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


smitpatel:
Added tests for Dictionary/Tuples too

PR #10033, opened 2017-10-10
Infrastructure changes to allow materializer factory factories to use more information

ajcvickers:
This is a prerequisite to moving type conversions to before the value is put in the buffer rather than when it is read from the buffer. Investigation of supporting more constructor patterns and context injection showed they would likely need these changes too.

Also, ran the SQL Server tests with the typed factory and fixed several issues. This included making exceptions more consistent between typed and untyped factory, and using the property in more places to provide additional info.


src/EFCore.Relational/Storage/IRelationalValueBufferFactoryFactory.cs
```diff
@@ -34,7 +34,15 @@ public interface IRelationalValueBufferFactoryFactory
...
+        [Obsolete("Use Create(IReadOnlyList<TypeMaterializationInfo>).")]
         IRelationalValueBufferFactory Create(
             [NotNull] IReadOnlyList<Type> valueTypes, [CanBeNull] IReadOnlyList<int> indexMap);
+
+        /// <summary>
+        ///     Creates a new <see cref="IRelationalValueBufferFactory" />.
+        /// </summary>
+        /// <param name="types"> Types and mapping for the values to be read. </param>
+        /// <returns> The newly created <see cref="IRelationalValueBufferFactoryFactory" />. </returns>
+        IRelationalValueBufferFactory Create([NotNull] IReadOnlyList<TypeMaterializationInfo> types);
```
ajcvickers: @divega @anpete @AndriySvyryd @smitpatel @maumar @bricelam Opinions on this change? It's technically a breaking change to a public interface, but I don't think that anybody is directly implementing this interface--they extend from our base classes, which are not broken. However, people have hacked this area, so there may be implementations out there. We could create a new interface instead, but it's considerably more ugly.

AndriySvyryd: We could ask/check known provider writers if they implement this directly, otherwise it should be ok

anpete: I'm fine with it.

bricelam: You know my motto: *Break early, listen for feedback.*

ajcvickers:
Ping @anpete

src/EFCore.Relational/Storage/RelationalValueBufferFactoryDependencies.cs
```diff
@@ -17,5 +20,43 @@ namespace Microsoft.EntityFrameworkCore.Storage
...
+        ///     Gets the type mapper.
+        /// </summary>
+        public IRelationalTypeMapper TypeMapper { get; }
+
+        /// <summary>
+        ///     Clones this dependency parameter object with one service replaced.
+        /// </summary>
+        /// <param name="typeMapper"> A replacement for the current dependency of this type. </param>
+        /// <returns> A new parameter object with the given service replaced. </returns>
+        public RelationalModelValidatorDependencies With([NotNull] IRelationalTypeMapper typeMapper)
```
smitpatel: typo.

src/EFCore.Relational/Storage/RelationalValueBufferFactoryDependencies.cs
```diff
@@ -17,5 +20,43 @@ namespace Microsoft.EntityFrameworkCore.Storage
     /// </summary>
     public sealed class RelationalValueBufferFactoryDependencies
     {
+        /// <summary>
+        ///     <para>
+        ///         Creates the service dependencies parameter object for a <see cref="RelationalModelValidator" />.
```
smitpatel: typo

PR #10038, opened 2017-10-10
Allow tracking deleted entities with same identity.

### LINKED ISSUE #7340: Unable to replace the dependent side of an identifying FK with a new instance
I have a One-to-ZeroOrOne relationship that uses a pattern that is different to the style given in [EF Core Relationships](https://docs.microsoft.com/en-us/ef/core/modeling/relationships#other-relationship-patterns) documentation. EF Core picks up my pattern, i.e. it knows it is a one-to-one relationship, but fails on an update when there is already an entry.

I have successfully used this one-to-one relationship pattern in EF6 and, to me anyway, it seems like a good pattern as it combines the Primary Key and Foreign Key.

NOTE: This is not urgent as I can swap to your recommended style for a one-to-one relationship, but it would be useful to know if you plan to fix this, as I use this pattern in an example.

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
The instance of entity type 'PriceOffer' cannot be tracked because another instance of this type with the same key is already being tracked. When adding new entities, for most key types a unique temporary key value will be created if no key is set (i.e. if the key property is assigned the default value for its type). If you are explicitly setting key values for new entities, ensure they do not collide with existing entities or temporary values generated for other new entities. When attaching existing entities, ensure that only one entity instance with a given key value is attached to the context.
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.StartTracking(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityGraphAttacher.PaintAction(EntityEntryGraphNode node)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph(EntityEntryGraphNode node, Func`2 handleNode)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntryNotifier.NavigationReferenceChanged(InternalEntityEntry entry, INavigation navigation, Object oldValue, Object newValue)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.ChangeDetector.DetectNavigationChange(InternalEntityEntry entry, INavigation navigation)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.ChangeDetector.DetectChanges(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.ChangeDetector.DetectChanges(IStateManager stateManager)
   at test.UnitTests.DataLayer.TestOneToOneUpdate.TestConnectedUpdateExistingRelationship() in C:\Users\Jon\Documents\Visual Studio 2015\Projects\EfCoreInAction\test\UnitTests\DataLayer\TestOneToOneUpdate.cs:line 84

```

### Steps to reproduce
Include a complete code listing (or project/solution) that we can run to reproduce the issue.

Partial code listings, or multiple fragments of code, will slow down our response or cause us to push the issue back to you to provide code to repoduce the issue.

The two main entity classes are:
```c#
public class Book                                   
{
    public int BookId { get; set; }
    public DateTime PublishedOn { get; set; }
    public string Publisher { get; set; }
    public decimal Price { get; set; }
    public string ImageUrl { get; set; }
    //----------------------------------------------
    //relationships
    public PriceOffer Promotion { get; set; }       
    public ICollection<Review> Reviews { get; set; }
    public ICollection<BookAuthor> AuthorsLink { get; set; }                   
}
````
And the one-to-one entity class, PriceOffer
```c#
public class PriceOffer                        
{
    public int BookId { get; set; }            
    public decimal NewPrice { get; set; }
    public string PromotionalText { get; set; }
}
```
My DbContext contains the information to define the key for `PriceOffer`
```c#
public class EfCoreContext : DbContext
{
    public DbSet<Book> Books { get; set; }            
    public DbSet<Author> Authors { get; set; }        
    public DbSet<PriceOffer> PriceOffers { get; set; }

    public EfCoreContext( DbContextOptions<EfCoreContext> options)      
        : base(options) {}                            

    protected override void  OnModelCreating(ModelBuilder modelBuilder)    
    {                                                 
        modelBuilder.Entity<BookAuthor>()             
            .HasKey(x => new {x.BookId, x.AuthorId}); 
        modelBuilder.Entity<PriceOffer>()             
            .HasKey(x => x.BookId);                   
    }                                                 
}
```
The problem comes when I try to replace/remove a `PriceOffer` from a `Book`, i.e. it only fails if there is a `PriceOffer` on the Book already. Here is the Unit Test that fails. It fails on the line `context.SaveChanges()`.
```c#
[Fact]
public void TestConnectedUpdateExistingRelationship()
{
    //SETUP
    var inMemDb = new SqliteInMemory();

    using (var context = inMemDb.GetContextWithSetup())
    {
        context.SeedDatabaseFourBooks();
        var book = context.Books
            .Include(p => p.Promotion)
            .First(p => p.Promotion != null);

        //ATTEMPT
        book.Promotion = new PriceOffer
        {
            NewPrice = book.Price / 2,
            PromotionalText = "Half price today!"
        };
        context.SaveChanges();

        //VERIFY
        var bookAgain = context.Books
            .Include(p => p.Promotion)
            .Single(p => p.BookId == book.BookId);
        bookAgain.Promotion.ShouldNotBeNull();
        bookAgain.Promotion.PromotionalText.ShouldEqual("Half price today!");
    }
}
```
One extra piece of information. If I set the book.Promotion to null and call SaveChanges then it works, i.e. it deletes the existing `PriceOffer`. See Unit Test below that does not fail
```c#
[Fact]
public void TestDeleteExistingRelationship()
{
    //SETUP
    var inMemDb = new SqliteInMemory();

    using (var context = inMemDb.GetContextWithSetup())
    {
        context.SeedDatabaseFourBooks();

        var book = context.Books
            .Include(p => p.Promotion)
            .First(p => p.Promotion != null);

        //ATTEMPT
        book.Promotion = null;
        context.SaveChanges();

        //VERIFY
        var bookAgain = context.Books
            .Include(p => p.Promotion)
            .Single(p => p.BookId == book.BookId);
        bookAgain.Promotion.ShouldBeNull();
        context.PriceOffers.Count().ShouldEqual(0);
    }
}
```
### Further technical details
EF Core version: "version": "1.1.0"
Database Provider: "Microsoft.EntityFrameworkCore.SqlServer": "1.1.0",
         *NOTE: It also fails on "Microsoft.EntityFrameworkCore.SqlServer": "1.0.1"/NET Core "version": "1.0.1", so its not a regression.*
Operating system: Window 10
IDE: Visual Studio 2015, update 3.

AndriySvyryd:
Fixes #7340

### DIFF
--- a/src/EFCore/Update/IUpdateEntry.cs
+++ b/src/EFCore/Update/IUpdateEntry.cs
@@ -29,6 +29,11 @@ namespace Microsoft.EntityFrameworkCore.Update
         EntityState EntityState { get; }
 
         /// <summary>
+        ///     The other entry that has the same key values, if one exists.
+        /// </summary>
+        IUpdateEntry SharedIdentityEntry { get; }
+
+        /// <summary>
         ///     Gets a value indicating if the specified property is modified. If true, the current value assigned
         ///     to the property should be saved to the database.
         /// </summary>
--- a/src/EFCore.InMemory/Storage/Internal/InMemoryStore.cs
+++ b/src/EFCore.InMemory/Storage/Internal/InMemoryStore.cs
@@ -138,6 +138,16 @@ namespace Microsoft.EntityFrameworkCore.Storage.Internal
                         _tables.Value.Add(key, table = _tableFactory.Create(entityType));
                     }
 
+                    if (entry.SharedIdentityEntry != null)
+                    {
+                        if (entry.EntityState == EntityState.Deleted)
+                        {
+                            continue;
+                        }
+
+                        table.Delete(entry);
+                    }
+
                     switch (entry.EntityState)
                     {
                         case EntityState.Added:
--- a/src/EFCore.Relational.Specification.Tests/TableSplittingTestBase.cs
+++ b/src/EFCore.Relational.Specification.Tests/TableSplittingTestBase.cs
@@ -9,14 +9,20 @@ using Microsoft.EntityFrameworkCore.TestUtilities;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Xunit;
+using Xunit.Abstractions;
 
 // ReSharper disable InconsistentNaming
 namespace Microsoft.EntityFrameworkCore
 {
     public abstract class TableSplittingTestBase
     {
+        protected TableSplittingTestBase(ITestOutputHelper testOutputHelper)
+        {
+            TestOutputHelper = testOutputHelper;
+        }
+
         [Fact(Skip = "#8973")]
-        public void Can_query_shared()
+        public virtual void Can_query_shared()
         {
             using (CreateTestStore(OnModelCreating))
             {
@@ -28,7 +34,7 @@ namespace Microsoft.EntityFrameworkCore
         }
 
         [Fact(Skip = "#8973")]
-        public void Can_query_shared_derived()
+        public virtual void Can_query_shared_derived()
         {
             using (CreateTestStore(OnModelCreating))
             {
@@ -40,13 +46,13 @@ namespace Microsoft.EntityFrameworkCore
         }
 
         [Fact(Skip = "#8973")]
-        public void Can_use_with_redundant_relationships()
+        public virtual void Can_use_with_redundant_relationships()
         {
             Test_roundtrip(OnModelCreating);
         }
 
         [Fact(Skip = "#8973")]
-        public void Can_use_with_chained_relationships()
+        public virtual void Can_use_with_chained_relationships()
         {
             Test_roundtrip(
                 modelBuilder =>
@@ -57,7 +63,7 @@ namespace Microsoft.EntityFrameworkCore
         }
 
         [Fact(Skip = "#8973")]
-        public void Can_use_with_fanned_relationships()
+        public virtual void Can_use_with_fanned_relationships()
         {
             Test_roundtrip(
                 modelBuilder =>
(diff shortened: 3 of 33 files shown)


ajcvickers:
What happens if I replace an owned entity more than once without saving?

AndriySvyryd:
@ajcvickers Can you review this again?
I changed the approach to having the update pipeline handle the diffing of the new instance with the old one. This allows it to handle discriminators and other readonly values properly, only send the update values and make it easier to handle entities with complex mappings in the future

src/EFCore/Update/IUpdateEntry.cs
```diff
@@ -29,6 +29,11 @@ public interface IUpdateEntry
         EntityState EntityState { get; }
 
         /// <summary>
+        ///     The other entry that has the same key values, if one exists.
+        /// </summary>
+        IUpdateEntry SharedIdentityEntry { get; }
```
ajcvickers: Breaking change? Will providers need to react if they have custom update-pipeline code?

AndriySvyryd: This interface is meant for consumption, not implementation.
Non-relational providers would need to react to support this, but it won't break anything that was working before.

PR #10039, opened 2017-10-11
Fix #9923 - Improved double dispose fix to not use AsyncLocal.

### LINKED ISSUE #9923: Perf Regression: AddDbContextPool in asp.net core System.InvalidOperationException
"AddDbContextPool in asp.net core System.InvalidOperationException" (https://github.com/aspnet/EntityFrameworkCore/pull/9871) seems to have caused a 30% regression in the TechEmpower "Single Query" and "Fortunes" scenarios, on 9/21/17 between 7:45PM and 8:28PM PST.

Scenario | RPS @ 9/21 7:45 PM | RPS @ 9/21 8:28 PM | Ratio
------------ | ------------- | ------------- | -------------
DbSingleQueryEf | 11,243 | 7,900 | 0.70
DbFortunesEf | 10,355 | 6,833 | 0.66

You can also see the regression in our charts at http://aka.ms/aspnet/benchmarks.

anpete:
Verified on TE benchmarks by @mikeharder.

### DIFF
--- a/src/EFCore/DbContext.cs
+++ b/src/EFCore/DbContext.cs
@@ -63,8 +63,6 @@ namespace Microsoft.EntityFrameworkCore
         private IDbContextPool _dbContextPool;
         private bool _initializing;
         private bool _disposed;
-
-        private AsyncLocal<bool> _leased;
         
         /// <summary>
         ///     <para>
@@ -281,7 +279,7 @@ namespace Microsoft.EntityFrameworkCore
         [DebuggerStepThrough]
         internal void CheckDisposed()
         {
-            if (_disposed || _leased?.Value == false)
+            if (_disposed)
             {
                 throw new ObjectDisposedException(GetType().ShortDisplayName(), CoreStrings.ContextDisposed);
             }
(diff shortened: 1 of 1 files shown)


src/EFCore/DbContext.cs
```diff
@@ -536,12 +542,13 @@ var resettableServices
         public virtual void Dispose()
         {
             if (_dbContextPool == null
-                || _leased.Value
+                || !ReferenceEquals(_lastExecutionContext, Thread.CurrentThread.ExecutionContext)
```
mikeharder: I think it's technically better to just use `_lastExecutionContext == Thread.CurrentThread.ExecutionContext`, and let the `==` operator decide if they are equal.  According to docs, `ExecutionContext` currently does not override `==` or `Equals()` so they should be equivalent to `ReferenceEquals()`, but in theory this could change in the future.

mikeharder: Should this check be inverted?  Don't we want to only call `_dbContextPool.Return()` if we're on the *same* thread which called `ResetState()`?

PR #10040, opened 2017-10-11
Fix #9956 - Add support for private and protected methods in DbFunctions

### LINKED ISSUE #9956: DbFunctions: Consider protected methods for discovering DbFunctionAttribute
At present we have set of BindingFlags to discover methods for UDFs, they explicitly ignores protected methods. But the attribute based discovery should be done regardless of protection level.
The protection level restricts user's ability to use the method inside of the EF Query but if in some way user can use it then we should translate it.
One way to use the method is to pass in IQueryable to public method in DbContext which would apply logics from protected/private methods by calling them and return it. (or there is always reflection)

pmiddleton:
Resolves #9956

Adds support for private, protected, internal, and protected internal

### DIFF
--- a/src/EFCore.Relational/Infrastructure/RelationalModelCustomizer.cs
+++ b/src/EFCore.Relational/Infrastructure/RelationalModelCustomizer.cs
@@ -72,13 +72,20 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
             Check.NotNull(modelBuilder, nameof(modelBuilder));
             Check.NotNull(context, nameof(context));
 
-            var functions = context.GetType().GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static | BindingFlags.FlattenHierarchy)
-                .Where(mi => mi.IsPublic
-                          && mi.GetCustomAttributes(typeof(DbFunctionAttribute)).Any());
+            var contextType = context.GetType();
 
-            foreach (var function in functions)
-            {
-                modelBuilder.HasDbFunction(function);
+            while(contextType != typeof(DbContext))
+            { 
+                var functions = contextType.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance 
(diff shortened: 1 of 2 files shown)


src/EFCore.Relational/Infrastructure/RelationalModelCustomizer.cs
```diff
@@ -72,13 +72,20 @@ protected virtual void FindDbFunctions([NotNull] ModelBuilder modelBuilder, [Not
...
-                          && mi.GetCustomAttributes(typeof(DbFunctionAttribute)).Any());
+            var contextType = context.GetType();
 
-            foreach (var function in functions)
-            {
-                modelBuilder.HasDbFunction(function);
+            while(contextType != typeof(DbContext))
+            { 
+                var functions = contextType.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance 
+                                                                | BindingFlags.Static | BindingFlags.DeclaredOnly)
```
smitpatel: > BindingFlags.DeclaredOnly to search only the methods declared on the Type, not methods that were simply inherited. 

I believe this should be different. When working with multi-context scenarios (using different contexts for read/write), user may want to just put method on common base instead of the exact context instance.

Also FlattenHierarchy should be there to find static members from base classes.

pmiddleton: The while loop will walk up the inheritance chain pulling methods until it gets to DbContext.

We need to do this because base class private methods are not returned when methods are pulled from derived types.  

FlattenHierarchy is disabled to prevent duplicate static methods from being returned during the inheritance walk.

smitpatel: Ah. didn't realize the loop.

smitpatel:
@pmiddleton - If you get some time then can you rebase this on latest dev?

smitpatel:
Merged via 843929da0c1bc90813178e8fa16edf9dbbeb58df
Thanks for contribution.

PR #10050, opened 2017-10-11
Fix #9923 - Perf Regression: AddDbContextPool in asp.net core System.InvalidOperationException

### LINKED ISSUE #9923: Perf Regression: AddDbContextPool in asp.net core System.InvalidOperationException

anpete:
Introduces a "lease" object, which is a scoped object that manages ownership of a pooled context instance. In particular, the lease object is responsible for returning instances to the pool at the end of the scope. This removes the complexity/race condition from DbContext.Dispose and greatly simplifies pooling in general.

### DIFF
--- a/src/EFCore/DbContext.cs
+++ b/src/EFCore/DbContext.cs
@@ -63,8 +63,6 @@ namespace Microsoft.EntityFrameworkCore
         private IDbContextPool _dbContextPool;
         private bool _initializing;
         private bool _disposed;
-
-        private AsyncLocal<bool> _leased;
         
         /// <summary>
         ///     <para>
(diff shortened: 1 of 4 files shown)


smitpatel:
CI failures.

PR #10075, opened 2017-10-13
Migrations: Use reflection order for columns in CreateTable

### LINKED ISSUE #2272: Column ordering

bricelam:
Fixes #2272

### DIFF
--- a/src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
+++ b/src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Reflection;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
 using Microsoft.EntityFrameworkCore.Infrastructure;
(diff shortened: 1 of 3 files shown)


src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs
```diff
@@ -605,6 +606,117 @@ protected virtual IEnumerable<MigrationOperation> Remove([NotNull] string source
...
+                types.GetOrAddNew(clrType)[index] = clrProperty;
+            }
+
+            foreach (var group in unorderedGroups)
+            {
+                groups.Add(group.Key, group.Value.Values.ToList());
+            }
+
+            foreach (var definingForeignKey in entityType.GetDeclaredReferencingForeignKeys()
+                .Where(fk => fk.DeclaringEntityType.DefiningEntityType == entityType))
```
bricelam: @AndriySvyryd is this this right? In Relational, do navigation-defined entity types always map to the principal entity type's table?

AndriySvyryd: No, that's just the default.

bricelam: K, will add a test and react

bricelam:
The more code I write, the less confident I become. 😄 I'm doing one last round of bashing on this before I merge...

Issue #9711, opened 2017-09-06
Relational: Investigate OracleRelationalCommandBuilderFactory "hacks" in Oracle sample


anpete:
Some ad hoc type mapping was required in OracleRelationalCommandBuilderFactory. We should make corresponding improvements to Relational/Core to remove the need for these changes.

PR #10076, opened 2017-10-14
Add non-hacked Oracle type mappings for bool, DateTimeOffset, and Guid

ajcvickers:
Part of issues #9711 and #242

Guid and bool required just type conversions (and some fixes in the product code).

DateTimeOffset cannot be read from the store without using an Oracle-specific call on the DbDataReader. To enable this there is now an IDataReaderMethodProvider service. This can then be used with a type conversion to get the correct behavior.

There are still some hacks remaining for other areas--specifically, the cursor parameter hack and some SQL manipulation.


samples/OracleProvider/src/OracleProvider/Storage/Internal/OracleDataReaderMethodProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+using JetBrains.Annotations;
+using Oracle.ManagedDataAccess.Client;
+using Oracle.ManagedDataAccess.Types;
+
+namespace Microsoft.EntityFrameworkCore.Storage.Internal
+{
+    public class OracleDataReaderMethodProvider : DefaultDataReaderMethodProvider
+    {
+        private static readonly IDictionary<Type, MethodInfo> _getXMethods
+            = new Dictionary<Type, MethodInfo>
```
anpete: YAGNI dictionary?

samples/OracleProvider/src/OracleProvider/Storage/Internal/OracleRelationalCommandBuilderFactory.cs
```diff
@@ -122,7 +122,7 @@ private static string AdjustCommandText(string commandText)
                                     = new RelationalDataReader(
                                         connection,
                                         dbCommand,
-                                        new DbDataReaderDecorator(dbCommand.ExecuteReader()),
+                                        dbCommand.ExecuteReader(),
```
anpete: 🎉

samples/OracleProvider/src/OracleProvider/Storage/Internal/OracleDataReaderMethodProvider.cs
ajcvickers: Yes. but as this is the sample I was showing a pattern and making it easy for people to just add new methods as needed. However, I can change if you prefer.

anpete: Makes sense

src/EFCore.Relational/Storage/IDataReaderMethodProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public interface IDataReaderMethodProvider
+    {
+        /// <summary>
+        ///     The method to use when reading values of the given type. The method must be defined
+        ///     on <see cref="DbDataReader" /> or one of its subclasses.
+        /// </summary>
+        /// <param name="type"> The type of the value to be read. </param>
+        /// <returns> The method to use to read the value. </returns>
+        MethodInfo GetReadMethod([NotNull] Type type);
```
anpete: Any reason to not just have this on RelationalTypeMapper?

src/EFCore/Query/ExpressionVisitors/Internal/ValueConverterExpressionVisitor.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Linq.Expressions;
+using JetBrains.Annotations;
+
+namespace Microsoft.EntityFrameworkCore.Query.ExpressionVisitors.Internal
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public class ValueConverterExpressionVisitor : ExpressionVisitorBase
```
anpete: Can remove this: See https://github.com/aspnet/EntityFrameworkCore/blob/61952a43a17785490c041a71ba573455a04f94b2/src/EFCore/Query/ExpressionVisitors/Internal/FilterApplyingExpressionVisitor.cs#L74

src/EFCore.Relational/Storage/IDataReaderMethodProvider.cs
ajcvickers: Hmm, I could try it. I kind of like small simple services, but it is part of type mapping, so it could be an optional method on the type mapper.

anpete: Could be easier for providers to do it in "one place".

src/EFCore/Query/ExpressionVisitors/Internal/ValueConverterExpressionVisitor.cs
ajcvickers: Nice. 👍 (I was thinking that there might be something that did this already, but I failed to find it.)

src/EFCore.Relational/Storage/IDataReaderMethodProvider.cs
ajcvickers: Yep. I'll change it.

Issue #1381, opened 2015-01-09
Enums Stored as Strings


motowilliams:
Will EF7 have optional (or default) support for storing the enum string values instead of the numeric values in the database?

rowanmiller:
The default will still be to store the underlying type of the enum. Swapping to use the string will be covered by https://github.com/aspnet/EntityFramework/issues/242

CharlesToniolo:
Is there any workaround to make it work with we have for now (RC1 or RC2)?
We are using automapper to translate these strings to enum (and enum to string), but doing this, almost of our queries have AsNoTracking and it is leading us in some troubles.

divega:
You could do the conversion in code in the entity and map the underlying string property using EF (you can make the property private and still map it) and leave the enum property unmapped (e.g. with `[NotMapped]`).

CharlesToniolo:
That is exactly what i thought. Thanks!

jeanmallet:
Worked fine for me. Thank you **divega**.

DantaliaN00:
How about LINQ?

I say about something like

```
enum MyEnum {
  SomeEnum
}

class DbEntity {
  [Column("my_val")]
  public MyEnum MyDbVal {get;set;}
}

//in DbContext
DbSet<DbEentity> Items {get;set;}

//and somewhere
db.Items.Single(i => i.MyDbVal == MyEnum.SomeEnum)

```

I know, I can use Mapped string and NotMapped Enum, but it's realy ugly with linq.

I actualy want to see strings in equation and in DB for many reasons
Maybe EF has some properties/options/settings about that?

divega:
@DantaliaN00 server-side querying based on equality should be enabled for the majority of mappings by the type conversion feature described in #242. We will start working on that feature soon, although we are planning to ship it after 1.1.

ajcvickers:
In 2.1, mapping an enum property to a string property will result in the enums being saved as strings in the database.

roji:
@ajcvickers, any chance that native database enums could be supported as well (i.e. #3620)? This would be great for PostgreSQL.

ajcvickers:
@roji Yes, but there are some more changes needed that have not yet been completed. They should be done relatively soon (depending on how much time I get to write code). The general idea is that:
* If the provider's type mapper now says that it supports the enum natively, then EF will not attempt to convert it
* If some other value converter has been supplied for the enum, then EF will use that
[...]

PR #10088, opened 2017-10-16
Use value converters for enums and allow enums to map to string columns

ajcvickers:
Part of issues #242 and #1381

Instead of using special code for materialization and parameter creation, type mappings now add an appropriate type converter for enum types that converts two and from the underlying numeric value. (Special code is still present for providers that are not updated and do not create mappings with converters.)

This means that if the column is instead a string column, then a different converter can be used which converts to and from the string values of the enum. SQL Server and most other providers will get this for free when updating to 2.0 and adding support for value conversions. SQLite is the exception since it has weird type mapping.


src/EFCore/Storage/Internal/EnumToStringConveter.cs
```diff
@@ -0,0 +1,33 @@
...
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public EnumToStringConveter()
+            : base(v => v.ToString(), v => ConvertToEnum(v))
+        {
+        }
+
+        private static TEnum ConvertToEnum(string value)
+            => Enum.TryParse<TEnum>(value, out var result)
```
AndriySvyryd: I suspect some users will want an option to ignore case.

ajcvickers: Good point. I'll add it.

Issue #9270, opened 2017-07-27
Set a default max batch size for SQL Server


divega:
@smitpatel ran a few custom batching benchmarks recently that showed interesting results. His benchmarks focuses on INSERT batching for the same entity, but similar analysis could be performed for other operations.

The test inserts 1000 rows on a table in SQL Server. There are two separate series, one for a local instance and one for a remote instance. The chart shows the time elapsed (on Y) for different batch sizes (on X):

![image](https://user-images.githubusercontent.com/1253414/28650948-ca42613a-7233-11e7-85df-916c47ce7375.png)

To make sense if this chart, keep in mind that at maximum batch size of 1, there is now batching, and we resort to different SQL that has a smaller fixed cost.

There are a few interesting things to observe:

1. For a local instance (i.e. low latency) the benefits of batching are very limited and in fact there only seem to be some improvement between 20 and 30 for batch size.  
2. For both local an remote things start to get worse after batch size of 40
3. A batch size of 2 seems to be a very bad idea, even for a remote (i.e. higher latency) instance.

Possible conclusions and follow up actions:

1. Currently if batch size is not specified we default to large batches limited only by the maximum number of parameters for a SQL statement which is 2100. It seems that we should pick a smaller default, e.g. 20.
2. There seems to be potential value in a minimum batch size setting, e.g. do not do batching unless there can be benefit. The threshold for the remote run seems to be 4, and for the local run closer to 20.
3. We should try to understand how this behaves for other operations. 
4. We should try to understand how this is sensitive to number and types of columns, and in general to the size of the data.
5. If possible we should experiment a bit with concatenated INSERTs as a strategy for batching INSERTs and compare the results. 
5. If possible we should experiment with higher latency database connections. Our current “remote” is not very remote.
6. Ultimately it would be cool if batch size could be adaptive :smile:

The code of the test follows:

``` C#
using System;
using System.Collections.Generic;
using System.Diagnostics;
using Microsoft.EntityFrameworkCore;
using Microsoft.Extensions.Logging;

namespace ConsoleApp6
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");

            new Test().Run();
        }

    }

    public class Test
    {
        protected OrdersContext Context;
        protected string ConnectionString = "Server=(localdb)\\mssqllocaldb;Database=Benchmarks;Trusted_Connection=True;Database=Perf_UpdatePipeline_Simple";

        public void Run()
        {
[...]

PR #10091, opened 2017-10-16
Add MinBatchSize option

AndriySvyryd:
Part of #9270

### DIFF
--- a/src/EFCore.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
+++ b/src/EFCore.Relational/Infrastructure/RelationalDbContextOptionsBuilder.cs
@@ -47,6 +47,15 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
             => WithOption(e => (TExtension)e.WithMaxBatchSize(maxBatchSize));
 
         /// <summary>
+        ///     Configures the minimum number of statements that are needed for a multi-statement command sent to the database
+        ///     during <see cref="DbContext.SaveChanges()" />.
+        /// </summary>
+        /// <param name="minBatchSize"> The minimum number of statements. </param>
+        /// <returns> The same builder instance so that multiple calls can be chained. </returns>
+        public virtual TBuilder MinBatchSize(int minBatchSize)
+            => WithOption(e => (TExtension)e.WithMinBatchSize(minBatchSize));
+
+        /// <summary>
         ///     Configures the wait time (in seconds) before terminating the attempt to execute a command and generating an error.
         /// </summary>
         /// <param name="commandTimeout"> The time in seconds to wait for the command to execute. </param>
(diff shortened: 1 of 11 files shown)


roji:
I'm synchronizing some old EF Core changes to Npgsql, and came across this issue. I can see that MinBatchSize defaults to a hardcoded 4 in `CommandBatchPreparer`, any particular reasoning for this? I saw some benchmarks in #9270, but those are SQL Server-specific and CommandBatchPreparer is used by other providers such as Npgsql.

Concretely I don't have reason to believe it's not a good idea to start batching from 2 statements (and no time to actually benchmark). Any chance you move the constant 4 to some provider-specific place where it would be easier to change? Otherwise I have to set it in NpgsqlOptionsExtension which isn't very nice.

AndriySvyryd:
@roji Why isn't it nice to do in NpgsqlOptionsExtension? Where would you want to do it?
 For Sql Server 3 and 4 are very similar, so we could lower the default if benchmarks show 3 is better for other providers. I suspect that 3 would still be better than 2 for Npgsql

roji:
My point was that I'd like to leave MinBatchSize null on NpgsqlOptionsExtension, to signify that the user hasn't specified it. If I set it to some value such as 2, we can no longer distinguish between a user-set value and a default value... Or maybe I misunderstood you suggestion...

Regarding 2 vs 3 vs 4, I should benchmark. However, on the wire, sending two messages as a batch or as two separate messages involved the exact same protocol messages, the only difference is in whether you wait for the first statement to complete before sending the second. So there really is no reason to think that batching would be bad for performance for any batch size...

I understand this works different for SQL Server, which is why I think it's better to manage the default separately in each provider.

AndriySvyryd:
I just can't think of a scenario where it would matter that the value was set by the user.

There is some overhead in EF when creating batches.

roji:
> I just can't think of a scenario where it would matter that the value was set by the user.

For example, when logging we could check whether the value was set by the user, and only only log if so. It's just a theoretical idea that the distinction is important, if we really don't care I can set it.

> There is some overhead in EF when creating batches.

Are you sure the slowdown you're seeing in the benchmarks comes from EF Core rather than the ADO.NET driver itself? Because in my experience the network round-trip you add by not batching dominates most things happening inside your application (especially as latency to your database server increases...)

AndriySvyryd:
> For example, when logging we could check whether the value was set by the user, and only only log if so. It's just a theoretical idea that the distinction is important, if we really don't care I can set it.

@divega, @ajcvickers ?

> Because in my experience the network round-trip you add by not batching dominates most things happening inside your application

Yes, that's why this has to be configurable. For the default I think we should consider the co-located scenario (<10ms latency)

roji:
>> Because in my experience the network round-trip you add by not batching dominates most things happening inside your application
>
> Yes, that's why this has to be configurable. For the default I think we should consider the co-located scenario (<10ms latency)

I agree, but the point is that in PostgreSQL I don't think there's any real latency low enough where not batching wouldn't make sense, even against localhost. I know it's bad to talk about perf in theory, but I have no real environment here to benchmark. Is there any chance you guys can run a quick benchmark with varying batch sizes against PostgreSQL? If not I'll at least run a localhost benchmark.

PR #10100, opened 2017-10-17
Update AppServiceProviderFactory to use WebHostFactoryResolver

javiercn:
Uses the new shared sources package we created to abstract the aspnet templates host initialization logic. It finds either
IWebHostBuilder CreateWebHostBuilder(string[] args) or
IWebHost BuildWebHost(string[] args)
and provides a factory method that can be used to produce an IWebHost instance

### DIFF
--- a/src/EFCore.Design/Design/Internal/AppServiceProviderFactory.cs
+++ b/src/EFCore.Design/Design/Internal/AppServiceProviderFactory.cs
@@ -3,9 +3,11 @@
 
 using System;
 using System.Reflection;
+using System.Diagnostics;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.AspNetCore.Hosting.WebHostBuilderFactory;
 
 namespace Microsoft.EntityFrameworkCore.Design.Internal
 {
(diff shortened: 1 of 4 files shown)


src/EFCore.Design/Design/Internal/AppServiceProviderFactory.cs
```diff
@@ -44,20 +45,20 @@ private IServiceProvider CreateFromBuildWebHost(string[] args)
...
+                    break;
+                case FactoryResolutionResultKind.NoEntryPoint:
+                    _reporter.WriteVerbose(DesignStrings.NoEntryPoint(_startupAssembly.GetName().Name));
+                    return null;
+                case FactoryResolutionResultKind.NoCreateWebHostBuilder:
+                case FactoryResolutionResultKind.NoBuildWebHost:
+                    _reporter.WriteVerbose(DesignStrings.NoBuildWebHost(webHostFactoryResult.ProgramType.DisplayName()));
+                    return null;
+                default:
+                    throw new InvalidOperationException("Can't get here.");
```
bricelam: I prefer this pattern:
``` csharp
Debug.Fail("Unexpected value: " + webHostFactoryResult.ResultKind);
return null;
```

src/EFCore.Design/Design/Internal/AppServiceProviderFactory.cs
```diff
@@ -44,20 +45,20 @@ private IServiceProvider CreateFromBuildWebHost(string[] args)
         {
             _reporter.WriteVerbose(DesignStrings.FindingBuildWebHost);
 
-            var programType = FindProgramClass();
-            if (programType == null)
+            var webHostFactoryResult = WebHostFactoryResolver.ResolveWebHostFactory<object, object>(_startupAssembly);
+            switch (webHostFactoryResult.ResultKind)
```
bricelam: It may not be important to preserve all these messages. If you think it's more maintainable to replace them with a generic "can't find web host factory" message, I won't push back.

Issue #10109, opened 2017-10-18
Better LINQ support for Microsoft.SqlServer.Types


ajcvickers:
Methods like STDistance are not easy to translate right now because they are t-sql "instance" methods called like this:
```sql
SELECT [Id], [Geography1].STDistance([Geography2]) as [Distance] From [Foos]
```

Function registration and method call translation does not support this out-of-the-box, but it shouldn't be too hard to add support. Once this is done, we should consider whether to write the common translations and ship them somewhere or maybe publish them as a sample.

ajcvickers:
@smitpatel Will do the initial work to support method call translation to this kind of function, and then pass of to @ajcvickers for testing.

smitpatel:
Assigning to @ajcvickers since query related work is finished.

ajcvickers:
Notes for triage: I was able to map t-sql functions and use them.

Functions look like:
```C#
public static class SqlGeographyExtensions
{
    public static bool Contains(this SqlGeography geography, SqlGeography otherGeography) 
        => throw new NotImplementedException();
    
    public static double Distance(this SqlGeography geography, SqlGeography otherGeography) 
        => throw new NotImplementedException();
}
```
Function mapping looks like:
```C#
protected override void OnModelCreating(ModelBuilder modelBuilder)
{
    modelBuilder.HasDbFunction(
        typeof(SqlGeographyExtensions).GetMethod(nameof(SqlGeographyExtensions.Contains)),
        b => b.HasTranslation(e =>
            new SqlFunctionExpression(e.First(), "STContains", typeof(bool), e.Skip(1))));

    modelBuilder.HasDbFunction(
        typeof(SqlGeographyExtensions).GetMethod(nameof(SqlGeographyExtensions.Distance)),
        b => b.HasTranslation(e =>
            new SqlFunctionExpression(e.First(), "STDistance", typeof(double), e.Skip(1))));
}
```
Usages:
```C#
var redmond = context.Cities.Single(c => c.Name == "Redmond");
var housesInRedmond = context.Houses.Where(h => redmond.Shape.Contains(h.Location)).ToList();
```
```C#
foreach (var house in context.Houses)
{
    foreach (var distanceInfo in context.Houses.Select(h => new
    {
        Name = h.Name,
        Distance = h.Location.Distance(house.Location)
    }))
    {
    }
}
```

Full sample:
```C#
using System;
using System.Data.SqlTypes;
using System.Linq;
[...]

Issue #10108, opened 2017-10-18
Figure out testing for EF use of Microsoft.SqlServer.Types


ajcvickers:
Testing for Microsoft.SqlServer.Types is problematic because:
* The types are not available on .NET Core
* Setting up a machine to have the correct versions of the correct types in the correct places is fragile and has traditionally causes C.I./build issues.

For now I have been using this little test app for most of my functional testing:
```C#
public class Program
{
    public static void Main()
    {
        var loggerProvider = new MyLoggerProvider();

        using (var context = new TestDbContext())
        {
            // Create UDT columns
            context.Database.EnsureDeleted();
            context.GetService<ILoggerFactory>().AddProvider(loggerProvider);
            context.Database.EnsureCreated();

            // Single save
            context.AddRange(NewFoo(0));
            context.SaveChanges();

            // Batch save
            context.AddRange(NewFoo(1), NewFoo(2));
            context.SaveChanges();
        }

        using (var context = new TestDbContext())
        {
            // Simple query
            var foos = context.Foos.OrderBy(e => e.Id).ToList();
            AssertFoo(foos[0], 0);
            AssertFoo(foos[1], 1);
            AssertFoo(foos[2], 2);

            // Equals translation works for HierarchyId but not for spatial types
            var foo = context.Foos.Single(e => e.HierarchyId.Equals(SqlHierarchyIds[1]));
            AssertFoo(foo, 1);

            // Simple prjection
            var projection = context.Foos.OrderBy(e => e.Id).Skip(1).Select(e => new {foo.Id, foo.Geography1}).First();
            Debug.Assert(projection.Geography1.STEquals(SqlGeographies[1]).IsTrue);

            // FromSql query using spatial function
            var bars = context.Bars.FromSql("SELECT [Id], [Geography1].STDistance([Geography2]) as [Distance] From [Foos]").OrderBy(e => e.Id).ToList();
            AssertDistance(bars[0].Distance, 0);
            AssertDistance(bars[1].Distance, 1);
            AssertDistance(bars[2].Distance, 2);

            // Manipulate and send database updates
            for (var i = 0; i < 3; i++)
            {
                foos[i].Geography1 = SqlGeographies[(i + 1) % 3];
                foos[i].Geography2 = SqlGeographies[(i + 2) % 3];
                foos[i].Geometry = SqlGeometries[(i + 1) % 3];
                foos[i].HierarchyId = SqlHierarchyIds[(i + 1) % 3];
            }
[...]

PR #10110, opened 2017-10-18
Allow mapping to SqlGeography, SqlGeometry, and HierarchyId columns on .NET Framework

ajcvickers:
Part of issue #1100

This change allows properties of type SqlGeography, SqlGeometry, and HierarchyId to be included in entity classes and mapped by convention to corresponding database columns.

WARNING: This is by no-means full spatial support for EF Core. These are the important limitations:
* It only works when running on .NET Framework. It will not work on .NET Core, since SqlClient for .NET Core does not support these types.
* Query support is limited:
  * LINQ queries that don't use the semantics of the SQL types should work.
  * Queries that need to use the type semantics (e.g. using STDistance) can be done with FromSql queries, but not LINQ
  * See issue #10108 for some test code and issue #10109 for some ideas to make LINQ work better
* The SQL Server spatial types do not provide a good client-side programming experience. A proper .NET spatial library is needed for a good, cross-platform experience.

Note that this change also allows more general mapping of types not directly referenced from the provider, and more specifically UDTs on SQL Server.

### DIFF
--- a/src/EFCore.Relational/Storage/RelationalTypeMapper.cs
+++ b/src/EFCore.Relational/Storage/RelationalTypeMapper.cs
@@ -26,6 +26,9 @@ namespace Microsoft.EntityFrameworkCore.Storage
     /// </summary>
     public abstract class RelationalTypeMapper : IRelationalTypeMapper
     {
+        private static readonly IReadOnlyDictionary<string, Func<Type, RelationalTypeMapping>> _emptyNamedMappings
+            = new Dictionary<string, Func<Type, RelationalTypeMapping>>();
+
         private static readonly MethodInfo _getFieldValueMethod
             = typeof(DbDataReader).GetTypeInfo().GetDeclaredMethod(nameof(DbDataReader.GetFieldValue));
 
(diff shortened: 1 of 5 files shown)


src/EFCore.Relational/Storage/RelationalTypeMapper.cs
```diff
@@ -154,11 +164,23 @@ public virtual RelationalTypeMapping FindMapping(Type clrType)
...
-                        ? (RelationalTypeMapping)mapping.Clone(CreateEnumToNumberConverter(k.ClrType))
-                        : mapping)
-                    : null);
+                k =>
+                    {
+                        var underlyingType = k.ClrType.UnwrapEnumType();
+
+                        if (!GetClrTypeMappings().TryGetValue(underlyingType, out var mapping))
+                        {
+                            if (GetClrTypeNameMappings().TryGetValue(underlyingType.FullName, out var mappingFunc))
```
anpete: Merge ifs

Xtrophic:
I tried it out and found out that sending null geography / geometry fields throws an exception:
ArgumentException: UdtTypeName property must be set only for UDT parameters.

Other than that it seems to work with CRUD ops.

ajcvickers:
@Xtrophic Thanks. I filed #10214 to track that issue.

PR #10136, opened 2017-10-22
Add GenerateCreateScript extension method on DatabaseFacade

### LINKED ISSUE #2943: Equivalent to ObjectContext.CreateDatabaseScript
This is one of my favorite features of the EF Power Tools, and I think we should have similar functionality in EF Core.

My proposal is an `db.Database.GenerateCreateScript()` extension method. Here's the implementation. :wink:

``` C#
public static string GenerateCreateScript(this DatabaseFacade database)
{
    var model = database.GetService<IModel>();
    var differ = database.GetService<IMigrationsModelDiffer>();
    var generator = database.GetService<IMigrationsSqlGenerator>();
    var sql = database.GetService<ISqlGenerationHelper>();

    var operations = differ.GetDifferences(null, model);
    var commands = generator.Generate(operations, model);

    var builder = new StringBuilder();
    foreach (var command in commands)
    {
        builder
            .Append(command.CommandText)
            .AppendLine(sql.BatchTerminator);
    }

    return builder.ToString();
}
```

ErikEJ:
fixes #2943

- also fixes a test localization bug causing test failure on non US systems.

### DIFF
--- a/src/EFCore.Relational/Storage/RelationalDatabaseCreator.cs
+++ b/src/EFCore.Relational/Storage/RelationalDatabaseCreator.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Transactions;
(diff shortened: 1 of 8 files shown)


src/EFCore.Relational/Storage/RelationalDatabaseCreator.cs
```diff
@@ -265,5 +266,28 @@ public virtual bool EnsureCreated()
...
+        /// </returns>
+        public virtual string GenerateCreateScript(string batchTerminator)
+        {
+            var commands = GetCreateTablesCommands();
+            var builder = new StringBuilder();
+            foreach (var command in commands)
+            {
+                builder
+                    .Append(command.CommandText)
+                    .AppendLine(batchTerminator);
```
bricelam: Any reason not to get this directly from `ISqlGenerationHelper` here?

bricelam: (Adding it to `RelationalDatabaseCreatorDependencies`)

ErikEJ: I actualky did this first, but got warnings about breaking changes due to constructor having been changed

smitpatel: @ErikEJ - Which ctor? ctor of `RelationalDatabaseCreatorDependencies` is fine to break.

ErikEJ: @smitpatel Yes, ctor of RelationalDatabaseCreatorDependencies  - I will break it then! 😄

ErikEJ: @smitpatel Broken!

PR #10141, opened 2017-10-23
Adds query filters sample.

anpete:


bricelam:
Do we want samples to build during CI? (If so, they should be added to `EFCore.sln`)

smitpatel:
Poor EFCore.sln 😞

samples/QueryFilters/Program.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        public DbSet<Blog> Blogs { get; set; }
+        public DbSet<Post> Posts { get; set; }
+
+        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
+        {
+            optionsBuilder
+                .UseSqlServer(
+                    @"Server=(localdb)\mssqllocaldb;Database=Demo.QueryFilters;Trusted_Connection=True;ConnectRetryCount=0;")
+                .UseLoggerFactory(new LoggerFactory().AddConsole((s, l) => l == LogLevel.Information && !s.EndsWith("Connection")));
```
ajcvickers: It's not a great idea to add a new LoggerFactory instance for every context instance since it causes the internal service provider to be rebuilt, but it doesn't matter too much here.

anpete: Thanks. This was copied from some old code, what is the preferred way to do this now?

ajcvickers: I'm not entirely sure. I took a note to go investigate the best way to do it now.

